PR #73, opened 2018-10-27
Add .pyc files to gitignore

svenseeberg:
* Also remove already commited .pyc files


svenseeberg:
Indeed, they were owned by root.

PR #76, opened 2018-10-28
Add global alerts and custom error pages

sascha11110:
[Django messages](https://docs.djangoproject.com/en/1.11/ref/contrib/messages/) can now be used on all pages that extend the _base template. The normal error pages for HTTP status codes and csrf errors are replaced.

### DIFF
--- a/backend/cms/templates/csrf_failure.html
+++ b/backend/cms/templates/csrf_failure.html
@@ -1,0 +1,15 @@
+{% extends "_raw.html" %}
+
+{% block raw_content %}
+<div id="no-session" class="px-4 py-10">
+    <div class="w-full max-w-xs mx-auto">
+        <h1 class="mb-4">CSRF Fehler</h1>
+        <p class="mb-4">
+            Bitte versuchen Sie die Seite neu zu laden.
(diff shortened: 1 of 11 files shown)


backend/cms/templates/csrf_failure.html
```diff
@@ -0,0 +1,15 @@
+{% extends "_raw.html" %}
+
+{% block raw_content %}
+<div id="no-session" class="px-4 py-10">
+    <div class="w-full max-w-xs mx-auto">
+        <h1 class="mb-4">CSRF Fehler</h1>
+        <p class="mb-4">
+            Bitte versuchen Sie die Seite neu zu laden.
```
svenseeberg: I would vote for directly using English language in source. And we should start with translatable strings.

backend/cms/templates/csrf_failure.html
```diff
@@ -0,0 +1,15 @@
...
+        </p>
+        <p>
+            <a href="/" class="text-grey-darker">Zur Startseite</a>
```
svenseeberg: Same.

PR #77, opened 2018-10-31
Refactoring of views

sascha11110:
- refactor views into multiple files
- introduce exemplary class-based views for pages and dashboard

My approach is to create a separate file for each view. For more complex views that process forms, I would use class-based views.


backend/cms/views/general/general.py
```diff
@@ -1,34 +1,29 @@
-from django.contrib.auth.decorators import login_required
 from django.shortcuts import render
 
 
-@login_required
-def index(request):
-    return render(request, 'dashboard.html')
-
 def handler400(request):
     ctx = {'code': 400, 'title': 'Bad request', 'message': 'Die Anfrage war fehlerhaft aufgebaut.'}
```
maxammann: I think these should be english by default. Maybe we can support i18n.

backend/cms/views/registration/forms.py
```diff
@@ -0,0 +1,27 @@
+from django import forms as django_forms
+from django.contrib.auth.forms import AuthenticationForm
+from django.contrib.auth.forms import PasswordResetForm as DjangoPasswordResetForm
+from django.contrib.auth.forms import SetPasswordForm
+
+
+class LoginForm(AuthenticationForm):
+    username = django_forms.CharField(widget=django_forms.TextInput(
+        attrs={'class': 'shadow appearance-none border rounded w-full py-2 px-3 text-grey-darker leading-tight focus:outline-none focus:shadow-outline', 'id': 'username', 'name': 'username', 'placeholder': 'Benutzername'}))
```
maxammann: These css classes should be defined in the templates imo and not in the view.

backend/cms/views/general/general.py
sascha11110: I will add i18n in a future pull request and I will also translate all hardcoded strings into English.

backend/cms/views/general/dashboard.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+@method_decorator(login_required, name='dispatch')
+class DashboardView(TemplateView):
+    template_name = 'general/dashboard.html'
+    base_context = {'current_menu_item': 'dashboard'}
+
+    def get(self, request):
+        val = 'To be defined'
+        return render(request, self.template_name, { 'key': val, **self.base_context })
```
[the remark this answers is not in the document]
sascha11110: Good point. Where would you put the linter?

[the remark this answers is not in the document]
maxammann: @svenseeberg said something about https://www.pylint.org/

PR #91, opened 2018-12-11
Jenkins integration with pylint

Schedulaar:
I added the Jenkins integreation together with pylint and created a default pylint config file.
I disabled the rules: missing-docstring and invalid-name, because there were too many errors involving these rules.
I adjusted our source files, so we have a build with no pylint warnings/errors.


svenseeberg:
Very cool!

PR #95, opened 2018-12-28
Feature/design models

Schedulaar:
Here is a first approach for the design of the models regarding sites, extras, pages, events, pois, push-notifications and their associations.

It's not finished yet, but I guess there's already a lot to discuss.

The event- and poi-models are taken from the feature/events branch. We should consider to use a pre-defined model for the events though (and only make our event-model a wrapper of that), since I guess this kind-of-work has already been done quite often (see https://djangopackages.org/grids/g/calendar/).

Also, I think my model of the extras-feature is very complex, but the most flexible (and cleanest) way to solve this in a relational way, but you can have a look yourselves^^

Any comments and suggestions are highly appreciated :D


backend/cms/models/language.py
```diff
@@ -0,0 +1,8 @@
+from django.db import models
+
+
+class Language(models.Model):
+    code = models.CharField(max_length=2, primary_key=True, validators=[lambda val: len(val) == 2])
```
Schedulaar: I'm aware that the validator doesn't work like this 😅

backend/cms/models/page.py
```diff
@@ -36,24 +40,25 @@ def get_tree_view(cls):
...
     )
-    status = models.CharField(max_length=10, choices=STATUS, default='draft')
     title = models.CharField(max_length=250)
+    status = models.CharField(max_length=10, choices=STATUS, default='draft')
     text = models.TextField()
-    permalink = models.CharField(max_length=60)
-    language = models.CharField(max_length=2)
+    language = models.ForeignKey(Language)
     version = models.PositiveIntegerField(default=0)
     active_version = models.BooleanField(default=False)
```
Schedulaar: what exactly does it mean for a version of a translation to be active? That it's the latest version? That it's the version that is going to be delivered via API? How does the status of a translation affect this?

backend/cms/models/poi.py
```diff
@@ -0,0 +1,42 @@
+from django.contrib.auth.models import User
+from django.db import models
+
+
+class POI(models.Model):
+    address = models.CharField(max_length=250)
```
Schedulaar: Whoops. Here's an association to a Site missing.

backend/cms/models/page.py
Ponitac: This isn't related to the translation status, but for the revision process. So we will have the same page with different revision states. But this field should only be true for one version, that is delivered via the API.

backend/cms/models/site.py
```diff
@@ -1,14 +1,35 @@
...
-    name = models.CharField(max_length=250)
-    slug = models.CharField(max_length=60)
+
+    title = models.CharField(max_length=200)
+    name = models.URLField(max_length=60, unique=True)
     status = models.CharField(max_length=4, choices=STATUS)
-    pub_date = models.DateTimeField(auto_now_add=True)
+    supported_languages = models.ManyToManyField(Language)
+
+    extras_enabled = models.BooleanField(default=True)
```
timoludwig: in my opinion, `extras_enabled` should not be a class-attribute of sites, but should be derived from the extras of the specific site. in other words, the API should return `true` if at least one extra is activated and `false` otherwise.

backend/cms/models/page.py
Schedulaar: Hmm... Not sure if I completely understood it: One page has several revisions and one revision (aka version) has several actual translations. Here, we have translations and revision packed into one model PageTranslation.
What happens if the active version has missing translations? Do we fallback to another version? If so, which one?

backend/cms/models/site.py
Schedulaar: Yes, we can do that.
`events_enabled` is appropriate though, right?

backend/cms/models/lehrstellenradar_extra.py
```diff
@@ -0,0 +1,25 @@
+from cms.models.extra import Extra
+
+
+class LehrstellenradarExtra(Extra):
```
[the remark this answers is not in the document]
Schedulaar: Yes, it may be a little bit over-engineered what I did here, but I still think it's appropriate :D
My model solves the problem, that we won't need to introduce strange attributes on the Site-model for determining what the extra should look like in the api (for example the wohnraumboerse-name, wohnraumboerse-url, wohnraumboerse-api, wohnraumboerse-thumbnail attributes in wp-cms just for one extra that is activated for two Sites), but have them defined in a relation table of wohnraumboerse-extra and Site. That way we don't pollute the Site model unnecessarily.

Also making them all just one unified extra would mean, that we have to hardcode their behaviour somewhere else (probably in the api then)...

What do the others think about that?

backend/cms/models/event.py
```diff
@@ -0,0 +1,69 @@
...
+
+class Event(models.Model):
+    site = models.ForeignKey(Site)
+    location = models.ForeignKey(POI, on_delete=models.PROTECT, null=True, blank=True)
+    start_date = models.DateTimeField()
+    end_date = models.DateTimeField()
+    picture = models.ImageField(null=True, blank=True, upload_to='events/%Y/%m/%d')
+    is_all_day = models.BooleanField(default=False)
+    is_recurring = models.BooleanField(default=False)
+    has_recurring_end_date = models.BooleanField(default=False)
```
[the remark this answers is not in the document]
Schedulaar: Yes, I think so. @florianhofhammer is that correct?
Should we use an already developed library for events/calendar stuff? We could have a look here: https://djangopackages.org/grids/g/calendar/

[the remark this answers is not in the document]
florianhofhammer: @timoludwig You're right, the model field `has_recurring_end_date` isn't necessary when using null in `recurring_end_date`. I'll remove this.
@Schedulaar I think using our own model could make our projekt sleeker. Only talking about the model, events aren't that hard to implement I think. I hope I don't overrate myself here regarding this topic :smile:

backend/cms/models/poi.py
florianhofhammer: As far as I'm not missing something the assiciations to a Site are missing in every model so far (Events, POIs, Pages).

backend/cms/models/page.py
sascha11110: > One page has several revisions

A page translation has several revision

> What happens if the active version has missing translations? Do we fallback to another version for the missing languages? If so, which one?

We should ensure that the active version is always a valid translation (if this is not the case, i.e. if there is a change to another language), then we inform the user about the need to update in the backend.

backend/cms/models/lehrstellenradar_extra.py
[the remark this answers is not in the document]
sascha11110: Fully agree

backend/cms/models/site.py
sascha11110: But we need to be able to determine which extras are activated. Am I getting something wrong?

timoludwig: What about something like `extras_enabled = models.ManyToManyField(Extra)`?

backend/cms/models/lehrstellenradar_extra.py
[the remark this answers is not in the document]
timoludwig: Ok, the wohnraumboerse is the only extra which would justify the division into all the extra subclasses. But I just came up with another idea how to solve this problem: What if we saved the two wohnraumboerse extras as two different extras (one called wohnraumboerse, the other one raumfrei) in the database and just activate one of them on the desired sites? This way we won't neither need the additional attributes in the site-model, nor the complex model-structure with hardcoded values. Additionally, it keeps the maintenance effort low because we can add/change/delete extras in the backend without changing the code.

backend/cms/models/page.py
Schedulaar: Hmm not sure what you mean by invalid translation... is it an invalid state, in which there's a language, that the Site supports, but doesn't have any active PageTranslation revision for that page?

Also: How can we track changes in the PageTranslations of the source language, so that we can inform the user that the other PageTranslations are outdated? Are the version numbers of the source-language PageTranslation transferred to the destination-languages PageTranslations once translated? And does a version-number increment on any change? Is there a possibility for the authors to change the text without changing the version? Do we always create a new PageTranslation object for every change?

backend/cms/models/poi.py
Schedulaar: Yeah, I've added the Site-associations in this PR, but forget about the POI model...

backend/cms/models/event.py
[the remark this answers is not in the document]
timoludwig: Yes, in WP we used the events-manager plugin.

backend/cms/models/lehrstellenradar_extra.py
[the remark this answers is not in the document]
timoludwig: Yeah, I share your concern that we don't want to have 10+ different extras for only a single site. So even if it is unlikely that we're going to have a widespread and configurable extra in the future, we need some kind of plan B what we do if this happens, right? What about an `ExtraSetting`-model which is associated to exactly one extra and one site? If I'm not mistaken, we could theoretically add this in the future without breaking changes to the existing extras, right?

backend/cms/models/site.py
timoludwig: If we follow the proposal of this PR, where all the extras are hardcoded in the models, then it makes sense to use an 1:n relation. However, if we prefer the solution i suggested [here](https://github.com/Integreat/cms-django/pull/95#discussion_r244511383), we should go for the n:n approach. Otherwise we would store the same extra information (name, alias, url, etc...) multiple times for every single site and if e.g. every site had its own lehrstellenradar extra and the lehrstellenradar url changed, we would have to change it for every site individually.

Schedulaar: @timoludwig Correct. Just mentioning, that in this current PR not *all* extras would be hardcoded models, but just those that have a special behavior (with per-site necessary settings). All the other extras are realized with manageable SimpleLinkExtras.

backend/cms/models/page.py
sascha11110: > Hmm not sure what you mean by invalid translation...

Sorry, my choice of words was quite confusing. Forget about the "invalid".

> Also: How can we track changes in the PageTranslations of the source language, so that we can inform the user that the other PageTranslations are outdated?

Good point. I think synchronizing the version numbers could become very complicated and error-prone. As in the current system we should mark one language as main language and save for every translation the version of the main language that the translation was derived from.

> And does a version-number increment on any change?

Yes, in my opinion.

> Is there a possibility for the authors to change the text without changing the version?

I think the version is just a internal state. But there should be the possibility for the user to change the page translation in the main language without demanding the other translations to adjust. This functionality is also available in the current CMS.
So we may need a field to mark the version (page translation) in the main language as a minor change that has no effect on the other translations.

> Do we always create a new PageTranslation object for every change?

I think so. Maybe we can implement the minor changes without the previously mentioned flag/field and change the current version directly. This would prevent the creation of a new PageTranslation object.

backend/cms/models/lehrstellenradar_extra.py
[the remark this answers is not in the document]
timoludwig: Hmm. I thought there would be a fancy way in django to merge the extras with the calling site's postal code when accessing it, but apparently there is not. So I think you were right in the first place and the cleaner model structure comes at the cost of some weird string replacements in the api or something similar. Sorry for the confusion! :sweat_smile: 
However, I'm still not 100% happy with the hardcoded urls, so I tried a more generic approach in the `feature/design-extra-models`-branch. Check it out if you're interested.

backend/cms/models/site.py
timoludwig: As stated above, I withdraw my proposal and agree on the 1:n relation of Site:Extra. So then we don't need the `extras_enabled` attribute, because we can just check if the `extra_set.all()` is empty, right?
`extras_enabled = Site.objects.get(id=site_id).extra_set.all() == []`

backend/cms/models/event.py
```diff
@@ -0,0 +1,69 @@
...
+    recurring_end_date = models.DateField(null=True, default=None, blank=True)
+    FREQUENCY = (
+        ('daily', 'Täglich'),
+        ('weekly', 'Wöchentlich'),
+        ('monthly', 'Monatlich'),
+        ('yearly', 'Jährlich')
+    )
+    frequency = models.CharField(max_length=7, choices=FREQUENCY, null=True, blank=True, default=None)
+
+    def clean(self):
```
svenseeberg: Just general note without need to change: I think "validate" is a better name for such functions? "clean" could implicate an state changing clean up method.

Schedulaar: That's a Django built-in function: https://docs.djangoproject.com/en/2.1/ref/models/instances/#validating-objects

backend/cms/models/page.py
svenseeberg: IMHO the model looks good so far. To clarify: "active" means that the text has been reviewed (whatever that means is up to the Verwalter / users). The latest "active" version is the page shown on the API / to the public. Correct?

svenseeberg: Ah and one other thing: do we need a "minor_edit" flag?

In Wordpress the minor edit flag for the main language implies that the translations are not outdated. This is used, for example, if a typo was fixed in the main language. Fixing typos does not (usually) invalidate translations.

svenseeberg:
Besides my 2 comments on the page translations, this is good to merge IMHO. It's probably not going to be the final state anyways ;-)

backend/cms/models/page.py
Schedulaar: > Ah and one other thing: do we need a "minor_edit" flag?

Either we add a `minor_edit` flag to the PageTranslation or we just change the current PageTranslation for minor edits (losing traceability and a proper history of the translation). I vote for the flag.

> To clarify: "active" means that the text has been reviewed (whatever that means is up to the Verwalter / users). The latest "active" version is the page shown on the API / to the public. Correct?

No. The `active` flag is true, if (and only if) the PageTranslation should be delivered via API. Thus I consider renaming it to a `public` flag.
The `active` (`public`) flag can only be true for one PageTranslation per page & language.
(In formula):
equation

The status of a PageTranslation on the other hand is only there for the revision process. Therefore I consider renaming the status choices from (draft, review, public) to (draft, in-review, reviewed).

Schedulaar:
@svenseeberg Yeah, sure, it's never the final state, but refactoring on the models will become more expensive the further we are with development of the functionalities...
If there won't be any other responses/reactions/requests I would merge this on tuesday

PR #106, opened 2019-01-29
Feature/legacy api

Schedulaar:
i've made some (further) significant changes to the models to add support for a language tree selection per site (to configure source/target languages) and "live-content" (mirroring) of pages.
There's a problem regarding the permalink, you might have an idea to solve it...

Other than that, I've implemented two endpoints: the sites and the site/languages endpoints.


backend/cms/models/page.py
```diff
@@ -40,13 +42,23 @@ def get_tree_view(cls):
...
+
+
+class MirrorPage(Page):
+    mirrored_page = models.ForeignKey(ContentPage)
+
 
 class PageTranslation(models.Model):
-    page = models.ForeignKey(Page, related_name='page_translations')
+    page = models.ForeignKey(ContentPage, related_name='page_translations')
     permalink = models.CharField(max_length=60)
```
Schedulaar: When mirroring a page we only have the permalink of the mirrored pageTranslation. We need to think about a way to generate permalinks for the mirror page as well...

sascha11110: Some possibilities:
1) Instead of mirroring the pages (by specifying foreign keys), we copy the page and then synchronize it afterwards.
2) Introduce further relations that assign the permalinks to the individual mirrorings.
3) Save only the slug of the permalink, keep it the same for all mirrorings and make the whole permalink distinguishable by the other components (like /augsburg/).

svenseeberg:
Let's merge this?! @Schedulaar

PR #111, opened 2019-03-13
Feature/feedback model

Thomas125:
This includes the PR of @Schedulaar about legacy-api.


Thomas125:
@timoludwig 
About the text comment field: This field already exists. It's called "text".

PR #115, opened 2019-03-22
WIP: Content creation (pages)

sascha11110:


backend/cms/models/page.py
```diff
@@ -86,10 +95,14 @@ class PageTranslation(models.Model):
     text = models.TextField()
     language = models.ForeignKey(Language, on_delete=models.CASCADE)
     version = models.PositiveIntegerField(default=0)
-    public = models.BooleanField(default=False)
+    choices = (
+        (True, 'Öffentlich'),
+        (False, 'Privat')
+    )
```
Schedulaar: `choices` is not an appropriate name. A boolean field `public` is just fine imo.

backend/cms/models/site.py
```diff
@@ -43,6 +43,18 @@ class Site(models.Model):
     matomo_token = models.CharField(max_length=150, blank=True, default='')
     matomo_ssl_verify = models.BooleanField(default=True)
 
+    @property
+    def languages(self):
+        language_tree_nodes = self.language_tree_nodes.prefetch_related('language').all()
+        languages = []
+        for language_tree_node in language_tree_nodes:
+            languages.append(language_tree_node.language)
+        return languages
```
timoludwig: Does anybody know how to solve this without a loop?
(And without importing `Language`, because this causes an import conflict)

backend/cms/templates/pages/tree_row.html
```diff
@@ -5,25 +5,22 @@
...
     </td>
     <td>
-        <a href="{% url 'edit_page' page_translation_id=node.id site_slug=site.slug %}" class="block py-3 px-2 text-grey-darkest">
-            {{ node }}
+        <a href="{% url 'edit_page' page_id=node.id site_slug=site.slug language_code=language.code %}" class="block py-3 px-2 text-grey-darkest">
+	        {% for page_translation in node.page_translations.all %}
+		        {% if page_translation.language == language %}
+			        {{ page_translation.title }}
+		        {% endif %}
+	        {% endfor %}
```
timoludwig: We should probably write a custom template filter to get the title in the current language.
Or has anybody a better way of solving this?

backend/cms/models/site.py
svenseeberg: Can you provide some more insight what you want to achieve by removing the loop? Performance? A list comprehension is still some sort of a loop:
````
return [language_tree_node.language for language_tree_node in language_tree_nodes]
````

Comment (Schedulaar) [reply backend/cms/models/site.py]: Actually I'd remove the `Site` foreignKey on `LanguageTreeNode` and only have a [1, 0..1] Site-LanguageTreeNodeRelation (introducing a `LanguageTreeNode` foreignKey on the `Site` model referencing the root of the tree) and then I believe you could do sth like `self.language_tree_root.get_family()`

Comment (timoludwig) [reply backend/cms/models/site.py]: @svenseeberg This was the first time I needed a loop like this in Django, so I assumed I'm doing it wrong. Usually, there are inbuilt shortcuts which are both shorter and easy to read, so I was hoping for something like `Language.objects.filter(language_tree_nodes__in=self.language_tree_nodes.all()` (this would work, but needs an import of `Language`). But I don't think, performance is an issue so like your proposal ;)

@Schedulaar `self.language_tree_root.get_family()` is essentially the same as `self.language_tree_nodes`. I went for this approach to make sure that every node belongs to exactly one site. Otherwise, some nodes may have no or multiple sites (without complicated validation). I think every site should have its own language tree to make sure they can make changes to their tree without affecting other sites (even if they have the same tree as another site in the beginning). Plus, this gives us the bonus of `LanguageTreeNode` being a Site-Language-Model with the attribute [`active`](https://github.com/Integreat/cms-django/blob/3029f82564d15a059898f605791947087b4b1708/backend/cms/models/language.py#L64), which can be used to add an inactive language to the tree (which allows pages to be created, but is not delivered via the api). But I'm open to your proposal as well, if you think it fits better to our use case.

Comment (Schedulaar) [reply backend/cms/models/site.py]: > I think every site should have its own language tree to make sure they can make changes to their tree without affecting other sites

Same for me :)

> Otherwise, some nodes may have no or multiple sites (without complicated validation).

Okay, true. With your approach, on the other hand, 
*  (1) nodes in the same tree might refer to different sites or
*  (2) there could be multiple trees for the same site

(without complicated validation), creating inconsistencies.

I think both approaches are fine, but we should make sure that inconsistencies cannot occur (so we need some validation at some point :) )

For my approach you could make the node-ForeignKey on the `Site` model *unique* and then add the custom validation, that either the `parent`-property or the `related-site`-property is null on the `LanguageTreeNode` model (such that only the root has the `related-site` property set).

In your approach this might be possible by
* (1) checking for all nodes that its family has the same `site` set on all nodes and
* (2) checking that all `LanguageTreeNode`s of a Site have the same [root](https://django-mptt.github.io/django-mptt/models.html#get-root).

File: backend/cms/views/language_tree/language_tree_node_form.py [hunk 4]
```diff
@@ -43,7 +44,24 @@ def save_page(self, site_slug, language_tree_node_id=None):
...
                 language=self.cleaned_data['language'],
-                site=Site.objects.get(slug=site_slug),
+                site=self.site,
                 active=self.cleaned_data['active'],
                 parent=self.cleaned_data['parent'],
             )
+
+        return language_tree_node
+
+    def clean(self):
```
timoludwig: @Schedulaar Do you think this appropriate for validating the language tree?

Issue #114, opened 2019-03-21
Multisite domains


svenseeberg:
This issue is about discussing the "multisite" support in our new CMS.

1) We should consider the naming. According to our naming convention we should probably name this "multi region support".

2) During our team call we discussed if this should be implemented with subdomains or as paths (https://[REGION].integreat-app.de/ or https://cms.integreat-app.de/[REGION]/). There is no clear preference yet. However, to implement APIv3 we need support the path option.

3) We discussed to have the API available under https://api.integreat.app. Is this still something we want to have? This would influence the discussion somewhat.

timoludwig:
We have implemented the second approach (`https://cms.integreat-app.de/[REGION]/`) and I don't see an urgent reason to add subdomains as well. If this pops up in the future, we can re-open this issue.

Issue #42, opened 2018-10-25
Admin: region management


svenseeberg:
- [x] add
- [x] change
- [x] hide
- [ ] delete
regions

timoludwig:
I think, this issue is to broad for one issue. I outsourced extra management to this issue: #160

PR #120, opened 2019-04-05
support multiple regions, i18n for backend

timoludwig:
Adds the basic functionality for multiple regions/sites.

Changes to models:
- add `Page.order`
- rename `Site.title` -> `Site.name`
- rename `Site.name` -> `Site.slug`

Changes to views and templates:
- added region overview
- added forms to add and edit regions
- divided the admin-layout into two different "stages": one for super-admins to edit regions, languages, users etc. and one for the region-manager to edit all the content etc.

connected to the following two issues: #42, #114 (but does not completely resolve them)


backend/cms/models/page.py
```diff
@@ -25,6 +25,7 @@ class Page(MPTTModel):
     icon = models.ImageField(blank=True,
                              null=True,
                              upload_to='pages/%Y/%m/%d')
+    order = models.IntegerField(null=True)
```
Schedulaar: I'm not sure about that, but it could be possible, that the MPTTModel already saves information about the order of nodes.

backend/cms/models/site.py
```diff
@@ -22,14 +22,14 @@ class Site(models.Model):
...
+    name = models.CharField(max_length=200)
+    slug = models.CharField(max_length=200, unique=True)
     status = models.CharField(max_length=4, choices=STATUS)
-    language_tree = models.ForeignKey(LanguageTree, null=True, on_delete=models.SET_NULL)
+    languages = models.ManyToManyField(Language)
 
     events_enabled = models.BooleanField(default=True)
     push_notifications_enabled = models.BooleanField(default=True)
-    push_notification_channels = ArrayField(models.CharField(max_length=60))
+    push_notification_channels = ArrayField(models.CharField(max_length=60), blank=True)
```
Schedulaar: why should this be blank? Imo it should always have some value, maybe default to empty array? :thinking:

backend/cms/models/site.py
```diff
@@ -22,14 +22,14 @@ class Site(models.Model):
         (ARCHIVED, 'Archived'),
     )
 
-    title = models.CharField(max_length=200)
-    name = models.URLField(max_length=60, unique=True)
+    name = models.CharField(max_length=200)
+    slug = models.CharField(max_length=200, unique=True)
     status = models.CharField(max_length=4, choices=STATUS)
-    language_tree = models.ForeignKey(LanguageTree, null=True, on_delete=models.SET_NULL)
+    languages = models.ManyToManyField(Language)
```
Schedulaar: Should we be able to enable / disable only a few languages of a Site? For example during the time in which a Site is working on a new language, it shouldn't be delivered to the API; what do you expect this `languages` field to look like in this situation? Should it include the new language?

Schedulaar: Also, since you're removing the tree structure: How do you save the information in which direction the pages should be translated? Afaik you said on the conference, that this should be remembered per page. Although the most common (or even the only) use case is that all pages of one site are translated in the same direction.

backend/cms/models/site.py
timoludwig: I didn't want to add any complex UI elements to the region form at this stage, therefore I just went with a string-representation of the array and the user can just input different channels separated by spaces. The `RegionForm` then parses the string with a `.split(" ")`, which results on an empty array if the string is empty. The `blank`-parameter is not database-related, but just handles the form-validation instead. At some point at the future we should probably add an UI element to add and remove array elements in the backend form, but I don't think this has a high priority.

Schedulaar: Alright, I was on a wrong track somehow; I thought, `blank=true` allows the field to be null, but probably it only allows it to be an empty array, right?

backend/cms/models/page.py
timoludwig: You're right! If I understood it right, we could use the fields `lft` or `rgt` of the MPTT model for ordering, as they should both represent the horizontal ordering. @sascha11110 can you confirm this?

backend/cms/models/site.py
timoludwig: Yes, I think only `null=True` allows the field to be actually `null` in the database.

backend/cms/models/site.py
timoludwig: tbh I didn't think of the case that they are working on a language which should not be published yet. The solution I think of is that we allow any language in the content creation, but deliver only the languages of this field in the api. If this isn't a good approach, we have to add another option as `api_languages` or so.

Concerning the translation direction:
As far as I understood, the XLIFF-files already contain the information, which language they're coming from. But you're right, in this case we'd need another field `translation_origin` or similar in the `PageTranslation` model. @sascha11110 @svenseeberg @Ponitac what do you think, would you prefer the language tree model?

backend/cms/models/page.py
timoludwig: Actually, it seems as we don't need to access these order fields directly (it would probably be a bad idea to mess with the tree database manually anyway), we can just use the `move_to(target, position)` method to change the order.

backend/cms/models/site.py
svenseeberg: Having an option to disable the listing of a language sounds useful to me. On the other hand, this issue did not come up yet.

And yes, I agree with Timo that we should store the translation origin on a per page basis. The XLIFF contains information about the source language.

timoludwig: The current workflow is as follows: When a new language is added, it is delivered via the API, but their translations are not public. This way, the language will appear in the WebApp, but the user can not change the language in the language switcher. Instead, the message "Translation not available" is shown when hovering over the language. I think this approach is also perfectly fine in the Django cms, so we don't need an additional field here.

Apart from that, I added `source_language` and `currently_in_translation` to the page translation model. @Schedulaar do you agree with this?

Schedulaar: > [...] I think this approach is also perfectly fine in the Django cms, so we don't need an additional field here.

Well, it's not exactly the same -- the language will be shown (and selectable) after selecting the city for example and then there would be no content in that site-language pair (unpleasant)...
Also these kind of things can only be introduced with a major model-change (like introducing a site-language-model and having a property `live: boolean` on that). That is true for any property a Site may want to set per language (an imaginable scenario imo). Changing a model once there's data in the DB is always a huge pain in the a** (if we want some kind of per-language-properties we need to do complex db-migrations), whereas adding a property to a model is very easy (if we chose now to add a language-site-model or similar). So we should be sure whether we should introduce sth like that now already (also balancing flexibility vs complexity in models).

> Apart from that, I added source_language [...] to the page translation

Well, I introduced the not-per-page-but-per-site-source-language(-tree) (:D) as a result of a discussion with @dkehne:
Some city had started translating their German pages to several destination pages (all sourced on the German language) manually on their own, but messed up with the translation sources (the sources weren't all set to German, but to mixed languages on mixed pages).
At some point they decided to have their instance translated by tolingo and then the export of the XLIFF was very complicated (or even faulty) because their database was kind-of messed up.
Also - as far as I remember - @dkehne stated, that the only use-case is that all pages of one language in a site have the same source-language. (correct me if I'm wrong) Therefore saving the source-language per page would be duplicated data. 

So this was my attempt to prevent this kind of mess.
Also this is just an explanation of my action, I don't really know what to recommend here.

timoludwig: Thanks for your detailed feedback and the good ideas. It took me until now to finish the changes, could you have a look at the updated models and check if they are fine now? Thanks in advance!

README.md
```diff
@@ -28,4 +28,11 @@
 * `docker stop $(docker ps -a -q)`
 * `docker rm $(docker ps -a -q)`
 * remove all images: `docker rmi $(docker images -a -q)`
-* remove all volumes: `docker volume prune`
\ No newline at end of file
+* remove all volumes: `docker volume prune`
+
+### i18n
+To make use of the translated backend, comile the django.po file as follows:
```
Thomas125:
```suggestion
To make use of the translated backend, compile the django.po file as follows:
```

backend/cms/models/page.py
```diff
@@ -75,21 +86,37 @@ class PageTranslation(models.Model):
...
+        ('draft', _('Draft')),
+        ('in-review', _('Pending Review')),
+        ('reviewed', _('Finished Review')),
     )
     title = models.CharField(max_length=250)
     status = models.CharField(max_length=9, choices=STATUS, default='draft')
     text = models.TextField()
-    language = models.ForeignKey(Language, on_delete=models.CASCADE)
+    language = models.ForeignKey(Language, related_name='pages', on_delete=models.CASCADE)
+    source_language = models.ForeignKey(
```
Schedulaar: Was it on purpose to add the language tree but keep the source_language property?

timoludwig: No sorry, this doesn't make sense anymore. Thanks for pointing that out.

PR #121, opened 2019-04-05
add multi-language support for backend (resolves #112)

### LINKED ISSUE #112: Admin: Multi-language support
Enable i18n support for the admin backend, to make it available in german and english. Should be extendable to other languages as well.

timoludwig:
- I converted all strings in the templates to translation tags (in english)
- I added a .po file for the german translation of all strings
- Resolves #112


backend/backend/settings.py
```diff
@@ -114,7 +115,16 @@
 # Internationalization
 # https://docs.djangoproject.com/en/1.11/topics/i18n/
 
-LANGUAGE_CODE = 'en-us'
+LANGUAGES = (
+    ('en', 'English'),
+    ('de', 'Deutsch'),
```
svenseeberg: We had a discussion about language labels. We discussed if we want to use the 2 or 5 character format ('en' vs 'en-us'). As far as I remember, we said that 2 characters are not sufficient. I tried to find the conversation in Slack without success.

timoludwig: This only handles the translation of the backend and does not involve the content. I thought we agreed on supporting only English and German on the mid-term and maybe adding a few more languages in the future.
But it's nearly no effort to change this to 5 chars just to be completely future-proof.

svenseeberg: Alright, then my comment might not be that relevant. One other question: does it make sense to integrate the content translations and back end translations somehow? This means, for example, that content languages are "the same" as back end interface languages (and not decoupled).

Schedulaar: I wouldn't recommend merging them; there's just no benefit as far as i can see and might add confusion.

Also we agreed on the bcp47 specification for content languages, which allows quite a lot of formats (to be future proof, if we want to support region-codes and / or "einfache Sprache" ): https://tools.ietf.org/html/bcp47

timoludwig: Yes, I agree that we should not mix the two levels up - we don't want to handle the content's translations in .po files and sometimes we want to edit an english page and keep the backend in german...

PR #127, opened 2019-05-05
[README] Update Django commands in Docker container

svenseeberg:
README reflects the sourced Python3 venv in the bashrc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ cd cms-django
 A docker-compose file is provided in the the repository. It will start one container with a PostgreSQL database and another one with the CMS.
 * `docker-compose up`
 * enter [http://localhost:8000](http://localhost:8000)
-* as long as there is no standard SQL dump, you have to create your own user: `docker exec -it $(docker-compose ps -q django) bash -c "integreat-cms createsuperuser"`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,9 +16,9 @@
 
 ### Migrations
 * change models
-* `docker exec -it $(docker-compose ps -q django) bash -c "python3 manage.py makemigrations [app]"`
-* optional, if you want to inspect the corresponding SQL syntax: `docker exec -it $(docker-compose ps -q django) bash -c "python3 manage.py sqlmigrate [app] [number]"`
-* `docker exec -it $(docker-compose ps -q django) bash -c "python3 manage.py migrate"`
+* `docker exec -it $(docker-compose ps -q django) bash -c "integreat-cms makemigrations [app]"`
```
timoludwig: The command `integreat-cms` doesn't work with docker exec, because `.bashrc` isn't sourced in this case (resulting in an `bash: integreat-cms: command not found` error). So either you get into the container shell first, end then execute `integreat-cms`, or you call docker exec with `.venv/bin/integreat-cms`.

svenseeberg:
True, I missed one bash command. I also removed the installation of dependencies from the requirements.txt

PR #128, opened 2019-05-06
[On hold ] Adjust Wiki for container dev

Ponitac:
Using the repo inside the docker container without .venv/bin/integreat-cms would result in a "command not found" error. In my opinion we should either activate venv by default or change the commands to ".venv/bin/integreat-cms".


svenseeberg:
Yes, this is definitely one option. The other one is sourcing before executing the command. However, the shortest option seems to be adding a `-i` parameter. With that, the bashrc will be sourced. Then it would be `bash -i -c integreat-cms`

svenseeberg:
I updated PR #127 accordingly.

Ponitac:
Agreed

PR #135, opened 2019-06-06
Xliff import and export

chenpengxiang:


chenpengxiang:
Add features:
1. html to xliff converter
2. xliff to html converter
3. export page translation to xliff file
4. import xliff file to page translation

backend/cms/page_xliff_converter.py
```diff
@@ -0,0 +1,312 @@
...
+
+        except Exception as ex:
+            pass
+
+        return result
+
+    def import_xliff_file(self, file_path, user):
+        result = False
+        if file_path.startswith(XLIFFS_DIR) and file_path.endswith(('.xlf', '.xliff')) and os.path.isfile(file_path):
+            with open(file_path, 'r') as f:
```
timoludwig: Same issue as above, in my docker environment the file gets opened in ascii encoding, which fails. Probably we should use  `open(file_path, 'r', encoding='utf-8')` here.

```
======================================================================
ERROR: test_import_xliff_file_translated (cms.tests.PageXliffHelperTest)
  File "/opt/integreat-cms/backend/cms/tests.py", line 540, in test_import_xliff_file_translated
    result = self.page_xliff_helper.import_xliff_file(file_path, self.user)
  File "/opt/integreat-cms/backend/cms/page_xliff_converter.py", line 292, in import_xliff_file
    xliff_content = f.read()
  File "/usr/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 282: ordinal not in range(128)
```

PR #138, opened 2019-06-06
#136 + use config from database and logging

JayeshS:
This branch builds on #136 and adds db-configuration and logging

### DIFF
--- a/backend/cms/models/configuration.py
+++ b/backend/cms/models/configuration.py
@@ -3,6 +3,6 @@ from django.db import models
 
 class Configuration(models.Model):
     key = models.CharField(max_length=100, unique=True, blank=False)
-    value = models.CharField(max_length=1000, unique=True, blank=False)
+    value = models.CharField(max_length=1000, blank=False)
     created_date = models.DateTimeField(auto_now_add=True)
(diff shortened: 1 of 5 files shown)


backend/cms/models/configuration.py
```diff
@@ -0,0 +1,8 @@
+from django.db import models
+
+
+class Configuration(models.Model):
```
svenseeberg: @timoludwig do you have any opinion on this model?

backend/cms/models/configuration.py
```diff
@@ -0,0 +1,8 @@
...
+    key = models.CharField(max_length=100, unique=True, blank=False)
+    value = models.CharField(max_length=1000, unique=True, blank=False)
```
svenseeberg: Is the unique flag really required for the value?

JayeshS: Oops.  :) Will fix.

backend/cms/models/configuration.py
timoludwig: I think it has a very limited use case (as we store all non-sensitive config in the settings.py), but for storing keys it's fine.

Issue #18, opened 2018-10-25
API: languages endpoint


svenseeberg:
Endpoint to get all languages for a site. This must be Integreat API v3 compatible.

https://github.com/Integreat/cms/wiki/REST-APIv3-Documentation#languages

timoludwig:
This issue was closed in #140.

Issue #16, opened 2018-10-25
API: sites endpoint


svenseeberg:
Endpoint of available sites. This must be compatible to Integreat API v3.

https://github.com/Integreat/cms/wiki/REST-APIv3-Documentation#locationsmultisitesinstances

timoludwig:
This issue was closed in #140.

PR #140, opened 2019-06-06
API: sites, languages

yanfeix:
- Django #16, sites
- Django #18, languages

### DIFF
--- a/backend/api/v3/sites.py
+++ b/backend/api/v3/sites.py
@@ -10,38 +10,64 @@ PREFIXES = [
     'Stadt',
 ]
 
+def strip_prefix(name):
+    for p in PREFIXES:
+        if name.startswith(p):
+            return p, name[len(p) + 1:]  # +1 for one whitespace
+    return None, name
+
+def transform_site(s):
+    prefix, name_without_prefix = strip_prefix(s.name)
+    return {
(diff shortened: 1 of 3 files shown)


backend/api/v3/sites.py
```diff
@@ -10,38 +10,64 @@
...
-            'events': s.events_enabled,
-            'push-notifications': s.push_notifications_enabled,
-            'longitude': s.longitude,
-            'langitude': s.latitude,
-            'aliases': None  # todo
-        }
+def transform_site(s):
+    prefix, name_without_prefix = strip_prefix(s.name)
+    return {
+        'id': s.slug,
```
timoludwig: just a minor detail: this sould be `s.id`.

backend/api/v3/sites.py
```diff
@@ -10,38 +10,64 @@
...
+        'push-notifications': s.push_notifications_enabled,
+        'longitude': s.longitude,
+        'langitude': s.latitude,
+        'aliases': None  # todo
+    }
+
+def transform_site_by_status(s):
+    prefix, name_without_prefix = strip_prefix(s.name)
+    return {
+        'id': s.slug,
```
timoludwig: here as well `s.id`.

PR #143, opened 2019-06-07
Fix migration issue

imnnquy:
Move the step init headers out of `__init__` method to make sure the migration successfully


imnnquy:
As @jshere mentioned here: https://github.com/Integreat/cms-django/pull/138#commitcomment-33833298
Closing this PR

PR #146, opened 2019-06-07
media library

RoyJia:
## Issue

https://github.com/Integreat/cms-django/issues/48

#### CURRENTLY:

 - So far, we are able to use django-filer with admin account and with its own UI(different with integreat) - http://localhost:8888/admin/filer/
 - But it could be config to other user by the Django authorzation system.

#### TO DO:

 - 1. Move django-filer to user interface form Django admin. Roy: I tried but I dont find a way to integrate it, may people know, who is more expertise for Django. 
 - 2. Fit django-filer UI to same with cms-django
 

 
#### CURRENTLY:

 - I created Media Library page
 - Able to upload/view/download/delete file, files is stored in `MEDIA_URL = '/media/'` and file info is stored with Document Model
 - Util method `save_file` is created for all files saving, because the file info would be saved to database

#### REVIEW:

 - "enable replacement of files (update files without changing download URL)" - could implement this with edit funtion  - A
 - "add alt text to enhance accessibility (probably as required field)" - since I did not differentiate file type, so I setup rel instead alt in some table cell
 - "security checks fo uploaded files" - C
 - "Nice to have feature: data deduplication.." - D
 - "Uploaded files should not immediately be publicly available" - How it should be? - E
 - "Scaling of images for the apps" - could be after support thumbnail - B
 - "Optimazation, for PNG lossless, for example with optipng, for JPEG with defined compression levels, i.e. 85." - F

#### TO DO:

 - A. Implement eidt
	 - issue: when update, new medel instance will be added
 - B. To support thumbnail
 - C. To support security checks fo uploaded files
 - D. Deduplication
 - E
 - F. Optimazation
 - G. To support pagination
 - H. To supoort search(within Media Library)
 - I. To support order by name/create_date...


RoyJia:
Hi @svenseeberg I'm looking at the conflict, is it okay after you merge it without fixing conflict

svenseeberg:
I fixed them :)

PR #148, opened 2019-06-12
Security fixes, refactoring of page form

### LINKED ISSUE #52: Pages: Make permalinks editable
Permalinks should be editable. The old permalinks must still be saved and delivered via the API.

Related to: https://github.com/Integreat/cms-django/issues/119

timoludwig:
### Upgrade Django from 1.11.18 to 1.11.20
This fixes CVE-2019-6975
### Update pyxdg from 0.25 to 0.26
This fixes CVE-2019-12761
### Improve prune docker tool
- Change the scope of the dev_tools/prune_docker.sh
- Delete only the django image and not all docker images
### Fix Integreat Logo
Remove css filter which alienates the Integreat logo
### Fix migration of push notification sender
Move init function to send to make sure the configuration is not used before it exists in the database
### Refactor page form
- Page form model: Split page form into two forms
    - PageForm for all page attributes
    - PageTranslationForm for all page translation attributes
    - Add slug field to form (fixes #52 make permalinks editable)
- Page models:
    - Add editors (prepares permissions on specific page instances)
    - Add publishers (also for page-instance-specific permissions)
    - Make the __str__ function more informative for logging
    - Improve the function for translation permalinks
    - Move the slug-generation-function into model
- Page views:
    - Make the page views more readable
    - Adapt the views to the new structure of the forms
    - Adapt side by side view to new forms
- Page templates:
    - Add page filters for translation information in page list
    - Add id, creator, last_updated and created_date to page list


backend/cms/views/pages/page_form.py
```diff
@@ -19,109 +39,155 @@ class PageForm(forms.ModelForm):
...
+
+        page_translation = super(PageTranslationForm, self).save(*args, **kwargs)
+
+        if not self.instance.id:
+            # only update these values when page translation is created
+            page_translation.page = page
+            page_translation.creator = user
+            page_translation.language = language
+
+        if not self.cleaned_data['slug']:
```
timoludwig: One thing I noticed here:
If the slug is present and changed, we have to check if it is unique per region and language.

backend/cms/models/page.py
```diff
@@ -142,13 +166,38 @@ class PageTranslation(models.Model):
...
+        ])
+
     @property
     def permalink(self):
-        permalink = self.page.site.slug + '/'
-        permalink += self.language.code + '/'
-        for ancestor in self.page.get_ancestors(include_self=True):
-            permalink += ancestor.page_translations.get(language=self.language).slug + '/'
-        return permalink
+        return '{}/{}/{}/{}'.format(
```
svenseeberg: We can define the permalink without the ancestors in the path. That would make things considerably easier, if the structure changes.

timoludwig: Hmm. I think the advantage of having informative URLs which tell me in which category the page lives outweighs the effort of managing multi-level permalinks. The hard part is to create a consistent database of permalinks, it shouldn't make much of a difference if we have to additionally store the ancestor path there. So I'd prefer having the ancestor path in the permalink...

backend/cms/templates/pages/tree_row.html
```diff
@@ -6,9 +6,12 @@
             <i data-feather="move" class="text-grey-darkest"></i>
         </span>
     </td>
+    <td>
+        {{ node.id }}
+    </td>
     <td>
         <a href="{% url 'edit_page' page_id=node.id site_slug=site.slug language_code=language.code %}" class="block py-3 px-2 text-grey-darkest">
```
Ponitac: We should remove px-2 due to alignment.

backend/cms/views/pages/page_form.py
timoludwig: resolved in the new version of the commit.

backend/cms/templates/pages/tree_row.html
timoludwig: resolved in the new version of the commit.

PR #149, opened 2019-06-12
User management, user-based access control

### LINKED ISSUE #44: Admin: manage users
Create, edit and delete users with e-mail, login, password, site access, role assignments.

Assign users to sites.

See also #45.

### LINKED ISSUE #56: Site selection
A user who has access to several sites must be able to switch between these instances at any time (without logging in again).

### LINKED ISSUE #59: User management
In accordance with #44 a similar user administration must also be available in the end customer backend.

### LINKED ISSUE #66: Cross link to admin panel
To facilitate the use for superadmins, we should provide crosslinks to the admin panel at the right places.

timoludwig:
### User management
- Add user model, views and templates
- Add organization model, views and templates
- Create, edit and delete superusers and staff (fixes #44)
- Create, edit and delete region users (fixes #59)
- Show region switcher based on user's permissions (fixes #56)
- Show link to network management if user is superuser or staff (fixes #66)
### User-based access control for admin area and regions
- Add @staff_required for admin area
- Add @region_permission_required for regions
- Decorate all views with the required function
### Page creation
- Fixed a small bug where the clean_slug method accessed fields of a form instance which didn't exist yet


backend/cms/templates/_base.html
```diff
@@ -30,15 +30,19 @@
 	        {% endif %}
         </div>
         <div id="instance-selector-list" class="absolute shadow rounded-b">
-	        {% for other_region in sites %}
-	            <a href="{% url 'dashboard' site_slug=other_region.slug %}" class="block px-4 py-3 text-grey-darkest hover:bg-grey">
-	                {{ other_region.name }}
-	            </a>
+	        {% for other_region in user.profile.regions.all %}
+		        {% if not other_region == site %}
```
florianhofhammer: Why is the current site excluded here? Did I miss a convention we agreed on or does this have technical reasons?

timoludwig: This is only the region selector in the frontend. The user is already operating on the current site (which is displayed in the div above), thus we don't have to provide a link to go to the current site.

florianhofhammer: Ok, I just pulled the corresponding branch and tested it myself. Now I get the reason, I just wasn't aware where this div is displayed and thought of a (complete) list in the body of the webpage. Thanks for the clarification!

florianhofhammer:
The new pages don't get activated via base_context dictionary yet and therefore don't get highlighted in the navigation sidebar. Apart from this, this PR looks good to me!

timoludwig:
@florianhofhammer thanks for your feedback! For me, new pages highlighted the right menu item, but for region_users it didn't work. Did you mean this? If yes, it should be fixed now.

florianhofhammer:
@timoludwig By "new pages" I meant region_users, roles, organizations, etc. pp. You fixed one of these, but the other ones still don't get highlighted in the navigation sidebar. As I already stated, this is no functional issue. I approve this merge request in regard of the new functionality, but we should either fix these display errors before merging or at least note this issue so that we don't forget to fix it later. Imho, this should be fixed before merging.

timoludwig:
Now I got it, thanks! I just fixed all the remaining menu items.

florianhofhammer:
Works like a charm! Thanks for the changes!

PR #151, opened 2019-06-12
Page instance permissions

### LINKED ISSUE #55: Entity based permissions
A few municipalities grant access rights to individual instances of an entity. For example, user A may only edit the pages X, Y and Z.

timoludwig:
### Add rule-based permissions for page instances (fixes #55)
- Add rules package to enable function-based permissions
- Add instance-specific permission rules for pages
- Add rule based permission checks for pages
- Add AJAX-based UI to manage instance-based permission for pages
### Add role fixtures
- Add possibility to populate the database with fixed values
- Add fixture file for user roles
- Apply fixtures on database migrations
#### This is a follow-up PR on #150 (which should be merged before), so the real diff which should be reviewed here is:
https://github.com/Integreat/cms-django/compare/feature/access-control...feature/page-instance-permissions

### DIFF
--- a/backend/backend/settings.py
+++ b/backend/backend/settings.py
@@ -99,6 +99,11 @@ DATABASES = {
     }
 }
 
+# Directory for initial database contents
+
+FIXTURE_DIRS = (
+    os.path.join(BASE_DIR, 'cms/fixtures/'),
+)
 
 # Authentication backends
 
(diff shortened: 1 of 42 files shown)


svenseeberg:
I just realized that this PR encompasses the previous ones as well. I did not read 100% of the code in this PR, but large parts and also played around in the running CMS. I'm confident enough that everything is fine and support merging this.

timoludwig:
Thanks! If your approval also covers the previous pull requests, can you approve them as well? This way, I can merge them in the correct order and the history is clean.

PR #165, opened 2019-06-28
Add Exceptions & Error messages

Ponitac:
Add Try-Catch block to prevent errors, if the api parameter aren't set correctly.


backend/cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -41,7 +45,10 @@ msgstr "Review abgeschlossen"
 msgid "Active"
 msgstr "Aktiv"
 
+<<<<<<< HEAD
```
timoludwig: you have to resolve those conflicts, these are no valid lines for po files

backend/cms/templates/_base.html
```diff
@@ -106,18 +106,26 @@
 	            <i data-feather="grid" class="absolute"></i>
 		        {% trans 'My Dashboard' %}
 	        </a>
+<<<<<<< HEAD
```
timoludwig: also, here are unsolved conflicts.

backend/cms/views/statistics/statistics.py
```diff
@@ -40,23 +44,29 @@ def prepare_csv(languages, hits, dates):
...
-                                                          period=request.GET.get('peri', 'day'),
-                                                          lang=lang[0])
-            temp_hits = []
+            try:
+                api_hits = api_man.get_visitors_per_timerange(date_string=start_date + ',' + end_date,
+                                                              site_id="2",
+                                                              period=request.GET.get('peri', 'day'),
+                                                              lang=lang[0])
+                temp_hits = []
+            except ConnectionError:
```
timoludwig: In addition to `ConnectionError`, I also received `InvalidURL` (if no matomo_url is defined) and `TypeError`(if https://statistics.integreat-app.de is given but the key is wrong or the region is not configured properly). So probably, we should also catch these errors.
However, we also could hide the statistics menu item if the url is not set for a specific region (or `statistics_enabled` is `false`).

Ponitac: Extra Try-Catch statements have been added

backend/cms/views/statistics/matomo_api_manager.py
```diff
@@ -50,21 +51,6 @@ def cleanmatomo_url(self):
         elif not bool(re.match("^https://", self.matomo_url)):  # check for "https://" and set it
             self.matomo_url = self.protocol + self.matomo_url
 
-    def api_request(self, method, curl):
```
timoludwig: This function is used several times in the api manager: line 80, line 90, line 101 and line 125.
If you delete it, you also have to delete the other functions (`create_instance`, `get_all_site_ids`, `get_all_sites_id_name` and `get_all_sites`). But I don't know if we need them in the future or not.

Ponitac: I removed the functions. These are all legacy code, that might come in handy, if we implement a new process for the creation of an instance. But at the moment they are obsolete.

Ponitac:
I additionally added an error message, to catch the case if a user didn't select a correct start or end date.

PR #177, opened 2019-08-03
Add pylint runner and make code pylint conform (fixes #167)

### LINKED ISSUE #167: Include pylint_runner in Jenkinsfile
- [x] Activate pylint for the Jenkins build again
- [x] Increase max line length to ~160 chars
- [x] Convert current code base to pylint conform code

timoludwig:
- Add pylint_runner to the Jenkins Pipeline again
- Make the code conform to pylint again
- Ignore pylint warnings on false positives 
- Relaxed pylint parameters:
    - max line length: 100 -> 160

Fixes #167


backend/cms/templatetags/page_filters.py
```diff
@@ -24,7 +24,7 @@ def page_translation_creator(page, language):
     page_translation = all_page_translations.filter(language__code=language.code)
     if page_translation.exists():
         return page_translation.first().creator
-    elif all_page_translations.exists():
+    if all_page_translations.exists():
```
svenseeberg: This is more than a code style change. Is this intended?

timoludwig: Since there is a ´return´ in the if-block, it will never execute the code below if the first condition is true. And if the first condition is false, then it makes no difference whether we have ´elif´ or just ´if´. So it is indeed just a matter of code style, and our pylint config doesn't like unnecessary ´elif´ or ´else´ statements...

backend/cms/views/pages/archive.py
```diff
@@ -4,7 +4,7 @@
 
 from django.contrib.auth.decorators import login_required
 from django.utils.decorators import method_decorator
-from django.shortcuts import render
+from django.shortcuts import render, redirect
```
Ponitac: Do we need this import as it seems to be working without it as well?

backend/.pylintrc
```diff
@@ -505,7 +505,7 @@ known-third-party=enchant
 [DESIGN]
 
 # Maximum number of arguments for function / method.
-max-args=5
+max-args=7
```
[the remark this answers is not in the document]
timoludwig: Because of [this line](https://github.com/Integreat/cms-django/blob/b3dc29f49535d143712115b394e1ad711d8a5e3b/backend/cms/views/push_notifications/push_notification_sender.py#L17), but probably it is better to leave the config as it is and just add an ignore-comment it in this line...

backend/cms/views/pages/archive.py
timoludwig: `redirect` is used in [this line](https://github.com/Integreat/cms-django/blob/b3dc29f49535d143712115b394e1ad711d8a5e3b/backend/cms/views/pages/archive.py#L33).

backend/.pylintrc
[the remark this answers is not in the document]
Ponitac: I guess we should leave it with 5 args and as you said ignore the few excecptions

backend/cms/views/pages/archive.py
Ponitac: But the code worked earlier as well, without the import. I have no problem with leaving it this way. I am just interested why we need the import now, but not earlier

timoludwig: If you look at the code, you can see that redirect is only executed if no language parameter is given and the site has a default language. So probably we just never tested this special case.

Ponitac: Ahh okay, I see. Thanks pyint :100: :grinning:

Ponitac:
I guess you can merge the PR now, if you want to

PR #179, opened 2019-08-05
POI management

### LINKED ISSUE #54: POI management
Locations should be management in a central way.

It should be discussed whether we can combine the locations for the **events** with the **POIs**. @Ponitac what do you think about that?

timoludwig:
- **Manage POIs**
	- Add views and templates for POI management
	- Fixes #54
- Refactor general views
	- Break all general views into the following submodules:
		- dashboard
		- error_handler
		- settings
		- utils
	- Move confirmation popups into the directories they belong
	- Prepare unique slug function for re-utilization in other views
-  Refactor dev-tools
	- Fix directory for venv
	- Add docker settings for use with postgres container


backend/cms/views/pois/poi_form.py
```diff
@@ -0,0 +1,126 @@
...
+            kwargs
+        )
+
+        # pop kwarg to make sure the super class does not get this param
+        poi = kwargs.pop('poi', None)
+        user = kwargs.pop('user', None)
+
+        if not self.instance.id:
+            # don't commit saving of ModelForm, because required fields are still missing
+            kwargs['commit'] = False
```
Ponitac: Did you add this, because this field isn't defined as mandatory by our model?

dev-tools/run.sh
```diff
@@ -1,15 +1,20 @@
 #!/bin/sh
```
Ponitac:
```suggestion
#!/bin/bash
```

Ponitac: As described in #180 we can secure the correct execution of the scripts.

dev-tools/prune_database.sh
```diff
@@ -1,6 +1,5 @@
 #!/bin/sh
```
Ponitac:
```suggestion
#!/bin/bash
```

dev-tools/migrate.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/sh
```
Ponitac:
```suggestion
#!/bin/bash
```

dev-tools/create_superuser.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/sh
```
Ponitac:
```suggestion
#!/bin/bash
```

backend/cms/views/pois/poi.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+@login_required
+@region_permission_required
+@permission_required('cms.manage_pois', raise_exception=True)
+def view_poi(request, poi_id, region_slug, language_code):
+    template_name = 'pois/poi_view.html'
+    poi = POI.objects.get(id=poi_id)
+
+    poi_translation = poi.get_translation(language_code)
```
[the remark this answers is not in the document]
timoludwig: Yes, good idea! I added the check.

backend/cms/templates/pois/poi.html
```diff
@@ -0,0 +1,138 @@
...
+        </div>
+    </div>
+
+    <div class="flex flex-wrap">
+        <div class="w-2/3 pr-2">
+            {{poi_form.errors}}
+            {{poi_translation_form.errors}}
+	        <div class="w-full p-4 mb-4 rounded border border-solid border-grey-light shadow bg-white">
+		        <label class="block mb-2 font-bold">{% trans 'Title' %}</label>
+		        {% trans 'Insert title here' as title_placeholder%}
```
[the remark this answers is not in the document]
timoludwig: Oh sorry, forgot to update the translation file. Should be fixed now.

backend/cms/templatetags/poi_filters.py
```diff
@@ -0,0 +1,55 @@
+from django import template
+
+register = template.Library()
+
+
+@register.filter
+def poi_translation_title(poi, language):
+    all_poi_translations = poi.translations
+    poi_translation = all_poi_translations.filter(language__code=language.code)
+    if poi_translation.exists():
```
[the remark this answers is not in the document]
timoludwig: Actually, I added this check only for debugging purposes. In theory, there should never exist a POI without any POI translation. So I think the order of this checks is ok for our use case.

backend/cms/views/pois/poi.py
```diff
@@ -0,0 +1,187 @@
+"""
+
+Returns:
+    [type]: [description]
```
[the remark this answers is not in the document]
timoludwig: Yeah, I think this comment is pretty useless :P

backend/cms/views/pois/poi.py
```diff
@@ -0,0 +1,187 @@
...
+        # TODO: error handling
+        if poi_form.is_valid() and poi_translation_form.is_valid():
+
+            poi = poi_form.save()
+            poi_translation = poi_translation_form.save(
+                poi=poi,
+                user=request.user,
+            )
+
+            if poi_form.has_changed() or poi_translation_form.has_changed():
```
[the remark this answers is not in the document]
timoludwig: Yes, good idea. If nothing changed, we don't need to save anything. I moved the check before the `save()`.

backend/cms/views/pois/poi_form.py
timoludwig: No, I add this because region is a required field, but at this time it is not part of the form model instance (because the region is not included in the html form, instead it is deducted from the currently active region). But Django refuses to save an object if a required field is still missing. But with `commit=false` however, it works. After adding the region, I can then finally save the object.

backend/cms/views/pois/poi.py
```diff
@@ -0,0 +1,187 @@
+"""
+
```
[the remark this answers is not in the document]
timoludwig: Thanks, I adapted the suggestion.

Issue #156, opened 2019-06-28
PageTranslations: Revisions


timoludwig:
We need a versioning mechanism for page translations. When changing a page, we should not change the content of the page translation instance, but create a new object with a new version number.
For each page and language, only the page translation with the highest version number should be delivered by the API. In case the highest version number has "draft" or "review required" as status, we take the highest version number which is "finished review".

This is part of #104.

PR #186, opened 2019-09-11
Add versioned PageTranslations

svenseeberg:
This PR tries to solve #156. It adds versions to PageTranslations with the help of the django-reversions app.

* Modify the PageTranslations model.
* Add Reversion to Page and PageTranslation models

image


timoludwig:
I pushed my changes just in case we continue with this approach despite its drawbacks.

timoludwig:
We concluded not to use the module "reversion".

PR #193, opened 2019-09-30
Version PageTranslations

svenseeberg:


svenseeberg:
@timoludwig @Ponitac do these changes make sense?

backend/cms/views/pages/page_form.py
```diff
@@ -210,10 +209,7 @@ def save(self, *args, **kwargs):
         page = kwargs.pop('page', None)
         user = kwargs.pop('user', None)
 
-        if not self.instance.id:
-            # don't commit saving of ModelForm, because required fields are still missing
-            kwargs['commit'] = False
-
+        kwargs['commit'] = False  # Don't save yet. We just want the object.
```
svenseeberg: @timoludwig can this be a problem in some special cases?

backend/cms/views/pages/page_form.py
```diff
@@ -222,9 +218,14 @@ def save(self, *args, **kwargs):
             page_translation.creator = user
             page_translation.language = self.language
 
+        page_translation.version = page_translation.version + 1
+        page_translation.id = None
```
svenseeberg: Instead of setting the `id = None`, we can alternatively call `save(force_insert=True)`

backend/cms/models/page.py
```diff
@@ -66,7 +66,8 @@ def languages(self):
 
     def get_translation(self, language_code):
         try:
-            page_translation = self.page_translations.get(language__code=language_code)
+            page_translation = self.page_translations.filter(
+                language__code=language_code).order_by('-version')[0]
```
svenseeberg: Do we prefer `.order_by('-version')[0]` or `.order_by('-version').first()` ?

florianhofhammer: I personally prefer the latter

backend/cms/views/pages/page_form.py
timoludwig: Yes, using `save(force_insert=True)` sounds good.

backend/cms/models/page.py
timoludwig: I agree with Flo.

PR #194, opened 2019-10-01
Add attach page / live content, fixes #13

### LINKED ISSUE #13: Pages: Attach content
Attach other pages from the same or other sites at the beginning or end of a page.

The other, attached pages should be shown while editing but not be editable.

Use case: different sites share common text templates which need to be amended by local addresses.

svenseeberg:
Change page selection to Django forms

### DIFF
--- a/backend/cms/views/pages/page.py
+++ b/backend/cms/views/pages/page.py
@@ -13,13 +13,12 @@ from django.contrib import messages
 from django.contrib.auth.decorators import login_required, permission_required
 from django.contrib.auth import get_user_model
 from django.contrib.auth.mixins import PermissionRequiredMixin
-from django.http import Http404
+from django.http import Http404, JsonResponse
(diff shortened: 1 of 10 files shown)


backend/cms/views/pages/page.py
```diff
@@ -508,3 +507,32 @@ def revoke_page_permission_ajax(request):
...
+    result = {}
+    for page in pages:
+        result[page.id] = ' -> '.join([page_iter.best_language_title() for page_iter in page.get_ancestors(include_self=True)])
+    return JsonResponse(result)
+
+    def label_from_instance(self, page):
+        return ' -> '.join([
+            page.get_first_translation([self.language.code]).title
+            for page in page.get_ancestors(include_self=True)
+        ])
```
Thomas125: No longer needed

PR #196, opened 2019-10-09
Extra & ExtraTemplate management

### LINKED ISSUE #159: Admin backend: Manage extra templates
In the network admin area, admins should be able to create extra templates which can be used for the extra objects in regions.
This involves all attributes of the `ExtraTemplate` model.

This issue is part of #99

timoludwig:
- Small changes to the extra & extra_template model to improve the handling of postal codes
- Add all views and html templates for extra templates
- Add basic extra templates as fixtures to be imported together with test data
- Add possiblity to activate and deactivate extras per region


timoludwig:
In addition to the global extra templates, I extended the PR to the management of extras per region.

PR #201, opened 2019-10-13
Improve dev-tools for docker users

timoludwig:
- Simplify checks for docker container
- Comment all important lines for better understanding
- Reset translation file if only the date changed
- Remove default environment variable (causes error for gettext and Django doesn't need it for default settings)


timoludwig:
I just fixed the script to work for users who don't have the german locale configured on their machines.

PR #205, opened 2019-10-17
Separate page and translation attributes in page form (fixes #154)

### LINKED ISSUE #154: Pages: Separate page and translation attributes in page form
At the moment, it is not clear which attributes of the page form refer to the `Page` object and which only handle `PageTranslation` objects (from a user's point of view). Probably, we should separate these two sections from each other.

This is part of #98.

timoludwig:
I redesigned the page form to point out which attributes belong to a translation and which ones belong to the entire page. I used a tab view for every language.
This is a completely new design approach which differs from the other pages, so if you have ideas how to implement the requirement more similar to the other forms or have other suggestions regarding the positioning or color choices, feel free to reach out to me.

Screenshot_2019-10-17 Integreat CMS

### DIFF
--- a/backend/cms/templates/pages/page.html
+++ b/backend/cms/templates/pages/page.html
@@ -7,8 +7,8 @@
 {% load rules %}
 <form method="post">
     {% csrf_token %}
-    <div class="flex flex-wrap mb-4">
-        <div class="w-2/5 flex flex-wrap flex-col justify-center">
+    <div class="flex flex-wrap">
+        <div class="w-4/5 flex flex-wrap flex-col justify-center mb-6">
             <h2 class="heading font-normal">
                 {% if page %}
                     {% if page_translation_form.instance.id %}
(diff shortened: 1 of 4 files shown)


backend/cms/templates/pages/page.html
```diff
@@ -22,195 +22,182 @@ <h2 class="heading font-normal">
...
+						        {% endif %}
+					        </div>
+				        {% endif %}
+			        {% endif %}
+		        </div>
+		        </div>
+	        </div>
         </div>
     </div>
 </form>
```
svenseeberg: In the beginning this file uses spaces for indentation, later tabs. I think we should use spaces everywhere because we're dong this in the Python code by default. Or is there a good reason to use tabs?

timoludwig: Thanks for pointing that out! Yes, I think mixing them both makes no sense.

PR #207, opened 2019-10-23
Translation status icons

### LINKED ISSUE #83: Translation: Status icons
Available icons should be:

* [x] no translation exists
* [ ] translation outdated
* [x] translation up to date
* [ ] translation in progress

timoludwig:
- Improve the translation models
- Add a few helper functions to the page translation model
- Add status icons to the page translation tree view

fixes #83

### DIFF
--- a/backend/cms/models/event.py
+++ b/backend/cms/models/event.py
@@ -170,12 +170,16 @@ class EventTranslation(models.Model):
     """
     Database object representing an event tranlsation
     """
+    DRAFT = 'DRAFT'
+    REVIEW_PENDING = 'PENDING'
+    REVIEW_FINISHED = 'FINISHED'
+
     STATUS = (
-        ('draft', 'Entwurf'),
-        ('in-review', 'Ausstehender Review'),
-        ('reviewed', 'Review abgeschlossen'),
+        (DRAFT, 'Entwurf'),
+        (REVIEW_PENDING, 'Ausstehender Review'),
+        (REVIEW_FINISHED, 'Review abgeschlossen'),
     )
-    status = models.CharField(max_length=9, choices=STATUS, default='draft')
+    status = models.CharField(max_length=9, choices=STATUS, default=DRAFT)
     title = models.CharField(max_length=250)
     description = models.TextField()
     permalink = models.CharField(max_length=60)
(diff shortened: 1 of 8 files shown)


backend/cms/models/event.py
```diff
@@ -170,12 +170,16 @@ class EventTranslation(models.Model):
     """
     Database object representing an event tranlsation
     """
+    DRAFT = 'DRAFT'
+    REVIEW_PENDING = 'PENDING'
+    REVIEW_FINISHED = 'FINISHED'
+
     STATUS = (
-        ('draft', 'Entwurf'),
```
timoludwig: Using constants here is more flexible and  less error-prone than inline strings

backend/cms/models/page.py
```diff
@@ -67,14 +67,14 @@ def languages(self):
     def get_translation(self, language_code):
         return self.page_translations.filter(
             language__code=language_code
-        ).order_by('-version').first()
```
timoludwig: We don't need this ordering anymore, because I addded the default order to the meta class

backend/cms/models/event.py
dkehne: agree on that.

backend/cms/models/page.py
```diff
@@ -200,8 +203,64 @@ def available_languages(self):
...
+        # If the source translation is outdated, this translation can not be up to date
+        if source_translation.is_outdated:
+            return True
+        self_revision = self.latest_major_public_revision
+        source_revision = source_translation.latest_major_public_revision
+        # If on of the translations has no major public revision, it cannot be outdated
+        if not self_revision or not source_revision:
+            return False
+        return self_revision.last_updated < source_revision.last_updated
+
```
dkehne: I don't get the idea here. Maybe someone else has to have a look at that.

backend/cms/models/page.py
svenseeberg: Good to know.

backend/cms/models/page.py
svenseeberg: I find this very easy to understand :-D

dkehne: this is because you like linux, open source and astronomy....

timoludwig: Isn't this the target group of our repository? :telescope:

PR #210, opened 2019-10-27
Duplicate regions

### LINKED ISSUE #43: Admin: duplicate regions
Duplicate a site with all its content except for users.

timoludwig:
Add the possibility to import content from other regions when creating a new region.
The duplication process of the language tree and the pages was a bit tricky because they both have the tree structure and the parents have to be adjusted as well. I solved this with a recursive function which I tried to document a bit more detailed than usually.

Fixes #43

Other implicit changes of this PR:
- Fixed a small bug of the page-status in the test data
- Fixed the CSRF handler (we had still the old view in the settings)


timoludwig:
Sorry, I fixed the pylint error now. The PR is ready for review :wink:

PR #211, opened 2019-10-28
Enhance page view and form

### LINKED ISSUE #82: Pages: Display title of backend language for translations
On translations, show the title of the main language. This is necessary when the admin is not speaking the specific language but needs to edit other settings.

### LINKED ISSUE #157: Pages: make read-only status visible
When a user is not allowed to change a page, make all form fields disabled and show a message for the user which explains the behaviour. E.g., explain that he does not have the required permissions and should send a message to the verwalter in case he needs it.

This is part of #98.

timoludwig:
- If content and backend language differ, I show two columns in the page view. In the first column there is the page translation title in the content language, in the second column in the backend language. This makes it easier for content creators to manage translations in unknown languages. (Fixes #82)
- If a user has the permissions to view all pages, but not to edit them, the page form gets disabled and a small notification is shown. Thus, it is clear whether or not someone can edit a page and why. (Fixes #157)
- Additionally, I fixed a small bug where the template for new pages could not be rendered. This was due to the tab view and the links to different translation of a not yet existing page. I solved it by showing only one tab when creating a page, and all additional tabs when editing a page.

### DIFF
--- a/backend/cms/templates/pages/tree.html
+++ b/backend/cms/templates/pages/tree.html
@@ -3,6 +3,7 @@
 {% load static %}
 {% load mptt_tags %}
 {% load page_filters %}
+{% load rules %}
 
 {% block content %}
 <div class="table-header">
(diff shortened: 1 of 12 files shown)


backend/cms/templates/pages/tree.html
```diff
@@ -54,26 +63,31 @@ <h2 class="heading font-normal mb-2">{% trans 'Pages' %}</h2>
...
                 <th class="text-sm text-left uppercase py-3 px-2">
                     <div class="lang-grid flags whitespace-no-wrap">
-	                    {% for lang in languages %}
-		                    <a href="{% url 'pages' region_slug=region.slug language_code=lang.code %}">
-			                    <img src="{% static '' %}images/flags/{{ lang.code }}.png" title="{{ lang.name }}" />
-		                    </a>
-	                    {% endfor %}
+                        {% for lang in languages %}
+                            <a href="{% url 'pages' region_slug=region.slug language_code=lang.code %}">
+                                <img src="{% static '' %}images/flags/{{ lang.code }}.png" title="{{ lang.name }}" />
```
Ponitac: I would recommend to add a language tooltip at this point

Ponitac:
```suggestion
                                <img src="{% static '' %}images/flags/{{ lang.code }}.png" title="{{ lang.name }}" />
                                <span title="{{ lang.translated_name }}">
```

Ponitac:
```suggestion
                                <span title="{{ lang.translated_name }}">
                                <img src="{% static '' %}images/flags/{{ lang.code }}.png" />
                                />
                                
                                
```

timoludwig: Thanks, I updated my commit with the working tooltip.

PR #212, opened 2019-11-02
Page review workflow, region model, pylint, package updates

### LINKED ISSUE #166: Make Page Permissions activable for each page
User shouldn't be able to see the personalized permissions for a page. Therefore a checkbox should appear at each page with an option.

### LINKED ISSUE #185: Add region prefixes
Regions not only have names. As they are different kinds of entities they need prefixes like 'Landkreis' (county) or 'Stadt' (city). This needs to be considered in the regions model.

### LINKED ISSUE #195: PageTranslations: Review workflow
We need a revision workflow for page translations. 
For each page and language, only the page translation with the highest version number should be delivered by the API. In case the highest version number has "draft" or "review required" as status, we take the highest version number which is "finished review".

Special cases we have to consider:
- In case an untrusted user changes a page translation, the review status (field `status` of `PageTranslation`) should not be manually changeable, but changes automatically to "draft".
- Untrusted users should have the possibility to click on a "submit for review"-button to set the status to "Pending Review". This way, users can make small changes without directly publishing them and notifying a verwalter.

Important:
If only the status of a page translation is changed (and no content etc.), we shouldn't create a new version. Otherwise, this could be misleading: e.g. if I change the status of version 3 to finished and after saving, the version 3 is still open for review and instead a new version 4 is now finished...

This issue also involves to merge the attributes `status` and `public` to one  set of choices. 
E.g.
```
    STATUS = (
        ('draft', _('Draft')),
        ('in-review', _('Pending Review')),
        ('public', _('Public')),
    )
```
This issue does not cover the User Interface. The UI is covered in #188.

This is part of #104.

### LINKED ISSUE #214: Add region aliases
In the new app, we will have a more powerful region-search which includes nearby cities etc.
This should be configurable by region aliases, an array of strings which return a specific region as result. Hence, we need a new database field for this and a possibility to manage these in the region form.

timoludwig:
- Refactor class constants
    - Create new module for global constants
    - Import constants in models and views
- Improve pylint config
    - Add foreign key related names to generated-members
    - Improve speed by using 2 threads
    - Remove unused directories from ignore
    - Change ignored error codes to strings
- Add translations and pylint to dev-tools
    - translations.sh re-generates and compiles the translation file
    - pylint.sh runs pylint_runner in the correct directory
- Add review workflow for page translations (fixes #195)
    - Users who can edit pages but not publish them have the possibility to save a page as draft or submit it for review
    - Users who can edit and publish pages can review changes and publish them
    - The current status of a page translation is shown in the page tree view
- Allow unicode characters in slugs
    - Adapt SlugField in models
    - Add correct unicode slugs to test data
- Add page permission setting for each region (fixes #166)
    - Add page_permissions_enabled to region model
    - Hide the permission settings in the page form if option is disabled
    - Raise permission denied on ajax requests if option is disabled
- Adapt to new pylint-django version
    - Make code valid for pylint-django version 2.0.12
    - Ignore unused arguments warnings
- Add administrative divisions of regions (fixes #185)
    - Add new administrative_division field to region model and form
    - Add all possible choices for regions in Germany
    - Show administrative division in region list
- Upgrade Django to version 2.2.7
    - Fix version of Django to 2.2.7
    - Remove fixed version of pylint
    - Adapt code to new version of pylint
- Add region aliases (fixes #214)
    - Add aliases field to region model and form
    - Adapt test data


timoludwig:
Sorry for all the mess, I won't add any commits anymore! :sweat_smile: 
Unfortunately, I cannot split this into separate PRs, because the pylint package got updated and the Jenkins build would fail on the earlier commits now.

Issue #221, opened 2019-11-16
Enable drag and drop in the language tree


timoludwig:
After implementing the drag and drop process for the page tree, we could use the same solution for the language tree.

Issue #50, opened 2018-10-25
Pages: New approach to control the arrangement of pages


sascha11110:
Currently there are two ways to determine the arrangement of the pages:
1) Tree view: Drag & Drop
2) Detailed view: Determining the order and the parent page

In particular, defining the order is not satisfactory, as the numbers are always assigned at different intervals.

My suggestion:
- define the parent page via select field
- then drag & drop between the existing children's pages

Are there any better suggestions?

timoludwig:
I like this approach!

Schedulaar:
Do we only implement one of the two? I'd prefer if we had both a global drag-and-droppable tree view and the option you suggested in the detailed view

timoludwig:
Yes, we provide both options. The drag & drop in the tree view is covered in this issue: #10.
Probably we can use the same drag & drop code snippets for both.

svenseeberg:
I'm not totally convinced yet. This makes changing the full tree very tedious.

Is there a reason to not support moving pages to other parent pages via drag & drop?

timoludwig:
@svenseeberg I think this is a misunderstanding... We DO want to support drag & drop in both the page tree view and the single page form.

svenseeberg:
Thanks for clarifying.

PR #223, opened 2019-11-17
Drag and drop in page tree

### LINKED ISSUE #10: Pages: Overview as tree
Create an overview of all pages in a tree.

This should include:
- [x] tree view
- [x] Drag & Drop page to other location in tree
- [x] Revision status
- [x] Translation status
- [x] Exclude archived pages
- [x] show only pages of current site

timoludwig:
This enables drag and drop to change the order and structure of the page tree. I implemented this from scratch without any additional libraries.

When a drag starts, the table is expanding to enable pages to be moved between other pages. When dragging over a region where the page can be dropped, the hover effect is a blue border. On pages where a move is not possible (e.g. because a page cannot be a child of itself or cannot be a sibling of one of its childs), the hover effect is red.

Please play around with it and tell me wether this is intuitive to use and works as expected.

If this is serving our purpose, I will also add this feature to the page ordering in single pages (#50) and the language tree (#221)

page-tree-drag-drop
Fixes #10


backend/cms/urls.py
```diff
@@ -173,6 +173,7 @@
                 url(r'^$', pages.PageTreeView.as_view(), name='pages'),
                 url(r'^new$', pages.PageView.as_view(), name='new_page'),
                 url(r'^upload$', pages.upload_page, name='upload_page'),
+                url(r'^upload$', pages.upload_page, name='upload_page'),
```
timoludwig: Oh, this was just an unintended line duplication. Will remove it before I merge.

ulliholtgrave:
Thank you Sascha, would love to hear that every day from you <3

PR #232, opened 2019-11-22
Fix usage of sudo command in dev tool

timoludwig:
Fix dev tools for esoterik operating systems like ubuntu which don't have a current version of npm in their repositories.


jnugh:
This works for me :)

PR #233, opened 2019-11-22
Feature/tinymce ui

### LINKED ISSUE #70: TinyMCE adjustments
Compared to the current editor, the following features are to be removed:
- "View" / "Ansicht"
- "abc"
- "File > new document" / "Datei > Neues Dokument"
- "Edit > Insert as text" / "Bearbeiten > Als Text einfügen"
- "Insert > Page break" / "Einfügen > Seitenumbruch"
- "Format > strike through, super-/subscript, code" / "Format > Durchgestrichen, Hochgestellt, Tiefgestellt, Code"

The Wordpress design of the editor should be retained.

Currently the menu as well as the icons are used. Should we unify this? Further suggestions for improvement?

ulliholtgrave:
Add icon buttons to TinyMCE.

Closes #70


backend/cms/templates/pages/page.html
```diff
@@ -241,11 +241,80 @@ <h2 class="heading font-normal">
...
+        language: '{{LANGUAGE_CODE|slice:"0:2"}}',
+        setup: (editor) => {
+            editor.ui.registry.addButton('notranslate', 
+            {
+                tooltip: '{% trans 'Do not translate the selected text.' %}',
+                icon: 'permanent-pen',
+                onAction: () => {
+                editor.focus();
+                val = tinymce.activeEditor.dom.getAttrib(tinyMCE.activeEditor.selection.getNode(), "translate", "yes");
+                console.log(editor.selection.getContent().length);
```
Thomas125: Remove console.log

Thomas125:
The SourceCode-Option is available via View -> SourceCode and Tools -> SourceCode.
Is there a special reason that it is in two places?

Thomas125:
The paragraph drop-down menu and the Format drop-down contain inline option which are redundant since there are special buttons for underline, bold text etc.
And it contains all four align-options, but I think we only wanted center, to avoid problems with ltr and rtl languages.

PR #244, opened 2019-11-23
Use objects.filter().first() instead of objects.get()

### LINKED ISSUE #243: Don't use objects.get in best_language_title()
Reproduce bug:
There must be at least one page without english translation. Then edit a page in the english backend. => Error

Reason:
To show the page in the page mirroring drop down menu, the root language of the region is used. A language node of the region is chosen by `LanguageTreeNode.objects.get(region__id=self.region.id)`.
But get() apparently only works if there is excactly one matching object. So it would only work if the region had only one language at all.

Thomas125:
Resolves #243

### DIFF
--- a/backend/cms/models/page.py
+++ b/backend/cms/models/page.py
@@ -11,7 +11,7 @@ from django.utils.translation import get_language
 
 from mptt.models import MPTTModel, TreeForeignKey
 
-from .language import Language, LanguageTreeNode
+from .language import Language
 from .region import Region
 from ..constants import status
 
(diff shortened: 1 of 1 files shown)


backend/cms/models/page.py
```diff
@@ -143,7 +143,7 @@ def get_tree(cls, region_slug, archived=False):
     def best_language_title(self):
         page_translation = self.page_translations.filter(language__code=get_language())
         if not page_translation:
-            alt_code = LanguageTreeNode.objects.get(region__id=self.region.id).get_root().language.code
+            alt_code = LanguageTreeNode.objects.filter(region__id=self.region.id).first().get_root().language.code
```
timoludwig: Sure this works, but in this case the shortcut `alt_code = self.region.default_language.code` would be even better.

Thomas125: Yeah, I've just seen this in ulli's pr. :laughing:

PR #248, opened 2019-11-27
Implement 2FA auth using webauthn

### LINKED ISSUE #131: 2 Factor Authentication
Implement 2FA via TOTP and, if possible, also U2F.

jnugh:
Add a settings view, allowing a user to register multiple keys and remove those.
Changing MFA settings is only possible if the user (re-)verified his identity (using a password) within the last 5 minutes.
Question is: Should 2FA be checked before changes are allowed to be made?
I don't see a thread model where this would help. The user already proved his identity by logging in in the first place and the password check is only there if someone gains access to the already logged in session. Asking for the password again makes sure that the attacker could not gain persistent access. If the password was leaked he would need the key to log in in the first place.
Asking for 2FA to edit 2FA settings would only help against an attacker that has the valid password and got access to a valid session somehow.

Add a new login step in case there is a key to verify.

Fixes #131 

This is my first PR, so there is most likely a lot of things that don't align with your standards and as this is a quite large PR as well I expect a lot of comments on this :P
I did not fix lining errors yet, as pylint_runner shows me a ton on unrelated files as well

### DIFF
--- a/backend/cms/locale/de/LC_MESSAGES/django.po
+++ b/backend/cms/locale/de/LC_MESSAGES/django.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2019-11-23 13:30+0000\n"
+"POT-Creation-Date: 2019-11-30 11:40+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: Timo Ludwig <ludwig@integreat-app.de>\n"
 "Language-Team:\n"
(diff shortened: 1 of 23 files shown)


backend/cms/templates/settings/user.html
```diff
@@ -0,0 +1,37 @@
+{% extends "_base.html" %}
+{% load i18n %}
+
+{% block content %}
+{% load static %}
+
+
+<h2 class="heading font-normal mb-2">{% trans 'User settings' %}</h2>
+<h3>{% trans 'Registered 2FA key' %}</h3>
```
svenseeberg: I'm wondering if we should name this "Registered FIDO/FIDO2 2-factor-authentication keys". Opinions? But on the other hand this can be explained in the documentation.

backend/cms/models/user_mfa.py
```diff
@@ -0,0 +1,23 @@
...
+    public_key = models.BinaryField(max_length=255, null=False)
+    sign_count = models.IntegerField(null=False)
+
+    def __str__(self):
+        return self.user.username + str(self.key_id)
+
+    class Meta:
+        default_permissions = ()
+        constraints = [
+            models.UniqueConstraint(fields=['user', 'name'], name='unique_key_nickname')
```
timoludwig: In another model we used `unique_together = (('user', 'name'))`, so maybe we should stay consistent to this (though it's just a matter of preference)...

backend/cms/templates/registration/login_mfa.html
```diff
@@ -0,0 +1,141 @@
...
+                    'Content-Type': 'application/json',
+                    'X-CSRFToken': window.django_csrf_token
+                },
+                body: JSON.stringify(transformedAssertionForServer)
+            });
+            const data = await result.json();
+            if(data.success) {
+                location.href = "/"
+            } else {
+                document.querySelector('.auth-error').classList.remove('hidden');
```
timoludwig: If you want to, you can use [umbrellaJS](https://umbrellajs.com/documentation), but vanilla JS is just fine :wink:

backend/cms/templates/settings/mfa/add_key.html
```diff
@@ -0,0 +1,137 @@
...
+
+{% block javascript_nocompress %}
+<script>
+    window.django_csrf_token = '{{ csrf_token }}';
+</script>
+{% endblock %}
+{% block javascript %}
+<script>
+    // Based on https://github.com/duo-labs/py_webauthn/blob/master/flask_demo/static/js/webauthn.js
+    function b64enc(arrayBuffer) {
```
timoludwig: Some JS code also seems to be duplicated, so if you want to you could outsource this code fragments to dedicated files and include them in the templates whenever necessary.

jnugh:
`dev-tools/pylint.sh` should probably be added to the readme file (this is where I looked) before trying to run `pylint_runner` which I found in the Jenkinsfile :smile:

backend/cms/templates/registration/login_mfa.html
jnugh: `umbrellajs` seems to always returns an array, but I want to query only one element when using `querySelector` (so I will keep the original code)

jnugh:
I noticed that the date columns in the key table were filled with dummy values only. Also fixed that :)

backend/cms/templates/settings/user.html
jnugh: I added a description next to the table (a paragraph did not look good). Please have a look at the text :smile:

backend/cms/templates/registration/login_mfa.html
timoludwig: `querySelector` always returns the first matched element, so to get the same functionality in umbrellajs, you'd either use an id-selector instead of classes `u('#auth-error').removeClass('hidden')` or explicitly choose the first element `u('.auth-error').first().removeClass('hidden')`. But that's just a side note, please don't rewrite the js-code, I mean this library should make life easier and not cause pointless work :grin:

backend/cms/templates/settings/user.html
timoludwig: Yeah I think this description is really good! :+1:

backend/cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -1885,6 +2032,22 @@ msgstr "Rolle wurde erfolgreich gespeichert."
...
+#, fuzzy
+#| msgid "The username or the password is incorrect."
+msgid "The provided password is not correct"
+msgstr "Der Benutzername oder das Passwort ist falsch."
+
+#: views/settings/mfa/mfa.py:101
+#, fuzzy
+#| msgid "Maybe it's already been used."
+msgid "This nickname has already been used"
+msgstr "Vielleicht wurde er schon genutzt."
```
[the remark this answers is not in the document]
timoludwig: Nevermind, I just noticed the PR has conflicts in the translation file anyway. I will fix these two entries in the merge commit.

PR #254, opened 2019-12-04
Check if dependencies are installed (fixes #252)

### LINKED ISSUE #252: Add npm version requirement to dev tools
The dev tools only work for npm version 5 or higher, so just checking if npm is installed is not enough in the dev tool `install.sh`. We should add a check whether the version is sufficient and provide information how to upgrade if no recent version of npm is contained in the official distro's repositories.

timoludwig:
Adds a little bit of bash magic to determine if the required version of npm is installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,13 +4,13 @@ This project aims to develop a content management system tailored to the needs o
 ## TL;DR
 
 ### Requirements
-Following packages are requiered for running the project (Install them with your package manager):
+Following packages are required for running the project (Install them with your package manager):
 * git
 * npm
 * python3
-* python3-dev 
-* postgresql *OR* docker
 * python-virtualenv
+* python3-dev (only on Ubuntu)
+* postgresql *OR* docker
 
 ### Installation
 ````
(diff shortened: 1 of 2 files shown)


timoludwig:
The package `python3-dev` is ubuntu-specific, so I do not want to add it to the universal requirements in `install.sh`. I hope the notice in the README.md is enough.

jnugh:
Just tried to run the install script. I had the same issue as with the run script where node (and npm) was not in PATH for root. So the install script should be updated as well.

I was wondering why sudo is used for npm install in the first place. It works perfectly fine when I remove `sudo` from this command and am not sure if npm should be executed with root privileges at all.

Without `sudo` I was able to run the script just fine and the npm version check works for me :smile:

timoludwig:
Ah yeah you're right, running `sudo npm` is generally considered bad practise, probably I needed it because my npm directory had the wrong file permissions. I should have looked into this problem earlier instead of just throwing sudo at it :sweat_smile:

PR #258, opened 2019-12-17
Fix mirrored page validation, fixes #259

### LINKED ISSUE #259: Creating pages not possible
Due to a JavaScript error, it is currently not possible to submit the page form if the page is not yet existing.

svenseeberg:
The mirrored_page_first field used some black Django magic. This commit uses the default render method for generating the input field.

Fixes #259


backend/cms/templates/pages/page.html
```diff
@@ -316,8 +319,10 @@ <h2 class="heading font-normal">
                         editor.selection.setContent('<span class="notranslate" translate="no">' + editor.selection.getContent() + '</span>');
                     }
                 }
-            }),
```
svenseeberg: @timoludwig shouldn't this be a ; ?

timoludwig: Yes, this is a syntax error. Probably not changing anything, but better fix it :wink:

PR #261, opened 2019-12-17
Fix bug where tinymce prevents form from being submitted (part of #259)

timoludwig:
This makes tinymce save everytime the input field changes.


timoludwig:
Is fixed in  #258

PR #264, opened 2019-12-18
Refactor generate_unique_slug

### LINKED ISSUE #260: Creating POIs not possible
Creating POIs fails with the following error:
```
AttributeError at /augsburg/pois/de-de/new

'NoneType' object has no attribute 'region'
```
This is due to a bug in either the slug utils or the poi post view.

### LINKED ISSUE #262: Slug of content types not always unique
When creating new content objects (pages, events and pois), the slug is not necessarily unique.
Steps to reproduce:

- Create a page with a permalink/slug which already exists
- It will be saved without modifications, resulting in a conflict

When saving an existing page, the slug gets unique automatically.

timoludwig:
I refactored the clean slug method because it was buggy for new objects.
The main changes include:
    - the region and language parameter for checking for existing slugs come from the form object now instead of the form object instance
    - the new property `translation.foreign_object` allows us to get the meta object for different content types, e.g. `page_translation.page` or `event_translation.event`
This changes also fixed the issue that POIs could not be created.

Fixes #262
Fixes #260

### DIFF
--- a/backend/cms/views/events/event_view.py
+++ b/backend/cms/views/events/event_view.py
@@ -97,6 +97,8 @@ class EventView(PermissionRequiredMixin, TemplateView):
         event_translation_form = EventTranslationForm(
             data=request.POST,
             instance=event_translation_instance,
+            region=region,
+            language=language,
         )
 
         if (
@@ -146,7 +148,6 @@ class EventView(PermissionRequiredMixin, TemplateView):
             )
             event_translation = event_translation_form.save(
                 event=event,
-                language=language,
                 user=request.user
             )
 
--- a/backend/cms/views/events/forms/event_translation_form.py
+++ b/backend/cms/views/events/forms/event_translation_form.py
@@ -24,8 +24,12 @@ class EventTranslationForm(forms.ModelForm):
             'minor_edit',
         ]
 
-    def __init__(self, data=None, instance=None, disabled=False):
+    #pylint: disable=too-many-arguments
+    def __init__(self, data=None, instance=None, disabled=False, region=None, language=None):
         logger.info('EventTranslationForm instantiated with data %s and instance %s', data, instance)
+
+        self.region = region
+        self.language = language
 
         # To set the status value through the submit button, we have to overwrite the field value for status.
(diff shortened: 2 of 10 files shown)


backend/cms/views/events/event_view.py
```diff
@@ -97,6 +97,8 @@ def post(self, request, **kwargs):
         event_translation_form = EventTranslationForm(
             data=request.POST,
             instance=event_translation_instance,
+            region=region,
```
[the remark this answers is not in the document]
timoludwig: Previously, I got the region via the form instance, but this didn't work when the object was just created and the instance was `None`. So I now get the region via the form directly. Page and POI translation forms already had it, so I only needed to add it to the event translation form.

timoludwig:
@ulliholtgrave @svenseeberg Thanks for your reviews, can you have a look at my recent changes? :slightly_smiling_face:

backend/cms/views/events/forms/event_translation_form.py
```diff
@@ -71,4 +75,7 @@ def save(self, event=None, language=None, user=None):
         return event_translation
 
     def clean_slug(self):
-        return generate_unique_slug(self, 'event')
+        unique_slug = generate_unique_slug(self, 'event')
+        self.data = self.data.copy()
```
florianhofhammer: Why is this copy necessary?
It only overwrites the previous data with the same data again, right?

timoludwig: In Django, the form data is an immutable dict. So in order to manipulate it, I had to copy it into a mutable dict. And I had to modify it in the first place, because when the form is saved successfully, the form is rendered again and then the data is not updated, so the form contains the old, "un-cleaned" slug as placeholder, even though the clean slug was successfully saved to the database. I'm not sure if this explanation is understandable :slightly_smiling_face: 
The other forms didn't need this adaption, because they never render the form after a submission, but redirect instead.

florianhofhammer: Ok, I just didn't notice that it is immutable. Then of course it makes sense to copy it in order to make it mutable!

PR #272, opened 2020-01-17
Improve PageForm

timoludwig:
- Move tinyMCE js and css to html head
- Improve JS handling in page form

### DIFF
--- a/backend/cms/static/js/poi.js
+++ b/backend/cms/static/js/poi.js
@@ -1,7 +1,16 @@
-// event handler for showing confirmation popups
-u('.confirmation-button').each(function(node) {
-    u(node).handle('click', show_confirmation_popup);
+u(document).handle('DOMContentLoaded', function() {
+
+    // event handler for showing confirmation popups
+    u('.confirmation-button').each(function(node) {
+        u(node).handle('click', show_confirmation_popup);
+    });
+    // event handler for closing confirmation popups
+    u('.confirmation-popup').each(function(node) {
+        u(node).find('button').handle('click', close_confirmation_popup);
+    });
+
 });
+
 function show_confirmation_popup(e) {
     let button = u(e.target).closest('button');
     let confirmation_popup = u(button.attr('data-confirmation-popup'));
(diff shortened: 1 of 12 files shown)


backend/cms/static/js/poi.js
```diff
@@ -1,7 +1,16 @@
...
+    // event handler for showing confirmation popups
+    u('.confirmation-button').each(function(node) {
+        u(node).handle('click', show_confirmation_popup);
+    });
+    // event handler for closing confirmation popups
+    u('.confirmation-popup').each(function(node) {
+        u(node).find('button').handle('click', close_confirmation_popup);
+    });
+
+};
```
ulliholtgrave:
```suggestion
});
```

ulliholtgrave: @timoludwig .... :1st_place_medal:

backend/cms/templates/_raw.html
```diff
@@ -13,11 +13,13 @@
...
+    {% block css %}{% endblock %}
     {% endcompress %}
     {% compress js %}
     <script src="{% static 'umbrellajs/umbrella.js' %}"></script>
     <script src="{% static 'feather-icons/dist/feather.js' %}"></script>
     <script src="{% static 'base64-js/base64js.min.js' %}"></script>
+    {% block javascript_head %}{% endblock %}
     {% endcompress %}
 </head>
 <body>
```
ulliholtgrave: We should put the feather.replace() call to a nocompress block, as the compression leads to not showing any icons in the CMS

backend/cms/static/js/poi.js
timoludwig: :see_no_evil:

backend/cms/templates/_raw.html
timoludwig: `feather.replace()` is also compressed in the [current develop branch](https://github.com/Integreat/cms-django/blob/351398d3e617f4f5aa0d42d4a7216295652e10ed/backend/cms/templates/_raw.html#L26)... so I don't think that's the issue here...

timoludwig:
I moved the css rules of tinyMCE before our custom rules now, this way the tinyMCE-styles do not override them anymore. Does it work for you now?

backend/cms/templates/_raw.html
ulliholtgrave: Yes, but now the compression differs in such way, that it ends up with ;feather.replace() instead of the semicolon on the other side

ulliholtgrave:
That seems to fix the problem :+1:

backend/cms/templates/_raw.html
timoludwig: Ok, this sounds like dark magic to me :fearful:

PR #274, opened 2020-01-20
Bugfix/opening delete confirmation twice

tailoric:
template was included twice


tailoric:
this was actually supposed to be a Pull Request for the coldaid fork

PR #295, opened 2020-02-28
Add check for changed SECRET_KEY, fixes #176

### LINKED ISSUE #176: Enforce new SECRET_KEY in production environments
The server should refuse to start if the SECRET_KEY does not differ from the default value in our repository.

The importance of this issue depends on whether the Federated Cloud feature is using this key or not.

svenseeberg:
We don't want the secret key from the repository to be present in any production environment. This change verifies that the key was changed if DEBUG == False.

### DIFF
--- a/backend/cms/apps.py
+++ b/backend/cms/apps.py
@@ -1,8 +1,12 @@
 """
 Django related class representing a config of an app
 """
+import logging
+import sys
+from django.conf import settings
 from django.apps import AppConfig
 
+logger = logging.getLogger(__name__)
 
 class CmsConfig(AppConfig):
     """
@@ -10,3 +14,8 @@ class CmsConfig(AppConfig):
     """
 
     name = 'cms'
+
+    def ready(self):
+        if settings.SECRET_KEY == '-!v282$zj815_q@htaxcubylo)(l%a+k*-xi78hw*#s2@i86@_' and not settings.DEBUG:
+            logger.error("You are running the Integreat CMS in production mode. Change the SECRET_KEY in the settings.py!")
+            sys.exit(1)


backend/cms/apps.py
```diff
@@ -1,12 +1,17 @@
...
 
-
 class CmsConfig(AppConfig):
     """
     Class inheriting the django AppConfig
     """
 
     name = 'cms'
+
+    if settings.SECRET_KEY == '-!v282$zj815_q@htaxcubylo)(l%a+k*-xi78hw*#s2@i86@_' and not settings.DEBUG:
```
timoludwig: I'm not sure, but as [I understand the docs](https://docs.djangoproject.com/en/dev/ref/applications/#django.apps.AppConfig.ready), we need a function which is called `ready`, and place the check inside this function...

backend/cms/apps.py
```diff
@@ -1,12 +1,17 @@
...
+        print("Your are running the Integreat CMS in production mode. Change the SECRET_KEY in the settings.py!")
```
florianhofhammer: Just a small thing, but here's a typo... Should be "you" instead of "your".
Apart from that, do we want to hard code the same string into the code twice (in `settings.py` and in `apps.py`) or should we move that to a constant?

backend/cms/apps.py
svenseeberg: This indeed fixes the issue of multiple execution. And yes, probably better than having this in the class itself :sweat_smile:

backend/cms/apps.py
svenseeberg: Done

svenseeberg:
Should we use `print` here or already use the logger?

timoludwig:
I agree, using a logger is probably the preferable solution.

PR #297, opened 2020-02-29
Change pages API endpoint, fixes #275

### LINKED ISSUE #275: Deliver content of mirrored page via API
So far the content of the mirrored page isn't delivered via the API

Steps to reproduce:

1. Add mirrored content to page
2. Save
3. Access API (for example http://0.0.0.0:8000/api/augsburg/de-de/pages/)

svenseeberg:
Add mirrored page translation to page content.

### DIFF
--- a/backend/api/v3/pages.py
+++ b/backend/api/v3/pages.py
@@ -23,7 +23,7 @@ def transform_page(page_translation):
         'title': page_translation.title,
         'modified_gmt': page_translation.last_updated,
         'excerpt': page_translation.text,
-        'content': page_translation.text,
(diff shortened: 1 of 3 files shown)


ulliholtgrave:
Isn't this PR missing the fact that we support adding an text before a page as well?

timoludwig:
@ulliholtgrave the part
```
if self.page.mirrored_page_first:
    return attached_text + self.text
return self.text + attached_text
```
does exactly that (note the order of the text parts)

ulliholtgrave:
Ahh okay, my fault.

PR #299, opened 2020-02-29
Add GVZ API app

svenseeberg:


backend/gvz_api/utils.py
```diff
@@ -0,0 +1,230 @@
+"""
+Helper classes for Gemeindeverzeichnis API
+"""
+import logging
+import json
+import requests
+from django.conf import settings  # pylint: disable=E0401
```
timoludwig:
```suggestion
from django.conf import settings
```

backend/gvz_api/utils.py
```diff
@@ -0,0 +1,230 @@
...
+from cms.constants import administrative_division
+
+logger = logging.getLogger(__name__)  # pylint: disable=C0103
```
timoludwig:
```suggestion
logger = logging.getLogger(__name__)
```

backend/gvz_api/utils.py
```diff
@@ -0,0 +1,230 @@
...
+        :rtype: list
+        """
+        response = requests.get("{}/searchcounty/{}".format(self.api_url,
+                                                            region_key))
+        content = json.loads(response.text)
+        if content:
+            return content[0]['children']
+        return []
+
+    def translate_type(self, region_type):  # pylint: disable=R0201
```
timoludwig:
```suggestion
    @staticmethod
    def translate_type(region_type):
```

backend/gvz_api/utils.py
```diff
@@ -0,0 +1,230 @@
...
+        Special filters for certain region types. For example, Kreisfreie
+        Städte exist multiple times on different levels. We want the lower
+        level that contain the coordinates.
+
+        :param region_details: dictionary with region details
+        :type region_details: dict
+        :return: indicates if a region is not interesting
+        :rtype: bool
+        """
+        if self.translate_type(
```
timoludwig:
```suggestion
    @classmethod
    def filter_region_types(cls, region_details):
        """
        Special filters for certain region types. For example, Kreisfreie
        Städte exist multiple times on different levels. We want the lower
        level that contain the coordinates.

        :param region_details: dictionary with region details
        :type region_details: dict
        :return: indicates if a region is not interesting
        :rtype: bool
        """
        if cls.translate_type(
```
This change is necessary if `translate_type()` is a static method.

Issue #332, opened 2020-03-07
Improve documentation of page xliff converter


timoludwig:
Explain how the page xliff converter works and add docstrings of the following format:
```
"""
[Summary]

:param [ParamName]: [ParamDescription], defaults to [DefaultParamVal]
:type [ParamName]: [ParamType](, optional)
...
:raises [ErrorType]: [ErrorDescription]
...
:return: [ReturnDescription]
:rtype: [ReturnType]
"""
```

svenseeberg:
Will be included in #341

svenseeberg:
Closed with cf73f4cf09850335e535aea450f98759e5f2b819

Issue #239, opened 2019-11-23
Export XLIFF and UI changes


svenseeberg:
* Remove single download button
* Add input type checkmark to page list (first column)
* After starting download, dont reload page. Keep check marks selected.
* Create export target languge drop down below page list
* The downloaded XLIFFs are zipped, place each target language in a subdirectory.
* XLIFF file name: [PAGE_ID]-[SOURCE_TRANSLATION_VERSION_NUMBER]-[TARGET_LANG_CODE]-[REGION]-[SLUG].xliff
* Zip file name: [REGION]-[DATE]-[SOURCE_LANG_CODE]-[TARGET_LANG_CODE].zip

svenseeberg:
Closed with cf73f4cf09850335e535aea450f98759e5f2b819

Issue #104, opened 2019-01-14
Meta: Translations


svenseeberg:
This is a meta issue with links to all Translations issues:
* [x] #38 Import XLIFF
* [x] #37 Export XLIFF
* [x] #85 Translation side-by-side view
* [x] #84 Export pages
* [x] #83 Status icons
* [x] #9 Translation coverage analytics
* [x] #155 Tab view for different languages
* [x] #156 Review workflow

timoludwig:
- Language Management: LanguageTree to List
- One translator per page
- One origin language per page translation (null for default language)

svenseeberg:
Candidates:

* https://pypi.org/project/django-xliff/ / https://github.com/callowayproject/django-xliff
* https://pypi.org/project/slc.xliff/ / https://github.com/collective/slc.xliff
* https://pypi.org/project/py-xliff-converter / https://github.com/romanvm/py-xliff-converter

timoludwig:
Ok. Another option might be https://pypi.org/project/django-modeltranslation-xliff/ / https://github.com/romanvm/django-modeltranslation-xliff, but I don't know if we can use this outside the standard django admin backend.

svenseeberg:
The Expedia team noticed that the django-modelstranslations is bound to the back end languages defined in the settings.py. Because our content languages are independent of the UI languages, we need our own implementation for exporting into XLIFF files. But we can copy functions for generating XLIFF files from existing modules to speed up the development.

svenseeberg:
I remove `#36 Automatic translation API (DeepL, Google) (Additional Features)` from this meta issue. I don't see this as part of it right now.

Issue #84, opened 2018-11-17
Translations: Export multiple pages at once


svenseeberg:
User needs to be able to export multiple pages (multiple XLIFF files) in a zip file for translation into a target language. Single page export into XLIFF is covered in #37 

Allow filtering: source language, target language, by page status

svenseeberg:
Closed with cf73f4cf09850335e535aea450f98759e5f2b819

PR #341, opened 2020-03-07
Improve XLIFF workflow

### LINKED ISSUE #238: Move XLIFF Import input below page list
Move XLIFF Import input below page list

svenseeberg:
* Refactor XLIFF converter, drop BS4, directly use LXML.
* Remove outdated tests, no current tests sadly.
* Enable exporting a selection of pages into XLIFF for one specific target language.
* Move export to bottom of page list.
* Add frame for bulk actions.
* Add doc comments to xliff helper classes.
* Clean up XLIFF helper classes.
* Fixes #238, #239, #332, #84, #104


src/cms/templates/pages/page_tree.html
```diff
@@ -125,6 +154,32 @@ <h2 class="heading font-normal mb-2">{% trans 'Pages' %}</h2>
...
+        function select_all_pages() {
+            var selected_pages = document.getElementsByName("page_selected");
+            if (document.getElementsByName("select_all_pages")[0].checked == true) {
+                selected_pages.forEach(e => e.checked = true);
+            } else {
+                selected_pages.forEach(e => e.checked = false);
+            }
+        }
+
+        function bulk_action_execute() {
```
svenseeberg, line 166: I think the biggest question is: do we want to handle bulk actions like this?

src/cms/templates/pages/page_tree.html
```diff
@@ -117,6 +100,52 @@ <h2 class="heading font-normal mb-2">{% trans 'Pages' %}</h2>
...
+                        {% for lang in languages %}
+                        {% if lang != region.default_language %}
+                        <option value="{{ lang.code }}">{% trans 'Export XLIFF for translation to' %} {{ lang.translated_name }}</option>
+                        {% endif %}
+                        {% endfor %}
+                    </select>
+                </div>
+                <div class="pl-3">
+                    <input type="button" onclick="bulk_action_execute();" class="inline-block cursor-pointer bg-blue-500 hover:bg-blue-600 text-white h-full font-bold py-2 px-4 rounded" value="{% trans 'Execute' %}" />
+                    <input type="hidden" id="bulk_action_url" value="{% url 'download_xliff' region_slug=region.slug language_code=language.code%}">
```
svenseeberg, line 123: I placed the URL in a hidden input to make it accessible to the compressed JS code. I think this is cleaner than having inline JS.

src/cms/views/pages/page_actions.py
```diff
@@ -104,39 +104,46 @@ def delete_page(request, page_id, region_slug, language_code):
...
-            page_xliff_helper.delete_tmp_in_xliff_folder(file_path)
-
+def upload_xliff(request, region_slug, language_code):
+    if request.method == 'POST' and 'xliff_file' in request.FILES and request.FILES['xliff_file']:
+        xliff_helper = PageXliffHelper()
+        upload_file = request.FILES['xliff_file']
+        upload_dir = os.path.join(XLIFFS_DIR, 'upload', str(uuid.uuid4()))
+        os.makedirs(upload_dir, exist_ok=True)
+        with open(os.path.join(upload_dir, upload_file.name), 'wb+') as upload_file:
+            for chunk in upload_file.chunks():
```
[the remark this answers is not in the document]
svenseeberg: I had an issue as well. I think it should be fixed with the latest commits.

src/cms/templates/pages/page_tree.html
```diff
@@ -117,6 +100,52 @@ <h2 class="heading font-normal mb-2">{% trans 'Pages' %}</h2>
...
+                <div class="pl-3">
+                    <select class="h-full pl-4 pr-4 rounded shadow" style="padding-top: 5px;" name="bulk_action" id="bulk_action">
+                        <option>{% trans 'Please select' %}</option>
+                        <option value="archive_pages">{% trans 'Archive pages' %}
+                        {% for lang in languages %}
+                        {% if lang != region.default_language %}
+                        <option value="{{ lang.code }}">{% trans 'Export XLIFF for translation to' %} {{ lang.translated_name }}</option>
+                        {% endif %}
+                        {% endfor %}
+                    </select>
```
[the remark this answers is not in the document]
svenseeberg: Actually the translation should be done according to the translation tree only. That means if German is translated to English and English to Arabic, the source for Arabic should be English when exporting to XLIFF. And I think we should only allow exactly this to prevent any misunderstandings.

svenseeberg:
> One general note:
> The file [src/cms/static/js/page-tree.js](https://github.com/Integreat/cms-django/blob/e884e95ae5fc7ae89bb2002568da988ee4a547e1/src/cms/static/js/page-tree.js) is somehow empty, but I couldn't tell whether there is actually code missing or the empty file was just committed unintentionally.

I think I added it in the beginning and totally forgot about it. I intended to move all JS code into this file. But it was much less JS Code than I anticipated.

src/cms/page_xliff_converter.py
```diff
@@ -1,280 +1,203 @@
...
+        :type page: cms.models.pages.Page
+        :return: file path to XLIFF file
+        :rtype: string or None
+        """
+        converter = TranslationXliffConverter(self.src_lang, self.tgt_lang)
 
-        if not filename:
-            filename = f'page_{source_translation_page.page.id}_{source_translation_page.language.code}_{target_language_code}.xliff'
+        src_trans = PageTranslation.objects.get(page=page, language=self.src_lang)
+        tgt_trans = PageTranslation.objects.get(page=page, language=self.tgt_lang)
```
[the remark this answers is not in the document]
svenseeberg: Indeed. Fixed.

src/cms/templates/pages/page_tree.html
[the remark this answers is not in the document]
timoludwig: Ah sorry, I didn't catch that the source language is already determined by the language tree. Perfect!
Yes, I totally agree with you. I just thought maybe it would be good to have the source language also in the select option field to make sure there are absolutely no misunderstandings about which language is the source, e.g.
"Export XLIFF for translation from German to English", "Export XLIFF for translation from English to Arabic", etc... but I think that's optional, since the region managers should know their "translation directions"...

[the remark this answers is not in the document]
svenseeberg: Yes, I thought about this option but went for the shorter version. It is easy to change at any time later though. Just a minor adjustment in the template.

src/cms/page_xliff_converter.py
```diff
@@ -1,358 +1,274 @@
...
+    def add_translation_unit(self, unit_id, source_text, target_text):
+        """
+        Add a translation unit to the file element
+
+        :param unit_id: value for the unit id attribute
+        :type unit_id: str
+        :param source_text: source text to be translated
+        :type source_text: str
+        :param target_text: already available translation
+        :type param: str
```
timoludwig, line 108:
```suggestion
        :type target_text: str
```

src/cms/page_xliff_converter.py
```diff
@@ -362,76 +278,132 @@ def _get_page_translation_slug(title):
...
+            return True
+        return False
+
+    def import_xliff_files(self, xliff_paths, user):
+        """
+        Import single (usually translated) XLIFF file and
+        save to page translation.
+
+        :param xlif_path: paths to XLIFF files
+        :type xliff_path: list [ str ]
```
timoludwig, line 329:
```suggestion
        :param xliff_paths: paths to XLIFF files
        :type xliff_paths: list [ str ]
```

src/cms/page_xliff_converter.py
```diff
@@ -362,76 +278,132 @@ def _get_page_translation_slug(title):
...
+        return file_paths
+
+    def generate_xliff_import_diff(self, xliff_paths):
+        """
+        Generate diff between XLIFF content and current translation content
+
+        :param xliff_paths: list of paths to XLIFF files
+        :type xliff_paths: [ str ]
+        :return: dictionaries containing diffs between XLIFF and current translation versions
+        :rtype: [ dict ]
```
timoludwig, line 372:
```suggestion
        :type xliff_paths: list [ str ]
        :return: dictionaries containing diffs between XLIFF and current translation versions
        :rtype: list [ dict ]
```

src/cms/page_xliff_converter.py
```diff
@@ -1,358 +1,274 @@
...
+        return etree.tostring(self.xliff, pretty_print=True)
+
+
+class PageXliffHelper():
+    """
+    Wrapper and helper methods for reading/writing XLIFF files and saving PageTranslation
+    """
+    def __init__(self, src_lang=None, tgt_lang=None):
+        """
+        :param src_lang: source language of tran
```
timoludwig, line 162:
```suggestion
```

src/cms/page_xliff_converter.py
```diff
@@ -1,358 +1,274 @@
...
-        if not filename:
-            filename = f'page_{source_translation_page.page.id}_{source_translation_page.language.code}_{target_language_code}.xliff'
+        # properties that make a translation unique: page id, target language, source version
+        xliff_id = self.tgt_lang.code + "_" + str(page.id) + "_" + str(src_trans.version)
 
-        file_path = None
+        filename = f"{region_slug}_{self.src_lang.code}__{xliff_id}__{slug}.xliff"
 
-        xliff_content = self.converter.page_translation_to_xliff(source_translation_page, target_language_code)
+        xliff_content = converter.translation_to_xliff(page.id, src_trans, tgt_trans, xliff_id)
```
timoludwig, line 218: I see two problems with this:

1. If the source language doesn't exist for one page, this still gives me a `DoesNotExist`-error
2. If an empty target translation is being created, it has the side-effect that the new target translation is marked with a :heavy_check_mark: because it is more recent than the source translation. So I would suggest to either allow the conversion functions to work with empty default values or to delete the `tgt_trans` object after it has been converted to xliff.

timoludwig: Apart from that, you could replace the conditional blocks with [get_or_create()](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#get-or-create)

src/cms/page_xliff_converter.py
```diff
@@ -362,76 +278,132 @@ def _get_page_translation_slug(title):
...
+
+        :param trans_fields: content extracted from XLIFF
+        :type trans_fields: dict
+        :param xliff_name: file name of XLIFF
+        :type xliff_name: str
+        :return: dictionary containing diff between XLIFF and current translation version
+        :rtype: dict
+        """
+        page = Page.objects.get(id=int(trans_fields["page_id"]))
+        tgt_trans = page.get_translation(trans_fields["tgt_lang_code"])
```
timoludwig, line 399: Also, I get an error when I want to import  a translation which does not yet exist... so I suggest to use [get_or_create()](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#get-or-create) here as well.

src/cms/page_xliff_converter.py
svenseeberg: 1. should be addressed now with a `filter()` instead of `get()`, while checking if there is exactly one result.
2. Even if the `tgt_trans` is not being saved?

timoludwig: 1. Yes this looks good :+1: 
2. Yes, but you could use the constructor `PageTranslation()` instead of `PageTranslation.objects.create()` to instantiate a translation object which is not saved if you don't call `save()`.

src/cms/page_xliff_converter.py
```diff
@@ -362,76 +336,132 @@ def _get_page_translation_slug(title):
...
+                tgt_trans = PageTranslation.objects.create(
+                    slug=slug,
+                    title=trans_fields["title"],
+                    text=trans_fields["content"],
+                    status=src_trans.status,
+                    language=tgt_lang,
+                    public=src_trans.public,
+                    page=page,
+                    creator=user
+                )
```
timoludwig, line 374:
```suggestion
        tgt_trans.title = trans_fields["title"]
        tgt_trans.text = trans_fields["content"]
        tgt_trans.slug = PageXliffHelper._get_page_translation_slug(tgt_trans.title)
        tgt_trans.version = tgt_trans.version + 1
```
After `get_or_create()`, the `tgt_trans`-object should exist in any case, right?

src/cms/page_xliff_converter.py
```diff
@@ -362,76 +336,132 @@ def _get_page_translation_slug(title):
...
+        :param trans_fields: content extracted from XLIFF
+        :type trans_fields: dict
+        :param xliff_name: file name of XLIFF
+        :type xliff_name: str
+        :return: dictionary containing diff between XLIFF and current translation version
+        :rtype: dict
+        """
+        page = Page.objects.get(id=int(trans_fields["page_id"]))
+        tgt_trans = page.get_translation(trans_fields["tgt_lang_code"])
+        old_lines_content = tgt_trans.text.splitlines()
```
timoludwig, line 458: `page.get_translation()` returns `None` if it doesn't exist yet. So in this case, `tgt_trans.text` will trigger the error `'NoneType' object has no attribute 'text'`...
But I guess my idea of using `get_or_create()` here wasn't correct here, because we don't want the translation object to be created before the import is confirmed... so maybe just add the same code as in line [260](https://github.com/Integreat/cms-django/blob/d3366f7db5b818329cbd56717605e363d83b4481/src/cms/page_xliff_converter.py#L260):
```
        if not tgt_trans:
            tgt_trans = PageTranslation(
                title="",
                text="",
                status=src_trans.status,
                language=self.tgt_lang,
                page=page)
```
or similar...

src/cms/templates/pages/page_tree.html
```diff
@@ -128,6 +157,32 @@ <h2 class="heading font-normal mb-2">{% trans 'Pages' %}</h2>
...
+            if ( action == "archive_pages" ) {
+                // TODO
+            } else { // no previous match, than language code -> XLIFF export
+                var page_selected = document.getElementsByName("page_selected");
+                var pages = [];
+                page_selected.forEach(async function (page_checkbox) {
+                    if(page_checkbox.checked) {
+                        pages = pages.concat(page_checkbox.value);
+                    }
+                });
```
timoludwig, line 181:
```suggestion
                page_selected.forEach(async function (page_checkbox) {
                    if(page_checkbox.checked) {
                        pages = pages.concat(page_checkbox.value);
                    }
                });
                if (pages.length == 0) {
                    return;
                }
```
I just noticed that the bulk action opens a new tab without doing anything when no page is selected... I think not opening a new tab makes more sense in this case (in the future, we could even disable the bulk action button when no page is selected, but for now I think it's sufficient).

svenseeberg:
> * It is possible to upload xliffs with empty titles, whereas it is not possible to add translations with empty titles via the normal form. But we should probably validate this at the model-level, not in this module, right?

I guess so, yes.

> * Unicode characters are encoded, e.g. `خوش آمدید` to `&#1582;&#1608;&#1588; &#1570;&#1605;&#1583;&#1740;&#1583;`. This leaves the diff unreadable and doesn't work in page titles. But I'm not sure what's causing this and if this should be part of this PR or addressed in another issue.

I guess that is due to the pre tag. I think we need to iterate on the diff preview at some point. But for now this works in my opinion. Most people importing text can't read Arabic anyways.

PR #343, opened 2020-03-25
Add POI selection to event form

### LINKED ISSUE #247: Link POIs to events in event form
Make it possible to add existing POIs as location for events or create a new POI if an event with a new location is added.
Ideally, make it possible to search for existing POIs instead of having a select tag with dozens of options.

florianhofhammer:
With this PR:
* The location of an event is shown in the event list (non-archived as well as archived).
* If an event has a location, the location is also shown in the event form.
* Inside an event form, a location for an event can be searched (implemented with AJAX calls).
* From the search results, either an existing POI can be selected or a new POI can be created.

Fixes #247

### DIFF
--- a/src/cms/templates/events/event_form.html
+++ b/src/cms/templates/events/event_form.html
@@ -260,26 +260,17 @@
                         </div>
 
                         <label class="mt-4 mb-2 block font-bold uppercase text-sm">{% trans 'Address' %}</label>
-                        <input class="appearance-none block w-full bg-gray-200 text-gray-600 border border-gray-200 rounded mb-2 py-3 px-4 leading-tight focus:outline-none focus:bg-white focus:border-gray-400"
-                               type="text" placeholder="{% trans 'Name of event location' %}">
+                        {% trans 'Name of event location' as poi_title_placeholder %}
+                        {% render_field poi_translation_form.title|add_error_class:"border-red-500" placeholder=poi_title_placeholder class="appearance-none block w-full bg-grey-lighter text-grey-darker border border-grey-lighter rounded mb-2 py-3 px-4 leading-tight focus:outline-none focus:bg-white focus:border-grey" %}
                         <small class="text-sm italic block mb-2">
                             {% trans 'Create an event location or start typing the name of an existing location' %}.
                         </small>
-                        <input class="appearance-none block w-full bg-gray-200 text-gray-600 border border-gray-200 rounded mb-2 py-3 px-4 leading-tight focus:outline-none focus:bg-white focus:border-gray-400"
-                               type="text" placeholder="{% trans 'Street' %}">
-                        <input class="appearance-none block w-full bg-gray-200 text-gray-600 border border-gray-200 rounded mb-2 py-3 px-4 leading-tight focus:outline-none focus:bg-white focus:border-gray-400"
-                               type="text" placeholder="{% trans 'City' %}">
-                        <div class="relative my-2">
-                            <select class="block appearance-none w-full bg-gray-200 border border-gray-200 text-gray-800 py-3 px-4 pr-8 rounded leading-tight focus:outline-none focus:bg-white focus:border-gray-400">
-                                <option>{% trans 'Germany' %}</option>
-                            </select>
-                            <div class="pointer-events-none absolute inset-y-0 right-0 flex items-center px-2 text-gray-800">
-                                <svg class="fill-current h-4 w-4" xmlns="http://www.w3.org/2000/svg"
-                                     viewBox="0 0 20 20">
-                                    <path d="M9.293 12.95l.707.707L15.657 8l-1.414-1.414L10 10.828 5.757 6.586 4.343 8z"/>
-                                </svg>
-                            </div>
-                        </div>
+                        {% trans 'Street' as street_placeholder %}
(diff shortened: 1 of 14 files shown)


src/cms/templates/events/event_form.html
```diff
@@ -399,89 +399,92 @@ <h3 class="font-bold my-3">{% trans 'Where' %}</h3>
...
             readonly : 1{% endif %}
-    });
+            });
 
-    custom_file_field();
-}, false);
-</script>
+            custom_file_field();
+        }, false);
+    </script>
```
timoludwig, line 489: This diff seems very broken, but all you added was a bit of indentation, right? :sweat_smile:

src/cms/views/events/event_actions.py
```diff
@@ -60,3 +64,32 @@ def delete(request, event_id, region_slug, language_code):
...
+    poi_query_result = []
+
+    if poi_query:
+        # Only query for pois from current region
+        for poi in region.pois.all():
+            for language in poi.languages:
+                poi_translation = poi.get_public_translation(language_code=language.code)
+                if poi_translation is not None and poi_query.lower() in poi_translation.title.lower():
+                    poi_query_result.append(poi)
+                    break
```
timoludwig, line 89: This could speed up the execution a bit (especially when there are many pois, the difference could be noticeable)...
I'm aware of the fact that with my solution we cannot keep track of the versions and might include results where only the title of an old version matches, but I'm not sure if this is a problem...

The field lookup selector `icontains` is the case-insensitive variant of `contains`.
```suggestion
    poi_query_result = region.pois.prefetch_related('translations').filter(
        archived=False,
        translations__title__icontains=poi_query,
        translations__status=status.PUBLIC
    ).distinct()
```

timoludwig: Oh, if you go with my solution, you have to include
```
```
in the header of course.

src/cms/templates/events/event_form.html
florianhofhammer: Yes, I included my own script above and wanted to have proper indentation. This looks weird in the generated HTML source if you look it up in the browser but imho it looks better in the IDE (which was more important to me). If you don't agree, I can of course easily remove the indentation.

src/cms/views/events/event_actions.py
florianhofhammer: I had very similar code for the query in the beginning but the problem was exactly what you mentioned above - old versions occurred in the results. This is especially confusing if we have an older version of a POI with a similar name than another POI in the current version (e.g. POI 1, version 1: "test location", version 2: "new location"; POI 2, version 1: "another test"). In the example, the query "test" would not yield both POIs, which could lead to confusion.

I therefore deliberately chose to implement it the way it currently is. I think that is a point we should discuss about maybe in the call tomorrow.

src/cms/templates/events/event_form.html
timoludwig: No, the indentation is fine, I just wondered whether I missed some other changes there.

src/cms/views/events/event_actions.py
timoludwig: Ok. In some cases, this might even be helpful (e.g. if a location is renamed and a user only remembers the old name), but especially when locations are created as draft with temporary titles, the results can be misleading.

Somehow I have the feeling that we are able to achieve the desired queries with the advanced Django expressions [Q()](https://docs.djangoproject.com/en/3.0/topics/db/queries/#complex-lookups-with-q-objects) and [F()](https://docs.djangoproject.com/en/3.0/ref/models/expressions/#f-expressions) :sweat_smile:
Maybe I'll have a look into it...

florianhofhammer: I've never heard of those options before but they actually sound pretty promising!

timoludwig: Ok, now I think I got it :sweat_smile: 
No need for Q&F, a subquery was the way to go:

```suggestion
    # All latest revisions of a POI (one for each language)
    latest_public_poi_revisions = POITranslation.objects.filter(
        poi=OuterRef('pk'),
        status=status.PUBLIC
    ).order_by('language', '-version').distinct('language').values('id')
    # All POIs which are not archived and have a latest public revision which contains the query
    poi_query_result = region.pois.prefetch_related('translations').filter(
        archived=False,
        translations__in=Subquery(latest_public_poi_revisions),
        translations__title__icontains=poi_query
    ).distinct()
```

timoludwig: (Additionally needs:
```
from django.db.models import OuterRef, Subquery

from ...models import Event, Region, POITranslation
```

src/cms/views/events/event_actions.py
```diff
@@ -1,11 +1,16 @@
+import json
+import logging
+
 from django.contrib import messages
 from django.contrib.auth.decorators import login_required
 from django.core.exceptions import PermissionDenied
```
timoludwig, line 6:
```suggestion
from django.core.exceptions import PermissionDenied
from django.db.models import OuterRef, Subquery
```

src/cms/views/events/event_actions.py
```diff
@@ -1,11 +1,16 @@
...
-from django.shortcuts import redirect
+from django.shortcuts import render, redirect
 from django.utils.translation import ugettext as _
```
timoludwig, line 9:
```suggestion

from ...constants import status
```

src/cms/views/events/event_actions.py
```diff
@@ -1,11 +1,16 @@
...
 
 from ...decorators import region_permission_required, staff_required
-from ...models import Event
+from ...models import Event, Region
```
timoludwig, line 11:
```suggestion
from ...models import Event, Region, POITranslation
```

Issue #333, opened 2020-03-07
Improve documentation of cms rules


timoludwig:
Explain the rules module and how it interacts with our permission management. Add docstrings of the following format:
```
"""
[Summary]

:param [ParamName]: [ParamDescription], defaults to [DefaultParamVal]
:type [ParamName]: [ParamType](, optional)
...
:raises [ErrorType]: [ErrorDescription]
...
:return: [ReturnDescription]
:rtype: [ReturnType]
"""
```

PR #345, opened 2020-04-03
Feature/single page endpoint

### LINKED ISSUE #14: API: single page endpoint
Endpoint for single pages. This must be Integreat APIv3 compatible. The page is selected by either a permalink or page ID.

https://github.com/Integreat/cms/wiki/REST-APIv3-Documentation#pageeventdisclaimer

ulliholtgrave:
PR adds the single page endpoint to the API as requested in #14 .

Fixes #14

### DIFF
--- a/src/api/v3/single_page.py
+++ b/src/api/v3/single_page.py
@@ -1,23 +1,41 @@
+"""
+View to return a JSON representation of a single page. The page can
+be selected via the id or the permalink.
+"""
 from django.http import JsonResponse, HttpResponse
+from django.shortcuts import get_object_or_404
 
 from cms.models import Region
 from .pages import transform_page
 
 def single_page(request, region_slug, language_code):
+    """
+    View function returning the desired page as a JSON or a 404 if the
+    requested page does not exist.
+    Args:
+        request : The request that has been send to the Django.
+        region_slug : Slug defining the region
+        language_code : Code to identify the desired language.
+
+    Returns:
+        HttpResponse: Return an JSON with the requested page and a HTTP 200 or an 404.
+    """
+    region = get_object_or_404(Region, slug=region_slug)
+
     if request.GET.get('id', ''):
-        region = Region.objects.get(slug=region_slug)
         page_translation = None
         page = region.pages.filter(id=int(request.GET.get('id', '')))
         if page:
             page_translation = page.first().get_public_translation(language_code)
-            return JsonResponse(transform_page(page_translation), safe=False)
+            if page_translation is not None:
+                return JsonResponse(transform_page(page_translation), safe=False)
+
     elif request.GET.get('url', ''):
-        region = Region.objects.get(slug=region_slug)
         for page in region.pages.all():
             page_translations = page.translations.all()
             for translation in page_translations:
-                if (translation.permalink == request.GET.get('url', '')) & (translation != None) :
+                if (translation.permalink == request.GET.get('url', '')) & (translation is not None):
                     return JsonResponse(transform_page(translation), safe=False)
-    
-    return HttpResponse('The requested Page does not exist', status=404)
+
+    return HttpResponse('The requested Page does not match any url or id', status=404)
     
--- a/src/api/v3/single_page.py
+++ b/src/api/v3/single_page.py
@@ -47,5 +47,5 @@ def single_page(request, region_slug, language_code):
         if page_translation.permalink == url:
             return JsonResponse(transform_page(page_translation), safe=False)
 
-    return HttpResponse('The requested Page does not match any url or id', status=404)
+    raise Http404('No Page matches the given url or id.')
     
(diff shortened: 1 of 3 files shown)


src/api/v3/single_page.py
```diff
@@ -0,0 +1,41 @@
...
+            page_translation = page.first().get_public_translation(language_code)
+            if page_translation is not None:
+                return JsonResponse(transform_page(page_translation), safe=False)
+
+    elif request.GET.get('url', ''):
+        for page in region.pages.all():
+            page_translations = page.translations.all()
+            for translation in page_translations:
+                if (translation.permalink == request.GET.get('url', '')) & (translation is not None):
+                    return JsonResponse(transform_page(translation), safe=False)
```
timoludwig, line 38:
```suggestion
    elif request.GET.get('url'):
        # Strip leading and trailing slashes to avoid ambiguous urls
        url = request.GET.get('url').strip('/')
        # Get potential page candidate by only filtering for the translation slug
        page = get_object_or_404(Page, region=region, translations__slug=url.split('/')[-1])
        # Get most recent public revision of the page
        page_translation = page.get_public_translation(language_code)
        # Check if the whole path is correct, not only the slug
        # TODO: Once we have a permalink mapping of old versions, we also have to check whether the permalink was valid in the past
        if page_translation.permalink == url:
            return JsonResponse(transform_page(page_translation), safe=False)
```
The for-loop is very inefficient, especially for regions with many pages this could lead to performance issues.
As the slug is unique across regions, I suggest to query for a potential page first by filtering for the slug, and then check if the whole permalink is correct as well.

src/api/v3/single_page.py
```diff
@@ -0,0 +1,41 @@
...
+
+    return HttpResponse('The requested Page does not match any url or id', status=404)
+
```
timoludwig, line 41:
```suggestion
    raise Http404('No Page matches the given url or id.')

```
This would be more standardized.

src/api/v3/single_page.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    View function returning the desired page as a JSON or a 404 if the
+    requested page does not exist.
+    Args:
+        request : The request that has been send to the Django.
+        region_slug : Slug defining the region
+        language_code : Code to identify the desired language.
+
+    Returns:
+        HttpResponse: Return an JSON with the requested page and a HTTP 200 or an 404.
```
timoludwig, line 21:
```suggestion

    :param request: The request that has been sent to the Django server
    :type request: django.http.HttpRequest

    :param region_slug: Slug defining the region
    :type region_slug: str

    :param language_code: Code to identify the desired language
    :type language_code: str

    :raises django.http.Http404: HTTP status 404 if the request is malformed or no page with the given id or url exists. 

    :return: Return a JSON with the requested page and a HTTP status 200.
    :rtype: django.http.JsonResponse
```
I'd suggest sticking to the official Sphinx docstring syntax as described in #333.

src/api/v3/single_page.py
```diff
@@ -0,0 +1,41 @@
+"""
+View to return a JSON representation of a single page. The page can
+be selected via the id or the permalink.
+"""
+from django.http import JsonResponse, HttpResponse
```
timoludwig, line 5:
```suggestion
from django.http import JsonResponse, Http404
```
This is needed if you apply my suggestion below.

src/api/v3/single_page.py
```diff
@@ -0,0 +1,41 @@
...
+from django.shortcuts import get_object_or_404
+
+from cms.models import Region
```
timoludwig, line 8:
```suggestion
from cms.models import Region, Page
```
This is needed if you apply my suggestion below.

src/api/v3/single_page.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    region = get_object_or_404(Region, slug=region_slug)
+
+    if request.GET.get('id', ''):
+        page_translation = None
+        page = region.pages.filter(id=int(request.GET.get('id', '')))
+        if page:
+            page_translation = page.first().get_public_translation(language_code)
+            if page_translation is not None:
+                return JsonResponse(transform_page(page_translation), safe=False)
```
timoludwig, line 31:
```suggestion
        page = get_object_or_404(Page, region=region, id=request.GET.get('id'))
        page_translation = page.get_public_translation(language_code)
        if page_translation:
            return JsonResponse(transform_page(page_translation), safe=False)
```
I'd suggest to use the `get_object_or_404`-function here as well.

src/api/v3/single_page.py
ulliholtgrave: Unfortunately I cannot add this suggestion to the batch as you changed the same lines in another proposed change

timoludwig: Oh Ok sorry :innocent:

PR #361, opened 2020-04-17
Refactor technical feedback

### LINKED ISSUE #360: Refactor technical feedback
The current implementation of technical feedback does not comply with the way it is used in the apps when submitting feedback. Instead of having one model for technical feedback, we should rather have a `technical` flag for each of the other feedback models. This flag should be set whenever the `category`-parameter contains "technisch" or "technical". In a future API version, we can convert this to a parameter, which can either be true or false in the feedback submission request.

Thomas125:
Fixes #360


src/cms/models/feedback/feedback.py
```diff
@@ -15,6 +15,7 @@ class Feedback(models.Model):
     )
     emotion = models.CharField(max_length=3, choices=EMOTION)
     comment = models.CharField(max_length=1000)
+    isTechnicalFeedback = models.BooleanField(default=False)
     readStatus = models.BooleanField(default=False)
```
timoludwig, line 19:
```suggestion
    is_technical = models.BooleanField(default=False)
    read_status = models.BooleanField(default=False)
```
The [official Django coding styleguides](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#model-style) recommend lowercase & underscore for field names instead of CamelCase.

PR #364, opened 2020-04-18
Api: feedback-handler-decorator + search result feedback

### LINKED ISSUE #24: API: search result feedback
Endpoint for posting feedback on search results.

https://github.com/Integreat/cms/wiki/REST-APIv3-Documentation#request-11

Thomas125:
This PR adds a decorator for functions that handle post-requests for feedback (feedback_handler) and uses this decorator to handle search result feedback.
The decorator handles the parameters and necessary checks that occur on each feedback endpoint and can be used for each feedback endpoint to avoid duplicated code.

Resolves #24


src/api/v3/feedback/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+from django.http import HttpResponse
+
+
+def feedback_handler(func):
+    def handle_feedback(request, region_slug, language_code):
+        if request.method != 'POST':
+            return HttpResponse('Invalid request.', status=405)
+        data = json.loads(request.body)
+
+        comment = data.pop("comment", "")
```
timoludwig, line 12: Just a hint: `request.POST` is immutable, so we have to use `request.POST.get()` instead of  `request.POST.pop()`.

src/api/v3/feedback/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+        rating = data.pop("rating", None)
+        category = data.pop("category", None)
+
+        if rating and rating != 'up' and rating != 'down':
```
timoludwig, line 16: You could also use:
```suggestion
        if rating not in [None, 'up', 'down']:
```

src/api/v3/feedback/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+            return HttpResponse("Invalid rating.", status=400)
+        if comment == "" and not rating:
+            return HttpResponse("Either comment or rating is required.", status=400)
+        if rating == 'up':
+            emotion = 'Pos'
+        elif rating == 'down':
+            emotion = 'Neg'
+        else:
+            emotion = "NA"
+        is_technical = category == "Technisches Feedback" #todo: determine is_technical
```
timoludwig, line 26: I clarified with the app team that this is a long-term legacy convention (so nothing else to do).
```suggestion
        is_technical = category == "Technisches Feedback"
```

src/api/v3/feedback/search_result_feedback.py
```diff
@@ -0,0 +1,13 @@
+from django.http import HttpResponse
+
```
timoludwig, line 2: I got a csrf-error when testing the api, so we might add this decorator:
```suggestion
from django.views.decorators.csrf import csrf_exempt
```

src/api/v3/feedback/search_result_feedback.py
```diff
@@ -0,0 +1,13 @@
...
+from api.v3.feedback import feedback_handler
+from cms.models import SearchResultFeedback
+
```
timoludwig, line 5: Disable csrf-protection for the feedback api:
```suggestion


@csrf_exempt
```

src/api/v3/feedback/__init__.py
```diff
@@ -0,0 +1,28 @@
+import json
+
+from django.http import HttpResponse
+
+
+def feedback_handler(func):
+    def handle_feedback(request, region_slug, language_code):
+        if request.method != 'POST':
+            return HttpResponse('Invalid request.', status=405)
```
timoludwig, line 9: The wordpress api answered `JsonResponse` here, so it's probably a good idea to keep it that way (also involves all other api responses)...

Thomas125: So, `return JsonResponse('Invalid request.', status=405, safe=False` ?

Thomas125: In the page feedback endpoint, HttpResponses are returned. Is that an error there?

src/api/v3/feedback/__init__.py
Thomas125: At the end of this decorator function, `data` should be a dictionary with all params but without comment, rating and category.
Online I found
```
request.GET
data = request.GET.copy()
data.pop('key') 
```
But that doesn't work. For a strange reason, a.get() returns the desired object, but a.pop() returns an array with this object.

Thomas125: The page feedback endpoint uses `json.loads(request.body)`.
Is that an error?

src/api/v3/feedback/__init__.py
timoludwig: The data should be passed as dict, so `return JsonResponse(['Invalid request.'], status=405, safe=False)`I think (but I didn't try it out)...
The page feedback endpoint was implemented during the expedia hackathon, where we didn't have much time for in-depth code reviews. It turns out, that we missed some error there.

src/api/v3/feedback/__init__.py
timoludwig: Ok that seems strange, whenever I used `request.GET.copy()`, it worked without problems. Are you sure it returned a list or dict? Because [the official python documentation](https://docs.python.org/3/library/stdtypes.html#dict.pop) does not describe a possibility how `pop()` should return anything other than a single value... (except for the case that this value is in fact a list :sweat_smile:)

The page feedback endpoint was implemented during the expedia hackathon, where we didn't have much time for in-depth code reviews. It turns out, that we missed some error there.

src/api/v3/feedback/__init__.py
timoludwig: Sorry, `['Invalid request.']` is of course a list and not a dict :man_facepalming: 
So, it would something like `{'error': 'Invalid request.'}`

src/api/v3/feedback/__init__.py
Thomas125: Ah, `request.GET/request.POST` isn't a dict but a `django.http.QueryDict`. Apparently, it is possible for such an object to contain multiple elements with the same key.
If a key exists multiple times, `get()` returns an arbitrary object, but `pop()` returns a  list of all objects :laughing: 
https://docs.djangoproject.com/en/3.0/ref/request-response/#django.http.QueryDict.pop

So, I think `data.pop('key')[0]` with a check if the array contains at least one element should work.

timoludwig: Ah, interesting! Yes, that should work!

Thomas125: `request.POST.copy().dict()` returns the QueryDict as a normal dict (eliminating multiple keys by choosing one arbitrary value). I think this is the best way.

Thomas125:
Is that pylint-fail really because I have a function with 6 arguments?
Can we turn off that rule? :laughing:

timoludwig:
Yes, I think this rule is a bit too restrictive. You can edit `src/.pylintrc` and change `max-args=5` to `max-args=7` or so. In cases, where we still need more arguments, we can use `# pylint: disable=too-many-arguments`...
I think the general rule makes sense, because in most cases many arguments can be refactored into a more readable form. But in your case I think it's fine :smile:

src/api/v3/feedback/search_result_feedback.py
```diff
@@ -0,0 +1,13 @@
...
+@feedback_handler
+# pylint: disable=unused-argument
+def search_result_feedback(data, region_slug, language_code, comment, emotion, is_technical):
+    query = data.get("query")
+    if not query:
+        return JsonResponse({'error': 'Search query is required.'}, status=400)
+    SearchResultFeedback.objects.create(searchQuery=query, emotion=emotion, comment=comment, is_technical=is_technical)
+    return JsonResponse({}, status=201)
```
timoludwig, line 13:
```suggestion
    return JsonResponse({'success': 'Feedback successfully submitted'}, status=201)
```
You could also add a success message here... sorry I didn't mention this earlier.

PR #379, opened 2020-04-24
Configure pylint for whole project

timoludwig:
- Move `.pylintrc` from `src`-directory to top level
- Ignore `.venv` and `node_modules` in `.pylintrc`
- Fix indendation error in `setup.py`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,9 +55,7 @@ jobs:
           at: .
       - run:
           name: Run pylint
-          command: |
-            cd src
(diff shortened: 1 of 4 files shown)


timoludwig:
I'm not sure if this makes sense. The pylintrc config is very comprehensive, so maybe it would bloat the setup.cfg file (the same applies to e.g. the `sphinx/conf.py`). I think both approaches have their pros and cons, I also like it to have one config per service, instead of one huge config file with everything mixed together... let's discuss this in the call.

PR #388, opened 2020-05-09
Fix .venv directory if path contains space

timoludwig:
Pipenv apparently has problems with paths which contain spaces :roll_eyes: 
My solution is to change the `.venv`-directory in the project directory to something auto-generated by pipenv (something like `~/.local/share/virtualenvs/django-4X_NwCeL`) which should not contain spaces... This is done automatically if the env variable `PIPENV_VENV_IN_PROJECT` is not set.


dev-tools/install.sh
```diff
@@ -46,4 +46,8 @@ fi
 
 cd $(dirname "$BASH_SOURCE")/..
 npm install
-PIPENV_VENV_IN_PROJECT=1 pipenv install
+# Check if working directory contains space (if so, pipenv in project won't work)
+if ! pwd | grep -q " "; then
```
thimmy687, line 50: should we add a notification here?

timoludwig: Good idea! I added a small warning.

Issue #325, opened 2020-03-07
Improve documentation of cms models


timoludwig:
Improve documentation of cms models with docstrings in the following format:
```
"""
[Summary]

:param [ParamName]: [ParamDescription], defaults to [DefaultParamVal]
:type [ParamName]: [ParamType](, optional)
...
:raises [ErrorType]: [ErrorDescription]
...
:return: [ReturnDescription]
:rtype: [ReturnType]
"""
```

PR #389, opened 2020-05-11
Improve model documentation

timoludwig:
This is the PR you all have been waiting for :tada: 
1600 lines of (more or less) pure documentation :heart_eyes: 

Fixes: #325
Depends on: #399

### DIFF
--- a/src/cms/models/events/event.py
+++ b/src/cms/models/events/event.py
@@ -1,8 +1,3 @@
-"""Class defining event-related database models
-
-Raises:
-    ValidationError: Raised when an value does not match the requirements
-"""
 from datetime import datetime, time, date
 from dateutil.rrule import weekday, rrule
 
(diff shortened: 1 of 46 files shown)


src/cms/models/events/event.py
```diff
@@ -46,34 +62,77 @@ def languages(self):
...
+
+        :return: Whether the event is recurring or not
+        :rtype: bool
+        """
         return bool(self.recurrence_rule)
 
     @property
     def is_all_day(self):
+        """
+        This property checks whether an event takes place the whole day by checking if start time in minimal and end
```
florianhofhammer, line 76: Typo: should be "start time is minimal"

timoludwig: Well spotted! It seems that you actually read all of these 1,6k lines  :laughing:

florianhofhammer: Not at all - I just started reading it! :laughing:

src/cms/models/languages/language_tree_node.py
```diff
@@ -37,45 +58,122 @@ class LanguageTreeNode(MPTTModel):
...
         """
+        Counts how many ancestors the node has. If the node is the root node, its depth is `0`.
 
+        :return: The depth of this language node
+        :rtype: str
+        """
         return len(self.get_ancestors())
 
+    # Explicitly define functions to show documentation of base model
+    @raise_if_unsaved
```
florianhofhammer, line 120: Just something that looks weird to me: why does useless-super-delegation have to be disabled for pylint for some functions but not for others? The functions below should theoretically all trigger this pylint warning/error, or am I missing something here?

timoludwig: Actually, I have no idea why. I think it's a bug in pylint :sweat_smile: 
When I ran pylint, only the three functions caused a warning and I don't see any reason why. But maybe I'm missing something.

src/cms/models/events/event.py
```diff
@@ -15,19 +10,33 @@
...
+    :param end_time: The time when the event ends
+    :param picture: The thumbnail image of the event
+    :param archived: Whether or not the event is archived
+
+    Relationship fields:
+
+    :param region: The region to which the event belongs (related name: ``events``)
+    :param location: The point of interest where the event takes place (related name: ``events``)
+    :param recurrence_rule: If the event is recurring, the recurrence rule contains all necessary information on the
+                            frequency, interval etc. which is needed to calculate the single instances of an recurring
```
florianhofhammer, line 29: Typo: "a recurring" instead of "an recurring" (yes, I'm pedantic from time to time :smile:)

src/cms/models/pages/page_translation.py
```diff
@@ -146,25 +263,94 @@ def combined_text(self):
 
     @classmethod
     def get_translations(cls, region, language):
+        """
+        This property requests retrieves the most recent versions of a :class:`~cms.models.regions.region.Region` and a
+        specific :class:`~cms.models.languages.language.Language`
```
florianhofhammer, line 268: This description doesn't sound right to me... I suggest something like
```suggestion
        This method retrieves the most recent versions of all :class:`~cms.models.pages.PageTranslation` objects of a 
        :class:`~cms.models.regions.region.Region` in a specific :class:`~cms.models.languages.language.Language`
```

If I understood correctly what the method's effect is, I think this description should fit the method better.

src/cms/models/users/organization.py
```diff
@@ -3,6 +3,20 @@
...
+    :param id: The database id of the organization
+    :param name: The name of the organization
+    :param slug: The slug of the organization
+    :param thumbnail: The url to the thumbnail image of the organization
+    :param created_date: The date and time when the organization was created
+    :param last_updated: The date and time when the organization was last updated
+
+    Reverse relationships:
+
+    :param members: All users who are members of organizations
```
florianhofhammer, line 18: Just a small change proposal concerning the wording as it's about the members of a specific organization:
```suggestion
    :param members: All users who are members of the organization
```

Issue #375, opened 2020-04-23
Automated doc updates


thimmy687:
The running build job should be able to _update_ and _commit_ the latests docs automatically.

timoludwig:
Very good idea! :+1:

PR #391, opened 2020-05-20
Automatically update documentation via CircleCI

timoludwig:
This PR adds the CircleCI config to automatically build the documentation and deploy it to GitHub-Pages: https://integreat.github.io/cms-django
The commits are authored by the user @IntegreatAPI.

The html files are now deleted from the repository, if somebody wants to read or test the documentation locally, he can still use the command `./dev-tools/generate_documentation.sh`.

Fixes: #375


.circleci/config.yml
```diff
@@ -1,5 +1,8 @@
 version: 2.1
 
+orbs:
+  gh-pages: sugarshin/gh-pages@volatile
```
maxammann, line 4: Beaware that sugarshin has access to all secrets of this project :)

timoludwig: Oh, I wasn't aware of that :scream: 
Then it's better to just copy their orb source into our config?

timoludwig: Ok, I added the relevant parts to our jobs, now it is more clear what happens and we don't depend on the third-party orb anymore.

PR #392, opened 2020-05-23
Auth user views

joelhmy:
adjust the UI


timoludwig:
Please select `digitalfabrik:develop` as base instead of `Integreat:develop` :slightly_smiling_face:

Issue #406, opened 2020-06-03
Add docker images for CircleCI build to git


timoludwig:
At the moment, the custom docker images for our CircleCI build are managed by me locally and pushed to [my private docker hub account](https://hub.docker.com/u/timoludwig).
We should move these docker files to this repository (probably into the .circleci-directory) and create a docker hub account for our organization.

PR #426, opened 2020-06-27
Add docker images for CircleCI build

timoludwig:
This adds
- The Dockerfiles
- The CircleCI config to build these images automatically and push them to the Docker Hub

Fixes: #406

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,7 +82,7 @@ jobs:
           path: htmlcov
   check-translations:
     docker:
-      - image: timoludwig/python-node-gettext:latest
+      - image: integreat/python-node-gettext:78c07dc
     resource_class: small
     steps:
       - checkout
@@ -93,7 +93,7 @@ jobs:
           command: ./dev-tools/check_translations.sh
   bundle-static-files:
     docker:
-      - image: timoludwig/python-node-gettext:latest
+      - image: integreat/python-node-gettext:78c07dc
     resource_class: small
     steps:
       - checkout
@@ -118,7 +118,7 @@ jobs:
             - src/cms/locale/de/LC_MESSAGES/django.mo
   packaging:
     docker:
-      - image: timoludwig/bionic-setuptools:latest
+      - image: integreat/bionic-setuptools:78c07dc
     resource_class: small
     steps:
       - checkout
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -188,6 +188,34 @@ jobs:
...
+                docker build -t integreat/bionic-setuptools:$(echo $CIRCLE_SHA1 | cut -c -7) -t integreat/bionic-setuptools:latest .circleci/images/bionic-setuptools
+            fi
+            if git diff --name-only HEAD~1 HEAD | grep -q .circleci/images/python-node-gettext; then
+                docker build -t integreat/python-node-gettext:$(echo $CIRCLE_SHA1 | cut -c -7) -t integreat/python-node-gettext:latest .circleci/images/python-node-gettext
+            fi
+      - run:
+          name: Push Docker images
+          command: |
+            if git diff --name-only HEAD~1 HEAD | grep -q .circleci/images; then
+                echo $DOCKER_PASSWORD | docker login -u integreat --password-stdin
```
svenseeberg, line 211: Does the `$DOCKER_PASSWORD` variable magically exist in Circle CI environments?

.circleci/config.yml
```diff
@@ -82,7 +82,7 @@ jobs:
           path: htmlcov
   check-translations:
     docker:
-      - image: timoludwig/python-node-gettext:latest
+      - image: integreat/python-node-gettext:78c07dc
```
svenseeberg, line 85: @timoludwig could you please explain why the version for the container is now fixed instead of `latest`?

.circleci/config.yml
timoludwig: The password is injected via a [CircleCI Context](https://app.circleci.com/settings/organization/github/Integreat/contexts).
I will explain this in detail when fixing #427 (Add CircleCI documentation).

.circleci/config.yml
timoludwig: This is done to enable a safe update process.
The images will be pushed to docker hub every time the Dockerfile changes, even on a feature branch. So I added a fixed tag which makes sure that other builds on other branches are still deterministic even if we work on another version of Dockerfiles in another branch. The alternative would have been to only push Dockerfiles when the changes are merged into develop, but then we can't test whether the images work because they are not available on feature branches...
So now, when we want to change something in our Dockerfiles, we

1. Commit the changes to the Dockerfiles
2. Look at the hash of the commit and insert it in the circleci config (and commit this as well)
3. Push the changes and see whether the build is successful on the new feature branch, while all other branches including develop and master still use the old docker images & config.

svenseeberg: I see. Thanks.

PR #432, opened 2020-07-12
Change verbosity for pipenv in venv

svenseeberg:
* If pipenv is running in a venv, it produces lots of noice.
  These notifications are only notices and spam the window.


svenseeberg:
There is a better environment variable to silence only warnings directly in relation to Virtualenv. I updated the PR.

PR #435, opened 2020-07-15
Fix non-existing language bug

### LINKED ISSUE #434: Manual creation of Languages results in Bugsplash at page view.
### Describe the Bug
If a system is set up manually without the test data. The manual language setup results in a error when opening the list view of the pages.
grafik

### Steps to Reproduce

1. Create Languages (Like German with code DE_de)
2. Create Region (Like Berlin)
3. Add Language Tree model like German -> English
4. Click on Pages and see error

### Expected Behavior
The usual list should be displayed. 

### Actual Behavior
Error message from Django.

### Additional Information
I guess this has something to do with manually setting the language code and this can't be matched by django.

timoludwig:
- This allows backend languages which are not contained in the content languages or have a different language code
- The table column of translation titles in the current backend languages is now hidden when the backend language does not exist in the content languages

Fixes #434

### DIFF
--- a/src/cms/locale/de/LC_MESSAGES/django.po
+++ b/src/cms/locale/de/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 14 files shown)


timoludwig:
WTF? Why is the PR not updated when I force-push to the branch? :thinking:

PR #442, opened 2020-07-23
Improve documentation of forms

### LINKED ISSUE #324: Improve documentation of forms
Improve documentation of the cms forms package, especially add additional docu pages about the handling of AJAX-form-manipulation and explain the interaction between form models, form views and form templates.
Add docstrings of the following format:
```
"""
[Summary]

:param [ParamName]: [ParamDescription], defaults to [DefaultParamVal]
:type [ParamName]: [ParamType](, optional)
...
:raises [ErrorType]: [ErrorDescription]
...
:return: [ReturnDescription]
:rtype: [ReturnType]
"""
```

timoludwig:
This makes the form docstrings more consistent.

Unfortunately, the form fields for model forms are still missing because there is a bug in our dependency: edoburu/sphinxcontrib-django#27. I opened an upstream PR: edoburu/sphinxcontrib-django#28. In the case that this PR is not merged, we could think about adding a patch to our repo or installing our custom fork of sphinxcontrib-django as dependency.

Fixes #324

### DIFF
--- a/src/cms/_tests.py
+++ b/src/cms/_tests.py
@@ -1,4 +1,4 @@
-from django.contrib.auth.models import User
+from django.contrib.auth import get_user_model
 from django.test import TestCase
 
 from .constants import position, status, region_status
@@ -51,7 +51,7 @@ class SetupClass(TestCase):
         return None
 
     def setUp(self):
-        self.user = User.objects.create_user('test', 'test@integreat.com', 'test')
+        self.user = get_user_model().objects.create_user('test', 'test@integreat.com', 'test')
         self.region = self.create_region({
             'name': 'demo',
             'events_enabled': True,
--- a/src/cms/forms/push_notifications/__init__.py
+++ b/src/cms/forms/push_notifications/__init__.py
@@ -1,1 +1,4 @@
+"""
+Forms for creating and modifying push notification and push notification translation objects
+"""
 from .push_notification_form import PushNotificationForm, PushNotificationTranslationForm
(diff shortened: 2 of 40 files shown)


src/cms/forms/push_notifications/__init__.py
```diff
@@ -1 +1,4 @@
+"""
+Forms for creating and modifying push notification and  push notification translation objects
```
florianhofhammer, line 2:
```suggestion
Forms for creating and modifying push notification and push notification translation objects
```

src/cms/_tests.py
```diff
@@ -4,7 +4,7 @@
 from zipfile import ZipFile
 from bs4 import BeautifulSoup
 
-from django.contrib.auth.models import User
+from django.contrib.auth import get_user_model
```
florianhofhammer, line 7: Seeing that you changed the `User` import to `get_user_model`, we should maybe think about following the official Django suggestion of implementing a custom user model to ease later extensibility (see [the official Django documentation](https://docs.djangoproject.com/en/2.2/topics/auth/customizing/#using-a-custom-user-model-when-starting-a-project)) as long as we haven't deployed the project yet and can still easily replace the user model.
Could this maybe also allow us to drop the `UserProfile` model?

timoludwig: Yeah, we should consider this option.
My initial thought when creating the UserProfile model was that that we could use the User model solely for authentication purposes and the Profile model for functional extensions. But you're right, most of docs & tutorials seem to recommend a custom user model.

PR #461, opened 2020-08-01
Add language support for Push Notifications, fixes #311

### LINKED ISSUE #311: Send push notifications in different languages
At the moment, push notifications can only be sent in one language and don't have translations. This should be changed to reflect the current features of WordPress.

svenseeberg:
## Changes
* Use basic design of PageTranslations form
* Use formsets to enter all translations at once
* Add send modes: only available translations, use main language

### DIFF
--- a/src/cms/models/push_notifications/push_notification_channel.py
+++ b/src/cms/models/push_notifications/push_notification_channel.py
@@ -8,7 +8,7 @@ class PushNotificationChannel(models.Model):
     :param channel: The channel of the push notification
     """
 
-    channel = models.CharField(max_length=60)
+    name = models.CharField(max_length=60)
 
     def __str__(self):
         """
@@ -17,4 +17,4 @@ class PushNotificationChannel(models.Model):
         :return: channel name
         :rtype: str
         """
-        return self.channel
+        return self.name
--- a/src/cms/models/push_notifications/push_notification_channel.py
+++ b/src/cms/models/push_notifications/push_notification_channel.py
@@ -1,0 +1,20 @@
+from django.db import models
+
+
+class PushNotificationChannel(models.Model):
+    """
+    Data model representing a push notification channels
+
+    :param channel: The channel of the push notification
+    """
+
+    channel = models.CharField(max_length=60)
+
+    def __str__(self):
+        """
+        This overwrites the default Python __str__ method, returns the channel name
+
+        :return: channel name
+        :rtype: str
+        """
+        return self.channel
--- a/src/cms/models/push_notifications/push_notification_translation.py
+++ b/src/cms/models/push_notifications/push_notification_translation.py
@@ -19,8 +19,8 @@ class PushNotificationTranslation(models.Model):
                      ``push_notification_translations``)
     """
 
-    title = models.CharField(max_length=250)
-    text = models.CharField(max_length=250)
+    title = models.CharField(max_length=250, blank=True)
+    text = models.CharField(max_length=250, blank=True)
     language = models.ForeignKey(
         "Language",
         related_name="push_notification_translations",
@@ -51,3 +51,4 @@ class PushNotificationTranslation(models.Model):
         """
 
         default_permissions = ()
+        unique_together = ["push_notification", "language"]
--- a/src/cms/templates/push_notifications/push_notification_form.html
+++ b/src/cms/templates/push_notifications/push_notification_form.html
@@ -4,10 +4,26 @@
 {% load static %}
 {% load widget_tweaks %}
 {% load push_notification_filters %}
+{% load page_filters %}
+{% load rules %}
+{% load compress %}
+
+<script>
+    function switch_language(language) {
+        var languages = [{% for lang in languages %}'{{ lang.id }}', {% endfor %}];
+        console.log(languages);
+        languages.forEach(function(language){
+            u("#tab-"+language).addClass("hidden");
+            u("#li-"+language).removeClass("z-10");
+        });
+        u("#tab-"+language).removeClass("hidden");
+        u("#li-"+language).addClass("z-10");
+}
+</script>
 <form method="post">
     {% csrf_token %}
     <div class="flex flex-wrap mb-4">
-        <div class="w-2/5 flex flex-wrap flex-col justify-center">
+        <div class="w-4/5 flex flex-wrap flex-col justify-center">
             <h2 class="heading font-normal">
                 {% if push_notification %}
                     {% if push_notification_translation_form.title.value %}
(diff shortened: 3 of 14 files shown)


src/cms/models/push_notifications/push_notification_translation.py
```diff
@@ -19,8 +19,8 @@ class PushNotificationTranslation(models.Model):
                      ``push_notification_translations``)
     """
 
-    title = models.CharField(max_length=250)
-    text = models.CharField(max_length=250)
+    title = models.CharField(max_length=250, blank=True)
+    text = models.CharField(max_length=250, blank=True)
```
timoludwig, line 23: Are you sure that we should allow empty fields here?

svenseeberg: Well, at least the current Wordpress behavior is this: if a translation for a language does not exist, the source language will be used. And if I do not support blank title & text, then the formset validation will almost always fail.

timoludwig: But wouldn't it be better not to save empty translations instead of storing empty objects in the database? Also, we can't use the source language if we "think" that the translation exists (and it's just empty)...
But if there's no way to save only the forms of a formset which were valid, I guess we can live with this downside.

svenseeberg: The alternative is overriding the validation checks. Not sure if there is any advantage doing it this way. In both options I'd need to independently check if there is a `title` or `text` string available, and if not, use the source language for one or both of them.

timoludwig: Should it really be an option to half-translate a PN? Imo, if e.g. the title is translated and the body isn't, I wouldn't send PNs with mixed languages, but raise an error so the sender can fill in both fields. Only if none of them is filled, it makes sense to fallback to the source language...
And then it does make a difference, because it is cleaner to check whether an object exists in the database vs. to check whether an existing object has only empty fields...

svenseeberg: There are more problems: if the model is invalid with a missing title, then handling the formset becomes complicated. When loading the formset for editing, it is easy to load existing objects for a model into the formset. However, we would then have to compute which languages are missing and create an initial dict to fill the vacant translations. While this is possible, I'd argue its much easier to understand (and compute) to simply store empty translations. And because we actually allow empty translations, this is not that wrong at all. Isn't it?

src/cms/views/push_notifications/push_notification_sender.py
```diff
@@ -1,43 +1,116 @@
...
+        Send all prepared push notification translations
+
+        :return: Success status
+        :rtype: bool
+        """
+        auth_key = self.get_auth_key()
+        status = True
+        for pnt in self.prepared_pnts:
+            payload = {
+                "to": f"/topics/{self.push_notification.region.id}-{pnt.language.code}-{self.push_notification.channel}",
```
svenseeberg, line 98: We defined the channels with the Wordpress Blog ID. This now poses a problem. We have 2 choices: we either import the Wordpress sites in a way that preserves their blog ID in Django. Alternatively, we can add a `blog_id` attribute to our region model.

src/cms/models/push_notifications/push_notification_channel.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class PushNotificationChannel(models.Model):
+    """
+    Data model representing a push notification channels
+
+    :param channel: The channel of the push notification
+    """
+
+    channel = models.CharField(max_length=60)
```
timoludwig, line 11:
```suggestion
    name = models.CharField(max_length=60)
```
Since the model is already the channel, I think it's inconsistent to have also a `channel` attribute...
This would result in e.g. `push_notification.channel.channel` instead of `push_notifications.channel.name` etc...<
But that's just my preference...

src/cms/templates/push_notifications/push_notification_form.html
```diff
@@ -4,10 +4,26 @@
...
+        var languages = [{% for lang in languages %}'{{ lang.id }}', {% endfor %}];
+        console.log(languages);
+        languages.forEach(function(language){
+            u("#tab-"+language).addClass("hidden");
+            u("#li-"+language).removeClass("z-10");
+        });
+        u("#tab-"+language).removeClass("hidden");
+        u("#li-"+language).addClass("z-10");
+}
+</script>
```
timoludwig, line 22: I'd propose to outsource any js code which is more than 2 lines to additional js files which are then included in the template instead of inline code.

src/cms/views/push_notifications/push_notification_sender.py
```diff
@@ -1,43 +1,133 @@
...
+        self.load_secondary_pnts()
+        self.auth_key = self.get_auth_key()
+
+    def load_secondary_pnts(self):
+        """
+        Load push notification translations in other languages
+        """
+        secondary_pnts = PushNotificationTranslation.objects.filter(
+            push_notification=self.push_notification
+        ).exclude(id=self.primary_pnt.id)
```
timoludwig, line 47: just a side note for the future: you can also use the reverse related name in such cases: `self.push_notification.translations.exclude(id=self.primary_pnt.id)`

src/cms/templates/push_notifications/push_notification_form.html
```diff
@@ -22,76 +38,71 @@ <h2 class="heading font-normal">
...
-                        {{ language.name }}
+        <div class="w-2/3 flex flex-wrap flex-col pr-2 tabbed">
+            {{page_form.errors}}
+            {{page_translation_form.errors}}
+            <ul class="flex flex-wrap pl-4">
+                {% for other_language in languages %}
+                <li id="li-{{ other_language.id }}" class="mr-1 {% if other_language.id == language.id %}z-10{% endif %}" style="margin-bottom: -2px" onclick="switch_language('{{ other_language.id }}')">
+                        <div class="bg-white text-blue-500 hover:bg-blue-500 hover:text-white border-l-2 border-t-2 border-r-2 border-blue-500 font-bold rounded-t-lg">
+                            <div class="border-b-2 border-white">
+                                <div class="p-4">{{ other_language }}</div>
```
timoludwig, line 53:
```suggestion
                                <div class="p-4">{{ other_language.translated_name }}</div>
```
I propose to use `other_language.translated_name` here to display the name of the language in the current backend language.

src/cms/views/push_notifications/push_notification_sender.py
```diff
@@ -1,43 +1,133 @@
...
+        :return: FCM API auth key
+        :rtype: str
+        """
         fcm_auth_config_key = "fcm_auth_key"
-        auth_key = Configuration.objects.filter(key=fcm_auth_config_key)
-        if auth_key.exists():
+        auth_key = Configuration.objects.filter(key=fcm_auth_config_key).first().value
+        if auth_key:
             self.logger.info("Got fcm_auth_key from database")
+            return auth_key
```
timoludwig, line 85: If the auth key does not exist in the database, this will result in the error: `'NoneType' object has no attribute 'value'`.
So I suggest to keep the previous if-condition.
```suggestion
        auth_key = Configuration.objects.filter(key=fcm_auth_config_key)
        if auth_key.exists():
            self.logger.info("Got fcm_auth_key from database")
            return auth_key.first().value
```

src/cms/views/push_notifications/push_notification_sender.py
```diff
@@ -1,43 +1,133 @@
...
+
+    def send_pn(self, pnt):
+        """
+        Send single push notification translation
+
+        :param pnt: the prepared push notification translation to be sent
+        :type pnt: ~cms.models.push_notifications.push_notification_translation.PushNotificationTranslation
+        """
+        if settings.DEBUG:
+            blog_id = 154  # Testumgebung Blog ID - prevent sending PNs to actual users
```
timoludwig, line 99: We could even add a new setting `TEST_BLOG_ID = 154` or similar in `settings.py` to make it easy to change it (in case it will)...

timoludwig (commented):
This looks good! :+1: 

I didn't test the actual sending of a PN, but the corresponding code looks reasonable.

src/cms/models/push_notifications/push_notification_channel.py
svenseeberg: Valid point.

src/cms/views/push_notifications/push_notification_sender.py
svenseeberg: Done.

src/cms/views/push_notifications/push_notification_sender.py
svenseeberg: Done

src/cms/templates/push_notifications/push_notification_form.html
svenseeberg: Done

src/cms/templates/push_notifications/push_notification_form.html
svenseeberg: Done

src/cms/views/push_notifications/push_notification_sender.py
timoludwig: I think it should be possible to import the WordPress blogs in a way that their blog id matches the region id.
Both database fields are identical (unique int, auto increement etc) and we don't have to merge it with other data sets...

timoludwig (approved):
:tada:

PR #463, opened 2020-08-12
Page order in page form

### LINKED ISSUE #50: Pages: New approach to control the arrangement of pages

timoludwig:
Control arrangement of pages within the page form:

1. Select parent page via dropdown
2. Change order between siblings via drag & drop

Screenshot Integreat CMS

Fixes #50

### DIFF
--- a/src/cms/forms/pages/page_form.py
+++ b/src/cms/forms/pages/page_form.py
@@ -4,13 +4,47 @@ from django import forms
 from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Permission
 from django.db.models import Q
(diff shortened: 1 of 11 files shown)


ulliholtgrave (approved):
Quite a lot, but seems to work and after a rough check everything looks okay to me

PR #468, opened 2020-08-13
Drag & drop in the language tree

### LINKED ISSUE #221: Enable drag and drop in the language tree

timoludwig:
- Refactor js and templatetags of pages
    This allows the scripts for drag & drop and the template filters to be re-used for other tree models (e.g. the language tree node)
- Enable drag and drop in the language tree

Fixes #221


ulliholtgrave (approved):
Your refactoring makes sense to me

PR #473, opened 2020-08-19
Add simple view tests

### LINKED ISSUE #471: Simple view tests
### Motivation

In our ongoing development process, we introduce new bugs from time to time, which results in some basic views not working anymore (e.g. #459, #436, #434,  #342). This is annoying and can easily be avoided without a complex testing-setup.

### Proposed Solution

Just create a few simple test cases which check if every basic view returns the http status code 200. We don't have to add complex form- and data handling yet.

### Alternatives

Just fix bugs when they're discovered...

### Additional Context

[Here](https://stackoverflow.com/questions/17194542/ensuring-all-django-views-return-200)'s an interesting snippet which allows to create multiple test functions to be created automatically from view names.

timoludwig:
-  Improve model permissions
    - Remove permissions for recurrence rule
    - Fix permissions for user management
    - Adapt `roles.json` fixture to changed permissions
- Add simple view tests
    - Check if all admin views and all region views return the HTTP status code 200.
    - Require static files before running tests in CircleCI

Fixes #471


ulliholtgrave (approved):
Solid improvement! :+1:

PR #477, opened 2020-08-26
Sitemap

### LINKED ISSUE #476: Create Sitemap
### Motivation
We need to support search engines indexing our content. Our content must be easily findable for users.

### Proposed Solution
Add a `/sitemap.xml` that contains links to all web pages (pages, events, locations, etc)

timoludwig:
Fortunately, I could use most of the functionality of the inbuilt package `django.contrib.sitemaps`.
However, there are a few things that didn't work out of the box, which required me to adapt quite a few things:
- The sitemaps were only loaded on application startup, but we have dynamic sitemaps when regions/languages are added/deleted/activated/deacticated
- The URLs in the sitemaps always are the same as the cms server, but we want to create sitemaps for the frontend
- Sitemaps are added to the index even if they're empty

Alternative translations are not yet supported by the standard package, but I already added the future interface (see https://github.com/django/django/commit/16218c20606d8cd89c5393970c83da04598a3e04), so once this is packed into a new release, the alternate tags will work as expected.

Fixes #476


ulliholtgrave (approved):
Looks good

PR #483, opened 2020-09-03
Update dependencies

timoludwig:
This should fix our current [security alert](https://github.com/Integreat/cms-django/network/alerts).


ulliholtgrave (approved):
Seems good, except for the super() stuff :D

PR #488, opened 2020-09-18
Add imprint model

### LINKED ISSUE #357: Add ImprintPage model
Add an `Imprint` and `ImprintTranslation` model which should be similar to `Page` and `PageTranslation`.

klinzo:
As this is my first PR in the django repo (and my first every python PR), please review very carefully.
There are several things I'm not quite sure about. I'll add comments for these things.
If I am not following your conventions somehow, please let me know.

Fixes #357

### DIFF
--- a/src/cms/models/pages/__init__.py
+++ b/src/cms/models/pages/__init__.py
@@ -1,5 +1,7 @@
 """
-This package contains both page-related data models:
-:class:`~cms.models.pages.page.Page` and
-:class:`~cms.models.pages.page_translation.PageTranslation`
+This package contains page-related data models:
+
+* :class:`~cms.models.pages.abstract_base_page.AbstractBasePage`
+* :class:`~cms.models.pages.page.Page` and
+  :class:`~cms.models.pages.page_translation.PageTranslation`
 """
--- a/src/cms/models/pages/__init__.py
+++ b/src/cms/models/pages/__init__.py
@@ -5,4 +5,5 @@ This package contains page-related data models:
   :class:`~cms.models.pages.abstract_base_page_translation.AbstractBasePageTranslation`
 * :class:`~cms.models.pages.page.Page` and
   :class:`~cms.models.pages.page_translation.PageTranslation`
+* :class:`~cms.models.pages.imprint_page.ImprintPage`
 """
--- a/src/cms/models/pages/__init__.py
+++ b/src/cms/models/pages/__init__.py
@@ -5,5 +5,6 @@ This package contains page-related data models:
   :class:`~cms.models.pages.abstract_base_page_translation.AbstractBasePageTranslation`
 * :class:`~cms.models.pages.page.Page` and
   :class:`~cms.models.pages.page_translation.PageTranslation`
-* :class:`~cms.models.pages.imprint_page.ImprintPage`
+* :class:`~cms.models.pages.imprint_page.ImprintPage` and
+  :class:`~cms.models.pages.imprint_page_translation.ImprintPageTranslation`
 """
--- a/src/cms/models/pages/__init__.py
+++ b/src/cms/models/pages/__init__.py
@@ -1,7 +1,8 @@
 """
 This package contains page-related data models:
 
-* :class:`~cms.models.pages.abstract_base_page.AbstractBasePage`
+* :class:`~cms.models.pages.abstract_base_page.AbstractBasePage` and
+  :class:`~cms.models.pages.abstract_base_page_translation.AbstractBasePageTranslation`
 * :class:`~cms.models.pages.page.Page` and
   :class:`~cms.models.pages.page_translation.PageTranslation`
 """
--- a/src/cms/models/pages/abstract_base_page.py
+++ b/src/cms/models/pages/abstract_base_page.py
@@ -1,0 +1,140 @@
+from django.db import models
+from django.utils.translation import get_language
+from django.utils import timezone
+
+from ...constants import status
+
+
+class AbstractBasePage(models.Model):
+    """
+    Abstract base class for page and imprint page models.
+
+    :param archived: Whether or not the page is archived
+    :param created_date: The date and time when the page was created
+    :param last_updated: The date and time when the page was last updated
+
+    Fields to be implemented in the inheriting model:
+
+    :param icon: The title image of the page
+    :param region: The :class:`~cms.models.regions.region.Region` to which the page belongs
+    """
+
+    archived = models.BooleanField(default=False)
+    created_date = models.DateTimeField(default=timezone.now)
+    last_updated = models.DateTimeField(auto_now=True)
+
+    @property
+    def icon(self):
+        """
+        The title image of the page
+        To be implemented in the inheriting model.
+        """
+        raise NotImplementedError
+
+    @property
+    def region(self):
+        """
+        The region to which the page belongs (related name: ``pages``)
+        To be implemented in the inheriting model.
+        """
+        raise NotImplementedError
+
+    @property
+    def languages(self):
+        """
+        This property returns a list of all :class:`~cms.models.languages.language.Language` objects, to which a page
+        translation exists.
+
+        :return: list of all :class:`~cms.models.languages.language.Language` a page is translated into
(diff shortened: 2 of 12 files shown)


src/cms/models/pages/abstract_base_page.py
```diff
@@ -0,0 +1,208 @@
...
+    def get_absolute_url(self):
+        """
+        This helper function returns the absolute url to the editing form of a page
+
+        :return: The absolute url of a page form
+        :rtype: str
+        """
+        pass
+
+    # TODO should this be implemented, passed or not be present here at all?
```
klinzo, line 151: Should this be implemented, passed or not be present here at all? Not sure if it is necessary for imprints at all (since there is probably not more than one anyway). Same for `get_archived` below.

src/cms/models/pages/imprint.py
```diff
@@ -0,0 +1,100 @@
...
+    editors = models.ManyToManyField(
+        settings.AUTH_USER_MODEL, related_name="editable_imprints", blank=True
+    )
+    publishers = models.ManyToManyField(
+        settings.AUTH_USER_MODEL, related_name="publishable_imprints", blank=True
+    )
+    created_date = models.DateTimeField(default=timezone.now)
+    last_updated = models.DateTimeField(auto_now=True)
+
+    # TODO implement
```
klinzo, line 50: Should this be implemented in this PR?

klinzo (commented):
I'm getting the following error:
image

Do you have a clue why this is the case?

timoludwig (commented):
> I'm getting the following error:
> image
> 
> Do you have a clue why this is the case?

This error typically appears when the database migrations were not successful and the current database structure differs from the defined models. In your case, the meta class definition is one problem:
```
TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```
I think this will be resolved if your don't inherit from ABC anymore (see my comment below). Apart from that, you probably won't be able to import the test data because of this change in the database, so you will need to create your own database contents. Once the final structure of the models is set, I can adjust the test data.

src/cms/models/pages/imprint_page_translation.py
```diff
@@ -0,0 +1,86 @@
...
+        :return: The permalink of the imprint
+        :rtype: str
+        """
+        return "/".join(
+            filter(
+                None,
+                [
+                    self.page.region.slug,
+                    self.language.code,
+                    "disclaimer",
```
klinzo, line 67: Should this be hardcoded here?

src/cms/models/pages/page_translation.py
```diff
@@ -44,34 +42,12 @@ class PageTranslation(models.Model):
...
-        This property is an alias of the page foreign key and is needed to generalize the :mod:`~cms.utils.slug_utils`
-        for all content types
-
-        :return: The page to which the translation belongs
-        :rtype: ~cms.models.pages.page.Page
-        """
-        return self.page
 
     @property
     def ancestor_path(self):
```
klinzo, line 53: All tree and mirroring related functions and properties are not included in the AbstractBasePage and AbstractBasePageTranslation models and therefore not available for the ImprintPage and ImrpintPageTranslation model. Is this correct or does it have to be available there as well?

Examples:
* ancestor_path, combined_text for translation models
* depth, get_tree, get_mirrored_text, get_previous_sibling for page models

klinzo (commented):
I also didn't add the `@classmethod` from PageTranslation to AbstractBasePageTranslation/ImprintPageTranslation. If they are needed there as well, let me know.
It is kind of weird imo anyway to have these functions as part of the PageTranslation class.

src/cms/models/pages/imprint_page_translation.py
klinzo: Probably not and also probably related to the question regarding tree related functions below: https://github.com/Integreat/cms-django/pull/488#discussion_r491673129

timoludwig (commented):
In general, this looks very good :+1:, Just had some minor comments.

Apart from that, I stumbled over the terminology: [Imprint](https://en.wikipedia.org/wiki/Imprint) is not the correct translation of [Impressum](https://en.wikipedia.org/wiki/Impressum), so I wondered if we might change the wording here to `DisclaimerPage`. Sorry that this was described incorrectly in the issue. What do you think about this?

klinzo:
> In general, this looks very good +1, Just had some minor comments.
> 
> Apart from that, I stumbled over the terminology: [Imprint](https://en.wikipedia.org/wiki/Imprint) is not the correct translation of [Impressum](https://en.wikipedia.org/wiki/Impressum), so I wondered if we might change the wording here to `DisclaimerPage`. Sorry that this was described incorrectly in the issue. What do you think about this?

I think we had the reverse discussion in Nürnberg on the DF Tagung and we decided to use the term `Imprint`. At least we renamed the issue from `Add DisclaimerPage model` to `Add ImprintPage model`. I think @svenseeberg was in favor of `imprint`.
image

Personally, both is fine to me. Maybe also @dkehne has an opinion? At least DeepL translates `Impressum` with `Imprint`.

src/cms/models/pages/__init__.py
```diff
@@ -1,5 +1,7 @@
 """
-This package contains both page-related data models:
+This package contains page-related data models:
+:class:`~cms.models.pages.abstract_base_page.AbstractBasePage` and
 :class:`~cms.models.pages.page.Page` and
 :class:`~cms.models.pages.page_translation.PageTranslation`
+:class:`~cms.models.pages.imprint_page.ImprintPage` and
```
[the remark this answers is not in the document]
klinzo: Good catch :+1:

dkehne:
agree on Imprint instead of Disclaimer.

src/cms/models/pages/__init__.py
```diff
@@ -1,5 +1,8 @@
...
+:class:`~cms.models.pages.imprint_page_translation.ImprintPageTranslation` and
```
timoludwig, line 7: Sorry, didn't notice that `AbstractBasePageTranslation` was also missing and the conjunctions are inconsistent. I propose a bullet list:
```suggestion
This package contains page-related data models:

* :class:`~cms.models.pages.abstract_base_page.AbstractBasePage` and
  :class:`~cms.models.pages.abstract_base_page_translation.AbstractBasePageTranslation`
* :class:`~cms.models.pages.page.Page` and
  :class:`~cms.models.pages.page_translation.PageTranslation`
* :class:`~cms.models.pages.imprint_page.ImprintPage` and
  :class:`~cms.models.pages.imprint_page_translation.ImprintPageTranslation`
```

You can test the resulting docs by running `./dev-tools/generate_documentation.sh` and open `./docs/autogenerated/cms.models.pages.html` in your browser afterwards.

src/cms/models/pages/imprint_page_translation.py
```diff
@@ -0,0 +1,96 @@
...
+    :param version: The revision number of the imprint translation
+    :param minor_edit: Flag to indicate whether the difference to the previous revision requires an update in other
+                       languages
+    :param created_date: The date and time when the imprint translation was created
+    :param last_updated: The date and time when the imprint translation was last updated
+
+    Relationship fields:
+
+    :param page: The imprint pag
+    e the translation belongs to (related name: ``translations``)
```
timoludwig, line 34:
```suggestion
    :param page: The imprint page the translation belongs to (related name: ``translations``)
```

timoludwig (commented):
Ok, then let's keep the name "imprint".

Once the docstrings and the formatting are fixed, the PR can be merged.
Doesn't the pre-commit hook work for you? The automatic formatting tool should make things easier in theory and not cause extra work :sweat_smile:

klinzo:
Which pre commit hook? :thinking:

timoludwig:
> Which pre commit hook? :thinking: 

If you install the cms with the `./dev-tools/install.sh` script or execute `pipenv run pre-commit install`, the tool black should automatically format your code according to our coding style conventions every time you commit something...

klinzo:
Thanks, working now :)

timoludwig (approved):
Thanks! You can not merge your first PR :tada:

PR #489, opened 2020-09-22
Change ordering live content pages, fix #464

### LINKED ISSUE #464: Embedding live content not working
### Describe the Bug

At the moment, embedding live content in pages is not working.

### Steps to Reproduce

1. Go to a region, e.g. "Augsburg"
2. Click on a page, e.g. "Welcome"
3. In the dropdown "Embed live content" select another  region, e.g. "Nürnberg"
4. Nothing happens

### Expected Behavior

The next dropdown with the page selection should appear

### Actual Behavior

Nothing appears
### Additional Information

The console and the network analysis tab reveal the traceback:

```
FieldError at /ajax/get_pages_list

Cannot resolve keyword 'title' into field.

/home/timo/job/integreat/django/src/cms/views/pages/page_actions.py in get_pages_list_ajax

432.    pages = Page.objects.filter(region__id=region_id).order_by("title")
```

So the problem is very straight forward: the `Page` model does not have a `title` attribute, thus we cannot sort by the title.

svenseeberg:
* Fix ording of live content pages by their title

### DIFF
--- a/src/cms/views/pages/page_actions.py
+++ b/src/cms/views/pages/page_actions.py
@@ -509,7 +509,7 @@ def get_pages_list_ajax(request):
         page.save()
         return JsonResponse({"nolist": True})
     region_id = decoded_json["region"]
-    pages = Page.objects.filter(region__id=region_id).order_by("title")
+    pages = Page.objects.filter(region__id=region_id)
     result = []
     for page in pages:
         result.append(
@@ -523,6 +523,7 @@ def get_pages_list_ajax(request):
                 ),
             }
         )
+    result = sorted(result, key=lambda k: k["name"])
     return JsonResponse(result, safe=False)
 
 


src/cms/views/pages/page_actions.py
```diff
@@ -523,6 +523,7 @@ def get_pages_list_ajax(request):
                 ),
             }
         )
+    result = sorted(result, key=lambda k: k['name'])
```
timoludwig, line 526:
```suggestion
    result = sorted(result, key=lambda k: k["name"])
```
Our coding style requires double quotes. If you install the cms with the `./dev-tools/install.sh` script or execute `pipenv run pre-commit install`, you don't have to bother about this and can just use our pre-commit hook to format your changed files.

timoludwig (changes requested):

src/cms/views/pages/page_actions.py
svenseeberg: Easier said than done on my machine :sweat_smile:

timoludwig (approved):

PR #491, opened 2020-09-24
Improve page revision slider

### LINKED ISSUE #487: Improve revision UI
### Motivation

The revision UI works at it is, but it can be further improved.

### Proposed Solution

- Add steps to the slider
- Label the steps with the according revision number
- ~Highlight the most recent public revision (which is shown in the apps)~

Test the solution on different browsers to make sure the experience is similar.

timoludwig:
This adds steps and labels to the revision slider:

Screenshot_2020-09-24 Integreat CMS

However, in my current Chromium version (85.0.4183.121), the `calc()`-statement doesn't work and the steps do not have the correct margins:

revisions_chromium

Does anybody have an idea how to resolve this (or do we just decide to not support this version of Chromium)?

Fixes #487

### DIFF
--- a/src/cms/locale/de/LC_MESSAGES/django.po
+++ b/src/cms/locale/de/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-10-09 10:26+0000\n"
+"POT-Creation-Date: 2020-10-09 10:49+0000\n"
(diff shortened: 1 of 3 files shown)


svenseeberg (approved):
No idea about the Chromium version, but works in Firefox 81.

timoludwig:
Yeah, I found the solution: I was using `<span>`-tags inside the `<datalist>` and `<option>`-tags which is not supported by the official html spec. In Firefox, this worked "accidentally", whereas in Chromium the behaviour was "correct" :sweat_smile: 
I managed to apply the css stylings directly onto the `<datalist>` and `<option>`-tags, which now works in both browsers without a problem :tada:

Issue #492, opened 2020-09-25
Rename points of interest


timoludwig:
### Motivation

The term "points of interest" (POIs) is not recognized as locations on the map.

### Proposed Solution

Change the wording, e.g.
- English: "Locations on map"
- German: "Orte auf Karte"

Yashs911:
Hey, @timoludwig I would like to work on this.

Yashs911:
Can you please guide in which file I need to rename?

timoludwig:
> Hey, @timoludwig I would like to work on this.

Nice, we really appreciate it!

> Can you please guide in which file I need to rename?

The string "Points of Interest" with "Locations on map" in following files:
- [src/cms/templates/_base.html](https://github.com/Integreat/cms-django/tree/develop/src/cms/templates/_base.html)
- [src/cms/templates/pois/poi_list.html](https://github.com/Integreat/cms-django/tree/develop/src/cms/templates/pois/poi_list.html)
- [src/cms/templates/pois/poi_list_archived.html](https://github.com/Integreat/cms-django/tree/develop/src/cms/templates/pois/poi_list_archived.html)

The string "POI" with "Locations on map" (only in strings, not in class names) in the following files:
- [src/cms/views/pois/poi_view.py](https://github.com/Integreat/cms-django/tree/develop/src/cms/views/pois/poi_view.py)
- [src/cms/views/pois/poi_actions.py](https://github.com/Integreat/cms-django/tree/develop/src/cms/views/pois/poi_actions.py)
- [src/cms/views/pois/poi_list_view.py](https://github.com/Integreat/cms-django/tree/develop/src/cms/views/pois/poi_list_view.py)

Please keep an eye on lower/upper case.
After that, execute `./dev-tools/translate.sh` to re-create the language file [src/cms/locale/de/LC_MESSAGES/django.po](https://github.com/Integreat/cms-django/tree/develop/src/cms/locale/de/LC_MESSAGES/django.po).
Then, insert the German translation "Orte auf Karte" in the related `msgstr` sections of the language file.

Feel free to ask any questions, if you feel uncomfortable with the German translations, you can also only change the English strings and we will add the rest. Thanks in advance!

Yashs911:
@timoludwig I made all the changes but I could not run the ./dev-tools/install.sh command hence the ./dev-tools/translate.sh.
I am using windows (cmd prompt) could you please guide me which software I need to download so that I can run ./dev-tools/translate.sh command.
[...]

PR #509, opened 2020-09-26
Rename points of interest

Yashs911:
Issue #492


timoludwig (changes requested):
Thanks for your first-time contribution! :rocket:

Since this feature isn't completed yet, I changed the base to our local feature branch instead of develop.

Please have a look at upper/lower case (e.g. "Archived **L**ocations on map" should be "Archived **l**ocations on map" and singular/plural (e.g. "Location**s on map** was successfully created and published." should be ideally be "Location was successfully created and published." - probably it's better to remove the "on map"-part when the singular is used).

Yashs911:
@timoludwig I made some changes as requested. If any other change is required feel free to point out. I will try my best.

timoludwig (approved):
Thanks! Looks good to me! :+1:

Issue #493, opened 2020-09-25
Add offers to sitemap


timoludwig:
### Motivation

At the moment, only pages, events and pois are included in the sitemap. Offers are missing.

### Proposed Solution

Add all routes to offers to the sitemap. Create a new class `OfferSitemap` in [sitemaps.py](https://github.com/Integreat/cms-django/blob/develop/src/sitemap/sitemaps.py) which inherits from `WebappSitemap` similar to e.g. `PageSitemap`.

The queryset should be:
```
queryset = Offer.objects.all()
```
which should then be filtered in the `__init__()`-method as follows:
```
self.queryset = self.queryset.filter(
    region=self.region
)
```
Then, add the function `get_absolute_url()` to the [Offer model](https://github.com/Integreat/cms-django/blob/develop/src/cms/models/offers/offer.py) which should return the absolute url in the following form:
`<region_slug>/<language_code>/offers/<offer_slug>`

In the end, add `OfferSitemap` to the `sitemaps` list in [src/sitemap/utils.py:27](https://github.com/Integreat/cms-django/blob/d905ab25d8425c829972e70da1e3b9f25ea61536/src/sitemap/utils.py#L27).

Please also add docstrings to all classes and methods.

stanislavkocan:
Hello, I would like to contribute and solve this issue. Can you please assign it to me ?

timoludwig:
Hey @stanislavkocan, cool! :rocket:  
I updated the issue description to give more details on the proposed solution.

stanislavkocan:
Thanks @timoludwig :) I would like to ask you, what should be the language_code ? If I understand correctly, offers are same in English and German version, that's why we filter only by region. It means that same offers should be in both sitemaps just with different language_code ? 

Right now, I set language_code to `self.region.default_language.code` (that was the only way how I knew to get language), but offers appear only in "German" sitemap with code _de-de_.

Thank you :)

timoludwig:
Ah sorry, I forgot about this problem. Yes, the offers are the same for different languages. But we want to include them in all sitemaps for all active languages. One example:

- https://integreat.app/augsburg/de/offers/sprungbrett
- https://integreat.app/augsburg/en/offers/sprungbrett

[...]

PR #511, opened 2020-09-27
Added Offers to sitemap

stanislavkocan:
I added Offers to sitemap. 

This code

`            try:
                url["alternates"] = url["item"].sitemap_alternates
            except AttributeError:
                logger.debug("%s has no attribute sitemap_alternates", url["item"])`

was added because of AttributeError "Offer object has no attribute sitemap_alternates". I know this may not be the best solution, just let me know and I will try to fix it in some other way.

### DIFF
--- a/src/sitemap/sitemaps.py
+++ b/src/sitemap/sitemaps.py
@@ -101,11 +101,12 @@ class WebappSitemap(ABC, Sitemap):
         urls = super()._urls(page, splitted_url.scheme, splitted_url.hostname)
         for url in urls:
             # Add information about alternative languages
-            try:
-                url["alternates"] = url["item"].sitemap_alternates
-            except AttributeError:
-                logger.debug("%s has no attribute sitemap_alternates", url["item"])
+            url["alternates"] = self.sitemap_alternates(url["item"])
         return urls
+
+    # pylint: disable=no-self-use
+    def sitemap_alternates(self, obj):
+        return obj.sitemap_alternates
 
 
 class PageSitemap(WebappSitemap):
@@ -255,7 +256,6 @@ class OfferSitemap(WebappSitemap):
         super().__init__(region, language)
         # Filter queryset based on region
         self.queryset = self.queryset.filter(region=self.region)
-        self.language = language
 
     def location(self, obj):
         """
@@ -265,3 +265,20 @@ class OfferSitemap(WebappSitemap):
         :type obj: ~cms.models.offers.offer.Offer
         """
         return "/" + "/".join([obj.region.slug, self.language.code, "offers", obj.slug])
+
+    def sitemap_alternates(self, obj):
+        """
+        This sitemap_alternates function returns the language alternatives of offers for the use in sitemaps.
+
+        :return: A list of dictionaries containing the alternative translations of offers
+        :rtype: list [ dict ]
+        """
+        return [
+            {
+                "location": f"{WEBAPP_URL}/{self.region.slug}/{language_tree_node.code}/offers/{obj.slug}",
+                "lang_code": language_tree_node.code,
+            }
+            for language_tree_node in self.region.language_tree_nodes.filter(
+                active=True
+            ).exclude(language=self.language)
+        ]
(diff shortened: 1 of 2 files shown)


stanislavkocan:
Fixes #493

src/sitemap/sitemaps.py
```diff
@@ -98,7 +101,10 @@ def _urls(self, page, protocol, domain):
         urls = super()._urls(page, splitted_url.scheme, splitted_url.hostname)
         for url in urls:
             # Add information about alternative languages
-            url["alternates"] = url["item"].sitemap_alternates
+            try:
+                url["alternates"] = url["item"].sitemap_alternates
+            except AttributeError:
+                logger.debug("%s has no attribute sitemap_alternates", url["item"])
```
timoludwig, line 107:
```suggestion
            url["alternates"] = self.sitemap_alternates(url["item"])
```
The sitemap alternates indicate the same content but in a different language. Thus, it makes sense to include them also for offers (sorry that I forgot about this in the issue). See my suggestion below how you could include them.

src/sitemap/sitemaps.py
```diff
@@ -98,7 +101,10 @@ def _urls(self, page, protocol, domain):
...
         return urls
```
timoludwig, line 109:
```suggestion

    # pylint: disable=no-self-use
    def sitemap_alternates(self, obj):
        return obj.sitemap_alternates

```

src/sitemap/sitemaps.py
```diff
@@ -214,3 +220,48 @@ def __init__(self, region, language):
...
+        self.language = language
+
+    def location(self, obj):
+        """
+        This location function returns the absolute path for a given object returned by items().
+
+        :param obj: Objects passed from items() method
+        :type obj: ~cms.models.offers.offer.Offer
+        """
+        return "/" + "/".join([obj.region.slug, self.language.code, "offers", obj.slug])
```
timoludwig, line 267:
```suggestion
        return "/" + "/".join([obj.region.slug, self.language.code, "offers", obj.slug])

    def sitemap_alternates(self, obj):
        return [
            {
                "location": f"{WEBAPP_URL}/{self.region.slug}/{language_tree_node.code}/offers/{obj.slug}",
                "lang_code": language_tree_node.code,
            }
            for language_tree_node in self.region.language_tree_nodes.filter(
                active=True
            ).exclude(language=self.language)
        ]
```
Just an idea how to add sitemap alternates for offers.

A bit of context to understand the `language_tree_nodes`: The language tree is a concept we introduced to manage translations in a pre-defined way (to make sure the initial content is added in the default language and the other translations follow the desired directions).

src/sitemap/sitemaps.py
```diff
@@ -214,3 +220,48 @@ def __init__(self, region, language):
...
+        :type region: ~cms.models.regions.region.Region
+
+        :param language: The language of this sitemap's urls
+        :type language: ~cms.models.languages.language.Language
+        """
+        # Instantiate WebappSitemap
+        super().__init__(region, language)
+        # Filter queryset based on region
+        self.queryset = self.queryset.filter(region=self.region)
+        self.language = language
```
timoludwig, line 258:
```suggestion
```
You can omit this line, since the language attribute already gets assigned in the `__init__()`-method of the [super class](https://github.com/Integreat/cms-django/blob/d905ab25d8425c829972e70da1e3b9f25ea61536/src/sitemap/sitemaps.py#L48).

timoludwig (commented):
Amazing, thank you for your contribution! :heart: :muscle: 

I really like your solution. I just have a small suggestion to add the alternate languages also for offers. In our current version of django, they are not included in the template yet, but soon they will (see https://github.com/django/django/commit/16218c20606d8cd89c5393970c83da04598a3e04).

stanislavkocan:
Thank you @timoludwig for your help and patience, I am still getting used to. I commited changes you mentioned, so I hope it will be Ok now.

timoludwig (approved):
Very cool, thank you! :rocket: :tada:

PR #512, opened 2020-09-30
Change Feedback API

### LINKED ISSUE #22: API: page feedback
Implement endpoint for posting feedback on a page.

The following definition should be supported: https://github.com/Integreat/cms/wiki/REST-APIv3-Documentation#pageeventdisclaimer

The API should accept an JSON object as defined in the above link, but also HTTP POST variables with the same names.

### LINKED ISSUE #23: API: offer feedback
Endpoint for posting feedback on an extra.

https://github.com/Integreat/cms/wiki/REST-APIv3-Documentation#extra

The API should accept an JSON object as defined in the above link, but also HTTP POST variables with the same names.

### LINKED ISSUE #27: API: events feedback
Endpoint for posting feedback on events.

https://github.com/Integreat/cms/wiki/REST-APIv3-Documentation#pageeventdisclaimer

The API should accept an JSON object as defined in the above link, but also HTTP POST variables with the same names.

### LINKED ISSUE #337: Improve documentation of api
Add docstrings to the api package (this means the developer docu, not the actual user docu of the api.
```
"""
[Summary]

:param [ParamName]: [ParamDescription], defaults to [DefaultParamVal]
:type [ParamName]: [ParamType](, optional)
...
:raises [ErrorType]: [ErrorDescription]
...
:return: [ReturnDescription]
:rtype: [ReturnType]
"""
```

### LINKED ISSUE #359: Add imprint feedback
Add the possibility to post feedback to the imprint/disclaimer.

This depends on #357.

svenseeberg:
* Split into three endpoints for pages, events, imprint pages each
* Edit page/event/disclaimer-endpoint to legacy endpoint
* Add one-to-one field for the association of regions to imprint pages
* Require slug as parameter instead of permalink
* Add API documentation

Resolves #22
Resolves #23 
Resolves #27 
Resolves #359 
Resolves #337

### DIFF
--- a/src/api/decorators.py
+++ b/src/api/decorators.py
@@ -11,10 +11,10 @@ def feedback_handler(func):
     Decorator definition for feedback API functions and methods
 
     :param func: decorated function
-    :type func: func
+    :type func: ~collection.abc.Callable
 
     :return: return handle_feedback function
-    :rtype: func
+    :rtype: ~collections.abc.Callable
     """
 
     @csrf_exempt
@@ -30,7 +30,7 @@ def feedback_handler(func):
         :type language_code: str
 
         :return: return decorated function
-        :rtype: func
+        :rtype: ~collections.abc.Callable
         """
         if request.method != "POST":
             return JsonResponse({"error": "Invalid request."}, status=405)
--- a/src/api/decorators.py
+++ b/src/api/decorators.py
@@ -25,9 +25,7 @@ def feedback_handler(func):
         if request.content_type == "application/json":
             data = json.loads(request.body.decode())
         else:
-            data = {}
-            for key in request.POST:
-                data[key] = request.POST.get(key)
+            data = request.POST.dict()
         comment = data.pop("comment", "")
         rating = data.pop("rating", None)
         category = data.pop("category", None)
--- a/src/api/decorators.py
+++ b/src/api/decorators.py
@@ -1,6 +1,8 @@
 from django.http import JsonResponse
 from django.views.decorators.csrf import csrf_exempt
 
+from cms.models import Region, Language
+
 
 def feedback_handler(func):
     @csrf_exempt
@@ -8,24 +10,36 @@ def feedback_handler(func):
         if request.method != "POST":
             return JsonResponse({"error": "Invalid request."}, status=405)
 
-        data = request.POST.dict()
-        comment = data.pop("comment", "")
-        rating = data.pop("rating", None)
-        category = data.pop("category", None)
+        try:
+            region = Region.objects.get(slug=region_slug)
+            language = Language.objects.get(code=language_code)
 
-        if rating not in [None, "up", "down"]:
-            return JsonResponse({"error": "Invalid rating."}, status=400)
-        if comment == "" and not rating:
+            data = request.POST.dict()
+            comment = data.pop("comment", "")
+            rating = data.pop("rating", None)
+            category = data.pop("category", None)
+
+            if rating not in [None, "up", "down"]:
+                return JsonResponse({"error": "Invalid rating."}, status=400)
+            if comment == "" and not rating:
+                return JsonResponse(
+                    {"error": "Either comment or rating is required."}, status=400
+                )
+            if rating == "up":
+                emotion = "POS"
+            elif rating == "down":
+                emotion = "NEG"
+            else:
+                emotion = "NA"
+            is_technical = category == "Technisches Feedback"
+            return func(data, region, language, comment, emotion, is_technical)
+        except Region.DoesNotExist:
             return JsonResponse(
-                {"error": "Either comment or rating is required."}, status=400
+                {"error": f'No region found with slug "{region_slug}"'}, status=404
             )
-        if rating == "up":
-            emotion = "Pos"
-        elif rating == "down":
-            emotion = "Neg"
-        else:
-            emotion = "NA"
-        is_technical = category == "Technisches Feedback"
-        return func(data, region_slug, language_code, comment, emotion, is_technical)
+        except Language.DoesNotExist:
+            return JsonResponse(
+                {"error": f'No language found with code "{language_code}"'}, status=404
+            )
 
     return handle_feedback
--- a/src/api/decorators.py
+++ b/src/api/decorators.py
@@ -7,8 +7,31 @@ from cms.models import Region, Language
 
 
 def feedback_handler(func):
+    """
+    Decorator definition for feedback API functions and methods
+
+    :param func: decorated function
+    :type func: func
+
+    :return: return handle_feedback function
+    :rtype: func
+    """
+
     @csrf_exempt
     def handle_feedback(request, region_slug, language_code):
+        """
+        Parse feedback API request parameters
+
+        :param request: Django request
+        :type request: ~django.http.HttpRequest
+        :param region_slug: slug of a region
+        :type region_slug: str
+        :param language_code: code of a language
+        :type language_code: str
+
+        :return: return decorated function
+        :rtype: func
+        """
         if request.method != "POST":
             return JsonResponse({"error": "Invalid request."}, status=405)
         try:
--- a/src/api/decorators.py
+++ b/src/api/decorators.py
@@ -1,3 +1,5 @@
+import json
+
 from django.http import JsonResponse
 from django.views.decorators.csrf import csrf_exempt
 
@@ -9,30 +11,9 @@ def feedback_handler(func):
     def handle_feedback(request, region_slug, language_code):
         if request.method != "POST":
             return JsonResponse({"error": "Invalid request."}, status=405)
-
         try:
             region = Region.objects.get(slug=region_slug)
             language = Language.objects.get(code=language_code)
-
-            data = request.POST.dict()
-            comment = data.pop("comment", "")
-            rating = data.pop("rating", None)
-            category = data.pop("category", None)
-
-            if rating not in [None, "up", "down"]:
-                return JsonResponse({"error": "Invalid rating."}, status=400)
-            if comment == "" and not rating:
-                return JsonResponse(
-                    {"error": "Either comment or rating is required."}, status=400
-                )
-            if rating == "up":
-                emotion = "POS"
-            elif rating == "down":
-                emotion = "NEG"
-            else:
-                emotion = "NA"
-            is_technical = category == "Technisches Feedback"
-            return func(data, region, language, comment, emotion, is_technical)
         except Region.DoesNotExist:
             return JsonResponse(
                 {"error": f'No region found with slug "{region_slug}"'}, status=404
@@ -41,5 +22,29 @@ def feedback_handler(func):
             return JsonResponse(
                 {"error": f'No language found with code "{language_code}"'}, status=404
             )
+        if request.content_type == "application/json":
+            data = json.loads(request.body.decode())
+        else:
+            data = {}
+            for key in request.POST:
+                data[key] = request.POST.get(key)
+        comment = data.pop("comment", "")
+        rating = data.pop("rating", None)
+        category = data.pop("category", None)
+
+        if rating not in [None, "up", "down"]:
+            return JsonResponse({"error": "Invalid rating."}, status=400)
+        if comment == "" and not rating:
+            return JsonResponse(
+                {"error": "Either comment or rating is required."}, status=400
+            )
+        if rating == "up":
+            emotion = "POS"
+        elif rating == "down":
+            emotion = "NEG"
+        else:
+            emotion = "NA"
+        is_technical = category == "Technisches Feedback"
+        return func(data, region, language, comment, emotion, is_technical)
 
     return handle_feedback
--- a/src/api/decorators.py
+++ b/src/api/decorators.py
@@ -4,6 +4,7 @@ from django.http import JsonResponse
 from django.views.decorators.csrf import csrf_exempt
 
 from cms.models import Region, Language
+from cms.constants import feedback_emotions
 
 
 def feedback_handler(func):
@@ -60,11 +61,11 @@ def feedback_handler(func):
                 {"error": "Either comment or rating is required."}, status=400
             )
         if rating == "up":
-            emotion = "POS"
+            emotion = feedback_emotions.POS
         elif rating == "down":
-            emotion = "NEG"
+            emotion = feedback_emotions.NEG
         else:
-            emotion = "NA"
+            emotion = feedback_emotions.NA
         is_technical = category == "Technisches Feedback"
         return func(data, region, language, comment, emotion, is_technical)
 
--- a/src/api/v3/feedback/event_feedback.py
+++ b/src/api/v3/feedback/event_feedback.py
@@ -23,7 +23,7 @@ def event_feedback(data, region, language, comment, emotion, is_technical):
     :type is_technical: bool
 
     :return: decorated function that saves feedback in database
-    :rtype: func
+    :rtype: ~collections.abc.Callable
     """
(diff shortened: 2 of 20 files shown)


src/api/v3/feedback/page_feedback.py
```diff
@@ -82,10 +94,12 @@ def feedback(request, region_slug, language_code):
...
         potential_page_translation = PageTranslation.objects.get(
-            slug=slug,
-            language__code=language_code,
+            slug=slug, language__code=language_code, page__region=region
         )
-        if potential_page_translation.permalink == feedback_data.permalink:
+        feedback_data.permalink = urlparse(feedback_data.permalink).path
+        if potential_page_translation.permalink.strip(
+            "/"
+        ) == feedback_data.permalink.strip("/"):
```
timoludwig, line 102:
```suggestion
        if potential_page_translation.permalink == feedback_data.permalink.strip("/"):
```
The `potential_page_translation.permalink` cannot have a leading or trailing slash by definition... so no `strip()` necessary (which also allows the condition to be written in one line)

timoludwig (dismissed):
LGTM :+1: 

Just found one optional improvement.

(And we have to keep in mind that the current feedback API does not work with JSON, but with form-data. So we'll have to support this as well.)

svenseeberg:
I did not know about the POST variables. I think we should amend this PR before merging then.

src/api/v3/feedback/page_feedback.py
```diff
@@ -23,32 +24,43 @@ class FeedbackData:
...
+    def __init__(self, page_id, permalink, comment, rating, category):
         self.page_id = page_id
         self.permalink = permalink
         self.comment = comment
-        self.emotion = emotion
+        self.category = category
+        self.rating = rating
 
     @classmethod
     def from_dict(cls, feedback_dict):
```
svenseeberg, line 35: @mael-fosso @emmakemb @AminAlAit as Timo mentioned, we are currently not using the JSON object but HTTP POST variables. So we'd need a second method that reads the POST variables and turns them into the class properties. Does one of you want to take care? Basically its a function `def from_post()` with `request.POST.get("permalink", "")`

Thomas125:
A while ago, I added a feedback-api decorator (https://github.com/Integreat/cms-django/blob/b08f2be60a9db6c8079ee923de2cd8912f550b12/src/api/decorators.py#L5) that is already in use for the feedbacks for event lists, offer lists, region and search result.
It takes care of the attributes comment, rating, category and is_technical (and returns proper error codes if one of these is not correct).
Maybe we can use it here as well.

Thomas125:
I think, I'm ready. Can someone help me, why the checks fail?

svenseeberg:
`docker-images - Unauthorized` - interesting.

timoludwig:
> `docker-images - Unauthorized` - interesting.

I though at least all members of Integreat had the permissions to access the docker secrets in the circleci environment...
I will look into this later, for now I just restarted the job with my permissions.

svenseeberg:
I'm not yet happy about the commit with the message `bugfix`. @Thomas125 do you want to do a rebase? If you like, I can do this as well.

src/api/v3/feedback/event_feedback.py
```diff
@@ -5,9 +5,15 @@
...
-# pylint: disable=unused-argument
 def event_feedback(data, region, language, comment, emotion, is_technical):
+    return event_feedback_internal(
+        data, region, language, comment, emotion, is_technical
+    )
+
+
+def event_feedback_internal(data, region, language, comment, emotion, is_technical):
     event_slug = data.get("slug")
+    print(event_slug)
```
svenseeberg, line 16: Do we need the print?

src/api/decorators.py
```diff
@@ -1,31 +1,45 @@
...
         if request.method != "POST":
             return JsonResponse({"error": "Invalid request."}, status=405)
 
-        data = request.POST.dict()
-        comment = data.pop("comment", "")
-        rating = data.pop("rating", None)
-        category = data.pop("category", None)
+        try:
+            region = Region.objects.get(slug=region_slug)
+            language = Language.objects.get(code=language_code)
```
timoludwig, line 15: I think it's best practise to only include the parts in the `try`-block which can actually fail with the catched exceptions.
Since the `return JsonResponse` causes following code not to be executed, I see no reason why we would need the whole function inside the `try`-block.

src/api/v3/feedback/event_feedback.py
timoludwig: If yes, please consider using `logger.debug()`.

svenseeberg:
If you are reviewing, the following `curl` commands may help you:

```
curl -X "POST" "http://localhost:8000/api/augsburg/de-de/feedback" -H 'Content-Type: application/json' -d '{"permalink": "/augsburg/de-de/willkommen", "comment": "adf", "rating": "down"}'
curl -X "POST" "http://localhost:8000/api/augsburg/de-de/feedback/page" -H 'Content-Type: application/json' -d '{"slug": "willkommen", "comment": "adf", "rating": "down"}'
curl -X "POST" "http://localhost:8000/api/augsburg/de-de/feedback/page"  -d 'slug=willkommen&comment=adf&rating=down' 
```

svenseeberg:
We found one command that produces an error:
`curl -X "POST" "http://localhost:8000/api/augsburg/de-de/feedback/offers" -d 'slug=lehrstellen-radar&comment=adf&rating=down'`

@mael-fosso and @emmakemb are looking into it.

src/api/v3/feedback/page_feedback.py
```diff
@@ -12,22 +12,21 @@ def page_feedback(data, region, language, comment, emotion, is_technical):
 
 
 def page_feedback_internal(data, region, language, comment, emotion, is_technical):
-    page_slug = data.get("slug")
-    if not page_slug:
+    page_slug = data.get("slug", None)
+    if page_slug is None or not page_slug:
```
Thomas125, line 16: This is redundant.
In the get-function, None is the default.
And `page_slug is None` implies `not page_slug`, so it is enough to check `not page_slug`

src/api/v3/feedback/page_feedback.py
```diff
@@ -12,22 +12,21 @@ def page_feedback(data, region, language, comment, emotion, is_technical):
...
-        ).page
-        PageFeedback.objects.create(
-            page=page,
-            language=language,
-            emotion=emotion,
-            comment=comment,
-            is_technical=is_technical,
-        )
-        return JsonResponse({"success": "Feedback successfully submitted"}, status=201)
-    except PageTranslation.DoesNotExist:
```
Thomas125: Is there a difference beteween your and mine solution in terms of behaviour?
Or is it just the case that your solution is cleaner than mine?

src/api/decorators.py
```diff
@@ -41,5 +22,29 @@ def handle_feedback(request, region_slug, language_code):
             return JsonResponse(
                 {"error": f'No language found with code "{language_code}"'}, status=404
             )
+        if request.content_type == "application/json":
+            data = json.loads(request.body.decode())
+        else:
+            data = {}
+            for key in request.POST:
+                data[key] = request.POST.get(key)
```
Thomas125, line 30: The for-loop is redundant.
Just use `data = request.POST.dict()`

src/api/v3/feedback/page_feedback.py
svenseeberg: Yes, its only a matter of style. As Timo wanted it the other way around, I simply changed it :)

src/api/decorators.py
svenseeberg: Does not work for me that way.

src/api/v3/feedback/page_feedback.py
Thomas125: And is there a reason, that you only changed this get-request?

svenseeberg: No, just did not come across the others.

src/api/v3/feedback/page_feedback.py
svenseeberg: Will update.

src/api/decorators.py
svenseeberg: Okay, tried again. It does work. There was probably some other reason for the issues before.

src/api/v3/feedback/page_feedback.py
svenseeberg: There was only one other instance I found: `src/api/v3/feedback/event_feedback.py`. This is now updated as well.

svenseeberg:
We do not use consistent URL endings: should API endpoints end with a `/` or without?

svenseeberg:
> curl -X "POST" "http://localhost:8000/api/augsburg/de-de/feedback/offers" -d 'slug=lehrstellen-radar&comment=adf&rating=down'

The endpoint is supposed to be `/extras`. `curl -X "POST" "http://localhost:8000/api/augsburg/de-de/feedback/extas" -d 'slug=lehrstellen-radar&comment=adf&rating=down'` works.

src/api/decorators.py
```diff
@@ -1,14 +1,54 @@
...
+            return JsonResponse(
+                {"error": f'No language found with code "{language_code}"'}, status=404
+            )
+        if request.content_type == "application/json":
+            data = json.loads(request.body.decode())
+        else:
+            data = request.POST.dict()
         comment = data.pop("comment", "")
         rating = data.pop("rating", None)
         category = data.pop("category", None)
```
mael-fosso, line 54: I found your code shorter than what I was planning to do: 
def from_post(cls, request):
         return cls(
            request.POST.get("id", None),
            request.POST.get("permalink", None),
            request.POST.get("comment", None),
            request.POST.get("emotion", None),
            request.POST.get("category", None),
        )

mael-fosso (approved):

src/api/decorators.py
```diff
@@ -1,14 +1,54 @@
...
 
 def feedback_handler(func):
+    """
+    Decorator definition for feedback API functions and methods
+
+    :param func: decorated function
+    :type func: func
+
+    :return: return handle_feedback function
+    :rtype: func
```
timoludwig, line 17:
```suggestion
    :type func:  ~collections.abc.Callable

    :return: return handle_feedback function
    :rtype:  ~collections.abc.Callable
```
`func` is not a valid python data type. Since `function` isn't either, I propose to use `Callable` in this case...

src/api/decorators.py
```diff
@@ -1,14 +1,54 @@
...
+
+        :param request: Django request
+        :type request: ~django.http.HttpRequest
+        :param region_slug: slug of a region
+        :type region_slug: str
+        :param language_code: code of a language
+        :type language_code: str
+
+        :return: return decorated function
+        :rtype: func
```
timoludwig, line 33:
```suggestion
        :rtype:  ~collections.abc.Callable
```

src/api/v3/feedback/event_feedback.py
```diff
@@ -0,0 +1,71 @@
...
+    :type language: ~cms.models.languages.language.Language
+    :param comment: The comment sent as feedback
+    :type comment: str
+    :param emotion: up or downvote, neutral
+    :type emotion: str
+    :param is_technical: is feedback on content or on tech
+    :type is_technical: bool
+
+    :return: decorated function that saves feedback in database
+    :rtype: func
```
timoludwig, line 26:
```suggestion
    :rtype:  ~collections.abc.Callable
```

src/api/v3/feedback/imprint_page_feedback.py
```diff
@@ -0,0 +1,66 @@
...
+    :type language: ~cms.models.languages.language.Language
+    :param comment: The comment sent as feedback
+    :type comment: str
+    :param emotion: up or downvote, neutral
+    :type emotion: str
+    :param is_technical: is feedback on content or on tech
+    :type is_technical: bool
+
+    :return: decorated function that saves feedback in database
+    :rtype: func
```
timoludwig, line 29:
```suggestion
    :rtype:  ~collections.abc.Callable
```

src/api/v3/feedback/legacy_feedback_endpoint.py
```diff
@@ -0,0 +1,50 @@
...
+    :type language: ~cms.models.languages.language.Language
+    :param comment: The comment sent as feedback
+    :type comment: str
+    :param emotion: up or downvote, neutral
+    :type emotion: str
+    :param is_technical: is feedback on content or on tech
+    :type is_technical: bool
+
+    :return: decorated function that saves feedback in database
+    :rtype: func
```
timoludwig, line 33:
```suggestion
    :rtype:  ~collections.abc.Callable
```

src/api/v3/feedback/page_feedback.py
```diff
@@ -1,102 +1,73 @@
...
+    :type language: ~cms.models.languages.language.Language
+    :param comment: The comment sent as feedback
+    :type comment: str
+    :param emotion: up or downvote, neutral
+    :type emotion: str
+    :param is_technical: is feedback on content or on tech
+    :type is_technical: bool
+
+    :return: decorated function that saves feedback in database
+    :rtype: func
```
timoludwig, line 29:
```suggestion
    :rtype:  ~collections.abc.Callable
```

src/api/v3/pages.py
```diff
@@ -1,9 +1,21 @@
...
 
 from cms.models import Region
 
 
 def transform_page(page_translation):
+    """
+    Function to create a JSON from a single page_translation Object.
+
+    :param page_translation: single page translation object
+    :type offer: ~cms.models.pages.page_translation.PageTranslation
```
timoludwig, line 14:
```suggestion
    :type page_translation: ~cms.models.pages.page_translation.PageTranslation
```

timoludwig (changes requested):

src/api/decorators.py
timoludwig: The first one is still missing

svenseeberg: Damnit, only grepped for `:rtype: func`

svenseeberg: Commit is amended.

src/api/decorators.py
```diff
@@ -20,12 +60,12 @@ def handle_feedback(request, region_slug, language_code):
                 {"error": "Either comment or rating is required."}, status=400
             )
         if rating == "up":
-            emotion = "Pos"
+            emotion = "POS"
```
timoludwig, line 63:
```suggestion
            emotion = feedback_emotions.POS
```
I think hardcoding the values doesn't make sense when we also have the constants. Import them with:
```
from cms.constants import feedback_emotions

File: src/api/decorators.py [hunk 17]
```diff
@@ -20,12 +60,12 @@ def handle_feedback(request, region_slug, language_code):
...
         elif rating == "down":
-            emotion = "Neg"
+            emotion = "NEG"
```
timoludwig, line 65:
```suggestion
            emotion = feedback_emotions.NEG
```

src/api/decorators.py
```diff
@@ -20,12 +60,12 @@ def handle_feedback(request, region_slug, language_code):
...
         else:
             emotion = "NA"
```
timoludwig, line 67:
```suggestion
            emotion = feedback_emotions.NA
```

timoludwig (commented):
Just found one additional improvement :slightly_smiling_face:

src/api/decorators.py
svenseeberg: Good catch!

timoludwig (approved):
I didn't test all of the endpoints, but from the code review alone I think this looks good! :rocket:

Issue #516, opened 2020-10-01
Disable h1-tags in content


timoludwig:
### Motivation

The apps use the `<h1>`-tags to display the title, so the content authors shouldn't use it in their content.

### Proposed Solution

Disable the creation of "Heading 1" in TinyMCE.

If it's not too much effort, we could also prevent the maual creation of `<h1>`-tags in the source code by validating the `text`/`description`-fields in the form's `clean()`-method.

aitorres:
Hello! I could work on this issue and submit a PR in a day or two, if that's okay

muditjain33:
Can I work on this issue?

aitorres:
@timoludwig Hello! I was checking the code and found tinymce is being used across several forms. I have two questions, for now:

1. Should I disable the "Heading 1" style in all places tinymce is being used?
2. I would also like to do the extra `clean()` method validation, so I was wondering what the expected behavior would be: whether to raise a validation error if `<h1>` tags are present in the field, remove any pair of `<h1>...</h1>` and kept the inner text intact, or substitute any `<h1></h1>` tags with, say, `<h2>` equivalent tags.

timoludwig:
>     1. Should I disable the "Heading 1" style in all places tinymce is being used?

Yes please. We wanted to simplify this by moving the tinymce to a separate file which is included from all content forms. For now, you can just copy + paste your solution to all tinymce scripts. (Or do the optimization yourself if you want to)

>     2. I would also like to do the extra `clean()` method validation, so I was wondering what the expected behavior would be: whether to raise a validation error if `<h1>` tags are present in the field, remove any pair of `<h1>...</h1>` and kept the inner text intact, or substitute any `<h1></h1>` tags with, say, `<h2>` equivalent tags.

Good question. I think I'd prefer the first option, so raise an error when the tags are present. Since this can also happen when you copy-paste a formatted text into the field, the error message should also be understandable for someone not knowing html.

aitorres:
Thanks for the clarifications yesterday! I submitted a pull request, I think I got all of it (except for the two things I mention in the PR body). If there's any suggestions, errors or things I should change just let me know!

aitorres:
Oh, and one more thing! (And sorry for the double reply). Yesterday night, DigitalOcean added a new requirement for repositories and pull requests to count towards the Hacktoberfest valid PR count, the information can be found here: https://hacktoberfest.digitalocean.com/hacktoberfest-update

[...]

PR #518, opened 2020-10-03
Disable Heading 1 style in certain TinyMCE fields

aitorres:
### Short description
This PR disables the possibility of using the Heading 1 style in TinyMCE fields for text and descriptions, throwing a validation error if the `<h1>` tags are (manually) inserted into the field's content.

### Proposed changes
First, the PR removes the possibility of selecting "Heading 1" as a style in certain TinyMCE fields by manually overriding which styles are visible in the TinyMCE styles list. The affected tinyMCE instances are in the following templates:

- `cms/templates/events/event_form.html`
- `cms/templates/pages/page_form.html`
- `cms/templates/pages/page_sbs.html`
- `cms/templates/pois/poi_form.html`

Second, the PR adds a specific `clean_<field>` method to the Django Forms being used in the aforementioned templates, that checks for the presence of `<h1>` tags in their fields and, if found, raises a validation error with an appropriate message.

The forms changed are:

- EventTranslationForm (field: description)
- PageTranslationForm (field: text)
- POITranslationForm (field: description)

These three forms are used in the templates above. This way, we prevent the use of "Heading 1" styles both in front-end and in back-end.

### Misc notes

I am not sure about the validation error message I used, which is: "Use of Heading 1 style not allowed."

Any suggestions would be accepted and changed into the Pull Request.

Also, ich spreche nicht Deutsch :-P so I wasn't able to add a properly translated string to the DE locale files. That would be a pending thing to do. If you can provide an equivalent message in Deutsch, I can add it to the proper locale file.

### Resolved issues

Fixes: #516

### DIFF
--- a/src/cms/forms/events/event_translation_form.py
+++ b/src/cms/forms/events/event_translation_form.py
@@ -1,6 +1,8 @@
 import logging
 
 from django import forms
+from django.core.exceptions import ValidationError
+from django.utils.translation import ugettext_lazy as _
 
 from ...constants import status
 from ...models import EventTranslation
@@ -89,3 +91,14 @@ class EventTranslationForm(forms.ModelForm):
         self.data = self.data.copy()
         self.data["slug"] = unique_slug
         return unique_slug
+
+    def clean_description(self):
+        description = self.data["description"]
+
+        if "<h1>" in description:
+            raise ValidationError(
+                _("Use of Heading 1 style not allowed."),
+                code="no-heading-1",
+            )
+
+        return description
--- a/src/cms/forms/pages/page_translation_form.py
+++ b/src/cms/forms/pages/page_translation_form.py
@@ -1,6 +1,8 @@
 import logging
 
 from django import forms
+from django.core.exceptions import ValidationError
+from django.utils.translation import ugettext_lazy as _
 
 from ...constants import status
 from ...models import PageTranslation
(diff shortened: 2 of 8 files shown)


timoludwig (changes requested):
Amazing, thank you for your contribution :heart: 

The form validation works as expected, thanks! However, I didn't get the tinymce setting to work. When I navigate to one of the forms, I can still select "Heading 1" as format. I think this is because we don't use the `formatselect` dropdown, but the `styleselect` instead. Thus, we have to adjust the option `style_formats` instead of `block_formats`. So i propose to remove heading 1 form the [default value of style_formats](https://www.tiny.cloud/docs/configure/editor-appearance/#style_formats) and include this array in the tinymce init function.

As for the translation, I propose the German translation:
> Bitte benutzen Sie keine Überschrift 1 in diesem Feld.

Thanks again for your effort! I added the `hacktoberfest`-tag to our repository to make sure PRs are eligible for this year's challenge.

aitorres:
Hey @timoludwig! Thank you so much for your comments! I'll be doing the proposed changes during the day and I'll let you know.

aitorres:
Hello @timoludwig! Sorry for the delay, I've pushed a couple of commits for both adding the translated string to the deutsch locale and explicitly using `styles_format`, I think this would solve the issue. Let me know if there's anything else I could do!

timoludwig (approved):
Thank you! Now everything works as expected. :+1:

Issue #524, opened 2020-10-07
Language code can contain slashes


timoludwig:
### Describe the Bug

At the moment, the language code field ist just a `CharField` which means it can contain any characters, including e.g. slashes. This is a problem because our routing system relies on the fact that language codes only contain characters and hyphens.

### Steps to Reproduce

1. Go to languages
2. Create a language with slash in code
3. Save

### Expected Behavior

The slash should not be allowed in that field.

### Actual Behavior

The language object can be saved which causes a lot of errors in the whole cms because the urls cannot be matched correctly anymore.

### Additional Information

Please change the `CharField` of the [language model](https://github.com/Integreat/cms-django/blob/b08f2be60a9db6c8079ee923de2cd8912f550b12/src/cms/models/languages/language.py#L32) to a `SlugField` to prevent harmful special characters.

S10MC2015:
I have sent a PR and I think I have done it correctly?

PR #525, opened 2020-10-07
Change things to SlugField

S10MC2015:
Fixes: #524
I don't know if I have done this right but I think I have?


timoludwig (changes requested):
Thanks, this was quite a quick reaction! :rocket: :racehorse: 

See my comment below, but the general solution works :slightly_smiling_face:

timoludwig (approved):
Thanks! :+1:

Issue #384, opened 2020-05-04
Short URL support


svenseeberg:
We may want so share pages with shorter URLs. In some cases, page URLs are very long. Therefore we could use a URL shortening tool, akin to bit.ly (but more privacy friendly). We could either have an integrated app (`shorturl`) that maps URLs like `cms.integreat-app.de/s/TOKEN` to the original URL, or use a completely outsourced tool.

stanislavkocan:
Hello, can you please assign it to me ? I will work on it. 

Is there any share functionality where you would like to include it ? Or how the user gets that short url ? Thank you.

svenseeberg:
I'd suggest that we have a button in the Page edit form that copies the short URL to the clip board. Additionally, we could have a button for this in the page list.

stanislavkocan:
Hello, I tried several methods (without using shortening tool aka bit.ly) but there are few problems. I found a plugin (https://pypi.org/project/short_url/) which takes integer and returns short unique string. The only integer which could be usable is page id, but there would be problems with translations, as they also have same page id. I also tried to encode url to integer, but the integer was very big, and the plugin I mentioned in this case would produce much longer string, which is not suitable for short url.

I plan to implement something like this, but please let me know, if you see there any potential problems which I did not realize: 

1) There would be button in Pages (next to View, Edit, Archive, Delete page) for each page called Copy link.
2) After click on that button for first time for some page, it would generate random short string (with library fe. shortuuid).
3) There would be new CharField in AbstractBasePageTranslation, which would store that short string or some default value in case random string was not generated yet for that page.
4) Copy link button would paste url with returned short string in right format (fe. app/s/dsjk5) to clipboard.
5) When user comes to link app/s/dsjk5, it would take dsjk5, filter the page and get absolute url of that page.
6) There would be redirect function, which would redict user to returned absolute url.

Let me know what you think.  Thanks.

PR #521, opened 2020-10-06
Add short url support

stanislavkocan:
### Short description

This PR adds link for each page in Pages (next to buttons View Page, Archive Page, ...)  which generates unique string of 5 characters and copies URL in format **url/s/<generated_string>** into clipboard. After user navigates to copied URL, it will redirect to corresponding page.

### Proposed changes

- There is link in Pages (next to View, Edit, Archive, Delete page) for each page with title Copy short link.
- In case page has no translation, link is visibly disabled with corresponding title.
- After click on that link for first time for some page, it generates random short string (using [shortuuid](https://pypi.org/project/shortuuid/) library), which is stored as new attribute called short_url_id  in AbstractBasePageTranslation class.
- Default value of attribute is empty string. New short_url_id is generated only when attribute has default value.
- Click on button also copies url in format **url/s/<generated_string>** into clipboard, with message informing user about successfuly copied url (using [pyperclip](https://pypi.org/project/pyperclip/) plugin).
- When user comes to copied short url, there is queryset which filters PageTranslation with given ID.
- User is redirected to page which is given by get_absolute_url() method

### Resolved issues

Fixes: #384

PR #528, opened 2020-10-08
Add short url support

timoludwig:
### Short description

This is essentially a copy of PR #521 by @stanislavkocan. See this PR of a detailed description of the changes.

### Proposed changes
- Added German translations (only review those)

### Resolved issues

Fixes: #384


src/cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -2742,48 +2750,56 @@ msgstr "Organisation wurde erfolgreich erstellt"
...
 
-#: views/pages/page_actions.py:101
+#: views/pages/page_actions.py:104
 msgid "Page was successfully deleted"
 msgstr "Seite wurde erfolgreich gelöscht"
 
-#: views/pages/page_actions.py:229
+#: views/pages/page_actions.py:129
+msgid "URL '{}' was successfully copied to clipboard."
+msgstr "Die URL '{}' wurde erfolgreich in die Zwischenablage kopiert."
```
svenseeberg, line 2771: Do we need the article "Die"? The English sentence does not start with "The" either. I removed the "Der" in front of "Ort" in the previous translation unification PR as well.

svenseeberg (approved):
Approved, just in case.

src/cms/locale/de/LC_MESSAGES/django.po
timoludwig: You're right, let's go for full consistency here :muscle:

PR #534, opened 2020-10-16
PDF Export, fixes #35

### LINKED ISSUE #35: Pages: export PDF
Export selected list of pages into PDF. Option to include table of content. Cache created PDFs.

It is probably a good thing to use the export PDF API for this feature. This ensures, that apps and back end produce the same results.

If possible we can include customization features for regions. For example, the cities could include their logos into the header or footer. This is covered in #213

melegiul:
### Short description
Added button to export pages in PDF format.

### Proposed changes

Uses xhtml2pdf library (based on reportlab) to convert the user input of the page editor (which styles are saved in html format) to the corresponding pdf document. When the selected page is a root page, a table of content is added to first page of the document containing all children of the root page.

### Resolved issues

Fixes: #35


svenseeberg:
@melegiul a note on your commit style: commit message headlines (first line) should have at most 50 characters. It is good practice to start them with one of the following words: Add, Change, Remove. Then only add another 3 to 5 words at most.

Everything else should go into the "body" of a commit message. Basically, just leave one blank line after the header and then put in the rest there. I think it is useful to use a bullet point style.

melegiul:
All right, i will put the messages into the body

svenseeberg:
Thanks for updating the commit messages.

I have one question we should discuss: do we need some sort of PDF cache? I guess that generating PDFs can take some time, especially when selecting many pages. We could basically create a hash across the versions of all included pages and name the created PDF with this hash. We then can store all created PDFs in a cache directory and just need to check if a PDF with this name already exists. If not, we can create it.

Regarding the code: looks good to me so far. I have not found any issues. Performance testing of exports is not so easy yet, as we would need hundreds of content pages for testing. But I really think we need some sort of caching.

src/cms/views/pages/page_actions.py
```diff
@@ -112,6 +116,86 @@ def delete_page(request, page_id, region_slug, language_code):
...
+        "html2pdf": True,
+        "region": region,
+        "page_level": page.level,
+        "pages": pages,
+        "language": language,
+    }
+    response = HttpResponse(content_type="application/pdf")
+    response["Content-Disposition"] = (
+        "attachment; filename=%s.pdf" % page_translation.title
+    )
```
melegiul, line 186: Since the created PDF is offered for download to the user, i suggest to use the local download folder as the cache. So i could just check this folder whether that hash already exists.

melegiul:
> We could create a hash across the slugs and versions of all included pages translations and name the created PDF with this hash.

I have seen that the page model covers an attribute called 'last_updated' containing the time when the page was last updated. Do you think, i could just use these concatenated timestamps + region_slug to create the hash value?

svenseeberg:
> I have seen that the page model covers an attribute called 'last_updated' containing the time when the page was last updated. Do you think, i could just use these concatenated timestamps + region_slug to create the hash value?

I see no problems with that. And I also see no problem with using the normal download directory.

src/cms/views/pages/page_actions.py
timoludwig: I'm not sure if I understood this correctly. Do you mean the local download folder of the user? The "expensive" operation is not the actual download of the file, but the server-side creation of the PDF. So what we want to achieve with the cache is that we store created PDFs on the server, so that when another user wants to download the same PDF, `pisa.CreatePDF` is not called again, but the PDF file from the cache is used. We don't need to take care about when the same user downloads the same file twice.

Pipfile
```diff
@@ -20,9 +20,12 @@ sphinx = "*"
 sphinxcontrib-django = "*"
 sphinx-last-updated-by-git = "*"
 sphinx-rtd-theme = "*"
+reportlab = "*"
+xhtml2pdf = "*"
 
 [packages]
 integreat-cms = {editable = true, path = "."}
+zipp = {editable = true, path = "."}
```
timoludwig, line 28:
```suggestion
```
This should not be there, I think this is a bug of Pipenv which adds this package as local package when new pip packages are added.
You can circumvent this problem if you delete the directories `build`, `.venv` and `src/integreat_cms.egg-info/` before doing the `pipenv install`.

Pipfile
```diff
@@ -20,9 +20,12 @@ sphinx = "*"
 sphinxcontrib-django = "*"
 sphinx-last-updated-by-git = "*"
 sphinx-rtd-theme = "*"
+reportlab = "*"
+xhtml2pdf = "*"
```
timoludwig, line 24:
```suggestion
```
I think reportlab can be removed since it is not used and xhtml2pdf can be removed since it is already contained in `setup.py`. If you freshly install the cms, all packages from setup.py get installed automatically and don't need to be contained in the Pipfile.

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta:__legacy__"
```
timoludwig, line 3: I don't know why this file is required. Did you add it on purpose?

src/cms/templates/_base.html
```diff
@@ -145,7 +145,7 @@
 					<a href="{% url 'pages' region_slug=region.slug %}" class="{% if current_menu_item == 'pages' %}active{% endif %}">
 						{% trans 'All pages' %}
 					</a>
-					{% if region.default_language == language and language != null %}
+					{% if language %}
```
timoludwig, line 148:
```suggestion
					{% if region.default_language == language %}
```
The condition in its original form was added on purpose, because it should prevent to create new pages in other languages than the default language.
So the "Create Page"-button should only appear if the current language is the default language. But the second condition was redundant I think.

src/cms/views/pages/page_actions.py
```diff
@@ -112,6 +116,86 @@ def delete_page(request, page_id, region_slug, language_code):
...
 
+# pylint: disable=unused-argument
+def link_callback(uri, rel):
+    """
+    according to xhtml2pdf documentation this function is neccessary for
+    resolving the django static files references and returning absolute paths
+    to the files
+
+    :param uri: uri that is generated by django template tag 'static'
+    :type uri: string
```
timoludwig, line 127:
```suggestion
    :type uri: str
```
Sphinx is a bit picky when it comes to types :slightly_smiling_face:

src/cms/views/pages/page_actions.py
```diff
@@ -112,6 +116,86 @@ def delete_page(request, page_id, region_slug, language_code):
...
+    :return: the absolute path according to django's static file settings
```
timoludwig, line 128:
```suggestion
    :return: the absolute path according to django's static file settings
    :rtype: str
```

src/cms/views/pages/page_actions.py
```diff
@@ -112,6 +116,86 @@ def delete_page(request, page_id, region_slug, language_code):
...
+@region_permission_required
+# pylint: disable=unused-argument
+def export_pdf(request, region_slug, page_id, language_code):
+    """
+    function for handling a pdf export request for a single page
+    if the page is a root page the corresponding django template
+    inserts the content of all children pages
+
+    :param request: request submitted for rendering pdf document
+    :type request: HttpRequest
```
timoludwig, line 159:
```suggestion
    :type request: ~django.http.HttpRequest
```

src/cms/views/pages/page_actions.py
```diff
@@ -112,6 +116,86 @@ def delete_page(request, page_id, region_slug, language_code):
...
+    :param page_id: database id for selected page
+    :type page_id: integer
```
timoludwig, line 161:
```suggestion
    :type page_id: int
```

src/cms/views/pages/page_actions.py
```diff
@@ -112,6 +116,86 @@ def delete_page(request, page_id, region_slug, language_code):
...
+    :param language_code: bcp47 code of the current language
+    :type language_code: string
```
timoludwig, line 163:
```suggestion
    :type language_code: str
```

src/cms/views/pages/page_actions.py
```diff
@@ -112,6 +116,86 @@ def delete_page(request, page_id, region_slug, language_code):
...
+    :raises PermissionDenied: user login and permissions required
+    :return: pdf document offered for download
+    :rtype: HttpResponse
```
timoludwig, line 166:
```suggestion
    :rtype: ~django.http.HttpRequest
```

timoludwig (commented):
Awesome, thank you for your contribution :heart: 

Apart from my comments, I noticed one problem: It should also be possible to create a pdf with the whole content of one region. Keep in mind that if there are multiple root pages, you also have multiple `tree_id`s in one region's pdf.



> I have seen that the page model covers an attribute called 'last_updated' containing the time when the page was last updated. Do you think, i could just use these concatenated timestamps + region_slug to create the hash value?

In some edge cases, there might be different page translations with the same `last_updated` value. So to be extra sure, include also the ids of this translations.

Pipfile
melegiul: Yes, we also talked about this with @svenseeberg, so i could implement this feature by adding a new action in the footer of the page_tree.html, so that when user sets the checkboxes for all pages, he gets a complete pdf document

melegiul: Referring to your comment: It should also be possible to create a pdf with the whole content of one region. Keep in mind that if there are multiple root pages, you also have multiple tree_ids in one region's pdf.

pyproject.toml
melegiul: No, this file just appeared, when i installed xhtml2pdf. I think it is kind of configuration file for packaging.

src/cms/views/pages/page_actions.py
melegiul: All right, so I will try to figure out, which is the recommended place for such cache inside our project folder.

timoludwig: You could e.g. use `src/cms/static/CACHE`, which is also used by the django-compressor.

src/cms/views/pages/page_actions.py
```diff
@@ -112,6 +118,99 @@ def delete_page(request, page_id, region_slug, language_code):
...
+        # retrieve the translation for each page
+        page_translation = page.get_translation(language_code)
+        pdf_key_list.append(page_translation.id)
+        pdf_key_list.append(page_translation.last_updated)
+    pdf_key_string = '_'.join(map(str, pdf_key_list))
+    pdf_hash_key = hashlib.sha256(bytes(pdf_key_string, 'utf-8')).hexdigest()
+    cached_response = cache.get(pdf_hash_key, 'has_expired')
+    if (cached_response is not 'has_expired'):
+        return cached_response
+    else:
```
melegiul, line 187: So basically this is where the caching happens, in the first lines it's just retrieving all queried pages from database. Then in line 182 all page_translation ids and timestamps are combined to a single string and hashed in the following line.

melegiul:
> It should also be possible to create a pdf with the whole content of one region.

I shifted the PDF export button to the footer of the page_tree.html, so that the user can control the exported pages by setting the checkboxes.

svenseeberg:
I noticed that we implemented a new feature in Wordpress since we opened the issue for Django: regions have the option to include icons in the header.

image

src/cms/views/pages/page_actions.py
```diff
@@ -112,6 +118,99 @@ def delete_page(request, page_id, region_slug, language_code):
...
+        )
+        template = get_template(template_path)
+        html = template.render(context)
+        pisa_status = pisa.CreatePDF(html, dest=response, link_callback=link_callback)
+        if pisa_status.err:
+            return HttpResponse(
+                "There was an error while rendering the following html <pre>"
+                + html
+                + "</pre>"
+            )
```
timoludwig, line 209:
```suggestion
            logger.error(
                "The following PDF could not be rendered: Region: %s, Language: %s, Pages: %s.",
                region,
                language,
                pages,
            )
            return HttpResponse(_("The PDF could not be successfully generated."))
```
The end user should not see the html, all he needs to know it that something went wrong. All debug information should only be logged internally.

timoludwig (commented):
Thanks for the caching functionality, very cool!
Apart from my inline comments, I have three additional comments:
- You have some unresolved conflicts, so please resolve them by rebasing your branch on the current upstream develop branch. If you need help with this, you can also grant me write permissions on your fork.
- The Pipfile.lock is not up to date. Please remove your local `build`, `.venv` and `src/integreat_cms.egg-info` directories and execute `./dev-tools/install.sh`. If you're not sure, I can also do this for you.
- After all your changes are made and the PR is ready to be merged, run `pipenv run black .` to apply our coding style. You can also do this automatically if you use our pre-commit hook.

src/cms/templates/pages/page_view.html
```diff
@@ -1,10 +1,65 @@
...
-<h1>{{page_translation.title}}</h1>
 
-<div style="padding-top: 30px">
-    {{ page_translation.text|safe }}
-</div>
+{% if html2pdf %}
+    {% if amount_pages > 1 %}
+        <!-- if the user selected several pages, additionally insert table of content -->
+        {% get_translation pages|first language.code as page_translation %}
+        <h1 id="title_page">{{page_translation.title}}</h1>
```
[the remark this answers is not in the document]
timoludwig: I added the new issue #550 for this.

melegiul:
> * You have some unresolved conflicts, so please resolve them by rebasing your branch on the current upstream develop branch. If you need help with this, you can also grant me write permissions on your fork.
> 
> * The Pipfile.lock is not up to date. Please remove your local `build`, `.venv` and `src/integreat_cms.egg-info` directories and execute `./dev-tools/install.sh`. If you're not sure, I can also do this for you.
> 
> * After all your changes are made and the PR is ready to be merged, run `pipenv run black .` to apply our coding style. You can also do this automatically if you use our pre-commit hook.

Ok, rebasing, reinstalling and code styling are all done.

melegiul:
I just had to fix a problem with arabic/persian letters during pdf export. Applying a different font family fixed that.

timoludwig (approved):
Thank you for your further improvements! :rocket: :muscle: 

I noticed that I made some mistakes during my review and provided incorrect suggestions.
Since this is my fault, I fixed them and added them to your branch:
- The public translation was only used for the cache hash calculation, not in the actual pdf. I fixed this by adding a new templatetag [get_public_translation](https://github.com/melegiul/cms-django/blob/9264af20fe835c7e0547d9d5367e6a00f235888e/src/cms/templatetags/content_filters.py#L34).
- I forgot to mention our pylint.sh style checker, which required some additional changes.

On doing so, I had a few ideas for more improvements:
- Squash the "Apply suggestions from code review" commits
- Split the templates [page_view.html](https://github.com/melegiul/cms-django/blob/feature/exportpdf/src/cms/templates/pages/page_view.html) and [page_pdf.html](https://github.com/melegiul/cms-django/blob/feature/exportpdf/src/cms/templates/pages/page_pdf.html), since they do not have much in common anymore.
- Add [intersphinx mappings](https://github.com/melegiul/cms-django/blob/9264af20fe835c7e0547d9d5367e6a00f235888e/sphinx/conf.py#L77) to the xhtml2pdf documentation
- Make text aligned to the right on languages which have right-to-left as text direction

If these changes are ok for you, you can hit that merge button :tada:

(In case you want to revert to your previous state, I created the branch [feature/exportpdf-backup](https://github.com/melegiul/cms-django/tree/feature/exportpdf-backup) to make sure nothing is lost.)

melegiul:
Thank you for your review. Nice to see that you solved the problem with the custom font-family inside the css file by re-implementing the link_callback function:tada::muscle:, looks good to me:rocket:. I will also keep in mind that commit squashing trick and the intersphinx mapping.

Issue #533, opened 2020-10-16
Remove shortuuid module


svenseeberg:
Currently, we use the shortuuid to map short urls to page translations. I see 2 problems with that:

1. The page translation ID in the database is already unique. Why do we need an additional module and another database column?  We could argue that this is needed if we want to map to other objects than pages as well. However, it would still be possible to do this via URL encoding, i.e. `https:/HOST/s/p/ID` for pages and `https://HOST/s/e/ID` for events etc. Additionally, this guarantees that we do not get any collisions. With the 5 characters shortuuid there is a reasonable chance for collisions. The number of pages we have is at least on the order of 10^4, that means we have about 10^5 translations. With 5 characters, excluding 0 and 1, there are less than 10^9 possible shortcodes. That means we have a probability for collisions on the order of 1/10^5 or even more.

2. We are using the module for these 2 lines:
   ```
   random_num = int(binascii.b2a_hex(os.urandom(length)), 16)
   return int_to_string(random_num, self._alphabet, padding=length)[:length]
   ```
   I see no need to include a full module as a dependency just to have these Python core functions mapped.

I suggest that we remove the dependency and either use the Python core function themselves to replicate the short UUID feature, or actually use database IDs. I prefer the latter, unless someone comes up with a good reason to not do this.

PR #535, opened 2020-10-16
Fix short URL feature

### LINKED ISSUE #532: Fix short URL copying to clipboard
### Describe the Bug
Currently, the short URL is copied to the clipboard on the server (`pyperclip` module), see function `copy_short_url()` in `src/cms/views/pages/page_actions.py`.

### Expected Behavior
Users in the Web GUI should be able to copy the short URL to the clipboard. We need a JavaScript function in the page view or pages list that copies the short URL into the clipboard in the browser. The `copy_short_url()` function can probably be dropped.

Example for copying to clipboard in JavaScript: https://www.w3schools.com/howto/howto_js_copy_clipboard.asp

svenseeberg:
- Copying the short URL to the clipboard was done on the back end, not in the browser.
- Remove short UUID dependency

Fixes: #532 
Fixes: #533


src/cms/static/js/pages/page_bulk_actions.js
```diff
@@ -31,3 +31,14 @@ function bulk_action_execute() {
...
+function short_url_to_clipboard(url) {
+    u('#mirrored_page_first_div').removeClass('hidden');
+    var copyText = document.getElementById("copy_to_clipboard");
+    copyText.value = "https://" + location.hostname + "/" + url;
+    u('#copy_to_clipboard').removeClass('hidden');
+    copyText.select();
+    copyText.setSelectionRange(0, 99999);
+    document.execCommand("copy");
+    u('#copy_to_clipboard').addClass('hidden');
+}
```
svenseeberg, line 44: This code currently exists twice. Have not yet found a place where to put it in order to merge it.

src/cms/templates/pages/page_tree_node.html
```diff
@@ -103,7 +103,7 @@
             {% endif %}
         {% endif %}
         {% if page_translation %}
-        <a href="{% url 'copy_short_url' page_id=page.id region_slug=region.slug language_code=language.code %}" title="{% trans 'Copy short link' %}" class="py-3 px-2">
+        <a href="#" onclick="short_url_to_clipboard('{{ page_translation.short_url }}');" title="{% trans 'Copy short link' %}" class="py-3 px-2">
```
svenseeberg, line 106: I think this needs to go away from the page list, because the user cannot directly choose the translation for the URL.

src/cms/static/js/pages/page_bulk_actions.js
timoludwig: I started to separate the js files by tasks in `src/cms/static/js/pages/`, so you could just create a new file `src/cms/static/js/pages/copy_to_clipboard.js` and include this in the page tree template...

src/cms/templates/pages/page_tree_node.html
timoludwig: Once we fix the language switcher, it will no longer be a problem to change e.g. to the english page tree and export the short url from there...

src/cms/views/pages/page_actions.py
```diff
@@ -112,46 +111,23 @@ def delete_page(request, page_id, region_slug, language_code):
...
-            _("URL '{}' was successfully copied to clipboard.".format(short_url)),
-        )
-    # pylint: disable=broad-except
-    except Exception as e:
-        logger.exception(e)
-        messages.warning(
-            request,
-            _("URL '{}' could not be copied to clipboard.".format(short_url)),
-        )
+    queryset = PageTranslation.objects.filter(short_url_id=short_url_id)
```
svenseeberg, line 118: Needs a check if a page is published.

src/cms/views/pages/page_actions.py
```diff
@@ -112,46 +111,23 @@ def delete_page(request, page_id, region_slug, language_code):
...
-@login_required
-def expand_short_url(request, short_url_id):
+def expand_page_translation_id(request, short_url_id):
     """
-    Searches for a page with requested short_url_id and redirects to that page.
+    Searches for a page translation with corresponding ID and redirects browser to web app
     """
-    queryset = PageTranslation.objects.filter(short_url_id=short_url_id)
-    page_translation = queryset.first()
+    page_translation = PageTranslation.objects.get(id=short_url_id)
```
svenseeberg, line 130: Needs a check if a page is published.

src/cms/static/js/pages/page_bulk_actions.js
svenseeberg: I modified the function to be used in other cases as well and moved it into a dedicted JS file.

src/cms/templates/pages/page_tree_node.html
svenseeberg: Alright.

src/cms/views/pages/page_actions.py
timoludwig: Also, there might be a newer revision of the page translation...
And we have to think about what to do on a 404... would we show the 404 of the cms, or redirect to the 404 in the webapp?

src/cms/views/pages/page_actions.py
```diff
@@ -112,48 +111,25 @@ def delete_page(request, page_id, region_slug, language_code):
...
+    page_translation = queryset.first()
 
-    return redirect(
-        "pages",
-        **{
-            "region_slug": region_slug,
-            "language_code": language_code,
-        },
-    )
+    if page_translation and page_translation.latest_public_revision:
```
svenseeberg, line 121: @timoludwig do we need to additionally check for archived pages here? I guess yes?

timoludwig: Yes, exactly.

svenseeberg: Alright then :)

src/cms/views/pages/page_actions.py
svenseeberg: Okay, we actually should redirect to the newest version (slug), for sure. About the 404: good question.

src/cms/views/pages/page_actions.py
```diff
@@ -112,48 +111,33 @@ def delete_page(request, page_id, region_slug, language_code):
...
-        "pages",
-        **{
-            "region_slug": region_slug,
-            "language_code": language_code,
-        },
-    )
+    if (
+        page_translation
+        and page_translation.latest_public_revision
+        and page_translation.page.archived == False
```
timoludwig, line 124: I think pylint wants to see this:
```suggestion
        and not page_translation.page.archived
```

src/cms/views/pages/page_actions.py
```diff
@@ -112,48 +111,33 @@ def delete_page(request, page_id, region_slug, language_code):
...
     """
-    queryset = PageTranslation.objects.filter(short_url_id=short_url_id)
-    page_translation = queryset.first()
+    page_translation = PageTranslation.objects.get(id=short_url_id)
 
-    if page_translation:
+    if (
+        page_translation
+        and page_translation.latest_public_revision
+        and page_translation.page.archived == False
```
timoludwig, line 139:
```suggestion
        and not page_translation.page.archived
```

setup.py
```diff
@@ -42,6 +42,7 @@
         "rules",
         "six",
         "webauthn",
+        "shortuuid",
```
timoludwig, line 45:
```suggestion
```
We can drop this now, right?

src/cms/views/pages/page_actions.py
timoludwig: I think the webapp has no generic 404 page, if we redirect e.g. to https://integreat.app/@@redux-first-router/NOT_FOUND, it shows `NetworkError when attempting to fetch resource.` which is a bit misleading. And when we redirect to https://integreat.app/404, it says `Sorry, but this city could not be found.`, which is also not entirely correct. So I propose to keep it as it is and show the error message of the cms...

setup.py
svenseeberg: True. I missed this line.

src/cms/views/pages/page_actions.py
svenseeberg: I'm now redirecting to the latest public revision. Should do the trick, right?

svenseeberg: In this case the check is not required anymore :roll_eyes:

timoludwig (approved):
:rocket: :tada:

PR #537, opened 2020-10-18
Add page translation ID based short URLs, fixes #533

### LINKED ISSUE #533: Remove shortuuid module

svenseeberg:
* Support short URLs based on page translation IDs.
* Remove `shortuuid` module

### DIFF
--- a/src/cms/urls.py
+++ b/src/cms/urls.py
@@ -42,6 +42,11 @@ urlpatterns = [
                     r"^(?P<short_url_id>[-\w]+)/",
                     pages.expand_short_url,
                     name="expand_short_url",
+                ),
+                url(
+                    r"^p/(?P<short_url_id>[0-9]+)$",
+                    pages.expand_page_translation_id,
+                    name="expand_page_translation_id",
                 )
             ]
         ),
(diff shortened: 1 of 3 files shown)


svenseeberg:
Continued in #535

PR #538, opened 2020-10-19
Page API: visibility of children, fixes #386

### LINKED ISSUE #386: Pages endpoint: skip pages if parent is not visible
If a parent page in the page tree is in any state that makes it invisible in the API, we do not want the children to appear in the pages endpoint. This should allow for easy deactivating of "full chapters".

svenseeberg:
* Only show page if parent is visible.

The solution is pretty straight forward: start with the main level of pages (the ones without parents). If they have visible content, start a recursive function that fetches the children.

Fixes #386

### DIFF
--- a/src/api/v3/pages.py
+++ b/src/api/v3/pages.py
@@ -1,6 +1,6 @@
 from django.http import JsonResponse
 
-from cms.models import Region, Page
+from cms.models import Region
 
 
 def transform_page(page_translation):
@@ -36,20 +36,20 @@ def transform_page(page_translation):
 def pages(request, region_slug, language_code):
     region = Region.get_current_region(request)
     result = []
-    for page in Page.objects.filter(region=region, parent=None):  # get main level
+    for page in region.pages.filter(archived=False, parent=None):  # get main level
         page_translation = page.get_public_translation(language_code)
         if page_translation:
             result.append(transform_page(page_translation))
-            result = get_children(region, page, language_code, result)
+            result = get_children(page, language_code, result)
     return JsonResponse(
         result, safe=False
     )  # Turn off Safe-Mode to allow serializing arrays
 
 
-def get_children(region, parent, language_code, result):
-    for page in Page.objects.filter(region=region, parent=parent):
+def get_children(parent, language_code, result):
+    for page in parent.children.filter(archived=False):
         page_translation = page.get_public_translation(language_code)
         if page_translation:
             result.append(transform_page(page_translation))
-            result = get_children(region, page, language_code, result)
+            result = get_children(page, language_code, result)
     return result
--- a/src/api/v3/pages.py
+++ b/src/api/v3/pages.py
@@ -1,6 +1,6 @@
 from django.http import JsonResponse
 
-from cms.models import Region
+from cms.models import Region, Page
 
 
 def transform_page(page_translation):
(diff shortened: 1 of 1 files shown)


src/api/v3/pages.py
```diff
@@ -36,10 +36,20 @@ def transform_page(page_translation):
 def pages(request, region_slug, language_code):
     region = Region.get_current_region(request)
     result = []
-    for page in region.pages.all():
+    for page in Page.objects.filter(region=region, parent=None):  # get main level
```
timoludwig, line 39:
```suggestion
    for page in region.pages.filter(archived=False, parent=None):  # get main level
```
archived pages should be excluded from the result.

src/api/v3/pages.py
```diff
@@ -36,10 +36,20 @@ def transform_page(page_translation):
...
         if page_translation:
             result.append(transform_page(page_translation))
+            result = get_children(region, page, language_code, result)
     return JsonResponse(
         result, safe=False
     )  # Turn off Safe-Mode to allow serializing arrays
+
+
+def get_children(region, parent, language_code, result):
+    for page in Page.objects.filter(region=region, parent=parent):
```
timoludwig, line 50:
```suggestion
    for page in parent.children.filter(archived=False):
```
- `children` is the reverse related name of `parent` and thus contains all direct children
- filter for region is not required because the children always have the same region as the parent
- exclude archived children

src/api/v3/pages.py
```diff
@@ -36,10 +36,20 @@ def transform_page(page_translation):
...
         page_translation = page.get_public_translation(language_code)
         if page_translation:
             result.append(transform_page(page_translation))
+            result = get_children(region, page, language_code, result)
     return JsonResponse(
         result, safe=False
     )  # Turn off Safe-Mode to allow serializing arrays
+
+
+def get_children(region, parent, language_code, result):
```
timoludwig, line 49:
```suggestion
def get_children(parent, language_code, result):
```
remove unused argument

src/api/v3/pages.py
```diff
@@ -36,10 +36,20 @@ def transform_page(page_translation):
...
+    for page in Page.objects.filter(region=region, parent=parent):
+        page_translation = page.get_public_translation(language_code)
+        if page_translation:
+            result.append(transform_page(page_translation))
+            result = get_children(region, page, language_code, result)
```
timoludwig, line 54:
```suggestion
            result = get_children(page, language_code, result)
```
remove unused argument

src/api/v3/pages.py
```diff
@@ -36,10 +36,20 @@ def transform_page(page_translation):
 def pages(request, region_slug, language_code):
     region = Region.get_current_region(request)
     result = []
-    for page in region.pages.all():
+    for page in Page.objects.filter(region=region, parent=None):  # get main level
         page_translation = page.get_public_translation(language_code)
         if page_translation:
             result.append(transform_page(page_translation))
+            result = get_children(region, page, language_code, result)
```
timoludwig, line 43:
```suggestion
            result = get_children(page, language_code, result)
```
remove unused argument

src/api/v3/pages.py
```diff
@@ -1,6 +1,6 @@
 from django.http import JsonResponse
 
-from cms.models import Region
+from cms.models import Region, Page
```
timoludwig, line 3:
```suggestion
from cms.models import Region
```
remove unused import

timoludwig (approved):
Looks good to me! :+1:

PR #562, opened 2020-10-28
Change FCM channel name

svenseeberg:
The native app released with region slugs in the channel name instead of the ID. This change is already reflected in WP and the documentation.


timoludwig (approved):
Looks good to me! :+1:

PR #564, opened 2020-10-30
Rename repository

timoludwig:
### Short description

Just in case we decide on changing our repository name, this replaces all hardcoded occurrences of `cms-django` by `integreat-cms`. This mostly involves the URLs to the repository on GitHub.


ulliholtgrave (commented):
I guess we should rename the github repo, before we apply this change. Especially all github.io/ urls aren't supported right now.
Or maybe we are doing the renaming in two steps?
First: Rename/Migrate all documentation
Second: Apply the change in this PR.

timoludwig:
Yes, I will merge this PR right after the repository is renamed and the documentation is deployed to the new URL. I just wanted to wait until this PR is approved so that it can be merged right after the renaming.

melegiul (changes requested):
Screenshot from 2020-11-18 20-53-14
I found some more appearances of 'cms-django' in the documentation, was this you're intention? If not, you could use a script like this to remove them:
 ```
for file in *.rst
do
sed -i 's/cms-django/integreat-cms/g' $file
done
```

timoludwig:
> I found some more appearances of 'cms-django' in the documentation, was this you're intention?

Oh you're right, somehow my search+replace didn't work correctly. However, I found a way to completely get rid of the hardcoded urls :tada:

melegiul (approved):

melegiul:
Nice :tada:. Looks good to me.

Issue #494, opened 2020-09-25
Drag & drop in page tree not working in Chromium


timoludwig:
### Describe the Bug

In the page tree, only the root page can be dragged and all sub-pages can't.

### Steps to Reproduce

1. Go to a Region
2. Click on Pages
3. Try to drag & drop a page other than a page on the root level

### Expected Behavior

The page can be dragged and dropped on another page or on a gap between pages.

### Actual Behavior

The page cannot be dragged.

### Additional Information

Reproduced in Chromium Version 85.0.4183.121 (Officicial Build) Arch Linux (64-Bit)

PR #565, opened 2020-10-30
fixed drag & drop bug in page tree

melegiul:
### Short description

fixed drag & drop bug in page tree

### Proposed changes

- the bug was caused by dom manipulation during the dragstart event
- dom manipulation must be delayed in this case by timeout
- this problem only occurs when using chrome browser

### Resolved issues

Fixes: #494


timoludwig (approved):
Cool that you found the cause of the bug, I'm impressed! :star_struck:

Issue #554, opened 2020-10-25
Public PDF export API


svenseeberg:
In addition to the "internal" PDF export functionality, we need to have a public version that implements https://github.com/Integreat/cms/wiki/REST-API-PDF

This is used by the PDF export button of the web app.

timoludwig:
The option with `pdf?id={id}` is currently not used, so only `pdf` and `pdf?url={permalink}` is required.

melegiul:
Is the api response expected to be in json format (i.e. Django JsonResponse)? If so can you tell the expected format, like the key values? I think in this case you want a key-value pair containing the link to the PDF document.

timoludwig:
No, our API should directly return the pdf document, like e.g. here:
https://cms.integreat-app.de/augsburg/de/wp-json/ig-mpdf/v1/pdf

Issue #119, opened 2019-03-29
Map old permalinks


svenseeberg:
Pages should be mapped by perma links, the perma link is based on the title. If the title changes, a new perma link should be generated. In this case, the old perma link should continue to work. Therefore a mapping must exist.

timoludwig:
With the versioning system of page translations, this is no longer an issue. As we create a new version whenever we change the slug of a page translation, we can just perform a simple database access and query for the translation with the given slug and return the most recent version of it.

timoludwig:
I just noticed that I confused two things here: Yes, the slugs of the translations are stored automatically and we don't need any additional mapping for it, but in case we change the structure of the pages (e.g. move a page to another parent page), the permalink will change and we currently don't track that.
So I think that this _is_ a non-trivial issue to solve and revise my comment above.

svenseeberg:
So what we need is two models: `PageTranslationPermalinkMap` and `EventTranslationPermalinkMap`, with each having an string attribute `permalink` and a foreign key to existing page or event translations. Right?

timoludwig:
If we really need a permalink storage, yes. But I thought the current status of our Slack discussion was that we don't.
Since we store old slugs in the translation model (just in old revisions), the only reason we would need the storage is to track the hierarchy-changes of pages. But we can as well ignore them and just take the last URL parameter of the path as slug. The worst thing that can happen is that we map a permalink which never existed to an existing page (e.g. /augsburg/de-de/non-existent/welcome). But I see no downside for the real use case here and would prefer less models and a slimmer database structure.

svenseeberg:
@timoludwig does mapping old slugs for page translations already work? If yes, we can close this issue.

timoludwig:
Yes, you can always do:
```
old_revision = PageTranslation.objects.filter(slug="old-page-translation-slug").first()
new_revision = region.pages.get(id=old_revision.page.id).get_translation(old_revision.language.code)
```
or:
```
page = region.pages.filter(translations__slug="old-page-translation-slug").distinct().first()
```

PR #572, opened 2020-11-05
Add PDF API

melegiul:
### Short description

Added PDF api for single/all pages.

### Proposed changes

- When the user sends a request to the API with empty query string, all
pages of the corresponding region are contained in PDF document.
- If the user specifies explicitly a single page by setting the parameter url to the permalink (e.g. url=augsburg/de-de/willkommen/willkomen-in-augsburg) of a page translation, the page and all descendants are returned.

### Resolved issues

Fixes: #554


src/api/v3/pdf_export.py
```diff
@@ -0,0 +1,70 @@
...
+    # since instances of GET are immutable,
+    # they must be copied before modified
+    request.GET = request.GET.copy()
+    if request.GET.get("url"):
+        # remove leading and trailing slashed to avoid ambiguos urls
+        url = request.GET.get("url").strip("/")
+        # get page candidate by filtering for translation slug
+        page = get_object_or_404(
+            Page, region=region, translations__slug=url.split("/")[-1]
+        )
```
timoludwig, line 46:
```suggestion
        # the last path component of the url is the page translation slug
        page_translation_slug = url.split("/")[-1]
        # get page by filtering for translation slug and translation language code
        page = get_object_or_404(
            region.pages,
            translations__slug=page_translation_slug,
            translations__language__code=language_code,
        )
```
I know you just copied+pasted the code from the single page endpoint, but it currently has a flaw: Translation slugs are only unique by region + language, so if two translations from two different languages have the same slug (this usually happens when cities duplicate e.g. the German version to other languages without really translating them), this raises the exception `cms.models.pages.page.Page.MultipleObjectsReturned: get() returned more than one Page`.

src/api/v3/pdf_export.py
```diff
@@ -0,0 +1,70 @@
...
+            request.GET.update({"api": pages})
+        else:
+            logger.error(
+                "The requested url: %s does not match the permalink for this page translation: %s",
+                url,
+                page_translation.permalink,
+            )
+            return HttpResponse(
+                "There exists no appropriate page for the requested url."
+            )
```
timoludwig, line 62:
```suggestion
```
This part is also outdated - in the meantime, we decided not to keep track of old permalinks (see #119), so we will ignore the given permalink/url and just process the last url component (=slug).

src/api/v3/pdf_export.py
```diff
@@ -0,0 +1,70 @@
...
+    else:
+        request.GET.update({"api": region.pages.filter(archived=False)})
+    response = export_pdf(request, region_slug, language_code)
```
timoludwig, line 65: I'm not sure about this "view-hacking" here. I think it would be cleaner to add a separate utility function (e.g. in `src/cms/utils/pdf_utils.py`) which does the pdf-processing and call this function in both pdf views.

src/cms/views/pages/page_actions.py
```diff
@@ -163,9 +163,9 @@ def link_callback(uri, rel):
 # pylint: disable=unused-argument, too-many-locals
 def export_pdf(request, region_slug, language_code):
     """
-    Function for handling a pdf export request for a single page.
-    If the page is a root page, the corresponding django template
-    inserts the content of all children pages.
+    Function for handling a pdf export request for pages.
+    The pages were either selected by cms user (see :func:`cms.static.js.pages.page_bulk_action.bulk_action_execute`)
```
timoludwig, line 167: You can only reference Python classes/functions/objects with Sphinx, no JavaScript functions.

timoludwig (commented):
Thanks! In general, your solution works. :+1: 

Unfortunately, there are some flaws in the single page endpoint where you copied code from. Sorry about that, I wasn't aware of this.

src/api/v3/pdf_export.py
melegiul: Sorry, i don't get this. The reason why i'm passing here that extra argument "api" with the HTTP GET attribute is, that i don't want to repeat these database queries unnecessaryly. So in this function i have to query the db anyway for compareing the permalink, another solution would be to declare an optional argument for the export_pdf function and pass that api variable that way, but i don't see there any advantage. Can you explain to me, what you mean?

timoludwig: The function `export_pdf(request, region_slug, language_code)` is the pdf view for logged in backend users, not api users.
So this function should only be called by Django in the internal request processing, not manually.
To prevent code duplications, you can put the contents of this function in a separate utility function and call this new utility function from both views.

melegiul: I see, kind of "seperation of concerns" technique. Ok, I can fix that:ok_hand:

src/cms/views/pages/page_actions.py
melegiul: Yes, until now this won't work. If you like, I could add sphinx-js to our pipfile, this should enable auto-generating docs from the js files too? See [https://hacks.mozilla.org/2017/07/introducing-sphinx-js-a-better-way-to-document-large-javascript-projects/](url)

timoludwig: Very cool, I didn't know of this extension. I will open a new issue for this feature.

src/api/v3/pdf_export.py
melegiul: Alright, utility function for PDF generation is now ready.

src/cms/utils/pdf_utils.py
```diff
@@ -0,0 +1,142 @@
...
+
+from backend.settings import STATIC_URL, MEDIA_URL
+from ..constants import text_directions
+from ..models import Language
+
+logger = logging.getLogger(__name__)
+
+
+# pylint: disable=too-many-locals
+def generate_pdf(region_slug, language_code, region, pages):
```
timoludwig, line 21: You don't need both the `region_slug` and the `region` parameter, one is enough.

src/cms/utils/pdf_utils.py
```diff
@@ -0,0 +1,142 @@
...
+        "html2pdf": True,
+        "right_to_left": language.text_direction == text_directions.RIGHT_TO_LEFT,
+        "region": region,
+        "pages": pages,
+        "language": language,
+        "amount_pages": amount_pages,
+    }
+    response = HttpResponse(content_type="application/pdf")
+    if amount_pages == 0:
+        return HttpResponse(_("No valid pages selected for PDF generation."))
```
timoludwig, line 72:
```suggestion
        return HttpResponse(_("No valid pages selected for PDF generation."), status=400)
```
You could also use `400 Bad Request` here (which makes it easier for the client to detect the error.

Also, you could move this condition right after `amount_pages = pages.count()` to make sure no unnecessary code is executed if no pages are given.

src/cms/utils/pdf_utils.py
```diff
@@ -0,0 +1,142 @@
...
+        html, dest=response, link_callback=link_callback, encoding="UTF-8"
+    )
+    if pisa_status.err:
+        logger.error(
+            "The following PDF could not be rendered: Region: %s, Language: %s, Pages: %s.",
+            region,
+            language,
+            pages,
+        )
+        return HttpResponse(_("The PDF could not be successfully generated."))
```
timoludwig, line 100:
```suggestion
        return HttpResponse(_("The PDF could not be successfully generated."), status=500)
```
Use `500 Internal Server Error` to help the client understand the problem.

timoludwig (commented):
Thanks! Yes, this is what I meant in my review :+1: 
Just found some minor improvements.

src/cms/utils/pdf_utils.py
melegiul: That's true, i just changed it.

timoludwig (approved):
Amazing, thanks! :+1: 
I think this is ready to merge now :tada:

Issue #576, opened 2020-11-09
Old permalinks not mapped in single page API endpoint


timoludwig:
### Describe the Bug

The permalink is checked for correctness in the single page API endpoint, but in #119 we decided not to keep track of old permalinks. Instead, we will just map permalinks based on the last url parameter (the slug) and ignore the rest.

### Steps to Reproduce

1. Copy one page permalink
2. Move that page to another parent to change the permalink
3. Request this page in the single page API endpoint with the old permalink from step 1

### Expected Behavior

The page should be found even if the permalink is old

### Actual Behavior

The error `No Page matches the given url or id` is returned.

Issue #575, opened 2020-11-09
MultipleObjectsReturned in single page API endpoint


timoludwig:
### Describe the Bug

The page translation slug is unique only per region and language. So if page translations are only queried by region and slug, there might be multiple objects with that conditions.

### Steps to Reproduce

1. Modify test data so that two translations of the same region but with different languages have the same slug
2. Retrieve the single page endpoint with this slug

### Expected Behavior

The page translation of the requested languages should be returned

### Actual Behavior

The following error appears:
```
cms.models.pages.page.Page.MultipleObjectsReturned: get() returned more than one Page -- it returned 2!
```

### Proposed solution

[Line 44](https://github.com/Integreat/cms-django/blob/acfe4cc079ac0326f92c22a0089019b3e6225fb7/src/api/v3/single_page.py#L44)
```
        page = get_object_or_404(
            Page, region=region, translations__slug=url.split("/")[-1]
        )
```
should become
```
        # the last path component of the url is the page translation slug
        page_translation_slug = url.split("/")[-1]
        # get page by filtering for translation slug and translation language code
        page = get_object_or_404(
            region.pages,
            translations__slug=page_translation_slug,
            translations__language__code=language_code,
        )
```

PR #577, opened 2020-11-09
Fix single page API endpoint

timoludwig:
### Short description

Fix two bugs in the single page API endpoint

### Proposed changes

- Fix Queryset in single page API endpoint
- Remove permalink check in single page API endpoint

### Resolved issues

Fixes: #575
Fixes: #576


melegiul (approved):
I just reproduced the bug, the solution looks good to me:thumbsup:

PR #544, opened 2020-10-21
Add Feedback overview

mael-fosso:
### Short description
Add view for feedback

### Proposed changes
- Overview page of feedback ordered by submission time.

- Distinction between web and native app. It should also be shown how much up and downvotes there are in sum for this page.

It should be possible to search for for feedback about individual pages.

### Resolved issues

Fixes: #32

PR #580, opened 2020-11-10
Improve feedback list view

timoludwig:
### Short description

These are my suggestions to improve PR #544

### Proposed changes

- Split feedback view into `AdminFeedbackListView` for technical feedback and `RegionFeedbackListView` for non-technical feedback
- Add access control for the feedback view
- Improve feedback models
- Apply other suggestions from code review


src/cms/views/feedback/region_feedback_list_view.py
```diff
@@ -0,0 +1,71 @@
...
+        )
+        event_list_feedback = region.event_list_feedback.filter(is_technical=False)
+        offer_feedback = region.offer_feedback.filter(is_technical=False)
+        offer_list_feedback = region.offer_list_feedback.filter(
+            region=region, is_technical=False
+        )
+        search_result_feedback = region.search_result_feedback.filter(
+            is_technical=False
+        )
+        imprint_page_feedback = region.imprint_page_feedback.filter(is_technical=False)
```
emmakemb, line 42: Hi, 
Thanks for your Sugestion Changes...
it means that is_technical=False means that there is no Feedback??

timoludwig: No, `is_technical=False` means that the feedback is targeted at the content creators, not at the programmers.
If you go to https://integreat.app/testumgebung/de and click on the emoticons on the left side, you can leave feedback for a specific page. The feedback function also has a dropdown "Wofür ist das Feedback?" where you can also select either "Inhalte von Testumgebung" or "Technische Funktionen". When you select "Technische Funktionen", it will be saved as `is_technical=True`. So the idea behind this is that feedback to "Inhalte von Testumgebung" is relevant for the region managers who created the content and feedback to "Technische Funktionen" is relevant for us programmers (thus visible at the admin level, not at the region level).

PR #583, opened 2020-11-11
Less verbosity for pipenv in venv

timoludwig:
### Short description

I noticed that the pipenv verbosity was inconsistent between different dev tools.
I fixed it by making pipenv less verbose when running in a venv in all remaining dev tools as well.

### Proposed changes

- Set `PIPENV_VERBOSITY=-1` in venv


melegiul (approved):
Looks good to me:thumbsup:

Issue #546, opened 2020-10-22
Add icon to region model


timoludwig:
### Motivation

At the moment, the icon field in the region form is just a mock-up.

### Proposed Solution

Add a field
```
icon = models.ImageField(blank=True, null=True, upload_to="regions/%Y/%m/%d")
```
to the region model. Then, add this also to the region form & template and make sure its saved when the form gets submitted.

PR #585, opened 2020-11-17
Added region icon to form/model

melegiul:
### Short description

Added region icon to form/model

### Proposed changes

-The region icon may be set in the region form, after setting the image
it is displayed in the form.
-For region forms without user defined icon, there is a sample mockup
icon presented in region form.

### Resolved issues

Fixes: #546

### DIFF
--- a/src/cms/locale/de/LC_MESSAGES/django.po
+++ b/src/cms/locale/de/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-11-20 11:53+0000\n"
+"POT-Creation-Date: 2020-11-22 18:36+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: Integreat <info@integreat-app.de>\n"
 "Language-Team: Integreat <info@integreat-app.de>\n"
@@ -2184,9 +2184,9 @@ msgstr "Ja, diese Region jetzt löschen"
 
 #: templates/regions/region_icon_widget.html:18
 msgid "No icon selected"
-msgstr "Keine Icon wurde ausgewählt"
+msgstr "Kein Icon ausgewählt"
 
-#: templates/regions/region_icon_widget.html:20
+#: templates/regions/region_icon_widget.html:21
 msgid "Clear"
 msgstr "Löschen"
 
--- a/src/cms/locale/de/LC_MESSAGES/django.po
+++ b/src/cms/locale/de/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-11-17 17:10+0000\n"
+"POT-Creation-Date: 2020-11-20 11:53+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: Integreat <info@integreat-app.de>\n"
 "Language-Team: Integreat <info@integreat-app.de>\n"
@@ -33,7 +33,7 @@ msgid "Rural district"
 msgstr "Landkreis"
 
 #: constants/administrative_division.py:55
-#: templates/regions/region_form.html:204
+#: templates/regions/region_form.html:192
 msgid "Region"
 msgstr "Region"
 
@@ -355,7 +355,7 @@ msgstr "Gespiegelte Seite nach dieser Seite einbinden"
 msgid "Private"
 msgstr "Privat"
 
-#: forms/regions/region_form.py:21
+#: forms/regions/region_form.py:25
 msgid "Do no import initial content"
 msgstr "Keine Inhalte übernehmen"
 
@@ -420,7 +420,7 @@ msgstr "Mein Dashboard"
 msgid "Translation Report"
 msgstr "Übersetzungs-Bericht"
 
-#: templates/_base.html:122 templates/regions/region_form.html:134
+#: templates/_base.html:122 templates/regions/region_form.html:122
 #: templates/statistics/statistics_dashboard.html:8
 msgid "Statistics"
 msgstr "Statistiken"
@@ -438,7 +438,7 @@ msgid "Upload new media file"
 msgstr "Neue Mediendatei hochladen"
 
 #: templates/_base.html:142 templates/pages/page_tree.html:19
-#: templates/regions/region_form.html:208
+#: templates/regions/region_form.html:196
 msgid "Pages"
 msgstr "Seiten"
 
@@ -452,7 +452,7 @@ msgid "Create page"
 msgstr "Seite erstellen"
 
 #: templates/_base.html:158 templates/events/event_list.html:27
-#: templates/regions/region_form.html:209
+#: templates/regions/region_form.html:197
 msgid "Events"
 msgstr "Veranstaltungen"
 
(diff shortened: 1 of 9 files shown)


src/cms/static/css/style.less
```diff
@@ -344,13 +344,9 @@ h2.heading {
...
-    width: 0.1px;
-	height: 0.1px;
-	opacity: 0;
-	overflow: hidden;
-	position: absolute;
-    z-index: -1;
+.image-field {
+    display: none;
+    position: absolute;
 }
```
melegiul, line 350: I don't understand here the author's purpose. Setting such small width/height and opacity to 0 is just like hiding the element so I reduced this code. I kept on hiding it, because that Image Field Widget (see [https://github.com/django/django/blob/master/django/forms/templates/django/forms/widgets/clearable_file_input.html](url) can not be customized with css in accordance to the form styles.

melegiul: But hiding the elements means also, there is no more a checkbox (included in the widget) for deleting the image. If you like to, I could add a customized button for deleting the icon.

timoludwig: I don't get this width/height parameters either... So yes, `display:none` should be the way to go. Just a hint: you can directly assign the [tailwind class](https://tailwindcss.com/docs/display#class-reference) `hidden` instead of `image-field`.

Yes, the checkbox to remove the icon would be cool. You don't have to leave the design as it is, If you wanted to style the widget yourself, you can always define a custom widget which inherits from the default one in the form class:
```
class IconWidget(ClearableFileInput):
    template_name = "/path/to/custom/widget/template"

class RegionForm(forms.ModelForm):
    ...
    class Meta:
        ...
            "icon",
        ],
        widgets = {
            "icon": IconWidget,
        }
```

timoludwig: Ah, and one general note regarding your GitHub comments: I noticed that your links always point to https://github.com/Integreat/cms-django/pull/url.
So probably you're using the markdown feature incorrecly. it's `[link-text](url)` but you tend to fill your urls in the first part where the link text should go and leave the `url` part unchanged...
Apart from that: If you just want to paste the url without a different link text, you don't need special markdown syntax for this. just insert the url and GitHub will automatically convert it to a link.

melegiul: Thank you, I just fixed the link. Now it should work.

melegiul: You're proposed IconWidget sounds good for me. I will check that.

timoludwig: Yes, works now. And apart from the remove-checkbox: I think it would also be cool to see the filename anywhere after a file has been selected. This way you can be sure the selection worked before submitting the form.

src/cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -2101,87 +2101,95 @@ msgstr "Geographische Breite hier eingeben"
...
 msgstr "Orte"
 
-#: templates/regions/region_form.html:213
+#: templates/regions/region_form.html:204
 msgid "Yes, delete this region now"
 msgstr "Ja, diese Region jetzt löschen"
 
+#: templates/regions/region_icon_widget.html:18
+msgid "No icon selected"
+msgstr "Keine Icon wurde ausgewählt"
```
timoludwig, line 2187:
```suggestion
msgstr "Kein Icon ausgewählt"
```

src/cms/static/js/region-form-icon.js
```diff
@@ -0,0 +1,20 @@
...
+ * listen for user input (selection in file browser)
+ * for uploading region icons
+ * updates the icon preview and checkbox for clearing
+ */
+function updateIconDisplay() {
+    const curFiles = u('#icon').first().files;
+    if(curFiles.length === 0) {
+        u('#icon_status').text("No icon selected");
+        u('#clear_icon').addClass("hidden");
+        u('#clear_icon').removeClass("block");
```
timoludwig, line 13:
```suggestion
```
If the div is hidden, you don't need to remove the block class as well. It's not shown anyway.

src/cms/static/js/region-form-icon.js
```diff
@@ -0,0 +1,20 @@
...
+    } else {
+        u('#icon_status').text(`${curFiles[0].name}`);
+        u('#icon_preview').attr({src: URL.createObjectURL(curFiles[0])})
+        u('#clear_icon').addClass("block");
```
timoludwig, line 17:
```suggestion
```
I think a div with is not hidden is displayed as a block per default, right? So you don't have to add this class at all.

src/cms/templatetags/form_icon_filters.py
```diff
@@ -0,0 +1,20 @@
+import ntpath
```
timoludwig, line 1:
```suggestion
import os
```
I propose to use the standard os package instead of the windows-style ntpath.

src/cms/templatetags/form_icon_filters.py
```diff
@@ -0,0 +1,20 @@
...
+def strip_path(icon):
+    """
+    strips the internal path of the icon file to show user only the file name
+
+    :param icon: relative path inside the media folder
+    :type icon: str
+    :return: name of icon file
+    :rtype: str
+    """
+    return ntpath.basename(icon)
```
timoludwig, line 20:
```suggestion
    return os.path.basename(icon)
```

timoludwig (commented):
Really cool, thanks! In general, I like the implementation! :+1:

I have one general note: If you dynamically show the "clear" button after somebody chose an icon, it should at least be possible to use it. But if you click the clear-button right after changing the icon without saving in between, you submit the form with both the image field and the clear field, which causes the error "Please either submit a file or check the clear checkbox, not both."
So I would change it in one of the following two ways:
1. (recommended) Add a new event handler which empties the file-input-field and removes the previewed image when the clear-button is clicked
2. Do not show the clear-button dynamically after selecting a file, so that at first the users have to either submit the form or discard their file without saving before they can use the clear-button

melegiul:
> But if you click the clear-button right after changing the icon without saving in between, you submit the form with both the image field and the clear field, which causes the error "Please either submit a file or check the clear checkbox, not both."

I didn't note that yet, so option one sounds cool. I will fix that.

timoludwig (commented):
Thanks, this works as expected now! :tada: 

I'm very sorry, but I have one final request: Unfortunately, I didn't notice earlier, that the text "No icon selected" isn't translated when it's inserted by JavaScript dynamically.
So I'd propose to work with two separate divs for the "No icon selected"-message and the file path. And instead of changing the content of the `#icon_status`-div, just hide/show both divs so the translated message from the Django template gets used.

melegiul:
> Thanks, this works as expected now! tada
> 
> I'm very sorry, but I have one final request: Unfortunately, I didn't notice earlier, that the text "No icon selected" isn't translated when it's inserted by JavaScript dynamically.
> So I'd propose to work with two separate divs for the "No icon selected"-message and the file path. And instead of changing the content of the `#icon_status`-div, just hide/show both divs so the translated message from the Django template gets used.

Thanks for you're review. No problem, you're right. That js string won't get translated. I fix that.

src/cms/static/js/region-form-icon.js
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * make sure that on enabled icon clear checkbox the icon reference
+ * of the file input is removed
+ * @param {Event} event - emitted when the form is finally send to server
+ */
+function checkInput(event){
+    if(u('#icon-clear_id').is(':checked')) {
+        u('#icon').first().value = '';
+    }
+}
```
melegiul, line 61: Beside the problem of the missing translation, I had also to fix another problem by adding that listener. This is because of commands like this for removing and resetting file input are not possible because of security:
```
iconValue = u('input[type=file').first().value; 
u('input[type=file]').first().value = '';
... and later
u('input[type=file]').first().value = iconValue;
```
It's not possible to reset the file once it is removed, so I only remove it when the form is submitted and checkbox is checked.

timoludwig: Interesting. On my system, there was no security warning or similar, and everything worked. Maybe this is browser-dependent? But cool that you noticed.

timoludwig (approved):
Thanks, this works perfectly! :rocket:

Issue #578, opened 2020-11-10
API partially returns HTTP instead of JSON


timoludwig:
### Motivation

In some cases, the API returns HTTP responses instead of JSON.
This is typically the case when the shortcut function `get_object_or_404()` is used or explicitly `HttpResponse`/`Http404` is returned.

### Proposed Solution

Add new shortcuts `get_object_or_404_json()` and `Json404` which should be used in the API instead of the HTTP responses.

timoludwig:
I see a problem with my proposed solution:
`Http404` is an exception, not a valid response. Any only this exception is converted to a `HttpResponse`. So a custom exception `Json404` would not work.

Then I thought about middlewares, but middlewares can only modify request objects (the replacement of the exceptions are done internally and not exposed to the middlewares), and it's kind of hacky to convert a `HttpResponse` to a `JsonResponse` under certain conditions.

Thus, I think a custom decorator which catches the `Http404` might be the best solution here.

PR #586, opened 2020-11-17
Always return JSON in API

timoludwig:
### Short description

This adds a custom decorator which makes sure the API always returns a JSON response, even if the error `Http404` is raised.

### Proposed changes

- Make use of shortcut functions in API
- Add json_response decorator

### Resolved issues

Fixes: #578

### DIFF
--- a/src/api/decorators.py
+++ b/src/api/decorators.py
@@ -1,6 +1,8 @@
 import json
 
-from django.http import JsonResponse
+from functools import wraps
+
+from django.http import JsonResponse, Http404
 from django.views.decorators.csrf import csrf_exempt
 
 from cms.models import Region, Language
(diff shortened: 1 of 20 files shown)


melegiul (commented):
Works like expected, but I got one question. When I try the url: api/augsburg/de-de/imprint I got a RelatedObjectDoesNotExist exception. Should the decorator transform this messages also to JSON?

timoludwig:
> Works like expected, but I got one question. When I try the url: api/augsburg/de-de/imprint I got a RelatedObjectDoesNotExist exception. Should the decorator transform this messages also to JSON?

I guess we should rather use e.g. `get_object_or_404(region.imprint)` in the imprint api to make sure it throws a `Http404` instead of `RelatedObjectDoesNotExist`. This way, we are consistent to the other endpoints and don't need special error handling for other exceptions. I will add this to the first commit of this PR.

timoludwig:
Actually, I changed my mind about this: Probably, it makes sense to just return an empty response in case the imprint doesn't exist instead of an error. At least this is the current behavior of the WordPress APIv3.

melegiul (approved):
Ok, looks good to me. I found another problem with the url: api/augsburg/de-de/offers, here we got a TypeError Exception, but I don't thinks that's related to the decorator.

timoludwig:
> Ok, looks good to me. I found another problem with the url: api/augsburg/de-de/offers, here we got a TypeError Exception, but I don't thinks that's related to the decorator.

Thanks for reporting! It took me a moment to figure out what's going on there. It turns out, the OR-clauses in the urls.py `(offers|extras)` were captured as regex-match and thus the string "offers" was treated like another parameter which is passed to the view. I fixed this by using the syntax for non-capturing regex groups: `(?:offers|extras)`.

Issue #132, opened 2019-05-21
Pages: Considerations on how to deal with child pages when archiving the parent page


sascha11110:
**Open questions:**

1. How do we deal with it if only the child pages are restored? (Ordering of pages)
2. How do we handle the archived status of child pages? Do we set the appropriate flag for each child page or do we simply derive the status from the parent pages?
3. What do we do if a child page has been archived in advance and then the parent page is archived? Do we restore all pages or only pages that have not been archived in advance?

**The following tasks resulted from the considerations so far:**

- [ ] In the list of archived pages and in the tree view, the child pages must also be displayed accordingly if the parent page has been archived [_only if we do not set the archived flag for all child pages_]

svenseeberg:
The API pages endpoint only shows pages where the parent is also visible. See #538

Regarding the question: I would not change the status of children when changing the parent.

timoludwig:
> Regarding the question: I would not change the status of children when changing the parent.

I agree. So for this issue to be closed, we just have to show non-archived pages in the archived page tree if their parent is archived. Otherwise, non-archived children of archived pages just completely vanish.

PR #598, opened 2020-11-21
Improve handling of archived pages

timoludwig:
### Short description

In this PR, I implemented the necessary changes for the archiving workflow for pages.

### Proposed changes

- New archiving concept for hierarchical pages
    - Pages are explicitly archived when their attribute explicitly_archived is True (former name: archived)
    - Pages are implicitly archived when at least one of their ancestors is explicitly archived
    - Pages are archived when they're either explicitly or implicitly archived

- Improve handling of archived pages
    - Fix aggregation of archived/non-archived pages in page tree, api and sitemap
    - Fix archiving workflow
    - Fix error/success messages

### Resolved issues

Fixes: #132


src/cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -1698,88 +1708,114 @@ msgstr "Revision"
 msgid "Show revisions"
 msgstr "Revisionen anzeigen"
 
-#: templates/pages/page_form.html:175 templates/pages/page_revisions.html:66
+#: templates/pages/page_form.html:160
+#: templates/pages/page_tree_archived_node.html:25
+msgid "Implicitly archived by parent page"
+msgstr "Implizit archiviert durch übergeordnete Seite"
```
svenseeberg, line 1714: While this message is technically 100% correct, I'm not sure if users will understand the meaning. Do we need to make this more descriptive? Something like "A parent of this page is archived, therefore this page is also archived."

src/cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -1698,88 +1708,114 @@ msgstr "Revision"
...
-#: templates/pages/page_form.html:299
-#: templates/pages/page_tree_archived_node.html:92
-#: templates/pages/page_tree_node.html:96
+#: templates/pages/page_form.html:323
 msgid "You cannot delete a page which has children."
 msgstr "Sie können keine Seite löschen, die Unterseiten besitzt."
 
-#: templates/pages/page_form.html:303
+#: templates/pages/page_form.html:324
+msgid "To delete this page, you have to delete or move its subpage first:"
```
svenseeberg, line 1808: Just stumbled upon the change from "child" to "subpage". There are still some occurences of "child" in the .po file, which we should change accordingly (lines 1804, 2925, 167, 171)?

svenseeberg (approved):
Looks good to me, I only have some non-blocking questions about the wording.

src/cms/locale/de/LC_MESSAGES/django.po
timoludwig: I wasn't sure about this, too. I thought it might help region managers to understand what's going on if we use the wording explicitly/implicitly archived also in the UI. This message in particular should be the shortest possible representation of the status (where other pages got "Public", "In Review", or "Archived"). Do you have an idea for an equally short message?

Apart from that, there are also more descriptive messages in other parts of the template:

---------------------

Screenshot_2020-12-02 Integreat Backend

---------------------

Screenshot_2020-12-02 Integreat Backend(2)

src/cms/locale/de/LC_MESSAGES/django.po
timoludwig: Good point. I had a look at WordPress and noticed that the typical wording is "subpage" instead of "child page", so probably it's a good idea to keep this consistent.

src/cms/locale/de/LC_MESSAGES/django.po
svenseeberg: The shortest possible alternative I can come up with right now is "A parent is archived therefore this page is archived".

timoludwig: @svenseeberg Do you think that the new wording is easier to understand?

timoludwig: @svenseeberg ping

Issue #505, opened 2020-09-26
Enable deletion of language tree nodes


timoludwig:
### Motivation

At the moment, language tree nodes can only be deleted via the Django Admin backend

### Proposed Solution

Add the possibility to delete language tree nodes via the language tree or the language tree node form.

### Additional Context

Also keep in mind what to do if content in this language exists.
Since the content is only related to the language objects in the database (not the language tree nodes), there is no inbuilt deletion protection here.

melegiul:
What's about the descendants of a language tree node? Should they be deleted too or is deletion blocked as long as there exist children of a language tree node (like the page deletion)?

timoludwig:
The deletion should be blocked as long as there are descendants.

melegiul:
I tried to stick close close to the projects conventions for confirmation popups, but I encountered some redundant templates for the popup dialogs. For each template folder like pages, pois, events, etc. there is a subdirectory for a confirmation dialog with three different templates. These templates are quiet the same, they only differ in single words like 'Are you sure you want to delete/archive/restore?'. So since I need myself a confirmation dialog for deleting the language nodes, I suggest a generic central template called 'confirmation-dialog'. This template is empty by default and only gets populated at runtime with the appropriate text according to current task (delete/archive/restore). What you think about this?

timoludwig:
This sounds good. We just have to make sure that the new code is actually less complex than the current solution, otherwise I'd prefer a few duplications (but easier readability).

melegiul:
No worry, complexity stays the same, no great changes. Next days I will open a draft pull request, you can check it out there.

Issue #466, opened 2020-08-12
Page form changes not correctly detected


timoludwig:
### Describe the Bug

The page form is always marked as changed, even if not changes are performed.
(The success message is always "Translation was successfully published." when it should say "No changes detected.")

### Steps to Reproduce

1. Go to a Region, e.g. "Augsburg"
2. Click on Pages
3. Click on any page
4. Do not change anything
5. Press "publish"

### Expected Behavior

The message "No changes detected." should appear

### Actual Behavior

The message "Translation was successfully published." appears

### Additional Information

The form's changed data always contains 'mirrored_page_first', even if this field was not changed.

PR #605, opened 2020-12-07
Added option to delete langauge tree node

melegiul:
### Short description

Feature for deleting language tree nodes and all their page translations.

### Proposed changes

- Button for deleting a language tree node and all the page translation
  for this language.
- Also inserted a generic confirmation dialog template, which is empty
  by default. The appropriate message is loaded in the template at
  runtime by retrieving messages from data- attributes of the clicked
  button.

### Resolved issues

Fixes: #505


src/cms/views/language_tree/language_tree_actions.py
```diff
@@ -67,3 +68,43 @@ def move_language_tree_node(
         logger.exception(e)
 
     return redirect("language_tree", **{"region_slug": region_slug})
+
+
+def delete_language_tree_node(request, region_slug, language_tree_node_id):
```
timoludwig, line 73:
```suggestion
@login_required
@region_permission_required
@permission_required("cms.manage_language_tree", raise_exception=True)
def delete_language_tree_node(request, region_slug, language_tree_node_id):
```
Please check the required permissions for the view, otherwise anonymous users, which are not logged in, can delete arbitrary language tree nodes.

src/cms/views/language_tree/language_tree_actions.py
```diff
@@ -67,3 +68,43 @@ def move_language_tree_node(
...
+        language_node.delete()
+        messages.success(
+            request,
+            _(
+                'The language tree node "{}" and all corresponding page translations were successfully deleted.'
+            ).format(language_node.translated_name),
+        )
+    except (ObjectDoesNotExist) as e:
+        messages.error(request, e)
+        logger.exception(e)
```
timoludwig, line 109:
```suggestion
    # get current region
    region = Region.get_current_region(request)
    # get current selected language node
    language_node = get_object_or_404(region.language_tree_nodes, id=language_tree_node_id)
    # get all page translation assigned to the language node
    page_translations = language_node.language.page_translations
    # filter these translation that belong to the region
    current_region_translations = page_translations.filter(
        page__region=region
    )
    current_region_translations.delete()
    language_node.delete()
    messages.success(
        request,
        _(
            'The language tree node "{}" and all corresponding page translations were successfully deleted.'
        ).format(language_node.translated_name),
    )
```
I propose the following changes:
- Use `get_object_or_404()` instead of try-catch-block
- Only delete a language tree node if it belongs to the current region (otherwise, region users could delete other region's language tree ndoes)

timoludwig (commented):
Thanks! The deletion process works as expected.
Please also delete all poi translations and event translations of the deleted language. Otherwise, these unused objects remain in the database.

Apart from that, the confirmation popups are also used in other places, which are not working yet with the new system:
- POI form
- POI list
- POI archived list
- Event form
- Event list
- Event archived list

Please fix them so they work again, preferably with the new system :)

ulliholtgrave:
Thanks for the PR. It seems that you need to take a look at language_tree_actions.py, as the checks for pylint are failing :)
image

timoludwig:
> Thanks for the PR. It seems that you need to take a look at language_tree_actions.py, as the checks for pylint are failing :)
> image

That's my fault, because I didn't care about pylint in my review suggestions :sweat_smile:

melegiul:
> @ulliholtgrave That's my fault, because I didn't care about pylint in my review suggestions sweat_smile

And mine too, forgotten to set back to draft:smiley:

src/cms/views/events/event_list_view.py
```diff
@@ -5,7 +5,7 @@
 from django.contrib.auth.mixins import LoginRequiredMixin, PermissionRequiredMixin
 from django.shortcuts import render, redirect
 from django.utils.decorators import method_decorator
-from django.utils.translation import ugettext as _
+from django.utils.translation import ugettext as _, ugettext_lazy
```
timoludwig, line 8:
```suggestion
from django.utils.translation import ugettext as _
```
You only need `ugettext_lazy` in forms and models. In views, you can use `ugettext` without problems.

src/cms/views/events/event_list_view.py
```diff
@@ -33,6 +33,27 @@ class EventListView(LoginRequiredMixin, PermissionRequiredMixin, TemplateView):
...
+        "restore_dialog_text": ugettext_lazy(
+            "All translations of this event will also be restored."
+        ),
+        "delete_dialog_title": ugettext_lazy(
+            "Please confirm that you really want to delete this event"
+        ),
+        "delete_dialog_text": ugettext_lazy(
+            "All translations of this event will also be deleted."
+        ),
+    }
```
timoludwig, line 56: If you have context you want to inject into multiple views, you should do this with the [ContextMixin](https://docs.djangoproject.com/en/2.2/ref/class-based-views/mixins-simple/#contextmixin):

`event_mixin.py`
```
from django.views.generic.base import ContextMixin

class EventMixin(ContextMixin):
    extra_context = {
        ...
    }
```

`event_list_view.py` and `event_view.py`
```
from .event_mixin import EventMixin

class EventListView(LoginRequiredMixin, PermissionRequiredMixin, TemplateView, EventMixin):
# or
class EventView(PermissionRequiredMixin, TemplateView, EventMixin):

...

        context = self.get_context_data(**kwargs)
        return render(
            request,
            self.template_name,
            {
                **context,
                ...
            } 
```
This also applies to the other content types.

src/cms/templates/events/event_form.html
```diff
@@ -297,13 +297,21 @@ <h3 class="font-bold my-3">{% trans 'Where' %}</h3>
                         <div class="pt-2 pb-4">
                             {% if event_form.instance.archived %}
                                 <span class="block font-bold mb-4">{% trans 'Restore event' %}</span>
-                                <button class="confirmation-button w-full bg-blue-500 hover:bg-blue-600 text-white font-bold py-2 px-4 rounded" data-confirmation-title="{% if event_translation_form.instance %}{{ event_translation_form.instance.title }}{% endif %}" data-action="{% url 'restore_event' event_id=event_form.instance.id region_slug=region.slug language_code=language.code %}" data-confirmation-popup="#confirm-restore-event">
+                                <button class="confirmation-button w-full bg-blue-500 hover:bg-blue-600 text-white font-bold py-2 px-4 rounded"
+                                    data-confirmation-title="{% trans 'Please confirm that you really want to restore this event' %}"
```
timoludwig, line 301:
```suggestion
                                    data-confirmation-title="{{ restore_dialog_title }}"
```
You wanted to use the context syntax also for events, right?

src/cms/views/events/event_list_view.py
melegiul: Sorry, but this is wrong. The reason you need lazy translation is, that you just just want to store a reference to the string and translate that as needed (when the template is rendered: https://docs.djangoproject.com/en/2.2/topics/i18n/translation/#lazy-translations). You can see what I mean when you do the following:

- Use ugettext as you suggest and save these in context (I propose the page menu, there you got delete/archive buttons)
- Choose english language in the upper right corner
- Click for example the delete button for arbitrary page
- Then you see that german messages are rendered in the dialog instead of english as you would expect

timoludwig: Oh sorry! I didn't realize the fact that the context is defined as class attribute in comparison to other strings in the view's `get()`-method. You right, in this case the lazy method is required in the view.

src/cms/templates/events/event_form.html
```diff
@@ -297,13 +297,21 @@ <h3 class="font-bold my-3">{% trans 'Where' %}</h3>
...
+                                    data-confirmation-text="{% trans 'All translations of this event will also be restored.' %}"
+                                    data-confirmation-subject="{% if event_translation_form.instance %}{{ event_translation_form.instance.title }}{% endif %}"
```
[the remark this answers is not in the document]
melegiul: Yes, that's right. I didn't think about that, because that was the way, it was coded before and I just renamed the attribute to 'subject'. But since this attribute is only saved as string, the if statement does not make a difference. I will remove them from all content types.

src/cms/views/events/event_list_view.py
melegiul: When I was preparing the content types for mixins and while I was testing the context rendering, I experienced a problem on develop branch with the page form. The form does no long detects, when the form is saved with no changes. This was earlier displayed by a message. Do you know, whats going on there?

timoludwig: Yes, I already opened the issue #466 (Page form changes not correctly detected) some time ago. You can ignore this in your PR.

melegiul: Thanks for the info, i'm just surprised because when I was fixing the icon upload bug, this 'no changes' message occurred to me. Seems like it works occasionally.

timoludwig (approved):
Thanks :+1: 
Works as expected and looks good! :rocket:

Issue #39, opened 2018-10-25
Add Imprint form


svenseeberg:
Add a form to edit the imprint.
This involves the form-model, form views and form templates.
This should be similar to the page form, so you can probably simply copy + paste existing code.

Depends on #357 :heavy_check_mark:

ulliholtgrave:
This might be an issue that is related to the one you already solved @klinzo ,or?

svenseeberg:
Definitely related, but not yet solved. We just have the model yet. The form is still needed.

PR #608, opened 2020-12-09
Add Imprint form

AminAlAit:
### Short description

Creation of Imprint form. Adjusted the code from the page form, page template and page view to suit that of the imprint.

### Proposed changes

- Inside _cms.view.imprints.imprint_view.py_, `page_id` and `new_page` variables were replaced by `imprint_id` and `new_imprint` with no noticeable effect.  

### Resolved issues

- Imprint tab can be clicked from the menu

Fixes: #39

### DIFF
--- a/src/cms/forms/imprint/__init__.py
+++ b/src/cms/forms/imprint/__init__.py
@@ -1,0 +1,6 @@
+"""
+Forms for creating and modifying imprint and imprint translation objects
+"""
+
+from .imprint_form import ImprintForm
+from .imprint_translation_form import ImprintTranslationForm
--- a/src/cms/forms/imprint/imprint_form.py
+++ b/src/cms/forms/imprint/imprint_form.py
@@ -1,0 +1,57 @@
+import logging
+
+from django import forms
+
+from ...models import ImprintPage
+
+
+logger = logging.getLogger(__name__)
+
+
+class ImprintForm(forms.ModelForm):
+    """
+    Form for creating and modifying imprint objects
+    """
+
+    class Meta:
+        model = ImprintPage
+        fields = ["icon"]
+
+    def __init__(self, *args, **kwargs):
+
+        logger.info(
+            "New ImprintForm instantiated with args %s and kwargs %s", args, kwargs
+        )
+
+        # pop kwarg to make sure the super class does not get this param
+        self.region = kwargs.pop("region", None)
+        disabled = kwargs.pop("disabled", None)
+
+        # instantiate ModelForm
+        super().__init__(*args, **kwargs)
+
+        # If form is disabled because the user has no permissions to manage the imprint, disable all form fields
+        if disabled:
+            for _, field in self.fields.items():
+                field.disabled = True
+
+    # pylint: disable=signature-differs
+    def save(self, *args, **kwargs):
+
+        logger.debug(
+            "ImprintForm saved with args %s, kwargs %s, cleaned data %s and changed data %s",
+            args,
+            kwargs,
+            self.cleaned_data,
+            self.changed_data,
+        )
+
+        # don't commit saving of ModelForm, because required fields are still missing
+        kwargs["commit"] = False
+        imprint = super().save(*args, **kwargs)
+
+        if not self.instance.id:
+            # only update these values when imprint is created
+            imprint.region = self.region
+        imprint.save()
+        return imprint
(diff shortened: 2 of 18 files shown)


src/cms/forms/imprints/imprint_translation_form.py
```diff
@@ -0,0 +1,100 @@
...
+            self.cleaned_data,
+            self.changed_data,
+        )
+
+        # pop kwarg to make sure the super class does not get this param
+        page = kwargs.pop("page", None)
+        user = kwargs.pop("user", None)
+
+        kwargs["commit"] = False  # Don't save yet. We just want the object.
+        page_translation = super().save(*args, **kwargs)
```
svenseeberg, line 72: I think we should rename `page_translation` to `imprint_translation`. Makes sense, doesn't it?

src/cms/templates/imprints/imprint_form.html
```diff
@@ -0,0 +1,475 @@
...
+            {% if can_edit_page %}
+                <input type="submit" name="submit_draft" class="bg-gray-500 hover:bg-gray-600 cursor-pointer text-white font-bold py-3 px-4 rounded mr-2" value="{% trans 'Save as draft' %}" />
+            {% endif %}
+            {% has_perm 'cms.publish_page' request.user page as can_publish_page %}
+            {% if can_publish_page %}
+                <input type="submit" name="submit_public" class="cursor-pointer bg-blue-500 hover:bg-blue-600 text-white font-bold py-3 px-4 rounded mr-2" value="{% trans 'Publish' %}" />
+            {% else %}
+                <input type="submit" name="submit_review" class="cursor-pointer bg-blue-500 hover:bg-green-600 text-white font-bold py-3 px-4 rounded mr-2" value="{% trans 'Submit for review' %}" />
+            {% endif %}
+            <button type="button" onclick="short_url_to_clipboard('s/p/{{ page_translation_form.instance.id }}');" class="cursor-pointer bg-blue-500 hover:bg-blue-600 text-white font-bold py-3 px-4 rounded">
```
svenseeberg, line 78: Not sure if we need a short URL at all for the imprint. But if we have it, we need to change the path. I'd say `i` for imprint? -> `s/i/IMPRINT_TRANSLATION_ID`.

svenseeberg:
The overall concept seems valid. Thank you for the contribution. Currently, lots of code is doubled. It would be cool to have some abstract classes and maybe even templates, so that the code can be shared with the page translations. @timoludwig do you agree?

timoludwig (changes requested):
Thanks for your work. :+1: 

Please have a look at the [Page](https://integreat.github.io/integreat-cms/autogenerated/cms.models.pages.html#module-cms.models.pages.page) and [ImprintPage](https://integreat.github.io/integreat-cms/autogenerated/cms.models.pages.html#module-cms.models.pages.imprint_page) models and understand, that both have the same base class [AbstractBasePage](https://integreat.github.io/integreat-cms/autogenerated/cms.models.pages.html#module-cms.models.pages.abstract_base_page), but are different models with slightly different attributes etc.
There are multiple places where you just copied+pasted code for the wrong model. Copy+paste is a good start here, because the models are very similar and it doesn't make sense to re-invent the wheel, but since they're not the same, you have to understand the difference between these models and adapt the files you copied.
This includes e.g.:
- Please make sure to use the correct models (`ImprintPage` instead of `Page` etc)
- Use the correct permissions everywhere (`manage_imprint` instead of `edit_page` etc)
- Please remove any fields which are only relevant for the Page model
- Rename all variables

I tried to make suggestions, but probably didn't find all occurrences of wrong model usages etc, so please have a look again at the code you have copied + pasted and try to understand what it's doing.

Apart from that: Please log into [CircleCI](https://app.circleci.com/pipelines/github/Integreat/integreat-cms?branch=develop-amin), grant it access to your GitHub-account and restart the build to fix the "unauthorized" error

timoludwig:
> The overall concept seems valid. Thank you for the contribution. Currently, lots of code is doubled. It would be cool to have some abstract classes and maybe even templates, so that the code can be shared with the page translations. @timoludwig do you agree?

@svenseeberg I have noticed a lot of potential for improvement in our views/forms/templates code (not only page/imprint) and will probably make a major refactoring in the next couple of weeks. But I think it would ok to merge this PR and refactor afterwards.

AminAlAit (commented):
.

timoludwig (approved):
@AminAlAit I rebased your branch onto the current develop branch and fixed all problems I noticed in the views. I think this is ready to merge now :tada:

PR #609, opened 2020-12-09
Read settings from environment, fixes #74

### LINKED ISSUE #74: Setup different settings files (development, production)
Use DJANGO_SETTINGS_MODULE as corresponding environment variable 

* remove EMAIL_BACKEND and setup a production ready mail server
* set DEBUG to false
* django won't handle static files anymore

svenseeberg:
Relevant settings can now be set with environment variables. If a environment variable is not set, the default value from `settings.py` is used. The `wsgi.py` needs to read the call environment variables and turn them into os.environ variables. An example config for Apache2 is included.

Fixes: #74


examples/apache2-vhost.conf
```diff
@@ -0,0 +1,40 @@
+WSGIScriptAlias / /opt/integreat-cms/src/backend/wsgi.py
```
timoludwig, line 1: I'm not sure if "examples" is a good directory name here, maybe rather "apache-conf"/"prod-conf"/"example-conf" or similar?

timoludwig (approved):
Looks good to me! :+1: :tada:

examples/apache2-vhost.conf
svenseeberg: My idea here was to potentially collect other example config files here as well. What about `example-configs` ?

timoludwig: ok, sounds good :+1:

ulliholtgrave:
Please wait a second till you merge the PR. I am currently setting up the system on a clean system and I want to try, if everything works out :)

ulliholtgrave:
Okay, everything works as expected :) 👍

PR #611, opened 2020-12-11
Fixed detecting no changes for page form

melegiul:
### Short description

Fixes change detection in page form

### Proposed changes

- Problem was that the mirror_page_first field (which was a ChoiceField) normalizes their values
  to strings (see https://docs.djangoproject.com/en/2.2/ref/forms/fields/#choicefield).
- So when comparing the initial value of the select widget True with the
  current value of the form field 'True',
  it always evaluated True != 'True'.

### Resolved issues

Fixes: #466


timoludwig (approved):
Perfect, thanks! :+1:

Issue #5, opened 2018-10-25
RSS-Feed on the Region Dashboard


svenseeberg:
### Motivation
As we are supporting a RSS-Feed on the current wordpress, we should deliver the same feature within our release. It should show the last blog entries from our website.
It not necessary to support a "already read" function. Showing the latest 5 news should be sufficient.

### Proposed Solution
It should be kind of equal to the current widget:
![grafik](https://user-images.githubusercontent.com/15608678/97785143-8af71580-1ba3-11eb-8159-36da574c65c5.png)

timoludwig:
Check existing libraries for this task, e.g. [django-feed-reader](https://pypi.org/project/django-feed-reader/) or [feedparser](https://pypi.org/project/feedparser/).

PR #613, opened 2020-12-17
Add RSS feed to dashboard

melegiul:
### Short description

Added RSS feed to dashboard

### Proposed changes

- Five latest news fetched from integreat website news feed.
- Rough outline of the remaining dashboard sections.

### Resolved issues

Fixes: #5

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,6 +34,7 @@ setup(
         "django-filer",
         "django-mptt",
         "django-widget-tweaks",
+        "feedparser",
         "idna",
         "lxml",
         "psycopg2-binary",
--- a/src/cms/static/js/toggle-dashboard-section.js
+++ b/src/cms/static/js/toggle-dashboard-section.js
@@ -1,0 +1,18 @@
+u('.collapsible').each(function(node){
+    u(node).on('click', toggle_dashboard_section);
+});
+
+/**
+ * Handles toggling the RSS feed
+ * @param {Event} e Click event for RSS feed
+ */
+function toggle_dashboard_section(e){
+    u('.collapsible-content').toggleClass('active');
+    if (u('.collapsible-content').hasClass('active')){
+        u('#up-arrow').removeClass('hidden');
+        u('#down-arrow').addClass('hidden');
+    } else {
+        u('#up-arrow').addClass('hidden');
+        u('#down-arrow').removeClass('hidden');
+    }
+}
--- a/src/cms/templates/dashboard/dashboard.html
+++ b/src/cms/templates/dashboard/dashboard.html
@@ -1,5 +1,45 @@
 {% extends "_base.html" %}
 {% load i18n %}
+{% load static %}
+{% load parse_struct_time %}
 {% block content %}
-    Dashboard
+    <div class="dashboard">
+        <h1 class="font-bold">Dashboard</h1>
+        <div class="flex items-start">
+            <div class="flex flex-col w-1/2 ">
+                <div class="bg-white p-4 my-3 mr-3 shadow">
+                    <h2 class="font-bold">{% trans 'At a glance' %}</h2>
+                </div>
+                <div class="bg-white p-4 my-3 mr-3 shadow">
+                    <h2 class="font-bold">{% trans 'Author Chat' %}</h2>
+                </div>
+            </div>
+            <div class="flex flex-col bg-white w-1/2 px-4 mt-3 ml-3 shadow">
+                <button type="button" class="collapsible font-bold py-4 text-left relative focus:outline-none">
+                    <span class="font-bold">Tür an Tür - Digitalfabrik News</span>
+                    <i data-feather="chevron-up" id="up-arrow" class="inline absolute right-0"></i>
+                    <i data-feather="chevron-down" id="down-arrow" class="inline hidden absolute right-0"></i>
+                </button>
+                <div class="collapsible-content flex flex-col active">
+                    <div class="flex py-4">
+                        <img src="{% static 'images/favicon.ico' %}"/>
+                        <div class="flex flex-col mx-auto">
+                            <h3 class="text-gray-700 font-bold">Tür an Tür - Digitalfabrik</h2>
+                            <a href="{{ home_page }}" class="font-bold text-blue-500 hover:text-blue-700">{{ domain }}</a>
+                        </div>
+                    </div>
+                    {% for entry in feed.entries %}
+                        <div class="py-4 border-b border-gray-400">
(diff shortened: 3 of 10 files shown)


Pipfile
```diff
@@ -20,6 +20,7 @@ sphinx-last-updated-by-git = "*"
 sphinx-rtd-theme = "*"
 
 [packages]
+feedparser = "*"
```
timoludwig, line 23:
```suggestion
```
Please add dependencies which are required on the production server to [setup.py](https://github.com/Integreat/integreat-cms/blob/2c5d505fd975819e461b0ab5cc24e18b7c0473a2/setup.py#L30). At the moment, the Pipfile is just for dev dependencies.

src/cms/static/js/toggle-dashboard-section.js
```diff
@@ -0,0 +1,18 @@
...
+function toggle_dashboard_section(e){
+    u('.collapsible-content').toggleClass('active');
+    if (u('.collapsible-content').hasClass('active')){
+        u('#up-arrow').removeClass('hidden');
+        u('#down-arrow').addClass('hidden');
+    } else {
+        u('#up-arrow').addClass('hidden');
+        u('#down-arrow').removeClass('hidden');
+    }
+}
```
timoludwig, line 18:
```suggestion
}

```
Please include newlines at the end of files

src/cms/views/dashboard/dashboard_view.py
```diff
@@ -36,4 +40,16 @@ def get(self, request, *args, **kwargs):
         """
 
         val = "To be defined"
-        return render(request, self.template_name, {"key": val, **self.base_context})
+        language_code = translation.get_language()
+        if language_code == "en-us":
+            feed = feedparser.parse("https://integreat-app.de/en/feed/")
+        elif language_code == "de-de":
+            feed = feedparser.parse("https://integreat-app.de/feed/")
```
timoludwig, line 47: I propose to define a new setting in [src/backend/settings.py](https://github.com/Integreat/integreat-cms/blob/develop/src/backend/settings.py), e.g.
```
RSS_FEED_URLS = {
    "en-us": "https://integreat-app.de/en/feed/",
    "de-de": "https://integreat-app.de/feed/",
}
```
and use this like
```suggestion
        language_code = translation.get_language()
        feed = feedparser.parse(RSS_FEED_URLS[language_code])
```

timoludwig (commented):
Thanks, this looks very cool and works like a charm! :+1:

I just have some minor comments:

src/cms/templates/dashboard/dashboard.html
```diff
@@ -1,5 +1,45 @@
...
+                    <span class="font-bold">Tür an Tür - Digitalfabrik News</span>
+                    <i data-feather="chevron-up" id="up-arrow" class="inline absolute right-0"></i>
+                    <i data-feather="chevron-down" id="down-arrow" class="inline hidden absolute right-0"></i>
+                </button>
+                <div class="collapsible-content flex flex-col active">
+                    <div class="flex py-4">
+                        <img src="{% static 'images/favicon.ico' %}"/>
+                        <div class="flex flex-col mx-auto">
+                            <h3 class="text-gray-700 font-bold">Tür an Tür - Digitalfabrik</h2>
+                            <a href="https://integreat-app.de/" class="font-bold text-blue-500 hover:text-blue-700">integreat-app.de</a>
```
timoludwig, line 28: Oh, and please don't hardcode this URL here as well, but add another setting :)

src/cms/templates/dashboard/dashboard.html
```diff
@@ -1,5 +1,45 @@
...
+                    <span class="font-bold">Tür an Tür - Digitalfabrik News</span>
+                    <i data-feather="chevron-up" id="up-arrow" class="inline absolute right-0"></i>
+                    <i data-feather="chevron-down" id="down-arrow" class="inline hidden absolute right-0"></i>
+                </button>
+                <div class="collapsible-content flex flex-col active">
+                    <div class="flex py-4">
+                        <img src="{% static 'images/favicon.ico' %}"/>
+                        <div class="flex flex-col mx-auto">
+                            <h3 class="text-gray-700 font-bold">Tür an Tür - Digitalfabrik</h2>
+                            <a href="{{ home_page }}" class="font-bold text-blue-500 hover:text-blue-700">integreat-app.de</a>
```
timoludwig, line 28: Maybe also replace the shown part of the link? In case the setting changes, this should probably change as well...
If you don't want to show the protocol, you can extract the domain with
```
from urllib.parse import urlparse

domain = urlparse(RSS_FEED_URLS["home-page"]).netloc
```
in the view.

setup.py
```diff
@@ -34,6 +34,7 @@
         "django-filer",
         "django-mptt",
         "django-widget-tweaks",
+        "feedparser",
```
timoludwig, line 37: Good, and now execute `./dev-tools/install.sh` one time to make sure the dependencies end up in the Pipfile.lock

melegiul: Yes, I'm just installing right now, but my pipfile.lock usually  needs some time;)

src/cms/templates/dashboard/dashboard.html
melegiul: > Maybe also replace the shown part of the link? In case the setting changes, this should probably change as well...
> If you don't want to show the protocol, you can extract the domain with
> 
> ```
> from urllib.parse import urlparse
> 
> domain = urlparse(RSS_FEED_URLS["home-page"]).netloc
> ```
> 
> in the view.

Finally the pipfile made it:smiley:, also with the domain included

src/cms/views/dashboard/dashboard_view.py
```diff
@@ -36,4 +42,23 @@ def get(self, request, *args, **kwargs):
...
-        return render(request, self.template_name, {"key": val, **self.base_context})
+        language_code = translation.get_language()
+        feed = feedparser.parse(RSS_FEED_URLS[language_code])
+        # select five most recent feeds
+        feed["entries"] = feed["entries"][:5]
+        # decode html entities like dash and split after line break
+        for entry in feed["entries"]:
+            entry["summary"] = html.unescape(entry["summary"]).split("\n")[0]
+        domain = urlparse(RSS_FEED_URLS["home-page"]).netloc
+        # breakpoint()
```
timoludwig, line 53:
```suggestion
```
I think as long as we don't use the python debugger actively in our development tooling, we should remove such comments.

src/cms/templates/dashboard/dashboard.html
timoludwig: Thanks! Apart from the superfluous comment, it looks good!

timoludwig (approved):
Perfect :+1:

PR #617, opened 2021-01-06
Fix xhtml2pdf entry in Pipfile.lock

timoludwig:
### Short description

Sometimes, pipenv does strange things, like e.g. marking dependencies as local editable packages... 
I didn't yet find the reason for this, so we have to look out for this in the future. For now, this fixes the Pipfile.lock by reverting the changes introduced in 7a61ce9f9eb333f1c3386e8bc62578a01a6f88ce.

### Proposed changes

- Fix xhtml2pdf dependency in Pipfile.lock

### Resolved issues

This should fix the failing build in #616

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -471,8 +471,10 @@
             "version": "==0.5.1"
         },
         "xhtml2pdf": {
-            "editable": true,
-            "path": "."
+            "hashes": [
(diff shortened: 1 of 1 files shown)


ulliholtgrave (approved):
Nothing really special, but with the change everything seems to work on this branch.

PR #625, opened 2021-01-09
Pass environment variables only if they are set

timoludwig:
### Short description

If I try to start the development server while no environment variables are set, I get the following error:
```
  File "/usr/lib/python3.7/wsgiref/handlers.py", line 137, in run
    self.result = application(self.environ, self.start_response)
  File "/home/timo/integreat/integreat-cms/.venv/lib/python3.7/site-packages/django/contrib/staticfiles/handlers.py", line 65, in __call__
    return self.application(environ, start_response)
  File "/home/timo/integreat/integreat-cms/src/backend/wsgi.py", line 16, in application
    os.environ["DJANGO_SECRET_KEY"] = environ["DJANGO_SECRET_KEY"]
KeyError: 'DJANGO_SECRET_KEY'
Jan 09 12:29:40 INTEGREAT CMS - ERROR: "GET / HTTP/1.1" 500 59
```

### Proposed changes

- Only pass the environment variables if they are set

### DIFF
--- a/src/backend/wsgi.py
+++ b/src/backend/wsgi.py
@@ -13,17 +13,9 @@ from django.core.wsgi import get_wsgi_application
 
 
 def application(environ, start_response):
-    os.environ["DJANGO_SECRET_KEY"] = environ["DJANGO_SECRET_KEY"]
-    os.environ["DJANGO_DEBUG"] = environ["DJANGO_DEBUG"]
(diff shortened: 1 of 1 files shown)


svenseeberg (approved):
Yes, that was a stupid mistake I made. Thanks for fixing

Issue #547, opened 2020-10-22
Add region icon to exported PDFs


timoludwig:
### Motivation

The exported PDFs should contain the region's icon in the header, see e.g. https://github.com/Integreat/cms-django/pull/534#issuecomment-713603724.

### Proposed Solution

Add the region icon in the header of the exported PDF.

### Additional Context

This is blocked by:
- [x] #35 (Pages: export PDF)
- [x] #546 (Add icon to region model)

PR #627, opened 2021-01-11
Added icon to page pdf export

melegiul:
### Short description

Added region icon to page pdf export

### Proposed changes

- Added header to PDF exports with right aligned region icon.
- Fixed footer bug: compression to a single footer, which takes the
  image size into account to avoid problems.
- Added JPG integreat logo, because xhtml2pdf still got problems with
  rendering PNG images with transparency

### Resolved issues

Fixes: #547 

### Screenshot
pdfExport


src/backend/settings.py
```diff
@@ -137,6 +137,7 @@
             "context_processors": [
                 "django.template.context_processors.debug",
                 "django.template.context_processors.request",
+                "django.template.context_processors.media",
```
timoludwig, line 140: Why is this context processor needed?
[The docs](https://docs.djangoproject.com/en/2.2/ref/templates/api/#django-template-context-processors-media) say that it only injects the `MEDIA_URL` variable, but this does not seem to be used in the template?
Or are there any other implications?

timoludwig (approved):
Works like a charm :+1:

src/backend/settings.py
melegiul: Ups, forgotten to remove that, since I decided to go just with `region.icon.path` as image source, I will remove that.

timoludwig (approved):
Event better now :+1:
Just resolve the translation conflicts and this PR can be merged.

src/cms/templates/_raw.html
```diff
@@ -11,18 +11,9 @@
...
-        These files are not included when the template is rendered in the html2pdf view,
-        because xhtml2pdf does not work in combination with tailwind css.
-        Tailwind css uses the @media without a mediatype, but xhtml2pdf expects the media identifier.
-        See https://github.com/xhtml2pdf/xhtml2pdf/issues/498 for more information.
-        Once this issue is fixed, we can use tailwind in the pdf context.
-    {% endcomment %}
     <link href="{% static 'css/style.min.css' %}" rel="stylesheet" type="text/css">
     <link href="{% static 'css/upload_form.css' %}" rel="stylesheet" type="text/css">
     <link href="{% static 'tailwindcss/dist/tailwind.css' %}" rel="stylesheet" type="text/css">
-    {% endif %}
```
melegiul, line 25: Ok, I've also shifted that warning about tailwind into the PDF html, because for now we don't need inheritance from _raw.html

timoludwig: sounds good!

Issue #631, opened 2021-01-13
get_last_root_page template tag not working


timoludwig:
### Describe the Bug

The `get_last_root_page` template tag is not working:
https://github.com/Integreat/integreat-cms/blob/develop/src/cms/templatetags/page_filters.py#L10-L20

This is because in the page QuerySet aggregation, the difference() function is used:
https://github.com/Integreat/integreat-cms/blob/a285cb3c74b2a2c501147076338e2b2a70c89bd6/src/cms/models/regions/region.py#L177

After difference(), it is not possible to use filter(), see https://docs.djangoproject.com/en/2.2/ref/models/querysets/#union

This leads to the last drop-region in the page tree being incorrectly tied to the last page in the list, not necessarily the last root page.

### Steps to Reproduce

1. Go to Page tree
2. Drag & drop page to most bottom line

### Expected Behavior

The page should be placed next to the last root page

### Actual Behavior

The page is placed next to the last page in the list

PR #632, opened 2021-01-13
Fix get_last_root_page template tag

timoludwig:
### Short description

See bug description of #631 for details.

### Proposed changes

- Use python list operations instead of Django QuerySet filters

### Resolved issues

Fixes: #631

### DIFF
--- a/src/cms/templatetags/page_filters.py
+++ b/src/cms/templatetags/page_filters.py
@@ -12,9 +12,9 @@ def get_last_root_page(pages):
     This tag returns the last page on the root level.
 
     :param pages: The requested page tree
-    :type pages: ~mptt.querysets.TreeQuerySet [ ~cms.models.pages.page.Page ]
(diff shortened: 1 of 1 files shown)


melegiul (approved):
Looks good to me, now there is no way to append existing pages to the end of a list of children, but I don't think it's a problem. Maybe we could later insert another drop field, so user can choose between options 'last child' or 'new root'.

timoludwig:
> Looks good to me, now there is no way to append existing pages to the end of a list of children, but I don't think it's a problem. Maybe we could later insert another drop field, so user can choose between options 'last child' or 'new root'.

Currently, this is possible by dropping a page onto the parent page of the last child. Hopefully the user tests will reveal whether this is intuitive or not...

melegiul:
> Currently, this is possible by dropping a page onto the parent page of the last child. Hopefully the user tests will reveal whether this is intuitive or not...

Sure, that's better than another drop field, I should have guessed that

Issue #619, opened 2021-01-08
Preview for live content


timoludwig:
### Motivation

At the moment, the aggregated content of a page including live/mirrored content can not be previewed.

### Proposed Solution

Add a possibility to preview the live/mirrored content at its specified position in combination with the original content

### Alternatives

Already implement #496 if feasible for the feature completion milestone

### Additional Context

This has a higher priority than the improved general preview (issue #496), because the wysiwyg editor already shows a rough preview of the normal content field.

melegiul:
I'm not sure if I got this right, so lets say I got page X and I set page Y as live-content after page X. When I now click the preview button ("eye icon"), then there supposed to be page X followed by page Y in the preview. Is this right?

timoludwig:
> I'm not sure if I got this right, so lets say I got page X and I set page Y as live-content after page X. When I now click the preview button ("eye icon"), then there supposed to be page X followed by page Y in the preview. Is this right?

yes, correct.

Issue #496, opened 2020-09-25
Improve content preview


timobrembeck:
### Motivation

At the moment, the page preview does not look similar to the web-app.

### Proposed Solution

Clarify with the App-Team whether it is feasible to create unique preview-links to render drafts in the web-app. This would create an exact preview of the content and the most powerful solution.

### Alternatives

If this is too much effort or not possible, recreate the look of the web-app in our local preview-view.

This includes, among others:
- [x] Correct colors of links etc
- [x] Bullet points in lists
- [x] Correct margins and max width etc.
- [x] Optional: Dummy-header that looks similar to the app

New tasks after #3396 
- [ ] Adjust the preview window so that the text is only as wide as it would be in the app
- [ ] Fix headlines
- [ ] Fix ordered list

MizukiTemma:
As #3396 the requirements for the alternativ seem to be fulfilled.

hauf-toni:
@MizukiTemma is right, the alternative is also covered. @osmers was just wondering whether the width of the preview window should be adjusted so that the text is only as wide as it would be in the app

PR #634, opened 2021-01-13
Added live content preview for pages

melegiul:
### Short description

Add live content to page preview

### Proposed changes

- Pages with embedded live content are previewed with their live
  content.
- This does not address issue #496 (general improved preview) due to
  missing links for a web-app based preview.

### Resolved issues

Fixes: #619

### DIFF
--- a/src/cms/templates/pages/page_view.html
+++ b/src/cms/templates/pages/page_view.html
@@ -2,9 +2,20 @@
 {% load i18n %}
 
 {% block raw_content %}
-<h1>{{page_translation.title}}</h1>
-
-<div style="padding-top: 30px">
-    {{ page_translation.text|safe }}
+<div class="px-4">
+    <h1 class="pt-4 font-bold text-lg">{{page_translation.title}}</h1>
+    {% if mirrored_translation and mirrored_page_first %}
+        <div class="pt-2">
+            {{ mirrored_translation.text|safe }}
+        </div>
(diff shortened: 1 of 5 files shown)


src/cms/templates/pages/page_view.html
```diff
@@ -2,9 +2,22 @@
...
+        <div class="pt-2">
+            {{ mirrored_translation.text|safe }}
+        </div>
+    {% endif %}
+    <h1 class="pt-4">{{page_translation.title}}</h1>
+    <div class="pt-2">
+        {{ page_translation.text|safe }}
+    </div>
+    {% if mirrored_translation and not mirrored_page_first %}
+        <h1 class="pt-4">{{mirrored_translation.title}}</h1>
```
svenseeberg, line 17: Usually we just concatenate the strings of the content. I'm not sure if we want to show the title here. In my opinion we should just add the two content strings in the correct order and display them.

svenseeberg (commented):
That's an elegant solution, I like it  :+1:. Just the one thing with the headline.

src/cms/templates/pages/page_view.html
melegiul: Alright, so that means we just want to show the title of the main page and omit the title of the mirrored page. I will adjust that.

timoludwig (approved):
Looks good now :+1:

Issue #635, opened 2021-01-14
Page saving message not correct


melegiul:
### Describe the Bug

Wrong information message when editing live/mirrored content of pages

### Steps to Reproduce

1. Go to regions page menu and select a page
2. Add some live/mirrored content to the page
3. Publish/save changes
4. The message says no changes were made

### Expected Behavior

Message should confirm, that changes were successfully saved.

### Actual Behavior

No changes were detected

### Additional Information

This bug only concerns the displayed message, nevertheless live/mirrored content is saved correctly to database and also shows up in the populated form.

timoludwig:
This problem seems to be related to #466.

PR #637, opened 2021-01-14
Refactor mirrored page field

timoludwig:
### Short description

Refactor mirrored page field

### Proposed changes

- Save field only when form is submitted (fixes the submission message of the page form)
- Prevent page embedding itself
- Add warning when archived page is embedded
- Add warning when page from archived or hidden region is embedded

### Resolved issues

Fixes: #635

### DIFF
--- a/src/cms/forms/pages/page_form.py
+++ b/src/cms/forms/pages/page_form.py
@@ -1,13 +1,16 @@
 import logging
 
+from html import escape
+
 from django import forms
 from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Permission
 from django.db.models import Q
 from django.urls import reverse
 from django.utils.translation import ugettext as _, get_language
+from django.utils.safestring import mark_safe
 
-from ...constants import position
+from ...constants import position, region_status, mirrored_page_first
 from ...models import Page, Region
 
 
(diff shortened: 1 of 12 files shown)


src/cms/forms/pages/page_form.py
```diff
@@ -76,12 +76,30 @@ class MirrorPageField(forms.ModelChoiceField):
...
-        return " -> ".join(
+        label = " 🡒 ".join(
             [
                 page_iter.best_language_title()
                 for page_iter in page.get_ancestors(include_self=True)
             ]
         )
+        if page.archived:
+            label += " (⚠ " + _("Archived") + ")"
+        return label
```
melegiul, line 87: Seems like some kind of encoding error
### Before
before
### Afterward
afterwards

melegiul (changes requested):
LGTM, fixes the problem about wrong messages totally.
Apart from that I just found a minor encoding issue and I got one question.

src/cms/urls.py
```diff
@@ -291,6 +291,18 @@
...
+                    include(
+                        [
+                            url(
+                                r"^mirrored_page_field/",
+                                pages.render_mirrored_page_field,
+                                name="render_mirrored_page_field",
+                            ),
+                        ]
+                    ),
+                ),
```
melegiul, line 305: Why is it necessary to use ajax here? The only case I can imagine is, when the user opens the form and wants keep track of changes to all other regions pages without refreshing, is this right?

melegiul, line 305: Why is it necessary to use ajax here? The only case I can imagine is, when the user opens the form and wants keep track of changes to all other regions pages without refreshing, is this right?

src/cms/forms/pages/page_form.py
timoludwig: Interesting, in my browser it works:tm:
I tried a different approach now, can you check whether this works on your system? Thanks!

src/cms/urls.py
timoludwig: Yes, it's not technically necessary, it just avoids a huge overhead when rendering the page form... otherwise the `mirrored_page_field` would have to be rendered for every city in the cms beforehand (and thus one option for every page object in the whole cms... in the actual live data, this can exceed a few thousand entries). Since the live content region is only changed rarely, we don't need to render a few thousand option tags every time a page form gets opened.

src/cms/forms/pages/page_form.py
melegiul: Yes, now the arrow is rendered correctly:tada:. I'm on firefox, so might have been a firefox issue.

melegiul (approved):

src/cms/urls.py
melegiul: I see, so it prevents database from being overload:thumbsup:.

Issue #292, opened 2020-02-14
Consistent list representations


florianhofhammer:
~~Currently, the event list is somewhat broken and not displayed correctly (see figure 1 below). This should be fixed.~~
Apart from that, the different lists (especially for pages, events and pois) aren't rendered consistently. For example, they use different shades of grey for the table borders. It should be investigated in which points they really differ (apart from the given example) as well as those differences should be eliminated in order to have a consistent UI experience.

The work on this issue probably doesn't contain any logic fiddling and only requires changes to the templates (i.e. probably only changing some css classes of html elements).

![Screenshot_2021-03-18 Integreat Redaktionssystem(1)](https://user-images.githubusercontent.com/16021269/111649780-fc5afc00-8804-11eb-941a-1f148d480a85.png)
![Screenshot_2021-03-18 Integreat Redaktionssystem](https://user-images.githubusercontent.com/16021269/111649793-febd5600-8804-11eb-9b28-21d950b07c4a.png)

timoludwig:
The list width is fixed in  #286. Not sure about the shadows though.

florianhofhammer:
I'm sorry, I missed this fix. I'll lower the priority of this issue.

svenseeberg:
Also, its worth noting that we have #172 still open. It would be cool to start creating meta classes for lists.

ulliholtgrave:
This issue will be resolved, when we fully cover our CSS classes (as suggested in https://github.com/Integreat/integreat-cms/issues/172). Besides that, we solved a lot of the problems in the last merges by accident :)

PR #639, opened 2021-01-20
Consistent list representations

mael-fosso:
### Short description
changing some css classes of html elements.

### Proposed changes
Apart from that, the different lists (especially for pages, events and pois) aren't rendered consistently. For example, they use different shades of grey for the table borders. It should be investigated in which points they really differ (apart from the given example) as well as those differences should be eliminated in order to have a consistent UI experience.

### Resolved issues

Fixes: #292


timoludwig:
There are no real changes, did you forget to commit something?
Keep in mind that changes to css should usually go into the file [style.less](https://github.com/Integreat/integreat-cms/blob/develop/src/cms/static/css/style.less) which is then compiled to the style.min.css at runtime.
If you did not commit something on purpose and just wanted to create the draft PR, please don't request a review from us unless you want us to take a look at the PR. Thanks!

timoludwig:
@emmakemb @mael-fosso I force-pushed your branch `feature/list-representations` to the current state of the develop-branch to make sure you don't run into nasty rebasing issues once you start with your work. Please execute
```
git fetch --all
git reset --hard origin/feature/list-representations
```
before making any other changes to ensure your local branch isn't out of date.
I wasn't aware that GitHub automatically closes a PR if it's even with the develop-branch, so just re-open it again after you pushed your new changes.

Issue #618, opened 2021-01-08
Use currently_in_translation flag


timoludwig:
### Motivation

Make use of the `currently_in_translation` flag in the content translation models.

### Proposed Solution

- [x] Set `currently_in_translation` flag when XLIFF is exported
- [x] Unset `currently_in_translation` flag when XLIFF is imported
- [x] Add possibility to manually unset `currently_in_translation` flag if translation process is cancelled
- [x] Show new icon for translations if the `currently_in_translation` flag is set (to make sure it is not shown as "outdated" anymore)
- [x] Show warning if a page is edited while the `currently_in_translation` flag is set (**optional/open debate**)

timoludwig:
@melegiul as discussed in Mattermost, when xliffs are exported for a non-existing translation, we should **not** create an empty translation just to set the "currently in translation" flag. The reason behind this is that often times, xliffs are just exported to get an offer from the translation service, but if that's too expensive, the translation process is not completed and then all empty translations have to be deleted again. Thus, they should only be created when the finished xliff is imported again.

melegiul:
>     * Show new icon for translations if the `currently_in_translation` flag is set (to make sure it is not shown as "outdated" anymore)

So when I understand right, that means we just want to update state "outdated" -> "in process", that means actually setting the "currently_in_translation" field in database. But for all "non-existing" translations we want to "ignore" the xliff export in terms of currently_in_translation flag. Is this right?

timoludwig:
Yes, exactly.

PR #640, opened 2021-01-21
Update currently in translation state on import/export XLIFF

melegiul:
### Short description

Update the translation state when XLIFF is imported or exported

### Proposed changes

- When XLIFF is exported, ajax request is send to update database entry of the translation and display clock icon to signal the successful update. 
- When XLIFF is imported, translation is created with currently_in_translation set to false.
- Warning in page form, when page is in translation process + ability to undo process

### Resolved issues

Fixes: #618

<del>### Open issue
I experienced behavior, that maybe you might not expect:</del>
<del>1. go to page tree of augsburg
2. export english XLIFF, which has 'up-to-date' arabic  pages (e.g. page "Stadtplan" from Augsburg)
3. refresh the website
4. see that initially 'up-to-date' arabic is no outdated</del>

<del>I guess that's due to the translation changes affecting the last_updated value, up for discussion if this needs a fix or not.</del>

### DIFF
--- a/src/cms/static/js/pages/page_bulk_actions.js
+++ b/src/cms/static/js/pages/page_bulk_actions.js
@@ -2,44 +2,58 @@
  * This file contains all functions which are needed for the bulk actions for pages.
  */
 
-function select_all_pages() {
-    var selected_pages = document.getElementsByName("page_selected");
-    if (document.getElementsByName("select_all_pages")[0].checked == true) {
-        selected_pages.forEach(e => e.checked = true);
-    } else {
-        selected_pages.forEach(e => e.checked = false);
-    }
-}
-
-function bulk_action_execute() {
-    var action = document.getElementById('bulk_action').value;
-    // match action types
-    if ( action == "archive_pages" ) {
-        // TODO
-    } else if (action == "pdf_export") {
+u(document).on('DOMContentLoaded', function () {
+    
+    u('[name="select_all_pages"]').on('click', select_all_pages);
+    function select_all_pages() {
         var selected_pages = document.getElementsByName("page_selected");
-        pages = [];
-        selected_pages.forEach(e => {
-            if(e.checked) {
-                pages.push(e.value)
-            }
-        })
-        if (pages.length != 0) {
-            var pdf_export_url = document.getElementById("pdf_export_url").value;
-            window.open(pdf_export_url+"?pages="+pages.join(','));
-        } 
-    } else { // no previous match, than language code -> XLIFF export
-        var page_selected = document.getElementsByName("page_selected");
+        if (document.getElementsByName("select_all_pages")[0].checked == true) {
+            selected_pages.forEach(e => e.checked = true);
+        } else {
+            selected_pages.forEach(e => e.checked = false);
+        }
+    }
+
+    u("#bulk_action_btn").on('click', bulk_action_execute);
+    function bulk_action_execute() {
+        var action = u('#bulk_action').first().value;
         var pages = [];
-        page_selected.forEach(async function (page_checkbox) {
-            if(page_checkbox.checked) {
-                pages = pages.concat(page_checkbox.value);
-            }
-        });
+        u('input').filter(':checked').each((node, i) => pages.push(node.value));
         if (pages.length == 0) {
+            // abort if no pages selected
             return;
         }
-        var bulk_action_url = document.getElementById("bulk_action_url").value;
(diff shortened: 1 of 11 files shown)


src/cms/static/js/pages/update_translation_state.js
```diff
@@ -0,0 +1,21 @@
+/*
```
svenseeberg, line 1: Just a note for discussion: until now we put all JS functions for one page in one file. Nothing wrong with splitting it into more files.

src/cms/templates/pages/page_form.html
```diff
@@ -161,7 +174,18 @@ <h2 class="heading font-normal">
                         <a href="{{ page_translation_form.instance.short_url }}" target="_blank" rel="noopener noreferrer" class="text-blue-500 hover:underline">{{ page_translation_form.instance.short_url }}</a>
                         <a href="#" onclick="short_url_to_clipboard('{{ page_translation_form.instance.short_url }}');" title="{% trans 'Copy to clipboard' %}" class="px-2 text-gray-800 hover:text-blue-500">
                             <i data-feather="copy" class="inline-block"></i>
-                        </a>
+                        </a><br>
+                        {% if page_translation_form.instance.currently_in_translation %}
+                            <i data-feather="alert-triangle" class="inline-block text-red-800"></i>
+                            <label class="inline-block mb-2 font-bold">{% trans 'Warning' %}:</label>
+                            {% trans 'The page is currently in translation for this language' %}
```
svenseeberg, line 181: What about a simple `Translation in progress` as an alternative to `The page is currently in translation for this language`? Would be shorter and I think it conveys the same information.

src/cms/static/js/pages/page_bulk_actions.js
```diff
@@ -2,44 +2,58 @@
...
+            var pdf_export_url = document.getElementById("pdf_export_url").value;
+            window.open(pdf_export_url+"?pages="+pages.join(','));
+        } else { // no previous match, than language code -> XLIFF export
+            const state = true;
+            const url = u("#ajax_url").first().dataset.url;
+            var bulk_action_url = document.getElementById("bulk_action_url").value;
+            // download XLIFF
+            window.open(bulk_action_url+"?target_lang="+action+"&pages="+pages.join(','), "_blank");
+            // update state of translations to currently_in_translation
+            post_translation_state(url, state, pages, action)
```
svenseeberg, line 39: This seems inefficient to me. We could set the translation state during the export instead of sending a second HTTP request to the server. However, I see the advantage of having just one mechanism for setting the state. This could be done somewhere around https://github.com/Integreat/integreat-cms/blob/develop/src/cms/views/pages/page_actions.py#L313

src/cms/static/js/pages/update_translation_state.js
timoludwig: I started separating the JS files for pages a while ago (c3b1153d1775bf75c1978020e3358faf886bb45a), because the single file was a bit overloaded and some functions are required on multiple templates, others only on individual ones.
I think it makes sense to have one JS-file per "functionality", even if this means many small files...

svenseeberg: Alright then, I'll try to remember to do this myself :+1:

src/cms/static/js/pages/page_bulk_actions.js
melegiul: Yes, the reason for two requests was, that one responses a zip file while the other responses those selected pages, that actually have an existing translation for updating the icon to "translation in progress". But of course I can circumvent that second one by saving all existing translation languages of a page somewhere in the DOM and just take that value. That would halve the number of requests and even remove one of the scattered js files, which I added for this PR.

src/cms/static/js/pages/page_bulk_actions.js
```diff
@@ -2,44 +2,58 @@
...
+        }
+    }
+
+    function updateTranslationIcon(pages, languageCode) {
+        for (page of pages){
+            let pageNode = u(`#page-${page}`);
+            // get language for updating translation state icon
+            let targetLanguage = pageNode.find(`.${languageCode}`);
+            if (!targetLanguage.find("#translation-icon").find("span").hasClass("no-trans")) {
+                // if the page already contains a translation for the exported language
```
melegiul, line 48: Second server request is no replaced by finding the existing translations of selected pages on DOM-side.

svenseeberg: Awesome!

svenseeberg (approved):
Works for me! Thanks.

PR #641, opened 2021-01-21
XLIFF: import error for missing page, fixes #622

### LINKED ISSUE #622: XLIFF import after page deletion
### Describe the Bug

When the user imports XLIFF file to non existing page, nothing happens.

### Steps to Reproduce

1. Select a page
2. export it as XLIFF
3. delete the page
4. import the XLIFF again

### Expected Behavior

The user should be notified, that he tries to add a translation for a page that does not exist.

### Actual Behavior

Nothing shows up

### Additional Information

When a bundle of XLIFFs is imported, only these with existing pages should actually be imported, for all others there should be a notification: "Action not possible".

svenseeberg:
### Short description
When importing a XLIFF file, the attempt to import a translation for a not existing pages or language produces an error in the diff list.

image

### Proposed changes
This PR hands the information about failed imports to the template, where an alternative to the diff is rendered.

### Resolved issues
Fixes: #622

### DIFF
--- a/src/cms/page_xliff_converter.py
+++ b/src/cms/page_xliff_converter.py
@@ -479,10 +479,10 @@ class PageXliffHelper:
         :param xliff_paths: list of paths to XLIFF files
         :type xliff_paths: [ str ]
 
-        :return: dictionaries containing diffs between XLIFF and current translation versions
+        :return: dictionaries containing diffs between XLIFF and current translation versions or an error message
         :rtype: list [ dict ]
         """
-        diffs = []
+        results = []
         for xliff_path in xliff_paths:
             if xliff_path.endswith((".xlf", ".xliff")) and os.path.isfile(xliff_path):
                 with open(xliff_path, "r", encoding="utf-8") as f:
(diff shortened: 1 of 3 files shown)


src/cms/page_xliff_converter.py
```diff
@@ -491,11 +491,20 @@ def generate_xliff_import_diff(self, xliff_paths):
...
-                        self.generate_translation_diff(
-                            trans_fields, os.path.basename(xliff_path)
-                        )
+                    diff = self.generate_translation_diff(
+                        trans_fields, os.path.basename(xliff_path)
                     )
+                    if diff is None:
+                        diffs.append(
+                            {
+                                "error": True,
```
svenseeberg, line 500: I'm not a 100% happy with this solution, as this is an "abuse" of the structure of the `diffs` list/dict. However, I did like the alternative of returning 2 variables or even restructuring everything even less.

src/cms/templates/pages/page_xliff_confirm.html
```diff
@@ -29,21 +29,31 @@
...
     </div>
     {% for xliff_diff in translation_diffs %}
-    <div class="border border-green-500 m-4">
+    <div class="border border-{% if xliff_diff.error %}red{% else %}green{% endif %}-500 m-4">
+        {% if xliff_diff.error %}
+        <h2 class="heading font-normal m-4">{% trans 'Translation cannot be imported' %}</h2>
+        <p class="m-4">{% trans 'The page or target language does not exist.' %}
+        <ul class="list-disc m-4 pl-4">
+            <li class="">{% trans 'Title' %}: {{ xliff_diff.title }}</li>
+            <li class="">{% trans 'Page ID' %}: {{ xliff_diff.page_id }}</li>
```
svenseeberg, line 46: AFAICT this is the only place where we actually show the page ID in the user interface (beyond the address bar). However, I consider this a necessary information in this case, as the ID is used for mapping the XLIFF to a page and this is the only clue about what went wrong.

src/cms/page_xliff_converter.py
timoludwig: We could rename the variable `diff` to `result` to indicate that it can hold also non-diff-objects, but in general I think this is totally fine.

timoludwig (approved):
Works as expected and I don't have any objections against putting errors in the diff-list. :+1:

PR #643, opened 2021-01-23
Combined last updated date for page with live content

### LINKED ISSUE #628: Changed date of page with live content
### Describe the Bug

At the moment, live content is not taken into account by the `modified_gmt`/`last_updated` field in the API. 

### Steps to Reproduce

1. Add live content to a page 
2. Retrieve page api endpoint 
3. See that the `modified_gmt` field did not change 

### Expected Behavior

- Use the source page time stamp, if the source page is making up 100% of the content
- Use the target page timestamp if content is mixed.

### Actual Behavior

The live content page does not impact the changed date.

svenseeberg:
### Short description
If a page is mirrored, the modified date in the API should be the one of the mirrored page, if the page translation itself has no content.

### Proposed changes
This adds a property `combined_last_updated` to the `PageTranslation` model that works in a similar way as the combined text property.

### Resolved issues

Fixes: #628

### DIFF
--- a/src/cms/models/pages/page_translation.py
+++ b/src/cms/models/pages/page_translation.py
@@ -119,6 +119,20 @@ class PageTranslation(AbstractBasePageTranslation):
             return attached_text + self.text
         return self.text + attached_text
 
+    @property
+    def combined_last_updated(self):
+        """
+        This function combines the last_updated date from this PageTranslation and from a mirrored page.
+        If this translation has no content, then the date from the mirrored translation will be used. In
+        other cases, the date from this translation will be used.
+
+        :return: The last_upated date of this or the mirrored page translation
+        :rtype: ~datetime.datetime
+        """
+        if self.page.get_mirrored_page and not self.text:
+            return self.page.get_mirrored_page(self.language.code).last_updated
+        return self.last_updated
+
     @classmethod
     def get_translations(cls, region, language):
         """
--- a/src/api/v3/pages.py
+++ b/src/api/v3/pages.py
@@ -40,7 +40,7 @@ def transform_page(page_translation):
         "url": page_translation.permalink,
         "path": page_translation.slug,
         "title": page_translation.title,
-        "modified_gmt": page_translation.last_updated,
+        "modified_gmt": page_translation.combined_last_updated,
         "excerpt": page_translation.text,
         "content": page_translation.combined_text,
         "parent": parent,
--- a/src/cms/models/pages/page.py
+++ b/src/cms/models/pages/page.py
@@ -137,7 +137,7 @@ class Page(MPTTModel, AbstractBasePage):
         :rtype: str
         """
         if self.mirrored_page:
-            return self.mirrored_page.get_translation(language_code)
+            return self.mirrored_page.get_public_translation(language_code)
         return None
 
     class Meta:


src/cms/models/pages/page_translation.py
```diff
@@ -119,6 +119,20 @@ def combined_text(self):
...
 
+    @property
+    def combined_last_updated(self):
+        """
+        This function combines the last_updated date from this PageTranslation and from a mirrored page.
+        If this translation has no content, then the date from the mirrored translation will be used. In
+        other cases, the date from this translation will be used.
+
+        :return: The last_upated date of this or the mirrored page translation
+        :rtype: ~django.db.models.DateTimeField
```
timoludwig, line 130:
```suggestion
        :rtype: ~datetime.datetime
```
The property doesn't return the field, it returns the content of the field. Thus, the return type is `datetime` directly.

src/cms/models/pages/page_translation.py
```diff
@@ -119,6 +119,20 @@ def combined_text(self):
...
+        """
+        if self.page.get_mirrored_page is not None and not self.text:
```
timoludwig, line 132:
```suggestion
        if self.page.get_mirrored_page and not self.text:
```
There is no way a foreign key contains something which evaluates to `False` but is not `None`, so we don't need to be that explicit here.

src/cms/models/pages/page_translation.py
```diff
@@ -119,6 +119,20 @@ def combined_text(self):
...
+            return self.page.get_mirrored_page(self.language.code).last_updated
```
timoludwig, line 133: I noticed that `get_mirrored_page()` uses `get_translation()` internally:
https://github.com/Integreat/integreat-cms/blob/000f2abdebd49ef17d13d7640c3b96f2413ca44c/src/cms/models/pages/page.py#L140
While you're at it, would it make sense to change that to `get_public_translation()`?

timoludwig (commented):
Thanks!
This only affects the date when the text is empty, but in the issue description is stated that the date of the live content should also be used if the mirrored page is newer than the page containing the live content. Is this a mistake in the issue, or should we change this in the `combined_last_updated()`?

src/cms/models/pages/page_translation.py
svenseeberg: Yes, I can do that.

svenseeberg:
> This only affects the date when the text is empty, but in the issue description is stated that the date of the live content should also be used if the mirrored page is newer than the page containing the live content. Is this a mistake in the issue, or should we change this in the `combined_last_updated()`?

I just looked it up in the chat history. I made a final suggestion with one "thumbs up" and no "against" votes. That suggestion was:
* Use the source page time stamp, if the source page is making up 100% of the content
* Use the target page timestamp if content is mixed.

src/cms/models/pages/page_translation.py
svenseeberg: I need to get used to this concept :grimacing:

svenseeberg:
@timoludwig does this need more discussion or should I change the issue for reference?

timoludwig (approved):
> @timoludwig does this need more discussion and/or should I change the issue for reference?

No, thanks for the clarification. I'm still not happy with using the term "last updated" if it doesn't reflect the date when the content changed the last time, but that seems more of an app issue.
Maybe, if we change the wording to "last reviewed" in the app, it makes sense to also rename the field in Django, but this is out of scope for this PR. :slightly_smiling_face:

PR #644, opened 2021-01-23
Update production info in README

svenseeberg:
Minor updates to the README to improve the info about the current production setup.


timoludwig (approved):
So need to run `collectstatic` in the moment? Are static files served through Django in this setup?

svenseeberg:
In the Apache2 example configuration we have an external static directory.

timoludwig:
> In the Apache2 example configuration we have an external static directory.

But how will the static files end up there? I think it makes sense to add an explanatory step to the guide if there is something which has to be done.

PR #645, opened 2021-01-27
API imprint: return default language, fixes #621

### LINKED ISSUE #621: Deliver fallback of missing imprint translations in API
### Motivation

The imprint is mandatory for all regions and languages.

### Proposed Solution

Always return a result in the [imprint API](https://github.com/Integreat/integreat-cms/blob/develop/src/api/v3/imprint.py). If the translation is missing, deliver the imprint in the region's default language.

svenseeberg:
If no imprint translation is available in the selected language, the imprint is returned in the default region language.

Co-authored-by: Sandro Wiedemann <sandro.wiedemann@gmail.com>

Fixes: #621

### DIFF
--- a/src/api/v3/imprint.py
+++ b/src/api/v3/imprint.py
@@ -41,7 +41,9 @@ def transform_imprint(imprint_translation):
 # pylint: disable=unused-argument
 def imprint(request, region_slug, language_code):
     """
-    Get imprint for language and return JSON object to client
+    Get imprint for language and return JSON object to client. If no imprint translation
+    is available in the selected language, try to return the translation in the region
+    default language.
 
     :param request: Django request
     :type request: ~django.http.HttpRequest
@@ -58,5 +60,11 @@ def imprint(request, region_slug, language_code):
         imprint_translation = region.imprint.get_public_translation(language_code)
         if imprint_translation:
             return JsonResponse(transform_imprint(imprint_translation))
+        if region.default_language:
+            imprint_default_translation = region.imprint.get_public_translation(
+                region.default_language.code
+            )
+            if imprint_default_translation:
+                return JsonResponse(transform_imprint(imprint_default_translation))
     # If imprint does not exist, return an empty response. Turn off Safe-Mode to allow serializing arrays
     return JsonResponse([], safe=False)


timoludwig (approved):
Looks good! :+1:

src/api/v3/imprint.py
```diff
@@ -58,5 +60,10 @@ def imprint(request, region_slug, language_code):
         imprint_translation = region.imprint.get_public_translation(language_code)
         if imprint_translation:
             return JsonResponse(transform_imprint(imprint_translation))
+        imprint_default_translation = region.imprint.get_public_translation(
+            region.default_language.code
```
timoludwig, line 64: If we want to be 100% sure to not throw an error, we should check whether `region.default_language` exists before accessing the `code` field. But this only affects an edge case where a new region is created without any languages...

svenseeberg: @Wiedemsa do you want to give it a try? :-)

Wiedemsa: i will check it out :)

svenseeberg:
@timoludwig do have any idea why docker-images says unauthorized?

src/api/v3/imprint.py
```diff
@@ -58,5 +60,10 @@ def imprint(request, region_slug, language_code):
         imprint_translation = region.imprint.get_public_translation(language_code)
         if imprint_translation:
             return JsonResponse(transform_imprint(imprint_translation))
+        if region.default_language:
+		    imprint_default_translation = region.imprint.get_public_translation(
+	            region.default_language.code
+        	)
+        	return JsonResponse(transform_imprint(imprint_default_translation))
```
timoludwig, line 67: Even if the default language exists, we don't know if there's a public translation of the imprint in the default language.
Thus, I think we should also check this here (`transform_imprint` will throw an error when called with the argument `None`).

timoludwig (commented):
> @timoludwig do have any idea why docker-images says unauthorized?

Yes, The problem is, that Sandro has to [connect his GitHub account to CircleCI](https://circleci.com/vcs-authorize/), so that CircleCI can verify he is really part of the Integreat/cms team and grant him access to the password of our organization's docker hub account (which is required for the `docker-images`-job). I also added this troubleshooting guide in PR #568 (2f11689304c51160b183decc951353897e94fa8a).

src/api/v3/imprint.py
svenseeberg: Yes, we found this 2 seconds after the push :D

timoludwig (approved):

Issue #6, opened 2018-10-25
Chat for authors


svenseeberg:
Create a chat that can be used to communicate between CMS users. It should be possible to create different channels, for example within one site, or a global chat. If messages are new (unread), they should be highlighted. A timestamp and the sender name should be displayed besides the message.

The chat should be integrated in the Dashboard.

timoludwig:
Before this feature is implemented, it should be evaluated whether there is an open source solution which can be integrated without much effort.

PR #646, opened 2021-01-28
Author Chat

timoludwig:
### Short description

This PR adds a simple chat for authors. It is a global chat where all users of all regions can communicate.

### Features

- Asynchronous messaging
- Highlight unread messages
- Convert URLs in messages to links
- Add Integreat icon next to the sender's name for Integreat staff members
- Enable message deletion (Integreat staff can delete all messages, other users just their own)

author-chat

### Technical details
- Add setting `AUTHOR_CHAT_HISTORY_DAYS` to configure the number of days in the chat history
- Add model `ChatMessage` with custom manager `ChatHistoryManager` which returns the chat history based on the setting
- Add field `chat_enabled` to region model
- Add field `chat_last_visited` to user profile model to determine read-status
- Add rule-based permission `cms.delete_chat_message` to determine whether a user is allowed to delete a specific chat message
- Add `ChatContextMixin` which can be inherited in all views which want to include the chat template (currently: admin dashboard and region dashboard)
- Improve confirmation popups to support ajax actions
- Handle network & server errors

### Resolved issues

Fixes: #6


melegiul (approved):
Awesome chat functions, works as expected.
I could just find non-blocking minor design issue - probably related to firefox
and empty parentheses for unset user email, I think in this case the parentheses should not show up.
chat-review

timoludwig:
> I could just find non-blocking minor design issue - probably related to firefox

I think this might be the spacer for the scroll bar - I changed the css property from `overflow-scroll` to `overflow-auto`, can you check whether this fixes the issue for you?

> and empty parentheses for unset user email, I think in this case the parentheses should not show up.

Thanks, totally forgot that email is an optional field...

melegiul:
> > I could just find non-blocking minor design issue - probably related to firefox
> 
> I think this might be the spacer for the scroll bar - I changed the css property from `overflow-scroll` to `overflow-auto`, can you check whether this fixes the issue for you?
> 
> > and empty parentheses for unset user email, I think in this case the parentheses should not show up.
> 
> Thanks, totally forgot that email is an optional field...

Yes, perfect. With `overflow-auto` it looks better now and the parentheses are no longer displayed too:tada:

melegiul:
This might be beyond the scope of this PR, but have you noted that timestamps for chat messages (and timestamps for other models as well) are displayed as UTC time, which has 1 hour deviation from our timezone (Europe/Berlin)?

timoludwig:
> This might be beyond the scope of this PR, but have you noted that timestamps for chat messages (and timestamps for other models as well) are displayed as UTC time, which has 1 hour deviation from our timezone (Europe/Berlin)?

The timezone is configured here: https://github.com/Integreat/integreat-cms/blob/c491eef06180c9370fcdfbc2b34181d729a14158/src/backend/settings.py#L230

I think this is a problem of all timestamps in the whole CMS - probably we have to convert all timestamps to local time before showing them to the users... Good point, I will open an issue.

PR #647, opened 2021-01-28
Add troubleshooting for developing on arm macs

jnugh:
### Short description
I had some issues while getting the application running on an M1 Macbook but was eventually able to solve the issues.

### Proposed changes
I explain:
* how to use a different python version
* how to compile postgres to get psycopg2 installed

It seems, like both issues will eventually go await once everyone adopts their software for the arm platform, but maybe it makes sense to have those steps in here for the meantime 😄 .


timoludwig (approved):
Amazing, thanks! Glad it worked out in the end :+1:

Issue #499, opened 2020-09-25
Simple mode/expert mode


timoludwig:
### Motivation

Some functions are hard to use or more or less in a beta-state. To ensure that no additional training is needed after the migration of the CMS from WordPress to Django, we should disable these functions by default and only enable them on request.

Expert functions:
- Live Content / Mirror pages
- Do-Not-Translate-Tag
- Recurring Events
- Page-based permissions ("editors"/"publishers")
- XLIFF functions (XLIFF export and XLIFF import)
- Page filter

### Proposed Solution

Add two modes ("Simple mode"/"Expert mode"). Make some functionalities only accessible for ~regions~ users which have the "expert mode" activated.

As far as I can tell, the preferred place to save the expert mode enabled/disabled state is a field in the UserProfile model.

timoludwig:
We could also think of this feature from a pure esthetics standpoint and just hide expert functions with css without bothering for permissions.

melegiul:
![expert-mode](https://user-images.githubusercontent.com/47010475/106130565-cdf46500-6161-11eb-8e3c-50c37efcee94.png)

So I suggest the expert-modes interface looking like this. But now I see two options how to elaborate this.
1. I could take the existing database model ```config/configuration.py``` and create a key/value pair in the database for each settings entry like "expert-mode" and save the settings that way.
2. I could create a new database model + form called "user-settings" to make the settings to  a form like the already existing ones (page form, etc.) including form validation, ...

In both cases I would tend to sending asynchronous messages to server rather than a submission button, especially for the expert mode checkbox. The reason for that is, the user should be able to just set/unset the checkbox and update database instantly without taking care about the other settings entries and the submission button. 

@timoludwig @ulliholtgrave @svenseeberg Which one would you prefer?

melegiul:
> We could also think of this feature from a pure esthetics standpoint and just hide expert functions with css without bothering for permissions.

Whats about using context variables and template tags instead of css. Like in the view:
```
context = {
// true or false
"expert-mode": models.settings.expert_mode
}
```
... and in the template:
```
{% if expert-mode %}
// only show up when expert mode is enabled
{% endif %}
```

timoludwig:
> So I suggest the expert-modes interface looking like this. But now I see two options how to elaborate this.
> 
> 1. I could take the existing database model `config/configuration.py` and create a key/value pair in the database for each settings entry like "expert-mode" and save the settings that way.
> 
[...]

PR #648, opened 2021-01-29
Added expert mode

melegiul:
### Short description

Added expert mode for displaying additional features.

### Proposed changes

Enabling expert mode includes the following features:
- Page mirroring
- Recurring events
- Do-Not-Translate-Tag in page editor
- Page-based permissions
- XLIFF import/export
- Page filtering

### Additional context

Expert mode setting is saved per user account, additional expert-mode-checkbox for (region-)user supplied.
Expert mode does not add an additional role to database, because expert mode does not deal with any additional permissions when enabled. Disabling expert mode just simplifies the GUI by hiding the aforementioned features.

### Resolved issues

Fixes: #499

### Screenshot

expert_mode

### DIFF
--- a/src/cms/templates/users/admin/user.html
+++ b/src/cms/templates/users/admin/user.html
@@ -84,6 +84,14 @@
                 </div>
 
                 <div class="py-4 border-b solid border-gray-200">
+                    <label for="id_expert_mode" class="font-bold mr-2 cursor-pointer">{% trans 'Expert mode' %}</label>
+                    {% render_field user_profile_form.expert_mode %}
+                    <p class="pt-2">
+                        {% trans 'Enable this option to show up additional features like XLIFF import/export, page filtering, mirrored pages, page-based permissions, Do-Not-Translate-Tag and recurring events' %}
+                    </p>
+                </div>
+                
+                <div class="py-4 border-b solid border-gray-200">
                     <label for="id_roles" class="font-bold block mb-2 cursor-pointer">{% trans 'Roles' %}</label>
                     {% render_field user_form.roles class+="w-full" %}
                 </div>
--- a/src/cms/fixtures/test_data.json
+++ b/src/cms/fixtures/test_data.json
@@ -23,7 +23,8 @@
     "fields": {
       "user": 1,
       "organization": null,
-      "regions": []
+      "regions": [],
+      "expert_mode": true
     }
   },
   {
--- a/src/cms/forms/users/region_user_profile_form.py
+++ b/src/cms/forms/users/region_user_profile_form.py
@@ -15,7 +15,7 @@ class RegionUserProfileForm(UserProfileForm):
 
     class Meta:
         model = UserProfile
-        fields = ["organization"]
+        fields = ["organization", "expert_mode"]
 
     def __init__(self, data=None, instance=None):
 
--- a/src/cms/forms/users/user_profile_form.py
+++ b/src/cms/forms/users/user_profile_form.py
@@ -18,7 +18,7 @@ class UserProfileForm(forms.ModelForm):
 
     class Meta:
         model = UserProfile
-        fields = ["regions", "organization"]
+        fields = ["regions", "organization", "expert_mode"]
 
     def __init__(self, data=None, instance=None):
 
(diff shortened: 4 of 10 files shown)


timoludwig:
@melegiul Thanks for your PR!
I'm sorry, it turned out that our requirements for the expert mode were not well thought out and need some further discussions. Thus, I think we should conclude the discussion in Mattermost and issue #499 before proceeding this this PR. Sorry if this involves that you might have to discard parts of your work :confused:

melegiul:
> @melegiul Thanks for your PR!
> I'm sorry, it turned out that our requirements for the expert mode were not well thought out and need some further discussions. Thus, I think we should conclude the discussion in Mattermost and issue #499 before proceeding this this PR. Sorry if this involves that you might have to discard parts of your work confused

No problem, I just tried to follow the discussion on mattermost, but i'm not sure if get it right, so it's probably good to list some open question.

svenseeberg:
This is the result of the discussion (please correct me where I'm wrong):
* We consider translation experts a subgroup of expert users.
* We hide GUI elements relevant for translation from normal users, for example XLIFF import and export and filters.
* After we implemented all translation functionality (most important filtering), we iterate over the expert mode concept and could introduce an additional "translation expert role/mode". But this is not part of this issue/pull request.

melegiul:
> This is the result of the discussion (please correct me where I'm wrong):
> 
>     * We consider translation experts a subgroup of expert users.
> 
>     * We hide GUI elements relevant for (mass-)translation from normal users, for example XLIFF import and export and filters.
> 
>     * After we implemented all translation functionality (most important filtering), we iterate over the expert mode concept and could introduce an additional "translation expert role/mode". But this is not part of this issue/pull request.

Yes, that should clarify the open questions, so I won't consider the topic about the "translation expert role/model" for now. That means an user operating in "expert mode" can not access XLIFF import/export elements, so I leave these elements for the next iteration, when the "translation expert mode" is implemented.

timoludwig:
> That means an user operating in "expert mode" can not access XLIFF import/export elements, so I leave these elements for the next iteration, when the "translation expert mode" is implemented.

Just for reference: As we just discussed in our call, it is exactly the other way round: So users with "expert mode" should be able to access XLIFF functions and users without "expert mode" should not see this functions.
The next iteration is only optional if we determine that one mode is not enough and we need an additional setting.

src/cms/templates/users/admin/user.html
```diff
@@ -83,6 +83,11 @@ <h3 class="heading font-bold text-white">{% trans 'Extended Settings' %}</h3>
                     {% render_field user_form.is_active %}
                 </div>
 
+                <div class="py-4 border-b solid border-gray-200">
+                    <label for="id_expert_mode" class="font-bold mr-2 cursor-pointer">{% trans 'Expert mode' %}</label>
+                    {% render_field user_profile_form.expert_mode %}
```
timoludwig, line 88: I'd like to have an explanation here what features exactly are enabled with expert mode (and in the region user form as well).

timoludwig (commented):
Thanks! In general, everything works as expected! :+1:
I just have one request regarding an explanation in the user form and a small improvement for the context.

src/cms/views/events/event_mixin.py
```diff
@@ -23,3 +23,9 @@ class EventMixin(ContextMixin):
             "All translations of this event will also be deleted."
         ),
     }
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        if "user" in kwargs:
+            context["expert_mode"] = kwargs["user"].profile.expert_mode
```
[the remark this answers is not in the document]
melegiul: Well, then let's go with the `{% if request.user.profile.expert_mode %}` template, this seems to be more compact. I will remove the `get_context_data()` function entirely from both Mixins.

src/cms/templates/users/admin/user.html
melegiul: Alright, I added an explanation paragraph

timoludwig (commented):
Looks good :+1: 

Sorry, I missed this earlier on: Could you also add the field `"expert_mode": true` to the test root user? This way, the expert mode is enabled by default for any new dev environments...
https://github.com/Integreat/integreat-cms/blob/ca8fc99d5e24a444891292c0811b5b3e772b7d2f/src/cms/fixtures/test_data.json#L2-L19

melegiul:
> Looks good +1
> 
> Sorry, I missed this earlier on: Could you also add the field `"expert_mode": true` to the test root user profile? This way, the expert mode is enabled by default for any new dev environments...

No problem, expert mode is now enabled in test data root

timoludwig (approved):
Amazing, thanks! :rocket:

Issue #653, opened 2021-02-01
Region and user deletion not working


timoludwig:
### Describe the Bug

I couldn't find the exact commit which introduced this bug, but since a year or so the confirmation popup for deleting regions and users is broken.

### Steps to Reproduce

1. Go to regions
2. Click on a region
3. Scroll down and click "Delete region ..."

### Expected Behavior

There should be a confirmation popup which warns the user about which resources are actually deleted

### Actual Behavior

The region/user form gets submitted

### Additional Information

Fix this in the
- [ ] region form
- [ ] admin user form
- [ ] region user form

PR #654, opened 2021-02-02
Fix region and user deletion

timoludwig:
### Short description

Fix region and user deletion

### Proposed changes

- Fix old confirmation popups
- Fix `ProtectedError` when children are deleted before their parents
- Fix incorrect condition in `delete_region_user()`

### Resolved issues

Fixes: #653


src/cms/templates/users/region/user.html
```diff
@@ -88,13 +88,22 @@ <h3 class="heading font-bold text-white">{% trans 'Extended Settings' %}</h3>
...
-
-		    {% include "../confirmation_popups/delete_region_user.html" %}
+            <button title="{% trans 'Delete user' %}" class="confirmation-button w-full bg-red-500 hover:bg-red-600 text-white font-bold py-2 px-4 rounded"
+                    data-confirmation-title="{% trans 'Please confirm that you really want to delete this user' %}"
+                    data-confirmation-text="{% trans 'This cannot be reversed.' %}"
+                    data-confirmation-subject="{{ user_form.username.value }}"
+                    data-action="{% url 'delete_region_user' region_slug=region.slug user_id=user_form.instance.id %}">
+                <i data-feather="trash-2" class="inline-block mr-2"></i>
+                {% trans 'Delete this user' %}
+            </button>
```
Thomas125, line 98: If the user also has other regions, the button should state something like "Remove user from region" instead of "Delelte user" and the text in the confirmation popup should be adapted.

Thomas125 (commented):
On deletion of a region, no users are deleted. Just there membership to the region is deleted. Is this the inteded behavoiur? Or should all users with no regions left be deleted?
If first: Then the text in the confirmation popup should be adapted.
If second: Actually delete them.
But in both cases: The confirmation text should serve the information, that users that still have other regions, will not be deleted.

src/cms/templates/users/region/user.html
timoludwig: You're right, I added two versions of the button.

timoludwig:
> On deletion of a region, no users are deleted. Just there membership to the region is deleted. Is this the inteded behavoiur? Or should all users with no regions left be deleted?

Good point, I added an extra query to remove orphan users (users, who are neither superusers nor staff and don't have a region assigned anymore) after a region has been deleted.

> But in both cases: The confirmation text should serve the information, that users that still have other regions, will not be deleted.

OK, I adapted the text of the confirmation popup.

Thomas125 (approved):
Looks good to me.

PR #655, opened 2021-02-08
Improve form templates

timoludwig:
### Short description

This adds improvements for form templates

### Proposed changes

- Add `PlaceholderModelForm` for default field placeholders
  - Adds "Enter {field_name} here" as default placeholder
  - Only one translation for all placeholders instead of translating both the field name and the placeholder separately
- Use `verbose_name` and `help_text` in templates
  - Deduplications of translated strings
  - The same labels and help texts are now used for templates, Django admin and the Sphinx documentation
- Refactor inline JS code 
  - Deduplications of JS code
  - Enables compression of the tinymce init script

### Resolved issues

Just improvements, no issues resolved


ulliholtgrave (approved):
Checked it out and seems to work smoothly.

Issue #314, opened 2020-03-03
Add page owner


timoludwig:
A page should optionally have a page owner, which is used to determine a logo beneath the content.
We could either solve this
 - via a user, which is linked to an organization
 - or directly via an organization

Which solution do you prefer? I think directly choosing an organization is a good idea. This way, we could make sure all members of an organization can change this page without adding multiple page editors/publishers.

svenseeberg:
I vote for org as well.

Issue #591, opened 2020-11-18
Show region/organization logo in footer


timoludwig:
### Motivation

The region and/or organization logo should be shown in the footer to increase trustworthiness.

### Proposed Solution

~~Add the region logo whenever it exists and add the organization logo whenever the page/event/poi is not empty and was created by a user who belongs to an organization which has a logo.~~
- If the page has a responsible organization: show the organization logo
- Else: Show the region logo

Probably it makes sense to add these in the API processing, but also think about alternative solutions.
### Additional context

Optionally, we could add a setting for regions to turn this feature on or off.

This is blocked by #314 Add page owner

PR #656, opened 2021-02-10
Add responsible organization to Page model

timoludwig:
### Short description

Adding an organization as responsible for a page means all members of the organization can edit and publish the page.
Later on, this can be used to add the organization's logo in the footer of the page (see issue #591).
For now, I added this as another function available in the expert mode.

### Proposed changes

- Add field `organization` to `Page` model and `PageForm`
- Add predicate `is_in_responsible_organization()` to check permissions

### Resolved issues

Fixes: #314


melegiul (approved):
Awesome, I tested the page form for adding responsible organizations and it worked like expected.

But I found a problem concerning the user permissions. In my opinion this problem could also be addressed in another issue, so for now I can approve. To reproduce the problem please try the following:
1. Create a region user (e.g. augsburg user) without a role but as member of an organization.
2. Make this organization responsible for a page
3. Login as the new created user
4. Go to page menu
5. Then user gets permission denied message

The reason for that is that the organization member has no "cms.view_pages" permission, that means the user has no chance to reach the page for editing, because he cannot enter the page tree. See https://github.com/Integreat/integreat-cms/blob/8ae252809b6b0e1c496b549af5c872b19339bdf6/src/cms/views/pages/page_tree_view.py#L25
Quick fix for that would be to add a new role called "Organization member" and set the permission "cms.view_pages" for that role. In this case no further code adjustment is necessary except that it must be ensured that an organizations member has this role selected.

melegiul:
And I think we should also change the role.json file and add to "Redakteur" the number 9 to permissions array.
``
{
    "model": "auth.group",
    "pk": 2,
    "fields": {
      "name": "Redakteur",
      "permissions": [
        10,
        11,
        12,
        13
      ]
    }
  },
``
### Currently
currently
### With permission 9 added to list
withViewPerm

timoludwig:
> The reason for that is that the organization member has no "cms.view_pages" permission, that means the user has no chance to reach the page for editing, because he cannot enter the page tree. See
> 
> https://github.com/Integreat/integreat-cms/blob/8ae252809b6b0e1c496b549af5c872b19339bdf6/src/cms/views/pages/page_tree_view.py#L25
> 
> Quick fix for that would be to add a new role called "Organization member" and set the permission "cms.view_pages" for that role. In this case no further code adjustment is necessary except that it must be ensured that an organizations member has this role selected.

Thanks for your review! I solved it in the following way: The `view_pages` permission is now rule-based, so every user who can at least edit/publish one page, can view the page tree.

melegiul:
> Thanks for your review! I solved it in the following way: The `view_pages` permission is now rule-based, so every user who can at least edit/publish one page, can view the page tree.

Very cool, that solves the problem about the missing permission:tada:
From my side this PR is ready now.

Issue #587, opened 2020-11-18
Insert mailto: and tel: links automatically


timoludwig:
### Motivation

Whenever email addresses and phone numbers are used in the content, we should replace them with the according `mailto:` and `tel:` links.

### Proposed Solution

We could e.g. replace such strings in the clean-method of the content forms.

ulliholtgrave:
The mailto feature can already be used by adding the autolink-Plugin (https://www.tiny.cloud/docs/plugins/opensource/autolink/) to the tinymce, but tel: is not supported. Therefore I am proposing extending the Plugin by the following RegEx to detect phone numbers:
/+?[0-9][0-9()-\s+]{4,20}[0-9]/

svenseeberg:
I think we can make it much simpler: `([0-9\.\-\ \,\+\(\)]{3,30})`

We need to support numbers like:
* 110
* 1234
* +49 (0) 123 4567890
* 0049 123 4567890
* 0123 4567-890
* 01234567890
etc

IMHO there are no patterns we can rely on. There is only the limited character set that we can use to identify telephone numbers and the length. To my knowledge (and quick scan of https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains) there are no TLDs with numbers only. That means we can safely assume that no domain can be numbers only, and therefore also no e-mail address.

timoludwig:
> I think we can make it much simpler: `([0-9\.\-\ \,\+\(\)]{3,30})`

But we also have to think about false positives here. Do we really want to turn nearly every number into a phone number link? This would also affect e.g. years/dates (2021 or  2021-02-09 or 09.02.2021), normal numbers ("Almost 300,000 people from around 170 different nations"), typical float numbers (82.6), postal codes (86150 Augsburg), etc.
I'm afraid we clutter the whole content with phone links if we're too inclusive here.
Maybe a more complex regex is better than too many phone links? Or we define short phone numbers (110, 112) in a global setting and only match regular phone numbers with the regex? Or we ask the user every time we automatically convert a number into a phone number link?

svenseeberg:
Okay, sorry. I understood that this is about recognizing user defined links as telephone numbers. In this case I think this issue should move to the additional features milestone.

PR #657, opened 2021-02-11
Setup autolink & custom_plugin

ulliholtgrave:
### Short description
This PR actives the inbuilt autolink plugin for the automated detection of mail addresses and enforces HTTPS for all links in our content. The found links are directly converted into a Link-Object and mail addresses are converted into :mailto-links.

Additionally, I created a custom tinymce plugin to detect phone numbers as well. Although, this isn't ready yet.

### Proposed changes
- Activate autolink plugin.
- Enforce HTTPS for all links (to be discussed).
- Add custom plugin to detect and modify phone numbers.

### Resolved issues

Fixes: #587

### DIFF
--- a/src/cms/static/js/tinymce_plugins/autolink_phone/index.js
+++ b/src/cms/static/js/tinymce_plugins/autolink_phone/index.js
@@ -1,0 +1,3 @@
+// Entry point for Tinymce hook.
+// Author: Jan-Ulrich Holtgrave (holtgrave@integreat-app.de)
+require('./plugin.js');
--- a/src/cms/static/js/tinymce_plugins/autolink_phone/plugin.js
+++ b/src/cms/static/js/tinymce_plugins/autolink_phone/plugin.js
@@ -1,0 +1,166 @@
+/**
+ * Copyright (c) Tiny Technologies, Inc. All rights reserved.
+ * Licensed under the LGPL or a commercial license.
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ *
+ * Version: 5.7.0 (2021-02-10)
+ */
+(function () {
+    'use strict';
+
+    var global = tinymce.util.Tools.resolve('tinymce.PluginManager');
+
+    var global$1 = tinymce.util.Tools.resolve('tinymce.Env');
+
+    var rangeEqualsDelimiterOrSpace = function (rangeString, delimiter) {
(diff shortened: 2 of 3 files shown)


src/cms/static/js/tinymce_plugins/autolink_phone/index.js
```diff
@@ -0,0 +1,3 @@
+// Entry point for Tinymce hook.
+// Author: Jan-Ulrich Holtgrave (holtgrave@integreat-app.de)
+require('./plugin.js');
```
timoludwig, line 3: This npm-style syntax normally doesn't work in browsers, right? Or does tinymce perform some magic here?
Or can this file just be omitted, since only `plugin.js` is listed in the `external_plugins`?

src/cms/static/js/tinymce_plugins/autolink_phone/plugin.js
```diff
@@ -0,0 +1,166 @@
+/**
+ * Copyright (c) Tiny Technologies, Inc. All rights reserved.
+ * Licensed under the LGPL or a commercial license.
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ *
+ * Version: 5.7.0 (2021-02-10)
+ */
```
timoludwig, line 8: Do you have a source for this script? The copyright notice suggests that this is an official plugin? Is there a possibility to install it somehow without having to track it in the vcs?

src/cms/static/js/tinymce_plugins/autolink_phone/index.js
ulliholtgrave: I guess we might remove this as we are not really publishing the plugin, but Tinymce does recommend that a plugin should include at least index.js, plugin.js and plugin.min.js

src/cms/static/js/tinymce_plugins/autolink_phone/plugin.js
ulliholtgrave: That's the official autolink plugin, but as it doesn't support the phone-detection I came up with a own version, which only searches for phone numbers. I don't think, that we can expand this by a patch, as we need to provide some serious customization, which includes:
- Adding additional RegEx
- Adding custom Links
- (Most important) We need to change the window-offset, which is used in multiple spots in the plugin. We do need to change this to identify numbers with only 3 digits.

Issue #503, opened 2020-09-25
Clarify hierarchy in page tree


timoludwig:
### Motivation

At the moment, it is not completely clear how pages are structured in the page tree and how the page order can be changed.

### Proposed Solution

Label the first page tree column with "hierarchy" and add a message box or tooltip to point out the possibility of drag & drop in the page tree.

### Alternatives

If you have better ideas how to achieve this goal, give them a try!

PR #658, opened 2021-02-11
Improve hierarchy in page tree

timoludwig:
### Short description

This adds improvements to the hierarchy of the page tree.

### Proposed changes

- Clarify hierarchy in page tree
- Enable collapsing of subpages

### Resolved issues

Fixes: #503

Screenshot_2021-02-12 Integreat CMS

### DIFF
--- a/src/cms/forms/pages/page_form.py
+++ b/src/cms/forms/pages/page_form.py
@@ -7,6 +7,7 @@ from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Permission
 from django.db.models import Q
 from django.urls import reverse
+from django.utils.text import capfirst
 from django.utils.translation import ugettext as _, get_language
 from django.utils.safestring import mark_safe
 
@@ -179,7 +180,10 @@ class PageForm(forms.ModelForm):
     """
 
     parent = ParentField(
-        queryset=Page.objects.all(), required=False, widget=ParentFieldWidget()
+        queryset=Page.objects.all(),
+        required=False,
+        widget=ParentFieldWidget(),
+        label=capfirst(Page._meta.get_field("parent").verbose_name),
     )
     related_page = forms.ModelChoiceField(
         queryset=Page.objects.all(), required=False, widget=forms.HiddenInput()
--- a/src/cms/locale/de/LC_MESSAGES/django.po
+++ b/src/cms/locale/de/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2021-02-12 10:22+0000\n"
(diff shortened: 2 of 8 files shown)


melegiul (approved):
Awesome, works like a charm.:tada:
There just one topic rather for discussion. I found that at level 3 or greater the row height of a page grows, but whether this is a problem depends on how many hierarchy levels are expected. Maybe we could increase the hierarchy columns max-width a bit. 
page-hierarchie

timoludwig:
Even with many levels, this looks fine for me:
Screenshot_2021-02-19 Integreat CMS

Can you check whether your dependencies (especially tailwindcss) are up to date? The css class `whitespace-nowrap` was added recently and should in theory make it impossible for the two icons to wrap into a new line.

ulliholtgrave:
I couldn't reproduce it, too. I tried it with multiple levels, on firefox & chrome, on two different screens and with different resolutions. Still seems to work.
grafik

melegiul:
> Even with many levels, this looks fine for me:
> Screenshot_2021-02-19 Integreat CMS
> 
> Can you check whether your dependencies (especially tailwindcss) are up to date? The css class `whitespace-nowrap` was added recently and should in theory make it impossible for the two icons to wrap into a new line.

Ok, then this problem just exists on my side, I will check my dependencies. It was my fault, sorry for that.

timoludwig:
> Ok, then this problem just exists on my side, I will check my dependencies. It was my fault, sorry for that.

No worries, since we want to support all more or less modern browsers, it would be still an issue if the problem persisted on your system.

melegiul:
> No worries, since we want to support all more or less modern browsers, it would be still an issue if the problem persisted on your system.

So I found that the problem still exists on Firefox (version 85.0.1 which should be the latest), but I just tried to reproduce on Chrome and there all is fine, so this might be a problem related to firefox.

melegiul:
I also checked my tailwind installation and `npm list` gave me, that there is version `2.0.3` installed. Which one is yours?

melegiul:
But I'm still wondering why on @ulliholtgrave firefox everything is fine also.

timoludwig:
> I also checked my tailwind installation and `npm list` gave me, that there is version `2.0.3` installed. Which one is yours?

I've also installed tailwind version `2.0.3` as defined in package.json: https://github.com/Integreat/integreat-cms/blob/63dccd6fbe99a3857681d82d211557406e47d0c6/package.json#L39
And my Firefox is just a minor version ahead (85.0.2), so this shouldn't make the difference...

Maybe it's a caching issue? Can you clear your cache and reload?

melegiul:
> I've also installed tailwind version `2.0.3` as defined in package.json:
> 
> https://github.com/Integreat/integreat-cms/blob/63dccd6fbe99a3857681d82d211557406e47d0c6/package.json#L39

Yes, I cleared the cache and now it works on my firefox, too. Thanke you:tada:

PR #661, opened 2021-02-13
No external links in content

### LINKED ISSUE #604: Don't open external links in new tab
### Motivation

To achieve a consistent behavior of external links in the content, we should prevent that content creators mark links as "open in new tab" or use `target="_blank"` in the html source code

### Proposed Solution

- [ ] Remove the option to open links in a new tab from the tinymce link settings
- [ ] Validate all content fields that they do not contain links which are opened in a new tab (similar to the check for `<h1>`-tags) 

### Additional Context

Possible regex:
```
r = re.compile(r"target=[\"']?_blank[\"']?")
urls = r.findall(text)
if urls:
    raise ValidationError(
        _("Use of external links is not allowed!"),
        code="no-external-links",
    )
```

AminAlAit:
### Short description
This function should serve in eliminating all external links inputted by the user. 

### Proposed changes
The added function, clean_text2, relies on regex to match the pattern in text.

### Resolved issues
Fixes #604 (Don't open external links in new tab)


src/cms/forms/pages/page_translation_form.py
```diff
@@ -137,8 +138,24 @@ def clean_text(self):
 
         if "<h1>" in text:
             raise ValidationError(
-                _("Use of Heading 1 style not allowed."),
+                _("Use of Heading 1 style is not allowed."),
                 code="no-heading-1",
             )
 
+    def clean_text2(self):
```
timoludwig, line 145: Quote from the [Django Documentation](https://docs.djangoproject.com/en/2.2/ref/forms/validation/):
"The clean_\<fieldname\>() method is called on a form subclass – where \<fieldname\> is replaced with the name of the form field attribute."

In other words, please do not create a second method `clean_text2()`, but put all validation code for the `text` field in the `clean_text()`-method. Django will call them magically when `form.is_valid()` is executed. It's no big problem that the code after the first `ValidationError` isn't run - it will only happen very rarely that the text field contains both `<h1>`-tags and external links, and in this case it's fine to fix one error at a time (e.g. after the error "Use of Heading 1 style is not allowed." the user removes the `<h1>`-tags, tries to submit the form again, gets the error "Use of external links is not allowed!" and removes the external links).

src/cms/forms/pages/page_translation_form.py
```diff
@@ -137,8 +138,24 @@ def clean_text(self):
...
+        text = self.data["text"]
+
+        # r'href=[\'"]?([^\'" >]+)'
+        # r(?<=href=").*?(?=")+.*?(?<=target=")
+        r = re.compile(r"(\B|\b)<a+.+\/a>(\B|\b)")
```
timoludwig, line 150:
```suggestion
        r = re.compile(r"target=[\"']?_blank[\"']?")
```
I'd suggest the following regex. We only want to prevent `target="_blank"` (with all different possibilities regarding quotation marks). It should be allowed to use normal links with `<a>`-tags.

src/cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -365,10 +369,15 @@ msgstr ""
...
 
+<<<<<<< HEAD
+#: forms/events/event_translation_form.py:100
+#: forms/pois/poi_translation_form.py:103
+=======
 #: forms/events/event_translation_form.py:135
 #: forms/imprint/imprint_translation_form.py:122
 #: forms/pages/page_translation_form.py:140
 #: forms/pois/poi_translation_form.py:149
+>>>>>>> 2682b9718108d6151c2b25194ff78b399038cc8a
```
timoludwig, line 380: The sections marked with `<<<<<<< HEAD`, `=======` and `>>>>>>>` denote [git merge conflicts](https://www.atlassian.com/git/tutorials/using-branches/merge-conflicts). Please always resolve conflicts before committing changes.
For the translation file, we luckily have the dev tool `./dev-tools/resolve_translation_conflicts.sh` which can resolve these conflicts automatically.

timoludwig (commented):
Thanks for your PR! :rocket: 

Apart from my comments below, a general note: The submodule `integreat-cms` inside this repository doesn't quite make sense. Please delete that file again. Do you know when this file appeared?

Screenshot_2021-02-13 Feature external links by AminAlAit · Pull Request #661 · Integreat integreat-cms

timoludwig (commented):
@AminAlAit Are you willing to complete this PR, or should we take over?

Issue #652, opened 2021-02-01
Convert datetime according to user's timezone


timoludwig:
### Motivation

Our server's timezone (currently set to "UTC") does not necessarily correspond to the user's timezone. Thus, we have to convert every `datetime` input & output in the templates to local time.

### Proposed Solution

First problem: How to get the user's timezone?

  - ~Option A: Deriving from the language setting (bad, because it is ambiguous e.g. for English)~
  - Option B: Add another field to the user settings to store the timezone
  - Option C: Use geolocation services to automatically derive the user's timezone, e.g.:
    - [django-tz-detect](https://pypi.org/project/django-tz-detect/)
    - [django-easy-timezones-redux](https://pypi.org/project/django-easy-timezones-redux/)
    - [django-visitor-information-middleware](https://pypi.org/project/django-visitor-information-middleware/)

Second problem: How to convert the datetime?

  - https://docs.djangoproject.com/en/2.2/topics/i18n/timezones/
  - https://tommikaikkonen.github.io/timezones/

### Alternatives

If we make the assumption that all our users are located in the Central European Time (CET) timezone, we just can set the server time to CET and don't need further adjustments. However, this will not suffice in the long term.

### Additional Context

We also have to think about what to do with events:

  - Option A: Store event time in the server's timezone
    - Convert input value before writing it into the database
    - Convert datetime whenever it gets printed ot delivered via the API
  - Option B: Store event time in the event's local time
    - Do not convert input value on saving
    - Do not convert datetime on output
    - Can we assume creator's timezone == event's timezone?

Thomas125:
Is there always a canonical event's local time?
For example, regarding a zoom-meeting of some german people with some australian people. What is the event's local time?
Is this relevant for us?

melegiul:
I would also prefer Option B, as the django docs suggest [selecting-current-timezone](https://docs.djangoproject.com/en/2.2/topics/i18n/timezones/#selecting-the-current-time-zone), it's common practice to save a local timezone to each user profile. That will keep our timestamps internally as "UTC", but formats the output according to users local timezone. However this does not address the problem about events timezones. Per default events should inherit their timezone from the creators user profile, maybe extended with some further customization options, e.g. "Select a custom timezone for this event"-button.

Thomas125:
Is it possible to use the timezone of the end user's computer?

timoludwig:
>* Is there always a canonical event's local time?
[...]

Issue #663, opened 2021-02-17
Timezone aware events


Thomas125:
### Motivation
Currently, the date- and time-information of events is naive (i.e. has no timezone information). It should be aware.
This is independent of #662 since events should be stored with timezone-information even in a single-timezone-situation (regarding migration, change of timezone etc.)

### Proposed Solution
Store date and time of events in a datetime-attribute in the models, instead of one date- and one time-attribute.

### Alternatives
None (I think)

### Additional Context
#652 contains information about how timezone-issues are resolved in django.

timoludwig:
Regarding the problem on all-day-events:

I think, the best option might be to extend the event model with an additional `CharField` which contains the timezone of the event. For now, the default value of this field should be read from the following setting: https://github.com/Integreat/integreat-cms/blob/0d623d97bfd2bc91765fbc3e93c4f37a7f15b0dc/src/backend/settings.py#L571
In the future, we might make this configurable in the form or derive it from the creator's local time.
Then, in the event form's `clean()` method, we could convert the UTC date to the local time of the `timezone` field and check whether the local time is `datetime.time.min` resp. `datetime.time.max`...

@Thomas125 What do you think about this? If you think this isn't a good solution, I'm also fine with adding a boolean field `all_day` to the event model.

timoludwig:
This was solved in #1642 :tada:

Issue #662, opened 2021-02-17
Add support for multiple end-user-timezones


Thomas125:
### Motivation
#652 adds a setting for the `current time zone` that is used for rendering.
This is hardcoded and systemwide. So, dates and times are always rendered with respect to that timezone.

### Proposed Solution
Eliminate the setting.
Instead, implement a way, that the `current time zone` is each user's actual time zone.

### Alternatives
This is issue is quite optional. Since in most cases, all users are in the same time zone, the solution of #652 is already sufficient for most cases. For discussion about this, see https://github.com/Integreat/integreat-cms/issues/652#issuecomment-775275083 and https://github.com/Integreat/integreat-cms/issues/652#issuecomment-775289621

### Additional Context
#652 contains information about how timezone-issues are resolved in django.

MizukiTemma:
This issue seems to have become outdated. In the current implementation, a time zone can be selected for each region and all time related information (for examp,e start and end time of events) are shown in the selected time zone.

A possible extension would be to show all time infos in the actual time zone of user: if user sees an event of a region in Germany and is currently in Japan, start/end time of the event will be shown in Japanese time. This would make sense for virtual events, but is probably no use for present events.
@laravoelk @hauf-toni Do we need this?

laravoelk:
In my opinion we do not need it. :-)

PR #664, opened 2021-02-17
Add current time zone

Thomas125:
### Short description
Adds a setting for `current time zone` that is used for rendering.

### Proposed changes
Add a setting `CURRENT_TIME_ZONE`
Add a middleware-function that sets the current time zone according to this setting

### Resolved issues

Fixes: #652

### Additonal
For further timezone handling, #662 and #663 have been created.

### DIFF
--- a/src/backend/settings.py
+++ b/src/backend/settings.py
@@ -144,6 +144,7 @@ MIDDLEWARE = [
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
+    "cms.middleware.timezone_middleware.TimezoneMiddleware",
 ]
 
 #: Default URL dispatcher (see :setting:`django:ROOT_URLCONF`)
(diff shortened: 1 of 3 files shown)


src/backend/settings.py
```diff
@@ -407,6 +408,10 @@
 #: (see :setting:`django:TIME_ZONE` and :doc:`topics/i18n/index`)
 TIME_ZONE = "UTC"
 
+#: A string representing the time zone that is used for rendering
+#: (see :setting:`django:CURRENT_TIME_ZONE` and :doc:`topics/i18n/index`)
```
timoludwig, line 412: Just for your information: The `:setting:` text role works only for official Django settings, not for our custom ones.
If you want to link to this custom setting from anywhere else, use:
```
:attr:`~backend.settings.CURRENT_TIME_ZONE`
```

timoludwig (approved):
Very clean and elegant solution! :+1:

timoludwig:
For the future: You don't need to merge the develop branch into your feature branch before merging your branch back into develop. This just clutters the git history and as long as there are no conflicts, we don't have to fear problems (aside from some edge cases)...

Issue #600, opened 2020-11-25
Improve feedback list


timoludwig:
### Motivation

The feedback list works, but has room for improvement. This issue should document ideas on how the view could be improved

### Proposed Solution

- Find a way to generalize the different types of feedback so only one loop is required for all feedback entries
- Add filters and sorting to the list
- Add pagination
- Show user who marked feedback as read

PR #669, opened 2021-02-19
Refactor Feedback

timoludwig:
### Short description

This adds all kinds of improvements for the feedback functionality and some improvements under the hood.
Also, this should enable the feedback list to get pagination (see #660).

### Proposed changes

- Generalize filter form scripts
- Generalize bulk actions
- Refactor feedback models, views and templates 
- Add bulk actions to feedback list 
- Add filters to feedback list 

### Resolved issues

Fixes: #600

Screenshot_2021-02-19 Integreat CMS


ulliholtgrave:
Unfortunately I won't really be able to provide a completet feedback, but I found a few UI problems/bugs. 
Feedback-Bugs
Firstly: The table has a strange border on the left side. As this is different from the other tables we should remove this or apply this styling to the other tables.
Secondly, you gave the negative-vote a left-padding of .25em. If there are only negative votes this results in placeholder. I propose, that you add a if-clause checking whether there are any positive votes and apply either pl-r or pl-l.

timoludwig:
> Firstly: The table has a strange border on the left side. As this is different from the other tables we should remove this or apply this styling to the other tables.

I liked the left border on unread feedback in WordPress, so I tried to recreate it.
Screenshot_2021-02-22 Kommentare (9) ‹ Testumgebung — WordPress
Unfortunately, I couldn't get the checkboxes of the read entries lined up without the border on the gray rows, so I just added the border on all rows. I see two options now:

1) Completely remove the left border
2) Try further to remove the border on gray rows and only show it on the highlighted rows. I mean, somehow it has to work :grin: 

What would you prefer?

ulliholtgrave:
I guess, what bothers me the most is the border at the table header. The rest is fine for me :)

timoludwig:
I added the following changes to this PR:
- Make left border bold only when feedback is unread
- Align positive and negative feedback, even if only one of both is given
- Add pagination for feedback
- Add feedback for POIs and the map incl. new API endpoints
- Remove the possibility to add the base feedback objects in admin.py and add the possibility to create all submodels
- Change feedback `read` flag to `read_by` field containing the user who marked the feedback as read

ulliholtgrave (approved):
Looks reasonable to me :) We might wanna do some small improvements to the UI of the feedback lists, but this is ready to ship for now :+1:

timoludwig:
I added some more minor improvements:
- Hover effect in the feedback list
- The possibility to select/deselect the checkbox on a row by clicking anything in the row which isn't a link
- The displayed object name now always uses the name of the current backend language if available and the region's default language as fallback
- The rating is not represented with "Positive" or "Negative" anymore, but the thumbs icons directly

@ulliholtgrave if you are fine with these changes, I will merge this PR.

Issue #674, opened 2021-02-20
POI field in event form not working


timoludwig:
### Describe the Bug

In commit 1da1ab033777bd94a172368b36845735ad68a875, I forgot to include `event_query_pois.js` in the event form.
This results in the POI field not working anymore.

PR #675, opened 2021-02-20
Fix POI field in event form

timoludwig:
### Short description

Fix the POI field in event form

### Proposed changes

- Include `event_query_pois.js` in the event form

### Resolved issues

Fixes: #674


ulliholtgrave (approved):
Resolves the issue.

Issue #673, opened 2021-02-19
Mirrored page not working on page creation form


timoludwig:
### Describe the Bug

When a new page is created, it is not possible to fill in the mirrored page field

### Steps to Reproduce

1. Go to a region
2. Create page
3. Select region for mirrored page
4. See "Internal Server Error"

### Expected Behavior

The select of valid page choices should appear

### Actual Behavior

The ajax call returns `Internal Server Error`

### Additional Information

```
ValueError at /ajax/render/mirrored_page_field/

invalid literal for int() with base 10: 'None'

file: /home/timo/job/integreat/integreat-cms/src/cms/views/pages/page_actions.py in render_mirrored_page_field
line: 850
```

PR #690, opened 2021-02-23
Fixes page mirroring bug

melegiul:
### Short description

Fixes page mirroring bug, which is caused by trying to set a mirrored page, when a completely new page is created.

### Proposed changes

- When a region for page mirroring is selected, an ajax request is
  triggered, which tries to retrieve the current page from database.
- So when there is a completely new page created, which has no entry in the
  database yet, this ajax request fails.
- That means the page mirroring field in the page form should be hidden
  for newly created pages.

### Alternatives
The proposed solution is the easiest, but if you wish, there would also be an ability to adjust the `render_mirrored_page_field` view, so that it can handle pages without a database entry:
https://github.com/Integreat/integreat-cms/blob/420ecee0f727fa4097d5f241f65d75a96a759dae/src/cms/views/pages/page_actions.py#L850

### Resolved issues

Fixes: #673

### DIFF
--- a/src/cms/templates/pages/page_form.html
+++ b/src/cms/templates/pages/page_form.html
@@ -245,7 +245,7 @@
                                 <img src="{% static 'svg/select-down-arrow.svg' %}" class="fill-current h-4 w-4" />
                             </div>
                         </div>
-                        <div class="relative my-2 {% if not page.mirrored_page %}hidden{% endif %}" id="mirrored_page_div" data-url="{% url 'render_mirrored_page_field' %}?page_id={{ page_form.instance.id }}&region_id=">
+                        <div class="relative my-2 {% if not page.mirrored_page %}hidden{% endif %}" id="mirrored_page_div" data-url="{% url 'render_mirrored_page_field' %}?{% if page_form.instance.id %}page_id={{ page_form.instance.id }}&{% endif %}region_id=">
                             {% include "pages/_mirrored_page_field.html" %}
                         </div>
                         <div class="relative my-2 pb-2 {% if not page.mirrored_page %}hidden{% endif %}" id="mirrored_page_first_div">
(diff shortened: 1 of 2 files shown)


timoludwig:
I think it would be cool to enable the mirrored page field also on the new page form.
As long as there are no technical reasons why it's impossible, I think it makes sense (e.g. some pages may be created only to make use of the mirrored page field - people may get confused that the field is missing when they create a page)...

melegiul:
> I think it would be cool to enable the mirrored page field also on the new page form.
> As long as there are no technical reasons why it's impossible, I think it makes sense (e.g. some pages may be created only to make use of the mirrored page field - people may get confused that the field is missing when they create a page)...

Alright, I've added an extra if condition, that should do the trick

timoludwig (commented):
Thanks! This definitely works, but it's kind of strange to send the string "None" in the first place (which isn't your fault), so I'd prefer to fix this directly in the template:
https://github.com/Integreat/integreat-cms/blob/420ecee0f727fa4097d5f241f65d75a96a759dae/src/cms/templates/pages/page_form.html#L248
Change this line to:
```
<div class="relative my-2 {% if not page.mirrored_page %}hidden{% endif %}" id="mirrored_page_div" data-url="{% url 'render_mirrored_page_field' %}?{% if page_form.instance.id %}page_id={{ page_form.instance.id }}&{% endif %}region_id=">
``` 

This way, the value of the `page_id` will be `None` directly instead of the string `"None"`...

timoludwig (approved):
Thanks! :+1:

melegiul:
> This way, the value of the `page_id` will be `None` directly instead of the string `"None"`...

Yes, that `None` string was the root of the problem, makes sense to update the template.

Issue #624, opened 2021-01-09
Add table sorting


timoludwig:
### Motivation

At the moment, only the event list can be sorted.

### Proposed Solution

Add the same [tablesort](https://www.npmjs.com/package/tablesort) functionality to all our tables:

- [x] Events
- [x] Archived Events
- [x] POIs
- [x] Archived POIs
- [ ] Regions
- [ ] Languages
- [ ] Admin Users
- [ ] Region Users
- [ ] Organizations
- [ ] Offer Templates
- [ ] Admin Feedback
- [ ] Region Feedback
- [ ] Push Notifications

ulliholtgrave:
No longer needed. See #885

PR #707, opened 2021-02-24
Improve POI list

mael-fosso:
### Short description
changing some css classes of html elements.

### Proposed changes
Apart from that, the different lists (especially for pages, events and pois) aren't rendered consistently. For example, they use different shades of grey for the table borders. It should be investigated in which points they really differ (apart from the given example) as well as those differences should be eliminated in order to have a consistent UI experience.

-
-
-

### Resolved issues

This fixes the POI part of #292 and #624

### DIFF
--- a/src/cms/templates/pois/poi_list.html
+++ b/src/cms/templates/pois/poi_list.html
@@ -37,32 +37,32 @@
 </div>
 
 <div class="table-listing">
-    <table class="w-full mt-4 rounded border border-solid border-gray-400 shadow bg-white table-fixed">
+    <table data-enable-table-sort class="sort w-full mt-4 rounded border border-gray-200 shadow bg-white table-fixed">
         <thead>
-            <tr class="border-b border-solid border-gray-400">
+            <tr class="border-b border-gray-200">
                 <th class="text-sm text-left uppercase py-3 pl-4 pr-2 min">{% trans 'ID' %}</th>
                 <th class="text-sm text-left uppercase py-3 pr-2 min">{% trans 'Version' %}</th>
-                <th class="text-sm text-left uppercase py-3">{% trans 'Status' %}</th>
-                <th class="text-sm text-left uppercase py-3">{% trans 'Title in' %} {{ language.translated_name }}</th>
+                <th class="text-sm text-left uppercase py-3 pr-2">{% trans 'Status' %}</th>
+                <th class="text-sm text-left uppercase py-3 pr-2" data-sort-default>{% trans 'Title in' %} {{ language.translated_name }}</th>
                 {% get_current_language as LANGUAGE_CODE %}
                 {% get_language LANGUAGE_CODE as backend_language %}
                 {% if backend_language and backend_language != language %}
-                    <th class="text-sm text-left uppercase py-3 px-2">{% trans 'Title in' %} {{ backend_language.translated_name }}</th>
+                    <th class="text-sm text-left uppercase py-3 pr-2">{% trans 'Title in' %} {{ backend_language.translated_name }}</th>
                 {% endif %}
-                <th class="text-sm text-left uppercase py-3 px-2">
-                    <div class="lang-grid flags" style="white-space: nowrap;">
-	                    {% for lang in languages %}
-		                    <a href="{% url 'pois' region_slug=region.slug language_slug=lang.slug %}">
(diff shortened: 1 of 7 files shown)


src/cms/templates/regions/region_list.html
```diff
@@ -22,9 +22,9 @@ <h2>
 </div>
 
 <div class="table-listing">
-    <table class="w-full mt-4 rounded border border-solid border-gray-200 shadow bg-white">
+    <table class="w-full mt-4 rounded border border-solid border-gray-400 shadow bg-white">
```
timoludwig, line 25: I don't like this dark border very much. I would rather suggest to make the border of the POI list more light instead of all other list dark.

src/cms/templates/regions/region_list.html
```diff
@@ -22,9 +22,9 @@ <h2>
...
         <thead>
-            <tr class="border-b border-solid border-gray-200">
+            <tr class="border-b border-solid border-gray-400">
```
timoludwig, line 27: This also applies for this border.

src/cms/templates/pois/poi_list.html
```diff
@@ -37,13 +50,13 @@ <h2 class="heading font-normal mb-2">{% trans 'Locations on map' %}</h2>
...
 <div class="table-listing">
-    <table class="w-full mt-4 rounded border border-solid border-gray-400 shadow bg-white table-fixed">
+    <table id="pois-list" class=" sort w-full mt-4 rounded border border-solid border-gray-400 shadow bg-white table-fixed">
         <thead>
             <tr class="border-b border-solid border-gray-400">
                 <th class="text-sm text-left uppercase py-3 pl-4 pr-2 min">{% trans 'ID' %}</th>
                 <th class="text-sm text-left uppercase py-3 pr-2 min">{% trans 'Version' %}</th>
-                <th class="text-sm text-left uppercase py-3">{% trans 'Status' %}</th>
-                <th class="text-sm text-left uppercase py-3">{% trans 'Title in' %} {{ language.translated_name }}</th>
+                <th class="text-sm text-left uppercase py-3 px-2">{% trans 'Status' %}</th>
```
timoludwig, line 58:
```suggestion
                <th class="text-sm text-left uppercase py-3 pr-2">{% trans 'Status' %}</th>
```
This is going into the wrong direction. Please do not add the class `px-2` to all classes, but rather `pr-2`. The reason is that we want to align all table fields to their left edge. If both the table head and the table cell have the same padding, they are indeed aligned, but together with the padding of the previous cells, we have a total padding of 4 between the cells and this is too much. Just add a padding-right of 2 to all cells (`pr-2`) and remove the `px-2` where you find it.

src/cms/templates/pois/poi_list.html
```diff
@@ -52,20 +65,20 @@ <h2 class="heading font-normal mb-2">{% trans 'Locations on map' %}</h2>
...
                 </th>
-                <th class="text-sm text-left uppercase py-3">{% trans 'Street' %}</th>
-                <th class="text-sm text-left uppercase py-3">{% trans 'Postal Code' %}</th>
-	            <th class="text-sm text-left uppercase py-3">{% trans 'City' %}</th>
-	            <th class="text-sm text-left uppercase py-3">{% trans 'Country' %}</th>
-                <th class="text-sm text-right uppercase py-3 pl-2 pr-4 min">{% trans 'Options' %}</th>
+                <th class="text-sm text-left uppercase py-3 px-2">{% trans 'Street' %}</th>
+                <th class="text-sm text-left uppercase py-3 px-2">{% trans 'Postal Code' %}</th>
+	            <th class="text-sm text-left uppercase py-3 px-2">{% trans 'City' %}</th>
+	            <th class="text-sm text-left uppercase py-3 px-2">{% trans 'Country' %}</th>
```
timoludwig, line 80: Here as well, please use `pr-2` instead of `px-2`.

timoludwig (commented):
Thanks! The sorting functionality is cool! :rocket: 

As for the consistent list representations: Please have a look at my comments, I hope you find them helpful.

timoludwig (approved):
I reverted your changes to the `region_list.html` because they did not make much sense and had nothing to do with this PR.
Apart from that, I added some adjustments to fix the tablesort mechanism with the new typescript bundling. Now I think the PR is ready to merge! :+1:

Issue #693, opened 2021-02-23
Improve event list


timoludwig:
### Proposed Solution

- Remove ID and version columns
- Fix icon of edit action
- Make text color of filter button always white

PR #708, opened 2021-02-24
Improve event list

mael-fosso:
### Short description
 Edit  Event Page
### Proposed changes

    -Remove ID and version columns
    -Fix icon of edit action
    -Make text color of filter button always white

### Resolved issues

Fixes: #693

### DIFF
--- a/src/cms/locale/de/LC_MESSAGES/django.po
+++ b/src/cms/locale/de/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2021-03-01 00:04+0000\n"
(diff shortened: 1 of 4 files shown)


timoludwig (approved):
Thanks you. this looks good! :+1:

For the future: If you are working on multiple issues simultaneously (like here with PR #707 and #708), start each feature branch on the develop branch. Otherwise, it is hard to recognize which changes belong to which PR and it's not possible to merge one PR while waiting for the other.
In this case, your commit 53433dcd71490a51be092676a44b757502308f17 of PR #707 was also included in this PR. But since this commit isn't ready yet, I removed it from this branch.
Other than that, I also removed the dark border around the page tree because it has nothing to do with this PR and also see [my comment on the other PR](https://github.com/Integreat/integreat-cms/pull/707#discussion_r582345574) about the dark border on the region list.

But all changes to the event list look good, this PR is ready to merge! :tada:

Issue #706, opened 2021-02-24
Pages API not working


ulliholtgrave:
### Describe the Bug
We run into an error, when we call any of the pages API (see screenshot).

This is probably broken since the adding of the page mirroring feature.

### Additional Information
![grafik](https://user-images.githubusercontent.com/15608678/109014324-522ffe80-76b4-11eb-8e33-dee02bf6a250.png)

PR #709, opened 2021-02-24
Fix mirrored text in page API

timoludwig:
### Short description

Fix mirrored text in page API

### Proposed changes

- Add extra checks to determine if mirrored page exists before accessing the mirrored text

### Resolved issues

Fixes: #706


ulliholtgrave (approved):
Works for me :)

PR #710, opened 2021-02-25
Add simple API view tests

timoludwig:
### Short description

Since we break things from time to time, it's good to have at least some basic testing functionality to leave less room for errors.
This should in theory prevent bugs like #706 and #711 in the future.

### Proposed changes

This adds a few simple API view tests which check whether the API views return status code 200.

Missing API views (implementation a bit more complicated):
- Feedback API
- Push content API
- Offer API
- Single page API

Note: Since this test detects bugs which are not yet solved on the develop branch, I expect the tests to fail. Once the bugs are fixed, I can rebase this branch and check whether they succeed.

This PR depends on
- [x] #709 Fix mirrored text in page API
- [x] #715 Fix FieldError in PDF API

### DIFF
--- a/src/api/urls.py
+++ b/src/api/urls.py
@@ -26,29 +26,42 @@ from .v3.offers import offers
 from .v3.single_page import single_page
 
 urlpatterns = [
-    url(r"^regions/?$", regions),
-    url(r"^regions/live/?$", liveregions),
-    url(r"^regions/hidden/?$", hiddenregions),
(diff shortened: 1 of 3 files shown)


timoludwig:
I rebased this branch on the current develop branch and it seems that the test now succeed :tada:

ulliholtgrave (approved):
Nice, this should improve the stability of our project :+1:

Issue #484, opened 2020-09-04
Do not serve node_modules under /static


timoludwig:
### Motivation

It's a bit overkill to serve the complete `node_modules` directory under `/static` just to make our npm dependencies work as expected.

### Proposed Solution

Instead, we should use browserify in combination with babel to bundle all npm dependencies we need (tinymce, umbrellajs, feather-icons, base64-js and chart.js) in the default static directory.

### Alternatives

We could also use another framework than browserify, but as we use this already for other tasks, we might as well stick with it.

### Additional Context

- For tailwind, we could use [django-tailwind](https://pypi.org/project/django-tailwind/)
- After this task is done, remove `STATICFILES_DIRS` from `settings.py`.

PR #712, opened 2021-02-25
Do not serve tailwind from node_modules

jnugh:
### Short description
Use django-tailwind to generate a tailwind css. This
allows for using a tailwind configuration file as
well as using purgeCss for production builds.

### Proposed changes
Remove the way tailwind is currently included (in styles.less and statically from node modules)
Use the `django-tailwind` package and a standard tailwind theme that includes the `tailwind` config and additional styles.
Configure tailwind builds to use purgeCss by analyzing the html files from `cms/templates`.

### Resolved issues
See #484 (just a fraction)

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "6231e8e61ae84ee477e43dfc77e329cbb28cdafe8a39d8134521c29e0eb855c4"
+            "sha256": "6bffb3891f2f23d48b0836684cfbd88d3dd636106187263a9d69abd4081bd698"
         },
         "pipfile-spec": 6,
         "requires": {
@@ -147,14 +147,6 @@
                 "sha256:9d886f19f031d26bb1391c055ed9be06fb226a04a4cec1842b372c58873b3caa"
             ],
             "version": "==3.0.0"
-        },
-        "django-tailwind": {
-            "hashes": [
-                "sha256:2c1bf6c91fc54c844b85e8b379c211c971a2827c3d5b84f7e0464485ef9b8adf",
-                "sha256:d3aadba6aa046ead22efc68584b9ddccd0f25ccfbad433bb4727bee684a02177"
-            ],
-            "markers": "python_version >= '3.7' and python_version < '4'",
-            "version": "==1.1.1"
         },
         "django-widget-tweaks": {
             "hashes": [
@@ -672,6 +664,14 @@
             "index": "pypi",
             "version": "==1.8.0"
         },
+        "django-tailwind": {
+            "hashes": [
+                "sha256:2c1bf6c91fc54c844b85e8b379c211c971a2827c3d5b84f7e0464485ef9b8adf",
+                "sha256:d3aadba6aa046ead22efc68584b9ddccd0f25ccfbad433bb4727bee684a02177"
+            ],
+            "index": "pypi",
+            "version": "==1.1.1"
+        },
         "docutils": {
             "hashes": [
                 "sha256:0c5b78adfbf7762415433f5515cd5c9e762339e23369dbe8000d84a4bf4ab3af",
@@ -724,7 +724,7 @@
                 "sha256:c729845434366216d320e936b8ad6f9d681aab72dc7cbc2d51bedc3582f3ad1e",
                 "sha256:fff4f0c04e1825522ce6949973e83110a6e907750cd92d128b0d14aaaadbffdc"
             ],
-            "markers": "python_version >= '3.6' and python_version < '4'",
+            "markers": "python_version >= '3.6' and python_version < '4.0'",
             "version": "==5.7.0"
         },
         "jinja2": {
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -148,6 +148,14 @@
             ],
             "version": "==3.0.0"
         },
+        "django-tailwind": {
+            "hashes": [
+                "sha256:2c1bf6c91fc54c844b85e8b379c211c971a2827c3d5b84f7e0464485ef9b8adf",
+                "sha256:d3aadba6aa046ead22efc68584b9ddccd0f25ccfbad433bb4727bee684a02177"
+            ],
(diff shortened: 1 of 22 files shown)


dev-tools/run.sh
```diff
@@ -115,6 +110,7 @@ else
     fi
 
     # Start Integreat CMS
+    pipenv run integreat-cms-cli tailwind start &\
```
timoludwig, line 113:
```suggestion
    sudo -u $SUDO_USER env PATH="$PATH" pipenv run integreat-cms-cli tailwind start &
```
In this part of the script, all commands are executed as sudo to make sure the docker commands work even for users who aren't in the docker group. This means that for all commands not requiring sudo, we have to de-escalate privileges to the user who ran `sudo` in the first place (which is automatically stored in the env var `$SUDO_USER`).

dev-tools/run.sh
```diff
@@ -46,6 +43,7 @@ if nc -w1 localhost 5432; then
     ./dev-tools/migrate.sh
 
     # Start Integreat CMS
+    pipenv run integreat-cms-cli tailwind start &\
```
timoludwig, line 46:
```suggestion
    pipenv run integreat-cms-cli tailwind start &
```
I don't understand the backslash at the end of the line, is this a mac os thing?

Pipfile.lock
```diff
@@ -148,6 +148,14 @@
             ],
             "version": "==3.0.0"
         },
+        "django-tailwind": {
```
timoludwig, line 151: Please also add this dependency to the `Pipfile` (Technically, it will be installed if it's only contained in `Pipfile.lock`, but the next time someone changes the `Pipfile`, this dependency will be kicked out from `Pipfile.lock`).

timoludwig (commented):
Thanks for this PR! :rocket: 
I really like the idea of being able to customize our tailwind config!

Unfortunately, I couldn't get the setup running. Executing `pipenv run integreat-cms-cli tailwind install` succeeded, but somehow `node-sass` isn't installed correctly:
<details>
  <summary>pipenv run integreat-cms-cli tailwind start</summary>

  ```
> start
> npm run dev


> dev
> watch "npm run dev:sass && npm run dev:postcss" ./src

> Watching ./src

> dev:sass
> node-sass --output-style expanded --source-map true src/styles.scss ../static/css/styles.css

internal/fs/utils.js:269
    throw err;
    ^

Error: ENOENT: no such file or directory, scandir '/home/timo/job/integreat/integreat-cms/src/tailwind_theme/static_src/node_modules/node-sass/vendor'
    at Object.readdirSync (fs.js:955:3)
    at Object.getInstalledBinaries (/home/timo/job/integreat/integreat-cms/src/tailwind_theme/static_src/node_modules/node-sass/lib/extensions.js:134:13)
    at foundBinariesList (/home/timo/job/integreat/integreat-cms/src/tailwind_theme/static_src/node_modules/node-sass/lib/errors.js:20:15)
    at foundBinaries (/home/timo/job/integreat/integreat-cms/src/tailwind_theme/static_src/node_modules/node-sass/lib/errors.js:15:5)
    at Object.module.exports.missingBinary (/home/timo/job/integreat/integreat-cms/src/tailwind_theme/static_src/node_modules/node-sass/lib/errors.js:45:5)
    at module.exports (/home/timo/job/integreat/integreat-cms/src/tailwind_theme/static_src/node_modules/node-sass/lib/binding.js:15:30)
    at Object.<anonymous> (/home/timo/job/integreat/integreat-cms/src/tailwind_theme/static_src/node_modules/node-sass/lib/index.js:13:35)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Module.load (internal/modules/cjs/loader.js:863:32) {
  errno: -2,
  syscall: 'scandir',
  code: 'ENOENT',
  path: '/home/timo/job/integreat/integreat-cms/src/tailwind_theme/static_src/node_modules/node-sass/vendor'
}
npm ERR! code 1
npm ERR! path /home/timo/job/integreat/integreat-cms/src/tailwind_theme/static_src
npm ERR! command failed
npm ERR! command sh -c node-sass --output-style expanded --source-map true src/styles.scss ../static/css/styles.css

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/timo/.npm/_logs/2021-02-26T12_33_59_460Z-debug.log
```
  
</details>

Maybe you have an idea how to debug this?

dev-tools/run.sh
jnugh: The backslash escapes the newline. It would evaluate to this form: `pipenv run integreat-cms-cli tailwind start & pipenv run integreat-cms-cli runserver localhost:8000`.
I added it because the script will run in the background and watch for configuration/scss changes which should end when the run.sh script is terminated. But as I now learned a terminated shell script will kill it's child processes. 👍🏽

Pipfile.lock
jnugh: I added it to `setup.py` and ran `./dev-tools/update_dependencies.sh` which added it to `Pipfile.lock`. When to use `Pipfile` vs `setup.py`?

dev-tools/run.sh
timoludwig: Interesting, I didn't know that it's possible to write subsequent commands after `&`...

jnugh:
I had a global `node-sass`. I'm able to reproduce after removing it. Strange, I'll have a look. I see `node-sass` in the `node_modules` directory.

Pipfile.lock
timoludwig: Ah, I didn't see the entry in `setup.py`. The difference is whether the dependency is required on the production server or not ([see docu](https://integreat.github.io/integreat-cms/virtualenv.html#add-dependencies)).
In thought that the `tailwind build` step produces output which can be served as static files, which means that we only need this dependency on dev environments (=> `Pipfile`)?

jnugh: Got it, thanks 😄 prod only needs the css file. I'll change it accordingly.

Issue #711, opened 2021-02-25
PDF API not working


timoludwig:
### Describe the Bug

My new API view tests (see #710) revealed a new bug: The PDF API isn't working anymore.

### Steps to Reproduce

1. Go to http://localhost:8000/api/augsburg/de-de/pdf

### Expected Behavior

The PDF should be generated.

### Actual Behavior

The following error is shown:
```
FieldError at /api/augsburg/de-de/pdf

Cannot resolve keyword 'archived' into field. Choices are: children, created_date, editors, explicitly_archived, icon, id, last_updated, level, lft, mirrored_page, mirrored_page_first, mirrored_page_id, mirroring_pages, organization, organization_id, parent, parent_id, publishers, region, region_id, rght, translations, tree_id
```

PR #715, opened 2021-02-25
Fix FieldError in PDF API

timoludwig:
### Short description

Fix FieldError in PDF API

### Proposed changes

- Add possibility to return unrestricted queryset of all pages of a region
- Adapt PDF API to new concept of archived pages introduced in #598

### Resolved issues

Fixes: #711


ulliholtgrave (approved):
Checked it out locally and everything seems to work again. Perfect :+1:

PR #716, opened 2021-02-26
Access and set HTML content according to ESLint

ulliholtgrave:
### Short description
So far, we are accessing and changing elements by using the un-sanitized innerhtml()- and html()-method. According to ESLint this is bad behavior. I managed to use the correct methods for the innerhtml-parts, but we should do this for the html()-Methods as well.

See page_mirroring.js

### DIFF
--- a/src/cms/static/js/pages/page_mirroring.js
+++ b/src/cms/static/js/pages/page_mirroring.js
@@ -9,7 +9,7 @@ function render_mirrored_page_field(event){
     // Check if a region was selected
     if(event.target.value === "") {
         // Remove mirrored page field to make sure no old value is selected
-        u('#mirrored_page_div').html("");
+        u('#mirrored_page_div').empty();
         // Hide fields
         u('#mirrored_page_div').addClass('hidden');
         u('#mirrored_page_first_div').addClass('hidden');
@@ -27,7 +27,8 @@ function render_mirrored_page_field(event){
             u('#mirrored_page_first_div').removeClass('hidden');
         }).catch(function (error) {
             // Show error message instead of field
-            u('#mirrored_page_div').html('<div class="bg-red-100 border-l-4 border-red-500 text-red-500 px-4 py-3 my-4" role="alert"><p>' + error.message + '</p></div>');
+            u('#mirrored_page_div').html('<div class="bg-red-100 border-l-4 border-red-500 text-red-500 px-4 py-3 my-4" role="alert"><p id="mirrored-page-error"></p></div>');
+            u('#mirrored-page-error').text(error.message);
             u('#mirrored_page_div').removeClass('hidden');
             u('#mirrored_page_first_div').addClass('hidden');
         });
--- a/src/cms/static/js/revisions.js
+++ b/src/cms/static/js/revisions.js
@@ -40,7 +40,7 @@ function handle_revision_slider_input(event) {
     // The last updated date of the revision
     const revision_date = u("#revision-" + current_revision).data("date");
     // Update the revision info box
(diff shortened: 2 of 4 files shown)


src/cms/static/js/revisions.js
```diff
@@ -40,7 +40,7 @@ function handle_revision_slider_input(event) {
     // The last updated date of the revision
     const revision_date = u("#revision-" + current_revision).data("date");
     // Update the revision info box
-    revision_info.innerHTML = "Revision: " + current_revision + "<br>" + revision_date;
+    revision_info.textContent = `Revision: ${current_revision}\r\n${revision_date}`;
```
timoludwig, line 43: Nice, I didn't know this syntax :smile: 
However, it seems that [this is not supported in IE11](https://kangax.github.io/compat-table/es6/#test-template_literals), so maybe we should use normal string concatenation here?

ulliholtgrave: Sad :( But okay, I'll revert it :D

timoludwig: Oh, but wait: The revisions.js in going through browserify anyway (because of the htmldiff), so I think this is converted to something IE11 understands... :laughing:

timoludwig: Indeed, I just checked: after executing
```
npx browserify src/cms/static/js/revisions.js -o src/cms/static/js/revisions_browserified.js -t [ babelify --presets [ @babel/preset-env ] ]
```
the file `revisions_browserified.js` contains the line:
```
revision_info.textContent = "Revision: ".concat(current_revision, "\r\n").concat(revision_date);
```

ulliholtgrave: So, we can use this syntax in this file, but not in the rest of the project, right?

src/cms/static/js/pages/page_mirroring.js
```diff
@@ -22,12 +22,13 @@ function render_mirrored_page_field(event){
             throw new Error(response.statusText);
         }).then(function (html) {
             // Load rendered field into div and show it
-            u('#mirrored_page_div').html(html);
+            u('#mirrored_page_div').empty();
+            u('#mirrored_page_div').append(html);
```
timoludwig, line 26:
```suggestion
            u('#mirrored_page_div').html(html);
```
This essentially equivalent, so I propose to keep the simpler form of it and mark the warning as false positive.
Since the html is a direct response from our server, I don't see a security threat here.

src/cms/static/js/pages/page_mirroring.js
```diff
@@ -22,12 +22,13 @@ function render_mirrored_page_field(event){
...
             u('#mirrored_page_div').removeClass('hidden');
             u('#mirrored_page_first_div').removeClass('hidden');
         }).catch(function (error) {
             // Show error message instead of field
-            u('#mirrored_page_div').html('<div class="bg-red-100 border-l-4 border-red-500 text-red-500 px-4 py-3 my-4" role="alert"><p>' + error.message + '</p></div>');
+            u('#mirrored_page_div').append('<div class="bg-red-100 border-l-4 border-red-500 text-red-500 px-4 py-3 my-4" role="alert"><p>' + error.message + '</p></div>');
```
timoludwig, line 31: This isn't much different either. Maybe this is "safer":
```suggestion
            u('#mirrored_page_div').html('<div class="bg-red-100 border-l-4 border-red-500 text-red-500 px-4 py-3 my-4" role="alert"><p id="mirrored-page-error"></p></div>');
            u('#mirrored-page-error').text(error.message);
```
But still, the error message of the fetch function can never return anything which can be considered dangerous.

timoludwig (approved):

Issue #713, opened 2021-02-25
Unable to drag&drop page in page form more then once


jnugh:
### Describe the Bug

I'm able to drag & drop a page in the right sidebar once. After the drop the item keeps it's blue border and is not draggable anymore.

### Steps to Reproduce

1. Go to the page list
2. Open any page for edit that has at least one sibling
3. Drag and drop the page
4. See error!

### Expected Behavior
It should work multiple times.

### Actual Behavior
It only works once

### Additional Information
Latest Chrome on MacOS

![Kapture 2021-02-25 at 16 43 26](https://user-images.githubusercontent.com/143341/109177927-9c82af80-7788-11eb-90e2-d87e3637e3d3.gif)

PR #717, opened 2021-02-26
Fix drag & drop in page form

timoludwig:
### Short description

Fix drag & drop in page form. The cause of the bug was identical to #494:
In Chromium, manipulating the DOM during dragstart event fires a dragend event immediately

### Proposed changes

- Add delay to dragstart in order table of page form

### Resolved issues

Fixes: #713

### DIFF
--- a/src/cms/static/js/pages/page_order.js
+++ b/src/cms/static/js/pages/page_order.js
@@ -58,8 +58,10 @@ function register_event_handlers() {
     u('.drop').each(function(node) {
         u(node).off('dragleave,dragover,drop');
     });
-    // Event handlers for drag events
-    u('.drag').on('dragstart', dragstart);
+    // Event handlers for drag events (delay because of behaviour in Chrome browser)
+    u('.drag').on('dragstart', function(event) {
+        window.setTimeout(dragstart, 0, event);
+    });
     u('.drag').handle('dragend', dragend);
     // Event handlers for drop events
     u('.drop').each(function(node) {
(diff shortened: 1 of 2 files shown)


src/cms/static/js/pages/page_order.js
```diff
@@ -58,8 +58,10 @@ function register_event_handlers() {
     u('.drop').each(function(node) {
         u(node).off('dragleave,dragover,drop');
     });
-    // Event handlers for drag events
-    u('.drag').on('dragstart', dragstart);
+    // Event handlers for drag events (delay because of behaviour in Chrome browser)
+    u('.drag').on('dragstart', function(event) {
+        window.setTimeout(dragstart, 50, event);
```
jnugh, line 63:
```suggestion
        window.setTimeout(dragstart, 0, event);
```

You are using `setTimeout` to schedule the function call after the current execution finished. Waiting for at least 50ms does not make much sense.

jnugh (approved):
Working for me :+1:

src/cms/static/js/pages/page_order.js
timoludwig: A ok, I didn't really understand the problem, but just copied from
https://github.com/Integreat/integreat-cms/blob/4b918461b9315001da7f81fa6c165755b90d1b14/src/cms/static/js/tree_drag_and_drop.js#L13
Anyway, I changed both timeouts to `0` now. Thanks for pointing that out!

PR #718, opened 2021-02-26
Fix get_pages() for archived pages

timoludwig:
### Short description

In da858ad7199e27e49116bfc49ff722462f32ce2e (#715), I made a stupid mistake which resulted in `region.get_pages(archived=True)` returning all non-archived pages.
### Proposed changes

- Do not overwrite the `pages` variable with the non archived pages


ulliholtgrave (approved):
Nobody's perfect :P

Issue #606, opened 2020-12-08
Upgrade Django to version 3


timoludwig:
### Motivation

In April 2021, The most recent Django version 3 will receive long-term-support with 3.2 LTS until 2024 (see [supported versions](https://www.djangoproject.com/download/#supported-versions)). Since our used version 2.2 LTS is "only" supported until April 2022, we have to upgrade until then anyway. Since major version changes come with the risk of breaking changes, I think it makes sense to upgrade before our first stable release and the migration of WordPress to Django.

### Proposed Solution

Follow the guide [Upgrading Django to a newer version](https://docs.djangoproject.com/en/3.1/howto/upgrade-version/)

### Alternatives

Wait another year before upgrading

Issue #457, opened 2020-07-29
Add sqlite database back end


svenseeberg:
### Motivation
Currently, a PostgreSQL server is needed to run the CMS during development. PostgreSQL can either be installed on the host system or via docker container. It would be much easier to support sqlite as a (default) alternative. That would enable developers to just install a virtual environment (pipenv) and get going.

### Proposed Solution
Add a file `src/backend/sqlite_settings.py` and add a sqlite database back end as default. The `dev-tools` need adjustments, as they check for open local ports and running docker containers.

svenseeberg:
Mael, Emmanuel and I looked into using SQlite as an alternative to PostgreSQL while developing. Currently, we are using PostgreSQL specific model properties, which cannot simply be translated to SQLite. Should we close this issue or change our models? Currently, the problem is `from django.contrib.postgres.fields import ArrayField`.

timoludwig:
Hmm, I think this depends on how you prioritise this issue. For me personally, I think we already have an easy way to get started by installing docker. Sure, SQLite would be even simpler, but I don't think it's necessary if it requires much effort (You would not only need to come up with a new data type for arrays in the models, but also adjust the form handling and saving process).

I'd suggest leaving the issue open for now but as it is with low priority and milestone "additional features". If at some point somebody wants to implement it, why not...

timoludwig:
- `JSONField` isn't an issue anymore after Django 3.1 (see #723)
- Suggestion for `ArrayField`:
    - Convert `Region.push_notification_channels` (Array of strings) to a normal ForeignKey to the `PushNotificationChannel` model
    - Convert `RecurrenceRule.weekdays_for_weekly` (Array of ints) to a `JSONField`

JoeyStk:
> I'd vote for closing this ticket as wontfix 😉

Agreed

JoeyStk:
In a popular vote within the team it was decided that there is no need to implement this any time soon.

PR #723, opened 2021-03-02
Upgrade Django to version 3

timoludwig:
### Short description

The only problem I faced was that the documentation build failed with Django 3.1.
Therefore, I implemented a fix in [sphinxcontrib-django2](https://github.com/timoludwig/sphinxcontrib-django2) (see https://github.com/timoludwig/sphinxcontrib-django2/commit/2bbbcf20cb5eb9e7931caad04391048ba0c7266b and https://github.com/timoludwig/sphinxcontrib-django2/commit/b1a596da474f35447552a0f7b760494362430e9e).
I couldn't find any further problems in the running system with version 3.1 and I don't think we're using anything which is listed in the deprecated features of [3.0](https://docs.djangoproject.com/en/3.1/releases/3.0/#features-deprecated-in-3-0) and [3.1](https://docs.djangoproject.com/en/3.1/releases/3.1/#features-deprecated-in-3-1).
This should make it trivial to upgrade to Django 3.2 LTS as soon as it's released.

### Proposed changes

- Upgrade Django to version 3.1
- Remove deprecated postgres JSONField
- Upgrade [sphinxcontrib-django2](https://github.com/timoludwig/sphinxcontrib-django2) to make sure the documentation build still works

### Resolved issues

Fixes: #606

### DIFF
--- a/sphinx/conf.py
+++ b/sphinx/conf.py
@@ -12,6 +12,8 @@ import sys
 import inspect
 import importlib
 
+from django import VERSION as django_version
+
 from backend.settings import VERSION
 
 # Append project source directory to path environment variable
@@ -20,6 +22,8 @@ sys.path.append(os.path.abspath("../src/"))
 sys.path.append(os.path.abspath("./"))
 #: The path to the django settings module (see :doc:`sphinxcontrib-django2:readme`)
 django_settings = "backend.sphinx_settings"
+#: The "major.minor" version of Django
+django_doc_version = f"{django_version[0]}.{django_version[1]}"
 
 # -- Project information -----------------------------------------------------
 
@@ -57,7 +61,10 @@ extensions = [
 ]
 #: Enable cross-references to other documentations
 intersphinx_mapping = {
-    "python": ("https://docs.python.org/3.7/", None),
+    "python": (
+        f"https://docs.python.org/{sys.version_info.major}.{sys.version_info.minor}/",
+        None,
+    ),
     "pipenv": ("https://pipenv.pypa.io/en/latest/", None),
     "requests": ("https://requests.readthedocs.io/en/master/", None),
     "sphinx": ("https://www.sphinx-doc.org/en/master/", None),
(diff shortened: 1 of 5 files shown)


sphinx/conf.py
```diff
@@ -20,6 +22,8 @@
 sys.path.append(os.path.abspath("./"))
 #: The path to the django settings module (see :doc:`sphinxcontrib-django2:readme`)
 django_settings = "backend.sphinx_settings"
+# Split by "." and take the first two parts to get "major.minor"
+django_minor_version = ".".join(django_version.split(".")[:2])
```
svenseeberg, line 26: Totally okay, but did you know that there is also another solution:
```
>>> import django
>>> django.VERSION
(2, 0, 0, 'final', 0)
```

svenseeberg (approved):
Looks like it wasn't as much work as I expected :+1:

src/cms/models/offers/offer_template.py
```diff
@@ -32,7 +31,7 @@ class OfferTemplate(models.Model):
         verbose_name=_("URL"),
         help_text=_("This will be an external API endpoint in most cases."),
     )
-    post_data = JSONField(
+    post_data = models.JSONField(
```
svenseeberg, line 34: This is awesome. I think that was the main blocker for using SQLite in Dev model.

sphinx/conf.py
timoludwig: I did, but at the time I couldn't come up with a better solution than `".".join(map(str, django.VERSION[:2]))` (which isn't that much better)... after giving it some thought, maybe `f"{django.VERSION[0]}.{django.VERSION[1]}"` is indeed more readable?

src/cms/models/offers/offer_template.py
timoludwig: For `Region.push_notification_channels` (Array of strings), I think we should just go for a normal ForeignKey to the `PushNotificationChannel` model instead of `ArrayField`...
For `RecurrenceRule.weekdays_for_weekly` (Array of ints) we can maybe also use a `JSONField`?

timoludwig: Let's continue this discussion in #457

sphinx/conf.py
timoludwig: I updated the PR accordingly.

Issue #724, opened 2021-03-03
Enhancement for rss feed icon


melegiul:
### Motivation

Reshape the rss feed icon to meet the origins wordpress shape. 

### Proposed Solution

Currently the rss feed icon is in a smaller cube shape, to make it more similar to its origin wordpress version it should be added rounded corners and the size should be increased.

### Alternatives

Keep it as it is.

### Original wordpress rss feed icon

![rssfeed](https://user-images.githubusercontent.com/47010475/109797372-ab021880-7c19-11eb-9f94-b52e6c239deb.png)

timoludwig:
Just for reference: The further adjustments of the RSS feed are tracked in  #687.

PR #725, opened 2021-03-03
Updated RSS feed icon shape

melegiul:
### Short description

Reshape rss feed icon

### Proposed changes

- Added rounded corners and increased icon size.

### Resolved issues

Fixes: #724

### Proposed icon shape

reshaped

### DIFF
--- a/src/cms/templates/dashboard/dashboard.html
+++ b/src/cms/templates/dashboard/dashboard.html
@@ -24,7 +24,7 @@
                 </button>
                 <div class="collapsible-content flex flex-col active">
                     <div class="flex py-4">
-                        <img src="{% static 'images/favicon.ico' %}"/>
+                        <img src="{% static 'images/favicon.ico' %}" class="rounded-full h-20"/>
                         <div class="flex flex-col mx-auto">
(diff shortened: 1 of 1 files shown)


timoludwig (commented):
Thanks for this PR, but I'm not sure if that's the right way to go. I see two problems:

1. The increased size just takes away space from the dashboard. The users know how the Integreat logo looks, and also the logo is always visible in the top left corner, so I see no reason to place it prominently in the RSS feed again.
2. The Integreat logo has slight rounded edges, but a completely round circle does not reflect the corporate design of our logo. I don't know why it was designed like this in Wordpress, and I don't see a reason why we would need the exact copy in Django. Probably it's the best idea to upload another correct logo with rounded edges instead of trying to mimic it in css.

And just for reference: The current look of the RSS feed doesn't event contain a logo in WordPress:
Screenshot_2021-03-03 Dashboard ‹ Integreat — WordPress


Let's discuss this topic in today's team call!

melegiul:
Yes, we can discuss it later. My reference for that PR is the screenshot from #5, there you can check out my "role model" for the RSS feed layout. But #5 is from 2018 and I have no access to the current wordpress cms, that might be the cause for the deviation to your posted screenshot.

PR #732, opened 2021-03-04
Improve output of check-translations.sh script

timoludwig:
### Short description

It bothered me that it was sometimes unclear why the translation check fails and it took multiple fixing loops until everything was fine. This should ideally fix this by showing the exact lines which cause the warnings.

### Proposed changes

- Print which translations exactly are missing or incorrect
- Show diff if translation file is not up to date
- Show hint how to disable the checking script
- Fix bug where translation file is considered out of date if the last change was less than a minute ago
- Lots of unnecessary bash magic to make everything beautiful

check-translations

### DIFF
--- a/dev-tools/check_translations.sh
+++ b/dev-tools/check_translations.sh
@@ -2,48 +2,120 @@
 
 # This script can be used to check the translation file for missing or empty entries.
 
-if [[ "$VIRTUAL_ENV" != "" ]]
-then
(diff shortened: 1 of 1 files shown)


ulliholtgrave (approved):
Looks very good. Just tested your branch and the output is waaaaaaaaaaaay better then before, thank you! :)

Issue #691, opened 2021-02-23
Improve page form


timoludwig:
### Motivation

At the moment, the page form needs further adjustments:

### Proposed Solution

- Language tabs smaller
- Title field at the top
- After title has changed: permalink should be automatically filled (needs ajax request to generate unique slug)
- Permalink only editable after clicking on edit button
- Revision renamed to "Version"
- Version and Status moved to the top right
- URL parameter renamed to "Page link" / "Seitenlink"
- If most recent version is public, the "Publish" button should be renamed to "Update" / "Aktualisieren"
- Change wording of message if newest version is draft but old version is public:
    - "The latest changes have only been saved as a draft. Currently, version {Revision ID} of this page is displayed in the app."
    - "Die letzten Änderungen wurden nur als Entwurf gespeichert. Aktuell wird Version {Revision ID} dieser Seite in der App angezeigt."
- Title label renamed to "Title of the page" / "Titel der Seite"
- Content label renamed to "Content of the page"/ "Inhalt der Seite"
- Rename "Settings for all translations" to "Settings of the page" / "Einstellungen der Seite"
- Rename "Thumbnail icon" / "Vorschaubild" to "Icon" / "Icon"
- Rename status "public" to "published" / "veröffentlicht"
- Rename "side by side view" to "Translator view" / "Übersetzeransicht"
- Move "archive page" and "delete page" into separate box called "Actions"
- Move "minor edit"-box into separate box in the right sidebar at the top (near the publish button)
- Make right sidebar more narrow (in WordPress, the sidebar has 280px)

### Additional Context

![Unbenannt](https://user-images.githubusercontent.com/16021269/108826054-25042300-75c4-11eb-92fb-0905eddfda31.PNG)

PR #731, opened 2021-03-04
Switch JS/CSS bundling process to webpack

jnugh:
### Draft
- [x] I would like to check if CI works. 
- [x] I will add some docs before removing draft state.
- [x] Chart.JS needs to be handled @ulliholtgrave is currently working in this area - we need to sync up on this before proceeding

### Short description
Switch JS/CSS bundling to webpack in order to streamline the way, frontend assets are processed and imported.

### Proposed changes
This PR introduces a webpack configuration that is able to generate two JS/CSS bundles that can be used across the application instead of building a separate set for certain pages each time.
Bundles are:
* editor.js/css - includes TinyMCE and the configuration parser because this is rather large
* index.js/css - includes everything else

Webpack takes the following steps:
1. files are written in TypeScript -> `ts-loader`
2. Run files trough babel to make sure they work in target browser (which are defined in `.browserslistrc`)
3. Process included (s)css files and run postcss to purge tailwind classes and make sure css works in target browser (`.browserslistrc` again)
4. Copy fonts to target directory
5. Copy `src/frontend/tinymce-plugins` to `generated/tinymce-plugins` because TinyMCE wants to load some plugins directly by making an individual http call. Those files are not handled by the steps described above

I also removed browserify which was used for some files.
I also changed calles to `u` (umbrellaJS) to standard APIs (e.g. `document.querySelector`).

### Resolved issues

Fixes: #484
Supersedes #712

PR #733, opened 2021-03-05
Improve page form

melegiul:
### Short description

Improves page form layout and renames some labels.

### Proposed changes

- [x] Language tabs smaller
- [x] Title field at the top
- [x] After title has changed: permalink should be automatically filled (needs ajax request to generate unique slug)
- [x] Permalink only editable after clicking on edit button
- [x] Revision renamed to "Version"
- [x] Version and Status moved to the top right
- [x]  URL parameter renamed to "Page link" / "Seitenlink"
- [x]  If most recent version is public, the "Publish" button should be renamed to "Update" / "Aktualisieren"
- [x]  Change wording of message if newest version is draft but old version is public:
        "The latest changes have only been saved as a draft. Currently, version {Revision ID} of this page is displayed in the app."
        "Die letzten Änderungen wurden nur als Entwurf gespeichert. Aktuell wird Version {Revision ID} dieser Seite in der App angezeigt."
- [x]  Title label renamed to "Title of the page" / "Titel der Seite"
- [x] Content label renamed to "Content of the page"/ "Inhalt der Seite"
- [x] Rename "Settings for all translations" to "Settings of the page" / "Einstellungen der Seite"
- [x] Rename "Thumbnail icon" / "Vorschaubild" to "Icon" / "Icon", currently missing because I can't this label
- [x] Rename status "public" to "published" / "veröffentlicht"
- [x] Rename "side by side view" to "Translator view" / "Übersetzeransicht"
- [x] Move "archive page" and "delete page" into separate box called "Actions"
- [x] Move "minor edit"-box into separate box in the right sidebar at the top
- [x] Make right sidebar more narrow, corresponding to WordPress, the sidebar has 280px width fixed now

### Resolved issues

Fixes: #691

### Screenshots

Screenshot from 2021-03-08 15-51-35

Screenshot from 2021-03-05 11-56-02

### DIFF
--- a/src/cms/static/js/forms/update-permalink.js
+++ b/src/cms/static/js/forms/update-permalink.js
@@ -1,0 +1,25 @@
+/* this file updates the permalink of the current page form
+when the user edited the page translations title */
+
+document.getElementById("id_title").addEventListener("focusout", function(e){
+    let currentTitle = e.target.value;
+    let url = document.querySelector('[for="id_title"]').dataset.slugifyUrl;
+    slugify(url, {"title": currentTitle})
+        .then(response => {
+        document.getElementById('id_slug').setAttribute("value", response["unique_slug"]);
+        });
+});
+
+async function slugify(url, data) {
+    const response = await fetch(url, {
+        method: 'POST',
+        headers: {
+            'Content-Type': 'application/json',
+            'HTTP_X_REQUESTED_WITH': 'XMLHttpRequest',
+            'X-CSRFToken': document.getElementsByName('csrfmiddlewaretoken')[0].value
+        },
+        body: JSON.stringify(data)
+    })
+    .then(response => response.json())
+    return response;
+}
--- a/src/cms/templates/pages/page_form.html
+++ b/src/cms/templates/pages/page_form.html
@@ -39,8 +39,8 @@
 {% has_perm 'cms.edit_page' request.user page as can_edit_page %}
 <form enctype="multipart/form-data" method="post" id="content_form">
     {% csrf_token %}
-    <div class="flex flex-wrap">
-        <div class="w-4/5 flex flex-wrap flex-col justify-center mb-6">
+    <div class="flex flex-wrap items-start">
+        <div class="w-4/5 mb-6">
             <h2 class="heading font-normal">
                 {% if page %}
                     {% if page_translation_form.instance.id %}
@@ -68,22 +68,27 @@
             {% endif %}
             {% has_perm 'cms.publish_page' request.user page as can_publish_page %}
             {% if can_publish_page %}
-                <input type="submit" name="submit_public" class="cursor-pointer bg-blue-500 hover:bg-blue-600 text-white font-bold py-3 px-4 rounded mr-2" value="{% trans 'Publish' %}" />
+                <input type="submit" name="submit_public" class="cursor-pointer bg-blue-500 hover:bg-blue-600 text-white font-bold py-3 px-4 rounded mr-2 whitespace-nowrap"
+                    value="{% if page_translation_form.instance.status == PUBLIC %}
+                            {% trans 'Update' %}
+                        {% else %}
+                            {% trans 'Publish' %}
+                        {% endif %}"/>
             {% else %}
                 <input type="submit" name="submit_review" class="cursor-pointer bg-blue-500 hover:bg-green-600 text-white font-bold py-3 px-4 rounded mr-2" value="{% trans 'Submit for review' %}" />
             {% endif %}
         {% endif %}
         </div>
-        <div class="w-2/3 flex flex-wrap flex-col pr-2">
+        <div class="flex flex-wrap flex-col flex-grow pr-2 w-2/3">
             {{page_form.errors}}
             {{page_translation_form.errors}}
-            <ul class="flex flex-wrap pl-4">
+            <ul class="flex flex-wrap items-end pl-4">
                 {% for other_language in languages %}
                     <li class="mr-1 {% if other_language == language %}z-10{% endif %}" style="margin-bottom: -2px">
                         <div class="bg-white text-blue-500 {% if other_language != language %}hover:bg-blue-500 hover:text-white{% endif %} border-l-2 border-t-2 border-r-2 border-blue-500 font-bold rounded-t-lg">
                             <div class="border-b-2 border-white">
                                 {% if other_language == language %}
-                                    <div class="{{ language.slug }} p-4">
+                                    <div class="{{ language.slug }} px-4">
                                         {% if page %}
                                             {% if other_language in page.languages %}
                                                 {% if page_translation_form.instance.is_outdated %}
@@ -124,7 +129,7 @@
                                         {{ other_language.translated_name }}
                                     </div>
                         {% else %}
-                            <a class="p-4 block" style="color: inherit;" href="{% url 'edit_page' page_id=page.id region_slug=region.slug language_slug=other_language.slug %}">
+                            <a class="px-4 block" style="color: inherit;" href="{% url 'edit_page' page_id=page.id region_slug=region.slug language_slug=other_language.slug %}">
                                 {% if other_language in page.languages %}
                                     {% get_translation page_form.instance other_language.slug as other_page_translation %}
                                     {% if other_page_translation.is_outdated %}
@@ -152,21 +157,37 @@
                         </div>
                     </li>
                 {% endfor %}
+                {% if page_translation_form.instance.id %}
+                    <li class="self-start ml-auto px-2">
+                        {% trans 'Version' %}:
+                        {{ page_translation_form.instance.version }}
+                        (<a href="{% url 'page_revisions' page_id=page.id region_slug=region.slug language_slug=language.slug %}"
+                            class="text-blue-500 hover:underline">{% trans 'Show versions' %}</a>)
+                    </li>
+                {% endif %}
             </ul>
             <div class="w-full mb-4 rounded border-2 border-blue-500 bg-white flex-auto">
                 <div class="w-full p-4">
+                    <div class="flex justify-between">
+                        <label for="{{ page_translation_form.title.id_for_label }}"
+                            data-slugify-url="{% url 'slugify_ajax' region_slug=region.slug language_slug=language.slug %}{% if page_form.instance.id %}?page={{ page_form.instance.id }}{% endif %}"
+                            class="mb-2 mt-4 font-bold">{{ page_translation_form.title.label }}
+                        </label>
+                        {% if page_translation_form.instance.id %}
+                            <span class="mb-2 mt-4 mr-8">
+                                <span class="font-bold">{% trans 'Status' %}:</span>
+                                {% if page.explicitly_archived %}
+                                    {% trans 'Archived' %}
+                                {% elif page.implicitly_archived %}
+                                    {% trans 'Archived, because a parent page is archived' %}
+                                {% else %}
+                                    {{ page_translation_form.instance.get_status_display }}
+                                {% endif %}
+                            </span>
+                        {% endif %}
+                    </div>
+                    {% render_field page_translation_form.title class="appearance-none block w-full bg-gray-200 text-xl text-gray-800 border border-gray-200 rounded py-3 px-4 mb-2 leading-tight focus:outline-none focus:bg-white focus:border-gray-400" %}
                     {% if page_translation_form.instance.id %}
-                        <span class="inline-block mt-4 mb-2 font-bold">{% trans 'Revision' %}:</span>
-                        {{ page_translation_form.instance.version }}
-                        (<a href="{% url 'page_revisions' page_id=page.id region_slug=region.slug language_slug=language.slug %}" class="text-blue-500 hover:underline">{% trans 'Show revisions' %}</a>)<br>
-                        <span class="inline-block mb-2 font-bold">{% trans 'Status' %}:</span>
-                        {% if page.explicitly_archived %}
-                            {% trans 'Archived' %}
-                        {% elif page.implicitly_archived %}
-                            {% trans 'Archived, because a parent page is archived' %}
-                        {% else %}
-                            {{ page_translation_form.instance.get_status_display }}
-                        {% endif %}<br>
                         <span class="inline-block mb-2 font-bold">{% trans 'Short URL' %}:</span>
                         <a href="{{ page_translation_form.instance.short_url }}" target="_blank" rel="noopener noreferrer" class="text-blue-500 hover:underline">{{ page_translation_form.instance.short_url }}</a>
                         <a href="#" onclick="short_url_to_clipboard('{{ page_translation_form.instance.short_url }}');" title="{% trans 'Copy to clipboard' %}" class="px-2 text-gray-800 hover:text-blue-500">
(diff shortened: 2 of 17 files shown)


timoludwig:
Amazing, thanks for this PR!

> Rename "Thumbnail icon" / "Vorschaubild" to "Icon" / "Icon", currently missing because I can't this label

I renamed this already in c59b59acc4317f5cb9b216f602ed1360b831a757, sorry I forgot to update it in the issue!

melegiul:
> I renamed this already in [c59b59a](https://github.com/Integreat/integreat-cms/commit/c59b59acc4317f5cb9b216f602ed1360b831a757), sorry I forgot to update it in the issue!

Thank you for your hint, so only 2 more left, we're getting closer :smiley:

melegiul:
There is one difference to the proposed solution: instead of using an ajax request for slug generation, I found a neat javascript function `slugify` that does the trick

timoludwig (commented):
Thanks for this PR! :rocket: 

In general, it looks very good! I just have some questions regarding the new slugify mechanism and some minor suggestions on implementation details. Sorry if I'm too nitpicking regarding variable names etc. - you don't have to adopt my suggestions if you don't agree :smile:

src/cms/static/js/forms/update-permalink.js
```diff
@@ -0,0 +1,25 @@
+/* this file updates the permalink of the current page form
+when the user edited the page translations title */
+
+u("#id_title").on('focusout', function(e){
```
timoludwig, line 4: Oh, and one additional comment: Since umbrellaJS is deprecated and will be removed in PR #731, it probably makes sense to use the standard JS-syntax already. Sorry, that you had to learn a new syntax which is now being discarded :unamused: 
```suggestion
document.getElementById("id_title").addEventListener("focusout", function(e){
```

melegiul: I remember, you said that recently during a call, so using umbrella here is not "future-proof" :sweat_smile:
I will update to standard js, that's not much effort for that PR, but it was fun working with umbrella.
Umbrella has so many common features to jquery, I think gaining experience in this area is an advantage :smile:

src/cms/templates/pages/page_form.html
```diff
@@ -152,21 +153,37 @@ <h2 class="heading font-normal">
                         </div>
                     </li>
                 {% endfor %}
+                {% if page_translation_form.instance.id %}
+                    <li class="self-start bg-white ml-auto px-2">
```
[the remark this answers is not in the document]
melegiul: Yes, looks better, I stuck too close to the mock-up

src/cms/utils/slug_utils.py
```diff
@@ -13,6 +13,34 @@
...
+        "region": form_object.region,
+        "language": form_object.language,
+        "slug": form_object.cleaned_data["slug"],
+        "manager": form_object.Meta.model.objects,
+        "page_translation": form_object.instance,
+    }
+    return generate_unique_slug_helper(foreign_model, **kwargs)
+
+
+def generate_unique_slug_helper(foreign_model, **kwargs):
```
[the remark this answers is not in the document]
melegiul: No, there is no reason to exclude `foreign_model` from kwargs, treating the parameters equally sounds reasonable to me
I will also exchange the function names, my choice was just because of the compatibility for all "users" of that function

src/cms/templates/pages/page_form.html
```diff
@@ -186,37 +203,55 @@ <h2 class="heading font-normal">
...
                     {% render_field page_translation_form.text class="bg-gray-200 w-full p-2 border border-gray-200 focus:outline-none focus:bg-white focus:border-gray-400 rounded tinymce_textarea" %}
-                    <span class="block mb-2 mt-4 font-bold">{% trans 'Implications on other translations' %}</span>
-                    {% render_field page_translation_form.minor_edit %}
-                    <label for="{{ page_translation_form.minor_edit.id_for_label }}" class="text-s">{{ page_translation_form.minor_edit.label }}</label>
-                    <div class="my-2 text-s text-gray-600">{{ page_translation_form.minor_edit.help_text }}</div>
                 </div>
             </div>
         </div>
-        <div class="w-1/3 pl-4 flex flex-wrap flex-col">    
+        <div class="pl-4 flex flex-wrap flex-col" style="width: 280px">
```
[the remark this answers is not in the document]
melegiul: I just tried 400px, it's a good compromise and there is less scroll effort

src/cms/templates/pages/page_form.html
```diff
@@ -68,22 +68,23 @@ <h2 class="heading font-normal">
             {% endif %}
             {% has_perm 'cms.publish_page' request.user page as can_publish_page %}
             {% if can_publish_page %}
-                <input type="submit" name="submit_public" class="cursor-pointer bg-blue-500 hover:bg-blue-600 text-white font-bold py-3 px-4 rounded mr-2" value="{% trans 'Publish' %}" />
+                <input type="submit" name="submit_public" class="cursor-pointer bg-blue-500 hover:bg-blue-600 text-white font-bold py-3 px-4 rounded mr-2"
+                    value="{% if page %}{% trans 'Update' %}{% else %}{% trans 'Publish' %}{% endif %}" />
```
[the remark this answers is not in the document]
melegiul: Ok, I've updated the submit button, taking into account whether the page form is public or draft, see 897a62b785ef2d8d5644dbda2dda7281b4bddf2d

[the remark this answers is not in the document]
melegiul: Ok, I've updated the submit button, taking into account whether the page form is public or draft, see https://github.com/Integreat/integreat-cms/pull/733/commits/897a62b785ef2d8d5644dbda2dda7281b4bddf2d

src/cms/utils/slug_utils.py
```diff
@@ -26,76 +54,77 @@ def generate_unique_slug(form_object, foreign_model):
...
+        logger.info('region: "%s"', kwargs["region"])
+        logger.info('language: "%s"', kwargs["language"])
 
-    slug = form_object.cleaned_data["slug"]
+    slug = kwargs["slug"]
 
-    # if slug is empty, generate from title/name
-    if not slug:
+    # if slug is empty and page already exists, generate from title/name
+    if not slug and kwargs["page_translation"]:
```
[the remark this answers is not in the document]
melegiul: This is also addressed in: fad14931a736fd68794249704f997ffcd216af64

src/cms/templates/pages/page_form.html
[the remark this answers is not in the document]
timoludwig: Since we also have the status ["Pending Review"](https://github.com/Integreat/integreat-cms/blob/b0594e0b4ad816029835b8f8188e69d3eead9908/src/cms/constants/status.py#L10) (in this case the button should also be called "Publish" instead of "Update"), I think it would make sense to check `page_translation_form.instance.status != PUBLIC` instead of `page_translation_form.instance.status == DRAFT`...

src/cms/utils/slug_utils.py
```diff
@@ -26,76 +61,77 @@ def generate_unique_slug(form_object, foreign_model):
...
     :return: An unique slug identifier
     :rtype: str
     """
+    slug = kwargs["slug"]
+    foreign_model = kwargs["foreign_model"]
+    object_instance = kwargs["object_instance"]
+    fallback = kwargs["fallback"]
+    if args:
+        region = args[0]
+        language = args[1]
```
timoludwig, line 82: Is there a specific reason to use `args` instead of `kwargs`? 
```suggestion
    region = kwargs.get("region")
    language = kwargs.get("language")
```

src/cms/utils/slug_utils.py
```diff
@@ -12,7 +12,42 @@
...
+        fallback = "name"
+    kwargs = {
+        "slug": form_object.cleaned_data["slug"],
+        "cleaned_data": form_object.cleaned_data,
+        "manager": form_object.Meta.model.objects,
+        "object_instance": form_object.instance,
+        "foreign_model": foreign_model,
+        "fallback": fallback
+    }
+    return generate_unique_slug(*args, **kwargs)
```
timoludwig, line 47:
```suggestion
    kwargs = {
        "slug": form_object.cleaned_data["slug"],
        "cleaned_data": form_object.cleaned_data,
        "manager": form_object.Meta.model.objects,
        "object_instance": form_object.instance,
        "foreign_model": foreign_model,
        "fallback": "name"
    }
    if foreign_model in ["page", "event", "poi"]:
        kwargs.update({
            "region": form_object.region,
            "language": form_object.language,
            "fallback": "title,
        })
    return generate_unique_slug(**kwargs)
```

src/cms/utils/slug_utils.py
```diff
@@ -26,76 +61,77 @@ def generate_unique_slug(form_object, foreign_model):
...
+                "language": language,
             }
         )
 
     # generate new slug while it is not unique
     while True:
         # get other objects with same slug
         other_objects = pre_filtered_objects.filter(slug=unique_slug)
-        if form_object.instance.id:
+        if object_instance:
```
timoludwig, line 123:
```suggestion
        if object_instance.id:
```
I think you need the check for `id` here, because form instances can exist before being saved to the database. These instances evaluate to ``True``, but ``object_instance.foreign_object`` fails with a ``RelatedObjectDoesNotExist``-error (Try creating a page while manually providing an empty slug).

src/cms/utils/slug_utils.py
```diff
@@ -26,76 +61,77 @@ def generate_unique_slug(form_object, foreign_model):
...
-            fallback = "title"
-        else:
-            fallback = "name"
-        if fallback not in form_object.cleaned_data:
+        logger.info('region: "%s"', region)
+        logger.info('language: "%s"', language)
+
+    # if slug is empty and fallback field is set, generate from fallback:title/name
+    if not slug and fallback:
+        if not hasattr(object_instance, fallback):
```
timoludwig, line 92:
```suggestion
        if fallback not in cleaned_data:
```
Now that you're not accessing the fallback-field of the `object_instance`, but the cleaned data instead via `kwargs["cleaned_data"][fallback]`, you also need to check whether the value is present in the `cleaned_data` instead of the `object_instance`...

src/cms/utils/slug_utils.py
```diff
@@ -26,76 +61,77 @@ def generate_unique_slug(form_object, foreign_model):
...
     :raises ~django.core.exceptions.ValidationError: When no slug is given and there is also no field which can be used
                                                      as fallback (either ``title`` or ``name``).
 
     :return: An unique slug identifier
     :rtype: str
     """
+    slug = kwargs["slug"]
+    foreign_model = kwargs["foreign_model"]
+    object_instance = kwargs["object_instance"]
+    fallback = kwargs["fallback"]
```
timoludwig, line 79:
```suggestion
    slug = kwargs.get("slug")
    foreign_model = kwargs.get("foreign_model")
    object_instance = kwargs.get("object_instance")
    fallback = kwargs.get("fallback")
```
While you're at it, I'd suggest the `get()`-syntax here as well, because then all values which are not given are just ``None`` instead of raising an exception.

src/cms/views/pages/page_actions.py
```diff
@@ -858,3 +860,39 @@ def render_mirrored_page_field(request):
...
+        language=language,
+    ).first()
+    kwargs = {
+        "slug": form_title,
+        "manager": PageTranslation.objects,
+        "object_instance": page_translation,
+        "foreign_model": "page",
+        "fallback": None
+    }
+    unique_slug = generate_unique_slug(region, language, **kwargs)
```
timoludwig, line 897:
```suggestion
        "region": region,
        "language": language,
    }
    unique_slug = generate_unique_slug(**kwargs)
```

src/cms/utils/slug_utils.py
```diff
@@ -13,6 +13,34 @@
...
+
+    :raises ~django.core.exceptions.ValidationError: When no slug is given and there is also no field which can be used
+                                                     as fallback (either ``title`` or ``name``).
+
+    :return: An unique slug identifier
+    :rtype: str
+    """
+    kwargs = {
+        "region": form_object.region,
+        "language": form_object.language,
```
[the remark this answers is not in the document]
timoludwig: Yeah, I think this is a good approach! :+1:

src/cms/templates/pages/page_form.html
```diff
@@ -186,37 +203,55 @@ <h2 class="heading font-normal">
                             </div>
                         {% endif %}
                     {% endif %}
-                    <label for="{{ page_translation_form.slug.id_for_label }}" class="block mb-2 font-bold">{{ page_translation_form.slug.label }}</label>
+                    <label for="{{ page_translation_form.slug.id_for_label }}" class="block mb-2 font-bold">
```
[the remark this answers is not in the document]
timoludwig: Since you didn't address this point yet, I think it makes sense to separate it from this PR. I opened a new issue: #735

timoludwig (commented):
Thanks for your changes! Nearly everything is fixed, but I noticed a few unresolved issues.

src/cms/utils/slug_utils.py
melegiul: Good point, I forgot to revert the if statement as well

src/cms/utils/slug_utils.py
melegiul: No, I have no preferences in this case, we can take kwargs

src/cms/templates/pages/page_form.html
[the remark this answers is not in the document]
melegiul: Alright, this one 46d7ce65c955f514798adced7febbdc9a98a6280 should cover pending review state too

melegiul:
Thank you for your further review, hopefully this should now cover all remaining issues.
I will wait for you to check the new changes, if you agree with them, I will do first a git squash before rebasing,
to keep conflict resolving shorter.

src/cms/utils/slug_utils.py
```diff
@@ -26,76 +63,72 @@ def generate_unique_slug(form_object, foreign_model):
...
+                "language": language,
             }
         )
 
     # generate new slug while it is not unique
     while True:
         # get other objects with same slug
         other_objects = pre_filtered_objects.filter(slug=unique_slug)
-        if form_object.instance.id:
+        if object_instance.id:
```
timoludwig, line 120:
```suggestion
        if object_instance and object_instance.id:
```
Sorry, I just noticed my suggestion fixes the form submission when the page doesn't exist, but the ajax call doesn't work in this case (it results in an error `AttributeError: 'NoneType' object has no attribute 'id'`)... so I think we have to check both - whether `object_instance` evaluates to `True` and whether it has an `id` already...

timoludwig (commented):
> Thank you for your further review, hopefully this should now cover all remaining issues.
> I will wait for you to check the new changes, if you agree with them, I will do first a git squash before rebasing,
> to keep conflict resolving shorter.

Thanks, this sounds good. I'm sorry, I made a wrong suggestion previously, so I think this has to be fixed again :neutral_face: 
But apart from that, I think this PR looks good finally! You can go ahead with rebasing and I will have one final look afterwards :+1:

src/cms/utils/slug_utils.py
melegiul: No problem, when it comes to non existing page forms, I getting always into trouble too :sweat_smile:

src/cms/views/pages/page_actions.py
```diff
@@ -858,3 +860,40 @@ def render_mirrored_page_field(request):
             "page_form": page_form,
         },
     )
+
+
+@login_required
+@region_permission_required
+@permission_required("cms.edit_pages", raise_exception=True)
+# pylint: disable=unused-argument
+def slugify_ajax(request, region_slug, language_code):
```
timoludwig, line 869:
```suggestion
def slugify_ajax(request, region_slug, language_slug):
```

src/cms/views/pages/page_actions.py
```diff
@@ -858,3 +860,40 @@ def render_mirrored_page_field(request):
...
+    """checks the current user input for page title and generates unique slug for permalink
+
+    :param request: The current request
+    :type request: ~django.http.HttpResponse
+    :param region_slug: region identifier
+    :type region_slug: str
+    :param language_code: language slug
+    :type language_code: str
```
timoludwig, line 877:
```suggestion
    :param language_slug: language slug
    :type language_slug: str
```

src/cms/views/pages/page_actions.py
```diff
@@ -858,3 +860,40 @@ def render_mirrored_page_field(request):
...
+    :return: unique page translation slug
+    :rtype: str
+    """
+    json_data = json.loads(request.body)
+    form_title = slugify(json_data["title"], allow_unicode=True)
+    region = Region.get_current_region(request)
+    language = get_object_or_404(region.languages, code=language_code)
```
timoludwig, line 884:
```suggestion
    language = get_object_or_404(region.languages, slug=language_slug)
```

src/cms/urls.py
```diff
@@ -362,6 +362,11 @@
                     name="get_page_order_table_ajax",
                 ),
                 url(r"^search_poi$", events.search_poi_ajax, name="search_poi_ajax"),
+                url(
+                    r"^(?P<region_slug>[-\w]+)/(?P<language_code>[-\w]+)/slugify$",
```
timoludwig, line 366:
```suggestion
                    r"^(?P<region_slug>[-\w]+)/(?P<language_slug>[-\w]+)/slugify$",
```

src/cms/templates/pages/page_form.html
```diff
@@ -68,22 +68,27 @@ <h2 class="heading font-normal">
...
             {{page_translation_form.errors}}
-            <ul class="flex flex-wrap pl-4">
+            <ul class="flex flex-wrap items-end pl-4">
                 {% for other_language in languages %}
                     <li class="mr-1 {% if other_language == language %}z-10{% endif %}" style="margin-bottom: -2px">
                         <div class="bg-white text-blue-500 {% if other_language != language %}hover:bg-blue-500 hover:text-white{% endif %} border-l-2 border-t-2 border-r-2 border-blue-500 font-bold rounded-t-lg">
                             <div class="border-b-2 border-white">
                                 {% if other_language == language %}
-                                    <div class="{{ language.slug }} p-4">
+                                    <div class="{{ language.code }} px-4">
```
timoludwig, line 91:
```suggestion
                                    <div class="{{ language.slug }} px-4">
```

src/cms/templates/pages/page_form.html
```diff
@@ -124,7 +129,7 @@ <h2 class="heading font-normal">
                                         {{ other_language.translated_name }}
                                     </div>
                         {% else %}
-                            <a class="p-4 block" style="color: inherit;" href="{% url 'edit_page' page_id=page.id region_slug=region.slug language_slug=other_language.slug %}">
+                            <a class="px-4 block" style="color: inherit;" href="{% url 'edit_page' page_id=page.id region_slug=region.slug language_code=other_language.code %}">
```
timoludwig, line 132:
```suggestion
                            <a class="px-4 block" style="color: inherit;" href="{% url 'edit_page' page_id=page.id region_slug=region.slug language_slug=other_language.slug %}">
```

src/cms/templates/pages/page_form.html
```diff
@@ -152,21 +157,37 @@ <h2 class="heading font-normal">
                         </div>
                     </li>
                 {% endfor %}
+                {% if page_translation_form.instance.id %}
+                    <li class="self-start ml-auto px-2">
+                        {% trans 'Version' %}:
+                        {{ page_translation_form.instance.version }}
+                        (<a href="{% url 'page_revisions' page_id=page.id region_slug=region.slug language_code=language.code %}"
```
timoludwig, line 164:
```suggestion
                        (<a href="{% url 'page_revisions' page_id=page.id region_slug=region.slug language_slug=language.slug %}"
```

src/cms/templates/pages/page_form.html
```diff
@@ -152,21 +157,37 @@ <h2 class="heading font-normal">
...
+                            class="text-blue-500 hover:underline">{% trans 'Show versions' %}</a>)
+                    </li>
+                {% endif %}
             </ul>
             <div class="w-full mb-4 rounded border-2 border-blue-500 bg-white flex-auto">
                 <div class="w-full p-4">
+                    <div class="flex justify-between">
+                        <label for="{{ page_translation_form.title.id_for_label }}"
+                            data-slugify-url="{% url 'slugify_ajax' region_slug=region.slug language_code=language.code %}{% if page_form.instance.id %}?page={{ page_form.instance.id }}{% endif %}"
```
timoludwig, line 173:
```suggestion
                            data-slugify-url="{% url 'slugify_ajax' region_slug=region.slug language_slug=language.slug %}{% if page_form.instance.id %}?page={{ page_form.instance.id }}{% endif %}"
```

melegiul:
Thank you for renaming the language codes, I've set those names before our recent language model improvement, so rebasing and squashing is now done

timoludwig (approved):
Amazing, thanks! :muscle:

PR #738, opened 2021-03-11
Improve logging

### LINKED ISSUE #450: Import classes at package level
### Motivation

At the moment, we have to reference e.g. forms and views by their sub-package paths (whereas in models, we can directly import `cms.models.Region`. It would be nice to have a consistent solution for all other Django packages, e.g. that forms and views get imported directly in the `__init__.py` of `cms.forms` and not in e.g. `cms.forms.pages`.
We don't have any namespace problems with this, as our models always contain the parent package name anyways (e.g. `PageForm` is a unique class name in our project).

### Proposed Solution

Move all imports in the `cms/forms/*/__init__.py` files to `cms/forms/__init__.py` and do the same for the other Django modules which we split up into packages.
If we run into cyclic imports, we might have to refactor smaller parts to circumvent import problems.

### Alternatives

We could at least reference the classes in first-level sub-packages of the standard Django modules (`cms.models.pages.Page`, `cms.forms.pages.PageForm`), but I think the proposed solution (`cms.models.Page`, `cms.forms.PageForm`) is cleaner.

### LINKED ISSUE #507: Improve logging levels
### Motivation

At the moment, some requests are logged very verbose, whereas other things aren't logged at all. This does not reflect the standards on logging levels: `debug`, `info`, `warning`, `error` and `critical`

### Proposed Solution

Change the logging levels to the values you think are reasonable and add debug information where it is missing.

E.g. logging similar to this: `logger.info("Form instantiated with data %s and instance %s", data, instance)` should rather be at the `debug` level.

Also improve the [settings file](https://github.com/Integreat/cms-django/blob/ae15187bc6e2526d554ada07936c66900dc9fd96/src/backend/settings.py#L214-L284) to make it easy to change the logging levels for all loggers at the same time without having to change many values.

timoludwig:
### Short description

This adds some improvements to the logging functionality. In preparation for this, I also refactored the forms a bit so we don't have to duplicate the logging for every form, but have one base form class which takes care of the logging.

### Proposed changes

- Refactor forms (fixes #450)
    - Import all forms in `cms.forms`
    - Split multiple classes into separate files
- Improve logging (fixes #507)
    - Add additional handlers for logfile and email to admins
    - Make logging levels consistent
    - Use `repr()` instead of `str()` when logging model instances

### Resolved issues

Fixes: #450
Fixes: #507

Notice: I moved the three previous commits to #739 (Adjust backend languages codes) and #740 (Improve models)

### DIFF
--- a/sphinx/conf.py
+++ b/sphinx/conf.py
@@ -121,6 +121,7 @@ nitpick_ignore = [
     ("py:attr", "django.contrib.auth.models.User.imprint_translations"),
     ("py:attr", "django.contrib.auth.models.User.chat_messages"),
     ("py:func", "django.contrib.sitemaps.Sitemap._urls"),
+    ("py:func", "django.utils.text.capfirst"),
     ("py:class", "django.core.handlers.WSGIHandler"),
     ("py:class", "django.forms.models.ModelChoiceIterator"),
     ("py:class", "mptt.models.MPTTModel"),
--- a/src/backend/settings.py
+++ b/src/backend/settings.py
@@ -14,6 +14,7 @@ For production use, the following settings can be set with environment variables
 * :attr:`~backend.settings.WEBAPP_URL`
 * :attr:`~backend.settings.STATIC_ROOT`
 * :attr:`~backend.settings.MEDIA_ROOT`
+* :attr:`~backend.settings.LOGFILE`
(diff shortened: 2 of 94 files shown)


ulliholtgrave (approved):
I wasn't able to test all the error or debug functions, but so far this looks pretty good :)

PR #739, opened 2021-03-12
Adjust backend languages codes

timoludwig:
### Short description

I noticed that PR #738 (Improve logging) got a little bit out of control, so I will split it into separate PRs.
This one is about the backend language codes. Recently, we changed our language models and used two-letter slugs for identifying languages. Since we need a mapping from backend to content languages, I propose to use "de" and "en" instead of "de-de" and "en-us" as backend languages as well.
 
### Proposed changes

- Adapt the backend languages to the changed content language model
- Remove unused `unify_language_slug` template tag ("en-us" and "en-gb" are both just "en" now)
- Fix language slugs in tests

### DIFF
--- a/src/backend/settings.py
+++ b/src/backend/settings.py
@@ -52,8 +52,8 @@ TEST_BLOG_ID = 154
 
 #: RSS feed URLs to the Integreat blog
 RSS_FEED_URLS = {
-    "en-us": "https://integreat-app.de/en/feed/",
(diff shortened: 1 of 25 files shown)


ulliholtgrave (approved):
Looks good and I've seen no significant changes in the behavior :)

svenseeberg (approved):

PR #740, opened 2021-03-12
Improve models

timoludwig:
### Short description

I noticed that PR #738 (Improve logging) got a little bit out of control, so I will split it into separate PRs.
This one provides a few enhancements for the models.

### Proposed changes

-  Improve handling of backend translations
    - Distinguish between `backend_translation`, `default_translation` and `best_translation` in the content models:
        - `backend_translation`: The translation of the page in the current backend language (used in the content lists in the column "Title in [backend language]")
        - `default_translation`: The translation of the page in the region's default language
        - `best_translation`: If `backend_translation` exists, it is returned and `default_translation` as fallback (used anywhere else in the backend where `None` is no acceptable result)
    - Remove some complexity in the templates
-  Improve string representations of models
    - Make output of `__str__()` method consistent (used as readable name in model choice fields and in Django admin) - This makes all custom ModelChoiceFields obsolete because the `__str__()` method is used as default label for the model instances
    - Make output of `__repr__()` method consistent (designed for logging)
    - Register all models in Django admin

### DIFF
--- a/src/cms/admin.py
+++ b/src/cms/admin.py
@@ -1,68 +1,9 @@
 """
 File routing to the admin region
 """
-
-
+from django.apps import apps
 from django.contrib import admin
 
-from .models import ChatMessage
-from .models import Configuration
-from .models import Event
-from .models import EventTranslation
-from .models import Offer
-from .models import OfferTemplate
-from .models import Language
-from .models import LanguageTreeNode
-from .models import Organization
-from .models import Page
-from .models import PageTranslation
-from .models import POI
-from .models import POITranslation
-from .models import PushNotificationChannel
-from .models import Region
-from .models import RecurrenceRule
-from .models import UserMfa
-from .models import UserProfile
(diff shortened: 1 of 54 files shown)


src/cms/admin.py
```diff
@@ -1,68 +1,9 @@
...
-admin.site.register(MapFeedback)
-admin.site.register(OfferFeedback)
-admin.site.register(OfferListFeedback)
-admin.site.register(PageFeedback)
-admin.site.register(POIFeedback)
-admin.site.register(RegionFeedback)
-admin.site.register(SearchResultFeedback)
-admin.site.register(ImprintPage)
-admin.site.register(ImprintPageTranslation)
+for model in apps.get_app_config("cms").get_models():
```
svenseeberg, line 8: Good idea :+1:

ulliholtgrave (approved):
Looks good and seems to work accordingly :+1: :)

PR #743, opened 2021-03-17
Enable colored logging

timoludwig:
### Short description

I have no idea why I missed this opportunity in #738 (Improve logging) - I mean who doesn't want colors, right? 

### Proposed changes

- Add `ColorFormatter` which inherits from the default Python logging Formatter
- Colorize logging output based on logging level
- This only affects the console output which is used if `DEBUG` is `True`.

logging
### Resolved issues

Fixes the problem that our console log looks boring

### DIFF
--- a/src/backend/logging_formatter.py
+++ b/src/backend/logging_formatter.py
@@ -1,0 +1,35 @@
+import logging
+
+
+class ColorFormatter(logging.Formatter):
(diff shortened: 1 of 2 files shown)


ulliholtgrave (approved):
Delightful improvement :P

Issue #735, opened 2021-03-10
Improve page link field


timoludwig:
### Motivation

After merging PR #733 (Improve page form), the page link looks like this:

![Screenshot_2021-03-10 Integreat Redaktionssystem](https://user-images.githubusercontent.com/16021269/110677979-13329a80-81d6-11eb-9e80-4118f733a2b6.png)

Clicking on the edit-button enables the page link to be edited, otherwise it's disabled.

### Proposed Solution

I would love it if the page link looked similar to the short url as long as the user doesn't want to edit it:

![Screenshot_2021-03-08 Integreat Redaktionssystem-pre](https://user-images.githubusercontent.com/16021269/110366497-a8068e00-8046-11eb-88ad-2048e7b1d398.png)

Then, after clicking on the edit-button, a small form field should appear:

![Screenshot_2021-03-08 Integreat Redaktionssystem(1)-post](https://user-images.githubusercontent.com/16021269/110366521-ae950580-8046-11eb-970e-75072afc4d43.png)

After that, the user can enter a custom slug, click on the save-icon, which hides the form field again and updates the page link, or click on abort, which resets the form field and shows the old value again...

PR #744, opened 2021-03-17
Improve page link field

melegiul:
### Short description

Improve page link field 

### Proposed changes

- Per default page link field is read-only.
- Deactivating read-only displays a form field, where user can edit manually.
- After editing user can click the save button to confirm the changes (saves to the input field, changes only get persistent on form       submission)
- When clicking the abort button, the previous slug value is restored

### Resolved issues

Fixes: #735

### Screenshots

Screenshot from 2021-03-17 11-24-41

Screenshot from 2021-03-17 12-58-30

### DIFF
--- a/src/cms/templates/pages/page_form.html
+++ b/src/cms/templates/pages/page_form.html
@@ -182,26 +182,29 @@
                             </div>
                         {% endif %}
                     {% endif %}
-                    <label for="{{ page_translation_form.slug.id_for_label }}" class="block mb-2 font-bold">
-                        {{ page_translation_form.slug.label }}
-                        <button type="button" class="p-1 bg-gray-200 rounded ml-2"
-                            onclick='document.getElementById("id_slug").readOnly = false;'>
-                            <i data-feather="edit-2" class="inline-block"></i>
+                    <div id="link-container" class="flex items-center">
+                        <label for="{{ page_translation_form.slug.id_for_label }}" class="mr-2 font-bold">
+                            {{ page_translation_form.slug.label }}:
+                        </label>
+                        <a href="{{ page_link }}{{ page_translation_form.instance.slug }}" class="mr-2 text-blue-500 hover:underline">{{ page_link }}{{ page_translation_form.instance.slug }}</a>
+                        <div class="flex mr-2 hidden flex-grow">
+                            {% spaceless %}
+                            <span>{{ page_link }}</span>
+                            {% render_field page_translation_form.slug|add_error_class:"slug-error" class="flex-grow appearance-none rounded bg-gray-200" %}
+                            {% endspaceless %}
+                        </div>
+                        <button id="edit-slug-btn" type="button" class="mr-2 text-gray-800 hover:text-blue-500">
+                            <i data-feather="edit-3"></i>
                         </button>
-                    </label>
-                    <div class="my-2 text-s text-gray-600">{{ page_translation_form.slug.help_text }}</div>
-                    <div id="slug-div" class="appearance-none w-full bg-gray-200 text-xl text-gray-800 border border-gray-200 rounded py-3 px-4 leading-tight focus:outline-none focus:bg-white focus:border-gray-400">
-                        {% spaceless %}
-                            <div style="display: table; white-space: nowrap;">
-                                <span style="display: table-cell;">https://integreat.app/{{ region.slug }}/{{ language.slug }}/</span>
-                                {% if page_translation_form.instance.ancestor_path %}
-                                    <span style="display: table-cell;">{{ page_translation_form.instance.ancestor_path }}/</span>
-                                {% endif %}
-                                <span class="w-full table-cell">
-                                    {% render_field page_translation_form.slug|add_error_class:"slug-error" class="appearance-none w-full rounded" %}
-                                </span>
-                            </div>
-                        {% endspaceless %}
+                        <button id="copy-slug-btn" type="button" class="text-gray-800 hover:text-blue-500">
+                            <i data-feather="copy"></i>
+                        </button>
+                        <button id="save-slug-btn" type="button" class="mr-2 text-gray-800 hover:text-blue-500 hidden">
+                            <i data-feather="save"></i>
+                        </button>
+                        <button id="restore-slug-btn" type="button" class="text-gray-800 hover:text-blue-500 hidden">
+                            <i data-feather="x-circle"></i>
+                        </button>
                     </div>
                     <label for="{{ page_translation_form.text.id_for_label }}" class="block mb-2 mt-4 font-bold">{{ page_translation_form.text.label }}</label>
                     {% render_field page_translation_form.text class="bg-gray-200 w-full p-2 border border-gray-200 focus:outline-none focus:bg-white focus:border-gray-400 rounded tinymce_textarea" %}
--- a/src/cms/views/pages/page_view.py
+++ b/src/cms/views/pages/page_view.py
@@ -10,6 +10,8 @@ from django.utils.decorators import method_decorator
 from django.utils.translation import ugettext as _
 from django.views.generic import TemplateView
 
+from backend.settings import WEBAPP_URL
+
 from ...constants import status
 from ...decorators import region_permission_required
 from ...forms import PageForm, PageTranslationForm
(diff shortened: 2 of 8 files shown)


src/cms/views/pages/page_view.py
```diff
@@ -134,6 +135,9 @@ def get(self, request, *args, **kwargs):
         else:
             siblings = page.parent.children.all()
         context = self.get_context_data(**kwargs)
+        page_link = f"https://integreat.app/{region.slug}/{language.slug}/"
```
timoludwig, line 138:
```suggestion
        page_link = f"{WEBAPP_URL}/{region.slug}/{language.slug}/"
```

https://github.com/Integreat/integreat-cms/blob/3482ec00b5ef9c6393da4d5e5d3b0a2878f9c26e/src/backend/settings.py#L54

src/frontend/css/style.scss
```diff
@@ -542,9 +542,23 @@ to add animation */
...
 /* page form styling */
-#slug-div {
-  input:read-only {
-      --tw-bg-opacity: 1;
-      background-color: rgba(229, 231, 235, var(--tw-bg-opacity));
+#content_form {
+  #link-container {
+    div {
+      --tw-text-opacity: 1;
+      color: rgba(59, 130, 246, var(--tw-text-opacity));
```
timoludwig, line 549: I don't think it's a good idea to hardcode tailwind colors in the stylesheet. Why not directly toggling the class `text-blue-500`? I think the stylesheet should only be used if something is not achievable with tailwind.

src/frontend/js/copy-clipboard.ts
```diff
@@ -4,15 +4,7 @@ window.addEventListener("load", () => {
...
-      const tmpInput = document.createElement("input");
-      tmpInput.type = "text";
-      document.body.appendChild(tmpInput);
-      tmpInput.value = value;
-      tmpInput.select();
-      tmpInput.setSelectionRange(0, 99999);
-
-      document.execCommand("copy");
-      document.body.removeChild(tmpInput);
+      navigator.clipboard.writeText(value);
```
timoludwig, line 7: `navigator.clipboard` is [not supported on Internet Explorer](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard). @jnugh do you know if there is a way to tell Webpack to convert it in some way (if that's even possible)?

But today I learned that `document.execCommand("copy")` [is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand) and shouldn't be used either (even though it continues to work on all modern browsers at the moment - but this might change in the future)... :unamused:

src/frontend/js/forms/update-permalink.ts
```diff
@@ -16,12 +26,53 @@ window.addEventListener("load", () => {
...
+      } else {
+          // reset slug field to read-only and restore slug value
+          slugField.value = currentSlug;
+          replaceButtonIcon("edit-3", editButton.querySelector("svg"));
+          replaceButtonIcon("copy", copyButton.querySelector("svg"));
+          slugField.readOnly = true;
+          slugDiv.classList.toggle("editable");
+          feather.replace();
+      }
+  });
```
timoludwig, line 75: I would not use the same buttons for different actions. I think this makes it hard to maintain the source code in the future. Just use separate buttons and add the class "hidden" to the buttons which should not be displayed at the moment.

src/cms/templates/pages/page_form.html
```diff
@@ -182,27 +182,24 @@ <h2 class="heading font-normal">
...
-                                {% if page_translation_form.instance.ancestor_path %}
-                                    <span style="display: table-cell;">{{ page_translation_form.instance.ancestor_path }}/</span>
-                                {% endif %}
-                                <span class="w-full table-cell">
-                                    {% render_field page_translation_form.slug|add_error_class:"slug-error" class="appearance-none w-full rounded" %}
-                                </span>
-                            </div>
-                        {% endspaceless %}
                     </div>
+                    <div class="my-2 text-s text-gray-600">{{ page_translation_form.slug.help_text }}</div>
```
timoludwig, line 202: I think the help text can be omitted here. Since the slug is updated automatically when the title changes, there is no need for the case what happens when the field is empty. And the users will notice that illegal characters are filtered out once they submit the form with an invalid slug.

```suggestion
```

src/cms/templates/pages/page_form.html
```diff
@@ -182,27 +182,24 @@ <h2 class="heading font-normal">
...
+                        <div class="mr-2">
+                            {% spaceless %}
+                            <span>{{ page_link }}</span>
+                            {% render_field page_translation_form.slug|add_error_class:"slug-error" class="inline appearance-none rounded" size=page_translation_form.instance.slug|length %}
+                            {% endspaceless %}
+                        </div>
+                        <button id="edit-btn" type="button" class="mr-4">
+                            <i data-feather="edit-3"></i>
+                        </button>
+                        <button id="copy-btn" type="button">
```
timoludwig, line 198:
```suggestion
                        <button id="edit-btn" type="button" class="mr-4 text-gray-800 hover:text-blue-500">
                            <i data-feather="edit-3"></i>
                        </button>
                        <button id="copy-btn" type="button" class="text-gray-800 hover:text-blue-500">
```
I'd also add hover effects...

timoludwig (commented):
Amazing, I think it looks cool! :+1: 

One idea for improvement: You could also add separate spans which are toggled when the edit button is clicked. This way, the read-only mode could be a regular clickable link and there wouldn't be any weird spacing after the slug (and the slug would not be truncated if it exceeds the input field's width)... and the slug field wouldn't have to be manually set to read-only, it could just be hidden when the updated link is displayed... what do you think?

src/frontend/js/forms/update-permalink.ts
melegiul: I have to admit, that this "double functional" buttons caused some headache while coding :sweat_smile: 
As they have not really common properties except for their position in the form, we can of course donate two more buttons

src/frontend/css/style.scss
melegiul: So the main reason for me to involve this file was, that I couldn't find a tailwind feature for pseudo-class read-only, but this pseudo-class is just required for the input element (background toggling), but for the text color we can take tailwind as well

src/frontend/js/forms/update-permalink.ts
```diff
@@ -16,12 +24,59 @@ window.addEventListener("load", () => {
...
+  }
+
+
+  function updatePermalink(currentSlug: string){
+    // get complete permalink string
+    let currentLink = linkContainer.querySelector("a").innerHTML;
+    // remove trailing slug from link
+    let updatedLink = currentLink.substr(0, currentLink.lastIndexOf("/")+1);
+    // update only inner html and keep href until form submission
+    linkContainer.querySelector("a").innerHTML = updatedLink.concat(currentSlug);
```
timoludwig, line 53:
```suggestion
    linkContainer.querySelector("a").innerHTML = updatedLink.concat(currentSlug);
    linkContainer.querySelector("a").href = updatedLink.concat(currentSlug);
```
Can you also update the link `href` here? :wink:

melegiul:
> You could also add separate spans which are toggled when the edit button is clicked. This way, the read-only mode could be a regular clickable link

This is possible, but I recommend to keep the current href attribute, even if the user modifies title or slug manually to avoid displaying invalid clickable links (see [update-permalink](https://github.com/Integreat/integreat-cms/blob/ac35a2ddcc8b620e8ccbffaf9b8d3aa391c5906d/src/frontend/js/forms/update-permalink.ts#L47))

src/frontend/js/forms/update-permalink.ts
melegiul: I'm not sure if this is a good idea, we might break a link that was prior valid, when we update the href attribute before the form is submitted

timoludwig: Ah ok, yeah this makes sense to me! I also checked the current WordPress behavior and there it's the same: the link name is the new changed link and the link target is the old, valid link... let's keep it the way it is!

timoludwig (approved):
Awesome, I really like it! :+1:

src/frontend/js/copy-clipboard.ts
timoludwig: So just for reference: For the moment, we decided to keep the deprecated `document.execCommand("copy")` solution and hope that the Browser support will continue for a while (ideally until we don't need to support IE anymore)

PR #745, opened 2021-03-17
Apache2: Add app online ping rule

svenseeberg:
Add Apache2 config entry to return a 204 for the `/ping` URL. This enables the app to check its online status.


timoludwig (approved):
Looks good! :+1:

PR #747, opened 2021-03-17
Add clean parameter to install.sh 

svenseeberg:
### Short description
If `./node_modules` or `./src/cms/static/*` contain legacy JS files, this may result in Webpack compilation errors.

### Proposed changes
The `install.sh` removes the `./node_modules` and `./src/cms/static/*` directories.

### DIFF
--- a/sphinx/troubleshooting.rst
+++ b/sphinx/troubleshooting.rst
@@ -4,6 +4,14 @@ Troubleshooting
 
 .. highlight:: bash
 
+.. admonition:: General Advice
+
+    If you are experiencing problems of any kind, make sure you have a clean working environment by executing::
+
+        ./dev-tools/install.sh --clean
+        ./dev-tools/prune_database.sh
+
+    before trying anything else.
 
 Pipenv not found
 ================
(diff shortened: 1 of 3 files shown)


timoludwig (commented):
Could you add a small note to the documentation?
```
.. Note::

    Pass the flag ``--clean`` to remove old node modules and static files if your webpack compilation process does not succeed.
```
to the following sections?
https://github.com/Integreat/integreat-cms/blob/3482ec00b5ef9c6393da4d5e5d3b0a2878f9c26e/sphinx/installation.rst#L66-L77

https://github.com/Integreat/integreat-cms/blob/3482ec00b5ef9c6393da4d5e5d3b0a2878f9c26e/sphinx/dev-tools.rst#L9-L15

If you're highly motivated, you could even add a new section to the [troubleshooting guide](https://github.com/Integreat/integreat-cms/blob/develop/sphinx/troubleshooting.rst) :smile:

svenseeberg:
I think the installation section itself does not need any addition, as we should not complicate the initial installation procedure. However I added more info to the developer tools and troubleshooting sections.

sphinx/troubleshooting.rst
```diff
@@ -127,3 +127,18 @@ MacOS on M1
 
         pipenv run pip uninstall psycopg2-binary
         pipenv run pip install psycopg2-binary --no-cache-dir
+
+
+Webpack Compilation Errors
+==========================
+.. Error::
+
+    | ERROR in /path/to/integreat-cms/src/cms/static/@nodelib/...
```
timoludwig, line 136:
```suggestion
    | ERROR in /path/to/integreat-cms/src/cms/static/\@nodelib/...
```
Somehow, Sphinx thinks that this is a valid email address and converts it into a `mailto`-link :smile: 
So I suggest escaping the `@`...

timoludwig (commented):
Thanks, looks good :+1: 
Just found one minor issue.

sphinx/troubleshooting.rst
timoludwig: @svenseeberg can you include this change? After that, I think this PR can be merged.

timoludwig (approved):

Issue #751, opened 2021-03-22
Bold/italic font not displayed in PDF exports


melegiul:
### Describe the Bug

When the user adds bold/italic font styles to page, this styles are not visible in PDF exports

### Steps to Reproduce

1. Go to a Page Form and save some bold/italic styles to the page content
2. Export this page as PDF
3. Notice that this styles are not visible in the document

### Expected Behavior

The PDF document should display user defined font styles

### Actual Behavior

PDF document displays only normal font weight

PR #752, opened 2021-03-22
Fix PDF font style bug

melegiul:
### Short description

Fixes bold/italic font styles in PDF export

### Proposed changes

- The current font for PDF generation is DejaVu, which is a static
  font-family. That means different font wheights and styles are shiped
  in different files (see folder `cms.frontend.fonts.dejavu-sans`).
- xhtml2pdf + DejaVu accept arabic letters, but not in combination with
  italic style, so italic styles are excluded from arabic documents.

### Resolved issues

Fixes: #751


timoludwig (approved):
Looks good to me! :+1:

Issue #753, opened 2021-03-22
Non-existing page translation PDF export bug


melegiul:
### Describe the Bug

When user selects a non existing page translation for PDF export, there comes up an uncaught KeyError exception.

### Steps to Reproduce

1. Go to page tree
2. Click on language flag with non-existing translation, e.g. arabic
3. Set checkbox for this translation
4. Select PDF export

### Expected Behavior

Nothing should happen due to missing translation
Alternative: pop up dialog to inform user about non existing page translation

### Actual Behavior

KeyError exception occurs

timoludwig:
Since this only happens when only non-existing translation are selected (as long as there is one existing translation, the non-existing ones just get filtered out and the PDF is generated successfully), I propose to show an error message in this case.

Even though it's not ideal from a UX perspective to open the new tab just for showing an error message, I think in this case it's acceptable because I see no simple way around it and I don't think we should build a complex solution for this rare edge case... just add the error message in the PDF export action and redirect to the page tree in the new tab.

melegiul:
I had just a deeper look at the code and concerning UX perspective I could offer also a third alternative:
This function [toggleBulkActionButton](https://github.com/Integreat/integreat-cms/blob/12b0c12ed86a7aad8f059c12fea51f712ca90c34/src/frontend/js/bulk-actions.ts#L66) is used to toggle the buttons disabled property, so I could add a function that compares some DOM elements (like [no-trans-class](https://github.com/Integreat/integreat-cms/blob/12b0c12ed86a7aad8f059c12fea51f712ca90c34/src/cms/templates/pages/page_tree_node.html#L83)) and disable the button if need.
I think the complexity would be not very high, what you think about this idea?

PR #754, opened 2021-03-22
Fixes non-existing translation PDF export

melegiul:
### Short description

Fixes KeyError exception when only non-existing page translations are selected for PDF export

### Proposed changes

- Adds a function to check, whether at least one translation exists for
  the selected pages.
- If this is the case, PDF export for the selection is allowed else
  denied.

### Resolved issues

Fixes: #753

### Additional context

This new function should enable both proposed approaches: User dialog or disabled bulk-action-button.

### DIFF
--- a/src/frontend/js/bulk-actions.ts
+++ b/src/frontend/js/bulk-actions.ts
@@ -65,11 +65,14 @@ window.addEventListener("load", () => {
 
   function toggleBulkActionButton() {
     // Only activate button if at least one item and the action is selected
+    // also check if at least one page translation exists for PDF export before activation
+    let selectedAction = bulkAction.options[bulkAction.selectedIndex];
     if (
       !selectItems
         .filter(isInputElement)
         .some((el) => el.checked) ||
-      bulkAction.selectedIndex === 0
+      bulkAction.selectedIndex === 0 ||
+      (selectedAction.id === "pdf-export-option" && !hasTranslation())
     ) {
       bulkActionButton.classList.remove(
         "bg-blue-500",
@@ -102,5 +105,18 @@ window.addEventListener("load", () => {
     }
     // Submit form and execute bulk action
     form.submit();
+  }
+
+  function hasTranslation(): boolean {
+    // checks if at least one of the selected pages has a translation for the current language
+    let languageSlug = document.getElementById("pdf-export-option").dataset.languageSlug;
+    return selectItems
+      .filter(isInputElement)
+      .filter(inputElement => inputElement.checked)
+      .some(inputElement => {
+        return inputElement
+          .closest("tr")
+          .querySelector(`.lang-grid .${languageSlug} .no-trans`) === null;
+      });
   }
 });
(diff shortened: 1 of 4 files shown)


src/frontend/js/bulk-actions.ts
```diff
@@ -103,4 +106,29 @@ window.addEventListener("load", () => {
...
+      let translationState = translation
+        .closest("tr")
+        .querySelector('.lang-grid')
+        .querySelector(`.${languageSlug}`);
+      if (translationState.querySelector(".no-trans") === null)
+        // if at least one page does not contain a no-trans-class for this language
+        // then there exists a translation for the given page-language-combination
+        return true;
+    }
+    return false;
```
timoludwig, line 132:
```suggestion
    // check if at least one input element has a translation
    return selectItems
      .filter(isInputElement)
      .filter(inputElement => inputElement.checked)
      .some(inputElement => {
        return inputElement
          .closest("tr")
          .querySelector(`.lang-grid .${languageSlug} .no-trans`) === null;
      });
```
This can be simplified a little bit...

timoludwig (commented):
Cool idea, yes I think it is better if the PDF action is not even triggered when no translation exist! :+1:

src/frontend/js/bulk-actions.ts
```diff
@@ -103,4 +106,29 @@ window.addEventListener("load", () => {
...
   }
+
+  function hasTranslation(): boolean {
+    // checks if at least one of the selected pages has a translation for the current language
+    let urlPath = window.location.pathname;
+    // remove trailing slash from url path
+    urlPath = urlPath.replace(/\/$/, '');
+    // extract current language slug from url path
+    let languageSlug = urlPath.substr(urlPath.lastIndexOf('/') + 1);
+    // retrieve all checked input elements
```
timoludwig, line 117: This seems kind of hacky to me. Can you maybe add another `data`-attribute for the language slug?

src/frontend/js/bulk-actions.ts
melegiul: I like this `some()` function, I didn't know it yet :rocket:

src/frontend/js/bulk-actions.ts
melegiul: Ok, data attribute is added

timoludwig (commented):
Perfect :+1:

One thing I noticed: We already have a server side error message for this case:
https://github.com/Integreat/integreat-cms/blob/def8a2ce24764c7be47f75dc7708ccabb4b2d55b/src/cms/utils/pdf_utils.py#L64-L67

But this doesn't get shown because in the action `export_pdf()`, the view expects the `Content-Disposition` header to be set. Even though this shouldn't happen anymore due to the new check in the frontend, I think it would be good to fix the bug in the backend as well...
Could you maybe add:
```
if response.status_code == 200:
```
in front of the following line:
https://github.com/Integreat/integreat-cms/blob/def8a2ce24764c7be47f75dc7708ccabb4b2d55b/src/cms/views/pages/page_actions.py#L261
Thanks!

melegiul:
That's a good idea, this will prevent the exception raising again, when we decide to exchange the frontend once

timoludwig (approved):
:+1:

Issue #703, opened 2021-02-23
Improve region form


timoludwig:
### Proposed Solution

- Add new field "Include administrative division in region name" which determines whether the name is listed as e.g. "Augsburg" or "Stadt Augsburg" in the apps (sorting is always by the name, independently from the administrative division"
- Rename "Gemeindeverband" to "Verbandsgemeinde"

PR #755, opened 2021-03-23
Improved region form

melegiul:
### Short description

Improves region form

### Proposed changes

- Renamed "Gemeindeverbande" to "Verbandsgemeinde"
- Added field "administrative division included" to include adm. div.
  into region name
- Updated PDF documents according to aforementioned field
### Resolved issues

Fixes: #703

### DIFF
--- a/src/cms/models/regions/region.py
+++ b/src/cms/models/regions/region.py
@@ -158,6 +158,19 @@ class Region(models.Model):
         ),
     )
 
+    administrative_division_included = models.BooleanField(
+        default=False,
+        verbose_name=_("include administrative division into name"),
+        help_text=__(
+            _(
+                "Determines whether the administrative division is displayed next to the region name."
+            ),
+            _(
+                "Sorting is always based on the name, independently from the administrative division."
+            ),
+        ),
+    )
+
     @property
     def languages(self):
         """
--- a/src/cms/templates/regions/region_form.html
+++ b/src/cms/templates/regions/region_form.html
@@ -40,8 +40,16 @@
                             <img src="{% static 'svg/select-down-arrow.svg' %}" class="fill-current h-4 w-4" />
                         </div>
                     </div>
+
+                    <!-- Option for including/excluding administrative devision from region name -->
+                    <div class="p-2 mt-1">
+                        {% render_field form.administrative_division_included %}
+                        <label for="{{ form.administrative_division_included.id_for_label }}" class="font-bold cursor-pointer p-2">{{ form.administrative_division_included.label }}</label>
+                        <div class="my-2 text-s text-gray-600">{{ form.administrative_division_included.help_text }}</div>
+                    </div>
+
                     <!-- General Options for region management -->
-                    <label for="{{ form.name.id_for_label }}" class="font-bold block p-2 mb-1 mt-5 cursor-pointer">{{ form.name.label }}</label>
+                    <label for="{{ form.name.id_for_label }}" class="font-bold block p-2 mb-1 cursor-pointer">{{ form.name.label }}</label>
                     {% render_field form.name class="appearance-none block w-full bg-gray-200 text-xl text-gray-800 border border-gray-200 rounded py-3 px-4 leading-tight focus:outline-none focus:bg-white focus:border-gray-400" %}
 
                     <label for="{{ form.common_id.id_for_label }}" class="font-bold block p-2 mb-1 mt-5 cursor-pointer">{{ form.common_id.label }}</label>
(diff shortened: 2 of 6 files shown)


src/cms/models/regions/region.py
```diff
@@ -158,6 +158,15 @@ class Region(models.Model):
         ),
     )
 
+    administrative_division_included = models.BooleanField(
+        default=False,
+        verbose_name=_("include administrative division into name"),
+        help_text=_(
+            "Determines whether the administrative division is displayed next to the region name.\
+            Sorting is always based on the name, independently from the administrative division."
```
timoludwig, line 166:
```suggestion
        help_text=__(
            _("Determines whether the administrative division is displayed next to the region name."),
            _("Sorting is always based on the name, independently from the administrative division."),
```
I think using the lazy formatting method `ugettext_many_lazy()` makes it a bit more readable...

src/cms/templates/regions/region_form.html
```diff
@@ -154,11 +154,19 @@ <h3 class="heading font-bold text-white">{% trans 'Extended Settings' %}</h3>
...
                     <div class="my-2 text-s text-gray-600">{{ form.page_permissions_enabled.help_text }}</div>
                 </div>
+
+                <!-- Option for including/excluding administrative devision from region name -->
+                <span class="block p-2 mt-2 text-xl font-bold">{% trans 'Region name option' %}</span>
+                <div class="p-2">
+                    {% render_field form.administrative_division_included %}
+                    <label for="{{ form.administrative_division_included.id_for_label }}" class="font-bold cursor-pointer p-2">{{ form.administrative_division_included.label }}</label>
+                    <div class="my-2 text-s text-gray-600">{{ form.administrative_division_included.help_text }}</div>
+                </div>
```
timoludwig, line 169: I'd propose to move this field beneath the `administrative_division` field...

src/cms/models/regions/region.py
```diff
@@ -158,6 +158,15 @@ class Region(models.Model):
         ),
     )
 
+    administrative_division_included = models.BooleanField(
+        default=False,
+        verbose_name=_("include administrative division into name"),
```
timoludwig, line 163:
```suggestion
        verbose_name=_("prepend administrative division to name"),
```
I think this would make it more clear the the administrative division goes before the name...
(German translation: "Verwaltungseinheit dem Namen voranstellen")

timoludwig (commented):
Looks good! :+1:

I just have a few small ideas for improvement.

src/cms/models/regions/region.py
melegiul: Thank you for your suggestion, but actually I'm fine with include, so I want to keep it

src/cms/templates/regions/region_form.html
melegiul: Yes, you are right, at this position it's more reasonable :+1:

src/cms/templates/regions/region_form.html
```diff
@@ -40,6 +40,14 @@ <h3 class="heading font-bold text-white">{% trans 'General Settings' %}</h3>
...
+
+                    <!-- Option for including/excluding administrative devision from region name -->
+                    <div class="p-2 mt-1">
+                        {% render_field form.administrative_division_included %}
+                        <label for="{{ form.administrative_division_included.id_for_label }}" class="font-bold cursor-pointer p-2">{{ form.administrative_division_included.label }}</label>
+                        <div class="my-2 text-s text-gray-600">{{ form.administrative_division_included.help_text }}</div>
+                    </div>
+
                     <!-- General Options for region management -->
                     <label for="{{ form.name.id_for_label }}" class="font-bold block p-2 mb-1 mt-5 cursor-pointer">{{ form.name.label }}</label>
```
timoludwig, line 52:
```suggestion
                    <label for="{{ form.name.id_for_label }}" class="font-bold block p-2 mb-1 cursor-pointer">{{ form.name.label }}</label>
```
Now I think this spacing is a little bit too big.

src/cms/templates/regions/region_form.html
```diff
@@ -154,12 +162,12 @@ <h3 class="heading font-bold text-white">{% trans 'Extended Settings' %}</h3>
 
                 <!-- Options related to page-instance-specific permissions -->
                 <span class="block p-2 mt-2 text-xl font-bold">{% trans 'Page permissions' %}</span>
-                <div class="p-2">
+                <div class="p-2 pb-4 border-b solid border-gray-400">
```
timoludwig, line 165:
```suggestion
                <div class="p-2">
```
And I'd revert this bottom border because it's the last entry again.

timoludwig (approved):
:+1:

PR #756, opened 2021-03-24
Improve handling of administrative division

timoludwig:
### Short description

In issue #703, I forgot to mention that the new field `administrative_division_included` should also determine the name of the regions in API responses. This PR now fixes this addendum.

### Proposed changes

- Show administrative division in API only when included in the name
- Add `prefix` and `full_name` properties to `Region` model
- Return translation of administrative division always in the region's default language


melegiul (approved):
Very clean and concise method, I like it :rocket:

Issue #697, opened 2021-02-23
Change user creation workflow


timoludwig:
### Proposed Solution

- Add option to create user without defining password
- User gets email with activation link
- On the first visit, the user can define the password
- User gets activated when password is set
- Show current activation status in user list and make it possible to re-send invitation email directly from user list

### Additional Context

Is this an additional feature or already required for feature completion?

PR #761, opened 2021-03-30
Update user creation workflow

melegiul:
### Short description

Add option to create inactive user without setting the password.
User gets activation link per email to set the password and activate the account.

### Proposed changes

- User and region-user form extended with activation link option.
- Setting the options disables password-field and active-field to create inactive account without password.
- Sends activation link to user (changed email field to required).
- This PR reuses the existing template of the cms app for the password reset workflow extended by custom title and error message.
- Also uses the built-in view/form of the django authentication app extended to activate user accounts after password was  successfully set.
- Option to resend the activation link to user within the user/region-user menu, only available for inactive accounts.

### Additional

When a user is saved with send_activation_link enabled, an inactive account is saved.
The password and active field for such accounts are hidden at the user form, to avoid manual interaction until the user has clicked the link.

### Resolved issues

Fixes: #697


src/cms/views/users/region_user_view.py
```diff
@@ -94,6 +94,9 @@ def post(self, request, *args, **kwargs):
             request.POST, instance=user_profile_instance
         )
 
+        if user_profile_instance and user_profile_instance.send_activation_link:
```
timoludwig, line 97:
```suggestion
        if user_instance and user_instance.active:
```
If you go with my suggestion of removing the `send_activation_link` field from the database, you can also use the `active` field here.

src/cms/views/users/user_view.py
```diff
@@ -101,22 +102,40 @@ def post(self, request, *args, **kwargs):
...
-                if user_form.has_changed() or user_profile_form.has_changed():
-                    if user_instance:
-                        messages.success(request, _("User was successfully saved"))
+                # check if either activation link is send or password is set or user already exists
+                if (
+                    user_profile_form.cleaned_data["send_activation_link"]
+                    or (
+                        not user_profile_form.cleaned_data["send_activation_link"]
+                        and user_form.cleaned_data["password"]
+                    )
```
timoludwig, line 111:
```suggestion
                    or user_form.cleaned_data["password"]
```
If an if-block has multiple `OR` conditions, Python will go into the if-block as soon as it sees a condition which evaluates to `True`. This means that on every subsequent condition, you can be sure that the first condition evaluates to `False` and you don't have to test for it...

src/cms/urls.py
```diff
@@ -303,6 +312,27 @@
...
+    url(
+        r"^activation/",
+        include(
+            [
+                url(
+                    r"^(?P<uidb64>[0-9A-Za-z]+)/(?P<token>.+)/$",
+                    auth_views.PasswordResetConfirmView.as_view(
+                        template_name="authentication/account_activation_form.html",
+                        success_url=reverse_lazy("activation_complete"),
+                        form_class=ActivationForm,
```
timoludwig, line 324: Cool idea to use the inbuilt view for this!

dev-tools/generate_documentation.sh
```diff
@@ -31,6 +31,9 @@ fi
 
 cd $(dirname "${BASH_SOURCE}")/..
 
+# remove stale doc files
+rm -rf ./docs ./sphinx/ref ./sphinx/ref-ext
```
timoludwig, line 35:
```suggestion
# remove stale doc files
if [ "$1" == "--clean" ]; then
    echo "Removing temporary documentation files."
    rm -rf ./docs ./sphinx/ref ./sphinx/ref-ext
fi
```
This makes the doc generation process significantly slower, so I would only perform this step if absolutely necessary...
Maybe it would also be good to add information about the new `--clean` flag to the documentation itself? But I can also do this in a separate PR.

timoludwig (commented):
Thanks, really cool feature! :rocket:
One step further towards feature completion :smile: 

Apart from my comments, I have one additional suggestion: I think it would be good if the activation-workflow would be the default, so that the checkbox is pre-selected and the password-field is hidden by default. Just if an admin actively deactivates this option, it should be possible to set the password on account creation.

src/cms/forms/authentication/activation_form.py
```diff
@@ -0,0 +1,22 @@
...
+    def save(self, commit=True):
+        """Overrides save method to additionally activate inactive accounts
+
+        :return: the user instance that was saved
+        :rtype: ~django.contrib.auth.models.User
+        """
+        super().save(commit)
+        self.user.is_active = True
+        self.user.save()
+        self.account_activation_done.send(sender=self.__class__, user_id=self.user.id)
```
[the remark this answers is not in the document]
melegiul: No there was no specific reason and I won't never reject the zen masters advice :relaxed:

dev-tools/generate_documentation.sh
melegiul: That's ok for me, this command just needed execution for a single time on the remote repo to fix a doc generation exception

melegiul:
I was just trying to follow your suggestion about removing the database field `send_activation_link`, but finally I see there a problem arising: removing this field would not allow to differ between user that are currently in the account activation process and users that were deactivated on purpose, like users that leave the organization, when we want to deny access but keep the user data. So when you create a user entry with `send_activation_link` enabled without corresponding model field, later we can not hide the password field and the active field, because we can't tell whether this account should be activated soon by the user himself or whether this is a retired user account. So I think this would just lead to a confusing situation, where inactive user accounts can have multiple reasons and this reasons are not transparent to the person, who maintains the user list. So I would vote for keeping this `send_activation_link` in the database which gives is a reliable information, whether the user is currently stuck in the activation process or just an inactive user. What you think about this?

melegiul:
> I was just trying to follow your suggestion about removing the database field `send_activation_link`, but finally I see there a problem arising: removing this field would not allow to differ between user that are currently in the account activation process and users that were deactivated on purpose, like users that leave the organization, when we want to deny access but keep the user data. So when you create a user entry with `send_activation_link` enabled without corresponding model field, later we can not hide the password field and the active field, because we can't tell whether this account should be activated soon by the user himself or whether this is a retired user account. So I think this would just lead to a confusing situation, where inactive user accounts can have multiple reasons and this reasons are not transparent to the person, who maintains the user list. So I would vote for keeping this `send_activation_link` in the database which gives is a reliable information, whether the user is currently stuck in the activation process or just an inactive user. What you think about this?
> But I also should rename `send_activation_link` to `account_activation_in_progress` to avoid misunderstanding of its purpose.

For reference: we decided to not distinguish between user with pending activation link and other inactive users

src/cms/templates/users/admin/list_row.html
```diff
@@ -47,4 +49,17 @@
...
+				title="{% trans 'Resend activation link' %}">
+				<i data-feather="send" class="inline-block text-gray-800"></i>
+			</a>
+		{% else %}
+			<a href="#"
+				title="{% trans 'You cannot send activation links to active users' %}">
+				<i data-feather="send" class="inline-block text-gray-400"></i>
+			</a>
+		{% endif %}
+	</td>
```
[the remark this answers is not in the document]
melegiul: This is already done altogether in 253d686f325d75ed58d8814db3b1d82175a67993 with the code snippet below your suggestion

src/cms/urls.py
melegiul: Inspired by the password reset process:smiley:

src/cms/views/authentication/authentication_actions.py
```diff
@@ -270,9 +271,31 @@ def password_reset_complete(request):
...
+    View to indicate that the account was successfully activated
+
+    :param request: The current request
+    :type request: ~django.http.HttpRequest
+
+    :return: Redirect to login page after activation
+    :rtype: ~django.http.HttpResponseRedirect
+    """
+
+    logger.info("Account activation for user %r was successful", request.user.profile)
```
timoludwig, line 293:
```suggestion
```
This will raise the error `'AnonymousUser' object has no attribute 'profile'` if the user is not authenticated while he is setting his password (which will almost always be the case). Also, there is no guarantee that the currently logged in user is the same user as the one being activated. So I'd propose to move this logging message into the `save()`-method of the `ActivationForm`.

src/cms/utils/account_activation_utils.py
```diff
@@ -0,0 +1,48 @@
...
+
+    Making use of ~django.core.mail.send_mail to prevent malicious header injections
+    by raising an ~django.core.mail.BadHeaderError exception
+    if a newline character is detected in subject, recipient_list, etc.
+
+    :param request: The current http request
+    :type request: ~django.http.HttpRequest
+
+    :param user_instance: the users profile to send the email
+    :type user_instance: ~cms.models.users.user_profile.UserProfile
```
timoludwig, line 23:
```suggestion
    :param user_instance: The user to whom the e-mail is to be sent
    :type user_instance: ~django.contrib.auth.models.User
```
I think this is the user model and not the user profile model, right?

src/cms/templates/users/region/list_row.html
```diff
@@ -26,11 +28,13 @@
...
 	</td>
-	<td class="pl-2">
+	<td class="px-2">
 		{% if user.is_active %}
 		<i data-feather="check"></i>
 		{% else %}
-		<i data-feather="x"></i>
+		<a href="{% url 'resend_activation_link_region' region_slug=region.slug user_id=user.pk %}" class="hover:underline text-blue-500 hover:text-blue-600 whitespace-nowrap">
+			{% trans 'Resend activation link' %}
+		</a>
```
timoludwig, line 37:
```suggestion
			<i data-feather="x"></i>
			<a href="{% url 'resend_activation_link_region' region_slug=region.slug user_id=user.pk %}" class="hover:underline text-blue-500 hover:text-blue-600">
				{% trans 'Resend activation link' %}
			</a>
```
I think it would be cool to have the x-icon together with the resend-link...

src/cms/templates/users/region/list_row.html
```diff
@@ -26,11 +28,13 @@
 			{% endif %}
         </a>
 	</td>
-	<td class="pl-2">
+	<td class="px-2">
```
timoludwig, line 31:
```suggestion
	<td class="px-2 whitespace-nowrap">
```

src/cms/templates/users/admin/list_row.html
```diff
@@ -28,9 +30,11 @@
 	</td>
 	<td class="pl-2">
```
timoludwig, line 31:
```suggestion
	<td class="px-2 whitespace-nowrap">
```

src/cms/templates/users/admin/list_row.html
```diff
@@ -28,9 +30,11 @@
...
 		{% if user.is_active %}
-		<i data-feather="check"></i>
+			<i data-feather="check"></i>
 		{% else %}
-		<i data-feather="x"></i>
+		<a href="{% url 'resend_activation_link' user_id=user.pk %}" class="hover:underline text-blue-500 hover:text-blue-600 whitespace-nowrap">
+			{% trans 'Resend activation link' %}
+		</a>
```
timoludwig, line 37:
```suggestion
			<i data-feather="x"></i>
			<a href="{% url 'resend_activation_link' user_id=user.pk %}" class="hover:underline text-blue-500 hover:text-blue-600">
				{% trans 'Resend activation link' %}
			</a>
```

timoludwig (commented):
Thanks! :+1:
I have a few more comments, but most of them are just minor details.

src/cms/views/authentication/authentication_actions.py
melegiul: Interesting, I never saw this exception because I always was logged in as root, good point :+1:

src/cms/views/users/user_view.py
```diff
@@ -101,22 +100,40 @@ def post(self, request, *args, **kwargs):
...
-                    messages.info(request, _("No changes detected."))
+                    messages.error(
+                        request,
+                        _(
+                            "Please choose either to send an activation link or set a password."
+                        ),
+                    )
             else:
                 messages.error(
                     request,
```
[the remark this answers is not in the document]
melegiul: This is done in d9428312651c445d14ba75b565f3991e15a49416

timoludwig (approved):
Thanks! :+1:
After squashing all commits, this PR is ready to merge! :rocket:

Issue #768, opened 2021-04-13
User creation not working


timoludwig:
### Describe the Bug

Due to a problem in the user form's logging, an error occurs when new users are created and a new role is assigned in the creation process

### Steps to Reproduce

1. Go to the user management
2. Create new user and select at least one role

### Expected Behavior

The user including a corresponding user profile should be created

### Actual Behavior

An error occurs and the user is created without a user profile
```
RelatedObjectDoesNotExist at /augsburg/users/new

User has no profile.
```
### Additional Information

<details>

Environment:

Request Method: POST
Request URL: http://localhost:8000/augsburg/users/new

Django Version: 3.1.8
Python Version: 3.7.10
Installed Applications:
['cms.apps.CmsConfig',
 'gvz_api.apps.GvzApiConfig',
 'django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.messages',
 'django.contrib.sessions',
 'django.contrib.sitemaps',
 'django.contrib.staticfiles',
 'corsheaders',
 'widget_tweaks',
 'easy_thumbnails',
 'filer',
 'mptt',
 'rules.apps.AutodiscoverRulesConfig']
Installed Middleware:
['corsheaders.middleware.CorsMiddleware',
 'django.middleware.security.SecurityMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.locale.LocaleMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'cms.middleware.timezone_middleware.TimezoneMiddleware']

Traceback (most recent call last):
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.7/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.7/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.7/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
[...]

PR #769, opened 2021-04-13
Fix user creation

timoludwig:
### Short description

 Fix `RelatedObjectDoesNotExist` error in user form

### Proposed changes

- Only use `user.profile` in logging if user already has a profile

### Resolved issues

Fixes: #768


ulliholtgrave (approved):
I tested it and this seems to eliminate the error :)

PR #770, opened 2021-04-14
Add table sorting 

emmakemb:
### Short description

At the moment, only the event list can be sorted.
Proposed Solution 
### Proposed changes

Add the same tablesort functionality to all our tables:
- Regions
- Languages
- Admin Users
- Region Users
- Organizations
- Offer Templates
- Admin Feedback
- Region Feedback
- Push Notifications
 
### Resolved issues

Fixes: #624

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "integreat-cms",
       "version": "1.0.0",
       "license": "Apache-2.0",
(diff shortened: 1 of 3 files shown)


timoludwig:
@mael-fosso @emmakemb I don't understand why you opened a PR if you have no intention of contributing even a single line of code that makes sense... Simply writing "sorted" under a heading does not make a table sorted, this doesn't make sense on so many levels...

PR #771, opened 2021-04-14
Add node-js requirements to install script

timoludwig:
### Short description

Since our new webpack bundling process has new constraints to our node-js version, I added the requirements check to the install.sh script.

### Proposed changes

- Check if nodejs version 12 or version 14 or higher is installed


ulliholtgrave (approved):
Looks good :)

PR #773, opened 2021-04-15
Check write permissions of logfile

timoludwig:
### Short description

This enables the possibility to start the server in production mode, even if no writable logfile location is given.

### Proposed changes

- Check write permissions before falling back to the default logfile location


ulliholtgrave (approved):
Looks reasonable :+1:

Issue #746, opened 2021-03-17
Only activate django.contrib.admin in DEBUG mode


timoludwig:
### Motivation

Changes made in the Django admin console are not validated with the rules defined in our forms and views. This means, incautious actions can lead to an inconsistent database, which should not be possible on a production system.

### Proposed Solution

- Only add `"django.contrib.admin"` to the `INSTALLED_APPS` in [src/backend/settings.py](https://github.com/Integreat/integreat-cms/blob/develop/src/backend/settings.py) if `DEBUG` is `False`.
- Only include `admin.site.urls` in [src/backend/urls.py](https://github.com/Integreat/integreat-cms/blob/develop/src/backend/urls.py) if `"django.contrib.admin"` is in `INSTALLED_APPS`
- Only show [the link to Django Admin](https://github.com/Integreat/integreat-cms/blob/3482ec00b5ef9c6393da4d5e5d3b0a2878f9c26e/src/cms/templates/_base.html#L60-L63) in the menu bar if `"django.contrib.admin"` is in `INSTALLED_APPS`

### Alternatives

Generalize the validation so it applies directly to the Models instead of the ModelForms, use the signals `pre_save` and `post_save` to handle constraints between models (e.g. there should not be a Page instance without translations etc.).
This might get unnecessarily complex so I propose to deactivate Django admin instead.

Issue #689, opened 2021-02-22
Improve help button


timoludwig:
### Motivation

At the moment, the help button is poorly placed and doesn't serve its purpose.

### Proposed Solution

- Move help button directly in top menu bar (not in the user submenu)
- Add new setting `WIKI_URL` to settings.py for the url to the integreat wiki
- Add link to `WIKI_URL` when clicking on the help button

### Additional Context

![image_2021-02-22_17-00-04](https://user-images.githubusercontent.com/16021269/108734176-7adabb80-752f-11eb-8fa5-ae20793a0ae7.png)

PR #774, opened 2021-04-15
Add functionality to the help button & Deactivate Django.Admin in production

ulliholtgrave:
### Short description
This PR introduces a global setting for the wiki-URL and deactivates the django.contrib.admin app, if the system is run on production mode.

### Proposed changes
- Introduce WIKI_URL in settings.py
- Create a redirect to the Wiki
- Only add django.contrib.admin to INSTALLED_APPS, if `DEBUG=True`
- Only add `/admin` endpoint to `urls.py` if `DEBUG=True`
- Only show Django-Administration Link in base.html, if debug is true.

### Resolved issues

Fixes: #746
Fixes: #689

### DIFF
--- a/src/backend/settings.py
+++ b/src/backend/settings.py
@@ -76,6 +76,9 @@ BLOG_URLS = {
     "de": f"{WEBSITE_URL}/blog/",
 }
 
+#: URL to the Integreat wiki
+WIKI_URL = "https://wiki.integreat-app.de"
+
 #: RSS feed URLs to the Integreat blog
 RSS_FEED_URLS = {
     "en": f"{WEBSITE_URL}/en/feed/",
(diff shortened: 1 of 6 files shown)


timoludwig (commented):
Thanks! :+1: 
In general, this works. However, I get an error because `admin.site.register(model)` in [src/cms/admin.py](https://github.com/Integreat/integreat-cms/blob/06b29d05984b6fc3ccf40b6f3275a48a2c862d5e/src/cms/admin.py#L9) fails.
I think the following lines should prevent that:
```
from django.apps import apps
from django.contrib import admin


if settings.DEBUG:
    for model in apps.get_app_config("cms").get_models():
        admin.site.register(model)
```

timoludwig (approved):
Looks good to me! :+1:

Issue #775, opened 2021-04-16
Page Hierachie cannot be changed via page form


ulliholtgrave:
### Describe the Bug
If some top-level page is selected in the Page Form and saved, nothing changed.

### Steps to Reproduce

1. Go to a Page Form
2. Select a top-level page in the right menu.
3. Click on save
4. Return to Hierarchy-View and see no changes.

### Expected Behavior
The selected top-level page should be saved.

### Actual Behavior
The system informs about saved changes, but nothing is changed.

### Additional Information
The hierarchy can actually be changed via the drag-and-drop on the page-overview page.

PR #779, opened 2021-04-16
Fix event handler of page order table

timoludwig:
### Short description

Initializes the event listener on window load, not on window change

### Resolved issues

Fixes: #775


ulliholtgrave (approved):
Works again :)

Issue #784, opened 2021-04-19
Page not found error at page form


melegiul:
### Describe the Bug

Modifying the page order throws error (page form)

### Steps to Reproduce

1. Go to the page form
2. Use the drag and drop feature to move a page (Settings of the page -> Positioning -> Order)
3. See error "Page not found" 404

### Expected Behavior

The page order should be changed

### Actual Behavior

The error occurs

### Additional Information

![Screenshot from 2021-04-19 11-12-51](https://user-images.githubusercontent.com/47010475/115212136-7df1c280-a100-11eb-9937-3ca2683ba916.png)

PR #785, opened 2021-04-19
Fix page drag & drop

timoludwig:
### Short description

Both drag & drop scripts were working concurrently, because both set event listeners to the same class names.

### Proposed changes

- Only activate tree drag & drop on page & language tree
- Only activate page order drag & drop on page form
- For tree drag & drop: Move function definitions from inside the event listener to the global scope of the script

### Resolved issues

Fixes: #784


melegiul (approved):
Thank you, that solves the problem :rocket:

PR #787, opened 2021-04-22
Dev Tools: use python3.7

svenseeberg:
### Short description
Fix python3 version to 3.7 in dev-tools scripts.

### Resolved issues
Fixes not working dev-tools scripts on machines where the default python3 version is not 3.7.

### DIFF
--- a/dev-tools/install.sh
+++ b/dev-tools/install.sh
@@ -48,8 +48,8 @@ node_numeric_version="${node_version:1}"
 # Strip trailing minor and patch version
 node_major_version="${node_numeric_version%%.*}"
 # Check node version requirements (12 or higher but not 13)
-if [ "${node_major_version}" -lt 12 ] || [ "${node_major_version}" -eq 13 ]; then
-    echo "nodejs version 12 or version 14 or higher is required, but version $node_version is installed. Please install a recent version manually and run this script again." >&2
+if ! [[ "${node_major_version}" =~ ^(12|14|15)$ ]] ; then
+    echo "nodejs version 12 or version 14 or 15 is required, but version $node_version is installed. Please install a supported version manually and run this script again." >&2
     exit 1
 fi
 
--- a/dev-tools/install.sh
+++ b/dev-tools/install.sh
@@ -85,7 +85,7 @@ else
 fi
 
 # Install pip dependencies
-pipenv install --dev
+pipenv --python 3.7 install --dev
 
 # Install pre-commit hook for black code style
-pipenv run pre-commit install
+pipenv --python 3.7 run pre-commit install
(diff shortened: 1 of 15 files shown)


timoludwig:
I'm not sure if that's the correct way to go. 
`pipenv run` automatically chooses the correct python version which is specified in the Pipfile and the virtual environment:
```
> python --version
Python 3.9.3
```
```
> pipenv run python --version
Python 3.7.10
```
If this doesn't work for you, the problem probably lies somewhere else... maybe you need to re-create the virtual environment?

Moreover, if we merge this PR, it will no longer be easy to work on a different python version, if Python3.7 cannot be easily installed...

dev-tools/install.sh
```diff
@@ -72,6 +72,7 @@ cd $(dirname "$BASH_SOURCE")/..
 if [ "$1" == "--clean" ]; then
   echo "Removing installed dependencies and compiled static files."
   rm -rf .venv node_modules src/cms/static
+  python3 -m venv .venv
```
timoludwig, line 75: Pipenv should automatically create the `.venv` directory in case it doesn't exist... maybe we need another `--pipenv-in-venv` parameter or similar for people who want to install pipenv inside the venv and not via their package manager or `pip --user`...

dev-tools/install.sh
```diff
@@ -72,6 +72,7 @@ cd $(dirname "$BASH_SOURCE")/..
...
 fi
```
timoludwig, line 77: Maybe something along the lines of:
```suggestion
if [ "$1" == "--pipenv-in-venv" ] && [ ! -d ".venv" ]; then
    python3.7 -m venv .venv
    . .venv/bin/activate
    pip3 install pipenv
fi
```

dev-tools/install.sh
svenseeberg: This seems to be a left over from previous experiments I did.

dev-tools/install.sh
```diff
@@ -48,8 +48,8 @@ node_numeric_version="${node_version:1}"
 # Strip trailing minor and patch version
 node_major_version="${node_numeric_version%%.*}"
 # Check node version requirements (12 or higher but not 13)
-if [ "${node_major_version}" -lt 12 ] || [ "${node_major_version}" -eq 13 ]; then
-    echo "nodejs version 12 or version 14 or higher is required, but version $node_version is installed. Please install a recent version manually and run this script again." >&2
+if [ "${node_major_version}" == 12 ] || [ "${node_major_version}" == 14 ] || [ "${node_major_version}" == 15 ] ; then
```
timoludwig, line 51: Now it will show an error whenever the version is correct, so you need to invert the conditions.

PR #788, opened 2021-04-22
Change installation nodejs version check

svenseeberg:
### Short description
Change the install.sh to check for NodeJS versions 12, 14 or 15.

### DIFF
--- a/dev-tools/install.sh
+++ b/dev-tools/install.sh
@@ -48,8 +48,8 @@ node_numeric_version="${node_version:1}"
 # Strip trailing minor and patch version
 node_major_version="${node_numeric_version%%.*}"
 # Check node version requirements (12 or higher but not 13)
-if [ "${node_major_version}" -lt 12 ] || [ "${node_major_version}" -eq 13 ]; then
-    echo "nodejs version 12 or version 14 or higher is required, but version $node_version is installed. Please install a recent version manually and run this script again." >&2
+if ! [[ "${node_major_version}" =~ ^(12|14|15)$ ]] ; then
+    echo "nodejs version 12, 14 or 15 is required, but version $node_version is installed. Please install a supported version manually and run this script again." >&2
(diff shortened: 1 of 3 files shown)


dev-tools/install.sh
```diff
@@ -48,8 +48,8 @@ node_numeric_version="${node_version:1}"
 # Strip trailing minor and patch version
 node_major_version="${node_numeric_version%%.*}"
 # Check node version requirements (12 or higher but not 13)
-if [ "${node_major_version}" -lt 12 ] || [ "${node_major_version}" -eq 13 ]; then
-    echo "nodejs version 12 or version 14 or higher is required, but version $node_version is installed. Please install a recent version manually and run this script again." >&2
+if ! [[ "${node_major_version}" =~ ^(12|14|15)$ ]] ; then
+    echo "nodejs version 12 or version 14 or 15 is required, but version $node_version is installed. Please install a supported version manually and run this script again." >&2
```
timoludwig, line 52:
```suggestion
    echo "nodejs version 12, 14 or 15 is required, but version $node_version is installed. Please install a supported version manually and run this script again." >&2
```

timoludwig (commented):
Looks good!
Could you maybe also update the version info here:

https://github.com/Integreat/integreat-cms/blob/0d623d97bfd2bc91765fbc3e93c4f37a7f15b0dc/sphinx/installation.rst#L17
https://github.com/Integreat/integreat-cms/blob/0d623d97bfd2bc91765fbc3e93c4f37a7f15b0dc/README.md#L19

dev-tools/install.sh
svenseeberg: Indeed.

timoludwig (approved):
> I think we need to check again if version 16 works after the first minor version release.

Yes, sounds reasonable. :+1:

Issue #699, opened 2021-02-23
Move offer selection to region form


timoludwig:
### Proposed Solution

Move the offer page to the region form. Change the selection e.g. to checkboxes. Optionally, we could simplify the models here by removing the offer model and adding a manytomany field to the region model where the offer templates can be selected.

melegiul:
When I understand this right, then this issue would consists of the following steps:

1. Remove the menu "Offers", currently located at the region specific sites
2. This menu mainly allows to activate offers, so this offer activation should be available in the region form, located at the network management site

The "offer template" menu + corresponding form should stay as it is (currently also on network management site)
Is this right?

timoludwig:
> When I understand this right, then this issue would consists of the following steps:
> 
>     1. Remove the menu "Offers", currently located at the region specific sites
> 
>     2. This menu mainly allows to activate offers, so this offer activation should be available in the region form, located at the network management site
> 
> 
> The "offer template" menu + corresponding form should stay as it is (currently also on network management site)
> Is this right?

Yes, correct.

Regarding the idea of simplifying the models in the proposed solution above: I'm not sure if I thought this through, but in theory I think we could remove the Offer model (since all it stores in the database is a mapping from regions to offer templates), but we'd have to move the logic of the properties somewhere... either into the region model or we leave it as class which can be instantiated (but is not saved to the database)... since I didn't try it out yet, I'm not sure which of both works best... do you have any thoughts on this?

timoludwig:
Another possibility would be to move the functionality of transforming `OfferTemplate` objects to `Offer` objects right into the [Offer API](https://github.com/Integreat/integreat-cms/blob/develop/src/api/v3/offers.py). Since the information about Offers isn't needed anywhere else in the CMS, this also sounds like a valid option to me.

PR #790, opened 2021-04-26
Refactor offer model

melegiul:
### Short description

Refactor offer model

### Proposed changes

- Added ManyToManyField to region model referencing OfferTemplates, which replaces the current Offer model as intermediate between OfferTemplate and Region. Added corresponding field to region form.
- Removed Offer model
- Replaced references to Offer with OfferTemplate
- Updated APIs region/offer-feedback database access
- Fixed broken feedback links to offer templates
- Update sitemap app

### Resolved issues

Fixes: #699


src/sitemap/sitemaps.py
```diff
@@ -238,33 +239,39 @@ def __init__(self, region, language):
...
         :return: The absolute path of the given offer object
         :rtype: str
         """
-        return "/" + "/".join([obj.region.slug, self.language.slug, "offers", obj.slug])
+        return reverse(
+            "edit_offer_template",
+            kwargs={
+                "offer_template_id": obj.id,
+            },
+        )
```
timoludwig, line 259:
```suggestion
        return "/" + "/".join([obj.region.slug, self.language.slug, "offers", obj.slug])
```
The sitemap does not generate all urls available in the CMS, but all urls of the WebApp.
Thus, this should be the absolute url to the offer in the WebApp (e.g. https://integreat.app/augsburg/de/offers/sprungbrett), not to the edit form in the CMS.
I noticed that this doesn't work for offers which are immediately opened in an external website, but this is not part of this PR, I will open a new issue for it.

src/sitemap/sitemaps.py
```diff
@@ -238,33 +239,39 @@ def __init__(self, region, language):
...
+        :type obj: ~cms.models.offers.offer_template.OfferTemplate
 
         :return: A list of dictionaries containing the alternative translations of offers
         :rtype: list [ dict ]
         """
         return [
             {
-                "location": f"{WEBAPP_URL}/{self.region.slug}/{language_tree_node.slug}/offers/{obj.slug}",
+                "location": f"{WEBAPP_URL}"
+                + reverse("edit_offer_template", kwargs={"offer_template_id": obj.id}),
```
timoludwig, line 274:
```suggestion
                "location": f"{WEBAPP_URL}/{self.region.slug}/{language_tree_node.slug}/offers/{obj.slug}",
```

timoludwig (commented):
Thanks, this works exactly as expected! :+1: 
I just have a few small ideas for improvement, but in general this looks good!

src/api/v3/feedback/offer_feedback.py
```diff
@@ -37,7 +37,7 @@ def offer_feedback(data, region, language, comment, rating, is_technical):
     :return: JSON object according to APIv3 offer feedback endpoint definition
     :rtype: ~django.http.JsonResponse
     """
-    offer = get_object_or_404(Offer, region=region, template__slug=data.get("slug"))
+    offer = get_object_or_404(OfferTemplate, regions=region.id, slug=data.get("slug"))
```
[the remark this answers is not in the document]
melegiul: Good hint, reading the documentation is an advantage :smile:

src/sitemap/sitemaps.py
melegiul: Yes it was my fault to set the link to the CMS here, but I'm not sure, if that would work, because I called the relationship `regions`. As the the queryset is filtered to only OfferTemplates for a single region, we could use `obj.regions.first().slug` or simpler `self.region.slug`.
On the other side my intention to use reverse here was to not violate the [DRY principle](https://docs.djangoproject.com/en/3.1/topics/http/urls/#reverse-resolution-of-urls). So this hardcoded string would correspond to `reverse("api_offers", kwargs={"region_slug": self.region.slug, "language_slug": self.language.slug}) + obj.slug`.
What do you think about this, does DRY matter here?

timoludwig: Yes, in general you're right, but we can't use the `reverse()` function for this, because links to the webapp are not defined in `urls.py` and will never be.
A different approach could be to use the function [get_absolute_url()](https://docs.djangoproject.com/en/3.1/ref/models/instances/#get-absolute-url) for all model instances and calculate the url to the webapp there. On the long term, we should also do this for pages, events, etc. but it's a good start to do this for offers now...

timoludwig: And you're also right about the region parameter, just `obj.region` won't work. However, your solution doesn't work as well, because even if you have the queryset of all offer templates of region A, it could still be that `obj.regions.first()` is region B if it is activated in that region as well. So maybe we need a similar method to the ones in the API `get_url(offer, region)` which calculates the correct URL. And `get_absolute_url()` can't be used for this, because it doesn't have access to the region object. So I suggest to either leave the urls hardcoded in the sitemaps.py, or add a property `get_url(self, region)` to the offer template model...
Or do you have any other ideas?

src/api/v3/regions.py
```diff
@@ -107,7 +111,9 @@ def hiddenregions(_):
         map(
             transform_region_by_status,
             Region.objects.filter(status=region_status.HIDDEN).annotate(
-                offers_enabled=Exists(Offer.objects.filter(region=OuterRef("pk")))
+                offers_enabled=Exists(
+                    OfferTemplate.objects.filter(regions=OuterRef("pk"))
+                )
```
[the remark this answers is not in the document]
melegiul: Much simpler, is done in 1ab25eb82ffcc46a5ae3e1809e0a9047400b4336

src/sitemap/sitemaps.py
melegiul: This ManyToManyField pitfalls :smile: 
So I like your idea about the offer template property. I will add a commit soon.

melegiul: Alright, this is done in d2b7a3e8ac442880d4c24f073106e1f028f42ad2, finally I just left it hardcoded, but just updated the region slug reference

timoludwig (approved):
Cool, thanks for your changes! :muscle:

melegiul:
One thing I forgot to mention:
Former there was a template tag for retrieving the date, when an offer was activated:
https://github.com/Integreat/integreat-cms/blob/5425956ffa8a1c0c9c36953e86c57f7593c86352/src/cms/templatetags/offer_filters.py#L10
Shall I delete the template tag or do we need that feature now in the region form (maybe with a custom widget for the MultipleCheckboxInput)?

timoludwig:
> One thing I forgot to mention:
> Former there was a template tag for retrieving the date, when an offer was activated:
> 
> https://github.com/Integreat/integreat-cms/blob/5425956ffa8a1c0c9c36953e86c57f7593c86352/src/cms/templatetags/offer_filters.py#L10
> 
> Shall I delete the template tag or do we need that feature now in the region form (maybe with a custom widget for the MultipleCheckboxInput)?

Ah okay, I don't think we need this anymore, you can delete the template tag.

PR #797, opened 2021-04-27
Change GVZ API wrapper

### LINKED ISSUE #796: GVZ API client not working anymore
### Describe the Bug

Since the GVZ API server changed recently, our GVZ API client doesn't work anymore:
```
Apr 26 18:05:17 INFO gvz_api.apps - GVZ API is not available. You won't be able to automatically import coordinates and region aliases.
```

### Steps to Reproduce

1. Start development server
4. See GVZ API logging message

### Expected Behavior

The GVZ API is available, so it should be recognized as available
### Actual Behavior

The GVZ API server is mistakenly recognized as down

### Additional Information

If interacting with the GVZ API is a common task for other use cases, it could also make sense to outsource our gvz api client to a separate repository and distribute it as pip package.

svenseeberg:
This PR adopts the GVZ API v2.

To accommodate the API changes, the handling of requests is also changed. First, the children need to be searched recursively. Second, all details are directly provided in the answers. Therefore the coordinates dictionary for children can directly be created.

Fixes #796

### DIFF
--- a/src/cms/forms/regions/region_form.py
+++ b/src/cms/forms/regions/region_form.py
@@ -132,10 +132,10 @@ class RegionForm(CustomModelForm):
         if apps.get_app_config("gvz_api").api_available:
             gvz_region = GvzRegion(
                 region_name=cleaned_data["name"],
-                region_key=cleaned_data["common_id"],
+                region_ags=cleaned_data["common_id"],
                 region_type=cleaned_data["administrative_division"],
             )
-            if gvz_region.aliases and cleaned_data["aliases"] == "":
+            if gvz_region.aliases and cleaned_data["aliases"] in [{}, ""]:
                 cleaned_data["aliases"] = gvz_region.aliases
             if gvz_region.longitude and cleaned_data["longitude"] == 0.0:
                 cleaned_data["longitude"] = gvz_region.longitude
--- a/src/gvz_api/apps.py
+++ b/src/gvz_api/apps.py
@@ -28,11 +28,9 @@ class GvzApiConfig(AppConfig):
         if "runserver" in sys.argv or "APACHE_PID_FILE" in os.environ:
             if settings.GVZ_API_ENABLED:
                 try:
-                    response = requests.get(
-                        f"{settings.GVZ_API_URL}/search/expect_empty_json", timeout=3
-                    )
-                    # Require the response to be empty, otherwise it's probably an error
-                    assert not json.loads(response.text)
+                    response = requests.get(f"{settings.GVZ_API_URL}/api/", timeout=3)
+                    # Require the response to return a valid JSON, otherwise it's probably an error
+                    assert json.loads(response.text)
                 except (
                     json.decoder.JSONDecodeError,
                     requests.exceptions.RequestException,
--- a/src/gvz_api/utils.py
+++ b/src/gvz_api/utils.py
@@ -5,7 +5,7 @@ import logging
 import json
 import requests
 from django.conf import settings
-from cms.constants import administrative_division
+from cms.constants import administrative_division as ad
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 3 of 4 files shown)


src/cms/forms/regions/region_form.py
```diff
@@ -132,7 +132,7 @@ def clean(self):
         if apps.get_app_config("gvz_api").api_available:
             gvz_region = GvzRegion(
                 region_name=cleaned_data["name"],
-                region_key=cleaned_data["common_id"],
+                region_ags=cleaned_data["common_id"],
                 region_type=cleaned_data["administrative_division"],
             )
             if gvz_region.aliases and cleaned_data["aliases"] == "":
```
timoludwig, line 138:
```suggestion
            if gvz_region.aliases and cleaned_data["aliases"] == in ["", "{}"]:
```
An emty JSON encoded as string is `{}`, so I suggest to update the aliases in this case as well.

timoludwig (commented):
Thanks, the GVZ API client works again! :+1:

src/gvz_api/utils.py
```diff
@@ -141,21 +159,6 @@ def best_match(self, region_name, region_type=None):
                 results_literal.append(region)
         if len(results_literal) == 1:
```
melegiul, line 160: Very cool GVZ API :rocket:
Coincidentally I had just a look at the code, too.
I found that saving a region called "Augsburg" with administrative division "Rural district" ("Landkreis")
displays the logger message: "GVZ API did not find type match for 'Augsburg'",
because the `results_literal` list contained two entries: "Kreisfreie Stadt (Kreisebene)" and "Landkreis".
Maybe we would need here some additional filters?

timoludwig: If I understood the [constants](https://github.com/Integreat/gemeindeverzeichnis-django/blob/main/src/gvz/constants.py) defined in the GVZ API correctly, the mapping should be the following:
```
Name             | Django         | GVZ

Kreis            | DISTRICT       | CT_ABSTRACT (40) or CT_COUNTY (43)
Landkreis        | RURAL_DISTRICT | CT_RURAL_COUNTY (44)
Kreisfreie Stadt | URBAN_DISTRICT | CT_URBAN_MUNICIPALITY (41) or MP_URBAN_MUNICIPALITY (61)
```
But in the [utils](https://github.com/Integreat/integreat-cms/blob/7b86675a13e75dfac0798ddcd7411e17b9fb956b/src/gvz_api/utils.py#L106-L127), this is defined a bit different :thinking: 

I think the optimal way would be if we'd find a definition we can all agree on and use the same one for Django and for the GVZ API? Then we wouldn't need a hardcoded mapping between the two systems... which source did you use to create the constants for the GVZ API?

src/cms/fixtures/test_data_reduced.json
```diff
@@ -0,0 +1,30 @@
...
+    "model": "cms.userprofile",
+    "pk": 1,
+    "fields": {
+      "user": 1,
+      "organization": null,
+      "regions": [],
+      "expert_mode": true
+    }
+  }
+]
```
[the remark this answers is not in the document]
svenseeberg: This file should not be in this commit. This is something from the migrations process and accidentally landed in here. I'll remove the file.

src/gvz_api/utils.py
svenseeberg: The names come from https://www.destatis.de/DE/Themen/Laender-Regionen/Regionales/Gemeindeverzeichnis/Administrativ/beschreibung-gebietseinheiten.pdf?__blob=publicationFile. However, the translations into English constants are a bit tricky/not perfect.

With carefully selecting all required objects from the GVZ, we could get around the mapping. However, the official values are very redundant right now, and I'm not sure if we really like to do this. Right now I'd recommend to keep the mapping.

src/gvz_api/apps.py
```diff
@@ -28,11 +28,9 @@ def ready(self):
         if "runserver" in sys.argv or "APACHE_PID_FILE" in os.environ:
             if settings.GVZ_API_ENABLED:
                 try:
-                    response = requests.get(
-                        f"{settings.GVZ_API_URL}/search/expect_empty_json", timeout=3
-                    )
+                    response = requests.get(f"{settings.GVZ_API_URL}/api/", timeout=3)
                     # Require the response to be empty, otherwise it's probably an error
-                    assert not json.loads(response.text)
+                    assert json.loads(response.text)
```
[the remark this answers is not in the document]
svenseeberg: It seems I already fixed this :D

src/gvz_api/utils.py
timoludwig: Ok, I see. Also the issue with e.g. "Landkreis Augsburg" persists (that no unambiguous match is found even if there are not multiple Landkreise with the name "Augsburg"), but if there is no easy fix to it, we should probably open a new issue for more refined constants and a better mapping...

timoludwig (approved):
Apart from the constants mapping, everything looks fine :+1:
But we can fix this later on.

Issue #698, opened 2021-02-23
Improve push notification form


timoludwig:
### Proposed Solution

- [x] Make input field to text area
- [x] Show how many characters are remaining
- [x] Rename "Push-Notifications" to "News" / "Nachrichten"
- [x] Change not sent symbol "x" to text "Message not sent yet"

### Additional Context
![Bildschirmfoto vom 2021-04-13 16-42-11](https://user-images.githubusercontent.com/15608678/114572956-3be7fd00-9c78-11eb-928e-b156602af7ac.png)

Files that need some editing for that issue are:

- src/cms/templates/push_notifications/push_notification_form.html (for everything related to the HTML-Input Field)
- src/cms/locale/de/LC_MESSAGES/django.po (for everything that needs some new translation. Search for "Push Notification" and change the german translation to "Nachricht/Nachrichten"

PR #802, opened 2021-05-03
Improve push notification form

MizukiTemma:
### Short description

Changes for the Issue #698 Improve push notification form are done.

### Proposed changes

- Renaming : Pusch nofification -> News message
- Message field is now multi line
- Letter count is added below the message field

### Resolved issues

Fixes: #698

_Note: This supersedes #791_

### DIFF
--- a/src/cms/locale/de/LC_MESSAGES/django.po
+++ b/src/cms/locale/de/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2021-05-03 11:27+0000\n"
+"POT-Creation-Date: 2021-05-05 13:32+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: Integreat <info@integreat-app.de>\n"
 "Language-Team: Integreat <info@integreat-app.de>\n"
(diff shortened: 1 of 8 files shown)


timoludwig (commented):
Thanks for your PR! :+1: 

In general this works for our test data, but we have to deal with a dynamic number of languages, so I added a suggestion on how to handle it in a generic way.

timoludwig (commented):
Thanks for your changes! :+1:
The length-checking of the messages now works as expected.

One task of issue #698 is still missing though:
  - [ ] Change not sent symbol "x" to text "Message not sent yet"

Since the "sent"-column in the does not look very appealing yet:

Screenshot_2021-05-05 Integreat Redaktionssystem(2)

I suggest to change it to something like this:

Screenshot_2021-05-05 Integreat Redaktionssystem(1)

Which means changing the lines 
https://github.com/Integreat/integreat-cms/blob/226a292ded181dfafe536ff0b4e5f24e5bbdc071/src/cms/templates/push_notifications/push_notification_list_row.html#L26-L32
to
```
	<td class="px-2 whitespace-nowrap">
		{% if push_notification.sent_date %}
			<i data-feather="check" class="text-green-500 align-text-top"></i> {{ push_notification.sent_date }}
		{% else %}
			<i data-feather="x" class="text-red-500 align-text-top"></i> {% trans 'Message not sent yet' %}
		{% endif %}
	</td>
```
And adding the line `{% load i18n %}` in top of the file to activate translatable strings.

timoludwig (approved):
Thanks! You can now merge this PR! :rocket:

Issue #592, opened 2020-11-18
Dashboard selection for multiple region users


timoludwig:
### Motivation

Users, who are not admins but have access to multiple regions, need some kind of "landing page" where they can select which region dashboard they want to open.

### Proposed Solution

Add a custom view for this use case and redirect all users who aren't admins but have access to multiple regions to this view.

PR #803, opened 2021-05-03
Implement regions selection for multiple region users

david-venhoff:
### Short description
This pr adds a region selection screen for users who have access to multiple regions, which is shown after login.

### Resolved issues

Fixes: #592

### DIFF
--- a/src/cms/templates/dashboard/region_selection.html
+++ b/src/cms/templates/dashboard/region_selection.html
@@ -1,0 +1,24 @@
+{% extends "authentication/_base.html" %}
+
+{% load i18n %}
+{% load static %}
+
+{% block heading %}
+    {% trans "Select region" %}
+{% endblock %}
+
+{% block content %}
+    <table class="w-full mt-4 rounded border border-solid border-gray-200 shadow bg-white">
+        <tbody>
+        {% for region in request.user.profile.regions.all %}
+        <tr class="border-t border-solid border-gray-200 hover:bg-gray-100">
+            <td class="pl-2">
+                <a href="{% url 'dashboard' region_slug=region.slug %}" class="block py-3 px-2 text-gray-800">
+                    {{ region.full_name }}
+                </a>
+            </td>
+        </tr>
+        {% endfor%}
+        </tbody>
+    </table>
+{% endblock %}
--- a/src/cms/views/dashboard/redirect_view.py
+++ b/src/cms/views/dashboard/redirect_view.py
@@ -32,9 +32,14 @@ class RedirectView(TemplateView):
         """
 
         user = request.user
+        regions = user.profile.regions
         if user.is_superuser or user.is_staff:
             return redirect("admin_dashboard")
-        regions = user.profile.regions
-        if regions.exists():
+
+        if regions.count() == 1:
             return redirect("dashboard", region_slug=regions.first().slug)
+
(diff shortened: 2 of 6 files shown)


src/cms/views/dashboard/redirect_view.py
```diff
@@ -32,9 +32,14 @@ def get(self, request, *args, **kwargs):
         """
 
         user = request.user
+        regions = user.profile.regions
         if user.is_superuser or user.is_staff:
             return redirect("admin_dashboard")
-        regions = user.profile.regions
-        if regions.exists():
+
+        if regions.count == 1:
```
timoludwig, line 39:
```suggestion
        if regions.count() == 1:
```
`count` itself is only the method, and since this method does not equal `1`, this check will always be `False`.

src/cms/templates/dashboard/region_selection.html
```diff
@@ -0,0 +1,24 @@
...
+{% load static %}
+
+{% block heading %}
+    {% trans "Select region" %}
+{% endblock %}
+
+{% block content %}
+    <table class="w-full mt-4 rounded border border-solid border-gray-200 shadow bg-white">
+        <tbody>
+        {% for region in user_regions %}
```
timoludwig, line 13:
```suggestion
        {% for region in request.user.profile.regions.all %}
```
No need for an extra variable, because the user object is already passed to the template via the [Request Context Processor](https://docs.djangoproject.com/en/3.2/ref/templates/api/#django-template-context-processors-request).

src/cms/templates/dashboard/region_selection.html
```diff
@@ -0,0 +1,24 @@
...
+            <td class="pl-2">
+                <a href="{% url 'dashboard' region_slug=region.slug %}" class="block py-3 px-2 text-gray-800">
+                    {{ region.full_name }}
+                </a>
+            </td>
+        </tr>
+        {% endfor%}
+        </tbody>
+    </table>
+{% endblock %}
```
timoludwig, line 24:
```suggestion
{% endblock %}

```
Please try not to add spaces at the end of a line and to end files with a newline.

src/cms/views/dashboard/region_selection.py
```diff
@@ -0,0 +1,41 @@
...
+        user = request.user
+        regions = user.profile.regions
+        return render(
+            request,
+            self.template_name,
+            {
+                "user_regions": regions.all(),
+                **self.get_context_data(**kwargs),
+            },
+        )
```
timoludwig, line 41:
```suggestion
```
Since this template view doesn't need extra context (see my comment above), you can just omit this method altogether and rely on the inbuilt Django [TemplateView](https://docs.djangoproject.com/en/3.2/ref/class-based-views/base/#django.views.generic.base.TemplateView) for rendering the request.

timoludwig (commented):
Amazing, thanks! :+1:
Nearly works as expected, just found a small bug and suggested a few improvements.

src/cms/views/dashboard/redirect_view.py
Inky-developer: Oops!

timoludwig (approved):
Thanks for your changes!
I squashed your commits into one again... after the CI pipeline is finished, you can merge this PR :tada:

Issue #805, opened 2021-05-04
Install script python3.7 checks failing


melegiul:
### Describe the Bug

Install script bug during python3.7 check

### Steps to Reproduce

1. Run the install script
2. See the warning about the python3.7-dev

### Expected Behavior

the script should pass the check, if python3.7-dev is installed

### Actual Behavior

Instead the script warns, that python3.7 is not installed

PR #806, opened 2021-05-04
Fixes the install script python3.7-dev check

melegiul:
### Short description

Fixes the install script python3.7-dev checker

### Proposed changes

Referencing @timoludwig 
"`grep -q` exits immediately after it found the first match, but dpkg -l is not done writing its output, so it sends the error 141 SIGPIPE because it can no longer pipe its output to grep because the process is dead"

### Resolved issues

Fixes: #805


timoludwig (approved):
Thanks! :+1:

PR #807, opened 2021-05-06
Improve dev tools and installation workflow

### LINKED ISSUE #728: Document / install development requirements
### Motivation
It seems that our CMS currently runs without problems on Ubuntu 20.04 and Debian 10. However, starting from a fresh (netinstall) setup, some requirements need to be fulfilled for install.sh to work.

### Proposed Solution
Adding a line `sudo apt install rustc build-essential openssl-dev pipenv python3-pip python3-setuptools` should do the trick.

timoludwig:
### Short description

I found another edge case where my developer tools didn't work properly, and this PR should fix this.
Additionally, I used the opportunity to implement some other improvements.

### Proposed changes

- Fix execution of pipenv with sudo
- Check if CMS is installed before running
- Make output of WebPack less verbose
- Remove dependency `python3.7-dev` on Ubuntu
- Add detailed commands for the installation of the prerequisites on popular linux distributions (fixes #728):
    - Ubuntu 20.04 LTS
    - Debian 10
    - Arch Linux
### Resolved issues

Fixes: #728

### DIFF
--- a/dev-tools/_functions.sh
+++ b/dev-tools/_functions.sh
@@ -11,6 +11,8 @@ if [[ "$*" == *"--verbose"* ]]; then
     set -vx
 fi
 
+# The Port on which the Integreat CMS development server should be started (do not use 9000 since this is used for webpack)
+INTEGREAT_CMS_PORT=8000
 # The name of the used database docker container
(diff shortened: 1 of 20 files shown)


svenseeberg (approved):
Successfully guided my through installation. IMHO good to merge.

Issue #671, opened 2021-02-19
Collapse pages per default above threshold


timoludwig:
### Motivation

Since pagination doesn't work well on trees, we should implement a different measure to prevent too many entries which can be overwhelming at first sight.

### Proposed Solution

Collapse all subpages until the total amount of visible pages in the tree is below a certain threshold.
Probably use the same setting which is planned for pagination: https://github.com/Integreat/integreat-cms/blob/15099b61c28afa5f290e8ace5da904d139ddc075/src/backend/settings.py#L533

Different collapsing algorithms are possible:
1. Radically collapse all subpages
    - Pro: less effort
    - Con: More effort to expand the subpages
2. Collapse only subpages of root pages
    - Pro: Also not much effort
    - Con: Many pages expanded at once, which can be confusing
3. Collapse and expand subpages based on number of pages on the sub-level (e.g. if the number of descendants is above threshold, only expand the direct children, otherwise expand all descendants)
    - Pro: Good balance between clicks needed to expand tree and number of entries expanded at once... so best user experience???
    - Con: More effort

### Additional Context

Additionally, add the possibility to collapse and expand all pages at once.

ulliholtgrave:
As our current system uses the first option, we should collapse all sub pages by default.

PR #808, opened 2021-05-06
Collapse pages per default

timoludwig:
### Short description

This collapses all pages per default, so only root pages are visible when opening the page tree.

### Proposed changes

- Collapse pages per default

### Resolved issues

Fixes: #671


ulliholtgrave (approved):
Nice, quick and easy. Thanks for the PR! :)

Issue #794, opened 2021-04-26
Add API endpoint for POIs


timoludwig:
### Motivation

At the moment, there is no API endpoint for POIs / locations

### Proposed Solution

Create the endpoint as specified in the wiki of the legacy CMS:
https://github.com/Integreat/cms/wiki/REST-APIv3-Documentation#pageseventslocationsdisclaimer

PR #821, opened 2021-05-11
Add api endpoint for locations

david-venhoff:
### Short description
This pr adds support for the locations api endpoint.

### Resolved issues
Fixes: #794

### DIFF
--- a/src/api/v3/locations.py
+++ b/src/api/v3/locations.py
@@ -1,0 +1,73 @@
+from django.http import JsonResponse
+
+from backend.settings import WEBAPP_URL
+from cms.models import Region
+from ..decorators import json_response
+
+
+def transform_poi_translation(poi_translation):
+    """
+    Function to create a JSON from a single poi_translation object.
+
+    :param poi_translation: The poi translation object which should be converted
+    :type poi_translation: ~cms.models.pois.poi_translation.POITranslation
+
+    :return: return data necessary for API
+    :rtype: dict
+    """
+
+    poi = poi_translation.poi
+    return {
+        "id": poi_translation.id,
+        "url": WEBAPP_URL + poi_translation.get_absolute_url(),
+        "path": poi_translation.get_absolute_url(),
+        "title": poi_translation.title,
+        "modified_gmt": poi_translation.last_updated.strftime("%Y-%m-%d %H:%M:%S"),
+        "excerpt": poi_translation.short_description,
+        "content": poi_translation.description,
+        "available_languages": poi_translation.available_languages,
+        "thumbnail": poi.icon.url if poi.icon else None,
+        "location": {
+            "id": poi.id,
+            "name": poi_translation.title,
+            "address": poi.address,
+            "town": poi.city,
+            "state": None,
+            "postcode": poi.postcode,
+            "region": None,
+            "country": poi.country,
+            "latitude": poi.latitude,
+            "longitude": poi.longitude,
+        },
+        "hash": None,
+    }
+
+
+@json_response
+# pylint: disable=unused-argument
+def locations(request, region_slug, language_slug):
+    """
+    List all POIs of the region and transform result into JSON
+
+    :param request: The current request
+    :type request: ~django.http.HttpRequest
+
+    :param region_slug: The slug of the requested region
+    :type region_slug: str
+
+    :param language_slug: The slug of the requested language
+    :type language_slug: str
+
+    :return: JSON object according to APIv3 locations endpoint definition
+    :rtype: ~django.http.JsonResponse
+    """
+    region = Region.get_current_region(request)
+    result = []
+    for poi in region.pois.filter(archived=False):
+        translation = poi.get_public_translation(language_slug)
+        if translation:
+            result.append(transform_poi_translation(translation))
+
+    return JsonResponse(
+        result, safe=False
+    )  # Turn off Safe-Mode to allow serializing arrays
--- a/src/api/urls.py
+++ b/src/api/urls.py
@@ -18,6 +18,7 @@ from .v3.feedback import (
 )
 from .v3.imprint import imprint
 from .v3.languages import languages
+from .v3.locations import locations
 from .v3.pages import pages
 from .v3.pdf_export import pdf_export
 from .v3.push_notifications import sent_push_notifications
@@ -41,6 +42,7 @@ urlpatterns = [
                     include(
                         [
                             url(r"^pages/?$", pages, name="api_pages"),
+                            url(r"^locations/?$", locations, name="api_locations"),
                             url(
                                 r"^page/?$",
                                 single_page,


src/api/v3/locations.py
```diff
@@ -0,0 +1,68 @@
...
+
+    :return: return data necessary for API
+    :rtype: dict
+    """
+
+    # is there a better way to access this?
+    try:
+        icon_url = poi_translation.icon.url
+    except AttributeError:
+        icon_url = ""
```
david-venhoff, line 22: This is probably not what I am supposed to do, how should I access the url correctly?

src/api/v3/locations.py
```diff
@@ -0,0 +1,68 @@
...
+
+    poi = poi_translation.poi
+    return {
+        "id": poi_translation.id,
+        "url": poi_translation.permalink,
+        "path": poi_translation.slug,
+        "title": poi_translation.title,
+        "modified_gmt": poi_translation.last_updated,
```
david-venhoff, line 30: This does not match the exact format required by https://github.com/Integreat/cms/wiki/REST-APIv3-Documentation#pageseventslocationsdisclaimer. However, other parts of the api send the `modified_gmt` in the exact same way. Should I change this to match the exact format?

src/api/v3/locations.py
```diff
@@ -0,0 +1,68 @@
...
+        "excerpt": poi_translation.short_description,
+        "content": poi_translation.description,
+        "available_languages": poi_translation.available_languages,
+        "thumbnail": icon_url,
+        "location": {
+            "id": poi.id,
+            "name": poi_translation.title,
+            "address": poi.address,
+            "town": poi.city,
+            "state": None,
```
david-venhoff, line 40: I could not find a matching field, is it okay to send a None here?

timoludwig: I just asked the app team, and they're not using the field anyway, so I think we can just set `None` here.

src/api/v3/locations.py
```diff
@@ -0,0 +1,68 @@
...
+            "postcode": poi.postcode,
+            "region": poi.region.slug,
```
timoludwig, line 42:
```suggestion
            "region": None,
```
I think the field `region` of the WordPress events plugin means a different thing than what we refer to as "region"... so I would just set this field to `None` as well... (I couldn't find a location in wordpress which had set any value in this field)

src/api/v3/locations.py
```diff
@@ -0,0 +1,68 @@
...
+            "country": poi.country,
+            "latitude": poi.latitude,
+            "longitude": poi.longitude,
+        },
+        "hash": None,
```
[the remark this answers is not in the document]
timoludwig: I think this is fine, the hash-functionality of the APIv3 isn't used anyway, so we don't have to implement this yet.

src/api/v3/locations.py
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+
+@json_response
+# pylint: disable=unused-argument
+def locations(_, region_slug, language_slug):
```
timoludwig, line 53:
```suggestion
# pylint: disable=unused-argument
def locations(request, region_slug, language_slug):
```
I think we need the request to get the current region

src/api/v3/locations.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    List all poi and transform result into JSON
```
timoludwig, line 55:
```suggestion
    List all POIs of this region and transform result into JSON

    :param request: The current HTTP request
    :type request: ~django.http.HttpRequest

    :param region_slug: The slug of the requested region
    :type region_slug: str

    :param language_slug: The slug of the requested language
    :type language_slug: str
```

src/api/v3/locations.py
```diff
@@ -0,0 +1,68 @@
...
+
+    :return: JSON object according to APIv3 locations endpoint definition
+    :rtype: ~django.http.JsonResponse
+    """
+    result = []
```
timoludwig, line 60:
```suggestion
    region = Region.get_current_region(request)
    result = []
```

src/api/v3/locations.py
```diff
@@ -0,0 +1,68 @@
...
+    for poi in POI.objects.all():
```
timoludwig, line 61:
```suggestion
    for poi in region.pois.filter(archived=False):
```
The endpoint should just return the POIs of the requested region, not all POIs.

src/api/v3/locations.py
```diff
@@ -0,0 +1,68 @@
+from django.http import JsonResponse
+
+from cms.models import POI
```
timoludwig, line 3:
```suggestion
from backend.settings import WEBAPP_URL
from cms.models import Region
```

src/api/v3/locations.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    List all poi and transform result into JSON
+
+    :return: JSON object according to APIv3 locations endpoint definition
+    :rtype: ~django.http.JsonResponse
+    """
+    result = []
+    for poi in POI.objects.all():
+        translation = poi.get_public_translation(language_slug)
+        if translation is not None:
```
timoludwig, line 63:
```suggestion
        if translation:
```
In Python, objects can never evaluate to False, no matter their type or attributes (exception: if it implements a special method `__bool __()` - and in this case the boolean evaluation is explicitly wanted, and not an unexpected side-effect). So it's safe to evaluate the object directly instead of checking whether it is not `None`. This is only required when you want to differ between `[]` and `None` or `""` and `None` or other falsey values etc...

timoludwig (commented):
Thanks!
In general it looks good, I just commented some clarifications about the specific fields.

src/api/v3/locations.py
```diff
@@ -0,0 +1,68 @@
...
+    return {
+        "id": poi_translation.id,
+        "url": poi_translation.permalink,
+        "path": poi_translation.slug,
+        "title": poi_translation.title,
+        "modified_gmt": poi_translation.last_updated,
+        "excerpt": poi_translation.short_description,
+        "content": poi_translation.description,
+        "available_languages": poi_translation.available_languages,
+        "thumbnail": icon_url,
```
timoludwig, line 34:
```suggestion
        "thumbnail": poi.icon.url if poi.icon else None,
```

src/api/v3/locations.py
timoludwig, line 22:
```suggestion
```

src/api/v3/locations.py
```diff
@@ -0,0 +1,64 @@
...
+        "hash": None,
+    }
+
+
+@json_response
+# pylint: disable=unused-argument
+def locations(request, region_slug, language_slug):
+    """
+    List all POIs of the region and transform result into JSON
+
```
timoludwig, line 51:
```suggestion

    :param request: The current request
    :type request: ~django.http.HttpRequest

    :param region_slug: The slug of the requested region
    :type region_slug: str

    :param language_slug: The slug of the requested language
    :type language_slug: str

```

timoludwig (commented):
Thanks!
Could you please add the missing parameter docstring? After that, the PR is ready to merge! :tada:

timoludwig (approved):
Thank you! :+1:

PR #823, opened 2021-05-12
Remove `node_modules` from static files

### LINKED ISSUE #484: Do not serve node_modules under /static

timoludwig:
### Short description

Since our change to WebPack, we luckily don't need to serve the `node_modules` under the `static` path anymore.
Interestingly, we had an issue for this (#484 "Do not serve `node_modules` under `/static`"), but closed it even though the `node_modules` were still served.

### Proposed changes

- Remove `node_modules` from static files

### Resolved issues

Fixes #484

### DIFF
--- a/src/backend/settings.py
+++ b/src/backend/settings.py
@@ -612,12 +612,6 @@ else:
 #: (see :setting:`django:STATIC_URL` and :doc:`Managing static files <django:howto/static-files/index>`)
 STATIC_URL = "/static/"
 
-#: This setting defines the additional locations the staticfiles app will traverse
-#: (see :setting:`django:STATICFILES_DIRS` and :doc:`Managing static files <django:howto/static-files/index>`)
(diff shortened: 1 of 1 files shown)


ulliholtgrave (approved):
:+1:

PR #828, opened 2021-05-13
Improve translation conflict resolution

timoludwig:
### Short description

Some minor improvements to the git translation conflict resolution tool.

### Proposed changes

- Enable resolving of git stash conflicts
- Improve error output in case `msguniq` fails
- Improve error output in case conflicts cannot be solved automatically


dev-tools/resolve_translation_conflicts.sh
```diff
@@ -9,20 +9,32 @@ source "$(dirname "${BASH_SOURCE[0]}")/_functions.sh"
...
 # Check if conflicts remain
-if grep -q "#-#-#-#-#" django.po; then
-    echo "Not all conflicts could be solved automatically. Please resolve remaining conflicts manually (marked with \"#-#-#-#-#\")."
-else
-    # Fix line numbers and empty lines
-    bash "${DEV_TOOL_DIR}/translate.sh"
+if grep --quiet "#-#-#-#-#" "$TRANSLATION_FILE"; then
+    echo -e "\n$TRANSLATION_FILE"
+    grep --before-context=2 --after-context=1 --line-number "#-#-#-#-#" "$TRANSLATION_FILE" | format_grep_output | print_with_borders
+    echo -e "❌ Not all conflicts could be solved automatically" | print_error
```
ulliholtgrave, line 32: Just a small thing: Do you want to add the "\n" as in line 23?

timoludwig: Since the function [print_with_borders](https://github.com/Integreat/integreat-cms/blob/4882518cef39864502051860d6b17b1d7707d284/dev-tools/_functions.sh#L88) already outputs one additional newline at its end, this is not required here. In both cases, there is an empty new line between the previous line and the error message. You can try this out be recreating both error messages manually (a syntax error is pretty easy by just breaking something, and a `#-#-#-#-#` conflict can be created by inserting two different translation for the same source string).

ulliholtgrave: Ahh I see :)

ulliholtgrave (approved):
Seems to work on my local branch :+1:

Issue #831, opened 2021-05-13
`UnboundLocalError` in imprint form


timoludwig:
### Describe the Bug

The imprint can only be created when it does not yet exist.
Once it exists, it cannot be changed anymore.

### Steps to Reproduce

1. Go to the imprint form
2. Create imprint
3. Change existing imprint translation

### Expected Behavior

The imprint translation should be changed

### Actual Behavior

`UnboundLocalError at /augsburg/imprint/de/`
`local variable 'imprint' referenced before assignment`
### Additional Information

<details>
 <summary>Traceback</summary>

```
Environment:


Request Method: POST
Request URL: http://localhost:8000/augsburg/imprint/de/

Django Version: 3.2.2
Python Version: 3.7.10
Installed Applications:
['cms.apps.CmsConfig',
 'gvz_api.apps.GvzApiConfig',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.messages',
 'django.contrib.sessions',
 'django.contrib.sitemaps',
 'django.contrib.staticfiles',
 'corsheaders',
 'widget_tweaks',
 'easy_thumbnails',
 'filer',
 'mptt',
 'rules.apps.AutodiscoverRulesConfig',
 'linkcheck',
 'django.contrib.admin']
Installed Middleware:
['corsheaders.middleware.CorsMiddleware',
 'django.middleware.security.SecurityMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.locale.LocaleMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'cms.middleware.timezone_middleware.TimezoneMiddleware']

Traceback (most recent call last):
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.7/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.7/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
[...]

PR #832, opened 2021-05-13
Fix `UnboundLocalError` in imprint form

timoludwig:
### Short description

Fix `UnboundLocalError` in imprint form

### Proposed changes

- Set `imprint` to `imprint_instance` if imprint already exists

### Resolved issues

Fixes: #831


ulliholtgrave (approved):
Fixes the problem :)

Issue #833, opened 2021-05-14
Query string for pagination not working


melegiul:
### Describe the Bug

The query string for pagination defined in the links of the pagination template has a divergent name to most views that use pagination. The paginators get method also works with empty query strings, but only returns the first page in this case.

### Steps to Reproduce

1. Go to a view that uses pagination like the user lists
2. Add as many user to exceed the number per page
3. Try to proceed to another page than the first
4. All links go back to the first page

### Expected Behavior

Browsing between the different pages should be possible

### Actual Behavior

All pagination links route back to the first page

### Additional Information

The query string was renamed in the pagination template in accordance to the frameworks default name

PR #834, opened 2021-05-14
Rename query string for pagination in views

melegiul:
### Short description

Fixes the links for all pagination links after the first page

### Proposed changes

Rename the paginators query string in the list views

### Resolved issues

Fixes: #833


timoludwig (approved):
Thanks, looks good to me! :+1:

Issue #793, opened 2021-04-26
Add API endpoint for events


timoludwig:
### Motivation

At the moment, there is no API endpoint for events

### Proposed Solution

Create the endpoint as specified in the wiki of the legacy CMS:
https://github.com/Integreat/cms/wiki/REST-APIv3-Documentation#pageseventslocationsdisclaimer

PR #835, opened 2021-05-14
Add api endpoint for events

david-venhoff:
### Short description
This pr adds support for the events api endpoint.

### Resolved issues

Fixes: #793

### DIFF
--- a/src/api/v3/events.py
+++ b/src/api/v3/events.py
@@ -1,0 +1,96 @@
+from django.http import JsonResponse
+from django.utils import timezone
+
+from backend.settings import WEBAPP_URL, CURRENT_TIME_ZONE
+from cms.models import Region
+from .locations import transform_poi
+from ..decorators import json_response
+
+
+def transform_event(event):
+    """
+    Function to create a JSON from a single event object.
+
+    :param event: The event which should be converted
+    :type event: ~cms.models.events.event.Event
+
+    :return: return data necessary for API
+    :rtype: dict
+    """
+    return {
+        "id": event.id,
+        "start_date": event.start_date,
+        "end_date": event.end_date,
+        "all_day": event.is_all_day,
+        "start_time": event.start_time,
+        "end_time": event.end_time,
+        "recurrence_id": event.recurrence_rule.id if event.recurrence_rule else None,
+        "timezone": CURRENT_TIME_ZONE,
+    }
+
+
+def transform_event_translation(event_translation):
+    """
+    Function to create a JSON from a single event_translation object.
+
+    :param event_translation: The event translation object which should be converted
+    :type event_translation: ~cms.models.events.event_translation.EventTranslation
+
+    :return: return data necessary for API
+    :rtype: dict
+    """
+
+    event = event_translation.event
+    if event.location:
+        location_translation = (
+            event.location.get_public_translation(event_translation.language.slug)
+            or event.location.default_translation
+        )
+        location = transform_poi(event.location, location_translation)
+    else:
+        location = None
+
+    return {
+        "id": event_translation.id,
+        "url": WEBAPP_URL + event_translation.get_absolute_url(),
+        "path": event_translation.get_absolute_url(),
+        "title": event_translation.title,
+        "modified_gmt": event_translation.last_updated.strftime("%Y-%m-%d %H:%M:%S"),
+        "excerpt": event_translation.description,
+        "content": event_translation.description,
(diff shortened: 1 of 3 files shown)


src/api/v3/events.py
```diff
@@ -0,0 +1,91 @@
...
+    :param event_translation: The event translation object which should be converted
+    :type event_translation: ~cms.models.events.event_translation.EventTranslation
+
+    :return: return data necessary for API
+    :rtype: dict
+    """
+
+    event = event_translation.event
+    if event.location:
+        location = transform_poi(event.location, event_translation.language.slug)
```
timoludwig, line 44:
```suggestion
        location_translation = (
            event.location.get_public_translation(event_translation.language.slug)
            or event.location.default_translation
        )
        location = transform_poi(event.location, location_translation)
```
Keep in mind that we don't know if there is a translation of the location in the same language as the event translation. Thus, we have to add a fallback to the default translation of the POI, otherwise you might get an `AttributeError: 'NoneType' object has no attribute 'title'`.

timoludwig (commented):
Thanks!
In general, it looks good. I like the idea of using `transform_poi()` for both the POI and the events API.

src/api/v3/events.py
```diff
@@ -0,0 +1,95 @@
+from django.http import JsonResponse
+
+from backend.settings import WEBAPP_URL, TIME_ZONE
```
timoludwig, line 3:
```suggestion
from backend.settings import WEBAPP_URL, CURRENT_TIME_ZONE
```
`TIMEZONE` is the timezone of our server, which will always be `UTC`. The setting we're looking for is `CURRENT_TIME_ZONE`, which is the timezone we expect our users to be in (and thus also the events which are created by them). Maybe the event model will be extended with it's own timezone setting in the future (see #663), but until then, I sugges to use `CURRENT_TIME_ZONE`.

src/api/v3/events.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    return {
+        "id": event.id,
+        "start_date": event.start_date,
+        "end_date": event.end_date,
+        "all_day": event.is_all_day,
+        "start_time": event.start_time,
+        "end_time": event.end_time,
+        "recurrence_id": event.recurrence_rule.id if event.recurrence_rule else None,
+        "timezone": TIME_ZONE,
```
timoludwig, line 27:
```suggestion
        "timezone": CURRENT_TIME_ZONE,
```

timoludwig (commented):
Thanks for your changes!
Sorry, I missed one thing in my review, see my comment below.

Apart from that: I also completely forgot that we have to calculate the single instances for recurring events, but I will open a new issue for this after this PR has been merged.

src/api/v3/events.py
```diff
@@ -0,0 +1,95 @@
...
+
+    :param language_slug: The slug of the requested language
+    :type language_slug: str
+
+    :return: JSON object according to APIv3 events endpoint definition
+    :rtype: ~django.http.JsonResponse
+    """
+    region = Region.get_current_region(request)
+    result = []
+    for event in region.events.filter(archived=False):
```
timoludwig, line 88:
```suggestion
    for event in region.events.filter(archived=False, end_date__gte=timezone.now()):
```
Please only return events in the future (or those currently taking place)...

src/api/v3/events.py
```diff
@@ -0,0 +1,95 @@
+from django.http import JsonResponse
+
```
timoludwig, line 2:
```suggestion
from django.utils import timezone

```

timoludwig (commented):
Thanks for your changes! :+1:

I'm so sorry, but I forgot yet another thing :neutral_face: 
But after that, I hopefully won't find anything to complain about :grinning:

src/api/v3/events.py
david-venhoff: Oh! I wasn't aware that this is a requirement.

timoludwig: Yes, I'm sorry I forgot this in the issue description and it's not documented, it's just how it's [currently implemented in WordPress](https://github.com/Integreat/cms/blob/30099ce9cdf76f72c34501fa77c7cfc6923ef935/wp-content/plugins/ig-wp-api-extensions/endpoints/v3/APIv3_Posts_Events.php#L28).

timoludwig (approved):
Thank you! :100:

Issue #729, opened 2021-03-04
Change `LanguageTreeNode` permissions


svenseeberg:
Give only superusers (not necessarily all staff) the permission to delete the language tree nodes - probably blocked by #684

Issue #696, opened 2021-02-23
Improve user list & form


timoludwig:
### Proposed Solution

List:

- Add role column
- ~Show email incl. email verified field~

Form:

- Fix translation of "Roles" label
- Make multi roles select field to single select
- Make role obligatory for non-superuser users

ulliholtgrave:
Files that need adjustments are:
src/cms/templates/users/region/list.html (Create new roles / email verified columns)
src/cms/templates/users/region/list_row.html (Load user data in new columns)
src/cms/models/users/user_profile.py (To adjust the property of the roles for a user)
src/cms/forms/users/user_form.py (Adjust user fields & **Translation**)
src/cms/models/users/user_profile.py (Adjust user field)

Issue #684, opened 2021-02-22
Set default permissions for staff members


timoludwig:
### Proposed Solution

Staff members should be granted permissions to manage all network resources, except:

- Delete regions
- Delete languages
- Delete organization
- Delete offer templates
- Delete roles
- Delete users

melegiul:
I've looked a bit in the documentation and I found that the staff field of user object is just a boolean, which allows to enter the built-in django admin site ([django:staff](https://docs.djangoproject.com/en/3.1/ref/contrib/auth/#django.contrib.auth.models.User.is_staff)). So I would propose two steps to address this issue:
1. Create a role named staff to assign all permission except for the above listed ones. At the moment our roles are defined in the file `cms.fixtures.roles.json` so I could append this staff role to this file.
2. Assign this role to users, which enable the staff checkbox in the admin user form.

So do I understand this issue right or is there another recommended approach?

timoludwig:
> So do I understand this issue right or is there another recommended approach?

Yes, I think this is the correct way to go. We have to think about one problem though: We don't want to make it possible to assign non-staff users to this special role. And I don't think we should hardcode this role to be excluded from the set of normal roles etc...
So I thought maybe it's best to create a new model with a onetoone field to the original `Group` model (similar to the `User` and `UserProfile` models) where we store the information which roles are meant for staff members and which aren't... This way, we can also add multiple roles for staff members if we need to add more fine-graded access control.
I started with this idea on the branch [feature/staff-roles](https://github.com/Integreat/integreat-cms/tree/feature/staff-roles) (see commit 5a23a5b5343b550c14cdc7837d74c5f3d1a53ff6 "Add support for staff and non-staff roles"), so if you think this makes sense, you can start your work off of this branch.

melegiul:
I didn't know that you already working on this issue and you made quite some progress. So you introduce a new boolean field to the group model to indicate that staff role has some special meaning, makes sense to me. Feel free to go ahead, until now I have not written any code yet.

timoludwig:
> I didn't know that you already working on this issue and you made quite some progress. So you introduce a new boolean field to the group model to indicate that staff role has some special meaning, makes sense to me. Feel free to go ahead, until now I have not written any code yet.

No, I haven't been working on it before you commented, I just wanted to try out my idea yesterday to see if it works... Not that I suggest something which doesn't 😄
I didn't mean to steal the issue from you, if you want to continue, feel free to do so 😉

[...]

Issue #172, opened 2019-06-29
Meta-css-classes for UI elements


timoludwig:
Currently, we use sets of ~10 css classes per UI element. If we want to make small changes to the elements (e.g. change the appearance of all buttons), we have to change these classes in all buttons in every template manually.
This is not practically, thus we need some kind of meta-classes to bundle these stylings and can be managed centralized. If necessary, include a webpack etc.
This also has the advantage of making the html source code much more readable.

sascha11110:
Should be fixed by https://github.com/Integreat/cms-django/commit/645dbf4ae4fb531306ff79232e9e2f55e235c915

timoludwig:
Actually, I meant something else by this issue: At the moment, we use certain combinations of css classes very often. I would like to add meta-classes which compile to a set of classes, so we have like "integreat-input-field" and "integreat-button", and when we want to change the appearance of all buttons, we only have to change this meta-class... do you know what I mean? Is that possible?

The commit you're mentioning fixes #280.

timoludwig:
I started working on this on the branch [feature/tailwind-css](https://github.com/Integreat/cms-django/tree/feature/tailwind-css)... So anyone interested in this issue can continue there.

svenseeberg:
I think we should postpone this to the Additional Features milestone. We do not have any problems with the current status.

timoludwig:
That would be fine with me.
Just for reference: Here is a good guide on how to achieve de-duplication of css classes:
https://tailwindcss.com/docs/extracting-components
(TL;RD: Try to extract components into sub-templates first, only use meta-classes where strictly necessary)

PR #838, opened 2021-05-17
Refactor forms and permissions

timoludwig:
### Short description

I refactored both the backend and the frontend of the forms and added more fine-graded permissions

### Proposed changes

- Centralized CSS for form fields
  - Add [@tailwindcss/forms](https://github.com/tailwindlabs/tailwindcss-forms) plugin for decent base styling
  - Put additional styling in the stylesheet
  - Remove duplicated css classes in html code
  - Make form styling more consistent
- Add support for staff and non-staff roles
  - Add dedicated Role model with one to one mapping to `django.contrib.auth.models.Group`
  - Add boolean field `staff_role`
- Refactor forms
  - Set additional instance attributes in `CustomModelForm`
- Refactor Permissions
  - Use more fine-graded permissions for all models
  - Use inbuilt permissions where possible
  - Check for permissions instead of superuser/staff status

### Resolved issues

Fixes (partially): #172
Fixes: #684
Fixes: #696
Fixes: #729


timoludwig:
I'm aware that this PR cannot be reviewed in detail. But I suggest to do a superficial review and merge this PR soon anyway instead of leaving it open for weeks until somebody finds the time to check it extensively...

melegiul (changes requested):
Awesome, getting rid of some css classe sounds like a great plan :rocket: 
But there is one thing I stumbled over on my system, just concerning the test data fixtures.
Locally running the tests fails due to HTTP 403 code in seven cases (all located in RegionViewTest):
Screenshot from 2021-07-02 20-05-43
One way to let tests succeed is to extend the permissions of our test user with the role "management" by adding to `src/cms/fixtures/roles.json`:
- 'Can view media file' (46)
- 'Can view page' (22)
- 'Can view location' (27)

But I wonder why circleci is not detecting this permission problem :thinking:

timoludwig:
> One way to let tests succeed is to extend the permissions of our test user with the role "management" by adding to `src/cms/fixtures/roles.json`
>
> But I wonder why circleci is not detecting this permission problem thinking

Actually, the management role on this branch has this permission already, but unfortunately permissions are changing their ids each time the database changes, so for now there is no way of persisting these in an existing database.
To fix the permissions on your machine locally, you have to execute `./dev-tools/prune_database.sh` before loading the new roles fixture. In the future, this problem will be fixed by resolving issue #748.

melegiul (approved):
Thank you very much, I was not aware of this issue so pruning the database worked for me :tada:

Issue #813, opened 2021-05-07
Remove `PushNotificationChannel` model


timoludwig:
### Motivation

At the moment, we have only one push notification channel, and also in the future we will probably have only a few and they won't change often. Thus, we won't store in the database, but in the settings file instead.

### Proposed Solution

- [ ] Remove `PushNotificationChannel` model
- [ ] Add setting (list) for push notification channels, at the moment only the channel "news" exists
- [ ] Remove the fields [Region.push_notification_channels](https://integreat.github.io/integreat-cms/ref-ext/cms.models.regions.html#cms.models.regions.region.Region.push_notification_channels) and [RegionForm.push_notification_channels](https://integreat.github.io/integreat-cms/ref-ext/cms.forms.regions.html#cms.forms.regions.region_form.RegionForm)
- [ ] Change the field [PushNotification.channel](https://integreat.github.io/integreat-cms/ref-ext/cms.models.push_notifications.html#cms.models.push_notifications.push_notification.PushNotification.channel) to a `CharField` with the choices from the settings file

PR #845, opened 2021-05-24
Remove PushNotificationChannel model

MizukiTemma:
### Short description

This PR contains changes for the issue Remove PushNotificationChannel model 

### Proposed changes

- The model PushNotificationChannel is deleted.
- The fields Region.push_notification_channels and RegionForm.push_notification_channels are removed.
- The field PushNotification.channel is changed to a CharField with the choices.
- The choices are added in the setting file.

### Resolved issues

Fixes: #813


melegiul (changes requested):
Very cool new choice field :rocket:
I just could find some old references to the regions push notifications field:
1. In the file src/cms/fixtures/test_data.json there are two regions, where push notification channel needs to be removed
2. In file src/cms/tests/pages/models.py there is also a region created with the channel field
3. Same in src/cms/tests/views/region_view_test.py

Then the tests should succeed and this should be ready to merge :tada:

melegiul (approved):
Awesome, this fixes the tests and the fixtures :tada:
So this one should be now ready to merge :+1:

Issue #824, opened 2021-05-12
Dashboard widget for Broken link checker


melegiul:
### Motivation

Currently on the dashboard there is only a link to the broken link checker, but there should be also some summary about the results of the broken link checker.

### Proposed Solution

The dashboard widget needs a concise summary (not a list of all the links), e.g. number of links in state "broken", "valid", etc.

### Additional Context

There are also some other dashboard widget (statistics) for reference of the expected scope.
The dummy template is in templates/analytics.

![Image Pasted at 2021-5-12 13-16](https://user-images.githubusercontent.com/47010475/117969428-ab104a00-b327-11eb-9422-282e8621191f.png)

### Current wordpress version

![Image Pasted at 2021-5-12 13-14](https://user-images.githubusercontent.com/47010475/117969349-959b2000-b327-11eb-98c4-d6281508eb0d.png)

PR #847, opened 2021-05-24
Adds brief summary about linkcheck to dashboard

melegiul:
### Short description

Adds brief summary about linkcheck to dashboard

### Proposed changes

Briefly reports the number of valid, invalid, etc. links on dashboard and analytics overview

### Screenshot
Screenshot from 2021-05-24 17-18-43

### Resolved issues

Fixes: #824

### DIFF
--- a/src/cms/templates/analytics/_broken_links_widget.html
+++ b/src/cms/templates/analytics/_broken_links_widget.html
@@ -14,14 +14,8 @@
             {% trans "Details" %}
         </a>
     </div>
-    <div class="mx-8">
-    <table class="w-1/2 pl-8">
-        <thead>
-            <tr>
-                <th class="w-1/2"></th>
-                <th class="w-1/2"></th>
-            </tr>
-        </thead>
+    <div class="mt-4">
+    <table class="w-full max-w-xl">
         <tbody>
             <tr>
                 <td>{% trans 'Invalid links' %}</td>
@@ -29,7 +23,7 @@
             </tr>
             <tr>
                 <td>{% trans 'Valid links' %}</td>
-                <td class="text-right">{{ invalid_links }}</td>
+                <td class="text-right">{{ valid_links }}</td>
             </tr>
             <tr>
                 <td>{% trans 'Ignored links' %}</td>
--- a/src/cms/templates/analytics/_broken_links_widget.html
+++ b/src/cms/templates/analytics/_broken_links_widget.html
@@ -14,7 +14,36 @@
             {% trans "Details" %}
         </a>
     </div>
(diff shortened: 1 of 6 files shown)


src/cms/templates/analytics/_broken_links_widget.html
```diff
@@ -14,7 +14,36 @@
             {% trans "Details" %}
         </a>
     </div>
-    <div class="p-3 w-full">
-        <!-- TODO -->
+    <div class="mx-8">
```
timoludwig, line 17:
```suggestion
    <div class="mt-4">
```
I would rather add a small margin to the top instead of to the sides.

src/cms/templates/analytics/_broken_links_widget.html
```diff
@@ -14,7 +14,36 @@
...
+    <table class="w-1/2 pl-8">
```
timoludwig, line 18:
```suggestion
    <table class="w-full max-w-xl">
```
I think this looks better on small screen sizes...

src/cms/utils/filter_links.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def filter_links(region_slug):
+    """[summary]
+
+    :param region_slug: [description]
+    :type region_slug: [type]
+    :return: [description]
+    :rtype: [type]
+    """
```
timoludwig, line 13: Could you update these docstrings please?

timoludwig (commented):
Thanks! :+1:
Looks good, I just have some minor suggestions.

src/cms/utils/filter_links.py
melegiul: Yes, this was a mistake, I've updated comments to something more meaningful

src/cms/views/analytics/analytics_overview.py
```diff
@@ -2,7 +2,10 @@
 from django.utils.decorators import method_decorator
 from django.views.generic import TemplateView
 
+from cms.utils.filter_links import filter_links
+
```
timoludwig, line 6:
```suggestion
```
`filter_links` is already imported in line 8 below. I wonder why pylint isn't complaining about this :thinking:

timoludwig (commented):
Thanks! Just a tiny detail:

src/cms/views/analytics/analytics_overview.py
melegiul: This was not my intention, but both have different paths
Seems like a way to trick pylint :joy:

timoludwig (approved):
:+1:

Issue #629, opened 2021-01-12
Add possibility to refresh `last_updated` date


timoludwig:
### Motivation

If region managers want to refresh the changed date to make sure the page doesn't look outdated to the users when in fact it isn't, they have to change anything in the content to make the form save. Otherwise the form is not saved with the message "No changes detected" or similar. 

### Proposed Solution

~Add the possibility to update the date without changing the content.~
~E.g. Show the last update date and a button "Refresh" next to it, which changes the date to the current moment.~

### Alternatives

Make it possible to submit the form without changes to update the date. 
Adapt the message to show something like "No changes detected, but date updated" or similar.

### Additional Context

Optionally add a bulk action to update the changed date of all selected pages

svenseeberg:
I vote for the alternative. I think we should keep the buttons to a minimum. One alternative I can think of: We could change the caption of the "Save" button. For example if no changes were made. the caption is "Refresh Date", and if changes were made, the caption changes to "Save changes". Something like that.

timoludwig:
Ok. I think changing the title of the "save"-button might be confusing, but I think it should be clear what happens when we adapt the success message which is shown.

MizukiTemma:
I think, too, it is confusing if the button is renamed (the editor may think they cannot change the page contents anymore when they see "Refresh date" but no "update") and more intuitiv to refresh the date is clicking update/aktuasilieren (not renamed button) without any change.

Shall I even so let the buttun name be changing depend on there is any change on the page?

MizukiTemma:
P.S. After clicking update without change the refresh confirmation message will be shown. I thinki this is necessary.

PR #849, opened 2021-05-27
Add possibility to refresh last_updated date

MizukiTemma:
### Short description

It is now possible to update a page without any change to refresh the last update date.

### Proposed changes

- By clicking the update/aktualisieren button the last update date of the page will be refreshed to the current time, even if there is no change made in the page contents.
- After refreshing a message is shown ("no changes detected but date refreshed") .

- The button name does not change (stays always "Update/Aktualisieren") in the current commit, because I think changing the button name is probably confusing (Please see the conversation on the issue page). If the changing name is even though wished, I will add it.

### Resolved issues

Fixes: #629

### DIFF
--- a/src/cms/locale/de/LC_MESSAGES/django.po
+++ b/src/cms/locale/de/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2021-05-29 15:29+0000\n"
+"POT-Creation-Date: 2021-05-31 17:37+0000\n"
 "Last-Translator: Integreat <info@integreat-app.de>\n"
 "Language-Team: Integreat <info@integreat-app.de>\n"
 "Language: German\n"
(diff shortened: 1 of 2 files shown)


src/cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -5404,11 +5403,15 @@ msgstr ""
 "Version <a href='%(revision_url)s' class='underline hover:no-"
 "underline'>Version %(revision)s</a> dieser Seite in der App angezeigt."
 
-#: views/pages/page_view.py:281
+#: views/pages/page_view.py:256
+msgid "No changes detected, but date refreshed"
+msgstr "Keine Änderungen vorgenommen, datum aktualisiert"
```
timoludwig, line 5408:
```suggestion
msgstr "Keine Änderungen vorgenommen, aber Datum aktualisiert"
```

src/cms/views/pages/page_view.py
```diff
@@ -275,25 +262,28 @@ def post(self, request, *args, **kwargs):
...
-            if published:
-                messages.success(
-                    request, _("Translation was successfully created and published")
-                )
-            else:
-                messages.success(request, _("Translation was successfully created"))
-        else:
-            if published:
-                messages.success(request, _("Translation was successfully published"))
+        if not unchanged:
```
timoludwig, line 265:
```suggestion
        if not page_form.has_changed() and not page_translation_form.has_changed():
            messages.info(request, _("No changes detected, but date refreshed"))
        else:
```
I think this can be solved without the additional variable.

src/cms/views/pages/page_view.py
```diff
@@ -248,25 +248,12 @@ def post(self, request, *args, **kwargs):
...
-                    "right_to_left": language.text_direction
-                    == text_directions.RIGHT_TO_LEFT,
-                },
-            )
+        unchanged = (
+            not page_form.has_changed() and not page_translation_form.has_changed()
+        )
+
+        if unchanged:
+            messages.info(request, _("No changes detected, but date refreshed"))
```
timoludwig, line 257:
```suggestion
```

timoludwig (commented):
Thank you! :+1:

I think this looks very good - changing the message but not the button should provide the best user experience.

I just have a small request regarding the translation and the re-arrangement of the "unchanged"-message.

timoludwig (approved):
Thanks! :+1:

Issue #846, opened 2021-05-24
Remove unused `created_date` & `last_updated` fields


svenseeberg:
### Motivation
We currently have a `last_updated` in the Page model and a `created_date` in the PageTranslation model which are not really used. There is no functional need for these dates. To the contrary: displaying all these dates would be confusing for users.

The `last_updated` date of a page can be considered the newest `last_updated` date of all its translations and the `created_date` of a page translation is in most cases the `last_updated` date of the first version of this translation, so if this fields are required in the future, they can be replaced by properties.

### Proposed Solution
- Remove the `last_updated` field from the content models:
  - `Page` model
  - `PushNotification` model
- Remove the `created_date` field from the translation models:
  - `PageTranslation` model
  - `EventTranslation` model
  - `POITranslation` model
  - `PushNotificationTranslation` model

### Alternatives
Keep as is.

PR #850, opened 2021-05-29
Remove unused created_date and last_updated fields

david-venhoff:
### Short description
This pr removes the `created_date` and `last_updated` fields from some models as specified in #846.
Additionally, this pr removes the `created_date` field from `ImprintPageTranslation` and the `last_updated` field from `ImprintPage`.

### Resolved issues
Fixes: #846

### DIFF
--- a/src/cms/fixtures/test_data.json
+++ b/src/cms/fixtures/test_data.json
@@ -279,7 +279,6 @@
       "version": 1,
       "minor_edit": false,
       "creator": 1,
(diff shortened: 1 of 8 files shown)


timoludwig (approved):
Thanks, this looks good! :+1:

You we're right to remove the fields from the `AbstractBasePage` and `AbstractBasePageTranslation` model instead of `Page` and `PageTranslation`.

Issue #825, opened 2021-05-12
Set/unset the state of links to ignored


melegiul:
### Motivation

Set/unset the state of links to ignored

### Proposed Solution

Like in the wordpress version of the broken link checker, user need a feature to transfer the state of e.g. invalid/valid links into state "ignore" or from "ignore" back to "invalid/valid".
The wordpress version seemed to support this feature by checkboxes and bulk actions.

### Alternatives

Each row of the link table could contain a button to ignore this single link, but ignoring a group of links get tedious this way.

### Additional Context

![image](https://user-images.githubusercontent.com/47010475/117967965-f32e6d00-b325-11eb-8cb7-17667835bc4a.png)

PR #851, opened 2021-06-01
Adds actions to brokenlink checker

melegiul:
### Short description

Adds actions to brokenlink checker

### Proposed changes

- Ignore, unignore and recheck actions added
- Link selection via checkboxes
- Only non-ignored links can be either ignored or rechecked
- Ignored links can only execute action "unignore"

### Resolved issues

Fixes: #825

### DIFF
--- a/src/cms/views/linkcheck/linkcheck.py
+++ b/src/cms/views/linkcheck/linkcheck.py
@@ -1,8 +1,12 @@
+from django.contrib import messages
 from django.db.models import OuterRef, Subquery
 from django.urls import reverse
+from django.shortcuts import redirect
+from django.utils.translation import gettext as _
 from django.views.generic import ListView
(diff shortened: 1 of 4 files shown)


src/cms/views/linkcheck/linkcheck.py
```diff
@@ -63,6 +67,39 @@ def get_queryset(self):
         )
         return page_links.union(event_links, poi_links)
 
+    def post(self, request, *args, **kwargs):
+        """
+        Applies selected action for selected links
+
+        :param request: The current request
+        :type request: ~django.http.HttpResponse
```
timoludwig, line 75:
```suggestion
        :type request: ~django.http.HttpRequest
```

timoludwig (commented):
Thanks! Looks good and works as expected :+1:

Issue #852, opened 2021-06-01
XLIFF export uses debug serve function


timoludwig:
### Motivation

In the XLIFF export, the debugging function [serve()](https://github.com/Integreat/integreat-cms/blob/0dae73237b23fe5b8c1bc97a7d95ad340ed19af1/src/cms/views/pages/page_actions.py#L336-L338) is used to deliver the XLIFF files to the user.
> **Warning**
>
> This view will only work if `DEBUG` is `True`.
>
> That’s because this view is **grossly inefficient and probably insecure**. This is only intended for local development, and should **never be used in production**.

– [Django Documentation](https://docs.djangoproject.com/en/3.2/ref/contrib/staticfiles/#django.contrib.staticfiles.views.serve)

### Proposed Solution

Use a custom [FileSystemStorage](https://docs.djangoproject.com/en/3.2/ref/files/storage/#the-filesystemstorage-class) and serve XLIFF files like the other static files via the Apache/Nginx webserver.

### Alternatives

Serve the file as normal `HttpResponse` by manually setting the correct headers

svenseeberg:
I think I'd prefer the `HttpResponse` solution. While the content is probably not really sensible, it is quite possible that there is unpublished content which should not be available publicly. Therefore it would be nice to serve the content in a protected end point.

timoludwig:
> I think I'd prefer the `HttpResponse` solution. While the content is probably not really sensible, it is quite possible that there is unpublished content which should not be available publicly. Therefore it would be nice to serve the content in a protected end point.

Do you think putting each file in a random UUID directory might suffice as "not publicly available" anymore? From a performance standpoint, I'd prefer the static files solution. But sure, requiring the login would be bullet-proof and since the zip only contains text-files, the file size is probably manageable by Django.

svenseeberg:
Okay. Lets hope that nobody enables directory listing by accident ;-)

Issue #734, opened 2021-03-08
Support XLIFF version 1.2


timoludwig:
### Motivation

At the moment, we can only import and export XLIFF files of version 2.0. To support the legacy import of old translation and the export for translation services which did not yet adopt the new standard, it would be beneficial to also allow conversions into the old format.

### Proposed Solution

There is an [xliff serializer for Django](https://github.com/callowayproject/django-xliff) which works with version 1.2. Since this package hasn't been updated since a few years, we should make sure it doesn't have any flaws and check whether everything works with the new Django version 3.1. But apart from that, it shouldn't be that much effort to get this converter to work. Probably it makes sense to include an adapted version of the code into our repository instead of including an outdated dependency.

### Alternatives

Just hope that we'll never need to import/export anything in the old format...

svenseeberg:
We also discussed to just use the basic XML serializer and simply create dictionaries/objects in the necessary structure.

Issue #722, opened 2021-03-01
Region checks missing in XLIFF import


timoludwig:
### Describe the Bug

It is possible to manipulate pages of all regions with the XLIFF import

### Steps to Reproduce

1. Export any page as XLIFF file
2. Edit the page id in the `<file original="{page_id}">` field
3. Import the XLIFF again
4. See that contents are applied to the new page, regardless of the page's region

### Expected Behavior

It should not be allowed to circumvent permission checks with the XLIFF import

### Actual Behavior

A user who can edit pages one one specific region can edit all pages in all regions

Issue #455, opened 2020-07-29
Improve XLIFF import confirmation page


timoludwig:
### Motivation

The XLIFF confirmation page has already decent functionality, but may be improved in the future.

### Proposed Solution

- Investigate about unicode encodings and how to dispay diffs accordingly (e.g. monospaced font instead of `<pre>`-tag)
- Think about showing the rendered html instead of the diff of the raw html source code (e.g. [htmldiff-js](https://www.npmjs.com/package/htmldiff-js))
- Test whether a side-by-side view has advantages here

svenseeberg:
I think we could see this as having two different modes:

* Advanced mode that actually shows the byte diff for users that know what is meant, including the normal unified diff format.
* Simplified view that shows the "visible" diff. That means marks removed words red and new word green and does not show the source code.

timoludwig:
Since we included htmldiff-js in our repo now, you can have a look at [revisions.js](https://github.com/Integreat/cms-django/blob/develop/src/cms/static/js/revisions.js) how to use it.
Keep in mind that you have to browserify the js you use in [run.sh](https://github.com/Integreat/cms-django/blob/b3c4ccfbd32c7720c6cc6886d578cee781e36abf/dev-tools/run.sh#L40).

PR #853, opened 2021-06-01
Refactor XLIFF import/export

timoludwig:
### Short description

This PR refactors the XLIFF import/export and resolves all current XLIFF issues.

### Proposed changes

- Make use of Django's unbuilt XML Serializer
- Support both XLIFF 1.2 and 2.0
- Show rendered HTML import Diff

### Resolved issues

Fixes: #454
Fixes: #455
Fixes: #722
Fixes: #734
Fixes: #852


src/cms/models/languages/language.py
```diff
@@ -106,6 +106,18 @@ def translated_name(self):
...
 
+    def get_source_language(self, region):
+        """
+        This returns the source language of this language in the given region
+
+        :return: The source language of the language
+        :rtype: ~cms.models.languages.language.Language
+        """
+        language_tree_node = self.language_tree_nodes.filter(region=region).first()
+        if language_tree_node and language_tree_node.parent:
```
svenseeberg, line 117: Why is the first item in the queryset expected to be the parent? Wouldn't it be necessary to iterate over all languages until the parent is found?

src/cms/templates/pages/page_tree.html
```diff
@@ -109,32 +109,44 @@ <h1 class="heading">{% trans 'Pages' %}</h1>
...
+              action="{% url 'upload_xliff' region_slug=region.slug language_slug=language.slug %}">
+            <label id="xliff_file_label" for="xliff_file" class="inline-block cursor-pointer bg-blue-500 hover:bg-blue-600 text-white font-bold py-2 px-4 m-0 rounded">
+                <i data-feather="upload" class="inline-block pr-1"></i>
+                {% trans 'Select files' %}
+            </label>
+            <span id="xliff_file_label_multiple" class="hidden"> {% trans 'and {} other files' %}</span>
+            <input id="xliff_file" type="file" name="xliff_file" class="hidden" accept=".zip, .xlf, .xliff" multiple>
+            {% csrf_token %}
+            <input id="xliff_file_submit" type="submit" value="{% trans 'Import' %}"
+                   class="inline-block bg-gray-500 text-white font-bold py-2 px-4 rounded cursor-not-allowed" disabled/>
```
svenseeberg, line 147: I'm not sure I like the Python/Black style indentation with HTML ;-) Works though.

src/cms/models/languages/language.py
timoludwig: `self` is an instance of the `Language` model, so `self.language_tree_nodes` is the queryset of all all instances of the `LanguageTreeNode` model which are mapped to this language (but across all regions). Since there cannot be two language tree nodes with the same language and the same region (because of [this constraint](https://github.com/Integreat/integreat-cms/blob/4e5560ddd0a2cd4708ae544b36358eb835bec41b/src/cms/models/languages/language_tree_node.py#L186-L191)), we know that the queryset `self.language_tree_nodes.filter(region=region)` can contain only one element (or none if the language doesn't exist in the requested region).
But yeah, I'll add a comment for better understanding.

src/cms/templates/pages/page_tree.html
timoludwig: I don't this this has anything to do with black, I just indented it manually this way. We don't have a clear formatting guide on html & js files, if you have any opinion on this, feel free to comment #749 with your suggestions. I'm also not happy with the current situation of our html files^^

src/cms/views/pages/page_xliff_import_view.py
```diff
@@ -0,0 +1,147 @@
...
+        self.language = get_object_or_404(
+            self.region.languages, slug=kwargs.get("language_slug")
+        )
+        # Get directory path of the uploaded XLIFF files
+        self.xliff_dir = os.path.join(XLIFF_UPLOAD_DIR, kwargs.get("xliff_dir"))
+
+        if not os.path.isdir(self.xliff_dir):
+            messages.error(
+                request,
+                _("This XLIFF import is no longer available."),
```
svenseeberg, line 88: In which circumstances would we expect this to happen? Reloading the import URL?

svenseeberg: AFAICT this should never happen?! I'd present a more dramatic message to the user. Something like "Something went wrong. If this occurs again, please contact the support team".

src/cms/templates/pages/page_tree.html
svenseeberg: I think I can live with any indentation. This was just ... unexpected :)

src/cms/views/pages/page_xliff_import_view.py
timoludwig: Reloading should not change anything, I didn't add a mechanism to delete XLIFF files from the server (at least in the migration phase I thought it might be useful to know exactly which files have been uploaded and when). Maybe a cronjob could regularly delete old XLIFF files which are older than a specific time? Or when everything works super stable, we can also delete the files right after a successful (or intentionally aborted) import...

svenseeberg (commented):
Overall the code looks good in my opinion. I'll test it tomorrow.

src/cms/models/languages/language.py
svenseeberg: You're right. I was not thinking hard enough about the language tree. That makes totally sense.

svenseeberg:
When the browser width is only about 1000px, the bulk action button is sort of broken:

image

svenseeberg:
When importing a file with no title, the following message appears:

> Page with the title in German: "Kontakt zu App Team Augsburg" could not be imported successfully because of the errors: slug: This field cannot be blank., title: This field cannot be blank., creator: This field cannot be blank.

I would reduce this message to "Title is missing", or at least place the info about the title first.

svenseeberg (changes requested):
Works well, I'd only request a minor change: the error message with a missing title during import is too confusing for normal users in my opinion.

Issue #786, opened 2021-04-19
Admin menu visible in user settings


timoludwig:
### Describe the Bug

Since the user settings are region-independent, we didn't include the region parameter in the [user_settings](https://github.com/Integreat/integreat-cms/blob/0d623d97bfd2bc91765fbc3e93c4f37a7f15b0dc/src/cms/urls.py#L258) url.
However, even for non-staff users, this means there are links to the admin dashboard which shouldn't be the case.

### Steps to Reproduce

1. Create a non-staff user
2. Go to the user settings

### Expected Behavior

The menu should correspond to one region of the user. Ideally the same region in which the user was before he clicked on the user settings

### Actual Behavior

Even non-staff users are in the network settings

### Additional Information

Probably it would be sufficient to add a new url route which contains the `region_slug` paramter if users visit the user settings coming from a region. If possible, admin users should still see the admin menu when they access the user settings while they are in the network settings.

![Screenshot_2021-04-19 Integreat Redaktionssystem](https://user-images.githubusercontent.com/16021269/115249843-4f3c1200-a129-11eb-930c-6bc78f5e6657.png)

PR #856, opened 2021-06-03
Admin menu visible in user settings #786

MizukiTemma:
### Short description

This PR modifies the link of the user setting so that the network setting/admin dashboard is not visible to the non-staff users. 

### Proposed changes

- added region slug in the link to the user settings.
-
-

### Resolved issues

Fixes: #786

### DIFF
--- a/src/cms/templates/_base.html
+++ b/src/cms/templates/_base.html
@@ -61,13 +61,13 @@
         </div>
     </div>
     <div id="user-info" class="relative px-2 text-gray-800 flex flex-col justify-center cursor-pointer hover:bg-gray-200">
-        <a href="{% url 'user_settings' %}" class="relative block whitespace-nowrap">
+        <a href="{% if region %}{% url 'user_settings' region_slug=region.slug %}{% else %}{% url 'user_settings' %}{% endif %}" class="relative block whitespace-nowrap">
             <i data-feather="user"></i>
             {{ request.user.profile.full_user_name }}
             <i data-feather="chevron-down"></i>
         </a>
         <div id="user-menu" class="absolute hidden shadow rounded-b top-full right-0 bg-gray-200">
-            <a href="{% url 'user_settings' %}" class="relative block whitespace-nowrap px-4 py-3 text-gray-800 hover:bg-gray-400">
+            <a href="{% if region %}{% url 'user_settings' region_slug=region.slug %}{% else %}{% url 'user_settings' %}{% endif %}" class="relative block whitespace-nowrap px-4 py-3 text-gray-800 hover:bg-gray-400">
                 <i data-feather="settings"></i>
                 {% trans 'User Settings' %}
             </a>
--- a/src/cms/urls.py
+++ b/src/cms/urls.py
@@ -938,6 +938,11 @@ urlpatterns = [
                         ]
                     ),
                 ),
+                url(
+                    r"^user_settings/$",
+                    settings.UserSettingsView.as_view(),
+                    name="user_settings",
+                ),
             ]
         ),
     ),
--- a/src/cms/views/settings/mfa/register_user_mfa_key_view.py
+++ b/src/cms/views/settings/mfa/register_user_mfa_key_view.py
@@ -87,4 +87,14 @@ class RegisterUserMfaKeyView(CreateView):
                 'The 2-factor authentication key "{}" was successfully registered.'
             ).format(new_key.name),
         )
-        return JsonResponse({"success": True, "successUrl": reverse("user_settings")})
+        # Determine success url
+        if request.user.is_superuser or request.user.is_staff:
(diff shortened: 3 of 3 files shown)


src/cms/urls.py
```diff
@@ -938,6 +938,45 @@
...
+                                            name="register_new_mfa_key",
+                                        ),
+                                        url(
+                                            r"^delete/(?P<key_id>\d+)$",
+                                            settings.DeleteUserMfaKeyView.as_view(),
+                                            name="delete_mfa_key",
+                                        ),
+                                    ]
+                                ),
+                            ),
```
melegiul, line 976: I think in the moment we do not need this block, because the 2FA button of the user settings for now points to `/user_settings/mfa/register` and not to `{region_slug}/user_settings/mfa/register`
You can see this in href of the following link tag, where region slug is not included:
https://github.com/Integreat/integreat-cms/blob/ed22b124c22912e78bc6cc6e4c6eff1e9635cff1/src/cms/templates/settings/user.html#L118
And because 2FA is not part of the issue, I think we can just skip this block

melegiul (changes requested):
Awesome, this solves the issue about the menu bar :tada: 
I have just one idea for improvement, then this should be ready to merge.:+1:

melegiul (approved):
Thank you very much :+1:
Just one missing rebase and this one is ready to merge :rocket:

src/cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -5001,9 +5001,7 @@ msgstr ""
 
 #: views/events/event_list_view.py:95
 msgid "You don't have the permission to edit or create events."
-msgstr ""
-"Sie haben nicht die nötige Berechtigung, um Veranstaltungen zu bearbeiten "
-"oder zu erstellen."
+msgstr ", um Veranstaltungen zu bearbeiten oder zu erstellen."
```
timoludwig, line 5004:
```suggestion
msgstr ""
"Sie haben nicht die nötige Berechtigung, um Veranstaltungen zu bearbeiten "
"oder zu erstellen."
```
This was probably just a mistake, right?

src/cms/templates/_base.html
```diff
@@ -61,16 +61,32 @@
...
+                {{ request.user.profile.full_user_name }}
+                <i data-feather="chevron-down"></i>
+            </a>
+        {% else %}
+            <a href="{% url 'user_settings' %}" class="relative block whitespace-nowrap">
+                <i data-feather="user"></i>
+                {{ request.user.profile.full_user_name }}
+                <i data-feather="chevron-down"></i>
+            </a>
+        {% endif %}
```
timoludwig, line 76:
```suggestion
        <a href="{% if region %}{% url 'user_settings' region_slug=region.slug %}{% else %}{% url 'user_settings' %}{% endif %}" class="relative block whitespace-nowrap">
            <i data-feather="user"></i>
            {{ request.user.profile.full_user_name }}
            <i data-feather="chevron-down"></i>
        </a>
```
I think, in this case I'd prefer the slightly more concise syntax. But your solution is also fine.

src/cms/templates/_base.html
```diff
@@ -61,16 +61,32 @@
...
+                    <i data-feather="settings"></i>
+                    {% trans 'User Settings' %}
+                </a>
+            {% else %}
             <a href="{% url 'user_settings' %}" class="relative block whitespace-nowrap px-4 py-3 text-gray-800 hover:bg-gray-400">
                 <i data-feather="settings"></i>
                 {% trans 'User Settings' %}
             </a>
+            {% endif %}
+
```
timoludwig, line 89:
```suggestion
            <a href="{% if region %}{% url 'user_settings' region_slug=region.slug %}{% else %}{% url 'user_settings' %}{% endif %}" class="relative block whitespace-nowrap px-4 py-3 text-gray-800 hover:bg-gray-400">
                <i data-feather="settings"></i>
                {% trans 'User Settings' %}
            </a>
```
here as well...

src/cms/urls.py
```diff
@@ -938,6 +938,18 @@
...
+                    r"^user_settings/",
+                    include(
+                        [
+                            url(
+                                r"^$",
+                                settings.UserSettingsView.as_view(),
+                                name="user_settings",
+                            ),
+                        ]
+                    ),
```
timoludwig, line 951:
```suggestion
                    r"^user_settings/$",
                    settings.UserSettingsView.as_view(),
                    name="user_settings",
```
Since we only have one url here, we can write this a little bit shorter.

timoludwig (commented):
Thank you! :tada:

This works as expected, I just have a few small suggestions to write things a little bit more concise.
I will apply my suggestions and rebase the branch in a moment.

timoludwig (approved):

Issue #842, opened 2021-05-18
Return all recurring events in API


timoludwig:
### Motivation

At the moment, only the first occurrence of recurring events is returned in the API. Since the App does not know how often a recurring event takes place, we have to expand a recurring event into all its occurrence.

### Proposed Solution

- Add a setting which denotes the timedelta up to which events should be returned.
- Calculate all occurrences of recurring events based on the recurrence rule and return all instances up to the defined maximum.
- Find a way to create unique ids for the events and event translations which do not exist in the database. (Since the app doesn't use the ids at the moment anyway, we might as well omit this if it is too hard to implement)
- Create unique slugs/paths by appending the date in the iso format to the slug of the first occurrence of the recurring event

### Additional Context

<details>
<summary>Example return JSON</summary>

```
  {
    "id": 4862,
    "url": "https://cms-test.integreat-app.de/testumgebung/de/events/wiederkehrende-test-veranstaltung-2021-05-18/",
    "path": "/testumgebung/de/events/wiederkehrende-test-veranstaltung-2021-05-18/",
    "title": "Wiederkehrende Test-Veranstaltung",
    "modified_gmt": "2021-05-14 18:54:50",
    "excerpt": "Test",
    "content": "<p>Test</p>\n",
    "available_languages": [],
    "thumbnail": null,
    "event": {
      "id": 169,
      "start_date": "2021-05-18",
      "end_date": "2021-05-18",
      "all_day": false,
      "start_time": "09:30:00",
      "end_time": "10:45:00",
      "timezone": "Europe/Berlin",
      "recurrence_id": "165"
    },
    "location": {
      "id": null,
      "name": null,
      "address": null,
      "town": null,
      "state": null,
      "postcode": null,
      "region": null,
      "country": null,
      "latitude": null,
      "longitude": null
    },
    "featured_image": null,
    "hash": "ab801b3eee92acf0da73d17ef810c38b"
  },
  {
    "id": 4863,
    "url": "https://cms-test.integreat-app.de/testumgebung/de/events/wiederkehrende-test-veranstaltung-2021-05-19/",
    "path": "/testumgebung/de/events/wiederkehrende-test-veranstaltung-2021-05-19/",
    "title": "Wiederkehrende Test-Veranstaltung",
    "modified_gmt": "2021-05-14 18:54:50",
    "excerpt": "Test",
    "content": "<p>Test</p>\n",
    "available_languages": [],
    "thumbnail": null,
    "event": {
[...]

PR #857, opened 2021-06-04
Return recurrent events in the events api

david-venhoff:
### Short description
This pr allows recurrent events to be returned by the events api.

### Proposed changes
Add a method `iter_after` to  the`RecurrenceRule` model to generate all valid dates covered by this rule after a start date.
Use this method in the events api endpoint to additionally return these events, up to a maximum time span.
These events and their translations will have an id of `null`. Additionally, a unique path will be created for each event translation.

### Resolved issues
Fixes: #842

### DIFF
--- a/src/api/v3/events.py
+++ b/src/api/v3/events.py
@@ -1,8 +1,15 @@
+from datetime import timedelta
 from django.http import JsonResponse
 from django.utils import timezone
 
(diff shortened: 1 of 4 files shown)


melegiul (approved):
Thank you very much, this works as expected :+1:
I especially like how you made use of pythons generators :heart: 
Very cool :rocket:

Issue #529, opened 2020-10-09
Page/Event-Translation should be allowed to be blank in creation/editing-form but not in the database


Thomas125:
### Describe the Bug

In the PageTranslation- and the EventTranslation-Model, the slug is allowed to be blank.
But no blank slug should be allowed in the database.

### Additional Information

In the creation/editing-form, no slug is required (which is the desired behaviour). In case no slug is specified, a unique slug is generated when the form is submitted.
For generating a unique slug, the method "clean_slug" is used.

### Proposed Solution

Remove the mentioned constraint in the two models. But make sure that one is still allowed to leave the fields empty in the form and that a slug is generated if it is left empty.

timoludwig:
This is also linked to #454, which is about the whole problem that our only validation is placed in the forms.

PR #858, opened 2021-06-12
Disallow empty slugs for event translations and page translations

david-venhoff:
### Short description
Currently, the two models `PageTranslation` and `EventTranslation` may contain empty slugs. This pr removes that possibility. 

### Proposed changes
- remove the `blank=True` parameter for slug field
- manually set the slug model field to be not required

### Resolved issues
Fixes: #529


timoludwig (commented):
Thanks for your PR! :+1:

I just noticed that the same problem exists for the following models, even though they're not mentioned in the issue:

- OfferTemplate
- POITranslation
- Region

Do you think you could add the same solution for these models as well? Thanks!

david-venhoff:
I included offer templates, poi translations and regions now.

timoludwig (approved):

Issue #855, opened 2021-06-02
Implement WordPress API endpoint URL patterns


svenseeberg:
### Motivation
With Django we decided to introduce a new, clean API URL pattern. However, the apps do not use this API yet. We already have exports of the content and the app team could start end to end testing, as soon as the app can interface with the CMS.

### Proposed Solution
We need to implement the URL pattern defined in https://github.com/Integreat/cms/wiki/REST-APIv3-Documentation
The WordPress URL pattern should be available in addition to the current (clean) URLs, so that the app team can switch over.

Please test the integration with the web app by [changing the back end URL](https://github.com/Integreat/cms/wiki/REST-APIv3-Documentation#test-api-with-webapp) to the Django system. If you need help finding issues or interpreting error messages, feel free to contact the app team for help.

### Alternatives
None, needs to be done.

svenseeberg:
Currently, the following URLs work:
* https://cms-django-dev.integreat-app.de/de/wp-json/extensions/v3/sites/
* https://cms-django-dev.integreat-app.de/wp-json/extensions/v3/augsburg/de/pages
* [...]

However, the region and language slugs need to be at the beginning of the URL:
* https://cms-django-dev.integreat-app.de/augsburg/de/wp-json/extensions/v3/pages
* https://cms-django-dev.integreat-app.de/augsburg/de/wp-json/extensions/v3/events
* [...]

PR #860, opened 2021-06-15
Implement WordPress API endpoint URL patterns

MizukiTemma:
### Short description

This PR contains the newly implemented URL patterns for WordPress API endpoint.

### Proposed changes

- New URL pattern was implemented.
-
-

### Resolved issues

Fixes: #855

### DIFF
--- a/src/api/urls.py
+++ b/src/api/urls.py
@@ -27,11 +27,7 @@ from .v3.regions import regions, liveregions, hiddenregions, pushnew
 from .v3.offers import offers
 from .v3.single_page import single_page
 
-urlpatterns = [
-    url(r"^regions/?$", regions, name="api_regions"),
-    url(r"^regions/live/?$", liveregions, name="api_regions_live"),
-    url(r"^regions/hidden/?$", hiddenregions, name="api_regions_hidden"),
-    url(r"^regions/pushnew/?$", pushnew, name="api_pushnew"),
+api_urlpatterns = [
     url(
         r"^(?P<region_slug>[-\w]+)/",
         include(
@@ -134,4 +130,19 @@ urlpatterns = [
             ]
         ),
     ),
+]
+
+region_api_urlpatterns = [
+    url(r"^$", regions, name="api_regions"),
+    url(r"^live/?$", liveregions, name="api_regions_live"),
+    url(r"^hidden/?$", hiddenregions, name="api_regions_hidden"),
+    url(r"^pushnew/?$", pushnew, name="api_pushnew"),
+]
+
+
+urlpatterns = [
+    url(r"^api/", include(api_urlpatterns)),
+    url(r"^api/regions/", include(region_api_urlpatterns)),
+    url(r"^wp-json/extensions/v3/", include(api_urlpatterns)),
+    url(r"^wp-json/extensions/v3/sites/", include(region_api_urlpatterns)),
 ]
--- a/src/backend/urls.py
+++ b/src/backend/urls.py
@@ -22,7 +22,7 @@ from django.contrib import admin
 
 
 urlpatterns = [
-    url(r"^api/", include("api.urls")),
+    url(r"^", include("api.urls")),
     url(r"^i18n/", include("django.conf.urls.i18n")),
 ]
 


src/api/urls.py
```diff
@@ -135,3 +135,7 @@
         ),
     ),
 ]
+
+apiurl = [
+    url(r"^/wp-json/extensions/v3/$", include(urlpatterns)),
+]
```
melegiul, line 141: According to the docs [(URLconf)](https://docs.djangoproject.com/en/3.1/topics/http/urls/#how-django-processes-a-request) django will load the url config and look for a variable called `urlpatterns`. So one option would be to append this list to `urlpatterns` sequence. But then there is the problem, that in our ROOT_URLCONF [API-urls](https://github.com/Integreat/integreat-cms/blob/e3b7c733822b3be05a5202b9b2e1a6c605e4e681/src/backend/urls.py#L25) all urls in this file get a preceeding `api` and this violates wordpress API specification.
Do you have an idea how to solve this problem?

melegiul (changes requested):
Thank you very much for your PR :+1: 
Your idea is basically right to support the wordpress API
But there is a remaining problem that we have to address

src/api/urls.py
timoludwig: @melegiul Good catch! I suggest to move the "api" prefix to the [src/api/urls.py](https://github.com/Integreat/integreat-cms/blob/563c174ac36cde9e922aa2b3a31c245e9e1da93b/src/api/urls.py) file in the following way:

```
api_urlpatterns = [
    # All existing urlpatterns here
]

urlpatterns = [
    url(r"^api/", include(api_urlpatterns)),
    url(r"^wp-json/extensions/v3/", include(api_urlpatterns)),
]
```
Apart from that, there are also a few endoints which are named differently (e.g. "sites" vs "regions"), so for this cases, we'd need a new additional rules.

melegiul:
That's an interesting problem that you encountered here, which is not you're fault
The problem is that our tests use [reverse](https://docs.djangoproject.com/en/3.1/ref/urlresolvers/#reverse) for the urls of [views](https://github.com/Integreat/integreat-cms/blob/develop/src/cms/tests/views/api_view_test.py), but in the docs it says:
```
The reverse() function can reverse a large variety of regular expression patterns for URLs,
but not every possible one. The main restriction at the moment is that the pattern cannot contain
alternative choices using the vertical bar ("|") character.
```
So I could offer two options:
1. Revolve the regex and split it in two entries, one for "regions" and one for "sites", so that reversing in the test can succeed
2. We could open an issue to alter the test case setup to support vertical bar regex. In this case please comment out the critical test cases: "api_regions", "api_regions_live", "api_regions_hidden" in the file [https://github.com/Integreat/integreat-cms/blob/develop/src/cms/tests/views/api_view_test.py](https://github.com/Integreat/integreat-cms/blob/develop/src/cms/tests/views/api_view_test.py) and add a comment: 
```
# TODO: find way to test API without using reverse
```
I have no preference between these options, in my opinion you can choose which one you like

melegiul:
@timoludwig What do you think, is there another option to solve this problem?

timoludwig:
> @timoludwig What do you think, is there another option to solve this problem?

In this case, I think I'd suggest to introduce another variable and add the urls with the prefix `regions/` to the normal api endpoint and the legacy `sites/` only to the wordpress-api-patterns:

```
api_urlpatterns = [
    # All existing urlpatterns without the region-patterns here
]

region_api_urlpatterns = [
    url(r"^$", regions, name="api_regions"),
    url(r"^live/?$", liveregions, name="api_regions_live"),
    url(r"^hidden/?$", hiddenregions, name="api_regions_hidden"),
    url(r"^pushnew/?$", pushnew, name="api_pushnew"),
]

urlpatterns = [
    url(r"^api/", include(api_urlpatterns)),
    url(r"^api/regions/", include(region_api_urlpatterns)),
    url(r"^wp-json/extensions/v3/", include(api_urlpatterns)),
    url(r"^wp-json/extensions/v3/sites/", include(region_api_urlpatterns)),
]
```

timoludwig (approved):
Thanks! :+1:
I think this looks good and works as expected! :tada:

I combined all your commits into one again.

melegiul (approved):
Thank you :+1:

Issue #861, opened 2021-06-15
Event forms location field broken


melegiul:
### Describe the Bug

When entering a location for an event, the cms throws a ProgrammingError exception.

### Steps to Reproduce

1. Go to the event form
2. Enter some characters to the location field
3. Watch the consoles output

### Expected Behavior

When I understand the help text right, existing location should be looked up for selection.

### Actual Behavior

Exception thrown

### Additional Information

```
File "/home/giuliano/dev/integreat-cms/.venv/lib/python3.7/site-packages/django/db/utils.py", line 90, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/home/giuliano/dev/integreat-cms/.venv/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
django.db.utils.ProgrammingError: SELECT DISTINCT ON expressions must match initial ORDER BY expressions
LINE 1: ..."cms_poitranslation"."id" IN (SELECT DISTINCT ON (U0."langua...
                                                             ^

Jun 15 09:44:42 ERROR django.server - "POST /ajax/search_poi HTTP/1.1" 500 224363
```

PR #862, opened 2021-06-15
Fix broken poi field in event form

melegiul:
### Short description

Fix broken poi field in event form

### Proposed changes

According to the docs:
```
Keep in mind that order_by() uses any default related model ordering that has been defined. 
You might have to explicitly order by the relation _id or referenced field to make sure 
the DISTINCT ON expressions match those at the beginning of the ORDER BY clause. 
``` 
Since the language model is ordered by bcp47_tag, docs suggest to explicitly order by the relations id. 

### Resolved issues

Fixes: #861


timoludwig (approved):
Thanks, this fixes the problem! :+1:

Issue #804, opened 2021-05-04
Generalize slugify ajax view in content forms


timoludwig:
### Motivation

When changing a page's title in the page form, the value is automatically slugified and inserted in the slug field. When clicking on an editing button, the value can be changed but it will also check whether that slug is already existent in the database and appends a number until it is unique (see [slugify_ajax()](https://github.com/Integreat/integreat-cms/blob/6a2099336bcee9f70d5b682bcbecec23edcfd27d/src/cms/views/pages/page_actions.py#L902) and [generate_unique_slug()](https://github.com/Integreat/integreat-cms/blob/6a2099336bcee9f70d5b682bcbecec23edcfd27d/src/cms/utils/slug_utils.py#L52)).

![Screenshot_2021-05-04 Integreat Redaktionssystem(2)](https://user-images.githubusercontent.com/16021269/116982540-330fa780-acc9-11eb-9460-69b8a95e4872.png)
![Screenshot_2021-05-04 Integreat Redaktionssystem(1)](https://user-images.githubusercontent.com/16021269/116982547-360a9800-acc9-11eb-8085-9b4460710ae2.png)

The same UI also exists for POIs, but the checking mechanism for slugs still only works for pages.

### Proposed Solution

- Generalize the slugify AJAX view so it works for pages, events and POIs
- Adapt the UI of pages and POIs also to events
- Make the look and behaviour consistent on all three content forms

### Additional Context

https://github.com/Integreat/integreat-cms/blob/6a2099336bcee9f70d5b682bcbecec23edcfd27d/src/cms/views/pages/page_actions.py#L902-L932

PR #863, opened 2021-06-18
generalize slugify_ajax

david-venhoff:
### Short description
This pr makes the design of the slug input field of page form, event form and poi form more inline with each other.
Additionally, the methods `slugify_ajax` can now work for pois and events too, while it previously just worked for pages.

### Proposed changes
- Update the slug input field of events
- Make the design of the slug input field uniform across pages, events and pois
- Make the `slugify_ajax` function more generic in order to allow it to also function with events and pois

### Resolved issues
Fixes: #804

### DIFF
--- a/src/cms/templates/events/event_form.html
+++ b/src/cms/templates/events/event_form.html
@@ -120,20 +120,33 @@
                         <span class="inline-block mb-2 font-bold">{% trans 'Status' %}:</span>
                         {{ event_translation_form.instance.get_status_display }}
                     {% endif %}
-                    <label for="{{ event_translation_form.title.id_for_label }}" class="block mb-2 mt-4 font-bold">{{ event_translation_form.title.label }}</label>
+                    <label for="{{ event_translation_form.title.id_for_label }}" 
+                        data-slugify-url="{% url 'slugify_ajax' region_slug=region.slug language_slug=language.slug model_type='event' %}{% if event_form.instance.id %}?model_id={{ event_form.instance.id }}{% endif %}"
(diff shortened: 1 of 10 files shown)


src/cms/templates/events/event_form.html
```diff
@@ -120,20 +120,33 @@ <h1 class="heading">
...
-                        {% spaceless %}
-                            <div style="display: table; white-space: nowrap;">
-                                <span style="display: table-cell;">https://integreat.app/{{ region.slug }}/{{ language.slug }}/events/</span>
-                                {% if event_translation_form.instance.ancestor_path %}
-                                    <span style="display: table-cell;">{{ event_translation_form.instance.ancestor_path }}/</span>
-                                {% endif %}
-                                <span style="display: table-cell; width: 100%;">{% render_field event_translation_form.slug|add_error_class:"slug-error" class="w-full rounded" %}</span>
-                            </div>
-                        {% endspaceless %}
+                    <div id="link-container" class="flex items-center">
```
melegiul, line 127: Could you please just add a little margin between the two form fields for events and pois?
Depending on the browser there is no/little gap between currently.
On chrome:
Screenshot from 2021-06-21 18-06-54
On firefox:
Screenshot from 2021-06-21 18-07-50

melegiul (changes requested):
Very clean and concise solution, this works as expected :tada:
I could just find a minor problem.

src/cms/templates/events/event_form.html
david-venhoff: Sure! I should have tested this with chrome, too.

melegiul (approved):
Thank you very much
Looks good to me :+1:

Issue #864, opened 2021-06-19
Crash when trying to delete a region


david-venhoff:
### Describe the Bug
Deleting a region via this page: 
![grafik](https://user-images.githubusercontent.com/78504586/122641895-558a4280-d108-11eb-93db-2bd36f8b3ce2.png)
fails with a `FieldError`:
![grafik](https://user-images.githubusercontent.com/78504586/122641888-42777280-d108-11eb-9344-86ecfb40b78a.png)

### Steps to Reproduce

1. Go to the regions tab
2. Edit any region
3. Scroll down and click on delete. Click ok on the warning.
4. See the error

### Expected Behavior
The server does not fail and deletes the region

### Actual Behavior
The server crashes and does not delete the region

### Additional Information
/

PR #865, opened 2021-06-21
Fix region deletion

timoludwig:
### Short description

Fix the region deletion, the solution was similar to #862.

### Proposed changes

- Fix ordering of language tree nodes
- Remove mapping of event location to prevent `ProtectedError`

### Resolved issues

Fixes: #864


melegiul (approved):
Thank you very much :+1:
Looks good to me

PR #872, opened 2021-06-29
Change translation title length

svenseeberg:
As UTF-8 characters can take up to 4 bytes, a limit of 250 bytes is very limiting for some languages. Migrating existing titles requires about 1024 bytes in length.


timoludwig (approved):
Looks good! :+1:

timoludwig:
@svenseeberg I increased the lengths of the other content models as well and also extended the slug fields (which are often times directly derived from the titles). If you're fine with these changes, we can merge this PR.

Issue #877, opened 2021-06-29
Menu links not working anymore when media library is opened


timoludwig:
### Describe the Bug

When the media library is opened, all other links outside the media library are not working anymore and are just interpreted as directory ids for the media library.
E.g., when clicking on the link http://localhost:8000/admin_dashboard/, the preact router catches this URL and actually opens http://localhost:8000/media-library/#/admin_dashboard/, which causes the server error `ValueError: Field 'id' expected a number but got 'admin_dashboard'` and the in the frontend, the error `An error has occurred. Please try again later.` appears.

### Steps to Reproduce

1. Go to the media library
2. Click on any link outside the media library

### Expected Behavior

Normal links should still work

### Actual Behavior

All links are treated as internal media library links of the preact router

ulliholtgrave:
This seems to be only a problem, when we are in the admin view of the library. The Preact component seems to call the directory content API everytime a link is clicked.

PR #879, opened 2021-06-30
Use preact router only for specific links

timoludwig:
### Short description

Use preact router only for specific links

### Proposed changes

Instead of going opt-out on preact routing links by adding the attribute `native` to all non-preact-links, I made preact routing opt-in by adding the attribute `media-library-link`.
All links without the `media-library-link` attribute now get the `native` attribute by default.
This means that only directory links are opened in preact, and all other links work as expected.

### Resolved issues

Fixes: #877


ulliholtgrave (approved):
Fixes the bug. Thanks :)

Issue #801, opened 2021-05-02
Load RSS feed dynamically


timoludwig:
### Motivation

When the blog is unreachable or takes very long to respond, the region dashboard cannot be opened.

### Proposed Solution

Load the RSS feed dynamically via ajax without interfering with the region dashboard view.
This mechanism can also be used to load more items dynamically if the user wants to read more without leaving the dasboard.

### Alternatives

- [x] At least add a timeout and error handling to the [RSS feed request](https://github.com/Integreat/integreat-cms/blob/0980373b98e4f616614e6e8f90e04b4c98218bd6/src/cms/views/dashboard/dashboard_view.py#L46) in the dashboard view. - **already implemented**

PR #881, opened 2021-07-03
Minimal solution for RSS timeout exception

membralala:
### Short description
Implements a minimal solution to deal with the timeout-exception in the dashboard RSS feed. Does not yet implement the proposed dynamic solution via ajax. 

### Proposed changes
- added Timeout exception 
- added debug information
- adjusted template

### Resolved issues
#801 

### Additional Information
Handled the exception via the requests module as purposed by the author of the feedparser package: https://github.com/kurtmckee/feedparser/pull/80
For testing purposes, set the timeout-argument to a very small value and load the region dashboard.


src/cms/views/dashboard/dashboard_view.py
```diff
@@ -67,18 +68,33 @@ def get(self, request, *args, **kwargs):
...
-        feed["entries"] = feed["entries"][:3]
-        # decode html entities like dash and split after line break
-        for entry in feed["entries"]:
-            entry["summary"] = html.unescape(entry["summary"]).split("\n")[0]
+        has_timed_out = False
+        feed = []
+        rss_feed_url = RSS_FEED_URLS[language_slug]
+
+        try:
+            requests.get(rss_feed_url, timeout=5.0)
```
timoludwig, line 76: This will do one additional request just to check whether the feed is available... This will make the process slower each time the feed is actually available (which is 99.9% of the time), so I don't think it's a good idea. It would be better to actually use the response of the first request to directly parse the RSS feed:
```suggestion
            response = requests.get(rss_feed_url, timeout=5)
            feed = feedparser.parse(response.text)
```

src/cms/views/dashboard/dashboard_view.py
```diff
@@ -67,18 +68,33 @@ def get(self, request, *args, **kwargs):
...
+            )
+            has_timed_out = True
+
+        if not has_timed_out:
+            feed = feedparser.parse(rss_feed_url)
+            # select five most recent feeds
+            feed["entries"] = feed["entries"][:5]
+            # decode html entities like dash and split after line break
+            for entry in feed["entries"]:
+                entry["summary"] = html.unescape(entry["summary"]).split("\n")[0]
```
timoludwig, line 89: Normally, in a try-catch block, you just continue the execution in the try-block as long as you want to rely on the "tried" command:
```suggestion
            # select five most recent feeds
            feed["entries"] = feed["entries"][:5]
            # decode html entities like dash and split after line break
            for entry in feed["entries"]:
                entry["summary"] = html.unescape(entry["summary"]).split("\n")[0]
        except requests.ReadTimeout:
            logger.debug(
                "Rss-feed timed out. Continue loading site without news column."
            )
            has_timed_out = True
```

timoludwig (commented):
Thanks! :+1:
In general, this works. I just have a small idea for improvement.
After this is merged, the priority of issue #801 will be lower, since dynamic loading via AJAX would still be nice, but is not explicitly required for our release.

src/cms/views/dashboard/dashboard_view.py
membralala: Thanks for this hint!

timoludwig (approved):
Thanks! :+1:

Issue #590, opened 2020-11-18
Dynamic internal link target suggestions


timoludwig:
### Motivation

When adding internal links, we should dynamically suggest targets while typing.
Searching for page/event/poi titles should be possible as well slugs.

### Proposed Solution

Perform the dynamic search via AJAX. Also have a look at similar implementations, e.g. the dynamic POI suggestions in the event form.

david-venhoff:
Should the event handler that makes the ajax calls be active in the editor or just in the `insert/link` prompt? If the event handler would just run in the `insert/link` prompt, it might make sense to just preload the available links, maybe roughly like that:
![grafik](https://user-images.githubusercontent.com/78504586/123636215-e1911e00-d81c-11eb-8864-1ff7fcb7dfb7.png)

What do you think about this approach?

timoludwig:
I think in this case, we have to make the behavior somewhat similar to wordpress, otherwise the municipalities might be confused after the migration. This means, the suggestions need to be active in both cases, when marking text in the editor and when inserting a completely new link. Also, the suggestions should adapt to the text the user is typing, not only listing all available options. I'm not sure how it works in wordpress, it seems like not only the titles, but also the content of the targets are searched and suggested. For reference, here the inline suggestion of wordpress:
![Screenshot 2021-06-28 at 14-46-17 Seite bearbeiten ‹ Nürnberg — WordPress](https://user-images.githubusercontent.com/16021269/123641045-1b185800-d822-11eb-8e9d-3da6a333e1e4.png)
![Screenshot 2021-06-28 at 14-48-19 Seite bearbeiten ‹ Nürnberg — WordPress](https://user-images.githubusercontent.com/16021269/123641053-1e134880-d822-11eb-8a36-0172d1eb7de7.png)
And here the design of the "insert link"-promt:
![Screenshot 2021-06-28 at 14-49-29 Seite bearbeiten ‹ Nürnberg — WordPress](https://user-images.githubusercontent.com/16021269/123641370-7cd8c200-d822-11eb-85f5-4f02a60b7843.png)
Does that help? Do you think that's doable? The design doesn't have to be 100% identical, but the workflow of inserting internal links should roughly stay the same.
Since at the moment, we just have the prompt, even if inline text is selected for a link, I think it's ok to start with that and only implement the small inline suggestion box if needed.

david-venhoff:
[...]

PR #890, opened 2021-07-11
Add internal link suggestions for tinymce editor

david-venhoff:
### Short description
This pr adds support for searching internal links for pois, events and pages in the tinymce editor.
image
image

### Proposed changes
- Replace the link plugin with a custom implementation of similar functionality
- Add an ajax endpoint for querying pois, events and pages for the current region that match a query string

### Resolved issues
Fixes: #590

### DIFF
--- a/src/cms/views/utils/search_content_ajax.py
+++ b/src/cms/views/utils/search_content_ajax.py
@@ -1,0 +1,126 @@
+import logging
+import json
+
+from django.contrib.auth.decorators import login_required
+from django.db.models import Q
+from django.http import JsonResponse
+from django.views.decorators.http import require_POST
+
+from backend.settings import WEBAPP_URL
+from ...constants import status
+from ...decorators import region_permission_required
+from ...models import Region, EventTranslation, PageTranslation, POITranslation
+
+logger = logging.getLogger(__name__)
+
+# The maximum number of results returned by `search_content_ajax`
+MAX_RESULT_COUNT = 20
+
+
+def format_object_translation(object_translation, typ):
+    """
+    Formats the [poi/event/page]-translation as json
+
+    :param object_translation: A translation object which has a title and a permalink
+    :type object_translation: ~cms.models.events.event.Event or ~cms.models.pages.page.Page or ~cms.models.pois.poi.POI
+    :param typ: The type of this object
+    :type typ: str
+    :return: A dictionary with the title, url and type of the translation object
+    :rtype: dict
+    """
+    return {
+        "title": object_translation.title,
+        "url": f"{WEBAPP_URL}/{object_translation.permalink}",
+        "type": typ,
+    }
+
+
+def find_objects(manager, obj_type, region, language_slug, query, hierarchical=False):
+    """
+    Filters all object translations from ``manager`` of this language that match the ``query``
+
+    :param manager: The manager for a specific model
+    :type manager: ~django.db.models.Manager
+    :param obj_type: The content type of the queried translations
+    :type obj_type: str
+    :param region: The current region
+    :type region: ~cms.models.regions.region.Region
+    :param language_slug: The language slug
+    :type language_slug: str
+    :param query: The query string used for filtering the objects
+    :type query: str
+    :param hierarchical: Whether or not the given model is hiearchical (which impacts the method for filtering out
+                         archived results)
+    :type hierarchical: bool
+    :return: A list of translation objects (formatted as dict)
+    :rtype: list [ dict ]
+    """
+    archived_flag = "explicitly_archived" if hierarchical else "archived"
+    result = (
+        manager.objects.filter(
+            **{
+                obj_type + "__region": region,
+                obj_type + "__" + archived_flag: False,
+                "language__slug": language_slug,
+                "status": status.PUBLIC,
+            }
+        )
+        .filter(Q(slug__icontains=query) | Q(title__icontains=query))
+        .distinct(obj_type)
+    )
+    # If the object type is hierarchical, filter out implicitly archived objects (objects with archived ancestors)
+    if hierarchical:
+        result = filter(lambda x: not getattr(x, obj_type).implicitly_archived, result)
+    return [format_object_translation(obj, obj_type) for obj in result]
+
+
+@require_POST
+@login_required
+@region_permission_required
+# pylint: disable=unused-argument
+def search_content_ajax(request, region_slug, language_slug):
+    """Searches all pois, events and pages for the current region and returns all that
+    match the search query. Results which match the query in the title or slug get ranked
+    higher than results which only match through their text content.
+
+    :param request: The current request
+    :type request: ~django.http.HttpResponse
+    :param region_slug: region identifier
+    :type region_slug: str
+    :param language_slug: language slug
+    :type language_slug: str
+    :return: Json object containing all matching elements, of shape {title: str, url: str, type: str}
+    :rtype: ~django.http.JsonResponse
+    """
+
+    region = Region.get_current_region(request)
+    query = json.loads(request.body.decode("utf-8"))["query_string"]
+
+    logger.debug(
+        "Ajax call: Live search for pois, events and pages with query %r", query
+    )
+
+    results = []
+
+    user = request.user
+    if user.has_perm("cms.view_event"):
+        results.extend(
+            find_objects(EventTranslation, "event", region, language_slug, query)
+        )
+
+    if user.has_perm("cms.view_page"):
+        results.extend(
+            find_objects(
+                PageTranslation, "page", region, language_slug, query, hierarchical=True
+            )
+        )
+
+    if user.has_perm("cms.view_poi"):
+        results.extend(
+            find_objects(POITranslation, "poi", region, language_slug, query)
+        )
+
+    # sort alphabetically by title
+    results.sort(key=lambda k: k["title"])
+
+    return JsonResponse({"data": results[:MAX_RESULT_COUNT]})
--- a/src/frontend/js/tinymce-plugins/custom_link_input/plugin.js
+++ b/src/frontend/js/tinymce-plugins/custom_link_input/plugin.js
@@ -1,0 +1,320 @@
+import { getCsrfToken } from "../../utils/csrf-token";
+
+(function () {
+    'use strict'
+
+    const tinymceConfig = document.getElementById("tinymce-config-options");
+
+    async function getCompletions(query, id) {
+        const url = tinymceConfig.getAttribute("data-link-ajax-url");
+        const response = await fetch(url, {
+            method: "POST",
+            headers: {
+                "X-CSRFToken": getCsrfToken(),
+            },
+            body: JSON.stringify({
+                query_string: query
+            })
+        });
+        if (response.status != 200) {
+            return [];
+        }
+
+        const data = await response.json();
+        return [data.data, id];
+    }
+
+    // Checks if the url is likely missing the https:// prefix and add it if that is the case
+    function checkUrlHttps(url) {
+        // This regex matches domains without protocol (strings which contain a dot without a preceding colon or slash)
+        const re = new RegExp("^[^:/]+[.].+");
+        if (re.test(url)) {
+            return "https://" + url;
+        }
+        return url;
+    }
+
+    function updateLink(editor, anchorElm, text, linkAttrs) {
+        anchorElm.textContent = text;
+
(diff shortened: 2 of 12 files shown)


src/cms/views/utils/search_content_ajax.py
```diff
@@ -0,0 +1,103 @@
...
+            "url": f"{WEBAPP_URL}/{object_translation.permalink}",
+            "type": typ,
+        }
+
+    def find_objects(object_manager, query, language_slug, body_name):
+        """Filters all object translations from ``object_manager`` of this language
+        that match ``query`` and returns them and their ranking"""
+        for obj in object_manager.all():
+            if obj.archived:
+                continue
```
timoludwig, line 51: I think it might be a bit inefficient to query all objects first, then check whether they're archived, then do a regex matching on the translation fields...
Maybe we can do the search on the SQL level with the Django QuerySet features (see [docu](https://docs.djangoproject.com/en/3.2/topics/db/search/))? I'm imagining something like:
```
    Q(page__region=region, language=language) & 
    (Q(slug__unaccent__icontains=query) | Q(title__unaccent__icontains=query))
).distinct("page")
```
But that's probably hard to get right and generalize for all content types (and we have the problem with the priorities, but if it's too complicated I'd rather remove the body-search). I will think about this and get back to you in a few days.

src/cms/views/utils/search_content_ajax.py
```diff
@@ -0,0 +1,103 @@
...
+        'Ajax call: Live search for pois, events and pages with query "%r"', raw_query
+    )
+
+    results = []
+
+    user = request.user
+    if user.has_perm("cms.view_event"):
+        results.extend(
+            (prio, format_object_translation(i, "event"))
+            for (prio, i) in find_objects(
```
timoludwig, line 81: I always associate `i` with some kind of integer counter, so in this case I'd rather use `obj` or similar...

src/frontend/js/tinymce-plugins/custom_link_input/plugin.js
```diff
@@ -0,0 +1,291 @@
...
+                            items: [
+                                {
+                                    type: "input",
+                                    name: "search",
+                                },
+                                {
+                                    type: "selectbox",
+                                    name: "completions",
+                                    items: completion_items
+                                }
```
timoludwig, line 164: Do you think you could add a default (but disabled) option for the select field like " - no results - " or similar when there's nothing to select? Only if it's not too complicated :sweat_smile:

timoludwig (commented):
Amazing, thanks! :heart_eyes: 
I really like the solution and I think the municipalities will be able to use this without any problems when they're used to the WordPress link plugin.
Kudos, I always get a headache when dealing with these tinymce plugins :joy:

src/frontend/js/tinymce-plugins/custom_link_input/plugin.js
```diff
@@ -0,0 +1,291 @@
...
+    // Ask if the user wants to append https if that prefix is likely missing
+    function checkUrlHttps(editor, url, handler) {
+        if (url.toLowerCase().startsWith("www.")) {
+            editor.windowManager.confirm(tinymceConfig.getAttribute("data-link-dialog-confirm_https-text"), function (s) {
+                if (s)
+                    handler("https://" + url);
+                else {
+                    handler(url);
+                }
+            });
```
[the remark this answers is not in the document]
david-venhoff: It seems like the static assets loaded for the tinymce editor contain relative paths: https://github.com/Integreat/integreat-cms/blob/e1af620e72d586e12379282c0b01f5870f7a308b/src/cms/templates/_tinymce_config.html#L33

Apparently, tinymce also automatically prefixes those with the `document_base_url` setting, which leads to invalid urls:
grafik

I am not quite sure how to handle this in the best way. Should i prefix these assets with the actual url?

src/cms/views/utils/search_content_ajax.py
david-venhoff: Okay. I have removed the body search for now.

src/frontend/js/tinymce-plugins/custom_link_input/plugin.js
```diff
@@ -0,0 +1,291 @@
...
+                        // restore the original user url
+                        api.setData({ url: user_url });
+                    }
+                }
+                prev_selected_completion = data.completions;
+
+                // Automatically update the text input to the url by default
+                data = api.getData();
+                if (data.url != data.text && prev_link_url == data.text) {
+                    api.setData({ text: data.url });
```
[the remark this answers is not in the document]
david-venhoff: I implemented it, but with a restriction that the link text does not get overwritten if it is not empty. I think otherwise this behavior could be annoying for the users, if they already entered a link text and then started to search for an internal link, since in this case their link text input would get overwritten.

src/frontend/js/tinymce-plugins/custom_link_input/plugin.js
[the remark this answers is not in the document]
timoludwig: Hmm, that's sad. But yes, I think it's ok to prefix the static paths of the assets with the cms host, you can e.g. get this from the setting `BASE_URL`... Probably that's less effort than to deal with all the problems which arise when tinymce builds the absolute urls...

src/cms/views/utils/search_content_ajax.py
```diff
@@ -0,0 +1,118 @@
...
+        if not object_translation:
+            continue
+
+        title = object_translation.title
+        slug = object_translation.slug
+
+        if re.search(query, title, re.IGNORECASE):
+            yield (2, object_translation)
+        elif re.search(query, slug, re.IGNORECASE):
+            yield (2, object_translation)
```
timoludwig, line 64:
```suggestion
def find_objects(manager, obj_type, region, language_slug, query, hierarchical=False):
    """
    Filters all object translations from ``manager`` of this language that match the ``query``

    :param manager: The manager for a specific model
    :type manager: ~django.db.models.Manager

    :param obj_type: The content type of the queried translations
    :type obj_type: str

    :param region: The current region
    :type region: ~cms.models.regions.region.Region

    :param language_slug: The language slug
    :type language_slug: str

    :param query: The query string used for filtering the objects
    :type query: str

    :param hierarchical: Whether or not the given model is hiearchical (which impacts the method for filtering out
                         archived results)
    :type hierarchical: bool

    :return: A list of translation objects (formatted as dict)
    :rtype: list [ dict ]
    """
    archived_flag = "explicitly_archived" if hierarchical else "archived"
    result = (
        manager.objects.filter(
            **{
                obj_type + "__region": region,
                obj_type + "__" + archived_flag: False,
                "language__slug": language_slug,
                "status": status.PUBLIC,
            }
        )
        .filter(Q(slug__icontains=query) | Q(title__icontains=query))
        .distinct(obj_type)
    )
    # If the object type is hierarchical, filter out implicitly archived objects (objects with archived ancestors)
    if hierarchical:
        result = filter(lambda x: not getattr(x, obj_type).implicitly_archived, result)
    return [format_object_translation(obj, obj_type) for obj in result]
```
Ok, so this would be my approach to do the search on the database level instead of doing a regex search in python.
Especially with many entries in the database, this should give a significant performance improvement, although I didn't do any benchmarking on this.

src/cms/views/utils/search_content_ajax.py
```diff
@@ -0,0 +1,118 @@
...
+    :type region_slug: str
+    :param language_slug: language slug
+    :type language_slug: str
+    :return: Json object containing all matching elements, of shape {title: str, url: str, type: str}
+    :rtype: ~django.http.JsonResponse
+    """
+
+    region = Region.get_current_region(request)
+    raw_query = json.loads(request.body.decode("utf-8"))["query_string"]
+    query = re.escape(raw_query)
```
timoludwig, line 88:
```suggestion
    query = json.loads(request.body.decode("utf-8"))["query_string"]
```

src/cms/views/utils/search_content_ajax.py
```diff
@@ -0,0 +1,118 @@
...
+
+    logger.debug(
+        'Ajax call: Live search for pois, events and pages with query "%r"', raw_query
```
timoludwig, line 91:
```suggestion
        "Ajax call: Live search for pois, events and pages with query %r", query
```
The representation modifier `%r` already puts quotes around string variabels.

src/cms/views/utils/search_content_ajax.py
```diff
@@ -0,0 +1,118 @@
...
+    )
+
+    results = []
+
+    user = request.user
+    if user.has_perm("cms.view_event"):
+        results.extend(
+            (prio, format_object_translation(obj, "event"))
+            for (prio, obj) in find_objects(region.events, query, language_slug)
```
timoludwig, line 100:
```suggestion
            find_objects(EventTranslation, "event", region, language_slug, query)
```

src/cms/views/utils/search_content_ajax.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+
+    if user.has_perm("cms.view_page"):
+        results.extend(
+            (prio, format_object_translation(obj, "page"))
+            for (prio, obj) in find_objects(region.pages, query, language_slug)
```
timoludwig, line 106:
```suggestion
            find_objects(PageTranslation, "page", region, language_slug, query, True)
```

src/cms/views/utils/search_content_ajax.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+
+    if user.has_perm("cms.view_poi"):
+        results.extend(
+            (prio, format_object_translation(obj, "poi"))
+            for (prio, obj) in find_objects(region.pois, query, language_slug)
```
timoludwig, line 112:
```suggestion
            find_objects(POITranslation, "poi", region, language_slug, query)
```

src/cms/views/utils/search_content_ajax.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+
+    # sort first by highest priority and then alphabetically by title
+    results.sort(key=lambda res: (-res[0], res[1]["title"]))
```
timoludwig, line 116:
```suggestion
    # sort alphabetically by title
    results.sort(key=lambda k: k["title"])
```

src/cms/views/utils/search_content_ajax.py
```diff
@@ -0,0 +1,118 @@
...
+
+    return JsonResponse({"data": [i[1] for i in results[:MAX_RESULT_COUNT]]})
```
timoludwig, line 118:
```suggestion
    return JsonResponse({"data": results[:MAX_RESULT_COUNT]})
```

src/cms/views/utils/search_content_ajax.py
```diff
@@ -0,0 +1,118 @@
+import re
```
timoludwig, line 1:
```suggestion
```

src/cms/views/utils/search_content_ajax.py
```diff
@@ -0,0 +1,118 @@
...
+import logging
+import json
+
+from django.contrib.auth.decorators import login_required
+from django.http import JsonResponse
```
timoludwig, line 6:
```suggestion
from django.db.models import Q
from django.http import JsonResponse
```

src/cms/views/utils/search_content_ajax.py
```diff
@@ -0,0 +1,118 @@
...
+from django.views.decorators.http import require_POST
+
+from backend.settings import WEBAPP_URL
+from ...decorators import region_permission_required
+from ...models.regions.region import Region
```
timoludwig, line 11:
```suggestion
from ...constants import status
from ...decorators import region_permission_required
from ...models import Region, PageTranslation, EventTranslation, POITranslation
```

timoludwig (commented):
Amazing, thanks for your changes! :rocket: 

I made suggestions for doing the filtering in sql instead of Python. I'ts a bit complicated because of the different archiving concept of the hierarchcal page model, so If you see any problems with my code or further room for improvement, feel free to change it.
I noticed that I needed to change the default sorting of the event translation model to `ordering = ["event__pk", "-version"]` in [src/cms/models/events/event_translation.py](https://github.com/Integreat/integreat-cms/blob/8219586acbb98c681635210d0c45817285a0c13e/src/cms/models/events/event_translation.py#L313) in order for the sql query to work.

timoludwig (approved):
Thank you! :1st_place_medal: 
I think that's it, you can merge the PR :tada:

Issue #889, opened 2021-07-07
Fix `Organization` logo field


timoludwig:
### Describe the Bug

When implementing the new media library and adapting the icon and logo fields to it, we forgot to adapt the organization form to the new system.

### Steps to Reproduce

1. Go to 'Organizations'
2. Create Organization
3. Try to set logo

### Expected Behavior

The media library overlay should be shown so a logo can be selected from the media library

### Actual Behavior

Nothing happens

### Additional Information

Have a look at e.g. the solution of the region logo and re-use or copy/paste the existing and working code

PR #891, opened 2021-07-14
Fix logo issue in organizations

membralala:
### Short description
Fixed a bug, where the organization form wasn't updated yet to work properly with the new media library. 

### Proposed changes
- updated organization_view
- changed the organization model's icon entry 
- added the logo as a thumbnail to the organization list 

### Resolved issues

Fixes: #889

### DIFF
--- a/src/cms/models/users/organization.py
+++ b/src/cms/models/users/organization.py
@@ -2,6 +2,8 @@ from django.db import models
 from django.utils import timezone
 from django.utils.translation import ugettext_lazy as _
 
+from ..media.media_file import MediaFile
+
 
 class Organization(models.Model):
     """
@@ -16,11 +18,14 @@ class Organization(models.Model):
         verbose_name=_("slug"),
         help_text=_("Unique string identifier without spaces and special characters."),
     )
-    icon = models.ImageField(
-        null=True,
-        blank=True,
-        upload_to="organizations/%Y/%m/%d",
+
+    icon = models.ForeignKey(
+        MediaFile,
         verbose_name=_("logo"),
+        on_delete=models.SET_NULL,
+        related_name="icon_organizations",
+        blank=True,
+        null=True,
     )
 
     created_date = models.DateTimeField(
--- a/src/cms/templates/organizations/organization_list_row.html
+++ b/src/cms/templates/organizations/organization_list_row.html
@@ -11,8 +11,8 @@
 	</td>
 	<td class="pr-2">
         {% if organization.icon %}
-            <a href="{{ organization.icon.url }}" target="_blank" rel="noopener noreferrer" class="bg-transparent hover:underline text-blue-500 hover:text-blue-600 cursor-pointer">
-                {{ organization.icon.url }}
+            <a href="{% url 'edit_organization' organization_id=organization.id %}">
+                <img src="{{organization.icon.url}}" alt="{{organization.icon.url}}" style="max-height:48px">
             </a>
         {% else %}
             <i data-feather="x" class="text-red-500"></i>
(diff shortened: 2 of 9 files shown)


src/cms/models/users/organization.py
```diff
@@ -16,11 +16,14 @@ class Organization(models.Model):
...
-    icon = models.ImageField(
-        null=True,
-        blank=True,
-        upload_to="organizations/%Y/%m/%d",
+
+    icon = models.ForeignKey(
+        "cms.MediaFile",
         verbose_name=_("logo"),
+        on_delete=models.SET_NULL,
+        related_name="icon_org",
```
timoludwig, line 24:
```suggestion
        related_name="organization_icon_of",
```
Just to stay consistent with the other models, I'd choose this related name.
I get that's not an ideal name, but if we change the naming scheme, we should probably change it for all models with an icon (regions, organizations, pages, events and POIs). Maybe `organizations_with_this_file_as_icon` would be most understandable, but sure that's a bit verbose :sweat_smile: 
Maybe, for the sake of simplicity, just `organizations`? if you go with this approach, please also rename the related name in all other models as well. The exact name is less relevant than the consistency :smile:

timoludwig (commented):
Thanks! This works exactly as expected! :+1:

src/cms/models/users/organization.py
membralala: Okay, yeah, I get this one! I might go with ```modelname_icon```, if that's fine.

There is another small thing, which came to my mind: In the organization_list_view the logo is represented as a link to the image source. Would it make sense to either represent the actual logo as an image instead of the link or just not show it as a table property?

timoludwig: > Okay, yeah, I get this one! I might go with `modelname_icon`, if that's fine.

I think this sounds as if the attribute would return an icon instance instead of the model. What about `icon_modelname_plural`, so e.g. `icon_organizations` (similar to what you initially suggested)?

> There is another small thing, which came to my mind: In the organization_list_view the logo is represented as a link to the image source. Would it make sense to either represent the actual logo as an image instead of the link or just not show it as a table property?

Yes, I really like this idea. A small thumbnail image would be cool! :heart_eyes_cat:

membralala: > I think this sounds as if the attribute would return an icon instance instead of the model. What about `icon_modelname_plural`, so e.g. `icon_organizations` (similar to what you initially suggested)?

That's fine to me as well, I'll go with it!

src/cms/templates/organizations/organization_list_row.html
```diff
@@ -11,9 +11,9 @@
 	</td>
 	<td class="pr-2">
         {% if organization.icon %}
-            <a href="{{ organization.icon.url }}" target="_blank" rel="noopener noreferrer" class="bg-transparent hover:underline text-blue-500 hover:text-blue-600 cursor-pointer">
-                {{ organization.icon.url }}
-            </a>
+        <a href="{% url 'edit_organization' organization_id=organization.id %}">
+            <img src="{{organization.icon.url}}" alt="{{organization.icon.url}}" style="height:60px">
```
timoludwig, line 15:
```suggestion
            <img src="{{organization.icon.url}}" alt="{{organization.icon.url}}" style="max-height:48px">
```
I'd go with this max height, because the normal text cells also have this height (24px for the content and 12px for top and bottom padding)... this way, all rows are of equal height, no matter whether an icon is set or not.

timoludwig (commented):
Amazing, thanks! The preview looks cool!

timoludwig (approved):
:+1:

Issue #893, opened 2021-07-14
Provide default order for region model


ulliholtgrave:
### Motivation
Currently the region and user lists does not have a default order when they get queried. This results in lists with entries that are hard to find. I suggest that we provide a default alphabetical order by the name.

![grafik](https://user-images.githubusercontent.com/15608678/125632813-53ed3865-c4f1-4dba-bab0-c831515261d8.png)

PR #894, opened 2021-07-15
Provide default order for region model

MizukiTemma:
### Short description

This PR contains a change to show regions alphabetically sorted by dafault.

### Proposed changes

- Regions are now sorted by default when they are shown.
- Userlist was already sorted in region user list and network management.
-

### Resolved issues

Fixes: #893

### DIFF
--- a/src/cms/views/regions/region_list_view.py
+++ b/src/cms/views/regions/region_list_view.py
@@ -44,7 +44,7 @@ class RegionListView(TemplateView):
         """
         regions = Region.objects.all()
         # for consistent pagination querysets should be ordered
-        paginator = Paginator(regions.order_by("created_date"), PER_PAGE)
+        paginator = Paginator(regions, PER_PAGE)
         chunk = request.GET.get("page")
         region_chunk = paginator.get_page(chunk)
         return render(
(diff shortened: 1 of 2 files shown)


src/cms/views/regions/region_list_view.py
```diff
@@ -44,7 +44,7 @@ def get(self, request, *args, **kwargs):
         """
         regions = Region.objects.all()
         # for consistent pagination querysets should be ordered
-        paginator = Paginator(regions.order_by("created_date"), PER_PAGE)
+        paginator = Paginator(regions.order_by("name"), PER_PAGE)
```
timoludwig, line 47: Please apply the sorting directly in the [Meta class](https://docs.djangoproject.com/en/3.2/ref/models/options/#django.db.models.Options.ordering) of the region model, not in the region list view. This way, all region querysets will be sorted, no matter where they're queried.

timoludwig:
```suggestion
        paginator = Paginator(regions, PER_PAGE)
```
Now you can remove the ordering in the paginator, because it's already sorted by default.

timoludwig (commented):
Thanks! :+1:
For regions, it looks good now, but for users, we need to do the ordering in the paginators of the list views just like you suggested because we don't user a custom user model yet.

MizukiTemma:
@timoludwig 

In the region user list view and the networt management user list view, paginator with user name is already used.
Shall it be yet somehow changed?

timoludwig (commented):
Oh sorry, then it's a mistake in the issue. The user list is fine!

timoludwig (approved):

Issue #895, opened 2021-07-16
Error when creating a new page/event/poi without description


david-venhoff:
### Describe the Bug
Creating a new page/event/poi that has no description causes this error:
![grafik](https://user-images.githubusercontent.com/78504586/125942150-453235c1-580b-45c7-b037-a1a5865c424e.png)

### Steps to Reproduce
1. Create a new event, page or poi
2. Fill in all required fields, but leave the description empty
3. Click on save

### Expected Behavior
The server saves the new page, event or poi

### Actual Behavior
The server shows an error message.

PR #896, opened 2021-07-16
Fix crash when creating a new page/event/poi with an empty description

david-venhoff:
### Short description
This pr fixes the crash that occurs when creating a page/event/poi that has an empty description or a description that is for any other reason no valid html.

### Proposed changes
- Handle exceptions thrown by the function that parses html strings and return the string unmodified in case of an error

### Resolved issues
Fixes: #895


timoludwig (approved):
Thanks, this looks good! :+1:

Issue #714, opened 2021-02-25
Autosave changes previous version to draft


Thomas125:
### Describe the Bug
On autosave the prior revision is changed to draft-status.

### Steps to Reproduce

1. Go to pages
2. Open page-editor for an existing page which last revision is PUBLIC.
3. Make some change (e.g. in the text)
4. Wait two minutes (until the autosave is triggered)

### Expected Behavior
A new revision is created with status DRAFT. The revision that already existed, remains unchanged with status PUBLIC.

### Actual Behavior
A new revision is created with status DRAFT. The revision that already existed, remains unchanged, but gets status DRAFT.

### Additional Information
Probably, it also makes sense to add a new status for autosave translations.

This issue depends on:
- Issue #869
- Issue #870

timoludwig:
This is also related to bug #672 maybe this can be fixed at the same time?

MizukiTemma:
I have tried to reproduce the bug but it didn't go as described.

In the editor of the published page with title "initial" und text "initial", an additional word "asdf" was added in the text.
Leaving the editor after an auto save was triggered without any explicit saving, no new revision (version) appeared and the status stayed published (veröffentlicht).

Has the behaviour somehow changed inbetween?

timoludwig:
To reproduce and finally fix this issue, we have to fix #869 and #870 first.

About the issue itself: I think the problem is that a new version is only created if one of the fields "slug", "title" or "text" change:

https://github.com/Integreat/integreat-cms/blob/5d3a30ad71108662c3f45ce8674a71096bdf59f7/src/cms/forms/pages/page_translation_form.py#L103-L107

So if the autosave gets triggered while these values are unchanged, the existing version gets saved with the new draft status.
Normally, this doesn't happen so much, because tinymce is smart enough to trigger the autosave only when something changed, but if we e.g. insert a letter and remove it directly afterwards, tinymce will trigger an autosave but Django will still recognize that the values are unchanged and just sets the status to draft, which is probably not what the user expects.
To circumvent this, we could either create a new version each time the form is submitted, no matter which fields changed, or somehow make sure that the form isn't saved at all when nothing changed and the save is an autosave... For this, we might have to either send an additional parameter to the form submission, or create a new status "autosave" which impacts the form submission in a way that an unchanged form is no longer saved.

MizukiTemma:
@timoludwig 

[...]

PR #897, opened 2021-07-16
Fix autosave changing previous version to draft

MizukiTemma:
### Short description

This PR proposes a solution to the problem that the status of page becomes Draft when some changes are made but then deleted before the autosave,

### Proposed changes

- when a POST request, which changes the status of a page to Draft, is sent by an auto save but the changes are deleted and therefore the contents have not changed, the status will be overwritten to Public.
-
-

### Resolved issues

Fixes: #714

### DIFF
--- a/src/cms/forms/pages/page_translation_form.py
+++ b/src/cms/forms/pages/page_translation_form.py
@@ -44,7 +44,9 @@ class PageTranslationForm(CustomContentModelForm):
             # Copy QueryDict because it is immutable
             data = kwargs.pop("data").copy()
             # Update the POST field with the status corresponding to the submitted button
-            if "submit_draft" in data:
+            if "submit_auto" in data:
+                data["status"] = status.AUTO_SAVE
+            elif "submit_draft" in data:
                 data["status"] = status.DRAFT
             elif "submit_review" in data:
                 data["status"] = status.REVIEW
--- a/src/cms/locale/de/LC_MESSAGES/django.po
+++ b/src/cms/locale/de/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2021-07-20 09:29+0000\n"
+"POT-Creation-Date: 2021-07-21 10:03+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: Integreat <info@integreat-app.de>\n"
 "Language-Team: Integreat <info@integreat-app.de>\n"
@@ -1283,17 +1283,21 @@ msgstr "App-Team"
 msgid "Promo team"
 msgstr "Promo-Team"
 
-#: constants/status.py:16
+#: constants/status.py:18
 msgid "Draft"
 msgstr "Entwurf"
 
-#: constants/status.py:17
+#: constants/status.py:19
 msgid "Pending Review"
 msgstr "Review ausstehend"
 
-#: constants/status.py:18
+#: constants/status.py:20
 msgid "Published"
 msgstr "Veröffentlicht"
+
+#: constants/status.py:21
+msgid "Auto Save"
+msgstr "Automatische Speicherung"
 
 #: constants/text_directions.py:14
 msgid "Left to right"
@@ -5257,12 +5261,17 @@ msgstr ""
 "veröffentlichen, aber Sie können Änderungen vornehmen und diese zur "
 "Überprüfung vorlegen."
 
-#: views/events/event_view.py:196
+#: views/events/event_view.py:182 views/imprint/imprint_view.py:199
+#: views/pages/page_view.py:244 views/pois/poi_view.py:157
+msgid "No changes detected, autosave skipped"
+msgstr "Keine Änderungen vorgenommen, automatisches Speichern übersprungen"
+
+#: views/events/event_view.py:203
 msgid "Event \"{}\" was successfully created"
 msgstr "Veranstaltung \"{}\" wurde erfolgreich erstellt"
 
-#: views/events/event_view.py:213 views/pages/page_view.py:261
-#: views/pois/poi_view.py:173
+#: views/events/event_view.py:220 views/pages/page_view.py:268
+#: views/pois/poi_view.py:179
 msgid "No changes detected, but date refreshed"
 msgstr "Keine Änderungen vorgenommen, aber Datum aktualisiert"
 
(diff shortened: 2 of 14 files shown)


src/cms/forms/pages/page_translation_form.py
```diff
@@ -78,6 +78,8 @@ def save(self, commit=True):
         if not {"slug", "title", "text"}.isdisjoint(self.changed_data):
             self.instance.version += 1
             self.instance.pk = None
+        else:
+            self.instance.status = status.PUBLIC
```
timoludwig, line 82: Hmm, I'm not sure if this is the right way to go. This will set the status to public each time a new page is created, but sometimes, the users want to create a page as draft...
I think we need to introduce a special `AUTOSAVE` status to the [constants](https://github.com/Integreat/integreat-cms/blob/develop/src/cms/constants/status.py) which would also be helpful for people to decide whether they want to stick with the autosaved version or restore the last "normal" version...

MizukiTemma: When some changes are really made (not afterwards deleted before the auto save), it creates a new version as draft and the original stays public.

timoludwig: Oh right, sorry! But still there is the case when you make no changes but click on the button "Save as draft", it makes the status public although it should be draft...

MizukiTemma: I introduced, as you suggested, a new status AUTO for auto save and updated the behaviour of auto save.
Now the user can save the page as draft by cliking "save as draft" also when no real change has been made.

timoludwig (commented):
Thanks! :+1:
Yes, I think the new status is a good way to differentiate between the save methods.

For readability reasons, I propose to rename `AUTO` to `AUTO_SAVE`, and for this you probably need to increase the `max_length` of the `status` field to `9` in [src/cms/models/pages/abstract_base_page_translation.py](https://github.com/Integreat/integreat-cms/blob/8219586acbb98c681635210d0c45817285a0c13e/src/cms/models/pages/abstract_base_page_translation.py#L15)

If you apply my suggestions, we will have the original bug again (only with status `AUTO_SAVE` instead of `DRAFT`), so to fix this, I suggest to add the new case to [src/cms/views/pages/page_view.py](https://github.com/Integreat/integreat-cms/blob/8219586acbb98c681635210d0c45817285a0c13e/src/cms/views/pages/page_view.py#L239):
```
        elif (
            page_translation_form.instance.status == status.AUTO_SAVE
            and not page_form.has_changed()
            and not page_translation_form.has_changed()
        ):
            messages.info(request, _("No changes detected, autosave skipped"))
```

MizukiTemma:
Thank you for the suggestions!

I have changed the max length of status to 9. To compile,the  status in event_translation.py and poi_translation.py needed to be changed same too, so those are also in the committ.

src/cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -5758,11 +5762,15 @@ msgstr ""
 "Sie haben nicht die nötige Berechtigung, um diese Seite zu veröffentlichen, "
 "aber Sie können Änderungen vornehmen und diese zur Überprüfung vorlegen."
 
-#: views/pages/page_view.py:247
+#: views/pages/page_view.py:244
+msgid "No changes detected, autosave skipped"
+msgstr "Keine Änderungen vorgenommen, aber Datum aktualisiert"
```
timoludwig, line 5767:
```suggestion
msgstr "Keine Änderungen vorgenommen, automatisches Speichern übersprungen"
```

timoludwig (commented):
Thank you! I completely forgot about the other models, could you also add the `AUTO_SAVE` status to the following locations?

- [src/cms/forms/events/event_translation_form.py](https://github.com/Integreat/integreat-cms/blob/8219586acbb98c681635210d0c45817285a0c13e/src/cms/forms/events/event_translation_form.py#L46-L51)
- [src/cms/forms/pois/poi_translation_form.p](https://github.com/Integreat/integreat-cms/blob/8219586acbb98c681635210d0c45817285a0c13e/src/cms/forms/pois/poi_translation_form.py#L52-L55)
- [src/cms/forms/imprint/imprint_translation_form.py](https://github.com/Integreat/integreat-cms/blob/8219586acbb98c681635210d0c45817285a0c13e/src/cms/forms/imprint/imprint_translation_form.py#L35-L38)

as well as the form views?
- [src/cms/views/events/event_view.py](https://github.com/Integreat/integreat-cms/blob/8219586acbb98c681635210d0c45817285a0c13e/src/cms/views/events/event_view.py#L176)
- [src/cms/views/pois/poi_view.py](https://github.com/Integreat/integreat-cms/blob/8219586acbb98c681635210d0c45817285a0c13e/src/cms/views/pois/poi_view.py#L152)
- [src/cms/views/imprint/imprint_view.py](https://github.com/Integreat/integreat-cms/blob/8219586acbb98c681635210d0c45817285a0c13e/src/cms/views/imprint/imprint_view.py#L194)

Thanks in advance!

src/cms/views/events/event_view.py
```diff
@@ -173,6 +173,12 @@ def post(self, request, **kwargs):
             event_form.add_error_messages(request)
             event_translation_form.add_error_messages(request)
             recurrence_rule_form.add_error_messages(request)
+        elif (
+            event_translation_form.instance.status == status.AUTO_SAVE
+            and not event_form.has_changed()
+            and not event_translation_form.has_changed()
+        ):
```
timoludwig, line 180:
```suggestion
            and not recurrence_rule_form.has_changed()
        ):
```
I think the recurrence rule form has to be checked here as well...

src/cms/views/imprint/imprint_view.py
```diff
@@ -191,6 +191,12 @@ def post(self, request, *args, **kwargs):
         if not imprint_translation_form.is_valid():
             # Add error messages
             imprint_translation_form.add_error_messages(request)
+        elif (
+            imprint_translation_form.instance.status == status.AUTO_SAVE
+            and not imprint_form.has_changed()
```
timoludwig, line 196:
```suggestion
```
For the imprint, there is no `ImprintForm`.

src/cms/views/imprint/imprint_view.py
```diff
@@ -191,6 +191,12 @@ def post(self, request, *args, **kwargs):
         if not imprint_translation_form.is_valid():
             # Add error messages
             imprint_translation_form.add_error_messages(request)
+        elif (
+            imprint_translation_form.instance.status == status.AUTO_SAVE
```
timoludwig, line 195: You have to import status from the constants.

src/cms/views/pois/poi_view.py
```diff
@@ -149,6 +149,12 @@ def post(self, request, *args, **kwargs):
             # Add error messages
             poi_form.add_error_messages(request)
             poi_translation_form.add_error_messages(request)
+        elif (
+            poi_translation_form.instance.status == status.AUTO_SAVE
```
timoludwig, line 153: Here as well, please import the status constant.

timoludwig (commented):
Thanks! :+1:
Just a few small details left :smile:

timoludwig (approved):
Thank you! :1st_place_medal: 
I think this PR is ready to be merged :tada:

Issue #873, opened 2021-06-29
Filenames longer than 64 chars not uploadable


timoludwig:
### Describe the Bug

Files with a name of more than 64 characters cannot be uploaded.

### Steps to Reproduce

1. Go to the media library
2. Click on "Upload File"
3. Select a file with a name of 65 characters or more

### Expected Behavior

Sensible lengths of filenames should be supported, I think at least 255 characters. If the filename is too long, a good error message should indicate the problem to the user.

### Actual Behavior

The message "A network error has occurred. Please try again later." is shown to the user, which is not really what happened.
This is the case because the server did not respond with valid JSON.

### Additional Information

<details>
<summary>Server log</summary>

```
Jun 29 15:39:16 ERROR django.request - Internal Server Error: /ajax/media/upload-file
Traceback (most recent call last):
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.StringDataRightTruncation: value too long for type character varying(100)


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.7/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.7/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.7/site-packages/django/views/decorators/http.py", line 40, in inner
    return func(request, *args, **kwargs)
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.7/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/home/timo/job/integreat/integreat-cms/src/cms/decorators.py", line 53, in wrap
    return function(request, *args, **kwargs)
  File "/home/timo/job/integreat/integreat-cms/src/api/decorators.py", line 94, in wrap
    return function(request, *args, **kwargs)
  File "/home/timo/job/integreat/integreat-cms/src/cms/views/media/media_actions.py", line 132, in upload_file_ajax
    media_file = media_file_form.save()
[...]

PR #903, opened 2021-07-23
Change max size of file names

ulliholtgrave:
### Short description
Changes the max_size settings for the different file objects. This allows files with longer file names to be uploaded.

Fixes: #873


timoludwig (approved):
Thanks! :+1:

Issue #887, opened 2021-07-07
`TypeError` in event form


melegiul:
### Describe the Bug

TypeError is thrown for newly created events

### Steps to Reproduce

1. Go to form for creating a new event
2. Enter valid form data
3. Click save
4. See the type error

### Expected Behavior

New event is created by form submission

### Actual Behavior

Exception occurs

### Additional Information

```
TypeError at /augsburg/events/de/new
has_changed() missing 2 required positional arguments: 'initial' and 'data'
```
![Screenshot from 2021-07-07 13-46-24](https://user-images.githubusercontent.com/47010475/124753881-cfecfc00-df29-11eb-9343-7da29fca5054.png)

david-venhoff:
The error originates from this line: https://github.com/Integreat/integreat-cms/blob/3c93fcba93637d700c66a740de788c4a81625705/src/cms/views/events/event_view.py#L187
Apparently, the method `has_changed` takes two arguments which are not specified there. A possible fix would be to replace this line with
```python
            elif "is_recurring" in event_form.changed_data:
```

Unfortunately, this causes another error, when the user removes an existing recurrence rule:
![grafik](https://user-images.githubusercontent.com/78504586/125940529-d5979be8-4d1a-40d6-a1eb-5d84a91b5d8a.png)

This error originates from there:
https://github.com/Integreat/integreat-cms/blob/3c93fcba93637d700c66a740de788c4a81625705/src/cms/views/events/event_view.py#L190

To me it seems like the line above causes this error, because the deletion does not set the recurrence rule to null, just its foreign key. Adding a line below 189:
```python
                event_form.instance.recurrence_rule = None
```
fixes that. However I am not sure why exactly django is throwing the new error, so this is most likely not the best fix.

florianhofhammer:
A possible issue is that `has_changed` determines whether a field value has changed or not based on the field's initial value and the newly assigned (through the `POST`) value (see the [Django docs](https://docs.djangoproject.com/en/3.2/ref/forms/api/#django.forms.Form.has_changed)).
When doing some introspection on `event_form.fields["is_recurring"]`, we see that it's `initial` value is set to `None`. I assume that this is the reason for the error because the comparison then obviously fails.
I'm digging into it!

florianhofhammer:
[...]

PR #905, opened 2021-07-24
Fix `TypeError` in `EventView` caused by missing method arguments

florianhofhammer:
### Short description

This PR fixes a `TypeError` occurring when trying to save a newly created event. The error is due to missing arguments to the `has_changed` method invoked on a field of the form.

### Proposed changes

Steps to fix the error:

- Provide the necessary arguments to the method
- Check the arguments for validity (i.e., not invoking `NoneType` objects) before

Even though the `elif` clause looks a little bit messy and overloaded now, I chose to stick with the method call instead of another workaround because the method invoked on the field does checks that we'd have to account for in our workaround otherwise.

### Resolved issues

Fixes: #887

### DIFF
--- a/src/cms/views/events/event_view.py
+++ b/src/cms/views/events/event_view.py
@@ -191,7 +191,9 @@ class EventView(TemplateView, EventContextMixin, MediaContextMixin):
             if event_form.cleaned_data.get("is_recurring"):
                 # If event is recurring, save recurrence rule
                 event_form.instance.recurrence_rule = recurrence_rule_form.save()
-            elif event_form.fields["is_recurring"].has_changed():
+            elif event_instance and event_form.fields["is_recurring"].has_changed(
+                event_instance.is_recurring, event_form.cleaned_data.get("is_recurring")
+            ):
                 # If the event is not recurring but it was before, delete the associated recurrence rule
                 event_form.instance.recurrence_rule.delete()
             event_translation_form.instance.event = event_form.save()
--- a/src/cms/views/events/event_view.py
+++ b/src/cms/views/events/event_view.py
@@ -191,11 +191,10 @@ class EventView(TemplateView, EventContextMixin, MediaContextMixin):
             if event_form.cleaned_data.get("is_recurring"):
                 # If event is recurring, save recurrence rule
                 event_form.instance.recurrence_rule = recurrence_rule_form.save()
-            elif event_instance and event_form.fields["is_recurring"].has_changed(
-                event_instance.is_recurring, event_form.cleaned_data.get("is_recurring")
-            ):
+            elif event_form.instance.recurrence_rule:
                 # If the event is not recurring but it was before, delete the associated recurrence rule
                 event_form.instance.recurrence_rule.delete()
+                event_form.instance.recurrence_rule = None
             event_translation_form.instance.event = event_form.save()
             event_translation_form.save()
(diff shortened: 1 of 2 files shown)


src/cms/views/events/event_view.py
```diff
@@ -191,7 +191,9 @@ def post(self, request, **kwargs):
             if event_form.cleaned_data.get("is_recurring"):
                 # If event is recurring, save recurrence rule
                 event_form.instance.recurrence_rule = recurrence_rule_form.save()
-            elif event_form.fields["is_recurring"].has_changed():
+            elif event_instance and event_form.fields["is_recurring"].has_changed(
+                event_instance.is_recurring, event_form.cleaned_data.get("is_recurring")
+            ):
```
timoludwig, line 196:
```suggestion
            elif event_instance and event_instance.recurrence_rule:
```
Maybe, we can simplify this? If I understand it correctly, we want to delete the recurrence rule object in this case:
- The event exists
- The event was recurring before
- The event should not be recurring anymore

Since the last condition is already checked by the fact that the first if case `event_form.cleaned_data.get("is_recurring")` was not matched, I think we can just check the first two conditions instead of relying on the Django functionality `has_changed`... or am I missing something here?

src/cms/views/events/event_view.py
```diff
@@ -191,7 +191,9 @@ def post(self, request, **kwargs):
...
                 # If the event is not recurring but it was before, delete the associated recurrence rule
                 event_form.instance.recurrence_rule.delete()
```
timoludwig, line 198:
```suggestion
                event_form.instance.recurrence_rule.delete()
                event_form.instance.recurrence_rule = None
```
I still get the error `save() prohibited to prevent data loss due to unsaved related object 'recurrence_rule'.` when I change an event from recurring to non-recurring. I'm not completely sure what causes the error, but if I understood it correctly, it's the following:
- The `delete()` deletes the recurrence rule in the database, but it does not change the recurrence rule of the event instance
- The event instance still has the recurrence rule as before, with the only difference being that the recurrence rule is not in the database anymore
- Django sees that you want to save the event which has an unsaved foreign key object and throws the error

So I presume that we need both: deleting the recurrence object (so the database doesn't get polluted with old and unused recurrence rule objects) and to set the recurrence rule of the event explicitly to `None`.

timoludwig (commented):
Good to see you're back :rocket: :muscle: :1st_place_medal: 

Thanks for your PR, I'm not sure whether that's a problem on my side, but it doesn't fix the issue for me. See my comments below for my suggestions :blush:

src/cms/views/events/event_view.py
florianhofhammer: Sorry for taking so long to respond.

You're right, my solution checked the change more explicitly but as I mentioned already above, I was not very happy with my solution anyways.
Since we already know through the first ` if event_form.cleaned_data.get("is_recurring")` which evaluated to false that there should be no recurrence now, we implicitly know that there was a change if the object still exists.

src/cms/views/events/event_view.py
florianhofhammer: That is absolutely right, I totally missed that part.
Since I was only looking at the creation of new events according to the linked issue, this problem absolutely slipped through and I didn't test for this functionality.

timoludwig (approved):
Thank you! :+1: :tada:

Issue #898, opened 2021-07-17
Broken icon widget preview


timoludwig:
### Describe the Bug

At the moment, we have two problems with the preview of our custom icon widget:

1. If the media library overlay is closed without selecting a file, the preview shows a broken image placeholder.

    **Suggested Solution:**
    Only show the preview when the file is selected, not when the media library overlay is opened.

2. If a PDF file is selected as icon, the following `ValueError` is shown:
    ```
    ValueError at /organizations/1/edit
    
    The 'thumbnail' attribute has no file associated with it.
    ```

    **Suggested Solution:**
    Only make it possible to select images as icons, and not text documents. This may require some changes to the media library (especially the overlay which is used to select the files). Additionally, include a fallback for when the file is an image but has no valid thumbnail. In this case, the actual file should be used for the preview.

PR #909, opened 2021-07-24
Restrict file selection to images on icon fields.

ulliholtgrave:
### Short description
This PR adds a variable to the selection component to restrict the selection of icons fields.

### Proposed changes
- Add boolean field 'onlyImage' to differentiate between images and other documents.
- Only show icon field if image is set by the media library.

### Resolved issues

Fixes: #898 
Fixes: #915


src/frontend/js/media-management/library.tsx
```diff
@@ -40,14 +41,17 @@ export default function Library({
...
 }: Props) {
   // The directory path contains the current directory and all its parents
   const [directoryPath, setDirectoryPath] = useState<Directory[]>([]);
   // The current directory is the last element of the directory path
   const directory = directoryPath[directoryPath.length - 1];
   // The directory content contains all subdirectories and files of the current directory
-  const [directoryContent, setDirectoryContent] = useState<MediaLibraryEntry[]>([]);
+  const [directoryContent, setDirectoryContent] = useState<MediaLibraryEntry[]>(
+    []
+  );
```
timoludwig, line 54: I'm not sure if I really find this more readable than before... Maybe you can configure prettier with a slightly longer max line length? E.g. 88 characters like black or 100 chars?

timoludwig (commented):
Thanks :muscle: 

This fixes both issues with the icon field, I just experience a problem in the edge case where an image does not have a thumbnail (you can test this behaviour by setting the thumbnail to `null` in Django admin after uploading a file)...
Screenshot 2021-07-27 at 18-51-10 Integreat Redaktionssystem
Can you maybe create a `thumbnail_url` property which returns `thumbnail.url` if the thumbnail is set and has a fallback to `file.url` if the media file is an image without thumbnail?
Maybe we can also use this in the library itself to show the original image in the sidebar if the tumbnail doesn't exist (I think we would have to use the new `thumbnail_url` property in the `serialize()` method for this)...

src/frontend/js/media-management/library.tsx
ulliholtgrave: Yeah, Will do so :)

ulliholtgrave:
@timoludwig Can you think of any other cases then the media upload (e.g. https://github.com/Integreat/integreat-cms/blob/a76074db3cdb39356d625d3998cdfc2bec258103/src/cms/views/media/media_actions.py#L123) where a thumbnail doesn't gets created or doesn't exist?

ulliholtgrave:
@timoludwig I applied your proposal with the additional property for non-existing thumbnails. As I was already touching the media library, I implemented the Loading Spinner from #915 as well. Feel free to review my changes :)

src/frontend/js/media-management/library.tsx
```diff
@@ -237,17 +247,26 @@ export default function Library({
...
+              breadCrumbs={directoryPath}
               mediaTranslations={mediaTranslations}
-              selectionMode={selectionMode}
-              globalEdit={globalEdit}
             />
           </div>
+          {isLoading ? (
+            <div class="flex h-screen">
+            <Loader class="animate-spin m-auto w-1/6 h-1/6"></Loader>
+            </div>
```
timoludwig, line 258:
```suggestion
            <Loader class="absolute w-32 h-32 -mt-9 -ml-16 inset-1/2 text-gray-600 animate-spin"/>
```
I think it looks a bit strange when the loading screen exceeds the available screen space and one needs to scroll into empty space... Maybe this looks a bit better?
This just requires the `relative` class on the [parent element](https://github.com/Integreat/integreat-cms/blob/b2f3fce6d97fe28946b38559d12f6f7afdc59323/src/frontend/js/media-management/library.tsx#L246).

timoludwig (commented):
Thank you! :+1:

In general this looks very good, I just have two small ideas for improvement (see my comments below) and two bugs I experience:
1. When an image without thumbnail is selected as icon and the form is opened again, the error `The 'thumbnail' attribute has no file associated with it.` appears, so I propose to use the new property `thumbnail_url` [here](https://github.com/Integreat/integreat-cms/blob/b2f3fce6d97fe28946b38559d12f6f7afdc59323/src/cms/templates/icon_widget.html#L7) as well.
2. The icon widget doesn't work anymore after the page form has been submitted via POST, because the media context variables are missing... probably `context = self.get_context_data(**kwargs)` as to be added in the [POST view](https://github.com/Integreat/integreat-cms/blob/b2f3fce6d97fe28946b38559d12f6f7afdc59323/src/cms/views/pages/page_view.py#L273-L277) as well. This is not really a part of this issue, but maybe you can fix this so we don't have to open a separate issue for it?

sphinx/conf.py
```diff
@@ -118,6 +118,7 @@
     ("py:attr", "django.contrib.auth.models.User.feedback"),
     ("py:attr", "django.contrib.auth.models.User.mfa_keys"),
     ("py:attr", "django.contrib.auth.models.User.imprint_translations"),
+    ("py:attr", "django.utils.datastructures.MultiValueDict"),
```
timoludwig, line 121:
```suggestion
    ("py:class", "django.utils.datastructures.MultiValueDict"),
```
I think this is a class and not an attribute.

sphinx/conf.py
```diff
@@ -118,8 +118,10 @@
     ("py:attr", "django.contrib.auth.models.User.feedback"),
     ("py:attr", "django.contrib.auth.models.User.mfa_keys"),
     ("py:attr", "django.contrib.auth.models.User.imprint_translations"),
-    ("py:attr", "django.contrib.auth.models.User.page_translations"),
+    ("py:class", "django.utils.datastructures.MultiValueDict"),
+    ("py:class", "django.contrib.auth.models.User.page_translations"),
```
timoludwig, line 122:
```suggestion
```
The `class`-line can be removed since you added the correct `attr` two lines after

src/cms/models/media/media_file.py
```diff
@@ -130,6 +130,21 @@ def url(self):
...
+        """
+        Return the path of the image that should be used as the thumbnail.
+
+        :return: The path of the file. If the file is an image and no thumbnail could be generated the file itself will be returned.
+        :rtype: str
+        """
+        if not self.thumbnail:
+            if self.type.startswith("image"):
+                #: Returns the path to the file itself
+                return self.url
```
timoludwig, line 144:
```suggestion
                return BASE_URL + self.url
```
For consistency reasons, it's probably good to append the base url here as well...

src/frontend/js/media-management/library.tsx
```diff
@@ -237,17 +247,24 @@ export default function Library({
           onClick={() => setFileIndex(null)}
```
timoludwig, line 247: Could you add the class `relative` as suggested [here](https://github.com/Integreat/integreat-cms/pull/909#discussion_r698574827) to the div in line 246? I can't make an automated suggestion because it's too far away from your changes.
Otherwise, the spinner is a bit off the center (which shows especially on small screens):
Screenshot 2021-09-16 at 02-35-10 Integreat Redaktionssystem

timoludwig (commented):
Thanks a lot! :+1:

Just some minor details until this PR can be merged!

timoludwig (approved):
Amazing, thanks! :+1: :100:

Issue #907, opened 2021-07-24
Fix Translator View.


ulliholtgrave:
### Describe the Bug
At the moment our Translation View is broken, as a NoReverseMatch error arises.

### Steps to Reproduce

1. Go to a page
2. Click on 'Side by Side View'

### Expected Behavior
The side-by-side view is shown.

### Actual Behavior
See screenshot.

### Additional Information
![grafik](https://user-images.githubusercontent.com/15608678/126866957-665936c4-1e01-46f4-b6f0-56ad0c9c4ecf.png)

PR #911, opened 2021-07-28
Fix translator view

david-venhoff:
### Short description
This pr fixes the error that shows up when trying to open the page translation view.
Additionally, this pr fixes a problem where the page translation loads as readonly, even if the user has all required permissions.

### Proposed changes
- Provide the target language to `_tinymce_config.html` in the sbs view
- Fix the variable `can_change_page_object` not getting set at all

### Resolved issues
Fixes: #907


timoludwig (commented):
Thanks! :+1:
This works as expected and fixes the issue, can you adapt the same solution for the imprint side by side view as well?

timoludwig (approved):
Thanks! :+1:

Issue #912, opened 2021-07-28
Link context widget shows empty url when clicking on image


david-venhoff:
### Describe the Bug
In the tinymce editor the link information widget shows up when clicking on an element that is a child of a link. This causes the widget to show an empty url:

![grafik](https://user-images.githubusercontent.com/78504586/127308993-2c5e37a2-7036-4bcc-94d2-4485f09cc7ea.png)

### Steps to Reproduce
1. Insert an element that is child of a link into the tinymce editor (for example an image via the media library)
2. click onto that element
3. Notice that the link popup shows up and that it shows an empty url string

### Expected Behavior
The popup does not show up at all or the popup shows up and contains the correct url

### Actual Behavior
The popup shows up without url

PR #913, opened 2021-07-28
Show the correct url when clicking on a node that is a child of a link

david-venhoff:
### Short description
This pr fixes the tinymce link popup to show the correct url when the selected element is just a child of a link.

### Proposed changes
- Make the `getAnchor` function walk the node hierarchy until it finds an anchor element

### Resolved issues
Fixes: #912

### DIFF
--- a/src/frontend/js/tinymce-plugins/custom_link_input/plugin.js
+++ b/src/frontend/js/tinymce-plugins/custom_link_input/plugin.js
@@ -35,7 +35,9 @@ import { getCsrfToken } from "../../utils/csrf-token";
     }
 
     function updateLink(editor, anchorElm, text, linkAttrs) {
-        anchorElm.textContent = text;
+        if (text !== null) {
+            anchorElm.textContent = text;
+        }
 
         editor.dom.setAttribs(anchorElm, linkAttrs);
         editor.selection.select(anchorElm);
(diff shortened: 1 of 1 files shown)


timoludwig (approved):

I noticed one problem when I select an image and use the custom link pugin to create a link with this image via "Insert -> Link", it replaces the image with the link text... but I'm not sure how the original link plugin handled this case and I don't think it's related to this issue/PR.

david-venhoff:
I just checked and the original link plugin handled this case correctly, by only changing the parent elements url. I think it makes sense to include that in this pr for consistency, but I can also open a new issue and make a separate pr. What do you think?

timoludwig:
> I just checked and the original link plugin handled this case correctly, by only changing the parent elements url. I think it makes sense to include that in this pr for consistency, but I can also open a new issue and make a separate pr. What do you think?

I think both is fine, do whatever you prefer.

david-venhoff:
Now the text field in the link popup is disabled if the current link contains anything else other than text, because the popup can only handle text. This also includes simple styling, for example editing a link with a cursive word disables the text field. 
Also clicking `unlink` on a link that has nested elements should keep those elements and only delete the link itself now.

timoludwig (approved):
Amazing, works like a charm! :+1:

Issue #672, opened 2021-02-19
Autosave on new page form creates multiple new pages


timoludwig:
### Describe the Bug

If the new page form is opened, every 2 minutes a new page is created by the autosave functionality. We need to somehow keep track of the first autosave and update the existing draft afterwards instead of creating new ones.

### Steps to Reproduce

1. Go to a region
2. Create new page
3. Type in a valid title
4. Wait at least 2 minutes
5. See new page draft in the page tree

Additionally, the form gets also submitted when the user leaves the page without saving.

### Expected Behavior

Autosaves of new pages should update the same draft and not create new instances every 2 minutes

### Actual Behavior

Every 2 minutes a new page is created

### Additional Information

If the expected behavior is too hard to implement, we could also deactivate autosave on the page creation form.

This issue depends on:
- Issue #869
- Issue #870

MizukiTemma:
Here too I tried reproduce of the bag and found some stuffs.

In the text editor of a published page with title "inital" and text "inital, I added a new word "asdf" in the text.
After an auto save, the text change was not shown in the console, but by clicking "save as draft" the change appered in the console.

So I would like to suggest deactivating the autosave (because it does not reflect the change) and saving the contents in the same way with an explicit "save as draft" when the use leaves the editor.

*comment on the screen shot: the first two marked are from auto save, the third is from click on "save as draft"
![Inkedautosave1_LI](https://user-images.githubusercontent.com/82345046/122192783-4307d300-ce94-11eb-8470-c95982d25df2.jpg)

timoludwig:
@MizukiTemma I'm sorry it took me so long to get back to your question!

In general, we cannot deactivate autosave completely because our clients need this feature. Thus, we have to find out why it's not working and fix it.

The first steps are:
- For easier testing, I would recommend to set a lower `autosave_interval` in [src/frontend/js/forms/tinymce-init.ts](https://github.com/Integreat/integreat-cms/blob/5d3a30ad71108662c3f45ce8674a71096bdf59f7/src/frontend/js/forms/tinymce-init.ts#L89) so you don't have to wait 2 minutes each time you want to test the autosave feature
[...]

Issue #899, opened 2021-07-17
Autosave changes unversioned `Page` model


timoludwig:
### Describe the Bug

When an autosave is triggered, the complete `Page` and `PageTranslation` model instances are saved.
Since just the `PageTranslation` is versioned, this might lead to unexpected changes on public pages.

### Steps to Reproduce

1. Go to a page form
2. Edit something in the content textarea to trigger the tinymce autosave
3. Change some field in the page form, e.g. embed live content
4. Wait for the autosave to get triggered
5. Return without saving

### Expected Behavior

All page fields should be unchanged, since they are not versioned and there is no backup of the "public" state. If the page field is exited without saving, the page model instance should be exactly the same as before.

### Actual Behavior

All fields are automatically saved and the page instance is saved as well.

### Additional Information

Probably, this can be solved by limiting the `FormData` in [src/frontend/js/forms/autosave.ts](https://github.com/Integreat/integreat-cms/blob/8219586acbb98c681635210d0c45817285a0c13e/src/frontend/js/forms/autosave.ts#L7) so only the fields `slug`, `title` and `text`/`description` are saved.
_Edit:_ Probably not, since this would cause the validation to fail or remove specific values... so maybe we need a completely new view for the autosave?

PR #918, opened 2021-07-31
Autosave on new page form creates multiple new pages

MizukiTemma:
### Short description

This PR solves the problem that the auto save creates multiple of new versions instead of update the first one.

### Proposed changes

- When there is a change made, the latest version of the translation will be checked.
- If it has status Autosave, this one will be deleted and replaced with the new version by not increasing the version number.

In the second commit changed
- autosave.ts is changed as in the comment suggested.
- Line 101 of forms.pages.page_translation_form.py, which was already commented out, is deleted.
- Now only one page is created also for a content which has no published version. The problem that auto save changes unversioned page (page without published version) will be solved in the issue #899.

### Resolved issues

Fixes: #672

### DIFF
--- a/src/frontend/js/forms/autosave.ts
+++ b/src/frontend/js/forms/autosave.ts
@@ -6,11 +6,13 @@ export async function autosaveEditor() {
   tinymce.triggerSave();
   let formData = new FormData(form);
   formData.append("submit_auto", "true");
-  const data = await fetch(window.location.href, {
+  const data = await fetch(form.action, {
     method: "POST",
     headers: {
       "X-CSRFToken": getCsrfToken(),
     },
     body: formData
   });
+  // Set the form action to the url of the server response to make sure new pages aren't created multiple times
+  form.action = data.url;
 }


timoludwig (commented):
Thanks for your PR! :+1:

Unfortunately, this approach doesn't work. When a new page is created, there is no information about existing pages in the submitted form data. So the form instance of the creation form always has the version 0 and will be set to 1 and then saved. This means, that multiple pages are created for the same content.
The only way to fix this, is to let the server know whether a page is completely new or whether an autosaved version already exists in the database.
I described one way to do this in [this comment](https://github.com/Integreat/integreat-cms/issues/672#issuecomment-868985930):
> Then, we can think about solving the issue described in the original bug report. One possibility I see is that after the very first autosave, we change the action of the `#content_form` to the url which is returned by the ajax call. When we submit data to the url `http://localhost:8000/augsburg/pages/de/new`, it will redirect to the new url of the new created page, e.g. `http://localhost:8000//augsburg/pages/de/81/edit`. If we now replace the form action and use the form action for future ajax calls instead of `window.location.href` [here](https://github.com/Integreat/integreat-cms/blob/5d3a30ad71108662c3f45ce8674a71096bdf59f7/src/frontend/js/forms/autosave.ts#L6), the page will be updated in the future instead of creating new page instances every time.

Do you understand what I mean or should I elaborate this approach?

timoludwig (approved):
Thank you! :+1:

I reverted your changes of the first commit, because it doesn't help with fixing the issue and it's not required to delete the previous autosaved versions when the form is submitted. But the new changes are fine and the problem with multiple pages being created is resolved :tada:

Issue #694, opened 2021-02-23
Improve event form


timoludwig:
### Proposed Solution

- [ ] Move status and version to the top right corner (see #691)
- [x] Switch labels for end-date and end-time
- [x] Add german translation of "Erstelle einen Veranstaltungsort oder beginne den Namen eines bestehenden Veranstaltungsortes einzutippen." to "Erstellen Sie einen Veranstaltungsort oder beginnen Sie den Namen eines bestehenden Veranstaltungsortes einzutippen."
- [ ] Add checkbox "Event does not have a physical location" and only show location fields if checkbox is not checked
- [ ] Fix the poi ajax query for the non-staff-users (verwaltung, event planner are getting 403 here)
- [ ] On change of start-date and start-time, copy the value into end-date and end-time if they are empty
- [x] Remove minor edit field
- [ ] Use the same dynamic slug/link update mechanism like for pages

ulliholtgrave:
A good first approach to that issue is the page_form template, which includes a lot of the desired UI changes.

PR #919, opened 2021-08-05
Improve event form

melegiul:
### Short description

Improves/repairs a couple of features for the event form

### Proposed changes

- [x] Moves status and version to top right corner
- [x] Adds checkbox "Event does not have physical location for disabling location field with corresponding form validation
- [x] Fix poi ajax query permission problem
- [x] On change of start date/time auto completes end date/time
- [x] Repair slug/link update mechanism (was just a permission issue)

### Resolved issues

Fixes: #694


src/cms/templates/events/event_form.html
```diff
@@ -109,19 +109,27 @@ <h1 class="heading">
...
+                        {% if event_translation_form.instance.id %}
+                            <span class="mr-8">
+                                <label class="inline-block">{% trans 'Status' %}:</label>
+                                {{ event_translation_form.instance.get_status_display }}
+                            </span>
+                        {% endif %}
+                    </div>
                     {% render_field event_translation_form.title|add_error_class:"border-red-500" %}
                     <div id="link-container" class="flex items-center mt-4">
                         <label for="{{ event_translation_form.slug.id_for_label }}">
```
timoludwig, line 135:
```suggestion
                        <label for="{{ event_translation_form.slug.id_for_label }}" class="mr-2">
```
I think adding a small margin here would look better.

src/cms/templates/events/event_form.html
```diff
@@ -227,32 +235,36 @@ <h1 class="heading">
...
-                        <div class="relative my-2">
-                            <input id="poi-query-input" type="search" autocomplete="off" class="pr-8 appearance-none block w-full bg-white text-gray-800 placeholder-gray-800 border border-gray-400 rounded py-3 px-4 leading-tight focus:outline-none focus:bg-white focus:placeholder-gray-600"{% if event_form.disabled %} disabled{% endif %} placeholder="{% if poi %}{{ poi|poi_translation_title:current_language }}{% else %}{{ poi_title_placeholder }}{% endif %}" data-url="{% url 'search_poi_ajax' %}" data-region-slug="{{ region.slug }}" data-default-placeholder="{{ poi_title_placeholder }}">
-                            <div class="absolute inset-y-0 right-0 flex items-center px-2 text-gray-800">
-                                <button id="poi-remove" title="{% trans 'Remove location from event' %}"{% if event_form.disabled %} disabled{% endif %}>
-                                    <i data-feather="trash-2" class="h-5 w-5"></i>
-                                </button>
+                        {% render_field event_form.has_not_location %}
+                        {{ event_form.has_not_location.label_tag }}
```
timoludwig, line 239: Interesting, I didn't know about this tag.
I think we should use this for all forms in the future :smile:

src/frontend/js/events/auto-complete.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * this file contains function for autocompletion of the event form fields
+ */
+
+window.addEventListener("load", () => {
+    if (document.getElementById("id_start_date") && document.getElementById("id_start_time")) {
+        document.getElementById("id_start_date").addEventListener("change", autocompleteEndDate);
+        document.getElementById("id_start_time").addEventListener("change", autocompleteEndTime);
```
timoludwig, line 8:
```suggestion
        document.getElementById("id_start_date").addEventListener("focusout", autocompleteEndDate);
        document.getElementById("id_start_time").addEventListener("focusout", autocompleteEndTime);
```
I would set the value only after the input field looses focus, because otherwise the value gets set as soon as the start field has a valid input, e.g.:
- Type `01.01.2021` in the start date, then the end date will be `01.01.0002`
- Type `10:30` in the start time, the the end time will be `10:03`

So with `focusout`, I think the most obvious problems should be fixed.

src/frontend/js/events/conditional-fields.ts
```diff
@@ -76,4 +76,18 @@ window.addEventListener("load", () => {
...
+  );
+  if (locationCheckbox) {
+    locationCheckbox.addEventListener("click", () => {
+      document
+        .getElementById("location-block")
+        .classList.toggle("hidden");
+      let locationInput = document.getElementById("id_location") as HTMLInputElement;
+      locationInput.disabled = !locationInput.disabled;
+    });
+  }
```
timoludwig, line 92:
```suggestion
  document
    .getElementById("id_has_not_location")
    ?.addEventListener("click", () => {
      document
        .getElementById("location-block")
        .classList.toggle("hidden");
      const locationInput = document.getElementById("id_location") as HTMLInputElement;
      locationInput.disabled = !locationInput.disabled;
    });
```
I think with TypeScript, we can simplify this a little bit.

timoludwig (commented):
Thanks a lot! :+1:

Everything seems to work fine, I just have a few small comments:

src/frontend/js/events/conditional-fields.ts
melegiul: I just googled that syntax, its called the Elvis operator. Best name for an operator so far :smile:

melegiul:
I forgot to fix a problem with the form validation, too.
Screenshot from 2021-08-06 10-46-51
I will have a look at this.

src/frontend/js/events/conditional-fields.ts
timoludwig: Haha, indeed that's a good name :laughing: 
However, in this context, `?.` is not the Elvis operator (which returns the first part if it evaluates to true), but rather reads as "continue with this evaluation if it's true until here", see documentation about [Optional Chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html)...
This way, we don't have to check explicitly, whether an element exists, but can just do something if it exists and ignore the code otherwise.

src/cms/views/events/event_view.py
```diff
@@ -172,7 +172,9 @@ def post(self, request, **kwargs):
             # Add error messages
             event_form.add_error_messages(request)
             event_translation_form.add_error_messages(request)
-            recurrence_rule_form.add_error_messages(request)
+            # do not call recurrence rule form clean method when recurrence rule is not set
+            if event_form.cleaned_data["is_recurring"]:
+                recurrence_rule_form.add_error_messages(request)
```
melegiul, line 177: Ok, so this one way to make sure that in case of an invalid event form the recurrence rule form is not validated, when the rule is not set. This is because `add_error_messages` will trigger a call to the recurrence rule forms clean method, which is not expected when the recurrence rule is disabled

src/frontend/js/events/conditional-fields.ts
melegiul: Its a handy operator, I will keep it in mind :+1:

timoludwig (approved):
Thanks a lot, looks good! :+1: :tada:

PR #923, opened 2021-08-18
Autosave changes unversioned Page model

MizukiTemma:
### Short description

This PR contains a suggestion to solve the problem that autosave changes unversioned page models.

### Proposed changes

- Before saving page and pagetranslation objects, it will be checked in page_view.py, whether the save is triggered by an auto save, a page objects already exisits, and there is a public version backup.-

### Resolved issues

Fixes: #899

### DIFF
--- a/src/cms/views/pages/page_view.py
+++ b/src/cms/views/pages/page_view.py
@@ -244,8 +244,13 @@ class PageView(TemplateView, PageContextMixin, MediaContextMixin):
             messages.info(request, _("No changes detected, autosave skipped"))
 
         else:
-            # Save forms
-            page_translation_form.instance.page = page_form.save()
+            # Only save page form if page does not yet exist or if translation is no auto save
+            if (
+                not page_instance
+                or page_translation_form.instance.status != status.AUTO_SAVE
+            ):
+                page_translation_form.instance.page = page_form.save()
+            # Save page translation form
             page_translation_form.save()
             # Add the success message and redirect to the edit page
             if not page_instance:
--- a/src/cms/locale/de/LC_MESSAGES/django.po
+++ b/src/cms/locale/de/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2021-08-25 15:43+0000\n"
+"POT-Creation-Date: 2021-08-30 10:06+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: Integreat <info@integreat-app.de>\n"
 "Language-Team: Integreat <info@integreat-app.de>\n"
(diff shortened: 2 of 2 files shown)


src/cms/views/pages/page_view.py
```diff
@@ -245,30 +245,46 @@ def post(self, request, *args, **kwargs):
...
-                    **{
-                        "page_id": page_form.instance.id,
-                        "region_slug": region.slug,
-                        "language_slug": language.slug,
-                    },
+            if (page_translation_form.instance.status != status.AUTO_SAVE) or (
+                not (
+                    page_form.instance.id
+                    and (
+                        not Page.objects.filter(id=page_form.instance.id)
```
timoludwig, line 252: This query doesn't make much sense. Why do an additional database query to get the page instance with `Page.objects.filter(id=page_form.instance.id).first()` when you already have the instance with `page_form.instance`?

src/cms/views/pages/page_view.py
```diff
@@ -245,30 +245,46 @@ def post(self, request, *args, **kwargs):
 
         else:
             # Save forms
```
timoludwig, line 247:
```suggestion
            # Only save page form if page does not yet exist or if translation is no auto save
```

src/cms/views/pages/page_view.py
```diff
@@ -245,30 +245,46 @@ def post(self, request, *args, **kwargs):
...
+                        ),
+                    )
+                    return redirect(
+                        "edit_page",
+                        **{
+                            "page_id": page_form.instance.id,
+                            "region_slug": region.slug,
+                            "language_slug": language.slug,
+                        },
+                    )
```
timoludwig, line 276: I don't really understand what's the purpose of this if condition. If I try to translate it into human language, it does the following:

_If the translation status is not `AUTO_SAVE` or if there is no public translation of the page, save both the page translation form and the page form_

But I don't really see how that fixes the issue. This completely ignores the autosaved version if a public version exists (what is not what should happen) and it saves the page form if no public version exists (which should also not happen). Instead, we want that on every autosave, the page translation form is saved (no matter if there exists a public version or not) and the page form should be ignored (with the exception when no page exists at all - then we need the page form to be saved, otherwise the page translation form cannot be saved as well).
```suggestion
            if not page_instance or page_translation_form.instance.status != status.AUTO_SAVE:
                page_translation_form.instance.page = page_form.save()
            # Save page translation form
            page_translation_form.save()
            # Add the success message and redirect to the edit page
            if not page_instance:
                messages.success(
                    request,
                    _('Page "{}" was successfully created').format(
                        page_translation_form.instance.title
                    ),
                )
                return redirect(
                    "edit_page",
                    **{
                        "page_id": page_form.instance.id,
                        "region_slug": region.slug,
                        "language_slug": language.slug,
                    },
                )
```

src/cms/views/pages/page_view.py
```diff
@@ -245,30 +245,46 @@ def post(self, request, *args, **kwargs):
...
-            if not page_form.has_changed() and not page_translation_form.has_changed():
-                messages.info(request, _("No changes detected, but date refreshed"))
+                if (
+                    not page_form.has_changed()
+                    and not page_translation_form.has_changed()
+                ):
+                    messages.info(request, _("No changes detected, but date refreshed"))
+                else:
+                    # Add the success message
+                    page_translation_form.add_success_message(request)
```
timoludwig, line 285:
```suggestion
            if not page_form.has_changed() and not page_translation_form.has_changed():
                messages.info(request, _("No changes detected, but date refreshed"))
```

src/cms/views/pages/page_view.py
```diff
@@ -245,30 +245,46 @@ def post(self, request, *args, **kwargs):
...
             else:
-                # Add the success message
-                page_translation_form.add_success_message(request)
+                print("No change saved")
```
timoludwig, line 287:
```suggestion
                # Add the success message
                page_translation_form.add_success_message(request)
```

MizukiTemma: Suggestion comitted

timoludwig (approved):

I stashed all commits and resolved the conflicts in the translation file.

PR #924, opened 2021-08-18
Ajax RSS feed

membralala:
### Short description
Re-implementation of the rss-news-feed in ajax. 

### Proposed changes
This PR features the discussed rss-feed improvements so that it allows an asynchronous loading of news-entries. This allows an independent timeout handling and also makes it possible to load more feed-entries dynamically. 

Fixes: #801

### DIFF
--- a/src/frontend/js/toggle-dashboard-section.ts
+++ b/src/frontend/js/toggle-dashboard-section.ts
@@ -2,7 +2,9 @@ window.addEventListener("load", () => {
   document.querySelectorAll(".collapsible").forEach((node) => {
     node.addEventListener("click", toggleDashboardSection);
   });
+  loadRssFeed();
 });
+
 
 /**
  * Handles toggling boxes on the dashboard
@@ -17,4 +19,93 @@ async function toggleDashboardSection({ target }: Event): Promise<void> {
   const content = collapsible.parentNode.querySelector(".collapsible-content");
   content.classList.toggle("max-h-160");
   content.classList.toggle("max-h-0");
+}
+
+
+async function loadRssFeed() {
+  const feedWidget = document.querySelector(".rss-feed");
+
+  if (feedWidget) {
+    const feedLoading = document.getElementById("feed-loading");
+    const reloadButton = document.getElementById("reload");
+
+    const handleError = () => {
+      feedLoading.classList.add("hidden");
+      reloadButton.classList.remove("hidden");
+      reloadButton.addEventListener("click", loadRssFeed);
+    };
+    
+    feedLoading.classList.remove("hidden");
+    reloadButton.classList.add("hidden");
+    const url = feedWidget.attributes[1].value;
+
+    // set timeout
+    const controller = new AbortController();
+    const timeoutId = setTimeout(() => {
+      controller.abort();
+      handleError();
+    }, 5000);
+
+    fetch(url, {signal: controller.signal})
+      .then((response: Response) => {
+        if (response.ok && response.status === 200) {
+          return response.text();
+        } 
+        throw new Error("Error during fetching of RSS feed");
+      }).then((rawFeed: string) => {
+        feedLoading.classList.add("hidden");
+        clearTimeout(timeoutId);
+        let parser = new DOMParser;
+        let feedHtml = parser.parseFromString(rawFeed, "text/xml");
+        console.log(feedHtml);
+
+        let firstEntry = 0, lastEntry = 3; 
+        feedWidget.innerHTML += loadRssEntries(feedHtml, firstEntry, lastEntry);
+        const readMoreButton = document.getElementById("read-more")
+        readMoreButton.classList.remove("hidden");
+        readMoreButton.addEventListener("click", () => {
+          
+          firstEntry = lastEntry;
+          lastEntry += 3;
+
+          feedWidget.innerHTML += loadRssEntries(feedHtml, firstEntry, lastEntry);
+
+          if (feedHtml.querySelectorAll("item").length - 1 <= firstEntry) {
+            readMoreButton.classList.add("hidden");
+
+            let entries = document.querySelectorAll(".rss-entry");
+            entries[entries.length - 1].classList.remove("border-b")
+            entries[entries.length - 1].classList.remove("border-gray-400")
+            entries[entries.length - 1].classList.remove("mb-2");
+          }     
+        }); 
+    }).catch ((err) => {
+      // console.error(err);
+      handleError();
+    });
+  }
+}
+
+
+function loadRssEntries(doc: Document, firstEntry: number, lastEntry: number) : String {
+  let feed = "";
+
+  Array.from(doc.querySelectorAll("item")).slice(firstEntry, lastEntry).forEach((node) => {
+    let link:String = node.querySelector("link").innerHTML;
+    let title:string = node.querySelector("title").innerHTML;
+    let content:String = node.querySelector("description").textContent;
+    if (title && link && content) {
+      let rssEntry = '<div class="rss-entry pb-2 border-b border-gray-400 mb-2">'
+        + '<a href="' 
+        + link 
+        + '" target="_blank" rel="noopener noreferrer" class="block py-b font-bold text-blue-500 hover:text-blue-700">'
+        + title 
+        + '</a>' + '<div class="text-gray-600 py-2"></div>'
+        + '<p class="pb-1">' + content
+        + '</div>'
+
+      feed += rssEntry;
+    }
+  }); 
+  return feed;
 }
--- a/src/cms/locale/de/LC_MESSAGES/django.po
+++ b/src/cms/locale/de/LC_MESSAGES/django.po
@@ -8,7 +8,11 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: \n"
+<<<<<<< HEAD
 "POT-Creation-Date: 2021-08-20 07:57+0000\n"
+=======
+"POT-Creation-Date: 2021-08-18 14:02+0000\n"
+>>>>>>> 57b8f69f ( implemented ajax rbased rss news feed with timeout handling)
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: Integreat <info@integreat-app.de>\n"
 "Language-Team: Integreat <info@integreat-app.de>\n"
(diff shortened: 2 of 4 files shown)


src/frontend/js/toggle-dashboard-section.ts
```diff
@@ -18,3 +21,91 @@ async function toggleDashboardSection({ target }: Event): Promise<void> {
...
+    const url = feedWidget.attributes[1].value;
+
+    // set timeout
+    const controller = new AbortController();
+    const timeoutId = setTimeout(() => {
+      controller.abort();
+    }, 5000);
+
+    fetch(url, {signal: controller.signal}).then(async (response: Response) => {
+      if (response.status === 200) {
```
melegiul, line 44: I justed tested the error handling case e. g. with an invalid url like integreat-app.de/fed and noticed that we need two more things:
1. On a controller timeout we need similar functionality (replacing the loading icon by the reload button) like in the catch block to prevent an infinite loading spin
2. The catch block cannot capture 404 errors so I think something like `if !reponse.ok { throw new Error("some error")}` would be helpful: 
```
A fetch() promise will reject with a TypeError when a network error is encountered or CORS is misconfigured on the
server-side, although this usually means permission issues or similar — a 404 does not constitute a network error, for
example. An accurate check for a successful fetch() would include checking that the promise resolved, then checking
that the Response.ok property has a value of true. 
```

melegiul (changes requested):
Thank you very much, thats a very concise solution. :+1: 
I just tested it in both languages and everything just works.
Only for the error handling I would suggest some improvements

src/frontend/js/toggle-dashboard-section.ts
membralala: Thanks for your feedback! I'm glad that the major part works fine. As far as I understand it, my changes should fix the two issues you pointed out.

src/frontend/js/toggle-dashboard-section.ts
```diff
@@ -18,3 +21,96 @@ async function toggleDashboardSection({ target }: Event): Promise<void> {
...
+      handleError();
+    }, 5000);
+
+    fetch(url, {signal: controller.signal}).then(async (response: Response) => {
+      if (response.status === 200 && response.ok) {
+        feedLoading.classList.add("hidden");
+        clearTimeout(timeoutId);
+        let parser = new DOMParser;
+        let feedHtml = parser.parseFromString(
+          parser.parseFromString(await response.text(), "text/xml").activeElement.innerHTML, "text/html");
```
melegiul, line 56: On firefox everything is fine, but on chrome the inner call of `parseFromString` returns a xmlDocument with activeElement set to null which throws a TypeError, which you can see if you pass an argument called `error` to the catch function.
For a solution I could found one common element of the xmlDocument (set in chrome and in firefox), which is called `documentElement`, which we could use here instead.
Can you confirm the problem on chrome?

melegiul (changes requested):
Thanks a lot for your changes.
I'm sorry, but I just found another problem when using chrome.

src/frontend/js/toggle-dashboard-section.ts
membralala: No worries as you are absolutely right again. I can confirm the issue and your proposed solution seems to work perfectly fine in chrome as well as in firefox. Thanks again!

src/frontend/js/toggle-dashboard-section.ts
```diff
@@ -18,3 +20,97 @@ async function toggleDashboardSection({ target }: Event): Promise<void> {
...
+  Array.from(doc.querySelectorAll("item")).slice(firstEntry, lastEntry).forEach((node) => {
+    let title: String, link: String, content: String;
+
+    // get title, link and content this way as they don't have a consistent index in the HTMLCollection
+    Array.from(node.children).forEach(element => {
+      if (element.localName === "title") title = element.textContent;
+      if (element.localName === "link") link = element.nextSibling.nodeValue;
+      if (element.localName === "content:encoded") content = element.textContent;
+    });
+
```
melegiul, line 100:
```suggestion
    let link:String = node.querySelector("link").innerHTML;
    let title:string = node.querySelector("title").innerHTML;
    let content:String = node.querySelector("description").textContent;
```
Instead of looping through all elements we can use queryselectors here since node is a document

src/frontend/js/toggle-dashboard-section.ts
```diff
@@ -18,3 +20,97 @@ async function toggleDashboardSection({ target }: Event): Promise<void> {
...
+    if (title && link && content) {
+      let rssEntry = '<div class="rss-entry pb-2 border-b border-gray-400 mb-2">'
+        + '<a href=' 
+        + link 
+        + 'target="_blank" rel="noopener noreferrer" class="block py-b font-bold text-blue-500 hover:text-blue-700">'
+        + title 
+        + '</a>' + '<div class="text-gray-600 py-2"></div>'
+        + '<p class="pb-1">' + content.split(" ").slice(0, 25).join(" ") + ' . . . </p>'
```
melegiul, line 108:
```suggestion
        + '<p class="pb-1">' + content + '</p>'
```

src/frontend/js/toggle-dashboard-section.ts
```diff
@@ -18,3 +20,97 @@ async function toggleDashboardSection({ target }: Event): Promise<void> {
...
+      if (element.localName === "title") title = element.textContent;
+      if (element.localName === "link") link = element.nextSibling.nodeValue;
+      if (element.localName === "content:encoded") content = element.textContent;
+    });
+
+    if (title && link && content) {
+      let rssEntry = '<div class="rss-entry pb-2 border-b border-gray-400 mb-2">'
+        + '<a href=' 
+        + link 
+        + 'target="_blank" rel="noopener noreferrer" class="block py-b font-bold text-blue-500 hover:text-blue-700">'
```
melegiul, line 105:
```suggestion
        + '<a href="' 
        + link 
        + '" target="_blank" rel="noopener noreferrer" class="block py-b font-bold text-blue-500 hover:text-blue-700">'
```
This just wraps the link in quotes

src/frontend/js/toggle-dashboard-section.ts
```diff
@@ -18,3 +20,97 @@ async function toggleDashboardSection({ target }: Event): Promise<void> {
...
+
+    // set timeout
+    const controller = new AbortController();
+    const timeoutId = setTimeout(() => {
+      controller.abort();
+      handleError();
+    }, 5000);
+
+    fetch(url, {signal: controller.signal}).then(async (response: Response) => {
+      if (response.status === 200 && response.ok) {
```
melegiul, line 50:
```suggestion
    fetch(url, {signal: controller.signal})
      .then((response: Response) => {
        if (response.ok && response.status === 200) {
          return response.text();
        } 
        throw new Error("Error during fetching of RSS feed");
      }).then((rawFeed: string) => {
```
This is useful, because we can resolve the promise first and don't have to use async function latter for parsing once we have received the RSS feed

src/frontend/js/toggle-dashboard-section.ts
```diff
@@ -18,3 +20,97 @@ async function toggleDashboardSection({ target }: Event): Promise<void> {
...
+
+            let entries = document.querySelectorAll(".rss-entry");
+            entries[entries.length - 1].classList.remove("border-b")
+            entries[entries.length - 1].classList.remove("border-gray-400")
+            entries[entries.length - 1].classList.remove("mb-2");
+          }     
+        }); 
+      } else {
+        throw Error("NetworkError. " + url + "does not respond.");
+      }
```
melegiul, line 79:
```suggestion
```

src/frontend/js/toggle-dashboard-section.ts
```diff
@@ -18,3 +20,97 @@ async function toggleDashboardSection({ target }: Event): Promise<void> {
...
+      handleError();
+    }, 5000);
+
+    fetch(url, {signal: controller.signal}).then(async (response: Response) => {
+      if (response.status === 200 && response.ok) {
+        feedLoading.classList.add("hidden");
+        clearTimeout(timeoutId);
+        let parser = new DOMParser;
+        let feedHtml = parser.parseFromString(
+          parser.parseFromString(await response.text(), "text/xml").documentElement.innerHTML, "text/html");
```
melegiul, line 55:
```suggestion
       let feedHtml = parser.parseFromString(rawFeed, "text/xml");
```
Just calling the inner `parseFromString` should be enough to get the HTML document

src/frontend/js/toggle-dashboard-section.ts
```diff
@@ -18,3 +20,97 @@ async function toggleDashboardSection({ target }: Event): Promise<void> {
...
+
+        let firstEntry = 0, lastEntry = 3; 
+        feedWidget.innerHTML += await loadRssEntries(feedHtml, firstEntry, lastEntry);
+        const readMoreButton = document.getElementById("read-more")
+        readMoreButton.classList.remove("hidden");
+        readMoreButton.addEventListener("click", async () => {
```
melegiul, line 61:
```suggestion
        readMoreButton.addEventListener("click", () => {
```

src/frontend/js/toggle-dashboard-section.ts
```diff
@@ -18,3 +20,97 @@ async function toggleDashboardSection({ target }: Event): Promise<void> {
...
+          
+          firstEntry = lastEntry;
+          lastEntry += 3;
+
+          feedWidget.innerHTML += await loadRssEntries(feedHtml, firstEntry, lastEntry);
```
melegiul, line 66:
```suggestion
          feedWidget.innerHTML += loadRssEntries(feedHtml, firstEntry, lastEntry);
```

src/frontend/js/toggle-dashboard-section.ts
```diff
@@ -18,3 +20,97 @@ async function toggleDashboardSection({ target }: Event): Promise<void> {
...
+    fetch(url, {signal: controller.signal}).then(async (response: Response) => {
+      if (response.status === 200 && response.ok) {
+        feedLoading.classList.add("hidden");
+        clearTimeout(timeoutId);
+        let parser = new DOMParser;
+        let feedHtml = parser.parseFromString(
+          parser.parseFromString(await response.text(), "text/xml").documentElement.innerHTML, "text/html");
+
+        let firstEntry = 0, lastEntry = 3; 
+        feedWidget.innerHTML += await loadRssEntries(feedHtml, firstEntry, lastEntry);
```
melegiul, line 58:
```suggestion
        feedWidget.innerHTML += loadRssEntries(feedHtml, firstEntry, lastEntry);
```

src/frontend/js/toggle-dashboard-section.ts
```diff
@@ -18,3 +20,97 @@ async function toggleDashboardSection({ target }: Event): Promise<void> {
...
+      }
+    }).catch ((err) => {
+      // console.error(err);
+      handleError();
+    });
+  }
+}
+
+
+async function loadRssEntries(doc: Document, firstEntry: number, lastEntry: number) : Promise<String> {
```
melegiul, line 88:
```suggestion
function loadRssEntries(doc: Document, firstEntry: number, lastEntry: number) : Promise<String> {
```

melegiul (changes requested):
Thanks again, now everything works on both browsers.
I found just a few last improvements and then this is ready to merge :tada:

src/frontend/js/toggle-dashboard-section.ts
membralala: That's much better for sure.

src/frontend/js/toggle-dashboard-section.ts
membralala: But this one would always render the full article to the news feed, no matter how long it is, wouldn't it? Wouldn't it make sense to limit it to snippet, so that interested readers just check the link and get the full and well formatted article in a new window as it was in the old feed?

melegiul (approved):
Awesome, now this should be ready to merge :muscle: 
Do you know how to squash commits?
If no thats no problem I can do that for you, too

Issue #925, opened 2021-08-25
Children API endpoint default behavior


melegiul:
### Describe the Bug

When no query string is present in the url, all root nodes should be selected

### Steps to Reproduce

1. Go to [https://cms-django-dev.integreat-app.de/augsburg/de/wp-json/extensions/v3/children](https://cms-django-dev.integreat-app.de/augsburg/de/wp-json/extensions/v3/children)
2. See that no pages were selected

### Expected Behavior

All root nodes should be selected and returned with their children (with specified depth)

### Actual Behavior

When neither id nor url is present in the query string, the API returns an error

PR #926, opened 2021-08-25
Children API: use root pages as ancestors per default

melegiul:
### Short description

Fixes API error when no query string is supplied

### Proposed changes

When neither id nor url is supplied by API children request, use all root pages + their children (with specified depth)

### Resolved issues

Fixes: #925


ulliholtgrave (approved):
I checked it locally and it seems to fix the problem :+1:

Issue #63, opened 2018-10-26
Object-based search on list overviews


sascha11110:
To speed up navigation, an entity based search would also help, limiting it to only the entity displayed in the current list.

svenseeberg:
Entity again is page/event/location?

sascha11110:
Yep

ulliholtgrave:
@svenseeberg I suggest to move the issue to the Version 1.0 milestone as search fields are important feature that we have in wordpress as well. Your opinion?

timoludwig:
This is needed for the following model lists:

- Region list
- Admin Media Library
- Admin User list
- Admin/Technical Feedback
- Region Media Library
- Page Tree
- Event list
- POI List
- Push News
- Region Feedback
- Region User list

The implementation for all lists should be nearly identical, so please try to use shared utility functions as much as possible.
Only the search for the media library is different, since it is implemented in preact native.

If feasible, try to do dynamic AJAX calls which then render a sub-template of the table contents instead of reloading the whole page.
Also think about including additional fields where suitable, e.g. also searching the content of pages if there is no exact match in the titles... (or directly search all fields but show matches in the name/title fields at the top) - but that's optional and can be skipped if it's too complicated.


Lists which can probably be omitted:
- Languages
- Roles
- Organizations
- Offer Templates
- Language Tree

ulliholtgrave:
First solution for version 1.0 can be the search by extending the views. Then the page is reloaded each time, but it is by far the simplest variant. However, it needs another button for the search field.

svenseeberg:
The page tree view should implement the search in JavaScript, as all pages are already present in the browser and searching in the back end will take a relatively long time.

PR #929, opened 2021-08-31
Search functionality

david-venhoff:
### Short description
This pr implements search functionality for various views, including: region list, admin/region user list, admin/region feedback, pages, pois, events, push news.
Still missing:
  * Media library search (#931)

### Proposed changes
- Extend the `search_content_ajax.py` endpoint for more models
- Implement a generic search input template which makes ajax requests to `search_content_ajax`
- Modify views to allow filtering the rendered objects

### Resolved issues
Fixes: #63

### DIFF
--- a/src/cms/forms/events/event_filter_form.py
+++ b/src/cms/forms/events/event_filter_form.py
@@ -62,3 +62,12 @@ class EventFilterForm(forms.Form):
     )
 
     poi_id = forms.IntegerField(widget=forms.HiddenInput, initial=-1, required=False)
+
+    query = forms.CharField(required=False)
+
+    def filters_visible(self):
+        """
+        :return: whether any filters (other than search were changed)
+        :rtype: bool
+        """
+        return self.has_changed() and self.changed_data != ["query"]
--- a/src/cms/forms/feedback/region_feedback_filter_form.py
+++ b/src/cms/forms/feedback/region_feedback_filter_form.py
@@ -56,3 +56,11 @@ class RegionFeedbackFilterForm(forms.Form):
         ),
         required=False,
     )
+    query = forms.CharField(required=False)
+
+    def filters_visible(self):
+        """
+        :return: whether any filters (other than search were changed)
+        :rtype: bool
+        """
+        return self.has_changed() and self.changed_data != ["query"]
--- a/src/cms/models/events/event_translation.py
+++ b/src/cms/models/events/event_translation.py
@@ -1,6 +1,7 @@
 from django.conf import settings
 from django.contrib.contenttypes.fields import GenericRelation
 from django.db import models
+from django.db.models import Q
 from django.utils.translation import ugettext_lazy as _
 
 from linkcheck.models import Link
@@ -283,6 +284,31 @@ class EventTranslation(models.Model):
         :rtype: bool
         """
         return not self.currently_in_translation and not self.is_outdated
+
+    @classmethod
+    def search(cls, region, language_slug, query):
+        """
+        Searches for all event translations and match the given `query` in their title or slug.
+
+        :param region: The current region
+        :type region: ~cms.models.regions.region.Region
+        :param language_slug: The language slug
+        :type language_slug: str
+        :param query: The query string used for filtering the events
+        :type query: str
+        :return: A query for all matching objects
+        :rtype: ~django.db.models.query.QuerySet
+        """
+        return (
+            cls.objects.filter(
+                **{
+                    "event__region": region,
+                    "language__slug": language_slug,
+                }
+            )
+            .filter(Q(slug__icontains=query) | Q(title__icontains=query))
+            .distinct("event")
+        )
 
     def __str__(self):
         """
--- a/src/cms/views/events/event_list_view.py
+++ b/src/cms/views/events/event_list_view.py
@@ -13,7 +13,7 @@ from backend.settings import PER_PAGE
 
 from ...constants import all_day, recurrence
 from ...decorators import region_permission_required, permission_required
-from ...models import Region
+from ...models import Region, EventTranslation
 from ...forms import EventFilterForm
 from .event_context_mixin import EventContextMixin
 
@@ -48,6 +48,7 @@ class EventListView(TemplateView, EventContextMixin):
         return self.template_archived if self.archived else self.template
 
     # pylint: disable=too-many-branches
+    # pylint: disable=too-many-locals
     def get(self, request, *args, **kwargs):
         """
         Render events list for HTTP GET requests
@@ -93,6 +94,8 @@ class EventListView(TemplateView, EventContextMixin):
 
         # all events of the current region in the current language
         events = region.events.filter(archived=self.archived)
+
+        query = None
 
         # Filter events according to given filters, if any
         filter_data = kwargs.get("filter_data")
@@ -157,6 +160,13 @@ class EventListView(TemplateView, EventContextMixin):
                 ):
                     # Only non-recurring events
                     events = events.filter(recurrence_rule__isnull=True)
+                # Filter events by the search query
+                query = event_filter_form.cleaned_data["query"]
+                if query:
+                    event_ids = EventTranslation.search(
+                        region, language_slug, query
+                    ).values("event__pk")
+                    events = events.filter(pk__in=event_ids)
         else:
             event_filter_form = EventFilterForm()
             event_filter_form.changed_data.clear()
@@ -178,6 +188,7 @@ class EventListView(TemplateView, EventContextMixin):
                 "languages": region.languages,
                 "filter_form": event_filter_form,
                 "filter_poi": poi,
+                "search_query": query,
             },
         )
 
--- a/src/cms/views/pois/poi_list_view.py
+++ b/src/cms/views/pois/poi_list_view.py
@@ -9,9 +9,11 @@ from django.utils.translation import ugettext as _
 from django.views.generic import TemplateView
 
 from backend.settings import PER_PAGE
+from cms.models.pois.poi_translation import POITranslation
 
 from ...decorators import region_permission_required, permission_required
 from ...models import Region, Language
+from ...forms import ObjectSearchForm
 from .poi_context_mixin import POIContextMixin
 
 logger = logging.getLogger(__name__)
@@ -43,6 +45,7 @@ class POIListView(TemplateView, POIContextMixin):
         """
         return self.template_archived if self.archived else self.template
 
+    # pylint: disable=too-many-locals
     def get(self, request, *args, **kwargs):
         """
         Render POI list
@@ -98,7 +101,19 @@ class POIListView(TemplateView, POIContextMixin):
                 )
                 % {"language": region.default_language.translated_name},
             )
+
         pois = region.pois.filter(archived=self.archived)
+        query = None
+
+        search_data = kwargs.get("search_data")
+        search_form = ObjectSearchForm(search_data)
+        if search_form.is_valid():
+            query = search_form.cleaned_data["query"]
+            poi_keys = POITranslation.search(region, language_slug, query).values(
+                "poi__pk"
+            )
+            pois = pois.filter(pk__in=poi_keys)
+
         # for consistent pagination querysets should be ordered
         paginator = Paginator(pois.order_by("region__slug"), PER_PAGE)
         chunk = request.GET.get("page")
@@ -113,6 +128,22 @@ class POIListView(TemplateView, POIContextMixin):
                 "archived_count": region.pois.filter(archived=True).count(),
                 "language": language,
                 "languages": region.languages,
+                "search_query": query,
                 **context,
             },
         )
+
+    def post(self, request, *args, **kwargs):
+        """
+        Apply the query and filter the rendered pois
+
+        :param request: The current request
+        :type request: ~django.http.HttpResponse
+        :param args: The supplied arguments
+        :type args: list
+        :param kwargs: The supplied keyword arguments
+        :type kwargs: dict
+        :return: The rendered template response
+        :rtype: ~django.template.response.TemplateResponse
+        """
+        return self.get(request, *args, **kwargs, search_data=request.POST)
--- a/src/cms/views/push_notifications/push_notification_list_view.py
+++ b/src/cms/views/push_notifications/push_notification_list_view.py
@@ -10,8 +10,9 @@ from django.views.generic import TemplateView
 
 from backend.settings import PER_PAGE
 
+from ...forms import ObjectSearchForm
 from ...decorators import region_permission_required, permission_required
-from ...models import Language, Region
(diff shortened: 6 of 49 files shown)


timoludwig (commented):
Thanks, really cool! :+1: :rocket: 

I really like the dynamic part of the search! However, I think we also need the filtered result list, since it's not obvious what the user wants to do with the result. For example, in the region list, the user might want to edit the region or he might want to open the region dashboard.

So I see two options:
1. When an entry in the dynamic result list is selected, fill in the search field with this value and trigger an enter, which should reload the page and pass the search parameter to the view (ideally via GET), which should then show the filtered list.
2.  Alternatively, we could modify the ajax endpoint to return already the rendered list instead of a JSON response. The JS should then replace the existing list with the rendered html from the response. The only issue I see is that I'm not sure about how pagination would work here...

If you have any detailed questions, we can also schedule a short call about this topic.

david-venhoff:
I Implemented the first approach you described: When the user hits enter after typing in a search query, the page will reload and only show the matching objects. For now I decided against using GET, so that the search is more consistent with other filters. I can still change the method to use GET, but maybe it would make sense to also switch to that method for the other filters?
Also some views (For example the [page tree view](https://github.com/Integreat/integreat-cms/blob/e3c49bb45efdbce49b461eaf4eee80b8976a35d4/src/cms/views/pages/page_tree_view.py#L32) contain an archived property, which, when `True`, only shows archived objects. I did not include this in the search functionality yet. How and when is this variable used?

src/cms/views/utils/search_content_ajax.py
```diff
@@ -91,35 +62,141 @@ def search_content_ajax(request, region_slug, language_slug):
...
+            }
+            for region in Region.search(query)
+        )
+
+    if user.has_perm("auth.view_user") and "user" in object_types:
+        if (
+            user.is_superuser
+            or user.is_staff
+            or (region and region in user.profile.regions.all())
+        ):
```
timoludwig, line 185: I'm not sure, this should already be checked by the decorator `@region_permission_required`, or am I missing something?

src/frontend/css/style.scss
```diff
@@ -142,14 +142,12 @@ label:not([for]) {
...
-      left: 14px;
-    }
+#table-search {
+  svg {
+    width: 16px;
+    height: 16px;
+    top: 14px;
+    left: 14px;
   }
 }
```
timoludwig, line 152:
```suggestion
```
This can be deleted as mentioned earlier. Additionally, the lines
```
  &.rounded-r-none {
    border-top-right-radius: 0;
    border-bottom-right-radius: 0;
  }
```
should be added in line 95 (the definitions for form elements).

timoludwig (commented):
Awesome, thanks for your changes! :rocket: 

I still have some suggestions for improvement, feel free to ask if anything is unclear.

One general problem seems to be that we have different requirements for the search in the lists and the search of the link suggestion tool: In the first one, we want to include archived elements when we're filtering the archived list, and in the latter, we only want public and non-archived results. If it gets too complicated to fit all these cases into one big function, it might be a better idea to use different views for the filter search and the link suggestion search... 

Apart from my detailed comments, I have to additional ideas:
- When a search doesn't return results, I think it's better to show "No ... found" instead of "No ... available yet", which is a bit misleading.
- All archived lists don't work with searching at the moment. It would be cool to support it for archived lists as well.

Both of these ideas can be addressed in separate issues, since it's already quite some work do to in the current scope.

src/cms/views/pois/poi_list_view.py
```diff
@@ -98,9 +101,32 @@ def get(self, request, *args, **kwargs):
...
+                    )
+                ]
+
+        pois = (
+            pois
+            if pois is not None
+            else list(
+                region.pois.filter(archived=self.archived).order_by("region__slug")
+            )
+        )
```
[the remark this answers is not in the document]
david-venhoff: Yes, this makes a lot more sense. I was initially a bit confused about the archived variable.

src/cms/views/push_notifications/push_notification_list_view.py
```diff
@@ -74,9 +75,29 @@ def get(self, request, *args, **kwargs):
...
+                        for push_translation in PushNotificationTranslation.search(
+                            region, language_slug, query
+                        ).order_by("push_notification__created_date")
+                    ]
+
+        if push_notifications is None:
+            query = None
+            push_notifications = list(
+                region.push_notifications.all().order_by("created_date")
+            )
```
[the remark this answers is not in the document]
david-venhoff: Right now, the search is only invalid if the length of the query is 0. Wouldn't it make sense in this case to not filter at all?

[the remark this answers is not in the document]
timoludwig: Ah okay. Maybe you could do something like this?
```
search_form = ObjectSearchForm(kwargs.get("search_data"))
if search_form.is_valid():
    ...
else:
   ...
```
(initializing the `ObjectSearchForm` even if the `search_data` is `None` has a little bit of overhead, but it shouldn't matter much)

Another approach might be to perform the filtering on an existing queryset instead of the class `PushNotificationTranslation`, but I'm afraid that's a bit too much work for now...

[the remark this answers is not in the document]
david-venhoff: Since queries only get evaluated when they are used, are there any downsides with this approach?
```python
push_notifications = region.push_notifications.all()
query = None

search_data = kwargs.get("search_data")
if search_data:
    search_form = ObjectSearchForm(search_data)
    if search_form.is_valid():
        query = search_form.cleaned_data["query"]
        if query:
            push_notification_keys = set(
                push_translation.push_notification.pk
                for push_translation in PushNotificationTranslation.search(
                    region, language_slug, query
                )
            )
            push_notifications = push_notifications.filter(pk__in=push_notification_keys)
```
This has the upside to also work if at some point other filters specific to the view get implemented, like at the event/page/poi views

src/frontend/js/search-query.ts
```diff
@@ -0,0 +1,106 @@
...
+            }
+            // Schedule function execution
+            scheduledFunction = setTimeout(
+                queryObjects,
+                300,
+                table_search_input.getAttribute("data-url"),
+                table_search_input.getAttribute("data-object-type"),
+                table_search_input.value,
+            );
+        });
```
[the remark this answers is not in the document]
david-venhoff: Thank you, but I reused mostly the same code from [there](https://github.com/Integreat/integreat-cms/blob/develop/src/frontend/js/events/event-query-pois.ts) 🙂

src/frontend/js/search-query.ts
```diff
@@ -0,0 +1,106 @@
...
+            let search_suggestion = document.getElementById("table-search-suggestions");
+            search_suggestion.classList.add("hidden");
+        });
+
+    table_search
+        .addEventListener("focusin", (event) => {
+            let search_suggestion = document.getElementById("table-search-suggestions");
+            search_suggestion.classList.remove("hidden");
+        });
+}
```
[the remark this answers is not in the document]
david-venhoff: This should make sense. One thing I am not too sure about is whether this is actually required, since a user can just click search (or press enter) when the specific object they are searching for shows up. So I thought that it might be useful to have a "quick action", but I understand it it might be too much.

src/frontend/js/search-query.ts
```diff
@@ -0,0 +1,106 @@
...
+            );
+        });
+
+    let table_search = document.getElementById("table-search");
+
+    table_search
+        .addEventListener("focusout", (event) => {
+            if (event.relatedTarget && (event.relatedTarget as HTMLElement).closest("#table-search")) {
+                return;
+            }
```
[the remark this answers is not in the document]
david-venhoff: I think this was used for an earlier version of the search input and then I forgot to remove it...

src/cms/views/utils/search_content_ajax.py
david-venhoff: No, I missed that...

src/cms/forms/events/event_filter_form.py
```diff
@@ -62,3 +62,12 @@ class EventFilterForm(forms.Form):
     )
 
     poi_id = forms.IntegerField(widget=forms.HiddenInput, initial=-1, required=False)
+
+    query = forms.CharField(required=False)
+
+    def filters_visible(self):
```
svenseeberg, line 68: I'd add a `@property` decorator here.

src/cms/forms/feedback/region_feedback_filter_form.py
```diff
@@ -56,3 +56,11 @@ class RegionFeedbackFilterForm(forms.Form):
         ),
         required=False,
     )
+    query = forms.CharField(required=False)
+
+    def filters_visible(self):
```
svenseeberg, line 61: This could also have a `@property`.

src/cms/models/events/event_translation.py
```diff
@@ -284,6 +285,30 @@ def is_up_to_date(self):
         """
         return not self.currently_in_translation and not self.is_outdated
 
+    @classmethod
+    def search(cls, region, language_slug, query):
+        """
+        Searches for all event translations and match the given `query` in their title or slug.
```
svenseeberg, line 291: Not really an issue, but I think we usually put an empty line between the description, parameters and return values.

src/cms/views/events/event_list_view.py
```diff
@@ -94,6 +95,8 @@ def get(self, request, *args, **kwargs):
         # all events of the current region in the current language
         events = region.events.filter(archived=self.archived)
```
svenseeberg, line 96: This method needs some refactoring. But if you like you could move this line right in front of the paginator and only set it if it is not set by any of the many if branches.

src/cms/views/regions/region_list_view.py
```diff
@@ -42,11 +43,40 @@ def get(self, request, *args, **kwargs):
         :return: The rendered template response
         :rtype: ~django.template.response.TemplateResponse
         """
+
         regions = Region.objects.all()
+        query = None
+
+        search_data = kwargs.get("search_data")
+        if search_data:
+            search_form = ObjectSearchForm(search_data)
```
svenseeberg, line 52: If `ObjectSearchForm` could deal with a `None` parameter, then the previous check (in some other places as well) would not be necessary.

src/cms/views/regions/region_list_view.py
```diff
@@ -42,11 +43,40 @@ def get(self, request, *args, **kwargs):
...
+            if search_form.is_valid():
```
svenseeberg, line 53: Does it make sense to check for `search_form.cleaned_data["query"]` in the `is_valid()` method of `ObjectSearchForm` Then we could also skip the following check here and other places.

svenseeberg (changes requested):
Overall, this already looks pretty solid. Some of my previous notes are just rambling. However, I'd ask you to improve the 3 consecutive if checks when loading the search form that occurs several times in this PR.

src/cms/views/events/event_list_view.py
timoludwig: @svenseeberg I'm not sure if that works, since the `events` variable is not completely replaced in the if-branches, but only filtered further.
However, in the initial solution @david-venhoff implemented it similar like you suggested (see [this comment](https://github.com/Integreat/integreat-cms/pull/929#discussion_r709652412)), but I requested a change earlier.

In view of our performance problems, I even have a completely new suggestion: instead of implementing the search as a model method, we could define the search function as a QuerySet method which can be chained to other database filters and thus reducing the amount of queries used:
```
class EventQuerySet(models.QuerySet):
    def search(self, language_slug, query):
        return self.filter(
            translations__language__slug=language_slug,
            translations__title__icontains=query,
        )

class Event(models.Model):
    ...
    objects = models.Manager.from_queryset(EventQuerySet)()
    ...
```
Then, it could be used e.g. like this:
```
events = region.events.filter(archived=self.archived)
...
query = event_filter_form.cleaned_data["query"]
if query:
    events = events.search(language_slug, query)
```
and in the end, only one sql would be performed, since [querysets are lazy](https://docs.djangoproject.com/en/3.2/topics/db/queries/#querysets-are-lazy)... what do you think of this approach?
Since all models have slightly different search filters, we would probably need one custom QuerySet class for each model...

src/cms/views/regions/region_list_view.py
timoludwig: Since `ObjectSearchForm` is just a regular Django form, passing in `None` as data is not a problem (in fact, it's a keyword parameter, so it makes no difference if you call `ObjectSearchForm()`, `ObjectSearchForm(None)` or `ObjectSearchForm(data=None)`).

src/cms/forms/events/event_filter_form.py
timoludwig: Just out of curiosity, I checked in which cases Django uses the `@property` decorator for forms, but I couldn't find a consistent pattern :sweat_smile: 
For example, similar form-methods like `form.is_valid()`, `form.has_changed()` or `non_field_errors()` are methods, but `form.errors` and `form.changed_data` are properties...

timoludwig (commented):
Thanks a lot for your changes! :1st_place_medal: 
I'm very sorry about the back and forth, but as described below, I have yet another idea how we could improve the performance and readability of the code :sweat_smile:

src/cms/views/events/event_list_view.py
svenseeberg: Oh sorry, I did not read carefully enough the following lines. Just merely saw the `events =`

david-venhoff: Should this be implemented for this pr, or should we open an issue for that?

david-venhoff: That makes sense. However, I am not sure how to handle the ajax search with this implementation. One option would be to keep it as it is right now, but that would leave us with some duplicate search methods (one version to search the objects, and one version to search the object translations). Or we could remove the old search methods and only use the new `QuerySet` based methods, but then there would be two sql queries required for the ajax search.

timoludwig: Ah okay, sorry, I wasn't aware that this requires such substantial changes. In this case, I think we should just remember this idea for later (maybe opening a new refactoring issue with low priority) and merge the current solution. The user experience of the search is already pretty good and a huge step towards the migration :+1:

timoludwig (commented):
Really cool, thanks!
I think all what's left to do now is to clean up the git history a bit and rebase your branch onto develop to resolve the conflicts.

david-venhoff:
I squashed the commits and rebased the branch onto develop.. There were a few merge conflicts regarding the user changes, but everything should work now.

timoludwig (approved):
Really cool, thanks a lot! :100: :tada:

Issue #928, opened 2021-08-30
API: page attributes path and url differ with WordPress


svenseeberg:
### Describe the Bug
https://cms-dev.integreat-app.de/bocholt/en/wp-json/extensions/v3/children?depth=1 and https://cms-dev.integreat-app.de/bocholt/en/wp-json/extensions/v3/children?depth=1 return different content for page / page translation objects. In WordPress, the full URL is returned in `url` and in `path` the full path part of the URL is returned. Django only returns the page slug itself in `path` and `url` is missing the domain and schema.

WordPress:
```
url: "https://cms-dev.integreat-app.de/bocholt/en/corona-virus/"
path: "/bocholt/en/corona-virus/"
```

Django:
```
url: "bocholt/en/information-on-the-corona-virus"
path: "information-on-the-corona-virus"
```

### Expected Behavior
The values in Django need to be the same as in WordPress.

PR #930, opened 2021-09-01
Fix API page response format for url and path

melegiul:
### Short description

Fix API page response format for url and path

### Proposed changes

- Prepends domain of backend to url translation path, reads domain from settings value `BASE_URL` (or is `WEBAPP_URL` expected here?)
- Sets path to permalink surrounded by slashes

Also updates url and path for these objects (included in page/children response):
1. parent
2. available_languages (path was not present before)

### Resolved issues

Fixes: #928

### DIFF
--- a/src/api/v3/pages.py
+++ b/src/api/v3/pages.py
@@ -29,17 +29,23 @@ def transform_page(page_translation):
             "id": page_translation.page.parent.id,
             "url": page_translation.page.parent.get_translation(
                 page_translation.language.slug
-            ).permalink,
-            "path": page_translation.page.parent.get_translation(
+            ).backend_base_link,
+            "path": "/"
+            + page_translation.page.parent.get_translation(
                 page_translation.language.slug
-            ).slug,
+            ).permalink
+            + "/",
         }
     else:
-        parent = None
+        parent = {
+            "id": 0,
+            "url": None,
+            "path": None,
+        }
     return {
         "id": page_translation.id,
-        "url": page_translation.permalink,
-        "path": page_translation.slug,
+        "url": page_translation.backend_base_link,
+        "path": "/" + page_translation.permalink + "/",
         "title": page_translation.title,
         "modified_gmt": page_translation.combined_last_updated,
         "excerpt": page_translation.text,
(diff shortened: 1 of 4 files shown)


svenseeberg:
The url and path variables look good now. I can see that there are more pages returned in https://cms-django-dev.integreat-app.de/albdonaukreis/en/wp-json/extensions/v3/children?depth=1 than in https://cms-dev.integreat-app.de/albdonaukreis/en/wp-json/extensions/v3/children?depth=1. I guess that the depth parameter is used differently. Also, the app currently breaks with the response. I asked the app team for help

src/api/v3/pages.py
```diff
@@ -29,17 +29,19 @@ def transform_page(page_translation):
             "id": page_translation.page.parent.id,
```
svenseeberg, line 29: Until now the API returned a `0` if there is no page parent. With Django, the response currently contains `null`. We need to change this as well.

svenseeberg (changes requested):

melegiul:
> The `url` and `path` variables look good now. I can see that there are more pages returned in https://cms-django-dev.integreat-app.de/albdonaukreis/en/wp-json/extensions/v3/children?depth=1 than in https://cms-dev.integreat-app.de/albdonaukreis/en/wp-json/extensions/v3/children?depth=1. I guess that the depth parameter is used differently. Also, the app currently breaks with the response. I asked the app team for help.

I was wondering about that behavior, too. cms-django-dev with depth=0 (no children) seems to return same response (9 elements) like cms-dev with depth=1 (with immediate childern), same seems to continue for other depth values:
- cms-django-dev with depth=1: 49 elements
- cms-dev with depth=2: 50 elements

Maybe related to WP-45 issue?

svenseeberg:
I think the difference is the following: our Django CMS assumes that the default depth is 0 while WordPress assumes the default depth is 1. Also the interpretation of the values may differ. For WP a depth of 1 means the page itself plus its direct children. In case of a missing `url` parameter that means some sort of virtual root element plus its direct children, which corresponds to the first level of pages. If a page is provided, the page itself plus its children are returned. That may explain the "of by one" problem.

svenseeberg (approved):

Issue #927, opened 2021-08-26
Creating a new push notification causes the server to crash


david-venhoff:
### Describe the Bug
The server crashes with "'str' object has no attribute 'exists'" when creating a new push notification:
![grafik](https://user-images.githubusercontent.com/78504586/130939788-6b3dad7b-c430-4772-878b-3787de645482.png)

It looks like the push notification gets saved anyway, because it shows up in the push notifications list after the crash.

### Steps to Reproduce
1. Go to the news tab in a region
2. Click on 'create new message'
3. Fill the form and click on 'save & send'

### Expected Behavior
The server saves the push notification without crashing

### Actual Behavior
The server crashes

PR #932, opened 2021-09-01
Fix push notifications API token

membralala:
### Short description
This should fix the server crash when trying to send a push notification.

### Proposed changes
- The parsing of the fcm-authentification-token was deprecated. Now it will handle the token as a string and not as a database entry.

Fixes: #927

### DIFF
--- a/src/cms/views/push_notifications/push_notification_sender.py
+++ b/src/cms/views/push_notifications/push_notification_sender.py
@@ -38,7 +38,10 @@ class PushNotificationSender:
         if len(self.primary_pnt.title) > 0:
             self.prepared_pnts.append(self.primary_pnt)
         self.load_secondary_pnts()
-        self.auth_key = self.get_auth_key()
+
+        self.auth_key = settings.FCM_KEY
+        if not self.auth_key:
+            logger.warning("Could not get a proper fcm_auth_key")
 
     def load_secondary_pnts(self):
         """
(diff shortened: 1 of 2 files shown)


src/cms/views/push_notifications/push_notification_sender.py
```diff
@@ -81,14 +81,10 @@ def get_auth_key():
...
-        if auth_key.exists():
-            logger.debug("Got fcm_auth_key from database")
-            return auth_key.first().value
-        logger.warning(
-            "Could not get %r from configuration database", fcm_auth_config_key
-        )
+        if auth_key and isinstance(auth_key, str):
+            return auth_key
+        logger.warning("Could not get a proper fcm_auth_key")
         return None
```
timoludwig, line 88: Actually, I think if we just set the default missing key in the settings file to `FCM_KEY = None` instead of `FCM_KEY = ""`, and an existing key to a non-empty string, we could remove the `get_auth_key()` function completely. Just set the auth key to the key specified in the settings file in line 41 and everything should work as expected...
(the logging warning can also be shown in the `__init__` function if the key is `None`...

timoludwig (commented):
Thanks! This works as expected, however, I have a small idea for improvement:

src/cms/views/push_notifications/push_notification_sender.py
membralala: I thought about this, but wasn't super sure, because it might be less intuitive, that the `FCM_KEY` in settings has to be a string. But I can just add this information as a comment.

timoludwig: A valid auth token for the firebase service is always a string, there is no other possibility to define this value. The only ambiguity we have is how an "undefined" key should be represented.

We could also rule out any misunderstandings here if we replace all occurrences of `if self.auth_key is None:` with `if not self.auth_key:`. This way, all "falsey" values (e.g. `""`, `False`, `None`, `0`, `{}`, ...) would disable the sending of push notifications regardless their type.

svenseeberg:
A minor note: the title of the pull request is a little ambiguous. "Fix push notifications API token" or something like that would be more insightful.

src/cms/views/push_notifications/push_notification_sender.py
membralala: Thanks for your explanation! I implemented the changes you considered.

PR #933, opened 2021-09-08
Change path of medialibrary, fixes #902

### LINKED ISSUE #902: Include region id in media upload directory
### Motivation

At the moment, media files are uploaded in the directory structure `%Y/%m/%d` (see [src/cms/models/media/media_file.py](https://github.com/Integreat/integreat-cms/blob/fe5639a0e783c5e3622982da2340d04887d533ba/src/cms/models/media/media_file.py#L21)) containing the date of the upload.

### Proposed Solution

To make the migration of wordpress files and debugging in the future easier, we should include the region id parameter in the upload directories, so that the files of each region are separated. This can be done by changing the `upload_to` parameter of the `FileField` to a function which returns the path with the region id (see [docu](https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.FileField.upload_to)).

### Alternatives

The change is not necessarily required, we could also leave the directory structure as it is.

JoeyStk:
### Short description

Change path of media library, fixes #902

### Proposed changes

-
-
-

### Resolved issues

Fixes: #902

### DIFF
--- a/dev-tools/prune_database.sh
+++ b/dev-tools/prune_database.sh
@@ -53,3 +53,7 @@ fi
 # Remove migrations
 rm -rfv "${BASE_DIR}/src/cms/migrations"
 echo "Removed database migrations" | print_info
+
+# Remove media files (because they are no longer usable without the corresponding database entries)
+rm -rfv "${BASE_DIR}/src/media"
+echo "Removed media files" | print_info
--- a/src/backend/settings.py
+++ b/src/backend/settings.py
@@ -620,9 +620,13 @@ STATICFILES_FINDERS = (
     "django.contrib.staticfiles.finders.AppDirectoriesFinder",
 )
 
+
+#################
+# MEDIA LIBRARY #
+#################
+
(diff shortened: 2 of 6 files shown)


timoludwig (approved):
@JoeyStk Thanks for your PR! :tada:

I had a look at the thumbnail-problem and resolved it myself, since it was a bit more complex than initially thought. But from my side, this PR is ready to merge now. :+1:

I changed the thumbnail generation process to make sure the `upload_to`-function is actually used by Django to determine the thumbnail upload path (before, the path was calculated manually).
Furthermore, I added two small changes:
- Add a separate `upload_to`-function for the thumbnails to make sure that files and their respective thumbnail files have the exact same file name (apart from the `_thumbnail`-suffix)
- Remove media files when deleting the database (since media files aren't usable anyway when the corresponding database entries are missing).

@ulliholtgrave since you are also working on the thumbnails, could you maybe have a quick look at my changes to sign off on them? Thanks!

ulliholtgrave:
@timoludwig Yeah, this requires some changes for my implementation, but it's a useful approach. i guess @JoeyStk can merge it now :)

Issue #934, opened 2021-09-09
Page filter subpages not shown


david-venhoff:
### Describe the Bug
When filtering for pages which are not root-pages, these subpages will not be shown in the page tree view.

### Steps to Reproduce
1. Filter on the page view for any page which is not a root-page

### Expected Behavior
All matching pages are shown

### Actual Behavior
Only the root pages (and matching subpages of the root pages) get shown

PR #935, opened 2021-09-16
Include subpages in filtered page results

MizukiTemma:
### Short description

This PR is to fix the problem that subpages are not shown when the filter is applied.

### Proposed changes

- Subpages are now shown too
- Pages are shown even when their root page and/or parent page is filterd out.
- 

### Resolved issues

Fixes: #934

### DIFF
--- a/src/cms/templates/pages/page_tree.html
+++ b/src/cms/templates/pages/page_tree.html
@@ -86,6 +86,7 @@
             {% get_last_root_page pages as last_root_page %}
             {% recursetree pages %}
                 {% get_translation node language.slug as page_translation %}
+                {% get_depth_in node pages as page_depth %}
                 {% include "pages/page_tree_node.html" with page=node %}
                 {% if not node.is_leaf_node %}
                     {{children}}
--- a/src/cms/templates/pages/page_tree_node.html
+++ b/src/cms/templates/pages/page_tree_node.html
@@ -2,8 +2,8 @@
 {% load content_filters %}
 {% load page_filters %}
 {% load tree_filters %}
-<tr id="page-{{ page.id }}-drop-left" data-drop-id="{{ page.id }}" data-drop-position="left" class="drop {% if page.depth == 0 %} drop-between {% endif %} h-3 -m-3 hidden level-{{page.depth}}"><td colspan="9"><div><span></span></div></td></tr>
-<tr id="page-{{ page.id }}" data-drop-id="{{ page.id }}" data-drop-position="last-child" class="drop drop-on border-t border-b border-solid border-gray-200 hover:bg-gray-100 child {% if page.depth > 0 %} hidden level-{{page.depth}}{% endif %}">
+<tr id="page-{{ page.id }}-drop-left" data-drop-id="{{ page.id }}" data-drop-position="left" class="drop {% if page_depth == 0 %} drop-between {% endif %} h-3 -m-3 hidden level-{{page_depth}}"><td colspan="9"><div><span></span></div></td></tr>
+<tr id="page-{{ page.id }}" data-drop-id="{{ page.id }}" data-drop-position="last-child" class="drop drop-on border-t border-b border-solid border-gray-200 hover:bg-gray-100 child {% if page_depth > 0 %} hidden level-{{page_depth}}{% endif %}">
     <td class="pl-4">
         <input type="checkbox" name="selected_ids[]" value="{{ page.id }}" class="bulk-select-item">
     </td>
--- a/src/cms/templatetags/page_filters.py
+++ b/src/cms/templatetags/page_filters.py
@@ -19,3 +19,40 @@ def get_last_root_page(pages):
     """
     root_pages = list(filter(lambda p: not p.parent, pages))
     return root_pages[-1] if root_pages else None
+
+
+@register.simple_tag
+def get_depth_in(node, pageset):
+    """
+    This tag returns the depth of node whithin the tree/pages in pageset.
+
+    :param node : the page
+    :type node : ~cms.models.pages.page.Page
+
+    :param pageset: The pages (all pages or pages chosen by filter)
+    :type pageset: list [ ~cms.models.pages.page.Page ]
+
+    :return: the depth of node whithin the tree/pages in pageset
+    :rtype: int
+    """
+    if not node.parent in pageset:
+        return 0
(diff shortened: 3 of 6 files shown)


src/cms/templates/pages/page_tree.html
```diff
@@ -86,11 +86,14 @@ <h1 class="heading">{% trans 'Pages' %}</h1>
...
+                {% get_depth_in node pages as page_depth %}
+                {% is_shown_by_root node last_root_page pages as is_shown %}
                 {% include "pages/page_tree_node.html" with page=node %}
                 {% if not node.is_leaf_node %}
                     {{children}}
                 {% endif %}
-                {% if node == last_root_page %}
+                {% if node == last_root_page or not is_shown %}
                     <tr data-drop-id="{{ node.id }}" data-drop-position="right" class="drop drop-between h-3 hidden" title="test"><td colspan="9"><div><span></span></div></td></tr>
                 {% endif %}
```
melegiul, line 98: These lines have the purpose to add a drop region after the last root page.
As drag&drop is disabled, when the page filter feature is active, we won't need this template tag here for now.

src/cms/templates/pages/page_tree_node.html
```diff
@@ -23,7 +23,7 @@
         <span class="collapse-subpages cursor-pointer inline-block align-middle"
               title="{% trans 'Collapse all subpages' %}"
               data-page-id="{{ page.id }}"
-              data-page-children="{{ page|get_children }}">
+              data-page-children="{{ children_in_filter }}">
```
melegiul, line 26: This template tag is not required, because the `get_children` filter will only return the filtered children, too.
But in this case where we have a partial `TreeQuerySet` only, the `get_children` filter is buggy unfortunately.
So we should fix that over here: https://github.com/Integreat/integreat-cms/blob/38c0601eb4cc94f652fc40d1933adc530d74cab8/src/cms/templatetags/tree_filters.py#L35
This line should actually be:
```
return [child.id for child in node.get_children()]
```
Then the filter should work with partial trees, too.

melegiul (changes requested):
Thank you very much, I have tested your bugfix on chrome and firefox and everything works great
I could just find some minor issues, then this should be ready to merge :tada:

src/cms/templates/pages/page_tree.html
```diff
@@ -86,11 +86,12 @@ <h1 class="heading">{% trans 'Pages' %}</h1>
             {% get_last_root_page pages as last_root_page %}
             {% recursetree pages %}
                 {% get_translation node language.slug as page_translation %}
+                {% get_depth_in node pages as page_depth %}
                 {% include "pages/page_tree_node.html" with page=node %}
                 {% if not node.is_leaf_node %}
                     {{children}}
                 {% endif %}
-                {% if node == last_root_page %}
+                {% if node == last_root_pages %}
```
melegiul, line 94:
```suggestion
                {% if node == last_root_page %}
```
This was just a typo, right?

src/cms/templates/pages/page_tree_node.html
melegiul, line 26:
```suggestion
              data-page-children="{{ page|get_children }}">
```
As we fixed now the buggy filter, we can apply it now

src/cms/templatetags/page_filters.py
```diff
@@ -19,3 +19,79 @@ def get_last_root_page(pages):
...
+    :param node : the page
+    :type node : ~cms.models.pages.page.Page
+
+    :param pageset: The pages (all pages or pages chosen by filter)
+    :type pageset: list [ ~cms.models.pages.page.Page ]
+
+    :return: The list of IDs of children pages of node that are in pageset
+    :rtype: list [~cms.models.pages.page.Page]
+    """
+    return [child.id for child in node.children.all() if child in pageset]
```
melegiul, line 38:
```suggestion
```
Now we can remove this unused tag

src/cms/templatetags/page_filters.py
```diff
@@ -19,3 +19,79 @@ def get_last_root_page(pages):
...
+
+    :param pageset: The pages (all pages or pages chosen by filter)
+    :type pageset: list [ ~cms.models.pages.page.Page ]
+
+    :return: wheter the node is shown as descendants of root or another page in pageset
+    :rtype: bool
+    """
+    if root:
+        return node in root.get_descendants(include_self=True) and node in pageset
+    return False
```
melegiul, line 97:
```suggestion
```
This one can now removed safely too

src/cms/templatetags/page_filters.py
```diff
@@ -19,3 +19,79 @@ def get_last_root_page(pages):
...
+
+    :param pageset: The pages (all pages or pages chosen by filter)
+    :type pageset: list [ ~cms.models.pages.page.Page ]
+
+    :return: the depth of node whithin the tree/pages in pageset
+    :rtype: int
+    """
+    if not node.parent in pageset:
+        return 0
+    return node.depth - get_highest_anscentor_in(node, pageset).depth
```
melegiul, line 57: Really cool how you handled the depth calculation of partial trees :muscle: 
Per default only pages from level 0 (the "roots") are shown in the tree,
So we just needed this solution to decide, which pages are the "roots" in a partial tree

melegiul (changes requested):
Great, now beside my comments only thing left is doing a rebase and commit squashing would be nice too.
If you need help resolving the conflicts please let me know.

melegiul (approved):
Awesome :tada:

timoludwig:
Please wait with merging this PR until I squashed all commits, thanks!

timoludwig:
@MizukiTemma Thank for you PR, you can now merge it :tada:

For the future: Please use the command `git rebase develop` instead of `git merge develop` when you want to include the recent changes of the develop branch in your feature branch like described in [our documentation](https://integreat.github.io/integreat-cms/git-flow.html#update-feature-branch). This makes the git history a bit cleaner and makes it easier to squash the commits later on.
Also, please change the subjects of your commit messages so that they describe the solution to the problem, not the problem itself. E.g., If the problem is that the subpages are not shown, the commit that fixes that issue should be named "Include subpages in filtered page results" or similar.

timoludwig:
@MizukiTemma This exactly what I'm talking about, please don't merge the develop branch into your feature branch, but use rebasing instead :slightly_smiling_face: 
I will rebase this branch now again before merging.

Issue #822, opened 2021-05-12
Invalidate cache for static files after they changed


timoludwig:
### Motivation

Since browsers want to cache our bundled static files output (`main.js`, `editor.js`, `pdf.js` etc...), we have to find a way how to invalidate the cache after these files changed.

### Proposed Solution

- Add the content hash in the output filename as described [here](https://webpack.js.org/guides/caching/)
- Find a way how the updated content hash can end up in the html template, e.g. [build an html](https://webpack.js.org/guides/output-management/#setting-up-htmlwebpackplugin) file which can be included (I'm not sure about that, probably there is a better way)

### Alternatives

- Add a short server-side `Cache-Control` `max-age`
- Completely disable caching by adding the current timestamp after the filename, e.g. `<script src="{% static 'main.js' %}?v={% now 'U' %}"></script>`

PR #936, opened 2021-09-20
Adds contents hash to support browsers cache invalidation

melegiul:
### Short description

Invalidates the browser cache by adding content hashes to output file names of webpack

### Proposed changes

- Sets webpack output file name to [name].[contenthash].js
- To keep only the latest file hash, the webpack output.clean option is used 
- To use output.clean it is necessary to switch from `webpack serve` to `webpack watch` (see [861](https://github.com/webpack/webpack-dev-middleware/issues/861))
- To include the renamed webpack output into djangos templates `django-webpack-loader` is used

### Alternatives

Instead of `django-webpack-loader` we could also use [HTMLwebpackPlugin](https://webpack.js.org/guides/output-management/#setting-up-htmlwebpackpluginl) to include the webpack output files. but this approach seemed to have problems with djangos templates format (maybe I just used the wrong template loader, I'm not sure)

### Resolved issues

Fixes: #822

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -125,6 +125,7 @@ jobs:
           paths:
             - src/cms/static
             - src/cms/locale/de/LC_MESSAGES/django.mo
+            - webpack-stats.json
   packaging:
     docker:
       - image: integreat/bionic-setuptools:78c07dc
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -245,6 +245,13 @@
             "markers": "python_version >= '3.6'",
             "version": "==0.13.3"
         },
+        "django-webpack-loader": {
+            "hashes": [
+                "sha256:37b7b5bcc2f4c1dc63de503e7d3bd61859abd6c8ffaf885bb75975890606946e",
+                "sha256:80a9dd7266366b28078466e359685370a10a6247492662085acb4809cb831264"
+            ],
+            "version": "==1.3.0"
+        },
         "django-widget-tweaks": {
             "hashes": [
                 "sha256:9f91ca4217199b7671971d3c1f323a2bec71a0c27dec6260b3c006fa541bc489",
--- a/dev-tools/run.sh
+++ b/dev-tools/run.sh
@@ -36,7 +36,11 @@ echo -e "Starting WebPack dev server in background..." | print_info | print_pref
 deescalate_privileges npm run dev 2>&1 | print_prefix "webpack" 36 &
 
 # Waiting for initial WebPack dev build
-while [[ ! -f "${BASE_DIR}/src/cms/static/main.js" ]]; do sleep 1; done
+compgen -G "${BASE_DIR}/src/cms/static/main.*.js" > /dev/null
+while [[ $? -eq 1 ]]; do 
+    sleep 1
+    compgen -G "${BASE_DIR}/src/cms/static/main.*.js" > /dev/null
+done
 
 # Show success message once dev server is up
 listen_for_devserver &
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,7 @@ sphinx/templates/breadcrumbs.html*
 .directory
 cms-django.iml
 integreat-cms.iml
+webpack-stats.json
 
 # Media Library
 src/media/
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,6 +25,7 @@
         "tailwindcss": "^2.2.4",
         "tinymce": "^5.8.2",
         "tinymce-i18n": "20.12.25",
+        "webpack-bundle-tracker": "^1.3.0",
         "whatwg-fetch": "^3.6.2"
       },
       "devDependencies": {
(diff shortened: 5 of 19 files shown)


.circleci/config.yml
```diff
@@ -148,7 +148,7 @@ jobs:
           at: .
       - run:
           name: Generate documentation
-          command: pipenv run ./dev-tools/generate_documentation.sh
+          command: pipenv run ./dev-tools/generate_documentation.sh --clean
```
timoludwig, line 151: The documentation bulid in circleci is always clean, because the directory in which the intermediate sphinx files are stored is not passed to later builds. So no need to pass this parameter explicitly.

Pipfile.lock
```diff
@@ -1184,11 +1252,11 @@
...
             "hashes": [
-                "sha256:32bd3b5d13dc8186d7a42fc816a23d32e83a4827d7d9882948e7b837c232da5a",
-                "sha256:4a05bdbe8b1446d77a01e20a23ebc6777c74f43237035e76be89699308987d6f"
+                "sha256:4d35a56f4508cfee4c4fb604373ede6feae2a306731d533f409ef5c3496fdbd8",
+                "sha256:eec6d497e4c2195fa0e8b2016b337532b8a699a68bcb22a512870e16925c6a5c"
             ],
             "index": "pypi",
-            "version": "==0.5.2"
+            "version": "==1.0.0"
         },
```
timoludwig, line 1260: One quick fix for the documentation build would be to restore the old version here manually.
Or you could find out what in the footer termplate changed and fix the patch file...

melegiul: Thank you, that sounds good to me.
It was not even my intention to update so many parts of this lock file,
but when I added a dependency to setup.py and ran `pipenv install -e .` like our docu says,
I expected `django-webpack-loader` to show up in the pipfile.lock, but it was not listed in the lock file.

So I applied this patch: I ran `pipenv install -e .` followed by `pipenv lock` to move the newly installed `django-webpack-loader` into the pifile.lock, but this has the drawback, that other unnecessary entries in the pipfile.lock updated, too.

Do you know a better approach for the pipfile.lock sync?
Maybe we could try this tool [pipenv-setup](https://github.com/pypa/pipenv/issues/1263#issuecomment-524382945)?

melegiul: Yes, this would be a great solution and probably also fix the docu generation problem, but are you sure that `django-webpack-loader` is a dev-dependency? I first thought we need this django app also installed in production, because It provides template tags like `{% render_bundle some_webpack_entry_point %}` which it is basically used for creating the script tag for webpack build output (replacing the ¸<script src="{% static 'main.js' %}" >) and not part of the webpack build process.

So we need this in production too or am I wrong?

timoludwig: Ah okay sorry, I wasn't aware how the plugin works. But yes, should be a production dependency...
So it's probably best to revert the sphinx theme version manually...

melegiul: You are right, reverting the sphinx theme version worked great to fix the docu generation, thank you :tada: 
Just some pylint and test issues remain, I will have a look at them

Pipfile
```diff
@@ -15,7 +15,7 @@ pylint-runner = "*"
 shellcheck-py = "*"
 sphinx = "*"
 sphinx-last-updated-by-git = "*"
-sphinx-rtd-theme = "*"
+sphinx-rtd-theme = "~=0.5"
```
svenseeberg, line 18: You can rebase on current develop and revert this.

dev-tools/run.sh
```diff
@@ -36,7 +36,11 @@ echo -e "Starting WebPack dev server in background..." | print_info | print_pref
 deescalate_privileges npm run dev 2>&1 | print_prefix "webpack" 36 &
 
 # Waiting for initial WebPack dev build
-while [[ ! -f "${BASE_DIR}/src/cms/static/main.js" ]]; do sleep 1; done
+compgen -G "${BASE_DIR}/src/cms/static/main.*.js" > /dev/null
+while [[ $? -eq 1 ]]; do 
+    sleep 1
+    compgen -G "${BASE_DIR}/src/cms/static/main.*.js" > /dev/null
+done
```
svenseeberg, line 43: Is this loop intended to wait for a user fixing something? Or do we expect that the command simply fails sometimes and we need to try again?

melegiul: This loop is only entered, when webpack compilation is not finished at this moment i.e. the frontend part is ready to use
The previous version was just a bash test command, checking if webpack is ready: 
```
[[ ! -f "${BASE_DIR}/src/cms/static/main.js" ]]
```
But as far as I know the command is not appropriate for glob/regex patterns like main._some-hash_.js,
so I replaced it by this compgen, which serves the same purpose [https://stackoverflow.com/questions/6363441/check-if-a-file-exists-with-wildcard-in-shell-script](https://stackoverflow.com/questions/6363441/check-if-a-file-exists-with-wildcard-in-shell-script), i.e. compgen returns exit status 1 as long as the file does not exists

svenseeberg: Ah okay, waiting for `webpack`. Thanks, that helps me understand.

svenseeberg (approved):
Looks good to me! :+1:

Issue #937, opened 2021-09-21
Rename Help menu item


ulliholtgrave:
### Motivation
For a unified domain we are supposed to rename our "Help" Link to "Online Documentation".

### Proposed Solution
Rename 'Help' to 'Online Documentation'. Besides that is it requested to add a contour only Integreat Icon to show the connection.
Resulting format would be: [INTEGREAT-ICON] Online Documentation

PR #941, opened 2021-09-22
Add legacy file type support & Redesign Help Link (fixes #910 & #937)

### LINKED ISSUE #910: Support legacy file types in media library
### Motivation

To support a few legacy file types from WordPress, we need a few adaptions to the media library.

### Proposed Solution

Unsupported media types should be able to remain in the database and the file system. The media library should not have any displaying glitches or weird error messages when legacy files are opened, changed or selected, only the upload of new unsupported files should be forbidden.
The legacy file types are:
- `application/msword`
- `application/vnd.ms-excel`
- `application/vnd.ms-powerpoint`
- `application/vnd.openxmlformats-officedocument.presentationml.presentation`
- `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`
- `application/vnd.openxmlformats-officedocument.wordprocessingml.document`
- `image/gif`
- `image/svg+xml`

ulliholtgrave:
### Short description
This PR adds the legacy MIME type to our list of accepted files, but denies their upload. To provide an additional possibility to upload them, the PR adds a setting to enable uploads temporary.

### Proposed changes
- Extend possible media types with the ones found in Wordpress
- Split enabled media in accepted media types and up-loadable file types. 
- Renames Help menu item, according to the naming of the cities team

### Resolved issues

Fixes: #910 #937


src/cms/templates/_base.html
```diff
@@ -18,8 +18,8 @@
     {% endif %}
     <a href="{% url 'wiki_redirect' %}" target="_blank" rel="noopener noreferrer"
        class="relative px-2 pt-4 text-gray-800 hover:bg-gray-200 border-r border-gray-400">
-        <i data-feather="help-circle"></i>
-        {% trans 'Help' %}
+        <img class="inline-block" fill="none" width="21" height="21" src="{% static 'images/integreat-icon-contur-only.png' %}">
+        {% trans 'Wiki / Documentation' %}
```
svenseeberg, line 22: I would name it `Documentation (Wiki)`, because the important part here is the `Documentation` and only some may think of our Documentation as "The Wiki".

ulliholtgrave: I talked to Daniel about this and he came up with this wording. But as this will only take like 1 minute to change, I would suggest to leave it as Wiki / Documentation for now. Maybe he can take the discussion into the cities-team :)

svenseeberg (approved):
Looks good! I only have one "issue" with the mentioned documentation link, but this may or may not be changed.

src/cms/templates/_base.html
svenseeberg: Well, "Documentation" is the standing term for Documentation ;-) We're always discussing "intuition", this is one of these moments. Only a small minority of web users knows what the term wiki actually refers to.

ulliholtgrave: Yeah, but there seem to be a time, when we referred to our Documentation as the "Integreat Wiki". So we are at this point again where we have to value between Habit and Conformity :D

PR #948, opened 2021-09-24
Improve user related performance

svenseeberg:


svenseeberg:
```
Sep 24 17:53:21 DEBUG django.db.backends - (0.000) SELECT "cms_role"."id", "cms_role"."name", "cms_role"."group_id", "cms_role"."staff_role" FROM "cms_role" WHERE "cms_role"."group_id" = 1 LIMIT 21; args=(1,)
```

This query is executed multiple times on each page call. Probably permission checks? Can we cache this?

src/cms/views/roles/role_view.py
```diff
@@ -45,10 +44,13 @@ def get(self, request, *args, **kwargs):
...
-        role_instance = Role.objects.filter(id=kwargs.get("role_id")).first()
-        group_instance = Group.objects.filter(role=role_instance).first()
+        role_instance = (
+            Role.objects.filter(id=kwargs.get("role_id"))
+            .select_related("group")
+            .first()
+        )
         role_form = RoleForm(instance=role_instance)
-        group_form = GroupForm(instance=group_instance)
+        group_form = GroupForm(instance=getattr(role_instance, "group", None))
```
timoludwig, line 53: Why not just `role_instance.group`? If the group is not yet set, it's `None` anyway, right?

src/cms/views/roles/role_view.py
```diff
@@ -72,10 +74,15 @@ def post(self, request, *args, **kwargs):
...
-        group_instance = Group.objects.filter(role=role_instance).first()
+        role_instance = (
+            Role.objects.filter(id=kwargs.get("role_id"))
+            .select_related("group")
+            .first()
+        )
         role_form = RoleForm(data=request.POST, instance=role_instance)
-        group_form = GroupForm(data=request.POST, instance=group_instance)
+        group_form = GroupForm(
+            data=request.POST, instance=getattr(role_instance, "group", None)
```
timoludwig, line 84: here as well maybe?

timoludwig (commented):
Thanks! :+1:
I just reviewed commits cb2340320f5294767b721ccaa1f8e3c6f09327cb and 82a8fa8db32b1f44c4326c377466cc207d3d763a which look good to me.

src/cms/views/roles/role_view.py
svenseeberg: Been there, done that ;-) `role_instance` itself can be `None` and then it will throw an error that `None` has no attribute `group`.

timoludwig: Ah yeah, sorry :facepalm:

melegiul (approved):
Awesome, select_related is a great choice when it comes to looking up foreign key relations :rocket:
Maybe we could use a similar approach for the user groups relationship?
But that could be the next iteration, if at all needed

timoludwig (approved):
Prefetching the groups and role objects reduced the amount of queries even further: Instead of doing two additional queries for each user to retrieve the groups and the role of the user, this does now two queries in the beginning to get the groups and roles for all users at once, so a constant number of queries independently from the number of users in the database. :tada:

Issue #874, opened 2021-06-29
Add possibility to replace media file


timoludwig:
### Motivation

At the moment, when a file gets deleted and then a new version uploaded, the file paths are not identical. This means that all links in the content are outdated.

### Proposed Solution

We need the possibility to replace a media file so its physical path on the hard drive stays unchanged.

### Additional Context

Probably this can be solved by adding the `file` field to the [MediaFileForm](https://github.com/Integreat/integreat-cms/blob/develop/src/cms/forms/media/media_file_form.py) and making sure that any uploaded file gets saved to the same location the existing file was stored. This should only apply to files which are uploaded directly via the file editing sidebar, not the original upload form.

This could look e.g. like this:
![Screenshot 2021-06-29 at 17-54-51 Integreat Editorial System](https://user-images.githubusercontent.com/16021269/123829909-20999f00-d903-11eb-9b83-f556a2177ea8.png)

timoludwig:
Duplicate of #69

PR #949, opened 2021-09-25
Add Media File Replacement 

membralala:
### Short description

This PR is not ready to merge! It should handle media file replacement, however there are still some issues, so I'm asking for advice. 

### Proposed changes

I implemented the UI logic for a replacement button in Preact and managed to change the file somehow expanding the `media_actions.edit_file_ajax` function. However, it still does not the main trick, as I struggle to manage a proper communication between the file_form and the database. Is there a possibility to link the new file to the old path? Or would it be better to delete the old file and database entry and create a new one to the old path? 

### Resolved issues

Should fix: #874


src/cms/forms/media/media_file_form.py
```diff
@@ -23,4 +24,23 @@ class Meta:
...
+        :return: The cleaned form data
+        :rtype: dict
+        """
+        
+        cleaned_data = super().clean()
+        file = cleaned_data.get("file")
+        if file:
+            cleaned_data["thumbnail"] = generate_thumbnail(file, 300, 300, False)
+
+        return cleaned_data
```
timoludwig, line 46: This is the right idea - however, we need a bit more complex validation and save methods for replacement, therefore I suggest to use a whole new form class for this task. So the `MediaFileForm` should be unchanged, but you can create a new `ReplaceMediaFileForm` with copy + paste from `UploadMediaFileForm`.

At first, we need to specify the fields for this form - since the `parent_directory` and `type` should be identical, we can remove it from the `fields` list.

Then, in the `clean()`-method, we have to do the following:
- Does the actual file type of the new file (checked with `magic.from_buffer()`) match the original file type? If not, throw a `ValidationError`. We only want to allow the replacement with new files of the same type.
- Set `cleaned_data["name"] = file.name` to make sure the new name in the database is the name of the new file
- Set `file.name = self.instance.file.name` to make sure the new name on the hard drive is the name of the old file
- If everything goes well and no errors occur, generate a thumbnail with the new file
- Set the thumbnail filename to the existing thumbnail filename of the form instance

Then, in the save-method, we have to do the following:
- Delete the old file with `self.instance.file.delete()` (otherwise Django would automatically append a random string to the new file's name because the name is not available)
- Delete the old thumbail
- Call `super().save()` to save the new files

One pitfall we have to keep in mind: the [upload_path()](https://github.com/Integreat/integreat-cms/blob/3632dd978cac673809cb72077aee1dc80c4c774b/src/cms/models/media/media_file.py#L24-L45)-method will return the directory of the current date instead of the initial file's date, so we have to find a way around this. However, I think it will be easier to tackle this problem when everything else is working (so we can try out a few different things), so let's not worry about that now :wink:

src/cms/views/media/media_actions.py
```diff
@@ -175,8 +177,13 @@ def edit_file_ajax(request, region_slug=None):
...
     )
-
-    media_file_form = MediaFileForm(data=request.POST, instance=media_file)
+    
+    if request.FILES:
+        # date = media_file.uploaded_date.strftime("%d/%m/%y")
+        media_file_form = MediaFileForm(data=request.POST, instance=media_file, files=request.FILES)
+        logger.warn(media_file_form.__dict__)
+    else:
+        media_file_form = MediaFileForm(data=request.POST, instance=media_file)
```
timoludwig, line 186: I would suggest to create a new ajax endpoint just for handling the replacement. Also, you don't need the differentiation with `if request.FILES:`, because `Form(...)` and `Form(..., files=None)` are equivalent (if you have a look at the [source code](https://github.com/django/django/blob/93f3b7a0ba2969109940775a18515728603afac6/django/forms/forms.py#L72), you see that `files` is a keyword argument with a default value of `None`).

src/frontend/js/media-management/component/edit-sidebar.tsx
```diff
@@ -253,6 +254,40 @@ export default function EditSidebar({
...
+                    type="file"
+                    maxLength={255}
+                    style={{display: "none"}}
+                    disabled={isLoading}
+                    onChange={
+                      () => {
+                        document.getElementById("submit-replace-file").click();
+                      }
+                    }
+                  />
```
timoludwig, line 288: I would add `accept={selectedFile.type}` here to make it harder for the user to select invalid files with a different file type than the original file.

src/frontend/js/media-management/component/edit-sidebar.tsx
```diff
@@ -253,6 +254,40 @@ export default function EditSidebar({
                       {mediaTranslations.btn_save_file}
                     </button>
                   )}
+     
+                  <button
```
timoludwig, line 258: I think it makes sense to define this as `<label for="replace-file-input" ...>` instead of a button with a custom onclick property.

src/frontend/js/media-management/component/edit-sidebar.tsx
```diff
@@ -253,6 +254,40 @@ export default function EditSidebar({
                       {mediaTranslations.btn_save_file}
                     </button>
                   )}
+
```
timoludwig, line 257: Here, you would have to add a new `<form>`-tag for the new `ReplaceMediaFileForm`.

timoludwig (commented):
Thanks for you PR! :+1:

In general, it has the right ideas. I posted some feedback below to point you in the next direction. I think, the first approach should be to try to keep the same database entry and just replace the file on the hard drive. I'm not completely sure about all the implications, but at least that's how I would try it. Could also be that I forgot something, so if you encounter any problems along the way, don't hesitate to ask.

timoludwig:
Oh, and apart from that: Your branch is a bit out-of-date, and since there were a few changes to the media library, I suggest you [rebase your branch](https://integreat.github.io/integreat-cms/git-flow.html#update-feature-branch) onto the current develop-state before proceeding. If you're unsure, I can also do this for you (or together during pair-programming).

src/cms/forms/media/media_file_form.py
membralala: Thanks for your advice! Most of your ideas seem to work flawless. However, the `self.instance.file.delete()` with in the save-method seems not to work and I couldn't really figure out why. Instead I resolved it by "manually" deleting the file with `os.remove()` and getting the correct path from the `models.media.media_file.upload_path()`-method.

src/cms/views/media/media_actions.py
```diff
@@ -216,6 +219,71 @@ def edit_file_ajax(request, region_slug=None):
...
+                        "text": _("No changes detected"),
+                    }
+                ],
+                "file": media_file.serialize(),
+            }
+        )
+
+    # Save form
+    media_file_form.save()
+
```
timoludwig, line 272:
```suggestion
    logger.info("%r was replaced by %r", media_file, request.user)

```
Maybe add a quick notice that a file was replaced, this might help debug problems on the production environment later on.

src/frontend/js/media-management/component/edit-sidebar.tsx
```diff
@@ -64,7 +65,7 @@ export default function EditSidebar({
   // The current directory is the last element of the directory path
   const file = directoryContent[fileIndex] as File;
   // This state is a buffer for the currently changed file
-  const [changedFile, setChangedFile] = useState<File>(file);
+  const [changedFile, setChangedFile] = useState<File | null>(file);
```
timoludwig, line 68:
```suggestion
  const [changedFile, setChangedFile] = useState<File>(file);
```
I don't think this can be `null` - since the edit sidebar is only rendered when a file is selected, we can be sure that there is always a file object in this buffer. Or am I missing something? Was there a specific reason you changed the line?

src/frontend/js/media-management/component/edit-sidebar.tsx
```diff
@@ -277,6 +278,22 @@ export default function EditSidebar({
                       {mediaTranslations.btn_save_file}
                     </button>
                   )}
+
+                  <label
+                    for="replace-file-input"
+                    // title={mediaTranslations.btn_replace_file}
```
timoludwig, line 284:
```suggestion
                    title={mediaTranslations.btn_replace_file} 
```
I think we can very well use this title here.

src/frontend/js/media-management/component/edit-sidebar.tsx
```diff
@@ -277,6 +278,22 @@ export default function EditSidebar({
...
+                      "w-full text-white text-center font-bold py-2 px-4 mb-4 rounded",
+                      { "cursor-not-allowed bg-gray-500": isLoading },
+                      { "bg-blue-500 hover:bg-blue-600": !isLoading }
+                    )}
+                    disabled={isLoading} 
+                  >
+                    <RefreshCw class="mr-1 inline-block h-5" />
+                    {mediaTranslations.btn_replace_file}
+                  </label>
+                  {/* <button id="submit-replace-file" class="hidden" type="submit"> </button> */}
```
timoludwig, line 295:
```suggestion
```
This isn't used anymore, right?

src/frontend/js/media-management/component/edit-sidebar.tsx
```diff
@@ -305,6 +322,31 @@ export default function EditSidebar({
...
+        action={apiEndpoints.replaceFile}
+        class="hidden"
+      >
+        <input name="id" type="hidden" value={file.id} />
+        <input
+          id="replace-file-input"
+          name="file"
+          type="file"
+          maxLength={255}
+          accept={changedFile.type}
```
timoludwig, line 338:
```suggestion
          accept={file.type}
```
`changedFile` is a "special" buffer for changes to the name or alt tags, so I think it's more canonical to use just `file` here.

src/frontend/js/media-management/component/edit-sidebar.tsx
```diff
@@ -305,6 +322,31 @@ export default function EditSidebar({
...
+          style={{display: "none"}}
```
timoludwig, line 339:
```suggestion
```
In general: Please use `class="hidden"` instead of `style={{display: "none"}}`. In this case however, since the form isn't shown anyway, it's not required at all.

timoludwig (commented):
Thanks a lot! :+1: :tada:

This PR already works pretty well and solves most parts of the issue. :muscle: 
See my comments below for my feedback.

In addition, I have two general requests:
1. I noticed that the thumbnails of a replaced image are not immediately updated, because the browser still has a copy of the image in it's cache and doesn't know it has been updated on the server side. Thus, I'd suggest to add
    ```
    thumbnail_url = self.thumbnail_url
    if thumbnail_url:
        # Make sure replaced images are not cached
        try:
            thumbnail_url += f"?{os.path.getmtime(self.file.path)}"
        except OSError as e:
            logger.error(
                "The physical file of %r could not be accessed: %r", self, e
            )
    ```
    to the [serialize()](https://github.com/Integreat/integreat-cms/blob/eba04af7d19f6015e224439952a136ee7808a7ff/src/cms/models/media/media_file.py#L148)-method to add a last-modified-timestamp to the thumbnail url.

2. I noticed some problems with the logging methods in case something goes wrong and the media file doesn't have a physical file etc... so I suggest to change the [\_\_repr__()](https://github.com/Integreat/integreat-cms/blob/eba04af7d19f6015e224439952a136ee7808a7ff/src/cms/models/media/media_file.py#L177)-method as follows:
    ```
    file_path = f"path: {self.file.path}, " if self.file else ""
    return f"<MediaFile (id: {self.id}, name: {self.name}, {file_path}{region})>"
    ```
    I hope this is a little more fail-safe.

PR #954, opened 2021-09-27
Fix `compgen` usage in `run.sh`

timoludwig:
### Short description

`compgen` does not exit with code 1 when no matches are found on all distros, therefore we should inspect the standard output rather than the exit code.

### Proposed changes

- Check whether output is empty instead of whether exit code is 1


melegiul (approved):
Thank you very much :+1:

PR #956, opened 2021-09-27
Update django-webpack-loader from 1.3.0 to 1.4.0

timoludwig:
### Short description

This update seems to introduce some problems, so we have to figure out how to fix the tests.
I assume it has something to do with https://github.com/django-webpack/django-webpack-loader/pull/297.

### Proposed changes

- Update django-webpack-loader from 1.3.0 to 1.4.0

### DIFF
--- a/src/cms/utils/pdf_utils.py
+++ b/src/cms/utils/pdf_utils.py
@@ -4,7 +4,7 @@ import logging
 from django.contrib.staticfiles import finders
 from django.core.cache import caches
 from django.db.models import Min
-from django.http import HttpResponse
+from django.http import HttpRequest, HttpResponse
 from django.template.loader import get_template
 from django.utils.translation import ugettext as _
 from django.views.decorators.cache import never_cache
@@ -90,6 +90,7 @@ def generate_pdf(region, language_slug, pages):
         "language": language,
         "amount_pages": amount_pages,
         "prevent_italics": ["ar", "fa"],
+        "request": HttpRequest(),
     }
     response = HttpResponse(content_type="application/pdf")
     response["Content-Disposition"] = f'filename="{filename}"'
--- a/src/cms/views/error_handler/error_handler.py
+++ b/src/cms/views/error_handler/error_handler.py
@@ -25,7 +25,6 @@ def render_error_template(context):
     return render_to_string("error_handler/http_error.html", context)
 
 
-# pylint: disable=unused-argument
 def handler400(request, exception):
     """
     Render a HTTP 400 Error code
@@ -40,6 +39,7 @@ def handler400(request, exception):
     :rtype: ~django.template.response.TemplateResponse
     """
     context = {
+        "request": request,
         "code": 400,
         "title": _("Bad request"),
         "message": _("There was an error in your request."),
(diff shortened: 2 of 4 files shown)


melegiul:
The new webpack loader version assumes that the key "request" is set in the templates context. As far as I know this happens per default in most circumstances, but in case of our pdf export we creating a custom context which does not include the request key.
So one way to fix this would be to add a request object to the pdf context, which is passed by the view over here: https://github.com/Integreat/integreat-cms/blob/3632dd978cac673809cb72077aee1dc80c4c774b/src/cms/utils/pdf_utils.py#L85
If you agree, I could also push a commit for this?

timoludwig:
> The new webpack loader version assumes that the key "request" is set in the templates context. As far as I know this happens per default in most circumstances, but in case of our pdf export we creating a custom context which does not include the request key.
> So one way to fix this would be to add a request object to the pdf context, which is passed by the view over here:

Ah okay. Do you think it's sufficient to add a dummy request object with `context["request"] = object()`? Then we don't have to refactor as much...
 
> If you agree, I could also push a commit for this?

Sure, go ahead!

timoludwig:
This also affects out error templates which are rendered via [render_to_string()](https://docs.djangoproject.com/en/3.2/topics/templates/#django.template.loader.render_to_string) without the request context.

melegiul:
Yes, dummy request is possible too, but object() does not allow setting attributes, which is required, so I used type() instead

src/cms/views/error_handler/error_handler.py
```diff
@@ -22,6 +22,15 @@ def render_error_template(context):
...
     """
+    # django-webpack-loader assumes request object in context
+    # and that this request object has a specific attribute set
+    context.update(
+        {
+            "request": type(
+                "dummy_request", (object,), {"_webpack_loader_used_tags": set()}
+            )()
+        }
+    )
```
timoludwig, line 33: I think in this case we should just use the real request of the `handler400()`, `handler403()` etc...

src/cms/utils/pdf_utils.py
```diff
@@ -82,13 +82,19 @@ def generate_pdf(region, language_slug, pages):
             title = region.name
     language = Language.objects.get(slug=language_slug)
     filename = f"Integreat - {language.translated_name} - {title}.pdf"
+    # django-webpack-loader assumes request object in context
+    # and that this request object has a specific attribute set
+    dummy_request = type(
+        "dummy_request", (object,), {"_webpack_loader_used_tags": set()}
+    )()
```
timoludwig, line 89: `dummy_request = type("", (), {})()` should be enough, right? The attribute `_webpack_loader_used_tags` is set by webpack loader if not existent on the request object anyway...
Or, maybe even better: do an `from django.http import HttpRequest` and use `HttpRequest()` as dummy request? maybe a bit more readable...

melegiul: Ok, I took the real request for error templates and a dummy HttpRequest object for the PDF export

timoludwig:
@melegiul Thanks a lot! I just resolved the conflicts and added the request object to the context directly instead of passing it to the function and then updating the context.
Can you have a look at it again and approve it if you're ok with the changes?

melegiul (approved):
Thank you, looks great now :+1:

PR #960, opened 2021-09-29
Fix permission checks

timoludwig:
### Short description

I noticed a few smaller problems with our permission checks:

Sometimes, the permissions of the `auth` app are checked for permissions of the `cms` app and vice versa.
This is not a security concern, the worst case scenario is that user didn't have access to certain functions which they should have had (since permission checks for non-existent permissions always return `False`).
Also, I improved a few permissions for the `User` model and the media models.


ulliholtgrave (approved):
Looks good. I created some users with the rolls and I didn't find any bad behavior :)

PR #976, opened 2021-10-14
Set webpack stats file path in settings

svenseeberg:
Set the path to the webpack-stats.json file in the settings.py. Does this make sense @melegiul ? It seems we use a different path, but basically this is almost the same as the example in https://github.com/django-webpack/django-webpack-loader#default-configuration.


timoludwig:
I tried to address this in #983:
https://github.com/Integreat/integreat-cms/blob/6ab2f3a9d1084d903ff8776aa2ea6a4eafff9b8a/webpack.config.js#L119

PR #981, opened 2021-10-16
Replace console.log with debug or error

michael-markl:


maxammann:
By @Schedulaar 's hand

Issue #829, opened 2021-05-13
Make pylint docstring config more strict


timoludwig:
### Motivation

At the moment, we use the default configuration for the plugin [pylint.extensions.docparams](http://pylint.pycqa.org/en/latest/technical_reference/extensions.html#parameter-documentation-checker).
The [default options](http://pylint.pycqa.org/en/latest/technical_reference/extensions.html#parameter-documentation-checker-options) are:

- `accept-no-param-doc`
    Whether to accept totally missing parameter documentation in the docstring of a function that has parameters.
    Default: `yes`
- `accept-no-raise-doc`
    Whether to accept totally missing raises documentation in the docstring of a function that raises an exception.
    Default: `yes`
- `accept-no-return-doc`
    Whether to accept totally missing return documentation in the docstring of a function that returns a statement.
    Default: `yes`
- `accept-no-yields-doc`
    Whether to accept totally missing yields documentation in the docstring of a generator.
    Default: `yes`

### Proposed Solution

I think we should set all those options to `no` to assure all parameters and return values are documented without exceptions.
Setting this options will probably cause a lot of warnings which have to be fixed.

PR #984, opened 2021-10-20
Make pylint docstring config more strict

JoeyStk:
### Short description
Make pylint docstring config more strict

### Proposed changes

- Add params & type to documentation
- Add yield to documentation
- Add return & rtype to documentation

### Resolved issues

Fixes: #829

### DIFF
--- a/src/cms/models/languages/language_tree_node.py
+++ b/src/cms/models/languages/language_tree_node.py
@@ -1,5 +1,5 @@
 from mptt.fields import TreeForeignKey
-from mptt.models import MPTTModel, raise_if_unsaved
+from mptt.models import MPTTModel
 
 from django.db import models
 from django.utils import timezone
@@ -114,48 +114,6 @@ class LanguageTreeNode(MPTTModel):
         :rtype: str
         """
         return len(self.get_ancestors())
-
-    # Explicitly define functions to show documentation of base model
-    @raise_if_unsaved
-    def get_ancestors(self, ascending=False, include_self=False):
-        return super().get_ancestors(ascending, include_self)
-
-    # pylint: disable=useless-super-delegation
-    @raise_if_unsaved
-    def get_family(self):
-        return super().get_family()
-
-    @raise_if_unsaved
-    def get_children(self):
-        return super().get_children()
-
-    @raise_if_unsaved
-    def get_descendants(self, include_self=False):
-        return super().get_descendants(include_self)
-
-    def get_descendant_count(self):
-        return super().get_descendant_count()
-
-    @raise_if_unsaved
-    def get_root(self):
-        return super().get_root()
-
-    # pylint: disable=useless-super-delegation
-    def insert_at(
-        self,
-        target,
-        position="first-child",
-        save=False,
-        allow_existing_pk=False,
-        refresh_target=True,
-    ):
-        return super().insert_at(
-            target, position, save, allow_existing_pk, refresh_target
-        )
-
-    # pylint: disable=useless-super-delegation
-    def move_to(self, target, position="first-child"):
-        return super().move_to(target, position)
 
     def __str__(self):
         """
--- a/src/cms/models/pages/page.py
+++ b/src/cms/models/pages/page.py
@@ -168,16 +168,49 @@ class Page(MPTTModel, AbstractBasePage):
         return Page.objects.filter(region__slug=region_slug, parent=None)
 
     def get_previous_sibling(self, *filter_args, **filter_kwargs):
+        """
+        This function gets all previous siblings of a page
+
+        :param filter_args: The supplied arguments
+        :type filter_args: list
+
+        :param filter_kwargs: The supplied kwargs
+        :type filter_kwargs: list
+
+        :return: The previous sibling
+        :rtype: ~cms.models.pages.page.Page
+        """
         # Only consider siblings from this region
         filter_kwargs["region"] = self.region
         return super().get_previous_sibling(*filter_args, **filter_kwargs)
 
     def get_next_sibling(self, *filter_args, **filter_kwargs):
+        """
+        This function gets the next sibling of a page
+
+        :param filter_args: The supplied arguments
+        :type filter_args: list
+
+        :param filter_kwargs: The supplied kwargs
+        :type filter_kwargs: list
+
+        :return: The next sibling
+        :rtype: ~cms.models.pages.page.Page
+        """
         # Only consider siblings from this region
         filter_kwargs["region"] = self.region
         return super().get_next_sibling(*filter_args, **filter_kwargs)
 
     def get_siblings(self, include_self=False):
+        """
+        This function gets all siblings of a page
+
+        :param include_self: gives state of include_self
+        :type include_self: bool
+
+        :return: All siblings of this page
+        :rtype: ~django.db.models.query.QuerySet [ ~cms.models.pages.page.Page ]
+        """
         # Return only siblings from the same region
         return (
             super().get_siblings(include_self=include_self).filter(region=self.region)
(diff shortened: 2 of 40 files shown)


src/cms/views/authentication/account_activation_view.py
```diff
@@ -71,6 +71,13 @@ def dispatch(self, *args, **kwargs):
     def form_valid(self, form):
         """
         If the form is valid, show a success message.
+
+        :param form: The supplied form
+        :type form: ~django.contrib.auth.forms.AuthenticationForm
```
timoludwig, line 76:
```suggestion
        :type form: ~django.contrib.auth.forms.SetPasswordForm
```

src/cms/views/authentication/account_activation_view.py
```diff
@@ -71,6 +71,13 @@ def dispatch(self, *args, **kwargs):
...
+        
+        :return: form validation
+        :rtype: HttpResponse
```
timoludwig, line 79:
```suggestion
        :rtype: ~django.http.HttpResponseRedirect
```

src/cms/views/authentication/account_activation_view.py
```diff
@@ -71,6 +71,13 @@ def dispatch(self, *args, **kwargs):
     def form_valid(self, form):
         """
         If the form is valid, show a success message.
+
+        :param form: The supplied form
+        :type form: ~django.contrib.auth.forms.AuthenticationForm
+        
+        :return: form validation
```
timoludwig, line 78:
```suggestion
        :return: A redirection to the ``success_url``
```

src/cms/views/authentication/password_reset_confirm_view.py
```diff
@@ -61,6 +61,13 @@ def dispatch(self, *args, **kwargs):
     def form_valid(self, form):
         """
         If the form is valid, show a success message.
+        
+        :param form: The supplied form
+        :type form: ~django.contrib.auth.forms.AuthenticationForm
```
timoludwig, line 66:
```suggestion
        :type form: ~django.contrib.auth.forms.SetPasswordForm
```

src/cms/views/authentication/password_reset_view.py
```diff
@@ -47,6 +47,14 @@ def dispatch(self, *args, **kwargs):
 
     def form_valid(self, form):
         """
+        This function validates the form and sends a message depending if the function was executed successfully
+        
+        :param form: The supplied form
+        :type form: ~django.contrib.auth.forms.AuthenticationForm
```
timoludwig, line 53:
```suggestion
        :type form: ~django.contrib.auth.forms.SetPasswordForm
```

src/cms/views/chat/chat_actions.py
```diff
@@ -20,6 +20,9 @@ def send_chat_message(request):
     """
     Send chat message
 
+    :param request: The current request
+    :type request: ~django.http.HttpResponse
```
timoludwig, line 24:
```suggestion
    :type request: ~django.http.HttpRequest
```
request and response are two different things :wink:

src/cms/views/chat/chat_actions.py
```diff
@@ -54,6 +57,12 @@ def delete_chat_message(request, message_id):
     """
     Delete chat message
 
+    :param request: The current request
+    :type request: ~django.http.HttpResponse
```
timoludwig, line 61:
```suggestion
    :type request: ~django.http.HttpRequest
```

src/cms/views/chat/chat_actions.py
```diff
@@ -54,6 +57,12 @@ def delete_chat_message(request, message_id):
...
+
+    :param message_id: The id of the message
+    :type message_id: int
```
timoludwig, line 64:
```suggestion
    :type message_id: ~django.http.JsonResponse
```

src/cms/views/pages/page_tree_view.py
```diff
@@ -122,6 +122,15 @@ def get(self, request, *args, **kwargs):
...
 
                     def page_filter(page):
+                        """
+                        Filters page
+
+                        :param page: The according page
+                        :type page: list
+
+                        :return: translation status
+                        :rtype: translation_status
```
timoludwig, line 132:
```suggestion
                        :rtype: bool
```

src/cms/views/pages/page_tree_view.py
```diff
@@ -123,6 +123,15 @@ def get(self, request, *args, **kwargs):
                     enable_drag_and_drop = False
 
                     def page_filter(page):
+                        """
+                        Filters page
+
+                        :param page: The according page
+                        :type page: list
+
+                        :return: translation status
```
timoludwig, line 132:
```suggestion
                        :return: Whether or not the page should be filtered based on its translation status
```

src/cms/models/pages/page.py
```diff
@@ -168,16 +168,50 @@ def get_root_pages(cls, region_slug):
...
         filter_kwargs["region"] = self.region
         return super().get_next_sibling(*filter_args, **filter_kwargs)
 
     def get_siblings(self, include_self=False):
+        """
+        This function gets all siblings of a page
+
+        :param include_self: gives state of include_self
+        :type include_self: bool
+
```
timoludwig, line 210:
```suggestion
```

src/cms/models/languages/language_tree_node.py
```diff
@@ -118,26 +118,72 @@ def depth(self):
...
     def get_root(self):
+        """
+        This function gets the root of the language tree
+
+        :return: The root of this language tree
+        :rtype: ~cms.models.languages.language_tree_node.LanguageTreeNode
+        """
         return super().get_root()
 
     # pylint: disable=useless-super-delegation
```
timoludwig, line 189:
```suggestion
```
Initially I added these methods so the docstrings of the parent class will be shown in this model's documentation. But this is no longer the case when we overwrite the docstrings, so I propose to delete them altogether. Since the docstrings of the parent class is still visible if somebody opens the "extended view" (in the top right corner), I think this is fine.

src/cms/models/languages/language_tree_node.py
```diff
@@ -149,12 +195,46 @@ def insert_at(
...
+        :param target: the current target
+        :type target: str
+
+        :param position: the current position
+        :type position: str
+
+        :return: The moved node
+        :rtype: ~cms.models.languages.language_tree_node.LanguageTreeNode
+        """
         return super().move_to(target, position)
```
timoludwig, line 239:
```suggestion
```
Same justification like above, and please also remove the lines
```
def insert_at(
        self,
        target,
        position="first-child",
        save=False,
```
which I cannot suggest changes for. Thanks!

timoludwig (commented):
Thanks for your changes! I still found some room for improvement, but I think this is the final iteration :smile: 

@melegiul could you have a look over my suggestions and apply them if you think they make sense? Thanks a lot!

src/cms/models/languages/language_tree_node.py
melegiul: this is done in 7f6b735a48fe6df23b9849e56a232aee64f6e5ef
but i still have to fix a commit message problem...

timoludwig: What problem do you mean?

melegiul:
@timoludwig Much better now, Thank you for your review

src/cms/models/languages/language_tree_node.py
melegiul: I already fixed it, it was about the co-authored statement in commit messages

timoludwig (commented):
Amazing, thanks! :+1:

Finally, I don't find anything to improve anymore :laughing: 
Could you squash the commits into one before merging?

timoludwig (approved):
:tada:

Issue #965, opened 2021-10-11
Links to Webapp not working for `Page`/`Event`/`POI` lists


timoludwig:
### Describe the Bug

At the moment, we only have a CMS-preview for pages. Since the preview still looks a bit different than the Webapp, we should also add links to the content objects in the webapp for all content types:

- [x] `Page` list
- [x] `Event` list
- [x] `POI` list

### Steps to Reproduce

1. Go to e.g. event list
2. Click on the "eye" icon
3. Nothing happens

### Expected Behavior

The event should be opened in the webapp

### Actual Behavior

Nothing happens

### Additional Information

- Also think about what happens if an object doesn't have a public version - probably the "preview" button should just be hidden (except for pages where we can use the CMS-preview for drafts)
- What to do for pages? How do we differentiate between the CMS-preview and the link to the webapp?

PR #985, opened 2021-10-22
Implement links to webapp for page, event and POI lists

membralala:
### Short description

This PR implements an external-link option for the page tree as well as the event and POI lists. 

### Proposed changes

- Implemented the link option as an addition to the Preview mode (pages only).
- Changed link icons in all cases from "eye" to "external link" to prevent ambiguity with the page tree view.
- Link icons are disabled, if the the page/event/POI is not published.

### Resolved issues

Fixes: #965


timoludwig (approved):
Thanks :tada: :muscle:

Issue #997, opened 2021-10-30
Redirect after password change wrong


maxammann:
### Describe the Bug
The redirect after pw change is wrong.

### Steps to Reproduce

1. Go to initial page after logoin
2. Click on User Settings
3. Change password
4. you are redirected to https://malte-test.tuerantuer.org/user_settings/ instead of https://malte-test.tuerantuer.org/testumgebung/user_settings/

### Expected Behavior
see above
### Actual Behavior
see above

PR #998, opened 2021-10-30
Fix redirect to region user settings

timoludwig:
### Short description

This fixes the redirect after a region user changed his email or password to make sure the menu sidebar shows the region menu instead of the network management area.

### Proposed changes

- Redirect to region user settings after form submission

### Resolved issues

Fixes: #997


ulliholtgrave (approved):
Fixes the problem :+1:

Issue #989, opened 2021-10-26
Page specific permissions are not working anymore


timoludwig:
### Describe the Bug

I just noticed that our page specific permissions are not working anymore.

### Steps to Reproduce

1. Activate page specific permissions in the region form
2. Create a user who doesn't have the permission to edit pages (e.g. event manager)
3. Go to any page form and try to set this user as editor/publisher
4. Nothing happens

### Expected Behavior

The user should be shown in the sidebar

### Actual Behavior

Nothing happens and in the javascript console the error appears:
```
Uncaught (in promise) TypeError: button is null
    _callee$ page-permissions.ts:28
    tryCatch runtime.js:63
    invoke runtime.js:293
    defineIteratorMethods runtime.js:118
    __awaiter main.7ee7f149a66cb0c21348.js:4686
    __awaiter main.7ee7f149a66cb0c21348.js:4665
    grantPagePermission main.7ee7f149a66cb0c21348.js:4727
    setPagePermissionEventListeners page-permissions.ts:14

```

PR #999, opened 2021-10-30
Fix specific user permissions bug 

membralala:
### Proposed changes

- change the way, page forms are initialized in the page_actions.py 

### Resolved issues

Fixes: #989


timoludwig (approved):
Thanks a lot, this fixes the issue! :+1: :tada:

PR #1002, opened 2021-10-30
Fix colored output in bash scripts

timoludwig:
### Short description

This increases the portability and makes the coloured output work on more devices

### Proposed changes

- Use `\x1b` instead of `\e` as escape sequence in bash scripts

### Resolved issues

Fixes the color problems e.g. on mac os


svenseeberg (approved):
I see no problems with this.

Issue #991, opened 2021-10-26
Improve Broken Link Checker


timoludwig:
### Motivation

The broken link checker can detect broken links, but still has potential for improvement.

### Proposed Solution

- [ ] Figure out what to do with old revisions - either make it obvious that a link does not belong to a public version in the table by showing the version and linking to it in the version history, or completely hide outdated links
- [ ] Add links to the source, so broken links can easily be fixed
- [ ] Fix the double arrow in the select field and the padding to the button

PR #1007, opened 2021-10-30
Improve link checker

david-venhoff:
### Short description
This pr adds the improvements specified by #991 to the link checker.

### Proposed changes
- Hide links from object translations which are outdated
- Add links to the source
- Fix some css issues

### Resolved issues
Fixes: #991

### DIFF
--- a/src/cms/utils/filter_links.py
+++ b/src/cms/utils/filter_links.py
@@ -1,45 +1,78 @@
-from django.db.models import Q
-
-from linkcheck.models import Link
+from cms.models import PageTranslation, EventTranslation, POITranslation
 
 
-def filter_links(region_slug):
+def filter_links(region_slug, link_filter=None):
     """
-    First filters links by current region
-    Then filters by the link state
-    Finally returns resulting query sets
+    Filter all links of one region by the given category
 
     :param region_slug: the current regions slug
     :type region_slug: str
-    :return: dictionary of filtered querysets
-    :rtype: dict
-    """
-    qset = Link.objects.filter(
-        Q(page_translations__page__region__slug=region_slug)
-        | Q(event_translations__event__region__slug=region_slug)
-        | Q(poi_translations__poi__region__slug=region_slug)
-    )
-    qset = qset.order_by("-url__last_checked")
 
-    qset = list(
+    :param link_filter: which links should be returned (one of ``valid``, ``invalid``, ``ignored``, ``unchecked``).
+                        If parameter is not in these choices or omitted, all links are returned by default.
+    :type link_filter: str
+
+    :return: a tuple of the requested links and a dict containing the counters of all remaining links
+    :rtype: tuple
+    """
+    # Get all translation objects of this region and prefetch its links
+    page_translations = list(
+        PageTranslation.objects.filter(page__region__slug=region_slug)
+        .distinct("page")
+        .select_related("language")
+        .prefetch_related("links__url")
+    )
+    event_translations = list(
+        EventTranslation.objects.filter(event__region__slug=region_slug)
+        .distinct("event")
+        .select_related("language")
+        .prefetch_related("links__url")
+    )
+    poi_translations = list(
+        POITranslation.objects.filter(poi__region__slug=region_slug)
+        .distinct("poi")
+        .select_related("language")
+        .prefetch_related("links__url")
+    )
+    # Filter out translations with no links
+    translations = list(
         filter(
-            lambda link: (link.content_object.latest_revision == link.content_object),
-            qset,
+            lambda translation: translation.links,
+            page_translations + event_translations + poi_translations,
         )
     )
-
-    valid_links = list(filter(lambda link: (not link.ignore and link.url.status), qset))
-    unchecked_links = list(
-        filter(lambda link: (not link.ignore and link.url.last_checked is None), qset)
-    )
-    ignored_links = list(filter(lambda link: (link.ignore), qset))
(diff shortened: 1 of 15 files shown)


timoludwig (commented):
Thanks a lot for your PR! :+1:

In general, this works just fine. However, I noticed a few performance problems once the database is filled up with links: E.g. if there are 100 links in the database, the current implementation needs 500 sql queries to render the link list. And in the production database, we can easily expect several thousand links...

I think the main problem is that for every link item, the access of `link.content_object` triggers a database access and each calculation of the backend edit link triggers another one because it needs the language slug of the current translation object...
I propose a complete refactoring of the link list by querying the latest versions of all translations first, and then query for the links of these versions afterwards.

Since GitHub's suggestion tool doesn't allow me to suggest such substantial changes, I will just push a new commit, and then you can have a look yourself if you think these changes make sense and whether you find something else we could possibly optimize.

src/cms/utils/filter_links.py
```diff
@@ -1,33 +1,78 @@
...
+    # Return the requested links
+    if link_filter == "valid":
+        links = valid_links
+    elif link_filter == "unchecked":
+        links = unchecked_links
+    elif link_filter == "ignored":
+        links = ignored_links
+    elif link_filter == "invalid":
+        links = invalid_links
+    return links, count_dict
```
david-venhoff, line 78: In the original version of this function, the links where ordered by `-url__last_checked`. Should we keep this behavior?

src/cms/utils/filter_links.py
```diff
@@ -1,33 +1,78 @@
...
-        Q(page_translations__page__region__slug=region_slug)
-        | Q(event_translations__event__region__slug=region_slug)
-        | Q(poi_translations__poi__region__slug=region_slug)
+    # Get all translation objects of this region and prefetch its links
+    page_translations = list(
+        PageTranslation.objects.filter(page__region__slug=region_slug)
+        .distinct("page")
+        .select_related("language")
+        .prefetch_related("links__url")
+    )
```
david-venhoff, line 24: It might make sense to also fetch `content_type` since that is accessed in `link_list_row.html`, too.
It looks like this change reduces the number of queries by one per link.

```suggestion
        .prefetch_related("links__content_type")
    )
```

src/cms/utils/filter_links.py
```diff
@@ -1,33 +1,78 @@
...
+    event_translations = list(
+        EventTranslation.objects.filter(event__region__slug=region_slug)
+        .distinct("event")
+        .select_related("language")
+        .prefetch_related("links__url")
+    )
```
david-venhoff, line 30:
```suggestion
        .prefetch_related("links__content_type")
    )
```

src/cms/utils/filter_links.py
```diff
@@ -1,33 +1,78 @@
...
+    poi_translations = list(
+        POITranslation.objects.filter(poi__region__slug=region_slug)
+        .distinct("poi")
+        .select_related("language")
+        .prefetch_related("links__url")
+    )
```
david-venhoff, line 36:
```suggestion
        .prefetch_related("links__content_type")
    )
```

david-venhoff (commented):
I think your changes look really good. I just got a question regarding the `filter_links` function.

src/cms/utils/filter_links.py
timoludwig: Ah, very good idea! :+1:

Although, now that I think about it: probably we can circumvent the query altogether by getting the type from the `item.translation` object directly via `item.translation._meta.verbose_name`. The only problem is that private members cannot be referenced in the template, so maybe we have to add another templatetag to return the readable type from a model instance?

```
@register.simple_tag
def get_model_verbose_name(instance):
    return instance._meta.verbose_name
```
and then in `link_list_row.html`:
```
<td class="pr-2">{% get_model_verbose_name item.translation %}</td>
```

src/cms/utils/filter_links.py
timoludwig: I'm not sure, I think it could also be helpful to have the links grouped by their translations... In the long term, we should add sorting & filters like for other lists, but at the moment this is fine I think.

timoludwig:
I noticed another problem with re-checking links: The `map()`-function is evaluated lazily and doesn't do anything if the return value is unused. So I converted it into a regular loop.

src/cms/utils/filter_links.py
timoludwig: I tried to address the sql queries of `content_type`, but somehow they still appear in the Django debug toolbar... It seems that `prefetch_related("links__url")` implicitly also queries for the `content_type`, even if it's unused in the template. Do you have another idea what might cause this? If not, no problem, since this is only one query per content type indepdendently from the number of links in the database, I don't this this is causing too much overhead.

timoludwig:
@david-venhoff thanks for reviewing my changes, can you have another look?
I also rebased the branch on the current develop branch and resolved the conflicts, so if you are fine with the changes, you can merge the PR!

timoludwig (approved):

src/cms/utils/filter_links.py
david-venhoff: I cannot find an sql query that queries for `content_type`. Can you maybe copy the query so I can take a look at it?

david-venhoff:
The changes look good to me 😃
Before merge I'd like to understand the problem with unnecessary queries that you mentioned.

PR #1009, opened 2021-10-31
Improve browser compatibility of automatic file download

timoludwig:
### Short description

On some browsers, I noticed that the automatic download doesn't work - instead, the xliff file is directly opened as XML file in the browser.

### Proposed changes

- Use `download` link attribute
- Trigger click of link instead of setting `window.location.href`


ulliholtgrave:
Do you know which browser are affected? I can't really reproduce it with my firefox and chromium

ulliholtgrave (approved):
Seems to fix the problem :+1:

PR #1010, opened 2021-10-31
Add missing access control to link checker

timoludwig:
### Short description

At the moment, there is no access control for the broken link checker.
Since the links itself are publicly available via the API, the worst that can happen is that unauthorized users trigger a re-check of links or ignore/unignore broken links. Nonetheless, we should fix this asap.

### Proposed changes

- Only allow link checker page for logged in users
- Only allow link checker page for users of that region


ulliholtgrave (approved):
Just tested it out. Thank you for fixing that and closing that security flaw :)

PR #1016, opened 2021-11-10
Stop overwriting link text

ulliholtgrave:
### Short description
This PR changes the behavior that an existing link text gets overwritten by the link, if we select some text input that we want use for linking.

### DIFF
--- a/src/frontend/js/tinymce-plugins/custom_link_input/plugin.js
+++ b/src/frontend/js/tinymce-plugins/custom_link_input/plugin.js
@@ -86,7 +86,7 @@ import { getCsrfToken } from "../../utils/csrf-token";
                 let url_changed_by_search = false;
                 // Check if the selected completion changed
                 if (prev_selected_completion != data.completions) {
-                    // find the correct thext currently shown in the completion items box
+                    // find the correct text currently shown in the completion items box
                     if (completion_items.length > 0) {
                         const current_completion = completion_items.find((completion) => completion.value == data.completions);
                         // Don't set the completion text to `- no results -`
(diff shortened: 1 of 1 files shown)


src/frontend/js/tinymce-plugins/custom_link_input/plugin.js
```diff
@@ -116,7 +116,7 @@ import { getCsrfToken } from "../../utils/csrf-token";
 
                 // Automatically update the text input to the url by default
                 data = api.getData();
-                if (!text_disabled && !url_changed_by_search && data.url != data.text && prev_link_url != data.url && (data.text != user_data.text || !data.text)) {
+                if (!text_disabled && !url_changed_by_search && data.url != data.text && prev_link_url != data.url && (data.text != user_data.text) && !data.text) {
```
david-venhoff, line 119: Probably the text should only be set to the url if the current text is equal to the previous url
 
This could simplify the condition a bit:

```suggestion
                if (!text_disabled && !url_changed_by_search && data.text == prev_link_url) {
```

david-venhoff (commented):
Unfortunately this change causes the text to not update, even if it should (for example if the text field is empty).
I think the condition could be simplified, such that the text only gets automatically updated if it matches the automatic text for the previous url.
Another option would be to disable the automatic text updating all together, if it causes too much confusion.

src/frontend/js/tinymce-plugins/custom_link_input/plugin.js
ulliholtgrave: Thanks, I added your suggestion to the PR :)

ulliholtgrave:
@david-venhoff I added you suggestion to the PR state

Issue #931, opened 2021-09-01
Search for media library


david-venhoff:
Additionally to the search functionality for other views, the media library should be searchable, too.
For more information see #63.

PR #1021, opened 2021-11-14
Add search function for Media Library

MizukiTemma:
### Short description
The goal of this PR is to add  search funcionality for media library so users can look for directories and files that match key words.

So far keyword search of items is functioning, but the result is shown as Json string not properly displayed.

At the moment the chanes are applied only for admin media library but we can add the changes also to non-admin media library once everything is done.

I appreciate your help and advice!

### What already done

- Search box and button
- Search suggestion by ajax call
- Search for directories and media files that much the keyword

### What not yet functioning
- Display of searched items

### Resolved issues

Fixes: #931

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ UNRELEASED
 * Fully functional media library in selection window
 * Align language flags and translation status icons
 * Fix error when replacing media files without thumbnail
+* Add search function for media library
 
 
 2021.11.0-beta
--- a/integreat_cms/cms/models/media/directory.py
+++ b/integreat_cms/cms/models/media/directory.py
@@ -1,4 +1,5 @@
 from django.db import models
+from django.db.models import Q
 from django.utils import timezone
 from django.utils.formats import localize
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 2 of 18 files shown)


melegiul:
>The goal of this PR is to add search funcionality for media library so users can look for directories and files that match key words.
So far keyword search of items is functioning, but the result is shown as Json string not properly displayed.

Looks great to me so far, could you please make a rebase to the current develop branch followed by a clean install?
That would make it easier for us to step into your branch and help you, because the current develop branch has a different project structure

timoludwig (commented):
@MizukiTemma @membralala I refactored the media library a bit to make the search functionality work:
- I added two different types of media library components: `DirectoryContentLibrary` for the "normal" view of directory content and `SearchResultLibrary` to view all search results for a given query.
- The preact router now extracts either the directory id or the search query from the url and renders the respective component
- The components either load the normal directory content or the search results, store the result in the state `mediaLibraryContent` and render the generic `Library` component.

Could you have a look at my changes and see if you think they make sense and if you have any suggestions for improvement? Please also tell if you don't understand something, then probably I should improve the documentation of the code.

timoludwig (approved):
From my perspective, this is ready to merge. :rocket: 
@MizukiTemma @membralala any objections?

membralala:
Thanks for your support, @timoludwig, as far as I can see everything works flawless. Thanks also for the detailed comments!

Issue #943, opened 2021-09-23
Remove redundant database calls


ulliholtgrave:
### Motivation
Our current code basis includes a lot of duplicate calls to our database. They should be removed towards an efficient performance level. 

### Proposed Solution
Check the following:

- Redundant calls for object properties (e.g. page.default_translation called multiple times)
- Redundant calls for foreign keys (e.g. page.translations.X ... e.g. page.translations.Y in the same model)
- Redundant calls for methods with the same arguments (e.g. get_translation('de') ... get_translations('de'))

For the following models the views, templates, models and api should be examined. Furthermore, their respective occurrence in utils should be investigated.
- [ ] chat
- [ ] events (Timo)
- [ ] feedback (Guiliano)
- [ ] languages & language tree (Guiliano)
- [ ] media (Ulli)
- [ ] offers
- [ ] pages (Timo)
- [ ] pois (Timo)
- [ ] push_notifications (Sven)
- [ ] regions (Ulli)
- [ ] statistics (Ulli)
- [ ] users, roles & organizations (Sven)

svenseeberg:
FYI: one quick way to output the executed SQL queries is adding a a logger to the settings.py:

```
        "django.db.backends": {
            "level": "DEBUG",
            "handlers": [],
        },
```

membralala:
FYI: A while ago I closed #944 without further ado, however meanwhile I found, that there are some cases, where there are duplicated queries in the same call. In this cases it seems, that using djangos ```@cached_property``` fixes this and even reduces the loading times, when redis is active.

ulliholtgrave:
We may consider to cache the result or Matomo requests

svenseeberg:
@ulliholtgrave I removed the "statistics" from the list as we do not have any relevant database calls in our CMS regarding statistics. Right?

PR #1022, opened 2021-11-17
Improve region performance

david-venhoff:
### Short description
This pr removes some database queries related to regions.
Related issue: #943 

### Proposed changes
- For region list view: pass a list of regions to the paginator, not a query. This prevents an unnecessary `SELECT Count(*) ...` query. Is it worth doing this for every model list view?
- Use `select_related` on the `Region.default_language` property. this reduces the number of queries on every view by one for every region which has `Include administrative division into name` checked. It would be possible to go even further and include the prefix as part of the name so that no additional queries are required at all, but the situation where the root language of a region gets changed would have to be handled correctly.

### DIFF
--- a/integreat_cms/cms/views/regions/region_list_view.py
+++ b/integreat_cms/cms/views/regions/region_list_view.py
@@ -55,7 +55,6 @@ class RegionListView(TemplateView):
             regions = regions.filter(pk__in=region_keys)
 
         chunk_size = int(request.GET.get("size", settings.PER_PAGE))
-        # for consistent pagination querysets should be ordered
         paginator = Paginator(regions, chunk_size)
         chunk = request.GET.get("page")
         region_chunk = paginator.get_page(chunk)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 UNRELEASED
 ----------
 
+* Improve performance of region list
+
 
 2021.11.0-beta
 --------------
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -18,81 +18,81 @@
     "default": {
         "aiohttp": {
             "hashes": [
-                "sha256:09754a0d5eaab66c37591f2f8fac8f9781a5f61d51aa852a3261c4805ca6b984",
-                "sha256:097ecf52f6b9859b025c1e36401f8aa4573552e887d1b91b4b999d68d0b5a3b3",
-                "sha256:0a96473a1f61d7920a9099bc8e729dc8282539d25f79c12573ee0fdb9c8b66a8",
-                "sha256:0af379221975054162959e00daf21159ff69a712fc42ed0052caddbd70d52ff4",
-                "sha256:0d7b056fd3972d353cb4bc305c03f9381583766b7f8c7f1c44478dba69099e33",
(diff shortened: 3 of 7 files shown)


timoludwig (commented):
Thanks for your PR! :+1:

I had some major suggestions which I couldn't add via GitHub's UI, so I directly pushed a commit, I hope that's ok. The most significant change is a custom model manager for regions which always prefetches the default language of a region. I think this makes sense because nearly every time a region object is fetched from the database, sooner or later `region.full_name` is called and it comes in handy if the necessary information doesn't need to be fetched from the database.
We can't include the prefix as part of the name, because the app needs the name without prefix for their sorting in the region list (so e.g. "Stadt Augsburg" is listed under "A" instead of "S")... we only could store the prefix in its own charfield which doesn't need to be translated in the first place, but for now I would keep the database structure as it is and see how much performance improvement we can achieve by just adapting the queries.

Could you have a look at my changes and check whether you see any flaws or room for improvement? Thanks!

integreat_cms/cms/views/regions/region_list_view.py
```diff
@@ -55,7 +55,6 @@ def get(self, request, *args, **kwargs):
             regions = regions.filter(pk__in=region_keys)
 
         chunk_size = int(request.GET.get("size", settings.PER_PAGE))
-        # for consistent pagination querysets should be ordered
         paginator = Paginator(regions, chunk_size)
```
timoludwig, line 58: I think it's a bad idea to convert this queryset to a list manually - if this would be more efficient in all scenarios, I think the Paginator class would do this itself.
The problem is, that calling `list()` instantly forces the queryset to evaluate, before it's even passed to the paginator. So if there are 1000 objects in the database, and the `chunk_size` is 10, passing the queryset as list would mean to retrieve all 1000 objects from the database, and then throwing 990 objects away to get the current page. In contrast, when we pass a queryset, it can do one query to determine that there are 1000 objects and then it can do a second query to retrieve only the 10 objects it needs. So even if we'll probably never have 1000 regions, I think it makes sense to keep the current behaviour and do an additional `COUNT()` for every pagination.

david-venhoff: Of course you are right!

david-venhoff:
These changes look good to me. I think that one problem with this approach is that this could lead to unnecessary prefetch calls in the future if the language tree nodes are not required for some database calls (?), but I cannot think of a practical better approach.

timoludwig (approved):
> I think that one problem with this approach is that this could lead to unnecessary prefetch calls in the future if the language tree nodes are not required for some database calls (?), but I cannot think of a practical better approach.

Yes you're right, this might become a problem. But we can always define a second model manager without the prefetching and use that one if needed. At the moment, I think it's important to focus on potential bottlenecks when multiple regions are loaded (e.g. in the region list, the region API or in the model choice fields where regions can be selected) and in all of these cases, the `full_name` is used afterwards. If there are views where only a single region is requested (without a call to `full_name`), it's no big deal when one single useless query is performed... it only becomes a problem, if the same thing happens for _n_ regions etc... but yeah, let's keep an eye on this.

For now, I'd suggest to merge this PR, put a check mark behind "regions" and move on to the next model :wink:

PR #1027, opened 2021-11-23
Fix cacheops config

timoludwig:
### Short description

Fix a few config problems with cacheops

### Proposed changes

- Tell cacheops about redis socket if configured
- Cache cms and linkcheck database tables


membralala (approved):
Thanks for this! Everything seems to work fine and fixes the issue, that I had.

Issue #1031, opened 2021-11-23
Empty `creator` fields not handled correctly


timoludwig:
### Describe the Bug

Normally, the `creator` field of translations should always have a valid relation to the user model.
However, if a user gets deleted, this field becomes `None`, which is not handled correctly at the moment.

### Actual Behavior

- When duplicating regions, it's not possible to duplicate pages when the creator field of a translation is `None` because the model cannot be validated.
- In the version list, just "Author:" is shown, which could be misleading

### Expected Behavior

- The pages should be duplicated with the same empty creator field
- We should show "Author: _deleted user_" or similar in the version list

PR #1032, opened 2021-11-23
Handle empty `creator` fields correctly 

timoludwig:
### Short description

Handle empty creator fields correctly 

### Proposed changes

- Fix typo in roles fixture
- Handle empty `creator` fields correctly 
    - Show 'deleted user' instead of empty string in revision view
    - Allow duplication of translations with empty `creator` fields

### Resolved issues

Fixes: #1031


ulliholtgrave (approved):
Tested it locally and seems to work :+1:

Issue #1018, opened 2021-11-12
Update css-loader


timoludwig:
### Motivation

When [css-loader](https://www.npmjs.com/package/css-loader) is updated to a later version than `6.0.0`, our css isn't working correctly anymore.

### Proposed Solution

- Check the [:warning: BREAKING CHANGES](https://github.com/webpack-contrib/css-loader/blob/master/CHANGELOG.md#600-2021-07-14) section in the changelog
- Figure out what we have to change to make our bundling work with the new version

PR #1034, opened 2021-11-24
Update npm dependencies

timoludwig:
### Proposed changes

- Update our npm dependencies again (except css-loader because it breaks our css, see #1018).
This should hopefully fix our [dependabot alerts](https://github.com/Integreat/integreat-cms/security/dependabot).
- Drop `node-sass`, `cssnano` and `jquery` (I couldn't find any usages of these packages and everything seems to work fine without them. Probably there are a few similar packages which could be removed)
- Use `window.setTimeout` explicitly to fix TypeScript error


package.json
```diff
@@ -1,75 +1,58 @@
 {
-  "name": "integreat-cms",
```
ulliholtgrave, line 2: I would suggest that we keep the "name" field as it seems to be generated every time we run the installation procedure.

ulliholtgrave (approved):
I went to through the whole installation process and everything seems to work.

Issue #1028, opened 2021-11-23
Page permissions aren't working anymore


timoludwig:
### Describe the Bug

Since our change to a custom user model, our page object-specific permissions aren't working anymore.
Apparently, django-rules has problems with this (see https://github.com/dfunckt/django-rules/issues/151).
The biggest problem is that it breaks the normal page permissions, even if page-specific permissions are disabled.

### Steps to Reproduce

1. Create a non-superuser which should have the permissions to edit all pages
2. Login with this user and open a page form

### Expected Behavior

The user should be able to edit the page

### Actual Behavior

It's not possible to edit/publish the page and the following messages are shown:
![Screenshot 2021-11-23 at 15-26-35 Integreat Editorial System](https://user-images.githubusercontent.com/16021269/143042237-b6db9457-8ebe-403a-8c40-fae64d180c30.png)

PR #1041, opened 2021-11-24
Fix page permissions

david-venhoff:
### Short description
This pr fixes a bug where non-admin users were not granted permission to edit pages, even if they were allowed to.
It looks like the bug might not actually be related to a django-rules issue. Instead, our [rules.py](https://github.com/Integreat/integreat-cms/blob/develop/integreat_cms/cms/rules.py) did not get loaded, which lead to the custom permission rules never being registered. This might be caused by the recent refactoring of the project structure.

### Proposed changes
- Tell the rules app to autodiscover rules config files

### Resolved issues
Fixes: #1028

### DIFF
--- a/integreat_cms/core/settings.py
+++ b/integreat_cms/core/settings.py
@@ -174,7 +174,7 @@ INSTALLED_APPS = [
     "corsheaders",
     "linkcheck",
     "mptt",
-    "rules",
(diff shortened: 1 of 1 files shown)


timoludwig (approved):
Ahh okay, makes sense. Sorry for messing that up and thanks a lot for fixing it! :+1: :muscle:

PR #1042, opened 2021-11-25
Fix timeout bug by using the specific window method

ulliholtgrave:
### Short description
After the update of our JS dependencies, we can no longer use false for checking the existence of the timeout function.
`[webpack]       TS2322: Type 'Timeout' is not assignable to type 'number | false'.
[webpack]   Type 'Timeout' is not assignable to type 'false'.
[webpack] 
[webpack] webpack compiled with 1 error
`
### Proposed changes

- Change the Timeout type to null and change check accordingly.
- Use window.setTimeout() to distinguish from further Timeout functions.


david-venhoff:
Looks good to me 🙂

david-venhoff (approved):

PR #1044, opened 2021-11-25
Improve production setup

timoludwig:
### Short description

This adds various improvements for the production setup routine.

### Proposed changes

- Move dev dependencies to `setup.py` again (This makes it easier to install the debug dependencies via pip, e.g. `pip install integreat-cms[debug]`)
- Change handling of `SECRET_KEY` - Django already handles empty keys
- Refactor settings
    - Enable more settings to be set via environment variables
    - Change the environment variable prefix from `DJANGO_` to `INTEGREAT_CMS_`
    - Correctly parse boolean values that are passed as strings
- Read settings from config file `/etc/integreat-cms.ini`
- Update production documentation
- Add `MANIFEST.in` file to exclude some more files from the package

Unfortunately, `pipenv lock -r > requirements.txt` outputs all dependencies including dev dependencies and I found no way to exclude them. So for now, we don't have a good way to make sure the package versions on the dev environment and the production server match.


ulliholtgrave (approved):
Good job! Let's see how the whole pipeline will behave :)

PR #1050, opened 2021-12-04
Changes language tabs, fixes #1030

### LINKED ISSUE #1030: Language tabs scale poorly
### Describe the Bug

Our language tab design doesn't scale well if the number of languages is increasing or the screen size is small.

Screenshot 2021-11-23 at 19-22-46 Integreat Redaktionssystem

### Expected Behavior

I see three main problems:
- The "Version" link should not cause the language tabs to float. Instead, we should maybe move it back into the white box to the rest of the form?
- Tabs in the second row should expand until the form-box (so no light blue background should be visible between the tabs and the box). But not sure whether that's doable, since the space underneath is in theory taken by the following flex items...
- If a tab in the top row is selected, it shouldn't have a bottom white border (because this causes it to "melt" into the bottom tab which makes no sense - the idea was that it should "melt" into the box itself). But no idea whether we can safely detect whether a tab is in the top bar or the bottom bar, since it's just one flexbox...

### Actual Behavior

Everything looks weird

JoeyStk:
### Short description
This PR changes the look and function of the language tabs.

### Proposed changes
So there are four things I changed in order to find a solution. All of this is up for discussion, of course. I just wanted to show you some ideas of what I think can be fixed. 

- Change the order of the language row. Maybe this make a lot of sense - or just makes sense, when there are more than one row. But in case of more than one rows the active tab would be in the bottom row, which is the Windows solutions.
- Change the background color to blue for non-active tabs, change the background-color to white for active tabs. That way it seems like the active tab is merging into the content-card. This is the most common solution I found when I was googling reference for tabs.
- Add some margin, so the tabs don't overlap
- Change the active effect to look more like the hover effect.

I'm aware that all of the proposed changes are just quick fixes for now. For a more sustainable solution in the long run it might be a good idea to involve someone from UI/UX here. 
Maybe we should have a discussion about this in the Mattermost channel. I wasn't sure what the common procedure with such issues is.

Looking forward for ideas and feedback! 

Fixes: #1030

### DIFF
--- a/integreat_cms/cms/templates/pages/page_form.html
+++ b/integreat_cms/cms/templates/pages/page_form.html
@@ -55,11 +55,11 @@
             {% endif %}
         </div>
         <div class="flex flex-wrap flex-col flex-grow pr-2 w-2/3">
-            <ul class="flex flex-wrap items-end pl-4">
+            <ul class="flex flex-wrap flex-wrap-reverse items-end pl-4">
                 {% for other_language in languages %}
-                    <li class="mr-1 {% if other_language == language %}z-10{% endif %}" style="margin-bottom: -2px">
-                        <div class="bg-white text-blue-500 {% if other_language != language %}hover:bg-blue-500 hover:text-white{% endif %} border-l-2 border-t-2 border-r-2 border-blue-500 font-bold rounded-t-lg">
-                            <div class="border-b-2 border-white">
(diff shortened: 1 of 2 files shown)


ulliholtgrave:
Hey Johannes, Thank you for your effort, but I think that we'll want to do a little bit more in this field. I talked with Nikola about this and we had some ideas, how we want to change this. I tried to put my ideas in the related issue and did some "mockup" with my limited design skills :D

PR #1058, opened 2021-12-07
Fix console error on page load

david-venhoff:
### Short description
Right now, when loading a page a console error shows up. 

### Proposed changes
- Don't try to access a property of null
- Alternatively, we could maybe not load this file when it is not needed


timoludwig (approved):
Good catch! :muscle:

Issue #1029, opened 2021-11-23
Language grid a little bit off


timoludwig:
### Describe the Bug

On all list with translations, the language flags and the corresponding status icons are a little bit misplaced.
With a few languages, this doesn't really matter, but since the difference accumulates, it becomes visible with very many languages:
![Screenshot 2021-11-23 at 19-17-28 Integreat Redaktionssystem](https://user-images.githubusercontent.com/16021269/143081685-45d5064f-0869-45ec-815f-7b2ba82002cd.png)

### Expected Behavior

The language flags and the status icons should align nicely

### Actual Behavior

They are a little bit off

PR #1059, opened 2021-12-08
Sync language grid width for table header and body

melegiul:
### Short description

Sync language grid width for table header and body

### Proposed changes

The problem is that despite both Element nodes (a-html-tags) in the header and in body of the the table have the same width of 18.67px, as inline-block elements they are followed whitespace Text nodes with different widths (4,08331px and 3,36664px)
So I updated the Elements node width so that both header and body have in total the exact same width of 22,0333px

Screenshot from 2021-12-08 15-33-03

### Resolved issues

Fixes: #1029

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ UNRELEASED
 * Allow configuration via /etc/integreat-cms.ini
 * Fix dependency versions for production setup
 * Fully functional media library in selection window
+* Align language flags and translation status icons
 
 
 2021.11.0-beta
--- a/integreat_cms/cms/templates/events/event_list.html
+++ b/integreat_cms/cms/templates/events/event_list.html
@@ -51,11 +51,13 @@
                 {% endif %}
                 <th class="text-sm text-left uppercase py-3">
                     <div class="lang-grid flags whitespace-nowrap">
-                        {% for lang in languages %}
-                            <a href="{% url 'events' region_slug=region.slug language_slug=lang.slug %}">
-                                <span class="fp fp-rounded fp-{{ lang.primary_country_code }}" title="{{ lang.translated_name }}"></span>
-                            </a>
-                        {% endfor %}
+                        {% spaceless %}
+                            {% for lang in languages %}
+                                <a href="{% url 'events' region_slug=region.slug language_slug=lang.slug %}">
+                                    <span class="fp fp-rounded fp-{{ lang.primary_country_code }}" title="{{ lang.translated_name }}"></span>
+                                </a>
+                            {% endfor %}
+                        {% endspaceless %}
                     </div>
                 </th>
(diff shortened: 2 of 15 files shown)


timoludwig (commented):
Thanks a lot! :+1:
Feels kinda hacky, but if it works :smile: 
Could you add a small notice that this problem is now fixed to our [changelog](https://github.com/Integreat/integreat-cms/blob/develop/CHANGELOG.md)?

melegiul:
Thank you for your review, yes it is hacky at the moment, because I did not found a proper way to sync the width of text node "whitespace". So far I tried by syncing parent nodes font-size, height, etc without success, but I will investigate this topic further in the future. I also updated now the changelog

timoludwig (commented):
Thanks! :+1:
Sorry, I just noticed that it would be good to have the same change also for the events, POIs and push notification lists. Can you maybe add the same class in these lists as well?

melegiul:
> Can you maybe add the same class in these lists as well?

Alright, I've now applied the changes to all other translatable content types

melegiul:
But alternatively we could also go like this:
```
{% for language in languages %}<a href="..."><a/>{% endfor %}
```
Removing the line in front of the a tag and after the closing a tag would remove all these whitespace characters, which are harder to sync by css

timoludwig (approved):
> But alternatively we could also go like this:
> 
> ```
> {% for language in languages %}<a href="..."><a/>{% endfor %}
> ```
> 
> Removing the line break in front of the beginning a tag and after the closing a tag would remove all these whitespace characters, which are harder to sync by css

Oh, I like that idea! Maybe we could even use the `{% spaceless %}` template tag for this:
```
{% spaceless %}
    {% for language in languages %}
        <a href="..."><a/>
    {% endfor %}
{% endspaceless %}
```
But if you don't want to change it further, the current solution would also be fine for me.

melegiul:
> Oh, I like that idea! Maybe we could even use the {% spaceless %} template tag for this:

I like the spaceless tag more, too. So I replced all whitespaces with a little fixed sized right margin, which syncs the elements width, too.

timoludwig (commented):
Thanks, relly cool! :rocket: 

But I'm sorry, now I noticed another thing: The language flags are also used in the lists of archived pages and POIs... could you add the spaceless tag for these templates as well? :see_no_evil:

melegiul:
> The language flags are also used in the lists of archived pages and POIs

Ok, I've added spaceless to archived pages, but in archived pois html this column seems not to exist. Is this intentional?

timoludwig:
> Ok, I've added spaceless to archived pages, but in archived pois html this column seems not to exist. Is this intentional?

Sorry, my bad, I meant archived events.

timoludwig (approved):
Thanks a lot, I think that's it :rocket: :muscle:

PR #1060, opened 2021-12-08
Persist setup.cfg to workspace after CircleCI version bump

timoludwig:
### Short description

The version is now contained in setup.cfg instead of setup.py, so we have to persist a different file after the version has been bumped. This should have been included in 36618d5fac30e0803c3da91d7d3cf4221c531517, but unfortunately I missed it.

### Proposed changes

- Persist `setup.cfg` to workspace after CircleCI version bump

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -261,7 +261,7 @@ jobs:
       - persist_to_workspace:
           root: .
           paths:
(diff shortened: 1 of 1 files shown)


ulliholtgrave (approved):
Let's try :D

PR #1064, opened 2021-12-14
Improve production setup

timoludwig:
### Short description

During my tests with the integreat cms pypi package, I noticed a few small problems:
- If the config file contains a `%`, configparser tries to do a string interpolation
- Some tailwind classes don't seem to be correctly included in the webpack build. This seems to be an issue of NodeJS v17.
- The version in the `__init__.py` was not contained in the final python package because a source path was incorrect
- The file based cache resides inside the virtual environment

### Proposed changes

- Do not interpolate variables with configparser
- Use NodeJS v16 in CircleCI again
- Fix source directory in CircleCI config
- Add option to configure file cache location

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,7 +91,7 @@ jobs:
             - integreat_cms/locale/de/LC_MESSAGES/django.mo
   webpack:
     docker:
-      - image: 'cimg/node:current'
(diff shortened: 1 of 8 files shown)


ulliholtgrave (approved):
Just went trough the files and looks good. I guess the real consequences are only seen after the release :) :+1:

Issue #1062, opened 2021-12-08
Replacing PDF documents doesn't work


timoludwig:
### Describe the Bug

When trying to replace a PDF document, the message:

> A network error has occurred. Please try again later.

appears and the file is not replaced.

### Steps to Reproduce

1. Go to the media library
2. Upload a PDF
3. Try to replace the PDF with another file

### Expected Behavior

The PDF should be replaced

### Actual Behavior

An error occurs

### Additional Information

The traceback in the console:

```
Dec 08 21:32:49 ERROR django.request - Internal Server Error: /ajax/media/replace-file
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.7/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.7/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.7/site-packages/django/views/decorators/http.py", line 40, in inner
    return func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.7/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/integreat-cms/integreat_cms/cms/decorators.py", line 114, in wrap
    return function(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.7/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/integreat-cms/integreat_cms/api/decorators.py", line 112, in wrap
    return function(request, *args, **kwargs)
  File "/opt/integreat-cms/integreat_cms/cms/views/media/media_actions.py", line 248, in replace_file_ajax
    data=request.POST, instance=media_file, files=request.FILES
  File "/opt/integreat-cms/integreat_cms/cms/forms/media/replace_media_file_form.py", line 63, in __init__
    self.original_thumbnail_path = getattr(instance.thumbnail, "path", None)
  File "/opt/integreat-cms/.venv/lib/python3.7/site-packages/django/db/models/fields/files.py", line 58, in path
    self._require_file()
[...]

PR #1067, opened 2021-12-15
Fix crash when replacing a file without thumbnail

david-venhoff:
### Short description

This pr fixes a crash when a user attempts to replace a file in the media library that has no thumbnail.

### Proposed changes

- Check if the file has a valid thumbnail before accessing its path

### Resolved issues

Fixes: #1062

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ UNRELEASED
 * Fix dependency versions for production setup
 * Fully functional media library in selection window
 * Align language flags and translation status icons
+* Fix error when replacing media files without thumbnail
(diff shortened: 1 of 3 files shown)


david-venhoff:
The CI error does not seem to be related to this pr, I'll try to rerun the scan tomorrow

timoludwig (approved):
Thanks a lot! :+1:
I restarted the build, seemed to be a temporary server outage.

timoludwig (commented):
Oh, just one request: Could you add
>* Fix error when replacing media files without thumbnail

to the changelog?

PR #1073, opened 2021-12-22
Use TLS for SMTP connection per default

timoludwig:
### Short description

On the dev system, I noticed that sending emails doesn't work because of an `SMTPNotSupportedError`.
Apparently, `EMAIL_USE_TLS` has to be explicitly activated when trying to send emails via port `587`.

### Proposed changes

- Use TLS for SMTP connection per default
- Allow secure connection type configuration via environment & config file
- Add new settings to example config


ulliholtgrave (approved):
Looks reasonable :+1:

PR #1074, opened 2021-12-22
Fix tag in comparison with previous release

timoludwig:
### Short description

In our [latest GitHub release](https://github.com/digitalfabrik/integreat-cms/releases/tag/2021.12.0-beta), the comparison with the previous tag didn't work as expected. Instead of 

> Compare changes: 2021.11.0-beta → 2021.12.0-beta

it initially showed:

> Compare changes: 2021.11.0-betanexport → 2021.12.0-beta

So presumably, passing the environment variables in CircleCI didn't work correctly.

### Proposed changes

- Write two separate export statements instead of one


ulliholtgrave (approved):
Maybe this helps :D

PR #1081, opened 2022-01-04
Improve performance for poi views

david-venhoff:
### Short description
This pr improves the performance of poi views.
Co-authored-by: philip popien <membralala@users.noreply.github.com>

### Proposed changes
- Load poi translations in one query
- Load the current language just once

### DIFF
--- a/integreat_cms/cms/templates/pois/poi_list.html
+++ b/integreat_cms/cms/templates/pois/poi_list.html
@@ -5,6 +5,9 @@
 {% load content_filters %}
 
 {% block content %}
+
+{% get_current_language as LANGUAGE_CODE %}
+{% get_language LANGUAGE_CODE as backend_language %}
 <div class="table-header">
     <div class="flex flex-wrap justify-between">
         <h1 class="heading">{% trans 'Locations on map' %}</h1>
@@ -33,8 +36,6 @@
         <thead>
             <tr class="border-b border-gray-200">
                 <th class="text-sm text-left uppercase py-3 pl-4 pr-2">{% trans 'Title in' %} {{ language.translated_name }}</th>
-                {% get_current_language as LANGUAGE_CODE %}
-                {% get_language LANGUAGE_CODE as backend_language %}
                 {% if backend_language and backend_language != language %}
                     <th class="text-sm text-left uppercase py-3 pr-2">{% trans 'Title in' %} {{ backend_language.translated_name }}</th>
                 {% endif %}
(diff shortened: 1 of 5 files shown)


integreat_cms/cms/templates/pois/poi_list.html
```diff
@@ -60,7 +61,7 @@ <h1 class="heading">{% trans 'Locations on map' %}</h1>
         <tbody>
         {% for poi in pois %}
             {% get_translation poi language.slug as poi_translation %}
-            {% include "pois/poi_list_row.html" %}
+            {% include "pois/poi_list_row.html" with LANGUAGE_CODE=LANGUAGE_CODE backend_language=backend_language %}
```
timoludwig, line 64:
```suggestion
            {% include "pois/poi_list_row.html" %}
```
Is this really necessary? I thought the context is passed to the included template anyway, the `with` parameters are only required, if the name differs...

integreat_cms/cms/templatetags/content_filters.py
```diff
@@ -30,6 +30,11 @@ def get_translation(instance, language_slug):
     return instance.get_translation(language_slug)
 
 
+@register.simple_tag
+def get_poi_translations(instance):
+    return instance.translations
+
+
```
timoludwig, line 37:
```suggestion
```
I don't think we need a custom template tag for a simple reverse foreign key... You can just use `poi.translations` in the template directly.

integreat_cms/cms/templates/pois/poi_list_row.html
```diff
@@ -32,9 +30,9 @@
     {% endif %}
 	<td class="pr-2 lang-grid">
         {% spaceless %}
-            {% for other_language in languages %}
-                <a href="{% url 'edit_poi' poi_id=poi.id region_slug=region.slug language_slug=other_language.slug %}">
-                    {% get_translation poi other_language.slug as other_translation %}
+            {% get_poi_translations poi as translations %}
+            {% for other_translation in translation %}
+                <a href="{% url 'edit_poi' poi_id=poi.id region_slug=region.slug language_slug=other_translation.language.slug %}">
```
timoludwig, line 35: The problem with this approach is that languages with missing translations are skipped. Also the order might differ from the languages in the grid, so you would have to make sure these are always identical.

timoludwig (commented):
Thanks, looks good!
Could you resolve the conflicts, fix the translations and make the commit message more meaningful? Then this is ready to merge if you don't see any other potential for improvement.

timoludwig (approved):

Issue #1065, opened 2021-12-14
`MultipleObjectsReturned` in page children API endpoint


timoludwig:
### Describe the Bug

When the children of a page with more than one version with the same slug are requested via the `url` parameter, the error `MultipleObjectsReturned` gets thrown.

### Steps to Reproduce

1. Go to the region Augsbug
2. Create at least one additional version of the Willkommen page with the same slug
3. Open http://localhost:8000/api/augsburg/de/children?url=willkommen

### Expected Behavior

The children of the page should be returned

### Actual Behavior

The following error appears:
```
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.8/site-packages/integreat_cms/api/decorators.py", line 112, in wrap
    return function(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.8/site-packages/integreat_cms/api/v3/pages.py", line 184, in children
    root_pages = [get_single_page(request, language_slug)]
  File "/opt/integreat-cms/.venv/lib/python3.8/site-packages/integreat_cms/api/v3/pages.py", line 116, in get_single_page
    page = get_object_or_404(
  File "/opt/integreat-cms/.venv/lib/python3.8/site-packages/django/shortcuts.py", line 76, in get_object_or_404
    return queryset.get(*args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.8/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.8/site-packages/cacheops/query.py", line 351, in get
    return qs._no_monkey.get(qs, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.8/site-packages/django/db/models/query.py", line 439, in get
    raise self.model.MultipleObjectsReturned(
integreat_cms.cms.models.pages.page.Page.MultipleObjectsReturned: get() returned more than one Page -- it returned more than 20!
```

### Additional Information

[...]

PR #1082, opened 2022-01-04
API: change single page filtering

svenseeberg:
### Short description
Fix filtering of pages in the APIv3 single page endpoint with the `url` parameter.

### Proposed changes
Get distinct page(s) for a language and translation slug. Raise a `Http404` if the result is not exactly (less or more than) one page.

### Resolved issues
Fixes: #1065

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 UNRELEASED
 ----------
 
+* Fix APIv3 single page endpoint for multiple translation versions
+
 
 2021.12.0-beta
 --------------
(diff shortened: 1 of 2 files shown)


timoludwig (commented):
Thanks, looks good to me! :+1:
Could you add this fix to the [CHANGELOG.md](https://github.com/digitalfabrik/integreat-cms/blob/develop/CHANGELOG.md)?

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 UNRELEASED
 ----------
-
+* Fix APIv3 single page endpoint for multiple translation versions
```
timoludwig, line 3:
```suggestion

* Fix APIv3 single page endpoint for multiple translation versions

```
A little caveat of our Changelog: In order to be correctly parsed by [awk magic](https://github.com/digitalfabrik/integreat-cms/blob/25b8706a1eb5004e96133fd73c567704238ba54e/.circleci/config.yml#L342), it needs exactly one new line after the heading and two new lines to the following release... Probably I should document this somewhere^^

timoludwig (approved):
Thanks! :+1:

PR #1084, opened 2022-01-05
Fix install script when clean install cannot remove anything

timoludwig:
### Short description

If the directories `.venv`, `node_modules` and `integreat_cms/static/dist` do not exist and a clean installation is executed, the script aborts because grep returns a non-zero exit code when no match is found.

### Proposed changes

- Allow the grep command after the `--clean` install to fail


svenseeberg (approved):
Does what it is supposed to do.

PR #1089, opened 2022-01-05
Improve performance for feedback views

membralala:
### Short description

This PR should improve performance for the admin and region feedback views as it reduces the amount of database queries. 

### Proposed changes

- Add some cached_properties in the feedback models
- Add an external dependency ```django-polymorphic``` ([https://django-polymorphic.readthedocs.io/en/stable/](https://django-polymorphic.readthedocs.io/en/stable/))


svenseeberg:
@membralala can you please elaborate why the polymorphic module improves performance? The reasons for making the change is not really clear to me. I expect that you did not find this by chance?

timoludwig:
> @membralala can you please elaborate why the polymorphic module improves performance? The reasons for making the change is not really clear to me. I expect that you did not find this by chance?

This is the explanation: https://django-polymorphic.readthedocs.io/en/stable/performance.html#performance
The `get_real_instance()` mentioned there is roughly equivalent to our [Feedback.submodel_instance()](https://github.com/digitalfabrik/integreat-cms/blob/80963f12685ba37bf9e46dc71e4096e2d3e0e995/integreat_cms/cms/models/feedback/feedback.py#L62).

timoludwig (approved):
Thanks a lot :heart_eyes: 

Since `django-polymorphic` improves the performance pretty significantly and makes it possible to remove some of our "magic" submodel-instance code (which ultimately improves maintainability of our code base), I think it's justified to add another dependency here. Also, the other changes look very clean and straight-forward to me, good job!

Issue #1023, opened 2021-11-20
Simple API tests


timoludwig:
### Motivation

At the moment, we postponed all "real" testing to the Version 1.2 milestone (see #765).
However, before we release the CMS in production, we should have at least a few basic API tests since this is the most critical part of the system and the it's not possible to use the apps when our API isn't working properly.

### Proposed Solution

For our [test data](https://github.com/Integreat/integreat-cms/blob/develop/integreat_cms/cms/fixtures/test_data.json), define the expected JSON result for each API endpoint and add unit tests to check whether they're equivalent.
If the test data set is missing some models (which means that the API endpoint would return an empty result), add some dummy data to the test data before defining the expected result.

### Additional Context

While you're at it, you could also review our basic view-tests in [integreat_cms/cms/tests/views](https://github.com/Integreat/integreat-cms/tree/328dcd5c8137d9790b5da7e9ab24828003630e49/integreat_cms/cms/tests/views) and check whether there are any views which can be added without much effort (e.g. because they don't expect custom GET/POST params).

PR #1095, opened 2022-01-06
Simple API tests

melegiul:
### Short description

Adds api tests for most important endpoints 

### Proposed changes

- Compares test data with api return values
- Checks json format and status codes

### Resolved issues

Fixes: #1023

### DIFF
--- a/dev-tools/generate_documentation.sh
+++ b/dev-tools/generate_documentation.sh
@@ -72,7 +72,7 @@ find ${SPHINX_DIR}/${SPHINX_APIDOC_EXT_DIR} -type f -name "*.rst" -print0 | xarg
     -e 's/Cms/CMS/g;s/Api/API/g;s/Poi/POI/g;s/Mfa/MFA/g;s/Pdf/PDF/g;s/Xliff1/XLIFF 1.2/g;s/Xliff2/XLIFF 2.0/g;s/Xliff/XLIFF/g' # Make specific keywords uppercase
 
 # Remove inherited members from tests
-find ${SPHINX_DIR}/${SPHINX_APIDOC_EXT_DIR} -type f -name "integreat_cms.cms.tests*.rst" -print0 | xargs -0 --no-run-if-empty sed --in-place '/:inherited-members:/d'
+find ${SPHINX_DIR}/${SPHINX_APIDOC_EXT_DIR} -type f -name "integreat_cms.*.tests.*rst" -print0 | xargs -0 --no-run-if-empty sed --in-place '/:inherited-members:/d'
 
 # Include _urls in sitemap automodule
 # shellcheck disable=SC2251
--- a/integreat_cms/api/tests/api_tests.py
+++ b/integreat_cms/api/tests/api_tests.py
@@ -1,0 +1,69 @@
+from django.test import TestCase
+from linkcheck.listeners import disable_listeners
+from .api_tests_utils import generate_test_functions
+
+endpoint_data = [
+    (
+        "/api/regions/",
+        "/wp-json/extensions/v3/sites/",
+        "integreat_cms/api/tests/expected-outputs/regions.json",
+    ),
+    (
+        "/api/regions/live",
+        "/wp-json/extensions/v3/sites/live",
+        "integreat_cms/api/tests/expected-outputs/regions_live.json",
+    ),
+    (
+        "/api/augsburg/de/pages",
+        "/augsburg/de/wp-json/extensions/v3/pages",
+        "integreat_cms/api/tests/expected-outputs/augsburg_de_pages.json",
+    ),
+    (
+        "/api/augsburg/de/locations",
+        "/augsburg/de/wp-json/extensions/v3/locations",
+        "integreat_cms/api/tests/expected-outputs/augsburg_de_locations.json",
+    ),
+    (
+        "/api/augsburg/de/children",
+        "/augsburg/de/wp-json/extensions/v3/children",
+        "integreat_cms/api/tests/expected-outputs/augsburg_de_children.json",
+    ),
+    (
+        "/api/augsburg/de/children?depth=2",
+        "/augsburg/de/wp-json/extensions/v3/children?depth=2",
+        "integreat_cms/api/tests/expected-outputs/augsburg_de_pages.json",
+    ),
+    (
+        "/api/augsburg/de/events",
+        "/augsburg/de/wp-json/extensions/v3/events",
+        "integreat_cms/api/tests/expected-outputs/augsburg_de_events.json",
+    ),
+    (
+        "/api/augsburg/de/page?id=1",
+        "/augsburg/de/wp-json/extensions/v3/page?id=1",
+        "integreat_cms/api/tests/expected-outputs/augsburg_de_page_1.json",
+    ),
+]
+
+
+class ApiTest(TestCase):
+    """
+    This test class checks all endpoints defined in `endpoint_data`.
+    It verifies that the content delivered by the endpoint is equivalent with the data
+    provided in the corresponding json file.
+    """
+
+    fixtures = [
+        "integreat_cms/cms/fixtures/roles.json",
+        "integreat_cms/cms/fixtures/test_data.json",
+    ]
+
+    @classmethod
+    def setUpClass(cls):
+        with disable_listeners():
+            super().setUpClass()
+
+    generate_test_functions(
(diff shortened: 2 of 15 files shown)


dev-tools/generate_documentation.sh
```diff
@@ -73,6 +73,7 @@ find ${SPHINX_DIR}/${SPHINX_APIDOC_EXT_DIR} -type f -name "*.rst" -print0 | xarg
 
 # Remove inherited members from tests
 find ${SPHINX_DIR}/${SPHINX_APIDOC_EXT_DIR} -type f -name "integreat_cms.cms.tests*.rst" -print0 | xargs -0 --no-run-if-empty sed --in-place '/:inherited-members:/d'
+find ${SPHINX_DIR}/${SPHINX_APIDOC_EXT_DIR} -type f -name "integreat_cms.api.tests*.rst" -print0 | xargs -0 --no-run-if-empty sed --in-place '/:inherited-members:/d'
```
timoludwig, line 76:
```suggestion
find ${SPHINX_DIR}/${SPHINX_APIDOC_EXT_DIR} -type f -name "integreat_cms.*.tests.*rst" -print0 | xargs -0 --no-run-if-empty sed --in-place '/:inherited-members:/d'
```
I think one wildcard matching for all apps is fine here.

integreat_cms/api/tests/api_tests.py
```diff
@@ -0,0 +1,64 @@
...
+
+    fixtures = [
+        "integreat_cms/cms/fixtures/roles.json",
+        "integreat_cms/cms/fixtures/test_data.json",
+    ]
+
+    @classmethod
+    def setUpClass(cls):
+        with disable_listeners():
+            super().setUpClass()
```
timoludwig, line 59:
```suggestion
```
When you rebase your branch onto the current develop branch, I think this part isn't required anymore. The corresponding issue of the linkcheck listeners should be fixed...

integreat_cms/api/tests/api_tests_utils.py
```diff
@@ -0,0 +1,36 @@
...
+    This function generates test functions for a list of views.
+
+    :param class_variables: The class variable dict in which the functions should be inserted
+    :type class_variables: dict
+
+    :param endpoint_data: The list of endpoints for which the test functions should be generated
+    :type endpoint_data: list
+    """
+    for data in endpoint_data:
+        endpoint, wp_endpoint, fixture = data
```
timoludwig, line 19:
```suggestion
    for endpoint, wp_endpoint, expected_result in endpoint_data:
```
You can expand the parameters directly in the for loop.

integreat_cms/api/tests/api_tests_utils.py
```diff
@@ -0,0 +1,36 @@
...
+        endpoint_str = re.sub("[^a-zA-Z0-9]", "_", endpoint.strip("/"))
+
+        def test_function(
+            self, endpoint=endpoint, fixture=fixture, wp_endpoint=wp_endpoint
```
timoludwig, line 23:
```suggestion
            self, endpoint=endpoint, wp_endpoint=wp_endpoint, expected_result=expected_result
```
I'd rename `fixture` to `expected_result` to point out that these json files are not intended to be imported as Django fixtures.

integreat_cms/api/tests/api_tests_utils.py
```diff
@@ -0,0 +1,36 @@
...
+        ):
+            response = self.client.get(endpoint, format="json")
+            self.assertEqual(response.status_code, 200)
+            response_wp = self.client.get(wp_endpoint, format="json")
+            self.assertEqual(response_wp.status_code, 200)
+            with open(fixture, encoding="utf-8") as f:
```
timoludwig, line 29:
```suggestion
            with open(expected_result, encoding="utf-8") as f:
```

integreat_cms/api/tests/api_tests.py
```diff
@@ -0,0 +1,64 @@
+from django.test import TestCase
+from linkcheck.listeners import disable_listeners
```
timoludwig, line 2:
```suggestion
```

timoludwig (commented):
Thanks a lot! :muscle:  :rocket: 
Looks very good and is easily extendable in the future, I like it! :+1:
See my comments below for a few minor improvements.

integreat_cms/api/tests/expected-outputs/regions_live.json
```diff
@@ -0,0 +1,30 @@
+[
+    {
+        "id": 1,
+        "name": "City Augsburg",
```
timoludwig, line 4:
```suggestion
        "name": "Stadt Augsburg",
```
The language of the prefix should always be in the default language of the region. If it is not, we have found a bug in the API.

integreat_cms/api/tests/expected-outputs/regions_live.json
```diff
@@ -0,0 +1,30 @@
...
+        "path": "augsburg",
+        "prefix": "City",
```
timoludwig, line 6:
```suggestion
        "prefix": "Stadt",
```

integreat_cms/api/tests/api_tests.py
```diff
@@ -0,0 +1,64 @@
...
+from .api_tests_utils import generate_test_functions
+
+endpoint_data = [
+    (
```
timoludwig, line 6:
```suggestion
    (
        "/api/regions/",
        "/wp-json/extensions/v3/sites/",
        "integreat_cms/api/tests/expected-outputs/regions.json",
    ),
    (
```
You already have the expected json for all regions, so why not just test it as well?

melegiul: Testing API all region method has a problem:
The test always fails because the http status code is 301 permanently moved
But we can test it easily when dropping the status 200 check
as we compare the json response anyways, imo we can drop the status check

melegiul: Alternatively we could allow 200 and 301 as assertions

timoludwig: Are you sure you didn't just forget a trailing slash? Then, the 301 is expected... we can always directly check the url where the 301 redirect is pointing to...

integreat_cms/api/tests/api_tests.py
timoludwig: Damn , I really thought the new linkcheck version would fix that :(
Sorry about that!

integreat_cms/api/tests/api_tests.py
melegiul: Indeed, the trailing slash seems to fix the 301 code, I didn't now that
Thank you :smile:

integreat_cms/api/tests/expected-outputs/regions_live.json
melegiul: Strange, only circle ci api returns here "city", my local dev api return "Stadt"
```
FAIL: test_api_regions (integreat_cms.api.tests.api_tests.ApiTest)
Test if the view ``api_regions`` returns the HTTP status code ``200`` with correct data
  File "/home/circleci/project/integreat_cms/api/tests/api_tests_utils.py", line 33, in test_function
    self.assertEqual(regions, response.json())
AssertionError: Lists differ: [{'id': 1, 'name': 'Stadt Augsburg', 'path': 'augsburg', 'liv[514 chars]"]'}] != [{'id': 1, 'name': 'City Augsburg', 'path': 'augsburg', 'live[512 chars]"]'}]

First differing element 0:
{'id': 1, 'name': 'Stadt Augsburg', 'path': 'augsburg', 'liv[224 chars]g"]'}
{'id': 1, 'name': 'City Augsburg', 'path': 'augsburg', 'live[222 chars]g"]'}
```

melegiul: Both use the same fixture, so I think it is something setting/database related :thinking:

melegiul: No, actually the translation file is not existing in this step at circle ci

timoludwig: Ahhhh ok this makes sense. Thanks for digging into that rabbit hole :rabbit:

timoludwig (approved):
Thanks a lot, looks good now :+1:
After cleaning up the Git history a bit, you can merge this PR :tada:

PR #1096, opened 2022-01-06
APIv3: add parents/ancestors end point, fixes #19

### LINKED ISSUE #19: API: parents endpoint
Endpoint to get parents of a page up to the root. This must be compatible to Integreat API v3.

https://github.com/Integreat/cms/wiki/REST-APIv3-Documentation#parents-of-a-post

@maxammann @Simple-codinger is this used by any app? Otherwise we could move this to another milestone.

svenseeberg:
### Short description
Add parents (actually ancestors) APIv3 end point.

### Proposed changes
Add a new function and URL to return all ancestors of a page.

### Resolved issues

Fixes: #19

### DIFF
--- a/integreat_cms/api/v3/pages.py
+++ b/integreat_cms/api/v3/pages.py
@@ -200,3 +200,33 @@ def children(request, region_slug, language_slug):
             if public_translation:
                 result.append(transform_page(public_translation))
     return JsonResponse(result, safe=False)
+
+
+@json_response
+# pylint: disable=unused-argument
+def parents(request, region_slug, language_slug):
+    """
+    Retrieves all ancestors (parent and all nodes up to the root node) of a page
+
+    :param request: The request that has been sent to the Django server
+    :type request: ~django.http.HttpRequest
+
+    :param region_slug: Slug defining the region
+    :type region_slug: str
+
+    :param language_slug: Code to identify the desired language
+    :type language_slug: str
+
+    :raises ~django.http.Http404: HTTP status 404 if the request is malformed or no page with the given id or url exists.
+
+    :return: JSON with the requested page ancestors
+    :rtype: ~django.http.JsonResponse
+    """
+    current_page = get_single_page(request, language_slug)
+    result = []
+    for ancestor in current_page.get_ancestors(include_self=False):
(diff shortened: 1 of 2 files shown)


integreat_cms/api/v3/pages.py
```diff
@@ -200,3 +200,32 @@ def children(request, region_slug, language_slug):
...
+
+    :return: JSON with the requested page ancestors
+    :rtype: ~django.http.JsonResponse
+    """
+    current_page = get_single_page(request, language_slug)
+    result = []
+    for ancestor in current_page.get_ancestors(include_self=False):
+        public_translation = ancestor.get_public_translation(language_slug)
+        if public_translation:
+            result.append(transform_page(public_translation))
```
timoludwig, line 230: I'm not sure what the expected output would be if one of the ancestors didn't have a public translation...
E.g, consider the following sub-tree:
`Public Root Page` -> `Hidden Category` -> `Public Page`
When the parents-endpoint for the `Public Page` is retrieved, do we really want to return just the `[Public Root Page]`, or should we rather throw a `Http404` for the whole request (since the `Public Page` shouldn't be shown in the app anyway because one of its parents is hidden)?
Or do we not care and leave it up to the app to make sure not to retrieve the endpoint for pages with hidden parents?
@klinzo do you have more insight or opinion on this?

klinzo: I guess a 404 would make sense. If a user just enters an url, we have no way of checking if that page (or its parents) exist without making these requests. So if your plan in the CMS is that pages with a non public parent should not be public, a 404 is the way to go imo.

timoludwig: Alright, thanks! Then I'd suggest the following:
```suggestion
    for ancestor in current_page.get_ancestors(include_self=False):
        public_translation = ancestor.get_public_translation(language_slug)
        if not public_translation:
            raise Http404("No Page matches the given url or id.")
        result.append(transform_page(public_translation))
```

klinzo: Sounds legit :+1:

timoludwig (approved):
Thanks! :100:

PR #1102, opened 2022-01-13
Fix bug when creating a new offer template

membralala:
### Short description

This PR fixes a tiny bug, where a wrong type was passed, when creating a new offer template as admin

### Proposed changes

- Change cleaned_post_data from ```dict``` to ```str```

### DIFF
--- a/integreat_cms/cms/views/offer_templates/offer_template_view.py
+++ b/integreat_cms/cms/views/offer_templates/offer_template_view.py
@@ -83,7 +83,7 @@ class OfferTemplateView(TemplateView):
             # Save form
             form.save()
             # Add the success message and redirect to the edit page
-            if offer_template_instance:
+            if not offer_template_instance:
                 messages.success(
                     request,
                     _('Offer template "{}" was successfully created').format(
@@ -98,5 +98,6 @@ class OfferTemplateView(TemplateView):
                 request,
                 _('Offer template "{}" was successfully saved').format(form.instance),
             )
+            form = OfferTemplateForm(instance=form.instance)
 
         return render(request, self.template_name, {"form": form, **self.base_context})


timoludwig (commented):
Thanks a lot for reporting this issue! :+1: 

The solution however isn't completely ideal I think, because the cleaned data should typically contain the submitted data converted into its desired type. So even though the JSON is submitted in a textarea and sent to the server as string, the Django representation of JSON is a Python datastructure like `dict`, `list` or `None` (see [documentation](https://docs.djangoproject.com/en/3.2/ref/forms/fields/#jsonfield)).

Instead, I suggest to have a look to the view and change `if offer_template_instance` to `if not offer_template_instance` in line [86](https://github.com/digitalfabrik/integreat-cms/blob/896f4e10ddc1081e93e400dfce7fa7cce47a2be1/integreat_cms/cms/views/offer_templates/offer_template_view.py#L86) of `offer_template_view.py`:

https://github.com/digitalfabrik/integreat-cms/blob/896f4e10ddc1081e93e400dfce7fa7cce47a2be1/integreat_cms/cms/views/offer_templates/offer_template_view.py#L83-L95

membralala:
Thanks for your feedback. Your point makes totally sense, however your solution would run into the same issue, when entering a not empty JSON in the form. I found that the issue does not show up when the form is re-initialized, so adding 
`form = OfferTemplateForm(instance=form.instance)` before rendering the template seems to fix it. Would this be a solution? Or do you have any better suggestions?

timoludwig:
> Thanks for your feedback. Your point makes totally sense, however your solution would run into the same issue, when entering a not empty JSON in the form. I found that the issue does not show up when the form is re-initialized, so adding `form = OfferTemplateForm(instance=form.instance)` before rendering the template seems to fix it. Would this be a solution? Or do you have any better suggestions?

I could not reproduce the issue with my change, can you tell me what values you entered exactly?
In theory, if you change `if offer_template_instance` to `if not offer_template_instance` it should do a `redirect` to the corresponding "edit form" (instead of remaining on the "creation form" - which doesn't make sense anyway because it would create another instance if it were submitted twice)... and after the redirect, the form is re-initialized anyway...

timoludwig (approved):
Thanks a lot! :tada:
I still don't have an idea what causes the issue, but this solves it for the moment :+1:

PR #1110, opened 2022-01-20
Fix region selection

timoludwig:
### Short description

Fix region selection

### Proposed changes

- Fix incorrect replacement of `request.region` in the region selection template
- Fix missing redirection in case an admin user opens the region selection by removing the original dashboard redirect view and handling the redirection directly in the region selection view


ulliholtgrave (approved):
Ok... :see_no_evil:

PR #1113, opened 2022-01-20
Improve performance for push notification

MizukiTemma:
### Short description

This PR tries to reduce SQL-queries for push notification page.

Related to #943 

### Proposed changes

-  collect all languages for which a translation exists before showing a push notification message in the list to avoid same/similar database calls at language flags (to choose edit icon or plus icon)
-
-

### DIFF
--- a/integreat_cms/cms/models/push_notifications/push_notification.py
+++ b/integreat_cms/cms/models/push_notifications/push_notification.py
@@ -1,5 +1,6 @@
 from django.conf import settings
 from django.db import models
+from django.utils.functional import cached_property
 from django.utils.translation import get_language, ugettext_lazy as _
 
 from ...constants.push_notifications import PN_MODES
@@ -51,13 +52,14 @@ class PushNotification(models.Model):
         ),
     )
 
-    @property
+    @cached_property
     def languages(self):
         """
-        This property returns a QuerySet of all :class:`~integreat_cms.cms.models.languages.language.Language` objects, to which an event
-        translation exists.
+        This property returns a QuerySet of all :class:`~integreat_cms.cms.models.languages.language.Language` objects,
+        to which a push notification translation exists.
 
-        :return: QuerySet of all :class:`~integreat_cms.cms.models.languages.language.Language` an event is translated into
+        :return: QuerySet of all :class:`~integreat_cms.cms.models.languages.language.Language` a push notification is
+                 translated into
         :rtype: ~django.db.models.query.QuerySet [ ~integreat_cms.cms.models.languages.language.Language ]
         """
         return Language.objects.filter(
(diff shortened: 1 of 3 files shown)


integreat_cms/cms/models/push_notifications/push_notification.py
```diff
@@ -50,6 +51,17 @@ class PushNotification(models.Model):
         ),
     )
 
+    @property
```
timoludwig, line 54:
```suggestion
    @cached_property
```
Can you cache this property?

integreat_cms/cms/models/push_notifications/push_notification.py
```diff
@@ -50,6 +51,17 @@ class PushNotification(models.Model):
...
+    def languages(self):
+        """
+        This property returns a QuerySet of all :class:`~integreat_cms.cms.models.languages.language.Language` objects, to which a POI
```
timoludwig, line 57:
```suggestion
        This property returns a QuerySet of all :class:`~integreat_cms.cms.models.languages.language.Language` objects, to which a push notification
```

integreat_cms/cms/models/push_notifications/push_notification.py
```diff
@@ -50,6 +51,17 @@ class PushNotification(models.Model):
...
+        translation exists.
+
+        :return: QuerySet of all :class:`~integreat_cms.cms.models.languages.language.Language` a POI is translated into
```
timoludwig, line 60:
```suggestion
        :return: QuerySet of all :class:`~integreat_cms.cms.models.languages.language.Language` a push notification is translated into
```

integreat_cms/cms/templates/push_notifications/push_notification_list_row.html
```diff
@@ -18,9 +19,7 @@
 				{% spaceless %}
 					{% for other_language in languages %}
 					<a href="{% url 'edit_push_notification' push_notification_id=push_notification.id region_slug=region.slug language_slug=other_language.slug %}">
-						{% with pnt=push_notification|translation:other_language %}
-						<i data-feather="{% if pnt %}edit-2{% else %}plus{% endif %}" class="text-gray-800"></i>
-						{% endwith %}
+						<i data-feather="{% if other_language in existing_languages %}edit-2{% else %}plus{% endif %}" class="text-gray-800"></i>
```
timoludwig, line 22:
```suggestion
						<i data-feather="{% if other_language in push_notification.languages %}edit-2{% else %}plus{% endif %}" class="text-gray-800"></i>
```
Once the property is cached, there is no reason anymore to wrap it into a `{% with %}` tag.

timoludwig (commented):
Thanks! :+1:
I just tested this a little bit with ~100 push notification objects, and while it reduces the number of SQL queries from ~500 to ~200, I think it is still not enough. Ideally, we should get to a constant number of queries independently from the numbers of entries, so can you try to prefetch the languages and translations for each push notification in one query? If you're unsure how to do it, we can also merge this PR as an preliminary step and you can move forward to another issue.

timoludwig (approved):
Even though I think there is still potential for further reductions, let's merge this PR for the time being.

Issue #1112, opened 2022-01-20
Version slider UI has a wrong alignment and Z-Index


ulliholtgrave:
### Describe the Bug

At the moment the slider item is behind the background which results in a very small drag & drop zone.

![grafik](https://user-images.githubusercontent.com/15608678/150324495-6237a46e-8cf0-4284-8439-6dcab5846ed0.png)

### Expected Behavior

The grey background point should be the same size then the blue point and should be in front of it.

Issue #1111, opened 2022-01-20
Add new language tabs for imprint form


timoludwig:
### Motivation

We recently changed the design of the language tabs in the content forms, but we forgot to change them in the imprint form

### Proposed Solution

copy the solution from pages, events and POIs to the imprint form as well

PR #1114, opened 2022-01-21
Fix UI issues and add language switcher to imprint

ulliholtgrave:
### Short description

This PR fixes the UI of the versioning view, makes some small changes to the page form and adds the language switcher to the imprint form.

### Proposed changes

- Change CSS of versioning slider
- Add language switcher to imprint form and add check if content has no id
- Add some small UI tweaks to the language switcher

### Resolved issues

Fixes: #1112 #1111

### DIFF
--- a/integreat_cms/cms/templates/content_form_language_switcher.html
+++ b/integreat_cms/cms/templates/content_form_language_switcher.html
@@ -4,12 +4,12 @@
 {% load i18n %}
 
 <div id="language-switcher" class="flex flex-col z-10 text-lg">
-    <span id="language-switcher-current" href="{% build_url target=target region_slug=request.region.slug language_slug=language.slug content_field=content_field content_id=content_id %}" class="font-bold flex justify-between bg-white py-2 px-4 rounded shadow hover:bg-gray-200 border-2 border-white hover:border-gray-200">
-        Weitere Sprachen
+    <span id="language-switcher-current" class="font-bold flex justify-between bg-white py-2 px-4 rounded shadow hover:bg-gray-200 border-2 border-white hover:border-gray-200">
+        {% trans 'More Languages' %}
     </span>
     <div id="language-switcher-list" class="h-0 hidden">
         {% for other_language in languages|sort_languages:language|slice:"4:" %}
-            <a href="{% build_url target=target region_slug=request.region.slug language_slug=other_language.slug content_field=content_field content_id=content_id %}" class="block flex justify-between whitespace-nowrap {% if forloop.last %} rounded-b {% endif %} shadow px-4 py-3 text-gray-800 hover:bg-gray-200 bg-white">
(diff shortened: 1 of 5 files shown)


integreat_cms/cms/templates/content_form_language_switcher.html
```diff
@@ -4,12 +4,12 @@
 {% load i18n %}
 
 <div id="language-switcher" class="flex flex-col z-10 text-lg">
-    <span id="language-switcher-current" href="{% build_url target=target region_slug=request.region.slug language_slug=language.slug content_field=content_field content_id=content_id %}" class="font-bold flex justify-between bg-white py-2 px-4 rounded shadow hover:bg-gray-200 border-2 border-white hover:border-gray-200">
-        Weitere Sprachen
+    <span id="language-switcher-current" class="font-bold flex justify-between bg-white py-2 px-4 rounded shadow hover:bg-gray-200 border-2 border-white hover:border-gray-200">
```
timoludwig, line 7:
```suggestion
    <span id="language-switcher-current" class="font-bold flex justify-between bg-white py-2 px-4 rounded shadow border-2 border-white">
```
I'm not sure about this, but if this area itself isn't a link anymore and clicking on it does nothing, maybe we should remove the hover effects and the pointer? But I don't know what's best practice for UI elements which trigger an action on _hovering_, not on _clicking_...

integreat_cms/cms/templates/content_form_language_switcher.html
```diff
@@ -4,12 +4,12 @@
...
+        {% trans 'More Languages' %}
     </span>
     <div id="language-switcher-list" class="h-0 hidden">
         {% for other_language in languages|sort_languages:language|slice:"4:" %}
-            <a href="{% build_url target=target region_slug=request.region.slug language_slug=other_language.slug content_field=content_field content_id=content_id %}" class="block flex justify-between whitespace-nowrap {% if forloop.last %} rounded-b {% endif %} shadow px-4 py-3 text-gray-800 hover:bg-gray-200 bg-white">
+            <a href="{% if target != 'edit_imprint' %}{% build_url target=target region_slug=request.region.slug language_slug=other_language.slug content_field=content_field content_id=content_id %}{% else %} {% url 'edit_imprint' region_slug=request.region.slug language_slug=other_language.slug %} {% endif %}" class="block flex justify-between whitespace-nowrap {% if forloop.last %} rounded-b {% endif %} shadow px-4 py-3 text-gray-800 hover:bg-gray-200 bg-white">
```
timoludwig, line 12: I guess this condition could also be handled in the `build_url` template tag by passing the `content_field` to the reverse function only when it's not `None`...

timoludwig (approved):
Thanks a lot, looks good to me! :+1:
Just some minor details, but they are optional I think.

ulliholtgrave:
All changes are included in #1116

PR #1115, opened 2022-01-21
Show package version at bottom navigation sidebar

timoludwig:
### Short description

When we have different instances of CMS servers deployed (production, cms-test, cms-dev), it's probably handy to see the installed version of the cms at a glance, especially for bug reports etc.
I hope the placement at the bottom of the navigation bar is not too obtrusive from a UI/UX perspective...

### Proposed changes

- Show version at bottom navigation sidebar


ulliholtgrave (approved):
Looks nice :)

PR #1116, opened 2022-01-23
Improve forms

timoludwig:
### Short description

This PR adds various improvements for our forms

### Proposed changes

- Make use of Django's inbuilt `FormViews`
    - Use consistent naming scheme for form views and templates
    - Use Django's `CreateView` and `UpdateView` for the follwing form views:
        - `LanguageForm`
        - `OfferTemplateForm`
        - `OrganizationForm`
        - `RegionForm`
- Always redirect after successful form submission
    - Django expects that a form view is reloaded after a successful form submission.
      Also, this prevents accidental re-submission when refreshing the form view.
- Improve performance of content forms
    - Optimize calculation of translation states in content forms
    - Use generic template for language tabs in forms
    - Add new language tabs to imprint form

### Resolved issues

This is part of #943
Fixes: #1111 
Fixes: #1112 

### Additional Context

@ulliholtgrave To avoid conflicts with your PR #1114 I included your changes to the page form and imprint form in my PR as well.


ulliholtgrave (approved):
Very nice! I tested it locally with about 400 pages and I can see a significant performance improvement! :+1: 
I also like the way you changed my initial UI proposal for the language switcher :)

svenseeberg:
Let's merge and continue testing in the test systems :)

timoludwig:
Alright! Unfortunately, this doesn't reduce the queries to a constant number yet, because calculating the labels for the parent field retrieves the ancestors for each other page and since Django's `ModelChoiceField` expects normal querysets, I cannot pass the cached tree structure here. My guess is that this could be fixed by subclassing this field and providing a custom widget, but I can also implement this in a subsequent PR.

PR #1120, opened 2022-01-23
Streamline API documentation and remove dummy function

ulliholtgrave:
### Short description

As we are about to release our first version, I aligned the documentation of our API functions for a better representation in our sphinx docu and removed some dummy lines of code.


timoludwig (approved):
Thanks! :+1:

PR #1121, opened 2022-01-24
Add region name to MatomoException token message

svenseeberg:
### Short description
When access to Matomo is denied, a mail with the error message is dispatched. But due to missing region name/id it is not possible to fix the prohblem.

### Proposed changes
Add the region name to the error message.


ulliholtgrave (approved):
Makes sense :+1:

PR #1122, opened 2022-01-24
Improve forms

timoludwig:
### Short description

Some further improvements for forms

### Proposed changes

- Reduce number of SQL queries in page form
- Fix resetting to default value in page order table
- Fix permission checks in generic form views


ulliholtgrave (approved):
This fixes the issue with the permissions and I can see a reduction of queries on my local system :+1:

Issue #1123, opened 2022-01-24
Privilege escalation for staff users


timoludwig:
### Describe the Bug

At the moment, staff users can declare themselves admins by setting the corresponding checkbox in the admin user form.

### Steps to Reproduce

1. Create a user who is staff, but not superuser
2. Go to admin user forrm of this user
3. Set checkbox "is administrator"
4. Save form

### Expected Behavior

I expect two things:
1. The field should be hidden if the request user isn't superuser (client-side UI improvement)
2. There should be an error if the user manages to set the hidden field nonetheless (server-side security check)

### Actual Behavior

The form is saved and the user becomes a superuser

PR #1127, opened 2022-01-24
Fix superuser permissions mechanic

membralala:
### Short description

This PR should fix an issue, where a user without superuser permissions was able to change his/her superuser state. 

### Proposed changes

- Change the user form template, so that the superuser option is hidden, when the user isn't a superuser him/herself
- Change the user form view, so that a server-side error message is rendered in the same case
- Change 'superuser' translations from 'Superadmin' to 'Admin'

### Resolved issues

Fixes:  #1123

### DIFF
--- a/integreat_cms/cms/templates/users/user_form.html
+++ b/integreat_cms/cms/templates/users/user_form.html
@@ -80,7 +80,7 @@
 			</div>
 			<div class="px-4 pb-4 divide-y divide-gray-200 space-y-4">
                 <div class="divide-y divide-gray-200">
-                    <div>
+                    <div class="{% if not request.user.is_superuser %}hidden{% endif %}">
                         {% render_field user_form.is_superuser %}
                         <label for="{{ user_form.is_superuser.id_for_label }}">{{ user_form.is_superuser.label }}</label>
                         <div class="help-text">{{ user_form.is_superuser.help_text }}</div>
--- a/integreat_cms/cms/views/users/user_form_view.py
+++ b/integreat_cms/cms/views/users/user_form_view.py
@@ -100,6 +100,11 @@ class UserFormView(TemplateView):
                     "A user has to be either staff/superuser or needs to be restricted to at least one region."
                 ),
             )
+        elif not request.user.is_superuser and "is_superuser" in user_form.changed_data:
+            messages.error(
+                request,
+                _("Superuser permissions need to be set by another superuser."),
+            )
         elif not user_form.has_changed():
             # Add "no changes" messages
             messages.info(request, _("No changes made"))
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -5620,7 +5620,7 @@ msgstr "Feedback wurde erfolgreich gelöscht"
 #: cms/views/settings/user_settings_view.py:94
 #: cms/views/settings/user_settings_view.py:111
 #: cms/views/users/region_user_form_view.py:98
-#: cms/views/users/user_form_view.py:105
+#: cms/views/users/user_form_view.py:110
 msgid "No changes made"
 msgstr "Keine Änderungen vorgenommen"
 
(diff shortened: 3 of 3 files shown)


membralala:
When creating the PR I just realised, that I missed something. So this needs a little more work.

integreat_cms/cms/templates/users/user_form.html
```diff
@@ -80,7 +80,7 @@ <h3 class="heading font-bold text-white"><i data-feather="key" class="mr-2"></i>
 			</div>
 			<div class="px-4 pb-4 divide-y divide-gray-200 space-y-4">
                 <div class="divide-y divide-gray-200">
-                    <div>
+                    <div class="{% if not current_user.is_superuser %}hidden{% endif %}">
```
timoludwig, line 83:
```suggestion
                    <div class="{% if not request.user.is_superuser %}hidden{% endif %}">
```
The `request` object is available in all templates because of the [request context processor](https://docs.djangoproject.com/en/3.2/ref/templates/api/#django-template-context-processors-request)

integreat_cms/cms/views/users/user_form_view.py
```diff
@@ -47,6 +47,7 @@ def get(self, request, *args, **kwargs):
         """
 
         user = get_user_model().objects.filter(id=kwargs.get("user_id")).first()
+        current_user = get_user_model().objects.filter(id=request.user.id).first()
```
timoludwig, line 50:
```suggestion
        current_user = get_user_model().objects.filter(id=request.user.id).first()
```
This is not required. `request.user` is already the user from the database, so no reason to query it again.

integreat_cms/cms/views/users/user_form_view.py
```diff
@@ -59,6 +60,7 @@ def get(self, request, *args, **kwargs):
             {
                 **self.base_context,
                 "user_form": user_form,
+                "current_user": current_user,
```
timoludwig, line 63:
```suggestion
```

integreat_cms/cms/views/users/user_form_view.py
```diff
@@ -83,6 +85,8 @@ def post(self, request, *args, **kwargs):
             get_user_model().objects.filter(id=kwargs.get("user_id")).first()
         )
 
+        current_user = get_user_model().objects.filter(id=request.user.id).first()
+
```
timoludwig, line 89:
```suggestion
```

integreat_cms/cms/views/users/user_form_view.py
```diff
@@ -100,6 +104,11 @@ def post(self, request, *args, **kwargs):
                     "A user has to be either staff/superuser or needs to be restricted to at least one region."
                 ),
             )
+        elif not current_user.is_superuser:
```
timoludwig, line 107:
```suggestion
        elif not request.user.is_superuser:
```

integreat_cms/cms/views/users/user_form_view.py
```diff
@@ -140,5 +149,6 @@ def post(self, request, *args, **kwargs):
             {
                 **self.base_context,
                 "user_form": user_form,
+                "current_user": current_user,
```
timoludwig, line 152:
```suggestion
```

timoludwig (commented):
Thanks! It still doesn't work as expected, but here are a few first hints to work with :wink:

membralala:
Thanks for your feedback! Apart from the very good ```request.user``` hint, in which situation does the behaviour does not match your expectations?

integreat_cms/cms/views/users/user_form_view.py
timoludwig, line 107:
```suggestion
        elif not request.user.is_superuser and "is_superuser" in user_form.changed_data:
```
The problem is that with your suggestion, non-superuser cannot save the form at all, which solves the privilege escalation but unfortunately we need normal staff users to be able to change other user fields.
Thus, I suggest to only raise the error when the non-superuser tries to change the `is_superuser`-field.

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -6323,32 +6323,38 @@ msgstr "Benutzer {} wurde erfolgreich von dieser Region entfernt."
...
 "least one region."
 msgstr ""
 "Ein Benutzer muss entweder Mitarbeiter/Super-Administrator sein, oder "
 "mindestens einer Region zugewiesen sein."
 
+#: cms/views/users/user_form_view.py:110
+msgid "Superuser permissions need to be set by another superuser."
+msgstr ""
+"Super-Administratorrechte müssen von einem anderen Superadministrator "
+"vergeben werden."
```
timoludwig, line 6356:
```suggestion
"Administratorrechte müssen von einem anderen Administrator vergeben werden."
```
Since we only have one level of admins, I'd go with "superuser" in English and "Administrator" in German. ("Super-Administrator" sounds a bit like there are different, normal "Administrators" out there...)
I'm aware that there are inconsistencies in our existing translations, so if you want to, you can also fix these, but they are not necessarily part of this issue.

integreat_cms/cms/views/users/user_form_view.py
membralala: Yeah, of course you're right...

timoludwig (approved):
Thanks, looks good now! :tada:

PR #1129, opened 2022-01-25
Fix permission check in user search

timoludwig:
### Short description

Just a small permission check problem which was caused by the change from the default Django user model to our own...


ulliholtgrave (approved):
Fixes it :+1:

PR #1133, opened 2022-01-25
Fix matomo API token check

timoludwig:
### Short description

For legacy reasons, our API tokens don't have consistent permissions levels.
Thus, we can just check for 'at least view access' to determine the correct matomo id.

### Proposed changes

- Use `getSitesIdWithAtLeastViewAccess` instead of `getSitesIdWithAdminAccess` for Matomo id check


ulliholtgrave (approved):
Nice that you found that so fast :+1:

Issue #1103, opened 2022-01-17
Add dropdown in order to execute bulk actions in events


JoeyStk:
### Motivation
Multiple posts can be selected to execute bulk actions. How ever this isn't possible for events and pois, which is necessary in order to automatically translate selected events and pois.

### Proposed Solution
Add the dropdown from posts to events and pois using the existing bulk action template. This breaks down into three smaller tasks: 
- [ ] add the bulk action template
- [ ] add the dropdown to select multiple events
- [ ] add an option to auto translate the selected events

### Additional Context
This is part of the project for Johannes' final exam.

PR #1135, opened 2022-01-25
Add automatic translations via DeepL API

JoeyStk:
### Short description
This is my first draft for the DeepL API. This is for my final exam.

### Proposed changes
- [x] Check if the selected language is supported by deepl
- [x] Get the selected post, which was selected to be translated
- [x] Create a connection to DeepL and translate
- [x] Set the translation, that DeepL returned, to the database

### Resolved issues
So far I only implemented the first step. However this already raised a few questions for me, so Giuliano proposed to me to upload what I had as a draft, so you can have a look at it already.
The API is so far only implemented for the pois. At a later point this will be implemented for the events in milestone 1.1.
I want to "finish" the API around the 11th of february, so I still have enough time to write the required documentation for the IHK. 

First, let's get to the general questions:
- Do you think the functions I outlined make sense, or do you think there is a redundant function or one that's missing?
- With the structure of the file I tried to stay close to the gvz_api, however I'm not sure if the structure of the files make sense, or if there is something I can improve 

Secondly, the more specific questions:
When I implemented the function "check_availability" I was struggling with a decision I had to make.
So far I hardcoded all languages, that are supported by DeepL inside a list. This, of course, is bad practice.
However, I realized there was a larger issue at hand, and I couldn't decide on a better way. There is an option inside the DeepL library that can check, if a language is supported. This works via python, thus we could only catch the exception and give the user feedback through an error message. To me this isn't a great solution (at least not in the long run).
I think it would be better to communicate through the ui/ux if a language can be translated automatically via DeepL. I was wondering, if @ulliholtgrave maybe had an idea for this, since you are in close contact with Nikola.
My solution would be to implement via python & error message temporarily, so my project is complete. But later improve this for the users. 

Fixes: #1103

### DIFF
--- a/integreat_cms/deepl_api/utils.py
+++ b/integreat_cms/deepl_api/utils.py
@@ -1,0 +1,146 @@
+import logging
+import deepl
+
+from django.conf import settings
+from django.contrib import messages
+from django.utils.translation import ugettext as _
+
+logger = logging.getLogger(__name__)
+
+
+class DeepLApi:
+    """
+    DeepL API to auto translate selected posts.
+    """
+
+    def __init__(self):
+        """
+        Initialize the DeepL client
+        """
+        self.translator = deepl.Translator(settings.DEEPL_AUTH_KEY)
+
+    def check_availability(self, request, language_slug):
+        """
+        This function checks, if the selected language is supported by DeepL
+
+        :param request: request that was sent to the server
+        :type request: ~django.http.HttpRequest
+
+        :param language_slug: current language slug
+        :type language_slug: str
+
+        :return: true or false
+        :rtype: bool
+        """
+        supported_source_languages = [
+            source_language.code.lower()
+            for source_language in self.translator.get_source_languages()
+        ]
+        supported_target_languages = [
+            target_languages.code.lower()[:2]
+            for target_languages in self.translator.get_target_languages()
+        ]
+        source_language = request.region.get_source_language(language_slug)
+        return (
+            source_language
+            and source_language.slug in supported_source_languages
+            and language_slug in supported_target_languages
+        )
+
+    def deepl_translation(self, request, content_objects, language_slug, form_class):
+        """
+        This functions gets the translation from DeepL
+
+        :param request: passed request
+        :type request: ~django.http.HttpRequest
+
+        :param content_objects: passed content objects
+        :type content_objects: ~django.db.models.query.QuerySet [ ~integreat_cms.cms.models.abstract_content_model.AbstractContentModel ]
+
+        :param language_slug: current GUI language slug
+        :type language_slug: str
+
+        :param form_class: passed Form class of content type
+        :type form_class: ~integreat_cms.cms.forms.custom_content_model_form.CustomContentModelForm
+        """
+        # Get target language
+        target_language = request.region.get_language_or_404(language_slug)
+        source_language = request.region.get_source_language(language_slug)
+        for content_object in content_objects:
+            source_translation = content_object.get_translation(source_language.slug)
+            if not source_translation:
+                messages.error(
+                    request,
+                    _('No source translation could be found for {} "{}".').format(
+                        type(content_object)._meta.verbose_name.title(),
+                        content_object.best_translation.title,
+                    ),
+                )
+                continue
+            existing_target_translation = content_object.get_translation(
+                target_language.slug
+            )
+            # For some languages, the DeepL client expects the BCP tag instead of the short language code
+            if target_language.slug in ("en", "pt"):
+                target_language_key = target_language.bcp47_tag
+            else:
+                target_language_key = target_language.slug
+            target_title = self.translator.translate_text(
+                source_translation.title,
+                source_lang=source_language.slug,
+                target_lang=target_language_key,
+            )
+            data = {
+                "title": target_title,
+                "status": existing_target_translation.status
+                if existing_target_translation
+                else source_translation.status,
+            }
+            if source_translation.content:
+                data["content"] = self.translator.translate_text(
+                    source_translation.content,
+                    source_lang=source_language.slug,
+                    target_lang=target_language_key,
+                )
+            # for pois adds a short description
+            if hasattr(source_translation, "short_description"):
+                data["short_description"] = self.translator.translate_text(
+                    source_translation.short_description,
+                    source_lang=source_language.slug,
+                    target_lang=target_language_key,
+                )
+            content_translation_form = form_class(
+                data=data,
+                instance=existing_target_translation,
+                additional_instance_attributes={
+                    "creator": request.user,
+                    "language": target_language,
+                    source_translation.foreign_field(): content_object,
+                },
+            )
+            # Validate event translation
+            if content_translation_form.is_valid():
+                content_translation_form.save()
+                logger.debug(
+                    "Successfully translated for: %r", content_translation_form.instance
+                )
+                messages.success(
+                    request,
+                    _('{} "{}" has been successfully translated.').format(
+                        type(content_object)._meta.verbose_name.title(),
+                        source_translation.title,
+                    ),
+                )
+            else:
+                logger.error(
+                    "Automatic translation for %r could not be created because of %r",
+                    content_object,
+                    content_translation_form.errors,
+                )
+                messages.error(
+                    request,
+                    _('{} "{}" could not be automatically translated.').format(
+                        type(content_object)._meta.verbose_name.title(),
+                        source_translation.title,
+                    ),
+                )
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-03-20 18:58+0000\n"
+"POT-Creation-Date: 2022-03-23 12:23+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: Integreat <info@integreat-app.de>\n"
 "Language-Team: Integreat <info@integreat-app.de>\n"
@@ -3121,7 +3121,7 @@ msgstr "Feedback"
 
 #: cms/templates/_base.html:181
 #: cms/templates/push_notifications/push_notification_list.html:9
-#: core/settings.py:712
+#: core/settings.py:716
 msgid "News"
 msgstr "Nachrichten"
 
@@ -5699,19 +5699,21 @@ msgstr ""
 "Adresse eingegeben haben, mit der Sie sich registriert haben, und überprüfen "
 "Sie Ihren Spam-Ordner."
 
-#: cms/views/bulk_action_views.py:123
-msgid "Automatic translations are not fully implemented yet"
-msgstr "Automatische Übersetzungen sind noch nicht vollständig implementiert"
-
-#: cms/views/bulk_action_views.py:126
+#: cms/views/bulk_action_views.py:124
 msgid "Automatic translations are disabled"
 msgstr "Automatische Übersetzungen sind deaktiviert"
 
-#: cms/views/bulk_action_views.py:164
+#: cms/views/bulk_action_views.py:137
+msgid "This language is not supported by DeepL. Please try another language"
+msgstr ""
+"Diese Sprache wird von DeepL nicht unterstützt. Bitte versuchen Sie eine "
+"andere Sprache"
+
+#: cms/views/bulk_action_views.py:177
 msgid "The selected {} were successfully archived"
 msgstr "Die ausgewählten {} wurden erfolgreich archiviert"
 
-#: cms/views/bulk_action_views.py:203
+#: cms/views/bulk_action_views.py:216
 msgid "The selected {} were successfully restored"
 msgstr "Die ausgewählten {} wurden erfolgreich wiederhergestellt"
 
@@ -6599,6 +6601,18 @@ msgid "Superuser permissions need to be set by another superuser."
 msgstr ""
 "Administratorrechte müssen von einem anderen Administrator vergeben werden."
 
+#: deepl_api/utils.py:74
+msgid "No source translation could be found for {} \"{}\"."
+msgstr "Es konnte kein Quelltext für {} \"{}\" gefunden werden."
+
+#: deepl_api/utils.py:129
+msgid "{} \"{}\" has been successfully translated."
+msgstr "{} \"{}\" wurde erfolgreich übersetzt"
+
+#: deepl_api/utils.py:142
+msgid "{} \"{}\" could not be automatically translated."
+msgstr "{} \"{}\" konnte nicht automatisch übersetzt werden."
+
 #: xliff/utils.py:152
 msgid ""
 "Page {} does not have a source translation in {} and therefore cannot be "
@@ -6650,6 +6664,10 @@ msgstr ""
 "Diese Seite konnte nicht importiert werden, da sie zu einer anderen Region "
 "gehört ({})."
 
+#~ msgid "Automatic translations are not fully implemented yet"
+#~ msgstr ""
+#~ "Automatische Übersetzungen sind noch nicht vollständig implementiert"
+
 #~ msgid "Filetype:"
 #~ msgstr "Dateityp:"
 
@@ -6662,6 +6680,9 @@ msgstr ""
 
 #~ msgid "Export XLIFF for translation to"
 #~ msgstr "Exportiere XLIFF für Übersetzung nach"
+
+#~ msgid "POI has been successfully translated"
+#~ msgstr "POI wurde erfolgreich übersetzt."
 
 #~ msgid "Access token to update the page content"
 #~ msgstr "Zugangs-Token um Seiten-Inhalte zu aktualisieren"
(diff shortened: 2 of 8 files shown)


integreat_cms/deepl_api/utils.py
```diff
@@ -0,0 +1,87 @@
...
+            "sv",
+            "zh",
+        ]
+        if language_slug in deepl_langs:
+            return language_slug in deepl_langs
+        else:
+            messages.warning(
+                request,
+                _("This language isn't supported by DeepL. Please try another one"),
+            )
```
timoludwig, line 57:
```suggestion
        supported_source_languages = [
            source_language.code.lower()
            for source_language in self.translator.get_source_languages()
        ]
        supported_target_languages = [
            target_languages.code.lower()[:2]
            for target_languages in self.translator.get_target_languages()
        ]
        return (
            request.region.default_language.slug in supported_source_languages
            and language_slug in supported_target_languages
        )
```
I'd probably go with something along the lines of this - I think requesting the supported languages via the deepl api client is easy enough that we can directly implement it like this without the intermediary solution of defining constants.

integreat_cms/deepl_api/utils.py
```diff
@@ -0,0 +1,87 @@
...
+        default_language = region.default_language.slug
+        print(default_language)
+
+    def get_post(self, poi_ids):
+        """
+        # zu implementieren am 21.1.
+        # nachdem alle Infos zum Beitrag übergeben werden,
+        # wird nur der tatsächliche Inhalt des Posts geholt
+        """
+        print(poi_ids)
```
timoludwig, line 73:
```suggestion
```
This method is probably also not necessary - you can just access `poi.content` to get the content of a POI without an additional function.

integreat_cms/deepl_api/utils.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def get_translation(self, translator, post, target_language):
+        """
+        # zu implementieren am 24.1.
+        This function calls the translate-function of DeepL and
+        """
+        translator = deepl.Translator(os.getenv("de7e6874-aebd-0823-a278-c281a3760c64"))
```
timoludwig, line 80:
```suggestion
        translator = deepl.Translator(settings.DEEPL_AUTH_KEY)
```
When handling sensible data, it's usually a good idea to read the corresponding usage instructions:
> Create a Translator object providing your DeepL API authentication key.
> To avoid writing your key in source code, you can set it in an environment
> variable DEEPL_AUTH_KEY, then read the variable in your Python code:
> `translator = deepl.Translator(os.getenv("DEEPL_AUTH_KEY"))`

The reason not to write your secrets into source code is pretty simple: You start testing a few things and after that you forget about it and commit your source code to git - where the secret is now published... No worries, I already invalidated the token, so don't beat yourself up about it, but the correct workflow would be:
1. Add a new setting for this in `integreat_cms/core/settings.py`, e.g.: `DEEPL_AUTH_KEY = os.environ.get("INTEGREAT_CMS_DEEPL_AUTH_KEY")`
1. Read the secret into a environment variable, e.g. via the configuration file `/etc/integreat.ini` or via the command line: `read -rs INTEGREAT_CMS_DEEPL_AUTH_KEY && export INTEGREAT_CMS_DEEPL_AUTH_KEY` and then start the dev server which can now read this environment variable which is separated from the source code in the repository
2. In your DeepL api client, use `from django.conf import settings` and then you can use `deepl.Translator(settings.DEEPL_AUTH_KEY)` to initialize the translator object.

integreat_cms/deepl_api/utils.py
```diff
@@ -0,0 +1,87 @@
...
+        result = translator.translate_text(post, target_lang=target_language)
+        return result
+
+    def set_translation_in_database(self, data):
+        """
+        # zu implementieren am 25.1.
+        """
```
timoludwig, line 87: I'm not sure whether you need a separate method for saving elements to the database - normally, you can just save Django models directly via calling their `save()`-method, e.g.:
```
translated_content = self.get_translation(source_translation, target_language)
target_translation = POITranslation.create(language=target_language, content=translated_content)
target_translation.save()
```

timoludwig (commented):
Thanks! In general, this looks very sensible. See my comments below where I see room for improvement.

In addition to that, I would suggest you add an `__init__()`-method which should initialize the translator object and assign it to `self.translator`. This method is automatically executed when the Api client is created with `deepl = DeepLApi()`. Then, all other methods don't need to initialize the translator again but can just write `self.translator. ...` when making API calls.
For the design of the translation methods, you could e.g. have a `translate` method which takes a queryset of objects and is called from the view via `deepl.translate(pois)` which then iterates over all given objects and calls `get_translation` and saves them to the database.

Regarding your specific question: Yes, your approach sounds good, I think it's fine to just show an error at the moment if a language isn't supported. However, I think it would not be too much work to fix this already in the bulk action select items - we could just call the `check_availability()`-method from the content views that embed the bulk action select input and only show the "Automatic translation" option for those languages that are supported. Or show them for all languages, but make them disabled for unsupported languages and show a tool-tip that explains why the automatic translation isn't possible for this language. Anyway, this is optional stuff you could think about if you have time later on.

JoeyStk:
First of all, thank you for reviewing my PR. I still have two thoughts on this PR: 
- at the moment there is a lot of redundant code in the event_action.py and the poi_action.py files. So I was wondering, if the degree of redundancy is still okay, or if it is too much. If so I think I could move the code to another function and then just call it inside the event_action and poi_action. 
- I'm not sure how robust the API is, at this point. Do you have concerns about this, and ideas what kind of exceptions would make sense, and which ones aren't necessary? 
Thank you for your feedback!

timoludwig (commented):
Thanks a lot! :heart: 
Really cool feature, and works very well for all languages that DeepL supports.

A few general comments:
- Since the feature also supports translating existing versions, I would adapt the text in the bulk action from "Create missing translations automatically" to "Automatically translate to <target_language>"
- ~Since the supported languages will most likely not change very frequently, it's probably a bit too much overhead to call the DeepL API every time for this. Maybe we could retrieve the supported languages initially during the server start similar to the [GVZ API](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/gvz_api/apps.py)? And then, the supported languages could e.g. be retrieved via `apps.get_app_config("deepl_api").supported_languages` or similar. This would also make it easier to disable the bulk action when the language is not supported.~ Edit: Not relevant for this PR

And regarding your questions:

> at the moment there is a lot of redundant code in the event_action.py and the poi_action.py files. So I was wondering, if the degree of redundancy is still okay, or if it is too much. If so I think I could move the code to another function and then just call it inside the event_action and poi_action.

Good point, I also noticed this during my review and only annotated the view for events, please also consider the same comments to be valid for POIs. Now that I see that the code is indeed very similar and quite a few lines, I guess outsourcing this to another function in the deepl utils would make sense. You would need to pass the object itself, but also the form class associated with that object.

> I'm not sure how robust the API is, at this point. Do you have concerns about this, and ideas what kind of exceptions would make sense, and which ones aren't necessary?

From what I experienced during my tests, I don't see any performance issues. At some point we could test what happens when e.g. 100 events are translated at the same time, but I don't think this is a common use case and for now the implementation is fine as it is I guess.

integreat_cms/deepl_api/utils.py
```diff
@@ -0,0 +1,161 @@
...
+        :param content_label: passed content type, either event or poi
+        :type content_label: str
+        """
+        # Get target language
+        target_language = request.region.get_language_or_404(language_slug)
+        source_language = request.region.get_source_language(language_slug)
+        for content_translation in content_translations:
+            source_translation = content_translation.get_translation(
+                source_language.slug
+            )
```
timoludwig, line 96: If the source translation does not exist, I get the following error:
```
AttributeError at /augsburg/events/en/auto-translate/

'NoneType' object has no attribute 'title'
```
So I suggest to show a warning instead.

integreat_cms/deepl_api/utils.py
```diff
@@ -0,0 +1,161 @@
...
+                else source_translation.status,
+            }
+            if source_translation.content:
+                data["content"] = self.get_translation(
+                    source_translation.content,
+                    source_language.slug,
+                    target_language_key,
+                )
+            # for pois adds a short description
+            if content_label == "poi":
```
timoludwig, line 123:
```suggestion
            if hasattr(source_translation, "short_description"):
```
I would just check whether the source translation has the attribute `short_description` instead of hard-coding the content label. This way, the short description is automatically translated when it's added to other models in the future.

integreat_cms/deepl_api/utils.py
```diff
@@ -0,0 +1,161 @@
...
+                    source_language.slug,
+                    target_language_key,
+                )
+            content_translation_form = form_class(
+                data=data,
+                instance=existing_target_translation,
+                additional_instance_attributes={
+                    "creator": request.user,
+                    "language": target_language,
+                    content_label: content_translation,
```
timoludwig, line 135:
```suggestion
                    source_translation.foreign_field(): content_object,
```
The content label is already stored in the translation model, so no need to pass it to this function.

integreat_cms/deepl_api/utils.py
```diff
@@ -0,0 +1,161 @@
...
+        :return: translated post in target language
+        :rtype: str
+        """
+        translated_content = self.translator.translate_text(
+            content, source_lang=source_language_slug, target_lang=target_language_key
+        )
+        return translated_content
+
+    def deepl_translation(
+        self, request, content_translations, language_slug, form_class, content_label
```
timoludwig, line 70:
```suggestion
        self, request, content_objects, language_slug, form_class
```
The `content_label` can be derived from the translation objects and `content_translations` isn't an ideal variable name because it contains the content objects, not the translation objects.

integreat_cms/deepl_api/utils.py
```diff
@@ -0,0 +1,161 @@
...
+        :type form_class: ~integreat_cms.cms.forms.custom_content_model_form.CustomContentModelForm
+
+        :param content_label: passed content type, either event or poi
+        :type content_label: str
+        """
+        # Get target language
+        target_language = request.region.get_language_or_404(language_slug)
+        source_language = request.region.get_source_language(language_slug)
+        for content_translation in content_translations:
+            source_translation = content_translation.get_translation(
```
timoludwig, line 94:
```suggestion
        for content_object in content_objects:
            source_translation = content_object.get_translation(
```

integreat_cms/deepl_api/utils.py
```diff
@@ -0,0 +1,161 @@
...
+        self, request, content_translations, language_slug, form_class, content_label
+    ):
+        """
+        This functions gets the translation from DeepL
+
+        :param request: passed request
+        :type request: ~django.http.HttpRequest
+
+        :param content_translations: passed content
+        :type content_translations: ~django.db.models.query.QuerySet [ ~integreat_cms.cms.models.abstract_content_translation.AbstractContentTranslation ]
```
timoludwig, line 79:
```suggestion
        :param content_objects: passed content objects
        :type content_objects: ~django.db.models.query.QuerySet [ ~integreat_cms.cms.models.abstract_content_model.AbstractContentModel ]
```
This param contains the content objects, not the translation objects

integreat_cms/deepl_api/utils.py
```diff
@@ -0,0 +1,161 @@
...
+        :type content_label: str
+        """
+        # Get target language
+        target_language = request.region.get_language_or_404(language_slug)
+        source_language = request.region.get_source_language(language_slug)
+        for content_translation in content_translations:
+            source_translation = content_translation.get_translation(
+                source_language.slug
+            )
+            existing_target_translation = content_translation.get_translation(
```
timoludwig, line 97:
```suggestion
            existing_target_translation = content_object.get_translation(
```

integreat_cms/deepl_api/utils.py
```diff
@@ -0,0 +1,161 @@
...
+                messages.success(
+                    request,
+                    _('Event "{}" has been successfully translated.').format(
+                        source_translation.title
+                    ),
+                )
+            else:
+                logger.error(
+                    "Automatic translation for %r could not be created because of %r",
+                    content_translation,
```
timoludwig, line 153:
```suggestion
                    content_object,
```

integreat_cms/deepl_api/utils.py
```diff
@@ -0,0 +1,161 @@
...
+            # Validate event translation
+            if content_translation_form.is_valid():
+                content_translation_form.save()
+                logger.debug(
+                    "Successfully translated for: %r", content_translation_form.instance
+                )
+                messages.success(
+                    request,
+                    _('Event "{}" has been successfully translated.').format(
+                        source_translation.title
```
timoludwig, line 147:
```suggestion
                    _('{} "{}" has been successfully translated.').format(
                        type(content_object)._meta.verbose_name.title(),
                        source_translation.title,
```
Since the function is not only for events, I suggest to adapt the success message here.

integreat_cms/deepl_api/utils.py
```diff
@@ -0,0 +1,161 @@
...
+            else:
+                logger.error(
+                    "Automatic translation for %r could not be created because of %r",
+                    content_translation,
+                    content_translation_form.errors,
+                )
+                messages.error(
+                    request,
+                    _('Event "{}" could not be automatically translated.').format(
+                        source_translation.title
```
timoludwig, line 159:
```suggestion
                    _('{} "{}" could not be automatically translated.').format(
                        type(content_object)._meta.verbose_name.title(),
                        source_translation.title,
```

integreat_cms/cms/views/pois/poi_actions.py
```diff
@@ -181,21 +185,28 @@ def automatic_translation(request, region_slug, language_slug):
...
-    pois = region.pois.filter(id__in=poi_ids).prefetch_translations()
+    pois = request.region.pois.filter(
+        id__in=request.POST.getlist("selected_ids[]")
+    ).prefetch_translations()
     logger.debug("Automatic translation for POIs: %r", pois)
-
-    if settings.DEEPL_ENABLED:
+    deepl = DeepLApi()
+    if deepl.check_availability(request, language_slug):
+        deepl.deepl_translation(request, pois, language_slug, POITranslationForm, "poi")
```
timoludwig, line 204:
```suggestion
        deepl.deepl_translation(request, pois, language_slug, POITranslationForm)
```

integreat_cms/cms/views/events/event_actions.py
```diff
@@ -243,20 +246,30 @@ def automatic_translation(request, region_slug, language_slug):
...
+    events = request.region.events.filter(id__in=event_ids).prefetch_translations()
     logger.debug("Automatic translation for events: %r", events)
-
-    if settings.DEEPL_ENABLED:
-        messages.warning(
-            request, _("Automatic translations are not fully implemented yet")
+    deepl = DeepLApi()
+    if deepl.check_availability(request, language_slug):
+        deepl.deepl_translation(
+            request, events, language_slug, EventTranslationForm, "event"
```
timoludwig, line 266:
```suggestion
            request, events, language_slug, EventTranslationForm)
```

timoludwig (commented):
Thanks a lot! :+1:
I like the separate utility function!
I just found a few smaller issues, see my suggestions below :slightly_smiling_face:

integreat_cms/deepl_api/utils.py
```diff
@@ -0,0 +1,161 @@
...
+        # Get target language
+        target_language = request.region.get_language_or_404(language_slug)
+        source_language = request.region.get_source_language(language_slug)
+        for content_object in content_objects:
+            if source_translation:
+                source_translation = content_object.get_translation(
+                    source_language.slug
+                )
+            else:
+                messages.error(request, _("No source translation could be found."))
```
timoludwig, line 94:
```suggestion
            source_translation = content_object.get_translation(source_language.slug)
            if not source_translation:
                messages.error(
                    request,
                    _('No source translation could be found for {} "{}".').format(
                        type(content_object)._meta.verbose_name.title(),
                        content_object.best_translation.title,
                    ),
                )
                continue
```
You need to reverse the order here, first you need to get the translation, then you can ceck whether the translation exists. Otherwise you get the error `UnboundLocalError: local variable 'source_translation' referenced before assignment`. Also, you need to do a `continue` to skip the rest of the loop-iteration in case of an error, otherwise it will add the error message, but will just continue with the code until the `AttributeError` occurs.
Apart from that, I'd also add the "best translation title" to the message in case multiple objects were selected and it's not clear which of them were successful and which of them failed.

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -6544,6 +6546,18 @@ msgid "Superuser permissions need to be set by another superuser."
 msgstr ""
 "Administratorrechte müssen von einem anderen Administrator vergeben werden."
 
+#: deepl_api/utils.py:94
+msgid "No source translation could be found."
+msgstr "Es konnte kein Quelltext gefunden werden."
+
+#: deepl_api/utils.py:144
+msgid "{} \"{}\" has been successfully translated."
+msgstr "Die Veranstaltung wurde erfolgreich übersetzt"
```
timoludwig, line 6555:
```suggestion
msgstr "{} \"{}\" wurde erfolgreich übersetzt"
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -6544,6 +6546,18 @@ msgid "Superuser permissions need to be set by another superuser."
...
+
+#: deepl_api/utils.py:157
+msgid "{} \"{}\" could not be automatically translated."
+msgstr "Die Veranstaltung konnte nicht erfolgreich übersetzt."
```
timoludwig, line 6559:
```suggestion
msgstr "{} \"{}\" konnte nicht automatisch übersetzt werden."
```

timoludwig (commented):
Thanks! :muscle: 
Still a few problems left, and could you add a note to the [CHANGELOG.md](https://github.com/digitalfabrik/integreat-cms/blob/develop/CHANGELOG.md) into the "unreleased" section?

timoludwig (approved):
Amazing, thanks a lot! :+1:
Works very well :tada: :rocket:

Issue #1128, opened 2022-01-25
Page search not working anymore


timoludwig:
### Describe the Bug

Since the performance refactoring, searching for pages doesn't work anymore

### Steps to Reproduce

1. Go to page tree
2. Search for a page

### Expected Behavior

The filtered results should be shown

### Actual Behavior

The error `'list' object has no attribute 'cache_tree'` appears

PR #1136, opened 2022-01-27
Fix page filters

timoludwig:
### Short description

Fixes the handling of page filters (by search query and by status filter)

### Proposed changes

- Make the page filters work with the cached tree structure

Caveats:
I'm still not sure how the hierarchy should be handled when filtering the tree. At the moment, I just disabled expanding/collapsing subpages, because it would require quite some computation to get all subpages that are in the same filtered set...
Do you think it would make more sense to completely hide the "hierarchy" column when filters are applied? Since it's not a real tree structure in this case anyway...

### Resolved issues

Fixes: #1128

### DIFF
--- a/integreat_cms/api/v3/pages.py
+++ b/integreat_cms/api/v3/pages.py
@@ -72,9 +72,11 @@ def pages(request, region_slug, language_slug):
     """
     region = request.region
     result = []
-    for page in region.get_pages(
-        prefetch_translations=True, prefetch_public_translations=True
-    ).cache_tree():
+    # The preliminary filter for explicitly_archived=False is not strictly required, but reduces the number of entries
+    # requested from the database
+    for page in region.pages.filter(explicitly_archived=False).cache_tree(
+        archived=False
+    )[0]:
         page_translation = page.get_public_translation(language_slug)
         if page_translation:
             result.append(transform_page(page_translation))
--- a/integreat_cms/cms/forms/pages/page_form.py
+++ b/integreat_cms/cms/forms/pages/page_form.py
@@ -88,11 +88,7 @@ class PageForm(CustomModelForm, MoveNodeForm):
         )
 
         # Limit possible parents to pages of current region
-        parent_queryset = (
-            self.instance.region.pages.all()
-            .prefetch_translations()
-            .prefetch_public_translations()
-        )
+        parent_queryset = self.instance.region.pages.all()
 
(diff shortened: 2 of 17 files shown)


ulliholtgrave:
I agree with your point regarding the hierarchy. But I guess we should indicate somehow that the user is some different "view". Maybe we should add some kind of red "Reset" button that brings the user back to the overview. Or we could implement the AJAX call that could load all subpages of a parent, if we want to keep the expansion in this view.

ulliholtgrave (changes requested):
I guess I would go for your solution right now, but from a user perspective I would make the drag&drop and expand icon gray to tell the user that they aren't active.

timoludwig:
> I guess, I would go with your solution right now, but from a user perspective I would make the drag&drop and expand icon gray to tell the user that they aren't active.

I just tried that out, but I'm not sure whether that's enough... For example, consider this page tree:

Screenshot 2022-01-28 at 11-15-12 Integreat Redaktionssystem

And then searched for "augsburg", the tree structure now looks like this:

Screenshot 2022-01-28 at 11-18-17 Integreat Redaktionssystem

This kind of indicates that "Wissenswertes über Augsburg" would be the parent pages of the other search results, which it isn't... 
Also, having ajax in this view wouldn't change much, because the parents/anncestors of the search results would still be invisible...
I kind of tend to my initial suggestion of completely hide the "hierarchy" column when the result is filtered.

@nikolahoff maybe you have an idea how to visualize a filtered tree structure?

ulliholtgrave (approved):

ulliholtgrave:
I agree with your point and from my perspective we can merge it like it is. I'll talk with Nikola about this. Maybe she has a smart solution for that issue :)

nikolahoff:
If it's possible I would suggest to hide both interaction elements (open tree arrow and move-arrow) as they can't be used in that case and might be confusing to the user.

timoludwig:
> If it's possible I would suggest to hide both interaction elements (open tree arrow and move-arrow) as they can't be used in that case and might be confusing to the user.

@nikolahoff Thanks for your input, I now removed the "hierarchy" column and added a reset-button to the search input:

Screenshot 2022-02-02 at 01-56-17 Integreat Redaktionssystem

ulliholtgrave (approved):
Thanks :+1: Looks good

PR #1137, opened 2022-01-27
Reduce queries for translation coverage to a constant amount

david-venhoff:
### Short description
This pr reduces the amount of database queries required to load the translation coverage view to a constant number.
Related issue: #943 

### Proposed changes

- Use the new page query api
- Don't show languages in the outdated translation word count table if no translation for that language is outdated

### DIFF
--- a/integreat_cms/cms/views/analytics/translation_coverage_view.py
+++ b/integreat_cms/cms/views/analytics/translation_coverage_view.py
@@ -1,4 +1,5 @@
 import logging
+from collections import Counter
 
 from django.contrib.auth.decorators import login_required
 from django.utils.decorators import method_decorator
@@ -6,7 +7,6 @@ from django.views.generic import TemplateView
 from django.shortcuts import render
 
 from ...constants import translation_status
-from ...models import PageTranslation
 from ...decorators import region_permission_required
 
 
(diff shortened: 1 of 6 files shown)


integreat_cms/cms/views/analytics/translation_coverage_view.py
```diff
@@ -44,67 +44,41 @@ def get(self, request, *args, **kwargs):
...
         region = request.region
 
-        total_number_pages = region.pages.count()
-
-        translation_coverage_data = []
-        outdated_word_count = {}
+        translation_coverage_data = {}
+        outdated_word_count = Counter()
 
+        pages = region.pages.all().prefetch_translations()
```
timoludwig, line 50:
```suggestion
        pages = (
            region.pages.filter(explicitly_archived=False)
            .prefetch_translations()
            .cache_tree()
        )
```
I think it would make sense to only show non-archived pages here.
Unfortunately, we don't have an efficient way to get all non-archived pages at once apart from `cache_tree()` (`region.get_pages(archived=False` doesn't perform well) at the moment, so I'd go with this approach even if it does a little bit of unnecessary tree structure caching...

integreat_cms/cms/views/analytics/translation_coverage_view.py
```diff
@@ -44,67 +44,41 @@ def get(self, request, *args, **kwargs):
...
+                data,
             )
-
         logger.debug("Outdated word count: %r", outdated_word_count)
 
         # Assemble the ChartData in the format expected by ChartJS (one dataset for each translation status)
         chart_data = {
             "labels": [
-                language.translated_name for language in region.active_languages
+                language.translated_name for language in translation_coverage_data
```
timoludwig, line 73:
```suggestion
                language.translated_name for language in translation_coverage_data.keys()
```
 I know that the default dict-to-list conversion is taking the keys and not the values, but to improve readability here, I'd go with the explicit approach.

timoludwig (commented):
Thanks a lot, this works like a charm! :tada:
I think this change is substantial enough to deserve a small note in the Changelog.

integreat_cms/cms/views/analytics/translation_coverage_view.py
david-venhoff: If I understand correctly, these pages should be filtered later using e.g. `if not page.archived:`?

integreat_cms/cms/views/analytics/translation_coverage_view.py
david-venhoff: I actually wrote it that way first, but pylint did not seem to like it :sweat_smile:

integreat_cms/cms/views/analytics/translation_coverage_view.py
```diff
@@ -44,67 +45,48 @@ def get(self, request, *args, **kwargs):
...
             )
-
         logger.debug("Outdated word count: %r", outdated_word_count)
 
         # Assemble the ChartData in the format expected by ChartJS (one dataset for each translation status)
         chart_data = {
             "labels": [
-                language.translated_name for language in region.active_languages
+                language.translated_name
+                for language in translation_coverage_data.keys()
```
timoludwig, line 81:
```suggestion
                language.translated_name for language in translation_coverage_data
```
Ok, I'm very sorry, I thought `keys()` would be more readable, but if pylint prefers the direct dict iteration, maybe that's better code style :information_desk_person:

integreat_cms/cms/views/analytics/translation_coverage_view.py
```diff
@@ -25,6 +25,7 @@ class TranslationCoverageView(TemplateView):
     #: The context dict passed to the template (see :class:`~django.views.generic.base.ContextMixin`)
     base_context = {"current_menu_item": "translation_coverage"}
 
+    # pylint: disable=consider-iterating-dictionary
```
timoludwig, line 28:
```suggestion
```

integreat_cms/cms/views/analytics/translation_coverage_view.py
```diff
@@ -44,67 +45,48 @@ def get(self, request, *args, **kwargs):
...
+        translation_coverage_data = {}
+        outdated_word_count = Counter()
 
+        pages = [
+            page
+            for page in region.pages.filter(explicitly_archived=False)
+            .prefetch_translations()
+            .cache_tree()
+            if not page.archived
+        ]
```
timoludwig, line 57:
```suggestion
        pages = (
            region.pages.filter(explicitly_archived=False)
            .prefetch_translations()
            .cache_tree()
        )
```
Actually, `cache_tree()` already skips nodes that don't have a parent node in the queryset, so if it's filtered by `explicitly_archived=False`, we don't have to perform additional filters.

timoludwig (commented):
Thanks! Could you squash you commits and resolve the conflicts with the develop branch before merging?

timoludwig (approved):
:tada:

PR #1138, opened 2022-01-27
Fix search bar

david-venhoff:
### Short description
This pr fixes a few problems I noticed with the search input bar.

### Proposed changes
- Fix the `page.implicitly_archived` method returning a list instead of a boolean, causing the search to return no results
- ~When filtering for pages in `page_tree_view` create a query instead of a list so that the later call to `pages.cache_tree()` works~ I just noticed that this is fixed by Timo's previous pr #1136
- Don't treat a click on the scrollbar of the search suggestions as simultaneously clicking all suggestions


timoludwig (approved):

david-venhoff:
Yes, I will look into that :+1:

PR #1142, opened 2022-02-02
Update/dev dependencies

ulliholtgrave:
### Short description

This PR updates both python and JS dependencies. Nevertheless, I experienced some strange behavior with our static files, which forced me to change the file-loader path creation in the webpack configuration. I keep the PR as a draft and try to add the real names to the paths again and maybe @timoludwig could give me an idea, if `npx` is required. The former implementation with npx was interpreted by my system as if the flags are related to npx and therefore it always broke.


timoludwig (approved):
Thanks! :+1:

PR #1144, opened 2022-02-02
Add setting to enable/disable automatic translations

timoludwig:
### Short description

Add setting to enable/disable automatic translations

### Proposed changes

- Add settings `DEEPL_AUTH_KEY` and `DEEPL_ENABLED`
- Hide automatic translation bulk actions if DeepL is deactivated


ulliholtgrave (approved):
Thanks! :)

PR #1145, opened 2022-02-02
Improve performance of content searches

david-venhoff:
### Short description
This pr improves the performance when performing a content search.
In general, the search was fairly optimized and the biggest changes could be made to pages.
Related issue: #943

### Proposed changes
- Prefetch some attributes for pages, events and pois
- Make use of cached data when searching pages

### DIFF
--- a/integreat_cms/cms/views/utils/search_content_ajax.py
+++ b/integreat_cms/cms/views/utils/search_content_ajax.py
@@ -12,7 +12,6 @@ from ...decorators import region_permission_required
 from ...models import (
     Region,
     EventTranslation,
-    PageTranslation,
     POITranslation,
     Feedback,
     PushNotificationTranslation,
@@ -78,9 +77,11 @@ def search_content_ajax(request, region_slug=None, language_slug=None):
     user = request.user
     if user.has_perm("cms.view_event") and "event" in object_types:
         object_types.remove("event")
-        event_translations = EventTranslation.search(
-            region, language_slug, query
-        ).filter(event__archived=archived_flag, status=status.PUBLIC)
+        event_translations = (
+            EventTranslation.search(region, language_slug, query)
+            .filter(event__archived=archived_flag, status=status.PUBLIC)
+            .select_related("event__region", "language")
+        )
         results.extend(
             format_object_translation(obj, "event") for obj in event_translations
         )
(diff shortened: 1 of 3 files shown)


integreat_cms/cms/models/abstract_content_model.py
```diff
@@ -268,7 +268,15 @@ def __repr__(self):
...
         class_name = type(self).__name__
-        return f"<{class_name} (id: {self.id}, region: {self.region.slug}, slug: {self.best_translation.slug})>"
+        # Only log the slug if doing so does not require a database query
+        slug = (
+            self.best_translation.slug
+            if hasattr(self, "prefetched_translations")
+            else "<...>"
+        )
+        return (
+            f"<{class_name} (id: {self.id}, region: {self.region.slug}, slug: {slug})>"
```
david-venhoff, line 278: I am a bit confused why pylint does such aggressive line breaks. Why would it break up the format string, even though it got shortened?

david-venhoff:
On a side note, I wondered why only public pages get filtered. Wouldn't it make sense to allow searching through all pages of the page tree?
This can be especially confusing in the archived pages view, where pages with e.g. `status: Draft` are visually indistinguishable from other pages, but are not taken into account when searching.

integreat_cms/cms/views/utils/search_content_ajax.py
```diff
@@ -98,21 +100,35 @@ def search_content_ajax(request, region_slug=None, language_slug=None):
...
+        for page_translation in query:
+            # Copy the cached attributes from the ``pages`` query to this page translation to prevent database queries
+            # pylint: disable=protected-access
+            cached_page = pages[page_translation.page.id]
+            page_translation.page._cached_ancestors = cached_page._cached_ancestors
+            page_translation.page.prefetched_public_translations = (
+                cached_page.prefetched_public_translations
+            )
+
+            results.append(format_object_translation(page_translation, "page"))
```
timoludwig, line 124:
```suggestion
        pages = region.pages.all().cache_tree(archived=archived_flag)[0]
        for page in pages:
            page_translation = page.get_translation(language_slug)
            if page_translation and (
                query.lower() in page_translation.slug
                or query.lower() in page_translation.title.lower()
            ):
                results.append(format_object_translation(page_translation, "page"))
```
Since we already fetched all pages including their translations from the database, we can also do the search in Python instead of another database query.

timoludwig (commented):
Thanks a lot, looks very good! :+1:

> On a side note, I wondered why only public pages get filtered. Wouldn't it make sense to allow searching through all pages of the page tree? This can be especially confusing in the archived pages view, where pages with e.g. `status: Draft` are visually indistinguishable from other pages, but are not taken into account when searching.

You're completely right, this is definitely a problem. I think the origin of the problem is that this view is generalized a bit too much and has to handle a lot of different use cases. One of them is the link suggestion in the content forms, where we only want to have public translations. So probably we need another flag similar to `archived_flag` which tells us whether to return only public translations or not. But this is probably a bit out of scope for this PR, I think you can either leave it as it is or remove the "public"-filter and I will open another issue to fix this in the future.

integreat_cms/cms/models/abstract_content_model.py
```diff
@@ -268,7 +268,15 @@ def __repr__(self):
         :rtype: str
         """
         class_name = type(self).__name__
-        return f"<{class_name} (id: {self.id}, region: {self.region.slug}, slug: {self.best_translation.slug})>"
+        # Only log the slug if doing so does not require a database query
```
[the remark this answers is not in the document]
david-venhoff: Yeah, this is probably not necessary. I added this, because the `PageQuerySet.cache_tree` method logs all pages, where this method requires a databse call for every page. I did not keep in mind that this would only affect debug mode.

david-venhoff:
Okay, I just committed your suggestions, which removed the public filter.

timoludwig (approved):
Thank you! I just resolved the conflicts with the develop branch and will merge this PR as soon as the checks are passed. :tada:

Issue #1134, opened 2022-01-25
Fix XLIFF segmentation


timoludwig:
### Motivation

In turns out, that the segmentation of XLIFF files is more complicated than initially thought.
The current Wordpress + WPML setup does the following:

1. The content in the database is just plain text with `\n`, `\r` and `\t` characters to represent paragraphs
2. The corresponding `<p>` tags are only inserted in the page API, not in the XLIFF export/import
3. The content is segmented in MemoQ with a specialized WPML filter that presumably converts each sentence into one XLIFF segment

When we add `<p>`-tags during the data migration from WordPress to Django (see https://github.com/digitalfabrik/integreat-cms/issues/1118 and https://github.com/digitalfabrik/cms/pull/1126), segmentation breaks and our translation memories don't work anymore.

### Proposed Solution

I see three options:
1. Just keep the wordpress database structure without `<p>`-tags and add them on runtime in the API - probably bad for performance and it doesn't make too much sense
2. The quick and dirty approach: Just reverse the `wpautop()` function and apply it before each XLIFF export in Django. After that, the result should be identical to the content WPML is exporting and the MemoQ-WPML-filter can also be applied to the Django-XLIFF.
3. The long-term solution: Find out what the MemoQ-WPML-filter is doing and do the segmentation already during the XLIFF-export in Django. If we manage to create the same segments, this would be better because the idea of the XLIFF standard in the first place is to put each segment into one `<trans-unit>` tag and not to put the entire content in one large `<![CDATA[]]>`-section and apply some dark magic to get segments out of this large blob (then why use XLIFF in the first place?)...

ulliholtgrave:
So, a short summary of the found solution: 

1. For our XLIFF-Im/Exports we need the specific WPML-header in our XLIFF-File, so the WPML filter successfully detects an WPML file:
`<header><phase-group><phase`  phase-name="shortcodes" process-name="Shortcodes  identification"></phase><phase phase-name="post_type"  process-name="Post  type"></phase></phase-group></header>`
2. The id field of the <trans unit> has to be the same as in the old XLIFF files as segments with different context ids will never be considered as the same context and even though they are the exact content this won't result in an 101% match. Therefore, the id field should be "body" to match our old TMs
3. Consequences for the translation process with MemoQ: To match XLIFF-files created by our django system, we have to import them with the WPML filter and then the matches with our TM should work out :slightly_smi

PR #1147, opened 2022-02-03
Fix XLIFF segmentation

timoludwig:
### Short description

Fix XLIFF segmentation

### Proposed changes

- Add legacy `<header>` in XLIFF 1.2
- Use legacy field ids in XLIFF 1.2

### Resolved issues

Fixes: #1134

### DIFF
--- a/integreat_cms/xliff/xliff1_serializer.py
+++ b/integreat_cms/xliff/xliff1_serializer.py
@@ -1,6 +1,7 @@
 import logging
 from urllib.parse import urlparse
 
+from django.conf import settings
 from django.core.serializers import base
 
 from ..cms.models import Page, PageTranslation, Language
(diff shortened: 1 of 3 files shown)


integreat_cms/xliff/xliff1_serializer.py
```diff
@@ -54,6 +55,23 @@ def start_object(self, obj):
...
+            attrs={
+                "phase-name": "shortcodes",
+                "process-name": "Shortcodes identification",
+            },
+        )
+        self.xml.addQuickElement(
+            "phase", attrs={"phase-name": "post_type", "process-name": "Post type"}
+        )
+        self.xml.endElement("phase-group")
+        self.xml.endElement("header")
```
timoludwig, line 74: This produces the following output:
```
<header>
    <phase-group>
        <phase phase-name="shortcodes" process-name="Shortcodes identification"/>
        <phase phase-name="post_type" process-name="Post type"/>
    </phase-group>
</header>
```
If re really need both the open and close tags as stated in #1134, I think we have to insert a dummy note into these phases, otherwise the XML parser will always use `<phase/>` instead of `<phase></phase`.

ulliholtgrave: Maybe we can just try it out on our system and Vijay can check whether it works or not?

timoludwig: Ok, sounds good :+1:

ulliholtgrave (approved):
Let's try it out :)

Issue #1101, opened 2022-01-12
Use AJAX for the page tree view


svenseeberg:
### Motivation
Loading all pages in the back end at once consumes lots of CPU and also creates a very huge DOM tree which is slow in some browsers.

### Proposed Solution
We only load the first level and load further pages if the tree is expanded. When doing bulk actions we assume that everything applies to children, unless the children are expanded.

### Alternatives
Non really.

melegiul:
When I understand this right, the preferred approach is to start with a DOM tree only containing root pages and creating new DOM nodes on every click on collapse.
As the django template [integreat_cms/cms/templates/pages/page_tree_node.html](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/templates/pages/page_tree_node.html) contains 160 lines, creating this rows with typescript will get very tedious and after expanding some pages the DOM tree is getting huge, too
Also we would need an extended children endpoint, to get currently missing information like page_translation states (outdated, up-to-date, etc)

So I want to suggest a different approach: the django admin site also uses pagination in combination with drag and drop operation, which are limited to the current selected page
To overcome this limitation, we could introduce a new bulk action "Move", to shift content between different pages
The advantage in this case is, that we can keep our current template loop for populating the table rows (=> less effort), while pagination should also reduce CPU load and DOM tree size at the page tree 

What you think about this?

![image](https://user-images.githubusercontent.com/47010475/152573381-e0809da2-c0bd-472f-96d4-a9ca178792fe.png)

timoludwig:
> As the django template [integreat_cms/cms/templates/pages/page_tree_node.html](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/templates/pages/page_tree_node.html) contains 160 lines, creating this rows with typescript will get very tedious and after expanding some pages, the DOM tree is getting huge, too Also we would need an extended children endpoint, to get currently missing information like page_translation states (outdated, up-to-date, etc)

I think we should not build the complete tree with typescript, but have an ajax endpoint that already returns the rendered html of the partial tree including all translation states etc... so the ajax view takes a parent page id as parameter and returns the rendered html tree of all direct children of that parent. I believe this could work without too much work, in typescript we then only have to insert this html code at the correct position...

timoludwig:
We use a similar approach for the page order table in the page form:
[...]

PR #1149, opened 2022-02-07
Page tree ajax request to fetch children

melegiul:
### Short description

Fetch page tree children by ajax request

### Proposed changes

- Page tree starts with root nodes only
- On expanding new page children get queried and added to the DOM

### Resolved issues

Fixes: #1101

### DIFF
--- a/integreat_cms/cms/templates/pages/page_tree_node.html
+++ b/integreat_cms/cms/templates/pages/page_tree_node.html
@@ -4,8 +4,8 @@
 {% load tree_filters %}
 
 {% get_translation page language.slug as page_translation %}
-<tr id="page-{{ page.id }}-drop-left" data-drop-id="{{ page.id }}" data-drop-position="left" class="drop {% if page.relative_depth == 1 %} drop-between {% endif %} h-3 -m-3 hidden level-{{page.relative_depth}}"><td colspan="9"><div><span></span></div></td></tr>
-<tr id="page-{{ page.id }}" data-drop-id="{{ page.id }}" data-drop-position="last-child" class="drop drop-on border-t border-b border-solid border-gray-200 hover:bg-gray-100 {% if page.relative_depth > 1 %} level-{{page.relative_depth}} {% if not filter_form.has_changed %} hidden {% endif %} {% endif %}">
+<tr id="page-{{ page.id }}-drop-left" data-drop-id="{{ page.id }}" data-drop-position="left" class="drop drop-between h-3 -m-3 hidden level-{{page.relative_depth}}"><td colspan="9"><div><span></span></div></td></tr>
+<tr id="page-{{ page.id }}" data-drop-id="{{ page.id }}" data-drop-position="last-child" class="drop drop-on border-t border-b border-solid border-gray-200 hover:bg-gray-100 {% if page.relative_depth > 1 %} level-{{page.relative_depth}} {% endif %}{% if parent_id %}parent-page-{{ parent_id }}{% endif %}">
     <td class="pl-4">
         <input type="checkbox" name="selected_ids[]" value="{{ page.id }}" class="bulk-select-item">
     </td>
@@ -18,13 +18,15 @@
                   title="{% trans 'Change the order and position of the pages with drag & drop.' %}">
                 <i data-feather="move"></i>
             </span>
-            {% if page.cached_children|length > 0 %}
+            {% if not page.is_leaf %}
                 <span class="collapse-subpages cursor-pointer inline-block align-middle"
                       title="{% trans 'Expand all subpages' %}"
                       data-alt-title="{% trans 'Collapse all subpages' %}"
                       data-page-id="{{ page.id }}"
-                      data-page-children="{{ page|get_children_ids }}">
+                      data-page-children="{{ page|get_children_ids }}"
+                      data-page-children-url="{% url 'get_page_children_ajax' region_slug=request.region.slug language_slug=language.slug tree_id=page.tree_id lft=page.lft rgt=page.rgt depth=page.depth %}">
                     <i data-feather="chevron-right"></i>
+                    <svg class="animate-spin h-5 w-5 border-4 border-indigo-200 rounded-full border-t-blue-500 hidden" viewBox="0 0 24 24"></svg>
                 </span>
             {% endif %}
         </td>
--- a/integreat_cms/cms/views/pages/partial_page_tree_view.py
+++ b/integreat_cms/cms/views/pages/partial_page_tree_view.py
@@ -1,0 +1,78 @@
+from django.conf import settings
+from django.contrib.auth.decorators import login_required
+from django.db.models import Q
+from django.shortcuts import render
+from django.utils.decorators import method_decorator
+from django.views.generic import TemplateView
+
+from integreat_cms.cms.views.pages.page_context_mixin import PageContextMixin
+
+from ...constants import status, translation_status
+from ...decorators import region_permission_required, permission_required
+
+
+@method_decorator(login_required, name="dispatch")
+@method_decorator(region_permission_required, name="dispatch")
+@method_decorator(permission_required("cms.view_page"), name="dispatch")
+class PartialPageTreeView(TemplateView, PageContextMixin):
+    """
+    View for rendering a partial page tree
+    """
+
+    #: Template for a partial page tree
+    template = "pages/_page_tree_children.html"
+
+    # pylint: disable=unused-argument
+    def get(self, request, *args, **kwargs):
+        r"""
+        Retrieve the rendered subtree of a given parent page
+
+        :param request: The current request
+        :type request: ~django.http.HttpResponse
+
+        :param \*args: The supplied arguments
+        :type \*args: list
+
+        :param \**kwargs: The supplied keyword arguments
+        :type \**kwargs: dict
+
+        :return: The rendered template response
+        :rtype: ~django.template.response.TemplateResponse
+        """
+        tree_id = int(kwargs.get("tree_id"))
+        lft = int(kwargs.get("lft"))
+        rgt = int(kwargs.get("rgt"))
+        depth = int(kwargs.get("depth"))
+        region = request.region
+        language = region.get_language_or_404(
+            kwargs.get("language_slug"), only_active=True
+        )
+        # List of all ancestors, the requested parent and all direct children
+        pages = region.pages.filter(
+            # This condition queries the parent and all direct children
+            Q(tree_id=tree_id, lft__range=(lft, rgt - 1), depth__lte=depth + 1)
+            # This condition queries all ancestors
+            | Q(tree_id=tree_id, lft__lt=lft, rgt__gt=rgt)
+        ).cache_tree(False)[0]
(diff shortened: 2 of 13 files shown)


timoludwig (commented):
Amazing, thanks a lot! :muscle: :rocket: :heart_eyes: 
I noticed a few problems, have a look at my comments below for details.

Apart from that, I think it would be cool to load the next children in advance after the page load and before the sub-pages are expanded, so that when the user clicks on expand, the results are already there... but this is probably out of scope for this PR, let's open another issue after this is finished...

integreat_cms/cms/templates/pages/page_tree_node.html
```diff
@@ -4,8 +4,8 @@
 {% load tree_filters %}
 
 {% get_translation page language.slug as page_translation %}
-<tr id="page-{{ page.id }}-drop-left" data-drop-id="{{ page.id }}" data-drop-position="left" class="drop {% if page.relative_depth == 1 %} drop-between {% endif %} h-3 -m-3 hidden level-{{page.relative_depth}}"><td colspan="9"><div><span></span></div></td></tr>
-<tr id="page-{{ page.id }}" data-drop-id="{{ page.id }}" data-drop-position="last-child" class="drop drop-on border-t border-b border-solid border-gray-200 hover:bg-gray-100 {% if page.relative_depth > 1 %} level-{{page.relative_depth}} {% if not filter_form.has_changed %} hidden {% endif %} {% endif %}">
+<tr id="page-{{ page.id }}-drop-left" data-drop-id="{{ page.id }}" data-drop-position="left" class="drop drop-between h-3 -m-3 hidden level-{{page.relative_depth}}"><td colspan="9"><div><span></span></div></td></tr>
```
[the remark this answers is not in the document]
melegiul: Thank you very much for your review, I applied your suggestions so far and they work greatly.
Here I think we have a problem, committing this suggestion will break drag and drop, because the relative depth for children of roots is 2, so they will not render this additional drop region above the current page after expanding, I remove this if statement therefore.
But I can try to find an if statement, so that it will render for children and not for search results and archived pages

integreat_cms/static/src/js/pages/collapse-subpages.ts
```diff
@@ -65,8 +76,44 @@ function toggleSubpagesRecursive(collapseSpan: HTMLElement) {
...
+    let newCollapseSpan = rowToinsert.querySelector(".collapse-subpages");
+    if (newCollapseSpan) {
+      // add event listener if not leaf node
+      newCollapseSpan.addEventListener("click", toggleSubpages);
+    }
+    insertAfter(rowToinsert, currentRow);
+    currentRow = rowToinsert;
+  });
+  // render rows buttton icons
+  feather.replace({ class: 'inline-block' });
```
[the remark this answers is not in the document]
melegiul: Thanks for reporting this, in 902aeb5b8d86d9a1305b59cbdbe5420c5c767e26 I have added the listeners and also set the confirmation context by creating a new constant, which is used now in the view and in the mixin
Now the confirmation dialogs should work

integreat_cms/static/src/js/pages/collapse-subpages.ts
```diff
@@ -14,12 +16,25 @@ window.addEventListener("load", () => {
...
+  let children = collapseSpan.closest("tbody").querySelectorAll(".parent-page-" + pageId);
+  if (children.length === 0) {
+    // Show loading spinner
+    collapseSpan.querySelector("svg").classList.toggle("hidden");
+    collapseSpan.querySelector(".animate-spin").classList.toggle("hidden");
+    await fetchSubpages(collapseSpan)
+    .catch(e => {
+      console.log('There has been a problem with your fetch operation: ' + e.message);
+    });
+    addDragAndDropListeners();
```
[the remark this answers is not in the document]
melegiul: e9dcf9c50e259166e259e231b12a29534b0a6a8d this will iterate over all ancestors and add the recently added children to their list of descendants on children expansion

integreat_cms/cms/views/pages/page_actions.py
```diff
@@ -928,3 +929,70 @@ def render_mirrored_page_field(request, region_slug):
...
+        # This condition queries the parent and all direct children
+        Q(tree_id=tree_id, lft__range=(lft, int(rgt) - 1), depth__lte=depth + 1)
+        # This condition queries all ancestors
+        | Q(tree_id=tree_id, lft__lt=lft, rgt__gt=rgt)
+    ).cache_tree()[0]
+    # If the depth is 1, the first element must be the direct parent, etc.
+    parent = pages[depth - 1]
+    # The remaining pages are the direct children
+    children = pages[depth:]
+    ancestors = [ancestor.id for ancestor in pages[:depth]]
```
timoludwig, line 983:
```suggestion
    # For every depth level, exactly one ancestor is in the list
    ancestor_ids = [ancestor.id for ancestor in pages[:depth]]
```
I think it's better to make it clear what this list only contains ids and not the objects itself.

integreat_cms/cms/templates/pages/page_tree_node.html
```diff
@@ -18,13 +18,14 @@
                   title="{% trans 'Change the order and position of the pages with drag & drop.' %}">
                 <i data-feather="move"></i>
             </span>
-            {% if page.cached_children|length > 0 %}
+            {% if not page.is_leaf %}
```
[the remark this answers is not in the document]
melegiul: What you think about this solution d280d82955ee5ad9a307c3207ed5761e53c3e4ff?
expansion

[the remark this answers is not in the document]
timoludwig: To be honest, I'm not 100% convinced, but I think as an intermediary solution, this is fine. Probably this only happens in rare edge cases, so shouldn't be a big problem. In the long term, we can maybe start the fetching of the subpages before the user clicks on the button, so we can remove the expansion button dynamically where the subpages don't exist.

integreat_cms/cms/templates/messages.html
```diff
@@ -24,3 +26,6 @@
     {% endfor %}
 </div>
 {% endif %}
+<div id="archived-subpages-only" class="bg-blue-100 border-l-4 border-blue-500 text-blue-500 px-4 py-3 mb-5 hidden" role="alert">
+    <p>{% trans 'All subpages have been archived' %}</p>
+</div>
```
timoludwig, line 31: Since the message template is embedded directly before the normal content, I think we should place this div at the beginning of the content of the page tree view template instead of the end of the messages template...
https://github.com/digitalfabrik/integreat-cms/blob/e76ad67d2e165702acafb089aec676d9203a9b56/integreat_cms/cms/templates/_base.html#L265-L266
Also, I'd probably change the message to "All subpages of this page are archived".

timoludwig (commented):
Thanks for your changes! Only a few details are missing now.
And could you add a small message to the changelog about this?

melegiul:
@timoludwig Thanks for your hints, I have now added a changelog entry, moved the archived message box and switched to a class based view
But there still remain probelms at the archived page tree view and with page filtering
I will have a look at them

melegiul:
dab25ce22301e04de593f206a8f884a14af73f15 this should fix the broken archived page tree
Page filters seem to be fine too

integreat_cms/cms/views/pages/partial_page_tree_view.py
```diff
@@ -0,0 +1,78 @@
...
+        rgt = int(kwargs.get("rgt"))
+        depth = int(kwargs.get("depth"))
+        region = request.region
+        language = region.get_language_or_404(
+            kwargs.get("language_slug"), only_active=True
+        )
+        # List of all ancestors, the requested parent and all direct children
+        pages = region.pages.filter(
+            # This condition queries the parent and all direct children
+            Q(tree_id=tree_id, lft__range=(lft, int(rgt) - 1), depth__lte=depth + 1)
```
timoludwig, line 53:
```suggestion
            Q(tree_id=tree_id, lft__range=(lft, rgt - 1), depth__lte=depth + 1)
```
Since you're casting `rgt` to int anyway, it's not required to do it again.

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -4557,12 +4557,16 @@ msgstr ""
 " Dieses Feld freilassen, um eine eindeutigen Permalink aus dem Titel zu "
 "generieren"
 
-#: cms/templates/pages/page_tree.html:13
+#: cms/templates/pages/page_tree.html:12
+msgid "All subpages of this page are archived"
+msgstr "Alle Unterseiten von dieser Seite sind archiviert"
```
timoludwig, line 4562:
```suggestion
msgstr "Alle Unterseiten dieser Seite sind archiviert"
```

integreat_cms/static/src/js/pages/collapse-subpages.ts
```diff
@@ -65,8 +91,91 @@ function toggleSubpagesRecursive(collapseSpan: HTMLElement) {
...
+    (<HTMLElement>childrenDoc.querySelector("#ancestors")).dataset.ancestors
+  ) as number[];
+  ancestors.forEach(function(ancestor: number) {
+    // as new children were added to the DOM, all ancestors have to update their descendants 
+    let ancestorDragSpan = <HTMLElement>document.querySelector("#page-" + ancestor + " .drag");
+    const descendants = JSON.parse(
+      ancestorDragSpan.dataset.nodeDescendants
+    ) as number[];
+    // update descendants
+    ancestorDragSpan.dataset.nodeDescendants = "[" + descendants.concat(childrenIds).join(",") + "]";
```
timoludwig, line 165:
```suggestion
    ancestorDragSpan.dataset.nodeDescendants = `[${descendants.concat(childrenIds)}]`;
```
Lists are automatically joined with commas when converted to a string.

integreat_cms/static/src/js/pages/collapse-subpages.ts
```diff
@@ -65,8 +91,91 @@ function toggleSubpagesRecursive(collapseSpan: HTMLElement) {
...
+ * Add recently fetched children ids to collapse span dataset
+ * 
+ * @param collapseSpan parent element to be updated
+ * @param childrenIds ids to be appended
+ */
+function updateChildren(collapseSpan: HTMLElement, childrenIds: Array<number>) {
+  const currentChidlrenIds = JSON.parse(
+    collapseSpan.dataset.pageChildren
+  ) as number[];
+  collapseSpan.dataset.pageChildren = "[" + currentChidlrenIds.concat(childrenIds).join(",") + "]";
```
timoludwig, line 179:
```suggestion
  collapseSpan.dataset.pageChildren = `[${currentChidlrenIds.concat(childrenIds)}]`;
```

timoludwig (commented):
Awesome, thanks! :rocket: 
Just a few tiny details, then this PR is ready to merge.
Could you rebase your branch to resolve the conflicts and clean up the git history a bit?

timoludwig (approved):
Amazing, thanks! :pray:

PR #1150, opened 2022-02-08
Change Webpack serving of inline assets & Align forms

ulliholtgrave:
### Short description

As of our recent update, I had to change the way of serving images/SVGs etc. as we are having some static paths in our files that can't be served with hashes.

### Proposed Changes
- Change Webpack-Config for serving inline assets
- Update form templates to match the UI of the page form


integreat_cms/static/src/js/forms/tinymce-init.ts
```diff
@@ -157,19 +161,18 @@ window.addEventListener("load", () => {
...
         addIcon(editor, tinymceConfig, "www");
         addIcon(editor, tinymceConfig, "email");
         addIcon(editor, tinymceConfig, "call");
         addIcon(editor, tinymceConfig, "clock");
         addIcon(editor, tinymceConfig, "idea");
-        editor.ui.registry.addIcon(
-          "no_translate",
-          parseSvg(require("../../svg/no_translate.svg").default)
-        );
+        addNoTranslateIcon(editor);
```
timoludwig, line 175:
```suggestion
        editor.ui.registry.addIcon(
          "notranslate",
          parseSvg(require("../../svg/notranslate.svg"))
        );
```

I see no reason why we should add a new function for something that's only used once... And I'd use the same name for the tinymce icon and the file name, so either `no_translate` or `notranslate` in both cases.

timoludwig (commented):
Thanks, this fixes the issue :+1:

ulliholtgrave:
@timoludwig I added your mentioned suggestions :)

webpack.config.js
```diff
@@ -16,7 +16,7 @@ module.exports = {
     filename: "[name].[contenthash].js",
     path: path.resolve(__dirname, "integreat_cms/static/dist"),
     clean: true,
-    assetModuleFilename: 'assets/[name]-[hash][ext][query]'
+    assetModuleFilename: 'assets/[name][ext][query]'
```
timoludwig, line 19: Don't we get caching problems when we remove the hash here? :thinking: 
If we change the files without renaming them, users have to manually clear their cache in order for them to receive the new files...

timoludwig (approved):
Thanks a lot, this fixes a lot of issues :tada:

Issue #1151, opened 2022-02-08
Add possibility to delete languages


timoludwig:
### Motivation

At the moment, it's only possible to create new languages, not to delete old ones.

### Proposed Solution

Add the possibility to delete languages, both from the language list and the language form.

Issue #1143, opened 2022-02-02
Hide "Responsible organization" field in page form if no organizations exist


timoludwig:
### Motivation

The field doesn't make sense when there are no choices which can be selected.

### Proposed Solution

Remove this field in the page form if no organizations exist.

Issue #1106, opened 2022-01-18
Enable deletion of offer templates


timoludwig:
### Motivation

Currently, it's only possible to add offer templates and not to delete them

### Proposed Solution

Add a mechanism to delete them similar to the other models

PR #1156, opened 2022-02-10
Enable deletion of languages and offer templates

timoludwig:
### Short description

This PR contains a few small improvements and adds the possibility to delete languages and offer templates

### Proposed changes

- Hide organization field in page form if no organizations exist
- Generalize deletion views
- Add possibility to delete languages 
- Add possibility to delete offer templates

### Resolved issues

Fixes: #1143
Fixes: #1151
Fixes: #1106


integreat_cms/cms/templates/languages/language_form.html
```diff
@@ -79,5 +79,24 @@ <h3 class="font-bold text-white"><i data-feather="settings" class="mr-2"></i> {%
             </div>
         </div>
     </div>
+    </div>
+    {% if form.instance.id and perms.cms.delete_language %}
```
svenseeberg, line 83: Should we add the permission `delete_language` in the migrations for at least one role?

svenseeberg (changes requested):
Not sure if I have a change request or simply a stupid question ;-)

integreat_cms/cms/templates/languages/language_form.html
timoludwig: I'm not sure, as this function is probably very rarely used, I wouldn't have a problem with an administrator having to do this.
But since languages are protected from deletion when used in a region, I don't see a problem with giving the delete permission to the municipality team either.
Maybe we should keep this in mind and do a general permission refactoring when everything is settled? I'd avoid having too many back-and-forth migrations for the permissions...

ulliholtgrave (approved):
Looks good to me. @svenseeberg I guess it's fine if the superuser to have the permission to delete languages for now

Issue #988, opened 2021-10-26
Add browser warning on unsaved content changes


ulliholtgrave:
### Motivation
Our current wordpress system raises a browser warning if someone tries to leave an editing page (see screenshot). Our current django implementation should do the same.
![grafik](https://user-images.githubusercontent.com/15608678/138903655-04bae3a8-e3f6-4b71-a869-2b19c68537cf.png)

### Proposed Solution
Add the browser warning to all content types.

PR #1160, opened 2022-02-11
Add browser warning

MizukiTemma:
### Short description

Browser warning is added so users will be warned when they are leaving a content editing page wihout saving or publishing.
### Proposed changes

- Input and click on the form section will be detected.
- If the user leaves the edit page clicking on "save as Draft", "Save as review" or "Publish"/"Update", nothing new happens. A warning will be shown before leaving otherwise. 

### Resolved issues

Fixes: #988

### DIFF
--- a/integreat_cms/cms/templates/language_tree/language_tree_node_form.html
+++ b/integreat_cms/cms/templates/language_tree/language_tree_node_form.html
@@ -6,7 +6,7 @@
 {% load widget_tweaks %}
 
 {% block content %}
-<form method="post">
+<form method="post" data-unsaved-warning>
     {% csrf_token %}
     <div class="flex flex-wrap mb-4">
         <div class="w-2/5 flex flex-wrap flex-col justify-center">
--- a/integreat_cms/static/src/js/unsaved-warning.ts
+++ b/integreat_cms/static/src/js/unsaved-warning.ts
@@ -1,0 +1,30 @@
+/**
+ * This file contains a function to warn the user when they leave a content without saving
+ */
+
+let confirmed = false;
+let edited = false;
+
+
+window.addEventListener("beforeunload", (event) => {
+  // trigger only when something is edited and no submit/save button clicked
+  if (edited && !confirmed) {
+    event.preventDefault();
+    event.returnValue = "This content is not saved. Would you leave the page?";
+  }
+});
+
+
+window.addEventListener("load", () => {
+  const form = document.querySelector("[data-unsaved-warning]");
+  // checks whether the user typed something in the content
+  form?.addEventListener("input", () => {
+    edited = true;
+    console.debug("editing detected, enabled beforeunload warning");
+  }, { once: true });
+  // checks whether the user has saved or submitted the content
+  form?.addEventListener("submit", () => {
+    confirmed = true;
+    console.debug("form submitted, disabled beforeunload warning");
+  });
+});
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ UNRELEASED
 * [ [#1190](https://github.com/digitalfabrik/integreat-cms/issues/1190) ] Add possibility to set custom region prefix
 * [ [#1164](https://github.com/digitalfabrik/integreat-cms/issues/1164) ] Fix possibility to cancel translation process
 * [ [#1175](https://github.com/digitalfabrik/integreat-cms/issues/1175) ] Don't show empty tag if the page has subpages
+* [ [#988](https://github.com/digitalfabrik/integreat-cms/issues/988) ] Add browser warning when leaving unsaved forms
 
 
 2022.2.0-beta
--- a/integreat_cms/cms/templates/events/event_form.html
+++ b/integreat_cms/cms/templates/events/event_form.html
@@ -10,6 +10,7 @@
     <form id="content_form"
           method="post"
           enctype="multipart/form-data"
+          data-unsaved-warning
           {% if event_form.disabled %}data-disable-poi-query{% endif %}>
         {% csrf_token %}
         {% get_current_language as LANGUAGE_CODE %}
--- a/integreat_cms/cms/templates/imprint/imprint_form.html
+++ b/integreat_cms/cms/templates/imprint/imprint_form.html
@@ -7,7 +7,7 @@
 {% load rules %}
 {% load render_bundle from webpack_loader %}
 {% block content %}
-    <form method="post" id="content_form">
+    <form method="post" id="content_form" data-unsaved-warning>
         {% csrf_token %}
         <div class="w-full flex flex-wrap justify-between gap-4 mb-4">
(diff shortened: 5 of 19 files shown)


integreat_cms/static/src/js/unsaved-warning.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * This file contains a function to warn the user when they leave a content without saving
+ */
+
+ var confirmed = false;
+ var edited = false;
```
membralala, line 6:
```suggestion
let confirmed = false;
let edited = false;
```
As far as I know  `let` is preferred over `var` because of the scoping.

integreat_cms/static/src/js/unsaved-warning.ts
```diff
@@ -0,0 +1,36 @@
...
+
+
+ window.addEventListener('beforeunload', function(e) {
```
membralala, line 9:
```suggestion
window.addEventListener("beforeunload", (event) => {
```

integreat_cms/static/src/js/unsaved-warning.ts
```diff
@@ -0,0 +1,36 @@
...
+    /*trigger only when something is edited and no submit/save button clicked*/
+   if( edited && !confirmed) {
+     e.preventDefault();
+     e.returnValue = 'This content is not saved. Would you leave the page?';
```
membralala, line 13:
```suggestion
    event.preventDefault();
    event.returnValue = "This content is not saved. Would you leave the page?";
```

integreat_cms/static/src/js/unsaved-warning.ts
```diff
@@ -0,0 +1,36 @@
...
+
+
+window.addEventListener("load", () => {
+    /*checks whether the user typed something in the content*/
+    var form = document.getElementById("content_form");
+    form.addEventListener("input", function(e){
+        
+        edited = true;
+        console.log("editing detected");
+    });
```
membralala, line 27:
```suggestion   
    const form = document.getElementById("content_form");
    if (form) {
        form.addEventListener("input", () => {
            edited = true;
            console.log("editing detected");
        });
   }
```

Otherwise there was an Error, when form was null

integreat_cms/static/src/js/unsaved-warning.ts
```diff
@@ -0,0 +1,36 @@
...
+    /*checks whether the user has saved or submitted the content*/
+    var buttons = document.querySelectorAll("button");
```
membralala, line 29:
```suggestion 
    const buttons = document.querySelectorAll("button");
```

integreat_cms/static/src/js/unsaved-warning.ts
```diff
@@ -0,0 +1,36 @@
...
+    buttons.forEach((button) =>{
+        button.addEventListener("click", function(e){
```
membralala, line 31:
```suggestion 
       button.addEventListener("click", () => {
```

membralala:
Thanks! For me this basically works. You might need to consider a few more forms, e.g. the user form and all admin forms, which should also implement this feature, I guess. Apart from this I suggest a few small changes just for the sake of consistency. However, @timoludwig might want to have a look as well 👀.

membralala (commented):
Thanks! For me this basically works. You might need to consider a few more forms, e.g. the user form and all admin forms, which should also implement this feature, I guess. Apart from this I suggest a few small changes just for the sake of consistency. However, @timoludwig might want to have a look as well 👀.

integreat_cms/cms/templates/language_tree/language_tree_node_form.html
```diff
@@ -6,7 +6,7 @@
 {% load widget_tweaks %}
 
 {% block content %}
-<form method="post">
+<form method="post" id="content_form">
```
timoludwig, line 9: The problem with using this id is that it also enables autosave, which we don't want for all the other forms.
So probably it would be better, to use a separate class or data attribute to determine the forms which should be used for the browser warning.

timoludwig (commented):
Thanks a lot! :+1:
In general it works as expected, I just have a few ideas for improvement.
Also, could you add a small notice to the changelog? (Probably you need to rebase you PR onto the current develop branch, otherwise you still have the old changelog format)

membralala:
I tackled the issue, that @timoludwig mentioned regarding the auto-save-function by doing the following: 
- Change the `id="content_form"`, that were added with this PR to a `data-unsaved-warning` attribute 
- Add the `data-unsaved-warning` attribute to all forms, that were originally tagged as content forms

timoludwig (commented):
@membralala 
Could you also add a small note to the changelog for this issue?

integreat_cms/static/src/js/unsaved-warning.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+});
+
+
+window.addEventListener("load", () => {
+  const form = document.querySelector("[data-unsaved-warning]");
+  // checks whether the user typed something in the content
+  form?.addEventListener("input", () => {
+    edited = true;
+    console.debug("editing detected, enabled beforeunload warning");
```
svenseeberg, line 23: The debug messages should be removed :-)

svenseeberg (changes requested):

integreat_cms/static/src/js/unsaved-warning.ts
svenseeberg: Can we unregister the event listener after the first change? I find it somewhat unnecessary to trigger an event for every follow-up change :D

svenseeberg (changes requested):
Sorry, I messes up with the Github review function :sweat_smile: 

Overall this PR does what it is supposed to do. Thanks a lot.

integreat_cms/static/src/js/unsaved-warning.ts
timoludwig: I'm not sure, do the debug logs really harm? I mean it's not a `console.log()`, it is only shown when a user enables debug logging...

svenseeberg: Yes, its not harmful. But to me it is similar to print statements in Python code. Valid for testing stuff, but can be removed when done :shrug:

timoludwig (approved):
I added the option `{ once: true }` to the event listener and added a note to the changelog.
From my side, this PR is ready to merge :tada:

integreat_cms/static/src/js/unsaved-warning.ts
timoludwig: Yeah, but we also have debug logging in python and do not remove it every time we solve an issue... It also makes the general debugging of problems easier in the future if all the debug logging is already there and you don't have to insert a lot of statements to track down a specific bug...

svenseeberg (approved):

svenseeberg:
Good to go then.

PR #1162, opened 2022-02-13
Refactor views

timoludwig:
### Short description

Refactor our views and urlpatterns to make use of more inbuilt Django functionality and reduce duplicated code fragments

### Proposed changes

- Use `get_context_data()` instead of `get()` in template views
    - Only provide the context data, let Django handle the rendering process automatically
- Use correct attribute for template view context
    - The attribute is called `extra_context` instead of `base_context`.
    - Also, it's more flexible to use `get_context_data()` in separate context mixins, because this method can be called from the inherited models even if the `extra_context` attribute is overwritten.
- Use app namespace for API and sitemap url patterns
    - This makes it possible to recognize whether an url is an API or sitemap url by inspecting the resovler match object.
    - Also, it avoids accidental name conflicts with cms urls.
- Check permissions in region middleware
    - Add `AccessControlMiddleware` which checks the authentication status on every request, independently from the views
    - This makes all view-based checks for login, staff status or region permissions obsolete
    - Split cms urls in protected urls and public urls with separate namespace
    - All api urls and the urls whitelisted in the `cms.urls.public` module do not require a login
- Refactor list views
    - Add generic list view for simple object lists without customization
    - Use this generic list for languages, organizations, offer templates, roles and the language tree
    - Provide default ordering for all models
- Migrate `URLconf` to new syntax
    - Use `path()` instead of `url()`
    - Use simple argument parsing instead of regex
    - Consistent use of trailing slashes
    - Use dash instead of underscore in url patterns
- Migrate to pytest for testing
    - Use pytest instead of standard unittests
    - Add test users for all roles to test data
    - Test views for all roles and unauthorized users
    - Use parametrization to generate view tests
    - Use pytest-xdist to run tests in parallel
    - Add tests package to documentation
- Fix permission check in form views
    - It should be possible to view forms with the view permission if they are only retrieved and not submitted
- Fix deletion permission of imprint
    - Hide actions tab if user does not have the permission to delete the imprint
    - Grant imprint deletion permission to management role
- Add view tests with custom parameters
- Add fallback representation for all models
- Add tests for POST views
- Fix ValueError in rules module
    - This fixed the `ValueError: '<Page (id: None)>' needs to have a value for field 'id' before this many-to-many relationship can be used.`

### Resolved issues

This doesn't fix any issues, just introduces new ones.


ulliholtgrave (approved):
Besides the bugs I found earlier, I couldn't find more. Great job! :)

Issue #1165, opened 2022-02-15
Selecting subpages does not activate the bulk action button


timoludwig:
### Describe the Bug

On the page tree, only root pages activate the bulk action button and subpages do not.
This means, that at the moment, bulk actions can only be performed when at least one root page is selected as well.

### Steps to Reproduce

1. Go to the page tree
2. Expand a root page
3. Select a subpage
4. Select a bulk action, e.g. PDF export

### Expected Behavior

The bulk action button should now be active

### Actual Behavior

The button is still disabled

### Additional Information

This bug was probably introduced with the change to ajax page tree fetching, maybe the event handlers have to be renewed after the subpages are loaded.

PR #1169, opened 2022-02-15
Add sub page event listeners for bulk action button

melegiul:
### Short description

Activate bulk action button on selecting sub pages

### Proposed changes

- After selecting a sub page the bulk action button is enabled
- Clicking on "Select all checkboxes" also toggles sub page checkboxes

### Resolved issues

Fixes: #1165

### DIFF
--- a/integreat_cms/static/src/js/bulk-actions.ts
+++ b/integreat_cms/static/src/js/bulk-actions.ts
@@ -34,17 +34,15 @@ function isInputElement(el: Element): el is HTMLInputElement {
 
 window.addEventListener("load", () => {
   const selectAllCheckbox = document.getElementById("bulk-select-all");
-  const bulkAction = document.getElementById("bulk-action") as HTMLSelectElement;
   const bulkActionForm = document.getElementById("bulk-action-form");
   const selectItems = Array.from(document.getElementsByClassName("bulk-select-item"));
-  const bulkActionButton = document.getElementById(
-    "bulk-action-execute"
-  ) as HTMLButtonElement;
+  const bulkAction = document.getElementById("bulk-action") as HTMLSelectElement;
 
   if (selectAllCheckbox && isInputElement(selectAllCheckbox)) {
     selectAllCheckbox.addEventListener("click", () => {
       const value = selectAllCheckbox.checked;
-      selectItems
+      // get all currently rendered page checkboxes
+      Array.from(document.getElementsByClassName("bulk-select-item"))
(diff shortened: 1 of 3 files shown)


integreat_cms/static/src/js/bulk-actions.ts
```diff
@@ -28,23 +28,27 @@
  *
  */
 
+var bulkActionButton: HTMLButtonElement;
+var bulkAction: HTMLSelectElement;
```
timoludwig, line 32: Is it really necessary to define these as global variables and set them on window load?
I think it would be better to just do the `document.getElementById("bulk-action")` every time these variables are used (and as far as I can see, it's only one location anyway?)

integreat_cms/static/src/js/bulk-actions.ts
```diff
@@ -95,16 +82,35 @@ window.addEventListener("load", () => {
...
       .filter(isInputElement)
-      .filter(inputElement => inputElement.checked)
-      .some(inputElement => {
-        return inputElement
-          .closest("tr")
-          .querySelector(`.lang-grid .${languageSlug} .no-trans`) === null;
-      });
+      .some((el) => el.checked) ||
+    bulkAction.selectedIndex === 0 ||
+    (selectedAction.id === "pdf-export-option" && !hasTranslation(selectItems))
```
timoludwig, line 97: Instead of hardcoding the option id here, I'd suggest to add a custom data attribute, e.g. `data-require-existing-translation` or similar, which could then be reused for other bulk actions in the future which also might require a translation.

melegiul: I think the "files changed" tab is a bit misleading here :joy: 
Actually all I did was moving 2 function (toggleBulkActionButton and hasTanslation) out of the inner load window scope to apply the export expression, because 'import' and 'export' may only appear at the top level
But sure we can take the opportunity to improve here a bit

timoludwig: Oh ok, sorry didn't notice :see_no_evil: 
Well then let's leave it as it is, we can still do this once we have another bulk action which requires it^^

timoludwig (commented):
Thanks! Can you squash your commits? Then it's ready to merge :tada:

melegiul:
Thanks for your review, commits are squashed now :slightly_smiling_face:

timoludwig (approved):

Issue #1173, opened 2022-02-15
Unused event location preselected in new Event form


timoludwig:
### Describe the Bug

When events are created and a POI exists which is not set as location for another event, it is preselected as location for the new event

### Steps to Reproduce

1. Create POI which is not used in other events
2. Create new event

### Expected Behavior

A new event should not have a POI preselected as location

### Actual Behavior

The unused POI is selected as location for the new event

Issue #1172, opened 2022-02-15
Filtering for POIs in event list not working


timoludwig:
### Describe the Bug

When filters for POIs in the event list are applied, all events are returned, independent from their location

### Steps to Reproduce

1. Go to event list
2. Filter for event location

### Expected Behavior

Only the events with the selected location should be shown

### Actual Behavior

All events are shown

Issue #1166, opened 2022-02-15
POIs can no longer be created from event form


ulliholtgrave:
### Describe the Bug

Even for superuser it is no longer possible to create a POI from the event form. Probably a permission issue.

### Additional Information

![image](https://user-images.githubusercontent.com/15608678/154031085-74de7527-3189-474e-8ad3-e29683312c12.png)

ulliholtgrave:
Seemed to be a local caching problem.

ulliholtgrave:
Okay, this is a browser-specific event. Firefox seems to work, but chrome and opera not

timoludwig:
I could reproduce this in Chromium.
When the POI creation is the only option in the dropdown, the entry is automatically selected and apparently Chromium does not trigger an event when this element is clicked.
On firefox, this problem doesn't exist, because the event is triggered every time interdependently of whether it's already pre-selected.

PR #1174, opened 2022-02-15
Fix POI integration in event form and list

timoludwig:
### Short description

This fixes a few problems with POIs in the event form and list

### Proposed changes

- Fix filtering for locations in event list
- Redesign POI selection in event form
- Do not preselect unused location for new event

### Resolved issues

Fixes: #1166
Fixes: #1172
Fixes: #1173


ulliholtgrave (approved):
Looks nice and works as expected :+1: Thanks :)

Issue #1182, opened 2022-02-16
Password reset not working


timoludwig:
### Describe the Bug

The password reset does not work at the moment

### Steps to Reproduce

1. Go to login form
2. Try to reset password
3. See error

### Expected Behavior

The user should get an email with the password reset link

### Actual Behavior

A `NoReverseMatch` at `/reset-password/` appears.

### Additional Information

<details>
  <summary>Traceback</summary>
    
  ```javascript
Feb 16 09:19:34 ERROR   django.request - Internal Server Error: /reset-password/
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/views/authentication/password_reset_view.py", line 46, in dispatch
    return super().dispatch(*args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/utils/decorators.py", line 130, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/views.py", line 222, in dispatch
    return super().dispatch(*args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/generic/edit.py", line 142, in post
    return self.form_valid(form)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/views/authentication/password_reset_view.py", line 78, in form_valid
    return super().form_valid(form)
[...]

PR #1183, opened 2022-02-16
Fix password reset

timoludwig:
### Short description

Fix the password rest urls

### Proposed changes

- Add tests for public views
    - These views are just tested for anonymous users because they cause a redirect for logged-in users
- Fix password reset urls
    - The uidb64 of the password reset views are not real uuids, just treat them as strings
    - Add test to prevent this error in the future

### Resolved issues

Fixes: #1182


ulliholtgrave (approved):
Thanks, works expected :+1: This was something I unfortunately didn't check after your urls PR.

Issue #1176, opened 2022-02-16
Live content for current region


svenseeberg:
### Describe the Bug
It is currently not possible to select a live content page for the own/current region

### Steps to Reproduce
1. Create 2 pages in a region
2. Edit page 1 and try to select page 2 as live content

### Expected Behavior
The own region should be available.

### Actual Behavior
The own region should is not listed and no pages are listed under `------`

PR #1184, opened 2022-02-16
Allow user to embed live content from current region

membralala:
### Short description

This PR should enable the option to embed live content from a site which belongs to the currently selected region. 

### Proposed changes

- Remove an initial query, that filters the current region from the 'mirrored_page_region' 

### Resolved issues

Fixes: #1176


svenseeberg (approved):

timoludwig (approved):
Thanks! :+1:

Issue #1164, opened 2022-02-15
Cancelling translation process not possible


timoludwig:
### Describe the Bug

When a page is currently in translation, its not possible to cancel this process anymore.

### Steps to Reproduce

1. Go to page tree
2. Export page as xliff
3. Open page form that is in translation
4. Click on "Abort translation process"

### Expected Behavior

The process should be aborted and the UI should be updated

### Actual Behavior

Nothing happens and the ajax view returns a 500 error code

### Additional Information

<details>
  <summary>Traceback</summary>
  
  ```
Feb 15 01:15:50 ERROR   django.request - Internal Server Error: /ajax/nuernberg/post_translation_state
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.8/site-packages/django/views/decorators/http.py", line 40, in inner
    return func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.8/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.8/site-packages/integreat_cms/cms/decorators.py", line 112, in wrap
    return function(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.8/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.8/site-packages/integreat_cms/cms/views/pages/page_actions.py", line 395, in post_translation_state_ajax
    page_translation = page.get_translation(target_language.slug)
AttributeError: 'str' object has no attribute 'slug'
  ```
  
</details>

PR #1186, opened 2022-02-16
Fix possibility to cancel translation process

timoludwig:
### Short description

Fix possibility to cancel translation process

### Proposed changes

- Fix possibility to cancel translation process
- Set default django settings in dev tools for non-docker users

### Resolved issues

Fixes: #1164


svenseeberg:
Do we need a bulk action to cancel many translations at once?

timoludwig:
> Do we need a bulk action to cancel many translations at once?

I think if so, we should open another feature request for this. Currently, we only have the (broken) possibility to cancel the translation via the form...

svenseeberg:
Okay, the rest works in my tests. Except for uploading XLIFF files. But that is another issue,

Issue #1185, opened 2022-02-16
Feedback returns 405 Method Not Allowed


svenseeberg:
### Describe the Bug
When sending feedback, for example about a page or search result, the API returns a `405 Method Not Allowed`

### Steps to Reproduce

1. Go to https://webnext.malte-app.de
2. Open the JS console and enter `window.localStorage.setItem('api-url', 'https://malte-test.tuerantuer.org');`
3. Send feedback
4. Look into the network developer tool or log file

### Expected Behavior
The API should say 200 OK and save the feedback.

### Actual Behavior
```
Feb 16 16:53:47 WARNING django.request - Method Not Allowed: /baddoberan/de/wp-json/extensions/v3/feedback/search/
```

PR #1187, opened 2022-02-16
Fix feedback API endpoints

timoludwig:
### Short description

Fix feedback API endpoints

### Proposed changes

- Add tests for feedback API (unfortunately, they just test the canonical URL at the moment, so they would not have caught this bug - but better than nothing)
- Fix feedback API endpoints
    - Enable both views - with trailing slash and without

### Resolved issues

Fixes: #1185


svenseeberg (approved):
Tested: works! Thanks a lot.

Issue #1190, opened 2022-02-17
Add possibility to enter custom prefix


timoludwig:
### Motivation

In some cases, the predefined choices for the administrative divisions are not sufficient.

### Proposed Solution

Add the possibility to enter custom prefixes, e.g. with an additional database field

### Alternatives

Add the possibility to override the available choices for the administrative division. This however has the problem that the GVZ API won't work anymore for these custom choices, which means we would have to implement special case handling for these.

PR #1191, opened 2022-02-18
Add possibility to set custom region prefix

timoludwig:
### Short description

Add possibility to set custom region prefix

### Proposed changes

- Add field 'custom_prefix' to region model
- Add field to region form and add custom validation to make sure it is not used when the prefix is identical to an administrative division
- Add custom prefix to test data

### Resolved issues

Fixes: #1190


integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -139,6 +141,40 @@ def clean(self):
...
+        # Validate custom prefix
+        if cleaned_data.get("custom_prefix"):
+            # Get the administrative divisions as conflicting options
+            administrative_divisions = [
+                label
+                for choice, label in self.fields["administrative_division"].choices
+            ]
+            if self.instance.default_language:
+                # If region already has a default language, translate the labels
+                with override(self.instance.default_language.slug):
```
svenseeberg, line 153: We could say that we check if the string is in any existing translation of prefixes. But that is probably not a case that will happen that often.

svenseeberg (approved):
Works as expected. I see no problems in the PR.

integreat_cms/cms/forms/regions/region_form.py
timoludwig: @svenseeberg I changed the validation method to check for all existing translation. Can you test again? Thanks!

svenseeberg: Works as expected. Thanks again :-)

PR #1194, opened 2022-02-18
Fix `TypeError` in XLIFF import view

timoludwig:
### Short description

This fixes the error:
`TypeError: join() argument must be str, bytes, or os.PathLike object, not 'UUID'`


ulliholtgrave (approved):
Thanks, that fixes the error :)

Issue #1175, opened 2022-02-16
Empty tag for empty pages with non-archived children


svenseeberg:
### Motivation
At the moment, the "empty" page label only works when a page has an empty content and no subpages.
There is an edge case where an empty page has only subpages which are archived, so just inspecting the page object with `is_leaf()` is not sufficient to check whether the empty tag should be shown. Instead, the database needs to be queried to check whether all subpages are archived or not.

### Proposed Solution
Since this means that we need to query `n+1` levels from the database to show `n` levels correctly in the tree, we should start with fetching the additional layer right after the page has finished loading:
- #1158

After this is implemented, we can use the additional information to correctly render the empty tag and insert it in those cases where only archived children exist. This is also connected to the following issue:
- #1244

which can probably solved together with this one.
### Alternatives
Alternatively, we can update the page tree view to query the first two levels instead of just the root pages on initial load. And on every expanding operation, all potential next levels need to be fetched to show the empty page tag correctly for subpages.

david-venhoff:
Note: This is right now blocked on #1158

svenseeberg:
@david-venhoff wouldn't it make more sense to calculate the data on the server side?

timoludwig:
@svenseeberg In any case, we need to query n+1 levels of pages from the database to show n levels in the page tree. So even when we use the new function, it would still increase the database load, because it would require to fetch the first two layers of pages instead of just the root pages. This was my thought process behind the proposal of combining this ticket with #1158 and query the additional levels dynamically.
Apart from that, this ticket is only about the archived flag (since it's about the page tree and we want to show pages without public translation in the page tree)...

svenseeberg:
@timoludwig can you update the initial post? I'm a little confused right now what this issue is actually about.

MizukiTemma:
Currently no empty tag is shown even if a page has only empty descendants. Should this remain so?

Then, I guess we only have to check whether a page as at least one non-archied child, becuase all the later descendants (grand children, grand grand children......) are all implecity but archived if all the children pages are archived. But I feel I'm missing something 🤔

svenseeberg:
> Currently no empty tag is shown even if a page has only empty descendants. Should this remain so?

Yes. I think we should check the content of child nodes. That would be too much logic and would become increasingly difficult to explain.

[...]

Issue #1158, opened 2022-02-10
Prefetch subpages in advance


timoludwig:
### Motivation

In PR #1149, we introduced the dynamic fetching of subpages in the page tree. Everytime the child pages of a parent page are expanded, an ajax request is performed to retrieve the result from the server, which causes the expansion to take a small loading time:

![Peek 2022-02-10 13-11](https://user-images.githubusercontent.com/16021269/153406471-9dbdf081-ce78-4b49-abab-9cce3240e6b2.gif)

### Proposed Solution

It would be cool if this delay could be reduced, e.g. by automatically loading the subpages before the expansion button is clicked (but after the rest of the root pages are rendered completely, it shouldn't slow down the loading time of the initial page tree). We could recursively walk the tree completely or only load n+1 levels depending on which pages are currently expanded.
The subpages should then expand instantly after the button is clicked:

![Peek 2022-02-10 13-11-2](https://user-images.githubusercontent.com/16021269/153406503-60415b62-60c6-4021-a3fb-c27cb265d7f4.gif)

### Additional Context

There are two tickets that depend on this issue:
- #1175
- #1244

david-venhoff:
When implementing this, #1175 can probably implemented, too.
Right now this is only partially implemented: The empty label only gets shown if a page has no subpages at all, but it should also be shown, if e.g. all its subpages are archived

PR #1199, opened 2022-02-19
Don't show empty tag if the page has no subpages

david-venhoff:
### Short description
This pr hides the empty for a page if the page has no subpages.
Generally, this tag should be hidden if the page has no publicly visible subpages, however this is not implemented as part of this pr.
This work should be done together with #1158.

Additionally, this pr fixes a crash due to an invalid url when moving a page.

Partially solves: #1175 

### Proposed changes

- Check if the page is a leaf before adding the empty tag


timoludwig (commented):
Thanks a lot! :+1:
Could you add a small note to the changelog for the empty page tag?
The bug with drag & drop doesn't need an entry in the changelog, it was introduced in the current unreleased version, so it is not a real change in comparison to the last version.

david-venhoff:
Added the note :+1:

timoludwig (approved):
:tada:

> Also, should this pr close #1175?

No, I think it's good to leave it open as reference and add a note that it's blocked by the other issue.

svenseeberg:
I think the title of this PR is misleading :sweat_smile: We're not showing the empty tag if a page has children OR has no children and is empty.

PR #1200, opened 2022-02-19
Fix parent page select input

timoludwig:
### Short description

Fix parent page select input

### Proposed changes

- Make it possible again to select no parent page for root pages
- Optimize url patterns for page order table


svenseeberg (approved):
I cannot find any issues with the changes. Thanks!

Issue #1196, opened 2022-02-19
API hit tracking with Matomo


svenseeberg:
### Motivation
We need to track API hits in Matomo.

### Proposed Solution
Track the single page and full content endpoint requests with Matomo.

### Alternatives
None. Needs to be done.

### Additional Context
ASAP.

PR #1201, opened 2022-02-19
Track API requests with Matomo

svenseeberg:
### Short description
Track single page and pages API endpoint requests with Matomo.

### Proposed changes
Add a decorator to allow tracking API endpoints. The decorator spawns a thread which executes the HTTP request to Matomo. This improves the performance of the CMS API, as the request does not have to wait for the response from Matomo.

### Resolved issues
Fixes: #1196 

This mostly works, however the `cip` parameter (https://developer.matomo.org/api-reference/tracking-api) breaks the request.

### DIFF
--- a/integreat_cms/api/decorators.py
+++ b/integreat_cms/api/decorators.py
@@ -2,14 +2,23 @@
 This module includes functions that are used as decorators in the API endpoints.
 """
 import json
+import threading
+import logging
+import random
+import re
 
 from functools import wraps
 
+from urllib import request, parse, error
+
 from django.http import JsonResponse, Http404
 from django.views.decorators.csrf import csrf_exempt
+from django.conf import settings
 
 from ..cms.models import Region, Language
 from ..cms.constants import feedback_ratings
+
+logger = logging.getLogger(__name__)
 
 
 def feedback_handler(func):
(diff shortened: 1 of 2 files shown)


integreat_cms/api/decorators.py
```diff
@@ -117,3 +125,74 @@ def wrap(request, *args, **kwargs):
...
+
+        :param \**kwargs: The supplied kwargs
+        :type \**kwargs: dict
+
+        :return: The response of the given function or an 404 :class:`~django.http.JsonResponse`
+        :rtype: ~django.http.JsonResponse
+        """
+        data = {
+            "idsite": request.region.matomo_id,
+            "token_auth": request.region.matomo_token,
```
timoludwig, line 164: Since the `matomo_id` and `matomo_token` are not mandatory fields in the region form, I think we should check `if request.region.statistics_enabled:` before doing the matomo request.

integreat_cms/api/decorators.py
```diff
@@ -117,3 +125,74 @@ def wrap(request, *args, **kwargs):
...
+            "rec": 1,
+            "url": request.build_absolute_uri(),
+            "urlref": settings.BASE_URL,
+            "ua": request.META["HTTP_USER_AGENT"]
+            if "HTTP_USER_AGENT" in request.META
+            else "unknown user agent",
```
timoludwig, line 170:
```suggestion
            "ua": request.META.get("HTTP_USER_AGENT", "unknown user agent"),
```
I think using a default dict value here is more readable.

integreat_cms/api/decorators.py
```diff
@@ -117,3 +125,74 @@ def wrap(request, *args, **kwargs):
...
+        :type data: dict
+        """
+        data = parse.urlencode(data)
+        url = f"{host}/matomo.php?{data}"
+        req = request.Request(url)
+        try:
+            with request.urlopen(req):
+                pass
+        except error.HTTPError as e:
+            logger.error("Matomo Tracking API request failed with: %s", e)
```
timoludwig, line 196:
```suggestion
            logger.error(
                "Matomo Tracking API request to %r failed with: %s",
                # Mask auth token in log
                re.sub(r"&token_auth=[^&]+", "&token_auth=********", url),
                e,
            )
```
Maybe it could be helpful to know which request failed exactly?

timoludwig (commented):
Thanks a lot! :tada:
The async requests work as expected, I didn't test whether the tracking works correctly on Matomo, but looks good.

integreat_cms/api/decorators.py
svenseeberg: Indeed. I had the same though already as well.

integreat_cms/api/decorators.py
svenseeberg: Good point.

integreat_cms/api/decorators.py
svenseeberg: Indeed more readable.

integreat_cms/api/decorators.py
```diff
@@ -117,3 +125,79 @@ def wrap(request, *args, **kwargs):
...
+        url = f"{host}/matomo.php?{data}"
+        req = request.Request(url)
+        try:
+            with request.urlopen(req):
+                pass
+        except error.HTTPError as e:
+            logger.error(
+                "Matomo Tracking API request to %r failed with: %s",
+                # Mask auth token in log
+                re.sub(r"&token_auth=[^&]+", "&token_auth=********", url),
```
timoludwig, line 199: I guess `import re` is still missing. Apart from that, it looks good! :+1:

svenseeberg: Wondering why pylint did not catch this in my IDE :thinking:

svenseeberg: Answer: I did not source the environment and therefore pylint did not work as expected.

timoludwig (approved):
:rocket:

PR #1203, opened 2022-02-20
Change colouring, remove dummy items & streamline boxes

ulliholtgrave:
### Short description

This PR prepares the UI to be in productive mode.

### Proposed changes

- Introduce new color class "water" and use it for all kind of boxes (e.g. dashboard, content forms & other model forms)
- Remove all dummy elements from UI (e.g. Unused widgets, dummy views for App-size and their corresponding appearance in templates)
- Remove unused views from tests
- Streamline HTML content forms according to [djlint](https://www.djlint.com/)


integreat_cms/cms/templates/analytics/analytics_overview.html
```diff
@@ -7,19 +7,18 @@
         <h1 class="heading">{% trans 'Analytics' %}</h1>
         <p class="py-3 text-lg"> {% trans 'Here you will find a set of tools to check the success and correctness of your app. Click on the Details button of a tool to get further information.' %}</p>
         <div class="xl:flex gap-3 pt-4">
-            <div class="xl:w-1/2">
-                <!-- Broken Link Widget -->
-                {% include "analytics/_broken_links_widget.html" %}
-                <!-- Translation Status Widget -->
-                {% include "analytics/_translation_overview_widget.html" %}
```
timoludwig, line 14: Maybe we still need the dummy widget for the translation coverage, because it looks a bit weird to have an overview site for analytics, but then there is only the broken link checker aligned to the right and the translation coverage is only visible when clicking on the menu item? the current solution is more confusing than a dummy widget for the translation coverage I think...

timoludwig (commented):
Thanks a lot! :+1:

I noticed that the media library and user settings still use the old colors for their boxes, could you adapt them as well?

Also, I think the alignment of the language tabs in forms looks a bit weird now, maybe you could replace `py-2` with `py-[calc(0.75rem+2px)]` in [line 13](https://github.com/digitalfabrik/integreat-cms/blob/f5b8d7ae4c6612d8a9ea822b4c8022acf7f2177c/integreat_cms/cms/templates/_form_language_tabs.html#L13) and [line 15](https://github.com/digitalfabrik/integreat-cms/blob/f5b8d7ae4c6612d8a9ea822b4c8022acf7f2177c/integreat_cms/cms/templates/_form_language_tabs.html#L15) of [integreat_cms/cms/templates/_form_language_tabs.html](https://github.com/digitalfabrik/integreat-cms/blob/f5b8d7ae4c6612d8a9ea822b4c8022acf7f2177c/integreat_cms/cms/templates/_form_language_tabs.html)? Sure, this padding isn't optimal, but at least it's now the same height as the sidebar header. The language tab design still looks a bit different than the sidebar boxes now, so probably we need a redesign of those in the future to make sure it looks like a unified design.

integreat_cms/cms/templates/dashboard/_dummy_widget.html
```diff
@@ -1,18 +0,0 @@
...
-    eye
-{% endblock %}
-
-{% block collapsible_box_title %}
-    {% trans 'At a glance' %}
-{% endblock %}
-
-{% block collapsible_box_content %}
-    <div class="p-3 w-full">
-        <!-- TODO -->
```
svenseeberg, line 16: We can add a text like "Upcoming feature" in here to make it easier to understand why it's there.

ulliholtgrave: This is just for the admins and I guess that we'll have some changes to the dashboard as well. Therefore, I prefer to just get rid of it :)

svenseeberg:
Also thanks for your work on this. I think it would be cool to merge this before the migration tomorrow :-) If you have time to include Timo's suggestions, I'd welcome this. But in general I have no problem with merging this as is and iterating/improving later.

integreat_cms/cms/templates/analytics/analytics_overview.html
ulliholtgrave: You are right, it's back again.

PR #1208, opened 2022-02-21
Update permissions of roles

timoludwig:
### Short description

Update permissions of roles

### Proposed changes

- Grant permission to publish events to editor
- Revoke permission to view feedback and imprint for editor and event manager roles


svenseeberg (approved):
Looks like magic :magic_wand:

PR #1210, opened 2022-02-21
Support legacy PDF API, fixes #1209

### LINKED ISSUE #1209: PDF Export URL pattern
### Describe the Bug
The web app calls `/REGION/LANG/wp-json/ig-mpdf/v1/pdf` to export a PDF which returns a 404. Our API currently uses `REGION/LANG/pdf`.

The normal mapping does not work, as we

### Steps to Reproduce

```
curl 'https://malte-test.tuerantuer.org/joerdenstorf/de/wp-json/ig-mpdf/v1/pdf'
```

svenseeberg:
### Short description
Route the PDF creation endpoint to the Integreat APIv3 PDF creation URL.

### Proposed changes
Add a URL pattern that routes the view `pdf_export` to `/REGION/LAN/wp-json/ig-mpdf/v1/pdf`

### Resolved issues
Fixes: #1209

### DIFF
--- a/integreat_cms/api/urls.py
+++ b/integreat_cms/api/urls.py
@@ -136,6 +136,11 @@ urlpatterns = [
                     "<slug:language_slug>/wp-json/extensions/v3/",
                     include(content_api_urlpatterns),
                 ),
+                path(
+                    "<slug:language_slug>/wp-json/ig-mpdf/v1/pdf/",
+                    pdf_export,
+                    name="pdf_export",
+                ),
             ]
         ),
     ),


integreat_cms/api/urls.py
```diff
@@ -136,6 +136,11 @@
                     "<slug:language_slug>/wp-json/extensions/v3/",
                     include(content_api_urlpatterns),
                 ),
+                path(
+                    "<slug:language_slug>/wp-json/ig-mpdf/v1/pdf",
```
timoludwig, line 140:
```suggestion
                    "<slug:language_slug>/wp-json/ig-mpdf/v1/pdf/",
```
I think we should use the canonical url with trailing slash here - Django will then automatically redirect to this path if the trailing slash is missing (this doesn't work the other way round, so if the path is defined without trailing shlash, requesting the endpoint with trailing slash will lead to a 404. But ok, since it's a legacy endpoint, we probably don't need to support the url with trailing slash).

timoludwig (approved):
Thanks! :+1:
Can you squash the commits?

PR #1212, opened 2022-02-21
Only show xliff export option for expert users

timoludwig:
### Short description

Only show xliff export option for expert users


svenseeberg (approved):
LGTM

PR #1214, opened 2022-02-21
Fix API return format of event location

timoludwig:
### Short description

Fix API return format of event location


svenseeberg (approved):
That was quick. Looks good!

PR #1218, opened 2022-02-21
Fix saving of first root node

timoludwig:
### Short description

Fix saving of first root node

### Proposed changes

- Add `CustomTreeNodeForm` for pages and language tree nodes
- Allow moving to siblings of other region's root nodes


svenseeberg (approved):
I tested:
- Editing the first page
- Editing other pages
- Moving other pages into pole position and then editing
- Moving pages in all layers to first, last and in between positions
- Moving pages in the page form and page tree

everything worked

timoludwig:
Ok, thanks! I just updated a few docstrings and added debug logging.

PR #1219, opened 2022-02-21
Fix layout in new page form on large screens

timoludwig:
### Short description

Fix layout in new page form on large screens (when you don't have a large screen you can see the broken behaviour when you zoom out)

### Proposed changes

- Show high dpi grid only when page already exists

This is a low prio PR, does not need to be reviewed today.


svenseeberg (approved):

ulliholtgrave (approved):
Seems so be okay :)

PR #1223, opened 2022-02-21
Remove icon from imprint api response

ulliholtgrave:
### Short description

This PR removes the icon field from the api response as our imprint model doesn't have it anymore.


timoludwig (commented):
Thanks, looks good! :+1:

Can you add a small note to the changelog?

timoludwig (approved):
:+1:

Issue #1227, opened 2022-02-21
Imprint broken


svenseeberg:
### Describe the Bug
The app cannot display the data from our imprint endpoint.

### Steps to Reproduce

1. Open any imprint on malteapp.de

ulliholtgrave:
Fixed in the last release :)

svenseeberg:
Sadly there is still another problem.

svenseeberg:
https://cms.malteapp.de/barth/de/wp-json/extensions/v3/disclaimer is missing the path
https://malteser.tuerantuer.org/barth/de/wp-json/extensions/v3/disclaimer is missing the path

PR #1231, opened 2022-02-22
Add path field to imprint response

ulliholtgrave:
### Short description

This PR is correcting the path and url field in the api response for the imprint

Fixes: #1227

### DIFF
--- a/integreat_cms/api/v3/imprint.py
+++ b/integreat_cms/api/v3/imprint.py
@@ -1,6 +1,7 @@
 """
 This module includes functions related to the imprint API endpoint.
 """
+from django.conf import settings
 from django.http import JsonResponse
 
 from ..decorators import json_response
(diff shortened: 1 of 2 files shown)


svenseeberg (approved):
Works. Changelog still missing.

integreat_cms/api/v3/imprint.py
```diff
@@ -18,7 +19,8 @@ def transform_imprint(imprint_translation):
     """
     return {
         "id": imprint_translation.id,
-        "url": imprint_translation.get_absolute_url(),
+        "url": settings.BASE_URL + imprint_translation.get_absolute_url(),
+        "path": imprint_translation.get_absolute_url(),
```
timoludwig, line 23: I believe `get_absolute_url()` needs a database request, so probably it should be executed only once and be stored to a variable?

ulliholtgrave: Okay... :joy:

timoludwig (approved):
:+1:

Issue #1228, opened 2022-02-22
Fix development release pipeline


svenseeberg:
### Describe the Bug
The development release pipeline is currently broken due to a problem with detecting the next version.

Solution: we use a different naming pattern. I'd suggest one of the following:
- `YYYY.MM-COMMIT_HASH`
- `YYYY.MM-COMMIT_NUM` (`COMMIT_NUM` is the number of the commit since first commit)

### Steps to Reproduce

1. Merge a PR to develop

### Expected Behavior
Create a beta release.

### Actual Behavior
The script to detect the next version runs in a circle.

PR #1237, opened 2022-02-22
Fix bumping of alpha versions

timoludwig:
### Short description

Fix bumping of alpha versions

### Proposed changes

- Do not iterate in a loop but query the latest version and do the bumping job twice

### Resolved issues

Fixes: #1228


ulliholtgrave (approved):
Let's try it :)

PR #1230, opened 2022-02-22
API: do not use cached properties, fixes #1226

svenseeberg:
Do not use cached properties.

Fixes #1226

PR #1238, opened 2022-02-22
Fix handling of archived pages in API

### LINKED ISSUE #1222: Error in children page endpoint
### Describe the Bug

When a page is opened in the app that does not have a public parent page, the children endpoint returns an error

### Steps to Reproduce

1. Create parent page without translation in a specific language
2. Create a sub page with a translation in this language
3. Access children endpoint

### Expected Behavior

Not sure what the expected output is here. Should we not show anything because the page is also hidden from the category view, or should we create the impression that the page would be a root page?

### Actual Behavior

```
Feb 21 21:23:13 ERROR   django.request - Internal Server Error: /gemeinschaftsunterkunft/en/wp-json/extensions/v3/children/
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/api/decorators.py", line 164, in wrap
    return func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/api/decorators.py", line 124, in wrap
    return function(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/api/v3/pages.py", line 205, in children
    result.append(transform_page(public_translation))
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/api/v3/pages.py", line 24, in transform_page
    parent_absolute_url = parent_page.get_public_translation(
AttributeError: 'NoneType' object has no attribute 'get_absolute_url'
Feb 21 21:23:13 WARNING django.request - Not Found: /gemeinschaftsunterkunft/en/wp-json/extensions/v3/parents/
```

### Additional Information

If we decide to show nothing, we have to make sure that the `available_languages` mapping of all other translations of this page do not contain this translation.

### LINKED ISSUE #1235: Archive pages show up in childen/parents end points
### Describe the Bug
The children and parents end points return archived pages in their result.

### Steps to Reproduce

1. Archive a child page
2. Open the children end point for the parent page.

### Expected Behavior
Archived pages should not be in any result.

svenseeberg:
* Add translation check in cache_tree function. This builds a list of pages that should be available via the API.
* The children endpoint validates returned pages against the cache_tree with a language filter.
* Skip page in transform function if parent is archived.
* Raise 404 if the slug argument resolves to an archived page.

### Resolved issues
- Fixes #1222 
- Fixes #1235


integreat_cms/cms/models/pages/page.py
```diff
@@ -25,14 +25,24 @@ class PageQuerySet(NS_NodeQuerySet, ContentQuerySet):
...
         """
         Caches a page tree queryset in a python data structure.
 
+        Support for combinations of parameters:
+        |                  | language_slug = None | language_slug = str |
+        | ---------------- | -------------------- | ------------------- |
+        | archived = None  | yes                  | no                  |
+        | archived = True  | yes                  | no                  |
+        | archived = False | yes                  | yes                 |
+
```
timoludwig, line 38:
```suggestion
```
Since there is only one valid combination where `language_slug` is not `None`, I'd just mention this in the param documentation.

integreat_cms/cms/models/pages/page.py
```diff
@@ -25,14 +25,24 @@ class PageQuerySet(NS_NodeQuerySet, ContentQuerySet):
...
         :param archived: Whether the pages should be limited to either archived  or non-archived pages.
                          If not passed or ``None``, both archived and non-archived pages are returned.
         :type archived: bool
 
+        :param language_slug: Code to identify the desired language
```
timoludwig, line 43:
```suggestion
        :param language_slug: Code to identify the desired language (optional, requires ``archived`` to be ``False``)
```

integreat_cms/cms/models/pages/page.py
```diff
@@ -25,14 +25,24 @@ class PageQuerySet(NS_NodeQuerySet, ContentQuerySet):
...
+        :type language_slug: str
+
         :return: A list of pages with cached children, descendants and ancestors and a list of all skipped pages
         :rtype: tuple [ list, list ]
         """
```
timoludwig, line 48:
```suggestion
        """
        if language_slug is not None and archived is not False:
                raise ValueError("archived must be False in order to filter for public translations by language_slug")
```

timoludwig (commented):
Thanks a lot! :+1:

This fixes the two linked issues at acceptable performance drawbacks. I noticed we have pretty significant problems with larger numbers of pages (e.g. with 512 pages, the complete pages endpoint took over 1000 sql queries and 30 seconds to load while the children endpoint of root pages with depth 2 took 2500 queries and over 1 min) which should be a remnant of #1230 and I'm not quite sure how easy it is to get the number down again after merging these changes, but since we have to do a thorough makeover of the children endpoint anyway, I guess this is acceptable.

integreat_cms/api/v3/pages.py
```diff
@@ -212,20 +228,26 @@ def children(request, region_slug, language_slug):
         # like in wordpress depth = 0 will return no results in this case
         depth = depth - 1
     result = []
+    # pylint: disable=unused-variable
+    public_region_pages, skipped_pages = request.region.pages.filter(
+        explicitly_archived=False
+    ).cache_tree(archived=False, language_slug=language_slug)
```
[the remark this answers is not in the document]
svenseeberg: Yes, indeed. It will mostly reduce bandwidth required for the time being. We then need a fast method that builds the tree for one root node / tree ID.

CHANGELOG.md
```diff
@@ -3,7 +3,7 @@ UNRELEASED
 
 * [ [#1227](https://github.com/digitalfabrik/integreat-cms/issues/1227) ] Correct URL and Path field in imprint API
 * [ [#1226](https://github.com/digitalfabrik/integreat-cms/issues/1226) ] Skip cache for childen and parents end points
-
+* [ [#1222](https://github.com/digitalfabrik/integreat-cms/issues/1222) ] Fix missing translations and archived pages in API
```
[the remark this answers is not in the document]
svenseeberg: And I thought I had done that :joy:. Thanks for catching that.

integreat_cms/api/v3/pages.py
```diff
@@ -142,6 +151,12 @@ def get_single_page(request, language_slug):
     else:
         raise RuntimeError("Either the id or the url parameter is required.")
 
+    if page.explicitly_archived:
+        raise Http404("No matching page translation found for url.")
+
+    # Check if any ancestor is archived -> will raise a 404
+    get_public_ancestor_translations(page, language_slug)
```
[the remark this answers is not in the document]
timoludwig: Given the urgency of this PR, the comment above is not important enough I think... Let's ignore it for the time being and improve later.

integreat_cms/cms/models/pages/page.py
```diff
@@ -212,6 +227,28 @@ def implicitly_archived(self):
...
+        if not bool(self.explicitly_archived_ancestors):
+            return False
+        return bool(
+            [
+                ancestor
+                for ancestor in self.get_cached_ancestors()
+                if ancestor.get_translation(language_slug)
+            ]
+        )
+
```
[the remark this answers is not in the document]
svenseeberg: True.

integreat_cms/api/v3/pages.py
```diff
@@ -20,25 +20,34 @@ def transform_page(page_translation):
...
     :return: data necessary for API
     :rtype: dict
     """
+    fallback_parent = {
+        "id": 0,
+        "url": None,
+        "path": None,
+    }
+
     parent_page = page_translation.page.parent
```
timoludwig, line 34:
```suggestion
    parent_page = page_translation.page.cached_parent
```

Do you think we could add this cached value with this PR again? With cacheops disabled, I think this will work fine and should reduce the number of queries tremendously.

integreat_cms/api/v3/pages.py
```diff
@@ -20,25 +20,34 @@ def transform_page(page_translation):
...
+        )
+        if parent_public_translation:
+            parent_absolute_url = parent_public_translation.get_absolute_url()
+            parent = {
+                "id": parent_page.id,
+                "url": settings.BASE_URL + parent_absolute_url,
+                "path": parent_absolute_url,
+            }
+        else:
+            parent = fallback_parent
```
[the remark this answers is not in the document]
timoludwig: If I understood correctly, we already circumvent this problem earlier by not calling this method for pages without public parents, right? So just throwing an exception without handling it could also be fine.

integreat_cms/api/v3/pages.py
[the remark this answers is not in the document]
svenseeberg: This works. We now only create the tree for the mptt trees we actually need (AFAICT).

integreat_cms/api/v3/pages.py
svenseeberg: Sure.

integreat_cms/api/v3/pages.py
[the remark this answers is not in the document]
svenseeberg: Okay, for future reference: we should probably also use the cache_tree with a query set filtered by tree_id. That is something we need to do to be safe in most cases.

integreat_cms/api/v3/pages.py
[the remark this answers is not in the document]
svenseeberg: Indeed. I think I had to implement this before the cache_tree already filtered by publicly visible content. And we don't want 404s in the normal pages endpoint ;-) Fixed now.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ UNRELEASED
 
 * [ [#1227](https://github.com/digitalfabrik/integreat-cms/issues/1227) ] Correct URL and Path field in imprint API
 * [ [#1226](https://github.com/digitalfabrik/integreat-cms/issues/1226) ] Skip cache for childen and parents end points
```
```suggestion
```
I'm not sure, should we remove entries from the changelog that do not make it into the release? Strictly speaking, it's not a change in comparison to the previous version.

integreat_cms/api/v3/pages.py
```diff
@@ -20,25 +20,34 @@ def transform_page(page_translation):
...
+        )
+        if parent_public_translation:
+            parent_absolute_url = parent_public_translation.get_absolute_url()
+            parent = {
+                "id": parent_page.id,
+                "url": settings.BASE_URL + parent_absolute_url,
+                "path": parent_absolute_url,
+            }
+        else:
+            raise Http404("No Page matches the given url or id.")
```
timoludwig, line 47:
```suggestion
            logger.info(
                "The parent %r of %r does not have a public translation in %r",
                parent_page,
                page_translation.page,
                page_translation.language,
            )
            raise Http404("No Page matches the given url or id.")
```
Since this error message is identical to the other ones that occur when the page really doesn't exist, I'd like to have a little more information to detect potential problems.

integreat_cms/api/v3/pages.py
```diff
@@ -212,20 +228,26 @@ def children(request, region_slug, language_slug):
         # like in wordpress depth = 0 will return no results in this case
         depth = depth - 1
     result = []
+    # pylint: disable=unused-variable
+    public_region_pages, skipped_pages = request.region.pages.filter(
+        explicitly_archived=False, tree_id__in=[page.tree_id for page in root_pages]
+    ).cache_tree(archived=False, language_slug=language_slug)
```
timoludwig, line 234: Could you add the `language_slug=language_slug` argument to the "all pages" endpoint as well? Currently, the endpoint throws a 404 if there exists at least one page with a non-archived parent page without public translation. In theory, we could also add this special edge case to the tests, but not urgent I think, let's add this later.

timoludwig (commented):
Thanks! I noticed one remaining crucial problem (which should be fixable by adding the new `language_slug` argument for this endpoint as well), otherwise the PR looks good! :+1:

integreat_cms/api/v3/pages.py
svenseeberg: Sure.

timoludwig (approved):
Awesome, thanks a lot! :tada:

Issue #1236, opened 2022-02-22
Saving of Somali translation in Karlsruhe (Malte) does not work due to "relativ zu: ... 1522 ist keine gültige Auswahl" 


FritjofK:
### Describe the Bug

Saving of a translation in Somali is not possible. 

### Steps to Reproduce

1. Go to ['...'] https://cms.malteapp.de/leakarlsruhefelsstrasse/pages/so/1523/edit/
2. Click on '....' Aktualisieren
3. 

![image](https://user-images.githubusercontent.com/16462172/155142139-3380d096-f76e-411e-8594-f5135bf1b393.png)

### Additional Information

Experienced before during "migration night" but not exactly sure, where.

timoludwig:
This is probably related to #1131 - changing the language tree has weird consequences for pages.

Issue #1131, opened 2022-01-25
Language Tree isn't handled correctly by cache in Page Tree


ulliholtgrave:
### Describe the Bug
So far, our Table rows aren't consistent after we delete a language tree node and recreate it. We should probably empty the cacheops cache after we delete a language tree node.

PR #1241, opened 2022-02-22
Flush Cacheops when modifying laguage tree nodes

svenseeberg:
### Short description
Flush the Cacheops Cache when modifying the language tree of a region.

I'm not really sure if this actually fixes the issue at hand. This probably needs some thorough testing. If nobody sees a reason why this should not help, then I suggest to deploy to the test server and invest some time into clicking around.

### Proposed changes
When moving a language tree node, flush all related objects in the region from the cache.

### Resolved issues
Fixes: #1131
Fixes: #1236

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ UNRELEASED
 
 * [ [#1227](https://github.com/digitalfabrik/integreat-cms/issues/1227) ] Correct URL and Path field in imprint API
 * [ [#1222](https://github.com/digitalfabrik/integreat-cms/issues/1222) ] Fix missing translations and archived pages in API
+* [ [#1131](https://github.com/digitalfabrik/integreat-cms/issues/1131) ] Flush Cache of related objects when changing a tree
 
 
 2022.2.3
--- a/integreat_cms/cms/forms/language_tree/language_tree_node_form.py
+++ b/integreat_cms/cms/forms/language_tree/language_tree_node_form.py
@@ -4,6 +4,8 @@ from django import forms
 from django.utils.text import capfirst
 from django.utils.translation import ugettext_lazy as _
 
+from cacheops import invalidate_obj
+
 from ..custom_model_form import CustomModelForm
 from ..custom_tree_node_form import CustomTreeNodeForm
 from ...models import Language, LanguageTreeNode
@@ -119,3 +121,25 @@ class LanguageTreeNodeForm(CustomModelForm, CustomTreeNodeForm):
             "LanguageTreeNodeForm validated [2] with cleaned data %r", cleaned_data
         )
         return cleaned_data
+
+    def save(self, commit=True):
+        """
+        This method extends the default ``save()``-method of the base :class:`~django.forms.ModelForm` to flush
+        the cache after commiting.
+
+        :param commit: Whether or not the changes should be written to the database
+        :type commit: bool
+
+        :return: The saved page translation object
+        :rtype: ~integreat_cms.cms.models.pages.page_translation.PageTranslation
+        """
+        # Save CustomModelForm and flush Cache
+        result = super().save(commit=commit)
+
+        for page in self.instance.region.pages.all():
+            invalidate_obj(page)
+        for poi in self.instance.region.pois.all():
+            invalidate_obj(poi)
+        for event in self.instance.region.events.all():
+            invalidate_obj(event)
+        return result
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-02-23 09:11+0000\n"
+"POT-Creation-Date: 2022-02-23 15:52+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: Integreat <info@integreat-app.de>\n"
 "Language-Team: Integreat <info@integreat-app.de>\n"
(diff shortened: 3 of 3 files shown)


svenseeberg:
We could even try to reinstate the changes made in #1230 wit this PR, deploy to the test system, and see what happens.

integreat_cms/cms/forms/language_tree/language_tree_node_form.py
```diff
@@ -119,3 +121,19 @@ def clean(self):
...
+
+        :param commit: Whether or not the changes should be written to the database
+        :type commit: bool
+
+        :return: The saved page translation object
+        :rtype: ~integreat_cms.cms.models.pages.page_translation.PageTranslation
+        """
+        # Save CustomModelForm and flush Cache
+        result = super().save(commit=commit)
+        invalidate_model(LanguageTreeNode)
```
timoludwig, line 138:
```suggestion
        invalidate_model(Page)
        invalidate_model(Event)
        invalidate_model(POI)
```
This already happens implicitly here:
https://github.com/digitalfabrik/integreat-cms/blob/8c06962b2549f3c6839bc38b4b8f25d95a6ce7e6/integreat_cms/cms/models/abstract_tree_node.py#L296-L300
Each node is "moved" after saving, no matter if the position changed or not.

However, I guess it would be good to invalidate all content models when the language tree is changed. I think cacheops would do this automatically if the translation models had a foreign key to the language tree node model, but since they're just related to the language model, cacheops doesn't think they have anything to do with the language tree. Maybe it's also required to invalidate the translation models, but I presume they are implicitly invalidated because they have a foreign key relation to the content models.

integreat_cms/cms/forms/pages/page_form.py
```diff
@@ -196,3 +198,19 @@ def get_publisher_queryset(self):
...
+        :param commit: Whether or not the changes should be written to the database
+        :type commit: bool
+
+        :return: The saved page translation object
+        :rtype: ~integreat_cms.cms.models.pages.page_translation.PageTranslation
+        """
+        # Save CustomModelForm and flush Cache
+        result = super().save(commit=commit)
+        invalidate_model(Page)
+        return result
```
timoludwig, line 216:
```suggestion
```
This also happens anyway because the `move()` method is called whenever a page is saved. And I guess it's not required to invalidate the `PageTranslation` model because it has a foreign key relation to the page model and thus invalidated anyway.

integreat_cms/cms/forms/pages/page_translation_form.py
```diff
@@ -84,8 +85,10 @@ def save(self, commit=True):
             self.instance.version += 1
             self.instance.pk = None
 
-        # Save CustomModelForm
-        return super().save(commit=commit)
+        # Save CustomModelForm and flush Cache
+        result = super().save(commit=commit)
+        invalidate_model(PageTranslation)
+        return result
```
timoludwig, line 91: I'm not sure, I think this is also implicitly done because the page translation is saved together with a page and saving a page triggers cache invalidation through the move method? Also, I'm not sure whether we really should invalidate the whole model here - maybe it's enough to invalidate all parents and children in case the slug changed (because urls might have changed?)...

timoludwig (commented):
Thanks! I would like to investigate further what exactly is happening and why. Otherwise, it might be better for performance to work without a cache at all than to have a cache and invalidate it all the time...

svenseeberg:
@timoludwig what do you think about not flushing the full model but iterating over all tree nodes of one region and flushing them? This would be a little more expensive during saving. But pretty safe to assume that we have lots more read than write operations. We could even reload the Cache after the write is done.

timoludwig:
> what do you think about not flushing the full model but iterating over all tree nodes of one region and flushing them?

@svenseeberg yes good idea! I'm kind of surprised that cacheops doesn't provide an invalidation method for querysets, but iterating should be fine.

integreat_cms/cms/forms/pages/page_translation_form.py
svenseeberg: Okay, lets drop this.

integreat_cms/cms/views/pages/page_actions.py
```diff
@@ -561,6 +562,9 @@ def move_page(request, region_slug, language_slug, page_id, target_id, position)
         messages.error(request, e)
         logger.exception(e)
 
+    for page in Page.objects.filter(region=region):
+        invalidate_obj(page)
+
```
timoludwig, line 567:
```suggestion
```
Again, I don't think this changes anything, because the page model inherits from abstract tree node and will get invalidated during move anyway:
https://github.com/digitalfabrik/integreat-cms/blob/8c06962b2549f3c6839bc38b4b8f25d95a6ce7e6/integreat_cms/cms/models/abstract_tree_node.py#L296-L300
We could think about invalidating only the region's tree nodes there, but actually I think this isn't enough because moving a root page inside a region can change all tree_ids of all other regions as well.

integreat_cms/cms/forms/language_tree/language_tree_node_form.py
```diff
@@ -4,9 +4,11 @@
 from django.utils.text import capfirst
 from django.utils.translation import ugettext_lazy as _
 
+from cacheops import invalidate_obj
+
 from ..custom_model_form import CustomModelForm
 from ..custom_tree_node_form import CustomTreeNodeForm
-from ...models import Language, LanguageTreeNode
+from ...models import Language, LanguageTreeNode, Page, Event, POI
```
timoludwig, line 11:
```suggestion
from ...models import Language, LanguageTreeNode
```

integreat_cms/cms/views/pages/page_actions.py
svenseeberg: Ah sorry, I misunderstood the implications.

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ UNRELEASED
 
 * [ [#1227](https://github.com/digitalfabrik/integreat-cms/issues/1227) ] Correct URL and Path field in imprint API
 * [ [#1222](https://github.com/digitalfabrik/integreat-cms/issues/1222) ] Fix missing translations and archived pages in API
+* [ [#1226](https://github.com/digitalfabrik/integreat-cms/issues/1226) ] Skip cache for childen and parents end points
```
```suggestion
```

timoludwig (approved):
:tada:

PR #1243, opened 2022-02-23
Add option to activate Matomo tracking, fixes #1242

### LINKED ISSUE #1242: Disable Matomo tracking setting
### Motivation
We want our test system to work normally, but we do not want to track API requests in the statistics. We should have a setting that only disables tracking API requests.

### Proposed Solution
Add a Setting to disable the tracking function (decorator).

### Alternatives
Use a different Matomo.

svenseeberg:
### Short description
Add an option to disable tracking of API requests. This is necessary when we want to connect to a Matomo instance in general but do not want to track API requests.

### Proposed changes
Add a `MATOMO_TRACKING` setting.

### Resolved issues
Fixes: #1242

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ UNRELEASED
 * [ [#1227](https://github.com/digitalfabrik/integreat-cms/issues/1227) ] Correct URL and Path field in imprint API
 * [ [#1222](https://github.com/digitalfabrik/integreat-cms/issues/1222) ] Fix missing translations and archived pages in API
 * [ [#1131](https://github.com/digitalfabrik/integreat-cms/issues/1131) ] Flush Cache of related objects when changing a tree
+* [ [#1242](https://github.com/digitalfabrik/integreat-cms/issues/1242) ] Add setting to activate Matomo tracking
 
 
 2022.2.3
--- a/integreat_cms/core/settings.py
+++ b/integreat_cms/core/settings.py
@@ -31,6 +31,11 @@ MATOMO_URL = os.environ.get(
     "INTEGREAT_CMS_MATOMO_URL", "https://statistics.integreat-app.de"
 )
 
+#: Enable tracking of API requests in Matomo
+MATOMO_TRACKING = bool(
+    strtobool(os.environ.get("INTEGREAT_CMS_MATOMO_TRACKING", "False"))
+)
+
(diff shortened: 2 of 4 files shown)


integreat_cms/core/settings.py
```diff
@@ -31,6 +31,9 @@
     "INTEGREAT_CMS_MATOMO_URL", "https://statistics.integreat-app.de"
 )
 
+#: Enable tracking of API requests in Matomo
+MATOMO_TRACKING = os.environ.get("INTEGREAT_CMS_MATOMO_TRACKING", False)
```
timoludwig, line 35:
```suggestion
MATOMO_TRACKING = bool(
    strtobool(os.environ.get("INTEGREAT_CMS_MATOMO_TRACKING", "False"))
)
```
If we just get the environment variable without parsing, we have the problem that setting `MATOMO_TRACKING = False` in the config file would evaluate to `True` because it's a non-empty string.

timoludwig (commented):
Thanks! Could you add a note to the changelog and document the new setting in the [example config](https://github.com/digitalfabrik/integreat-cms/blob/develop/example-configs/integreat-cms.ini)?

integreat_cms/core/settings.py
svenseeberg: Right, thanks :see_no_evil:

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ UNRELEASED
 
 * [ [#1227](https://github.com/digitalfabrik/integreat-cms/issues/1227) ] Correct URL and Path field in imprint API
 * [ [#1222](https://github.com/digitalfabrik/integreat-cms/issues/1222) ] Fix missing translations and archived pages in API
+* [ [#1226](https://github.com/digitalfabrik/integreat-cms/issues/1226) ] Skip cache for childen and parents end points
```
```suggestion
```

svenseeberg: Where the hack is this coming from?! I need to investigate what is going on here. This happens whenever I rebase on develop.

timoludwig (approved):
:rocket:

svenseeberg:
No idea why I have this wrong changelog entry appearing in the last PRs again and again :see_no_evil:

CHANGELOG.md
timoludwig: Maybe you have [rerere](https://git-scm.com/book/de/v2/Git-Tools-Rerere) enabled and git has remembered how you resolved this conflict the first time and repeats it now everytime it sees it again :see_no_evil:

PR #1246, opened 2022-02-23
Add push page API end point, fixes #1099

### LINKED ISSUE #1099: Push Content API
### Motivation
The push content APIv3 endpoint is still actively used to publish COVID-19 incidence information. If COVID-19 is still relevant after the Omicron surge, then this is blocking the migration.
https://github.com/digitalfabrik/cms/wiki/REST-APIv3-Documentation#push-content

Also, we either need to migrate the access tokens or update the Corona Push config.

### Proposed Solution
Add the end point. 

### Alternatives
Nobody cares about COVID-19 anymore

### Additional Context
This API endpoint provides a general solution for pushing data into Integreat and is probably always beneficial.

svenseeberg:
### Short description
Add the `pushpage` API end point (https://github.com/digitalfabrik/cms/wiki/REST-APIv3-Documentation#push-content)

### Proposed changes
- Add an `api_token` attribute to the Page model.
- Add an API endpoint.
- Add a Check Box to the page form that generates a token. The user is not allowed to enter a token manually.

### Resolved issues
Fixes: #1099


integreat_cms/static/src/js/pages/page-api-token.ts
```diff
@@ -0,0 +1,36 @@
...
+    let checkBox = document.getElementById("checkbox_api_token") as HTMLInputElement;
+    let text = document.getElementById("id_api_token") as HTMLInputElement;
+    let random = "";
+    if (checkBox.checked == true){
+      let random = "";
+      const characters ='ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+      for ( let i = 0; i < 24; i++ ) {
+        random += characters.charAt(Math.floor(Math.random() * characters.length));
+      }
+      text.value = random;
```
timoludwig, line 32:
```suggestion
    if (checkBox.checked){
      text.value = window.crypto.randomUUID();
      apiTokenContainer.classList.remove("hidden");
```
Since `Math.random()` is not cryptocraphically secure, I suggest using `window.crypto.randomUUID()` for this task (https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID). This only has the side-effect that the tokens are now 36 instead of 24 chars (but I see no reason why we should lower the entropy here). And it requires an update of  typescript (run `npm update`) because the type definition of this module was recently updated.
Oh, and [browser support](https://caniuse.com/mdn-api_crypto_randomuuid) isn't optimal, but I'm not sure if webpack does any magic here to make it available in older browsers... we should probably test this and maybe provide a fallback?

integreat_cms/cms/migrations/0010_page_api_token.py
```diff
@@ -0,0 +1,27 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="page",
+            name="api_token",
+            field=models.CharField(
+                blank=True,
+                help_text="API token to allow writing content to translations.",
+                max_length=24,
```
timoludwig, line 22:
```suggestion
                max_length=36,
```
If we use `Crypto.randomUUID()`, we need to increase the max length.

integreat_cms/cms/models/pages/page.py
```diff
@@ -188,6 +188,13 @@ class Page(AbstractTreeNode, AbstractBasePage):
             "This allows all members of the organization to edit and publish this page."
         ),
     )
+    api_token = models.CharField(
+        null=True,
+        blank=True,
+        max_length=24,
```
timoludwig, line 194:
```suggestion
        max_length=36,
```

timoludwig (commented):
Awesome, thanks a lot! :heart_eyes: 

In addition to my comments, I noticed that the api token gets removed when a non-expert user submits the form. So I suggest we remove the field from the form completely in such cases (which means that Django won't expect it to be submitted and won't overwrite the value in the database) by adding the follwing lines:
```
if not request.user.expert_mode:
    del page_form.fields["api_token"]
```
to [page_form_view.py:228](https://github.com/digitalfabrik/integreat-cms/blob/ff8efa4f689b61b974ab7c8fcdc61b63c871801f/integreat_cms/cms/views/pages/page_form_view.py#L228)

integreat_cms/static/src/js/pages/page-api-token.ts
timoludwig: I fit's too hard to do it cryptographically secure in the browser, we could also just say that the user has to enable the checkbox and submit the form, then the server calculates a secret, and the user can copy it afterwards.

integreat_cms/api/v3/pages.py
```diff
@@ -302,3 +305,58 @@ def get_public_ancestor_translations(current_page, language_slug):
...
+
+    :raises ~django.http.Http404: HTTP status 404 if the request is malformed or no page with the given id or url exists.
+
+    :return: JSON with the requested page ancestors
+    :rtype: ~django.http.JsonResponse
+    """
+    try:
+        data = json.loads(request.body)
+    except json.decoder.JSONDecodeError:
+        data = {}
```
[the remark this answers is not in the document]
svenseeberg: We could even say this is a 500 error. But yes, 405 makes more sense than

integreat_cms/static/src/js/pages/page-api-token.ts
svenseeberg: Yes, using a UUID is a good idea.

integreat_cms/static/src/js/pages/page-api-token.ts
```diff
@@ -0,0 +1,30 @@
...
+    let text = document.getElementById("id_api_token") as HTMLInputElement;
+    let apiTokenContainer = document.getElementById("api-token-container") as HTMLInputElement;
+    if (checkBox.checked){
+      text.value = window.crypto.randomUUID();
+      apiTokenContainer.classList.remove("hidden");
+    } else {
+      text.value = "";
+      apiTokenContainer.classList.add("hidden");
+    }
+}
```
timoludwig, line 30:
```suggestion
}

/**
 * Copy API token to clipboard
 *
 * @param event Event - The click event
 */
function copyApiToken(event: Event) {
    event.preventDefault();
    let text = document.getElementById("id_api_token") as HTMLInputElement;
    let copyButton = document.getElementById("copy-api-token");
    let successBox = document.getElementById("copy-api-token-success");
    // Copy value to clipboard
    copyToClipboard(text.value);
    // Indicate success by changing button
    copyButton.classList.add("hidden");
    successBox.classList.remove("hidden");
    // Revert button after 3 seconds
    setTimeout(function() {
        successBox.classList.add("hidden");
        copyButton.classList.remove("hidden");
    }, 3 * 1000);
}

```
Oh, I forgot to suggest my `copyApiToken()` function.

svenseeberg: Thanks, I'll copy this.

integreat_cms/static/src/js/pages/page-api-token.ts
```diff
@@ -0,0 +1,60 @@
+/*
+ * Generate and (un)set the API token.
+ */
+
+// Hack to get around the fact that randomUUID is not yet a part of the default TypeScript libs
+declare global {
+  interface Crypto {
+    randomUUID: () => string;
+  }
+}
```
timoludwig, line 10:
```suggestion
```
I'd prefer to just do an `npm update`, because newer versions of typescript already support this.

integreat_cms/static/src/js/pages/page-api-token.ts
```diff
@@ -0,0 +1,60 @@
...
+
+import { copyToClipboard } from "../copy-clipboard";
+
+// Set event handler
+window.addEventListener("load", () => {
+    document.getElementById("id_enable_api_token")?.addEventListener("change", setApiToken);
+    let checkbox = document.getElementById("checkbox_api_token") as HTMLInputElement;
```
timoludwig, line 17:
```suggestion
```

integreat_cms/cms/forms/pages/page_form.py
```diff
@@ -46,6 +46,12 @@ class PageForm(CustomModelForm, CustomTreeNodeForm):
         required=False,
         label=_("Source region for live content"),
     )
+    api_token = forms.CharField(
+        max_length=24,
+        min_length=24,
+        required=False,
+        label=_("Access token to update the page content"),
```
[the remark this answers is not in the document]
timoludwig: Adding `del self.cleaned_data["enable_api_token"]` to `_clean_cleaned_data()` is not resolved yet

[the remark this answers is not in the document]
timoludwig: This is required, otherwise submitting the new page form will fail.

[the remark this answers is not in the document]
svenseeberg: Done now :)

Issue #1052, opened 2021-12-07
Select all subpages when checking parent page


timoludwig:
### Motivation

In the page tree, typically a parent page should include all its subpages for all bulk actions.

### Proposed Solution

When selecting a parent page in the page tree, all its sub-pages should be selected as well. This also applies to de-selecting a page. After that, it should be possible to (de-)select individual subpages without changing the selection status of the parent page.

svenseeberg:
Very much related: the "select all" checkbox only selects parent pages now. I think it should also apply to children once they're expanded.

svenseeberg:
Basic concept: when executing a bulk action (XLIFF & PDF export), the view needs to check which checkboxes/pages are in the HTTP Post variables and which pages actually exist. All existing pages that are not in the HTTP POST inherit the "checked" status from the parent that is in the post.

Issue #1004, opened 2021-10-30
Create possibility to expand/collapse all pages at once


timoludwig:
### Motivation

For lists with tree nodes, it might be a tedious task to expand or collapse all items in the tree.

### Proposed Solution

Add a button to expand/collapse all items at once.
Use a generic implementation for all trees, so it can be used in the page tree and the language tree.

ulliholtgrave:
Is this really necessary for the Version 1.0? At the moment you can use the search for finding a page rather quickly. I would suggest to put it into Version 2.0 with a prio:high label.

PR #1250, opened 2022-02-25
Fix page selection

MizukiTemma:
### Short description

This PR changes behaviour of page selection, so the child pages are selected too when a page is checked.

### Proposed changes

- when a page is selected, expand the child pages to show the user explicitely that they are checked too.
- the check boxes of the child pages become checked too.
-

### Resolved issues

Fixes: #1052


timoludwig (commented):
Thanks a lot! :muscle: :rocket: 

In general, the approach of dynamically loading the subpages as soon as the checkbox is clicked is very good!
When the select items are clicked, this does work as expected. However, when the normal expand button is clicked (without selecting the checkbox), the recursive function somehow traverses the tree and tries to load additional items multiple times:

Screenshot 2022-03-02 at 12-54-34 Integreat Redaktionssystem

(Notice the duplicated pages "Ausländerbehörde" and "Gesundheitsamd" which should be hidden when only the first layer is expanded - but they are shown twice)

I think this is due to the fact that `listenSelectItemsRecursively()` also listens to changes that originate from the normal collapsing mechanism (and not only to those triggered by selecting a checkbox input).

Also, I'm not sure whether it's a good idea to expand all the subpages when the parent page is selected - do you think it would be possible to load them in the background, select the hidden checkboxes but leave the subpages hidden until the collapse button is explicitly clicked? Maybe this would also solve the issue above.
Probably, you could make use of [fetchSubpages()](https://github.com/digitalfabrik/integreat-cms/blob/c4e882513634e9f352a08dc9025e373f5ed00285/integreat_cms/static/src/js/pages/collapse-subpages.ts#L111) instead of triggering the `click` event on the collapse/expand button?

svenseeberg:
Also thanks! A little different than I expected, but definitely a totally valid approach.

> Also, I'm not sure whether it's a good idea to expand all the subpages when the parent page is selected 

That looks a lot like #1004. Maybe that could be added to this PR as well? I know it's a different issue. But now that it is solved with JS, its probably not complicated to have both issues solved.

timoludwig (commented):
Thanks a lot! :muscle: 

There are still quite a few cases that don't work correctly:
- When all pages are expanded/collapsed, it does not expand or collapse all, but just toggle all. This means That if the tree is already partially expanded, it just "inverts" the result which looks kind of weird.
- If there are more than 2 levels, the layout is a bit broken:
Screenshot 2022-03-06 at 23-39-44 Integreat Redaktionssystem
- The mechanism does not work for parent pages that are not on the root level:
Screenshot 2022-03-06 at 23-43-48 Integreat Redaktionssystem
- The empty page tags are not working correctly:
Screenshot 2022-03-06 at 23-43-08 Integreat Redaktionssystem

Also, I feel like with all these new functionalities, the code is becoming more and more complex and probably there are a few more problems hidden that I didn't catch.
Maybe it would make sense to implement #1158 first and have a clean differentiation between the functions "fetch subpages", "expand/collapse subpages" and "bulk select pages"? Otherwise, we have limitless nested blocks and edge cases etc...
Sorry for all the back and forth, I can also do this refactoring for you so the scope of this PR doesn't grow endlessly :sweat_smile:

timoludwig (approved):
Awesome, thanks a lot! :heart_eyes_cat: 
Really cool, this will save a lot of time for the content creators.

In general, the functionality worked as expected! I just refactored the structure a bit to make sure that all bulk action functionality is in the bulk action script, the fetching is in the fetching script etc... Since I changed quite a few things, maybe somebody could re-review this as well?

svenseeberg (approved):
Works as expected.

timoludwig:
> I'm a little confused about the prefetching feature. Instead of loading the full tree in one request the loading is now split into the view with the root nodes and then one AJAX request for each tree that loads the complete tree. While the trees are being loaded, I cannot interact with the browser. Without having timed the results, I have the feeling that waiting for the pages to load now takes longer than it took when the full tree was rendered in the view. My understanding was that we only wanted to load the next level and ignore pages with a greater depth and then only load deeper pages when required (first level is expanded). I'd expect that it would be faster to fetch all trees in one AJAX request instead of one request per root node.


Yes, what you are describing is exactly what we planned and what @MizukiTemma originally implemented. I was just not happy because this lead to a lot of weird edge cases that some of the tree functionality is enabled and some of it is disabled etc, a lot of event listeners that constantly have to be added and removed and it made everything very complicated. But maybe it just didn't put enough thought into it 😅

Issue #1086, opened 2022-01-05
Missing slug in poi view


Thomas125:
### Describe the Bug

The slug is missing in the url of a poi in the poi view

### Steps to Reproduce

1. Go to 'Locations on Map'
2. Open an arbitratry Location
3. Scroll down to 'URL parameter'

### Expected Behavior

The url is shown with slug
The link references the url with slug

### Actual Behavior

The url is shown without slug
The link references the url without slug

PR #1254, opened 2022-02-28
Deliver correct link to poi form

ulliholtgrave:
### Short description

This PR adds a correct calculation of the link to an POI and adds it to the POI form.

### Proposed changes

- Create link in view
- Use link in POI form

### Resolved issues

Fixes: #1086


integreat_cms/cms/views/pois/poi_form_view.py
```diff
@@ -70,6 +71,7 @@ def get(self, request, *args, **kwargs):
         poi_translation_form = POITranslationForm(
             instance=poi_translation, disabled=disabled
         )
+        url_link = f"{settings.WEBAPP_URL}/{region.slug}/{language.slug}/locations/"
```
timoludwig, line 74:
```suggestion
        url_link = f"{settings.WEBAPP_URL}/{region.slug}/{language.slug}/{poi_translation_form.instance.url_infix}/"
```
Instead of hardcoding the url part here, could you adapt the [url_infix](https://github.com/digitalfabrik/integreat-cms/blob/6a4be17ff0562b7c57e75c09cfce7af6064bc688/integreat_cms/cms/models/pois/poi_translation.py#L62) to `locations`?

Also, could you add the same context variable to the POST-view (in case the form is rendered with errors)...

timoludwig (commented):
Thanks! :+1:
In general, it works as expected. I just have a small idea for improvement.

timoludwig (approved):
Thanks a lot! :muscle:

PR #1260, opened 2022-03-04
Fix Firebase Notifications

svenseeberg:
### Short description
Add the legacy URL (`/REGION_SLUG/LANGUAGE_SLUG/wp-json/extensions/v3/fcm`) for the push notifications endpoint.

Also, some more data is required in the FCM JSON body. This PR includes all data that is included in https://github.com/digitalfabrik/cms/blob/develop/wp-content/plugins/firebase-notifications/classes/service.php#L116


integreat_cms/cms/views/push_notifications/push_notification_sender.py
```diff
@@ -87,17 +87,30 @@ def send_pn(self, pnt):
         :rtype: ~requests.Response
         """
         if settings.DEBUG:
-            region_slug = Region.objects.get(
-                id=settings.TEST_BLOG_ID
```
svenseeberg, line 91: Well, we could set a "TEST_BLOG_SLUG" in the settings?

integreat_cms/api/urls.py
```diff
@@ -44,6 +44,11 @@
         sent_push_notifications,
         name="sent_push_notifications",
     ),
+    path(
+        "fcm/",
```
timoludwig, line 48: Since the existing endpoint `sent_push_notifications` isn't documented anywhere, we could also just replace the existing one instead of adding a new one...

integreat_cms/cms/views/push_notifications/push_notification_sender.py
```diff
@@ -87,17 +87,30 @@ def send_pn(self, pnt):
...
-            ).slug  # Testumgebung - prevent sending PNs to actual users in development
+            region = Region.objects.get(
+                slug="testumgebung"
+            )  # Testumgebung - prevent sending PNs to actual users in development
         else:
-            region_slug = self.push_notification.region.slug
+            region = self.push_notification.region
         payload = {
-            "to": f"/topics/{region_slug}-{pnt.language.slug}-{self.push_notification.channel}",
+            "to": f"/topics/{region.slug}-{pnt.language.slug}-{self.push_notification.channel.lower()}",
```
timoludwig, line 96: Instead of `.lower()`, we could also make sure the value in the settings file is in lower case:
```
CHANNELS = (("news", "News"),) 
```
instead of:

https://github.com/digitalfabrik/integreat-cms/blob/f708ba12b2fa49163cc284ba0fd250e7f9736219/integreat_cms/core/settings.py#L702

integreat_cms/cms/views/push_notifications/push_notification_sender.py
timoludwig: sounds good.

timoludwig (approved):
Thanks a lot! :+1:
I only found optional improvements.
And maybe we should add this change to the changelog?

integreat_cms/cms/views/push_notifications/push_notification_sender.py
timoludwig: And since the tab is called "Nachrichten" on german, we could also think about translating the second "News" tuple value...

integreat_cms/cms/views/push_notifications/push_notification_sender.py
svenseeberg: Changed.

integreat_cms/api/urls.py
svenseeberg: Agreed.

integreat_cms/cms/views/push_notifications/push_notification_sender.py
svenseeberg: Done.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ UNRELEASED
 * [ [#1086](https://github.com/digitalfabrik/integreat-cms/issues/1086) ] Provide correct URL for POI
 * [ [#1247](https://github.com/digitalfabrik/integreat-cms/issues/1247) ] Update translation status on source status changes
 * [ [#1251](https://github.com/digitalfabrik/integreat-cms/issues/1251) ] Fix change detection in page form
+* [ [#1260](https://github.com/digitalfabrik/integreat-cms/pull/1260) ] Fix Firebase messaging
```
```suggestion
* [ [#1260](https://github.com/digitalfabrik/integreat-cms/issues/1260) ] Fix Firebase messaging
```
Unfortunately, our regex only handles links to `issues`... (and github redirects to the correct PR)

timoludwig (changes requested):

CHANGELOG.md
svenseeberg: Okay :D

timoludwig (approved):

Issue #1177, opened 2022-02-16
Lock for page editing


svenseeberg:
### Motivation
When a page is being edited by one users, other users should be prevented from accidentally editing it at the same time.

### Proposed Solution
Lock a page when one user is editing it. If one user is already editing, other users should be blocked from editing the form. It should be possible to take over the form from another user.

It should be sufficient to keep this "lock" status in our Redis cache, we do not have to add this to our SQL database.

PR #1261, opened 2022-03-04
Implement lock for content editing

david-venhoff:
### Short description
This pr implements support for locking a content object so it can be edited by only a single user.
When a user tries to edit some content that is already being edited by another user, a popup will show up and ask the user to go back or to take over.

### Proposed changes
- Track which objects are locked using django's caching system
- Don't allow edits for content forms if the editor does not have the exlusive lock on the content
- Send a "heartbeat" every few seconds to extend the lock and check if the page is still locked for the user
- Create an auto-save if editing permission is lost (version history is not implemented right now for pois and events though)
- Show a popup notification if the user does not have editing permission

### Resolved issues
Fixes: #1177

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ UNRELEASED
 * [ [#1319](https://github.com/digitalfabrik/integreat-cms/issues/1319) ] Fix error on Imprint API
 * [ [#1103](https://github.com/digitalfabrik/integreat-cms/issues/1103) ] Add automatic translations via DeepL API
 * [ [#1024](https://github.com/digitalfabrik/integreat-cms/issues/1024) ] Add URL search-replace for linkchecker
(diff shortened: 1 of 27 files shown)


timoludwig (commented):
Thanks a lot, really awesome feature! :heart: 
Sorry that it took me so long to review your PR, now I kind of regret it :grin: 
In general, I'm very happy with how it works from a UX-perspective and how it is implemented. I could not find any bugs, I just have a few suggestions how to show the other user in the popup to make it clear who is currently editing the form and who took over.
Thanks again! :muscle:

david-venhoff:
I added your suggestion and rebased the pr

timoludwig (approved):
Thanks again! Sorry, it seems like yet another rebase is required :see_no_evil:

PR #1264, opened 2022-03-05
Add FCM API tests

svenseeberg:
### Short description
Add sent push news to the test data and tests for the fcm API endpoint.


timoludwig (approved):
Thanks, looks good! :+1:
You could even add a push notification, which has not been sent yet, to the test data to make sure these news are not contained in the API result.

svenseeberg:
> You could even add a push notification, which has not been sent yet, to the test data to make sure these news are not contained in the API result.

Good idea. I amended the commit.

Issue #1257, opened 2022-03-02
`TypeError` when creating language tree node


timoludwig:
### Describe the Bug

Creating language tree nodes is not possible at the moment.

### Steps to Reproduce

1. Go to language tree
2. Create new language tree node
3. See error

### Expected Behavior

The new tree node should be created

### Actual Behavior

`TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'`

### Additional Information

<details>
  <summary>Traceback</summary>
    
  ```
Mar 02 15:44:40 ERROR   django.request - Internal Server Error: /stralsund/language-tree/new/
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/views/language_tree/language_tree_node_form_view.py", line 101, in post
    elif not language_tree_node_form.has_changed():
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/forms/forms.py", line 425, in has_changed
[...]

PR #1265, opened 2022-03-05
Fix creating language tree nodes

timoludwig:
### Short description

The fix in bbe3e55e2a57179ce6c31de5efa44ecdb7b5e68a did not work correctly, so a few more substantial changes were required.

### Proposed changes

- Add test for creating language tree nodes
- Fix creating language tree nodes
    - coerce ref_node field to int was not possible because the reference to no node cannot be expressed by an integer
    - Use strings for values of ref_node fields where possible
    - Fix parent field after `move()` and not before `save()` because it caused problems for language tree nodes
- Provide fallback translation if default translation doesn't exist
    - Under normal circumstances, this should not happen. But when the default language is changed into a language that does not have a translation for each page, this might lead to problems.
- Show content list views only for active languages
- Do not return skipped nodes in `cache_tree()` method

### Resolved issues

Fixes: #1257

### DIFF
--- a/integreat_cms/api/v3/events.py
+++ b/integreat_cms/api/v3/events.py
@@ -49,7 +49,7 @@ def transform_event_translation(event_translation):
     if event.location:
         location_translation = (
             event.location.get_public_translation(event_translation.language.slug)
-            or event.location.default_translation
+            or event.location.best_translation
         )
     else:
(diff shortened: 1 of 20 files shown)


svenseeberg (approved):
LGTM

PR #1267, opened 2022-03-06
Fix push notifications attribute name in API

svenseeberg:
### Short description
The attribute that indicates activated push notifications in the regions/sites endpoint should be `push_notifications`.

### Proposed changes
Change the attribute name `push-notifications` to `push_notifications` in the regions API endpoint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 UNRELEASED
 ----------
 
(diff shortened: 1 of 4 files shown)


timoludwig (commented):
Thanks! Maybe add at least one entry to the changelog so we don't have a release without any release notes?

timoludwig (approved):
:+1:

Issue #1271, opened 2022-03-06
Feedback endpoint not working if multiple translations exist for same slug


timoludwig:
### Describe the Bug

When a translation has multiple versions with the same slug, the feedback endpoint throws an error:

### Steps to Reproduce

1. Create a second translation for `willkommen` page with same slug
2. execute `curl -X "POST" "http://localhost:8000/api/augsburg/de/feedback/page" -d 'slug=willkommen&rating=up'`

### Expected Behavior

`{"success": "Feedback successfully submitted"}`

### Actual Behavior

Internal server error

### Additional Information

<details>
  <summary>Traceback</summary>
    
  ```
Mar 06 20:43:18 ERROR django.request - Internal Server Error: /api/augsburg/de/feedback/page
Traceback (most recent call last):
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.9/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/home/timo/job/integreat/integreat-cms/integreat_cms/api/decorators.py", line 87, in handle_feedback
    return func(data, region, language, comment, rating_normalized, is_technical)
  File "/home/timo/job/integreat/integreat-cms/integreat_cms/api/decorators.py", line 124, in wrap
    return function(request, *args, **kwargs)
  File "/home/timo/job/integreat/integreat-cms/integreat_cms/api/v3/feedback/page_feedback.py", line 38, in page_feedback
    return page_feedback_internal(data, region, language, comment, rating, is_technical)
  File "/home/timo/job/integreat/integreat-cms/integreat_cms/api/v3/feedback/page_feedback.py", line 67, in page_feedback_internal
    page_translation = get_object_or_404(
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.9/site-packages/django/shortcuts.py", line 76, in get_object_or_404
    return queryset.get(*args, **kwargs)
[...]

PR #1272, opened 2022-03-06
Fix feedback endpoint

timoludwig:
### Short description

This fixes the bug in the feedback endpoint

### Proposed changes

- Add second version for content translations to test data (this reproduces the bug)
- Fix `MultipleObjectsReturned` in feedback endpoint

### Resolved issues

Fixes: #1271

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 UNRELEASED
 ----------
 
+* [ [#1271](https://github.com/digitalfabrik/integreat-cms/issues/1271) ] Fix feedback API endpoint
+
 
 2022.3.2
 --------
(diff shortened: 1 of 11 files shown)


svenseeberg (approved):
Thanks! Looks good to me. I'm only wondering if the raised error with the message `"This event translation slug is not unique, please contact your server administrator."` is helping anyone ^^

timoludwig:
> Thanks! Looks good to me. I'm only wondering if the raised error with the message `"This event translation slug is not unique, please contact your server administrator."` is helping anyone ^^

Your're right, and since the new exception is not caught, it isn't shown to the end user anyway (the api call just returns an internal server error, which is pretty much the correct response I think^^). But when we only see the message in the log, it doesn't make sense to address the server administrator here... I'll change the wording.

PR #1275, opened 2022-03-08
Add timezone setting to region model

JoeyStk:
### Short description
So far the events haven't been timezone aware. This PR changes the model to be timezone aware, following the suggestion of Timo inside the PR. 

### Proposed changes
The PR adds a charfield. The default value of the charfield is the timezone that was previously defined inside the setting.py

### Resolved issues
Fixes: #1365

Please feel free to give feedback on whether or not I understood the issue correctly. Thank you a lot in advance! :)


svenseeberg:
I think the issue has a larger scope than simply converting the timestamp. We probably need a timezone setting for regions and all events created for this region use this timezone. Alternatively, the time zone could be a part of the user profile. But putting it in a region is probably better.

All dates entered for an event should then be based on the region time zone.

timoludwig:
> But putting it in a region is probably better.

Yes, I agree.

Maybe we could even increase the scope of this to all region content, not only the events. We could combine the existing [region middleware](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/core/middleware/region_middleware.py) with the [timezone middleware](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/middleware/timezone_middleware.py) and activate the region's timezone on all views that have `request.region` set. This would mean that all `last_updated` timestamps etc. also reflect the current region's timezone, and not the system timezone anymore. The system timezone would then only be used for the network admin area.

JoeyStk:
Please find the updated commit to the PR. This is the first commit, which so far only adds the timezone field to the region model. I would be interested to know if I managed to find a proper solution for the issue. 
Also I still have a question left. I'm not very happy with the content of the dropdown so far. I don't find this to be very user-friendly for two reasons:
- there is still content like "etc/gmt+11", which I'm conflicted about whether it's necessary in this context or not
- the options inside the dropdown are to me quite unsorted, which I found quite unintuitive. 
I would be interested to know how you think about this :)

integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -61,6 +61,7 @@ class Meta:
             "offers",
             "short_urls_enabled",
             "custom_prefix",
+            "custom_timezone",
```
timoludwig, line 64:
```suggestion
            "timezone",
```

If you want to make the selection really effortless, you could add the field `timezone_region = forms.ChoiceField(choices=get_timezone_region_choices())` to the region form - this additional field does not belong to the region model and ist just there to simplify the selection by selecting the global area first and then have a second dropdown for the actual timezone.
For the choices, I suggest the following snippet:
```
def get_timezone_region_choices():
    timezone_regions = list(
        {tz.split("/")[0] for tz in available_timezones() if "/" in tz}
    )
    timezone_regions.sort()
    return [("", "---------")] + [(tzr, tzr) for tzr in timezone_regions]
```
Then, the next step would be to implement a conditional selection in Javascript - each time the global field changes, we should hide/show some of the options in the timezone dropdown to make sure only the options of the currently selected global area are visible.

integreat_cms/cms/templates/regions/region_form.html
```diff
@@ -99,6 +99,12 @@ <h3 class="heading font-bold text-black"><i data-feather="map" class="mr-2"></i>
                 <h3 class="heading font-bold text-black"><i data-feather="settings" class="mr-2"></i> {% trans 'Extended Settings' %}</h3>
             </div>
             <div class="px-4 pb-4 divide-y divide-gray-200 space-y-2">
+                <div>
+                    <!-- Options related to Events -->
+                    <label for="{{ form.custom_timezone.id_for_label }}">{{ form.custom_timezone.label }}</label>
+                    {% render_field form.custom_timezone %}
+                    <div class="help-text">{{ form.custom_timezone.help_text }}</div>
```
timoludwig, line 106:
```suggestion
                    <!-- Options related to Timezones -->
                    <label for="{{ form.timezone.id_for_label }}">{{ form.timezone.label }}</label>
                    {% render_field form.timezone %}
                    <div class="help-text">{{ form.timezone.help_text }}</div>
```

timoludwig (commented):
Thanks a lot! :muscle: 

> I would be interested to know if I managed to find a proper solution for the issue.

Yes, this is going in the right direction. :+1:

>  There is still content like "etc/gmt+11", which I'm conflicted about whether it's necessary in this context or not

Probably not necessary, but at the moment I think it's easier to just let the python library handle all choices, so we don't have to care about it.

> The options inside the dropdown are to me quite unsorted, which I found quite unintuitive.

Good point, I agree. Sorting is quite easy, but we could even further improve it by having two dropdowns: At first, we select the global area (the part before the first slash) and then, in the second dropdown we list only those entries of that area. This should prevent a lot of unnecessary scrolling etc. Downside is we need a little bit of javascript to implement this logic.

timoludwig (commented):
Thanks a lot! :muscle: 
I noticed that the output `available_timezones()` somehow depends on the system's locale files or similar, so we cannot rely on the fact that they are deterministic.
See my comments below how I suggest to get around this problem.

timoludwig (commented):
Thanks! Now, could you have a look at all circleci errors and try to fix them?

integreat_cms/cms/templates/regions/region_form.html
```diff
@@ -99,6 +99,18 @@ <h3 class="heading font-bold text-black"><i data-feather="map" class="mr-2"></i>
...
             </div>
             <div class="px-4 pb-4 divide-y divide-gray-200 space-y-2">
+                <div>
+                    <!-- Options related to Timezones -->
+                    <label for="{{ form.timezone_area.id_for_label }}">{{ form.timezone_area.label }}</label>
+                    {% render_field form.timezone_area %}
+                    <div class="help-text">{{ form.timezone_area.help_text }}</div>
+                </div>
+                <div>
+                    <!-- Options related to Timezones -->
```
timoludwig, line 109:
```suggestion
```
Since both of these dropdowns have the same topic, I think we don't need a separator between them.

integreat_cms/static/src/js/regions/conditional-fields.ts
```diff
@@ -28,4 +28,33 @@ window.addEventListener('load', () => {
...
+        otherOptions = timezoneDropdown.querySelectorAll('option');
+    } else {
+        otherOptions = timezoneDropdown.querySelectorAll('option:not([value*="' + timezoneAreaDropdown.value + '"])');
+        selectedOptions = timezoneDropdown.querySelectorAll('option[value*="' + timezoneAreaDropdown.value + '"]');
+    }
+    otherOptions.forEach((option) => option.classList.add('hidden'));
+    selectedOptions.forEach((option) => option.classList.remove('hidden'));
+
+    timezoneDropdown.querySelector('option[value=""]').classList.remove('hidden');
+}
```
timoludwig, line 60:
```suggestion
function updateTimezoneAreaDropdown(event?: Event) {
    const timezoneAreaDropdown = document.getElementById("id_timezone_area") as HTMLSelectElement;
    const timezoneDropdown = document.getElementById("id_timezone") as HTMLSelectElement;
    // Reset timezone value when timezone region is changed
    if (event !== undefined) {
        timezoneDropdown.value = "";
    }
    // Hide all sub-timzones of other areas
    timezoneDropdown
        .querySelectorAll('option:not([value*="' + timezoneAreaDropdown.value + '"],[value=""])')
        .forEach((option) => option.classList.add("hidden"));
    // Show all sub-timezones of the selected area
    timezoneDropdown
        .querySelectorAll('option[value*="' + timezoneAreaDropdown.value + '"]')
        .forEach((option) => option.classList.remove("hidden"));
}

```
I found a way to simplify this function - it turns out we don't need the if clause...

timoludwig (commented):
Thanks a lot, this works very well! :+1:
Just two small suggestions - and could you add a small note to the changelog? Thanks!

timoludwig (approved):
Amazing, thanks! :rocket:

Issue #1277, opened 2022-03-08
Changed XLIFFs are incorrectly detected as unchanged


timoludwig:
### Describe the Bug

At the moment, importing XLIFFs does not work when a translation for this page already exists in the database.
The new import is incorrectly detected as "unchanged".

### Steps to Reproduce

1. Export XLIFF file for existing translation
2. Change title or content in file
3. Import XLIFF again

### Expected Behavior

The changed diff should be shown

### Actual Behavior

The content is correctly displayed, but it states "No changes detected" as if the new content would already be in the database

Issue #1276, opened 2022-03-08
Allow unchanged XLIFF files to be uploaded


timoludwig:
### Motivation

When uploading XLIFF files, it currently causes an error when the files does not change the existing content.
However, it is quite common that translation were already up to date when exporting, so no changes are a perfectly valid result.

### Proposed Solution

Instead of showing the error, do the following:
- Remove the `currently_in_translation` flag of the existing translation
- Update the `last_updated` date of the existing translation
- Show an info message instead of the error

PR #1278, opened 2022-03-08
Fix XLIFF import

timoludwig:
### Short description

This fixes the XLIFF import

### Proposed changes

- Add tests for xliff import/export
- Fix change detection for XLIFF import
- Allow importing unchanged XLIFF files

### Resolved issues

Fixes: #1276
Fixes: #1277

### DIFF
--- a/integreat_cms/xliff/utils.py
+++ b/integreat_cms/xliff/utils.py
@@ -8,8 +8,6 @@ import glob
 import os
 import uuid
 
-from copy import deepcopy
-
 from django.conf import settings
 from django.contrib import messages
 from django.core import serializers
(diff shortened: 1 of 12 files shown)


integreat_cms/xliff/utils.py
```diff
@@ -375,6 +398,9 @@ def get_xliff_import_errors(request, page_translation):
     :param page_translation: The page translation which is being imported
     :type page_translation: ~integreat_cms.cms.models.pages.page_translation.PageTranslation
 
+    :param add_message_if_unchanged: Whether a message should added if  no changes are detected
```
svenseeberg, line 401:
```suggestion
    :param add_message_if_unchanged: Whether a message should added if no changes are detected
```

tests/xliff/xliff_test.py
```diff
@@ -0,0 +1,183 @@
...
+        # If the role should be allowed to access the view, we expect a successful result
+        assert response.status_code == 302
+        page_tree = reverse(
+            "pages", kwargs={"region_slug": "augsburg", "language_slug": "en"}
+        )
+        assert response.headers.get("Location") == page_tree
+        response = client.get(page_tree)
+        print(response.headers)
+        # The download link to the xliff file is contained in the message
+        match = re.search(
```
svenseeberg, line 47: I'd prefer a real XML parser in such cases, but I do not object a regex for this limited scope.

svenseeberg (approved):
As always just some comments, no requests for changes. Awesome tests :rocket:

tests/xliff/xliff_test.py
timoludwig: Good idea, I replaced the regex with:
```
parsed_content = fromstring(response.content.decode("utf-8"))
download_links = [
    link.attrib["href"]
    for link in parsed_content.iter("a")
    if "data-auto-download" in link.attrib
]
```

PR #1285, opened 2022-03-09
Do not send a push notification twice, fixes #1281

### LINKED ISSUE #1281: Push notifications can be sent multiple times
### Describe the Bug

It should not be possible to send push notifications multiple times to prevent the accidental re-sending of news.
Instead, only the "save" button should be enabled which can be used to update the messages that are delivered via the API endpoint.

### Steps to Reproduce

1. Go to News
2. Send new push notification
3. Send again

### Expected Behavior

The news should only be sent the first time and only saved on all subsequent submissions

### Actual Behavior

The news is sent twice

svenseeberg:
### Short description
Prevent sending push notifications (again) that have a `sent_date`.

### Proposed changes
- Hide the Send button if a sent_date exists
- When submitting the form do not send the news if a sent_date exists

### Resolved issues
Fixes: #1281


timoludwig (approved):

PR #1286, opened 2022-03-09
Improve push notifications handling

### LINKED ISSUE #760: Add attributes to push notification data for native apps
### Motivation

In order to be able to directly open the correct local news in the native apps upon clicking on a push notification, the native apps need some additional attributes to map the push notification to the correct local news.

### Proposed Solution

Include the attributes `city_code` and `news_id` to the push notification data similar to the `language_code` attribute that is already present. The `city_code` should be equivalent to the path attribute from the [sites endpoint](https://cms.integreat-app.de/wp-json/extensions/v3/sites) (if possible without slashes), e.g. `augsburg` for the city of Augsburg. The `news_id` should be equivalent to the `id` from the [fcm endpoint](https://cms.integreat-app.de/testumgebung/de/wp-json/extensions/v3/fcm?channel=news).

### Additional Context

Please leave us a comment in [this issue](https://issues.tuerantuer.org/browse/IGAPP-263) once this is done.

timoludwig:
### Short description

- Improve the feedback a user gets when something goes wrong with push notifications
- Enable linking of push notifications to local news in native apps

Fixes #760

### DIFF
--- a/integreat_cms/cms/views/push_notifications/push_notification_sender.py
+++ b/integreat_cms/cms/views/push_notifications/push_notification_sender.py
@@ -5,6 +5,7 @@ import logging
 import requests
 
 from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
 
 from ...models import PushNotificationTranslation
 from ...models import Region
@@ -13,7 +14,6 @@ from ...constants import push_notifications as pnt_const
 logger = logging.getLogger(__name__)
 
 
-# pylint: disable=too-few-public-methods
 class PushNotificationSender:
     """
     Sends push notifications via FCM HTTP API.
@@ -28,6 +28,9 @@ class PushNotificationSender:
 
         :param push_notification: the push notification that should be sent
         :type push_notification: ~integreat_cms.cms.models.push_notifications.push_notification.PushNotification
+
+        :raises ~django.core.exceptions.ImproperlyConfigured: If the auth key is missing or the system runs in debug
+                                                              mode but the test region does not exist.
         """
         self.push_notification = push_notification
         self.prepared_pnts = []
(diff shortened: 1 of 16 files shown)


integreat_cms/cms/views/push_notifications/push_notification_sender.py
```diff
@@ -86,19 +102,14 @@ def send_pn(self, pnt):
...
-                slug=settings.TEST_REGION_SLUG
-            )  # Testumgebung - prevent sending PNs to actual users in development
-        else:
-            region = self.push_notification.region
         payload = {
-            "to": f"/topics/{region.slug}-{pnt.language.slug}-{self.push_notification.channel}",
+            "to": f"/topics/{self.region.slug}-{pnt.language.slug}-{self.push_notification.channel}",
             "notification": {"title": pnt.title, "body": pnt.text},
             "data": {
+                "news_id": str(pnt.id),
```
timoludwig, line 109: @klinzo Is this what you had in mind with https://github.com/digitalfabrik/integreat-cms/issues/760?

Also, can we use this opportunity to get some consistency into the data? We somehow mix `camelCase` and `snake_case` here and use `blog_id` and `city` for regions... Or are these fields needed exactly like this for backwards compatibility?

klinzo: Yes, that is definitely what I had in mind with #760, is this the same id that will later be used as `id` in this endpoint: https://github.com/digitalfabrik/cms/wiki/REST-APIv3-Documentation#get-messages-for-channel

Feel free to adjust the attributes as needed, we don't use those yet since they are not available in wordpress afaik, right? So you can definitely remove/change `blog_id` and change `lanCode` to `language_code`. If you want you can also change `city` to `city_code`.

timoludwig: Alright, thanks! :+1:

> at least if this is the same id that will later be used as `id` in this endpoint

Yes, it is :ok_hand:

integreat_cms/core/settings.py
```diff
@@ -80,6 +80,10 @@
 #: Authentication token for the Firebase API. This needs to be set for a correct usage of the messages feature.
 FCM_KEY = os.environ.get("INTEGREAT_CMS_FCM_KEY")
 
+#: Whether push notifications via Firebase are enabled.
+#: This is ``True`` if :attr:`~integreat_cms.core.settings.FCM_KEY` is set, ``False`` otherwise.
+FCM_ENABLED = bool(FCM_KEY)
```
ulliholtgrave, line 85: I would prefer to provide some dummy FCM key, if DEBUG is activated. Otherwise, we always have to set some env variable to see the whole Push Notification area.

ulliholtgrave:
Another small feature request: Could we add a status field to our PN form? I guess it's kinda helpful to identify the the PN as a user.

ulliholtgrave (approved):
Thanks. This fixes the linking issue and improves the UI of the notifications a lot! :)

Issue #1282, opened 2022-03-09
Cache not invalidated when feedback changes


timoludwig:
### Describe the Bug

When feedback objects change (e.g. when new feedback is submitted, or the read-status of feedback entries is changed), the change is not immediately visible in the feedback view.

### Steps to Reproduce

1. Enable redis cache
2. Go to feedback
3. Mark entry as read

### Expected Behavior

The entry should be marked as read

### Actual Behavior

The entry is still unread

PR #1287, opened 2022-03-10
Fix feedback cache invalidation

melegiul:
### Short description

Fix feedback cache invalidation

### Proposed changes

django-cacheops does not support multi-table inheritance (https://github.com/Suor/django-cacheops/issues/31)
The suggested workaround is to use signal handlers to invalidate the cache manually (https://github.com/Suor/django-cacheops/issues/31#issuecomment-673564324), so signal handlers are used to invalidate the cache on feedback creation and deletion.
But marking feedback as read/unread uses the query set update method and does not trigger a save signal, so in case of feedback updates cache invalidation is performed in the view instead.

### Resolved issues

Fixes: #1282

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ UNRELEASED
 
 * [ [#1301](https://github.com/digitalfabrik/integreat-cms/issues/1301) ] Fix order of push notifications
 * [ [#1296](https://github.com/digitalfabrik/integreat-cms/issues/1296) ] Fix page tree after resetting filters
+* [ [#1282](https://github.com/digitalfabrik/integreat-cms/issues/1282) ] Fix feedback cache invalidation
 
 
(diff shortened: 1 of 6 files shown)


integreat_cms/cms/signals/FeedbackUpdateHandler.py
```diff
@@ -0,0 +1,60 @@
+from django.db.models.signals import post_delete, post_save
```
timoludwig, line 1: Python usually uses `snake_case` for file naming, so I suggest to call this file `feedback_signals.py` or similar.

melegiul: Thank you for your review, I have renamed now the file as you suggested :+1:

timoludwig (approved):
Thanks! :rocket:

Issue #1256, opened 2022-03-02
Remove 'Media Replacement' button from media selection popup


ulliholtgrave:
### Motivation

The functionality to replace media elements shouldn't be present in the media selection for content pages as it doesn't serve a purpose in this context.

### Proposed Solution

Add some check to the preact component to exclude the 'media replacement' button.

PR #1295, opened 2022-03-14
Remove replace media from media selection

ulliholtgrave:
### Short description

This PR removes the replace button from the media selection.

### Proposed changes

- Only add replace button if media component isn't in selection mode
- Fix some code smell issues

### Resolved issues

Fixes:  #1256


timoludwig (approved):
Thanks a lot! :+1:
Didn't review the whole diff, but tested the feature and it works as expected.

PR #1300, opened 2022-03-15
Publish dev package on branches matching pattern

timoludwig:
### Short description

In some cases, we want to build specific branches as dev packages before merging the PR into develop. For this use case, we can now just append `-publish-dev-package` to the branch name and push the changes, which triggers the dev release workflow.
Since all push events trigger the CircleCI build, It is not required to open a separate PR for the new branch.
After the dev package has been published, it can be installed on the test system to check how the changes affect a production-like environment.

Since the resulting packages might not be in the correct order, this needs a bit of care when database migrations are involved. In this case, make sure to only use migrations that can be reversed and reverse them before you install another package version which might not contain the migration file of the feature branch.

### Proposed changes

- Publish dev packages on every branch ending with `-publish-dev-package` 
- Update documentation

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -453,7 +453,9 @@ workflows:
       - bump-dev-version:
           filters:
             branches:
-              only: develop
+              only:
+                - develop
+                - /.*-publish-dev-package/
           requires:
             - pipenv-install
       - build-package:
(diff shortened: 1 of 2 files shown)


ulliholtgrave (approved):
LGTM

Issue #1301, opened 2022-03-16
Push notifications ordered incorrectly


timoludwig:
### Describe the Bug

In both the push notification list and the API, the results are ordered incorrectly.

### Steps to Reproduce

1. Go to push notifications
2. See order

### Expected Behavior

- In the push notification list, the result is ordered descending by creation date
- In the push notification API, the result is ordered by the push notification id

### Actual Behavior

- In the push notification list, the result should be ordered by the last updated date of the translation, and ascending instead of descending
- In the push notification API, the result should be ordered by the timestamp (which should be the last updated date instead of the sent date)

PR #1302, opened 2022-03-16
Bugfix/push notification order

timoludwig:
### Short description

This fixes the order of push notifications

### Proposed changes

- Fix order of push notifications
- Fix push notification API response format
    - Use localized last updated as timestamp instead of sent date
    - Add channel to API result
- Set news as default push notification channel
- Translate push notification channel (Show "Nachrichten" instead of "News")

### Resolved issues

Fixes: #1301


svenseeberg (approved):
LGTM

PR #1303, opened 2022-03-16
Improve translation of disabled creation button

timoludwig:
### Short description

Improve translation of disabled creation button

### Proposed changes

- Improve translation of disabled creation button


ulliholtgrave (approved):
Thanks, Works as expected :)

Issue #1296, opened 2022-03-15
Page tree broken after search is cleared


timoludwig:
### Describe the Bug

When the search form in the page tree is cleared, all pages are expanded by default, but the collapsing/expanding does not work as expected.

Maybe the source of the bug is the same as #1179 and it's not a browser caching problem after all? If so, maybe fixing this also fixes #1245 when we enable the browser cache again after the back-button.

### Steps to Reproduce

1. Go to the page tree
2. Enter something into the seach bar and submit
3. Clear the search via the "x"
4. See broken page tree

### Expected Behavior

The page tree should be fully collapsed and expanding should work normal

### Actual Behavior

Everything is expanded and the tree becomes weird on toggling subpages

### Additional Information

E.g. the following subtree:

![Screenshot 2022-03-15 at 16-46-48 Integreat Redaktionssystem](https://user-images.githubusercontent.com/16021269/158416817-ab6c7c4a-ba81-4e11-be24-68cff46930c1.png)

Toggling the subpages only hides the middle level and the leaf pages are still visible

![Screenshot 2022-03-15 at 16-46-58 Integreat Redaktionssystem](https://user-images.githubusercontent.com/16021269/158416811-f7467cdb-c09b-45d8-9d34-225e00864c5a.png)

PR #1304, opened 2022-03-16
Fix page filter form

timoludwig:
### Short description

Fix page tree after resetting filters

### Proposed changes

- Refactor page filter form
- Add page filter tests

### Resolved issues

Fixes: #1296


ulliholtgrave (approved):
Thanks, Works as expected! :)

Issue #1279, opened 2022-03-09
News Form Attribute Error


membralala:
### Describe the Bug

When saving the news form without entering title and content before, it throws an Attribute Error.

### Steps to Reproduce
Try to save a news entry without title and content.

### Expected Behavior

A warning message should appear as it is implemented in other forms when critical data is missing.

### Actual Behavior

The push notification is still saved to the database, however, the form could not be rendered afterwards, if there is no title specified.

timoludwig:
I cannot reproduce this error. When I save the form without a title, the title is empty, but it does not show an error.
Push news are a bit different to other models because an empty title is actually allowed in the database.
Can you post the traceback for the AttributeError?

ulliholtgrave:
I am also not able to reproduce this. The only issue I see is the maybe arbitrary error message that says that text is missing, but the real issue is the missing FCM key.

membralala:
@timoludwig sorry for late reply, I missed your comment somehow. 

I tried it again. Providing some input text for either title or content seems to work fine, but saving without any of those gives me the following: 
``` 
Traceback (most recent call last):
  File "/Users/philip/.local/share/virtualenvs/Integreat-4loMQVvc/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/Users/philip/.local/share/virtualenvs/Integreat-4loMQVvc/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/Users/philip/.local/share/virtualenvs/Integreat-4loMQVvc/lib/python3.9/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/Users/philip/.local/share/virtualenvs/Integreat-4loMQVvc/lib/python3.9/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/Users/philip/.local/share/virtualenvs/Integreat-4loMQVvc/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/Users/philip/.local/share/virtualenvs/Integreat-4loMQVvc/lib/python3.9/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
[...]

PR #1307, opened 2022-03-19
Fix AttributeError in news form

timoludwig:
### Short description

Fix AttributeError in news form

### Proposed changes

- Make title field of push notification translations mandatory
- Indicate which language of the formset is responsible for an error
- Require at least the form of the default language to contain data
- Add red borders around form fields that are not valid

I also updated the pip dependencies because I got a [weird CircleCI error](https://app.circleci.com/pipelines/github/digitalfabrik/integreat-cms/4017/workflows/ea6b2832-b502-4a8e-aaff-05b21e355487) and I assume an outdated cache is responsible.

### Resolved issues

Fixes: #1279


ulliholtgrave (approved):

Issue #1024, opened 2021-11-22
Centralized Link / Content Replacement


svenseeberg:
The Broken Link Checker in WP has a function to replace links in the content. This a often used function and is needed before we migrate the Integreat CMS.

We need an easy to use function close to the broken links feature, that lets users replace old links in the content. Basically this is a search and replace for all pages.

PR #1310, opened 2022-03-20
Implement URL search-replace for linkchecker

membralala:
### Short description

This PR provides a basic search and replace solution for URLs. It therefore enhances the linkchecker's list view

### Proposed changes

- Add an URL edit form to the linkchecker's list view
- Replace the URL in pages contents 

### Resolved issues

Fixes: #1024

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ UNRELEASED
 
 * [ [#1319](https://github.com/digitalfabrik/integreat-cms/issues/1319) ] Fix error on Imprint API
 * [ [#1103](https://github.com/digitalfabrik/integreat-cms/issues/1103) ] Add automatic translations via DeepL API
+* [ [#1024](https://github.com/digitalfabrik/integreat-cms/issues/1024) ] Add URL search-replace for linkchecker
 
 
 2022.3.6
--- a/integreat_cms/cms/urls/protected.py
+++ b/integreat_cms/cms/urls/protected.py
@@ -597,8 +597,20 @@ urlpatterns = [
                                         ),
                                         path(
                                             "<slug:link_filter>/",
-                                            linkcheck.LinkListView.as_view(),
-                                            name="linkcheck",
+                                            include(
+                                                [
+                                                    path(
+                                                        "",
+                                                        linkcheck.LinkListView.as_view(),
+                                                        name="linkcheck",
+                                                    ),
+                                                    path(
+                                                        "<int:link_id>/",
+                                                        linkcheck.LinkListView.as_view(),
+                                                        name="edit_link",
+                                                    ),
+                                                ]
+                                            ),
                                         ),
                                     ]
                                 ),
--- a/integreat_cms/cms/views/linkcheck/linkcheck.py
+++ b/integreat_cms/cms/views/linkcheck/linkcheck.py
@@ -1,7 +1,9 @@
+import re
+
 from django.conf import settings
 from django.contrib import messages
 from django.urls import reverse
-from django.shortcuts import redirect
+from django.shortcuts import get_object_or_404, redirect
 from django.utils.translation import gettext as _
 from django.views.generic import ListView
 from django.views.generic.base import RedirectView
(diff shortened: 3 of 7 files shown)


membralala:
With the way this is working now, I noticed an issue, that the replacement fails due to the form validation, if the URL, that you want to use is already in the database and I'm not quite sure, how to tackle this one. Could we even ignore the validation as there is already some browser-side validation and the actual model is never updated anyway?

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -4301,10 +4302,14 @@ msgstr "Anzahl an Regionen"
...
 msgstr "Noch keine Sprachen vorhanden."
 
-#: cms/templates/linkcheck/link_list_row.html:24
+#: cms/templates/linkcheck/link_list_row.html:25
 msgid "Go to source"
 msgstr "Inhalt aufrufen"
 
+#: cms/templates/linkcheck/link_list_row.html:35
+msgid "Edit URL"
+msgstr "URL bearbeiten"
```
timoludwig, line 4311:
```suggestion
msgid "Replace URL globally"
msgstr "URL zentral ersetzen"
```

integreat_cms/cms/views/linkcheck/linkcheck.py
```diff
@@ -6,9 +8,10 @@
 from django.views.generic import ListView
 from django.views.generic.base import RedirectView
 
-from linkcheck.models import Link
+from linkcheck.models import Link, Url
 
 from ...utils.filter_links import filter_links
+from ...forms.linkchecker.edit_url_form import EditUrlForm
```
timoludwig, line 14:
```suggestion
from ...forms.linkcheck.edit_url_form import EditUrlForm
```
I'd rename the directory from `linkchecker` to `linkcheck` to be more consistent with the rest of our linkcheck files.

integreat_cms/cms/views/linkcheck/linkcheck.py
```diff
@@ -1,3 +1,5 @@
+import re
+
 from django.conf import settings
 from django.contrib import messages
 from django.urls import reverse
```
timoludwig, line 5:
```suggestion
from django.urls import reverse
from django.shortcuts import get_object_or_404, redirect
```

timoludwig (commented):
Thanks a lot! :muscle: 
In general, it works as expected, but I noticed a few problems: It replaces the content of all versions, which means a lot of overhead for replaced links that are never actually used or even viewed in the link list. It turns out my initial idea didn't work as I thought, so I came up with a different idea to filter only the latest versions of the translations, see my comments below for details.

> With the way this is working now, I noticed an issue, that the replacement fails due to the form validation, if the URL, that you want to use is already in the database and I'm not quite sure, how to tackle this one. Could we even ignore the validation as there is already some browser-side validation and the actual model is never updated anyway?

I suggest to circumvent this problem by using a basic `forms.Form` instead of `forms.ModelForm`, but see my comments below for details.

integreat_cms/cms/urls/protected.py
```diff
@@ -590,8 +590,32 @@
...
+                                                                    linkcheck.LinkListView.as_view(),
+                                                                    name="edit_link",
+                                                                ),
+                                                                path(
+                                                                    "edit/",
+                                                                    linkcheck.LinkListView.as_view(),
+                                                                    name="edit_url",
+                                                                ),
+                                                            ]
+                                                        ),
```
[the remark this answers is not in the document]
membralala: Indeed!

membralala:
Thanks for your feedback! I applied your suggestions and rebased the branch.

membralala:
Thanks for your feedback! I applied your suggestions and rebased the branch.

timoludwig:
Oh, and just one additional request: could you add a note to the changelog file? Thanks!

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 UNRELEASED
 ----------
-
+* [ [#1024](https://github.com/digitalfabrik/integreat-cms/issues/) ] Add URL search-replace for linkchecker
```
timoludwig, line 3: Could you add this to the last line? :see_no_evil:

Issue #1290, opened 2022-03-11
Delete multiple feedback with different types


melegiul:
### Describe the Bug

Delete multiple feedback with differnt types does not work

### Steps to Reproduce

1. Go to the region feedback view
2. Select at least 2 feedback entries with different types
3. Click on bulk delete

### Expected Behavior

Selected feedback should be removed

### Actual Behavior

Selected feedback persists also after cache clearing, persists on database 

### Additional Information

No traceback or error message
Deleting multiple feedback with same type works fine

timoludwig:
Thanks for the report! I can reproduce this issue with the following traceback:

<details>
<summary>Traceback</summary>

```
Environment:


Request Method: POST
Request URL: http://localhost:8000/augsburg/feedback/delete/

Django Version: 3.2.12
Python Version: 3.9.10
Installed Applications:
['integreat_cms.cms',
 'integreat_cms.gvz_api',
 'integreat_cms.linkcheck.apps.ModifiedLinkcheckConfig',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.messages',
 'django.contrib.sessions',
 'django.contrib.sitemaps',
 'django.contrib.staticfiles',
 'corsheaders',
 'rules.apps.AutodiscoverRulesConfig',
 'treebeard',
 'webpack_loader',
 'widget_tweaks',
 'polymorphic',
 'django.contrib.admin',
 'debug_toolbar']
Installed Middleware:
['debug_toolbar.middleware.DebugToolbarMiddleware',
 'corsheaders.middleware.CorsMiddleware',
 'django.middleware.security.SecurityMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.locale.LocaleMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'integreat_cms.core.middleware.RegionMiddleware',
 'integreat_cms.core.middleware.AccessControlMiddleware',
 'integreat_cms.cms.middleware.TimezoneMiddleware',
 'integreat_cms.api.middleware.JsonDebugToolbarMiddleware']



Traceback (most recent call last):
[...]

PR #1312, opened 2022-03-21
Fix deleting polymorphic feedback

david-venhoff:
### Short description

This pr implements a fix for deleting multiple feedback items of different types, which does not work because of a bug in django-polymorphic ([229](https://github.com/django-polymorphic/django-polymorphic/issues/229)).

### Proposed changes

The fix implements the recommended workaround for this bug, which is to only retrieve the base class before deletion.

### Resolved issues

Fixes: #1290

### DIFF
--- a/integreat_cms/cms/models/feedback/feedback.py
+++ b/integreat_cms/cms/models/feedback/feedback.py
@@ -5,6 +5,7 @@ from django.utils.functional import cached_property
 from django.utils.translation import ugettext_lazy as _
 
 from polymorphic.models import PolymorphicModel
+from polymorphic.managers import PolymorphicQuerySet, PolymorphicManager
 
 from ...constants import feedback_ratings
 from ...utils.translation_utils import ugettext_many_lazy as __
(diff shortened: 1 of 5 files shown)


david-venhoff:
I am a bit confused that the documentation fails to build. What reference `polymorphic.query.PolymorphicQuerySet` is the error message about?

timoludwig (commented):
Thanks, looks good! :rocket:

> I am a bit confused that the documentation fails to build. What reference `polymorphic.query.PolymorphicQuerySet` is the error message about?

Since our documentation build is "nitpicky", we have to add all classes that do not have a documentation to [nitpick_ignore](https://github.com/digitalfabrik/integreat-cms/blob/13aec2421525e77318c110b705c37453ea099fd4/sphinx/conf.py#L124).
However, in this case I think a documentation is available when we add `"django-polymorphic": ("https://django-polymorphic.readthedocs.io/en/stable/", None)` to our [intersphinx_mapping](https://github.com/digitalfabrik/integreat-cms/blob/13aec2421525e77318c110b705c37453ea099fd4/sphinx/conf.py#L63).

david-venhoff:
It looks like we already have such an entry: https://github.com/digitalfabrik/integreat-cms/blob/13aec2421525e77318c110b705c37453ea099fd4/sphinx/conf.py#L96-L99
So if this is an error on the side of django-polymorphic, should I just add the class to nitpick_ignore?

timoludwig:
> So if this is an error on the side of django-polymorphic, should I just add the class to nitpick_ignore?

Ah, interesting, I was just confused because the documentation for `PolymorphicQuerySet` exists, it's just listed under a different path:
[polymorphic.managers.PolymorphicQuerySet](https://django-polymorphic.readthedocs.io/en/latest/api/polymorphic.managers.html#polymorphic.managers.PolymorphicQuerySet) instead of `polymorphic.query.PolymorphicQuerySet`.

I can probably also fix this in [sphinxcontrib-django2](https://github.com/timoludwig/sphinxcontrib-django2), I will have a look.

david-venhoff:
Okay, I updated the file now

PR #1314, opened 2022-03-22
Fix layout of media library on small screens

timoludwig:
### Short description

Fix layout of media library on small screens


ulliholtgrave (approved):
Looks reasonable :)

Issue #1319, opened 2022-03-22
Error on disclaimer API


svenseeberg:
### Describe the Bug

Needs to be reproduced. Trace:

```
Mar 22 09:08:17 ERROR   django.request - Internal Server Error: /neuss/de/wp-json/extensions/v3/disclaimer/
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/api/decorators.py", line 124, in wrap
    return function(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/api/v3/imprint.py", line 56, in imprint
    imprint_translation = region.imprint.get_public_translation(language_slug)
AttributeError: 'NoneType' object has no attribute 'get_public_translation'
```

PR #1321, opened 2022-03-22
Use None-Check instead of hasattr() in imprint api

ulliholtgrave:
### Short description

Even though our region model has no `impint` attribute, the hasattr() method return true. Therefore, I replaced it with a None check.

### Resolved issues

Fixes: #1319

### DIFF
--- a/integreat_cms/api/v3/imprint.py
+++ b/integreat_cms/api/v3/imprint.py
@@ -1,10 +1,14 @@
 """
 This module includes functions related to the imprint API endpoint.
 """
+import logging
+
 from django.conf import settings
 from django.http import JsonResponse
 
 from ..decorators import json_response
+
+logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


integreat_cms/api/v3/imprint.py
```diff
@@ -63,4 +68,9 @@ def imprint(request, region_slug, language_slug):
             if imprint_default_translation:
                 return JsonResponse(transform_imprint(imprint_default_translation))
     # If imprint does not exist, return an empty response. Turn off Safe-Mode to allow serializing arrays
+    logger.info(
+        "The imprint for region %r in the language %s was request, but could not be returned",
```
timoludwig, line 72:
```suggestion
        "The imprint for region %r in the language %s does not exist",
```

timoludwig (commented):
Thanks 👍
I'd just change the message a bit

PR #1324, opened 2022-03-23
Update dependencies to mitigate minimist vulnerability

ulliholtgrave:
### Short description

Updates our Python and JS libraries to the newest stable versions. This also comes with a small performance update for the JS compilation :tada: 

Fixes: https://github.com/digitalfabrik/integreat-cms/security/dependabot/6


timoludwig (approved):
LGTM! :+1:

Issue #1054, opened 2021-12-07
Deliver fallback translations for live content


timoludwig:
### Motivation

When live content of other regions is included, we want to make sure the content is available in all languages, even if the source region doesn't provide them.

### Proposed Solution

In all places where live content is embedded (API and page preview), check whether the translation of the live content in the specific language exists and if not, provide the live content in the source region's default language.

### Additional Context

This is especially useful for attached addresses.

PR #1331, opened 2022-03-31
Provide fallback translations for live content

david-venhoff:
### Short description

This pr provides an error message for mirrored pages where the selected languages does not exist, and provides existing languages as links.

### Proposed changes

- Add two fields to the language model for translating the error message
- Provide the error message in the api in the page preview

### Resolved issues

Fixes: #1054


integreat_cms/cms/models/pages/page_translation.py
```diff
@@ -94,24 +94,51 @@ def short_url(self):
...
+            else self.language.message_live_content_not_available
+        )
+        choices = (
+            self.page.mirrored_page.prefetched_major_public_translations_by_language_slug
+        )
+        error_message += "<br>" + "<br>".join(
+            f"&emsp;* <a href={settings.WEBAPP_URL + choice.get_absolute_url()}>{choice.language.native_name}</a>"
+            for choice in choices.values()
+        )
+        return f"<p>{error_message}</p>"
```
timoludwig, line 126:
```suggestion
        # Get all translations of this page which have a corresponding translation of the mirrored page
        translations = filter(
            None,
            [
                self.page.get_public_translation(language_slug)
                for language_slug in self.page.mirrored_page.prefetched_major_public_translations_by_language_slug.keys()
            ],
        )
        # Create bullet points for each available translation
        choices = "</li><li>".join(
            f"<a href={settings.WEBAPP_URL + translation.get_absolute_url()}>{translation.language.native_name}</a>"
            for translation in translations
        )
        return f"<p>{error_message}</p><ul><li>{choices}</li></ul>"
```
Two ideas:
1. I think it would be better to link to the translations of the current page instead of the mirrored page. Also, we should probably only list options where both translations exist - those of the current page and those of the mirrored page.
2. I think bullet points make more sense in this case. Sure, they don't look good in the preview at the moment, but this is a general problem (see issue  #496) and isn't necessarily part of this PR.

integreat_cms/cms/views/pages/page_actions.py
```diff
@@ -168,14 +168,13 @@ def view_page(request, page_id, region_slug, language_slug):
     template_name = "pages/page_view.html"
 
     page_translation = page.get_translation(language_slug)
-    mirrored_translation = page.get_mirrored_page_translation(language_slug)
-
+    is_mirrored = bool(page.mirrored_page)
```
timoludwig, line 171:
```suggestion
```

timoludwig (commented):
I would just adapt the list of available languages to link to the current page instead of the mirrored page and use bullet points instead of asterisks.

david-venhoff:
Thank you for the feedback, I added your suggestions now :+1:

timoludwig (approved):
Thanks a lot! :tada:

Issue #1320, opened 2022-03-22
Different tree_id of page/child pages


svenseeberg:
### Describe the Bug
In Malte somehow the tree_id of a subpage got manually changed to another id when saving a new translation: The page with id 3240 had the tree id 417 and its parent 3239 had the tree id 418. I now manually assigned tree id 418 to the subpage and hope that it fixes the problem for now. But we have to investigate how this could happen.

svenseeberg:
2 observations:
- schneeberg was migrated from Wordpress - this could be the source of the issue
- Creating new pages with sub pages does work without issues

svenseeberg:
As described in #1317, a race condition could be the source of the problem: https://github.com/django-treebeard/django-treebeard/issues/53

svenseeberg:
Same error again due to broken tree_id:
```
  id  |         created_date          | explicitly_archived | lft | rgt | tree_id | depth | mirrored_page_first | icon_id | mirrored_page_id | organization_id | parent_id | region_id | api_token 
------+-------------------------------+---------------------+-----+-----+---------+-------+---------------------+---------+------------------+-----------------+-----------+-----------+-----------
 3300 | 2022-03-28 10:39:38.571114+00 | f                   |   2 |   3 |     929 |     2 | f                   |         |             2759 |                 |      3299 |       165 | 
 3301 | 2022-03-28 10:40:18.186689+00 | f                   |   4 |   5 |     929 |     2 | f                   |         |             2761 |                 |      3299 |       165 | 
 3299 | 2022-03-28 10:39:04.33347+00  | f                   |   1 |   6 |     930 |     1 | f                   |   18628 |             2760 |                 |           |       165 | 
 ```

All edits were performed consecutively in one browser tab. No other user was present. That makes a race condition less likely.

PR #1334, opened 2022-04-01
Use transactions for tree manipulations

svenseeberg:
Apply changes to MPTT tree structures in SQL tables with transactions to prevent race conditions with other requests.

### Short description
Use the `@transaction.atomic` decorator for all views that update MPTT trees.

Fixes: #1320 (maybe)

### DIFF
--- a/integreat_cms/cms/views/pages/page_form_view.py
+++ b/integreat_cms/cms/views/pages/page_form_view.py
@@ -7,6 +7,7 @@ from django.urls import reverse
 from django.utils.decorators import method_decorator
 from django.utils.translation import ugettext as _
 from django.views.generic import TemplateView
+from django.db import transaction
 
 from ...constants import status, text_directions
 from ...decorators import permission_required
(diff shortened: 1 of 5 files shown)


integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -176,6 +177,7 @@ def get(self, request, *args, **kwargs):
             },
         )
 
+    @transaction.atomic
```
timoludwig, line 180: Maybe let's add this to the post-function of `LanguageTreeNodeFormView` as well?

timoludwig (approved):
Thanks! :+1:
Let's try it out!

integreat_cms/cms/views/pages/page_form_view.py
svenseeberg: Yes, I missed that one.

timoludwig (approved):
:tada:

Issue #1325, opened 2022-03-24
Cannot delete mirrored page


david-venhoff:
### Describe the Bug

Right now a crash occurs when trying to delete a mirrored page:
![grafik](https://user-images.githubusercontent.com/78504586/159928712-32ffbe18-6266-487a-b033-bc62fb57ed10.png)

### Steps to Reproduce

1. Go to a mirrored page
2. Click on delete

### Expected Behavior

Error message which says that it is not possible to delete a mirrored page 

### Actual Behavior

Crash

PR #1338, opened 2022-04-02
Fix error when deleting page which was embedded as live content

timoludwig:
### Short description

Fix error when deleting page which was embedded as live content

### Proposed changes

- Disable possibility to delete a page which was referenced in another page as live content

### Resolved issues

Fixes: #1325


ulliholtgrave (approved):
Thanks. That looks very nice and I like that you also added the corresponding pages that are hindering the deletion :+1:

Issue #1001, opened 2021-10-30
Add option to deliver fallback for missing translations


timoludwig:
### Motivation

Parts of the content are less language-dependent than others (e.g. lists of contact-addresses) or change so frequently that they are not translated via the normal workflow. Instead, some regions want to show the translation in the region's default language per default for translations that are missing.

### Proposed Solution

- For each content type (`Page`, `ImprintPage`, `Event`, `POI`), add a new option to the region model to determine whether missing translations should be delivered in the region's default language.
- In the API, deliver the fallback translation for each missing translation if this option is enabled
- Adapt the translation report and icons in the page list to make it more obvious that the translations are in fact missing, but the fallback translations are shown in the apps instead.

ulliholtgrave:
Is this the same behavior as in the Wordpress system? As you can see in the screenshots, we have pages that are available in different languages (see the extra language for the corona information).
![grafik](https://user-images.githubusercontent.com/15608678/139572981-34a992a4-4595-4a97-b1ab-21fa8a91f15d.png)

![grafik](https://user-images.githubusercontent.com/15608678/139573031-ce75c952-ed05-4fab-bb91-1c8a32df1928.png)

svenseeberg:
As the app uses slugs for identifying objects (pages, news, etc) we need to ask the app team what happens if the same slug is used for another language.

timoludwig:
> As the app uses slugs for identifying objects (pages, news, etc) we need to ask the app team what happens if the same slug is used for another language.

I don't think this has anything to do with this issue. We already allow the same slug for different languages, even for manually created translations, so if this is a problem, it's one independently from this issue.

But this doesn't seem to be an issue, e.g.
- https://integreat.app/augsburg/de/corona-virus
- https://integreat.app/augsburg/en/corona-virus

work just fine. Probably the app uses the whole path `augsburg/de/corona-virus` and `augsburg/en/corona-virus` as identifier for a translation.

timoludwig:
This issue is no longer as relevant, because we need duplicates for individual pages anyway (see issue #1053).
So we should keep this idea in our minds, but let's move that to the backlog and think about it another time.

PR #1339, opened 2022-04-04
Add option to show content in default language as fallback

JoeyStk:
### Short description
This PR adds an option to show the content for POIs and events in the default language if there is no translation.

### Proposed changes
- I added the checkbox to the Region model and form
- I added the logic to the API

### Resolved issues
Fixes: #1001 
Fixes: #1411

### DIFF
--- a/tests/api/expected-outputs/api_augsburg_de_children_depth_2.json
+++ b/tests/api/expected-outputs/api_augsburg_de_children_depth_2.json
@@ -1,432 +1,432 @@
 [
-    {
-        "id": 75,
-        "url": "http://localhost:8000/augsburg/de/willkommen/",
-        "path": "/augsburg/de/willkommen/",
-        "title": "Willkommen",
-        "modified_gmt": "2019-08-12T07:57:42.456Z",
-        "excerpt": "Willkommen in Augsburg",
-        "content": "<p>Willkommen in Augsburg</p>",
-        "parent": {
-            "id": 0,
-            "url": null,
-            "path": null
-        },
-        "order": 2,
-        "available_languages": {
-            "en": {
-                "id": 7,
-                "url": "http://localhost:8000/augsburg/en/welcome/",
-                "path": "/augsburg/en/welcome/"
-            },
-            "ar": {
-                "id": 13,
-                "url": "http://localhost:8000/augsburg/ar/\u0645\u0639\u0644\u0648\u0645\u0627\u062a-\u0627\u0644\u0648\u0635\u0648\u0644/",
-                "path": "/augsburg/ar/\u0645\u0639\u0644\u0648\u0645\u0627\u062a-\u0627\u0644\u0648\u0635\u0648\u0644/"
-            },
-            "fa": {
-                "id": 11,
-                "url": "http://localhost:8000/augsburg/fa/\u06af\u0627\u0645-\u0646\u062e\u0633\u062a/",
-                "path": "/augsburg/fa/\u06af\u0627\u0645-\u0646\u062e\u0633\u062a/"
-            }
-        },
-        "thumbnail": null,
-        "hash": null
+  {
+    "id": 75,
+    "url": "http://localhost:8000/augsburg/de/willkommen/",
+    "path": "/augsburg/de/willkommen/",
+    "title": "Willkommen",
+    "modified_gmt": "2019-08-12T07:57:42.456Z",
+    "excerpt": "Willkommen in Augsburg",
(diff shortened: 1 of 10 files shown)


timoludwig (commented):
Thanks a lot! :rocket:

In general, it works as expected, however there are a few minor problems:
1. When the default translation does only exist as draft, it throws the error `'NoneType' object has no attribute 'event'`
2. The [available_languages](https://github.com/digitalfabrik/integreat-cms/blob/f44890a48ab520eaaf7217dfcfc3a0f51bb587a7/integreat_cms/cms/models/abstract_content_translation.py#L183) attribute of the translation does not list the fallback translations
3. In the sitemap, the fallback translations do not appear - but if this is too complicated, we can also open another issue for this.

JoeyStk:
This push should fix the issues that you listed, but I'm not confident, that how I implemented it, is the best way. Maybe you have an idea for a more simple way to solve the issues.

integreat_cms/api/v3/events.py
```diff
@@ -152,12 +157,18 @@ def events(request, region_slug, language_slug):
...
     region = request.region
     result = []
+    language = region.get_language_or_404(language_slug, only_active=True)
     now = timezone.now().date()
     for event in region.events.prefetch_public_translations().filter(archived=False):
         event_translation = event.get_public_translation(language_slug)
+        if not event_translation and region.use_fallback_language:
+            event_translation = event.get_public_translation(
+                region.default_language.slug
+            )
```
timoludwig, line 167: If we do not set the language attribute of the fallback translation, then the link will be incorrect in the API response.
What problems exactly did you face with the "fake" language?
E.g. for the test data, the endpoint `http://localhost:8000/api/augsburg/ar/events/` should return the test event with the "fake" url `http://localhost:8000/augsburg/ar/events/test-veranstaltung/` and the available languages `de`, `en` and `fa` where the link/path always points to the available language, no matter whether it's a real translation or a fake one:
```
"id": 2,
"url": "http://localhost:8000/augsburg/ar/events/test-veranstaltung/",
"path": "/augsburg/ar/events/test-veranstaltung/",

...

"available_languages": {
    "de": {
        "id": 2,
        "url": "http://localhost:8000/augsburg/de/events/test-veranstaltung/",
        "path": "/augsburg/de/events/test-veranstaltung/"
    },
    "en": {
        "id": 3,
        "url": "http://localhost:8000/augsburg/en/events/test-event/",
        "path": "/augsburg/en/events/test-event/"
    },
    "fa": {
        "id": 2,
        "url": "http://localhost:8000/augsburg/fa/events/test-veranstaltung/",
        "path": "/augsburg/fa/events/test-veranstaltung/"
    },
},
```
I don't know yet what to do about the ids, but since the app is not using them anyway, I think it's not that important that the ids make sense.

integreat_cms/cms/models/abstract_content_translation.py
```diff
@@ -200,12 +200,23 @@ def available_languages(self):
...
+                    continue
             other_translation = self.foreign_object.get_public_translation(
                 language.slug
             )
+            if not other_translation and self.foreign_object.region.use_fallback_language:
+                other_translation = self.foreign_object.get_public_translation(
+                    self.foreign_object.region.default_language.slug
+                )
+                if other_translation:
+                    other_translation.language = self.foreign_object.region.default_language
```
timoludwig, line 219: Keep in mind that the abstract content translation is used for all content types, but fallback translation should only be used for events and locations.

timoludwig (commented):
Thanks! I got a few errors and could not test this in detail.
I'm not sure how the current code is supposed to work exactly, so I don't have any ideas yet on how to improve it. Could you maybe take another look at it?

integreat_cms/cms/templates/regions/region_form.html
```diff
@@ -128,6 +128,12 @@ <h3 class="heading font-bold text-black"><i data-feather="star" class="mr-2"></i
                         <label for="{{ form.locations_enabled.id_for_label }}">{{ form.locations_enabled.label }}</label>
                         <div class="help-text">{{ form.locations_enabled.help_text }}</div>
                     </div>
+                    <div>
+                        <!-- Options related to Fallback language -->
+                        {% render_field form.fallback_translations_enabled %}
+                        <label for="{{ form.fallback_translations_enabled.id_for_label }}">{{ form.fallback_translations_enabled.label }}</label>
```
ulliholtgrave, line 134: Could you add some text-ellipsis to this label. Otherwise, I have some strange display due to the length of the text.
grafik

PR #1340, opened 2022-04-04
Update pre-commit hooks

timoludwig:
### Short description

There seems to be a problem with our black pre-commit hook, so I suggest to update it.

### Proposed changes

- Update pre-commit hooks


ulliholtgrave (approved):
Looks good :+1:

Issue #1226, opened 2022-02-21
Issue with Cache in children endpoint


svenseeberg:
**This issue needs further investigation/testing**. For Malte the cache is currently disabled and the performance is acceptable. The priority of this issue depends on the performance of the migrated Integreat CMS.

### Describe the Bug
When calling random pages we sometimes get the following stack trace:
```
Feb 21 22:53:01 ERROR   django.request - Internal Server Error: /leakarlsruhefelsstrasse/de/wp-json/extensions/v3/children/
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/api/decorators.py", line 178, in wrap
    return func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/api/decorators.py", line 124, in wrap
    return function(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/api/v3/pages.py", line 205, in children
    result.append(transform_page(public_translation))
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/api/v3/pages.py", line 22, in transform_page
    parent_page = page_translation.page.cached_parent
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/utils/functional.py", line 48, in __get__
    res = instance.__dict__[self.name] = self.func(instance)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/models/abstract_tree_node.py", line 210, in cached_parent
    return self.get_cached_ancestors()[-1]
IndexError: list index out of range
```
This trace persists until the cache is flushed.

### Steps to Reproduce

Happens randomly. Visit pages until it happens.

For the time being we could use a simple filter instead of a cache operation in https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/api/v3/pages.py#L231

timoludwig:
[...]

PR #1343, opened 2022-04-05
Fix cache invalidation after page tree move

melegiul:
### Short description

Fix cache invalidation in children endpoint

### Proposed changes

When a complete page tree is moved, it's tree_id is changed, so it is necessary to clear the cache for the page translation - page relation.

### Resolved issues

Fixes: #1226

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ UNRELEASED
 * [ [#1198](https://github.com/digitalfabrik/integreat-cms/issues/1198) ] Check availability for DeepL bulk actions
 * [ [#1293](https://github.com/digitalfabrik/integreat-cms/issues/1293) ] Enable login via email address
 * [ [#1327](https://github.com/digitalfabrik/integreat-cms/issues/1327) ] Fix page PDF export
+* [ [#1226](https://github.com/digitalfabrik/integreat-cms/issues/1226) ] Fix page tree fields cache invalidation
 
 
 2022.3.6
--- a/integreat_cms/cms/models/pages/page.py
+++ b/integreat_cms/cms/models/pages/page.py
@@ -8,6 +8,8 @@ from django.utils.functional import cached_property
 from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext_lazy as _
 
+from cacheops import invalidate_model
+
 from treebeard.ns_tree import NS_NodeQuerySet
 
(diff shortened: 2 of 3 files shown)


integreat_cms/cms/models/abstract_tree_node.py
```diff
@@ -284,6 +285,10 @@ def move(self, target, pos=None):
         invalidate_model(self.__class__)
         super().move(target=target, pos=pos)
         invalidate_model(self.__class__)
+        if self.depth == 1:
+            # Moving a complete tree causes a tree_id change in the page model
+            # so the cache of the page translation - page relation has to be cleared
+            invalidate_model(PageTranslation)
```
timoludwig, line 291: This model is also used as base for the language tree node, so I think we should move this invalidation to the page model.
Also, I think just to be 100% save we should invalidate the translation model in any case - even though the `tree_id` does not change, the `lft` and `rgt` values change which might also cause caching issues in some weird edge cases.

timoludwig (commented):
Great, thank you for investigating the bug and fixing the issue! :clap:

integreat_cms/cms/models/abstract_tree_node.py
melegiul: Thank you very much for your review, you are right, I moved the invalidation to the page model

integreat_cms/cms/views/pages/page_actions.py
```diff
@@ -434,6 +436,9 @@ def move_page(request, region_slug, language_slug, page_id, target_id, position)
         # (The move()-method executes raw sql which might cause problems if the instance isn't fetched again)
         page = Page.objects.get(id=page_id)
         page.save()
+        # Moving tree nodes potentially causes changes to the fields tree_id, lft and rgt in the tree node model
+        # so the cache of page translations has to be cleared, because of it's relation to the page model
+        invalidate_model(PageTranslation)
```
timoludwig, line 441: Sorry, another request: Could you mode this to the actual `move()`-method of the page model (you probably need to create this first and do a call to `super().move()` inside the new method)? Since moving can also be performed via the page form (not only drag & drop), I feel like that's the more flexible solution.

melegiul: Ah, yes, I missed the page form :see_no_evil: 
Thanks again for catching this

timoludwig (approved):
Thanks a lot! :pray:

Issue #1354, opened 2022-04-12
Order field in page API not working on root level


timoludwig:
### Describe the Bug

We use the `lft` field of the MPTT model to convert them to `order` fields which is used by the app.
However, on the root level this does not work, because all root pages have the same `lft` values.

### Steps to Reproduce

Access API and inspect root pages

### Expected Behavior

The root pages should be ordered by their tree ids

### Actual Behavior

All root pages have the `order` field `1` because it's their `lft` value

PR #1355, opened 2022-04-12
Fix order of root pages

timoludwig:
### Short description

Fix order of root pages

### Proposed changes

- Order root pages by the `tree_id` instead of the `lft` field

### Resolved issues

Fixes: #1354


ulliholtgrave (approved):
Perfect :+1:

Issue #1353, opened 2022-04-12
Add tunews setting to region model


timoludwig:
### Motivation

To activate the tunews in specific regions, we need to store this config in the region model and deliver it via the API.

### Proposed Solution

- Add new boolean field `tunews_enabled` (with default `False`) to the region model
- Add the new field to the region form
- Deliver this field in the region API with the key `tunews`

PR #1358, opened 2022-04-13
Add tunews setting to region

david-venhoff:
### Short description
This pr adds a tunews setting to the region model

### Proposed changes

- Add field to region model
- Allow changing that field in the region form
- Return the field via the region api endpoint

### Resolved issues

Fixes: #1353


timoludwig (approved):
Thanks a lot, works exactly as expected! :+1:

PR #1361, opened 2022-04-15
Update packages and fix pylint issues in the ini file

ulliholtgrave:
### Short description

This PR updates our JS and python packages to their newest version. 

We should keep in mind that our setup.cfg will be deprecated by some point and we should consider switching to a pyproject.toml setup (see Issue #1362 ).

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -286,7 +286,7 @@
                 "sha256:2857e29ff0d34db842cd7ca3230549d1a697f96ee6d3fb071cfa6c7393832597",
                 "sha256:6881edbebdb17b39b4eaaa821b438bf6eddffb4468cf344f09f89def34a8b1df"
             ],
-            "markers": "python_full_version >= '3.5.0'",
+            "markers": "python_version >= '3'",
             "version": "==2.0.12"
         },
         "click": {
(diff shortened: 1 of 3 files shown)


integreat_cms/integreat-cms-cli
```diff
@@ -32,7 +40,7 @@ def main():
             # pylint: disable=import-outside-toplevel,unused-import
             import django
         except ImportError:
-            raise ImportError(
+            raise ImportError from Exception(
```
timoludwig, line 43: This does not make sense to me, as it _looses_ information in comparison to the previous solution. I guess what you intended was something like this?
```
        except ImportError as e:
            raise ImportError(
                "Couldn't import Django. Are you sure it's installed and "
                "available on your PYTHONPATH environment variable? Did you "
                "forget to activate a virtual environment?"
            ) from e
```
I cannot make this suggestion because it's out of reach.

integreat_cms/integreat-cms-cli
```diff
@@ -18,7 +21,12 @@ def read_config():
 
 
 def main():
-    """" Application entry point """
+    """ "
```
timoludwig, line 24:
```suggestion
    """
```

integreat_cms/integreat-cms-cli
```diff
@@ -18,7 +21,12 @@ def read_config():
...
+    Application entry point
+
+    :raises ImportError: If the Django framework is not installed or not in the path, an import Error is risen.
+
```
timoludwig, line 28:
```suggestion
    :raises ImportError: If the Django framework is not installed or not in the path, an import error is risen.
```

Pipfile.lock
```diff
@@ -1972,11 +1975,11 @@
...
         "pylint": {
             "hashes": [
-                "sha256:9d945a73640e1fec07ee34b42f5669b770c759acd536ec7b16d7e4b87a9c9ff9",
-                "sha256:daabda3f7ed9d1c60f52d563b1b854632fd90035bcf01443e234d3dc794e3b74"
+                "sha256:c149694cfdeaee1aa2465e6eaab84c87a881a7d55e6e93e09466be7164764d1e",
+                "sha256:dab221658368c7a05242e673c275c488670144123f4bd262b2777249c1c0de9b"
             ],
             "index": "pypi",
-            "version": "==2.12.2"
+            "version": "==2.13.5"
```
timoludwig, line 1982:
```suggestion
            "version": "==2.12.2"
```
The problem with updating pylint is that it currently has a bug in one of its dependencies (see https://github.com/PyCQA/pylint/issues/6012 and https://github.com/tartley/colorama/issues/86) which is not yet released (see https://github.com/tartley/colorama/issues/341), so we'll have to wait until they publish a new release before we can update pylint to the next version.

Pipfile.lock
```diff
@@ -1972,11 +1975,11 @@
         },
         "pylint": {
             "hashes": [
-                "sha256:9d945a73640e1fec07ee34b42f5669b770c759acd536ec7b16d7e4b87a9c9ff9",
-                "sha256:daabda3f7ed9d1c60f52d563b1b854632fd90035bcf01443e234d3dc794e3b74"
+                "sha256:c149694cfdeaee1aa2465e6eaab84c87a881a7d55e6e93e09466be7164764d1e",
+                "sha256:dab221658368c7a05242e673c275c488670144123f4bd262b2777249c1c0de9b"
```
timoludwig, line 1979:
```suggestion
                "sha256:9d945a73640e1fec07ee34b42f5669b770c759acd536ec7b16d7e4b87a9c9ff9",
                "sha256:daabda3f7ed9d1c60f52d563b1b854632fd90035bcf01443e234d3dc794e3b74"
```

timoludwig (commented):
Thanks! :+1:

Do you have a resource that claims that `setup.cfg` is deprecated (soonish)? I know that `pyproject.toml` is the "new" format, but at the moment, it cannot replace `setup.cfg`, can it?
(it can just replace some of the configuration values, but at the moment this would just mean one additional file, I'm not sure whether that's an improvement...)

ulliholtgrave:
I resolved all issues and created https://github.com/digitalfabrik/integreat-cms/issues/1362 as a reminder to keep track of the current development in Setup Tools.

timoludwig (approved):
Thanks! :+1:

Issue #1093, opened 2022-01-06
Add possibility to set individual logo for CMS branding


ulliholtgrave:
### Motivation

Provide the possibility to change the logo to a specific one to provide the opportunity to brand the cms to your organization. 

### Proposed Solution

Add a setting that defines the path to the logo and use it in all relevant files (e.g. templates and tailwind with integreat_logo as default value).

svenseeberg:
@ulliholtgrave I just realized that we do not have a custom logo for Malte in the current Malte CMS. Let's skip this issue for now, as this is more work than expected. We probably want to change this issue into "Support custom CSS" or similar.

ulliholtgrave:
@svenseeberg But aren't we using the default Malteser logo in the current cms.

svenseeberg:
No, we're not :)

PR #1364, opened 2022-04-19
Add Malte and Aschaffenburg brandings

timoludwig:
### Short description

This adds the option to brand the CMS for Malte and Aschaffenburg.
In the long run, we might also add the possibility to set completely custom logos and themes (see #1093), but this is a first start to support a fixed set of inbuilt themes.

### Proposed changes

- Add option to support different brandings
- Use svg logos for Integreat instead of jpg & png (only use png for PDF export)
- Add logos and accent colors for Malte and Aschaffenburg

You can test the different logos by editing your local config file in `/etc/integreat-cms.ini` or set the environment variable `export INTEGREAT_CMS_BRANDING=malte` or `export INTEGREAT_CMS_BRANDING=aschaffenburg` before starting the dev server.

### Resolved issues

Fixes partially: #1093


svenseeberg:
I think we should add a note to our README file that the license does not cover logos in this repository.

timoludwig:
> I think we should add a note to our README file that the license does not cover logos in this repository.

Good point! I was unsure about this because we're also not doing this in our [other repos](https://github.com/digitalfabrik/integreat-app/tree/main/build-configs/aschaffenburg/assets) and I thought it might be covered by this section of the Apache license:

>    6. Trademarks. This License does not grant permission to use the trade
      names, trademarks, service marks, or product names of the Licensor,
      except as required for reasonable and customary use in describing the
      origin of the Work and reproducing the content of the NOTICE file.

Or isn't this sufficient when the logos aren't registered trademarks? Do we need a different license file for the logos, or is it just sufficient to exclude them from the general license, leaving them "without license"?

README.md
```diff
@@ -48,3 +48,10 @@ cd integreat-cms
 For detailed instructions, tutorials and the source code reference have a look at our great documentation:
 
 <p align="center">:notebook: https://digitalfabrik.github.io/integreat-cms/</p>
+
+
+## License
+
+This project is licensed under the Apache 2.0 License, see [LICENSE.txt](./LICENSE.txt)
+
+All files in [./integreat_cms/static/src/logos/](./integreat_cms/static/src/logos/) are not covered by this license and may only be used with specific permission of the copyright holder.
```
timoludwig, line 57: @svenseeberg Do you think this is sufficient?

svenseeberg: Sounds good, thanks!

ulliholtgrave (approved):
I tried all variants and the mentioned functions. Looks good to me and thanks for your effort! :+1:

Issue #1366, opened 2022-04-22
Monthly recurring event does not work for a "Monday"


FritjofK:
### Describe the Bug

For a monthly recurring event which is a **Monday** the error 

"Wochentag: Kein Wochentag für die monatliche Wiederholung eingegeben" 

is shown. It happens on all weeks for a monthly recurring event. 

### Steps to Reproduce

1. Go to Veranstaltungen in Malte CMS
2. Click on Veranstaltung erstellen -> wiederholend
3. Select any date and chose "monatlich" and "Montag" as day
4. See error "Wochentag: Kein Wochentag für die monatliche Wiederholung eingegeben" 

### Expected Behavior

Just like any other day. 

### Additional Information

<details>
  <summary>Traceback</summary>

![image002](https://user-images.githubusercontent.com/16462172/164741667-8c629fba-c92c-4272-9beb-1ec226921ee8.png)

![image001](https://user-images.githubusercontent.com/16462172/164741711-d163d3e0-f6cf-40cd-95be-6985008cd1e8.png)

</details>

PR #1367, opened 2022-04-22
Allow events to be monthly recurring on mondays

timoludwig:
### Short description

Allow events to be monthly recurring on mondays

### Proposed changes

- Check for `None` instead of implicitly casting the value into a boolean (which doesn't work because monday is defined as the integer `0` which evaluates to `False`...

### Resolved issues

Fixes: #1366

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 UNRELEASED
 ----------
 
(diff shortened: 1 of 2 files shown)


ulliholtgrave (approved):
Thanks. That fixes the error :+1:

Issue #1351, opened 2022-04-11
[XLIFF] New translations are imported with empty slugs


timoludwig:
### Describe the Bug

When translations that do not yet exist in the database are imported via XLIFF, the resulting objects have empty slugs.

### Steps to Reproduce

1. Export XLIFF file for pages that do not have a translation in the target language
2. Write dummy titles into the XLIFF files
3. Import XLIFF files

### Expected Behavior

The translation slug should be automatically derived from the translated title

### Actual Behavior

The slug is empty and the translation object is saved with an empty slug

PR #1372, opened 2022-04-26
Modify slug in translation import via xliff

MizukiTemma:
### Short description

This PR solves the problem that a translation is saved with an empty slug when it is imported via XLIFF.

### Proposed changes

- Check the slug before saving, when a translation is imported.
- If it is empty, generate a slug from the title.

### Resolved issues

Fixes: #1351


timoludwig (commented):
Thanks! :+1:
This does set the slug, however it does not make sure that it's unique per region and language.
Could you use the [generate_unique_slug()](https://github.com/digitalfabrik/integreat-cms/blob/810bd2f3be861db0148a5e0410b5878ee59bfa17/integreat_cms/cms/utils/slug_utils.py#L52) function to generate a slug?
Since this step is already done in the process when the imported translation [is validated](https://github.com/digitalfabrik/integreat-cms/blob/c51bd2c365a407d792ebe30c420423e3641dd201/integreat_cms/xliff/utils.py#L469) (because the slug is generated in the [form's clean-method](https://github.com/digitalfabrik/integreat-cms/blob/c51bd2c365a407d792ebe30c420423e3641dd201/integreat_cms/cms/forms/pages/page_translation_form.py#L90)), maybe we can somehow extract the slug or write it back into the translation object?
```
page_translation.slug = page_translation_form.instance.slug
```
But that's just an idea, I didn't try this out...

integreat_cms/xliff/utils.py
```diff
@@ -473,6 +473,8 @@ def get_xliff_import_errors(request, page_translation, add_message_if_unchanged=
                 for message in page_translation_form.get_error_messages()
             ]
         )
+    else:
+        page_translation.slug = page_translation_form.instance.slug
```
timoludwig, line 477:
```suggestion
        # Use the unique slug generated by the form's clean-method for imported page translations
        page_translation.slug = page_translation_form.instance.slug
```

timoludwig (commented):
Thanks, this seems to work as expected! :+1:
Since this is a bit "hacky" (because a method that only reports errors is usually not considered to change the given object as side effect), I would add a few explanatory comments.
In addition to my inline suggestion, I would also add a short sentence to the method docstring of `get_xliff_import_errors()`, e.g.:
```
Validate an imported page translation and return all errors.
As a side effect, a unique slug is generated for the translation if it does not yet have one.
```
Apart from that, could you resolve the conflicts, fix the circleci errors and add a note to the changelog? Thanks!

timoludwig (commented):
Thanks! :+1:
Now, could you squash your commits? Then, this PR is ready to merge! :tada:

timoludwig (approved):
Thanks! :tada:

Issue #1369, opened 2022-04-23
CSS cached after change


timoludwig:
### Describe the Bug

While our JS-code is bundled to a file which contains the hash of its content, we do not have the same workflow for our css files.
So whenver we change something in the css files, the users might still have an old copy of the css file in their cache.

### Steps to Reproduce

See e.g. html source code:

```
<script src="/static/main.98399dbf796c38c23dbd.js" ></script>
<link href="/static/main.css" rel="stylesheet" />
```

### Expected Behavior

We should add the same hash-based-filename for css files like for js files

### Actual Behavior

The css filename does not change when its content does

PR #1373, opened 2022-04-27
Use contenthash for css files

ulliholtgrave:
### Short description

This PR should fix the misbehavior of the cache dealing with CSS files. However, I am not completely sure if this affects all relevant CSS files. @timoludwig Can you verify this? :)

### Proposed changes

- Add contenthash to MiniCSSPlugin

### Resolved issues

Fixes: #1369

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 UNRELEASED
 ----------
 
+* [ [#1369](https://github.com/digitalfabrik/integreat-cms/issues/1369) ] Add contenthash to CSS files for correct cache handling
+
 
 2022.4.2
 --------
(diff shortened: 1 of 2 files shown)


ulliholtgrave:
Okay, There seems to be a problem with the PDF-Export... I'll look into it

timoludwig:
> Okay, There seems to be a problem with the PDF-Export... I'll look into it

The css of the pdf should not be minified, because there was a bug with minified font face definitions (see https://github.com/xhtml2pdf/xhtml2pdf/pull/609)... my PR was merged, but no new version was released, so for the moment we have to keep excluding the pdf css from the minifier. Probably, we need change the `exclude: "pdf.css",` to a regex which contains a variable hash... (or we don't use the hash in the pdf because this is just internal for the pdf generation and never really cached by the browser.

ulliholtgrave:
Okay, or we could just wait for the next release of XHTML2PDF and can merge this afterwards.

timoludwig (approved):
And yes, this handles all our css files:
- main.[hash].css
- editor_content.[hash].css
- editor.[hash].css (not generated at the moment because this module only contains js code)
- pdf.css (excluded from hash)

Issue #1350, opened 2022-04-11
Improve PDF export


timoludwig:
### Motivation

At the moment, the PDF export feels a bit rudimentary and buggy.

### Proposed Solution

- Add cover page with city & integreat logos
- Visualize the hierarchy in the table of contents
- Fix font of headlines (should be raleway)
- Fix images & icons in the content as well as line breaks:

| Content        | PDF           |
| ------------- |:-------------:|
| ![pdf-source](https://user-images.githubusercontent.com/16021269/162759931-b9c0769e-eabf-419b-9474-e1da453d19bb.png) | ![pdf-pdf](https://user-images.githubusercontent.com/16021269/162759908-ea7a9c23-684b-49cc-87d8-abb5644457e4.png) |

osmers:
The Malte PDF should have the Malte Logo in the bottom right hand of the PDF, right?

timoludwig:
> The Malte PDF should have the Malte Logo in the bottom right hand of the PDF, right?

At the moment, we don't have the possibility to set individual logos other than the Integreat logo, but it's planned (see issue #1093).
Did the PDF export in WordPress have a Malte logo?

osmers:
I don't remember whether it had a Malte Logo - I just found the Integreat Logo in the Malte PDF confusing...especially for people who only know Malte...
Does the individual logo issue include Kommunen logos? Bcs for Integreat it would be great if they could appear additionally to the Integreat Logo in the top right hand corner (or bottom left hand maybe?)

timoludwig:
Regarding the Malte Logo I opened PR #1364 which handles all logos in the CMS, so as soon as this is merged, there will the Malte logo in the PDF footer.
And the individual region logos are already shown in the top right corner, they just have to be set in the region form.

timoludwig:
@osmers the PR is now merged and released, so the Malte CMS is now styled in the Malte design which includes the logo in the footer of exported PDFs.

ulliholtgrave:
![grafik](https://user-images.githubusercontent.com/15608678/165490539-f6e245e0-5652-47c1-8061-900da921ec83.png)
@timoludwig I guess the problems with the images and icons are related to our implementations, but our file migration. It seems a few static files are also linking to the old TUM-Server, which are not redirected on the test system. With new or updated images this is no longer an issue.

[...]

PR #1375, opened 2022-04-30
Fix legacy media urls in PDF export

timoludwig:
### Short description

(Hopefully) fix legacy media urls in PDF export.
Since the directory structure of the media export is still in progress, we might need to adapt this slightly in the future, but I hope this already fixes most problems.

### Proposed changes

- Convert image urls to absolute paths on allowed hosts
- Convert the legacy url pattern `/region-slug/wp-content/uploads/` to the new path `/media/`

### Resolved issues

Fixes (partially): #1350


ulliholtgrave (approved):
Thanks :+1:  This seems to fix the problem for the case you mentioned in the last call. I propose that we merge if for now and deploy it on the test system with the Apache etc. to see if this fixes all the problems :)

Issue #726, opened 2021-03-03
Extend `POI` model


svenseeberg:
### Motivation
The map feature requires to have additional fields for our POI model. 

### Proposed Solution
The following (optional) attributes are necessary:
* Website address (URL)
* Contact e-mail address
* Contact phone number
* (Opening hours) TBD
* Icon (Original)
* Icon (Thumbnail_150x150)

This issue is deeply connected to #951 and could be solved together with it.

timoludwig:
When implementing this, the POI-model parts of #695 and #522 could be included as well:
- Add boolean field `show_location_on_map`
- Add text field `meta_description`

PR #1376, opened 2022-04-30
Add additional fields to location model

timoludwig:
### Short description

Add additional fields to location model

### Proposed changes

Add the following fields to the POI model:
- `website`
- `email`
- `phone_number`

### Resolved issues

Fixes: #726


ulliholtgrave:
I am not sure, but your PR still kinda misses the icon field, right?

timoludwig:
> I am not sure, but your PR still kinda misses the icon field, right? I guess it's kinda essential and we should at least add one field for it.

What do you mean? The POI model already has an [icon field](https://github.com/digitalfabrik/integreat-cms/blob/e975f628ebed012cd0177d2617819dbf5849c7fc/integreat_cms/cms/models/pois/poi.py#L41-L47) which is delivered [via the api](https://github.com/digitalfabrik/integreat-cms/blob/e975f628ebed012cd0177d2617819dbf5849c7fc/integreat_cms/api/v3/locations.py#L73).

ulliholtgrave (approved):
Ahh, yes you are right. Then I have no objections :+1:

Issue #1180, opened 2022-02-16
Save location on map without coordinates


svenseeberg:
### Describe the Bug
Not all users are able to search for coordinates. If the location is not set to be shown on the map (checkbox), it should be allowed to have empty coordinates.

Related: #1170

timoludwig:
I think the coordinates are required to render the POIs correctly on the app's map.
So we can't really make this optional, maybe only if the option to not show the location on the map is enabled?
Anyway, the ideal solution would be to request the coordinates via an API and auto-fill them (see also #1000)

svenseeberg:
Yes, the automatic solution would be best. However, we need some sort of validation function at some point anyways. Therefore 0, 0 should be okayish enough and could be the default values.

Issue #1000, opened 2021-10-30
Automatically validate location data of `POI` model


timoludwig:
### Motivation

At the moment, it is possible to insert incorrect data for the exact position of `POI`s. Since there are many possible pitfalls when creating `POI`s, we should validate the input and auto-fill as much as possible.
And since the map feature of the app breaks when we deliver incorrect data, this should ideally be implemented before the initial production release.

### Proposed Solution

Possible ideas (check effort vs. usability on each step):

- [ ] Validate whether an address format is correct
- [ ] Validate whether a street exists in the given city
- [ ] Suggest available streets on the selected city
- [ ] Auto-fill city after zip code has been inserted
- [ ] Auto-fill coordinates after address is complete
- [ ] If coordinates are manually changed after auto-fill, check for consistency with the address data and raise a warning if the positions are completely off

Check which data sources and APIs can be used for this task.

svenseeberg:
Honestly, I do not think that this is that important. As with all content that is created in the app, the content creators need to take care that everything is "correct". I think it would be most important to show a preview of the location on a small embedded map to simply show where the location actually is on a map. I would also argue that this is not necessary for a release.

ulliholtgrave:
I agree with Sven, a validation is surely a nice thing to have, but the wordpress doesn't have it. Therefore I suggest to postpone it to Version 2.0 and add the prio:high label.

timoludwig:
Alright, seems reasonable! :+1:

timoludwig:
We could use our geocoding service for this:
http://nominatim.maps.tuerantuer.org/nominatim/search.php?street=Am%20B%C3%BChl&postalcode=86199

API documentation: https://nominatim.org/release-docs/develop/api/Overview/

dkehne:
Isn't that also related to #1035 ?

timoludwig:
> Isn't that also related to #1035 ?

Yes, it's related, but can be implemented independently of each other.

timoludwig:
Since this is now mostly done, I decreased the prio of the remaining tasks and moved it back into the backlog

JoeyStk:
[...]

PR #1377, opened 2022-04-30
Automatically derive location coordinates from address

timoludwig:
### Short description

Automatically derive location coordinates from address

### Proposed changes

- Add geopy for Nominatim API client
- Automatically derive location coordinates from address
- Show warning when manual coordinates are too far away from Nominatim API result
- Make coordinates optional for locations not on map

### Resolved issues

Fixes (partially): #1000
Fixes: #1180


ulliholtgrave (changes requested):
Thank you very much. I tried it and everything seems to work just as expected :+1: 
Just one small issue: Could you add some initial debug message that the Normatim API is available/used? So something similar like the GVZ and DeepL API?

ulliholtgrave (approved):

ulliholtgrave:
Thanks for adding this API check :+1: I guess this is good to go, if the pylint issues are adressed :tada:

Issue #1245, opened 2022-02-23
Keep status of expanded tree when navigating back in browser


svenseeberg:
### Motivation
When opening a page for navigation in browser, the state of the tree should be kept (which parts of the tree are expanded/collapsed).

### Proposed Solution
Track the status of the tree in a Cookie(?) and automatically expand the tree as needed/set.

### Alternatives
Make it work in the browser history, which we disabled.

### Additional Context
We intentionally disabled browser cache right now.

timoludwig:
So #1179 was actually a feature and not a bug? :laughing:

timoludwig:
Why would we need a cookie for this? Can't we just revert the changes from #1192?

svenseeberg:
The issue was that the HTML tree seems to be in the history correctly, but then there are some issues with drag & drop.

svenseeberg:
https://developer.mozilla.org/en-US/docs/Web/API/History_API

timoludwig:
@svenseeberg but I guess the normal history API is also used on the back button? If the tree is broken after using the back-button, I see no reason why it should work using `window.history.back()`. But maybe I'm missing something :sweat_smile:

svenseeberg:
@timoludwig I was thinking that we can maybe fix things by pushing states to the history whenever we're firing AJAX calls.

timoludwig:
I enabled the browser cache of the page tree again in 0e2b49c2db6de28bfc01027ca541748ff6920aa7, but this does not lead to the tree being expanded on the back button. So apparently, the source of bug #1179 was not the browser cache, but something different, and removing the browser cache just was a workaround.

Regarding this issue: Yes, your approach makes sense, but is definitely not an easy task. With all our event handlers, this increases the potential for bugs which we cannot prevent with unit tests. Maybe the latest improvements to the page tree already provide most of the functionality that was the motivation of this issue?

svenseeberg:
I think we should redefine the scope of this issue. We could limit it to keeping the expanded/collapsed state of single nodes. And this could easily stored in a cookie.

timoludwig:
We could, yes. To me, it still sounds complicated and I'm not sure whether the benefit of the feature is big enough to justify the effort, but maybe I'm overestimating the complexity :wink:

PR #1379, opened 2022-05-02
Persist status of page tree expansion

david-venhoff:
### Short description

This pr stores the expansion state of the page tree in the browser session storage in order to reconstruct the previous state after navigating back to the page tree.
Since loading the dynamic page tree takes a few moments, it could make sense to display some sort of loading indicator, before the page tree can get expanded.

### Proposed changes

- Keep track of page row expansions
- On page load, or when the dynamic subpages loaded, restore the previous tree expansion

### Resolved issues

Fixes: #1245

### DIFF
--- a/integreat_cms/cms/views/pages/partial_page_tree_view.py
+++ b/integreat_cms/cms/views/pages/partial_page_tree_view.py
@@ -1,62 +1,77 @@
-from django.shortcuts import render
-from django.utils.decorators import method_decorator
-from django.views.generic import TemplateView
+from collections import defaultdict
+import json
 
-from integreat_cms.cms.views.pages.page_context_mixin import PageContextMixin
+from django.http import JsonResponse
+from django.template.loader import render_to_string
+from django.utils.translation import get_language
+from django.views.decorators.http import require_POST
 
+from ..pages.page_context_mixin import PageContextMixin
+from ...models.languages.language import Language
 from ...decorators import permission_required
 
 
-@method_decorator(permission_required("cms.view_page"), name="dispatch")
-class PartialPageTreeView(TemplateView, PageContextMixin):
+@permission_required("cms.view_page")
+@require_POST
+# pylint: disable=unused-argument
+def render_partial_page_tree_views(request, region_slug, language_slug):
+    r"""
+    Retrieve the rendered subtree of a given root page
+
+    :param request: The current request
+    :type request: ~django.http.HttpRequest
+
+    :param region_slug: The slug of the current region
+    :type region_slug: str
+
+    :param language_slug: The slug of the current language
+    :type language_slug: str
+
+    :return: The rendered template responses
+    :rtype: ~django.http.JsonResponse
     """
-    View for rendering a partial page tree
-    """
+    requested_tree_ids = [int(i) for i in json.loads(request.body.decode("utf-8"))]
 
-    #: Template for a partial page tree
-    template = "pages/_page_tree_children.html"
+    region = request.region
+    language = region.get_language_or_404(language_slug, only_active=True)
 
-    # pylint: disable=unused-argument
-    def get(self, request, *args, **kwargs):
-        r"""
-        Retrieve the rendered subtree of a given root page
+    backend_language = Language.objects.filter(slug=get_language()).first()
 
-        :param request: The current request
-        :type request: ~django.http.HttpRequest
+    all_pages = (
+        region.pages.filter(tree_id__in=requested_tree_ids)
+        .prefetch_major_translations()
+        .prefetch_related("mirroring_pages")
+        .cache_tree(archived=False)
+    )
 
-        :param \*args: The supplied arguments
-        :type \*args: list
+    pages_by_id = defaultdict(list)
+    for page in all_pages:
+        pages_by_id[page.tree_id].append(page)
 
-        :param \**kwargs: The supplied keyword arguments
-        :type \**kwargs: dict
-
-        :return: The rendered template response
-        :rtype: ~django.template.response.TemplateResponse
-        """
-        tree_id = int(kwargs.get("tree_id"))
-        region = request.region
-        language = region.get_language_or_404(
-            kwargs.get("language_slug"), only_active=True
-        )
+    sub_trees = []
+    for tree_id in requested_tree_ids:
+        # Skip page trees which do not exist
+        if tree_id not in pages_by_id:
+            continue
         # Get the tree of the given id
-        pages = (
-            region.pages.filter(tree_id=tree_id)
-            .prefetch_major_translations()
-            .prefetch_related("mirroring_pages")
-            .cache_tree(archived=False)
-        )
+        pages = pages_by_id[tree_id]
         # The first element must be the root node
         parent = pages[0]
         # The remaining pages are the descendants
         children = pages[1:]
-        return render(
-            request,
-            self.template,
-            {
-                **self.get_context_data(**kwargs),
-                "pages": children,
-                "language": language,
-                "languages": region.active_languages,
-                "parent_id": parent.id,
-            },
+        sub_trees.append(
+            render_to_string(
+                "pages/_page_tree_children.html",
+                {
+                    **PageContextMixin().get_context_data(),
+                    "backend_language": backend_language,
+                    "pages": children,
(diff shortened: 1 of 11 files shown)


david-venhoff:
The latest commit should improve the performance of restoring the page tree state a bit. In order to remove some sql queries I moved some processing from the databse into python, but that could be reverted if it turns out that it does not perform great on the server.
I also noticed another performance issue on the client. It seems like the browser spends a large chunk of time parsing the svg strings created by `feather.replace();` calls. These could probably be cached but it seems like feather-icons is not accepting pr's at the moment.

timoludwig:
> But it seems like feather-icons is not accepting pr's at the moment.

Thanks! I did not yet have the time to look into it in detail, but since we migrated to [lucide](https://lucide.dev/) for our preact icons, we could also do the same for the regular ones, maybe this gives us a bit more flexibility...

ulliholtgrave:
Lucide also runs with a treeshake-approach to reduce the replacement time (e.g. see https://github.com/lucide-icons/lucide/tree/main/packages/lucide#lucide). Maybe this could also help us improve the general performance of our CMS by defining our used icons on each view.

timoludwig (changes requested):
Unfortunately, we still need to improve the performance of this feature before it can be merged :neutral_face:

david-venhoff:
I added a small change which first load the trees that should get re-expanded and then the rest. This seems to improve performance quite a lot, since usually only a few sub-trees are expanded.

integreat_cms/cms/views/pages/partial_page_tree_view.py
```diff
@@ -1,61 +1,74 @@
...
-        language = region.get_language_or_404(
-            kwargs.get("language_slug"), only_active=True
-        )
+    backend_language = Language.objects.filter(slug=get_language()).first()
+
+    all_pages = (
+        region.pages.filter(tree_id__in=requested_tree_ids)
+        .prefetch_major_public_translations()
+        .prefetch_related("mirroring_pages")
+        .cache_tree(False)
```
timoludwig, line 44:
```suggestion
        .cache_tree(archived=False)
```
I'd write the kwarg explicitly here to make it more obvious what is happening.

integreat_cms/cms/views/pages/partial_page_tree_view.py
```diff
@@ -1,61 +1,74 @@
...
+    )
+
+    pages_by_id = defaultdict(list)
+    for page in all_pages:
+        pages_by_id[page.tree_id].append(page)
```
timoludwig, line 49:
```suggestion
    pages_by_id = {page.tree_id: page for page in all_pages}
```
Using a `defaultdict` only makes sense when it actually prevents errors - in this case however it does only delay it.
`pages = pages_by_id[tree_id]` does not throw a `KeyError`, but the next statement `parent = pages[0]` does if the default dict returns an empty list.
So probably it's best to go with boring old dictionaries here.

integreat_cms/cms/views/pages/partial_page_tree_view.py
```diff
@@ -1,61 +1,74 @@
...
+
+    sub_trees = []
+    for tree_id in requested_tree_ids:
         # Get the tree of the given id
```
timoludwig, line 53:
```suggestion
        # Skip page trees which do not exist
        if tree_id not in pages_by_id:
            continue
        # Get the tree of the given id
```

integreat_cms/static/src/js/pages/fetch-subpages.ts
```diff
@@ -3,42 +3,64 @@
...
-    })
+
+    let expandedRows: Set<number> = new Set(JSON.parse(window.sessionStorage.getItem("page_state")));
+    const openPages = rootPages.filter(page => expandedRows.has(Number(page.dataset.pageId)));
+    const closedPages = rootPages.filter(page => !expandedRows.has(Number(page.dataset.pageId)));
+
+    await fetchAllSubpages(url, openPages);
+    restorePageTreeLayout();
+
+    await fetchAllSubpages(url, closedPages);
```
timoludwig, line 29: Very cool idea :ok_hand:

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ UNRELEASED
 * [ [#1526](https://github.com/digitalfabrik/integreat-cms/issues/1526)] Fix sending push notifications in one language
 * [ [#1630](https://github.com/digitalfabrik/integreat-cms/issues/1630)] Fix not recognized sent status of push notifications
 * [ [#1683](https://github.com/digitalfabrik/integreat-cms/issues/1683) ] Improve organization management
+* [ [#1245](https://github.com/digitalfabrik/integreat-cms/issues/1245) ] Restore expanded state of page tree
```
timoludwig, line 21:
```suggestion
```
This should now reside in the unreleased section

timoludwig (changes requested):
Amazing, thank you so much for your efforts to keep improving this solution and constantly resolving all conflicts! :pray:
It works like a charm and is very fast, so cool! :rocket:

integreat_cms/static/src/js/pages/fetch-subpages.ts
david-venhoff: Thanks!

integreat_cms/cms/views/pages/partial_page_tree_view.py
david-venhoff: The idea behind this code is that `all_pages` contains all pages that are requested, but they still need to get grouped according to their trees, so that they can be sent to the client. I think this suggestion changes the behavior: previously a page tree could contain a list of pages, but with this change only a single page.

timoludwig: Ahh sorry! So the defaultdict should not prevent the key error, but make it easier to directly append the pages instead of checking whether the `tree_id` is already contained in `pages_by_id`? Yes makes sense! :+1:

integreat_cms/static/src/js/pages/fetch-subpages.ts
timoludwig: Ok, so the only issue I see with it is that it takes longer than before when no page was expanded or all pages were. Do you think it would be feasible to keep the previous solution of fetching each tree individually and just perform them in two stages?
E.g. like this?
```
await Promise.all(openPages.map(fetchSubpages))
restorePageTreeLayout();
await Promise.all(closedPages.map(fetchSubpages))
```

david-venhoff: So if I understand correctly, do you want to go back to sending multiple requests at once? This does not seem to be faster for me, at least locally

david-venhoff:
I am not sure about whether our redis cache will still work now that the requests use the `POST` method, so this might be something to change

timoludwig:
> I am not sure about whether our redis cache will still work now that the requests use the `POST` method, so this might be something to change

Do we even have request-level caching enabled? I'm not sure...
And the database-level cache should work independently from the request method, right?

integreat_cms/static/src/js/pages/fetch-subpages.ts
svenseeberg: I think this depends a lot. Our server definitely has multiple processes running, which is a difference to local test environments. I think it is very valid to send multiple requests at once.

timoludwig: I set up a multi-process-environment locally and checked whether the alternative fetching approach was really faster, but it wasn't. Could be my specific setup, but I don't think it's worth the effort testing this on the real server, probably the current solution is fine. Feeling a little bit ashamed that my objection caused this PR to be open for nearly a year :see_no_evil: :crying_cat_face:

timoludwig (approved):
I'm deeply sorry it took me so long to revisit this :disappointed_relieved:

I finally had the time to test my alternative suggestion, but it did not increase performance. So all that was left was to resolve the conflicts.

Thanks again for this awesome work, really cool feature :pray: :heart_eyes: 

wait

ulliholtgrave (approved):
Thanks, This really is a cool feature and it works very smoothly on  my machine. I am really happy that we can finally release this :+1: :tada:

Issue #1380, opened 2022-05-02
Non-ascii chars in region slug breaks region list


timoludwig:
### Describe the Bug

Regions can be created with non-ascii characters in the slug, but the view "edit_region" only supports `-a-zA-Z0-9_`:
```
django.urls.exceptions.NoReverseMatch: Reverse for 'edit_region' with keyword arguments '{'slug': 'münchen1'}' not found. 1 pattern(s) tried: ['regions/(?P<slug>[-a-zA-Z0-9_]+)/edit/\\Z']
```

### Steps to Reproduce

1. Create region with umlaut in slug
4. Try to open region list

### Expected Behavior

Since we allow unicode characters in the database, we should also allow unicode characters in the url pattern

### Actual Behavior

Unicode characters don't work

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
May 02 17:50:39 ERROR   django.request - Internal Server Error: /regions/
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/views/regions/region_list_view.py", line 84, in post
    return self.get(request, *args, **kwargs, search_data=request.POST)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/views/regions/region_list_view.py", line 58, in get
    return render(
[...]

PR #1381, opened 2022-05-03
Use str path converter for unicode slugs

timoludwig:
### Short description

The slug path converter only matches strings consisting of ASCII letters or numbers, plus the hyphen and underscore characters.
The default str path converter matches any non-empty string, excluding the path separator, '/'.

### Proposed changes

- Use str path converter for unicode slugs

### Resolved issues

Fixes: #1380


ulliholtgrave (approved):
Thanks, works accordingly :+1:

timoludwig:
@ulliholtgrave never mind, I reverted my changes, let's just adapt the subdomains to redirect to the canonical login url.

Issue #1384, opened 2022-05-04
Phone numbers and emails addresses marked as invalid links


timoludwig:
### Describe the Bug

In the broken link checker, phone numbers and email addresses are marked as invalid links.
Although the status says "Phone number (not automatically checked)", it still creates the impression that something is wrong with these links.

### Steps to Reproduce

1. Add phone number / email links in the content
2. See broken link checker

### Expected Behavior

The phone numbers and emails should not be listed as broken links

### Actual Behavior

they are

PR #1385, opened 2022-05-04
Remove emails and phone numbers from link list and invalidate cache after bulk option

ulliholtgrave:
### Short description

This PR removes unwanted Link elements from the LinkList and invalidates the cache after bulk actions are performed.

### Resolved issues

Fixes: #1384


integreat_cms/cms/utils/filter_links.py
```diff
@@ -47,6 +47,15 @@ def filter_links(region_slug, link_filter=None):
...
         for link in translation.links.all()
     ]
+    # Define found link types that are not supposed to be shown in the List (e.g. phone numbers and Emails)
+    ignored_link_types = [
+        "Phone number (not automatically checked)",
+        "Email link (not automatically checked)",
+    ]
+    links = [
+        link for link in links if link["link"].url.message not in ignored_link_types
+    ]
```
timoludwig, line 57: I wouldn't rely on the hardcoded messages here since they might change in the future. Instead, we can also query for the type.
Also, we can include this filter in the previous loop, so we don't need to loop over all link twice.
I cannot make an automated suggestion, because it's out of range, but this would be my preferred solution:
```
    # Define found link types that are not supposed to be shown in the List (e.g. phone numbers and Emails)
    ignored_link_types = [
        "mailto",
        "phone",
        "anchor",
    ]
    # Convert translations to list of dicts to enable filtering of links in python
    links = [
        {"translation": translation, "link": link}
        for translation in translations
        for link in translation.links.all()
        if link.url.type not in ignored_link_types
    ]
```

timoludwig (commented):
Thanks! :+1:
Works as expected, I just have an idea how to improve it.

ulliholtgrave:
@timoludwig You suggestions make a lot of sense. I added them :+1:

Issue #1311, opened 2022-03-21
`last_updated` field of translations incorrect when cloning regions


timoludwig:
### Describe the Bug

When cloning regions, the `last_updated` field of translations is incorrectly marked as the time when the region was cloned.
This is problematic because it destroys the editing history and translation status of interdependent translations.

### Steps to Reproduce

1. Clone region
2. Go to page tree
3. See incorrect dates

### Expected Behavior

The dates should be preserved

### Actual Behavior

The dates are set to the current time

PR #1386, opened 2022-05-04
Fix last_updated field when cloning regions

melegiul:
### Short description

Preserves the last_updated fields for page translations after cloning regions

### Proposed changes

Overrides the predefined model method `save` for instances of `AbstractContentTranslation`. The approach is similar to the current implementation, `auto_now` uses the fields `pre_save` method to override `last_updated` [https://docs.djangoproject.com/en/3.2/ref/models/instances/#what-happens-when-you-save](https://docs.djangoproject.com/en/3.2/ref/models/instances/#what-happens-when-you-save)
Only cloned page translations preserve their last_updated field, not language tree nodes

### Resolved issues

Fixes: #1311


timoludwig (commented):
Just found a tiny typo... and could you add a new check to the region duplication test to make sure we don't break anything in the future?
E.g. by adding the line
```
assert source_page_translation.last_updated == target_page_translation.last_updated
```
after [this line](https://github.com/digitalfabrik/integreat-cms/blob/d52ca0de5ac5c543df74d464ea64e7f29f4a8dce/tests/cms/test_duplicate_regions.py#L67)?

melegiul:
> Thanks a lot, works as expected! +1 Just found a tiny typo... and could you add a new check to the region duplication test to make sure we don't break anything in the future? E.g. by adding the line
> 
> ```
> assert source_page_translation.last_updated == target_page_translation.last_updated
> ```
> 
> after [this line](https://github.com/digitalfabrik/integreat-cms/blob/d52ca0de5ac5c543df74d464ea64e7f29f4a8dce/tests/cms/test_duplicate_regions.py#L67)?

Thanks a lot for your review :smile: 
Are you sure you want me to add this line?
On my machine the tests seem to fail also with the current dict compare test:

```
assert source_page_translation_dict == target_page_translation_dict
E               AssertionError: assert {'content': '...uage': 1, ...} == {'content': '...uage': 1, ...}
E                 Omitting 9 identical items, use -vv to show
E                 Differing items:
E                 {'last_updated': datetime.datetime(2019, 8, 12, 7, 57, 42, 456000, tzinfo=<UTC>)} != {'last_updated': datetime.datetime(2022, 5, 6, 8, 27, 25, 126266, tzinfo=<UTC>)}
```
Earlier tests falsely succeeded, because `model_to_dict` only returns editable fields and `last_updated` with the `auto_now` option was non editable https://github.com/django/django/blob/21d8ea4eb3e22d458515b5278c8cd3a15b069799/django/forms/models.py#L108.

timoludwig (approved):
> Are you sure you want me to add this line? On my machine the tests seem to fail also with the current dict compare test:
> 
> Earlier tests falsely succeeded, because `model_to_dict` only returns editable fields and `last_updated` with the `auto_now` option was non editable https://github.com/django/django/blob/21d8ea4eb3e22d458515b5278c8cd3a15b069799/django/forms/models.py#L108.

Oh, you're right, I also looked into the `model_to_dict()` function and wondered why our tests were succeeding before but didn't realize the field has become editable by your change and thus is now contained in the test. Alright, then let's leave it like this, even better :+1:

PR #1389, opened 2022-05-05
Change media library upload path

timoludwig:
### Short description

Change media library upload path to keep the directory structure cleaner and avoid overlaps of region ids with the date directories

### Proposed changes

- Use `regions/` instead of `sites/` for region media library
- Use `global/` for global media library


timoludwig:
Oh, yes... existing files will continue to work without problems, it's just that we probably want to have a consistent directory structure on the live system...
However, even without merging this PR, our directory structure is very dirty on Malte, because the migration files from wordpress put the files directly into the `media` directory while newly uploaded files went into the `media/sites` directory...
So maybe we can fix these problems together manually... or do you think we should create a Django migration script for this?

timoludwig:
@ulliholtgrave Alright, I added a migration file to handle the renaming of the existing files.
I couldn't get the OSSAR scan to work on this branch, no matter what I did, so I temporarily disabled it.

ulliholtgrave (approved):
Very nice :+1: This should remove the double structure and result in a cleaner dealing with our media files :)

ulliholtgrave:
Thanks for the last editing. I guess this PR is ready to merge :+1: :)

Issue #1371, opened 2022-04-25
Fallback texts for pages


svenseeberg:
### Motivation
Wordpress had a function to duplicate from one language into another. We do not want mixed content anymore and we are not going to migrate content of duplicated pages to Django. However, we want to show users a link to existing languages/translations. We already have a similar functionality for live content (see #1331).

### Proposed Solution
Deliver a fallback text with links to existing translations in the current language, if the page title exists but the content is empty. We need translations of this text in all relevant languages. Translating the sentences with Google Translate should be sufficient for now.

### Additional Context

https://issues.tuerantuer.org/browse/WP-65

PR #1393, opened 2022-05-06
Fallback text for empty pages

david-venhoff:
### Short description

This pr adds support for showing a fallback text for empty pages which links to other non-empty translations of the same page.

### Proposed changes

- Check whether a page translation is empty when generating an api response
- If it is empty, show a message which provides some alternatives
- Reuse the existing translations for mirrored pages for this message

### Resolved issues

Fixes: #1371


svenseeberg (approved):
Thanks a lot! There are some unnecessary changes in this PR with changed file endings of JSON files. That could be changed. Beyond that I cannot see any issues. Lets hope that a rebase on develop fixes the OSSAR problems?

PR #1396, opened 2022-05-07
Update GitHub actions

timoludwig:
### Short description

Since we faced a few problems with the OSSAR action in PRs #1393 and #1389, I played around with the github actions a bit, but couldn't find the issue...
Does anybody have an idea how to fix it? If not, I'd just disable it temporarily, until they fix their upstream problem...
Nonetheless, I think we could merge this changes to use the newest versions of the actions and add a new dependency check...

### Proposed changes

- Update CodeQL and OSSAR
- Add Dependency review

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -2,12 +2,11 @@ name: "CodeQL"
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


ulliholtgrave (approved):
Looks good, by have also no idea why this check isn't working anymore :thinking:

PR #1397, opened 2022-05-08
Do not run dependency review on push event

timoludwig:
### Short description

Dependency review apparently does not support the push event and just works on pull requests...


ulliholtgrave (approved):
Lgtm :+1:

PR #1417, opened 2022-05-11
Don't show page fallback text if there is no alternative translation

david-venhoff:
### Short description

This pr prevents the fallback text for pages to be shown when the alternatives linklist would be empty.
This should fix a bug where the fallback text was shown for empty pages which are only used for the tree structure.


svenseeberg:
@david-venhoff thanks a lot! Would you be willing to add a test case which returns the fall back test? That would probably catch most related problems in the future.

david-venhoff:
I've added a test case now (The fallback text is at line 426)

svenseeberg:
@timoludwig should we add an entry to the changelog?

timoludwig:
> @timoludwig should we add an entry to the changelog?

Yes, sounds good :+1:

svenseeberg:
@david-venhoff you cannot directly link to this PR, that will mess with the build pipeline. But `https://github.com/digitalfabrik/integreat-cms/issues/1417` does work as well for PRs.

Issue #1410, opened 2022-05-10
HTML entities not stripped in excerpt


timoludwig:
### Describe the Bug

The `excerpt` field in the API should return parts of the content without HTML entities, but at the moment it's just a copy of the content

### Steps to Reproduce

1. Retrieve API

### Expected Behavior

The `excerpt` field should not contain HTML

### Actual Behavior

The `excerpt` field contains HTML

### Additional Information

https://webnext.integreat.app/testumgebung/de/events

PR #1418, opened 2022-05-11
Strip HTML in excerpt field and adjust test data

ulliholtgrave:
### Short description

This PR adds a functionality to strip the HTML entities in the excerpt field and adjust the test data accordingly.

### Resolved issues

Fixes: #1410


timoludwig (commented):
Is there a reason you didn't use the inbuilt Django functionality?
```
from django.utils.html import strip_tags

excerpt": strip_tags(page_translation.content),
```
I feel like a custom html parser is a bit overkill for this task...
Apart from that: could you add a note to the changelog?

ulliholtgrave:
@timoludwig You are right, i just missed this inbuild method and changed it accordingly. I also added a Changelog entry and fixed all the test cases (which was a hell of a job :see_no_evil: )

Issue #1404, opened 2022-05-09
Fix performance issue for "select all" in big page trees


ulliholtgrave:
### Describe the Bug

In very large page trees (e.g. Munich) clicking on the select all checkbox results in a browser crash.

### Additional Information

This is probably a result of the used recursion in the implementation. This could be fixed by just using a QuerySelectorAll as an event and remove the current implementation.

PR #1420, opened 2022-05-13
Fix performance issues for "select all"

ulliholtgrave:
### Short description

Restructure handling of updating the counter. This leads to some massive increases in performance.

### Resolved issues

Fixes: #1404


ulliholtgrave:
@timoludwig The missing Changelog-Entry is on purpose as I am not sure, if maybe another iteration is required :)

timoludwig (approved):
Awesome, thanks! :+1:

Issue #1292, opened 2022-03-11
Allow uploading multiple files at once in media library


timoludwig:
### Motivation

At the moment, it is very tedious to upload a large set of files one by one.

### Proposed Solution

- Make it possible to select and upload multiple files at once
- Optional: Allow uploading a zip file which is then extracted into single files in the media library

PR #1421, opened 2022-05-13
Add drag and drop for media file upload and enable multiple files

ulliholtgrave:
### Short description

This pr substitutes our current file upload form with a drag and drop area. This also enables to upload multiple files due through an internal queue.

### Proposed changes

- Add Dropzone library
- Use it in the media library

### Resolved issues

Fixes: #1292

### DIFF
--- a/integreat_cms/static/src/js/media-management/component/upload-file.tsx
+++ b/integreat_cms/static/src/js/media-management/component/upload-file.tsx
@@ -1,11 +1,13 @@
 /*
  * This component renders a file upload field
  */
-import { StateUpdater, useState } from "preact/hooks";
-import { FilePlus, FileText, Upload, Image } from "preact-feather";
-import cn from "classnames";
+import { StateUpdater, useEffect, useRef, useState } from "preact/hooks";
+import { FilePlus } from "preact-feather";
+import Dropzone, { DropzoneFile } from "dropzone";
+import "dropzone/dist/dropzone.css";
 
 import { Directory, MediaApiPaths } from "../index";
+import { getCsrfToken } from "../../utils/csrf-token";
 
 interface Props {
   directory: Directory;
@@ -15,6 +17,7 @@ interface Props {
   mediaTranslations: any;
   submitForm: (event: Event, successCallback: () => void) => any;
   isLoading: boolean;
+  refreshState: [boolean, StateUpdater<boolean>];
 }
 export default function UploadFile({
   directory,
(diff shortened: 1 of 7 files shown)


integreat_cms/static/src/js/media-management/component/upload-file.tsx
```diff
@@ -24,73 +28,46 @@ export default function UploadFile({
   mediaTranslations,
   submitForm,
   isLoading,
+  refreshListing,
 }: Props) {
   // This state is a buffer which contains the currently selected file (the browser type File)
   const [selectedFile, setSelectedFile] = useState<File | null>(null);
```
timoludwig, line 34:
```suggestion
```
This buffer is no longer required, right?

integreat_cms/static/src/js/media-management/component/upload-file.tsx
```diff
@@ -24,73 +28,46 @@ export default function UploadFile({
...
+      headers: { "X-CSRFToken": getCsrfToken() },
+      dictDefaultMessage: mediaTranslations.text_upload_area,
+      acceptedFiles: allowedMediaTypes,
+    });
+    dropZone.on("queuecomplete", () => {
+      refreshListing();
+      dropZone.removeAllFiles(true);
+    });
+    dropZone.on("complete", () => {
+      refreshListing();
```
timoludwig, line 49: I don't really understand why, but this does not seem to work for me.
When I upload files, only sometimes the new file is listed in the library and sometimes I need to reload the page to see the files.

timoludwig (commented):
Thanks a lot! :+1:

In general, it seems to work, however I noticed some weird behaviour I can't really deterministically reproduce, so I'm a bit confused...
When I upload many files at once, a few of the requests just never get answered... I think this might have to do with the Django development server which can't handle many requests at the same time (I also noticed similar problems on the page tree with many root pages) - but the production system seems to handle concurrent request quite reliable, so probably this is just a development problem and will be fixed on the server... Nevertheless, I'd like to have some kind of user feedback whether a request was successful, maybe you could show a message when a file has been successfully uploaded and an error when the server does not answer?
Also, I think it feels a bit "glitchy" when files are uploaded very fast, maybe it's also ok to leave the uploaded files in the dropzone for a few moments to give visual feedback to the user that the upload was successful? But not sure...

integreat_cms/static/src/js/media-management/component/upload-file.tsx
```diff
@@ -24,73 +27,48 @@ export default function UploadFile({
...
+      setTimeout(() => {
+        refreshListing();
+        dropZone.removeAllFiles(false);
+      }, 500);
+    });
+    dropZone.on("complete", (file) => {
+      setTimeout(() => {
+        dropZone.removeFile(file);
+      }, 2000);
+    });
```
timoludwig, line 52:
```suggestion
```
I'm not sure what this event really does - if the action of the `queuecomplete` event is executed first, this will do nothing, right? If already all files have been removed, there is no need to remove the individual files as well...

integreat_cms/static/src/js/media-management/component/upload-file.tsx
```diff
@@ -24,73 +27,48 @@ export default function UploadFile({
...
+      headers: { "X-CSRFToken": getCsrfToken() },
+      dictDefaultMessage: mediaTranslations.text_upload_area,
+      acceptedFiles: allowedMediaTypes,
+      parallelUploads: 1,
+    });
+    dropZone.on("queuecomplete", () => {
+      setTimeout(() => {
+        refreshListing();
+        dropZone.removeAllFiles(false);
+      }, 500);
```
timoludwig, line 46:
```suggestion
      }, 2500);
```
I still think this is a bit short...

integreat_cms/static/src/js/media-management/component/upload-file.tsx
```diff
@@ -24,73 +27,48 @@ export default function UploadFile({
...
+      url: apiEndpoints.uploadFile,
+      headers: { "X-CSRFToken": getCsrfToken() },
+      dictDefaultMessage: mediaTranslations.text_upload_area,
+      acceptedFiles: allowedMediaTypes,
+      parallelUploads: 1,
+    });
+    dropZone.on("queuecomplete", () => {
+      setTimeout(() => {
+        refreshListing();
+        dropZone.removeAllFiles(false);
```
timoludwig, line 45: Could we maybe filter out all files that were not uploaded successfully? I feel like even if we show the "x" on the image for 2 seconds, it might not be enough to understand that there was an error uploading the file...

timoludwig (commented):
Thanks! :+1:

Using the timeout definitely helps, however, I still would like some kind of error handling that exceeds the possibility to see an "x" for 2 seconds.
Apart from that: Do you think there is the possibility to show the most recently uploaded file in the sidebar like before, or is that not feasible? 

Also, could you add a changelog entry?

timoludwig (approved):
Thanks a lot, looks good to me! :+1:

PR #1424, opened 2022-05-14
Invalidate translations when removing LanguageTreeNode, fixes #1403

### LINKED ISSUE #1403: Cache does not get invalidated correctly when LanguageTreeNode is deleted
### Describe the Bug

If you delete a language the page tree still shows the translation status of the corresponding language. This is probably a caching issue.

svenseeberg:
This change flushes all content models after removing a LanguageTreeNode. It seems the problem described in #1403 only exists for page translations and no other content models, but I added them to be safe in the future.

One caveat though: Initially I tried to flush the PageTranslation objects which seemed to make more sense to me but that did not help. I'm not 100% sure why this fixes the issue. https://github.com/Suor/django-cacheops/blob/master/README.rst does not really explain how related objects are handled.

Fixes #1403

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ UNRELEASED
 * [ [#1402](https://github.com/digitalfabrik/integreat-cms/issues/1402) ] Also duplicate imprints for new regions
 * [ [#1408](https://github.com/digitalfabrik/integreat-cms/issues/1408) ] Remove duplication of push API tokens for pages during duplication process
 * [ [#1404](https://github.com/digitalfabrik/integreat-cms/issues/1404) ] Fix performance issue for select all on huge page trees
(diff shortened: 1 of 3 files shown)


timoludwig (approved):
Thanks a lot, fixes the issue :+1:

> One caveat though: Initially I tried to flush the PageTranslation objects which seemed to make more sense to me but that did not help. I'm not 100% sure why this fixes the issue. 

I also don't understand it completely, but my take on it is: Probably invalidating the page translation objects is not enough because the translations are stored inside the cached page tree structure `page.prefetched_translations` etc. and while it usually follows foreign key relationships, probably in this case it does not know about the prefetched objects cache...

timoludwig:
If you want, you could add a changelog entry :wink:

PR #1425, opened 2022-05-14
API: slugify page slugs, fixes #1401

### LINKED ISSUE #1401: API: Wordpress slug compatibility
### Motivation
Django does not allow special characters in slugs, while WordPress does allow them. We're stripping them during the migration. It would be nice if links all over the internet and within our content to pages with special characters in their slugs do not break.

### Proposed Solution
If a slug transmitted via the API does not match a translation, run the slugify method on it and then try again to match. This is the same operation we perform on slags after migrating. Alternatively, the slugify method could be run on all slugs if the compute time is not too long?

### Alternatives
Old links break or need to be fixed manually over time.

### Additional Context
https://issues.tuerantuer.org/browse/INFRA-161

svenseeberg:
### Short description
Apply the slugify function on all API page endpoint slugs. This should make all calls to children endpoints compatible with old WordPress slugs. This should do no harm (except of some compute time), as we only have slugs in our database that were cleaned by the slugify function.

### Proposed changes
Use the [slugify](https://docs.djangoproject.com/en/4.0/ref/utils/#django.utils.text.slugify) function to clean page translation slugs in API requests.

### Resolved issues
Fixes: #1401

### DIFF
--- a/integreat_cms/api/v3/pages.py
+++ b/integreat_cms/api/v3/pages.py
@@ -9,6 +9,7 @@ from django.http import JsonResponse, Http404
 from django.shortcuts import get_object_or_404
 from django.views.decorators.csrf import csrf_exempt
 from django.utils.html import strip_tags
+from django.utils.text import slugify
 
 from ...cms.models import Page
 from ...cms.forms import PageTranslationForm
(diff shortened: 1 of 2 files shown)


svenseeberg:
... tests proved this wrong. I'll take another look ;-)

integreat_cms/api/v3/pages.py
```diff
@@ -142,7 +143,7 @@ def get_single_page(request, language_slug):
         # Strip leading and trailing slashes to avoid ambiguous urls
         url = request.GET.get("url").strip("/")
         # The last path component of the url is the page translation slug
-        page_translation_slug = url.split("/")[-1]
+        page_translation_slug = slugify(url.split("/")[-1])
```
timoludwig, line 146:
```suggestion
        page_translation_slug = slugify(url.split("/")[-1], allow_unicode=True)
```
I guess we have to allow unicode slugs here. Timing should be irrelevant in comparison to the database queries etc, right? Otherwise we could think about using the slugify method only if the regular url did not return a response...

timoludwig (approved):
LGTM :+1:

Issue #1438, opened 2022-05-17
Page-specific permissions cause error in page form


timoludwig:
### Describe the Bug

When page-specific permissions are enabled in the region form, the page form throws an error

### Steps to Reproduce

1. Enable page-specific permissions for a region
2. Go to a page form in this region
3. See error

### Expected Behavior

The page-specific permissions should be manageable when they're activated

### Actual Behavior

A `NoReverseMatch` error appears

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
render
    bit = node.render_annotated(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/defaulttags.py", line 315, in render
    return nodelist.render(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/defaulttags.py", line 315, in render
    return nodelist.render(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/loader_tags.py", line 195, in render
    return template.render(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/base.py", line 172, in render
    return self._render(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/base.py", line 162, in _render
    return self.nodelist.render(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
[...]

PR #1439, opened 2022-05-17
Fix error in page form when page-specific permissions are enabled

timoludwig:
### Short description

Fix error in page form when page-specific permissions are enabled

### Proposed changes

- Fix URL tags in template

### Resolved issues

Fixes: #1438


ulliholtgrave (approved):
Thanks :+1:

Issue #1461, opened 2022-05-21
No warning displayes when using the back button while editing a page


karthik-integreat:
### Describe the Bug

A warning is not shown, as it is in all other menu items when clicking on the back button while editing a page.

### Steps to Reproduce

1. Go to 'Seiten'
2. Click on 'Alles ausklappen'
3. Select a page and edit the page
3. Click on the back button of the browser
4. No warning message is displayed. "Are you sure do you want to leave this page"?

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

PR #1465, opened 2022-05-21
show unsaved warning when only changing tinymce editor

david-venhoff:
### Short description

This pr fixes a bug where no autosave warning was shown after editing the body of a page.
The bug is caused by tinymce not (anymore?) bubbling up the input event, which is used by the autosave mechanism to determine when to show the warning.
Additionally this pr fixes a related bug that caused `unsaved-warnings.ts` to be loaded twice.

### Proposed changes

- Manually create an input event when the editor state changes to 'dirty'
- Use a custom event to notify the unsaved-changes script that an autosave occrured

### Resolved issues

Fixes: #1461


timoludwig (approved):
Thanks a lot! :+1:

I noticed that the content is marked as "not-dirty" even when "stay on page" is clicked on the warning popup, so it does not work a second time if no changes are made in between. But I think that's a quite rare edge case and I'm not sure whether this has anything to do with the PR, probably it was this way before (even when the warning did work).

Apart from that: I guess this fix is significant enough to deserve an entry in the changelog if you want.

david-venhoff:
Weird, the warning does not show up a second time, because the page is stored as draft when the user clicks decides to not leave the page on the popup. So I think there should at least be no data loss, but I don't understand why there happens a store as draft at all.

Issue #1471, opened 2022-05-22
Statistic settings missing in region form


timoludwig:
### Describe the Bug

In the region form, the fields for statistic settings are missing
culprit: 57909d02c7044f3a6c1a3a880002efbcf385fbc8

### Expected Behavior

The region settings should be changeable in the region form

### Actual Behavior

They are not

PR #1472, opened 2022-05-22
Add statistic settings to region form again

timoludwig:
### Short description

Add statistic settings to region form again

### Resolved issues

Fixes: #1471

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 UNRELEASED
 ----------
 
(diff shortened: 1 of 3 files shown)


ulliholtgrave (approved):
I am not really sure how and why there have been stripped from the template, but your PR are adding them again. So Thanks :+1:

timoludwig:
> I am not really sure how and why there have been stripped from the template

Actually, it was my own fault, when I redesigned the region form in #1412, I initially wanted to use a separate box for statistics, but then I didn't like it and removed it, but apparently forgot to add the fields to the original box again :see_no_evil:

PR #1473, opened 2022-05-23
Adapt offers API to web app

svenseeberg:
### Short description
The web app expects a null value instead of an empty dict in the offers API for the `post` attribute.

### Proposed changes
Convert the empty dict to null when delivering the information via API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ UNRELEASED
 ----------
 
 * [ [#1471](https://github.com/digitalfabrik/integreat-cms/issues/1471) ] Add statistic settings to region form again
+* [ [#1473](https://github.com/digitalfabrik/integreat-cms/issues/1473) ] Fix offers compatibility with web app
 
 
 2022.5.1
--- a/integreat_cms/api/v3/offers.py
+++ b/integreat_cms/api/v3/offers.py
@@ -50,6 +50,8 @@ def get_post_data(offer, region):
     post_data = offer.post_data
     if offer.use_postal_code == postal_code.POST:
         post_data.update({"search-plz": region.postal_code})
+    if not post_data:
+        return None
     return post_data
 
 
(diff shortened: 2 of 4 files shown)


svenseeberg:
We could add a test before merging :D

ulliholtgrave (approved):
I guess there should be some more elegant solutions, but I guess this is an emergency :+1:

svenseeberg:
As I already fixed this in the production environment, we have the time to add a test case.

timoludwig (approved):

I guess it would be a bit cleaner to put it into the `get_post_data()` function, but that's just personal preference...
```
if not post_data:
    return None
return post_data
```

svenseeberg:
We have one additional bug: we need to replace `"search-plz":"{plz}"` with the actual zip code of each region.

svenseeberg:
> We have one additional bug: we need to replace `"search-plz":"{plz}"` with the actual zip code of each region.

Oh wait, the code is already in there.

timoludwig (approved):
:+1:

svenseeberg:
I guess we ignore the documentation build error (`intersphinx inventory 'https://docs.python-requests.org/en/master/objects.inv' not fetchable due to <class 'requests.exceptions.SSLError'>: HTTPSConnectionPool(host='docs.python-requests.org', port=443): Max retries exceeded with url: /en/master/objects.inv (Caused by SSLError(CertificateError("hostname 'docs.python-requests.org' doesn't match either of '*.ispapi.net', 'ispapi.net'")))`)?

timoludwig:
> I guess we ignore the documentation build error

Last time we had this problem, we had weeks of failing builds until the documentation of pipenv was fixed... maybe we should temporarily remove the line
https://github.com/digitalfabrik/integreat-cms/blob/1b898ac34655530c734ef90ac08965ea45f1ac1d/sphinx/conf.py#L79
and add the few requests-references to the [nitpicky-ignore](https://github.com/digitalfabrik/integreat-cms/blob/1b898ac34655530c734ef90ac08965ea45f1ac1d/sphinx/conf.py#L129) setting?
But I don't know, could also only take hours until the certificate of the docs is fixed...

svenseeberg:
It seems it was fixed while I added a line to the changelog ;-)

PR #1480, opened 2022-05-23
Improve list paddings

timoludwig:
### Short description

Fix a few paddings and text distributions


ulliholtgrave (approved):
Thanks, LGTM. I guess we could also consider to use the same overflow techniques as in the page tree, but right now this stream lines our UI :+1:

Issue #1470, opened 2022-05-22
Change flags of Farsi and Arabic


dkehne:
After we had the discussion of using non-political flags (which cannot be assigned to one state) already in 2016 i would recommend to use the same flags for FA and AR as in WordPress.

Media files are downloadable there.

timoludwig:
You mean this one, right?
![ar](https://user-images.githubusercontent.com/16021269/169854377-8a68684b-1d0e-4a0a-ab4d-9d6a17b1d1ec.png)
Do you know where we would find it as svg?

PR #1482, opened 2022-05-23
Add non-political flags for Arabic and Farsi

timoludwig:
### Short description

Add non-political flags for Arabic and Farsi

### Proposed changes

- Update dependencies
- Add new choices to country selector

### Resolved issues

Fixes: #1470


ulliholtgrave (commented):
grafik
Are these really the same flags that we are using in the old system? They seem a little bit de-placed in my opinion :thinking:

timoludwig:
> Are these really the same flags that we are using in the old system? They seem a little bit de-placed in my opinion thinking

No, the old flags were only available as png and very poor quality:

fa
ar

so I tried to redesign them as svgs:

fa
ab


On my system, they look a bit different, but your screenshot looks strange, I agre. Probably I should increase the spacing between the letters and the border...

ulliholtgrave:
Yes, that would probably help :)

ulliholtgrave:
I don't know if this is a server thing, but even in Github your SVGs are looking the same as on the dev server :D
grafik

svenseeberg:
Do we know what these flags stand for? AFAICT they are just characters from their corresponding alphabet, right?

timoludwig:
> Do we know what these flags stand for? AFAICT they are just characters from their corresponding alphabet, right?

Yes, indeed. There is no deeper meaning.
Do you have a better idea for the representation? Using these would have the advantage that the users recognize them from WordPress.

I have yet to improve the svgs to also have proper alignment on ullis machine...

svenseeberg:
> Do you have a better idea for the representation?

No, I just wanted to make sure I understand what is going on :D

timoludwig:
Ok, I think I found my problem: I defined the letters as [actual text in the font "Courier New"](https://github.com/timoludwig/flagpack-dart-sass/blob/a79294cdf7ff034063535b4498a21419e0388da9/flags/4x3/ab.svg?short_path=6eb8809#L114).
On systems which don't have this font installed, it looks a bit weird.
I now updated the flags to be normal paths which should look the same on all systems:

ab
fa

ulliholtgrave (approved):
Thanks, Works as expected and we can finally close this PR :tada:

Issue #1416, opened 2022-05-11
Not able to save staff users in region user form


timoludwig:
### Describe the Bug

At the moment, an error occurs when a staff user is changed via the region user form

### Steps to Reproduce

1. Create a staff user with a preferred region
2. Edit this user from the user list of that region
3. Select a non-staff role (otherwise the form cannot be submitted)

### Expected Behavior

It should be possible to save the user without changing its role.
We could also think about filtering this list and not showing staff users in the region user list.

### Actual Behavior

An error occurs

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
Internal Server Error: /augsburg/users/1/edit/
Traceback (most recent call last):
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.9/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.9/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/home/timo/job/integreat/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/home/timo/job/integreat/integreat-cms/integreat_cms/cms/views/users/region_user_form_view.py", line 98, in post
    region_user_form.save()
[...]

PR #1483, opened 2022-05-24
Improve user list

timoludwig:
### Short description

Improve user list

### Proposed changes

- Hide staff users from region user list
- Add filters to admin user list

### Resolved issues

Fixes: #1416


ulliholtgrave (approved):

Issue #1460, opened 2022-05-21
Hide Status in the Broken Link checker


karthik-integreat:
### Motivation

The Status under the broken link checker is confusing. Sometimes it just says "Fehler", sometimes it says "Forbidden". It is probably better to hide the status in the broken link checker.

### Proposed Solution

Hide the Status in the broken link checker

dkehne:
Only the "Status" coloumn should be moved into expert mode. If this is not possible, just remove the Status for all users.
It is not understable without technical knowledge.

dkehne:
@maxammann had some ideas which status we might ignore - e.g. if anchor is wrong.
currently the broken link checker might be to sensitive of W3C-conform-problems

dkehne:
we just want to have broken links that lead to the fact that users are not able to access to content behind the link. all other technical problems are no real broken links for users.

maxammann:
I think should differentiate between:

* Temporary Error (Gateway Timeout, Service Unavailable)
* Permanent Error (404)
* Warnings (someone reported that there are "broken links" which are not really broken

PR #1484, opened 2022-05-24
Only show status in broken link checker for expert users

timoludwig:
### Short description

Only show status in broken link checker for expert users

### Resolved issues

Fixes: #1460


ulliholtgrave (approved):

Issue #742, opened 2021-03-16
Add map bounding box to region model


svenseeberg:
### Motivation
We want to implement maps in the Integreat app. The app function needs to know which map tiles it needs to download. For this we need to provide two bounding box coordinates.

### Proposed Solution
Add a property to region model and output in the APIv3.

The property should be named `bounding_box`. Values are nested arrays consisting of two numbers: `[[32.958984, -5.353521],[43.50585, 5.615985]]`

Reference in WordPress CMS: https://issues.tuerantuer.org/browse/WP-42

svenseeberg:
We do not need to implement this before the migration, if we do not have the time. The maps feature will not be released in may. Please address other issues first.

PR #1485, opened 2022-05-24
Add default bounding box to region API

timoludwig:
### Short description

Add default bounding box to region API

### Resolved issues

Part of: #742


ulliholtgrave (approved):
LGTM :+1:

Issue #1478, opened 2022-05-23
Page with archived sibling cannot be saved


timoludwig:
### Describe the Bug

A page with an archived ref node cannot be saved in the page form

### Steps to Reproduce

Not completely sure how to reproduce this. The tree library has its own "algorithm" to determine the reference node of each node - this is the left neighbour most of the time I think. When this page is pre-selected in the ref-node-field, but it's missing from the form choices (because it's archived), the form throws an error.

### Expected Behavior

The page should be saveable no matter whether a sibling is archived

### Actual Behavior

The ref node field is not recognized as valid choice

PR #1488, opened 2022-05-24
Fix bug where page with archived sibling cannot be saved

timoludwig:
### Short description

Fix bug where page with archived sibling cannot be saved.
This should only make the page form a bit slower, but as a quick fix I think it's sufficient.

### Proposed changes

- Allow archived pages as ref nodes

### Resolved issues

Fixes: #1478


ulliholtgrave (approved):
I tested the case with the archived page locally and this seem to fix the issue with the migrated data :+1:

PR #1490, opened 2022-05-24
Use legacy format for bounding box

timobrembeck:
### Short description

It seems that the app needs a new release to support the canonical flat list for bounding boxes, so in order to roll out our own release sooner, let's use the legacy nested list format for now.
(for reference see https://github.com/digitalfabrik/integreat-app/pull/675)


ulliholtgrave (approved):
Looks reasonable :+1: Can you update the expected test output before you merge it? :)

Issue #1454, opened 2022-05-21
Recurring events should be standard-functionality (instead of expert mode)


dkehne:
We have discussed internally that the recurring events should be in the standard-mode not in the expert mode. Should be done until end of June.

PR #1493, opened 2022-05-24
Enable recurring events for non-expert users

timoludwig:
### Short description

Enable recurring events for non-expert users

### Proposed changes

- Show recurrence input fields for all users, not only expert users
- Update help text for the expert user field

### Resolved issues

Fixes: #1454


ulliholtgrave (approved):

Issue #1503, opened 2022-05-27
Expanding feedback not working


timoludwig:
### Describe the Bug

The expanding button for feedback does not work at the moment

### Steps to Reproduce

1. Submit feedback that exceeds the length of 20 words
2. Go to feedback list and try to expand the truncated text

### Expected Behavior

The text should be expanded

### Actual Behavior

Nothing happens

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

PR #1506, opened 2022-05-27
Fix feedback comments not being expandable

david-venhoff:
### Short description

This pr fixes a bug where feedback comments were not expandable in some cases. This bug existed because the responsible event handlers were (depending on page load order) set up on the `<i>`  elements, which got later replaced by `<svg>` elements.

### Proposed changes

- Make sure that the image elements are already converted when setting up the event handlers

### Resolved issues

Fixes: #1503 

I also noticed that the feedback text only gets collapsed when it has more than one line or when it has more than 20 words.
I don't know if this is a problem in practice, but a text with few but long words would break the layout:
grafik


integreat_cms/static/src/js/feedback/toggle-feedback-comments.ts
```diff
@@ -1,7 +1,10 @@
 /*
  * This file provides a simple spoiler for feedback comments with hide/show functionality.
  */
+import feather from "feather-icons";
+
 window.addEventListener('load', () => {
+    feather.replace({ class: 'inline-block' });
```
timoludwig, line 7: Since the icons are also replaced in the index.ts, this will cause the icons to be replaced exactly twice on load:
https://github.com/digitalfabrik/integreat-cms/blob/25b2b7b54d942fa3f58fa7fc5ef6ee8387d45e34/integreat_cms/static/src/index.ts#L74-L76
Maybe it would be better to reduce this, e.g. by exporting the feedback event handler function and call it in index.ts after the icons have been replaced or set the event handler on the parent `<span>` element?

timoludwig (commented):
Thanks a lot! :+1:

This fixes the issue, however I think there might be a more efficient solution...
Also, could you add a note to the changelog?

> I also noticed that the feedback text only gets collapsed when it has more than one line or when it has more than 20 words. I don't know if this is a problem in practice, but a text with few but long words would break the layout

Good point, maybe it makes sense to include this block as last `elif` case before printing the whole comment?
```
{% elif feedback.comment|length > 120 %}
    <span>
        {{ feedback.comment|truncatechars:120 }} <i data-feather="chevron-down"
        class="toggle-feedback-comment transform cursor-pointer hover:scale-125"></i>
    </span>
    <span class="hidden">
        {{ feedback.comment }} <i data-feather="chevron-up"
        class="toggle-feedback-comment transform cursor-pointer hover:scale-125"></i>
    </span>
{% else %}
    {{ feedback.comment }}
{% endif %}
```

david-venhoff:
I have now removed the call to `feather.replace` and added your suggestions for truncating too long feedback. On my screen the limit of 120 character was quite high, so I went for 60 instead.

timoludwig:
@david-venhoff I think my initial solution with truncating the feedback in the django template doesn't work well for all edge cases (e.g. comments with multiple lines that contain very long words or less than 20 words but very long ones etc.), so I tried to do the truncation in css + js...
The solution was a bit too complex for regular GitHub-suggestions, so I directly pushed a commit. Could you have a look at it and check if everything works for you and whether you're ok with my changes? Thanks a lot!

david-venhoff:
Looks good to me. :+1:

timoludwig (approved):
:tada:

Issue #1514, opened 2022-05-31
Incorrrect format of region aliases in API


timoludwig:
### Describe the Bug

The region aliases should be returned as JSON objects, not as JSON string.

### Steps to Reproduce

Go to https://admin.integreat-app.de/api/regions/

### Expected Behavior

The region aliases should be objects

### Actual Behavior

The region aliases are a JSON string

steffenkleinle:
Thanks :tada:

PR #1515, opened 2022-05-31
Fix format of region aliases in API

timoludwig:
### Short description

Fix format of region aliases in API

### Proposed changes

Use `JSONField` instead of `TextField`, this automatically fixes the formatting

### Resolved issues

Fixes: #1514


ulliholtgrave (approved):
Tested it locally and seems to work as expected :+1:

PR #1522, opened 2022-06-01
Fix documentation build

timobrembeck:
### Short description

Apparently, Django changed something in its documentation, which breaks our intersphinx mappings.

### Proposed changes

- Change link to xml serializer


ulliholtgrave (approved):
Ohh magic Timo, as you are on github. Fix all our issues and prevents us from bugs :P :+1:

Issue #1502, opened 2022-05-27
Hide links on archived pages in broken link checker


timoludwig:
### Motivation

When broken links are on archived pages, they are less meaningful and can be ignored.

### Proposed Solution

Remove links on archived pages from the broken link checker. We have to evaluate which of the following is more reasonable and efficient:
1. Leave the link objects in the database but filter them in the database query and do not show them in the list
2. Whenever a page is archived, delete all corresponding link objects. On restoring, we have to call save() on the page translations as well to make sure the links are being created again.

osmers:
If the effort is low, then I would consider this high priority bcs we have explicitly told the Kommunen that they should have very few broken links. As long as broken links on archived pages are shown, they have no overview of how many links really are in need of fixing!

osmers:
Not sure, but I think we need to open this up again. In Dachau for example I found some links on archived pages that show up: https://admin.integreat-app.de/stadtdachau/analytics/linkcheck/invalid/
See Quelle interpreter/translator

timoludwig:
That's unfortunate :cry:
We have to do further research to tell whether this is actually a problem of the code or some kind of follow-up problem caused by our inconsistent production database.

timoludwig:
I could verify that this is directly happening when scanning the links.
I ran the link scanning on the production system again and afterwards, many links were found in archived pages:
```
In [1]: from linkcheck.models import *

In [2]: Link.objects.filter(page_translation__page__explicitly_archived=True).count()
Out[3]: 25554
```
So something with this filter seems to be wrong:
https://github.com/digitalfabrik/integreat-cms/blob/a0b389f3d28f8c66ba152ec52246cd244bf93433/integreat_cms/cms/linklists.py#L50-L53

timoludwig:
Ok, I found out the problem during the findlinks command on the prod system: We use a [wrapper around the cli](https://git.tuerantuer.org/DF/salt/src/commit/3c454b47af43c656789bf69346b2326252f2fe97/states/integreat-cms/files/integreat-cms-cli) which did not pass the environment variable `LINKCHECK_EXCLUDE_ARCHIVED_PAGES` to the actual management command.
So the rest of this task remains an infrastructure ticket.

osmers:
@timoludwig somehow broken links in archived pages and past events are showing up again: https://admin.integreat-app.de/kreishoexter/analytics/linkcheck/ignored/
We put them on ignore, but can you have a look nonetheless?

[...]

PR #1524, opened 2022-06-05
Do not check links in archived pages

timoludwig:
### Short description

Do not check links in archived pages

### Proposed changes

- Delete link objects when pages/event/locations are archived
- Create link objects when pages/event/locations are restored

### Resolved issues

Fixes: #1502

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,9 @@ UNRELEASED
 ----------
 
 * [ [#1516](https://github.com/digitalfabrik/integreat-cms/issues/1516) ] Fix save buttons alignment
-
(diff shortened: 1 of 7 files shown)


ulliholtgrave (dismissed):
Thanks, Works as expected :+1: . However, We need to restart the link finding process on the system, as the links for the archived pages are still in the db.

timoludwig:
> However, We need to restart the link finding process on the system, as the links for the archived pages are still in the db.

Yes, but speaking of which: I forgot to adapt the filter for the normal scanning process, so links in archived pages will still be checked if we just restart the scan.

Issue #1520, opened 2022-06-01
Update button text incorrect in side by side view


timoludwig:
### Describe the Bug

When a page/event/location is already published, the publish-button should be called "Update" instead of "Publish".

### Steps to Reproduce

1. Go to the side by side view
2. See incorrect button text

### Expected Behavior

Update

### Actual Behavior

Publish

lucky-chap:
Will do that @timoludwig

timoludwig:
@lucky-chap could you give a bit more information? Which OS are you using, what's the output of `command -v python3.9` and `python3.9 --version`?

timoludwig:
Ok, and which OS? I guess you can remove this check temporarily:
https://github.com/digitalfabrik/integreat-cms/blob/25f64a41e03c819d812bd4bfe2e8209043055084/dev-tools/install.sh#L14-L16
And hopefully no other problems arise.

lucky-chap:
Gonna try that. And I'm on Windows 11

lucky-chap:
Now it's telling me to install pipenv for python3. version 2018.10.9 or later. I have done that but still not working. Dunno what's wrong.

timoludwig:
Sorry, the dev-tools are more or less targeted at unix-based systems. Are you using the windows subsystem for linux?
If not, you can try to install the dependencies manually without pipenv and the dev-tools, but we don't have documentation for this unfortunately.

PR #1525, opened 2022-06-05
Fix button name in side-by-side view

MizukiTemma:
### Short description

This PR fixes the text on the button in side-by-side view so that it says "Publish" when the shown translation is not yet published but "Update" when already published.

### Proposed changes

- The change has been applied for side-by-side view for page.
- I cannot find any side-by-side view for event and location. Could anybody show me where the feature is?
- but for Imprint side-by-side view is available. Should the change be also here applied?

### Resolved issues

Fixes: #1520

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ UNRELEASED
 
 * [ [#1516](https://github.com/digitalfabrik/integreat-cms/issues/1516) ] Fix save buttons alignment
 
+* [ [#1520](https://github.com/digitalfabrik/integreat-cms/issues/1520) ] Fix button name in side-by-side view
 
 2022.6.0
 --------
--- a/integreat_cms/cms/templates/imprint/imprint_sbs.html
+++ b/integreat_cms/cms/templates/imprint/imprint_sbs.html
@@ -22,7 +22,13 @@
                     </a>
                     {% if perms.cms.change_imprintpage %}
                         <button name="submit_draft" class="btn btn-gray">{% trans 'Save as draft' %}</button>
-                        <button name="submit_public" class="btn">{% trans 'Publish' %}</button>
+                        <button name="submit_public" class="btn">
+                            {% if imprint_translation_form.instance.status == PUBLIC %}
+                                {% trans 'Update' %}
+                            {% else %}
+                                {% trans 'Publish' %}
+                            {% endif %}
+                        </button>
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,7 @@ UNRELEASED
 ----------
 
 * [ [#1516](https://github.com/digitalfabrik/integreat-cms/issues/1516) ] Fix save buttons alignment
-
+* [ [#1520](https://github.com/digitalfabrik/integreat-cms/issues/1520) ] Fix button name in side-by-side view
```
```suggestion

* [ [#1520](https://github.com/digitalfabrik/integreat-cms/issues/1520) ] Fix button name in side-by-side view
```
Unfortunately, the changelog format is very strict - it needs exactly one line after the heading and two lines between the last bullet point and the next heading.

timoludwig (commented):
Thanks, looks good! :+1:
Could you repeat the process for the imprint side by side view?
And squash your commits into one [as described here](https://digitalfabrik.github.io/integreat-cms/git-flow.html#clean-up-git-history) afterwards?

integreat_cms/cms/templates/imprint/imprint_sbs.html
```diff
@@ -22,7 +22,13 @@ <h1 class="heading">
                     </a>
                     {% if perms.cms.change_imprintpage %}
                         <button name="submit_draft" class="btn btn-gray">{% trans 'Save as draft' %}</button>
-                        <button name="submit_public" class="btn">{% trans 'Publish' %}</button>
+                        <button name="submit_public" class="btn">
+                            {% if imprint_translation_form.instance.status == 'PUBLIC' %}
```
timoludwig, line 26:
```suggestion
                            {% if imprint_translation_form.instance.status == PUBLIC %}
```
I think to stay consistent with the other templates, we should not hardcode this value (although it is unlikely that it will change in the near future, but rather pass the `PUBLIC` variable to the template context like in the other views...
You could add e.g. `from ...constants import status` and `"PUBLIC": status.PUBLIC,` here: https://github.com/digitalfabrik/integreat-cms/blob/1b78f774ad9d766db630aaa5f975c1e3400315b9/integreat_cms/cms/views/imprint/imprint_sbs_view.py#L28

timoludwig (approved):
Awesome, that's it! :rocket: :muscle:

Issue #1516, opened 2022-06-01
Save buttons jump to the left if title is too long


timoludwig:
### Describe the Bug

On most forms, the "save"/"update"-buttons jump to the left if the title is too long.

### Steps to Reproduce

1. Create page with very long title (or make browser window smaller)

### Expected Behavior

The buttons should always be aligned to the right, no matter whether they're in the same line as the title or below.

### Actual Behavior

They break into the next line but are left aligned

### Additional Information

![image-2022-06-01-10-20-00-486](https://user-images.githubusercontent.com/16021269/171390458-53de30dc-841a-411e-a61e-0a5f6cda2b89.png)

PR #1527, opened 2022-06-06
Fix horizontal align of page form save buttons

david-venhoff:
### Short description

This pr fixes the horizontal alignment of the page form save buttons and prevents horizontal scroll for very long page titles
grafik

### Proposed changes

- Specify alignment for the button row
- Add overflow: ellipsis to page title
- Specify word wrap-all behavior for the page order table to show long titles in multiple lines

### Resolved issues

Fixes: #1516

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 UNRELEASED
 ----------
 
+* [ [#1516](https://github.com/digitalfabrik/integreat-cms/issues/1516) ] Fix save buttons alignment
+
 
 2022.6.0
 --------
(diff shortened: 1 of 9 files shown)


ulliholtgrave:
Thanks, that works very well. However, Can you also add this to the forms for events and pois? :)

david-venhoff:
I added the same styling to the events, pois and push notification forms

david-venhoff:
Long names can cause some more issues in general, e.g. this is how long event names currently get displayed in the table. Should I open an issue for that?
grafik

david-venhoff:
The pr now additionally includes users, regions and languages

ulliholtgrave (approved):
Thanks, looks good now :+1:

Issue #1258, opened 2022-03-02
Provide possibility to confirm the up-to-dateness of a page 


ulliholtgrave:
### Motivation

If our content has not been changed for too long, it can mean a loss of credibility. We should therefore provide the possibility for editors to confirm the timeliness.

### Proposed Solution

The page form should have a function under "Actions" that allows an editor to confirm that all translations are up to date. Then all current translations of the page should be updated and the last_modified field should be set to the current time.

svenseeberg:
IMHO this should be a post migration issue.

timoludwig:
> IMHO this should be a post migration issue.

No, this is a feature request. If the content of a page hasn't changed in two years, but the content is still valid, the content managers want to have the possibility to refresh the date to make it obvious that the information is not outdated. At the moment, this takes `2n` clicks for `n` languages, so for cities with many languages, this soon becomes a time-consuming task if performed for multiple pages.

svenseeberg:
Yes, but this is not a function of WP either.

timoludwig:
Oh sorry, I misunderstood your objection. Yes, I agree, should be fine for the 1.2 milestone. If somebody wants to implement it earlier, fine, but this is not a hard requirement.

dkehne:
Also we could work within the analytics-page to show up the last x pages which are the oldest ones OR we show pages there which haven't been updated for x years.

PR #1529, opened 2022-06-07
Confirm pages timeliness

membralala:
### Short description

Users can now confirm pages timeliness and refresh the date/timestamp of the latest version of all translations of a page. 

### Proposed changes

- Enhances page form actions with a 'refresh date' button
- Adds a 'refresh dates' bulk action to the page tree view 

### Resolved issues

Fixes: #1258


integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -4971,20 +4971,29 @@ msgid "Archive this page"
...
 
 #: cms/templates/pages/page_form.html:329
-#: cms/templates/pages/page_form.html:393
+msgid "Refresh date"
+msgstr "Datum aktuallisieren"
+
+#: cms/templates/pages/page_form.html:332
+#: cms/templates/pages/page_form.html:335
+msgid "Refresh"
+msgstr "Aktuallisieren"
```
ulliholtgrave, line 4980: Hier gibt es nen kleinen Typo

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -4971,20 +4971,29 @@ msgid "Archive this page"
 msgstr "Diese Seite archivieren"
 
 #: cms/templates/pages/page_form.html:329
-#: cms/templates/pages/page_form.html:393
+msgid "Refresh date"
+msgstr "Datum aktuallisieren"
```
ulliholtgrave, line 4975: Typo :)

integreat_cms/cms/views/pages/page_actions.py
```diff
@@ -827,3 +828,45 @@ def render_mirrored_page_field(request, region_slug, language_slug):
...
+    :param language_slug: The slug of the current language
+    :type language_slug: str
+
+    :return: A redirection to the :class:`~integreat_cms.cms.views.pages.page_form_view.PageFormView`
+    :rtype: ~django.http.HttpResponseRedirect
+    """
+
+    page_translations = PageTranslation.objects.filter(page_id=page_id)
+    for page_translation in page_translations:
+        if page_translation is page_translation.latest_version:
```
timoludwig, line 862: `is` is the wrong operator here - it checks whether the two objects reference the same memory location, but since `.latest_version` does a database request and returns a new python object (of the same database object), `is` will always return `False` here. But that just as a side note, since there is a better solution anyway.
```
$ python
>>> a = [1,2,3]
>>> b = [1,2,3]
>>> a == b
True
>>> a is b
False
```

integreat_cms/cms/views/pages/page_actions.py
```diff
@@ -827,3 +828,45 @@ def render_mirrored_page_field(request, region_slug, language_slug):
...
+    :type region_slug: str
+
+    :param language_slug: The slug of the current language
+    :type language_slug: str
+
+    :return: A redirection to the :class:`~integreat_cms.cms.views.pages.page_form_view.PageFormView`
+    :rtype: ~django.http.HttpResponseRedirect
+    """
+
+    page_translations = PageTranslation.objects.filter(page_id=page_id)
```
timoludwig, line 860: Just filtering for the page translations is not enough here - theoretically, a malicious user could mark pages of other regions as up to date :smiling_imp:

integreat_cms/cms/views/pages/page_actions.py
```diff
@@ -827,3 +828,45 @@ def render_mirrored_page_field(request, region_slug, language_slug):
             "page_form": page_form,
         },
     )
+
+
```
timoludwig, line 832:
```suggestion

@require_POST
```
Since this view changes database contents, we need to protect it against CSRF-attacks, otherwise attackers could send refresh-links to users (or redirect to them from their site) and mark pages as up to date which are really outdated (which would be catastrophic). The simple way to prevent this is just to convert it to a post-only view, because then Django automatically requires the CSRF-token.

integreat_cms/cms/views/pages/page_actions.py
```diff
@@ -827,3 +828,45 @@ def render_mirrored_page_field(request, region_slug, language_slug):
...
+    :type language_slug: str
+
+    :return: A redirection to the :class:`~integreat_cms.cms.views.pages.page_form_view.PageFormView`
+    :rtype: ~django.http.HttpResponseRedirect
+    """
+
+    page_translations = PageTranslation.objects.filter(page_id=page_id)
+    for page_translation in page_translations:
+        if page_translation is page_translation.latest_version:
+            page_translation.save()
```
timoludwig, line 863:
```suggestion
    page = get_object_or_404(request.region.pages, id=page_id)
    for page_translation in page.translations.distinct("page__pk", "language__pk"):
        page_translation.save()
```
The simple way to get all newest versions of all languages is a distinct query - this makes sure only one result is returned for each page/language combination, and since they are already ordered by version, the returned element will be the latest version.
Only remaining problem: Now the order of the translations are more or less random, but we need to save them in the correct order to preserve their state based on the region's language tree.
So the default language needs to be saved first, then the languages that are direct children of the default language and so on... I leave this task up to you :innocent:

integreat_cms/cms/views/pages/page_actions.py
```diff
@@ -827,3 +828,45 @@ def render_mirrored_page_field(request, region_slug, language_slug):
...
+
+    # Get url of current page form for the redirect
+    language = get_object_or_404(Language, slug=language_slug)
+    page_translation = page_translations.filter(language=language).first()
+    redirect_url = page_translation.backend_edit_link
+    messages.success(
+        request, _("Successfully refreshed the date for all translations.")
+    )
+    return redirect(redirect_url)
```
timoludwig, line 872:
```suggestion
    return redirect(
        "edit_page",
        **{
            "page_id": page_id,
            "language_slug": language_slug,
            "region_slug": request.region.slug,
        },
    )
```
This does a few unnecessary database requests - I think this is enough.

integreat_cms/cms/views/pages/page_bulk_actions.py
```diff
@@ -114,3 +114,38 @@ def post(self, request, *args, **kwargs):
...
+        :return: The redirect
+        :rtype: ~django.http.HttpResponseRedirect
+        """
+        # Get all page translations to the corresponding pages of the given queryset
+        page_translations = PageTranslation.objects.filter(
+            page__id__in=self.get_queryset()
+        )
+        for page_translation in page_translations:
+            # Only refresh the date for the last version of a page translation
+            if page_translation is page_translation.latest_version:
```
timoludwig, line 145:
```suggestion
        for page in self.get_queryset():
            # Only refresh the date for the last version of a page translation
            for page_translation in page.translations.distinct("page__pk", "language__pk"):
```
I think this might be more performant - however, the same problem as with the individual view: We need to preserve the order of the save operations according to the region's language tree.

integreat_cms/cms/templates/pages/page_form.html
```diff
@@ -324,6 +324,17 @@ <h3 class="heading font-bold text-black">
                                         {% trans 'Archive this page' %}
                                     </button>
                                 {% endif %}
+                                {% if perms.cms.change_page and not page.archived %}
+                                    <label>
+                                        {% trans 'Refresh date' %}
+                                    </label>
+                                    <a href="{% url 'refresh' page_id=page_form.instance.id region_slug=request.region.slug language_slug=language.slug  %}"
```
timoludwig, line 331: Since this view changes the database, we need to convert it into a form + button instead of normal link.

timoludwig (commented):
Thanks a lot! :+1:
This is going in the right direction, but a few changes are still required.

timoludwig:
@membralala Ok, so the conclusion was to remove the bulk action and only have the button in the form to create the incentive to look at the page before confirming the actuality.

ulliholtgrave:
@membralala As the result of the discussion it seems like we don't really need any functionality in the tree view. Sou you can remove it from there and only keep the function on the page form.

membralala:
@ulliholtgrave , @timoludwig  Thanks for your detailed feedback, I removed the bulk action and enhanced my first solution.

integreat_cms/cms/views/pages/page_actions.py
```diff
@@ -827,3 +828,52 @@ def render_mirrored_page_field(request, region_slug, language_slug):
...
+    :type language_slug: str
+
+    :return: A redirection to the :class:`~integreat_cms.cms.views.pages.page_form_view.PageFormView`
+    :rtype: ~django.http.HttpResponseRedirect
+    """
+    region = request.region
+    page = get_object_or_404(region.pages, id=page_id)
+
+    # Consider only the last version of each translation
+    page_translations = page.translations.distinct("page__pk", "language__pk")
```
timoludwig, line 864:
```suggestion
    page_translations = page.translations.filter(
        language__in=region.active_languages
    ).distinct("page__pk", "language__pk")
```
We could also filter the translations to return only active ones.

integreat_cms/cms/views/pages/page_actions.py
```diff
@@ -827,3 +828,52 @@ def render_mirrored_page_field(request, region_slug, language_slug):
...
+    # Refresh page translations respective their languages depth level in the
+    # language tree, to make sure, that translations won't count as outdated.
+    language_nodes = LanguageTreeNode.objects.filter(region=region).order_by("depth")
```
timoludwig, line 867:
```suggestion

    # Sort page translations according to the position of their languages in the
    # language tree to ensure that the translations are not considered outdated.
    page_translations = sorted(
        page_translations,
        key=lambda page_translation: region.active_languages.index(
            page_translation.language
        ),
    )

```
Instead of querying the language tree nodes and then iterating over it, we could also sort the page translation queryset by the in index of their language in the region's language tree (I noticed that tree nodes are ordered by their `tree_id` and `lft` fields [by default](https://github.com/django-treebeard/django-treebeard/blob/70fd011f5a3bc7fe21653cbdba2c7d6e931aaec2/treebeard/ns_tree.py#L106) so the tree structure is preserved anyway and we can use the prefetched region properties), this way the order is correct, we save a few database queries and we don't have the problem of requesting languages that don't have a corresponding page translation...
(your code would throw the error `'NoneType' object has no attribute 'save'` in this case)

integreat_cms/cms/views/pages/page_actions.py
```diff
@@ -827,3 +828,52 @@ def render_mirrored_page_field(request, region_slug, language_slug):
...
+    for language in language_nodes:
+        page_translations.filter(language=language.language).first().save()
```
timoludwig, line 869:
```suggestion
    # Update timestamps of all translations
    for page_translation in page_translations:
        page_translation.save()
```

timoludwig (commented):
Thanks a lot :+1:
Only a few final touches are missing.
And could you add a note to the changelog?

timoludwig (approved):
Awesome, thanks! :rocket:

Issue #1530, opened 2022-06-07
Improve error messages when duplicate translation imported via XLIFF import


timoludwig:
### Describe the Bug

When a zip file is uploaded which contains multiple XLIFF files for the same page, all these files will be imported with the same version number, which breaks database consistency.

### Steps to Reproduce

1. Export page as XLIFF
2. Copy file and create zip archive
3. Upload file
4. Open page revision view
5. Infinite redirection loop
6. See error on next "normal" page

### Expected Behavior

The import should fail with an error message

### Actual Behavior

The import succeeds with duplicated translation versions

MizukiTemma:
Is this solved by #1993 ?

Issue #1518, opened 2022-06-01
Race conditions in tree operations causing Internal Server Errors


timobrembeck:
### Describe the Bug

When content forms are submitted several times in succession too quickly, the same translation version can be created multiple times.

### Steps to Reproduce

Send edit requests multiple times after another (for me, it didn't work to push the "save" button multiple times, but I could simulate this beaviour and we know that it already happened in the production system, so somehow this problem definitely exists)

Maybe, this is a combination of autosave and the real saving process?

### Expected Behavior

There should only exist exactly one database object for each translation version.

### Actual Behavior

In some cases, a version can be created multiple times, which breaks the CMS functionality in multiple places.

ulliholtgrave:
As a first approach we could deactivate the form buttons after they got clicked. That won't really hurt and we can eliminate this possible source for the problem.

dkehne:
Can you specify that a bit more that we can priorize it in our service team?

timoludwig:
> Can you specify that a bit more that we can priorize it in our service team?

We already released a quick fix so no duplicate versions are created and an internal server error is shown instead.
If you received no complaints from cities that they get errors when trying to save pages, this does not seem to be a big problem anymore. But we can also evaluate our log files to estimate how frequently this error occurs.

ulliholtgrave:
However, this is probably the same problem that created the two Landshuts yesterday. As this error always requires some manual cleanup I would prioritize it rather high.

timoludwig:
For reference: This is the corresponding issue in our tree library which could have an impact on this:
https://github.com/django-treebeard/django-treebeard/issues/53

timoludwig:
One example traceback during link replacement:
```
The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 98, in dispatch
[...]

PR #1538, opened 2022-06-09
Add database constraints to enforce unique version numbers

timoludwig:
### Short description

Ensure version numbers are unique for all content objects in each language.

### Proposed changes

- Add database `UniqueConstraint`s for all content translation models
- Adapt xliff test case because it is now failing due to the constraint

### Resolved issues

My hope is that this is a quick temporary fix for #1518 and #1530 to make sure that at least the database integrity is preserved and an internal server error is thrown instead of creating duplicate translation versions.


ulliholtgrave (approved):
Seems reasonable and for my manipulated XLIFF this seems to work. I think we should merge this rather fast and bring this live

Issue #1539, opened 2022-06-13
Urlencode permalinks when copying to clipboard


ulliholtgrave:
### Describe the Bug

Permalinks get created with "Umlauten" which results in links that are not usable on social media (e.g. instagram). A sample can be seen in the screenshot.
We also have the same issue for cyrillic letters, but the german letters have a greater priority right now.

### Expected Behavior

~Especially the german special cases should be translated to accepted translations (e.g. ae for ä).~
Urlencode all permalinks when the "copy to clipboard" button is clicked. This is also the same behavior when urls are copied from the url bar of most browsers.

### Additional Information

![grafik](https://user-images.githubusercontent.com/15608678/173300479-16bda5d1-da97-4ee0-992a-37d1b71dd4cf.png)

svenseeberg:
UTF-8 characters should be okay. There is no difference between Chinese or Arabic letters and German Umlauts. Therefore we should probably tackle this issue by modifying the copy link button to export a URL encoded link. That should always work.

PR #1543, opened 2022-06-13
Urlencode links when copied to clipboard

david-venhoff:
### Short description

This pr changes the copy to clipboard button in the content forms to encode links, so they can be used on platforms which don't have unicode support.

### Proposed changes

- Apply the builtin `encodeURI` function to the copied text 

### Resolved issues

Fixes: #1539


timoludwig (approved):

Issue #1542, opened 2022-06-13
Button for "copy url" of Short-URLs is misplaced


ulliholtgrave:
### Describe the Bug

The copy button of the ShortURL feature is located in a new line instead of behind the field

### Additional Information

![grafik](https://user-images.githubusercontent.com/15608678/173309922-f9e60699-0866-4929-8edb-584f4418e5ca.png)

PR #1544, opened 2022-06-13
Fix short url copy button position

david-venhoff:
### Short description

This pr fixes the position of the copy short url button in the page form and imprint form to make it consistent with the normal page url.

grafik

### Resolved issues

Fixes: #1542


timoludwig (approved):

Issue #1517, opened 2022-06-01
Set all pages to draft when duplicating region


timoludwig:
### Motivation

When a region is cloned, the content needs to be reviewed before it is usable

### Proposed Solution

When a region is cloned, all pages of the new region should be set to draft.

dkehne:
@timoludwig we have discussed internally and we do not need the bulk action BUT we would like to have the behaviour that after a duplication (of an instance) all pages are on "draft"-status.

dkehne:
prio medium-high if not much effort.

PR #1546, opened 2022-06-14
Set all pages to draft when duplicating regions

timoludwig:
### Short description

Set all pages to draft when duplicating regions

### Resolved issues

Fixes: #1517


ulliholtgrave (approved):
Thanks, That does the trick :+1:

Issue #1550, opened 2022-06-14
Revisions always restored as public


timoludwig:
### Describe the Bug

No matter which button is used to restore page revisions, the revision is always restored as public.

### Steps to Reproduce

1. Go to page revisions
2. Restore a version as draft
3. The status is identical to the restored version, no matter which button is used

### Expected Behavior

When it's imported as draft, it should be a draft, and when it's imported as public, it should be public.

### Actual Behavior

The status does not change, no matter which button is used

PR #1551, opened 2022-06-14
Fix status change when restoring revisions

timoludwig:
### Short description

Fix status change when restoring revisions

### Resolved issues

Fixes: #1550


membralala (approved):
Thank you, this works as expected!

PR #1553, opened 2022-06-15
Add map bounding box to region model

timoludwig:
### Short description

Add map bounding box to region model

### Proposed changes

- Add bounding coordinates fields to region model
- Automatically retrieve them via Nominatim API
- Show warning when locations are enabled but bounding box is unset

### Resolved issues

Fixes: #742


ulliholtgrave:
> > we should probably change the PLZ to required
> 
> The postal code is already required.
> 
Ahh yes, you are right. :see_no_evil: 

> > maybe add a descriptive text that mentions the usage.
> 
> Don't you think this is already sufficient?

Your changes are fine, but I think we should provide some description to the original plz field, as it has some new implications. Especially if we set a specific PLZ which is not a central within our region area, this has some direct implications to the bounding box.
grafik

ulliholtgrave (approved):
Looks good :+1: Nominatim is still kind of a black box for me, but your new calculations are returning better results :tada:

Issue #1528, opened 2022-06-06
Long object titles break list view layouts


david-venhoff:
### Describe the Bug

When choosing a very long title for some kind of object, the corresponding list view layouts can break:
![grafik](https://user-images.githubusercontent.com/78504586/172213560-681b3c89-71ba-4589-b333-22d6fe5d9db2.png)

This seems to affect: events, pois, news, users, languages, regions

### Expected Behavior

No horizontal scrollbar appears

### Actual Behavior

The table gets wider than the screen width

PR #1554, opened 2022-06-15
Fix list view layout for long titles

membralala:
### Short description

This PR should fix some misalignments in the list view for long titles / small screens.

### Proposed changes

- Enable horizontal scrolling for the ```table-listing```  css class
- Truncate titles to 35 chars 

### Resolved issues

Fixes: #1528

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 UNRELEASED
 ----------
 
+* [ [#1528](https://github.com/digitalfabrik/integreat-cms/issues/1528) ] Fix list view layouts for long titles
+
 
 2022.6.3
 --------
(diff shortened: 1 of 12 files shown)


timoludwig (approved):

The only disadvantage I see is that on most larger screens there is quite a bit of unused screen real estate... The truncation to a fixed character limit is not really suitable for responsive design, so in the long term, I would prefer doing the truncation in css similar to how it's done in the [feedback list](https://github.com/digitalfabrik/integreat-cms/blob/c7411b5b6cc9be8da7f4622e47b9298b08f62d4b/integreat_cms/cms/templates/feedback/region_feedback_list.html#L60).

But since this is a reasonable improvement over the current solution, let's merge it! :rocket:

membralala:
Thanks for your feedback! Should I open a follow-up issue for the enhancement of this solution?

Issue #1497, opened 2022-05-25
Setting an event/location to draft noes not change older versions


timoludwig:
### Describe the Bug

When an event/location is saved as draft, only the latest version is marked as "draft" and older public versions are still delivered via the API

### Steps to Reproduce

1. Publish event
2. Change the content and save as draft

### Expected Behavior

The event should not be delivered at all - since the status is now "draft"

### Actual Behavior

The old published version is delivered

### Additional Information

We already solved this for pages (all past versions need to be stored as draft when an event is stored as draft) and also need to fix this for events and pois.
Since changing all past versions to draft also means that users without publishing permissions could un-publish events/locations, we have to repeat the fixes for #1445 for events/locations as well.

Issue #1445, opened 2022-05-17
Do not allow roles without publishing rights to un-publish pages


timoludwig:
### Motivation

At the moment, user roles who do not have the permission to publish pages, can un-publish pages by storing them as a draft (including all past versions)

### Proposed Solution

User roles who can only change pages and not publish pages should only have the permission to store something as "in review" and not as draft (which also affects past public versions)

PR #1555, opened 2022-06-16
Set old versions to draft, when saved as draft (Event, Location)

MizukiTemma:
### Proposed changes

- When an event/locations is saved as draft, its older versiond as well as the translations will be marked as draft.
- The same changes as in #1445 are applied for event  (no "save as draft" button available for users without publishing permission, permission check in the view) but not for location, since there is no publish permission defined for locations.

### Resolved issues

Fixes: #1497


timoludwig (approved):

Issue #1509, opened 2022-05-29
Legacy sitemap URLs missing


timoludwig:
### Describe the Bug

The web app sitemap https://integreat.app/sitemap.xml uses the old sitemap endpoint from wordpress, which returrns 404 at the moment.

### Expected Behavior

Support the following URL patterns:

Sitemap-Index: https://cms.integreat-app.de/wp-json/ig-sitemap/v1/sitemap-index.xml
Region sitemaps: https://cms.integreat-app.de/region-slug/language-slug/wp-json/ig-sitemap/v1/sitemap.xml

### Actual Behavior

Only the new URLs

Sitemap-Index: https://cms.integreat-app.de/sitemap.xml
Region sitemaps: https://cms.integreat-app.de/region-slug/language-slug/sitemap.xml

work and the old URLs return 404.

PR #1556, opened 2022-06-16
Add legacy sitemap url patterns

david-venhoff:
### Short description

This pr adds legacy sitemap url patterns:
 * https://cms.integreat-app.de/wp-json/ig-sitemap/v1/sitemap-index.xml
 * https://cms.integreat-app.de/region-slug/language-slug/wp-json/ig-sitemap/v1/sitemap.xml

### Resolved issues

Fixes: #1509

### DIFF
--- a/integreat_cms/sitemap/urls.py
+++ b/integreat_cms/sitemap/urls.py
@@ -10,7 +10,7 @@ See :mod:`~integreat_cms.core.urls` for the other namespaces of this application
 
 For more information on this file, see :doc:`topics/http/urls`.
 """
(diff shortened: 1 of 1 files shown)


ulliholtgrave (changes requested):
Thanks, the index sitemap is working as expected, but the sitemaps for the region are returned without XML-formatting (see screenshot). I guess, we should fix that before we enable the legacy route again and maybe break the app with it :see_no_evil: 
grafik

The index sitemap looks like this for me:
grafik

Issue #1510, opened 2022-05-30
Limit duration of events to 7 days


timoludwig:
### Motivation

Events with a duration of more than 7 days are usually not really continuous and should rather be replaced by recurring events.

### Proposed Solution

Limit the duration of events to 7 days

PR #1557, opened 2022-06-16
Limit event duration to 7 days

MizukiTemma:
### Short description

This PR limits the maximal duration of an event to 7 days.

### Proposed changes

- It is no longer allowed to create an one-time event with duration longer than 7 days or to change the duration of an one-time event to over 7 days.
- A small explanation for this limitation is added on the event editor page above the date inputs fields.

### Resolved issues

Fixes: #1510

### DIFF
--- a/integreat_cms/cms/forms/events/event_form.py
+++ b/integreat_cms/cms/forms/events/event_form.py
@@ -1,6 +1,6 @@
 import logging
 
-from datetime import time
+from datetime import time, timedelta
 
 from django import forms
 from django.utils.translation import ugettext_lazy as _
@@ -95,6 +95,16 @@ class EventForm(CustomModelForm):
         # make self.data mutable to allow values to be changed manually
         self.data = self.data.copy()
 
+        if cleaned_data["end_date"] - cleaned_data["start_date"] > timedelta(6):
+            self.add_error(
+                "end_date",
+                forms.ValidationError(
+                    _(
+                        "The maximum duration for events is 7 days. Consider using recurring events if the event is not continuous."
+                    ),
+                    code="invalid",
+                ),
+            )
         if cleaned_data.get("is_all_day"):
             cleaned_data["start_time"] = time.min
             self.data["start_time"] = time.min
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-06-24 08:36+0000\n"
+"POT-Creation-Date: 2022-06-24 13:50+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: Integreat <info@integreat-app.de>\n"
 "Language-Team: Integreat <info@integreat-app.de>\n"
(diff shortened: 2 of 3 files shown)


integreat_cms/cms/templates/events/event_form.html
```diff
@@ -142,6 +142,10 @@ <h3 class="heading font-bold text-black">
                     </div>
                     <div class="w-full p-4 rounded bg-white">
                         <div>
+                            <label>{% trans 'Maximal duration allowed for one-time event is 7 days' %}</label>
+                            <div class="help-text">
+                                {% trans 'Set an event longer than 7 days as recurring event.' %}
+                            </div>
```
timoludwig, line 148:
```suggestion
```
I don't think we need this as permanent info - the warning when somebody tries to save an event with more than 7 days should be enough.

integreat_cms/cms/forms/events/event_form.py
```diff
@@ -1,6 +1,7 @@
 import logging
 
 from datetime import time
+import datetime
```
timoludwig, line 4:
```suggestion
from datetime import time, timedelta
```
We can combine these imports

integreat_cms/cms/forms/events/event_form.py
```diff
@@ -95,7 +96,21 @@ def clean(self):
         # make self.data mutable to allow values to be changed manually
         self.data = self.data.copy()
 
-        if cleaned_data.get("is_all_day"):
+        if (
+            cleaned_data["end_date"] - cleaned_data["start_date"]
+            > datetime.timedelta(6)
+            and not cleaned_data["is_recurring"]
+        ):
```
timoludwig, line 103:
```suggestion
        if cleaned_data["end_date"] - cleaned_data["start_date"] > timedelta(6):
```
I guess each occurrence of recurring events should also not take place longer than 7 days. Only the timedelta between the initial start date and the recurrence end date may be longer than 7 days.

integreat_cms/cms/forms/events/event_form.py
```diff
@@ -95,7 +96,21 @@ def clean(self):
...
+            self.add_error(
+                "end_date",
+                forms.ValidationError(
+                    _(
+                        "Event duration longer than 7 days is not allowed for one-time event."
```
timoludwig, line 108:
```suggestion
                        "The maximum duration for events is 7 days. Consider using recurring events if the event is not continuous."
```

integreat_cms/cms/forms/events/event_form.py
```diff
@@ -95,7 +96,21 @@ def clean(self):
...
+                    ),
+                    code="required",
```
timoludwig, line 110:
```suggestion
                    code="invalid",
```
The problem is not really that no value was entered at all, but an invalid value was entered.

integreat_cms/cms/forms/events/event_form.py
```diff
@@ -95,7 +96,21 @@ def clean(self):
...
+                ),
+            )
+        elif cleaned_data.get("is_all_day"):
```
timoludwig, line 113:
```suggestion
        if cleaned_data.get("is_all_day"):
```
I don't think we should use an `elif` here, since both `if` cases are possible at the same time.

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -3981,67 +3985,75 @@ msgstr "Auswirkungen auf andere Übersetzungen"
...
-#: cms/templates/events/event_form.html:244
+#: cms/templates/events/event_form.html:145
+msgid "Maximal duration allowed for one-time event is 7 days"
+msgstr "Maximale zugelassene Dauer für einmalige Veranstaltung ist 7 Tage"
+
+#: cms/templates/events/event_form.html:147
+msgid "Set an event longer than 7 days as recurring event."
+msgstr "Stelle eine längere Veranstaltung als wiederholend ein."
+
+#: cms/templates/events/event_form.html:248
```
timoludwig, line 3996:
```suggestion
#: cms/templates/events/event_form.html:244
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -1567,46 +1567,50 @@ msgstr "{} \"{}\" wurde erfolgreich veröffentlicht"
...
 msgstr ""
 "Entweder muss der Veranstaltungsort deaktiviert werden oder es muss ein "
 "gültiger Ort angegeben werden"
 
-#: cms/forms/events/event_form.py:110
+#: cms/forms/events/event_form.py:108
+msgid "Event duration longer than 7 days is not allowed for one-time event."
+msgstr "Einmalige Veranstaltung darf nicht länger als 7 Tage sein."
+
+#: cms/forms/events/event_form.py:125
```
timoludwig, line 1601:
```suggestion
#: cms/forms/events/event_form.py:103
msgid ""
"The maximum duration for events is 7 days. Consider using recurring events "
"if the event is not continuous."
msgstr ""
"Die maximale Veranstaltungsdauer beträgt 7 Tage. Verwenden Sie "
"wiederkehrende Veranstaltungen, wenn die Veranstaltung nicht kontinuierlich "
"stattfindet."

#: cms/forms/events/event_form.py:120
```

timoludwig (commented):
Thanks! :+1:
In general it looks good, just a few ideas for improvement.

timoludwig (approved):
Thanks! :rocket:

Issue #1563, opened 2022-06-20
Writing translation in split-screen-view not possible


FritjofK:
### Describe the Bug
If I want to translate a text via the split-screen-view, I can only write a title but cannot write text in the editor. 
I have permission to work in translations (it works in single-page-view). 

### Steps to Reproduce

1. Go to any page and click on a foreign language
2. Click on "Übersetzeransicht anzeigen"
3. Writing in editor not possible

### Expected Behavior
Writing texts should be possible

### Actual Behavior
Field is "blocked"

### Additional Information
has been reproduced by one other team member

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

PR #1564, opened 2022-06-20
Fix permission checks in side-by-side view

timoludwig:
### Short description

Fix permission checks in side-by-side view.
Along the way, I noticed a few other inconsistencies in our permission checks.

### Proposed changes

Implemented the following constraints:
- only view permission: make form disabled, remove save buttons, show warning
- only edit permission: only show "submit for review" button
- publish permission: show "save as draft" and "publish/update" buttons

### Resolved issues

Fixes: #1563


ulliholtgrave (changes requested):
I know that this is not directly part of your issue, but as you are already touching the side-by-side template: Can you fix the text alignment of the "back to the editor" button (I guess the main problem is that it's coded as an <a>-tag and is not a real button)

timoludwig:
> Can you fix the text alignment of the "back to the editor" button

Alright, done. I didn't like the way the buttons increased their height when the heading breaks into the new line anyway, so I adapted the flex behavior of the buttons instead.

ulliholtgrave (approved):

Issue #1511, opened 2022-05-30
PDF generation fails if filename is too long


timoludwig:
### Describe the Bug

When the title of a PDF file is too long, the PDF generation fails

### Steps to Reproduce

1. Create page with very long title
2. Export as PDF

alternatively, click on the PDF-icon on this page:

https://integreat.app/dillingenanderdonau/fa/infobriefe-fuer-alle-haupt-und-ehrenamtlich-taetigen-sowie-projekttraeger-in-den-bereichen-asyl-und-integration/infobriefe-fuer-alle-haupt-und-ehrenamtlich-taetigen-sowie-projekttraeger/15-infobrief-vom-18-februar2021-fuer-alle-haupt-und-ehrenamtlich-taeti-gen-sowie-projekttraeger-in-den-bereichen-asyl-und-integration-22-02-2021

which redirects to:

https://cms.integreat-app.de/dillingenanderdonau/fa/wp-json/ig-mpdf/v1/pdf/?url=%2Fdillingenanderdonau%2Ffa%2Finfobriefe-fuer-alle-haupt-und-ehrenamtlich-taetigen-sowie-projekttraeger-in-den-bereichen-asyl-und-integration%2Finfobriefe-fuer-alle-haupt-und-ehrenamtlich-taetigen-sowie-projekttraeger%2F15-infobrief-vom-18-februar2021-fuer-alle-haupt-und-ehrenamtlich-taeti-gen-sowie-projekttraeger-in-den-bereichen-asyl-und-integration-22-02-2021

### Expected Behavior

The filename should be truncated if too long

### Actual Behavior

an `Internal Server Error` (`OSError`) is thrown

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
May 30 09:30:42 ERROR   django.request - Internal Server Error: /dillingenanderdonau/fa/wp-json/ig-mpdf/v1/pdf/
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/api/decorators.py", line 124, in wrap
    return function(request, *args, **kwargs)
[...]

PR #1570, opened 2022-06-22
Fix PDF generation for long filenames

membralala:
### Short description

This PR fixes the ```OSError``` when exporting pages with very long titles. It does this by truncating filenames to at most 250 characters (and an additional file extension). 

### Resolved issues

Fixes: #1511

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ UNRELEASED
 * [ [#1504](https://github.com/digitalfabrik/integreat-cms/issues/1504) ] Keep filters on pagination
 * [ [#1585](https://github.com/digitalfabrik/integreat-cms/issues/1585) ] Hide news after 28 days
 * [ [#1600](https://github.com/digitalfabrik/integreat-cms/issues/1600) ] Improve XLIFF export bulk option description
+* [ [#1511](https://github.com/digitalfabrik/integreat-cms/issues/1511) ] Fix PDF generation for long filenames
 
 
 2022.6.3
--- a/integreat_cms/cms/utils/pdf_utils.py
+++ b/integreat_cms/cms/utils/pdf_utils.py
@@ -18,6 +18,8 @@ from django.views.decorators.cache import never_cache
 
 from xhtml2pdf import pisa
 
+
+from .text_utils import truncate_bytewise
 from ..constants import text_directions
 from ..models import Language, Page
 
@@ -83,7 +85,15 @@ def generate_pdf(region, language_slug, pages):
             # In any other case, take the region name
             title = region.name
     language = Language.objects.get(slug=language_slug)
-    filename = f"{pdf_hash}/{capfirst(settings.BRANDING)} - {language.translated_name} - {title}.pdf"
+    # Make sure, that the length of the filename is valid. To prevent potential
+    # edge cases, shorten filenames to 3/4 of the allowed max length.
+    ext = ".pdf"
+    try:
+        max_len = ((os.statvfs(settings.PDF_ROOT).f_namemax // 4) * 3) - len(ext)
+    except FileNotFoundError:
+        max_len = 192 - len(ext)
+    name = f"{capfirst(settings.BRANDING)} - {language.translated_name} - {title}"
+    filename = f"{pdf_hash}/{truncate_bytewise(name, max_len)}{ext}"
     # Only generate new pdf if not already exists
     if not pdf_storage.exists(filename):
         # Convert queryset to annotated list which can be rendered better
--- a/integreat_cms/cms/utils/text_utils.py
+++ b/integreat_cms/cms/utils/text_utils.py
@@ -17,3 +17,23 @@ def lowfirst(string):
     :rtype: str
     """
     return string and str(string)[0].lower() + str(string)[1:]
+
+
+def truncate_bytewise(string, length):
+    """
+    Truncate a UTF-8 encoded string to a maximum byte length.
+
+    :param string: The input text
+    :type string: str
+
+    :param length: The maximum length of the text in byte representation
+    :type length: int
+
+    :return: The truncated text
+    :rtype: str
+    """
+    encoded_string = string.encode()
+    try:
+        return encoded_string[:length].decode()
+    except UnicodeDecodeError as e:
+        return encoded_string[: e.start].decode()
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -6032,12 +6032,12 @@ msgstr "Unbekannt"
 msgid "CW"
 msgstr "KW"
 
-#: cms/utils/pdf_utils.py:69
+#: cms/utils/pdf_utils.py:71
 msgid "No valid pages selected for PDF generation."
 msgstr ""
 "Es wurden keine gültigen Seiten zur Erzeugung der PDF-Datei ausgewählt."
 
-#: cms/utils/pdf_utils.py:117
+#: cms/utils/pdf_utils.py:127
 msgid "The PDF could not be successfully generated."
 msgstr "PDF-Datei konnte nicht erfolgreich erzeugt werden."
 


integreat_cms/cms/utils/pdf_utils.py
```diff
@@ -83,7 +83,11 @@ def generate_pdf(region, language_slug, pages):
             # In any other case, take the region name
             title = region.name
     language = Language.objects.get(slug=language_slug)
-    filename = f"{pdf_hash}/{capfirst(settings.BRANDING)} - {language.translated_name} - {title}.pdf"
+    # Make sure, that the length of the filename is valid
+    truncated_filename = Truncator(
+        f"{pdf_hash}/{capfirst(settings.BRANDING)} - {language.translated_name} - {title}"
+    ).chars(250, truncate="")
+    filename = f"{truncated_filename}.pdf"
```
timoludwig, line 90:
```suggestion
    ext = ".pdf"
    max_len = os.statvfs(settings.PDF_ROOT).f_namemax - len(ext)
    name = f"{capfirst(settings.BRANDING)} - {language.translated_name} - {title}"
    filename = f"{pdf_hash}/{name[:max_len]}{ext}"
```
Since the maximum filename length depends on the file system used by the OS, I guess it's better to ask the system instead of hardcoding the value (although 255 chars seems to be pretty standard...)

timoludwig (commented):
Thanks! :+1:
This works as expected, however I have two small suggestions for improvement.

membralala:
Hey @timoludwig, thanks for your feedback! I realised some other issue in that I am not allowed to save the generated PDF due to it's long name and therefore have to manually shorten it. The limits of allowed chars seem to differ with the browser I am using. Can you reproduce this behaviour?

timoludwig:
> I realised some other issue in that I am not allowed to save the generated PDF due to it's long name and therefore have to manually shorten it. The limits of allowed chars seem to differ with the browser I am using. Can you reproduce this behaviour?

Interesting, no I cannot. Maybe your download directory is on another hard disk partition which uses a different filesystem than the directory where the CMS is served from? Could you check this with `python3 -c "import os; print(os.statvfs('/').f_namemax)"` by replacing `'/'` with these directories? Or maybe, this is indeed a browser issue? In which browser did you face this issue?

membralala:
```python3 -c "import os; print(os.statvfs('/').f_namemax)"``` returns 255. But I tested it again and found, that it depends on the characters used in the title (e.g. Umlaute), so I guess it is about the url encoding. It actually does not differ with the browsers.

**Comment by timoludwig** [comment]
> But I tested it again and found, that it depends on the characters used in the title (e.g. Umlaute), so I guess it is about the url encoding.

Ahh, of course, now I understand the problem. A utf-8 character may take multiple bytes in memory but only counts as one character when truncating the string. According to the corresponding stackoverflow-questions it seems to be surprisingly hard to truncate a utf-8 string to a given byte-length in python. 
Maybe our next best shot is so convert the name to ASCII before truncating the string, maybe via the slugify function?

**Comment by membralala** [comment]
I realised, that if we do this, we might get some trouble with the languages with non-latin alphabet, as slugify will ignore symbols, that it could not simplify. That might cause, that a lot of filenames of the same city and language will be reduced to the name pattern ```integreat-{language}```. 
As far as my research goes, the symbols that are encoded with more than 3 bytes might appear quite rarely in our context, so we could restrict the title length to about a third of the max length allowed by the filesystem. Or do you think, that it's worth the effort to build a solution, that truncates the title based on the given characters?

**Comment by timoludwig** [comment]
> I realised, that if we do this, we might get some trouble with the languages with non-latin alphabet, as slugify will ignore symbols, that it could not simplify. That might cause, that a lot of filenames of the same city and language will be reduced to the name pattern `integreat-{language}`.

Ok, yeah, then this reducing the filenames to ascii characters is probably an insufficient solution.

> As far as my research goes, the symbols that are encoded with more than 2 bytes might appear quite rarely in our context, so we could restrict the title length to about half of the max length allowed by the filesystem. Or do you think, that it's worth the effort to build a solution, that truncates the title based on the given characters?

Hmm, actually we have a few special character sets, not only Arabic, but also Cyrillic and Chinese, and I assume it's very possible that we end up with a title with characters that consists of more than 2 bytes on average.
However, after looking more into all the stackoverflow questions about this topic, I think I found an [acceptable solution](https://stackoverflow.com/a/70304695):
```
def truncate(string, length):
    encoded_string = string.encode()
    try:
        return encoded_string[:length].decode()
    except UnicodeDecodeError as e:
        return encoded_string[:e.start].decode()
```
The trick is that `.encode()` converts the string into a byte-sequence which can then be truncated to the desired number of bytes. Since this could cause multi-byte-characters to be cut in the middle which causes a `UnicodeDecodeError`, the try/except block makes sure that the string is truncated to the next best location where the string can still be decoded. The only disadvantage is that the last character might be different to the real character, but I think we can accept this limitation.

**Comment by membralala** [comment]
I implemented @timoludwig 's suggestion, which shortens the URLs as expected, however I still got the same strange behaviour and had to reduce the title manually by a few signs. So I limited the title  length to 3/4 of the allowed max length to prevent possible edge cases. This works for me, however it still does not feel like a beautiful solution, but I'm not sure, if it is worth to further investigate?

File: integreat_cms/cms/utils/text_utils.py [hunk 2]
```diff
@@ -17,3 +17,24 @@ def lowfirst(string):
     :rtype: str
     """
     return string and str(string)[0].lower() + str(string)[1:]
+
+
+def truncate_bytewise(string, length):
+    """
+    Truncate a string to a maximum byte length.
```
svenseeberg, line 24:
```suggestion
    Truncate a UTF-8 encoded string to a maximum byte length.
```

integreat_cms/cms/utils/text_utils.py
```diff
@@ -17,3 +17,24 @@ def lowfirst(string):
...
+    :type length: int
+
+    :return: The truncated text
+    :rtype: str
+    """
+    encoded_string = string.encode()
+    try:
+        print(len(encoded_string[:length]))
+        return encoded_string[:length].decode()
+    except UnicodeDecodeError as e:
```
svenseeberg, line 39: IMHO we could actually implement a correct parsing of the UTF-8 bytes, which is not that difficult: https://en.wikipedia.org/wiki/UTF-8#Encoding

We just need to jump to the position indicated by the `length` parameter and bytewise go back until the byte does not start with `10xxxxxx`. The check could be implemented like this: `if myByte&1 == 1 and myByte&2 == 0:`. I'm not a bit wizard though and there may be even better solutions ;)

integreat_cms/cms/utils/pdf_utils.py
```diff
@@ -83,7 +85,14 @@ def generate_pdf(region, language_slug, pages):
             # In any other case, take the region name
             title = region.name
     language = Language.objects.get(slug=language_slug)
-    filename = f"{pdf_hash}/{capfirst(settings.BRANDING)} - {language.translated_name} - {title}.pdf"
+    # Make sure, that the length of the filename is valid.
+    ext = ".pdf"
+    try:
+        max_len = ((os.statvfs(settings.PDF_ROOT).f_namemax // 4) * 3) - len(ext)
```
svenseeberg, line 91: You could add a comment here that we're limiting the length to 3/4 of the allowed length to navigate around potential edge cases.

svenseeberg (changes requested):
Thanks for the updated PR. I'm wondering if we should implement a correct UTF-8 parsing. That would not be that hard and avoid wrong characters at the end. We would have to loop through the last 3 bytes at most and do a relatively simple check to find out whether we're at the beginning of a UTF-8 character.

membralala:
Thanks for your feedback! Regarding the UTF-8 truncation I implemented a suggestion based on your approach (although using slightly different bit magic) , however I don't really know, if this has any benefits compared to Timo's suggestion. What do you think?

svenseeberg:
> Thanks for your feedback! Regarding the UTF-8 truncation I implemented a suggestion based on your approach (although using slightly different bit magic) , however I don't really know, if this has any benefits compared to Timo's suggestion. What do you think?

AFAICT Timos suggestion altered the last character, right? With this correctly implemented UTF-8 character parsing we should not have this issue.

membralala:
I rolled back to Timo's suggestion regards bytewise truncation, as both solutions should have the same behaviour, but Timo's is easier to read and maybe catches some edge cases, that I might have missed. @svenseeberg , are you fine with that?

CHANGELOG.md
```diff
@@ -16,7 +17,6 @@ UNRELEASED
 * [ [#1567](https://github.com/digitalfabrik/integreat-cms/issues/1567) ] Hide organization field in user form
 * [ [#1563](https://github.com/digitalfabrik/integreat-cms/issues/1563) ] Fix permission checks in side-by-side view
 
-
```
svenseeberg, line 19: Please keep the additional empty line here :)

integreat_cms/cms/utils/pdf_utils.py
```diff
@@ -18,6 +18,8 @@
 
 from xhtml2pdf import pisa
 
+
+from ..utils.text_utils import truncate_bytewise
```
timoludwig, line 22:
```suggestion
from .text_utils import truncate_bytewise
```
Just a minor detail - since you're already in the utils package, you don't need to traverse to the parent package to import from the text utils.

integreat_cms/cms/utils/pdf_utils.py
```diff
@@ -83,7 +85,15 @@ def generate_pdf(region, language_slug, pages):
             # In any other case, take the region name
             title = region.name
     language = Language.objects.get(slug=language_slug)
-    filename = f"{pdf_hash}/{capfirst(settings.BRANDING)} - {language.translated_name} - {title}.pdf"
+    # Make sure, that the length of the filename is valid. To prevent potential
+    # edge cases, shorten filenames to 3/4 of the allowed max length.
+    ext = ".pdf"
+    try:
+        max_len = ((os.statvfs(settings.PDF_ROOT).f_namemax // 4) * 3) - len(ext)
+    except (FileNotFoundError, ZeroDivisionError):
```
timoludwig, line 93: Just out of curiosity: How can this throw a `ZeroDivisionError`? Or did it just happen and you're unsure why? :slightly_smiling_face:

timoludwig (approved):
Thanks a lot! :+1:
I guess this is fine, if we're really bored we could investigate why this doesn't work in some edge cases and under which circumstances the last character is changed and whether this is confusing for the end users, but I assume our backlog will never get that empty :slightly_smiling_face:

svenseeberg (approved):
:tada:

PR #1576, opened 2022-06-25
Keep filters on pagination

david-venhoff:
### Short description

This draft implements persistent filters for the event list.
I am not quite sure how to best handle GET-forms best and would like to get some feedback about it.

### Proposed changes
- Modify pagination.html to not overwrite existing queries
- Change the event filter form method to 'GET'
- include some existing queries in the event filter form as hidden inputs


integreat_cms/cms/views/events/event_list_view.py
```diff
@@ -110,10 +110,10 @@ def get(self, request, *args, **kwargs):
         query = None
 
         # Filter events according to given filters, if any
-        filter_data = kwargs.get("filter_data")
-        if filter_data is not None:
+        # TODO How to check for a submitted form??
+        if "filter" in request.GET or "query" in request.GET:
```
david-venhoff, line 114: What would a better check for whether there are form data look like?

timoludwig: I think we could do this similar to the page filter form:

https://github.com/digitalfabrik/integreat-cms/blob/ab0a89576ae901f4b30aa8e9c65ff43c44654a80/integreat_cms/cms/views/pages/page_tree_view.py#L91-L99

https://github.com/digitalfabrik/integreat-cms/blob/ab0a89576ae901f4b30aa8e9c65ff43c44654a80/integreat_cms/cms/forms/pages/page_filter_form.py#L45-L53

timoludwig: Maybe you could even add a base class for all filter forms with such shared functionality

PR #1580, opened 2022-06-29
Improve user list

timoludwig:
### Short description

Just some small improvements for the user list

### Proposed changes

- Remove links to form from all columns except username
- Remove organization column
- Add region column for global user list

### DIFF
--- a/integreat_cms/cms/templates/users/region_user_list_row.html
+++ b/integreat_cms/cms/templates/users/region_user_list_row.html
@@ -6,43 +6,28 @@
             {{ user.username }}
         </a>
     </td>
-	<td>
-        <a href="{% url 'edit_region_user' region_slug=request.region.slug user_id=user.id %}" class="block py-3 pr-2 text-gray-800">
-            {{ user.first_name }}
-        </a>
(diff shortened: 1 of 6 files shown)


integreat_cms/cms/templates/users/region_user_list_row.html
```diff
@@ -6,43 +6,28 @@
...
-	</td>
-	<td>
-		<a href="{% url 'edit_region_user' region_slug=request.region.slug user_id=user.id %}" class="block py-3 pr-2 text-gray-800">
-			{{ user.role }}
-		</a>
+	<td class="pr-2">
+		{{ user.role }}
 	</td>
-	<td>
-		<a href="{% url 'edit_user' user_id=user.id %}" class="block py-3 pr-2 text-gray-800 whitespace-nowrap">
```
timoludwig, line 37: This faulty url (`edit_user` instead of `edit_region_user`) was leading to a few permission denied errors when the user form was opened by clicking on the login date

ulliholtgrave: Good catch

PR #1581, opened 2022-06-29
Improve wording of minor edit label

timoludwig:
### Short description

Improve wording of minor edit label

### Proposed changes

Before, the label always was "implications on other translations", but for translations that are not the default translation, this is not correct. For languages that are further down the language tree (either leaf nodes or somewhere between the default language and a leaf node), the minor edit field not only impacts the other translations, but the current translation as well.
This change of the wording reflects the actual implications of the minor edit field.

### DIFF
--- a/integreat_cms/cms/templatetags/content_filters.py
+++ b/integreat_cms/cms/templatetags/content_filters.py
@@ -84,6 +84,28 @@ def get_language(language_slug):
 
 
 @register.simple_tag
+def minor_edit_label(region, language):
+    """
+    This tag returns the label of the minor edit field of the given form
+
+    :param region: current region
+    :type region: ~integreat_cms.cms.models.regions.region.Region
+
+    :param language: The current language
+    :type language: ~integreat_cms.cms.models.languages.language.Language
+
+    :return: The minor edit label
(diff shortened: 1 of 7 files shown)


integreat_cms/cms/templatetags/content_filters.py
```diff
@@ -83,6 +83,28 @@ def get_language(language_slug):
...
+
+    :return: The minor edit label
+    :rtype: str
+    """
+    language_node = region.language_node_by_slug[language.slug]
+    if language_node.is_leaf():
+        return _("Implications on the translation status")
+    if language_node.is_root():
+        return _("Implications on other translations")
+    return _("Implications on the status of this and other translations")
```
ulliholtgrave, line 105: Is this case really necessary. Do we have some stand-alone language nodes?

timoludwig: Not in the production database, but our whole model is built on the assumption that we might have language tree nodes that are neither root nodes nor leaf nodes.

ulliholtgrave (approved):
Thanks, looks good :+1:

PR #1582, opened 2022-06-30
Add testing documentation

timoludwig:
### Short description

Add basic information about testing to our documentation

### Proposed changes

- Add basic info about pytest
- Add hint how to change API url in web app


sphinx/testing.rst
```diff
@@ -0,0 +1,30 @@
...
+
+After each run, the test coverage is uploaded to `CodeClimate <https://codeclimate.com/github/digitalfabrik/integreat-cms>`__ (see :ref:`circleci-upload-test-coverage`).
+
+
+Test API with WebApp
+====================
+
+To test the API in the web app with a different CMS server, open the JavaScript console of the web app and execute::
+
+    window.localStorage.setItem('api-url', 'your test api url')
```
ulliholtgrave, line 30: Could you add a example for the usage? E.g.

`window.localStorage.setItem('api-url',` 'http://localhost:8000')`

timoludwig: That's probably a bad example since localhost is not allowed by the CSP policy, but yes, I can add an example.

ulliholtgrave (approved):
Thanks :+1:

Issue #1464, opened 2022-05-21
Minor edit with only one public version breaks translation status


osmers:
### Describe the Bug

When I publish a translation (from draft to published) with the tick set at small edit, the status is not updated
### Steps to Reproduce

1. Go to a page with translation (in draft status)
2. Click on any translation 
3. Change something
4. Put the tick at small edit
5. Press update/publish

### Expected Behavior

The current version, as indicated in the timeline, should be the one which is published. If the status was draft, the status should change to published after clicking publish. 

### Actual Behavior

In the case of a draft page, the status is not changed.

osmers:
I can try and make a video to show the behaviour
Possible solution in the case of updating a draft page - automatically disable/ignore the small edit check box once published is pressed as generally the intention upon pressing publish should be that the page is published.
In the case of an already published page, the check box should not affect the delivered version.

MizukiTemma:
I have tried to reproduce the bug in the following steps:
1. create a new page and publish it (one public version, version 1).
2. make some changes and save it as draft (translation in draft status, version2).
3. make some changes and publish it as minor edit (version3).

All those are done in the german version.

The translation version3 is marked as published and also as the version shown in the app.
Though "translation missing" is shown.

Is this "translation missing" the problem?

The status change between draft and published seems to be working.

timoludwig:
> The translation version3 is marked as published and also as the version shown in the app. Though "translation missing" is shown.
> 
> Is this "translation missing" the problem?

Yes, I think this is essentially the same problem as in the initial issue description.
Another way to reproduce the bug is to create a page as public, then without making changes click on "minor edit" and "update" - then, it's still only one version, which is both public and minor, which seems to break the status.

MizukiTemma:
Ah, ok. Thank you. Now I got it :)

svenseeberg:
@osmers we need to revisit this. Initially you stated:

> When I publish a translation (from draft to published) with the tick set at small edit, the status is not updated

However, this is exactly what the "minor edit" function is about. If you want to update the translation status, the "minor edit" should not be checked.

Maybe we need to rename "minor edit" to "ignore in translation status" ?

timoludwig:
> Whether a page is a draft or published should have no influence on the state of a translation.

That's not entirely correct, since we defined draft versions as "missing" translation because they are not shown in the apps.
[...]

PR #1588, opened 2022-07-05
Fix translation status of minor edit

MizukiTemma:
### Short description

This PR fixes the translation status of a minor publishing.

### Proposed changes

- <s>takes not only public major versions but also public minor version into consideration when checking the translation status.</s>
- When there is no major public version, check if there is a minor public version. If exists, change the translation status to up-to -date.
- No behavior change when there is at least one major public version.

### Resolved issues

Fixes: #1464

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ UNRELEASED
 * [ [#1512](https://github.com/digitalfabrik/integreat-cms/issues/1512) ] Deliver location names in the api in the default language only
 * [ [#1581](https://github.com/digitalfabrik/integreat-cms/issues/1581) ] Improve wording of minor edit label
 * [ [#1580](https://github.com/digitalfabrik/integreat-cms/issues/1580) ] Improve user list
+* [ [#1504](https://github.com/digitalfabrik/integreat-cms/issues/1504) ] Keep filters on pagination
 
 
 2022.6.3
--- a/integreat_cms/cms/forms/custom_filter_form.py
+++ b/integreat_cms/cms/forms/custom_filter_form.py
@@ -1,0 +1,46 @@
+from django import forms
+
+
+class CustomFilterForm(forms.Form):
+    """
+    Base class for filter forms
+    """
+
+    def __init__(self, **kwargs):
+        r"""
+        Initialize the custom filter form
+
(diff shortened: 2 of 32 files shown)


MizukiTemma:
Thank you for the comment.

I modified the changes as shown in the table below and the minor public version will be checked only when there is no public version.

casetable

timoludwig (commented):
Thanks a lot for your effort! :muscle: 

> I modified the changes as shown in the table below and the minor public version will be checked only when there is no major public version.

In general, this sounds reasonable. I just have concerns about the following edge case:

1. Start with the default test data (German is source language, English depends on it), status of page "Willkommen" is up to date for all languages
2. Mark the English translation as draft -> now the translation state is "missing"
3. Perform major changes in German and publish -> now the other translations are outdated, English is still "draft"/"missing" but the content is outdated
4. Perform a minor change in English (e.g. change a phone number) and publish -> the translation state is now "up to date" even if the content is still outdated (and should still be recognized as outdated since the last change was only minor)

Since the current bug is also just an edge case, I think we should not create other broken edge cases when fixing this.

Unfortunately, the topic is very complex, so I'm not 100% sure what the best solution would be here. Maybe we could also leave the translation status calculation as it is (with just ignoring minor versions for the translation state like it was originally intended) and just prevent the minor/public combinations that don't make sense.
For example, we could also add validation to the page translation form which makes impossible to mark the first version of a translation as minor (assuming that the initial content always is "relevant" for other languages as well) and we could e.g. mark all past version as public when publishing a page (similar to what we do when saving a page as draft) to make sure that the previous version of a minor edit is always public when the minor edit is published...
@svenseeberg what are your thoughts on this?

svenseeberg:
> For example, we could also add validation to the page translation form which makes impossible to mark the first version of a translation as minor (assuming that the initial content always is "relevant" for other languages as well) and we could e.g. mark all past version as public when publishing a page (similar to what we do when saving a page as draft) to make sure that the previous version of a minor edit is always public when the minor edit is published...

Yes, you're totally right. The first version should never allow the minor edit flag. That just makes no sense.

Regarding changing the status of past versions, I'm not 100% sure. I think we would have to change the button caption in the history to "Publish this version" or similar to ensure that nobody expects that returning to an old version would bring back a draft status?

Issue #1585, opened 2022-07-04
News are not disabled/unpublished after 28 days


dkehne:
In Castrop-Rauxel there is still a news enabled that is older that 28days so process does not seem to work...
![Unbenannt2](https://user-images.githubusercontent.com/4182469/177116605-0f866212-67fb-4891-afcd-580202cea76e.PNG)

PR #1598, opened 2022-07-16
Hide news after 28 days

timoludwig:
### Short description

Hide news after 28 days

### Proposed changes

- Add new setting `FCM_HISTORY_DAYS`
- Filter news in api endpoint that are older than the configured value

### Resolved issues

Fixes: #1585


ulliholtgrave (approved):
Thanks, that works as expected :+1: It's a little bit sad that we'll loose the format test for push notifications, but I guess adding some dynamic PN based of the current date might be a little bit too much effort for now.

PR #1602, opened 2022-07-18
Exclude autosaves and not yet reviewed versions from XLIFF export

### LINKED ISSUE #1599: Exclude autosaves and not yet reviewed versions from XLIFF export
### Describe the Bug

~While exporting XLIFFs, you have the choice to explicitly export only published sites. However it exports also automatically saved pages. These are not shown in the App and are not published sites. These should not be exported.~
When exporting all pages as XLIFF, also auto saved pages and those in review are included, even if only public versions and drafts should be included.

### Steps to Reproduce

1. Go to 'Deutschlandvorlage'
2. Click on 'Sites'. Change the Language in the drop down Menu to French. Apply the Filter "Translation is old" or "Translation does not exist"
3. Scroll down to 'the drop down menu'. Choose "Export all pages for the translations into French".
4. The Export is generated.

### Expected Behavior

The export should have all pages that do not have a translation or where the translation is old, but only if the status is either "PUBLIC" or "DRAFT", not "AUTOSAVE" or "IN_REVIEW".

Also add test data for those two status and make sure they don't end up in the exported test XLIFF files.

### Actual Behavior

In the export are automatically saved pages as well as those in review.

### Additional Information

Bild eingefügt am 2022-7-18 08-38

svenseeberg:
### Short description
Change the "all pages" XLIFF export so that only translations with draft or public status are being exported.

### Proposed changes
- Add new methods to the abstract content models that return/cache/filter page translations with status draft or public.
- Change the XLIFF export bulk action for exporting all pages to use these new methods.

### Resolved issues
Fixes: #1599


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ UNRELEASED
 * [ [#1504](https://github.com/digitalfabrik/integreat-cms/issues/1504) ] Keep filters on pagination
 * [ [#1585](https://github.com/digitalfabrik/integreat-cms/issues/1585) ] Hide news after 28 days
 * [ [#1600](https://github.com/digitalfabrik/integreat-cms/issues/1600) ] Improve XLIFF export bulk option description
+* [ [#1599](https://github.com/digitalfabrik/integreat-cms/issues/1599) ] XLIFF only includes draft and public translations
```
svenseeberg, line 12: Maybe `XLIFFs never include translations in review or auto saved` is explaining it better?

timoludwig (approved):
@svenseeberg Thanks a lot and sorry it has taken me so long to get to review this.

Since this has been stale for so long, I took the liberty of rebasing the branch onto the current develop branch and directly added my feedback:
1. I reduced a bit of duplicated code fragments for the prefetched translation methods
2. I noticed that you only excluded autosaves and pending reviews for the source translation - however, I think it also makes sense to exclude them in the target translation as well:
  https://github.com/digitalfabrik/integreat-cms/blob/9ee98e572e3c6ca892b0d1563a711e39d6bca439/integreat_cms/xliff/utils.py#L142-L144

If you're fine with these changes, I think this PR is finally ready to be merged! :tada:

svenseeberg:
Thanks, looks good to me.

Issue #1596, opened 2022-07-14
Ignore missing default translations in translation coverage


timoludwig:
### Motivation

When the translations in the default language are either "in review" or "draft", the translation status is considered "missing".
This is confusing in the translation coverage view, since we only want to know how many translations need to be translated by professional translators, not how many pages need content revisions.

### Proposed Solution

Ignore missing translations in the default language in the translation coverage view.
These translations should also be ignored in all other languages and not counted as missing/outdated words.

dkehne:
Please speak to UI/UX to solve it useful. I like the idea more of Timo, than dropping of the source language or put in any more text. 
it should be self-explaining...

svenseeberg:
> it should be self-explaining...

It always should be. It never is, and we need to discuss how to get there.

Issue #1440, opened 2022-05-17
Add missing translation to translation report


ulliholtgrave:
### Motivation

Besides the outdated translations we also want to see how many words are missing translations in each language.

### Proposed Solution

Add counter of missing translations and extend the table on the translation coverage.

osmers:
We advocate for implementing this feature sooner rather than later, especially if the effort is low. Daniel went to check the WP system today to get a number for a translation cost estimate. 
The question remains whether outdated and missing translations should be combined in one figure or could be shown in two columns or something...what do you think?

PR #1607, opened 2022-07-26
Refactor translation coverage and add missing translations to word count

ulliholtgrave:
### Short description

This PR updates the UI of our translation report and adds the word calculation for missing translations.

### Proposed changes

- Change axis of chart to scale vertically with additional languages
- Add word column for missing translations
- Downgrade postcss/autoprefixer to get rid of deprecation warning

### Resolved issues

Fixes: #1440
Fixes: #1596

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ UNRELEASED
 * [ [#1566](https://github.com/digitalfabrik/integreat-cms/issues/1566) ] Manage organizations per region
 * [ [#1440](https://github.com/digitalfabrik/integreat-cms/issues/1440) ] Add missing word count to translation coverage report
 * [ [#1596](https://github.com/digitalfabrik/integreat-cms/issues/1596) ] Ignore default language in translation report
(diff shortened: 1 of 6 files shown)


timoludwig (approved):
Thanks a lot! :muscle: 

In general I like the changed layout of the chart and the added counter! :+1:
I found two problems:
1. The default language was excluded, but the missing translations in the default language were not. So all translations that were in "draft" state for the default language still counted as outdated/missing in the other languages. I fixed this in da54176af4d5ac4c57f78f0065d4fd692ea89a36.
2. I think we should count the words based on the source language instead of the target language. In other words, if in the beginning we have two words per translation (and all of them are up to date), and then we add 100 words in the default translation... in this case, I would expect the word counter to show 100 words (that need to be translated) instead of 2 words in the outdated translation (since the translation is outdated, it is irrelevant how many words it contains, right?)... I tried to address this in 7790991fe3fa376e6940e967c0727d3085f5e24f, let me know what you think.

If you're fine with my changes, I think this PR can be merged! :tada:

Issue #1571, opened 2022-06-23
Show "offline downloads" in statistics


dkehne:
### Motivation
We promised municipalities that due to the migration there will be no disadvantage/major change related to the statistics. Currently we don't have a change in collection of the data but in how we visualize them. The hits from the native app where separated and summarized in a single coloumn "Offline Downloads". These hits were subtracted from the language-specific-web-app-page-hits.

### Proposed Solution

Filter API hits by `children` and  `pages` endpoints. All hits on the `children` endpoint are single page downloads (see #1097), while all hits on the `pages` endpoint are "offline downloads" (by the native app).

The data should be available in the graph and in the CSV export.

### Additional Context

This was already implemented in https://github.com/digitalfabrik/helper-bash-scripts/blob/master/matomo/statistics.py#L168

PR #1609, opened 2022-07-26
Add visualization of offline downloads to the statistics

ulliholtgrave:
### Short description

This PR adds the hits from the mobile applications to our statistic and export functions.

### Proposed changes

- Add specific async call to our matomo requests

### Resolved issues

Fixes: #1571

### DIFF
--- a/integreat_cms/cms/utils/matomo_api_manager.py
+++ b/integreat_cms/cms/utils/matomo_api_manager.py
@@ -28,7 +28,7 @@ class MatomoApiManager:
     This class helps to interact with Matomo API.
     There are three functions which can be used publicly:
 
-    * :func:`~integreat_cms.cms.utils.matomo_api_manager.MatomoApiManager.get_matomo_id`: Retrieve the Metomo ID belonging to the given Matomo access token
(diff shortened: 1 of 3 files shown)


integreat_cms/cms/utils/matomo_api_manager.py
```diff
@@ -336,6 +344,14 @@ def get_visits_per_language(self, start_date, end_date, period):
                     # In this case, it matches the languages to their respective dataset (because the datasets are ordered)
                     for language, dataset in zip(languages, datasets)
                 ]
+                # The dataset for offline downloads
+                + [
+                    {
+                        "label": _("Offline Downloads"),
+                        "borderColor": colors.DEFAULT,
```
timoludwig, line 351:
```suggestion
                        "borderColor": next(color_cycle),
```
I would separate the offline download from the sum of all languages by using another color, so either choosing a random one from the dict or pre-defining a new constant for the offline downloads...

timoludwig (commented):
Thanks a lot! :+1:
In general, the calculation works as expected.
I'm just unsure whether we're now duplicating some requests?
If we count `/de/wp-json` for e.g. German and `pages` for the offline downloads, aren't the German offline downloads counted twice? Or am I missing something?

timoludwig (approved):
Thanks! :+1:
Now, I question the issue itself, but at least this PR solves it :sweat_smile:

Issue #1535, opened 2022-06-08
Event API slow if many recurring events exist


timoludwig:
### Describe the Bug

If many recurring events exist, the event API becomes very slow, e.g.:
https://integreat.app/muenchen/de/events

### Steps to Reproduce

1. Create many recurring events
2. Load events in app

### Expected Behavior

Since calculating all occurrences takes a lot of time, we should cache the results for a given time

### Actual Behavior

No caching makes the API very slow.

PR #1611, opened 2022-07-27
Fix slow recurring event api

david-venhoff:
### Short description
This pr fixes the performance problem regarding the events api.

### Proposed changes
- Don't create new events for recurrences
- Create unique slugs using the `$`-character, which is not allowed in 'normal' slugs

### Resolved issues

Fixes: #1535

### DIFF
--- a/integreat_cms/api/v3/events.py
+++ b/integreat_cms/api/v3/events.py
@@ -2,8 +2,6 @@
 This module includes functions related to the event API endpoint.
 """
 
-from copy import deepcopy
-
 from datetime import timedelta
 
 from django.conf import settings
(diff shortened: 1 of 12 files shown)


integreat_cms/api/v3/events.py
```diff
@@ -59,14 +68,49 @@ def transform_event_translation(event_translation):
...
+    languages = {}
+
+    for translation in event_translation.foreign_object.available_translations():
+        if translation.language == event_translation.language:
+            continue
+
+        slug = f"{translation.slug}${recurrence_date}"
+        path = translation.url_prefix + slug + "/"
+        languages[translation.language.slug] = {
+            "id": translation.id,
```
timoludwig, line 106:
```suggestion
            "id": None,
```

timoludwig (commented):
Thanks a lot! :muscle: 
Sorry that it took me so long to review this PR.
This seems to fix the issue :+1:
I would just prefer that the id field of non existing objects would return `None`/`null`...

timoludwig (approved):
Thanks a lot! :pray:

PR #1612, opened 2022-07-27
Limit number of recurring events returned from the api

david-venhoff:
### Short description

This pr limits the number of recurring events returned from the events api until a better solution is implemented.
This change is because the app struggles with many events.

### Proposed changes

- Add a setting to control the maximum number of recurring events, currently 30
- Limit number of recurring  events returned


david-venhoff:
I will update the new failing tests for this after #1611 gets merged in order to avoid updating them twice

PR #1613, opened 2022-07-29
Fix status of minor edit translation with only one public version

MizukiTemma:
### Short description

This PR is to solve the problem that the status of translations with only single public version is broken when the public version is minor edit.

This is a successor of PR https://github.com/digitalfabrik/integreat-cms/pull/1588 .

### Proposed changes

- If the translation is the first version, save it as major version.
- Otherwise, change the past versions to public. 
- No change in status calculation.

### Resolved issues

Fixes: #1464

### DIFF
--- a/integreat_cms/cms/views/pages/page_form_view.py
+++ b/integreat_cms/cms/views/pages/page_form_view.py
@@ -288,6 +288,30 @@ class PageFormView(
                 page_translation_form.instance.page.translations.filter(
                     language__in=languages
                 ).update(status=status.DRAFT)
+            # If the translation is minor change and there is no major public version
+            if (page_translation_form.instance.minor_edit is True) and (
+                not page_translation_form.instance.major_public_version
+            ):
+                # and if it is the first version, save it as major public
+                if page_translation_form.instance.version == 1:
+                    page_translation_form.instance.minor_edit = False
+                    page_translation_form.save()
+                    messages.info(
+                        request,
+                        _(
+                            'Page "{}" was saved as major version. The first version is not allowed to be a minor edit.'
+                        ).format(page_translation_form.instance.title),
+                    )
+                # or if there are only draft and/or minor public version, change them to public major version
+                else:
+                    language_tree_node = region.language_node_by_slug.get(language.slug)
+                    languages = [language] + [
+                        node.language for node in language_tree_node.get_descendants()
+                    ]
+                    page_translation_form.instance.page.translations.filter(
+                        language__in=languages
+                    ).update(status=status.PUBLIC)
+
             # Add the success message and redirect to the edit page
             if not page_instance:
                 messages.success(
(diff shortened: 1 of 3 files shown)


integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -6810,7 +6810,15 @@ msgstr ""
 "Sie haben nicht die nötige Berechtigung, um diese Seite zu veröffentlichen, "
 "aber Sie können Änderungen vornehmen und diese zur Überprüfung vorlegen."
 
-#: cms/views/pages/page_form_view.py:295
+#: cms/views/pages/page_form_view.py:302
+msgid ""
+"Page \"{}\" was saved as major version. The first version is not allowed to "
+"be a minor edit."
+msgstr ""
+"Seite \"{}\" wurde als NICHT-geringfügige Änderung gespeichert. Die erste "
```
svenseeberg, line 6818: I'd stick to the English original "Seite wurde als große Änderung gespeichert."

integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -288,6 +288,30 @@ def post(self, request, *args, **kwargs):
...
+                            'Page "{}" was saved as major version. The first version is not allowed to be a minor edit.'
+                        ).format(page_translation_form.instance.title),
+                    )
+                # or if there are only draft and/or minor public version, change them to public major version
+                else:
+                    language_tree_node = region.language_node_by_slug.get(language.slug)
+                    languages = [language] + [
+                        node.language for node in language_tree_node.get_descendants()
+                    ]
+                    page_translation_form.instance.page.translations.filter(
```
svenseeberg, line 311: Just a thought: we could only change only one version instead of all versions. What is better? And why?

integreat_cms/locale/de/LC_MESSAGES/django.po
MizukiTemma: done!

integreat_cms/cms/views/pages/page_form_view.py
MizukiTemma: I did it so as the case of marking a version to draft, where all the past versions are changed to draft and thoght it avoids other edge cases.

svenseeberg: Alright. No further objections/questions :)

MizukiTemma:
I think it's better to release the fix of this edge case first.

The get method is large too, so we can probably have an independent issue for refactoring of this view.

svenseeberg (approved):
Works as expected!

timoludwig:
> Alternatively, we can squash merge here.

I disabled squash merging for this repo because I thought it might be good to use one consistent merging strategy... And squashing the commits manually takes ~30 seconds, so I hope this process doesn't slow down our development process, but obviously this is open for discussion 😅

integreat_cms/cms/views/pages/page_form_view.py
timoludwig: Well, to me this also doesn't make sense. Now, if there are only draft versions and you save another version as minor draft, all versions are made public?! And the message is still "The page was successfully saved as draft", even if it is now public. I guess I will open an issue for it.

svenseeberg: https://github.com/digitalfabrik/integreat-cms/issues/1622 for reference

Issue #1549, opened 2022-06-14
Bundled XLIFF export via pop-up


ulliholtgrave:
### Motivation

Currently, when exporting the same pages into multiple languages, you have to click through all the languages and select the pages again. This is very time consuming and should be simplified.

### Proposed Solution

In addition to the existing export functions, we want another function for Integreat employees. So that they don't have to click through all the languages and select all the pages again for exports, there should be another bulk function that only Integreat employees see. This function should open a pop-up that provides an overview of all languages and their respective checkboxes. There, the user can then mark the languages into which the selected pages are to be translated.

PR #1614, opened 2022-07-30
Add bulk action for multilingual XLIFF export

membralala:
### Short description

This PR adds a bulk action and a pop-up window for XLIFF export in multiple languages. 

### Proposed changes

- Add bulk action option
- Add a language selection and confirmation pop-up 

### Resolved issues

Fixes: #1549

### DIFF
--- a/integreat_cms/cms/templates/pages/_page_xliff_export_overlay.html
+++ b/integreat_cms/cms/templates/pages/_page_xliff_export_overlay.html
@@ -1,0 +1,47 @@
+{% load i18n %}
+
+<div id="xliff_export_overlay" class="hidden fixed items-center justify-center inset-0 bg-opacity-75 bg-gray-800 z-50 cursor-pointer">
+    <div class="cursor-auto justify-center w-[400px] max-h-fit px-[10px] z-50 m-auto">
+        <div class="bg-white opacity-100 content rounded shadow-md w-full">
+            <div class="flex items-center w-full rounded p-4 bg-water-500">
+                <h2 class="font-bold font-default">{% trans 'Multilingual XLIFF export' %}</h2>
+                <button type="button" id="btn-close-xliff-export" class="flex ml-auto rounded-full hover:bg-blue-500 hover:text-white">
+                    <i data-feather="x-circle" class="h-8 w-8"></i>
+                </button>
+            </div> 
+            <div class="w-full p-4 rounded shadow">
+                <h3>{% trans 'Please select languages for multilingual export.' %}</h3>
+                <table class="mt-4">
+                    <tbody>
+                        {% for lang in languages %}
+                            <tr class="h-[38px]">
+                                {% if lang == request.region.default_language %}
+                                    <td></td>
+                                    <td class="text-gray-400" title="{% trans 'You cannot export XLIFF files for the default language' %}"> {{ lang.translated_name }} </td> 
+                                    <td>
+                                        <span class="left-2 fp fp-rounded fp-{{ lang.primary_country_code }}" title="{{ lang.translated_name }}"></span>
+                                    </td>  
+                                {% else %}
+                                    <td> 
+                                        <input id="xliff_export_select_lang_{{ lang.slug }}" type="checkbox" form="bulk-action-form" name="selected_language_slugs[]" value="{{ lang.slug }}" class="bulk-select-language"/>
+                                    </td>
+                                    <td><label for="xliff_export_select_lang_{{ lang.slug }}">{{ lang.translated_name }}</label></td> 
+                                    <td>
+                                        <label for="xliff_export_select_lang_{{ lang.slug }}"><span class="left-2 fp fp-rounded fp-{{ lang.primary_country_code }}" title="{{ lang.translated_name }}"></span></label>
+                                    </td> 
+                                {% endif %}
+                            </tr>
+                        {% endfor %}
+                    </tbody>
+                </table>
+                <div class="font-bold mt-4">
+                    <input id="select-all-languages" type="checkbox"/>
+                    <label for="select-all-languages">{% trans 'Select all languages.' %}</label>
+                </div>
+                <div class="flex justify-end"> 
+                    <button id="xliff-overlay-bulk-action-execute" class="btn" form="bulk-action-form" type="submit" disabled>{% trans 'Execute' %}</button>
+                </div>
+            </div> 
+        </div>
+    </div>
+</div>
--- a/integreat_cms/cms/templates/pages/page_tree.html
+++ b/integreat_cms/cms/templates/pages/page_tree.html
@@ -180,6 +180,12 @@
                             {% endif %}
                         {% endwith %}
                     {% endif %}
+                    {% if request.user.is_superuser or request.user.is_staff %}
+                        <option id="xliff-multiple-languages-option" data-show-xliff-export-overlay
+                            data-bulk-action="{% url 'download_xliff_multiple_languages' region_slug=request.region.slug language_slug=language.slug %}">
+                            {% trans 'Export selected pages for multilingual translation' %} ({{ XLIFF_EXPORT_VERSION|upper }})
+                        </option>
+                    {% endif %}
                 </select>
                 <button id="bulk-action-execute" class="btn" disabled>{% trans 'Execute' %}</button>
             </div>
@@ -213,4 +219,5 @@
         {% include "../generic_confirmation_dialog.html" %}
         {% include "../tutorials/page_tree.html" with tutorial_id="page-tree" hidden=request.user.page_tree_tutorial_seen %}
         {% include "pages/_page_preview_overlay.html" %}
+        {% include "pages/_page_xliff_export_overlay.html" %}
     {% endblock content %}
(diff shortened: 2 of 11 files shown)


integreat_cms/cms/views/pages/page_bulk_actions.py
```diff
@@ -118,3 +121,62 @@ def post(self, request, *args, **kwargs):
...
+        :rtype: ~django.http.HttpResponseRedirect
+        """
+
+        target_languages = [
+            get_object_or_404(
+                self.request.region.language_tree_nodes,
+                language__slug=language_slug,
+                parent__isnull=False,
+            ).language
+            for language_slug in self.request.POST.getlist("selected_language_ids[]")
```
timoludwig, line 158:
```suggestion
            language
            for language in self.request.region.active_languages
            if language.slug in self.request.POST.getlist("selected_language_ids[]")
```
This is a bit more complicated (and does more database queries) than needed.

integreat_cms/cms/templates/pages/page_tree.html
```diff
@@ -180,6 +180,12 @@ <h1 class="heading">
                             {% endif %}
                         {% endwith %}
                     {% endif %}
+                    {% if request.user.is_staff %}
```
timoludwig, line 183:
```suggestion
                    {% if request.user.is_superuser or request.user.is_staff %}
```
In theory, the user model allows superusers to not be staff members, so for consistency, I'd also check superuser together with the staff flag everywhere it's used.

integreat_cms/static/src/js/pages/xliff-export-overlay.ts
```diff
@@ -0,0 +1,83 @@
...
+    selector.addEventListener("change", () => {
+      let val = selector.getAttribute("data-value");
+      let lastVal = selector.getAttribute("data-last-value");
+        selector.setAttribute("data-value", selector.value); 
+      if (val != lastVal) {
+        selector.setAttribute("data-last-value", val); 
+      }
+    });
+  }
+}
```
timoludwig, line 83: I don't understand the JS part completely to be honest and feel like it is more complicated than needed, but other than that: Could you open the selection window only when the action is actually executed, not when the option is selected from the dropdown?
(I guess this would also solve the problem that it's currently possible to open the overlay when no pages are selected at all and thus the multi lang export itself cannot be executed....)

integreat_cms/cms/templates/pages/_page_xliff_export_overlay.html
```diff
@@ -0,0 +1,49 @@
...
+                            <tr class="h-[38px]">
+                                {% if lang == request.region.default_language %}
+                                    <td></td>
+                                    <td class="text-gray-400" title="{% trans 'You cannot export XLIFF files for the default language' %}"> {{ lang.translated_name }} </td> 
+                                    <td>
+                                        <span class="left-2 fp fp-rounded fp-{{ lang.primary_country_code }}" title="{{ lang.translated_name }}"></span>
+                                    </td>  
+                                {% else %}
+                                    <td> 
+                                        <input type="checkbox" form="bulk-action-form" name="selected_language_ids[]" value="{{ lang.slug }}" class="bulk-select-language"/>
```
timoludwig, line 26:
```suggestion
                                        <input type="checkbox" form="bulk-action-form" name="selected_language_slugs[]" value="{{ lang.slug }}" class="bulk-select-language"/>
```
I would rename the form field to match the values it contains (or alternatively pass the ids instead of the slugs).

timoludwig (commented):
Thanks a lot, very cool feature and works as expected! :+1:
Just a few ideas for improvement.

membralala:
Thanks for your feedback! I changed the behaviour based on your suggestions and I think you're right, it feels a bit more natural this way. Does this match your expectations as well? 🙂

integreat_cms/cms/templates/pages/_page_xliff_export_overlay.html
```diff
@@ -0,0 +1,49 @@
...
+                                {% endif %}
+                            </tr>
+                        {% endfor %}
+                    </tbody>
+                </table>
+                <div class="font-bold mt-4">
+                    <span>
+                        <input id="select-all-languages" type="checkbox"/>
+                    </span> 
+                    <span>{% trans 'Select all languages.' %}</span>
```
timoludwig, line 41:
```suggestion
                    <input id="select-all-languages" type="checkbox"/>
                    <label for="select-all-languages">{% trans 'Select all languages.' %}</label>
```
I think we could add a clickable label here as well.
Also, the `<span>`-elements without attributes don't do much.

integreat_cms/static/src/js/pages/xliff-export-overlay.ts
```diff
@@ -0,0 +1,71 @@
...
+function addXliffExportOverlayListeners(callback: (overlay: HTMLElement) => void) {
+  const overlay = document.getElementById("xliff_export_overlay");
+  if (overlay) {
+    // Set listener for bulk action execute button.
+    document.getElementById("bulk-action-execute")?.addEventListener("click", (event) => {
+      const xliffOptionIndex = (document.getElementById("xliff-multiple-languages-option") as HTMLOptionElement).index;
+      const selectedIndex = (document.getElementById("bulk-action") as HTMLSelectElement).selectedIndex; 
+      if (xliffOptionIndex === selectedIndex) {
+        event.preventDefault(); 
+        callback(overlay);
```
timoludwig, line 26: Is there a specific use case why this callback was outsourced? I cannot think of any other action one might want to do here...
```suggestion
        overlay.classList.remove("hidden");
        overlay.classList.add("flex");
```

integreat_cms/static/src/js/pages/xliff-export-overlay.ts
```diff
@@ -0,0 +1,71 @@
...
+  document.getElementById("select-all-languages")?.addEventListener("change", (toggle) => {
+    document.querySelectorAll(".bulk-select-language").forEach((el) => {
+      (el as HTMLInputElement).checked = (toggle.target as HTMLInputElement).checked; 
+      toggleXliffBulkActionButton(); 
+    });
+  });
+});
+
+
+function addXliffExportOverlayListeners(callback: (overlay: HTMLElement) => void) {
```
timoludwig, line 17:
```suggestion
function addXliffExportOverlayListeners() {
```

integreat_cms/static/src/js/pages/xliff-export-overlay.ts
```diff
@@ -0,0 +1,71 @@
+window.addEventListener("load", () => {
+  addXliffExportOverlayListeners(openXliffExportOverlay);
```
timoludwig, line 2:
```suggestion
  addXliffExportOverlayListeners();
```

integreat_cms/static/src/js/pages/xliff-export-overlay.ts
```diff
@@ -0,0 +1,71 @@
...
+      }
+    });
+  }
+}
+
+
+function openXliffExportOverlay(overlay: HTMLElement) {
+  overlay.classList.remove("hidden");
+  overlay.classList.add("flex");
+}
```
timoludwig, line 45:
```suggestion
```

integreat_cms/static/src/js/pages/xliff-export-overlay.ts
```diff
@@ -0,0 +1,71 @@
...
+  // Only activate button if at least one language item is selected
+  const selectLanguages = <HTMLInputElement[]>(
+    Array.from(document.getElementsByClassName("bulk-select-language"))
+  )
+  const multiLanguageXliffButton = document.getElementById("xliff-overlay-bulk-action-execute") as HTMLButtonElement; 
+  if (selectLanguages.some((el) => el.checked)) {
+    multiLanguageXliffButton.disabled = false;
+  } else {
+    multiLanguageXliffButton.disabled = true; 
+  }
```
timoludwig, line 70:
```suggestion
  multiLanguageXliffButton.disabled = !selectLanguages.some((el) => el.checked);
```
This expression can be simplified.

integreat_cms/static/src/js/pages/xliff-export-overlay.ts
```diff
@@ -0,0 +1,71 @@
...
+
+
+function closeXliffExportOverlay(overlay: HTMLElement) {
+  overlay.classList.add("hidden");
+  overlay.classList.remove("flex");
+  let selector = (document.getElementById("bulk-action") as HTMLSelectElement);
+  if (selector.hasAttribute("data-last-value")) {
+    selector.value = selector.getAttribute("data-last-value");
+    selector.setAttribute("data-value", selector.value); 
+  }
```
timoludwig, line 55:
```suggestion
```
The recovering of the last value isn't needed, right?

timoludwig (commented):
Thanks a lot! :+1:
Yes, this looks better! Still have a few ideas for improvement left :wink:
After that, could you resolve the conflicts and squash the commits? Thanks!

integreat_cms/static/src/js/pages/xliff-export-overlay.ts
membralala: I think, you're right!

integreat_cms/static/src/js/pages/xliff-export-overlay.ts
membralala: Yes, I missed this, it was part of the previous UI solution

timoludwig (approved):
Thanks! :rocket:

Issue #1604, opened 2022-07-20
Broken Links from old revisions are shown


dkehne:
### Describe the Bug
After cloning lkkarlsruhe to mannheim we found broken links from old revisions of pages in "draft" status. Those pages/old revision should not be crawled by the broken-link checker.

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

timoludwig:
Probably this is related to https://github.com/digitalfabrik/integreat-cms/issues/1572

timoludwig:
I was not able to reproduce this locally. When I save a new revision of a page, the old link objects are being deleted:

https://github.com/digitalfabrik/integreat-cms/blob/02a4d49bc0ec47ed0c2bb2786e9a0c61dd7dad73/integreat_cms/cms/forms/pages/page_translation_form.py#L84-L85

If anyone manages to reproduce this problem, try to describe it as detailed as possible.
Maybe this is not a problem of the code base, but a problem resulting of inconsistencies in the production database instead?
Requires further investigation before we can fix this issue.

timoludwig:
@dkehne and just to clarify: Broken links in draft translations (if they are the most recent ones) should be shown, right? Just not in the old versions of it?
E.g. if we have a translation with
- version 1: broken links A, B, X, Y
- version 2: broken links B, C, Y, Z

Then, in the broken link checker, the links B, C, Y, Z should be listed, even though the page is a draft, right? And only A and X should be excluded because they're outdated...

MizukiTemma:
I'm not sure whether this is the case which @dkehne has originally reported but found a suspicious case:

1) Publish a new page with 1 link. This link is detected by the link checker.
2) Add another link to the page and update it. Those both links are detected by the link checker.
3) Restore and publish the version with one link. The linkchecker still shows both the links.

timoludwig:
@MizukiTemma good catch! I can also reproduce this bug. This could definitely be the cause of this problem.

PR #1617, opened 2022-08-03
Show no broken links from old versions

MizukiTemma:
### Short description

This PR solves the problem that links in old versions are shown in the link checker.

### Proposed changes

-<s> Links in draft versions are not shown in the link check.</s>
-<s>Only links in public versions are detected.</s>
- This bug happens when an old version is restored, because deletion of the old links will not be called during restoration, unlike case of creating a new version.
- The old links will be now deleted at restoration too.

### Resolved issues

Fixes: #1604

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ UNRELEASED
 * [ [#1623](https://github.com/digitalfabrik/integreat-cms/issues/1623) ] Fix imprint publish/update button
 * [ [#1534](https://github.com/digitalfabrik/integreat-cms/issues/1534) ] Invalidate cache after moving nodes
 * [ [#1535](https://github.com/digitalfabrik/integreat-cms/issues/1535) ] Fix event api performance
+* [ [#1604](https://github.com/digitalfabrik/integreat-cms/issues/#1604) ] Show no broken links from restored versions
 
 
 2022.7.0
--- a/integreat_cms/cms/views/pages/page_revision_view.py
+++ b/integreat_cms/cms/views/pages/page_revision_view.py
@@ -169,7 +169,7 @@ class PageRevisionView(TemplateView):
                     "language_slug": language.slug,
                 },
             )
-
+        current_revision.links.all().delete()
         revision.pk = None
         revision.version = current_revision.version + 1
 


timoludwig:
Hmm. Yes, we could use this workaround, but it makes the link lists slower and it does not solve the original bug.
In theory, old versions are already excluded because the old Link objects are deleted when new versions are created:

https://github.com/digitalfabrik/integreat-cms/blob/02a4d49bc0ec47ed0c2bb2786e9a0c61dd7dad73/integreat_cms/cms/forms/pages/page_translation_form.py#L84-L85

I would much rather prefer if we found the original bug and solved it directly instead of this workaround...

MizukiTemma:
All right, i'll try!

timoludwig:
Oh, and I think there is a misunderstanding regarding the issue: Broken links in draft versions should be shown, not only of public versions. The ticket is about outdated versions I think, the draft status was just additional information to relate the problem to the production database.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ UNRELEASED
 * [ [#1464](https://github.com/digitalfabrik/integreat-cms/issues/1464) ] Fix status of translation with only minor public version
 * [ [#1623](https://github.com/digitalfabrik/integreat-cms/issues/1623) ] Fix imprint publish/update button
 * [ [#1534](https://github.com/digitalfabrik/integreat-cms/issues/1534) ] Invalidate cache after moving nodes
+* [ [#1604](https://github.com/digitalfabrik/integreat-cms/issues/#1604) ] Show no broken links from draft/old versions
```
timoludwig, line 10:
```suggestion
* [ [#1604](https://github.com/digitalfabrik/integreat-cms/issues/#1604) ] Show no broken links from restored versions
```

timoludwig (commented):
Could you apply the same changes for the imprint revision view?

MizukiTemma:
> Could you apply the same changes for the imprint revision view?

Link check is not applied for imprint right now.
Should it be implemented now also for imprint?

timoludwig:
> Link check is not applied for imprint right now. Should it be implemented now also for imprint?

Ah ok sorry, no then it's fine. Then please just resolve the conflict and squash the commits :slightly_smiling_face:

timoludwig (approved):
:+1:

Issue #1534, opened 2022-06-08
Cache not invalidated correctly when LanguageTreeNodes are moved


timoludwig:
### Describe the Bug

When language tree nodes are moved, the pages/events/pois of regions are still incorrectly cached, leading to incorrect translation status icons for each language column.

### Steps to Reproduce

1. Enable redis cache
2. Change order of language tree nodes
3. See old oder in page tree

### Expected Behavior

The page transaltions should now be in the new order.
The cache should be invalidated when language tree nodes are moved.

### Actual Behavior

They are still in the cached order

### Additional Information

This is essentially the same bug as https://github.com/digitalfabrik/integreat-cms/issues/1403 and https://github.com/digitalfabrik/integreat-cms/issues/1487, so the solution can just be copied from there.

Issue #1618, opened 2022-08-03
Investigate why the standard cacheops invalidation procedure isn't triggered


ulliholtgrave:
### Describe the Bug

We seem to have a problem with cacheops not invalidating our cache after we delete or update our objects (e.g. language nodes). To mitigate this behavior we manually invalidate the cache (see #1534), but it would be better to get the in-built function working.

### Additional Information

See cacheops documentation: https://github.com/Suor/django-cacheops

ulliholtgrave:
Right now I can think of three possible problems:

-     The invalidation is handled asynchronous and opening the page tree right after the deletion/movement of the nodes results in a display of the not-yet invalidated state
-     Our inheritance model isn't handled correctly by cacheops
-     The caching of the views isn't correctly invalidated

timoludwig:
I'm not sure whether we really have a general problem here. The way I understand it is that the cacheops invalidation is working exactly as documented. The problem is just that cacheops cannot know that e.g. the pages are affected from changes of the language tree, since the pages are only linked to the language model and not to the language tree node model. So I guess for this specific case we need the manual invalidation anyway and for normal cases the cacheops invalidation works fine. If we notice other edge cases where the standard invalidation doesn't work, we should open separate issues for it...

ulliholtgrave:
Okay, I understand your problem with the issue. I somehow would prefer a more elegant way than iterating through all objects, but I guess we can close the issue.

PR #1619, opened 2022-08-03
Invalidate cache after moving nodes

ulliholtgrave:
### Short description

This PR invalidates the cache of all lanuage-related objects after a language node is moved.

### Proposed changes

- Export the manual invalidation into a separate function
- Use the function after moving and deleting language nodes

However, our cache behavior is still pretty strange and I opened issue #1618 for a deeper investigation to get rid of this workarounds.

### Resolved issues

Fixes: #1534

### DIFF
--- a/integreat_cms/cms/views/language_tree/language_tree_actions.py
+++ b/integreat_cms/cms/views/language_tree/language_tree_actions.py
@@ -62,6 +62,7 @@ def move_language_tree_node(
         # (The move()-method executes raw sql which might cause problems if the instance isn't fetched again)
         language_tree_node = LanguageTreeNode.objects.get(id=language_tree_node_id)
         language_tree_node.save()
+        manually_invalidate_models(region)
         messages.success(
             request,
             _('The language tree node "{}" was successfully moved.').format(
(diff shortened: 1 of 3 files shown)


integreat_cms/cms/views/language_tree/language_tree_actions.py
```diff
@@ -146,3 +140,21 @@ def delete_language_tree_node(request, region_slug, language_tree_node_id):
         ).format(language_node.translated_name),
     )
     return redirect("language_tree", **{"region_slug": region_slug})
+
+
+def manually_invalidate_models(region):
+    """
+    This is a helper function to iterate through all affected objects and invalidate their cache.
+    This is somehow necessary as the original cache invalidation of cacheops doesn't seem to trigger.
```
timoludwig, line 148:
```suggestion
    This is necessary as the original cache invalidation of cacheops only triggers for direct foreign key relationships.
```

timoludwig (commented):
Thanks a lot! :+1:

timoludwig (approved):
:rocket:

Issue #1573, opened 2022-06-24
Show diff to last version in side-by-side view


FritjofK:
### Motivation

Integreat's content are updated over time and the translations need to follow. Sometimes there is a quite a lot of time between changes of the same page and the information on a page is quite long. 
It would be helpful for translators working directly in the system to get a view on the changes since the last translation was done. 

### Proposed Solution

Compare time stamp of last source language change and translation change and then show all changes in source language since that point in time? 

### Alternatives

Show the version history on the translation view and maybe highlight the time of the last change that was translated in the version history slider? 

### Additional Context

This is how it was done in WordPress. By clicking on "Show differences" in the translation view, the changes are shown in HTML-view.

![Integreat_WordPress_Screenshots from show differences since last translation_2](https://user-images.githubusercontent.com/16462172/175505217-16a38e3b-08f6-4fe8-8842-9ea5339fa325.jpg)
![Integreat_WordPress_Screenshots from show differences since last translation_1](https://user-images.githubusercontent.com/16462172/175505220-49954033-2e47-49fe-af6a-b9ce8d9f4151.jpg)

FritjofK:
Question came from pilot on Lesvos where they had translators working directly in the system. In Germany this (translators working within Integreat with the side-by-side view) happens very rarely, I think. 
In international context it might be used more often.

timoludwig:
Duplicate of #1449

Issue #1449, opened 2022-05-17
Show diff to last source version in side-by-side view


svenseeberg:
### Motivation

When updating a translation it would be helpful to see the difference in the source since the last translation.

### Proposed Solution

Add a toggle button to show the diff of the source language since the last translation in the side-by-side view.

PR #1621, opened 2022-08-03
Add diff-view for source translation in sbs-editor

david-venhoff:
### Short description

This pr adds functionality to render the diff of the source translation since the last update of the target translation in the side-by-side page editor.

### Proposed changes

- Render the diff of the source translation in sbs-view
- Add a button to toggle between the normal view and the diff view
- Remove disabled tinymce editor instance in source translation for better consistency

### Resolved issues

Fixes: #1449,  Fixes: #1573

### DIFF
--- a/integreat_cms/cms/templates/pages/page_sbs.html
+++ b/integreat_cms/cms/templates/pages/page_sbs.html
@@ -70,19 +70,38 @@
                                type="text"
                                value="{{ source_page_translation.title }}"
                                disabled>
-                        <div class="mt-4">
-                            <label class="block mt-4 font-bold">{{ page_translation_form.content.label }}</label>
-                            <textarea id="source_translation_tinymce" cols="40" rows="10" class="tinymce_textarea" disabled>{{source_page_translation.content}}</textarea>
+                        <div>
+                            <label class="block font-bold">{{ page_translation_form.content.label }}</label>
+                            <textarea 
+                                id="source_translation_tinymce" 
+                                cols="40" 
+                                rows="10" 
+                                class="tinymce_textarea" 
+                                disabled
+                                data-old="{{ old_translation_content }}" 
+                                data-new="{{ source_page_translation.content }}"
+                            >
+                                {{source_page_translation.content}}
+                            </textarea>
                         </div>
-                        <label>{% trans 'Copy content' %}</label>
-                        <button id="copy-translation-content"
-                                title="{% trans 'Copy content' %}"
-                                class="btn w-full bg-blue-500 hover:bg-blue-600 text-white font-bold py-2 px-4 rounded">
-                            {% with source_language=source_page_translation.language.translated_name target_language_name=target_language.translated_name %}
-                                {% blocktrans %}Copy content of {{ source_language }} to {{ target_language_name }}{% endblocktrans %}
-                            {% endwith %}
-                            <i data-feather="arrow-right-circle" class="mr-2"></i>
-                        </button>
+                        <label>{% trans 'Actions' %}</label>
+                        <div class="flex flex-wrap gap-4">
+                            <button id="toggle-translation-diff"
+                                    class="btn grow bg-blue-500 hover:bg-blue-600 text-white font-bold py-2 px-4 rounded"
+                                    title="{% if page_translation_form.instance.is_up_to_date %}{% trans 'No changes have been made to the source translation.' %}{% else %}{% trans 'Toggle source translation differences' %}{% endif %}"
+                                    {% if page_translation_form.instance.is_up_to_date %}disabled{% endif %}>
+                                <div class="hidden toggle">{% trans 'Hide source translation differences' %}</div>
+                                <div>{% trans 'Show source translation differences' %}</div>
+                            </button>
+                            <button id="copy-translation-content"
+                                    title="{% trans 'Copy content' %}"
+                                    class="btn grow bg-blue-500 hover:bg-blue-600 text-white font-bold py-2 px-4 rounded">
+                                {% with source_language=source_page_translation.language.translated_name target_language_name=target_language.translated_name %}
+                                    {% blocktrans %}Copy content of {{ source_language }} to {{ target_language_name }}{% endblocktrans %}
+                                {% endwith %}
+                                <i data-feather="arrow-right-circle" class="mr-2"></i>
+                            </button>
+                        </div>
                     </div>
                 </div>
                 <div class="border border-blue-500 rounded shadow-2xl">
--- a/integreat_cms/cms/views/pages/page_sbs_view.py
+++ b/integreat_cms/cms/views/pages/page_sbs_view.py
@@ -101,6 +101,10 @@ class PageSideBySideView(TemplateView, PageContextMixin):
             disabled=disabled,
         )
 
+        old_translation_content = get_old_source_content(
+            page, source_language, target_language
+        )
+
         return render(
             request,
             self.template_name,
@@ -109,6 +113,7 @@ class PageSideBySideView(TemplateView, PageContextMixin):
                 "page_translation_form": page_translation_form,
                 "source_page_translation": source_page_translation,
                 "target_language": target_language,
+                "old_translation_content": old_translation_content,
             },
         )
 
@@ -214,6 +219,10 @@ class PageSideBySideView(TemplateView, PageContextMixin):
             # Add the success message
             page_translation_form.add_success_message(request)
 
+        old_translation_content = get_old_source_content(
+            page, source_language_node.language, target_language
+        )
+
         return render(
             request,
             self.template_name,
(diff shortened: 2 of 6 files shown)


integreat_cms/cms/templates/pages/page_sbs.html
```diff
@@ -72,8 +72,20 @@ <h1 class="heading">
                                disabled>
                         <div class="mt-4">
                             <label class="block mt-4 font-bold">{{ page_translation_form.content.label }}</label>
-                            <textarea id="source_translation_tinymce" cols="40" rows="10" class="tinymce_textarea" disabled>{{source_page_translation.content}}</textarea>
```
timoludwig, line 75: The tinymce textarea had the advantage that all formatting etc. in html is already rendered and shown exactly like intended... (now, e.g. headings and links are styled differently) did it not work to show the additions/deletions in the tinymce field?
Also, I think the margins etc. should look similar on the left and right sides, to be able to better compare them.

integreat_cms/cms/views/pages/page_sbs_view.py
```diff
@@ -71,6 +71,14 @@ def get(self, request, *args, **kwargs):
         source_page_translation = page.get_translation(source_language.slug)
         target_page_translation = page.get_translation(target_language.slug)
 
+        # For the text diff, use the latest source translation that was created before the target translation
+        filter_args = {"language": source_language}
+        if target_page_translation:
+            filter_args["last_updated__lte"] = target_page_translation.last_updated
+        source_previous_translation = (
+            page.translations.filter(**filter_args).order_by("-last_updated").first()
+        )
```
timoludwig, line 80:
```suggestion
        old_translation_content = ""
        if target_page_translation:
            source_previous_translation = (
                page.translations.filter(
                    language=source_language,
                    last_updated__lte=target_page_translation.last_updated,
                )
                .order_by("-last_updated")
                .first()
            )
            if source_previous_translation:
                old_translation_content = source_previous_translation.content
```
If no target_page_translation exists, I would argue that the old content should be empty and the diff is "everything"...

integreat_cms/cms/views/pages/page_sbs_view.py
```diff
@@ -109,6 +117,7 @@ def get(self, request, *args, **kwargs):
                 "page_translation_form": page_translation_form,
                 "source_page_translation": source_page_translation,
                 "target_language": target_language,
+                "old_translation": source_previous_translation,
```
timoludwig, line 120:
```suggestion
                "old_translation_content": old_translation_content,
```

integreat_cms/cms/templates/pages/page_sbs.html
```diff
@@ -72,8 +72,20 @@ <h1 class="heading">
...
+                            <div class="p-2 rounded border border-solid border-gray-200 shadow bg-white">
+                                <div class="toggle-diff"> {{ source_page_translation.content|safe }} </div>
+    
+                                <div 
+                                    class="toggle-diff text-diff hidden"
+                                    data-old="{{ old_translation.content }}"
```
timoludwig, line 80:
```suggestion
                                    data-old="{{ old_translation_content }}" 
```

timoludwig (commented):
Thanks a lot! :+1:
This nearly works as expected, I would just adapt the look of the source translation to the target translation to make it easier to compare both sides (so a fixed size even for very short or very long pages would be good and add a bit of margin so they are horizontally aligned). If possible, the disabled tinymce area would solve this because it already looks like the right side without any adaptions, but if this doesn't work well with the html diff, we could try to at least make it somehow similar.

integreat_cms/cms/templates/pages/page_sbs.html
david-venhoff: I implement it like this so that it is looks the same like it does in revision view, I'll try to make it work on an editor instance

integreat_cms/cms/templates/pages/page_sbs.html
```diff
@@ -70,10 +70,25 @@ <h1 class="heading">
...
+                                data-new="{{ source_page_translation.content }}"
+                            >
+                                {{source_page_translation.content}}
+                            </textarea>
                         </div>
+                        <label>{% trans 'Toggle source translation differences' %}</label>
+                        <button id="toggle-translation-diff" title="{% trans 'Toggle source translation differences' %}" class="btn w-full bg-blue-500 hover:bg-blue-600 text-white font-bold py-2 px-4 rounded">
+                            <div class="hidden toggle">{% trans 'Hide source translation differences' %}</div>
+                            <div>{% trans 'Show source translation differences' %}</div>
+                        </button>
```
timoludwig, line 91: What do you think about combining the two buttons into one label? I think it's kind of redundant to have the label say the same thing than the button text. I cannot make an automated suggestion, because the copy content button was not changed by you, but this snippet is what I would use:
(I would also make the button disabled when there are no changes in the source translation and the target translation is already up to date)
```
<label>{% trans 'Actions' %}</label>
<div class="flex flex-wrap gap-4">
    <button id="toggle-translation-diff"
            class="btn grow bg-blue-500 hover:bg-blue-600 text-white font-bold py-2 px-4 rounded"
            title="{% if page_translation_form.instance.is_up_to_date %}{% trans 'No changes have been made to the source translation.' %}{% else %}{% trans 'Toggle source translation differences' %}{% endif %}"
            {% if page_translation_form.instance.is_up_to_date %}disabled{% endif %}>
        <div class="hidden toggle">{% trans 'Hide source translation differences' %}</div>
        <div>{% trans 'Show source translation differences' %}</div>
    </button>
    <button id="copy-translation-content"
            title="{% trans 'Copy content' %}"
            class="btn grow bg-blue-500 hover:bg-blue-600 text-white font-bold py-2 px-4 rounded">
        {% with source_language=source_page_translation.language.translated_name target_language_name=target_language.translated_name %}
            {% blocktrans %}Copy content of {{ source_language }} to {{ target_language_name }}{% endblocktrans %}
        {% endwith %}
        <i data-feather="arrow-right-circle" class="mr-2"></i>
    </button>
</div>
```

timoludwig (commented):
Amazing, really cool that the styling works in the tinymce area! :+1:
Now I just have a small idea for the button's appearance.

timoludwig (commented):
Awesome, thanks a lot for your changes! :+1:

Almost looks perfect, but sorry, another thought just crossed my mind: When the diff is viewed in the source editor, and then the source content is copied into the target editor, the complete diff including the red/green formatting is copied, which is probably not what was intended. Do you think we could always copy the new source content, no matter whether the diff is currently shown or not?

david-venhoff:
Oh, somehow missed that...
Should be fixed now!

integreat_cms/cms/views/pages/page_sbs_view.py
```diff
@@ -71,6 +71,20 @@ def get(self, request, *args, **kwargs):
         source_page_translation = page.get_translation(source_language.slug)
         target_page_translation = page.get_translation(target_language.slug)
```
timoludwig, line 73:
```suggestion
        major_target_page_translation = page.translations.filter(
            language__slug=target_language.slug, minor_edit=False
        ).first()

```
I think that we should use the latest major version of the target translation to determine the correct date of the source translation... if I perform a minor edit in the target language (e.g. fix a typo), then this version should not be used to determine the diff of the source translation...
Sure, this is an edge case, but since we had a few problems with edge cases of translation states in the past, I think it would be better to fix it now instead of hoping nobody notices :see_no_evil:

integreat_cms/cms/views/pages/page_sbs_view.py
```diff
@@ -71,6 +71,20 @@ def get(self, request, *args, **kwargs):
...
 
+        # For the text diff, use the latest source translation that was created before the target translation
+        old_translation_content = ""
+        if target_page_translation:
```
timoludwig, line 76:
```suggestion
        if major_target_page_translation:
```

integreat_cms/cms/views/pages/page_sbs_view.py
```diff
@@ -71,6 +71,20 @@ def get(self, request, *args, **kwargs):
...
+            source_previous_translation = (
+                page.translations.filter(
+                    language=source_language,
+                    last_updated__lte=target_page_translation.last_updated,
```
timoludwig, line 80:
```suggestion
                    last_updated__lte=major_target_page_translation.last_updated,
```

timoludwig (commented):
Thanks a lot! :pray: 

I'm sorry about my iterative review loop, but I have just found another problem :crying_cat_face: 

Could you inject the `old_translation_content` template context in both the GET and POST view? At the moment, the diff only works correctly when requesting the form with GET, but after e.g. a submission without changes, the diff does not work correctly...

david-venhoff:
I have update the pr and the issues with minor edits should be resolved now.

timoludwig (approved):
Thank you! :+1: :100:

PR #1623, opened 2022-08-04
Fix text of imprint publish/update button

timoludwig:
### Short description

Fix text of imprint publish/update button


svenseeberg (approved):
Looks good, thanks for the quick fix.

Issue #1622, opened 2022-08-03
Minor edits are published when saved as draft


timoludwig:
### Describe the Bug

When a page is
- marked as minor edit and
- saved as draft

the current and all past versions are made public

### Steps to Reproduce

1. Go to the page form of a page
2. mark as minor edit
3. Save page as draft

### Expected Behavior

When a page is saved as draft, it should actually be saved as draft

### Actual Behavior

The page is made public

svenseeberg:
I did not notice this edge case while testing :-/ This needs to be fixed before we can deploy #1620. I'll look into this issue.

svenseeberg:
So before I start fixing this, I want to document the different states and their transitions:

- If a version is the first version, it can (obviously) never be minor.
- If a change is a minor edit, the translation status should not be changed. That basically means that it keeps being outdated, if it was outdated before. If it acts as a source for another language, then the other language should stay up to date.
- A minor change has no influence on the translation status.
- The translation status is calculated based on major versions. If the major version of a translation is newer than the source, it is up to date. If it is older, it is outdated.
- The fact whether a translation version is draft or public has no influence on the translation status.
- The history should not change (??).
-

PR #1624, opened 2022-08-04
Fix translation state when pubishing minor edit

timoludwig:
### Short description

This would be my suggestion to fix the translation state calculation

### Proposed changes

- Don't allow the first version to be a minor edit (identical to before)
- When a page is published and a minor edit, also publish the past versions to make sure the status is calculated correctly (and the minor change is actually a minor change and not a transition from a draft page to a public page)

### Resolved issues

Fixes: #1464
Fixes: #1622

### DIFF
--- a/integreat_cms/cms/views/pages/page_form_view.py
+++ b/integreat_cms/cms/views/pages/page_form_view.py
@@ -288,29 +288,28 @@ class PageFormView(
                 page_translation_form.instance.page.translations.filter(
                     language__in=languages
                 ).update(status=status.DRAFT)
-            # If the translation is minor change and there is no major public version
-            if (page_translation_form.instance.minor_edit is True) and (
-                not page_translation_form.instance.major_public_version
+            # If this is the first version and the minor edit checkbox is checked, remove it
+            if (
+                page_translation_form.instance.version == 1
+                and page_translation_form.instance.minor_edit
             ):
-                # and if it is the first version, save it as major public
-                if page_translation_form.instance.version == 1:
-                    page_translation_form.instance.minor_edit = False
-                    page_translation_form.save()
-                    messages.info(
-                        request,
-                        _(
-                            'Page "{}" was saved as major version. The first version is not allowed to be a minor edit.'
-                        ).format(page_translation_form.instance.title),
-                    )
-                # or if there are only draft and/or minor public version, change them to public major version
-                else:
-                    language_tree_node = region.language_node_by_slug.get(language.slug)
-                    languages = [language] + [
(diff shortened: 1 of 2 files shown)


integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -288,29 +288,28 @@ def post(self, request, *args, **kwargs):
                 page_translation_form.instance.page.translations.filter(
                     language__in=languages
                 ).update(status=status.DRAFT)
-            # If the translation is minor change and there is no major public version
-            if (page_translation_form.instance.minor_edit is True) and (
-                not page_translation_form.instance.major_public_version
+            # If this is the first version and the minor edit checkbox is checked, remove it
```
svenseeberg, line 291: Totally agree here. I had the same solution in mind.

integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -288,29 +288,28 @@ def post(self, request, *args, **kwargs):
...
+                page_translation_form.instance.save()
+                messages.info(
+                    request,
+                    _(
+                        'The "minor edit" option was disabled because the first version can never be a minor edit.'
+                    ).format(page_translation_form.instance.title),
+                )
+            # When a version is published and a minor edit, also publish the past versions
+            # (to make sure the translation state is updated correctly)
+            if (
```
svenseeberg, line 306: As of discussion in https://github.com/digitalfabrik/integreat-cms/issues/1464 this feels wrong to me, because we're mixing the intention of 2 settings. But if everyone feels this is good, then I won't object.

I personally feel we should definitely untangle the draft/public state from minor edits (ignore in translation status).

svenseeberg: But maybe we need to rephrase the concept of the translation status. And maybe that is something that only needs to happen in my head. But if we label the translation status to "Translation status of public version", then this sure makes sense. Or maybe we just need to differentiate between these to options: we can have public up to date translations and we can have draft up to date translations.

timoludwig: This probably needs separate clarification with the service team. But in the mean time, I guess we should leave the current behaviour (that draft pages are considered missing) because this is what we teach our users at the moment.

timoludwig: But in general I agree that there is probably a better solution out there :smile:

svenseeberg: Okay, let's continue this discussion after merging ^^

svenseeberg (approved):

PR #1625, opened 2022-08-05
Fix corrupted changelog

david-venhoff:
Fixes the Changlog, which had an entry at the wrong section


timoludwig (approved):
Thanks, didn't notice...

Issue #1628, opened 2022-08-09
Add dutch po file


JoeyStk:
### Motivation
We want to prototype a version in dutch for a possible expansion to municipalities in the Netherlands 

### Proposed Solution
Translate the existing po file to dutch (BCP-Code nl-NL)

### Alternatives
none

### Additional Context
none

PR #1629, opened 2022-08-09
Add dutch translation file

JoeyStk:
### Short description
This PR adds a dutch translation file

### Proposed changes
- Add dutch translation file

### Resolved issues

Fixes: #1628

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -89,7 +89,7 @@ jobs:
       - persist_to_workspace:
           root: .
           paths:
-            - integreat_cms/locale/de/LC_MESSAGES/django.mo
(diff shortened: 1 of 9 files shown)


timoludwig (approved):
Thanks a lot! :muscle: 

There were still a few changes missing until the translation file could be used as an UI language in the CMS, but since automated suggestions didn't work here, I allowed myself to implement the changes myself :innocent: 

Apart from that, there were two small problems with your translation file:
1. The compiled format of the translation file should be excluded from git and just compiled on the local device
2. All variable names in the form of `%(variable)s` were translated, however they need to be preserved in order for the string replacement to work.

I fixed these problems in your commit, so this PR is ready to go! :tada:

Issue #1011, opened 2021-11-02
Format datetime fields in API as `ISO8601`


timoludwig:
### Motivation

When implementing the API, we replicated the WordPress API as similar as possible, including some weird date formats which are not ideal for processing them in the apps.

### Proposed Solution

- For every datetime field, add a new field to represent the same date in [ISO_8601](https://de.wikipedia.org/wiki/ISO_8601) format. This affects e.g.:
    - The `start_date`, `start_time`/ `end_date`, `end_time` fields of events - combine them into `start`/`end`
    - The `sent_date` field of push notifications
    - The `modified_gmt` field of all other objects (maybe also rename this to `last_updated` or `last_modified`)

steffenkleinle:
Could we move this a little bit up please? We still have to live with moment :/

svenseeberg:
Should be possible, yes :)

JanErikHuehne:
Is it correct, that the sent_date and modified_gmt have already been modified beforehand?

timoludwig:
> Is it correct, that the sent_date and modified_gmt have already been modified beforehand?

The `modified_gmt` field of pages & imprint is already correct, but e.g. the `modified_gmt` field of the event and location API reponse and the `timestamp` field of the  push notification API are in the `Y-m-d H:i:s` format, but the desired format is `YYYY-MM-DDTHH:MM:SS.fff` - the representation of [datetime.isoformat()](https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat) (which is also the default string representation of a [DateTimeField](https://docs.djangoproject.com/en/4.1/ref/models/fields/#datetimefield) in Django, so just use the unformatted - but timezone aware - model field in the API response and Django will automatically format it correctly).

PR #1642, opened 2022-08-21
Use ISO datetime format

JanErikHuehne:
### Short description
Adjusted API datetime field to match an ISO standard

### Proposed changes
- Remove date and time fields from event model, add datetime fields for start and end
- Use the event view to convert date and time fields of the event form to datetime fields of event model
- Adjust API datetime fields: Rename them to "last_updated" and let them return the datetime in ISO format
- For Event-API: Make use new event fields start and end to return ISO-datetime fields instead of start_time, start_date, end_time, end_date
- Adjust the input test data json and the expected test result json files to match the new input and output formats
### Resolved issues
Fixes: #1011


integreat_cms/cms/views/events/event_list_view.py
```diff
@@ -103,16 +103,19 @@ def get(self, request, *args, **kwargs):
...
             chunk_size,
         )
         chunk = request.GET.get("page")
         event_chunk = paginator.get_page(chunk)
-
+        for event in event_chunk:
+            setattr(event, "start_date", event.start.date())
+            setattr(event, "start_time", event.start.time())
+            setattr(event, "end_date", event.end.date())
+            setattr(event, "end_time", event.end.time())
```
timoludwig, line 118:
```suggestion
```
This looks very ugly to me. Could we rather adapt the templates as well so they also use the new fields?

timoludwig (commented):
Awesome, congratulations to your first PR and thanks a lot! :tada: :rocket:

First of all: I'm sorry I labeled this issue as "good first issue" - it was definitely not :wink:
I am all the more impressed that you have fought your way through it. :muscle: 

In general, everything seems to work as expected :+1:

I see two major points that should be fixed before merging this PR:
1. We need to migrate our existing event data, otherwise all start and end dates will be reset to the point in time when we install this update on our server.
2. We should also adapt the event list templates to use the same new fields instead of setting custom attributes.

If you have any troubles implementing these changes, don't hesitate to ask!

integreat_cms/cms/forms/events/event_form.py
```diff
@@ -43,19 +63,11 @@ class Meta:
         model = Event
         #: The fields of the model which should be handled by this form
         fields = [
```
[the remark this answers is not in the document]
JanErikHuehne: I have implemented it this way, works fine. I am still not sure if we have to use HiddenInputs as widgets for the `start` and `end`. I have implemented it to be sure.

ulliholtgrave:
@digitalfabrik/app-team FYI We are currently implementing your required ISO format :)

integreat_cms/cms/views/events/event_form_view.py
```diff
@@ -212,7 +222,20 @@ def post(self, request, **kwargs):
...
+                event_form.cleaned_data["start_date"],
+                event_form.cleaned_data["start_time"],
+            )
+            event.end = datetime.combine(
+                event_form.cleaned_data["end_date"],
+                event_form.cleaned_data["end_time"],
+            )
+            event.save()
+            event_form.save_m2m()
+            event_translation_form.instance.event = event
```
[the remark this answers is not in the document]
JanErikHuehne: Done

integreat_cms/cms/views/events/event_list_view.py
JanErikHuehne: Also done.

integreat_cms/api/v3/events.py
```diff
@@ -26,15 +26,21 @@ def transform_event(event, custom_date=None):
...
+    start = datetime.combine(start_date, event.start.time(), tzinfo=event.start.tzinfo)
+    end = datetime.combine(end_date, event.end.time(), tzinfo=event.end.tzinfo)
     return {
         "id": event.id if not custom_date else None,
-        "start_date": custom_date or event.start_date,
-        "end_date": custom_date + (event.end_date - event.start_date)
-        if custom_date
-        else event.end_date,
+        "start": start,
+        "end": end,
```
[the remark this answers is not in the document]
JanErikHuehne: done

integreat_cms/api/v3/events.py
```diff
@@ -26,15 +26,28 @@ def transform_event(event, custom_date=None):
...
+        #  deprecated field in the future
+        "end_date": end.date(),
+        #  deprecated field in the future
+        "end_time": end.time(),
         "all_day": event.is_all_day,
-        "start_time": event.start_time,
-        "end_time": event.end_time,
         "recurrence_id": event.recurrence_rule.id if event.recurrence_rule else None,
         "timezone": settings.CURRENT_TIME_ZONE,
```
timoludwig, line 52:
```suggestion
        "timezone": event.timezone,
```

integreat_cms/cms/views/events/event_form_view.py
```diff
@@ -73,9 +72,15 @@ def get(self, request, *args, **kwargs):
         ).first()
         if event_instance:
             poi_instance = event_instance.location
+            initial = {
+                "start_date": event_instance.start.date(),
+                "start_time": event_instance.start.time(),
+                "end_date": event_instance.end.date(),
+                "end_time": event_instance.end.time(),
```
timoludwig, line 79:
```suggestion
                "start_date": event_instance.start_local.date(),
                "start_time": event_instance.start_local.time(),
                "end_date": event_instance.end_local.date(),
                "end_time": event_instance.end_local.time(),
```

timoludwig (commented):
Amazing, thanks! :pray: 

I have to admit that I misunderstood how Django handles timezones :sweat_smile: 
Even if dates are given in timezone aware local times, Django converts them to UTC and does not keep track of the original timezone when writing them to the database. So every time we retrieve a date from the database, we have to convert it to the local time in the region's timezone before we either display it in the CMS or deliver it via the API.
This makes everything a bit more complicated, I hope I did not mess up my suggestions.

integreat_cms/cms/models/events/event.py
```diff
@@ -85,10 +85,8 @@ class Event(AbstractContentModel):
         on_delete=models.PROTECT,
         verbose_name=_("location"),
     )
-    start_date = models.DateField(verbose_name=_("start date"))
-    start_time = models.TimeField(blank=True, verbose_name=_("start time"))
-    end_date = models.DateField(verbose_name=_("end date"))
-    end_time = models.TimeField(blank=True, verbose_name=_("end time"))
+    start = models.DateTimeField(verbose_name="start", default=datetime.now)
+    end = models.DateTimeField(verbose_name="end", default=datetime.now)
```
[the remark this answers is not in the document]
JanErikHuehne: 1. Missed those thanks! 
2. Yes. But I agree, that your suggested work around in the data migration is better :+1:

integreat_cms/cms/migrations/0033_event_datetime.py
```diff
@@ -0,0 +1,76 @@
...
+                default=datetime.datetime.now, verbose_name="end"
+            ),
+        ),
+        migrations.AddField(
+            model_name="event",
+            name="start",
+            field=models.DateTimeField(
+                default=datetime.datetime.now, verbose_name="start"
+            ),
+        ),
```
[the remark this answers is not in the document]
JanErikHuehne: Makes sense. I think this might effect the internal representation in the database and in what order it is handled. So I am not sure of the effect of it.

integreat_cms/cms/migrations/0033_event_datetime.py
```diff
@@ -0,0 +1,76 @@
...
+        migrations.RunPython(start_and_end_init),
+        migrations.RemoveField(
```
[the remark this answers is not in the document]
JanErikHuehne: I like this workaround :+1:

integreat_cms/cms/migrations/0033_event_datetime.py
```diff
@@ -0,0 +1,76 @@
...
+    :param apps: The configuration of installed applications
+    :type apps: ~django.apps.registry.Apps
+
+    :param schema_editor: The database abstraction layer that creates actual SQL code
+    :type schema_editor: ~django.db.backends.base.schema.BaseDatabaseSchemaEditor
+    """
+    events = apps.get_model("cms", "Event")
+    for event in events.objects.all():
+        event.start = datetime.datetime.combine(event.start_date, event.start_time)
+        event.end = datetime.datetime.combine(event.end_date, event.end_time)
```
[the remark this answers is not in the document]
JanErikHuehne: Timezones overall seem to be huge pain in the neck. I think your handling makes a lot of sense though. I think it might get a bit more complicated when we want to convert the timezones to the local times of the end-user when displaying them. I think this will need a lot of special treatment, especially when checking for all day events. We will have to be very careful with the logic and ordering of the conversions then, in order for it to function in the excepted way.

integreat_cms/cms/migrations/0033_event_datetime.py
```diff
@@ -0,0 +1,76 @@
...
+
```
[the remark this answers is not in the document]
JanErikHuehne: Okay, I didn´t know about this aspect yet. But it makes a lot of sense :+1:

integreat_cms/cms/migrations/0033_event_datetime.py
```diff
@@ -0,0 +1,76 @@
...
+            ),
+        ),
+        migrations.AddField(
+            model_name="event",
+            name="start",
+            field=models.DateTimeField(
+                default=datetime.datetime.now, verbose_name="start"
+            ),
+        ),
+        migrations.RunPython(start_and_end_init),
```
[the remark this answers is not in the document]
JanErikHuehne: @timoludwig  
So this has no effect in the normal migration? Since the fields exist when we call the script? But are they not deleted right after in any case? Maybe I´m wrong here. But from the logic standpoint I see it as: If the fields start_time etc. do not exist, create them and then delete them right after?

integreat_cms/cms/forms/events/event_form.py
```diff
@@ -43,20 +63,17 @@ class Meta:
...
         #: The widgets which are used in this form
         widgets = {
-            "start_date": forms.DateInput(format="%Y-%m-%d", attrs={"type": "date"}),
-            "end_date": forms.DateInput(format="%Y-%m-%d", attrs={"type": "date"}),
-            "start_time": forms.TimeInput(format="%H:%M", attrs={"type": "time"}),
-            "end_time": forms.TimeInput(format="%H:%M", attrs={"type": "time"}),
             "icon": IconWidget(),
+            # Not sure if they have to be set to HiddenInputs to work correct
+            "start": forms.widgets.HiddenInput(),
+            "end": forms.widgets.HiddenInput(),
```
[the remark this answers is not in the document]
JanErikHuehne: Alright, good to know! :v:

JanErikHuehne:
I think that we also have to update the expected test results as well to match the timezone. I think that part is still missing, or?

integreat_cms/cms/models/events/event.py
```diff
@@ -149,7 +147,7 @@ def is_all_day(self):
         :return: Whether event takes place all day
         :rtype: bool
         """
-        return self.start_time == time.min and self.end_time == time.max.replace(
+        return self.start.time() == time.min and self.end.time() == time.max.replace(
             second=0, microsecond=0
         )
```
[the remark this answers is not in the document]
JanErikHuehne: Makes sense. LGTM!

integreat_cms/cms/migrations/0033_event_datetime.py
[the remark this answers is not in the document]
timoludwig: I'm not sure whether I understand your comment correctly... Yes, the reversing code has no effect on the normal migration, but only because it is not executed, not because it has no side effects...
The first arguments of [RunPython()](https://docs.djangoproject.com/en/3.2/ref/migration-operations/#runpython) is called only when the migration is applied, and the second argument only when the migration is unapplied. When [reversing migrations](https://docs.djangoproject.com/en/3.2/topics/migrations/#reversing-migrations), the opposite operations are executed in the reverse order, so:
#### Applying this migration:
1. Create new `start` and `end` fields
2. Execute `start_and_end_init()` to populate new fields with the combined values of the old fields
3. Make `start` and `end` fields non-nullable
4. Remove old `start_date`, `start_time`, `end_date` and `end_time` fields
#### Reversing this migration:
1. Create old `start_date`, `start_time`, `end_date` and `end_time` fields
2. Make `start` and `end` fields nullable
3. Execute `start_and_end_reverse()` to populate the old fields with the split values of the new fields
4. Remove new `start` and `end` fields

integreat_cms/api/v3/events.py
```diff
@@ -2,13 +2,13 @@
...
 
-from datetime import timedelta
+from datetime import datetime, timedelta
 
 from django.conf import settings
 from django.http import JsonResponse
 from django.utils import timezone
 from django.utils.html import strip_tags
-
+from django.utils import timezone
```
timoludwig, line 11:
```suggestion

```
`timezone` is already imported in line 9

integreat_cms/cms/forms/events/event_filter_form.py
```diff
@@ -94,9 +94,14 @@ def apply(self, events, region, language_slug):
             pass
         elif events_time_range.CUSTOM in cleaned_time_range:
             # Filter events for their start and end
-            events = events.filter_upcoming(
-                self.cleaned_data["date_from"] or date.min
-            ).filter(end_date__lte=self.cleaned_data["date_to"] or date.max)
+            tzinfo = zoneinfo.ZoneInfo(region.timezone)
```
timoludwig, line 97: I guess we need to `import zoneinfo` in this file.

integreat_cms/cms/forms/events/event_form.py
```diff
@@ -1,6 +1,6 @@
 import logging
```
timoludwig, line 2:
```suggestion
import zoneinfo

```

integreat_cms/cms/migrations/0033_event_datetime.py
```diff
@@ -0,0 +1,108 @@
+# Generated by Django 3.2.14 on 2022-08-21 13:03
+import zoneinfo
+from datetime import datetime
+from django.db import migrations, models
+from django.utils import timezone
```
timoludwig, line 5:
```suggestion
```
This import is now unused

integreat_cms/cms/models/events/event.py
```diff
@@ -35,9 +35,9 @@ def filter_upcoming(self, from_date=None):
         :return: The Queryset of events after the given date
         :rtype: ~integreat_cms.cms.models.events.event.EventQuerySet
         """
-        from_date = from_date or date.today()
+        from_date = from_date or timezone.now().date()
```
timoludwig, line 38: `from django.utils import timezone` is missing at the top of the file

integreat_cms/cms/models/events/event.py
```diff
@@ -149,10 +147,43 @@ def is_all_day(self):
         :return: Whether event takes place all day
         :rtype: bool
         """
-        return self.start_time == time.min and self.end_time == time.max.replace(
+        return self.start_local.time() == time.min and self.end_local.time() == time.max.replace(
             second=0, microsecond=0
```
timoludwig, line 151:
```suggestion
        return (
            self.start_local.time() == time.min
            and self.end_local.time() == time.max.replace(second=0, microsecond=0)
```
Our code formatting tool [black](https://github.com/digitalfabrik/integreat-cms/blob/develop/dev-tools/black.sh) wants to reformat these lines...

timoludwig (commented):
Awesome, thanks a lot! :muscle: 

Now there is not too much left to do I think. After you fixed the last issues and adapted the tests, could you clean up the git history and rebase the branch? After that, I will test everything extensively again :grin:

> Timezones overall seem to be huge pain in the neck

Yes, they give me quite a headache as well :sweat_smile: 

> I think that we also have to update the expected test results as well to match the timezone. I think that part is still missing, or?

Yes, I didn't make suggestions for this part because it's probably easier to just retrieve the endpoints and copy + paste the result into the files instead of replacing the dates manually...

integreat_cms/api/v3/events.py
JanErikHuehne: Wait my IDE still marked it as missing. Maybe something was still loading, my bad :D

integreat_cms/cms/forms/events/event_form.py
JanErikHuehne: Datetime did also still miss here. I added both

integreat_cms/cms/models/events/event.py
JanErikHuehne: Done

integreat_cms/cms/models/events/event.py
JanErikHuehne: Already resolved by [c79e687](https://github.com/digitalfabrik/integreat-cms/pull/1642/commits/c79e6870b2025d6d81e2ff24dcb2a496891c0b04)

integreat_cms/cms/forms/events/event_filter_form.py
JanErikHuehne: Resolved

integreat_cms/cms/migrations/0033_event_datetime.py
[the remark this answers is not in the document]
JanErikHuehne: Ah got it! I miss understand that the second argument of RunPython is only run when the migration is unapplied. That is actually a cool concept, I have to dive deeper into the django migrations at some point I guess :D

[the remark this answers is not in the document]
JanErikHuehne: Added it!

JanErikHuehne:
> Now there is not too much left to do I think. After you fixed the last issues and adapted the tests, could you clean up the git history and rebase the branch? After that, I will test everything extensively again 

Yes sure. But I´m not sure when I will get to it since I am going to Munich today. Maybe I will find time and space in the train. In any other case, I should have time for it sometime next week

JanErikHuehne:
I completed the rebase and everything seems to work fine. I squashed the commits into 1 and added one slightly longer commit description. @timoludwig if you find something while re-checking everything, let me know. I added the change in the unreleased block inside the changelog, I hope this is fine.

timoludwig (approved):
Great, thank you so much! :pray: 
Works flawlessly :heart_eyes:

Issue #1450, opened 2022-05-17
Add API from SUMM.AI to create easy-understable German


dkehne:
### Motivation
Summ.ai is an MT-service for easy understable german. They developed an API-service for us so that we can offer municipalities https://summai.notion.site/SUMM-API-Documentation-external-fc41fa5ef1434ad2b311be4876b4685a

### Proposed Solution
Implement the Summ.ai API in a similar way to the DeepL API from Johannes' Project. The Summ-Team asked as if we can deliver the "instance slug" as user (muenchen, augsburg, lkaugsburg) so that they can track which city uses the API most often. 

credentials for the API are in Passbolt

dkehne:
Is this sth. we good handover to kiitos or @JoeyStk ? 
Still got one deadline for 30th September 2022

PR #1652, opened 2022-08-26
Add Summ AI API

JoeyStk:
### Short description
This PR adds Summ.AI and therefore the option to automatically translate German events, POIs and pages to simple German.

### Proposed changes

- This PR adds the option to automatically translate to simple German

### Resolved issues

Fixes: #1450


integreat_cms/cms/templates/events/event_list.html
```diff
@@ -130,6 +130,9 @@ <h1 class="heading">{% trans 'Events' %}</h1>
                         </option>
                     {% endif %}
                 {% endif %}
+                <option data-bulk-action="{% url 'simple_german_translation_pages' %}" title=" {% trans 'Automatically translate this page to easy language' %}">
+                    {% trans 'Automatically translate this page to easy language' %}
+                </option>
```
timoludwig, line 135: I guess it makes sense to show this option only when the `SUMM_AI_ENABLED` setting is `True`, right?
The same goes for the other content types.

integreat_cms/cms/templates/events/event_list.html
```diff
@@ -130,6 +130,9 @@ <h1 class="heading">{% trans 'Events' %}</h1>
                         </option>
                     {% endif %}
                 {% endif %}
+                <option data-bulk-action="{% url 'simple_german_translation_pages' %}" title=" {% trans 'Automatically translate this page to easy language' %}">
+                    {% trans 'Automatically translate this page to easy language' %}
```
timoludwig, line 134:
```suggestion
                    {% trans 'Automatically translate this page to simple German' %}
```
The API only handles German and no other languages, right?

integreat_cms/summ_ai/utils.py
```diff
@@ -0,0 +1,106 @@
...
+
+# from django.conf import settings
+from django.apps import apps
+from django.contrib import messages
+from django.utils.translation import ugettext as _
+
+logger = logging.getLogger(__name__)
+
+# pylint: disable=too-few-public-methods
+class SummAiAPI:
```
timoludwig, line 13:
```suggestion
class SummAiApiClient:
```
To be extra nit-picky, this class is the client to interact with the API, not the API itself :grin:
Also, since this file only contains the class, I would rename it from `utils.py` to `summ_ai_api_client.py`.

integreat_cms/summ_ai/utils.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+    Summ AI API to get german posts in easy german language.
+    """
+
+    def __init__(self):
+        """
+        Initialize the Summ AI client
+        """
+        url = "https://backend.summ-ai.com/translate/"
+        self.translator = requests.get(url)
```
timoludwig, line 23: The translator is typically a client object itself. If we don't have a fully featured api client for the API yet, we have to implement the request methods ourselves, but it does not make sense to make only one dummy request in the init method if the subsequent requests cannot build on it.

timoludwig (commented):
Thanks a lot! :+1:
Since this is only a draft, I won't concentrate on the missing functionality, I have just a few preliminary thoughts...

timoludwig (commented):
I have now implemented the SUMM.AI API client. I faced a few challenges, because
1. The API returns weird strings when given HTML
2. It takes very long for whole pages with a few thousand characters (multiple minutes, I faced request timeouts very often)

So the solution I came up with was HTML-segmentation by `<p>` and `<li>` tags in combination with asynchronous requests. So the way it works at the moment:

1. Segment the HTML content of all given pages into a long list of translatable strings
2. Translate all these strings in parallel
3. Wait for the requests to finish 
4. Assemble the modified HTML tree back to content strings
5. And write the changes to the database

This also has disadvantages, first of all because the context information might be lost for two subsequent paragraphs and because too many requests very quickly run into responses like:
> Too many requests. Please wait and resend your request.

So whether or not these two points are show-stoppers remains to be tested.

Have fun :joy_cat: 

## How to test?
1. Get the API key from [passbolt](https://passbolt.tuerantuer.org/app/passwords/view/211b7ee1-4ba4-4df5-b30b-d4538dac5d2d)
6. Add `SUMM_AI_API_KEY = <the api-key>` to `/etc/integreat-cms.ini`
7. Run server
8. Enable SUMM.AI in the region form
9. Execute bulk action from page tree

## How to get real translations?

6. Start server without `DEBUG` mode or add `SUMM_AI_TEST_MODE = False` to `/etc/integreat-cms.ini`

## What to watch out for?

The rate limit of the SUMM.AI API is quite low, so be a bit cautious, don't translate too many pages at once (or too large pages). Even in test mode, the API server is pretty easily DDoSable through our asynchronous requests.
Also, please don't overuse our credits, at the moment we have ~35k characters left, but yesterday I used our previous limit of 18k characters within a few requests, so only translate small pages and don't repeat to often :sweat_smile:

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -5337,27 +5372,29 @@ msgstr ""
 "Nicht (⚠️) veröffentlichte und veröffentlichte Seiten für die "
 "%(language_translated_name)s-Übersetzung exportieren"
 
-#: cms/templates/pages/page_tree.html:186
+#: cms/templates/pages/page_tree.html:191
 msgid "Export selected pages for multilingual translation"
-msgstr "Ausgewählte Seiten für die mehrsprachige Übersetzung exportieren"
+msgstr ""
+"Ausgewählte Seiten für die %(language_translated_name)s-Übersetzung "
```
ulliholtgrave, line 5378: In this part of the tree template we don't have access to the `language_translated_name` variable.

timoludwig: Thanks! Not exactly sure how this translation got changed...

integreat_cms/cms/views/mixins.py
```diff
@@ -105,3 +106,33 @@ def get_context_data(self, **kwargs):
...
+        context = super().get_context_data(**kwargs)
+        context["SUMM_AI_ENABLED"] = (
+            settings.SUMM_AI_ENABLED
+            and self.request.region.summ_ai_enabled
+            and kwargs.get("language_slug")
+            in [
+                settings.SUMM_AI_GERMAN_LANGUAGE_SLUG,
+                settings.SUMM_AI_EASY_GERMAN_LANGUAGE_SLUG,
+            ]
+        )
```
timoludwig, line 137: @JoeyStk in theory, this code should make sure that the SUMM.AI option is only shown in the template when all three conditions are `True`:

1. SUMM.AI is globally enabled (aka an API key is set)
2. SUMM.AI is enabled in the current region (via the region form)
3. The current language slug is either `settings.SUMM_AI_GERMAN_LANGUAGE_SLUG` or `settings.SUMM_AI_EASY_GERMAN_LANGUAGE_SLUG`...

Do you see a potential problem with my code which causes the condition to not work on your machine?

tests/cms/views/view_config.py
```diff
@@ -814,7 +814,7 @@
                         "form-0-language": 1,
                         "form-1-title": "",
                         "form-1-text": "",
-                        "form-1-language": 2,
+                        "form-1-language": 6,
```
timoludwig, line 817: @JoeyStk Regarding 4:
I wondered why the tests of the push notifications failed after I added a new language tree node - in theory, I didn't change anything related to push notifications.
Then, I noticed that the push notification formset expects the forms to be in the same order as the language tree... and since I added the new node Easy German (with id 6) in the language tree right after German (where previously the second language was English), the formset was confused because in the second form now there was language with id 2 but the second language in the tree is now de-si with id 6... so I just changed the test config to adapt to this new language tree structure...
Not sure whether I explained that well, to me it's still kind of confusing nevertheless :see_no_evil:

timoludwig (commented):
@JoeyStk Thanks a lot for your detailed review! :pray:

Regarding 1.:

I'm not sure what our current credit usage is, because our SUMM.AI dashboard still shows "16799 / 18000", even though the 18k are definitely consumed and we got additional 36k characters... However, for me it still works, so I assume the problem lies elsewhere... Maybe you can try again and have a look at the server log and check whether there is a more detailed error description? (sometimes I get the error when I try to translate a page with too many paragraphs, because the API has a very strict rate limiting and doesn't like too many requests per minute...

Regarding 2., 3. & 5.:

Interesting, I cannot reproduce this. In theory, the option should not be shown in all three cases...
I'm not completely sure what could go wrong here, maybe the first step would be to check whether we have the identical git state, maybe during the reset from your old state to the new one git magically merged some files or so?
What is the output of `git fetch --all && git status`? And does the output change when you execute `git reset --hard origin/feature/summ_ai`?

tests/cms/views/view_config.py
JoeyStk: Aha! That's very good to know. I never would have guessed this :thinking:

integreat_cms/summ_ai_api/utils.py
```diff
@@ -0,0 +1,323 @@
...
+        Parse the HTML string into an lxml tree object and split into segments
+
+        :param html: The HTML string content of this field
+        :type html: str
+        """
+        if html:
+            self.html = fromstring(html)
+            # Translate all <p> and <li> tags (and filter out empty segments)
+            self.segments = [
+                HTMLSegment(segment=segment) for segment in self.html.iter("p", "li")
```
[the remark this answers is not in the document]
JoeyStk: I was able to provoke an error, when I used quote and h2 elements (s. screenshot). I asked the service-team if these are commonly used elements or not. I haven't yet gotten an response. As soon as I have one, I'll comment here again :) 
image

JoeyStk (commented):
Now that I was able to make everything work, everything seems to work for me!
Regarding to your question about the html splitting: I tried to write multiple sentences containing the same context and at least as far as my tests went, I wasn't able to break it. The translation response still was senseable.
However I still think the translations themselves have a few issues from time to time and act a bit silly here and there. For example, I wrote a page titled "Leichte Sprache" - translated it, and received the very catchy headline "Leichte Sprache ist eine bestimmte Art zu sprechen. Leichte Sprache spricht man so das alle verstehen können. Zum Beispiel: Menschen mit Lern-Schwierigkeiten Menschen mit kleinen Kindern Menschen mit Demenz".
I will approve this PR as soon as I received an answer from the service-team regarding your question :)

integreat_cms/cms/views/mixins.py
JoeyStk: Please consider this to be resolved. I was able to fix the issue on my side and could test this PR properly now. Sorry for the flawed review so far. Please only consider the new review below :blush:

integreat_cms/summ_ai_api/utils.py
```diff
@@ -0,0 +1,323 @@
...
+    #: The form class of the current content type
+    form_class = None
+    #: The current object instance to be translated
+    object_instance = None
+    #: The German source translation of the object instance
+    german_translation = None
+    #: Wether or not the translation was successfull
+    valid = True
+    #: The title field of this helper
+    title_field = None
```
ulliholtgrave, line 185: Can you remove the title field from the automatic translation and always keep the german one? So far, the API seems to work rather bad with short texts (wrong spelling, super long translations and so on). This is also communicated with the service team

timoludwig: Alright! Should I just disable it temporarily, so we might be able to translate it in the future, or should I simplify the process and remove all title functionality altogether?

ulliholtgrave: I would remove the functionality in total. Right now the code for it only creates some extra complexity and we could still take your code from the commit history of this repo if we would need in the end :)

integreat_cms/summ_ai_api/utils.py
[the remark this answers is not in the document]
JoeyStk: I got a reply from the service-team, that h2 to h6 and quote-tags are rarely used. I think it's okay not to implement it now. If it comes up in the future it can still be implemented then

JoeyStk:
Since I'm still the author of this PR, I can't approve it. So here is the oral approval, if someone wants to press the button for me

integreat_cms/summ_ai_api/utils.py
timoludwig: Due to my desire to force-push, this would not be preserved in the git history :sweat_smile: 
But I found a way to make this a bit more generic without hardcoding the field names, so we can adapt them in the future.

JoeyStk:
I tested everything again. And for me everything worked! I still approve this PR :D

timoludwig (approved):
> I tested everything again. And for me everything worked! I still approve this PR :D

Thanks a lot, I will do the approve for you

ulliholtgrave (approved):
I am also fine with the PR and like how you approached the issue with the title :+1:

timoludwig:
Ok, then I only have to figure out what the hell is happening with the tests and why commit 75276114edee785eae37e879f3db1cbcbcb2d841 is required...

timoludwig:
Unfortunately, I could not find the source of the problem.
For now, I think this might be a bug in Django's test client `AsyncClient`. I will open a ticket in the Django issue tracker and see whether I get more insight there.
In the meantime, I think we can merge this PR, since the workaround should not have any impact on our production environment.

timoludwig:
> Unfortunately, I could not find the source of the problem. For now, I think this might be a bug in Django's test client `AsyncClient`. I will open a ticket in the Django issue tracker and see whether I get more insight there.

Interesting, this actually seems to be a bug of the Django testing framework:
https://code.djangoproject.com/ticket/34063

PR #1656, opened 2022-08-31
Only validate event duration if dates are valid

timoludwig:
### Short description

If one of the dates are missing or invalid, the previous solution caused a `KeyError`

### Proposed changes

- Only validate the event duration if the dates itself are valid


ulliholtgrave (approved):
Good catch :+1:

Issue #1533, opened 2022-06-08
Meta: PDF


timoludwig:
This is a meta issue for all PDF issues:

- https://github.com/digitalfabrik/integreat-cms/issues/35
- https://github.com/digitalfabrik/integreat-cms/issues/213
- https://github.com/digitalfabrik/integreat-cms/issues/547
- https://github.com/digitalfabrik/integreat-cms/issues/550
- https://github.com/digitalfabrik/integreat-cms/issues/554
- https://github.com/digitalfabrik/integreat-cms/issues/711
- https://github.com/digitalfabrik/integreat-cms/issues/751
- https://github.com/digitalfabrik/integreat-cms/issues/753
- https://github.com/digitalfabrik/integreat-cms/issues/876
- https://github.com/digitalfabrik/integreat-cms/issues/1108
- https://github.com/digitalfabrik/integreat-cms/issues/1206
- https://github.com/digitalfabrik/integreat-cms/issues/1209
- https://github.com/digitalfabrik/integreat-cms/issues/1327
- https://github.com/digitalfabrik/integreat-cms/issues/1349
- https://github.com/digitalfabrik/integreat-cms/issues/1350
- https://github.com/digitalfabrik/integreat-cms/issues/1406
- https://github.com/digitalfabrik/integreat-cms/issues/1498
- https://github.com/digitalfabrik/integreat-cms/issues/1499
- https://github.com/digitalfabrik/integreat-cms/issues/1511
- https://github.com/digitalfabrik/integreat-cms/issues/1532

laravoelk:
Hello Together, I am not sure, if this is the right place to write about known problems with PDF. If not, just let me know and I will open a new issue. ;-) As you know, languages written from right to left (Arabic, Farsi) can not be transferred into PDFs in Integreat. Is that something you are currently working on and will be changed in the future? Thanks a lot for your help.

osmers:
All issues are done

Issue #1499, opened 2022-05-25
False positive error message for urlencoded filenames in PDF export


timoludwig:
### Describe the Bug

When filenames which contain umlaute are urlencoded in the src attribute, they cause error messages in the PDF export 

### Steps to Reproduce

Create PDF which contains images with umlaute in their filename

### Expected Behavior

The images are contained in the PDF, so no error should appear

### Actual Behavior

An error is displayed in the logfile:

  ```
May 25 12:26:44 ERROR   integreat_cms.cms.utils.pdf_utils - The file '/var/www/integreat-cms/media/regions/206/2022/01/h%C3%A4usliche-gewalt.jpg' was not found in the media directories.
  ```

Even though the file exists:

  ```
$ ls /var/www/integreat-cms/media/regions/206/2022/01/häusliche-gewalt.jpg
/var/www/integreat-cms/media/regions/206/2022/01/häusliche-gewalt.jpg
  ```

jonbulz:
TL,DR: I think even though the error appears in the logs, images with Umlauten are actually rendered in PDFs, but the rendering is broken in general and images are weirdly stacked on top of each other.


I'm not sure if I'm catching all there is here, but this is difficult to reproduce for me. I'll try my best to explain what I experienced so far. I'm having two files, [test.jpg](https://user-images.githubusercontent.com/101657168/186887971-43e98f7c-385a-43c1-87a6-f3ea0bbc7227.jpg) and [täst.jpg](https://user-images.githubusercontent.com/101657168/186888141-326eb7cb-d1a4-4fac-bc74-50516e5b1021.jpg).

First off, I'm always getting the error in the logs when using image files with Umlaute as src. However, the PDF still seems to render correctly. Eg. 
```<p><a href="http://localhost:8000/media/regions/1/2022/08/t%C3%A4st.jpg"><img src="http://localhost:8000/media/regions/1/2022/08/t%C3%A4st.jpg" alt=""></a></p>```
gives the expected [täst.pdf](https://github.com/digitalfabrik/integreat-cms/files/9432450/tast.pdf).


BUT as soon as I include multiple images, it gets weirder. For example
[...]

Issue #1537, opened 2022-06-09
Inline Icons in PDF export broken


svenseeberg:
### Describe the Bug
When using icons within a line of text, the PDF export creates a white bar across the icon/image with the height of the text line. If the icon itself has about the height of the text, then the icon is mostly invisible. See https://integreat.app/testumgebung/de/alltag/information as an example.

### Steps to Reproduce

1. Create a page with an icon at the beginning and another in the middle of a line of text.
2. Save the page and export a PDF
3. See that the icon at the beginning of the line is fully visible, while the icon in the middle is partially hidden:
  ![image](https://user-images.githubusercontent.com/15197000/172789546-d0350844-5b54-435b-bc8f-8634c1891a6d.png)
### Expected Behavior
Icons/images should be visible regardless of their position in text.

### Actual Behavior
Only images at the beginning or outside of text work as expected.

### Additional Information
I guess that this is a bug in https://github.com/xhtml2pdf/xhtml2pdf/

charludo:
I've found the issue: in [this xhtml2pdf file](https://github.com/xhtml2pdf/xhtml2pdf/blob/master/xhtml2pdf/default.py) on line 403, setting the background color (any background color, doesn't matter the value - color, transparent, none,...) makes inline-icons disappear.

There's certainly some deeper underlying issue, the question is, is it worth looking into? We could simply use a version of this default css without that line.

PR #1661, opened 2022-09-04
Change PDF library from xhtml2pdf to weasyprint

jonbulz:
### Short description
Replace xhtml2pdf by weasyprint, as there seem to be several formatting issues in pdfs generated with xhtml2pdf.

### Resolved Issues
I didn't get to do a lot of testing, but I believe #1499 and #1533 are better now.


jonbulz:
I didn't check this as thoroughly as I would have liked, but I just wanted to open this PR now because I won't be free the next couple of days. This way, maybe somebody else can already have a look.

Oh, and obviously this requires the weasyprint module, which I just installed locally via pip. Would this need to be somewhere in this PR?

timobrembeck (commented):
Thanks a lot! :+1:

In general, I approve the idea of moving towards a more stable PDF library, especially considering all the problems we had with xhtml2pdf in the past.

> I didn't get to do a lot of testing, but I believe #1499 and #1533 are better now.

Yes, this change seems to fix all of the weird PDF-bugs we have at the moment, also e.g. #1537! :+1:
I also did not have the time to do in-depth testing, but just from the first look, I already see a few points we need to address before making the change final.

- [ ] Fonts are not working yet (we should try to make the PDF look similar to the web app which includes using the same fonts for headings and the content)
- [ ] The table of contents is not working yet (all page titles aka `<h1>`-elements should be listed in the TOC)
- [ ] Images are not resized when they overlap the page
- [ ] The footer with page numbers and Integreat logo (and optionally the region logo) is missing


Probably, there are some more points I forgot, but we have to make sure that moving to another library does not re-open issues we already have fixed for xhtml2pdf.

> Oh, and obviously this requires the weasyprint module, which I just installed locally via pip. Would this need to be somewhere in this PR?

Adding functional dependencies is documented [here](https://digitalfabrik.github.io/integreat-cms/virtualenv.html#add-dependencies):

> When adding new functional dependencies, please add them to https://github.com/digitalfabrik/integreat-cms/blob/develop/setup.cfg and execute:
>
>     pipenv update integreat-cms

jonbulz:
You're right, I didn't even notice. That doesn't look good at all. Unfortunately, wkhtmltopdf also doesn't seem to work right out of the box, so I guess it might be necessary to make some upstream contributions to one of the libraries to fix our issues.
I'll try to look into what would be the least effort to get where we want to. ^^

timobrembeck (changes requested):
I'm just changing the status to make it easier to keep the overview which PRs still need reviews and which don't :wink:

ulliholtgrave:
@jonbulz Thank you very much for your effort so far, but as we discussed in our last teamcall, we decided to fix our current setup with XHTML2PDF and maybe provide some upstream patches :+1: 

Therefore I am closing this PR for housekeeping :)

Issue #1660, opened 2022-09-01
Moving page to first level from "detailed page view" does not work. 


FritjofK:
### Describe the Bug
In the CMS of VernetztesObdach I tried to move a page from a lower "level" (a children page) to the first level while being on the detailed page view. 

Drag and Drop from lower level to first level on the Page Tree View does(!) work. 

### Steps to Reproduce

1. Go to any childre page to "Reihenfolge" 
2. Click on "Übergeordnete Seite"
3. Select "---------------------" and click on "Aktualisieren"
4. No error message, but confirmation is shown, but the page stays at its previous spot. 

### Expected Behavior
Move the page to first level and once selecting it also show all pages on the first level to order it in the desired place. 

### Actual Behavior
No error message, but confirmation is shown indicating that it worked, but the page stays at its previous spot. 

### Additional Information

CMS of VernetztesObdach which sometimes is behind in releases. I did not test it in Malte nor Integreat. 

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

timoludwig:
This was closed in #1662

PR #1463, opened 2022-05-21
Hide archived pages in various selections

david-venhoff:
### Short description

This pr hides archived pages in the page form in the parent page selection, the page order selection, and the mirrored page selection.
Additionally, the page order selection should update correctly now when selecting '----------' as parent page after selecting something else as parent page.

### Proposed changes

- Add filters to exclude archived pages

### Resolved issues

Fixes: #1283

PR #1662, opened 2022-09-05
Fix page order table for root pages

timoludwig:
### Short description

The page order table in the page form did not work correctly for root pages.

### Proposed changes

- Fix page order table for root pages

### Resolved issues

Fixes: #1660

### DIFF
--- a/integreat_cms/cms/views/pages/page_actions.py
+++ b/integreat_cms/cms/views/pages/page_actions.py
@@ -771,20 +771,17 @@ def get_page_order_table_ajax(request, region_slug, parent_id=None, page_id=None
 
     region = request.region
 
-    if page_id:
-        page = get_object_or_404(region.pages, id=page_id)
-    else:
-        page = None
+    page = get_object_or_404(region.pages, id=page_id) if page_id else None
 
-    if parent_id or page:
-        parent = (
-            get_object_or_404(region.pages, id=parent_id) if parent_id else page.parent
-        )
-        siblings = parent.cached_children
+    if parent_id:
+        parent = get_object_or_404(region.pages, id=parent_id)
+        siblings = [
+            sibling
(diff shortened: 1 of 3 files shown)


integreat_cms/cms/views/pages/page_actions.py
```diff
@@ -771,20 +771,17 @@ def get_page_order_table_ajax(request, region_slug, parent_id=None, page_id=None
...
+            if not sibling.explicitly_archived
+        ]
     else:
-        page = None
-
-    if parent_id or page:
-        parent = (
-            get_object_or_404(region.pages, id=parent_id) if parent_id else page.parent
-        )
-        siblings = parent.cached_children
```
timoludwig, line 783: @david-venhoff do you remember why you added the [`or page` case](https://github.com/digitalfabrik/integreat-cms/commit/7c505dab0e123302eab5fee49c0ad6a6a6478e94#diff-4956f040b51352b8fde20d1ff3cd57f3879253563dd16d10a75abb5550915cd4) in  in PR #1463?
I believe this is the source of the bug - since it always loads the "old" siblings of the page when no `parent_id` is given (but in this case we want the root pages of the region since we want to move the page to a new location)...
Or am I missing something?

david-venhoff: I think at some point there was a bug that the parent page input would default to "-----", while the page order table showed all the pages siblings by default (as it still does). Then, when changing the parent page input to something and then going back to "-----", the page order table showed something different. I fixed that inconsistency in the pr not realizing that "-----" was supposed to mean "no parent page" instead of "default". 
So from my point of view there are no objections to this change, this seems much better :+1:

timoludwig: Alright, thanks for the feedback!

Issue #1647, opened 2022-08-23
Adjust migrations to provide deletion capabilities to service team


ulliholtgrave:
### Motivation

The service team requested to have the permission to delete the following two models:

- Regions
- PushNotifications

These permissions need to be added to the migrations to withstand updating our systems.

Issue #1583, opened 2022-06-30
Add chat permission


ulliholtgrave:
### Motivation

Right now every role can see the chat if it is activated for a region. In the old system the chat was only visible to the "Verwalter" role.

### Proposed Solution

- Add the "change" permission to the [chat message model](https://github.com/digitalfabrik/integreat-cms/blob/839ceea71fae49654ab00155100473f41c3a0c56/integreat_cms/cms/models/chat/chat_message.py#L82) and give it to the "Verwalter" and all IG roles

ulliholtgrave:
> The chat is already enabled by default:
> 
> https://github.com/digitalfabrik/integreat-cms/blob/839ceea71fae49654ab00155100473f41c3a0c56/integreat_cms/cms/models/regions/region.py#L245-L246

I updated the issue.

ulliholtgrave:
As we are already touching the chat widget we should also come up with an idea about how we handle the widgets if statistics and chat are disabled. Our current implementation leaves a big blank whole in the middle of the screen.

PR #1671, opened 2022-09-07
Add chat permission and give deletion rights to service team

JoeyStk:
### Short description
Change chat permission

### Proposed changes
- Restrict change chat permission to management and IG roles only
- Give delete chat permission to all roles
- Give deletion permission for push notifications and regions to service team

### Resolved issues

Fixes: #1583
Fixes: #1647


integreat_cms/cms/constants/roles.py
```diff
@@ -75,6 +75,8 @@
     "change_imprintpage",
     "change_pushnotification",
     "change_user",
+    "change_chatmessage",
+    "delete_chatmessage",
```
timoludwig, line 79:
```suggestion
```
I don't see a reason why we should allow the management role to delete all chat messages (including those by other users)... they are already allowed to delete their own messages...

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ UNRELEASED
 
 * [ [#1664](https://github.com/digitalfabrik/integreat-cms/issues/1664) ] Fix media library and content edit lock in sbs view
 * [ [#1660](https://github.com/digitalfabrik/integreat-cms/issues/1660) ] Fix moving pages to the root level from the page form
+* [ [#1583](https://github.com/digitalfabrik/integreat-cms/issues/1583) ] Add chat permission
```
```suggestion
* [ [#1583](https://github.com/digitalfabrik/integreat-cms/issues/1583) ] Only enable the chat for staff and management roles
```

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ UNRELEASED
...
+* [ [#1647](https://github.com/digitalfabrik/integreat-cms/issues/1647) ] Give service team delete permissions for regions and pns
```
```suggestion
* [ [#1647](https://github.com/digitalfabrik/integreat-cms/issues/1647) ] Give service team delete permissions for regions and push notifications
```

integreat_cms/cms/templates/dashboard/dashboard.html
```diff
@@ -14,7 +14,7 @@ <h1 class="heading">{% trans 'My Dashboard' %}</h1>
                     {% include "statistics/_statistics_widget.html" %}
                 {% endif %}
                 <!-- Chat Widget -->
-                {% if chat_form %}
+                {% if chat_form and perms.cms.delete_event %}
```
timoludwig, line 17:
```suggestion
                {% if chat_form and perms.cms.change_chatmessage %}
```

integreat_cms/cms/migrations/0002_roles.py
```diff
@@ -127,9 +127,11 @@
             "delete_poi",
             "view_poi",
             "change_pushnotification",
+            "delete_pushnotification",
```
timoludwig, line 130:
```suggestion
```
I would not suggest to edit existing migrations.
This only works on newly setup systems and all existing databases will go into an inconsistent state (because they think they already have applied the migration, but the migration is now different)...
For more context, see [this SO answer](https://stackoverflow.com/a/37097258).

integreat_cms/cms/migrations/0002_roles.py
```diff
@@ -127,9 +127,11 @@
...
             "send_push_notification",
             "view_pushnotification",
             "change_region",
+            "delete_region",
```
timoludwig, line 134:
```suggestion
```

integreat_cms/cms/migrations/0034_alter_chatmessage_options.py
```diff
@@ -0,0 +1,25 @@
...
+    operations = [
+        migrations.AlterModelOptions(
+            name="chatmessage",
+            options={
+                "default_permissions": ("delete", "change"),
+                "ordering": ["-sent_datetime"],
+                "verbose_name": "chat message",
+                "verbose_name_plural": "chat messages",
+            },
+        ),
```
timoludwig, line 24:
```suggestion
        ),
        migrations.RunPython(add_roles, remove_roles),
```
You need to implement to functions to assign/unassign the new permission to users.

integreat_cms/cms/constants/roles.py
```diff
@@ -116,6 +117,7 @@
     "change_poi",
     "change_pushnotification",
     "change_region",
+    "change_chatmessage",
```
timoludwig, line 120: I would argue that the marketing team should also be allowed to use the chat

timoludwig (commented):
Thanks a lot! :+1:

Unfortunately, the assignment of the new roles does not work as expected, since we need to adjust them in the new migration instead of editing past migrations.

integreat_cms/cms/constants/roles.py
timoludwig: @JoeyStk If you resolve conversations without making changes, please write a short comment why you think it's irrelevant :slightly_smiling_face:

timoludwig (commented):
Thanks a lot! :+1:

Two minor points:
1. I think it would make sense that all staff roles can use the chat, so starting with the marketing team
2. Having two different commits for this PR only makes sense if the changes for "adding chat permission" and "grant deletion permissions to service team" are really separated from another. But when the diff is mixed anyway and the commit description does not match what is happening, I suggest to combine the commits into one instead.

integreat_cms/cms/migrations/0035_alter_chatmessage_options.py
```diff
@@ -0,0 +1,52 @@
...
+
+    :param apps: The configuration of installed applications
+    :type apps: ~django.apps.registry.Apps
+
+    :param schema_editor: The database abstraction layer that creates actual SQL code
+    :type schema_editor: ~django.db.backends.base.schema.BaseDatabaseSchemaEditor
+    """
+    Group = apps.get_model("auth", "Group")
+    Permission = apps.get_model("auth", "Permission")
+
```
timoludwig, line 20:
```suggestion

    # Emit post-migrate signal to make sure the Permission objects are created before they can be assigned
    emit_post_migrate_signal(2, False, "default")

```
Wow, this really confused me... I couldn't figure out why the migration wouldn't work for the `change_chatmessage` permission, apparently the migration changing the model options does not create the new permission, but the auth app waits for the migrations to finish and only after that it creates them... Unless we emit this signal manually beforehand.

timoludwig (commented):
It took me a bit of time to realize why the heck the migration wasn't working, but now I think I got it :smile:

timoludwig (approved):
Thanks, now it works for me! :+1:

Issue #1591, opened 2022-07-12
Add bulk actions for language tree nodes


dkehne:
### Motivation
After we have duplicated the e.g. Deutschlandvorlage we disable all foreign languages to not confuse users because the decision on what languages they want to use, is taken later on in the onboarding process.

In WordPress CMS it was possible to unmark language-checkboxes so that the languages were disabled (or deleted if no content was existing). In Django CMS this process needs to be done seperatly for each language taking up 2-3 minutes instead of 5 seconds.

If there is any solution to enable the deletion of more than 1 language at one, it would be a good enhancement.

no prio.

PR #1674, opened 2022-09-08
Improve language tree views

timoludwig:
### Short description

A few improvements for our language tree views

### Proposed changes

- Refactor language tree views to make use of the [new](27ecf162c33ef6e0c8c7853c741f8435f4a69585) generic views
- Improve language tree node creation
    - Disable changing the language after a node has been created
    - Preselect the region's default language as source language
    - Add button to add more nodes
- Generalize bulk action views
    - Make it easier to create new bulk actions which set boolean fields
- Add bulk actions for language tree nodes

### Resolved issues

Fixes: #1591


ulliholtgrave (approved):
Thanks, everything works as expected and I like the idea of more generalization :+1:

Issue #951, opened 2021-09-27
Categories for POIs


svenseeberg:
If many locations/POIs are displayed on the map within the app, we want to add a category field to our POIs that can later be used by the apps for filtering etc. The categories should be dynamic and not be hardcoded.

Example categories are: 
Beratungsstelle - Advice center
Behörde/Amt - Authority/Office
Treffpunkt - Hangout
Freizeit - Recreation
Einkaufen - Shopping
Dienstleistung - Service
Mobilität - Mobility
Gesundheit - Health
Kultur - Culture
Gastronomie - Gastronomy

timoludwig:
Probably it makes sense to create a new model `POICategory` or similar where we can create new entries in the network area which can then be selected in a choice field in the `POI` model, right?

dkehne:
I would suggest to use the same categories as we use in social pass and volunteering card + additionally Behörde/Amt

Behörde/Amt - Authority/Office
Einrichtung - Institution
Treffpunkt - Hangout
Freizeit - Recreation
Einkaufen - Shopping
Dienstleistung - Service
Mobilität - Mobility
Gesundheit - Health
Kultur - Culture
Gastronomie - Gastronomy
Multimedia - Multimedia
Sonstiges - Others

timoludwig:
@dkehne The categories are only stored in the database and can be managed in the CMS without having to change the source code.

PR #1677, opened 2022-09-11
Add Model POI Category

MizukiTemma:
### Short description

This PR introduces a new model POICategory and add an input field "Category" to POIs. 

### Proposed changes

- A new model POICategory is created.
- New POI Categories can be added/re-named in the network management.
- One category can be chosen for each POI.

### Resolved issues

Fixes: #951


integreat_cms/cms/models/poi_categories/poi_category.py
```diff
@@ -0,0 +1,53 @@
...
+from django.utils.translation import ugettext_lazy as _
+
+from ..abstract_base_model import AbstractBaseModel
+
+
+class POICategory(AbstractBaseModel):
+    """
+    Data model representing a POI category.
+    """
+
```
timoludwig, line 11:
```suggestion

    slug = models.SlugField(
        max_length=1024,
        unique=True,
        allow_unicode=True,
        verbose_name=_("slug"),
    )
```
I guess it makes sense to also add a slug identifier in addition to the `id` field.

integreat_cms/cms/models/poi_categories/poi_category.py
```diff
@@ -0,0 +1,53 @@
...
+        blank=False,
+        verbose_name=_("category name"),
+        help_text=_("The name of the POI category."),
+    )
+    english_category_name = models.CharField(
+        max_length=250,
+        blank=False,
+        verbose_name=_("english category name"),
+        help_text=_("The english name of the POI category."),
+    )
```
timoludwig, line 23:
```suggestion
```



I'm not sure how that's supposed to work. Since the categories should be used in the apps, it's not sufficient to translate them into German and English. Instead, we need one translation per language that exists in the CMS. In other words, we probably need an additional translation model which holds two foreign keys - one to the `POICategory`, one to the `Language` model, and one field which contains the translated category name.
In the CMS, we should probably show all those translated fields one one page, so one doesn't have to reload the page for every single translation in other languages ([formsets](https://docs.djangoproject.com/en/3.2/topics/forms/formsets/) is the keyword here).

```
class POICategoryTranslation(AbstractBaseModel):
    """
    Data model representing a POI category translation.
    """

    category = models.ForeignKey(
        POICategory,
        on_delete=models.CASCADE,
        verbose_name=_("category"),
    )
    language = models.ForeignKey(
        Language,
        on_delete=models.CASCADE,
        verbose_name=_("language"),
    )
    name = models.CharField(
        max_length=250,
        blank=False,
        verbose_name=_("category name"),
        help_text=_("The name of the POI category."),
    )
```

timoludwig (commented):
Thanks a lot! :+1:

In general, this is a good start, I'm just thinking that we need to rework the translation part, since these categories are to be used in the apps...

Also, could you deliver the new field "category" in the POI API response? Thanks!

timoludwig (changes requested):
I'm going to change the state now from "review required" to "changes requested".
@MizukiTemma feel free to re-request a review once you implemented the current suggestions.

MizukiTemma:
@timoludwig 
Thank you for your review :D
After changes I'll re-request review.

integreat_cms/cms/migrations/0037_poicategory.py
```diff
@@ -0,0 +1,51 @@
...
+                        auto_created=True,
+                        primary_key=True,
+                        serialize=False,
+                        verbose_name="ID",
+                    ),
+                ),
+                (
+                    "category_name",
+                    models.CharField(
+                        help_text="The name of the poi category.",
```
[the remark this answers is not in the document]
MizukiTemma: Oh :D That's nice. Thank you!

MizukiTemma:
@timoludwig 
Thank you for suggestions :) The following changes are now implemented:
- translation model for POI Category (POICategoryTranslation)
- form was modified so that translations for each languages can be edited after creating a poi category first time
- overview of translation availability for each language is added in the poi category list
- new field "category" is added in the POI API response
- squashing/cleaning the migrarion file

PS. I have no idea what is missing for build-documentation......

ulliholtgrave:
@MizukiTemma This documentation problem is probably related to some problems with the internal links that can be solved afterwards. However, there are some other checks failing. Can you fix them before? E.g. Adjust the tests for the new output

MizukiTemma:
@ulliholtgrave 
The other checks are now passed and the test is adjusted.

timoludwig (commented):
@MizukiTemma Thanks a lot! :+1:

Everything works nearly as expected!
I really like your solution, especially the formset usage. :ok_hand: 

Initially I only had a few ideas for improvements, but my urge for perfectionism led me down a rabbit hole, so I ended up changing quite a few things which go beyond the GitHub suggestion feature, so I pushed my suggestions directly to your branch.

The main points I addressed:
- The `POICategory` had fields which belong to the translation model
- It was required to click the "save" button once until the translations could be created
- Translation objects were created even if the name field was empty
- Adding languages after creating a POI category broke the form workflow
- There were no permission checks for the form view
- I added a few test cases for the categories
- I refactored the views to use more of Django's class based views
- I implemented the possibility to delete categories
- I added a separate API endpoint for the categories

Could you have a look at the new state and test whether everything works for you or if you would like to change something? Thanks!

MizukiTemma:
@timoludwig 
Thank you for your improvement 😍 It's so beautiful!!!!!!!

timoludwig (approved):

ulliholtgrave (approved):
Thanks you two, that looks very neat :+1:

Issue #1678, opened 2022-09-13
Internal server error when creating new page translations


timoludwig:
### Describe the Bug

The page form throws an internal server error when the translation does not yet exist

### Steps to Reproduce

1. Go to a page form where the translation does not yet exist

### Expected Behavior

A new translation form should be opened

### Actual Behavior

An internal server error occurs

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
Sep 13 08:33:31 ERROR   django.request - Internal Server Error: /lahndillkreis/pages/fr/5665/edit/
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/views/pages/page_form_view.py", line 92, in get
    if page_translation.status != status.PUBLIC:
AttributeError: 'NoneType' object has no attribute 'status'
  ```

</details>

PR #1679, opened 2022-09-13
Fix `AttributeError` when creating new page translations

timoludwig:
### Short description

Fix `AttributeError` when creating new page translations

### Proposed changes

- Check whether the page translation exists before accessing its status

### Resolved issues

Fixes: #1678


ulliholtgrave (approved):
Good catch! :+1: Thag fixes the issue :)

Issue #1653, opened 2022-08-26
"Minor Edit"-Button changes translation status nevertheless


dkehne:
The "Minor Change - please do not translate button" does not work. The translations are updated nevertheless.

### Steps to Reproduce
I updated the icon of this page: https://admin.integreat-app.de/bochum/pages/de/1252/edit/
I checked the "Geringfügige Änderung, please do not update translation status"-mark and clicked on save
The translation status of alle translations changed.

### Expected Behavior
Translation status should NOT be updated, when the checkbox is ticked.

dkehne:
Prio High.

svenseeberg:
I cannot confirm that the problem exists. When setting the minor edit flag the status of depending translations does not change when changing the page icon or content.

timoludwig:
I think the problem arose because no changes were made to the content of the translation - so no new version was created, and just the "minor edit" field of the existing translation was toggled and the date was updated, which caused the dependent translations to be outdated.

One way to fix this issue would be to create new translation versions even if the content did not change... this inflates the version history a bit, but would reflect the translation status better.

Another possibility would be to not update the timestamp of the translation when no changes are made, but this behavior was explicitly requested to be able to refresh the date without performing changes.

dkehne:
I do know if the fact that live-content is also used on this page, matters.

svenseeberg:
> One way to fix this issue would be to create new translation versions even if the content did not change... this inflates the version history a bit, but would reflect the translation status better.

Agreed, we should probably do that.

ulliholtgrave:
> I think the problem arose because no changes were made to the content of the translation - so no new version was created, and just the "minor edit" field of the existing translation was toggled and the date was updated, which caused the dependent translations to be outdated.

But as we confirm the up-to-dateness of a translation with no changing content, we should probably consider not changing the the translation status at all. The content is the same and in my opinion the "marks as up-to-date" functionality should be independent.

timoludwig:
> But as we confirm the up-to-dateness of a translation with no changing content, we should probably consider not changing the the translation status at all. The content is the same and in my opinion the "marks as up-to-date" functionality should be independent.

Yes, I think this would be the case when we just create a new translation object (with automatically setting the minor edit field to true) instead of modifying the last updated date of the existing translation (like we do at the moment).

PR #1691, opened 2022-09-16
Fix translation status in edge cases

timoludwig:
### Short description

The translation status was broken in certain edge cases and this PR hopefully fixes this by always creating new versions, even if the content did not change.
Also, I refactored some things I stumbled over while solving this issue.

### Proposed changes

- Remove code duplication in content models
- Create new versions even if content did not change to preserve translation status (fixes: #1653)
- Inject more settings into template context
- Refactor translation status buttons
    - Pass translation status directly via the value of the submission buttons
    - Inject the status constants in all templates
    - Remove long `if`/`elif` blocks to determine the status

### Resolved issues

Fixes: #1653


ulliholtgrave:
I am not really sure if this problem was introduced in this PR, but I am no longer able to update pages without being a minor edit :D
Even if I change something, set "minor edit" to false and click update, "minor edit" is set after the redirect and the status of all pages is set to "up-to-date". 
grafik

timoludwig (commented):
> I am not really sure if this problem was introduced in this PR, but I am no longer able to update pages without being a minor edit :D Even if I change something, set "minor edit" to false and click update, "minor edit" is set after the redirect and the status of all pages is set to "up-to-date". 

Oh yes, this was my fault. Just a `not` too much, so it did the exact opposite of what it was supposed to do :sweat_smile:

MizukiTemma (approved):
I think too this solves the problem and brings no negative effect on other functionalities 👍

PR #1693, opened 2022-09-16
Fix JS error in page form

timoludwig:
### Short description

Fix small JS error in page form

### Proposed changes

- Fix `Uncaught TypeError: toggleButton is null`


ulliholtgrave (approved):
Thanks :+1:

Issue #1683, opened 2022-09-13
Additional organization management


timoludwig:
### Motivation

With #1668, the can now be managed per region.
However, a few adaptions are needed before region managers can use the feature properly.

I see main two problems:
1. They do not understand what a "slug" is
2. They don't see at a glance which members and pages an organizations has

### Proposed Solution

1. Remove the slug field from the form and automatically derive it from the name
2. Add two additional boxes where the users and pages of the current organizations are listed
3. Grant the management role to edit organizations

PR #1712, opened 2022-09-23
Additional organization management

JanErikHuehne:
### Short description

The organization management inside a region has been modified. 

### Proposed changes

- Removing the slug input field when creating a new organization (the slug is now creating internally and it is taken care of that it is unique)
- Moved the deletion button of an already existing organization up in the page
- Added a display of maintained pages and members of that organization 
- Added permissions of management role to edit and delete organizations 

### Resolved issues
Fixes: #1683

### Aspects that need additional adjustments / input welcomed

- So I just added a template for the display of pages that are maintained by an organization. I have trouble figuring out a good way of displaying the information of the pages, since they might differ in the available languages etc. 
So ideas about the following point are very well welcomed :innocent: :

How to display the pages in the organization edit page (what information, should we add a language selecting gadget here as well?)
- So I implemented one possible way of selecting a organization slug in the following template: {organization_name}_{Region_Name}_{counter_if_necessary}. Maybe there is a better way to deal with organization with the same name in the same region (not allow it?) ?
- I´m not sure about the changes to the permissions of the management role, since the required permissions were already displayed at the user roles even before the made changes. 
___________________________________________________


integreat_cms/cms/templates/organizations/member_list_row.html
```diff
@@ -0,0 +1,15 @@
...
+    <td class="text-sm text-left py-3 pl-4 pr-2">
+            {{ member.username }}
+    </td>
+    
+    <td class="text-sm text-left py-3 pl-4 pr-2">
+        {% if member.role %}
+            {{ member.role }}
+        {% else %}
+          - 
+        {% endif %}
```
timoludwig, line 13:
```suggestion
        {{ member.role }}
```
It's not possible to create users without a role anyway

integreat_cms/cms/templates/organizations/organization_form.html
```diff
@@ -15,9 +18,25 @@ <h1 class="heading">
...
+					<button title="{% trans 'Delete organization' %}" class="btn confirmation-button btn-red"
+							data-confirmation-title="{{ delete_dialog_title }}"
+							data-confirmation-text="{{ delete_dialog_text }}"
+							data-confirmation-subject="{{ form.name.value }}"
+							data-action="{% url 'delete_organization' region_slug=request.region.slug slug=form.instance.slug %}">
+						<i data-feather="trash-2" class="mr-2"></i>
+						{% trans 'Delete this organization' %}
+					</button>
+				</div>
+			{% endif %}
```
timoludwig, line 38:
```suggestion
			{% if form.instance.id and perms.cms.delete_organization %}
				<div class="flex flex-wrap gap-4">
					<button title="{% trans 'Delete organization' %}" class="btn confirmation-button btn-red"
							data-confirmation-title="{{ delete_dialog_title }}"
							data-confirmation-text="{{ delete_dialog_text }}"
							data-confirmation-subject="{{ form.name.value }}"
							data-action="{% url 'delete_organization' region_slug=request.region.slug slug=form.instance.slug %}">
						<i data-feather="trash-2" class="mr-2"></i>
						{% trans 'Delete this organization' %}
					</button>
				</div>
			{% endif %}
			{% if perms.cms.change_organization %}
				<div class="flex flex-wrap gap-4">
					<button class="btn">{% trans 'Save' %}</button>
				</div>
			{% endif %}
```
I'd put the deletion button to the left

integreat_cms/cms/templates/organizations/organization_form.html
```diff
@@ -43,20 +59,85 @@ <h3 class="heading font-bold text-black"><i data-feather="settings" class="mr-2"
...
+			<h3 class="heading font-bold text-black"> <i data-feather="layout" class="mr-2"></i>
+				{% trans "Pages" %}
+			</h3>
+			</div>
+			<div class="table-listing w-full p-2">
+				<table class="w-full mt-4 rounded border border-solid border-gray-200 shadow bg-white">
+					<thead>
+						<tr class="border-b border-solid border-gray-200">
+							<th class="text-sm text-left uppercase py-3 pl-4 pr-2">
+								{% trans "Name in Englisch" %}
```
timoludwig, line 78: > How to display the pages in the organization edit page (what information, should we add a language selecting gadget here as well?)

I think a language selector would be overkill, but maybe two columns: The region's default language and the current backend UI language? (And if both are identical, then just one column?)

timoludwig (changes requested):
Thanks a lot, this is really helpful in managing the organizations! :muscle: 

> Maybe there is a better way to deal with organization with the same name in the same region (not allow it?) ?

Yes sounds reasonable to me. Probably, we should add a separate clean method for the name field and prevent that two organizations with the same name in the same region are created...

> I´m not sure about the changes to the permissions of the management role, since the required permissions were already displayed at the user roles even before the made changes.

I'm not sure how it happened, but when I start on the develop branch, execute the `prune_database.sh` dev tool and start the server again, the management role does not have the permission to view or edit organizations...

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -4907,24 +4909,46 @@ msgstr "Noch keine Angebots-Vorlagen vorhanden."
...
+msgid "Name in Englisch"
+msgstr "Name auf Englisch"
+
+#: cms/templates/organizations/organization_form.html:90
+msgid "This organization currently has no maintained pages."
+msgstr "Von dieser Organization werden zur Zeit keine Seiten verwaltet."
+
+#: cms/templates/organizations/organization_form.html:104
+msgid "Members"
+msgstr "November"
```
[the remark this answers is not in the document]
JanErikHuehne: Good to know! :+1:

integreat_cms/cms/forms/organizations/organization_form.py
```diff
@@ -1,6 +1,10 @@
+from distutils.command.clean import clean
```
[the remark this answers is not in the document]
JanErikHuehne: I have no idea where this comes from, might be some IDE auto suggestion that I missed.

integreat_cms/cms/forms/organizations/organization_form.py
```diff
@@ -1,6 +1,10 @@
...
 from ...models import Organization
 from ..icon_widget import IconWidget
 from ..custom_model_form import CustomModelForm
+import logging
```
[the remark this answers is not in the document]
JanErikHuehne: fixed

integreat_cms/cms/templates/organizations/organization_form.html
JanErikHuehne: Alright :D

integreat_cms/cms/templates/organizations/member_list_row.html
JanErikHuehne: fixed

integreat_cms/cms/templates/organizations/organization_form.html
JanErikHuehne: do we want to add a redirect to the pages somewhere or just simply display them?

timoludwig: Yes, good idea :+1:
Maybe just make the table cells link to the page form (similar to the regular page tree)?

JanErikHuehne: Yes, thats what I had in mind as well  :+1:

integreat_cms/cms/constants/roles.py
```diff
@@ -75,15 +75,16 @@
     "change_imprintpage",
     "change_pushnotification",
     "change_user",
-    "change_chatmessage",
+    "change_organization" "change_chatmessage",
```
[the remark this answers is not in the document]
JanErikHuehne: Is is sufficient to add the permissions in [here](https://github.com/digitalfabrik/integreat-cms/blob/184f0302cd573e4c2ff19f9b343ea9013e84fc71/integreat_cms/cms/migrations/0002_roles.py) and not to create a new migration file?

integreat_cms/cms/templates/organizations/pages_list_row.html
```diff
@@ -0,0 +1,8 @@
+{% load i18n %}
+
+<tr class="border-b border-solid border-gray-200">
+    <td class="text-sm text-left py-3 pl-4 pr-2">
+            {{ page_translation.title }}
+    </td>
+    
+</tr>
```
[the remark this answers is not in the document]
JanErikHuehne: Fixed

integreat_cms/cms/constants/roles.py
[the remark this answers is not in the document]
JanErikHuehne: Update: Just adding it in roles.py does the trick already. I tested it without and with the changes by logging in as as management user. Before the changes you can only view organizations, afterwards you can also delete and create them. I feel like it does it since it is used in the update roles migration :question:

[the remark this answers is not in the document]
timoludwig: Yeah, the problem is that this only works for newly set up systems. Django keeps track of which migrations were already applied and on production systems which already have the migration `0002_roles` applied, nothing will change. You can simulate this behavior by:
1. Check out the `develop` branch
2. Prune the database
3. Run the server
4. Stop it again
5. Check out the `feature/better_organization_management` branch
6. Run the server again

Then, it should have the `0002_roles` migration applied, but in its develop-state, so without the additional permission...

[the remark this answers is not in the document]
JanErikHuehne: Ah that makes sense, I will put it in a new migration file :+1:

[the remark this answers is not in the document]
JanErikHuehne: Alright, it should be fixed now

JanErikHuehne:
I added a check for a already existing name in the organization forms clean method. We might want to check later if we also want to not allow similar names such as names that only differ in a lower case / upper case character

JanErikHuehne:
@timoludwig  As for the tests, For the permissions they also rely on the roles.py defined in constants. Should I add the new permissions there anyways, so that the tests get the current permissions or add them in the conftest.py file manually?

integreat_cms/cms/forms/organizations/organization_form.py
```diff
@@ -26,3 +34,48 @@ class Meta:
...
+    def clean(self):
+        """
+        Validate if form fields name is not already in use for another organization in the same region, see :meth:`django.forms.Form.clean`
+
+        :return: The cleaned form data
+        :rtype: dict
+        """
+        cleaned_data = super().clean()
+
+        if Organization.objects.filter(name=cleaned_data["name"]).exists():
```
timoludwig, line 71: > I added a check for a already existing name in the organization forms clean method.

Thanks! Three small points:
1. If you validate only one field, I'd prefer the `clean_<fieldname>` over the general `clean()` method.
2. If you don't exclude the current organization instance, you can only create organizations and not save them (to e.g. change the icon), because the currently edited organization blocks its own name...
3. You did not limit the queryset to the current region, so it checks whether the value is in use globally instead of only region-wide.


```suggestion
    def clean_name(self):
        """
        Validate if form fields name is not already in use for another organization in the same region
        (see :ref:`overriding-modelform-clean-method`)

        :return: The name which is unique per region
        :rtype: str
        """
        cleaned_name = self.cleaned_data["name"]
        if (
            Organization.objects.exclude(id=self.instance.id)
            .filter(region=self.instance.region, name=cleaned_name)
            .exists()
        ):
```

> We might want to check later if we also want to not allow similar names such as names that only differ in a lower case / upper case character

Yeah, I think for now it's ok, no need for a more sophisticated check.

integreat_cms/cms/forms/organizations/organization_form.py
```diff
@@ -26,3 +34,48 @@ class Meta:
...
+            self.add_error(
+                "name",
+                forms.ValidationError(
+                    _(
+                        "A organization with the same name already exists in this region. Please choose another name."
```
timoludwig, line 76:
```suggestion
                        "An organization with the same name already exists in this region. Please choose another name."
```

integreat_cms/cms/forms/organizations/organization_form.py
```diff
@@ -26,3 +34,48 @@ class Meta:
...
+                    ),
+                    code="invalid",
+                ),
+            )
+        return cleaned_data
```
timoludwig, line 81:
```suggestion
        return cleaned_name
```

timoludwig (changes requested):

> @timoludwig As for the tests, For the permissions they also rely on the roles.py defined in constants. Should I add the new permissions there anyways, so that the tests get the current permissions or add them in the conftest.py file manually?

Not in `conftest.py`, but we need to adapt the view config, yes:
https://github.com/digitalfabrik/integreat-cms/blob/4b51ce11662f045a2b7b13ed2cd08622c6c59b30/tests/cms/views/view_config.py#L102-L103
(`STAFF_ROLES]` should be replaced by `STAFF_ROLES + [MANAGEMENT]`) because we expect the test to succeed for the management role as well.

And also, for consistency reasons, I think it's best to add the permissions to the [/integreat_cms/cms/constants/roles.py](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/constants/roles.py) as well, even though you didn't use them in the migration file.
But sometimes when we do more complex permission updates, we use migrations like [this one](https://github.com/digitalfabrik/integreat-cms/blob/4b51ce11662f045a2b7b13ed2cd08622c6c59b30/integreat_cms/cms/migrations/0035_alter_chatmessage_options.py#L26-L33) and we don't want the permissions to be reset afterwards.

JanErikHuehne:
> (`STAFF_ROLES]` should be replaced by `STAFF_ROLES + [MANAGEMENT]`) because we expect the test to succeed for the management role as well.

I´m sorry for the delay, I moved these past days. As for the change in conftest.py I added these roles for the Augsburg test region. Adding them to the Nuernberg region will result in an error, since the management user that is used is listed for the Augsburg region. i hope this is how it is intended.

integreat_cms/cms/forms/organizations/organization_form.py
JanErikHuehne: Fixed

integreat_cms/cms/forms/organizations/organization_form.py
JanErikHuehne: Adjusted :+1:

integreat_cms/cms/constants/roles.py
```diff
@@ -75,15 +75,16 @@
...
     "delete_directory",
     "delete_feedback",
-    "delete_mediafile",
+    "delete_organization" "delete_mediafile",
```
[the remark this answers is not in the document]
JanErikHuehne: As we discussed I kept it in there

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -1800,6 +1802,14 @@ msgstr ""
 "Die Datei kann nicht in einem Verzeichnis einer anderen Region hochgeladen "
 "werden."
 
+#: cms/forms/organizations/organization_form.py:79
+msgid ""
+"An organization with the same name already exists in this region. Please "
+"choose another name."
+msgstr ""
+"Eine Organization mit diesem Namen existiert bereits in dieser Region. Bitte "
```
timoludwig, line 1810:
```suggestion
"Eine Organisation mit diesem Namen existiert bereits in dieser Region. Bitte "
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -4907,24 +4922,55 @@ msgstr "Noch keine Angebots-Vorlagen vorhanden."
...
+
+#: cms/templates/organizations/organization_form.html:195
+#: cms/templates/users/region_user_list.html:21
+#: cms/templates/users/user_list.html:29
+msgid "Username"
+msgstr "Benutzername"
+
+#: cms/templates/organizations/organization_form.html:217
+msgid "This organization currently has no members."
+msgstr "Diese Organization has zur Zeit keine Mitglieder."
```
timoludwig, line 4972:
```suggestion
msgstr "Diese Organisation hat zur Zeit keine Mitglieder."
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -4907,24 +4922,55 @@ msgstr "Noch keine Angebots-Vorlagen vorhanden."
...
+#: cms/templates/organizations/organization_form.html:95
+#: cms/templates/organizations/organization_form.html:138
+#: cms/templates/organizations/organization_form.html:152
+msgid "View page"
+msgstr "Weitere Seiten-Infos"
+
+#: cms/templates/organizations/organization_form.html:108
+#: cms/templates/organizations/organization_form.html:170
+msgid "This organization currently has no maintained pages."
+msgstr "Von dieser Organization werden zur Zeit keine Seiten verwaltet."
```
timoludwig, line 4958:
```suggestion
msgstr "Von dieser Organisation werden zur Zeit keine Seiten verwaltet."
```

integreat_cms/cms/templates/organizations/organization_form.html
```diff
@@ -15,9 +19,25 @@ <h1 class="heading">
...
+		<div class="flex flex-wrap justify-between gap-4"> 
+			{% if form.instance.id and perms.cms.delete_organization %}
+				<div class="flex flex-wrap gap-4">
+					<button title="{% trans 'Delete organization' %}" class="btn confirmation-button btn-red"
+							data-confirmation-title="{{ delete_dialog_title }}"
+							data-confirmation-text="{{ delete_dialog_text }}"
+							data-confirmation-subject="{{ form.name.value }}"
+							data-action="{% url 'delete_organization' region_slug=request.region.slug slug=form.instance.slug %}">
+						<i data-feather="trash-2" class="mr-2"></i>
+						{% trans 'Delete this organization' %}
```
timoludwig, line 31:
```suggestion
						{% trans 'Delete' %}
```
I think we should make these buttons more consistent.

integreat_cms/cms/templates/organizations/organization_form.html
```diff
@@ -43,20 +60,175 @@ <h3 class="heading font-bold text-black"><i data-feather="settings" class="mr-2"
...
+						<table class="w-full mt-4 rounded border border-solid border-gray-200 shadow bg-white">
+						
+							<thead>
+								<tr class="border-b border-solid border-gray-200">
+									<th class="text-sm text-left uppercase py-3 pl-4 pr-2">
+										{% trans "Name in " %} {{ region_default_language.translated_name }}
+									</th>
+									<th class="text-sm text-left uppercase py-3 pl-4 pr-2">
+										{% trans "Name in "%} {{ backend_language.translated_name }}
+									</th>
```
timoludwig, line 126:
```suggestion
									{% if region_default_language != backend_language %}
										<th class="text-sm text-left uppercase py-3 pl-4 pr-2">
											{% trans "Name in "%} {{ backend_language.translated_name }}
										</th>
									{% endif %}
```
Could you reduce code duplications by showing the default language in any case and just showing the additional column if the backend language differs? If you use the same condition in every row, I guess it's save to say that each row will have the same number of columns in the end.

integreat_cms/cms/templates/organizations/organization_form.html
```diff
@@ -43,20 +60,175 @@ <h3 class="heading font-bold text-black"><i data-feather="settings" class="mr-2"
...
+
+												</a>
+											</td>
+										{% else %}
+											<td class="text-sm text-left py-3 pl-4 pr-2">
+
+												<i>{% trans 'Translation not available' %}</i>
+
+											</td>
+										{% endif %}
```
timoludwig, line 164:
```suggestion
										{% if region_default_language != backend_language %}
											{% if backendlang_page_translation %}
												<td class="text-sm text-left py-3 pl-4 pr-2">
													<a href="{% url 'edit_page' page_id=page.id region_slug=request.region.slug language_slug=backend_language.slug %}" class="block py-1.5 px-2 overflow-hidden max-w-xs whitespace-nowrap text-ellipsis text-gray-800" title="{% trans 'View page' %}">
	
														{{ backendlang_page_translation.title }}
	
													</a>
												</td>
											{% else %}
												<td class="text-sm text-left py-3 pl-4 pr-2">
	
													<i>{% trans 'Translation not available' %}</i>
	
												</td>
											{% endif %}
										{% endif %}
```

integreat_cms/cms/templates/organizations/organization_form.html
```diff
@@ -43,20 +60,175 @@ <h3 class="heading font-bold text-black"><i data-feather="settings" class="mr-2"
...
+									<td colspan="6" class="px-4 py-3">
+										{% trans "This organization currently has no maintained pages." %}
+									</td>
+								</tr>
+							
+								{% endfor %}
+							</tbody>
+							
+						</table>
+					{% else %}
```
timoludwig, line 116:
```suggestion
```

integreat_cms/cms/templates/organizations/organization_form.html
```diff
@@ -43,20 +60,175 @@ <h3 class="heading font-bold text-black"><i data-feather="settings" class="mr-2"
...
+										{% trans "This organization currently has no maintained pages." %}
+
+									</td>
+								</tr>
+							
+								{% endfor %}
+							</tbody>
+							
+						</table>
+					{% endif %}
```
timoludwig, line 179:
```suggestion
```

integreat_cms/cms/templates/organizations/organization_form.html
```diff
@@ -43,20 +60,175 @@ <h3 class="heading font-bold text-black"><i data-feather="settings" class="mr-2"
...
+									</th>
+								</tr>
+
+							</thead>
+
+							<tbody>
+								{% for page in form.instance.pages.all %}
+									{% get_translation page region_default_language.slug as regionlang_page_translation %}
+									{% get_translation page backend_language.slug as backendlang_page_translation %}
+									<tr class="border-b border-solid border-gray-200">
```
timoludwig, line 135:
```suggestion
									<tr class="border-b border-solid border-gray-200 hover:bg-gray-100">
```
Maybe a small hover effect similar to the page tree?

timoludwig (changes requested):
Awesome, thanks a lot! :tada:

Everything looks nearly perfect, I just have a few ideas for small aesthetic changes and one possibility to reduce a few lines of code.

> I´m sorry for the delay, I moved these past days.

Absolutely no worries, happy settling in :wink:

> As for the change in conftest.py I added these roles for the Augsburg test region. Adding them to the Nuernberg region will result in an error, since the management user that is used is listed for the Augsburg region. i hope this is how it is intended.

Yes, exactly! :+1:
We use the tests to check whether access to the region "augsburg" works and whether access to the region "nurnberg" is denied as expected.

integreat_cms/cms/templates/organizations/organization_form.html
JanErikHuehne: Fixed

integreat_cms/cms/templates/organizations/organization_form.html
JanErikHuehne: Fixed, at least I hope I did is as you intended it. The code suggestions here are a bit hard to read :D

integreat_cms/cms/templates/organizations/organization_form.html
JanErikHuehne: Yeah sure. Added it

JanErikHuehne:
> Absolutely no worries, happy settling in wink

Haha, thanks! Quite a trip down here

JanErikHuehne:
Should be good now. I will rebase it on top of the develop branch

.circleci/config.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     steps:
       - checkout
       - restore_cache:
-          key: pipenv-{{ checksum "Pipfile.lock" }}-v1
+          key: pipenv-{{ checksum "Pipfile.lock" }}-v2
```
timoludwig, line 14:
```suggestion
          key: pipenv-{{ checksum "Pipfile.lock" }}-v1
```
After rebasing, the increment should no longer be required (but yeah, it doesn't hurt either :laughing:)

integreat_cms/cms/constants/roles.py
```diff
@@ -73,16 +73,19 @@
 MANAGEMENT_PERMISSIONS = EDITOR_PERMISSIONS + [
     "change_feedback",
     "change_imprintpage",
+    "change_organization",
```
timoludwig, line 76: Ah sorry, I totally forgot: Could you grant the `view_organization` permission to the marketing team and the `change_organization` and `delete_organization` permissions to the service team?

integreat_cms/cms/forms/organizations/organization_form.py
```diff
@@ -26,3 +34,51 @@ class Meta:
...
+
+        :param \**kwargs: The supplied keyword arguments
+        :type \**kwargs: dict
+        """
+        super().__init__(**kwargs)
+        if self.instance:
+            # logger.info(Organization.users.all())
+            logger.info(self.instance.slug)
+            logger.info(self.instance.pages.all())
+            # logger.info(self.instance.users.all())
```
timoludwig, line 50:
```suggestion
```
The instance is already logged [here](https://github.com/digitalfabrik/integreat-cms/blob/5c3265e28770a5545ddd9dc881ecfef41b8e6c25/integreat_cms/cms/forms/custom_model_form.py#L47-L61).

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -1766,7 +1766,9 @@ msgstr ""
 
 #: cms/forms/media/create_directory_form.py:66
 msgid "A directory with the name \"{}\" already exists."
-msgstr "Ein Ordner mit diesem Namen existiert bereits"
+msgstr ""
+"Ein Ordner mit diesem Namen existiert bereitstext-sm text-left py-3 pl-4 "
+"pr-2 text-ellipsis text-gray-800"
```
timoludwig, line 1771:
```suggestion
"Ein Ordner mit diesem Namen existiert bereits"
```

integreat_cms/cms/migrations/0038_add_organization_permissions_to_management_role.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    # We can't import the Person model directly as it may be a newer
+    # version than this migration expects. We use the historical version.
+    Group = apps.get_model("auth", "Group")
+    Permission = apps.get_model("auth", "Permission")
+
+    management_group = Group.objects.filter(name="MANAGEMENT").first()
+    management_group.permissions.remove(
+        *Permission.objects.filter(codename__in=NEW_MANAGEMENT_PERMISSIONS)
+    )
```
timoludwig, line 52:
```suggestion
    Group = apps.get_model("auth", "Group")
    Permission = apps.get_model("auth", "Permission")

    # Clear and update permissions according to new constants
    for role_name in dict(roles.CHOICES):
        group = Group.objects.filter(name=role_name).first()
        # Clear permissions
        group.permissions.clear()
        # Set permissions
        group.permissions.add(
            *Permission.objects.filter(codename__in=roles.PERMISSIONS[role_name])
        )
```
Since we're changing permissions quite frequently, I guess it's justifiable to use a more generic approach here (with the advantage that we can always be 100% sure that they are in sync with the constants module and the disadvantage that we cannot reverse them)...

integreat_cms/cms/migrations/0038_add_organization_permissions_to_management_role.py
```diff
@@ -0,0 +1,70 @@
...
+from django.db import migrations
+
+
+NEW_MANAGEMENT_PERMISSIONS = [
+    "change_organization",
+    "view_organization",
+    "delete_organization",
+]
+# pylint: disable=unused-argument
+def add_organization_mangement_permissions(apps, schema_editor):
```
timoludwig, line 11:
```suggestion
# pylint: disable=unused-argument
def update_roles(apps, schema_editor):
```

integreat_cms/cms/migrations/0038_add_organization_permissions_to_management_role.py
```diff
@@ -0,0 +1,70 @@
...
+
+    dependencies = [
+        ("cms", "0037_event_datetime"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            add_organization_mangement_permissions,
+            remove_organization_mangement_permissions,
+        ),
```
timoludwig, line 69:
```suggestion
        migrations.RunPython(update_roles, migrations.RunPython.noop),
```

timoludwig (changes requested):
Great, thanks a lot! :rocket: :muscle: 

One additional idea I had:
Could you remove the "slug" column in the organization list and replace if by the columns "number of pages" and "number of members"?

integreat_cms/cms/templates/organizations/organization_form.html
timoludwig: If it's too hard to read you can also commit the suggested changes directly via the GitHub UI and hope I didn't mess up anything :see_no_evil:

integreat_cms/cms/constants/roles.py
timoludwig: Oh, so sorry, the marketing team already has the permission to view them and the service team already the permission to change them...
So please only grant the `delete_organization` permission to the service team :slightly_smiling_face:

.circleci/config.yml
JanErikHuehne: I know, but wanted to do the rebase close to merging so I dont have to do it again when things are merged before this one :smile:

integreat_cms/cms/forms/organizations/organization_form.py
JanErikHuehne: your right :D

integreat_cms/cms/migrations/0038_add_organization_permissions_to_management_role.py
JanErikHuehne: Yes, but I think it is kind of odd since it is basically the same as migration no. 25. But if you want we can do it that way

JanErikHuehne:
> Great, thanks a lot! rocket muscle
> 
> One additional idea I had: Could you remove the "slug" column in the organization list and replace if by the columns "number of pages" and "number of members"?

Seems like a good idea. I will  implement it

.circleci/config.yml
timoludwig: Ah ok :smile: 
Probably you already knew, but just in case you didn't: I wrote a small [bash script](https://github.com/digitalfabrik/integreat-cms/blob/develop/dev-tools/resolve_translation_conflicts.sh) to resolve the conflicts in the translation file automatically, so as long as you only have conflicts in the translation file (and you only have one commit), you can rebase automatically by executing:
```
git fetch --all && GIT_SEQUENCE_EDITOR=true git rebase origin/develop || ./dev-tools/resolve_translation_conflicts.sh && GIT_EDITOR=true git rebase --continue && git push --force-with-lease
```

integreat_cms/cms/migrations/0038_add_organization_permissions_to_management_role.py
timoludwig: Yeah, merging those two won't work for the [same reason](https://github.com/digitalfabrik/integreat-cms/pull/1712#discussion_r979807938) we cannot put the code in the initial role migration `0002_roles.py`: The production system already knows that it has applied the migration `0025_update_roles.py` and won't apply it again, even if we change it.
In the end, the explicit role change you suggested is better 1. because it's more clear what happens and 2. it's reversible, but it's also more work, so if you don't want to add the role for the service team manually as well, I'm also fine with the more generic approach (which just needs to be copied + pasted each time a permission is changed)...

JanErikHuehne: Makes sense, so I did implement it in the lazy way now :D

integreat_cms/cms/models/users/organization.py
```diff
@@ -61,6 +61,23 @@ def get_repr(self):
...
         return f"<Organization (id: {self.id}, slug: {self.slug}, region: {self.region.slug})>"
 
+    @property
+    def num_pages(self):
+        """
+
+        :return: the current number of maintained pages of an organization object
+        :rtype: int
+        """
+        return self.pages.count
```
timoludwig, line 71:
```suggestion
        return self.pages.count()
```
In templates, this works since callables are automatically called, but if this property is used e.g. in a view, this would return only the method `count` and not the actual value.

integreat_cms/cms/models/users/organization.py
```diff
@@ -61,6 +61,23 @@ def get_repr(self):
...
+
+    @property
+    def num_members(self):
+        """
+        :return: the current number of members of an organization object
+        :rtype: int
+        """
+        return self.members.count
```
timoludwig, line 79:
```suggestion
        return self.members.count()
```

integreat_cms/cms/templates/organizations/organization_form.html
```diff
@@ -43,20 +60,134 @@ <h3 class="heading font-bold text-black"><i icon-name="settings" class="mr-2"></
...
+											</td>
+										{% else %}
+											<td class="text-sm text-left py-3 pl-4 pr-2">
+												<i>{% trans 'Translation not available' %}</i>
+											</td>
+										{% endif %}
+
+										{% if region_default_language != backend_language %}
+											{% if backendlang_page_translation %}
+											{% endif %}
```
timoludwig, line 114:
```suggestion
```
This `endif` tag is placed incorrectly

integreat_cms/cms/templates/organizations/organization_form.html
```diff
@@ -43,20 +60,134 @@ <h3 class="heading font-bold text-black"><i icon-name="settings" class="mr-2"></
...
+														{{ backendlang_page_translation.title }}
+
+													</a>
+												</td>
+											{% else %}
+												<td class="text-sm text-left py-3 pl-4 pr-2">
+
+													<i>{% trans 'Translation not available' %}</i>
+
+												</td>
```
timoludwig, line 127:
```suggestion
												</td>
											{% endif %}
```
It belongs here

timoludwig (changes requested):
Thanks a lot! :+1:
Just two small issues remaining :slightly_smiling_face: 
And could you add a changelog entry for this?

integreat_cms/cms/templates/organizations/organization_form.html
```diff
@@ -15,9 +19,25 @@ <h1 class="heading">
...
-		{% endif %}
+		<div class="flex flex-wrap justify-between gap-4"> 
+			{% if form.instance.id and perms.cms.delete_organization %}
+				<div class="flex flex-wrap gap-4">
+					<button title="{% trans 'Delete organization' %}" class="btn confirmation-button btn-red"
+							data-confirmation-title="{{ delete_dialog_title }}"
+							data-confirmation-text="{{ delete_dialog_text }}"
+							data-confirmation-subject="{{ form.name.value }}"
+							data-action="{% url 'delete_organization' region_slug=request.region.slug slug=form.instance.slug %}">
+						<i data-feather="trash-2" class="mr-2"></i>
```
timoludwig, line 30:
```suggestion
						<i icon-name="trash-2" class="mr-2"></i>
```
Ah, and we currently changed our icon dependency

integreat_cms/cms/templates/organizations/organization_form.html
```diff
@@ -43,20 +60,134 @@ <h3 class="heading font-bold text-black"><i icon-name="settings" class="mr-2"></
...
-		</button>
+	{% if form.instance.id %}
+
+	{% get_current_language as LANGUAGE_CODE %}
+	{% get_language LANGUAGE_CODE as backend_language %}
+	{% with request.region.default_language as  region_default_language %}
+		<div class="grid xl:grid-cols-2 gap-4">
+			<div class="mb-4 rounded border border-solid border-blue-500 shadow-2xl bg-white">
+				<div class="p-4 rounded bg-water-500">
+				<h3 class="heading font-bold text-black"> <i data-feather="layout" class="mr-2"></i>
```
timoludwig, line 71:
```suggestion
				<h3 class="heading font-bold text-black">
				    <i icon-name="layout" class="mr-2"></i>
```

integreat_cms/cms/templates/organizations/organization_form.html
```diff
@@ -43,20 +60,134 @@ <h3 class="heading font-bold text-black"><i icon-name="settings" class="mr-2"></
...
+							
+								{% endfor %}
+							</tbody>
+						</table>
+				</div>
+			</div>
+		{% endwith %}
+		<div class="mb-4 rounded border border-solid border-blue-500 shadow-2xl bg-white">
+			<div class="p-4 rounded bg-water-500">
+				<h3 class="heading font-bold text-black"> <i data-feather="users" class="mr-2"></i>
```
timoludwig, line 145:
```suggestion
				<h3 class="heading font-bold text-black">
				    <i icon-name="users" class="mr-2"></i>
```

timoludwig (changes requested):

integreat_cms/cms/models/users/organization.py
JanErikHuehne: My bad. I did not notice it since I just used it in the template. :man_facepalming:

integreat_cms/cms/templates/organizations/organization_form.html
JanErikHuehne: :+1:

integreat_cms/cms/templates/organizations/organization_form.html
JanErikHuehne: Alright. Good to know for the future :D

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -1781,7 +1781,7 @@ msgstr ""
 
 #: cms/forms/media/create_directory_form.py:66
 msgid "A directory with the name \"{}\" already exists."
-msgstr "Ein Ordner mit diesem Namen existiert bereits"
+msgstr "Ein Ordner mit diesem Namen existiert bereitst"
```
timoludwig, line 1784:
```suggestion
msgstr "Ein Ordner mit diesem Namen existiert bereits"
```
Small typo

timoludwig (changes requested):
Thanks! :+1:
Sorry, one small typo has suddenly appeared :see_no_evil: 
And could you add a changelog entry?

JanErikHuehne:
> Thanks! +1
> Sorry, one small typo has suddenly appeared see_no_evil
> And could you add a changelog entry?

Oh, my bad. I think somethings went wrong at the rebasing, thats why the typo is there from an earlier commit and no changlog is added. I will add it. Sry for the mess, its been such a stressful week :D

integreat_cms/locale/de/LC_MESSAGES/django.po
JanErikHuehne: fixed

JanErikHuehne:
Alright, should be fixed/included now. If you spot anything just let me know and I will fix it.

timoludwig (approved):
Awesome, I think that's it! :tada: :muscle:

MizukiTemma:
@JanErikHuehne 
Thank you! It works as expected 😃 

When the merge conflict is resolved, I think this PR is ready for merge 👍

MizukiTemma (approved):

timoludwig:
@JanErikHuehne I couldn't resist including this feature in the current release. I'm sorry to have taken away the joy of merging. :laughing:

Issue #1526, opened 2022-06-06
Push news translations in main language not sent


svenseeberg:
### Describe the Bug
When a push notification is sent with the mode "Use main language if no translation is available", no translations beside the main language get sent.

### Steps to Reproduce

1. Go to New
2. Enter a notification text and title in the main language
3. Select the "Use main language if no translation is available" mode and send the notification
4. Look into the log file and observe that only one notifications (main language) has been sent, but no other translations

### Expected Behavior
Notifications in all translations should be sent.

### Actual Behavior
Because the form for all other translations is empty, no translation object is being created. However, during the sending process, the [`load_secondary_pnts` method](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/views/push_notifications/push_notification_sender.py#L59) iterates over all existing translations. Therefore, no additional notification besides the main translation is being sent.

### Solution
Iterate over all language tree nodes of the region, then check for an existing translation. If no translation exists, create a new PN translation and save the text to the database as well.

timoludwig:
> Solution
>
> Iterate over all language tree nodes of the region, then check for an existing translation. If no translation exists, create a new PN translation and save the text to the database as well.

Instead of saving the German text to all languages and create dummy translation objects in the database, we could also change the sending function to send the German translation for other languages as well without saving the content to the database.
We just have to think about whether we want the dummy translations in the API response as well - in this case, we'd need to create temporary translation objects in the api similar to what we do for fallback translations for events and POIs:
https://github.com/digitalfabrik/integreat-cms/blob/a5787fd1df20e1e3f19a7bc5f075f5ad5653f837/integreat_cms/cms/models/abstract_content_model.py#L213-L227

svenseeberg:
> We just have to think about whether we want the dummy translations in the API response as well - in this case, we'd need to create temporary translation objects in the api similar to what we do for fallback translations for events and POIs:

I guess we want that. Would you still dynamically create the content instead of just saving the source language for all other languages?

timoludwig:
> I guess we want that. Would you still dynamically create the content instead of just saving the source language for all other languages? I think there is a bigger chance of problems in the future with that approach. Having the translations in the database feels sort of safe to me.

[...]

PR #1713, opened 2022-09-23
Fix sending Push Notifications

svenseeberg:
### Short description
This PR fixes the handling of push notification translations when the main language should be used for other translations as well. If the main language should be used, only the notification via Firebase contains the original language. The translation in the API will contain a fallback translation that links to available translations.

The main challenge here is the default behavior of Django formsets. Django usually removes empty forms when validating/cleaning the formset. This PR introduces an override for one specific case: if the user chooses to use the main language for other translations as well, empty database objects will be created for empty push notification translations.

### Proposed changes
- Allow an empty title in the push notification translation model.
- Allow saving empty push notification translation forms when choosing the `USE_MAIN_LANGUAGE` mode.
- Change the API to deliver a fallback translation for empty database objects.
- The resulting translation object in the API looks like this:
  ```json
  [{"id": "81", "title": "Test DE", "message": "This page does not exist in the selected language. It is however available in these languages:\nDeutsch: https://integreat.app/testumgebung/de/news/local/80\nEnglish: https://integreat.app/testumgebung/en/news/local/82", "timestamp": "2022-09-23T16:39:54.843Z", "last_updated": "2022-09-23T18:39:54.843+02:00", "channel": "news"}]
  ```

### Side effects
- Push Notifications still not using unified translation models. I think that this is still a valid approach as the content model has much less complexity/information (no versions, no HTML body, etc).
- The new Formset model save method does not provide the commit parameter. I cannot think of a case where this should be an issue (at the moment).
- We do not have HTML in Push Notifications. The fallback translation links are plain text.

### Resolved issues

Fixes: #1526


osmers:
We had another question from a partner about this today - when can we close the issue?

timoludwig:
> We had another question from a partner about this today - when can we close the issue?

@digitalfabrik/cms-team anyone willing to review this? :innocent:

ulliholtgrave (changes requested):
I am not really sure about this, but I think that your changes are breaking our debug setup for FCM. Right now, I only can send PNs to our real FCM channel, but they are no longer printed to the log if our system is in DEBUG mode :thinking:

integreat_cms/cms/models/push_notifications/push_notification_translation.py
```diff
@@ -56,6 +55,43 @@ def search(cls, region, language_slug, query):
...
+        if (
+            self.push_notification.mode == pnt_const.USE_MAIN_LANGUAGE
+            and self.text == ""
+        ):
+            translations = "\n".join(
+                [
+                    f"{translation.language.native_name}: {settings.WEBAPP_URL}{translation.get_absolute_url()}"
+                    for translation in self.push_notification.translations.exclude(
+                        text=""
+                    )
```
timoludwig, line 89: Instead of `filter(~Q(text=""))` we can just use `exclude(text="")`.

integreat_cms/cms/forms/push_notifications/push_notification_form.py
```diff
@@ -7,6 +7,21 @@ class PushNotificationForm(CustomModelForm):
...
+        :param \**kwargs: The supplied keyword arguments
+        :type \**kwargs: dict
+        """
+        # Instantiate CustomModelForm
+        super().__init__(**kwargs)
+
+        # Make fields disabled when push notification was already sent
+        if self.instance.sent_date:
+            self.fields["channel"].disabled = True
+            self.fields["mode"].disabled = True
```
timoludwig, line 23: If the mode is changed to "only available" after sending, the messages are contained in the API response, but empty:
```
$ curl -s http://localhost:8000/api/testumgebung/fa/fcm/ | python -m json.tool

[
    {
        "id": "8",
        "title": "",
        "message": "",
        "timestamp": "2022-10-14T13:30:34.942Z",
        "last_updated": "2022-10-14T15:30:34.942+02:00",
        "channel": "news"
    }
]
```

So I suggest to disable these fields after a notification was sent.

integreat_cms/cms/views/push_notifications/push_notification_form_view.py
```diff
@@ -96,17 +93,17 @@ def get(self, request, *args, **kwargs):
         push_notification_form = PushNotificationForm(instance=push_notification)
 
         num_languages = len(region.active_languages)
-        PNTFormset = modelformset_factory(
-            PushNotificationTranslation,
+        PushNotificationTranslationFormSet = inlineformset_factory(
```
timoludwig, line 96: I suggest to use the `inlineformset_factory` instead of the `modelformset_factory` because the [BaseInlineFormSet](https://docs.djangoproject.com/en/3.2/topics/forms/modelforms/#django.forms.models.BaseInlineFormSet) is exactly designed for our purpose - saving multiple objects with a foreign key to the same object instance. Also, since the formset returns a class, I guess it's justified to write the variable name upper case.

integreat_cms/cms/forms/push_notifications/push_notification_translation_form.py
```diff
@@ -50,3 +51,19 @@ def add_error_messages(self, request):
...
+        :return: Whether the form has changed
+        :rtype: bool
+        """
+        if (
+            hasattr(self.instance, "push_notification")
+            and self.instance.push_notification.mode
+            == push_notifications.USE_MAIN_LANGUAGE
+        ):
+            return True
+        return super().has_changed()
```
timoludwig, line 69: Instead of iterating though all forms, setting the data manually and overwriting the save method, we can also overwrite the form's `has_changed()` method to convince the formset to save these forms as well.

integreat_cms/cms/views/push_notifications/push_notification_form_view.py
```diff
@@ -96,24 +93,26 @@ def get(self, request, *args, **kwargs):
...
+            instance=push_notification,
             # Add initial data for all languages which do not yet have a translation
             initial=[
                 {"language": language}
                 for language in region.active_languages
                 if language not in existing_languages
             ],
         )
+        # Make title of default language required
+        pnt_formset[0].fields["title"].required = True
```
timoludwig, line 115: If the title field is no longer required, we need to make sure to require at least one title - otherwise, the form can just be saved with no input at all.

timoludwig (commented):
Thanks a lot, this fixes the issue! :+1:

I found two functional problems:
- It was possible to save the form without any input
- When the mode was changed from "use main language" to "only available" after sending, the API response contained empty titles & texts

Apart from that, I found a workaround so we can remove the loop and different save method.

Since some of the changes were not in the area Github allowed me to create suggestions for, I directly pushed my suggestions to the branch. Feel free to revert them if you want :wink:

> I am not really sure about this, but I think that your changes are breaking our debug setup for FCM. Right now, I only can send PNs to our real FCM channel, but they are no longer printed to the log if our system is in DEBUG mode thinking

@ulliholtgrave Hmm, somehow I cannot reproduce your problem, when I tried it I was able to send push notifications to the testumgebung and also got the "invalid key" errors in the console...

ulliholtgrave (approved):
I also tested yours and Timo's changes and everything seems to work accordingly :+1:

Issue #1714, opened 2022-09-23
Log GET parameters of failed HTTP requests


timoludwig:
### Motivation

In our error log, only the base path of the URL that caused the error is shown, e.g.:
```
Sep 23 07:38:43 ERROR   django.request - Internal Server Error: /bodenseekreis/ar/wp-json/ig-mpdf/v1/pdf/
Sep 23 07:59:19 ERROR   django.request - Internal Server Error: /weilheim-schongau/fr/wp-json/extensions/v3/children/
Sep 23 16:48:28 WARNING django.request - Not Found: /lkkarlsruhe/ro/wp-json/extensions/v3/children/
```

### Proposed Solution

To get more insight what went wrong, I suggest to log the GET parameters as well.

PR #1715, opened 2022-09-23
Log GET parameters of failed HTTP requests

timoludwig:
### Short description

Log GET parameters of failed HTTP requests

### Proposed changes

- Add logging formatter to modify the log records of the `django.request` module

### Side effects

If we receive any kind of secrets/keys via GET-parameters, this would expose them to the log.
However, I don't think that we do this anywhere...

### Resolved issues

Fixes: #1714

### DIFF
--- a/integreat_cms/core/logging_formatter.py
+++ b/integreat_cms/core/logging_formatter.py
@@ -33,3 +33,36 @@ class ColorFormatter(logging.Formatter):
         # pylint: disable=protected-access
         self._style._fmt = color + fmt + "\x1b[0m"
         return super().format(record)
+
+
+class RequestFormatter(logging.Formatter):
+    """
+    Logging Formatter to log the GET parameters of a failed HTTP request
+    """
+
+    def format(self, record):
+        """
+        Format the specified record including the request if possible (see :meth:`python:logging.Formatter.format`).
+
+        :param record: The log record
+        :type record: ~logging.LogRecord
+
+        :return: The formatted logging message
+        :rtype: str
+        """
+        message = super().format(record)
+        # Check whether this record belongs to a request
+        if record.name == "django.request":
+            # Prepend HTTP status code to the message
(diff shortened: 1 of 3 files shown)


ulliholtgrave (approved):
Thanks, that does the trick and should help us for issues on the production :+1:

membralala:
Thank you! For me this works as the query parameters are shown as expected. 
However, I found it sometimes a little hard to find the GET parameters as they were shown after traceback other log information. E.g. I provoked a permission denied error  and got this as log: 
```
Sep 28 10:07:33 WARNING django.request - 403 Forbidden (Permission denied): /nurnberg/pages/de/
  File "/Users/philip/.local/share/virtualenvs/Integreat-4loMQVvc/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
  File "/Users/philip/Documents/Integreat/integreat-cms/integreat_cms/core/middleware/access_control_middleware.py", line 67, in __call__
    raise PermissionDenied(
django.core.exceptions.PermissionDenied: <SimpleLazyObject: <User (id: 3, username: editor, role: Editor, region: augsburg)>> does not have the permission to access <Region (id: 2, slug: nurnberg)>?test=1
Forbidden (Permission denied): /nurnberg/pages/de/
```
where the `?test=1` is not so easy to see. Not sure if this was intentional or if it is even a realistic problem. 👀 But maybe something like this would keep the information together (although I admit it the code does not look as nice as before): 
```
query = f"?{record.request.META['QUERY_STRING']} " if record.request.META['QUERY_STRING'] else ""
message = message.replace(
    "django.request - ", f"django.request - {record.status_code} {query}"
)
```

timoludwig:
> However, I found it sometimes a little hard to find the GET parameters as they were shown after traceback other log information. 

Good catch! I tried to improve the error message for multi-line exceptions now.

Issue #1720, opened 2022-09-28
SUMM.AI translation not working for `<p>&nbsp;</p>` tags


timoludwig:
### Describe the Bug

Our content contains a few "strange" new lines which consist of `<p>`-tags and one space character.
The SUMM.AI returns the error "Please check your input text. It should not be empty." for these tags.

### Steps to Reproduce

1. Create page with `<p>&nbsp;</p>` tag in content
2. Try to translate into easy German
3. See error

### Expected Behavior

The elements should just stay as they are - just a non-breaking space.

### Actual Behavior

The API returns an error

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
Sep 28 15:31:11 ERROR   integreat_cms.summ_ai_api.summ_ai_api_client - SUMM.AI translation of <HTMLSegment (text:  )> failed because of <class 'integreat_cms.summ_ai_api.summ_ai_api_client.SummAiException'>: API error: 400 - Please check your input text. It should not be empty.
  ```

</details>

PR #1721, opened 2022-09-28
Fix translation of empty tags with SUMM.AI API for Easy German

timoludwig:
### Short description

Fix translation of empty tags with SUMM.AI API for Easy German

### Proposed changes

- Filter out all empty texts and don't translate them

### Resolved issues

Fixes: #1720

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,13 @@
 UNRELEASED
 ----------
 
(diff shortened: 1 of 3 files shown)


ulliholtgrave (approved):
I didn't do an insight analysis of the behavior but for empty HTML tags this works fine :+1:

Issue #1717, opened 2022-09-26
Submitting feedback to imprint in non-existing language breaks the feedback list


timoludwig:
### Describe the Bug

When feedback is submitted about a fallback translation of the imprint, the feedback list breaks

### Steps to Reproduce

1. Submit feedback about imprint in a language where no imprint exists
2. Open feedback list

### Expected Behavior

The feedback should be shown

### Actual Behavior

An internal server error occurs:
```
AttributeError: 'NoneType' object has no attribute 'title'
```

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
_render
    return self.nodelist.render(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/loader_tags.py", line 62, in render
    result = block.nodelist.render(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/loader_tags.py", line 62, in render
    result = block.nodelist.render(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/defaulttags.py", line 214, in render
    nodelist.append(node.render_annotated(context))
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/base.py", line 988, in render
    output = self.filter_expression.resolve(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/base.py", line 671, in resolve
[...]

PR #1730, opened 2022-10-04
Handle imprint feedback given in languages, that don't exist in a region.

membralala:
### Short description

This PR should fix the feedback view error, that was shown, when imprint feedback was submitted in a language, that does not exist for this regions feedback. To be honest, I wasn't so sure, how the API is usually used, but as far as I understand, this should do it. 

### Proposed changes

- Throw 404 for imprint feedback via API, if no imprint for the selected language exists.
- If somehow still a feedback was made in a not existing language, provide a fallback for the "Feedback on" column in the feedback list view. 

### Side Effects
- Introduce a new database field "imprint page translation" to ```ImprintPageFeedback```

### Resolved issues

Fixes: #1717

### DIFF
--- a/integreat_cms/summ_ai_api/summ_ai_api_client.py
+++ b/integreat_cms/summ_ai_api/summ_ai_api_client.py
@@ -97,7 +97,7 @@ class SummAiApiClient:
                 # Let the field handle the translated text
                 text_field.translate(response_data["translated_text"])
                 return text_field
-        except (aiohttp.ClientError, SummAiException) as e:
+        except (aiohttp.ClientError, asyncio.TimeoutError, SummAiException) as e:
             logger.error(
                 "SUMM.AI translation of %r failed because of %s: %s",
                 text_field,


membralala:
I realised some issues and inconsistencies with this solution, so I will change it and follow the approach, that e.g. page feedbacks use.

MizukiTemma (commented):
Thank you for work!
I've tested it so far by creating feedback objects in Django-admin and it works for list displaying as expected 👍

Still to test is how it works when a feed back is sent in the same way as users would. Clicking an icon for feed back, (tipping comments, and) sending it by a button: how can we simulate this way?

integreat_cms/api/v3/feedback/imprint_page_feedback.py
```diff
@@ -68,13 +68,16 @@ def imprint_page_feedback_internal(
...
 
     :return: JSON object according to APIv3 imprint feedback endpoint definition
     :rtype: ~django.http.JsonResponse
     """
-    if region.imprint:
+    imprint = region.imprint
+
+    if imprint and language in region.languages:
         ImprintPageFeedback.objects.create(
+            imprint_page_translation=imprint.get_translation(language.slug),
```
timoludwig, line 80: Just because `language` is in `region.languages` does not mean that this translation exists - and if it doesn't, this means that the API does not return a 404, but throws an internal server error instead (

integreat_cms/cms/models/feedback/imprint_page_feedback.py
```diff
@@ -1,16 +1,25 @@
...
     Database model representing feedback about imprint pages.
     """
 
+    imprint_page_translation = models.ForeignKey(
+        ImprintPageTranslation,
+        on_delete=models.CASCADE,
+        related_name="feedback",
+        verbose_name=_("imprint page translation"),
+    )
+
```
timoludwig, line 22:
```suggestion
```
I don't think it's required to add this field - in most cases, the imprint is just duplicated into all languages which means only for the default language this translation does exist - for all other languages it doesn't...

timoludwig (changes requested):
Thanks a lot! :muscle: 

The part of showing entries with non-existing translations probably works as expected :+1:
(I couldn't test this, since the new database structure does not allow `null`-values in this field.) 

However, I see two major problems:
1. The current migration cannot be applied to the production system, because our prod database already contains imprint feedback entries without translation and the new database structure does not allow empty values in this field.
2. With your solution, it isn't possible to create new feedback entries for fallback translations anymore. However, the app has no idea which languages exist in our database and which of them are just fallback translations, so it shows the feedback icons for every language. Thus, we also have to accept feedback for non-existing languages.

> Still to test is how it works when a feed back is sent in the same way as users would. Clicking an icon for feed back, (tipping comments, and) sending it by a button: how can we simulate this way?

@MizukiTemma you can simulate the behaviour of the app by executing the following:

```
curl -X "POST" "http://localhost:8000/api/augsburg/de/feedback/imprint-page/" -d 'comment=Illegales%20Impressum!&rating=down'
```
(Change the language slug `de` to e.g. `ar` if you want to submit feedback for another language or change the endpoint `imprint-page` if you want to submit feedback for other content types.)

membralala:
@MizukiTemma , @timoludwig  thank you for your feedback. 

> 1. The current migration cannot be applied to the production system, because our prod database already contains imprint feedback entries without translation and the new database structure does not allow empty values in this field.

Thank you for the advice, I did not know that this won't work. But then we need to make sure, that imprint feedbacks are deleted when the imprint page is deleted via the imprint page model, don't we? Otherwise there could be feedbacks for a not existing page. 

> 2. With your solution, it isn't possible to create new feedback entries for fallback translations anymore. However, the app has no idea which languages exist in our database and which of them are just fallback translations, so it shows the feedback icons for every language. Thus, we also have to accept feedback for non-existing languages.

Okay I see. So just to clarify this for me, this point refers to the restrictions I introduced to the API endpoint, right?

timoludwig:
>  But then we need to make sure, that imprint feedbacks are deleted when the imprint page is deleted via the imprint page model, don't we? Otherwise there could be feedbacks for a not existing page.

Yes, correct, this is already handled:

https://github.com/digitalfabrik/integreat-cms/blob/d68c72f9274ec0d7691dc914250194ac062691a1/integreat_cms/cms/views/imprint/imprint_actions.py#L48

> Okay I see. So just to clarify this for me, this point refers to the restrictions I introduced to the API endpoint, right?

Not only the endpoint, also the changed model has an impact on this: When we store the imprint translation as foreign key field, we cannot create new feedback entries which don't have a corresponding translation.

PR #1731, opened 2022-10-05
Ensure .postgres directory exists when creating container

PeterNerlich:
### Short description
When starting the development server for the first time using docker as the platform for Postgres, `./dev-tools/run.sh` fails to create the container because it tries to bind the yet non-existent directory `.postgres` to it.

### Proposed changes
Ensure the directory is created if necessary when creating the container.

Opting for `mkdir -p` because
- it won't throw an error if the directory already does exist
- it will create the directory if it does not exist
- it *will* throw an error if it exists and is anything *but* a directory.

### Side effects
None come to mind. If anything, it should be limited to fresh starts only, as the change is only in the `create_docker_container` helper function.


ulliholtgrave (approved):
Thanks, I guess this won't harm and may prevent some problems :+1:

timoludwig (approved):
Congratz to your first PR and welcome to the team :tada:

Yes, I agree, this looks good! :rocket:

Issue #1630, opened 2022-08-11
Push Notification sent but status not updated due to missing firebase response ID


ulliholtgrave:
### Describe the Bug

One customer had a bug about sending a push notification. It was received on the smartphones, but it had a "not sent yet" status. In our log I found error which indicates that no id was created (see Traceback).

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/views/push_notifications/push_notification_form_view.py", line 255, in post
    if push_sender.send_all():
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/views/push_notifications/push_notification_sender.py", line 140, in send_all
    logger.info("%r sent, FCM id: %r", pnt, res.json()["message_id"])
KeyError: 'message_id'

  ```

</details>

PR #1740, opened 2022-10-06
Fix Spontaneous Multiplication of Push Notifications

PeterNerlich:
### Short description
Sometimes, the Firebase API seems to not provide a `message_id` for submitted push notifications, while it is still received on subscribed devices. As it still sent HTTP status code 200, we assumed `message_id` was there and tried logging it, causing an exception. This lead to resend attempts and thus multiplied push notifications.

### Proposed changes
- Log API full response if no `message_id` is received on submitting a push notification, to gather further information
- Deliberately don't raise an exception since PNs seem to be sent correctly, (but avoid previous exception from failing to log non-existent dict field)

### Side effects
- Scenarios where a push notification might fail with an error in the API response, but HTTP code 200, will not be retried. Instances of this will only be visible from the logs.

### Resolved issues
Fixes: #1630

### DIFF
--- a/integreat_cms/cms/views/push_notifications/push_notification_sender.py
+++ b/integreat_cms/cms/views/push_notifications/push_notification_sender.py
@@ -142,7 +142,12 @@ class PushNotificationSender:
         for pnt in self.prepared_pnts:
             res = self.send_pn(pnt)
             if res.status_code == 200:
-                logger.info("%r sent, FCM id: %r", pnt, res.json()["message_id"])
+                if "message_id" in res.json():
+                    logger.info("%r sent, FCM id: %r", pnt, res.json()["message_id"])
+                else:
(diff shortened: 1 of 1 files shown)


ulliholtgrave (approved):
Looks good. I wasn't able to reproduce this issue locally, but your explanation makes sense and I would be up for bringing this on production :+1:

timoludwig (approved):
Looks good! :+1:
Thanks a lot for investigating the problem and for the detailed explanation :pray:

PeterNerlich:
Further detail that might be helpful: If I didn't get it wrong while quickly googling this, [this section of the API doc](https://firebase.google.com/docs/cloud-messaging/http-server-ref#table9) outlines some error conditions while still sending HTTP status code 200.

PR #1744, opened 2022-10-10
Auto-complete address and coordinates of locations

MizukiTemma:
### Short description

Currently coordinates of a POI appears (if left blank) first after the POI is saved.
This PR auto fills coordinates live while editing when the address of POI is completed.

### Proposed changes

- Add event listeners in the address fields.
- Coordinates will be searched live while an user is filling the address fields.

### Side effects

- So far not known

### Resolved issues

Part of: #1000

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ UNRELEASED
 * [ [#1742](https://github.com/digitalfabrik/integreat-cms/issues/1742) ] Add last modified date to media sidebar
 * [ [#1703](https://github.com/digitalfabrik/integreat-cms/issues/1703) ] Remove pending account activation warning when user form is submitted with errors
 * [ [#1684](https://github.com/digitalfabrik/integreat-cms/issues/1684) ] Set filesize limit for uploads to 3MB
+* [ [#1000](https://github.com/digitalfabrik/integreat-cms/issues/1000) ] Auto-complete address and coordinates of locations
 
 
 2022.10.0
--- a/integreat_cms/cms/templates/pois/poi_form.html
+++ b/integreat_cms/cms/templates/pois/poi_form.html
@@ -140,6 +140,7 @@
                     <div class="flex flex-col flex-auto px-4 pb-4 divide-y divide-gray-200 space-y-2">
                         <div>
                             <label>{% trans 'Address' %}</label>
+                            <div id="nominatim-error" class="bg-red-100 border-l-4 border-red-500 text-red-500 px-4 py-3 my-2 hidden" role="alert"></div>
                             <label for="{{ poi_form.address.id_for_label }}" class="secondary">{{ poi_form.address.label }}</label>
                             {% render_field poi_form.address|add_error_class:"border-red-500" %}
                             <label for="{{ poi_form.postcode.id_for_label }}" class="secondary">{{ poi_form.postcode.label }}</label>
@@ -156,9 +157,13 @@
                         </div>
                         <div>
                             <label>{% trans 'Position' %}</label>
-                            <div class="help-text">
-                                {% trans 'If you leave these fields blank, they are automatically derived from the address.' %}
+                            <div>
+                                <input type="checkbox" id="auto-fill-coordinates" data-url="{% url 'auto_complete_poi_address' %}" checked>
+                                <label for="auto-fill-coordinates" class="secondary !inline">
+                                    {% trans 'Derive coordinates automatically from address' %}
+                                </label>
                             </div>
+
                             <label for="{{ poi_form.latitude.id_for_label }}" class="secondary">{{ poi_form.latitude.label }}</label>
                             {% render_field poi_form.latitude|add_error_class:"border-red-500" %}
                             <div class="help-text">{{ poi_form.latitude.help_text }}</div>
--- a/integreat_cms/cms/views/pois/poi_actions.py
+++ b/integreat_cms/cms/views/pois/poi_actions.py
@@ -1,18 +1,23 @@
 """
 This module contains view actions for objects related to POIs.
 """
+import json
 import logging
 
+from django.conf import settings
 from django.contrib import messages
-from django.http import Http404
+from django.http import Http404, HttpResponse, JsonResponse
 from django.shortcuts import render, redirect
 from django.utils.translation import ugettext as _
 from django.views.decorators.http import require_POST
 
 from linkcheck.models import Link
 
+from ....api.decorators import json_response
 from ...decorators import permission_required
 from ...models import POI
+
+from ....nominatim_api.nominatim_api_client import NominatimApiClient
 
 logger = logging.getLogger(__name__)
 
@@ -170,3 +175,51 @@ def view_poi(request, poi_id, region_slug, language_slug):
         raise Http404
 
     return render(request, template_name, {"poi_translation": poi_translation})
+
+
+@json_response
+@require_POST
+@permission_required("cms.view_poi")
+def auto_complete_address(request):
+    """
+    Autocomplete location address and coordinates
+
+    :param request: The current request
+    :type request: ~django.http.HttpRequest
+
+    :raises ~django.http.Http404: If no location was found for the given address
+
+    :return: The address and coordinates of the location
+    :rtype: ~django.http.JsonResponse
+    """
+    data = json.loads(request.body.decode("utf-8"))
+
+    if not settings.NOMINATIM_API_ENABLED:
+        return HttpResponse(_("Location service is disabled"), status_code=503)
+
+    street_input = data.get("street")
+    postcode_input = data.get("postcode")
+    city_input = data.get("city")
+
+    nominatim_api_client = NominatimApiClient()
+
+    result = nominatim_api_client.search(
+        street=street_input,
+        postalcode=postcode_input,
+        city=city_input,
+        addressdetails=True,
+    )
+
+    if not result:
+        raise Http404(_("Address could not be found"))
+
+    address = result.raw.get("address", {})
+    return JsonResponse(
+        data={
+            "postcode": address.get("postcode"),
+            "city": address.get("city"),
+            "country": address.get("country"),
+            "longitude": result.longitude,
+            "latitude": result.latitude,
+        }
+    )
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-10-15 16:25+0000\n"
+"POT-Creation-Date: 2022-10-15 23:05+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: Integreat <info@integreat-app.de>\n"
 "Language-Team: Integreat <info@integreat-app.de>\n"
@@ -4182,7 +4182,7 @@ msgstr "Auf Google Maps öffnen"
 #: cms/templates/imprint/imprint_form.html:120
 #: cms/templates/pages/page_form.html:285 cms/templates/pages/page_sbs.html:87
 #: cms/templates/poicategories/poicategory_list.html:31
-#: cms/templates/pois/poi_form.html:233
+#: cms/templates/pois/poi_form.html:238
 #: cms/templates/regions/region_list.html:25
 #: cms/templates/settings/user_settings.html:93
 msgid "Actions"
@@ -5637,58 +5637,54 @@ msgstr "Neuen Ort erstellen"
 msgid "The name is currently displayed to users only in the default language"
 msgstr "Der Name wird Benutzern momentan nur in der Standardsprache angezeigt"
 
-#: cms/templates/pois/poi_form.html:137 cms/templates/pois/poi_form.html:158
+#: cms/templates/pois/poi_form.html:137 cms/templates/pois/poi_form.html:159
 msgid "Position"
 msgstr "Position"
 
-#: cms/templates/pois/poi_form.html:160
-msgid ""
-"If you leave these fields blank, they are automatically derived from the "
-"address."
-msgstr ""
-"Wenn Sie diese Felder leer lassen, werden sie automatisch aus der Adresse "
-"abgeleitet."
+#: cms/templates/pois/poi_form.html:163
+msgid "Derive coordinates automatically from address"
+msgstr "Koordinaten automatisch aus der Adresse ableiten"
 
-#: cms/templates/pois/poi_form.html:177
+#: cms/templates/pois/poi_form.html:182
 msgid "Contact details"
 msgstr "Kontaktdaten"
 
-#: cms/templates/pois/poi_form.html:240 cms/templates/pois/poi_form.html:242
+#: cms/templates/pois/poi_form.html:245 cms/templates/pois/poi_form.html:247
 #: cms/templates/pois/poi_list_archived_row.html:65
 msgid "Restore location"
 msgstr "Ort wiederherstellen"
 
-#: cms/templates/pois/poi_form.html:248
+#: cms/templates/pois/poi_form.html:253
 msgid "Restore this location"
 msgstr "Diesen Ort wiederherstellen"
 
-#: cms/templates/pois/poi_form.html:252 cms/templates/pois/poi_form.html:254
+#: cms/templates/pois/poi_form.html:257 cms/templates/pois/poi_form.html:259
 #: cms/templates/pois/poi_list_row.html:112
 msgid "Archive location"
 msgstr "Ort archivieren"
 
-#: cms/templates/pois/poi_form.html:260
+#: cms/templates/pois/poi_form.html:265
 msgid "Archive this location"
 msgstr "Diesen Ort archivieren"
 
-#: cms/templates/pois/poi_form.html:267 cms/templates/pois/poi_form.html:289
+#: cms/templates/pois/poi_form.html:272 cms/templates/pois/poi_form.html:294
 #: cms/templates/pois/poi_list_archived_row.html:78
 #: cms/templates/pois/poi_list_row.html:125
 msgid "Delete location"
 msgstr "Ort löschen"
 
-#: cms/templates/pois/poi_form.html:273
+#: cms/templates/pois/poi_form.html:278
 msgid "You cannot delete a location which is referenced by an event."
 msgstr ""
 "Der Ort kann nicht gelöscht werden, da er von einem Event verwendet wird."
 
-#: cms/templates/pois/poi_form.html:275
+#: cms/templates/pois/poi_form.html:280
 msgid "To delete this location, you have to delete this event first:"
 msgid_plural "To delete this location, you have to delete these events first:"
 msgstr[0] "Löschen Sie zuerst diese Veranstaltung, um den Ort zu löschen:"
 msgstr[1] "Löschen Sie zuerst diese Veranstaltungen, um den Ort zu löschen:"
 
-#: cms/templates/pois/poi_form.html:295
+#: cms/templates/pois/poi_form.html:300
 msgid "Delete this location"
 msgstr "Diesen Ort löschen"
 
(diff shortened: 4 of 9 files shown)


sarahsporck (approved):
Very nice work!
Tested and seems to work correctly 👌🏼 

I think you can adjust the help text of the longitude/latitude field in the poi_form, as they are now inferred live :) Maybe something like `Geographische Länge und Breite werden automatisch aus der Adresse abgeleitet`. 

Further I'm a little bit worried about people who want to set longitude and latitude to a different value as the address. In this case changing the Address would mean to overwrite the longitude and latitude value. Maybe the users should be able to opt-out from inferring by adding a tick box `Länge und Breite aus der Addresse ableiten`?
This would also make the help text for longitude and latitude irrelevant, I guess.

MizukiTemma:
@sarahsporck 
Thnak you for your review :)

Now the tick box is added so users can choose whether the coordinates are to be auto-filled.
When they once checked the box but uncheck it afterwords, the original value of coordinates will be restored.

integreat_cms/cms/templates/pois/poi_form.html
```diff
@@ -156,8 +156,9 @@ <h3 class="heading font-bold text-black">
                         </div>
                         <div>
                             <label>{% trans 'Position' %}</label>
-                            <div class="help-text">
-                                {% trans 'If you leave these fields blank, they are automatically derived from the address.' %}
+                            <div>
+                                <input type="checkbox" id="auto-fill-coordinates">
+                                <label>{% trans 'Calcurate coordinates automatically from address' %}</label>
```
david-venhoff, line 161:
```suggestion
                                <label for="auto-fill-coordinates">{% trans 'Calculate coordinates automatically from address' %}</label>
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -5637,54 +5637,50 @@ msgstr "Der Name wird Benutzern momentan nur in der Standardsprache angezeigt"
...
-#: cms/templates/pois/poi_form.html:160
-msgid ""
-"If you leave these fields blank, they are automatically derived from the "
-"address."
-msgstr ""
-"Wenn Sie diese Felder leer lassen, werden sie automatisch aus der Adresse "
-"abgeleitet."
+#: cms/templates/pois/poi_form.html:161
+msgid "Calcurate coordinates automatically from address"
+msgstr "Länge und breite aus der Adresse ableiten"
```
david-venhoff, line 5642:
```suggestion
msgid "Calculate coordinates automatically from address"
msgstr "Länge und Breite aus der Adresse ableiten"
```

integreat_cms/cms/templates/pois/poi_form.html
timoludwig:
```suggestion
                                <label for="auto-fill-coordinates" class="secondary !inline">
                                    {% trans 'Derive coordinates automatically from address' %}
                                </label>
```
In addition to @david-venhoff's comment I would also suggest to make the label smaller and define it as inline element so it aligns next to the checkbox instead of breaking into the next line.

integreat_cms/static/src/js/pois/poi-actions.ts
```diff
@@ -0,0 +1,68 @@
...
+
+window.addEventListener("load", () => {
+    document.getElementById("id_address")?.addEventListener("keyup", fetchCoordinates);
+    document.getElementById("id_postcode")?.addEventListener("keyup", fetchCoordinates);
+    document.getElementById("id_city")?.addEventListener("keyup", fetchCoordinates);
+
+    document.getElementById("auto-fill-coordinates").addEventListener("input", fetchCoordinates);
+
+    initialLongitude = ( document.getElementById("id_longitude") as HTMLInputElement).value;
+    initialLatitude = (document.getElementById("id_latitude") as HTMLInputElement).value;
```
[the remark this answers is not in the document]
MizukiTemma: @david-venhoff Thank you for your correction 🙈 so much of typos...

When it's a new poi, blank value will be restored if the box is not checked or unchecked. If the POI is saved with blank coordinates, those will be automatically derived as before.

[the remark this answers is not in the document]
david-venhoff: There is a small edge case that if the user creates a new poi, fills in the coordinates manually first and then enters the poi address, then the coordinates would get erased. But since this is probably not going to happen I think the current behavior is fine :)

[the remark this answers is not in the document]
MizukiTemma: Ah ok, I got it!
I'll fix it too :) Thank you for detecting it.

integreat_cms/static/src/js/pois/poi-actions.ts
```diff
@@ -0,0 +1,85 @@
+import { getCsrfToken } from "../utils/csrf-token";
+
+let initialLongitude: string, initialLatitude: string;
+let userInputLng: string, userInputLat: string;
+
+window.addEventListener("load", () => {
+    document.getElementById("id_address")?.addEventListener("keyup", fetchCoordinates);
+    document.getElementById("id_postcode")?.addEventListener("keyup", fetchCoordinates);
+    document.getElementById("id_city")?.addEventListener("keyup", fetchCoordinates);
```
timoludwig, line 9: If we perform an ajax call for each keyup event, we will spam the nominatim server with very many request.
This could cause race conditions and puts unnecessary strain on the service, so I'd suggest to either include a timeout here similar how we do it for [search suggestions](https://github.com/digitalfabrik/integreat-cms/blob/33a4ccf6fb2734fd2f89f077a045136cae88b2d6/integreat_cms/static/src/js/search-query.ts#L72-L89) or use another event, e.g. `focusout`.
```suggestion
    document.getElementById("id_address")?.addEventListener("focusout", fetchCoordinates);
    document.getElementById("id_postcode")?.addEventListener("focusout", fetchCoordinates);
    document.getElementById("id_city")?.addEventListener("focusout", fetchCoordinates);
```

integreat_cms/cms/templates/pois/poi_form.html
```diff
@@ -156,8 +156,9 @@ <h3 class="heading font-bold text-black">
                         </div>
                         <div>
                             <label>{% trans 'Position' %}</label>
-                            <div class="help-text">
-                                {% trans 'If you leave these fields blank, they are automatically derived from the address.' %}
+                            <div>
+                                <input type="checkbox" id="auto-fill-coordinates">
+                                <label for="auto-fill-coordinates">{% trans 'Calculate coordinates automatically from address' %}</label>
                             </div>
```
timoludwig, line 162:
```suggestion
                            </div>
                            <div id="nominatim-error" class="bg-red-100 border-l-4 border-red-500 text-red-500 px-4 py-3 mb-5 hidden" role="alert"></div>

```
I'd also show an error if anything went wrong with fetching.

integreat_cms/static/src/js/pois/poi-actions.ts
```diff
@@ -0,0 +1,85 @@
...
+    document.getElementById("id_longitude").addEventListener("keyup", saveActualCoordinates);
+    document.getElementById("id_latitude").addEventListener("keyup", saveActualCoordinates);
+});
+
+function saveActualCoordinates(){
+    userInputLng = (document.getElementById("id_longitude") as HTMLInputElement).value;
+    userInputLat = (document.getElementById("id_latitude") as HTMLInputElement).value;
+}
+
+async function fetchCoordinates() {
```
timoludwig, line 25:
```suggestion
async function fetchCoordinates() {
    let error = document.getElementById("nominatim-error");
    error.classList.add("hidden");

    let longitude = (document.getElementById("id_longitude") as HTMLInputElement);
    let latitude = (document.getElementById("id_latitude") as HTMLInputElement);
```

integreat_cms/static/src/js/pois/poi-actions.ts
```diff
@@ -0,0 +1,85 @@
...
+                postcode: postcode,
+                city: city,
+            }),
+        });
+
+        if (response.status != 200) {
+            return;
+        }
+
+        const data = await response.json();
```
timoludwig, line 53:
```suggestion
        const data = await response.json();

        if (response.status != 200) {
            error.textContent = data.error;
            error.classList.remove("hidden");
            return;
        }
```

integreat_cms/static/src/js/pois/poi-actions.ts
```diff
@@ -0,0 +1,85 @@
...
+
+        if (data) {
+            // Set and display new data
+            let longitude = data.longitude;
+            let latitude = data.latitude;
+
+            ( document.getElementById("id_longitude") as HTMLInputElement).value = longitude;
+            ( document.getElementById("id_latitude") as HTMLInputElement).value = latitude; 
+        }
```
timoludwig, line 62:
```suggestion
        // Set and display new data
        longitude.value = data.longitude;
        latitude.value = data.latitude;
        longitude.classList.add("!border-green-500");
        latitude.classList.add("!border-green-500");
        setTimeout(function() {
            longitude.classList.remove("!border-green-500");
            latitude.classList.remove("!border-green-500");
        }, 5000);
```
We could also show a green border after fetching to indicate the side effect

integreat_cms/static/src/js/pois/poi-actions.ts
```diff
@@ -0,0 +1,85 @@
...
+
+        if (initialLatitude){
+            nextLatitude = initialLatitude;
+        }
+        else{
+            nextLatitude = userInputLat;
+        }
+
+        ( document.getElementById("id_longitude") as HTMLInputElement).value = nextLongitude;
+        ( document.getElementById("id_latitude") as HTMLInputElement).value = nextLatitude;
```
timoludwig, line 83:
```suggestion
    } else {
        if (initialLongitude) {
            longitude.value = initialLongitude;
        } else {
            longitude.value = userInputLng;
        }
        if (initialLatitude) {
            latitude.value = initialLatitude;
        } else {
            latitude.value = userInputLat;
        }
```

integreat_cms/cms/views/pois/poi_actions.py
```diff
@@ -170,3 +174,36 @@ def view_poi(request, poi_id, region_slug, language_slug):
         raise Http404
 
     return render(request, template_name, {"poi_translation": poi_translation})
+
+
```
timoludwig, line 178:
```suggestion

@json_response
```
Since it's an AJAX view, I suggest to always return a JSON, even on errors

integreat_cms/cms/views/pois/poi_actions.py
```diff
@@ -1,10 +1,12 @@
 """
 This module contains view actions for objects related to POIs.
 """
+import json
 import logging
 
+from django.conf import settings
 from django.contrib import messages
-from django.http import Http404
+from django.http import Http404, JsonResponse
```
timoludwig, line 9:
```suggestion
from django.http import Http404, HttpResponse, JsonResponse
```

integreat_cms/cms/views/pois/poi_actions.py
```diff
@@ -14,6 +16,8 @@
 from ...decorators import permission_required
```
timoludwig, line 16:
```suggestion
from ....api.decorators import json_response
from ...decorators import permission_required
```

integreat_cms/cms/templates/pois/poi_form.html
```diff
@@ -156,8 +156,9 @@ <h3 class="heading font-bold text-black">
                         </div>
                         <div>
                             <label>{% trans 'Position' %}</label>
-                            <div class="help-text">
-                                {% trans 'If you leave these fields blank, they are automatically derived from the address.' %}
+                            <div>
+                                <input type="checkbox" id="auto-fill-coordinates">
+                                <label for="auto-fill-coordinates">{% trans 'Calculate coordinates automatically from address' %}</label>
```
timoludwig, line 161:
```suggestion
                                <label for="auto-fill-coordinates" class="secondary !inline">
                                    {% trans 'Calculate coordinates automatically from address' %}
                                </label>
```
I would also suggest to make the label smaller and define it as inline element so it aligns next to the checkbox instead of breaking into the next line.

MizukiTemma:
@timoludwig 
Cool improvement 😄 Thank you!

Sorry that I have misunderstoot the issue. I'll now add other (not yet checked) features 🙂

integreat_cms/cms/views/pois/poi_actions.py
```diff
@@ -170,3 +175,39 @@ def view_poi(request, poi_id, region_slug, language_slug):
...
+@require_POST
+@permission_required("cms.view_poi")
+def deliver_coordinates(request):
+    """
+    Live search for POI Coordinates
+
+    :param request: The current request
+    :type request: ~django.http.HttpRequest
+
+    :raises ~django.http.Http404: If Nominatim API is not enabled
```
timoludwig, line 190:
```suggestion
    :raises ~django.http.Http404: If no coordinates were found for the given address
```

integreat_cms/cms/templates/pois/poi_form.html
```diff
@@ -156,9 +156,14 @@ <h3 class="heading font-bold text-black">
                         </div>
                         <div>
                             <label>{% trans 'Position' %}</label>
-                            <div class="help-text">
-                                {% trans 'If you leave these fields blank, they are automatically derived from the address.' %}
+                            <div>
+                                <input type="checkbox" id="auto-fill-coordinates" checked>
```
timoludwig, line 160:
```suggestion
                                <input type="checkbox" id="auto-fill-coordinates" data-url="{% url 'get_coordinates' %}" checked>
```
Ah, one additional thing: I would not hardcode the url, but get it via the template tag instead.

integreat_cms/static/src/js/pois/poi-actions.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+async function fetchCoordinates() {
+    let error = document.getElementById("nominatim-error");
+    error.classList.add("hidden");
+
+    let longitude = (document.getElementById("id_longitude") as HTMLInputElement);
+    let latitude = (document.getElementById("id_latitude") as HTMLInputElement);
+    if ((document.getElementById("auto-fill-coordinates") as HTMLInputElement).checked) {
+        let url = "/ajax/coordinates/";
```
timoludwig, line 32:
```suggestion
    let longitude = document.getElementById("id_longitude") as HTMLInputElement;
    let latitude = document.getElementById("id_latitude") as HTMLInputElement;
    let autoFillCoordinates = document.getElementById("auto-fill-coordinates") as HTMLInputElement;

    if (autoFillCoordinates.checked) {
```

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ UNRELEASED
 * [ [#951](https://github.com/digitalfabrik/integreat-cms/issues/951) ] Add possibility to create categories for POIs
 * [ [#1742](https://github.com/digitalfabrik/integreat-cms/issues/1742) ] Add last modified date to media sidebar
 * [ [#1703](https://github.com/digitalfabrik/integreat-cms/issues/1703) ] Remove pending account activation warning when user form is submitted with errors
+* [ [#1000](https://github.com/digitalfabrik/integreat-cms/issues/1000) ] Auto-fill Coordinate after address is complete
```
```suggestion
* [ [#1000](https://github.com/digitalfabrik/integreat-cms/issues/1000) ] Auto-fill coordinates after address is complete
```

timoludwig (changes requested):
Thanks! A few small details, then this PR is ready to be merged! :tada:

> I'll now add other (not yet checked) features :slightly_smiling_face:

Well, I mean those are just ideas, if you notice that some of them are too much work for the benefit they provide, you can also edit the issue and explain why you reject a specific point :grin:

MizukiTemma:
Now the ajax call for the coordination will be called if street and postcode are filled (city is not anymore an necessary firing condition).

When coordinates are found from the street and postcode, this proves (I think) the validity of street input. Users can know something is not correctly given, seeing an error from this calculation.

From those coordinates the city will be calculated.
This is needed because data calculated by the combination of street and postcode has the address as one string without attribute like "city" or "comunity (Landkreis)" and there is no guaranty on which positioin the city name stands: whether the address is in a Landkreis or kreisfreie Stadt makes a difference on the position of city name.

The calculated city will be auto-filled (or replace the old input if the new address is in another city). The country field will be now auto-filled too.

MizukiTemma:
> Thanks! A few small details, then this PR is ready to be merged! 🎉
> 
> > I'll now add other (not yet checked) features 🙂
> 
> Well, I mean those are just ideas, if you notice that some of them are too much work for the benefit they provide, you can also edit the issue and explain why you reject a specific point 😁

Oh oh... I read this comment just now...... 😅

integreat_cms/cms/templates/pois/poi_form.html
```diff
@@ -140,6 +140,7 @@ <h3 class="heading font-bold text-black">
                     <div class="flex flex-col flex-auto px-4 pb-4 divide-y divide-gray-200 space-y-2">
                         <div>
                             <label>{% trans 'Address' %}</label>
+                            <div id="nominatim-error" class="bg-red-100 border-l-4 border-red-500 text-red-500 px-4 py-3 my-2 hidden" role="alert"></div>
```
timoludwig, line 143: Now that also the address is completed, I moved the error div up to the address field

integreat_cms/cms/views/pois/poi_actions.py
```diff
@@ -170,3 +175,51 @@ def view_poi(request, poi_id, region_slug, language_slug):
...
+    city_input = data.get("city")
+
+    nominatim_api_client = NominatimApiClient()
+
+    result = nominatim_api_client.search(
+        street=street_input,
+        postalcode=postcode_input,
+        city=city_input,
+        addressdetails=True,
+    )
```
timoludwig, line 211: > From those coordinates the city will be calculated. This is needed because data which geocode calclulates with the combination of street and postcode has the address as one string without attribute like "city" or "comunity (Landkreis)" and there is no guaranty on which positioin the city name stands: whether the address is in a Landkreis or kreisfreie Stadt makes a difference on the position of city name.

Instead of performing two searches - first for the coordinates and then the reverse search for the address - we can also combine them by setting `addressdetails=True` (see [docs](https://geopy.readthedocs.io/en/stable/#nominatim)).
Then the address is given as dictionary and not as string.
Also, instead of importing Nominatim directly in this view, I think it's better to adapt our Nominatim api client to our needs and then call the changed methods of the client.

integreat_cms/static/src/js/pois/poi-actions.ts
```diff
@@ -0,0 +1,93 @@
...
+    } else {
+      // Reset to initial value
+      longitude.value = longitude.dataset.initial;
+      latitude.value = latitude.dataset.initial;
+    }
+  });
+
+  // Add possibility to reset coordinates to initial values
+  if (longitude?.value) {
+    longitude.dataset.initial = longitude.value;
```
timoludwig, line 24: I'd store the initial values in the data attributes rather than in global variables.

integreat_cms/static/src/js/pois/poi-actions.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+});
+
+async function autoCompleteAddress() {
+  let street = (document.getElementById("id_address") as HTMLInputElement).value;
+  let postcode = (document.getElementById("id_postcode") as HTMLInputElement).value;
+  let city = (document.getElementById("id_city") as HTMLInputElement).value;
+
+  // Only try auto filling if street and either postcode or city are given
+  if (street.trim().length == 0 || (postcode.trim().length == 0 && city.trim().length == 0)) {
```
timoludwig, line 46: I think we can also perform the search if only address and city are given - in this case, we can try to automatically fill the postcode.

integreat_cms/static/src/js/pois/poi-actions.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  if (response.status != 200) {
+    error.textContent = data.error;
+    error.classList.remove("hidden");
+    return;
+  }
+
+  updateField("postcode", data.postcode);
+  updateField("city", data.city);
+  updateField("country", data.country);
```
timoludwig, line 76: I moved the border animation into a separate function to reduce code duplication.

integreat_cms/static/src/js/pois/poi-actions.ts
```diff
@@ -0,0 +1,95 @@
...
+  let longitude = document.getElementById("id_longitude") as HTMLInputElement;
+  let latitude = document.getElementById("id_latitude") as HTMLInputElement;
+
+  document.getElementById("auto-fill-coordinates")?.addEventListener("input", ({ target }) => {
+    if ((target as HTMLInputElement).checked) {
+      autoCompleteAddress();
+    } else {
+      // Reset to initial value
+      longitude.value = longitude.dataset.initial;
+      latitude.value = latitude.dataset.initial;
```
timoludwig, line 17: I think when the checkbox is un-checked, we don't need to perform the ajax call, we can just reset the values.

timoludwig (approved):
Thanks a lot, I really think this makes creating POIs much easier! :rocket: 
I had a few ideas for improvement which were outside the area GitHub allowed me to post suggestions for, so I directly pushed to your branch :innocent: 
I tried to explain my changes in the comments below:

integreat_cms/cms/views/pois/poi_actions.py
MizukiTemma: Oh, I overlooked it 😅 it is what I wanted to have!

MizukiTemma:
@timoludwig 
Cool improvement! Thank you for your aid 😃

PR #1748, opened 2022-10-11
Only check availability of external APIs once

timoludwig:
### Short description

I was bothered by the verbosity of each server reload after a change to a Python file, so I removed the API availability checks in those cases. Especially when working on the road, this is a bit annoying.
I think it is enough to check the availability once after the first start of the server.

### Proposed changes

- Disable the re-check on every autoreload of the dev server
- Decrease verbosity of availability check exceptions

### DIFF
--- a/integreat_cms/cms/apps.py
+++ b/integreat_cms/cms/apps.py
@@ -1,4 +1,6 @@
 import logging
+import os
+import sys
 
 from django.apps import AppConfig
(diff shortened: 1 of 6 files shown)


ulliholtgrave (approved):
Looks good :+1:

svenseeberg (approved):
LGTM

Issue #1710, opened 2022-09-22
Last tab of sidebar hidden on small screens


membralala:
### Describe the Bug

At the moment on small screens the last tab of the sidebar is hidden behind the version display. For the regions e.g. this affects the tab "language tree". 

### Steps to Reproduce

1. Go to any region.
2. Lower the height of your screen until the sidebar is scrollable.
3. Try to reach the last tab of the sidebar. 

### Expected Behavior

The sidebar's scroll behaviour should respect the version display. 

### Actual Behavior

The version display overlays the sidebar.

### Additional Information

<img width="422" alt="Bildschirmfoto 2022-09-22 um 10 16 44" src="https://user-images.githubusercontent.com/66683622/191694973-1075685d-da68-4fa2-8250-f61f483175ef.png">

PR #1754, opened 2022-10-12
Add spacing to sidebar to improve view on small screens

JoeyStk:
### Short description
I added spacing to the sidebar to improve the view on small screens. At the moment the last menu point ("Language Tree")  disappears behind the version number on small screens.

### Proposed changes

- I added 3rem padding-bottom to the sidebar hard coded in the scss file
- Initially I wanted to use the tailwind class "pb-12" since this equals to 3rem, but for me it didn't work. Neither Chrome nor Firefox registered this class. However p-12 worked, which is strange since if p-12 is working, also pb-12 should be working. [See more about this here](https://tailwindcss.com/docs/customizing-spacing#default-spacing-scale).
- Optionally I could also add a custom class for "pb-12" equaling to 3rem to the tailwind config file, but I'm not sure if we're going to use that class ever again

### Side effects

- Couldn't find any in Chrome or Firefox

### Resolved issues

Fixes: #1710 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ UNRELEASED
 * [ [#1000](https://github.com/digitalfabrik/integreat-cms/issues/1000) ] Auto-complete address and coordinates of locations
 * [ [#1434](https://github.com/digitalfabrik/integreat-cms/issues/1434) ] Add display of HIX values to nudge users to write easier texts
 * [ [#1770](https://github.com/digitalfabrik/integreat-cms/issues/1770) ] Fix error in SUMM.AI translation if paragraph contains only special characters
+* [ [#1710](https://github.com/digitalfabrik/integreat-cms/issues/1710) ] Add spacing to sidebar to improve view on small screens
 
 
 2022.10.0
--- a/integreat_cms/static/src/css/style.scss
+++ b/integreat_cms/static/src/css/style.scss
@@ -52,7 +52,7 @@ header {
 
 nav#primary-navigation {
   #menu {
-    max-height: calc(100vh - 50px);
+    max-height: calc(100vh - 56px);
     a {
       @apply relative block p-3 pl-6;
     }
--- a/integreat_cms/cms/templates/_base.html
+++ b/integreat_cms/cms/templates/_base.html
@@ -96,7 +96,7 @@
                 <div class="h-full w-full bg-{{ BRANDING }}-logo-white hover:bg-{{ BRANDING }}-logo-hover bg-contain bg-center bg-no-repeat"></div>
             </a>
         </div>
-        <div id="menu" class="pb-2 overflow-y-auto {{ BRANDING }}-branding">
+        <div id="menu" class="pb-10 overflow-y-auto {{ BRANDING }}-branding">
             {% if request.region %}
                 <a href="{% url 'dashboard' region_slug=request.region.slug %}"
                    class="{% if current_menu_item == 'region_dashboard' %} active{% endif %}">


integreat_cms/static/src/css/style.scss
```diff
@@ -53,6 +53,7 @@ header {
 nav#primary-navigation {
   #menu {
     max-height: calc(100vh - 50px);
+    padding-bottom: 3rem;
```
svenseeberg, line 56: AFAICT `2.5 rem` should be sufficient? But it works as is. It could be a little more space efficient though :D

svenseeberg (changes requested):
Looks good, thanks. 2 minor comments though :)

CHANGELOG.md
```diff
@@ -4,14 +4,14 @@ UNRELEASED
...
 * [ [#1703](https://github.com/digitalfabrik/integreat-cms/issues/1703) ] Remove pending account activation warning when user form is submitted with errors
+* [ [#1710](https://github.com/digitalfabrik/integreat-cms/issues/1710) ] Add spacing to sidebar to improve view on small screens
 
 
 2022.10.0
 ---------
 
 * [ [#1734](https://github.com/digitalfabrik/integreat-cms/issues/1734) ] Increase timeout for SUMM.AI API client
 
-
```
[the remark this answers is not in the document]
JoeyStk: Oh yes!

integreat_cms/static/src/css/style.scss
JoeyStk: I changed it to 2.5rem now. However with this change 10px get cut off at the bottom. Can we maybe settle at 2.75rem? :D

timoludwig: I agree with @svenseeberg, `2.5rem` should be the perfect value. The version string has a line height of `1.5rem` and a padding of `0.5rem`, so `1.5 + 0.5 + 0.5 = 2.5`... I cannot reproduce that anything is cut off or 10px are missing somewhere...

timoludwig: Oh, I'm sorry, now I can indeed reproduce the small overlap of 6px :see_no_evil: 
I think this is due to the incorrect calculation of the max height... the logo has a height of 56px, but only 50px are deducted from 100vh for the menu...

integreat_cms/static/src/css/style.scss
```diff
@@ -53,6 +53,7 @@ header {
 nav#primary-navigation {
   #menu {
     max-height: calc(100vh - 50px);
```
timoludwig, line 55:
```suggestion
    max-height: calc(100vh - 56px);
```
The logo has a height of 56px, so we should calculate the correct value here.

integreat_cms/static/src/css/style.scss
```diff
@@ -53,6 +53,7 @@ header {
...
+    padding-bottom: 2.5rem;
```
timoludwig, line 56:
```suggestion
```
I would put this into the template by changing `pb-2` to `pb-10` here:
https://github.com/digitalfabrik/integreat-cms/blob/33a4ccf6fb2734fd2f89f077a045136cae88b2d6/integreat_cms/cms/templates/_base.html#L99

timoludwig (changes requested):

> Initially I wanted to use the tailwind class "pb-12" since this equals to 3rem, but for me it didn't work. Neither Chrome nor Firefox registered this class. However p-12 worked, which is strange since if p-12 is working, also pb-12 should be working. [See more about this here](https://tailwindcss.com/docs/customizing-spacing#default-spacing-scale).

How did you try this, did you just use the Browser developer tools? Then this is the expected behavior, since we have a [CssMinimizerWebpackPlugin](https://github.com/digitalfabrik/integreat-cms/blob/33a4ccf6fb2734fd2f89f077a045136cae88b2d6/webpack.config.js#L121) which removes any unused css classes from the static file bundle. To test this properly, you need to really edit the template, wait for the webpack build to finish and then reload the page. All tailwind classes which are used somewhere else in our templates will also work dynamically in the Browser's developer tools, but the rest needs to be registered first.

JoeyStk:
Ah yes, I tried it using the developer tools. Thank you for the explanation. Alternatively I could also remove the pb-10 class and instead just set the max-height to 100vh - 96px. This includes the logo on the top and the version number on the bottom. This would safe us using one class. But this way it's also working

timoludwig (approved):
Thanks! :+1:

> Alternatively I could also remove the pb-10 class and instead just set the max-height to 100vh - 96px. This includes the logo on the top and the version number on the bottom.

Sure, this would work as well. Just use whichever version you prefer :slightly_smiling_face:

PR #1755, opened 2022-10-12
Fix broken inline icons in PDF exports

charludo:
### Short description
Fix broken inline icons in PDF exports by removing the css rule causing the issue.

### Proposed changes
- `default_css` for PDF exports is no longer provided by `xhtml2pdf`, but by a local copy with the fix applied
- I'll open a PR with `xhtml2pdf` to fix this upstream, so we can remove the local copy again

### Side effects
- none that I know of

### Resolved issues

Fixes: #1537

### DIFF
--- a/integreat_cms/cms/utils/pdf_utils.py
+++ b/integreat_cms/cms/utils/pdf_utils.py
@@ -15,8 +15,8 @@ from django.template.loader import get_template
 from django.utils.text import capfirst
 from django.utils.translation import ugettext as _
 from django.views.decorators.cache import never_cache
-
 from xhtml2pdf import pisa
+from xhtml2pdf.default import DEFAULT_CSS
 
 
 from .text_utils import truncate_bytewise
@@ -110,10 +110,18 @@ def generate_pdf(region, language_slug, pages):
         html = get_template("pages/page_pdf.html").render(context)
         # Save empty file
         pdf_storage.save(filename, ContentFile(""))
+
+        # Get fixed version of default pdf styling (see https://github.com/digitalfabrik/integreat-cms/issues/1537)
+        fixed_css = DEFAULT_CSS.replace("background-color: transparent;", "", 1)
+
         # Write PDF content into file
         with pdf_storage.open(filename, "w+b") as pdf_file:
             pisa_status = pisa.CreatePDF(
-                html, dest=pdf_file, link_callback=link_callback, encoding="UTF-8"
+                html,
+                dest=pdf_file,
+                link_callback=link_callback,
+                encoding="UTF-8",
+                default_css=fixed_css,
             )
         # pylint: disable=no-member
         if pisa_status.err:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ UNRELEASED
 ----------
 
 * [ [#1759](https://github.com/digitalfabrik/integreat-cms/issues/1759) ] Add line break between images in PDF exports
+* [ [#1537](https://github.com/digitalfabrik/integreat-cms/issues/1537) ] Fix broken inline icons in PDF exports
 * [ [#951](https://github.com/digitalfabrik/integreat-cms/issues/951) ] Add possibility to create categories for POIs
 * [ [#1742](https://github.com/digitalfabrik/integreat-cms/issues/1742) ] Add last modified date to media sidebar
 * [ [#1703](https://github.com/digitalfabrik/integreat-cms/issues/1703) ] Remove pending account activation warning when user form is submitted with errors
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -6394,7 +6394,7 @@ msgid "No valid pages selected for PDF generation."
 msgstr ""
 "Es wurden keine gültigen Seiten zur Erzeugung der PDF-Datei ausgewählt."
 
-#: cms/utils/pdf_utils.py:127
+#: cms/utils/pdf_utils.py:135
 msgid "The PDF could not be successfully generated."
 msgstr "PDF-Datei konnte nicht erfolgreich erzeugt werden."
 


integreat_cms/cms/utils/pdf_utils.py
```diff
@@ -110,10 +110,22 @@ def generate_pdf(region, language_slug, pages):
         html = get_template("pages/page_pdf.html").render(context)
         # Save empty file
         pdf_storage.save(filename, ContentFile(""))
+
+        # Get fixed version of default pdf styling (see https://github.com/digitalfabrik/integreat-cms/issues/1537)
+        css_file_path = os.path.join(
+            os.path.dirname(__file__), "../../static/src/css/pdf_default_styles.css"
```
svenseeberg, line 116: Maybe `settings.STATIC_ROOT` is a better source for accessing the `static` directory?

svenseeberg (commented):
Generally this looks valid, thanks. One question though: is the change to the CSS too specific to create an upstream pull request for the CSS file?

integreat_cms/cms/utils/pdf_utils.py
charludo: Ah, right. Will do, thanks.

charludo:
> Generally this looks valid, thanks. One question though: is the change to the CSS too specific to create an upstream pull request for the CSS file?

As I mentioned in the PR, I'll create another PR with `xhtml2pdf`. But seeing as they have some open PRs dating back to March, I think we should fix this locally, at least as a temporary measure.

membralala (commented):
Thanks, @charludo, for me this works fine. Apart from @svenseeberg's suggestion, I think this fix is worth a mention in our ```CHANGELOG.md```. 🙂

charludo:
> Thanks, @charludo, for me this works fine. Apart from @svenseeberg's suggestion, I think this fix is worth a mention in our `CHANGELOG.md`. slightly_smiling_face

Oh, right! Sorry, still getting to know the ropes around here 😄

integreat_cms/cms/utils/pdf_utils.py
timoludwig: Unfortunately, this isn't as straightforward as it might seem: `STATIC_ROOT` is only used in production mode and we probably want to fix the bug in dev mode as well.

The universal alternative to get the path of static files in both production and develop modee would be 
```
from django.contrib.staticfiles import finders

css_file_path = finders.find('css/pdf_default_styles.css')
```
However, this doesn't work out of the box because we have an intermediary bundling step of the static files via webpack - so in order for this to work, we would have to create a new file `pdf_default_styles.ts` which imports the css file and is then referenced in the [webpack config](https://github.com/digitalfabrik/integreat-cms/blob/develop/webpack.config.js).

So in this case, I guess I would prefer it if we would bypass the whole problem and just modify the existing default css:

```
from xhtml2pdf.default import DEFAULT_CSS

# Remove first occurrence of default html background color style
fixed_css = DEFAULT_CSS.replace("background-color: transparent;", "", 1)
```

timoludwig (changes requested):
Thanks a lot! :+1:

Due to our static files bundling setup, I'd suggest to remove the hard coded fixed file and modify the default CSS in place instead.

integreat_cms/cms/utils/pdf_utils.py
timoludwig: Oh, or another solution which would be a bit more robust against changes to the upstream repo would be to add the lines
```
html {
    background-color: inherit;
}
```
to the file [integreat_cms/static/src/css/pdf_page_export.css](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/static/src/css/pdf_page_export.css)

charludo: @timoludwig ah, the first suggestion is much nicer, agreed.

The second one (resetting the affected style) unfortunately doesn't work. I initially tried to solve the problem that way, but having *any* `background-color: ` in the `html` style rules breaks the icons - `inherit`, `initial`, `unset`, rgb colors, even invalid values. That's why I think there's some underlying issue, not necessarily with the CSS itself (otherwise the `transparent` background color should be fine as well)

timoludwig: Ah okay, I tried it earlier and it seemed to work... 
One potential pitfall: our PDFs are cached in the src/integreat_cms/pdf directory, so maybe the broken version of the pdf was still cached and you didn't notice that the fix works? Could you delete the contents of that directory and try again? Sorry if I mixed something up and this does indeed not work :innocent:

timoludwig (changes requested):
Thanks a lot, this fixes the issue :ok_hand: 

Could you rebase your branch and squash your commits to clean up the git history a bit? :innocent:

charludo:
> Thanks a lot, this fixes the issue ok_hand
> 
> Could you rebase your branch and squash your commits to clean up the git history a bit? innocent

Ahh sorry, I merged instead of rebasing. Do you have a suggestion on how to squash from here?

timoludwig:
> Ahh sorry, I merged instead of rebasing. Do you have a suggestion on how to squash from here?

Yes, two possibilities:

1. Either with interactive rebasing:
  ```
  git fetch --all
  git rebase -i origin/develop
  # Only leave "pick" in the first line and replace all other "pick" by "fixup"
  # Resolve conflicts with ./dev-tools/resolve_translation_conflicts.sh
  git push --force-with-lease
  ```

2. Or via a temporary branch:
  ```
  git fetch --all
  git checkout -b tmp develop
  git merge --squash bug/inline-icons-in-pdf-export-broken
  git commit -m "Fix broken inline icons in PDF exports"
  git checkout bug/inline-icons-in-pdf-export-broken
  git reset --hard tmp
  git push --force-with-lease
  ```

charludo:
> > Ahh sorry, I merged instead of rebasing. Do you have a suggestion on how to squash from here?
> 
> Yes, two possibilities:
> 
>     1. Either with interactive rebasing (the merge commit will just disappear afterwards):
> 
> 
> ```
> git fetch --all
> git rebase -i origin/develop
> # Only leave "pick" in the first line and replace all other "pick" by "fixup"
> # Resolve conflicts with ./dev-tools/resolve_translation_conflicts.sh
> git push --force-with-lease
> ```
> 
>     2. Or via a temporary branch:
> 
> 
> ```
> git fetch --all
> git checkout -b tmp origin/develop
> git merge --squash bug/inline-icons-in-pdf-export-broken
> git commit -m "Fix broken inline icons in PDF exports"
> git checkout bug/inline-icons-in-pdf-export-broken
> git reset --hard tmp
> git push --force-with-lease
> ```

Thanks! Also, I'm bookmarking this.

timoludwig (approved):
Thanks :ok_hand:

ulliholtgrave (approved):
Very nice :+1: I guess this PR is ready to merge after you fix the last conflict :)

Issue #1672, opened 2022-09-07
Switch to lucide as an icon provider


david-venhoff:
### Motivation

Feather has some performance problems and is not well maintained.
> Lucide also runs with a treeshake-approach to reduce the replacement time (e.g. see https://github.com/lucide-icons/lucide/tree/main/packages/lucide#lucide). Maybe this could also help us improve the general performance of our CMS by defining our used icons on each view.

https://github.com/digitalfabrik/integreat-cms/issues/1379#issuecomment-1235752087

### Alternatives

Keep using both feather and  lucide

PR #1757, opened 2022-10-12
Use lucide icon provider

david-venhoff:
### Short description

This pr uses lucide instead of feather for rendering icons in the browser, further improving the performance of https://github.com/digitalfabrik/integreat-cms/pull/1379 to make it hopefully usable for large page trees.

### Proposed changes

- Switch to lucide
- Don't query every icon on the document when dynamically rendering an image after an ajax call
- rename the `data-feather` attribute to `icon-name`

### Unresolved questions:
- ~Should this pr include the current implementation of `create-icons.ts` or should I create a pr to lucide?~
- ~Why are most dependencies dev-dependencies? And should lucide be added to the dev-dependency section too?~

### Side effects

- New image icons must use `icon-name="..."` instead of `data-feather="..."`

### Resolved issues

Fixes: #1672 

### DIFF
--- a/integreat_cms/static/src/css/style.scss
+++ b/integreat_cms/static/src/css/style.scss
@@ -86,6 +86,13 @@ nav#primary-navigation {
   }
 }
 
+i[icon-name] {
+  width: 24px;
+  height: 24px;
+  display: inline-block;
+  vertical-align: middle;
+}
+
 h1.heading {
   @apply text-3xl mb-2;
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,10 +4,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "integreat-cms",
-      "dependencies": {
-        "dropzone": "^6.0.0-beta.2"
-      },
       "devDependencies": {
         "@babel/cli": "^7.18.10",
         "@babel/core": "^7.19.0",
@@ -16,7 +12,6 @@
         "@babel/preset-env": "^7.19.0",
         "@tailwindcss/forms": "^0.5.3",
         "@types/dropzone": "^5.7.4",
-        "@types/feather-icons": "^4.7.0",
         "@types/tinymce": "^4.6.5",
         "autoprefixer": "10.4.8",
         "babel-loader": "^8.2.5",
@@ -27,12 +22,13 @@
         "core-js": "^3.25.1",
         "css-loader": "^6.7.1",
         "css-minimizer-webpack-plugin": "^4.1.0",
+        "dropzone": "^6.0.0-beta.2",
         "element-closest": "^3.0.2",
-        "feather-icons": "^4.29.0",
         "file-loader": "^6.2.0",
         "flagpack-dart-sass": "^1.3.0",
         "history": "^5.3.0",
         "htmldiff-js": "1.0.5",
+        "lucide": "^0.91.0",
         "lucide-preact": "^0.88.0",
         "mini-css-extract-plugin": "^2.6.1",
         "postcss-cli": "^10.0.0",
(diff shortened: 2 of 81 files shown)


ulliholtgrave (approved):
Thank you very much for your effort :+1:
I clicked through a lot of our views and I couldn't find any issues with your changes:

Regarding your questions:
- The main difference between devDependencies and "normal" dependencies are the usage in the bundled frontend version. For example: We are using some minifiers and converters like babel that are required for the bundling of the resulting JS file. As lucide is a pure frontend library that is required for our frontend operations I would put it as a "normal" dependency
- I really like your approach of not rendering each icon after another, so it would be nice if you could open a proposal in the lucide repository. However, I would keep your solution and merge it in our repo first and maybe we can adjust it afterwards if the accept your proposal :+1:

package.json
```diff
@@ -24,18 +23,17 @@
     "css-loader": "^6.7.1",
     "css-minimizer-webpack-plugin": "^4.1.0",
     "element-closest": "^3.0.2",
-    "feather-icons": "^4.29.0",
     "file-loader": "^6.2.0",
     "flagpack-dart-sass": "^1.3.0",
     "history": "^5.3.0",
     "htmldiff-js": "1.0.5",
+    "lucide-preact": "^0.88.0",
```
timoludwig, line 30:
```suggestion
    "lucide": "^0.91.0",
    "lucide-preact": "^0.88.0",
```
> The main difference between devDependencies and "normal" dependencies are the usage in the bundled frontend version. For example: We are using some minifiers and converters like babel that are required for the bundling of the resulting JS file. As lucide is a pure frontend library that is required for our frontend operations I would put it as a "normal" dependency

I would argue that all of our dependencies are dev dependencies, since we don't use npm in production.
Normally, the separation is used to execute `npm install --production` on your production system which would then only install the dependencies without the dev dependencies. In our case however, the prod server does not have npm installed and doesn't know our `package.json` and just gets the bundled static files.

package.json
```diff
@@ -24,18 +23,17 @@
     "css-loader": "^6.7.1",
     "css-minimizer-webpack-plugin": "^4.1.0",
```
timoludwig, line 24:
```suggestion
    "css-minimizer-webpack-plugin": "^4.1.0",
    "dropzone": "^6.0.0-beta.2",
```

package.json
```diff
@@ -55,6 +53,7 @@
     "whatwg-fetch": "^3.6.2"
   },
   "dependencies": {
```
timoludwig, line 55:
```suggestion
  }
```

package.json
```diff
@@ -55,6 +53,7 @@
...
-    "dropzone": "^6.0.0-beta.2"
+    "dropzone": "^6.0.0-beta.2",
+    "lucide": "^0.91.0"
   }
```
timoludwig, line 58:
```suggestion
```

timoludwig (changes requested):
Thanks a lot, really cool! :+1:

Unfortunately, I still see the same small glitching (e.g. in the menu) when clicking through the CMS like with feather... it still takes a perceivable amount of time until the icons are rendered.
To be honest: I don't really understand the replacement code and am not sure how it could potentially optimized. But I'm fine with just merging this and try it out. It does at least not seem to harm the performance :sweat_smile:

integreat_cms/static/src/css/style.scss
```diff
@@ -86,6 +86,13 @@ nav#primary-navigation {
   }
 }
 
+i[icon-name] {
+  width: 24px;
+  height: 24px;
+  display: inline-block;
+  vertical-align: middle;
+}
```
timoludwig, line 94: Ah, very cool idea to allocate the space from the beginning! :ok_hand:

package-lock.json
```diff
@@ -6,7 +6,8 @@
     "": {
       "name": "integreat-cms",
       "dependencies": {
-        "dropzone": "^6.0.0-beta.2"
+        "dropzone": "^6.0.0-beta.2",
+        "lucide": "^0.91.0"
       },
```
timoludwig, line 11: Could you execute `npm install` once and add the changes to `package-lock.json` to this PR as well?

timoludwig (changes requested):
Very cool, thanks! :rocket: 

Could you update the `package-lock.json` as well and squash your commits? Then this PR is ready! :tada:

timoludwig (approved):
Thanks! :muscle:

PR #1729, opened 2022-10-04
Catch `asyncio.TimeoutError` in SUMM.AI API client

timoludwig:
### Short description

Catch `asyncio.TimeoutError` in SUMM.AI API client

### Resolved issues

Fixes: #1728

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #1760, opened 2022-10-14
Fix false positive error message for urlencoded filenames in PDF export

charludo:
### Short description
Media files containing special chars were falsely reported as not found

### Proposed changes
- Use `urllib.parse.unquote` to decode the path

### Side effects

- none I am aware of 

### Resolved issues

Fixes: #1499

### DIFF
--- a/integreat_cms/cms/utils/pdf_utils.py
+++ b/integreat_cms/cms/utils/pdf_utils.py
@@ -2,7 +2,7 @@ import hashlib
 import logging
 import os
 
-from urllib.parse import urlparse
+from urllib.parse import urlparse, unquote
 
 from django.conf import settings
 from django.contrib.staticfiles import finders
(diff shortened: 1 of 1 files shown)


ulliholtgrave (approved):
Good catch! :) This seems to fix the issue :+1:

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 UNRELEASED
 ----------
 
+* [ [#1499](https://github.com/digitalfabrik/integreat-cms/issues/1499) ] Fix false positive error message for urlencoded filenames in PDF export
```
timoludwig, line 4:
```suggestion
```
We use the changelog to keep other teams and users of the CMS informed about relevant changes.
Normally we don't include changes here that are not noticeable to them.
Since the error message was only displayed in the server log and the PDF export actually worked properly, I would consider this an implementation detail that is not relevant to the changelog.

timoludwig (changes requested):
Thanks a lot, this fixes the issue! :muscle: 

Could you remove the changelog entry in this case and squash your commits?

CHANGELOG.md
ulliholtgrave: I am not really sure if I can agree on that. As we have an open issue that is visible to other team members I would argue that it is a valid point to keep track of its fix in our changelog :thinking:

timoludwig: Ok, but the team members who keep track of GitHub issues also can see when they're closed... I thought we maintain our changelog for people who don't understand `git log` or are not able to track the progress just by following the repository.

Actually I'm fine with both ways - either list all closed issues or only those relevant for cms users - but we should stay consistent in our guidelines. And in the past, we also didn't include changelog entries for purely technical changes, even if they had corresponding issues, see e.g. #1757,  #1729, #1715 and many more.
Maybe let's discuss this in the next team call :wink:

timoludwig (approved):
Thanks! :pray:

PR #1764, opened 2022-10-15
Provide fallback translations for imprint feedbacks

membralala:
### Short description

This PR should fix the feedback view error, that was shown, when imprint feedback was submitted in a language, that does not exist for this regions feedback.

### Proposed changes

- Use fallback translation for the title of an imprint feedback, if the language exists in this region
- Use default backend language, otherwise

### Resolved issues

Fixes:  #1717

There was a previous PR, where I used a different solution which did not work out. When I dropped the changes, the PR was closed automatically. However, for full discussion also see #1730.

### DIFF
--- a/integreat_cms/api/v3/feedback/imprint_page_feedback.py
+++ b/integreat_cms/api/v3/feedback/imprint_page_feedback.py
@@ -73,7 +73,7 @@ def imprint_page_feedback_internal(
     :return: JSON object according to APIv3 imprint feedback endpoint definition
     :rtype: ~django.http.JsonResponse
     """
-    if region.imprint:
+    if region.imprint and language in region.visible_languages:
         ImprintPageFeedback.objects.create(
             region=region,
             language=language,
@@ -88,4 +88,6 @@ def imprint_page_feedback_internal(
         region,
         language,
     )
-    raise Http404("The imprint does not exist in this region")
+    raise Http404(
+        "The imprint does not exist in this region for the selected language."
+    )
--- a/integreat_cms/cms/models/feedback/imprint_page_feedback.py
+++ b/integreat_cms/cms/models/feedback/imprint_page_feedback.py
@@ -20,7 +20,11 @@ class ImprintPageFeedback(Feedback):
         :rtype: str
         """
         try:
-            return self.region.imprint.get_translation(self.language.slug).title
+            translation = (
+                self.region.imprint.get_translation(self.language.slug)
+                or self.region.imprint.default_translation
+            )
+            return translation
         except ImprintPage.DoesNotExist:
             return _("Imprint")
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ UNRELEASED
 
 * [ [#1718](https://github.com/digitalfabrik/integreat-cms/issues/1718) ] Enable submitting feedback about fallback translations of events and pois
 * [ [#1793](https://github.com/digitalfabrik/integreat-cms/issues/1793) ] Fix sending feedback for recurring events
+* [ [#1717](https://github.com/digitalfabrik/integreat-cms/issues/1717) ] Provide fallback translations for imprint feedbacks
 
 
 2022.10.2
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-10-27 09:09+0000\n"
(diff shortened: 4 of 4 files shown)


integreat_cms/cms/models/feedback/imprint_page_feedback.py
```diff
@@ -20,7 +20,11 @@ def object_name(self):
         :rtype: str
         """
         try:
-            return self.region.imprint.get_translation(self.language.slug).title
+            if self.language in self.region.languages:
+                return self.region.imprint.get_public_translation(
+                    self.language.slug
+                ).title
+            return self.region.imprint.default_translation.title
```
david-venhoff, line 27:
```suggestion
            return self.region.imprint.get_public_translation(
                self.language.slug
            ).title
```
The if-statement does not guarantee that the imprint translation exists in this region, since a region could e.g. just have a german imprint while supporting other languages. However, it looks like just using `get_public_translation` may be enough in this case, because it returns a fallback translation for imprints when there is no translation of the actual language found.

david-venhoff (commented):
Thanks, I think this fixes the issue. I just noticed a small potential improvement.

integreat_cms/cms/models/feedback/imprint_page_feedback.py
timoludwig:
```suggestion
            translation = (
                self.region.imprint.get_translation(self.language.slug)
                or self.region.imprint.default_translation
            )
            return translation.title
```
In addition to @david-venhoff's comment: In the case that the imprint only exists in draft state, `get_public_translation()` does not return a translation, even if fallback translations are enabled.

timoludwig (changes requested):
Thanks a lot! :+1:

Unfortunately this still throws an `AttributeError` when the imprint only exists as draft.

integreat_cms/cms/models/feedback/imprint_page_feedback.py
membralala: Thanks for your feedbacks! I'm sorry for struggling around with such a small issue, I'm getting really confused with the possible ways of translation and their impact in different scenarios. 🙈  I think @timoludwig 's suggestion works for me in that it does not produce an error anymore. 
However I have a few more questions: 
1. I implemented `if self.language in self.region.languages:` because at the moment the API allows to send a feedback in a language, that does not exist for this region (means it does not exist in the regions language tree). Therefore e.g. in @david-venhoff's example if someone would send an API-call with a language, that exists in the database in general, but not for the specific region, an error would show up again. With Timo's soultion this would not happen, but doesn't it still make sense to not allow this in the API and return a 404 instead? 
2. A more general question: What is the proper meaning of a fallback translation in the Integreat context? Is it always the translation of the default language of a region or is it the next parent translation of a language in the language tree, that exists? E.g say that 
```
DE (default)
   EN 
      ES 
```
is the language tree and a page exists in DE and EN, what is the fallback translation for ES? 
I'm sorry, I just found that I really have a lack of understanding here. 👀

timoludwig: > I'm sorry for struggling around with such a small issue, I'm getting really confused with the possible ways of translation and their impact in different scenarios.

No worries, we should probably write more documentation on this topic! 

> 1. But doesn't it still make sense to not allow this in the API and return a 404 instead? 

Yes, correct, in the API this makes sense, but for the representation of the existing objects I wouldn't limit it, since e.g. a feedback entry could be created while a language is active, then the language is disabled but the feedback entry still exists in the database and somehow needs to be displayed in the list.

> 2. What is the proper meaning of a fallback translation in the Integreat context? Is it always the translation of the default language of a region? 

Yes, always the default language! The order of the tree only counts for the internal translation management and has no influence on the content delivered to the users.

membralala: Okay, thank you very much for clarification, then I will just add the additional catch in the API.

integreat_cms/api/v3/feedback/imprint_page_feedback.py
```diff
@@ -73,7 +73,7 @@ def imprint_page_feedback_internal(
     :return: JSON object according to APIv3 imprint feedback endpoint definition
     :rtype: ~django.http.JsonResponse
     """
-    if region.imprint:
+    if region.imprint and language in region.languages:
```
timoludwig, line 76:
```suggestion
    if region.imprint and language in region.visible_languages:
```
Probably it makes sense to only allow visible languages here.

timoludwig (changes requested):
Thanks a lot! :+1:
Sorry for taking so long to review this, somehow I didn't realize the feedback was already implemented :smile:
Just a minor change request, other than that please just rebase and fix the changelog conflict, then it's good to go! :tada:

timoludwig (approved):
Thanks! :muscle:

david-venhoff (approved):

PR #1768, opened 2022-10-17
Change statistics calculcation, fixes #1767

### LINKED ISSUE #1767: Change statistics calculcation
### Motivation
In the past we always included all downloads (single pages and full content downloads) in the per language download count. The offline download number only gave an rough indication of how large the portion of app users is. This should not influence the language counter.

### Proposed Solution
Also count offline downloads in the languages.

### Alternatives
Rewrite history?

### Additional Context
https://wiki.integreat-app.de/statistiken
https://github.com/digitalfabrik/helper-bash-scripts/blob/master/matomo/statistics.py#L168

svenseeberg:
### Short description
Include offline downloads in per language download numbers.

### Proposed changes
Change Matomo API request to count all downloads per language.

### Side effects
Invalidates numbers from last months, but is compatible with statistics over last years.

### Resolved issues
Fixes: #1767

### DIFF
--- a/integreat_cms/cms/utils/matomo_api_manager.py
+++ b/integreat_cms/cms/utils/matomo_api_manager.py
@@ -209,7 +209,7 @@ class MatomoApiManager:
                 "labels": self.simplify_date_labels(dataset.keys(), period),
                 "datasets": [
                     {
-                        "label": _("All languages"),
+                        "label": _("Total Accesses"),
                         "borderColor": colors.DEFAULT,
                         "data": list(dataset.values()),
                     }
@@ -249,7 +249,7 @@ class MatomoApiManager:
                     self.fetch(
                         session,
                         **query_params,
-                        segment=f"pageUrl=@/{language.slug}/wp-json/;pageUrl!@/pages/",
+                        segment=f"pageUrl=@/{language.slug}/wp-json/extensions/v3/",
                     )
                 )
                 for language in languages
@@ -257,10 +257,24 @@ class MatomoApiManager:
             # Create separate task to gather offline download hits
             tasks.append(
                 loop.create_task(
-                    self.fetch(session, **query_params, segment="pageUrl=@/pages/"),
+                    self.fetch(
+                        session,
+                        **query_params,
+                        segment="pageUrl=@/wp-json/extensions/v3/pages/",
+                    ),
                 )
             )
-            # Create task for total visits in all languages
+            # Create separate task to gather WebApp download hits
+            tasks.append(
+                loop.create_task(
+                    self.fetch(
+                        session,
+                        **query_params,
+                        segment="pageUrl=@/wp-json/extensions/v3/children/",
+                    ),
+                )
+            )
+            # Create task for all downloads
             tasks.append(
                 loop.create_task(
                     self.fetch(
@@ -322,6 +336,8 @@ class MatomoApiManager:
         logger.debug("All asynchronous fetching tasks have finished.")
         # The last dataset contains the total visits
         total_visits = datasets.pop()
+        # Get the separately created datasets for webapp downloads
+        webapp_downloads = datasets.pop()
         # Get the separately created datasets for offline downloads
         offline_downloads = datasets.pop()
 
@@ -347,15 +363,23 @@ class MatomoApiManager:
                 # The dataset for offline downloads
                 + [
                     {
-                        "label": _("Offline Downloads"),
+                        "label": _("Offline Accesses"),
                         "borderColor": next(color_cycle),
                         "data": list(offline_downloads.values()),
+                    }
+                ]
+                # The dataset for online/web app downloads
+                + [
+                    {
+                        "label": _("WebApp Accesses"),
+                        "borderColor": next(color_cycle),
+                        "data": list(webapp_downloads.values()),
                     }
                 ]
                 # The dataset for total visits
                 + [
                     {
-                        "label": _("All languages"),
+                        "label": _("Total Accesses"),
                         "borderColor": colors.DEFAULT,
                         "data": list(total_visits.values()),
                     }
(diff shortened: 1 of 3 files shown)


ulliholtgrave (commented):
I am pretty sure that me or @timoludwig added this extra filter for a reason. If I remember it correctly we are counting calls double as an offline and a single page hit :thinking:

svenseeberg:
> I am pretty sure that me or @timoludwig added this extra filter for a reason. If I remember it correctly we are counting calls double as an offline and a single page hit thinking

Exactly this is intended.

timoludwig (commented):
> Exactly this is intended. The numbers are not supposed to add up, because languages and offline/online are totally different categories.

In #1571, the request was to separate them: 

>  The hits from the native app where separated and summarized in a single coloumn "Offline Downloads". These hits were subtracted from the language-specific-web-app-page-hits.

However, if we don't, I think we should redesign the graph & table because in its current form it is not comprehensible which numbers exactly were collected.

Two ideas:
1. Just change the label from:
    - "All languages" to "All languages (online & offline downloads)" and from
    - "Offline Downloads" to "All languages (only offline downloads)" or similar
    
    to point out that these numbers are already contained in all other lines
3. Add a new select filter to the graph to show three options:
    - "Online accesses"
    - "Offline downloads"
    - "Online accesses & Offline downloads"
    
    and based on this value we set the segment filter for the matomo api.

svenseeberg:
We could/should add another column "Online/Web App Downloads" in the future and then this is very easy to understand. But at least with the status quo we're in a state that editors are used to.

charludo:
> We need to revert this to the old numbers the editors are used to. This is a good start for enhancing later. I'll rename the columns for this PR, but then let's please merge and create a release together with #1800 ASAP.

If possible, I think including #1802 in that release would be a good idea. Working on it now.

timoludwig (approved):
> We need to revert this to the old numbers the editors are used to. This is a good start for enhancing later. I'll rename the columns for this PR, but then let's please merge and create a release together with #1800 ASAP.

Alright! :+1:

svenseeberg:
Okay, I added online downloads. It now looks like this:

image

integreat_cms/cms/utils/matomo_api_manager.py
```diff
@@ -249,18 +249,24 @@ async def get_visits_per_language_async(self, loop, query_params, languages):
...
                 loop.create_task(
-                    self.fetch(session, **query_params, segment="pageUrl=@/pages/"),
+                    self.fetch(session, **query_params, segment="pageUrl=@/wp-json/extensions/v3/pages/"),
                 )
             )
-            # Create task for total visits in all languages
+            # Create separate task to gather online/web app download hits
+            tasks.append(
+                loop.create_task(
+                    self.fetch(session, **query_params, segment="pageUrl=@/wp-json/extensions/v3/children/"),
```
timoludwig, line 266: I don't think the children endpoint is enough here... this will just cause a lot more inconsistencies since pages endpoint + children endpoint will not match the total number of all downloads.
I think we should invert the condition for offline downloads to get the number which we refer to as "online downloads"...

integreat_cms/cms/utils/matomo_api_manager.py
```diff
@@ -352,10 +360,18 @@ def get_visits_per_language(self, start_date, end_date, period):
                         "data": list(offline_downloads.values()),
                     }
                 ]
+                # The dataset for online/web app downloads
+                + [
+                    {
+                        "label": _("Online/WebApp Downloads"),
```
timoludwig, line 366: "Online download" sounds kind of strange to me, but not sure whether my suggestion of "online accesses" is really better :sweat_smile:

timoludwig (changes requested):

integreat_cms/cms/utils/matomo_api_manager.py
svenseeberg: Or simply "WebApp Downlods" ?

integreat_cms/cms/utils/matomo_api_manager.py
svenseeberg: Why not? We only exactly track these 2 endpoints. And the numbers do add up: 
image

timoludwig: Ahh, I wasn't aware that the single page endpoint isn't used at all and all other endpoints are not tracked via Matomo. Alright, sorry, then it's fine.

integreat_cms/cms/utils/matomo_api_manager.py
svenseeberg: Agreed ;-)

svenseeberg: I changed it to "Accesses" and "Zugriffe" everywhere.

integreat_cms/cms/utils/matomo_api_manager.py
```diff
@@ -347,15 +363,23 @@ def get_visits_per_language(self, start_date, end_date, period):
                 # The dataset for offline downloads
                 + [
                     {
-                        "label": _("Offline Downloads"),
+                        "label": _("Offline Accesses"),
```
timoludwig, line 366:
```suggestion
                        "label": _("Offline Downloads"),
```
Maybe offline is the only use case where we should avoid talking about "accesses", because we don't know how often the users actually access the app? Not sure though, would also be fine for me to use the same wording for all columns.

timoludwig (approved):

integreat_cms/cms/utils/matomo_api_manager.py
svenseeberg: Well, I would stay consistent over all columns. Especially because we summarize the offline downloads/accesses in the total accesses column as well.

timoludwig: Alright, makes sense! :+1:

ulliholtgrave (approved):
LGTM :+1:

Issue #1170, opened 2022-02-15
Add map preview on POI form


timoludwig:
### Motivation

When POIs are created, in wordpress there is a map preview for the address/coordinates of the POI.

### Proposed Solution

- [x] Remove placeholder "Google Map vs. OSM"
- [ ] Add map preview for location

### Additional Context

In the long term, we can also use this map to set the address (see #1035)

PR #1775, opened 2022-10-17
Add map preview on POI form

MizukiTemma:
### Short description

This PR adds a map preview function on the POI form.

### Proposed changes

- Add a button "Preview on Map".
- The location will be shown on the map as a pin, if coordinates are given (or auto-filled). Otherwise an error message* (will be decorated later).
- Uses https://github.com/maplibre/maplibre-gl-js
- Preview in popup as hidden overlay

*I'm not sure what kind of error message is good for the case that no coordinates are given. Should the button apper first after the coordinates are given/auto-calculated?

### Side effects

- The POI form page might be slow now, having an ajax, many events and live map popup.

### Resolved issues

Fixes: #1170 

### DIFF
--- a/integreat_cms/cms/views/pois/poi_actions.py
+++ b/integreat_cms/cms/views/pois/poi_actions.py
@@ -221,7 +221,9 @@ def auto_complete_address(request, region_slug):
     return JsonResponse(
         data={
             "postcode": address.get("postcode"),
-            "city": address.get("city"),
+            "city": address.get("city")
+            or address.get("town")
+            or address.get("village"),
             "country": address.get("country"),
             "longitude": result.longitude,
             "latitude": result.latitude,
--- a/integreat_cms/static/src/index.ts
+++ b/integreat_cms/static/src/index.ts
@@ -74,7 +74,7 @@ import "./js/tutorial-overlay.ts";
 
 import "./js/unsaved-warning";
 
-import "./js/pois/poi-actions.ts"
+import "./js/pois/poi-actions.ts";
 
 window.addEventListener('DOMContentLoaded', () => {
   create_icons_at(document.documentElement);
--- a/integreat_cms/static/src/js/pois/poi-actions.ts
+++ b/integreat_cms/static/src/js/pois/poi-actions.ts
@@ -15,6 +15,9 @@ window.addEventListener("load", () => {
       // Reset to initial value
       longitude.value = longitude.dataset.initial;
       latitude.value = latitude.dataset.initial;
+
+      longitude.dispatchEvent(new Event("focusout"));
+      latitude.dispatchEvent(new Event("focusout"));
     }
   });
 
@@ -79,6 +82,12 @@ async function autoCompleteAddress() {
     updateField("longitude", data.longitude);
     updateField("latitude", data.latitude);
   }
+
+  let longitude = document.getElementById("id_longitude") as HTMLInputElement;
+  let latitude = document.getElementById("id_latitude") as HTMLInputElement;
+  
+  longitude.dispatchEvent(new Event("focusout"));
+  latitude.dispatchEvent(new Event("focusout"));
 }
 
 function updateField(fieldName: string, value: string) {
--- a/integreat_cms/static/src/js/pois/poi-map.ts
+++ b/integreat_cms/static/src/js/pois/poi-map.ts
@@ -1,0 +1,40 @@
+import maplibregl from "maplibre-gl";
+
+window.addEventListener("load", () => {
+  let longitude = document.getElementById("id_longitude") as HTMLInputElement;
+  let latitude = document.getElementById("id_latitude") as HTMLInputElement;
+
+  showPreview();
+  longitude.addEventListener("focusout", showPreview);
+  latitude.addEventListener("focusout", showPreview);
+});
+
+
+function showPreview(){
+  let container = document.getElementById("map");
+  let lng = (document.getElementById("id_longitude") as HTMLInputElement).value;
+  let lat = (document.getElementById("id_latitude") as HTMLInputElement).value;
+  
+  if (lng && lat){
+    document.getElementById("no_coordinates_alert")?.classList.add("hidden");
+    
+    let map = new maplibregl.Map({
+        container: container, //'map',
+        style: 'https://maps.tuerantuer.org/styles/integreat/style.json',
+        center: [parseFloat(lng), parseFloat(lat)], // starting position [lng, lat]
+        zoom: 15 // starting zoom
+    });
+    
+    let marker = new maplibregl.Marker()
+    .setLngLat([parseFloat(lng), parseFloat(lat)])
+    .addTo(map);
+    
+    map.on("load", map.resize);
+    map.addControl(new maplibregl.FullscreenControl({container: container}));
+
+    container.classList.remove("hidden");
+  } else {
+    container.classList.add("hidden");
+    document.getElementById("no_coordinates_alert")?.classList.remove("hidden");
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ UNRELEASED
 * [ [#1350](https://github.com/digitalfabrik/integreat-cms/issues/1350) ] Various small PDF export improvements
 * [ [#1777](https://github.com/digitalfabrik/integreat-cms/issues/1777) ] Fix autocompleting POI address for non-staff users
 * [ [#1749](https://github.com/digitalfabrik/integreat-cms/issues/1749) ] Fix region deletion error if media library has nested structure
+* [ [#1170](https://github.com/digitalfabrik/integreat-cms/issues/1170) ] Add map preview on POI form
 
 
 2022.10.1
--- a/integreat_cms/cms/templates/pois/poi_form.html
+++ b/integreat_cms/cms/templates/pois/poi_form.html
@@ -171,6 +171,13 @@
                             {% render_field poi_form.longitude|add_error_class:"border-red-500" %}
                             <div class="help-text">{{ poi_form.longitude.help_text }}</div>
                         </div>
+                        <div>
+                            <label>{% trans 'Map' %}</label>
+                            <div id="no_coordinates_alert" class="bg-blue-100 border-l-4 border-blue-500 text-blue-500 px-4 py-3 my-2 hidden" role="alert">
+                                {% trans 'Map preview will be available when coordinates are given.' %}
+                            </div>
+                            <div id='map' class="aspect-video"></div>
+                        </div>
                     </div>
                 </div>
             </div>
@@ -324,3 +331,7 @@
 {% block javascript %}
     {% render_bundle 'editor' 'js' %}
 {% endblock javascript %}
+{% block javascript_head %}
+    {% render_bundle 'map' 'js' %}
+    {% render_bundle 'map' 'css' %}
+{% endblock %}
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -4211,7 +4211,7 @@ msgstr "Auf Google Maps öffnen"
 #: cms/templates/imprint/imprint_form.html:120
 #: cms/templates/pages/page_form.html:289 cms/templates/pages/page_sbs.html:87
 #: cms/templates/poicategories/poicategory_list.html:31
-#: cms/templates/pois/poi_form.html:238
+#: cms/templates/pois/poi_form.html:245
 #: cms/templates/regions/region_list.html:25
 #: cms/templates/settings/user_settings.html:93
 msgid "Actions"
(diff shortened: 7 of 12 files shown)


JoeyStk (changes requested):
Thank you so much for this PR. It's a really cool feature!
I have a few suggestions here: 

1. I don't know if UI/UX have given you notes on the popup. If not, I would suggest to align the popup vertically to the center. I think it looks a little bit strange sticking to the top
2. I think I would add some padding to the error message. I know it's just an error message, but I think some padding could increase readability.
3. The error message itself, I think, is fine. But yes, maybe I would disable the button until coordinates are input
4. I didn't have an issue with the loading speed, at least not with ten events using that POI

integreat_cms/cms/templates/_raw.html
```diff
@@ -9,6 +9,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=1.0, shrink-to-fit=no, user-scalable=no" />
     <title>{{ BRANDING|title }} {% trans 'Editorial System' %}</title>
     <link rel="shortcut icon" type="image/png" href="{% static 'logos/'|add:BRANDING|add:'/'|add:BRANDING|add:'-icon.svg' %}">
+    <link href='https://unpkg.com/maplibre-gl@latest/dist/maplibre-gl.css' rel='stylesheet' />
```
timoludwig, line 12: Rather than loading this from a CDN, would it also be possible to include this in our webpack build (and thus the CSS minifier which drops all unnused css classes etc)...

integreat_cms/cms/views/pois/poi_actions.py
```diff
@@ -214,10 +214,15 @@ def auto_complete_address(request):
...
+
+    city_town = address.get("city")
+    if not city_town:
+        city_town = address.get("town")
+
     return JsonResponse(
         data={
             "postcode": address.get("postcode"),
-            "city": address.get("city"),
+            "city": city_town,
```
timoludwig, line 225:
```suggestion
            "city": address.get("city") or address.get("town"),
```
Could you give an example address where Nominatim returns `town` instead of `city`?

integreat_cms/cms/views/pois/poi_actions.py
```diff
@@ -214,10 +214,15 @@ def auto_complete_address(request):
         raise Http404(_("Address could not be found"))
 
     address = result.raw.get("address", {})
+
+    city_town = address.get("city")
+    if not city_town:
+        city_town = address.get("town")
+
```
timoludwig, line 221:
```suggestion
```

integreat_cms/static/src/js/pois/poi-actions.ts
```diff
@@ -35,8 +36,43 @@ window.addEventListener("load", () => {
       latitude.dataset.initial = latitude.value;
     });
   }
+
+  document.getElementById("map_preview_button").addEventListener("click", showPreview);
 });
 
+function showPreview(){
+  const script = document.createElement("script");
+  script.src = "https://unpkg.com/maplibre-gl@latest/dist/maplibre-gl.js";
```
timoludwig, line 45: I don't think that's a good way to include the script - isn't it possible to import the things we need and let webpack bundle the stuff for us?

integreat_cms/static/src/js/pois/poi-actions.ts
```diff
@@ -35,8 +36,43 @@ window.addEventListener("load", () => {
...
+  document.body.appendChild(script);
+
+  var lng = (document.getElementById("id_longitude") as HTMLInputElement).value;
+  var lat = (document.getElementById("id_latitude") as HTMLInputElement).value;
+
+  if (lng && lat){
+    var map = new maplibregl.Map({
+      container: 'map',
+      style: 'https://api.maptiler.com/maps/streets/style.json?key=get_your_own_OpIi9ZULNHzrESv6T2vL', // stylesheet location
```
timoludwig, line 54:
```suggestion
      style: 'https://maps.tuerantuer.org/styles/integreat/style.json',
```
Probably not a good idea to include the key in the source code.
Also maybe we can use our own tileserver https://maps.tuerantuer.org/?

integreat_cms/cms/templates/pois/_map_preview_overlay.html
```diff
@@ -0,0 +1,17 @@
...
+            </button> 
+            
+            <div id='map' style='width: 400px; height: 300px;'>
+                <div id="no_coordinates_alert" class ="hidden">
+                    {% trans 'Map preview will be available when coordinates are given.' %}
+                </div>
+            </div>
+        </div>
+    </div>
+</div>
```
timoludwig, line 17: Do you think you could render this directly in the sidebar below "Position"?

integreat_cms/cms/views/pois/poi_actions.py
MizukiTemma: Windgäßchen, 86720 Nördlingen, for example.
Bamberg is also attributed as "town".

timoludwig: Interesting, thanks!

integreat_cms/cms/templates/pois/_map_preview_overlay.html
MizukiTemma: Do you mean the map preview shall be an embedded part of the form, not a pop-up?

timoludwig: I would prefer that, yes. Maybe with an additional button to show a larger map which opens the pop-up, but it would be cool if a small map would directly show the position at the first glance.

integreat_cms/cms/templates/_raw.html
MizukiTemma: I've tried it in the 5. commit. I'm not sure if I understoot it correctly 😅

timoludwig: Yes, exactly like this! :muscle: 
Now I ask myself whether we should outsource this into a separate webpack entrypoint to avoid the download for users who only use the map rarely... 
Maybe add a new line
```
map: "./integreat_cms/static/src/map.ts",
```
here:
https://github.com/digitalfabrik/integreat-cms/blob/15ce171cc8041bac2d24823e271632db56eaebf1/webpack.config.js#L10-L13

And import the CSS and JS used for the map in the new file...
Then we would have to add the lines
```
{% block javascript_head %}
    {% render_bundle 'map' 'js' %}
    {% render_bundle 'map' 'css' %}
{% endblock %}
```
to the poi form template.

integreat_cms/static/src/index.ts
```diff
@@ -74,7 +74,8 @@ import "./js/tutorial-overlay.ts";
 
 import "./js/unsaved-warning";
 
-import "./js/pois/poi-actions.ts"
+import "./js/pois/poi-actions.ts";
+import "maplibre-gl";
```
timoludwig, line 78:
```suggestion
```
I think this isn't even required, the import in poi actions should be enough.

integreat_cms/cms/templates/pois/_map_preview_overlay.html
MizukiTemma: Ok👍

MizukiTemma:
@JoeyStk @timoludwig 
Your suggestions are implementet:
1) added a small map preview and a button unter address and coordinates input fields. They appears only when the coordinates are given or auto filled. Otherwise an message. It changes live depend on the actual input value.
2) Click on the button triggers a pop-up which shows the location on a larger map.
3) The pop-up is aligned vartically center.
4) Loading of CSS for the map is out sourced.

integreat_cms/static/src/js/pois/poi-actions.ts
```diff
@@ -35,8 +45,51 @@ window.addEventListener("load", () => {
...
+  if (lng && lat){
+    container.classList.remove("hidden");
+    document.getElementById("large_map_preview_button")?.classList.remove("hidden");
+    document.getElementById("no_coordinates_alert")?.classList.add("hidden");
+
+    var map = new maplibregl.Map({
+      container: container, //'map',
+      style: 'https://maps.tuerantuer.org/styles/integreat/style.json',
+      center: [parseFloat(lng), parseFloat(lat)], // starting position [lng, lat]
+      zoom: 12 // starting zoom
```
timoludwig, line 72:
```suggestion
      zoom: 15 // starting zoom
```
I'd go with a slightly enlarged zoom factor

timoludwig (changes requested):
Thanks a lot, very cool! :muscle:

I have a few suggestions to improve the map view and to separate the webpack bundling further.

MizukiTemma:
@timoludwig 
Your suggestions are now applied.

Do you have any idea why the test is failing with the PDFs?

timoludwig:
> Do you have any idea why the test is failing with the PDFs?

Uhh, very crazy stuff going on here :sweat_smile:

This is essentially the same problem like @charludo experienced in #1781...

I noticed that in both cases the PDF tests were parallelized together with the asynchronous SUMM AI tests in the [same batch](https://app.circleci.com/pipelines/github/digitalfabrik/integreat-cms/5852/workflows/723a6eec-e72c-4986-b866-8621ff3ec044/jobs/52167/parallel-runs/12):

```
tests/summ_ai_api/summ_ai_test.py ....................                   [ 80%]
tests/pdf/test_pdf_export.py FFFF.                                       [100%] 
```

Which leads me to my current assumption: I think this is kind of a race condition with the SUMM AI test which changes the augsburg region by enabling the SUMM AI option:
https://github.com/digitalfabrik/integreat-cms/blob/b71eb5f232a3fffe74912ed7d84746c76718d285/tests/summ_ai_api/summ_ai_test.py#L46-L48
this sets the `last_updated` field of the region the current timestamp, and this timestamp is used to generate the hashes for the PDFs:
https://github.com/digitalfabrik/integreat-cms/blob/b71eb5f232a3fffe74912ed7d84746c76718d285/integreat_cms/cms/utils/pdf_utils.py#L55

And since the hashes do not match the expected PDF hashes, the tests fail.

So this is not a problem of your branch, but just bad luck... and somehow the CircleCI cache seems to persist the order of the tests on each branch, so the build on this branch will keep failing until we fix the problem and e.g. the branch `Map_preview2` [succeeds](https://app.circleci.com/pipelines/github/digitalfabrik/integreat-cms/5858/workflows/d9c7cac8-26f3-413e-a69c-b2ff5d2d7a1b) despite being on exactly the same state.

MizukiTemma:
uhhh, that's pretty mean...... 😓 
Thank you for the explanation 👍

charludo:
> Which leads me to my current assumption: I think this is kind of a race condition with the SUMM AI test which changes the augsburg region by enabling the SUMM AI option:


What's the reason this didn't previously happen?

timoludwig:
> What's the reason this didn't previously happen?

I added this test case [26 days ago](https://github.com/digitalfabrik/integreat-cms/commit/21e668d1bdf1224be9bad38e7d7800716ad769d8), maybe it's just a coincidence that it first happened on your PDF branch... (still weird though)

integreat_cms/static/src/js/pois/poi-map.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  document.getElementById("auto-fill-coordinates")?.addEventListener("input", ({ target }) => {
+    if (!(target as HTMLInputElement).checked) {
+      showPreview(map);
+    }
+  });
+});
+
+
+function showPreview(container:HTMLElement){
```
timoludwig, line 25:
```suggestion
function showPreview(){
  let container = document.getElementById("map");
```
Since we only have one map element now, I guess we can remove this function argument.

integreat_cms/static/src/js/pois/poi-map.ts
```diff
@@ -0,0 +1,51 @@
...
+  var lng = (document.getElementById("id_longitude") as HTMLInputElement).value;
+  var lat = (document.getElementById("id_latitude") as HTMLInputElement).value;
```
timoludwig, line 27:
```suggestion
  let lng = (document.getElementById("id_longitude") as HTMLInputElement).value;
  let lat = (document.getElementById("id_latitude") as HTMLInputElement).value;
```
I'd always use `let` instead of `var`.

integreat_cms/static/src/js/pois/poi-map.ts
```diff
@@ -0,0 +1,51 @@
...
+    document.getElementById("no_coordinates_alert")?.classList.add("hidden");
+    
+    var map = new maplibregl.Map({
+        container: container, //'map',
+        style: 'https://maps.tuerantuer.org/styles/integreat/style.json',
+        center: [parseFloat(lng), parseFloat(lat)], // starting position [lng, lat]
+        zoom: 15 // starting zoom
+    });
+    
+    var marker = new maplibregl.Marker()
```
timoludwig, line 39:
```suggestion
    let marker = new maplibregl.Marker()
```

integreat_cms/static/src/js/pois/poi-map.ts
```diff
@@ -0,0 +1,51 @@
...
+  latitude.addEventListener("focusout",  ()=> showPreview(map));
+
+  longitude.addEventListener("input", ()=> showPreview(map));
+  latitude.addEventListener("input", ()=> showPreview(map));
+
+  document.getElementById("auto-fill-coordinates")?.addEventListener("input", ({ target }) => {
+    if (!(target as HTMLInputElement).checked) {
+      showPreview(map);
+    }
+  });
```
timoludwig, line 21:
```suggestion
```
I'm not sure, do we really need this event listener? Probably it's sufficient that enabling the option causes the reload of the coordinates which dispatches the input event and already updates the map...

integreat_cms/static/src/js/pois/poi-actions.ts
```diff
@@ -15,6 +15,9 @@ window.addEventListener("load", () => {
       // Reset to initial value
       longitude.value = longitude.dataset.initial;
       latitude.value = latitude.dataset.initial;
+
+      longitude.dispatchEvent(new Event("input"));
+      latitude.dispatchEvent(new Event("input"));
```
timoludwig, line 20:
```suggestion
      longitude.dispatchEvent(new Event("focusout"));
      latitude.dispatchEvent(new Event("focusout"));
```
I think we can just dispatch the `focusout` event here - then, we only need one event listener on the other side...

integreat_cms/static/src/js/pois/poi-actions.ts
```diff
@@ -79,6 +82,12 @@ async function autoCompleteAddress() {
     updateField("longitude", data.longitude);
     updateField("latitude", data.latitude);
   }
+
+  let longitude = document.getElementById("id_longitude") as HTMLInputElement;
+  let latitude = document.getElementById("id_latitude") as HTMLInputElement;
+  
+  longitude.dispatchEvent(new Event("input"));
+  latitude.dispatchEvent(new Event("input"));
```
timoludwig, line 90:
```suggestion
  longitude.dispatchEvent(new Event("focusout"));
  latitude.dispatchEvent(new Event("focusout"));
```

timoludwig (changes requested):
Amazing, thanks a lot! :muscle: 
Works exactly as expected, I'd just change a few implementation details...

MizukiTemma:
Yes, the test is fixed 😃 Thank you for your speedy work 👍 

Now somehow build-documentation is failing 😢

timoludwig:
> Now somehow build-documentation is failing cry

It seems that the lxml docs host is a bit unstable at the moment:


```
intersphinx inventory 'https://lxml.de/apidoc/objects.inv' not fetchable due to <class 'requests.exceptions.ReadTimeout'>: HTTPSConnectionPool(host='lxml.de', port=443): Read timed out. (read timeout=5)
```

I'd just re-run the test until it succeeds.

integreat_cms/cms/views/pois/poi_actions.py
```diff
@@ -221,7 +221,7 @@ def auto_complete_address(request, region_slug):
     return JsonResponse(
         data={
             "postcode": address.get("postcode"),
-            "city": address.get("city"),
+            "city": address.get("city") or address.get("town"),
```
timoludwig, line 224:
```suggestion
            "city": address.get("city")
            or address.get("town")
            or address.get("village"),
```
I noticed the response can also contain `village`, e.g. for:
[Hauptstraße 85a
35716 Ewersbach](http://nominatim.maps.tuerantuer.org/nominatim/search?street=Hauptstra%C3%9Fe%2085a&postalcode=35716&city=Ewersbach&addressdetails=1)

timoludwig (changes requested):

ulliholtgrave (approved):
That's a very nice feature :+1:

timoludwig:
@MizukiTemma I'm going to merge this now before further conflicts arise :grin:

Issue #1700, opened 2022-09-19
Provide better error message for matomo timeout


ulliholtgrave:
### Motivation

Right now our system throws an internal server error when our Matomo server has an issue (mostly timeouts) and users are told to contact their administrator. This results in a lot of calls towards the service team, but normally only a reload of the page is required.

### Proposed Solution

Specify specific error message that takes the matomo server into account. For example: "The statistics network is currently experiencing heavy traffic. Please try again in a few minutes."

PR #1776, opened 2022-10-18
Add adjusted error message for statistics page

SteffiStuffel:
### Short description

Adjusted the error message. 

### Resolved issues

Fixes: #1700 

### DIFF
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-10-21 13:13+0000\n"
+"POT-Creation-Date: 2022-10-24 11:48+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: Integreat <info@integreat-app.de>\n"
 "Language-Team: Integreat <info@integreat-app.de>\n"
@@ -1701,7 +1701,7 @@ msgstr "Alle Regionen"
 #: cms/templates/feedback/region_feedback_list.html:35
 #: cms/templates/languagetreenodes/languagetreenode_list.html:24
 #: cms/templates/pages/_page_xliff_import_diff.html:30
-#: cms/templates/statistics/statistics_overview.html:49
+#: cms/templates/statistics/statistics_overview.html:50
 msgid "Language"
 msgstr "Sprache"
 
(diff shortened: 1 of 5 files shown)


timoludwig (commented):
Thanks a lot! :+1:

This is definitely an improvement, I was just wondering whether this is now misleading in the case that something else went wrong and it's not the Matomo timeout causing the error...
Do you think it is worth looking into the different possible exceptions the [MatomoApiManager](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/utils/matomo_api_manager.py) can throw and then differentiating them [here](https://github.com/digitalfabrik/integreat-cms/blob/15ce171cc8041bac2d24823e271632db56eaebf1/integreat_cms/cms/views/statistics/statistics_actions.py#L94-L96) to show two different kind of server error messages in the frontend [here](https://github.com/digitalfabrik/integreat-cms/blob/15ce171cc8041bac2d24823e271632db56eaebf1/integreat_cms/cms/templates/statistics/statistics_overview.html#L19)?
Not sure whether it's worth the effort though.

timoludwig:
@SteffiStuffel sorry for misleading you, I just noticed that the MatomoApiManager does in fact not handle timeout errors gracefully since the exception thrown is an `asyncio.TimeoutError` which is no subclass of `aiohttp.ClientError`.
This means we could e.g. handle the error directly in the view with:
```
    except asyncio.TimeoutError:
        return JsonResponse(
            {"error": "Timeout during request to Matomo API"}, status=504
        )
```
in the views [here](https://github.com/digitalfabrik/integreat-cms/blob/15ce171cc8041bac2d24823e271632db56eaebf1/integreat_cms/cms/views/statistics/statistics_actions.py#L47-L51) and [here](https://github.com/digitalfabrik/integreat-cms/blob/15ce171cc8041bac2d24823e271632db56eaebf1/integreat_cms/cms/views/statistics/statistics_actions.py#L92-L96) (also add `import asyncio`) and then show the corresponding div based on the return status code.

You can test the code by replacing
```
        async with aiohttp.ClientSession() as session:
```
by
```
        timeout = aiohttp.ClientTimeout(total=1)  # 1 sec timeout
        async with aiohttp.ClientSession(timeout=timeout) as session:
```
in the API manager.

SteffiStuffel:
@timoludwig thanks for explanation, adjusted it like you suggested.

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -6186,42 +6188,47 @@ msgstr "Neuen Schlüssel hinzufügen"
 msgid "Number of online accesses over the last 14 days."
 msgstr "Anzahl der Online-Zugriffe der letzten 14 Tage."
 
+#: cms/templates/statistics/_statistics_widget.html:23
+#: cms/templates/statistics/statistics_overview.html:19
+msgid "The statistics network is currently experiencing heavy traffic."
+msgstr "Der Statsitik-Server hat aktuell eine hohe Auslastung."
```
timoludwig, line 6194:
```suggestion
msgstr "Der Statistik-Server hat aktuell eine hohe Auslastung."
```
Small typo

timoludwig (changes requested):
Thanks a lot! :+1:
Just found a small typo.

timoludwig (approved):
Thanks! :ok_hand:

Issue #1749, opened 2022-10-12
Deleting regions not possible if media library has nested structure


timoludwig:
### Describe the Bug

Normally, media files and directories are [cascade](https://github.com/digitalfabrik/integreat-cms/blob/a10d0b86986d3bce9cf0a1e981a44a4e38333eea/integreat_cms/cms/models/media/media_file.py#L135) [deleted](https://github.com/digitalfabrik/integreat-cms/blob/a10d0b86986d3bce9cf0a1e981a44a4e38333eea/integreat_cms/cms/models/media/directory.py#L21) when regions are deleted.
However, this fails if the protection rules between media files and directories prevent the cascade deletion.

### Steps to Reproduce

1. Create nested directory structure with media elements in subfolders
2. Try to delete region
3. See error

### Expected Behavior

The region should be deleted and all media elements and directories with it.
This can either be achieved by deleting them in the correct order (by deleting the sub-folders and elements first) or removing the protected relationship by setting `parent = None` on all directories and `parent_directory = None` on all media files before deleting.

### Actual Behavior

A `ProtectedError` occurs

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
Oct 12 08:25:05 ERROR   django.request - 500 Internal Server Error: /regions/saarbrückenalt/delete/
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/decorators/http.py", line 40, in inner
    return func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/views/regions/region_actions.py", line 46, in delete_region
    deleted_objects = region.delete()
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/db/models/base.py", line 966, in delete
    collector.collect([self], keep_parents=keep_parents)
[...]

PR #1778, opened 2022-10-18
Fix region deletion error if media library has nested structure

seluianova:
### Short description
When region has nested directory structure with media elements in subfolders, attempt to delete such region produces an error:
```
django.db.models.deletion.ProtectedError: ("Cannot delete some instances of model 'Region' because they are referenced through protected foreign keys: 'Director
y.region'.", {<MediaFile (id: 66048, name: arriving.png, path: /var/www/integreat-cms/media/regions/252/2022/06/arriving.png, Landeshauptstadt Saarbrücken (alt)
(&#9888; Versteckt))>})
```

### Proposed changes
Protected relationship is removed by setting parent = None on all directories and parent_directory = None on all media files before deleting the region

### Side effects
Didn't find any

### Resolved issues
Fixes: #1749

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ UNRELEASED
 
 * [ [#1350](https://github.com/digitalfabrik/integreat-cms/issues/1350) ] Various small PDF export improvements
 * [ [#1777](https://github.com/digitalfabrik/integreat-cms/issues/1777) ] Fix autocompleting POI address for non-staff users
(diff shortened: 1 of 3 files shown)


charludo (approved):
Nice. Works like intended, including for deeply nested structures.

MizukiTemma (approved):
Works as expected :) deletion successful, no error, media files/folders are deleted too 👍 Thank you!

timoludwig (approved):
Thanks a lot, works as expected! :muscle:

Issue #1777, opened 2022-10-18
Autocompleting POI address not working for non-staff users


timoludwig:
### Describe the Bug

At the moment, the auto completion feature introduced in #1744 does only work for staff users.

### Steps to Reproduce

1. Login as non staff user
2. Try to autocomplete POI address
3. See error in JS and Django console

### Expected Behavior

The address completion should also work for non staff users.
To solve this, the `region_slug` parameter needs to be passed to the view, even if it's unused.
(Because our [access control](https://github.com/digitalfabrik/integreat-cms/blob/15ce171cc8041bac2d24823e271632db56eaebf1/integreat_cms/core/middleware/access_control_middleware.py#L59-L69) treats all views without the `region_slug` parameter as "staff area" and denies region users to access them.)

### Actual Behavior

A `PermissionDenied` error occurs

### Additional Information

At least the error is not caught in the JS code, so there is no error message shown to users, but the feature just doesn't work.

<details>
  <summary>Traceback</summary>

  ```
Oct 18 06:50:51 WARNING django.request - 403 Forbidden (Permission denied): /ajax/locations/auto-complete-address/
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/core/middleware/access_control_middleware.py", line 67, in __call__
    raise PermissionDenied(
django.core.exceptions.PermissionDenied: <SimpleLazyObject: <User (id: 930, username: maucher, role: Management, region: lkmuenchen)>> does not have the permission to access the staff area
Oct 18 06:50:54 WARNING django.request - 403 Forbidden (Permission denied): /ajax/locations/auto-complete-address/
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/core/middleware/access_control_middleware.py", line 67, in __call__
    raise PermissionDenied(
django.core.exceptions.PermissionDenied: <SimpleLazyObject: <User (id: 930, username: maucher, role: Management, region: lkmuenchen)>> does not have the permission to access the staff area
[...]

PR #1783, opened 2022-10-19
Fix autocompleting POI address for non-staff users

seluianova:
### Short description
At the moment, the auto completion feature introduced in https://github.com/digitalfabrik/integreat-cms/pull/1744 does only work for staff users.

### Proposed changes
Pass region_slug parameter into the view.
(Because our [access control](https://github.com/digitalfabrik/integreat-cms/blob/15ce171cc8041bac2d24823e271632db56eaebf1/integreat_cms/core/middleware/access_control_middleware.py#L59-L69) treats all views without the region_slug parameter as "staff area" and denies region users to access them.)

### Side effects
Didn't find any

### Resolved issues
Fixes: #1777

### DIFF
--- a/integreat_cms/cms/urls/protected.py
+++ b/integreat_cms/cms/urls/protected.py
@@ -429,11 +429,6 @@ urlpatterns = [
             [
                 path("", include(media_ajax_urlpatterns)),
                 path(
-                    "locations/auto-complete-address/",
-                    pois.auto_complete_address,
-                    name="auto_complete_poi_address",
-                ),
-                path(
                     "chat/",
                     include(
                         [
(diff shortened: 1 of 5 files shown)


charludo (approved):
Works like a charm!

integreat_cms/cms/urls/protected.py
```diff
@@ -429,7 +429,7 @@
             [
                 path("", include(media_ajax_urlpatterns)),
                 path(
-                    "locations/auto-complete-address/",
+                    "<region_slug>/locations/auto-complete-address/",
                     pois.auto_complete_address,
                     name="auto_complete_poi_address",
                 ),
```
david-venhoff, line 435: One small improvement for consistency could be to move this down to the other ajax endpoints that require a region slug. It could for example be grouped with `search-poi` at line 626.

david-venhoff (approved):
Thanks, this fixes the issue!

integreat_cms/cms/urls/protected.py
seluianova: Thank you, done!

PR #1784, opened 2022-10-20
Various small PDF export improvements

charludo:
### Short description

Two small improvements:
- fix styling of nested `div`s in `li`s, which now have proper line height and list-style icons.
- use `DejaVu` for LTR texts, so arabic links in latin texts are displayed correctly

### Proposed changes

- fix styling (`ul li div` should display as `inline-block` instead of `block`) and opened PR with [xhtml2pdf/xhtml2pdf](https://github.com/xhtml2pdf/xhtml2pdf/pull/635) to fix this upstream
- replace `Open Sans` with `DejaVu` for LTR PDFs

### Side effects

- none

### Resolved issues

Fixes: #1350 (partly; see discussion there)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 UNRELEASED
 ----------
 
+* [ [#1350](https://github.com/digitalfabrik/integreat-cms/issues/1350) ] Various small PDF export improvements
 
 2022.10.1
 ---------
--- a/integreat_cms/cms/templates/pages/page_pdf.html
+++ b/integreat_cms/cms/templates/pages/page_pdf.html
@@ -28,7 +28,7 @@
                 }
             {% else %}
                 html, body {
-                    font-family: "Open Sans", sans-serif;
+                    font-family: "DejaVu", sans-serif;
                 }
                 pdftoc, h1 {
                     font-family: "Raleway", sans-serif;
--- a/integreat_cms/static/src/css/pdf_page_export.css
+++ b/integreat_cms/static/src/css/pdf_page_export.css
@@ -121,6 +121,9 @@ p, ol, ul {
 li {
     padding: 1px 0px;
 }
+ul li div:first-child {
+    display: inline-block;
+}
 .right-to-left {
     text-align: right;
 }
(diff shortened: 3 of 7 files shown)


ulliholtgrave (approved):
The newly generated PDFs look better and better with each of your PRs :)

charludo:
> The newly generated PDFs look better and better with each of your PRs :)

Haha, thanks. Though it feels kinda weird that all of the fixes so far have been a single line of CSS 😆

integreat_cms/cms/utils/pdf_utils.py
```diff
@@ -195,3 +196,25 @@ def link_callback(uri, rel):
...
+    """
+    # keep the first occurrence, since it's intentionally
+    # set in page_pdf.html
+    html_parts = html.partition("font-family: ")
+    html = (
+        html_parts[0]
+        + html_parts[1]
+        + re.sub(r"font-[a-zA-Z]+:.*?;", "", html_parts[2])
+    )
+    return html
```
timoludwig, line 220: In my small test this worked as expected, but I'm not sure whether it's robust enough against invalid HTML (which we unfortunately might have in our content), e.g. if the semicolon is missing after the font definition, some more content than desired might be removed etc...
```suggestion
    content = fromstring(html)
    for element in content.iter():
        # Remove all inline font style definitions
        style = element.attrib.pop("style", None)
        if style:
            element.attrib["style"] = re.sub(r"font-[a-zA-Z]+:.*?;", "", style)
    return tostring(content, with_tail=False).decode("utf-8")
```
(This requires `from lxml.html import fromstring, tostring`)

timoludwig (changes requested):
Thanks a lot, looks better now! :+1:
I just have a few concerns that the regex might be a bit too error-prone if the HTML isn't completely valid.

integreat_cms/cms/utils/pdf_utils.py
charludo: Yeah, you are absolutely right. Your suggestion is a lot safer, and does away with partitioning out the wanted styling.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 UNRELEASED
 ----------
 
+* [ [#1350](https://github.com/digitalfabrik/integreat-cms/issues/1350) ] Various small PDF export improvements
```
```suggestion


```
The first entry always needs to add an additional new line :see_no_evil:

timoludwig (changes requested):

timoludwig (approved):
Thanks! :pray:

Issue #170, opened 2019-06-29
Meta: :iphone: Responsive Design


timobrembeck:
Make the CMS responsive with at least these features:

- [x] #1816 
- [x] #1815
- [x] #1907
- [ ] Possibility to change basic content and settings (publish/hide pages etc.)

If doable without too much effort, we also could support all features of the CMS in mobile-friendly views. This would be a cool step towards future-proof technology and digitalization of local minicipalities.

osmers:
Closing as there are no more related issues so far - future issues can get the label topic:responsive-design

PR #1785, opened 2022-10-24
Add mobile menu

JoeyStk:
### Short description
I reduced this PR to only adding a mobile menu. I think we should make #170 a Meta Issue and split it up in separate issues. 

### Proposed changes
- On screens smaller than lg the navigation disappears to the side and instead a mobile menu appears
- A click on the mobile menu is supposed to open the navigation bar
- Disband grid system for mobile menus in event and POIs and add some spacing to the top
- Remove topbar for mobile screens, since this is only for developers and we don't expect developer to do their work on a mobile screen.

Thank you a lot in advance! I'm very much looking forward to your feedback :) 

### Side effects
Couldn't find any yet.

### Resolved issues
Parts of #1816 


timoludwig (commented):
Since #1787 is now merged, can you rebase your branch onto the develop branch and check whether this fixes the tests?

david-venhoff (commented):
Thanks, this looks really cool! :+1:
The conditional grid styling works great, probably the Imprint view should have it too.

Regarding the topbar we can probably just not display the text (and only the icon), e.g. for the region selector:
```html
<div class="hidden lg:inline">
    {{ request.region.full_name }}
</div>
```

Also I noticed that it is not really possible to expand the region selector or user selector in the topbar, since tapping on it causes the browser to follow the link instead. In this case, the link could probably just be removed since it is not required, but this would have to be done for all dropdowns (e.g. langauge selectors for pages, pois, events) so maybe there is a better solution.



Probably the best solution for the _form_language_tabs is to render them in javascript dynamically. They also seem to have a problem on small screens:
grafik

integreat_cms/cms/templates/_base.html
```diff
@@ -67,7 +67,7 @@
             </div>
         </div>
         <div id="user-info"
-             class="relative px-2 text-gray-800 flex flex-col justify-center cursor-pointer hover:bg-gray-200">
+             class="hidden relative px-2 text-gray-800 lg:flex flex-col justify-center cursor-pointer hover:bg-gray-200">
```
timoludwig: Just hiding this removes two essential functions: Accessing the user settings and logout.
I like @david-venhoff's idea of hiding some of the text and removing the link to make it possible to tap on the icon to expand the menu.

integreat_cms/cms/templates/_base.html
```diff
@@ -29,7 +29,7 @@
             {% trans "Documentation (Wiki)" %}
         </a>
         <div id="region-selector"
-             class="relative flex flex-col justify-center px-2 cursor-pointer text-gray-800 hover:bg-gray-200 border-r border-gray-400">
+             class="hidden relative lg:flex flex-col justify-center px-2 cursor-pointer text-gray-800 hover:bg-gray-200 border-r border-gray-400">
```
timoludwig: Here as well - we need at least some kind of region selector, otherwise it's not possible to navigate back to the network area from within a region.
(This is not only relevant for developers, but also for e.g. the service team members, who maybe also want to check a few things on the go)...

integreat_cms/static/src/index.ts
```diff
@@ -76,6 +76,8 @@ import "./js/unsaved-warning";
 
 import "./js/pois/poi-actions.ts";
 
+import "./js/dashboard.ts";
```
timoludwig:
```suggestion
import "./js/menu.ts";
```
Dashboard doesn't really make sense when the menu is visible on every page, not only the dashboard.

timoludwig (approved):
Thanks a lot! :+1:
Most of this looks good to me, just a few details were missing and I think we cannot hide all functions from the top menu on small screens.
Since my suggestions for refactoring the menu were out of range of the GitHub suggestion feature, I directly pushed to your branch, I tried to explain my changes below:

integreat_cms/static/src/index.ts
JoeyStk: Yes, I agree :) Thank you!

integreat_cms/cms/templates/_base.html
```diff
@@ -88,9 +88,13 @@
                 </a>
             </div>
         </div>
+        <div id="mobile-menu"
+             class="flex flex-col justify-center mr-10 px-2 lg:hidden border-l border-gray-400">
```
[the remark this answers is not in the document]
JoeyStk: Oh yes, that makes sense!

integreat_cms/cms/templates/_base.html
JoeyStk: I understand :)

integreat_cms/static/src/js/dashboard.ts
```diff
@@ -0,0 +1,8 @@
+document.addEventListener("DOMContentLoaded", function () {
+    const mobile = document.querySelector("#mobile-menu");
+    const sidebar = document.querySelector("#primary-navigation");
+
+    mobile.addEventListener("click", () => {
+        sidebar.classList.toggle("-translate-x-full");
+    });
+});
```
david-venhoff, line 8: I think this file is not required anymore

david-venhoff (approved):
Thanks, apart from the duplicate file this looks good to me!

integreat_cms/static/src/js/dashboard.ts
JoeyStk: Oh yes, thank you!

PR #1795, opened 2022-10-25
Reduce CircleCI resource classes to save credits

timoludwig:
### Short description

Our credit usage has nearly doubled in the current month:

Screenshot 2022-10-25 at 12-56-22 Insights

So maybe it's time to reduce CircleCI resource classed to save credits

Reducing the `resource_class` of the `build-documentation` did not work:

```
Exception occurred:
  File "/home/circleci/.pyenv/versions/3.9.14/lib/python3.9/multiprocessing/connection.py", line 388, in _recv
    raise EOFError
```
... so somehow the job needs the cores?

### Side effects

Increased workflow duration, not sure how much, maybe 1-2 mins?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,6 +40,7 @@ jobs:
   black:
     docker:
       - image: cimg/python:3.9
+    resource_class: small
     steps:
       - checkout
       - attach_workspace:
@@ -50,7 +51,6 @@ jobs:
   pylint:
     docker:
       - image: cimg/python:3.9
-    resource_class: xlarge
     steps:
       - checkout
       - attach_workspace:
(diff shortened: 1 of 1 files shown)


ulliholtgrave (approved):
LGTM

svenseeberg (approved):
Let's try. If we feel it takes too long to finish the CI jobs, we can discuss again? As far as I understand we're now falling back to the default setting which is medium, right? That means 2 instead of 4 cores.

timoludwig:
> Let's try. If we feel it takes too long to finish the CI jobs, we can discuss again? As far as I understand we're now falling back to the default setting which is medium, right? That means 2 instead of 4 cores.

Ahh, yes, good point, we could also use the `resource_class: small` for quite a few jobs that don't need two cores.

Issue #1333, opened 2022-04-01
Mark external links, phone numbers, etc.


svenseeberg:
### Motivation
We want to mark external links, phone numbers, etc with decicated icons to make it easier for users to understand what clicking on them will do:
- Will it open an external / third party website or the phone app
- Users should know if a resource is available offline in the app
- Increased trust

### Proposed Solution
When inserting a link with TinyMCE, external, phone or e-mail links (not pointing to the CMS domain) should get an additional CSS class. This class can then be used by the app to render an additional icon in front of the link.

### Alternatives
Don't.

### Additional Context
The app team needs to be informed when this issue is closed: https://issues.tuerantuer.org/browse/IGAPP-973

PR #1796, opened 2022-10-25
Mark external links with special class

charludo:
### Short description
This PR adds the class `link-external` to links inserted or updated with TinyMCE.

### Proposed changes

- check if link is external (i.e. does not contain `integreat.app`), if so, add extra class when inserting
- re-check when updating the link, both through the main plugin modal and the inline popup

### Side effects

- none

### Resolved issues

Fixes: #1333


svenseeberg (changes requested):
The basic JS code works, thanks! However, we should mark more domains as supported.

integreat_cms/static/src/js/tinymce-plugins/custom_link_input/plugin.js
```diff
@@ -36,6 +36,10 @@ import { getCsrfToken } from "../../utils/csrf-token";
     return url;
   }
 
+  function isExternalUrl(url) {
+    return !url.includes("integreat.app");
```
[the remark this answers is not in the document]
charludo: No, that sounds reasonable for this issue/PR imo.
Though I'm a bit unsure about how to pass the domains from django to webpack. Is there an example in the project where that's already happening?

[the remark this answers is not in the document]
timoludwig: When we want to pass variables from python to JS, we typically do one of to things:
1. create a div with data attributes, e.g. like in [_tinymce_config.html](https://github.com/digitalfabrik/integreat-cms/blob/c87c51f80943dfc58cc249f698d85dfae1a1ad5b/integreat_cms/cms/templates/_tinymce_config.html#L17)
2. pass the data via [`json_script`](https://docs.djangoproject.com/en/4.1/ref/templates/builtins/#json-script), e..g like in [media_list.html](https://github.com/digitalfabrik/integreat-cms/blob/c87c51f80943dfc58cc249f698d85dfae1a1ad5b/integreat_cms/cms/templates/media/media_list.html#L7)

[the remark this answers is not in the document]
charludo: Thanks, that was very helpful!

integreat_cms/core/settings.py
```diff
@@ -114,6 +114,11 @@
     os.environ.get("INTEGREAT_CMS_DEFAULT_REQUEST_TIMEOUT", 10)
 )
 
+#: The URLs which are treated as internal in TinyMCE custom link plugin
+CUSTOM_INTERNAL_URLS = os.environ.get("INTEGREAT_INTERNAL_URLS", "").split()
+INTERNAL_URLS = " ".join(
+    url for url in ["integreat.app", ".integreat-app.de"] + CUSTOM_INTERNAL_URLS if url
+)
```
timoludwig, line 121:
```suggestion
#: The URLs which are treated as internal in TinyMCE custom link plugin
INTERNAL_URLS = [BASE_URL, WEBAPP_URL, WEBSITE_URL] + list(
    filter(
        None,
        (
            x.strip()
            for x in os.environ.get("INTEGREAT_CMS_INTERNAL_URLS", "").splitlines()
        ),
    )
)
```
1. I wouldn't hardcode "integreat.app" and "integreat-app.de" here since we use the same code for whitelabeled versions like the malte app and the aschaffenburg app. Instead, we can rely on the existing settings for these domains. Also, this requires to move either this setting down or the `BASE_URL` up so it can be used here.
2. To stay consistent with our other settings which contain lists (e.g. [`ALLOWED_HOSTS`](https://github.com/digitalfabrik/integreat-cms/blob/c87c51f80943dfc58cc249f698d85dfae1a1ad5b/integreat_cms/core/settings.py#L336-L346) or [`LANGUAGES`](https://github.com/digitalfabrik/integreat-cms/blob/c87c51f80943dfc58cc249f698d85dfae1a1ad5b/integreat_cms/core/settings.py#L649-L662)), I would split by newline characters instead of spaces, even if this makes the code a bit less readable. This makes it possible to write e.g.
    ```
    INTERNAL_URLS =
	    internal.host.org
	    my.app.de
    ```
into the config file. Speaking of which, could you add an example of this to [example-configs/integreat-cms.ini](https://github.com/digitalfabrik/integreat-cms/blob/develop/example-configs/integreat-cms.ini)?

integreat_cms/cms/templatetags/settings_tags.py
```diff
@@ -28,3 +28,15 @@ def get_base_url():
...
+@register.simple_tag
+def get_internal_urls():
+    """
+    This tag returns the stringified list of url names
+    which should be treated as internal
+
+    :return: stringified list of internal url names
+    :rtype: str
+    """
+    return settings.INTERNAL_URLS
```
timoludwig, line 42:
```suggestion
    return " ".join(settings.INTERNAL_URLS)
```
When used in a data attribute, I guess it's fine to split by spaces.

timoludwig (changes requested):

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 UNRELEASED
 ----------
 
+* [ [#1333](https://github.com/digitalfabrik/integreat-cms/issues/1333) ] Mark external links with special class
```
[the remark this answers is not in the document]
charludo: Since that's the second time it's come up: 2 empty lines below the last entry, 1 empty line above the first entry?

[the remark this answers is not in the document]
timoludwig: Yes, correct! :+1:
And if you have an idea how to improve the [parsing command](https://github.com/digitalfabrik/integreat-cms/blob/c87c51f80943dfc58cc249f698d85dfae1a1ad5b/.circleci/config.yml#L412) to make the format more flexible, let us know :sweat_smile:

integreat_cms/core/settings.py
timoludwig: Good point, this looks way more readable! :+1:
Could you change the syntax of the other list settings as well? :wink:

charludo: Please have a look if you're happy with this. Unfortunately `BASE_URL` and so forth all come with an attached `https://`, but we need to not only capture `https://integreat-app.de` but also (for example) `cms.integreat-app.de` as @svenseeberg has already remarked, so I had to do some slight post-processing.

timoludwig: Ah okay, I thought the `BASE_URL` would already account for the CMS url. But yes, marking all subdomains as internal as well makes sense I guess.

integreat_cms/core/settings.py
```diff
@@ -888,6 +883,16 @@
...
 
 
+#################
+# INTERNAL URLS #
+#################
+
+#: The URLs which are treated as internal in TinyMCE custom link plugin
+INTERNAL_URLS = [BASE_URL, WEBAPP_URL, WEBSITE_URL] + [
+    x.strip(x) for x in os.environ.get("INTEGREAT_CMS_INTERNAL_URLS", "").splitlines()
+]
```
timoludwig, line 893:
```suggestion
INTERNAL_URLS = (
    ALLOWED_HOSTS
    + [WEBAPP_URL, WEBSITE_URL]
    + [
        x.strip(x)
        for x in os.environ.get("INTEGREAT_CMS_INTERNAL_URLS", "").splitlines()
    ]
)
```
Oh, I just noticed we sometimes have [multiple CMS domains](https://git.tuerantuer.org/DF/salt/src/commit/80f38189a906116872590cdc733683e47e95d913/pillars/letsencrypt/org/tuerantuer/malte-prod.sls#L4-L6) we add to the allowed hosts, probably all of them should be implicitly be marked as internal and `BASE_URL` isn't enough...

integreat_cms/core/settings.py
```diff
@@ -888,6 +883,21 @@
...
+#################
+# INTERNAL URLS #
+#################
+
+#: The URLs which are treated as internal in TinyMCE custom link plugin
+INTERNAL_URLS = (
+    ALLOWED_HOSTS
+    + [WEBAPP_URL, WEBSITE_URL]
+    + [
+        x.strip(x)
```
timoludwig, line 895:
```suggestion
        x.strip()
```
Oh, `x.strip(x)` will always return an empty string.

integreat_cms/core/settings.py
```diff
@@ -888,6 +883,21 @@
...
+        for x in os.environ.get("INTEGREAT_CMS_INTERNAL_URLS", "").splitlines()
+    ]
```
timoludwig, line 897:
```suggestion
        if x
    ]
```
Ah, and sorry for not noticing this earlier: I think the `if x` is required because configparser returns a trailing new line for multi-line config values.

timoludwig (changes requested):

integreat_cms/core/settings.py
charludo: Whoops! Good catch, also didn't see that.

integreat_cms/core/settings.py
charludo: Yup! Also noticed that. Sorry for not noticing it in the last suggestion.

I'll have another thorough look and then squash the history.

timoludwig (approved):
Thanks! :pray:
Now, could you squash your commits again? :see_no_evil:

svenseeberg (approved):
Looks good to me, thanks.

Issue #1746, opened 2022-10-11
Change User right for role Terminplaner


nassabay:
### Motivation
user with the role Terminplaner should not be able to see any part of the analysis section.

### Proposed Solution
change user right for role Terminplaner to only see event and media library

timoludwig:
@digitalfabrik/service-team does this also apply to the roles "author" (users who can submit pages for review) and "editor" (users who can publish pages but nothing else)?

osmers:
Ok, finally we discussed this - event manager and author role should not see any analytics
Editor role can see broken links but nothing else in Analytics

osmers:
So editor role can stay the way it is afaict

PR #1797, opened 2022-10-25
Hide analytics for author and event manager

MizukiTemma:
### Short description

Availability of analysis menu will be changed as below:
      -      Management: statistics (if enabled), translation report, feedback, broken links
      -      Editor: translation report, broken links
      -      Autor: none
      -      Event manager (Terminplaner): none

### Proposed changes

- create new permissions "view_translation_report" etc by introducing a dummy model "PermissionManager" (because there is no related model except feedback).
- use permission check to show/hide analytics menus.

### Side effects

- so far not found

### Resolved issues

Fixes: #1746 


integreat_cms/cms/models/permission_manager/permission_manager.py
```diff
@@ -0,0 +1,17 @@
+from django.db import models
+
+
+class PermissionManager(models.Model):
```
ulliholtgrave, line 4: I guess I am fine with introducing some dummy model for additional permissions, but I would prefer the name `CustomPermissionManager`, otherwise it's kinda suggesting that it takes care of everything.

Could you also change the directory name into `custom_permissions` to match the naming?

integreat_cms/cms/templates/_base.html
```diff
@@ -105,33 +105,39 @@
...
-                    <a href="{% url 'region_feedback' region_slug=request.region.slug %}"
-                       class="{% if current_menu_item == 'region_feedback' %} active{% endif %}">
-                        <i icon-name="thumbs-up"></i>
-                        {% trans "Feedback" %}
-                    </a>
+                {% if perms.cms.view_analytics %}
+                    <div class="min-h-3 pl-2">
+                        <p class="font-bold">{% trans "Analytics" %}</p>
+                    </div>
+                    {% if request.region.statistics_enabled and perms.cms.view_statics %}
```
ulliholtgrave, line 112:
```suggestion
                    {% if request.region.statistics_enabled and perms.cms.view_statistics %}
```

integreat_cms/cms/models/permission_manager/permission_manager.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+    Dummy class to implement permissions that are not related to models
```
ulliholtgrave, line 6:
```suggestion
    Custom manager that introduces permissions that are not related to real models
```

ulliholtgrave (changes requested):
Thanks, I guess your approach does the trick, but I found some small issues that should be addressed

integreat_cms/cms/models/permission_manager/permission_manager.py
MizukiTemma: Done 👍

integreat_cms/cms/models/permission_manager/permission_manager.py
MizukiTemma: Also done :)

integreat_cms/cms/templates/_base.html
MizukiTemma: done

integreat_cms/cms/models/custom_permissions/custom_permission.py
```diff
@@ -0,0 +1,17 @@
+from django.db import models
+
+
+class CustomPermissionManager(models.Model):
```
timoludwig, line 4: This is mostly personal preference I guess, but I'm not really sure about creating a separate model for this - what do you think about adding the permissions to the user model instead?

integreat_cms/cms/templates/_base.html
```diff
@@ -105,33 +105,39 @@
...
-                    <i icon-name="check-circle"></i>
-                    {% trans "Translation Report" %}
-                </a>
-                {% if perms.cms.view_feedback %}
-                    <a href="{% url 'region_feedback' region_slug=request.region.slug %}"
-                       class="{% if current_menu_item == 'region_feedback' %} active{% endif %}">
-                        <i icon-name="thumbs-up"></i>
-                        {% trans "Feedback" %}
-                    </a>
+                {% if perms.cms.view_analytics %}
```
timoludwig, line 108:
```suggestion
                {% if request.region.statistics_enabled and perms.cms.view_statistics or perms.cms.view_translation_report or perms.cms.view_feedback or perms.cms.view_broken_links %}
```
I don't think the `view_analytics` permission makes sense on its own - we should instead just show the analytics section whenever the user is allowed to see at least one part of it.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ UNRELEASED
 * [ [#1170](https://github.com/digitalfabrik/integreat-cms/issues/1170) ] Add map preview on POI form
 * [ [#1579](https://github.com/digitalfabrik/integreat-cms/issues/1579) ] Fix auto-filling of coordinates for multiple street numbers
 * [ [#1767](https://github.com/digitalfabrik/integreat-cms/issues/1767) ] Revert statistics calculation to original & add online downloads column
+* [ [#1746](https://github.com/digitalfabrik/integreat-cms/issues/1746) ] Hide analytics section (partially) for author, editor and terminplaner
```
timoludwig, line 22:
```suggestion
* [ [#1746](https://github.com/digitalfabrik/integreat-cms/issues/1746) ] Hide analytics section (partially) for author, editor and event manager
```

timoludwig (changes requested):
Thanks, this gets the job done :muscle:

integreat_cms/cms/models/custom_permissions/custom_permission.py
ulliholtgrave: That would also be fine for me 👍

MizukiTemma: Moved to the user model 👍

integreat_cms/cms/templates/_base.html
MizukiTemma: Done.

CHANGELOG.md
MizukiTemma: Modified. I moved this entry to UNRELREASED.

integreat_cms/cms/constants/roles.py
```diff
@@ -67,6 +67,9 @@
 #: The permissions of the editor role
 EDITOR_PERMISSIONS = AUTHOR_PERMISSIONS + [
     "publish_page",
+    "view_analytics",
```
timoludwig, line 70:
```suggestion
```

integreat_cms/cms/migrations/0044_alter_user_options.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 3.2.16 on 2022-11-06 10:53
+
+from django.db import migrations
+
+
```
timoludwig, line 5:
```suggestion

# pylint: disable=unused-argument
def update_roles(apps, schema_editor):
    """
    Update analytics permissions

    :param apps: The configuration of installed applications
    :type apps: ~django.apps.registry.Apps

    :param schema_editor: The database abstraction layer that creates actual SQL code
    :type schema_editor: ~django.db.backends.base.schema.BaseDatabaseSchemaEditor
    """
    Group = apps.get_model("auth", "Group")
    Permission = apps.get_model("auth", "Permission")


    # Emit post-migrate signal to make sure the Permission objects are created before they can be assigned
    emit_post_migrate_signal(2, False, "default")


    # Clear and update permissions according to new constants
    for role_name in dict(roles.CHOICES):
        group = Group.objects.filter(name=role_name).first()
        # Clear permissions
        group.permissions.clear()
        # Set permissions
        group.permissions.add(
            *Permission.objects.filter(codename__in=roles.PERMISSIONS[role_name])
        )


```
Just changing the constants won't change the permissions on already deployed systems. You always have to re-assign the roles afterwards via a migration.

integreat_cms/cms/migrations/0044_alter_user_options.py
```diff
@@ -0,0 +1,30 @@
...
+                "ordering": ["username"],
+                "permissions": (
+                    ("view_translation_report", "view_translation_report"),
+                    ("view_broken_links", "view_broken_links"),
+                    ("view_statistics", "view_statistics"),
+                ),
+                "verbose_name": "user",
+                "verbose_name_plural": "users",
+            },
+        ),
```
timoludwig, line 29:
```suggestion
        ),
        migrations.RunPython(update_roles, migrations.RunPython.noop),
```

integreat_cms/cms/migrations/0044_alter_user_options.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 3.2.16 on 2022-11-06 10:53
+
```
timoludwig, line 2:
```suggestion

from django.core.management.sql import emit_post_migrate_signal
```

integreat_cms/cms/models/users/user.py
```diff
@@ -219,3 +219,9 @@ class Meta:
         default_permissions = ("change", "delete", "view")
         #: The fields which are used to sort the returned objects of a QuerySet
         ordering = ["username"]
+        #: Permissions that are not related to real models
+        permissions = (
+            ("view_translation_report", "view_translation_report"),
+            ("view_broken_links", "view_broken_links"),
+            ("view_statistics", "view_statistics"),
```
timoludwig, line 226: I mean that's open for debate, but if we're using permission objects to hide menu items, we could also treat them like "real" permissions and also deny access to the analytics views itself... otherwise, users can still view these pages by typing in the link directly... (in theory, this shouldn't be a problem, I'm just wondering if we should take these measures to have a consistent behaviour when it comes to permissions.)

timoludwig (changes requested):

integreat_cms/cms/migrations/0044_alter_user_options.py
MizukiTemma: oh right 😅 I've totally forgotten it.

integreat_cms/cms/migrations/0044_alter_user_options.py
```diff
@@ -0,0 +1,62 @@
...
+from django.core.management.sql import emit_post_migrate_signal
+from django.db import migrations
+
+
```
timoludwig, line 6:
```suggestion
from ..constants import roles


```
Sorry, forgot this import

MizukiTemma: No problem, I added it :)

timoludwig (approved):
Thanks! :+1:

integreat_cms/cms/models/users/user.py
MizukiTemma: I think it's anyway nice to have, so implemented 🙂

ulliholtgrave (approved):
Thanks, After you solve the last conflicts this seems good to go :+1:

Issue #1732, opened 2022-10-05
CircleCI cache of virtual environment outdated


timoludwig:
### Describe the Bug

After a specific time, the CircleCI cache does not work the way we expect it to.

### Steps to Reproduce

1. Do not update the Pipfile.lock for a while
2. See `FileNotFoundError`s in CircleCI

### Expected Behavior

When CircleCI's cache is invalid, it should not be used.

### Actual Behavior

Somehow, CircleCI thinks the cache exists, but the restored files do not contain the required executables.
The check here:
https://github.com/digitalfabrik/integreat-cms/blob/947d227efc57b52d3a2269bfe0e40d8e36248c16/.circleci/config.yml#L18
returns `true`, so CirleCI does restore something (the directory `.venv`), but it does not restore its contents.

### Additional Information

Maybe it's already sufficient to change 
```
if [[ -d ".venv" ]]; then
```
```
if [[ -x ".venv/bin/integreat-cms-cli" ]]; then
```
Edit: No, this does not help, see 1ae628d41a85738ebc4758c1a287a5272b4e4dbf and [this build](https://app.circleci.com/pipelines/github/digitalfabrik/integreat-cms/5509/workflows/50614f95-d8f5-4c64-9921-c59f7e5e0438/jobs/48429).

<details>
  <summary>Traceback</summary>

  ```
Traceback (most recent call last):
  File "/home/circleci/.pyenv/versions/3.9.14/lib/python3.9/site-packages/pipenv/core.py", line 2627, in _launch_windows_subprocess
    return subprocess.Popen([command] + script.args, **options)
  File "/home/circleci/.pyenv/versions/3.9.14/lib/python3.9/subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/home/circleci/.pyenv/versions/3.9.14/lib/python3.9/subprocess.py", line 1821, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: '/home/circleci/project/.venv/bin/integreat-cms-cli'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/circleci/.pyenv/versions/3.9.14/bin/pipenv", line 8, in <module>
    sys.exit(cli())
  File "/home/circleci/.pyenv/versions/3.9.14/lib/python3.9/site-packages/pipenv/vendor/click/core.py", line 1128, in __call__
[...]

PR #1798, opened 2022-10-25
Fix CircleCI cache invalidation

timoludwig:
### Short description

Ok, this was another crazy rabbit hole I got myself into :laughing: 

Because of another error caused by an updated CircleCI docker image (see https://github.com/CircleCI-Public/cimg-python/issues/161) I noticed that the CircleCI cache is broken exactly at the same time a new docker image version is released:

- version `3.9.15` released today caused [failing build](https://app.circleci.com/pipelines/github/digitalfabrik/integreat-cms/5923/workflows/aa940c9a-0eba-47c1-ad2f-41ba3eecd3e4) for 3b8744ff25bef11e0f38599aae6e96513d736b1a
- version `3.9.14` released on Oct 4 caused [failing build](https://app.circleci.com/pipelines/github/digitalfabrik/integreat-cms/5501/workflows/08ba3c5b-eb04-4c32-97ee-1a0b19e31435) for d23ed936708102bb56dc2fa58df76a377379207b
- version `3.9.13` released on May 28 caused [failing build](https://app.circleci.com/pipelines/github/digitalfabrik/integreat-cms/4648/workflows/27185013-fe0c-482a-bf7f-8ed3f352c414) for 57a5aa6b4792538a8cb03170dc165378f85d7f14
- etc.

So I investigated a bit further and noticed that the CircleCI virtual environment uses files which depend on the specific python version, e.g.:

```
  File "/home/circleci/.pyenv/versions/3.9.14/lib/python3.9/site-packages/pipenv/core.py", line 2627, in _launch_windows_subprocess
    return subprocess.Popen([command] + script.args, **options)
```

So the cached virtual environment was for python version `3.9.14`, but the new docker image had version `3.9.15` and could not use the restored venv.
So to solve this, we have to pin the exact python version and remember to increment the cache key version number each time we upgrade the docker image.

### Proposed changes

- Update the pip dependencies once again
- Downgrade CircleCI docker image to python 3.9.14

### Side effects

- We won't use the "bleeding edge" docker image anymore and have to keep in mind to upgrade it manually in the future

### Resolved issues

Fixes: #1732

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,13 +5,13 @@ orbs:
 jobs:
   pipenv-install:
     docker:
-      - image: cimg/python:3.9
+      - image: cimg/python:3.9.14
     environment:
       PIPENV_VENV_IN_PROJECT: true
     steps:
       - checkout
       - restore_cache:
-          key: pipenv-{{ checksum "Pipfile.lock" }}-v2
+          key: pipenv-{{ checksum "Pipfile.lock" }}-v1
       - run:
           name: Install pip dependencies
           command: |
@@ -26,7 +26,7 @@ jobs:
               fi
             fi
       - save_cache:
-          key: pipenv-{{ checksum "Pipfile.lock" }}-v2
+          key: pipenv-{{ checksum "Pipfile.lock" }}-v1
           paths:
             - .venv
             - integreat_cms.egg-info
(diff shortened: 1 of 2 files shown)


svenseeberg (approved):
The fact that the tests work speaks for itself. ;-)

ulliholtgrave (approved):
Let's try :+1:

Issue #1800, opened 2022-10-26
Exclude archived pages from PDF exports


charludo:
### Describe the Bug
Currently, archived pages are still included in PDF exports, even after updating a sibling page to enforce re-export of the PDF.

### Steps to Reproduce

1. export a parent page with children
2. visit the CMS, mark one of these child pages as archived
3. edit and update one of its siblings to ensure the PDF is being regenerated
4. export parent page again
5. archived page is still included in newly generated PDF

### Expected Behavior
Archived pages are never included in PDF exports

### Actual Behavior
Archived pages are included in PDF export of parent

PR #1801, opened 2022-10-26
Exclude archived pages from PDF exports

charludo:
### Short description
Exclude archived pages from PDF exports

### Proposed changes

- check if paghe is archived before adding it to list of to-be-exported pages

### Side effects

- none

### Resolved issues

Fixes: #1800


svenseeberg:
Looks good IMHO and we should merge this sooner than later. I have not looked into PDF tests yet, but if we could add some sort of test for this it would probably be a very good idea.

timoludwig (approved):
Thanks a lot for your speedy fix! :tada:

For context about the failing build, see #1798. After merging that and rebasing this, the tests should hopefully succeed.

svenseeberg (approved):

integreat_cms/cms/utils/pdf_utils.py
```diff
@@ -56,7 +56,7 @@ def generate_pdf(region, language_slug, pages):
     for page in pages:
         # add translation id and last_updated to hash key list if they exist
         page_translation = page.get_public_translation(language_slug)
-        if page_translation:
+        if page_translation and not page.archived:
```
timoludwig, line 59: I have a few concerns about performance, if a page has very many archived children this causes a lot of additional database queries, but let's merge it first and then see whether it is actually a problem.

charludo: Did not realize this, sorry. Do you have an alternative suggestion?

timoludwig: An alternative solution would be to call `.cache_tree(archived=False)` on the queryset - but then it's not a queryset anymore, but a list, and we would have to perform the other calculations like `aggregate` in python... so a bit more effort, not sure whether it's justified, so let's try the simple approach first.

charludo:
> Looks good IMHO and we should merge this sooner than later. I have not looked into PDF tests yet, but if we could add some sort of test for this it would probably be a very good idea.

Agreed, but I don't see a quick way of doing this, at least not today.

I could add an archived page to an existing test, but not sure how meaningful that would be? Is it OK with you to if we put this off for now?

timoludwig:
> I could add an archived page to an existing test, but not sure how meaningful that would be? Is it OK with you to if we put this off for now?

I'd say the change is trivial enough that we don't need a test now - we can just keep it in mind for the future.

Issue #1793, opened 2022-10-25
Sending feedback for recurring events not working


timoludwig:
### Describe the Bug

The feedback endpoint only accepts event slugs which really exist in the database, so it fails for all recurring events where we generate a bunch of date-specific slugs.

### Steps to Reproduce

1. Create recurring event
2. Retrieve event API endpoint to see which slug the recurring event now has
3. Try to submit feedback about the event with this specific slug

### Expected Behavior

The feedback should be submitted

### Actual Behavior

A 404 error occurs: `No matching event found for slug.`

### Additional Information

This is similar to https://github.com/digitalfabrik/integreat-cms/issues/1718, but only from the concept, the solution needs to be different.

seluianova:
Do we need to validate a date-part of the slug? Check that it corresponds to the event?
Another option would be to simply remove the date from the slug, before we retrieve the event from DB. Does it sound okay?

timoludwig:
> Another option would be to simply remove the date from the slug, before we retrieve the event from DB. Does it sound okay?

Yes, I think removing the date string is perfectly fine, probably only relevant for edge cases to know which exact occurrence of a recurring event the feedback refers to.

PR #1810, opened 2022-10-26
Fix sending feedback for recurring events

seluianova:
### Short description
Slugs for recurring events contain date (in format: base-slug$yyyy-mm-dd).
As a result, when we receive a feedback for recurring event, event can't found by slug in DB and response 404 is returned.

### Proposed changes
Remove the date from the slug during feedback processing

### Side effects
- If someone manually creates event with the slug ending with the date in format $yyyy-mm-dd, feedback for such event won't be processed (upd: actually not possible, symbol $ is not allowed in the slug)
- There is no validation of the date in slug. If we receive feedback for an existing event but with an nonexistent date, it will be processed successfully.
- We lose information about which date the feedback belongs to.

### Resolved issues
Fixes: #1793

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ UNRELEASED
 ----------
 
 * [ [#1718](https://github.com/digitalfabrik/integreat-cms/issues/1718) ] Enable submitting feedback about fallback translations of events and pois
+* [ [#1793](https://github.com/digitalfabrik/integreat-cms/issues/1793) ] Fix sending feedback for recurring events
 
 
 2022.10.2
--- a/integreat_cms/api/v3/feedback/event_feedback.py
+++ b/integreat_cms/api/v3/feedback/event_feedback.py
@@ -1,4 +1,5 @@
 import logging
+import re
 
 from django.http import JsonResponse, Http404
 
@@ -67,10 +68,11 @@ def event_feedback_internal(data, region, language, comment, rating, is_technica
     :return: JSON object according to APIv3 single page feedback endpoint definition
     :rtype: ~django.http.JsonResponse
     """
-    event_translation_slug = data.get("slug")
+    # Remove date from the slug for recurrung events
+    event_translation_slug = re.sub(r"\$\d{4}-\d{2}-\d{2}$", "", data.get("slug"))
 
     events = region.events.filter(
-        translations__slug=data.get("slug"),
+        translations__slug=event_translation_slug,
         translations__language=language,
     ).distinct()
 


integreat_cms/api/v3/feedback/event_feedback.py
```diff
@@ -67,10 +68,11 @@ def event_feedback_internal(data, region, language, comment, rating, is_technica
     :return: JSON object according to APIv3 single page feedback endpoint definition
     :rtype: ~django.http.JsonResponse
     """
-    event_translation_slug = data.get("slug")
+    # Remove date from the slug for recurrung events
+    event_translation_slug = re.sub("\$\d{4}-\d{2}-\d{2}$", "", data.get("slug"))
```
david-venhoff, line 72: You should probably mark the string as a raw string literal:
```suggestion
    event_translation_slug = re.sub(r"\$\d{4}-\d{2}-\d{2}$", "", data.get("slug"))
```

seluianova: yea, already done.
it worked without it though

david-venhoff (approved):
Thanks, this fixes the issue!
Since normal slugs should not be able to contain the `$` symbol, it should be impossible to manually create a slug that gets changed by this approach.

charludo (approved):
Looks good & works as intended for me!

timoludwig (approved):
LGTM :+1:

timoludwig (changes requested):
@seluianova could you execute `git rebase origin/develop` to remove the merge commit? Thanks!

CHANGELOG.md
```diff
@@ -15,7 +16,6 @@ UNRELEASED
 * [ [#1749](https://github.com/digitalfabrik/integreat-cms/issues/1749) ] Fix region deletion error if media library has nested structure
 * [ [#1170](https://github.com/digitalfabrik/integreat-cms/issues/1170) ] Add map preview on POI form
 * [ [#1579](https://github.com/digitalfabrik/integreat-cms/issues/1579) ] Fix auto-filling of coordinates for multiple street numbers
-* [ [#1767](https://github.com/digitalfabrik/integreat-cms/issues/1767) ] Revert statistics calculation to original & add online downloads column
```
timoludwig, line 18: This line should probably stay...

timoludwig (changes requested):

CHANGELOG.md
seluianova: Sure, fixed. Removed accidentally during merge

Issue #1035, opened 2021-11-24
Enable setting POI position via drag & drop on map


svenseeberg:
### Motivation
We need good way for content creators to enter location coordinates in POIs. We cannot content creators to know the difference between longitude and latitude and are capable of retrieving the numbers from a map.

WordPress currently uses Google Maps to resolve the location of an address and then display a marker on a map.

### Proposed Solution
Allow dragging and dropping a marker on a map.

We have our own tile server: maps.tuerantuer.org. The style can be edited as well: https://github.com/Integreat/integreat-maplibre-style#edit-the-style

We should make use of our new UX & design expert to define a clean process for users.

timoludwig:
This kind of depends on #1170

PR #1814, opened 2022-10-29
Enable setting POI position via drag & drop on map

MizukiTemma:
### Short description

This PR enables users to set address and coordinates of POI by drag & drop of pin on the map.

### Proposed changes

- Set the map pin draggable.
- After each drag&drop (change in coordinates) the address will be derived.
- New address and coordinates are suggested, user can choose to update the both or one of them or to discard the change.
- If it is a new POI and there is no address/coordinates, the map and pin show the center coordinates of the region.

### Side effects

- 

### Resolved issues

Fixes: #1035


integreat_cms/cms/templates/pois/poi_form.html
```diff
@@ -187,7 +196,43 @@ <h3 class="heading font-bold text-black">
...
+                                </div>
+                                <div id="coordinates_found">
+                                </div>
+                                <input type="radio" id="radio_1" name="radio" value="radio_1"/>
+                                <label for="radio_1">
+                                    {% trans "use address and coodinates of map pin" %}
+                                </label>
+                                <br/>
+                                <input type="radio" id="radio_2" name="radio" value="radio_2"/>
+                                <label for="radio_1">
```
seluianova, line 219: copy-paste issue i guess:
`<label for="radio_1">` here suppose to be `<label for="radio_2">`
and the same below for radio_3

seluianova (changes requested):

seluianova (commented):
Works good! Thank you.
One little improvement idea I got: may be "Discard" should be another button next to "Confirm"? Rather than radio option. Seems like more common UI to me.

integreat_cms/cms/templates/pois/poi_form.html
MizukiTemma: oh, right 😅 Now fixed!

MizukiTemma:
Map preview is adjusted so it shows the last status when the discard button is clicked.

integreat_cms/cms/templates/pois/poi_form.html
```diff
@@ -191,45 +182,89 @@ <h3 class="heading font-bold text-black">
                             <label>
                                 {% trans "Map" %}
                             </label>
-                            <div id="no_coordinates_alert"
-                                 class="bg-blue-100 border-l-4 border-blue-500 text-blue-500 px-4 py-3 my-2 hidden"
-                                 role="alert">
-                                {% trans "Map preview will be available when coordinates are given." %}
```
timoludwig, line 188: I think with the new feature, it makes sense to show the map all the time and allow users to either set the position via drag & drop or via clicking on the map. No need to hide and show the map when there's always something to do.

integreat_cms/cms/templates/pois/poi_form.html
```diff
@@ -191,45 +182,89 @@ <h3 class="heading font-bold text-black">
...
-                            </div>
                             <div id="map"
                                  class="aspect-video"
-                                 data-url="{% url "get_address_from_coordinates" region_slug=request.region.slug %}">
+                                 data-url="{% url "get_address_from_coordinates" region_slug=request.region.slug %}"
+                                 data-bounding-box="{{ request.region.bounding_box.api_representation }}">
```
timoludwig, line 188: Instead of the center of the region, I'd go with the bounding box when no coordinates are set.

integreat_cms/cms/templates/pois/poi_form.html
```diff
@@ -191,45 +182,89 @@ <h3 class="heading font-bold text-black">
...
+                                    </p>
+                                    <p id="updated_country">
+                                    </p>
                                 </div>
-                                <input type="radio" id="radio_1" name="radio" value="radio_1"/>
-                                <label for="radio_1" class="secondary !inline">
-                                    {% trans "use address and coodinates of map pin" %}
+                                <input type="radio"
+                                       id="update_address_and_coordinates"
+                                       name="update_position"
```
timoludwig, line 226: I'd use more meaningful ids and names for the radio buttons

integreat_cms/cms/templates/pois/poi_form.html
```diff
@@ -191,45 +182,89 @@ <h3 class="heading font-bold text-black">
...
+                                       required
+                                       checked/>
+                                <label for="update_coordinates" class="secondary !inline">
+                                    {% trans "Update only coordinates" %}
                                 </label>
                                 <br/>
+                                <div id="update_address_error"
+                                     class="bg-red-100 border-l-4 border-red-500 text-red-500 px-4 py-3 my-2 hidden"
+                                     role="alert">
+                                    {% trans "Please choose one option or cancel" %}
```
timoludwig, line 259: I'd show an error when no option is selected

integreat_cms/cms/views/pois/poi_actions.py
```diff
@@ -13,8 +13,6 @@
 
 from linkcheck.models import Link
 
-from geopy.point import Point
```
timoludwig: I think the view should not need to import any third-party code to use public functions of the nominatim API client - instead, we should perform the conversion from float numbers to a `Point` object in the client itself.

integreat_cms/cms/views/pois/poi_actions.py
```diff
@@ -276,8 +268,8 @@ def get_address_from_coordinates(request, region_slug):
 
     return JsonResponse(
         data={
-            "number": address.get("house_number") or "",
-            "street": address.get("road") or "",
+            "number": address.get("house_number"),
+            "street": address.get("road"),
```
timoludwig, line 272: I think `None` is more meaningful than `""` here (and we're not showing these values if they're empty anyway)

integreat_cms/nominatim_api/nominatim_api_client.py
```diff
@@ -242,47 +243,29 @@ def get_region_bounding_box(self, name, aliases=None):
...
-        :rtype: geopy.location.Location
+        :return: The address at these coordinates
+        :rtype: dict
         """
-        if query_str and query_dict:
-            raise RuntimeError(
-                "You can either specify query_str or pass additional keyword arguments, not both."
-            )
-
-        query = query_str or query_dict
```
timoludwig: This does not make sense in this case because the [`reverse()` function](https://geopy.readthedocs.io/en/stable/#geopy.geocoders.Nominatim.reverse) of the underlying library only accepts a `query` parameter and no dictionary.

integreat_cms/nominatim_api/nominatim_api_client.py
```diff
@@ -242,47 +243,29 @@ def get_region_bounding_box(self, name, aliases=None):
...
-
+        coordinates = Point(latitude, longitude)
         try:
-            result = self.geolocator.reverse(
-                query,
-                exactly_one=exactly_one,
-                addressdetails=addressdetails,
```
timoludwig: `exactly_one` and `addressdetails` are `True` by default for the reverse function

integreat_cms/static/src/js/pois/poi-map.ts
```diff
@@ -1,165 +1,242 @@
...
+    // Render the map initially
+    const [map, marker] = renderMap() as [Map, Marker];
 
-    showPreview();
-    longitude.addEventListener("focusout", showPreview);
-    latitude.addEventListener("focusout", showPreview);
+    // Update the map every time a coordinate changes
+    const [latitudeInput, longitudeInput] = getCoordinateInputFields();
+    latitudeInput.addEventListener("focusout", updateMap.bind(null, map, marker));
+    longitudeInput.addEventListener("focusout", updateMap.bind(null, map, marker));
```
timoludwig, line 12: I think it doesn't make sense to render the complete map new each time the coordinates change - just render the map once initially and after that, we can change the position via [`flyTo()`](https://maplibre.org/maplibre-gl-js-docs/api/map/#map#flyto) (which also looks a bit nicer)...

integreat_cms/static/src/js/pois/poi-map.ts
```diff
@@ -1,165 +1,242 @@
...
-    let address = construct_address(data.number, data.street, data.postcode, data.city, data.country);
-    let coordinates = "Latitude: " + lngLat.lat + ", Longitude: " + lngLat.lng;
-
-    document.getElementById("address_found").textContent = address;
-    document.getElementById("coordinates_found").textContent = coordinates;
-
-    let street_number = data.street + " " + data.number;
-    document.getElementById("confirm_map_input").addEventListener("click", () => {
-        use_map_pin_adresse(street_number, data.postcode, data.city, data.country, lngLat.lat, lngLat.lng);
-    });
```
timoludwig: If you set the event listener each time the coordinates change, you pile up a bunch of event handlers and if you use the feature multiple times subsequently, the function will run too often... Instead, we should set this event handler on `load` and think about another way to pass the new values to the update function (I did it by splitting the fields into separate `<span>` elements which can just be retrieved when the new position is confirmed).

integreat_cms/static/src/js/pois/poi-map.ts
```diff
@@ -1,165 +1,242 @@
...
         }),
     });
 
-    const data = await response.json();
-
-    let error = document.getElementById("nominatim-error");
-    if (response.status != 200) {
-        error.textContent = data.error;
-        error.classList.remove("hidden");
-        return;
```
timoludwig: I think we shouldn't abort on error - even if the address cannot be found, it might be useful to update only the coordinates without the address.

integreat_cms/static/src/js/pois/poi-map.ts
```diff
@@ -1,165 +1,242 @@
...
+function constructStreet(street: string, number: string) {
+    // Only append street number if street is given
+    if (street && number) {
+        return street + " " + number;
     }
-    return address;
+    return street;
 }
 
-function use_map_pin_adresse(
```
timoludwig: To avoid too many conflicts with #1820, I'd suggest to use the proper naming conventions already even if they're not enforced yet. So in JS functions and variables, we're going to use CamelCase from now on...

integreat_cms/static/src/js/pois/poi-map.ts
```diff
@@ -1,165 +1,242 @@
...
 
-function construct_address(number: string, street: string, postcode: string, city: string, country: string) {
-    let address = ", " + postcode + ", " + city + ", " + country;
-
-    if (number.length != 0 && street.length != 0) {
-        address = street + " " + number + address;
-    } else if (street.length != 0) {
-        address = street + address;
-    } else if (number.length != 0) {
-        address = number + address;
```
timoludwig: I think a number without street does not make sense... I would only show the number when the street is found as well.

timoludwig (approved):
Thanks a lot, very cool feature! :heart_eyes: 
I absolutely love the new maps integration and being able to set the position via the marker really makes a difference! :rocket: 

I had quite a few ideas how to improve the code which didn't really work with the suggestion feature, so I directly committed them (once again :innocent:)
I tried to explain my changes below, feel free to ask when something isn't clear or when you disagree with my ideas.

MizukiTemma:
@timoludwig 
Thank you for the improvement and explanation. It looks and works way better 😻

ulliholtgrave (approved):
Thanks, looks good and is a great improvement :+1:

PR #1821, opened 2022-11-02
Hide line numbers in translation file

timoludwig:
### Short description

I hope this reduces our merge conflicts with the increasing number of simultaneously open PRs.

### Proposed changes

- Only show the filename and omit the line numbers to prevent merge conflicts

### Side effects

- It's a bit harder to find an occurrence of a translated string which potentially makes it harder for translators to get the context of a specific string


seluianova (approved):
Looks good to me!
As I understand, context can be searched within the file by msgid anyway

JoeyStk (approved):
I agree with @seluianova. I think this is an improvement to before and one can still search for the msgid. Thank you a lot for this PR :)

Issue #1806, opened 2022-10-26
Mark POI category as no visible in app


ulliholtgrave:
### Motivation

As the app is not using our POI categories right now, we should inform our users about it.

### Proposed Solution

Just use the same solution we had for our contact Form (see https://github.com/digitalfabrik/integreat-cms/pull/1565)

seluianova:
Didn't you mean different link in the description?

timoludwig:
> Didn't you mean different link in the description?

Maybe he meant: https://github.com/digitalfabrik/integreat-cms/pull/1565 or #1608?

PR #1822, opened 2022-11-02
Mark POI category as not visible in app

seluianova:
### Short description
As the app is not using our POI categories right now, we should inform our users about it.

### Proposed changes
Add text 'currently not publicly visible' into Category section of POI form

### Side effects
Didn't find any

### Resolved issues
Fixes: #1806

### DIFF
--- a/integreat_cms/cms/templates/pois/poi_form.html
+++ b/integreat_cms/cms/templates/pois/poi_form.html
@@ -306,6 +306,7 @@
                             <i icon-name="tag" class="pb-1"></i>
                             {{ poi_form.category.label }}
                         </h3>
+                        ({% trans "Currently not visible in apps" %})
                     </div>
                     <div class="px-4 pb-4 divide-y divide-gray-200 space-y-2">
                         <div>
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-11-08 09:49+0000\n"
+"POT-Creation-Date: 2022-11-08 15:35+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: Integreat <info@integreat-app.de>\n"
 "Language-Team: Integreat <info@integreat-app.de>\n"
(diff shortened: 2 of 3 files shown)


JoeyStk (approved):
Thank you! Yes, I think that should fix it :)

integreat_cms/cms/templates/pois/poi_form.html
```diff
@@ -228,6 +228,7 @@ <h3 class="heading font-bold text-black">
                             <i icon-name="tag" class="pb-1"></i>
                             {{ poi_form.category.label }}
                         </h3>
+                        ({% trans "currently not publicly visible" %})
```
ulliholtgrave, line 231:
```suggestion
                        ({% trans "not visible in app" %})
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -5756,42 +5756,46 @@ msgstr ""
 msgid "Contact details"
 msgstr "Kontaktdaten"
 
-#: cms/templates/pois/poi_form.html:270 cms/templates/pois/poi_form.html:272
+#: cms/templates/pois/poi_form.html:231
+msgid "currently not publicly visible"
```
ulliholtgrave, line 5760:
```suggestion
msgid "not visible in app"
msgst "nicht sichtbar in der App"
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -5756,42 +5756,46 @@ msgstr ""
...
+msgstr "aktuell nicht öffentlich sichtbar"
```
ulliholtgrave, line 5761:
```suggestion

```

ulliholtgrave (changes requested):
Thanks for the quick fix. I would only change the wording a little bit

integreat_cms/cms/templates/pois/poi_form.html
timoludwig: @ulliholtgrave are you sure you want to remove the "currently"? I can imagine that the users won't ever fill in these categories when they think that this is just for their own categorization and won't ever be shown in the app...
And it would be better if at least some of the categories would be filled in as soon as the app implements this feature, right? :thinking:

ulliholtgrave: Yeah, maybe you are right and we should keep the "currently". However, I would strip the "publicly" from the wording as it's also not available in some "private" area.

seluianova: okay, will do "currently not visible in the app"

ulliholtgrave (approved):
Thanks, Looks good now :+1:

Issue #1824, opened 2022-11-03
Tests for api_client_firebase


Gaston69:
### Motivation
The firebase push-notification-sender functions have no unit-tests, that ensure the correct function

### Proposed Solution
writing tests that use a local firebase-emulator

Issue #1738, opened 2022-10-05
Move `PushNotificationSender` into separate app


timoludwig:
### Motivation

We outsourced all clients of external  APIs (e.g. DeepL, Nominatim, SUMM.AI, GVZ) into separate Django apps.

### Proposed Solution

For consistency reasons, I suggest to move the [Firebase API client](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/views/push_notifications/push_notification_sender.py) to a separate app as well.

### Alternatives

Just keep it the way it is

PR #1825, opened 2022-11-03
Move Firebase API client to separate app

Gaston69:
### Short description
Moved pushNotificationSender - functions to separate api_client_firebase folder and added some tests

### Proposed changes
- Separated the functions and configs to api_client_firebase folder.
- Tests for validate PushNotifications are added.

### Resolved issues

Fixes: #1738
Fixes: #1824


Gaston69:
I gave the test a wrong name, should be test_api_client_firebase.py instead of test_firebase_client_api.py ... 
I ll fix this while adding the tests for the firebase-api / issue #1824

MizukiTemma (commented):
Thank you for your contribution and congratulations for the first PR in our repository!

I havetwo changes:

1. add entry to CHANGELOG.md as others do in the section UNRELEASED.
2. rename api_client_firebase to firebase_api for consistency (like summ_ai_api etc)

integreat_cms/api_client_firebase/apps.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+
+from django.apps import apps, AppConfig
+from django.conf import settings
+
+logger = logging.getLogger(__name__)
+
+
+class PushNotificationSenderApiConfig(AppConfig):
```
timoludwig, line 16:
```suggestion
class FirebaseApiConfig(AppConfig):
```
Our internal team vote has shown that we have a preference to keep the current naming scheme of `<service>_api`, so could you rename the directory to `firebase_api`? Thanks!

integreat_cms/api_client_firebase/apps.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    PushNotificationSender API config inheriting the django AppConfig
+    """
+
+    name = "integreat_cms.push_notification_sender_api"
```
timoludwig, line 21:
```suggestion
    name = "integreat_cms.firebase_api"
```

integreat_cms/api_client_firebase/apps.py
```diff
@@ -0,0 +1,54 @@
...
+    api_available = False
```
timoludwig, line 22:
```suggestion
```
I don't think it makes sense to keep track of the API availability... it's an interesting info at server start so we notice potential problems in our server logs, but I would assume that even if the API is unavailable at the time of the check, it is probably a temporary downtime and we don't need to "store" that the api is down - just try it again when a push notification is sent, and hope that the server is up by then...

integreat_cms/api_client_firebase/apps.py
```diff
@@ -0,0 +1,54 @@
...
+        """
+        # Only check availability if running a server
+        if apps.get_app_config("cms").test_external_apis:
+            if settings.FCM_ENABLED:
+                try:
+                    response = requests.get(
+                        f"{settings.FCM_URL}/api/", timeout=3
+                    )
+                    # Require the response to return a valid JSON, otherwise it's probably an error
+                    assert json.loads(response.text)
```
timoludwig, line 36: This availability check does not make sense.
https://fcm.googleapis.com/fcm/send/api is no valid endpoint of the firebase API, so wen cannot use it to check the availability.
Either you find a more suitable endpoint for testing, or we just skip the availability test - see my comment above, I think even if it's not available for a short time, it will probably be up again and just because it had a short down time during server start does not mean we shouldn't try to send push notifications...

integreat_cms/api_client_firebase/push_notification_sender.py
```diff
@@ -0,0 +1,156 @@
...
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+
+from ..cms.forms.push_notifications.push_notification_translation_form import PushNotificationTranslation
+from ..cms.constants import push_notifications as pnt_const
+from ..cms.models import Region
+logger = logging.getLogger(__name__)
+
+# Firebase Push Notifications / Firebase Cloud Messaging #
+class PushNotificationSender:
```
timoludwig, line 14:
```suggestion
class FirebaseApiClient:
```
I would rename this file to "firebase_api_client.py"

integreat_cms/api_client_firebase/push_notification_sender.py
```diff
@@ -0,0 +1,156 @@
+
```
timoludwig, line 1: Could you delete the old file [integreat_cms/cms/utils/push_notification_sender.py](https://github.com/digitalfabrik/integreat-cms/blob/enhancement/push_notification_sender_api/integreat_cms/cms/utils/push_notification_sender.py)? Otherwise we will just have the same code twice.

integreat_cms/core/settings.py
```diff
@@ -114,9 +114,13 @@
...
-###############################
-# Firebase Push Notifications #
-###############################
+
+##############################################################
+# Firebase Push Notifications (Firebase Cloud Messaging FCM) #
+##############################################################
+
+#: API Url
+FCM_URL = "https://fcm.googleapis.com/fcm/send"
```
timoludwig, line 123: If you're introducing a new setting for the URL (which is a good idea), you should also make use of it and use this setting in https://github.com/digitalfabrik/integreat-cms/blob/a723854ef7d330fe00299a6b89d8f19a515625dd/integreat_cms/api_client_firebase/push_notification_sender.py#L20 instead of the hardcoded class variable.

tests/cms/test_firebase_client_api.py
```diff
@@ -0,0 +1,44 @@
...
+from integreat_cms.cms.models.push_notifications.push_notification import PushNotification
+from django.core.exceptions import ImproperlyConfigured
+
+#######################################################################
+# Test for api_client_firebase / push_notification_sender-class (pns) 
+# @TODO
+# def test_pns_send_pn()
+# def test_pns_send_all()
+#######################################################################
+class TestApiClientFirebase:
```
timoludwig, line 12: What's the plan with the tests, do you want to include basic tests in this PR as well, or should they come with #1839? If the latter, I would remove them from this branch for now.

integreat_cms/api_client_firebase/apps.py
```diff
@@ -0,0 +1,54 @@
+"""
+Configuration of PushNotificationSender API app
```
timoludwig, line 2:
```suggestion
Configuration of the Firebase API app
```

integreat_cms/api_client_firebase/apps.py
```diff
@@ -0,0 +1,54 @@
...
+
+from django.apps import apps, AppConfig
+from django.conf import settings
+
+logger = logging.getLogger(__name__)
+
+
+class PushNotificationSenderApiConfig(AppConfig):
+    """
+    PushNotificationSender API config inheriting the django AppConfig
```
timoludwig, line 18:
```suggestion
    Firebase API config inheriting the django AppConfig
```

integreat_cms/api_client_firebase/apps.py
```diff
@@ -0,0 +1,54 @@
...
+                except (
+                    json.decoder.JSONDecodeError,
+                    requests.exceptions.RequestException,
+                    requests.exceptions.Timeout,
+                    AssertionError,
+                ) as e:
+                    logger.error(e)
+                    logger.error(
+                        "PushNotificationSender API is unavailable. You won't be able to "
+                        "automatically import region coordinates and aliases."
```
timoludwig, line 51: This doesn't seem right, the firebase API has nothing to do with region coordinates.
Probably this was just copied too hastily...

integreat_cms/cms/views/push_notifications/push_notification_form_view.py
```diff
@@ -11,12 +11,9 @@
 from django.utils.translation import ugettext as _
 from django.views.generic import TemplateView
 
-from ...utils.push_notification_sender import PushNotificationSender
+from ....api_client_firebase.push_notification_sender import PushNotificationSender
```
timoludwig, line 14:
```suggestion
from ....firebase_api.firebase_api_client import FirebaseApiClient
```

integreat_cms/api_client_firebase/push_notification_sender.py
```diff
@@ -0,0 +1,156 @@
...
+import logging
+import requests
+
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+
+from ..cms.forms.push_notifications.push_notification_translation_form import PushNotificationTranslation
```
timoludwig, line 8:
```suggestion
from ..cms.forms.push_notifications.push_notification_translation_form import (
    PushNotificationTranslation,
)
```
Our automatic formatting tool wants to change these lines.
Please run `./dev-tools/black.sh` before you push your code or execute `pipenv run pre-commit install` do enable the pre commit hooks which do this automatically before every commit.

timoludwig (changes requested):
Thanks a lot! :+1:
And congratz to your first PR :tada: :muscle: 

I did not yet test the push notification sending, but since you did not change any of the actual sending code, I assume everything will be fine. I will do proper testing once this PR is completely finalized, but until then I have a few preliminary change requests.

Also, could you remove the empty file [integreat_cms/api_client_firebase/utils.py](https://github.com/digitalfabrik/integreat-cms/blob/enhancement/push_notification_sender_api/integreat_cms/api_client_firebase/utils.py)?

> Add entry to CHANGELOG.md as others do in the section UNRELEASED.

Sorry, @MizukiTemma, I have to disagree :sweat_smile: 
Since this change is puerly technical and has no impact on our end users, I would not put this into the changelog. This is only relevant for deverlopers, and developers can also do `git log` to see what's changed :wink:

integreat_cms/api_client_firebase/apps.py
Gaston69: -- Änderung abwarten, RM von Johannes/Timon

integreat_cms/api_client_firebase/apps.py
Gaston69: Good issue. - we should talk, how solve this in total. Johannes will talk to you. Then we ll change it as negotiated. 
My suggest: put this or a similar function as integration-test to staging-tests. We could ignore it, if it fails < 3 times, something like this. (Checking the availability of the service does make sense in my opinion. Specially as it is legacy-code / legacy API)

timoludwig: I removed the availability check for now because in its current form it doesn't work.

timoludwig (approved):
Thanks a lot! :+1:

I made a few changes to finalize this PR:
- I moved the test file from `tests/cms` into `tests/firebase_api` because it tests the `firebase_api` app and not the `cms` app
- I deduplicated the URL settings (was defined twice)
- I removed availability check since it makes no sense in the current form
- I added the `requests-mock` dependency to the Pipfile
- I added added the new app to to `INSTALLED_APPS` so the code defined in the `ready()` method actually runs
- I adapted the docstrings to use the correct types and linking to other documentations

Also, I made one final test to check whether sending still works as expected:
photo_2022-11-10_19-27-21

So I think this is good to go! :rocket:

MizukiTemma (approved):
👍

PR #1829, opened 2022-11-03
Add contributor list to GitHub releases

timoludwig:
### Short description

This mentions all contributors of a given release in the release body which automatically adds a contribution footer similar to e.g.:

Screenshot 2022-11-04 at 00-19-44 Build software better together

### Proposed changes

- Add new script to fetch contributors between two given git references
- Add new contributors section in GitHub release body

### Side effects

- We cannot really test the new workflow until the next release, so it's not unlikely that I screwed something up and the notification won't work next time :see_no_evil:  (this shouldn't affect the PyPI release however, just the GitHub release and the Mattermost notification)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -427,15 +427,23 @@ jobs:
           name: Request installation access token to authorize as Deliverino app
           command: echo "export DELIVERINO_ACCESS_TOKEN=$(pipenv run ./.circleci/scripts/get_access_token.py)" >> $BASH_ENV
       - run:
-          name: Assemble release body
+          name: Get previous version tag
           command: |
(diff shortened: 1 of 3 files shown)


MizukiTemma (approved):
I don't find anything that can be potentially harmfull🚀

david-venhoff (approved):
Let's try it!
A small thing I noticed is that commit co-authors would get ignored, but this probably won't have an effect in practice.

timoludwig:
> Let's try it! A small thing I noticed is that commit co-authors would get ignored, but this probably won't have an effect in practice.

Yes, indeed, but I did not find any information about co-authors in the respective API response...
Probably we would have to do an expensive regex-matching in the commit body and additional API calls to get the usernames to the referenced email addresses to retrieve them, but I was a bit too lazy for that...

Issue #1791, opened 2022-10-25
Live content not shown in PDFs


timoludwig:
### Describe the Bug

The PDFs only contain the direct content of the page translations, not the mirrored live content which might be embedded.

### Steps to Reproduce

1. Embed live content in a page
2. Export this page as PDF

### Expected Behavior

The live content should be shown in the PDF

### Actual Behavior

Only the content of the page itself is shown

### Additional Information

![live](https://user-images.githubusercontent.com/16021269/197714940-11c2001f-91d1-49ae-b38d-7157b6f8453b.png)

PR #1835, opened 2022-11-07
Render live content in pdfs

david-venhoff:
### Short description

This pr fixes a bug that live content does not get rendered in pdfs

### Proposed changes

- Include the live content when rendering the pdf
- If no matching translation of the live content exists, render the default error message which lists alternative languages which work

### Side effects

/ 

### Resolved issues

Fixes: #1791

### DIFF
--- a/integreat_cms/cms/models/pages/page_translation.py
+++ b/integreat_cms/cms/models/pages/page_translation.py
@@ -103,6 +103,22 @@ class PageTranslation(AbstractBasePageTranslation):
     @cached_property
     def mirrored_translation_text(self):
         """
+        This method returns the text of the mirrored translation or an empty string,
+        if it is not available in this language
+
+        :return: The text
+        :rtype: str
+        """
+        if self.page.mirrored_page:
+            if translation := self.page.get_mirrored_page_translation(
+                self.language.slug
+            ):
+                return translation.content
+        return ""
+
+    @cached_property
+    def mirrored_translation_text_or_fallback_message(self):
+        """
         This method returns the text of the mirrored translation.
         If there is no mirrored translation, it returns an empty string.
         If there is a mirrored page but no translation, a html error message will be returned with languages that are translated as options.
@@ -113,9 +129,8 @@ class PageTranslation(AbstractBasePageTranslation):
         if not self.page.mirrored_page:
             return ""
 
-        translation = self.page.get_mirrored_page_translation(self.language.slug)
-        if translation and translation.content:
-            return translation.content
+        if content := self.mirrored_translation_text:
+            return content
 
         error_message = (
             self.language.message_partial_live_content_not_available
@@ -179,8 +194,11 @@ class PageTranslation(AbstractBasePageTranslation):
         :rtype: str
         """
         if self.page.mirrored_page_first:
-            return self.mirrored_translation_text + self.display_content
-        return self.display_content + self.mirrored_translation_text
+            return (
+                self.mirrored_translation_text_or_fallback_message
+                + self.display_content
+            )
+        return self.display_content + self.mirrored_translation_text_or_fallback_message
 
     @cached_property
     def combined_last_updated(self):
--- a/integreat_cms/cms/templates/pages/page_pdf.html
+++ b/integreat_cms/cms/templates/pages/page_pdf.html
@@ -87,7 +87,13 @@
                         {% endif %}
                         {% get_public_translation page language.slug as page_translation %}
                         <h1 class="level-{{ page.depth|add:"-1" }}">{{ page_translation.title }}</h1>
-                        <div class="content" style="padding-bottom: 30px;">{{ page_translation.content|pdf_strip_fontstyles|safe }}</div>
+                        <div class="content" style="padding-bottom: 30px;">
+                            {% if page.mirrored_page_first %}{{ page_translation.mirrored_translation_text|pdf_strip_fontstyles|safe }}{% endif %}
+                            {{ page_translation.content|pdf_strip_fontstyles|safe }}
+                            {% if not page.mirrored_page_first %}
+                                {{ page_translation.mirrored_translation_text|pdf_strip_fontstyles|safe }}
+                            {% endif %}
+                        </div>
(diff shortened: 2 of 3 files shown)


MizukiTemma (approved):
Works as expected with both region internal and external live contents🙂

integreat_cms/cms/templates/pages/page_pdf.html
```diff
@@ -87,7 +87,13 @@ <h1 id="title_page">{{ language.table_of_contents }}</h1>
                         {% endif %}
                         {% get_public_translation page language.slug as page_translation %}
                         <h1 class="level-{{ page.depth|add:"-1" }}">{{ page_translation.title }}</h1>
-                        <div class="content" style="padding-bottom: 30px;">{{ page_translation.content|pdf_strip_fontstyles|safe }}</div>
+                        <div class="content" style="padding-bottom: 30px;">
+                            {% if page.mirrored_page_first %}{{ page_translation.mirrored_translation_text|pdf_strip_fontstyles|safe }}{% endif %}
+                            {{ page_translation.content|pdf_strip_fontstyles|safe }}
+                            {% if not page.mirrored_page_first %}
+                                {{ page_translation.mirrored_translation_text|pdf_strip_fontstyles|safe }}
```
timoludwig, line 94: Since you're not using `display_content` I assume you don't want to show the fallback texts in the PDF - which makes sense to me - but the `mirrored_translation_text` also might return a fallback text when the mirrored translation does not exist... so maybe it would be better to use a different property here which is just empty when the mirrored content does not exist? :thinking:

timoludwig (commented):
Thanks, this works as expected! :+1:
I just have one concern about fallback texts in mirrored translations.

integreat_cms/cms/templates/pages/page_pdf.html
david-venhoff: I decided to use this tag because I think it could be confusing if some part of a page would just silently not render, and the error message would still be helpful in pdfs to inform the reader.
If you want though, I can change it to not render the message

integreat_cms/cms/models/pages/page_translation.py
```diff
@@ -100,8 +100,37 @@ def short_url(self):
             "public:expand_page_translation_id", kwargs={"short_url_id": self.id}
         )
 
+    @cached_property
+    def mirrored_translation_text_internal(self):
```
timoludwig, line 104:
```suggestion
    def mirrored_translation_text(self):
```
I think we can combine these methods into one.

integreat_cms/cms/models/pages/page_translation.py
```diff
@@ -100,8 +100,37 @@ def short_url(self):
...
+        """
+        This method returns the text of the mirrored translation or None,
```
timoludwig, line 106:
```suggestion
        This method returns the text of the mirrored translation or an empty string,
```

integreat_cms/cms/models/pages/page_translation.py
```diff
@@ -100,8 +100,37 @@ def short_url(self):
...
+        if it is not available in this language
+
+        :return: The text or None
+        :rtype: str or None
```
timoludwig, line 110:
```suggestion
        :return: The text
        :rtype: str
```

integreat_cms/cms/models/pages/page_translation.py
```diff
@@ -100,8 +100,37 @@ def short_url(self):
...
+        """
+        if not self.page.mirrored_page:
+            return None
+
+        translation = self.page.get_mirrored_page_translation(self.language.slug)
+        if translation and translation.content:
+            return translation.content
+
+        return None
```
timoludwig, line 119:
```suggestion
        if self.page.mirrored_page:
            if translation := self.page.get_mirrored_page_translation(self.language.slug):
                return translation.content
        return ""
```
I think this can be simplified a bit

integreat_cms/cms/models/pages/page_translation.py
```diff
@@ -100,8 +100,37 @@ def short_url(self):
...
     def mirrored_translation_text(self):
+        """
+        This method returns the text of the mirrored translation or an empty string,
+        if it is not available in this language
+
+        :return: The text
+        :rtype: str
+        """
+        return self.mirrored_translation_text_internal or ""
+
```
timoludwig, line 131:
```suggestion
```
I think we don't need a function for such a trivial task

integreat_cms/cms/models/pages/page_translation.py
```diff
@@ -113,9 +142,9 @@ def mirrored_translation_text(self):
         if not self.page.mirrored_page:
             return ""
 
-        translation = self.page.get_mirrored_page_translation(self.language.slug)
-        if translation and translation.content:
-            return translation.content
+        content = self.mirrored_translation_text_internal
+        if content:
+            return content
```
timoludwig, line 147:
```suggestion
        if content := self.mirrored_translation_text:
            return content
```

integreat_cms/cms/models/pages/page_translation.py
```diff
@@ -179,8 +208,10 @@ def combined_text(self):
         :rtype: str
         """
         if self.page.mirrored_page_first:
-            return self.mirrored_translation_text + self.display_content
-        return self.display_content + self.mirrored_translation_text
+            return (
+                self.mirrored_translation_text_or_error_message + self.display_content
```
timoludwig, line 212:
```suggestion
                self.mirrored_translation_text_or_fallback_message + self.display_content
```

integreat_cms/cms/models/pages/page_translation.py
```diff
@@ -179,8 +208,10 @@ def combined_text(self):
...
+            )
+        return self.display_content + self.mirrored_translation_text_or_error_message
```
timoludwig, line 214:
```suggestion
        return self.display_content + self.mirrored_translation_text_or_fallback_message
```

timoludwig (approved):
Thank you! :pray:

Issue #1832, opened 2022-11-05
Add opening hours to the POI form and api response


ulliholtgrave:
### Motivation

We need a possibility to add opening hours to POIs as structured data. These also should be delivered via the API. See https://github.com/digitalfabrik/integreat-cms/issues/959 for more information.

PR #1837, opened 2022-11-07
Add opening hours widget to location form

timoludwig:
### Short description

This adds the possibility to manage opening hours for locations

### Proposed changes

- Add JSON field to POI model
- Add widget to render current opening hours in sidebar of editing form
- Add widget to modify opening hours as popup to allow to change the opening hours for multiple days at once
- Deliver opening hours in API

### Missing

- Server-side validation of format
- Client-side validation of overlapping time slots

### Resolved issues

Fixes: #1832


MizukiTemma (changes requested):
Thank you for the PR, it looks really cool and very intuitive 😃 

Sadly I found a case where the week day choice is not working well. To reproduce:
1) Click "Edit Mo-Fr"
2) Give all the week days 10:00-18:00 and save.
3) Click "Edit Mo-Fr" again
4) Choose only monday (click out Tue - Fr), give 10:00-15:00 and save.
5) For all weekdays the opening hours are changed to 10:00-15:00.

The same happens also when you choose a weekday by clicking on the week day row in step 3.
If the poi form itself is saved (published) between step 2 and 3, it works fine.

Another thing to adjust is the layout of popup in English version.
The weekday circles are not lining up well (see the screen shot).
English ver

timoludwig:
> Sadly I found a case where the week day choice is not working well.

Oh, good catch! The reason was that even though `{ ...openingHoursBuffer }` correctly created an independent copy of the object, the `timeSlot` objects inside that object were still given as reference and not duplicated, so even if the `allDay` and `closed` fields were independent, the time slots of all edited days were "connected" to the same objects in memory, meaning they would also change when one of the other days is updated later... 

0 RGt-o4ovYiIt_9nS

I solved this problem now by doing `JSON.parse(JSON.stringify(openingHoursBuffer))`, not sure if there is a better way :see_no_evil: 

> Another thing to adjust is the layout of popup in English version. The weekday circles are not lining up well (see the screen shot).

Ah, yeah, should be fixed now as well.

MizukiTemma:
Yes, it's fixed! Thank you 👍

MizukiTemma (approved):

ulliholtgrave (approved):
Thanks, That looks really fancy :+1:

Issue #1827, opened 2022-11-03
Multi-language-XLIFF export popup not scrollable


timoludwig:
### Describe the Bug

When very many languages exist in a region and/or the user has a very small screen, not all of the languages fit onto the selection popup and the user has to zoom out to be able to submit the export.

### Steps to Reproduce

1. Create many languages or make window size smaller
2. Try to export multiple languages at once

### Expected Behavior

The overflowing content of the popup should be scrollable.

### Actual Behavior

The size is fixed and overflowing content is hidden

### Additional Information

Probably, this affects other popups as well. Ideally, this can be solved for all of them at once.

![Screenshot (777)](https://user-images.githubusercontent.com/16021269/199816077-9946e8b8-e580-4337-b1d0-303e8f3b74d3.png)

PR #1848, opened 2022-11-09
Make Multi-language-XLIFF export popup scrollable

JoeyStk:
### Short description
Make multi language popup scrollable so that even on smaller screens the end doesn't get cut off

### Proposed changes
- Add view height to popups, set to 80vh
- Add overflow scroll
- Change only for this popup. Couldn't find a popup that has the same issue. All other popups I considered had less height and therefore didn't have this issue
- Add custom height class for 80vh

### Side effects
There are some screens which would have just a minimal scroll which can be a bit ugly and/or annoying, but I hope that's okay

### Resolved issues

Fixes: #1827

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 UNRELEASED
 ----------
 
+* [ [#1827](https://github.com/digitalfabrik/integreat-cms/issues/1827) ] Make Multi-language-XLIFF export popup scrollable
+
 
 2022.11.1
 ---------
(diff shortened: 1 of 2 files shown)


JoeyStk:
Yes, makes sense. That makes it a lot easier!

membralala (approved):
Thank you, this fixes the issue! 🙂

timoludwig (approved):
Thanks! :pray:

Issue #1820, opened 2022-11-02
Add configs for eslint and prettier


charludo:
### Motivation

Currently, `eslint` and `prettier` may heavily format files depending on the developer's local setup. Adding project-specific configuration for both would standardize how our files are formatted. 

### Proposed Solution

Copy `.eslintrc` and `.prettierd` from digitalfabrik/integreat-app

### Alternatives

Write our own

Issue #1066, opened 2021-12-15
Add option to integrate Sentry


timobrembeck:
### Motivation

To monitor Javascript errors on the user clients, it should be possible to integrate [Sentry](https://sentry.io/).

### Proposed Solution

Follow the documentation:
- https://docs.sentry.io/platforms/javascript/

### Additional Context

It should be possible to configre both the sentry server and the key via the config file `/etc/integreat-cms.ini` (which means that the values need to be read from the environment variables in `settings.py`).

svenseeberg:
Does this cover JavaScript or Python errors? If this is about Python errors, I would currently not use Sentry for this and keep our mechanism with logging + InfluxDB + Grafana.

timoludwig:
The linked docs only describe how to track Python errors, so yes, it's a bit redundant. But according to the docs, it should be fairly easy to integrate everything, also the Javascript errors: https://docs.sentry.io/platforms/javascript/

svenseeberg:
Can we make this issue about tracking JS errors in the browsers? I would prefer to not transfer any server data to sentry :D

timoludwig:
Alright, makes sense. I updated the issue description.

JoeyStk:
We would need a knowledge exchange for this

PR #1849, opened 2022-11-09
Add configs for eslint and prettier

charludo:
### Short description
This PR adds:
- a `.prettierrc` and `.prettierignore` file usable by most modern IDEs and the existing `prettier.sh` dev tool
- a `.eslintrc.js` adapted from digitalfabrik/integreat-app, albeit heavily modified for our needs, also working with most modern IDEs and the new `eslint.sh` dev tool
- a CircleCI job for eslint 

### Proposed changes

- added mentioned files
- modified almost every JS file in the `/static/` subdir in order to fix linting errors

### Side effects

- merge conflicts with (most likely) every currently open PR touching any static file
- it is possible that parts of the CMS silently broke during this refactor. Still checking that.

### Resolved issues

Fixes: #1820


timoludwig (changes requested):
Awesome, thanks a lot! :pray:

I love the attention to detail and level of perfectionism you have put into this task :heart_eyes: 

I have not read every single line yet (duh!), but I could not find major problems in the refactored code. Obviously there will be a few errors hidden in the large diff, but I doubt that we'll find them with reasonable effort.... instead, I suggest that for once we follow the example of every other IT company and let our users find the bugs for us.
We just have to make sure we're available in the days after the next release to fix any urgent problems.
Also, we could think about prioritizing #1066 a bit higher and also include this in the next release to get better feedback about errors on the user's devices.

But before the conflicts pile up again, I'd vote for a yolo merge :cat:

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 UNRELEASED
 ----------
 
+* [ [#1820](https://github.com/digitalfabrik/integreat-cms/issues/1820) ] Add configs for eslint and prettier
```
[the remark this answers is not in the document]
charludo: True :D

dev-tools/eslint.sh
```diff
@@ -0,0 +1,12 @@
...
+
+# This script can be used to lint the static files with eslint
+
+# Import utility functions
+# shellcheck source=./dev-tools/_functions.sh
+source "$(dirname "${BASH_SOURCE[0]}")/_functions.sh"
+
+# Run prettier
+echo "Starting code linting with eslint..." | print_info
+eslint .
```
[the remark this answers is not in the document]
charludo: Ah right, didn't think anout that.

integreat_cms/static/src/js/media-management/component/message.tsx
```diff
@@ -4,56 +4,57 @@
...
-                        )}
-                        role="alert"
-                    >
-                        {message.text}
-                    </div>
-                );
-            })}
+            {messageQueue.map((message) => (
+                <div
+                    key={window.btoa(message.text).slice(-1 * messageKeyLength)}
```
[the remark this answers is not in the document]
charludo: Yeah, same thing in React, it's a bit of a bothersome error. 

Good catch. I really wanted to use a proper hashing function here, turns out, JS doesn't come eith one OOTB, and I didn't want to implement / import one just for that.

Fair point though, I'll commit your suggestion!

(Btw: a review this thorough, at 3:29 in the morning? Are you alright?? 😅 😂 )

.circleci/config.yml
```diff
@@ -156,7 +156,18 @@ jobs:
           at: .
       - run:
           name: Check formatting of CSS & JS files
-          command: npx prettier --print-width 120 --tab-width 4 --check integreat_cms/static/src
+          command: npx prettier --check .
+  eslint:
```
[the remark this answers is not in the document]
charludo: Thanks. Haven't worked with circleCI before 😅

charludo:
Does anyone have an idea for how to proerly escape DOM attributes (see CodeQL complaints)?

.eslintrc.js
```diff
@@ -0,0 +1,143 @@
...
+    },
+    ignorePatterns: [
+        "**/dist/",
+        "**/htmlcov/**",
+        "**/node_modules/",
+        "*.config.js",
+        "*.json",
+        "*.lock",
+        ".eslintrc.js",
+        ".venv/",
```
timoludwig, line 14:
```suggestion
        ".venv/",
        "build/",
        "docs/",
```
Sometimes I build the package locally with `python setup.py sdist bdist_wheel`, so ignoring `build/` would help with that. And the documentation build in `docs/` also produced a lot of errors for me.

timoludwig (changes requested):
Thanks! Sorry for not noticing this the last time, I would like to exclude two more directories from eslint. Other than that, it looks fine! :+1:

> Does anyone have an idea for how to properly escape DOM attributes (see CodeQL complaints)?

Hmm, not quite sure what CodeQL wants us to do. If we do some kind of manual escaping, it will probably not detect it (like your attempts), and there does not seem to be a standard way of doing this (yet?)... and I'm not even sure if the complaints are valid. E.g. when I put `'">` as a POI street name and use this POI as event location, the link to google maps just looks like this:
```
<a id="poi-google-maps-link"
    href="https://www.google.com/maps/search/?api=1&query=&#x27;&quot;&gt;&lt;img src=&quot;x&quot; onerror=alert(document.cookie);&gt;,Augsburg,Deutschland"
    class="text-blue-500 hover:underline"
    target="_blank"
    rel="noopener noreferrer">
    Auf Google Maps öffnen
</a>
```
So it's definitely escaped... probably by Django though, that's why CodeQL doesn't notice it...
That's also the case for almost all other warnings, the values are directly coming from our CMS and there is no possibility for users to intercept them. The only exception is maybe the RSS feed parsing code which looks a it suspicious to me, but at least we're only parsing our own feed and not from untrusted sources. I would still recommend to revisit this topic at some point, but I guess for this PR it's out of scope. In the future, there seems to be the [HTML Sanitizer API](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Sanitizer_API) for this stuff, so maybe let's keep an eye on this.

.eslintrc.js
charludo: yup, good point. Probably also exclude them from prettier?

So I'll just undo the escape-attempts and dismiss the complaints for now as wontfix?

timoludwig: Yes, sounds good :+1:

timoludwig (approved):
Thanks a lot! :pray:

I'd suggest the following timeline:

We try to finalize as many of our open PRs as possible until tomorrow, then we'll do another release to get these features live, then you resolve the conflicts one last time and squash the commits and we'll merge this PR. Then we'll have the weekend and maybe a few days after (until the next urgent bug appears :grin:) to test everything and notice potential problems.
Do you think this sounds reasonable?

charludo:
> Thanks a lot! pray
> 
> I'd suggest the following timeline:
> 
> We try to finalize as many of our open PRs as possible until tomorrow, then we'll do another release to get these features live, then you resolve the conflicts one last time and squash the commits and we'll merge this PR. Then we'll have the weekend and maybe a few days after (until the next urgent bug appears grin) to test everything and notice potential problems. Do you think this sounds reasonable?

Yeah, that sounds sensible! Guess it's only fair that I should be responsible for the merge conflicts I cause :D

ulliholtgrave (commented):
Looks good 👍 But could you squash your commits? :) 136 commits are a little bit too much in my opinion :D

charludo:
> Looks good +1 But could you squash your commits? :) 136 commits are a little bit too much in my opinion :D

Absolutely, just thought it would be easier to review the individual changes this way :D

ulliholtgrave:
Can you do another rebase on develop? After the release the develop state is a little bit different :)

timoludwig:
A rebase will be a lot of effort to fix the conflicts again which already have been resolved by the merge commits...
I'd suggest
```
# Set local feature/add-configs-for-eslint-and-prettier branch to the state of develop
git reset --hard origin/develop
# Create one commit with all commits of the origin branch
git merge --squash origin/feature/add-configs-for-eslint-and-prettier
# Commit the squash merge
git commit
```

charludo:
Goog idea, worked like a charm :)

integreat_cms/static/src/index.ts
```diff
@@ -5,14 +5,10 @@
  *
  *     https://preactjs.github.io/preact-devtools/
  */
-if (process.env.NODE_ENV !== "production") {
-    require("preact/debug");
-}
 /* Babel requirements & polyfills */
```
timoludwig, line 8:
```suggestion
/* eslint-disable import/first */
if (process.env.NODE_ENV !== "production") {
    require("preact/debug");
}
/* Babel requirements & polyfills */
```
Unfortunately the preact debug imports [needs to be imported at the top](https://preactjs.com/guide/v10/debugging/#strip-devtools-from-production), even if it uses the require syntax due to the exclusion of production mode...

integreat_cms/static/src/index.ts
```diff
@@ -89,8 +85,16 @@ import "./js/pois/opening-hours/index.tsx";
 
 import "./js/menu.ts";
 
+if (process.env.NODE_ENV !== "production") {
+    /* eslint-disable-next-line global-require */
+    require("preact/debug");
+}
```
timoludwig, line 91:
```suggestion
```

ulliholtgrave (approved):
Buckle up and let's go 🙏  👍

PR #1850, opened 2022-11-10
Reset `POT-Creation-Date` in translation file to avoid git conflicts

timoludwig:
### Short description

I noticed that #1821 wasn't quite enough to reduce the git conflicts in the translation file due to the constantly changing date...
Since this information isn't really that useful anyway, I suggest to remove it in our dev tools.

### Proposed changes

- Reset `POT-Creation-Date` in translation file to avoid git conflicts


JoeyStk (approved):
Thank you a lot! This is a very helpful PR :) 
I'm absolutely not a bash expert but as far as I can tell this looks good :) Maybe it would be good if someone with more experience in bash could also take a look at this :)

timoludwig:
> Maybe it would be good if someone with more experience could also take a look at this :)

@svenseeberg I can picture you really wanting to read some Bash :joy:

dev-tools/translate.sh
```diff
@@ -12,17 +12,21 @@ ensure_not_root
...
-fi
+# Reset POT-Creation-Date to avoid git conflicts
+sed --in-place --regexp-extended 's/^"POT-Creation-Date: [0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}\+[0-9]{4}\\n"$/"POT-Creation-Date: YEAR-MO-DA HO:MI+ZONE\\n"/' "${TRANSLATION_FILE}"
 
-# Compile translation file
-echo "Compiling translation file..." | print_info
-pipenv run integreat-cms-cli compilemessages --verbosity "${SCRIPT_VERBOSITY}"
+# Skip compilation if --skip-compile option is given
+if [[ "$*" == *"--skip-compile"* ]]; then
+    # Compile translation file
```
charludo, line 27:
```suggestion
# Skip compilation if --skip-compile option is given
if [[ "$*" != *"--skip-compile"* ]]; then
    # Compile translation file
```
😉

charludo (changes requested):
Agree, really useful change!! Other than the one comment, LGTM

dev-tools/translate.sh
timoludwig: oh haha :sweat_smile:

david-venhoff: Seems to still be `==`

timoludwig: :facepalm:

david-venhoff (approved):
Looks good to me now :tada:

Issue #1844, opened 2022-11-09
Missing gender sensitive language


osmers:
### Describe the Bug
Gender sensitive language for some cases is missing

### Steps to Reproduce

1. Go to Benutzer:innen
2. Click to create a new account
3. See that the user name is still called Benutzername
4. See that the description for active status is as follows.
Legt fest, ob dieser Benutzer aktiv ist. Kann deaktiviert werden, anstatt Benutzer zu löschen.

### Expected Behavior
All words should be gendered or gender neutral

### Actual Behavior
The male word Benutzer is used

### Additional Information/suggestion
Rename single word Benutzer to Benutzer:in or Benutzer:innen depending on context. If it is a composite word, like Benutzername, use Kontoname
Possible need to change the english original accordingly from user name to account name

JoeyStk:
This is not as straightforward as it might seem on first glance. This issue is due to the django base user form. It has nothing to do with our translation (file). There is a way to change this, but [the Django documentation](https://docs.djangoproject.com/en/4.1/topics/auth/customizing/#specifying-custom-user-model) warns to do so. I think we have to reconsider, if we really want to do this

PR #1852, opened 2022-11-10
Use gender sensitive language in user form

timoludwig:
### Short description

A few fields in the user form were still missing because we're inheriting the base user model from `django.contrib.auth`, so we had to manually overwrite them.

### Proposed changes

- Use gender sensitive language in user form

### Resolved issues

Fixes: #1844


svenseeberg (approved):
Except for the migrations this should be good to go. The tests will tell :D

timoludwig:
> Except for the migrations this should be good to go. The tests will tell :D

Ah yeah, sorry, missed them :sweat_smile:

JoeyStk (approved):
Ah yes! That is a lot better than what I have tried! Thank you so much!

Issue #1788, opened 2022-10-24
Broken translation status if only minor public versions exist


timoludwig:
### Describe the Bug

When a POI or Event is published as minor edit, the translation status shows "Fallback translation", which does not make sense.

### Steps to Reproduce

1. Create POI as draft
2. Do not change content and publish - this should create a new versions which is marked as minor edit
3. See translation status in language tab - it is marked as fallback translation

### Expected Behavior

The translation should be up to date.
We can achieve this by setting all past versions to public, similar to how we do it for pages:

https://github.com/digitalfabrik/integreat-cms/blob/f90e5aacf3095537a6e19e30d5e2689fccf45a13/integreat_cms/cms/views/pages/page_form_view.py#L342-L350

### Actual Behavior

The translation status is incorrect

### Additional Information

![Screenshot 2022-10-24 at 16-45-23 Integreat Redaktionssystem](https://user-images.githubusercontent.com/16021269/197555076-01633fd3-49f3-42ac-94c7-093091bb4d72.png)

PR #1854, opened 2022-11-10
Fix broken translation status if only minor public versions exists

JoeyStk:
### Short description
Fixes the issue with the broken translation status if only minor public versions exists for POI and event

### Proposed changes
- Add extra if-condition to ensure the correct translation status is set

### Side effects
I don't think there are any.  

### Resolved issues
Fixes: #1788

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ UNRELEASED
 * [ [#1791](https://github.com/digitalfabrik/integreat-cms/issues/1791) ] Render live content in pdfs
 * [ [#1869](https://github.com/digitalfabrik/integreat-cms/issues/1869) ] Fix error in imprint side by side view
 * [ [#1786](https://github.com/digitalfabrik/integreat-cms/issues/1786) ] Remove texblock option in editor, add button to clear all formatting
+* [ [#1788](https://github.com/digitalfabrik/integreat-cms/issues/1788) ] Fix broken translation status of events & locations if only minor public versions exists
 
 
 2022.11.3
--- a/integreat_cms/cms/views/events/event_form_view.py
+++ b/integreat_cms/cms/views/events/event_form_view.py
@@ -236,6 +236,13 @@ class EventFormView(
                     language__in=languages
                 ).update(status=status.DRAFT)
 
+            elif (
+                event_translation_form.instance.status == status.PUBLIC
+                and event_translation_form.instance.minor_edit
+            ):
+                event_translation_form.instance.event.translations.filter(
+                    language=language
+                ).update(status=status.PUBLIC)
             # Show a message that the slug was changed if it was not unique
             if user_slug and user_slug != event_translation_form.cleaned_data["slug"]:
                 other_translation = EventTranslation.objects.filter(
--- a/integreat_cms/cms/views/pois/poi_form_view.py
+++ b/integreat_cms/cms/views/pois/poi_form_view.py
@@ -177,6 +177,13 @@ class POIFormView(
                 poi_translation_form.instance.poi.translations.filter(
                     language__in=languages
                 ).update(status=status.DRAFT)
+            elif (
+                poi_translation_form.instance.status == status.PUBLIC
+                and poi_translation_form.instance.minor_edit
+            ):
+                poi_translation_form.instance.poi.translations.filter(
+                    language=language
+                ).update(status=status.PUBLIC)
 
             # Show a message that the slug was changed if it was not unique
             if user_slug and user_slug != poi_translation_form.cleaned_data["slug"]:


integreat_cms/cms/views/events/event_form_view.py
```diff
@@ -236,6 +236,13 @@ def post(self, request, **kwargs):
                     language__in=languages
                 ).update(status=status.DRAFT)
 
+            if (
+                event_translation_form.instance.status == status.PUBLIC
+                and event_translation_form.instance.minor_edit
+            ):
+                event_translation_form.instance.poi.translations.filter(
```
MizukiTemma, line 243:
```suggestion
                event_translation_form.instance.event.translations.filter(
```

MizukiTemma (changes requested):
Thank you for PR 👍 

I think this is just a copy & paste thing though have to suggest a change 😅 See code suggestion.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ UNRELEASED
 * [ [#1521](https://github.com/digitalfabrik/integreat-cms/issues/1521) ] Show warning when user entered slug got changed
 * [ [#1807](https://github.com/digitalfabrik/integreat-cms/issues/1807) ] Remove short description field from POI form
 * [ [#522](https://github.com/digitalfabrik/integreat-cms/issues/522) ] Add region setting to activate SEO section, add SEO section to POI form
+* [ [#1788](https://github.com/digitalfabrik/integreat-cms/issues/1788) ] Fix broken translation status if only minor public versions exists
```
timoludwig, line 8:
```suggestion
* [ [#1788](https://github.com/digitalfabrik/integreat-cms/issues/1788) ] Fix broken translation status of events & locations if only minor public versions exists
```

integreat_cms/cms/views/events/event_form_view.py
```diff
@@ -236,6 +236,13 @@ def post(self, request, **kwargs):
                     language__in=languages
                 ).update(status=status.DRAFT)
 
+            if (
```
timoludwig, line 239:
```suggestion
            elif (
```
We don't need to check this condition when the status is `DRAFT`

integreat_cms/cms/views/pois/poi_form_view.py
```diff
@@ -177,6 +177,13 @@ def post(self, request, *args, **kwargs):
                 poi_translation_form.instance.poi.translations.filter(
                     language__in=languages
                 ).update(status=status.DRAFT)
+            if (
```
timoludwig, line 180:
```suggestion
            elif (
```

timoludwig (changes requested):

integreat_cms/cms/views/events/event_form_view.py
timoludwig: It's a bit sad that our tests did not reveal this error. Unfortunately we can't add automated tests for every edge case of the form submissions... at least for the future, we could test this specific case by adding
```
            (
                "edit_event",
                PRIV_STAFF_ROLES + REGION_ROLES,
                {
                    "title": "new title",
                    "start_date": "2030-01-01",
                    "end_date": "2030-01-01",
                    "is_all_day": True,
                    "status": status.PUBLIC,
                    "minor_edit": True
                },
            ),
```
to the [view test config](https://github.com/digitalfabrik/integreat-cms/blob/3ed0649012dfa9945ea183e35ae401ef0a950413/tests/cms/views/view_config.py#L736-L746).

JoeyStk:
Thank you :) All of it makes sense :)

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ UNRELEASED
 * [ [#522](https://github.com/digitalfabrik/integreat-cms/issues/522) ] Add region setting to activate SEO section, add SEO section to POI form
 * [ [#1858](https://github.com/digitalfabrik/integreat-cms/issues/1858) ] Enable submitting feedback about fallback translations of recurring events
 * [ [#1865](https://github.com/digitalfabrik/integreat-cms/issues/1865) ] Fix TextLab HIX widget for non-staff users
+* [ [#1788](https://github.com/digitalfabrik/integreat-cms/issues/1788) ] Fix broken translation status of events & locations if only minor public versions exists
```
timoludwig, line 15: Could you put this entry into the "unreleased" section?

timoludwig (changes requested):

CHANGELOG.md
JoeyStk: Oh shit, yes of course

timoludwig (approved):

MizukiTemma (approved):

timoludwig:
> When DRAFT event/POI is published, translation status is correct now ✅
> 
> However when status is still DRAFT, it shows "translation missing". Is it an expected behavior or another issue?

This was defined as expected behavior by the service team since all draft content is not visible in the app and is thus "missing".

seluianova:
@timoludwig looks a bit strange to me, because "missing" is more like "not exists". And to mark thing as "visible" there is a "published" status... But okay, thanks for the answer :)

timoludwig:
@seluianova yes, I agree, this probably needs to be revisited in the future together with the service team and UI/UX.

Issue #1494, opened 2022-05-25
Add a role without page editing permissions


dkehne:
### Describe the Bug
The page-pased feature is sth. we were already used to in WordPress. Users can be granted access to only single pages within the system. The feature worked as a restriction, so that only pages are allowed to be seen and edited, which are granted by the page-based-permission.

E.g. Mr. X from the Jobcenter only wants to see and edit the Jobcenter and Agentur für Arbeit-pages. So his user role will be Redaktion or Autor (depending on if he can directly publish) and then he will be granted to the two pages. Then he only sees these two pages and can edit them.

Currently the page-based feature is not combinable with the user roles and only use-ful in combination with Terminplaner, because the feature grants access to pages, where not access exists yet.

svenseeberg:
In order to use the per page editing permissions, we need a user role that has no page editing permissions in general. This role can then be assigned to users that should only be able to edit a few pages.

MizukiTemma:
Should the page-based permission be achived by organisation ferture or literally per-page independet of organisation?

For the example of Mr. X from the jobcenter, if he is added to organisation "Jobcenter" he gains permission to the pages which belongs to the organisation "Jobcenter".

timoludwig:
> Should the page-based permission be achived by organisation ferture or literally per-page independent of organisation?

Both should be possible - via the orga and via page-based-permissions. Both features are already implemented, but not really useful if all users already have the editing permissions. Thus, we want to add another role which is not allowed to edit pages which makes them automatically eligible to be used with the individual permissions levels.

MizukiTemma:
Ok. And I've found the per-page permission grant system 🙂

PR #1855, opened 2022-11-11
Add a role without page editing permission

MizukiTemma:
### Short description

This PR introduces a new user role which does not have any permission to edit page, POI or event in default.
Users of this role can edit only pages, which belong to their organisation or which are shared with them by per-page permission. 
Whether a user can only edit or also publish, it depends on the per-page permission. Permission per organization gives always publish permission too.
The new role is to be used for users who are/will be allowed to edit only some single pages. 

### Proposed changes

- add a new role <s>"Limited page manager" </s> "Observer" with no page editing permission in default
- add a new user of this role in test_data.json
- If the user is in this role, a message will be shown in the page list view which pages they are allowed to edit.

### Side effects

- <s>I don't come up with any good naming for this new role. "Limited_page_manager" is a bit long and not so self-explaining...... </s> 

### Resolved issues

Fixes: #1494


integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -1433,6 +1433,10 @@ msgstr "App-Team"
 msgid "Marketing team"
 msgstr "Marketing-Team"
 
+#: cms/constants/roles.py
+msgid "Limited page manager"
+msgstr "Limited page manager:in"
```
JoeyStk, line 1438: I spoke to @karthik-integreat and we came up with "Beobachter:in" or "Leser:in". The first one is prefered

integreat_cms/cms/constants/roles.py
```diff
@@ -57,6 +63,11 @@
     "view_poi",
 ]
 
+#: The permissions of the limited_page_manager
+LIMITED_PAGE_MANAGER_PERMISSIONS = EVENT_MANAGER_PERMISSIONS + [
+    "view_page",
```
JoeyStk, line 68: I'm not entirely sure about this. If I'm interpreting the issue correctly this is only about pages. So I don't think the limited page manager should inherit the permissions of the event manager. But maybe I'm incorrect about this

JoeyStk (changes requested):
Thank you so much for this PR. Please find two suggestions of mine below. I wasn't able to understand all the details. Hopefully someone else can give more insights here

timoludwig (changes requested):
Thanks a lot! :+1:
In general everything works as expected, I just have a few ideas to improve the constants and the warning message.

integreat_cms/cms/constants/roles.py
```diff
@@ -41,6 +43,10 @@
 #: Choices to use these constants in a database field
 CHOICES = ROLES + STAFF_ROLES
 
+EXTENDED_CHOICES = CHOICES + [
+    (LIMITED_PAGE_MANAGER, _("Limited page manager")),
+]
+
```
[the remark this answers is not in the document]
MizukiTemma: I did so to avoid errors in the old migration files without changing them.
If the new role is added to the constant ROLES, role update functions in migration files try an update for the new role, which didn't exist at the time and gives an error (I think that was None Type Object).

Is there a way to avoid this or is it okay to change the old migration files?

integreat_cms/cms/migrations/0052_alter_role_name.py
```diff
@@ -0,0 +1,70 @@
...
+        # Clear permissions
+        group.permissions.clear()
+        # Set permissions
+        group.permissions.add(
+            *Permission.objects.filter(codename__in=roles.PERMISSIONS[role_name])
+        )
+
+    # Create the new role Observer
+    observer, _ = Group.objects.get_or_create(name="OBSERVER")
+    Role.objects.get_or_create(name="OBSERVER", group=observer)
```
timoludwig, line 37:
```suggestion
```
The group is already created in the loop, right?
If not, the permissions would be missing.

integreat_cms/cms/constants/roles.py
```diff
@@ -57,6 +60,20 @@
...
 ]
 
+#: The permissions of the limited_page_manager
+OBSERVER_PERMISSIONS = [
+    "add_directory",
+    "change_directory",
+    "view_directory",
+    "change_mediafile",
+    "replace_mediafile",
+    "upload_mediafile",
```
timoludwig, line 70: Hmm, these permissions do not really fit the role name "observer", but ok, don't see much alternative to it :sweat_smile:

integreat_cms/cms/views/pages/page_tree_view.py
```diff
@@ -81,13 +85,33 @@ def get(self, request, *args, **kwargs):
...
+            access_granted_pages = Page.objects.filter(
+                Q(editors=request.user) | Q(publishers=request.user)
+            ).filter(region=request.region)
+            if request.user.organization:
+                access_granted_pages = access_granted_pages.union(
+                    Page.objects.filter(organization=request.user.organization)
+                )
+            if len(access_granted_pages) > 0:
+                page_list = ", ".join(
+                    f"<a href='{escape(translation.backend_edit_link)}' class='hover:underline'>{escape(translation.title)}</a>"
```
timoludwig: With my PR #1862 now merged, messages containing links have to be formatted by [translate_link](https://github.com/digitalfabrik/integreat-cms/blob/a5006d1c1f69a5d24a9cc4d8b424f9bc79844f2b/integreat_cms/cms/utils/translation_utils.py#L26) (if it's only one link) or [format_html](https://docs.djangoproject.com/en/4.1/ref/utils/#django.utils.html.format_html)/[format_html_join](https://docs.djangoproject.com/en/4.1/ref/utils/#django.utils.html.format_html_join) (when having multiple links), otherwise they will be escaped.

integreat_cms/cms/migrations/0025_update_roles.py
```diff
@@ -34,7 +34,7 @@ def update_roles(apps, schema_editor):
 
     # Assign the correct permissions
     for role_name in dict(roles.CHOICES):
-        group = Group.objects.filter(name=role_name).first()
+        group, _ = Group.objects.get_or_create(name=role_name)
```
timoludwig, line 37: Ah, stupid me that I have not thought of this case when creating those migrations. Still not feeling completely comfortable with changing the past migrations, but in this case I think the production server should be fine.

timoludwig (changes requested):
Thanks a lot! :+1:
Now to validate whether the new role really is not allowed to see/change things it cannot see, could you add it to the view tests and put it into the view test config for every view it can access?

integreat_cms/cms/views/pages/page_tree_view.py
```diff
@@ -81,13 +85,33 @@ def get(self, request, *args, **kwargs):
...
+            access_granted_pages = Page.objects.filter(
+                Q(editors=request.user) | Q(publishers=request.user)
+            ).filter(region=request.region)
+            if request.user.organization:
+                access_granted_pages = access_granted_pages.union(
+                    Page.objects.filter(organization=request.user.organization)
+                )
+            if len(access_granted_pages) > 0:
+                page_list = ", ".join(
+                    translation.title
```
timoludwig, line 101: I think it was cool showing the links to the edit forms, see e.g. an example here how it is done:
https://github.com/digitalfabrik/integreat-cms/blob/a5006d1c1f69a5d24a9cc4d8b424f9bc79844f2b/integreat_cms/cms/views/regions/region_actions.py#L47-L78

integreat_cms/cms/migrations/0052_alter_role_name.py
MizukiTemma: yes, the group is created in the loop before.

MizukiTemma: Yes, the groop is created in the loop.

integreat_cms/cms/views/pages/page_tree_view.py
MizukiTemma: Ah ok now I got it why the link in the messsage did not work from the certain moment.
I reverted it back.

tests/cms/views/view_config.py
```diff
@@ -161,10 +162,10 @@
             ),
             ("events", ROLES),
             ("events_archived", ROLES),
-            ("new_event", ROLES),
+            ("new_event", ROLES - [OBSERVER]),
```
timoludwig, line 165: Hmm, I find this syntax not ideal, could you add a new constant, maybe `WRITE_ROLES` or something with all roles which are allowed to write contents?

MizukiTemma: OK, I"ll do so :)

MizukiTemma:
Update

The new role Observer is added to the test and the test is adjusted.

integreat_cms/cms/views/pages/page_tree_view.py
```diff
@@ -81,13 +84,46 @@ def get(self, request, *args, **kwargs):
...
-            )
+            access_granted_pages = Page.objects.filter(
+                Q(editors=request.user) | Q(publishers=request.user)
+            ).filter(region=request.region)
+            if request.user.organization:
+                access_granted_pages = access_granted_pages.union(
+                    Page.objects.filter(organization=request.user.organization)
+                )
+            if len(access_granted_pages) > 0:
+                messages.error(
```
timoludwig, line 99:
```suggestion
                messages.info(
```
I wouldn't consider this an error

integreat_cms/cms/views/pages/page_tree_view.py
```diff
@@ -81,13 +84,46 @@ def get(self, request, *args, **kwargs):
...
+                                    translation.backend_edit_link,
+                                    translation.title,
+                                )
+                                for translation in [
+                                    page.get_translation(
+                                        request.region.default_language.slug
+                                    )
+                                    or page.best_translation
+                                    for page in access_granted_pages
+                                ]
```
timoludwig, line 118:
```suggestion
                                    page.best_translation.backend_edit_link,
                                    page.best_translation.title,
                                )
                                for page in access_granted_pages
```
I would directly prefer `best_translation` here - we specifically created this property for this use case, so if the user has the UI in English, the English translation is shown (if it exists), otherwise the default language is used as fallback.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ UNRELEASED
 * [ [#1786](https://github.com/digitalfabrik/integreat-cms/issues/1786) ] Remove textblock option in editor, add button to clear all formatting
 * [ [#1788](https://github.com/digitalfabrik/integreat-cms/issues/1788) ] Fix broken translation status of events & locations if only minor public versions exists
 * [ [#1688](https://github.com/digitalfabrik/integreat-cms/issues/1688) ] Specify protected pages when trying to delete regions
+* [ [#1494](https://github.com/digitalfabrik/integreat-cms/issues/1494) ] Add a role without page editing permissions
```
timoludwig, line 16: This needs to be moved into the "unreleased" section now

timoludwig (changes requested):
Thanks a lot! :pray:

Almost perfect, just a small detail in the message! :muscle:

timoludwig (approved):
:tada:

JoeyStk (approved):
Thank you so much
This looks good now :)

Issue #1786, opened 2022-10-24
Disable texblock option in editor


ulliholtgrave:
### Motivation

The textblock option creates various problems within the PDF generation process due to additional div elements. Therefore, It should be deactivated in the editor.

seluianova:
Could you please clarify what exactly you mean by "textblock option" ? 
Is it this menu item?
![image](https://user-images.githubusercontent.com/115008338/200934606-0a38e94e-162a-4bba-9020-0eb0bd15689c.png)

ulliholtgrave:
Exact. We would like to remove this "Blocks" section entirely.

PR #1861, opened 2022-11-14
Remove texblock option in editor, add button to clear formatting

seluianova:
### Short description
The textblock option creates various problems within the PDF generation process due to additional div elements. Therefore, It should be deactivated in the editor.

### Proposed changes
Remove "Blocks" format option from the editor.
Also I suggest to add "Clear formatting" button, as a convenient way to remove all applied formatting from an element.

### Side effects
Didn't find any

### Resolved issues
Fixes: #1786


charludo (commented):
Very nice, I really like the remove formatting option!

Minor: Not sure about removing blockquote along with div though, I could potentially see some usecases..? On the other hand, there doesn't seem to be any styling attached to blockquotes, so maybe they aren't used after all.

charludo (approved):

MizukiTemma (approved):
Thank you 👍 Works as expected!

timoludwig (approved):

timoludwig:
@seluianova could you resolve the conflict again? :slightly_smiling_face:

PR #1862, opened 2022-11-15
Correctly escape success/error messages

timoludwig:
### Short description

The assumption that our messages were `safe` was incorrect, since they also contain user input (e.g. when the title/name of an object is contained in the message).

### Proposed changes

- Escape the message per default by removing the [safe](https://docs.djangoproject.com/en/4.1/ref/templates/builtins/#std-templatefilter-safe) template filter
- Provide a util function `translate_link()` which can be used to render links in messages correct (even if they are escaped)

### Side effects

- Not sure if I forgot any messages which used HTML syntax... if so, the messages would now contain the escaped tags and the HTML code would be literally printed out.

### DIFF
--- a/integreat_cms/cms/templates/messages.html
+++ b/integreat_cms/cms/templates/messages.html
@@ -4,25 +4,25 @@
             {% if msg.level_tag == 'info' %}
                 <div class="bg-blue-100 border-l-4 border-blue-500 text-blue-500 px-4 py-3 mb-5"
                      role="alert">
-                    <p>{{ msg.message|safe }}</p>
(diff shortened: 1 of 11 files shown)


david-venhoff (approved):
LGTM

PeterNerlich (approved):
@membralala und ich haben uns die PR angesehen und keine direkt mit dem Code zusammenhängenden Probleme gefunden.

Folgendes ist uns aber aufgefallen, was möglicherweise mit diesen Änderungen zu tun haben könnte:
Screenshot aus der Region edit view. The save success message seems escaped one time too many.

Issue #1688, opened 2022-09-15
Specify protected pages when trying to delete regions


ulliholtgrave:
### Motivation

If someone tries to delete a region that has pages that are mirrored somewhere else Django throws a `ProtectedError` and our server returns a internal server error. As the user has no information about the reason and the conflicting pages we should return an error message that contains the reason and the affected pages.

### Additional Context

For reproduction purpose you can mirror a page from a region that you try to delete afterwards. You will then get a 501 and see the stack trace in the console.

PR #1863, opened 2022-11-15
Specify protected pages when trying to delete regions

JoeyStk:
### Short description
It's not possible to delete a region of which pages are mirrored in other regions. However, up until now there was only a Django default error and there wasn't an informative output to the user

### Proposed changes

- Give warning to user why deleting the region didn't work 
- Output which pages are causing the issue

- I thought maybe it would also be interesting to know in which region the page is mirrored to. But since this wasn't in the issue I didn't do that yet. What do you think? Do I need to implement this? 

### Side effects

- I mean there has to be a simpler way to do this than what I did. I don't know why but I really struggled to make this more simple. I would be very interested to hear your ideas on how to refactor this

### Resolved issues
Fixes: #1688

### DIFF
--- a/integreat_cms/cms/views/regions/region_actions.py
+++ b/integreat_cms/cms/views/regions/region_actions.py
@@ -6,11 +6,12 @@ import logging
 from django.contrib import messages
 from django.contrib.auth import get_user_model
 from django.shortcuts import get_object_or_404, redirect
+from django.utils.html import format_html, format_html_join
 from django.utils.translation import ugettext as _
 from django.views.decorators.http import require_POST
 
 from ...decorators import permission_required
-from ...models import Region
+from ...models import Region, Page
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


integreat_cms/cms/views/regions/region_actions.py
```diff
@@ -45,25 +46,42 @@ def delete_region(request, *args, **kwargs):
...
+        for page in mirrored_pages:
+            if page is not None:
+                if page.region == region:
+                    p = p + str(page)
+        messages.error(
+            request,
+            _(
+                "Region could not be deleted, because these pages are {} mirrored in other regions"
+            ).format(p),
+        )
```
charludo, line 85: I think this can be simplified:

```suggestion
    except ProtectedError:
        pages = Page.objects.filter(region=region)
        messages.error(
            request,
            _(
                "Region could not be deleted, because the following page(s) are mirrored in other regions: {}"
            ).format(", ".join([str(page) for page in pages])),
        )
```

charludo (changes requested):
The error reporting for protected pages can be simplified. Other than that though, looks good to me!

integreat_cms/cms/views/regions/region_actions.py
timoludwig: I agree on using the `join` instead of the loop, but this lists just all pages of the region and not those which are embedded in other regions as live content...

integreat_cms/cms/views/regions/region_actions.py
```diff
@@ -37,6 +38,51 @@ def delete_region(request, *args, **kwargs):
     """
 
     region = get_object_or_404(Region, slug=kwargs.get("slug"))
+
+    # Check whether region can be safely deleted
+    mirrored_pages = Page.objects.filter(
+        region=region, mirroring_pages__isnull=False
+    ).distinct()
+    if len(mirrored_pages) > 0:
```
timoludwig, line 46: Since a lot of database operations are performed before the Protection error occurs, I would suggest to check in advance whether the deletion can be executed. Otherwise the region would still exist but be in an inconsistent state (e.g. the complete page hierarchy would be lost, the media library etc...)

integreat_cms/cms/views/regions/region_actions.py
```diff
@@ -37,6 +38,51 @@ def delete_region(request, *args, **kwargs):
...
+                            format_html_join(
+                                ", ",
+                                "<a href='{}' class='underline hover:no-underline'>{}</a>",
+                                [
+                                    (
+                                        mirrored_page.best_translation.backend_edit_link,
+                                        mirrored_page.best_translation.title,
+                                    )
+                                    for mirrored_page in page.mirroring_pages.all()
+                                ],
```
timoludwig, line 71: I would suggest to link all pages that mirror the listed page so the service team can directly remove any mirrored relationships if needed.
Just listing the mirrored pages is not very helpful if we don't know where they are embedded from.

timoludwig (approved):
Thanks! :+1:

A few aspects:
1. Just handling the exception is a bad idea if other deletion-related operations are executed before the error occurs.
2. Assembling strings in loops can often times be replaced by `join` as @charludo pointed out
3. We can retrieve all mirrored pages in one database query instead of looping through the pages
4. I would add links to the page forms to make it easier to change any mirroring relationships if needed

Since my suggestions were kind of complex, I took the liberty of directly committing them :innocent: 
Maybe someone can double-check again if I messed something up? :laughing:

integreat_cms/cms/views/regions/region_actions.py
charludo: Ah, you're absolutely right, I misunderstood what was being done & had an unfortunate test case.

ulliholtgrave:
@JoeyStk Thanks, for your effort 🙏  . I guess you can see the technical feedback also in the merged PR :)

Issue #1818, opened 2022-11-01
Broken Link Checker stops when facing URL longer than 255 chars


svenseeberg:
### Describe the Bug
It seems that the broken link check LIFO queue is not being checked to the end if an error occurs. That means that if many links are checked at once (queue is being filled with many links) and one error occurs, many links are not being checked. The queue seems to live in memory and not in the database.

### Steps to Reproduce

1. I replaced links with the shell as described in https://issues.tuerantuer.org/browse/INFRA-217
2. After all links have been replaced, the link checker returns some warnings which indicate it is running
3. After some time, an error occurs and no further warnings appear. This indicates that the link validation process has stopped

### Expected Behavior
Errors should either be catched and the LIFO queue should normally be finished. Alternatively, the queue could be persisted, for example in the database, and continued with the next cron job.

### Actual Behavior
One error stops the link check process.

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
[...]
/opt/integreat-cms/.venv/lib/python3.9/site-packages/urllib3/connectionpool.py:1045: InsecureRequestWarning: Unverified HTTPS request is being made to host 'integreat.app'. Adding certificate verification is str
ongly advised. See: https://urllib3.readthedocs.io/en/1.26.x/advanced-usage.html#ssl-warnings
  warnings.warn(
/opt/integreat-cms/.venv/lib/python3.9/site-packages/urllib3/connectionpool.py:1045: InsecureRequestWarning: Unverified HTTPS request is being made to host 'integreat.app'. Adding certificate verification is str
ongly advised. See: https://urllib3.readthedocs.io/en/1.26.x/advanced-usage.html#ssl-warnings
  warnings.warn(
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/db/models/query.py", line 581, in get_or_create
    return self.get(**kwargs), False
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/cacheops/query.py", line 353, in get
    return qs._no_monkey.get(qs, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/db/models/query.py", line 435, in get
    raise self.model.DoesNotExist(
linkcheck.models.DoesNotExist: Url matching query does not exist.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #1873, opened 2022-11-18
Increase max URL length for linkchecker

timoludwig:
### Short description

Increase max URL length for linkchecker

### Proposed changes

- Increase the length from currently 255 characters to 1024

### Side effects

This results in a migration of the linkcheck app. Probably we need to manually execute
```
integreat-cms-cli makemigrations linkcheck && integreat-cms-cli migrate
```
on all our test & prod systems...

### Resolved issues

Part of: #1818


ulliholtgrave (approved):
Good catch, but wouldn't it be necessary to add a migration for all the linkcheck objects? Or do we want to adjust it on the production database?

timoludwig:
> Good catch, but wouldn't it be necessary to add a migration for all the linkcheck objects? Or do we want to adjust it on the production database?

The problem is that we cannot add migrations for models which don't belong to our own app...
If I add a migration file for this, it tries to find the `Url` model in the `cms` app which obviously doesn't work... so unfortunately I see no other way than generating the migrations on the prod systems :cry:

ulliholtgrave:
I see. Thanks for the explanation. I guess the PR is good to go then 👍

svenseeberg:
But if we add a manual migration, would that not create issues for all future app releases?

ulliholtgrave:
@svenseeberg I guess it won't be an issue as the linkcheck app is an separate app and after this PR it should generate new link URL objects with the newly defined field length

svenseeberg:
But if the app releases an additional migration themselves, we would be required to run the migration merge command, don't we? Django will not be able to sort out the migration conflicts itself.

timoludwig:
> But if we add a manual migration, would that not create issues for all future django-linkcheck app releases that include migrations themselves?

oh yes, you're right. This will generate conflicts which will be very cumbersome to fix :grimacing:

ulliholtgrave:
Wouldn't it be better then to delete all the links and create the objects again with the new defined variable?

svenseeberg:
> oh yes, you're right. This will generate conflicts which will be very cumbersome to fix grimacing

I think it would be okay to do this. We just need to make sure that the production CMS is not being killed by our state.highstate and an incomplete migration process. I'm not really sure how we can prevent this - we would have to check for new migrations in advance and throw errors?

timoludwig:
> Wouldn't it be better then to delete all the links and create the objects again with the new defined variable?

The problems aren't the existing database objects (they are all smaller than the current max size), but the database schema.

svenseeberg:
> we would have to check for new migrations in advance and throw errors?

I think we can add a pseudo state in our Salt config that uses https://docs.github.com/en/rest/repos/contents#get-repository-content to check for new migrations. We can keep track of fixed migrations in our Pillar?

timoludwig:
Ok, I remembered that we did something similar with the auth app for the lunes project [a while ago](https://github.com/digitalfabrik/lunes-cms/commit/c6196980bd8fe5b6aac1796b393b1d3b3ab9a6bc#diff-32088c2d908f60a4bb90b97ebc7fa6dfc370c8142773975c4f609a31d0d5b680)... I guess the same workaround should do the trick here as well. Looks ugly but is probably a bit safer and less work than the check in salt.

svenseeberg:
> Ok, I remembered that we did something similar with the auth app for the lunes project [a while ago](https://github.com/digitalfabrik/lunes-cms/commit/c6196980bd8fe5b6aac1796b393b1d3b3ab9a6bc#diff-32088c2d908f60a4bb90b97ebc7fa6dfc370c8142773975c4f609a31d0d5b680)... I guess the same workaround should do the trick here as well. Looks ugly but is probably a bit safer and less work than the check in salt.

I think that looks like a good solution :)

ulliholtgrave:
I can't really test it with the current state of my local db, but I would trust you that this does the trick 👍 :)

svenseeberg:
> I can't really test it with the current state of my local db, but I would trust you that this does the trick +1 :)

I tested the migration. It works. I did not, however, create a test setup with migrations within the linkcheck app.

Issue #1132, opened 2022-01-25
Add support for TOTP 2FA


timoludwig:
### Motivation

At the moment, we only support 2FA via U2F/FIDO2.

### Proposed Solution

Since this requires a separate hardware token and a bit of technical knowledge, I suggest to add support for the more established TOTP method via e.g. Google Authenticator App.

PR #1884, opened 2022-11-21
Add TOTP authentication 

ulliholtgrave:
### Short description

This PR adds TOTP as another second factor for 2FA. It also adds another passwordless authentication route that only requires the user to authenticate via FIDO2 or TOTP.

### Proposed changes

- Add TOTP registration and authentication view and methods
- Add passwordless authentication routine

### Side effects

- I had to implement another PasswordAuthenticationView to handle the authentication for the passwordless variant which might be a little bit overkill
- After my study I would suggest to deactivate TOTP for passwordless but keep the FIDO2 support.

### Resolved issues

Fixes: #1132 


integreat_cms/cms/views/settings/totp_delete_view.py
```diff
@@ -0,0 +1,73 @@
...
+
+        user.totp_key = None
+        user.save()
+        messages.success(
+            request,
+            _(
+                "You have successfully disconnected your account from your authentication app."
+            ),
+        )
+        return redirect("user_settings", **kwargs)
```
timoludwig, line 73:
```suggestion
        kwargs = (
            {"region_slug": self.request.region.slug} if self.request.region else {}
        )
        return redirect("user_settings", **kwargs)
```
Here we should probably be redirect to the user settings in a specific region for non-staff-users...

timoludwig (changes requested):
Thanks a lot, really cool feature :heart_eyes: 

In general everything works as expected, just a few implementation details I would like to change.

integreat_cms/cms/templates/settings/user_settings.html
```diff
@@ -81,50 +81,84 @@ <h3 class="heading font-bold text-black">
...
+                    <i icon-name="shield" class="mr-2"></i> {% trans "My 2-Factor-Authentication methods" %}
                 </h3>
             </div>
             <div class="p-4">
-                {% trans "You can use your FIDO2 keys to secure your account. Once you added a key to your account you won't be able to log in without using the key." %}
+                {% trans "To further protect your account you can add additional authentication methods (2-factor authentication). The system currently supports the setup of FIDO2 keys and authentication via authentication apps. Once you add a method, it will always be requested when logging in and potential attackers will have a harder time gaining control over your account." %}
                 <br />
-                {% trans "To make sure that you don't loose access to your account when you loose your FIDO2 key it is recommended to add multiple keys. Just make sure to keep at least one key at a secure place." %}
+                <strong>{% trans "Warning: It is recommended not to activate only one method, but to add multiple items so that you do not lock yourself out in case of loss." %}</strong>
```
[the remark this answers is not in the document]
ulliholtgrave: I removed the sentence and opened #1902 for a recovery

integreat_cms/cms/utils/totp_utils.py
```diff
@@ -0,0 +1,92 @@
...
+import base64
+import io
+import datetime
+import math
+import hmac
+import hashlib
+
+from datetime import datetime, timedelta
+
+import qrcode
```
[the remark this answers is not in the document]
ulliholtgrave: I would stick with the qrcode library as it's rather small and [django-qr-code](https://pypi.org/project/django-qr-code/) introduces about 4 new dependencies. However, I added it to the Pipfile as an explicit dependency

integreat_cms/cms/views/settings/totp_register_view.py
```diff
@@ -0,0 +1,93 @@
...
+        :param \*args: The supplied arguments
+        :type \*args: list
+
+        :param \**kwargs: The supplied keyword arguments
+        :type \**kwargs: dict
+
+        :return: A redirection to the account settings
+        :rtype: ~django.http.HttpResponseRedirect
+        """
+
```
timoludwig, line 41:
```suggestion
        if request.user.totp_key:
            messages.error(
                request,
                _(
                    "You have already registered a TOTP key. Please disconnect your app before adding a new one."
                ),
            )
            kwargs = {"region_slug": request.region.slug} if request.region else {}
            return redirect("user_settings", **kwargs)
```
I guess we could make sure a key is not already configured to prevent accidental overwriting of existing keys...

integreat_cms/cms/templates/authentication/delete_totp.html
```diff
@@ -0,0 +1,20 @@
+{% extends "authentication/_base.html" %}
+{% load i18n %}
+{% block heading %}
+    {% trans "Add Authentication-App" %}
+{% endblock heading %}
+{% block content %}
+    <div class="mb-3">
+        {% trans "You are about to disconnect your authenticator app from your account and deactivate TOTP authentication. Are you sure about that?" %}
+    </div>
+    <form method="post" action="{% url 'delete_totp' %}">
```
timoludwig, line 10:
```suggestion
    <form method="post">
```
This action also has to be removed to allow non-staff users to disconnect their apps.

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -7542,6 +7690,22 @@ msgstr ""
 "Der 2-Faktor-Authentifizierungsschlüssel \"{}\" wurde erfolgreich "
 "registriert."
 
+#: cms/views/settings/totp_delete_view.py
+msgid ""
+"You have successfully disconnected your account from your authentication app."
+msgstr ""
+"Sie haben erfolgreich Ihr Konto mit der Authentifizierungs-App getrennt."
```
timoludwig, line 7697:
```suggestion
"Sie haben Ihr Konto erfolgreich von der Authenticator-App getrennt."
```

integreat_cms/cms/templates/authentication/register_totp.html
```diff
@@ -0,0 +1,32 @@
...
+    <p class="mb-3">
+        {% blocktrans trimmed with BRANDING=BRANDING|capfirst %}
+            Please enter the 6-digit code for "{{ BRANDING }}" to  confirm the registration:
+        {% endblocktrans %}
+    </p>
+    <form method="post"
+          action="{% url 'register_totp' %}"
+          class="flex flex-nowrap gap-4">
+        {% csrf_token %}
+        <input id="totp"
```
timoludwig, line 24:
```suggestion
        <input autocomplete="false"
               id="totp"
```

integreat_cms/cms/templates/authentication/register_totp.html
```diff
@@ -0,0 +1,32 @@
...
+               name="totp"
+               type="text"
+               minlength="6"
+               maxlength="6"
+               pattern="\d{6}"/>
```
timoludwig, line 29:
```suggestion
               pattern="\d{6}"
               autofocus/>
```

integreat_cms/cms/templates/authentication/totp_login.html
```diff
@@ -0,0 +1,27 @@
...
+    <form method="post"
+          action="{% url 'public:login_totp' %}"
+          class="flex flex-nowrap gap-4">
+        {% csrf_token %}
+        <input id="totp_code"
+               name="totp_code"
+               type="text"
+               minlength="6"
+               maxlength="6"
+               pattern="\d{6}"/>
```
timoludwig, line 24:
```suggestion
               pattern="\d{6}"
               autofocus/>
```

integreat_cms/cms/templates/authentication/totp_login.html
```diff
@@ -0,0 +1,27 @@
...
+    <div class="mb-3">
+        {% blocktrans trimmed with BRANDING=BRANDING|capfirst %}
+            Please open your authenticator app on your device and enter the 6-digit code for "{{ BRANDING }}" to verify your identity:
+        {% endblocktrans %}
+    </div>
+    <form method="post"
+          action="{% url 'public:login_totp' %}"
+          class="flex flex-nowrap gap-4">
+        {% csrf_token %}
+        <input id="totp_code"
```
timoludwig, line 19:
```suggestion
        <input autocomplete="false"
               id="totp_code"
```

integreat_cms/cms/templates/authentication/totp_login.html
```diff
@@ -0,0 +1,27 @@
...
+    <i icon-name="smartphone"></i>
+    {% trans "Authentication" %}
+{% endblock heading %}
+{% block content %}
+    <div class="mb-3">
+        {% blocktrans trimmed with BRANDING=BRANDING|capfirst %}
+            Please open your authenticator app on your device and enter the 6-digit code for "{{ BRANDING }}" to verify your identity:
+        {% endblocktrans %}
+    </div>
+    <form method="post"
```
timoludwig, line 15:
```suggestion
    <form method="post"
          autocomplete="off"
```

integreat_cms/cms/templates/authentication/register_totp.html
```diff
@@ -0,0 +1,32 @@
...
+         alt="QR code for TOTP registration"
+         src="{{ qr }}"/>
+    <p class="mb-3">
+        {% blocktrans trimmed with BRANDING=BRANDING|capfirst %}
+            Please enter the 6-digit code for "{{ BRANDING }}" to  confirm the registration:
+        {% endblocktrans %}
+    </p>
+    <form method="post"
+          action="{% url 'register_totp' %}"
+          class="flex flex-nowrap gap-4">
```
timoludwig, line 22:
```suggestion
    <form method="post" autocomplete="off" class="flex flex-nowrap gap-4">
```
This url is only accessible for staff users, so I guess it's better to just POST to the same url which is currently opened (which should already have the region slug for region users).
Also, I guess disabling autocomplete probably makes sense for OTPs...

integreat_cms/cms/urls/protected.py
```diff
@@ -139,6 +139,30 @@
...
+                            ),
+                        ]
+                    ),
+                ),
+                path(
+                    "totp/",
+                    include(
+                        [
+                            path(
+                                "delete",
```
timoludwig, line 159:
```suggestion
                                "delete/",
```
I would prefer URLs ending with a trailing slash

integreat_cms/cms/urls/public.py
```diff
@@ -62,6 +62,18 @@
                     authentication.LoginView.as_view(extra_context=auth_context),
                     name="login",
                 ),
+                path(
+                    "passwordless",
```
timoludwig, line 66:
```suggestion
                    "passwordless/",
```

integreat_cms/cms/urls/public.py
```diff
@@ -62,6 +62,18 @@
...
+                    authentication.PasswordlessLoginView.as_view(
+                        extra_context=auth_context
+                    ),
+                    name="passwordless_login",
+                ),
+                path(
+                    "totp",
```
timoludwig, line 73:
```suggestion
                    "totp/",
```

integreat_cms/cms/models/users/user.py
```diff
@@ -103,6 +104,22 @@ class User(AbstractUser, AbstractBaseModel):
...
         ),
     )
+    totp_key = models.CharField(
+        default=None,
+        null=True,
+        blank=True,
+        max_length=128,
+        verbose_name=_("TOTP key"),
+        help_text=_("Will be used to generate TOTP keys"),
+        validators=[MinLengthValidator(64)],
```
timoludwig, line 114: Hmm well ok, I guess my idea with the validator didn't make sense since we're never actually validating this value.
We're generating keys of length 32 now, so this would be always invalid, but since we're not calling `full_clean()`, it does not check anything... I guess then we can also remove the check since it creates a deceptive impression of security.
```suggestion
```

integreat_cms/cms/utils/totp_utils.py
```diff
@@ -0,0 +1,59 @@
...
+
+    :param given_totp_code: The entered TOTP code form the user
+    :type given_totp_code: str
+
+    :param encoded_key: The key that is used for the generation of the true code
+    :type encoded_key: str
+
+    :return: True if the given code matches the initiated
+    :rtype: bool
+    """
```
timoludwig, line 56:
```suggestion
    """
    if not key:
        return False
```
I know, this is probably overly cautious, but since an empty string is a valid totp seed according to the spec, I would just avoid these complications altogether...

timoludwig (changes requested):

Now there isn't much left to do, mainly problems with the url scope of region users, but the rest are just minor details.

I can't wait for this feature to get merged :tada:

integreat_cms/cms/models/users/user.py
```diff
@@ -80,7 +81,7 @@ class User(AbstractUser, AbstractBaseModel):
         related_name="members",
         verbose_name=_("organization"),
         help_text=_(
-            "This allows the user to edit and publish all pages for which the organisation is registered as the responsible organisation"
+            "This allows the user to edit and publish all pages for which the organization is registered as the responsible organization"
```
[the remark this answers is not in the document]
ulliholtgrave: Removed it and I agree that we should solve this inconsistencies all together at a later point :+1:

integreat_cms/cms/views/settings/totp_delete_view.py
```diff
@@ -0,0 +1,73 @@
...
+
+        :param \**kwargs: The supplied keyword arguments
+        :type \**kwargs: dict
+
+        :return: A redirection to the account settings
+        :rtype: ~django.http.HttpResponseRedirect
+        """
+
+        user = request.user
+
```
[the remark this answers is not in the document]
ulliholtgrave: Yeah, sorry for the late response. I'll open a separate issue for this...

integreat_cms/cms/templates/authentication/delete_totp.html
```diff
@@ -0,0 +1,20 @@
+{% extends "authentication/_base.html" %}
```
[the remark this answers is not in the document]
ulliholtgrave: You are right, I adjusted the location of the files

integreat_cms/cms/forms/users/user_form.py
```diff
@@ -111,6 +112,12 @@ def __init__(self, **kwargs):
             self.fields["is_staff"].label = _("Integreat team member")
         self.fields["email"].required = True
 
+        # Check if passwordless authentication is possible for the user
+        # if self.instance.totp_key is None and not self.instance.mfa_keys.exists():
```
ulliholtgrave, line 116: @timoludwig Hiermit kann ich zumindest client-seitig auch nochmal eine extra Validierung hinzufügen, aber wenn ich das auskommentiere, failen eine Reihe von Tests, was ich mich nicht ganz erklären kann :thinking:

timoludwig: This is because the region user form inherits from the user form, but with different fields... meaning that the same `__init__()` method is called, but accessing a non-existing field results in a key error. So I'd suggest to check whether the field really exists before making it disabled.

integreat_cms/cms/forms/users/user_form.py
```diff
@@ -111,6 +112,16 @@ def __init__(self, **kwargs):
             self.fields["is_staff"].label = _("Integreat team member")
         self.fields["email"].required = True
 
+        # Check if passwordless authentication is possible for the user
+        if (
+            "passwordless_authentication_enabled" in self.fields
```
timoludwig, line 117: I think this part was missing

integreat_cms/cms/views/users/user_form_view.py
```diff
@@ -102,6 +102,14 @@ def post(self, request, *args, **kwargs):
...
                 _("Superuser permissions need to be set by another superuser."),
             )
+        elif (
+            not request.user.is_superuser
+            and "passwordless_authentication_enabled" in user_form.changed_data
+        ):
+            messages.error(
+                request,
+                _("Only superusers can enable or disable passwordless authentication."),
+            )
```
timoludwig, line 112: I moved this check into the view - this way, we can access `request.user`.

setup.cfg
```diff
@@ -63,8 +63,10 @@ install_requires =
 	lxml
 	Pillow
 	psycopg2-binary
+	pyotp
 	python-dateutil
 	python-magic
+	qrcode
```
timoludwig, line 69: Since these dependencies are needed on the prod server, it is not sufficient to add them to the Pipfile, we need to add them to setup.cfg as well.

timoludwig (approved):
Awesome, thanks a lot for your changes! :muscle: 

I just adjusted the validation step in the user form and improved a few translations.

From my perspective this is ready to be merged, really cool feature! :tada:

PR #1888, opened 2022-11-22
Process test_path argument in test.sh

seluianova:
### Short description
To make it easier to maintain/develop new tests, I suggest adding the ability to pass test_path as an argument to test.sh script.
So that only particular tests can be run like this:
./test.sh tests/api/test_api_feedback.py

### Proposed changes
* Pass TEST_PATH argument to "pipenv run pytest" command
* Generate code coverage report only when all tests are run. Not sure, if it makes sense otherwise.

### Side effects
Didn't find any

### Resolved issues
N/A

### DIFF
--- a/dev-tools/_functions.sh
+++ b/dev-tools/_functions.sh
@@ -185,7 +185,7 @@ function ensure_not_root {
 function ensure_root {
     # Check if script is not running as root
     if ! [[ $(id -u) == 0 ]]; then
-        echo "The script ${SCRIPT_NAME} needs root privileges to connect to the docker deamon. It is be automatically restarted with sudo." | print_warning
+        echo "The script ${SCRIPT_NAME} needs root privileges to connect to the docker daemon. It will automatically be restarted with sudo." | print_warning
         # Call this script again as root (pass -E because we want the user's environment, not root's)
         sudo --preserve-env=HOME,PATH env "${SCRIPT_PATH}" "${SCRIPT_ARGS[@]}"
         # Exit with code of subprocess
--- a/dev-tools/test.sh
+++ b/dev-tools/test.sh
@@ -6,8 +6,9 @@
 # shellcheck source=./dev-tools/_functions.sh
 source "$(dirname "${BASH_SOURCE[0]}")/_functions.sh"
 
-# Delete outdated code coverage
-rm -rf "${BASE_DIR:?}/htmlcov/"
+# Delete outdated code coverage report
+CODE_COVERAGE_DIR="${BASE_DIR:?}/htmlcov"
+rm -rf "${CODE_COVERAGE_DIR}"
 
 require_installed
 require_database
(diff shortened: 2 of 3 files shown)


dev-tools/_functions.sh
```diff
@@ -185,7 +185,7 @@ function ensure_not_root {
 function ensure_root {
     # Check if script is not running as root
     if ! [[ $(id -u) == 0 ]]; then
-        echo "The script ${SCRIPT_NAME} needs root privileges to connect to the docker deamon. It is be automatically restarted with sudo." | print_warning
+        echo "The script ${SCRIPT_NAME} needs root privileges to connect to the docker deamon. It will be automatically restarted with sudo." | print_warning
```
charludo, line 188:
```suggestion
        echo "The script ${SCRIPT_NAME} needs root privileges to connect to the docker daemon. It will automatically be restarted with sudo." | print_warning
```
(Not your typo, but I just saw that codespell is complaining)

dev-tools/test.sh
```diff
@@ -6,17 +6,28 @@
...
+    deescalate_privileges pipenv run pytest "$TEST_PATH" --disable-warnings --quiet --numprocesses=auto
+fi
+
 echo "✔ Tests successfully completed " | print_success
 
-echo -e "Open the following file in your browser to view the test coverage:\n" | print_info
-echo -e "\tfile://${BASE_DIR}/htmlcov/index.html\n" | print_bold
+if [ -d "$CODE_COVERAGE_DIR" ]; then
+    echo -e "Open the following file in your browser to view the test coverage:\n" | print_info
+    echo -e "\tfile://${CODE_COVERAGE_DIR}/index.html\n" | print_bold
```
charludo, line 32: Minor: at least in my case, the path looks like `file:///home/[...]/htmlcov//index.html` (with two slashes at the end). IDK if that happens on all platforms though.

charludo (approved):
Very nice, this is really useful.

I know it's way out of the scope of this PR, but maybe we could think about using something like [pytest-testmon](https://pypi.org/project/pytest-testmon/) in the future and add a `--changed` flag to the script?

Since right now it's not possible to exclusively test changed files sitting in different directories.

david-venhoff (approved):
Thanks, Looks good to me!

dev-tools/test.sh
seluianova: Ah, it's my mistake that I didn't notice. Thanks

dev-tools/test.sh
```diff
@@ -6,17 +6,28 @@
...
 export INTEGREAT_CMS_SUMM_AI_API_KEY="dummy"
 
-deescalate_privileges pipenv run pytest --disable-warnings --quiet --numprocesses=auto --cov=integreat_cms --cov-report html
+# If only particular tests should be run, test path can be passed as CLI argument
+TEST_PATH=$1;
+
+if [[ -z "$TEST_PATH" ]]; then
+    deescalate_privileges pipenv run pytest --disable-warnings --quiet --numprocesses=auto --cov=integreat_cms --cov-report html
+else
+    deescalate_privileges pipenv run pytest "$TEST_PATH" --disable-warnings --quiet --numprocesses=auto
```
timoludwig, line 25:
```suggestion
    echo -e "Running all tests..." | print_info
    deescalate_privileges pipenv run pytest --disable-warnings --quiet --numprocesses=auto --cov=integreat_cms --cov-report html
elif [[ -e "$TEST_PATH" ]]; then
    echo -e "Running tests in ${TEST_PATH}..." | print_info
    deescalate_privileges pipenv run pytest "$TEST_PATH" --quiet --disable-warnings --numprocesses=auto
else
    echo -e "${TEST_PATH}: No such file or directory" | print_error
    exit 1
```
Maybe we could show an error if the given path does not exist?
Otherwise the output is just
```
bringing up nodes...

11985 warnings in 5.42s
```
without any error or success message, which could be misinterpreted as "nothing went wrong"...

timoludwig (changes requested):
Thanks a lot, I wanted to do this for a long time :smile: 

Could you also add the new argument to the [documentation](https://github.com/digitalfabrik/integreat-cms/blob/develop/sphinx/dev-tools.rst#testing)?

> I know it's way out of the scope of this PR, but maybe we could think about using something like [pytest-testmon](https://pypi.org/project/pytest-testmon/) in the future and add a `--changed` flag to the script?

@charludo Yes, cool idea! Could you open an issue for this?

dev-tools/test.sh
seluianova: Agree

sphinx/dev-tools.rst
```diff
@@ -84,7 +84,11 @@ Testing
 
 Run tests and generate coverage report with :github-source:`dev-tools/test.sh`::
 
-    ./dev-tools/test.sh
+    ./dev-tools/test.sh [test_path]
+
+**Arguments:**
+
+* ``test_path``: Run only tests from the given path
```
timoludwig, line 91:
```suggestion
    ./dev-tools/test.sh [TEST_PATH]

**Arguments:**

* ``TEST_PATH``: Run only tests in ``TEST_PATH``
```
I guess for "variable" arguments, there are two conventions, either angular brackets `<var>` or upper case `VAR`, to differentiate them from the literal argument `var` (see e.g. http://docopt.org/)...

seluianova: Probably [<test_path>] would be better actually...
I only see the use of uppercase for the argument of option 
`--input=ARG`

timoludwig: Well, I think both are fine, e.g. the Django CLI also uses upper case:
```
pipenv run integreat-cms-cli shell -h
usage: integreat-cms-cli shell [-h] [--no-startup] [-i {ipython,bpython,python}] [-c COMMAND]
                               [--version] [-v {0,1,2,3}] [--settings SETTINGS]
                               [--pythonpath PYTHONPATH] [--traceback] [--no-color]
                               [--force-color]
```
Choose whichever you prefer :smile:

seluianova: `-c COMMAND`
in your example it is also the argument of an option. but our TEST_PATH is a positional argument :)

PR #1891, opened 2022-11-22
Only run translation hook if python or HTML files were modified

timoludwig:
### Short description

The checking script should only run when there actually something to check.

### Proposed changes

- Only run translation hook if python or HTML files were modified

### Side effects

- Faster execution times of the pre commit hooks if no Python and HTML files are modified


charludo (approved):
LGTM

ulliholtgrave (approved):
Thanks :+1:

Issue #1895, opened 2022-11-23
Add option to only test changed files


charludo:
### Motivation

Currently we can either run all tests, or those in a specific directory (see #1888). However it would be nice to have the option to run all tests (and only those) that are affected by un-commited changes. 

### Proposed Solution

[pytest-testmon](https://pypi.org/project/pytest-testmon/) offers this functionality. Add a `--changed` flag to our test script which runs pytest with testmon support. Maybe also make this flag play nice with the existing folder-specific solution.

### Alternatives

There may be alternative frameworks out there. This one just seems popular and well-documented.

### Additional Context

The solution would probably not be perfect, since testmon can only monitor files/branches with existing test coverage.

### Design Requirements

n/a

PR #1897, opened 2022-11-23
Add option to only test changed files

charludo:
### Short description

This PR adds the `--changed` flag to our testing devtool. If passed, only files that have recently been changed will be tested. Plays nicely with the `TEST_PATH` argument added in #1888.

### Proposed changes

- add `pytest-testmon` as a dependency
- if `--changed` is passed to the script, only test recently changed files

### Side effects

- none

### Resolved issues

Fixes: #1895 


Gaston69:
Peter and me say: 
everything works fine :-)
We made a benchmark: 

Ingos Laptop t470s, postgres native: 
real	5m40,563s
user	8m50,080s
sys	0m31,970s

running 4 tests
real	1m19,451s
user	1m48,395s
sys	0m5,579s

Peters Laptop n240bu/ 4core, postgres in container: 
all tests:
real	3m45.063s
user	7m47.509s
sys	0m29.674s

4 tests in one file
real	0m52.908s
user	1m28.823s
sys	0m4.344s

Gaston69 (approved):

PeterNerlich:
One afterthought: Should `.testmondata` be added to the `.gitignore`?

charludo:
> One afterthought: Should `.testmondata` be added to the `.gitignore`?

Yes, since its content is dependent on your system/where you store the project, afaik.

dev-tools/test.sh
```diff
@@ -16,7 +16,8 @@ require_database
 # Set dummy key to enable SUMM.AI during testing
 export INTEGREAT_CMS_SUMM_AI_API_KEY="dummy"
 
-while test $# -gt 0; do
+# Parse given commend line arguments
+while [[ $# -gt 0 ]]; do
```
timoludwig, line 20: Since we're also using the extended test command in other places, we could also use it here for consistency reasons

dev-tools/test.sh
```diff
@@ -25,38 +26,51 @@ while test $# -gt 0; do
...
+        PYTEST_ARGS+=("--testmon-nocollect")
+        CHANGED_MESSAGE=" affected by recent changes"
+    else
+        # Inform that all tests will be run
         echo -e "\nIt looks like you have not run pytest without the \"--changed\" flag before." | print_warning
         echo -e "Pytest has to build a dependency database by running all tests without the flag once.\n" | print_warning
+        # Override test path argument
         unset TEST_PATH
-        TESTMON="--testmon"
-        unset CHANGED_MESSAGE
```
timoludwig: Instead of always setting the message and unsetting it in the if case, I directly converted it to an if/else

dev-tools/test.sh
```diff
@@ -25,38 +26,51 @@ while test $# -gt 0; do
     esac
 done
 
-if [[ $CHANGED -gt 0 ]]; then
-    # if --changed flag was passed: only run changed tests and don't update dependency database
-    TESTMON="--testmon-nocollect"
-    CHANGED_MESSAGE=" affected by recent changes"
+# The default pytests args we use
+PYTEST_ARGS=("--quiet" "--disable-warnings" "--numprocesses=auto")
```
timoludwig, line 30: we could keep track of the pytest args in an array

dev-tools/test.sh
```diff
@@ -25,38 +26,51 @@ while test $# -gt 0; do
...
-if [[ -z "$TEST_PATH" ]]; then
-    echo -e "Running all tests$CHANGED_MESSAGE..." | print_info
-    deescalate_privileges pipenv run pytest $TESTMON --disable-warnings --quiet --numprocesses=auto --cov=integreat_cms --cov-report html
-elif [[ -e "$TEST_PATH" ]]; then
-    echo -e "Running tests in ${TEST_PATH}$CHANGED_MESSAGE..." | print_info
-    deescalate_privileges pipenv run pytest "$TEST_PATH" $TESTMON --quiet --disable-warnings --numprocesses=auto
-else
+# Determine whether coverage data should be collected
+if [[ -z "${CHANGED}" && -z "${TEST_PATH}" ]]; then
+    PYTEST_ARGS+=("--cov=integreat_cms" "--cov-report html")
```
timoludwig, line 55: only collect coverage when neither the changed flag, nor the test path are given

timoludwig (approved):
Thanks a lot, really cool feature! :+1:

I had a bunch of warnings when I executed with the `--changed` flag but without the `TEST_PATH` given - it tried to collect coverage data, but all the unchanged nodes did not return any.
So my suggestion would be to only collect coverage when the script is neither run with the test path, nor with the changed flag... or is there a better way?

My suggestions didn't work with the GitHub suggestions feature, so I directly created a commit with them... could you have a look the the changes, and check whether I messed something up or whether you want to improve anything?

charludo:
> Thanks a lot, really cool feature! +1
> 
> I had a bunch of warnings when I executed with the `--changed` flag but without the `TEST_PATH` given - it tried to collect coverage data, but all the unchanged nodes did not return any. So my suggestion would be to only collect coverage when the script is neither run with the test path, nor with the changed flag... or is there a better way?
> 
> My suggestions didn't work with the GitHub suggestions feature, so I directly created a commit with them... could you have a look the the changes, and check whether I messed something up or whether you want to improve anything?

Very nice, thanks for the changes. Much cleaner.

However, `"${PYTEST_ARGS[@]}"` only appends the last element of the args-array to the command in my case. It might very well be that this is a peculiarity with zsh though. I've pushed a workaround, mind taking a final look? 😄

dev-tools/test.sh
```diff
@@ -16,23 +16,61 @@ require_database
...
+    TEST_MESSAGE=" in ${TEST_PATH}"
+    PYTEST_ARGS+=("${TEST_PATH}")
+elif [[ -n "${TEST_PATH}" ]]; then
+    # If the test path does not exist but was non-zero, show an error
     echo -e "${TEST_PATH}: No such file or directory" | print_error
     exit 1
 fi
 
+echo -e "Running all tests${TEST_MESSAGE}${CHANGED_MESSAGE}..." | print_info
+eval "deescalate_privileges pipenv run pytest ${PYTEST_ARGS[*]}"
```
timoludwig, line 70: Hmm, I'm not convinced that this is a valid use case for `eval`...
Maybe we could debug the problems with the previous solution?
Since `#!/bin/bash` is used as shebang, your default shell should have no influence on the fact that the script is executed by bash... but it could be something platform-specific, what's the output of your `/bin/bash --version`?
Could you check out my commit and insert `set -x` somewhere above the pytest command? This should print out what exactly is being run.
E.g. when I run ` ./dev-tools/test.sh --changed tests/api`, it prints out the following:
```
Running all tests in tests/api affected by recent changes...
+ IFS=
+ read -r line
+ deescalate_privileges pipenv run pytest --quiet --disable-warnings --numprocesses=auto --testmon-nocollect tests/api
++ id -u
+ [[ 1000 == 0 ]]
+ env pipenv run pytest --quiet --disable-warnings --numprocesses=auto --testmon-nocollect tests/api
```

charludo: Ah, good call - the issue was the argument `--cov-report html`, which got auto-escaped. Replaces it with `--cov-report=html`, now everything works as intended.

Thanks btw, didn't know about `set -x`

timoludwig: Ahh ok makes sense, thanks for figuring that out! :+1:

timoludwig (approved):

Issue #1889, opened 2022-11-22
Make translation status independent from publishing status


timoludwig:
### Motivation

At the moment, the translation status is "missing" if only a draft translation exists.
This causes several problems and makes it hard to differentiate the "translation status" from the "status of the translation".

### Proposed Solution

Calculate the translation status always for the latest major version, independently whether this version is published or not.

### Alternatives

Keep the current implementation

PR #1899, opened 2022-11-24
Make translation status indepentend from publishing status

MizukiTemma:
### Short description

This PR makes translation status (up-to-date, outdated, missing, fallback) independent from publishing status (public, draft, preview, auto-save) by comparing the latest major version of each language, while currently only the latest major "public" versions are taken into consideration for translation status calculation.

### Proposed changes

- Use the latest major (= not minor edit) version of each language to calculate translation status (up-to-date, outdated, missing, fallback), regardless of publishing status (public, draft, preview, auto-save).
- This change applies for Page, POI and Event.

### Side effects

- pylint: disable=too-many-public-methods is introduced into the class AbstractContentModel. Otherwise having more than 20 methods will not be allowed.
- I have probably misunderstood the problem or imagined something other than what actually the user wish is 🙈 

### Resolved issues

Fixes: #1889


integreat_cms/cms/fixtures/test_data.json
```diff
@@ -1863,7 +1863,7 @@
       "currently_in_translation": false,
       "version": 2,
       "minor_edit": false,
-      "last_updated": "2022-09-11T08:04:39.746Z",
+      "last_updated": "2019-08-12T08:04:49.746Z",
       "creator": 9,
```
JoeyStk, line 1867: I don't really understand this change? Is it a necessary change?

JoeyStk (changes requested):
Thank you a lot for this PR. I think in general this looks good. However I have one question left, but maybe that is just on me. In the page view template it still says "Status" - I was wondering if that should also be changed to publication status

MizukiTemma:
> Thank you a lot for this PR. I think in general this looks good. However I have one question left, but maybe that is just on me. In the page view template it still says "Status" - I was wondering if that should also be changed to publication status

It's a good idea 👍

integreat_cms/cms/fixtures/test_data.json
MizukiTemma: Ah, I should have mentioned to this point in PR description😅 
This is to keep the translation status of English version be "outdated" even after the change in translation calculation.

For a page content "Wissenswertes über Augusburg (Trivia about Augusburg)" we had the following test data.

- German ("Wissenswertes über Augusburg")
          Version 1: public, 2019-08-12T08:05:39.369Z
- English ("Trivia about Augusburg")
          Version 1: public, 2019-08-12T08:04:39.746Z
          Version 2: review, 2022-09-11T08:04:39.746Z

Before the change in translation status calculation, the english translation was judged as "outdated" since the english version 1 was used for translation status calculation and the english version 1 was created earlier than the german version 1.

Now with the change the latest major version is used for translation status calculation regardless of published or not. This means the english version 2 is now compared to the german version 1 in translation status calculation. To keep "Trivia about Augusburg" sill "outdated" after this change, the date of english version2 should be between the date of English version 1 and that of German version 1.

JoeyStk: Ah okay, that makes sense :)

JoeyStk (approved):
I checked it again. The publication statuses change as expected :)

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -4124,6 +4122,13 @@ msgstr "Veröffentlichen"
 msgid "Submit for review"
 msgstr "Zur Überprüfung vorlegen"
 
+#: cms/templates/events/event_form.html cms/templates/events/event_list.html
+#: cms/templates/pages/page_form.html cms/templates/pages/page_revisions.html
+#: cms/templates/pages/page_tree.html cms/templates/pois/poi_form.html
+#: cms/templates/pois/poi_list.html
+msgid "Publication status"
+msgstr "Veröffentlichung"
```
timoludwig, line 4130:
```suggestion
msgstr "Veröffentlichungsstatus"
```
Maybe a more literal translation?

integreat_cms/cms/models/abstract_content_model.py
```diff
@@ -367,7 +394,7 @@ def get_translation_state(self, language_slug):
         :return: A string describing the state of the translation, one of :data:`~integreat_cms.cms.constants.translation_status.CHOICES`
         :rtype: str
         """
-        translation = self.get_major_public_translation(language_slug)
+        translation = self.get_major_translation(language_slug)
         if not translation:
             if self.fallback_translations_enabled:
                 fallback_translation = self.get_public_translation(
```
timoludwig, line 400:
```suggestion
                fallback_translation = self.get_translation(
```
I guess to be consistent here, we'd need to show the "fallback" icon always if there is a fallback translation, independently from their publication status?
This sure might be a bit confusing in the beginning, because if the fallback is a draft, then also the non-existing translation is not shown in the app. But if we're being strict here, we should also make the fallback status independent from publication...

timoludwig (changes requested):
Thanks a lot! :+1:

From my perspective, this looks good! However, as always with changes in the translation/publication status area and everything corresponding with minor edits, I expect some weird edge cases where this fix breaks other stuff etc... I guess we'll have to live with that fear :sweat_smile: 

> I have probably misunderstood the problem or imagined something other than what actually the user wish is

No, actually I think this is exactly what the issue was about!

MizukiTemma:
> Thanks a lot! 👍
> 
> From my perspective, this looks good! However, as always with changes in the translation/publication status area and everything corresponding with minor edits, I expect some weird edge cases where this fix breaks other stuff etc... I guess we'll have to live with that fear 😅
>

yes, I'm actually afraid of that 😅 though I'm always there to fix if any edge case appears  💪

timoludwig (approved):
Thanks! :+1:

Issue #1876, opened 2022-11-18
Make items in the sidebar of content forms expandable and save the setting as a cookie


ulliholtgrave:
### Motivation

Right now our content forms (e.g. page form) are pretty stuffed with elements and the design team is working on it. However, as a quick fix and considering the additonal SEO, HIX and AT section we want to reduce the elements ASAP.

### Proposed Solution

Make the sidebar and SEO section expandable and save the last setting in a cookie to load it if the user returns to that view.
- [x] Page form
- [x] Event form
- [x] POI form

This blocks:
- [ ] #1769

svenseeberg:
Is this about collapsing the full sidebar (to the right) or single boxes within the side bar?

ulliholtgrave:
I guess as we are not even showing all items to non-expert users I wouldn't really go with a role-based view.

In my opinion we should add an expandable option to all 5 boxes on the right side as some of them (e.g. Settings of the page) aren't really used that often. I wouldn't mind providing expanding on a more granular level, but as a quick fix I would start with the big boxes.

svenseeberg:
Okay, I guess we agree that this is about collapsing single boxes.

ulliholtgrave:
Ahh yeah, right. I misunderstood your last statement. I will adjust the title of issue.

timoludwig:
This was fixed in #1972

PR #1903, opened 2022-11-28
Make sidebar items toggleable and save setting in cookie

JoeyStk:
### Short description
This PR adds the functionality to make items in the sidebar of content forms expandable and save the setting as a cookie.

### Proposed changes

- Add a class and ids to the items
- Add functionality to make them toggleable (Side notes: a. only the head of the cards is the trigger, please let me know if you think the entire card should be clickable, b. I didn't add an icon to keep it minimal, please let me know if you prefer an icon, e.g. "+", "-" as commonly used for accordions)
- Safe value in cookies and retrieve them to remember previous decision

### Side effects

- If you toggle the "Settings on page" section on Desktop screens it still leaves some white space and looks a bit ugly. However my screen is too small to test and/or inspect this issue. If requested I could do this in the next few days on a big screen at the office

### Resolved issues

Fixes (partially):  #1876


integreat_cms/static/src/js/pages/page-clickable-sidebar.ts
```diff
@@ -0,0 +1,23 @@
...
+    for (let i = 0; i < boxes.length; i++) {
+        boxes[i].firstElementChild.addEventListener("click", () => {
+            boxes[i].lastElementChild.classList.toggle("hidden");
+            let cookieValue = "hidden";
+            if (boxes[i].lastElementChild.classList.contains("hidden")) {
+                cookieValue = "hidden";
+            } else {
+                cookieValue = "show";
+            }
+            document.cookie = boxes[i].id + "=" + cookieValue + ";SameSite=Lax";
```
charludo, line 13: use
```suggestion
            document.cookie = `${boxes[i].id}=${cookieValue};SameSite=Lax`;
```
to pacify the `prefer-template` linter warning

integreat_cms/static/src/js/pages/page-clickable-sidebar.ts
```diff
@@ -0,0 +1,23 @@
...
+    const cookies = document.cookie.split("; ").map((cookie) => cookie.split("="));
+    for (let i = 0; i < boxes.length; i++) {
+        boxes[i].firstElementChild.addEventListener("click", () => {
+            boxes[i].lastElementChild.classList.toggle("hidden");
+            let cookieValue = "hidden";
+            if (boxes[i].lastElementChild.classList.contains("hidden")) {
+                cookieValue = "hidden";
+            } else {
+                cookieValue = "show";
+            }
```
charludo, line 12:
```suggestion
            const cookieValue = boxes[i].lastElementChild.classList.contains("hidden") ? "hidden" : "show";
```
is a bit cleaner/more readable, I think

charludo (changes requested):
Thanks, this is very useful, and works great.

My only complaint (besides the two small suggestions below) is that this sets a cookie for every box. Wouldn't it make more sense to set a single cookie containing a serialized JSON object with box names as the key / hidden status as the value?

integreat_cms/cms/templates/pages/page_form.html
```diff
@@ -165,7 +165,8 @@ <h1 class="heading">
                     {% include "../hix_widget.html" %}
                 {% endif %}
                 {% if page %}
-                    <div class="mt-4 lg:mt-0 rounded border border-blue-500 shadow-2xl bg-white">
+                    <div id="minor-edit"
+                         class="sidebar-box mt-4 lg:mt-0 rounded border border-blue-500 shadow-2xl bg-white">
```
timoludwig, line 169: We already have a mechanism to toggle boxes (they are currently only used on the dashboard):
[integreat_cms/cms/templates/_collapsible_box.html](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/templates/_collapsible_box.html) in combination with [integreat_cms/static/src/js/collapsible-boxes.ts](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/static/src/js/collapsible-boxes.ts)...
Probably it would be a bit more effort to use the sub-template in all boxes, but in the end, it would make the bit a bit more clean when we don't have duplicated functionality for this.
Also, I think the chevron icon of the existing boxes look nicer and the animation when toggling the boxes is more smooth when toggling the height instead of the visibility.
And we would have the advantage that the collapsed state of all dashboard boxes are remembered as well.
I also think we should add the same change for boxes in the event and POI form for consistency.

integreat_cms/static/src/js/pages/page-clickable-sidebar.ts
```diff
@@ -0,0 +1,23 @@
+document.addEventListener("DOMContentLoaded", () => {
+    const boxes = document.querySelectorAll(".sidebar-box");
+    const cookies = document.cookie.split("; ").map((cookie) => cookie.split("="));
+    for (let i = 0; i < boxes.length; i++) {
```
timoludwig, line 4: The for-loop with a counter seems a bit old fashioned to me, I guess a `forEach` would be the more modern approach?
```suggestion
    boxes.forEach((box) => {
```

integreat_cms/static/src/js/pages/page-clickable-sidebar.ts
```diff
@@ -0,0 +1,23 @@
...
+            document.cookie = boxes[i].id + "=" + cookieValue + ";SameSite=Lax";
+        });
+    }
+    for (let j = 0; j < cookies.length; j++) {
+        for (let k = 0; k < boxes.length; k++) {
+            if (cookies[j][0] === boxes[k].id && cookies[j][1] === "hidden") {
+                document.querySelector("#" + boxes[k].id).lastElementChild.classList.add("hidden");
+            }
+        }
+    }
```
timoludwig, line 22:
```suggestion
    cookies.forEach(([id, value]) => {
        document.getElementById(id)?.lastElementChild.classList.add("hidden");
    })
```
I think this loop is a bit inefficient - the cookie already holds the id of the box, so no need to iterate over all boxes just to fine the correct box with the right id.
Also, I think we should not overcomplicate matters and only store the cookie when the box is actually hidden - and just remove the entry when it's shown again.

timoludwig (changes requested):
Thanks a lot! :tada:

> My only complaint (besides the two small suggestions below) is that this sets a cookie for every box. Wouldn't it make more sense to set a single cookie containing a serialized JSON object with box names as the key / hidden status as the value?

Yes, definitely agree! :+1:

I would even argue we don't need the value here - just store list of all ids which should be hidden and remove elements from that list which should be shown.

Also, from a security standpoint, it's kind of risky to set cookies to dynamic keys which are derived from the document. I didn't find an immediate way to exploit this, but imagine being able to set the `csrftoken` cookie to a trivial value just by managing to create a HTML element with the id `csrftoken` - definitely not ideal...

integreat_cms/cms/templates/pages/page_form.html
JoeyStk: I think this should be fixed :thinking:

integreat_cms/cms/templates/pages/page_form.html
```diff
@@ -167,288 +167,16 @@ <h1 class="heading">
...
-                            <label>{% minor_edit_label request.region language %}</label>
-                            {% render_field page_translation_form.minor_edit %}
-                            <label for="{{ page_translation_form.minor_edit.id_for_label }}"
-                                   class="secondary">
-                                {{ page_translation_form.minor_edit.label }}
-                            </label>
-                            <div class="help-text">{% minor_edit_help_text request.region language page_translation_form %}</div>
-                        </div>
-                    </div>
+                    {% include "./page_form_sidebar/_page_minor_edit_template.html" with box_id="page-minor-edit" %}
```
timoludwig, line 170:
```suggestion
                    {% include "./page_form_sidebar/minor_edit_box.html" with box_id="page-minor-edit" %}
```
I think "template" is kind of redundant - since this file is already in the `templates` directory and ends with `.html`, I guess it's clear that it's a template (the same goes for the other boxes).

integreat_cms/static/src/js/collapsible-boxes.ts
```diff
@@ -26,18 +29,35 @@ const toggleDashboardSection = async ({ target }: Event): Promise<void> => {
...
         content.style.height = `${content.clientHeight}px`;
         // Use timeout to allow a smooth transition
         setTimeout(() => {
             content.style.height = "0px";
         });
+        collapsible.dataset.collapsed = "";
     }
+    // Store the currently collapsed boxes in a cookie
+    const collapsedBoxes = Array.from(document.querySelectorAll("[data-collapsed]:not([id=''])")).map((box) => box.id);
+    document.cookie = `collapsed-boxes=${JSON.stringify(collapsedBoxes)}`;
```
timoludwig, line 44: This is just a preliminary suggestion - it doesn't work since it does not preserve e.g. the state of the page form, when dashboard boxes are toggled and vice versa. But maybe this is a good enough starting point for you to figure something out.

timoludwig (changes requested):
I really like that we now re-use existing functionality.

> And then we can have a chat about how it would be best to set the cookies

I committed a suggestion for the cookie handling - not really perfect yet, but maybe you can play around with it and improve it.
The goal should be to make this also re-usable for other boxes (e.g. on the dashboard) and remember their state as well.

integreat_cms/static/src/js/collapsible-boxes.ts
timoludwig: Ah, a funny think I just noticed: This nearly works, because the cookie only counts for the current path if not specified otherwise, so the page form cookie does not overwrite the dashboard cookie, but they just are independent of each other and we store one list of ids for each path:

cookie

In general, this is a good thing, because it makes things a bit easier - but it also creates one additional problem: The cookie is now valid only for one specific page, but probably we want to collapse the same box on all page forms, independently of the page id.

So probably, we need something along the lines of:
```suggestion
    const collapsedBoxes = Array.from(document.querySelectorAll("[data-collapsed]:not([id=''])")).map((box) => box.id);
    const cookiePath = window.location.pathname.split("/").slice(0, 3).join("/");
    document.cookie = `collapsed-boxes=${JSON.stringify(collapsedBoxes)};path=${cookiePath}`;
```

integreat_cms/static/src/js/collapsible-boxes.ts
```diff
@@ -26,18 +29,35 @@ const toggleDashboardSection = async ({ target }: Event): Promise<void> => {
...
     });
+    // Restore all collapsed boxes
+    const collapsedBoxes = getCookie("collapsed-boxes");
+    if (collapsedBoxes) {
+        JSON.parse(collapsedBoxes).forEach((id: string) => {
+            const collapsible = document.querySelector(`.collapsible[id='${id}']`);
+            if (collapsible) {
+                collapsible.querySelector(".up-arrow").classList.add("hidden");
+                collapsible.querySelector(".down-arrow").classList.remove("hidden");
+                collapsible.parentNode.querySelector(".collapsible-content").classList.add("h-0");
```
timoludwig, line 59:
```suggestion
            const collapsible = document.querySelector(`.collapsible[id='${id}']`) as HTMLElement;
            if (collapsible) {
                collapsible.querySelector(".up-arrow").classList.add("hidden");
                collapsible.querySelector(".down-arrow").classList.remove("hidden");
                collapsible.parentNode.querySelector(".collapsible-content").classList.add("h-0");
                collapsible.dataset.collapsed = "";
```
And here we should probably also set the data attribute which is later used to store the cookie.

integreat_cms/static/src/js/collapsible-boxes.ts
JoeyStk: Yes, this solves the issue. I was working on a different idea in the meantime, but my solution only fixed some of the issues. So let's go with your solution. As far as I tested it, it's working reliably now :) Thank you a lot

timoludwig (approved):
Thanks a lot, really cool feature! :rocket: 

I noticed that it does not fix #1876 completely (since the issue was about all content forms, also events and POIs), but I think we should merge this PR now since it's a good improvement on its own, and add the event and poi forms later.

PR #1905, opened 2022-11-28
Enable dev tools for rootless docker

timoludwig:
### Short description

I hope this makes it possible to use our dev tools even if docker is running in rootless mode.
@membralala could you check whether this works for you? 

### Proposed changes

- Instead of requiring root permissions for all docker operations, check whether `docker ps` already succeeds without sudo


svenseeberg (approved):
Works as expected on Linux :+1: Thanks.

membralala (approved):
I'm sorry, I missed that one, somehow overread the notification 🙈 This works great for me, thank you very much!

Issue #1345, opened 2022-04-06
Track DeepL API usage


svenseeberg:
### Motivation
We have an automatic translation feature via DeepL. However, we have to pay for translations. We do want to keep track of our used quota.

### Proposed Solution
Query our used words quota before and after sending translation requests and store the difference in our database (InfluxDB?)

### Alternatives
The total amount of words translated could be stored in an attribute for the region.

charludo:
@svenseeberg I'm a bit unsure why we'd want to store the used quota. Wouldn't it make more sense to just query the DeepL API (see Ulli's link) for our usage any time it's relevant (e.g. before sending translation request, in the planned [translation budget widget](https://www.figma.com/file/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=166%3A1980), ...)? That way it's always up to date.

So for now I'd add a function to check the usage, and raise an exception in the `deepl_translation`  function if the translation would exceed the limit.

timoludwig:
> Wouldn't it make more sense to just query the DeepL API (see Ulli's link) for our usage any time it's relevant (e.g. before sending translation request, in the planned [translation budget widget](https://www.figma.com/file/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=166%3A1980), ...)? That way it's always up to date.

I think DeepL only tracks our global credit count, but we want to keep track of which portion of that is used by which region... this is not only done to prevent errors during the automatic translation, but also for our business calculation to see whether the usage roughly matches the price we charge the regions for this feature. This way, we could identify e.g. "power-user" regions which use the feature above average and maybe need an individual rate or whatever...

> So for now I'd add a function to check the usage, and raise an exception in the `deepl_translation` function if the translation would exceed the limit.

Not sure whether that's required - can we just "try" to translate and then handle the error thrown by DeepL when the quota is reached? I'm also not sure whether we're able to calculate the limit in exactly the same way as DeepL... what about special HTML characters, white spaces etc., are they billed exactly like translatable text?

timoludwig:
> from what I can see in the [wiki](https://wiki.integreat-app.de/maschinelleuebersetzungen), the plan is to offer a budget of 50.000 words. However, as far as I can tell, DeepL tracks usage by character. Is this an oversight, or are we actually tracking and displaying the translation usage of each community by word, regardless of word lengths/character counts?

[...]

PR #1909, opened 2022-11-29
Track DeepL API usage

charludo:
### Short description

This PR enables tracking of the number of words translated with DeepL by a region. Translations of more words than the remaining budget allows are blocked.

### Proposed changes

- add various new fileds required for tracking the credit usage, budget year and add-on booking to the region model
- check if enough translation budget remains before attempting any translation
- if the budget is exhausted, don't translate and show an error message 
- unencode content before sending it to DeepL to fix broken Umlaut-translation

### Side effects

- (hopefully) none

### Resolved issues

Fixes: #1345, #1932

#### About testing out this PR
I don't know if there's an easier way, but what worked for me was adding `or True` to [this line](https://github.com/digitalfabrik/integreat-cms/blob/9af8ee7014b0b284e7a88fd66ab1a9c18544451f/integreat_cms/deepl_api/apps.py#L30) and a (free) DeepL auth-key [here](https://github.com/digitalfabrik/integreat-cms/blob/9af8ee7014b0b284e7a88fd66ab1a9c18544451f/integreat_cms/core/settings.py#L699).


JoeyStk (changes requested):
Thank you so much for this PR! It's a really important feature!
I have two requests: 
- Could you please squash your commits. I think at least the "fixup:commit" could be squashed
- As far as I can tell we're still missing a mechanism that decides in which time interval the budget should be counted. @osmers what mechanism do you think would make most sense?

osmers:
Could you please add the possibility of choosing two dates?

- A required field to set the renewal date
- An optional field to determine the starting date for edge cases where the starting date is not the renewal date in the first year. 

Information regarding the optional field - if possible only show it after checking a check box called something like "Unterjähriger Start" - if the optional field has a date, the remaining budget should be calculated by dividing 1.000.000 words by 12 months and then multiplying by however many months are left till the renewal date.

And, pls disable remaining budget, as we do not wish to change that information

integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -424,6 +426,27 @@ def clean_hix_enabled(self):
...
+        if cleaned_deepl_usage_remaining > cleaned_deepl_usage_limit:
+            self.add_error(
+                "deepl_usage_remaining",
+                _(
+                    "The remaining number of words can not be larger than the total number of allowed words."
+                ),
+            )
+            return False
+        return self.cleaned_data.get("deepl_usage_remaining")
+
```
timoludwig, line 449:
```suggestion
```
As @osmers suggested, I'd make this a readonly field - and then, we also don't need to validate the input.

charludo:
> dividing 1.000.000 words by 12 months

Just for clarification - it's 1M free words? SInce in the wiki, it says 50k afaik.

timoludwig:
> Just for clarification - it's 1M free words? Since in the wiki, it says 50k afaik.

The way I understood it: 50k words [are free](https://wiki.integreat-app.de/maschinelleuebersetzungen), 1M words can be booked as [add-on](https://wiki.integreat-app.de/uebersetzungspaket)?

charludo:
> > Just for clarification - it's 1M free words? Since in the wiki, it says 50k afaik.
> 
> The way I understood it: 50k words [are free](https://wiki.integreat-app.de/maschinelleuebersetzungen), 1M words can be booked as [add-on](https://wiki.integreat-app.de/uebersetzungspaket)?

So if there are no custom packages, and the add-on is only bookable once (?) a checkbox "add-on booked"  is sufficient and I can do away with the manual entry fields?

integreat_cms/cms/models/regions/region.py
```diff
@@ -321,6 +321,22 @@ class Region(AbstractBaseModel):
         ),
     )
 
+    deepl_usage_limit = models.PositiveIntegerField(
+        default=50_000,
+        blank=True,
+        verbose_name=_("Yearly DeepL machine translation budget"),
+        help_text=_("Number of words that can be machine translated by DeepL per year"),
+    )
```
timoludwig, line 329: > So if there are no custom packages, and the add-on is only bookable once (?) a checkbox "add-on booked" is sufficient and I can do away with the manual entry fields?

Yes, exactly. I think it would make sense to convert the limit to a global setting instead and only add a boolean field to the region model which tells us whether the additional package was booked.
This would also make it easier to adjust these values in the future.

integreat_cms/cms/models/regions/region.py
```diff
@@ -321,6 +321,22 @@ class Region(AbstractBaseModel):
...
+
+    deepl_usage_remaining = models.PositiveIntegerField(
+        default=50_000,
+        verbose_name=_("Remaining DeepL machine translation budget"),
+        blank=True,
+        help_text=_(
+            "Remaining number of words that can be translated by DeepL this year"
+        ),
+    )
```
timoludwig, line 338: Instead of keeping track of the remaining words, I would rather start from 0 and count the consumed words. This makes it easier to change the limit independently from the used words.

integreat_cms/deepl_api/utils.py
```diff
@@ -139,3 +187,9 @@ def deepl_translation(self, request, content_objects, language_slug, form_class)
                         source_translation.title,
                     ),
                 )
+
+            # Update remaining DeepL usage for the region
+            request.region.deepl_usage_remaining = max(
+                0, request.region.deepl_usage_remaining - translation_word_count
+            )
+            request.region.save()
```
timoludwig, line 195: Probably a smart region manager with enough scripting skills can trick this counter by performing all translations at the same time :smile: 
I'm not even sure whether we can use [`select_for_update()`](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#select-for-update) here since the region object is already queried from the database in the [`RegionMiddleware`](https://github.com/digitalfabrik/integreat-cms/blob/9ba3bfbe5a8a4185258e9d4ebb969974522b8897/integreat_cms/core/middleware/region_middleware.py#L57)... but maybe re-querying at this point in time still works?

integreat_cms/deepl_api/utils.py
```diff
@@ -42,6 +43,38 @@ def check_availability(request, language_slug):
...
+        """
+        # Gather content to be translated and calculate total word count
+        content_to_translate = []
+        for attr in self.translatable_attributes:
+            if hasattr(source_translation, attr):
+                content_to_translate.append(getattr(source_translation, attr))
+        content_to_translate = " ".join(content_to_translate)
+        for char in "-,.\n":
+            content_to_translate = content_to_translate.replace(char, " ")
+        word_count = len(content_to_translate.split())
```
timoludwig, line 67: As mentioned in the issue description, I guess this will be almost certainly wrong. A better estimation would maybe be to use [`strip_tags()`](https://docs.djangoproject.com/en/3.2/ref/utils/#django.utils.html.strip_tags), but then again I'm not sure if we're using much more of DeepLs credits than what we charge our own customers.
Would also be interesting to know how tinymce calculates the words in the editor^^

timoludwig (changes requested):
Thanks a lot, very cool! :tada:

osmers:
> > > Just for clarification - it's 1M free words? Since in the wiki, it says 50k afaik.
> > 
> > 
> > The way I understood it: 50k words [are free](https://wiki.integreat-app.de/maschinelleuebersetzungen), 1M words can be booked as [add-on](https://wiki.integreat-app.de/uebersetzungspaket)?
> 
> So if there are no custom packages, and the add-on is only bookable once (?) a checkbox "add-on booked" is sufficient and I can do away with the manual entry fields?

We would still need the date field, right? Because otherwise we do not know the renewal date for the 1.000.000 words. And in case a Kommune wants to start before their renewal date (which is tied to their contract start date, thus not freely choosable) we would need some way of calculating the portion of 1.000.000 words that remain for the period before the renewal date, right?

integreat_cms/deepl_api/utils.py
charludo: I unfortunately wasn't able to find the source code for the TinyMCE plugin responsible, but I think your `strip_tags()` suggestion plus the `tang_handling="html"` option should probably be accurate enough

timoludwig: Sounds good! :+1:

integreat_cms/deepl_api/utils.py
charludo: This appears to be working, with one caveat: importing `Region` at the top of the module crashes Django with `django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet`. Moving the import inside the class fixes the error, though obviously this is not ideal either.

Do you happen to know what the reason behind this is?

JoeyStk (changes requested):
Thank you so much for the changes! Maybe this is on me, but I couldn't find an option to see the remaining deepl credits. However if I understood @osmers correctly this field was needed, but had to be disabled. The rest looks very good though

charludo:
> Thank you so much for the changes! Maybe this is on me, but I couldn't find an option to see the remaining deepl credits. However if I understood @osmers correctly this field was needed, but had to be disabled. The rest looks very good though

No, there currently isn't one, that's true. One of the requested changes was to switch from counting down the remaining credits to simply countign how many have been used, and calculating the remaining ones on the fly.

I think #1926 covers showing the remaining credits, and this is a bit out of scope for this PR. I guess we could add "(xx.xxx credits remaining)" or similar to the successfull translation message, though I don't think that info is useful in most cases.

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -3310,6 +3315,56 @@ msgid ""
 msgstr ""
 "Ob automatische Übersetzungen in Leichte Sprache mit SUMM.AI aktiviert sind"
 
+#: cms/models/regions/region.py
+msgid "DeepL credits renewal date"
+msgstr "DeepL Credits Zurücksetzungsdatum"
+
+#: cms/models/regions/region.py
+msgid "Day of the year on which the budget usage is reset"
+msgstr "Tag des Jahres zu welchem das Budget zurück gesetzt wird"
```
JoeyStk, line 3324: There is a tiny typo here: zurückgesetzt is [one word](https://www.duden.de/rechtschreibung/zuruecksetzen)

JoeyStk (approved):
Okay, I understand. 
I found one typo and @osmers is still going to discuss if the 1m words are on top of the 50k free words (as it is in your PR) or if in this case the free budget gets cancelled and instead it's just 1m words (and not 1.050.000). But @osmers will let you know about this.
For me this looks good now :)

integreat_cms/deepl_api/utils.py
```diff
@@ -58,84 +112,117 @@ def deepl_translation(self, request, content_objects, language_slug, form_class)
...
+                    # Only translate existing, non-empty attributes
+                    if hasattr(source_translation, attr) and getattr(
+                        source_translation, attr
+                    ):
+                        # data has to be unescaped for DeepL to recognize Umlaute
+                        data[attr] = self.translator.translate_text(
+                            unescape(getattr(source_translation, attr)),
+                            source_lang=source_language.slug,
+                            target_lang=target_language_key,
+                            tag_handling="html",
```
timoludwig, line 186: Since only the `content` field is HTML and `title` and `meta_description` are plain text, I'm not sure whether we need to differentiate here...
But probably there should be no big problems with enabling this option for non-html content, right?

integreat_cms/deepl_api/utils.py
```diff
@@ -58,84 +112,117 @@ def deepl_translation(self, request, content_objects, language_slug, form_class)
...
-                    "creator": request.user,
-                    "language": target_language,
-                    source_translation.foreign_field(): content_object,
-                },
+        # Re-select the region from db to prevent simultaneous
+        # requests exceeding the DeepL usage limit
+
+        with transaction.atomic():
+            region = (
+                apps.get_model("cms", "Region")
```
timoludwig, line 120: I think this is fine! I see no way around it since the `DeepLApi` class is imported in `deepl_api`'s `apps.py` which has a [`ready()`](https://github.com/digitalfabrik/integreat-cms/blob/ad73c79b7b994b756997fb1019b63662345ff543/integreat_cms/deepl_api/apps.py#L27)-function which is run when the deepl api app is initialized.
So my theory why the `AppRegistryNotReady` error appears is that the `ready()`-function is executed before the models are loaded, meaning we cannot import them at this point in time.

integreat_cms/cms/models/regions/region.py
```diff
@@ -321,6 +321,45 @@ class Region(AbstractBaseModel):
...
+
+    deepl_midyear_start_enabled = models.BooleanField(
+        default=False,
+        verbose_name=_("DeepL budget year start differs from renewal date"),
+        help_text=_(
+            f"Enable to set an add-on starting date differing from the renewal date. Budget will be set as a monthly fraction of {settings.DEEPL_CREDITS_ADDON:,} credits"
+        ),
+    )
+
```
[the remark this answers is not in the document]
charludo: I kind of agree, however this is something @osmers specifically requested:

> Could you please add the possibility of choosing two dates?
> 
>     * A required field to set the renewal date
> 
>     * An optional field to determine the starting date for edge cases where the starting date is not the renewal date in the first year.
> 
> 
> Information regarding the optional field - if possible only show it after checking a check box called something like "Unterjähriger Start"

I also don't think clicking the tiny "x" to the right of the date field to clear it is an intuitive way of disabling the feature.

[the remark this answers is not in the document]
timoludwig: Yes, I think it's a good idea to keep the checkbox - I mean just that we don't need to store this information in the database.

integreat_cms/cms/models/regions/region.py
```diff
@@ -321,6 +321,45 @@ class Region(AbstractBaseModel):
...
+    deepl_midyear_start_date = models.DateField(
+        default=django_timezone.now,
```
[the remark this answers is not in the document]
charludo: Yep, good point. Forgot to change the default to None afterhaving made the field nullable.

integreat_cms/deepl_api/utils.py
charludo: Thanks for the explanation. After asking I realized the workaround with `get_model` was used in lots of places, I guess this has happened before? :D

integreat_cms/cms/models/regions/region.py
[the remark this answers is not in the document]
charludo: Oh, I see!

integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -125,12 +125,18 @@ class Meta:
...
+            "deepl_midyear_start_enabled",
+            "deepl_midyear_start_date",
         ]
         #: The widgets which are used in this form
         widgets = {
             "timezone": forms.Select(choices=get_timezone_choices()),
             "icon": IconWidget(),
             "offers": forms.CheckboxSelectMultiple(),
+            "deepl_renewal_date": forms.DateInput(attrs={"type": "date"}),
+            "deepl_midyear_start_date": forms.DateInput(attrs={"type": "date"}),
```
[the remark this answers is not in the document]
timoludwig: Hmm, good question. I thought maybe we could omit the day and just assume the 1. of each month? @osmers do you think this is "good enough"?

integreat_cms/cms/models/regions/region.py
```diff
@@ -726,6 +765,29 @@ def get_repr(self):
...
+        if self.deepl_addon_booked and self.deepl_midyear_start_enabled:
+            # only the months matter, the exact dates are disregarded
+            # calculate the difference between the two dates in months
+            midyear_start_month = getattr(self.deepl_midyear_start_date, "month")
+            renewal_month = getattr(self.deepl_renewal_date, "month")
+            months_difference = 12 - ((midyear_start_month - renewal_month) % 12)
+            self.deepl_addon_multiplier = months_difference / 12
+
+        super().save(*args, **kwargs)
+
```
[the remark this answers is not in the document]
timoludwig: Ah ok, yeah makes sense I guess. Would be cool if a monthly cron job would be sufficient, let's wait for Svenjas reponse.

timoludwig (changes requested):
Amazing, thanks a lot! :muscle: 
This works as expected! :rocket: 

As always, just a few nit-picky details :wink:

integreat_cms/deepl_api/utils.py
```diff
@@ -42,6 +46,74 @@ def check_availability(request, language_slug):
...
+
+        :return: the region's budget
+        :rtype: int
+        """
+        # calculate the fractional modifier for differing add-on start/renewal dates
+        multiplier = 1.0
+        if region.deepl_addon_booked and region.deepl_midyear_start_date:
+            midyear_start_month = region.deepl_midyear_start_date + 1
+            renewal_month = region.deepl_renewal_date + 1
+            months_difference = 12 - ((midyear_start_month - renewal_month) % 12)
```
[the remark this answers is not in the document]
charludo: I'll leave this up to you. The `+ 1` is there just to have the `% 12`, which I think makes the calculation a little more intuitive.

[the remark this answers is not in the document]
timoludwig: I'm not sure I understand - wouldn't the `%12` just stay the same?

osmers:
> I found one typo and @osmers is still going to discuss if the 1m words are on top of the 50k free words (as it is in your PR) or if in this case the free budget gets cancelled and instead it's just 1m words (and not 1.050.000). But @osmers will let you know about this.

So it was explained to me the way that 50.000 is a Freigrenze, meaning, they have a total of 1.000.000 words and can use 50.000 for free
image
In the end, if they buy the package, they get 1.000.000 not 1.050.000

timoludwig (approved):
> In the end, if they buy the package, they get 1.000.000 not 1.050.000

Alright, I adapted the calculation now. Could somebody review again to check whether I messed something up?

> Also, I just saw that you requested my review - not sure I'd be the best person to review, since I have no idea how to do that :D

Haha yeah, I guess this was referring to the comments, whether you notice any misconceptions in our conversation, not about the actual code :sweat_smile:

integreat_cms/deepl_api/utils.py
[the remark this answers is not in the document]
timoludwig: I simplified the calculation a bit now - I hope it still works :see_no_evil: 
https://github.com/digitalfabrik/integreat-cms/blob/cf99bf80f183236ac35711b1726c73003464a83b/integreat_cms/cms/models/regions/region.py#L765-L769

integreat_cms/cms/templates/regions/region_form.html
```diff
@@ -259,6 +260,76 @@ <h3 class="heading font-bold text-black">
...
+                                </p>
+                            </div>
+                            <div class="flex flex-wrap justify-between">
+                                <label class="secondary m-0">
+                                    {% trans "Remaining words" %}:
+                                </label>
+                                <p>
+                                    {{ form.instance.deepl_budget_remaining|intcomma }}
+                                </p>
+                            </div>
```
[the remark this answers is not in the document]
timoludwig: Thanks a lot for checking! I fixed the CircleCI key to make sure the workflow can run again and rebased one more time to get rid of the merge-commit. But now this should be ready to be merged :tada:

PR #1913, opened 2022-11-29
Fix availability check of DeepL API during development

timoludwig:
### Short description

The change introduced in #1748 broke the DeepL functionality during local development.
Not sure if this solution is ideal tbh, maybe we should rather store the supported languages in the cache rather than the app config? :thinking: 

### Proposed changes

-  Fix availability check of DeepL API during development

### Side effects

- Slightly longer restarting times when the server reloads after a file has been edited

### DIFF
--- a/integreat_cms/deepl_api/apps.py
+++ b/integreat_cms/deepl_api/apps.py
@@ -1,11 +1,13 @@
 """
 Configuration of DeepL API app
 """
+import os
+import sys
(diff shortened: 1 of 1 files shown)


charludo (approved):
LGTM!

ulliholtgrave (approved):
Thanks, Seems to work for me 👍

Issue #1908, opened 2022-11-29
Overly verbose logging of `RelatedObjectDoesNotExist` error in `repr()` methods 


membralala:
### Describe the Bug

When creating a new POI, an error is thrown: 
```integreat_cms.cms.models.pois.poi.POI.region.RelatedObjectDoesNotExist: POI has no region.```

### Steps to Reproduce

1. Go to 'Locations'
2. Click on 'Create location'
3. See error

### Expected Behavior

A debug log message should show, how the POI was created

### Actual Behavior

The error occurs

<details>
  <summary>Traceback</summary>

```
Nov 29 10:01:41 DEBUG integreat_cms.cms.models.abstract_base_model - repr() for object <POI (id: None)> failed because of <class 'integreat_cms.cms.models.pois.poi.POI.region.RelatedObjectDoesNotExist'>: POI has no region.
Traceback (most recent call last):
  File "/Users/philip/Documents/Integreat/integreat-cms/integreat_cms/cms/models/abstract_base_model.py", line 51, in __repr__
    return self.get_repr()
  File "/Users/philip/Documents/Integreat/integreat-cms/integreat_cms/cms/models/abstract_content_model.py", line 433, in get_repr
    return f"<{class_name} (id: {self.id}, region: {self.region.slug}, slug: {self.best_translation.slug})>"
  File "/Users/philip/.local/share/virtualenvs/Integreat-4loMQVvc/lib/python3.9/site-packages/django/db/models/fields/related_descriptors.py", line 197, in __get__
    raise self.RelatedObjectDoesNotExist(
integreat_cms.cms.models.pois.poi.POI.region.RelatedObjectDoesNotExist: POI has no region.
```

</details>

timoludwig:
This is expected behavior - the repr method tries to be as informative as possible, but when the poi is logged during its creation, the related fields cannot be logged. 
I just added the debug message so we notice unexpected errors in the repr methods - but when the poi was not saved yet, this is totally normal.

PR #1916, opened 2022-11-30
Reduce logging verbosity if related object does not exist

timoludwig:
### Short description

Reduce logging verbosity if related object does not exist

### Proposed changes

- Skip the logging call in this case
- Add a notice that logged errors might be false positives and should be excluded when they occur

### Resolved issues

Fixes: #1908


ulliholtgrave (approved):
LGTM

MizukiTemma (approved):
Thank you 👍 Looks good.

Issue #1756, opened 2022-10-12
Imprint sbs-view has no media library, no content-edit-lock and no diff-view


david-venhoff:
### Describe the Bug

Looks like the imprint sbs-view was forgotten when these features where implemented.
This is similar to https://github.com/digitalfabrik/integreat-cms/pull/1666

Issue #1870, opened 2022-11-16
Copy source content in imprint side by side view not working anymore


timoludwig:
### Describe the Bug

The "Copy content from ... to ..." button in the imprint side by side view does not work anymore.
I assume this is since the update from the button in the page side by side view (see #1621).

### Steps to Reproduce

1. Go to imprint side by side view
2. Try to copy content from source to target
3. Nothing happens
4. See error in JS console

### Expected Behavior

The content should be copied. Ideally, we would implement exactly the same behaviour like in the page side by side view with the two buttons "Show source translation difference" and "Copy content...".

### Actual Behavior

Nothing happens and an error appears in the JS console:
```
Uncaught TypeError: content is undefined
```

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
Uncaught TypeError: content is undefined
    setContentString tinymce.js:15385
    setContentInternal tinymce.js:15439
    fold tinymce.js:262
    setContentInternal tinymce.js:15438
    setContent tinymce.js:17710
    setContent$2 tinymce.js:17941
    setContent tinymce.js:19678
    setContent tinymce.js:28841
    ts sbs-copy-content.ts:20
tinymce.js:15385

  ```

</details>

seluianova:
I'm fixing it within the task https://github.com/digitalfabrik/integreat-cms/issues/1756

PR #1919, opened 2022-11-30
Fix imprint sbs-view

seluianova:
### Short description
Lately introduced features were not added to imprint sbs-view:
media library, content-edit-lock, diff-view

### Proposed changes
- Add media library, content-edit-lock, diff-view
- Fix copy source content (old_translation_content variable was missed)

### Side effects
Didn't find any

### Resolved issues
Fixes: #1756, #1870


integreat_cms/cms/views/imprint/imprint_sbs_view.py
```diff
@@ -8,6 +8,9 @@
 from django.utils.translation import ugettext as _
```
david-venhoff, line 8:
```suggestion
from django.utils.translation import ugettext as _
from django.urls import reverse
```

integreat_cms/cms/views/imprint/imprint_sbs_view.py
```diff
@@ -8,6 +8,9 @@
...
 from django.views.generic import TemplateView
 
+from ..media.media_context_mixin import MediaContextMixin
+from ..mixins import ContentEditLockMixin
```
david-venhoff, line 12:
```suggestion
```

integreat_cms/cms/views/imprint/imprint_sbs_view.py
```diff
@@ -17,13 +20,17 @@
 
 @method_decorator(permission_required("cms.view_imprintpage"), name="dispatch")
 @method_decorator(permission_required("cms.change_imprintpage"), name="post")
-class ImprintSideBySideView(TemplateView):
+class ImprintSideBySideView(TemplateView, MediaContextMixin, ContentEditLockMixin):
```
david-venhoff, line 23:
```suggestion
class ImprintSideBySideView(TemplateView, MediaContextMixin):
```

integreat_cms/cms/views/imprint/imprint_sbs_view.py
```diff
@@ -17,13 +20,17 @@
...
     """
     View for the imprint side by side form
     """
 
     #: The template to render (see :class:`~django.views.generic.base.TemplateResponseMixin`)
     template_name = "imprint/imprint_sbs.html"
+
+    #: The url name of the view to show if the user decides to go back (see :class:`~integreat_cms.cms.views.mixins.ContentEditLockMixin`)
+    back_url_name = "edit_imprint"
+
```
david-venhoff, line 33:
```suggestion
```
I think it would be better to redirect the user to the dashboard, because if someone else has taken over, the `edit_imprint` page also can't be edited.
Unfortunately the `ContentEditLockMixin` cannot handle redirecting to the dashboard, so it has to be done manually. Github does not let me add a suggestion, but you can add this:
```python
    def get_context_data(self, **kwargs):
        r"""
        Returns a dictionary representing the template context
        (see :meth:`~django.views.generic.base.ContextMixin.get_context_data`).

        :param \**kwargs: The given keyword arguments
        :type \**kwargs: dict

        :return: The template context
        :rtype: dict
        """
        context = super().get_context_data(**kwargs)
        context.update(
            {
                "back_url": reverse(
                    "dashboard",
                    kwargs={
                        "region_slug": kwargs["region_slug"],
                    },
                )
            }
        )
        return context

```

integreat_cms/cms/templates/imprint/imprint_sbs.html
```diff
@@ -160,10 +171,12 @@ <h1 class="heading">
             </div>
         </div>
     </form>
+    {{ media_config_data|json_script:"media_config_data" }}
     {% if not perms.cms.change_imprintpage %}
         {% include "../_tinymce_config.html" with readonly=1 language=target_language %}
     {% else %}
         {% include "../_tinymce_config.html" with language=target_language %}
+        {% include "../_content_edit_lock.html" with lock_key=source_imprint_translation.page.edit_lock_key %}
     {% endif %}
```
david-venhoff, line 180:
```suggestion
    {% endif %}
    {% include "../generic_confirmation_dialog.html" %}
```
The confirmation dialog also needs to be included

david-venhoff (commented):
Thanks! This fixes most problems for me.
I am not certain, is the `.testmondata-journal` supposed to be included in commits?

charludo:
> I am not certain, is the `.testmondata-journal` supposed to be included in commits?
No, it shouldn't be - my mistake, I'll open a PR

integreat_cms/cms/views/imprint/imprint_sbs_view.py
seluianova: May I ask you to share user steps to get redirected to the dashboard?
If I wanted to go back, I would use the button "Back to the imprint form" or "back" in the browser...
I can't get the other options :(

david-venhoff (approved):
Thanks, LGTM!

integreat_cms/cms/views/imprint/imprint_sbs_view.py
```diff
@@ -225,5 +262,43 @@ def post(self, request, *args, **kwargs):
                 "imprint_translation_form": imprint_translation_form,
                 "source_imprint_translation": source_imprint_translation,
                 "target_language": target_language,
+                "old_translation_content": old_translation_content,
             },
         )
+
+
+def get_old_source_content(imprint, source_language, target_language):
```
timoludwig, line 270: This function is just copied + pasted from the page sbs view, right?
Probably we should think about combining these views into more generic ones to remove such duplications. But for now, I guess it's fine... I'll open another issue for this.

timoludwig (approved):
Looks good, thanks a lot! :+1:

integreat_cms/cms/views/imprint/imprint_sbs_view.py
seluianova: Do you think AbstractContentModel would be the right place for this function?

timoludwig: Would be one possibility, yes. But I opened #1946 for this, so you don't have to change it now if you don't want to.

Issue #1885, opened 2022-11-22
Ongoing translation cannot be cancelled if newer version exists


timoludwig:
### Describe the Bug

When a translation is "currently in translation", and then a new minor version is created, the translation status is still "in translation", but it cannot be cancelled anymore, since the corresponding check in the page form is only shown when the latest version (no matter if major or minor) is currently in translation.

### Steps to Reproduce

1. Export a page as XLIFF which has a pubic version but the latest version is a draft with the "only public" option
2. Open page form

### Expected Behavior

When a translation is currently in translation, it should be possible to cancel the process:

![Screenshot 2022-11-22 at 15-08-40 Integreat Editorial System](https://user-images.githubusercontent.com/16021269/203335080-73742b17-7cf6-4555-b792-42284cc4d298.png)

### Actual Behavior

The cancel button is not shown when the latest version is not in translation:
![Screenshot 2022-11-22 at 15-08-05 Integreat Editorial System](https://user-images.githubusercontent.com/16021269/203335050-c38d4ffe-4795-49ff-84f4-7981e69decd5.png)

PR #1929, opened 2022-11-30
Fix ongoing page translation not cancellable

david-venhoff:
### Short description

This pr fixes a bug where an ongoing translation was not cancellable if the latest version is a minor edit.
It also fixes two related bugs I noticed, where the status was always set to `up-to-date` after cancelling an ongoing translation, and a bug where the the translation status would not update after an xliff import/export if the latest version is a minor edit.

### Proposed changes

- Store the translation status in the latest version (as opposed to the latest major version previously)
- Don't update the objects timestamp after changing the translation state

### Side effects

/

### Resolved issues

Fixes: #1885

### DIFF
--- a/integreat_cms/cms/models/abstract_content_translation.py
+++ b/integreat_cms/cms/models/abstract_content_translation.py
@@ -431,7 +431,8 @@ class AbstractContentTranslation(AbstractBaseModel):
             # If the page does not have a major public version, it is considered "missing" (keep in mind that it might
             # have draft versions or public versions that are marked as "minor edit")
             return translation_status.MISSING
-        if translation.currently_in_translation:
+        # For "currently in translation", we consider the latest version instead of the latest major version
+        if self.currently_in_translation:
             return translation_status.IN_TRANSLATION
         if not self.source_language:
             # If the language of this translation is the root of this region's language tree, it is always "up to date"
--- a/integreat_cms/xliff/utils.py
+++ b/integreat_cms/xliff/utils.py
@@ -206,9 +206,14 @@ def page_to_xliff(page, target_language, dir_name, only_public=False):
     logger.debug("Created XLIFF file %r", actual_filename)
 
     # Set "currently in translation" status for existing target translation
-    if target_page_translation.id:
-        target_page_translation.currently_in_translation = True
-        target_page_translation.save()
+    if latest_version := page.get_translation(target_language.slug):
+        latest_version.currently_in_translation = True
+        latest_version.save(update_timestamp=False)
+        logger.debug(
+            "Updated translation status of %r to %r",
+            latest_version,
+            latest_version.translation_state,
+        )
 
     return download_storage.path(actual_filename)
 
@@ -391,23 +396,6 @@ def xliff_import_confirm(request, xliff_dir):
                         ).format(page_translation.readable_title, error_list),
                     )
                     success = False
-                elif not has_changed:
-                    # Update existing translation
-                    existing_translation = page_translation.latest_version
-                    existing_translation.currently_in_translation = False
-                    existing_translation.save()
-                    logger.info(
-                        "%r of XLIFF file %r was imported without changes by %r",
-                        existing_translation,
-                        xliff_file,
-                        request.user,
-                    )
-                    messages.info(
-                        request,
-                        _("Page {} was imported without changes.").format(
-                            page_translation.readable_title
-                        ),
-                    )
                 else:
                     # Check if previous version already exists
                     existing_translation = page_translation.latest_version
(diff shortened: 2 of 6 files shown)


MizukiTemma (commented):
Thank you for the PR 👍 It looks good as far as I tested.

For information: I tested in the following way.
1) Create a new page (German version 1 public)
2) Create an english translation (English version 1 public)
3) Create German Version 2 (public), making the english translation out-dated.
4) Export an xliff for English, making the english translation "currently in translation"
5) Create a new English "minor" version in Django Admin respectively in public, draft, auto-save or review.
6) Check if the cancel option is shown in the form.
In this test, it works as expected 😃 

And it's really cool that you fixed two other bugs on the way together 💪 
The bug that the translation status always turns to "up-to-date" after cancellation is fixed.

Now the second bug is left: I couldn't reproduce it in the develop branch. Can you describe how to trigger it? After this, I can probably approve your PR 👍

timoludwig (changes requested):
Thanks! :+1:
This fixes the problems partially, but causes another problem:
The "currently in translation" status is no longer inherited for new versions if the latest version is only minor.

I see three ways around it:
1. Just update all versions from the latest major version to the latest version
2. Update both the latest major version and the latest version (and not all in between)
3. Change the translation status function to consider the latest version for the currently in translation state instead of the latest major version

david-venhoff:
I changed the implementation to @timoludwig 's third idea (Only use the latest version to calculate the translation status), in hope that it is the simplest solution. Hopefully I did not does not miss any edge case! :grimacing:

integreat_cms/xliff/utils.py
```diff
@@ -208,7 +208,7 @@ def page_to_xliff(page, target_language, dir_name, only_public=False):
     # Set "currently in translation" status for existing target translation
     if target_page_translation.id:
         target_page_translation.currently_in_translation = True
```
timoludwig, line 210:
```suggestion
    if latest_version := page.get_translation(target_language.slug):
        latest_version.currently_in_translation = True
```
I guess in this case we should only update the status of the latest version (since if e.g. only public versions are exported to XLIFF, the `target_page_translation` will be the latest public version and not the latest version), otherwise a similar problem as before exists :sweat_smile:

integreat_cms/xliff/utils.py
```diff
@@ -208,7 +208,7 @@ def page_to_xliff(page, target_language, dir_name, only_public=False):
...
-        target_page_translation.save()
+        target_page_translation.save(update_timestamp=False)
```
timoludwig, line 211:
```suggestion
        latest_version.save(update_timestamp=False)
        logger.debug(
            "Updated translation status of %r to %r",
            latest_version,
            latest_version.translation_state,
        )
```

integreat_cms/cms/models/abstract_content_translation.py
```diff
@@ -431,7 +431,7 @@ def translation_state(self):
             # If the page does not have a major public version, it is considered "missing" (keep in mind that it might
             # have draft versions or public versions that are marked as "minor edit")
             return translation_status.MISSING
-        if translation.currently_in_translation:
+        if self.currently_in_translation:
```
timoludwig, line 434:
```suggestion
        # For "currently in translation", we consider the latest version instead of the latest major version
        if self.currently_in_translation:
```
In case we change something related to the translation state in the future (which isn't that unrealistic I think :grin:), I'd like to make sure we remember why we chose a different version here than for the other states.

integreat_cms/xliff/utils.py
```diff
@@ -395,7 +395,7 @@ def xliff_import_confirm(request, xliff_dir):
                     # Update existing translation
                     existing_translation = page_translation.latest_version
                     existing_translation.currently_in_translation = False
-                    existing_translation.save()
+                    existing_translation.save(update_timestamp=False)
```
timoludwig, line 398: I think in this case, the behavior was actually intended (see issue #1276): If a page was outdated and is then imported unchanged, this should update the status to "up to date".
However, with our recent changes of always creating new versions rather than modifying existing versions, I guess it would be more consistent to also create a new version here than to update the last updated date.
We can just do so by combining the last two conditions (so `elif not has_changed:` and `else` into one case and always delete the links and execute `page_translation.save()` and only adapt the info message depending on whether the page was actually changed or not.
Even though this is not strictly speaking part of this issue, could you add a commit with this change as well? Thanks in advance!

timoludwig (changes requested):
> Hopefully I did not does not miss any edge case!

You jinxed it :see_no_evil:

timoludwig (approved):
Thanks a lot! 
I think that's it! :muscle:

MizukiTemma (approved):
Looks good 🚀 Thank you!

Issue #1772, opened 2022-10-17
Hidden languages are listed in fallback translation text


timoludwig:
### Describe the Bug

The fallback text for missing page translations also lists existing public translations in languages which are disabled/hidden.

### Steps to Reproduce

1. Create a page which has no translation in e.g. "ar" but a public translation in "fa"
2. Mark "fa" as "not visible in app"
3. Request API for language "ar"
4. See fallback text which lists the option for "fa"

### Expected Behavior

Hidden and disabled languages should not be listed in the fallback translations

### Actual Behavior

They are listed if the translations itself are public

### Additional Information

Maybe this also has implications in other places of the cms - e.g. available languages in the API or in the sitemap. Maybe we should exclude these languages in the `prefetch_public_translations()` queryset method?

seluianova:
I tried to reproduce the issue within the PR review and could only get the described result when "ar" translation is created, but with empty content.

But when there is no "ar" translation (as described), the page in the response is not listed at all.

Did I misunderstand something?

Issue #1689, opened 2022-09-16
Add CircleCI check for changelog format


timoludwig:
### Motivation

The changelog is parsed by the following dark magic bash command:
https://github.com/digitalfabrik/integreat-cms/blob/9bc0251e90f312e80ed39226924b7993678d3790/.circleci/config.yml#L366
which requires exactly one line after the version heading and two new lines after the bullet points.

### Proposed Solution

Add a script which checks the format in CircleCI

### Alternatives

Improve the bash command to handle more changelog formats

timoludwig:
It's quite a common problem that the changelog entry does not end up in the unreleased section after rebasing, so we should definitely add a check for this as well.

david-venhoff:
Not sure if it is worth the effort in our case, but I found this solution interesting where every changelog entry got its own file: https://about.gitlab.com/blog/2018/07/03/solving-gitlabs-changelog-conflict-crisis/

timoludwig:
Yes, this is exactly the idea I had in mind when creating #1763.
Also not sure, but if we count all the developer hours which are spent with conflict resolution, maybe it will pay off in the long term :smile:

Issue #1441, opened 2022-05-17
Additional status for fallback translations


svenseeberg:
### Motivation
Currently, we show an "outdated translation" status for pages with no content but existing title.

### Proposed Solution
Show an additional status icon for translations with fallback translations.

timoludwig:
> Design Requirements
> 
> We need an icon for the translation status for fallback translations (@hauf-toni)

@JoeyStk I'm not sure whether we really need a design for this - we only use icons provided by https://lucide.dev/ inside the cms. So we would have to make sure the new design matches the existing icon set. Instead, I suggest that we just choose one of the available options.

I would suggest e.g.

- info, because the page contains an information text in which languages the page is available:
  ![info](https://user-images.githubusercontent.com/16021269/226762582-a87134cd-7603-416b-b7d8-04246e44810d.png)
- milestone, because the page redirects the user to another language:
  ![milestone](https://user-images.githubusercontent.com/16021269/226762707-97806adb-9033-46d6-af9a-fd9a23af12dc.png)

Just a reminder - these are the existing status icon:

| status | icon |
|---|---|
| up-to-date | ![check](https://user-images.githubusercontent.com/16021269/226762127-5d3cc198-6a21-4f26-9591-2616fb1f8d2b.png) |
| in translation | ![clock](https://user-images.githubusercontent.com/16021269/226762144-ff079c3e-5c8b-49b0-8c89-7aa6b34b24cd.png) |
| outdated | ![alert-triangle](https://user-images.githubusercontent.com/16021269/226762195-ec52d500-1130-4166-8deb-3a59d67bdd62.png) |
| fallback (when duplicated) | ![copy](https://user-images.githubusercontent.com/16021269/226761857-c1839f7c-8efc-4111-80df-4f5f9f6c92c4.png) |
| machine translated | ![bot](https://user-images.githubusercontent.com/16021269/226762181-61a2b617-ef8c-4fa2-baa7-b1121c07755e.png) |
| missing | ![x](https://user-images.githubusercontent.com/16021269/226762189-53c8db52-c118-498c-87f3-15edca97cadc.png) |

osmers:
Question regarding the suggestions: 
info - doesn't the page always contain an info text telling the user that the page is not available in their language and thus redirects them to another language. Or are you referring to pages where not all of the page is translated? E.g. when part of the text is provided by the Kommune and another part is live-content from another Kommune that is not available in all languages?
[...]

PR #1930, opened 2022-11-30
Bugfix: do not list hidden languages in fallback translation text in api

Gaston69:
### Short description

When a language is marked as hidden, it currently is still shown as part of the fallback translations for missing page translations when a list of all pages are requested from the api for a specific language. 
("This page does not exist in the selected language. It is however available in these languages: A, B, C")

Co-authored-by: @PeterNerlich 
### Proposed changes

- In order for a page translation to be included in the fallback-translation-text, the corresponding language has to be visible in the region

We added the following test case to the sample data / api response tests:
- A dummy language which is set to active, but not visible for region Augsburg
- A corresponding page translation on a page without an English translation
- An API response test for all pages in English, expecting the dummy language not to appear as a fallback. This failed previously, and now succeeds.

### Side effects

Test data has been extended and expected output for /api/augsburg/en/pages/ in test-system has been added

### Resolved issues

Fixes: #1772 


david-venhoff (approved):
Thanks, LGTM!
I guess it is not strictly required for this pr, but I noticed that there are other places in the cms where hidden languages are still shown. 
One example are mirrored pages: The alternative languages message can contain languages that are not visible. That should be fixable by modifying https://github.com/digitalfabrik/integreat-cms/blob/ad73c79b7b994b756997fb1019b63662345ff543/integreat_cms/cms/models/pages/page_translation.py#L142-L148
For example:
```python
translations = filter(
          None,
          [
              self.page.get_public_translation(language_slug)
              for (language_slug, translation) in self.page.mirrored_page.prefetched_major_public_translations_by_language_slug.items()
              if translation.language in self.page.region.visible_languages
          ],
      )
```
However as Timo mentioned in #1772 this might affect even more places and maybe its better to fix that with a more general solution in another pr.

integreat_cms/cms/fixtures/test_data.json
```diff
@@ -1885,6 +1919,40 @@
...
   },
+  {
+    "model": "cms.pagetranslation",
+    "pk": 82,
+    "fields": {
+      "title": "Teste Hidden Language",
+      "slug": "teste-hidden-language",
+      "status": "PUBLIC",
+      "content": "<p>Dies ist ein Inhalt in \"Hidden language\"</p>",
+      "language": 7,
```
seluianova, line 1930: Also it's not possible to create a page without German translation.
So it's probably better to have it in test data as well, for consistency

integreat_cms/cms/fixtures/test_data.json
```diff
@@ -355,6 +355,23 @@
       "message_partial_live_content_not_available": "Ein Teil von der Seite ist nicht in Ihrer Sprache. Der Teil ist aber in diesen Sprachen verfügbar:"
     }
   },
+  {
+    "model": "cms.language",
+    "pk": 7,
+    "fields": {
+      "slug": "hidden",
```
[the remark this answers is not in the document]
timoludwig: I think it's ok to add a new language to allow a few more edge cases in the test data which should increase test coverage in the long term.
Probably that's a bit nit-picky, but I would prefer a "real" language in the test data, e.g. Spanish, French or something (and set a flag).

integreat_cms/cms/fixtures/test_data.json
```diff
@@ -355,6 +355,23 @@
...
+      "bcp47_tag": "hid",
+      "native_name": "Hidden language",
+      "english_name": "Hidden language",
+      "text_direction": "LEFT_TO_RIGHT",
+      "primary_country_code": "hi",
+      "created_date": "2022-11-30T09:16:00.000Z",
+      "last_updated": "2022-11-30T09:16:00.000Z",
+      "table_of_contents": "Table of Contents",
+      "message_content_not_available": "This Language may exist but is marked as hidden. We test the behavior when it is marked as hidden:",
+      "message_partial_live_content_not_available": "This Language may exist but is marked as hidden."
```
timoludwig, line 372: Also, I'd use the correct fallback text here. The text above does not really make sense...

integreat_cms/cms/fixtures/test_data.json
```diff
@@ -508,6 +525,23 @@
       "last_updated": "2022-09-14T15:07:49.750Z"
     }
   },
+  {
+    "model": "cms.languagetreenode",
+    "pk": 10,
+    "fields": {
+      "lft": 3,
+      "rgt": 4,
+      "tree_id": 1,
```
timoludwig, line 534: Adding this kind of works, but leaves inconsistencies in the database. Inserting an element in an MPTT structure always changes at least the `rgt` value of the root node - each `lft` and `rgt` value may only exist once in each tree:

mptt

But you don't have to understand mptt in order to create correct tree structures - our tree library will do this for us. Just create the language via the GUI and then export the table via:
```
export INTEGREAT_CMS_SECRET_KEY="dummy"
export INTEGREAT_CMS_DEBUG=1
pipenv run integreat-cms-cli dumpdata cms.languagetreenode --settings=integreat_cms.core.docker_settings
```
And then overwrite the existing test tree nodes with the output.

integreat_cms/cms/fixtures/test_data.json
timoludwig: Yes, I agree! :+1:

timoludwig (changes requested):
Thanks a lot :+1:
This fixes the issue! However, I'd suggest to adapt the test data a bit.

integreat_cms/cms/fixtures/test_data.json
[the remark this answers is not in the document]
PeterNerlich: We discussed that and decided that it would make sense to separate test cases, especially here to make a dummy language that would not be used for real in the future and thus be less prone to accidental activation of the `visible` setting, breaking this test from a seemingly unrelated context.

integreat_cms/cms/fixtures/test_data.json
seluianova: @timoludwig side question: in this test we have english translation with title, but with empty content.
But when there is no english translation at all (no entry in DB), API request for "en" language doesn't return this page at all. Is it an expected behavior? Shouldn't it also return a page with the same fallback message?

timoludwig: No, this is the expected behavior. For pages, the fallback texts are only shown then the object exists in the database, but with an empty title.
Reason for this is that some regions want to create content for a specific target group, e.g. they add an info for people who speak Romanian but since this info is irrelevant for all other people, they want to hide it for them.
So we decided that the fallback mechanism for pages only works when at least the title is translated.
Also, a related issue is #1441.

CHANGELOG.md
```diff
@@ -1,7 +1,33 @@
...
+* [ [#1890](https://github.com/digitalfabrik/integreat-cms/issues/1890) ] Add hint about icon aspect ratio to location form
+* [ [#1889](https://github.com/digitalfabrik/integreat-cms/issues/1889) ] Make translation status independent from publishing status
+* [ [#1494](https://github.com/digitalfabrik/integreat-cms/issues/1494) ] Add a role without page editing permissions
+* [ [#1914](https://github.com/digitalfabrik/integreat-cms/issues/1914) ] Always uncheck minor edit field by default
+* [ [#1934](https://github.com/digitalfabrik/integreat-cms/issues/1934) ] Make sure translations are never a minor version after XLIFF import
+* [ [#1864](https://github.com/digitalfabrik/integreat-cms/issues/1864) ] Fix possibility to mark page as up-to-date without performing changes
+* [ [#1885](https://github.com/digitalfabrik/integreat-cms/issues/1885) ] Fix ongoing translation cancel button
+* [ [#1942](https://github.com/digitalfabrik/integreat-cms/issues/1942) ] Fix auto save functionality
+* [ [#1922](https://github.com/digitalfabrik/integreat-cms/issues/1922) ] Fix html escape in xliff import error message
```
timoludwig, line 30: This diff is very strange, could you squash your commits and rebase your branch onto the current develop branch?

integreat_cms/cms/fixtures/test_data.json
```diff
@@ -379,12 +379,29 @@
...
+  {
+    "model": "cms.language",
+    "pk": 7,
+    "fields": {
+      "slug": "hidden",
+      "bcp47_tag": "hid",
+      "native_name": "Hidden language",
+      "english_name": "Hidden language",
+      "text_direction": "LEFT_TO_RIGHT",
+      "primary_country_code": "hi",
```
timoludwig, line 391:
```suggestion
      "primary_country_code": "at",
```
`hi` is not part of the valid [choices](https://github.com/digitalfabrik/integreat-cms/blob/29313949a4756643590a83d93532c859bc4f269c/integreat_cms/cms/constants/countries.py#L10) for this model, so I suggest to use a different dummy-country-code.

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 UNRELEASED
 ----------
 
+* [ [#1772](https://github.com/digitalfabrik/integreat-cms/issues/1772) ] Fix hidden languages being listed in fallback translation text on the API endpoint
+
+
```
```suggestion
```
Exactly two new lines between the last bullet point and the next version heading are required. Not sure why the check didn't complain, probably this is an open todo for #1689.

timoludwig (approved):
Sorry it took me so long to revisit this.
Thanks a lot for your changes, other than two small details it looks good! :+1:

timoludwig:
@Gaston69 not a big deal, but next time try to rebase onto develop instead of merging when your feature branch is outdated and clean the git history up as described in the [docs](https://digitalfabrik.github.io/integreat-cms/git-flow.html#rebasing) before merging the PR, thanks!

Issue #1934, opened 2022-12-05
Minor edit field not changed after XLIFF import


timoludwig:
### Describe the Bug

When the newest version of a translation is a minor edit and this page is imported as XLIFF, the `minor_edit` field is set even after the import, meaning the translation status does not change.

### Steps to Reproduce

1. Create latest version with minor edit in target language
2. Export as XLIFF and re-import without changes
3. Confirm
4. See translation status still outdated

### Expected Behavior

An XLIFF import should never be considered a minor edit and always cause the page to be up to date after import

### Actual Behavior

The translation status stays outdated when latest version is minor edit

svenseeberg:
I have not yet spent time investigating the cause of #1914 - they could be related.

timoludwig:
> I have not yet spent time investigating the cause of #1914 - they could be related.

I think they are related from their topic, but require different solutions. I already have an idea, should I take over #1914 as well?

timoludwig:
Also, #1871 is related but a different problem I think (since in that case the publication status is different, not the minor edit field)

PR #1935, opened 2022-12-05
Make sure translations are never a minor version after XLIFF import

timoludwig:
### Short description

Make sure translations are never a minor version after XLIFF import

### Proposed changes

- When deserializing XLIFF files, always reset the minor edit field to `False`, independently of whether the existing database object is a minor or major edit.

### Resolved issues

Fixes: #1934


svenseeberg (approved):
The solution definitely ensures that this can never happen again :D

david-venhoff (approved):
Thanks, LGTM!

PR #1947, opened 2022-12-07
Suggest to make two reviews for every PR opened

timoludwig:
### Short description

We require two reviews per PR, so the math only works out when everybody makes two reviews for each of their own PRs they open. Otherwise, the pile of open PRs gets bigger and bigger until someone reviews more than twice as many PRs as they open.

### Proposed changes

- Suggest to make two reviews for every PR opened

### DIFF
--- a/sphinx/pull-request-review-guide.rst
+++ b/sphinx/pull-request-review-guide.rst
@@ -23,7 +23,7 @@ Goals and purpose
 
 .. tip::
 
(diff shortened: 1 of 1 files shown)


JoeyStk (approved):
Yeah! I'm halfway done until I can open a PR again! 
Thank you so much for this PR

david-venhoff (approved):

Issue #1953, opened 2022-12-08
Readable, discernable diffs


PeterNerlich:
### Motivation

Currently, our tests consist overwhelmingly of API tests, comparing the response of one broad API call with a saved string of the expected result.
This is a problem in and of itself, but the diffs generated when something breaks are far from helpful. It is not clear which part(s) have changed, and easy to miss a small change in a field somewhere when a whole subtree has changed in a different place.

### Proposed Solution

Use a pytest plugin producing discernable diffs, like `pytest-diff`.

### Alternatives

Use other pytest plugins, implement our own, live with extreme manual effort to make sense of failed tests, or, best of all, improve the tests to be more general and their output more meaningful.

### Additional Context

This issue became clear while working on #1930.

PR #1955, opened 2022-12-08
Readable test diffs with pytest-icdiff

PeterNerlich:
### Short description

This PR makes the output of failed API tests legible, by producing deep diffs of the JSON responses. It is possible to quickly see which values differ from the expected output.

### Proposed changes

- Add pytest-diff as dependency (no additional setup seems to be required)

### Side effects

- Tests are now producing different output. I hope this is not being parsed after the fact somewhere? (since pytest-diff is a plugin for pytest, pytest itself should not have issues)

### Resolved issues

Fixes: #1953 

### DIFF
--- a/dev-tools/test.sh
+++ b/dev-tools/test.sh
@@ -24,13 +24,21 @@ while [[ $# -gt 0 ]]; do
     case $1 in
         # If only tests affected by recent changed should be run, --changed can be passed as a flag
         --changed) CHANGED=1;shift;;
+        # Verbosity for pytest
+        -v|-vv|-vvv|-vvvv) VERBOSITY="$1";shift;;
         # If only particular tests should be run, test path can be passed as CLI argument
         *) TEST_PATH=$1;shift;;
     esac
 done
 
 # The default pytests args we use
-PYTEST_ARGS=("--quiet" "--disable-warnings" "--numprocesses=auto")
+PYTEST_ARGS=("--disable-warnings" "--color=yes")
+
+if [[ -n "${VERBOSITY}" ]]; then
+    PYTEST_ARGS+=("$VERBOSITY")
+else
+    PYTEST_ARGS+=("--quiet" "--numprocesses=auto")
+fi
 
 # Check if --changed flag was passed
 if [[ -n "${CHANGED}" ]]; then
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -89,6 +89,7 @@ dev = [
     "pytest-django",
     "pytest-testmon",
     "pytest-xdist",
+    "pytest-icdiff",
     "requests-mock",
     "shellcheck-py",
     "sphinx",
--- a/sphinx/dev-tools.rst
+++ b/sphinx/dev-tools.rst
@@ -87,7 +87,7 @@ Testing
 
 Run tests and generate coverage report with :github-source:`dev-tools/test.sh`::
 
-    ./dev-tools/test.sh [--changed] [TEST_PATH]
+    ./dev-tools/test.sh [--changed] [-v[v[v[v]]]] [TEST_PATH]
 
 **Arguments:**
 
@@ -96,6 +96,10 @@ Run tests and generate coverage report with :github-source:`dev-tools/test.sh`::
 **Options:**
 
 * ``--changed``: Run only tests affected by recent changes
+
+* ``-v``, ``-vv``, ``-vvv``, ``-vvvv``: Verbosity levels, passed directly to pytest.
+  Notice that if none are specified, we automatically pass ``--quiet``
+  and run tests on multiple CPUs using xdist, resulting in much shorter wait times.
 
 
 Code Quality


JoeyStk (approved):
Thank you so much for your PR. I had a little read on pytest-diff, and as for me we can try this out. But it would be good if someone with more experience on this could also have a look :)

timoludwig (commented):
Thanks! :+1:
In general I agree that the current test output isn't ideal and I'm fine with changing it (and we also don't parse this anywhere).

I'm not sure what I'm doing wrong, but somehow I cannot get this to work... I just changed a few small details in the expected API output, and the error still looks very unreadable:
```
>           assert result == response.json()
E           AssertionError: assert
E             [{'id': 35, 'url': 'http://localhost:8000/augsburg/de/behörden-und-beratung/behörden/', 'path': '/augsburg/de/behörden-und-beratung/behörden/', 'title': 'Behörden', 'modified_gmt': '2022-01-16T16:36:57.038Z', 'last_updated': '2022-01-16T17:36:57.038+01:00', 'excerpt': '', 'content': '', 'parent': {'id': 14, 'url': 'http://localhost:8000/augsburg/de/behörden-und-beratung/', 'path': '/augsburg/de/behörden-und-beratung/'}, 'order': 2, 'available_languages': {'en': {'id': 47, 'url': 'http://localhost:8000/augsburg/en/authorities-and-counselling/public-authorities/', 'path': '/augsburg/en/authorities-and-counselling/public-authorities/'}...
E
E             ...Full output truncated (11 lines hidden), use '-vv' to show

tests/api/test_api_result.py:46: AssertionError
```
And I'm pretty sure it looked exactly the same before. But maybe I have a different problem, does it work for you?
Also, I think the dependency itself isn't a good choice, mainly for these reasons:
1. It's only version number `0.1.14` and hasn't been updated for over three years, so probably it died during early development stage
2. The author of the package either deleted or renamed their GitHub username (https://github.com/username/pytest-diff isn't reachable anymore), so not sure where any maintenance would happen... If you think we should use this dependency nevertheless, I'd suggest you download the source code, create a new repository under the Digitalfabrik orga and we start maintaining this dependency ourselves. But probably this isn't worth the effort, maybe you could do a little bit of research to find a different package? Or maybe simply adding `-vv` as the error message suggests works as well?

Apart from that, the dev dependencies also need to be added to the Pipfile.lock, only adding them to the Pipfile leads to an inconsistent state and means that we don't install the same dependency versions on all machines anymore.

So please run `pipenv install --dev <package-name>` as described in the [docs](https://digitalfabrik.github.io/integreat-cms/virtualenv.html#add-dependencies) to update the lockfile as well, thanks!

svenseeberg:
Considering the state of pytest-diff, do you intend to continue on this PR? Otherwise we can probably close it and revisit the issue another time?

PeterNerlich:
Thank you for your patience! I did a bit more research and compiled some screenshots.

For the sake of comparison, **this is how it looks currently:**
image

---

And **this is how it looks with `pytest-diff`** – if you indeed pass `-vv` to pytest as well (it first regurgitates the whole strings, it looks like):
image
I guess when we tried it, Ingo was running individual tests and using `-vv` anyway to cut down on waiting time, so we didn't notice. The following options also require `-vv` to be helpful, but there is an [issue](https://github.com/pytest-dev/pytest/issues/3962) and [a PR](https://github.com/pytest-dev/pytest/pull/10588) upstream that would separate that out to its own setting.

---

But, `pytest-diff` still is of little help if the individual lines are very long, like in `tests/api/expected_outputs/`, which was what originally motivated this search. **Here `pytest-icdiff` seems pretty cool:**
image
image
image
This displays the versions side by side in two columns, which might be annoying if the detected width is not your terminal width (e.g. when viewing the Circle CI logs) and is not my first preference (I like the way GitHub highlights differences). But it cuts down on the clutter very nicely. Additions, changes and removals are highlighted on a character level.

---

Lastly, **there is `pytest-clarity`:**
image
image
image
This makes it a bit easier to spot the changes since lines with changes are printed in color with the concrete changed characters inverted, but it displays the whole thing – I left out *way more* from this than from the screenshots of `pytest-icdiff`.

---

**My personal verdict is that I like `pytest-icdiff` better than the other presented options**, including the option of not changing anything.

What I have not been able to figure out is how exactly running pytest as part of the `dev-tools/test.sh` is different from running e.g. `pipenv run pytest tests/api/test_api_result.py::test_api_result[/api/augsburg/de/pages/-/augsburg/de/wp-json/extensions/v3/pages/-tests/api/expected-outputs/augsburg_de_pages.json-200] -s --disable-warnings --ds=integreat_cms.core.docker_settings` directly. Running `dev-tools/test.sh` somehow shows the whole diff in red in addition to apparently not being able to identify the terminal width:
image

timoludwig:
Thanks for revisiting this! :+1:

> **My personal verdict is that I like `pytest-icdiff` better than the other presented options**, including the option of not changing anything.

I agree, this tools looks very promising!

> Running `dev-tools/test.sh` somehow shows the whole diff in red in addition to apparently not being able to identify the terminal width

Both problems seems to be caused by an incompatibility with [pytest-xdist](https://github.com/pytest-dev/pytest-xdist/), possibly related to https://github.com/pytest-dev/pytest-xdist/issues/819. I suggest to:
- [Open an issue in pytest-xdist](https://github.com/pytest-dev/pytest-xdist/issues/new)
- Provide a workaround to pass an additional option to the `test.sh` tool to disable the parallel execution to improve the output (run without the `--numprocesses=auto` flag).

PeterNerlich:
Found out that getting the colored output seems to be as easy as adding `--color=yes` to the pytest arguments. Improving the columns is not as easy and I will probably be submitting a PR to the project allowing for all of the arguments in icdiff to also be specified for pytest-icdiff.

I'd also like to wait until #2027 is merged

PeterNerlich:
I think I found a quite workable solution for now. Running `./dev-tools/test.sh` as usual now includes a properly coloured icdiff, through it thinks the terminal is 80 characters wide due to xdist specifics. For easy diffs with singular diffs or where the changes are near the start, this already helps a lot:
image

But sometimes the output is cut before the actual changes, when icdiff is still only showing the context.
In that case, one can now actually follow the instructions and add `-vv` to get verbose output (specifically, I made sure to pass on all levels from `-v` to `-vvvv` while I was at it). In that case, I opted to also disable parallel tests using xdist, meaning the tests will take more time, but the output won't be overloaded with xdist specific verbose messages in addition to the diffs utilising the full width of the terminal:
image

dev-tools/test.sh
```diff
@@ -18,19 +18,23 @@ export INTEGREAT_CMS_SUMM_AI_API_KEY="dummy"
...
+        # also disable multiple processes by xdist so diffs are full terminal width
+        -v|-vv|-vvv|-vvvv) VERBOSITY="$1v";NUMPROCESSES="";shift;;
         # If only particular tests should be run, test path can be passed as CLI argument
         *) TEST_PATH=$1;shift;;
     esac
 done
 
 # The default pytests args we use
-PYTEST_ARGS=("--quiet" "--disable-warnings" "--numprocesses=auto")
+PYTEST_ARGS=("--quiet" "$VERBOSITY" "--disable-warnings" "$NUMPROCESSES" "--color=yes")
```
timoludwig, line 37:
```suggestion
PYTEST_ARGS=("--disable-warnings" "--color=yes")

if [[ -n "${VERBOSITY}" ]]; then
    PYTEST_ARGS+=("$VERBOSITY")
else
    PYTEST_ARGS+=("--quiet" "--numprocesses=auto")
fi
```

I think it would make sense to only start with the arguments which are present for all test calls and then append all additional arguments based on the command line input

dev-tools/test.sh
```diff
@@ -18,19 +18,23 @@ export INTEGREAT_CMS_SUMM_AI_API_KEY="dummy"
...
+NUMPROCESSES="--numprocesses=auto"
 
 # Parse given command line arguments
 while [[ $# -gt 0 ]]; do
     case $1 in
         # If only tests affected by recent changed should be run, --changed can be passed as a flag
         --changed) CHANGED=1;shift;;
+        # Verbosity for pytest (one extra v to counteract --quiet)
+        # also disable multiple processes by xdist so diffs are full terminal width
+        -v|-vv|-vvv|-vvvv) VERBOSITY="$1v";NUMPROCESSES="";shift;;
```
timoludwig, line 30:
```suggestion
        # Verbosity for pytest
        -v|-vv|-vvv|-vvvv) VERBOSITY="$1";shift;;
```
Then we don't need to counteract `--quiet` here

dev-tools/test.sh
```diff
@@ -18,19 +18,23 @@ export INTEGREAT_CMS_SUMM_AI_API_KEY="dummy"
 
 # Disable linkcheck listeners during testing
 export INTEGREAT_CMS_LINKCHECK_DISABLE_LISTENERS=1
+NUMPROCESSES="--numprocesses=auto"
```
timoludwig, line 21:
```suggestion
```
And can remove this variable

timoludwig (changes requested):
Thanks a lot, makes sense! :+1:

Could you also [document](https://github.com/digitalfabrik/integreat-cms/blob/develop/sphinx/dev-tools.rst#testing) the new verbose flags and mention that they implicate performance drawbacks?

dev-tools/test.sh
PeterNerlich: Ah, I tried to group related arguments (quiet/verbose). Will do!

timoludwig (approved):
Thanks! :+1:

JoeyStk (approved):
LGTM

Issue #1956, opened 2022-12-08
PDF export not working for Amharic


timoludwig:
### Describe the Bug

The alphabet of Amharic is not supported by our PDF export at the moment

### Steps to Reproduce

1. Go to a page in Amharic, e.g. https://integreat.app/nuernberg/am/%E1%89%A3%E1%88%85%E1%88%8D/%E1%89%A3%E1%88%85%E1%88%8B%E1%8B%8A-%E1%8A%A0%E1%8B%B5%E1%88%AB%E1%88%BE%E1%89%BD
2. Export as PDF
3. See broken PDF

### Expected Behavior

We should support all languages in our ecosystem

### Actual Behavior

Amharic alphabet does not work

### Additional Information

https://cms.integreat-app.de/pdf/c94ef6023a/Integreat%20-%20Amharisch%20-%20%E1%89%A3%E1%88%85%E1%88%8B%E1%8B%8A%20%E1%8A%A0%E1%8B%B5%E1%88%AB%E1%88%BE%E1%89%BD.pdf
![Screenshot (866)](https://user-images.githubusercontent.com/16021269/206446947-41591a1d-512c-4c1b-95da-bfcc92f0a015.png)

PR #1707, opened 2022-09-20
Fix Cyrillic TOC

MizukiTemma:
### Short description

This PR tries to solve the problem that table of content in cyrillic letters is not shown in exported PDF.

Though there is an on-going discussion about PDF (library), I open this PR because this issue is probably urgent for the current situation.

### Proposed changes

- Replace the font set with a version which includes also cyrillic.
- Raleway font is used as before, but this does not support greek letters (also as before). The problem stays unsolved for greek letters but will be sloved for cyrillic. 

### Side effects

- so for not known

### Resolved issues

Fixes: #1532 

<hr>

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #1961, opened 2022-12-09
Add Amharic font and fix PDF export

MizukiTemma:
### Short description

This PR adds a font set for Amharic language so that PDF export in this language works.

The approach is the same with #1707 .

Attension: if you have exported some broken Amharic PDFs to reproduce the problem in other branches, please delete those PDFs in the folder integreat_cms/pdf/. Otherwise they stay cashed.

### Proposed changes

- Add the set of noto-sans-ethiopic. 

### Side effects

- hopefully none.

It's probably better to list all the languages accross all the regions and check whether they are all covered by our PDF feature, to prevent the same bug in future.

### Resolved issues

Fixes: #1956 


integreat_cms/static/src/fonts/noto-sans-ethiopic/noto-sans-ethiopic.css
```diff
@@ -0,0 +1,35 @@
...
+    src: url("./noto-sans-ethiopic-v42-latin-ext_latin_ethiopic-regular.eot"); /* IE9 Compat Modes */
+    src: local("Noto Sans Ethiopic"), local("NotoSansEthiopic-Regular"),
+        /* Local installed font */ url("./noto-sans-ethiopic-v42-latin-ext_latin_ethiopic-regular.eot?#iefix")
+            format("embedded-opentype"),
+        /* IE6-IE8 */ url("./noto-sans-ethiopic-v42-latin-ext_latin_ethiopic-regular.woff2") format("woff2"),
+        /* Super Modern Browsers */ url("./noto-sans-ethiopic-v42-latin-ext_latin_ethiopic-regular.woff") format("woff"),
+        /* Modern Browsers */ url("./noto-sans-ethiopic-v42-latin-ext_latin_ethiopic-regular.ttf") format("truetype"); /* Safari, Android, iOS */
+}
+
+/* noto-sans-sc-700 - chinese-simplified */
```
JoeyStk, line 23: I think this is just a copy-paste-issue. I think it's supposed to be "ethiopic -  simplified" or even "amharic-simplified", since their are a few languages spoken in Ethiopia which we might integrate in our system

JoeyStk (approved):
Thank you so much. This is an important feature for our CMS! 
I only found one tiny thing. Other than that the PDF export worked as expected :)

integreat_cms/static/src/fonts/noto-sans-ethiopic/noto-sans-ethiopic.css
MizukiTemma: Oh 😓 Yes, this was looked over. I'll change it to "ethiopic".
(I've chosen the term "ethiopic" since the font set is called so. If you think "it's better to use the term "amharic", I'll change them.)

timoludwig (approved):
If you want to, you could also add a language tree node and a page translation in Amharic, which would enable us to use this language in the tests, and then we could add another test PDF for this language.
But that's optional :wink:

MizukiTemma:
That's a good idea 😃 
I'll do it but in a separate PR so this one can be merged asap.

Issue #1945, opened 2022-12-07
Make message and button in list and form of page/event/poi uniform for observer users


MizukiTemma:
### Motivation

For users with observer roll, a message "You don't have the permission to edit or create pages/events." will be shown when they open the event list or poi list, and they are also not allowed to type anything in the form when they open a page or event. However the message will not be shown in the location list and they can type in a location form.

Furthermore, the create page button is hidden for them in the page list, but the create button in event and location list is disabled but shown.

### Proposed Solution

- Add the message in the location list.
- Disable the location form for observer users.
- Hide the create buttons in event and location lists.

### Additional Context

To see those behaviours, log in as observer user (for example as Region Observer, which is given in the default data).

PR #1966, opened 2022-12-12
Make events, pois and pages uniform for observer users

david-venhoff:
### Short description

There are some inconsistencies regarding the poi/event/page forms and list views for observer users.
This pr tries to fix that.

### Proposed changes

- Add no editing permission message to the location list view
- Disable the location form for users without editing permission
- Hide the create new poi/event buttons for users without the create object permission
- Don't show invalid warning in the event form for users without publishing and editing permissions

### Side effects

/

### Resolved issues

Fixes: #1945 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 UNRELEASED
 ----------
 
+* [ [#1945](https://github.com/digitalfabrik/integreat-cms/issues/1945) ] Make message and button in list and form of page/event/poi uniform for observer users
+
 
 2022.12.2
 ---------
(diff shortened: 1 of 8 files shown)


JoeyStk (changes requested):
First of all, thank you a lot for this PR. I wasn't sure, if the PR is already up for review, but since it's not a draft I was guessing yes. 
So I think the functionality is there already and working as expected. All three requirements that were defined in the issue are met and work on my machine.
Other than that I only have thoughts on the failed test and I think you could squash your commit history a little bit, but I'm guessing that you already have an eye on that

JoeyStk (approved):
Thank you for updating, I think this looks good now! :)

MizukiTemma (changes requested):
Thank you for your PR 🙂 It looks nice.

I have only a small suggestion.

MizukiTemma (approved):
Thank you! very cool 🚀 

After squashing the commits, it's ready for merge 😃

Issue #1736, opened 2022-10-05
Fields for page-specific permissions are inconsistently named


timoludwig:
### Motivation

The fields for page-specific permissions of the `Page` model:

- The `editors` field contains all users who have the permission to edit (but not publish) the page:
   https://github.com/digitalfabrik/integreat-cms/blob/d68c72f9274ec0d7691dc914250194ac062691a1/integreat_cms/cms/models/pages/page.py#L159-L170
- The `publishers` field contains all users who have the permission to edit and publish the page:
   https://github.com/digitalfabrik/integreat-cms/blob/d68c72f9274ec0d7691dc914250194ac062691a1/integreat_cms/cms/models/pages/page.py#L171-L182

### Proposed Solution

Since our roles are called ["Author"](https://github.com/digitalfabrik/integreat-cms/blob/d68c72f9274ec0d7691dc914250194ac062691a1/integreat_cms/cms/constants/roles.py#L20) for the users who have the permission to edit (but not publish) a page and ["Editor"](https://github.com/digitalfabrik/integreat-cms/blob/d68c72f9274ec0d7691dc914250194ac062691a1/integreat_cms/cms/constants/roles.py#L19) for the users who have the permission to edit & publish the page, I suggest to rename
- `editors` to `authors` and
- `publishers` to `editors`.

### Additional Context

Make sure that the data migration preserves existing users in these fields (so the renaming should not lead to privilege escalations of the users specified in the `editors` field).

PR #1971, opened 2022-12-14
Rename role names

JoeyStk:
### Short description
Rename two role names

### Proposed changes

- Rename editors to authors
- Rename publishers to editors

### Side effects
I hope none

### Resolved issues

Fixes: #1736 


integreat_cms/cms/forms/pages/page_form.py
```diff
@@ -181,10 +181,10 @@ def _clean_cleaned_data(self):
 
     def get_editor_queryset(self):
```
david-venhoff, line 182: Maybe this function should be renamed? Right now I find it a bit confusing since it does not actually return the page editors anymore

integreat_cms/cms/forms/pages/page_form.py
```diff
@@ -202,16 +202,16 @@ def get_editor_queryset(self):
         )
         if self.instance.id:
             users_without_permissions = users_without_permissions.difference(
-                self.instance.editors.all()
+                self.instance.authors.all()
             )
         return users_without_permissions
 
     def get_publisher_queryset(self):
```
david-venhoff, line 209: Same here

integreat_cms/cms/rules.py
```diff
@@ -26,7 +26,7 @@
 @predicate
 def is_page_editor(user, page):
```
david-venhoff, line 27: These two methods could probably also get renamed

david-venhoff (commented):
Thanks! I tested the pr and it mostly works, but I found a problem with the migration: If a user has a page-specific permission to edit (not publish) a page then after the migration the user will have permission to also publish the page. This is because the `editors` field changes its meaning but keeps the same list of users. 
What should happen I think is that the users of the old `editor` field should get added to the `author` field and users of the `publisher` field should get added to the new `editors` field

david-venhoff (commented):
I think for consistency the other editor & publisher occurences in the top-comment in this file could also be changed, 
https://github.com/digitalfabrik/integreat-cms/blob/1e7e04e6ac81be0f8542e02276919560be9a4247/integreat_cms/cms/rules.py

integreat_cms/cms/templates/pages/_page_permission_table.html
```diff
@@ -74,11 +74,11 @@
...
 </div>
 <div class="w-full flex flex-wrap mt-2">
-    {% render_field page_form.publishers %}
+    {% render_field page_form.editors %}
     <button class="btn grant-page-permission mt-2 w-full"
             data-url="{% url 'grant_page_permission_ajax' region_slug=request.region.slug %}"
             data-page-id="{{ page.id }}"
             data-permission="publish">
-        <i icon-name="plus-circle" class="mr-2"></i> {% translate "Add to editors" %}
+        <i icon-name="plus-circle" class="mr-2"></i> {% translate "Add to authors" %}
```
timoludwig, line 82:
```suggestion
        <i icon-name="plus-circle" class="mr-2"></i> {% translate "Add to editors" %}
```
In this particular case, the string was already correct (this is the text shown in the GUI which was fixed earlier)...

integreat_cms/cms/migrations/0055_alter_author_and_editor_role_name.py
```diff
@@ -0,0 +1,43 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    """
+    Add a migration to rename the author and editor roles for more consistency
+    """
+
+    dependencies = [
+        ("cms", "0054_user_totp_key"),
```
timoludwig, line 13:
```suggestion
        ("cms", "0055_track_region_deepl_api_usage"),
```

timoludwig (changes requested):
Thanks! :+1:

A few things were still missing, have a look at my suggestions below.

Also, there is this documentation patch which needs to be adapted as well:
https://github.com/digitalfabrik/integreat-cms/blob/develop/sphinx/patches/cms.diff

MizukiTemma:
@JoeyStk 

I fixed some fields that were not adjusted to the new role names so the test runs successfully.
If I have overlooked any logic or constraints and this is really not a solution, please drop the last 2 commits (I hope in this case the rebase won't harm anything at least).

david-venhoff (approved):
Thanks, I think this is good to go :rocket:

Issue #1906, opened 2022-11-28
Fix link escape in message in imprint form


MizukiTemma:
### Describe the Bug

A message will be shown in the imprint form if the standard language is currently duplicated for the given language and it contains a link to the version which is shown in the app. Thogh the link is not well escaped and the tag is directly shown.

### Steps to Reproduce

1. Go to Imprint in the menu.
2. Click on the tab of a translation for which the default language is duplicated.
3. See the message on the top of the form.

### Expected Behavior

A link is shown.

### Actual Behavior

A tag is shown.

### Additional Information

![not_well_escaped_link](https://user-images.githubusercontent.com/82345046/204389202-d3e95ff6-7d71-4c38-8a36-a386df4a2089.png)

The link target is probably also incorrect. It doesn't make sense that revision 1 is shown while there is no version yet for the language. (See the screen shot. No version has been yet created (and therefore the default language is duplicated) for Easy German but the message says version 1 of Easy German is shown in the app.)

seluianova:
> It doesn't make sense that revision 1 is shown while there is no version yet for the language

So, if there is no version yet, I would guess we don't need this message at all. Do we?

Also, there is no such message for other entities (pages, pois, events).
Why do we have it for imprint only?

seluianova:
> Also, there is no such message for other entities (pages, pois, events).

I just did some digging into the code and found that we have a similar message for pages:

`                    message = _(
                        "Currently, <a>version {}</a> of this page is displayed in the app."
                    ).format(public_translation.version)`

However if I understand it correctly, it is not reachable. Because when the last page translation is DRAFT, all previous are also set to DRAFT.
If this is correct, we should probably also remove this block of code from page view to avoid confusion

timoludwig:
> However if I understand it correctly, it is not reachable. Because when the last page translation is DRAFT, all previous are also set to DRAFT. If this is correct, we should probably also remove this block of code from page view to avoid confusion

There are two exceptions to this rule:
1. When there are autosaves or in-review pages, not the newest version is shown the app, but only the newest public version.
2. When old revisions are restored as draft from the revision slider, only the restored version is a draft and the previous version will be still published.

We still have it on our plate to create a better UI for these kind of workflows, so in theory it should be a bit easier to work on a new state while still delivering the old state in the apps, but I'm not sure whether there is even an open ticket for it.

seluianova:
@timoludwig now I could reproduce it, thank you!

PR #1976, opened 2022-12-16
Fix link escape in message in imprint form

seluianova:
### Short description
Link is displayed as HTML tags in imprint form:
https://user-images.githubusercontent.com/82345046/204389202-d3e95ff6-7d71-4c38-8a36-a386df4a2089.png
Also link itself is incorrect, when there is no public version for the translation (it routes to Revision 1, but this revision is in draft state)

### Proposed changes
- Use format_html function to display link as a link
- Don't show the message, when there is no public version of the translation in the requested language.
I added condition by 'public_translation.id' for that. When it is 'None', it means that fallback translation is used.

### Side effects
Didn't find any

### Resolved issues
Fixes: #1906

### DIFF
--- a/integreat_cms/cms/views/imprint/imprint_form_view.py
+++ b/integreat_cms/cms/views/imprint/imprint_form_view.py
@@ -15,6 +15,7 @@ from ...forms import ImprintTranslationForm
 from ...models import ImprintPageTranslation, ImprintPage
 from ...constants import status
 from ...utils.content_edit_lock import get_locking_user
+from ...utils.translation_utils import gettext_many_lazy as __, translate_link
 
 logger = logging.getLogger(__name__)
 
@@ -92,25 +93,38 @@ class ImprintFormView(TemplateView, MediaContextMixin):
 
         disabled = False
         if imprint:
-            # Show information if latest changes are only saved as draft
+            # Show information if latest changes are only saved as draft, but there is an earlier public version of this translation
             public_translation = imprint.get_public_translation(language.slug)
-            if public_translation and imprint_translation != public_translation:
+            if (
+                public_translation
+                and imprint_translation != public_translation
+                and public_translation.id  # checking that public translation is not a fallback translation
+            ):
+                public_translation_url = reverse(
+                    "imprint_revisions",
+                    kwargs={
+                        "region_slug": region.slug,
+                        "language_slug": language.slug,
+                        "selected_revision": public_translation.version,
+                    },
+                )
+                message = __(
+                    _(
+                        "This is not the most recent public revision of this translation."
+                    ),
+                    _(
+                        "Instead, <a>revision {}</a> is shown in the apps.",
+                    ).format(public_translation.version),
+                )
                 messages.info(
                     request,
-                    _(
-                        "This is <b>not</b> the most recent public revision of this translation. Instead, <a href='%(revision_url)s' class='underline hover:no-underline'>revision %(revision)s</a> is shown in the apps."
-                    )
-                    % {
-                        "revision_url": reverse(
-                            "imprint_revisions",
-                            kwargs={
-                                "region_slug": region.slug,
-                                "language_slug": language.slug,
-                                "selected_revision": public_translation.version,
-                            },
-                        ),
-                        "revision": public_translation.version,
-                    },
+                    translate_link(
+                        message,
+                        attributes={
+                            "href": public_translation_url,
+                            "class": "underline hover:no-underline",
+                        },
+                    ),
                 )
 
         # Make form disabled if user has no permission to manage the imprint
@@ -122,11 +136,6 @@ class ImprintFormView(TemplateView, MediaContextMixin):
 
         imprint_translation_form = ImprintTranslationForm(
             instance=imprint_translation, disabled=disabled
-        )
-
-        # Pass side by side language options
-        side_by_side_language_options = self.get_side_by_side_language_options(
-            region, language, imprint
         )
 
         # If the imprint does not exist yet, create the key manually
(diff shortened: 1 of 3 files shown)


integreat_cms/cms/views/imprint/imprint_form_view.py
```diff
@@ -92,25 +93,34 @@ def get(self, request, *args, **kwargs):
...
         if imprint:
-            # Show information if latest changes are only saved as draft
+            # Show information if latest changes are only saved as draft, but there is an earlier public version of this translation
             public_translation = imprint.get_public_translation(language.slug)
-            if public_translation and imprint_translation != public_translation:
+            if (
+                public_translation
+                and imprint_translation != public_translation
+                and public_translation.id
+            ):
```
timoludwig, line 102:
```suggestion
            if public_translation and imprint_translation != public_translation:
```
Hmm, I don't really understand this change. `get_public_translation()` either returns `None`, or a translation object. If it returns a translation object, `id` will definitely be set, so I think in this context, the conditions `if public_translation` and `if public_translation.id` are equivalent... or am I missing something here?

integreat_cms/cms/views/imprint/imprint_form_view.py
```diff
@@ -92,25 +93,34 @@ def get(self, request, *args, **kwargs):
...
-                            "imprint_revisions",
-                            kwargs={
-                                "region_slug": region.slug,
-                                "language_slug": language.slug,
-                                "selected_revision": public_translation.version,
-                            },
+                    format_html(
+                        "{} <a href='{}' class='underline hover:no-underline'>{} {}</a> {}",
+                        _(
+                            "This is not the most recent public revision of this translation. Instead,"
```
timoludwig, line 117:
```suggestion
                    translate_link(
                        message,
                        attributes={
                            "href": public_translation_url,
                            "class": "underline hover:no-underline",
                        },
```
The problem with splitting the sentence is that the order of the words might be different in other languages and it makes it harder to grasp the context of the translations.
Instead, I think this would be a good use case for my [translate_link()](https://github.com/digitalfabrik/integreat-cms/blob/45a836a6fb8426d0183a8c760b1895dba930c733/integreat_cms/cms/utils/translation_utils.py#L26-L48) utility I recently implemented.

integreat_cms/cms/views/imprint/imprint_form_view.py
```diff
@@ -92,25 +93,34 @@ def get(self, request, *args, **kwargs):
...
                         ),
-                        "revision": public_translation.version,
-                    },
+                        public_translation_url,
+                        _("revision"),
+                        public_translation.version,
+                        _("is shown in the apps"),
+                    ),
```
timoludwig, line 123:
```suggestion
```

integreat_cms/cms/views/imprint/imprint_form_view.py
```diff
@@ -92,25 +93,34 @@ def get(self, request, *args, **kwargs):
...
+            ):
+                public_translation_url = reverse(
+                    "imprint_revisions",
+                    kwargs={
+                        "region_slug": region.slug,
+                        "language_slug": language.slug,
+                        "selected_revision": public_translation.version,
+                    },
+                )
+
```
timoludwig, line 111:
```suggestion
                message = __(
                    _(
                        "This is not the most recent public version of this translation."
                    ),
                    _(
                        "Instead, <a>version {}</a> is shown in the apps.",
                    ).format(public_translation.version),
                )
```

timoludwig (commented):
Thanks a lot! :+1:
Works as expected, I just have a small idea for improvement.

integreat_cms/cms/views/imprint/imprint_form_view.py
seluianova: > If it returns a translation object, id will definitely be set

Not really.
When public_translation is not None AND public_translation.id is not None --> returned translation is in the requested language.
When public_translation is not None BUT public_translation.id is None --> returned translation is in fallback language.

seluianova: So with the new condition by public_translation.id we won't display the message, when public translation is a fallback translation, otherwise it produced incorrect link.

integreat_cms/cms/views/imprint/imprint_form_view.py
seluianova: Yep, looks better indeed. Thanks

integreat_cms/cms/views/imprint/imprint_form_view.py
timoludwig: Ahh ok got it, thanks for the explanation! And also thanks for the comment, I think this is very helpful so we don't accidentally remove this condition in the future.

integreat_cms/cms/views/imprint/imprint_form_view.py
```diff
@@ -8,13 +8,16 @@
 from django.utils.translation import gettext as _
 from django.views.generic import TemplateView
 
+from integreat_cms.cms.utils.translation_utils import translate_link
+
```
timoludwig, line 12:
```suggestion
```
Could you also use a relative import for this?

integreat_cms/cms/views/imprint/imprint_form_view.py
```diff
@@ -8,13 +8,16 @@
...
 from ...constants import translation_status
 from ..media.media_context_mixin import MediaContextMixin
 from ...decorators import permission_required
 from ...forms import ImprintTranslationForm
 from ...models import ImprintPageTranslation, ImprintPage
 from ...constants import status
 from ...utils.content_edit_lock import get_locking_user
+from ...utils.translation_utils import gettext_many_lazy as __
```
timoludwig, line 20:
```suggestion
from ...utils.translation_utils import gettext_many_lazy as __, translate_link
```

integreat_cms/cms/views/imprint/imprint_form_view.py
seluianova: Sure

seluianova:
Applied suggested changes and removed one local variable, because pylint complained "Too many local variables"

timoludwig (approved):
LGTM, thanks! :+1:

MizukiTemma (approved):
Thank you 😃 It works as expected 👍

Issue #1695, opened 2022-09-17
Add warning category to broken link checker


david-venhoff:
### Motivation

For some pages the link checker shows an error, even though the page technically works for users. For this reason, a warning category should be added.

### Proposed Solution

Add a warning category which should include the following links:

- Links to pages with status code 3xx
- Links with invalid fragment
- Links to pages with status code 403 because the page might just not be available to the link checker, even though it works for users.

timoludwig:
> And where are the links saved (db-table)?

We use [django-linkcheck](https://github.com/DjangoAdminHackers/django-linkcheck) for this tasks, this library ships two [models](https://github.com/DjangoAdminHackers/django-linkcheck/blob/master/linkcheck/models.py): `Url` which contains the url string, the status and message, and `Link`, which holds a many to many mapping from translation models to the `Url` model (to de-duplicate the checking process and check urls which appear in multiple pages only once).

> I would add them to the fixtures.

Please do not include the real life example links below into our fixtures, they are just for testing purposes. Instead, I would recommend to use https://httpstat.us/ for the fixtures... you can e.g. generate links like:

- https://httpstat.us/200#broken-anchor
- https://httpstat.us/301 
- https://httpstat.us/302
- https://httpstat.us/403

(Only adding the `Link` objects leads to an inconsistent test database - ideally you would also create a few new page translations which contain these links... then you could also do this the other way round: add the links via the content editor first, then export the data via the [dumpdata](https://docs.djangoproject.com/en/4.1/ref/django-admin/#dumpdata) command)

> Can anybody please give examples for the mentioned links:
3xx, 200 with invalid fragments, 403

Some example links for the warning category:

### "Broken external hash anchor"
When the message is "Broken external hash anchor", the link usually works, some examples:

- https://www.augsburg.de/umwelt-soziales/gesundheit/coronavirus/diagnosestelle#c1098258
- https://www.vhs-lernportal.de/wws/9.php#/wws/home.php
- https://www.krankenkassenzentrale.de/wiki/international#

So the intention of this warning category would be to encourage content creators to either fix the anchor or remove everything after the `#`...

### "301 Moved Permanently" and "302 Found"
When the message is "301 Moved Permanently" or "302 Found", the link also works, but is just a redirect...

- https://www.zusammengegencorona.de/impfen/basiswissen-zum-impfen/impfmythen/#faqitem=b3a900e6-8fee-5b1d-828b-430395c12ac2
[...]

PR #1977, opened 2022-12-18
Update linkcheck dependency

timoludwig:
### Short description

Finally, our upstream contributions to the linkcheck library have been released! :tada:

### Proposed changes

- Update dependencies (including linkcheck version [2.0.0](https://pypi.org/project/django-linkcheck/2.0.0/))
- Remove workaround for linkcheck app
- Fix the link test data

### Side effects

A lot of [other improvements](https://github.com/DjangoAdminHackers/django-linkcheck/blob/02c6db5d4bcc2db6480637a27314a59a2ba56c05/CHANGELOG#L3-L23) to the link checker, the most important being:
- No crash of the process anymore when long URLs occur
- The `findlinks` command now deletes outdated links - no need to delete the entire link database anymore
- Broken anchors are now considered valid links instead of broken links (this means we do not need a separate warning category, see #1695)
- SSL certificates are now verified

### Resolved issues

Fixes: #1695
Fixes: #1818


ulliholtgrave (approved):
This looks very nice and seems to solve some issues :star_struck: 

As we are already updating our dependencies I would like to ask you to update our whole dependencies and bump TinyMCE to "^5.10.7". This would also fix the vulnerability of https://github.com/digitalfabrik/integreat-cms/security/dependabot/17.

JoeyStk (approved):
Thank you a lot! This also worked on my machine. My only thought was that maybe we could add an entry to the changelog. I know that for changes that are mostly relevant to developers we don't add an entry, but since this PR fixes issues that are also relevant for members of the service-team it could be beneficial to add one. What do you think? :)

timoludwig:
> My only thought was that maybe we could add an entry to the changelog. I know that for changes that are mostly relevant to developers we don't add an entry, but since this PR fixes issues that are also relevant for members of the service-team it could be beneficial to add one. What do you think? :)

Ok, I added one entry for the broken hash anchors, or which changes did you mean?
The other changes are probably more infrastructure-bound, since the issues will not be fixed as soon as the new version is released and installed, but only when we perform the next re-scan of the links...

Issue #1739, opened 2022-10-05
Translating pages with many paragraphs via SUMM.AI runs into rate limiting


timobrembeck:
### Describe the Bug

Translating a page which has many paragraphs (or bullet list items) via SUMM.AI may be impossible due to their rate limiting of 15 requests per second (in reality, I managed to send more requests than documented, but at least this is the official limit they're claiming).

### Steps to Reproduce

1. Create page with many HTML segments (`<p>` or `<li>` tags)
2. Translate page into Easy German via SUMM.AI
3. See error (perhaps)

### Expected Behavior

The SUMM.AI API client should inspect the response, and if it is the error which indicates that too many request were sent (HTTP status `429`, see [docs](https://summai.notion.site/SUMM-API-Documentation-fc41fa5ef1434ad2b311be4876b4685a)), delay all other parallel tasks, wait for one minute, resend the failed request(s) and continue.

### Actual Behavior

Currently, the client does not respect the quota and just sends as many requests as there are HTML segments.
Also, it does not stop if it encounters one error - it just finishes all other requests which causes the server to crash and return a less meaningful error message (just internal server error in HTML instead of JSON).

PR #1980, opened 2022-12-19
Respect rate limiting of SUMM.AI API

timoludwig:
### Short description

@Gaston69 @PeterNerlich I took the liberty to open a draft PR from your branch to provide a bit of preliminary feedback.

### Proposed changes

- Pause requests to SUMM.AI API when rate limiting error occurs

### Resolved issues

Fixes: #1739


timoludwig (commented):
Thanks a lot for tackling this issue :muscle: 
I realize this a very complex topic, so kudos that it already goes in the right direction :smile: 

I like the general approach of being "greedy" with the rate limiting and only wait if the server returns the error code, in my experience the server can handle a lot more load than the 15 requests/second which are stated in the API documentation :sweat_smile: 

Nevertheless, I see a few problems with this:
1. The rate limiting code does not seem to work yet, when I tried this out with e.g. 100 pages, it immediately sent all requests out and only waited for 10 of them at a time, but the API server just cares about when a request is started, not when it's response is received on our end.
2. The tests are only testing small components of the implemented classes, and not how they are used in context. So instead of testing each single private method of the API client and the queue, I think it makes more sense to test the view itself - use the existing `test_auto_translate_easy_german` as a guideline how this is done and experiment a bit with it - modify the settings to use lower thresholds, export more pages at once to create more tasks in the queue and check in the end whether each page has been translated completely - then, we can be much more sure that our complete implementation works, not only the internal helper classes, but also how they are used inside the CMS. In my opinion, this also makes the queue tests a bit superfluous - I guess if we translate a queue of 100 pages and in the end all text segments are correctly translated, this kind of proofs that the `next()` method of the queue works etc...
For this purpose, I would also really recommend to go with the existing testing framework of `aiohttp` instead of the custom `MockedRequest` - this way, we could also program a small dummy server which really runs into rate limiting sometimes:
```
async def fake_summ_ai_server(request):
    """
    Create fake responses which simulate the SUMM.AI API server

    :param request: The request
    :type request: aiohttp.web.Request

    :return: The response
    :rtype: aiohttp.web.Response
    """
    if random.getrandbits(1):
        return web.json_response(
            {
                "translated_text": "Hier ist Ihre Leichte Sprache Übersetzung",
                "jobid": "9999",
            }
        )
    else:
        return web.json_response(
            {
                "error": "Too many requests. Please wait and resend your request.",
            }, status=429
        )
``` 
Obviously, it would be even better if this fake server would deterministically return the rate limiting error, e.g. always on the first time, so we can correctly test whether our view re-tries failed requests correctly.

Apart from that, you seem to have some misunderstandings in your git workflow - Some commits which are already on the develop branch, are copied on your branch with a different commit hash (e.g. 0f886f5d14b018a1178186b693a71a39b2cc383e is the correct commit on develop, and on your branch you duplicated it as 4c312cb51f1cb11032bf6614667969d5a3ebc8d1). This is also the reason that the diff of this PR looks weird and includes changes which are already merged into develop. I can only guess how you ended up with this git history, but my assumption is that instead of rebasing your feature-branch onto develop, you seem to either do the opposite (rebasing the develop branch onto your feature branch) or cherry-pickinng specific commits... Also, commit 30c6ed66c905cdac358614ae64c525b9d630b538 looks a bit weird - did you just copy + paste the latest changes from develop into your branch and created a commit from it? This is normally not the intended way of resolving git conflicts - I would suggest to strictly follow the commands described in the [documentation](https://digitalfabrik.github.io/integreat-cms/git-flow.html#rebasing) - and if they don't work for you, we can try to find out the reason for this.

integreat_cms/summ_ai_api/summ_ai_api_client.py
```diff
@@ -85,21 +92,19 @@ async def translate_text_field(self, session, text_field):
...
-                    raise SummAiException(
-                        f"Unexpected API result: {response.status} - {response_data!r}"
-                    )
-                # Let the field handle the translated text
-                text_field.translate(response_data["translated_text"])
-                return text_field
+                if self.check_rate_limit_exceeded(response.status):
+                    # !!! we are not lucky with switching types of return values (text_field when suceeded, string when waiting, none when error)
+                    # what do you suggest?
+                    return self.SIGNAL_WAIT_FOR_NEXT_REQUEST
```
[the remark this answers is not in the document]
PeterNerlich: Yes! This was also what I was thinking about, but in that case I'd strongly prefer if it would be a subclass of `SummAiException` because it just seems sensible – but then we would need to somehow catch any `SummAiException` *except* if it's a `SummAiRateLimitingExceeded`. I looked into this a bit, and I don't think there is a clean way to do this without simply re-raising the exception.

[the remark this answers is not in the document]
timoludwig: Yes, I agree. I mean you could also use `SummAiException` as base class and define `SummAiRateLimitingExceeded(SummAiException)` and `SummAiRuntimeException(SummAiException)` or whatever as subclasses of it. And then raise a `SummAiRuntimeException` or whatever in all non-rate-limiting cases...

integreat_cms/summ_ai_api/summ_ai_api_client.py
```diff
@@ -130,16 +137,43 @@ async def translate_text_fields(self, loop, text_fields):
...
+                This form makes it easy to always have at most n concurrent tasks
+                as well as intermittent wait times through the task generator.
+
+                :returns: A list of task-results
+                :rtype: List
+                """
+                completed = []
+                while len(task_generator):
+                    task = next(task_generator)
+                    result = await task
```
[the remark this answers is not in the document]
PeterNerlich: I see. Yes, we didn't test this out yet. Ideally we would have worked our way through to 100% test coverage on roughly all abstraction levels, but we already had enough trouble trying to get the little we already have, so I don't think we will complete this before opening the PR for real.
Moving `loop.create_task()` down into the worker should not yield any complications I can think of.

integreat_cms/summ_ai_api/utils.py
```diff
@@ -392,3 +395,96 @@ def __repr__(self):
...
+
+        :param task: The task that failed because of the rate limiting
+        :type task: asyncio.Future
+        """
+
+        # Only save current timestamp if this is the first failed request reported
+        if (
+            self.last_rate_limit is None
+            or time.time() - self.last_rate_limit < self.wait_time
+        ):
```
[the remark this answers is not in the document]
PeterNerlich: This is a mistake from copying the condition from lines 444–448, thanks for noticing. We expected to find these kinds of mistakes by writing a sufficient number of tests.

tests/summ_ai_api/summ_ai_test.py
```diff
@@ -222,3 +232,142 @@ async def test_summ_ai_error_handling(
...
+    response_data_translation_succeeded = {
+        "dummy": "dummy",
+        "translated_text": "Das ist einfaches Deutsch.",
+    }
+    assert (
+        SummAiApiClient(
+            MockedRequest(), "form_dummy"
+        ).check_translated_text_is_in_response(response_data_translation_succeeded, 200)
+        is True
+    ), "if translated_text found in reponse-data, check_translated_text_is_in_response should return True"
```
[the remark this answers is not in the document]
PeterNerlich: This is intended to test `SummAiApiClient.check_translated_text_is_in_response()`. Currently it *is* mostly a stub that seems obvious and slightly unnecessary. For philosophy about why this might still be sensible in the long run I suggest asking Ingo or me separately from this PR, I have a feeling this will become quite the extensive discussion.

tests/summ_ai_api/summ_ai_test.py
```diff
@@ -222,3 +232,142 @@ async def test_summ_ai_error_handling(
...
+
+    tasks = [1, 2, 3]
+    # for the test, waiting-time is reduced to 0.5s, default is 60s defined in settings.SUMM_AI_WAIT_FOR_NEXT_REQUEST_AFTER_LIMIT_EXCEEDED
+    task_generator = PatientTaskQueue(tasks, 0.5)
+
+    # Rate limit hit
+    task = next(task_generator)
+    assert task in tasks
+    # we call out a reach the rate limit
+    task_generator.hit_rate_limit(task)
```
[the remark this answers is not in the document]
PeterNerlich: We intended to write tests for (nearly) all abstraction levels, both high level, overarching ones, similar to those that currently make up the bulk of all tests, as well low level ones, testing only single components, ensuring the origin of a breaking change can be quickly located.

tests/summ_ai_api/summ_ai_test.py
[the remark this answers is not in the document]
timoludwig: Don't get me wrong - philosophically we are 100% on the same page. I would love to see us get such test coverage in our CMS!
However, since we have unlimited potential for improvement but limited resources, I think this is a utopia we will never achieve. That doesn't mean we shouldn't try - but I think we should carefully choose the order in which we do things. That's also why we spend so much time prioritizing issues - because we know we can't solve all problems at once, we do the things first that will have the biggest impact. To apply this principle to testing, I would suggest that we write tests for complex functions that are most likely to break first, and only then write tests for trivial functions (which may or may not become complex in the future).

tests/summ_ai_api/summ_ai_test.py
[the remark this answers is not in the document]
Gaston69: Hi @Timo, 
Thanks for the very good review, the helpful hints and especially your close look at the logic! I have never met a developer who did this so well.
The git issues were almost certainly me screwing up (again). I'm continuing to make a genuine effort and ask for your utmost indulgence - I'm quite annoyed with myself.
As far as the testing discussion is concerned, I would like to take up the cudgels for extending unit tests and also testing encapsulated, very simple methods whenever we touch code. The functional tests (e.g. API tests) have cost us an incredible amount of time to fix. When functional tests fail, it is almost impossible to know where exactly the problem is without knowing the full context. As projects become more complex and the number of developers involved increases, this becomes a bigger and bigger problem. I have made the painful experience that a larger test coverage (only) of functional tests created far more problems than it solved. 
Should we test (in Python, after all, only defined by usage) private methods? Actually not useful. However, where we can reduce complexity by outsourcing logic to our own methods, we should do so anyway - my opinion. The extra effort will benefit us in two ways: we will be able to test much more granularly and see more and more where a problem occurs, we will reduce the complexity of methods and generate even more readable code (production as well as test code). Where new code is written test driven, it will no longer be necessary/meaningful to test private methods.
Furthermore, our attention should be focused on reducing test execution times. The current waiting times cost us a lot of productivity and will slow down the acceptance of TDD. In my view, this includes keeping tests with database access to a minimum and using a memory DB as the test database. 
As @Peter noted, this could become a discussion :-).

integreat_cms/summ_ai_api/summ_ai_api_client.py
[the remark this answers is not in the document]
PeterNerlich: Ah, I see! That idea still eluded me, thanks!

PeterNerlich:
Except for [one Sphinx-related TODO](https://github.com/digitalfabrik/integreat-cms/pull/1980/files#diff-c014aeeee747170bda5cb4823f67f5c355754ba552097871badb8e2fd590bdd9R403-R410) this is now ready for review!

timoludwig (commented):
Thanks a lot for your changes! :muscle: 

This is significantly better now, I can really see how much effort went into getting this to work! :rocket:

Only requesting `n` fields at a time and waiting when the request limit is hit already works as expected, thanks! However, the re-scheduled tasks are not executed again when appended to the queue, because of this error: `RuntimeError: cannot reuse already awaited coroutine`.
So this needs to be fixed before the functionality can be used as intended.

Following up on our philosophical discussion, this again raises the question about the significance of the current tests. I mean, in my opinion the main purpose of shipping tests with a PR is to make sure the PR really fixes the issue it intents to. If you have 600 lines of tests which are all succeeding, but your PR does not fix the issue, then the tests are somehow testing the wrong things. I can see how much thought you put into getting a good coverage, but I feel like you got lost in the details and created an artificial testing environment which is too far away from the real world use case. Yes, faster test execution times are important for productivity, but even more important is that the tests reflect a realistic scenario. If our CMS only works on the test bench but fails in production, we can also use [volkswagen](https://github.com/auchenberg/volkswagen) to run our tests. :grin:

In general, you convinced me that it makes sense to also test the small (obvious) things, but in my opinion this is an optional improvement which only makes sense when the bigger picture already works. For future PRs, I would suggest the following approach: At first, write integration tests which really proof your PR is working end-to-end (simulate a user request in the CMS and check whether its outcome is correct) and only when this part is working, go the extra mile and test all the helper functions etc...

integreat_cms/summ_ai_api/summ_ai_api_client.py
```diff
@@ -122,27 +128,36 @@ async def translate_text_fields(self, loop, text_fields):
...
         :rtype: list [ ~integreat_cms.summ_ai_api.utils.TextField ]
         """
+
         # Set a custom SUMM.AI timeout
         timeout = aiohttp.ClientTimeout(total=60 * settings.SUMM_AI_TIMEOUT)
         async with aiohttp.ClientSession(timeout=timeout) as session:
             # Create tasks for each text field
             tasks = [
-                loop.create_task(self.translate_text_field(session, text_field))
+                lambda: self.translate_text_field(session, text_field)
```
timoludwig, line 137:
```suggestion
                self.translate_text_field(session, text_field)
```
There is no such thing like an async lambda function, adding support for them was rejected by the Python core developers: https://bugs.python.org/issue33447

integreat_cms/summ_ai_api/utils.py
```diff
@@ -392,3 +412,125 @@ def __repr__(self):
...
+        :rtype: asyncio.Future
+        """
+        now = time.time()
+        if (
+            self.last_rate_limit is not None
+            and now - self.last_rate_limit < self.wait_time
+        ):
+            # If we are currently waiting out a rate limit,
+            # the worker should sleep for the remaining time before getting the next real task
+            return lambda: asyncio.sleep(self.wait_time - (now - self.last_rate_limit))
```
timoludwig, line 474:
```suggestion
            return asyncio.sleep(self.wait_time - (now - self.last_rate_limit))
```

integreat_cms/summ_ai_api/utils.py
```diff
@@ -392,3 +412,125 @@ def __repr__(self):
...
+    as well as intermittent wait times through the task generator.
+
+    :returns: A list of task-results
+    :rtype: List
+    """
+    completed = []
+    while len(task_generator):
+        task = next(task_generator)
+        try:
+            result = await loop.create_task(task())
```
timoludwig, line 511:
```suggestion
            result = await loop.create_task(task)
```

timoludwig (commented):
With commit 962527b8c1fb243e0789d42bce745166ad40792a, you introduced a new bug by using synchronous lambda functions in an asynchronous context. This is also the reason for the somewhat misleading error messages of the failing `test_auto_translate_easy_german` test.

You don't need an extra wrapper around the functions because calling an async function itself does not yet run it. It just returns a [coroutine](https://docs.python.org/3/library/asyncio-task.html#coroutine) which is only executed once it's scheduled inside an async event loop. So there is no problem in defining all coroutines in the beginning and pass them all to the task queue.

integreat_cms/summ_ai_api/summ_ai_api_client.py
```diff
@@ -122,27 +140,56 @@ async def translate_text_fields(self, loop, text_fields):
...
+                return completed
+
             # Wait for all tasks to finish and collect the results
-            # (the results are sorted in the order the tasks were created)
-            return await asyncio.gather(*tasks)
+            worker_results = await asyncio.gather(
+                *[worker() for i in range(settings.SUMM_AI_MAX_CONCURRENT_REQUESTS)]
+            )
+            # Put all results in one single list
+            all_results = flattened_list(worker_results)
```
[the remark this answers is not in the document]
PeterNerlich: No, not at all.
Before our PR, all requests were done simultaneously and all results were gathered in one big list. With our pattern, there is a limited amount of workers each accumulating the results of a subset of all requests, and each only returning the list of results they achieved. So in the end, we get a list of lists which we want to flatten to get a list of all results.

integreat_cms/summ_ai_api/utils.py
```diff
@@ -392,3 +394,103 @@ def __repr__(self):
...
+        if (
+            self.last_rate_limit is None
+            or time.time() - self.last_rate_limit < self.wait_time
+        ):
+            self.last_rate_limit = time.time()
+        # Reschedule the failed task
+        self.append(task)
+
+
+def flattened_list(l):
```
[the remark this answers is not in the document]
timoludwig: Ok, makes sense, never thought of that differentiation. I have the feeling that somehow even Python doesn't stay consistent with this pattern (e.g. [enumerate()](https://docs.python.org/3/library/functions.html#enumerate) is not called `enumerated()`), but your choice sounds reasonable to me.

integreat_cms/summ_ai_api/summ_ai_api_client.py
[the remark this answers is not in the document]
timoludwig: Ah okay, yeah I get it. But without the results from the sleeping tasks, you don't have to check `if isinstance(x, list)` and can also use the inbuilt [`itertools.chain()`](https://docs.python.org/3/library/itertools.html#itertools.chain), right?

PeterNerlich:
> With commit [962527b](https://github.com/digitalfabrik/integreat-cms/commit/962527b8c1fb243e0789d42bce745166ad40792a), you introduced a new bug by using synchronous lambda functions in an asynchronous context. This is also the reason for the somewhat misleading error messages of the failing `test_auto_translate_easy_german` test.
> 
> You don't need an extra wrapper around the functions because calling an async function itself does not yet run it. It just returns a [coroutine](https://docs.python.org/3/library/asyncio-task.html#coroutine) which is only executed once it's scheduled inside an async event loop. So there is no problem in defining all coroutines in the beginning and pass them all to the task queue.

True, but this is the source of the bug you mentioned previously, which we had missed, that tasks failed to be executed again after running into the rate limit. We had simply put the coroutines back in the queue, but it had already been awaited before and the internal state didn't match. From what we understand, the only way to redo a task is to generate a new coroutine object, by calling the original function with its parameters. Because those are unknown to the worker, we tried just wrapping it in a lambda function, unaware that it wouldn't work because we failed to adjust the tests we had written.
With a normal (synchronous) function it seems to work (coming up in a mammoth commit with lots of little changes which I'm finishing right now).

integreat_cms/summ_ai_api/utils.py
```diff
@@ -392,3 +416,169 @@ def __repr__(self):
...
+    result = []
+    for x in l:
+        if isinstance(x, list):
+            result.extend(flattened_list(x))
+        else:
+            result.append(x)
+    return result
+
+
+def delay_function_call(func, args=None):
```
timoludwig, line 561: Just for the future: When you're working with a dynamic number of arguments (which can also be no arguments at all), it's often time easier to work with the asterisk notation `*args` in the function signature (see [Arbitrary Argument Lists](https://docs.python.org/3/tutorial/controlflow.html#arbitrary-argument-lists)).

timoludwig (commented):
> True, but this is the source of the bug you mentioned previously, which we had missed, that tasks failed to be executed again after running into the rate limit. We had simply put the coroutines back in the queue, but it had already been awaited before and the internal state didn't match. From what we understand, the only way to redo a task is to generate a new coroutine object, by calling the original function with its parameters. Because those are unknown to the worker, we tried just wrapping it in a lambda function, unaware that it wouldn't work because we failed to adjust the tests we had written.
With a normal (synchronous) function it seems to work (coming up in a mammoth commit with lots of little changes which I'm finishing right now).

Ahh okay, thanks for figuring that out! :ok_hand: 

A different idea I had in mind was to put only the text fields in the queue, pass the `session` object to the worker and create the coroutines there. But your solution with delaying the function execution also looks good! :+1:

integreat_cms/summ_ai_api/summ_ai_api_client.py
```diff
@@ -173,3 +204,52 @@ def translate_queryset(self, queryset):
         # Commit changes to the database
         for translation_helper in translation_helpers:
             translation_helper.commit(easy_german)
+
+    @staticmethod
+    def validate_response(response_data, response_status):
```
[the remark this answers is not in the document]
PeterNerlich: Ah right. I was postponing looking that up and forgot.

integreat_cms/summ_ai_api/utils.py
```diff
@@ -392,3 +416,169 @@ def __repr__(self):
...
+        args = []
+
+    def delayed():
+        """
+        :return: A function taking no arguments
+        :rtype: Callable
+        """
+        return func(*args)
+
+    return delayed
```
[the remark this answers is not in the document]
PeterNerlich: Well, would you look at that. Thanks!

integreat_cms/summ_ai_api/summ_ai_api_client.py
```diff
@@ -122,27 +136,44 @@ async def translate_text_fields(self, loop, text_fields):
...
             ]
+            # A "patient" task queue which only hands out sleep tasks after a task was reported as failed
+            task_generator = PatientTaskQueue(tasks)
+
             # Wait for all tasks to finish and collect the results
-            # (the results are sorted in the order the tasks were created)
-            return await asyncio.gather(*tasks)
+            worker_results = await asyncio.gather(
+                *[
+                    worker(loop, task_generator, str(i))
```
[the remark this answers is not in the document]
PeterNerlich: Yeah, but it won't be any significant overhead to keep in case we want to instantiate different groups of workers and distinguish their log entries down the road.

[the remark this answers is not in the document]
timoludwig: In logging, the `%s` formatter already converts ints to strings, so you wouldn't have to change the logging calls.

integreat_cms/summ_ai_api/utils.py
```diff
@@ -392,3 +394,103 @@ def __repr__(self):
...
+        :rtype: asyncio.Future
+        """
+        now = time.time()
+        if (
+            self.last_rate_limit is not None
+            and now - self.last_rate_limit < self.wait_time
+        ):
+            # If we are currently waiting out a rate limit,
+            # the worker should sleep for the remaining time before getting the next real task
+            return asyncio.sleep(self.wait_time - (now - self.last_rate_limit))
```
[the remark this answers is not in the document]
PeterNerlich: Since we're not sure of the larger scale implications of this and decide we are not going to spend time on testing alternative solutions, we will not handle this for now.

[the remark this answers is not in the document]
timoludwig: Alright, fine for me

integreat_cms/summ_ai_api/summ_ai_api_client.py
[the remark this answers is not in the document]
PeterNerlich: That's a separate issue. The result of the sleeping task is `None` and an element of the list returned by the worker among the other text fields. So we can use `itertools.chain()`, and now that I'm less confused about it I'm also all for it, but the resulting (pseudo) list will be the same.

integreat_cms/summ_ai_api/utils.py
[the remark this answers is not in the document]
PeterNerlich: Thought about this a bit more after all. Since the returned text fields are never used (as they are the same objects that were initially given to the `PatientTaskQueue`) it doesn't really matter when there are a few extra `None` among them, and if they are going to be, they will be easy to filter out – but yes, it does feel hacky. I'm flinching away from letting the whole thread sleep for this long of a time span (without having a concrete understanding of the consequences) though, and experimented with making `PatientTaskQueue` an asynchronous iterator as well.

integreat_cms/summ_ai_api/utils.py
```diff
@@ -393,3 +418,137 @@ def __repr__(self):
...
+            and now - self.last_rate_limit < self.wait_time
+        ):
+            # If we are currently waiting out a rate limit,
+            # the worker should sleep for the remaining time before getting the next real task.
+            # Return a function without arguments that the worker can evaluate
+            # to obtain a coroutine.
+            # (see the corresponding comment in SummAiApiClient…translate_text_fields())
+            return partial(asyncio.sleep, self.wait_time - (now - self.last_rate_limit))
+        return self.popleft()
+
```
timoludwig, line 479:
```suggestion
```
Since we're not using the queue as sync iterator anymore, I think we should remove this code block.

tests/summ_ai_api/summ_ai_test.py
```diff
@@ -224,3 +477,147 @@ async def test_summ_ai_error_handling(
...
+    task_generator = PatientTaskQueue(tasks, 1)
+
+    # we check that list comes reverse (fifo-queue)
+    for i in range(len(task_generator)):
+        assert next(task_generator) == tasks[i]
+
+    assert len(task_generator) == 0
+    # we check that next() did remove elements from the queue
+    with pytest.raises(StopIteration):
+        next(task_generator)
```
timoludwig, line 564:
```suggestion
async def test_patient_task_queue_normal_deque():
    """
    Test for PatientTaskQueue Class. Tests that a first-in/first out (fifo) deque-queue is build from a task-list
    Checks that __anext__() removes one element each from the queue
    """

    tasks = [1, 2, 3]
    task_generator = PatientTaskQueue(tasks, 1)

    # we check that list comes reverse (fifo-queue)
    i = 0
    async for task in task_generator:
        assert task == tasks[i]
        i += 1

    assert len(task_generator) == 0
    # we check that __anext__() did remove elements from the queue
    with pytest.raises(StopAsyncIteration):
        await task_generator.__anext__()
```
If we're using `PatientTaskQueue` as asynchronous iterator, we should also test the async iter features instead of the sync ones.

tests/summ_ai_api/summ_ai_test.py
```diff
@@ -224,3 +477,147 @@ async def test_summ_ai_error_handling(
...
+    # Cleanup coroutine without awaiting
+    task.close()
+
+    await asyncio.sleep(0.3)
+
+    # Next task is normal again
+    task = next(task_generator)
+    assert (
+        task in tasks
+    ), "when wait-mode passed, next() is expected to return the next task"
```
timoludwig, line 623:
```suggestion
    # Test first task
    task = await task_generator.__anext__()
    assert task == tasks[0], "PatientTaskQueue should return the first element"

    # Test second task with timing
    start = time.time()
    task = await task_generator.__anext__()
    end = time.time()
    assert task == tasks[1], "PatientTaskQueue should return the second element"
    assert (
        end - start < 0.5
    ), "PatientTaskQueue should return the second element very fast"

    # Hit the rate limit and check whether the second task is rescheduled
    task_generator.hit_rate_limit(task)
    start = time.time()
    task = await task_generator.__anext__()
    end = time.time()
    assert task == tasks[1], "PatientTaskQueue should return the second element again"
    assert end - start > 0.5, "PatientTaskQueue should return the second element slower"

    # Test third task
    start = time.time()
    task = await task_generator.__anext__()
    end = time.time()
    assert task == tasks[2], "PatientTaskQueue should return the third element"
    assert (
        end - start < 0.5
    ), "PatientTaskQueue should return the third element very fast"

    assert (
        len(task_generator) == 0
    ), "PatientTaskQueue should be empty after three tasks have been removed"
```
Here as well, we should test the async queue instead of the sync one.

tests/summ_ai_api/summ_ai_test.py
```diff
@@ -1,11 +1,28 @@
+import asyncio
```
timoludwig, line 1:
```suggestion
import asyncio
import time
```

tests/summ_ai_api/utils.py
```diff
@@ -0,0 +1,94 @@
...
+        :param request: The request
+        :type request: aiohttp.web.Request
+
+        :return: The response
+        :rtype: aiohttp.web.Response
+        """
+        return aiohttp.web.json_response(data=response_data, status=status)
+
+    return fake_summ_ai_server
+
```
timoludwig, line 34:
```suggestion


def make_mock_summ_ai_server_rate_limited():
    """
    Creates a mock of SUMM.AI server with rate limiting

    :return: Mocked SUMM.AI server with rate limiting
    :rtype: aiohttp.web.Application
    """
    app = aiohttp.web.Application()
    app["attempt"] = 0

    async def fake_summ_ai_server_rate_limited(request):
        request.app["attempt"] += 1
        if request.app["attempt"] < 3:
            return aiohttp.web.json_response(
                data={"error": "rate limit exceeded"}, status=429
            )
        return aiohttp.web.json_response(
            data={
                "translated_text": "Hier ist Ihre Leichte Sprache Übersetzung",
                "jobid": "9999",
            },
            status=200,
        )

    app.router.add_post("/", fake_summ_ai_server_rate_limited)
    return app


```

Since the queue is still not working as expected despite all tests succeeding, I fear we can not avoid adding an integration test for this any longer. I suggest e.g. this fake server which combines the ideas of `test_worker()` and `make_mock_summ_ai_server()`.

integreat_cms/summ_ai_api/utils.py
```diff
@@ -3,6 +3,10 @@
 """
 import logging
 import itertools
+import time
+import asyncio
+from collections import deque
+from functools import partial
```
timoludwig, line 9:
```suggestion
```

tests/summ_ai_api/summ_ai_test.py
```diff
@@ -100,7 +353,14 @@ async def test_auto_translate_easy_german(
...
     # Setup a dummy server to fake responses from the SUMM.AI API server (an instance of aiohttp.test_utils.RawTestServer)
-    fake_server = await aiohttp_raw_server(fake_summ_ai_server)
+    fake_server = await aiohttp_raw_server(
+        make_mock_summ_ai_server(
+            response_data={
+                "translated_text": "Hier ist Ihre Leichte Sprache Übersetzung",
+                "jobid": "9999",
+            }
+        )
+    )
```
timoludwig, line 363:
```suggestion
    # Setup a dummy server to fake responses from the SUMM.AI API server
    # (an instance of aiohttp.test_utils.TestServer)
    fake_server = await aiohttp_server(make_mock_summ_ai_server_rate_limited())
    # Avoid unnecessary long wait times during tests
    settings.SUMM_AI_RATE_LIMIT_COOLDOWN = 0.1
```
In the real integration test I'd suggest to use the rate limited fake server to really test this feature.
This requires
```
from .utils import make_mock_summ_ai_server_rate_limited
```

and passing the fixture `aiohttp_server` instead of `aiohttp_raw_server` to `test_auto_translate_easy_german`.

timoludwig (commented):
Sorry it took me so long to get back to this! :innocent: 

Thanks a lot for your improvements, I get the feeling that this PR is finally approaching its merge :grinning: 

Three main points I'd like to address:
1. The most pressing issue are race conditions in your queuing code. When you first check the length of the queue and then request an element from it afterwards, it might no longer exist. This pattern goes against Python's [EAFP](https://docs.python.org/3/glossary.html?highlight=eafp#term-EAFP) principle - Easier to ask for forgiveness than permission.
2. Closely connected to 1.: I won't repeat all my arguments from our previous discussion, but I think the fact that the current code still doesn't work proves my point that we finally need to add a real integration test. We can always try to get as much coverage as possible with fine-graded sub-tests of specific components, but we will probably always forget some edge case or specific combination of steps. In the end, I want to know whether everything works when a user clicks the translate button. In order to achieve this, I added a suggestion for a mocked server which only returns the correct result when the request is tried again after an initial rate limiting response.
3. I like the async variation of the queue and I think this is much more Pythonic than the previous solution which handed out waiting tasks. I'd just suggest to remove all sync iterator code since it's just used in the tests and we will probably not need it in the future. Also, I added suggestions how we could test the async functionality of the queue specifically.

integreat_cms/summ_ai_api/utils.py
PeterNerlich: My question with this is: What is your take on the issue of inheriting from `deque`? I like the idea of having a subclass that behaves like a special form of the base class, which is why I opted to add the async version of `__next__()` in addition to this. I understand why removing `__next__()` instead is a good thing (it's dead code for the current state of the software after all) – but it also breaks the idea that the class does it's special thing but can also be used just as one would use the base class (which in this case doesn't technically break anything but our overarching intent with the class).
It's another vague thing where I have nothing more than two conflicting gut feelings, which don't play nicely with a habit of wanting to do things perfectly in first approximation... :grin:

tests/summ_ai_api/utils.py
```diff
@@ -0,0 +1,94 @@
...
+    Creates a mock of SummAiServer-(Response)
+
+    :param response_data: data that mocked server returns
+    :type response_data: dict
+
+    :param status: reponse_data that mocked server returns (default = 200)
+    :type status: int
+
+    :return: mocked SummAiServer
+    :rtype: aiohttp.web.Response
```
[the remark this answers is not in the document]
PeterNerlich: Good catch! Has been a while since I worked on this code by now, but I think we got quite a bit lost in return types and such

integreat_cms/summ_ai_api/utils.py
timoludwig: The way I understand it: We're mixing concepts here, since queues are different from iterators, and we need some functionality from both. The `next()` is something defined on iterators, not on queues. So by removing the `__next__` methods, we're not breaking the concept of the base class, it's still a normal queue and you can pop tasks synchronously by using the inbuilt methods `.popleft()` etc... just when you want to use the "patient" feature of the class, then you can use it's async iterator interface...
Theoretically, we could also use [async queues](https://docs.python.org/3/library/asyncio-queue.html), but they are not thread-safe. So we need `dequeue` to get thread-safety, and we need the async iterator stuff to get the queue items asynchronously.
I mean if we really wanted to separate these concepts, we could also define two classes: One queue which contains the tasks, and one iterator which handles the fetching from the queue. But I guess it's also fine in its current form.

MizukiTemma (changes requested):
Thank you so much for all the work you did for this PR 💪 

I successfully reproduced the reported bug in the develop branch and tested this PR with the same data: it works as expected wiith the bulk action 😃

But somehow no more easy German translation can be created/updated via check box in the section `Implications on other translations` in the form. This is the case even the original text is very short and has no bullet list or paragrahs.

PeterNerlich:
Thanks for the review @MizukiTemma!

> But somehow no more easy German translation can be created/updated via check box in the section `Implications on other translations` in the form.

Shoot. What exactly would be the expected behaviour in this case? I have no understanding of translations beyond the superficial and what we learned while working on this PR.

timoludwig:
> > But somehow no more easy German translation can be created/updated via check box in the section `Implications on other translations` in the form.
> 
> Shoot. What exactly would be the expected behaviour in this case? I have no understanding of translations beyond the superficial and what we learned while working on this PR.

@PeterNerlich That's not exactly your fault, the MT translation feature via the content forms is relatively new, so it did not yet exist when you worked on this issue.
But yes, it would be cool if you could check out the new structure and rebase the branch to resolve the conflicts. I can also assist with this since the breaking changes are mostly my fault :innocent:

PeterNerlich:
We rebased the changes on develop, please check if that issue persists @MizukiTemma. If not, please continue to recheck whether this PR does what it should, as we are not sure how it interacts with the rest of the system. The existing tests all seem to pass, but who knows.

MizukiTemma (approved):
@PeterNerlich Thank you for update. I tested it again and it works as expected 😃

seluianova (commented):
I tested the feature a bit, it works basically fine.
There is one concern I got:
CMS waits for SUMM_AI_RATE_LIMIT_COOLDOWN, when it gets error 429/529 only during the first iteration.
But if the error occurs again, it doesn't wait anymore, but re-sends the requests constantly.

Steps to reproduce:
1. Setup SUMM AI mock server to return error 429
2. Translate any page (let's say with 5 paragraphs)

Result:
CMS sends 5 requests --> gets error 429 --> waits 30 seconds.
After 30 seconds, it continues to send requests continuously, despite the error 429.

Do we want to fix it?

**UPD**: I also found one critical side effect.
If the response error not in (500, 429, 529), CMS keeps constantly re-sending requests.
Can be reproduced, for example, with the response codes 400, 404, etc.

integreat_cms/summ_ai_api/summ_ai_api_client.py
```diff
@@ -118,22 +135,38 @@ async def translate_text_fields(self, loop, text_fields):
...
```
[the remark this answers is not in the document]
PeterNerlich: Since workers only try to continually obtain and complete one task from the queue and return when it is empty, it barely has any effect. I feel like making the line longer is not worth it, the overall area is already confusing to understand...

tests/summ_ai_api/summ_ai_test.py
```diff
@@ -228,3 +452,180 @@ async def test_summ_ai_error_handling(
...
+        "translated_text": "Das ist einfaches Deutsch.",
+    }
+    assert (
+        SummAiApiClient(MockedRequest(), PageTranslationForm).validate_response(
+            response_data_translation_succeeded, 200
+        )
+        is True
+    ), "if translated_text found in reponse-data, validate_response should return True"
+
+    response_data_translation_failed = {
```
seluianova, line 473: I think it could be 2 separate cases, because there are basically 2 cases: one positive and one negative.
The problem with combining them is that if the first assertion fails, the second one won't be reached.

tests/summ_ai_api/summ_ai_test.py
```diff
@@ -188,25 +402,35 @@ async def test_summ_ai_error_handling(
     # The region we want to use for testing
     region_slug = "augsburg"
     # Setup a dummy server to fake responses from the SUMM.AI API server (an instance of aiohttp.test_utils.RawTestServer)
-    fake_server = await aiohttp_raw_server(broken_fake_summ_ai_server)
```
seluianova, line 191: Function broken_fake_summ_ai_server can be removed, if it's not used anymore

tests/summ_ai_api/utils.py
```diff
@@ -0,0 +1,164 @@
...
+                "jobid": "9999",
+            },
+            status=200,
+        )
+
+    app.router.add_post("/", fake_summ_ai_server_rate_limited)
+    return app
+
+
+def make_broken_json_summ_ai_server(response_text, status=200):
```
seluianova, line 73: This method is just the same as make_mock_summ_ai_server().
I guess we could use make_mock_summ_ai_server() as a universal one.

PeterNerlich: It is not, this one allows us to inject any *text* into the request, while the other is more convenient as it takes a `dict()` and turns that into JSON. We need this separate instance because we test the behaviour on bogus data that cannot be parsed as JSON. However, just looking at this collection of functions in `utils.py`, this context is lost, the difference between the two functions is not obvious at all at first (or second) glance and I worry this might be a big source of confusion in the future. Do you think adding additional doctrings would be sufficient?

PeterNerlich:
> CMS sends 5 requests --> gets error 429 --> waits 30 seconds. After 30 seconds, it continues to send requests continuously, despite the error 429.
> 
> Do we want to fix it?

I think I fixed it, please retry :pray:  @seluianova

integreat_cms/summ_ai_api/utils.py
```diff
@@ -388,3 +418,139 @@ def __repr__(self):
...
+            self.last_rate_limit = time.time()
+            logger.debug(
+                "PatientTaskQueue.hit_rate_limit() [%s +1] (blocking for %s s)",
+                len(self),
+                self.wait_time,
+            )
+        else:
+            logger.debug(
+                "PatientTaskQueue.hit_rate_limit() [%s +1] (already known, %s/%s s left)",
+                len(self),
```
timoludwig, line 517:
```suggestion
                "PatientTaskQueue hit rate limit during %r (already known, %s/%ss elapsed)",
                task,
```

tests/summ_ai_api/summ_ai_test.py
```diff
@@ -228,3 +437,186 @@ async def test_summ_ai_error_handling(
...
+    Tests if SummAiInvalidJSONError is raised when an invalid JSON is sent as response.
+    """
+    session = aiohttp.ClientSession()
+    fake_server = await aiohttp_raw_server(
+        make_broken_json_summ_ai_server(
+            response_text="""{
+                "translated_text": "a text in an unfinished json response",
+                "jobid": "9999",
+            """,
+            status=200,
```
timoludwig, line 486:
```suggestion
            response_text="<h1>Internal Server Error</h1>",
            status=500,
```
A more realistic example wold be an html-response...

tests/summ_ai_api/summ_ai_test.py
```diff
@@ -228,3 +437,186 @@ async def test_summ_ai_error_handling(
...
+
+    assert (
+        SummAiApiClient(MockedRequest(), PageTranslationForm).check_rate_limit_exceeded(
+            500
+        )
+        is False
+    ), "if response-status not in (429, 529), check-rate-limit should return False"
+
+
+# pylint: disable=compare-to-zero
```
timoludwig, line 549:
```suggestion
```

tests/summ_ai_api/summ_ai_test.py
```diff
@@ -228,3 +437,186 @@ async def test_summ_ai_error_handling(
...
+    # we check that list comes reverse (fifo-queue)
+    i = 0
+    async for task in task_generator:
+        assert task == tasks[i]
+        i += 1
+
+    assert len(task_generator) == 0
+    # we check that __anext__() did remove elements from the queue
+    with pytest.raises(StopAsyncIteration):
+        # await task_generator.__anext__()
```
timoludwig, line 568:
```suggestion
```

timoludwig (commented):
Awesome, thanks a lot for all your improvements! :+1: :rocket: 

> > If the response error not in (500, 429, 529), CMS keeps constantly re-sending requests too.
> 
> This is the intended behaviour for now. Even 500 requests just throw a `SummAiRuntimeError` which simply gets caught and logged, without triggering a wait time. I guess if we observe other cases that happen often, we will add behaviour here.

Hmm, I guess this is not quite ideal, because if the server has a longer down time, we make it harder for it to recover if we flood it with additional requests potentially indefinitely.
Sorry, I guess I did not think about this problem when suggesting to resend these requests, maybe we need some kind of "max-retries" setting to avoid an infinite loop?

integreat_cms/summ_ai_api/utils.py
```diff
@@ -388,3 +418,139 @@ def __repr__(self):
...
+        :returns: a task of the queue
+        :rtype: asyncio.Future
+        """
+        now = time.time()
+        if (
+            self.last_rate_limit is not None
+            and now - self.last_rate_limit < self.wait_time
+        ):
+            # Bail out early when the queue is empty
+            if len(self) == 0:
```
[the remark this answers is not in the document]
PeterNerlich: Yes, but I strongly feel like `len(self) == 0` is more explicit and less confusing in this case

PeterNerlich:
> @timoludwig: [...] if the server has a longer down time, we make it harder for it to recover if we flood it with additional requests potentially indefinitely.

I guess we could keep the fixed wait time on 429 and 529 status codes and implement exponential backoff on everything outside the 200s block + failed JSON parsings.
I'm not sure where exactly to implement the exponential backoff though. `PatientTaskQueue` seems ideal at first because it already implements the normal wait, but then we need workers to call some kind of `PatientTaskQueue.task_was_successful()` method to reset the backoff, instead of only calling `hit_rate_limit()` when it failed – I'm not sure whether this is a problem

seluianova:
> > If the response error not in (500, 429, 529), CMS keeps constantly re-sending requests too.
> 
> This is the intended behaviour for now. Even 500 requests just throw a `SummAiRuntimeError` which simply gets caught and logged, without triggering a wait time. I guess if we observe other cases that happen often, we will add behaviour here.

Right, but I mean 500 stops the execution and doesn't re-send the request (I believe it's okay, we shouldn't try to kill the server even more).
But 400, 404 and others keeps re-sending requests, and I'm not sure if that makes sense.
If we get 429 and 529, we expect the issue to resolve itself soon.
But if we get 400 or 404, it normally means that there is something wrong on our side.

seluianova:
> > If the response error not in (500, 429, 529), CMS keeps constantly re-sending requests too.
> 
> This is the intended behaviour for now. Even 500 requests just throw a `SummAiRuntimeError` which simply gets caught and logged, without triggering a wait time. I guess if we observe other cases that happen often, we will add behaviour here.

**UPD**. Reworked my last answer

So. The correct statement would be:

If the response code is not in (200, 429, 529, 500) AND the response body contains JSON (with the "error" tag), then CMS does not send repeated requests, but stops execution and returns an error message to the user (it looks good to me).

BUT:

If the response code is not in the list above AND the response body type is not JSON, then the execution loops.

I carefully double-checked some error codes from the SUMM AI documentation, and here are the results:

400 - execution stops (requests are not re-sent). 
Previously I tested this status a bit incorrect, because I didn’t include JSON into the response in the mocked server. But as I understand, SUMM AI always include JSON in case of 400, so I believe there is no problem here.

404 - execution loops (requests are constantly re-sent)
Because 404 usually has HTML in the response body.

500 - execution stops, because we handle this error separately and throw an exception.

503 - execution loops, because there is no JSON in the response body.

I think the expected behavior would be:

If we get 429 or 529, we re-send requests (it works now)
If we get other errors, there is no need to re-send requests, because it’s either the server error (and we don’t want to overload it more) or the problem is on our side.

P.S. sorry for the long-read 🙈
I will read the other comments tomorrow.

tests/summ_ai_api/summ_ai_test.py
seluianova: I think there would be 2 separate cases:
1. Unexpected html-response with 200 status code
2. 500 status code

Or even 3:
3. 200 status code, json response without translated_text tag

tests/summ_ai_api/utils.py
seluianova: @PeterNerlich what confused me here: 
This method is named as “broken” but it accepts any text and can basically accept a valid json too.

timoludwig:
@seluianova I see your point, and I'm not sure if the problem persists, but some time ago I tested this feature and realized that error 500 could be viewed as some sort of "extreme variant" of 529, since if we send a lot of requests at the same time, the server crashed in some cases, which led to the "incorrect" error 500 instead of 529 (and html response instead of json). These crashes however were only temporary and resolved itself after a short time, so re-sending after the waiting period could also make sense in this case.
I will contact SUMM.AI and ask whether this problem was resolved in the meantime or whether this could still happen.

seluianova:
@timoludwig okay, but if so, I am concerned anyway, because if server is really down, we will make the user wait for 30 secs * 20 retries

timoludwig:
> @timoludwig okay, but if so, I am concerned anyway, because if server is really down, we will just make the user wait for 30 secs * 20 retries

Yeah you're right, the 20 retries do not make sense for a cumulative waiting time of 30 seconds... Maybe 5 retries make more sense? :sweat_smile:

PeterNerlich:
After a break from this neverending hunk of a PR, I thought I could resume work maybe. So I finally got around to implementing retries.

timobrembeck:
@PeterNerlich Thanks for revisiting this, could you have a look at the 12 unresolved conversations and check whether they still require some changes or whether everything has been resolved?

PeterNerlich:
> Thanks for revisiting this, could you have a look at the 12 unresolved conversations and check whether they still require some changes or whether everything has been resolved?

Sorry, I lost track by now.

For the topic regarding worker identifiers as strings or not, we have differing opinions and I am unable to discern which is the more sensible one. Here we probably require a more detailed talk to progress, but I believe GitHub or other text forms are not a good fit for that. Unfortunately, I won't be at the upcoming conference either.

For the rest, I should be able to adopt them as early as Soon™.

PeterNerlich:
@timobrembeck Now only our stalemate should be left

integreat_cms/summ_ai_api/utils.py
```diff
@@ -392,3 +416,169 @@ def __repr__(self):
...
+        if (
+            self.last_rate_limit is None
+            or time.time() - self.last_rate_limit < self.wait_time
+        ):
+            self.last_rate_limit = time.time()
+        # Reschedule the failed task
+        self.appendleft(task)
+
+
+async def worker(loop, task_generator, i=None):
```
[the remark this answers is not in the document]
timobrembeck: Even if you insist on using string ids, `None` is no string, so I would remove this default value in any case. Also, I would encourage developers to always use this function in combination with an id to be able to associate specific logging messages with their origin.

timobrembeck (commented):
@PeterNerlich Thanks a lot for your latest changes, they look good! :rocket: 

> Now only our stalemate should be left

Well, I still don't see any advantage of your approach and in my opinion, it's a bit counter-intuitive from a conceptual point of view, but it's also pretty irrelevant, so I won't stop this PR from being merged this way :laughing: 

So the only thing that's left to do now is squashing the commits, and then this PR is fiiiiiiinally good to go :joy: :innocent:

timobrembeck:
@PeterNerlich could you append the "add release note" commit to the commit "SUMM.AI: gracefully handle rate limiting" that actually contains this change? Thanks!

seluianova:
I think I see one issue 🙈
If we get 429/529, and the error does not go away after retries, the message "successfully translated" is still displayed.
Although the page is not actually translated, and the "German" content is simply copied as is into "German (easy)".

PeterNerlich:
Okay, radical idea:
This PR has added substantial complexity in order to process a queue of tasks in batches and retry failing ones and continue after waiting out a hold off period.
Further increase complexity by putting this completely into the background, meaning the users page loads with no greater delay than any other page, and give them a way to see the progress of running jobs. Also prevent new jobs from involving tasks already queued by another job.

timobrembeck:
> Okay, radical idea: This PR has added substantial complexity in order to process a queue of tasks in batches and retry failing ones and continue after waiting out a hold off period. Further increase complexity by putting this completely into the background, meaning the users page loads with no greater delay than any other page, and give them a way to see the progress of running jobs. Also prevent new jobs from involving tasks already queued by another job.
> 
> Is it worth it? I have no clue. But it's how I feel long running processes should look like from a users perspective.

This seems like the right approach, yes.

Then we'd have to decide whether we use the simple approach:
https://github.com/digitalfabrik/integreat-cms/blob/37dc98314b23efe92e125d490701be34d52c64a0/integreat_cms/api/decorators.py#L178-L180
or include [celery](https://docs.celeryq.dev/en/stable/) for this (but if we do so, we maybe should also have used celery for the queue tasks in the first place, for our custom queue I think the simple threading library is enough)?

PeterNerlich:
I'm not familiar with challenges and advantages of either approach, but I think the simple one should do the job just fine…? The website of celery states *"It’s a task queue with focus on real-time processing, while also supporting task scheduling."* – both of which are not interesting to this PR.

PeterNerlich:
Only fixed @seluianova s findings for now. I'm debating whether further expanding on it should be a separate PR.

If we don't want to spend a lot of time making the loop a background action and developing the overview necessary to be transparent to the user, I think we might also just throw up a helpful "have patience" kind of popup in JS.

<details><summary>Diff for the commit (which was squashed into the main SUMM AI commit)</summary>

```diff
commit 1edb8afbed793dc0350497d07fae0692a6a6a853
Author: Peter Nerlich <peter.nerlich@tuerantuer.org>
Date:   Sun Oct 15 21:42:10 2023 +0200

    add PatientTaskQueue.abort(), prevent reporting failed translations as success

diff --git a/integreat_cms/summ_ai_api/summ_ai_api_client.py b/integreat_cms/summ_ai_api/summ_ai_api_client.py
index 9b525dc2e..7fe5b2bb5 100644
--- a/integreat_cms/summ_ai_api/summ_ai_api_client.py
+++ b/integreat_cms/summ_ai_api/summ_ai_api_client.py
@@ -84,7 +84,7 @@ class SummAiApiClient(MachineTranslationApiClient):
             # This is normally filtered out before this function is called,
             # something must have gone wrong.
             # Raise an exception without immediately catching it!
-            raise SummAiRuntimeError("text_field to translate is none or text is empty")
+            raise SummAiRuntimeError("Field to translate is None or empty")
         try:
             async with session.post(
                 settings.SUMM_AI_API_URL,
@@ -150,8 +150,17 @@ class SummAiApiClient(MachineTranslationApiClient):
                 partial(self.translate_text_field, session, text_field)
                 for text_field in text_fields
             ]
+
+            # If the translation is aborted, set the exception field
+            # to both signal that this wasn't translated and to display a reason why
+            def abort_function(task, reason):
+                # Retrieve field from arguments to translate_text_field()
+                field = task.args[1]
+                # Set the exception
+                field.exception = f"Machine translation aborted: {reason}"
+
             # A "patient" task queue which only hands out sleep tasks after a task was reported as failed
-            task_generator = PatientTaskQueue(tasks)
+            task_generator = PatientTaskQueue(tasks, abort_function=abort_function)
 
             # Wait for all tasks to finish and collect the results
             worker_results = await asyncio.gather(
diff --git a/integreat_cms/summ_ai_api/utils.py b/integreat_cms/summ_ai_api/utils.py
index 206d8e2e1..27c71de7b 100644
--- a/integreat_cms/summ_ai_api/utils.py
+++ b/integreat_cms/summ_ai_api/utils.py
@@ -452,6 +452,7 @@ class PatientTaskQueue(deque):
         tasks,
         wait_time=settings.SUMM_AI_RATE_LIMIT_COOLDOWN,
         max_retries=settings.SUMM_AI_MAX_RETRIES,
+        abort_function=None,
     ):
         """
         Constructor initializes the class variables
@@ -464,11 +465,19 @@ class PatientTaskQueue(deque):
 
         :param max_retries: Maximum retries before giving up
         :type max_retries: int
+
+        :param abort_function: Function to call for each unfinished task if the queue is aborted.
+                               Takes two arguments: The task (:class:`asyncio.Future`) and the reason given (:class:`str`).
+                               Can be `None` instead to do nothing.
+        :type abort_function: asyncio.Future
         """
         super().__init__(tasks)
         self.wait_time = wait_time
         self.max_retries = max_retries
         self.retries = 0
+        self.abort_function = abort_function
+        self._aborted = False
+        self._in_progress = []
 
     def __aiter__(self):
         return self
@@ -481,7 +490,7 @@ class PatientTaskQueue(deque):
         :returns: a task of the queue
         :rtype: asyncio.Future
         """
-        if self.retries > self.max_retries:
+        if self._aborted:
             raise StopAsyncIteration
 
         now = time.time()
@@ -501,7 +510,9 @@ class PatientTaskQueue(deque):
             )
             await asyncio.sleep(wait_time_remaining)
         try:
-            return self.popleft()
+            task = self.popleft()
+            self._in_progress.append(task)
+            return task
         except IndexError as e:
             raise StopAsyncIteration from e
 
@@ -512,6 +523,9 @@ class PatientTaskQueue(deque):
         :param task: The task that failed because of the rate limiting
         :type task: asyncio.Future
         """
+        assert (
+            task in self._in_progress
+        ), f"PatientTaskQueue: Task not known as in progress: {task}"
 
         # Only save current timestamp if this is the first failed request reported
         if (
@@ -534,10 +548,16 @@ class PatientTaskQueue(deque):
                 time.time() - self.last_rate_limit,
                 self.wait_time,
             )
+
         # Reschedule the failed task
+        self._in_progress.remove(task)
         self.appendleft(task)
 
-    # pylint: disable=unused-argument
+        if self.retries > self.max_retries and not self._aborted:
+            self.abort(
+                f"Retried tasks a consecutive {self.max_retries} times. Giving up."
+            )
+
     def completed(self, task):
         """
         A task was completed, reset the retry counter
@@ -546,6 +566,20 @@ class PatientTaskQueue(deque):
         :type task: asyncio.Future
         """
         self.retries = 0
+        self._in_progress.remove(task)
+
+    def abort(self, reason="Aborted"):
+        """
+        Abort the Queue, handling unfinished task according to the supplied abort function.
+
+        :param reason: The reason why the queue was aborted that is to be handed to the supplied abort function.
+        :type reason: str
+        """
+        self._aborted = True
+        logger.debug("PatientTaskQueue aborted: %s", reason)
+        if self.abort_function:
+            for unfinished_task in list(self) + self._in_progress:
+                self.abort_function(unfinished_task, reason)
 
 
 async def worker(loop, task_generator, i):
 ```
 
 </details>

File: integreat_cms/summ_ai_api/summ_ai_api_client.py [hunk 34]
```diff
@@ -80,20 +98,19 @@ async def translate_text_field(self, session, text_field):
...
-                # Let the field handle the translated text
-                text_field.translate(response_data["translated_text"])
-                return text_field
-        except (aiohttp.ClientError, asyncio.TimeoutError, SummAiException) as e:
+                try:
+                    response_data = await response.json()
+                except aiohttp.ContentTypeError as e:
+                    raise SummAiInvalidJSONError(
+                        f"API delivered invalid JSON: {response.status} - {await response.text()}"
+                    ) from e
```
seluianova, line 106:
```suggestion
                    logger.error(
                        "Failed to parse JSON from API response: %s",
                        response.status,
                    )
                    raise SummAiInvalidJSONError from e
```

integreat_cms/summ_ai_api/summ_ai_api_client.py
```diff
@@ -172,3 +215,72 @@ def translate_queryset(self, queryset, language_slug):
...
+        :type response_data: dict
+
+        :param response_status: The response-status form SummAiApi-Request
+        :type response_status: int
+
+        :returns: True or False
+        :rtype: bool
+
+        :raises SummAiRuntimeError: One of the parameters is not an integer.
+        :raises SummAiInvalidJSONError: Invalid JSON response returned by the API.
```
seluianova, line 234: This exception is not actually returned by this function

```suggestion
```

integreat_cms/summ_ai_api/summ_ai_api_client.py
```diff
@@ -172,3 +215,72 @@ def translate_queryset(self, queryset, language_slug):
...
+        :param response_data: The response-data from SummAiApi
+        :type response_data: dict
+
+        :param response_status: The response-status form SummAiApi-Request
+        :type response_status: int
+
+        :returns: True or False
+        :rtype: bool
+
+        :raises SummAiRuntimeError: One of the parameters is not an integer.
```
seluianova, line 233: Not sure if you really meant "not an integer" :)

```suggestion
        :raises SummAiRuntimeError: Invalid JSON response returned by the API
```

integreat_cms/summ_ai_api/summ_ai_api_client.py
seluianova: Or perhaps you really wanted to raise it instead of SummAiRuntimeError?

integreat_cms/summ_ai_api/summ_ai_api_client.py
seluianova, line 106: I suggest adding this logging because the message of this exception itself is not logged.
If execution falls into this "except" block, there is not clue in the log at all to what the problem was.

Use case example:
If API returns 404 - in this case, the response body is usually html --> json can't be parsed --> but you'd never guess it from the log without explicit logging

```suggestion
                    logger.error(
                        "Failed to parse JSON from API response: %s",
                        response.status,
                    )
                    raise SummAiInvalidJSONError from e
```

integreat_cms/summ_ai_api/summ_ai_api_client.py
PeterNerlich: Trying to recall what happened, seems like I forgot this when raising SummAiInvalidJSONError was moved to outside of and before calling this function in `translate_text_field()` (line 104 currently)

integreat_cms/summ_ai_api/summ_ai_api_client.py
PeterNerlich: Do you have a specific intention in leaving out the `await response.text()` from the logging?

seluianova: Yes, but I don't insist here.

I assumed that the status code will give us enough information about the problem, while the response text may not be as useful in cases where it is not JSON. 
_(I decided so by looking at the repeated html 'not found' in the log)_

PeterNerlich: I decided to keep the exception text and the `await response.text()` (but only there), without separating it out into one call to `….text()` and parsing it ourselves, because it would basically be duplicated code from a foreign library that would only be run in the event of an error, which shouldn't be often and thus bring no big performance improvement.

Furthermore, while the resulting string of `….text()` is not being cached, if I interpret the client library correctly, the only substantial wait time in calling it arises from not having received the whole response yet. This means that calling it a second time should not result in any significant delay, since the byte reader can read the whole data sequence in one go.

integreat_cms/summ_ai_api/summ_ai_api_client.py
```diff
@@ -80,20 +98,28 @@ async def translate_text_field(self, session, text_field):
...
-                        raise SummAiException(
-                            f"API error: {response.status} - {response_data['error']}"
-                        )
-                    raise SummAiException(
-                        f"Unexpected API result: {response.status} - {response_data!r}"
+                try:
+                    response_data = await response.json()
+                except (
+                    aiohttp.ContentTypeError,
+                    aiohttp.web_exceptions.HTTPBadRequest,
```
seluianova, line 105: it produces AttributeError at /augsburg/pages/de-si/machine-translate/
`module 'aiohttp' has no attribute 'web'`

seluianova, line 105: It produces AttributeError now at /augsburg/pages/de-si/machine-translate/
`module 'aiohttp' has no attribute 'web_exceptions'`

It requires an import to fix it, but as far as I understand, this exception can't be actually raised here anyway?

PeterNerlich: Aah dang, I forgot the import. I added it because I looked at the implementation, and [it actually seems to be raised if the mime-type doesn't match](https://github.com/aio-libs/aiohttp/blob/54a0d9a3f1fad037c812e75b432054d7c5a9bf45/aiohttp/web_request.py#L674-L681).

seluianova: Your link refers to a different entity.
This is the right one:
https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client_reqrep.py#L1061

PeterNerlich:
I'm sorry for taking so long again. Bringing the branch up to date with develop and thinking through the original intention on both sides of each merge conflict was daunting and I always had something to distract me from it until yesterday.
There will still be some mistakes with the type declarations I bet, but the mypy tool was very useful! At least once I found out it exists from the CI because I didn't look at the types PR or the up to date documentation :grin:

integreat_cms/summ_ai_api/summ_ai_api_client.py
```diff
@@ -175,3 +223,61 @@ def translate_queryset(self, queryset: list[Page], language_slug: str) -> None:
...
+        return True
+
+    @staticmethod
+    def check_internal_server_error(response_status: int) -> bool:
+        """
+        Checks if we got a HTTP 500 error
+
+        :param response_status: The response-status form SummAiApi-Request
+
+        :returns: False (if the response_status is neither 429 nor 529)
```
seluianova, line 258:
```suggestion
        :returns: False (if the response_status is not 500)
```

seluianova (approved):
Soo 🙃
Thanks for tackling a bunch of my comments.
I think it looks good now.

I still have one small doubt, but it may probably be fixed later (or not fixed at all):
We re-send requests not only when we get 429 or 529, but also when the response  type is not json. This happens, for example, when we get 404.
As we discussed above, we consider that as okay.
But:
You can't read this behavior from the code.
So it works like this as a side effect actually.
I think it would be ideal to refactor the code (some day) to make the desired behavior more explicit.

PeterNerlich:
> We re-send requests not only when we get 429 or 529, but also when the response type is not json. This happens, for example, when we get 404. As we discussed above, we consider that as okay. But: You can't read this behavior from the code.

I would argue that you can read it from code, since the implementation of `worker()` in `utils.py` includes:

```python
        except (SummAiRateLimitingExceeded, SummAiInvalidJSONError):
            task_generator.hit_rate_limit(task)
```

But I concede that I, as the original author of this functionality, had to search through the code again to find where this happens. So this *is* far from ideal.
Do you think adding this to the docstring of `worker()` alleviates this situation in any significant way?

> Catches SummAiRateLimitingExceeded and SummAiInvalidJSONError and counts them as rate limit hits in order enqueue them again.

As well as this to the docstring of `SummAiApiClient.translate_text_field()`:

> Note that worker currently not only counts SummAiRateLimitingExceeded but also SummAiInvalidJSONError as a rate limit hit and enqueues the task again.

integreat_cms/core/settings.py
```diff
@@ -959,6 +959,20 @@
 #: The timeout in minutes for requests to the SUMM.AI API
 SUMM_AI_TIMEOUT: Final[int] = 10
 
+#: The limit for "Too many requests".
+SUMM_AI_MAX_CONCURRENT_REQUESTS = int(
+    os.environ.get("INTEGREAT_CMS_SUMM_AI_MAX_CONCURRENT_REQUESTS", 20)
+)
+
+#: Waiting time after "Too many request" response was sent
```
seluianova, line 967:
```suggestion
#: Waiting time after "Too many requests" response was sent
```

seluianova:
@PeterNerlich ah, sorry, I was too focused on the validate_response() function where the most of the validation happens.
It's better with comments, thank you!

PR #1984, opened 2022-12-22
Fix wrong parameter type in page order, fix #1983

### LINKED ISSUE #1983: Change page parent broken in page form
### Describe the Bug
When trying to change the page parent of a page on level 2 and move it into level 3 by changing the page parent, a JavaScript error occurs.

### Steps to Reproduce

1. Open any page on level 2
2. Choose another level 2 page as new parent
3. An error occurs in `Order` box instead of the drop down
4. When saving the page the page parent is not updated

### Expected Behavior
A list of pages should appear.

### Actual Behavior
image

### Additional Information
* No browser console errors
* The error fist appeared with the [2022.12.0](https://github.com/digitalfabrik/integreat-cms/pull/1944) release
* The error message reads `raw is undefined` in development mode

svenseeberg:
### Short description
Fix #1983 where an unexpected parameter type was passed to the `escapeMeta` function.

### Proposed changes
Allow the escapeMeta function to handle any parameter type and always return a valid (String) result.

### Side effects
I thought about splitting the nested function calls in https://github.com/digitalfabrik/integreat-cms/pull/1984/files#diff-12d9bde41ce20ae100fac96355585b5bd8415aeb05cc981c8bbb3b9f1e30f8e6R146 and only pass it to escapeMeta if there is a element. However, this made `getPageOrderTable()` even longer and the result seems to be mostly the same - except for the "unclean" parameter type.

### Resolved issues
Fixes: #1983 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ UNRELEASED
 * [ [#1900](https://github.com/digitalfabrik/integreat-cms/issues/1900) ] Exclude users without view_page permission from page-specific permissions
 * [ [#1956](https://github.com/digitalfabrik/integreat-cms/issues/1956) ] Add Amharic fonts and fix PDF export in Amharic
 * [ [#1906](https://github.com/digitalfabrik/integreat-cms/issues/1906) ] Fix link escape in message in imprint form
+* [ [#1983](https://github.com/digitalfabrik/integreat-cms/issues/1983) ] Fix broken page form ordering box
 
 
 2022.12.2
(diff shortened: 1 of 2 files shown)


svenseeberg:
I'd trigger a release after merging this.

MizukiTemma (approved):
Thank you for speedy fix 🚀

MizukiTemma:
Isn't the issue number in the "Resolved issues" wrong?

svenseeberg:
> Isn't the issue number in the "Resolved issues" wrong?

Definitely yes :D I seems I had the wrong text in my clipboard.

PR #1986, opened 2022-12-23
Limit log message filepath length

svenseeberg:
### Short description
We sometimes get multiple MB of base64 coded files in the log file. This is not helping. I guess we need a deeper fix, but the output should be limited anyways.

### Proposed changes
Just in case limit the path length to 1024 characters in log messages.

### Side effects
If the path is longer then it would be cut. Should not happen too often though.

### Resolved issues
None, this is it.


MizukiTemma (approved):
👍 Thank you!

JoeyStk (approved):
Thank you a lot. Yes, this makes sense

Issue #1968, opened 2022-12-13
Fix permission for Observer role in Media Library


MizukiTemma:
### Describe the Bug

Currently users with role "Observer" can replace files in the media library, and a network error occurs when they try to delete a file.

However, they should not be allowed to delete or replace any files. Only uploading shoud be possible for them.

### Steps to Reproduce
1. Upload some files as root user.
2. Log out once and log in as observer.
3. Go to Media Library.
4. Try to delete/replace a file.

### Expected Behavior

Buttons "Replace file" and "Delete file" are not available for observers, and replacing/deleting requests will be rejected if the user is an observer.

### Actual Behavior

The buttons are shown, deleting will be rejected but triggers a network error, replacing will not be rejected.

### Additional Information

Add also a warning "You don't have permission to delete or replace files but you can upload files" on the top part of  the media library, as in the page tree, poi/event lists.

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

PR #1988, opened 2022-12-30
Fix observer permissions in media library

MizukiTemma:
### Short description

This PR adjusts the permission of observer role so observers can neither delete nor replace media files.

### Proposed changes

- remove the permission replace_mediafile from observer role (delete_media_file was not listed).
- save whether the user is allowed to replace/delete media files <s>in the attribute "value" in the template media_list(_admin)</s> as canReplaceFile and canDeleteFile in MediaContextMixin.
- hide the replace and delete buttons depend on <s>the "value" attribute</s> canReplaceFile and canDeleteFile.
- check permission before replacing/deleting the file and show an warning message if an observer user manages to send a delete/replace request.

### Side effects

- I hope none.
- There is probably a better way to check the permission to hide the buttons......

### Resolved issues

Fixes: #1968

### DIFF
--- a/integreat_cms/cms/migrations/0056_update_observer_permissions.py
+++ b/integreat_cms/cms/migrations/0056_update_observer_permissions.py
@@ -1,0 +1,43 @@
+from django.db import migrations
+from ..constants import roles
+
+
+# pylint: disable=unused-argument
+def update_roles(apps, schema_editor):
+    """
+    Update the role definitions
+
+    :param apps: The configuration of installed applications
+    :type apps: ~django.apps.registry.Apps
+
+    :param schema_editor: The database abstraction layer that creates actual SQL code
+    :type schema_editor: ~django.db.backends.base.schema.BaseDatabaseSchemaEditor
+    """
+    # We can't import the Person model directly as it may be a newer
+    # version than this migration expects. We use the historical version.
+    Group = apps.get_model("auth", "Group")
+    Permission = apps.get_model("auth", "Permission")
+
+    # Assign the correct permissions
+    for role_name in dict(roles.CHOICES):
+        group, _ = Group.objects.get_or_create(name=role_name)
+        # Clear permissions
+        group.permissions.clear()
+        # Set permissions
+        group.permissions.add(
+            *Permission.objects.filter(codename__in=roles.PERMISSIONS[role_name])
+        )
+
+
+class Migration(migrations.Migration):
+    """
+    Migration file to update permissions of observer
+    """
+
+    dependencies = [
+        ("cms", "0055_track_region_deepl_api_usage"),
+    ]
+
+    operations = [
+        migrations.RunPython(update_roles, migrations.RunPython.noop),
+    ]
--- a/integreat_cms/static/src/js/media-management/component/edit-directory-sidebar.tsx
+++ b/integreat_cms/static/src/js/media-management/component/edit-directory-sidebar.tsx
@@ -18,6 +18,7 @@ type Props = {
     globalEdit?: boolean;
     submitForm: (event: Event, successCallback?: (data: any) => void) => void;
     isLoading: boolean;
+    canDeleteDirectory: boolean;
 };
 const EditDirectorySidebar = ({
     directory,
@@ -26,6 +27,7 @@ const EditDirectorySidebar = ({
     globalEdit,
     submitForm,
     isLoading,
+    canDeleteDirectory,
 }: Props) => {
     // This state is a buffer for the currently changed directory
     const [changedDirectory, setChangedDirectory] = useState<Directory>(directory);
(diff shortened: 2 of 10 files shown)


integreat_cms/cms/views/media/media_actions.py
```diff
@@ -275,6 +274,20 @@ def replace_file_ajax(request, region_slug=None):
...
+                "messages": [
+                    {
+                        "type": "error",
+                        "text": _(
+                            "You don't have the permission to replace media files."
+                        ),
+                    }
+                ],
+            }
+        )
```
timoludwig, line 289: If I got it right, this error message is never shown to the user anyway, because the replace button isn't shown for users who don't have the permission to replace media files, right? Then I guess we don't need the verbose error message and a generic 403 error raised by the more concise `@permission_required` decorator is sufficient...

timoludwig (changes requested):
Thanks! :tada:
From what I can tell, this works as expected.
However, I suggest to change the way the permission value is passed to preact.

integreat_cms/cms/templates/media/media_list.html
```diff
@@ -1,8 +1,14 @@
...
 {% block content %}
-    <integreat-media-management
-    class="flex w-full h-full"
+    <integreat-media-management id="integreat-media-management"
+    class="flex w-full h-full" value="
+    {% if not perms.cms.replace_mediafile %}
+        hide
+    {% else %}
+        show
+    {% endif %}
```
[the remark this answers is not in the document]
MizukiTemma: Thank you 🙂 I passed it by MediaContextMixin.

integreat_cms/cms/views/media/media_actions.py
MizukiTemma: hmm, somehow the 403 page will not be called as I tried with @permission_required, showing the button temporary for the observer user and using it to emulate the case an observer makes it to send an delete request 🤔

integreat_cms/cms/migrations/0056_update_observer_permissions.py
```diff
@@ -0,0 +1,47 @@
+from django.db import migrations
+
+ROLES = [
+    {
+        "name": "EDITOR",
+        "remove_permissions": [
+            "replace_mediafile",
+        ],
+    },
+]
```
timoludwig, line 10: Hmm, I see that you tried to optimize the "typical" role-update-migration a bit, but this has the downside of including one potential source of errors... in this case, you used "EDITOR" instead of "OBSERVER", and instead of fixing it, I would just suggest to use our generic approach for the role update... this way, we only have to define our permissions once in the constants and don't have to think about them again in the migrations... we just apply the config from the constants.

timoludwig (changes requested):

integreat_cms/cms/templates/media/media_list.html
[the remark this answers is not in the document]
timoludwig: Thanks, looks very clean now! :+1:

integreat_cms/cms/views/media/media_actions.py
timoludwig: I'm not sure what you mean, if the `@permission_required` really doesn't work, we have a very big problem :sweat_smile:
We also use this decorator in a lot of other views, would be a very critical bug...
Maybe it was connected to the mistake in the migration, so that the observer user still had the permission to replace media files?

timoludwig: It would be very cool to have view tests for the media actions to be absolutely sure we're not messing something up here, but that's not completely trivial... I guess I'll open another issue for it :grin:

timoludwig: For reference: I opened #1997 for that matter.

integreat_cms/cms/migrations/0056_update_observer_permissions.py
MizukiTemma: Oh, what a leathal mistake 😨 

Do you mean a migration like 0043, taking the constants and updating all the permissions and roles?

timoludwig: Yes, exactly! :+1:

integreat_cms/cms/views/media/media_actions.py
MizukiTemma: Thank you for opening an issue for that 👍 

It seems @permission_required itsself is indeed working. I got the 403 error page as I tried to publish a page as observer without permission for it.

However the 403 error page will not be shown when a delete file request is sent by an observer. This remains so also after modification of the migration. 
With @permission_required it returns 403 forbidden to file deleting requests, though no redirect to the 403 error page. I think it has something to do with preact, for the console shows "Uncaught (in promise) Error: 403: Forbidden" but I don't know how and what to change 😓

integreat_cms/cms/migrations/0056_update_observer_permissions.py
MizukiTemma: Done 👍

integreat_cms/cms/views/media/media_actions.py
timoludwig: Ah ok, so the user is correctly prevented from replacing/deleting files (even if they manage to show the button in the developer tools), this is all we're caring about.
It doesn't have to look nice in this case, because it will only happen if the user tampers with the HTML source code. Users who do that are not entitled to nice error messages :sweat_smile:

MizukiTemma: Yes, permission check and resulting rejection of permissionless delete/replace requests are successfull.
I'll implement with @permission_required. It brings more consistency 🙂

MizukiTemma:
The button to delete a directory is now hidden for users who are not allowed to.

integreat_cms/static/src/js/media-management/component/edit-directory-sidebar.tsx
```diff
@@ -124,6 +126,7 @@ const EditDirectorySidebar = ({
                                     }`}
                                     className={cn("btn", {
                                         "btn-red": !isLoading && directory.numberOfEntries === 0,
+                                        "hidden": !canDeleteDirectory,
```
timoludwig, line 129: Normally, in (p)react, you don't add the class "hidden" to elements which should not be displayed, but you just don't render it, see e.g. [Conditional Rendering](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator):

```
{canDeleteDirectory && (
    <button
        ...
    </button>
)}
```

integreat_cms/static/src/js/media-management/component/edit-sidebar.tsx
```diff
@@ -247,7 +251,10 @@ const EditSidebar = ({
                                         className={cn(
                                             "w-full text-white text-center font-bold py-3 px-4 m-0 rounded",
                                             { "cursor-not-allowed bg-gray-500": isLoading },
-                                            { "bg-blue-500 hover:bg-blue-600": !isLoading }
+                                            { "bg-blue-500 hover:bg-blue-600": !isLoading },
+                                            {
+                                                hidden: !canReplaceFile,
+                                            }
```
timoludwig, line 257: In this case, I'd also use `{!selectionMode && canReplaceFile && (` in line 247 instead of the hidden class.

integreat_cms/static/src/js/media-management/component/edit-sidebar.tsx
```diff
@@ -256,7 +263,13 @@ const EditSidebar = ({
...
                                 <button
                                     title={mediaTranslations.btn_delete_file}
-                                    className={cn("btn", { "btn-red": !isLoading })}
+                                    className={cn(
+                                        "btn",
+                                        {
+                                            hidden: !canDeleteFile,
+                                        },
+                                        { "btn-red": !isLoading }
+                                    )}
```
timoludwig, line 272: And here also:

```
{canDeleteFile && (
    <button
        ...
    </button>
)}
```

timoludwig (changes requested):
Thanks! :+1:
Sorry for not noticing earlier, I have another suggestion to improve the preact rendering.
Also, the failed CircleCI check was my fault, I fixed this in #2000. After this is merged and you rebased your branch, the error should be fixed.

integreat_cms/static/src/js/media-management/component/edit-directory-sidebar.tsx
MizukiTemma: That's nice to know 😃 Thank you for the advice! Now applied.

seluianova (approved):
Migration works ok, 'replace_mediafile' removed for Observer (/)
Buttons to delete directory / delete file / replace file are not available now for Observer (/)

Issue #1769, opened 2022-10-17
Evenly distribute form sidebar boxes


ulliholtgrave:
### Motivation

Right now our sidebars in the content forms have a different behavior on different screen sizes. With the addition of the HIX widget the height is no longer aligned and we should think of some stable alignment mechanism for all content forms.

- [x] Page form (#1990)
- [ ] Event form (#2019)
- [ ] Location form (#2019)

This depends on:
- [x] #1876

![image](https://user-images.githubusercontent.com/15608678/196170622-8891a16f-6045-4f73-8653-1c486a1d904a.png)

PeterNerlich:
Where is this behaviour visible, considering the test data we have?

timoludwig:
You can insert a fake-key for the Textlab API, create a file `/etc/integreat-cms.ini` with the contents:
```
[develop]
TEXTLAB_API_KEY = dummy
```
Then restart the server, enable HIX in a region and after that, the widget should be shown (although a bit smaller because there is an error due to the incorrect token)... you can however manually increase its size to simulate the production UI by temporarily adding the class `h-[700px]` to [this line](https://github.com/digitalfabrik/integreat-cms/blob/307d495402e58822c33616227118de366fcd586e/integreat_cms/cms/templates/hix_widget.html#L4) (it's important to actually change the source code and not add the class in the Browser console, because webpack needs to create this class first via the Tailwind JIT compiler).

Gaston69:
Peter and me searched for a solution, using "column-2" class from tailwind. 
Unfortunately that breaks grouped content (by <div>) into two columns as well in Chrome (seams to work good in Firefox). So we decided leave it as it is. If someone has any other suggestions / solutions please tell us.

PeterNerlich:
```diff
diff --git a/integreat_cms/cms/fixtures/test_data.json b/integreat_cms/cms/fixtures/test_data.json
index 9425818b4..c9983d9fd 100644
--- a/integreat_cms/cms/fixtures/test_data.json
+++ b/integreat_cms/cms/fixtures/test_data.json
@@ -226,7 +226,7 @@
       "admin_mail": "augsburg@integreat-app.de",
       "created_date": "2019-08-12T07:48:17.997Z",
       "last_updated": "2019-08-12T07:48:17.998Z",
-      "statistics_enabled": false,
+      "statistics_enabled": true,
       "matomo_id": null,
       "matomo_token": "",
       "offers": [1, 2, 3]
[...]

PR #1990, opened 2023-01-10
Improve spacing in page form

JoeyStk:
### Short description
This PR is supposed to improve the spacing inside the page form. Up until now there was some empty space when the "Settings" tab was closed. This PR solves this issue.

### Proposed changes

- Remove empty space when the "Settings"-tab is closed
- Remove nesting of divs inside the sidebar

### Side effects

- For screens larger than 2100px in width there is no third column anymore because I changed the nesting of the divs. I think this is okay, but if would also be fine with finding a better solution. Please let me know what you think. Take a look at the changes before/after in the screenshots below.

Before: 
grafik

After;
grafik

### Resolved issues

Fixes: #1769 


timoludwig:
> For screens larger than 2100px in width there is no third column anymore because I changed the nesting of the divs. I think this is okay, but if would also be fine with finding a better solution. Please let me know what you think. Take a look at the changes before/after in the screenshots below.

Hmm, yeah so I would actually prefer that we use available screen real estate on larger screens. I don't exactly know what screen sizes are common in our user group, but probably the fraction of people with larger screens will grow in the future.
If there is only one column, it's hardly usable when all boxes are expanded, because one needs a lot of vertical scrolling to view all the options, even though most of the remaining screen is empty...
So in my opinion, we should look for a solution with two columns, e.g. like [I suggested in the issue comments](https://github.com/digitalfabrik/integreat-cms/issues/1769#issuecomment-1322075492).

JoeyStk:
Thank you a lot for your feedback. I understand the issue. I took a look at the suggestion but then thought that there might be a simpler solution to this. Feel free to check out my new solution. It should solve the mentioned issue. So far there are still some minor issues with the margins to the side and the width, but I wanted to make sure that the underlying logic is okay before I work on the details. If there are still issues with this solution that I didn't think of, I will try your suggestion next :)! I hope that is okay :)

timoludwig:
Hmm, I'm not really sure I understand - for me, the screen looks like this above 2100px:

(So not really evenly distributed... even when the margins are fixed, the boxes of the left sidebar column take more than twice as much space than the boxes of the right column... maybe you could try to explain your approach and what you were trying to achieve? I don't really understand the changes you made to the classes...)

Screenshot 2023-01-10 at 19-32-23 Integreat Editorial System

JoeyStk:
Thank you again for your feedback. 

Which challenge did I try to solve
- Remove the empty spacing when "Settings" are closed
grafik
- Remove the empty spacing inside the editor when Settings are opened
grafik
- Have two columns below the editor on Tablet screens, have one column on the right side of the editor on smaller Desktop screens and have two columns on bigger Desktop screens 


How I did it:
- Split the page form into two main divs: the editor and the sidebar. Then inside the sidebar div create two divs for the two columns in the sidebar (for challenges 1 & 3). Depending on screen size enable or disable flex/block (flex for Tablet and big Desktop screens, block for smaller Desktop screens)(for the third challenge)
- Remove class flex-auto from the editor's div to remove the white space inside the editor (for challenge nr. 2)


So mainly I was trying to avoid changing the order of the content boxes and also tried to avoid the boxes overflowing to the next column. At least for me it's less painful as a user to do some vertical scrolling than having a different arrangement of the content boxes dependent on which screen I'm using. But I understand now, that my solution only solves side effects of the actual issue. So I guess I'm coming to the conclusion that at least I cannot solve the issue with plain CSS, so I will try your TS solution next

timoludwig (approved):
Amazing, thanks a lot! :heart_eyes: 
Works like a charm!
I've added three small changes:
- I think the left column should be the larger one
- I refactored the partitioning function so we don't have to search boxes by their height... this has the advantage that it works reliable even when two boxes have the same height
- I also converted the hix widget to a collapsible box to enable it for this feature

.eslintrc.js
```diff
@@ -106,7 +106,7 @@ module.exports = {
         "no-magic-numbers": [
             "error",
             {
-                ignore: [-1, 0, 1, 100],
+                ignore: [-1, 0, 1, 2, 100],
```
charludo, line 109: This addition probably makes sense, though in that case, please remove 
```
/* eslint-disable-next-line no-magic-numbers */
```
where the magic number 2 is currently being used in the project, since the line is not necessary anymore. Alternatively, ignore the 2 in your new file and remove it from the ignore list.

charludo (changes requested):
Very nice! Much better use of space in the sidebar(s), thank you :)

.eslintrc.js
timoludwig: Makes sense! Since I changed the config when reworking this, I also removed the unnecessary exemptions now.

charludo (approved):

Issue #1552, opened 2022-06-15
Slug field layout broken when permalink is too long


timoludwig:
### Describe the Bug

When the permalink of the parent page is too long, the slug field cannot be edited in the page form.

### Steps to Reproduce

1. Create page with very long slug
2. Create child-page of this page
3. Try to edit slug of child page

### Expected Behavior

The slug field should be scrollable horizontally

### Actual Behavior

The slug field is fixed and the slug cannot be edited because it is hidden beneath the other elements

### Additional Information

![Screenshot 2022-06-15 at 10-24-08 Integreat Redaktionssystem](https://user-images.githubusercontent.com/16021269/173780180-89c98c30-be6f-43cf-b2bb-a1b1c748cacf.png)

PR #1991, opened 2023-01-10
Make slugfield scrollable

PeterNerlich:
### Short description

Make slugfield usable, even if URI prefix is very long.

### Proposed changes

- make slugfield scrollable by setting `overflow: auto` on the container
- ensure `input` is never at 0 width
- reduce label opacity to $\frac{2}{3}$ for better distinction from input

### Side effects

- Other places using `.slug-field` will be affected. We estimate this to be fine.

### Resolved issues

Fixes: #1552 (Slug field layout broken when permalink is too long)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ UNRELEASED
 * [ [#1769](https://github.com/digitalfabrik/integreat-cms/issues/1769) ] Evenly distribute page, event and poi form sidebar boxes
 * [ [#1876](https://github.com/digitalfabrik/integreat-cms/issues/1876) ] Make items in the sidebar of poi and event form toggleable
 * [ [#2023](https://github.com/digitalfabrik/integreat-cms/issues/2023) ] Rephrase texts in text understandability-box
+* [ [#1552](https://github.com/digitalfabrik/integreat-cms/issues/1552) ] Make slugfield scrollable
 
 
 2023.1.1
--- a/integreat_cms/static/src/css/style.scss
+++ b/integreat_cms/static/src/css/style.scss
@@ -203,17 +203,20 @@ select,
 
 .slug-field {
     @apply flex px-3 py-2 whitespace-nowrap;
+    overflow: auto;
     &:focus-within {
         @apply bg-white border-blue-500 ring-1 ring-blue-500 outline-none;
     }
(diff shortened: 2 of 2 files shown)


timoludwig (commented):
Thanks a lot, this fixes the issue! :+1:

If the slug itself is longer than the input field, we have a "stacked" scrolling now, I was wondering whether we could combine this, so that the input field automatically grows when longer slugs are entered so we only have one layer of scrolling, but I can imagine this is probably a bit complicated, so only improve this if it doesn't cost you too much time :sweat_smile:
I'm already happy with the current solution! :+1:

Screenshot 2023-01-10 at 19-04-46 Integreat Redaktionssystem



> I guess the issue number mensioned in the PR description is not correct...

True, I fixed it

timoludwig:
Oh, and just a minor addition to my comment:
Please [separate the commit subject from body with a blank line](https://cbea.ms/git-commit/#separate), commit subjects with more than 72 characters are a bit hard to read:
Screenshot 2023-01-10 at 20-22-35 Make slugfield scrollable by PeterNerlich · Pull Request #1991 · digitalfabrik_integreat-cms

So I would use e.g. the following commit message instead:
```
Make slugfield scrollable

- Make slugfield scrollable by setting `overflow: auto` on the container
- Ensure `input` is never at 0 width
- Reduce label opacity to 60% for better distinction from input
```

charludo (approved):
LGTM, thanks!

PeterNerlich:
> If the slug itself is longer than the input field, we have a "stacked" scrolling now, I was wondering whether we could combine this, so that the input field automatically grows when longer slugs are entered so we only have one layer of scrolling, but I can imagine this is probably a bit complicated [...]

Yes, that is not possible without using JS/TypeScript magic that might act differently for different browsers, and since it would add probably quite a bunch of complexity and we don't have a very clear structure where this logic resides and how it is tested, if at all (at least that we know of), Ingo strongly favoured not investing more time, and I agree. I would have ideas how to hack it, though.

timoludwig (changes requested):
> Ingo strongly favoured not investing more time, and I agree.

Sounds reasonable! Yes, this leaves us with the changelog entry and the commit message.

timoludwig (approved):
:+1:

PR #1993, opened 2023-01-10
Skip duplicate translations in XLIFF

svenseeberg:
### Short description
When importing an XLIFF zip file that contains 2 files for the same page translation, the CMS currently throws an internal error because the same new version is created twice. Instead of throwing an error we should display a warning to the user in the diff view that the zip file contains duplicates page translations.

### Proposed changes
- When creating the diff, check for duplicate page translations in the diff list. Show warnings for duplicates.
- The import function itself logs an error but continues importing other page translations.

### Side effects
- If the user is not careful and overlooks the message, a translation may receive the wrong text. However, I consider this a minor inconvenience as this can be fixed at any later point in time. 
- Users may disregard the error notification and thus a page translation may not be up to date.

### Resolved issues
https://chat.tuerantuer.org/digitalfabrik/pl/3ecz8rnmxibmtenkmeiz9t1qfc


integreat_cms/xliff/utils.py
```diff
@@ -319,6 +319,10 @@ def get_xliff_import_diff(request, xliff_dir):
                 page=page_translation.page,
                 language=page_translation.language,
             )
+            if existing_translation in [diff_item["existing"] for diff_item in diff]:
+                duplicate_message = _("Duplicate page translation in XLIFF, skipping")
+                messages.warning(request, f"{duplicate_message} {xliff_file}")
+                continue
```
timoludwig, line 325: Maybe we should place this check in the `xliffs_to_pages` function? If we only adapt the diff, I think the actual translations which are imported will still contain the duplicate...

timoludwig:
Ah, and I guess you have to re-authenticate to CircleCI:

Screenshot 2023-01-10 at 19-11-54 Pipelines - digitalfabrik_integreat-cms

integreat_cms/xliff/utils.py
svenseeberg: Yes, indeed the better place. Also this iterator would not be necessary there.

svenseeberg: ... on the other hand we would have to introduce higher level functionality to the `xliff_to_pages` function. Currently, it does not know anything about page translations, existing versions, etc. This could work by comparing the raw values of the page id and translation language code. Maybe it is actually better to check in both locations where there already is some more context?

timoludwig: Ah ok true, yeah or we check it only in `get_xliff_import_diff()` and `xliff_import_confirm()` and accept the fact that `xliff_to_pages()` might return duplicates?

svenseeberg: I think I found a somewhat satisfying solution: We warn the user in the diff view about the duplicate and then in the actual import catch the error and print it to the log file. This has the generic advantage that just because saving one page translations fails the full XLIFF import is not aborted.

image

image

```
Jan 10 19:35:09 ERROR integreat_cms.xliff.utils - Error when importing new new version for <PageTranslation (id: 85, page_id: 23, language: en, slug: interpret-and-translate)> from 'augsburg_de_en_23_1_dolmetschen-und-übersetzen.xliff' by <SimpleLazyObject: <User (id: 1, username: root, team: CMS team, superuser)>>: (IntegrityError('duplicate key value violates unique constraint "pagetranslation_unique_version"\nDETAIL:  Key (page_id, language_id, version)=(23, 2, 2) already exists.\n'))
```

MizukiTemma (approved):
Thank you 🙂 Looks good, working as expected 👍

integreat_cms/xliff/utils.py
```diff
@@ -319,39 +320,74 @@ def get_xliff_import_diff(request, xliff_dir):
...
+                    format_html(
+                        __(
+                            _(
+                                "Page <b>{}</b> from file <b>{}</b> was also translated in file <b>{}</b>."
+                            ),
+                            _(
+                                "Please check which of the files contains the most recent version and upload only this file."
+                            ),
+                            _(
+                                "If you confirm this import, only the first file will be imported and the latter will be ignored."
```
svenseeberg, line 337: Can we guarantee ordering? I'm not sure.

integreat_cms/xliff/utils.py
```diff
@@ -414,34 +448,62 @@ def xliff_import_confirm(request, xliff_dir, machine_translated):
...
+                                    ),
+                                    _(
+                                        "If the problem persists, contact an administrator."
+                                    ),
+                                ),
+                                page_translation.readable_title,
+                                xliff_file,
                             ),
                         )
+                    else:
```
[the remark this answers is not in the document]
svenseeberg: Good idea.

integreat_cms/xliff/utils.py
```diff
@@ -319,39 +320,72 @@ def get_xliff_import_diff(request, xliff_dir):
...
-                                )
-                            )[2:]
+            if existing_translation in diff:
+                # Show global error to indicate duplicate translation
+                messages.error(
+                    request,
+                    format_html(
+                        __(
+                            _(
+                                "Page <b>{}</b> from file <b>{}</b> was also translated in file <b>{}</b>."
```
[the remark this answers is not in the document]
svenseeberg: I thought about that and decided it was not really important as the XLIFF files usually are named similar. At least if they were exported from our CMS. But sure, this is even better.

integreat_cms/xliff/utils.py
svenseeberg: I think not. https://docs.python.org/3/library/glob.html#module-glob states in the first sentence:

> results are returned in arbitrary order

svenseeberg: I'd suggest that we simply call `sort()` when iterating over the returned list.

timoludwig: Indeed, in my tests I also noticed that the order is arbitrary. I meant "first" in relation to the position in the sentence. So in the message "Page X from file Y was also translated in file Z.", I interpreted "Y" as the "first" file, and "Z" as the latter...
Do you have a better idea for the wording? Maybe "Page X was translated in two files, only file Y will be imported and file Z will be ignored"?

timoludwig: Calling `sort()` won't change much, since if the user interprets "first" in relation of the file in the zip file, they will open their zip file and think that the first file which appears is the first file. But if we call sort, this is not the case.

svenseeberg: Okay, we could just rephrase to `One of the following files will not be imported` ?

timoludwig: But I mean we already know that the existing diff entry `diff[translation_key]["file"]` will be imported and the file `xliff_file` which comes later in the loop will be ignored. I tried to phrase the sentence in a way that points that out. Do you have a better wording in mind?

timoludwig: Ahh wait, I'm stupid, sorry 😁
I never experienced differences in the order between the diff calculation and the actual import, but yes, according to the docs the order might differ. So yes I agree that it makes sense to run sort on the list to make the order deterministic and we're able to tell the user which version will be imported...

svenseeberg: Alright, thanks for figuring it out :)

svenseeberg:
I cannot officially add an approve, but this is good to go IMHO. As the major changes are now contributed by @timoludwig , I'd leave the merge button to you.

Issue #1831, opened 2022-11-04
Create management section for machine translation and add MT settings


JoeyStk:
### Motivation
We want a new management section for automated translations

### Proposed Solution
- [ ] Add a management section for automated translations and add settings to enable/disable machine translation for specific content types. 
- [ ] Add menu item but make it hidden (only visible in DEBUG mode) until the follow-up issues are finished and the view is completed

### Additional Context
This is part of the issue family to enable automated translations in our system.
This blocks:
- [ ] #1926
- [ ] #1927

![management](https://user-images.githubusercontent.com/16021269/204849061-fc697686-64cf-42bf-9b12-82c464018023.png)
— [Source](https://www.figma.com/file/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=166%3A1980)

Issue #1927, opened 2022-11-30
Add possibility to disable specific languages for machine translations


JoeyStk:
### Motivation
In our machine translation management section we want to add an overview of the available languages.

### Proposed Solution
- [ ] Add a new (array ?) field to the region model to store all enabled/disabled languages for machine translations
- [ ] Add a form to change these languages to the machine translation management section
- [ ] In all places where machine translations are generated, first check whether the corresponding language is enabled for machine translations in this region.

### Additional Context
Part of our meta issue for machine translations

This depends on:
- [x] #1831 

### Design Requirements

![language selection](https://user-images.githubusercontent.com/16021269/204850134-4db8ca5c-a706-4597-80ad-257c13e49860.png)
— [Source](https://www.figma.com/file/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=166%3A1980)

timoludwig:
@nikolahoff How was this supposed to work, is this just a read-only overview over all languages which are available for machine translations, or should it also be possible to disable specific languages (even if they are available)?

osmers:
@timoludwig it should be possible to disable specific languages, even if they are theoretically available for MT. The idea is that certain partners might say that the quality of a specific language is too poor for their standard, which is why they disable it altogether.

charludo:
Is this supposed to disable translations *to* a language, or *from* that language?

And regardless of the answer, don't we also need the opposite direction as well?

e.g. a region does not want to use SUMM.AI, so they disable translation *to* "German (easy)". But they also want to disable translations *from* any language except "German" (e.g. to prevent machine translating machine translations).

Unless this has already been discussed and decided elsewhere, I think the PR for this issue should add 2 new columns, one *from*, one *to*.

timoludwig:
I think this setting is about the target languages. And probably this is sufficient, because the source language is implicitly stated by the language tree? A from & to setting would probably be too complex anyways?

charludo:
> the source language is implicitly stated by the language tree

So it's not possible to translate from anything but the "root" language?

But yes, in that case it's clear.

timoludwig:
Except a region has a nested language tree, e.g. "de -> en -> fr" or something...
Then it would be up to the region managers to disable MT for"en" if they also want to enable them for "fr" and don't want to machine translate already machine translated pages...

charludo:
[...]

Issue #1926, opened 2022-11-30
Add translation budget graph to machine translation management section


timoludwig:
### Motivation

After #1831 is implemented, we want to extend the view with a nice graph to visualize the translation budget usage of the region.

### Proposed Solution

- Add a graph which shows the usage of the current budget
- Show date when the budget is next reset

### Additional Context

This is blocked by 
- [x] #1345
- [x] #1831

![usage-dashboard](https://user-images.githubusercontent.com/16021269/204849678-79a969b4-4aac-468f-8330-4d6fa82c5a39.png)
— [Source](https://www.figma.com/file/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=166%3A1980)

Issue #1857, opened 2022-11-11
Manage permission in fixture again instead of migrations


timoludwig:
### Motivation

In the early stages of this project, we stored the permission in a [fixture](https://github.com/digitalfabrik/integreat-cms/blob/00610f05efaeb8bd67b63f74110a07db58d1e1df/integreat_cms/cms/fixtures/roles.json) which we could just update by running `integreat-cms-cli loadddata roles`.

The main problem with this approach was that we referenced permissions by ids, and on the dev systems the ids of the permissions changed each time the database was reset after a model change. Thus, I suggested the current mechanism of managing roles via migrations (see #748).

However, I wasn't aware at the time that the readable (and constant) permission names could also be used as keys in the fixture instead of the numeric ids. Considering the current overhead of having to change the constants and then generate a migration which applies the changed constants, I think the previous approach wasn't too bad if we change the way we reference the permissions.

### Proposed Solution

- Create a fixture `integreat_cms/cms/fixtures/permissions.json`
- Convert the current definitions from [integreat_cms/cms/constants/roles.py](https://github.com/digitalfabrik/integreat-cms/blob/a0b389f3d28f8c66ba152ec52246cd244bf93433/integreat_cms/cms/constants/roles.py) to json and put it into the new file
- Remove all permissions from roles.py (but keep the role definitions)
- [Squash the migrations](https://docs.djangoproject.com/en/3.2/topics/migrations/#migration-squashing) to get rid of all the role-changing migrations and speed up the setup of new systems and the tests (which need to run the migrations before each test run)

### Alternatives

Just keep the current setup

timoludwig:
The file https://github.com/digitalfabrik/integreat-cms/blob/d3bd93af510d28d576d5a5f6629445b44fc17c48/integreat_cms/cms/fixtures/test_roles.json introduced in #2126 could be a good place to start for this.

osmers:
Stale

PR #1994, opened 2023-01-11
Create management section for machine translations

charludo:
### Short description
Add machine translations management view and form. Users of adequate permission level (manager and up) can change which type of content can be machine translated in a region. These settings are currently without effect. 

### Proposed changes

- add "Translations" section in the sidebar (currently only viewable in debug mode)
- create translations management form
- alter the region model to store the new settings 

### Side effects

- none 

### Resolved issues

Fixes: #1831


JoeyStk (changes requested):
Thank you so much! This is PR is so cool!
I have four thoughts:
1. I think this PR deserves an entry in our changelog, because it's a feature that is also interesting to other teams, for example the service-team
2. I think you can already remove the translation report from the Analysis section. AFAIK it's only supposed to be inside the translations section
3. AFAIK @Gaston69 and @PeterNerlich are working on an issue, that introduces the third section. So maybe it could be useful to them to know that this PR exists and that it already introduces the "translation" section
4. @osmers Could you please double-check if everything is as it's supposed to be? 

grafik
grafik

integreat_cms/cms/constants/roles.py
```diff
@@ -101,6 +101,7 @@
     "delete_mediafile",
     "delete_organization",
     "grant_page_permissions",
+    "manage_translations",
```
timoludwig, line 104: Setting new permissions also require a database migration to apply them on the existing production systems.

integreat_cms/cms/constants/machine_translation_permissions.py
```diff
@@ -0,0 +1,20 @@
+"""
+This module contains string representations of permission levels for machine
+translations, used by :class:`~integreat_cms.cms.models.regions.region.Region`.
+"""
+from django.utils.translation import gettext_lazy as _
+
+
+#: Machine Translations are disabled
+NOONE = 0
```
timoludwig, line 9: I'd find `NO_ONE` or `NOBODY` more readable...

integreat_cms/cms/constants/machine_translation_permissions.py
```diff
@@ -0,0 +1,20 @@
...
+#: Everyone can use Machine Translation
+EVERYONE = 1
+#: Only Managers can use Machine Translations
+MANAGERS = 2
+
+#: Choices to use these constants in a database field
+CHOICES = (
+    (NOONE, _("No")),
+    (EVERYONE, _("Yes")),
+    (MANAGERS, _("Yes, only Managers")),
```
timoludwig, line 19:
```suggestion
    (MANAGERS, _("Yes, only managers")),
```

integreat_cms/cms/models/regions/region.py
```diff
@@ -370,6 +375,24 @@ class Region(AbstractBaseModel):
...
+    machine_translate_events = models.PositiveIntegerField(
+        choices=machine_translation_permissions.CHOICES,
+        default=machine_translation_permissions.EVERYONE,
+        verbose_name=_("Events"),
+    )
+
+    machine_translate_pois = models.PositiveIntegerField(
+        choices=machine_translation_permissions.CHOICES,
+        default=machine_translation_permissions.EVERYONE,
+        verbose_name=_("POIs"),
```
timoludwig, line 393:
```suggestion
        verbose_name=_("Locations"),
```
`POIs` is only our internal model name for historic reasons, in the UI we use the term `Location`.

integreat_cms/cms/templates/translations/translations_management.html
```diff
@@ -0,0 +1,43 @@
...
+                        <label for="{{ form.machine_translate_pois.id_for_label }}">{{ form.machine_translate_pois.label }}</label>
+                        {% render_field form.machine_translate_pois %}
+                    </div>
+                </div>
+            </div>
+        </div>
+    </div>
+</div>
+</form>
+{{ media_config_data|json_script:"media_config_data" }}
```
timoludwig, line 41:
```suggestion
```
There will probably not be a media library on this page, so we can remove the corresponding config data

charludo:
Thanks for the review!

> 1. I think this PR deserves an entry in our changelog

Yes, I agree, though I thought about leaving it out until all related issues are closed and the management page becomes public.

Btw, would it be alright if I did the two blocked issues (#1926, #1927) in this PR as well? They're not too much extra effort on top. In that case I'd put this PR back into draft mode.

> 2. I think you can already remove the translation report from the Analysis section. AFAIK it's only supposed to be inside the translations section

Ah, thanks for bringing that up! I can't for the life of me find the issue again, but there was a discussion between @timoludwig and @osmers that concluded the report would stay in the analysis section. Since the design of this issue still has it in the translation section though, I wasn't sure what the final decision was. But one of them will have to be removed, yes.

integreat_cms/cms/views/translations/translations_management_view.py
```diff
@@ -0,0 +1,95 @@
...
+
+        :return: The rendered template response
+        :rtype: ~django.template.response.TemplateResponse
+        """
+
+        if not request.user.has_perm("cms.manage_translations"):
+            raise PermissionDenied(
+                f"{request.user!r} does not have the permission to manage machine translation settings."
+            )
+
```
timoludwig, line 77:
```suggestion
```
Since you already required this permission via the
```
@method_decorator(permission_required("cms.manage_translations"), name="dispatch")
```
line for all http methods (`dispatch()` is called for both `get()` and `post()`, you don't need to check it here again.

integreat_cms/cms/views/translations/translations_management_view.py
```diff
@@ -0,0 +1,95 @@
...
+        form = TranslationsManagementForm(
+            data=request.POST,
+            instance=request.region,
+        )
+
+        if not form.is_valid():
+            # Add error messages
+            form.add_error_messages(request)
```
timoludwig, line 85: Typically, an incorrectly submitted form is directly rendered again instead of redirecting to keep the entered values and allow the template tags like `|add_error_class:"border-red-500"` to work (this needs the invalid form object for rendering, whereas when you redirect to the get view again, the form is initialized again with no data).

integreat_cms/cms/views/translations/translations_management_view.py
```diff
@@ -0,0 +1,95 @@
...
+        else:
+            # Save the machine translation settings
+            form.save()
```
timoludwig, line 88: I'd show a success message in this case

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -1280,6 +1280,18 @@ msgstr "Funktionierender interner Link"
...
+#: cms/constants/machine_translation_permissions.py
+msgid "No"
+msgstr "Nein"
+
+#: cms/constants/machine_translation_permissions.py
+msgid "Yes"
+msgstr "Ja"
+
+#: cms/constants/machine_translation_permissions.py
+msgid "Yes, only Managers"
```
timoludwig, line 1292:
```suggestion
msgid "Yes, only managers"
```

timoludwig (changes requested):
Thanks a lot! :+1:
Works flawlessly, I only have a few minor comments.

> Yes, I agree, though I thought about leaving it out until all related issues are closed and the management page becomes public.

Exactly, since even our test system does not run in `DEBUG` mode, only developers can see the new section at the moment. So it makes sense to include the changelog entry only when the service team is able to use it.

> Btw, would it be alright if I did the two blocked issues (#1926, #1927) in this PR as well? They're not too much extra effort on top. In that case I'd put this PR back into draft mode.

Sure, if you want to. But since the review of this PR is almost finished, it's probably easier for the reviewers if we keep the chunks smaller.

> Ah, thanks for bringing that up! I can't for the life of me find the issue again, but there was a discussion between @timoludwig and @osmers that concluded the report would stay in the analysis section. Since the design of this issue still has it in the translation section though, I wasn't sure what the final decision was. But one of them will have to be removed, yes.

Since the new section is only for machine translations and the translation report also contains the human translations, I think it makes more sense to keep that view in the analytics section (see https://github.com/digitalfabrik/integreat-cms/issues/1830#issuecomment-1333669373)

integreat_cms/cms/views/translations/translations_management_view.py
timoludwig: Oh and by the way, could you also add tests for the new view (both `get()` and `post()`) to the [view config](https://github.com/digitalfabrik/integreat-cms/blob/f973afc2eb7813ea96dd434c235c69cb3f4afd50/tests/cms/views/view_config.py#L31)?

osmers:
@charludo yes, I agree with @timoludwig and I think it is also what we agreed upon previously, that the translation report should remain under analysis and not be available under translations.
Under the management section of translations, it should also be possible to select which languages should be machine translatable, right?

integreat_cms/cms/views/translations/translations_management_view.py
charludo: > (this needs the invalid form object for rendering, whereas when you redirect to the get view again, the form is initialized again with no data)

Is the other method preferred, i.e. should I change the way Iḿ handling the form?

charludo:
@osmers yes, the translatable language settings will be implemented in #1927

integreat_cms/cms/views/translations/translations_management_view.py
timoludwig: For successful form submissions, it's good to do a `redirect` afterwards (e.g. to prevent accidental re-submits when the page is refreshed etc), but when there are errors, I suggest to directly call `render` instead of the redirect to keep all the invalid values in the form and make it easier to correct them.

integreat_cms/cms/constants/roles.py
charludo: I'm not really clear on this. Do I need to add a migration like [this one](https://github.com/digitalfabrik/integreat-cms/blob/f973afc2eb7813ea96dd434c235c69cb3f4afd50/integreat_cms/cms/migrations/0005_grant_imprint_deletion_permission.py)? Attempting this results in an `Permission matching query does not exist.` error during migration.

timoludwig: Yes, when you change permissions which are created in the same migration, you have to emit the post migrate signal in order for the auth app to pick up the model changes and create the permission object, e.g. like here:

https://github.com/digitalfabrik/integreat-cms/blob/f973afc2eb7813ea96dd434c235c69cb3f4afd50/integreat_cms/cms/migrations/0044_alter_user_options.py#L22-L23

timoludwig: Or wait, where is this permission assigned? If it doesn't exist yet, you have to add it to one of the models, probably the user model?

charludo: > probably the user model

Thanks, that did it!

integreat_cms/cms/migrations/0058_translations_management_settings.py
```diff
@@ -0,0 +1,86 @@
...
+    translations_management_permission = Permission.objects.get(
+        codename="manage_translations"
+    )
+
+    for role_name in dict(roles.STAFF_ROLES):
+        group = Group.objects.get(name=role_name)
+        group.permissions.add(translations_management_permission)
+
+    management_group = Group.objects.get(name="MANAGEMENT")
+    management_group.permissions.add(translations_management_permission)
```
timoludwig, line 30: I have thought about this a lot - I think my current preference would be to just use the sledgehammer approach and unassign all permissions and re-assign them according to the config dict - this way, we avoid any mistakes that result from inconsistencies between those two definitions. 

```suggestion
    # Assign the correct permissions
    for role_name in dict(roles.CHOICES):
        group, _ = Group.objects.get_or_create(name=role_name)
        # Clear permissions
        group.permissions.clear()
        # Set permissions
        group.permissions.add(
            *Permission.objects.filter(codename__in=roles.PERMISSIONS[role_name])
        )
```

However, I'm still a bit unhappy with our current migration workflow, so we should probably find something more sensible in the long term (see #1857)

integreat_cms/cms/views/translations/translations_management_view.py
```diff
@@ -0,0 +1,105 @@
...
+logger = logging.getLogger(__name__)
+
+
+@method_decorator(permission_required("cms.manage_translations"), name="dispatch")
+class TranslationsManagementView(TemplateView):
+    """
+    View for showing the machine translations management options
+    """
+
+    template_name = "translations/translations_management.html"
```
timoludwig, line 23:
```suggestion
    #: The template to render (see :class:`~django.views.generic.base.TemplateResponseMixin`)
    template_name = "translations/translations_management.html"

    #: The context dict passed to the template (see :class:`~django.views.generic.base.ContextMixin`)
    extra_context = {"current_menu_item": "translations_management"}
```
We could define the current menu item context here to avoid the duplication

integreat_cms/cms/views/translations/translations_management_view.py
```diff
@@ -0,0 +1,105 @@
...
+
+        if not form.is_valid():
+            # Add error messages
+            form.add_error_messages(request)
+            return render(
+                request,
+                self.template_name,
+                {
+                    **self.get_context_data(**kwargs),
+                    "current_menu_item": "translations_management",
```
timoludwig, line 87:
```suggestion
```

integreat_cms/cms/migrations/0058_translations_management_settings.py
```diff
@@ -0,0 +1,86 @@
...
+
+    :param apps: The configuration of installed applications
+    :type apps: ~django.apps.registry.Apps
+
+    :param schema_editor: The database abstraction layer that creates actual SQL code
+    :type schema_editor: ~django.db.backends.base.schema.BaseDatabaseSchemaEditor
+    """
+    Group = apps.get_model("auth", "Group")
+    Permission = apps.get_model("auth", "Permission")
+
```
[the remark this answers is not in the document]
charludo: Ah, thanks, I thought the comments were an either/or thing.
And thank you for the thorough re-review :D 
Hope everything's OK now.

timoludwig (approved):
Awesome, I think that's it! :tada:

JoeyStk:
Amazing!

JoeyStk (approved):

Issue #1954, opened 2022-12-08
Text direction not working anymore in PDF export


timoludwig:
### Describe the Bug

Somehow RTL languages are printed LTR in the PDF export

### Steps to Reproduce

1. Go to a right-to-left written language, e.g. https://integreat.app/kreishoexter/ar/%D8%A7%D9%84%D8%AD%D9%8A%D8%A7%D8%A9-%D8%A7%D9%84%D9%8A%D9%88%D9%85%D9%8A%D8%A9/%D8%A7%D9%84%D8%B3%D9%83%D9%86/%D8%A7%D9%84%D9%84%D9%88%D8%A7%D8%B2%D9%85-%D8%A7%D9%84%D8%A3%D8%B3%D8%A7%D8%B3%D9%8A%D8%A9-%D9%84%D8%AA%D8%AC%D9%87%D9%8A%D8%B2-%D8%B4%D9%82%D8%A9
2. Export as PDF
3. See that the text is written left-to-right

### Expected Behavior

RTL languages should be exported in RTL

### Actual Behavior

RTL languages are exported in LTR

### Additional Information

![Screenshot (862)](https://user-images.githubusercontent.com/16021269/206442583-f38aca8a-472b-46fe-bd40-ab3571910931.png)

timoludwig:
Our [Arabic test PDF](https://github.com/digitalfabrik/integreat-cms/blob/develop/tests/pdf/files/3b02f5ea5b/Integreat%20-%20Arabisch%20-%20%D9%85%D8%B9%D9%84%D9%88%D9%85%D8%A7%D8%AA%20%D8%A7%D9%84%D9%88%D8%B5%D9%88%D9%84.pdf) looks good, so not sure whether this problem only occurs in some special cases.

seluianova:
> Our Arabic test PDF looks good

I haven't dealt with Arabic languages before, so I'm not sure about my findings. But here they are:

Titles in the table content have a different direction, than in the CMS:
![image](https://user-images.githubusercontent.com/115008338/207369833-3fdceaf6-5d36-42ea-89ad-aaaf3de6e66d.png)

Then, on the page itself, the title looks correct to me, but the content has the wrong direction:
![image](https://user-images.githubusercontent.com/115008338/207371308-be08e981-7c7d-40df-8f17-a77d1bd5adf6.png)

[...]

PR #1996, opened 2023-01-12
Fix RTL text direction in PDF export

seluianova:
### Short description
Somehow RTL languages are printed LTR in the PDF export

Title has correct direction, but content is incorrect:
image

### Proposed changes
Text direction is broken when text passes through pdf_strip_fontstyles filter.
Adding "unescape" in return statement solves the issue.

### Side effects
I also updated pdfs in tests for other languages, because tests were failing.
But visually there was no difference except for Arabisch.
pdf_test_result

### Resolved issues
Fixes: #1954


JoeyStk (changes requested):
Thank you so much for this PR. It took me a while to wrap my head around the issue. For Arabic this works for me well now. However for Farsi I still have the issue. 
To me, the part that confused me the most, is that not only the direction of the entire line is mixed up, the direction is also mixed up inside a word. Since in Arabic and Farsi every letter looks different dependent on the position it has within a word, the words don't even look the same in Pdf and the editor.
grafik

seluianova:
@JoeyStk 
Cool finding, thanks. I was looking at Arabisch for so long that my eyes blurred.

I have just reviewed Arabic and Farsi one more time, and here is my summary:
As I see, issues below are actually the same for both Arabic and Farsi.

1. When Arabic/Farsi sentence **contains a Latin word**, weird metamorphoses take place:
I've marked identical pieces of text with the same color. So, we can see that these pieces of text change their position relative to the English word.
What is besonders strange to me - they change not their direction (they still RTL), but position in the sentence.

2. When there is a line break, the newline goes up, not down.
3. Bullet points are on the left, but should be on the right.

image

svenseeberg:
Have you tried setting an `dir="rtl"` attribute for a list item, for example? Would that improve the layout?

seluianova:
As we discussed on the team call:
- the first issue related to mixed content - we will leave it as is
- for second (line breaks) and third (bullet points) and ToC - I will raise issues for xhtml2pdf library

JoeyStk (approved):
For the new scope we're going for, everything seems to work.
I have tested it again and didn't run into issues :)

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ UNRELEASED
 * [ [#1345](https://github.com/digitalfabrik/integreat-cms/issues/1345) ] Track DeepL API usage by regions
 * [ [#1695](https://github.com/digitalfabrik/integreat-cms/issues/1695) ] Treat URLs with broken hash anchors as valid in link checker
 * [ [#1969](https://github.com/digitalfabrik/integreat-cms/issues/1969) ] Hide edit link button in content forms for users without edit permission
+* [ [#1954](https://github.com/digitalfabrik/integreat-cms/issues/1954) ] Fix RTL text direction in PDF export
```
timoludwig, line 8: Due to the long time this PR is open now, this needs to be rebased and then moved up into the unreleased section again.

timoludwig (changes requested):
Thanks a lot! :+1:
I hope that with this fix, the Arabic texts are at least partly readable with some effort...

timoludwig (approved):

PR #1941, opened 2022-12-06
Introduce new status for automatic translations

SteffiStuffel:
### Short description
Added new machine_translated status to translation status.
Added bot icon to show if a translation is machine_translated
Works for SUMM.AI, XLIFF-Imports can be marked with a checkbox as machine translated, DeepL is also implemented, but I couldn't test it, as I don't know how to activate it.
Works for pages, events and pois

### Proposed changes
Removed unnecessary verbose name in AbtractContentTranslations
Moved shared Meta fields of translation forms to customContentModelFrom

### Resolved issues
Fixes: #1883 

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #2010, opened 2023-01-17
Fix changelog and migrations

timoludwig:
### Short description

Two problems occurred during conflict resolution of #1941:
- The changelog entry was inserted at the wrong position
- The previous migrations were renamed and re-ordered, which does not work on production systems which already have them applied. Migrations depend on each other in a specific order to produce a deterministic state.

### Proposed changes

- Move changelog entry
- Reorder migrations to their previous position


MizukiTemma (approved):
👍

JoeyStk (approved):
Now this works also for me! Thank you very much :)

Issue #1752, opened 2022-10-12
Show chapter for internal link suggestions


timoludwig:
### Motivation

When adding an internal link via tinymce and use the suggestion feature to link to existing content, sometimes the titles are ambiguous.

### Proposed Solution

Indicate the whole chapter path of the suggestion here:
![Screenshot 2022-10-12 at 11-11-50 Integreat Redaktionssystem](https://user-images.githubusercontent.com/16021269/195302907-168e58dd-9e69-48e0-8449-ef77036175d0.png)

e.g. similar how we do it for embedding live content:

![Screenshot 2022-10-12 at 11-15-20 Integreat Redaktionssystem](https://user-images.githubusercontent.com/16021269/195303328-0a769667-ecd9-4d44-9a08-a9b987ccd23a.png)

### Additional Context

Maybe the popup needs to be resized to fit longer nested chapter structures

osmers:
Could this be a good first issue? Got a call today from a Kommune who's struggeling with internal links bcs of not having unique names for pages. 
Any way that we could maybe tackle this in the not too distant future?

PR #2016, opened 2023-01-21
Show chapter for internal link suggestions

david-venhoff:
### Short description

This pr changes the internal link suggestions to show the full path to an object translation instead of just its title.
As a sideeffect, the title will be shown in the backend language instead of the object translation's language.
I did not change the general search input feature, because there it is not ambiguous which object is which after performing the search.

### Proposed changes

- Add  a `path` to the JSON objects returned by `search_content_ajax`
- Display the path instead of the title in the internal link suggestions

### Side effects

/

### Resolved issues

Fixes: #1752

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ UNRELEASED
 * [ [#2008](https://github.com/digitalfabrik/integreat-cms/issues/2008) ] Fix end date of recurring events in API
 * [ [#1954](https://github.com/digitalfabrik/integreat-cms/issues/1954) ] Fix RTL text direction in PDF export
 * [ [#1883](https://github.com/digitalfabrik/integreat-cms/issues/1883) ] Add icon and color to POI Category
+* [ [#1752](https://github.com/digitalfabrik/integreat-cms/issues/1752) ] Show chapter for internal link suggestions
 
 
 2023.1.1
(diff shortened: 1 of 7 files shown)


charludo (dismissed):
This works really nice, except for the results in the wrong language. It's a bit unintuitive to search (e.g.) for "Willko" and receive "Welcome" or "Welcome -> ..." as a result. As a user I'd be worried that I'm linking to the English version of content even when I don't mean to.

Since `language_slug` is being passed to `search_content_ajax` in any case, would it maybe be possible to return the correct language paths?

JoeyStk (changes requested):
Thank you a lot for this PR
I also thought it was strange that the languages get mixed. Other than that it also worked for me :)

david-venhoff:
Sorry for the delay..
The path should now be displayed in the current language.

ulliholtgrave (approved):
Thanks, Looks good to me and works nicely :+1:

JoeyStk:
Yes, I agree. The issue we had before is fixed now. Thank you a lot!

JoeyStk (approved):

ulliholtgrave:
Hey @david-venhoff Since we are adding your change to today's release, I'll take the liberty of merging your PR. Nevertheless, many many thanks for your work and best regards from the conference in Dortmund! :)

david-venhoff:
Thanks you! :)

PR #2017, opened 2023-01-23
Make Matomo filters compatible to WP URLs

svenseeberg:
### Short description
Make our statistics filters compatible to the old WP URL pattern (w/o slashes behind endpoints). This should not interfere with the Django pattern.

### Proposed changes
- Filter URLs without the ending slash.
- Disable tracking for the single page endpoint. This should have no consequences at all as the end point is not being used (and I did not find any relevant results for it in the Matomo database).

### Side effects
- None.

### Resolved issues
Fixes: None

### DIFF
--- a/integreat_cms/api/v3/pages.py
+++ b/integreat_cms/api/v3/pages.py
@@ -178,7 +178,6 @@ def get_single_page(request, language_slug):
     return page
 
 
-@matomo_tracking
 @json_response
 # pylint: disable=unused-argument
 def single_page(request, region_slug, language_slug):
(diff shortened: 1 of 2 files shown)


timoludwig (approved):
Thanks! :+1:

Do you think it would also make sense to remove this line:
https://github.com/digitalfabrik/integreat-cms/blob/01d2c78a3ceb4625e34e07e8b8cb110065eadf2f/integreat_cms/api/v3/pages.py#L181
to disable matomo tracking of the single page endpoint?

PR #2028, opened 2023-01-26
Disable linkcheck listeners during tests

timoludwig:
### Short description

The linkcheck listeners do not have access to the database, which leads to very verbose errors during testing.
And since #1977, these errors get now logged to the console.
Since I don't know how to enable access for these asynchronous tasks (I mean there is no place I could add the `@django_db` etc), I thought it might be easier to just disable the listeners during testing.

### Proposed changes

- Disable linkcheck listeners during local tests
- Disable linkcheck listeners on CircleCI

### Side effects

- If we would have proper tests of the linkchecker, this would probably break our tests. But since we don't, it apparently makes no difference to disable the listeners...

### Resolved issues

This fixes a lot of verbose error messages during tests:
```
RuntimeError: Database access not allowed, use the "django_db" mark, or the "db" or "transactional_db" fixtures to enable it.
RuntimeError while running do_check_instance_links with args=(<class 'integreat_cms.cms.models.pages.page_translation.PageTranslation'>, <PageTranslation (id: 233, page_id: 47, language: ar, slug: الترجمة-الشفهية-والتحريرية)>, True) and kwargs={}: Database access not allowed, use the "django_db" mark, or the "db" or "transactional_db" fixtures to enable it.
```

### DIFF
--- a/dev-tools/test.sh
+++ b/dev-tools/test.sh
@@ -16,6 +16,9 @@ require_database
 # Set dummy key to enable SUMM.AI during testing
 export INTEGREAT_CMS_SUMM_AI_API_KEY="dummy"
 
+# Disable linkcheck listeners during testing
+export INTEGREAT_CMS_LINKCHECK_DISABLE_LISTENERS=1
+
 # Parse given command line arguments
 while [[ $# -gt 0 ]]; do
     case $1 in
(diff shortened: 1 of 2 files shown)


PeterNerlich (approved):
This indeed fixes the problem for me. And it seems simple enough that I cannot think of any weird side effects.

ulliholtgrave (approved):

Issue #1965, opened 2022-12-12
Combine recurrent events


steffenkleinle:
### Motivation
Currently the events endpoint and the apps UIs are clustered with multiple occurrences of recurring events because every occurrence is delivered as "own" event. This makes the API slow and the UI unclear and confusing for users. Additionally, this prevents the apps from connecting multiple events, for example to show the next occurrences.

### Proposed Solution
Instead of delivering every occurrence of recurring events as separate event, we could just add a new property `dates` or a similar name that can have an array with the next x (1..n) occurrences of the event as follows:
```javascript
[{
  ...
  "path": "/testumgebung/de/events/recurring-event",
  "dates": [
    { // 1st occurrence
      "start": string, // ISO 8601 date string
      "end": string, // ISO 8601 date string
      "all_day": boolean
    },
    { // 2nd occurrence
      "start": string, // ISO 8601 date string
      "end": string, // ISO 8601 date string
      "all_day": boolean
    },
    ...
  ],
  ...
}]
```

Probably a bump of the api to v4 is necessary to ensure backward compatibility of the whole api. The version of all endpoints should be bumped to make it easier for the frontends. Together with this version bump we could get rid of legacy properties/endpoints that are not used by the frontends anymore (e.g. `start_date`, `end_time`, `timezone`, `event`, `modified_gmt`, `url`...). We should probably discuss this in another call though.

### Alternatives
Deliver only the first occurrence and some kind of frequency, e.g. `weekly`, `daily`, ... While this would be possible, it would probably quite error prone and we would have to implement the logic twice (backend + frontend).  Depends on the possible frequencies event creators can set in the CMS.

### Additional Context

The following App-issue depends on this: https://issues.tuerantuer.org/browse/IGAPP-1177

### Design Requirements
None.

timoludwig:
> ### Alternatives
> 
> Deliver only the first occurrence and some kind of frequency, e.g. `weekly`, `daily`, ... While this would be possible, it would probably quite error prone and we would have to implement the logic twice (backend + frontend). Depends on the possible frequencies event creators can set in the CMS.

I would actually prefer this solution - this would give the app a lot more freedom how to display the recurrence information, how many occurrences to show, etc.
This also doesn't mean we have to implement this logic twice - we only calculate the recurrences for the API, so if we can just pass our internal database fields to the API, we can remove our calculation code which also makes the API faster.
[...]

PR #2029, opened 2023-01-26
Combine recurring events and add rrule

SteffiStuffel:
### Short description
For the new endpoint /events/?combine_recurring=True only one event object is returned. This object contains an rrule, which defines how the element is repeated.

### Proposed changes
Added this new endpoint, we need to keep the old one for backwards compatibility.
If we want to remove the old endpoint in the future (we should wait at least 6 months), i can open a ticket, i think this would be nice to remove the legacy code.

### Resolved issues

Fixes: #1965 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ UNRELEASED
 * [ [#1989](https://github.com/digitalfabrik/integreat-cms/issues/1989) ] Improve load time of locations endpoint
 * [ [#1578](https://github.com/digitalfabrik/integreat-cms/issues/1578) ] Add option to hide files in global media library
 * [ [#1752](https://github.com/digitalfabrik/integreat-cms/issues/1752) ] Show chapter for internal link suggestions
+* [ [#1965](https://github.com/digitalfabrik/integreat-cms/issues/1965) ] Add ical rrule for recurring events
 
 
 2023.2.0
--- a/integreat_cms/api/v3/events.py
+++ b/integreat_cms/api/v3/events.py
@@ -89,6 +89,9 @@ def transform_event_translation(event_translation, recurrence_date=None):
         "location": transform_poi(event.location),
         "event": transform_event(event, recurrence_date),
         "hash": None,
+        "recurrence_rule": event.recurrence_rule.to_ical_rrule_string()
+        if event.recurrence_rule
+        else None,
     }
 
 
@@ -136,17 +139,17 @@ def transform_event_recurrences(event_translation, today):
     :return: An iterator over all future recurrences up to ``settings.API_EVENTS_MAX_TIME_SPAN_DAYS``
     :rtype: Iterator[:class:`~datetime.date`]
     """
+
     event = event_translation.event
     recurrence_rule = event.recurrence_rule
     if not recurrence_rule:
         return
 
-    # In order to avoid unnecessary computations, check if any future event
-    # may be valid and return early if that is not the case
-    if (
+    event_is_invalid = (
         recurrence_rule.recurrence_end_date
         and recurrence_rule.recurrence_end_date < today
-    ):
+    )
+    if event_is_invalid:
         return
 
     start_date = event.start_local.date()
@@ -185,16 +188,17 @@ def events(request, region_slug, language_slug):
     region = request.region
     # Throw a 404 error when the language does not exist or is disabled
     region.get_language_or_404(language_slug, only_active=True)
+
     result = []
     now = timezone.now().date()
     for event in region.events.prefetch_public_translations().filter(archived=False):
-        event_translation = event.get_public_translation(language_slug)
-        if event_translation:
-            if event.end_local.date() >= now:
+        if event_translation := event.get_public_translation(language_slug):
+            # Either it's in the future or it's recurring and the last recurrence is >= now
+            if not event.is_past:
                 result.append(transform_event_translation(event_translation))
-
-            for future_event in transform_event_recurrences(event_translation, now):
-                result.append(future_event)
+            if "combine_recurring" not in request.GET:
+                for future_event in transform_event_recurrences(event_translation, now):
+                    result.append(future_event)
 
     return JsonResponse(
         result, safe=False
--- a/integreat_cms/cms/models/events/recurrence_rule.py
+++ b/integreat_cms/cms/models/events/recurrence_rule.py
@@ -1,9 +1,13 @@
-from datetime import date, timedelta
+import datetime
+from datetime import date, datetime, time, timedelta
+from dateutil import rrule
+
 from django.contrib.postgres.fields import ArrayField
 from django.core.validators import MinValueValidator
 from django.db import models
 from django.utils.translation import gettext
 from django.utils.translation import gettext_lazy as _
+from django.utils.timezone import make_aware
 
 from ..abstract_base_model import AbstractBaseModel
 from ...constants import frequency, weekdays, weeks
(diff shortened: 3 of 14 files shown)


JoeyStk (approved):
Thank you very much. Works as expected! Only note would be to update the changelog file :)

timoludwig (changes requested):
Thanks a lot, very cool feature! :tada:

In general, it almost works as expected :+1:

There is just one edge case when the initial occurrence of an recurrent event is already in the past, then we would completely hide this event in the app if we omit the future occurrences in the API.

Also, I had the idea to use the dateutil dependency to generate the rrule strings, even if it's not too complicate to assemble the strings manually.

integreat_cms/api/v3/events.py
```diff
@@ -89,6 +89,9 @@ def transform_event_translation(event_translation, recurrence_date=None):
         "location": transform_poi(event.location),
         "event": transform_event(event, recurrence_date),
         "hash": None,
+        "recurrence_rule": event.recurrence_rule.to_ical_rrule_string()
+        if event.recurrence_rule
+        else None,
```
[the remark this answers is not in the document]
SteffiStuffel: I don't think so, event holds the information for one specific event, but recurrence_rule is shared between all events, like e.g. location.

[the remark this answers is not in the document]
timoludwig: Ah yes makes sense :+1:

SteffiStuffel:
> 
> There is just one edge case when the initial occurrence of an recurrent event is already in the past, then we would completely hide this event in the app if we omit the future occurrences in the API.
> 

Fix this and added a test for this case.

tests/cms/models/events/test_recurrence_rule.py
```diff
@@ -0,0 +1,108 @@
...
+        )
+
+    def test_api_rrule_bimonthly_until(self):
+        recurrence_rule = RecurrenceRule(
+            frequency="MONTHLY",
+            interval=2,
+            weekdays_for_weekly=None,
+            weekday_for_monthly=6,
+            week_for_monthly=1,
+            recurrence_end_date=datetime.datetime(2030, 10, 19),
```
timoludwig, line 92:
```suggestion
            recurrence_end_date=datetime.date(2030, 10, 19),
```
Since the database field contains a `date` and not a `datetime`, it's maybe better to use this in the test as well (though it doesn't seem to break anything in this case since `datetime.combine()` apparently also works when given a `datetime`)...

tests/cms/models/events/test_recurrence_rule.py
```diff
@@ -0,0 +1,108 @@
...
+
+from integreat_cms.cms.models import RecurrenceRule, Event
+
+
+# pylint: disable=missing-docstring
+
+
+class TestCreatingIcalRule:
+    """
+    Test whether to_ical_rrule_string function is calculating the rrule corretly
```
timoludwig, line 16:
```suggestion
    Test whether to_ical_rrule_string function is calculating the rrule correctly
```

tests/cms/models/events/test_recurrence_rule.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+
+    test_event = Event(
+        start=datetime.datetime(2030, 1, 1, 11, 30, 0, 0, pytz.UTC),
+        end=datetime.datetime(2030, 1, 1, 12, 30, 0, 0, pytz.UTC),
+    )
+
+    def test_api_rrule_every_year_start_date_in_the_past(self):
```
timoludwig, line 24: > > There is just one edge case when the initial occurrence of an recurrent event is already in the past, then we would completely hide this event in the app if we omit the future occurrences in the API.
>
> Fix this and added a test for this case.

Actually, this test case doesn't really cover the previous bug (since the real API code is not executed here)... if we really wanted to make sure this isn't reintroduced in the future, we would have to add a past event to the test data and adapt all the expected API responses again... I'll leave this up to you if you're still motivated :sweat_smile:

tests/api/expected-outputs/augsburg_ar_events.json
```diff
@@ -315,6 +320,7 @@
       "recurrence_id": 1,
       "timezone": "Europe/Berlin"
     },
-    "hash": null
+    "hash": null,
+    "recurrence_rule": "DTSTART:20300101T120000\nRRULE:FREQ=WEEKLY;INTERVAL=2;UNTIL=20310101T235959;BYDAY=TU,TH"
   }
-]
+]
```
timoludwig, line 326:
```suggestion
]

```
I'd prefer newlines at the end of files

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ UNRELEASED
 
 * [ [#2037](https://github.com/digitalfabrik/integreat-cms/issues/2037) ] Add organization and barrier-free fields to POI
 * [ [#2064](https://github.com/digitalfabrik/integreat-cms/issues/2064) ] Show Linkcheck status codes to all users
+* [ [#1965](https://github.com/digitalfabrik/integreat-cms/issues/1965) ] Add ical rrule for recurring events
```
timoludwig, line 6: Sorry, released a new version yet again :see_no_evil:

timoludwig (changes requested):
Thanks a lot for your changes! :muscle: 

I just have a few more little nitpicky comments, then this is ready to be merged :tada:

integreat_cms/api/v3/events.py
```diff
@@ -136,17 +139,17 @@ def transform_event_recurrences(event_translation, today):
...
 
-    # In order to avoid unnecessary computations, check if any future event
-    # may be valid and return early if that is not the case
-    if (
+    event_is_invalid = (
         recurrence_rule.recurrence_end_date
         and recurrence_rule.recurrence_end_date < today
-    ):
+    )
+    if event_is_invalid:
```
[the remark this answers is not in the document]
SteffiStuffel: Leads to unused variable linting error. Will keep it as it is.

integreat_cms/api/v3/events.py
```diff
@@ -185,16 +188,20 @@ def events(request, region_slug, language_slug):
...
     result = []
     now = timezone.now().date()
     for event in region.events.prefetch_public_translations().filter(archived=False):
         event_translation = event.get_public_translation(language_slug)
         if event_translation:
-            if event.end_local.date() >= now:
+            if event.end_local.date() >= now or (
+                event.is_recurring
+                and event.recurrence_rule.to_ical_rrule().after(timezone.now(), True)
+            ):  # or is recurring and last recurring is >= now
```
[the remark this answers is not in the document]
SteffiStuffel: I added this comment, but in general i don't think these types of comments are helpful or are any good when thinking about clean code principles.

integreat_cms/cms/models/events/recurrence_rule.py
```diff
@@ -201,6 +205,44 @@ def get_repr(self):
         """
         return f"<RecurrenceRule (id: {self.id}, event: {self.event.best_translation.slug})>"
 
+    def to_ical_rrule(self):
+        """
+        Calculates the ical standardized rrule for a recurring rule. See details of the rrule here:
+        https://icalendar.org/iCalendar-RFC-5545/3-8-5-3-recurrence-rule.html
+
+        :return: The ical rrule for the recurrence rule
+        :rtype: rrule
```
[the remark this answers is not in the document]
SteffiStuffel: Nice, thank you :rocket:

tests/cms/models/events/test_recurrence_rule.py
SteffiStuffel: You are right, actually this only shows me, there is too much logic in the endpoints, that should be in the models.

timoludwig: Yes, agreed :grin:

integreat_cms/api/v3/events.py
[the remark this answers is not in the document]
timoludwig: Ok, for me it would also be fine to remove the comment.
But the previous version with the hanging inline comment after a multi-line if-statement looked less readable to me, this is why I suggested to move the comment into its own line.
Maybe it would be even better to create a new property to the event model which explains what this if-condition is doing? E.g. something along the lines of:

```
@cached_property
def is_past(self):
    """
    This property checks whether an event lies in the past, including potential future recurrences.

    :return: Whether event lies in the past
    :rtype: bool
    """
    now = timezone.now()
    duration = event.end_local - event.start_local
    future_recurrence = (
        self.is_recurring
        and self.recurrence_rule.to_ical_rrule().after(now - duration, True)
    )
    return self.end_local.date() < now.date() and not future_recurrence
```

then, we could use the simpler and more readable property in the API:

```
if not event.is_past:
    result.append(transform_event_translation(event_translation))
```

I also noticed that the `rrule.after()` method only takes the start datetime into account, so in case the last occurrence is currently ongoing, the condition `after(now)` would not return the current instance. So I suggest to calculate the potential start date of the currently ongoing occurrence by subtracting the duration of the event from the current time.
This still has the caveat of an inconsistency (non-recurring events are still shown on their last day even after they're over, and recurring events disappear as soon as the end time has been reached), but I'm not sure which of both behaviors is the desired one?

timoludwig (approved):
Awesome, thanks! :rocket:

Issue #2031, opened 2023-01-29
Respect settings of machine translation management section


timoludwig:
### Motivation

At the moment, the changes introduced in #1831 / #1994 and #1927 / #2036 are not respected by the machine translation options in the CMS.

### Proposed Solution

Before showing the bulk action to machine translate evens/pois via DeepL or pages via SUMM.AI, check whether the current user is allowed to according to the values of these region fields:
- `machine_translate_pages`
- `machine_translate_events`
- `machine_translate_pois`

Also, please enforce these permissions in bulk action views (in case a malicious user submits this bulk action anyways, even if it's not part of the UI). Potentially, this can be done by adding a rule-based permission defined in [rules.py](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/rules.py).

charludo:
In addition to the proposed solution, it should also be checked whether `machine_translation_enabled` is `True` for the target language(s).

PR #2036, opened 2023-01-30
Add possibility to disable specific languages for machine translations

charludo:
### Short description

Add the possibility to disable machine translations into specific languages from the machine translation management section

### Proposed changes

- add field `machine_translation_enabled` to the `LanguageTreeNode`
- add the field to the `LanguageTreeNodeForm`
- add a list of checkboxes to the MT management section to quickly make a selection for all active languages of the region

### Side effects

- none I'm aware of

### Resolved issues

Fixes: #1927 


charludo:
I wasn't entirely sure how to handle languages without a MT provider, even after the discussion in #1927.
The current behaviour is now as follows:
- by default, MT is enabled for all languages, both new and old
- MT can be enabled/disabled for a language regardless of whether or not a provider is available in the `LanguageTreeNodeForm`
- in the MT management section, languages without a MT provider are unchecked and disabled (I found leaving them checked and disabled a lot more confusing)
- however, saving the `TranslationsManagementForm` will not affect those languages without a provider, i.e. when a provider eventually becomes available, `machine_translation_enabled` is `True` by default, unless it has been changed in the `LanguageTreeNodeForm`

JoeyStk (approved):
Thank you a lot! I'm very impressed by this. The only thing I noticed - and I'm not sure if this really is an issue or not. I was wondering if it makes sense to highlight "Deutsch (DeepL)" as source language / default language to make sure, that translations to German aren't possible / wouldn't make sense. But I guess we can also skip it :thinking:

charludo:
> I was wondering if it makes sense to highlight "Deutsch (DeepL)" as source language / default language to make sure, that translations to German aren't possible / wouldn't make sense.

Thanks for the review!

Yeah, I also wasn't sure about this. I guess the way everything is currently handled, it's impossible to translate into the root of the language node tree.

integreat_cms/cms/forms/translations/translations_management_form.py
```diff
@@ -26,3 +76,52 @@ class Meta:
...
+        :param \**kwargs: The supplied keyword arguments
+        :type \**kwargs: dict
+        """
+        super().__init__(*args, **kwargs)
+
+        self.fields[
+            "machine_translatable_languages"
+        ].queryset = self.instance.language_tree_nodes.filter(active=True).order_by(
+            "language"
+        )
```
timoludwig, line 96: I think it would be cool to group the languages by MT provider instead of ordering them by the language:

Screenshot 2023-02-07 at 13-15-03 Integreat Editorial System

In order to do so, we need to explicitly set the choices of the field in addition to the queryset:

```suggestion
        # Exclude inactive languages and the root node
        languages = self.instance.language_tree_nodes.filter(
            active=True, parent__isnull=False
        )
        mt_providers = defaultdict(list)
        self.unavailable_languages = []
        for language in languages:
            if language.mt_provider:
                mt_providers[language.mt_provider].append(
                    (language.id, language.translated_name)
                )
            else:
                self.unavailable_languages.append(language.translated_name)
        self.fields["machine_translatable_languages"].choices = mt_providers.items()
        self.fields["machine_translatable_languages"].queryset = languages.filter(
            id__in=[id for languages in mt_providers.values() for id, _ in languages]
        )
```

integreat_cms/cms/forms/translations/translations_management_form.py
```diff
@@ -1,17 +1,67 @@
...
+
+        option = super().create_option(name, value, *args, **kwargs)
+        mt_provider = value.instance.mt_provider
+        if mt_provider:
+            option["label"] += f" ({mt_provider})"
+        else:
+            option["attrs"]["disabled"] = ""
+            option["attrs"]["checked"] = False
+            option["attrs"]["class"] = "bg-gray-100 border-gray-300"
+        return option
```
timoludwig, line 50:
```suggestion
    #: The template to use when rendering this widget
    template_name = "translations/languages_input.html"
```
If we group the languages by MT provider, we don't need to overwrite the option dict.
We can however overwrite the template to improve the aesthetics of the input fields.
My suggestion would be to create a file `translations/languages_input.html` with the following content:
```
<ul id="{{ widget.attrs.id }}" class="divide-y divide-gray-200">
    {% for group, options, index in widget.optgroups %}
        <li>
            <label class="text-lg">{{ group }}</label>
            <ul id="{{ widget.attrs.id }}_{{ index }}">
                {% for option in options %}
                    <li>{% include option.template_name with widget=option %}</li>
                {% endfor %}
            </ul>
        </li>
    {% endfor %}
</ul>
```

integreat_cms/cms/forms/translations/translations_management_form.py
```diff
@@ -1,17 +1,67 @@
 import logging
```
timoludwig, line 2:
```suggestion
from collections import defaultdict

```

integreat_cms/cms/forms/translations/translations_management_form.py
```diff
@@ -1,17 +1,67 @@
...
+
 class TranslationsManagementForm(CustomModelForm):
     """
     Form for modifying machine translation settings of a region
     """
 
+    machine_translatable_languages = forms.ModelMultipleChoiceField(
+        widget=CustomCheckboxSelectMultiple(),
+        queryset=None,
+        required=False,
```
timoludwig, line 61:
```suggestion
        required=False,
        label=_("Language Selection"),
```
To also translate the field name for error messages, I suggest to set an explicit label here.

integreat_cms/cms/templates/translations/translations_management.html
```diff
@@ -34,6 +34,20 @@ <h3 class="heading font-bold text-black">
                     </div>
                 </div>
             </div>
+            <div class="rounded border border-solid border-blue-500 shadow-2xl bg-white">
+                <div class="p-4 rounded bg-water-500">
+                    <h3 class="heading font-bold text-black">
+                        <i icon-name="flag" class="mr-2"></i> {% translate "Language Selection" %}
```
timoludwig, line 40:
```suggestion
                        <i icon-name="flag" class="mr-2"></i> {{ form.machine_translatable_languages.label }}
```

integreat_cms/cms/forms/translations/translations_management_form.py
```diff
@@ -1,17 +1,67 @@
...
+        help_text=_("Allow machine translations into the following languages:"),
+    )
+
```
timoludwig, line 64:
```suggestion

    #: The languages that cannot be selected at the moment
    unavailable_languages = []

```
We could define a class attribute to keep track of the unavailable languages

integreat_cms/cms/forms/translations/translations_management_form.py
```diff
@@ -26,3 +76,52 @@ class Meta:
...
+        # Since cleaned_data only contains checked checkboxes, iterate over all region languages
+        for language in self.instance.language_tree_nodes.filter(active=True):
+            if language in self.cleaned_data["machine_translatable_languages"]:
+                language.machine_translation_enabled = True
+                language.save()
+            elif language.mt_provider:
+                # do not alter languages without a MT provider, i.e. those that
+                # were unchecked and disabled in CustomCheckboxSelectMultiple
+                language.machine_translation_enabled = False
+                language.save()
```
timoludwig, line 125:
```suggestion
        # Iterate over all machine translatable languages and toggle them based on the checkbox value
        for language in self.fields["machine_translatable_languages"].queryset:
            language.machine_translation_enabled = (
                language in self.cleaned_data["machine_translatable_languages"]
            )
            language.save()
```
If we correctly limit the queryset to all languages which are possible, we can let Django handle the validation and only need to care about saving here.

integreat_cms/cms/models/languages/language_tree_node.py
```diff
@@ -93,6 +100,25 @@ def text_direction(self):
         """
         return self.language.text_direction
 
+    @cached_property
+    def mt_provider(self):
+        """
+        Return the name of the machine translation provider it it exists,
```
timoludwig, line 106:
```suggestion
        Return the name of the machine translation provider if it exists,
```

integreat_cms/cms/models/languages/language_tree_node.py
```diff
@@ -93,6 +100,25 @@ def text_direction(self):
...
+        or empty string otherwise
+
+        :return: Name of the MT provider for the language
+        :rtype: str
+        """
+        deepl_config = apps.get_app_config("deepl_api")
+        if (
+            self.slug == settings.SUMM_AI_EASY_GERMAN_LANGUAGE_SLUG
+            and settings.SUMM_AI_ENABLED
```
timoludwig, line 115:
```suggestion
            and settings.SUMM_AI_ENABLED
            and self.region.summ_ai_enabled
```
We also need to check here whether SUMM.AI is enabled in the current region

integreat_cms/cms/templates/translations/translations_management.html
```diff
@@ -34,6 +34,20 @@ <h3 class="heading font-bold text-black">
...
+                    </h3>
+                </div>
+                <div class="px-4 pb-4 divide-y divide-gray-200 space-y-4">
+                    <div>
+                        <!-- Options related to which languages can be machine translated -->
+                        <div class="help-text pt-4">{{ form.machine_translatable_languages.help_text }}</div>
+                        {% render_field form.machine_translatable_languages %}
+                    </div>
```
timoludwig, line 48:
```suggestion
                    <!-- Select which languages can be machine translated -->
                    {% if form.machine_translatable_languages.field.queryset.exists %}
                        <div>
                            <div class="help-text mt-4">{{ form.machine_translatable_languages.help_text }}</div>
                            {% render_field form.machine_translatable_languages %}
                        </div>
                    {% endif %}
                    {% if form.unavailable_languages %}
                        <div>
                            <div class="help-text mt-4">{% translate "These languages are currently unavailable for machine translations" %}:</div>
                            <p>{{ form.unavailable_languages|join:", " }}</p>
                        </div>
                    {% endif %}
```

timoludwig (changes requested):
Thanks a lot, very cool feature! :tada:

I just have a few ideas how we could improve the looks of the form a bit.

osmers:
@JoeyStk and I just tested the feature and test-translated a location. We noticed that even though we disabled the language, the bulk action could still be carried out. I think this should not be possible. What do you think?

timoludwig:
@osmers We created a separate issue for this: https://github.com/digitalfabrik/integreat-cms/issues/2031

In this PR, we only add the UI for the settings.

osmers:
Ah perfect - didn't know that :)

charludo:
> Thanks a lot, very cool feature! tada
> 
> I just have a few ideas how we could improve the looks of the form a bit.

Thanks for the review! Agreed, your suggestion looks a lot cleaner.

sphinx/changelog.rst
```diff
@@ -0,0 +1,602 @@
+Changelog
```
timoludwig, line 1: This file results from a failed documentation build, could you remove it?
If you want to, you can also directly add it to `.gitignore` to prevent similar problems in the future.

timoludwig (changes requested):
Thanks a lot! Just one tiny detail missing, then this is ready to merge :tada:

integreat_cms/cms/migrations/0060_alter_languagetreenode.py
```diff
@@ -0,0 +1,25 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    """
+    Add field for en-/dis-abling machine translations to language tree model
+    """
+
+    dependencies = [
+        ("cms", "0059_add_poicategory_icon_color"),
```
timoludwig, line 12: Could you rebase your branch again and rename this file to get a linear migration history?

timoludwig (changes requested):

integreat_cms/cms/migrations/0060_alter_languagetreenode.py
charludo: Oh whoops, of course. Is done.

timoludwig (approved):
:rocket:

Issue #2032, opened 2023-01-30
Gap in sidebar when boxes are collapsed


timoludwig:
### Describe the Bug

When there is only one sidebar and some boxes in the upper part are collapsed, there is a bit of unused space between the two segments of the sidebar (Also see https://github.com/digitalfabrik/integreat-cms/issues/1769#issuecomment-1344607096).

### Steps to Reproduce

1. Go to page form
2. Adapt screen resolution until exactly one sidebar is shown
3. Collapse box in upper part of the sidebar
4. See spacing

### Expected Behavior

The collapsing should cause the lower boxes to move upwards. Maybe we should define a grid that only has one column when there's only space for one sidebar and two columns otherwise, instead of wrapping the two columns around which causes the space?

### Actual Behavior

There is spacing between them

### Additional Information

![Screenshot 2022-12-09 at 19-02-15 Integreat Editorial System](https://user-images.githubusercontent.com/16021269/206765078-120d0068-6624-4b84-bcc1-52a9e57b26e3.png)

PR #2043, opened 2023-01-31
Remove gap between sidebar boxes (for POIs)

JoeyStk:
### Short description
With introducing the collapsible boxes to the page, event and POI form we created a gap between the boxes when certain boxes are collapsed. This is mainly due to two factors: a. using grid instead of flex, b. the two sidebars for medium and 4xl screens aren't wrapped inside one parent div. These changes are going to be relevant for all post types, however in this PR (at least up until now) I focused on POIs only. If we fixed everything for POI I will extend these changes to events and pages one by one.

For a graphical summary on what I tried to fix see the screenshots here: 
Before: 
grafik

After: 
grafik

### Proposed changes

- Using grid only for 3xl screens
- Add parent div for two sidebars (for medium and 4xl screens)
- Remove flex-auto for the editor so it doesn't create blank space
- Adjust spacing between the boxes

### Side effects
While testing this, I didn't run into any. However I can't imagine that there are now, since this is a major change. So I would ask you to test this PR thoroughly :blush: 
Please also check if you can find classes, that aren't used anymore, so I can remove them :) 

### Resolved issues
Fixes: Parts of #2032 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ UNRELEASED
 ----------
 
 * [ [#2056](https://github.com/digitalfabrik/integreat-cms/issues/2056) ] Change restriction of event duration to 28 days
+* [ [#2032](https://github.com/digitalfabrik/integreat-cms/issues/2032) ] Remove gap between sidebar boxes in location form
 
 
 2023.2.0
--- a/integreat_cms/cms/templates/_collapsible_box.html
+++ b/integreat_cms/cms/templates/_collapsible_box.html
@@ -1,5 +1,5 @@
 {# Set collapsed=True when including this template to show the box collapsed in the beginning #}
-<div class="rounded shadow-2xl border border-blue-500 bg-white mb-3">
+<div class="rounded shadow-2xl border border-blue-500 bg-white mb-4">
     <div {% if box_id %}id="{{ box_id }}"{% endif %}
          class="collapsible cursor-pointer rounded p-4 bg-water-500 text-black font-bold flex justify-between"
          {% if collapsed %}data-collapsed{% endif %}>
--- a/integreat_cms/cms/templates/pois/poi_form.html
+++ b/integreat_cms/cms/templates/pois/poi_form.html
@@ -43,10 +43,10 @@
                 </div>
             {% endif %}
         </div>
-        <div class="pt-4 lg:pt-0 md:grid grid-cols-2 3xl:grid-cols-[minmax(0px,_1fr)_400px] 4xl:grid-cols-[minmax(0px,_1fr)_400px_400px] gap-4">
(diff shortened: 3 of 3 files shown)


MizukiTemma (approved):
Thank you for PR :+1 Looks very nice 😃 

Do we need an entry in changelog?

seluianova (changes requested):
No gap now ✅
But there is a regression change:
If I collapse some blocks and re-open form, blocks are reordered randomly.
image
It's not reproduced on develop branch, all blocks always retain their position.
Do we want and can avoid it?

JoeyStk:
Thank you for your reviews!

@MizukiTemma I will add an entry to the changelog once I have extended this to events and pages :)
@selulanova This is expected behaviour, that was part of [this issue](https://github.com/digitalfabrik/integreat-cms/issues/1769). However since I created the PR we have added the option to switch off the automatic distribution of the boxes. I have rebased my branch now, which means the boxes shouldn't be automatically reorganized by default, but you could enable it in the user settings.

seluianova (commented):
Thank you 👌
One more regression issue I have noticed:
If I minimize the window, boxes go down and there is no gap between left and right columns.
image

JoeyStk:
Oh, good catch. I don't know why I didn't see this. The gap-2 is of course not working anymore. I should have thought of that. I think it's fixed now :)

timoludwig (changes requested):
Hmm, somehow I noticed problems on all tested screen resolutions...
In my opinion, the spacing between the boxes and the columns should be consistent. I think, 4rem looks better than 2rem, but the most important thing is that it's consistent.

<details>
  <summary>1700px and below</summary>

Screenshot 2023-02-10 at 13-16-16 Integreat Editorial System

</details>

<details>
  <summary>1700px - 2100px</summary>

Screenshot 2023-02-10 at 13-16-40 Integreat Editorial System


</details>

<details>
  <summary>2100px and above</summary>

Screenshot 2023-02-10 at 13-15-32 Integreat Editorial System


</details>

JoeyStk:
Very good catch. Thank you for the detailed feedback. In some parts this goes beyond what I wanted to do with this PR and in other parts I created this issue during a rebase. I adjusted the spacing for all boxes and gaps to class category 4 (e.g. gap-4 or mb-4, =1rem). I was only able to double check if my changes work for screens smaller than 2100px. For screens above 2100px I can only test it earliest next week

seluianova (approved):
Looks good to me!

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 UNRELEASED
 ----------
 
+* [ [#2032](https://github.com/digitalfabrik/integreat-cms/issues/2032) ] Remove gap between sidebar boxes (for POIs)fields to POI
```
timoludwig, line 4:
```suggestion
* [ [#2032](https://github.com/digitalfabrik/integreat-cms/issues/2032) ] Remove gap between sidebar boxes in location form
```

timoludwig (changes requested):
Thanks! :+1:

Now I cannot find anymore problems, looks good!
I would just reword the changelog entry.

timoludwig (approved):

Issue #959, opened 2021-09-29
Extend POI model/view/template according to cities requirements


ulliholtgrave:
### Motivation

At the moment our POI implementation doesn't support all required fields defined by the cities team. 

### Proposed Solution

Therefore, the following changes to our models, views and templates should be applied:

- [x] #726:
    - Add phone number field (not required)
    - Add E-Mail field (not required)
- [x] #951:
    - Add category field (Advice centre, Authority/Office, Meeting place, Leisure, Shopping, Service, Mobility, Health, Culture, Gastronomy)
- [x] #1832:
    - Add opening hours field (not required)
- [x] #2037:
    - Add organization relation (not required)
    - Add barrier-free field (boolean, not required)

- [ ] Add spoken languages field (design TBD) // re-opened by service-team in 02/2023, see discussion below

Rejected ideas:
- ~~Add target group field (max 50 characters, not required)~~ //removed by service team on 18.02.23)
- ~~Change further important information field (max 50 chars, not required)~~ // removed after a discussion with @osmers in 02/23
- ~~Add "Stops in the vicinity" field~~ //removed by service team on 18.02.23)

### Design Requirements

@hauf-toni

User-Story: I, as user of the Integreat App, would like to see if a POI is multilingual or not

svenseeberg:
This requires coordination with the app team to support the data types and extension of the API definition.

ulliholtgrave:
@timoludwig and @svenseeberg This could become relevant for the Coldaid-Project. @JoeyStk I also contacted the App-Team if this requirements are still up to date

f1sh1918:
for integreat we currently use this design. At the moment we don't need the additional field, but as @ulliholtgrave  mentioned, we have to discuss what we need for cold aid, but i dunno know if you implement different fields in the backend, since it could be a little bit overwhelming for the integreat purpose to have all the coldaid attributes
![image](https://user-images.githubusercontent.com/37902063/163977014-0e8b2e95-836a-4cc6-87d7-b2548fd1eb0a.png)

ulliholtgrave:
I moved it to the next milestone as our current implementation will be productive within this month and we should evaluate how the feature is used.

sarahsporck:
I know for IGAPP-920 we need thumbnail images. Besides, that which fields should be added for coldaid?

ulliholtgrave:
I guess the critical issues are the categories (done), the thumbnail and the opening hours.

[...]

PR #2045, opened 2023-01-31
Add fields to POI form

JoeyStk:
### Short description
This PR adds the following three new sidebar boxes to the POI form.
- Stops in the vicinity
- Spoken languages
- Target group

### Proposed changes

For each one of these fields

- create according model
- add field to the form 
- add field to the API
- add according sidebar box

Maybe @osmers you can have a look at the screenshots below and decide if all requirements for each new field are met.
Stops in the vicinity:
grafik

Spoken languages: 
grafik

Target group:
grafik

### Side effects

- Dependent on how the spoken languages are being displayed inside the app it could be interesting to give the user a dropdown of all languages to help them select a language flag and language name 
- As @MizukiTemma already pointed out inside the issue description also here it could be interesting to add a dropdown button where the user can select from a list of pre-defined target groups

### Resolved issues

Fixes: Parts of #959


timobrembeck (changes requested):
Thanks a lot! :+1:

I think in this case, it would have been better to clarify the existing questions before implementing the solution.

> Dependent on how the spoken languages are being displayed inside the app it could be interesting to give the user a dropdown of all languages to help them select a language flag and language name

Indeed, we already have a language model in our database - it would be stupid not to use it. Free text fields are always prone to error and do not help much when the fields should be used in the apps for filtering, because the same languages might be entered with slightly different spellings. A `forms.ModelMultipleChoiceField` with a `forms.CheckboxSelectMultiple` widget for example? Might need additional thoughts from UI/UX perspective.

> As @MizukiTemma already pointed out inside the issue description also here it could be interesting to add a dropdown button where the user can select from a list of pre-defined target groups

Completely agree, since this field probably is used as filter as well, it would be better to have a fixed set of possible choices here.

Apart from that, I question the free text field for "stops in the vicinity" - can't we use one of our location services to query the nearby stations automatically? Seems a lot of unnecessary effort for region managers to me.

But that's just my opinion, probably this needs further input to answer all open questions.

osmers:
> Apart from that, I question the free text field for "stops in the vicinity" - can't we use one of our location services to query the nearby stations automatically? Seems a lot of unnecessary effort for region managers to me.

I completely agree - especially since they are already complaining that maintaining the POIs is too much work

osmers:
Regarding the other two boxes planned - I think drop downs make a lot of sense - or maybe they can start typing a language and the correct selection comes up. That way they don't need to scroll through the whole list of languages :)

MizukiTemma:
I think it's better to implement a possibility to create and manage target groups and their translations as we did for POI Category and Language.
Then it'll be easier to show and let users choose target groups and they are centrally managed and work well as filter.

dkehne:
It might be useful that these three input-field (or 2 input-field if Haltestellen is generated automatically) are only maintable in expert-mode. As already mentioned it's at the moment already a lot of work to maintain the POIs and also if the argument is that you do not have to maintain the field, the municipalities often start with it, because it is possible and then they regret. 

Espacially because this is not derived from a need, but just a guess that this might be useful, we should not simply make it available for everyone.

JoeyStk:
The requirements have been changed a lot and therefore this PR has become completely obsolete

Issue #1005, opened 2021-10-30
Make it possible to filter for unused media elements


timoludwig:
### Motivation

In the media library, it should be possible to filter for all media elements which are not embedded in any pages/events/pois and are not used as icon of any object.

### Proposed Solution

Evaluate possible approaches for this task, e.g. whether we should store this information somewhere in the database and update it every time a media element is inserted into a page, or whether this should be "calculated" at runtime by scanning the whole content for occurrences of the media file's absolute paths.

ulliholtgrave:
I agree that this should be come in handy, but as for the migration we can focus on only migrating used media files. Therefore, I would leave it on the Version 2.0 milestone.

JanErikHuehne:
Is there any consent on how this should be implemented? Scanning at runtime sounds very slow to me. So should we only perform this one time scan during migration and add fields to media files that get set if they are added / removed somewhere? Maybe a decorator implementation to all functions involving  media files might be a good idea to check it? Otherwise, there will a lot of redundant code I assume.

JanErikHuehne:
I see. I´m on it. Thanks for the input :+1:

PR #2048, opened 2023-02-06
Add filter to show unused media files

MizukiTemma:
### Short description

This PR adds a filter function to show unused media files in the media library and a description in side bar of media files where the file is used.

### Proposed changes

- Add a button for the filter in the media library.
- Use link checker as mentioned in the issue description.
- Get all the media files that are used neither as icon nor embedded item in contents per ajax and show them. 
- Add a new description in the file side bar to tell in which content the file is used.

### Side effects

- None (i hope)

- In region media library currently, only files that belong to the region will be checked and filtered, becuase region users cannot change or delete global files. Should global files too be filtered shown in region media library?

### Resolved issues

Fixes: #1005 


integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -7403,6 +7403,10 @@ msgstr "Datei auswählen"
 msgid "Replace file"
 msgstr "Datei ersetzen"
 
+#: cms/views/media/media_context_mixin.py
+msgid "Show unused media files"
+msgstr "Unbenutzte Mediadateien zeigen"
```
JoeyStk, line 7408: I would prefer "Unbenutzte Mediendateien zeigen"

JoeyStk (changes requested):
Thank you a lot!
In general I think this feature works well. When I use media inside a page, the filter acts correctly and doesn't show the media anymore - this goes both for the regional and the global library. But I still made a few observations: 

- Maybe I'm doing something wrong, but I wasn't able to uncheck the filter and think it would be good, if it would also be possible to reset the filter after using it once. So far it is only possible when I remove the filter-unused-media-files from the URL
- This is probably unrelated to this PR, but when I try to access the media library from the editor the loading time is very long (three times it took forever until I restarted the server). 
- I have one suggestion for a different translation

I think this is it for now :)

MizukiTemma:
@JoeyStk 
> * Maybe I'm doing something wrong, but I wasn't able to uncheck the filter and think it would be good, if it would also be possible to reset the filter after using it once. So far it is only possible when I remove the filter-unused-media-files from the URL

I've introduced an navigation titling "Home > Unused media files" like when a search funcion is being used. so users can go back to the home directory where all the files are shown just in the same way with after key word search.

Screenshot 2023-03-03 151733
Screenshot 2023-03-03 151709



> * This is probably unrelated to this PR, but when I try to access the media library from the editor the loading time was very long (three times it took forever until I restarted the server).

I think, as you meant, this is not because of PR, for I've seen this problem before this PR sometimes. If you think this cannot be ignored anymore, we can have a closer look at it 😃 

> * I have one suggestion for a different translation

Applied!

integreat_cms/cms/models/media/media_file.py
```diff
@@ -202,6 +204,48 @@ def thumbnail_url(self):
...
+        :rtype: str
+        """
+
+        as_icon = (
+            list(self.icon_organizations.all())
+            + list(self.icon_regions.all())
+            + list(self.events.all())
+            + list(self.pages.all())
+            + list(self.pois.all())
+        )
```
timoludwig, line 222: This will produce a LOT of queries (5 queries per media element)...
Some regions have over 1000 media files in their library, which means this will cause 5k database queries and make the library veeeery slow. Please prefetch these reverse relationships for the whole queryset before using them on each individual item.

integreat_cms/cms/models/media/media_file.py
```diff
@@ -202,6 +204,48 @@ def thumbnail_url(self):
...
+
+        as_content = []
+        try:
+            url = Url.objects.get(url=self.url)
+            as_content = list(url.links.all())
+        except Url.DoesNotExist:
+            as_content = []
```
timoludwig, line 229:
```suggestion
        as_content = list(Link.objects.filter(url__url=self.url))
```
You can also go directly via the link model.
But we really have to think about performance here, since it's impossible to prefetch all link objects for the media elements since `url` is a property of the media model and not a database field. Maybe we have to drop this idea :(
Or do you have another idea how to optimize this?

integreat_cms/cms/models/media/media_file.py
```diff
@@ -202,6 +204,48 @@ def thumbnail_url(self):
...
+                + ")"
+            ]
+        for content in as_content:
+            result = result + [
+                str(content.content_object) + " (" + str(content.content_type)[6:] + ")"
+            ]
+
+        result = list(dict.fromkeys(result))
+
+        return ", ".join(result)
```
timoludwig, line 247: I think it would be cool to have the links to the respective content edit forms here.

integreat_cms/cms/views/media/media_actions.py
```diff
@@ -129,6 +131,62 @@ def get_query_search_results_ajax(request, region_slug=None):
...
+    """
+    result = (
+        list(media_file.icon_organizations.all())
+        + list(media_file.icon_regions.all())
+        + list(media_file.events.all())
+        + list(media_file.pages.all())
+        + list(media_file.pois.all())
+    )
+
+    return result
```
timoludwig, line 187: I would move this into the model and use it in the `used_in` property.

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -7449,6 +7453,10 @@ msgstr "Datei auswählen"
 msgid "Replace file"
 msgstr "Datei ersetzen"
 
+#: cms/views/media/media_context_mixin.py
+msgid "Show unused media files"
+msgstr "Unbenutzte Mediendateien zeigen"
```
timoludwig, line 7458:
```suggestion
msgstr "Ungenutzte Mediendateien zeigen"
```

integreat_cms/static/src/js/media-management/component/breadcrumbs.tsx
```diff
@@ -18,6 +18,10 @@ type Props = {
 
 const Breadcrumbs = ({ breadCrumbs, searchQuery, mediaTranslations, allowDrop, dropItem }: Props) => {
     const [currentDragTarget, setCurrentDragTarget] = useState<null | number>(null);
+
+    const actualUrl = window.location.href.split("/");
+    const isUnusedFiltered = actualUrl[actualUrl.length - 2] === "filter-unused-media-files";
```
timoludwig, line 23: I'd rather add a new property here that's passed from `UnusedFilterResultLibrary`.

integreat_cms/static/src/js/media-management/component/breadcrumbs.tsx
```diff
@@ -48,6 +52,14 @@ const Breadcrumbs = ({ breadCrumbs, searchQuery, mediaTranslations, allowDrop, d
...
                     </Link>
                 </li>
+                {isUnusedFiltered && (
+                    <>
+                        <li class="py-2">
+                            <ChevronRight />
+                        </li>
+                        <li class="px-3 py-2">{mediaTranslations.heading_show_unused_files}</li>
+                    </>
+                )}
```
timoludwig, line 62: This returns weird results when in a subdirectory:
Screenshot 2023-03-04 at 14-25-19 Integreat Redaktionssystem

integreat_cms/static/src/js/media-management/unused-filter-library.tsx
```diff
@@ -0,0 +1,37 @@
...
+    // The directory content contains all subdirectories and files of the current directory
+    const [_mediaLibraryContent, setMediaLibraryContent] = mediaLibraryContentState;
+    // The file index contains the index of the file which is currently opened in the sidebar
+    const [_fileIndex, setFileIndex] = fileIndexState;
+    // This state is used to refresh the media library after changes were made
+    const [refresh, _setRefresh] = refreshState;
+
+    useEffect(() => {
+        const urlParams = new URLSearchParams({});
+        console.debug(`Loading search result for query "${searchQuery}"...`);
```
timoludwig, line 27:
```suggestion
        console.debug("Loading unused media files...");
```

integreat_cms/static/src/js/media-management/unused-filter-library.tsx
```diff
@@ -0,0 +1,37 @@
...
+        // Load the search result
+        ajaxRequest(filterUnusedMediaFiles, urlParams, setMediaLibraryContent);
+        // Close the file sidebar
+        setFileIndex(null);
+        /* eslint-disable-next-line react-hooks/exhaustive-deps */
+    }, [searchQuery, refresh]);
```
timoludwig, line 33:
```suggestion
    }, [refresh]);
```

integreat_cms/static/src/js/media-management/library.tsx
```diff
@@ -177,57 +177,77 @@ const Library = ({
...
+                            onClick={async () => {
+                                try {
+                                    route(`/filter-unused-media-files/`);
+                                } catch {
+                                    console.debug("Error");
+                                }
+                            }}>
+                            <Filter className="inline-block mr-2 h-5" />
+                            {mediaTranslations.btn_filter_unused_media_files}
+                        </button>
```
timoludwig, line 249: > Maybe I'm doing something wrong, but I wasn't able to uncheck the filter and think it would be good, if it would also be possible to reset the filter after using it once. So far it is only possible when I remove the filter-unused-media-files from the URL

I agree with this - What about a second button here "Show all media files" which is hidden by default and only shown when the filter is applied? (similarly, the filter button could only be shown when it's not applied)

timoludwig (changes requested):
Thanks a lot! :+1:

Apart from my comments below, I'd also suggest to disable the "Upload" and "Create directory" buttons when the filter is applied, since it might not be clear to which directories the new files are uploaded.

integreat_cms/static/src/js/media-management/component/breadcrumbs.tsx
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable */
+
```
MizukiTemma, line 2: Sorry, I don't know how to solve "Do not nest ternary expressions" Error on the line 54 and disabled eslint temporary for this file...

integreat_cms/static/src/js/media-management/index.tsx
```diff
@@ -89,6 +92,8 @@ const MediaManagement = (props: Props) => {
     const [refresh, setRefresh] = useState<boolean>(false);
     // This state contains a file which should be opened in the sidebar after the content has been refreshed
     const [sidebarFile, setSidebarFile] = useState<File>(null);
+    // whether the filter for unused media files is being activated
+    const [isUnsedFiltered, setFilterState] = useState<boolean>(false);
```
timoludwig: I don't think we need a separate state for this - when we define a new route, the preact router automatically passes the query params to the rendered components.

integreat_cms/cms/models/media/media_file.py
```diff
@@ -115,6 +119,33 @@ def file_size_limit(value):
...
+            )
+        )
+        return self.annotate(is_embedded=Exists(urls)).filter(
+            icon_organizations__isnull=True,
+            icon_regions__isnull=True,
+            events__isnull=True,
+            pages__isnull=True,
+            pois__isnull=True,
+            is_embedded=False,
+        )
```
timoludwig, line 146: This query finds all usages with one sql query - this will probably be much faster than querying all media objects from the database and checking whether they're used in a loop.

integreat_cms/cms/views/media/media_actions.py
```diff
@@ -129,6 +129,60 @@ def get_query_search_results_ajax(request, region_slug=None):
...
+    """
+
+    file = get_object_or_404(
+        MediaFile.objects.filter(
+            Q(region=request.region) | Q(region__isnull=True, is_hidden=False),
+        ),
+        id=request.GET.get("file"),
+    )
+
+    return JsonResponse({"data": file.serialize_usages()})
```
timoludwig, line 156: I did however not find a solution to prefetch all usages in a reasonable amount of time, so I went with another dynamic ajax view which only queries the usages when they are expanded in the sidebar.

integreat_cms/static/src/js/media-management/library.tsx
```diff
@@ -177,59 +180,107 @@ const Library = ({
...
+                    </div>
+                    <div>
+                        {mediaFilter ? (
+                            <button
+                                id="show-all-files-btn"
+                                title={mediaTranslations.btn_reset_filter}
+                                className="btn"
+                                type="submit"
+                                onClick={() => route("/")}>
+                                <FilterX className="inline-block mr-2 h-5" />
```
timoludwig, line 263: I updated lucide icons to make use of the `FilterX` icon...

timoludwig (approved):
Thanks a lot! :+1:

I refactored the structure "a bit", mostly to improve performance and to somehow implement my suggestion of showing the usages in the sidebar (even though I did not manage to find a query that is fast enough to include it on every request, so I added an additional ajax step).

So from my side this is good to go, but could you have a detailed look again at my changes and check whether I made some other mistakes somewhere? Thanks!

integreat_cms/static/src/js/media-management/component/breadcrumbs.tsx
timoludwig: I solved this by introducing another variable `const showFullPath = !(searchQuery || mediaFilter);`

JoeyStk (approved):
Thank you very much for all the changes. It looks a lot better now :smile:

MizukiTemma:
@timoludwig 

I really appreciate your suggestions and detailed explanation 🥺 🙏 it's full of improvements and I find nothing bad 😃

integreat_cms/cms/models/media/media_file.py
```diff
@@ -202,6 +204,69 @@ def thumbnail_url(self):
...
+                    + str(icon.region.slug)
+                    + "/"
+                    + icon._meta.default_related_name
+                    + "/"
+                    + str(icon.region.default_language.slug)
+                    + "/"
+                    + str(icon.id)
+                    + "/"
+                    + "edit"
+                )
```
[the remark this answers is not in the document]
MizukiTemma: Yes, I will 💪

integreat_cms/cms/models/media/media_file.py
MizukiTemma: It's so cool 😻 Thank you!

Issue #1989, opened 2023-01-10
Slow loading of map endpoint


maxammann:
### Describe the Bug
Loading the following endpoint takes 5-6s:

https://cms.integreat-app.de/muenchen/de/wp-json/extensions/v3/locations/?on_map=1

### Steps to Reproduce

1. Click on the above link

### Expected Behavior
Should be lower than 500ms.

### Actual Behavior
Takes 5-6s

### Additional Information
None

timoludwig:
Short term solution could be to enable caching in Apache: https://issues.tuerantuer.org/browse/INFRA-300

PR #2057, opened 2023-02-09
Improve load time of location endpoint

SteffiStuffel:
### Short description
Reduced the load time of this endpoint, by reducing db calls and removing deep copy

### Proposed changes
I used cProfiler to get insights to the loading times and snakeviz to visualize them, getting this output for the location endpoint: 
before

We can see that from 1.70sec total time of this request, 1.078s were spent in deepcopy. Also we see that only 583 calls of the 394670 calls are non-recursive, indication a infinite loop. After removing the deepcopy, we get this result: 
result-after-removing-deepcopy

The total time already decreased to 0.717s and the most time is spent in 102 database calls. These were made due to getting the default translations of objects. Getting the default translations at once, we can reduce the number of database calls to 5, and the total time to 0.075s. 
at-the-end

So at the end with 2 small changes the load time decreased from 1.7s to 0.08s. 
The test-scenario contains about 120 locations.

### Resolved issues

Fixes: #1989


integreat_cms/api/v3/locations.py
```diff
@@ -124,6 +124,9 @@ def locations(request, region_slug, language_slug):
         pois = pois.filter(location_on_map=location_on_map)
     for poi in pois:
         translation = poi.get_public_translation(language_slug)
+        poi.default_translation = poi.get_public_translation(
+            region.default_language.slug
+        )
```
timoludwig, line 129: This changes the behavior of the response (because it now returns the title of the default public translation instead of the newest default translation.
Also, I find it a bit hacky to overwrite a read-only property - I would prefer it if we'd
1. Also prefetch the default translations by adding e.g. `.prefetch_translations()` to the queryset in line 118
2. (preferred) Add a new property `default_public_translation` to the abstract content model which works exactly like `default_translation` with the difference that it caches the default public translation instead of the newest default translation (which doesn't make sense for the API anyway, since we only want to deliver public translations). And then we use `default_public_translation` instead of `default_translation` in the API response.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ UNRELEASED
 * [ [#2008](https://github.com/digitalfabrik/integreat-cms/issues/2008) ] Fix end date of recurring events in API
 * [ [#1954](https://github.com/digitalfabrik/integreat-cms/issues/1954) ] Fix RTL text direction in PDF export
 * [ [#1883](https://github.com/digitalfabrik/integreat-cms/issues/1883) ] Add icon and color to POI Category
+* [ [#1989](https://github.com/digitalfabrik/integreat-cms/issues/1989) ] Improve load time of locations endpoint
```
timoludwig, line 19: This has to be moved to the UNRELEASED section

timoludwig (changes requested):
Thanks, this helps a lot! :+1:

In addition to my comments, I noticed that the poi categories also cause a lot of duplicated queries, I would suggest to prefetch them (including their translations) as well.

integreat_cms/core/settings.py
```diff
@@ -276,6 +276,7 @@
     "integreat_cms.core.middleware.RegionMiddleware",
     "integreat_cms.core.middleware.AccessControlMiddleware",
     "integreat_cms.core.middleware.TimezoneMiddleware",
+    "django_cprofile_middleware.middleware.ProfilerMiddleware",
```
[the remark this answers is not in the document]
SteffiStuffel: ok, removed this, but added a "Debugging" section to the docs.

integreat_cms/cms/models/abstract_content_model.py
```diff
@@ -256,8 +255,8 @@ def get_public_translation(self, language_slug):
                 )
             )
             if public_translation:
-                # Create copy in memory to make sure original translation is not affected by changes
-                public_translation = deepcopy(public_translation)
+                public_translation = copy(public_translation)
```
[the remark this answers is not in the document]
SteffiStuffel: Tested with Redis and could not find any problems yet. Will test again with productive data as soon as i get the dump.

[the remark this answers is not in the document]
timoludwig: I added them to the fixtures directory. The easiest way to import them is probably to edit the [loadtestdata.sh](https://github.com/digitalfabrik/integreat-cms/blob/28277a074118b7f6f6307ea6b1a87901a4de9977/dev-tools/loadtestdata.sh#L14) dev tool and to replace `test_data.json` with `test_pois.json`.

SteffiStuffel:
> 
> In addition to my comments, I noticed that the poi categories also cause a lot of duplicated queries, I would suggest to prefetch them (including their translations) as well.

I also saw this yesterday and I am struggling with this. I try to pretech POICategoryTranslation of all Pois of a region, but POICategory inherites from AbstractBaseModel and not from AbstractContentModel, therefore I cannot use functions like "prefetch_public_translations" to prefetch them. 
So I thought if might make sense to let POICategoryTranslation inherit from AbstractContentModel, but then it would need to have a "region" and this does not make sense here.
Other thought I have is to add a "prefetch_public_translations" method to either POICategory or AbstractBaseModel. To AbstractBaseModel does not really make sense, as not every class inheriting from AbstractBaseModel needs or has translations. But to move it to POICategory seems to create a lot of overhead and also won't work, as POICategroyTranslations is missing all the translation-related-fields like "status, ...". 
So I decided to add (copy) the prefetch methods to poi_category.py, but the result is not saved, "prefetched_translations" is not available, so always a db call is made. I do not really know how i should call these methods in locations.py. Can you help please?

integreat_cms/api/v3/locations.py
```diff
@@ -115,7 +116,13 @@ def locations(request, region_slug, language_slug):
     # Throw a 404 error when the language does not exist or is disabled
     region.get_language_or_404(language_slug, only_active=True)
     result = []
-    pois = region.pois.prefetch_public_translations().filter(archived=False)
+    pois = (
+        region.pois.prefetch_public_translations()
+        .filter(archived=False)
+        .select_related("category")
+    )
+    # TODO category translations need to be prefetched
```
timoludwig, line 124:
```suggestion
        .prefetch_related("category__translations__language")
    )
```
> So I decided to add (copy) the prefetch methods to poi_category.py, but the result is not saved, "prefetched_translations" is not available, so always a db call is made. I do not really know how i should call these methods in locations.py. Can you help please?

For the other content models, the prefetching is implemented as custom queryset methods:

https://github.com/digitalfabrik/integreat-cms/blob/28277a074118b7f6f6307ea6b1a87901a4de9977/integreat_cms/cms/models/abstract_content_model.py#L39-L48

However, since POI categories are a lot simpler, we can just rely on the inbuilt Django queryset method [`prefetch_related()`](https://docs.djangoproject.com/en/4.1/ref/models/querysets/#prefetch-related).

integreat_cms/cms/models/poi_categories/poi_category.py
```diff
@@ -43,6 +43,50 @@ def name(self):
...
+                .select_related("language")
+                .order_by("language__id")
+                .distinct("language__id")
+                .all()
+            )
+        return {
+            translation.language.slug: translation
+            for translation in prefetched_translations
+        }
+
```
timoludwig, line 78:
```suggestion
```
Fortunately, the POI categories are very simple, so we don't need this complicated caching mechanism (for the other content methods, this is only required because the default prefetching mechanism of Django does not allow to filter the prefetched queryset, so we need to build one prefetched attribute for each filtered queryset combination we want to have for each status etc...)

integreat_cms/cms/models/poi_categories/poi_category.py
```diff
@@ -43,6 +43,50 @@ def name(self):
...
+    @cached_property
+    def prefetched_translations_by_language_slug(self):
+        """
+        This method returns a mapping from language slugs to their public translations of this object
+
+        :return: The object translation in the requested :class:`~integreat_cms.cms.models.languages.language.Language` or
+                 :obj:`None` if no translation exists
+        :rtype: dict
+        """
+        return self.get_prefetched_translations_by_language_slug()
```
timoludwig, line 88:
```suggestion
        return {
            translation.language.slug: translation
            for translation in self.translations.all()
        }
```
Here, we can just use `self.translations.all()`. Django is smart enough that it uses the prefetched translations in case they exist and does a db query otherwise.

integreat_cms/cms/models/poi_categories/poi_category.py
```diff
@@ -54,7 +98,8 @@ def get_translation(self, language_slug):
                  if no translation is saved for the language, the category name of the POICategory
         :rtype: ~integreat_cms.cms.models.poi_categories.poi_category_translation.POICategoryTranslation
         """
-        return self.translations.filter(language__slug=language_slug).first()
+        # return self.translations.filter(language__slug=language_slug).first()
```
timoludwig, line 101:
```suggestion
```

timoludwig (changes requested):
> I also saw this yesterday and I am struggling with this. I try to pretech POICategoryTranslation of all Pois of a region, but POICategory inherites from AbstractBaseModel and not from AbstractContentModel, therefore I cannot use functions like "prefetch_public_translations" to prefetch them. So I thought if might make sense to let POICategoryTranslation inherit from AbstractContentModel, but then it would need to have a "region" and this does not make sense here. Other thought I have is to add a "prefetch_public_translations" method to either POICategory or AbstractBaseModel. To AbstractBaseModel does not really make sense, as not every class inheriting from AbstractBaseModel needs or has translations. But to move it to POICategory seems to create a lot of overhead and also won't work, as POICategroyTranslations is missing all the translation-related-fields like "status, ...".

Yeah, it's a bit unfortunate that our models are similar enough to get these code duplications but still different enough to make it hard to generalize it for all of them. One possibility would be to add another shared abstract base model, but in this case the methods are not completely identical, so probably we wouldn't gain much here. So I suggest to leave it like this for the time being and maybe we can come up with something better in the future.

integreat_cms/cms/models/poi_categories/poi_category.py
SteffiStuffel: Thanks for the explanation :heart:

SteffiStuffel:
Thanks for reviewing. I adjusted the prefetching of category translations.

I started to test with test_poi data, but there are some problems:
- loading the pois works fine, with the difference, that "organization" and "barrier_free" options are not passed, and on develop they are passed but null. 
- When i want to change the category of a poi in the ui, i get an error 
```
Exception Type: ValueError
Exception Value: not enough values to unpack (expected 3, got 1)
Exception Location: /home/steffi/Dokumente/integreat-cms/integreat_cms/cms/utils/translation_utils.py, line 40, in translate_link
```

As this is not ready and needs fixes, no need to review now. I will continue to work on this next week.

timoludwig (approved):
> * loading the pois works fine, with the difference, that "organization" and "barrier_free" options are not passed, and on develop they are passed but null.

Good point, the `barrier_free` field is probably irrelevant for performance, but the organization field is another related field which potentially causes a lot of sub-queries. I added this to the `select_related` query as well.

> * When i want to change the category of a poi in the ui, i get an error
> 
> ```
> Exception Type: ValueError
> Exception Value: not enough values to unpack (expected 3, got 1)
> Exception Location: /home/steffi/Dokumente/integreat-cms/integreat_cms/cms/utils/translation_utils.py, line 40, in translate_link
> ```

This error was caused by a combination of inconsistencies in the test data I provided and aa small bug I fixed in #2069. I revised the test data and hopefully eliminated all problems, so I think this is good to go now :tada:

ulliholtgrave (approved):
Thanks, this looks and feels like a real improvement for me :tada: :rocket:

maxammann:
Awesome work and great investigation!

Issue #2056, opened 2023-02-07
Change restriction of event duration to 28 days


osmers:
### Motivation
Quite a few Kommunen requested that the event duration be extended to more than 7 days (see #1510) again, at the Dialogforum or in the Kommunenumfrage - 

### Proposed Solution
Restrict it to 28 days to avoid super long events but make it possible for Kommunen to e.g. include the Interkulturelle Woche

### Alternatives
Not to restrict it at all and have events that last 1 year or longer like before or keep the current restriction and have Kommunen be unhappy

### Additional Context
From Clara's report on the Kommunenumfrage
![image](https://user-images.githubusercontent.com/64901227/217218492-e095fbbd-fc62-4a8e-b0c5-8dc03ea5bbac.png)

PR #2059, opened 2023-02-10
Change max. event duration to 28 Days

MizukiTemma:
### Short description

This PR changes the maximal event duration to 28 days, which is currently 7 days.

### Proposed changes

- Change the condition of duration check to 28 days.

### Side effects

- I think none.

### Resolved issues

Fixes: #2056 

### DIFF
--- a/integreat_cms/cms/forms/events/event_form.py
+++ b/integreat_cms/cms/forms/events/event_form.py
@@ -4,6 +4,7 @@ from datetime import time, timedelta, datetime
 
 from django import forms
 from django.utils.translation import gettext_lazy as _
+from django.conf import settings
 
 from ...models import Event
 from ..custom_model_form import CustomModelForm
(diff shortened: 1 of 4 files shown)


ulliholtgrave (approved):
LGTM :+1:

integreat_cms/cms/forms/events/event_form.py
```diff
@@ -171,12 +171,12 @@ def clean(self):
...
                             ),
                         )
-            elif cleaned_data["end_date"] - cleaned_data["start_date"] > timedelta(6):
+            elif cleaned_data["end_date"] - cleaned_data["start_date"] > timedelta(27):
                 self.add_error(
                     "end_date",
                     forms.ValidationError(
                         _(
-                            "The maximum duration for events is 7 days. Consider using recurring events if the event is not continuous."
+                            "The maximum duration for events is 28 days. Consider using recurring events if the event is not continuous."
```
timoludwig, line 179: To me, the new maximum duration also sounds a bit random, maybe we should add a new setting for this to be able to change it a bit easier in the future?

MizukiTemma: A nice idea 😃 I added it to `settings`.

We can also implement in such a way that service team can change it by themselvs, but it's an over kill in my opinion 😅

timoludwig: Indeed, no need to store this in the database (yet :laughing:)

integreat_cms/core/settings.py
```diff
@@ -83,6 +83,9 @@
 #: The time span up to which recurrent events should be returned by the api
 API_EVENTS_MAX_TIME_SPAN_DAYS = 31
 
+#: The maximum duration of an event
+MAX_EVENT_DURATION = 28
```
timoludwig, line 87:
```suggestion
MAX_EVENT_DURATION = int(os.environ.get("INTEGREAT_CMS_MAX_EVENT_DURATION", 28))
```
I would make this also configurable via the settings file. Otherwise we would still have to release a new version to change this value.

timoludwig (changes requested):

timoludwig (approved):

PR #2061, opened 2023-02-10
Respect settings of machine translation management section

charludo:
### Short description

In #1831 / #1994 and #1927 / #2036 settings were added to control who is allowed to machine translate which content, and into which languages MTs are allowed. This PR implements the necessary backend logic to ensure only permitted requests are being passed on to SUMM.AI/DeepL.

### Proposed changes

- add a utility function responsible for determining whether or not a request should be permitted
- call this function in (page/event/poi)_list_views to be able to disable the bulk action option
- also call it within the DeepL and SUMM.AI translation functions to prevent malicious circumvention of those rules

### Side effects

- none that I could find

### Resolved issues

Fixes: #2031 

### DIFF
--- a/integreat_cms/cms/templates/events/event_list.html
+++ b/integreat_cms/cms/templates/events/event_list.html
@@ -120,27 +120,23 @@
                 <option data-bulk-action="{% url 'bulk_archive_events' region_slug=request.region.slug language_slug=language.slug %}">
                     {% translate "Archive events" %}
                 </option>
-                {% if DEEPL_ENABLED %}
-                    {% if language == request.region.default_language %}
-                        <option disabled
-                                title="{% translate "You cannot translate into the default language" %}">
-                            {% translate "Create missing translations automatically" %}
+                {% if MT_PERMITTED %}
+                    {% if SUMM_AI_ENABLED %}
+                        <option data-bulk-action="{% url 'auto_translate_easy_german_events' region_slug=request.region.slug language_slug=language.slug %}">
+                            {% translate "Automatically translate this event into Easy German" %}
                         </option>
-                    {% elif DEEPL_AVAILABLE %}
-                        <option data-bulk-action="{% url 'automatic_translation_events' region_slug=request.region.slug language_slug=language.slug %}">
-                            {% translate "Create missing translations automatically" %}
-                        </option>
-                    {% else %}
-                        <option disabled
-                                title="{% translate "This language is not supported by DeepL" %}">
-                            {% translate "Create missing translations automatically" %}
-                        </option>
+                    {% elif DEEPL_ENABLED %}
+                        {% if DEEPL_AVAILABLE %}
+                            <option data-bulk-action="{% url 'automatic_translation_events' region_slug=request.region.slug language_slug=language.slug %}">
+                                {% translate "Create missing translations automatically" %}
+                            </option>
+                        {% else %}
+                            <option disabled
+                                    title="{% translate "This language is not supported by any available machine translation provider." %}">
+                                {% translate "Create missing translations automatically" %}
+                            </option>
+                        {% endif %}
                     {% endif %}
-                {% endif %}
-                {% if SUMM_AI_ENABLED %}
-                    <option data-bulk-action="{% url 'auto_translate_easy_german_events' region_slug=request.region.slug language_slug=language.slug %}">
-                        {% translate "Automatically translate this event into Easy German" %}
-                    </option>
                 {% endif %}
             </select>
             <button form="bulk-action-form" id="bulk-action-execute" class="btn" disabled>{% translate "Execute" %}</button>
--- a/integreat_cms/cms/templates/pois/poi_list.html
+++ b/integreat_cms/cms/templates/pois/poi_list.html
@@ -107,27 +107,23 @@
                 <option data-bulk-action="{% url 'bulk_archive_pois' region_slug=request.region.slug language_slug=language.slug %}">
                     {% translate "Archive locations" %}
                 </option>
-                {% if DEEPL_ENABLED %}
-                    {% if language == request.region.default_language %}
-                        <option disabled
-                                title="{% translate "You cannot translate into the default language" %}">
-                            {% translate "Create missing translations automatically" %}
+                {% if MT_PERMITTED %}
+                    {% if SUMM_AI_ENABLED %}
+                        <option data-bulk-action="{% url 'auto_translate_easy_german_pois' region_slug=request.region.slug language_slug=language.slug %}">
+                            {% translate "Automatically translate this location into Easy German" %}
                         </option>
-                    {% elif DEEPL_AVAILABLE %}
-                        <option data-bulk-action="{% url 'automatic_translation_pois' region_slug=request.region.slug language_slug=language.slug %}">
-                            {% translate "Create missing translations automatically" %}
-                        </option>
-                    {% else %}
-                        <option disabled
-                                title="{% translate "This language is not supported by DeepL" %}">
-                            {% translate "Create missing translations automatically" %}
-                        </option>
+                    {% elif DEEPL_ENABLED %}
+                        {% if DEEPL_AVAILABLE %}
+                            <option data-bulk-action="{% url 'automatic_translation_pois' region_slug=request.region.slug language_slug=language.slug %}">
+                                {% translate "Create missing translations automatically" %}
+                            </option>
+                        {% else %}
+                            <option disabled
+                                    title="{% translate "This language is not supported by any available machine translation provider." %}">
+                                {% translate "Create missing translations automatically" %}
+                            </option>
+                        {% endif %}
                     {% endif %}
-                {% endif %}
-                {% if SUMM_AI_ENABLED %}
-                    <option data-bulk-action="{% url 'auto_translate_easy_german_pois' region_slug=request.region.slug language_slug=language.slug %}">
-                        {% translate "Automatically translate this location into Easy German" %}
-                    </option>
                 {% endif %}
             </select>
             <button form="bulk-action-form" id="bulk-action-execute" class="btn" disabled>{% translate "Execute" %}</button>
--- a/integreat_cms/cms/utils/translation_utils.py
+++ b/integreat_cms/cms/utils/translation_utils.py
@@ -1,10 +1,15 @@
 """
 This module contains helpers for the translation process.
 """
+import logging
 import re
 
 from django.utils.html import format_html, format_html_join
 from django.utils.text import format_lazy
+
+from ..constants import machine_translation_permissions as mt_perms
+
+logger = logging.getLogger(__name__)
 
 
 def gettext_many_lazy(*strings):
@@ -46,3 +51,78 @@ def translate_link(message, attributes):
         link_text,
         after,
     )
+
+
+def mt_is_permitted(region, user, content_type, target_lang_slug):
+    """
+    Checks if a machine translation is permitted, i.e. if for the
+    given region, MT of the given content type is allowed and
+    MT into the target language is enabled for the requesting user.
+
+    :param region: the region attempting to machine translate
+    :type region: ~integreat_cms.cms.models.regions.region.Region
+
+    :param user: the user requesting the translation
+    :type user: ~django.contrib.auth.models.User
+
+    :param content_type: type of content which would be translated
+    :type content_type: ~integreat_cms.cms.models.abstract_content_model.AbstractContentModel
+
+    :param target_lang_slug: slug of the MT target language
+    :type target_lang_slug: str
+
+    :return: if the translation is permitted
+    :rtype: bool
+    """
+    permission_settings = {
+        "events": (region.machine_translate_events, "cms.change_event"),
+        "pages": (region.machine_translate_pages, "cms.change_page"),
+        "pois": (region.machine_translate_pois, "cms.change_poi"),
+    }
+
+    mt_perms_setting, required_perm = permission_settings[content_type]
+
+    if mt_perms_setting == mt_perms.NO_ONE:
+        logger.debug(
+            "Machine translations are disabled for content type %r in %r.",
+            content_type,
+            region,
+        )
+        return False
+
+    mt_perm = "cms.manage_translations"
+    if mt_perms_setting == mt_perms.MANAGERS and not user.has_perm(mt_perm):
+        logger.debug(
+            "Machine translations are only enabled for content type %r in %r for users with the permission %r.",
+            content_type,
+            region,
+            mt_perm,
+        )
+        return False
+
+    if not user.has_perm(required_perm):
+        logger.debug(
+            "Machine translations are only enabled for content type %r in %r for users with the permission %r.",
+            content_type,
+            region,
+            required_perm,
+        )
+        return False
+
+    target_lang = region.language_node_by_slug[target_lang_slug]
+
+    if target_lang.is_root():
+        logger.debug(
+            "Machine translations are disabled for the default language %r in %r.",
+            target_lang,
+            region,
+        )
+        return False
+
+    if not target_lang.machine_translation_enabled:
+        logger.debug(
+            "Machine translations are disabled for %r in %r.", target_lang, region
+        )
+        return False
+
+    return True
--- a/integreat_cms/cms/views/mixins.py
+++ b/integreat_cms/cms/views/mixins.py
@@ -131,9 +131,6 @@ class SummAiContextMixin(ContextMixin):
             settings.SUMM_AI_ENABLED
             and self.request.region.summ_ai_enabled
             and kwargs.get("language_slug")
-            in [
-                settings.SUMM_AI_GERMAN_LANGUAGE_SLUG,
-                settings.SUMM_AI_EASY_GERMAN_LANGUAGE_SLUG,
-            ]
+            == settings.SUMM_AI_EASY_GERMAN_LANGUAGE_SLUG
         )
         return context
--- a/integreat_cms/deepl_api/utils.py
+++ b/integreat_cms/deepl_api/utils.py
@@ -9,6 +9,8 @@ from django.utils.html import strip_tags
 from django.apps import apps
 from django.db import transaction
 
+from ..cms.utils.translation_utils import mt_is_permitted
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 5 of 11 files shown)


MizukiTemma:
One question to the required behaviours: poi and events are allowed to be translated by both DeepL and SAMM.AI (as long as those are activated, users have permission for automatic trasnlation and automatic translation is activated for the content type) but no page translation by DeepL?

MizukiTemma (changes requested):
Thank you for PR 💪 

I have two small suggestion:

1) deactivate (show in gray letters) the SAMM.AI option when the current language is not Easy German even if the conditions (user permission, content type, etc...) are met.
Currently the option is shown normal even when the current language is German (not Easy German). I think this is more consistent with behaviour for DeepL option, which is grayed out when automatic translation is not available for the current language.

2) Expand the message `Machine translations have been disabled for this content type or for this language` with `or you do not have permission to automatic translate this content`, since users lower than Manager fall into this case when automatic translation of the content is allowed only for Managers (and higher).

MizukiTemma:
Open question for all members: Observer users.
Automatic translations are shown in the bulk action normally (not gray out) for observers too, as long as automatic translation is available for all users.
Is it sufficient if requests of automatic translation from observers are rejected followed by redirect to the 403 page as it is now, or should they be disabled?

timoludwig:
> 1. deactivate (show in gray letters) the SUMM.AI option when the current language is not Easy German even if the conditions (user permission, content type, etc...) are met.
> Currently the option is shown normal even when the current language is German (not Easy German). I think this is more consistent with behavior for DeepL option, which is grayed out when automatic translation is not available for the current language.

Hmm, I guess it was specifically requested by the service team that the SUMM.AI option is visible for both German and Easy German. @osmers is this still the desired behavior?

> 2. Expand the message `Machine translations have been disabled for this content type or for this language` with `or you do not have permission to automatic translate this content`, since users lower than Manager fall into this case when automatic translation of the content is allowed only for Managers (and higher).

I'm not even sure if we need this entry in the first place. In my opinion, the UI would be a bit cleaner if the MT option would not be shown at all in the drop-down if they are disabled for either the content type/language or the user role.

> Open question for all members: Observer users. Automatic translations are shown in the bulk action normally (not gray out) for observers too, as long as automatic translation is available for all users. Is it sufficient if requests of automatic translation from observers are rejected followed by redirect to the 403 page as it is now, or should they be disabled?

Yes, definitely agree, this is not solved ideal at the moment. In my opinion, those options should just be hidden from the UI. The observer users don't need to know that there is more functionality that they cannot use. They should just see a simplified UI with everything they _can_ do.

charludo:
> One question to the required behaviours: poi and events are allowed to be translated by both DeepL and SAMM.AI (as long as those are activated, users have permission for automatic trasnlation and automatic translation is activated for the content type) but no page translation by DeepL?

I had the same question, see Daniel's answer in the cms-team channel.

> I'm not even sure if we need this entry in the first place. In my opinion, the UI would be a bit cleaner if the MT option would not be shown at all in the drop-down if they are disabled for either the content type or the user role. Only for the language-specific settings it might make sense to explain why some languages are machine translatable and others aren't...

I wasn't really sure about this either, so opted for sticking to the current implementation of having an explanation (when a language was unavailable an explanation was already shown). I think for users who know they generally have the permission to translate, it's better to offer them an explanation why they can't translate in *this* case (content/language permission/availability issue), but I do agree with this:

> > Open question for all members: Observer users. Automatic translations are shown in the bulk action normally (not gray out) for observers too, as long as automatic translation is available for all users. Is it sufficient if requests of automatic translation from observers are rejected followed by redirect to the 403 page as it is now, or should they be disabled?
> 
> Yes, definitely agree, this is not solved ideal at the moment. In my opinion, those options should just be hidden from the UI. The observer users don't need to know that there is more functionality that they cannot use. They should just see a simplified UI with everything they _can_ do.

Somewhat related: I couldn't find a specific permission that allows users to use MT if the setting in the management section is "Yes" for a content type. Is it just everyone except observers?

osmers:
I am not sure I understand the question completely - so when should we gray out SUMM.AI? Are we talking about the management section or the bulk action?

timoludwig:
> I am not sure I understand the question completely - so when should we gray out SUMM.AI? Are we talking about the management section or the bulk action?

About the bulk action. At the moment, we show the SUMM.AI bulk action for both "normal" German and "easy" German. @MizukiTemma suggested to show the bulk action only when the current language is "easy" German (to be more consistent with the other bulk actions, which always conform to the target language and not the source language). I thought I remembered that the service team wanted the SUMM.AI bulk action to be shown for both variants of the German language to reduce the number of clicks necessary to bulk translate into SUMM.AI, but today I learned that the bulk action is only intended for staff members anyway (see #2065), so I guess this isn't important anymore... (initial bulk translations are very rare, so no need to reduce clicks in that case I guess)...

osmers:
@timoludwig sounds good to me - I'd either way keep it in tone with the other bulk actions :)

MizukiTemma:
> > One question to the required behaviours: poi and events are allowed to be translated by both DeepL and SAMM.AI (as long as those are activated, users have permission for automatic trasnlation and automatic translation is activated for the content type) but no page translation by DeepL?
> 
> I had the same question, see Daniel's answer in the cms-team channel.
Ah, OK. Thank you 🙂 


> > > Open question for all members: Observer users. Automatic translations are shown in the bulk action normally (not gray out) for observers too, as long as automatic translation is available for all users. Is it sufficient if requests of automatic translation from observers are rejected followed by redirect to the 403 page as it is now, or should they be disabled?
> > 
> > 
> > Yes, definitely agree, this is not solved ideal at the moment. In my opinion, those options should just be hidden from the UI. The observer users don't need to know that there is more functionality that they cannot use. They should just see a simplified UI with everything they _can_ do.
> 
> Somewhat related: I couldn't find a specific permission that allows users to use MT if the setting in the management section is "Yes" for a content type. Is it just everyone except observers?
Basically yes, though using `has_perm("cms.change_event")` and `has_perm("cms.change_poi)` and `has_perm("cms.change_page_object", page)` or in the templates `can_change` helps us better probably, since this is more general and can stay functioning even if new changes are introduced in permissions settings.

charludo:
The behavior implemented in this PR, taking into considerations the comments from @MizukiTemma and @timoludwig above as well as what has been discussed on monday, looks as follows:

- in the root language, no MT option is ever shown.
- unprivileged users are not shown any MT options.
    - this includes observers.
    - this includes content/language specific permissions set in the MT management section.
- otherwise: 
    - if the language is German (easy) and SUMM.AI is enabled, show the corresponding option. Do not show the DeepL option.
    - for all other languages:
        - if DeepL supports it, show the DeepL option.
        - if not, still show it, but disable it and hint that the language is not supported by DeepL.

Is this behavior acceptable to everyone?

MizukiTemma (changes requested):
Thank you for updates 💪 
It's almost there! I noticed only one small case.

When SAMM_AI is not enabled but DEEPL is enabled, grayed out DEEPL option is shown in Easy German in event and poi list bulk action I guess this should not be shown even as grayed out option since we do not show DEEPL option in Easy German. If this is a misunderstanding, just dismiss 😉

integreat_cms/cms/templates/pois/poi_list.html
```diff
@@ -106,28 +106,24 @@ <h1 class="heading">{% translate "Locations" %}</h1>
...
-                {% else %}
-                    <option disabled
-                            title="{% translate "This language is not supported by DeepL" %}">
-                        {% translate "Create missing translations automatically" %}
+            {% if MT_PERMITTED and language != request.region.default_language %}
+                {% if SUMM_AI_ENABLED %}
+                    <option data-bulk-action="{% url 'auto_translate_easy_german_pois' region_slug=request.region.slug language_slug=language.slug %}">
+                        {% translate "Automatically translate this location into Easy German" %}
                     </option>
+                {% elif DEEPL_ENABLED %}
```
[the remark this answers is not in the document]
charludo: Thanks for the review!! 🎉

This is kind of intentional, though I agree that it's not optimal. My thought process was that there should probably be the same indication that the language can't be translated into despite it being technically allowed like we have with all other languages.

What do you think about dropping the "DeepL" part from the hint/reason? And instead just putting "Language not supported by any available machine translation provider"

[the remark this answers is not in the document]
MizukiTemma: Sorry, somehow the review comment was sent duplicated 😓 

That sounds a good idea 👍

integreat_cms/deepl_api/utils.py
```diff
@@ -110,6 +112,24 @@ def deepl_translation(self, request, content_objects, language_slug, form_class)
...
+                request.user,
+                type(content_objects.first())._meta.default_related_name,
+                language_slug,
+            ):
+                messages.error(
+                    request,
+                    _(
+                        'Machine translations are disabled for content type "{}" or for language "{}".'
+                    ).format(
+                        type(content_objects.first())._meta.verbose_name.title(),
```
timoludwig, line 127:
```suggestion
            if content_objects and not mt_is_permitted(
                region,
                request.user,
                type(content_objects[0])._meta.default_related_name,
                language_slug,
            ):
                messages.error(
                    request,
                    _(
                        'Machine translations are disabled for content type "{}", language "{}" or the current user.'
                    ).format(
                        type(content_objects[0])._meta.verbose_name.title(),
```
I see three potential problems:
1. When `content_objects` is empty, this will raise an `AttributeError: type object 'NoneType' has no attribute '_meta'`
2. The message is incorrect since the current user could also be the reason for the denied check
3. Calling `content_objects.first()` on a queryset you're going to need afterwards is inefficient, since it causes a separate db query to run. It's better to directly force the evaluation of the whole queryset (also using one query, but this time the result can be used afterwards when the whole queryset is iterated).

integreat_cms/summ_ai_api/summ_ai_api_client.py
```diff
@@ -150,6 +153,23 @@ def translate_queryset(self, queryset):
...
+            self.request.user,
+            type(queryset.first())._meta.default_related_name,
+            settings.SUMM_AI_EASY_GERMAN_LANGUAGE_SLUG,
+        ):
+            messages.error(
+                self.request,
+                _(
+                    'Machine translations are disabled for content type "{}" or for language "{}".'
+                ).format(
+                    type(queryset.first())._meta.verbose_name.title(),
```
timoludwig, line 167:
```suggestion
        if queryset and not mt_is_permitted(
            self.request.region,
            self.request.user,
            type(queryset[0])._meta.default_related_name,
            settings.SUMM_AI_EASY_GERMAN_LANGUAGE_SLUG,
        ):
            messages.error(
                self.request,
                _(
                    'Machine translations are disabled for content type "{}", language "{}" or the current user.'
                ).format(
                    type(queryset[0])._meta.verbose_name.title(),
```
here as well?

sphinx/conf.py
```diff
@@ -182,6 +182,7 @@
     ("py:attr", "linkcheck.models.Link.page_translation"),
     ("py:attr", "linkcheck.models.Link.poi_translations"),
     ("py:attr", "linkcheck.models.Link.poi_translation"),
+    ("py:mod", "sphinxcontrib_django2"),
```
timoludwig, line 185: This is not required anymore since #2103
```suggestion
```

integreat_cms/cms/views/mixins.py
```diff
@@ -132,7 +132,6 @@ def get_context_data(self, **kwargs):
             and self.request.region.summ_ai_enabled
             and kwargs.get("language_slug")
             in [
```
timoludwig, line 134:
```suggestion
            == settings.SUMM_AI_EASY_GERMAN_LANGUAGE_SLUG
```
Maybe we don't need `in` for a single value?

integreat_cms/cms/views/mixins.py
```diff
@@ -132,7 +132,6 @@ def get_context_data(self, **kwargs):
...
-                settings.SUMM_AI_GERMAN_LANGUAGE_SLUG,
                 settings.SUMM_AI_EASY_GERMAN_LANGUAGE_SLUG,
             ]
```
timoludwig, line 136:
```suggestion
```

integreat_cms/cms/utils/translation_utils.py
```diff
@@ -46,3 +49,55 @@ def translate_link(message, attributes):
...
+    :param target_lang_slug: slug of the MT target language
+    :type target_lang_slug: str
+
+    :return: if the translation is permitted
+    :rtype: bool
+    """
+    permission_settings = {
+        "events": (region.machine_translate_events, "cms.change_event"),
+        "pages": (region.machine_translate_pages, "cms.change_poi"),
+        "pois": (region.machine_translate_pois, "cms.change_page_object"),
```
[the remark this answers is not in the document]
timoludwig: Also, maybe we should use `._meta.default_related_name` here as well instead of hardcoding the values?

timoludwig (changes requested):
Thanks a lot! :muscle:

integreat_cms/deepl_api/utils.py
charludo: 1. yep, good point, thanks for fixing this.
2. thanks for the more complete error message (the chances of this being shown because of user permission issues are probably very low for legitimate use cases though)
3. did not know, TIL

integreat_cms/cms/utils/translation_utils.py
```diff
@@ -46,3 +49,55 @@ def translate_link(message, attributes):
...
+        return False
+
+    # else: user is permitted to edit and translate this content;
+    # determine if translation into language is permitted
+    target_lang_allowed = get_object_or_404(
+        region.language_tree_nodes,
+        language__slug=target_lang_slug,
+    ).machine_translation_enabled
+
+    return target_lang_allowed
```
timoludwig, line 103:
```suggestion
    mt_perms_setting, required_perm = permission_settings[content_type]

    if mt_perms_setting == mt_perms.NO_ONE:
        logger.debug(
            "Machine translations are disabled for content type %r in %r.",
            content_type,
            region,
        )
        return False

    mt_perm = "cms.manage_translations"
    if mt_perms_setting == mt_perms.MANAGERS and not user.has_perm(mt_perm):
        logger.debug(
            "Machine translations are only enabled for content type %r in %r for users with the permission %r.",
            content_type,
            region,
            mt_perm,
        )
        return False

    if not user.has_perm(required_perm):
        logger.debug(
            "Machine translations are only enabled for content type %r in %r for users with the permission %r.",
            content_type,
            region,
            required_perm,
        )
        return False

    target_lang = region.language_node_by_slug[target_lang_slug]

    if target_lang == region.default_language:
        logger.debug(
            "Machine translations are disabled for the default language %r in %r.",
            target_lang,
            region,
        )
        return False

    if not target_lang.machine_translation_enabled:
        logger.debug(
            "Machine translations are disabled for %r in %r.", target_lang, region
        )
        return False

    return True
```
During testing, I noticed that the `manage_translations` permission was not correctly checked, because the app label `cms.` was missing...
I took me a while to figure this out, so I suggest to add a bit of debug logging to make it a bit easier to follow the reason why the function returns `False`. This should also make the comment obsolete.

Also, querying the language via `region.language_node_by_slug` instead of `get_object_or_404` prevents an additional query because the language tree nodes are already prefetched for the region model anyways.

Apart from that, I think we could also check whether a language is the default language here? Then we don't have to check it each template...

integreat_cms/cms/utils/translation_utils.py
```diff
@@ -3,9 +3,12 @@
 """
 import re
 
+from django.shortcuts import get_object_or_404
```
timoludwig, line 6:
```suggestion
import logging
import re

```

integreat_cms/cms/utils/translation_utils.py
```diff
@@ -3,9 +3,12 @@
...
 from django.utils.html import format_html, format_html_join
 from django.utils.text import format_lazy
 
+from ..constants import machine_translation_permissions as mtp
```
timoludwig, line 10:
```suggestion
from ..constants import machine_translation_permissions as mt_perms

logger = logging.getLogger(__name__)
```

integreat_cms/cms/templates/events/event_list.html
```diff
@@ -119,28 +119,24 @@ <h1 class="heading">{% translate "Events" %}</h1>
...
-                    </option>
-                {% elif DEEPL_AVAILABLE %}
-                    <option data-bulk-action="{% url 'automatic_translation_events' region_slug=request.region.slug language_slug=language.slug %}">
-                        {% translate "Create missing translations automatically" %}
-                    </option>
-                {% else %}
-                    <option disabled
-                            title="{% translate "This language is not supported by DeepL" %}">
-                        {% translate "Create missing translations automatically" %}
+            {% if MT_PERMITTED and language != request.region.default_language %}
```
timoludwig, line 122:
```suggestion
            {% if MT_PERMITTED %}
```

integreat_cms/cms/templates/pois/poi_list.html
```diff
@@ -106,28 +106,24 @@ <h1 class="heading">{% translate "Locations" %}</h1>
...
-                    </option>
-                {% elif DEEPL_AVAILABLE %}
-                    <option data-bulk-action="{% url 'automatic_translation_pois' region_slug=request.region.slug language_slug=language.slug %}">
-                        {% translate "Create missing translations automatically" %}
-                    </option>
-                {% else %}
-                    <option disabled
-                            title="{% translate "This language is not supported by DeepL" %}">
-                        {% translate "Create missing translations automatically" %}
+            {% if MT_PERMITTED and language != request.region.default_language %}
```
timoludwig, line 109:
```suggestion
            {% if MT_PERMITTED %}
```

timoludwig (changes requested):
Thanks for your changes! :+1:
I still noticed a few problems in `mt_is_permitted`, see my comments below.

integreat_cms/cms/utils/translation_utils.py
charludo: Oh no, sorry to hear that 😅  And thank you for investing the time to add the debug logic already!

Good idea about checking for the default language.

integreat_cms/cms/utils/translation_utils.py
```diff
@@ -46,3 +51,78 @@ def translate_link(message, attributes):
...
+            "Machine translations are only enabled for content type %r in %r for users with the permission %r.",
+            content_type,
+            region,
+            required_perm,
+        )
+        return False
+
+    target_lang = region.language_node_by_slug[target_lang_slug]
+
+    if target_lang == region.default_language:
```
timoludwig, line 114:
```suggestion
    if target_lang.is_root():
```
Ah sorry, I messed up my suggestion. The `target_lang` is a `LanguageTreeNode` object, the `region.default_language` a `Language` object...
We could check `target_lang.language == region.default_language` but I think this is even simpler.

timoludwig (changes requested):
Thanks! Just a small detail (my mistake), then this is ready to go! :tada:

timoludwig (approved):

MizukiTemma (approved):
Thank you so much 🎉

Issue #1875, opened 2022-11-18
Add checkbox `Automatische Übersetzung für veraltete Übersetzung` for automatic translation on page form


ulliholtgrave:
### Motivation

In the daily work of the municipalities the normal workflow includes updating the german content and directly wanting to translate them to other languages. Therefore, our automatic translation workflow should include a possibility to trigger automatic translations directly in the page form for now outdated translation. 

### Proposed Solution

Add another checkbox below/above (TBD) "minor_edit" that is exclusive to "minor_edit". If it is checked and the page gets updated, all translations that would be outdated and are supported by DeepL are automatically translated.

The result would be a manual created german translation and up-to-date translations in other languages.

The translation budget needs to be considered for the translation.

### Additional Context

Related to #1727 and #1663

nikolahoff:
@ulliholtgrave if the budget for translations is not sufficient is it possible
a) to show an info: "Your budget is not sufficient. You want to translate 123 words but you only have credits for 56 words."
b) to disable the "Update" button in that case?

ulliholtgrave:
@nikolahoff Yes, this is possible. 

However, We should consider the case when a user is trying to translate into 5 languages but his budget only allows translations for 3. In this case we would stop the translation process after 3 and inform the user about 2 missing languages due to insufficient funds. Does that sound like a reasonable solution for you?

nikolahoff:
Sounds good. Can we show an info-box in that case?

osmers:
@ulliholtgrave @nikolahoff how do we deal with the possible issue that Kommunen will call us and say that, had they known, they would have chosen different languages to translate the text into? Not sure how probably this is, since the budgets should be large enough. Just wanted to alert to the possibility...maybe we can stop the process before anything is translated, saying that the budget would only be large enough for x languages, and to please go to the settings and choose the ones they want?

osmers:
@ulliholtgrave maybe @dkehne can give his input here bcs he was not a fan of our prioritising - it would be too confusing.

timoludwig:
We're talking about a single page here, right?
I doubt that the difference between one single page being translated into language X but not into language Y makes such a huge difference.
If they really cannot live with a partially translated page, they can also manually copy the content into the DeepL UI to finish the missing languages by hand.
I doubt the programming effort is justified by such an edge case which does not have drastic implications.

osmers:
> I doubt the programming effort is justified by such an edge case which does not have drastic implications.

[...]

Issue #1459, opened 2022-05-21
Simplify the DeepL translation process for events & locations


karthik-integreat:
### Motivation

The procedure right now is as follows:
Publish an event. go to another language, choose the option, translate using deepL for available languages.
It would be easier to have the possibility to do it immediately after creating an event.

### Proposed Solution

Copy the solution for pages from #1875 for Event and POIs

### Additional Context

Depends on:
- [x] #1875

osmers:
Could we just use the same process as for pages? Like you can add a check mark if you wanna translate the event into all available languages? Or the pop-up idea to be able to choose the languages. Though I heard pop-ups are not that popular...

timoludwig:
@osmers I agree, I would put this on hold until #1875 is implemented and then copy the solution from there.

osmers:
@timoludwig just not sure if that works, bcs what if I do not want all languages? The check box works for the pages, bcs we are only translating Aktualisierungsübersetzungen. For events it would almost always be an Initialübersetzung. And then we'd have no  way of choosing the languages, right?

timoludwig:
@osmers what do you think of just creating initial translations for all missing translations which are supported by DeepL?
So in order to "exclude" one language from the initial translation, one would have to
1. Create the event without ticking the checkbox
2. Create all translations which should be done manually and not via DeepL
3. Save the default translation again with the checkbox checked to create all missing translations automatically?

osmers:
Assuming that for some languages they might not want a translation at all, neither machine nor human, is there any way to do that?

osmers:
Bcs I think that if there are languages that they'd want translated by a human, they can also add that (replace the machine translation) afterwards. It is more the case where they do not want any translations for specific languages. Not sure, how relevant it is...the case I could imagine is that it is an event directed at a specific group of people, thus only their language should be translated (plus maybe English?!)

PR #2062, opened 2023-02-10
Add automatic translation checkbox to Pages, Events, POIs

charludo:
### Short description

Adds a checkbox "automatic translation" to Page/Event/POI form views that is exclusive to "minor edit". If checked, the respective content is translated into all languages that are descendants of the source language (if they are supported by DeepL, that is).

### Proposed changes

- added a form field `automatic_translation` to the (new) `MachineTranslationForm`
- when checked, allow selection of translations to update/create
- if checked and minor edit is not checked, pass to DeepL for translation 
- hide this checkbox if DeepL is not enabled or the language has no descendants

### Side effects

- none that I could find, however I'd be glad if whoever reviews this keeps an eye on the translation statuses. AFAIK everything is working as intended though. 

### Additional information
- moved the TS logic for hiding/showing conditional fields into its own, view-independent file, since the logic was already re-usable and I saw no point in duplicating it

### Resolved issues

Fixes: #1875
Fixes: #1459

### DIFF
--- a/integreat_cms/cms/forms/translations/machine_translation_form.py
+++ b/integreat_cms/cms/forms/translations/machine_translation_form.py
@@ -1,0 +1,108 @@
+import logging
+
+from django import forms
+from django.utils.translation import gettext_lazy as _
+
+from ...utils.translation_utils import mt_to_lang_is_permitted
+
+logger = logging.getLogger(__name__)
+
+
+class MachineTranslationForm(forms.Form):
+    """
+    Form for selecting target languages for machine translation of a content object.
+    """
+
+    automatic_translation = forms.BooleanField(
+        widget=forms.CheckboxInput(),
+        required=False,
+        label=_("Automatic translation"),
+        help_text=_(
+            "Tick if updating this content should automatically refresh or create its translations."
+        ),
+    )
+
+    translations_to_update = forms.ModelMultipleChoiceField(
+        widget=forms.CheckboxSelectMultiple(),
+        queryset=None,
+        required=False,
+        label=_("Update existing translations:"),
+    )
+
+    translations_to_create = forms.ModelMultipleChoiceField(
+        widget=forms.CheckboxSelectMultiple(),
+        queryset=None,
+        required=False,
+        label=_("Create new translations:"),
+    )
+
+    def __init__(self, instance, region, language, **kwargs):
+        r"""
+        Initialize MT translation form
+
+        :param \**kwargs: The supplied keyword arguments
+        :type \**kwargs: dict
+        """
+        super().__init__(**kwargs)
+
+        parent_node = region.language_node_by_slug.get(language.slug)
+        translation_targets = [
+            language_node
+            for language_node in region.language_tree
+            if language_node.parent_id
+            and language_node.parent_id == parent_node.id
+            and language_node.mt_provider
+            and mt_to_lang_is_permitted(region, language_node.slug)
+        ]
+
+        # completely hide options if no target languages exist
+        if not translation_targets:
+            del self.fields["automatic_translation"]
+            return
+
+        to_update, to_create = [], []
+        for target in translation_targets:
+            target_type = (
+                to_update
+                if instance and instance.get_translation(target.slug)
+                else to_create
+            )
+            target_type.append(target.id)
+
+        self.fields[
+            "translations_to_update"
+        ].queryset = region.language_tree_nodes.filter(id__in=to_update)
+        self.fields[
+            "translations_to_create"
+        ].queryset = region.language_tree_nodes.filter(id__in=to_create)
+
+        self.initial["translations_to_update"] = to_update
+
+    def clean(self):
+        """
+        Validate form fields which depend on each other, see :meth:`django.forms.Form.clean`
+
+        :return: The cleaned form data
+        :rtype: dict
+        """
+        cleaned_data = super().clean()
+
+        if not cleaned_data["automatic_translation"]:
+            cleaned_data["translations_to_update"] = []
+            cleaned_data["translations_to_create"] = []
+        return cleaned_data
+
+    def get_target_language_slugs(self):
+        """
+        Return the slugs of all selected target languages
+
+        :return: The target language slugs
+        :rtype: list [ str ]
+        """
+        return (
+            self.cleaned_data["translations_to_update"]
+            .union(self.cleaned_data["translations_to_create"])
+            .values_list("language__slug", flat=True)
+            if self.is_valid()
+            else []
+        )
--- a/integreat_cms/cms/templates/events/event_form_sidebar/minor_edit_box.html
+++ b/integreat_cms/cms/templates/events/event_form_sidebar/minor_edit_box.html
@@ -7,14 +7,41 @@
     feather
 {% endblock collapsible_box_icon %}
 {% block collapsible_box_title %}
-    {% translate "Minor edit" %}
+    {% minor_edit_label request.region language %}
 {% endblock collapsible_box_title %}
 {% block collapsible_box_content %}
-    <label class="mt-0">{% minor_edit_label request.region language %}</label>
-    {% render_field event_translation_form.minor_edit %}
-    <label for="{{ event_translation_form.minor_edit.id_for_label }}"
-           class="secondary">
-        {{ event_translation_form.minor_edit.label }}
-    </label>
-    <div class="help-text">{% minor_edit_help_text request.region language event_translation_form %}</div>
+    <div>
+        {% render_field event_translation_form.minor_edit class+="mutually-exclusive-checkbox" %}
+        <label for="{{ event_translation_form.minor_edit.id_for_label }}"
+               class="secondary">
+            {{ event_translation_form.minor_edit.label }}
+        </label>
+        <div class="help-text">{% minor_edit_help_text request.region language event_translation_form %}</div>
+    </div>
+    {% if MT_ENABLED %}
+        <div>
+            {% render_field machine_translation_form.automatic_translation class+="mutually-exclusive-checkbox" %}
+            <label for="{{ machine_translation_form.automatic_translation.id_for_label }}"
+                   class="secondary">
+                {{ machine_translation_form.automatic_translation.label }}
+            </label>
+            <div class="help-text">{{ machine_translation_form.automatic_translation.help_text }}</div>
+            <div id="language-options" class="pl-2 hidden">
+                {% if machine_translation_form.translations_to_update.field.queryset.exists %}
+                    <label for="{{ machine_translation_form.translations_to_update.id_for_label }}"
+                           class="secondary">
+                        {{ machine_translation_form.translations_to_update.label }}
+                    </label>
+                    {% render_field machine_translation_form.translations_to_update %}
+                {% endif %}
+                {% if machine_translation_form.translations_to_create.field.queryset.exists %}
+                    <label for="{{ machine_translation_form.translations_to_create.id_for_label }}"
+                           class="secondary">
+                        {{ machine_translation_form.translations_to_create.label }}
+                    </label>
+                    {% render_field machine_translation_form.translations_to_create %}
+                {% endif %}
+            </div>
+        </div>
+    {% endif %}
 {% endblock collapsible_box_content %}
--- a/integreat_cms/cms/views/events/event_form_view.py
+++ b/integreat_cms/cms/views/events/event_form_view.py
@@ -9,11 +9,17 @@ from django.utils.translation import gettext as _
 from django.views.generic import TemplateView
 
 
+from ....deepl_api.utils import DeepLApi
 from ...constants import status, translation_status
 from ...decorators import permission_required
-from ...forms import EventForm, EventTranslationForm, RecurrenceRuleForm
+from ...forms import (
+    EventForm,
+    EventTranslationForm,
+    RecurrenceRuleForm,
+    MachineTranslationForm,
+)
 from ...models import Language, Event, EventTranslation, RecurrenceRule, POI
-from ...utils.translation_utils import translate_link
+from ...utils.translation_utils import mt_is_permitted, translate_link
 from .event_context_mixin import EventContextMixin
 from ..media.media_context_mixin import MediaContextMixin
 from ..mixins import ContentEditLockMixin
@@ -103,6 +109,18 @@ class EventFormView(
         recurrence_rule_form = RecurrenceRuleForm(
             instance=recurrence_rule_instance, disabled=disabled
         )
+
+        machine_translation_form = MachineTranslationForm(
+            instance=event_instance, region=region, language=language
+        )
+
+        # Check for MT availability for automatic translation
+        MT_ENABLED = (
+            settings.DEEPL_ENABLED
+            and region.language_node_by_slug.get(language.slug).get_descendants()
+            and mt_is_permitted(region, request.user, Event._meta.default_related_name)
+        )
+
         url_link = f"{settings.WEBAPP_URL}/{region.slug}/{language.slug}/{event_translation_form.instance.url_infix}/"
         return render(
             request,
@@ -112,6 +130,7 @@ class EventFormView(
                 "event_form": event_form,
                 "event_translation_form": event_translation_form,
                 "recurrence_rule_form": recurrence_rule_form,
+                "machine_translation_form": machine_translation_form,
                 "poi": event_instance.location if event_instance else None,
                 "language": language,
                 "languages": region.active_languages if event_instance else [language],
@@ -119,10 +138,11 @@ class EventFormView(
                 "translation_states": event_instance.translation_states
                 if event_instance
                 else [],
+                "MT_ENABLED": MT_ENABLED,
             },
         )
 
-    # pylint: disable=too-many-locals,too-many-branches
+    # pylint: disable=too-many-locals,too-many-branches,too-many-statements
     def post(self, request, **kwargs):
         r"""
         Save event and ender event form for HTTP POST requests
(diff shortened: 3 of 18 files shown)


integreat_cms/cms/templates/events/event_form_sidebar/minor_edit_box.html
```diff
@@ -7,14 +7,26 @@
     feather
 {% endblock collapsible_box_icon %}
 {% block collapsible_box_title %}
-    {% translate "Minor edit" %}
+    {% translate "Minor edit and automatic translations" %}
```
JoeyStk, line 10: According to the design (s. screenshot) the title is supposed to be different. In a chat with @osmers last week it was reinforced, that the new title is supposed to be the one from the design. 

grafik
```suggestion
    {% translate "Translations" %}
```

JoeyStk (changes requested):
Thank you very much for this PR! 
All in all this looks good. I'm double checking with @osmers if the different possible states of the radio buttons are as they are required. If so, I will let you know. Other than that I only have one small remark about the title of the box

Update 15/02/23: All possible states work as requested :)

integreat_cms/cms/templates/events/event_form_sidebar/minor_edit_box.html
charludo: Thanks, I was not aware. Where did that discussion happen? Because I couldn't find it via the issue(s)

JoeyStk: Yes, that's unfortunate. The design (and therefore also the discussion) were made in Figma, and not everyone has access to the board. That is definitively not great

JoeyStk: Here is [the link](https://www.figma.com/file/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=166%3A1980) for Figma, in case you don't already have it

charludo: Thanks for the link, I did not have it yet, I think.
Just so I'm clear (maybe @osmers can help out here), only already existing translations should be updated? New ones should not be created, even if the language is available through DeepL?

osmers: @charludo correct - we will use the management section for creating new translations

osmers: @charludo just discussed the title of the box as well - could you change it from "Übersetzungen" to "Auswirkungen auf andere Übersetzungen" and delete the subtitle (the one that is currently "Auswirkungen auf andere Übersetzungen")?

charludo: > the one that is currently "Auswirkungen auf andere Übersetzungen"

AFAIK this changes dependent on if the language that is currently being edited is the root of the language tree ("Auswirkungen auf andere Übersetzungen"), a leaf ("Auswirkungen auf den Übersetzungsstatus"), or has both parent and child elements ("Auswirkungen auf den Status dieser und anderer Übersetzungen").

Do you want me to move this behaviour into the title of the box, or actually remove this feature and keep it to a static "Auswirkungen auf andere Übersetzungen"?

charludo: > correct - we will use the management section for creating new translations

In that case, shouldn't there (eventually, not necessarily in this PR) be a second checkbox that translates to all child languages, at least for new pages/events/pois? I would imagine it's a lot more convenient to just check a box and hit "Save" as opposed to having to split this into two steps on different forms

osmers: > > the one that is currently "Auswirkungen auf andere Übersetzungen"
> 
> AFAIK this changes dependent on if the language that is currently being edited is the root of the language tree ("Auswirkungen auf andere Übersetzungen"), a leaf ("Auswirkungen auf den Übersetzungsstatus"), or has both parent and child elements ("Auswirkungen auf den Status dieser und anderer Übersetzungen").
> 
> Do you want me to move this behaviour into the title of the box, or actually remove this feature and keep it to a static "Auswirkungen auf andere Übersetzungen"?

Could we maybe have a short call for this bcs I'm not sure I completely understand :D

osmers: > > correct - we will use the management section for creating new translations
> 
> In that case, shouldn't there (eventually, not necessarily in this PR) be a second checkbox that translates to all child languages, at least for new pages/events/pois? I would imagine it's a lot more convenient to just check a box and hit "Save" as opposed to having to split this into two steps on different forms

We have actually discussed this a lot during the conference and changed a lot regarding what was planned. So we have three cases in which we need a translation and at least one of them requires a bulk action which will now be added in the page tree - but your suggestions is something we might think of later in case Kommunen complain about the bulk action for single pages. The only "problem" would be that we'd need a possibility to choose languages for new translations, which might be difficult/work with a pop up, if we incorporate it directly into the page form.

charludo: @osmers thanks for letting me know about the changed plans, yes, I think a call would be great. I'll message you on Mattermost.

integreat_cms/deepl_api/utils.py
```diff
@@ -232,3 +249,41 @@ def deepl_translation(self, request, content_objects, language_slug, form_class)
...
+
+        :param request: passed request
+        :type request: ~django.http.HttpRequest
+
+        :param source_object: passed content object
+        :type source_object: ~integreat_cms.cms.models.abstract_content_model.AbstractContentModel
+
+        :param source_language_slug: slug of the source language from which to translate
+        :type source_language_slug: str
+
```
timoludwig, line 274:
```suggestion
```

integreat_cms/deepl_api/utils.py
```diff
@@ -232,3 +249,41 @@ def deepl_translation(self, request, content_objects, language_slug, form_class)
...
+        :type form_class: ~integreat_cms.cms.forms.custom_content_model_form.CustomContentModelForm
+        """
+        for language_slug in target_language_slugs:
+            if self.check_availability(request, language_slug):
+                self.deepl_translation(
+                    request,
+                    [source_object],
+                    language_slug,
+                    form_class,
+                    source_language_slug,
```
timoludwig, line 288:
```suggestion
```

integreat_cms/cms/views/events/event_form_view.py
```diff
@@ -220,6 +244,26 @@ def post(self, request, **kwargs):
...
+                and machine_translation_form.is_valid()
+                and machine_translation_form.data.get("automatic_translation")
+                and not event_translation_form.data.get("minor_edit")
+            ):
+                event_translation_instance.refresh_from_db()
+                deepl = DeepLApi()
+                deepl.deepl_translate_to_languages(
+                    request,
+                    event_translation_instance.event,
+                    language.slug,
```
timoludwig, line 260:
```suggestion
```

integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -325,6 +348,26 @@ def post(self, request, *args, **kwargs):
...
+                and machine_translation_form.is_valid()
+                and machine_translation_form.data.get("automatic_translation")
+                and not page_translation_form.data.get("minor_edit")
+            ):
+                page_translation_instance.refresh_from_db()
+                deepl = DeepLApi()
+                deepl.deepl_translate_to_languages(
+                    request,
+                    page_translation_instance.page,
+                    language.slug,
```
timoludwig, line 364:
```suggestion
```

integreat_cms/cms/views/pois/poi_form_view.py
```diff
@@ -173,6 +196,27 @@ def post(self, request, *args, **kwargs):
...
+                and machine_translation_form.is_valid()
+                and machine_translation_form.data.get("automatic_translation")
+                and not poi_translation_form.data.get("minor_edit")
+            ):
+                poi_translation_instance.refresh_from_db()
+                deepl = DeepLApi()
+                deepl.deepl_translate_to_languages(
+                    request,
+                    poi_translation_instance.poi,
+                    language.slug,
```
timoludwig, line 213:
```suggestion
```

timoludwig (changes requested):
Thanks a lot, really awesome feature :heart_eyes: 

My comments are mostly related to implementation details, the functionality itself is working as expected :rocket:

JoeyStk (changes requested):
Thank you very much for your changes!
I still noticed a few things:

- I think this also already resolves issue #2110, which is awesome
- Thank you for already implementing the option to create new translations, which also is amazing
- I think @timoludwig also already noticed this: in our live system we sometimes use BCP-47 codes such as "pt" instead of "pt-PT". When I tried translation a language whose BCP-47 wasn't properly filled in, I got an Django error. And also just for the case that someone mistypes the BCP-47 code we should provide an error message 
grafik

integreat_cms/cms/forms/translations/machine_translation_form.py
```diff
@@ -0,0 +1,112 @@
...
+    def __init__(self, instance, region, language, **kwargs):
+        r"""
+        Initialize MT translation form
+
+        :param \**kwargs: The supplied keyword arguments
+        :type \**kwargs: dict
+        """
+        super().__init__(**kwargs)
+
+        self.initial["automatic_translation"] = False
```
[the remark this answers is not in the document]
charludo: Sorry, I went through a couple iterations with the implementation in this PR, looks like I missed a handful of "Überbleibsel" 😬

integreat_cms/cms/forms/translations/machine_translation_form.py
```diff
@@ -0,0 +1,112 @@
...
+
+        translation_targets = [
+            language
+            for language in region.language_tree_nodes.filter(parent=language.id)
+            if language.mt_provider and mt_to_lang_is_permitted(region, language.slug)
+        ]
```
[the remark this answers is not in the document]
charludo: I was not aware, thanks to you and @JoeyStk further down for pointing this out.

And since using the prefetched tree isn't really less readable, even minimal performance improvement sounds good to me, so let's go with that version 😆

integreat_cms/cms/forms/translations/machine_translation_form.py
```diff
@@ -0,0 +1,112 @@
...
+        ].queryset = region.language_tree_nodes.filter(
+            id__in=[language[0] for language in to_update]
+        )
+        self.fields[
+            "translations_to_create"
+        ].queryset = region.language_tree_nodes.filter(
+            id__in=[language[0] for language in to_create]
+        )
+
+        self.initial["translations_to_update"] = [language[0] for language in to_update]
```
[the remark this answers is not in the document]
charludo: Good to know. The only reason I was using `queryset` here at all is because I saw in your implementation [here](https://github.com/digitalfabrik/integreat-cms/blob/c6a29e4ef0c73ec29641f468b0aabb7b60825ad2/integreat_cms/cms/templates/translations/translations_management.html#L45) that it was possible to check for the existence of choices in a template this way, and it seems with just `choices` that's unfortunately not possible. Is there a reason both `choices` and `queryset` in the [corresponding form](https://github.com/digitalfabrik/integreat-cms/blob/c6a29e4ef0c73ec29641f468b0aabb7b60825ad2/integreat_cms/cms/forms/translations/translations_management_form.py#L81)?

integreat_cms/cms/forms/translations/machine_translation_form.py
```diff
@@ -0,0 +1,112 @@
...
+        cleaned_data["translation_target_languages"].extend(
+            language.slug for language in cleaned_data.pop("translations_to_update", [])
+        )
+        cleaned_data["translation_target_languages"].extend(
+            language.slug for language in cleaned_data.pop("translations_to_create", [])
+        )
+
+        if not cleaned_data["automatic_translation"]:
+            cleaned_data["translation_target_languages"] = []
+        return cleaned_data
```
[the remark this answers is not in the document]
charludo: Yeah, fair point. I guess it was working fine, but I see how it could lead to confusion. Plus I like your solution of calling `get_target_language_slugs` from the form views!

integreat_cms/static/src/js/mutually_exclusive_checkboxes.ts
```diff
@@ -0,0 +1,23 @@
...
+    });
+    // eslint-disable-next-line no-param-reassign
+    clickedCheckbox.checked = true;
+};
+
+window.addEventListener("load", () => {
+    const checkboxes = <HTMLInputElement[]>(<any>document.querySelectorAll(".mutually-exclusive-checkbox"));
+    checkboxes.forEach((checkbox) => {
+        // eslint-disable-next-line no-param-reassign
+        checkbox.addEventListener("click", () => enforceMutualExclusion(checkbox));
```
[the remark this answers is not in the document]
charludo: Oh wow, no idea how I missed that! That's a bit embarassing.

Hm, the only other way I see is an additional event listener for the `automatic_translation` checkbox which adds the `hidden` class back to the `language-options` on uncheck. It might be a bit less "hacky", but not sure if it's worth the effort, since currently the mutual exclusion is not used anywhere else.

Your call, if you are reasonably certain we won't be needing to reuse this anywhere else, I'll go with your original suggestion.

integreat_cms/cms/forms/translations/machine_translation_form.py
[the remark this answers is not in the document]
timoludwig: I think `queryset` is always required for a model (multiple) choice form, and in this case I explicitly used choices to allow the grouping by mt provider (there is no possibility to set [optgroups](https://github.com/digitalfabrik/integreat-cms/blob/c6a29e4ef0c73ec29641f468b0aabb7b60825ad2/integreat_cms/cms/templates/translations/languages_input.html#L2) just by setting the queryset I think.

In this case, we could also have used a normal choice field instead of a model choice field to save a few sql queries, but in the end, a model choice field is probably the cleaner way to do it, so I'd leave it that way.

integreat_cms/deepl_api/utils.py
```diff
@@ -111,7 +121,12 @@ def deepl_translation(self, request, content_objects, language_slug, form_class)
             )
             # Get target language
             target_language = region.get_language_or_404(language_slug)
-            source_language = region.get_source_language(language_slug)
+
+            source_language = (
+                region.get_language_or_404(source_language_slug)
+                if source_language_slug
+                else region.get_source_language(language_slug)
+            )
```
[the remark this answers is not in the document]
charludo: Ahh, this is another one of those remnants of an earlier iteration on this. I was originally always passing the region's root language here, assuming that it would be preferable to avoid "machine translation chains" like DE->EN->FR->ES. Since it's been decided that only one translation step (direct children) should be taken, you are right, this is now superfluous.

charludo:
> * I think this also already resolves issue #2110, which is awesome
Unfortunately I don't think so, right now everything is being passed to DeepL, which simply discards simplified German. Might be a good idea to move #2110 up to the currenct milestone though, otherwise it might be confusing for regions why the translation doesn't work?

> * Thank you for already implementing the option to create new translations, which also is amazing
It was agreed upon, after all 😆 

> * I think @timoludwig also already noticed this: in our live system we sometimes use BCP-47 codes such as "pt" instead of "pt-PT". When I tried translation a language whose BCP-47 wasn't properly filled in, I got an Django error. And also just for the case that someone mistypes the BCP-47 code we should provide an error message

Yes, thanks, I was not aware. I think the suggested solution by @timoludwig is already sufficient, right? There should no longer be the need for a warning?

integreat_cms/cms/forms/translations/machine_translation_form.py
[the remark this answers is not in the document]
charludo: Yes, it originally was a non-model `MultipleChoiceForm`. But that overly complicated checking if the list of choices was empty inthe template.

integreat_cms/static/src/js/mutually_exclusive_checkboxes.ts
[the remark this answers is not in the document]
timoludwig: I'm not sure I understand, just setting `checkbox.checked = false;` somewhere doesn't fire any events, not even change events, right?
Also, my suggestion should be reusable for other use cases as well, or am I missing something?

JoeyStk:
Yes, I can open an issue for it

integreat_cms/static/src/js/mutually_exclusive_checkboxes.ts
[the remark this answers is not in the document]
charludo: Just tested it, yep you're right, that doesn't fire the event. Hm. The only thing I'm taking an issue with is removing and readding the event listener.

I just pushed an alternative suggestion, does this work for you?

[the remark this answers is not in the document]
timoludwig: Ah yeah, dispatching the change event makes even more sense to avoid having to unregister the listener :+1:
Then we just have to make sure we remember this when we want to use the click event on mutually exclusive fields in the future :grinning:

timoludwig (approved):
:rocket:

charludo:
Is someone constantly re-starting the develop workflow on CircleCI?

timoludwig:
> Is someone constantly re-starting the develop workflow on CircleCI?

Yes, I'm not sure why, but something is causing the builds to fail :(
I cannot identify any relation to this PR, so I opened issue #2119 for this.
I restarted the whole build now (instead of only the failing job), maybe this causes the parallelized containers to be assigned in a different way, which might fix the issue for now.

timoludwig:
These race conditions are downright bullying me.
Well, I didn't find a way around them, so I disabled the test for now. This topic continues in #2119, but this PR can be merged now I guess.

JoeyStk (approved):
:tada:

timoludwig:
@charludo I'm going to merge this now in order to avoid conflicts with #2114, ok?

Issue #1970, opened 2022-12-13
Disable page order section for observers without edit/change permission


MizukiTemma:
### Describe the Bug

Observers can change the page order by drag&drop even if they are not allowed to change/edit the page. Even though such changes will not be saved when no change/edit permission is granted, it is confusing and therefore must be disabled.

### Steps to Reproduce

1. Log in as observer
2. Go to a page for which this observer has no permission to change/edit
3. Go to "Order" section
4. See drag&drop is possible

### Expected Behavior

Observers cannot drag&drop the page order if they do not have change/edit permission for this page.

### Actual Behavior

Observers can drag&drop.

PR #2072, opened 2023-02-16
Deactivate page order drag and drop for observers

MizukiTemma:
### Short description

This PR deactivates page order change by drag&drop in the page form if user does not have permission to change the page.

While working on this issue, I found a similar problem with icon setting of page/poi/event. Icon choise is now deactivated too when user does not have change permission.

### Proposed changes

- Set the page node in the page order section undraggable if user has no change permission.
- Deactivate all the event listners of the page order section.
- Hide icon set/change buttons in page/poi/event form if user has no change permission.

### Side effects

- None (I hope)

### Resolved issues

Fixes: #1970 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ UNRELEASED
 * [ [#1322](https://github.com/digitalfabrik/integreat-cms/issues/1322) ] Add organization name & logo to the page API response
 * [ [#2105](https://github.com/digitalfabrik/integreat-cms/issues/2105) ] Indicate that categories are now shown in the apps
 * [ [#2065](https://github.com/digitalfabrik/integreat-cms/issues/2065) ] Only allow staff members to use SUMM.AI bulk action
+* [ [#1970](https://github.com/digitalfabrik/integreat-cms/issues/1970) ] Disable drag & drop in page order for users without editing permission
+* [ [#2112](https://github.com/digitalfabrik/integreat-cms/issues/2112) ] Hide form buttons and bulk actions for users without editing permissions
 
 
 2023.2.2
--- a/integreat_cms/cms/templates/events/event_list_archived.html
+++ b/integreat_cms/cms/templates/events/event_list_archived.html
@@ -81,17 +81,19 @@
             </tbody>
         </table>
     </div>
-    <div class="flex flex-wrap gap-2 mt-4">
-        <select id="bulk-action" class="w-auto max-w-full">
-            <option>
-                {% translate "Select bulk action" %}
-            </option>
-            <option data-bulk-action="{% url 'bulk_restore_events' region_slug=request.region.slug language_slug=language.slug %}">
-                {% translate "Restore events" %}
-            </option>
-        </select>
-        <button form="bulk-action-form" id="bulk-action-execute" class="btn" disabled>{% translate "Execute" %}</button>
-    </div>
+    {% if perms.cms.change_event %}
+        <div class="flex flex-wrap gap-2 mt-4">
+            <select id="bulk-action" class="w-auto max-w-full">
+                <option>
+                    {% translate "Select bulk action" %}
+                </option>
+                <option data-bulk-action="{% url 'bulk_restore_events' region_slug=request.region.slug language_slug=language.slug %}">
+                    {% translate "Restore events" %}
+                </option>
+            </select>
+            <button form="bulk-action-form" id="bulk-action-execute" class="btn" disabled>{% translate "Execute" %}</button>
+        </div>
+    {% endif %}
     {% include "../generic_confirmation_dialog.html" %}
     {% url "events" as url %}
     {% include "pagination.html" with chunk=events %}
--- a/integreat_cms/cms/templates/pages/_page_order_table.html
+++ b/integreat_cms/cms/templates/pages/_page_order_table.html
@@ -2,7 +2,7 @@
 {% load content_filters %}
 {% load page_filters %}
 <div class="table-listing">
-    <table data-activate-page-order
+    <table {% if can_edit_page %}data-activate-page-order{% endif %}
            class="w-full mt-4 rounded border border-solid border-gray-200 shadow bg-white table-auto cursor-default break-all">
         <tbody>
             {# djlint:off H020 #}
@@ -12,8 +12,9 @@
                         <td class="single_icon min">
                             <span data-contained-in-siblings="true"
                                   class="drag text-gray-800 block py-3 px-2 cursor-move"
-                                  {% if not page.archived %} draggable="true" title="{% translate "Change the position of the page with drag & drop." %}" {% endif %}>
-                                <i icon-name="move"></i>
+                                  draggable="{% if not page.archived and can_edit_page %}true{% else %}false{% endif %}"
+                                  title="{% translate "Change the position of the page with drag & drop." %}">
+                                {% if can_edit_page %}<i icon-name="move"></i>{% endif %}
                             </span>
                         </td>
                         <td id="page_title"
@@ -77,8 +78,8 @@
                     <tr class="border-2 border-blue-500 hover:bg-gray-100">
                         <td class="single_icon min">
                             <span class="drag text-gray-800 block py-3 px-2 cursor-move"
-                                  draggable="true">
-                                <i icon-name="move"></i>
+                                  draggable="{% if can_edit_page %}true{% else %}false{% endif %}">
+                                {% if can_edit_page %}<i icon-name="move"></i>{% endif %}
                             </span>
                         </td>
                         <td id="page_title" data-default-title="{% translate "New Page" %}">{% translate "New Page" %}</td>
@@ -87,8 +88,8 @@
                     <tr class="border-2 border-blue-500 hover:bg-gray-100">
                         <td class="single_icon min">
                             <span class="drag text-gray-800 block py-3 px-2 cursor-move"
-                                  draggable="true">
-                                <i icon-name="move"></i>
+                                  draggable="{% if can_edit_page %}true{% else %}false{% endif %}">
+                                {% if can_edit_page %}<i icon-name="move"></i>{% endif %}
                             </span>
                         </td>
                         <td id="page_title"
--- a/integreat_cms/cms/templates/pages/page_tree_archived.html
+++ b/integreat_cms/cms/templates/pages/page_tree_archived.html
@@ -5,6 +5,7 @@
 {% load page_filters %}
 {% load rules %}
 {% block content %}
+    {% has_perm 'cms.change_page_object' request.user as can_edit_pages %}
     {% with filter_form.filters_visible as filters_visible %}
         <div class="table-header">
             <div class="flex flex-wrap justify-between gap-4">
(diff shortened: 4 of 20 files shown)


integreat_cms/cms/templates/pages/_page_order_table.html
```diff
@@ -77,7 +78,7 @@
                     <tr class="border-2 border-blue-500 hover:bg-gray-100">
                         <td class="single_icon min">
                             <span class="drag text-gray-800 block py-3 px-2 cursor-move"
-                                  draggable="true">
+                                  draggable="{% if can_edit_page %}true{% else %}false{% endif %}">
```
membralala, line 81: I think it could maybe even make sense to remove the "move-cursor" and maybe even remove the move-icon.

membralala (commented):
Thanks, I think that you fixed the issues that you described. However I noticed a few more places, where the UI for an observer needs a few restrictions: 
- [x] on the POI form it is possible to change coordinates on the map (not through the input fields) 
- [ ] on the POI form it is possible to change the opening hours
- [ ] on the page tree view it is allowed to move page rows via drag and drop
- [x] in the side-by-side view it is possible to use the "copy content" button

integreat_cms/cms/templates/pages/_page_order_table.html
MizukiTemma: Applied!

seluianova (commented):
Looks good to me!
I also noticed, that "Archive" and "Restore" actions are available for Observer, however they return 403. Relevant for pages/events/pois, as button in the table as well as in the bulk actions.
Do we want to fix it as well?

**UPD**. One more questionable thing:
As Observer I can see "Import XLIFF files" function, but it returns 403 even though I want to load a translation for a page for which I have write permissions.
What should be an expected behavior?
- Observer should be able to upload XLIFF for editable pages
or
- "Import XLIFF files" function should not be visible to Observer at all

**UPD-2**. From the answer in the channel:
_"Import XLIFF files" function should not be visible to Observer at all_
Do we want to fix it here or separately?

membralala:
Apart from the things @seluianova has mentioned, everything looks good to me now 🙂.

MizukiTemma:
@seluianova Nice catches and thank you for the update with the answer from the channel 😃 
I've fixed them. Can you review again?

seluianova (commented):
Thanks for updates!
Findings from another review round:

1. When I change a pin location on the map, it produces TS error:
image

2. Empty bulk action list looks maybe not ideal IMO, but we can think about it separately:
image

MizukiTemma:
@seluianova 
Thank you for the review 😃 

The error with the map pin appears at loading of the form, right? So far as I tested, observers cannot move the pin. If you know any case tha obervers can move the pin, please tell me the details 💪 The error at loading must be solved now.

I'll ask Service Team about the bulk action and will opne an issue when changes are needed.

MizukiTemma:
@seluianova 
The empty bulk actions in POI and Event list are hidden now for observers.

MizukiTemma:
Thank you for all the careful reviews!

We probably find more elements that should be deactivated/hidden for observers in the review process.
I would like to but first complete this PR to release the solution for the initial problem with drag&drop in page order table, so please entry any further feature changes to the issue #2112. Thank you in advance 😃

MizukiTemma:
> It seems, the initial issue with drag-and-drop re-appeared. As observer, I can change coordinates now

Really? I can't reproduce it...
if you mean grapping and moving the map around, I think that's okay, because it doesn't change the coordinates (same with zooming in and out).

seluianova:
> Really? I can't reproduce it...

Nope, sorry. I am updating my comment 🙈

MizukiTemma:
Ah ok, now I see the problem  😓  Can you add it to #2112 .

The drag&drop problem initially reported is the one in the page order section in page form (see #1970). To release the solution for the page section problem, I prefer to collect all the other problems in #2112.

seluianova (approved):
@MizukiTemma sure! thank you. the changes look good to me!

integreat_cms/cms/templates/events/event_form_sidebar/icon_box.html
```diff
@@ -11,5 +11,17 @@
...
+        {% if event_form.instance.icon.url %}
+            {# djlint:off H006 #}
+            <img src="{{ event_form.instance.icon.url }}"
+                 class="h-40 m-auto"
+                 alt="{{ event_form.instance.icon.url }}"/>
+            {# djlint:on #}
+        {% else %}
+            {% translate "No icon is set" %}
+        {% endif %}
+    {% endif %}
```
timoludwig, line 26:
```suggestion
    {% render_field event_form.icon label=event_form.icon.label %}
```
I think it would be more in line with Django's design principles if we always render the form field here - and just don't show the edit buttons in the template of the icon widget:

Wrap the following block with `{% if not widget.attrs.disabled %}`:

https://github.com/digitalfabrik/integreat-cms/blob/3a27a1732671c41c1d9854f1a293798837616459/integreat_cms/cms/templates/icon_widget.html#L17-L45

integreat_cms/cms/templates/events/event_list_archived.html
```diff
@@ -86,9 +86,11 @@ <h1 class="heading">{% translate "Archived events" %}</h1>
...
                 {% translate "Select bulk action" %}
             </option>
-            <option data-bulk-action="{% url 'bulk_restore_events' region_slug=request.region.slug language_slug=language.slug %}">
-                {% translate "Restore events" %}
-            </option>
+            {% if perms.cms.change_event %}
+                <option data-bulk-action="{% url 'bulk_restore_events' region_slug=request.region.slug language_slug=language.slug %}">
+                    {% translate "Restore events" %}
+                </option>
+            {% endif %}
```
timoludwig, line 93: Here I think we can also hide the entire bulk actions

integreat_cms/cms/templates/pages/page_form_sidebar/settings_box.html
```diff
@@ -12,7 +12,19 @@
...
+            {% if page_form.instance.icon.url %}
+                {# djlint:off H006 #}
+                <img src="{{ page_form.instance.icon.url }}"
+                     class="h-40 m-auto"
+                     alt="{{ page_form.instance.icon.url }}"/>
+                {# djlint:on #}
+            {% else %}
+                {% translate "No icon is set" %}
+            {% endif %}
+        {% endif %}
```
timoludwig, line 27:
```suggestion
        {% render_field page_form.icon label=page_form.icon.label %}
```
Here as well - I'd handle this in the widget template

integreat_cms/cms/templates/pages/page_tree_archived.html
```diff
@@ -94,9 +95,11 @@ <h1 class="heading">{% translate "Archived Pages" %}</h1>
...
                         {% translate "Select bulk action" %}
                     </option>
-                    <option data-bulk-action="{% url 'bulk_restore_pages' region_slug=request.region.slug language_slug=language.slug %}">
-                        {% translate "Restore pages" %}
-                    </option>
+                    {% if can_edit_pages %}
+                        <option data-bulk-action="{% url 'bulk_restore_pages' region_slug=request.region.slug language_slug=language.slug %}">
+                            {% translate "Restore pages" %}
+                        </option>
+                    {% endif %}
```
timoludwig, line 102: Here hide all bulk actions as well?

integreat_cms/cms/templates/pois/poi_form_sidebar/icon_box.html
```diff
@@ -10,6 +10,18 @@
...
+        {% if poi_form.instance.icon.url %}
+            {# djlint:off H006 #}
+            <img src="{{ poi_form.instance.icon.url }}"
+                 class="h-40 m-auto"
+                 alt="{{ poi_form.instance.icon.url }}"/>
+            {# djlint:on #}
+        {% else %}
+            {% translate "No icon is set" %}
+        {% endif %}
+    {% endif %}
```
timoludwig, line 26:
```suggestion
    <div class="help-text">{{ poi_form.icon.help_text }}</div>
    {% render_field poi_form.icon label=poi_form.icon.label %}
```

integreat_cms/cms/templates/pois/poi_list_archived.html
```diff
@@ -55,9 +55,11 @@ <h1 class="heading">{% translate "Archived locations" %}</h1>
...
                 {% translate "Select bulk action" %}
             </option>
-            <option data-bulk-action="{% url 'bulk_restore_pois' region_slug=request.region.slug language_slug=language.slug %}">
-                {% translate "Restore locations" %}
-            </option>
+            {% if perms.cms.change_poi %}
+                <option data-bulk-action="{% url 'bulk_restore_pois' region_slug=request.region.slug language_slug=language.slug %}">
+                    {% translate "Restore locations" %}
+                </option>
+            {% endif %}
```
timoludwig, line 62: And here as well hide the whole bulk actions?

integreat_cms/cms/views/pois/poi_context_mixin.py
```diff
@@ -72,6 +72,7 @@ def get_context_data(self, **kwargs):
                     "Opening time is identical with the closing time of the previous slot"
                 ),
             },
+            "canChangeEvent": self.request.user.has_perm("cms.change_event"),
```
timoludwig, line 75:
```suggestion
            "canChangeLocation": self.request.user.has_perm("cms.change_poi"),
```
I think we should only check for the poi permission when editing pois, right?
This also impacts all other occurrences of this string.

integreat_cms/static/src/js/pois/opening-hours/component/opening-hours-display.tsx
```diff
@@ -25,8 +26,8 @@ const OpeningHoursDisplay = ({ openingHours, setSelectedDays, translations, days
                     key={translations.weekdays[day]}
                     title={translations.editWeekdayLabel}
                     className="flex flex-wrap justify-between gap-2 hover:bg-gray-50 p-4 border-b cursor-pointer"
```
timoludwig, line 28:
```suggestion
                    className={cn("flex flex-wrap justify-between gap-2 p-4 border-b", {
                        "hover:bg-gray-50 cursor-pointer": canChangeEvent,
                    })}
```
Hmm, I think the hover effect and the pointer cursor somehow arise the impression that there is something to be clicked, so I'd remove the classes without the edit permissions.
This requires the import `import cn from "classnames";`.

timoludwig (changes requested):
Thanks a lot! :rocket: 
This is a crazy rabbit hole :rabbit: :sweat_smile:

MizukiTemma:
@timoludwig 
Thank you for detailed reveiw with code suggestions!
The changes are now applied 😃

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ UNRELEASED
 * [ [#1322](https://github.com/digitalfabrik/integreat-cms/issues/1322) ] Add organization name & logo to the page API response
 * [ [#2105](https://github.com/digitalfabrik/integreat-cms/issues/2105) ] Indicate that categories are now shown in the apps
 * [ [#2065](https://github.com/digitalfabrik/integreat-cms/issues/2065) ] Only allow staff members to use SUMM.AI bulk action
+* [ [#1970](https://github.com/digitalfabrik/integreat-cms/issues/1970) ] Disable drag&drop in page order for users without permission
```
timoludwig, line 9:
```suggestion
* [ [#1970](https://github.com/digitalfabrik/integreat-cms/issues/1970) ] Disable drag & drop in page order for users without editing permission
* [ [#2112](https://github.com/digitalfabrik/integreat-cms/issues/2112) ] Hide form buttons and bulk actions for users without editing permissions
```
I think you can credit yourself for all the additional things you have fixed in this PR :grin:

timoludwig (approved):
Thanks a lot! :muscle:

PR #2082, opened 2023-02-18
Add mutex to tree move function, fix #1518

### LINKED ISSUE #1518: Race conditions in tree operations causing Internal Server Errors

svenseeberg:
### Short description
Prevent race conditions in page tree. This is no replacement for database transactions. This would not catch problems where the process dies in the middle of changing a tree structure.

### Proposed changes
* Use django-db-mutex module to implement a per class mutex that allows only one page tree move operation at a time.
* This can be replaced with basically the same functionality within tree beard if it works.
* As we have multiple processes in production, the database back end is used for communication. We could also use Redis, which would actually be a little faster. However, the app implements it this way and it should be good enough for testing if we continue to have 500 errors in our logs.
* Saving of other data should continue even when the move action fails. The user receives a notification about the failure to move the page.

### Side effects
* In a few cases users may have to wait a little longer.
* If getting the mutex fails, the user as to redo the action.
* The mutex timeout / TTL seems to have no effect that I could verify. However this should still be better than the status quo. In a test I added a sleep for 60s in the block that requires the mutex and set the mutex TTL to 5s. No error was raised.

### Resolved issues
Fixes: #1518

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -506,6 +506,13 @@
             "markers": "python_version >= '3.7'",
             "version": "==3.13.0"
         },
+        "django-db-mutex": {
+            "hashes": [
+                "sha256:a2c882ec677b80b931a33ae902fe347a634de64a43b5ac3da4e5d64230ae2afe",
+                "sha256:dadbd6fdae070749ddbb76a86b9fb68a5e45228b366f707868267468f1aad774"
+            ],
+            "version": "==3.0.0"
+        },
         "django-debug-toolbar": {
             "hashes": [
                 "sha256:24ef1a7d44d25e60d7951e378454c6509bf536dce7e7d9d36e7c387db499bc27",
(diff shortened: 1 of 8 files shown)


svenseeberg:
I will improve the user notification and error handling a little before we merge this.

integreat_cms/cms/models/abstract_tree_node.py
```diff
@@ -282,7 +287,21 @@ def move(self, target, pos=None):
         # Moving a node can modify all other nodes via raw sql queries (which are not recognized by cachalot),
         # so we have to invalidate the whole model manually.
         invalidate_model(self.__class__)
-        super().move(target=target, pos=pos)
+        try:
+            with db_mutex(f"tree-page-{self.region}"):
```
timoludwig, line 291:
```suggestion
            with db_mutex("page-tree"):
```
This is not sufficient, since all page trees of all regions are stored in the same database table and moving a page in one region can also change the `tree_id`s of pages in other regions (which is probably the main source of the race conditions, in most cases there aren't multiple pages moved in the same region at the same time - only because of the interference between regions, this is such a big problem).

Apart from that: Do you think the Python inbuilt [`multiprocessing.Lock`](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Lock) could also be used for this task?

integreat_cms/cms/models/abstract_tree_node.py
```diff
@@ -282,7 +287,21 @@ def move(self, target, pos=None):
...
+            raise DBMutexError(
+                # pylint: disable=consider-using-f-string
+                _('Could not change position in tree of "{}".'.format(self))
+            ) from e
+        except DBMutexTimeoutError as e:
+            logger.error("Retrieving database mutex timed out while saving %s", self)
+            raise DBMutexTimeoutError(
+                # pylint: disable=consider-using-f-string
+                _('Could not change position in tree of "{}".'.format(self))
+            ) from e
```
timoludwig, line 304: These exceptions should probably be caught in the view?
https://github.com/digitalfabrik/integreat-cms/blob/534f226f80618be350bb10c1e6d7c7448147db66/integreat_cms/cms/views/pages/page_actions.py#L494

integreat_cms/cms/models/abstract_tree_node.py
```diff
@@ -282,7 +287,21 @@ def move(self, target, pos=None):
...
         invalidate_model(self.__class__)
-        super().move(target=target, pos=pos)
+        try:
+            with db_mutex(f"tree-page-{self.region}"):
+                super().move(target, pos)
+        except DBMutexError as e:
+            logger.error("Could not retrieve database mutex to save %s", self)
+            raise DBMutexError(
+                # pylint: disable=consider-using-f-string
+                _('Could not change position in tree of "{}".'.format(self))
```
timoludwig, line 297:
```suggestion
                _('Could not change position in tree of "{}".').format(self)
```
You need to translate first, then perform the string formatting.

integreat_cms/cms/models/abstract_tree_node.py
```diff
@@ -282,7 +287,21 @@ def move(self, target, pos=None):
...
+            ) from e
+        except DBMutexTimeoutError as e:
+            logger.error("Retrieving database mutex timed out while saving %s", self)
+            raise DBMutexTimeoutError(
+                # pylint: disable=consider-using-f-string
+                _('Could not change position in tree of "{}".'.format(self))
```
timoludwig, line 303:
```suggestion
                _('Could not change position in tree of "{}".').format(self)
```

Pipfile
```diff
@@ -29,6 +29,7 @@ twine = "*"
 
 [packages]
 integreat-cms = {editable = true, path = "."}
+django-db-mutex = "*"
```
timoludwig, line 32:
```suggestion
```
Functional dependencies need to be added to the [setup.cfg](https://github.com/digitalfabrik/integreat-cms/blob/534f226f80618be350bb10c1e6d7c7448147db66/setup.cfg#L43) (At least until #2027 is merged...)

timoludwig (changes requested):
Thanks, this will hopefully reduce our problems! :crossed_fingers:

integreat_cms/cms/models/abstract_tree_node.py
svenseeberg: Global lock: Yeah, you're right. I missed that.

About [multiprocessing.Lock](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Lock): I don't think so, as Apache2 starts the processes independently of each other (AFAICT). In the example of the documentation the lock needs to be passed to each process: https://docs.python.org/3/library/multiprocessing.html#synchronization-between-processes

integreat_cms/cms/models/abstract_tree_node.py
svenseeberg: :facepalm:

integreat_cms/cms/models/abstract_tree_node.py
svenseeberg: Ah yes, it should. Interestingly, I could not provoke the (timeout) exception yet. In all my tests (mostly with sleep) all processes waited for others to finish. And everything finished orderly.

integreat_cms/cms/models/abstract_tree_node.py
timoludwig: Ah, yeah sounds plausible.

integreat_cms/cms/models/abstract_tree_node.py
```diff
@@ -282,7 +287,19 @@ def move(self, target, pos=None):
         # Moving a node can modify all other nodes via raw sql queries (which are not recognized by cachalot),
         # so we have to invalidate the whole model manually.
         invalidate_model(self.__class__)
-        super().move(target=target, pos=pos)
+        try:
+            with db_mutex("page-tree"):
```
timoludwig, line 291:
```suggestion
            with db_mutex(self.__class__.__name__):
```
Ah, just noticed this is the shared model between language tree and page tree, so technically we can use two locks for those two models :sweat_smile: 

Also, I thought maybe it would make sense to put the constraint check in the lines 270-286 into the lock-protected area as well? Otherwise the condition might not be true anymore once the lock is acquired.

integreat_cms/cms/models/abstract_tree_node.py
```diff
@@ -282,7 +287,19 @@ def move(self, target, pos=None):
...
+                super().move(target, pos)
+        except DBMutexError as e:
+            logger.error("Could not retrieve database mutex to save %s", self)
```
timoludwig, line 294:
```suggestion
            logger.error("Could not retrieve database mutex to save %r", self)
```
For debugging purposes, I always find it more helpful to get the page id etc. in the log instead of the title

integreat_cms/cms/models/abstract_tree_node.py
```diff
@@ -282,7 +287,19 @@ def move(self, target, pos=None):
...
+            raise DBMutexError(
+                _('Could not change position in tree of "{}".').format(self)
+            ) from e
+        except DBMutexTimeoutError as e:
+            logger.error("Retrieving database mutex timed out while saving %s", self)
```
timoludwig, line 299:
```suggestion
            logger.error("Retrieving database mutex timed out while saving %r", self)
```

integreat_cms/cms/models/abstract_tree_node.py
```diff
@@ -282,7 +287,19 @@ def move(self, target, pos=None):
...
         # so we have to invalidate the whole model manually.
         invalidate_model(self.__class__)
-        super().move(target=target, pos=pos)
+        try:
+            with db_mutex("page-tree"):
+                super().move(target, pos)
+        except DBMutexError as e:
+            logger.error("Could not retrieve database mutex to save %s", self)
+            raise DBMutexError(
+                _('Could not change position in tree of "{}".').format(self)
```
timoludwig, line 296:
```suggestion
                _('Could not change position in tree of "{}".').format(self.best_translation.title)
```
For sub-pages, this results in:

> Could not change position in tree of "Willkommen \&rarr; Stadtplan".

because since #2016 the string-repreesentation was changed into the path instead the title (which I would revert at some point, but right now I'd also be more explicit how the page is represented, especially since `&rarr;` is printed literally and not rendered to an arrow).

integreat_cms/cms/models/abstract_tree_node.py
```diff
@@ -282,7 +287,19 @@ def move(self, target, pos=None):
...
+            ) from e
+        except DBMutexTimeoutError as e:
+            logger.error("Retrieving database mutex timed out while saving %s", self)
+            raise DBMutexTimeoutError(
+                _('Could not change position in tree of "{}".').format(self)
```
timoludwig, line 301:
```suggestion
                _('Could not change position in tree of "{}".').format(self.best_translation.title)
```

timoludwig (changes requested):

integreat_cms/cms/models/abstract_tree_node.py
svenseeberg: Ah yes, I missed that after moving the code from the Page class to the abstract class.

svenseeberg:
```
Using pylint 2.16.2 for python 3.9.15
pylint: Command line or configuration file:1: UserWarning: Specifying exception names in the overgeneral-exceptions option without module name is deprecated and support for it will be removed in pylint 3.0. Use fully qualified name (maybe 'builtins.Exception' ?) instead.
************* Module integreat_cms.cms.models.abstract_tree_node
integreat_cms/cms/models/abstract_tree_node.py:297:20: E1101: Instance of 'AbstractTreeNode' has no 'best_translation' member (no-member)
integreat_cms/cms/models/abstract_tree_node.py:304:20: E1101: Instance of 'AbstractTreeNode' has no 'best_translation' member (no-member)
```
... a few more things need fixing.

integreat_cms/cms/models/abstract_tree_node.py
timoludwig: Ah sorry, in this case I missed that the language tree nodes do not have a translation... So maybe we have to settle for the default string representation anyways...

svenseeberg: I like the idea of having readable titles. But at the same time I think it is highly unlikely that any user will ever see this message :D So yeah, lets just use %r.

integreat_cms/cms/models/abstract_tree_node.py
```diff
@@ -282,7 +287,23 @@ def move(self, target, pos=None):
         # Moving a node can modify all other nodes via raw sql queries (which are not recognized by cachalot),
         # so we have to invalidate the whole model manually.
         invalidate_model(self.__class__)
-        super().move(target=target, pos=pos)
+        try:
+            with db_mutex(self.__class__.__name__):
+                super().move(target, pos)
+        except DBMutexError as e:
+            logger.error("Could not retrieve database mutex to save %r", self)
```
timoludwig, line 294:
```suggestion
```
Ah, since the exception is logged in the page actions anyway, maybe we don't need to log it here again?

integreat_cms/cms/models/abstract_tree_node.py
```diff
@@ -282,7 +287,23 @@ def move(self, target, pos=None):
...
+            raise DBMutexError(
+                _('Could not change position in tree of "{}".').format(
+                    self.best_translation.title
+                )
+            ) from e
+        except DBMutexTimeoutError as e:
+            logger.error("Retrieving database mutex timed out while saving %r", self)
```
timoludwig, line 301:
```suggestion
```

integreat_cms/cms/models/abstract_tree_node.py
```diff
@@ -282,7 +287,23 @@ def move(self, target, pos=None):
...
-        super().move(target=target, pos=pos)
+        try:
+            with db_mutex(self.__class__.__name__):
+                super().move(target, pos)
+        except DBMutexError as e:
+            logger.error("Could not retrieve database mutex to save %r", self)
+            raise DBMutexError(
+                _('Could not change position in tree of "{}".').format(
+                    self.best_translation.title
+                )
```
timoludwig, line 298:
```suggestion
                _('Could not change position in tree of "{}".').format(self)
```

integreat_cms/cms/models/abstract_tree_node.py
```diff
@@ -282,7 +287,23 @@ def move(self, target, pos=None):
...
+            ) from e
+        except DBMutexTimeoutError as e:
+            logger.error("Retrieving database mutex timed out while saving %r", self)
+            raise DBMutexTimeoutError(
+                _('Could not change position in tree of "{}".').format(
+                    self.best_translation.title
+                )
```
timoludwig, line 305:
```suggestion
                _('Could not change position in tree of "{}".').format(self)
```

Pipfile.lock
```diff
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "b5ee7f98f76daedb5e5b92e8d12ba8a8e0bf8890f4db84deba904cfe0dd7e20e"
+            "sha256": "1715c4c8cb6049cbbb98afcc0f5a885ebd2c9b6f5df357e83c20fbd860472c01"
```
timoludwig, line 4: When I run `./dev-tools/install --clean`, the `Pipfile.lock` is changed again... Doesn't this happen on your machine?
In theory, we have a circleci check for this:
https://github.com/digitalfabrik/integreat-cms/blob/534f226f80618be350bb10c1e6d7c7448147db66/.circleci/config.yml#L23-L27

But somehow, this does not seem to work here :thinking:
Probably, because the old venv state is cached based on the `Pipfile.lock` hash (which didn't change obviously)...

Anyways, could you commit the diff in the `Pipfile.lock` as well?

timoludwig (changes requested):

Pipfile.lock
svenseeberg: Sure. Did that. No idea though what happened and why? :D

integreat_cms/cms/models/abstract_tree_node.py
timoludwig: What do you think about including the constraint check in the mutex area as well?

Pipfile.lock
timoludwig: I think for pipenv it makes a difference whether a package is a direct dependency in the Pipfile, or an implicit dependency by some package in the Pipfile (in this case `integreat-cms`)... After you moved the dependency into `setup.cfg`, the lock file needed to be regenerated...

integreat_cms/cms/models/abstract_tree_node.py
svenseeberg: You mean the database constraints for tree consistency? I'm not yet sure how that would work. The constraints just make some (bad) queries fail, don't they?

timoludwig: I mean those in the lines above:

https://github.com/digitalfabrik/integreat-cms/blob/97fe210503c4233566303fa6c26cc192db4f064e/integreat_cms/cms/models/abstract_tree_node.py#L269-L286

I can imagine a rare edge case where this condition is `False` before acquiring the lock, but then in the mean time until the view waits for the lock, the condition becomes `True` which should cause the operation to fail in theory?

svenseeberg: If we move the last cache flush into the mutex protected block as well, couldn't we omit the first cache flush?

timoludwig: Not sure, I guess there will still be a lot of cases where there is either a non-move operation before or after the move, which requires the flushing on both ends, right?
Only in the edge cases that two moves are directly consecutive, we could omit the flushing?

svenseeberg: Alright, I'll keep it and do a push.

timoludwig (approved):
:tada:

svenseeberg:
@timoludwig @ulliholtgrave lets merge and deploy? I can have a close look at the prod CMS the remaining day.

ulliholtgrave (approved):
Fine by me

timoludwig:
One caveat I just noticed: The pages are also moved when the page form is submitted, so in theory we also need to put the page form into the same mutex? :thinking: 
But this PR can be merged anyways, it's definitely an improvement!

svenseeberg:
Yes lets merge and continue that later.

Issue #1727, opened 2022-10-02
Build page tree for new section "Translations"


JoeyStk:
### Motivation
Since HIX is more or less implemented now, we should extend the MT via DeepL to pages. So far it's only implemented for events and POIs.

### Proposed Solution

- [x] Build page tree template which is the base for the new translation interface
- [x] Add menu item but make it hidden (only visible in DEBUG mode) until the follow-up issues are finished and the view is completed

### Additional Context
This is part of the issue family to enable automated translations

![base-page-tree](https://user-images.githubusercontent.com/16021269/204845635-b7b4da5c-0331-453e-842b-b73affdc77bd.png)
— [Source](https://www.figma.com/file/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=166%3A1980)

The following issues are blocked by this issue and are follow-up issues:
- [x] #1923
- [x] #1924
- [x] #1925

svenseeberg:
We discussed this yesterday: there are not many use cases for this additional view. We may not need it. Cities will probably need a "bulk translate everything" once. After that, pages will simply be auto translated when being edited/updated in the source language.

PR #2086, opened 2023-02-19
Page tree for machine translations

Gaston69:
### Short description
Page tree template which is the base for the new translation interface inkl. menu item (visible on debug only) 
!!! As I saw just now: no longer wanted? https://github.com/digitalfabrik/integreat-cms/issues/1727#issuecomment-1435599281

My comments and questions I still put in here, maybe there is either still a use of the whole or also the one or other answer for other tasks. 

### Questions and comments: 
- To the 'Product-Owners': please check especially all links on the page. Do they lead where intended? 
- In the template a help is intended as mouse-over (see photo in the issue). What content should be displayed there? 
- Where should the form submit to? 
I feel like tackling the follow-up ticket (issue no.) as well. I would like to start this in pair. - Who would have time & want? 

- CheckboxSelectMultiple is coded directly in the html-template (translate_pages_tree_node.html), because I didn't find a simple solution to change css-styles for the label-element wrtitten by the django-method. Feel free to build it into a form-class. The code I started with you find here: /integreat-cms/integreat_cms/cms/forms/translations/translate_pages_form.py 

- The link "Adjust HIX value" I have deviated from the template as button-enabled, because i think, it should link to the page to edit (so as not to leave the user out in the cold). If otherwise desired, please let me know. 

What - from my point of view - is missing: 
- tests for TS-scripts
- tests for view-methods
- integration test (IMHO less needed then those above ;-))

What I find very disturbing: 
- css-style instructions are given in TS code - that s not pretty (translate-pages-forms.ts)

@TODO in the aftermath: 
- go through again which elements are redundantly used in pages and translate_pages and swap them out to templates.

Considerations: 
- for a - small - usability improvement we buy a lot of complexity: 
instead of a CheckboxSelectMultiple I would like to promote to use a simple Html multiselectbox here and - as originally thought - a checkbox option for 'all languages'. 
The whole ugly js then falls away. I myself fell into the trap of converting by template, instead of pursuing the question, what would be the first simple step to let an automatic translation (with perhaps greater limitations/not so good usability) go productive.
- 

### Side effects

- 

### Resolved issues

Fixes: #
 Build page tree for new section "Translations" #1727 


timobrembeck (commented):
@Gaston69 I apologize on behalf of the entire team for the miscommunicatio nand misplanning, sorry for the time you have already invested in this issue!

PR #2087, opened 2023-02-19
Add dev tool to check current review count

timoludwig:
### Short description

Add a small dev-tool which might help us to keep track of our own reviewing workflow.
Requires [`github-cli`](https://cli.github.com/).

### Example usage:

- `./dev-tools/have_i_reviewed_enough.sh`
- `./dev-tools/have_i_reviewed_enough.sh --since "2023-01-01"`
- `./dev-tools/have_i_reviewed_enough.sh --since "-2 months"`
- `./dev-tools/have_i_reviewed_enough.sh --since "last week"`

### DIFF
--- a/dev-tools/have_i_reviewed_enough.sh
+++ b/dev-tools/have_i_reviewed_enough.sh
@@ -1,0 +1,87 @@
+#!/bin/bash
+
+# This script retrieves stats about the pull request reviews from GitHub
+
+# Import utility functions
+# shellcheck source=./dev-tools/_functions.sh
+source "$(dirname "${BASH_SOURCE[0]}")/_functions.sh"
+
+# Check if requirements are satisfied
+if [[ ! -x "$(command -v gh)" ]]; then
+    echo "The GitHub cli is not installed. Please install github-cli manually and run this script again."  | print_error
+    exit 1
+fi
+
+# Parse command line arguments
+while [ "$#" -gt 0 ]; do
+    case "$1" in
+        --since) SINCE="$2"; shift 2;;
+        --since=*) SINCE="${1#*=}"; shift 1;;
+        *) echo "Unknown option: $1" | print_error; exit 1;;
+    esac
+done
+
+if [[ -n "${SINCE}" ]]; then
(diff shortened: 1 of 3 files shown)


JoeyStk (approved):
That is very cool! Thank you a lot

dev-tools/have_i_reviewed_enough.sh
```diff
@@ -0,0 +1,87 @@
...
+    --jq "${JQ_SPLIT_BY_CREATION}" \
+    ${UPDATED_FLAG} \
+))
+OPENED_TOTAL=$((OPENED[0]+OPENED[1]))
+
+# Do not count review comments on own PRs into the score
+# See https://stedolan.github.io/jq/manual/ for JQ documentation
+JQ_REMOVE_OWN="del(.[] | select(.author.login == \"${USERNAME}\")) | length"
+# shellcheck disable=SC2086
+REVIEWED_TOTAL=$(gh search prs \
```
seluianova, line 64: It basically counts not the number of PRs reviewed by a user within a certain time range, but the number of PRs that have ever been reviewed by the user and updated within a certain period of time.

So, if I review some PR, but it gets reworked/updated every week during the next month, it will be counted as +1 for me every week till it actually merged. 

Not sure if it can be done more precisely though.

seluianova (approved):
Looks cool, works fine!
Noted one corner case, not sure if it's worth fixing.

dev-tools/have_i_reviewed_enough.sh
timoludwig: Yes, correct :see_no_evil: 
The problem is that I did not find an easy way of retrieving all pull request reviews in a certain time frame - for more detailed information, there is the [pull request reviews](https://docs.github.com/en/rest/pulls/reviews) endpoint, but this needs a specific PR number. So I would have to first query all PRs in the time period and then query all reviews for each PR, and then count the number of reviews for the specific user - a lot of effort for a simple script which should just help us a bit :smile:

In most cases this will be good enough I think :sweat_smile: 
Especially since usually when a PR is updated, a new review is requested as well...

Issue #2012, opened 2023-01-17
Links found in archived pages after cloning region


timoludwig:
### Describe the Bug

When a region is cloned, the link checker of the new region contains links in archived pages, even if these links were not present in the source region.

### Steps to Reproduce

1. Create region with archived pages that contain links
2. See link checker of that region which does not contain the links in archived pages
3. Clone that region to a new region
4. See link checker of the new region which now contains the links in archived pages

### Expected Behavior

The cloned region should also exclude links in archived pages

### Actual Behavior

The cloned region has links in archived pages

JoeyStk:
Workaround: Archived pages are not supposed to be cloned

PR #2089, opened 2023-02-20
Exclude archived pages when cloning a region

seluianova:
### Short description
When a region is cloned, the link checker of the new region contains links in archived pages, even if these links were not present in the source region.

### Proposed changes
Do not copy archived pages when cloning a region

### Side effects
Didn't find any

### Resolved issues
Fixes: #2012


membralala (approved):
I tested it for "explicit" archived pages as well as for pages that were archived because a parent node was archived and it worked in both cases. 🙂 The fact, that the `duplicate_pages` method from now on does not duplicate archived pages might be worth a note in the doc string as well, I guess.

timoludwig (approved):
Thanks a lot! :+1:

PR #2093, opened 2023-02-22
Fix events endpoint for recurring events

SteffiStuffel:
### Short description
For every recurring event the first event was appended, even if it is in the past. Fixed this.

### Proposed changes
The first event of a recurring event is only appended if recurring events are delivered as one event with rrule.

### Resolved issues

Fixes: #

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ UNRELEASED
 ----------
 
 * [ [#2012](https://github.com/digitalfabrik/integreat-cms/issues/2012) ] Exclude archived pages when cloning a region
+* [ [#2093](https://github.com/digitalfabrik/integreat-cms/issues/2093) ] Fix recurring events endpoint
 
 
 2023.2.1
--- a/integreat_cms/api/v3/events.py
+++ b/integreat_cms/api/v3/events.py
@@ -161,7 +161,7 @@ def transform_event_recurrences(event_translation, today):
             days=settings.API_EVENTS_MAX_TIME_SPAN_DAYS
         ):
             break
-        if recurrence_date < today or recurrence_date == start_date:
+        if recurrence_date < today:
             continue
 
         yield transform_event_translation(event_translation, recurrence_date)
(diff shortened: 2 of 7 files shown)


integreat_cms/api/v3/events.py
```diff
@@ -193,8 +193,12 @@ def events(request, region_slug, language_slug):
...
+            event_is_not_recurring = not event.recurrence_rule
+            recurring_event_should_be_appended_once = "combine_recurring" in request.GET
             # Either it's in the future or it's recurring and the last recurrence is >= now
-            if not event.is_past:
+            if not event.is_past and (
+                event_is_not_recurring or recurring_event_should_be_appended_once
+            ):
                 result.append(transform_event_translation(event_translation))
             if "combine_recurring" not in request.GET:
                 for future_event in transform_event_recurrences(event_translation, now):
```
timoludwig, line 204: I think we could simplify the nested structure of this loop a bit:
```
        if not event.is_past and (
            event_translation := event.get_public_translation(language_slug)
        ):
            combine_recurring_events = "combine_recurring" in request.GET
            if event.is_recurring and not combine_recurring_events:
                result.extend(iter(transform_event_recurrences(event_translation, now)))
            else:
                result.append(transform_event_translation(event_translation))
```

integreat_cms/api/v3/events.py
```diff
@@ -161,7 +161,7 @@ def transform_event_recurrences(event_translation, today):
             days=settings.API_EVENTS_MAX_TIME_SPAN_DAYS
         ):
             break
-        if recurrence_date < today or recurrence_date == start_date:
+        if recurrence_date < today:
```
timoludwig, line 164: Adding the first occurrence to this function definitely makes sense! The problem is that it doesn't quite work when the recurring event goes on for several days and the start_date was yesterday or earlier.
But I think this should be a rather rare edge case in comparison to the bug that is fixed by this PR, so I guess we should merge and fix this problem later.

timoludwig (changes requested):
Sorry for my incorrect suggestion from the last PR and thanks a lot for the quick fix! :innocent: :pray: 

Just have a small suggestion for refactoring the loop and noticed a minor problem which I didn't see an obvious solution for, so I think we can live with that caveat for the moment.

timoludwig:
Oh, and could you add a changelog entry for this? :smile:

JoeyStk (commented):
Thank you a lot for your PR
While testing I had the same suggestions that are mentioned above

integreat_cms/api/v3/events.py
SteffiStuffel: done

CHANGELOG.md
```diff
@@ -1,8 +1,9 @@
+
```
timoludwig, line 1:
```suggestion
```

CHANGELOG.md
```diff
@@ -1,8 +1,9 @@
...
 UNRELEASED
 ----------
 
 * [ [#2012](https://github.com/digitalfabrik/integreat-cms/issues/2012) ] Exclude archived pages when cloning a region
-
+* [ [#2093](https://github.com/digitalfabrik/integreat-cms/issues/2093) ] Fix recurring events endpoint
```
```suggestion


```

timoludwig (changes requested):

CHANGELOG.md
SteffiStuffel: Timo on lightspeed :rocket:
already fixed

timoludwig (approved):
Thanks a lot! :muscle:

JoeyStk (approved):

PR #2095, opened 2023-02-22
Add gzip middleware

SteffiStuffel:
### Short description
Added gzip middleware for compression.

### Side effects

- "Security researchers recently revealed that when compression techniques (including GZipMiddleware) are used on a website, the site may become exposed to a number of possible attacks. Before using GZipMiddleware on your site, you should consider very carefully whether you are subject to these attacks. If you’re in any doubt about whether you’re affected, you should avoid using GZipMiddleware."
- I checked the [BREACH-Paper](https://www.breachattack.com/resources/BREACH%20-%20SSL,%20gone%20in%2030%20seconds.pdf
). We should not be affected by this security risk, as we never pass CSRF-Tokens (or other secrets) in the body. Is this correct?

### Resolved issues

https://issues.tuerantuer.org/browse/IGAPP-1168

Fixes: #

### DIFF
--- a/integreat_cms/core/settings.py
+++ b/integreat_cms/core/settings.py
@@ -277,6 +277,7 @@ MIDDLEWARE = [
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
+    "django.middleware.gzip.GZipMiddleware",
     "integreat_cms.core.middleware.RegionMiddleware",
     "integreat_cms.core.middleware.AccessControlMiddleware",
     "integreat_cms.core.middleware.TimezoneMiddleware",


timoludwig (commented):
> I checked the [BREACH-Paper](https://www.breachattack.com/resources/BREACH%20-%20SSL,%20gone%20in%2030%20seconds.pdf). We should not be affected by this security risk, as we never pass CSRF-Tokens (or other secrets) in the body. Is this correct?

In fact, we do, e.g. on the login page:

https://github.com/digitalfabrik/integreat-cms/blob/a15be1005910cdacb0e62e8d11f5149b8da75d71/integreat_cms/cms/templates/authentication/login.html#L10

What do you think about removing the middleware and using the [`@gzip_page`](https://docs.djangoproject.com/en/3.2/topics/http/decorators/#django.views.decorators.gzip.gzip_page) decorator only on all API views?

svenseeberg:
IMHO gzip compression should not be part of the Django application and rather be activated in the web server (Apache2).

svenseeberg:
> IMHO gzip compression should not be part of the Django application and rather be activated in the web server (Apache2). It should be quite easy to activate gzip for all `/api` paths.

interesting: http://blog.dscpl.com.au/2009/10/details-on-wsgi-10-amendmentsclarificat.html 

However, we should investigate if this 13 years old article is still valid ;-)

SteffiStuffel:
@svenseeberg can you please explain, why you prefer doing this in the web server than in django? Would like to understand.

Not only the api is affected by missing compression, also e.g. https://cms-test.integreat-app.de/static/logos/integreat/integreat-icon.svg is not compressed.

svenseeberg:
> @svenseeberg can you please explain, why you prefer doing this in the web server than in django? Would like to understand.
> 
> Not only the api is affected by missing compression, also e.g. https://cms-test.integreat-app.de/static/logos/integreat/integreat-icon.svg is not compressed.

I think you partially provided the answer yourself: Apache2 will also do it for all other resources (if we allow it). To be honest: I expected mod_deflate to be active. Also, I think Apache2 will do it much more efficiently than a Python application. But that is just an assumption.

SteffiStuffel:
Thanks. Some kind of compression is definitely there, as when looking at the cms-test.integreat.app, nearly everything is compressed.

So preferred solution is: We also enabled compression for svg's in the server config and also for the `/api`

svenseeberg:
I added the following 2 lines to our Apache configuration on integreat-test.tuerantuer.org:

```
AddOutputFilterByType DEFLATE application/json
AddOutputFilterByType DEFLATE image/svg+xml
```

It seems that the API responses and also .svg files are now compressed. Do you want to validate?

SteffiStuffel:
> I added the following 2 lines to our Apache configuration on integreat-test.tuerantuer.org:
> 
> ```
> AddOutputFilterByType DEFLATE application/json
> AddOutputFilterByType DEFLATE image/svg+xml
> ```
> 
> It seems that the API responses and also .svg files are now compressed. Do you want to validate?

Should we make sure the login page is not deflated? Not really sure, but seems to me to be a security issue.

svenseeberg:
AFAICT we're not vulnerable to BREACH. It requires 3 ingredients:
1. Be served from a server that uses HTTP-level compression
2. Reflect user-input in HTTP response bodies
3. Reflect a secret (such as a CSRF token) in HTTP response bodies

I don't think that any user input is returned in our login page. AFAICT it would be if we returned a string like "Username X is invalid". But we do not do that.

timoludwig:
> I don't think that any user input is returned in our login page. That means we're not providing ingredient number 2. AFAICT it would be if we returned a string like "Username X is invalid". But we do not do that.

Well, the form still contains the username in the `value` attribute of the form field.
But we don't compress the login page now, since it uses html and not json?
But a few multi-factor-authentication views return secrets via json I think. Not sure whether this is really exploitable, but it might be.

> Our page forms might be vulnerable though.

Definitely, we have a few ajax views which might contain confidential data.
Would it be possible to tell apache to only deflate json views in a specific namespace?

svenseeberg:
> Well, the form still contains the username in the `value` attribute of the form field.

:see_no_evil: I looked at the wrong (initial) request in the Browser Development tools. Right.

> Would it be possible to tell apache to only deflate json views in a specific namespace?

Yes.

But looking more into the publication I don't think that this is actually a significant problem for us. Yes, in theory CSRF tokens can be leaked. But the headers, that contain the authentication, cannot be leaked without using TLS compression. This is disabled. As we only use body compression, only secrets within the body can be leaked. I'm not aware of any relevant secret in our response bodies besides CSRF tokens, which I would consider a second layer of security. That means that an attacker would need quite a lot of things to get an actual attack working:
- Sniff on the network traffic to get the response size. Probably not a realistic scenario for our CMS.
- Actually inject text into the body (probably the easiest)
- To exploit the CSRF tokens one would still require valid credentials. That would make the whole thing pointless. With access to credentials there would be no need to extract CSRF tokens.

timoludwig:
> * To exploit the CSRF tokens one would still require valid credentials. That would make the whole thing pointless. With access to credentials there would be no need to extract CSRF tokens.

Not necessarily, they could also craft a malicious website with a button "get 999999$", and when a user clicks this button while having a valid session in the cms, they could perform a post-request to the cms doing arbitrary things.

svenseeberg:
Okay, I can think of one scenario were we could actually leak information: the e-mail address of users could be leaked in the user form.

Something else:

> The BREACH attack can be exploited with just a few thousand requests, and can be executed in under a minute. The number of requests required will depend on the secret size.

A few thousand requests per minute would make our CMS crumble under the load. So in most practical scenarios the user would probably already have used the CSRF token by saving the dialog before token can be guessed.

Considering the requirements and results of the attack, this would not cause any sleepless nights for me.

svenseeberg:
> Not necessarily, they could also craft a malicious website with a button "get 999999$", and when a user clicks this button while having a valid session in the cms, they could perform a post-request to the cms doing arbitrary things.

Yes, I already edited my previous post to include that. But that means we have now 4 requirements for the attack:
- measure network responses
- inject data into requests
- extract the CSRF token in time
- have the user visit an malicious external form

I would not say this is impossible. But I think there are other things we can spend our time on that would improve security much more ^^

svenseeberg:
We also have the page API tokens. Alright alright, I'll disable gzip for all */edit/, user form and login paths ;-)

SteffiStuffel:
Will close this, further discussions in Svens gitea PR.

PR #2099, opened 2023-02-28
Improve caching for endpoints

SteffiStuffel:
### Short description
Improved caching for endpoints

### Proposed changes
- added [caching](https://docs.djangoproject.com/en/4.1/topics/cache/#the-per-site-cache) 
- set cache max-age time to one hour
- allowed cache-control header to be set in the request

### Side effects
should not affect cms sites, tested, there the no-cache header is still set.

### Resolved issues

Fixes: #

### DIFF
--- a/integreat_cms/core/settings.py
+++ b/integreat_cms/core/settings.py
@@ -270,9 +270,11 @@ if DEBUG:
 MIDDLEWARE = [
     "corsheaders.middleware.CorsMiddleware",
     "django.middleware.security.SecurityMiddleware",
+    "django.middleware.cache.UpdateCacheMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.middleware.locale.LocaleMiddleware",
     "django.middleware.common.CommonMiddleware",
+    "django.middleware.cache.FetchFromCacheMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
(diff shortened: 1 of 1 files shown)


SteffiStuffel:
Closed, see [here ](https://github.com/digitalfabrik/integreat-app/pull/888)

timoludwig (commented):
> Closed, see [here ](https://github.com/digitalfabrik/integreat-app/pull/888)

We still could enable caching for the API endpoints if we can live with the content being potentially outdated for some time...

svenseeberg:
> We still could enable caching for the API endpoints if we can live with the content being potentially outdated for some time...

AFAICT editors frequently look at their work in the app. They would probably be confused if they cannot immediately see the results of their work.

timoludwig:
> AFAICT editors frequently look at their work in the app. They would probably be confused if they cannot immediately see the results of their work.

Ok, sounds reasonable. Then let's keep the current setup and maybe only investigate server side caches which can be invalidated as soon as the content is saved?
(I mean technically we could look into [If-Modified-Since](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since) requests, but probably not worth the effort?)

SteffiStuffel:
> > We still could enable caching for the API endpoints if we can live with the content being potentially outdated for some time...
> 
> AFAICT editors frequently look at their work in the app. They would probably be confused if they cannot immediately see the results of their work.

Ye, but we should make an App, which works good for our target group, e.g. using not unnecessary much mobil data. Creators can learn to use a incognito tab.

timoludwig:
On [Mattermost](https://chat.tuerantuer.org/digitalfabrik/pl/19txygsfdpymu8urejiu6qswny), we concluded that this kind of caching won't help us very much, mostly because of these reasons:
1. The endpoints are not fetched often enough that a caching time of 1 hour creates many cache hits (we'd need to start at 24h which could cause serious delay in information transfer).
2. The API response times are not the bottleneck of the app rendering process - we should probably rethink our API design in general and e.g. find ways how we can retrieve only the information that the app needs to render a specific page (e.g. at the moment, it fetches _all_ regions and _all_ languages before it even starts to fetch the pages of one specific region in one specific language).

Instead, we could think about a persistent server side cache which is explicitly invalidated when database objects change - we already have the library https://github.com/Suor/django-cacheops for database-level cache, we could also use this to cache the rendered API results.

Issue #2065, opened 2023-02-13
Only allow staff members to use SUMM.AI bulk action


timoludwig:
### Motivation

We only use the SUMM.AI bulk action in the page tree for initial translations by our team.
Thus, it should also be limited to staff members.
Region managers are supposed to translate pages in SUMM.AI via the new machine translation interface (#1727).

### Proposed Solution

- Either check `is_staff` in the bulk action view or create a new permission and assign it to the service- and cms team
- Hide bulk action for users who are not allowed to use it

JoeyStk:
As a work-around the function was disabled, therefore this issue isn't prio high anymore

timoludwig:
Doesn't this issue contradict our decision to cancel the automatic translation page tree (see #1727)? If we don't have the other automatic translation interface, then normal users should be able to use the bulk action, otherwise they don't have the possibility to translate at all...

PR #2104, opened 2023-03-06
Only allow staff members to use the SUMM AI bulk action

JoeyStk:
### Short description
This PR restricts the usage of the SUMM AI bulk action to staff members only

### Proposed changes

- Add a check in the template if the user is staff member or not
- Add an entry to the changelog to let others teams know that we have improved this (please correct me, if we should remove the entry in case it's not meant to be seen by everyone) 

### Side effects
I think none. I tested it for a few roles and didn't notice any side effects. 
One thing we should keep in mind. As for now the SUMM AI function was disabled globally. So once this PR is merged, we could turn it on again

### Resolved issues

Fixes: #2065 

### DIFF
--- a/integreat_cms/cms/templates/pages/page_tree.html
+++ b/integreat_cms/cms/templates/pages/page_tree.html
@@ -156,7 +156,7 @@
                 <option data-bulk-action="{% url 'bulk_archive_pages' region_slug=request.region.slug language_slug=language.slug %}">
                     {% translate "Archive pages" %}
                 </option>
-                {% if SUMM_AI_ENABLED %}
+                {% if SUMM_AI_ENABLED and request.user.is_staff %}
                     <option data-bulk-action="{% url 'auto_translate_easy_german_pages' region_slug=request.region.slug language_slug=language.slug %}">
                         {% translate "Automatically translate this page into Easy German" %}
                     </option>
--- a/integreat_cms/cms/views/bulk_action_views.py
+++ b/integreat_cms/cms/views/bulk_action_views.py
@@ -6,6 +6,7 @@ import logging
 from django.conf import settings
 from django.contrib import messages
 from django.contrib.auth.mixins import PermissionRequiredMixin
+from django.core.exceptions import PermissionDenied
 from django.http import Http404
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
@@ -169,6 +170,10 @@ class BulkActionEasyGermanView(BulkActionView):
         :return: The redirect
         :rtype: ~django.http.HttpResponseRedirect
         """
+        if not request.user.is_staff:
+            raise PermissionDenied(
+                "Only staff users have the permission to bulk translate pages via SUMM.AI"
+            )
         if not settings.SUMM_AI_ENABLED or not request.region.summ_ai_enabled:
             if not settings.SUMM_AI_ENABLED:
                 logger.warning("SUMM.AI globally disabled")
@@ -178,7 +183,9 @@ class BulkActionEasyGermanView(BulkActionView):
             return super().post(request, *args, **kwargs)
         # Collect the corresponding objects
         logger.info(
-            "%r started SUMM.AI translation for: %r", request.user, self.get_queryset()
+            "%r started SUMM.AI translation for: %r",
+            request.user,
+            self.get_queryset(),
         )
         api_client = SummAiApiClient(request, self.form)
         api_client.translate_queryset(self.get_queryset())
--- a/tests/summ_ai_api/summ_ai_test.py
+++ b/tests/summ_ai_api/summ_ai_test.py
@@ -10,9 +10,6 @@ from integreat_cms.cms.models import Page, Region
 from ..conftest import (
     ANONYMOUS,
     PRIV_STAFF_ROLES,
-    MANAGEMENT,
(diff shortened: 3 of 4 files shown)


timoludwig (commented):
I'm not sure if this is required anymore - if we don't implement #1727, then we need a bulk action for non-staff users...

JoeyStk:
After a quick chat with @dkehne I was reassured that this PR - although it's not very urgent - is still wanted, because it's the cleaner solution to our current workaround. At the moment there is still an edge case in which the municipalities have a short time window in which they theoretically could use the bulk action, although they shouldn't be able to. Sure, this isn't very likely but still at least a theoretic possibility

timoludwig (changes requested):
Ok, but in the long term region managers should be able to use the bulk action, right? We only do the initial translations ourselves, but if they want to update the existing content, they should be able to do this on their own?

Apart from that: Please also implement the check in the bulk action view itself, otherwise this just makes the option invisible, but does not prevent users to use it.

JoeyStk:
Yes, that's how I understood it as well

Okay. I will do so. Thank you!

osmers:
Couldn't we just use the same principle for updating translations that we use for all other DeepL languages? If SUMM is active in the region and allowed for translation via the management section, then Deutsch (leicht) is an option that can be chosen on the page form?

timoludwig:
> Couldn't we just use the same principle for updating translations that we use for all other DeepL languages? If SUMM is active in the region and allowed for translation via the management section, then Deutsch (leicht) is an option that can be chosen on the page form?

@osmers Good idea, I opened #2110 for this.

osmers:
@timoludwig perfect - I also just asked Charlotte to post the latest design and ideas for language selection into the issue #1875 so that that is up to date

integreat_cms/cms/templates/pages/page_tree.html
```diff
@@ -157,9 +157,11 @@ <h1 class="heading">
                     {% translate "Archive pages" %}
                 </option>
                 {% if SUMM_AI_ENABLED %}
-                    <option data-bulk-action="{% url 'auto_translate_easy_german_pages' region_slug=request.region.slug language_slug=language.slug %}">
-                        {% translate "Automatically translate this page into Easy German" %}
-                    </option>
+                    {% if request.user.is_staff %}
```
charludo, line 160: I think you can combine these

charludo (approved):
Very nice, works as expected! Apart from the one small comment, looks good to me! :)

integreat_cms/cms/templates/pages/page_tree.html
```diff
@@ -157,9 +157,11 @@ <h1 class="heading">
                     {% translate "Archive pages" %}
                 </option>
                 {% if SUMM_AI_ENABLED %}
```
charludo, line 159: I think you can combine this check with the one below.

integreat_cms/cms/views/bulk_action_views.py
```diff
@@ -169,20 +169,23 @@ def post(self, request, *args, **kwargs):
...
+                return super().post(request, *args, **kwargs)
+            # Collect the corresponding objects
+            logger.info(
+                "%r started SUMM.AI translation for: %r",
+                request.user,
+                self.get_queryset(),
+            )
+            api_client = SummAiApiClient(request, self.form)
+            api_client.translate_queryset(self.get_queryset())
+            # Let the base view handle the redirect
```
timoludwig, line 188:
```suggestion
        if not request.user.is_staff:
            raise PermissionDenied(
                f"Only staff users have the permission to bulk translate pages via SUMM.AI"
            )
        if not settings.SUMM_AI_ENABLED or not request.region.summ_ai_enabled:
            if not settings.SUMM_AI_ENABLED:
                logger.warning("SUMM.AI globally disabled")
            if not request.region.summ_ai_enabled:
                logger.warning("SUMM.AI disabled in %r", request.region)
            messages.error(request, _("Translations to Easy German are disabled"))
            return super().post(request, *args, **kwargs)
        # Collect the corresponding objects
        logger.info(
            "%r started SUMM.AI translation for: %r",
            request.user,
            self.get_queryset(),
        )
        api_client = SummAiApiClient(request, self.form)
        api_client.translate_queryset(self.get_queryset())
        # Let the base view handle the redirect
```
In my opinion, we should always raise errors when something is not supposed to happen - just silently doing nothing makes debugging very hard when something is not working.
So in this case, raising a permission denied error is the way to go I think.
(requires `from django.core.exceptions import PermissionDenied`)

tests/summ_ai_api/summ_ai_test.py
```diff
@@ -148,6 +148,9 @@ async def test_auto_translate_easy_german(
             response.headers.get("location")
             == f"{settings.LOGIN_URL}?next={translate_easy_german}"
         )
+    elif role in [MANAGEMENT, EDITOR, AUTHOR]:
+        assert response.status_code == 302
+        assert response.headers.get("location") == "/augsburg/pages/de/"
```
timoludwig, line 153:
```suggestion
```
I think the `else` case is much more the expected outcome.

timoludwig (changes requested):

timoludwig (approved):

Issue #2075, opened 2023-02-18
Expert mode: Ability to mark/umark "Ignore Textverständlichkeitsindex" on a content page


dkehne:
### Motivation
Some pages deliberately have a bad HIX value, e.g. bulleted lists or pages that only contain addresses. Expert users should have the possibility to switch off the HIX calculation for a page. These pages should then not be included in the overall HIX calculation (across pages), and at the same time the page should not be translatable via DeepL.

### Proposed Solution
Checkbox on the page-detail page to mark/umark, only visible in expert mode

seluianova:
Ability to ignore calculation of HIX value for certain pages was implemented.

> At the same time the page should not be translatable via DeepL.

Can't do this part because we don't currently use DeepL for pages.
I added this requirement into #2102

PR #2111, opened 2023-03-07
Add ability to ignore calculation of HIX value for certain pages

seluianova:
### Short description
Some pages deliberately have a bad HIX value, e.g. bulleted lists or pages that only contain addresses. Expert users should have the possibility to switch off the HIX calculation for a page.

### Proposed changes
- Add hix_ignore attribute to the page model and "Ignore HIX value" checkbox on the page form
- Calculate HIX value only for pages with hix_ignore=false
- Change the grayscale for the disabled checkboxes because it was hard to recognize if the checkbox was checked or not.

### Side effects
Didn't find any

### Resolved issues
Fixes: #2075

Will do next:
Pages with hix_ignore=true should not be translatable via DeepL.


MizukiTemma:
Thank you for the PR 💪 

I have two question/suggestion (open for all members). 

1) What about hiding the checkbox and show whether the Hix is activated for the page instead of diactivating and graying out the check box, when the user is not allowed to change it? 

2) Schall we check whether the user is an expert user when any change on hix_ignore is included in a request, just like publishing permission check in draft/public change requests? Though this may be an over kill 😅

seluianova:
@MizukiTemma thanks for the feedback :)
I will comment to the first point, because at first I implemented it that way, and then changed it.

Argument 1: 
We already disable other checkboxes when the user does not have edit rights. Such as "Minor edit", for example.

Argument 2:
When the user is not an expert and hix_ignore= true, we could additionally display some text like "HIX calculation is disabled for this page", but then the whole block seems idle to me.
I thought the disabled checkbox looks nicer and clearer.

But I am open to opinions anyway.

integreat_cms/static/src/css/style.scss
```diff
@@ -228,7 +228,7 @@ select,
     @apply rounded text-blue-500 mr-2 align-text-top;
 
     &:disabled {
-        @apply bg-gray-100 border-gray-300;
+        @apply bg-gray-300 border-gray-300;
```
JoeyStk, line 231:
```suggestion
        @apply bg-gray-200 border-gray-200;
```

JoeyStk (changes requested):
Thank you for this PR. To me mostly everything looks good. 

- I think the new greyscale on the disabled checkbox looks a lot better. However, and I should have thought about this earlier when I merged my PR, that it would be good to use the same greyscales throughout the CMS. Therefore I suggest to use grey-200, because we use it for the title bar already (s.screenshot)
- About the discussion for the expert mode. I was reminded, that the expert mode was invented primarily to ease the user experience. Therefore elements that can't be used by non-experts shouldn't be displayed, so user aren't overloaded with elements (I double-checked this with @dkehne)

MizukiTemma:
@seluianova @JoeyStk
Thank you for input and comments 😃 

hmm, to be consistent with the principle "users see only what they can do", should we hide both the ignore_hix check box and the minor edit check box (and others too which are similar to them, if there is any)?

seluianova:
UPD: don't show the checkbox to non-expert users.

JoeyStk (approved):
Thank you for your changes. To me this looks good now :smile:

integreat_cms/cms/forms/pages/page_form.py
```diff
@@ -44,6 +44,11 @@ class PageForm(CustomModelForm, CustomTreeNodeForm):
         required=False,
         label=_("Enable write access via API for this page"),
     )
+    hix_ignore = forms.BooleanField(
+        required=False,
+        label=_("Ignore HIX value"),
+        help_text=_("Tick if HIX value should not be calculated for this page."),
+    )
```
timoludwig, line 51:
```suggestion
```
You only have to define non-model form fields here. All fields which are derived from the model are automatically created if added to the meta class.

integreat_cms/cms/forms/pages/page_form.py
```diff
@@ -89,6 +95,9 @@ def __init__(self, **kwargs):
         self.fields["api_token"].widget.attrs["readonly"] = True
         self.fields["enable_api_token"].initial = bool(self.instance.api_token)
 
+        # HIX value is not ignored by default
+        self.fields["hix_ignore"].initial = False
+
```
timoludwig, line 100:
```suggestion
```
I'd prefer to use the initial value of the model instance here.

integreat_cms/cms/templates/hix_widget.html
```diff
@@ -13,41 +14,51 @@
...
-        </div>
-        <div data-hix-state="outdated" class="bg-yellow-400 rounded p-2 hidden">
-            <i icon-name="alert-circle"></i>
-            {% translate "HIX value is outdated." %}
+    <label>{% translate "Improve the understandability of the page content" %}</label>
+    {% if request.user.expert_mode %}
+        {% render_field page_form.hix_ignore %}
+        <label for="{{ page_form.hix_ignore.id_for_label }}" class="secondary">{{ page_form.hix_ignore.label }}</label>
+        <div class="help-text">{{ page_form.hix_ignore.help_text }}</div>
+    {% elif page_form.hix_ignore.value %}
```
timoludwig, line 22:
```suggestion
    {% elif page_form.instance.hix_ignore %}
```
I think using the model instance here makes it more clear where the value is coming from

integreat_cms/cms/templates/hix_widget.html
```diff
@@ -13,41 +14,51 @@
...
+        <div class="help-text">{% translate "HIX value calulation is disabled for this page." %}</div>
+    {% endif %}
+    <div id="hix-block"
+         class="{% if page_form.hix_ignore.value %}hidden{% endif %}">
```
timoludwig, line 26:
```suggestion
         class="{% if page_form.instance.hix_ignore %}hidden{% endif %}">
```

integreat_cms/static/src/css/style.scss
```diff
@@ -228,7 +228,7 @@ select,
     @apply rounded text-blue-500 mr-2 align-text-top;
 
     &:disabled {
-        @apply bg-gray-100 border-gray-300;
+        @apply bg-gray-200 border-gray-200;
```
timoludwig, line 231:
```suggestion
        @apply bg-gray-200 border-gray-200;
        &:checked {
            @apply bg-gray-400 border-gray-400;
        }
```
I like the idea, but I noticed the contrast between a checked and unchecked disabled checkbox is extremely low, so I suggest to make checked disabled checkboxes a bit darker.

timoludwig (changes requested):
Thanks a lot! :+1:
Works as expected, just have a few minor comments.

integreat_cms/static/src/css/style.scss
seluianova: I already committed another suggestion about the gray-scale from @JoeyStk.
Pls take a look at his comment above.
The last one wins or..? 😀

integreat_cms/cms/forms/pages/page_form.py
seluianova: Good to know 🙈

integreat_cms/static/src/css/style.scss
timoludwig: Yes, I did see that, and I'm not contradicting this suggestion (using `gray-200` for both background and border looks fine to me), but I'd extend this to make sure the difference between checked disabled and unchecked disabled can be perceived...

| | Unchecked | Checked |
| --- | --- | --- |
| Before | Screenshot 2023-03-10 at 13-34-43 Integreat Redaktionssystem| Screenshot 2023-03-10 at 13-34-57 Integreat Redaktionssystem |
| After @JoeyStk's suggestion | Screenshot 2023-03-10 at 13-33-16 Integreat Redaktionssystem | Screenshot 2023-03-10 at 13-33-47 Integreat Redaktionssystem |
| After my suggestion | Screenshot 2023-03-10 at 13-35-18 Integreat Redaktionssystem | Screenshot 2023-03-10 at 13-35-35 Integreat Redaktionssystem |

seluianova: Ah sorry, I misread that

timoludwig (changes requested):
Ah, could you squash the two commits?

timoludwig (approved):

PR #2114, opened 2023-03-11
Add isort

timoludwig:
### Short description

This sorts the python import statements

### Proposed changes

- Add isort to dev dependencies
- Add dev tool for isort
- Apply isort
- Add isort pre-commit hook
- Add isort to CircleCI workflow


charludo (approved):
Awesome! 🎉 I know there already was a warning beforehand, but not having to worry about this anymore is very welcome, thank you!

integreat_cms/cms/models/pois/poi.py
```diff
@@ -1,15 +1,14 @@
-from django.db import models
 from django.core.validators import MaxValueValidator, MinValueValidator
+from django.db import models
 from django.utils.translation import gettext_lazy as _
-
 from linkcheck.models import Link
 
+from ...utils.translation_utils import gettext_many_lazy as __
```
JoeyStk, line 6: I'm not completly understanding why this import is on top of the list, since it's not in alphabetical order :thinking:

integreat_cms/cms/urls/protected.py
```diff
@@ -6,51 +6,50 @@
...
     OrganizationForm,
-    RegionForm,
-    EventTranslationForm,
-    POITranslationForm,
     PageTranslationForm,
+    POITranslationForm,
+    RegionForm,
 )
 from ..models import (
+    POI,
```
JoeyStk, line 19: I understand that POI is on top of this list because it's all caps, but I think I would rather have POI where it was before behind Page and before POICategory. At least that's where I would search for it. But maybe I'm just being picky

JoeyStk (changes requested):
Thank you for your PR. I think all in all a sensible improvement. I looked through all files and to me they look fine. I found a few smaller issues/questions, but maybe they are just on me

tools/code_style.sh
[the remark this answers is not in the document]
timoludwig: Good idea, fixed :+1:

integreat_cms/cms/models/pois/poi.py
timoludwig: There are three categories which are more important than the alphabetical order:
1. Imports from the python standard library go on top (e.g. `logging`, `os`, `sys`, etc.)
2. Imports from third-party packages go second (`django`, `linkcheck`, etc.)
3. Relative imports go third (`from ...x import y` etc.)

JoeyStk: Yes, that I understand. I was just wondering why it was on top of the relative imports, since within the block of relative imports it wasn't ordered alphabetically. But seems to be resolved now anyway!

JoeyStk (approved):
:tada:

integreat_cms/cms/models/pois/poi.py
timoludwig: Ah ok, yeah within the relative imports, the leading dots are counted as well. So
`..b` is sorted before `.a` and so on...

timoludwig: So imports from "far away packages" are sorted first, and then as we descend the list we get nearer to the current code :)

JoeyStk: Okay, that's explains it. Thank you very much!

PR #2115, opened 2023-03-11
Make pylint config more strict

timoludwig:
### Short description

I played around with the pylint config a bit and enabled a few more checks.

### Proposed changes

Enable and fix the following new checks:
- `I0021`: `useless-suppression`
- `W0201`: `attribute-defined-outside-init`
- `W0511`: `fixme`
- `C0325`: `superfluous-parens`
- `R1706`: `consider-using-ternary`
- `C2001`: `compare-to-zero`
- `C1901`: `compare-to-empty-string`
- `R6101`: `consider-using-namedtuple-or-dataclass`
- `R6102`: `consider-using-tuple`
- `R6103`: `consider-using-assignment-expr`
- `R6104`: `consider-using-augmented-assign`

### Side effects

I'm pretty confident that these changes don't affect functionality, so I'd try to focus on maintainability when reviewing this PR.
These new rules definitely improve consistency of the code style, but I'm not completely convinced whether all of these changes really improve the readability. Let me know if you'd like to revert some lines or disable some of these checks entirely.


charludo (commented):
Generally, I think this is a great idea and goes a long way towards cleaning up and unifying the style.

For the following rules however, I'm not completely conviced:
- `consider-using-tuple`: functionally, there appears to be no difference (?) - from the docs: *Only for style consistency! Emitted where an in-place defined ``list`` can be replaced by a ``tuple``. Due to optimizations by CPython, there is no performance benefit from it.* Personally I find the `[]` style a little more readable. But this is a minor thing, changing to `()` is NBD, if you think it benefits the overall style.
- `consider-using-namedtuple-or-dataclass`: all I could find in the changed files were two supressions of this rule, or did I miss something? I find this style to be a *lot* less readable than dicts, and from the documentation it appears there's a need for an additional import (`from typing import NamedTuple`).

timoludwig:
> For the following rules however, I'm not completely convinced

Yes, completely agree with your reasoning :+1:

charludo (approved):
LGTM 🎉

MizukiTemma (approved):
Thank you 🎉

Issue #2118, opened 2023-03-14
Move and rename menu button for automatic translation management


timoludwig:
### Motivation

Since #1727 is no longer planned, there is no need for a separate menu section anymore.
Move the translation management menu button to the bottom.

Also, we could think about removing the "debug mode" constraint and always show the button when at least one MT provider is enabled in that region.

### Current state
![status-quo](https://user-images.githubusercontent.com/16021269/224976137-1310b55a-f267-440f-ae9f-ec6d4764bfc5.png)

### Proposed Solution

![vorschlag](https://user-images.githubusercontent.com/16021269/224976166-932401c0-d0af-4fee-83c1-cad62c569268.png)

I would prefer it if the button could fit into one line, so I'd make the menu a bit wider (17rem?)...

PR #2123, opened 2023-03-14
Move and rename MT management

MizukiTemma:
### Short description

This PR moves and renames the machine translation management section.

### Proposed changes

- Move the section to the settings area of menu bar.
- Rename it to Machine Translations.
- Remove the debug mode constraint.

### Side effects

- I hope none.

### Resolved issues

Fixes: #2118 

### DIFF
--- a/integreat_cms/cms/templates/_base.html
+++ b/integreat_cms/cms/templates/_base.html
@@ -98,10 +98,10 @@
         </div>
     </header>
     <nav id="primary-navigation"
-         class="w-56 h-full fixed left-0 inset-y-0 text-gray-200 bg-gray-700 z-50 transform -translate-x-full lg:translate-x-0 motion-safe:transition-transform">
+         class="w-72 h-full fixed left-0 inset-y-0 text-gray-200 bg-gray-700 z-50 transform -translate-x-full lg:translate-x-0 motion-safe:transition-transform">
         <div class="w-full p-1 h-14 bg-gray-800">
             <a href="{% url 'public:region_selection' %}">
-                <div class="h-full w-full bg-{{ BRANDING }}-logo-white hover:bg-{{ BRANDING }}-logo-hover bg-contain bg-center bg-no-repeat">
+                <div class="h-full w-full bg-{{ BRANDING }}-logo-white hover:bg-{{ BRANDING }}-logo-hover bg-contain ml-2 bg-no-repeat">
                 </div>
             </a>
         </div>
@@ -144,21 +144,6 @@
                             <i icon-name="link"></i>
                             {% translate "Broken Links" %}
                         </a>
-                    {% endif %}
-                {% endif %}
-                <!-- Machine Translation Section -->
-                {% if debug %}
-                    {% if perms.cms.manage_translations %}
-                        <div class="min-h-3 pl-2">
-                            <p class="font-bold">{% translate "Translations" %}</p>
-                        </div>
-                        {% if perms.cms.manage_translations %}
-                            <a href="{% url 'translations_management' region_slug=request.region.slug %}"
-                               class="{% if current_menu_item == 'translations_management' %} active{% endif %}">
-                                <i icon-name='sliders'></i>
-                                {% translate "Manage Translations" %}
-                            </a>
-                        {% endif %}
                     {% endif %}
                 {% endif %}
                 <!-- Content Section -->
@@ -211,11 +196,19 @@
                         {% translate "Imprint" %}
                     </a>
                 {% endif %}
-                {% if perms.cms.view_user or perms.cms.view_languagetreenode %}
+                {% get_mt_visibility request.region perms as show_mt_section %}
+                {% if show_mt_section or perms.cms.view_user or perms.cms.view_languagetreenode %}
                     <!-- Settings Section-->
                     <div class="min-h-3 pl-2">
                         <p class="font-bold">{% translate "Settings" %}</p>
                     </div>
+                {% endif %}
+                {% if show_mt_section %}
+                    <a href="{% url 'translations_management' region_slug=request.region.slug %}"
+                       class="{% if current_menu_item == 'translations_management' %} active{% endif %}">
+                        <i icon-name='bot'></i>
+                        {% translate "Machine Translations" %}
+                    </a>
                 {% endif %}
                 {% if perms.cms.view_organization %}
                     <a href="{% url 'organizations' region_slug=request.region.slug %}"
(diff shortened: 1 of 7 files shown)


integreat_cms/cms/templates/_base.html
```diff
@@ -211,12 +196,19 @@
                         {% translate "Imprint" %}
                     </a>
                 {% endif %}
-                {% if perms.cms.view_user or perms.cms.view_languagetreenode %}
+                {% if perms.cms.manage_translations or perms.cms.view_user or perms.cms.view_languagetreenode %}
                     <!-- Settings Section-->
```
JoeyStk, line 200: As @timoludwig suggest in the issue report I think we should also check if at least one MT provider has been enabled in this region. Therefore I think we should add the check "if DEEPL_ENABLED"

timoludwig: I'd show the entry when at least either DeepL or SUMM.AI are enabled.

integreat_cms/cms/templates/_base.html
```diff
@@ -211,12 +196,19 @@
...
                     <div class="min-h-3 pl-2">
                         <p class="font-bold">{% translate "Settings" %}</p>
                     </div>
                 {% endif %}
+                {% if perms.cms.manage_translations %}
+                    <a href="{% url 'translations_management' region_slug=request.region.slug %}"
+                       class="{% if current_menu_item == 'translations_management' %} active{% endif %}">
+                        <i icon-name='bot'></i>
+                        {% translate "Machine Translations" %}
```
JoeyStk, line 209: I agree with @timoludwig in the issue report, that we should widen the navigation bar, so "machine translation" fits in one row

JoeyStk (changes requested):

integreat_cms/cms/templates/_base.html
JoeyStk: Oh yes, of course. I forgot about SUMM.AI for a second :see_no_evil:

MizukiTemma: Sorry, I overlooked it 🙈 
Now MT providers are checked too.

integreat_cms/cms/templates/_base.html
MizukiTemma: Oh no, I missed this point too 😓 Now it's widened.

integreat_cms/cms/templates/_base.html
```diff
@@ -211,12 +196,19 @@
                         {% translate "Imprint" %}
                     </a>
                 {% endif %}
-                {% if perms.cms.view_user or perms.cms.view_languagetreenode %}
+                {% if SUMM_AI_ENABLED and perms.cms.manage_translations or DEEPL_ENABLED and perms.cms.manage_translation or perms.cms.view_user or perms.cms.view_languagetreenode %}
```
timoludwig, line 199:
```suggestion
                {% get_mt_visibility request.region perms as show_mt_section %}
                {% if show_mt_section or perms.cms.view_user or perms.cms.view_languagetreenode %}
```
I suggest to also add a check whether summ ai is enabled in the current region and to move this complex condition into a template tag (e.g. to [base_filters.py](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/templatetags/base_filters.py)):
```
@register.simple_tag
def get_mt_visibility(region, perms):
    return "cms.manage_translations" in perms and (
        settings.DEEPL_ENABLED or settings.SUMM_AI_ENABLED and region.summ_ai_enabled
    )
```

integreat_cms/cms/templates/_base.html
```diff
@@ -211,12 +196,19 @@
...
                     <!-- Settings Section-->
                     <div class="min-h-3 pl-2">
                         <p class="font-bold">{% translate "Settings" %}</p>
                     </div>
                 {% endif %}
+                {% if SUMM_AI_ENABLED and perms.cms.manage_translations or DEEPL_ENABLED and perms.cms.manage_translations %}
```
timoludwig, line 205:
```suggestion
                {% if show_mt_section %}
```

timoludwig (changes requested):

integreat_cms/cms/templates/_base.html
MizukiTemma: It's a good idea 😃 Thank you for suggestion!

integreat_cms/cms/templatetags/base_filters.py
```diff
@@ -35,3 +36,22 @@ def get_private_member(element, key):
...
+    This tag checks whether either DEEPL or SUMM.AI is activated in the region and the user has permission to manage automatic translations.
+
+    :param region: The rcurrent region
+    :type region: ~integreat_cms.cms.models.regions.region.Region
+
+    :param perms: Permission of the user
+    :type perms: ~django.contrib.auth.context_processors.PermWrapper
+
+    :return: Whether the MT section should be shown
+    :rtype: boolean
```
timoludwig, line 53:
```suggestion
    :rtype: bool
```

integreat_cms/cms/templatetags/base_filters.py
```diff
@@ -35,3 +36,22 @@ def get_private_member(element, key):
...
+@register.simple_tag
+def get_mt_visibility(region, perms):
+    """
+    This tag checks whether either DEEPL or SUMM.AI is activated in the region and the user has permission to manage automatic translations.
+
+    :param region: The rcurrent region
+    :type region: ~integreat_cms.cms.models.regions.region.Region
+
+    :param perms: Permission of the user
+    :type perms: ~django.contrib.auth.context_processors.PermWrapper
```
timoludwig, line 50: Could you add this to Sphinx' nitpicky ignore list?

timoludwig (changes requested):

integreat_cms/cms/templatetags/base_filters.py
MizukiTemma: Applied 👍 Thank you!

timoludwig (approved):

JoeyStk (changes requested):
I noticed one more thing. Since the navigation bar is wider now, it gets very clear that the logo is centered and not aligned with the navigation links. After talking to @timoludwig I would suggest to remove bg-center and replace it with ml-2 [here](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/templates/_base.html#L104). 
grafik

MizukiTemma:
@JoeyStk 

ohhhhn 🤦 now fixed!

JoeyStk (approved):
Awesome! Thank you!

Issue #2116, opened 2023-03-14
`DeepLException` during MT for languages with faulty bcp tag 


JoeyStk:
### Describe the Bug
If you try to use the MT for languages with a faulty bcp tag Django errors occur. This issue is quite relevant, because in our live system quite a few bcp codes are filled out "incorrectly", and therefore we would run into a lot of issues when releasing our implementation in April.

### Steps to Reproduce

1. Go to ''languages" and create a language that a. instead of the bcp tag uses a slug and b. create another language that has an incorrect bcp tag
2. Then click on 'event' or 'locations', ǵo to either one of the newly created languages
3. Select pages and use the bulk action for machine translation
4. In both cases, see the error 
![grafik](https://user-images.githubusercontent.com/72705147/224963832-05dd6e2e-8df1-40c1-8a60-1240969928e5.png)

### Expected Behavior

Even in cases where instead of the bcp code the slug is filled in the translation should work. And for cases in which the bcp code is filled in incorrectly (e.g. pt_PT instead of pt-PT) there should be an informative error message.

### Actual Behavior

In both cases that I described above a django error message appears (see screenshot)

### Additional Information

The underlying problem is probably that we check the language slugs when testing for DeepL availability, but when we're actually translating we're sending the bcp tag. This error also appears while using the translation box in our page form, introduced by #2062

timoludwig:
I could only reproduce this bug for English and Portuguese:
https://github.com/digitalfabrik/integreat-cms/blob/c6a29e4ef0c73ec29641f468b0aabb7b60825ad2/integreat_cms/deepl_api/utils.py#L150-L154

For all other languages, the bcp tag is not used during DeepL translation, right?

PR #2124, opened 2023-03-14
Fixes MT for languages with faulty BCP code

JoeyStk:
### Short description
This PR fixes an error that occurred while using the MT for English or Portuguese. With this PR now it gets checked if a. the BCP47 code is in the correct format as expected, and either returns an error message or for the edge case that the BCP47 tag for English or Portuguese were incorrect sets en-GB respectively pt-PT as default.

### Proposed changes

- Add check if BCP47 code is correct
- Set pt-PT and en-GB as default values if the BCP47 code wasn't correct
- Return an error message for all other cases in which the BCP47 code wasn't correct
- Remove the check from the main function and put it in a separate function in order to pass pylint (too many if-branches in one function 15 out of 12). However I had to add the check "if target_language_key is None" to the main function, which is quite ugly. If someone else has a better idea for this, please let me know :smile: 

### Side effects

- I didn't into any

### Resolved issues

Fixes: #2116

### DIFF
--- a/integreat_cms/cms/templates/languages/language_form.html
+++ b/integreat_cms/cms/templates/languages/language_form.html
@@ -54,7 +54,20 @@
                         </div>
                         <label for="{{ form.bcp47_tag.id_for_label }}" class="block mb-2">{{ form.bcp47_tag.label }}</label>
                         {% render_field form.bcp47_tag|add_error_class:"border-red-500" %}
-                        <div class="mb-2 text-s text-gray-600">{{ form.bcp47_tag.help_text }}</div>
+                        <div class="mb-2 text-s text-gray-600">
+                            {{ form.bcp47_tag.help_text }}
+                            {% if DEEPL_ENABLED %}
+                                <br />
+                                <br />
+                                <i class="h-5" icon-name="alert-triangle"></i>
+                                {# djlint:off #}
+                                {% blocktranslate trimmed with link="href='https://www.deepl.com/de/docs-api/translate-text/translate-text/' target='_blank' class='underline'" %}
+                                    For some languages, DeepL needs a correct BCP47 tag (for example English or Portuguese).
+                                    For more information, please checkout the <a {{ link }}>DeepL documentation</a>.
+                                {% endblocktranslate %}
+                                {# djlint:on #}
+                            {% endif %}
+                        </div>
                     </div>
                 </div>
             </div>
--- a/integreat_cms/cms/views/events/event_list_view.py
+++ b/integreat_cms/cms/views/events/event_list_view.py
@@ -109,7 +109,7 @@ class EventListView(TemplateView, EventContextMixin, SummAiContextMixin):
 
         if settings.DEEPL_ENABLED:
             deepl = DeepLApi()
-            DEEPL_AVAILABLE = deepl.check_availability(request, language_slug)
+            DEEPL_AVAILABLE = deepl.check_availability(request, language)
         else:
             DEEPL_AVAILABLE = False
         MT_PERMITTED = mt_is_permitted(
--- a/integreat_cms/deepl_api/utils.py
+++ b/integreat_cms/deepl_api/utils.py
@@ -27,26 +27,46 @@ class DeepLApi:
         self.translatable_attributes = ["title", "content", "meta_description"]
 
     @staticmethod
-    def check_availability(request, language_slug):
+    def check_availability(request, target_language):
         """
         This function checks, if the selected language is supported by DeepL
 
         :param request: request that was sent to the server
         :type request: ~django.http.HttpRequest
 
-        :param language_slug: current language slug
-        :type language_slug: str
+        :param target_language: transmitted target language
+        :type target_language: ~integreat_cms.cms.models.languages.language.Language
 
         :return: true or false
         :rtype: bool
         """
         deepl_config = apps.get_app_config("deepl_api")
-        source_language = request.region.get_source_language(language_slug)
+        source_language = request.region.get_source_language(target_language.slug)
         return (
             source_language
             and source_language.slug in deepl_config.supported_source_languages
-            and language_slug in deepl_config.supported_target_languages
+            and (
+                target_language.slug in deepl_config.supported_target_languages
+                or target_language.bcp47_tag.lower()
+                in deepl_config.supported_target_languages
+            )
         )
+
+    def get_target_language_key(self, target_language):
+        """
+        This function decides the correct target language key
+
+        :param target_language: the target language
+        :type target_language: ~integreat_cms.cms.models.abstract_content_translation.AbstractContentTranslation
+
+        :return: target_language_key which is 2 characters long for all languages except English and Portugese where the BCP tag is transmitted
+        :rtype: str
+        """
+        deepl_config = apps.get_app_config("deepl_api")
+        for code in [target_language.slug, target_language.bcp47_tag]:
+            if code.lower() in deepl_config.supported_target_languages:
+                return code
+        return None
 
     def check_usage(self, region, source_translation):
         """
@@ -130,6 +150,8 @@ class DeepLApi:
                 )
                 return
 
+            target_language_key = self.get_target_language_key(target_language)
+
             for content_object in content_objects:
                 source_translation = content_object.get_translation(
                     source_language.slug
@@ -146,12 +168,6 @@ class DeepLApi:
 
                 existing_target_translation = content_object.get_translation(
                     target_language.slug
-                )
-                # For some languages, the DeepL client expects the BCP tag instead of the short language code
-                target_language_key = (
-                    target_language.bcp47_tag
-                    if target_language.slug in ("en", "pt")
-                    else target_language.slug
                 )
 
                 # Before translating, check if translation would exceed usage limit
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -5128,6 +5128,17 @@ msgid "or"
 msgstr "oder"
 
 #: cms/templates/languages/language_form.html
+#, python-format
+msgid ""
+"For some languages, DeepL needs a correct BCP47 tag (for example English or "
+"Portuguese). For more information, please checkout the <a %(link)s>DeepL "
+"documentation</a>."
+msgstr ""
+"Für einige Sprachen benötigt DeepL den korrekten BCP47-Tag (zum Beispiel für "
+"Englisch oder Portugiesisch). Für weitere Informationen lesen Sie bitte   "
+"die <a %(link)s>DeepL Dokumentation</a>."
+
+#: cms/templates/languages/language_form.html
 msgid "Translations"
 msgstr "Übersetzungen"
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ UNRELEASED
 * [ [#2048](https://github.com/digitalfabrik/integreat-cms/issues/2048) ] Add possibility to show media file usages in sidebar
 * [ [#1875](https://github.com/digitalfabrik/integreat-cms/issues/1875) ] Add automatic translation options to Pages, Events and Locations
 * [ [#2118](https://github.com/digitalfabrik/integreat-cms/issues/2118) ] Move and rename menu button for automatic translation management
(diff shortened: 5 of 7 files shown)


integreat_cms/deepl_api/utils.py
```diff
@@ -48,6 +48,42 @@ def check_availability(request, language_slug):
...
+            else:
+                messages.error(
+                    request,
+                    _("Your bcp code for the selected language seems to be incorrect."),
+                )
+                target_language_key = None
+        else:
+            target_language_key = target_language.slug
+        return target_language_key
+
```
charludo, line 86:
```suggestion
        if target_language.slug not in ["en", "pt"]:
            return target_language.slug

        if len(target_language.bcp47_tag) == 5 and "-" in target_language.bcp47_tag:
            return target_language.bcp47_tag

        defaults = {
            "en": "en-GB",
            "pt": "pt-PT",
        }
        return defaults.setdefault(target_language.slug, None)
```
I think this might be a little better readable, and is easily extendable should the need arise.

I'd then move the error message into the `deepl_translation`  function (see below)

integreat_cms/deepl_api/utils.py
```diff
@@ -48,6 +48,42 @@ def check_availability(request, language_slug):
             and language_slug in deepl_config.supported_target_languages
         )
 
+    def get_target_language_key(self, request, target_language):
+        """
+        This function decides the correct target language key
+
+        :param request: the transmitted request
+        :type request: ~django.http.HttpRequest
```
charludo, line 56:
```suggestion
```
then the request also does not need to be passed any longer

integreat_cms/deepl_api/utils.py
```diff
@@ -147,12 +183,11 @@ def deepl_translation(self, request, content_objects, language_slug, form_class)
...
                 )
-                # For some languages, the DeepL client expects the BCP tag instead of the short language code
-                if target_language.slug in ("en", "pt"):
-                    target_language_key = target_language.bcp47_tag
-                else:
-                    target_language_key = target_language.slug
-
+                target_language_key = self.get_target_language_key(
+                    request, target_language
+                )
```
charludo, line 188:
```suggestion
                target_language_key = self.get_target_language_key(target_language)
```

charludo (changes requested):
Thanks a lot! 🎉 

I only have a couple suggestions, mostly to make `get_target_language_key` a little more succinct.

integreat_cms/deepl_api/utils.py
```diff
@@ -147,12 +183,11 @@ def deepl_translation(self, request, content_objects, language_slug, form_class)
...
+                if target_language_key is None:
+                    break
```
charludo, line 190:
```suggestion
                if not target_language_key:
                    # re-add error message here
                    return
```

I would also move this up and out of the loop, right below `target_language = ...`

integreat_cms/deepl_api/utils.py
timoludwig: @charludo I like your simpler version of the code!
However, I think in this case we don't even need the hard-coded fallback mechanism - I would suggest to handle this case already at the "check availability"-stage and just not enable DeepL translations for en or pt if the bcp tag is faulty. I think this is too much of an edge case to provide detailed user feedback here. Also, the assumption that a valid bcp tag always has 5 characters is wrong...

```suggestion
        deepl_config = apps.get_app_config("deepl_api")
        for code in [target_language.slug, target_language.bcp47_tag]:
            if code in deepl_config.supported_source_languages:
                return code
        return None

```

My suggestion would be to make this completely language agnostic and just remove the `[:2]` here: https://github.com/digitalfabrik/integreat-cms/blob/e90baf5755c9ef18e7b943c6308c9c37b8ec1426/integreat_cms/deepl_api/apps.py#L51

This would mean we would store the bcp tag in the `supported_source_languages` config and could then check whether either the slug or the bcp tag are supported.

This would also require the `check_availability()` function to be rewritten to also check for the bcp tag if the slug is not valid.

timoludwig (changes requested):

integreat_cms/deepl_api/utils.py
```diff
@@ -61,28 +61,11 @@ def get_target_language_key(self, request, target_language):
...
-                    request,
-                    _("Your bcp code for the selected language seems to be incorrect."),
-                )
-                target_language_key = None
-        else:
-            target_language_key = target_language.slug
-        return target_language_key
+        deepl_config = apps.get_app_config("deepl_api")
+        for code in [target_language.slug, target_language.bcp47_tag]:
+            if code in deepl_config.supported_source_languages:
```
charludo, line 66:
```suggestion
            if code in deepl_config.supported_target_languages:
```
@timoludwig shouldn't it be this? (Not sure if those two are the same, but since we already differentiate, just use `target` for consistency's sake)

timoludwig: They are indeed not identical, the source languages do not differentiate between the bcp tags...

integreat_cms/deepl_api/utils.py
JoeyStk: Is there a preferred way on how I should go about it? Would it be better if I added a second parameter for the bcp code or should I rather change the language_slug parameter to a language parameter, that contains both slug and bcp47 code as attributes?

timoludwig: I think one single language parameter makes the most sense...

integreat_cms/cms/templates/_base.html
```diff
@@ -101,7 +101,7 @@
          class="w-56 h-full fixed left-0 inset-y-0 text-gray-200 bg-gray-700 z-50 transform -translate-x-full lg:translate-x-0 motion-safe:transition-transform">
         <div class="w-full p-1 h-14 bg-gray-800">
             <a href="{% url 'public:region_selection' %}">
-                <div class="h-full w-full bg-{{ BRANDING }}-logo-white hover:bg-{{ BRANDING }}-logo-hover bg-contain bg-center bg-no-repeat">
+                <div class="h-full w-full bg-{{ BRANDING }}-logo-white hover:bg-{{ BRANDING }}-logo-hover ml-2 bg-contain bg-no-repeat">
```
timoludwig, line 104: This change does not belong into this PR, does it? Could you rebase on develop again?

integreat_cms/cms/templates/languages/language_form.html
```diff
@@ -54,7 +54,13 @@ <h3 class="font-bold text-black">
                         </div>
                         <label for="{{ form.bcp47_tag.id_for_label }}" class="block mb-2">{{ form.bcp47_tag.label }}</label>
                         {% render_field form.bcp47_tag|add_error_class:"border-red-500" %}
-                        <div class="mb-2 text-s text-gray-600">{{ form.bcp47_tag.help_text }}</div>
+                        <div class="mb-2 text-s text-gray-600">
+                            {{ form.bcp47_tag.help_text }}
+                            </ br>
+                            {% if DEEPL_ENABLED %}
+                                <i class="mr-2" icon-name="alert-triangle"></i>{% translate "For some language DeepL needs a correct BCP47 code (for example English or Portuguese). For exact information which languages require this, please checkout the" %} <a href="https://www.deepl.com/de/docs-api/translate-text/translate-text/">{% translate "DeepL documentation" %}</a>.
```
timoludwig, line 61:
```suggestion
                            {% if DEEPL_ENABLED %}
                                <br />
                                <br />
                                <i class="h-5" icon-name="alert-triangle"></i>
                                {# djlint:off #}
                                {% blocktranslate trimmed with link="href='https://www.deepl.com/de/docs-api/translate-text/translate-text/' target='_blank' class='underline'" %}
                                    For some languages, DeepL needs a correct BCP47 code (for example English or Portuguese).
                                    For more information, please checkout the <a {{ link }}>DeepL documentation</a>.
                                {% endblocktranslate %}
                                {# djlint:on #}
```
Two problems with this part:
1. `</ br>` is an incorrect tag
2. Please do not split translated text into chunks - always keep the text as a whole to make it possible to translate the sentence in context, and also the position of a specific part of the sentence may be different in other languages.

integreat_cms/cms/views/events/event_list_view.py
```diff
@@ -109,7 +109,7 @@ def get(self, request, *args, **kwargs):
 
         if settings.DEEPL_ENABLED:
             deepl = DeepLApi()
-            DEEPL_AVAILABLE = deepl.check_availability(request, language_slug)
+            DEEPL_AVAILABLE = deepl.check_availability(request, region.default_language)
```
timoludwig, line 112:
```suggestion
            DEEPL_AVAILABLE = deepl.check_availability(request, language)
```
Why the default language here?

timoludwig (changes requested):

integreat_cms/deepl_api/utils.py
```diff
@@ -27,27 +27,46 @@ def __init__(self):
...
         deepl_config = apps.get_app_config("deepl_api")
-        source_language = request.region.get_source_language(language_slug)
+        source_language = request.region.get_source_language(target_language.slug)
         return (
             source_language
             and source_language.slug in deepl_config.supported_source_languages
-            and language_slug in deepl_config.supported_target_languages
+            and (
+                target_language.slug in deepl_config.supported_target_languages
+                or target_language.bcp47_code in deepl_config.supported_target_languages
```
timoludwig, line 50:
```suggestion
                or target_language.bcp47_tag.lower() in deepl_config.supported_target_languages
```

integreat_cms/deepl_api/utils.py
```diff
@@ -27,27 +27,46 @@ def __init__(self):
...
         )
 
+    def get_target_language_key(self, target_language):
+        """
+        This function decides the correct target language key
+
+        :param target_language: the target language
+        :type target_language: ~integreat_cms.cms.models.abstract_content_translation.AbstractContentTranslation
+
+        :return: target_language_key which is 2 characters long for all languages except English and Portugese where the BCP code is transmitted
```
timoludwig, line 61:
```suggestion
        :return: target_language_key which is 2 characters long for all languages except English and Portugese where the BCP tag is transmitted
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -5127,6 +5127,17 @@ msgstr "Das ist in vielen Fällen ein einzelner (Makro-)Sprachen-Tag aus der"
 msgid "or"
 msgstr "oder"
 
+#: cms/templates/languages/language_form.html
+#, python-format
+msgid ""
+"For some languages, DeepL needs a correct BCP47 code (for example English or "
```
timoludwig, line 5133:
```suggestion
"For some languages, DeepL needs a correct BCP47 tag (for example English or "
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -5127,6 +5127,17 @@ msgstr "Das ist in vielen Fällen ein einzelner (Makro-)Sprachen-Tag aus der"
...
+"Portuguese). For more information, please checkout the <a %(link)s>DeepL "
+"documentation</a>."
+msgstr ""
+"Für einige Sprachen benötigt DeepL den korrenten BCP47 Code (zum Beispiel "
```
timoludwig, line 5137:
```suggestion
"Für einige Sprachen benötigt DeepL den korrekten BCP47-Tag (zum Beispiel "
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -5127,6 +5127,17 @@ msgstr "Das ist in vielen Fällen ein einzelner (Makro-)Sprachen-Tag aus der"
...
+"für Englisch oder Portugiesisch). Für weitere Informationen lesen Sie bitte  "
+"<a %(link)s> die DeepL documentation</a>."
```
timoludwig, line 5139:
```suggestion
" die <a %(link)s>DeepL Dokumentation</a>."
```

timoludwig (changes requested):

integreat_cms/deepl_api/utils.py
```diff
@@ -27,27 +27,46 @@ def __init__(self):
...
+        :rtype: str
+        """
+        deepl_config = apps.get_app_config("deepl_api")
+        for code in [target_language.slug, target_language.bcp47_tag]:
+            if code in deepl_config.supported_target_languages:
```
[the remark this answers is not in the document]
JoeyStk: Oh, that was what was missing! Thank you!

Issue #1383, opened 2022-05-03
Update internal links when cloning regions


svenseeberg:
### Motivation
When a region is cloned, all links in the content still point to pages in the old region. However, in almost all cases these links should point to pages in the newly created region. We're already planing to have a centralized search & replace function. This function can be used to update all pages when cloning a region. This prevents any human errors that could occur when manually replacing the URLs.

### Proposed Solution
Call the search & replace function created in #1382 to replace the base URL + region slug for all pages (including all versions if easily possible) without adding new page versions. A minor edit should be okay.

timoludwig:
This either depends on #1414 or the media links have to be excluded from the replacement, otherwise the images will be broken.

timoludwig:
> This either depends on #1414 or the media links have to be excluded from the replacement, otherwise the images will be broken.

My earlier statement is not true since media links use the region id and not the region slug in links. So replacing the region slug should be fine.

PR #2126, opened 2023-03-16
Update internal links when cloning regions

timoludwig:
### Short description

This adds a few utils for link replacements and applies the replacement mechanism on cloned regions.

### Proposed changes

- Add utility function to replace links in content
- Update internal links when cloning regions
- Add management command to replace links in content

### Side effects

Well, testing this was quite a journey. The main problem was that our normal test cases not not allow interactions with the database, and I noticed very weird side effects from enabling transaction support only for few test cases.
In particular, the problem was that the transaction test cases roll back not only the changes from the particular test case, but also remove the test data including the roles that are generated during the migrations.
I think I now got around them by adding the extension `pytest-order` and telling pytest it should execute the transaction test cases last. Also, I needed to provide another test fixture just for roles to make sure I can re-populate the database after it has been rolled back by another transaction test case.

### Resolved issues

Fixes: #1383


MizukiTemma (approved):
Thank you for PR 😃 This is a very nice feature!
Link replace works as expected during region cloning and by command. 


I just noticed one case (this is not your fault, though).
When the source regions has no impressum, `replace_links` will not be run. In this case the error page will be shown but a new region will be created with not updated links. I think this happens with source reagions without page (or language_tree) too.

How to reproduce:
1) Go to the region Nürnberg (default data, no impressum).
2) Put some internal links in a page.
3) Clone Nürnberg.
4) See unupdated links in the page in the new region.

Please dismiss this if an error page is enough or we can assume that a source region has always language_tree, page and impressum.

timoludwig:
@MizukiTemma Thanks a lot for the report, since this is a separate problem that already exists on the develop branch, I opened issue #2153 for this and fixed it in #2154.

Issue #1763, opened 2022-10-15
Add changelog building step to release workflow


timoludwig:
### Motivation

With our growing number of contributors, resolving conflicts in the changelog file becomes more and more effort.

### Proposed Solution

Add changelog entries as separate files first and assemble them later in the release workflow, similar to [how the app team is doing it](https://github.com/digitalfabrik/integreat-app/tree/main/release-notes/unreleased)

### Alternatives

- Add a specific line to a commit message body, e.g. `[changelog]`, and then write a script which automatically uses the commit message subject as changelog entry of the issue which is fixed by that commit
- Add a section to PR descriptions which can be parsed by a script and assembled into a changelog file
- Completely automate the changelog and generate one entry for each fixed issue since the last release

PR #2131, opened 2023-03-20
Add release notes bundling tool

timoludwig:
### Short description

I was really annoyed by the changelog format and the constant merge conflicts, so I implemented a completely over-engineered release note tool.

To generate the good old markdown changelog, execute e.g.:
```
./tools/make_release_notes.sh --format=md --output=CHANGELOG.md
```

You can create new release notes by copying+pasting the existing ones, or by simply executing:
```
./tools/new_release_note.sh 1689 en "Changelog entries are now automatically genereated"
```
```
Checking if Integreat CMS is installed...
✔ Integreat CMS is installed
Creating new release note...
✔ The new release note has been generated:

integreat_cms/release_notes/current/unreleased/1689.yml
┌──────────────────────────────────────
│ en: Changelog entries are now automatically genereated
│ de: Changelog-Einträge werden nun automatisch generiert
└──────────────────────────────────────
```
or:
```
./tools/new_release_note.sh 1763 de "Das Änderungsprotokoll steht jetzt auch auf Deutsch zur Verfügung"
```
```
Checking if Integreat CMS is installed...
✔ Integreat CMS is installed
Creating new release note...
✔ The new release note has been generated:

integreat_cms/release_notes/current/unreleased/1763.yml
┌──────────────────────────────────────
│ en: The change log is now also available in German
│ de: Das Änderungsprotokoll steht jetzt auch auf Deutsch zur Verfügung
└──────────────────────────────────────
```
And if a DeepL auth key is set in the config file, the changelog entries are automatically translated.

### Proposed changes

- Split changelog into separate files
- Remove obsolete changelog checking script
- Add dev tool to build release notes
- Add dev tool to create new release note
- Adapt documentation script to new release notes
- Adapt CircleCI workflow to new release notes
- Show release notes in CMS

### Side effects

- Whether the release workflow works as expected remains to be seen

### Resolved issues

Fixes: #1689
Fixes: #1763


CHANGELOG.md
charludo, line 1: Is this intentional? If so, we should probably add the changelog to `.gitignore`

tools/make_release_notes.sh
charludo, line 1: I have a couple of notes on the (default) behaviour and conventions here:
- to keep with `generate_documentation.sh`, maybe it would make sense to rename this script to `generate_release_notes.sh`? (just preference though)
- executing only `./tools/make_release_notes.sh` currently simply echoes the entire changelog. I don't see how this is useful. My suggestion would be to either:
    - only print the unreleased section, so we do not have to scroll endlessly up to see the entries we're most likely most interested in, and add a `--all` flag to get the current behaviour
    - **or** instead of echoing it, somehow open the output with `less` to make it easily navigatable and jump to the top of the output initially
    - **or** (and this is what I would prefer for daily workflow) make `--format=md --output=CHANGELOG.md` the default behaviour if no options are given. Then simply add  `--dry` and `--all` flags to echo only the unreleased section, or all sections in the console.

If it is worth the effort to implement these changes, I don't know; if we're not going to create the changelog ourselves before commiting (see the question about `.gitignore`) and this script is only used by CircleCI workflows, then probably not.

CHANGELOG.md
timoludwig: Yes, it was intentional, but it's open for debate.
Do we want the file for people who visit our repo and want a quick overview without having to generate the file itself or look into the [documentation](https://digitalfabrik.github.io/integreat-cms/changelog.html)?
Or do we want to omit the file because it's redundant information?
Whether or not it belongs in the `.gitignore` depends on how often we're going to generate it - I personally see no use case to do this a lot locally, I guess I would just look into the docs or the new page in the CMS when I want the complete list.

charludo (changes requested):
Very nice! 🎉 This will indeed prevent many merge conflicts 😄 
Just a couple of notes:

CHANGELOG.md
charludo: Hm, it would still be nice to have the log visible in the repo, I think. Maybe an additional pre-commit hook?

tools/make_release_notes.sh
timoludwig: > * to keep with `generate_documentation.sh`, maybe it would make sense to rename this script to `generate_release_notes.sh`? (just preference though)

I agree with the idea, but I would rather rename the docs script to `make_docs.sh` if that's ok?

> * executing only `./tools/make_release_notes.sh` currently simply echoes the entire changelog. I don't see how this is useful. My suggestion would be to either:
>       
>   * only print the unreleased section, so we do not have to scroll endlessly up to see the entries we're most likely most interested in, and add a `--all` flag to get the current behaviour
>   * **or** instead of echoing it, somehow open the output with `less` to make it easily navigatable and jump to the top of the output initially
>   * **or** (and this is what I would prefer for daily workflow) make `--format=md --output=CHANGELOG.md` the default behaviour if no options are given. Then simply add  `--dry` and `--all` flags to echo only the unreleased section, or all sections in the console.

I guess I chose "raw" and "all" as default values because that's how the script is used in the circleci workflow (it's easier to redirect stdout into a bash variable than generating a file and then read from it), but I agree the default values should match the daily use case, since the cirlceci config only has to be written once :grin:
Probably, I'd only print the latest version most of the time, and not the whole history...

CHANGELOG.md
timoludwig: But if we keep it in the repo and modify it with each PR, this would defy the whole reason for a new tool, because we'd have the same merge conflicts as before (although it would be a little bit easier to resolve them)...
If we decide to keep it in the repo, I would only update it with every automated release commit (so it would not contain the unreleased section)...

charludo: Oh... Good point. 

> I would only update it with every automated release commit (so it would not contain the unreleased section)...

Either that, or could we write an action to generate the changelog whenever a PR is merged into `develop`? Since that appears to be the default branch when visiting the repo, so this is where visitors would look for the log.

charludo: > Maybe just a small placeholder file which links to the documentation?
> Or a symlink to the `ìntegreat_cms/release_notes` directory?

Those would work, yes! You're right, too much effort wouldn't be justified here.

timoludwig: Hmm, symlinks don't quite work in the UI the way I hoped (clicking  on it just opens https://github.com/digitalfabrik/integreat-cms/blob/changelog-bundling/release-notes)...

Maybe a badge which links to the release notes?

[Release Notes](https://digitalfabrik.github.io/integreat-cms/changelog.html)

tools/make_release_notes.sh
timoludwig: I tried to change the behavior as described above, hope I didn't mess up :sweat_smile:

CHANGELOG.md
charludo: Sure, why not! As long as the changelog is easy to find, any solution is probably fine.

tools/make_release_notes.sh
charludo: Works like a charm! 🎉

charludo (approved):

MizukiTemma (approved):
Thank you 👍

timoludwig:
> So, where exactly is the config file again? A quick search in the documentation doesn't bring anything up. Is this a missing section?

In `/etc/integreat-cms.ini`

There's a small sentence here: https://digitalfabrik.github.io/integreat-cms/ref/integreat_cms.core.html#settings

But yeah, could probably be more prominent. And I'm not sure why the search isn't returning this result...

Issue #2101, opened 2023-03-03
Store the HIX value of pages in the database


charludo:
### Motivation
We want to prevent the automatic translation of hard to understand page texts, where hardness is measured by the HIX score. Calculation of this score is too slow/expensive to do "on the fly" for large numbers of pages, so instead we need to store it in the database.

### Proposed Solution

- [x] Store the score in a new field on the page model and recalculate and update it every time the page is saved.
- [ ] Add management command to populate the hix values of all page translations which have not been saved yet

### Alternatives

Store the entire history of HIX scores per page, opening up analytics possibilities. 

### Additional Context

- Quick side note from a chat with @dkehne. The automatic translations should be possible if the HIX value is greater or equal to 15 and not greater than 15. (Author: @JoeyStk, March 6th 2023)

### Blocking:
- #2088 
- #2102

david-venhoff:
Johannes and I have thought a bit about possible implementations for this issue, but we are not really sure how to proceed.

What we thought about:
- Store the hix value in the database when the user clicks on update in the hix widget: This would probably cause many pages to have no hix score, so automatic translations are not possible / only possible after the user manually updates every hix score
- Store the hix value when the user updates a page: This could make saving a page quite slow, since we would have to wait for the hix api to answer
- Calculate the hix score in the background: probably the simplest solution for users, but could be difficult to implement. We might need to have a persistent queue of pages, a bit similar to the link checker.

Are there maybe better solutions than the ones listed above? Or do you have a preference for one of these?

timoludwig:
In order for this to work, we need two components that are separate from one another:

1. We need a possibility to populate the database with all hix values, independently from their saving process. For this task, I suggest implementing a management command (after #2071 is merged) which iterates over all pages and queries the hix value (ideally, this would make use of aiohttp to query multiple values at the same time, not sure what the rate limits are here, but we can play with this around later). We could then start this command e.g. on a weekend to avoid putting unnecessary strain on the server.

2. We need a possibility to update the hix value after a page is saved, I like your idea with the background task similar to the link checker. I'm not even sure if we even need a persistent queue... Couln't we just add a normal post save signal handler that queries the hix score and stores it? There won't be much race conditions like in the link checker, because the hix values of pages are completely independently from another (unlike the link checker where many links can point to the same URL)...

I also think these two components can be developed separately, maybe 2. first and then 1. only needs to calculate those values that are missing after some time (when we decide to launch the MT feature)...

[...]

PR #2135, opened 2023-03-22
Fix usage of new DeepL availability check

timoludwig:
### Short description

This should fix all the remnants we missed in #2124.

### Proposed changes

- Fix usage of new DeepL availability check

### Resolved issues

Fixes: #2134

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #2138, opened 2023-03-22
Store hix value in the database

david-venhoff:
### Short description

This pr stores the hix values of page translations in the database. 

### Proposed changes

- Add a Page translation signal hook that updates the hix value for changed translations

### Side effects

- Currently, the score is not updated after a save if it has been calculated before. This could become a problem when the content of a page translation changes without creating a new version
- The score is calculated for all page translations, including drafts. I don't think that this is a issue as the information might also be useful but I can change this behavior if not wanted

### Resolved issues

Partially resolves issue #2101


integreat_cms/core/signals/hix_signals.py
```diff
@@ -0,0 +1,56 @@
...
+    :type instance: ~integreat_cms.cms.models.pages.page_translation.PageTranslation
+
+    :param \**kwargs: The supplied keyword arguments
+    :type \**kwargs: dict
+    """
+    if instance.hix_score is not None:
+        return
+
+    if not hix_enabled(instance):
+        return
```
charludo, line 28:
```suggestion
    if instance.hix_score or not hix_enabled(instance):
        return
```
Unless I'm mistaken here, is there a special reason to compare with `None`?

I'm also not really clear on why we need to ckeck if the score exists already, is it saved multiple times otherwise?

integreat_cms/core/signals/hix_signals.py
```diff
@@ -0,0 +1,56 @@
...
+
+    if score := lookup_hix_score(instance.content) is not None:
```
charludo, line 30:
```suggestion
    if score := lookup_hix_score(instance.content):
```
same here

timoludwig (commented):
> HIX is only for German, right? Then it may be better to store HIX-value in the page model not in the translation model so translations of other languages don't have the field unnecessarly.

Hmm, good question, however I would prefer to leave it in the translation model, for two reasons:
1. I think having a clear model is more important than saving a few bits of storage in the database, and the hix score is clearly connected to the German version of a page, so it makes more sense in the translation model
2. Storing the hix score not only per language, but also per version allows us to track the hix score over time and show e.g. charts how it developed... could be a cool feature for our okr metrics.

charludo (changes requested):
Thank you for getting this done so quickly! 🎉 

I have a couple of small nitpicks, see below.

Please rebase again after #2135 has been merged, that will make testing everything out a little easier 😄

integreat_cms/core/signals/hix_signals.py
david-venhoff: If we don't check whether the score was already calculated, the server ends up in an infinite loop, because calculating and saving the hix score causes the save signal handler to run again. Without the check it would calculate the hix score again, save it and then run again due to the save.

I just compared to None because I find it easier understandable. Do we have a guideline whether to use `is not None` or the implicit version? It seems like we use both variants in the codebase

charludo: > If we don't check whether the score was already calculated, the server ends up in an infinite loop, because calculating and saving the hix score causes the save signal handler to run again. Without the check it would calculate the hix score again, save it and then run again due to the save.

Ah, gotcha. That makes sense.


> I just compared to None because I find it easier understandable. Do we have a guideline whether to use `is not None` or the implicit version? It seems like we use both variants in the codebase

Don't know if we have guidelines, and there are cases when `if A` =/= `if A is not None`, but I did not see that here.

TBH I feel like with your explanation about score re-calculation it makes even more intuitive sense to say "if the score exists, return" 😅 But ultimately it's just a style preference.

timoludwig: @david-venhoff In my opinion, we should use `is not None` only if we want to differentiate between `False`, `0`, `""` (or other falsy values) and `None`. In all cases where it doesn't make a difference, I'd prefer `not x`.
Especially when handling with the Django ORM, we can safely assume that any object that is not `None` automatically evaluates to any truthy value, and I would also apply this pattern on our own custom methods.

integreat_cms/core/signals/hix_signals.py
```diff
@@ -0,0 +1,53 @@
...
+    r"""
+    Calculates a hix store for a page translation after saving
+
+    :param instance: The page translation that gets saved
+    :type instance: ~integreat_cms.cms.models.pages.page_translation.PageTranslation
+
+    :param \**kwargs: The supplied keyword arguments
+    :type \**kwargs: dict
+    """
+    if instance.hix_score or not hix_enabled(instance):
```
timoludwig, line 24:
```suggestion
    if kwargs.get("raw") or instance.hix_score or not hix_enabled(instance):
```
We have to ignore raw save signals (e.g. during fixture loading), because they will cause errors when loading the test data into a fresh database.

> `raw`
> 
> A boolean; `True` if the model is saved exactly as presented (i.e. when loading a fixture). One should not query/modify other records in the database as the database might not be in a consistent state yet.

– https://docs.djangoproject.com/en/3.2/ref/signals/#post-save

integreat_cms/core/signals/hix_signals.py
```diff
@@ -0,0 +1,53 @@
...
+        return
+
+    if score := lookup_hix_score(instance.content):
+        logger.debug(
+            "Storing hix score %s for new page translation %s", score, instance.title
```
timoludwig, line 29:
```suggestion
            "Storing hix score %s for new page translation %r", score, instance
```
I'd prefer the repr for logging instead only the title.

integreat_cms/core/signals/hix_signals.py
```diff
@@ -0,0 +1,53 @@
...
+        )
+        instance.hix_score = score
+        instance.save(update_timestamp=False)
+    else:
+        logger.warning(
+            "Could not store the hix score for new page translation %s", instance.title
```
timoludwig, line 35:
```suggestion
            "Could not store the hix score for new page translation %r", instance
```
This makes debugging a bit easier

integreat_cms/cms/forms/pages/page_translation_form.py
```diff
@@ -20,4 +20,4 @@ class Meta:
         #: The model of this :class:`django.forms.ModelForm`
         model = PageTranslation
         #: The fields of the model which should be handled by this form
-        fields = CustomContentModelForm.Meta.fields + ["slug"]
+        fields = CustomContentModelForm.Meta.fields + ["slug", "hix_score"]
```
timoludwig, line 23: This line currently has no effect, right?
But ok, since we're probably adding this field to the form at some point in the future anyway, I guess it can stay.

timoludwig (changes requested):
Thanks a lot! :+1:
Works as expected, except when loading the test data into a new database.

charludo (approved):
🎉  Awesome, everything (apart from what TImo mentioned ofc) is working as far as I can tell!

PR #2155, opened 2023-03-24
Small improvements for HIX API client

timoludwig:
### Short description

Small improvements for HIX API client

### Proposed changes

- Allow long text for post save signals (the limitation only applies to the ajax view)
- Catch the urllib error in the API client and not in the view utility (The previous solution did not log the HTTPError)
- Do not execute post save signal when the content is empty


david-venhoff (approved):
LGTM

Issue #2137, opened 2023-03-22
Outdated machine translations not marked as outdated


timoludwig:
### Describe the Bug

Machine translations are marked with the "bot" icon indefinitely, even if they're outdated.

### Steps to Reproduce

1. Create machine translation
2. Change source translation without updating the bot translation
3. See bot icon

### Expected Behavior

Outdated translations should be flagged as outdated, no matter where they come from.
The bot icon should only be used for machine translations that are up to date.

### Actual Behavior

Currently, the bot icon overrides the outdated icon

PR #2157, opened 2023-03-24
Fix translation status for outdated mts

JoeyStk:
### Short description
The translation status for machine translations used to still show the bot icon even when the translation wasn't up-to-date anymore. This PR fixes this and only shows the bot icon for machine translations that are up-to-date and shows the warning icon for machine translations that are outdated. 

### Proposed changes

- Adds distinction between up-to-date and outdated machine translations and decides for the correct translation status and icon

### Side effects

- I restricted the query for machine translations and this could cause some side effects, although I don't have an exact idea how this might show 

### Resolved issues

Fixes: #2137

### DIFF
--- a/integreat_cms/cms/models/abstract_content_translation.py
+++ b/integreat_cms/cms/models/abstract_content_translation.py
@@ -436,25 +436,26 @@ class AbstractContentTranslation(AbstractBaseModel):
             # If the page does not have a major public version, it is considered "missing" (keep in mind that it might
             # have draft versions or public versions that are marked as "minor edit")
             return translation_status.MISSING
-        if translation.machine_translated:
-            return translation_status.MACHINE_TRANSLATED
         if translation.currently_in_translation:
             return translation_status.IN_TRANSLATION
         if not self.source_language:
             # If the language of this translation is the root of this region's language tree, it is always "up to date"
             return translation_status.UP_TO_DATE
-        if not (source_translation := self.major_source_translation):
+        if (
             # If the source language does not have a major public version, the translation is considered "outdated",
             # because the content is not in sync with its source translation
-            return translation_status.OUTDATED
-        if source_translation.translation_state == translation_status.OUTDATED:
+            not (source_translation := self.major_source_translation)
             # If the source translation is already outdated, this translation is as well
+            or source_translation.translation_state == translation_status.OUTDATED
+            # If the translation was edited before the last major change in the source language, it is outdated
+            or translation.last_updated <= source_translation.last_updated
+        ):
             return translation_status.OUTDATED
-        if translation.last_updated > source_translation.last_updated:
-            # If the translation was edited after the source translation, we consider it up to date
-            return translation_status.UP_TO_DATE
-        # If the translation was edited before the last major change in the source language, it is outdated
-        return translation_status.OUTDATED
+        if translation.machine_translated:
+            # If the translation has been made by machine translation and is up to date, show the bot icon
+            return translation_status.MACHINE_TRANSLATED
+        # If the translation was edited after the source translation, we consider it up to date
+        return translation_status.UP_TO_DATE
 
     @classmethod
     def search(cls, region, language_slug, query):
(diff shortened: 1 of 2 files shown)


integreat_cms/cms/models/abstract_content_translation.py
```diff
@@ -447,6 +445,18 @@ def translation_state(self):
             # If the source language does not have a major public version, the translation is considered "outdated",
             # because the content is not in sync with its source translation
             return translation_status.OUTDATED
+        if (
```
ulliholtgrave, line 448: I guess you can optimize this a little bit. See suggestion:
```suggestion
        if translation.machine_translated:
            if translation.last_updated >= source_translation.last_updated:
                # If the translation has been made by machine translation and is up to date, show the bot icon
                return translation_status.MACHINE_TRANSLATED
            # If the translation has been made by machine translation but is not up to date, show the outdated icon
            return translation_status.OUTDATED
```

ulliholtgrave (changes requested):
In general, this seems to fix the issue, but I have one small suggestion for an improvement :)

integreat_cms/cms/models/abstract_content_translation.py
timoludwig: I agree that it can be optimized, but I would go one step further: We don't even need the nested `if` condition if we just move the "machine translated" check down and use "up to date" as last return value:
```
    def translation_state(self):
        """
        This function returns the current state of a translation in the given language.

        :return: A string describing the state of the translation, one of :data:`~integreat_cms.cms.constants.translation_status.CHOICES`
        :rtype: str
        """
        if not (translation := self.major_version):
            # If the page does not have a major public version, it is considered "missing" (keep in mind that it might
            # have draft versions or public versions that are marked as "minor edit")
            return translation_status.MISSING
        if translation.currently_in_translation:
            return translation_status.IN_TRANSLATION
        if not self.source_language:
            # If the language of this translation is the root of this region's language tree, it is always "up to date"
            return translation_status.UP_TO_DATE
        if not (source_translation := self.major_source_translation):
            # If the source language does not have a major public version, the translation is considered "outdated",
            # because the content is not in sync with its source translation
            return translation_status.OUTDATED
        if source_translation.translation_state == translation_status.OUTDATED:
            # If the source translation is already outdated, this translation is as well
            return translation_status.OUTDATED
        if translation.last_updated < source_translation.last_updated:
            # If the translation was edited before the last major change in the source language, it is outdated
            return translation_status.OUTDATED
        if translation.machine_translated:
            # If the translation has been made by machine translation and is up to date, show the bot icon
            return translation_status.MACHINE_TRANSLATED
        # If the translation was edited after the source translation, we consider it up to date
        return translation_status.UP_TO_DATE
```

integreat_cms/release_notes/current/unreleased/2155.yml
```diff
@@ -0,0 +1,2 @@
+en: Fix translation status for outdated translations after machine translation
+de: Behebe Fehler beim Übersetzungsstatus für abgelaufener Übersetzung nach machineller Übersetzung
```
timoludwig, line 2:
```suggestion
de: Behebe Fehler beim Übersetzungsstatus für veraltete maschinelle Übersetzungen
```

timoludwig (changes requested):

integreat_cms/cms/models/abstract_content_translation.py
```diff
@@ -436,25 +436,26 @@ def translation_state(self):
...
             # If the source language does not have a major public version, the translation is considered "outdated",
             # because the content is not in sync with its source translation
-            return translation_status.OUTDATED
-        if source_translation.translation_state == translation_status.OUTDATED:
+            not (source_translation := self.major_source_translation)
             # If the source translation is already outdated, this translation is as well
+            or source_translation.translation_state == translation_status.OUTDATED
+            # If the translation was edited before the last major change in the source language, it is outdated
+            or translation.last_updated < source_translation.last_updated
```
timoludwig, line 451:
```suggestion
            or translation.last_updated <= source_translation.last_updated
```
Haha, I thought the case that two translations are created in exactly the same second is a rare enough edge case that we can ignore it - apparently, in the test data there are translations with exactly the same date and they expect the translations to be flagged as outdated in this case...
I'm fine with either way - adapting the tests or make this if-condition compliant with the tests^^

timoludwig (approved):
:+1:

ulliholtgrave (approved):
Looks good 👍

Issue #1904, opened 2022-11-28
Long Links in PDFs are not wrapped around


nassabay:
### Describe the Bug
if a Link is longer than 1 line, it just runs over the line break

### Steps to Reproduce

1. Go to '[Kreis Soest](https://integreat.app/kreis-soest/de/corona-virus/corona-test)'
2. Click on 'PDF erstellen'
3. In the PDF Links run over the line end.

### Expected Behavior
Line break in Link

### Actual Behavior
No Link break in Link

![Screenshot (804)](https://user-images.githubusercontent.com/65529262/204301279-c11b4097-57e5-4a8b-acb6-575fbe358afc.png)

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

mael-fosso:
wo in backend kann ich zugreifen?

svenseeberg:
https://www.w3schools.com/cssref/css3_pr_overflow-wrap.php sadly is only available with CSS 3. The module only supports CSS 2. I was wondering if we can solve this by allowing `&#x200B;` (zero width space characters) in long links as a workaround. I tried to insert them but they get stripped from the HTML code.

seluianova:
There is a discussion on this issue:
https://github.com/xhtml2pdf/xhtml2pdf/issues/339

But the proposed property '-pdf-word-wrap: CJK;' only works for 'div' and 'p' elements.
I couldn't get it to apply to links.

charludo:
I know this is far from ideal, but until a fix (eventually) lands in `xhtml2pdf`, could we not simply truncate links over a certain length?

E.g. turn [https://thisisaverylonglink.com/it/just/keeps/going/on/and/on/and/on/and/on/and/on/and/on/and/so/forth/etc/etc/etc/etc/enough.html](https://thisisaverylonglink.com/it/just/keeps/going/on/and/on/and/on/and/on/and/on/and/on/and/so/forth/etc/etc/etc/etc/enough.html) 

into either

[https://thisisaverylonglink.com/it/just/keeps/going/on/and/on/and[...]](https://thisisaverylonglink.com/it/just/keeps/going/on/and/on/and/on/and/on/and/on/and/on/and/so/forth/etc/etc/etc/etc/enough.html)

or

[...]

PR #2164, opened 2023-03-26
Truncate long links in pdf exports to a maximum of 50 chars

membralala:
### Short description

This PR should provide a temporary solution for long links in PDFs. Instead of overflowing, long links are now limited to 50 chars. Longer links will be truncated. 

### Proposed changes

- Add a filter that truncates long links 
- Apply filter on pages in the PDF template

### Side effects

- This might be a naive solution. I'm not 100% confident, that there won't be edge cases, where this might affect links in an unexpected way. 

### Resolved issues

Fixes: #1904

### DIFF
--- a/integreat_cms/cms/fixtures/test_data.json
+++ b/integreat_cms/cms/fixtures/test_data.json
@@ -1695,7 +1695,7 @@
       "title": "Alltag",
       "slug": "alltag",
       "status": "PUBLIC",
-      "content": "<a href=\"https://integreat-app.de/\">G&uuml;ltiger Link</a><br><a href=\"https://integreat.app\">Ungepr&uuml;fter Link</a><br><a href=\"https://invalid.integreat.app/\">Ung&uuml;ltiger Link</a><br><a href=\"https://invalid2.integreat.app/\">Ignorierter Link</a><br><br><a href=\"https://integreat.app/augsburg/de/willkommen/\">Interner Link</a>",
+      "content": "<a href=\"https://integreat-app.de/\">G&uuml;ltiger Link</a><br><a href=\"https://integreat.app\">Ungepr&uuml;fter Link</a><br><a href=\"https://invalid.integreat.app/\">Ung&uuml;ltiger Link</a><br><a href=\"https://invalid2.integreat.app/\">Ignorierter Link</a><br><a href=\"https://integreat.app/augsburg/de/willkommen/\">Interner Link</a><br><a href=\"https://integreat-app.de/\">Ein langer Link zum Testen des Verhaltens von Links im PDF Export. Text mit Leerzeichen sollte gebrochen werden, W&#246;rter mit mehr als 50 Zeichen sollten hingegen gek&#252;rzt werden. www.dashieristeinsehrlangerlinkalszusammenh&#228;ngendezeichenkette.de</a>",
       "language": 1,
       "currently_in_translation": false,
       "version": 1,
@@ -3563,6 +3563,18 @@
       "field": "content",
       "url": 7,
       "text": "Interner Link",
+      "ignore": false
+    }
+  },
+  {
+    "model": "linkcheck.link",
+    "pk": 12,
+    "fields": {
+      "content_type": 23,
+      "object_id": 61,
+      "field": "content",
+      "url": 3,
+      "text": "Ein langer Link zum Testen des Verhaltens von Links im PDF Export. Text mit Leerzeichen sollte gebrochen werden, Wörter mit mehr als 50 Zeichen sollten hingegen gekürzt werden. www.dashieristeinsehrlangerlinkalszusammenhängendezeichenkette.de",
       "ignore": false
     }
   },
--- a/integreat_cms/cms/templates/pages/page_pdf.html
+++ b/integreat_cms/cms/templates/pages/page_pdf.html
@@ -102,11 +102,11 @@
                 <h1 class="level-{{ page.depth|add:"-1" }}">{{ page_translation.title }}</h1>
                 <div class="content">
                     {% if page.mirrored_page_first %}
-                        {{ page_translation.mirrored_translation_text|pdf_strip_fontstyles|safe }}
+                        {{ page_translation.mirrored_translation_text|pdf_strip_fontstyles|pdf_truncate_links:50|safe }}
                     {% endif %}
-                    {{ page_translation.content|pdf_strip_fontstyles|safe }}
+                    {{ page_translation.content|pdf_strip_fontstyles|pdf_truncate_links:50|safe }}
                     {% if not page.mirrored_page_first %}
-                        {{ page_translation.mirrored_translation_text|pdf_strip_fontstyles|safe }}
+                        {{ page_translation.mirrored_translation_text|pdf_strip_fontstyles|pdf_truncate_links:50|safe }}
                     {% endif %}
                 </div>
                 {% for close in info.close %}
--- a/integreat_cms/cms/templatetags/pdf_filters.py
+++ b/integreat_cms/cms/templatetags/pdf_filters.py
@@ -5,6 +5,7 @@ import re
 from html import unescape
 
 from django import template
+from django.utils.text import Truncator
 from lxml.etree import ParserError
 from lxml.html import fromstring, tostring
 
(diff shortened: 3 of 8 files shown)


integreat_cms/cms/templatetags/pdf_filters.py
```diff
@@ -30,3 +30,30 @@ def pdf_strip_fontstyles(instance):
...
+    :return: The instance with truncated links
+    :rtype: ~integreat_cms.cms.models.pages.page.Page
+    """
+    try:
+        content = fromstring(instance)
+    except ParserError:
+        return instance
+    truncation_postfix = "[...]"
+    truncation_length = max_chars - len(truncation_postfix)
+    for elem, _, _, _ in content.iterlinks():
```
seluianova, line 55: I'm not very familiar with Python conventions, so this is more of a question:
Isn't it better to use meaningful names, even if these variables are not used?

timoludwig: Good point. In general, the underscore is a very common convention in Python for unused variables, and we also use this in our code base. However, it's not completely ideal because it clashes with the alias import for translations, e.g. imagine
```
from django.utils.translation import gettext as _
```
on top of a file, then having the "unused" pattern
```
var, _ = some_func()
```
and later you want to translate something
```
text = _("Translated text")
```
This won't work, since `_` has been overwritten.
So yes, I's agree that in this case it's better to use readable variable names and then ignore the corresponding pylint warning.

seluianova (changes requested):
@JoeyStk it works for me with the same link, on the same page.
Have you done something specific? :)
image

Looks good to me, I couldn't break it anyhow.
Only if I add the link as text and not as a link. But I believe it's okay.

I would suggest adding a long link to the test data and adjusting the pdf test, so that this defect is covered.

integreat_cms/cms/templatetags/pdf_filters.py
membralala: Okay, I will change that, the translation alias was not on my mind, makes totally sense!

integreat_cms/cms/templates/pages/page_pdf.html
```diff
@@ -102,11 +102,11 @@ <h1 id="title_page">{{ language.table_of_contents }}</h1>
                 <h1 class="level-{{ page.depth|add:"-1" }}">{{ page_translation.title }}</h1>
                 <div class="content">
                     {% if page.mirrored_page_first %}
-                        {{ page_translation.mirrored_translation_text|pdf_strip_fontstyles|safe }}
+                        {{ page_translation.mirrored_translation_text|pdf_strip_fontstyles|pdf_truncate_links:50|safe }}
```
seluianova, line 105: Questionable thing, but we could probably allow more symbols.
Maybe 100?
So that links that do not exceed line length are left as is.

membralala:
@JoeyStk I will have a deeper look into that tomorrow, but might it be, that the pdf with the wrong link was already cached in your /integreat-cms/pdf directory? Could you maybe delete all cached files in there and then try it again and give a feedback, if the issue still persists? 👀

integreat_cms/cms/templates/pages/page_pdf.html
membralala: Yeah, I don't really have a clear opinion on that one, as I can see pros and cons in both cases. 
On the one hand more links might be truncated, when only 50 chars are allowed, on the other hand shorter links read nicer (we might also prevent weird looking line breaks etc) and with longer links the chance of edge cases might become bigger (e.g. what if links are indented, e.g because of a bullet point list?)

seluianova: Bullet point is a good point 😀
But I found one more case to think of:
If the link has a normal text, and this text is longer than 50 characters, it is truncated too
image

timoludwig: How about sticking with 50 chars but only truncate the specific words of the text?
Screenshot 2023-03-28 at 23-52-01 Integreat - Englisch - Daily Life pdf

integreat_cms/cms/templatetags/pdf_filters.py
```diff
@@ -30,3 +30,30 @@ def pdf_strip_fontstyles(instance):
...
+        if elem.text and len(elem.text) > max_chars:
+            elem.text = f"{elem.text[:truncation_length]}{truncation_postfix}"
```
timoludwig, line 57:
```suggestion
    for elem, _, _, _ in content.iterlinks():
        elem.text = " ".join(
            Truncator(word).chars(max_chars) for word in elem.text.split(" ")
        )
```
You could also make use of Django's `Truncator` which already handles the truncation postfix on its own...
Also, I suggest to only truncate the specific words in the link text, not the whole text itself.

(this requires `from django.utils.text import Truncator`)

timoludwig (changes requested):

integreat_cms/cms/templates/pages/page_pdf.html
membralala: Yes, I like this solution.

seluianova (commented):
Looks good to me 👍
But test is still missed :)

integreat_cms/cms/templatetags/pdf_filters.py
```diff
@@ -30,3 +31,30 @@ def pdf_strip_fontstyles(instance):
...
+    :param max_chars: The maximal length of a link before it will be truncated
+    :type max_chars: int
+
+    :return: The instance with truncated links
+    :rtype: ~integreat_cms.cms.models.pages.page.Page
+    """
+    try:
+        content = fromstring(instance)
+    except ParserError:
+        return instance
```
timoludwig, line 54:
```suggestion
# pylint: disable=unused-variable
def pdf_truncate_links(page_content, max_chars):
    """
    This tag returns the page content with truncated link texts.

    :param page_content: The content of the page
    :type page_content: str

    :param max_chars: The maximal length of a link before it will be truncated
    :type max_chars: int

    :return: The content with truncated links
    :rtype: str
    """
    try:
        content = fromstring(page_content)
    except ParserError:
        return page_content
```
Ah, I just noticed that the types are incorrect.

timoludwig (changes requested):

integreat_cms/cms/fixtures/test_data.json
```diff
@@ -3566,6 +3566,18 @@
...
   },
+  {
+    "model": "linkcheck.link",
+    "pk": 12,
+    "fields": {
+      "content_type": 23,
+      "object_id": 61,
+      "field": "content",
+      "url": 3,
+      "text": "Gültiger Link",
```
timoludwig, line 3577:
```suggestion
      "text": "Ein langer Link zum Testen des Verhaltens von Links im PDF Export. Text mit Leerzeichen sollte gebrochen werden, Wörter mit mehr als 50 Zeichen sollten hingegen gekürzt werden. www.dashieristeinsehrlangerlinkalszusammenhängendezeichenkette.de",
```
When I import the test data and run `./tools/integreat-cms-cli findlinks`, I get the following error:
```
linkcheck.models.Link.MultipleObjectsReturned: get() returned more than one Link -- it returned 2!
```

So apparently it's not possible to have the same combination of link text and url twice.

integreat_cms/cms/fixtures/test_data.json
```diff
@@ -1695,7 +1695,7 @@
       "title": "Alltag",
       "slug": "alltag",
       "status": "PUBLIC",
-      "content": "<a href=\"https://integreat-app.de/\">G&uuml;ltiger Link</a><br><a href=\"https://integreat.app\">Ungepr&uuml;fter Link</a><br><a href=\"https://invalid.integreat.app/\">Ung&uuml;ltiger Link</a><br><a href=\"https://invalid2.integreat.app/\">Ignorierter Link</a><br><br><a href=\"https://integreat.app/augsburg/de/willkommen/\">Interner Link</a>",
+      "content": "<a href=\"https://integreat-app.de/\">G&uuml;ltiger Link</a><br><a href=\"https://integreat.app\">Ungepr&uuml;fter Link</a><br><a href=\"https://invalid.integreat.app/\">Ung&uuml;ltiger Link</a><br><a href=\"https://invalid2.integreat.app/\">Ignorierter Link</a><br><a href=\"https://integreat.app/augsburg/de/willkommen/\">Interner Link</a><br><a href=\"https://integreat-app.de/\">Ein langer Link zum Testen des Verhaltens von Links im PDF Export. Text mit Leerzeichen sollte gebrochen werden, W&#246;rter mit mehr als 50 Zeichen sollten hingegen gek&#252;rzt werden. www.dashieristeinsehrlangerlinkalszusammenh&#228;ngendezeichenkette.de</a>",
```
timoludwig, line 1698: Was it intentional that the long text and `www.dashieristeinsehrlangerlinkalszusammenh&#228;ngendezeichenkette.de` are one and the same link? Or did you mean to create two different links?

timoludwig (changes requested):
Awesome, thanks a lot! :+1:
Just the link in the test data isn't using the correct link text...

integreat_cms/cms/fixtures/test_data.json
membralala: Oh yes, I missed that 🙈

integreat_cms/cms/fixtures/test_data.json
membralala: This was intentional. I did not see a reason to create a new link to test this behaviour as it is just about the text content and not about the actual link reference. However, I could of course also split it, if you think this is a better solution.

timoludwig: Alright, no that's fine!

timoludwig (approved):
Thanks a lot! :+1:

In this case, you could also clean up the commit description a bit, but it's not that important, the few lines do not really matter...

JoeyStk (approved):
Works for me too :tada:

Issue #1999, opened 2023-01-14
Add management command to reset DeepL budget


timoludwig:
### Motivation

At the moment, we track the DeepL budget on a yearly basis, but we don't ever reset it.

### Proposed Solution

Probably it's best to do this via a monthly cron-job which is configured in our production deployment config on salt.
However, I think the actual functionality how the reset mechanism works should be implemented as part of the CMS.
A good place for this would be a [management command](https://docs.djangoproject.com/en/3.2/howto/custom-management-commands/) in my opinion:

- Create new command following [these examples](https://github.com/digitalfabrik/integreat-cms/tree/develop/integreat_cms/core/management/commands)
    - Document this command in [docs/src/management-commands.rst](https://github.com/digitalfabrik/integreat-cms/blob/develop/docs/src/management-commands.rst)
    - Test this command in [tests/core/management/commands](https://github.com/digitalfabrik/integreat-cms/tree/develop/tests/core/management/commands)

This command could perform the following steps:

- Query all regions where `deepl_renewal_month` is the current month and:
    - Reset `deepl_budget_used` to `0`
    - Reset the `deepl_midyear_start_month` to the current month, because the difference is only used in the first year of the add-on

### Alternatives

Write a small script on the production system that directly interacts with the database

### Additional Context

Also add the documentation about the cronjob to the [production documentation](https://github.com/digitalfabrik/integreat-cms/blob/4dab5e522b0781be9d7acf2e4faf8ee537ff4e68/sphinx/prod-server.rst).

PR #2181, opened 2023-03-29
Add management command to reset DeepL budget

MizukiTemma:
### Short description

This PR adds a management command that resets the DeepL budget to 0 if a region's renewal month is the current month.

### Proposed changes

- Query regions whose renewal month is the current month, reset their `deepl_budget_used` to 0, set `deepl_midyear_start_month` to the current month.
- Add a test for the command (and some test data for it too).

### Side effects

- It seems `datetime.now()` returns the current time of GMT. If the cronjob starts between 00:00 and 02:00 of the first day of the month, the returned value must be adjusted.

### Resolved issues

Fixes: #1999 

### DIFF
--- a/docs/src/management-commands.rst
+++ b/docs/src/management-commands.rst
@@ -137,6 +137,18 @@ Translate an entire region into Easy German via SUMM.AI::
 * ``--initial``: Whether existing translations should not be updated
 
 
+``reset_deepl_budget``
+~~~~~~~~~~~~~~~~~~~~~~
+
+Reset DeepL budget of regions whose renewal month is the current month::
+
+    integreat-cms-cli reset_deepl_budget [--force]
+
+**Options:**
+
+* ``--force``: Allow to reset the budget even if it's not the first day of the month
+
+
 Create new commands
 -------------------
 
--- a/integreat_cms/core/management/commands/reset_deepl_budget.py
+++ b/integreat_cms/core/management/commands/reset_deepl_budget.py
@@ -1,0 +1,55 @@
+import logging
+from datetime import datetime
+
+from django.core.management.base import CommandError
+
+from ....cms.models import Region
+from ..log_command import LogCommand
+
+logger = logging.getLogger(__name__)
+
+
+class Command(LogCommand):
+    """
+    Management command to reset DeepL budget
+    """
+
+    help = "Reset DeepL budget"
+
+    def add_arguments(self, parser):
+        """
+        Define the arguments of this command
+
+        :param parser: The argument parser
+        :type parser: ~django.core.management.base.CommandParser
+        """
+        parser.add_argument(
+            "--force",
+            action="store_true",
+            help="Whether the reset should be run despite it's not the 1st day of the month",
+        )
+
+    # pylint: disable=arguments-differ
+    def handle(self, *args, force, **options):
+        """
+        Try to run the command
+        """
+
+        current_day = datetime.now().day
+        current_month = datetime.now().month - 1
+
+        if current_day != 1 and not force:
+            raise CommandError(
+                "It is not the 1st day of the month. If you want to reset DeepL budget despite that, run the command with --force"
+            )
+
+        if not (regions := Region.objects.filter(deepl_renewal_month=current_month)):
+            self.print_info(
+                "✔ There is no region whose DeepL budget needs to be reset."
+            )
+        else:
+            for region in regions:
+                region.deepl_budget_used = 0
+                region.deepl_midyear_start_month = None
+                region.save()
+            self.print_info("✔ DeepL budget has been reset.")
--- a/tests/core/management/commands/test_reset_deepl_budget.py
+++ b/tests/core/management/commands/test_reset_deepl_budget.py
@@ -1,0 +1,82 @@
+from datetime import datetime
+from unittest.mock import patch
+
+import pytest
+from django.core.management.base import CommandError
+
+from integreat_cms.cms.models import Region
+
+from ..utils import get_command_output
+
+
+# pylint: disable=too-few-public-methods
+class datetime_not_first_day:
+    """
+    Fake datetime object where now() is never the first day of the month
+    """
+
+    @classmethod
+    def now(cls):
+        """
+        generate a date with day=2
+
+        :returns: a date which is never the 1st day of the month
+        :rtype: datetime.datetime
+        """
+        real_now = datetime.now()
+        return real_now.replace(day=2)
+
+
+@patch("datetime.datetime", datetime_not_first_day)
+def test_not_first_day():
+    """
+    Ensure that the command will not run when it's not the 1st day of the month without --force
+    """
+    with pytest.raises(CommandError) as exc_info:
+        assert not any(get_command_output("reset_deepl_budget"))
+        assert (
+            str(exc_info.value)
+            == "It is not the 1st day of the month. If you want to reset DeepL budget despite that, run the command with --force"
+        )
+
+
+# pylint: disable=unused-argument
+@pytest.mark.order("last")
+@pytest.mark.django_db(transaction=True, serialized_rollback=True)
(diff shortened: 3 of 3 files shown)


integreat_cms/core/management/commands/reset_deepl_budget.py
```diff
@@ -0,0 +1,35 @@
...
+        current_month = datetime.now().month - 1
+
+        if not (regions := Region.objects.filter(deepl_renewal_month=current_month)):
+            self.print_info(
+                "✔ There is no region whose DeepL budget needs to be reset."
+            )
+        else:
+            for region in regions:
+                region.deepl_budget_used = 0
+                region.deepl_midyear_start_month = current_month
```
charludo, line 31: I think this should be reset to `None`, which is tthe default for non-midyear-starting regions, which the regions turns into after passing `deepl_renewal_month` for the first time.

MizukiTemma: It is so specified in the issue description. How do you think @timoludwig ?

timoludwig: Ah sorry, this was a mistake in the issue description. It won't make any difference in the calculation, but I agree with @charludo that it makes more sense to have the midyear start disabled (by setting it to `None`) rather than enabling it and set it to the same month as the regular renewal month...

MizukiTemma: Applied :slightly_smiling_face:

timoludwig:
> Something that just occured to me: with this, it's rather easy to reset the budget multiple times (accidentally), i.e. it doesn't matter whether the command is used on the 01. or 31. of a month.
> 
> The only way around it that I see is storing the `last_reset_year` or something n the region model.
> 
> Do you think this is a big enough risk to justify the extra effort?

Hmm, I think a new field in the region model is a bit overkill, but one solution I could think of:
Maybe only do something when it's the 1st day of the month, and show a warning otherwise?
And add a new command line parameter `--force` or similar which overwrites the warning and allows to reset the counter even if it's not the first day? Then we have some kind of safety against accidental executions...

tests/conftest.py
```diff
@@ -73,6 +73,7 @@ def load_test_data_transactional(transactional_db, django_db_blocker):
     with django_db_blocker.unblock():
         call_command("loaddata", "integreat_cms/cms/fixtures/test_roles.json")
         call_command("loaddata", "integreat_cms/cms/fixtures/test_data.json")
+        call_command("loaddata", "integreat_cms/cms/fixtures/test_regions.json")
```
timoludwig, line 76:
```suggestion
```
I think loading an additional fixture for all test cases is a bit overkill if only one test case is using it. Instead, either load the fixture in the specific test case or generate the required objects on demand.

MizukiTemma:
@charludo It's a good idea to check/prevent resetting more than once per month. Since the region model already has many fields, I'll adopt the idea from @timoludwig.

tests/core/management/commands/test_reset_deepl_budget.py
```diff
@@ -0,0 +1,33 @@
...
+@pytest.mark.order("last")
+@pytest.mark.django_db(transaction=True, serialized_rollback=True)
+def test_reset_deepl_budget(load_test_data_transactional):
+    """
+    Ensure that DeepL budget gets reset successfully
+    """
+
+    current_month = datetime.now().month - 1
+    target_region_id = (
+        Region.objects.filter(deepl_renewal_month=current_month).first().id
```
timoludwig, line 20:
```suggestion
    region = Region.objects.create(
        slug="deepl_test",
        deepl_renewal_month=current_month,
        deepl_midyear_start_month=current_month + 1,
        deepl_budget_used=42,
```
I'd just generate the required region here.

integreat_cms/cms/fixtures/test_regions.json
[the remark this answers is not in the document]
MizukiTemma: Thank you for review and suggestions 😃 I applied them!

tests/core/management/commands/test_reset_deepl_budget.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    Ensure that the command will not run when it's not the 1st day of the month without --force
+    """
+    with pytest.raises(CommandError) as exc_info:
+        assert not any(get_command_output("reset_deepl_budget"))
+        if datetime.now().day != 1:
+            assert (
+                str(exc_info.value)
+                == "It is not the 1st day of the month. If you want to reset DeepL budget despite that, run the command with --force"
+            )
```
timoludwig, line 21:
```suggestion
from unittest.mock import patch

import pytest
from django.core.management.base import CommandError

from integreat_cms.cms.models import Region

from ..utils import get_command_output


class datetime_not_first_day:
    """
    Fake datetime object where now() is never the first day of the month
    """

    @classmethod
    def now(cls):
        real_now = datetime.now()
        return real_now.replace(day=2)


@patch("datetime.datetime", datetime_not_first_day)
def test_not_first_day():
    """
    Ensure that the command will not run when it's not the 1st day of the month without --force
    """
    with pytest.raises(CommandError) as exc_info:
        assert not any(get_command_output("reset_deepl_budget"))
        assert (
            str(exc_info.value)
            == "It is not the 1st day of the month. If you want to reset DeepL budget despite that, run the command with --force"
        )
```
This test case will fail on the first day of the month since it will not raise the given exception.
We could solve this by a conditional, but I would prefer if we could patch the datetime class to just never return the first day of the month...

tests/core/management/commands/test_reset_deepl_budget.py
```diff
@@ -0,0 +1,83 @@
...
+        ), "The DeepL budget should have been reset to 0."
+        assert (
+            region2.deepl_budget_used == 42
+        ), "The DeepL budget should remain unchanged."
+        assert (
+            region1.deepl_midyear_start_month is None
+        ), "The midyear start month should have been reset to None."
+        assert (
+            region2.deepl_midyear_start_month == current_month + 1
+        ), "The midyear start month should have been reset to None."
```
timoludwig, line 83:
```suggestion
    out, err = get_command_output("reset_deepl_budget", "--force")
    region1.refresh_from_db()
    region2.refresh_from_db()
    assert "✔ DeepL budget has been reset." in out
    assert not err
    assert (
        not region1.deepl_budget_used
    ), "The DeepL budget of region 1 should have been reset to 0."
    assert (
        region2.deepl_budget_used == 42
    ), "The DeepL budget of region 2 should remain unchanged."
    assert (
        region1.deepl_midyear_start_month is None
    ), "The midyear start month of region 1 should have been reset to None."
    assert (
        region2.deepl_midyear_start_month == current_month + 1
    ), "The midyear start month of region 2 should not have been reset to None."
```
These are two times the same assertions, right?
I guess then it's fine to just pass `--force` always, even when it's the first day of the month.

timoludwig (changes requested):

docs/src/management-commands.rst
```diff
@@ -137,6 +137,14 @@ Translate an entire region into Easy German via SUMM.AI::
 * ``--initial``: Whether existing translations should not be updated
 
 
+``reset_deepl_budget``
+~~~~~~~~~~~~~~~~~~~~~~
+
+Reset DeepL budget of regions whose renewal month is the current month::
+
+    integreat-cms-cli reset_deepl_budget
```
timoludwig, line 145:
```suggestion
    integreat-cms-cli reset_deepl_budget [--force]

**Options:**

* ``--force``: Allow to reset the budget even if it's not the first day of the month
```
could you document the new option?

timoludwig (changes requested):

timoludwig (approved):
Great, thanks! :+1:

charludo (approved):
🎉 Nice, the warning/force flag sound like a good solution

Issue #1851, opened 2022-11-10
Add a region link replacement function


svenseeberg:
### Motivation
Sometimes we rename or copy regions. In these cases the links in the content are not changed and point to the source or no longer existent region. Also the link status should be reset. That means the checker needs to check the link first, before we show a status.

### Proposed Solution
Add a feature that allows to easily replace links to suit a newly created or rename region. This can probably be located near the Broken Link Checker in the GUI.

This function should only be available to expert users.

### Alternatives
We currently use the following code in the Django shell to manually solve this problem:
```python
from copy import deepcopy
from integreat_cms.cms.models import *
from linkcheck.models import *
from linkcheck import update_lock
from lxml.html import rewrite_links
from functools import partial

def replace_link(old_url, new_url, link):
    if link == old_url:
        return new_url
    return link

with update_lock:
    all_pages = Region.objects.filter(slug='deutschlandvorlage').first().pages.all()
    for page in all_pages:
        for language in page.languages:
            trans = page.get_translation(language.slug).latest_version
            if trans is None:
                continue
            trans_new = deepcopy(trans)
            changed = False
            for link in Link.objects.filter(object_id=trans.id):
                if str(link.url).startswith("https://integreat.app/deutschlandvorlage-textlab/"):
                    print(link.url)
                    fixed_url = str(link.url).replace("https://integreat.app/deutschlandvorlage-textlab/", "https://integreat.app/deutschlandvorlage/")
                    trans_new.content = rewrite_links(
                        trans_new.content,
                        partial(replace_link, link.url.url, fixed_url),
                    )
                    link.delete()
                    changed = True
            if changed:
                trans_new.id = None
                trans_new.version = trans.version + 1
                trans_new.minor_edit = True
                trans_new.save()
```

### Additional Context
We also want a string replacement function (#1382). But it is probably safer to implement this first.

svenseeberg:
https://github.com/digitalfabrik/integreat-cms/issues/1383 is even more relevant, I guess.

timoludwig:
The utility function introduced in #2126 could be used for this as well, so I suggest to wait until that PR is merged before working on this issue.

PR #2193, opened 2023-04-03
Add a region link replacement function

MizukiTemma:
### Short description

This PR adds a possibility to replace region links in Broken link checker section.

### Proposed changes

- Add a small GUI in the broken link section.
- <s>Check if the new slug exists.</s>
- Use the replace function from #2126 

### Side effects

- None (probably).

### Resolved issues

Fixes: #1851 


seluianova (commented):
I tested the branch a bit, I think it works basically well, thanks!

I see one potential _side effect_:

If the region name is also part of some external links, they will also be updated (and therefore broken).
Do we want to avoid it?

And a couple of _conceptual questions_:

1. When we clone a region, we automatically update the links now. 
Why not do the same when changing a region slug? (I assume this was the only case for which the current change was requested, right?)

2. "Broken Links" view has a regional slug, so only links in the selected region are displayed.
But with "Replace Region Link" we ignore the slug and can replace any region across pages in all regions.
So I can open "Broken Links" in Augsburg region and replace "dresden" with "nurnberg".
I think it's a bit confusing.
As a user, I think I would expect only links in Augsburg containing "dresden" to be updated.
What do you think?

integreat_cms/cms/views/linkcheck/linkcheck_actions.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    region = request.region
+    data = json.loads(request.body.decode("utf-8"))
+    search = data.get("oldSlug")
+    replace = data.get("newSlug")
+
+    if not search or not replace:
+        return JsonResponse(
+            data={
+                "response": "Fail",
```
seluianova, line 34: Maybe it's a picky thing, but I would suggest to return an error response status when the request has not been fulfilled, for example 400.

integreat_cms/cms/views/linkcheck/linkcheck_actions.py
```diff
@@ -0,0 +1,49 @@
...
+            }
```
seluianova, line 35:
```suggestion
            }, 
            status=400,
```

MizukiTemma:
Thank you for review!


> I tested the branch a bit, I think it works basically well, thanks!
> 
> I see one potential _side effect_:
> 
> If the region name is also part of some external links, they will also be updated (and therefore broken). Do we want to avoid it?
> 
> Random example: Some page contains an external link containing "nurnberg" https://sz.fau.eu/2022/05/the-language-centre-of-fau-erlangen-nurnberg-is-looking-to-fill-the-following-positions-from-oct-1-2022/ If we want to change the region slug from "nurnberg" to "nuernberg" and update links, the link above will be broken.

Right, this case must be taken into consideration. I'll investigate possibilities.

> And a couple of _conceptual questions_:
> 
> 1. When we clone a region, we automatically update the links now.
>    Why not do the same when changing a region slug? (I assume this was the only case for which the current change was requested, right?)

Good idea 🙂 We can add an automatic link replacement into the process of renaming a region. I think though we need anyway this replacement feature in the broken link view, because automatic update is not possible when a reagion whose content is reffered as link in another region, is deleted and users probably need possibility to update links by themselves.

> 2. "Broken Links" view has a regional slug, so only links in the selected region are displayed.
>    But with "Replace Region Link" we ignore the slug and can replace any region across pages in all regions.
>    So I can open "Broken Links" in Augsburg and replace "dresden" with "nurnberg".
>    I think it's a bit confusing.
>    As a user, I think I would expect only links in Augsburg containing "dresden" to be updated.
>    What do you think?

Do you mean links with "dresden" in contents of region Hamburg can be replaced when an user does a replacement "dresden" -> nurnberg" in the broken link view of Augsburg?
This will be prevented because region argument is passed to `replace_links` which is called in `replace_region_links`. Users replace only links in the content of their region.

seluianova:
> Do you mean links with "dresden" in contents of region Hamburg can be replaced when an user does a replacement "dresden" -> nurnberg" in the broken link view of Augsburg?

Sorry, I don't know how I tested that it didn't work, but now I've re-checked it and it does work.
Pls ignore this item 🙈

> when a reagion whose content is reffered as link in another region, is deleted

But when a region is deleted (not cloned, not updated), its pages are also deleted.
How would replacing regional slugs help in such a case?

integreat_cms/cms/templates/linkcheck/links_by_filter.html
```diff
@@ -131,4 +131,42 @@ <h1 class="heading">
...
+                <i icon-name="arrow-right" class="inline-block ml-2"></i>
+                <input id="replace_region_to"
+                       placeholder="{% translate "New region slug" %}"
+                       class="pr-8 appearance-none block bg-white text-gray-800 placeholder-gray-800 border border-gray-400 rounded py-3 px-4 leading-tight ml-2"/>
+                <label id="region-replace-btn"
+                       class="inline-block cursor-pointer bg-blue-500 hover:bg-blue-600 text-white font-bold py-2 px-4 m-0 rounded ml-2"
+                       data-url="{% url 'replace_region_link' region_slug=request.region.slug %}">
+                    <i icon-name="link" class="inline-block pr-1"></i>
+                    {% translate "Replace links" %}
+                </label>
```
seluianova, line 150:
```suggestion
            <div class="flex mt-2 mb-2 gap-2">
                <input id="replace_region_from"
                       placeholder="{% translate "Old region slug" %}"
                       class="pr-8 appearance-none block bg-gray text-gray-800 placeholder-gray-800 border border-gray-400 rounded py-3 px-4 leading-tight"/>
                <i icon-name="arrow-right" class="inline-block"></i>
                <input id="replace_region_to"
                       placeholder="{% translate "New region slug" %}"
                       class="pr-8 appearance-none block bg-white text-gray-800 placeholder-gray-800 border border-gray-400 rounded py-3 px-4 leading-tight"/>
                <button id="region-replace-btn"
                        class="btn"
                        data-url="{% url 'replace_region_link' region_slug=request.region.slug %}">
                    <i icon-name="link"></i>
                    {% translate "Replace links" %}
                </button>
```

seluianova, line 150: Another tiny suggestion - to use the button tag so that our common css style for buttons is applied

```suggestion
                <button id="region-replace-btn"
                        class="btn"
                        data-url="{% url 'replace_region_link' region_slug=request.region.slug %}">
                    <i icon-name="link"></i>
                    {% translate "Replace links" %}
                </button>
```

seluianova, line 150: Another tiny suggestion - to use the button tag for the "Replace links" button so that our common css style for buttons is applied

```suggestion
            <div class="flex mt-2 mb-2 gap-2">
                <input id="replace_region_from"
                       placeholder="{% translate "Old region slug" %}"
                       class="pr-8 appearance-none block bg-gray text-gray-800 placeholder-gray-800 border border-gray-400 rounded py-3 px-4 leading-tight"/>
                <i icon-name="arrow-right" class="inline-block"></i>
                <input id="replace_region_to"
                       placeholder="{% translate "New region slug" %}"
                       class="pr-8 appearance-none block bg-white text-gray-800 placeholder-gray-800 border border-gray-400 rounded py-3 px-4 leading-tight"/>
                <button id="region-replace-btn"
                        class="btn"
                        data-url="{% url 'replace_region_link' region_slug=request.region.slug %}">
                    <i icon-name="link"></i>
                    {% translate "Replace links" %}
                </button>
```

MizukiTemma:
Updates after the discussion in chat.

1) In future an automatic link replacement will be introduced in the region renaming process etc, yet this replacement feature is to implement because 
     a) there are already links that went broken becuase of region renaming etc.
     b) this feature can be used to replace not only links to Integreat contents but also for E-mail adresses and phone numbers.

2) To avoid the problem that external links become replaced too, `replace_links` function is extended by a new optional parameter `only_internal`, which decides only internal links shoud be replaced or external links too.

3) To achieve the extension of 1-b, this feature is modified so it generally replaces matched parts of links. Now E-mail addresses and phone numbers can be updated.

4) To show users how the replacement works, some examples are added. I hope this is intuitive enough.

integreat_cms/core/settings.py
```diff
@@ -923,6 +923,8 @@
 # DJANGO LINKCHECK #
 ####################
 
+LINKCHECK_SITE_DOMAINS = [WEBAPP_URL]
```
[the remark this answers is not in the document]
timoludwig: It is used by the linkcheck library: https://github.com/DjangoAdminHackers/django-linkcheck#site_domain-and-linkcheck_site_domains

MizukiTemma:
@Gaston69 @PeterNerlich Thank you for review! I applied your suggestions 😃  The test will be adjusted after #2207 is merged.

integreat_cms/cms/templates/linkcheck/links_by_filter.html
```diff
@@ -173,4 +173,63 @@ <h1 class="heading">
...
+                    <span class="bg-red-300">/augsburg/de/willkommen</span> -> <span class="bg-green-300">/augsburg/de/hallo</span>
+                    <div class="pl-10">
+                        https://integreat.app<span class="bg-red-300">/augsburg/de/willkommen</span>-in-augsburg -> https://integreat.app<span class="bg-green-300">/augsburg/de/hallo</span>-in-nurnberg
+                    </div>
+                </div>
+                <span class="font-bold">{% translate "External links will not be changed." %}</span>
+            </div>
+            <div class="flex mt-2 mb-2 gap-2">
+                <input id="replace_region_from"
+                       placeholder="{% translate "Old region slug" %}"
```
timoludwig, line 206:
```suggestion
                       placeholder="{% translate "Search" %}"
```

integreat_cms/cms/templates/linkcheck/links_by_filter.html
```diff
@@ -173,4 +173,63 @@ <h1 class="heading">
...
+                       class="pr-8 appearance-none block bg-gray text-gray-800 placeholder-gray-800 border border-gray-400 rounded py-3 px-4 leading-tight" />
+                <i icon-name="arrow-right" class="mt-auto mb-auto"></i>
+                <input id="replace_region_to"
+                       placeholder="{% translate "New region slug" %}"
```
timoludwig, line 210:
```suggestion
                       placeholder="{% translate "Replace" %}"
```

integreat_cms/cms/templates/linkcheck/links_by_filter.html
```diff
@@ -173,4 +173,63 @@ <h1 class="heading">
...
+                    <span>{% translate "Example 3: " %}</span>
+                    <span class="bg-red-300">/augsburg/de/willkommen</span> -> <span class="bg-green-300">/augsburg/de/hallo</span>
+                    <div class="pl-10">
+                        https://integreat.app<span class="bg-red-300">/augsburg/de/willkommen</span>-in-augsburg -> https://integreat.app<span class="bg-green-300">/augsburg/de/hallo</span>-in-nurnberg
+                    </div>
+                </div>
+                <span class="font-bold">{% translate "External links will not be changed." %}</span>
+            </div>
+            <div class="flex mt-2 mb-2 gap-2">
+                <input id="replace_region_from"
```
timoludwig, line 205: You could also define these input fields via a real Django form and render them with `{% render_field %}` and perform the replacement step inside the form's `save()` method - I guess this would be a little more Django-standard...

integreat_cms/static/src/js/link-check/link-replace.ts
timoludwig, line 1: If we have the replacement on a separate page, we don't need to submit the form via ajax... I think a normal post-request to submit the form is fine.

integreat_cms/cms/utils/linkcheck_utils.py
```diff
@@ -229,7 +236,9 @@ def replace_links(search, replace, region=None, user=None, commit=True):
                 new_translation = deepcopy(translation)
                 for link in translation.links.select_related("url"):
                     url = link.url.url
-                    if search in url:
+                    if search in url and (
+                        not only_internal or link.url.type != "external"
```
timoludwig, line 240: Since the scope of this issue changed a little bit (renaming regions is no longer the main use case), maybe allowing the replacement of external links as well isn't too bad? E.g. if we have many external links:

https://old-domain.tld/first-page
https://old-domain.tld/second-page

and they need to be replaced with:

https://new-domain.tld/first-page
https://new-domain.tld/second-page

But I wonder whether we should make this configurable by the user? E.g. a select field to limit the replacement to a specific link type, including an "all types" option to not limiting it?
Not sure though, I guess leaving it as it is for the moment is fine, and we can still think about extending it in the future...

integreat_cms/cms/views/linkcheck/linkcheck_actions.py
```diff
@@ -0,0 +1,43 @@
...
+
+    if not search or not replace:
+        return JsonResponse(
+            data={
+                "response": "Fail",
+            },
+            status=400,
+        )
+
+    replace_links(search, replace, region, commit=True, only_internal=True)
```
timoludwig, line 38: It would be really amazing if we could have something like a confirmation-stage, where the replacement is run with `commit=False` and the changes are only printed out, and not yet performed... But I guess this would require some refactoring of the existing function, so let's open a new issue for this once this is done :smile:

integreat_cms/release_notes/current/unreleased/1851.yml
```diff
@@ -0,0 +1,2 @@
+en: Add a region link replacement function
+de: Füge Region-Link-Ersetzungsfunktion hinzu
```
timoludwig, line 2:
```suggestion
en: Add search & replace function to link checker
de: Ermögliche Suchen & Ersetzen im Link-Checker
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -5647,6 +5647,58 @@ msgstr "Erneut prüfen"
...
+msgid "Replace region link"
+msgstr "Regionlink ersetzen"
+
+#: cms/templates/linkcheck/links_by_filter.html
+msgid ""
+"Here you can replace internal links, e-mail addresses and phone numbers "
+"partially or completely."
+msgstr ""
+"Here können Sie interne Links, E-mail Adresse und Telefonnummer partiell "
+"oder komplett ersetzen."
```
timoludwig, line 5660:
```suggestion
"Hier können Sie interne Links, E-Mail-Adressen und Telefonnummern teilweise "
"oder komplett ersetzen."
```

timoludwig (changes requested):
Thanks for you changes! :muscle:

In general it looks good, my main concern is that it makes the normal link list a bit overloaded, so maybe move it to a separate page?

integreat_cms/cms/utils/linkcheck_utils.py
MizukiTemma: Choice of link type is now available. I hope this helps users to avoid replacing a lot of links accidentally (like, many links to the website of city office or health department get replaced together while users wanted to replace only internal links..)

integreat_cms/cms/forms/linkcheck/link_replace_form.py
```diff
@@ -0,0 +1,49 @@
...
+        label=_("Replace"),
+        max_length=64,
+        help_text=_("Enter new link to replace the old link with"),
+    )
+    link_types = forms.MultipleChoiceField(
+        label=_("Link type"),
+        widget=forms.CheckboxSelectMultiple(attrs={"class": "default-checked"}),
+        choices=LINK_TYPES,
+        required=True,
+    )
```
timoludwig, line 26: Could you make "internal" checked by default?

integreat_cms/cms/forms/linkcheck/link_replace_form.py
```diff
@@ -0,0 +1,49 @@
...
+        label=_("Search"), max_length=64, help_text=_("Enter old link to be replaced")
+    )
+    replace = forms.CharField(
+        label=_("Replace"),
+        max_length=64,
+        help_text=_("Enter new link to replace the old link with"),
+    )
+    link_types = forms.MultipleChoiceField(
+        label=_("Link type"),
+        widget=forms.CheckboxSelectMultiple(attrs={"class": "default-checked"}),
```
timoludwig, line 23: The class "default-checked" is not required here I think...

timoludwig (changes requested):
Awesome, I like the new view and the option to select the link types! :rocket:

integreat_cms/cms/templates/linkcheck/links_by_filter.html
```diff
@@ -2,23 +2,31 @@
...
+            {% elif LINKCHECK_PHONE_ENABLED and view.kwargs.url_filter == 'phone' %}
+                {% translate "Phone links" %}
+            {% else %}
+                {% translate "All links" %}
+            {% endif %}
+        </h1>
+        {% if request.user.expert_mode %}
+            <a href="{% url 'search_and_replace_link' region_slug=request.region.slug %}"
+               class="font-bold text-sm text-gray-800 flex items-center gap-1 mb-2 hover:underline">
+                <span><i icon-name="replace" class="align-top h-5"></i> {% translate "Search & replace" %}</span>
```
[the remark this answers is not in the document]
MizukiTemma: Ohw 🙈 I didn't notice the icon wasn't shown...

integreat_cms/cms/views/linkcheck/link_replace_view.py
```diff
@@ -0,0 +1,96 @@
...
+        :rtype: ~django.http.HttpResponseRedirect
+        """
+        form = LinkReplaceForm(data=request.POST, region=self.request.region)
+
+        if not form.is_valid():
+            messages.error(
+                request,
+                _("Links could not be replaced."),
+                _("Please try again later or contact an administrator."),
+            )
```
timoludwig, line 74:
```suggestion
            for field in form:
                for error in field.errors:
                    messages.error(request, f"{_(field.label)}: {_(error)}")
            for error in form.non_field_errors():
                messages.error(request, _(error))
```
I think it makes more sense to tell the user what exactly went wrong

timoludwig (changes requested):
Thanks! :+1:
Just a small detail I'd like to change...

timoludwig (approved):
:tada:

Issue #2198, opened 2023-04-04
Unable to replace malformed `mailto:` or `tel:` links


timoludwig:
### Describe the Bug

Email and tel links are hidden in the broken link checker, so normally they do not show up.
However, when they are malformed in any way, e.g. a link was created `https://tel:12345`, it's impossible to fix this problem in the broken link checker, because it only accepts valid URLs, not valid email or tel links.

### Steps to Reproduce

1. Create malformed tel link in content
2. Try to fix problem in broken link checker
3. See validation error

### Expected Behavior

It should be possible to replace incorrect links, even if they use `mailto:` or `tel:` targets.

### Actual Behavior

Only links with `http://`/`https://` can be replaced

PR #2199, opened 2023-04-04
Add emails and phone link checker

timobrembeck:
### Short description

Add the possibility to replace `mailto:` and `tel:` links with the link checker

### Proposed changes

- Don't ignore them by default anymore (but leave the possibility to ignore them in the future again) 
- Extend the validation to support `mailto:` and `tel:` links
  - Enforce email validation on mailto links
  - Do not enforce validation on phone links

Caveats: Only the link target is replaced, not the link text. So if this feature is used without caution, it might create inconsistencies (e.g. tel links with a different phone number target than their displayed text etc.)... Thus, I decided to only make the new categories visible for expert users.

### Resolved issues

Fixes: #2198

### DIFF
--- a/integreat_cms/cms/forms/linkcheck/edit_url_form.py
+++ b/integreat_cms/cms/forms/linkcheck/edit_url_form.py
@@ -2,6 +2,7 @@ import logging
 
 from django import forms
 from django.core.validators import EmailValidator, URLValidator
+from django.utils.translation import gettext as _
 
 logger = logging.getLogger(__name__)
 
@@ -73,3 +74,4 @@ class EditUrlForm(forms.Form):
     """
 
     url = LinkField()
+    text = forms.CharField(widget=forms.TextInput(attrs={"readonly": "readonly"}))
--- a/integreat_cms/cms/templates/linkcheck/link_list_row.html
+++ b/integreat_cms/cms/templates/linkcheck/link_list_row.html
@@ -5,102 +5,120 @@
 {% load widget_tweaks %}
 {% load linkcheck_filters %}
 {% object_translation_has_view_perm request.user url.region_links.0.content_object as show_source_link %}
-<tr class="border-t border-gray-200 hover:bg-gray-100 text-gray-800"
-    {% if url.id == view.kwargs.url_id %} id="replace-url"{% endif %}>
-    <td class="py-3 pl-4">
-        <input type="checkbox"
-               name="selected_ids[]"
-               value="{{ url.id }}"
-               class="bulk-select-item"/>
-    </td>
-    <td class="pr-2 whitespace-nowrap max-w-[75px] sm:max-w-[100px] md:max-w-[100px] lg:max-w-[100px] xl:max-w-[150px] 2xl:max-w-[200px] 3xl:max-w-[260px] 4xl:max-w-[400px] overflow-hidden text-ellipsis"
-        title="{{ url.url }}">
-        <a href="{{ url.url }}"
-           target="_blank"
-           rel="noopener noreferrer"
-           class="text-blue-500 hover:underline">{{ url.url }}</a>
-        {% if url.redirect_to %}
-            <br />
-            <i icon-name="corner-down-right"></i>
-            <a href="{{ url.redirect_to }}"
+<tbody class="hover:bg-gray-100">
+    <tr class="border-t border-gray-200 text-gray-800"
+        {% if url.id == view.kwargs.url_id %} id="replace-url"{% endif %}>
+        <td class="py-3 pl-4">
+            <input type="checkbox"
+                   name="selected_ids[]"
+                   value="{{ url.id }}"
+                   class="bulk-select-item"/>
+        </td>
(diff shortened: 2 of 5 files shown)


seluianova:
> Caveats: Only the link target is replaced, not the link text. So if this feature is used without caution, it might create inconsistencies (e.g. tel links with a different phone number target than their displayed text etc.)... Thus, I decided to only make the new categories visible for expert users.

There is the same thing with URLs. If text is the same as the link, non-expert users can make an inconsistency too.
Do we want to hide only phones and mails?
Also, if phones/mails are broken, they are still available under "Invalid" section for non-expert users.
So do we really need this limitation?

integreat_cms/cms/forms/linkcheck/edit_url_form.py
```diff
@@ -1,9 +1,74 @@
...
+        :rtype: str
+        """
+        if value.startswith("mailto:"):
+            email = value[7:]
+            logger.debug(
+                "Value %r is an email link, enforcing EmailValidator on %r",
+                value,
+                email,
+            )
+            self.validators.append(EmailValidator())
```
seluianova, line 58: > Enforce email validation on mailto links
I tried to add an invalid email, but it's displayed as a normal one.
Is it expected? Did I misunderstood something? 
image

timoludwig: Ah yeah, this was maybe a bit misleading, the email validation is only made when trying to replace an email.
Maybe this can be improved after #2207 is merged, but for now it's ok I guess.

timoludwig:
> If phones/mails are broken, they are still available under "Invalid" section for non-expert users (thus can be damaged).

True, but I'd argue that an invalid phone/mail link is important enough to appear in the invalid section, so that also non-experts are aware of the issue.

> Also there is the same thing with URLs. If text is the same as the link, non-expert users can make an inconsistency too.
> Was it considered?

Yes, but I think for links the link text is only identical in very rare cases, whereas for mails and phone numbers this will be the case in nearly all links (at least this is my assumption)... I'd say it's a good enough starting point, maybe the service team can provide feedback after we tested this for a while...

example-configs/integreat-cms.ini
```diff
@@ -115,6 +115,10 @@ EMAIL_USE_SSL = False
 [linkcheck]
 # Whether link check should be disabled [optional, defaults to False]
 LINKCHECK_DISABLE_LISTENERS = False
+# URL types that are not supposed to be shown in the link list (e.g. phone numbers and emails)
+LINKCHECK_IGNORED_URL_TYPES =
```
seluianova, line 119: Do I understand correctly that these settings should be applied automatically, or do I have to do something extra for that?
Currently the value of settings.LINKCHECK_IGNORED_URL_TYPES for me is resolved as [].

timoludwig: The .ini file in the repository is just an example config to illustrate what values are possible.
I added this so we can quickly return to ignoring emails and phone numbers if it's too confusing for the users, but per default I wouldn't ignore them, that's why I set the empty list as default value in the settings file.

seluianova: Ah, okay, thanks. I thought this config is loaded somehow

seluianova (approved):
Looks good to me as far as I tested 👌

I would only note one thing, not sure how important:
If I try to save an invalid email, the error is "Url: Enter a valid URL"
I assume the expected one would be "Enter a valid email address"
image

JoeyStk (approved):
Thank you :tada: 
As far as I tested it, it seems to work. I have two thoughts, but I'm not sure how relevant they are. 
1. `Caveats: Only the link target is replaced, not the link text.` I found this piece of information very important to know. So I think it definitely should be documented. If you haven't already, I also would think about maybe adding a warning about this inside the CMS, probably rather just a small note after clicking on the replace button? 
2.  I noticed, that all email addresses I entered were found in the link checker, valid and invalid ones. I assumed, that this is the expected behavior, since I can just ignore the ones that are correct, but I thought it was still worth mentioning it

timoludwig:
> I would only note one thing, not sure how important: If I try to save an invalid email, the error is "Url: Enter a valid URL" I assume the expected one would be "Enter a valid email address"

@seluianova Good idea, I adapted the error message in this case.

> 1. `Caveats: Only the link target is replaced, not the link text.` I found this piece of information very important to know. So I think it definitely should be documented. If you haven't already, I also would think about maybe adding a warning about this inside the CMS, probably rather just a small note after clicking on the replace button?

@JoeyStk Yes, makes sense. I agree that it might be confusing, especially if the link text is a URL itself and not a descriptive title. I added a second field now that indicates that the link text is read-only, what do you think?

link-text-replace-disabled

> 2. I noticed, that all email addresses I entered were found in the link checker, valid and invalid ones. I assumed, that this is the expected behavior, since I can just ignore the ones that are correct, but I thought it was still worth mentioning it

Indeed, since the link checker itself does not check them (as indicated by the message "Email link (not automatically checked)")... maybe we could provide an upstream contribution to https://github.com/DjangoAdminHackers/django-linkcheck to also include (at least format-checking) of email and phone number links in the future...

Issue #2109, opened 2023-03-06
Archive feedback instead of deleting it


osmers:
### Motivation
Regions want a possibility of shortening the list of viewable feedback without deleting read feedback immediately. 

### Proposed Solution
Regions should have the option of archiving feedback instead of deleting it - archived tab as already implemented under pages should be created

### Alternatives
Keep it as it is with regions either deleting feedback or marking it as read.

### Additional Context
![image](https://user-images.githubusercontent.com/64901227/223234830-27cee58b-aec0-4a11-8f14-49da2a8c4ac7.png)
![image](https://user-images.githubusercontent.com/64901227/223234897-7f03dc8d-ec91-44f2-a66d-9662e089db22.png)

PR #2202, opened 2023-04-05
Add the ability to archive feedback

seluianova:
### Short description
Regions want a possibility of shortening the list of viewable feedback without deleting read feedback immediately.

### Proposed changes
- Add the ability to archive and restore feedbacks.
- Make it available to users with change_feedback permission (by analogy with other entities, is it okay?)

### Side effects
I didn't find any

### Resolved issues
Fixes: #2109 


integreat_cms/cms/templates/feedback/admin_feedback_list.html
```diff
@@ -147,6 +157,9 @@ <h1 class="heading">{% translate "Technical Feedback" %}</h1>
                     <option data-bulk-action="{% url 'mark_admin_feedback_as_unread' %}">
                         {% translate "Mark as unread" %}
                     </option>
+                    <option data-bulk-action="{% url 'archive_admin_feedback' %}">
+                        {% translate "Archive" %}
+                    </option>
```
MizukiTemma, line 162:
```suggestion
                    {% if perms.cms.change_feedback %}
                        <option data-bulk-action="{% url 'archive_admin_feedback' %}">
                            {% translate "Archive" %}
                        </option>
                    {% endif %}
```

Marketing team users can view but cannot change feedback. I think we can hide the archive bulk action to them.

integreat_cms/cms/templates/feedback/admin_feedback_list_archived.html
```diff
@@ -0,0 +1,166 @@
...
+                        {% translate "Restore" %}
+                    </option>
+                    {% if perms.cms.delete_feedback %}
+                        <option data-bulk-action="{% url 'delete_admin_feedback' %}">
+                            {% translate "Delete" %}
+                        </option>
+                    {% endif %}
+                </select>
+                <button id="bulk-action-execute" class="btn" disabled>{% translate "Execute" %}</button>
+            </div>
```
MizukiTemma, line 163:
```suggestion
            {% if perms.cms.change_feedback or perms.cms.delete_feedback %}
                <div class="flex self-start gap-2 mt-2">
                    <select id="bulk-action">
                        <option>
                            {% translate "Select bulk action" %}
                        </option>
                        {% if perms.cms.change_feedback %}
                            <option data-bulk-action="{% url 'restore_admin_feedback' %}">
                                {% translate "Restore" %}
                            </option>
                        {% endif %}
                        {% if perms.cms.delete_feedback %}
                            <option data-bulk-action="{% url 'delete_admin_feedback' %}">
                                {% translate "Delete" %}
                            </option>
                        {% endif %}
                    </select>
                    <button id="bulk-action-execute" class="btn" disabled>{% translate "Execute" %}</button>
                </div>
            {% endif %}
```

Here it's similar but we can here additionally hide the entire bulk action for marketing team users because both the current options are not available for them.

MizukiTemma (changes requested):
Thank you for PR 🙂 So far I tested it works as expected.

I have only two small suggestion for showing/hiding archive/restore options depend on permissions users have.

integreat_cms/cms/templates/feedback/admin_feedback_list_archived.html
```diff
@@ -0,0 +1,166 @@
+{% extends "_base.html" %}
+{% load i18n %}
+{% load static %}
+{% load widget_tweaks %}
+{% load text_filters %}
+{% block content %}
+    {% with filter_form.filters_visible as filters_visible %}
+        <div class="table-header">
+            <div class="flex flex-wrap justify-between">
+                <h1 class="heading">{% translate "Archived feedbacks" %}</h1>
```
JoeyStk, line 10:
```suggestion
                <h1 class="heading">{% translate "Archived feedback" %}</h1>
```
This is picky, but as far as I could research the plural of feedback is feedback

seluianova: Just a matter of curiosity: if I say in German “die Feedbacks”, plural has “s”, right?

integreat_cms/cms/templates/feedback/admin_feedback_list.html
seluianova: Ah, I guess then we can hide the whole list of bulk actions, because they don't have access to any actions

integreat_cms/cms/templates/feedback/admin_feedback_list_archived.html
seluianova: As I understand, if user has the delete_feedback permission, he also has the change_feedback permission.
So I only keep change_feedback condition.
Pls let me know if you actually think that both conditions are must.

JoeyStk (approved):
Very cool! :tada:

integreat_cms/release_notes/current/unreleased/2109.yml
```diff
@@ -0,0 +1,2 @@
+en: Add the ability to archive feedback
+de: Hinzufügen der Möglichkeit, Feedback zu archivieren
```
timoludwig, line 2:
```suggestion
de: Füge die Möglichkeit hinzu, Feedback zu archivieren
```

timoludwig (approved):
Thanks! :+1:

MizukiTemma (approved):
🎉

PR #2211, opened 2023-04-05
Fix pylint useless suppression warnings

charludo:
### Short description

Fix pylint useless suppression warnings

### Proposed changes

- remove the useless suppressions

### Side effects

- none

### Resolved issues

Fixes: n/a

### DIFF
--- a/integreat_cms/cms/models/abstract_content_translation.py
+++ b/integreat_cms/cms/models/abstract_content_translation.py
@@ -425,7 +425,6 @@ class AbstractContentTranslation(AbstractBaseModel):
         """
         return self.translation_state == translation_status.UP_TO_DATE
 
-    # pylint: disable=too-many-return-statements
     @cached_property
     def translation_state(self):
         """
--- a/integreat_cms/cms/views/pois/poi_form_ajax_view.py
+++ b/integreat_cms/cms/views/pois/poi_form_ajax_view.py
@@ -45,7 +45,6 @@ class POIFormAjaxView(TemplateView, POIContextMixin):
             },
         )
 
-    # pylint: disable=unused-argument
     def post(self, request, *args, **kwargs):
         r"""Add a new POI to the database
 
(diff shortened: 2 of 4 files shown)


timoludwig (approved):
Thanks!
Is there any possibility to make the pylint job fail when this warning is thrown? Otherwise we will probably re-introduce similar useless suppression statements in the future...

charludo:
> he `enable` to enable the warning and `fail-on` to set the return code

tried that as well, same result. Using only `fail-on` was more of a sanity check to see if pylint even recognized the setting

> > Good idea! It _should_ be possible, see: [pylint-dev/pylint#4398](https://github.com/pylint-dev/pylint/pull/4398), and changing `enable` to `fail-on` in `pyproject.toml` still keeps the warnings, but the exit code in both cases is `0`.
> 
> I think it's not an either/or, but it needs both statements... the `enable` to enable the warning and `fail-on` to set the return code... Also, I think the `fail-on` needs to be in the `[tool.pylint.main]` section?
> 
> Just setting
> 
> ```
> fail-on = [
>     "all"
> ]
> ```
> 
> might work as well?

Good idea, I'll have a try...

charludo:
> > he `enable` to enable the warning and `fail-on` to set the return code
> 
> tried that as well, same result. Using only `fail-on` was more of a sanity check to see if pylint even recognized the setting
> 
> > Just setting
> > ```
> > fail-on = [
> >     "all"
> > ]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > might work as well?
> 
> Good idea, I'll have a try...

No luck, unfortunately

MizukiTemma (approved):
✅ thank you 🙂

timoludwig:
> If it's alright with you, I'll merge as-is, without squashing, since the commits are all about different "issues"

I agree :+1:

Issue #2197, opened 2023-04-04
Event API caches old result after setting an event to draft


timoludwig:
### Describe the Bug

Events are still delivered via the events API when the result is cached via django-cachalot.

### Steps to Reproduce

1. Enable redis cache
2. Fetch events API to populate cache
3. Set event to draft
4. Fetch events API again

### Expected Behavior

An event that is set to draft should not be delivered via the API

### Actual Behavior

The event is still delivered when the result is cached and not manually invalidated

PR #2215, opened 2023-04-06
Invalidate cache after events have been updated

timoludwig:
### Short description

I have to admit that I neither understand why the bug is happening, nor why exactly this line is fixing it (I also tried `invalidate_obj(region)`, `invalidate_obj(event)`, `invalidate_obj(event_translation)`, `invalidate_model(Region)`, `invalidate_model(Event)`, neither of which worked), but the important thing is that it seems to fix the issue :see_no_evil: 

### Proposed changes

- Invalidate event translation cache after events have been updated

### Side effects

- The cache in all regions is invalidated when an event in one region is updated

### Resolved issues

Fixes: #2197

### DIFF
--- a/integreat_cms/cms/views/events/event_form_view.py
+++ b/integreat_cms/cms/views/events/event_form_view.py
@@ -1,5 +1,6 @@
 import logging
 
+from cacheops import invalidate_model
 from django.conf import settings
 from django.contrib import messages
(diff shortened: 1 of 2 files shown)


ulliholtgrave (approved):
Thanks, seems to work with my local Redis setup :+1:

Issue #2190, opened 2023-04-03
Bumping dev versions not working in CircleCI when prod release was first in month


timobrembeck:
### Describe the Bug

When the first release of a month is a prod release and not a dev release, the CircleCI dev workflow fails for the subsequent run.

### Steps to Reproduce

1. Create a production release (by merging a PR into main) in a month before creating a dev release (by merging a PR into develop)
2. Merge a PR into develop
3. See failing build

### Expected Behavior

The build should succeed

### Actual Behavior

The build fails, see e.g. https://app.circleci.com/pipelines/github/digitalfabrik/integreat-cms/8235/workflows/4e96f674-fa79-41e8-be3a-5b32f525b0a0/jobs/86178
```
ERROR   - Invariant violated: New version must be greater than old version 
ERROR   -   Failed Invariant: '2023.3.44-alpha' > '2023.4.0'
ERROR   - Invalid argument --set-version='2023.3.44-alpha'
```

timobrembeck:
Apparently, this is still a thing:
https://app.circleci.com/pipelines/github/digitalfabrik/integreat-cms/11446/workflows/10a093bd-2d2f-46d6-8252-6329d1e9c2d0/jobs/138736

PR #2216, opened 2023-04-06
Fix dev version bump when latest release is newer

timoludwig:
### Short description

Fix dev version bump when latest release is newer

### Proposed changes

- Only bump the dev version to the current alpha version if this version is newer than the current production version

### Resolved issues

Fixes: #2190

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -419,9 +419,14 @@ jobs:
             echo "Most recent version on TestPyPI: ${CURRENT_ALPHA_VERSION}"
             CURRENT_ALPHA_VERSION=$(echo "${CURRENT_ALPHA_VERSION}" | sed "s/integreat-cms (\([^()]*\)a0)/\1-alpha/")
             echo "Version converted to alternative format: ${CURRENT_ALPHA_VERSION}"
-            # Bump version to current alpha version
-            echo "Bump to the currently existing version"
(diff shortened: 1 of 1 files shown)


MizukiTemma (approved):

ulliholtgrave (commented):
I am not really sure if this fixes the issue I am thinking of :thinking: 

Even if I restart the CircleCI jobs the following package will be built:
`Successfully built integreat-cms-2023.4.0.tar.gz and integreat_cms-2023.4.0-py3-none-any.whl

CircleCI received exit code 0
`
However, our develop branch and the version on pypi is at least 2023.4.3. Do I miss something or is this only executed if we merge on develop?

timoludwig:
> Do I miss something or is this only executed if we merge on develop?

Exactly, as you can see here:
https://github.com/digitalfabrik/integreat-cms/blob/cbb1dca55b25b8b013789df9f0fcd777bc4c3f46/.circleci/config.yml#L597-L602
The `bump-dev-version` job is only executed on workflows triggered on the `develop` branch or branches ending in `-publish-dev-package`.

Also, you wouldn't notice a difference since the bug is only triggered when a production release was the first in a month and there are no dev versions in the current month.
But we can try to validate this fix by preparing a prod release on the first of May without merging any PR into develop on that day.

ulliholtgrave (approved):
Thanks for the explanation @timoludwig . This is good to go :+1:

Issue #2102, opened 2023-03-03
Add DeepL bulk action to the page tree view


charludo:
### Motivation

We currently lack a way of bulk-translating pages into a supported language, even though the same feature exists for POIs and events. While a more in-depth translation manager is still being debated, we want to at least be able to offer a bulk action for this that is consistent with the one for POIs/events.  

### Proposed Solution

- enable the bulk action in the `page_tree` template (see existing implementations for other content types)
- add a confirmation/summary popup containing all relevant information (pages which will be translated, pages which cannot be translated due to poor hix value (only if source language is German!), remaining budget and how much of it will be used by the bulk action), as well as the option to change source and target languages (see mockup by @JoeyStk below)
- if the HIX value calculation is disabled for a page (hix_ignored=true), the page should not be translated via DeepL (see #2075)
- if the budget allows for the bulk action, enable the confirmation button

### Alternatives

- automatically un-select all low-HIX pages when this bulk action is selected

### Additional Context

Blocked by:
- [x] #2101 

### Design Requirements

(**update, not in mockup:** move budget section up above the untranslatable pages)
![image](https://user-images.githubusercontent.com/47758554/223746248-e811b4a0-dac1-4803-9e96-6419d7fcd3b4.png)

PR #2221, opened 2023-04-10
Add popup to DeepL bulk action of page tree view

JoeyStk:
### Short description
This PR adds the option to translate pages via MT through a bulk action inside the page tree view. Please note, this PR is not yet ready for review, since I still want to take the time to attempt refactoring it myself. This PR is supposed to fulfill the following use case: I, as CMS user, want to get an overview about my MT budget, the pages that will be able to be translated and those that can't be translated (due to too low HIX score), before translating them.

### Proposed changes

- Add release note
- Add a option for MT to the bulk action dropdown
- Add a popup to get a report about the attempted MT
  - Show from which to which language this translation is going to be
  - Show the pages that can be translated
  - Show the pages that can't be translated
  - Show the budget and calculate, if there is enough budget for it
- Request hix score and amount of words per page via AJAX  

### Background information
- After a chat with @osmers we decided to disable the option to switch the language inside the popup, since a. we don't want to encourage this behavior, and it would have increased the scope of this PR quite a bit
- After a chat with @hauf-toni we decided to keep most of the structure for the popup in comparison to the XLIFF popup, but add two significant changes: 1. remove the X to close the popup and instead adding a "Cancel" button, since it's more prominent and gives clearer instruction to the user of what to do and 2. widen the popup from 400px to 500px, because 400px were too narrow and added a lot of additional vertical scroll. 

### Side effects

- This could be effecting the MT via page, event or poi form
- There might be a bunch of merge conflicts after #2209 is merged
- Let's say you have 1 page that can be translated and one that can't be, you'll get two error messages for the page that can't be translated (one inside the popup via JS/TS and the second one after confirming via Dj/Py).
- I noticed, that currently I can re-translate a page via MT although the translation status is "Up to date" or "Machine translated and up to date". In order to be more user friendly and prevent unnecessary usage of MT budget, I would suggest adding a check if a machine translation is actually needed or not. First I think it's up for discussion if we really want this or not, and secondly I think that would also be out of scope for this PR and rather something for another issue  

### Resolved issues

Fixes: #2102 

### DIFF
--- a/integreat_cms/cms/templates/_machine_translation_overlay.html
+++ b/integreat_cms/cms/templates/_machine_translation_overlay.html
@@ -1,0 +1,117 @@
+{% load i18n %}
+<div id="machine-translation-overlay"
+     class="hidden fixed items-center justify-center inset-0 bg-opacity-75 bg-gray-800 z-50 cursor-pointer">
+    <div class="cursor-auto justify-center w-[500px] max-h-fit px-[10px] z-50 m-auto">
+        <div class="bg-white opacity-100 content rounded shadow-md w-full">
+            <div class="flex items-center w-full rounded p-4 bg-water-500">
+                <h2 class="font-bold font-default">
+                    <i icon-name="bot" class="pb-1"></i>
+                    {% translate "Machine translation" %}
+                </h2>
+            </div>
+            <div class="w-full p-4 rounded shadow overflow-scroll max-h-[80vh]">
+                <span class="block">{% translate "This report shows a summary of your attempted machine translation. Please confirm if everything is correct" %}</span>
+                <div class="mt-4">
+                    <h3>
+                        <b>{% translate "Language selection" %}</b>
+                    </h3>
+                    <div class="flex items-center mt-2">
+                        <div class="flex">
+                            <span class="fp fp-rounded fp-{{ source_language.primary_country_code }} w-7 mr-2"></span>
+                            <p>{{ source_language }}</p>
+                        </div>
+                        <i class="mx-4" icon-name="arrow-big-right"></i>
+                        <div class="flex">
+                            <span class="fp fp-rounded fp-{{ language.primary_country_code }} w-7 mr-2"></span>
+                            <p>{{ language }}</p>
+                        </div>
+                    </div>
+                </div>
+                <div class="mt-4">
+                    <h3>
+                        <b>{% translate "Pages to be translated" %} (<span id="machine-translation-overlay-pages-number"></span>/<span id="machine-translation-overlay-pages-total-number"></span>)</b>
+                    </h3>
+                    <div class="mt-2">
+                        <div id="machine-translation-overlay-pages-no-pages-warning"
+                             class="hidden bg-red-100 border-l-4 border-red-500 text-red-500 px-4 py-3 mb-5">
+                            {% if source_language.slug in textlab_languages %}
+                                {% blocktranslate trimmed with hix_threshold=hix_threshold|floatformat:0 %}
+                                    None of the selected pages fullfills the required HIX score of {{ hix_threshold }}.
+                                {% endblocktranslate %}
+                            {% else %}
+                                {% translate "None of the selected pages has a source translation." %}
+                            {% endif %}
+                        </div>
+                        <ul id="machine-translation-overlay-pages">
+                        </ul>
+                        <ul id="machine-translation-overlay-pages-optional" class="hidden">
+                        </ul>
+                        <a id="machine-translation-overlay-expansion-trigger"
+                           data-default-text="{% translate "Show all" %}"
+                           data-alternative-text="{% translate "Show less" %}"
+                           class="text-blue-500 hidden">{% translate "Show all" %}</a>
+                    </div>
+                </div>
+                <div class="mt-4">
+                    <h3>
+                        <b>{% translate "Budget" %}</b>
+                    </h3>
+                    <table>
+                        <tr>
+                            <td id="machine-translation-overlay-current-budget-label">{% translate "Current budget" %}</td>
+                            <td class="pl-2" id="machine-translation-overlay-current-budget-result">
+                                {{ request.region.deepl_budget_remaining }}
+                            </td>
+                        </tr>
+                        <tr>
+                            <td id="machine-translation-overlay-budget-usage-label">{% translate "Number of words you would use" %}</td>
+                            <td class="pl-2" id="machine-translation-overlay-budget-usage-result"></td>
+                        </tr>
+                        <tr class="border-solid border-t-2">
+                            <td id="machine-translation-overlay-budget-remains-label">{% translate "Remaining budget: " %}</td>
+                            <td class="pl-2" id="machine-translation-overlay-budget-remains-result"></td>
+                        </tr>
+                    </table>
+                </div>
+                <div class="mt-4 hidden" id="machine-translation-overlay-warning">
+                    <h3>
+                        <b>{% translate "Pages that can't be translated" %}</b>
+                    </h3>
+                    <div class="my-2 bg-yellow-100 border-l-4 border-yellow-500 text-yellow-600 px-4 py-3 mb-5"
+                         role="alert">
+                        <i class="mr-2" icon-name="alert-triangle"></i>
+                        {% blocktranslate trimmed %}
+                            There are currently <span id="not-translatable"></span> page(s) that can't be translated.
+                        {% endblocktranslate %}
+                        {% if source_language.slug in textlab_languages %}
+                            {% blocktranslate trimmed with hix_threshold=hix_threshold|floatformat:0 %}
+                                These pages can't be translated because the text understandibility is too low or because their hix value has been ignored. Please improve it to a value of at least {{ hix_threshold }} to translate these pages. You can identify the pages that couldn't be translated in the page tree via the translation status.
+                            {% endblocktranslate %}
+                        {% else %}
+                            {% translate "These pages can't be translated because they have no translation in the source language." %}
+                        {% endif %}
+                    </div>
+                    <ul class="hidden mb-2" id="machine-translation-overlay-warning-optional">
+                    </ul>
+                    <a id="machine-translation-overlay-warning-more" data-default-text="{% translate "Show pages" %}" data-alternative-text="{% translate "Don't show pages" %}"  class="text-blue-500">
+                        {% translate "Show pages" %}
+                    </a>
+                </div>
+                <div class="flex justify-between mt-4">
+                    <button type="button"
+                            id="btn-close-machine-translation-overlay"
+                            class="btn !bg-red-500">
+                        {% translate "Cancel" %}
+                    </button>
+                    <button id="machine-translation-overlay-bulk-action-execute"
+                            class="btn"
+                            form="bulk-action-form"
+                            type="submit"
+                            disabled>
+                        {% translate "Confirm" %}
+                    </button>
+                </div>
+            </div>
+        </div>
+    </div>
+</div>
--- a/integreat_cms/cms/views/pages/page_actions.py
+++ b/integreat_cms/cms/views/pages/page_actions.py
@@ -19,6 +19,7 @@ from django.views.decorators.http import require_POST
 from treebeard.exceptions import InvalidMoveToDescendant, InvalidPosition
 
 from ....api.decorators import json_response
+from ....textlab_api.utils import check_hix_score
 from ...constants import text_directions
 from ...decorators import permission_required
 from ...forms import PageForm
@@ -896,3 +897,70 @@ def refresh_date(
             "region_slug": request.region.slug,
         },
     )
+
+
+@require_POST
+@json_response
+# pylint: disable=unused-argument
+def post_hix_and_word_number_per_page(request, region_slug, language_slug):
+    """
+    This function collects the hix score and the amount of words per page from the source
+
+    :param request: The current request
+    :type request: ~django.http.HttpRequest
+
+    :param region_slug: The slug of the current region
+    :type region_slug: str
+
+    :param language_slug: The slug of the current language
+    :type language_slug: str
+
+    :return: A json response of {{"id": id, "title": title, "words": number of words, "hix": if HIX value is high enough}} in case of success
+    :rtype: ~django.http.JsonResponse
+    """
+    selected_page_ids = json.loads(request.body.decode("utf-8"))
+    selected_pages = request.region.get_pages(return_unrestricted_queryset=True).filter(
+        id__in=selected_page_ids
+    )
+
+    filtered_pages = {}
+    translatable = {}
+    not_translatable = {}
+
+    source_language = request.region.get_source_language(language_slug)
+
+    for page in selected_pages:
+        source_translation = page.get_translation(source_language.slug)
+        words = (
+            len(source_translation.content.split())
+            + len(source_translation.title.split())
+            if source_translation
+            else 0
+        )
+
+        if source_translation and check_hix_score(
+            request, source_translation, show_message=False
+        ):
+            page_data = {
+                "id": page.id,
+                "title": source_translation.title,
+                "words": words,
+                "hix": True,
+            }
+            translatable.update({page.id: page_data})
+        else:
+            page_data = {
+                "id": page.id,
+                "title": source_translation.title
+                if source_translation
+                else page.best_translation.title,
+                "words": words,
+                "hix": False,
+            }
+            not_translatable.update({page.id: page_data})
+
+    filtered_pages = {
+        "translatable": translatable,
+        "not_translatable": not_translatable,
+    }
+    return JsonResponse(filtered_pages)
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -3951,6 +3951,125 @@ msgstr "Standard-Sprache wird dupliziert"
 msgid "Other Languages"
 msgstr "Weitere Sprachen"
 
+#: cms/templates/_machine_translation_overlay.html
+msgid "Machine translation"
+msgstr "Maschinelle Übersetzung"
+
+#: cms/templates/_machine_translation_overlay.html
+msgid ""
+"This report shows a summary of your attempted machine translation. Please "
+"confirm if everything is correct"
+msgstr ""
+"Hier finden Sie eine Übersicht über die maschinelle Übersetzung. Bitte "
+"bestätigen Sie, dass alle Angaben korrekt sind."
+
+#: cms/templates/_machine_translation_overlay.html
+msgid "Language selection"
+msgstr "Sprachauswahl"
+
+#: cms/templates/_machine_translation_overlay.html
+msgid "Pages to be translated"
+msgstr "Zu übersetzende Seiten"
+
+#: cms/templates/_machine_translation_overlay.html
+#, python-format
+msgid ""
+"None of the selected pages fullfills the required HIX score of "
+"%(hix_threshold)s."
+msgstr ""
+"Keiner der ausgewählten Seiten erfüllt den benötigten HIX-Wert von "
+"%(hix_threshold)s."
+
+#: cms/templates/_machine_translation_overlay.html
+msgid "None of the selected pages has a source translation."
+msgstr "Keine der ausgewählten Seiten hat eine Quellübersetzung."
+
+#: cms/templates/_machine_translation_overlay.html
+msgid "Show all"
+msgstr "Alle anzeigen"
+
+#: cms/templates/_machine_translation_overlay.html
+msgid "Show less"
+msgstr "Weniger anzeigen"
+
+#: cms/templates/_machine_translation_overlay.html
+msgid "Budget"
+msgstr "Budget"
+
+#: cms/templates/_machine_translation_overlay.html
+msgid "Current budget"
+msgstr "Aktuelles Gesamtbudget"
+
+#: cms/templates/_machine_translation_overlay.html
+msgid "Number of words you would use"
+msgstr "Benötiges Budget"
+
+#: cms/templates/_machine_translation_overlay.html
+msgid "Remaining budget: "
+msgstr "Verbleibendes Budget:"
+
+#: cms/templates/_machine_translation_overlay.html
+msgid "Pages that can't be translated"
+msgstr "Seiten, die nicht übersetzt werden können"
+
+#: cms/templates/_machine_translation_overlay.html
+msgid ""
+"There are currently <span id=\"not-translatable\"></span> page(s) that can't "
+"be translated."
+msgstr ""
+"Es gibt aktuell <span id=\"not-translatable\"></span> Seiten, die nicht "
+"übersetzt werden können."
+
+#: cms/templates/_machine_translation_overlay.html
+#, python-format
+msgid ""
+"These pages can't be translated because the text understandibility is too "
+"low or because their hix value has been ignored. Please improve it to a "
+"value of at least %(hix_threshold)s to translate these pages. You can "
+"identify the pages that couldn't be translated in the page tree via the "
+"translation status."
+msgstr ""
+"Diese Seiten können nicht übersetzt werden, weil die Textverständlichkeit zu "
+"gering ist oder weil der HIX-Wert für diese Seiten ignoriert wurde. Bitte "
+"verbessern Sie den HIX-Wert auf einen Wert von mindestens %(hix_threshold)s, "
+"um diese Seiten übersetzen zu können. Sie können die Seiten, die nicht "
+"übersetzt werden konnten an ihrem Status im Seitenbaum erkennen."
+
+#: cms/templates/_machine_translation_overlay.html
+msgid ""
+"These pages can't be translated because they have no translation in the "
+"source language."
+msgstr ""
+"Diese Seiten können nicht übersetzt werden, weil sie keine Übersetzung in "
+"der Quellsprache haben."
+
+#: cms/templates/_machine_translation_overlay.html
+msgid "Show pages"
+msgstr "Seiten anzeigen"
+
+#: cms/templates/_machine_translation_overlay.html
+msgid "Don't show pages"
+msgstr "Seiten nicht anzeigen"
+
+#: cms/templates/_machine_translation_overlay.html
+#: cms/templates/_tinymce_config.html
+#: cms/templates/generic_confirmation_dialog.html
+#: cms/templates/linkcheck/link_list_row.html
+#: cms/templates/pois/poi_form_sidebar/position_box.html
+#: cms/views/pois/poi_context_mixin.py
+msgid "Cancel"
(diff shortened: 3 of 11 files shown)


charludo:
This is already looking fantastic! One issue I noticed (or maybe I'm just doing something wrong): The `Create missing translations automatically` bulk action is only shown to me on non-German page tree views, but selecting it then shows the error/warning about the too-low-for-translation HIX scores 😅

timoludwig (commented):
Wow, thanks a lot! :rocket:

The UI looks very cool! Before proceeding with the review, I think it would be good to resolve the conflicts. Since most of them are my fault, I could take care of that, or do you want to have a look yourself?

JoeyStk:
@timoludwig If you want to, feel free to resolve the conflicts. I'm a bit worried, that I might screw that up. If you don't find the time, I have a cautious attempt at it some time later this week

integreat_cms/static/src/js/pages/machine-translation-overlay.ts
```diff
@@ -0,0 +1,241 @@
...
+            list.textContent = title as string;
+            list.classList.add("list-disc");
+            list.classList.add("ml-4");
+            if (i <= 4) {
+                document.getElementById("machine-translation-overlay-pages").appendChild(list);
+            } else {
+                document.getElementById("machine-translation-overlay-pages-optional").appendChild(list);
+            }
+        }
+    });
```
timoludwig, line 124: I think building HTML elements dynamically in JS is a bit more complicated than needed - instead, I'd suggest to do the calculation which pages are allowed and which aren't in Python, as well as doing the rendering process in Django templates, and let the ajax view return the rendered HTML (see 
[`render_to_string()`](https://docs.djangoproject.com/en/3.2/topics/templates/#django.template.loader.render_to_string))...
```
return JsonResponse(
    {
        "all_pages": all_pages,
        "pages_to_be_translated": render_to_string(...),
        "pages_invalid": render_to_string(...),
    }
)
```

integreat_cms/cms/views/pages/page_actions.py
```diff
@@ -896,3 +896,54 @@ def refresh_date(
...
+                request.region.default_language.slug
+            )
+            == "UP_TO_DATE",
+            pages,
+        )
+    )
+    all_pages = {}
+    for language in request.region.active_languages:
+        default_language = request.region.default_language
+        if language == default_language:
```
timoludwig, line 936: Why this check? We also want to enable machine translations where the source language is not the default language.

integreat_cms/static/src/js/pages/machine-translation-overlay.ts
```diff
@@ -0,0 +1,241 @@
...
+    const usageOfBudget = document.getElementById("machine-translation-overlay-budget-usage-result")
+        .innerText as any as number;
+    const remainingBudget = currentBudget - usageOfBudget;
+    document.getElementById("machine-translation-overlay-budget-remains-result").innerHTML =
+        remainingBudget as unknown as string;
+    toggleMachineTranslationForPagesButton();
+};
+
+const setRequiredBudget = (sum: number) => {
+    document.getElementById("machine-translation-overlay-budget-usage-result").innerHTML = "";
```
timoludwig, line 57:
```suggestion
    document.getElementById("machine-translation-overlay-budget-usage-result").textContent = "";
```
I'd suggest to use `textContent` instead of `innerHTML` whenever you're not really inserting raw HTML. This also applies to all other occurrences of `innerHTML`.

timoludwig (changes requested):
Thanks a lot, really cool! :rocket: 

I know this wasn't ready for review yet, but I had a few preliminary comments nonetheless...

Also, I noticed two general things:
- The HIX value is also enforced when HIX is disabled in a region (or globally)... this does not make sense in my opinion, since we have different production systems where HIX might be disabled, but MTs should be available. Also, I think the budget warning could also be interesting for source languages where HIX is not enabled or events & pois which also don't have HIX enabled...
- I'm not completely sure why, but somehow the warning "None of the selected pages fullfills the required HIX score of 15." was shown for me, even when there exist pages with a sufficient HIX score... 


> Let's say you have 1 page that can be translated and one that can't be, you'll get two error messages for the page that can't be translated (one inside the popup via JS/TS and the second one after confirming via Dj/Py).

You could circumvent this by removing all invalid pages from the selected pages before submitting the form.

> I noticed, that currently I can re-translate a page via MT although the translation status is "Up to date" or "Machine translated and up to date". In order to be more user friendly and prevent unnecessary usage of MT budget, I would suggest adding a check if a machine translation is actually needed or not. First I think it's up for discussion if we really want this or not, and secondly I think that would also be out of scope for this PR and rather something for another issue

Yes good point, I suggest to open an issue for this.

integreat_cms/static/src/js/pages/machine-translation-overlay.ts
timoludwig: On second thought, I think this is fine. I mean the JS implementation works, so probably it would be a waste of time to implement the same thing in Python. Just keep in mind this option for next time 😁

integreat_cms/static/src/js/pages/machine-translation-overlay.ts
```diff
@@ -0,0 +1,241 @@
+/* eslint no-magic-numbers: ["error", { "ignore": [0, 1, 3, 4, 5, 15] }] */
```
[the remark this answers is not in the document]
JoeyStk: Oh, this is a  good point. Thank you! 
Would you prefer for me to get the variable from the settings (for example via a hidden HTML tag) or just use 15 as a constant for HIX?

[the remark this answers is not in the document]
timoludwig: Yes, retrieving it from the settings and passing it to the template (e.g. as data-attribute) makes the most sense I think.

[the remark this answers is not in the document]
JoeyStk: This should be applied now

integreat_cms/static/src/js/pages/machine-translation-overlay.ts
JoeyStk: Ok, thank you. I'll write this down in my notes

integreat_cms/static/src/js/pages/machine-translation-overlay.ts
JoeyStk: I tried to apply this as often as I thought it made sense. I keep innerHTML on two occasions on which it serves as a cleaner to get rid off the HTML child elements. I hope that decision was correct

integreat_cms/cms/views/pages/page_actions.py
JoeyStk: I changed this a little bit now, can you please have a look if I went into the right direction?

JoeyStk:
Finally I took the time to work on this some more, and came up with the following changes. Technically speaking it'S not really refactoring, but I think it can be considered as cleaning up the code a bit:

- I split the setAmount() into two functions/methods, calculateAmount() for doing the logic and setAmount() for actually setting it to the HTML element
- As descibed above I tried to use textContent on as many occasions as I think made sense. There are still two where I use innerHTML since the HTML element contains children which should be cleaned by innerHTML.
- I created more constants to avoid the abundance of document.getElement I had before
- I tried to improve the names of variables and functions in order to make them more self-describing
- I added a comment to every function about what it is doing
- I removed hixThreshold as a new TS constant and instead access the variable we have in our setting
- I changed check if enough budget is remaining from larger than 0 to equal or larger than 0
- I reduced  the amount of magic numbers by setting constants and using them a bit more wisely. The three remaining magic numbers are either used as indexes or as increments. I thought it didn't make sense to also declare them as constants
- I repaired the budget variable (I screwed this up while rebasing my local setup)
- I repaired the "Show all pages" trigger, since it had a logical error
- I changed the Cancel button to be red, since this was a request of @osmers 
- I also added the number of selected pages to the overview, since also this was a request of @osmers 

There are still a few things I didn't do in this go
- I didn't change the names/ids of HTML elements, since I didn't want to screw anything up and I didn't have great alternatives in mind anyway. So please feel free to leave suggestions here
- @osmers asked if we/I also could add the option for the pages that don't have a sufficient hix score to open the editor of the according page by clicking on it's page title. Since I'm reaching the limits of my brain, I would suggest to open a new issue for this and add this feature once this PR has been merged

MizukiTemma (changes requested):
Thank you for PR 💪 

Sadly this problem still exists, even when the HIX is activated and the score is good enough...
>I'm not completely sure why, but somehow the warning "None of the selected pages fullfills the required HIX score of 15." was shown for me, even when there exist pages with a sufficient HIX score...

JoeyStk:
> Sadly this problem still exists, even when the HIX is activated and the score is good enough...
> 
> > I'm not completely sure why, but somehow the warning "None of the selected pages fullfills the required HIX score of 15." was shown for me, even when there exist pages with a sufficient HIX score...

Thank you for your reply. Hmm, strange. I can't reproduce this. I get the correct selection of pages, when I open the popup. Can you please give me some more details?
grafik

JoeyStk:
> Thanks a lot, really cool! rocket
> 
> I know this wasn't ready for review yet, but I had a few preliminary comments nonetheless...
> 
> Also, I noticed two general things:
> 
>     * The HIX value is also enforced when HIX is disabled in a region (or globally)... this does not make sense in my opinion, since we have different production systems where HIX might be disabled, but MTs should be available. Also, I think the budget warning could also be interesting for source languages where HIX is not enabled or events & pois which also don't have HIX enabled...
> 
>     * I'm not completely sure why, but somehow the warning "None of the selected pages fullfills the required HIX score of 15." was shown for me, even when there exist pages with a sufficient HIX score...
> 
> 
> > Let's say you have 1 page that can be translated and one that can't be, you'll get two error messages for the page that can't be translated (one inside the popup via JS/TS and the second one after confirming via Dj/Py).
> 
> You could circumvent this by removing all invalid pages from the selected pages before submitting the form.
> 
> > I noticed, that currently I can re-translate a page via MT although the translation status is "Up to date" or "Machine translated and up to date". In order to be more user friendly and prevent unnecessary usage of MT budget, I would suggest adding a check if a machine translation is actually needed or not. First I think it's up for discussion if we really want this or not, and secondly I think that would also be out of scope for this PR and rather something for another issue
> 
> Yes good point, I suggest to open an issue for this.

I noticed, that I missed that bit of feedback. I will have to take a closer look at it next week

JoeyStk:
> Yes good point, I suggest to open an issue for this.
[Done!] (https://github.com/digitalfabrik/integreat-cms/issues/2247)

JoeyStk:
Thanks for the feedback. I'll look into it! :smiley:

MizukiTemma:
@JoeyStk 
Sorry for direct push, the changes were too large to suggest as comment.
The overlay should show now translatable and untranslatable pages correctly. HIX check is now working with page IDs not titles.

The remaining ToDo are
1) Bipass HIX check if HIX is disabled in the region or globally.
2) Exclude pages that did not pass the HIX check from translation request.

JoeyStk:
@MizukiTemma 
Thank you again for your push. I think I found something for Todo#1 and I fixed an issue with the javascript toggle ("Show more" / "Show less"). But I think there is still the issue, that we have to get only the selected pages from the body via post. However I wasn't able to find a solution for that, because as far as I understand the issue the post body is empty as soon as the function post_hix_and_word_number_per_page is called. Do you maybe have an idea for that?

MizukiTemma:
@JoeyStk Thank you for updates. 
IDs of the selected pages are now passed to the ajax function as boddy data appended in the fetch and the process works only with the selected pages.
HIX bipass now checks not only the global deactivation but also the cases where HIX is globally activated but deactivated in the region too.

integreat_cms/cms/templates/pages/_page_machine_translation_overlay.html
```diff
@@ -0,0 +1,101 @@
...
+                    </table>
+                </div>
+                <div class="mt-4 hidden" id="machine-translation-overlay-warning">
+                    <h3>
+                        <b>{% translate "Pages that can't be translated" %}</b>
+                    </h3>
+                    <div class="my-2 bg-yellow-100 border-l-4 border-yellow-500 text-yellow-600 px-4 py-3 mb-5"
+                         role="alert">
+                        <i class="mr-2" icon-name="alert-triangle"></i>
+                        {% translate "There are currently" %} <span id="machine-translation-overlay-warning-number"></span> {% translate "pages that can't be translated. These pages can't be translated because the text understandibility is too low." %} {% translate "Please improve it to a value of at least 15 to translate these pages. You can identify the pages, that couldn't be translated in the page tree via the translation status." %}
```
timoludwig, line 76: Also, I've explained quite often that it's bad practise to split up translatable sentences into multiple segments. There is no harm in using html markup inside translations, maybe shorten the id name a bit to make it better readable in the translation file.

integreat_cms/cms/templates/pages/_page_machine_translation_overlay.html
```diff
@@ -0,0 +1,101 @@
...
+                    </div>
+                </div>
+                <div class="mt-4">
+                    <h3>
+                        <b>{% translate "Pages to be translated" %}(<span id="machine_translation_overlay_pages_number"></span>/<span id="machine_translation_overlay_pages_total_number"></span>)</b>
+                    </h3>
+                    <div class="mt-2">
+                        <div id="machine-translation-overlay-pages-no-pages-warning"
+                             class="hidden bg-red-100 border-l-4 border-red-500 text-red-500 px-4 py-3 mb-5">
+                            {% translate "None of the selected pages fullfills the required HIX score of " %}  {{ hix_threshold }}.
```
timoludwig, line 38: Also here the variable should be included in the translation text since it's part of the same sentence.

integreat_cms/static/src/js/pages/machine-translation-overlay.ts
```diff
@@ -0,0 +1,267 @@
...
+    if (budget >= 0 && document.getElementById("machine-translation-overlay-pages").children.length > 0) {
+        MachineTranslationButton.disabled = false;
+    } else {
+        MachineTranslationButton.disabled = true;
+    }
+};
+
+const getSelectedItems = () => {
+    // get the HTML tr element of the selected pages
+    const checkboxes = document.querySelectorAll(".bulk-select-item");
```
timoludwig, line 22:
```suggestion
    const checkboxes = document.querySelectorAll(".bulk-select-item:checked");
```
You can directly retrieve the selected checkboxes.

integreat_cms/static/src/js/pages/machine-translation-overlay.ts
```diff
@@ -0,0 +1,267 @@
...
+    return selectedPages;
+};
+
+const getSelectedIDs = () => {
+    // get the HTML tr element of the selected pages
+    const checkboxes = document.querySelectorAll(".bulk-select-item");
+    const selectedPages: Array<number> = [];
+    checkboxes.forEach((checkbox) => {
+        if ((checkbox as HTMLInputElement).checked) {
+            selectedPages.push(parseInt(checkbox.closest("tr").getAttribute("data-drop-id"), 10));
```
timoludwig, line 38: Why go the complicated way via the `tr` and `data-drop-id` when you could just access the value of the checkbox directly?

integreat_cms/static/src/js/pages/machine-translation-overlay.ts
```diff
@@ -0,0 +1,267 @@
...
+        if ((checkbox as HTMLInputElement).checked) {
+            selectedPages.push(checkbox.closest("tr"));
+        }
+    });
+    return selectedPages;
+};
+
+const getSelectedIDs = () => {
+    // get the HTML tr element of the selected pages
+    const checkboxes = document.querySelectorAll(".bulk-select-item");
```
timoludwig, line 34:
```suggestion
    const checkboxes = document.querySelectorAll(".bulk-select-item:checked");
```
You can filter the checked checkboxes directly. Also, why is this queryselector duplicated?

integreat_cms/static/src/js/pages/machine-translation-overlay.ts
```diff
@@ -0,0 +1,267 @@
+/* eslint no-magic-numbers: ["error", { "ignore": [0, 1, 3] }] */
+import { getCsrfToken } from "../utils/csrf-token";
+
+const limitOfPreview = 5;
```
timoludwig, line 4: If the limit is a variable, I would not hardcode the number in the variable name `listOfFirstFivePages`.

integreat_cms/static/src/js/pages/machine-translation-overlay.ts
```diff
@@ -0,0 +1,267 @@
...
+
+const toggleMachineTranslationForPagesButton = () => {
+    // Only activate button if budget is enough
+    const budget = document.getElementById("machine-translation-overlay-budget-remains-result")
+        .innerText as any as number;
```
timoludwig, line 9: Yes another occurrence of this pattern.

integreat_cms/static/src/js/pages/machine-translation-overlay.ts
```diff
@@ -0,0 +1,267 @@
...
+    }
+};
+
+const getSelectedItems = () => {
+    // get the HTML tr element of the selected pages
+    const checkboxes = document.querySelectorAll(".bulk-select-item");
+    const selectedPages: Array<Element> = [];
+    checkboxes.forEach((checkbox) => {
+        if ((checkbox as HTMLInputElement).checked) {
+            selectedPages.push(checkbox.closest("tr"));
```
timoludwig, line 26: I don't see a reason to do this complicated stuff with the additional array of `tr`s when you just need the number of selected pages anyway.

integreat_cms/cms/templates/pages/_page_machine_translation_overlay.html
timoludwig, line 1: I don't see a reason to limit this to pages... why not make this generic for events and pois as well?

JoeyStk:
Please note: This PR is not ready for re-review. I just pushed the recent changes, so @MizukiTemma can help me working on it

integreat_cms/cms/templates/pages/_page_machine_translation_overlay.html
JoeyStk: The service-team agrees with you on that. We'll extend the popup to events and pois as well

integreat_cms/cms/views/pages/page_actions.py
```diff
@@ -896,3 +897,62 @@ def refresh_date(
...
+        id__in=selected_page_ids
+    )
+    hix_threshold = settings.HIX_REQUIRED_FOR_MT
+
+    filtered_pages = {}
+    translatable = {}
+    not_translatable = {}
+
+    for page in selected_pages:
+        translation = page.get_translation(request.region.default_language.slug)
```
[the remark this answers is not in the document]
timoludwig: In my opinion, we should just skip the HIX value check in this case. I would be surprised if the HIX value of the German source text had a huge influence on the quality of machine translations from and to completely different languages.
Also, this would reflect the same constraints that we enforce in the forms (where you can enable machine translations for any language other than German independently from the HIX score).

integreat_cms/cms/templates/pages/_page_machine_translation_overlay.html
MizukiTemma: In my opinion we can generalize the pop up in a separate issue/PR because it is beyond the original scope of issue and this PR is already kind of voluminous. We probably get some feedback and wishes from users for the page pop up and can apply them while generalizing.

timoludwig: Fine for me 👍
Just move the files to a more general location, so they don't have to be moved later on.

JoeyStk:
Thank you for all your suggestions and explanations. The changes should be applied now. We double-checked each other to make sure we don't miss any comment or misinterpret it, so hopefully the implemented changes are what you had in mind. Please let us know if something is not as intended

integreat_cms/static/src/js/machine-translation-overlay.ts
```diff
@@ -0,0 +1,216 @@
...
+    } else {
+        expansionTrigger.classList.add("hidden");
+        expansionTrigger.classList.remove("block");
+    }
+    toggleMachineTranslationForPagesButton();
+
+    if (expandableWarningListElement.children.length !== 0) {
+        warningListElement.classList.add("block");
+        warningListElement.classList.remove("hidden");
+    }
```
MizukiTemma, line 137:
```suggestion
    const expansionTrigger = document.getElementById("machine-translation-overlay-expansion-trigger");
    const noPageWarning = document.getElementById("machine-translation-overlay-pages-no-pages-warning");

    if (previewListOfPages.children.length === 0) {
        noPageWarning.classList.remove("hidden");
    } else {
        noPageWarning.classList.add("hidden");
    }

    if (listOfOptionalPages.children.length === 0) {
        expansionTrigger.classList.add("hidden");
        expansionTrigger.classList.remove("block");
    } else {
        expansionTrigger.classList.remove("hidden");
        expansionTrigger.classList.add("block");
    }

    toggleMachineTranslationForPagesButton();

    if (expandableWarningListElement.children.length !== 0) {
        warningListElement.classList.add("block");
        warningListElement.classList.remove("hidden");
    } else {
        warningListElement.classList.remove("block");
        warningListElement.classList.add("hidden");
    }
```

The warning test "None of the selected pages fullfills the required HIX...." was not toggled correctlly.

integreat_cms/static/src/js/machine-translation-overlay.ts
```diff
@@ -0,0 +1,216 @@
...
+    let numberOfTranslatablePages = resetValue;
+    let numberOfWords = resetValue;
+
+    pages.forEach((page) => {
+        console.log(page.title);
+        const list = document.createElement("li");
+        list.classList.add("list-disc", "ml-4");
+        list.textContent = page.title;
+
+        if (previewListOfPages.children.length <= limitOfPreview) {
```
MizukiTemma, line 78:
```suggestion
        if (previewListOfPages.children.length < limitOfPreview) {
```

integreat_cms/static/src/js/machine-translation-overlay.ts
```diff
@@ -0,0 +1,216 @@
...
+        document.getElementById("btn-close-machine-translation-overlay").addEventListener("click", () => {
+            closeMachineTranslationOverlay(overlay);
+        });
+        // Close window by clicking on backdrop.
+        overlay.addEventListener("click", (e) => {
+            if (e.target === overlay) {
+                closeMachineTranslationOverlay(overlay);
+            }
+        });
+    }
```
MizukiTemma, line 210:
```suggestion
                
                setAmountOfSelectedPages(selectedIDs.length);
            }
        });
        document.getElementById("btn-close-machine-translation-overlay").addEventListener("click", () => {
            closeMachineTranslationOverlay(overlay);
        });
        // Close window by clicking on backdrop.
        overlay.addEventListener("click", (e) => {
            if (e.target === overlay) {
                closeMachineTranslationOverlay(overlay);
            }
        });

        toggleOptionalText(
            document.getElementById("machine-translation-overlay-warning-more"),
            document.getElementById("machine-translation-overlay-warning-optional")
        );
        toggleOptionalText(
            document.getElementById("machine-translation-overlay-expansion-trigger"),
            document.getElementById("machine-translation-overlay-pages-optional")
        );
    }
```

Event listner on "Show pages" were piled up and click toggle didn't work sometimes.

MizukiTemma (changes requested):
Thank you for rebase and update 💪 

I've noticed some incorrect behaviour on toggling of the warning message and click event listner of showing more/less pages.

After those changes the PR is ready for meage in my opinion 🙂

JoeyStk:
Thank you for your suggestions. They are applied now :)

MizukiTemma (approved):
🎉

integreat_cms/cms/templates/_machine_translation_overlay.html
```diff
@@ -0,0 +1,117 @@
...
+                </div>
+                <div class="mt-4 hidden" id="machine-translation-overlay-warning">
+                    <h3>
+                        <b>{% translate "Pages that can't be translated" %}</b>
+                    </h3>
+                    <div class="my-2 bg-yellow-100 border-l-4 border-yellow-500 text-yellow-600 px-4 py-3 mb-5"
+                         role="alert">
+                        <i class="mr-2" icon-name="alert-triangle"></i>
+                        {% blocktranslate trimmed %}
+                            There are currently <span id="not-translateable"></span> page(s) that can't be translated.
```
timoludwig, line 84:
```suggestion
                            There are currently <span id="not-translatable"></span> page(s) that can't be translated.
```
Small typo

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -3951,6 +3951,125 @@ msgstr "Standard-Sprache wird dupliziert"
...
+#: cms/templates/_machine_translation_overlay.html
+msgid "Pages that can't be translated"
+msgstr "Seiten, die nicht übersetzt werden können"
+
+#: cms/templates/_machine_translation_overlay.html
+msgid ""
+"There are currently <span id=\"not-translateable\"></span> page(s) that "
+"can't be translated."
+msgstr ""
+"Es gibt aktuell <span id=\"not-translateable\"></span> Seiten, die nicht "
```
timoludwig, line 4020:
```suggestion
"There are currently <span id=\"not-translatable\"></span> page(s) that "
"can't be translated."
msgstr ""
"Es gibt aktuell <span id=\"not-translatable\"></span> Seiten, die nicht "
```

integreat_cms/static/src/js/machine-translation-overlay.ts
```diff
@@ -0,0 +1,225 @@
...
+
+    pages.forEach((page) => {
+        const list = document.createElement("li");
+        list.classList.add("list-disc", "ml-4");
+        list.textContent = page.title;
+        expandableWarningListElement.appendChild(list);
+        numberOfNotTranslatablePages += 1;
+    });
+
+    document.getElementById("not-translateable").textContent = numberOfNotTranslatablePages.toString();
```
timoludwig, line 106:
```suggestion
    document.getElementById("not-translatable").textContent = numberOfNotTranslatablePages.toString();
```

timoludwig (changes requested):
Thanks a lot, yes this looks a lot better now! :+1:

integreat_cms/cms/templates/_machine_translation_overlay.html
```diff
@@ -0,0 +1,117 @@
...
+                        {% endblocktranslate %}
+                        {% if source_language.slug == "de" %}
```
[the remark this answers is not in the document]
JoeyStk: This should be applied now

JoeyStk:
Thank you a lot for your feedback. The changes should be applied now, the PR should be rebased to only one commit and no merge conflict should be open anymore. With that I think this PR is ready for re-review :)

timoludwig (approved):
Thanks a lot! :rocket:

Issue #2230, opened 2023-04-17
"HIX value too low for MT" error message shown even when automatic MT is unchecked


charludo:
### Describe the Bug

When attempting to automatically translate a page with insufficient HIX score, an error message is shown to the user. However, this message is now also shown when the user did not attempt to automatically translate

### Steps to Reproduce

1. Go to any German page form view
2. Uncheck automatic MT checkbox
3. Chenge the text to something with a low HIX value (optional, makes no difference: refresh the HIX score)
4. click Publish
5. receive error message

### Expected Behavior

The error is only shown if the user actually attempted to use MT

### Actual Behavior

The error is always shown

PR #2231, opened 2023-04-17
Fix wrongfully shown 'HIX too low' error message

charludo:
### Short description

The "HIX score too low for MT" error message was shown whenever the HIX score was too low, even when MT had not been attempted by the user.

### Proposed changes

- if automatic MT was unchecked, all translation target languages were removed from the forms `cleaned_data`, however to check if MT was possible, the querysets assigned to the fields themselves were checked for emptiness -> fixed

### Side effects

- none 

### Resolved issues

Fixes: #2230

### DIFF
--- a/integreat_cms/cms/forms/machine_translation_form.py
+++ b/integreat_cms/cms/forms/machine_translation_form.py
@@ -82,9 +82,10 @@ class MachineTranslationForm(CustomContentModelForm):
         ].queryset = self.request.region.language_tree_nodes.filter(id__in=to_update)
         self.initial["mt_translations_to_update"] = to_update
 
-    def mt_is_enabled(self):
+    def mt_form_is_enabled(self):
         """
-        Validate form fields which depend on each other, see :meth:`django.forms.Form.clean`
+        Helper method to decide if this form should be shown, or if it should be hidden for
+        the current language due to a lack of MT-compatible child language nodes
 
         :return: Whether this form is enabled
         :rtype: bool
@@ -123,14 +124,10 @@ class MachineTranslationForm(CustomContentModelForm):
         """
         self.instance = super().save(commit, foreign_form_changed)
 
-        if (
-            commit
-            and self.mt_is_enabled
-            and check_hix_score(self.request, self.instance)
-        ):
-            language_nodes = self.cleaned_data["mt_translations_to_create"].union(
-                self.cleaned_data["mt_translations_to_update"]
-            )
+        language_nodes = self.cleaned_data["mt_translations_to_create"].union(
+            self.cleaned_data["mt_translations_to_update"]
+        )
+        if commit and language_nodes and check_hix_score(self.request, self.instance):
             for language_node in language_nodes:
                 logger.debug(
                     "Machine translation via %r into %r for: %r",
--- a/integreat_cms/cms/templates/events/event_form_sidebar/minor_edit_box.html
+++ b/integreat_cms/cms/templates/events/event_form_sidebar/minor_edit_box.html
@@ -18,7 +18,7 @@
         </label>
         <div class="help-text">{% minor_edit_help_text request.region language event_translation_form %}</div>
     </div>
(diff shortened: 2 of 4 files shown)


integreat_cms/cms/forms/machine_translation_form.py
```diff
@@ -125,7 +129,7 @@ def save(self, commit=True, foreign_form_changed=False):
```
timoludwig, line 129:
```suggestion

        language_nodes = self.cleaned_data["mt_translations_to_create"].union(
            self.cleaned_data["mt_translations_to_update"]
        )
```
What do you think about pulling out the language nodes from the inner block and only check the hix score if there are language nodes to begin with?

integreat_cms/cms/forms/machine_translation_form.py
```diff
@@ -125,7 +129,7 @@ def save(self, commit=True, foreign_form_changed=False):
 
         if (
             commit
-            and self.mt_is_enabled
+            and self.mt_is_enabled()
```
timoludwig, line 132:
```suggestion
            and self.mt_is_enabled()
            and language_nodes
```

timoludwig (changes requested):
Sorry for messing this up and thanks for the quick fix :innocent:

integreat_cms/cms/forms/machine_translation_form.py
timoludwig: I wanted to use the `mt_is_enabled()` method to check whether the MT form is shown on the page or not, see e.g. https://github.com/digitalfabrik/integreat-cms/blob/df2863ec4d47ce7f37aab2d8fdd4564514047431/integreat_cms/cms/templates/pages/page_form_sidebar/minor_edit_box.html#L21

The problem if we also use this method whether something was selected, is, that it will e.g. hide the form if MT was selected and the form validation of the page form or page translation form fails, the form is directly rendered again (with the error message indicating which form field caused the error), but the MT form will just be hidden in this case:

Screenshot 2023-04-17 at 11-08-12 Integreat Editorial System

So I think it would make sense to separate these two steps (showing the form and having something selected)... Maybe to reduce the potential for misunderstandings, we could adapt the docstring of the `mt_is_enabled()` method?

timoludwig: Or we could rename `mt_is_enabled()` to `show_mt_form` or similar? I guess I just wanted to make it sound similar to the inbuilt `is_valid()`, but if the meaning isn't clear, it's not a good choice.

charludo: Ahh, I see. Yes, I think renaming it would be a good idea, from the naming and logic inside it I was assuming this was just used to check if MT had been enabled (by the user) for this page.

How about `mt_form_is_enabled` as the method name and "Helper method to decide if this form should be shown, or if it should be hidden for the current language due to a lack of MT-compatible child language nodes"  or something?

timoludwig: Yes, sounds good! :+1:

integreat_cms/cms/forms/machine_translation_form.py
```diff
@@ -125,7 +129,7 @@ def save(self, commit=True, foreign_form_changed=False):
...
             and check_hix_score(self.request, self.instance)
         ):
             language_nodes = self.cleaned_data["mt_translations_to_create"].union(
```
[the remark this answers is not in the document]
charludo: Super pendantic, but I guess we can remove the `self.mt_form_is_enabled()` check here? Since we're checking if the `cleaned_data` is non-empty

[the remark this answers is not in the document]
timoludwig: Ah yes, correct :+1:

timoludwig (approved):
Thanks! :muscle:

MizukiTemma (approved):
Thank you, looks good 🎉

Issue #2225, opened 2023-04-13
Show HIX score in page revision view


MizukiTemma:
### Motivation

We store HIX values in page translations. Showing them in the reversion view probably motivates and helps users to see how the score has changes so far and which changes have made it better/worse.

### Proposed Solution

Show hix values in the page revision view, if HIX is activated globally and in the region and a score is saved.

### Alternatives

Leave it as it is.

PR #2233, opened 2023-04-18
Show Hix value in page revision view

MizukiTemma:
### Short description

This PR shows HIX value in page revision view.

### Proposed changes

- Show Hix score in the slide bar box and between Publication status and Author.
- Apply this only when HIX is globally and in the region activated and not deactivated for the page. If no HIX value is stored, leave the space empty (just like empty content).

### Side effects

- I hope none 🙈 

### Resolved issues

Fixes: #2225 

### DIFF
--- a/integreat_cms/cms/templates/pages/page_revisions.html
+++ b/integreat_cms/cms/templates/pages/page_revisions.html
@@ -33,6 +33,10 @@
                 <b>{% translate "Author" %}:</b> <span id="revision-editor"></span>
                 <br />
                 <b>{% translate "Date" %}:</b> <span id="revision-date"></span>
+                {% if page_translations.first.hix_enabled and not page_translations.first.hix_ignore %}
+                    <br />
+                    <b>{% translate "HIX" %}:</b> <span id="revision-hix"></span>
+                {% endif %}
             </output>
             <datalist id="steplist" class="w-full flex font-mono">
                 {% for page_translation in page_translations reversed %}
@@ -53,27 +57,38 @@
                  id="revision-{{ page_translation.version }}"
                  data-date="{{ page_translation.last_updated }}"
                  data-editor="{% firstof page_translation.creator deleted_user_text %}"
-                 data-status="{{ page_translation.status }}">
-                <label class="inline-block">{% translate "Publication status" %}:</label>
-                {{ page_translation.get_status_display }}
-                {% if page_translation == api_revision %}
-                    ({% translate "This is the version shown in the apps." %})
-                {% elif forloop.first %}
-                    {% if api_revision %}
-                        {% with api_revision.version as version %}
-                            ({% blocktrans %}This version is <b>not</b> displayed in the apps - but version {{ version }} instead{% endblocktrans %})
-                        {% endwith %}
-                    {% else %}
-                        ({% translate "This translation is <b>not</b> shown in the apps" %})
-                    {% endif %}
-                {% endif %}
-                {% if page_translation.minor_edit %}
-                    - <i>{% translate "Minor edit" %}</i>
-                {% endif %}
-                <span class="float-right">
-                    <label class="inline-block">{% translate "Author" %}:</label>
-                    {% firstof page_translation.creator deleted_user_text %}
-                </span>
+                 data-status="{{ page_translation.status }}"
+                 data-hix="{{ page_translation.hix_score|floatformat:2 }}">
+                <div class="flex justify-between">
+                    <span>
+                        <label class="inline-block">{% translate "Publication status" %}:</label>
+                        {{ page_translation.get_status_display }}
+                        {% if page_translation == api_revision %}
+                            ({% translate "This is the version shown in the apps." %})
+                        {% elif forloop.first %}
+                            {% if api_revision %}
+                                {% with api_revision.version as version %}
+                                    ({% blocktrans %}This version is <b>not</b> displayed in the apps - but version {{ version }} instead{% endblocktrans %})
+                                {% endwith %}
+                            {% else %}
+                                ({% translate "This translation is <b>not</b> shown in the apps" %})
+                            {% endif %}
+                        {% endif %}
+                        {% if page_translation.minor_edit %}
+                            - <i>{% translate "Minor edit" %}</i>
+                        {% endif %}
+                    </span>
+                    <span>
+                        {% if page_translation.hix_enabled and not page_translation.hix_ignore and page_translation.hix_score %}
+                            <span class="pr-4">
+                                <label class="inline-block">{% translate "HIX score" %}:</label>
+                                {{ page_translation.hix_score|floatformat:2 }}
+                            </span>
+                        {% endif %}
+                        <label class="inline-block">{% translate "Author" %}:</label>
+                        {% firstof page_translation.creator deleted_user_text %}
+                    </span>
+                </div>
                 <div class="revision-plain hidden">
                     <label>{% translate "Permalink" %}:</label>
                     {{ page_translation.get_absolute_url }}
--- a/integreat_cms/core/signals/hix_signals.py
+++ b/integreat_cms/core/signals/hix_signals.py
@@ -20,33 +20,31 @@ def page_translation_save_handler(instance, **kwargs):
     :param \**kwargs: The supplied keyword arguments
     :type \**kwargs: dict
     """
+
     if kwargs.get("raw"):
         return
 
-    content_unchanged = (
-        instance.latest_version
-        and instance.latest_version.content == instance.content
-        and instance.latest_version.hix_score is not None
-    )
-    if (
-        instance.hix_score
-        or instance.hix_ignore
-        or not instance.hix_enabled
-        or not instance.content.strip()
-        or content_unchanged
-    ):
+    if instance.hix_ignore or not instance.hix_enabled or not instance.content.strip():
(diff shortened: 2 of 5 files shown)


JoeyStk (changes requested):
Thank you!
In general this seems to work. 
However when I restored one version, it didn't re-calculate or restore the HIX score and therefore kept it blank (s.screenshot)
grafik

integreat_cms/cms/templates/pages/page_revisions.html
```diff
@@ -53,27 +57,39 @@ <h1 class="heading">
...
-                {% elif forloop.first %}
-                    {% if api_revision %}
-                        {% with api_revision.version as version %}
-                            ({% blocktrans %}This version is <b>not</b> displayed in the apps - but version {{ version }} instead{% endblocktrans %})
-                        {% endwith %}
-                    {% else %}
-                        ({% translate "This translation is <b>not</b> shown in the apps" %})
+                 data-status="{{ page_translation.status }}"
+                 data-hix="{{ page_translation.hix_score|floatformat:2 }}">
+                >
```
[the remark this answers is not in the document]
MizukiTemma: oops 😅 Thank you!

charludo (changes requested):
Thanks, this works really well! My only gripe is that maybe we should indicate when the HIX score is ignored - but I think since this is only defined for pages, not page translations, we are not saving the history of that setting...? So probably the effort is not justified 😅

MizukiTemma:
> Thanks, this works really well! My only gripe is that maybe we should indicate when the HIX score is ignored - but I think since this is only defined for pages, not page translations, we are not saving the history of that setting...? So probably the effort is not justified 😅

It would be ideal to show in which versions HIX was ignored, though I think it is too much work because as you mentioned `hix_ignore` is not in translations 😿 

Placeholder after "HIX" is now implemented, both in the slide bar box and between publication status and author 🙂

charludo (approved):
Thanks, LGTM! 🎉

integreat_cms/cms/templates/pages/page_revisions.html
```diff
@@ -53,27 +57,42 @@ <h1 class="heading">
...
                     {% endif %}
-                {% endif %}
-                {% if page_translation.minor_edit %}
-                    - <i>{% translate "Minor edit" %}</i>
-                {% endif %}
-                <span class="float-right">
-                    <label class="inline-block">{% translate "Author" %}:</label>
-                    {% firstof page_translation.creator deleted_user_text %}
-                </span>
+                    <span>
```
timoludwig, line 91:
```suggestion
                    <span>
                        {% if page_translation.hix_enabled and not page_translation.hix_ignore and page_translation.hix_score %}
                            <span class="pr-4">
                                <label class="inline-block">{% translate "HIX score" %}:</label>
                                {{ page_translation.hix_score|floatformat:2 }}
                            </span>
                        {% endif %}
```
I think it would look better if the HIX score would also be aligned to the right and not floating in the middle.
Also, we can make use of the translation properties here:
https://github.com/digitalfabrik/integreat-cms/blob/e654925ba72fe2930e8e36acd8eabac82af1c6c7/integreat_cms/cms/models/abstract_content_translation.py#L510-L533

This also makes sure that we don't show the HIX value on languages that are not enabled for HIX.

integreat_cms/cms/templates/pages/page_revisions.html
```diff
@@ -53,27 +57,42 @@ <h1 class="heading">
...
+                    </span>
+                    {% if TEXTLAB_API_ENABLED and request.region.hix_enabled and not page_translations.first.page.hix_ignore %}
+                        <span>
+                            <label class="inline-block">{% translate "HIX score" %}:</label>
+                            {% if page_translation.hix_score %}
+                                {{ page_translation.hix_score|floatformat:2 }}
+                            {% else %}
+                                -
+                            {% endif %}
+                        </span>
```
timoludwig, line 89:
```suggestion
```

integreat_cms/cms/templates/pages/page_revisions.html
```diff
@@ -53,27 +57,42 @@ <h1 class="heading">
...
                     {% endif %}
```
timoludwig, line 90:
```suggestion
```

MizukiTemma:
> Thank you! In general this seems to work. However when I restored one version, it didn't re-calculate or restore the HIX score and therefore kept it blank

Ohw, I didn't notice it 😅 Thank you for finding!

I modified `page_translation_save_handler` so it returns the existing HIX value if such extsts instead of skipping the process. HIX value must be now restored too. I hope this change doesn't bring any side effect to HIX process.

integreat_cms/core/signals/hix_signals.py
```diff
@@ -28,9 +28,10 @@ def page_translation_save_handler(instance, **kwargs):
         and instance.latest_version.content == instance.content
         and instance.latest_version.hix_score is not None
     )
+    if instance.hix_score:
+        return instance.hix_score
```
timoludwig: > I modified page_translation_save_handler so it returns the existing HIX value if such extsts instead of skipping the process. HIX value must be now restored too. I hope this change doesn't bring any side effect to HIX process.

Yes this makes sense, but it would be better to refactor the whole function in my opinion, since the whole if cases are nested more than they need to be.

What do you think about e.g. the following:

```
    if identical_version := instance.all_versions.filter(
        content=instance.content, hix_score__isnull=False
    ).first():
        logger.debug(
            "Content of %r is identical to %r, copying HIX score %r",
            instance,
            identical_version,
            identical_version.hix_score,
        )
        instance.hix_score = identical_version.hix_score
        return

    if instance.hix_ignore or not instance.hix_enabled or not instance.content.strip():
        logger.debug(
            "HIX calculation pre save signal skipped for %r (ignored=%s, enabled=%s, empty=%s)",
            instance,
            instance.hix_ignore,
            instance.hix_enabled,
            not bool(instance.content.strip()),
        )
        instance.hix_score = None
        return
```

integreat_cms/release_notes/current/unreleased/2225.yml
```diff
@@ -0,0 +1,2 @@
+en: Show HIX score in page revision view
+de: HIX-Punktzahl in der Seitenrevisionsansicht anzeigen
```
timoludwig, line 2:
```suggestion
de: Zeige HIX-Wert in der Seiten-Versionsansicht an
```

timoludwig (changes requested):

integreat_cms/core/signals/hix_signals.py
MizukiTemma: Looks cool 😄 Applied!

integreat_cms/core/signals/hix_signals.py
```diff
@@ -20,33 +20,31 @@ def page_translation_save_handler(instance, **kwargs):
     :param \**kwargs: The supplied keyword arguments
     :type \**kwargs: dict
     """
-    if kwargs.get("raw"):
+
+    if kwargs.get("raw") or instance.hix_score:
```
timoludwig, line 24:
```suggestion
    if kwargs.get("raw"):
```
This looks like a good optimization at first glance, but will often times cause incorrect HIX values.
Whenever we create new versions, we inherit all fields (including the HIX value) from the old version, and only overwrite the content manually. So we have to re-calculate the HIX score here, and can only keep it if the content is actually identical (thus the check for `identical_version` further below)...

timoludwig (changes requested):

integreat_cms/cms/templates/pages/page_revisions.html
```diff
@@ -33,6 +33,10 @@ <h1 class="heading">
                 <b>{% translate "Author" %}:</b> <span id="revision-editor"></span>
                 <br />
                 <b>{% translate "Date" %}:</b> <span id="revision-date"></span>
+                {% if page_translation.first.hix_enabled and not page_translation.first.hix_ignore %}
```
timoludwig, line 36:
```suggestion
                {% if page_translations.first.hix_enabled and not page_translations.first.hix_ignore %}
```
Sorry, I made a small typo in my suggestion

timoludwig (changes requested):
Awesome, thanks! 
Apart from a small mistake on my side, this looks good! :tada:

Could you squash your commits?

timoludwig (approved):
Amazing! :rocket: 

Just a side note: If you want to attribute the whole commit to @charludo and @JoeyStk as well, you could remove the new lines between the co-authored-messages, so change 

```
Show Hix value in page revision view

    Co-authored-by: JoeyStk <72705147+JoeyStk@users.noreply.github.com>

    Co-authored-by: Charlotte <47758554+charludo@users.noreply.github.com>

    Co-authored-by: Timo Ludwig <timo.ludwig@tuerantuer.org>
```
to:
```
Show Hix value in page revision view

    Co-authored-by: JoeyStk <72705147+JoeyStk@users.noreply.github.com>
    Co-authored-by: Charlotte <47758554+charludo@users.noreply.github.com>
    Co-authored-by: Timo Ludwig <timo.ludwig@tuerantuer.org>
```

which should then be recognized correctly by GitHub :wink:

Screenshot 2023-04-19 at 18-45-48 Show Hix value in page revision view by MizukiTemma · Pull Request #2233 · digitalfabrik_integreat-cms

Issue #2222, opened 2023-04-11
Broken HIX loading animation when DeepL is enabled


JoeyStk:
### Describe the Bug
When I open a page and the HIX widget is not loaded yet, it shows the message, that supposedly there was no content on my pages, which is not true. Usually this issue resolves itself as soon as the HIX widget is loaded, however when ever the HIX service is not available, this message stays for a while (ca. 30 seconds, before showing the error message, that HIX isn't available). This is not a huge probem, but it might cause some confusion.

### Steps to Reproduce

0. Enable DeepL by setting a key in the config file and enable HIX in the region settings
1. Go to pages
2. Select a page
3. See error in HIX box (error resolves itself as soon as the widget is loaded)

### Expected Behavior
Show an error message like "HIX widget is loading..." or something similar

### Actual Behavior

Shows untrue error message "No content on page"

### Additional Information
![Bildschirmfoto von 2023-04-11 17-08-10](https://user-images.githubusercontent.com/72705147/231209824-a72f4699-a368-4821-83de-7795e677c0d4.png)

timoludwig:
I think this only happens when DeepL is enabled, but I did not yet have a closer look to why that is...

PR #2235, opened 2023-04-19
Show less confusing message while HIX widget is loading

charludo:
### Short description

Show "HIX widget is loading..."  instead of "There is no page content." while the HIX widget is loading.

### Proposed changes

The issue was that the 1000ms timeout used to ensure tinyMCE was fully initialized before attempting to call `getContent()` was insufficient. Instead of increasing the timeout, I did the following:
- remove the timeout completely
- tinyMCE dispatches a new `tinyMCEInitialized` signal after completing initialization
- `initHixValue` is no longer called on page load, but only after that signal has been emitted
- some more refactoring and simplification of the `initHixValue` function
- unrelated: fixed an `uncaught TypeError` (it was such a small fix I did not think a separate issue was necessary)

### Side effects

- none that I could find. AFAIK the behaviour of the widget is as it was before, minus the confusing "no page content" message.

### Resolved issues

Fixes: #2222

### DIFF
--- a/integreat_cms/cms/templates/hix_widget.html
+++ b/integreat_cms/cms/templates/hix_widget.html
@@ -11,8 +11,8 @@
 {% block collapsible_box_content %}
     <div class="relative">
         <div id="hix-loading"
-             class="hidden absolute w-full h-full bg-white bg-opacity-75">
-            <div class="relative top-1/3 m-auto w-1/3">
+             class="{% if page_form.instance.hix_ignore %} hidden{% endif %} absolute w-full h-full bg-white bg-opacity-75">
+            <div class="relative top-1/3 text-center">
                 <i icon-name="loader" class="animate-spin"></i> {% translate "Loading..." %}
             </div>
         </div>
--- a/integreat_cms/static/src/js/analytics/hix-widget.ts
+++ b/integreat_cms/static/src/js/analytics/hix-widget.ts
@@ -62,11 +62,23 @@ const setHixLabelState = (state: string) => {
     } else if (state === "updated") {
         document.getElementById("hix-container").classList.remove("hidden");
     }
+
+    // Hide the button if there is no content or the value is up-to-date
+    const updateButton = document.getElementById("btn-update-hix-value");
+    if (state === "updated" || state === "no-content") {
+        updateButton.classList.add("hidden");
+    } else {
+        updateButton.classList.remove("hidden");
+    }
+
+    // Hide the loading spinner
+    document.getElementById("hix-loading")?.classList.add("hidden");
 };
 
-const getHixValue = async (url: string) => {
+const getHixValue = async () => {
+    const updateButton = document.getElementById("btn-update-hix-value");
     let result;
-    await fetch(url, {
+    await fetch(updateButton.dataset.url, {
         method: "POST",
         headers: {
             "X-CSRFToken": getCsrfToken(),
@@ -75,11 +87,7 @@ const getHixValue = async (url: string) => {
             text: getContent(),
         }),
     })
-        .then((response) => {
-            // Hide loading spinner
-            document.getElementById("hix-loading")?.classList.add("hidden");
-            return response.json();
-        })
+        .then((response) => response.json())
         .then((json) => {
             const labelState = json.error ? "error" : "updated";
             setHixLabelState(labelState);
@@ -93,8 +101,6 @@ window.addEventListener("load", async () => {
     if (!document.getElementById("hix-chart")) {
         return;
     }
-
-    document.getElementById("hix-loading")?.classList.remove("hidden");
 
     // Define dummy data for empty default chart.
     const dummyData = {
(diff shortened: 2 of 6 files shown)


integreat_cms/cms/templates/hix_widget.html
```diff
@@ -49,10 +49,15 @@
                     <i icon-name="info"></i>
                     {% translate "There is no page content." %}
                 </div>
+                <div data-hix-state="loading" class="bg-blue-400 rounded p-2">
+                    <i icon-name="info"></i>
+                    {% translate "HIX widget is loading..." %}
+                </div>
```
timoludwig, line 55: There is already a loading div which is working when DeepL is disabled:
https://github.com/digitalfabrik/integreat-cms/blob/c27133f3ac9ad5fe41941cb0b7ca3ab73e1f7d39/integreat_cms/cms/templates/hix_widget.html#L13-L18

Do you think you could merge them somehow, so we don't have two different loading elements?

charludo: We could just remove the new one completely, if that's OK with you? This just ensures that a message is always shown.

timoludwig: Ah sorry, now I get it, the new one is not the overlay, but just the message that is shown behind the overlay? Ok, yeah I guess it's fine removing it, since it is covered by the loading overlay anyway?

timoludwig: Ok, but somehow this change caused the overlay to not be shown anymore? :sweat_smile: 
Probably it makes sense to remove the `hidden` class here:
https://github.com/digitalfabrik/integreat-cms/blob/9536578edc22bc3df607622345b03951ac65fbf0/integreat_cms/cms/templates/hix_widget.html#L14

And only hide the loading overlay after the value really has been initialized?

charludo: Whoops, sorry. Yes.

integreat_cms/cms/templates/hix_widget.html
```diff
@@ -11,7 +11,7 @@
 {% block collapsible_box_content %}
     <div class="relative">
         <div id="hix-loading"
-             class="hidden absolute w-full h-full bg-white bg-opacity-75">
+             class="absolute w-full h-full bg-white bg-opacity-75">
             <div class="relative top-1/3 m-auto w-1/3">
```
timoludwig, line 15:
```suggestion
            <div class="relative top-1/3 text-center">
```
Not your fault, but I noticed the text is not really centered (only when the loading text has exactly the width of 1/3rd of the widget by chance)

integreat_cms/cms/templates/hix_widget.html
```diff
@@ -11,7 +11,7 @@
 {% block collapsible_box_content %}
     <div class="relative">
         <div id="hix-loading"
-             class="hidden absolute w-full h-full bg-white bg-opacity-75">
```
timoludwig, line 14: I guess this means that the following line can be deleted, right?
https://github.com/digitalfabrik/integreat-cms/blob/9b4fff616594ca995a9d24b79d74a3c0caf6ae3d/integreat_cms/static/src/js/analytics/hix-widget.ts#L98

integreat_cms/static/src/js/analytics/hix-widget.ts
```diff
@@ -167,14 +168,33 @@ window.addEventListener("load", async () => {
...
+        updateChart(chart, initialHixValue);
+        toggleMTAvailability(initialHixValue);
+    };
+
     // Set listener, that checks, if tinyMCE content has changed to update the
     // HIX value status
     document.querySelectorAll("[data-content-changed]").forEach((element) => {
+        // make sure initHixValue is called only after tinyMCE is initialized
+        element.addEventListener("tinyMCEInitialized", () => {
+            initHixValue();
```
timoludwig, line 195:
```suggestion
        element.addEventListener("tinyMCEInitialized", async () => {
            await initHixValue();
```
I believe this has to be async, otherwise the loading overlay is removed instantly, without waiting for the result.

timoludwig (changes requested):

integreat_cms/static/src/js/analytics/hix-widget.ts
```diff
@@ -64,9 +64,10 @@ const setHixLabelState = (state: string) => {
     }
```
timoludwig, line 64: Not sure whether that's a bad idea, but what do you think about hiding the button when there is nothing to by adding the block
```
    // Hide the button if there is no content or the value is up-to-date
    const updateButton = document.getElementById("btn-update-hix-value");
    if (state === "updated" || state === "no-content") {
        updateButton.classList.add("hidden");
    } else {
        updateButton.classList.remove("hidden");
    }
```
to the `setHixLabelState()` function? (And remove the `updateButton.disabled = ...` statements?)

timoludwig (commented):
Very cool improvement! :+1:

What do you think about adding a small release note for this? (I guess not obligatory since it's a minor change, but still visually noticeable)

integreat_cms/static/src/js/analytics/hix-widget.ts
charludo: Sure, why not! 😄

MizukiTemma (approved):
I'm for the idea of release note entry (though not necessary)🙂

timoludwig (approved):
Thank you! :tada:

timoludwig:
@charludo I think that's a really cool improvement which should be released soon, so I'm going to merge this now :wink:

charludo:
> @charludo I think that's a really cool improvement which should be released soon, so I'm going to merge this now :wink:

Yeah, sorry, go ahead. I remember waiting for the checks to pass... Looks like I forgot to check back 😄

Issue #2237, opened 2023-04-19
Link text sometimes differs in link replacement form


timoludwig:
### Describe the Bug

Since I added the disabled link text field to the link replacement form, there seem to be edge cases where the link text differs in the table row and the form field, which might be confusing. Both links belong to the same URL, but there seems to be no apparent reason why both variants should be shown here. Using the same would be better.

### Steps to Reproduce

1. Go to linkcheck list
2. Replace URL that has more than one link
3. See different link text (sometimes)

### Expected Behavior

The link texts should be identical

### Actual Behavior

They are differing:
![Screenshot 2023-04-19 at 19-48-27 Integreat Editorial System](https://user-images.githubusercontent.com/16021269/233158725-3d417d2f-9e8e-45a7-a815-d9e6f54d2a0e.png)

Issue #2226, opened 2023-04-15
Allow to replace link text in link checker


timobrembeck:
### Motivation

In many cases, links are not only contained in the `href` attribute of the specific link:
```
<a href="https://some-link.tld">Readable link text</a>
```
But also in the link text itself:
```
<a href="https://some-link.tld">https://some-link.tld</a>
```

At the moment, it is only possible to replace the link in the `href` attribute, and not the link text. This is mostly due to the simpler implementation (we can just make use of [`lxml.html.rewrite_links()`](https://lxml.de/lxmlhtml.html) instead of some complex regex) and the fact that the URL is identical for all replaced link objects, but the link texts might differ.

### Proposed Solution

At least in the case where all link texts are identical, I suggest to go the extra mile and allow the users to replace the link text as well. This would be a little bit more complicated, but can probably also be done with lxml (e.g. `iterlinks()`), so hopefully no regex matching.

Optionally, we could also think about allowing to "override" the link text even if it differs in the other sources.
### Alternatives

Leave it up to the content creators to go through all affected pages and replace the link texts manually. However, since only the first link text is shown in the link checker, it's practically impossible to find the other occurrences of the link.
Probably we could have to at least provide a simpler method of finding all occurrences of a specific link.

### User Story

As a region manager or editor, I want to replace not only the links but also the link texts (in case they are URLs) to make sure that the app users see the correct link text for the new link target.

timoludwig:
As mentioned in #2307, making the link text editable via the form field led to several problems.
Thus, I suggest to choose a more black-box feature instead:
- Try to determine whether the link text should be replaced as well
- Automatically replace the link text if it's identical or similar to the URL (e.g. `mailto:`, `tel:` links or shortened URL domains etc)
- Optionally show a warning if the "magic" method detects an inconsistency between URL and link text that it cannot fix automatically

osmers:
Closing for now as stale and no real requests for that from municipalities. Can reopen if reports by municipalities come in.

PR #2238, opened 2023-04-19
Improve UI of link checker

timoludwig:
### Short description

The author chat revealed that users do not understand the new columns "SSL" and "Code".
In my opinion, the difference between a secure and an insecure link as well as between a "404 No Found" and "403 Forbidden" should still be useful to uneducated users, thus I suggest to add a few help text, rather than just hide them from non-expert users...

### Proposed changes

- Fix differing link text in link replacement form
- Improve explanation of link checker columns

### Resolved issues

Fixes: #2237

### DIFF
--- a/integreat_cms/cms/templates/linkcheck/link_list_row.html
+++ b/integreat_cms/cms/templates/linkcheck/link_list_row.html
@@ -14,23 +14,25 @@
                    value="{{ url.id }}"
                    class="bulk-select-item"/>
         </td>
-        <td class="pr-2 whitespace-nowrap max-w-[75px] sm:max-w-[100px] md:max-w-[100px] lg:max-w-[100px] xl:max-w-[150px] 2xl:max-w-[200px] 3xl:max-w-[260px] 4xl:max-w-[400px] overflow-hidden text-ellipsis"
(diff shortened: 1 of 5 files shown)


MizukiTemma (approved):
LGTM 👍

timoludwig:
> * Do you think there is any chance to also add some explanation to the status? (for example: I find " New Connection Error: Failed to establish a new connection: " not very descriptive)

Hmm, the status is just the most readable explanation of the URL check, sometimes we can assemble a translated string from the URL fields, but sometimes there are unexpected errors, and I don't see an easy way around it than just printing out the error that the request library encountered...
After all, I think "failed to establish a connection" pretty much makes it clear what the problem is: The website of this link is either temporarily unavailable or does not exist (anymore)...
Maybe we could start translating some of these exceptions manually, but there are quite a few, so I don't know how much coverage we'll get...

> * And I was a bit confused by the read-only-field in the link text. Is there a reason why this is an input field and not plain text? thinking

I thought this would make it extra clear that this value is not changed by replacing the URL. Otherwise people might still think that the link text is maybe replaced as well?
Also, this is a preparation for #2226...

PR #2242, opened 2023-04-20
Add repair tree CLI command

svenseeberg:
### Short description
Add a shell command that repairs a tree of a given tree node ID.

### Proposed changes
- Loop through all branches of a tree beginning from root node.
- Sort existing tree nodes by sorting them by existing lft values.
- Re-calculate lft and rgt values by counting through the leafs.

### Side effects
- None

### Resolved issues
None


integreat_cms/core/management/commands/repair_tree.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def add_arguments(self, parser):
+        """
+        Define the arguments of this command
+
+        :param parser: The argument parser
+        :type parser: ~django.core.management.base.CommandParser
+        """
+        parser.add_argument(
+            "--page", help="The ID of a page belongig to the broken tree."
```
timoludwig, line 26:
```suggestion
            "page_id", type=int, help="The ID of a page belonging to the broken tree."
```
Command line options starting with `--` are usually optional, but in this case we want the id to be obligatory, right? If we remove the dashes, the parser will require this argument.

integreat_cms/core/management/commands/repair_tree.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+        parser.add_argument(
+            "--commit",
+            action="store_true",
+            help="Whether changes should be written to the database",
+        )
+
+    # pylint: disable=arguments-differ
+    def handle(self, *args, page, commit, **options):
```
timoludwig, line 35:
```suggestion
    def handle(self, *args, page_id, commit, **options):
```

integreat_cms/core/management/commands/repair_tree.py
```diff
@@ -0,0 +1,88 @@
...
+        r"""
+        Try to run the command
+
+        :param \*args: The supplied arguments
+        :type \*args: list
+
+        :param page: The page ID (node) of a broken tree
+        :type page: str
```
timoludwig, line 43:
```suggestion
        :param page_id: The page ID (node) of a broken tree
        :type page_id: int
```

integreat_cms/core/management/commands/repair_tree.py
```diff
@@ -0,0 +1,88 @@
...
+
+        :param \**options: The supplied keyword options
+        :type \**options: dict
+
+        :raises ~django.core.management.base.CommandError: When the input is invalid
+        """
+        if not page:
+            raise CommandError
+        root_node = Page.objects.get(id=page).get_root()
+        self.calculcate_left_right_values(root_node, 1, commit)
```
timoludwig, line 56:
```suggestion
        try:
            tree_id = Page.objects.get(id=page_id).tree_id
        except Page.DoesNotExist as e:
            raise CommandError(f'The page with id "{page_id}" does not exist.') from e
        try:
            root_node = Page.objects.get(tree_id=tree_id, parent=None)
        except (Page.DoesNotExist, Page.MultipleObjectsReturned) as e:
            raise CommandError(
                f'The root page of tree id "{tree_id}" cannot be determined: {e}'
            ) from e
        action = "Fixing" if commit else "Detecting problems in"
        self.print_info(f"{action} tree with id {tree_id}...")
        self.pages_seen = []
        self.calculate_left_right_values(root_node, 1, commit)
        self.check_for_orphans(tree_id)
```
We could improve the error messages here in case the input does not make sense.
Also, the `get_root()` uses the tree fields itself, so I would suggest to rather check for `parent=None` instead.

integreat_cms/core/management/commands/repair_tree.py
```diff
@@ -0,0 +1,88 @@
...
+
```
timoludwig, line 57:
```suggestion

    def check_tree_fields(self, tree_node, left, right):
        """
        Check whether the tree fields are correct

        :param tree_node: The current node
        :type tree_node: ~integreat_cms.cms.models.pages.page.Page

        :param left: The new left value of the node
        :type left: int

        :param right: The new right value of the node
        :type right: int

        :return: Whether the tree fields of the node are valid
        :rtype: bool
        """
        valid = True
        self.stdout.write(self.bold(f"Page {tree_node.id}:"))
        self.print_success(f"\tparent_id: {tree_node.parent_id}")
        if not tree_node.parent or tree_node.tree_id == tree_node.parent.tree_id:
            self.print_success(f"\ttree_id: {tree_node.tree_id}")
        else:
            self.print_error(
                f"\ttree_id: {tree_node.tree_id} → {tree_node.parent.tree_id}"
            )
            valid = False
        if tree_node.parent_id:
            if tree_node.depth == tree_node.parent.depth + 1:
                self.print_success(f"\tdepth: {tree_node.depth}")
            else:
                self.print_error(
                    f"\tdepth: {tree_node.depth} → {tree_node.parent.depth + 1}"
                )
                valid = False
        elif tree_node.depth == 1:
            self.print_success(f"\tdepth: {tree_node.depth}")
        else:
            self.print_error(f"\tdepth: {tree_node.depth} → 1")
            valid = False
        if tree_node.lft == left:
            self.print_success(f"\tlft: {tree_node.lft}")
        else:
            self.print_error(f"\tlft: {tree_node.lft} → {left}")
            valid = False
        if tree_node.rgt == right:
            self.print_success(f"\trgt: {tree_node.rgt}")
        else:
            self.print_error(f"\trgt: {tree_node.rgt} → {right}")
            valid = False
        return valid

    def check_for_orphans(self, tree_id):
        """
        Check whether orphans exist (pages with the same tree_id, but its ancestors are in another tree)

        :param tree_id: The current tree id
        :type tree_id: int
        """
        if orphans := Page.objects.filter(tree_id=tree_id).exclude(
            id__in=self.pages_seen
        ):
            self.print_error(
                "\n💣 Orphans detected! The following pages share the tree id "
                f"{tree_id} but don't have a relation to the root node:"
            )
            for orphan in orphans:
                self.stdout.write(self.bold(f"Page {orphan.id}:"))
                self.print_error(
                    f"\tparent_id: {orphan.parent_id}\n"
                    f"\tparent.tree_id: {orphan.parent.tree_id}"
                )
                self.stdout.write(
                    self.bold(
                        f"\tdepth {orphan.depth}\n"
                        f"\tlft: {orphan.lft}\n"
                        f"\trgt: {orphan.rgt}"
                    )
                )

```
I'd add two additional methods here:
- `check_tree_fields()` compares the tree fields to the new values and prints them in the according color
- `check_for_orphans()` Checks whether any pages were not handled by the recursive functions, which indicates that other trees might be broken as well.

integreat_cms/core/management/commands/repair_tree.py
```diff
@@ -0,0 +1,88 @@
...
+    def calculcate_left_right_values(self, tree_node, left, commit):
+        """
+        Recursively calculate the Left and Right value for a given node and its
```
timoludwig, line 60:
```suggestion
    def calculate_left_right_values(self, tree_node, left, commit):
        """
        Recursively calculate the left and right value for a given node and its
```
Small typo

integreat_cms/core/management/commands/repair_tree.py
```diff
@@ -0,0 +1,88 @@
...
+        :param commit: Whether changes should be written to the database
+        :type commit: bool
+
+        :return: The new right value of the node
+        :rtype: int
+        """
+        right = left
+
+        for child in tree_node.get_children().order_by("lft"):
+            right = self.calculcate_left_right_values(child, right + 1, commit)
```
timoludwig, line 78:
```suggestion
        for child in tree_node.children.all():
            right = self.calculate_left_right_values(child, right + 1, commit)
```
`get_children()` uses the internal tree fields to determine the children - this might not work when the structure is broken. Instead, I suggest to go via the `parent` field - or rather its reverse related name `children`.
Also, ordering is already by `lft` by default.

integreat_cms/core/management/commands/repair_tree.py
```diff
@@ -0,0 +1,88 @@
...
+
+        right += 1
+
+        logger.warning("id=%i, lft=%s, rgt=%s", tree_node.id, left, right)
+        if commit:
```
timoludwig, line 83:
```suggestion
        if not valid and commit:
            if tree_node.parent:
                tree_node.tree_id = tree_node.parent.tree_id
                tree_node.depth = tree_node.parent.depth + 1
```
I suggest to also update the internal fields `tree_id` and `depth` in case they are incorrect.

integreat_cms/core/management/commands/repair_tree.py
```diff
@@ -0,0 +1,88 @@
...
+            tree_node.rgt = right
+            tree_node.lft = left
+            tree_node.save()
+
```
timoludwig, line 87:
```suggestion
            logger.info("Fixed tree fields of %r", tree_node)

        self.pages_seen.append(tree_node.id)

```
Now, let's remember all pages that have been handled during this run in order to recognize orphans later on...

integreat_cms/core/management/commands/repair_tree.py
```diff
@@ -0,0 +1,88 @@
...
+logger = logging.getLogger(__name__)
+
+
+class Command(LogCommand):
+    """
+    Management command to repair a broken Treebeard page tree
+    """
+
+    help = "Search & replace links in the content"
+
```
timoludwig, line 17:
```suggestion
    pages_seen = None
```

timoludwig (changes requested):
Awesome, this is very much needed :heart_eyes: 

In general, it seems to work for a few test cases I set up.
However, it did not catch problems in these cases:
- Inconsistencies between the `tree_id` and the `parent` field
- Inconsistencies between the `depth` and the `parent` field
- When the `lft` and `rgt` values are incorrect in a way that the page is not included in `get_children()`

I added a few suggestions that take the `parent` field as single source of truth and adapts all internal tree fields accordingly. I hope this works completely independent of how bad the tree structure is broken.
However, I did not validate my suggestion against the real database, so I would be very interested in the result it produces.

integreat_cms/core/management/commands/repair_tree.py
```diff
@@ -0,0 +1,184 @@
...
+logger = logging.getLogger(__name__)
+
+
+class Command(LogCommand):
+    """
+    Management command to repair a broken Treebeard page tree
+    """
+
+    help = "Search & replace links in the content"
```
timoludwig, line 16:
```suggestion
    help = "Repair broken tree structure"
```

integreat_cms/core/management/commands/repair_tree.py
```diff
@@ -0,0 +1,184 @@
...
+        :return: The new right value of the node
+        :rtype: int
+        """
+        right = left
+
+        for child in tree_node.children.all():
+            right = self.calculate_left_right_values(child, right + 1, commit)
+
+        right += 1
+
```
timoludwig, line 172:
```suggestion

        valid = self.check_tree_fields(tree_node, left, right)

```
Sorry, somehow messed up my suggestions and forgot this call :sweat_smile:

integreat_cms/core/management/commands/repair_tree.py
```diff
@@ -0,0 +1,184 @@
...
+            root_node = Page.objects.get(tree_id=tree_id, parent=None)
+        except (Page.DoesNotExist, Page.MultipleObjectsReturned) as e:
+            raise CommandError(
+                f'The root page of tree id "{tree_id}" cannot be determined: {e}'
+            ) from e
+        action = "Fixing" if commit else "Detecting problems in"
+        self.print_info(f"{action} tree with id {tree_id}...")
+        self.pages_seen = []
+        self.calculate_left_right_values(root_node, 1, commit)
+        self.check_for_orphans(tree_id)
```
timoludwig, line 68:
```suggestion
            root_node = Page.objects.get(id=page_id)
        except Page.DoesNotExist as e:
            raise CommandError(f'The page with id "{page_id}" does not exist.') from e
        # Traversing to root node
        while root_node.parent:
            root_node = root_node.parent
        action = "Fixing" if commit else "Detecting problems in"
        self.print_info(f"{action} tree with id {root_node.tree_id}...")
        self.pages_seen = []
        self.calculate_left_right_values(root_node, 1, commit)
        self.check_for_orphans(root_node.tree_id)
```
Not sure if that's a bad idea, but we could also traverse the `parent` field to get the root node instead of going via the `tree_id`... this would have the advantage that we really fix the given page, also in the case where the given page is an orphan itself...

timoludwig (changes requested):

integreat_cms/core/management/commands/repair_tree.py
svenseeberg: Just noticed. Thanks for providing the fix before I stumble over the problem ^^

integreat_cms/core/management/commands/repair_tree.py
svenseeberg: As we decided that the parent field is our source of truth, this would be the right choice.

Issue #1632, opened 2022-08-12
Unsaved warning always shows up when navigating to sbs-view


david-venhoff:
### Steps to Reproduce

1. Go to a page
2. Select the translation direction
3. Click on 'Show translator view'
4. See warning, even though nothing was changed

### Expected Behavior

No warning occurs, the sbs-viewer opens

### Actual Behavior

The unsaved warning shows up

### Additional Information

Related to: #2013
This happens probably because the `direction of translation` selector is part of the page form, and the unsaved warning shows up if there is any input event within that form.

Issue #2013, opened 2023-01-18
Browser warning about unsaved content in side-by-side view


svenseeberg:
### Describe the Bug
When editing and saving information in the side by side view, some browsers display an unsaved content warning. This should be suppressed.

### Steps to Reproduce

1. Open any translation side by side view in Chrome 109 
2. Edit and click publish
3. See browser warning about loosing unsaved content

### Expected Behavior
The browser warning should be suppressed for the save action.

### Actual Behavior
Browser does not recognize that the publish button is saving content.

### Additional Information
Related to: #1632
https://chat.tuerantuer.org/digitalfabrik/pl/7mgbz37i9pgy9jpfbugq5htayo

JoeyStk:
The same bug occurs when leaving the German view to translation view

timoludwig:
> The same bug occurs when leaving the German view to translation view

Indeed, there is a different ticket for this: #1632

timoludwig:
This was apparently fixed in #2243.

david-venhoff:
I can still (or again) reproduce this, by opening the sbs view, writing something in the text field and then clicking save. Then the unsaved warning still shows up

osmers:
Closing as side by side view is rarely used, can reopen if reports about this behaviour come in from municipalities

PR #2243, opened 2023-04-23
Suppress unsaved warnings for form elements that don't affect content

membralala:
### Short description

This PR should prevent a browser alert for unsaved content caused by form elements that don't actually change the content. This was the case for:

- The selection input on page and imprint form view in the "Translator View" box
- The tinyMCE comparison window (right side) in the side-by-side view
- The minor edit checkbox on page form, POI form and event form

### Side effects

- I don't think, that there are side effects

### Resolved issues

Fixes: #1632, (#2013)

*As a side node:* Neither me nor the service team could reproduce the chrome related bug as described in #2013, however I found that the warning e.g. showed up after toggling the button which shows differences to a previous version. I think this might be the issue that occurred there, but I'd be glad, if any reviewers will also test this with chrome.

### DIFF
--- a/integreat_cms/cms/templates/events/event_form_sidebar/minor_edit_box.html
+++ b/integreat_cms/cms/templates/events/event_form_sidebar/minor_edit_box.html
@@ -11,7 +11,7 @@
 {% endblock collapsible_box_title %}
 {% block collapsible_box_content %}
     <div>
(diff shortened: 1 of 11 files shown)


MizukiTemma (approved):
Thank you for the PR 🙂 It works as expected!

JoeyStk (approved):
Thank you! I was testing this in Firefox and Chrome, and either way it worked. I only have two minor suggestions below

Issue #2246, opened 2023-04-24
`AttributeError` in PDF export


timoludwig:
### Describe the Bug

When a link has no link text, the PDF fails with an `AttributeError`.
This is the case when images are embedded with `<img>` tags.

### Steps to Reproduce

1. Insert img tag in page
2. Export as PDF

### Expected Behavior

The PDF should be created

### Actual Behavior

An error occurs:
```
AttributeError: 'NoneType' object has no attribute 'split'
```

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
Apr 24 10:25:04 ERROR   django.request - 500 Internal Server Error: /lkmuenchen/de/wp-json/ig-mpdf/v1/pdf/?url=%2Flkmuenchen%2Fde%2Finformationen-fuer-menschen-aus-der-ukraine%2Faufenthaltserlaubnis-beantragen
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/api/decorators.py", line 123, in wrap
    return function(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/api/v3/pdf_export.py", line 57, in pdf_export
    return generate_pdf(region, language_slug, pages)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/decorators/cache.py", line 44, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/utils/pdf_utils.py", line 107, in generate_pdf
    html = get_template("pages/page_pdf.html").render(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/backends/django.py", line 61, in render
    return self.template.render(context)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/template/base.py", line 170, in render
    return self._render(context)
[...]

PR #2248, opened 2023-04-25
Fix `AttributeError` in PDF export

MizukiTemma:
### Short description

This PR fixes `AttributeError` which occurs in PDF export when an img tag is included in the page.

### Proposed changes

- Truncate only when `elem.text` is not None.

### Side effects

- This change seems not to affect, so far as I tested, the link truncating, which is already introduced. There might be however a case I didn't catch or other similar problems. 

### Resolved issues

Fixes: #2246


JoeyStk (approved):
Thank you a lot for the quick fix.
I tried to break it, but couldn't. All of the following didn't break the creation of pdfs

- adding images
- adding links (they get truncated correctly)
- adding icons and special characters
- adding formatting and colors
- adding videos


To me it looks good

integreat_cms/release_notes/current/unreleased/2246.yml
```diff
@@ -0,0 +1,2 @@
+en: Fix AttributeError in PDF export
+de: AttributeFehler beim PDF-Export beheben
```
timoludwig, line 2:
```suggestion
en: Fix error in PDF export
de: Fehler beim PDF-Export beheben
```
The users never got to see the `AttributeError`, this is only shown in debug mode and in the server logs. As the release notes are targeted at the users of the system and not the developers, I think the error description should match what they experienced.

timoludwig (approved):

Issue #2054, opened 2023-02-06
 Do not check links in inactive languages


timoludwig:
### Motivation

At the moment, the link checker is active for all languages of a region, independent from their status.

### Proposed Solution

Ignore inactive languages in the link checker (invisible languages are still ok I think):
- Add a new `object_filter` (see [docs](https://github.com/DjangoAdminHackers/django-linkcheck#linklist-classes)) for all [link lists](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/linklists.py#L57)
- Whenever the `active` field of a language tree node is toggled from `True` to `False`, delete all link objects which belong to translations in that language
- Whenever the `active` field of a language tree node is toggled from `False` to `True`, trigger the link scan for all translations in that language

PR #2251, opened 2023-04-26
Do not check links in inactive languages

MizukiTemma:
### Short description

This PR modifies link check so it will not check links in translations of inactive languages.

### Proposed changes

- Delete all links in the language which is changed to inactive. 
- Trigger link check when a language gets set to active.
- Add object filters to exclude translations in inactive languages.

### Side effects

- Restration/deleting of links sometimes takes a while (doesn't finish in a second).

### Resolved issues

Fixes: #2054 

### DIFF
--- a/integreat_cms/cms/linklists.py
+++ b/integreat_cms/cms/linklists.py
@@ -1,10 +1,18 @@
 from django.conf import settings
+from django.db.models import Exists, OuterRef
 from linkcheck import Linklist
 
-from .models import Event, EventTranslation, Page, PageTranslation, POITranslation
+from .models import (
+    Event,
+    EventTranslation,
+    LanguageTreeNode,
+    Page,
+    PageTranslation,
+    POITranslation,
+)
 
 
-class ContentTranslationLinklist(Linklist):
+class ActiveLanguageLinklist(Linklist):
     """
     Base class for content translation link lists
     """
@@ -14,7 +22,7 @@ class ContentTranslationLinklist(Linklist):
     @classmethod
     def filter_callable(cls, objects):
         """
-        Get only latest versions for non-archived events/locations
+        Get only translations in active languages
 
         :param objects: Objects to be filtered
         :type objects: ~django.db.models.query.QuerySet
@@ -22,23 +30,33 @@ class ContentTranslationLinklist(Linklist):
         :return: Objects that passed the filter
         :rtype: ~django.db.models.query.QuerySet
         """
-        return objects.filter(
-            **{f"{objects.model.foreign_field()}__archived": False}
-        ).distinct(f"{objects.model.foreign_field()}__pk", "language__pk")
+        # Because of large overhead, we only want this filter during the findlinks management command.
+        if settings.LINKCHECK_COMMAND_RUNNING:
+            # Exclude links in inactive languages
+            objects = objects.annotate(
+                active=Exists(
+                    LanguageTreeNode.objects.filter(
+                        region=OuterRef(f"{objects.model.foreign_field()}__region"),
+                        language=OuterRef("language"),
+                        active=True,
+                    )
+                )
+            ).filter(active=True)
+
+        return objects
 
 
-class PageTranslationLinklist(Linklist):
+class PageTranslationLinklist(ActiveLanguageLinklist):
     """
     Class for selecting the PageTranslation model for link checks
     """
 
     model = PageTranslation
-    html_fields = ["content"]
 
     @classmethod
     def filter_callable(cls, objects):
         """
-        Get only latest versions for non-archived pages
+        Get only latest versions for non-archived pages in active languages
 
         :param objects: Objects to be filtered
         :type objects: ~django.db.models.query.QuerySet
@@ -46,14 +64,44 @@ class PageTranslationLinklist(Linklist):
         :return: Objects that passed the filter
         :rtype: ~django.db.models.query.QuerySet
         """
-        # Since filtering the archived pages causes a lot of overhead, we only want this during the findlinks management command.
-        if settings.LINKCHECK_EXCLUDE_ARCHIVED_PAGES:
+        # Apply filter of parent class
+        objects = super().filter_callable(objects)
+        # Because of large overhead, we only want this filter during the findlinks management command.
+        if settings.LINKCHECK_COMMAND_RUNNING:
+            # Exclude archived pages
             pages = Page.objects.all().cache_tree(archived=False)
             objects = objects.filter(page__in=pages)
+
         return objects.distinct("page__pk", "language__pk")
 
 
-class EventTranslationLinklist(ContentTranslationLinklist):
+class NonArchivedLinkList(ActiveLanguageLinklist):
+    """
+    Class for excluding archived events and locations
+    """
+
+    @classmethod
+    def filter_callable(cls, objects):
+        """
+        Get only latest translations for non-archived events/locations in acitive languages
+
+        :param objects: Objects to be filtered
+        :type objects: ~django.db.models.query.QuerySet
+
+        :return: Objects that passed the filter
+        :rtype: ~django.db.models.query.QuerySet
+        """
+        # Apply filter of parent class
+        objects = super().filter_callable(objects)
+        # Exclude archived events/locations
+        objects = objects.filter(
+            **{f"{objects.model.foreign_field()}__archived": False}
+        ).distinct(f"{objects.model.foreign_field()}__pk", "language__pk")
+
+        return objects
+
+
+class EventTranslationLinklist(NonArchivedLinkList):
     """
     Class for selecting the EventTranslation model for link checks
     """
@@ -63,7 +111,7 @@ class EventTranslationLinklist(ContentTranslationLinklist):
     @classmethod
     def filter_callable(cls, objects):
         """
-        Get only translations of upcoming events
+        Get only translations of upcoming events in active languages
 
         :param objects: Objects to be filtered
         :type objects: ~django.db.models.query.QuerySet
@@ -71,11 +119,16 @@ class EventTranslationLinklist(ContentTranslationLinklist):
         :return: Objects that passed the filter
         :rtype: ~django.db.models.query.QuerySet
         """
+        # Apply filter of parent class
+        objects = super().filter_callable(objects)
+        # Exclude past events
         upcoming_events = Event.objects.filter_upcoming()
-        return objects.filter(event__in=upcoming_events)
+        objects = objects.filter(event__in=upcoming_events)
+
+        return objects
 
 
-class POITranslationLinklist(ContentTranslationLinklist):
+class POITranslationLinklist(NonArchivedLinkList):
     """
     Class for selecting the POITranslation model for link checks
     """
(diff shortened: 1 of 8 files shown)


MizukiTemma:
@JoeyStk 
Thank you for testing! It was because of `object_filter`s which was not implemented correctly 😅 

Can you test this PR again?

JoeyStk (approved):
Yes, now it's working! Thank you very much!

integreat_cms/cms/views/language_tree/language_tree_node_form_view.py
```diff
@@ -42,3 +43,33 @@ def get_form_kwargs(self):
...
+            language = language_tree_node.language
+
+            models = [PageTranslation, EventTranslation, POITranslation]
+            for model in models:
+                filters = {
+                    f"{model.foreign_field()}__region": region,
+                    "language": language,
+                }
+                for translation in model.objects.filter(**filters):
+                    if request.POST.get("active") == "on":
```
timoludwig, line 70: Similar to above: I don't think it's a good idea to rely on the sent post data, instead I would wait for Django to validate and save all the data, and then access e.g. `self.get_object().first().active` to determine whether the selected nodes have been enabled or disabled.

timoludwig (changes requested):
Thanks a lot! :+1:

> Restration/deleting of links sometimes takes a while (doesn't finish in a second).

The time delay is not really the problem here, but the underlying reason for it: It puts a lot of strain on our server, especially when multiple users are working on the content at the same time, this quickly piles up and could cause a lot of server crashes. Therefore I think we need to optimize this before it can be merged.

integreat_cms/cms/views/language_tree/language_tree_node_form_view.py
```diff
@@ -42,3 +43,30 @@ def get_form_kwargs(self):
...
+            region = language_tree_node.region
+            language = language_tree_node.language
+            models = [PageTranslation, EventTranslation, POITranslation]
+            for model in models:
+                filters = {
+                    f"{model.foreign_field()}__region": region,
+                    "language": language,
+                }
+                for translation in model.objects.filter(**filters):
+                    if self.get_object().active:
```
timoludwig, line 68:
```suggestion
            models = [PageTranslation, EventTranslation, POITranslation]
            for model in models:
                filters = {
                    f"{model.foreign_field()}__region": language_tree_node.region,
                    "language": language_tree_node.language,
                }
                distinct = [f"{model.foreign_field()}__pk", "language__pk"]
                for translation in model.objects.filter(**filters).distinct(*distinct):
                    if language_tree_node.active:
```
I think we can limit the translations to the latest versions of each content type and language...

integreat_cms/cms/views/language_tree/language_tree_bulk_actions.py
```diff
@@ -122,3 +144,18 @@ class BulkDisableView(BulkActivateView):
...
+            models = [PageTranslation, EventTranslation, POITranslation]
+            for model in models:
+                filters = {
+                    f"{model.foreign_field()}__region": region,
+                    "language": language,
+                }
+                for translation in model.objects.filter(**filters):
+                    translation.links.all().delete()
+
+        return super().post(request, *args, **kwargs)
```
timoludwig, line 161:
```suggestion
```
Then we could delete this one.

integreat_cms/release_notes/current/unreleased/2054.yml
```diff
@@ -0,0 +1,2 @@
+en: Do not check links in inactive languages
+de: Prüfe keinen Link in inaktiven Sprachen
```
timoludwig, line 2:
```suggestion
en: Disable content in inactive languages for the link checker
de: Deaktiviere Inhalte in inaktiven Sprachen für den Link-Checker
```

timoludwig (changes requested):

In general it looks good! I just have a few smaller suggestions.

integreat_cms/cms/linklists.py
```diff
@@ -38,19 +46,31 @@ class PageTranslationLinklist(Linklist):
...
         """
-        # Since filtering the archived pages causes a lot of overhead, we only want this during the findlinks management command.
-        if settings.LINKCHECK_EXCLUDE_ARCHIVED_PAGES:
+        # Because of large overhead, we only want this filter during the findlinks management command.
+        if settings.LINKCHECK_COMMAND_RUNNING:
+            # Exclude links in inactive languages
+            objects = objects.annotate(
+                active=Exists(
+                    LanguageTreeNode.objects.filter(
+                        region=OuterRef("page__region"),
```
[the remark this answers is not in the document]
MizukiTemma: Nice idea :smiley: I tried and it seems to work well.

integreat_cms/cms/linklists.py
```diff
@@ -38,19 +46,31 @@ class PageTranslationLinklist(Linklist):
...
+                        active=True,
+                    )
+                )
+            ).filter(active=True)
+            # Exclude archived pages
             pages = Page.objects.all().cache_tree(archived=False)
             objects = objects.filter(page__in=pages)
-        return objects.distinct("page__pk", "language__pk")
+
+        return objects
```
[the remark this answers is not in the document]
MizukiTemma: Right...I think it has been removed by mistake :sweat: now revived.

integreat_cms/cms/linklists.py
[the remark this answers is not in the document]
MizukiTemma: Good idea :smiley: I just changed `self` to `objects`

integreat_cms/cms/linklists.py
```diff
@@ -22,6 +30,19 @@ def filter_callable(cls, objects):
...
+            # Exclude links in inactive languages
+            objects = objects.annotate(
+                active=Exists(
+                    LanguageTreeNode.objects.filter(
+                        region=OuterRef(f"{objects.model.foreign_field()}__region"),
+                        language=OuterRef("language"),
+                        active=True,
+                    )
+                )
+            ).filter(active=True)
```
timoludwig, line 44: Probably we need another intermediate class, because `PageTranslationLinklist` does not inherit from the existing `ContentTranslationLinklist` (this is because the `__archived` filter does nor work for pages...)

How about the following structure:
- Base class `ActiveLanguageLinklist(Linklist)` with the new `filter_callable` to exclude inactive languages
- Page link list `PageLinklist(ActiveLanguageLinklist)` inheriting the language filter
- Intermediate base class `NonArchivedLinklist(ActiveLanguageLinklist)` with the archived filter as base for events and pois
- Link lists `EventLinklist(NonArchivedLinklist)` and `POILinklist(NonArchivedLinklist)` inheriting the intermediate class

Also feel free to suggest a different naming of the classes if you have better ideas :laughing:

integreat_cms/cms/linklists.py
```diff
@@ -38,18 +59,20 @@ class PageTranslationLinklist(Linklist):
...
 
         :param objects: Objects to be filtered
         :type objects: ~django.db.models.query.QuerySet
 
         :return: Objects that passed the filter
         :rtype: ~django.db.models.query.QuerySet
         """
-        # Since filtering the archived pages causes a lot of overhead, we only want this during the findlinks management command.
-        if settings.LINKCHECK_EXCLUDE_ARCHIVED_PAGES:
+        # Because of large overhead, we only want this filter during the findlinks management command.
```
timoludwig, line 70:
```suggestion
        # Apply filter of parent class
        objects = super().filter_callable(objects)
        # Because of large overhead, we only want this filter during the findlinks management command.
```
Overriding a method will replace it completely, so the code defined in the parent class is not run in this case.
So we need to explicitly call `super()` if we want to extend the base class (and not replace it).
This is a pretty fundamental concept in Python, so keep this in mind.

integreat_cms/cms/linklists.py
```diff
@@ -63,16 +86,19 @@ class EventTranslationLinklist(ContentTranslationLinklist):
...
-        Get only translations of upcoming events
+        Get only translations of upcoming events in active languages
 
         :param objects: Objects to be filtered
         :type objects: ~django.db.models.query.QuerySet
 
         :return: Objects that passed the filter
         :rtype: ~django.db.models.query.QuerySet
         """
+        # Exclude past events
```
timoludwig, line 97:
```suggestion
        # Apply filter of parent class
        objects = super().filter_callable(objects)
        # Exclude past events
```
Looks like a bug to me, that this was missing? Interestingly, we should have links  in archived events at the moment? Never noticed this :sweat_smile:

integreat_cms/cms/linklists.py
```diff
@@ -82,6 +108,20 @@ class POITranslationLinklist(ContentTranslationLinklist):
...
+
+        :param objects: Objects to be filtered
+        :type objects: ~django.db.models.query.QuerySet
+
+        :return: Objects that passed the filter
+        :rtype: ~django.db.models.query.QuerySet
+        """
+
+        return objects
+
```
timoludwig, line 124:
```suggestion
```
This would just override any parent methods without doing anything

timoludwig (changes requested):
Thanks for your changes! :+1:

A few problems with the class inheritance remain I think...

integreat_cms/cms/linklists.py
```diff
@@ -1,10 +1,18 @@
...
+    EventTranslation,
+    LanguageTreeNode,
+    Page,
+    PageTranslation,
+    POITranslation,
+)
 
 
-class ContentTranslationLinklist(Linklist):
+class ActiveLanguageLink(Linklist):
```
timoludwig, line 15:
```suggestion
class ActiveLanguageLinklist(Linklist):
```
I think it makes sense to point out that this class is a link list, not just a link.

integreat_cms/cms/linklists.py
```diff
@@ -38,22 +57,52 @@ class PageTranslationLinklist(Linklist):
     @classmethod
```
timoludwig, line 57: We can remove `html_fields = ["content"]` now since it's inherited from the base class.

integreat_cms/cms/linklists.py
```diff
@@ -38,22 +57,52 @@ class PageTranslationLinklist(Linklist):
...
     def filter_callable(cls, objects):
         """
-        Get only latest versions for non-archived pages
+        Get only latest versions for non-archived pages in acitive languages
```
timoludwig, line 60:
```suggestion
        Get only latest versions for non-archived pages in active languages
```

integreat_cms/cms/linklists.py
```diff
@@ -38,22 +57,52 @@ class PageTranslationLinklist(Linklist):
...
+        if settings.LINKCHECK_COMMAND_RUNNING:
+            # Exclude archived pages
             pages = Page.objects.all().cache_tree(archived=False)
             objects = objects.filter(page__in=pages)
+
         return objects.distinct("page__pk", "language__pk")
 
 
-class EventTranslationLinklist(ContentTranslationLinklist):
+class NonArchivedLinkList(ActiveLanguageLink):
```
timoludwig, line 79:
```suggestion
class NonArchivedLinkList(ActiveLanguageLinklist):
```

timoludwig (approved):
:rocket:

PR #2253, opened 2023-05-01
Change secondary buttons to ghost and outline buttons

JoeyStk:
### Short description
This PR introduces outline buttons as new secondary buttons, and ghost buttons as buttons that fulfill a different task than the main purpose of the pages. Additionally I made some minor changes to existing buttons

### Proposed changes

- Add outline and ghost buttons, and change secondary buttons to outline buttons
- Change the gray tone of disabled buttons to a slightly lighter gray
- Change "Reset filter" button in feedback and archived feedback from a gray button to the light blue buttons
- Add documentation for it

### Side effects

- I tried to go through all buttons, but still might have overlooked some. Please check, if you can find any gray (not disabled) buttons.

### Resolved issues

Fixes: #


MizukiTemma (approved):
Thank you for the PR :+1:  it looks good!

I have one small suggestion (though not specified in the issue scope and not mandatory): can we maybe make the filter buttons (in Feedback list, Media library, page list, event list) look exactly like the new preview buttons (light sky blue background + blue font)?

And probably a new release note :wink:

JoeyStk:
Thank you for your review. 

- I added a release note 
- and for your suggestion I asked @hauf-toni for some input, and will get back to you as soon as I have a response

integreat_cms/cms/constants/colors.py
```diff
@@ -29,3 +29,7 @@
     "#fabebe",
     "#ffd8b1",
 ]
+
+#: Instruction for deciding between outline and ghost button.
+#: Please use outline buttons for secondary buttons - thus buttons that still fulfill the main purpose of the shown view, just in secondary hierarchy.
+#: Please use ghost buttons for buttons that have a different purpose than the other buttons on the shown view
```
timoludwig, line 35: Two problems:
1. This will not be shown in the generated html documentation, because the `#:` syntax only works as docstring for something specific (e.g. an attribute or a function), if you just place this as inline comment, it will probably never be read...
2. The placement does not really make sense, yes, this is somehow about colors, but only for graph colors and not related to UI design in any way. So I doubt devs will read this when wondering about general design questions. I would prefer it if we would find a separate place for this, e.g. a new documentation page "Design guidelines" or similar?

integreat_cms/cms/templates/feedback/_admin_feedback_filter_form.html
```diff
@@ -38,7 +38,7 @@
         </div>
     </div>
     <div class="w-full flex p-4 justify-end gap-4">
-        <a id="filter-reset" class="btn btn-gray">{% translate "Reset filter" %}</a>
+        <a id="filter-reset" class="btn btn-light-blue">{% translate "Reset filter" %}</a>
```
timoludwig, line 41:
```suggestion
        <a id="filter-reset" class="btn btn-ghost">{% translate "Reset filter" %}</a>
```

> I have one small suggestion (though not specified in the issue scope and not mandatory): can we maybe make the filter buttons (in Feedback list, Media library, page list, event list) look exactly like the new preview buttons (light sky blue background + blue font)?

I agree, I also understood the guideline in a way that we only have three different kind of buttons, and convert the light blue buttons to ghost buttons...

integreat_cms/release_notes/current/unreleased/2170.yml
```diff
@@ -0,0 +1,2 @@
+en: Change styling for secondary buttons
+de: Ändere das Styling für sekundäre Knöpfe
```
timoludwig, line 2:
```suggestion
de: Ändere das Styling für sekundäre Buttons
```

timoludwig (changes requested):
Thanks a lot! :+1:
I think this really makes it more clear what a button does.
I agree however that we should merge the light blue buttons and the ghost buttons, since they serve a similar purpose (buttons that don't trigger an immediate action, but just show/hide something) and already look similar.

integreat_cms/cms/constants/colors.py
JoeyStk: I added the documentation as proper .rst files to the "basic concepts" section of our documentation. I hope that was the solution you had in mind :)

docs/src/design-guidelines.rst
```diff
@@ -0,0 +1,11 @@
...
+Design Guidelines
+******************
+
+Buttons
+=======
+
+In our project we want to use three buttons all together: regular buttons (Tailwind class "btn"), outline buttons (Tailwind classes "btn btn-outline")
+and ghost buttons (Tailwind classes "btn btn-ghost"). Regular buttons are used, if they are the main call to action on the page.
+Please use outline buttons for secondary buttons - thus buttons that still fulfill the main purpose of the shown view, just in secondary hierarchy.
+Please use ghost buttons for buttons that have a different purpose than the other buttons on the shown view.
```
timoludwig, line 11:
```suggestion
In our project we use three types of buttons:

* Primary buttons (css class ``btn``):
  The main call to action on the page, e.g. submitting a form or opening another page
* Outline buttons (css classes ``btn btn-outline``):
  The secondary call to action on the page, e.g. submitting a form as draft
* Ghost buttons (css classes ``btn btn-ghost``):
  Buttons that don't trigger an immediate action, but only hide/show additional components, e.g. filters or previews.
```
I would format this a little bit differently to make it more readable.

timoludwig (changes requested):

JoeyStk:
Oh, good point. Looks like I messed that up while rebasing. The commits are squashed now

timoludwig (approved):

hauf-toni:
> Thank you for your review.
> 
> * I added a release note
> * and as for your suggestion I asked @hauf-toni for some input, and will get back to you as soon as I have a response



> Thank you for the PR 👍 it looks good!
> 
> I have one small suggestion (though not specified in the issue scope and not mandatory): can we maybe make the filter buttons (in Feedback list, Media library, page list, event list) look exactly like the new preview buttons (light sky blue background + blue font)?
> 
> And probably a new release note 😉

From an ui/ux perspective I would recommend using the new preview buttons as @MizukiTemma suggested. The usual/"soon to be usual" use case for this button is not to trigger an immediate action (like the CTA + secondary button), it would be best suited for the filter as well 👍

Issue #2252, opened 2023-04-27
Unable to type anything in the box"URL-Parameter" while creating a new region


karthik-integreat:
While creating a new region, I am not able to type anything on the box "URL-Parameter".
It is also the case for already created regions.

PR #2254, opened 2023-05-01
Change URL parameter field to make it editable

JoeyStk:
### Short description
This PR fixes the issue with the not editable URL parameter field by adding contenteditable to the parent element.

### Proposed changes

- ~~Add contenteditable to parent element~~
- Change the CSS selector from &:read-only to &[readonly] as suggested [here](https://stackoverflow.com/questions/69174538/css-read-only-is-applied-to-elements-that-are-not-readonly).

### Side effects

- ~~I'm not sure if this is the best solution. I guess the best solution would be to let the input field know it's writeable. However every trick I tried didn't work, and therefore this is the next best solution. Let me know, what you think.~~
- I checked the places I know the read-only field is needed, but I'm not sure I know all the places. Can you please check all the places you know where a readonly field is needed to see if this doesn't mess anything up? 

### Resolved issues

Fixes: #2252

### DIFF
--- a/integreat_cms/release_notes/current/unreleased/2252.yml
+++ b/integreat_cms/release_notes/current/unreleased/2252.yml
@@ -1,0 +1,2 @@
+en: Fix the problem of the URL parameter field not being editable
+de: Behebe das Problem, dass das URL-Parameterfeld nicht editierbar ist
--- a/integreat_cms/static/src/css/style.scss
+++ b/integreat_cms/static/src/css/style.scss
@@ -217,8 +217,7 @@ select,
 [type="tel"],
 [type="time"],
 [type="week"],
-textarea,
-.slug-field {
+textarea {
     &:read-only {
         @apply text-gray-500 pointer-events-none cursor-not-allowed;
     }


MizukiTemma:
In develop branch currently, not only the region slug in the region form but also slug in the content forms (page, event, poi) are not editable. But they must be editable too, right?

Then it looks working as expected to me 🙂

integreat_cms/release_notes/current/unreleased/2252.yml
```diff
@@ -0,0 +1,2 @@
+en: Fix not editable problem of URL parameter field
+de: Behebung des Problems, dass das URL-Parameterfeld nicht editierbar war
```
timoludwig, line 2:
```suggestion
en: Fix the problem of the URL parameter field not being editable
de: Behebe das Problem, dass das URL-Parameterfeld nicht editierbar ist
```
Similar to git commit messages, we use the imperative mood for our release notes.

timoludwig (changes requested):
Thanks, this works as expected! :+1:
I just have a small suggestion for improvement...

JoeyStk:
> In develop branch currently, not only the region slug in the region form but also slug in the content forms (page, event, poi) are not editable. But they must be editable too, right?

AFAIK yes

integreat_cms/static/src/css/style.scss
```diff
@@ -236,6 +235,9 @@ textarea,
         &:focus {
             box-shadow: none;
         }
+        &:read-only {
+            @apply text-gray-500 pointer-events-none cursor-not-allowed;
+        }
```
timoludwig, line 240:
```suggestion
```
Ah, I'm very sorry about my incorrect suggestion, this in fact no needed, since `[type="text"]` is already styled accordingly via the rule above in line 208, thus it does not need to be repeated here.

timoludwig (changes requested):

MizukiTemma (approved):
👍

timoludwig (approved):

Issue #2247, opened 2023-04-25
Only execute MT if translation is outdated


JoeyStk:
### Describe the Bug

At the moment it's possible to execute a MT even if the translation is up-to-date. This however is not really user-friendly and has lead to some confusion for CMS users.

### Steps to Reproduce

1. Go to a page/event or poi form that has an up-to-date translation
2. Click on Automatic Translation inside the "Implications on other translations" box
3. Check checkbox in "Update existing translations: " section for a language that has an up-to-date translation
4. Save page/event or poi form and see how budget gets effected by this translation

### Expected Behavior

- The translation can't be executed if there already is an up-to-date translation
- The user gets information that there already is an up-to-date translation
- The MT budget therefore is also not effected

### Actual Behavior
The translation can be executed and the budget is effected

### Additional Information
Take a look at the [pre-existing translation statuses](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/constants/translation_status.py). Please note to check if there is an up-to-date machine translation or if there is an up-to-date (human) translation. 
Also please be reminded, that this goes for all content types (namely pages, events and POI)

Issue #2119, opened 2023-03-14
Race condition in async tests


timoludwig:
### Describe the Bug

In some cases there seems to be a race condition in asynchronous tests again.

### Steps to Reproduce

No deterministic way of reproducing this, but at least two workflow runs showed this behaviour:
- https://app.circleci.com/pipelines/github/digitalfabrik/integreat-cms/7877/workflows/4cc87299-07f9-4b8a-83f5-3d75bd9ad5c7/jobs/80573
- https://app.circleci.com/pipelines/github/digitalfabrik/integreat-cms/7877/workflows/d0d6d4ca-310d-4363-89bc-6bf66cc5e139/jobs/80617

### Expected Behavior

Async tests should not influence the sync tests

### Actual Behavior

Something goes wrong when specific tests are parallelized together with the async SUMM.AI test

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
============================= test session starts ==============================
platform linux -- Python 3.9.16, pytest-7.2.2, pluggy-1.0.0
django: settings: integreat_cms.core.circleci_settings (from option)
rootdir: /home/circleci/project
plugins: circleci-parallelized-0.1.0, cov-4.0.0, django-4.5.2, icdiff-0.6, testmon-1.4.5, xdist-3.2.0, requests-mock-1.10.0
collecting 2499 items                                                          No timing found for "test_recurrence_rule.TestCreatingIcalRule"
collected 2499 items                                                           
running 69 items due to CircleCI parallelism

tests/summ_ai_api/summ_ai_test.py ......................                 [ 31%]
tests/api/test_api_result.py .....................................       [ 85%]
tests/cms/views/test_public_view_status_code.py ....                     [ 91%]
tests/core/management/commands/test_summ_ai_bulk.py ..

Too long with no output (exceeded 10m0s): context deadline exceeded
  ```

</details>

This is probably a similar problem like the one fixed in #1787.

timoludwig:
The underlying problem: `summ_ai_enabled` is set to `True` in async mode while other sync tests expect the value to be `False` (like in the test data) and the sync tests start before the test db is reset after the async test is finished.

This is a problem when these two tests are assigned to the same container on CircleCI.

Approaches I tried that all didn't work:

- Rewrite the async tests as sync (didn't work because aiohttp didn't work with `asyncio.run(aiohttp_raw_server(fake_summ_ai_server))`)
[...]

PR #2255, opened 2023-05-01
Only execute MT if translation is outdated

JoeyStk:
### Short description

This PR restricts machine translations to translations that are not up-to-date

### Proposed changes

- Check if translation status of target translation is neither UP_TO_DATE nor MACHINE_TRANSLATED
- Add release note

### Side effects

- I think none

### Resolved issues

Fixes: #2247


integreat_cms/deepl_api/deepl_api_client.py
```diff
@@ -141,6 +141,17 @@ def translate_queryset(self, queryset, language_slug):
                     target_language.slug
                 )
 
+                # Before translating, check if translation is not up-to-date
+                if existing_target_translation.translation_state in (
```
MizukiTemma, line 145:
```suggestion
                if existing_target_translation and existing_target_translation.translation_state in (
```

To avoid `Nonetype Object` error, check existence of translation first then status.

integreat_cms/deepl_api/deepl_api_client.py
```diff
@@ -141,6 +141,17 @@ def translate_queryset(self, queryset, language_slug):
...
+                    "UP_TO_DATE",
+                    "MACHINE_TRANSLATED",
+                ):
+                    messages.error(
+                        self.request,
+                        _("There already is an up-to-date translation"),
```
MizukiTemma, line 151:
```suggestion
                        _("There already is an up-to-date translation for {}").format(
                            content_object.best_translation.title
                        ),
```

It's more user friendly to indicate for which content a translation already actual and for which not especially when multiple of contents are translated at once.

MizukiTemma (changes requested):
Thank you for the PR 🙂 This is a nice feature letting users save translation budget.

I tested it with bulk action and have two suggestions.

integreat_cms/deepl_api/deepl_api_client.py
JoeyStk: Oh, that's a very good point. Thank you!

integreat_cms/deepl_api/deepl_api_client.py
JoeyStk: I had wanted to do that too, but couldn't get it to work so pylint wouldn't complain. But this seems to work. Thank you!

MizukiTemma:
MT by checking a box in "implications on other translation" is mentioned in the "Steps to reproduce" in issue description, though in this procedure I think triggering of MT is anyway wished, because it is meant that the content is somehow updated in the source language and the translation of target language to be updated too. But for bulk action the new feature is very helpfull :)

Can we probably introduce this feature also for SUMM.AI, or for SUMM.AI saving budget is not important (due to cost, ect)?

timoludwig:
> Can we probably introduce this feature also for SUMM.AI, or for SUMM.AI saving budget is not important (due to cost, ect)?

I agree, since the added check has nothing to do with DeepL in paarticular, I think the DeepL API client is the wrong place for it, we should rather move this check into the generic MT translation base classes so all MT providers benefit from it.

JoeyStk:
I won't be able to work on it this week anymore, but will move the check to the correct file next week

seluianova (changes requested):
I tested a bit and found one issue. 
Steps to reproduce:

1. Select 2 translations - one up-to-date and one outdated.
2. Execute MT in bulk actions

Result:
Error "There already is an up-to-date translation for {}" is shown for both pages, however only one of them is up-to-date.

I guess the expected result would be:
1 successful translation + 1 error message

integreat_cms/core/utils/machine_translation_provider.py
```diff
@@ -182,3 +182,38 @@ def is_permitted(region, user, content_type):
...
+                    "UP_TO_DATE",
+                    "MACHINE_TRANSLATED",
+                )
+            ):
+                logger.debug(
+                    "There already is an up-to-date translation for %s",
+                    content_object.best_translation.title,
+                )
+                return False
+            return True
```
MizukiTemma, line 219:
```suggestion
    def is_needed(self, queryset, target_language):
        """
        Checks if a machine translation is needed, thus checking if the
        translation status is UP_TO_DATE or MACHINE_TRANSLATED and then
        returns a lit of translations which are to be updated

        :param queryset: The content model which should be translated
        :type queryset: ~django.db.models.query.QuerySet [ ~integreat_cms.cms.models.abstract_content_model.AbstractContentModel ]

        :param target_language: The target language
        :type target_language: ~integreat_cms.cms.models.languages.language.Language

        :return: translations which need to be translated and updated
        :rtype: list
        """
        # Before translating, check if translation is not up-to-date
        to_translate = []
        for content_object in queryset:
            existing_target_translation = content_object.get_translation(
                target_language.slug
            )
            if (
                existing_target_translation
                and existing_target_translation.translation_state
                in (
                    "UP_TO_DATE",
                    "MACHINE_TRANSLATED",
                )
            ):
                logger.debug(
                    "There already is an up-to-date translation for %s",
                    content_object.best_translation.title,
                )
            else:
                to_translate = to_translate + [content_object]
        return to_translate
```

MizukiTemma (changes requested):
These are suggestions to fix the problem reported by @seluianova 

I couldn't add this suggestion in `bulk_action_views.py` since it was too large.
The following is from line 145 to the end of the post function.

if language_node.mt_provider.bulk_only_for_staff and not request.user.is_staff:
            raise PermissionDenied(
                f"Only staff users have the permission to bulk translate {self.form._meta.model._meta.model_name} via {language_node.mt_provider}"
            )

        to_translate = language_node.mt_provider.is_needed(
            request.region, self.get_queryset(), language_node
        )
        if not to_translate:
            messages.error(
                request,
                _("All the selected translations are already up-to-date."
                ),
            )
            return super().post(request, *args, **kwargs)

        for content_object in self.get_queryset():
            if not content_object in to_translate:
                messages.error(
                    request,
                    _("There already is an up-to-date translation for {}").format(
                        content_object.best_translation.title,
                    ),
                )
        
        logger.debug(
            "Machine translation via %s into %r for: %r",
            language_node.mt_provider.name,
            language_node.language,
            to_translate,
        )
        api_client = language_node.mt_provider.api_client(request, self.form)
        api_client.translate_queryset(to_translate, language_node.slug)

        # Let the base view handle the redirect
        return super().post(request, *args, **kwargs)

JoeyStk:
@seluianova @MizukiTemma Thank you so much for you contributions. 

- The requested changes should be applied now. 
- Following @seluianova's recommendation, I disabled the failing tests for now. As far as I understood the situation this PR is blocked until [Issue #2119](https://github.com/digitalfabrik/integreat-cms/issues/2119) is fixed

seluianova:
> I highly doubt that a race condition is really the cause when the tests are failing consistently

As far as I could see, these failures occur because several tests try to translate the same content at the same time.
As a result, one of these tests succeeds and the next one fails because the state is already "machine translated".

Before the introduced changes, these tests seemed more stable because simultaneous translation of the same entity was possible.

timoludwig:
> As far as I understand, these failures occur because several tests try to translate the same content at the same time. As a result, one of these tests succeeds and the next one fails because the state is already "machine translated".
> 
> Before the introduced changes, these tests seemed more stable because simultaneous translation of the same entity was possible, but now it's not.

Ah okay, thanks for the explanation. But in this case we could also define a mapping of page ids per role, such as they don't interfere with each other? And also use a completely different page (which is not translated during the `test_auto_translate_easy_german` test) for the error-check in `test_summ_ai_error_handling`?
Also, I think it would be cool to test the feature of this PR itself, meaning that we validate that really an error occurs when an already translated page is translated again.

seluianova (commented):
Fixed failing tests + added additional tests to check errors:
- There already is an up-to-date translation for {}
- All the selected translations are already up-to-date.

seluianova (approved):
My change request has been addressed, so I am adding an approval from my side

MizukiTemma (approved):
Looks good 😃 Thank you 🎉

Issue #1974, opened 2022-12-15
Allow user to select 2FA methods if they setup more than one


ulliholtgrave:
### Motivation

Right now user are forced to use FIDO2 even if they enabled TOTP, because we are checking FIDO2 first and redirect the user to the MFA authentication afterwards. The check for TOTP gets never reached if FIDO2 keys exist for the user. 

### Proposed Solution

We should provide an link for the user to choose TOTP as a backup if it is activated.

PR #2257, opened 2023-05-02
Refactor MFA setup and make fallback method available

ulliholtgrave:
### Short description

This PR refactors some of our MFA functionality to achieve consistent naming. It also adds a 

### Proposed changes

- Rename MFA to Fido or WebAuthn according to its prevalence
- Add fallback 2FA authentication if multiple methods are set up
- Deactivate TOTP for passwordless authentication endpoint

### Side effects

- Hopefully none

### Resolved issues

Fixes: #1974


integreat_cms/cms/templates/authentication/login_totp.html
```diff
@@ -27,4 +27,15 @@
...
+            <hr class="w-full h-px my-4 bg-gray-200 border-0 dark:bg-gray-300"/>
+            <span class="absolute px-3 font-medium text-gray-900 -translate-x-1/2 bg-white left-1/2">{% translate "or" %}</span>
+        </div>
+        <div class="mb-3 mt-4">
+            <a href="{% url 'public:login_totp' %}" class="btn w-full text-center">
+                <i icon-name="key"></i>
+                {% translate "Use Security Key" %}
+            </a>
+        </div>
+    {% endblock content %}
```
MizukiTemma, line 41:
```suggestion
    {% endif %}
{% endblock content %}
```

MizukiTemma (changes requested):
Thank you for the PR 💪 It works as expected after the small fix in the suggestion is applied 🙂

MizukiTemma (approved):
Thank you 😃 LGTM

integreat_cms/cms/models/users/user_fido_key.py
```diff
@@ -5,9 +5,9 @@
 from ..abstract_base_model import AbstractBaseModel
 
 
-class UserMfaKey(AbstractBaseModel):
+class UserFidoKey(AbstractBaseModel):
```
timoludwig, line 8: For consistency reasons, we could also rename the model to just `FidoKey` (since it's also called `Organization` instead of `UserOrganization`, `Role` instead of `UserRole` etc...)
But not a bit deal, I'm also fine with leaving it as it is.

integreat_cms/cms/views/authentication/totp_login_view.py
```diff
@@ -51,6 +51,29 @@ def dispatch(self, request, *args, **kwargs):
...
+        :type \**kwargs: dict
+
+        :return: Rendered login form
+        :rtype: ~django.http.HttpResponse
+        """
+
+        return render(
+            request,
+            self.template_name,
+            {"can_use_webauthn": self.user.mfa_keys.exists()},
```
timoludwig, line 74: If you just want to add a template context, it's not neceessary to overwrite `get()` - just use [`get_context_data()`](https://docs.djangoproject.com/en/4.2/topics/class-based-views/generic-display/#adding-extra-context) instead.
This also has the advantage that you can use
```
        return render(
            request,
            self.template_name,
            self.get_context_data(**kwargs),
        )
```
in the post view as well, and have the same context variable even if the form validation fails and the post request is rendered again without a call to `get()`.

integreat_cms/cms/templates/authentication/login_totp.html
```diff
@@ -27,4 +27,16 @@
                autofocus />
         <button class="btn-blue btn">{% translate "Confirm" %}</button>
     </form>
+    {% if can_use_webauthn %}
+        <div class="inline-flex items-center justify-center w-full">
+            <hr class="w-full h-px my-4 bg-gray-200 border-0 dark:bg-gray-300" />
+            <span class="absolute px-3 font-medium text-gray-900 -translate-x-1/2 bg-white left-1/2">{% translate "or" %}</span>
+        </div>
+        <div class="mb-3 mt-4">
+            <a href="{% url 'public:login_totp' %}" class="btn w-full text-center">
```
timoludwig, line 36:
```suggestion
            <a href="{% url 'public:login_webauthn' %}" class="btn w-full text-center">
```
I think here you want to link to the webauthn login, right?

integreat_cms/cms/utils/mfa_utils.py
```diff
@@ -24,3 +24,20 @@ def generate_challenge(challenge_len):
...
+    :type request: ~django.http.HttpRequest
+
+    :return: The user id
+    :rtype: int
+    """
+    if "mfa_user_id" in request.session:
+        return request.session["mfa_user_id"]
+    if "mfa_passwordless_user_id" in request.session:
+        return request.session["mfa_passwordless_user_id"]
+    return None
```
timoludwig, line 43: If we don't differentiate between the two session variables, do we really need both? Or could we just use `mfa_user_id` for all cases?
Also, why don't we also return the user object if the id exists? I mean this will be the next step anyway, whenever this function is called. So I guess it would also make sense to call this `get_mfa_user()` and return either a user object or `None`?

timoludwig (changes requested):

Just have a few smaller suggestions.

integreat_cms/cms/models/users/user_fido_key.py
```diff
@@ -5,9 +5,9 @@
...
     """
-    Data model representing a user's multi-factor-authentication (MFA) key
+    Data model representing a user's FIDO key
     """
 
     user = models.ForeignKey(
```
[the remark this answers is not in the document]
ulliholtgrave: I am not really sure about that as there is still some development ongoing with FIDO or if we might start supporting another standard. If so, we might add an type attribute to the model and keep the model and objects as they are. But if you prefer to be clear about the current keys, I would be fine with renaming them :man_shrugging:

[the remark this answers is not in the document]
timoludwig: Ok, but since the model already is called `UserFidoKey`, it will probably be renamed when we add more key types? If we want to stay flexible, shouldn't we just keep it `UserMfaKey`?
I mean the key model fields might also differ for different key types, so I think it would make sense to create an own model for each key type...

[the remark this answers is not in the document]
ulliholtgrave: Okay, I'll streamline it accordingly :+1:

integreat_cms/cms/utils/mfa_utils.py
ulliholtgrave: If I remember correctly, Jonas could think of a situation where attackers could mitigate the 2FA, if they somehow fake the passwordless id and use this function for a 2FA authentication, even if the user didn't activated it.

However, I applied your changes but kept the two separate IDs.

timoludwig: If this vulnerability exists, then your current code does not prevent it.
`get_mfa_user()` returns the user of the `mfa_passwordless_user_id` session variable even if it's called from the totp view and vice versa. If this is not a problem, we can use the same session variable, and if it is a problem, we have to use two different methods.

ulliholtgrave: I removed the extra ID and only use the mfa_user_id.

ulliholtgrave:
@timoludwig I applied all of your suggestions and I would suggest that we merge this PR in the next days to deactivate the TOTP version for passwordless authentication.
Jonas and I could think of a scenario to bypass the 2FA if you manipulate the session variable to skip the password authentication, but as you would still need an FIDO2 token for it I think we can merge this for now. I suggest we investigate this in a later stage and open another issue for it. Especially as this scenario might already be possible on the current development status.

timoludwig (changes requested):
Thanks for your changes! :+1:

There is now a weird bug I don't fully understand:
When passwordless authentication is enabled, the second factor is no longer required to login via the normal login form. You can just login using name and password, without the second factor whatsoever. Only if the login form is accessed from the special passwordless login view, the second factor is being requested (but not the password obviously).

Also, the selection of the backup method is also possible via the passwordless login, which means that TOTP can still be used as passwordless login, if a FIDO key is registered at the same time. Is this intended? Or did we want to remove TOTP completely for passwordless login?

integreat_cms/cms/forms/users/user_form.py
```diff
@@ -116,7 +116,7 @@ def __init__(self, **kwargs):
         if (
             "passwordless_authentication_enabled" in self.fields
             and self.instance.totp_key is None
```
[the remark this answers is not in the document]
ulliholtgrave: Nobody used TOTP for my experiment which probably means that we won't have a problem with your code suggestion. However, if somebody actually uses only TOTP for passwordless authentication I'll volunteer to fix in the DB :D

Issue #2053, opened 2023-02-06
Do not check links in past events


timoludwig:
### Motivation

At the moment, the link checker is active for all events, including past ones.

### Proposed Solution

Ignore past events in the link checker.
Probably, it does not make sense to directly remove them as soon as an event is over, since this would require some kind of cron-job or scheduled task.
In my opinion, it's sufficient to ignore them on the next link scan, so just add a new `object_filter` (see [docs](https://github.com/DjangoAdminHackers/django-linkcheck#linklist-classes)) here: https://github.com/digitalfabrik/integreat-cms/blob/7393f06527f0e5f6538ea8eeb2d97ac90aa0c6e0/integreat_cms/cms/linklists.py#L57

PR #2260, opened 2023-05-05
Do not check links in past events

MizukiTemma:
### Short description

This PR excludes past events from link check.

### Proposed changes

- add a new filter in `EventTranslationLinklist`

### Side effects

- This alone does not remove links in past events directly and immediately. When the link check is triggerd, for example by the command `findlinks` or other stimli, they dissapear from the link list. Is there already a scheduled job for link screening?

### Resolved issues

Fixes: #2053 

### DIFF
--- a/integreat_cms/cms/linklists.py
+++ b/integreat_cms/cms/linklists.py
@@ -1,7 +1,7 @@
 from django.conf import settings
 from linkcheck import Linklist
 
-from .models import EventTranslation, Page, PageTranslation, POITranslation
+from .models import Event, EventTranslation, Page, PageTranslation, POITranslation
 
 
 class ContentTranslationLinklist(Linklist):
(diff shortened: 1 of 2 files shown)


integreat_cms/cms/linklists.py
```diff
@@ -59,6 +59,21 @@ class EventTranslationLinklist(ContentTranslationLinklist):
     """
 
     model = EventTranslation
+    html_fields = ["content"]
```
timoludwig, line 62:
```suggestion
```
This is not required since `EventTranslationLinklist` inherits from `ContentTranslationLinklist` which already sets this field.

timoludwig (changes requested):

> However, I am a little bit unsure about recent events that we will scan and the links being displayed after the event has happened. @timoludwig But this links will probably be solved if we setup the weekly link scanning routine, or?

Exactly, I don't think we need a separate cronjob for this, once a week should be sufficient I guess...
(We just have to merge this salt PR sometime eventually :see_no_evil:)

Issue #2245, opened 2023-04-24
Allow to clone an instance and set all pages on status published


nassabay:
### Motivation
Sometimes a cloned instance (Einrichtung) is already ready to go online, and there is no need to adjust more than a few selected pages. In this case it is extremely tiresome to have to go to every page and set the status to 'published'. It would be great to have an easy solution to get all pages published in one go.

### Proposed Solution
The cloning process should have an optional setting (Haken) to automatically set all cloned pages to status 'published'. This option should not be set by default, 

### Alternatives
A bulk option to publish pages in the page menu.

PR #2261, opened 2023-05-08
Add option to duplicate regions with initial page status

ulliholtgrave:
### Short description

This PR enables team user to keep the status of the pages of a cloned region.

### Proposed changes

- Add boolean field to region_form
- Skip the status reset if the field is set

### Side effects

- If the initial cloning mechanism works as expected no issues should arise from this :man_shrugging: 

### Resolved issues

Fixes: #2245 


JoeyStk (approved):
Thank you very much!
It works for me! I only have one very picky suggestion :sweat_smile:

timoludwig (changes requested):
Awesome, works as expected! :muscle: 
I just suggest to adapt the wording a bit.

Issue #2262, opened 2023-05-08
New "education" icon for location categories


nassabay:
### Motivation
Category Education (Bildung) is missing for the map!  

### Proposed Solution
New Category 

### Additional Context
Bildungsträger, VHS etc fallen da drunter

timoludwig:
@nassabay could you please provide the icon field you want to use for the new category?

PR #2266, opened 2023-05-08
Add education icon to selection for location categories

timobrembeck:
### Short description

Add education icon to selection for location categories

### Proposed changes

- Add new icon
- Make icon available in selection for region categories

### Side effects

- Releasing this needs to be synced with the app

### Resolved issues

Fixes: #2262


MizukiTemma (approved):
LGTM 👍

JoeyStk (commented):
Thank you! This looks very good.
I have one genuine question. I noticed, that we have 13 icons, but only twelve colors now. Is that a problem? Is there a need for one distinct color for every icon?

timoludwig:
> I noticed, that we have 13 icons, but only twelve colors now. Is that a problem? Is there a need for one distinct color for every icon?

Yes, that's also what I mentioned in https://issues.tuerantuer.org/browse/KOMMUNE-569, and I think the service team would like to have an additional color, but they do not yet know which one...

seluianova (commented):
I observe some problem with the resizing of the new icon.
Is it only on my laptop?
image
image

timoludwig:
@seluianova ah, thanks for noticing! apparently, a `viewBox="0 0 40 40"` was missing... could you check again?

seluianova (approved):
Looks good now, thanks!

PR #2267, opened 2023-05-09
Add version view for events

JoeyStk:
### Short description
This PR adds the revision slider for events

### Proposed changes

- Add UI for revision slider for events 
- Add functionality to restore earlier revisions

### Side effects

- I'm not exactly sure if the permissions are correct
- blocked until #2268 has been merged

### Resolved issues

Fixes: #1154 

### DIFF
--- a/integreat_cms/cms/templates/events/event_form.html
+++ b/integreat_cms/cms/templates/events/event_form.html
@@ -69,6 +69,8 @@
                                     <div>
                                         {{ event_translation_form.instance.version }}
                                         <!-- TODO: Events Revisions -->
+                                        (<a href="{% url 'event_revisions' event_id=event.id region_slug=request.region.slug language_slug=language.slug %}"
+   class="text-blue-500 hover:underline">{% translate "Show" %}</a>)
                                     </div>
                                     <div class="font-bold pr-2">{% translate "Publication status" %}:</div>
                                     <div>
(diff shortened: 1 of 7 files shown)


integreat_cms/cms/templates/events/event_form.html
```diff
@@ -69,6 +69,8 @@ <h1 class="heading overflow-hidden text-ellipsis">
                                     <div>
                                         {{ event_translation_form.instance.version }}
                                         <!-- TODO: Events Revisions -->
```
timoludwig, line 71:
```suggestion
```
I guess TODO can be removed now?

integreat_cms/cms/templates/events/event_revisions.html
timoludwig, line 1: Have you considered creating a shared base template for all revision views instead of just copy+pasting the existing one? The current solution has the disadvantage that we always have to edit three templates when we want to change something in the revision views...

integreat_cms/cms/views/events/event_revision_view.py
timoludwig, line 1: Also this view is a lot of duplicated view. Why not create a shared parent class and only outsource the parts that differ into sub-methods that can be overwritten? (Or add class attributes if necessary)

timoludwig (changes requested):
Thanks! I did not yet have a closer look at this, but I think we should generalize the existing view & template instead of just copying + pasting everything.

timoludwig:
This is superseded by #2268.

Issue #1155, opened 2022-02-09
Add version view for locations


timoludwig:
### Motivation

At the moment, we only have the revision slider for pages and the imprint. Since we store all versions for events and POIs as well, we should also make use of it.

### Proposed Solution

Add the same UI for versions of the POI translation model with the possibility to view the diff between versions and to restore old versions.

### Additional Context

Is possible, use a shared view and template for all revision views of all content models.

Related issue:
- #1154

PR #2268, opened 2023-05-09
Add version view for events & locations

MizukiTemma:
### Short description

This PR adds version view for POI.

### Proposed changes

- Make view and template of page version view universal for page and POI

### Side effects

- <s>Note: this PR is still in progress but openred as draft to share progress etc with #2267 </s>
- If this approach looks reasonable, it will be extented for event.

### Resolved issues

Fixes: #1154
Fixes: #1155 


integreat_cms/cms/templates/pages/page_revisions.html
```diff
@@ -6,24 +6,24 @@
...
-                {% blocktrans %}Page versions of "{{ page_title }}"{% endblocktrans %}
+            {% with content_translations.first.title as content_title %}
+                {% blocktrans %}Versions of "{{ content_title }}"{% endblocktrans %}
             {% endwith %}
         </h1>
-        <a href="{% url 'edit_page' page_id=page.id region_slug=request.region.slug language_slug=language.slug %}"
-           class="btn btn-outline">
-            <i icon-name="arrow-left-circle" class="align-top"></i> {% translate "Back to the page form" %}
+        <a href={{form}}
+           class="btn btn-gray">
```
JoeyStk, line 14:
```suggestion
           class="btn btn-outline">
```
Small rebasing error

integreat_cms/cms/views/pages/page_revision_view.py
```diff
@@ -145,28 +175,30 @@ def post(self, request, *args, **kwargs):
             messages.error(request, _("This revision does not exist."))
             return redirect_to_page_revisions
```
JoeyStk, line 176:
```suggestion
            return redirect_to_content_revisions
```

JoeyStk (changes requested):
Thank you very much for the PR and taking over the responsibility to generalize this feature. I wasn't able to work on this PR myself, but have a few suggestions that hopefully help you in the process. 

- I think it would be best to rename page_revisions to generic_revisions or something similar. Same goes for the view. 
- I was looking into the question how to handle the context. I think one solution might be to create a new function get_context, that finds out the content type via the kwargs and then returns the navigation item that should be highlighted in the navigation bar

JoeyStk (approved):
To me this looks very good now. The only thing I'm not entirely sure about is, whether or not we should specify the content type in ["Back to form"](https://github.com/digitalfabrik/integreat-cms/pull/2268/files#diff-f99ce2980aa8c5acb976c0faae732414c0e70d66f329ac72fd22dd2f4ba4b7cfR14). To me either way is fine, but maybe we should get a second opinion on this

integreat_cms/cms/views/content_revision_view.py
```diff
@@ -0,0 +1,363 @@
...
+class ContentRevisionView(TemplateView):
+    """
+    View for browsing the page/poi revisions and restoring old page/poi revisions
+    """
+
+    #: The template to render (see :class:`~django.views.generic.base.TemplateResponseMixin`)
+    template_name = "content_revisions.html"
+    #: The context dict passed to the template (see :class:`~django.views.generic.base.ContextMixin`)
+
+    def get_context(self, request, *args, **kwargs):
```
timoludwig, line 30: No need to define a custom method here, there's already the inbuilt [`get_context_data()`](https://docs.djangoproject.com/en/4.2/topics/class-based-views/generic-display/#adding-extra-context)

integreat_cms/cms/views/content_revision_view.py
```diff
@@ -0,0 +1,363 @@
...
+        """
+        Returns the context of current content
+
+        To be implemented in the inheriting class
+        """
+        raise NotImplementedError
+
+    def can_change_content(self, request, content):
```
timoludwig, line 38: I like this idea, however there's already an inbuilt Mixin that handles permission in class-based views: [`PermissionRequiredMixin`](https://docs.djangoproject.com/en/4.2/topics/auth/default/#the-permissionrequiredmixin-mixin)

integreat_cms/cms/views/content_revision_view.py
```diff
@@ -0,0 +1,363 @@
...
+        """
+        content_type = None
+        content_name = ""
+
+        if kwargs.get("page_id"):
+            content_type = Page
+            content_name = "page"
+        elif kwargs.get("poi_id"):
+            content_type = POI
+            content_name = "poi"
```
timoludwig, line 181: I think it would make more sense to add a `model` attribute to the sub-classes and derive the content type and name from it...

integreat_cms/cms/views/content_revision_view.py
```diff
@@ -0,0 +1,363 @@
...
+
+        region = request.region
+        content = (
+            content_type.objects.filter(region=region)
+            .filter(id=kwargs.get(f"{content_name}_id"))
+            .first()
+        )
```
timoludwig, line 188: This query could be handled by the [`SingleObjectMixin`](https://docs.djangoproject.com/en/3.2/ref/class-based-views/mixins-single-object/#singleobjectmixin)

integreat_cms/cms/templates/content_revisions.html
```diff
@@ -6,24 +6,22 @@
...
-            {% with page_translations.first.title as page_title %}
-                {% blocktrans %}Page versions of "{{ page_title }}"{% endblocktrans %}
+            {% with content_translations.first.title as content_title %}
+                {% blocktrans %}"Versions of {{ content_title }}"{% endblocktrans %}
             {% endwith %}
         </h1>
-        <a href="{% url 'edit_page' page_id=page.id region_slug=request.region.slug language_slug=language.slug %}"
-           class="btn btn-outline">
-            <i icon-name="arrow-left-circle" class="align-top"></i> {% translate "Back to the page form" %}
+        <a href={{ form }} class="btn btn-outline">
```
timoludwig, line 13: Quotes are missing and `{{ form }}` usually holds an actual form object and not a URL, so I would rename this variable.

integreat_cms/cms/templates/content_revisions.html
```diff
@@ -6,24 +6,22 @@
...
+            <i icon-name="arrow-left-circle" class="align-top"></i> {% translate "Back to the form" %}
         </a>
     </div>
-    <form method="post"
-          action="{% url 'page_revisions' page_id=page.id region_slug=request.region.slug language_slug=language.slug %}">
+    <form method="post" action={{ revision }}>
```
timoludwig, line 17: Same here, the quotes are missing and `revision` is a bit ambiguous.

integreat_cms/cms/templates/content_revisions.html
```diff
@@ -6,24 +6,22 @@
...
         {% csrf_token %}
         <div class="w-3/5 m-auto mb-28 relative">
             <input class="relative z-10 pt-1"
                    type="range"
                    name="revision"
                    min="1"
-                   max="{{ page_translations.count }}"
+                   max="{{ content_translations.count }}"
```
timoludwig, line 24: While we're at it, we could also refactor the existing template and make it more performant - if we call `{{ content_translations.count }}` several times in a template, it will make a new sql query each time. It's better to force the evaluation into a python object and call `{{ content_translations|length }}` each time, which will do the computation in Python instead of SQL and does not retrieve the objects from the database again.

integreat_cms/cms/templates/content_revisions.html
```diff
@@ -108,10 +106,10 @@ <h1>{{ page_translation.title }}</h1>
             </div>
         </div>
         <div class="w-full p-4 flex justify-end gap-4 action-buttons">
-            {% if not page.archived %}
-                {% has_perm 'cms.change_page_object' request.user page as can_edit_page %}
-                {% has_perm 'cms.publish_page_object' request.user page as can_publish_page %}
-                {% if can_publish_page %}
+            {% if not content.archived %}
+                {% has_perm 'cms.change_{{ content_name }}_object' request.user content as can_edit %}
+                {% has_perm 'cms.publish_{{ content_name }}_object' request.user content as can_publish %}
```
timoludwig, line 111: The `cms.change_{{ content_name }}_object` permission only exists for pages, so we need to use a different syntax for all other content models.

integreat_cms/cms/views/content_revision_view.py
```diff
@@ -0,0 +1,363 @@
...
+                ).first(),
+                "selected_revision": selected_revision.first(),
+                "language": language,
+                "form": form,
+                "revision": revision,
+            },
+        )
+
+    # pylint: disable=too-many-branches
+    def post(self, request, *args, **kwargs):
```
timoludwig, line 155: There are quite a few duplications in `get()` and `post()`, whenever you notice this you can extract pieces of code into `dispatch()` which will be called in every case.

timoludwig (changes requested):
Thanks a lot, really cool! :rocket: 

I think generalizing this stuff is the right way to approach things, if this goes well we should also think about generalizing other views (e.g. side by side views), it really makes things easier when these things are at a single place and we don't have to edit the same thing for all content types.

Apart from my comments, I also took a more radical approach on refactoring this stuff, e.g. making more use of Django's inbuilt view mixins etc., so I also tried to make it compatible with the imprint version view as well...
Since my changes went a bit out of hand (as always), I'm now trying a different approach and I'm going to open another PR where you can have a look at the changes before merging them into your branch :sweat_smile:

timoludwig (approved):
Thanks a lot! :tada:

Issue #1690, opened 2022-09-16
Add API usage documentation


timoludwig:
### Motivation

At the moment, we only have the [API developer documentation](https://digitalfabrik.github.io/integreat-cms/ref/integreat_cms.api.html) and the [outdated API usage documentation from Wordpress](https://github.com/digitalfabrik/cms/wiki/REST-APIv3-Documentation).

### Proposed Solution

Add an updated version of the API usage documentation with all new fields etc.

### Additional Context

Also think about whether it is possible to use some kind of automated tool for this, so we don't have to update the documentation manually every time a field changes.

PeterNerlich:
Potentially interesting literature: https://www.django-rest-framework.org/topics/documenting-your-api/

timoludwig:
@PeterNerlich We'd have to rewrite our API using the django rest framework in order to be able to use its automatic documentation feature.
But yes, this could be less work than writing an automated generator for our custom API ourselves :sweat_smile:

david-venhoff:
I have experimented a bit with the django rest framework and I think that our current api cannot be converted to it well.
I uploaded a branch with some wip changes in case we decide to revisit it again at some point: [feature/django_rest_framework_api](https://github.com/digitalfabrik/integreat-cms/tree/feature/django_rest_framework_api)

PR #2272, opened 2023-05-11
Add API documentation

david-venhoff:
### Short description

This pr adds an api documentation.

### Proposed changes

- Add documentation to api/init.py

### Side effects

/

### Resolved issues

Fixes: #1690 

### DIFF
--- a/docs/src/api-docs.rst
+++ b/docs/src/api-docs.rst
@@ -1,0 +1,1039 @@
+*****************
+API Documentation
+*****************
+
+Sending the Development Header
+==============================
+
+To prevent tracking of API calls during tests and development, use the
+"X-Integreat-Development" header. Example:
+
+::
+
+   curl -H "X-Integreat-Development: true"  \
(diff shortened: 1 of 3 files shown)


svenseeberg:
Awesome, thanks! I'm wondering if we can establish a process that keeps the documentation and API in sync. I'm not sure how we can achieve this. Adding another check box to the pull request template is a bit mundane, right?

timoludwig (commented):
> I'm wondering if we can establish a process that keeps the documentation and API in sync. I'm not sure how we can achieve this. Adding another check box to the pull request template is a bit mundane, right? If nobody has a good idea, I would approve this PR after having validated the endpoints.

I guess we could write a CircleCI check that fails when files in the API source code are changed without adapting the documentation :smile:
And the check is only bypassed when `[no-doc-change]` is present in the commit summary :grin: 
But probably an idea for another issue...

timoludwig (changes requested):

timoludwig (approved):
:+1:

ulliholtgrave (approved):
Very nice! Thank you very much 😊

PR #2276, opened 2023-05-16
Fix remaining grayed button to outline button

JoeyStk:
### Short description
While reviewing, I noticed that I had missed out on the "Back to Form" button in the side-by-side view while I was working on  #2253. This PR fixes this.

### Proposed changes

- Changes "Back to form" button in side-by-side view from a grayed button (looks like it was disabled) to outline-button (introduced by PR  #2253)
- Remove unused / not rendered class btn-gray from "Order"-Section inside the collapsible box "Settings"

### Side effects

- I think none
- If you find the time, please double-check if you can find any grayed buttons that are not disabled. While going through the views and the code-base I didn't find any anymore, but maybe my keywords weren't great and I don't know all of the views
- I double-checked with UI/UX and they told me to use outline-buttons and not ghost-buttons for "Go back to..." buttons

### Resolved issues

Fixes: Leftover of #2253

### DIFF
--- a/integreat_cms/cms/templates/pages/page_form_sidebar/settings_box.html
+++ b/integreat_cms/cms/templates/pages/page_form_sidebar/settings_box.html
@@ -20,7 +20,7 @@
         {% render_field page_form|get_private_member:"ref_node_id" %}
         {% render_field page_form|get_private_member:"position" %}
         {% render_field page_form.parent id="parent" %}
-        <label class="btn-gray">{% translate "Order" %}</label>
+        <label>{% translate "Order" %}</label>
         <div id="page_order_table" class="mb-4">{% include "pages/_page_order_table.html" %}</div>
     </div>
     <div {% if not request.user.expert_mode %}hidden{% endif %}>
--- a/integreat_cms/cms/templates/pages/page_sbs.html
+++ b/integreat_cms/cms/templates/pages/page_sbs.html
@@ -14,7 +14,7 @@
             </h1>
             <div class="flex flex-wrap grow justify-between gap-2 mb-4">
                 <a href="{% url 'edit_page' page_id=source_page_translation.page.id region_slug=request.region.slug language_slug=target_language.slug %}"
-                   class="bg-gray-400 hover:bg-gray-500 text-center cursor-pointer text-white font-bold py-3 px-4 rounded">
+                   class="btn btn-outline">
                     <i icon-name="arrow-left-circle" class="align-top"></i> {% translate "Back to the page form" %}
                 </a>
                 <div class="flex flex-wrap gap-2">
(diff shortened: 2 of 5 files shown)


timoludwig (approved):
Thanks! :+1:

> If you find the time, please double-check if you can find any grayed buttons that are not disabled. While going through the views and the code-base I didn't find any anymore, but maybe my keywords weren't great and I don't know all of the views

I think we could argue that we don't need the class [btn-gray](https://github.com/digitalfabrik/integreat-cms/blob/c0bbd6c40708aac4695d458906876286a6ced4d5/integreat_cms/static/src/css/style.scss#L146-L151) anymore, since we only want to style buttons like this that are disabled (but then we don't need the additional class)...
There is only [one usage](https://github.com/digitalfabrik/integreat-cms/blob/c0bbd6c40708aac4695d458906876286a6ced4d5/integreat_cms/cms/templates/pages/page_form_sidebar/settings_box.html#L23) which can be removed I think...

integreat_cms/cms/templates/pages/page_sbs.html
```diff
@@ -14,7 +14,7 @@ <h1 class="heading">
             </h1>
             <div class="flex flex-wrap grow justify-between gap-2 mb-4">
                 <a href="{% url 'edit_page' page_id=source_page_translation.page.id region_slug=request.region.slug language_slug=target_language.slug %}"
-                   class="bg-gray-400 hover:bg-gray-500 text-center cursor-pointer text-white font-bold py-3 px-4 rounded">
+                   class="btn btn-outline hover:bg-gray-500 text-center cursor-pointer text-white font-bold py-3 px-4 rounded">
```
timoludwig, line 17:
```suggestion
                   class="btn btn-outline">
```
Ah, I think all those classes are not required anymore, since all the styling is already applied by the `btn` classes...

timoludwig (changes requested):

integreat_cms/cms/templates/pages/page_form_sidebar/settings_box.html
```diff
@@ -20,7 +20,7 @@
         {% render_field page_form|get_private_member:"ref_node_id" %}
         {% render_field page_form|get_private_member:"position" %}
         {% render_field page_form.parent id="parent" %}
-        <label class="btn-gray">{% translate "Order" %}</label>
```
timoludwig, line 23: Could you also delete the definition in the stylesheet? So we don't get tempted to use this again in the future :wink:

timoludwig (changes requested):

timoludwig (approved):

seluianova (changes requested):
I've found one gray button, it's [Back to the imprint form] in the side-by-side imprint view.
image

I would also discuss considering return buttons as ghost-buttons, because they don't trigger any content changes, what do you think?

timoludwig:
> I would also discuss considering return buttons as ghost-buttons, because they don't trigger any content changes, what do you think?

But our current guideline was that ghost buttons don't redirect the user to a different place, they just hide/show something dynamically...

seluianova (commented):
[Reset filter] is left (in Media Library) 

> But our current guideline was that ghost buttons don't redirect the user to a different place, they just hide/show something dynamically...

Okay, I'm not insisting here. Just a personal feeling.

seluianova (approved):
Thanks! Looks great to me ✅

timoludwig (commented):
I know, nit-picky, but could you also fix the typo in the commit summary? :grin:

Issue #1750, opened 2022-10-12
Improve loading times for sitemaps


timoludwig:
### Motivation

Our sitemaps are loading very slow and probably do a bunch of database queries which can be optimized.

### Proposed Solution

Investigate what is taking so long and try to reduce database queries similar to how we do it for the API - make use of prefetched translations and cache tree operations.

svenseeberg:
Loading times are indeed high. The cause could be a bug: https://integreat.app/lkalzey/de/sitemap.xml has 22MB after downloading.

The file starts with the following redundant entries:
```xml
<url><loc>https://integreat.app/lkalzey/de/ukraine/dringend-uebersetzer-fuer-ukrainisch-im-landkreis-alzey-worms-gesucht/</loc><lastmod>2023-01-31</lastmod><changefreq>monthly</changefreq><priority>1.0</priority><xhtml:link rel="alternate" hreflang="" href="https://integreat.app/lkalzey/en/ukraine/urgent-need-for-ukrainian-interpreters-in-the-alzey-worms-district/"/><xhtml:link rel="alternate" hreflang="" href="https://integreat.app/lkalzey/ru/украина/срочно-нужны-переводчики-украинског/"/><xhtml:link rel="alternate" hreflang="" href="https://integreat.app/lkalzey/uk/україна/терміново-шукаємо-усних-перекладачі/"/></url>

<url><loc>https://integreat.app/lkalzey/de/ukraine/dringend-uebersetzer-fuer-ukrainisch-im-landkreis-alzey-worms-gesucht/</loc><lastmod>2023-01-31</lastmod><changefreq>monthly</changefreq><priority>1.0</priority><xhtml:link rel="alternate" hreflang="" href="https://integreat.app/lkalzey/en/ukraine/urgent-need-for-ukrainian-interpreters-in-the-alzey-worms-district/"/><xhtml:link rel="alternate" hreflang="" href="https://integreat.app/lkalzey/ru/украина/срочно-нужны-переводчики-украинског/"/><xhtml:link rel="alternate" hreflang="" href="https://integreat.app/lkalzey/uk/україна/терміново-шукаємо-усних-перекладачі/"/></url>

[...]

PR #2282, opened 2023-05-16
Remove duplicate sitemap entries

svenseeberg:
### Short description
This improves the performance of the sitemap. This is quite relevant as some search engine crawlers frequently crash our server by inducing too much load.

Reference: https://integreat-test.tuerantuer.org/augsburg/de/sitemap.xml

- w/o patch: 24,4MB/2,03min
- w/ patch: 1kB/23ms

### Proposed changes
- Remove duplicate entries

### Side effects
- N/A

### Resolved issues
Fixes: #1750

### DIFF
--- a/integreat_cms/sitemap/sitemaps.py
+++ b/integreat_cms/sitemap/sitemaps.py
@@ -147,7 +147,7 @@ class PageSitemap(WebappSitemap):
         self.queryset = self.queryset.filter(
             page__in=self.region.get_pages(return_unrestricted_queryset=True),
             language=self.language,
-        )
+        ).distinct("page__pk")
 
 
 class EventSitemap(WebappSitemap):
@@ -182,7 +182,7 @@ class EventSitemap(WebappSitemap):
         # Filter queryset based on region and language
         self.queryset = self.queryset.filter(
             event__in=self.region.events.all(), language=self.language
-        )
+        ).distinct("event__pk")
 
 
 class POISitemap(WebappSitemap):
@@ -213,7 +213,7 @@ class POISitemap(WebappSitemap):
         # Filter queryset based on region and language
         self.queryset = self.queryset.filter(
             poi__in=self.region.pois.all(), language=self.language
-        )
+        ).distinct("poi__pk")
 
 
 class OfferSitemap(WebappSitemap):


timoludwig (approved):
Oh, good catch! :see_no_evil: 
Could you apply the same fix to events and pois as well?

MizukiTemma (approved):
Thank you 💪

svenseeberg:
I have no idea why the tests are failing right now o_O

integreat_cms/sitemap/sitemaps.py
```diff
@@ -144,9 +144,13 @@ def __init__(self, region, language):
...
-        self.queryset = self.queryset.filter(
-            page__in=self.region.get_pages(return_unrestricted_queryset=True),
-            language=self.language,
+        self.queryset = (
+            self.queryset.filter(
+                page__in=self.region.get_pages(return_unrestricted_queryset=True),
+                language=self.language,
+            )
+            .order_by("page", "-version")
+            .distinct("page")
```
timoludwig, line 153:
```suggestion
            .distinct("page_pk")
```
If you use `page_pk` as distinct value (like in the default ordering), you don't need to order first.

integreat_cms/sitemap/sitemaps.py
```diff
@@ -182,7 +186,7 @@ def __init__(self, region, language):
         # Filter queryset based on region and language
         self.queryset = self.queryset.filter(
             event__in=self.region.events.all(), language=self.language
-        )
+        ).order_by("-version")
```
timoludwig, line 189:
```suggestion
        ).distinct("event__pk")
```
This should be sufficient (same for pois)

timoludwig (changes requested):

integreat_cms/sitemap/sitemaps.py
svenseeberg: But that would list translations in ascending order? And don't we need to get the slug from the latest version?

timoludwig: No, translations are ordered by `["page__pk", "language__pk", "-version"]` by default.

Issue #2284, opened 2023-05-17
`TypeError` when non-staff user tries to grant page permissions


timoludwig:
### Describe the Bug

At the moment, an error occurs when a non-staff member tries to grant additional page permissions

### Steps to Reproduce

1. Login as "management"
2. Go to page form
3. Try to add the observer to the authors of the page

### Expected Behavior

The observer should be able to edit the page

### Actual Behavior

An error occurs:
```
May 17 10:25:10 ERROR integreat_cms.cms.views.pages.page_actions - argument of type 'ManyRelatedManager' is not iterable
Traceback (most recent call last):
  File "/home/integreat/integreat-cms/integreat_cms/cms/views/pages/page_actions.py", line 548, in grant_page_permission_ajax
    if page.region not in request.user.regions:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: argument of type 'ManyRelatedManager' is not iterable
```

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
May 17 10:25:10 DEBUG integreat_cms.cms.views.pages.page_actions - [AJAX] <SimpleLazyObject: <User (id: 2, username: management, role: Manager, region: augsburg)>> wants to grant <User (id: 10, username: observer, role: Observer, region: augsburg)> the permission to edit <Page (id: 21, region: augsburg, slug: alltag)>
May 17 10:25:10 ERROR integreat_cms.cms.views.pages.page_actions - argument of type 'ManyRelatedManager' is not iterable
Traceback (most recent call last):
  File "/home/integreat/integreat-cms/integreat_cms/cms/views/pages/page_actions.py", line 548, in grant_page_permission_ajax
    if page.region not in request.user.regions:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: argument of type 'ManyRelatedManager' is not iterable
May 17 10:25:10 DEBUG integreat_cms.cms.views.pages.page_actions - Ein Fehler ist aufgetreten. Bitte kontaktieren Sie eine:n Administrator:in.
  ```

</details>

svenseeberg:
Good catch :+1: I only grepped for the region name and did therefore not stumble over this error message. Can/Do we want to add the request URL to all logged errors? :D

timoludwig:
> Can/Do we want to add the request URL to all logged errors? :D

[...]

PR #2285, opened 2023-05-17
Fix error when granting additional page permissions

timoludwig:
### Short description

Fix error when granting additional page permissions

### Proposed changes

- Replace `user.regions` by `user.regions.all()`
- Add tests for page permissions

### Resolved issues

Fixes: #2284

### DIFF
--- a/integreat_cms/cms/views/dashboard/region_selection.py
+++ b/integreat_cms/cms/views/dashboard/region_selection.py
@@ -38,7 +38,6 @@ class RegionSelection(TemplateView):
         """
 
         user = request.user
-        regions = user.regions
         if user.is_superuser or user.is_staff:
             return redirect("admin_dashboard")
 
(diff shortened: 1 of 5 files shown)


svenseeberg (approved):
Thanks, looks good to me. AFAICT the tests do not cover the previous problem, but that is intended?

timoludwig:
> Thanks, looks good to me. AFAICT the tests do not cover the previous problem, but that is intended?

Yes they do, since previously these tests would fail for the MANAGEMENT role? :thinking:

svenseeberg:
> Yes they do, since previously these tests would fail for the MANAGEMENT role? thinking

Maybe something is wrong in my setup. When I remove the `.all()` the tests show no problem.

timoludwig:
> Maybe something is wrong in my setup. When I remove the `.all()` the tests show no problem.

Ahh sorry you're right, because the views still return `200` and the error is just logged...
Damn, yeah then it's a bit more complicated to test this and requires some refactoring... Probably I will do this later so we can merge this PR (now?)

timoludwig:
the broad except is the real problem here^^

svenseeberg:
Thanks

Issue #2289, opened 2023-05-17
Add single region endpoint


timoludwig:
### Motivation

The web app currently requests all regions on each request, which adds a lot of unnecessary overhead.

### Proposed Solution

Add the possibility to get a specific region by slug.
Also include the languages of that region in the request to save the subsequent request as well.

steffenkleinle:
Please let the app team know once this is done: https://issues.tuerantuer.org/browse/IGAPP-1267

PR #2291, opened 2023-05-17
Add single region endpoint

timoludwig:
### Short description

Add single region endpoint

### Proposed changes

- Deprecate `/live/` and `/hidden/` filters of region endpoint
- Add languages to the region API response
- Allow to request single region by region slug

### Side effects

- Regions can no longer be filtered by their status (but the app requests all regions anyway and does the filtering for themself)
- Languages of archived regions can no longer be retrieved (I guess this was a bug?)

### Resolved issues

Fixes: #2289


steffenkleinle (approved):
Thanks :pray:

PR #2297, opened 2023-05-22
test.sh: Allow specific and multiple tests

PeterNerlich:
### Short description
This PR adds comfort feature for development in making it easier to run one or a set of specific tests without running other modules or even undesired tests within the same module.
It is essentially an extension of the existing specify-a-single-tests-module feature, adding support for the pytest `::` syntax and passing through markers and keyword expressions. (https://docs.pytest.org/en/latest/how-to/usage.html#specifying-which-tests-to-run)

*Example: Selecting the one test that still fails for a specific role, to quickly test the difference of small changes without waiting for the other integration tests in the same test module:*
`tools/test.sh -vv tests/summ_ai_api/summ_ai_test.py::test_summ_ai_error_handling[OBSERVER-pyloop]`

### Proposed changes

- instead of saving arguments not matching any specific pattern into `TEST_PATH` (overriding potential previous contents), add them to the `TESTS` array
- loop over the array and strip everything from the first `::` before checking whether a file with that name exists
- look for `-m` and `-k` arguments and pass them to pytest as well
- display the selection sanely: keyword expressions are enclosed in quotes (`""`), tags are prefixed using `with `, and all individual criteria are joined with `, `.

### Side effects

None come to mind, but I don't trust myself to not forget something here

### DIFF
--- a/tools/_functions.sh
+++ b/tools/_functions.sh
@@ -434,3 +434,12 @@ function format_grep_output {
             -e "s/^([0-9]+)-(.*)/\1\2/"                     `# Remove dash of unmatched line`
     done
 }
+
+# Use (multi-char) delimiter to join strings
+# Taken from https://stackoverflow.com/a/17841619
+function join_by {
+    local d=${1-} f=${2-}
+    if shift 2; then
+        printf %s "$f" "${@/#/$d}"
+    fi
+}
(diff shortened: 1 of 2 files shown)


MizukiTemma (approved):
Thank you👍LGTM
This makes test procedure easier 😃

david-venhoff (approved):
Thanks!

PR #2299, opened 2023-05-23
Change verbose name of news in network settings

JoeyStk:
### Short description
This PR changes the verbose name of news in the network settings from TüNews to a more generic name that fits more than one provider 

### Proposed changes

- Change verbose name of news
- Add new migration file
- Add release note

### Side effects

- I think none

### Resolved issues

Fixes: #2298 


integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -3603,12 +3603,16 @@ msgid "Please check the box if you want to use short urls."
...
-msgid "Enable tunews"
-msgstr "tünews aktivieren"
+msgid "Enable external news"
+msgstr "Externe News aktivieren"
 
 #: cms/models/regions/region.py
-msgid "Enable to show a feed of tunews articles to users."
-msgstr "Auswählen, um anderen Benutzer:innen einen tünews-Feed anzuzeigen."
+msgid ""
+"Enable to display national news from external providers in addition to local "
```
ulliholtgrave, line 3611: I would rephrase that a little bit as we are not sure if we stick with national news and i would differentiate a little bit between push notifications and external news.
```suggestion
"Enable to display external articles in addition to local news managed by the CMS "
```

integreat_cms/cms/migrations/0071_update_external_news.py
```diff
@@ -0,0 +1,29 @@
...
+        migrations.RemoveField(
+            model_name="region",
+            name="tunews_enabled",
+        ),
+        migrations.AddField(
+            model_name="region",
+            name="external_news_enabled",
+            field=models.BooleanField(
+                default=False,
+                help_text="Enable to display national news from external providers in addition to local news",
```
ulliholtgrave, line 25:
```suggestion
                help_text="Enable to display external articles in addition to local news managed by the CMS",
```

integreat_cms/cms/models/regions/region.py
```diff
@@ -315,10 +315,12 @@ class Region(AbstractBaseModel):
...
     )
 
-    tunews_enabled = models.BooleanField(
+    external_news_enabled = models.BooleanField(
         default=False,
-        verbose_name=_("Enable tunews"),
-        help_text=_("Enable to show a feed of tunews articles to users."),
+        verbose_name=_("Enable external news"),
+        help_text=_(
+            "Enable to display national news from external providers in addition to local news"
```
ulliholtgrave, line 322:
```suggestion
            "Enable to display external articles in addition to local news managed by the CMS"
```

ulliholtgrave (changes requested):
Thanks for your PR :+1:  In general this looks alright to me, but I have some small suggestions for the help text. 

:warning:  As I didn't create the PO and migration file by myself their might be some conflicts with my suggestion if they diverge a little bit

integreat_cms/api/v3/regions.py
```diff
@@ -34,7 +34,8 @@ def transform_region(region):
         "latitude": region.latitude,
         "bounding_box": region.bounding_box.api_representation,
         "aliases": region.aliases,
-        "tunews": region.tunews_enabled,
+        "tunews": region.external_news_enabled,
```
JoeyStk, line 37: > I think it would make sense to rename the database field as well, and only keep the old name in the API for legacy reasons.

I changed the database name now and kept this entry in the API. But I switched it to the new field, since the old field was removed. Is that correct or should I revert the migration partly to restore the tunews_enabled, so nothing is changed?

integreat_cms/cms/models/regions/region.py
JoeyStk: Thank you for your suggestions. The changes should be applied now :)

integreat_cms/cms/migrations/0071_update_external_news.py
```diff
@@ -0,0 +1,29 @@
...
+        ),
+        migrations.AddField(
+            model_name="region",
+            name="external_news_enabled",
+            field=models.BooleanField(
+                default=False,
+                help_text="Enable to display external articles in addition to local news managed by the CMS",
+                verbose_name="Enable external news",
+            ),
+        ),
```
timoludwig, line 28: I'm pretty sure I've explained this a few times before: Removing a field and then adding a new one will cause the saved data to be lost.
[`RenameField`](https://docs.djangoproject.com/en/3.2/ref/migration-operations/#renamefield) is what you should use instead in these cases to preserve the data.
Autogenerating migration files saves a lot of time in most cases, but it doesn't mean we don't have to check whether it does what we want it to do.

tests/api/expected-outputs/regions.json
```diff
@@ -24,6 +24,7 @@
       "Schmiechen": { "latitude": 48.212774, "longitude": 10.965841 }
     },
     "tunews": true,
+    "external_news": false,
```
timoludwig, line 27:
```suggestion
    "external_news": true,
```

tests/api/expected-outputs/regions.json
```diff
@@ -61,8 +63,9 @@
       [5.98865807458, 47.3024876979],
       [15.0169958839, 54.983104153]
     ],
-    "aliases": { "Nuernberg": { "latitude": 49.450938, "longitude": 11.075269 } },
+    "aliases": { "Nuernberg": { "longitude": 11.075269, "latitude": 49.450938 } },
     "tunews": false,
+    "external_news": false,
```
timoludwig, line 68:
```suggestion
    "external_news": true,
```

tests/api/expected-outputs/regions_live.json
timoludwig, line 1: This files has been deleted on the develop branch, please don't re-add it in this PR. Sometimes, resolving a git conflict means to delete a file that's present on your branch but not anymore on the base branch.

timoludwig (changes requested):
Thanks for your changes! :+1:

Could you also rename the field in the test data to fix the tests, adjust the formatting? Thanks!

integreat_cms/api/v3/regions.py
timoludwig: Ah, and one thing I forgot: Since we have [API documentation](https://github.com/digitalfabrik/integreat-cms/blob/develop/docs/src/api-docs.rst) now, could you document the new field?

ulliholtgrave (approved):
I guess I fixed the issues with the tests now. @timoludwig Can you also take a last look on it? :)

timoludwig (approved):
Thanks @ulliholtgrave! :muscle:

Issue #2224, opened 2023-04-12
HIX value differs when calculated via tinymce vs pre save signal


timobrembeck:
### Describe the Bug

The HIX value differs depending on how it's calculated.

### Steps to Reproduce

1. Go to a page and calculate the HIX value via the "update" button in the hix widget
2. Save page
3. Inspect the object in the database or via Django admin: http://localhost:8000/admin/cms/pagetranslation/
4. See different value

### Expected Behavior

The value should be identical

### Actual Behavior

The value differs

### Additional Information

I assume this is due to the different way line breaks are handled in tinymce vs the database.
TinyMCE just uses `\n` as line break whereas the content in the database is `\r\n`.

charludo:
Could you name a page where this is happening? I'm having trouble reproducing this

timoludwig:
I noticed this with e.g. "Ausländerbehörde" in the test data...

I noticed that the content in the database (which is sent in thee pre-save signal) contains unicode (`&#228;`) characters and lines breaks in the form of `\r\n` (score: **17.46**):
```
[...]

PR #2302, opened 2023-05-24
Fix HIX value difference

MizukiTemma:
### Short description

This PR tries to keep the HIX value identical between pre save signal and tinymce.

### Proposed changes

- replace `\n` with `\r\n` before checking HIX value.
- replace entity reference with numeric reference (example: `&auml;` -> `&#228;`) before checking HIX value.

### Side effects

- Sadly I couldn't find a way to convert entity reference to numeric reference directly and needed some converting steps.
- It seems, so far observed in the test data, reference art of special letters does not cause difference in HIX value, but the style of new line does.
-  If "\n" appears in the text by chance, for example when an user writes "Please use \n as line break" in the editor, simple replacement might replace it too, even though it is not intended as a line break. I testet it and "\n" tipped in the editor was not replaced, but I cannot garant it...

### Resolved issues

Fixes: #2224 

### DIFF
--- a/integreat_cms/cms/views/utils/hix.py
+++ b/integreat_cms/cms/views/utils/hix.py
@@ -2,6 +2,7 @@
 This file contains functionality to communicate with the Textlab api to get the hix-value
 for a given text.
 """
+
 import json
 import logging
 from functools import lru_cache
@@ -10,6 +11,8 @@ from urllib.error import URLError
 from django.conf import settings
 from django.http import JsonResponse
 from django.views.decorators.http import require_POST
+from lxml.etree import LxmlError
+from lxml.html import fromstring, tostring
 
 from ....api.decorators import json_response
 from ....textlab_api.textlab_api_client import TextlabClient
@@ -59,12 +62,20 @@ def get_hix_score(request, region_slug):
     # Don't pass texts larger than 100kb to the api in order to avoid being vulnerable to dos attacks
     if len(request.body) > MAX_TEXT_LENGTH:
         return JsonResponse({"error": "Request too large"})
+
     body = json.loads(request.body.decode("utf-8"))
     text = body["text"]
 
     if not isinstance(text, str) or not text.strip():
         logger.warning("Received invalid text: %r", text)
         return JsonResponse({"error": f"Invalid text: '{text}'"})
+
+    # Normalize text to get same HIX result as with form submission
+    text = text.replace("\n", "\r\n")
+    try:
+        text = tostring(fromstring(text)).decode("utf-8")
+    except LxmlError:
+        pass
 
     if score := lookup_hix_score(text):
         return JsonResponse({"score": score})
--- a/integreat_cms/release_notes/current/unreleased/2224.yml
+++ b/integreat_cms/release_notes/current/unreleased/2224.yml
@@ -1,0 +1,2 @@
+en: Fix HIX value difference
+de: Behebe Differenz in HIX Wert


integreat_cms/cms/views/utils/hix.py
```diff
@@ -59,13 +60,21 @@ def get_hix_score(request, region_slug):
...
     body = json.loads(request.body.decode("utf-8"))
     text = body["text"]
 
     if not isinstance(text, str) or not text.strip():
         logger.warning("Received invalid text: %r", text)
         return JsonResponse({"error": f"Invalid text: '{text}'"})
 
-    if score := lookup_hix_score(text):
+    # convert special characters into plain letters
+    unescaped = html.unescape(text.replace("\n", "\r\n"))
```
timoludwig, line 72: > f "\n" appears in the text by chance, for example when an user writes "Please use \n as line break" in the editor, simple replacement might replace it too, even though it is not intended as a line break. I testet it and "\n" tipped in the editor was not replaced, but I cannot garant it...

This is totally fine, since a newline `\n` is one single special character, whereas when "\n" is typed in the content, it is considered two separate characters `\` and `n`, which is escaped in the content to `\\n` and not replaced with this statement :+1:

integreat_cms/cms/views/utils/hix.py
```diff
@@ -59,13 +60,21 @@ def get_hix_score(request, region_slug):
...
+    # convert special characters to numeric reference
+    to_numeric_reference = unescaped.encode("ascii", "xmlcharrefreplace")
+    # convert from byte to string
+    to_string = to_numeric_reference.decode("utf-8", "xmlcharrefreplace")
```
timoludwig, line 76: > Sadly I couldn't find a way to convert entity reference to numeric reference directly and needed some converting steps.

I think the different encoding is a result from our [custom clean method](https://github.com/digitalfabrik/integreat-cms/blob/0dc0592a301cf233300844a0a65e204e87f947da/integreat_cms/cms/forms/custom_content_model_form.py#L103), so you could also do:
```
from lxml.etree import LxmlError
from lxml.html import fromstring, tostring

try:
    text = tostring(fromstring(text)).decode("utf-8")
except LxmlError:
    pass
```
but that also isn't that much shorter...

timoludwig (approved):
Awesome, thanks a lot! :pray:

integreat_cms/cms/views/utils/hix.py
MizukiTemma: Applied 🙂

integreat_cms/cms/views/utils/hix.py
```diff
@@ -59,13 +62,21 @@ def get_hix_score(request, region_slug):
...
         return JsonResponse({"error": "Request too large"})
+
     body = json.loads(request.body.decode("utf-8"))
     text = body["text"]
 
     if not isinstance(text, str) or not text.strip():
         logger.warning("Received invalid text: %r", text)
         return JsonResponse({"error": f"Invalid text: '{text}'"})
 
+    # Replace line break \n with \r\n
```
timoludwig, line 73:
```suggestion
    # Normalize text to get same HIX result as with form submission
```
I think it's pretty obvious that the following line replaces `\n` with `\r\n`, so no need to duplicate this information in a comment.
We could however explain why we're doing it.

timoludwig (approved):

JoeyStk (approved):
Thank you for this important fix. I wasn't able to break it :+1:

Issue #2307, opened 2023-05-25
Remove link text field from link replacement form


timoludwig:
### Motivation

Initially, I thought having a disabled link text field would make sense to indicate that link texts are not replaced and to do some preliminary work for #2226.

However, introducing this had several side effects:
- The existence of the field led to confusion
- The replacement form required scrolling on smaller screens
- It was no longer possible to replace the `src` attributes of `img` tags

### Proposed Solution

Remove the field completely and return to the previous replacement form

PR #2316, opened 2023-06-12
Remove link text field from link replacement form

seluianova:
### Short description
Disabled link text field in the link replacement form caused several issues:
- The existence of the field led to confusion
- The replacement form required scrolling on smaller screens
- It was no longer possible to replace the src attributes of img tags

### Proposed changes
Revert the previous link replacement form (without the disabled link text field)

### Side effects
No?

However I am not quite sure what was meant by _It was no longer possible to replace the src attributes of img tags_ because I think I can replace src value for images on develop.

### Resolved issues
Fixes: #2307

### DIFF
--- a/integreat_cms/cms/templates/linkcheck/link_list_row.html
+++ b/integreat_cms/cms/templates/linkcheck/link_list_row.html
@@ -101,23 +101,11 @@
     </tr>
     {% if url.id == view.kwargs.url_id %}
         <tr>
-            <td colspan="{% if request.user.expert_mode %}6{% else %}5{% endif %}">
-                <div class="px-2 pb-2">
+            <td colspan="10">
+                <div class="flex gap-2 p-2">
                     {% render_field edit_url_form.url|add_error_class:"border-red-500" type="url" form="edit-url-form" %}
-                </div>
-            </td>
-            <td colspan="3"
-                title="{% translate "The link text cannot be replaced centrally, because it may differ in the other source content." %}">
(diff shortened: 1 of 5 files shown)


MizukiTemma (approved):
LGTM 👍 Thank you!

MizukiTemma:
Oh, we should probably add a release note, since this change is visible for users 👀

integreat_cms/cms/templates/linkcheck/link_list_row.html
```diff
@@ -5,122 +5,108 @@
 {% load widget_tweaks %}
 {% load linkcheck_filters %}
 {% object_translation_has_view_perm request.user url.region_links.0.content_object as show_source_link %}
-<tbody class="hover:bg-gray-100">
```
timoludwig, line 8: I liked the previous solution with the `tbody` because it caused the area of the form to also change the hovered background when the `tr` is hovered, which made it more obvious that the two rows actually belong together...

timoludwig (changes requested):
Thanks a lot! :+1:

In general it looks good, I'd just suggest to revert the previous `tbody` element and I agree that we could add a release note.

integreat_cms/cms/templates/linkcheck/link_list_row.html
seluianova: True. Please check now

timoludwig (approved):
Thanks! :tada:

Issue #2317, opened 2023-06-13
Empty pages count as "not ready for MT" in dashboard


timoludwig:
### Describe the Bug

The value of "ready for MT" in the dashboard calculates the percentage of all pages with HIX above the threshold in relation to all translations.
This is incorrect since e.g. empty pages without a HIX score are allowed for MT (but do not increase the percentage in the dashboard).

### Steps to Reproduce

1. Go to a region dashboard
2. See percentage of pages that are ready for machine translations
3. Check that the percentage gets lower for every empty page

### Expected Behavior

An empty page should be considered ready for MT by default

### Actual Behavior

An empty page is calculated as "not ready for MT" in the overall percentage counter in the dashboard

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

seluianova:
In addition to empty pages, we also have pages that are not empty but do not have HIX. For example, if TextLab API returned an error.
Should such pages be considered as ready for MT or not ready for MT?

osmers:
Maybe it would already help if we closed this issue: https://github.com/digitalfabrik/integreat-cms/issues/2296
I just had the case today that basically all pages have a HIX above 15 but the percentage says only 85% are ready for translation: https://admin.integreat-app.de/odenwaldkreis/ check the dashboard here

Maybe we can even separate it further and say: 
XX of YY pages are ready for translation. ZZ pages have no HIX value bcs they have no content. They can be translated. ÄÄ pages have a hidden HIX value and cannot be translated. 

XX = pages with HIX above 15
YY = pages with a HIX value
ZZ = pages with only a title
ÄÄ = pages with hidden HIX value

seluianova:
> Maybe it would already help if we closed this issue: https://github.com/digitalfabrik/integreat-cms/issues/2296

@osmers yes, that's basically what we are doing.

But we use a slightly different calculation.
The message is just:
`XX out of YY pages are ready for automatic translation.`
Where:
YY = all pages with HIX enabled (including empty pages and pages with HIX errors)
XX = YY - (pages with HIX below 15)

There was a discussion about it in PR #2320, you can take a look.

Issue #2310, opened 2023-05-30
`ZeroDivisionError` in region dash board


MizukiTemma:
### Describe the Bug

`ZeroDivisionError` occurs when there is no page but HIX is activated in the region.

### Steps to Reproduce

1. Go to a region
2. Archive all the pages
3. Activate HIX for the region
4. Go to the region dash board
5. see the error

### Expected Behavior

Dashboard will be shown.

### Actual Behavior

`ZeroDivisionError`. Users cannot see the dashboard

### Additional Information

<details>
  <summary>Traceback</summary>

   File "/home/user/integreat-cms/integreat_cms/cms/views/dashboard/dashboard_view.py", line 85, in get_hix_context
    ready_for_mt = math.trunc(100 * ready_for_mt_count / len(hix_translations))

Exception Type: ZeroDivisionError at /nurnberg/
Exception Value: division by zero

</details>

MizukiTemma:
This is related to #2296

PR #2320, opened 2023-06-13
Count empty translations as ready for MT in the dashboard

seluianova:
### Short description
The value of "ready for MT" in the dashboard calculates the percentage of all pages with HIX above the threshold in relation to all translations.
This is incorrect since e.g. empty pages without a HIX score are allowed for MT

### Proposed changes
Add the number of empty translations to the ready_for_mt_count value

### Side effects
No?

### Resolved issues
Fixes: #2317
Fixes: #2296
Fixes: #2310 (self-resolved as a side effect)

### DIFF
--- a/integreat_cms/cms/views/dashboard/dashboard_view.py
+++ b/integreat_cms/cms/views/dashboard/dashboard_view.py
@@ -1,5 +1,4 @@
 import logging
-import math
 
 from django.conf import settings
 from django.utils import translation
@@ -65,15 +64,11 @@ class DashboardView(TemplateView, ChatContextMixin):
         hix_pages = region.get_pages(return_unrestricted_queryset=True).filter(
             hix_ignore=False
         )
+
         # Get the latest versions of the page translations for these pages
         hix_translations = PageTranslation.objects.filter(
             language__slug__in=settings.TEXTLAB_API_LANGUAGES, page__in=hix_pages
         ).distinct("page_id", "language_id")
-
-        # Get all translations with empty content
-        hix_translations_with_empty_content = [
-            pt for pt in hix_translations if not pt.content.strip()
-        ]
 
         # Get all hix translations where the score is set
         hix_translations_with_score = [pt for pt in hix_translations if pt.hix_score]
@@ -83,16 +78,15 @@ class DashboardView(TemplateView, ChatContextMixin):
             hix_translations_with_score, key=lambda pt: pt.hix_score
         )
 
-        # Get the number of translations which are ready for MT
-        ready_for_mt_count = sum(
-            pt.hix_score >= settings.HIX_REQUIRED_FOR_MT
+        # Get the number of translations which are not ready for MT
+        not_ready_for_mt_count = sum(
+            pt.hix_score < settings.HIX_REQUIRED_FOR_MT
             for pt in hix_translations_with_score
-        ) + len(hix_translations_with_empty_content)
-
-        ready_for_mt = math.trunc(100 * ready_for_mt_count / len(hix_translations))
+        )
 
         return {
             "worst_hix_translations": worst_hix_translations,
             "hix_threshold": settings.HIX_REQUIRED_FOR_MT,
-            "ready_for_mt": ready_for_mt,
+            "ready_for_mt_count": len(hix_translations) - not_ready_for_mt_count,
+            "total_count": len(hix_translations),
         }
--- a/integreat_cms/cms/views/dashboard/dashboard_view.py
+++ b/integreat_cms/cms/views/dashboard/dashboard_view.py
@@ -70,6 +70,11 @@ class DashboardView(TemplateView, ChatContextMixin):
             language__slug__in=settings.TEXTLAB_API_LANGUAGES, page__in=hix_pages
         ).distinct("page_id", "language_id")
 
+        # Get all translations with empty content
+        hix_translations_with_empty_content = [
+            pt for pt in hix_translations if not pt.content.strip()
+        ]
+
         # Get all hix translations where the score is set
         hix_translations_with_score = [pt for pt in hix_translations if pt.hix_score]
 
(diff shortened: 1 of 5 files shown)


JoeyStk (approved):
Thank you a lot for the quick fix! :muscle: 

You could think about also including #2296 in this PR. In that issue we are asked to change the percentage (58% pages) to actual numbers (10 out of 17 pages). In combination with this PR the service-team asked us to write something like "10 out of 17 pages are ready for mt, 3 of them are empty pages and therefore not listed below" or the other way around "7 out of 17 pages are ready for mt, 3 more pages are empty and therefore not listed below". The service-team left it to us to decide which one we like better. If you think this would exceed the scope of this PR we can do it in a separate PR

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -70,18 +70,25 @@ def get_hix_context(self):
             language__slug__in=settings.TEXTLAB_API_LANGUAGES, page__in=hix_pages
         ).distinct("page_id", "language_id")
 
+        # Get all translations with empty content
+        hix_translations_with_empty_content = [
+            pt for pt in hix_translations if not pt.content
```
timoludwig, line 75:
```suggestion
            pt for pt in hix_translations if not pt.content.strip()
```
In the HIX signals, we also count a bunch of whitespaces as "empty"

timoludwig (changes requested):
Thanks! :+1:

seluianova:
Added changes for #2296.
Currently the widget looks like this:
image
I chose this sentence because in addition to empty pages, we may have pages with no HIX score, for example, if the TextLab API returned an error.
`For 16 pages the HIX value is not calculated.` - this number includes empty pages & non-empty pages without HIX (but with HIX enabled).
But let me know if that's not what's needed.

I also replaced the release notes because changes #2296 completely overwrote #2317.

timoludwig (changes requested):
> I chose this sentence because in addition to empty pages, we may have pages with no HIX score, for example, if the TextLab API returned an error. `For 16 pages the HIX value is not calculated.` - so this number includes empty pages & non-empty pages without HIX (but with HIX enabled). But let me know if that's not what's needed.

Hmm, yeah definitely a good question how to handle this case... :thinking: 
In my opinion, we don't need to tell the users that some values are missing... In the end, we just allow MTs for any page translation where we're in doubt (because the textlab api returned an error), so from the user's perspective, I think they should just count as valid...

Maybe we could add a management command to retrieve all pages where the HIX value could not be retrieved, so we can investigate the issue and find out why the textlab api calls are failing... but that's probably out of scope for this issue...

seluianova:
> from the user's perspective, I think they should just count as valid...

So if I take the message below, do you think the number of such pages should be included into YY?

`XX out of YY pages are ready for automatic translation. ZZ of them are empty and therefore not listed below.`

For example, we have:

1 page with HIX >= 15
1 page with HIX < 15
1 page with no HIX

The message will say:
"1 out of 3 pages are ready for automatic translation", but the table will contain 2 rows.

I'm worried that it will look like a bug because not obvious why "out of 3".

timoludwig:
> @timoludwig so, having the message below, do you think the number of such pages should be included into YY?
> 
> `XX out of YY pages are ready for automatic translation. ZZ of them are empty and therefore not listed below.`

In my opinion, the "ZZ of them are empty ..." is already too much information. I think a simple "XX out of YY pages are ready ..." is enough for the users to estimate whether they need to seriously rework their content or whether it's ok like this.

So my suggestion for the calculation would be:
YY = "page translations which have HIX enabled" (including empty pages or pages with HIX errors)
XX = YY - "page translations with HIX < 15"

So from the user's perspective, HIX errors don't matter because they don't prevent MT. HIX errors are just interesting for the development team to investigate the issue, but this information is not needed on the dashboard.

> For example, we have:
> 
> 1 page with HIX >= 15 1 page with HIX < 15 1 page with no HIX (but not empty)
> 
> The message will say: "1 out of 3 pages are ready for automatic translation", but the table will contain 2 rows.
> 
> I'm worried that it will look like a bug because not obvious why "out of 3".

I think this is a very rare edge case, in most cases the regions will have so many pages that they cannot realistically count the numbers of rows, so I doubt that they will be confused if the numbers don't exactly match.

> **UPD**. As one more option, probably we could just display such pages in the table? With the dash or N/A as HIX VALUE. In addition it will help to find such pages quickly, and update the value... I mean, a management command would help here, but if we display all the lines, the numbers will at least match.

In my opinion, this would just cause more confusion because they do not understand why these pages do not have a value... Since we treat such pages as "ready for MT", we could only show them with an artificial score of 20 or whatever... but I guess not listing them there is also fine.

> **UPD-2**. Or another option is to simply exclude them along with those who have HIX disabled.

I think this does not make sense, since HIX disabled means they cannot be automatically translated and HIX missing means they can.

seluianova:
Updated message to: 

`XX out of YY pages are ready for automatic translation.`

> YY = "page translations which have HIX enabled" (including empty pages or pages with HIX errors)
> XX = YY - "page translations with HIX < 15"

osmers:
So how would the numbers behave if all pages have a value above 15? They XX and YY should match, right?
Bcs I received the question yesterday from a municipality: Their number said 85% but the list did not show any pages they would have to work on. And I think that is very confusing. Then it is better that they get the information that ZZ pages don't have content and therefore no HIX but are translateable.

seluianova:
> So how would the numbers behave if all pages have a value above 15? They XX and YY should match, right?

@osmers right.

osmers:
> > So how would the numbers behave if all pages have a value above 15? They XX and YY should match, right?
> 
> @osmers right.

Sorry to ask again - but to me, that does not seem to be the case if YY includes empty pages. Unless we count XX as empty pages + pages with HIX > 15

I think my point is that is important that if all pages are ready for translation XX and YY have to match to avoid confusion

seluianova:
> Unless we count XX as empty pages + pages with HIX > 15

@osmers Yes, that's what we do. We count XX as all pages ready for MT: with HIX >= 15 or empty.

> I think my point is that is important that if all pages are ready for translation XX and YY have to match to avoid confusion

Yes, XX and YY will match, if all pages are ready for MT.

Examples:
If we have 3 pages, all have HIX >= 15 —> `3 out of 3 pages are ready for automatic translation.`
If we have 3 pages: 1 with HIX >= 15, 1 with HIX < 15, 1 empty —> `2 out of 3 pages are ready for automatic translation.`
If we have 3 pages, 1 empty, 2 with HIX < 15 —> `1 out of 3 pages are ready for automatic translation.`

osmers:
@seluianova perfect - thank you for clarifying :)

timoludwig (approved):
Awesome, thanks! :tada:

Issue #2318, opened 2023-06-13
Incorrect translation status after machine translation (XLIFF-exported pages)


MizukiTemma:
### Describe the Bug

Translation status of "currently in translation" pages does not transit correctly after machine translation.

### Steps to Reproduce

1. Go to a page.
2. Export the page for translation in a language for which machine translation is allowed
3. Machine translate it by bulk action.
4. See the status remains "currently in translation" with clock icon.

### Expected Behavior

After machine translation the status changes to "machine translation" (and the robot icon).

### Actual Behavior

Aftter machne translation the status remains "currently in translation" (with the clock icon).

### Additional Information

Related issue #2319

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

MizukiTemma:
It seems the problem is more generall: when Redis is activated, the icon doesn not transit qucikly (sometimes transits after a while/some other operations, sometimes remains unchanged) although the change is actually saved in the database.

timobrembeck:
> It seems the problem is more generall: when Redis is activated, the icon doesn not transit qucikly (sometimes transits after a while/some other operations, sometimes remains unchanged) although the change is actually saved in the database.

Thanks for investigating this issue. In this case, I'd say we re-open it and make sure to invalidate the cache correctly.

PR #2214, opened 2023-04-06
Fix translation state for minor edit currently in translation

timoludwig:
### Short description

Show translation status correctly if the latest version is a minor edit and is currently being translated

### Proposed changes

- Set the `currently_in_translation` flag for all versions, not only the latest version

### Resolved issues

Fixes: #2191

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #2323, opened 2023-06-20
Fix translation status of currently in translation pages after MT

MizukiTemma:
### Short description

This PR fixes the translation status change of `currently in translation` pages after they are machine translated.

### Proposed changes

- Count machine translation as cancellation of translation (XLIFF) process.
- Set "currently in translation" to false in the form.
- Revert the value of `currently in translation` in all the versions to false after machine translation (see #2214).

### Side effects

- I'm very afraid that this change affects other constraints/rules of translation status and its calculation, though so far it looks good to me 🙈 

### Note
This needs to be done before the release of #2221.

### Resolved issues

Fixes: #2318 

### DIFF
--- a/integreat_cms/deepl_api/deepl_api_client.py
+++ b/integreat_cms/deepl_api/deepl_api_client.py
@@ -165,6 +165,7 @@ class DeepLApiClient(MachineTranslationApiClient):
                     if existing_target_translation
                     else source_translation.status,
                     "machine_translated": True,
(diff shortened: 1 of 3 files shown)


ulliholtgrave (approved):
Thanks! That seems to fix the problem with the status :+1:

JoeyStk (approved):
Thank you! Works as expected.
I wasn't able to find unwanted side effects too :)

Issue #2273, opened 2023-05-11
`NotSupportedError` at feedback filter


MizukiTemma:
### Describe the Bug

In some combinations of status and rating choices, an error occurs when filter is applied for feedback list.

### Steps to Reproduce

1. Go to the global or a region feedback
2. Uncheck "unread" and "positive"
3. Apply filter 
4. See `NotSupportedError` (Calling QuerySet.select_related() after union() is not supported)

### Expected Behavior

Filter result will be shown.

### Actual Behavior

Error occurs.

### Additional Information

The same promlem can exist in filters for other contents.
Other but similar problems can exist too.

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

PR #2324, opened 2023-06-24
Fix feedback filter error

MizukiTemma:
### Short description

This PR fixes `NotSupportedError` which occurs in the feedback filter.

### Proposed changes

- Make an unrestricted query set after `.union()` so `select_related` can be operated on the query set (like in `get_pages` method of region).

### Side effects

- None (I hope), since no change on the filter query calcuration.

### Resolved issues

Fixes: #2273 

### DIFF
--- a/integreat_cms/release_notes/current/unreleased/2273.yml
+++ b/integreat_cms/release_notes/current/unreleased/2273.yml
@@ -1,0 +1,2 @@
+en: Fix feedback filter error
+de: Behebe Fehler im Feedback Filter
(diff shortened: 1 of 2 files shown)


david-venhoff:
Looks mostly good to me, maybe you can move the code below https://github.com/digitalfabrik/integreat-cms/blob/9f272712025903f2ed4209998416ab030302a1a6/integreat_cms/cms/forms/feedback/region_feedback_filter_form.py#L116
so that we don't need to make a new database request every time?

MizukiTemma:
@david-venhoff Nice idea 🙂 applid!

integreat_cms/release_notes/current/unreleased/2273.yml
```diff
@@ -0,0 +1,2 @@
+en: Fix feedback filter error
+de: Behebe Fehler in Feedback Filter
```
ulliholtgrave, line 2:
```suggestion
de: Behebe Fehler im Feedback Filter
```

ulliholtgrave (changes requested):
Besides this small typo, this works as expected and looks good to me :+1:

ulliholtgrave (approved):
Thanks, Looks good 👍

Issue #2325, opened 2023-06-26
Add url for organizations


steffenkleinle:
### Motivation

Our goal is to make information in integreat more trustworthy. One idea for this is to add organization footers (https://issues.tuerantuer.org/browse/IGAPP-1191) for pages that are edited and created by organizations in order to show users directly who created specific pages. To make this more trustworthy, we should not just show some text and logo but also add a link to the homepage of the organization.

### Proposed Solution
Add a url/website/link field to the organization edit screen AND deliver it with the API.

PR #2328, opened 2023-06-28
Add website field to organization

MizukiTemma:
### Short description

This PR adds a new field `website` to the model Organization and makes a logo mandatory for each organization.

### Proposed changes

- Add a field `website` where address of homepage etc can be given.
- Set `blank=False` (by letting the default valid) for icon and website to reject saves without icon or website.
- But `null=True` so existing organizations without icon/website don't cause any problem.
- Adjust test.

### Side effects

- In my opinion this feature does not bring that much progress on trustability of page/organization if the website field is allowed to be empty and nowadays (almost) every organization has a website, blog or something. So I set not only the logo but also the website mandatory (for new saves and creations) but this is of course open to discussion.

### Resolved issues

Fixes: #2325 


integreat_cms/api/v3/locations.py
```diff
@@ -98,6 +98,7 @@ def transform_poi_translation(poi_translation):
             "slug": poi.organization.slug,
             "name": poi.organization.name,
             "logo": poi.organization.icon.url if poi.organization.icon else None,
+            "website": poi.organization.website if poi.organization.website else None,
```
timoludwig, line 101:
```suggestion
            "website": poi.organization.website,
```
If we don't want `None` values there in the long time, I guess it's fine to go with an empty string for the time being. This also helps the apps if they can be sure that the result is a string and not a string or `None`.

integreat_cms/api/v3/pages.py
```diff
@@ -87,6 +87,7 @@ def transform_page(page_translation):
             "slug": organization.slug,
             "name": organization.name,
             "logo": organization.icon.url if organization.icon else None,
+            "website": organization.website if organization.website else None,
```
timoludwig, line 90:
```suggestion
            "website": organization.website,
```

integreat_cms/cms/fixtures/test_data.json
```diff
@@ -3643,6 +3643,7 @@
       "name": "Test Organization",
       "slug": "test-organization",
       "icon": null,
+      "website": null,
```
timoludwig, line 3646: If we want to enforce icon & website, I would also add one to the test data.

integreat_cms/cms/migrations/0071_add_organization_website.py
```diff
@@ -0,0 +1,33 @@
...
+
+    dependencies = [
+        ("cms", "0070_alter_poicategory_icon"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="organization",
+            name="website",
+            field=models.URLField(max_length=250, null=True, verbose_name="website"),
```
timoludwig, line 20:
```suggestion
            field=models.URLField(max_length=250, verbose_name="website"),
```

integreat_cms/cms/models/users/organization.py
```diff
@@ -43,6 +42,8 @@ class Organization(AbstractBaseModel):
         default=timezone.now, verbose_name=_("creation date")
     )
 
+    website = models.URLField(max_length=250, null=True, verbose_name=_("website"))
```
timoludwig, line 45:
```suggestion
    website = models.URLField(max_length=250, verbose_name=_("website"))
```
`null=True` on a char field doesn't make much sense anyway. Empty values are converted from `None` to an empty string the first time the form is submitted...

integreat_cms/cms/models/users/organization.py
```diff
@@ -28,7 +28,6 @@ class Organization(AbstractBaseModel):
         verbose_name=_("logo"),
         on_delete=models.SET_NULL,
```
timoludwig, line 29:
```suggestion
        on_delete=models.PROTECT,
```
I guess if we want to enforce logos, we have to protect them from being deleted?

integreat_cms/cms/models/users/organization.py
```diff
@@ -28,7 +28,6 @@ class Organization(AbstractBaseModel):
...
         related_name="icon_organizations",
-        blank=True,
         null=True,
```
timoludwig, line 31:
```suggestion
```

timoludwig (commented):
Thanks! :+1:

> But `null=True` so existing organizations without icon/website don't cause any problem.

I'm not sure this is a good idea. The app wants this rules to be enforced to make sure that their api client can fetch the api and get the results it is expecting. This does not work when we leave invalid objects in the database. So I suggest to write a migration to either delete all invalid organizations or set default values for all missing fields.

> So I set not only the logo but also the website mandatory (for new saves and creations) but this is of course open to discussion.

Makes sense to me :+1:

Also, I would suggest to update the [API documentation](https://github.com/digitalfabrik/integreat-cms/blob/develop/docs/src/api-docs.rst) with these changes.

integreat_cms/cms/fixtures/test_data.json
MizukiTemma: I could add an Integreat URL, for example, to `webseite` of the test data, but difficult is the icon. 

We currently don't have any test for the media library. To add an icon to the organization test data, I had to create first an MediaFile object, for which a picture should be stored in the folder `integreat_cms/media/` but this folder will not be stored in the git contro :/ Is there another way?

Another question is, whether it is okay that all the organizations will be deleted after the migration because none of them has right now `website` field. Default values may be cool in this case, tohugh I have no idea what value is reasonable 🤔

timoludwig: How about copying one static image into the global media library in the [loadtestdata.sh](https://github.com/digitalfabrik/integreat-cms/blob/develop/tools/loadtestdata.sh) script? And then add a media file object to the test data with this path?

As for the default values: I think it would be fine to set them to an empty string, sure that's not what the apps expect, but maybe less invasive than deleting all invalid organization objects... And then we can tell the service team that they should insert the missing information as soon as possible...

MizukiTemma: Thank you for advice 🙂 The branch has been updated.

steffenkleinle:
Any ideas when this will be merged/deployed? Thanks :)

timoludwig:
> Any ideas when this will be merged/deployed? Thanks :)

Depends on how urgent it is :wink:
If required, I guess we can merge & deploy this in the next days.

steffenkleinle:
> > Any ideas when this will be merged/deployed? Thanks :)
> 
> Depends on how urgent it is wink If required, I guess we can merge & deploy this in the next days.

I'd say not really urgent as it is a new feature. However, it still blocks the complete feature which would be ready otherwise. So soonish would be nice but definitely no stress or hurry needed.

integreat_cms/cms/models/users/organization.py
```diff
@@ -26,10 +26,9 @@ class Organization(AbstractBaseModel):
     icon = models.ForeignKey(
         MediaFile,
         verbose_name=_("logo"),
-        on_delete=models.SET_NULL,
+        on_delete=models.PROTECT,
         related_name="icon_organizations",
-        blank=True,
-        null=True,
+        default="",
```
timoludwig, line 31:
```suggestion
```
An empty string does not make sense for a foreign key field. A foreign key can only contain the reference to the foreign object, or be `None`. Since we don't want to allow the latter, we have to make sure that all organization objects have a valid icon selected before we execute this migration on the production system.

integreat_cms/cms/models/users/organization.py
```diff
@@ -43,6 +42,8 @@ class Organization(AbstractBaseModel):
         default=timezone.now, verbose_name=_("creation date")
     )
 
+    website = models.URLField(max_length=250, verbose_name=_("website"), default="")
```
timoludwig, line 45:
```suggestion
    website = models.URLField(max_length=250, verbose_name=_("website"))
```
It's similar for the icon field - at least, this is a CharField, so an empty string might be ok as temporary default for the existing entries, but I would not leave it in the model class.

integreat_cms/cms/migrations/0073_add_organization_website.py
```diff
@@ -0,0 +1,33 @@
...
+    dependencies = [
+        ("cms", "0072_alter_poicategory_icon"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="organization",
+            name="website",
+            field=models.URLField(default="", max_length=250, verbose_name="website"),
+        ),
```
timoludwig, line 21:
```suggestion
        ),
        migrations.AlterField(
            model_name="organization",
            name="website",
            field=models.URLField(max_length=250, verbose_name="website"),
        ),
```
Directly after creating the field and populating the existing entries with empty strings, I would alter the field so we don't have an empty string for new objects.

integreat_cms/cms/migrations/0073_add_organization_website.py
```diff
@@ -0,0 +1,33 @@
...
+        migrations.AlterField(
+            model_name="organization",
+            name="icon",
+            field=models.ForeignKey(
+                default="",
```
timoludwig, line 26:
```suggestion
```

integreat_cms/cms/templates/organizations/organization_list_row.html
```diff
@@ -31,6 +31,16 @@
...
     </td>
+    <td class="pr-2">
+        {% if organization.website %}
+            <a href="{{ organization.website }}"
+               title="{% translate "To website" %}">
+                <i icon-name="check" class="text-green-500"></i>
+            </a>
+        {% else %}
+            <i icon-name="x" class="text-red-500"></i>
+        {% endif %}
```
timoludwig, line 42:
```suggestion
        <a href="{{ organization.website }}"
           target="_blank"
           rel="noopener noreferrer"
           class="text-blue-500 hover:underline"
           title="{% translate "To website" %}">
            {{ organization.website }}
        </a>
```
I think we can directly show the website here.

integreat_cms/cms/templates/organizations/organization_list.html
```diff
@@ -21,6 +21,7 @@ <h1 class="heading">{% translate "Manage Organizations" %}</h1>
                     <th class="text-sm text-left uppercase py-3 px-2">{% translate "Members" %}</th>
                     <th class="text-sm text-left uppercase py-3 px-2">{% translate "Maintained Pages" %}</th>
                     <th class="text-sm text-left uppercase py-3 px-2">{% translate "Logo" %}</th>
+                    <th class="text-sm text-left uppercase py-3 px-2">{% translate "Website" %}</th>
```
timoludwig, line 24:
```suggestion
                    <th class="text-sm text-left uppercase py-3 pr-2">{% translate "Logo" %}</th>
                    <th class="text-sm text-left uppercase py-3 pr-2">{% translate "Website" %}</th>
```
I'd use the same padding for the table headline like for the table cells

integreat_cms/cms/templates/organizations/organization_form.html
```diff
@@ -66,6 +66,17 @@ <h3 class="heading font-bold text-black">
...
         </div>
+        <div class="mb-4 rounded border border-solid border-blue-500 shadow-2xl bg-white">
+            <div class="p-4 rounded bg-water-500">
+                <h3 class="heading font-bold text-black">
+                    <i icon-name="settings" class="mr-2"></i> {% translate "URL" %}
+                </h3>
+            </div>
+            <div class="px-4 pb-4">
+                <label for="{{ form.website.id_for_label }}">{{ form.website.label }}</label>
+                {% render_field form.website %}
```
timoludwig, line 77: I think we can move this field into the "General settings" box together with the "name" input field.

tests/api/api_config.py
```diff
@@ -44,14 +44,14 @@
         "/augsburg/de/wp-json/extensions/v3/pages/",
         "tests/api/expected-outputs/augsburg_de_pages.json",
         200,
-        5,
+        7,
```
timoludwig, line 47:
```suggestion
        5,
```
You can cut down those queries by selecting the related icon objects in the API.

Replace `region.pages.select_related("organization")` by `region.pages.select_related("organization__icon")` in `integreat_cms/api/v3/pages.py` and `.select_related("category", "organization")` by `.select_related("category", "organization__icon")` in `integreat_cms/api/v3/locations.py`.

tests/api/api_config.py
```diff
@@ -44,14 +44,14 @@
...
     ),
     (
         "/api/augsburg/ar/pages/",
         "/augsburg/ar/wp-json/extensions/v3/pages/",
         "tests/api/expected-outputs/augsburg_ar_pages.json",
         200,
-        5,
+        7,
```
timoludwig, line 54:
```suggestion
        5,
```

tests/api/api_config.py
```diff
@@ -65,21 +65,21 @@
         "/augsburg/de/wp-json/extensions/v3/locations/",
         "tests/api/expected-outputs/augsburg_de_locations.json",
         200,
-        5,
+        6,
```
timoludwig, line 68:
```suggestion
        5,
```

tests/api/api_config.py
```diff
@@ -65,21 +65,21 @@
...
     ),
     (
         "/api/augsburg/en/locations/",
         "/augsburg/en/wp-json/extensions/v3/locations/",
         "tests/api/expected-outputs/augsburg_en_locations.json",
         200,
-        5,
+        6,
```
timoludwig, line 75:
```suggestion
        5,
```

tests/api/api_config.py
```diff
@@ -65,21 +65,21 @@
...
     ),
     (
         "/api/augsburg/ar/locations/",
         "/augsburg/ar/wp-json/extensions/v3/locations/",
         "tests/api/expected-outputs/augsburg_ar_locations.json",
         200,
-        5,
+        6,
```
timoludwig, line 82:
```suggestion
        5,
```

tests/api/api_config.py
```diff
@@ -135,7 +135,7 @@
         "/augsburg/de/wp-json/extensions/v3/children/?depth=2",
         "tests/api/expected-outputs/augsburg_de_children_depth_2.json",
         200,
-        11,
+        13,
```
timoludwig, line 138:
```suggestion
        11,
```

timoludwig (changes requested):
@MizukiTemma Thanks a lot for your changes! :muscle: 

In my opinion, a few things still need to be addressed before this can be merged.

MizukiTemma:
@timoludwig thank you for suggestion and explanation! I applied them 🙂

timoludwig (approved):
Awesome, thanks! :rocket:

timoludwig:
@MizukiTemma One final request: could you add a release note for this change?

tests/api/expected-outputs/augsburg_ar_locations.json
```diff
@@ -85,7 +85,8 @@
       "id": 1,
       "slug": "test-organization",
       "name": "Test Organization",
-      "logo": null
+      "logo": "http://localhost:8000/media/global/integreat-icon.svg",
```
PeterNerlich, line 88: Is it secured that the server will always be at `localhost:8000` when running tests? @timoludwig

PeterNerlich (changes requested):
Except for the missing release notes the code looks good to me. It also seems to behave as advertised, only the migration doesn't work when there are existing organisations without a logo (as is the case with the previous test data from `develop`).

```python3
  Apply all migrations: admin, auth, cms, contenttypes, db_mutex, linkcheck, sessions
Running migrations:
  Applying cms.0073_add_organization_website...Traceback (most recent call last):
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
psycopg2.errors.NotNullViolation: column "icon_id" of relation "cms_organization" contains null values


  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/bin/integreat-cms-cli", line 52, in <module>
    main()
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/bin/integreat-cms-cli", line 48, in main
    execute_from_command_line(sys.argv)
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/core/management/base.py", line 354, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/core/management/base.py", line 398, in execute
    output = self.handle(*args, **options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/core/management/base.py", line 89, in wrapped
    res = handle_func(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/core/management/commands/migrate.py", line 244, in handle
    post_migrate_state = executor.migrate(
                         ^^^^^^^^^^^^^^^^^
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/db/migrations/executor.py", line 117, in migrate
    state = self._migrate_all_forwards(state, plan, full_plan, fake=fake, fake_initial=fake_initial)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/db/migrations/executor.py", line 147, in _migrate_all_forwards
    state = self.apply_migration(state, migration, fake=fake, fake_initial=fake_initial)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/db/migrations/executor.py", line 227, in apply_migration
    state = migration.apply(state, schema_editor)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/db/migrations/migration.py", line 126, in apply
    operation.database_forwards(self.app_label, schema_editor, old_state, project_state)
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/db/migrations/operations/fields.py", line 244, in database_forwards
    schema_editor.alter_field(from_model, from_field, to_field)
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/db/backends/base/schema.py", line 608, in alter_field
    self._alter_field(model, old_field, new_field, old_type, new_type,
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/db/backends/postgresql/schema.py", line 196, in _alter_field
    super()._alter_field(
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/db/backends/base/schema.py", line 765, in _alter_field
    self.execute(
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/db/backends/base/schema.py", line 145, in execute
    cursor.execute(sql, params)
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/db/backends/utils.py", line 98, in execute
    return super().execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/db/backends/utils.py", line 66, in execute
    return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/db/backends/utils.py", line 75, in _execute_with_wrappers
    return executor(sql, params, many, context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/db/backends/utils.py", line 79, in _execute
    with self.db.wrap_database_errors:
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/db/utils.py", line 90, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/home/peter/NewStructure/Projects/Code/integreat-cms-clone/.venv/lib/python3.11/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
django.db.utils.IntegrityError: column "icon_id" of relation "cms_organization" contains null values
```

</details>

How about having some 1×1 px or placeholder image that is set as default for those? Or would that circumvent the intention by giving the organisations another state where no immediate action is required?

tests/api/expected-outputs/augsburg_ar_locations.json
timoludwig: Yes, this might only be different if somebody sets another port explicitly and runs the tests locally, but I guess in this case the person just has to deal with the failing tests... We have to assume something here, we cannot put variables in json files.
This base url is also used in all other expected api responses:
https://github.com/digitalfabrik/integreat-cms/blob/7a8fda13ca5b35a90c79077174335e01c87beb2d/tests/api/expected-outputs/augsburg_de_pages.json#L4

timoludwig:
> Except for the missing release notes the code looks good to me.

The [release note](https://github.com/digitalfabrik/integreat-cms/pull/2328/files#diff-c3dc8ce89196fc5a00018b7fe3ec1ed972e6795e0b262664a66241ca2276a42c) was added in the mean time :smiley: 

> It also seems to behave as advertised, only the migration doesn't work when there are existing organisations without a logo (as is the case with the previous test data from `develop`).

Yes, this is definitely a valid concern, but I don't see a better solution to be honest. We can only require database constraints for all objects in a table, we cannot exclude the existing organizations from it.

> How about having some 1×1 px or placeholder image that is set as default for those? Or would that circumvent the intention by giving the organisations another state where no immediate action is required?

Hmm, it's hard to set a placeholder image as default, since we would have to add physical file handling as part of the migration file, which works (since it's just python code), but it kind of contradicts the principle that migrations should only alter the database scheme and contents, not the file system. So I opted for this caveat that we have to set an image for all existing organizations before we can install this update.

MizukiTemma:
@timoludwig should we then start to tell the service team that all the organisations now should have a logo? This will take a while.

timoludwig:
> @timoludwig should we then start to tell the service team that all the organisations now should have a logo? This will take a while.

No, I think we can do this via the command line and set an existing image in the production database for all of them, e.g. [this one](https://admin.integreat-app.de/media/global/2022/07/Alltag-Zusammenleben.png):
```
from integreat_cms.cms.models import *
default_icon = MediaFile.objects.get(file="global/2022/07/Alltag-Zusammenleben.png")
Organization.objects.all().update(icon=default_icon)
```

We just have to do this step for all prod environments separately (integreat, malte, obdach)...

Then after the update is done, we can tell the service team they should add icon + website for all organizations (thus it will be less work for them since they can add icon and website in one step).

MizukiTemma:
@timoludwig @PeterNerlich Thank you! I'll then merge this PR.

timoludwig:
> @timoludwig @PeterNerlich Thank you! I'll then merge this PR.

Sounds good, I updated all default icons of existing organization objects on integreat-test and malte-test, so we can verify that the planned approach works with the test package installation.

Issue #2171, opened 2023-03-27
Archiving mirrored pages causes inconsistent state


timoludwig:
### Describe the Bug

When a page, that is mirrored on another page, is archived, this leads to several problems:
1. The mirrored page is not selected in the dropdown anymore (because archived pages are removed from the queryset of possible mirroring targets), which means the first page of that region becomes the mirroring target when the save button is clicked, which is easily overlooked and means potentially unrelated content is inserted as live content
2. The content of the archived mirrored page is still live somewhere, which contradicts the purpose of archiving a page

### Steps to Reproduce

1. Mirror a page somewhere
2. Archive the mirrored page

### Expected Behavior

Either one of two things should happen:
1. Archiving a mirrored page should not be possible, similar to deleting a mirrored page
2. The live content should be disabled when a mirrored page is archived (in this case we could also allow deleting a page with the same effect)

### Actual Behavior

The mirrored page is archived, but still referenced as live content

PR #2329, opened 2023-07-01
Do not archive mirrored pages

MizukiTemma:
### Short description

This PR solves the inconsistent state of mirrored pages by prevanting archiving of mirrored pages.

### Proposed changes

- Deactivate the archiving icon of mirrored pages in the tree 
- Hide the archiving button in the form if the page is mirrored
- Check if there are mirorring pages before archiving a page (both in single request and bulk action)
- Adjust error/success messages of the bulk action and titles of the icon

### Side effects

- None?
- I have chosen the first one among the two suggested solutions (expected behaviour) because it seems to be more in accordance with the existing behaviour of page deletion and probably causes less confusion than the second option.

### Resolved issues

Fixes: #2171 

### DIFF
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -6035,6 +6035,28 @@ msgid "Archive this page"
 msgstr "Diese Seite archivieren"
 
 #: cms/templates/pages/page_form_sidebar/actions_box.html
+msgid ""
+"You cannot archive a page which is embedded as live content from another "
+"page."
+msgstr ""
+"Diese Seite kann nicht archiviert werden, da sie von einer anderen Seite als "
+"Live-Inhalt eingebunden wurde."
+
+#: cms/templates/pages/page_form_sidebar/actions_box.html
+msgid ""
+"To archive this page, you have to remove the embedded live content from this "
+"page first:"
+msgid_plural ""
+"To archive this page, you have to remove the embedded live content from "
+"these pages first:"
+msgstr[0] ""
+"Um diese Seite zu archivieren, müssen Sie den eingebetteten Live-Inhalt von "
+"dieser Seite entfernen:"
+msgstr[1] ""
+"Um diese Seite zu archivieren, müssen Sie den eingebetteten Live-Inhalt von "
+"diesen Seiten entfernen:"
+
+#: cms/templates/pages/page_form_sidebar/actions_box.html
 #: cms/templates/pages/page_tree_archived_node.html
 #: cms/templates/pages/page_tree_node.html
 msgid "Delete page"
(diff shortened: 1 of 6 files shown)


timoludwig (commented):
Thanks a lot, works as expected! :muscle: 

I thought that maybe it could be helpful to know which pages prevent the page from being archived, so in case they want to proceed they can e.g. contact the manager of the other region to remove the live content etc...

MizukiTemma:
@timoludwig In the action box it shows where the page is mirrored. I'll modify the explanation to "Um diese Seite zu löschen **oder zu archivieren.**....."

Where am I mirrored

timoludwig (approved):
Cool, thanks! :+1:

svenseeberg:
I'm a little worried that some regions have to maintain pages because other regions do mirror them. Can we think of a solution here? Maybe a button to remove all mirror pages by force from the source region and notify the maintainers via mail?

MizukiTemma:
That's a nice idea 😃 @osmers What way do you think is the best to inform content creators that a mirrored page has been removed if we introduce force archiving/deleting of mirrored pages? I have a bit of worry that not all of them check E-mails...

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -6034,6 +6034,28 @@ msgstr "Seite archivieren"
...
+"page first:"
+msgid_plural ""
+"To archive this page, you have to remove the embedded live content from "
+"these pages first:"
+msgstr[0] ""
+"Um diese Seite zu archivieren, müssen Sie den eingebetteten Live-Inhalt von "
+"dieser Seite entfernen:"
+msgstr[1] ""
+"Um diese Seite zu archivieren, müssen Sie den eingebetteten Live-Inhalt von "
+"dieser Seiten entfernen:"
```
david-venhoff, line 6057:
```suggestion
"diesen Seiten entfernen:"
```

david-venhoff (approved):
Looks good to me. As a small suggestion, you might consider to move the `is mirrored` checks into the `page.archive()` method, to reduce code duplication. I don't think this is blocking the pr though.

MizukiTemma:
> As a small suggestion, you might consider to move the `is mirrored` checks into the `page.archive()` method, to reduce code duplication. I don't think this is blocking the pr though.

Nice idea, though I'll leave the code as it is to keep the current check & message schema and to avoid bulk archive action getting inturruped at the first mirrored page 🙂

osmers:
@MizukiTemma good question - I think generally they are registered with their work account and should therefore read the emails. Maybe you can also always send an email to me so that I am aware and can inform the corresponding person from the service team to double check with the municipality. 
I just had the issue as well that I need to archive the corona pages but can't bcs they are mirrored in many regions. 
Did you introduce a solution already or not yet?

MizukiTemma:
@osmers Thank you for feedback! This notification function is not yet implemented. I have updated the issue for that #2338

osmers:
@MizukiTemma at the same time I noticed that archived pages are also shown as mirroring a page and created an issue for that bcs only live pages should be in the list. Otherwise people would get notified that content has been removed even though they already archived the page that had the removed content.

Issue #2270, opened 2023-05-10
[Obdach] POI category icons


f1sh1918:
### Motivation
For vernetztes Obdach the we need new poi category icons.

### Proposed Solution
 Please make the new icons available via dropdown and deliver it to the fronted via api:
- advice
- daily_routine
- finance
- medical_aid
- other_help
- structure

[integreat_categories_iconset cms.zip](https://github.com/digitalfabrik/integreat-cms/files/11443922/integreat_categories_iconset.cms.zip)

### Note
Probably there is no separate implementation for icons for vernetztes Obdach, so all integreat category icons should also be available there.

JoeyStk:
@f1sh1918 I have two questions:
1. question: 
> other_help

In our integreat categories we already have "other". Am I correct in assuming, that we don't need the separate category "other_help" anymore if you use the one from Integreat?

2. question:
What would be the best translation for both finance and structure? Would you prefer "Finanzielles" or "Finanzen", and the same for structure. Is "Strukturelles" or "Struktur" better?

Thank you in advance :)

f1sh1918:
@JoeyStk 

1. if you check the images in the zip file you can see the differences. so one is just like "Sonstiges" and the other "Sonstige Hilfen" as the images may describe. So we should have both. I mean some may not be useful for integreat but here i'm just talking about icons and not about categories we may have in integreat. I'm even a bit confused why we have to translate icon names. I think thats not needed but looks more consistent in the ui. Afaik only administrators like us can create categories :)


2.  I really don't mind since no user will see that names. In my opinion the dropdown should just consist the icons and not names so the user would know at instant how the icon looks like

3. the existing colors are fine

thx for taking the issue :)

PR #2331, opened 2023-07-03
[Obdach] Add new icons for poicategory

JoeyStk:
### Short description
This PR adds six new icons for poicategory

### Proposed changes

- Add svgs for the categories. I adapted the svgs a bit to fit the circle styling of the other svgs, as seen in the PR for the education category. I hope that was correct.
- I pushed a branch "feature/add_new_poi_categories" where I also created the POI categories. As far as my understanding goes, we only need the icons for now (therefore this new branch)

### Side effects

- None I can think of

### Resolved issues

Fixes: #2270


timoludwig:
> I pushed a branch "feature/add_new_poi_categories" where I also created the POI categories. As far as my understanding goes, we only need the icons for now (therefore this new branch)

This is a misunderstanding, the constants are not the poi categories, but only the choices for the icons. So yes, they need to be included in this pr, otherwise the icons are not usable.

And no, we don't need the actual poi categories in the repository, those are created in the database.

JoeyStk:
Thank you for clarifying. So it would have been better to use the other branch.

> And no, we don't need the actual poi categories in the repository, those are created in the database.

I'm not completely sure I understand correctly. For the icon names we still need a migration file, right?
If I'm reading this correctly I would add the icon names to the constants, add a migration file, extend the translation file and add a release note. Is that correct?

timoludwig:
> I'm not completely sure I understand correctly. For the icon names we still need a migration file, right? If I'm reading this correctly I would add the icon names to the constants, add a migration file, extend the translation file and add a release note. Is that correct?

Yes, I think that's it.

JoeyStk:
> Yes, I think that's it.

Okay, everything should be uploaded now :)

integreat_cms/release_notes/current/unreleased/2270.yml
```diff
@@ -0,0 +1,2 @@
+en: Add new POI category icons
+de: Füge neue Icons für POI-Kategoriesn hinzu
```
timoludwig, line 2:
```suggestion
de: Füge neue Icons für POI-Kategorien hinzu
```

timoludwig (approved):
Thanks! :+1:

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -1426,6 +1446,10 @@ msgstr "Dienstleistung"
 msgid "Shopping"
 msgstr "Einkaufen"
 
+#: cms/constants/poicategory.py
+msgid "Structure"
+msgstr "Stuktur"
```
ulliholtgrave, line 1451:
```suggestion
msgstr "Struktur"
```

ulliholtgrave (changes requested):
Please fix the typo and we are good to go :+1:

integreat_cms/locale/de/LC_MESSAGES/django.po
JoeyStk: Thank you, good catch :facepalm:

ulliholtgrave (approved):
Thanks, Looks good

Issue #2337, opened 2023-07-05
Add bulk action to publish pages & set to draft


timoludwig:
### Motivation

From time to time, the service team needs to adjust the publication status of multiple pages at once, so either publish multiple pages or set them to draft.
At the moment, the cms team needs to do this manually in the Django shell, which takes time and is error prone.

### Proposed Solution

Add bulk actions for these tasks, only available for staff members.
For consistency, I'd suggest to add them for events and pois, too.

### Alternatives

Do the task manually

### User Story

As a staff member, I want to bulk publish all selected items from the list or set them to draft.

### Additional Context
This was rejected earlier in #1517, but we realized this task is needed from time to time, even if pages can be cloned as draft.

PR #2341, opened 2023-07-06
Add drafting and publishing bulk action

JoeyStk:
### Short description
This PR adds the option to change pages, events and pois to draft or public in a bulk action

### Proposed changes

- Add release note
- Add function to change the translation status
- Add bulk action call to events, pois and pages

### Side effects

None I'm aware of

### Resolved issues

Fixes: #2337

### DIFF
--- a/integreat_cms/cms/views/utils/translation_status.py
+++ b/integreat_cms/cms/views/utils/translation_status.py
@@ -1,0 +1,43 @@
+"""
+This file contains the helper function to change the translation file
+"""
+from django.contrib import messages
+from django.utils.translation import gettext_lazy as _
+
+from ...constants import status
+
+
+def change_translation_status(request, selected_content, language_slug, desired_status):
+    """
+    Helper function to change the translation status
+
+    :param request: The current request
+    :type request: ~django.http.HttpRequest
+
+    :param selected_content: The current queryset
+    :type selected_content: ~integreat_cms.cms.models.abstract_content_model.ContentQuerySet
+
+    :param language_slug: The language slug of the current language
+    :type language_slug: str
+
+    :param desired_status: The desired status
+    :type desired_status: str
+    """
+
+    status_translation = {"PUBLIC": "Published", "DRAFT": "Draft"}
+
+    if request.user.is_superuser or request.user.is_staff:
+        for content in selected_content:
+            translation = content.get_translation(language_slug)
+            translation.status = desired_status
+            translation.pk = None
+            translation.version += 1
+            if desired_status == status.DRAFT:
+                translation.all_versions.update(status=desired_status)
+            translation.save()
+            messages.success(
+                request,
+                _('The status of "{}" was successfully changed to "{}"').format(
+                    translation, _(status_translation[desired_status])
+                ),
+            )
--- a/integreat_cms/cms/templates/events/event_list.html
+++ b/integreat_cms/cms/templates/events/event_list.html
@@ -138,6 +138,19 @@
                 <option data-bulk-action="{% url 'bulk_archive_events' region_slug=request.region.slug language_slug=language.slug %}">
                     {% translate "Archive events" %}
                 </option>
+                {% if request.user.is_superuser or request.user.is_staff %}
+                    {% translate "events" as content_type %}
+                    <option data-bulk-action="{% url 'publish_multiple_events' region_slug=request.region.slug language_slug=language.slug %}">
+                        {% blocktranslate trimmed %}
+                            Publish {{ content_type }}
+                        {% endblocktranslate %}
+                    </option>
+                    <option data-bulk-action="{% url 'draft_multiple_events' region_slug=request.region.slug language_slug=language.slug %}">
+                        {% blocktranslate trimmed %}
+                            Draft {{ content_type }}
+                        {% endblocktranslate %}
+                    </option>
+                {% endif %}
                 {% if MT_PERMITTED %}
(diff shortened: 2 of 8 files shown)


MizukiTemma (changes requested):
Thank you for the PR 💪 This is a really convenient feature 😃 

I have three suggestions:
1. It's probably more friendly to show success/error message after bulk action so one can be sure that the bulk action has been really orerated.
2. We may better check user role in the view too. The bulk publish and draft are hidden for normal users but make it sure won't harm.
3. Currently it toggles only translations of the default language independent of in which language tab the user is right now. I think this feature is needed also for other languages, especially because sometimes publication status of maschine translated content is somehow not correct......

MizukiTemma:
And those two points are subject to discussion:

Those two behaviours should be inherited in the bulk publishing/draft too for consisteny reason.

1. New version
When a content is published or set to draft by the buttons in the form, a new version is created even if there is no change. 
2. Set all to draft 
When a content is set to draft, all the version are set to draft too.

JoeyStk:
> And those two points are subject to discussion:
Those two behaviours should be inherited in the bulk publishing/draft too for consisteny reason, in my opinion.
New version
When a content is published or set to draft by the buttons in the form, a new version is created even if there is no change.
Set all to draft
When a content is set to draft, all the version are set to draft too.

Are there opinions about this? :smiley:

integreat_cms/cms/views/utils/translation_status.py
```diff
@@ -0,0 +1,36 @@
...
+        for content in selected_content:
+            translation = content.get_translation(language_slug)
+            translation.status = desired_status
+            translation.save()
+            messages.success(
+                request,
+                _("The status of {} was successfully changed to {}").format(
+                    translation, desired_status.lower()
+                ),
+            )
```
MizukiTemma, line 36: This is my suggestion to the translation problem:

```suggestion
    status = {
        "PUBLIC": "Published", 
        "DRAFT": "Draft"
    }

    if request.user.is_superuser or request.user.is_staff:
        for content in selected_content:
            translation = content.get_translation(language_slug)
            translation.status = desired_status
            translation.save()
            messages.success(
                request,
                _("The status of \"{}\" was successfully changed to \"{}\"").format(
                    translation, _(status[desired_status])
                ),
            )
```

MizukiTemma (changes requested):
> I had assumed that the status would be translated because in the constants file it's already translated. Do you have an idea on how to fix this?

The problem is that the parameter `desired_status` is a string, not a variable anymore, which means `desired_status = "PUBLIC"` (or "DRAFT") not  `desired_status = PUBLIC`.

One possible solution may be putting a small dictionary and convert `PUBLIC` to `Published`and to translate this.

integreat_cms/cms/views/utils/translation_status.py
MizukiTemma: Oh, and we can maybe add an error message for else-case. I know those options are not shown for non staff users but...... 😅

JoeyStk:
Thank you very much for the suggestion. Yes, this seems to fix the issue :)

timoludwig:
> > And those two points are subject to discussion:
> > Those two behaviours should be inherited in the bulk publishing/draft too for consisteny reason, in my opinion.
> > New version
> > When a content is published or set to draft by the buttons in the form, a new version is created even if there is no change.
> > Set all to draft
> > When a content is set to draft, all the version are set to draft too.
> 
> Are there opinions about this? :smiley:

@JoeyStk I agree with @MizukiTemma in both questions.

integreat_cms/cms/views/utils/translation_status.py
```diff
@@ -0,0 +1,42 @@
...
+    status = {"PUBLIC": "Published", "DRAFT": "Draft"}
+
+    if request.user.is_superuser or request.user.is_staff:
+        for content in selected_content:
+            translation = content.get_translation(language_slug)
+            translation.status = desired_status
+            translation.pk = None
+            translation.version += 1
+            if desired_status == "DRAFT":
+                translation.all_versions.update(status=desired_status)
```
JoeyStk, line 35: @MizukiTemma Is this what you had in mind? :)

MizukiTemma (approved):
Thank you for updates 💪 looks good 🎉

seluianova (commented):
Looks good to me!
I only found a little thing:
If the page status has not actually been changed (if it was already published or already draft), the message is anyway:
 _The status of "{page}" was successfully changed to "Published/Draft"_
Do we want to handle this case separately? To show another message maybe?

JoeyStk:
Thank you @seluianova for your review!

> Do we want to handle this case separately? To show another message maybe?

I'm fine of doing it either way. On one hand I think you're right and it would be nice to be more precise, on the other hand since this functionality is only going to be used by the service team, I think how it is at the moment should be sufficient as well. But maybe @ulliholtgrave, @osmers or @svenseeberg are more qualified to make this decision or give some insights :)

osmers:
I agree with @JoeyStk - also it is not 100% wrong since the page(s) are still in draft/published. So it's kinda reassuring. For me there is no need to have two separate messages :)

JoeyStk:
>  agree with @JoeyStk - also it is not 100% wrong since the page(s) are still in draft/published. So it's kinda reassuring. For me there is no need to have two separate messages :)

Thank you. @seluianova is this fine with you and if so, could you please review again? :)

seluianova (approved):

Issue #1997, opened 2023-01-12
Add tests for media library actions


timoludwig:
### Motivation

At the moment, we only have tests for viewing the media library:
https://github.com/digitalfabrik/integreat-cms/blob/f973afc2eb7813ea96dd434c235c69cb3f4afd50/tests/cms/views/view_config.py#L39-L41

### Proposed Solution

It would be very cool to have tests for interactions with the media library:

- [ ] Uploading files
- [ ] Renaming files
- [ ] Changing alt texts of files
- [ ] Replacing files
- [ ] Deleting files
- [ ] Creating directories
- [ ] Renaming directories
- [ ] Deleting directories

PR #2346, opened 2023-07-15
Update dependencies

timoludwig:
### Short description

Update dependencies

### Proposed changes

- Update npm depdendencies
- Update pip depdendencies

### Side effects

I don't really understand why djlint has such a huge diff again, but I don't see an immediate problem with it (apart from a few annoying git conflicts).
However, we could also try to identify which version bump caused these changes to then clarify whether this is a bug of djlint (which might get reverted in future versions) or a new feature / style guideline which will stay.

### Resolved issues

- https://github.com/digitalfabrik/integreat-cms/security/dependabot/32


ulliholtgrave (approved):
Lgtm :)

pyproject.toml
```diff
@@ -121,66 +121,68 @@ pinned = [
...
     "multidict==6.0.4",
-    "natsort==8.3.1",
+    "natsort==8.4.0",
     "oscrypto==1.3.0",
     "parso==0.8.3",
     "pexpect==4.8.0",
     "pickleshare==0.7.5",
-    "Pillow==9.5.0",
-    "prompt-toolkit==3.0.38",
+    "Pillow==10.0.0",
```
timoludwig, line 161: The update of the pillow library seems to break the media upload:
```
  File "/home/integreat/integreat-cms/integreat_cms/cms/utils/media_utils.py", line 53, in generate_thumbnail
    image.thumbnail((size, size), resample=Image.ANTIALIAS)
                                           ^^^^^^^^^^^^^^^
AttributeError: module 'PIL.Image' has no attribute 'ANTIALIAS'
```

This is another example why #1997 is needed :sweat_smile:

timoludwig: Ok, this was an easy fix :laughing: 
https://stackoverflow.com/questions/76616042/attributeerror-module-pil-image-has-no-attribute-antialias

MizukiTemma (approved):
Now media library works too. Thank you 👍

Issue #2107, opened 2023-03-06
Implement the new dashboard 


JoeyStk:
### Motivation
We want to include a dashboard that gives an overview over todos for the municipalities. 

### Proposed Solution

Implement the design you can find following the link to Figma (or in the screenshot). 

- The link for "Seiten, die zum Review vorliegen" should be [this](cms.integreat-app.de/%region%/pages/%language%/?query=&status=REVIEW&date_from=&date_to=&translation_status=UP_TO_DATE&translation_status=IN_TRANSLATION&translation_status=OUTDATED&translation_status=MISSING&translation_status=MACHINE_TRANSLATED) (pages filtered by status)
- The link for "Unglesenes Feedback" should be [this](https://admin.integreat-app.de/testumgebung/feedback/?query=&category=&language=&date_from=&date_to=&read_status=UNREAD&rating=True&rating=False&rating=&filter=)(feedback filtered by read_status)
- The link for "Fehlerhafte Links" should be a link to the broken link section
- "Seiten mit niedrigem HIX-Wert" needs one extra step:
  - Create a new box in the analysis > translation report section. Move the HIX section (s. screenshot) from the current dashboard there. 
![grafik](https://github.com/digitalfabrik/integreat-cms/assets/72705147/703e2f9c-b62f-4555-96ba-2646f6abfa53)
  - Link to analysis > translation report
-  The link for "Veraltete Seiten" should be to pages pre-filtered for pages that are older than 365 days

### Alternatives

None

### Additional Context
(For the implementation this can be ignored for now. Implementing the five tasks as seen in the screenshot is enough for the first iteration)
The following activities could be displayed and completed by the municipalities:

- unread feedback
- Faulty links
- Outdated pages (from 365 days, but can also be done earlier)
-  Pages awaiting revision (always display first)
-  Too large images or pages with too much content
-  Pictures without alternative texts --> maintain texts
- POIs without categories
-  Pages with HIX value below 10
-  Inactive users who have not logged in for 1 year
-  Missing icons on the non-bottom level

optionally expandable at some point:

- Possible duplicates for POIs
- Outdated translations that could also be made DeepL if budget is available

What should not go in:
- Live content that has changed at some point

### Design Requirements

@hauf-toni

Link to Figma: https://www.figma.com/file/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=518%3A3671 - please check for the project board with the according issue id.

[...]

PR #2348, opened 2023-07-17
Add new todo dashboard

JoeyStk:
### Short description
This PR adds the new todo dashboard to the dashboard. It takes the place of the HIX analysis box. The HIX analysis box is instead moved to the translation coverage section.

### Proposed changes

- Add new todo dashboard box
- Add template for todos
- Add six todos for the first version of the todo dashboard
- Move the HIX analysis to translation coverage

### Side effects

I think none

### Resolved issues

Fixes: #2107

### DIFF
--- a/integreat_cms/cms/models/pages/page.py
+++ b/integreat_cms/cms/models/pages/page.py
@@ -224,9 +224,10 @@ class Page(AbstractTreeNode, AbstractBasePage):
         """
         This returns all of the page's ancestors which are archived.
 
-        :return: The QuerySet of archived ancestors
+        :return: The list of archived ancestors
         :rtype: list [ ~integreat_cms.cms.models.pages.page.Page ]
         """
+
         return [
             ancestor
             for ancestor in self.get_cached_ancestors()
--- a/integreat_cms/cms/models/regions/region.py
+++ b/integreat_cms/cms/models/regions/region.py
@@ -631,31 +631,47 @@ class Region(AbstractBaseModel):
             raise Http404("No language matches the given query.") from e
 
     @cached_property
+    def explicitly_archived_ancestors_subquery(self):
+        """
+        This property returns a subquery for all explicitly archived anchestors of a given page.
+        Needs to be used as part of another query because in order to resolve :class:`~django.db.models.OuterRef`,
+        e.g. in a :class:`~django.db.models.Subquery` or in :class:`~django.db.models.Exists`.
+
+        :return: A queryset of the explicitly archived ancestors.
+        :rtype: ~django.db.models.query.QuerySet
+        """
+
+        return self.pages.filter(
+            tree_id=models.OuterRef("tree_id"),
+            lft__lt=models.OuterRef("lft"),
+            rgt__gt=models.OuterRef("rgt"),
+            explicitly_archived=True,
+        )
+
+    @cached_property
     def archived_pages(self):
         """
         This property returns a QuerySet of all archived pages and their descendants of this region.
 
-        Per default, the returned queryset has some limitations because of the usage of
-        :meth:`~django.db.models.query.QuerySet.union`. To perform the extra effort of returning an unrestricted
-        queryset, use :meth:`~integreat_cms.cms.models.regions.region.Region.get_pages` with the parameters ``archived`` and
-        ``return_unrestricted_queryset`` set to ``True``.
-
         :return: A QuerySet of all archived pages of this region
         :rtype: ~treebeard.ns_tree.NS_NodeQuerySet [ ~integreat_cms.cms.models.pages.page.Page ]
         """
-        # Queryset of explicitly archived pages
-        explicitly_archived_pages = self.pages.filter(explicitly_archived=True)
-        # Multiple order_by clauses are not allowed in sql queries, so to make combined queries with union() work,
-        # we have to remove ordering from the input querysets and apply the default ordering to the resulting queryset.
-        explicitly_archived_pages = explicitly_archived_pages.order_by()
-        # List of QuerySets of descendants of archived pages
-        implicitly_archived_pages = [
-            page.get_descendants().order_by() for page in explicitly_archived_pages
-        ]
-        # Merge explicitly and implicitly archived pages
-        archived_pages = explicitly_archived_pages.union(*implicitly_archived_pages)
-        # Order the resulting :class:`~treebeard.ns_tree.NS_NodeQuerySet` to restore the tree-structure
-        return archived_pages.order_by("tree_id", "lft")
+
+        return self.pages.filter(
+            id=models.Case(
+                models.When(
+                    models.Exists(
+                        self.explicitly_archived_ancestors_subquery.values("pk")
+                    ),
+                    then=models.F("pk"),
+                ),
+                models.When(
+                    explicitly_archived=True,
+                    then=models.F("pk"),
+                ),
+                default=None,
+            )
+        )
 
     @cached_property
     def non_archived_pages(self):
@@ -664,27 +680,26 @@ class Region(AbstractBaseModel):
         A page is considered as "non-archived" if its ``explicitly_archived`` property is ``False`` and all the
         page's ancestors are not archived as well.
 
-        Per default, the returned queryset has some limitations because of the usage of
-        :meth:`~django.db.models.query.QuerySet.difference` (see :meth:`~django.db.models.query.QuerySet.union` for some
-        restrictions). To perform the extra effort of returning an unrestricted queryset, use
-        :meth:`~integreat_cms.cms.models.regions.region.Region.get_pages` with the parameter
-        ``return_unrestricted_queryset`` set to ``True``.
-
         :return: A QuerySet of all non-archived pages of this region
         :rtype: ~treebeard.ns_tree.NS_NodeQuerySet [ ~integreat_cms.cms.models.pages.page.Page ]
         """
-        # Multiple order_by clauses are not allowed in sql queries, so to make combined queries with difference() work,
-        # we have to remove ordering from the input querysets and apply the default ordering to the resulting queryset.
-        archived_pages = self.archived_pages.order_by()
-        # Exclude archived pages from all pages
-        non_archived_pages = self.pages.difference(archived_pages)
-        # Order the resulting TreeQuerySet to restore the tree-structure
-        return non_archived_pages.order_by("tree_id", "lft")
+
+        return self.pages.filter(
+            id=models.Case(
+                models.When(
+                    models.Exists(
+                        self.explicitly_archived_ancestors_subquery.values("pk")
+                    ),
+                    then=None,
+                ),
+                default=models.F("pk"),
+            ),
+            explicitly_archived=False,
+        )
 
     def get_pages(
         self,
         archived=False,
-        return_unrestricted_queryset=False,
         prefetch_translations=False,
         prefetch_public_translations=False,
         annotate_language_tree=False,
@@ -694,17 +709,8 @@ class Region(AbstractBaseModel):
         To retrieve all pages independently of their archived-state, use the reverse foreign key
         :attr:`~integreat_cms.cms.models.regions.region.Region.pages`.
 
-        Per default, the returned queryset has some limitations because of the usage of
-        :meth:`~django.db.models.query.QuerySet.difference` and :meth:`~django.db.models.query.QuerySet.union`.
-        To perform the extra effort of returning an unrestricted queryset, set the parameter
-        ``return_unrestricted_queryset`` to ``True``.
-
         :param archived: Whether or not only archived pages should be returned (default: ``False``)
         :type archived: bool
-
-        :param return_unrestricted_queryset: Whether or not the result should be returned as unrestricted queryset.
-                                             (default: ``False``)
-        :type return_unrestricted_queryset: bool
 
         :param prefetch_translations: Whether the latest translations for each language should be prefetched
                                       (default: ``False``)
@@ -722,14 +728,6 @@ class Region(AbstractBaseModel):
         :rtype: ~treebeard.ns_tree.NS_NodeQuerySet [ ~integreat_cms.cms.models.pages.page.Page ]
         """
         pages = self.archived_pages if archived else self.non_archived_pages
-        if (
-            return_unrestricted_queryset
-            or prefetch_translations
-            or prefetch_public_translations
-        ):
-            # Generate a new unrestricted queryset containing the same pages
-            page_ids = [page.id for page in pages]
-            pages = self.pages.filter(id__in=page_ids)
         if prefetch_translations:
             pages = pages.prefetch_translations()
         if prefetch_public_translations:
@@ -756,7 +754,7 @@ class Region(AbstractBaseModel):
         :param query: The query string used for filtering the regions
         :type query: str
         :return: A query for all matching objects
-        :rtype: ~django.db.models.QuerySet
+        :rtype: ~django.db.models.query.QuerySet
         """
         return cls.objects.filter(name__icontains=query)
 
--- a/integreat_cms/cms/templates/dashboard/_todo_dashboard_row.html
+++ b/integreat_cms/cms/templates/dashboard/_todo_dashboard_row.html
@@ -1,0 +1,26 @@
+{% load i18n %}
+<div class="flex justify-between border-t-2 border-zinc-600 p-4 w-full">
+    <div class="flex items-center">
+        <div class="mr-6">
+            <i icon-name="{% spaceless %}{% block todo_dashboard_icon %}{% endblock todo_dashboard_icon %}{% endspaceless %}"></i>
+        </div>
+        <div>
+            <a href="{% spaceless %}{% block todo_dashboard_title_link %}{% endblock todo_dashboard_title_link %}{% endspaceless %}"
+               class="text-blue-500 font-bold underline">
+                {% block todo_dashboard_title %}
+                {% endblock todo_dashboard_title %}
+            </a>
+            {% block todo_dashboard_number %}
+                <span>({% translate "in total" %} {{ total }})</span>
+            {% endblock todo_dashboard_number %}
+            <p>
+                {% block todo_dashboard_description %}
+                {% endblock todo_dashboard_description %}
+            </p>
+        </div>
+    </div>
+    <div class="flex items-center justify-end w-80 text-center">
+        {% block todo_dashboard_button_link %}
+        {% endblock todo_dashboard_button_link %}
+    </div>
+</div>
--- a/integreat_cms/cms/templates/dashboard/_todo_dashboard_widget.html
+++ b/integreat_cms/cms/templates/dashboard/_todo_dashboard_widget.html
@@ -1,0 +1,33 @@
+{% extends "_collapsible_box.html" %}
+{% load i18n %}
+{% block collapsible_box_icon %}
+    clipboard-list
+{% endblock collapsible_box_icon %}
+{% block collapsible_box_title %}
+    {% translate "To-Do's" %}
+{% endblock collapsible_box_title %}
+{% block collapsible_box_content %}
+    <div class="flex flex-wrap gap-2">
+        <p class="pt-2 px-4">
+            {% blocktranslate trimmed %}
+                Every day you can find a list of ideas on how you can improve the content of your pages
+            {% endblocktranslate %}
+        </p>
+        {% if perms.cms.change_page %}
+            {% include "dashboard/todo_dashboard_rows/_unreviewed_pages_row.html" %}
+            {% include "dashboard/todo_dashboard_rows/_automatically_saved_pages_row.html" %}
+        {% endif %}
+        {% if perms.cms.change_feedback %}
+            {% include "dashboard/todo_dashboard_rows/_unread_feedback_row.html" %}
+        {% endif %}
+        {% if perms.cms.change_page and perms.cms.view_broken_links %}
+            {% include "dashboard/todo_dashboard_rows/_broken_links_row.html" %}
+        {% endif %}
+        {% if perms.cms.change_page and TEXTLAB_API_ENABLED and request.region.hix_enabled %}
+            {% include "dashboard/todo_dashboard_rows/_low_hix_row.html" %}
+        {% endif %}
+        {% if perms.cms.change_page %}
+            {% include "dashboard/todo_dashboard_rows/_outdated_pages_row.html" %}
+        {% endif %}
+    </div>
+{% endblock collapsible_box_content %}
--- a/integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_automatically_saved_pages_row.html
+++ b/integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_automatically_saved_pages_row.html
@@ -1,0 +1,37 @@
+{% extends "../_todo_dashboard_row.html" %}
+{% load i18n %}
+{% block todo_dashboard_icon %}
+    save
+{% endblock todo_dashboard_icon %}
+{% block todo_dashboard_title_link %}
+    {% url 'pages' region_slug=request.region.slug language_slug=default_language_slug %}?status=AUTO_SAVE
+{% endblock todo_dashboard_title_link %}
+{% block todo_dashboard_title %}
+    {% translate "Automatically saved pages" %}
+{% endblock todo_dashboard_title %}
+{% block todo_dashboard_number %}
+    {% with total=automatically_saved_pages|length %}
+        {{ block.super }}
+    {% endwith %}
+{% endblock todo_dashboard_number %}
+{% block todo_dashboard_description %}
+    {% if automatically_saved_pages %}
+        {% blocktranslate trimmed with single_automatically_saved_page=automatically_saved_pages.0 %}
+            The page <b>{{ single_automatically_saved_page }}</b> has been saved automatically. Please don't forget to publish this page once it's ready.
+        {% endblocktranslate %}
+    {% else %}
+        {% blocktranslate trimmed %}
+            At the moment there are no automatically saved pages. Good job!
+        {% endblocktranslate %}
+    {% endif %}
+{% endblock todo_dashboard_description %}
+{% block todo_dashboard_button_link %}
+    {% if automatically_saved_pages %}
+        <a class="btn !rounded-full"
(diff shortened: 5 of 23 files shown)


integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -4646,10 +4646,144 @@ msgstr "Seite neu laden"
 msgid "Continue reading"
 msgstr "Weiterlesen"
 
+#: cms/templates/dashboard/_todo_dashboard_widget.html
+msgid "To-Do's"
+msgstr "To-Do’s"
```
timoludwig, line 4651:
```suggestion
msgid "To-Do's"
msgstr "To-Dos"
```
To be fair, this was already incorrect in the design :sweat_smile:
https://www.duden.de/rechtschreibung/To_do
http://www.deppenapostroph.info/

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -42,51 +46,176 @@ def get_context_data(self, **kwargs):
...
+                    last_updated__lte=OUTDATED_THRESHOLD_DATE.date(),
+                    page__explicitly_archived=False,
+                    language__slug=self.request.region.default_language.slug)
+                .order_by("page__id", "language__id", "-version", "last_updated")
+                .distinct("page__id", "language__id")
+                .all()
+        )
+        print(OUTDATED_THRESHOLD_DATE, outdated_pages[0].last_updated)
+        for p in outdated_pages:
+            # TODO: why does last_updated sometimes differ??
```
JoeyStk, line 156: @david-venhoff This is the line we spoke about on Tuesday. If you checkout this branch and go to the dashboard view, you'll see in brackets the amount of pages that fit the criteria of being outdated (being older than 365 days). If you filter for the same criteria in the page tree (or click on the title link "outdated page") you might notice, that the amount of pages differ. In my case for 1088 pages the dashboard shows 24 outdated pages in the dashboard, but only 13 in the filtered page tree view

integreat_cms/cms/templates/dashboard/_todo_dashboard_widget.html
```diff
@@ -0,0 +1,25 @@
...
+    <div class="flex flex-wrap gap-2">
+        <p>
+            {% blocktranslate trimmed %}
+                Every day you can find a list of ideas on how you can improve the content of your pages
+            {% endblocktranslate %}
+        </p>
+        {% include "dashboard/todo_dashboard_rows/_unreviewed_pages_row.html" %}
+        {% include "dashboard/todo_dashboard_rows/_automatically_saved_pages_row.html" %}
+        {% include "dashboard/todo_dashboard_rows/_unread_feedback_row.html" %}
+        {% include "dashboard/todo_dashboard_rows/_broken_links_row.html" %}
```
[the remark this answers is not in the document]
JoeyStk: After a chat with @osmers it was decided to remove this section

integreat_cms/cms/templates/dashboard/_todo_dashboard_row.html
```diff
@@ -0,0 +1,24 @@
...
+    <div>
+        <a href="{% spaceless %}{% block todo_dashboard_title_link %}{% endblock todo_dashboard_title_link %}{% endspaceless %}"
+           class="text-blue-500 font-bold underline">
+            {% block todo_dashboard_title %}
+            {% endblock todo_dashboard_title %}
+        </a>
+        <span>&nbsp;(in total
+            {% block todo_dashboard_number %}
+            {% endblock todo_dashboard_number %}
+        )</span>
```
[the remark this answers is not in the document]
timobrembeck: This is correct, unfortunately the html spec has a few limitations here. But in my opinion, the UI that is visible to the user is more important than code style. Improvements to the code style are only desirable if they don't influence the layout.

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -42,51 +47,189 @@ def get_context_data(self, **kwargs):
...
+        r"""
+        Extend context by info on automatically saved pages
+
+        :return: Dictionary containing the context for auto saved pages
+        :rtype: dict
+        """
+        automatically_saved_pages = (
+            PageTranslation.objects.filter(
+                status=status.AUTO_SAVE,
+                page__explicitly_archived=False,
```
david-venhoff, line 219: It would probably also make sense to exclude implicitely archived pages here.

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -42,51 +47,189 @@ def get_context_data(self, **kwargs):
...
+            PageTranslation.objects.filter(
+                status=status.REVIEW,
+                page__explicitly_archived=False,
+                language__slug=self.request.region.default_language.slug,
+                page__id__in=Subquery(not_implicitly_archived_ids.values("pk")),
+            )
+            .order_by("page__id", "language__id", "-version")
+            .distinct("page__id", "language__id")
+            .all()
         )
```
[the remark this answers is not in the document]
david-venhoff: I mixed up the draft and review statuses

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -42,51 +47,189 @@ def get_context_data(self, **kwargs):
...
+        :rtype: dict
+        """
+        automatically_saved_pages = (
+            PageTranslation.objects.filter(
+                status=status.AUTO_SAVE,
+                page__explicitly_archived=False,
+                language__slug=self.request.region.default_language.slug,
+            )
+            .order_by("page__id", "language__id", "-version")
+            .distinct("page__id", "language__id")
```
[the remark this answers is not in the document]
PeterNerlich: I think this was required, but I don't remember how exactly it went wrong without. Anyway, this should not be a performance impact even if it should not be required, so we're being lazy and will just leave it in :upside_down_face:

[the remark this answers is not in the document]
timobrembeck: If you order by `language` instead of `language__id`, it will take the default ordering of the language model (the `bcp47_tag`):
https://github.com/digitalfabrik/integreat-cms/blob/fc529eaf8ccecbe21cfdff267c20dec07fbdfb70/integreat_cms/cms/models/languages/language.py#L215
Which means it needs to query the foreign object in order to perform the query. If we order by the language's `id` however, Django doesn't need to query this additional information from the db since the language id is already stored in the translation model.

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -42,51 +47,189 @@ def get_context_data(self, **kwargs):
...
+        :rtype: dict
+        """
+        OUTDATED_THRESHOLD_DATE = datetime.now() - relativedelta(
+            days=settings.OUTDATED_THRESHOLD_DAYS
+        )
+
+        outdated_pages = (
+            PageTranslation.objects.filter(
+                last_updated__lte=OUTDATED_THRESHOLD_DATE.date(),
+                page__explicitly_archived=False,
```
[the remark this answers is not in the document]
timobrembeck: @PeterNerlich Why did you resolve this conversation? I don't see how **im**plicitly archived pages are excluded in this query...

integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_automatically_saved_pages_row.html
```diff
@@ -0,0 +1,36 @@
+{% extends "../_todo_dashboard_row.html" %}
+{% load i18n %}
+{% block todo_dashboard_icon %}
+    save
+{% endblock todo_dashboard_icon %}
+{% block todo_dashboard_title_link %}
+    {% url 'pages' region_slug=request.region.slug language_slug=default_language_slug %}
```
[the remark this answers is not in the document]
timobrembeck: > Another (probably not very important) problem is that there seems to be no way to get rid of an autosave that should not get published.

@david-venhoff what do you mean by this? You can reject unwanted autosaves in the version view, right? Then it will revert to the latest public version which should result in the autosave no longer being present on the dashboard.

integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_unreviewed_pages_row.html
```diff
@@ -0,0 +1,36 @@
...
+    {% else %}
+        {% blocktranslate trimmed %}
+            At the moment there is no page waiting for a review. Good job!
+        {% endblocktranslate %}
+    {% endif %}
+{% endblock todo_dashboard_description %}
+{% block todo_dashboard_button_link %}
+    {% if unreviewed_pages %}
+        <a class="btn !rounded-full"
+           href="{% url 'edit_page' region_slug=request.region.slug language_slug=default_language_slug page_id=single_unreviewed_page.page_id %}">
```
timobrembeck, line 30:
```suggestion
           href="{% url 'edit_page' region_slug=request.region.slug language_slug=default_language_slug page_id=unreviewed_pages.0.page_id %}">
```
No need to define a new variable just to access the first element

integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_unreviewed_pages_row.html
```diff
@@ -0,0 +1,36 @@
...
+    {% translate "Pages that are waiting for review" %}
+{% endblock todo_dashboard_title %}
+{# djlint:off #}
+{% block todo_dashboard_number %}
+    {{ unreviewed_pages|length }}{% endblock todo_dashboard_number %}
+{# djlint:on #}
+{% block todo_dashboard_description %}
+    {% if unreviewed_pages %}
+        {% blocktranslate trimmed %}
+            The page <b>{{ single_unreviewed_page }}</b> is waiting for a review. Please review, if the content on the page is correct.
```
timobrembeck, line 19:
```suggestion
        {% blocktranslate trimmed with single_unreviewed_page=unreviewed_pages.0 %}
            The page <b>{{ single_unreviewed_page }}</b> is waiting for a review. Please review, if the content on the page is correct.
```

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -42,51 +47,192 @@ def get_context_data(self, **kwargs):
...
+            .distinct("page__id", "language__id")
+            .all()
         )
 
-        # Get the latest versions of the page translations for these pages
-        hix_translations = PageTranslation.objects.filter(
-            language__slug__in=settings.TEXTLAB_API_LANGUAGES, page__in=hix_pages
-        ).distinct("page_id", "language_id")
+        single_unreviewed_page = unreviewed_pages[0] if unreviewed_pages else None
+
```
timobrembeck, line 97:
```suggestion
```

integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_automatically_saved_pages_row.html
```diff
@@ -0,0 +1,36 @@
...
+{% block todo_dashboard_title %}
+    {% translate "Automatically saved pages" %}
+{% endblock todo_dashboard_title %}
+{# djlint:off #}
+{% block todo_dashboard_number %}
+    {{ automatically_saved_pages|length }}{% endblock todo_dashboard_number %}
+{# djlint:on #}
+{% block todo_dashboard_description %}
+    {% if automatically_saved_pages %}
+        {% blocktranslate trimmed %}
```
timobrembeck, line 18:
```suggestion
        {% blocktranslate trimmed with single_automatically_saved_page=automatically_saved_pages.0 %}
```

integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_automatically_saved_pages_row.html
```diff
@@ -0,0 +1,36 @@
...
+    {% else %}
+        {% blocktranslate trimmed %}
+            At the moment there are no automatically saved pages. Good job!
+        {% endblocktranslate %}
+    {% endif %}
+{% endblock todo_dashboard_description %}
+{% block todo_dashboard_button_link %}
+    {% if automatically_saved_pages %}
+        <a class="btn !rounded-full"
+           href="{% url 'edit_page' region_slug=request.region.slug language_slug=default_language_slug page_id=single_automatically_saved_page.page_id %}">
```
timobrembeck, line 30:
```suggestion
           href="{% url 'edit_page' region_slug=request.region.slug language_slug=default_language_slug page_id=automatically_saved_pages.0.page_id %}">
```

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -42,51 +47,192 @@ def get_context_data(self, **kwargs):
...
+            .all()
         )
 
-        # Get the number of translations which are not ready for MT
-        not_ready_for_mt_count = sum(
-            pt.hix_score < settings.HIX_REQUIRED_FOR_MT
-            for pt in hix_translations_with_score
+        single_automatically_saved_page = (
+            automatically_saved_pages[0] if automatically_saved_pages else None
         )
```
timobrembeck, line 232:
```suggestion
```

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -42,51 +47,192 @@ def get_context_data(self, **kwargs):
...
 
         return {
-            "worst_hix_translations": worst_hix_translations,
-            "hix_threshold": settings.HIX_REQUIRED_FOR_MT,
-            "ready_for_mt_count": len(hix_translations) - not_ready_for_mt_count,
-            "total_count": len(hix_translations),
+            "automatically_saved_pages": automatically_saved_pages,
+            "single_automatically_saved_page": single_automatically_saved_page,
```
timobrembeck, line 236:
```suggestion
```

integreat_cms/core/settings.py
```diff
@@ -1093,3 +1093,10 @@
 
 # Our database operations should never exceed 60 seconds
 DB_MUTEX_TTL_SECONDS = 60
+
+#############
+# Dashboard #
+#############
+
+# Days since a page is considered to be outdated in the todo dashboard
```
timobrembeck, line 1101:
```suggestion
#: Days after which a page is considered to be outdated in the todo dashboard
```

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -42,51 +47,192 @@ def get_context_data(self, **kwargs):
...
+
+        :return: Dictionary containing the context for broken links
+        :rtype: dict
+        """
+        invalid_urls = filter_urls(self.request.region.slug, "invalid")[0]
+        invalid_url = invalid_urls[0] if invalid_urls else None
+
+        relevant_translation = (
+            invalid_url.region_links[0].content_object if invalid_url else None
+        )
```
timobrembeck, line 203: This cannot work, because `Url` objects are different from `Link` objects. The `Url` only contains the reference to the `Link` objects, and only the `Link` objects references the translation objects.

integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_broken_links_row.html
```diff
@@ -0,0 +1,36 @@
...
+    {% else %}
+        {% blocktranslate trimmed %}
+            At the moment there are no broken links. Good job!
+        {% endblocktranslate %}
+    {% endif %}
+{% endblock todo_dashboard_description %}
+{% block todo_dashboard_button_link %}
+    {% if broken_links %}
+        <a class="btn !rounded-full"
+           href="{% url 'edit_url' region_slug=request.region.slug url_filter='invalid' url_id=relevant_url.page_id %}#replace-url">
```
timobrembeck, line 30: This cannot work, the `Url` model does not have a `page_id` property or field.

integreat_cms/cms/views/dashboard/dashboard_view.py
timobrembeck: @PeterNerlich Here as well, you resolved the conversation, but the issue does not seem to be resolved?

integreat_cms/cms/templates/dashboard/_todo_dashboard_row.html
```diff
@@ -0,0 +1,28 @@
...
+            <a href="{% spaceless %}{% block todo_dashboard_title_link %}{% endblock todo_dashboard_title_link %}{% endspaceless %}"
+               class="text-blue-500 font-bold underline">
+                {% block todo_dashboard_title %}
+                {% endblock todo_dashboard_title %}
+            </a>
+            {# djlint:off #}
+            <span>({% translate "in total" %}
+                {% block todo_dashboard_number %}
+                {% endblock todo_dashboard_number %})</span>
+            {# djlint:on #}
```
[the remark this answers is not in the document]
timobrembeck: @JoeyStk this is not yet resolved, you only implemented the suggested changes of the super template, but not in the sub templates. This means that you completely override the block content, because you didn't include the `{{ block.super }}` parts in the sub templates.

timobrembeck (changes requested):

integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_automatically_saved_pages_row.html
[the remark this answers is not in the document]
david-venhoff: Ah, I was not aware of that. Then you can ignore this point ^^

integreat_cms/cms/views/dashboard/dashboard_view.py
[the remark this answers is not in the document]
PeterNerlich: @timobrembeck Because Johannes and I went over this today. The changes are just not pushed yet. We will continue work on this tomorrow morning.

integreat_cms/cms/views/dashboard/dashboard_view.py
PeterNerlich: Because Johannes and I went over this today. The changes are just not pushed yet. We will continue work on this tomorrow morning.

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -42,51 +47,192 @@ def get_context_data(self, **kwargs):
...
+
+        not_implicitly_archived_ids = Page.objects.filter(
+            id=Case(
+                When(
+                    Exists(explicitly_archived_ancestors),
+                    then=None,
+                ),
+                default=F("pk"),
+            )
+        )
```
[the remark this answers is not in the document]
PeterNerlich: @timobrembeck I played around with this for a bit and the tests don't seem to break overall. Except for the PDF tests because they always make issues, and `test_sitemap` because the query count went down by one:

image

Should I add this as a separate PR rather than as part of this one? If so, this one should depend on it and only get merged after it, correct?

integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_unreviewed_pages_row.html
PeterNerlich: Do you have any preference of 
```jinja
        {% blocktranslate trimmed with single_unreviewed_page=unreviewed_pages.0 %}
            The page <b>{{ single_unreviewed_page }}</b> is waiting for a review. Please review, if the content on the page is correct.
```
over
```jinja
        {% blocktranslate trimmed %}
            The page <b>{{ unreviewed_pages.0 }}</b> is waiting for a review. Please review, if the content on the page is correct.
```
or is that just because that variable was used earlier?

integreat_cms/cms/views/dashboard/dashboard_view.py
PeterNerlich: But isn't that exactly what the line for `relevant_translation` does? It picks the first region link for an URL and selects its `content_object`

integreat_cms/cms/views/dashboard/dashboard_view.py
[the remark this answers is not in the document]
PeterNerlich: I committed my suggestion directly to this branch for now.

integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_unreviewed_pages_row.html
timobrembeck: Exactly, I think in general it's better to avoid the definition of new variable names for trivial stuff, but in the translation files it's preferable to have more speaking variable names to get the context for the translation.

integreat_cms/cms/views/dashboard/dashboard_view.py
timobrembeck: Ah damn, I'm sorry, I guess I missed the `region_links[0]` part somehow :see_no_evil:

integreat_cms/cms/models/pages/page.py
```diff
@@ -216,14 +216,11 @@ def explicitly_archived_ancestors(self):
...
+        :return: The list of archived ancestors
         :rtype: list [ ~integreat_cms.cms.models.pages.page.Page ]
         """
-        return [
-            ancestor
-            for ancestor in self.get_cached_ancestors()
-            if ancestor.explicitly_archived
-        ]
+
+        return self.region.explicitly_archived_ancestors_query(fixed=self)
```
timobrembeck, line 223:
```suggestion
        return [
            ancestor
            for ancestor in self.get_cached_ancestors()
            if ancestor.explicitly_archived
        ]
```
Hmm, I'd like to revert this change since it bypasses the cached ancestors and might cause additional queries in some circumstances.

integreat_cms/cms/models/regions/region.py
```diff
@@ -630,32 +630,73 @@ def get_language_or_404(self, language_slug, only_active=False, only_visible=Fal
...
+            rgt = fixed.rgt
+        else:
+            tree = models.OuterRef("tree_id")
+            lft = models.OuterRef("lft")
+            rgt = models.OuterRef("rgt")
+
+        return self.pages.filter(
+            tree_id=tree,
+            lft__lt=lft,
+            rgt__gt=rgt,
```
timobrembeck, line 672:
```suggestion
        return self.pages.filter(
            tree_id=models.OuterRef("tree_id"),
            lft__lt=models.OuterRef("lft"),
            rgt__gt=models.OuterRef("rgt"),
```
I think one method for the subquery is enough here.

integreat_cms/cms/models/regions/region.py
```diff
@@ -694,16 +736,15 @@ def get_pages(
...
-        To perform the extra effort of returning an unrestricted queryset, set the parameter
-        ``return_unrestricted_queryset`` to ``True``.
-
         :param archived: Whether or not only archived pages should be returned (default: ``False``)
         :type archived: bool
 
         :param return_unrestricted_queryset: Whether or not the result should be returned as unrestricted queryset.
                                              (default: ``False``)
+
+                                             .. deprecated:: 2023.10.0
```
timobrembeck, line 745: Hmm, I don't think we need deprecation grace periods for our functions, just removing them without notice is fine I guess.
Also, please remove all usages of the `return_unrestricted_queryset` parameter since it's not needed anymore.

integreat_cms/cms/views/dashboard/dashboard_view.py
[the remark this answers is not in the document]
timobrembeck: @PeterNerlich wow, that's really cool, awesome improvement! :muscle:
It's fine to include this change in this PR, however could you clean up the git history a bit to have one commit for the dashboard changes and one commit for the refactoring of the model methods?

timobrembeck (changes requested):
Thanks a lot! :muscle: 
The performance problems definitely have been resolved! :tada:
I'd just change a small implementation detail :wink:

timobrembeck (changes requested):
Great, thanks! :+1:

@PeterNerlich there is still one occurrence of the outdated argument:

https://github.com/digitalfabrik/integreat-cms/blob/0c742982837e51281b5a86a5d9c3b04b38876036/integreat_cms/cms/views/regions/region_update_view.py#L67

This will cause an error when archiving regions via the form. It's unfortunate that we don't have a test case for this, maybe we should add one? We could add a copy of this test case:

https://github.com/digitalfabrik/integreat-cms/blob/0c742982837e51281b5a86a5d9c3b04b38876036/tests/cms/views/view_config.py#L464-L479

just with `ARCHIVED` instead of `ACTIVE`, this should then correctly trigger the `if` condition and reveal the error.

PeterNerlich:
Memo to ourselves: Notify marketing team in time before this is released

timobrembeck:
@PeterNerlich Then we should probably notify the team now? I think this PR is ready to be merged soon and will probably be released next week?

JoeyStk:
@timobrembeck  & @PeterNerlich 
> @PeterNerlich Then we should probably notify the team now? I think this PR is ready to be merged soon and will probably be released next week?

I finally finished and uploaded the blog post for the to do board and it's currently being spell-checked and proof read by the marketing team. As soon as this is finished, we can release this PR :smiley: 

However I still have one question for @timobrembeck. At the conference we discussed that it's always desirable to load the dashboard via AJAX to make sure that in no instance the page doesn't load due to this widget and instead only this singluar widget is not being loaded (but navigation etc. are being loaded). What do you think? Do you agree and we should still implement the AJAX call for this widget, and if so, should we still do it before the release (and maybe even in this PR) or can we postpone this, since currently there is no bulk of database calls?

timobrembeck:
> However I still have one question for @timobrembeck. At the conference we discussed that it's always desirable to load the dashboard via AJAX to make sure that in no instance the page doesn't load due to this widget and instead only this singular widget is not being loaded (but navigation etc. are being loaded). What do you think? Do you agree and we should still implement the AJAX call for this widget, and if so, should we still do it before the release (and maybe even in this PR) or can we postpone this, since currently there is no bulk of database calls?

Thanks to @PeterNerlich's optimizations I think AJAX-calls are optional here, I did not experience longer page loading times with this PR.
After merging this to develop, we can do a bit more testing with (nearly) live data in the test cms, and if no issues arise, I think it's fine to release it in this state and implement the ajax calls later.

PeterNerlich:
Reminded me that I didn't fix the last bit @timobrembeck  mentioned yet. Done now!

timobrembeck (approved):
Thanks a lot! :+1:
I don't have time for a detailed review at the moment, but I don't see major blockers for this PR. Let's just merge it :shrug:

david-venhoff (approved):
I just did a quick review, but so far everything looks good 🎉

Issue #2046, opened 2023-02-01
No graceful error message when email authentication fails


timoludwig:
### Describe the Bug

When the email credentials in the config are not working, all functions which involve sending emails are not failing gracefully, but showing an internal server error instead.

### Steps to Reproduce

1. Input wrong email config
2. Resend e.g. an invitation link
3. See error

### Expected Behavior

A nice readable error message should be shown which explains the reason

### Actual Behavior

An internal server error occurs

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
Feb 01 09:57:40 ERROR   django.request - 500 Internal Server Error: /lkgoerlitz/users/1264/resend-activation-link/
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/decorators/http.py", line 40, in inner
    return func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/views/users/region_user_actions.py", line 86, in resend_activation_link_region
    send_welcome_mail(request, user, activation=True)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/utils/welcome_mail_utils.py", line 57, in send_welcome_mail
    send_mail(
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/utils/email_utils.py", line 73, in send_mail
    email.send()
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/mail/message.py", line 284, in send
    return self.get_connection(fail_silently).send_messages([self])
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/mail/backends/smtp.py", line 102, in send_messages
    new_conn_created = self.open()
[...]

PR #2349, opened 2023-07-17
Add graceful error messages when sending mail

PeterNerlich:
### Short description

Display a graceful error message instead of Internal Server Error when failing to send mail.

### Proposed changes

- Catch multiple types of exceptions
  - *BadHeaderError (was already there, now giving slightly more context)*
  - SMTPAuthenticationError (special SMTPException)
  - SMTPException
  - ConnectionRefusedError
- Every exception gets the same base message that already existed, with a brief hint on what the actual problem is. This might make them more content (they get a feeling of knowing what's wrong, not only that something is wrong) and possibly speed up the investigation process when they report it.

### Side effects

- Only a few hours ago I claimed I would be focusing on Reviews and not submit any new PRs for now. With this I kinda loose credibility, don't I? :sweat_smile: 
- Code should be fine

### Resolved issues

Fixes: #2046 

### DIFF
--- a/integreat_cms/cms/utils/welcome_mail_utils.py
+++ b/integreat_cms/cms/utils/welcome_mail_utils.py
@@ -2,12 +2,12 @@ import logging
 
 from django.conf import settings
 from django.contrib import messages
-from django.core.mail import BadHeaderError
 from django.utils.encoding import force_bytes
 from django.utils.http import urlsafe_base64_encode
 from django.utils.text import capfirst
 from django.utils.translation import gettext as _
 
+from ..utils.translation_utils import gettext_many_lazy as __
 from .account_activation_token_generator import account_activation_token_generator
 from .email_utils import send_mail
 
@@ -73,9 +73,11 @@ def send_welcome_mail(request, user, activation):
                 debug_mail_type, user.full_user_name
             ),
         )
-    except BadHeaderError as e:
-        logger.exception(e)
+    except RuntimeError as e:
         messages.error(
             request,
-            _("An error occurred! Could not send {}.").format(debug_mail_type),
+            __(
+                _("An error occurred! Could not send {}.").format(debug_mail_type),
+                e,
+            ),
         )
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -7370,6 +7370,22 @@ msgstr ""
 "verändern soll und keine Aktualisierung der Übersetzungen in anderen "
 "Sprachen erfordert."
 
+#: cms/utils/email_utils.py
+msgid "Malformed header data."
+msgstr "Fehlerhafte Header-Daten."
+
+#: cms/utils/email_utils.py
+msgid "Invalid email credentials."
+msgstr "Ungültige E-Mail-Anmeldedaten."
+
+#: cms/utils/email_utils.py
+msgid "An SMTP error occured."
+msgstr "Ein SMTP-Fehler ist aufgetreten."
+
+#: cms/utils/email_utils.py
+msgid "The email server refused the connection."
+msgstr "Der E-Mail-Server hat die Verbindung abgelehnt."
+
 #: cms/utils/internal_link_checker.py
 msgid "Imprint does not exist or is not public in this language"
 msgstr ""
@@ -7468,6 +7484,7 @@ msgid "{} was successfully sent to user {}."
 msgstr "{} wurde erfolgreich an {} gesendet."
 
 #: cms/utils/welcome_mail_utils.py
+#: cms/views/authentication/password_reset_view.py
 msgid "An error occurred! Could not send {}."
 msgstr "Es ist ein Fehler aufgetreten! Die {} konnte nicht versendet werden."
 
@@ -7546,6 +7563,10 @@ msgstr ""
 "Wenn Sie keine E-Mail erhalten, vergewissern Sie sich bitte, dass Sie die "
 "Adresse eingegeben haben, mit der Sie sich registriert haben, und überprüfen "
 "Sie Ihren Spam-Ordner."
+
+#: cms/views/authentication/password_reset_view.py
+msgid "password reset email"
+msgstr "E-Mail zum Zurücksetzen des Passworts"
 
 #: cms/views/authentication/totp_login_view.py
 msgid "Login failed."
@@ -8998,6 +9019,9 @@ msgid "This page belongs to another region ({})."
 msgstr ""
 "Diese Seite konnte nicht importiert werden, da sie zu einer anderen Region "
 "gehört ({})."
+
+#~ msgid "SMTP error."
+#~ msgstr "SMTP-Fehler."
 
 #~ msgid "You can however archive this page."
 #~ msgstr "Sie können diese Seite jedoch archivieren."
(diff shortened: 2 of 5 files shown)


integreat_cms/cms/utils/welcome_mail_utils.py
```diff
@@ -77,5 +81,35 @@ def send_welcome_mail(request, user, activation):
...
         messages.error(
             request,
-            _("An error occurred! Could not send {}.").format(debug_mail_type),
+            __(
+                error_occured,
+                _("Malformed header data."),
+            ),
+        )
+    except SMTPAuthenticationError as e:
+        logger.exception(e)
```
timoludwig, line 90:
```suggestion
        logger.error(e)
```
I think with these specific error messages, we don't need the whole stack trace from the smtp library, just the error is enough.

integreat_cms/cms/utils/welcome_mail_utils.py
```diff
@@ -77,5 +81,35 @@ def send_welcome_mail(request, user, activation):
...
+        messages.error(
+            request,
+            __(
+                error_occured,
+                _("Invalid email credentials."),
+            ),
+        )
+    except SMTPException as e:
+        logger.exception(e)
```
timoludwig, line 99:
```suggestion
        logger.error(e)
```

integreat_cms/cms/utils/welcome_mail_utils.py
```diff
@@ -77,5 +81,35 @@ def send_welcome_mail(request, user, activation):
...
+        messages.error(
+            request,
+            __(
+                error_occured,
+                _("SMTP error."),
+            ),
+        )
+    except ConnectionRefusedError as e:
+        logger.exception(e)
```
timoludwig, line 108:
```suggestion
        logger.error(e)
```

integreat_cms/release_notes/current/unreleased/2046.yml
```diff
@@ -0,0 +1,2 @@
+en: Show graceful error message when failing to send mail
+de: Zeige Fehlermeldung bei Fehlschlag, Mails zu verschicken
```
timoludwig, line 2:
```suggestion
de: Zeige hilfreichere Fehlermeldung, wenn das Senden von E-Mails fehlschlägt
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -7455,13 +7455,29 @@ msgstr "Aktivierungsmail"
...
+
 #: cms/utils/welcome_mail_utils.py
 msgid "{} was successfully sent to user {}."
 msgstr "{} wurde erfolgreich an {} gesendet."
 
 #: cms/utils/welcome_mail_utils.py
-msgid "An error occurred! Could not send {}."
-msgstr "Es ist ein Fehler aufgetreten! Die {} konnte nicht versendet werden."
+msgid "Malformed header data."
+msgstr "Fehlerhafte Kopfdaten."
```
timoludwig, line 7468:
```suggestion
msgstr "Fehlerhafte Header-Daten."
```

timoludwig (changes requested):
Thanks a lot! :+1:

I think it would make more sense to move the `try`-`except` block into [/integreat_cms/cms/utils/email_utils.py](https://github.com/digitalfabrik/integreat-cms/blob/7a8fda13ca5b35a90c79077174335e01c87beb2d/integreat_cms/cms/utils/email_utils.py#L73), because then the error handling could also be used for the password reset.

PeterNerlich:
> I think it would make more sense to move the `try`-`except` block into [/integreat_cms/cms/utils/email_utils.py](https://github.com/digitalfabrik/integreat-cms/blob/7a8fda13ca5b35a90c79077174335e01c87beb2d/integreat_cms/cms/utils/email_utils.py#L73), because then the error handling could also be used for the password reset.

Funny you should mention that, Ingo and I went back and forth about this for a while, but finally opted to leave it one step higher in the call stack because we would have to pass `request` and `debug_mail_type` in there and were unsure whether this is valid in all uses of this utility function. E.g. we could imagine a cron job or management command using that. But you can probably shine a light on whether this poses a problem or not.

timoludwig:
> Funny you should mention that, Ingo and I went back and forth about this for a while, but finally opted to leave it one step higher in the call stack because we would have to pass `request` and `debug_mail_type` in there and were unsure whether this is valid in all uses of this utility function. E.g. we could imagine a cron job or management command using that. But you can probably shine a light on whether this poses a problem or not.

Ah, sorry, I missed that it created quite a few problems to pass the `request` object to the util function (I think `debug_mail_type` is irrelevant since the user knows what type of email they were just sending, so I think it's not required to add that info if it's too complicated). I agree that it makes sense to keep the util function agnostic of the http request.

Ok, now thinking about it, I guess we need to add another `try`/`except` block in the password reset view here:

https://github.com/digitalfabrik/integreat-cms/blob/7a8fda13ca5b35a90c79077174335e01c87beb2d/integreat_cms/cms/views/authentication/password_reset_view.py#L82

something along the lines of:

```
try:
    return super().form_valid(form)
except RuntimeError as e:
    messages.error(request, e)
    return redirect("public:password_reset")
```

Also, to de-duplicate all the different exception messages, I think it would be fine to catch and unify these in the email utils, and just raise one type of error (e.g. `RuntimeError`) there, and then catch the runtime error in both the password reset view and the welcome/activation mail view.

timoludwig:
> I implemented it for the password reset view, but when testing I wondered whether this is actually what we want. The success message always says "if an account for that email exists…", but the error message will only appear if we actually send an email. The relevance of this is probably very low, but I can't help but wonder how the proper solution for this should actually look like

Thanks! This is exactly how I had envisioned it. It is intended that we don't show an error for non-existing emails to make it harder to brute-force registered email addresses.

After fixing the two remaining small suggestions and cleaning up the git history, I think this is ready to be merged! :+1:

PeterNerlich:
> It is intended that we don't show an error for non-existing emails to make it harder to brute-force registered email addresses.

But now one can still brute-force for registered email addresses if there is a system configuration error or the email server is unreachable, since those errors will only occur for actual email send attempts, correct? Though it is not really a regression to the HTTP 500 thrown before, it does now explicitly state that trying to send the email was what caused a problem.

timoludwig:
> But now one can still brute-force for registered email addresses if there is a system configuration error or the email server is unreachable, since those errors will only occur for actual email send attempts, correct?

Ok true, though I think we can live with this :smiling_face_with_tear:

timoludwig (changes requested):
@PeterNerlich could you clean up the git history a bit? Splitting the two commits for welcome email and password reset is fine for me, but the translations and release notes should be part of the commit they are actually referring to.

PeterNerlich:
In that case, it's easiest to squash it all together, since the translations and release notes are also relevant to the password reset form

timoludwig:
> In that case, it's easiest to squash it all together, since the translations and release notes are also relevant to the password reset form

I agree.

timoludwig (approved):
Great, thanks! :tada:

ulliholtgrave (approved):
Thanks, I only looked on the code and didn't provoke the Exceptions. However, I thinks that the differentiation is a great improvement to the current state :+1:

Issue #1547, opened 2022-06-14
Add banner for test systems


timoludwig:
### Motivation

Sometimes, it can be dangerous to mix up the test and prod systems.

### Proposed Solution

- Add new `TEST` setting independently from `DEBUG` mode
- Also support reading this setting from environment variables
- When this setting is `True`, add a test banner at the top of the screen

### Additional Context

Just for reference, e.g. PyPI does it the following way:

https://pypi.org/
![Screenshot 2022-06-14 at 10-42-58 PyPI · The Python Package Index](https://user-images.githubusercontent.com/16021269/173534875-d68e4533-ea2d-47cd-8c57-1a413ea8906a.png)

https://test.pypi.org/
![Screenshot 2022-06-14 at 10-43-15 TestPyPI · The Python Package Index](https://user-images.githubusercontent.com/16021269/173534871-4e50b2e2-9949-4ec5-b08d-8e78cf5c25b2.png)

steffenkleinle:
@timoludwig the text of the banner is not 100% accurate, the content of test cms is not visible in our beta versions. It is only visible in webnext/dev releases of the native apps.

timoludwig:
@steffenkleinle ah ok, that's unfortunate. The banner should reduce confusion, not add to it. Do you want to open an issue and suggest a different wording?

steffenkleinle:
@timoludwig done

PR #2354, opened 2023-07-23
Show test banner in test systems

timoludwig:
### Short description

With the increasing number of test & production systems of our different whitelabel solutions, I think this banner helps to keep the overview.

### Proposed changes

- Show test banner in test systems

### Resolved issues

Fixes: #1547

### DIFF
--- a/example-configs/integreat-cms.ini
+++ b/example-configs/integreat-cms.ini
@@ -8,8 +8,10 @@
 # or rearranged as desired.
 
 [base]
-# Whether to debug mode is enabled [optional, defaults to False]
+# Whether debug mode is enabled [optional, defaults to False]
 DEBUG = False
+# Whether test mode is enabled [optional, defaults to False]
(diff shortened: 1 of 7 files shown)


PeterNerlich:
Would it make sense to add a banner to the management commands on this occasion as well?

timoludwig:
> Would it make sense to add a banner to the management commands on this occasion as well?

No, since management commands are invoked via the shell, the admin needs to explicitly type the server they want to connect to, and also most shells show the host name on every prompt anyway, so if you see
```
timo@integreat-test ~ % integreat-cms-cli migrate
```
it's pretty obvious that the command only affects the `integreat-test` server.

ulliholtgrave (approved):
Thanks. Probably a good change/idea even though it kinda seems strange that we even need this :sweat_smile:

PeterNerlich (approved):
:+1:

Issue #2340, opened 2023-07-06
Make selection of category for pois required


steffenkleinle:
### Motivation

As a user I want to know what kind of category a poi is and not just have all pois without any categories. I also want to have the filter function be of actual use. Making the selection required would force the users of the CMS creating a poi to select the most suitable category.

### Proposed Solution

Make the selection of a category for pois required. For existing POIs just the `Sonstige` category should be used.

### Alternatives

We will just use the `other` category as default in the frontend for now (no translations available though).

PR #2370, opened 2023-07-26
Make selection of category mandatory for POI

MizukiTemma:
### Short description

This PR makes category selection mandatory for locations.

### Proposed changes

- Make the field `category` mandatory for validation and database objects
- Update API documentation
- Adjust test

### Side effects

I have to concerns/questions:
- Some exising POIs probably do not have category in the live system. Do we update them first then apply this change as we did for organizations in #2328 ? 
- Is there any way to cut down the queries?

### Resolved issues

Fixes: #2340 


integreat_cms/cms/migrations/0076_alter_poi_category.py
ulliholtgrave, line 1: This migration will break on production...
As we have a lot of POIs created without a category you need some kind of function that deals with already existing objects and assign a default value to them as they won't meet the new constraint.

ulliholtgrave (changes requested):

integreat_cms/cms/migrations/0076_alter_poi_category.py
ulliholtgrave: I guess we agreed with the app team that the default value should be "Sonstiges" if existing pois have no category yet

MizukiTemma: Can we assume that there is only one category with `icon=others`? Then this can be found unambiguously and added to the existing POIs without category.

timoludwig: @MizukiTemma I think to be completely safe we should create such a category in the migration if it doesn't exist.

MizukiTemma: The migration is updated 🙈 I hope this is going at least in the correct direction.

timoludwig: @MizukiTemma Yes, this is better!
Two additional suggestions:
- You could use [`get_or_create()`](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#get-or-create) to get the other category object and you could use `POI.objects.filter(category=None).update(category=default_category)` to set the default category in one query.

MizukiTemma: The suggestions are applied.

integreat_cms/cms/migrations/0078_alter_poi_category.py
```diff
@@ -0,0 +1,44 @@
...
+def check_and_update_poicategory(apps, schema_editor):
+    """
+    This function checks whether the default POI category "Other" exists and cretes it if not.
+    It checks furthermore whether each location has a category and assigns the default category if not.
+    """
+    POICategory = apps.get_model("cms", "POICategory")
+    POI = apps.get_model("cms", "POI")
+
+    default_category, _ = POICategory.objects.get_or_create(
+        icon=poicategory.OTHER, color=poicategory.DARK_BLUE
```
timoludwig, line 18:
```suggestion
        icon=poicategory.OTHER, defaults={"color": poicategory.DARK_BLUE},
```
The color is not so important I'd say, so I think it's enough to only set this when the other category needs to be created.
One thought I just had: It's possible to create multiple "other" categories, right? Then, this would throw an error :neutral_face: 
Maybe my solution wasn't such idea after all? :see_no_evil: 
Also, we might have to create at least one translation?

MizukiTemma: Right...`get_or_create` throws an error if there are more than one category with `icon=other` 🤦 I would query category objects for `icon=other` and take the first one if any exists and create a new one else, because this query will be anyway necessary if `get_or_create` throws the error.

To create a translation for the newly generated other category (if any does not exists), there is a difficulty: no language exists at the moment of reading this migration but `POITranslation` model needs one......

integreat_cms/cms/migrations/0078_alter_poi_category.py
```diff
@@ -0,0 +1,44 @@
...
+
+from django.db import migrations, models
+
+from ..constants import poicategory
+
+
+# pylint: disable=unused-argument
+def check_and_update_poicategory(apps, schema_editor):
+    """
+    This function checks whether the default POI category "Other" exists and cretes it if not.
```
timobrembeck, line 11:
```suggestion
    This function checks whether the default POI category "Other" exists and creates it if not.
```

integreat_cms/cms/migrations/0078_alter_poi_category.py
```diff
@@ -0,0 +1,44 @@
...
+    It checks furthermore whether each location has a category and assigns the default category if not.
+    """
```
timobrembeck, line 13:
```suggestion
    """
    Language = apps.get_model("cms", "Language")
    default_language = Language.objects.first()
    if not default_language:
        # If no languages exists, we can assume that also no POIs exist
        return

```
> To create a translation for the newly generated other category (if any does not exists), there is a difficulty: no language exists at the moment of reading this migration but POITranslation model needs one...... Of course there is at least one language in the real database, it won't work in the development environment. And creating a language object in the migration may bring a conflict with testdata or language objects in the real database......

Ah ok, but I'd say that if not language object exists, then we probably also don't have  POI object, since we cannot create a POI objects without a POITranslation and no POITranslation without a Language :grin:
So I'd argue we can assume the database is at least in a consistent state before running this migration (and just skip the function if no language exists)...

integreat_cms/cms/migrations/0078_alter_poi_category.py
```diff
@@ -0,0 +1,44 @@
...
+    POICategory = apps.get_model("cms", "POICategory")
+    POI = apps.get_model("cms", "POI")
+
+    default_category, _ = POICategory.objects.get_or_create(
+        icon=poicategory.OTHER, defaults={"color": poicategory.DARK_BLUE},
+    )
```
timobrembeck, line 19:
```suggestion
    default_category, created = POICategory.objects.get_or_create(
        icon=poicategory.OTHER,
        defaults={"color": poicategory.DARK_BLUE},
    )
    if created:
        POICategoryTranslation = apps.get_model("cms", "POICategoryTranslation")
        POICategoryTranslation.objects.create(
            category=default_category, language=default_language, name=poicategory.OTHER
        )

```
Maybe like this to create the translation?

timobrembeck (changes requested):

integreat_cms/cms/migrations/0079_alter_poi_category.py
```diff
@@ -0,0 +1,61 @@
...
+            icon=poicategory.OTHER,
+            defaults={"color": poicategory.DARK_BLUE},
+        )
+        if created:
+            POICategoryTranslation = apps.get_model("cms", "POICategoryTranslation")
+            POICategoryTranslation.objects.create(
+                category=default_category,
+                language=default_language,
+                name=poicategory.OTHER,
+            )
```
timobrembeck, line 35:
```suggestion
        default_category = POICategory.objects.create(
            icon=poicategory.OTHER,
            color=poicategory.DARK_BLUE,
        )
        POICategoryTranslation = apps.get_model("cms", "POICategoryTranslation")
        POICategoryTranslation.objects.create(
            category=default_category,
            language=default_language,
            name=poicategory.OTHER,
        )
```
Ah, I think if no "other" category exists, it makes no sense to use `get_or_create()`, since we will create one in any case...
Sorry, in my last suggestion, I forgot that we agreed on the fact that there might be multiple other categories :innocent:

timobrembeck (changes requested):
Thanks a lot! :+1:
Apart from a cosmetic change, I think this is good to go! :rocket:

timobrembeck (approved):
Thank you! :tada:

integreat_cms/cms/migrations/0080_alter_poi_category.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+# pylint: disable=unused-argument
+def check_and_update_poicategory(apps, schema_editor):
+    """
+    This function checks whether the default POI category "Other" exists and creates it if not.
+    It checks furthermore whether each location has a category and assigns the default category if not.
+    """
+    Language = apps.get_model("cms", "Language")
+    if not (default_language := Language.objects.first()):
```
seluianova, line 15: As I understand, first() function returns a random element.
For example, when I ran the migration, a new category was created in Amharisch.
Perhaps it would be better to select the default language by id=1?

integreat_cms/cms/migrations/0080_alter_poi_category.py
```diff
@@ -0,0 +1,60 @@
...
+
+from django.db import migrations, models
+
+from ..constants import poicategory
+
+
+# pylint: disable=unused-argument
+def check_and_update_poicategory(apps, schema_editor):
+    """
+    This function checks whether the default POI category "Other" exists and creates it if not.
```
seluianova, line 11: Probably I'm missing something, but if we have one set of categories for the entire CMS (not separate sets for each region), can we just check in the prod database, if we already have this category?

integreat_cms/cms/migrations/0080_alter_poi_category.py
```diff
@@ -0,0 +1,60 @@
...
+    ):
+        default_category = POICategory.objects.create(
+            icon=poicategory.OTHER,
+            color=poicategory.DARK_BLUE,
+        )
+        POICategoryTranslation = apps.get_model("cms", "POICategoryTranslation")
+        POICategoryTranslation.objects.create(
+            category=default_category,
+            language=default_language,
+            name=poicategory.OTHER,
```
seluianova, line 33: It creates a category with the name "others" (in English, all letters are small).
I guess it should be "Sonstiges", assuming the default language should be German?

integreat_cms/cms/migrations/0080_alter_poi_category.py
MizukiTemma: I think there is no absolute garantie that the default language always has `id=1` or German is id=1 when this migration file gets run in future at some point again, because it can be deleted before (although this is not like to happen).

integreat_cms/cms/migrations/0080_alter_poi_category.py
MizukiTemma: Like the language objects, even though the default category currently exists, it can have been deleted when this migration runs in future :/

seluianova:
Do we also want to get rid of this option on the form if it is no longer valid?
image

integreat_cms/cms/migrations/0080_alter_poi_category.py
seluianova: The category names on the POI form are displayed in the CMS language, i.e. either German or English.
So, I guess the new category should be added in either German or English (or both actually?).
Maybe, filtering by language slug makes more sense then?

timobrembeck: No, the `first()` method does not return a random element, but the first one in the default ordering.
For `Language` objects, that's [`bcp47_tag`](https://github.com/digitalfabrik/integreat-cms/blob/f0b28821ca25b879b7d2733141bfd9fd9d536d1d/integreat_cms/cms/models/languages/language.py#L215), so it's no accident that Amharic was selected as your first language.
I'm not sure how we should determine the default language for all regions, maybe via the setting [`LANGUAGE_CODE`](https://github.com/digitalfabrik/integreat-cms/blob/f0b28821ca25b879b7d2733141bfd9fd9d536d1d/integreat_cms/core/settings.py#L735)? And only, if a language with this code does not exist we take the first one?
I mean this is a very rare edge case anyways, it's just to prevent an error on newly installed systems.

integreat_cms/cms/migrations/0080_alter_poi_category.py
timobrembeck: Also the migration file should work on all systems, even new production systems we might set up in the future for additional whitelabeling versions (e.g. aschaffenburg).

integreat_cms/cms/migrations/0080_alter_poi_category.py
timobrembeck: Well, I'm not sure if we can always assume that that is the case...
Anyways, this is just to prevent error in some cases, it's very likely that this case will never get executed (but  we just need to be prepared when it is)... it is easy to fix the translation in the cms in case this really happens.

integreat_cms/cms/migrations/0080_alter_poi_category.py
seluianova: > it is easy to fix the translation in the cms in case this really happens
I guess it works here too then :)

MizukiTemma: Yes, in case a new category Other is created during this migration only with the first language (Amharic or another) which is not root, we can easilly add translations for German and other languages and/or fix the translation of Amharic :)

seluianova (approved):
Looks good to me, thanks!

integreat_cms/cms/migrations/0080_alter_poi_category.py
seluianova: Okay, my inner perfectionist insisted that I bring this up anyway :)

MizukiTemma: Thank you for the detailed check 🙂

PR #2373, opened 2023-07-26
Fix Push Notification Problems

PeterNerlich:
### Short description

Fix discovered bugs in the recently merged push notification features for testing

### Proposed changes

- Timezone information is not a property of PushNotifications anymore, scheduled date is now always rendered in the timezone of the currently active region
- Languages are sorted according to the language tree of the currently active region
- If the user has access to only one region, the form won't show the regions checkboxes and POST won't contain that data. Add it back so the form data can be validated and saved properly.

### Side effects

unknown

### DIFF
--- a/integreat_cms/cms/views/push_notifications/push_notification_form_view.py
+++ b/integreat_cms/cms/views/push_notifications/push_notification_form_view.py
@@ -80,7 +80,7 @@ class PushNotificationFormView(TemplateView):
             else None
         )
 
-        details = extract_pn_details(request, push_notification)
+        details = extract_pn_details(request, push_notification, sort_for_region=region)
         not_accessible_regions_warning = None
 
         if push_notification:
@@ -99,12 +99,12 @@ class PushNotificationFormView(TemplateView):
             if details["disable_edit"]:
                 not_accessible_regions_warning = mark_safe(
                     __(
+                        _("You are not allowed to edit this message."),
                         _(
-                            "This news message is also assigned to regions you don't have access to: {}."
+                            "This news message belongs to regions you don't have access to: {}."
                         ).format(", ".join(map(str, details["other_regions"]))),
-                        _("Thus you cannot edit or delete this news message."),
                         _(
-                            "Please ask the one responsible for all of these regions or contact an administrator."
+                            "Please ask a person responsible for all regions or contact an administrator."
                         ),
                     )
                 )
@@ -160,7 +160,7 @@ class PushNotificationFormView(TemplateView):
             },
         )
 
-    # pylint: disable=too-many-branches
+    # pylint: disable=too-many-branches,too-many-statements
     def post(self, request, *args, **kwargs):
         r"""
         Save and show form for creating or editing a push notification. Send push notification
@@ -196,7 +196,9 @@ class PushNotificationFormView(TemplateView):
             )
             raise PermissionDenied
 
-        details = extract_pn_details(request, push_notification_instance)
+        details = extract_pn_details(
+            request, push_notification_instance, sort_for_region=region
+        )
 
         pn_form = PushNotificationForm(
             data=request.POST,
@@ -261,6 +263,14 @@ class PushNotificationFormView(TemplateView):
             for form in pnt_formset:
                 if not form.is_valid():
                     form.add_error_messages(request)
+        elif set(pn_form.cleaned_data["regions"]).difference(request.available_regions):
+            logger.warning(
+                "%r is not allowed to post news to %r (allowed regions: %r)",
+                request.user,
+                pn_form.cleaned_data["regions"],
+                request.available_regions,
+            )
+            raise PermissionDenied
         else:
             # Save forms
             pnt_formset.instance = pn_form.save()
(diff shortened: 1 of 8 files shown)


integreat_cms/cms/views/push_notifications/push_notification_form_view.py
```diff
@@ -379,3 +393,27 @@ def extract_pn_details(request, push_notification):
         "all_languages": all_languages,
         "disable_edit": disable_edit,
     }
+
+
+def patch_missing_regions_field(request):
```
timoludwig, line 398: I think this is the wrong way of handling this problem. I'd prefer it if the value was contained in the html form already, and only hidden from the user via css.

PeterNerlich: Yes, but I figured we needed to validate it in this case anyway (whether the user has access to it). Do you have a quick suggestion for an alternative approach?

svenseeberg: In my opinion your suggestion looks valid @PeterNerlich , thanks. I would rephrase the logged message a bit (could be a bug, not a manipulation):
```python
            logger.warning(
                "%r is not allowed to post news to %r (allowed regions: %r)",
                request.user,
                push_notification_instance,
                request.available_regions
            )
            raise PermissionDenied
```

svenseeberg:
While testing I noticed that we could easily improve the error message about the not assigned regions a bit:
```
You are not allowed to edit this message. This news message is also assigned to regions you don't have access to: Nationalparklandkreis Nürnberg. Please ask the one responsible for all of these regions or contact an administrator.
```
The point being: first tell the user what is going on. Then explain why.

integreat_cms/cms/views/push_notifications/push_notification_form_view.py
```diff
@@ -99,12 +99,12 @@ def get(self, request, *args, **kwargs):
             if details["disable_edit"]:
                 not_accessible_regions_warning = mark_safe(
                     __(
+                        _("You are not allowed to edit this message."),
                         _(
-                            "This news message is also assigned to regions you don't have access to: {}."
+                            "This news message belongts to regions you don't have access to: {}."
```
PeterNerlich, line 104: Typo:
```suggestion
                            "This news message belongs to regions you don't have access to: {}."
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -8672,6 +8672,23 @@ msgstr ""
 "angezeigt, haben allerdings keine Auswirkung auf die gesendete Push-"
 "Benachrichtigung."
 
+#: cms/views/push_notifications/push_notification_form_view.py
+msgid "You are not allowed to edit this message."
+msgstr "Sie dürfen diese Nachricht nicht bearbeiten."
+
+#: cms/views/push_notifications/push_notification_form_view.py
+msgid "This news message belongts to regions you don't have access to: {}."
```
PeterNerlich, line 8680:
```suggestion
msgid "This news message belongs to regions you don't have access to: {}."
```

PeterNerlich:
Oh no, of course it's the pesky *Too many statements* again :see_no_evil:

integreat_cms/cms/views/push_notifications/push_notification_form_view.py
```diff
@@ -263,6 +263,14 @@ def post(self, request, *args, **kwargs):
             for form in pnt_formset:
                 if not form.is_valid():
                     form.add_error_messages(request)
+        elif set(pn_form.cleaned_data["regions"]).difference(request.available_regions):
+            logger.warning(
+                "%r is not allowed to post news to %r (allowed regions: %r)",
+                request.user,
+                push_notification_instance,
+                request.available_regions
+            )
```
PeterNerlich: This message only makes sense with `pn_form.cleaned_data["regions"]`:
```suggestion
            logger.warning(
                "%r is not allowed to post news to %r (allowed regions: %r)",
                request.user,
                pn_form.cleaned_data["regions"],
                request.available_regions,
            )
```

svenseeberg: True, I mixed up the parameters.

svenseeberg:
@PeterNerlich and I found an interesting issue: while the dev server was running, we sent a message via
```
integreat-cms-cli send_push_notifications
```
The database shows the message as sent, the log output was as expected:
```
integreat=# SELECT * FROM cms_pushnotification WHERE id=9;
 id | channel | draft |           sent_date           |         created_date          |      mode      |  scheduled_send_date   
----+---------+-------+-------------------------------+-------------------------------+----------------+------------------------
  9 | news    | f     | 2023-07-27 20:57:53.178401+02 | 2023-07-27 20:51:34.231419+02 | ONLY_AVAILABLE | 2023-07-27 20:55:00+02
```

However, the dev server shows the message as not being sent: http://localhost:8000/testumgebung/push-notifications/de/9/edit/ says `Message not sent yet`. Probably some weird caching bug? Restarting the dev server does not help, it seems.

svenseeberg (approved):

Issue #2185, opened 2023-03-30
Add push notication templates


ulliholtgrave:
### Motivation

We want to give our users the ability to mark push notifications as a template so that they can be selected as a template when a new one is created.

### Proposed Solution

When users create or edit a push notification, they should find a button that allows them to save the editor content as a template.
When creating a new push notification, there should be an option to load this content as a template. Maybe by adding a new "Create new from template" button?

### User Story

- As a CMS-User I want select a pre-defined push notification template for a new notification so that I don't have to write everything from scratch.
- As a CMS-User I want create a Push Notification that I can mark as a template to use it again to save time and effort.

timoludwig:
Hmm, sounds like a very complex solution for this problem.
Why not add a button "send again" that allows just that?

ulliholtgrave:
Well, that's what we sold :see_no_evil: :D

ulliholtgrave:
But providing our municipalities with templates for most of our content types is something we should aiming for. Providing them with some clean and structures templates to fill should definitely improve the content quality and reduce their effort.

timoludwig:
OK fine 👍

ulliholtgrave:
And they actually want to edit the Push Notifications before they sent them again, because it's most of the time some kind of reminder or a announcement that changes content-wise. If we just open an existing notification, change it's content and re-send it, we would probably just update the same object again and again which would result in some problems with the versioning and the display of the message as it would be the same id. We could also create a new object if we it resend an existing notification, but I don't know if the users would really understand that...

timoludwig:
Well, setting the id to None when the "send again" Button is pressed is trivial. I agree that resent notifications should be stored as new objects in the database.
But it doesn't matter, since the desicion has already been made, I'm fine with the templating solution :grin:

hauf-toni:
Design proposal can be found [here]([url](https://www.figma.com/file/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?type=design&node-id=708%3A4786&t=uQrQNiBvokP0rtOT-1)), tbd with @ulliholtgrave 
I also agree that resent notifications should be stored as new objects if the user decides to save the new notification again (can be done by selecting a checkbox under "settings", name/version can be edited as well)

hauf-toni:
as discussed today: the creation of push notifications should only be available to users who have the administrator status. The item "Regions to which message should be sent" will also only be displayed if these users manage more than one region.

PR #2378, opened 2023-07-28
Add push notification templates

ulliholtgrave:
### Short description

This PR adds the possibility to mark push notifications as templates and reuse them for further messages. 

### Proposed changes

- Add template and template name field to the model
- Add separate list view for templates
- Add functionality to generate new and independent notifications from a template by copying their whole structure and create a new one

### Side effects

- N/A

### Resolved issues

Fixes: #2185 


timoludwig (changes requested):
Awesome, thanks a lot! :+1:

In general, the functionality seems to match the issue description!
I see two major points I'd like to address before this change can be merged:
1. There is a lot of code duplication. I would prefer if we could generalize the existing views and just adapt the slightly different behavior based on the boolean field. If there isn't an easy way to achieve this, I would also be fine with refactoring this later, but to me it seems that de-duplicating it now could even make things easier.
2. I'm not quite sure about ~~edge~~ normal cases when an existing template is edited - what is the desired behavior? Create a new copy and save it as new template? Edit the existing template? Create a copy and save it as normal news?

integreat_cms/cms/views/push_notifications/push_notification_form_view.py
```diff
@@ -323,6 +324,23 @@ def post(self, request, *args, **kwargs):
                         pn_form.instance.sent_date = datetime.now()
                         pn_form.instance.draft = False
                         pn_form.instance.save()
```
PeterNerlich, line 326: Do we want to update this regardless of whether it was a template or not or only when it was a "normal" push notification?

integreat_cms/cms/views/push_notifications/push_notification_template_form_view.py
```diff
@@ -0,0 +1,381 @@
...
+    ).distinct()
+    disable_edit = bool(
+        not request.user.is_superuser and not request.user.is_staff and other_regions
+    )
+    return {
+        "all_regions": all_regions,
+        "other_regions": other_regions,
+        "all_languages": all_languages,
+        "disable_edit": disable_edit,
+    }
```
PeterNerlich, line 381: We might want to move this to a utils file in this case

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -2044,10 +2044,22 @@ msgstr "Geplanter Tag zum Senden"
 msgid "Scheduled send time"
 msgstr "Geplante Zeit zum Senden"
 
+#: cms/forms/push_notifications/push_notification_form.py
+msgid "Save this message as template"
+msgstr "Diese Seite als Vorlage speichern"
```
PeterNerlich, line 2049: Seite oder Nachricht?
```suggestion
msgstr "Diese Nachricht als Vorlage speichern"
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -3390,6 +3402,18 @@ msgstr ""
...
+msgid "push notification template"
+msgstr "Push-Benachrichtigungs-Vorlage"
+
+#: cms/models/push_notifications/push_notification.py
+msgid "push notification template name"
+msgstr "Name der Vorlage"
+
+#: cms/models/push_notifications/push_notification.py
+msgid "Provide a distinct name for the template"
+msgstr "Bitte vergeben Sie einen eindeutigen Namen für die Vorlage"
```
PeterNerlich, line 3415:
```suggestion
msgstr "Geben Sie der Vorlage einen eindeutigen Namen"
```

PeterNerlich (approved):
A bunch of thoughts from me that I don't really see as a blocker.

To be honest, I find the feature confusing to use currently, but think we should address that (soon) after releasing this.
As a user, I expect Templates to be different from Push Notifications themselves. Even if they are stored in the same DB table with only a flag differentiating them, I think about them as different categories entirely. Kind of like classes and instances of them. It is confusing to me that the form when clicking on a template in the templates overview looks the same as clicking on a normal Push Notification. It is confusing that I can accidentally send a normal Push Notification when I went to create it via the templates view. It is irritating to me that I can remove a template from the list and add it to the normal overview. I think we should re-think how we approach this / approach to communicate this in the near future.

timoludwig:
> To be honest, I find the feature confusing to use currently, but think we should address that (soon) after releasing this.
As a user, I expect Templates to be different from Push Notifications themselves. Even if they are stored in the same DB table with only a flag differentiating them, I think about them as different categories entirely. Kind of like classes and instances of them. It is confusing to me that the form when clicking on a template in the templates overview looks the same as clicking on a normal Push Notification. It is confusing that I can accidentally send a normal Push Notification when I went to create it via the templates view. It is irritating to me that I can remove a template from the list and add it to the normal overview. I think we should re-think how we approach this / approach to communicate this in the near future.

I agree, in the long term I think we should create one abstract base push notification model with the shared fields, and then two sub models for the templates and the actual news. 
But this would require quite some refactoring, probably too much work for a few days. Let's improve this later.
And yes, I would also make it more apparent what the user is currently editing - a normal news or a template...

osmers:
Any idea if this can be fixed/resolved/released today as well together with the other two new features? Bcs I know Philipp will ask at out 9am meeting

integreat_cms/cms/migrations/0076_auto_20230727_1652.py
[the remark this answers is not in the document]
svenseeberg: Done.

integreat_cms/cms/forms/push_notifications/push_notification_form.py
```diff
@@ -91,6 +98,13 @@ def clean(self):
                 ),
             )
 
+        # Check if a template name is set when the message should be used as a template
+        if cleaned_data.get("is_template") and not cleaned_data.get("template_name"):
+            self.add_error(
+                "template_name",
+                forms.ValidationError(_("Please provide a name for your template")),
+            )
```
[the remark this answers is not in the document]
svenseeberg: However, it is a bit nasty to access this information in this location, AFAICT :D

integreat_cms/cms/forms/push_notifications/push_notification_form.py
```diff
@@ -124,7 +138,14 @@ def clean(self):
...
     class Meta:
         model = PushNotification
-        fields = ["channel", "regions", "mode", "scheduled_send_date"]
+        fields = [
+            "channel",
+            "regions",
+            "mode",
+            "scheduled_send_date",
+            "is_template",
+            "template_name",
```
david-venhoff, line 147: I think these are not required, since their corresponding fields are already manually defined above

integreat_cms/cms/templates/push_notifications/push_notification_template_list.html
```diff
@@ -0,0 +1,63 @@
...
+{% block content %}
+    <div class="table-header">
+        <h1 class="heading">
+            {% translate "Templates" %}
+        </h1>
+        <div class="flex flex-wrap justify-end gap-4">
+            {% if perms.cms.change_pushnotification %}
+                <a href="{% url 'new_push_notification' region_slug=request.region.slug language_slug=language.slug %}"
+                   class="btn">
+                    {% translate "Create news message" %}
```
david-venhoff, line 14: Since we are in the template push notiifcation list view, I would expect that clicking this button creates a new template, not a 'normal' pn

david-venhoff (commented):
Apart from what was already said, I found no blocking issues

integreat_cms/cms/forms/push_notifications/push_notification_form.py
timoludwig: Ah yeah, but it would make more sense to remove the manually defined fields than this ones. The label should be the verbose name of the model field.

integreat_cms/cms/forms/push_notifications/push_notification_form.py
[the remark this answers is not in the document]
svenseeberg: Peter and I decided to ignore this ;-)

integreat_cms/cms/templates/push_notifications/push_notification_template_list.html
svenseeberg: @david-venhoff would it be okay if we automatically set the checkbox?

david-venhoff: sure!

svenseeberg: Fixed.

integreat_cms/cms/views/push_notifications/push_notification_form_view.py
svenseeberg: We changed that

integreat_cms/cms/views/push_notifications/push_notification_list_view.py
```diff
@@ -86,7 +86,7 @@ def get(self, request, *args, **kwargs):
                 },
             )
 
-        push_notifications = region.push_notifications.all()
+        push_notifications = region.push_notifications.filter(is_template=False)
```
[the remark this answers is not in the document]
svenseeberg: We de-duplicated some code but decided to leave this as is and create another issue for further code cleaning.

integreat_cms/cms/templates/push_notifications/push_notification_form.html
```diff
@@ -24,9 +32,15 @@ <h1 class="heading overflow-hidden text-ellipsis">
                             {% translate "Save" %}
                         </button>
                     {% endif %}
-                    {% if perms.cms.send_push_notification and not push_notification_form.instance.sent_date and not not_accessible_regions_warning %}
+                    {% if not push_notification_form.instance.id and push_notification_form.is_template.value %}
+                        {# hide button when creating a new template #}
```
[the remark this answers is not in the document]
svenseeberg: And yes, the if is a strange workaround for missing brackets in templates. But maybe it is even better to read this way.

svenseeberg (approved):

Issue #2172, opened 2023-03-27
Fallback texts of mirrored pages list empty pages


timoludwig:
### Describe the Bug

When a mirrored content is not available in the current language, we show a partial fallback text.
This is also shown when the referenced translation is empty, which makes no sense.

### Steps to Reproduce

1. Create an empty page
2. Reference another empty page as live content
3. Inspect `page_translation.mirrored_translation_text_or_fallback_message`

### Expected Behavior

Empty translations should not be listed as fallback translations of mirrored pages

### Actual Behavior

All available languages are listed, even those with empty translations

PR #2379, opened 2023-07-31
Do not show empty mirrored content as fallback

MizukiTemma:
### Short description

This PR fixes the bug that not only non-empty translations but also empty translasions are suggested as fallback of the embedded content.

### Proposed changes

- Filter empty translations out of the fallback list
- Return an empty string as in the case where the page has no mirrored content, if  there is no non-empty translations which can be a fallback. (Otherwise no language will be shown after the message _Part of the page does not exist in the selected language. It is however available in these languages:_ and it is confusing.)

### Side effects

- I may have misunderstood some part of the functionalitry and/or the wished status 🙈 

### Resolved issues

Fixes: #2172 

### DIFF
--- a/integreat_cms/cms/models/pages/page_translation.py
+++ b/integreat_cms/cms/models/pages/page_translation.py
@@ -132,26 +132,39 @@ class PageTranslation(AbstractBasePageTranslation):
         :return: The text, as specified
         :rtype: str
         """
+
+        # If there is no page embetted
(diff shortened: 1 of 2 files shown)


david-venhoff (approved):
LGTM

ulliholtgrave (approved):
Thanks, This seems to do the trick :+1:

Issue #1792, opened 2022-10-25
Exclude reserved namespaces from possible slugs


timobrembeck:
### Motivation

Some urls are reserved for the linking of the webapp, e.g.

- Global patterns:
    - https://integreat.app/landing
    - https://integreat.app/main-disclaimer
    - https://integreat.app/licenses
    - https://integreat.app/recommend

- Region-specific patterns:
    - https://integreat.app/testumgebung/de/disclaimer
    - https://integreat.app/testumgebung/de/news/local
    - https://integreat.app/testumgebung/de/events
    - https://integreat.app/testumgebung/de/locations
    - https://integreat.app/testumgebung/de/offers
    - https://integreat.app/testumgebung/de/search

### Proposed Solution

- We should prevent `landing`, `main-disclaimer`, `licenses` and `recommend` as slugs for regions 
- We should prevent `disclaimer`, `news`, `events`, `locations`, `offers` and `search` as slugs for pages.

(probably it makes sense to put these lists in the settings file)

Issue #522, opened 2020-10-07
Meta-description for pages/events/pois


timobrembeck:
### Motivation

To improve SEO, we want to add the possibility to add meta-descriptions for pages/events/pois.

### Proposed Solution

- [ ] Add a new field `meta_description = models.TextField(blank=True)` to `AbstractBasePageTranslation`, `EventTranslation` and `POITranslation` models.
- [ ] Add this field to the `PageTranslationForm`, `EventTranslationForm` and `POITranslationForm`.
- [ ] Add a new textarea beneath the `text`/`description` fields to the templates of that form (`page_form.html`, `event_form.html`, `poi_form.html`).
- [ ] Add `description` property to the `Region` model, which can be used for the content overview in the app.

svenseeberg:
The challenge here is that our content creators do not know anything about SEO. Therefore we need to steer them into a direction that in the end results in a good meta information. We could have an additional input field that asks for a 1 to 3 sentences summary of the page content.

JoeyStk:
@steffenkleinle could you please give an update if this is still wanted and if so, why :)

steffenkleinle:
I think this is not really my place to tell. This is more a question for the PO/the service team whether they deem it necessary/sensible that cities can edit the meta description to improve on the SEO.

hauf-toni:
waiting for input from the service team’s perspective — I'll get back to you asap @JoeyStk

laravoelk:
We will talk about this at the conference with the whole team and get back to you next week. :-)

steffenkleinle:
> We will talk about this at the conference with the whole team and get back to you next week. :-)

Perhaps this should even be an even bigger discussion including the marketing team as this is mainly relevant for SEO.

laravoelk:
We have discusses this issue at the conference and we were wondering if it would generally be possible to include AI, so that the Meta description is automatically generated and municipalities can just change it if they need.

Issue #402, opened 2020-06-02
Translatable Meta Description per Region


dkehne:
To enhance SEO every instance should have an own customizable meta description which can be maintained by the city, if they want to do so. The functionality could be implemented similar to the imprint, where a city has the chance to maintain an own page but still have the possiblity to export it via XLIFF to get translations automatically. 

If there should be a maximum length which is fixed technically by us, has to be evaluated.

timoludwig:
Things that have to be clarified from a technical side of view:
- Do we want this to be a field in the `Region` model or a separate content model? (because of the translations probably a separate model)
- Which maximum length do we use? For Google, only the width of the rendered text counts, which should be smaller than 920 pixels. This comes down to ~120 to 158 characters.

steffenkleinle:
Once done please notify the app team (https://issues.tuerantuer.org/browse/IGAPP-296).

PR #2391, opened 2023-08-07
Add API for social media tags

svenseeberg:
### Short description
Add a new API endpoint that returns a ready to use HTML snippet with social media meta tags. The resulting HTML code can be included with Nginx Server Side Includes into the index.html of the web app.

### Proposed changes
- Register path `/api/social/REGION_SLUG/LANGUAGE_SLUG/PATH`
- Search for suitable page objects and return tags for the first found object
- Provide fallback strings if no page is found.
- Introduce new settings: platform name & social media preview image

### Side effects
- Does not really work for other types. The issue here is that we plan to pass the path directly from the URL in the web app. That means events would come in as `/api/social/testumgebung/ar/events/sprachmentoring-treffen-2$2023-08-12`. Not sure if we can find a good way to handle this, as we definitely will have namespace collissions. However, we could add a dedicated route for `/api/social/REGION_SLUG/LANGUAGE_SLUG/{events,locations}/PATH` etc. Feedback is welcome.
- Tests are still missing.
- The template is in the `cms` app templates directory, not in the `api` app.
- We need special handling for the `hallo` region slug as this is hallo aschaffenburg with a dedicated app in the Integreat CMS.
- We currently have no concept of front end user languages in our CMS. With more translation files we could support localized preview descriptions.

### Resolved issues
Fixes: #2084

### Additional context

This page should basically be replaced by the social endpoint -> https://git.tuerantuer.org/Integreat/social-media-meta-tags/src/branch/master/meta.php

A few helpful links to get a better understanding of the why and all the follow ups:
- [Understanding google & SEO](https://developers.google.com/search/docs/fundamentals/seo-starter-guide)
- [Structured Data (json+ld)](https://developers.google.com/search/docs/appearance/structured-data/intro-structured-data)
- [Open Graph aka. `og:title`, ...](https://ogp.me/)


steffenkleinle:
> Does not really work for other types. The issue here is that we plan to pass the path directly from the URL in the web app. That means events would come in as `/api/social/testumgebung/ar/events/sprachmentoring-treffen-2$2023-08-12`. Not sure if we can find a good way to handle this, as we definitely will have namespace collissions. However, we could add a dedicated route for `/api/social/REGION_SLUG/LANGUAGE_SLUG/{events,locations}/PATH` etc. Feedback is welcome.

I am pretty sure that the suffix `$2023-08-12` is not the problem but if it is, we already have an issue to remove this again soon: https://github.com/digitalfabrik/integreat-app/issues/2031
So if that would not work for some time, not a problem imo.

sarahsporck:
As far as I'm concerned I think setting social media headers in the backend as much as possible should be the preferred solution. In the best case scenario this would also include json ld tags.

Therefore, the following tags seem to be missing:
```html
<meta property="og:site_name" content="integreat.app/malte.app/halloaschaffenburg.app">
<meta property="og:image" content="<%= config.icons.socialMediaPreview %>" />
<meta property="og:image:width" content="1200" />
<meta property="og:image:height" content="630" />
<meta property='og:url' content={window.location.href} /> // not sure about this one
```

And besides these there are the json+ld tags ([events](https://github.com/digitalfabrik/integreat-app/blob/main/web/src/components/JsonLdEvent.tsx), [breadcrumbs](https://github.com/digitalfabrik/integreat-app/blob/main/web/src/components/JsonLdBreadcrumbs.tsx)), which are a little bit more complicated, and adding these could also be discussed separately.

Any other tags that are calculated in the frontend should not have anything to do with how search engines or social media crawlers see our site.

sarahsporck:
@svenseeberg do you need more help or input for this? If you're too busy right now, I could take over :)

integreat_cms/api/urls.py
```diff
@@ -123,6 +124,16 @@
 urlpatterns = [
     path("api/regions/", include(region_api_urlpatterns)),
     path("wp-json/extensions/v3/sites/", include(region_api_urlpatterns)),
+    path(
+        "api/social/<slug:region_slug>/<slug:language_slug>/",
```
sarahsporck, line 128: [At least for google](https://developers.google.com/search/docs/appearance/site-names?hl=de#technical-guidelines) the root page (https://integreat.app) also needs the correct title and (maybe also some other configuration?). I think this would need separate handling.

svenseeberg: True, we need fallback strings.

svenseeberg:
We also have a problem with Aschaffenburg, because they are in the same CMS and I don't want the social media previews to be region specific. Well, we may have to create region settings with global fallbacks.

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -22,6 +22,14 @@ msgstr ""
 msgid "API"
 msgstr "API"
 
+#: api/v3/socialmedia_headers.py
+msgid ""
+"hallo aschaffenburg is your digital companion for the town of Aschaffenburg. "
+"Here you can find local information, advice centres and services."
+msgstr ""
+"hallo aschaffenburg is your digital companion for the town of Aschaffenburg. "
+"Here you can find local information, advice centres and services."
```
timobrembeck, line 31:
```suggestion
"hallo aschaffenburg ist Ihr digitaler Begleiter für die Stadt Aschaffenburg. "
"Hier finden Sie lokale Informationen, Beratungsstellen und Dienstleistungen."
```

integreat_cms/core/settings.py
```diff
@@ -68,6 +68,14 @@
 #: URL to the Integreat wiki
 WIKI_URL = os.environ.get("INTEGREAT_CMS_WIKI_URL", "https://wiki.integreat-app.de")
 
+#: Platform Name, used for app users and social media headers
+PLATFORM_NAME = os.environ.get("INTEGREAT_PLATFORM_NAME", "Integreat")
```
timobrembeck, line 72: I think it makes sense to combine this setting with the existing `BRANDING` value to avoid having to maintain both settings.
I opened #2434 for this matter, and in case that PR gets merged first, I'd suggest to remove the `PLATFORM_NAME` setting and use `BRANDING_TITLE` instead.

integreat_cms/api/v3/socialmedia_headers.py
```diff
@@ -0,0 +1,71 @@
+"""
+This module includes functions related to the pages API endpoint.
```
timobrembeck, line 2:
```suggestion
This module includes functions related to the social media headers API endpoint.
```

integreat_cms/api/v3/socialmedia_headers.py
```diff
@@ -0,0 +1,71 @@
...
+
+from ...cms.models import PageTranslation
+
+logger = logging.getLogger(__name__)
+
+
+# pylint: disable=unused-argument
+def socialmedia_headers(request, region_slug, language_slug, path):
+    """
+    Retrieves all children for a single page
```
timobrembeck, line 21:
```suggestion
    Renders the social media headers for a single page
```

integreat_cms/release_notes/2023/unreleased/2391.yml
```diff
@@ -0,0 +1,2 @@
+en: Add social media headers API endpoint
+de: Neuer Social Media Kopfzeilen API Endpunkt
```
timobrembeck, line 2:
```suggestion
de: Neuer API Endpunkt für Social-Media-Header
```

integreat_cms/api/v3/socialmedia_headers.py
```diff
@@ -0,0 +1,71 @@
...
+    :type path: str
+
+    :return: HTML meta headers required by social media platforms
+    :rtype: ~django.template.response.TemplateResponse
+    """
+    region = request.region
+    page_translation = PageTranslation.search(region, language_slug, path).first()
+    activate(language_slug)
+
+    if page_translation:
```
timobrembeck, line 42: This condition doesn't really make sense, does it?
Did you mean to check for the aschaffenburg slug here?
```suggestion
    if region.slug != "hallo":
```
But if we move aschaffenburg into its own CMS, this case is irrelevant anyways...

integreat_cms/api/urls.py
```diff
@@ -123,6 +124,16 @@
...
+        include(
+            [
+                re_path(
+                    r"^(?P<path>\w+)$", socialmedia_headers, name="socialmedia_headers"
```
timobrembeck, line 132:
```suggestion
                path(
                    "<path:path>/", socialmedia_headers, name="socialmedia_headers"
```
For me, this regex didn't work for non-root pages. However, we don't even need the regex since Django supports paths per default via the [`path` converter](https://docs.djangoproject.com/en/3.2/topics/http/urls/#path-converters).

integreat_cms/api/v3/socialmedia_headers.py
```diff
@@ -0,0 +1,71 @@
...
+        excerpt = (
+            unescape(strip_tags(page_translation.content))[:100]
+            if page_translation
+            else None
```
timobrembeck, line 46: `None` does also not make sense here, we'd probably want to return a 404 error in case the requested page does not exist (or is not public)?

integreat_cms/api/v3/socialmedia_headers.py
```diff
@@ -0,0 +1,71 @@
...
+        )
+        title = page_translation.title
+    else:
+        excerpt = _(
+            "hallo aschaffenburg is your digital companion for the town of Aschaffenburg. Here"
+            " you can find local information, advice centres and services."
+            if region == "hallo"
+            else "Your local information guide with over 100 regions in Germany."
+        )
+        title = "Local Information for You"
```
timobrembeck, line 56: Instead of hardcoding such a fallback value, we'd probably want to implement #402, right?

timobrembeck (changes requested):
Thanks a lot! :muscle: 

Since we're thinking about moving aschaffenburg into its own cms, this is probably not the final state anyways, but I just left my preliminary comments anyway :see_no_evil:

integreat_cms/core/settings.py
timobrembeck:
```suggestion
```
The PR is now merged, so you could rebase this one and replace `PLATFORM_NAME` with `BRANDING_TITLE` everywhere.

integreat_cms/locale/de/LC_MESSAGES/django.po
svenseeberg: I intentionally added the English translation here. As we do not have language tags here, we probably want to only use the English translations (as of now). But maybe I can fully remove the translation.

integreat_cms/api/v3/socialmedia_headers.py
svenseeberg: I'm not yet sure how we do that with the timing of switching the CMS. I would use your suggestion right now.

integreat_cms/api/v3/socialmedia_headers.py
svenseeberg: Yes, good idea.

integreat_cms/locale/de/LC_MESSAGES/django.po
timobrembeck: Ah ok, but then I'd use the
```
from django.utils import translation

with translation.override("en"):
    excerpt = _("...")
```
or wouldn't use the gettext-function at all, since this way we might forget about it and just have missings translations in our po file that we don't know about...

integreat_cms/api/v3/socialmedia_headers.py
svenseeberg: Yes, we could indeed just add a new region property, if we can agree that we support only a single language for the time being.

integreat_cms/locale/de/LC_MESSAGES/django.po
svenseeberg: yes, lets fully remove the translation.

integreat_cms/api/v3/socialmedia_headers.py
timobrembeck: Ah hmm yeah forgot about the translations... I guess translations are pretty important in the mid-term? Then I'm fine with using the hardcoded approach now and build a good solution with own translation models in the next weeks/months...

svenseeberg: Theoretically we could include the `<html>` tag in the endpoint. This would enable translations for the meta description as well.

svenseeberg: I think Nginx does not care about which part of the HTML document we include. I have no idea how we would dynamically set a header though ;)

sarahsporck:
As it seems that @svenseeberg doesn't find the time to finish this right now, I'll take over.

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -9340,6 +9340,15 @@ msgstr ""
...
 "gehört ({})."
 
+#~ msgid ""
+#~ "hallo aschaffenburg is your digital companion for the town of "
+#~ "Aschaffenburg. Here you can find local information, advice centres and "
+#~ "services."
+#~ msgstr ""
+#~ "hallo aschaffenburg is your digital companion for the town of "
+#~ "Aschaffenburg. Here you can find local information, advice centres and "
+#~ "services."
```
sarahsporck, line 9350: I guess this text also has to be added for integreat then. How do you add a translation again?

integreat_cms/api/urls.py
```diff
@@ -123,6 +128,28 @@
...
+                path(
+                    "<slug:region_slug>/<slug:language_slug>/<path:path>/",
+                    page_socialmedia_headers,
+                    name="page_socialmedia_headers",
+                ),
+                path(
+                    "<slug:region_slug>/<slug:language_slug>/",
+                    region_socialmedia_headers,
+                    name="region_socialmedia_headers",
+                ),
```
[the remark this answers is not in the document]
timobrembeck: I think we should store the web app's reserved routes in our constants anyway, e.g. to be able to prevent regions to be created with the slug "landing" (also see #1792), so maybe this can be resolved together? If it's too complicated, we could also merge this PR fist and open another issue for improving this in the future.

integreat_cms/api/v3/socialmedia_headers.py
```diff
@@ -0,0 +1,134 @@
...
+    PushNotificationTranslation,
+)
+
+from ...cms.models import PageTranslation
+
+logger = logging.getLogger(__name__)
+
+
+def root_socialmedia_headers(request, path):
+    title = f"{settings.BRANDING_TITLE} | Web-App | Local Information for You"
```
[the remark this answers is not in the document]
timobrembeck: If we want to translate this, we'd have to add a new field to the `Language` model, similar to e.g. the [translation for "Table of contents"](https://github.com/digitalfabrik/integreat-cms/blob/2c560214ffb3a8af52435dec32192d7dca730a21/integreat_cms/cms/models/languages/language.py#L102).

integreat_cms/api/v3/socialmedia_headers.py
```diff
@@ -0,0 +1,134 @@
...
+
+    :param language_slug: Code to identify the desired language
+    :type language_slug: str
+
+    :return: HTML meta headers required by social media platforms
+    :rtype: ~django.template.response.TemplateResponse
+    """
+    region = request.region
+    title = f"{region.name} | {settings.BRANDING_TITLE}"
+    excerpt = ""  # TODO: individual excerpt per city
```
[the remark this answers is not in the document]
timobrembeck: In the long term, this should be covered by #402. For now, an automatic excerpt from the content should be good enough, or what do you think?
For pages/events/pois, we have #522, for news I think it should just be the message itself, and for offers we don't have an issue yet, do you want to open one?

integreat_cms/locale/de/LC_MESSAGES/django.po
timobrembeck: For the backend UI translations (which the translation file belongs to), see: https://digitalfabrik.github.io/integreat-cms/internationalization.html

However, for the social media tags we can't really use this, since it only supports the backend UI languages English, German and Dutch. For the content languages, we need to store the translations in the database. Since we're currently lacking the appropriate model structure for this, I'm afraid this is out of scope for this PR...
Do you want to open a new issue for the meta description of the root level / region selection (simiar to #402 and #522)?

sarahsporck: See #2563 .

integreat_cms/api/v3/socialmedia_headers.py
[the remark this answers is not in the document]
sarahsporck: As offers will be removed in foreseeable future I'd not open an issue for that.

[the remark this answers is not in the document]
sarahsporck: > For now, an automatic excerpt from the content should be good enough, or what do you think?

I think that's what we are doing rn either way, so this should be fine.

integreat_cms/api/urls.py
```diff
@@ -128,10 +129,21 @@
...
 ]
 
+socialmedia_api_urlpatterns = [
+    path("", socialmedia_headers, name="root_socialmedia_headers"),
+    path(
+        "<path:path>",
+        socialmedia_headers,
+        name="socialmedia_headers",
+    ),
+]
```
sarahsporck, line 139: I realized that it is not really simple to correctly match the pages here. 
`region_slug/language_slug` refers to a region, right. But what about `landing/de`? Due to the region middleware this would match and return 404. When using the other approach to match only specific slugs (as implemented here #1792) is not really flexible and requires action as soon as another path/slug joins the pool.
Therfore, I moved the matching/detection logic to `socialmedia_headers.py`.

integreat_cms/api/v3/socialmedia_headers.py
```diff
@@ -0,0 +1,185 @@
...
+    region = (
+        Region.objects.filter(slug=potential_region_slug).first()
+        if potential_region_slug
+        else None
+    )
+
+    potential_language_slug = slugs[1] if len(slugs) > 1 else None
+    language = Language.objects.filter(
+        slug=potential_language_slug
+    ).first() or Language.objects.get(slug="de")
```
sarahsporck, line 61: Fallback to german, if the language cannot be found (as it is the current approach in the webapp).

integreat_cms/api/v3/socialmedia_headers.py
```diff
@@ -0,0 +1,185 @@
...
+    """
+    Renders the social media headers for a root region page
+
+    :param request: The request that has been sent to the Django Server
+    :param region: The region the request refers to
+    :param path: The path to the page
+    """
+    title = f"{region.name} | {settings.BRANDING_TITLE}"
+    url = f"{settings.WEBAPP_URL}/{path}"
+    return render_socialmedia_headers(request, title, None, url)
```
sarahsporck, line 111: Omit the excerpt for these pages. Even though, I'm sure google will handle this correctly and crawl the content, I'm actually not sure about `TwitterBot`, `LinkedInBot`, etc. ...

integreat_cms/api/v3/socialmedia_headers.py
```diff
@@ -0,0 +1,185 @@
...
+    elif route == "locations" and len(route_path):
+        content_translation = POITranslation.search(
+            region, language.slug, *route_path
+        ).first()
+    elif route == "news" and len(route_path):
+        content_translation = PushNotificationTranslation.search(
+            region, language.slug, *route_path
+        ).first()
+    else:
+        # TODO: add breadcrumb json-ld if content_translation exists
```
sarahsporck, line 143: Is it okay to leave the todos? I'd prefer doing json-ld in a separate PR

integreat_cms/cms/migrations/0084_language_socialmedia_webapp_title.py
```diff
@@ -0,0 +1,26 @@
...
+    dependencies = [
+        ("cms", "0083_add_page_based_offer_fields"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="language",
+            name="socialmedia_webapp_title",
+            field=models.CharField(
+                default="Integreat | Web-App | Lokale Informationen für Dich",
```
sarahsporck, line 20: Quite sure, this does not work correctly for malte & aschaffenburg. Should I omit the default value? Also, we actually already have translations for this in the frontend, which could be "ported" to the backend somehow(?/by hand?).

integreat_cms/locale/de/LC_MESSAGES/django.po
sarahsporck: For this text we'd also already have translations in the frontend.

integreat_cms/api/v3/socialmedia_headers.py
```diff
@@ -0,0 +1,185 @@
...
+    Available contexts are:
+    - root e.g. /, /landing, /disclaimer, /landing/de
+    - region root e.g. /testumgebung, /testumgebung/de, /testumgebung/de/events
+    - page e.g. /testumgebung/de/willkommen, /testumgebung/de/events/event
+
+    :param request: The request that has been sent to the Django server
+    :param path: The path of the page the social meddia headers should be created for
+
+    :return: HTML meta headers required by social media platforms
+    """
```
sarahsporck, line 48: Tests tbd soon.

integreat_cms/api/urls.py
```diff
@@ -128,10 +129,21 @@
...
+
 #: The url patterns of this module (see :doc:`django:topics/http/urls`)
 urlpatterns: list[URLPattern] = [
     path("api/regions/", include(region_api_urlpatterns)),
     path("wp-json/extensions/v3/sites/", include(region_api_urlpatterns)),
+    path("api/social/", include(socialmedia_api_urlpatterns)),
+    path("wp-json/extensions/v3/social", include(socialmedia_api_urlpatterns)),
```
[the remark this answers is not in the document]
timobrembeck: What do you mean exactly?

integreat_cms/api/v3/socialmedia_headers.py
timobrembeck: Sure :+1:

integreat_cms/api/v3/socialmedia_headers.py
```diff
@@ -0,0 +1,298 @@
...
+if TYPE_CHECKING:
+    from typing import Optional
+
+    from django.http import HttpRequest, HttpResponse
+
+logger = logging.getLogger(__name__)
+
+
+def socialmedia_headers(
+    request: HttpRequest, path: Optional[str] = None
```
timobrembeck, line 34:
```suggestion
    from django.http import HttpRequest, HttpResponse

logger = logging.getLogger(__name__)


def socialmedia_headers(
    request: HttpRequest, path: str | None = None
```
Although the python world [doesn't completely agree](https://discuss.python.org/t/clarification-for-pep-604-is-foo-int-none-to-replace-all-use-of-foo-optional-int/26945) on one consistent syntax, the `|` union separator seems to be the more modern way to express `Union` and `Optional` (see [PEP 604](https://peps.python.org/pep-0604/)).

integreat_cms/cms/templates/socialmedia_headers.html
```diff
@@ -0,0 +1,13 @@
+<title>{{ title }}</title>
```
timobrembeck, line 1:
```suggestion
<html lang="{{ language_code }}">
<title>{{ title }}</title>
```
What do you think about the idea to include the `<html>` tag here as well to make it possible for the bots to detect the content language of the page?

timobrembeck (changes requested):
Here as well, sorry for the long waiting period and thanks a lot for your changes! :muscle: 

I think this is a big step in the right direction! :+1:

One question I'd like to raise is how we should handle 404 errors: In my opinion, the social media preview should indicate that the link is not worth clicking on.

integreat_cms/api/urls.py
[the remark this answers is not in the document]
sarahsporck: Do you also maintain the wp endpoints for "new" endpoints was the question. ^^
The previous question was very sparsely formulated 🙈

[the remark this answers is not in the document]
timobrembeck: Ah, yeah I guess if we assume that #2617 is merged soon, then we don't really need the legacy endpoint.

integreat_cms/api/v3/socialmedia_headers.py
```diff
@@ -0,0 +1,298 @@
...
+    region = (
+        Region.objects.filter(slug=potential_region_slug).first()
+        if potential_region_slug
+        else None
+    )
+
+    potential_language = (
+        Language.objects.filter(slug=slugs[1]).first() if len(slugs) > 1 else None
+    )
+    language = potential_language or Language.objects.get(slug="de")
```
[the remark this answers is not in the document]
sarahsporck: We have "invalid" urls such as
- https://webnext.integreat.app/testumgebung/test
- https://webnext.integreat.app/testumg/de

And what you mean by throw an 404 is, that we do not want that these pages are shown in the search results and the server does not return any headers, right?

integreat_cms/api/v3/socialmedia_headers.py
```diff
@@ -0,0 +1,298 @@
...
+    if not region:
+        return root_socialmedia_headers(request, language, path or "")
+
+    content_headers = None
+    if len(slugs) >= 3:
+        content_headers = content_socialmedia_headers(
+            request, region, language, "/".join(slugs[2:])
+        )
```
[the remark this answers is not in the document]
sarahsporck: I was thinking about new pages/features that might come in the future. Everytime we add a new url we'd need to update this. Just think about the offers tab. But maybe we can also assume that google can do the fallback and stuff and we don't have to worry about this.

integreat_cms/api/v3/socialmedia_headers.py
```diff
@@ -0,0 +1,298 @@
...
+        (
+            "hallo aschaffenburg is your digital companion for the town of Aschaffenburg. Here"
+            " you can find local information, advice centres and services."
+        )
+        if settings.BRANDING == "aschaffenburg"
+        else (
+            settings.BRANDING_TITLE + " is your digital companion for Germany. Here"
+            " you can find local information, advice centres and services."
+        )
+    )
```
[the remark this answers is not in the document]
sarahsporck: I feel by keeping it empty we implicate ourselves. If we take 1 year until we implement #2563 the page score could have fallen quite visibly.

integreat_cms/api/v3/socialmedia_headers.py
```diff
@@ -0,0 +1,298 @@
...
+    :param request: The request that has been sent to the Django server
+    :param region: The region the page belongs to
+    :param language: Language of the page
+    :param path: Should be a page translation slug
+    :return: HTML meta headers required by social media platforms
+    """
+
+    path_elements = path.split("/", maxsplit=1)
+    route = path_elements[0] if len(path_elements) > 0 else None
+    slug = path_elements[1] if len(path_elements) == 2 else None
```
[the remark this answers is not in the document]
sarahsporck: very clever 👍🏼

integreat_cms/api/v3/socialmedia_headers.py
```diff
@@ -0,0 +1,298 @@
...
+    """
+    if not (
+        event_translation := EventTranslation.search(
+            region, language.slug, slug
+        ).first()
+    ):
+        return None
+
+    # TODO: add event json-ld
+    activate(language.slug)
```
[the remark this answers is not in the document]
sarahsporck: Not sure what activate does anyway 🙈

integreat_cms/api/v3/socialmedia_headers.py
```diff
@@ -0,0 +1,298 @@
...
+    :param language: The language the event is supposed to belong to
+    :param slug: The event slug
+    :return: HTML social meta headers required by social media platforms if the event page exists
+    """
+    if not (
+        event_translation := EventTranslation.search(
+            region, language.slug, slug
+        ).first()
+    ):
+        return None
```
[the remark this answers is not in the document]
sarahsporck: So we don't want to show any meta headers in this case? Basically same question as above.

integreat_cms/cms/fixtures/test_data.json
```diff
@@ -413,6 +415,7 @@
       "text_direction": "RIGHT_TO_LEFT",
       "primary_country_code": "ab",
       "table_of_contents": "محتويات",
+      "socialmedia_webapp_title": "Integreat | Web-App | Lokale Informationen für Dich",
```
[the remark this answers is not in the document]
sarahsporck: Are you just reffering to the tests or also in the live system? We also changed the remaining part, as it does make sense for e.g. Malte to have a different title.

integreat_cms/cms/templates/socialmedia_headers.html
sarahsporck: But we could put the canonical urls inside here https://github.com/digitalfabrik/integreat-app/blob/4dd43127ffae65ae1326640e67b5e6ab22ec2bc1/web/src/components/Helmet.tsx#L24

integreat_cms/api/v3/socialmedia_headers.py
```diff
@@ -0,0 +1,298 @@
...
+        if potential_region_slug
+        else None
+    )
+
+    potential_language = (
+        Language.objects.filter(slug=slugs[1]).first() if len(slugs) > 1 else None
+    )
+    language = potential_language or Language.objects.get(slug="de")
+    if not region:
+        return root_socialmedia_headers(request, language, path or "")
```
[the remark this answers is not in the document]
sarahsporck: maybe we could check for reserved urls and then return 404. Otherwise https://webnext.integreat.app/recommend/de, which is imo a very good page might not get a good (search result) score.

integreat_cms/core/settings.py
```diff
@@ -108,6 +108,13 @@
     "INTEGREAT_CMS_WIKI_URL", "https://wiki.integreat-app.de"
 )
 
+#: Platform Name, used for app users and social media headers
+
+#: Social media preview image
+SOCIAL_PREVIEW_IMAGE = os.environ.get(
+    "INTEGREAT_SOCIAL_PREVIEW_IMAGE", f"{WEBAPP_URL}/social-media-preview.png"
+)
```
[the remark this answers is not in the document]
sarahsporck: Jep. That would make sense. Even better if this can be configured per page at some point. In case of pois & events we actually could already use the thumbnail image. What do you think?

integreat_cms/cms/fixtures/test_data.json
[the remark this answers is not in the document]
timobrembeck: Also for production, I think in this case I was referring to the translated part, so I wondered whether it really makes sense to show the German "Web-App | Lokale Informationen für Dich" as fallback for Arabic content or whether we should rather just stick with the branding title without the rest... I mean in most cases we will have the correct translation stored in the language model, I was just thinking about cases where it might not be.

integreat_cms/core/settings.py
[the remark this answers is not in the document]
timobrembeck: Cool idea, yes. Do we need to generate the thumbnails in a specific aspect ratio for the social media previews? At the moment, they're just 1:1...

integreat_cms/api/v3/socialmedia_headers.py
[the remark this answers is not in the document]
timobrembeck: Yes, I agree. Let's show the generic fallback preview for the special reserved slugs, and only handle real 404 errors differently.

integreat_cms/api/v3/socialmedia_headers.py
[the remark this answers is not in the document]
timobrembeck: Initially, I thought maybe it makes sense to show something like "Region not found"/"Language not available" as title and maybe a special error image for the social media preview to indicate that the links are (no longer) valid, but if this means that an existing social media preview can cause a page to be indexed by search engines, then it's not that optimal and it would be better to just not show any social media headers in this case.

integreat_cms/api/v3/socialmedia_headers.py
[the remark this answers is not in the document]
timobrembeck: Hmm, but shouldn't we update the reserved slugs setting anyways every time the web app adds a new internal route? To make sure there are no conflicts with existing regions/languages/pages?
Maybe we can document this somewhere in the app's docs to make app devs more aware that new routes require a small adaption in the cms?

integreat_cms/cms/templates/socialmedia_headers.html
timobrembeck: The way I understood nginx server side includes is that they include arbitrary content into the final html document before the response is sent to the client, right? So if nginx doesn't know how valid HTML looks like, we could also include a partial document, what could possibly go wrong? :see_no_evil: 

Change [index.ejs#L2-L17](https://github.com/digitalfabrik/integreat-app/blob/4dd43127ffae65ae1326640e67b5e6ab22ec2bc1/web/src/index.ejs#L2-L17) to:
```
<!-- Nginx Server Side Include template for dynamic social media previews -->
<!--# if expr="$render_title = yes" -->
<!--# include virtual="/proxy/socialmeta/$request_uri" -->
<!--# else -->
<html lang="">
<head>
    <title><%= config.appName %></title>
    <meta name="apple-mobile-web-app-title" content="<%= config.appName %>">
    <meta property="og:title" content="<%= config.appName %>" data-react-helmet="true">
    <meta property="og:site_name" content="<%= config.hostName %>">
    <meta name="twitter:title" content="<%= config.appName %>">
<!--# endif -->
```

And then on the cms side, we only render the opening tag of the `head` without closing it:
```
<html lang="{{ language_code }}">
<head>
    <title>{{ title }}</title>
    ...
```

integreat_cms/api/urls.py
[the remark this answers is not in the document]
sarahsporck: As the pr is merged I remove the wp-endpoint. No need to maintain it for new endpoints.

integreat_cms/cms/migrations/0084_language_socialmedia_webapp_title.py
sarahsporck: We should do this migration relatively soonish to merging this pr then.

integreat_cms/cms/fixtures/test_data.json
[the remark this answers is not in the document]
sarahsporck: @dkehne, I guess? We've had quite a long discussion about this, already. 

We could also just show the branding title as fallback for e.g. Arabic.

integreat_cms/core/settings.py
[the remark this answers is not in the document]
sarahsporck: https://stackoverflow.com/a/15560644 I guess this says enough about the size. But you can also specify the size of the page https://ogp.me/#structured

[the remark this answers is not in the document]
sarahsporck: I also just saw that Malte **does** have a different social media preview image. Therefore, I might have to move some stuff around again.

integreat_cms/cms/templates/socialmedia_headers.html
sarahsporck: Hmm. then we'd have to persuade @svenseeberg to always serve the socialmediaheaders - independently of the User-Agent. Otherwise we'd serve invalid html 😅 I'm not yet persuaded...

timobrembeck: Well, but that's what the `<!--# if expr="$render_title = yes" -->` is for, right?
So if the value is set, the dynamic social media headers are included, and if not, we use the hardcoded values from the `<!--# else -->` case... so we'd serve valid HTML in any case :nerd_face:

sarahsporck: Ah right. Me not `ejs`-Master 😅 
Thanks for explaining.

integreat_cms/api/v3/socialmedia_headers.py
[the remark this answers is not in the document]
sarahsporck: I've thought about this again. This would cause problems with urls where we detect the language and then redirect. Consider https://webnext.integreat.app/testumgebung or just https://webnext.integreat.app/ (which is the most crucial root url). These are both definitely valid. I see the comfort of having a region not found or language not found preview. But I'm not sure if this can actually be implemented.

integreat_cms/api/v3/socialmedia_headers.py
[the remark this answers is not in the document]
sarahsporck: Jep you're right. Hopefully this is always practiced as such 😅 I created an issue for documentation in the app repo https://github.com/digitalfabrik/integreat-app/issues/2683

integreat_cms/api/v3/socialmedia_headers.py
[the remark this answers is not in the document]
sarahsporck: I've added raising an exception in this case.

integreat_cms/api/urls.py
```diff
@@ -129,10 +138,91 @@
...
+                        name=f"social_root_reserved_default_language_{reserved}",
+                    ),
+                ]
+            ),
+        ),
+        settings.RESERVED_REGION_SLUGS,
+    ),
+    path(
+        "<slug:social_region_slug>", region_social_media_headers, name="social_region"
+    ),
```
timobrembeck, line 167:
```suggestion
```
This is now no longer needed, right? Because of the next route?

integreat_cms/api/urls.py
```diff
@@ -129,10 +138,91 @@
...
+                    name="social_region_reserved_local_news",
+                ),
+                re_path(
+                    r"^news/tunews/[\w-]+/$",
+                    region_social_media_headers,
+                    name="social_region_reserved_tunews",
+                ),
+                *map(
+                    lambda reserved: path(
+                        reserved,
```
timobrembeck, line 190:
```suggestion
                        f"{reserved}/",
```
Just for consistency, I'd also use a trailing slash here.

integreat_cms/cms/models/languages/language.py
```diff
@@ -108,6 +109,23 @@ class Language(AbstractBaseModel):
...
+        ),
+    )
+    social_media_webapp_description = models.TextField(
+        max_length=200,
+        blank=True,
+        verbose_name=_("Social media description"),
+        help_text=_(
+            "Displayed description of the WebApp in the search results and on social media pages (max 200 characters)."
+        ),
+    )
```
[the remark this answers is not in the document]
timobrembeck: No that looks good! :+1:

integreat_cms/core/settings.py
```diff
@@ -160,6 +160,11 @@
 #: The readable title of the branding
 BRANDING_TITLE: Final[str] = AVAILABLE_BRANDINGS[BRANDING]
 
+#: Social media preview image
+SOCIAL_PREVIEW_IMAGE = os.environ.get(
```
timobrembeck, line 164:
```suggestion
SOCIAL_PREVIEW_IMAGE: Final[str] = os.environ.get(
```

integreat_cms/cms/models/languages/language.py
```diff
@@ -108,6 +109,23 @@ class Language(AbstractBaseModel):
             _("This is used in exported PDFs."),
         ),
     )
+    social_media_webapp_title = models.CharField(
+        max_length=100,
+        blank=False,
+        default=settings.BRANDING_TITLE,
```
timobrembeck, line 115:
```suggestion
        blank=True,
```
I think adding the dynamic branding title as default value here is not ideal since it would require a database migration each time we change the setting.

timobrembeck (dismissed):
@sarahsporck awesome, thanks a lot for your changes! :pray: 

I just have a few minor comments, in general this looks very cool, definitely a huge leap forward for our SEO! :heart_eyes:

svenseeberg (commented):
I just noted that the output is not yet exactly what we need. I have not invenstigated the reason.

The html and head tags should not be there.

```html
<html lang="de-DE">
    <head>
        <title>Alltag - Augsburg | Integreat</title>
        <meta name="apple-mobile-web-app-title" content="Alltag - Augsburg | Integreat">
        <meta name="twitter:title " content="Alltag - Augsburg | Integreat">
        <meta name="twitter:card" content="summary">
        <meta property="og:title" content="Alltag - Augsburg | Integreat">
        
            <meta property="og:description" content="Gültiger LinkUngeprüfter LinkUngültiger LinkIgnorierter LinkInterner LinkFehlerhafter interner LinkE">
            <meta name="twitter:description" content="Gültiger LinkUngeprüfter LinkUngültiger LinkIgnorierter LinkInterner LinkFehlerhafter interner LinkE">
        
        <meta property="og:image" content="http://localhost:8000/static/logos/integreat/social-media-preview.png" />
        <meta property="og:image:width" content="1200" />
        <meta property="og:image:height" content="630" />
        <meta property="og:url" content="https://integreat.app/augsburg/de/alltag/" />
```

By default a `HttpResponse` should return an empty document. So its probably something that we do.

timobrembeck:
@svenseeberg the partial html response was my idea to be able to set the language tag of the html tag.
If we adapt the including template accordingly, it should work fine. Only the final result needs to be valid html, the preliminary header is just raw text and does not need to be parsed.
For more information about this context, read the existing discussion 🙈

timobrembeck (approved):
Finally, I think this is ready to be merged! :tada:

I opened a PR on the web app's side to adapt their part of the including template as well:

https://github.com/digitalfabrik/integreat-app/pull/2809

Apart from that, I made some further changes to ensure that the response is also a partial HTML response when a 404 error occurs and I changed the tests to compare the literal html output instead of converting the API result to json first.

@david-venhoff could you check again whether everything makes sense at it is now? Thanks! :+1:

And @svenseeberg do you have any concerns about this idea with the partial HTML response?

integreat_cms/api/v3/social_media_headers.py
```diff
@@ -0,0 +1,346 @@
...
+        language.bcp47_tag,
+        language.social_media_webapp_description,
+        url,
+    )
+
+
+@partial_html_response
+# pylint: disable=unused-argument
+def region_social_media_headers(
+    request: HttpRequest, region_slug: str, language_slug: str = settings.LANGUAGE_CODE
```
david-venhoff, line 188: Is there a reason we use the global default language here? It might make more sense to use the root language of the region

david-venhoff (approved):
:tada:
This looks good to me. I found one question, but I think this pr is already in a good state and should not be delayed much further.

integreat_cms/api/v3/social_media_headers.py
timobrembeck: Good point! I changed it accordingly:
https://github.com/digitalfabrik/integreat-cms/compare/b38c6e7a6adea1fb440224e5ad865b20da2192ab..5158344574eb3b9ab0f6b60b6383d3802faa6d2b

david-venhoff (approved):

svenseeberg:
> And @svenseeberg do you have any concerns about this idea with the partial HTML response?

No, should be okay :tada:

Issue #2300, opened 2023-05-24
Copying text outdates the HIX-value


osmers:
### Describe the Bug
When I copy a text, the HIX-value shows up as outdated afterwards

### Steps to Reproduce

1. Go to any page with content
2. Mark part of the text
3. Press Ctrl+C
4. See outdated HIX-value

### Expected Behavior
HIX-value stays up to date

### Actual Behavior
HIX value is outdated

PR #2401, opened 2023-08-17
Fix "HIX score outdated"

PeterNerlich:
### Short description

Any action within the tinymce editor modifies the content, and should thus set the content to be *dirty*. At least, this is the assumption of tinymce. But there are actions that do not modify the content, like making a selection, or end up at a previous state, like adding a character back after accidentally removing it.

This PR fixes this problem by simply monkeypatching tinymce's `isDirty()` and `setDirty()` methods, always comparing it to a previous value and saving that on successful autosaves.
Evaluating the HIX score also sets such a value to use as basis for determining whether an earlier score is actually outdated or not.

### Proposed changes

- Monkeypatch tinymce `isDirty()` and `setDirty()` to actually represent the content being different from the last saved state
    - tinymce provides this nice `startContent` property that is initialised to the initial content and then never used anywhere
    - We compare against that and set it after successful autosaves (using the actual content that was sent when the autosave was triggered)
- Have a similar value to compare whether the content is different from the last HIX evaluation and display the hints based on that

### Side effects

- `isDirty()` behaves differently now (not relying on the `isNotDirty` property anymore) and `setDirty()` tries to keep `isNotDirty` in line with what the new `isDirty()` would probably return if no changes to this are missed (this might get out of sync when an autosave is performed)

### Additional notes

- Where should I best put the monkeypatch?
- Do we want a monkeypatch at all? (It seems very logical to me in this case, but it still feels… dirty :upside_down_face:)

### Resolved issues

Fixes: #2300

### DIFF
--- a/integreat_cms/release_notes/current/unreleased/2300.yml
+++ b/integreat_cms/release_notes/current/unreleased/2300.yml
@@ -1,0 +1,2 @@
+en: HIX value is no longer instantly outdated, if only some text was selected
+de: HIX-Wert ist nicht mehr sofort veraltet, wenn Text lediglich ausgewählt wurde
--- a/integreat_cms/static/src/js/analytics/hix-widget.ts
+++ b/integreat_cms/static/src/js/analytics/hix-widget.ts
@@ -16,6 +16,8 @@ import { getCsrfToken } from "../utils/csrf-token";
 // See https://www.chartjs.org/docs/latest/getting-started/integration.html#bundlers-webpack-rollup-etc for details
 Chart.register(DoughnutController, ArcElement, CategoryScale, LinearScale, Tooltip, Title);
 
+let initialContent: string = null;
+
 const updateChart = (chart: Chart, value: number) => {
     const hixMaxValue = 20;
     const hixThresholdGood = 15;
@@ -78,13 +80,14 @@ const setHixLabelState = (state: string) => {
 const getHixValue = async () => {
     const updateButton = document.getElementById("btn-update-hix-value");
     let result;
+    const sentContent = getContent().trim();
     await fetch(updateButton.dataset.url, {
         method: "POST",
         headers: {
             "X-CSRFToken": getCsrfToken(),
         },
         body: JSON.stringify({
-            text: getContent(),
+            text: sentContent,
         }),
     })
         .then((response) => response.json())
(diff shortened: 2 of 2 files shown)


integreat_cms/release_notes/current/unreleased/2300.yml
```diff
@@ -0,0 +1,2 @@
+en: HIX value is no longer instandly outdated, if only some text was selected
```
seluianova, line 1:
```suggestion
en: HIX value is no longer instantly outdated, if only some text was selected
```

integreat_cms/static/src/js/forms/tinymce-init.ts
```diff
@@ -58,10 +58,25 @@ const toggleNoTranslate = (editor: Editor) => {
 
 export const getContent = (): string => tinymce.activeEditor.getContent();
 
+const monkeypatchEditor = () => {
```
seluianova, line 61: Could you please clarify for which case we need this monkeypatch ?
It seems to me that changing the listener of the contentChanged event does the trick. Or am I missing something?

PeterNerlich: True, if I didn't miss anything just now, it should suffice for #2300 specifically.
I made the monkey-patch for `isDirty()` because this seems to be where the functionality should reside, but I agree that doing a monkey-patch is also not really a clean approach to do this. But it looks to me like the #2132 (or the underlying mechanism) should/could make use of this functionality?

MizukiTemma: I opened a new PR #2435 because #2132 has stucked due to Firefox. If the new PR looks good then you don't have to think about influence on `dirty`.

MizukiTemma (changes requested):
Thank you for the PR 💪 It fixes the problem 🙂 

But some side effect was found: 
1. an unsaved warning will be shown when some changes are really made and the page is published or saved as draft. For example, add some words in a page and save it.
2. After leaving this unsaved warning by chosing "Verlassen" and saving the page several times, `IntegrityError` will be raised.

PeterNerlich:
I cut out everything not related to this issue and put it in #2476

integreat_cms/static/src/js/forms/tinymce-init.ts
PeterNerlich: This was removed and put (further polished) in #2476

MizukiTemma (approved):
Thank you for the updates 🙂 Looks good!

seluianova (approved):
LGTM!

Issue #381, opened 2020-04-24
Update internal links when translating (with XLIFF & DeepL)


svenseeberg:
If a page contains an internal link, the link target will not change during the translation process. As a result, users of other languages than the source language end up in the source language page. It should be our goal that users see the translated page. ~~There might be some exceptions, where authors intentionally want users to end up in one page that is valid for all languages (we could argue that in these cases the live content function could be used).~~ As we now have the fallback translations, we always want to link to the current language.

During XLIFF import, all internal links should be checked if they refer to another language. If yes, the link should be redirected to the current language of the translation. The importer function could scan for all internal links, look for the translated page, and substitute the link target.

a-tags that link to internal pages can get additional attributes to make the process more efficient, for example:
```html
<a href="https://integreat.app/testumgegbung/en/welcome" internal="true" pageid="5">
```

We can then use an xpath expression to scan all pages for elements that are internal links:
```xpath
//a/[@internal='true']
```

The fix_internal_links command should probably be extended to add the required attributes for all links that are pointing to internal pages.

timoludwig:
Yes, although I would differ between the generic search/replace and link search/replace, because the latter can be done with the lxml function `rewrite_links()`, see e.g.:
https://github.com/digitalfabrik/integreat-cms/blob/7d91b2a40cbea712d16f1a734a1321d6dea6caad/integreat_cms/cms/views/linkcheck/linkcheck.py#L170-L173

Probably we can generalize the solution for the link replacement of the broken link checker and use it for this issue and #1383.

karthik-integreat:
We had a discussion in the Service Team, that it would be necessary to have this for the Maps feature as well.

dkehne:
We got some more requests regarding that feature...guess we need to reschedule it because Q3/2022 is not realistic

osmers:
Any news on this feature and its possible implementation?

timoludwig:
> Any news on this feature and its possible implementation?

No, sadly nobody wanted to tackle this complex issue yet. I increased the prio of the ticket to increase the incentive a bit.

svenseeberg:
The scope changed. We want to update all links during XLIFF import. That should make things easier.

svenseeberg:
This should be done along with #1383

svenseeberg:
HIX and DeepL features are more important.

osmers:
Is this an issue we could tackle in Q3 possibly? Since HIX and DeepL are done now :)

[...]

PR #2403, opened 2023-08-18
Update internal links after translation

david-venhoff:
### Short description

This pr adds a feature to automatically fix internal links after translations.
I decided against adding something like `?id=x` to all internal links, since that would have to get recalculated after each save anyways.
I am not sure if something like a per-url opt-out button would be useful. Are there any use cases where a link to a different language is desired?
This implementation will change links to their fallback translation if there doesn't exists a url for the desired language and fallback translations are enabled. If fallback translations are disabled, the url will not be changed at all.

### Proposed changes

- Update internal links after a new content translation has been created
- Also update the link texts if the previous link text was the title of the linked content object
- Update the `fix_internal_links` command to also fix urls which are pointing to the wrong language

### Side effects

- Updating link texts will not work in most cases when translating with deepl, since deepl will usually change the link texts. 
- The `fix_internal_links` command will not change any link texts at all (though it was already like that before this pr, and we can probably assume that at least the link texts were correctly translated)
- If a page contains a short url that points to itself, it will be updated with every save, since the short url id changes per version. Is that a problem?

### Resolved issues

Fixes: #381


david-venhoff:
Not a dumb comment! I might just have misunderstood something. Basically, I am not sure how short urls help with the issue that urls are pointing to the wrong language after translation. 

In my comment I only referred to the idea of adding the content id to urls. `integreat.app/augsburg/de/willkommen?page_id=1` for example or `<a href="integreat.app/augsburg/de/willkommen" data-content-id=1>...</a>`. In that case, the id would have to be recalculated after each save, since the url could have changed. 

Are you thinking of removing the whole path, like `integreat.app/s/p/105/`? Because we already have this feature, called short urls. Those urls will indeed not break when a change in the tree hierarchy occurs, however they still refer to a concrete language.

MizukiTemma (approved):
Thank you so much for tackling this complicated issue 💪 
It looks good to me, working as expected 😸 

>I am not sure if something like a per-url opt-out button would be useful. Are there any use cases where a link to a different language is desired?

I don't think this is often the case. If needed, users can manually adjust the link target.

>Updating link texts will not work in most cases when translating with deepl, since deepl will usually change the link texts.
The fix_internal_links command will not change any link texts at all (though it was already like that before this pr, and we can probably assume that at least the link texts were correctly translated)

In my opinion we can leave it so.

>If a page contains a short url that points to itself, it will be updated with every save, since the short url id changes per version. Is that a problem?

I can't imagin any self-linking is needed (Except for the case of mistake. And I don't think it happens so often that this lead to a problem). So the current implementation won't harm.

integreat_cms/core/management/commands/fix_internal_links.py
```diff
@@ -138,50 +102,37 @@ def handle(self, *args, region_slug, username, commit, **options):
...
-            # Find a content object by translation url
-            objects = (
-                get_content_model(content_type)
-                .objects.filter(
-                    translations__slug=slugify(translation_slug, allow_unicode=True),
-                    translations__language__slug=language_slug,
-                    region__slug=region_slug,
-                )
-                .distinct()
+        for url in Url.objects.all():
```
timobrembeck, line 105: This feels like a lot of overhead... why not filtering out all potential internal urls first? We have ~80k URL objects in the database and only ~8k of them are internal.

integreat_cms/cms/forms/custom_content_model_form.py
```diff
@@ -125,6 +126,19 @@ def clean_content(self):
             link.attrib.pop("target", None)
             self.logger.debug("Removed target attribute from link: %r", tostring(link))
 
+        # Update internal links
+        for link in content.iter("a"):
+            if href := link.attrib["href"]:
```
timobrembeck, line 131:
```suggestion
            if href := link.attrib.get("href"):
```
I think an invalid link object without a `href` attribute should not cause an internal server error.

timobrembeck (changes requested):
Awesome, thanks a lot! :heart_eyes: 

Looks very promising, just a question: Do you think we need the separation of the short urls / long urls at all? In my opinion, we could just use the long URL as canonical URL and even replace short URLs with long URLs in the source code? I think this will also help the apps with their internal routing, because they don't need to implement an extra step to resolve short URLs as internal URLs...

> > I am not sure if something like a per-url opt-out button would be useful. Are there any use cases where a link to a different language is desired?
>
> I don't think this is often the case. If needed, users can manually adjust the link target.

Not necessarily, right? The way I understand it is that the code now overrides the link every time, no matter if a different language is selected, right? One possibility (which would also kind of make sense from a technical point of view) is abusing the "do not translate" tag for this? But not very easy to implement, so maybe we leave it as it is and only develop this functionality if it's actually needed and cities complain?

> > Updating link texts will not work in most cases when translating with deepl, since deepl will usually change the link texts.
> > The fix_internal_links command will not change any link texts at all (though it was already like that before this pr, and we can probably assume that at least the link texts were correctly translated)
>
> In my opinion we can leave it so.

Yes, I agree :+1:

> > If a page contains a short url that points to itself, it will be updated with every save, since the short url id changes per version. Is that a problem?
>
> I can't imagin any self-linking is needed (Except for the case of mistake. And I don't think it happens so often that this lead to a problem). So the current implementation won't harm.

Yes, I also think it won't be a problem, but as described above, I think it would also make sense to replace short urls with their long counterparts and only use short URLs where they are visible to the user or printed out somewhere...

integreat_cms/cms/utils/internal_link_utils.py
```diff
@@ -0,0 +1,172 @@
...
+
+    if target_translation := source_translation.foreign_object.get_public_translation(
+        target_language_slug
+    ):
+        title = (
+            target_translation.title
+            if source_translation.title.lower() == link_text.strip().lower()
+            else link_text
+        )
+        return url_types.get_url(url_type, target_translation), title
```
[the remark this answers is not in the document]
david-venhoff: So if I understand correctly, would you be in favor of changing the behavior to 1. drop the url type constant and 2. change the url to SLUG_URL, unless the link text is equal to the link href?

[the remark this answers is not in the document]
timobrembeck: I would even say we don't need the special case with "link text equals href"... Just a simple "Always convert the href to slug url" should be fine. Even if the link text is a short url, I think the href should still be a slug url.

integreat_cms/core/management/commands/fix_internal_links.py
david-venhoff: With this pr, the fix internal links command now also fixes short urls which point to the wrong language. Unfortunately they are not considered internal, since they have the backend domain, not the app domain.

integreat_cms/cms/utils/internal_link_utils.py
```diff
@@ -0,0 +1,174 @@
...
+        elif current_link.strip() == link_text.strip():
+            title = fixed_link
+
+        return fixed_link, title
+
+    return None
+
+
+WEBAPP_NETLOC = urlparse(settings.WEBAPP_URL).netloc
+CMS_NETLOC = urlparse(settings.BASE_URL).netloc
```
timobrembeck, line 66:
```suggestion
SHORT_LINKS_NETLOC = urlparse(settings.SHORT_LINKS_URL).netloc
```

integreat_cms/cms/utils/internal_link_utils.py
```diff
@@ -0,0 +1,174 @@
...
+            region_slug,
+            language_slug,
+            object_slug,
+        )
+
+    instance = instances[0]
+    return instance.get_public_translation(language_slug)
+
+
+def get_public_translation_cms(path):
```
timobrembeck, line 144:
```suggestion
def get_public_translation_from_short_link(path):
```

david-venhoff:
Oh, right 😅
I pushed the changes now.

timobrembeck (approved):
Thanks you :pray: 

I didn't have time to do an in-depth review again, but at first glance everything looks good! Let's give it a go :rocket:

Issue #2122, opened 2023-03-14
Remove usage of deprecated `distutils.utils.strtobool()`


timobrembeck:
### Motivation

We currently use a deprecated function `distutils.utils.strtobool()` which [will be removed in Python 3.12](https://peps.python.org/pep-0632/).

### Proposed Solution

Find an alternative or [include a re-implementation](https://github.com/drgarcia1986/simple-settings/pull/281/files/) in our own code base

### Alternatives

Don't support Python 3.12

PR #2407, opened 2023-08-27
Fix pylint deprecation warnings

ulliholtgrave:
### Short description

I was kinda annoyed by the `distutils` deprecation warning and solved them for our project by following the PEP 632 recommendations (see [https://peps.python.org/pep-0632/](https://peps.python.org/pep-0632/)).

### Proposed changes

- Add a general utils file to the top level of our project if our modules are needing more shared methods.

### Resolved Issues
Fixes: #2122 

### DIFF
--- a/integreat_cms/api/v3/locations.py
+++ b/integreat_cms/api/v3/locations.py
@@ -1,8 +1,6 @@
 """
 This module includes functions related to the locations/POIs API endpoint.
 """
-from distutils.util import strtobool
-
 from django.conf import settings
 from django.db.models import Prefetch
 from django.http import JsonResponse
@@ -12,6 +10,7 @@ from django.utils.html import strip_tags
 from ...cms.constants import status
 from ...cms.models import POICategoryTranslation
 from ...cms.models.pois.poi import get_default_opening_hours
+from ...core.utils.strtobool import strtobool
 from ..decorators import json_response
 from .location_categories import transform_location_category
 
--- a/integreat_cms/core/settings.py
+++ b/integreat_cms/core/settings.py
@@ -10,7 +10,6 @@ See :doc:`/prod-server` for details.
 """
 
 import os
-from distutils.util import strtobool
 from urllib.parse import urlparse
 
 from django.core.exceptions import ImproperlyConfigured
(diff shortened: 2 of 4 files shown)


integreat_cms/default_utils.py
timobrembeck, line 1: Just a matter of preference, but I think plain `utils.py` is sufficient as filename, no need for the `default_` prefix.
We also used this convention in the other Django apps, so we either have a single `utils.py` file or a `utils` directory containing multiple utils with prefixes...

integreat_cms/release_notes/current/unreleased/2406.yml
timobrembeck, line 1: Implementation details are irrelevant for the end users and other teams, so I would suggest to remove this release note to stay consistent with our previous guidelines.

timobrembeck (changes requested):

integreat_cms/utils.py
timobrembeck, line 1: Sorry, I think I have expressed myself a little misleadingly... I think the location of the file inside the `core` directory was better than in the root directory, I just suggested it to rename it. So now the name is better but I'd suggest to move it back into the core dir :see_no_evil:

integreat_cms/api/v3/locations.py
```diff
@@ -12,6 +10,7 @@
 from ...cms.constants import status
 from ...cms.models import POICategoryTranslation
 from ...cms.models.pois.poi import get_default_opening_hours
+from ...default_utils import strtobool
```
timobrembeck, line 13:
```suggestion
from ...core.utils import strtobool
```

integreat_cms/core/settings.py
```diff
@@ -10,13 +10,13 @@
...
 import os
-from distutils.util import strtobool
 from urllib.parse import urlparse
 
 from django.core.exceptions import ImproperlyConfigured
 from django.utils.translation import gettext_lazy as _
 
 from ..nominatim_api.utils import BoundingBox
+from ..utils import strtobool
 from .logging_formatter import ColorFormatter, RequestFormatter
```
timobrembeck, line 20:
```suggestion
from .logging_formatter import ColorFormatter, RequestFormatter
from .utils import strtobool
```

timobrembeck (changes requested):

timobrembeck (approved):

PeterNerlich (approved):
LGTM!

Issue #2365, opened 2023-07-25
Remove archived pages from list of mirroring pages


osmers:
### Describe the Bug
A page lists the names of pages that mirror the page - this list includes archived pages

### Steps to Reproduce

1. Go to https://admin.integreat-app.de/integreat-inhalte/pages/de/10947/edit/
2. Scroll down to the list of pages that mirror the page
3. Click on the link for Neuss
4. See that the page is archived

### Expected Behavior
List shows only pages that are live

### Actual Behavior
List also shows archived pages

### Additional Information
Showing archived pages creates the impression that there are many more pages that are mirroring the content while only live pages are interesting and should prevent the page being archived itself. If any archived pages mirror a page, and the mirrored pages should be deleted, then the mirrored page should be force removed from the archived page.

PR #2409, opened 2023-09-07
Remove archived pages from mirorring page list

MizukiTemma:
### Short description

This PR removes archived pages from the list of mirroring pages.

### Proposed changes

<s>- Exclude explicitly or implicitly archived pages from `page.mirroring_pages`
- Change the field `mirrored_page` deletable
- Allow page deletion/archiving when there is no mirroring page or only mirorring pages that are archived (in other word: prevent deletion/archiving only when there are non-archived mirorring pages). </s>
- Make migration to set `mirrored_page` to `None` if a page is archived or belongs to an archived region.
- Set `mirrored_page` of a page object to `None` when it gets archived or belongs to a region which gets archived 

### Side effects

- When a page A is mirroring another page X and A gets archived, now X can be archived too because A is archived. But when A gets restored, the problem occurs that A is mirroring after the restration not X but the first page of the region, which was exactly the reason why we wanted to prevent archiving of mirrored pages and show mirroring pages as list. Therefore the field `mirrored_page` gets set to `None` when a page becomes archived. 

### Resolved issues

Fixes: #2365 


david-venhoff (approved):
This looks mostly good to me. I only wonder whether adding `mirroring_live_pages` is the best option. Since this pr make sure that there can't be archived pages which mirror another page, filtering archived pages out of the mirroring pages seems a bit unnecessary to me. We would only have to add a migration that sets `mirrored_page` to `None` for all archived pages.

MizukiTemma:
@david-venhoff that's a good idea 😃 If the next reviewes are for it, I'll change the implementation in that way.

integreat_cms/cms/models/pages/page.py
```diff
@@ -351,6 +355,15 @@ def restore(self):
...
                     translation.save(update_timestamp=False)
 
+    @cached_property
+    def mirroring_live_pages(self):
+        """
+        Returns live pages that are mirroring the given page. Archived pages are excluded.
+        """
+        all_mirroring_pages = self.mirroring_pages.all()
+        archived_page_ids = [page.id for page in all_mirroring_pages if page.archived]
+        return all_mirroring_pages.exclude(id__in=archived_page_ids)
```
timobrembeck, line 365: >> This looks mostly good to me. I only wonder whether adding mirroring_live_pages is the best option. Since this pr makes sure that there can't be archived pages which mirror another page, filtering archived pages out of the mirroring pages seems a bit unnecessary to me. We would only have to add a migration that sets mirrored_page to None for all archived pages.
>
> @david-venhoff that's a good idea 😃 If the next reviewes are for it, I'll change the implementation in that way.

I agree as well :innocent:

timobrembeck (changes requested):
Thanks a lot! :+1:

I also think it would be a good idea to keep the existing database consistent with the new behavior and remove all existing live content from archived pages.

Also, we could go one step further and also remove live content from public pages in archived regions? I guess this would also require to remove live content while archiving a region, which could be a bit cumbersome...

integreat_cms/cms/models/pages/page.py
```diff
@@ -136,7 +136,7 @@ class Page(AbstractTreeNode, AbstractBasePage):
         "self",
         null=True,
         blank=True,
-        on_delete=models.PROTECT,
+        on_delete=models.SET_NULL,
```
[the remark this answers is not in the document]
MizukiTemma: I was confused and misunderstood some stuffs 🙈 yes, this is unnecessary 😅

MizukiTemma:
@timobrembeck @david-venhoff 
Your sugestions are now applied 😸

integreat_cms/cms/migrations/0079_alter_page_mirrored_page.py
```diff
@@ -0,0 +1,61 @@
...
+
+    Page = apps.get_model("cms", "Page")
+
+    pages_archived_region = Page.objects.filter(region__status=region_status.ARCHIVED)
+    pages_archived_region.update(mirrored_page=None)
+
+    pages_explicitely_archived = Page.objects.filter(explicitly_archived=True)
+    pages_explicitely_archived.update(mirrored_page=None)
+
+    update_implicitly_archived_pages(apps, schema_editor, pages_explicitely_archived)
```
timobrembeck, line 47:
```suggestion

    pages_archived = [
        Page.objects.filter(
            tree_id=page.tree_id, lft__range=(page.lft, page.rgt - 1)
        ).order_by()
        for page in Page.objects.filter(explicitly_archived=True)
    ]

    page_ids = pages_archived_region.union(*pages_archived).values_list("id", flat=True)

    Page.objects.filter(id__in=page_ids).update(mirrored_page=None)
```
I think this can be optimized a bit.
We don't have access to the page model's methods like `get_tree()`, but we can use the [queryset filter in its source definition](https://github.com/django-treebeard/django-treebeard/blob/f9674d0adf2ae3077c4475895f0df074179102c7/treebeard/ns_tree.py#L627-L629).

integreat_cms/cms/migrations/0079_alter_page_mirrored_page.py
```diff
@@ -0,0 +1,61 @@
...
+    :param pages: The pages that are archived
+    :type pages: ~django.db.models.query.QuerySet [ ~integreat_cms.cms.models.Page ]
+    """
+
+    Page = apps.get_model("cms", "Page")
+    if child_pages := Page.objects.filter(parent__in=pages):
+        child_pages.update(mirrored_page=None)
+        update_implicitly_archived_pages(apps, schema_editor, child_pages)
+
+
```
timobrembeck, line 27:
```suggestion
```
Then we don't need this function as well...

timobrembeck (changes requested):

MizukiTemma:
It seems removing of archived pages does not work if redis is activ. I'll do an update for that.

MizukiTemma:
Removing of the target pages does not work correctly/quickly if redis is activated. Cache invalidation is added.

integreat_cms/cms/migrations/0079_alter_page_mirrored_page.py
```diff
@@ -0,0 +1,52 @@
...
+    :param schema_editor: The database abstraction layer that creates actual SQL code
+    :type schema_editor: ~django.db.backends.base.schema.BaseDatabaseSchemaEditor
+    """
+
+    Page = apps.get_model("cms", "Page")
+    Region = apps.get_model("cms", "Region")
+
+    pages_archived_region = Page.objects.filter(region__status=region_status.ARCHIVED)
+    invalidate_model(Region)
```
timobrembeck, line 25:
```suggestion
```

This probably needs to be moved to the end of the function to take effect, right?
I merged the `pages_archived_region` pages with the archived pages later to do the update in one combined query.

integreat_cms/cms/migrations/0079_alter_page_mirrored_page.py
```diff
@@ -0,0 +1,52 @@
...
+            tree_id=page.tree_id, lft__range=(page.lft, page.rgt - 1)
+        ).order_by()
+        for page in Page.objects.filter(explicitly_archived=True)
+    ]
+
+    page_ids = pages_archived_region.union(*pages_archived).values_list("id", flat=True)
+
+    Page.objects.filter(id__in=page_ids).update(mirrored_page=None)
+
+    invalidate_model(Page)
```
timobrembeck, line 38:
```suggestion
    invalidate_model(Region)
    invalidate_model(Page)
```

integreat_cms/release_notes/current/unreleased/2365.yml
```diff
@@ -0,0 +1,2 @@
+en: Remove archived pages from list of mirroring pages
+de: Entferne archivierte Seiten aus der Liste der spiegelnden Seiten
```
timobrembeck, line 2:
```suggestion
de: Entferne archivierte Seiten aus der Liste der gespiegelten Seiten
```

timobrembeck (changes requested):

timobrembeck (approved):
:tada:

Issue #2398, opened 2023-08-14
Add field for "only by appointment" to opening hours for POIs


osmers:
### Motivation
Some POIs are only open with prior appointment - this option is so far missing when specifying the opening hours. So we should add it.

### Proposed Solution
After clicking on "edit all opening hours" there should be another check box which allows users to specify that on that day the POI is only open with prior appointment. The link for booking the appointment can then either be in the content or we implement a text field that allows websites as an input.

### Alternatives
Leave it as it is which would mean that as soon as opening hours are edited, the days with prior appointment are either shown as open or closed but not with the necessary information for users.

### User Story
As a user of the Integreat App I want to see on which days I need an appointment to go to a POI so that I do not waste my time and have all the relevant information before going.
As a user of the CMS I want to have the possibility of including this kind of opening conditions so that our users get all the relevant information and do not just show up.

### Design Requirements
Not sure if much design is needed here - maybe a decision about whether we should have the information for appointment booking in the text or within the opening hours window and how to display it in the app, if it is part of the opening hours.

LeandraH:
@hauf-toni Do we have a plan for how this shows up in the app yet?

LeandraH:
@hauf-toni Sorry to have tagged you in this one, I'll tag you in the app one instead 😅

ztefanie:
@charludo Is this done? When will it approximately be deployed?

charludo:
> @charludo Is this done? When will it approximately be deployed?

Yep, it's already been merged and live on the test server, and I assume it's going to be included in the next release.

PR #2416, opened 2023-09-21
Add 'by appointment only' setting to POI opening hours

charludo:
### Short description

Add a per-day setting "By appointment only", as well as a URL field for a link where appointments can be booked to the opening hours settings of POIs.

### Proposed changes

- add fields in the model and templates
- adjust fixtures and testdata files

### Side effects

- none that I found 

### Resolved issues

Fixes: #2398 


integreat_cms/cms/constants/opening_hours.py
```diff
@@ -30,6 +31,6 @@
                 },
             },
         },
-        "required": ["allDay", "closed", "timeSlots"],
+        "required": ["allDay", "closed", "appointmentOnly", "timeSlots"],
```
charludo, line 34: I just noticed that this change is probably not a good idea - old POIs can no longer be saved. Reverting this would also allow us to leave all the fixtures and expected tes outputs unchanged, significantly reducing the diff of this PR.

The only alternative I can think of is using some JS to **always** mark the Opening Hours as changed, which is not exactly great either.

Would appreciate some input.

timobrembeck: Hmm, I think we should keep this line, because it also helps the app parsing the information if they know the exact JSON schema that they receive...
So I'd propose to write a data migration to change the JSON of all existing POIs... sure, more work in the short term, but at least we have a clear database state.

integreat_cms/static/src/js/pois/opening-hours/component/opening-hours-input-fields.tsx
```diff
@@ -79,9 +79,10 @@ const OpeningHoursInputFields = ({ openingHoursState, selectedDaysState, transla
         // Create new object instead of reference to existing one
         const newOpeningHoursBuffer = deepCopy(openingHoursBuffer);
         newOpeningHoursBuffer.closed = (target as HTMLInputElement).checked;
-        // If closed was checked, disable all day
+        // If closed was checked, disable all day and appointment only
         if (newOpeningHoursBuffer.closed) {
             newOpeningHoursBuffer.allDay = false;
+            newOpeningHoursBuffer.appointmentOnly = false;
```
timobrembeck, line 85: I think in line 99, we need the same statement, right? Otherwise the "appointment" checkbox stays set when "all day" is selected...

integreat_cms/static/src/js/pois/opening-hours/component/opening-hours-input-fields.tsx
```diff
@@ -101,6 +102,21 @@ const OpeningHoursInputFields = ({ openingHoursState, selectedDaysState, transla
...
 
+    // When the appointment only checkbox is changed, write the changes to the buffer
+    const updateAppointmentOnly = ({ target }: Event) => {
+        console.debug("Changed appointment only checkbox:", target);
+        // Create new object instead of reference to existing one
+        const newOpeningHoursBuffer = deepCopy(openingHoursBuffer);
+        newOpeningHoursBuffer.appointmentOnly = (target as HTMLInputElement).checked;
+        // If appointment only was checked, disable closed
+        if (newOpeningHoursBuffer.appointmentOnly && newOpeningHoursBuffer.closed) {
+            newOpeningHoursBuffer.closed = false;
```
timobrembeck, line 113:
```suggestion
        // If appointment only was checked, disable closed and all day
        if (newOpeningHoursBuffer.appointmentOnly) {
            newOpeningHoursBuffer.closed = false;
            newOpeningHoursBuffer.allDay = false;
```

timobrembeck (changes requested):
Awesome, thanks a lot! :tada:

integreat_cms/static/src/js/pois/opening-hours/component/opening-hours-input-fields.tsx
charludo: I don't think so - in the case that the location is open around the clock, you could conceivably still need an appointment. But if it is closed, it's closed, appointment or not.

timobrembeck: Really? As a user, I would consider these two options contradictory.
Either, I can come without appointment (then I have to come within the correct time slots or at any time if the location is open 24h), or I need an appointment, but then I don't care about the actual opening hours.
To me, it could create misunderstandings if a location is both "open all day" and "open only by appointment"...

charludo: But what's the difference to e.g. 8:00-17:00, only with appointment? To me those are separate things, one tells me when the place is open, and the other if I need to book an appointment for within that time range.

Setting both appointment only and open all day then just means: Come any time, provided you've called beforehand.

But maybe @osmers can chime in on this? 😄

timobrembeck: Ah ok, I missed that it's possible to set time slots while "only by appointment" is selected. In my opinion, this shouldn't be possible as well. The service can tell when they have time for an appointment via a third party timeslot booking system or manually via phone or email. Otherwise, I think people might just come during those hours even if they have no appointment. But ok, that's just my perspective, I'm open to input here.

osmers: I think @charludo is right bcs there are a lot of official places that write opening hours and by appointment only at the same time: 
image
I think by appointment only should be selectable when opening times are put in as well as when it's closed all day. And it obviously needs to be selectable on a day to day basis. So Monday should be able to be different from Tuesday and so on.

charludo: Alright, so no limitations at all then? Including "closed all day, by appointment only"?

osmers: I mean, it would make sense to let the municipalities decide, right? If that is how opening hours are communicated in their area, why not?

charludo: OK!

timobrembeck: Ok, as long as this "appointment only" hint is also shown in the app next to the time slot, I agree that this is the most informative and does not have much potential for misunderstandings :+1:

osmers: @timobrembeck true - it definitely needs to be displayed next to the specific day
Is there already an app issue for this feature?

integreat_cms/cms/migrations/0081_poi_appointment_url.py
```diff
@@ -0,0 +1,26 @@
...
+        migrations.AddField(
+            model_name="poi",
+            name="appointment_url",
+            field=models.URLField(
+                blank=True,
+                help_text="Link to an external website where an appointment for this location can be made.",
+                max_length=500,
+                verbose_name="appointment link",
+            ),
+        ),
```
timobrembeck, line 25:
```suggestion
        ),
        migrations.RunPython(set_default_appointment_only),
```
Since the two migrations relate to each other, I think we can move them to the same file.

integreat_cms/cms/migrations/0081_poi_appointment_url.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 3.2.20 on 2023-09-21 08:22
+
+from django.db import migrations, models
+
+
```
timobrembeck, line 5:
```suggestion

# pylint: disable=unused-argument
def set_default_appointment_only(apps, schema_editor):
    """
    Set default value for appointmentOnly
    :param apps: The configuration of installed applications
    :type apps: ~django.apps.registry.Apps
    :param schema_editor: The database abstraction layer that creates actual SQL code
    :type schema_editor: ~django.db.backends.base.schema.BaseDatabaseSchemaEditor
    """
    POI = apps.get_model("cms", "POI")
    for obj in POI.objects.all():
        for item in obj.opening_hours:
            item["appointmentOnly"] = False
        obj.save()


```

integreat_cms/cms/migrations/0081_poi_appointment_url.py
```diff
@@ -0,0 +1,26 @@
...
+class Migration(migrations.Migration):
+    """
+    Add field for setting appointment link
```
timobrembeck, line 8:
```suggestion
    Add field for setting appointment link and
    set default appointment only opening hours for all existing POIs
```

integreat_cms/cms/migrations/0082_set_default_appointment_only.py
timobrembeck, line 1: Then this file can be deleted

integreat_cms/static/src/js/pois/opening-hours/component/opening-hours-display.tsx
```diff
@@ -35,6 +35,7 @@ const OpeningHoursDisplay = ({ openingHours, setSelectedDays, translations, days
                     <div class="text-right">
                         {dayOpeningHours.closed && translations.closedLabel}
                         {dayOpeningHours.allDay && translations.allDayLabel}
+                        {dayOpeningHours.appointmentOnly && translations.appointmentOnlyLabel}
```
timobrembeck, line 38: Maybe we need a new line in case closed or open all day is checked as well?

Screenshot 2023-10-12 at 11-40-02 Integreat Redaktionssystem

timobrembeck (changes requested):
Thank you! :+1:

Just two small issues remaining...

integreat_cms/static/src/js/pois/opening-hours/component/opening-hours-display.tsx
charludo: My solution avoids unwanted line breaks at the cost of some empty `span`s. Please lmk if that's OK.

timobrembeck: Yes, looks good! :+1:

timobrembeck (approved):
Thanks! :pray:

Could you squash the commits before merging?

david-venhoff (approved):
Looks good to me. But shouldn't the `appointment_url` be also delivered in the api? Otherwise I am not sure what the purpose of it is.

charludo:
> Looks good to me. But shouldn't the `appointment_url` be also delivered in the api? Otherwise I am not sure what the purpose of it is.

Whoops, yes, good point! Added that, if you'd like to take another look before the merge.

integreat_cms/api/v3/locations.py
```diff
@@ -93,6 +93,7 @@ def transform_poi_translation(poi_translation):
         "temporarily_closed": poi.temporarily_closed,
         # Only return opening hours if not temporarily closed and they differ from the default value
         "opening_hours": opening_hours,
+        "appointment_url": poi.appointment_url or None,
```
timobrembeck, line 96: Could you add this to the API documentation as well?

integreat_cms/cms/models/pois/poi.py
```diff
@@ -87,6 +90,15 @@ class POI(AbstractContentModel):
             _("The opening hours remain and are only hidden."),
         ),
     )
+    appointment_url = models.URLField(
+        max_length=500,
+        blank=True,
+        null=True,
```
timobrembeck, line 96:
```suggestion
```
It doesn't make a huge difference, but I think `null=True` for char fields rarely makes sense, since it's only `None` until the form has been submitted once, after that it's an empty string.
So it leads to a bit of an inconsistent database where some fields are `None` and some fields are `""` when the URL is not set.
Anyways, not super important, just something to keep in mind I guess.

Issue #2319, opened 2023-06-13
Incorrect translation status after machine translation (outdated pages)


MizukiTemma:
### Describe the Bug

Translation status of "outdated" pages does not transit correctly after machine translation.

### Steps to Reproduce
Sadly I cannot reproduce this bug so far 😿 

### Expected Behavior

After machine translation the status changes from "outdated" (alert-triangle icon) to "machine translation" (robot icon)

### Actual Behavior

After machine translation the status changes from "outdated" (alert-triangle icon) to "currently in translation" (clock icon)

### Additional Information

Similar bug is reported #2318 

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

vthejay:
For example, Stadt Regensburg reported that the status of the sides (without previous translations) are saved as "Drafts" after a machine translation bulk action. For example:
The page https://admin.integreat-app.de/regensburg/pages/de/277/edit/
the translation in English was saved as Draft after the bulk action.

PR #2418, opened 2023-09-21
Fix transit of translation status icons

MizukiTemma:
### Short description

This PR fixes the problem that translation status icons sometimes do not transit correctly and quickly.

The problem can be observed when Redis is activated and especially in/after transit to/from `currently_in_translation`. Therefore I assume that cache of the field `currently_in_translation`, which suppresses other translation statuses, is the cause.

### Proposed changes

- Invalidate cache if Redis is activated.

### Side effects

- As I could not reproduce the exact cases reported in the issues, this solution may not cover all the cases.

### Resolved issues

Fixes: #2318
Fixes: #2319 

### DIFF
--- a/integreat_cms/core/settings.py
+++ b/integreat_cms/core/settings.py
@@ -307,8 +307,11 @@ INSTALLED_APPS = [
     "widget_tweaks",
 ]
 
+#: Check whether redis is activated
+REDIS_CACHE = bool(strtobool(os.environ.get("INTEGREAT_CMS_REDIS_CACHE", "False")))
+
 # Install cacheops only if redis cache is available
-if "INTEGREAT_CMS_REDIS_CACHE" in os.environ:
+if REDIS_CACHE:
     INSTALLED_APPS.append("cacheops")
 
 # The default Django Admin application and debug toolbar will only be activated if the system is in debug mode.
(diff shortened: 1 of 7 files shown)


MizukiTemma:
After @david-venhoff 's advice, invalidation by `invalidated_update` is being tried, but I have no idea why the test is failing with attribute errors while this does not occur while using XLIFF Export/Import etc.

timobrembeck (commented):
Didn't try out, but seems reasonable.
I'd just suggest to add a new setting for the redis cache, e.g. 
```
REDIS_CACHE = bool(strtobool(os.environ.get("INTEGREAT_CMS_REDIS_CACHE", "False")))
```

And then use this setting everywhere.

MizukiTemma:
@timobrembeck 
Nice idea 🙂 Applied

integreat_cms/core/settings.py
```diff
@@ -922,8 +922,11 @@
     },
 }
 
+#: whether redis is activated
+REDIS_CACHE = bool(strtobool(os.environ.get("INTEGREAT_CMS_REDIS_CACHE", "False")))
```
timobrembeck, line 926: Do you think you could move this up, so we could already profit from this in the following line?
https://github.com/digitalfabrik/integreat-cms/blob/78fbad03983656514b245ca58080cab34a326ce3/integreat_cms/core/settings.py#L311-L312

Otherwise we might get inconsistencies when we use e.g. the string "False" as environment variable, since `cacheops` will be enabled even if redis is disabled.

timobrembeck (approved):
Thanks! :+1:
After squashing the commits, this is good to go! :rocket:

integreat_cms/core/settings.py
MizukiTemma: Done 🙂 I hope this looks okay.

ulliholtgrave (approved):
Nice, works as expected :+1:

Issue #1836, opened 2022-11-07
"Last updated" field in region list should show the date of the last changed content


ulliholtgrave:
### Motivation

Right now the "last updated" field in the region list shows the date when the region form got last updated which isn't really an important information. Instead we should show the the date on which the content of the region got changed last to see how active the region is.

PR #2419, opened 2023-09-23
Change 'last updated' field in region list to refer to their content

charludo:
### Short description

Change 'last updated' field in region list to each region's content instead of the region object itself.

### Proposed changes

- add `last_content_update` cached property to region model
- use it instead of `last_updated` in the region list

### Side effects

- none I could find, keeping the `last_updated` field unmodified should prevent any backwards compatibility issues

### Resolved issues

Fixes: #1836 

### DIFF
--- a/integreat_cms/cms/models/regions/region.py
+++ b/integreat_cms/cms/models/regions/region.py
@@ -818,6 +818,42 @@ class Region(AbstractBaseModel):
             },
         )
 
+    @cached_property
+    def last_content_update(self):
+        """
+        Find the latest date at which any content of the
+        region has been modified.
+
+        :return: the last content update date
+        :rtype: datetime.datetime
+        """
+        latest_page_update = self.pages.aggregate(
+            latest_update=models.Max("translations__last_updated")
+        )["latest_update"]
+        latest_poi_update = self.pois.aggregate(
+            latest_update=models.Max("translations__last_updated")
+        )["latest_update"]
+        latest_event_update = self.events.aggregate(
+            latest_update=models.Max("translations__last_updated")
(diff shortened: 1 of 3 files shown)


integreat_cms/cms/models/regions/region.py
```diff
@@ -818,6 +818,32 @@ def backend_edit_link(self):
...
         )
 
+    @cached_property
+    def last_content_update(self):
+        """
+        Find the latest date at which any content of the
+        region has been modified.
+
+        :return: the last content update date
+        :rtype: str
```
david-venhoff, line 828:
```suggestion
        :rtype: datetime.datetime
```

integreat_cms/cms/models/regions/region.py
```diff
@@ -818,6 +818,32 @@ def backend_edit_link(self):
...
+        """
+        latest_page_update = self.pages.aggregate(
+            latest_update=models.Max("translations__last_updated")
+        )["latest_update"]
+        latest_poi_update = self.pois.aggregate(
+            latest_update=models.Max("translations__last_updated")
+        )["latest_update"]
+        latest_event_update = self.events.aggregate(
+            latest_update=models.Max("translations__last_updated")
+        )["latest_update"]
```
david-venhoff, line 838: Nitpicky, but maybe this should also include the imprint?

david-venhoff (approved):
Thanks!

timobrembeck (approved):
Thanks! :+1:
Apart from @david-venhoff's comments, I don't have any further objections, except maybe adding a changelog entry for it :sweat_smile:

Issue #2363, opened 2023-07-25
Outdated organization properties delivered in endpoints


steffenkleinle:
### Describe the Bug
If properties of an organization are changed and saved (e.g. website), the pages that have that organization set still have the outdated properties in the endpoints delivered (e.g. the childrens endpoint). Pages have to be manually updated for the changes to take effect.

### Steps to Reproduce

1. Go to https://cms-test.integreat-app.de/testumgebung/organizations/t%C3%BCr-an-t%C3%BCr/edit/
2. Change the website to https://example.com/home
3. Click on save
4. See old website on https://webnext.integreat.app/testumgebung/de/arztbesuch-2

### Expected Behavior
Not every page with an organization set has to be updated manually for the changes to take effect.

### Actual Behavior
Every page with an organization set has to be updated manually for the changes to take effect.

### Additional Information

> Ah, das liegt vermutlich am Cache - wenn Orgas geupdated werden, müsste man den page-Cache invalidieren, damit die neuen Werte ausgeliefert werden

PR #2422, opened 2023-09-23
Invalidate cache when organization was saved

david-venhoff:
### Short description

Another caching related bugfix, where the automatic invalidation system did not invalidate PageTranslations when an organization changed.

### Proposed changes

- Invalidate all PageTranslations when an organization gets saved

### Side effects

- Also unrelated PageTranslations will not be cached after an organization gets saved

### Resolved issues

Fixes: #2363 

### DIFF
--- a/integreat_cms/core/signals/organization_signals.py
+++ b/integreat_cms/core/signals/organization_signals.py
@@ -1,0 +1,21 @@
+"""
+This module contains signal handlers related to organization objects.
+"""
+
+from cacheops import invalidate_model
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from ...cms.models import Organization, Page, PageTranslation
+
+
+@receiver(post_save, sender=Organization)
+def organization_create_handler(**kwargs):
+    r"""
+    Invalidate page translation cache after organization creation
+
+    :param \**kwargs: The supplied keyword arguments
+    :type \**kwargs: dict
+    """
+    invalidate_model(Page)
+    invalidate_model(PageTranslation)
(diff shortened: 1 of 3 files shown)


timobrembeck (approved):
Didn't try to reproduce the bug locally (so didn't test the fix either), but seems reasonable :+1:

integreat_cms/core/signals/organization_signals.py
```diff
@@ -0,0 +1,20 @@
...
+
+@receiver(post_save, sender=Organization)
+def organization_create_handler(**kwargs):
+    r"""
+    Invalidate page translation cache after organization creation
+
+    :param \**kwargs: The supplied keyword arguments
+    :type \**kwargs: dict
+    """
+    invalidate_model(PageTranslation)
```
MizukiTemma, line 20:
```suggestion
    invalidate_model(Page)
    invalidate_model(PageTranslation)
```

You have to invalidate `Page`, because the organization field is stored in the page objects.
I tested with the children endpoint by changing the website field of an organization and it didn't work. After adding this line it worked.

MizukiTemma (changes requested):

integreat_cms/core/signals/organization_signals.py
david-venhoff: Oh. I tested with the specific page endpoint and there invalidating `PageTranslation` fixes the problem, but invalidating `Page` does not. I did not expect the behavior to differ between those endpoints and only tested the specific page endpoint. So to fix the problem completely, actually both models would need to be invalidated :/

MizukiTemma: What.........Me neither have expected that 😅

MizukiTemma (approved):
Thank you 🎉

PR #2426, opened 2023-09-23
Update dependencies

ulliholtgrave:
### Short description

Update our dependencies to fix ESLint memory problems.

### Side effects

- @david-venhoff I had to change your heartbeat implementation, but I am not really sure if I applied the correct changes for your implementation to work. Maybe you can review this and apply changes if necessary :)


MizukiTemma (approved):

integreat_cms/static/src/js/content-edit-lock.ts
```diff
@@ -28,7 +28,7 @@ const sendTakeOverMessage = async (url: string, payload: string) => {
 };
 
 const sendHeartbeat = async (heartbeatData: HTMLElement) => {
-    clearInterval(heartbeatInteval);
+    clearInterval(heartbeatInterval.ref.name);
```
timobrembeck, line 31: I don't really understand this change, why was it necessary?
With this line, I get the error:

> Uncaught (in promise) TypeError: heartbeatInterval is null

In the JS console.

Also, everywhere in the documentations about the `setInterval()`/`clearInterval()` functions it seems as we have to pass the return value of `setInterval()` directly to `clearInterval()` without any attributes:

```
intervalId = setInterval(function() { ... }, ...);
...
clearInterval(intervalId);
```

ulliholtgrave: I know and I am also very confused. But if you revert the change I get an error that this can't be assigned and webpack won't compile the js code :/

timobrembeck: I guess it's about the type of the variable.
Apparently, the return type was changed from `NodeJS.Timer` to `NodeJS.Timeout`...
We can prevent such errors in the future if we just use `ReturnType<typeof setTimeout>`, this should be platform independent...

integreat_cms/static/src/js/content-edit-lock.ts
```diff
@@ -7,7 +7,7 @@ import { showConfirmationPopupWithData } from "./confirmation-popups";
 import { storeDraft } from "./forms/tinymce-init";
 import { getCsrfToken } from "./utils/csrf-token";
 
-let heartbeatInteval: NodeJS.Timer | null = null;
+let heartbeatInterval: NodeJS.Timer | null = null;
```
timobrembeck, line 10:
```suggestion
let heartbeatInterval: ReturnType<typeof setTimeout> | null = null;
```

integreat_cms/static/src/js/content-edit-lock.ts
timobrembeck:
```suggestion
    clearInterval(heartbeatInterval);
```

timobrembeck (approved):

PR #2434, opened 2023-09-26
Improve handling of brandings

timobrembeck:
### Short description

This adds a few improvements to our whitelabel/branding functionality.

### Proposed changes

- ~Add possibility to set region-specific brandings~ (rejected, aschaffenburg is getting its own cms)
- Make radable branding title configurable
- Add more regions to the test data

### DIFF
--- a/integreat_cms/cms/fixtures/test_data.json
+++ b/integreat_cms/cms/fixtures/test_data.json
@@ -234,10 +234,7 @@
         "Steindorf": { "latitude": 48.215553, "longitude": 11.001126 },
         "Schmiechen": { "latitude": 48.212774, "longitude": 10.965841 }
       },
-      "events_enabled": true,
       "locations_enabled": true,
-      "chat_enabled": true,
-      "push_notifications_enabled": true,
       "short_urls_enabled": true,
       "page_permissions_enabled": true,
       "external_news_enabled": true,
@@ -251,9 +248,6 @@
       "admin_mail": "augsburg@integreat-app.de",
       "created_date": "2019-08-12T07:48:17.997Z",
       "last_updated": "2019-08-12T07:48:17.998Z",
-      "statistics_enabled": false,
-      "matomo_id": null,
-      "matomo_token": "",
       "offers": [1, 2, 3]
     }
   },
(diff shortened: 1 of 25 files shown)


timobrembeck:
> Add possibility to set region-specific brandings

I'm not sure however if this is a step in the right direction, if we're going to move hallo aschaffenburg to its own VM, then I'd remove this feature since there is no use case anymore.

The improvements to the readable title however still make sense in my opinion, especially with regard to #2391.

svenseeberg:
> if we're going to move hallo aschaffenburg to its own VM, then I'd remove this feature since there is no use case anymore.

I think that moving them into their own server would work best. The maintenance effort for an additional server is quite low.

PeterNerlich (approved):
I've looked at the code changes and tried running the CMS with each possible `INTEGREAT_CMS_BRANDING` env value to observe the branding, and I have no complaints.

I'm only wondering whether it would be useful / reasonably easy to put the primary color in the `integreat_cms/core/settings.py` as well (currently it's in `integreat_cms/static/src/css/style.scss:76`+) – then it would all be in one place.

timobrembeck:
> I'm only wondering whether it would be useful / reasonably easy to put the primary color in the `integreat_cms/core/settings.py` as well (currently it's in `integreat_cms/static/src/css/style.scss:76`+) – then it would all be in one place.

Hmm, I don't see an easy way to pass the settings from Python to CSS – also, the logos etc. reside in the static directory (which is where they belong), so I think keeping style information in the stylesheet makes sense here.

timobrembeck:
> > if we're going to move hallo aschaffenburg to its own VM, then I'd remove this feature since there is no use case anymore.
> 
> I think that moving them into their own server would work best. The maintenance effort for an additional server is quite low. The bigger issue would be the migration itself. But we can configure a reverse proxy for all hello paths.

@svenseeberg Sounds good! Maybe a few redirects are sufficient until the app adjusts the new path and most devices are updated?

timobrembeck:
Ok, I removed all code related to region-specific brandings, since there is no use case anymore if aschaffenburg is getting its own cms.

Issue #2128, opened 2023-03-17
Unsaved warning is shown despite content is autosaved on leave


timobrembeck:
### Describe the Bug

When changes are made to the tinymce content forms, the autosave is triggered on leaving the form, no matter whether the "unsaved warning" is ignored or not.

### Steps to Reproduce

1. Go to content form
2. Make changes
3. Leave the form without saving before the autosave interval (120s)
4. Ignore the browser warning and click on "leave page"
5. Return to editor and see that the changes have been autosaved

### Expected Behavior

Since the changes are saved, I don't expect the unsaved warning

### Actual Behavior

The browser's unsaved warning is shown despite the saved changes:
![unsavedWarning](https://user-images.githubusercontent.com/16021269/225890201-f6f15c21-14d0-4e86-96de-f295a17f66b3.png)

### Additional Information

Only the changes of the content translation forms are saved, the content models itself aren't.
So I would suggest to keep the warning when something in the sidebars is changed (which would indeed be lost).

An alternative solution would be to disable the preemptive saving when the form is left with ignoring the warning.

timobrembeck:
> As it turned out to be unexpectedly complex to achieve a consistent behaviour in different brouwers and the problem is not critical, I'll close this PR for now.

_Originally posted by @MizukiTemma in https://github.com/digitalfabrik/integreat-cms/issues/2132#issuecomment-1766595978_

PR #2435, opened 2023-09-26
Show only unsaved warning when changes are not saved

MizukiTemma:
### Short description

This is the second try after the approach in #2132 stucked due to the behaviour of FireFox.

### Proposed changes

- Benefit from #2243 😁 

### Side effects

- I hope there is none.

### Resolved issues

Fixes: #2128 

### DIFF
--- a/integreat_cms/cms/templates/events/event_form.html
+++ b/integreat_cms/cms/templates/events/event_form.html
@@ -130,7 +130,7 @@
                         <label for="{{ event_translation_form.content.id_for_label }}">
                             {{ event_translation_form.content.label }}
                         </label>
-                        {% render_field event_translation_form.content|add_error_class:"border-red-500" class+="tinymce_textarea flex-auto" %}
+                        {% render_field event_translation_form.content|add_error_class:"border-red-500" class+="tinymce_textarea flex-auto" data-unsaved-warning-exclude="" %}
                     </div>
                 </div>
             </div>
--- a/integreat_cms/cms/templates/imprint/imprint_form.html
+++ b/integreat_cms/cms/templates/imprint/imprint_form.html
@@ -105,7 +105,7 @@
                         <label for="{{ imprint_translation_form.content.id_for_label }}">
                             {{ imprint_translation_form.content.label }}
                         </label>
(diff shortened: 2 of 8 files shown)


david-venhoff (changes requested):
Unfortunately this still breaks the autosave functionality on firefox for me 🙈. 
However, there seems to be a method that might fix our problem: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon
Have we every tried that in the previous pr?

Apart from that, I think the `data-unsaved-warning-exclude` attribute should also be added to the `imprint_sbs` and the `page_sbs`.
Also, since the input events from tinymce forms are not required anymore, we can probably remove this in `tinymce-init.ts`:
```ts
                // When the editor becomes dirty, send an input event, so that the unsaved warning can be shown
                editor.on("dirty", () => {
                    if ((editor.targetElm as HTMLElement).hasAttribute("data-unsaved-warning-exclude")) {
                        return;
                    }
                    document.querySelectorAll("[data-unsaved-warning]").forEach((element) => {
                        element.dispatchEvent(new Event("input"));
                    });
                });
```

MizukiTemma:
@david-venhoff
ohh stupid me 🙈 I hadn't checked it with FireFox before opening the PR 🙈 

With `navigator.sendBeacon` auto save is working, but there is one problem that it return only `true` or `false` and we cannot set `form.action` to the url returned by the server. Without `form.action = data.url` multiple of new pages will be generated if an user is editing a brand new page, and I don't come up with any alternative way to catch the server response than staying with `fetch` 😿

david-venhoff:
Ahh okay. In that case we could maybe have two autosave implementations, one for normal timed autosaves and one for when the user is attempting to leave the page?
As far as I understand, we only have to update the form action when the user is not leaving the page, so in the problematic case we could use the beacon api.

MizukiTemma:
I tried Firefox-oriented some improvement, but it still needs adjustment for cases when a page already exists and the first translation of a non-root language is being edited, and it seems also unstable now in other browsers than Firefox.

Honestly saying, I'm thinking whether it's worth this much of effort 😢 (or just I'm not good enough at this thema)

MizukiTemma:
Same with #2132

Issue #2389, opened 2023-08-07
Do not show translation status of the root language


MizukiTemma:
### Motivation

Translation status of the root language is always `up to date`. Therefore it is not so meaningfull to show it in the content list, so let's omit it.

### Proposed Solution

Show only status of other languages (non-root).
This can be applied not only for pages but also for POI and event.

### Alternatives

Leave it as it is.

### Additional Context

![image](https://github.com/digitalfabrik/integreat-cms/assets/82345046/e0761213-b6b5-4453-a06b-ccb94199d848)

dkehne:
Does not really makes an effort, can be done if we are bored.

timobrembeck:
I updated my flag package with the new flag, so this issue is no longer blocked :innocent:

PR #2436, opened 2023-09-28
Do not show translation status of the root language

MizukiTemma:
### Short description

This PR removes the column of the root language in the translation status table in the event/POI list and page tree.

### Proposed changes

- Skip icon display if the language if the root of the region

### Side effects

- None?

### Resolved issues

Fixes: #2389 

### DIFF
--- a/integreat_cms/cms/templates/events/event_list.html
+++ b/integreat_cms/cms/templates/events/event_list.html
@@ -74,10 +74,12 @@
                         <div class="lang-grid flags whitespace-nowrap">
                             {% spaceless %}
                                 {% for lang in languages %}
-                                    <a href="{% url 'events' region_slug=request.region.slug language_slug=lang.slug %}">
-                                        <span class="fp fp-rounded fp-{{ lang.primary_country_code }}"
-                                              title="{{ lang.translated_name }}"></span>
-                                    </a>
+                                    {% if lang != request.region.default_language %}
+                                        <a href="{% url 'events' region_slug=request.region.slug language_slug=lang.slug %}">
+                                            <span class="fp fp-rounded fp-{{ lang.primary_country_code }}"
+                                                  title="{{ lang.translated_name }}"></span>
+                                        </a>
+                                    {% endif %}
                                 {% endfor %}
                             {% endspaceless %}
                         </div>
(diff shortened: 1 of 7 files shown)


timobrembeck:
One disadvantage of this change: When a region has German as default language and both German & Simple German enabled, only one German flag is shown in the translation table which could lead to confusion in the form of "Hmm, why is the German translation missing here?"

simple-german-fehlt

Hovering over the German flag would reveal that it's simple German, but nonetheless I think this is not 100% ideal...

MizukiTemma:
> One disadvantage of this change: When a region has German as default language and both German & Simple German enabled, only one German flag is shown in the translation table which could lead to confusion in the form of "Hmm, why is the German translation missing here?"
> 
> simple-german-fehlt
> 
> Hovering over the German flag would reveal that it's simple German, but nonetheless I think this is not 100% ideal...

@osmers What do you think?

PeterNerlich:
Sorry for just injecting myself into the conversation like this. I'm convinced we need to adapt our iconography (and make it a good one) to solve this problem. Unfortunately, most flags are geometrically already the simplest form they could be, so we cannot whip up a simplified version that intuitively shows that the language has been simplified. After all, the German flag is just three flat-colored stripes – there's no way to simplify that any further without losing fundamental features of the flag.

The best I can imagine on the fly is a generic added symbol. I tried it with an overlayed **E**, but if you have an idea that uses a non-verbal icon instead, that would be preferred as long as it is still readable on the flag at such a small size.

image

PeterNerlich:
Yeah, that signet sadly does not give me the impression that a graphic designer made it or at least had the last say in it… :sweat_smile:

MizukiTemma:
[Here](https://www.figma.com/file/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?type=design&mode=design) is the update from @hauf-toni

MizukiTemma:
This PR will be continued after the new flag icon for Easy German is added to [this library](https://github.com/timobrembeck/flagpack-dart-sass/pull/10).

MizukiTemma:
Closed as not urgent. The branch will remain there.

timobrembeck:
I released https://github.com/timobrembeck/flagpack-dart-sass/releases/tag/v1.5.0 now with the new flag, so this PR is no longer blocked :innocent: 
But apparently, GitHub does not allow re-opening PRs when their branches got force-pushed to while the PR was closed? :thinking:

Issue #2356, opened 2023-07-23
Broken images cause PDF export to fail


timobrembeck:
### Describe the Bug

When an image is broken, it causes the complete PDF export to fail

### Steps to Reproduce

https://integreat.app/testumgebung/de/test-broken-photo-pdf
https://cms.integreat-app.de/testumgebung/de/wp-json/ig-mpdf/v1/pdf?url=%2Ftestumgebung%2Fde%2Ftest-broken-photo-pdf

### Expected Behavior

We should ignore the image or show a placeholder, but generate the PDF

### Actual Behavior

An internal server error occurs

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
Jul 23 22:12:01 WARNING integreat_cms.cms.utils.pdf_utils - The file 'https://integreat.app/dortmund
/wp-content/uploads/sites/187/2017/11/calendar60-1.png' is not inside the static directories '/stati
c/' and '/media/'.
Jul 23 22:12:03 WARNING xhtml2pdf - Error in handling image
'<img class="alignnone wp-image-589" src="https://integreat.app/dortmund/wp-content/uploads/sites/187/2017/11/calendar60-1.png" alt="" width="15" height="15" dir="rtl"/>'
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/xhtml2pdf/xhtml2pdf_reportlab.py", line 323, in __init__
    self._image = self._read_image(self.fp)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/xhtml2pdf/xhtml2pdf_reportlab.py", line 353, in _read_image
    return PILImage.open(fp)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/PIL/Image.py", line 3298, in open
    raise UnidentifiedImageError(msg)
PIL.UnidentifiedImageError: cannot identify image file <_io.BytesIO object at 0x7fcde76a5360> fileName=<_io.BytesIO object at 0x7fcde76a5360>

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/xhtml2pdf/tags.py", line 354, in start
    img = PmlImage(
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/xhtml2pdf/xhtml2pdf_reportlab.py", line 467, in __init__
    img = self.getImage()
[...]

PR #2438, opened 2023-09-29
Fix PDF export fail with broken images

MizukiTemma:
### Short description

This PR fixes the problem that broken images cause an error during PDF export.

### Proposed changes

- If an image is broken, replace it with a placeholder image.
- In my opinion, showing a placeholder is better than just ignoring broken images because content managers and users can notice that there actually must be an image but something is wrong with it.
- Currently just a gray image is being used as placeholder, for we don't have any appropriate image for it in the repository and I am afraid of violating copyright or any constraints by using one found on the internet. Does anyone have a good candidate for the placeholder?

### Side effects

- I have reproduced the bug in the following steps and this is fixed by the PR. I hope exact the bug reported in the issue was successfully reproduced and fixed.
1. Put a broken img tag (for example <p>Logo Caritas Dortmund</p>)into a page.
2. Export the page as PDF
3. See the error message  `......is not inside the static directories '/static/' and '/media/'` in the console and failed PDF Export.

### Resolved issues

Fixes: #2356 


timobrembeck (changes requested):
@MizukiTemma Thanks a lot for giving this issue a shot, however there were a few misunderstandings:

1. The bug is not about images that are not in the static directory. In general, it should be allowed to load images from remote sources. E.g. insert the valid image
    ```
    <img class="alignnone" src="https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcT7VoXh8W7JsQw5pKkGlx-_V4ZykwIcYVQ7bqkmc4ETlUqyHfMUPA&amp;s" alt="Logo Amnesty International" width="173" height="63">
    ```
    into a page, on the develop branch this correctly renders the Amnesty International logo, but on your branch this changes to a placeholder, since the image is not located on the cms server.
2. The bug is about very specific images from remote sources which cannot be parsed as images. One example is if you insert the source code
    ```
    <img class="alignnone" src="https://foresight4youth.com/wp-content/uploads/2018/07/Logo-Caritas-Dortmund-300x124.jpg" alt="Logo Caritas Dortmund" width="174" height="72">
    ```
    Although the image looks ok in the browser when accessing https://foresight4youth.com/wp-content/uploads/2018/07/Logo-Caritas-Dortmund-300x124.jpg, our image library cannot render this image.
    In this case, a placeholder makes sense instead of returning an error 500 (or an empty PDF if it's loaded from the cache)...
    The exception I get with this example is:
    ```
    ZeroDivisionError at /augsburg/pages/de/export/

    float division by zero
    ```
    But the bug fix should not be limited to this specific error, but in general when an image file is broken.

PeterNerlich:
>    Although the image looks ok in the browser when accessing https://foresight4youth.com/wp-content/uploads/2018/07/Logo-Caritas-Dortmund-300x124.jpg, our image library cannot render this image.

Actually, for me it does not look ok. I get this friendly cookie message instead:
image

So while this is probably the explanation for only this specific case, it might help guess when this breaks in general: Whenever a path is returning something else than an image. Here the mime type is already changed to `text/html`, but there might also be just broken responses where the mime type is promising, but the data itself is garbage.

timobrembeck:
> Actually, for me it does not look ok. I get this friendly cookie message instead:

Ahh, good catch, yes this is obviously a mistake on their side of the wordpress setup, but I guess we should somehow deal with such mistakes.

> So while this is probably the explanation for only this specific case, it might help guess when this breaks in general: Whenever a path is returning something else than an image. Here the mime type is already changed to `text/html`, but there might also be just broken responses where the mime type is promising, but the data itself is garbage.

Exactly, there might be even cases where some programs that have less strict file format checks just display the image correctly, but the python library returns an error. So mime-checking is not sufficient, I would just recommend a broad `try`/`except` here to catch all potential exceptions.

timobrembeck:
I decided to fix this in the upstream library, so no need to address my requested changes.
See also: https://github.com/digitalfabrik/integreat-cms/issues/2356#issuecomment-1751815004

MizukiTemma:
@timobrembeck OK.

PR #2439, opened 2023-10-02
Fix different HIX value calculation

seluianova:
### Short description
There are still cases where we got different HIX values for the same content.
For example:
1. Create new page with the following content:
```
<p>Eine sehr schwierige Zeile</p>
<p>Eine weitere sehr schwierige Zeile</p>
```
2. Save or Publish it --> HIX is **19.17**
3. Press "Publish" again with no changes --> HIX is **19.86**

The behavior will be the same if step 3 is "Press Ctrl+A and reload the HIX value".

### Proposed changes
-  Rework text normalization before sending request to Textlab API

### Side effects
- Hope none, but careful testing is welcomed

### Resolved issues
Fixes: #2224

### DIFF
--- a/integreat_cms/core/signals/hix_signals.py
+++ b/integreat_cms/core/signals/hix_signals.py
@@ -35,16 +35,19 @@ def page_translation_save_handler(instance, **kwargs):
         instance.hix_score = None
         return
 
-    if identical_version := instance.all_versions.filter(
-        content=instance.content, hix_score__isnull=False
-    ).first():
+    latest_version = instance.latest_version
+
+    if (
+        latest_version
+        and latest_version.hix_score
+        and latest_version.content == instance.content
+    ):
         logger.debug(
-            "Content of %r is identical to %r, copying HIX score %r",
+            "Content of %r was not changed, copying the HIX score from the previous version: %r",
             instance,
-            identical_version,
-            identical_version.hix_score,
+            latest_version.hix_score,
         )
-        instance.hix_score = identical_version.hix_score
+        instance.hix_score = latest_version.hix_score
         return
 
     if score := lookup_hix_score(instance.content):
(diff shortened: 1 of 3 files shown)


integreat_cms/core/signals/hix_signals.py
```diff
@@ -35,16 +35,15 @@ def page_translation_save_handler(instance, **kwargs):
         instance.hix_score = None
         return
 
-    if identical_version := instance.all_versions.filter(
-        content=instance.content, hix_score__isnull=False
-    ).first():
+    latest_version = instance.latest_version
+
+    if latest_version.content == instance.content:
```
seluianova, line 40: Otherwise we won't be able to obtain the stable HIX value without changing the page content.
Because if some equal content with an old (broken) value is found, it will be constantly re-stored instead of sending a new request.
Alternatively, we can make a management command to fix the "broken" values?

integreat_cms/core/signals/hix_signals.py
```diff
@@ -35,16 +35,15 @@ def page_translation_save_handler(instance, **kwargs):
         instance.hix_score = None
         return
 
-    if identical_version := instance.all_versions.filter(
-        content=instance.content, hix_score__isnull=False
-    ).first():
+    latest_version = instance.latest_version
+
+    if latest_version and latest_version.content == instance.content:
```
seluianova, line 40: Otherwise we won't be able to obtain the stable HIX value without changing the page content.
Because if previous equal content with an old (broken) value is found, it will be constantly re-stored instead of sending a new request.
Alternatively, we can make a management command to fix the "broken" values?

MizukiTemma (approved):
LGTM 👍 Thank you

integreat_cms/core/signals/hix_signals.py
MizukiTemma: I think this way is fine 🙂

david-venhoff (approved):
Thanks, LGTM

PR #2441, opened 2023-10-09
Change print to logging in push notification sender

timobrembeck:
### Short description

Since the push notification sending command is used as cronjob and not as manual command (and because the [output of the crobjob is redirected to the logfile](https://git.tuerantuer.org/DF/salt/src/commit/c0ebd23bcffb5c8ce61fb4e2e3e2d88d65dcffeb/states/integreat-cms/files/cms-worker.service#L12)), the previous solution caused a few messages in the log that were very hard to identify, e.g.:
```
Done!
```

Instead of print statements (which are great for manual commands because they're better readable than log entries), I suggest to convert the output to regular logging calls.

### Proposed changes

- Change print to logging in push notification sender

### DIFF
--- a/integreat_cms/core/management/commands/send_push_notifications.py
+++ b/integreat_cms/core/management/commands/send_push_notifications.py
@@ -1,14 +1,17 @@
+import logging
+
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
-from django.core.management.base import CommandError
+from django.core.management.base import BaseCommand, CommandError
 from django.utils import timezone
 
 from ....cms.models import PushNotification, Region
 from ....firebase_api.firebase_api_client import FirebaseApiClient
-from ..log_command import LogCommand
+
+logger = logging.getLogger(__name__)
 
 
-class Command(LogCommand):
+class Command(BaseCommand):
     """
     Management command to send timed push notifications
     """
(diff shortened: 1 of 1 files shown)


MizukiTemma (approved):
Looks good 🙂

integreat_cms/core/management/commands/send_push_notifications.py
```diff
@@ -45,9 +49,14 @@ def handle(self, *args, **options):
             draft=False,
             scheduled_send_date__lte=timezone.now(),
         )
-        for pn in pending_push_notifications:
-            self.send_push_notification(pn)
-        self.print_success("Done!")
+        if pending_push_notifications:
+            for pn in pending_push_notifications:
+                self.send_push_notification(pn)
+            logger.info("All scheduled push notifications have been processed.")
```
ulliholtgrave, line 55: Could we add a count or some representation of the sent PNs? I guess this might come in handy if the feature isn't working or used as intended :)

ulliholtgrave (changes requested):

integreat_cms/core/management/commands/send_push_notifications.py
timobrembeck: @ulliholtgrave good idea, maybe like this?

ulliholtgrave (approved):
Works for me :+1:

Issue #2294, opened 2023-05-17
Add location url to event API response


timobrembeck:
### Motivation

In theory, event locations can already be mapped to their location object via the id, but when this information is not yet retrieved by the app, it considerably slows down the loading of an event.

### Proposed Solution

Add a new property to the `location` object of the event api response (either `slug` or `url`).
I guess the url would be not completely trivial since we don't know whether the location is available in the current language, so we have to put some thought into how to handle this case. Link to the default language?

### Alternatives

Do the mapping via the `id`

Related issue:
- https://github.com/digitalfabrik/integreat-cms/issues/2277

sarahsporck:
I think as fallback either linking to the default language or not linking at all would make the most sense. Small remark, linking to the default language could be bad UX, though.

timobrembeck:
True, although at the moment all of our 150 regions have the "fallback translation" feature enabled for events and pois, which means that all objects are visible in all languages, even if the content is not available in some languages:

(this can be set in the region form)

![Screenshot 2023-10-09 at 15-32-45 Integreat Redaktionssystem](https://github.com/digitalfabrik/integreat-cms/assets/16021269/6f53ce96-3f26-4ff5-9155-28ec8929a517)

But from the app's perspective, there is no difference, the translations look like normal translations even if they have content in the default language. So I guess we can just use our internal translation method [`POI.get_public_translation()`](https://github.com/digitalfabrik/integreat-cms/blob/e65275995c2a1775b54e70afe49d912d530554d8/integreat_cms/cms/models/abstract_content_model.py#L254) which automatically handles this case correctly and returns a translation if it exists (or fallback translations are enabled), and no translation otherwise.

PR #2447, opened 2023-10-10
Add `location_url` to event endpoint response

sarahsporck:
### Short description
This adds the `location_url` to the event endpoint. I decided against adding the url in the `transform_poi` function, as it is also used in the poi endpoint and there this information would be duplicated.

### Proposed changes

- add attribute `location_url` to event endpoint

### Resolved issues

Fixes: #2294 

### DIFF
--- a/integreat_cms/api/v3/events.py
+++ b/integreat_cms/api/v3/events.py
@@ -52,12 +52,17 @@ def transform_event(event, custom_date=None):
     }
 
 
-def transform_event_translation(event_translation, recurrence_date=None):
+def transform_event_translation(
+    event_translation, poi_translation, recurrence_date=None
+):
     """
     Function to create a JSON from a single event_translation object.
 
     :param event_translation: The event translation object which should be converted
     :type event_translation: ~integreat_cms.cms.models.events.event_translation.EventTranslation
+
+    :param poi_translation: The poi translation object which is associated to this event
+    :type poi_translation: ~integreat_cms.cms.models.pois.poi_translation.POITranslation
 
     :param recurrence_date: The recurrence date for the event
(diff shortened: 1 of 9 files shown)


integreat_cms/api/v3/events.py
```diff
@@ -88,6 +93,9 @@ def transform_event_translation(event_translation, recurrence_date=None):
         else event_translation.available_languages_dict,
         "thumbnail": event.icon.url if event.icon else None,
         "location": transform_poi(event.location),
+        "location_url": settings.BASE_URL + poi_translation.get_absolute_url()
+        if event.location
+        else None,
```
david-venhoff, line 98:
```suggestion
        "location_url": settings.BASE_URL + poi_translation.get_absolute_url()
        if poi_translation
        else None,
```
This would crash for an event translation if its location has not been translated into the language of the event translation and if fallback translations are disabled.

david-venhoff (approved):
Thanks, this looks mostly good to me. This update just needs to be reflected in our api documentation.

sarahsporck:
> Thanks, this looks mostly good to me. This update just needs to be reflected in our api documentation.

How do you update the api docs? I only found `./tools/make_docs.sh`

timobrembeck:
> How do you update the api docs? I only found `./tools/make_docs.sh`

Unfortunately the api docs cannot be automatically generated from the source code, so it has to be manually changed in this document:

https://github.com/digitalfabrik/integreat-cms/blob/01c9631b145b0ca62dd2d4be37ef6e047df351f5/docs/src/api-docs.rst

MizukiTemma (approved):
LGTM 👍

Can we maybe add a release note?

After squashing it's ready to merge 🙂

Issue #2452, opened 2023-10-11
Deletion of region causes Error 500


dkehne:
### Describe the Bug
Try to delete the Testumgebung-region: Error 500. Region not deleted.

### Steps to Reproduce
See bug description

![Screenshot 2023-10-11 110823](https://github.com/digitalfabrik/integreat-cms/assets/4182469/0dd9ddcd-330f-434f-a638-5954172770a9)

timobrembeck:
The internal traceback is:
```
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/decorators/http.py", line 40, in inner
    return func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/views/regions/region_actions.py", line 93, in delete_region
    deleted_objects = region.delete()
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/db/models/base.py", line 966, in delete
    collector.collect([self], keep_parents=keep_parents)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/db/models/deletion.py", line 302, in collect
    raise ProtectedError(
django.db.models.deletion.ProtectedError: ("Cannot delete some instances of model 'Region' because they are referenced through protected foreign keys: 'MediaFile.region'.", {<Organization (id: 1, slug: tür-an-tür, region: testumgebung)>})
```

timobrembeck:
In other words, the reason is a circular protection problem:
The media file `TaT_DF_LOGO.png` cannot be deleted, because it is still used as logo of the tür-an-tür organization
So a quick workaround would be to delete the [tür-an-tür](https://admin.integreat-app.de/testumgebung/organizations/t%C3%BCr-an-t%C3%BCr/edit/) organization first, and then delete the whole region.

PR #2455, opened 2023-10-11
Fix error when deleting regions with organization logos

timobrembeck:
### Short description

Fix error when deleting regions with organization logos

### Proposed changes

- Delete region's organizations before the region itself (including the region's media library)

### Side effects

- If an error occurs during `region.delete()`, the organization objects can not be recovered

### Resolved issues

Fixes: #2452


MizukiTemma (approved):
Thank you 😸 Works as expected

Issue #319, opened 2020-03-06
Add Python type hints


Thomas125:
Enforce type-annotation, see:
- https://docs.python.org/3/library/typing.html
- https://typing.readthedocs.io/en/latest/

timobrembeck:
To check the type hints, we can use mypy: https://mypy-lang.org/

Does anybody know a good tool to convert our current type hints from the docstrings into the real type hint syntax?
Essentially, this was what https://github.com/tox-dev/sphinx-autodoc-typehints is doing, but sphinx only uses this to write the documentation, not to update the source code itself. Maybe we could still use this somehow?

Another possibility is to create stub files with [stubgen](https://mypy.readthedocs.io/en/stable/stubgen.html) and read them into the source code with [retype](https://pypi.org/project/retype/), but this still requires a lot of manual work to fix all inconsistencies and incomplete types.

I started working on this on the [`typing`](https://github.com/digitalfabrik/integreat-cms/tree/typing) branch, but there is still much work to do.

PR #2459, opened 2023-10-11
Add type annotations

timobrembeck:
### Short description

From time to time, [we get TypeErrors](https://github.com/digitalfabrik/integreat-cms/issues?q=is%3Aissue+TypeError) in the production code, which we could prevent if we'd use static type checking from the beginning.

At the same time, I think it has the benefit that we as developers have to think more about the possible values a variable can have, which makes us become more aware for edge cases and the code we're writing in general.

### Proposed changes

- Update dependencies
- Use type annotations
- Check them with [my[py]](https://mypy.readthedocs.io/en/stable/index.html)

### Side effects

- The code gets more verbose
- Sometimes it's not completely obvious which type the return value of a third party library has etc, so it makes development a bit slower
- **I had to disable Sphinx' nitpicky mode due to https://github.com/sphinx-doc/sphinx/issues/9813 :sob:** 

### Resolved issues

Fixes: #319


timobrembeck (commented):
I know, the diff of this PR got a little bit out of hand, so I don't expect anyone to actually read all this :sweat_smile: 
Just a few random checks should be enough, since all the main functionality is also covered by the tests.

There will be a few mistakes here and there, but I'd suggest fixing them later instead of keeping this PR on hold until somebody had time to look at all the changes... :see_no_evil: 

Biggest drawback is definitely https://github.com/sphinx-doc/sphinx/issues/9813, but I think the benefits of the type annotations are worth it.

seluianova (approved):
I've done a few random checks and haven't faced any issues.

JoeyStk (approved):
Same goes for me. I also didn't run into any issues!

Issue #2449, opened 2023-10-10
`TypeError`: `>` not supported between instances of `NoneType` and `NoneType`


timobrembeck:
### Describe the Bug

#2419 introduced a new bug despite the tests were successful on the PR (yet another reason why we need #2060)

### Steps to Reproduce
https://app.circleci.com/pipelines/github/digitalfabrik/integreat-cms/9650/workflows/063d1449-2c90-4a7c-b98f-29558433e126/jobs/107953

### Expected Behavior

No error

### Actual Behavior

```
TypeError: 'Region' object is not subscriptable

During handling of the above exception, another exception occurred:

TypeError: '>' not supported between instances of 'NoneType' and 'NoneType'
```

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
Traceback (most recent call last):
  File "/home/circleci/project/.venv/lib/python3.9/site-packages/django/template/base.py", line 829, in _resolve_lookup
    current = current[bit]
TypeError: 'Region' object is not subscriptable

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/circleci/project/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/circleci/project/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/circleci/project/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/circleci/project/.venv/lib/python3.9/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/home/circleci/project/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/home/circleci/project/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/home/circleci/project/integreat_cms/cms/views/regions/region_list_view.py", line 58, in get
    return render(
  File "/home/circleci/project/.venv/lib/python3.9/site-packages/django/shortcuts.py", line 19, in render
    content = loader.render_to_string(template_name, context, request, using=using)
[...]

PR #2461, opened 2023-10-11
Fix internal server error in region list

timobrembeck:
### Short description

~Hot~fix for the bug in the region list

### Resolved issues

Fixes: #2449


timobrembeck:
> Ok, I found the culprit: There were a few regions in the production database which had an imprint set without translations. This is why the if condition did not work as intended.
I deleted the inconsistent imprints now, so this is no longer urgent.
Still makes sense to prevent this error in the future, but in general I think we can assume that our database is consistent.

– https://github.com/digitalfabrik/integreat-cms/issues/2449#issuecomment-1758565022

ulliholtgrave (approved):

charludo (approved):

Issue #2283, opened 2023-05-17
`UnreadablePostError` on content edit lock release


timobrembeck:
### Describe the Bug

```
May 17 09:39:22 ERROR   django.request - 500 Internal Server Error: /saarbruecken/ajax/content-edit-lock/release/
[...]
django.http.request.UnreadablePostError: Apache/mod_wsgi request data read error: Partial results are valid but processing is incomplete.
```
We get a lot of these error messages on the prod server, and while the error itself it probably not very crucial, it makes the log file barely readable, because 70% of the entries are related to this error.
This is not a new phenomenon, it has been around for a few months now...

### Steps to Reproduce

1. Deploy system to apache server
2. Leave any edit page
3. See error

### Expected Behavior

The `content_edit_lock_release` should not throw an error if the user forcefully ends the request

### Actual Behavior

An error occurs (not exactly sure in which cases)

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
May 17 09:39:22 ERROR   django.request - 500 Internal Server Error: /saarbruecken/ajax/content-edit-lock/release/
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/http/request.py", line 393, in read
    return self._stream.read(*args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/wsgi.py", line 34, in read
    result = self.buffer + self._read_limited()
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/wsgi.py", line 28, in _read_limited
    result = self.stream.read(size)
OSError: Apache/mod_wsgi request data read error: Partial results are valid but processing is incomplete.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/http/request.py", line 338, in body
    self._body = self.read()
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/http/request.py", line 395, in read
    raise UnreadablePostError(*e.args) from e
django.http.request.UnreadablePostError: Apache/mod_wsgi request data read error: Partial results are valid but processing is incomplete.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
[...]

PR #2467, opened 2023-10-16
Keep content edit lock release request alive

david-venhoff:
### Short description

This pr tries to fix the log spam caused by many content-edit-lock-release requests being aborted.
An error will still be logged if a request gets aborted, but I hope that this should only happen very rarely after this pr.
Unfortunately I cannot reproduce the actual issue locally, does someone know how to do that?

### Proposed changes

- Use the `navigator.sendBeacon` api which does not cancel requests
- use the `pagehide` event instead of `unload` (https://developer.mozilla.org/en-US/docs/Web/API/Window/unload_event#usage_notes)

### Side effects

- Ideally none

### Resolved issues

Fixes: #2283

### DIFF
--- a/integreat_cms/cms/views/utils/content_edit_lock.py
+++ b/integreat_cms/cms/views/utils/content_edit_lock.py
@@ -61,7 +61,7 @@ def content_edit_lock_release(request, region_slug=None):
     :return: Json object containing `success: true` if the content object could be unlocked
     :rtype: ~django.http.JsonResponse
     """
-    body = json.loads(request.body.decode("utf-8"))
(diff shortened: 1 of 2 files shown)


timobrembeck (approved):
Thanks a lot for investigating this problem!
I also didn't try it out to recreate a production environment locally, but this looks like it might actually work! :+1:

MizukiTemma (approved):
LGTM 👍

timobrembeck:
@david-venhoff worked like a charm, not a single `UnreadablePostError` in the logs today :pray:

Issue #2360, opened 2023-07-25
Page based offer


svenseeberg:
### Motivation
In the future we want to move offers into our page based content to avoid parallel structures.

### Proposed Solution
Define API endpoints that allow associating offers with specific pages in which the content should be displayed.

### Alternatives
N/A

### User Story
As a CMS editor I want embed an offer into a page so that app users find content at the expected locations.

### Additional Context
This is required for #

PR #2474, opened 2023-10-20
Page based offers

charludo:
### Short description

For offer providers for which support in the app has been added, add the option to embed offers in a page.

### Proposed changes

- add required fields to the `OfferTemplate` and `Page` models and forms
- amend API response
- amend fixture to utilize the feature and corresponding test cases

### Side effects

- not sure what is going on with the reformatting in the `test_data.json`, this is `prettier`s output 😦  

### Resolved issues

Fixes: #2360 


integreat_cms/api/v3/pages.py
timobrembeck, line 1: Could you also fetch the offer templates with `select_related` in order to avoid one additional query per page that has set an offer?

integreat_cms/cms/forms/pages/page_form.py
```diff
@@ -102,6 +103,11 @@ def __init__(self, **kwargs):
         # Let mirrored page queryset be empty per default and only fill it if a region is selected
         mirrored_page_queryset = Page.objects.none()
 
+        # Filter the offer provides available for embedding
+        self.fields["page_based_offer"].queryset = self.instance.region.offers.filter(
+            supported_by_app_in_content=True
+        )
```
timobrembeck, line 109: Since we only allow offers here that are available in the current region: Do we also need a post-save signal handler for the region model to remove an offer from all pages of that region when it's disabled in the region form?

timobrembeck (changes requested):
Awesome, thanks! :tada:
Works as expected, just a few minor nitpicks from my side :grin:

integreat_cms/cms/models/pages/page.py
```diff
@@ -197,6 +197,16 @@ class Page(AbstractTreeNode, AbstractBasePage):
         verbose_name=_("Ignore HIX value"),
         help_text=_("This allows to ignore HIX value for this specific page."),
     )
+    page_based_offer = models.ForeignKey(
```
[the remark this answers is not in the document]
charludo: Fair point. Just `offer` isn't really indicative of the intention though, maybe we could go with `embedded_offer`?

integreat_cms/cms/forms/pages/page_form.py
charludo: That's a good question I had not thought about yet. Actually I am not sure removing the offers is ideal - for example, what in case of accidental disables of the offer? That could potentially caue some frustration. I would throw
- just keep them embedded (plus maybe show a warning in the `OfferTemplate` form) and
- disable the disabling unless no page embeds those offers anymore
into the ring of potential solutions.

timobrembeck: Hmm yeah I agree this isn't ideal. But if the offer is really disabled, it could also be convenient to do this via the region form without having to click though all pages and remove it manually everywhere?
Hmm, not sure about this. I guess as quick fix it's safest to disable the disabling (:grin:) until we come up with a long term solution.

charludo: Alright, sounds good to me!

integreat_cms/cms/models/pages/page.py
```diff
@@ -197,6 +197,16 @@ class Page(AbstractTreeNode, AbstractBasePage):
         verbose_name=_("Ignore HIX value"),
         help_text=_("This allows to ignore HIX value for this specific page."),
     )
+    embedded_offer = models.ForeignKey(
+        "cms.OfferTemplate",
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
```
timobrembeck, line 204:
```suggestion
        "cms.Offer",
        null=True,
        blank=True,
        on_delete=models.PROTECT,
```
I realize it's quite late for this debate, but one idea I just had: We could also convert this to a foreign key to the `Offer` model, then we'd have the protection mechanism in the region form by default...

timobrembeck (changes requested):
Thanks! :+1:

And apparently, we need to be able to set multiple offers per page, see https://github.com/digitalfabrik/integreat-cms/issues/2360#issuecomment-1775141807

integreat_cms/cms/models/pages/page.py
timobrembeck: Haha oh I'm so sorry, somehow I missed that we refactored our offer models [two years ago](https://github.com/digitalfabrik/integreat-cms/pull/790) :joy:
We used to have an intermediary [`Offer` model](https://github.com/digitalfabrik/integreat-cms/blob/5425956ffa8a1c0c9c36953e86c57f7593c86352/src/cms/models/offers/offer.py#L10-L31) that acted as link between regions and offer templates^^
Ok well then my idea doesn't work of course, since protection only protects against deletion of the templates, not the removal of the regions :facepalm:

charludo: Oh, alright 😄 (Funnily enough I constantly find myself writing `Offer` or `OfferProvider`, seem I'm still influenced by the Compass)

integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -185,6 +185,20 @@ def __init__(self, *args, **kwargs):
...
+                .values_list("embedded_offers", flat=True)
+                .distinct()
+                if offer_template
+            )
+            if self.instance
+            else []
+        )
+        self.fields["offers"].queryset = self.fields["offers"].queryset.exclude(
+            id__in=non_disableable_offers
+        )
```
timobrembeck, line 201: This will do the opposite of what we want: If an offer is removed from the queryset, it is automatically disabled when the form is submitted.

integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -27,6 +27,28 @@
...
+    def create_option(self, *args, **kwargs):
+        """
+        Overwrites the parent's method in order to disable
+        a set of pre-determined options
+
+        :return: a single option
+        :rtype: dict
+        """
+        option = super().create_option(*args, **kwargs)
+        if option["value"] in self.disabled_options:
```
timobrembeck, line 47:
```suggestion
        if option["value"].instance in self.disabled_options:
```

integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -277,6 +312,18 @@ def clean(self):
         ):
             cleaned_data["deepl_midyear_start_month"] = None
 
+        # Make sure no non-disableable offers have been disabled
+        if (
+            self.disabled_offer_options
+            and not all(self.disabled_offer_options) in cleaned_data["offers"]
```
timobrembeck, line 318:
```suggestion
            and not all(offer in cleaned_data["offers"] for offer in self.disabled_offer_options)
```
Maybe this is what you intended?

integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -27,6 +27,28 @@
...
+        """
+        Overwrites the parent's method in order to disable
+        a set of pre-determined options
+
+        :return: a single option
+        :rtype: dict
+        """
+        option = super().create_option(*args, **kwargs)
+        if option["value"] in self.disabled_options:
+            option["attrs"]["disabled"] = True
```
timobrembeck, line 48:
```suggestion
            option["attrs"]["class"] = "fake-disable"
```
`disabled = True` prevents the field from being submitted, so this would be an alternative (though not really satisfying, since it's still possible to disable the field, I guess because of the label tag)...

timobrembeck (changes requested):

integreat_cms/cms/forms/regions/region_form.py
timobrembeck: I guess one solution would be to add
```
label:has(> .fake-disable) {
    @apply text-gray-600 pointer-events-none;
}
```
to our style sheet, but this only works in firefox if users enabled `layout.css.has-selector.enabled` in their config...
Not ideal, but maybe better than nothing?

charludo: Deja-vu. RIght, we've had this topic before, with the HIX/MT interaction 😄 

Back then we resorted to JS, so I guess I'll do the same here.

integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -185,6 +207,18 @@ def __init__(self, *args, **kwargs):
...
+            [
+                offer_template
+                for offer_template in Page.objects.filter(region=self.instance)
+                .values_list("embedded_offers", flat=True)
+                .distinct()
+                if offer_template
+            ]
+            if self.instance
+            else []
+        )
```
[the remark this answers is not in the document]
charludo: This is not working as intended, because apparently, when creating a new region, the form has its instance set as a hidden region(?), and somehow `OfferTemplate.objects.filter(pages__region=self.instance)` then returns 3 `OfferTemplate`s.

I am really not sure how or why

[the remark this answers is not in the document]
charludo: I've pushed the current state, if you could take a look at this I'd really appreciate it.

TBF the previous soution did work even if it was inelegant, so worst case we could also revert to it 😅

integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -185,6 +208,10 @@ def __init__(self, *args, **kwargs):
         self.fields["deepl_midyear_start_enabled"].initial = bool(
             self.instance.deepl_midyear_start_month
         )
+        self.disabled_offer_options = OfferTemplate.objects.filter(
+            pages__region=self.instance,
+        )
```
timobrembeck, line 213:
```suggestion
        self.disabled_offer_options = (
            OfferTemplate.objects.filter(pages__region=self.instance)
            if self.instance.id
            else OfferTemplate.objects.none()
        )
```
Ah ok, then maybe like this? :thinking:

timobrembeck: Interesting, probably we can remove all `if self.instance` checks in the region form since all they're doing is checking whether the instance is a correct region objects and has the attributes we expect. But after calling `super().__init__(*args, **kwargs)` on the model form, the instance will never be `None`... I hope I'm not missing something important...

timobrembeck (approved):
Great, thanks a lot! :+1:

Now I wonder whether we should apply the same principle to the `supported_by_app_in_content` checkbox in the offer template form – maybe we should prevent unchecking this when the offer is embedded in at least one page? :sweat_smile: 
But I'd also be fine with merging this first and opening a new issue for this...

integreat_cms/cms/forms/pages/page_form.py
```diff
@@ -102,6 +104,11 @@ def __init__(self, **kwargs):
         # Let mirrored page queryset be empty per default and only fill it if a region is selected
         mirrored_page_queryset = Page.objects.none()
 
+        # Filter the offer provides available for embedding
```
JoeyStk, line 107:
```suggestion
        # Filter the offer providers available for embedding
```

integreat_cms/cms/templates/offertemplates/offertemplate_form.html
```diff
@@ -71,6 +71,15 @@ <h3 class="heading font-bold text-black">
                         </div>
                         {% render_field form.thumbnail %}
                     </div>
+                    <div>
+                        <label for="{{ form.supported_by_app_in_content.id_for_label }}">
```
JoeyStk, line 75:
```suggestion
                        <label for="{{ form.supported_by_app_in_content.id_for_label }}" class="flex items-center">
```
I found that checkbox and label weren't aligned (s.screenshot). Items-center should do :)
grafik

JoeyStk (commented):
Thank you very much!
I wasn't able to do a detailed review yet. So far I only have these two small comments. I will hopefully complete my review in the next few days :)

charludo:
> Great, thanks a lot! 👍
> 
> Now I wonder whether we should apply the same principle to the `supported_by_app_in_content` checkbox in the offer template form – maybe we should prevent unchecking this when the offer is embedded in at least one page? 😅 But I'd also be fine with merging this first and opening a new issue for this...

Hm, good point. The only situation I see this ever being an issue is if an offer provider changes/breaks their API and the app is no longer able to display the content.

Not sure what should happen in that situation though.

Maybe the reverse could work: allowing unchecking "supported by app in content" but then showing "warning: some previously embedded offers can currently not be displayed in the app" above the offer selection in the page form?

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -6280,6 +6310,10 @@ msgstr "Reihenfolge"
 msgid "Embed live content"
 msgstr "Live-Inhalte einbinden"
 
+#: cms/templates/pages/page_form_sidebar/settings_box.html
+msgid "Embed offers"
+msgstr "Eingebettete Angebote"
```
JoeyStk, line 6315:
```suggestion
msgstr "Angebote einbinden"
```
This could be just personal preference, but I like "einbidne" a bit more because it suggests activity and therefore maybe helps to communicate the call to action a bit better?

JoeyStk (approved):
Thank you for answering my questions, they helped me to understand a bit better.
The last thought I had was again just minor regarding the translation file :)

PR #2476, opened 2023-10-23
Robust changed content detection

### LINKED ISSUE #2477: Unsaved changes indicator
### Motivation

As a user editing e.g. a page, I want to know whether I have unsaved changes, or notice when an autosave is successful.
This should result in slightly higher perceived reliability of the system through transparency and predictability.

### Proposed Solution

Add a bullet to the document title whenever the form is different from the most recently saved state.

### Alternatives

- Add a banner to the top of the page once changes have been saved
  Might get annoying, especially if it changes the layout / scroll position relative to the focused element. Very verbose compared to the minor role this indicator is supposed to achieve.
- Add an indicator to the tinymce editor
  Might be relatively complex

The proposed solution is not better than these alternatives. But it requires way less effort and is a good starting point.

### User Story

As an editor, I want to know whether I have unsaved changes, or notice when an autosave is successful, so that I have a higher trust in the system.

PeterNerlich:
### Short description

Rework changed content detection from using a dirty flag to comparing with the previous content and add a little indicator. With this we can reliably tell whether we need to autosave at all or not.

### Proposed changes

- Remove dirty flag and determine presence of changes by comparing with previously captured state
- Only attatch beforeunload listener when dirty to allow the browsers back/forward cache (per recommendation: https://developer.chrome.com/articles/page-lifecycle-api/#the-beforeunload-event)
- Prepend bullet to document title when dirty

### Side effects

- When closing the tab with unsaved changes, an autosave is triggered. However, the popup asking whether to really close the page is shown before the server can respond.
  This leads to situations where the content is saved despite a warning being shown. This cannot be changed, since it's impossible to predict whether it will be successful or not. See also #2132 and #2435.

- There is a potential for the value representing the last saved state to be out of sync with what was actually sent as an auto save. This is because of the separation of the component tracking changes for the whole form and the autosave plugin to tinymce which performs the actual autosave: the former only receives the signal that an autosave is being attempted and then that it was indeed successful, but never the actual data that was sent. It might theoretically be possible for the content to change in the tiny instance between the autosave gathering the data to send to the server and the unsaved warning module receiving the signal and itself gathering the data as a candidate for the next saved state.
  It might be possible to mitigate this by moving the autosave itself out of the editor plugin (then at most sending a signal to trigger it), since it requires the whole form data, which is technically out of the scope of a tinymce plugin (also more than one editor could be part of the same form, which would create problems, though we do this nowhere).
  However, this is hopefully very unlikely and should not do any damage if it should occur: The state would likely only be different by a single character until the next autosave or reload, and a wrong save decision because of it would not pose a problem be caught and corrected reasonably soon.

- The change indicator in the title relies on the title not being changed (by JS) over the life cycle of a page in the browser. This is because to reliably add the bullet, the original title is saved in a variable. Currently, we don't even give a different title to different CMS pages.

### Resolved issues

Fixes #2477 

### DIFF
--- a/integreat_cms/static/src/js/forms/tinymce-init.ts
+++ b/integreat_cms/static/src/js/forms/tinymce-init.ts
@@ -56,7 +56,14 @@ const toggleNoTranslate = (editor: Editor) => {
     }
 };
 
-export const getContent = (): string => tinymce.activeEditor.getContent();
+export const getContent = (args: object = {}): string => {
+    const defaultArgs = { source_view: true };
+    return tinymce.activeEditor.getContent({ ...{ defaultArgs }, ...args });
+};
+
+export const editors = tinymce.editors;
+
+export const isActuallyDirty = (editor: Editor) => editor.startContent !== editor.getContent({ source_view: true });
 
 /**
  * This file initializes the tinymce editor.
@@ -89,6 +96,7 @@ window.addEventListener("load", () => {
             },
             link_title: false,
             autosave_interval: "120s",
+            autosave_ask_before_unload: false, // We implement that ourselves
             forced_root_block: true,
             plugins: "code fullscreen autosave preview media image lists directionality wordcount hr charmap paste",
             external_plugins: {
--- a/integreat_cms/static/src/js/unsaved-warning.ts
+++ b/integreat_cms/static/src/js/unsaved-warning.ts
@@ -1,40 +1,109 @@
 /**
  * This file contains a function to warn the user when they leave a content without saving
  */
+import { editors, isActuallyDirty } from "./forms/tinymce-init";
 
-let dirty = false;
+const takeFormSnapshot = (form: HTMLFormElement) => new FormData(form);
 
-window.addEventListener("beforeunload", (event) => {
-    // trigger only when something is edited and no submit/save button clicked
-    if (dirty) {
-        event.preventDefault();
-        /* eslint-disable-next-line no-param-reassign */
-        event.returnValue = "This content is not saved. Would you leave the page?";
+const equalSnapshots = (a: FormData, b: FormData) => {
+    // We could inplement a proper comparison of nested arrays
+    // …ooor we could just take the hit and stringify everything
+    // to JSON, saving us maintenance on all these code lines
+    if (a == null || b == null) {
+        return false;
     }
-});
+    const withoutExcluded = (iterator: IterableIterator<[string, FormDataEntryValue]>) =>
+        Array.from(iterator).filter((item) => {
+            const target = document.querySelector(`[name="${item[0]}"]`);
+            return !target.hasAttribute("data-unsaved-warning-exclude");
+        });
+    const aJSON = JSON.stringify(withoutExcluded(a.entries()));
+    const bJSON = JSON.stringify(withoutExcluded(b.entries()));
+    return aJSON === bJSON;
+};
+
+const originalTitle = document.title;
+
+let snapshotCandidate: FormData | null = null;
+let savedSnapshot: FormData | null = null;
+
+const isDirty = (form: HTMLFormElement) => {
+    const now = takeFormSnapshot(form);
+    if (!Array.from(now.entries()).length) {
+        console.warn("form snapshot has no entries:", now.entries(), form);
+    }
+    if (!equalSnapshots(now, savedSnapshot)) {
+        return true;
+    }
+    // Are there any unsaved changes in the editors?
+    for (const editor of editors) {
+        if (isActuallyDirty(editor)) {
+            return true;
+        }
+    }
+    return false;
+};
+
+// Only temporarily add a beforeunload event
+// https://developer.chrome.com/articles/page-lifecycle-api/#the-beforeunload-event
+const beforeunload = (event: BeforeUnloadEvent) => {
+    console.debug("[beforeunload]");
+    event.preventDefault();
+    /* eslint-disable-next-line no-param-reassign */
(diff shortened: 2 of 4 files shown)


PeterNerlich:
On second thought, there is no reason for this to require merging *after* #2401

seluianova:
> On second thought, there is no reason for this to require merging after https://github.com/digitalfabrik/integreat-cms/pull/2401

@PeterNerlich so, is it still actual?

PeterNerlich:
> @PeterNerlich so, is it still actual?

Yes, feel free to review the PR and/or give an opinion on the feature with the bullet in the document title

seluianova (changes requested):
✳️ I'm observing some strange side effects on this branch.
For example:
1. Open Willkommen page in the editor
2. Add a point (.) at the end of the sentence
3. Publish --> success
4. Remove a point at the end of the sentence
5. Publish --> the point is returned back, "No changes detected, but date refreshed" message is shown.

I couldn't reproduce it on develop branch.

✳️ In addition, I have the following doubt:
I have a feeling that I, as a user, probably wouldn't notice the bullet point in the title...
(But I'm open for other opinions here)

✳️ Also, in some cases the indicator doesn't appear when changes are made, but so far I haven't been able to determine the exact steps to reproduce this (it seems pretty random?).

integreat_cms/static/src/js/unsaved-warning.ts
```diff
@@ -1,40 +1,101 @@
...
+    }
+    // Are there any unsaved changes in the editors?
+    for (const editor of editors) {
+        if (isActuallyDirty(editor)) {
+            return true;
         }
-        if (!dirty) {
-            console.debug("editing detected, enabled beforeunload warning");
+    }
+    return false;
```
david-venhoff, line 41: I could reproduce the same problem that @seluianova mentioned, were the modified input would not get submitted sometimes, and could not figure out how exactly it is caused. However, commenting out these lines fixes the problem for me somehow :shrug:

david-venhoff (changes requested):
Thanks for the pr! Unfortunately it seems to cause the editor to behave weirdly sometimes and I am not really sure why.

Currently we emit an input event for the form when the editor becomes dirty for the unsaved warning to update. This can probably be removed now: https://github.com/digitalfabrik/integreat-cms/blob/dbd5c7d0bc498a9e36bd1c90e370c26aada735d9/integreat_cms/static/src/js/forms/tinymce-init.ts#L222-L229
Also the `contentChanged` event that gets emitted on `keyUp` for the hix widget should probably be migrated to use your system: https://github.com/digitalfabrik/integreat-cms/blob/dbd5c7d0bc498a9e36bd1c90e370c26aada735d9/integreat_cms/static/src/js/forms/tinymce-init.ts#L231-L234

integreat_cms/static/src/js/unsaved-warning.ts
```diff
@@ -1,40 +1,101 @@
...
+    } else {
+        window.removeEventListener("beforeunload", beforeunload);
+        document.title = originalTitle;
+    }
+};
+
+window.addEventListener("load", () => {
+    const form = document.querySelector("[data-unsaved-warning]") as HTMLFormElement;
+    // Remember the state on initialization as saved
+    savedSnapshot = takeFormSnapshot(form);
```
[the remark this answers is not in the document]
PeterNerlich: Apparently it's perfectly fine. Proven by changing the selector to `[data-unsHaved-warning]`

PeterNerlich:
Thank you for bringing up that issue! I totally missed that this happens.
It was annoying to figure out how to fix, but this is how I think it occurred: TinyMCE hides the original text field it's supposed to represent and builds its own interface, requiring it to save its state back to the form element before the form data gets captured and sent in order to be up to date. This is taken care of for the autosave functionality, but this didn't happen for the regular form submit.

So in order to ensure this also for the normal submit, I added a *formdata* listener that modifies the formdata gathered on submit *(and also when calling `new FormData(form)`)* to contain the latest tinymce content.

integreat_cms/static/src/js/unsaved-warning.ts
PeterNerlich: Please see whether you can still reproduce it or whether my attempt to fix it was successful

seluianova:
Another odd find, no idea what it has to do with the current changes, but I've seen it several times on this PR-branch:
image

integreat_cms/static/src/js/forms/tinymce-init.ts
```diff
@@ -56,7 +56,14 @@ const toggleNoTranslate = (editor: Editor) => {
     }
 };
 
-export const getContent = (): string => tinymce.activeEditor.getContent();
+export const getContent = (args: object = {}): string => {
+    const defaultArgs = { format: "raw" };
```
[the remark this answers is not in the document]
PeterNerlich: > I cannot find any evidence that this does anything else than `format: "raw"`

Okay, it's just buried in piles upon piles of spaghetti code. Testing it out should have been the way to go after all. `source_view: true` seem reasonably stable compared to `format: 'raw'`

integreat_cms/static/src/js/unsaved-warning.ts
seluianova: @PeterNerlich I still can reproduce it

seluianova:
I can still reproduce the problem where changes made are lost after clicking the "Update" button.

JoeyStk:
Closed as stale

PR #2479, opened 2023-10-24
Add empty tag to empty pages with only archived subpages

MizukiTemma:
### Short description

This PR adds an empty tag to empty pages whose subpages are all archived.

### Proposed changes

- Add to the condition for the empty tag the case where all the subpages are archived 
- Adjust `get_non_archived_children` to choose whether the parent page should be included or not.

### Side effects

- I hope none
- but I may have misunderstood the desired behaviour 🙈 

### Resolved issues

Fixes: #1175 

### DIFF
--- a/integreat_cms/cms/models/pages/page_translation.py
+++ b/integreat_cms/cms/models/pages/page_translation.py
@@ -266,7 +266,7 @@ class PageTranslation(AbstractBasePageTranslation):
         This functions returns a list of translated tags which apply to this function.
         Supported tags:
         * Live content: if the page of this translation has live content
-        * Empty: if the page contains no text and has no subpages (TODO: Can also be empty if subpages are archived)
+        * Empty: if the page contains no text and has no non-archived subpages
 
         :return: A list of tags which apply to this translation
         """
@@ -275,7 +275,10 @@ class PageTranslation(AbstractBasePageTranslation):
         if self.page.mirrored_page:
             tags.append(_("Live content"))
 
-        if self.is_empty and self.page.is_leaf():
+        if self.is_empty and (
+            self.page.is_leaf()
+            or not self.page.children.filter(explicitly_archived=False).exists()
+        ):
             tags.append(_("Empty"))
 
         return tags
--- a/integreat_cms/release_notes/current/unreleased/1175.yml
+++ b/integreat_cms/release_notes/current/unreleased/1175.yml
@@ -1,0 +1,2 @@
+en: Add empty tag for empty pages with only archived subpages
+de: Füge "Leer" Tag zu leeren Seiten mit nur archivierten Unterseiten hinzu


david-venhoff (approved):
Thanks, LGTM!

integreat_cms/cms/models/pages/page.py
```diff
@@ -256,7 +256,7 @@ def archived(self) -> bool:
         """
         return self.explicitly_archived or self.implicitly_archived
 
-    def get_non_archived_children(self) -> Iterator[Page]:
+    def get_non_archived_children(self, include_parent: bool | None = True) -> Iterator[Page]:
```
timobrembeck, line 259:
```suggestion
    def get_non_archived_children(self, include_parent: bool = True) -> Iterator[Page]:
```
If an optional parameter has a default value that is not `None`, we don't need the `| None` as type hint: Either, `True`/`False` is directly passed, then it's a `bool`, or it is omitted, then it is set to `True` which is also a `bool`. So the variable `include_parent`  is always a `bool` and never `None`.

sarahsporck (approved):
Implementation looks good 👍🏼. 
Would be great to have tests for the tags though, but if it is not in scope for this pr, this can also be done later.

integreat_cms/cms/models/pages/page_translation.py
```diff
@@ -275,7 +275,10 @@ def tags(self) -> list[Promise]:
         if self.page.mirrored_page:
             tags.append(_("Live content"))
 
-        if self.is_empty and self.page.is_leaf():
+        if self.is_empty and (
+            self.page.is_leaf()
+            or not any(self.page.get_non_archived_children(include_parent=False))
```
timobrembeck, line 280:
```suggestion
            or not self.page.children.filter(explicitly_archived=False).exists()
```
I think it's unnecessary to query the entire subtree of a page to determine whether it has archived descendants. Shouldn't it be sufficient to look at the direct children, and if no un-archived direct children exist, then we know that all its grand-children are either explicitly or implicitly archived?

timobrembeck (changes requested):
Thanks! :+1:
I see a slight potential for improving the performance, and could you remove the "TODO" comment in line 269?

MizukiTemma:
> Implementation looks good 👍🏼. Would be great to have tests for the tags though, but if it is not in scope for this pr, this can also be done later.

hmm 🤔 I don't think test for the tags can be well integreated into the current view based test... I'll leave it as it is for now.

integreat_cms/cms/models/pages/page.py
```diff
@@ -256,7 +256,7 @@ def archived(self) -> bool:
         """
         return self.explicitly_archived or self.implicitly_archived
 
-    def get_non_archived_children(self) -> Iterator[Page]:
+    def get_non_archived_children(self, include_parent: bool = True) -> Iterator[Page]:
```
timobrembeck, line 259:
```suggestion
    def get_non_archived_children(self) -> Iterator[Page]:
```
Could you also revert this change that is no longer needed?

integreat_cms/cms/models/pages/page.py
```diff
@@ -272,7 +272,7 @@ def get_non_archived_children(self) -> Iterator[Page]:
                 or child_page.parent_id in archived_parents
             ):
                 archived_parents.add(child_page.id)
-            else:
+            elif include_parent or not child_page == self:
```
timobrembeck, line 275:
```suggestion
            else:
```

integreat_cms/cms/models/pages/page_translation.py
```diff
@@ -266,7 +266,7 @@ def tags(self) -> list[Promise]:
         This functions returns a list of translated tags which apply to this function.
         Supported tags:
         * Live content: if the page of this translation has live content
-        * Empty: if the page contains no text and has no subpages (TODO: Can also be empty if subpages are archived)
+        * Empty: if the page contains no text and has no subpages
```
timobrembeck, line 269:
```suggestion
        * Empty: if the page contains no text and has no non-archived subpages
```

timobrembeck (changes requested):

timobrembeck (commented):
Thanks! After squashing the commits, I have no further objections :grin:

timobrembeck (approved):

Issue #2466, opened 2023-10-16
Hide links from old versions in link checker


MizukiTemma:
### Describe the Bug

Internal server error will be caused during link replacement in the broken link list.

### Steps to Reproduce

1. Go to https://admin.integreat-app.de/lk-emmendingen/analytics/linkcheck/invalid/?page=2
2. Find the link with linktext "Hauptstraße 15, 79341 Kenzingen"
3. Replace the URL with "https://www.google.com/maps/place/Hauptstra%C3%9Fe+15,+79341+Kenzingen/@48.1918235,7.7665451,17z/data=!3m1!4b1!4m6!3m5!1s0x47913cec05b5a17f:0xe65c8600197c8b17!8m2!3d48.19182!4d7.76912!16s%2Fg%2F11c5fw72pv?entry=ttu"
4. . See error

### Expected Behavior

The link will be successfully updated.

### Actual Behavior

Internal server error occurs

### Additional Information

This issue is related to #2465 
<details>
  <summary>Traceback</summary>

  ```
Oct 16 09:24:55 ERROR   django.request - 500 Internal Server Error: /lk-emmendingen/analytics/linkcheck/invalid/746380/
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "pagetranslation_unique_version"
DETAIL:  Key (page_id, language_id, version)=(14470, 74, 5) already exists.

The above exception was the direct cause of the following exception:
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
[...]

PR #2481, opened 2023-10-25
Hide links from old versions in link checker

MizukiTemma:
### Short description

This PR adds an apropriate message when a link from old versions is about to be replaced

### Proposed changes

- Chekck whether all the queried tranlations are the latest of each translation
- Show the new detailed message

### Side effects

- I know the loop `for translation in translations` is doubled in the function. This is to prevent  it that replacement is successfull in some of the translations but not in the others, so either the link will be replaced in all of the related translations or not at all.
- I left the utility function `replace_links` unchanged, for this works only on the latest versions.

### Resolved issues

Fixes: #2466 

### DIFF
--- a/integreat_cms/cms/utils/linkcheck_utils.py
+++ b/integreat_cms/cms/utils/linkcheck_utils.py
@@ -9,13 +9,19 @@ from typing import TYPE_CHECKING
 from urllib.parse import ParseResult, unquote, urlparse
 
 from django.conf import settings
-from django.db.models import Prefetch, Q
+from django.db.models import Prefetch, Q, Subquery
 from linkcheck import update_lock
 from linkcheck.listeners import tasks_queue
 from linkcheck.models import Link, Url
 from lxml.html import rewrite_links
 
-from ..models import EventTranslation, PageTranslation, POITranslation
+from integreat_cms.cms.models import (
+    EventTranslation,
+    ImprintPageTranslation,
+    PageTranslation,
+    POITranslation,
+    Region,
+)
 
 if TYPE_CHECKING:
     from typing import Any
@@ -31,7 +37,7 @@ def get_urls(
     prefetch_content_objects: bool = True,
 ) -> list[Url]:
     """
-    Count all urls by status, either of a specific region or globally
+    Collect all the urls which appear in the latest versions of the contents of the region, filtered by ID or region if given.
 
     :param region_slug: The slug of the current region
     :param url_ids: The list of requested url ids
@@ -43,15 +49,39 @@ def get_urls(
         # If the results should be limited to specific ids, filter the queryset
         urls = urls.filter(id__in=url_ids)
     if region_slug:
+        region = Region.objects.get(slug=region_slug)
+        latest_pagetranslation_versions = Subquery(
+            PageTranslation.objects.filter(
+                page__id__in=Subquery(region.non_archived_pages.values("pk")),
+            )
+            .distinct("page__id", "language__id")
+            .values_list("pk", flat=True)
+        )
+        latest_poitranslation_versions = Subquery(
+            POITranslation.objects.filter(poi__region=region)
+            .distinct("poi__id", "language__id")
+            .values_list("pk", flat=True)
+        )
+        latest_eventtranslation_versions = Subquery(
+            EventTranslation.objects.filter(event__region=region)
+            .distinct("event__id", "language__id")
+            .values_list("pk", flat=True)
+        )
+        latest_imprinttranslation_versions = Subquery(
+            ImprintPageTranslation.objects.filter(page__region=region)
+            .distinct("page__id", "language__id")
+            .values_list("pk", flat=True)
+        )
         # Get all link objects of the requested region
         region_links = Link.objects.filter(
-            Q(page_translation__page__region__slug=region_slug)
-            | Q(imprint_translation__page__region__slug=region_slug)
-            | Q(event_translation__event__region__slug=region_slug)
-            | Q(poi_translation__poi__region__slug=region_slug)
+            Q(page_translation__id__in=latest_pagetranslation_versions)
+            | Q(imprint_translation__id__in=latest_imprinttranslation_versions)
+            | Q(event_translation__id__in=latest_eventtranslation_versions)
+            | Q(poi_translation__id__in=latest_poitranslation_versions)
         ).order_by("id")
+
         if prefetch_content_objects:
-            region_links = region_links.prefetch_related("content_object")
+            region_links = region_links.prefetch_related("content_object__language")
         # Prefetch all link objects of the requested region
         urls = urls.prefetch_related(
             Prefetch(
--- a/integreat_cms/release_notes/current/unreleased/2466.yml
+++ b/integreat_cms/release_notes/current/unreleased/2466.yml
@@ -1,0 +1,2 @@
+en: Hide links from old versions in link checker
+de: Blende Links aus alten Versionen im Link-Checker aus


timobrembeck (commented):
> This is to prevent it that replacement is successful in some of the translations but not in the others, so either the link will be replaced in all of the related translations or not at all.

Hmm, I'm not sure, maybe it makes sense to change this behavior? E.g. we could replace all links in the latest versions and just ignore the old versions.

MizukiTemma:
> > This is to prevent it that replacement is successful in some of the translations but not in the others, so either the link will be replaced in all of the related translations or not at all.
> 
> Hmm, I'm not sure, maybe it makes sense to change this behavior? E.g. we could replace all links in the latest versions and just ignore the old versions.

I thought it might be confusing for users when the link remains unchanged in some translations but changed in the others 🤔

timobrembeck:
> I thought it might be confusing for users when the link remains unchanged in some translations but changed in the others 🤔

But it'd not completely random, but pretty useful: The link is changed in all newest versions of the translations (which totally makes sense, since this is the content that is viewed by the users and where it's important that the links are working) and it's not changed in all older versions (which is kind of irrelevant, since the users don't see old versions anyway)...

Maybe we should even go one step further and accept the fact that we always will have inconsistencies in the database and already filter out invalid links when the list view is retrieved? We could even program some kind of self-fixing method every time somebody visits the link list view? :thinking:

MizukiTemma:
@timobrembeck 
Thank you for a nice idea 😸 

The second commit includes:
1. Skip link replacement if the link is in an older version, but do if in the latest versions.
2. Exclude link objects of older versions from `region_links`, so they won't shown to users at all. Their trace won't appear in the columns "Link text", "Source" and "Usage" too.

integreat_cms/cms/utils/linkcheck_utils.py
```diff
@@ -37,12 +37,17 @@ def get_urls(region_slug=None, url_ids=None, prefetch_content_objects=True):
...
+        all_region_links = Link.objects.filter(
             Q(page_translation__page__region__slug=region_slug)
             | Q(imprint_translation__page__region__slug=region_slug)
             | Q(event_translation__event__region__slug=region_slug)
             | Q(poi_translation__poi__region__slug=region_slug)
         ).order_by("id")
+        region_link_ids = []
+        for region_link in all_region_links:
+            if region_link.content_object == region_link.content_object.latest_version:
+                region_link_ids += [region_link.id]
```
timobrembeck, line 49: Hmm, I'm not too convinced about this loop, this function is too slow already (see #2539) and this loop will cause a lot of additional queries.
Maybe we could calculate the ids of the latest versions in one query beforehand, similar to how we did in the dashboard view? Or find another way how we could move this condition into the sql query itself?

Since we need to improve the performance of this function anyway, we could also dive deeper into what's costing the most time here and maybe refactor the entire function.

integreat_cms/release_notes/current/unreleased/2466.yml
```diff
@@ -0,0 +1,2 @@
+en: Show better error message when replacing links from old versions
+de: Zeige bessere Fehlermeldung beim Ersetzen von Links aus alten Versionen an
```
timobrembeck, line 2:
```suggestion
en: Hide links from old versions in link checker
de: Blende Links aus alten Versionen im Link-Checker aus
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -9347,6 +9347,14 @@ msgstr ""
...
 "gehört ({})."
 
+#~ msgid ""
+#~ "This link cannot be replaced becuase it does not exist in the latest "
+#~ "version of the translation. Please contact an administrator."
+#~ msgstr ""
+#~ "Dieser Link kann nicht ersetzt werden, weil er in der neuesten Version "
+#~ "dieser Übersetzung nicht mehr existiert. Bitte wenden Sie sich an einen "
+#~ "System-Administrator."
+
```
timobrembeck, line 9357:
```suggestion
```

integreat_cms/cms/views/linkcheck/linkcheck_list_view.py
```diff
@@ -175,18 +175,20 @@ def post(self, request, *args, **kwargs):
...
+                        logger.debug("Replacing links of %r", new_translation)
+                        new_translation.content = rewrite_links(
+                            new_translation.content,
+                            partial(self.replace_link, self.instance.url, new_url),
+                        )
+                        # Save translation with replaced content as new minor version
+                        new_translation.id = None
+                        new_translation.version += 1
+                        new_translation.minor_edit = True
+                        new_translation.save()
```
timobrembeck, line 191: This check can be removed if we make sure the region_links only contains the latest versions anyway.


```suggestion
                    new_translation = deepcopy(translation)
                    # Replace link in translation
                    logger.debug("Replacing links of %r", new_translation)
                    new_translation.content = rewrite_links(
                        new_translation.content,
                        partial(self.replace_link, self.instance.url, new_url),
                    )
                    # Save translation with replaced content as new minor version
                    new_translation.id = None
                    new_translation.version += 1
                    new_translation.minor_edit = True
                    new_translation.save()
```

timobrembeck (changes requested):

integreat_cms/cms/utils/linkcheck_utils.py
MizukiTemma: I'm not confident but changed the behaivor so IDs of latest versions of each content type will be collected and then used in the query for `region_link`

integreat_cms/cms/utils/linkcheck_utils.py
```diff
@@ -41,13 +47,37 @@ def get_urls(
...
+            POITranslation.objects.filter(poi__region=region)
+            .distinct("poi__id", "language__id")
+            .values_list("pk", flat=True)
+        )
+        latest_eventtranslation_versions = list(
+            EventTranslation.objects.filter(event__region=region)
+            .distinct("event__id", "language__id")
+            .values_list("pk", flat=True)
+        )
+        latest_imprinttranslation_versions = list(
```
timobrembeck, line 68:
```suggestion
        latest_pagetranslation_versions = Subquery(
            PageTranslation.objects.filter(
                page__id__in=Subquery(region.non_archived_pages.values("pk")),
            )
            .distinct("page__id", "language__id")
            .values_list("pk", flat=True)
        )
        latest_poitranslation_versions = Subquery(
            POITranslation.objects.filter(poi__region=region)
            .distinct("poi__id", "language__id")
            .values_list("pk", flat=True)
        )
        latest_eventtranslation_versions = Subquery(
            EventTranslation.objects.filter(event__region=region)
            .distinct("event__id", "language__id")
            .values_list("pk", flat=True)
        )
        latest_imprinttranslation_versions = Subquery(
```
Here I guess we could also use subqueries instead of plain lists.

integreat_cms/cms/utils/linkcheck_utils.py
```diff
@@ -41,13 +47,37 @@ def get_urls(
...
-            | Q(event_translation__event__region__slug=region_slug)
-            | Q(poi_translation__poi__region__slug=region_slug)
+            Q(page_translation__id__in=latest_pagetranslation_versions)
+            | Q(imprint_translation__id__in=latest_imprinttranslation_versions)
+            | Q(event_translation__id__in=latest_eventtranslation_versions)
+            | Q(poi_translation__id__in=latest_poitranslation_versions)
         ).order_by("id")
+
         if prefetch_content_objects:
             region_links = region_links.prefetch_related("content_object")
```
timobrembeck, line 82:
```suggestion
            region_links = region_links.prefetch_related("content_object__language")
```
I noticed that prefetching the language objects also reduces a few queries.
Just the duplicated region queries are hard to eliminate I guess. But that's not different on the develop branch.

timobrembeck (changes requested):
Thank you for your changes and sorry it took me so long to get back to this PR!
The queries look very good now, I think just a few small optimizations and this is good to go! :+1:

MizukiTemma:
@timobrembeck Thank you 🙏

timobrembeck (approved):
:rocket:

PeterNerlich (commented):
Looks good to me so far! I would need to know how to reproduce the original bug locally in order to confirm that this indeed prevents the Internal Server Error from happening.

Maybe the docstring of `get_urls()` could be adjusted to better reflect the intention of the function? Still speaks of "counting", which wasn't even the point before these changes, as far as I can see.

MizukiTemma:
> Looks good to me so far! I would need to know how to reproduce the original bug locally in order to confirm that this indeed prevents the Internal Server Error from happening.
> 


You can reproduce it with the following steps:
1. Go to a page
2. Insert a link to a website A and save (version n)
3. Delete the link and insert a new link to another website B, and save (version n+1)
5. Go to Django Admin, create an URL object with A and a link object with A and version n
6. You'll see an link to A from version n

MizukiTemma:
> Maybe the docstring of `get_urls()` could be adjusted to better reflect the intention of the function? Still speaks of "counting", which wasn't even the point before these changes, as far as I can see.

Done!

Issue #2482, opened 2023-10-25
Cannot assign authors and editors in a brand new page


MizukiTemma:
### Describe the Bug

It is not possible to add an user as author and editor while creating a brand new page, although this section is shown and there is no warning.

### Steps to Reproduce

1. Create a new page
2. Go to the section "Additional permissions for this page"
3. Add an user as author or editor
4. Save and open the page again to find the assignment not saved

### Expected Behavior

Either this section is hidden until the first save, or save the assignment successfully

### Actual Behavior

No error, page save succeeds but the assignment is not saved. See the trace back which appears when clicking on "Add to authours/editors"

### Additional Information

<details>
 <summary>Traceback (most recent call last):
</summary>

  ```
  File "/home/micucu/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/micucu/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/micucu/integreat-cms/.venv/lib/python3.9/site-packages/django/views/decorators/http.py", line 40, in inner
    return func(request, *args, **kwargs)
  File "/home/micucu/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/home/micucu/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/home/micucu/integreat-cms/integreat_cms/cms/views/pages/page_actions.py", line 628, in grant_page_permission_ajax
    "page": page,
UnboundLocalError: local variable 'page' referenced before assignment
  ```

</details>

JoeyStk:
[...]

PR #2494, opened 2023-10-29
Hide permission section in page form during page creation

JoeyStk:
### Short description

This PR fixes the issue that authors and observers can't be assigned during the page creation.

### Proposed changes

- Hide the permission section during the time that there is not a page id yet (for example page creation)
- Add help text for page creation to inform about this behavior

### Side effects

I think none

### Resolved issues

Fixes: #2482 

### DIFF
--- a/integreat_cms/cms/templates/pages/page_form_sidebar/settings_box.html
+++ b/integreat_cms/cms/templates/pages/page_form_sidebar/settings_box.html
@@ -90,12 +90,16 @@
                 <label>
                     {% translate "Additional permissions for this page" %}
                 </label>
-                <p class="italic">
-                    {% translate "This affects only users, who don't have the permission to change arbitrary pages anyway." %}
-                </p>
-                <div id="page_permission_table">
-                    {% include "pages/_page_permission_table.html" %}
-                </div>
+                {% if page_id %}
+                    <p class="italic">
+                        {% translate "This only affects users who do not have permission to edit arbitrary pages." %}
+                    </p>
+                    <div id="page_permission_table">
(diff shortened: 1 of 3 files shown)


MizukiTemma (approved):
Looks good :+1:

integreat_cms/cms/templates/pages/page_form_sidebar/settings_box.html
```diff
@@ -84,12 +84,16 @@
...
                 </label>
-                <p class="italic">
-                    {% translate "This affects only users, who don't have the permission to change arbitrary pages anyway." %}
-                </p>
-                <div id="page_permission_table">
-                    {% include "pages/_page_permission_table.html" %}
-                </div>
+                {% if page_id %}
+                    <p class="italic">
+                        {% translate "This affects only users, who don't have the permission to change arbitrary pages anyway." %}
```
charludo, line 89:
```suggestion
                        {% translate "This only affects users who do not have permission to edit arbitrary pages." %}
```
I know this is not really part of this PR, but I'd suggest somewhat changing this formulation. The current formulation is slightly confusing because it implies that *no* user has page editing permissions.

charludo (approved):
LGTM, thank you!

integreat_cms/cms/templates/pages/page_form_sidebar/settings_box.html
JoeyStk: Thank you! It took us a while to settle on a suitable German translation, but now we found something we're satisfied with :smile:

PR #2499, opened 2023-10-30
Fix crashing dashboard when region has no default language

PeterNerlich:
### Short description

Fixes a bug when the newly introduced dashboard (#2348) would cause a regression before a new region could be set up, because it assumed regions always have a default language. When trying to access its slug, it crashed, ironically preventing anyone from setting up the language tree through the CMS.

### Proposed changes

- Add checks to detect when there is no default language and don't display the items in that case

### Side effects

I couldn't think of any yet

### Resolved issues

Fixes: #2518

### DIFF
--- a/integreat_cms/cms/templates/dashboard/dashboard.html
+++ b/integreat_cms/cms/templates/dashboard/dashboard.html
@@ -11,7 +11,9 @@
             <div class="xl:w-1/2">
                 <!-- Task Widget -->
                 {% if perms.cms.change_page or perms.cms.change_feedback %}
-                    {% include "dashboard/_todo_dashboard_widget.html" with no_padding=True %}
+                    {% if request.region.default_language %}
+                        {% include "dashboard/_todo_dashboard_widget.html" with no_padding=True %}
+                    {% endif %}
                 {% endif %}
                 <!-- Statistic Widget -->
                 {% if request.region.statistics_enabled and perms.cms.view_statistics %}
--- a/integreat_cms/cms/views/dashboard/dashboard_view.py
+++ b/integreat_cms/cms/views/dashboard/dashboard_view.py
@@ -85,6 +85,9 @@ class DashboardView(TemplateView, ChatContextMixin):
         :return: Dictionary containing the context for unreviewed pages
         :rtype: dict
         """
+        if not self.request.region.default_language:
+            return {}
+
         unreviewed_pages = PageTranslation.objects.filter(
             language__slug=self.request.region.default_language.slug,
             id__in=self.latest_version_ids,
@@ -103,6 +106,9 @@ class DashboardView(TemplateView, ChatContextMixin):
         :return: Dictionary containing the context for auto saved pages
         :rtype: dict
         """
+        if not self.request.region.default_language:
+            return {}
+
         automatically_saved_pages = PageTranslation.objects.filter(
             language__slug=self.request.region.default_language.slug,
             id__in=self.latest_version_ids,
@@ -155,18 +161,19 @@ class DashboardView(TemplateView, ChatContextMixin):
         :return: Dictionary containing the context for pages with low hix value
         :rtype: dict
         """
-        if settings.TEXTLAB_API_ENABLED and self.request.region.hix_enabled:
-            translations_under_hix_threshold = PageTranslation.objects.filter(
-                language__slug__in=settings.TEXTLAB_API_LANGUAGES,
-                id__in=self.latest_version_ids,
-                page__hix_ignore=False,
-                hix_score__lt=settings.HIX_REQUIRED_FOR_MT,
-            )
+        if not settings.TEXTLAB_API_ENABLED or not self.request.region.hix_enabled:
+            return {}
 
-            return {
-                "pages_under_hix_threshold": translations_under_hix_threshold,
-            }
-        return {}
+        translations_under_hix_threshold = PageTranslation.objects.filter(
+            language__slug__in=settings.TEXTLAB_API_LANGUAGES,
+            id__in=self.latest_version_ids,
+            page__hix_ignore=False,
+            hix_score__lt=settings.HIX_REQUIRED_FOR_MT,
+        )
+
+        return {
+            "pages_under_hix_threshold": translations_under_hix_threshold,
+        }
 
     def get_outdated_pages_context(self):
         r"""
@@ -175,6 +182,9 @@ class DashboardView(TemplateView, ChatContextMixin):
         :return: Dictionary containing the context for outdated pages
         :rtype: dict
         """
+        if not self.request.region.default_language:
+            return {}
+
         OUTDATED_THRESHOLD_DATE = datetime.now() - relativedelta(
             days=settings.OUTDATED_THRESHOLD_DAYS
         )


integreat_cms/cms/templates/dashboard/_todo_dashboard_widget.html
```diff
@@ -13,7 +13,7 @@
                 Every day you can find a list of ideas on how you can improve the content of your pages
             {% endblocktranslate %}
         </p>
-        {% if perms.cms.change_page %}
+        {% if perms.cms.change_page and request.region.default_language is not None %}
```
PeterNerlich, line 16: Or maybe we should just check if the relevant keys are in the request context at all?

PeterNerlich:
There is also a `500 Internal Server Error` in the test deployment when going to *Machine Translations* of a region, but I couldn't reproduce that locally yet

MizukiTemma (approved):
Thank you for speedy fix 💪 It works as expected 🙂 

One code style suggestion: you don't need `is not None` when checking something is None or not. Though this is not so important.

MizukiTemma:
> There is also a `500 Internal Server Error` in the test deployment when going to _Machine Translations_ of a region, but I couldn't reproduce that locally yet

Could you open an issue for that?

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -64,23 +64,25 @@ def get_unreviewed_pages_context(self):
...
+                PageTranslation.objects.filter(
+                    status=status.REVIEW,
+                    language__slug=self.request.region.default_language.slug,
+                    page__id__in=Subquery(
+                        self.request.region.non_archived_pages.values("pk")
+                    ),
+                )
+                .order_by("page__id", "language__id", "-version")
+                .distinct("page__id", "language__id")
+                .all()
```
charludo, line 78: I think an early return would be cleaner (sorry, I cannot suggest changes for deleted lines, so tthere are some remnants that will have to be deleted manually). (THis goes for all the other functions as well, of course)

```suggestion
         if not self.request.region.default_language:
            return {}

        unreviewed_pages = (
            PageTranslation.objects.filter(
                status=status.REVIEW,
                language__slug=self.request.region.default_language.slug,
                page__id__in=Subquery(
                    self.request.region.non_archived_pages.values("pk")
                ),
            )
            .order_by("page__id", "language__id", "-version")
            .distinct("page__id", "language__id")
            .all()
        )

        return {
            "unreviewed_pages": unreviewed_pages,
            "default_language_slug": self.request.region.default_language.slug,
        }
```

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -89,25 +91,27 @@ def get_automatically_saved_pages(self):
...
-            .distinct("page__id", "language__id")
-        )
-        automatically_saved_pages = PageTranslation.objects.filter(
-            id__in=Subquery(last_versions.values("pk")),
-            status=status.AUTO_SAVE,
-        ).all()
+            automatically_saved_pages = PageTranslation.objects.filter(
+                id__in=Subquery(last_versions.values("pk")),
+                status=status.AUTO_SAVE,
+            ).all()
```
charludo, line 108:
```suggestion
            )
```
Adding `.all()` after `.filter(...)` has no effect. Same goes for `.all()` after '.distinct(...)` above.

charludo (changes requested):
This indeed fixes the issue at hand, but I am not sure if the more sensible solution wouldn't be to get rid of the default-language constraint all together (see comment in the integreeat-cms channel on Mattermost).

charludo:
> One code style suggestion: you don't need `is not None` when checking something is None or not. Though this is not so important.

In this case I 100% agree, however just to clarify, it is not always the same:

Python distinguishes between truthy/falsy values and actual boolean True/False values.

When you do:
```
assert 2 == True
```
this is False because (boolean) True is not the same as (truthy) 2. (assert bool(2) == True, on the other hand, is True.)

But this actually has a "deeper" meaning:
```
assert 2         # True
assert 2 == True # False
```
may not make sense at first glance, but it's an important distinction when working with (potentially unknown) variables:
```
assert x == True
```
is only True if x is a boolean and is True. However,
```
assert x
```
is True as soon as x has a value (that is not `0`, `0.0`, `[]`, or `None`). Slightly different example:
```
if get_return_value() is None:
    # throw error or something, idk
```
actually checks if x is None. If you get back an empty list, or 0, or False, the condition is false.
```
if not get_return_value():
    # throw error or something, idk
```
s true as soon as you get any non-truthy value.

PeterNerlich:
> > There is also a `500 Internal Server Error` in the test deployment when going to _Machine Translations_ of a region, but I couldn't reproduce that locally yet
> 
> Could you open an issue for that?

It's #2510 now

integreat_cms/cms/templates/dashboard/_todo_dashboard_widget.html
timobrembeck: No, since this template is only shown in the region dashboard, we can assume that `request.region` is a valid region object.

But I agree with the others that `is not None` is not required here, since it's either a `Language` object or `None` and a `Language` object can never evaluate to a Falsey value, so `False` and `None` are indeed equivalent in this context.
```suggestion
        {% if perms.cms.change_page and request.region.default_language %}
```

PeterNerlich: I meant we could make it data dependent and leave it to [the view](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/views/dashboard/dashboard_view.py#L51-L83) to populate the context with variables like `unreviewed_pages`, and if they are missing, just don't display the widget. That would also mean we don't have essentially the same check whether the Textlab API is enabled globally and for that region in both the view and [the template rendering it](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/templates/dashboard/_todo_dashboard_widget.html#L26-L28)

timobrembeck: Ah ok, sorry for the misunderstanding.
Hmm, I guess it would be a bit complicated to determine the difference between an empty variable and an unset variable in the template...
E.g. `{% if pages_under_hix_threshold %}` will be `True` only if pages with a low HIX value exist and `False` if Textlab is disabled or if it's enabled but all pages have a sufficient HIX value... But we want to hide the row if it's disabled and show "All good" if not, so I think it makes sense to keep the current conditions.
It would be possible to determine the difference e.g. in a custom template tag, but in my opinion this would make it more complicated and not easier. Or we could add new context variables for each box, e.g. `show_low_hix_row` and set this from the view...

seluianova (commented):
Nit-picky comment from me:
Maybe we want to check default language also in get_broken_links_context() and get_low_hix_value_context(), and not show those sections if its not set? 

Because no language = no content = no broken links and no bad hix ?
And since we hide other content-related sections instead of showing them with "in total 0" ?

PeterNerlich:
> no language = no content = no broken links and no bad hix ? Since we are hiding all other content-related sections rather than showing them with "in total 0"...

Hmm, that sounds to me like we need to look at what the life cycle of a region is and how the dashboard should behave at each step. What does it mean for a region not to have a default language, in terms of the user? If that only occurs in the beginning between *Creating a region* and *Region is ready to be populated with content*, we might just show a message "Set up languages in order to start adding content", or decide that it is negligible and not worth addressing at all.

timobrembeck:
> Hmm, that sounds to me like we need to look at what the life cycle of a region is and how the dashboard should behave at each step. What does it mean for a region not to have a default language, in terms of the user? If that only occurs in the beginning between _Creating a region_ and _Region is ready to be populated with content_, we might just show a message "Set up languages in order to start adding content", or decide that it is negligible and not worth addressing at all.

Exactly, this is the only case where no language exists normally. So I think showing a message like "Set up languages in order to start adding content" is not needed, just hiding the "to do" box is sufficient.

integreat_cms/cms/templates/dashboard/_todo_dashboard_widget.html
```diff
@@ -13,7 +13,7 @@
                 Every day you can find a list of ideas on how you can improve the content of your pages.
             {% endblocktranslate %}
         </p>
-        {% if perms.cms.change_page %}
+        {% if perms.cms.change_page and request.region.default_language %}
```
timobrembeck, line 16: What do you think about moving this condition [here](https://github.com/digitalfabrik/integreat-cms/blob/4d7020f07aa5952b96de612522baa795b973c59f/integreat_cms/cms/templates/dashboard/dashboard.html#L13) as @seluianova suggested?

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -155,18 +161,22 @@ def get_low_hix_value_context(self):
...
+
+        translations_under_hix_threshold = PageTranslation.objects.filter(
+            language__slug__in=settings.TEXTLAB_API_LANGUAGES,
+            page__id__in=Subquery(
+                self.request.region.non_archived_pages.filter(hix_ignore=False).values(
+                    "pk"
+                )
+            ),
+            hix_score__lt=settings.HIX_REQUIRED_FOR_MT,
+        ).distinct("page_id", "language_id")
```
timobrembeck, line 175:
```suggestion
            id__in=self.latest_version_ids,
            page__hix_ignore=False,
            hix_score__lt=settings.HIX_REQUIRED_FOR_MT,
        )
```
Small mistake in the conflict resolution

integreat_cms/cms/templates/dashboard/dashboard.html
```diff
@@ -10,7 +10,7 @@ <h1 class="heading">
         <div class="xl:flex gap-3 pt-2">
             <div class="xl:w-1/2">
                 <!-- Task Widget -->
-                {% if perms.cms.change_page or perms.cms.change_feedback %}
+                {% if (perms.cms.change_page or perms.cms.change_feedback) and request.region.default_language %}
```
timobrembeck, line 13: Parentheses are not allowed in Django template conditions, you need to stack two separate `{% if %}` tags if you need something like `if (A or B) and C`.
If the condition would be `if A or (B and C)`, you could just write `if A or B and C`, since `and` has higher precedence.

PeterNerlich: Yes, I found that out just now

timobrembeck (approved):

charludo (approved):

PR #2500, opened 2023-10-31
Fix English strings on new Dashboard

charludo:
### Short description

This is a very nitpicky PR and should have been a review (sorry about that).

### Proposed changes

- 4 of 6 sentences end in a period, so the remaining two should as well.
- "Please review, if the content on the page is correct." is not super clear since it appears to say that a review should be added in case the content is correct.

### Side effects

- 
- 

### Resolved issues

Fixes: n/a


MizukiTemma (approved):

JoeyStk (approved):
Thank you :+1:

Issue #2462, opened 2023-10-11
`IntegrityError` when deleting region with feedback


timobrembeck:
### Describe the Bug

It was not possible to delete a region which has feedback

### Steps to Reproduce

1. Go to https://admin.integreat-app.de/regions/testumgebung/
2. Click delete

### Expected Behavior

The region should be deleted

### Actual Behavior

```
psycopg2.errors.ForeignKeyViolation: update or delete on table "cms_feedback" violates foreign key c
onstraint "cms_mapfeedback_feedback_ptr_id_33c98f2f_fk_cms_feedback_id" on table "cms_mapfeedback"
DETAIL:  Key (id)=(159) is still referenced from table "cms_mapfeedback".
```

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/decorators/http.py", line 40, in inner
    return func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/views/regions/region_actions.py", line 95, in delete_region
    deleted_objects = region.delete()
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/db/models/base.py", line 967, in delete
    return collector.delete()
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/db/models/deletion.py", line 435, in delete
    signals.post_delete.send(
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/cacheops/transaction.py", line 77, in __exit__
    self._no_monkey.__exit__(self, exc_type, exc_value, traceback)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/db/transaction.py", line 246, in __exit__
    connection.commit()
[...]

PR #2506, opened 2023-10-31
Fix `IntegrityError` when deleting region with multiple feedback

charludo:
### Short description

Fix `IntegrityError` when deleting region with multiple feedback.

#### Steps that previously worked to reproduce the issue:
1. Submit at least *2 feedback forms of differing types*, for example:
```
curl --location --request POST 'http://localhost:8000/api/augsburg/de/feedback/imprint-page' --form 'comment="test"'
curl --location --request POST 'http://localhost:8000/api/augsburg/de/feedback/map' --form 'comment="test"'
```
2. It does not matter whether these are read or remain unread, but they **cannot** be for the same endpoint (e.g. two feedback submissions for the map)
3. Attempt to delete the region "Stadt Augsburg" and receive the following error:
```
django.db.utils.IntegrityError: update or delete on table "cms_feedback" violates foreign key constraint "
cms_imprintpagefeedb_feedback_ptr_id_8eb10618_fk_cms_feedb" on table "cms_imprintpagefeedback"
DETAIL:  Key (id)=(1) is still referenced from table "cms_imprintpagefeedback".
```
This should now no longer "work", i.e. be fixed.

### Proposed changes

- Explicitly delete all feedback before deleting the region

### Side effects

- none, I think

### Resolved issues

Fixes: #2462 

### DIFF
--- a/integreat_cms/cms/views/regions/region_actions.py
+++ b/integreat_cms/cms/views/regions/region_actions.py
@@ -91,6 +91,8 @@ def delete_region(request, *args, **kwargs):
     region.events.update(location=None)
     # Prevent ProtectedError when media files get deleted before their usages as organization logo
     region.organizations.all().delete()
+    # Prevent IntegrityError when multiple feedback objects exist
+    region.feedback.all().delete()
     # Delete region and cascade delete all contents
     deleted_objects = region.delete()
     logger.info(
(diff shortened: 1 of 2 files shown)


timobrembeck (approved):
Thanks! :pray:
Could you add a release note for this change?

JoeyStk (approved):
Thank you!

Issue #2478, opened 2023-10-23
Fix page archive message


MizukiTemma:
### Describe the Bug

When an user tries to restore a implicitly archived page by bulk action, it shows "The selected pages were successfully restored" but 
the page won't be (and should not be) restored as long as it has an archived anscestor.

The message needs to be adjusted.

### Steps to Reproduce

1. Go to the list of archived pages
2. Restore an implicitly archived page (status "Archived, because a parent page is archived") by bulk action
3. See the message "The selected pages were successfully restored"
4. See the page remains implecitely archived

### Expected Behavior

Show for each page "Page XYZ was successfully restored" if it was actually restored, otherwise "Page XYZ could not be restored because it has an archived anscester page".

### Actual Behavior

A success message is shown regardless of whether they are actually restored.

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

Issue #2475, opened 2023-10-23
Add expected error/success messages to view tests


timobrembeck:
### Motivation

After #2454, it is now possible to view the user messages in the server logs in debug mode.

### Proposed Solution

We could use this to further improve the view tests by adding a new "expected messages" column to the each entry in the [view config](https://github.com/digitalfabrik/integreat-cms/blob/44cf1912b98059db151ddc64c60ce8121c352dc6/tests/cms/views/view_config.py#L31).
E.g. for normal get-views, we could make sure that no error messages are present, or for form-submit-views we require the correct success message to exist.

Creating the initial baseline version for these kind of tests is probably very time consuming, maybe it's possible to somehow script this by inserting empty lists first and then extracting the actual messages from the failed tests?

PeterNerlich:
We are in a state of indecision regarding the view config, and we are already individually testing for messages in some tests. For the lack of concrete approaches we will put this on the graveyard of ideas for now.

PR #2513, opened 2023-10-31
Fix page archive message

JoeyStk:
### Short description

When attempting to restore an implicitly archived page by bulk action you get a success message, although this behavior is rightfully denied. This PR fixes the feedback message in order to match the behavior we want and we already have implemented.

### Proposed changes

- This PR changes that not one bulk message is given to the user, but instead the user receives feedback for every single page they are trying to restore. At first this gave me a bit of a stomach ache because to me it looked as the user is being spammed with messages, but @MizukiTemma pointed out that we already have this behavior for MT and therefore to the user this is the expected behavior. 
- Show an error message instead of a success message when attempting to restore an implicitly archived page.
- In order to not spam the logger @ulliholtgrave and I decided to create bulk messages for the logger.

### Side effects

- This PR gives out an error message, when attempting to restore a page that was implicitly archived. However if you don't do this via bulk action and instead do it via the single command you get an info message instead of an error message. The content and the context of these messages differ, but too me the situations are still close enough that we should at least think about matching these message to either be both info messages or both error message. I'm okay with all three options :)
- When restoring a page is not successfully the user is still redirected to the page tree. @ulliholtgrave and I also talked about this. However it's difficult to find a clear line where a redirect should go, especially for bulk actions when half of the pages were successfully restored and for the other half restoring was unsuccessfully. Therefore we left it out of this PR and rather suggest to first do some research on what the user expects, then discuss it in the team and then maybe open a new issue for it.

### Resolved issues

Fixes: #2478

### DIFF
--- a/integreat_cms/cms/views/bulk_action_views.py
+++ b/integreat_cms/cms/views/bulk_action_views.py
@@ -315,21 +315,40 @@ class BulkRestoreView(BulkActionView):
         :param \**kwargs: The supplied keyword arguments
         :return: The redirect
         """
+        restore_failed = []
         for content_object in self.get_queryset():
-            content_object.restore()
+            if self.get_queryset().model is Page and content_object.implicitly_archived:
+                messages.error(
+                    request,
+                    _(
+                        'The selected {} "{}" could not be restored because it has an archived ancestor page'
+                    ).format(
+                        self.model._meta.verbose_name,
+                        content_object.best_translation.title,
+                    ),
+                )
+                restore_failed.append(content_object)
+            else:
+                content_object.restore()
+                messages.success(
+                    request,
+                    _('The selected {} "{}" was successfully restored').format(
+                        self.model._meta.verbose_name,
+                        content_object.best_translation.title,
+                    ),
+                )
 
         # Invalidate cache
         invalidate_model(self.model)
         logger.debug(
             "Restored %r by %r",
-            self.get_queryset(),
+            self.get_queryset().exclude(id__in=[page.id for page in restore_failed]),
             request.user,
         )
-        messages.success(
-            request,
-            _("The selected {} were successfully restored").format(
-                self.model._meta.verbose_name_plural
-            ),
+        logger.debug(
+            "Could not be restored %r by %r",
+            restore_failed,
+            request.user,
         )
 
         return super().post(request, *args, **kwargs)
--- a/integreat_cms/cms/views/bulk_action_views.py
+++ b/integreat_cms/cms/views/bulk_action_views.py
@@ -18,6 +18,7 @@ from django.views.generic.list import MultipleObjectMixin
 
 from ..constants import status
 from ..models import Page
+from .utils.stringify_list import iter_to_string
 from .utils.translation_status import change_translation_status
 
 if TYPE_CHECKING:
@@ -318,38 +319,41 @@ class BulkRestoreView(BulkActionView):
         restore_failed = []
         for content_object in self.get_queryset():
             if self.get_queryset().model is Page and content_object.implicitly_archived:
-                messages.error(
-                    request,
-                    _(
-                        'The selected {} "{}" could not be restored because it has an archived ancestor page'
-                    ).format(
-                        self.model._meta.verbose_name,
-                        content_object.best_translation.title,
-                    ),
-                )
                 restore_failed.append(content_object)
             else:
                 content_object.restore()
-                messages.success(
-                    request,
-                    _('The selected {} "{}" was successfully restored').format(
-                        self.model._meta.verbose_name,
-                        content_object.best_translation.title,
+
+        if restore_failed:
+            messages.error(
+                request,
+                _(
+                    "The following {} could not be restored because they have at least one archived parent {}: {}"
+                ).format(
+                    self.model._meta.verbose_name_plural,
+                    self.model._meta.verbose_name,
+                    iter_to_string(
+                        [object.best_translation.title for object in restore_failed]
                     ),
-                )
+                ),
+            )
+
+        restore_succeeded = self.get_queryset().exclude(
+            id__in=[page.id for page in restore_failed]
+        )
+
+        if restore_succeeded:
+            messages.success(
+                request,
+                _("The following {} were successfully restored: {}").format(
+                    self.model._meta.verbose_name_plural,
+                    iter_to_string(
+                        [object.best_translation.title for object in restore_succeeded]
+                    ),
+                ),
+            )
 
         # Invalidate cache
         invalidate_model(self.model)
-        logger.debug(
-            "Restored %r by %r",
-            self.get_queryset().exclude(id__in=[page.id for page in restore_failed]),
-            request.user,
-        )
-        logger.debug(
-            "Could not be restored %r by %r",
-            restore_failed,
-            request.user,
-        )
 
         return super().post(request, *args, **kwargs)
 
--- a/integreat_cms/cms/views/utils/stringify_list.py
+++ b/integreat_cms/cms/views/utils/stringify_list.py
@@ -1,0 +1,21 @@
+from django.utils.translation import gettext_lazy as _
+
+
+def iter_to_string(iterable: list) -> str:
+    """
+    This is a helper function that turns a list into a string using different delimiters.
+
+    :param iterable: List of items
+    :return: Returns string with the single items separated by comma expect the last item which is separated by 'and'.
+    """
+    # Convert iterable to list to support querysets etc.
+    lst = list(iterable)
+    # If the list contains more than 1 element, save the last element for later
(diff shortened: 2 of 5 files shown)


timobrembeck:
> This PR changes that not one bulk message is given to the user, but instead the user receives feedback for every single page they are trying to restore. At first this gave me a bit of a stomach ache because to me it looked as the user is being spammed with messages, but @MizukiTemma pointed out that we already have this behavior for MT and therefore to the user this is the expected behavior.

Hmm, I don't really buy that argument. We just did it that way because it would have been complicated otherwise, not because it was the best solution from a UX perspective.
In my opinion, your initial doubts were justified and it would actually be better to collect all the succeeded translation in a list first and then show one single message with all pages that succeeded. And in the long term also adapt this behavior for MTs.

> In order to not spam the logger @ulliholtgrave and I decided to create bulk messages for the logger.

I don't think it's a good idea to modify logging behavior here, since we want to use the logging messages for testing later (see #2475) and it might cause misunderstandings if the logging messages do not match the real messages.

ulliholtgrave:
> > In order to not spam the logger @ulliholtgrave and I decided to create bulk messages for the logger.
> 
> I don't think it's a good idea to modify logging behavior here, since we want to use the logging messages for testing later (see #2475) and it might cause misunderstandings if the logging messages do not match the real messages.

Actually we didn't change the logging behavior or create bulk messages. We just removed the falsely mentioned pages and added another logger error with pages that couldn't be restored.

timobrembeck:
> Actually we didn't change the logging behavior or create bulk messages. We just removed the falsely mentioned pages and added another logger error with pages that couldn't be restored.

Ok then nevermind, I just wanted to mention it so you don't invest time into this approach.

JoeyStk:
> Hmm, I don't really buy that argument. We just did it that way because it would have been complicated otherwise, not because it was the best solution from a UX perspective.
In my opinion, your initial doubts were justified and it would actually be better to collect all the succeeded translation in a list first and then show one single message with all pages that succeeded. And in the long term also adapt this behavior for MTs.

Yes, seems reasonable. I will change this once we have the bugs on the dashboard under control. Until then, I assume that this PR has rather low priority.

JoeyStk:
The new commit does the following:

- adds release note
- bundles all error messages and all success message into one message each
- introduces a function to return a proper string from the list using different delimiter
- removes the logging within the bulk action function

While working on this two questions came up:

1. I saw that there is a counter for the number of events and POIs that have been archived, but there is no such counter for pages. This raised the question if this was on purpose or not. If not, I would open a new issue for it
2. Since we introduce bundled messages with this PR I wanted to suggest to open at least two subsequent issues a. to also bundle the feedback messages when archiving a page/poi/event and b. to also bundle the feedback message for MT. Is this okay and am I forgetting something?

integreat_cms/cms/views/utils/stringify_list.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    titles_as_str = ""
+    for index, item in enumerate(list_of_items):
+        if int(index) == len(list_of_items) - 2:
+            titles_as_str += item + " & "
+        elif int(index) == len(list_of_items) - 1:
+            titles_as_str += item
+        else:
+            titles_as_str += item + ", "
+    return titles_as_str
```
JoeyStk, line 19: I assume there is some Python magic how to do it in one line using join, but I couldn't get it to work. Suggestion how to make this more concise are very welcome

timobrembeck (commented):
> I saw that there is a counter for the number of events and POIs that have been archived, but there is no such counter for pages. This raised the question if this was on purpose or not. If not, I would open a new issue for it

Not completely sure, I think the reason was that it's a bit more complicated to calculate the number of archived pages, so it did slow down the page tree load. But maybe this is resolved in the mean time because of our new subqueries.

> Since we introduce bundled messages with this PR I wanted to suggest to open at least two subsequent issues a. to also bundle the feedback messages when archiving a page/poi/event and b. to also bundle the feedback message for MT. Is this okay and am I forgetting something?

Yes, I think it makes sense to use this new util for all bulk actions.

integreat_cms/cms/views/bulk_action_views.py
```diff
@@ -330,22 +331,44 @@ def post(self, request, *args, **kwargs):
...
+                content_object.restore()
+
+        if restore_failed:
+            messages.error(
+                request,
+                _(
+                    "The following {} could not be restored because they have an archived ancestor {}: {}"
+                ).format(
+                    self.model._meta.verbose_name_plural,
+                    self.model._meta.verbose_name_plural,
```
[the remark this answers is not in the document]
JoeyStk: Ah right, I adjusted it so both translations fit the singular form now

integreat_cms/cms/views/utils/stringify_list.py
JoeyStk: Ah yes, applied! I just renamed iter to _iter so we can get rid off the disable pylint and then pylint remarked that I can also remove the doc because for iter it's already obvious

JoeyStk:
> Not completely sure, I think the reason was that it's a bit more complicated to calculate the number of archived pages, so it did slow down the page tree load. But maybe this is resolved in the mean time because of our new subqueries.

Thank you for clarifying. Makes sense. What do you think would be the better way how to handle this? Create an issue and add it to the backlog and keep the option to reject it later on - or should we just leave it as it is?

> Yes, I think it makes sense to use this new util for all bulk actions.

Done!

MizukiTemma (approved):
Looks good 🙂 Thank you!

integreat_cms/cms/views/utils/stringify_list.py
timobrembeck: Ok, however, a underscore typically indicates that the variable is unused, right?
Maybe something like `iterable` makes more sense?

timobrembeck: And pylint did not mark the docstring as "obvious", but it complained the fact that you wanted to document an ignored parameter.
In the end, this would mean that this parameter is not visible in the docs, which is kind of sad when developers want to consult the documentation about it...

timobrembeck (approved):

integreat_cms/cms/views/utils/stringify_list.py
JoeyStk: > Ok, however, a underscore typically indicates that the variable is unused, right?

I didn't know this convention. [PEP8](
https://peps.python.org/pep-0008/#descriptive-naming-styles) recommends a trailing not a leading underscore in order not to interfere with any Python keywords. But I'm also fine with iterable, which to me is the more consistent solution

timobrembeck: Yeah, I guess the only "official" convention of the leading underscore is to indicate "private" methods and variables. However, for a function argument this does not make sense, since they only live in the local scope and are not accessible from the outside anyways.
So it just seems to be an unofficial convention to treat function arguments starting with an underscore as "dummy variables", e.g. pylint defines them here:

https://github.com/pylint-dev/pylint/blob/ae344fb04e9caddd69db4364d7c55b6e25395a0e/pylint/checkers/variables.py#L1213

I know a lot of projects that also use `_` as variable name for unused parameters, but this clashes with the gettext translation alias, so for Django projects this is not a good choice... (I noticed we still have a few occurrences of this pattern in our code base, maybe we should fix that some time :sweat_smile:)

The [Google style guide](https://google.github.io/styleguide/pyguide.html) even suggest completely to avoid underscores altogether and `del` the unused variables instead:

> Unused argument warnings can be suppressed by deleting the variables at the beginning of the function. Always include a comment explaining why you are deleting it. "Unused." is sufficient. For example:
> ```
> def viking_cafe_order(spam: str, beans: str, eggs: str | None = None) -> str:
>     del beans, eggs  # Unused by vikings.
>     return spam + spam + spam
> ```
> Other common forms of suppressing this warning include using `_` as the identifier for the unused argument or prefixing the argument name with `unused_`, or assigning them to `_`. These forms are allowed but no longer encouraged. These break callers that pass arguments by name and do not enforce that the arguments are actually unused.

As for the trailing underscore: Didn't know that, yeah also makes sense I guess, although I think e.g. `cls` is a lot more widespread than `class_`. In my opinion, variable names that have easy to understand alternatives should use them, but we could keep the trailing underscore option in mind when we face a similar problem in the future.

JoeyStk: That was very insightful, thank you! 

> (I noticed we still have a few occurrences of this pattern in our code base, maybe we should fix that some time 😅)

Yes, I will open an issue for that

>  I think e.g. cls is a lot more widespread than class_. In my opinion, variable names that have easy to understand alternatives should use them, but we could keep the trailing underscore option in mind when we face a similar problem in the future.

I agree!

Issue #2408, opened 2023-09-06
`IntegrityError` when deleting user who marked feedback as read


timobrembeck:
### Describe the Bug

When a user is deleted who marked a specific feedback entry as read, an internal server error occurs.

### Steps to Reproduce

1. Login as user who is deleted later
2. Mark feedback as read
3. Login from admin account and delete that user

### Expected Behavior

The user should be deleted, not sure what should happen to the feedback:
- It could be marked unread again if `SET_NULL` is used
- It could be deleted when `CASCADE` is used
- We could also set the `read_by` field to another user, either a dummy user or an existing one in the region
- We could define another field which denotes that the entry is read, but from a deleted user

### Actual Behavior

An internal server error occurs:
```
django.db.utils.IntegrityError: update or delete on table "cms_user" violates foreign key constraint "cms_feedback_read_by_id_fbaaab26_fk_cms_user_id" on table "cms_feedback"
DETAIL:  Key (id)=(668) is still referenced from table "cms_feedback".
```

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
Sep 05 09:27:58 ERROR   django.request - 500 Internal Server Error: /users/668/delete/
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/db/backends/base/base.py", line 242, in _commit
    return self.connection.commit()
psycopg2.errors.ForeignKeyViolation: update or delete on table "cms_user" violates foreign key constraint "cms_feedback_read_by_id_fbaaab26_fk_cms_user_id" on table "cms_feedback"
DETAIL:  Key (id)=(668) is still referenced from table "cms_feedback".

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/decorators/http.py", line 40, in inner
    return func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
[...]

PR #2514, opened 2023-10-31
Fix IntegrityError when deleting user

seluianova:
### Short description
This PR fixes the following case:
1. the user was assigned to more than one region (z.B. A and B)
2. the user read a feedback in the region A
3. the user was removed from the region A (but remained in the region B)

Then, if we try to remove the user from region B (or in general, in "Network management"), IntegrityError will occur.

```
django.db.utils.IntegrityError: update or delete on table "cms_user" violates foreign key constraint "cms_feedback_read_by_id_fbaaab26_fk_cms_user_id" on table "cms_feedback"
DETAIL:  Key (id)=(668) is still referenced from table "cms_feedback".
```

### Proposed changes
- Mark feedback read by the user as unread before deleting the user

### Side effects
- No?

### Resolved issues
Fixes: #2408

### DIFF
--- a/integreat_cms/cms/views/users/region_user_actions.py
+++ b/integreat_cms/cms/views/users/region_user_actions.py
@@ -36,9 +36,13 @@ def delete_region_user(request, region_slug, user_id):
 
     region = request.region
     user = get_object_or_404(region.region_users, id=user_id)
+
     if user.regions.count() == 1:
-        logger.info("%r deleted %r", request.user, user)
+        # Mark feedback read by the user as unread to prevent IntegrityError
+        user.feedback.update(read_by=None)
+
         user.delete()
+        logger.info("%r deleted %r", request.user, user)
         messages.success(
             request,
             _('Account "{}" was successfully deleted.').format(user.full_user_name),
--- a/integreat_cms/cms/views/users/user_actions.py
+++ b/integreat_cms/cms/views/users/user_actions.py
@@ -32,9 +32,12 @@ def delete_user(request, user_id):
     """
 
     user = get_object_or_404(get_user_model(), id=user_id)
-    logger.info("%r deleted %r", request.user, user)
-    user.delete()
 
+    # Mark feedback read by the user as unread to prevent IntegrityError
+    user.feedback.update(read_by=None)
+
+    user.delete()
+    logger.info("%r deleted %r", request.user, user)
     messages.success(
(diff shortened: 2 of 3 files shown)


integreat_cms/cms/views/users/region_user_actions.py
```diff
@@ -36,9 +37,13 @@ def delete_region_user(request, region_slug, user_id):
 
     region = request.region
     user = get_object_or_404(region.region_users, id=user_id)
+
     if user.regions.count() == 1:
-        logger.info("%r deleted %r", request.user, user)
+        # Mark feedback read by the user as unread to prevent IntegrityError
+        Feedback.objects.filter(read_by=user).update(read_by=None)
```
timobrembeck, line 43:
```suggestion
        user.feedback.update(read_by=None)
```
I think this would work as well, but doesn't make a huge difference.

integreat_cms/cms/views/users/region_user_actions.py
```diff
@@ -8,6 +8,7 @@
 from django.utils.translation import gettext_lazy as _
 from django.views.decorators.http import require_POST
 
+from ....cms.models import Feedback
```
timobrembeck, line 11:
```suggestion
```
Then this import wouldn't be required

integreat_cms/cms/views/users/user_actions.py
```diff
@@ -9,6 +9,7 @@
 from django.utils.translation import gettext_lazy as _
 from django.views.decorators.http import require_POST
 
+from ....cms.models import Feedback
```
timobrembeck, line 12:
```suggestion
```

integreat_cms/cms/views/users/user_actions.py
```diff
@@ -32,9 +33,12 @@ def delete_user(request, user_id):
     """
 
     user = get_object_or_404(get_user_model(), id=user_id)
-    logger.info("%r deleted %r", request.user, user)
-    user.delete()
 
+    # Mark feedback read by the user as unread to prevent IntegrityError
+    Feedback.objects.filter(read_by=user).update(read_by=None)
```
timobrembeck, line 38:
```suggestion
    user.feedback.update(read_by=None)
```

timobrembeck (approved):
Thanks! :+1:

MizukiTemma (approved):
Thank you 👍

Issue #2507, opened 2023-10-31
"In total" calculation of pages with low hix value is incorrect


JoeyStk:
### Describe the Bug

The section "pages with low hix value" also shows an incorrect "in total" calculation. For example for [LK Emmendingen](https://cms-test.integreat-app.de/lk-emmendingen/) it says, that there are 160 pages with insufficient HIX value. However if you follow the link, you'll see that indeed there are no pages with a HIX value under the threshold of 15. Similar picture for [Kreis Segeberg](https://cms-test.integreat-app.de/kreissegeberg/) and [Munich](https://cms-test.integreat-app.de/muenchen/)

### Steps to Reproduce

1. Go to [LK Emmendingen in our test system](https://cms-test.integreat-app.de/lk-emmendingen/)
2. See on dashboard there are 160 pages with insufficient HIX value
3. Click on link
4. See there are no pages with insufficient HIX value

### Expected Behavior

The both numbers should be the same

### Actual Behavior

They are not

### Additional Information
This issue is related to #2505 and #2501

MizukiTemma:
One possibe cause is probably that the `pages_under_hix_threshold` will not be updated after a page (and therefore its HIX value) was changed.

Issue #2501, opened 2023-10-31
In total calculation of "pages waiting for review" on new dashboard incorrect


charludo:
### Describe the Bug

- The "Pages that are waiting for review" link in the new Dashboard seems to be broken. Next to it, a hint "(in total x)" is shown, however when clicking on the link, no page is found (possibly due to the selected filters?)

#### Somewhat related:
- Clicking on "Pages with low HIX scores" scrolls down to the "Overview of the text understandability of the pages" box after loading the "Translation report" page, but should also expand the box in case the user has previously collapsed it.
- (optional) the "Go to Page" button is very prominent compared to the links taking you to all affected pages, possibly leading to confusion. Maybe a second, outlined button "See all" could be shown to the left of "Go to Page"?

### Steps to Reproduce

1. Go to https://cms-test.integreat-app.de/lkaugsburg/
2. See "Pages that are waiting for review (in total 17)"  in the To-Dos
3. Click the link
4. See "No pages found with this filter."

Same goes for example for [Lk Emmendingen](https://cms-test.integreat-app.de/lk-emmendingen/pages/de/?status=REVIEW) which shows 22 pages on the dashboard, but actually only has 2

### Expected Behavior

Number of pages found when clicking the link matches number of pages shown as "x in total"

### Actual Behavior

Mismatch

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

JoeyStk:
> The "Pages that are waiting for review" link in the new Dashboard seems to be broken. Next to it, a hint "(in total x)" is shown, however when clicking on the link, no page is found (possibly due to the selected filters?)

I think it's actually the other way around. The link (incl. the according filters) should be right. Instead the (in total x) is probably the source of the issue

charludo:
> > The "Pages that are waiting for review" link in the new Dashboard seems to be broken. Next to it, a hint "(in total x)" is shown, however when clicking on the link, no page is found (possibly due to the selected filters?)
> 
> I think it's actually the other way around. The link (incl. the according filters) should be right. Instead the (in total x) is probably the source of the issue

Alright, that would make sense as well! 😄

PR #2515, opened 2023-10-31
Fix "in total" in dashboard

MizukiTemma:
### Short description

This PR fixes the calculation of "in total" in the new dashboard.

The cause was that the query was returning the latest ones among the versions saved with low HIX value or as in-review, not the versions with low HIX or in-review among the latest.

### Proposed changes

- Adjust the query to the same style with `get_automatically_saved_pages`, which was already working fine.

### Side effects

- None? (I hope)

### Resolved issues

Fixes: #2501
Fixes: #2507

### DIFF
--- a/integreat_cms/cms/views/dashboard/dashboard_view.py
+++ b/integreat_cms/cms/views/dashboard/dashboard_view.py
@@ -22,6 +22,7 @@ class DashboardView(TemplateView, ChatContextMixin):
 
     #: The template to render (see :class:`~django.views.generic.base.TemplateResponseMixin`)
     template_name = "dashboard/dashboard.html"
+    latest_version_ids = []
 
     def get_context_data(self, **kwargs):
         r"""
@@ -36,6 +37,7 @@ class DashboardView(TemplateView, ChatContextMixin):
         context = super().get_context_data(**kwargs)
         # RSS FEED urls
         language_slug = translation.get_language()
+        self.latest_version_ids = self.get_latest_versions()
         context.update(
             {
                 "current_menu_item": "region_dashboard",
@@ -57,6 +59,25 @@ class DashboardView(TemplateView, ChatContextMixin):
 
         return context
 
+    def get_latest_versions(self):
+        """
+        Collects all the latest page translations of the current region
+
+        :return: all the latest page translations of the current region
+        :rtype: ~django.db.models.query.QuerySet
+
+        """
+        latest_version_ids = (
+            PageTranslation.objects.filter(
+                page__id__in=Subquery(
+                    self.request.region.non_archived_pages.values("pk")
+                ),
+            )
+            .distinct("page__id", "language__id")
+            .values_list("pk", flat=True)
+        )
+        return list(latest_version_ids)
+
     def get_unreviewed_pages_context(self):
         """
         Extend context by info on unreviewed pages
@@ -64,17 +85,10 @@ class DashboardView(TemplateView, ChatContextMixin):
         :return: Dictionary containing the context for unreviewed pages
         :rtype: dict
         """
-        unreviewed_pages = (
-            PageTranslation.objects.filter(
-                status=status.REVIEW,
-                language__slug=self.request.region.default_language.slug,
-                page__id__in=Subquery(
-                    self.request.region.non_archived_pages.values("pk")
-                ),
-            )
-            .order_by("page__id", "language__id", "-version")
-            .distinct("page__id", "language__id")
-            .all()
+        unreviewed_pages = PageTranslation.objects.filter(
+            language__slug=self.request.region.default_language.slug,
+            id__in=self.latest_version_ids,
+            status=status.REVIEW,
         )
 
         return {
@@ -89,20 +103,11 @@ class DashboardView(TemplateView, ChatContextMixin):
         :return: Dictionary containing the context for auto saved pages
         :rtype: dict
         """
-        last_versions = (
-            PageTranslation.objects.filter(
-                language__slug=self.request.region.default_language.slug,
-                page__id__in=Subquery(
-                    self.request.region.non_archived_pages.values("pk")
-                ),
-            )
-            .order_by("page__id", "language__id", "-version")
-            .distinct("page__id", "language__id")
-        )
         automatically_saved_pages = PageTranslation.objects.filter(
-            id__in=Subquery(last_versions.values("pk")),
+            language__slug=self.request.region.default_language.slug,
+            id__in=self.latest_version_ids,
             status=status.AUTO_SAVE,
-        ).all()
+        )
 
         return {
             "automatically_saved_pages": automatically_saved_pages,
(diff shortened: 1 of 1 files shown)


integreat_cms/release_notes/current/unreleased/2501.yml
timobrembeck, line 1: I don't think that this PR needs release notes, since the bugs only existed on the develop branch and were never released – which means that this is not a change in comparison to the last released version.

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -64,18 +64,20 @@ def get_unreviewed_pages_context(self):
         :return: Dictionary containing the context for unreviewed pages
         :rtype: dict
         """
-        unreviewed_pages = (
+        last_versions = (
```
timobrembeck, line 67: This query is now essentially repeated 4 times in the same view, right?
Maybe it would make sense to outsource the query for the latest versions (of all languages) into a separate function, set a class attribute with these ids and then use them in all the subsequent helper functions? We'd just have to move the respective language filters into the follow-up querysets.

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -64,18 +64,20 @@ def get_unreviewed_pages_context(self):
...
                 ),
             )
             .order_by("page__id", "language__id", "-version")
             .distinct("page__id", "language__id")
-            .all()
         )
+        unreviewed_pages = PageTranslation.objects.filter(
+            id__in=Subquery(last_versions.values("pk")),
+            status=status.REVIEW,
+        ).all()
```
timobrembeck, line 80:
```suggestion
        )
```
See https://github.com/digitalfabrik/integreat-cms/pull/2499#discussion_r1377257856

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -151,15 +153,20 @@ def get_low_hix_value_context(self):
...
-                language__slug__in=settings.TEXTLAB_API_LANGUAGES,
-                page__id__in=Subquery(
-                    self.request.region.non_archived_pages.filter(
-                        hix_ignore=False
-                    ).values("pk")
-                ),
+                id__in=Subquery(last_versions.values("pk")),
                 hix_score__lt=settings.HIX_REQUIRED_FOR_MT,
-            ).distinct("page_id", "language_id")
+            ).all()
```
timobrembeck, line 169:
```suggestion
            )
```

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -151,15 +153,20 @@ def get_low_hix_value_context(self):
...
         """
         if settings.TEXTLAB_API_ENABLED and self.request.region.hix_enabled:
+            last_versions = (
+                PageTranslation.objects.filter(
+                    language__slug__in=settings.TEXTLAB_API_LANGUAGES,
+                    page__id__in=Subquery(
+                        self.request.region.non_archived_pages.values("pk")
+                    ),
+                )
+                .order_by("page__id", "language__id", "-version")
```
timobrembeck, line 163:
```suggestion
```
Isn't this the default ordering anyways?

timobrembeck (changes requested):

integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -416,6 +417,9 @@ def post(self, request, *args, **kwargs):
                 # Add the success message
                 page_translation_form.add_success_message(request)
 
+            invalidate_obj(page_form.instance)
+            invalidate_obj(page_translation_form.instance)
```
MizukiTemma, line 421: This is to avoid Errors by cache.

integreat_cms/cms/views/dashboard/dashboard_view.py
MizukiTemma: Thank you for review 👍 The suggestions are applied.

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -15,6 +15,7 @@
 logger = logging.getLogger(__name__)
 
 
+# pylint: disable=attribute-defined-outside-init
```
timobrembeck, line 18:
```suggestion
```
You can set the attribute to an empty list at the class-level.

integreat_cms/cms/views/pages/page_form_view.py
timobrembeck: What errors exactly? The form instances should already be invalidated when the forms are saved. If that doesn't work, we have a bigger problem.

integreat_cms/cms/views/dashboard/dashboard_view.py
timobrembeck: Great! Yes, that's how I imagined it :+1:

integreat_cms/cms/views/pages/page_form_view.py
MizukiTemma: I saw something like `Key = (...) already exists` but cannot reproduce it 🤔 I'll remove those lines.

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -177,19 +179,9 @@ def get_outdated_pages_context(self):
...
-            .order_by("page__id", "language__id", "-version", "last_updated")
-            .distinct("page__id", "language__id")
-        )
-
         outdated_pages = PageTranslation.objects.filter(
-            id__in=Subquery(last_versions.values("pk")),
+            language__slug=self.request.region.default_language.slug,
+            id__in=self.latest_version_ids,
             last_updated__lte=OUTDATED_THRESHOLD_DATE.date(),
         )
```
timobrembeck, line 186:
```suggestion
        ).order_by("last_updated")
```

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -57,24 +59,36 @@ def get_context_data(self, **kwargs):
...
-                status=status.REVIEW,
-                language__slug=self.request.region.default_language.slug,
                 page__id__in=Subquery(
                     self.request.region.non_archived_pages.values("pk")
                 ),
             )
-            .order_by("page__id", "language__id", "-version")
             .distinct("page__id", "language__id")
-            .all()
+            .values("pk")
```
timobrembeck, line 77:
```suggestion
            .values_list("pk", flat=True)
```
I'm sorry for the many additional comments :see_no_evil: 

Apparently, `values()` still keeps the queryset lazy evaluated, which means its result will not be cached, but evaluated again in all the subqueries:
Screenshot 2023-11-01 at 19-54-52 Integreat Redaktionssystem


So we'd have to convert it to an actual list to make use of the ids:
Screenshot 2023-11-01 at 19-52-48 Integreat Redaktionssystem

Not sure how big the impact really is :thinking: 
In the end, this will actually execute one query more, but I assume the computation time will still be smaller, since the result is reused in 4 other queries...

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -57,24 +59,36 @@ def get_context_data(self, **kwargs):
...
+        )
+        return latest_version_ids
```
timobrembeck, line 79:
```suggestion
        return list(latest_version_ids)
```

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -22,6 +22,7 @@ class DashboardView(TemplateView, ChatContextMixin):
 
     #: The template to render (see :class:`~django.views.generic.base.TemplateResponseMixin`)
     template_name = "dashboard/dashboard.html"
+    latest_version_ids = PageTranslation.objects.none()
```
timobrembeck, line 25:
```suggestion
    latest_version_ids = []
```

MizukiTemma:
@timobrembeck Thank you so much 😃

timobrembeck (approved):
Great, thanks! :tada:
No further objections from my side!

JoeyStk (approved):
Amazing! Thank you!

Issue #2522, opened 2023-11-03
Fix line break in new dashboard in sections with checkmark icon


JoeyStk:
### Describe the Bug

When we have a success message (and therefore the checkmark icon) our new dashboard we have some strange line breaks. This is due to the width of the icon.

### Steps to Reproduce

1. Go to a region in our test system that has a success message on the dashboard, [for example Stadt Dachau](https://cms-test.integreat-app.de/stadtdachau/).
2. See strange word wrapping in the first row (or see screenshot below)

### Expected Behavior

Regular line break

### Actual Behavior

Strange line break

### Additional Information

![grafik](https://github.com/digitalfabrik/integreat-cms/assets/72705147/1a3a9f05-2eb2-4357-816c-ed6921348376)

This issue was wished in this chat: https://chat.tuerantuer.org/digitalfabrik/pl/kkasjb7hg7bbuqnuf3qmrpbmoy

PR #2526, opened 2023-11-03
Fix line break in dashboard sections

JoeyStk:
### Short description

This PR aims to fix the line break in dashboard sections. It's caused due to the width of the checkmark icon.

### Proposed changes

- Add basis class to the rows

### Side effects

- I wasn't able to test this on a screen larger than 14". If changes are needed to be made, someone else would need to take over from me :/
- There are some sizes (for example 1280px where the buttons don't look great. But I'm not sure how to improve it with flex. Maybe it would be better to switch to grid?

### Resolved issues

Fixes: #2522 

### DIFF
--- a/integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_automatically_saved_pages_row.html
+++ b/integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_automatically_saved_pages_row.html
@@ -32,6 +32,6 @@
             {% translate "Go to page" %}
         </a>
     {% else %}
-        <i class="mr-12" icon-name="check-circle-2"></i>
+        <i class="h-8 w-8 text-green-500" icon-name="check-circle-2"></i>
     {% endif %}
 {% endblock todo_dashboard_button_link %}
--- a/integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_broken_links_row.html
+++ b/integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_broken_links_row.html
@@ -32,6 +32,6 @@
             {% translate "Go to link" %}
         </a>
     {% else %}
-        <i class="mr-12" icon-name="check-circle-2"></i>
+        <i class="h-8 w-8 text-green-500" icon-name="check-circle-2"></i>
     {% endif %}
 {% endblock todo_dashboard_button_link %}
--- a/integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_low_hix_row.html
+++ b/integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_low_hix_row.html
@@ -32,6 +32,6 @@
             {% translate "Go to page" %}
         </a>
     {% else %}
-        <i class="mr-12" icon-name="check-circle-2"></i>
+        <i class="h-8 w-8 text-green-500" icon-name="check-circle-2"></i>
     {% endif %}
 {% endblock todo_dashboard_button_link %}
--- a/integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_outdated_pages_row.html
+++ b/integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_outdated_pages_row.html
@@ -32,6 +32,6 @@
             {% translate "Go to page" %}
         </a>
     {% else %}
-        <i class="mr-12" icon-name="check-circle-2"></i>
+        <i class="h-8 w-8 text-green-500" icon-name="check-circle-2"></i>
     {% endif %}
 {% endblock todo_dashboard_button_link %}
--- a/integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_unread_feedback_row.html
+++ b/integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_unread_feedback_row.html
@@ -32,6 +32,6 @@
             {% translate "Go to page" %}
         </a>
     {% else %}
-        <i class="mr-12" icon-name="check-circle-2"></i>
+        <i class="h-8 w-8 text-green-500" icon-name="check-circle-2"></i>
     {% endif %}
 {% endblock todo_dashboard_button_link %}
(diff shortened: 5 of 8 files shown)


integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_automatically_saved_pages_row.html
```diff
@@ -27,7 +27,7 @@
 {% endblock todo_dashboard_description %}
 {% block todo_dashboard_button_link %}
     {% if automatically_saved_pages %}
-        <a class="btn !rounded-full"
+        <a class="btn !rounded-full ml-2"
```
timobrembeck, line 30:
```suggestion
        <a class="btn !rounded-full"
```

integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_broken_links_row.html
```diff
@@ -27,7 +27,7 @@
 {% endblock todo_dashboard_description %}
 {% block todo_dashboard_button_link %}
     {% if broken_links %}
-        <a class="btn !rounded-full"
+        <a class="btn !rounded-full ml-2"
```
timobrembeck, line 30:
```suggestion
        <a class="btn !rounded-full"
```

integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_low_hix_row.html
```diff
@@ -27,7 +27,7 @@
 {% endblock todo_dashboard_description %}
 {% block todo_dashboard_button_link %}
     {% if pages_under_hix_threshold %}
-        <a class="btn !rounded-full"
+        <a class="btn !rounded-full ml-2"
```
timobrembeck, line 30:
```suggestion
        <a class="btn !rounded-full"
```

integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_outdated_pages_row.html
```diff
@@ -27,7 +27,7 @@
 {% endblock todo_dashboard_description %}
 {% block todo_dashboard_button_link %}
     {% if outdated_pages %}
-        <a class="btn !rounded-full"
+        <a class="btn !rounded-full ml-2"
```
timobrembeck, line 30:
```suggestion
        <a class="btn !rounded-full"
```

integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_unread_feedback_row.html
```diff
@@ -28,7 +28,7 @@
 {% block todo_dashboard_button_link %}
     {% if unread_feedback %}
         <a href="{% url 'region_feedback' region_slug=request.region.slug %}"
-           class="btn !rounded-full">
+           class="btn !rounded-full ml-2">
```
timobrembeck, line 31:
```suggestion
           class="btn !rounded-full">
```

integreat_cms/cms/templates/dashboard/todo_dashboard_rows/_unreviewed_pages_row.html
```diff
@@ -27,7 +27,7 @@
 {% endblock todo_dashboard_description %}
 {% block todo_dashboard_button_link %}
     {% if unreviewed_pages %}
-        <a class="btn !rounded-full"
+        <a class="btn !rounded-full ml-2"
```
timobrembeck, line 30:
```suggestion
        <a class="btn !rounded-full"
```

timobrembeck (changes requested):

JoeyStk:
Oh yeah, much better! Thank you!
However, I'm still not 100% satisfied with this view. 

- Do you think it's okay if I change the margin to ml-6 so it matches the margin between the icon on the left side and the text?
- And secondly, the checkmark icon is not centered (s. screenshot). One idea could be to adjust the margin from the right from mr-12 to mr-9 in all sub-templates. But it's still not exact and also not dynamic in case we change the text inside the button. My initial idea was to create a div that matches the width of the button and then center the icon within that width, but I couldn't make it work - it only made it worse. Do you have an idea how to fix this? My only other idea would have been to switch to grid but I'm not sure if that's worth the effort :thinking: 

Thank you a lot in advance!

timobrembeck:
> Do you think it's okay if I change the margin to ml-6 so it matches the margin between the icon on the left side and the text?

Yes, good idea! :+1:

> And secondly, the checkmark icon is not centered (s. screenshot). One idea could be to adjust the margin from the right from mr-12 to mr-9 in all sub-templates. But it's still not exact and also not dynamic in case we change the text inside the button. My initial idea was to create a div that matches the width of the button and then center the icon within that width, but I couldn't make it work - it only made it worse. Do you have an idea how to fix this? My only other idea would have been to switch to grid but I'm not sure if that's worth the effort 🤔

Ah, ok yeah I didn't think about this problem. If all buttons are aligned to the right, it's a bit weird if the check mark floats somewhere near the center of the buttons (but not all buttons have the same width, so it's never going to be exact...)

What do you think about using `basis-full` on the left column and adding `justify-center` and `w-36` to the right column? To me, that looks quite clean and does not waste screen space...

I would maybe also increase the size of the icons a bit (e.g. `h-8 w-8`) and make them green (`text-green-500`) as suggested in the [initial design](https://www.figma.com/file/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?type=design&node-id=533%3A3489&mode=dev)?

timobrembeck (approved):
Thanks, I think this looks cool! :+1:

MizukiTemma (changes requested):
Looks good 🙂 Thank you.

There is only one very mean request: put `, ` in the translation "gespeichert. Bitte vergessen Sie nicht die Seite zu veröffentlichen, sobald "

JoeyStk:
> There is only one very mean request: put , in the translation "gespeichert. Bitte vergessen Sie nicht die Seite zu veröffentlichen, sobald "

Oh, very good catch! Thank you so much :)

MizukiTemma (approved):
👍

Issue #2495, opened 2023-10-30
`TypeError: expected string or bytes-like object` in nominatim API client


timobrembeck:
### Steps to Reproduce

1. Go to https://admin.integreat-app.de/testumgebung-reg/pois/de/new/
2. Input something ?
3. Submit form

### Expected Behavior

A POI should be created and the coordinates should correctly be retrieved from the nominatim api

### Actual Behavior

An internal server error occurs

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
Oct 30 09:30:52 ERROR   django.request - 500 Internal Server Error: /testumgebung-reg/pois/de/new/
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/views/pois/poi_form_view.py", line 166, in post
    if not poi_form.is_valid() or not poi_translation_form.is_valid():
[...]

PR #2529, opened 2023-11-05
Fix error in location coordinates retrieving

MizukiTemma:
### Short description

This PR fixes `TypeError: expected string or bytes-like object`, which occurs in an edge case where only spaces are given as address and house number.

### Proposed changes

- Check whether the input for "addess and house number" contains any valid letter

### Side effects

- There may be other edge cases 🤔 

### Resolved issues

Fixes: #2495 

### DIFF
--- a/integreat_cms/nominatim_api/nominatim_api_client.py
+++ b/integreat_cms/nominatim_api/nominatim_api_client.py
@@ -71,7 +71,7 @@ class NominatimApiClient:
             raise RuntimeError(
                 "You can either specify query_str or pass additional keyword arguments, not both."
             )
-        if "street" in query_dict:
+        if street := query_dict.get("street"):
             # This expression matches a number optionally followed by a whitespace and one character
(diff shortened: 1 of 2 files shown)


integreat_cms/nominatim_api/nominatim_api_client.py
```diff
@@ -71,7 +71,7 @@ def search(
             raise RuntimeError(
                 "You can either specify query_str or pass additional keyword arguments, not both."
             )
-        if "street" in query_dict:
+        if "street" in query_dict and query_dict["street"]:
```
timobrembeck, line 74:
```suggestion
        if street := query_dict.get("street"):
```
Maybe like this and then replace `query_dict["street"]` with `street` in line 83?

timobrembeck (approved):
Thanks! :pray:

david-venhoff (approved):

Issue #1878, opened 2022-11-18
Auto create/update link texts to Integreat content


svenseeberg:
### Motivation
With the introduction of our location feature, we're more and more often referencing internal content. However, it is still a relatively high effort to create a new new link, for example to a location. First, a text has to be created. Then the text has to be selected and the link has to be inserted. If the title of the location changes in the structured database, the link description is not updated.

### Proposed Solution
When inserting a link without having a text selected, automatically create a useful text from the link target. This is already happening for page titles. If it is a location, it should insert the title and address, including icons.

Also, and that is the most interesting part, add an additional HTML attribute that marks the link text as being automatically created. The attribute could be invented, for example `integreat-auto-content=1`. We can then use the Django linkcheck database to update the link text whenever we update the location, page title, etc. That means when updating a location, for example, we retrieve all pages that link to this location. We then iterate over all a tags with links and our attribute and then update the link text. This could probably be a part of our abstract content translation model save method.

It should be made obvious to the user how to use this. When inserting a link, we already have a dialog for selecting an internal content to link to. This dialog needs something like a check box "automatically update link description". This check box should be checked when having no text selected. When having a text selected, the checkbox should be unchecked.

### Alternatives
None. It's a really cool killer feature :rocket: 

### Additional Context
It would be cool if translations are fixed automatically. That means if the location is translated into arabic, but the link in a page is still to the german location, this should be fixed automatically.

### Design Requirements

We should probably get some good drafts for the dialog boxes. And we also have to improve the "change link" box.

Also we need nice defaults for rendering locations and maybe also events.

david-venhoff:
> If it is a location, it should insert the title and address, including icons.

Shouldn't stuff like this ideally be implemented in the frontend? That way, they could decide how to display all the relevent poi information and we would only have to provide some metadata with the link

timobrembeck:
> Shouldn't stuff like this ideally be implemented in the frontend? That way, they could decide how to display all the relevent poi information and we would only have to provide some metadata with the link

I did not find our previous discussion about this topic, but I think I remember we rejected the idea of having something like an intermediary html syntax that needs to be rendered differently by the app.

Issue #2516, opened 2023-10-31
META: Large synthetic dataset for performance evaluation


PeterNerlich:
This issue is about **research** into improving the development workflow when interested in performance bottlenecks. While we could just create a copy of the live system for local experimentation (the same as we do with the test system every so often), it might contain personal information which we rather don't want to even be able to obtain as developers.

- **Devise a method of generating reasonably authentic data** in comparison with the live system, namely a similar number of regions with similar content.
- **Verify that** a dev environment with that data **behaves similarly** to the live environment in performance (mind resources available, which might vary greatly between the machines used for development; while e.g. the redis cache is probably only to be mentioned to the developer as a potentially deciding if not active).

This should be separate from the existing `test_data.json` fixture, as the small dataset is highly preferable during quick iterations on a feature, except when performance with large data is the focus. The developer should be able to switch between them with relative ease.

timobrembeck:
Just for completeness, I want to mention
```
./tools/integreat-cms-cli duplicate_pages augsburg
```
which can be used to generate a lot of pages, however it does not cover specific edge cases which are not reflected in the original test data. So one solution could be to create a more diverse baseline test data, which hopefully would result in a more realistic dataset if the duplication algorithm is executed a few times (~1k pages for large regions is realistic).

david-venhoff:
>  however it does not cover specific edge cases which are not reflected in the original test data

One edge case example would be #2530, where performance testing requires lots of different links, which cannot be created using the `duplicate_pages` tool

david-venhoff:
@timobrembeck when using dumpdata on the production database, the resulting file is about 8GB big. I assume this is too much to include in this repo? :smile: 
But maybe it is fine to just use database dumps if we need to test for performance?

PeterNerlich:
For eliminating sensitive data from the dump – are we looking at anything not detectable in a fixed format? Because otherwise we could just detect names, telephone numbers, email addresses etc. and replace them with generic ones. Maybe even build a dictionary of which ones we previously replaced and re-use the generated one whenever we encounter the same detail again, if we feel fancy.

timobrembeck:
> @timobrembeck when using dumpdata on the production database, the resulting file is about 8GB big. I assume this is too much to include in this repo? 😄 But maybe it is fine to just use database dumps if we need to test for performance?
> 
> Edit: When compressed it is about 740MB

[...]

Issue #1627, opened 2022-08-09
Add broken link checker for imprint


timobrembeck:
### Motivation

Apart form pages, events and pois, also the imprint can contain broken links that should be fixed when found.

### Proposed Solution

Add the same mechanism for the link checker for imprint translations

### Additional Context

Pay extra attention to the deletion of old links, e.g. in the imprint translation form and the imprint revision view outdated links need to be deleted (see e.g. #1604)

PR #2530, opened 2023-11-06
Automatically update internal links

david-venhoff:
### Short description

This pr adds a feature to automatically update internal links when the content object they link to gets updated.
This also works across regions.
Right now no address is shown as link text for links to pois as requested in #1878, is that still wanted (And if so, do you have a preferred format @svenseeberg)?
Also I still need to test if the performance is okay. Ideally I would like to do that on the test system, as the data are more realistic there.

### Proposed changes
 
- When saving a page, fix all links in that page that are outdated
- When saving a page, find all links that point to this page, and for each link update the content translation of that link
- Add the attribute `data-integreat-auto-update` and automatically create a link text for those links if possible
- Add icons before event and poi links

### Side effects

- When a user updates a page and if that causes another page to be updated, the author will be shown as the user who changed the other page, even if the user does not have permission to do so (Should alternatively no user be shown?)

### Resolved issues

Fixes: #1878


seluianova:
> Add icons before event and poi links
Should it be the default icon or a specific location/event icon?

david-venhoff:
> Should it be the default icon or an icon set for the particular location/event ?

Seems like a good idea to me

> Is it okay, that we add this attribute to external links as well?

While technically redundant, they are ignored internally, so imo that is not a problem

seluianova:
> While technically redundant, they are ignored internally, so imo that is not a problem

I'm not insisting here, but perhaps it can be understood to mean that we will update the link text for external links as well

david-venhoff:
I have added the functionality to load custom defined icons, which looks like this:
grafik

I have also adjusted the behavior of the link dialog, so that either the text and url input can be edited, or that the 'link to internal content' input and the checkbox can be edited. Unfortunately, tinymce is very buggy and does not visually display which inputs are enabled and which are disabled. I also tried disable the checkbox on initial load, but that causes another tinymce bug where it will stay visually disabled, even when it gets enabled later…
I feel like the current custom link input dialog is hard to maintain and the tinymce api is very limiting. Maybe we should create an issue to completely refactor it at some point? 

> 4. Is it okay that the link text with an icon will be shown like this?

I think yes, because it accurately represents what the link looks like and users can expand the text field to view the full link text. If you don't agree, would you prefer just  `[image]title` for links with both image and text?

integreat_cms/cms/utils/internal_link_utils.py
```diff
@@ -51,14 +60,14 @@ def update_link_language(
...
 
-        # Update the title if it was previously the translation title or the url
-        if link_text:
-            if source_translation.title.lower() == link_text.strip().lower():
-                link_text = target_translation.title
-            elif current_link.strip() == link_text.strip():
-                link_text = fixed_link
+        # Update the title if it was previously the url, otherwise use the new title
+        link_html = None
+        if len(link) == 0 and link.text and current_url.strip() == link.text.strip():
```
sarahsporck, line 65: What does the len(link) check for? Isn't the link an Element?

integreat_cms/cms/views/utils/search_content_ajax.py
```diff
@@ -44,9 +45,19 @@ def format_object_translation(
     :param typ: The type of this object
     :return: A dictionary with the title, path, url and type of the translation object
     """
+    if isinstance(object_translation.link_title, str):
```
sarahsporck, line 48: What is the difference between `object_translation.link_title` and `object_translation.title`?

integreat_cms/static/src/js/tinymce-plugins/custom_link_input/plugin.js
```diff
@@ -143,6 +144,16 @@ import { getCsrfToken } from "../../utils/csrf-token";
...
+                if (data.completions !== "") {
+                    api.disable("text");
+                    api.disable("url");
+                    api.enable("autoupdate");
+                } else {
+                    api.enable("text");
+                    api.enable("url");
+                    api.disable("autoupdate");
+                }
+
```
sarahsporck, line 156: So...what's this doing? And why? :)

integreat_cms/cms/utils/content_translation_utils.py
```diff
@@ -0,0 +1,170 @@
...
+        # Note that this should not be possible, since links to archived pages get deleted
+        if getattr(outdated_content_translation.foreign_object, "archived", False):
+            continue
+
+        new_content = get_cleaned_content(
+            outdated_content_translation.content,
+            outdated_content_translation.language.slug,
+        )
+        if new_content == outdated_content_translation.content:
+            continue
```
sarahsporck, line 136: What's this check for?

integreat_cms/cms/views/utils/search_content_ajax.py
david-venhoff: `link_title` is the method I defined in `integreat_cms/cms/models/abstract_content_translation.py` which can both return a `str` and an `Element`. I did this because the issue for this pr wants icons displayed next to the links for internal content translations. So in those cases, `link_title` returns the icon as an HTML Element in addition to the normal title

integreat_cms/static/src/js/tinymce-plugins/custom_link_input/plugin.js
david-venhoff: This enables the checkbox to automatically update an internal link only when there is a current completion, and it disables the manual inputs for a link, too in that case. Unfortunately it does not display correctly, due to a bug in tinymce

integreat_cms/cms/utils/content_translation_utils.py
david-venhoff: This is just an optimization so that no new translations get created when the content did not change. For example, this function gets executed when the status of a content translation changes to `PUBLIC`, but its title and content might not have changed and all existing links are still valid. In this case there would be no point in saving new versions with the same content.

seluianova (commented):
> Unfortunately, tinymce is very buggy and does not visually display which inputs are enabled and which are disabled

To be honest I think it was better when the user could control the checkbox regardless of the link type.

Because now you can open a dialog, can _uncheck_ the checkbox, but can't _check_ it back. 
Or, if you edit a link, where the checkbox was unchecked, you can _check_ it. But then you can't _uncheck_ it back!
So it's like - you still can edit a checkbox, but only once 😃
I think it's just more confusing and looks like a bug 🙈

MariaKabanova:
I think it’s an awesome change which adds a lot of value for customers. I’ve tested this pull request, and I found a couple of small things.

The first one is about an icon. Looks like even though the link title and url are auto-updated, the icon of the internal link is not updated automatically. You can follow the next steps to reproduce that: 

1. Open page, then insert the link on another page
2. Go to the linked page and change or delete an icon there
3. Return to the original page, where you can see that icon didn’t change. But if you click on “update”, the icon is changed after all

The second one contains the similar problem but in imprint. After inserting a link to the page in imprint, this link doesn’t automatically update. Steps to reproduce:

1. Open Imprint, then create the link to another page (insert/link)
2. Go to the linked page and change its title
3. Return to the imprint and see that the title didn’t change
4. Click on “update” and see that title has changed

I'm not sure it's worth fixing, please let me know what you think.

MariaKabanova:
@david-venhoff  

Today I've tested again this branch and now everything is great!  It means that  now we need to wait for #1627 is fixed. After that I'll retest imprints. :-)

integreat_cms/cms/models/events/event_translation.py
```diff
@@ -83,6 +85,24 @@ def backend_edit_link(self) -> str:
...
+        This property returns the html that should be used as a title for a link to this translation
+
+        :return: The link content
+        """
+        img = get_icon_html("clock")
+
+        if icon := self.event.icon:
+            if thumbnail := icon.thumbnail_url:
+                img = Element("img")
+                img.set("style", "height:15px;")
```
seluianova, line 100:
```suggestion
                img.set("style", "height:15px; margin-right:5px;")
```

integreat_cms/cms/models/pages/page_translation.py
```diff
@@ -385,6 +386,23 @@ def path(self) -> SafeString:
...
+    def link_title(self) -> Element | str:
+        """
+        This property returns the html that should be used as a title for a link to this translation
+
+        :return: The link content
+        """
+        if icon := self.page.icon:
+            if thumbnail := icon.thumbnail_url:
+                img = Element("img")
+                img.set("style", "height:15px;")
```
seluianova, line 399:
```suggestion
                img.set("style", "height:15px; margin-right:5px;")
```

integreat_cms/cms/models/pois/poi_translation.py
```diff
@@ -94,6 +96,24 @@ def backend_edit_link(self) -> str:
...
+        This property returns the html that should be used as a title for a link to this translation
+
+        :return: The link content
+        """
+        img = get_icon_html("pin")
+
+        if icon := self.poi.icon:
+            if thumbnail := icon.thumbnail_url:
+                img = Element("img")
+                img.set("style", "height:15px;")
```
seluianova, line 111:
```suggestion
                img.set("style", "height:15px; margin-right:5px;")
```

seluianova (approved):
LGTM, thanks!

**UPD**. tooo minor and may be fixed later or not fixed at all, but as a note:
If you update only an icon of the page/event/location, it's not updated automatically (as it works with titles), but only if you click "Update" on the page, where the corresponding link is used

integreat_cms/cms/utils/internal_link_utils.py
seluianova: `# Update the title if it was previously the url `

I haven't looked into it deeply, but it seems to work for pages, but not for events or locations 🤔

integreat_cms/cms/models/events/event_translation.py
```diff
@@ -83,6 +85,24 @@ def backend_edit_link(self) -> str:
             },
         )
 
+    @cached_property
+    def link_title(self) -> Element | str:
```
seluianova, line 89: And I think we could do some refactoring here to reduce code duplication.
The implementation is almost the same for all content types. I think we could basically keep one default method in abstract_content_translation.py and only re-assign the default icon (null for pages, not-null for events and locations)

integreat_cms/cms/models/events/event_translation.py
david-venhoff: I also added some left margin for rtl languages

david-venhoff:
> If you update only an icon of the page/event/location, it's not updated automatically (as it works with titles), but only if you click "Update" on the page, where the corresponding link is used

Oh, right. Unfortunately this is a little bit more complicated to fix, because the icon is not part of the translation forms. I'd prefer to fix this in a follow-up pr, if that is okay.

david-venhoff:
> If you update only an icon of the page/event/location, it's not updated automatically (as it works with titles), but only if you click "Update" on the page, where the corresponding link is used

That should already be implement (in 98b3c47fe5d87aa359e2f2f5fa179b9962ddd8b1)

MizukiTemma (approved):
Cool feature 😻 Thank you so much for tackling this issue!

As #1627 is in the backlog and therfore will not be implemented soon, I suggest to merge this PR already 👍

david-venhoff:
Quick status update: I would like to do some performance testing before merging this pr, so I'll block it on #2516

david-venhoff:
I tested this with production data and, as expected, the performance was really bad. I am currently working on increasing the performance, so that this pr can be merged.

david-venhoff:
@seluianova @MizukiTemma I have added an optimization to reduce the time it takes to save. On the production data saving a page now takes one or two seconds longer than without this pr, which I think is acceptable. Could you take a look at the changes?

MizukiTemma:
> @seluianova @MizukiTemma I have added an optimization to reduce the time it takes to save. On the production data saving a page now takes one or two seconds longer than without this pr, which I think is acceptable. Could you take a look at the changes?

Sorry for having overlooked this request 🙈 Could you rebase and resolve the conflicts? Then I will check this PR again 💪

david-venhoff:
The pr is now rebased onto develop :)

MizukiTemma:
@david-venhoff Thank you for rebasing 👍 
I found only one thins to fix. `Cannot resolve keyword 'slug' into field` occurs when a link is added to the imprint.
(Sorry, this is an extra work due to my recent PR adding Imprint to the link check 🙏 )

david-venhoff:
Great catch!

PR #2532, opened 2023-11-06
Update dependencies

timobrembeck:
### Resolved issues

Fixes:
- https://github.com/digitalfabrik/integreat-cms/security/dependabot/39
- https://github.com/digitalfabrik/integreat-cms/security/dependabot/40
- https://github.com/digitalfabrik/integreat-cms/security/dependabot/41


charludo (approved):
Looks good! QUestion about the `aria-label`s: Are these just added to prevent linter complaints? I'm asking because I'm pretty confident we have a lot more buttons 😄

MizukiTemma (approved):

timobrembeck:
> Looks good! Question about the `aria-label`s: Are these just added to prevent linter complaints? I'm asking because I'm pretty confident we have a lot more buttons 😄

Yes, exactly, there is this new rule [jsx-a11y/control-has-associated-label](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/control-has-associated-label.md#jsx-a11ycontrol-has-associated-label) which only checks jsx components. 
But in the end, this also makes sense for html buttons that only have an icon as content without any text, so maybe we should open a new issue for this?

Issue #2524, opened 2023-11-03
Internal server error when filtering events by end date


JoeyStk:
### Describe the Bug
When filtering events by end date an Internal server error occurs (both on live and test server).

### Steps to Reproduce

1. Go to events
2. Click on show filter
3. Select "custom time range", only fill in the end date and apply filter
4. See Internal server error in both test and live system.

### Expected Behavior

Events are filtered by end date

### Actual Behavior

Internal server error

### Additional Information

<details>
  <summary>Traceback</summary>

   ```
    return bound_method(*args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/views/events/event_list_view.py", line 95, in get
    events, poi, query = event_filter_form.apply(events, region, language_slug)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/forms/events/event_filter_form.py", line 105, in apply
    events = events.filter_upcoming(from_local).filter(end__lte=to_local)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/models/events/event.py", line 40, in filter_upcoming
    return self.filter(
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/db/models/query.py", line 941, in filter
    return self._filter_or_exclude(False, args, kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/db/models/query.py", line 961, in _filter_or_exclude
    clone._filter_or_exclude_inplace(negate, args, kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/db/models/query.py", line 968, in _filter_or_exclude_inplace
    self._query.add_q(Q(*args, **kwargs))
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/db/models/sql/query.py", line 1416, in add_q
    clause, _ = self._add_q(q_object, self.used_aliases)
[...]

PR #2543, opened 2023-11-10
Fix event search

JoeyStk:
### Short description
This PR changes the search query for events and therefore fixes the bug which occurs when filtering events only by end date

### Proposed changes

- Change search query for events- 

### Side effects

- I think none

### Resolved issues

Fixes: #2524 

### DIFF
--- a/integreat_cms/cms/forms/events/event_filter_form.py
+++ b/integreat_cms/cms/forms/events/event_filter_form.py
@@ -98,9 +98,10 @@ class EventFilterForm(CustomFilterForm):
         elif events_time_range.CUSTOM in cleaned_time_range:
             # Filter events for their start and end
             tzinfo = zoneinfo.ZoneInfo(region.timezone)
-            from_local = datetime.combine(
-                self.cleaned_data["date_from"] or date.min, time.min, tzinfo=tzinfo
-            )
+            if date_from := self.cleaned_data["date_from"]:
+                from_local = datetime.combine(date_from, time.min, tzinfo=tzinfo)
+            else:
+                from_local = datetime.min.replace(tzinfo=zoneinfo.ZoneInfo(key="UTC"))
             to_local = datetime.combine(
                 self.cleaned_data["date_to"] or date.max, time.max, tzinfo=tzinfo
             )
--- a/integreat_cms/release_notes/current/unreleased/2524.yml
+++ b/integreat_cms/release_notes/current/unreleased/2524.yml
@@ -1,0 +1,2 @@
+en: Fix internal server error when filtering events by end date
+de: Behebe den internen Serverfehler beim Filtern von Ereignissen nach Enddatum
--- a/tests/cms/views/test_public_view_status_code.py
+++ b/tests/cms/views/test_public_view_status_code.py
@@ -12,14 +12,16 @@ from .view_config import PARAMETRIZED_PUBLIC_VIEWS
 if TYPE_CHECKING:
     from _pytest.logging import LogCaptureFixture
 
+    from .view_config import PostData, ViewNameStr
+
 
 @pytest.mark.django_db
 @pytest.mark.parametrize("view_name,post_data", PARAMETRIZED_PUBLIC_VIEWS)
 def test_public_view_status_code(
     load_test_data: None,
     caplog: LogCaptureFixture,
-    view_name: str,
-    post_data: dict[str, str],
+    view_name: ViewNameStr,
+    post_data: PostData,
 ) -> None:
     """
     This test checks whether the given view return the correct status code for anonymous users
--- a/tests/cms/views/test_view_redirect.py
+++ b/tests/cms/views/test_view_redirect.py
@@ -13,16 +13,18 @@ if TYPE_CHECKING:
 
     from _pytest.logging import LogCaptureFixture
 
+    from .view_config import RedirectTarget, Roles, ViewKwargs, ViewNameStr
+
 
 @pytest.mark.django_db
 @pytest.mark.parametrize("view_name,kwargs,roles,target", PARAMETRIZED_REDIRECT_VIEWS)
 def test_view_redirect(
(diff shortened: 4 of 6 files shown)


david-venhoff (commented):
Thanks! I am not so sure about calling `date()` on `from_date`, since the documentation of `filter_upcoming` specifies that it should already be a `datetime.date`. I'd argue that the error comes from code that calls `filter_upcoming` with the wrong type.
As far as i can see, this seems to be the case in `event_filter_form.py:105`

timobrembeck (changes requested):
I don't completely remember, but I think we went for comparing datetimes here instead of dates to account for timezone differences, so for Germany it doesn't make a huge difference, since we have a maximum of 2h difference to UTC, but let's say we deploy an instance in another timezone, it might be weird when the results are +-1 day off since the event dates are stored in UTC in the database, but the dates that are shown in the UI are the local dates.

The underlying problem seems to be that a minimal date and time in the Europe/Berlin timezone do not have a corresponding value in the UTC timezone (since it would be a negative time):
```
from zoneinfo import ZoneInfo
from datetime import datetime, date, time

utc = ZoneInfo("UTC")
berlin = ZoneInfo("Europe/Berlin")

datetime.combine(date.min, time.min, tzinfo=berlin).astimezone(utc)

> OverflowError: date value out of range

```

So I'd also prefer to fix that instead of working around it by only comparing dates.
I guess we could directly define the minimum start date as
```
datetime.min.replace(tzinfo=zoneinfo.ZoneInfo(key="UTC"))
```
to work around this problem? :thinking:

timobrembeck (commented):
Apart from that, could you also add a few test cases for the event list with different filters applied? Just to make sure this does not happen again in the future?

JoeyStk:
@MizukiTemma You can push the tests for the event filter now :)

JoeyStk:
> Apart from that, could you also add a few test cases for the event list with different filters applied? Just to make sure this does not happen again in the future?

@MizukiTemma was already working on tests for the event filter view and will push them on this branch

> datetime.min.replace(tzinfo=zoneinfo.ZoneInfo(key="UTC"))

Applied!

integreat_cms/cms/models/events/event.py
```diff
@@ -44,16 +45,18 @@ def filter_upcoming(self, from_date: datetime | None = None) -> EventQuerySet:
         :param from_date: The date which should be used for filtering, defaults to the current date
         :return: The Queryset of events after the given date
         """
-        from_date = from_date or timezone.now().date()
+        from_date = (from_date or timezone.now()).min.replace(
+            tzinfo=ZoneInfo(key="UTC")
+        )
```
timobrembeck, line 50:
```suggestion
        from_date = from_date or timezone.now()
```
Hmm, this was not exactly what I meant, I meant to call the `.replace(tzinfo=ZoneInfo(key="UTC"))` [here](https://github.com/digitalfabrik/integreat-cms/blob/6e60e28067c1f525019e12e6e8671060e9784c46/integreat_cms/cms/forms/events/event_filter_form.py#L102) where the actual `OverflowError` is happening. It's only needed for the fallback value of `date.min` in combination with `time.min`. The other times are fine, Django is able to automatically convert between timezones, as long as the given datetime is timezone-aware (which `timezone.now()` is).

integreat_cms/cms/models/events/event.py
```diff
@@ -44,16 +45,18 @@ def filter_upcoming(self, from_date: datetime | None = None) -> EventQuerySet:
...
         return self.filter(
-            Q(end__gte=from_date)
+            Q(end__date__gte=from_date)
```
timobrembeck, line 52:
```suggestion
            Q(end__gte=from_date)
```
`end` is a datetime,`from_date` as well, so no need for `__date`.

timobrembeck (changes requested):

tests/cms/test_event_filters.py
```diff
@@ -0,0 +1,28 @@
...
+
+from urllib.parse import urlencode
+
+import pytest
+from django.test.client import Client
+from django.urls import reverse
+
+
+@pytest.mark.django_db
+def test_event_filters(load_test_data: None, admin_client: Client) -> None:
```
MizukiTemma, line 11: I didn't find any way to test the event filter by `view_config.py`.
If this looks like the right way to go, I'll add other test cases and the documentation.

timobrembeck: Maybe it's stupid idea, but a quick way to add get parameters to the existing view tests would be something like converting the view name to an optional tuple, e.g. instead of:
https://github.com/digitalfabrik/integreat-cms/blob/6e60e28067c1f525019e12e6e8671060e9784c46/tests/cms/views/view_config.py#L266

we write
```
(
    (
        "events",
        parse.urlencode({
            "events_time_range": "CUSTOM",
            "date_from": "2023-01-01",
            "date_to": "2030-12-31",
            "poi_id": -1,
            "all_day": 1,
            "recurring": 2,
        })
    ),
    ROLES
)
```

And then in the test utils, we can change:
https://github.com/digitalfabrik/integreat-cms/blob/6e60e28067c1f525019e12e6e8671060e9784c46/tests/cms/views/utils.py#L37

to:
```
if isinstance(view_name, tuple):
    url = reverse(view_name[0], kwargs=kwargs) + "?" + view_name[1]
else:
    url = reverse(view_name, kwargs=kwargs)
```

But I didn't test this out. This also makes the view config even more complex and also requires changes to the type annotations.

timobrembeck (approved):
Thanks for adding the tests! :+1:

I fixed the remaining problems and the type hints, I think this is good to go now.

MizukiTemma:
@timobrembeck Thank you so much 🙏

ulliholtgrave (approved):
Nice, looks pretty good now :+1:

Issue #2343, opened 2023-07-11
Perform link replacement in background for cloned region


MizukiTemma:
### Motivation

As mentioned in #2281 cloning region encounters often time out. This is a follow up issue to introduce link replacement in background.

### Proposed Solution

perform the link replacement asynchronously in the background

### Alternatives

Leave link replacement a manual work

PR #2344, opened 2023-07-11
Deactivate link replacement in region cloning

MizukiTemma:
### Short description

As a temporary solution this PR removes link replacement from region cloning process.

### Proposed changes

- Remove link replacement from region cloning
- Adjust test

### Side effects

- Now region cloning should function without time out, but link replacement should be done manually with the newly introduced link replacement GUI in the broken link section.
- Background link replacement should follow (#2343)

### Resolved issues

Fixes: #2281 

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #2558, opened 2023-11-15
Perform link replacement as background task

sarahsporck:
### Short description
Reverts #2344.

### Proposed changes

- ~add a global task queue (implemetation almost identical to [django-linkcheck task queue](https://github.com/DjangoAdminHackers/django-linkcheck/blob/3d6188e5053469a36fedd307b75ecb02b15d0087/linkcheck/listeners.py#L19))~ Instead just create a thread where the links are replaced
- replace links when cloning content
- readd checks in the tests

### Side effects

- Technically there should be no multithread issue here, as django-linkcheck creates a lock when doing sth. like this, but I also did not explicitly test for side effects.

### Resolved issues

Fixes: #2343 

### DIFF
--- a/integreat_cms/cms/forms/regions/region_form.py
+++ b/integreat_cms/cms/forms/regions/region_form.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import json
 import logging
+import threading
 import time
 from copy import deepcopy
 from typing import TYPE_CHECKING
@@ -15,6 +16,8 @@ from django.utils.translation import gettext_lazy as _
 from django.utils.translation import override
 from linkcheck.listeners import disable_listeners, tasks_queue
 from linkcheck.models import Link
+
+from integreat_cms.cms.utils.linkcheck_utils import replace_links
 
 from ....gvz_api.utils import GvzRegion
 from ....matomo_api.matomo_api_client import MatomoException
@@ -266,8 +269,9 @@ class RegionForm(CustomModelForm):
                     duplicate_imprint(source_region, region)
             # Duplicate media content
             duplicate_media(source_region, region)
-            # Create links for the most recent versions of all translations manually
-            find_links(region)
+
+            # Create links for the most recent versions of all translations manually and replace internal links
+            create_and_replace_links_async(source_region, region)
 
         return region
 
@@ -813,6 +817,29 @@ def duplicate_media(source_region: Region, target_region: Region) -> None:
     # TODO: implement duplication of all media files
 
 
+def create_and_replace_links_async(source_region: Region, region: Region) -> None:
+    """
+    Create all links for the latest versions of the region's page translations, then replace all links in the content.
+    This is run as a background task.
+
+
+    :param source_region: The region with the slug of the to be replaced links
+    :type source_region: ~integreat_cms.cms.models.regions.region.Region
+
+    :param region: The region in which the links should be replaced and created
+    :type region: ~integreat_cms.cms.models.regions.region.Region
+    """
+
+    def create_and_update_links() -> None:
+        find_links(region)
+        replace_internal_links(source_region, region)
+
+    t = threading.Thread(target=create_and_update_links, daemon=True)
+    t.start()
+    if not settings.BACKGROUND_TASKS_ENABLED:
+        t.join()
+
+
 def find_links(region: Region) -> None:
     """
     Find all link objects in the latest versions of the region's page translations
@@ -834,3 +861,18 @@ def find_links(region: Region) -> None:
     logger.debug(
         "Found links: %r", Link.objects.filter(Q(page_translation__page__region=region))
     )
+
+
+def replace_internal_links(source_region: Region, region: Region) -> None:
+    """
+    Replace all internal link objects with the latest versions of the region's page translations
+
+    :param source_region: The region with the slug of the to be replaced links
+    :type source_region: ~integreat_cms.cms.models.regions.region.Region
+
+    :param region: The region in which the links should be replaced
+    :type region: ~integreat_cms.cms.models.regions.region.Region
+    """
+    old_link = f"{settings.WEBAPP_URL}/{source_region.slug}/"
+    new_link = f"{settings.WEBAPP_URL}/{region.slug}/"
+    replace_links(old_link, new_link, region=region, link_types=["internal"])
--- a/integreat_cms/core/circleci_settings.py
+++ b/integreat_cms/core/circleci_settings.py
@@ -30,3 +30,5 @@ LOG_LEVEL = "DEBUG"
 LINKCHECK_DISABLE_LISTENERS = True
 #: Enable logging of all entries from the messages framework
 MESSAGE_LOGGING_ENABLED = True
+# Disable background tasks during testing
+BACKGROUND_TASKS_ENABLED = False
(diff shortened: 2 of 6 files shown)


integreat_cms/cms/utils/task_queue.py
```diff
@@ -0,0 +1,42 @@
+import logging
+from queue import Empty, LifoQueue
+import threading
+
+logger = logging.getLogger(__name__)
+
+class TaskQueue:
```
timobrembeck, line 7: Hmm, I'm not sure whether this use case really justifies a queue – after all, this link replacement function is only executed once per region cloning process, right? So I think it will be an extreme edge case that two regions are cloned at the same time and even if this happens, putting those two jobs into a queue wouldn't make much of a difference, I mean probably the normal CPU scheduler is capable of handling two background threads, right? :thinking:

Maybe I'm missing something, but I think simply this should do it:
```
    t = threading.Thread(
        target=replace_links,
        kwargs={
            "search": old_link,
            "replace": new_link,
            "region": region,
            "link_types": ["internal"],
        },
        daemon=True,
        name="replace_links",
    )
    t.start()
```

Or, if this does not work for some reason, I think we don't need to re-define the queue in order to use it – we could just import the linkcheck queue and add tasks to it.

timobrembeck (changes requested):
Thanks! :+1:

Especially with the current problems when cloning regions, I think this would be really helpful!

However, I'm not yet convinced that a queue is necessary here, I'd prefer a simple background thread.


> Technically there should be no multithread issue here, as django-linkcheck creates a lock when doing sth. like this, but I also did not explicitly test for side effects.

Unfortunately, the linkcheck lock is only thread-safe, not process-safe, see https://github.com/DjangoAdminHackers/django-linkcheck/issues/168. In our production environment, we start [wsgi with 8 processes](https://git.tuerantuer.org/DF/salt/src/commit/45bd245ee42928463760d6e9e8fec1728c6a7fc7/pillars/django/integreat-prod.sls#L4), meaning we would have 8 concurrent queues, so if there are race conditions between the background jobs, using this queue wouldn't prevent them.

integreat_cms/cms/utils/task_queue.py
sarahsporck: Jep that's actually what I started with.
However, I realized, that we need to have a way to check whether the thread has been finished in the tests. 
I figured that there are 2 ways to do this:
1. Add a check `if TESTING: t.join()` to run the thread synchronously in the tests.
2. Add a helper class

I chose the second option as I thought it could be helpful in the future to have a manager for background tasks in general. If that's not relevant rn I'm open to remove this.

timobrembeck: Ah ok, didn't realize this.
Hmm, I'm undecided on this, but leaning a bit towards the first solution. Maybe just a new setting `ENABLE_BACKGROUND_TASKS` or similar which is `True` by default but can be disabled during tests or when debugging problems on the production server?
I'm not yet sure whether we can reuse the queue – it depends a bit on our future use cases, I guess if we implement more complex background tasks, we would be better off relying on a queue worker that is independent from the wsgi processes that handle the webserver requests, e.g. by using Celery or similar. Also, we did not yet find a process-safe locking mechanism which could also influence this decision. So as long as we don't know in which direction we're heading, I would vote for removing the queue, but I'm also fine with keeping it for the moment.

sarahsporck: Sure. Fine with me. Especially as there is no actual use rn.

integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -252,6 +255,8 @@ def save(self, commit=True):
             # Create links for the most recent versions of all translations manually
             find_links(region)
```
timobrembeck, line 256: Maybe it would make sense to execute this in the background as well?

tools/test.sh
```diff
@@ -22,6 +22,9 @@ export INTEGREAT_CMS_DEEPL_AUTH_KEY="dummy"
 # Disable linkcheck listeners during testing
 export INTEGREAT_CMS_LINKCHECK_DISABLE_LISTENERS=1
 
+# Disable background tasks during testing
+export INTEGREAT_CMS_BACKGROUND_TASKS_ENABLED=0
```
timobrembeck, line 26: This also needs to be added to the [circleci_settings.py](https://github.com/digitalfabrik/integreat-cms/blob/2c560214ffb3a8af52435dec32192d7dca730a21/integreat_cms/core/circleci_settings.py) (the `test.sh` is only used for local testing)

integreat_cms/cms/forms/regions/region_form.py
sarahsporck: Let's see what the worst case scenario could be:
Source region A has a page /A/de/willkommen.
Further there is a target region B.

We start the thread find_links.
We start the thread replace_links.

find_links is creating links in the background, however /A/de/willkommen has not yet processed. Then /A/de/willkommen occurs in the content and replace_links updates the internal link to /B/de/willkommen. Later find_links creates the link /B/de/willkommen.
Everyone is happy =).

I'll also thread the find_links functionallity.

sarahsporck: Never mind. When running find_links in a thread, the links are not updated. Could it be that the save in find links overwrites the replaced links?

timobrembeck: Ah yes, `replace_links()` depends on `find_links()`. So maybe an intermediate function that runs them both synchronously, and just this intermediary function is started as one background thread?

integreat_cms/core/circleci_settings.py
```diff
@@ -28,6 +28,8 @@
 LINKCHECK_DISABLE_LISTENERS = True
 #: Enable logging of all entries from the messages framework
 MESSAGE_LOGGING_ENABLED = True
+# Disable background tasks during testing
+INTEGREAT_CMS_BACKGROUND_TASKS_ENABLED = True
```
timobrembeck, line 32:
```suggestion
INTEGREAT_CMS_BACKGROUND_TASKS_ENABLED = False
```

timobrembeck: Not only the variable name, but also the value needs to be changed :see_no_evil:

sarahsporck: 😅 wow. sometimes you actually need to mind every detail in a review. sorry for the delay and thanks for the awesome review 🎉

integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -249,8 +252,9 @@ def save(self, commit=True):
                     duplicate_imprint(source_region, region)
             # Duplicate media content
             duplicate_media(source_region, region)
-            # Create links for the most recent versions of all translations manually
-            find_links(region)
+
+            # Create linksfor the most recent versions of all translations manually and replace internal links
```
timobrembeck, line 256:
```suggestion
            # Create links for the most recent versions of all translations manually and replace internal links
```

integreat_cms/release_notes/current/unreleased/2343.yml
```diff
@@ -0,0 +1,2 @@
+en: When duplicating a region, perform internal link replacement as background task.
+de: Beim duplizieren des Inhalts einer Region werden interne Verlinkungen nun im Hintergrund ausgetauscht.
```
timobrembeck, line 2:
```suggestion
en: When duplicating a region, perform internal link replacement as background task
de: Beim Duplizieren des Inhalts einer Region werden interne Verlinkungen nun im Hintergrund ausgetauscht
```

timobrembeck (commented):
Sorry it took me so long to revisit this :innocent:

Could you resolve the git conflicts, fix the typos and squash your commits?
Then I think we can finally merge this PR :rocket:

timobrembeck (approved):
:tada:

MizukiTemma (approved):
Looks good 😸 Thank you!

PR #2559, opened 2023-11-17
Validate slugs against reserved patterns

sarahsporck:
### Short description

Currently regions can be created with slugs that are already in use in the webapp. The same goes for page patterns that should be reserved.

Some changes in this PR are required for #2391.

### Proposed changes

- Validate against reserved region patterns
- Validate against page patterns

### Resolved issues

Fixes: #1792 

### DIFF
--- a/integreat_cms/cms/utils/slug_utils.py
+++ b/integreat_cms/cms/utils/slug_utils.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 import logging
 from typing import TYPE_CHECKING
 
+from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.utils.text import slugify
 from django.utils.translation import gettext_lazy as _
@@ -127,9 +128,8 @@ def generate_unique_slug(**kwargs: Unpack[SlugKwargs]) -> str:
     if not slug:
         if fallback not in cleaned_data:
             raise ValidationError(
-                _("Cannot generate slug from %(fallback)s."),
+                _("Cannot generate slug from {}.").format(_(fallback)),
                 code="invalid",
-                params={"fallback": _(fallback)},
             )
         # Check whether slug field supports unicode
         allow_unicode = object_instance._meta.get_field("slug").allow_unicode
@@ -168,7 +168,17 @@ def generate_unique_slug(**kwargs: Unpack[SlugKwargs]) -> str:
             else:
                 # the current object is also allowed to have the same slug
                 other_objects = other_objects.exclude(id=object_instance.id)
-        if not other_objects.exists():
+        if (
+            not other_objects.exists()
+            and not (
+                foreign_model == "page"
+                and unique_slug in settings.RESERVED_REGION_PAGE_PATTERNS
+            )
+            and not (
+                foreign_model == "region"
+                and unique_slug in settings.RESERVED_REGION_SLUGS
+            )
+        ):
             break
         i += 1
         unique_slug = f"{slug}-{i}"
--- a/integreat_cms/core/settings.py
+++ b/integreat_cms/core/settings.py
@@ -64,6 +64,26 @@ IMPRINT_SLUG: Final[str] = os.environ.get("INTEGREAT_CMS_IMPRINT_SLUG", "disclai
 
 TEST_REGION_SLUG: Final[str] = "testumgebung"
 
+#: Reserved region slugs, that are used in the webapp
+RESERVED_REGION_SLUGS: Final[list[str]] = [
+    "landing",
+    "recommend",
+    "licenses",
+    "main-disclaimer",
+    "jpal",
+    "not-found",
+]
+
+#: Reserved region page patterns
+RESERVED_REGION_PAGE_PATTERNS: Final[list[str]] = [
+    IMPRINT_SLUG,
+    "news",
+    "events",
+    "locations",
+    "offers",
+    "search",
+]
+
 #: URL to the Integreat Website
 WEBSITE_URL: Final[str] = os.environ.get(
     "INTEGREAT_CMS_WEBSITE_URL", "https://integreat-app.de"
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -7906,9 +7906,8 @@ msgid "The PDF could not be successfully generated."
 msgstr "PDF-Datei konnte nicht erfolgreich erzeugt werden."
 
 #: cms/utils/slug_utils.py
-#, python-format
-msgid "Cannot generate slug from %(fallback)s."
-msgstr "Slug kann nicht aus dem %(fallback)s abgeleitet werden."
+msgid "Cannot generate slug from {}."
+msgstr "Slug kann nicht aus dem {} abgeleitet werden."
 
 #: cms/utils/welcome_mail_utils.py
 msgid "activation mail"
--- a/integreat_cms/release_notes/current/unreleased/1792.yml
+++ b/integreat_cms/release_notes/current/unreleased/1792.yml
@@ -1,0 +1,2 @@
+en: Disallow using region and page slugs with reserved patterns.
+de: Verbiete das Verwenden von reservierten URL-Parametern.
--- a/tests/cms/utils/test_slug_utils.py
+++ b/tests/cms/utils/test_slug_utils.py
@@ -1,0 +1,95 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import pytest
+from django.core.exceptions import ValidationError
+
+from integreat_cms.cms.models import PageTranslation, Region
+from integreat_cms.cms.utils.slug_utils import generate_unique_slug
+
+if TYPE_CHECKING:
+    from pytest_django.fixtures import SettingsWrapper
+
+    from integreat_cms.cms.utils.slug_utils import SlugKwargs
+
+
+@pytest.mark.django_db
+def test_generate_unique_slug_fallback(
+    settings: SettingsWrapper, load_test_data: None
+) -> None:
+    """
+    Test whether the :meth:`~integreat_cms.cms.utils.slug_utils.generate_unique_slug_helper` function correctly uses the fallback property
+    when no slug is provided
+    """
+    region = Region.objects.get(slug="augsburg")
+    kwargs: SlugKwargs = {
+        "cleaned_data": {"name": "unique_slug_fallback"},
+        "manager": Region.objects,
+        "object_instance": region,
+        "foreign_model": "region",
+        "fallback": "name",
+    }
+    assert (
+        generate_unique_slug(**kwargs) == "unique_slug_fallback"
+    ), "Name is not used as fallback when slug is missing"
+
+
+@pytest.mark.django_db
+def test_generate_unique_slug_reserved_region_slug(
+    settings: SettingsWrapper, load_test_data: None
+) -> None:
+    """
+    Test whether the :meth:`~integreat_cms.cms.utils.slug_utils.generate_unique_slug_helper` function returns the correct unique slug
+    when the new region slug is a reserved slug
+    """
+    region = Region.objects.get(slug="augsburg")
+    kwargs: SlugKwargs = {
+        "slug": "landing",
+        "cleaned_data": {},
+        "manager": Region.objects,
+        "object_instance": region,
+        "foreign_model": "region",
+        "fallback": "name",
+    }
+    assert (
+        generate_unique_slug(**kwargs) == "landing-2"
+    ), "Reserved region slug is not prevented"
+
+
+@pytest.mark.django_db
+def test_generate_unique_slug_reserved_page_slug(
+    settings: SettingsWrapper, load_test_data: None
+) -> None:
+    """
+    Test whether the :meth:`~integreat_cms.cms.utils.slug_utils.generate_unique_slug_helper` function  function returns the correct unique slug
+    when the new page slug is a reserved slug
+    """
+    page = PageTranslation.objects.filter(page__region__slug="augsburg").first()
+    kwargs: SlugKwargs = {
+        "slug": "disclaimer",
+        "cleaned_data": {},
+        "manager": PageTranslation.objects,
+        "language": page.language,
+        "object_instance": page,
+        "foreign_model": "page",
+    }
+    assert (
+        generate_unique_slug(**kwargs) == "disclaimer-2"
+    ), "Reserved imprint slug is not prevented for pages"
+
+
+def test_generate_unique_slug_no_fallback() -> None:
+    """
+    Test whether the :meth:`~integreat_cms.cms.utils.slug_utils.generate_unique_slug_helper` throws a :class:`django.core.exceptions.ValidationError`,
+    while the fallback property does not exist
+    """
+    kwargs: SlugKwargs = {
+        "cleaned_data": {},
+        "manager": [],
+        "object_instance": Region(),
+        "foreign_model": "region",
+        "fallback": "name",
+    }
+    with pytest.raises(ValidationError):
+        generate_unique_slug(**kwargs)


integreat_cms/cms/forms/pages/page_translation_form.py
```diff
@@ -23,3 +27,26 @@ class Meta:
...
         fields = MachineTranslationForm.Meta.fields + ["slug", "hix_score"]
+
+    def clean_slug(self):
+        """
+        Validate the slug field (see :ref:`overriding-modelform-clean-method`)
+
+        :return: A validated slug
+        :rtype: str
+        """
+        cleaned_slug = self.cleaned_data["slug"]
```
david-venhoff, line 38:
```suggestion
        super().clean_slug()
        cleaned_slug = self.cleaned_data["slug"]
```
This overwrites a the `clean_slug` method at https://github.com/digitalfabrik/integreat-cms/blob/2c560214ffb3a8af52435dec32192d7dca730a21/integreat_cms/cms/forms/custom_content_model_form.py#L170-L181

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -2009,6 +2009,11 @@ msgstr "Quell-Region für Live-Inhalte"
 msgid "Enable write access via API for this page"
 msgstr "Schreib-Zugriff via API für diese Seite aktivieren"
 
+#: cms/forms/pages/page_translation_form.py cms/models/regions/region.py
+#, python-format
+msgid "%(slug)s is a reserved slug."
+msgstr ""
```
david-venhoff, line 2015:
```suggestion
msgstr "$(slug)s ist ein reservierter URL-Parameter."
```

david-venhoff (approved):
Thanks! I just wonder whether it would be better to automatically fix the slug instead of showing an error message, since that is the current solution when e.g. a duplicate slug is detected. But that is not too important I guess.
This might also need a changelog entry.

integreat_cms/cms/forms/pages/page_translation_form.py
```diff
@@ -23,3 +27,26 @@ class Meta:
...
+        if is_root_slug and any(
+            map(
+                lambda reserved_pattern: cleaned_slug == reserved_pattern,
+                settings.RESERVED_REGION_PAGE_PATTERNS,
+            )
+        ):
+            raise ValidationError(
+                _("%(slug)s is a reserved slug."),
+                code="invalid",
+                params={"slug": cleaned_slug},
```
timobrembeck, line 49:
```suggestion
                _("{} is a reserved slug.").format(cleaned_slug),
                code="invalid",
```
I'm aware that this goes against the recommendations of the [Django docs](https://docs.djangoproject.com/en/3.2/ref/forms/validation/#raising-validationerror), but I think `{}` as placeholder is much more readable than the oldschool `%` formatting :sweat_smile: 
Also, for consistency reasons, since this is the formatting syntax we're using everywhere else in the project...

integreat_cms/cms/models/regions/region.py
```diff
@@ -48,6 +49,17 @@
...
 
+def validate_slug(slug):
+    """
+    Helper function to validate a slug is not a reserved slug
+
+    :param slug: The slug to be validated
+    :type slug: str
+    """
+    if slug in settings.RESERVED_REGION_SLUGS:
+        raise ValidationError(_("%(slug)s is a reserved slug."), params={"slug": slug})
```
timobrembeck, line 60: I would choose one consistent enforcement method, either validators or the clean method.

integreat_cms/locale/de/LC_MESSAGES/django.po
timobrembeck, line 2015:
```suggestion
msgid "{} is a reserved slug."
msgstr "{} ist ein reservierter URL-Parameter."
```

integreat_cms/core/settings.py
```diff
@@ -64,6 +64,26 @@
 
 TEST_REGION_SLUG: Final[str] = "testumgebung"
 
+#: Reserved region slugs, that are used in the webapp
+RESERVED_REGION_SLUGS = [
```
timobrembeck, line 68:
```suggestion
RESERVED_REGION_SLUGS: Final[list[str]] = [
```
Technically, this is not really required, but I think it's helpful to prevent the settings from being used incorrectly or being overridden...

integreat_cms/core/settings.py
```diff
@@ -64,6 +64,26 @@
...
+    "landing",
+    "recommend",
+    "licenses",
+    "main-disclaimer",
+    "jpal",
+    "not-found",
+]
+
+#: Reserved region page patterns
+RESERVED_REGION_PAGE_PATTERNS = [
```
timobrembeck, line 78:
```suggestion
RESERVED_REGION_PAGE_PATTERNS: Final[list[str]] = [
```

timobrembeck (changes requested):
Thanks! :+1:

> I just wonder whether it would be better to automatically fix the slug instead of showing an error message, since that is the current solution when e.g. a duplicate slug is detected.

Cool idea! If you decide to go for this, the cleaning logic could be moved here:
https://github.com/digitalfabrik/integreat-cms/blob/6e60e28067c1f525019e12e6e8671060e9784c46/integreat_cms/cms/utils/slug_utils.py#L171

integreat_cms/cms/forms/pages/page_translation_form.py
```diff
@@ -23,3 +27,26 @@ class Meta:
...
+        :rtype: str
+        """
+        cleaned_slug = self.cleaned_data["slug"]
+        is_root_slug = not self.instance.ancestor_path
+        if is_root_slug and any(
+            map(
+                lambda reserved_pattern: cleaned_slug == reserved_pattern,
+                settings.RESERVED_REGION_PAGE_PATTERNS,
+            )
+        ):
```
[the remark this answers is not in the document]
sarahsporck: ahh. you're right. Before I checked for slug.startsWith, as I thought the slug refers to the url_infix + slug. I'll simplify the check.

integreat_cms/cms/forms/pages/page_translation_form.py
```diff
@@ -23,3 +27,26 @@ class Meta:
...
+
+    def clean_slug(self):
+        """
+        Validate the slug field (see :ref:`overriding-modelform-clean-method`)
+
+        :return: A validated slug
+        :rtype: str
+        """
+        cleaned_slug = self.cleaned_data["slug"]
+        is_root_slug = not self.instance.ancestor_path
```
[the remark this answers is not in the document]
sarahsporck: In that case I'd prefer using a validator. When considering the ancestor_path this was not possible. So I'll move everything in a validator.

integreat_cms/cms/migrations/0084_alter_region_slug.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('cms', '0083_add_page_based_offer_fields'),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
sarahsporck, line 14: Question @timobrembeck. When I add a validator for the PageTranslation (via the `@modify_fields` wrapper) no automatic migration is generated. Should we then in general prefer using the form `clean_xyz()`-approach to validate such properties?

timobrembeck: Hmm, probably yes. Out of the top of my head, I don't have an idea how to overwrite the validators of an existing field if `@modify_fields` doesn't work...

sarahsporck: It does work. But the migrations don't work out of the box. But never mind. I'll just use the clean approach as it is more intuitive and probably less error prone 👍🏼

integreat_cms/cms/forms/pages/page_translation_form.py
sarahsporck: Linty doesn't like this -> `consider-using-f-string`

integreat_cms/cms/models/regions/region.py
sarahsporck: I decided to go for the clean method

integreat_cms/cms/forms/pages/page_translation_form.py
timobrembeck: Unfortunately, [gettext doesn't work with f-strings](https://savannah.gnu.org/bugs/?61596), so the `{}` syntax is as close as it gets.
But for me, pylint is correctly able to detect that using an f-string here is not an option... :thinking: 
What did you mean by "Linty", is this a separate tool to check the style conventions, or just a typo?

integreat_cms/release_notes/current/unreleased/1792.yml
```diff
@@ -0,0 +1,2 @@
+en: Disallow using region and page slugs with reserved patterns.
+de: Verbiete das verwenden von reservierten urls.
```
timobrembeck, line 2:
```suggestion
de: Verbiete das Verwenden von reservierten URL-Parametern.

```

integreat_cms/cms/forms/pages/page_translation_form.py
```diff
@@ -2,6 +2,8 @@
 
 import logging
 
+from django.utils.translation import gettext_lazy as _
+
```
timobrembeck, line 6:
```suggestion
```

integreat_cms/cms/utils/slug_utils.py
```diff
@@ -169,6 +170,24 @@ def generate_unique_slug(**kwargs: Unpack[SlugKwargs]) -> str:
...
+                )
+            if (
+                foreign_model in ["region"]
+                and unique_slug in settings.RESERVED_REGION_SLUGS
+            ):
+                raise ValidationError(
+                    _("%(slug)s is a reserved slug."),
+                    code="invalid",
+                    params={"slug": unique_slug}
+                )
```
timobrembeck, line 190:
```suggestion
        if not (
            other_objects.exists()
            or (
                foreign_model == "page"
                and unique_slug in settings.RESERVED_REGION_PAGE_PATTERNS
            )
            or (
                foreign_model == "region"
                and unique_slug in settings.RESERVED_REGION_SLUGS
            )
        ):
```
If you move the check into the slug helper, then I think it does not make sense to throw an error – just appending a number seems much more user friendly, as @david-venhoff suggested.
Then we also don't have to argue about formatting strings :sweat_smile:

integreat_cms/core/settings.py
```diff
@@ -64,6 +64,26 @@
...
+    "recommend",
+    "licenses",
+    "main-disclaimer",
+    "jpal",
+    "not-found",
+]
+
+#: Reserved region page patterns
+RESERVED_REGION_PAGE_PATTERNS: Final[list[str]] = [
+    "disclaimer",
```
timobrembeck, line 79:
```suggestion
    IMPRINT_SLUG,
```
Since the imprint slug is configurable, we should also include it here.

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -7910,6 +7910,11 @@ msgstr "PDF-Datei konnte nicht erfolgreich erzeugt werden."
 msgid "Cannot generate slug from %(fallback)s."
 msgstr "Slug kann nicht aus dem %(fallback)s abgeleitet werden."
 
+#: cms/utils/slug_utils.py
+#, python-format
+msgid "%(slug)s is a reserved slug."
+msgstr "%(slug)s ist ein reservierter URL-Parameter"
+
```
timobrembeck, line 7917:
```suggestion
```

timobrembeck (changes requested):

integreat_cms/cms/forms/pages/page_translation_form.py
sarahsporck: Oops a typo. At least locally i get the warning with pylint. I've commited it nevertheless. If it works in the ci this should be fine.

integreat_cms/locale/de/LC_MESSAGES/django.po
sarahsporck: Why remove these lines? :)

tests/cms/utils/test_slug_utils.py
```diff
@@ -0,0 +1,91 @@
...
+@pytest.mark.django_db
+def test_generate_unique_slug_fallback(
+    settings: SettingsWrapper, load_test_data: None
+) -> None:
+    """
+    Test whether the :meth:`~integreat_cms.cms.utils.slug_utils.generate_unique_slug_helper` function correctly uses the fallback property
+    when no slug is provided
+    """
+    settings.DEBUG = True
+    mock_object = Region.objects.get(slug="augsburg")
```
timobrembeck, line 24:
```suggestion
    region = Region.objects.get(slug="augsburg")
```
I think when you query an object from the database, it's not a mocked object, but a real one^^ Sorry for being so nitpicky :see_no_evil:

tests/cms/utils/test_slug_utils.py
```diff
@@ -0,0 +1,91 @@
...
+    print(mock_object)
```
timobrembeck, line 25:
```suggestion
```

tests/cms/utils/test_slug_utils.py
```diff
@@ -0,0 +1,91 @@
...
+    kwargs: SlugKwargs = {
+        "cleaned_data": {"name": "unique_slug_fallback"},
+        "manager": Region.objects,
+        "object_instance": mock_object,
```
timobrembeck, line 29:
```suggestion
        "object_instance": region,
```

tests/cms/utils/test_slug_utils.py
```diff
@@ -0,0 +1,91 @@
...
+        "foreign_model": "region",
+        "fallback": "name",
+    }
+    assert generate_unique_slug(**kwargs), "unique_slug_fallback"
```
timobrembeck, line 33:
```suggestion
    assert (
        generate_unique_slug(**kwargs) == "unique_slug_fallback"
    ), "Name is not used as fallback when slug is missing"
```
The syntax of the [assert](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement) statement only evaluates the first expression, the second one after the `,` is an optional message that is printed when the assertion fails. So `assert generate_unique_slug(**kwargs), "unique_slug_fallback"` only checks whether `generate_unique_slug(**kwargs)` evaluates to something "truthy" and would print the literal "unique_slug_fallback" when the result is falsey. This is also the reason why the other two tests are passing even though the if condition is incorrect.

tests/cms/utils/test_slug_utils.py
```diff
@@ -0,0 +1,91 @@
...
+    mock_object = Region.objects.get(slug="augsburg")
+    kwargs: SlugKwargs = {
+        "slug": "landing",
+        "cleaned_data": {"name": "unique_slug_fallback"},
+        "manager": Region.objects,
+        "object_instance": mock_object,
+        "foreign_model": "region",
+        "fallback": "name",
+    }
+    assert generate_unique_slug(**kwargs), "landing-2"
```
timobrembeck, line 54:
```suggestion
    region = Region.objects.get(slug="augsburg")
    kwargs: SlugKwargs = {
        "slug": "landing",
        "cleaned_data": {},
        "manager": Region.objects,
        "object_instance": region,
        "foreign_model": "region",
        "fallback": "name",
    }
    assert (
        generate_unique_slug(**kwargs) == "landing-2"
    ), "Reserved region slug is not prevented"
```

tests/cms/utils/test_slug_utils.py
```diff
@@ -0,0 +1,91 @@
...
+    mock_object = PageTranslation.objects.filter(page__region__slug="augsburg").first()
+    kwargs: SlugKwargs = {
+        "slug": "disclaimer",
+        "cleaned_data": {},
+        "manager": PageTranslation.objects,
+        "language": mock_object.language,
+        "object_instance": mock_object,
+        "foreign_model": "page",
+    }
+    assert generate_unique_slug(**kwargs), "disclaimer-2"
```
timobrembeck, line 75:
```suggestion
    page_translation = PageTranslation.objects.filter(page__region__slug="augsburg").first()
    kwargs: SlugKwargs = {
        "slug": "disclaimer",
        "cleaned_data": {},
        "manager": PageTranslation.objects,
        "language": page_translation.language,
        "object_instance": page_translation,
        "foreign_model": "page",
    }
    assert (
        generate_unique_slug(**kwargs) == "disclaimer-2"
    ), "Reserved imprint slug is not prevented for pages"
```

timobrembeck (changes requested):
Awesome, thanks for adding the tests! :muscle:

integreat_cms/locale/de/LC_MESSAGES/django.po
timobrembeck: Just because I saw that the same translation (just with the `{}` formatting) was already included in the translation file (commented out) and I thought we don't need to keep both. I intended to keep one of them, but now I saw you removed both of them, that's no problem, no need to re-add it :sweat_smile:

tests/cms/utils/test_slug_utils.py
```diff
@@ -0,0 +1,91 @@
...
+
+@pytest.mark.django_db
+def test_generate_unique_slug_fallback(
+    settings: SettingsWrapper, load_test_data: None
+) -> None:
+    """
+    Test whether the :meth:`~integreat_cms.cms.utils.slug_utils.generate_unique_slug_helper` function correctly uses the fallback property
+    when no slug is provided
+    """
+    settings.DEBUG = True
```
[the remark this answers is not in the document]
sarahsporck: Actually, I don't need this. Just saw it in another testfile, while getting the test_data loading and thought it was required. I'll remove it again.

tests/cms/utils/test_slug_utils.py
sarahsporck: But you're also completely right. originally I tried to mock this, but it didn't work out. Will be changed to your suggestion.

tests/cms/utils/test_slug_utils.py
sarahsporck: Uhhm. Oops :D. 🧑‍🏭

timobrembeck (approved):
Great, thanks! :tada:

timobrembeck (commented):
Could you squash your commits before merging?

Issue #2336, opened 2023-07-05
Deadlock when deleting regions


timobrembeck:
### Describe the Bug

Deleting regions via the region form is currently not possble

### Steps to Reproduce

1. Go to region form
2. Click on 'Delete region'
3. Wait

### Expected Behavior

The region should be deleted

### Actual Behavior

Nothing happens, the page is loading forever

### Additional Information

Seems to be related to this traceback:

<details>
  <summary>Traceback</summary>

  ```
In [3]: Region.objects.get(slug="kreis-re").delete()
^C---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
File /opt/integreat-cms/.venv/lib/python3.9/site-packages/django/db/models/fields/related_descriptors.py:173, in ForwardManyToOneDescriptor.__get__(self, instance, cls)
    172 try:
--> 173     rel_obj = self.field.get_cached_value(instance)
    174 except KeyError:

File /opt/integreat-cms/.venv/lib/python3.9/site-packages/django/db/models/fields/mixins.py:15, in FieldCacheMixin.get_cached_value(self, instance, default)
     14 try:
---> 15     return instance._state.fields_cache[cache_name]
     16 except KeyError:

KeyError: 'language'

During handling of the above exception, another exception occurred:

KeyboardInterrupt                         Traceback (most recent call last)
Cell In[3], line 1
----> 1 Region.objects.get(slug="kreis-re").delete()

File /opt/integreat-cms/.venv/lib/python3.9/site-packages/django/db/models/base.py:967, in Model.delete(self, using, keep_parents)
    965 collector = Collector(using=using)
    966 collector.collect([self], keep_parents=keep_parents)
--> 967 return collector.delete()

File /opt/integreat-cms/.venv/lib/python3.9/site-packages/django/db/models/deletion.py:404, in Collector.delete(self)
    402 for model, obj in self.instances_with_model():
    403     if not model._meta.auto_created:
--> 404         signals.pre_delete.send(
    405             sender=model, instance=obj, using=self.using
    406         )
    408 # fast deletes
    409 for qs in self.fast_deletes:

File /opt/integreat-cms/.venv/lib/python3.9/site-packages/django/dispatch/dispatcher.py:180, in Signal.send(self, sender, **named)
    177 if not self.receivers or self.sender_receivers_cache.get(sender) is NO_RECEIVERS:
    178     return []
--> 180 return [
[...]

PR #2560, opened 2023-11-18
Improve performance when deleting regions

david-venhoff:
### Short description

This pr suppresses most linkcheck deletion listeners when deleting regions, which hopefully improves performance enough that deleting big regions will be possible again.
Currently the `Link` post_delete handlers from linkcheck will still run, which is unnecessary. However this is hopefully not a big problem, since those only run per link and not per page translation. 

### Proposed changes

- Delete regions with the `disable_listeners` context manager

### Side effects

- The db may be in an inconsistent state after deleting a region. However, this should only result in internal links to content of the deleted region not being flagged as broken. Also the db will be fixed after the next run of the `findlinks` management command.

### Resolved issues

fixes: #2336


MizukiTemma (approved):
Looks good 😃 
Thank you for analysing the problem 👍

charludo (approved):

Issue #2561, opened 2023-11-20
JS Error in Region selection for News


svenseeberg:
### Describe the Bug
Selecting multiple regions for news fails with a JS error. It seems to be Browser specific behavior. It does not work on stable Chrome and Firefox releases but works in Firefox Nightly.

### Steps to Reproduce

1. Log in as root on a current dev setup in Chromium 119 or current Firefox (not Nightly)
2. Open http://localhost:8000/birkenfeld/push-notifications/de/new/ 
3. Open the developer tool
4. Select a region checkbox
5. See error

### Expected Behavior
Region checkboxes should be clickable

### Actual Behavior
Clicking checkboxes fails with JS error.

### Additional Information

![image](https://github.com/digitalfabrik/integreat-cms/assets/15197000/0adab51b-1f22-44ca-8971-f9c21caa5acf)

```
    organization.ts:11 Uncaught TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
    at Function.from (<anonymous>)
    at HTMLUListElement.updateOrganizationChoices (organization.ts:11:29)
updateOrganizationChoices @ organization.ts:11
```

svenseeberg:
`regions` in line organization.ts:11 is defined, `regions.selectedOptions` is not:
![image](https://github.com/digitalfabrik/integreat-cms/assets/15197000/a00fdf23-b0db-41c5-a4b4-94b0d71e1abd)
![image](https://github.com/digitalfabrik/integreat-cms/assets/15197000/158f3b22-ee6f-4eeb-985d-09af532312f8)

PR #2566, opened 2023-11-24
Fix JS error in Region selection for News

charludo:
### Short description

Fixes an JS error in the News creation page that was caused by an unrelated-to-the-news piece of JS being triggered by the presence of HTML elements with matchinf IDs

### Proposed changes

- check if other relevant fields are present (and the JS code should run), don't run the offending function otherwise.
- 

### Side effects

- I assume none (😅)
- 

### Resolved issues

Fixes: #2561


ulliholtgrave (approved):
Nice catch! :tada: :D

david-venhoff (approved):
Thanks, LGTM!
This might be worth a changelog entry, since it is affecting the users.

PR #2568, opened 2023-11-24
[WIP] Add chat bot integration

svenseeberg:
### Short description
Rough outline of how a chat bot for app users could be integrated into the Integreat CMS. The CMS only acts as a proxy/middleware between different components (Zammad, LLMs, App, etc) and its main task is to map sessions.

### Proposed changes
- N/A

### Side effects
- N/A

### Resolved issues
- #2619 

### DIFF
--- a/integreat_cms/chatapi/views.py
+++ b/integreat_cms/chatapi/views.py
@@ -1,0 +1,26 @@
+"""
+This module includes functions related to the regions API endpoint.
+"""
+from django.http import Http404, JsonResponse
+from .models import ChatSession
+from ..decorators import json_response
+
+
+@json_response
+# pylint: disable=unused-argument
+def region_by_slug(request: HttpRequest, region_slug: str) -> JsonResponse:
+    """
+    Retrieve a single region and transform result into JSON
+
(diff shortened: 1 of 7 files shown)


integreat_cms/chatapi/views.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+@json_response
+# pylint: disable=unused-argument
+def region_by_slug(request: HttpRequest, region_slug: str) -> JsonResponse:
+    """
+    Retrieve a single region and transform result into JSON
+
+    :return: JSON object according to APIv3 live regions endpoint definition
+    """
```
steffenkleinle, line 16: Should be adjusted, just copy/pasted I guess?

integreat_cms/chatapi/views.py
```diff
@@ -0,0 +1,26 @@
...
+    session_token = request.GET["session_id"]
+    chat_session = ChatSession.object.filter(session_token=session_token)
+    if not chat_session:
+        ZammadHandler.create_ticket()
```
steffenkleinle, line 20: Here we should probably persist a session token/ticket id pair, right?

PR #2575, opened 2023-11-28
Add test for media library

MizukiTemma:
### Short description

This PR adds test for media library (integreat_cms/cms/views/media/media_actions.py).

### Proposed changes

Expand the test for the following actions.
- [x] Uploading files
- [x] Renaming files
- [x] Changing alt texts of files
- [x] Replacing files
- [x] Deleting files
- [x] Creating directories
- [x] Renaming directories
- [x] Deleting directories

### Side effects

1. Add some new test data
2. Existing test media file objects are changed to .png, because thumbnail generation during replacing fails with .svg and replacement cannot be testet.
3. Expected API responses are adjusted due to 2.

### Resolved issues

Fixes: #1997 (#2582)

### DIFF
--- a/integreat_cms/cms/forms/media/replace_media_file_form.py
+++ b/integreat_cms/cms/forms/media/replace_media_file_form.py
@@ -131,9 +131,11 @@ class ReplaceMediaFileForm(CustomModelForm):
 
     def save(self, commit: bool = True) -> MediaFile:
         # Remove old file
-        os.remove(self.original_file_path)
-        logger.debug("Removed old file %r", self.original_file_path)
-
+        try:
+            os.remove(self.original_file_path)
+            logger.debug("Removed old file %r", self.original_file_path)
+        except FileNotFoundError:
+            logger.debug("The file %r could not be removed", self.original_file_path)
         # Remove old thumbnail
         if self.original_thumbnail_path:
             os.remove(self.original_thumbnail_path)
--- a/tests/cms/test_media_library.py
+++ b/tests/cms/test_media_library.py
@@ -1,0 +1,479 @@
+from __future__ import annotations
+
+import os
+from urllib import parse
+
+import pytest
+from django.conf import settings
+from django.core.files.uploadedfile import SimpleUploadedFile
+from django.test.client import Client
+from django.urls import reverse
+
+from integreat_cms.cms.models import Directory, MediaFile
+from tests.conftest import (
+    ANONYMOUS,
+    HIGH_PRIV_STAFF_ROLES,
+    PRIV_STAFF_ROLES,
+    STAFF_ROLES,
+)
+
+
+@pytest.mark.django_db
+def test_directory_path(
+    load_test_data: None,
+    login_role_user: tuple[Client, str],
+) -> None:
+    # Log the user in
+    client, role = login_role_user
+
+    query = parse.urlencode(
+        {
+            "directory": 1,
+        }
+    )
+    next_url = reverse("mediacenter_directory_path")
+    url = f"{next_url}?{query}"
+
+    response = client.get(url)
+
+    if role in STAFF_ROLES:
+        assert response.status_code == 200
+        # Check contents in the response
+        assert "Test Directory" in response.content.decode("utf-8")
+        assert not "Empty Directory" in response.content.decode("utf-8")
+    elif role == ANONYMOUS:
+        # For anonymous users, we want to redirect to the login form instead of showing an error
+        assert response.status_code == 302
+        assert (
+            response.headers.get("location") == f"{settings.LOGIN_URL}?next={next_url}"
+        )
+    else:
+        assert response.status_code == 403
+
+
+@pytest.mark.django_db
+def test_get_directory_content(
+    load_test_data: None,
+    login_role_user: tuple[Client, str],
+) -> None:
+    # Log the user in
+    client, role = login_role_user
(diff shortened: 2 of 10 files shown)


MizukiTemma:
The test is successfull in a local run but not on git. Could somebody help me with that? 🙏

seluianova:
> The test is successfull in a local run but not on git. Could somebody help me with that? 🙏

What I have found so far:

✳️ Files integreat-logo.png and malte-logo.png are not created in the directory /media/global on the CI server.
As I understand, they are created locally by loadtestdata.sh, but this script is not running on the CI server.

✳️ The replace media file action is the only action, that requires the file to actually exist.

tests/cms/views/utils.py
```diff
@@ -46,7 +49,39 @@ def check_view_status_code(
         # If the post data is a string, assume json as content type
         if isinstance(post_data, str):
             post_kwargs["content_type"] = "application/json"
-        response = client.post(url, **post_kwargs)
+        if view_name == "mediacenter_replace_file":
```
seluianova, line 52: I'm not insisting here, but would like to share some thoughts to consider.
I see some advantages in implementing tests for views as separate functions (rather than all in this one):
- if-s by specific view names increase complexity
- tests could be run separately per suite (as I understand, I can't run media library-only tests now?)
- we can easily enrich the tests with view-specific checks, e.g., in a database (to verify that media files have actually been created/updated/deleted)
- tests will be more like user scenarios

I'm open to other opinions

MizukiTemma: Thank you for the advice 🙂 I find it a good idea too and separated the tests for media library.

integreat_cms/cms/forms/media/replace_media_file_form.py
```diff
@@ -130,10 +130,12 @@ def clean(self) -> dict:
...
-        # Remove old file
-        os.remove(self.original_file_path)
-        logger.debug("Removed old file %r", self.original_file_path)
-
+        # Remove old
+        try:
+            os.remove(self.original_file_path)
+            logger.debug("Removed old file %r", self.original_file_path)
+        except FileNotFoundError:
+            logger.debug("The file %r could not be removed", self.original_file_path)
```
MizukiTemma, line 138: The test file is not stroed during the test and it caused an error as the path does not exists. So `try` and `except` are added. I hope this does not harm the functionality and bring some safety in practice.

david-venhoff (approved):
Thanks!
I think there are a few additional tests that we could perform:
- Verify the data received from the various media library endpoints
- Verify that a directory was actually created/renamed/deleted when requesting it
- Creating a directory as a subdirectory
- Verify that a file actually got uploaded/renamed/replaced/moved/deleted after the request

However this already looks like a good improvement and I am not sure how important these additional tests are.

MizukiTemma:
Thank you for review 😸 
I've added the last three of your suggestions.

> * Verify the data received from the various media library endpoints

What do you mean exactly here?

tests/cms/test_media_library.py
```diff
@@ -0,0 +1,465 @@
...
+
+    query = parse.urlencode(
+        {
+            "directory": 1,
+        }
+    )
+    next_url = reverse("mediacenter_directory_path")
+    url = f"{next_url}?{query}"
+
+    response = client.post(url)
```
seluianova, line 37:
```suggestion
    response = client.get(url)
```

We handle both methods, but I think it's more important to test GET because the GET request is sent from JS.

tests/cms/test_media_library.py
```diff
@@ -0,0 +1,465 @@
...
+
+    query = parse.urlencode(
+        {
+            "directory": 1,
+        }
+    )
+    next_url = reverse("mediacenter_get_directory_content")
+    url = f"{next_url}?{query}"
+
+    response = client.post(url)
```
seluianova, line 67:
```suggestion
    response = client.get(url)
```

tests/cms/test_media_library.py
```diff
@@ -0,0 +1,465 @@
...
+
+    query = parse.urlencode(
+        {
+            "file": 1,
+        }
+    )
+    next_url = reverse("mediacenter_get_file_usages")
+    url = f"{next_url}?{query}"
+
+    response = client.post(url)
```
seluianova, line 424:
```suggestion
    response = client.get(url)
```

tests/cms/test_media_library.py
```diff
@@ -0,0 +1,465 @@
...
+
+    query = parse.urlencode(
+        {
+            "query": "test",
+        }
+    )
+    next_url = reverse("mediacenter_get_search_result")
+    url = f"{next_url}?{query}"
+
+    response = client.post(url)
```
seluianova, line 454:
```suggestion
    response = client.get(url)
```

seluianova (approved):
> Verify the data received from the various media library endpoints

I think this means checking the json in the response itself. Might make sense especially for GET requests, because for POST we check the changes at least in the database now.

Whether you add this or not, current tests look good to me. Thanks :)

david-venhoff:
> I think this means checking the json in the response itself. Might make sense especially for GET requests, because for POST we check the changes at least in the database now.

Yes, this is what I meant :)

MizukiTemma:
@david-venhoff @seluianova 
Thank you for the replies. That's added now 👍

Issue #310, opened 2020-03-03
CSV export feedback


svenseeberg:
Some cities want to export their feedback.

timoludwig:
This is blocked by #32 (View feedback).

osmers:
Since I just got a request from a municipality regarding this issue, I wanted to open it up for discussion - especially how comprehensive it is? Is the effort low, medium or high? Then I can discuss it with the team - but since there is no other way for export (e.g. via API as with POIs) it might be a worthwhile issue to put into Q1

timobrembeck:
I'd say the effort is low - medium, so should be doable in 24Q1.

PR #2593, opened 2023-12-22
Add option to export feedback to CSV

JoeyStk:
### Proposed changes

- Add import-export library
- add bulk action option to export feedback to CSV
- add test for export

### Side effects

- This issue was moved from the current milestone to 24Q4 while Peter and I were working on it. So I suppose you can postpone reviewing this until then :sweat_smile:

### Resolved issues

Fixes: #310


docs/src/conf.py
```diff
@@ -210,6 +214,7 @@
     ("py:class", "xml.dom.minidom.Element"),
     ("py:func", "django.contrib.sitemaps.Sitemap._urls"),
     ("py:func", "django.utils.text.capfirst"),
+    ("py:class", "import_export.resources.Resource.get_instance"),
```
timobrembeck, line 217:
```suggestion
```
This doesn't work. You can only exclude entities that don't have documentation, you cannot ignore errors in existing docs here.

integreat_cms/cms/views/feedback/region_feedback_actions.py
```diff
@@ -169,3 +173,113 @@ def delete_region_feedback(
...
+        ::
+
+            class YourResource(ModelResource):
+                def skip_row(self, instance, original,
+                             row, import_validation_errors=None):
+                    # Add code here
+                    return super().skip_row(instance, original, row,
+                                            import_validation_errors=import_validation_errors)
+        """
+        return super().skip_row(instance, original, row, import_validation_errors)
```
timobrembeck, line 240:
```suggestion
    # pylint: disable=useless-parent-delegation
    def get_instance(self, *args: Any, **kwargs: Any) -> Any:
        """
        See :meth:`import_export.resources.Resource.get_instance`
        """
        return super().get_instance(*args, **kwargs)

    def import_data(self, *args: Any, **kwargs: Any) -> Any:
        """
        See :meth:`import_export.resources.Resource.import_data`
        """
        return super().import_data(*args, **kwargs)

    def import_row(self, *args: Any, **kwargs: Any) -> Any:
        """
        See :meth:`import_export.resources.Resource.import_row`
        """
        return super().import_row(*args, **kwargs)

    def save_instance(self, *args: Any, **kwargs: Any) -> Any:
        """
        See :meth:`import_export.resources.Resource.save_instance`
        """
        return super().save_instance(*args, **kwargs)

    def skip_row(self, *args: Any, **kwargs: Any) -> Any:
        """
        See :meth:`import_export.resources.Resource.skip_row`
        """
        return super().skip_row(*args, **kwargs)
```
Since the upstream documentation is just utterly broken, I wouldn't try to fix it on our side. Just link to the upstream docs, that's enough.

timobrembeck (changes requested):
Thanks a lot! :+1:

That's a very cool feature! Just needs a little bit of polishing, but in general it looks good!

integreat_cms/cms/views/feedback/region_feedback_actions.py
```diff
@@ -169,3 +173,113 @@ def delete_region_feedback(
...
 
     return redirect("region_feedback", region_slug=region_slug)
+
+
+# pylint: disable=too-few-public-methods
+class FeedbackResource(resources.ModelResource):
+    """
+    This is the Resource class that connects to the django-import-export library
+    """
+
```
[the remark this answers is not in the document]
JoeyStk: Makes sense! Should we also add the export for category then?

pyproject.toml
```diff
@@ -48,6 +48,7 @@ dependencies = [
     "django-cors-headers",
     "django-db-mutex",
     "django-debug-toolbar",
+    "django-import-export",
```
timobrembeck, line 51: Ah, and one thing I forgot: You have to execute the update dependency script to pin the version.

integreat_cms/cms/views/feedback/region_feedback_actions.py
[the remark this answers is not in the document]
timobrembeck: Yes, why not! :+1:

pyproject.toml
timobrembeck: You did not actually run the script, did you? Just appending the pinned version manually does not work, because the sub-dependencies are missing. `django-import-export` depends on `tablib` and `diff-match-patch` which are not yet installed. And those two might have even more sub-sub-dependencies, so it's always a bad idea to try to generate the pinned versions manually. Just execute the script.

timobrembeck (changes requested):

pyproject.toml
PeterNerlich: We did, and we saw that it introduced changes to about every second line, and thought that raising the versions of all dependencies to their newest available version wasn't really the atomic change we intended to make. We forgot sub-dependencies were a thing… :facepalm:

timobrembeck: I agree that it's not 100% optimal, but at the moment I don't see a way around it :see_no_evil: 
Writing a bash script which only pins the versions of the newly added dependency is possible but non-trivial I guess. At some point in the future I might have a look at this :grin:

JoeyStk: I ran it again and pushed the pyproject.toml file as it was updated

JoeyStk:
Thank you for your feedback. Hopefully I was able to resolve everything as requested. I will squash the commits at the end of the reviewing process to minimize the risk of messing something up

timobrembeck (commented):
> Hopefully I was able to resolve everything as requested.

Thanks a lot, looks good from my side :+1:

> I will squash the commits at the end of the reviewing process to minimize the risk of messing something up

Hmm, I think if you mess something up, it would be good if a reviewer could notice it. So in my opinion, you should squash now, and then I can approve the end result of what you want to merge :smile:

JoeyStk:
Sounds good. Done!

timobrembeck (approved):
Thanks! :tada:

If you want to attribute Peter's and Ulli's contribution to this PR, you can add the lines
```

Co-authored-by: Peter Nerlich <peter.nerlich@tuerantuer.org>
Co-authored-by: Ulli Holtgrave <ulli.holtgrave@tuerantuer.org>
```
to your commit message.

MizukiTemma (approved):
Looks good 😺 Thank you!

ulliholtgrave:
I guess we could also wait for my upstream patch to be released. I fixed a few of the issues and we might reduce our overwritten methods.

timobrembeck:
> I guess we could also wait for my upstream patch to be released. I fixed a few of the issues and we might reduce our overwritten methods.

They already have been resolved. So we just have to install the version 3.3.7, then we can remove the overwritten `skip_row()` method.

timobrembeck:
Since basically only the the resolving of the conflict and update of the dependency was missing, I updated the branch to be able to include this cool feature in today's release :tada:

Issue #1716, opened 2022-09-26
Deadlock during page move


timoludwig:
### Describe the Bug

In some edge cases, a deadlock can occur during moving pages.

### Steps to Reproduce

Not sure how to reproduce

### Expected Behavior

The page should be moved

### Actual Behavior

An internal server error occurs:
```
django.db.utils.OperationalError: deadlock detected
DETAIL:  Process 2031290 waits for ShareLock on transaction 13447834; blocked by process 2031285.
Process 2031285 waits for ShareLock on transaction 13447835; blocked by process 2031290.
HINT:  See server log for query details.
CONTEXT:  while updating tuple (2,47) in relation "cms_page"
```

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
    return self.cursor.execute(sql, params)
psycopg2.errors.DeadlockDetected: deadlock detected
DETAIL:  Process 2031290 waits for ShareLock on transaction 13447834; blocked by process 2031285.
Process 2031285 waits for ShareLock on transaction 13447835; blocked by process 2031290.
HINT:  See server log for query details.
CONTEXT:  while updating tuple (2,47) in relation "cms_page"


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/views/decorators/http.py", line 40, in inner
    return func(request, *args, **kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/usr/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/views/pages/page_actions.py", line 490, in move_page
    page.move(target, position)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/integreat_cms/cms/models/pages/page.py", line 303, in move
    super().move(target, pos)
[...]

PR #2596, opened 2023-12-29
Add deferrable Page tree constraints

svenseeberg:
### Short description
Add deferrable constraints for the page tree that do not obstruct the update operation of the Treebeard tree.

We tested this and it works :tada:. If the constraints are not deferrable, move operations break. If the constraints are deferrable, updates on trees work as expected.

### Proposed changes
- Add constraints to AbstractTreeNode
  - Add unique constraint for (tree_id, lft)
  - Add unique constraint for (tree_id, rgt)
  - Add check constraint to ensure smaller lft than rgt values

### Missing
- Front end error messages

### Resolved issues

Potentially fixes: #1518, #1716

Note to reviewers: ***You don't need to read the comments in this thread**, except maybe the very most recent ones. The earliest possibly helpful one is [this one](https://github.com/digitalfabrik/integreat-cms/pull/2596#issuecomment-2187069317) outlining a few assumptions at the core of the architecture of `@tree_mutex()`.*


svenseeberg:
Additional trigger constraints:

```sql
CREATE FUNCTION public.unique_combined_lft_rgt() RETURNS trigger
   LANGUAGE plpgsql AS
$$BEGIN
   IF (SELECT COUNT(combined) FROM (SELECT lft AS combined FROM cms_page
       WHERE tree_id=NEW.tree_id
       UNION
       SELECT rgt AS combined FROM cms_page
       WHERE tree_id=NEW.tree_id
      ) lr WHERE combined IN (NEW.lft, NEW.rgt)
      ) > 0
   THEN
      RAISE EXCEPTION 'lft or rgt value not unique.';
   END IF;
 
   RETURN OLD;
END;$$;
 
CREATE TRIGGER trigger_unique_combined_lft_rgt BEFORE UPDATE ON public.cms_page
   FOR EACH ROW EXECUTE PROCEDURE public.unique_combined_lft_rgt();


CREATE FUNCTION public.page_tree_limit_children() RETURNS trigger
   LANGUAGE plpgsql AS
$$BEGIN
   IF (SELECT (COUNT(*)+1)*2 FROM cms_page WHERE tree_id=NEW.tree_id AND lft>NEW.lft AND rgt<NEW.rgt) < (NEW.rgt-NEW.lft-1)
   THEN
      RAISE EXCEPTION 'The node has fewer children than indicated by the difference between lft and rgt values (one missing node is allowed!).';
   END IF;
   RETURN NEW;
END;$$;
 
CREATE TRIGGER trigger_page_tree_limit_children BEFORE UPDATE ON public.cms_page
   FOR EACH ROW EXECUTE PROCEDURE public.page_tree_limit_children();
```

svenseeberg:
Less restrictive working version:
```
CREATE FUNCTION public.unique_combined_lft_rgt() RETURNS trigger
   LANGUAGE plpgsql AS
$$BEGIN
   IF (OLD.tree_id=NEW.tree_id) AND
      ((SELECT COUNT(combined) FROM (SELECT lft AS combined FROM cms_page
       WHERE tree_id=NEW.tree_id
       UNION
       SELECT rgt AS combined FROM cms_page
       WHERE tree_id=NEW.tree_id
      ) lr WHERE combined IN (NEW.lft, NEW.rgt)
      ) > 2)
   THEN
      RAISE EXCEPTION 'lft or rgt value not unique enough.';
   END IF;
   RETURN NEW;
END;$$;
 
CREATE CONSTRAINT TRIGGER trigger_unique_combined_lft_rgt AFTER UPDATE ON public.cms_page DEFERRABLE INITIALLY DEFERRED
   FOR EACH ROW EXECUTE PROCEDURE public.unique_combined_lft_rgt();
```

PeterNerlich:
stress test to reproduce collisions:

`tools/integreat-cms-cli shell`

```python
from integreat_cms.cms.models import Page

i=0
alltag = Page.objects.get(id=21)
sprachkurse = Page.objects.get(id=20)

while True:
  alltag.move(sprachkurse, 'last-child')
  alltag.move(sprachkurse, 'right')
  print(f"i = {i}")
  i += 1
```
…and proceed to happily move pages about in the CMS

svenseeberg:
The constraints introduce issues with the repair_tree command as it is not yet build for updating all trees at once in one large transaction. And as Treebeard it will temporarily break the tree structure.

PeterNerlich:
mock `treebeard.models.Node._get_database_cursor()` within a custom `atomic` decorator (?) / context manager to always return the cursor of the connection of the currently wrapped function

integreat_cms/core/management/commands/find_database_inconsistencies.py
```diff
@@ -0,0 +1,125 @@
...
+                | Q(
+                    rgt__in=Subquery(
+                        model.objects.filter(tree_id=OuterRef("tree_id")).values_list(
+                            "lft", flat=True
+                        )
+                    )
+                )
+            )
+            if len(inconsistencies):
+                self.print_error(f"Oh oh, there are tree inconsistencies!")
```
PeterNerlich, line 72: *very important contribution*
```suggestion
                self.print_error(f"Uh oh, there are tree inconsistencies!")
```

integreat_cms/cms/utils/repair_tree.py
```diff
@@ -0,0 +1,259 @@
...
+            ) from e
+        root_node = None
+        for tree_node in mptt_fixer.get_fixed_tree_of_page(page.pk):
+            # the first node is always the root node
+            if root_node is None:
+                root_node = tree_node
+            action = "Fixing" if commit else "Detecting problems in"
+            printer.print(f"{action} tree with id {root_node.tree_id}...")
+            print_changed_fields(Page.objects.get(id=tree_node.pk), tree_node.lft, tree_node.rgt, printer)
+        check_for_orphans(root_node.tree_id, pages_seen, printer)
```
PeterNerlich, line 94: Record valid visited page as seen so it doesn't reported as orphan
```suggestion
            pages_seen.append(tree_node.pk)
        check_for_orphans(root_node.tree_id, pages_seen, printer)
```

integreat_cms/cms/utils/repair_tree.py
```diff
@@ -0,0 +1,259 @@
...
+    else:
+        for root_node in Page.objects.filter(lft=1):
+            action = "Fixing" if commit else "Detecting problems in"
+            printer.print(f"{action} tree with id {root_node.tree_id}...")
+            orphans.update(set(check_for_orphans(root_node.tree_id, pages_seen, printer)))
+
+        if commit:
+            for page in mptt_fixer.get_fixed_tree_nodes():
+                page.save()
```
PeterNerlich, line 103: same here
```suggestion
        fixed = mptt_fixer.get_fixed_tree_nodes()
        pages_seen.extend(map(lambda tree_node: tree_node.pk, fixed))
        for root_node in Page.objects.filter(lft=1):
            action = "Fixing" if commit else "Detecting problems in"
            printer.print(f"{action} tree with id {root_node.tree_id}...")
            orphans.update(set(check_for_orphans(root_node.tree_id, pages_seen, printer)))
 
        if commit:
            for page in fixed:
                page.save()
```

integreat_cms/cms/utils/repair_tree.py
```diff
@@ -0,0 +1,259 @@
...
+        node = self.fixed_nodes[parent.pk]
+        while node.parent:
+            self.fixed_nodes[parent.pk].rgt = node.rgt + 1
+            node = node.parent
+
+    def get_fixed_tree_nodes(self):
+        """
+        Yield all page tree nodes
+        """
+        yield self.fixed_nodes.values()
```
PeterNerlich, line 249: simply returning the `dict_values` object makes more sense, doesn't it? there are no values being yielded individually like this, which I guess was the intention
```suggestion
        return self.fixed_nodes.values()
```

integreat_cms/cms/utils/repair_tree.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def get_fixed_tree_of_page(self, node_id: int):
+        """
+        get all nodes of page tree, either identfied by one page or the (new) tree ID.
+        """
+        tree_id = self.fixed_nodes[node_id].tree_id
+        for node in self.fixed_nodes.values():
+            if node.tree_id == tree_id:
+                yield node
+        yield
```
PeterNerlich, line 259: this just produces None
```suggestion
```

PeterNerlich (changes requested):

PeterNerlich:
I had a crack at making `tree_mutex` be a transaction and inject its db cursor into treebeard, but it doesn't seem fully working yet: even though everything seems to work correctly now as far as I can tell, attempting a concurrent move operation causes one of the constraints to fail. Maybe this could be explained by the cache/db disparity we read about, I'm really not sure.
I pushed my changes to a separate branch for now: https://github.com/digitalfabrik/integreat-cms/compare/7bc796c...bugfix/treebeard_transaction

integreat_cms/cms/utils/repair_tree.py
```diff
@@ -0,0 +1,268 @@
...
+            return self.print
+        return self._write
+    @write.setter
+    def write(self, new):
+        self._write = new
+
+
+@transaction.atomic
+@tree_mutex
+def repair_tree(page_id: int = 0, commit: bool = False, printer: Printer = Printer(), fix_orphans: bool = True) -> None:
```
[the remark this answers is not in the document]
svenseeberg: AFAICT Django starts always with pk=1.

integreat_cms/cms/models/abstract_tree_node.py
```diff
@@ -230,41 +228,29 @@ def move(self, target: AbstractTreeNode, pos: str | None = None) -> None:
         :param pos: The new position of the page relative to the target
                     (choices: :mod:`~integreat_cms.cms.constants.position`)
         :raises ~treebeard.exceptions.InvalidPosition: If the node is moved to another region
-        :raises ~db_mutex.exceptions.DBMutexError: If the DB mutex could not be retrieved
-        :raises ~db_mutex.exceptions.DBMutexTimeoutError: If waiting for the DB mutex timed out
         """
         logger.debug("Moving %r to position %r of %r", self, pos, target)
-        try:
```
svenseeberg, line 237: We do not need the mutex here anymore, right? We directly implement them in the views.

svenseeberg:
> Maybe also risk a look at [b40e283...bugfix/treebeard_transaction](https://github.com/digitalfabrik/integreat-cms/compare/b40e283...bugfix/treebeard_transaction) – I rebased it just now onto these latest changes. I didn't test it though. Have a look through the mutex at least, I think I saw a few improvements there (e.g. I moved the uuid for some reason, I think it was critical for the mutex to work correctly)

Do you want to open a PR for that? In general this seems solid.

svenseeberg:
Ah never mind, I will just include the commit :D

integreat_cms/cms/views/pages/page_actions.py
```diff
@@ -42,6 +45,8 @@
 
 
 @require_POST
+@transaction.atomic
+@tree_mutex("page")
```
david-venhoff, line 49: That should probably also be added to `page_bulk_actions.py`, right?

integreat_cms/cms/utils/repair_tree.py
```diff
@@ -0,0 +1,284 @@
...
+    to fix hierarchy and sorts siblings by (potentially inconsistent) lft.
+    """
+
+    def __init__(self) -> None:
+        """
+        Creates a fixed tree when initializing class but does not save results
+        """
+        self.broken_nodes: list[Page] = list(
+            Page.objects.all().order_by("tree_id", "lft")
+        )
```
timobrembeck, line 194:
```suggestion
        self.broken_root_nodes: list[Page] = list(
            Page.objects.filter(parent=None)
        )
        self.broken_child_nodes: list[Page] = list(
            Page.objects.exclude(parent=None)
        )
```
If those subsets are only handled separately, I think it makes sense to store them independent of each other instead of using the loops later.
Also, the ordering is like this by default, so no need to state it explicitly.

integreat_cms/cms/utils/repair_tree.py
```diff
@@ -0,0 +1,284 @@
...
+        tree_counter = 1
+        for node in self.broken_nodes:
+            if not node.parent_id:
+                node.lft = 1
+                node.rgt = 2
+                node.depth = 1
+                node.fixed_children = []
+                node.tree_id = tree_counter
+                tree_counter = tree_counter + 1
+                self.fixed_nodes[node.pk] = node
```
timobrembeck, line 212:
```suggestion
        for tree_counter, node in enumerate(self.broken_root_nodes, 1):
            node.lft = 1
            node.rgt = 2
            node.depth = 1
            node.fixed_children = []
            node.tree_id = tree_counter
            self.fixed_nodes[node.pk] = node
```
Incrementing counters manually is oldschool :grin:

integreat_cms/cms/utils/repair_tree.py
```diff
@@ -0,0 +1,284 @@
...
+        Get all remaining (child) nodes, add add them to the new/fixed tree
+        """
+        for node in self.broken_nodes:
+            if node.parent_id:
+                parent = self.fixed_nodes[node.parent_id]
+                node.fixed_children = []
+                node = self.calculate_lft_rgt(node, parent)
+                self.fixed_nodes[node.pk] = node
+                self.fixed_nodes[parent.pk].fixed_children.append(node.pk)
+                self.update_ancestors_rgt(node.pk)
```
timobrembeck, line 225:
```suggestion
        for node in self.broken_child_nodes:
            parent = self.fixed_nodes[node.parent_id]
            node.fixed_children = []
            node = self.calculate_lft_rgt(node, parent)
            self.fixed_nodes[node.pk] = node
            self.fixed_nodes[parent.pk].fixed_children.append(node.pk)
            self.update_ancestors_rgt(node.pk)
```

integreat_cms/cms/utils/repair_tree.py
```diff
@@ -0,0 +1,284 @@
...
+                tree_counter = tree_counter + 1
+                self.fixed_nodes[node.pk] = node
+
+    def fix_child_nodes(self) -> None:
+        """
+        Get all remaining (child) nodes, add add them to the new/fixed tree
+        """
+        for node in self.broken_nodes:
+            if node.parent_id:
+                parent = self.fixed_nodes[node.parent_id]
```
timobrembeck, line 220: If the tree structure is broken, we cannot assume that the parent's `lft` value is necessarily smaller than the node's, so this might throw a `KeyError` if the parent is not yet fixed.

integreat_cms/cms/utils/repair_tree.py
```diff
@@ -0,0 +1,284 @@
...
+from ..models import Page
+from ..utils.tree_mutex import tree_mutex
+
+logger = logging.getLogger(__name__)
+
+
+class Printer:
+    """
+    Select printer for stdout or log file
+    """
```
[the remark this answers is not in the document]
PeterNerlich: Only because you jumped at it while we were still working on it :slightly_smiling_face: You're probably right, but I think we'll try to actually get it in working condition first

PeterNerlich:
@timobrembeck I cannot figure out how to capture the command output in `tests/core/management/commands/test_repair_tree.py` *(btw. ignore `tests/cms/models/pages/tree_mutex.py`, it is a stub that we just copied, as you can see)* – the management command is run successfully and I see its output in pytests `Captured stdout call`, but apparently using the following to capture it doesn't work. I don't think it makes much sense to use the log fixture for it as we do in other tests, because logically we don't care which way the messages take, only that they eventually end up in the terminal stdout when we call the command.

```python
def get_command_output(command: str, *args: str, **kwargs: Any) -> tuple[str, str]:
    out = StringIO()
    err = StringIO()
    call_command(
        command,
        *args,
        **kwargs,
        stdout=out,
        stderr=err,
    )
    return out.getvalue(), err.getvalue()
```

Do you have any idea?

timobrembeck:
@PeterNerlich No sorry, at the first glance I don't see the reason for this, but I assume it has something to do with the custom `Printer` class. In my opinion, we should remove it to reduce complexity and use consistent logging behavior in all management commands.

PeterNerlich:
Then I guess I'll try that next then. Though if it works as I expect it to, it should be rather transparent. It just hands the strings to the desired function, be it from logging or otherwise.

PeterNerlich:
Unfortunately, it's still the same

timobrembeck:
> Unfortunately, it's still the same

As described [here](https://github.com/digitalfabrik/integreat-cms/pull/2596#discussion_r1509938529), we have to log to the logger `integreat_cms.core.management.commands` in order to get the logging to stdout because it's defined here:
https://github.com/digitalfabrik/integreat-cms/blob/41b995e1304285c2d6853813a3168b672af840db/integreat_cms/core/settings.py#L685-L692

PeterNerlich:
Okay, I had to re-run the CI pipeline a few times until it went through. At least twice I had a nonsensical failure (not finding the region Augsburg for example) – I fear that the introduced mutex might not be completely innocent here, though I'd strongly expect a different result than the object being treated as not existing in the database.

tests/cms/models/pages/tree_mutex.py
```diff
@@ -0,0 +1,42 @@
...
+            week_for_monthly=None,
+            recurrence_end_date=None,
+        )
+        test_event = Event(
+            start=datetime.datetime(2020, 1, 1, 11, 30, 0, 0, ZoneInfo("UTC")),
+            end=datetime.datetime(2030, 1, 1, 12, 30, 0, 0, ZoneInfo("UTC")),
+        )
+        test_event.recurrence_rule = recurrence_rule
+        ical_rrule = recurrence_rule.to_ical_rrule_string()
+        assert ical_rrule == "DTSTART:20200101T113000\nRRULE:FREQ=YEARLY"
```
PeterNerlich, line 42: TODO: replace with actual test of `tree_mutex`

PeterNerlich:
could this maybe solve #1716?

PeterNerlich:
@timobrembeck Weird stuff is happening making tests fail and I have no clue why. `tests/api/test_api_push_page.py::test_api_push_page_content - assert 404 == 403`
Can you have a look at whether this PR is at fault or something else is going on?

PeterNerlich:
I'd still like to add a slightly more complex test to `tests/cms/utils/test_repair_tree.py`, but with the two tests for `@tree_mutex()` I finally think that this is ready to leave the draft status! :tada:

PeterNerlich:
Tests seemingly were a success this time, but I'm not confident that my last commit regarding that is 100% effective. It also needs cleanup for sure and likely should be refactored, even for such a band-aid solution.

PeterNerlich:
Even if I'm still investigating why the test don't get along as they should, I think the following log from `test_tree_mutex` is already very indicative of the effectiveness of `tree_mutex` (especially considering that `test_rule_out_false_positive` which does exactly the same thing but not using the mutex provokes a collision immediately in all of my observation so far)

<details><summary>test stdout of log messages from the mutex and patched cursor function</summary>

```
starting threads…
running 5-10-5 on contestant #21 with tree_mutex
    [#21] 0
running 5-10-5 on contestant #19 with tree_mutex
    [#19] 0
  Failed to acquire MUTEX_PAGE_TREE as 20570cbf-2cd3-43d5-aecf-c1b7092c99c8: MUTEX_page_TREE present (e04f1088-6fd0-492c-bd48-3048ef06bf6c). Waiting 0.1s…
   moving contestant #19…
someone is getting our monkeypatched db cursor (default)! <class 'integreat_cms.cms.models.pages.page.Page'>, write
  Failed to acquire MUTEX_PAGE_TREE as 20570cbf-2cd3-43d5-aecf-c1b7092c99c8: MUTEX_page_TREE present (e04f1088-6fd0-492c-bd48-3048ef06bf6c). Waiting 0.1s…
OK moving contestant #19!
  Releasing MUTEX_PAGE_TREE after 0.1849055290222168s (e04f1088-6fd0-492c-bd48-3048ef06bf6c)
   moving back contestant #19…
  Failed to acquire MUTEX_PAGE_TREE as 20570cbf-2cd3-43d5-aecf-c1b7092c99c8: MUTEX_page_TREE present (7ab36bba-e55a-4892-b50b-9ead0fdf26a7). Waiting 0.1s…
someone is getting our monkeypatched db cursor (default)! <class 'integreat_cms.cms.models.pages.page.Page'>, write
OK moving back contestant #19!
  Releasing MUTEX_PAGE_TREE after 0.11504745483398438s (7ab36bba-e55a-4892-b50b-9ead0fdf26a7)
    [#19] 1
   moving contestant #19…
  Failed to acquire MUTEX_PAGE_TREE as 20570cbf-2cd3-43d5-aecf-c1b7092c99c8: MUTEX_page_TREE present (1b4e1e43-b8f3-4323-89c7-d4887ba566e9). Waiting 0.1s…
someone is getting our monkeypatched db cursor (default)! <class 'integreat_cms.cms.models.pages.page.Page'>, write
OK moving contestant #19!
  Releasing MUTEX_PAGE_TREE after 0.10081100463867188s (1b4e1e43-b8f3-4323-89c7-d4887ba566e9)
   moving back contestant #19…
  Failed to acquire MUTEX_PAGE_TREE as 20570cbf-2cd3-43d5-aecf-c1b7092c99c8: MUTEX_page_TREE present (13bf03cf-20c2-4e6e-8d53-d443419e9ac3). Waiting 0.1s…
someone is getting our monkeypatched db cursor (default)! <class 'integreat_cms.cms.models.pages.page.Page'>, write
  Failed to acquire MUTEX_PAGE_TREE as 20570cbf-2cd3-43d5-aecf-c1b7092c99c8: MUTEX_page_TREE present (13bf03cf-20c2-4e6e-8d53-d443419e9ac3). Waiting 0.1s…
OK moving back contestant #19!
  Releasing MUTEX_PAGE_TREE after 0.14224028587341309s (13bf03cf-20c2-4e6e-8d53-d443419e9ac3)
    [#19] 2
   moving contestant #19…
someone is getting our monkeypatched db cursor (default)! <class 'integreat_cms.cms.models.pages.page.Page'>, write
  Failed to acquire MUTEX_PAGE_TREE as 20570cbf-2cd3-43d5-aecf-c1b7092c99c8: MUTEX_page_TREE present (0bc106e0-d09a-40ca-95bc-d7ff90e027cd). Waiting 0.1s…
OK moving contestant #19!
  Releasing MUTEX_PAGE_TREE after 0.15320730209350586s (0bc106e0-d09a-40ca-95bc-d7ff90e027cd)
   moving back contestant #19…
  Failed to acquire MUTEX_PAGE_TREE as 20570cbf-2cd3-43d5-aecf-c1b7092c99c8: MUTEX_page_TREE present (062407b2-4e86-4d72-9ea8-987059b19628). Waiting 0.1s…
someone is getting our monkeypatched db cursor (default)! <class 'integreat_cms.cms.models.pages.page.Page'>, write
OK moving back contestant #19!
  Releasing MUTEX_PAGE_TREE after 0.12149930000305176s (062407b2-4e86-4d72-9ea8-987059b19628)
    [#19] 3
   moving contestant #19…
  Failed to acquire MUTEX_PAGE_TREE as 20570cbf-2cd3-43d5-aecf-c1b7092c99c8: MUTEX_page_TREE present (893a6f38-82ff-449f-a3f8-676d901cca7b). Waiting 0.1s…
someone is getting our monkeypatched db cursor (default)! <class 'integreat_cms.cms.models.pages.page.Page'>, write
OK moving contestant #19!
  Releasing MUTEX_PAGE_TREE after 0.09466361999511719s (893a6f38-82ff-449f-a3f8-676d901cca7b)
   moving back contestant #19…
  Failed to acquire MUTEX_PAGE_TREE as 20570cbf-2cd3-43d5-aecf-c1b7092c99c8: MUTEX_page_TREE present (217eea57-26d7-4a00-ada7-ac5ad51fdbe9). Waiting 0.1s…
someone is getting our monkeypatched db cursor (default)! <class 'integreat_cms.cms.models.pages.page.Page'>, write
OK moving back contestant #19!
  Releasing MUTEX_PAGE_TREE after 0.11383414268493652s (217eea57-26d7-4a00-ada7-ac5ad51fdbe9)
    [#19] 4
   moving contestant #19…
  Failed to acquire MUTEX_PAGE_TREE as 20570cbf-2cd3-43d5-aecf-c1b7092c99c8: MUTEX_page_TREE present (883f0766-1eac-45ad-a5ec-70b68a052280). Waiting 0.1s…
someone is getting our monkeypatched db cursor (default)! <class 'integreat_cms.cms.models.pages.page.Page'>, write
  Failed to acquire MUTEX_PAGE_TREE as 20570cbf-2cd3-43d5-aecf-c1b7092c99c8: MUTEX_page_TREE present (883f0766-1eac-45ad-a5ec-70b68a052280). Waiting 0.1s…
OK moving contestant #19!
  Releasing MUTEX_PAGE_TREE after 0.21114397048950195s (883f0766-1eac-45ad-a5ec-70b68a052280)
   moving back contestant #19…
  Failed to acquire MUTEX_PAGE_TREE as 20570cbf-2cd3-43d5-aecf-c1b7092c99c8: MUTEX_page_TREE present (78c19334-ad1c-4b61-b45b-02139e74ea89). Waiting 0.1s…
someone is getting our monkeypatched db cursor (default)! <class 'integreat_cms.cms.models.pages.page.Page'>, write
OK moving back contestant #19!
  Failed to acquire MUTEX_PAGE_TREE as 20570cbf-2cd3-43d5-aecf-c1b7092c99c8: MUTEX_page_TREE present (78c19334-ad1c-4b61-b45b-02139e74ea89). Waiting 0.1s…
  Releasing MUTEX_PAGE_TREE after 0.11609148979187012s (78c19334-ad1c-4b61-b45b-02139e74ea89)
finished 5-10-5 of contestant #19 with tree_mutex
   moving contestant #21…
someone is getting our monkeypatched db cursor (default)! <class 'integreat_cms.cms.models.pages.page.Page'>, write
OK moving contestant #21!
  Releasing MUTEX_PAGE_TREE after 1.5792932510375977s (20570cbf-2cd3-43d5-aecf-c1b7092c99c8)
   moving back contestant #21…
someone is getting our monkeypatched db cursor (default)! <class 'integreat_cms.cms.models.pages.page.Page'>, write
OK moving back contestant #21!
  Releasing MUTEX_PAGE_TREE after 0.11186051368713379s (eaef70a7-b8f1-4a84-9a59-ad880b7f2c21)
    [#21] 1
   moving contestant #21…
someone is getting our monkeypatched db cursor (default)! <class 'integreat_cms.cms.models.pages.page.Page'>, write
OK moving contestant #21!
  Releasing MUTEX_PAGE_TREE after 0.14497041702270508s (f7c1c7c6-1422-4b10-95ce-a21af933090d)
   moving back contestant #21…
someone is getting our monkeypatched db cursor (default)! <class 'integreat_cms.cms.models.pages.page.Page'>, write
OK moving back contestant #21!
  Releasing MUTEX_PAGE_TREE after 0.20409369468688965s (77765b0e-4d76-4a83-a29e-5a2dfe91beba)
    [#21] 2
   moving contestant #21…
someone is getting our monkeypatched db cursor (default)! <class 'integreat_cms.cms.models.pages.page.Page'>, write
OK moving contestant #21!
  Releasing MUTEX_PAGE_TREE after 0.15691709518432617s (f4ea6dd6-62ea-4857-b0cf-8ff401d9a794)
   moving back contestant #21…
someone is getting our monkeypatched db cursor (default)! <class 'integreat_cms.cms.models.pages.page.Page'>, write
OK moving back contestant #21!
  Releasing MUTEX_PAGE_TREE after 0.1440751552581787s (ab41fc5e-95fd-4ad4-9809-d5edba714965)
    [#21] 3
   moving contestant #21…
someone is getting our monkeypatched db cursor (default)! <class 'integreat_cms.cms.models.pages.page.Page'>, write
OK moving contestant #21!
  Releasing MUTEX_PAGE_TREE after 0.1037137508392334s (3e7a11a3-3c02-45f9-bcc4-d2a6d532be3f)
   moving back contestant #21…
someone is getting our monkeypatched db cursor (default)! <class 'integreat_cms.cms.models.pages.page.Page'>, write
OK moving back contestant #21!
  Releasing MUTEX_PAGE_TREE after 0.17049074172973633s (53225650-6062-4861-a995-79b5e137f58d)
    [#21] 4
   moving contestant #21…
someone is getting our monkeypatched db cursor (default)! <class 'integreat_cms.cms.models.pages.page.Page'>, write
OK moving contestant #21!
  Releasing MUTEX_PAGE_TREE after 0.13613367080688477s (de91fa77-6826-4026-8f23-2f28dd279c5a)
   moving back contestant #21…
someone is getting our monkeypatched db cursor (default)! <class 'integreat_cms.cms.models.pages.page.Page'>, write
OK moving back contestant #21!
  Releasing MUTEX_PAGE_TREE after 0.11805582046508789s (109eb81d-eb78-4129-98bb-83bcc71b2647)
finished 5-10-5 of contestant #21 with tree_mutex
joined threads!
```

</details>

PeterNerlich:
Wait, we decorated a few functions with `@tree_mutex()`. But since e.g. to move a page we catch a few exceptions. Doesn't that mean that these never reach `@tree_mutex()`, meaning that the transaction is committed instead of being rolled back?

svenseeberg:
> Wait, we decorated a few functions with `@tree_mutex()`. But since e.g. to move a page we catch a few exceptions. Doesn't that mean that these never reach `@tree_mutex()`, meaning that the transaction is committed instead of being rolled back?

AFAICT no? Should be easy to test though :D

PeterNerlich:
New thought – I think I found a flaw in my thinking which could solve this. Please correct my assumptions:

1. We can successfully force **treebeard** to use transactions by **overwriting its internal function** to get a database cursor to give a prepared cursor that opened a transaction
2. As long as no other treebeard function is called from other contexts during that call, everything works fine. But if there are concurrent calls to treebeard from **contexts that share the memory for the imports**, they wrongly also use the transaction
3. Because **WSGI spawns individual python processes**, we **cannot use** locks to fence the code or **shared data structures** to govern which db cursor to hand out
4. The attempt to use **`traceback.extract_stack()`** to discern contexts from each other seems **insufficient**, as two concurrent calls of the function could possess indistinguishable call stacks
5. The attempt to use **`cache.get_or_set()`** to implement a custom mutex using the cache seems **insufficient**, as the assumption of *get if exists else set* action to be atomic proves to be false
6. New: *To isolate treebeard calls, we do not need a mutex that works across processes. It should be sufficient to **rely on locks within contexts with shared memory**, where they are useful, while **different processes will have a separate state of treebeard and thus a separate db cursor anyway**.*

svenseeberg:
> New thought – I think I found a flaw in my thinking which could solve this. Please correct my assumptions:
> 
>     1. We can successfully force **treebeard** to use transactions by **overwriting its internal function** to get a database cursor to give a prepared cursor that opened a transaction

:+1: 

>     2. As long as no other treebeard function is called from other contexts during that call, everything works fine. But if there are concurrent calls to treebeard from **contexts that share the memory for the imports**, they wrongly also use the transaction – _we have to either_ prevent concurrent calls with a lock _or_ hand out a different db cursor depending on the detected context

When would this happen? It does not sound very likely to me at first glance that this would actually happen. We do not have threading enabled. https://modwsgi.readthedocs.io/en/master/user-guides/processes-and-threading.html states that "concurrent requests [are] being handled in distinct threads executing within those processes".

>     3. Because **WSGI spawns individual python processes**, we **cannot use** locks to fence the code or **shared data structures** to govern which db cursor to hand out

:+1: 

>     4. The attempt to use **`traceback.extract_stack()`** to discern contexts from each other seems **insufficient**, as two concurrent calls of the function could possess indistinguishable call stacks

:+1: 

>     5. The attempt to use **`cache.get_or_set()`** to implement a custom mutex using the cache seems **insufficient**, as the assumption of _get if exists else set_ action to be atomic proves to be false – _there are race conditions where this is overwritten because the cache entry seemed entry, while another function put their value in in the meantime_

:+1: 

>     6. New: _To isolate treebeard calls, we do not need a mutex that works across processes. It should be sufficient to **rely on locks within contexts with shared memory**, where they are useful, while **different processes will have a separate state of treebeard and thus a separate db cursor anyway**._

This is a totally different root cause than we initially want to address? We currently have an issue with multiple processes continually rewriting the table while others are doing the same.

PeterNerlich:
> > 2. […] But if there are concurrent calls to treebeard from **contexts that share the memory for the imports**, they wrongly also use the transaction
> 
> When would this happen? […] We do not have threading enabled. […]

Maybe it's my fault for trying to consider these cases at all and making it even more complicated. 2. mainly reflects my mental model of what modifying properties of imports means for us, and I used threading in the tests to provoke and prevent race conditions.

> We assumed that have an issue with multiple processes continually rewriting the table while others are doing the same.

Yes, this was the original issue, before all the issues that arose while trying to fix it. The solution is to let the database place make a transaction. Due to our hacky fix it's just difficult to not make everything else use that very transaction in the meantime as well.

svenseeberg:
> Yes, this was the original issue, before all the issues that arose while trying to fix it. The solution is to let the database place make a transaction. Due to our hacky fix it's just difficult to not make everything else use that very transaction in the meantime as well.

Agreed. But AFAICT this is not an issue, as long as it is the same process that would create the same modifications to the database anyways. So maybe this is even wanted?

PeterNerlich:
I have to read through the diff again with a rested head, but I think I more or less got it to a reasonable state. Again. Except for more git history cleanup and one failing test ([→details](https://github.com/digitalfabrik/integreat-cms/pull/2919)).

integreat_cms/cms/models/languages/language_tree_node.py
```diff
@@ -161,7 +161,7 @@ def get_repr(self) -> str:
             f"<LanguageTreeNode (id: {self.id}{language_str}{parent_str}{region_str})>"
         )
 
-    class Meta:
+    class Meta(AbstractTreeNode.Meta):
```
PeterNerlich, line 164: The `Meta` class of models derived from `AbstractTreeNode` have to themselves descend from `AbstractTreeNode.Meta` in order to inherit the constraints.
Maybe we should add something to `AbstractTreeNode` that checks whether this is the case for any class derived from it and print a warning otherwise? Sounds hacky and would have to be executed once on subclass definition, not each time an instance is created :confused:

PeterNerlich:
@svenseeberg We originally started with finding constraints that break when the tree is broken in multiple degrees of freedom, giving just enough leeway for treebeard to do its efficient operations that temporarily break the tree a little.
If we managed to make treebeard use transactions, shouldn't we be able to create strict constraints now because they will be deferred to when the treebeard operations are done? *(=when our transaction is closed at the end of our context)*

tests/core/management/commands/test_repair_tree.py
```diff
@@ -45,6 +46,7 @@ def test_check_broken_tree_fields(load_test_data: None) -> None:
     assert "lft: 2 → 1" in err
 
 
+@pytest.mark.skip(reason="Broken tree caught by deferrable tree constraints")
 @pytest.mark.django_db
 def test_fix_broken_tree_fields(load_test_data: None) -> None:
```
david-venhoff, line 51: Is there a reason to keep these tests?

david-venhoff (approved):
The changes by this pr are straightforward and should help without causing any problems, so I think its finally time to merge this :tada:

Only question I have is why are we adding the constraints to the abstract tree node instead of treebeard directly (Or phrased differently, is there a plan to merge this upstream at some point)?

PeterNerlich:
> Only question I have is why are we adding the constraints to the abstract tree node instead of treebeard directly (Or phrased differently, is there a plan to merge this upstream at some point)?

I don't remember everything we considered at the time, but we decided at least about getting treebeard to use transactions internally was vanishingly improbable, so we saw hacking it "externally" like this was the only way with a viable value for effort ratio.

One problem with the constraints is that [any subclasses that define their own meta class need to have it inherit the parents meta class](https://github.com/digitalfabrik/integreat-cms/pull/2596/files#r1685816732), which you have to be aware of – otherwise they won't apply. Normally just defining a new meta class works, and indeed here it works too – there is no indication that something important is missing.
Though obviously it's debatable whether that's on the user, but I at least find it very confusing, even if logical when you take long enough to think about what is actually happening behind the scenes.

tests/core/management/commands/test_repair_tree.py
PeterNerlich: Yes, they will be re-enabled in #3075. I just skipped them here while dividing this PR into three because the tests failed in this kind of intermediate step.

MizukiTemma (approved):
Thank you so much 🚀

PR #2600, opened 2024-01-04
Write all management command output to logs

timobrembeck:
### Short description

In January, the cronjob for resetting the DeepL budget apparently did not work.
I noticed that its output is only written to stdout and not to the logfile, so I thought it would make sense to change this behavior, and while I'm at it for the other management commands as well.

### Proposed changes

- Use colored logging in CircleCI
- Write all management command output to logs

### Side effects

- It's not possible anymore to print something in white and/or bold in the output of management commands.

### DIFF
--- a/docs/src/management-commands.rst
+++ b/docs/src/management-commands.rst
@@ -190,7 +190,7 @@ When adding new custom commands, you can use the base classes:
 :class:`~integreat_cms.core.management.log_command.LogCommand`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-A base class to improve the logging output of management commands
+A base class for management commands to set the stream handler of the logger to the command's stdout wrapper
 
 
 :class:`~integreat_cms.core.management.debug_command.DebugCommand`
--- a/integreat_cms/core/apps.py
+++ b/integreat_cms/core/apps.py
@@ -6,10 +6,11 @@ import sys
 from typing import TYPE_CHECKING
 
 from django.apps import AppConfig
+from django.contrib.messages.constants import SUCCESS
 from django.utils.translation import gettext_lazy as _
 
 if TYPE_CHECKING:
-    from typing import Final
+    from typing import Any, Final
 
     from django.utils.functional import Promise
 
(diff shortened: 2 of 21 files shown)


ulliholtgrave (approved):
Thanks. This is much cleaner and might help us with future debug effort :+1:
Initially, I was a little bit confused about the SUCCESS logging-level which isn't listed at https://docs.python.org/3/library/logging.html#logging-levels but my manual tests worked fine with it :man_shrugging:

timobrembeck:
> Initially, I was a little bit confused about the SUCCESS logging-level which isn't listed at https://docs.python.org/3/library/logging.html#logging-levels but my manual tests worked fine with it 🤷‍♂️

The `SUCCESS` level is used by the Django messaging framework and defined as `25`:
https://github.com/django/django/blob/cc56c22a24ef717cc3111e92ca146136fa518d55/django/contrib/messages/constants.py#L3

And added as custom logging level here:
https://github.com/digitalfabrik/integreat-cms/blob/ff4775c1f837832c5db781fc46c7bd1d531ca936/integreat_cms/core/storages.py#L14-L15

not sure whether this is the right place though, should I move it to the settings file?

charludo (changes requested):
> - Use colored logging in CircleCI
😍 

Adapted from [this](https://stackoverflow.com/a/35804945/1691778) SO question and answer:
```python
import logging


def logForLevel(self, message, *args, **kwargs):
    if self.isEnabledFor(25):
        self._log(25, message, args, **kwargs)


def logToRoot(message, *args, **kwargs):
    logging.log(25, message, *args, **kwargs)


logging.addLevelName(25, "SUCCESS")
setattr(logging, "SUCCESS", 25)
setattr(logging.getLoggerClass(), "success", logForLevel)
setattr(logging, "success", logToRoot)
```
Pasting this right after the imports of the `settings.py` (prob not the right place, but still) works in allowing us to just do 
`logger.success("message")`
instead of the slightly unintuitive 
`logger.log(SUCCESS, "message")` (plus the preceding import of SUCCESS)

I think this could even be beneficial beyond usage in the management commands.

timobrembeck:
@charludo Cool idea! :+1:
The only downside of this approach is that mypy doesn't seem to be recognizing the custom logging function ([others seem to have the same problem](https://github.com/python/typing/discussions/980)), so for now we only can use it in combination with an ignore comment. But ok, maybe this changes in the future.

charludo:
> @charludo Cool idea! 👍 The only downside of this approach is that mypy doesn't seem to be recognizing the custom logging function ([others seem to have the same problem](https://github.com/python/typing/discussions/980)), so for now we only can use it in combination with an ignore comment. But ok, maybe this changes in the future.

Ah, dang, that is also not ideal. In that case I'm OK with both solutions, let's just go with what you prefer.

charludo (approved):

Issue #2585, opened 2023-12-06
Fix the error message when deleting a media file in use


seluianova:
### Describe the Bug
When deleting a media file that is in use, a network error is displayed
![image](https://github.com/digitalfabrik/integreat-cms/assets/115008338/7a7371aa-ff98-4fe8-88a9-eccd5659b3e0)

### Steps to Reproduce

1. Go to 'Media Library'
2. Try to delete a media file that is being used as the organisation icon
3. See error

### Expected Behavior
I think we could display a more understandable (user-friendly) message.
Like: "This media file can't be deleted because it is in use"

### Actual Behavior
The following error is displayed:
A network error has occurred. Please try again later.

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
Dec 06 11:32:21 ERROR django.request - 500 Internal Server Error: /ajax/media/file/delete/
Traceback (most recent call last):
  File "/home/tory/Integreat/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/tory/Integreat/integreat-cms/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/tory/Integreat/integreat-cms/.venv/lib/python3.9/site-packages/django/views/decorators/http.py", line 40, in inner
    return func(request, *args, **kwargs)
  File "/home/tory/Integreat/integreat-cms/.venv/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/home/tory/Integreat/integreat-cms/integreat_cms/api/decorators.py", line 115, in wrap
    return function(request, *args, **kwargs)
  File "/home/tory/Integreat/integreat-cms/integreat_cms/cms/views/media/media_actions.py", line 361, in delete_file_ajax
    media_file.delete()
  File "/home/tory/Integreat/integreat-cms/.venv/lib/python3.9/site-packages/django/db/models/base.py", line 966, in delete
    collector.collect([self], keep_parents=keep_parents)
  File "/home/tory/Integreat/integreat-cms/.venv/lib/python3.9/site-packages/django/db/models/deletion.py", line 302, in collect
    raise ProtectedError(
[...]

PR #2601, opened 2024-01-05
Disable delete button for used media files and improve error message

MizukiTemma:
### Short description

This PR disables the delete button if the media file is in use and improve as back up the error message which is shown when users try to delete an used media file,

### Proposed changes

- Disable the delete button for used media files
- Improve the error message in case a media file in use is about to be deleted.

### Side effects

- The delete button is disabled in another way than I suggested in the issue. In my opinion it's better to disable it in the same way as it is disabled for non-empty directories.

### Resolved issues

Fixes: #2585 

### DIFF
--- a/integreat_cms/cms/views/media/media_actions.py
+++ b/integreat_cms/cms/views/media/media_actions.py
@@ -354,6 +354,22 @@ def delete_file_ajax(
         MediaFile.objects.filter(region=region), id=request.POST.get("id")
     )
 
+    # Check if the media file is in use
+    if media_file.is_used:
+        return JsonResponse(
+            {
+                "messages": [
+                    {
+                        "type": "warning",
+                        "text": _(
+                            'File "{}" cannot be deleted because it is used as an icon or content.'
+                        ).format(media_file.name),
+                    }
+                ],
+            },
+            status=400,
+        )
+
     # Delete corresponding physical files
     media_file.file.delete()
     media_file.thumbnail.delete()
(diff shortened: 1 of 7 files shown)


integreat_cms/cms/views/media/media_actions.py
```diff
@@ -354,6 +354,22 @@ def delete_file_ajax(
...
 
+    # Check if the media file is in use
+    if media_file.is_used:
+        return JsonResponse(
+            {
+                "messages": [
+                    {
+                        "type": "warning",
+                        "text": _(
+                            'File "{}" cannot be deleted because it is used as icon or content.'
```
charludo, line 365:
```suggestion
                            'File "{}" cannot be deleted because it is used as an icon or content.'
```
"als Icon" works perfectly fine in German, but in English you need an indefinite article here, I am 99% sure

charludo (approved):
Very nice, thank you!! 🎉

integreat_cms/cms/views/media/media_context_mixin.py
```diff
@@ -43,6 +43,7 @@ def get_context_data(self, **kwargs: Any) -> dict[str, Any]:
                 "btn_upload_file": _("Upload file"),
                 "btn_save_file": _("Save file"),
                 "btn_delete_file": _("Delete file"),
+                "btn_delete_used_file": _("You can only delete unsed media files"),
```
seluianova, line 46:
```suggestion
                "btn_delete_used_file": _("You can only delete unused media files"),
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -8521,6 +8527,10 @@ msgstr "Datei speichern"
 msgid "Delete file"
 msgstr "Datei löschen"
 
+#: cms/views/media/media_context_mixin.py
+msgid "You can only delete unsed media files"
```
seluianova, line 8531:
```suggestion
msgid "You can only delete unused media files"
```

seluianova (approved):
Noted another typo, otherwise all good!

Issue #2290, opened 2023-05-17
New API URL patterns not versioned


timobrembeck:
### Describe the Bug

At the moment, we only add the version number `v3` for the legacy API endpoint patterns from wordpress, but not for the new ones.
On order for the app to migrate to the new patterns, we need to prefix them with the current version.

### Steps to Reproduce

1. Go to http://localhost:8000/api/regions/
2. See API result
### Expected Behavior

The API result should be available under http://localhost:8000/api/v3/regions/

### Actual Behavior

There is no version parameter

steffenkleinle:
Please notify the app team once this is done: https://issues.tuerantuer.org/browse/IGAPP-1317

PR #2617, opened 2024-01-23
Add version to api url patterns

sarahsporck:
### Short description
This PR adds a version path element to the new endpoints, so that the app & webapp can finally use them.

### Proposed changes

- Add version 'v3' to the pathname of all new urls.

### Side effects

- Does any external service use the urls already? 

### Resolved issues

Fixes: #2290


sarahsporck:
> Should we probably update the API docs with the additional "/v3"?

Jep. You're totally right! I've updated the docs.

timobrembeck (approved):
Thanks a lot! :+1:

Once we have more than one active API version, we might need to reflect this in our package structure as well ([similar to how we did it for lunes-cms](https://github.com/digitalfabrik/lunes-cms/tree/develop/lunes_cms/api)), but since we aren't planning v4 yet, I guess it's not necessary to do the extra work now.

Issue #1589, opened 2022-07-09
Reset filters broken in user filter form


david-venhoff:
### Steps to Reproduce

1. Go to user filter form
2. Change some filters
3. Click on 'reset filters'

### Expected Behavior

The filters should get reset

### Actual Behavior

Nothing happens

MizukiTemma:
Filter resetting is not working in the page list too (Drop down for "Publication status" won't be reset).

PR #2618, opened 2024-01-23
Fix filter reset function

MizukiTemma:
### Short description

This PR fixes the reset function for filter. 

### Proposed changes

- Set `value` of `select` Objects to "" (default).
- I decided against `HTMLFormElement.reset()` because this sets the choices back to the latest chosen values not the default.

### Side effects

- None?

### Resolved issues

Fixes: #1589 

### DIFF
--- a/integreat_cms/release_notes/current/unreleased/1589.yml
+++ b/integreat_cms/release_notes/current/unreleased/1589.yml
@@ -1,0 +1,2 @@
+en: Fix filter reset function
+de: Repariere die "Filter zurücksetzen"-Funktion
--- a/integreat_cms/static/src/js/filter-form.ts
+++ b/integreat_cms/static/src/js/filter-form.ts
@@ -14,6 +14,9 @@ const resetToDefaultValue = (node: HTMLSelectElement | HTMLInputElement) => {
         // Non-checkbox input marked to have a default value
         /* eslint-disable-next-line no-param-reassign */
         node.value = node.getAttribute("data-default-value");
+    } else if (node.matches("select")) {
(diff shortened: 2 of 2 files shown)


integreat_cms/static/src/js/filter-form.ts
```diff
@@ -14,6 +14,9 @@ const resetToDefaultValue = (node: HTMLSelectElement | HTMLInputElement) => {
         // Non-checkbox input marked to have a default value
         /* eslint-disable-next-line no-param-reassign */
         node.value = node.getAttribute("data-default-value");
+    } else if (node.matches("select")) {
+        const idName = node.id;
```
JoeyStk, line 18:
```suggestion
        const inputElement = document.getElementById(node.id) as HTMLInputElement
        inputElement.value = ""
```
I'd prefer to write it like this, as I feel this is better use of the variable and more legible. But I suppose this comes down to personal preference, so feel free to disagree and not follow the suggestion

JoeyStk (commented):
Thank you for the PR. It looks good to me so far. I only have one small comment :)

integreat_cms/release_notes/current/unreleased/1589.yml
```diff
@@ -0,0 +1,2 @@
+en: Fix filter reset function
+de: Repariere Filter-Reset-Funktion
```
JoeyStk, line 2:
```suggestion
de: Repariere die "Filter zurücksetzen"-Funktion
```
Found something else :see_no_evil:

timobrembeck (approved):
Thanks! :tada:

Issue #2627, opened 2024-01-30
Cursor to move page among siblings disappears


osmers:
### Describe the Bug
Cursor to move page among siblings disappears once parent page was changed from initial selection to something else

### Steps to Reproduce

1. Go to any page
2. Change the parent page
3. See how the cursor next to your page in the siblings' menu disappears

### Expected Behavior
The cursor stays so that I can change the order of the sibling pages

### Actual Behavior
The cursor disappears and changing the order becomes impossible (only possible after reloading/saving the page)

### Additional Information
![image](https://github.com/digitalfabrik/integreat-cms/assets/64901227/90c76d1d-8f2c-43d3-8e20-463c22f99ef5)

PR #2635, opened 2024-02-05
Fix page order not being editable after changing the page parent

charludo:
### Short description

Fix page order not being editable after changing the page's parent in the page form sidebar

### Proposed changes

- this was caused by the `can_edit_page` variable being set in the parent template, which isn't utilized in the ajax call. Now the permission check is handled in the ajax call and the passed along from there.

### Side effects

- none I am aware of

### Resolved issues

Fixes: #2627


MizukiTemma (approved):
Nice 😄

timobrembeck (approved):
Thanks! :+1:

Issue #2624, opened 2024-01-28
Page link is not updated when changing title


MariaKabanova:
### Describe the Bug

Page link doesn't update when title changes

### Steps to Reproduce

1. Open any page
2. Change a title 
3. Click ‘update’ (do not perform any other actions before this step)
4. See page link

### Expected Behavior

The page link is updated according new page title

### Actual Behavior

The page link is not updated according new page title

### Additional Information

If you click anywhere on the form after changing the title but before clicking "Update", the url will be updated successfully.
The problem is only reproduced if "Update" is clicked immediately after changing the title.

This issue exists not only on pages, but on events and locations too

PR #2636, opened 2024-02-05
Update page link before submitting page form

charludo:
### Short description

I had a bit of trouble reproducing the issue. The only way I could trigger it was by entering Firefox's Responsive Design Mode and enabling Touch Simulation (though I suspect this might also happen on real touch devices).

I don't think this is worth an entry in the change log, this really is an edge case.

### Proposed changes

- if the content form is submitted while we're still waiting on a response for `slugify`, prevent the submission, then trigger it again once the previous call completed.

### Side effects

- hopefully not

### Resolved issues

Fixes: #2624

### DIFF
--- a/integreat_cms/cms/templates/events/event_form.html
+++ b/integreat_cms/cms/templates/events/event_form.html
@@ -41,10 +41,14 @@
                     <div class="flex flex-wrap gap-4 ml-auto mr-0 items-center">
                         {% include "generic_auto_save_note.html" with form_instance=event_form.instance %}
                         {% if perms.cms.publish_event %}
-                            <button name="status" value="{{ DRAFT }}" class="btn btn-outline">
+                            <button name="status"
+                                    value="{{ DRAFT }}"
+                                    class="btn btn-outline no-premature-submission">
                                 {% translate "Save as draft" %}
                             </button>
-                            <button name="status" value="{{ PUBLIC }}" class="btn">
+                            <button name="status"
+                                    value="{{ PUBLIC }}"
+                                    class="btn no-premature-submission">
                                 {% if event_translation_form.instance.status == PUBLIC %}
                                     {% translate "Update" %}
                                 {% else %}
@@ -52,7 +56,9 @@
                                 {% endif %}
                             </button>
                         {% else %}
-                            <button name="status" value="{{ REVIEW }}" class="btn">
+                            <button name="status"
+                                    value="{{ REVIEW }}"
+                                    class="btn no-premature-submission">
                                 {% translate "Submit for approval" %}
                             </button>
                         {% endif %}
@@ -67,7 +73,8 @@
                     <div class="w-full p-4 flex flex-col flex-auto">
                         <div class="flex justify-between">
                             <label for="{{ event_translation_form.title.id_for_label }}"
-                                   data-slugify-url="{% url 'slugify_ajax' region_slug=request.region.slug language_slug=language.slug model_type='event' %}{% if event_form.instance.id %}?model_id={{ event_form.instance.id }}{% endif %}">
+                                   data-slugify-url="{% url 'slugify_ajax' region_slug=request.region.slug language_slug=language.slug model_type='event' %}"
+                                   {% if event_form.instance.id %}data-model-id="{{ event_form.instance.id }}"{% endif %}>
                                 {{ event_translation_form.title.label }}
                             </label>
                             {% if event_translation_form.instance.id %}
--- a/integreat_cms/cms/templates/imprint/imprint_form.html
+++ b/integreat_cms/cms/templates/imprint/imprint_form.html
@@ -43,8 +43,7 @@
                 <div class="w-full rounded border border-blue-500 bg-white flex flex-col flex-auto">
                     <div class="w-full p-4 flex flex-col flex-auto">
                         <div class="flex justify-between mr-2">
-                            <label for="{{ page_translation_form.title.id_for_label }}"
-                                   data-slugify-url="{% url 'slugify_ajax' region_slug=request.region.slug language_slug=language.slug model_type='page' %}{% if page_form.instance.id %}?model_id={{ page_form.instance.id }}{% endif %}">
+                            <label for="{{ page_translation_form.title.id_for_label }}">
                                 {{ imprint_translation_form.title.label }}
                             </label>
                             {% if imprint_translation_form.instance.id %}
--- a/integreat_cms/cms/templates/pages/page_form.html
+++ b/integreat_cms/cms/templates/pages/page_form.html
@@ -52,10 +52,14 @@
                     </button>
                     {% has_perm 'cms.publish_page_object' request.user page as can_publish_page %}
                     {% if can_publish_page %}
-                        <button name="status" value="{{ DRAFT }}" class="btn btn-outline">
+                        <button name="status"
+                                value="{{ DRAFT }}"
+                                class="btn btn-outline no-premature-submission">
                             {% translate "Save as draft" %}
                         </button>
-                        <button name="status" value="{{ PUBLIC }}" class="btn whitespace-nowrap">
+                        <button name="status"
+                                value="{{ PUBLIC }}"
+                                class="btn whitespace-nowrap no-premature-submission">
                             {% if page_translation_form.instance.status == PUBLIC %}
                                 {% translate "Update" %}
                             {% else %}
@@ -63,7 +67,9 @@
                             {% endif %}
                         </button>
                     {% elif can_edit_page %}
-                        <button name="status" value="{{ REVIEW }}" class="btn">
+                        <button name="status"
+                                value="{{ REVIEW }}"
+                                class="btn no-premature-submission">
                             {% translate "Submit for approval" %}
                         </button>
                     {% endif %}
@@ -92,7 +98,8 @@
                         {% endif %}
                         <div class="flex justify-between mr-2">
                             <label for="{{ page_translation_form.title.id_for_label }}"
-                                   data-slugify-url="{% url 'slugify_ajax' region_slug=request.region.slug language_slug=language.slug model_type='page' %}{% if page_form.instance.id %}?model_id={{ page_form.instance.id }}{% endif %}">
+                                   data-slugify-url="{% url 'slugify_ajax' region_slug=request.region.slug language_slug=language.slug model_type='page' %}"
+                                   {% if page_form.instance.id %}data-model-id="{{ page_form.instance.id }}"{% endif %}>
                                 {{ page_translation_form.title.label }}
                             </label>
                             {% if page_translation_form.instance.id %}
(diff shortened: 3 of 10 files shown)


seluianova (changes requested):
> though I suspect this might also happen on real touch devices

Right, I was able to reproduce the issue using the touch screen.

As I can see now, the proposed changes fix that! 🚀 
I think this can be applied to events and locations as well.

**UPD**.
And the same problem occurs when saving to draft.

charludo:
> I think this can be applied to events and locations as well.
> 
> **UPD**. And the same problem occurs when saving to draft or submitting for approval.

Right, thanks for catching that!

I've added a more general solution.

integreat_cms/static/src/js/forms/prevent-premature-submission.ts
```diff
@@ -0,0 +1,26 @@
+export default class SubmissionPrevention {
```
seluianova, line 1: Maybe add some short clarification here? Because from the code itself it is rather difficult to understand why it was added

seluianova (changes requested):
Opps, sorry, I actually found one issue:
"Submit for approval" action has been broken for Observer
image
It can be reproduced, if Observer changes the page title.

charludo:
> Opps, sorry, I actually found one issue: "Submit for approval" action has been broken for Observer 
[...]
It can be reproduced, if Observer changes the page title.

Sorry for taking so long to get back to you.

Looks like this also happens on the develop branch, so it's unrelated to the changes. However, I think it might be intentional - @osmers, should Observers be able to change titles of pages they can edit?

If yes, I can open a new issue;
if no, I'd use this PR to set the field disabled for Observers, if that's all right with both of you.

seluianova:
> Looks like this also happens on the develop branch, so it's unrelated to the changes

I have re-checked:
The error on the screen is reproduced on develop, so it's really unrelated.
But the action itself works on develop - I can submit a title change for approval.

seluianova:
> Looks like this also happens on the develop branch, so it's unrelated to the changes

I have re-checked:
The error on the screen is reproduced on develop, so it's really unrelated.
But the action itself works on develop - I can submit a title change for approval now.

But right, if the title field should be disabled, it will solve the problem anyway :)

charludo:
> > Looks like this also happens on the develop branch, so it's unrelated to the changes
> 
> I have re-checked: The error on the screen is reproduced on develop, so it's really unrelated. But the action itself works on develop - I can submit a title change for approval now.
> 
> But right, if the title field should be disabled, it will solve the problem anyway :)

AH, I mistakenly thought Observers could not change the title, but in fact it's just that they can't directly change the link (so the automatic link update is, of course, not an issue at all here.

I've removed the submission prevention for "submit for approval".

osmers:
> > > Looks like this also happens on the develop branch, so it's unrelated to the changes
> > 
> > 
> > I have re-checked: The error on the screen is reproduced on develop, so it's really unrelated. But the action itself works on develop - I can submit a title change for approval now.
> > But right, if the title field should be disabled, it will solve the problem anyway :)
> 
> AH, I mistakenly thought Observers could not change the title, but in fact it's just that they can't directly change the link (so the automatic link update is, of course, not an issue at all here.
> 
> I've removed the submission prevention for "submit for approval".

I think it makes sense that observers can change the title of the page they are allowed to edit. Either they were given the rights to publish right away which means they are trusted to only do what's expected or they can only submit for review, in which case the changes to the title could be rejected if not wanted, right?

seluianova:
Observer can now submit a title change for approval (the same behavior as in develop) ✅
But the original problem remains for the "Submit for approval" action:
- login as Author 
- edit some page title
- click 'Submit for approval' using the touch screen
--> the page is saved with a new title, but the link is not updated

charludo:
> > I mistakenly thought Observers could not change the title, but in fact it's just that they can't directly change the link
> 
> Actually it seems they can change a link directly (submit a link change for approval), I have just checked in develop.

Alright. I think everythin is working now the same as on develop, plus touch screens work. However, I have not fixed the underlying bug (which is that the `slugify_ajax` endpoint raises `PermissionDenied` for observers because they are not allowed to edit objects), but instead just moved the release of the submission lock to a `finally()` block. This way the `fetch` still throws an unhandled error, just like on develop.

I think we should either
- keep the `PermissionDenied` but handle it "properly" (aka hide it / prevent logging to console), or
- change the required permission in `slugify_ajax.py` to `view_page` etc.

I like the second option better, but wasn't sure if there's some unintended consequences.

JoeyStk (approved):
Thank you very much! Looks really good! I really would like to have a test for this, but since we don't have JS tests yet, I think we can give it a pass

charludo:
> Thank you very much! Looks really good! I really would like to have a test for this, but since we don't have JS tests yet, I think we can give it a pass

Yes, fair 😅 Any suggestions on how to handle the `PermissionDenied`?

JoeyStk:
Oh sorry, I forgot to mention that. I also tend towards the second option. I think the PermissionDenied indicates that somewhere we have a logical thinking error in our permissions. I think changing the permission should be the right thing :thinking:

seluianova:
> change the required permission in slugify_ajax.py to view_page etc.

Sounds better to me, because if Observer has the right to modify the link directly, then modifying the link should also work through ajax. I think that would be more consistent.

charludo:
Alright, I've lowered the required permissions. 

"Worst case" someone can generate a slug for a content objects they shouldn't - but since the function only returns the generated slug, not actually saves it, I think that's OK.

I have also let the client-side error handling in, just in case anything else occurs server-side we still release the submission lock.

seluianova:
> "Worst case" someone can generate a slug for a content objects they shouldn't

This could be eliminated if we check for "cms.change_page_object" permission instead of "is user.role == Observer".
This will also make the check more accurate, because the slug can actually be edited when the user has permissions to a particular page and being an Observer is not enough.
But it will require some re-work of slugify_ajax() function. Not sure it's worth it, given the criticality of the original issue.

integreat_cms/cms/templates/pages/page_sbs.html
```diff
@@ -36,7 +36,9 @@ <h1 class="heading">
                                 {% endif %}
                             </button>
                         {% elif can_edit_page %}
-                            <button name="status" value="{{ REVIEW }}" class="btn">
+                            <button name="status"
```
seluianova, line 39: Currently we don't have ajax slug generation in the sbs form (probably it's a bug though)

seluianova: I have created a separate issue for that

charludo:
> > "Worst case" someone can generate a slug for a content objects they shouldn't
> 
> This could be eliminated if we check for "cms.change_page_object" permission instead of "is user.role == Observer". This will also make the check more accurate, because the slug can actually be edited when the user has permissions to a particular page and being an Observer is not enough. But it will require some re-work of slugify_ajax() function. Not sure it's worth it, given the criticality of the original issue.
> 
> Smthg like:
> 
> ```
> request.user.has_perm("cms.change_page_object", object_instance.page)
> ```
> 
> But I don't insist here, let me know if you would prefer to keep the current implementation.

You're absolutely right, that makes a lot more sense. Tbh I completely forgot that checking permissions for a specific object is an option... 😅

integreat_cms/cms/views/utils/slugify_ajax.py
```diff
@@ -39,31 +40,36 @@ def slugify_ajax(
...
     manager = managers[model_type].objects
     object_instance = manager.filter(
         **{model_type: model_id, "language": language}
     ).first()
+    request.user.has_perm(required_permission, getattr(object_instance, model_type))
+    observer = [role[1] for role in roles.ROLES if role[0] == roles.OBSERVER][0]
+    if not (
+        request.user.has_perms((required_permission,))
+        or str(request.user.role) == observer
+    ):
```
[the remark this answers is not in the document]
charludo: Oh no, I think I forgot to complete the refactor 🙈 That's a bit embarassing.

integreat_cms/cms/templates/imprint/imprint_form.html
```diff
@@ -44,7 +44,8 @@ <h1 class="heading">
                     <div class="w-full p-4 flex flex-col flex-auto">
                         <div class="flex justify-between mr-2">
                             <label for="{{ page_translation_form.title.id_for_label }}"
-                                   data-slugify-url="{% url 'slugify_ajax' region_slug=request.region.slug language_slug=language.slug model_type='page' %}{% if page_form.instance.id %}?model_id={{ page_form.instance.id }}{% endif %}">
+                                   data-slugify-url="{% url 'slugify_ajax' region_slug=request.region.slug language_slug=language.slug model_type='page' %}"
```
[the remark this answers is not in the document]
charludo: I guess so!

integreat_cms/cms/templates/events/event_form.html
```diff
@@ -67,7 +73,8 @@ <h1 class="heading overflow-hidden text-ellipsis">
                     <div class="w-full p-4 flex flex-col flex-auto">
                         <div class="flex justify-between">
                             <label for="{{ event_translation_form.title.id_for_label }}"
-                                   data-slugify-url="{% url 'slugify_ajax' region_slug=request.region.slug language_slug=language.slug model_type='event' %}{% if event_form.instance.id %}?model_id={{ event_form.instance.id }}{% endif %}">
+                                   data-slugify-url="{% url 'slugify_ajax' region_slug=request.region.slug language_slug=language.slug model_type='event' %}"
```
seluianova, line 76: 'if' was necessary for the case where a new event is created.
there is a failure without if

integreat_cms/cms/templates/pages/page_form.html
```diff
@@ -92,7 +98,8 @@ <h1 class="heading">
                         {% endif %}
                         <div class="flex justify-between mr-2">
                             <label for="{{ page_translation_form.title.id_for_label }}"
-                                   data-slugify-url="{% url 'slugify_ajax' region_slug=request.region.slug language_slug=language.slug model_type='page' %}{% if page_form.instance.id %}?model_id={{ page_form.instance.id }}{% endif %}">
+                                   data-slugify-url="{% url 'slugify_ajax' region_slug=request.region.slug language_slug=language.slug model_type='page' %}"
```
seluianova, line 101: please return the if condition :)

integreat_cms/cms/templates/pois/poi_form.html
```diff
@@ -59,7 +61,8 @@ <h1 class="overflow-hidden heading text-ellipsis">
                         <div class="flex flex-col flex-auto w-full p-4">
                             <div class="flex justify-between">
                                 <label for="{{ poi_translation_form.title.id_for_label }}"
-                                       data-slugify-url="{% url 'slugify_ajax' region_slug=request.region.slug language_slug=language.slug model_type='poi' %}{% if poi_form.instance.id %}?model_id={{ poi_form.instance.id }}{% endif %}">
+                                       data-slugify-url="{% url 'slugify_ajax' region_slug=request.region.slug language_slug=language.slug model_type='poi' %}"
```
seluianova, line 64: same here

seluianova (approved):
I've re-tested everything again and found no new problems except the one caused by removing the if-condition in the templates.
As soon as the if-s are returned, I think it's good to go 🤞
Thanks for tackling this tricky issue!

charludo:
> I've re-tested everything again and found no new problems except the one caused by removing the if-condition in the templates. As soon as the if-s are returned, I think it's good to go 🤞 Thanks for tackling this tricky issue!

No, thank YOU for the thorough reviews and attention to detail!!

Seriously, without you, this PR *might* have solved the original issue, but would also have introduced multiple new bugs. Thank you 💛

Issue #2586, opened 2023-12-11
Improve Language Selection of Maschine Translation  


MizukiTemma:
### Motivation

In future more automatic translation providers will be connected to our CMS system. It's time to improve and/or re-design "Language Selection" box in the machine translation settings, to make it clear for users which language is to be translated by which provider etc.

### Proposed Solution

TBD with UI/UX team

Leave comments here if you have any opinion, idea, question or concern 😉  

### Alternatives

N/A

### User Story

N/A

### Additional Context

Introduction of new automatic translation providers #1436 

The affected zone:
![image](https://github.com/digitalfabrik/integreat-cms/assets/82345046/f4c753b4-dc96-46f9-9336-849beb5f58c1)

### Design Requirements

Following points are subject to discussion and attention:
- Whether the current listing style of languages pro provider is good enough, possibly how to improve it
- How to handle languages that are supported by more than one provider (Should each language have only one provider assigned or make it possible to translate by multiple providers to, for example, compare the quality and chose the best one?) (How to display the check box etc in the box)

MizukiTemma:
In my opinion only one provider should be assined for each language and users choose one (and change to another if they want) for each language in the setting. This lets users save translation credits, is easier to use (it will be too much work at some point to compare result translations of multiple providers or to choose one provider every time when machine translation is needed), and avoids much of possible errors and bad user experience.

JoeyStk:
My idea would be to have a list of all languages that are supported by at least one provider. Next to the language I would probably add a dropdown button where region managers can select the MT providers that should be used for each language. For the default values I think we can use some sort of provider hierarchy.
@hauf-toni What do you think about this and do you have some additional insight?

dkehne:
Yeah, I am with @JoeyStk. Dropdown to choose from the provdiers.

JoeyStk:
It probably makes sense to implement this either before or alternatively in parallel with [#1436](https://github.com/digitalfabrik/integreat-cms/issues/1436).

osmers:
Just to add onto what's been said - I think if users want to compare quality they can do that outside the CMS by simply copying and pasting the text to both DeepL and Google and let native speakers check the translations. That way they do not need to use credits at all.

hauf-toni:
from ui/ux side it is a go for @JoeyStk 's idea (dropdown menu). would you like a design for this?

osmers:
I think a design would be useful, if it is not a lot of work @hauf-toni

[...]

PR #2637, opened 2024-02-05
Add Google Translate as MT provider

MizukiTemma:
### Short description

This PR adds Google Translate as an additional machine translation provider.

Currently Google Translate is available only for the languages that are not supported by DeepL. Possibility to choose a provider per language will be introduced in #2586 .

I suggest to add tests separately as #2641 so this PR will not be extremely large after reviews 😅 

### Proposed changes

- Add Google Translate in the same way with DeepL
- Extend `settings` and `integreat-cms.ini` for Google Authentication
- The hix check and budget control are there too.

### Side effects

- So far as I understood, the credential must be stored somewhere and accessed by the path given in `integreat-cms.ini`. I'm afraid this might bring security concerns.
- Move `check_usage` and `translatable_attributes` to the class `MachineTranslationAPIClient` to avoid duplicates.

### Resolved issues

Fixes: #1436 

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #2642, opened 2024-02-06
Change MT provider selection to dropdowns

PeterNerlich:
### Short description

Offer setting the preferred machine translations provider for each language.

### Proposed changes

- Add the field `preferred_mt_provider` to the `LanguageTreeNode` model.
- All critical places should still access the property `mt_provider`. It has been altered to return the provider for the name in `preferred_mt_provider` if available for the language, or the next best entry from the `mt_providers` list, or `None` if no providers can translate to that language.
- In the machine translations dashboard, replace the checkboxes which language to enable machine translations for with a per-language dropdown for the preferred translations provider.
- Re-arrange the machine translations dashboard so the list of dropdowns has enough room downwards.

### Side effects

- We create the form fields by generating them on the fly per language when required by the template. Because the languages are different and changing between and potentially within regions, they are not part of the formal model declaration.
This means that data regarding them gets rejected from `cleaned_data`, and we circumvent that by initially saving the raw data the form gets initialised with as `uncleaned_data`. This is surely not the best way to handle this, so please feel free to suggest alternative methods – we decided not to waste time on researching and re-discovering long known practices or reinventing the wheel any more than this, and instead rely on you, at this moment. :stuck_out_tongue: 
- The translation providers now don't exist only as classes referenced in a list in `constants` anymore, but also with their names as choices in the `LanguageTreeNode` model (the preferred provider setting). Seeing how we have a rather loose dependency on it (only accessing via the `mt_provider` property to fall back to the next best available provider in case), we might also lift that restriction and just make it a field for arbitrary strings, so we don't need database migrations when we add, remove or rename providers.
- Loosely related to the previous point: We might want to introduce slugs or similar identifiers for translation providers to separate this setting from the user facing name and make this more robust.

### Resolved issues

Fixes: #2586

@JoeyStk 


MizukiTemma:
> * We create the form fields by generating them on the fly per language when required by the template. Because the languages are different and changing between and potentially within regions, they are not part of the formal model declaration.
>   This means that data regarding them gets rejected from `cleaned_data`, and we circumvent that by initially saving the raw data the form gets initialised with as `uncleaned_data`. This is surely not the best way to handle this, so please feel free to suggest alternative methods – we decided not to waste time on researching and re-discovering long known practices or reinventing the wheel any more than this, and instead rely on you, at this moment. 😛


Is it the problem that each region has a defferent set/number of languages? Then [Model Form Functions](https://docs.djangoproject.com/en/5.0/ref/forms/models/) may be an option (if you haven't yet tried). This is what is being used for POICategoryTranslation and we can handle exactly as many forms as the number of non-root-languages of the region by setting `max_num` and `min_num`. 

Though we can implement this not now but later as follow-up refactoring because the current implementation is working anyway.

MizukiTemma (approved):
I approve this PR already, for the actual implementation is working as expected 😸 

Let's see how it will be with #2637 (it looks already good as far as I tried locally merged) 😁

david-venhoff (commented):
Thanks, this looks mostly good to me. I also tested to add a new translation provider and everything seems to work well.

I just feel a bit uneasy about the form changes. I think you could use a django [MultiValueField](https://docs.djangoproject.com/en/5.0/ref/forms/fields/#django.forms.MultiValueField) to represent multiple languages with each different choices.

david-venhoff:
I tried to use the multiValueField, but I am also not sure if it is completely right for this case :see_no_evil: 

This is what I hacked together (and quite incomplete)
<details>

```python
class LanguagePreferencesWidget(MultiWidget):
    def __init__(self, languages_dict, attrs=None):
        widgets = {}
        for slug, language_options in languages_dict.items():
            widgets[slug] = Select(
                attrs=attrs,
                choices=language_options.choices,
            )
        super().__init__(widgets, attrs)

    def decompress(self, value):
        print(value)
        if value:
            return list(value.values())
        return [None for _ in self.widgets]


class LanguagePreferences(MultiValueField):
    def __init__(self, languages_dict, **kwargs):
        fields = []
        for slug, language_options in languages_dict.items():
            fields.append(forms.ChoiceField(
                label=language_options.language_tree_node.translated_name,
                choices=language_options.choices,
                initial=language_options.initial,
            ))
        self.slugs = list(languages_dict.keys())
        initial_data = {slug: language_options.initial for slug, language_options in languages_dict.items()}

        super().__init__(fields, widget=LanguagePreferencesWidget(languages_dict), error_messages={}, initial=initial_data, **kwargs)
```

</details>

Alternatively, maybe you could maybe remove the `get_language_fields` method and set the fields in `__init__`? Then there is no need for `uncleaned_data` at least

PeterNerlich:
> Alternatively, maybe you could maybe remove the `get_language_fields` method and set the fields in `__init__`? Then there is no need for `uncleaned_data` at least

I'm not sure, originally fields are defined as static for the class and their names are listed in the `Meta` class, which I haven't tried accessing from within the `__init__` method. I just assumed this wouldn't be possible. Might be worth a try

david-venhoff (approved):
Given that this has `prio: high` now, I will just approve it, we can always change the form later if we find problems with this approach :see_no_evil:

MizukiTemma:
@PeterNerlich @JoeyStk 
Sorry, I take the joy of pressing the merge button from you 😿 to test Google Translate with this PR together.

Issue #2280, opened 2023-05-16
Handle DeepL API errors when requesting a translation


seluianova:
### Describe the Bug
We don't currently handle the situation when DeepL works during initialization, but then returns an error when requesting a translation.
![image](https://github.com/digitalfabrik/integreat-cms/assets/115008338/595b8233-50c9-4c08-bae3-b91e2e296a5d)

### Steps to Reproduce
Issue can be reproduced using the test test_deepl_bulk_machine_translation_api_error added in https://github.com/digitalfabrik/integreat-cms/pull/2278

Or manually using a standalone mock-server to simulate an error response.

Or manually using this fast workaround:

1. Enable DeepL translations by setting the correct API key in the settings.
When the CMS starts, it should successfully load supported languages.
2. Put the wrong server_url directly into DeepLApiClient class, for example:
```
self.translator = deepl.Translator(
    auth_key=settings.DEEPL_AUTH_KEY, server_url="https://deepl.com"
)
```
So that we could get a successful response when requesting supported langs, but an error, when requesting a translation.
3. Execute translation of any page

### Expected Behavior
I would expect a user-friendly error message, similar to what we have for SUMM AI:

_Page "Willkommen" could not be automatically translated into English. Please try again later or contact an administrator_

### Actual Behavior
Exception is shown

seluianova:
Possible errors from DeepL API docs:
413, "The request size exceeds the limit"
429, "Too many requests. Please wait and resend your request"
456, "Quota exceeded. The character limit has been reached"
5**, "Internal error"

PR #2606, opened 2024-01-13
Bundle feedback messages for all bulk actions

david-venhoff:
### Short description

This pr bundles feedback messages for bulk actions to avoid showing a single message for each affected object.

### Proposed changes

- Keep track of affected objects in lists by message type

### Side effects

/

### Resolved issues

Fixes: #2540 

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #2644, opened 2024-02-09
Add user friendly message for failed Deepl API requests

MizukiTemma:
### Short description

This PR adds user friendly error messages for DeepL errors.

### Proposed changes

- add `try` and `except` with new messages

### Side effects

- I'll wait for #2606 to make it sure that new error messages won't spam users.

### Resolved issues

Fixes: #2280 

### DIFF
--- a/integreat_cms/deepl_api/deepl_api_client.py
+++ b/integreat_cms/deepl_api/deepl_api_client.py
@@ -5,6 +5,7 @@ from html import unescape
 from typing import TYPE_CHECKING
 
 import deepl
+from deepl.exceptions import DeepLException
 from django.apps import apps
 from django.conf import settings
 from django.contrib import messages
@@ -147,13 +148,23 @@ class DeepLApiClient(MachineTranslationApiClient):
                     if hasattr(source_translation, attr) and getattr(
                         source_translation, attr
                     ):
-                        # data has to be unescaped for DeepL to recognize umlauts
-                        data[attr] = self.translator.translate_text(
-                            unescape(getattr(source_translation, attr)),
-                            source_lang=source_language.slug,
-                            target_lang=target_language_key,
-                            tag_handling="html",
-                        )
+                        try:
+                            # data has to be unescaped for DeepL to recognize Umlaute
+                            data[attr] = self.translator.translate_text(
+                                unescape(getattr(source_translation, attr)),
+                                source_lang=source_language.slug,
+                                target_lang=target_language_key,
+                                tag_handling="html",
+                            )
+                        except DeepLException as e:
+                            messages.error(
+                                self.request,
+                                _(
+                                    "A problem with DeepL API has occurred. Please contact an administrator."
+                                ),
+                            )
+                            logger.error(e)
+                            return
 
                 content_translation_form = self.form_class(
                     data=data,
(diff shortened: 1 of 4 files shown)


MizukiTemma:
Currently there is a detaield message for each case listed [here](https://github.com/digitalfabrik/integreat-cms/issues/2280#issuecomment-1551639085)


Though I and @david-venhoff came to the idea hat the error message probably doesn't have to be so detailed as it is right now and one universal error message is enough, for the focus of the problem is that the users get interrupted by the untreated exceptions.

What do you think @osmers ?

timobrembeck:
@MizukiTemma @osmers yes I agree, we don't necessarily need the detailed errors, but it helps when users report the problem to us, so we can easier identify and reproduce the problem ourselves.

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -9609,6 +9609,12 @@ msgstr[1] ""
 "Die folgenden {model_name_plural} konnten nicht automatisch übersetzt "
 "werden: {object_names}"
 
+#: deepl_api/deepl_api_client.py
+msgid ""
+"An problem with DeepL API has occurred. Please contact an administrator."
```
david-venhoff, line 9614:
```suggestion
"A problem with DeepL API has occurred. Please contact an administrator."
```

tests/deepl_api/deepl_api_test.py
```diff
@@ -1048,14 +1048,20 @@ def test_deepl_bulk_mt_no_target_language(
...
 @pytest.mark.parametrize(
     "login_role_user", PRIV_STAFF_ROLES + [AUTHOR, MANAGEMENT, EDITOR], indirect=True
 )
-@pytest.mark.skip(reason="server error handling is currently not implemented")
+@pytest.mark.parametrize("api_errors", api_errors)
+# pylint: disable=redefined-outer-name
 def test_deepl_bulk_mt_api_error(
     load_test_data: None,
     login_role_user: tuple[Client, str],
+    api_errors: int,
```
david-venhoff, line 1064:
```suggestion
    error: int,
```
A bit nitpicky, but maybe you could rename this to `error`, since right now it sounds like it could store multiple errors

david-venhoff (approved):
Thanks, LGTM

integreat_cms/deepl_api/deepl_api_client.py
```diff
@@ -147,13 +150,18 @@ def translate_queryset(
...
+                            data[attr] = self.translator.translate_text(
+                                unescape(getattr(source_translation, attr)),
+                                source_lang=source_language.slug,
+                                target_lang=target_language_key,
+                                tag_handling="html",
+                            )
+                        except DeepLException as e:
+                            deepl_api_error = True
+                            logger.error(e)
+                            break
```
seluianova, line 164:
```suggestion
                        except DeepLException as e:
                            messages.error(
                                self.request,
                                _(
                                    "A problem with DeepL API has occurred. Please contact an administrator."
                                ),
                            )
                            logger.error(e)
                            return
```

Assuming that we don't need 2 error messages here?
image
The top message occurs because the condition `if content_translation_form.is_valid()` is not met. It also adds the following message to the log:
> Mar 26 10:01:35 ERROR integreat_cms.deepl_api.deepl_api_client - Automatic translation for <Page (id: 4, parent: 1, region: augsburg, slug: interesting-facts-about-augsburg)> could not be created because of {'title': ['This field is required.'], 'slug': ['Cannot generate slug from title.']}
Seems irrelevant to me when there's a deepl error.

MizukiTemma:
@seluianova 
Thank you 😸 I've adopted your idea

seluianova (approved):
LGTM, thanks!

Issue #2578, opened 2023-11-30
`fix_internal_links` command is very inefficient


timobrembeck:
### Describe the Bug

Currently, the `fix_internal_links` links is doing the following under the hood:

1. Get all internal URLs
2. For each URL:
    1. Get the translation object the URL points to
    2. Get all link objects of the URL
    3. For each link:
        1. Get the target translation
        2. If the permalink differs: call `replace_links()`:
            1. For each content model:
                1. For each translation of this model:
                    1. For each link in this translation:
                        1. If the link matches the given old link, it is replaced by the fixed link

### Expected Behavior

Instead of traversing all URLs, then all links, then all translations and then all links again, we should consolidate the nested levels and only query the objects from the database we really need to

### Actual Behavior

The current implementation is utterly inefficient, especially when limited to only one region.

PR #2649, opened 2024-02-10
Improve performance of the `fix_internal_links` management command

david-venhoff:
### Short description

This pr tries to improve the performance of the `fix_internal_links` management command.
I could not test it on production data, but I am hopeful that this can speed up the command a lot.

### Proposed changes

- Don't use the `replace_links` function and instead work directly on the already known translation (Also skips having to wait for linkcheck listeners + 0.1 seconds after each update)
- If a region is given, only include links of that region
- Make sure that not more than one new version per translation is created in the process

Also some minor improvements from reducing the number of database calls

### Side effects

Ideally none

### Resolved issues

Fixes: #2578

### DIFF
--- a/integreat_cms/cms/utils/internal_link_utils.py
+++ b/integreat_cms/cms/utils/internal_link_utils.py
@@ -10,13 +10,12 @@ from urllib.parse import unquote, urlparse
 
 from django.conf import settings
 
+from ..constants import status
 from ..models import (
-    Event,
-    ImprintPage,
+    EventTranslation,
     ImprintPageTranslation,
-    Page,
     PageTranslation,
-    POI,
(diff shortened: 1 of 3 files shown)


david-venhoff:
Seems like this depends on #2648 to be merged

ulliholtgrave:
After rebasing your commits, this seems good to go for me. I did some local testing with our page duplication and I noticed some performance improvements. As we don't really have additional ASAP fixes in our pipeline, I would vouch for merging this soon and test it on the test system.

MizukiTemma (approved):
Nice 🐱

ulliholtgrave (approved):
Thanks :+1:

Issue #2633, opened 2024-02-05
Add speech bubble icon to Icon List


osmers:
### Motivation
Some municipalities use the speech bubble icon to indicate which languages are spoken at a certain location within their written text

### Proposed Solution
Add speech bubble icon to drop down list of icons bcs it makes it easier for municipalities to use the icon

### Alternatives
Let them add it via the media library as they currently do

### User Story
As an admin I want to be able to choose the speech bubble icon from a drop down list of icons in order to minimize the time I need to create content

### Additional Context
Check https://integreat.app/muenchen/en/services-for-women-and-girls/equality-between-women-and-men-2/equality-office-for-women to see the use of the icon 

### Design Requirements
None, bcs icon already exists.

PR #2651, opened 2024-02-12
Add spoken languages icon to tinymce

JoeyStk:
### Short description

This PR adds the speech bubble icon to indicate spoken languages to the tinymce icon list and shortcuts.

### Proposed changes

- Add speech bubble to icon list
- Add shortcut for it

### Side effects

- I noticed one strange side effect when testing this. With the new icon we get a lot of unused white space underneath the icon option. However as I don't think it has anything to with my immediate changes and I also couldn't find any quick CSS fixes for it, I decided to leave it open for now. Does any of you have an idea how to fix this and could help me out? Thanks in advance!
grafik

### Resolved issues

Fixes: #2633 


JoeyStk:
@seluianova Oh, thank you very much! I thought the issue was something completely different

MizukiTemma (approved):
Looks good 😸

seluianova (approved):
LGTM!

**UPD**.
Or you could even set it to 40x40.
_I suggested 80x80 because the previous icon (contact.svg) was 80x80, but it still had an unnecessary space at the bottom:_
image
40x40 IMO would look better.

PeterNerlich:
> **UPD**. Or you could even set it to 40x40. _I suggested 80x80 because the previous icon (contact.svg) was 80x80, but it still had an unnecessary space at the bottom

Oh I finally got what you mean, probably! It's that the icon SVGs themselves have width/height html attributes, the last ones set to `80` each, despite the parent element being limited to 24px width and height with CSS. Somehow the width is squished to those 24px and the icon looks correct, but the height remains the same, invisibly sticks out over the borders of the popup items and pushes the boundaries of the popup itself.

image
I think we can remove the width and height html attributes from the SVGs, we might only need to check the viewBox – I noticed the light bulb icon and the one for opening hours are missing it and become weird when only removing width/height.

And while we're at it addressing cosmetics, we could also ensure the SVGs have a consistent look. I see that the light bulb has a fill set, the others have not, and it looks slightly different.

Issue #2621, opened 2024-01-25
Reversing changes in the editor does not outdate HIX value


osmers:
### Describe the Bug
When I change a text in the editor, the HIX value shows that it is outdated. I can then refresh the value. When I use the "reverse changes" button, the text is changed but I cannot update the HIX value.

### Steps to Reproduce

1. Go to any page with content, e.g. https://admin.integreat-app.de/testumgebung/pages/de/33112/edit/
2. Eliminate part of the text
3. Refresh the HIX value
4. Press the Undo button
5. See that the HIX value cannot be refreshed

### Expected Behavior
The HIX value shows that it is outdated and the button to refresh appears

### Actual Behavior
HIX value shows that it is up to date

Issue #2577, opened 2023-11-29
HIX value can be manipulated by adding empty lines in a page


nassabay:
### Describe the Bug
If the Integreat user just enters a lot of empty lines at the end of the page, they improve the HIX value without having to make any special adjustments to the text.
We have had cases in Soest and Erlensee, for example.

We have tried to reproduce this behaviour directly on the page of Textlab and there the HIX value stays the same.

### Steps to Reproduce

1. Create a page in a test environment with a short text of long words.
2. Note the bad HIX value
3. Add a few empty lines at the end by just pressing enter
4. Note the increased HIX value

### Expected Behavior
The HIX value should be independent of multiple empty lines in the text.

### Actual Behavior
The HIX value increases with every empty line.

PR #2652, opened 2024-02-12
Fix HIX value diff when page content has nested tags

seluianova:
### Short description
Using root.iter() was not a good choice to iterate over the paragraphs, because it goes "one-way" deep into the nested elements.
For example, if a page has the following structure:
`<div><p>Some image</p><p><a href="some.image"></a></p><p>&nbsp;</p></div>`
The iteration will proceed as follows:
```
<div>
<p>
<p>
<a>
<img>
--> end of iteration
```
And the last p-node will not be processed.

### Proposed changes
Use list(root) instead of root.iter()

### Side effects
No?

### Resolved issues
Another follow-up to #2577

**UPD**: added another small HIX-related fix
Fixes: #2621

### DIFF
--- a/integreat_cms/release_notes/current/unreleased/2621.yml
+++ b/integreat_cms/release_notes/current/unreleased/2621.yml
@@ -1,0 +1,2 @@
+en: Mark HIX value as outdated when changes are undone in the editor
+de: Markiere HIX-Wert als veraltet, wenn Änderungen im Editor rückgängig gemacht werden
--- a/integreat_cms/cms/views/utils/hix.py
+++ b/integreat_cms/cms/views/utils/hix.py
@@ -42,7 +42,7 @@ def normalize_text(text: str) -> str:
         root = fromstring(text)
 
         # Remove paragraphs without text (e.g. empty paragraphs or with an image only)
-        for node in root.iter():
+        for node in list(root):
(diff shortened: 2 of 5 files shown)


seluianova:
Added another small HIX-related fix for #2621

integreat_cms/release_notes/current/unreleased/2621.yml
```diff
@@ -0,0 +1,2 @@
+en: Outdate HIX value when reversing changes in the editor
+de: Veralte HIX-Wert beim Rückgängigmachen von Änderungen im Editor
```
seluianova, line 2: Not sure if it sounds okay in German?

timobrembeck: Maybe like this?
```suggestion
de: Markiere HIX-Wert als veraltet, wenn Änderungen im Editor rückgängig gemacht werden
```

Also the English text could be made a bit more clear, but I would also be fine with the current wording.
```
en: Mark HIX value as outdated when changes are undone in the editor
```

timobrembeck (approved):
Thanks! :+1:

JoeyStk (approved):
Thank you very much. LGTM!

PR #2654, opened 2024-02-12
Dynamic export formats

PeterNerlich:
### Short description

Allow export to other formats supported by tablib / django-import-export
We were troubleshooting what looked like an encoding problem with feedback exported as CSV, and by the time we found that EXCEL was at fault and our CSVs were flawless, this was already half done. I figured I might as well clean it up and submit it as a low prio PR.

### Proposed changes

- Extend feedback export URL endpoint by `format` parameter
- Check if requested format is available through the dependency library
- Automatically determine mime type

### Side effects

- We just throw a ValueError if the format is not available. This is likely not the most elegant solution, but at the moment we hard-code which format can be requested through the UI anyway.

### Resolved issues

none

### DIFF
--- a/integreat_cms/cms/templates/feedback/region_feedback_list.html
+++ b/integreat_cms/cms/templates/feedback/region_feedback_list.html
@@ -119,7 +119,7 @@
                             {% translate "Delete" %}
                         </option>
                     {% endif %}
-                    <option data-bulk-action="{% url 'export_region_feedback' region_slug=request.region.slug %}">
+                    <option data-bulk-action="{% url 'export_region_feedback' region_slug=request.region.slug file_format='csv' %}">
                         {% translate "Export to CSV" %}
                     </option>
                 </select>
--- a/integreat_cms/cms/views/feedback/region_feedback_actions.py
+++ b/integreat_cms/cms/views/feedback/region_feedback_actions.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 import logging
 from typing import TYPE_CHECKING
 
+import magic
 from cacheops import invalidate_model, invalidate_obj
 from django.contrib import messages
 from django.http import HttpResponse
@@ -14,6 +15,7 @@ from django.shortcuts import redirect
 from django.utils.translation import gettext_lazy as _
 from django.views.decorators.http import require_POST
 from tablib import Dataset
+from tablib.formats import registry as format_registry
 
 from ...decorators import permission_required
 from ...models import Feedback
@@ -181,13 +183,15 @@ def delete_region_feedback(
 def export_region_feedback(
     request: HttpRequest,
     region_slug: str,
+    file_format: str,
 ) -> HttpResponse:
     """
-    Export a list of feedback items (to CSV)
+    Export a list of feedback items
 
     :param request: Object representing the user call
     :param region_slug: The slug of the current region
-    :return: Response with CSV file
+    :param file_format: The export format
+    :return: Response with file
     """
     selected_ids = request.POST.getlist("selected_ids[]")
     selected_feedback = Feedback.objects.filter(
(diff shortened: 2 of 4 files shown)


MizukiTemma:
Is this to fix the problem reported [here](https://chat.tuerantuer.org/digitalfabrik/pl/ebbqa7bjo3fcpxcxoqzf4rmcxc)?

PeterNerlich:
@MizukiTemma More or less. The fix is to trick EXCEL into actually reading it as UTF-8. But this was the related discussion that resulted in this PR, yes.

integreat_cms/cms/templates/feedback/region_feedback_list.html
```diff
@@ -184,9 +184,12 @@ <h1 class="heading">
                                 {% translate "Delete" %}
                             </option>
                         {% endif %}
-                        <option data-bulk-action="{% url 'export_region_feedback' region_slug=request.region.slug %}">
+                        <option data-bulk-action="{% url 'export_region_feedback' region_slug=request.region.slug file_format='csv' %}">
                             {% translate "Export to CSV" %}
                         </option>
+                        <!--<option data-bulk-action="{% url 'export_region_feedback' region_slug=request.region.slug file_format='xlsx' %}">
```
JoeyStk, line 190:
```suggestion

```
Since we agreed that xlsx shouldn't be supported, we can remove this and the comment below

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -5578,6 +5578,10 @@ msgstr "Noch kein Feedback vorhanden."
 msgid "Export to CSV"
 msgstr "Als CSV-Datei exportieren"
 
+#: cms/templates/feedback/region_feedback_list.html
+msgid "Export to XLSX"
```
JoeyStk, line 5582:
```suggestion
```
And we can also undo the translation for it

JoeyStk (commented):
Thank you. I have two small comments before reviewing this properly :)

integreat_cms/core/management/commands/duplicate_feedback.py
```diff
@@ -0,0 +1,70 @@
+from __future__ import annotations
```
JoeyStk, line 1: As discussed this file should not be part of this PR :)

JoeyStk (approved):
Except from my comment below and resolving the failing tests, I don't have any further objections

JoeyStk:
Ah, and one last information I forgot to mention. The length of your commit message is currently at 72 characters. I was wondering, if this might be to long. However the [Linux kernel documentation](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst?id=bc7938deaca7f474918c41a0372a410049bd4e13#n664) says: 

> For these reasons, the ``summary`` must be no more than 70-75
characters, and it must describe both what the patch changes, as well
as why the patch might be necessary.  It is challenging to be both
succinct and descriptive, but that is what a well-written summary
should do.

So apparently this is not an issue. I just wanted to mention to you that it was something I thought about, and let you know what made me decided not to make it part of my review :)

integreat_cms/cms/views/feedback/region_feedback_actions.py
```diff
@@ -204,8 +208,13 @@ def export_region_feedback(
         data = resource.export_resource(obj)
         dataset.append(data)
 
-    response = HttpResponse(dataset.csv, content_type="text/csv")
+    if file_format in map(lambda f: f.title, format_registry.formats()):
```
david-venhoff, line 211:
```suggestion
    if file_format in (f.title for f in format_registry.formats()):
```
Since we recently had a pr removing all usages of map, I think we shouldn't use it anymore

david-venhoff (approved):
LGTM

Issue #2665, opened 2024-02-20
Cloning Region with page based offers


svenseeberg:
### Describe the Bug
When cloning a region that contains page based offers and the new region does not have the offers activated, it is not possible to save the region form.

> Einige Angebote konnten nicht deaktiviert werden, da sie mindestens in einer Seite eingebettet sind.

Addtionally, the Zammad URL field is not changeable, the help form checkbox as well.

### Steps to Reproduce

1. Create a region with a Zammad form added to a page
2. Clone the region into a new region w/o the offer being activated
3. Try to change any region setting and save the form
4. The error message appears

### Expected Behavior
2 Options:
- Either the offer needs to be deactivated during the cloning of the region
- the offer has to be force activated for the region

### Actual Behavior
The page offer is activated but the offer is not active in the region settings.

### Additional Information
![image](https://github.com/digitalfabrik/integreat-cms/assets/15197000/c3111510-ec48-4a20-918f-1b3fa4d7cce8)

PR #2667, opened 2024-02-21
Fix behavior when cloning region with page based offers

charludo:
### Short description

Currently, cloning a region which has embedded offfers into pages can lead to unexpected errors when the newly-created region does not have the same offers enabled.

### Proposed changes

- allow users to choose from two behaviors when cloning a region: either force-activate all required offers\* or discard embeddings for missing offers when cloning pages.
- \* small caveat: if no Zammad URL is provided for the new region, Zammad embeddings are discarded even when force-activation is selected

### Side effects

- none that I could find. As far as I can tell, results are correct in all cases (no cloning, cloning with force-activation or discarding, with/without Zammad URL) 

### Resolved issues

Fixes: #2665 

### DIFF
--- a/integreat_cms/cms/forms/regions/region_form.py
+++ b/integreat_cms/cms/forms/regions/region_form.py
@@ -22,7 +22,7 @@ from integreat_cms.cms.utils.linkcheck_utils import replace_links
 from ....gvz_api.utils import GvzRegion
 from ....matomo_api.matomo_api_client import MatomoException
 from ....nominatim_api.nominatim_api_client import NominatimApiClient
-from ...constants import status
+from ...constants import duplicate_pbo_behaviors, status
 from ...models import LanguageTreeNode, OfferTemplate, Page, PageTranslation, Region
 from ...models.regions.region import format_mt_help_text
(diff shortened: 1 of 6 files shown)


charludo:
One test i failing becuase the valid link count is somehow higher now after duplicating a region - I don't really know why, does someone have an idea?

integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -112,6 +114,26 @@ class RegionForm(CustomModelForm):
         ),
     )
 
+    duplication_pbo_behavior = forms.ChoiceField(
+        choices=(
```
david-venhoff, line 118: The choices could go into the constants directory, to be more consistent with the rest of the code (and prevent typos)

david-venhoff (approved):
LGTM

MizukiTemma (approved):
Thank you 👍 works as expected!

integreat_cms/cms/forms/regions/region_form.py
charludo: Right, good point.

david-venhoff (approved):
:tada:

PR #2668, opened 2024-02-21
Fix HIX value difference between CMS and Textlab Online service

seluianova:
### Short description
It was discovered that the HIX value in https://textlab.online and in CMS is not identical:
image
image

### Proposed changes
As I found, TextLab Online uses `<br>` for line breaks, while we used `\r\n`.
And that seems to be the root cause of the observed difference.

Then when I changed the line breaks in our text normalization function, I found that this actually makes the entire function unnecessary. Because it fixed all of our previous problems with the HIX value difference that I was aware of  🙈
Including:
- HIX value diff when adding empty lines
- HIX value diff between pre save signal and tinymce
etc.

### Side effects
I tested everything I could think of, but I may have missed something.
So additional testing is certainly welcomed.

### Resolved issues

Fixes: #


MizukiTemma (approved):
>  Because it fixed all of our previous problems with the HIX value difference that I was aware of 🙈

🙈 

Thank you so much for all the analysis and work 🙏 Looks good!

david-venhoff (approved):

Issue #2582, opened 2023-12-05
Bring test coverage to 82,5 %


MizukiTemma:
### Motivation

As a part of our "Zero bug policy" let us increase the test coverage continueing from #2492 

### Proposed Solution

Goal for the end of March 2024: 82,5% 
Current coverage:  [![](https://img.shields.io/codeclimate/coverage/digitalfabrik/integreat-cms)](https://codeclimate.com/github/digitalfabrik/integreat-cms/code?sort=-loc)

### Additional Context

For more information see #2492 
The goal percentage can be adjusted after all the test related PRs are closed that are currently open.

MizukiTemma:
Please leave comments here which part will be worked on, when you start, to avoid multiple people working on the same parts.

JoeyStk:
Peter and I decided to write tests for [page_form_view.py](integreat_cms/cms/views/pages/page_form_view.py) :).

seluianova:
@MariaKabanova and I will take hix.py

MizukiTemma:
Sorry, I've closed it accidentally by the latest PR. Still 0,5% to go 😅

seluianova:
0,4% to go :)

JoeyStk:
And we're back to 81,7% :sweat_smile:

timobrembeck:
> And we're back to 81,7% 😅

Well, that's on me :innocent: 
And will be fixed with #2772 hopefully.

MizukiTemma:
I hope #2696 will bring it up again 👀

MizukiTemma:
We've reached 82,5% 🎉 

Let's keep this good coverage 😸

MizukiTemma:
We've reached 82,5% 🎉

Let's keep this good coverage 😸

PR #2669, opened 2024-02-21
Add tests for HIX signals

MariaKabanova:
### Short description

This PR for HIX score representation on page edit

### Proposed changes

Tests for:
- HIX score is requested and saved when the page translation is updated
- HIX score is not updated when hix is disabled on region level
- HIX score is not updated when hix is enabled on page level but hix_ignore flag is set to "on"
- HIX score is not updated when hix is enabled on page level and page content is empty
- HIX score is not updated when hix is enabled on page level and page content is not updated
- HIX score is requested but response 400 received. Hix score is not updated

### Related issues

Relates to: #2582

### DIFF
--- a/integreat_cms/core/circleci_settings.py
+++ b/integreat_cms/core/circleci_settings.py
@@ -27,6 +27,10 @@ SUMM_AI_ENABLED = True
 DEEPL_AUTH_KEY = "dummy"
 #: Enable manually because existing setting derives from the unset env var
 DEEPL_ENABLED = True
+#: Set dummy Textlab key to test automatic translations via Textlab API
+TEXTLAB_API_KEY = "dummy"
+#: Enable manually because existing setting derives from the unset env var
+TEXTLAB_API_ENABLED = True
 #: Disable linkcheck listeners on CircleCI
 LINKCHECK_DISABLE_LISTENERS = True
 # Disable background tasks during testing
--- a/tests/textlab_api/textlab_api_test.py
+++ b/tests/textlab_api/textlab_api_test.py
@@ -1,0 +1,302 @@
+from __future__ import annotations
+
+import json
+import time
+from typing import TYPE_CHECKING
+
+import pytest
+from django.http import HttpResponse
+from django.test.client import Client
+from django.urls import reverse
+from werkzeug.wrappers import Request, Response
+
+from integreat_cms.cms.constants import status
+from integreat_cms.cms.constants.roles import EDITOR
+from integreat_cms.cms.models.pages.page import Page
+from integreat_cms.cms.models.pages.page_translation import PageTranslation
+from integreat_cms.cms.models.regions.region import Region
+from tests.mock import MockServer
+
+if TYPE_CHECKING:
+    from django.test.client import Client
+    from pytest_django.fixtures import SettingsWrapper
+    from pytest_httpserver.httpserver import HTTPServer
+
+
+def update_page_content(
+    admin_client: Client,
+    title: str,
+    content: str,
+    hix_ignore: bool = False,
+) -> tuple[str, HttpResponse]:
+
+    edit_page = reverse(
+        "edit_page",
+        kwargs={
+            "region_slug": "augsburg",
+            "language_slug": "de",
+            "page_id": 2,
+        },
+    )
+    edit_page_data = {
+        "title": title,
+        "content": content,
+        "mirrored_page_region": "",
+        "_ref_node_id": 3,
+        "_position": "right",
+        "status": status.PUBLIC,
+    }
+    if hix_ignore:
+        edit_page_data["hix_ignore"] = "on"
+
+    return (edit_page, admin_client.post(edit_page, edit_page_data))
+
+
+@pytest.mark.django_db
+def test_hix_score_update(
+    load_test_data: None,
+    admin_client: Client,
+    settings: SettingsWrapper,
+    mock_server: MockServer,
+) -> None:
+    """
+    Check that the HIX score is requested and saved when the page translation is updated
+
+    :param load_test_data: The fixture providing the test data (see :meth:`~tests.conftest.load_test_data`)
+    :param admin_client: The fixture providing the http client
+    :param settings: The fixture providing the django settings
+    :param mock_server: The fixture providing the dummy http server
+    """
+    # Setup a mocked Textlab API server with dummy responses
+    mock_server.configure("/user/login", 200, {"token": "dummy"})
+    mock_server.configure("/benchmark/420", 200, {"formulaHix": 15.48920568})
+
+    # Redirect call aimed at the Textlab API to the fake server
+    settings.TEXTLAB_API_URL = f"http://localhost:{mock_server.port}"
+
+    # Enable Textlab in the test region
+    Region.objects.filter(slug="augsburg").update(hix_enabled=True)
+
+    edit_page, response = update_page_content(
+        admin_client, "Willkommen in Augsburg", "Neuer Inhalt"
+    )
+
+    assert response.status_code == 302
+    assert response.headers.get("Location") == edit_page
+
+    # Check the updated HIX score
+    page_translation = Page.objects.get(id=2).get_translation("de")
+
+    assert page_translation.hix_score == 15.48920568
+    assert mock_server.requests_counter == 2
+
+
+@pytest.mark.django_db
+def test_hix_disable_on_region(
+    load_test_data: None,
+    admin_client: Client,
+    settings: SettingsWrapper,
+    mock_server: MockServer,
+) -> None:
+    """
+    Check that the HIX score is not updated when hix is disabled on region level. No request to TextLab API is expected.
+
+    :param load_test_data: The fixture providing the test data (see :meth:`~tests.conftest.load_test_data`)
+    :param admin_client: The fixture providing the http client
+    :param settings: The fixture providing the django settings
+    :param mock_server: The fixture providing the dummy http server
+    """
+
+    # Setup a mocked Textlab API server with dummy responses
+    mock_server.configure("/user/login", 200, {"token": "dummy"})
+    mock_server.configure("/benchmark/420", 200, {"formulaHix": 20.0})
+
+    # Redirect call aimed at the Textlab API to the fake server
+    settings.TEXTLAB_API_URL = f"http://localhost:{mock_server.port}"
+
+    # Disable Textlab in the test region
+    Region.objects.filter(slug="augsburg").update(hix_enabled=False)
+
+    edit_page, response = update_page_content(
+        admin_client, "Willkommen in Augsburg", "Neuer Inhalt1"
+    )
+
+    assert response.status_code == 302
+    assert response.headers.get("Location") == edit_page
+
+    # Check that HIX score was not updated
+    page_translation = Page.objects.get(id=2).get_translation("de")
+
+    assert page_translation.hix_score is None
+    assert mock_server.requests_counter == 0
+
+
+@pytest.mark.django_db
+def test_ignore_hix_on_page(
+    load_test_data: None,
+    admin_client: Client,
+    settings: SettingsWrapper,
+    mock_server: MockServer,
+) -> None:
+    """
+    Check that the HIX score is not updated when hix is disabled at the page level (hix_ignore=True). No request to TextLab API is expected.
+
+    :param load_test_data: The fixture providing the test data (see :meth:`~tests.conftest.load_test_data`)
+    :param admin_client: The fixture providing the http client
+    :param settings: The fixture providing the django settings
+    :param mock_server: The fixture providing the dummy http server
+    """
+
+    # Setup a mocked Textlab API server with dummy responses
+    mock_server.configure("/user/login", 200, {"token": "dummy"})
+    mock_server.configure("/benchmark/420", 200, {"message": "Error occurred"})
+
+    # Redirect call aimed at the Textlab API to the fake server
+    settings.TEXTLAB_API_URL = f"http://localhost:{mock_server.port}"
+
+    # Enable Textlab in the test region
+    Region.objects.filter(slug="augsburg").update(hix_enabled=True)
+
+    edit_page, response = update_page_content(
+        admin_client, "Willkommen in Augsburg", "Neuer Inhalt2", hix_ignore=True
+    )
+
+    assert response.status_code == 302
+    assert response.headers.get("Location") == edit_page
+
+    # Check that the HIX score has not been updated
+    page_translation = Page.objects.get(id=2).get_translation("de")
+
+    assert page_translation.hix_score is None
+    assert mock_server.requests_counter == 0
+
+
+@pytest.mark.django_db
+def test_hix_page_content_empty(
+    load_test_data: None,
+    admin_client: Client,
+    settings: SettingsWrapper,
+    mock_server: MockServer,
+) -> None:
+    """
+    Check that the HIX score is not updated when hix is enabled on page level and page content is empty. No request to TextLab API is expected.
+
+    :param load_test_data: The fixture providing the test data (see :meth:`~tests.conftest.load_test_data`)
+    :param admin_client: The fixture providing the http client
+    :param settings: The fixture providing the django settings
+    :param mock_server: The fixture providing the dummy http server
+    """
+
+    # Setup a mocked Textlab API server with dummy responses
+    mock_server.configure("/user/login", 200, {"token": "dummy"})
+    mock_server.configure("/benchmark/420", 500, {"message": "Error occurred"})
+
+    # Redirect call aimed at the Textlab API to the fake server
+    settings.TEXTLAB_API_URL = f"http://localhost:{mock_server.port}"
+
+    # Enable Textlab in the test region
+    Region.objects.filter(slug="augsburg").update(hix_enabled=True)
+
+    edit_page, response = update_page_content(
+        admin_client, "Willkommen in Augsburg", ""
+    )
+
+    assert response.status_code == 302
+    assert response.headers.get("Location") == edit_page
+
+    # Check that the HIX score has not been updated
+    page_translation = Page.objects.get(id=2).get_translation("de")
+
+    assert page_translation.hix_score is None
+    assert mock_server.requests_counter == 0
+
+
+@pytest.mark.django_db
+def test_hix_no_content_changes(
+    load_test_data: None,
+    admin_client: Client,
+    settings: SettingsWrapper,
+    mock_server: MockServer,
+) -> None:
+    """
+    Check that the HIX score is copied from the previous translation version if the page content is not changed. No request to TextLab API is expected.
+
+    :param load_test_data: The fixture providing the test data (see :meth:`~tests.conftest.load_test_data`)
+    :param admin_client: The fixture providing the http client
+    :param settings: The fixture providing the django settings
+    :param mock_server: The fixture providing the dummy http server
+    """
+
+    # Setup a mocked Textlab API server with dummy responses
+    mock_server.configure("/user/login", 200, {"token": "dummy"})
+    mock_server.configure("/benchmark/420", 500, {"message": "Error occurred"})
+
+    # Redirect call aimed at the Textlab API to the fake server
+    settings.TEXTLAB_API_URL = f"http://localhost:{mock_server.port}"
+
+    # Enable Textlab in the test region
+    Region.objects.filter(slug="augsburg").update(hix_enabled=True)
+
+    # Setting HIX score initial value
+    initial_hix_score = 11
+    PageTranslation.objects.filter(page__id=2, language__slug="de").update(
+        hix_score=initial_hix_score
+    )
+
+    # Check initial page properties
+    page_translation = Page.objects.get(id=2).get_translation("de")
+    content_before_change = page_translation.content
+    assert content_before_change != ""
+    assert page_translation.latest_version.hix_score == initial_hix_score
+
+    edit_page, response = update_page_content(
+        admin_client, "Neuer Titel", content_before_change
+    )
+
+    assert response.status_code == 302
+    assert response.headers.get("Location") == edit_page
+
+    # Check that the HIX score has been copied from the previous version
+    page_translation = Page.objects.get(id=2).get_translation("de")
+
+    assert page_translation.hix_score == initial_hix_score
+    assert page_translation.content == content_before_change
+    assert mock_server.requests_counter == 0
+
+
+@pytest.mark.django_db
+def test_hix_response_400(
+    load_test_data: None,
+    admin_client: Client,
+    settings: SettingsWrapper,
+    mock_server: MockServer,
+) -> None:
+    """
+    Check that the HIX score is not being updated when the API returns a HTTP 400 error.
+
+    :param load_test_data: The fixture providing the test data (see :meth:`~tests.conftest.load_test_data`)
+    :param admin_client: The fixture providing the http client
+    :param settings: The fixture providing the django settings
+    :param mock_server: The fixture providing the dummy http server
+    """
+    # Setup a mocked Textlab API server with dummy responses
+    mock_server.configure("/user/login", 200, {"token": "dummy"})
+    mock_server.configure("/benchmark/420", 400, {"message": "Error occurred"})
+
+    # Redirect call aimed at the Textlab API to the fake server
+    settings.TEXTLAB_API_URL = f"http://localhost:{mock_server.port}"
+
+    # Enable Textlab in the test region
+    Region.objects.filter(slug="augsburg").update(hix_enabled=True)
+
+    edit_page, response = update_page_content(
+        admin_client, "Willkommen in Augsburg", "Neuer Inhalt3"
+    )
+
+    assert response.status_code == 302
+    assert response.headers.get("Location") == edit_page
+
+    # Check that the HIX score has not been updated
+    page_translation = Page.objects.get(id=2).get_translation("de")
+
+    assert page_translation.hix_score is None
--- a/integreat_cms/core/settings.py
+++ b/integreat_cms/core/settings.py
@@ -259,11 +259,11 @@ TEXTLAB_API_URL: Final[str] = os.environ.get(
 )
 
 #: Key for the Textlab API
-TEXTLAB_API_KEY: Final[str | None] = os.environ.get("INTEGREAT_CMS_TEXTLAB_API_KEY")
+TEXTLAB_API_KEY: str | None = os.environ.get("INTEGREAT_CMS_TEXTLAB_API_KEY")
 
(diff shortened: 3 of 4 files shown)


tests/textlab_api/textlab_api_test.py
```diff
@@ -13,3 +28,353 @@ def test_normalize_text(
...
+) -> None:
+    """
+    Check that the HIX score is not updated when hix is disabled on region level. No request to TexLab API is expected.
+
+    :param load_test_data: The fixture providing the test data (see :meth:`~tests.conftest.load_test_data`)
+    :param login_role_user: The fixture providing the http client and the current role (see :meth:`~tests.conftest.login_role_user`)
+    :param settings: The fixture providing the django settings
+    :param httpserver: The fixture providing the dummy http server
+    """
+
```
seluianova, line 110: I think it makes sense to configure the mock server in all cases, otherwise the HIX value will be None because TextLab is unavailable, and not because "hix is disabled on region level", "hix_ignore flag is set to "on"", etc

tests/textlab_api/textlab_api_test.py
```diff
@@ -13,3 +28,353 @@ def test_normalize_text(
...
+    # Redirect call aimed at the Textlab API to the fake server
+    settings.TEXTLAB_API_URL = f"http://localhost:{httpserver.port}"
+
+    # Enable Textlab in the test region
```
seluianova, line 114:
```suggestion
    # Disable Textlab in the test region
```

tests/textlab_api/textlab_api_test.py
```diff
@@ -13,3 +28,353 @@ def test_normalize_text(
...
+        "_position": "right",
+        "status": status.PUBLIC,
+    }
+
+    response = client.post(edit_page, edit_page_data)
+
+    assert response.status_code == 302
+    assert response.headers.get("Location") == edit_page
+
+    # Check the updated HIX score
```
seluianova, line 144:
```suggestion
    # Check that the HIX score was not updated
```

tests/textlab_api/textlab_api_test.py
```diff
@@ -13,3 +28,353 @@ def test_normalize_text(
...
+@pytest.mark.django_db
+@pytest.mark.parametrize("login_role_user", [EDITOR], indirect=True)
+def test_ignore_hix_on_page(
+    load_test_data: None,
+    login_role_user: tuple[Client, str],
+    settings: SettingsWrapper,
+    httpserver: HTTPServer,
+) -> None:
+    """
+    Check that the HIX score is not updated when hix is enabled on page level but hix_ignore flag is set to "on". No request to TexLab API is expected.
```
seluianova, line 159:
```suggestion
    Check that the HIX score is not updated when hix is disabled on page level (hix_ignore=True). No request to TexLab API is expected.
```

tests/textlab_api/textlab_api_test.py
```diff
@@ -13,3 +28,353 @@ def test_normalize_text(
...
+    :param load_test_data: The fixture providing the test data (see :meth:`~tests.conftest.load_test_data`)
+    :param login_role_user: The fixture providing the http client and the current role (see :meth:`~tests.conftest.login_role_user`)
+    :param settings: The fixture providing the django settings
+    :param httpserver: The fixture providing the dummy http server
+    """
+
+    # Redirect call aimed at the Textlab API to the fake server
+    settings.TEXTLAB_API_URL = f"http://localhost:{httpserver.port}"
+
+    # Enable Textlab in the test page
```
seluianova, line 170:
```suggestion
    # Enable Textlab in the test region
```

tests/textlab_api/textlab_api_test.py
```diff
@@ -13,3 +28,353 @@ def test_normalize_text(
...
```
seluianova, line 227:
```suggestion
    # Enable Textlab in the test region
```

tsconfig.json
```diff
@@ -22,6 +22,6 @@
     /* Skip type checking of declaration files. */
     "skipLibCheck": true,
     /* Disallow inconsistently-cased references to the same file. */
-    "forceConsistentCasingInFileNames": true
-  }
+    "forceConsistentCasingInFileNames": true,
+  },
```
seluianova, line 26:
```suggestion
    "forceConsistentCasingInFileNames": true
  }
```

MariaKabanova: This was changed by ./tools/code_style.sh

tests/textlab_api/textlab_api_test.py
MariaKabanova: Yes, thanks, it makes sense. Fixed it

tests/textlab_api/textlab_api_test.py
MariaKabanova: Fixed

tests/textlab_api/textlab_api_test.py
MariaKabanova: Fixed

tests/textlab_api/textlab_api_test.py
MariaKabanova: Fixed

tests/textlab_api/textlab_api_test.py
MariaKabanova: Fixed

tests/textlab_api/textlab_api_test.py
MariaKabanova: Fixed

MariaKabanova:
One of new tests is failing. I think it would be better to fix it on CMS side. When TextLab API returns response with not a successful HTTP code (e.g. 400 like in test), it should ignore any value in the response body, even a valid one, like {"formulaHix": 15.48920568}

Could someone experienced please help with this? I am not sure how I can fix it in code...

tests/textlab_api/textlab_api_test.py
seluianova: There are currently 2 potential reasons why a HIX score will not be obtained in this test:
- HIX is disabled on region level
- Textlab API server (our mock) returns 500

If a defect occurs that will cause the hix_enabled attribute to be ignored, we won't notice it, because this test will remain "green".

Thus, to check that the HIX score is actually not updated, when hix is disabled at the region level, we need to configure a positive response from the mock server.

tests/textlab_api/textlab_api_test.py
```diff
@@ -13,3 +28,373 @@ def test_normalize_text(
...
+        "status": status.PUBLIC,
+        "hix_ignore": "on",
+    }
+
+    response = client.post(edit_page, edit_page_data)
+
+    assert response.status_code == 302
+    assert response.headers.get("Location") == edit_page
+
+    # Check the updated HIX score
```
seluianova, line 211:
```suggestion
    # Check that the HIX score has not been updated
```

seluianova (changes requested):
Thanks for the proposed tests!
I came up with a few suggestions.

I believe we could also introduce an explicit check for „No request to TextLab is expected“.
Pls take a look at respond_with_handler() function in docs:
https://pytest-httpserver.readthedocs.io/en/latest/howto.html#using-custom-request-handler
An Implementation idea is:
- create an empty list for incoming requests
- in the handler() function, perform adding the request to the list before returning Response()
- at the end, check the size of the request list (if it’s empty - no request was sent)

Feel free to ping me if you need any help or further clarification.

tests/textlab_api/textlab_api_test.py
seluianova: There are still 2 potential reasons why a HIX score will not be obtained in this test:

HIX is disabled on region level
Textlab API server (our mock) returns 500
Thus, if a defect occurs that will cause the hix_enabled attribute to be ignored, we won't notice it, because the HIX score will be None anyway.

But if we configure a positive response here, it will help to catch this kind of defect.

seluianova: There are still 2 potential reasons why a HIX score will not be obtained in this test:

- HIX is disabled on region level
- Textlab API server (our mock) returns 500

Thus, if a defect occurs that will cause the hix_enabled attribute to be ignored, we won't notice it, because the HIX score will be None anyway.

But if we configure a positive response here, it will help to catch this kind of defect.

tests/textlab_api/textlab_api_test.py
```diff
@@ -13,3 +28,373 @@ def test_normalize_text(
...
+
+    :param load_test_data: The fixture providing the test data (see :meth:`~tests.conftest.load_test_data`)
+    :param login_role_user: The fixture providing the http client and the current role (see :meth:`~tests.conftest.login_role_user`)
+    :param settings: The fixture providing the django settings
+    :param httpserver: The fixture providing the dummy http server
+    """
+    # Setup a mocked Textlab API server with dummy responses
+    httpserver.expect_request("/user/login").respond_with_json({"token": "dummy"})
+    httpserver.expect_request("/benchmark/5").respond_with_json(
+        {"formulaHix": 15.48920568}, status=400
```
seluianova, line 361: Also it doesn't make much difference in the current case, but normally error responses (400, 500) imply that the request was not fulfilled.
So we are unlikely to get the error status code and the successful hix calculation result in the same response.
It would be more natural to return smthg like:
`{"message": "Error occurred"}`

tests/textlab_api/textlab_api_test.py
MariaKabanova: Good point! Fixed

tests/textlab_api/textlab_api_test.py
```diff
@@ -13,3 +28,373 @@ def test_normalize_text(
...
+    # Redirect call aimed at the Textlab API to the fake server
+    settings.TEXTLAB_API_URL = f"http://localhost:{httpserver.port}"
+
+    # Enable Textlab in the test region
+    Region.objects.filter(slug="augsburg").update(hix_enabled=True)
+
+    # Log the user in
+    client, _role = login_role_user
+
+    # Update page content
```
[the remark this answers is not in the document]
MariaKabanova: Extracted a function. Looks cleaner now, thank you!

tests/textlab_api/textlab_api_test.py
MariaKabanova: Fixed in all places

tests/textlab_api/textlab_api_test.py
MariaKabanova: Thank you for the idea with respond_with_handler(). Just pushed the update.

tests/textlab_api/textlab_api_test.py
```diff
@@ -13,3 +32,316 @@ def test_normalize_text(
...
+        "content": content,
+        "mirrored_page_region": "",
+        "_ref_node_id": 3,
+        "_position": "right",
+        "status": status.PUBLIC,
+    }
+    if hix_ignore:
+        edit_page_data["hix_ignore"] = "on"
+
+    return (edit_page, client.post(edit_page, edit_page_data))
```
seluianova, line 67:
```suggestion

      edit_page = reverse(
        "edit_page",
        kwargs={
            "region_slug": "augsburg",
            "language_slug": "de",
            "page_id": 2,
        },
    )
    edit_page_data = {
        "title": title,
        "content": content,
        "mirrored_page_region": "",
        "_ref_node_id": 3,
        "_position": "right",
        "status": status.PUBLIC,
    }
    if hix_ignore:
        edit_page_data["hix_ignore"] = "on"

    return (edit_page, login_role_user[0].post(edit_page, edit_page_data))
```

tests/textlab_api/textlab_api_test.py
```diff
@@ -13,3 +32,316 @@ def test_normalize_text(
...
+    login_role_user: tuple[Client, str],
+    title: str,
+    content: str,
+    hix_ignore: bool = False,
+) -> tuple[str, HttpResponse]:
+
+    # Log the user in
+    client, _role = login_role_user
+
+    # Update page content
```
seluianova, line 47:
```suggestion
```
Nit-picky comment:
Since the whole function is called "update_page_content", I think the additional clarifying comment can be omitted.

seluianova, line 47:
```suggestion
```
Nit-picky comment:
Since the whole function is called "update_page_content", I think the additional clarifying comment can be omitted.

tests/textlab_api/textlab_api_test.py
seluianova, line 67:
```suggestion
    return (edit_page, login_role_user[0].post(edit_page, edit_page_data))
```
Another nit-peaky and very optional comment:
Since we don't use _role variable here anyway, we may not create it.

PeterNerlich:
Giving this high priority as per request from the tech meeting (any issues regarding HIX). Not sure this changes much here and it is only about testing the existing functionality anyway, so don't feel pressured to reschedule your work-life-balance just to get this out a day or two earlier.

integreat_cms/core/circleci_settings.py
```diff
@@ -25,6 +25,12 @@
 DEEPL_AUTH_KEY = "dummy"
 #: Enable manually because existing setting derives from the unset env var
 DEEPL_ENABLED = True
+#: Set dummy Textlab key to test automatic translations via Textlab API
+TEXTLAB_API_KEY = "dummy"
+#: Enable manually because existing setting derives from the unset env var
+TEXTLAB_API_ENABLED = True
+#: Use debug logging on CircleCI
+LOG_LEVEL = "DEBUG"
```
seluianova, line 33:
```suggestion
```

tests/textlab_api/textlab_api_test.py
```diff
@@ -0,0 +1,345 @@
...
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize("login_role_user", [EDITOR], indirect=True)
+def test_hix_score_update(
+    load_test_data: None,
+    login_role_user: tuple[Client, str],
+    settings: SettingsWrapper,
+    httpserver: HTTPServer,
+) -> None:
```
seluianova, line 65:
```suggestion
def test_hix_score_update(
    load_test_data: None,
    admin_client: Client,
    settings: SettingsWrapper,
    httpserver: HTTPServer,
) -> None:
```
I suggest using admin_client in these tests instead of login_role_user because we're testing hix_signals that trigger regardless of user role anyway.
If we use admin_client here, it will just simplify the code a little bit.

The update_page_content() function and all other tests should be updated accordingly then.

tests/textlab_api/textlab_api_test.py
```diff
@@ -0,0 +1,345 @@
...
+from integreat_cms.cms.models.regions.region import Region
+
+if TYPE_CHECKING:
+    from django.test.client import Client
+    from pytest_django.fixtures import SettingsWrapper
+    from pytest_httpserver.httpserver import HTTPServer
+
+
+def update_page_content(
+    login_role_user: tuple[Client, str],
```
seluianova, line 26:
```suggestion
    admin_client: Client,
```

tests/textlab_api/textlab_api_test.py
```diff
@@ -0,0 +1,345 @@
...
+    title: str,
+    content: str,
+    hix_ignore: bool = False,
+) -> tuple[str, HttpResponse]:
+
+    # Log the user in
+    client, _role = login_role_user
+
+    # Update page content
```
seluianova, line 35:
```suggestion
```

tests/textlab_api/textlab_api_test.py
```diff
@@ -0,0 +1,345 @@
...
+        "content": content,
+        "mirrored_page_region": "",
+        "_ref_node_id": 3,
+        "_position": "right",
+        "status": status.PUBLIC,
+    }
+    if hix_ignore:
+        edit_page_data["hix_ignore"] = "on"
+
+    return (edit_page, client.post(edit_page, edit_page_data))
```
seluianova, line 55:
```suggestion
    return (edit_page, admin_client.post(edit_page, edit_page_data))
```

tests/textlab_api/textlab_api_test.py
```diff
@@ -0,0 +1,345 @@
...
+        return Response(json.dumps({"message": "Error occurred"}), status=500)
+
+    # Setup a mocked Textlab API server with dummy responses
+    httpserver.expect_request("/user/login").respond_with_json({"token": "dummy"})
+    httpserver.expect_request("/benchmark/420").respond_with_handler(handler)
+
+    # Redirect call aimed at the Textlab API to the fake server
+    settings.TEXTLAB_API_URL = f"http://localhost:{httpserver.port}"
+
+    # Enable Textlab in the test page
```
seluianova, line 276:
```suggestion
    # Enable Textlab in the test region
```

tests/textlab_api/textlab_api_test.py
```diff
@@ -0,0 +1,345 @@
...
+    Region.objects.filter(slug="augsburg").update(hix_enabled=True)
+
+    # Setting HIX score initial value
+    initial_hix_score = 11
+    translationId = Page.objects.get(id=2).get_translation("de").latest_version.pk
+    translation = PageTranslation.objects.get(id=translationId)
+    translation.latest_version.hix_score = initial_hix_score
+    translation.save()
```
seluianova, line 284:
```suggestion
    PageTranslation.objects.filter(page__id=2, language__slug="de").update(hix_score=initial_hix_score)
```

A little shorter

tests/textlab_api/textlab_api_test.py
```diff
@@ -0,0 +1,345 @@
...
+
+    # Check that initial page properties
```
seluianova, line 286:
```suggestion
    # Check initial page properties
```

tests/textlab_api/textlab_api_test.py
```diff
@@ -0,0 +1,345 @@
...
+    assert page_translation.latest_version.hix_score == initial_hix_score
+
+    edit_page, response = update_page_content(
+        login_role_user, "Neuer Titel", content_before_change
+    )
+
+    assert response.status_code == 302
+    assert response.headers.get("Location") == edit_page
+
+    # Check that the HIX score has not been updated
```
seluianova, line 299:
```suggestion
    # Check that the HIX score has been copied from the previous version
```

tests/textlab_api/textlab_api_test.py
```diff
@@ -0,0 +1,345 @@
...
+@pytest.mark.django_db
+@pytest.mark.parametrize("login_role_user", [EDITOR], indirect=True)
+def test_hix_no_content_changes(
+    load_test_data: None,
+    login_role_user: tuple[Client, str],
+    settings: SettingsWrapper,
+    httpserver: HTTPServer,
+) -> None:
+    """
+    Check that the HIX score is not updated when hix is enabled on page level and page content is not updated. No request to TextLab API is expected.
```
seluianova, line 255:
```suggestion
    Check that the HIX score is copied from the previous translation version if the page content is not changed. No request to TextLab API is expected.
```

MariaKabanova:
Thank you @seluianova for your comments and suggestions!
Ready for final review round now

tests/textlab_api/textlab_api_test.py
```diff
@@ -0,0 +1,339 @@
...
+
+    # Check that HIX score was not updated
+    page_translation = Page.objects.get(id=2).get_translation("de")
+
+    assert page_translation.hix_score is None
+    assert text_api_requests_count == 0
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize("admin_client", [EDITOR], indirect=True)
```
seluianova, line 148:
```suggestion
```

tests/textlab_api/textlab_api_test.py
```diff
@@ -0,0 +1,339 @@
...
+
+    # Check that the HIX score has not been updated
+    page_translation = Page.objects.get(id=2).get_translation("de")
+
+    assert page_translation.hix_score is None
+    assert text_api_requests_count == 0
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize("admin_client", [EDITOR], indirect=True)
```
seluianova, line 242:
```suggestion
```

tests/textlab_api/textlab_api_test.py
```diff
@@ -0,0 +1,339 @@
...
+    # Check that the HIX score has been copied from the previous version
+    page_translation = Page.objects.get(id=2).get_translation("de")
+
+    assert page_translation.hix_score == initial_hix_score
+    assert page_translation.content == content_before_change
+    assert text_api_requests_count == 0
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize("admin_client", [EDITOR], indirect=True)
```
seluianova, line 302:
```suggestion
```

seluianova (approved):
Thanks for all the effort, looks good to me now.

There is also an opportunity to reduce code duplication in the mock server configuration, using the mock_server fixture, introduced in https://github.com/digitalfabrik/integreat-cms/pull/2748, but if this PR will be merged before 2748, I will adjust it myself.

seluianova:
@MariaKabanova I have merged the new mock_server fixture https://github.com/digitalfabrik/integreat-cms/pull/2748, could you please update your tests to use it?

seluianova:
@MariaKabanova the new mock_server fixture has been merged https://github.com/digitalfabrik/integreat-cms/pull/2748, could you please update your tests to use it?

tests/textlab_api/textlab_api_test.py
```diff
@@ -0,0 +1,334 @@
...
+
+    return (edit_page, admin_client.post(edit_page, edit_page_data))
+
+
+@pytest.mark.django_db
+def test_hix_score_update(
+    load_test_data: None,
+    admin_client: Client,
+    settings: SettingsWrapper,
+    httpserver: HTTPServer,
```
seluianova, line 59:
```suggestion
    mock_server: MockServer,
```

tests/textlab_api/textlab_api_test.py
```diff
@@ -0,0 +1,334 @@
...
+) -> None:
+    """
+    Check that the HIX score is requested and saved when the page translation is updated
+
+    :param load_test_data: The fixture providing the test data (see :meth:`~tests.conftest.load_test_data`)
+    :param admin_client: The fixture providing the http client
+    :param settings: The fixture providing the django settings
+    :param httpserver: The fixture providing the dummy http server
```
seluianova, line 67:
```suggestion
    :param mock_server: The fixture providing the dummy http server
```

tests/textlab_api/textlab_api_test.py
```diff
@@ -0,0 +1,334 @@
...
+    text_api_requests_count = 0
+
+    def handler(request: Request) -> Response:
+        nonlocal text_api_requests_count
+        text_api_requests_count += 1
+        return Response(json.dumps({"formulaHix": 15.48920568}), status=200)
+
+    # Setup a mocked Textlab API server with dummy responses
+    httpserver.expect_request("/user/login").respond_with_json({"token": "dummy"})
+    httpserver.expect_request("/benchmark/420").respond_with_handler(handler)
```
seluianova, line 78:
```suggestion
    # Setup a mocked Textlab API server with dummy responses
    mock_server.configure("/user/login", 200, {"token": "dummy"})
    mock_server.configure("/benchmark/420", 200, {"formulaHix": 15.48920568})
```

tests/textlab_api/textlab_api_test.py
```diff
@@ -0,0 +1,334 @@
...
+
+    # Redirect call aimed at the Textlab API to the fake server
+    settings.TEXTLAB_API_URL = f"http://localhost:{httpserver.port}"
```
seluianova, line 81:
```suggestion
    settings.TEXTLAB_API_URL = f"http://localhost:{mock_server.port}"
```

tests/textlab_api/textlab_api_test.py
```diff
@@ -0,0 +1,334 @@
...
+    )
+
+    assert response.status_code == 302
+    assert response.headers.get("Location") == edit_page
+
+    # Check the updated HIX score
+    page_translation = Page.objects.get(id=2).get_translation("de")
+
+    assert page_translation.hix_score == 15.48920568
+    assert text_api_requests_count == 1
```
seluianova, line 97:
```suggestion
    assert mock_server.requests_counter == 1
```

MariaKabanova:
@seluianova thank you for your suggestions! 
I updated my code respectively

.gitignore
```diff
@@ -52,6 +52,7 @@ docs/src/templates/breadcrumbs.html*
 cms-django.iml
 integreat-cms.iml
 webpack-stats.json
+.pytest_cache
```
PeterNerlich, line 55: Why do you add this to the `.gitignore`? I never noticed `.pytest_cache` before, but having a quick look at my system it seems that pytest automatically generates another `.gitignore` inside it which is also recognized by the top level git repository – making this change redundant.

tests/textlab_api/textlab_api_test.py
```diff
@@ -0,0 +1,302 @@
...
+
+    # Redirect call aimed at the Textlab API to the fake server
+    settings.TEXTLAB_API_URL = f"http://localhost:{mock_server.port}"
+
+    # Enable Textlab in the test region
+    Region.objects.filter(slug="augsburg").update(hix_enabled=True)
+
+    edit_page, response = update_page_content(
+        admin_client, "Willkommen in Augsburg", "Neuer Inhalt2", True
+    )
```
PeterNerlich, line 162: Slight preference from me: by the time I read to here I already forgot the fourth argument to `update_page_content()` and was confused about where you actually set the page to ignore the HIX score. If we pass it explicitly as a kwarg, this should be abundantly clear and no future Peter should be able to just overlook it. :grin: 

```suggestion
    edit_page, response = update_page_content(
        admin_client, "Willkommen in Augsburg", "Neuer Inhalt2", hix_ignore=True
    )
```

tests/textlab_api/textlab_api_test.py
```diff
@@ -0,0 +1,302 @@
...
+
+@pytest.mark.django_db
+def test_hix_response_400(
+    load_test_data: None,
+    admin_client: Client,
+    settings: SettingsWrapper,
+    mock_server: MockServer,
+) -> None:
+    """
+    Check that the HIX score is requested but response 400 received. Hix score is not updated
```
PeterNerlich, line 275: What do you think about this wording? I think it's a bit clearer.

```suggestion
    Check that the HIX score is not being updated when the API returns a HTTP 400 error.
```

PeterNerlich (approved):
Thanks for your work! :tada:  I really like the tests as I find them easy to understand, and the docstrings seem also very thoughtful.

I have a few comments but nothing that I find critical, so feel free to incorporate them or not based on your own opinion before merging.

.gitignore
MariaKabanova: Hi Peter. Thank you for your review!
Just checked, indeed this .pytest_cache is not required. Removed. I added it while switching my branch to python 3.11

PR #2680, opened 2024-03-01
complete tree_mutex

svenseeberg:


integreat_cms/cms/views/pages/page_actions.py
```diff
@@ -11,6 +11,7 @@
 import uuid
 from typing import TYPE_CHECKING
 
+from db_mutex import DBMutexError, DBMutexTimeoutError
```
svenseeberg, line 14: I guess this can be removed.

PR #2682, opened 2024-03-02
README: nodejs 20 is required

svenseeberg:
### Short description
It seems that Node.js 18.x is not longer sufficient to run eslint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ This content management system helps local integration experts to provide multil
 Following packages are required before installing the project (install them with your package manager):
 
 * `npm` version 7 or later
-* `nodejs` version 18 or later
+* `nodejs` version 20 or later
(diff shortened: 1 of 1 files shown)


timobrembeck (changes requested):
Thanks! :+1:

Could you also change this here:
https://github.com/digitalfabrik/integreat-cms/blob/d2e629f294d85dae1b83ed87a19a1346c4566652/tools/install.sh#L76
and here:
https://github.com/digitalfabrik/integreat-cms/blob/d2e629f294d85dae1b83ed87a19a1346c4566652/docs/src/installation.rst?plain=1#L17

david-venhoff (approved):

svenseeberg:
Maybe the root cause is something else. Let me check again.

Issue #2692, opened 2024-03-05
Adjust message for bulk action success/failure/warning messages


osmers:
### Motivation
The last update #2540 made feedback messages of bulk actions rather unreadable. The list of pages published/archived/set to draft/uploaded is too long. The list needs to be easier to read.

### Proposed Solution
Show only the title of 3-5 page names in the feedback message, and add the number of other pages that received a feedback message.
E.g. Der Veröffentlichunsstatus der folgenden Seiten wurde erfolgreich zu "Entwurf" geändert: "Willkommen", "Willkommen in [Stadt]", "Wissenswertes über [Stadt]", "Zusammenleben in Deutschland", "Stadtplan" und 97 weitere.

### Alternatives
Go back to the version before PR #2540 

### User Story
As a user of the CMS I want to be able to quickly see, how many pages received which feedback message so that I can evaluate whether everything worked the way I wanted it to. -->

### Additional Context
Currently, it looks like this:
![image](https://github.com/digitalfabrik/integreat-cms/assets/64901227/dc22f735-dd0f-4b8b-855f-a73d61635a86)
But it looks especially bad for translation uploads: 
![image (3)](https://github.com/digitalfabrik/integreat-cms/assets/64901227/89067c6d-35e9-4e32-9c23-9fe837e581f4)

### Design Requirements
Maybe decision on how many pages should be shown and also how to separate the page names more clearly.

PR #2693, opened 2024-03-11
Limit the maximum number of items displayed in feedback messages

david-venhoff:
### Short description

This pr limits the maximum number of items displayed in feedback messages in order to avoid visual clutter.

### Proposed changes

Add a parameter to `iter_to_str` to stop after a given number of objects (default 5).
If there are more objects to be displayed, the message will only note that there are more objects. 
 

### Resolved issues

Fixes: #2692 

### DIFF
--- a/integreat_cms/cms/utils/stringify_list.py
+++ b/integreat_cms/cms/utils/stringify_list.py
@@ -3,23 +3,37 @@ from collections.abc import Iterable
 from django.utils.translation import gettext_lazy as _
 
 
-def iter_to_string(iterable: Iterable, *, quotation_char: str = '"') -> str:
+def iter_to_string(
+    iterable: Iterable, *, quotation_char: str = '"', max_items: int = 5
+) -> str:
     """
     This is a helper function that turns a list into a string using different delimiters.
 
     :param iterable: List of items
     :param quotation_char: The character that is added around each item in the list
+    :param max_items: The maximum number of items in the iterable to display
     :return: Returns string with the single items separated by comma expect the last item which is separated by 'and'.
     """
     # Convert iterable to list to support querysets etc.
-    lst = list(iterable)
-    # If the list contains more than 1 element, save the last element for later
-    last_element = lst.pop() if len(lst) > 1 else None
+    elements = list(iterable)
+    last_element_to_display = elements.pop() if 1 < len(elements) <= max_items else None
     # Join remaining elements with commas and surrounding quotes
     separator = quotation_char + ", " + quotation_char
-    list_str = quotation_char + separator.join(map(str, lst)) + quotation_char
-    # Append the last element with "and"
-    if last_element:
-        list_str += " " + _("and") + f" {quotation_char}{last_element}{quotation_char}"
-    # Return final string
-    return list_str
+    elements_str = (
+        quotation_char + separator.join(map(str, elements[:max_items])) + quotation_char
+    )
+
+    # Handle the last element with 'and' and note that there are more elements if limited by `max_items`
+    number_elements_over_limit = len(elements) - max_items
+    if number_elements_over_limit == 1:
+        elements_str += " " + _("and 1 other")
+    elif number_elements_over_limit > 1:
+        elements_str += " " + _("and {} others").format(number_elements_over_limit)
+    elif last_element_to_display:
+        elements_str += (
+            " "
+            + _("and")
+            + f" {quotation_char}{last_element_to_display}{quotation_char}"
+        )
+
+    return elements_str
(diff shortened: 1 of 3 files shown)


MizukiTemma (approved):
Looks good 👍

integreat_cms/cms/utils/stringify_list.py
```diff
@@ -3,23 +3,30 @@
...
     """
     This is a helper function that turns a list into a string using different delimiters.
 
     :param iterable: List of items
     :param quotation_char: The character that is added around each item in the list
+    :param max_items: The maximum number of items in the iterable to display
     :return: Returns string with the single items separated by comma expect the last item which is separated by 'and'.
     """
     # Convert iterable to list to support querysets etc.
     lst = list(iterable)
```
JoeyStk, line 18:
```suggestion
    list_of_elements = list(iterable)
```

I'd like to make this a bit more specific and follow Robert C. Martin's recommendation in Clean Code chapter 2.5 to use pronounceable names as variables

integreat_cms/cms/utils/stringify_list.py
```diff
@@ -3,23 +3,30 @@
...
-    # If the list contains more than 1 element, save the last element for later
-    last_element = lst.pop() if len(lst) > 1 else None
+    last_element = lst.pop() if 1 < len(lst) <= max_items else None
     # Join remaining elements with commas and surrounding quotes
     separator = quotation_char + ", " + quotation_char
-    list_str = quotation_char + separator.join(map(str, lst)) + quotation_char
+    list_str = (
```
JoeyStk, line 22:
```suggestion
    list_of_elements_as_str = (
```

integreat_cms/cms/utils/stringify_list.py
```diff
@@ -3,23 +3,30 @@
...
+        quotation_char + separator.join(map(str, lst[:max_items])) + quotation_char
+    )
     # Append the last element with "and"
     if last_element:
         list_str += " " + _("and") + f" {quotation_char}{last_element}{quotation_char}"
-    # Return final string
+    elif len(lst) - max_items == 1:
```
JoeyStk, line 28: This might be controversial but in order to make the code a bit more readable I would prefer to put this conditional into a variable like:

only_one_more_element_left = len(lst) - max_items == 1

and then use it here as

```suggestion
    elif only_one_more_element_left:
```
In my option this improves the code a little. But since I couldn't find a clear guideline on this, I'd also be okay if you decide to reject this suggestion :)

If you decide to follow the suggestion, I'd like it if you could also adapt something like this for the line below

JoeyStk (approved):
Thank you very much. This looks good. I left a few ideas on how we could improve the readability of the surrounding code. I'm aware that this is on me as I didn't edit the code when I copied it from Lunes to Integreat, but I hope you are fine with doing the cleaning now :)
As mentioned below, I tried to base my suggestions on information gathered either from the presentation at the conference, the book the presentation was based on or on comments in forums, however I realize that there is still room for debate on whether or not the rules actually apply in these cases

integreat_cms/cms/utils/stringify_list.py
david-venhoff: I would also take the subchapter 'avoid encodings' of that book into account, and propose to just rename it to 'elements'. Do you agree?

integreat_cms/cms/utils/stringify_list.py
david-venhoff: I went ahead and added a variable `number_elements_over_limit` and also renamed the variable `last_element` to make it more descriptive. Could you take a look at the changes? :)

integreat_cms/cms/utils/stringify_list.py
```diff
@@ -3,23 +3,37 @@
...
-    return list_str
+    elements_str = (
+        quotation_char + separator.join(map(str, elements[:max_items])) + quotation_char
+    )
+
+    # Handle the last element with 'and' and note that there are more elements if limited by `max_items`
+    number_elements_over_limit = len(elements) - max_items
+    if number_elements_over_limit == 1:
+        elements_str += " " + _("and 1 other")
+    elif number_elements_over_limit > 0:
```
JoeyStk, line 30: Why did you choose to check if number_elements_over_limit is larger than 0 and not if it's larger than 1?

integreat_cms/cms/utils/stringify_list.py
JoeyStk: Looks good! I like the changes. This is better than what I had suggested. I only have one question left, which you can find below :)

integreat_cms/cms/utils/stringify_list.py
david-venhoff: For no particular reason I think. I justed wanted to check that we are over the limit at all.
But I'll change it now to 1.

PR #2694, opened 2024-03-11
Fix region selection in push news form

timobrembeck:
### Short description

Fix the bug where checkboxes became disabled because of the overflow-scroll property in Chromium and Firefox on Windows.


PeterNerlich (approved):
Cannot confirm that this fixes the bug that I could never reproduce on my machine, but it doesn't break existing functionality.

Issue #2670, opened 2024-02-27
Provide option to translate into Plain German


dkehne:
### Motivation
Marburg-Biedenkopf wants to have plain German in their translation of easy German. Therefore we need to overwrite the current translations once.

### Proposed Solution
- Add new setting for SUMM.AI to differentiate between plain and easy German (default: easy)
- Allow to set this setting via environment variable & config file
- When the value is set to "plain", set the `output_language_level` field in the SUMM.AI API request to `plain`

https://summai.notion.site/SUMM-AI-API-Documentation-fc41fa5ef1434ad2b311be4876b4685a

Hint: At the moment, this is just a test flight, so no need to store this value persistent per region etc., so when the translation is updated via the page form, the "easy" German translation can be used.

PR #2701, opened 2024-03-17
Add option to translate into Plain German

MizukiTemma:
### Proposed changes

- Add a new environment variable `SUMM_AI_PLAIN_GERMAN_REGIONS` to list regions that want to use plain German in the bulk MT action.
- Set `output_language_level` to `plain` if the region is listed in `SUMM_AI_PLAIN_GERMAN_REGIONS` **and** the MT translation request is made per bulk action.
- Change the language name in the description of bulk SUMM.AI translation to `German (plain)` if it is the case of `output_language_level = "plain"`.

### Side effects

- None? 🙈 

### Resolved issues

Fixes: #2670

### DIFF
--- a/integreat_cms/core/settings.py
+++ b/integreat_cms/core/settings.py
@@ -1011,6 +1011,14 @@ SUMM_AI_IS_INITIAL: Final[bool] = bool(
     strtobool(os.environ.get("INTEGREAT_CMS_SUMM_AI_IS_INITIAL", "True"))
 )
 
+# Slugs of regions that prefer Plain German over Easy German in the management command
+SUMM_AI_PLAIN_GERMAN_REGIONS: Final[list[str]] = [
+    x.strip()
+    for x in os.environ.get(
+        "INTEGREAT_CMS_SUMM_AI_PLAIN_GERMAN_REGIONS", ""
+    ).splitlines()
+]
+
 
 ################
 # STATIC FILES #
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -9835,6 +9835,15 @@ msgstr ""
 #~ msgid "The DeepL budget has been exceeded."
 #~ msgstr "Das DeepL-Budget wurde überschritten."
 
+#~ msgid "German (Plain)"
+#~ msgstr "Deutsch (leicht)"
+
+#~ msgid ""
+#~ "Machine translate %(content_type)s via %(provider)s to German (Plain)"
+#~ msgstr ""
+#~ "%(content_type)s maschinell via %(provider)s auf Deutsch (einfach) "
+#~ "übersetzen"
+
 #~ msgid "Language Selection"
 #~ msgstr "Sprachauswahl"
 
--- a/integreat_cms/summ_ai_api/summ_ai_api_client.py
+++ b/integreat_cms/summ_ai_api/summ_ai_api_client.py
@@ -92,6 +92,12 @@ class SummAiApiClient(MachineTranslationApiClient):
         logger.debug("Translating %r", text_field)
         # Use test region for development
         user = settings.TEST_REGION_SLUG if settings.DEBUG else self.region.slug
+        # Set the language level to "plain" if the region prefers Plain German
+        output_language_level = (
+            "plain"
+            if self.request.region.slug in settings.SUMM_AI_PLAIN_GERMAN_REGIONS
+            else "easy"
+        )
         if (
             text_field is None
             or (isinstance(text_field, TextField) and not text_field.text)
@@ -111,6 +117,7 @@ class SummAiApiClient(MachineTranslationApiClient):
                     "separator": settings.SUMM_AI_SEPARATOR,
                     "is_test": settings.SUMM_AI_TEST_MODE,
                     "is_initial": settings.SUMM_AI_IS_INITIAL,
+                    "output_language_level": output_language_level,
                 },
             ) as response:
                 # Wait for the response
@@ -164,7 +171,7 @@ class SummAiApiClient(MachineTranslationApiClient):
                 # asynchronously as a task. If we have to repeat the task
                 # (e.g. if we run into rate limiting and have to resend the request),
                 # we need a NEW coroutine object.
-                # For that case, we a representation of our function which can be
+                # For that case, we need a representation of our function which can be
                 # evaluated when needed, giving a new coroutine for the task each time.
                 partial(self.translate_text_field, session, text_field)
                 for text_field in text_fields
--- a/example-configs/integreat-cms.ini
+++ b/example-configs/integreat-cms.ini
@@ -143,6 +143,9 @@ SUMM_AI_EASY_GERMAN_LANGUAGE_SLUG = "de-si"
 SUMM_AI_SEPARATOR = "hyphen"
 # Whether the SUMM.AI translations are initial by default [optional, defaults to True]
 SUMM_AI_IS_INITIAL = True
+# Slugs of regions that prefer Plain German over Easy German in the management command
+SUMM_AI_PLAIN_GERMAN_REGIONS = "augsburg
+testregion"
 
 [MT global]
 # The amount of yearly translation credits each region receives for free


david-venhoff (commented):
Thanks, I tested this pr and it works.

As far as I understand from the issue, the use case is that a municipality wants to use plain German instead of easy German and we want to allow for this as a test flight. But is there an argument why we cannot add plain German as a new language and add it as a target language to the summ_ai_provider? The current way seems quite confusing to me because it conflates those two "languages". Also it seems a bit hacky to me to special case the ui just to differentiate between easy German and plain German. 

Another thing that was not clear to me from the issue is why we only want to apply this to bulk actions.

MizukiTemma:
> As far as I understand from the issue, the use case is that a municipality wants to use plain German instead of easy German and we want to allow for this as a test flight. But is there an argument why we cannot add plain German as a new language and add it as a target language to the summ_ai_provider? The current way seems quite confusing to me because it conflates those two "languages". Also it seems a bit hacky to me to special case the ui just to differentiate between easy German and plain German.
> 

I thought the minimal change is wished so Plain Germn can be tested with less waiting time for implementation, but we can do that too. What do you thingk @dkehne ?

> Another thing that was not clear to me from the issue is why we only want to apply this to bulk actions.

@dkehne ?

MizukiTemma:
@dkehne 
Then, is it enough for now that only the bulk action name is "Machine translate pages/events/locations via SUMM.AI to German (Plain)" insted of "... German (Easy)" and it translates into plain German with `output_language_level = "plain"` if the region is listed in `SUMM_AI_PLAIN_GERMAN_REGIONS` as wishing for plain German, without changing anything further?

MizukiTemma:
@dkehne 

The slug of region needs to be added once manually to the environment variable `SUMM_AI_PLAIN_GERMAN_REGIONS`

integreat_cms/cms/templates/events/event_list.html
```diff
@@ -158,9 +158,15 @@ <h1 class="heading">
...
-                            {% endblocktranslate %}
+                            {% if is_plain_german_region %}
+                                {% blocktranslate trimmed with provider=MT_PROVIDER.name %}
+                                    Machine translate {{ content_type }} via {{ provider }} to German (Plain)
+                                {% endblocktranslate %}
+                            {% else %}
+                                {% blocktranslate trimmed with provider=MT_PROVIDER.name %}
+                                    Machine translate {{ content_type }} via {{ provider }} to {{ language }}
+                                {% endblocktranslate %}
+                            {% endif %}
```
david-venhoff, line 169:
```suggestion
                            {% blocktranslate trimmed with provider=MT_PROVIDER.name %}
                                Machine translate {{ content_type }} via {{ provider }} to {{ language_display_name }}
                            {% endblocktranslate %}
```

integreat_cms/cms/templates/pages/page_tree.html
```diff
@@ -180,9 +180,15 @@ <h1 class="heading">
...
-                            {% endblocktranslate %}
+                            {% if is_plain_german_region %}
+                                {% blocktranslate trimmed with provider=MT_PROVIDER.name %}
+                                    Machine translate {{ content_type }} via {{ provider }} to German (Plain)
+                                {% endblocktranslate %}
+                            {% else %}
+                                {% blocktranslate trimmed with provider=MT_PROVIDER.name %}
+                                    Machine translate {{ content_type }} via {{ provider }} to {{ language }}
+                                {% endblocktranslate %}
+                            {% endif %}
```
david-venhoff, line 191:
```suggestion
                            {% blocktranslate trimmed with provider=MT_PROVIDER.name %}
                                Machine translate {{ content_type }} via {{ provider }} to {{ language_display_name }}
                            {% endblocktranslate %}
```

integreat_cms/cms/templates/pois/poi_list.html
```diff
@@ -147,9 +147,15 @@ <h1 class="heading">
...
-                            {% endblocktranslate %}
+                            {% if is_plain_german_region %}
+                                {% blocktranslate trimmed with provider=MT_PROVIDER.name %}
+                                    Machine translate {{ content_type }} via {{ provider }} to German (Plain)
+                                {% endblocktranslate %}
+                            {% else %}
+                                {% blocktranslate trimmed with provider=MT_PROVIDER.name %}
+                                    Machine translate {{ content_type }} via {{ provider }} to {{ language }}
+                                {% endblocktranslate %}
+                            {% endif %}
```
david-venhoff, line 158:
```suggestion
                            {% blocktranslate trimmed with provider=MT_PROVIDER.name %}
                                Machine translate {{ content_type }} via {{ provider }} to {{ language_display_name }}
                            {% endblocktranslate %}
```

integreat_cms/cms/views/events/event_list_view.py
```diff
@@ -117,5 +117,7 @@ def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
                 "search_query": query,
                 "source_language": region.get_source_language(language.slug),
                 "content_type": "events",
+                "is_plain_german_region": self.request.region.slug
+                in settings.SUMM_AI_PLAIN_GERMAN_REGIONS,
```
david-venhoff, line 121:
```suggestion
                "language_display_name": _("German (Plain)") if self.request.region.slug in settings.SUMM_AI_PLAIN_GERMAN_REGIONS else str(language)
```

integreat_cms/cms/views/pages/page_tree_view.py
```diff
@@ -154,5 +154,7 @@ def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
                 "filter_form": filter_form,
                 "XLIFF_EXPORT_VERSION": settings.XLIFF_EXPORT_VERSION,
                 "hix_threshold": settings.HIX_REQUIRED_FOR_MT,
+                "is_plain_german_region": self.request.region.slug
+                in settings.SUMM_AI_PLAIN_GERMAN_REGIONS,
```
david-venhoff, line 158:
```suggestion
                "language_display_name": _("German (Plain)") if self.request.region.slug in settings.SUMM_AI_PLAIN_GERMAN_REGIONS else str(language)
```

integreat_cms/cms/views/pois/poi_list_view.py
```diff
@@ -126,6 +126,8 @@ def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
                 "search_query": query,
                 "source_language": region.get_source_language(language.slug),
                 "content_type": "locations",
+                "is_plain_german_region": self.request.region.slug
+                in settings.SUMM_AI_PLAIN_GERMAN_REGIONS,
```
david-venhoff, line 130:
```suggestion
                "language_display_name": _("German (Plain)") if self.request.region.slug in settings.SUMM_AI_PLAIN_GERMAN_REGIONS else str(language)
```

integreat_cms/core/settings.py
```diff
@@ -1005,6 +1005,16 @@
...
+# Slugs of regions that prefer Plain German over Easy German
+SUMM_AI_PLAIN_GERMAN_REGIONS: Final[str | list[str] | None] = os.environ.get(
+    "INTEGREAT_CMS_SUMM_AI_PLAIN_GERMAN_REGIONS", []
+)
+
+# Path of MT bulk action
+BULK_MT_PATH: Final[str] = os.environ.get(
+    "INTEGREAT_CMS_BULK_MT_PATH", "machine-translate"
+)
+
```
david-venhoff, line 1017:
```suggestion
```

I wouldn't add a setting for that, since the path cannot change without a code update anyways

integreat_cms/summ_ai_api/summ_ai_api_client.py
```diff
@@ -207,6 +222,15 @@ def translate_queryset(self, queryset: list[Page], language_slug: str) -> None:
             settings.SUMM_AI_EASY_GERMAN_LANGUAGE_SLUG
         )
 
+        # Set the language level to "plain" if the region prefers Plain German and it is a bulk action
+        is_plain_german_region = (
+            self.request.region.slug in settings.SUMM_AI_PLAIN_GERMAN_REGIONS
+        )
+        is_bulk_action = self.request.path.split("/")[-2] == settings.BULK_MT_PATH
```
david-venhoff, line 229:
```suggestion
        bulk_mt_path = "machine-translate"
        is_bulk_action = self.request.path.split("/")[-2] == bulk_mt_path
```

david-venhoff (approved):
I'll approve now given the close deadline. However, just to be sure, the plan is to eventually remove this code again, right?

MizukiTemma:
@david-venhoff 
Thank you for the suggestions 🐱 They are now applied!

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -8346,6 +8349,11 @@ msgstr ""
 "Sie haben nicht die nötige Berechtigung, um Veranstaltungen zu bearbeiten "
 "oder zu erstellen."
 
+#: cms/views/events/event_list_view.py cms/views/pages/page_tree_view.py
+#: cms/views/pois/poi_list_view.py
+msgid "German (Plain)"
+msgstr "Deutsch (leicht)"
```
JoeyStk, line 8355: Am I understanding this correctly that for the municipality we just give them plain German instead of easy German but under the same language label or should the name of the language be renamed to Deutsch (einfach)? @dkehne

JoeyStk (commented):
Looks good so far. I only have two minor comments/questions

integreat_cms/core/settings.py
```diff
@@ -1005,6 +1005,11 @@
     strtobool(os.environ.get("INTEGREAT_CMS_SUMM_AI_IS_INITIAL", "True"))
 )
 
+# Slugs of regions that prefer Plain German over Easy German
+SUMM_AI_PLAIN_GERMAN_REGIONS: Final[str | list[str] | None] = os.environ.get(
+    "INTEGREAT_CMS_SUMM_AI_PLAIN_GERMAN_REGIONS", []
+)
```
PeterNerlich, line 1011: So how does this translate into specifying the environment variable?
```sh
$ INTEGREAT_CMS_SUMM_AI_PLAIN_GERMAN_REGIONS=augsburg, testregion
bash: testregion: command not found
$ INTEGREAT_CMS_SUMM_AI_PLAIN_GERMAN_REGIONS=augsburg,testregion python
Python 3.11.8 (main, Feb 12 2024, 14:50:05) [GCC 13.2.1 20230801] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.environ.get("INTEGREAT_CMS_SUMM_AI_PLAIN_GERMAN_REGIONS", [])
'augsburg,testregion'
>>> 
```
Well, we need to do an additional step here. Environmental variables will only be strings, but luckily we're using slugs that won't contain special characters like, say, commas, so we can use them as separators:
```suggestion
# Slugs of regions that prefer Plain German over Easy German
SUMM_AI_PLAIN_GERMAN_REGIONS: Final[str | list[str] | None] = [x.strip() for x in os.environ.get(
    "INTEGREAT_CMS_SUMM_AI_PLAIN_GERMAN_REGIONS", ""
).split(",")]
```

MizukiTemma: 🤔? I've imagined something like SUMM_AI_PLAIN_GERMAN_REGIONS = ["augsburg", "nurnberg"]

PeterNerlich: This works in python, but not in bash or anywhere you would set environment variables

PeterNerlich: Oh wait, I was hasty. `splitlines()` is used at a few places in the settings.py. Better be consistent, even if I wasn't sure whether it was trivial making environment variables contain newlines.
```suggestion
# Slugs of regions that prefer Plain German over Easy German
SUMM_AI_PLAIN_GERMAN_REGIONS: Final[str | list[str] | None] = [
    x.strip()
    for x in os.environ.get(
        "INTEGREAT_CMS_SUMM_AI_PLAIN_GERMAN_REGIONS", ""
    ).splitlines()
]
```

integreat_cms/summ_ai_api/summ_ai_api_client.py
```diff
@@ -226,7 +251,9 @@ def translate_queryset(self, queryset: list[Page], language_slug: str) -> None:
         asyncio.set_event_loop(loop)
 
         # Translate queryset asynchronously in parallel
-        loop.run_until_complete(self.translate_text_fields(loop, text_fields))
+        loop.run_until_complete(
+            self.translate_text_fields(loop, text_fields, output_language_level)
+        )
```
timobrembeck, line 256:
```suggestion
        loop.run_until_complete(self.translate_text_fields(loop, text_fields))
```

Hmm, as I think about it, I guess it makes sense to place the code into the `translate_text_field()` method itself, so we don't have to carry the argument around.
Just set the additional flag right before the API call to the SUMM.AI API:

https://github.com/digitalfabrik/integreat-cms/blob/c2e9ea75443abc2b640866631e32b85b7b6e4ed5/integreat_cms/summ_ai_api/summ_ai_api_client.py#L91

integreat_cms/summ_ai_api/summ_ai_api_client.py
```diff
@@ -166,7 +175,12 @@ async def translate_text_fields(
                 # we need a NEW coroutine object.
                 # For that case, we a representation of our function which can be
                 # evaluated when needed, giving a new coroutine for the task each time.
-                partial(self.translate_text_field, session, text_field)
+                partial(
+                    self.translate_text_field,
+                    session,
+                    text_field,
+                    output_language_level,
+                )
```
timobrembeck, line 183:
```suggestion
                partial(self.translate_text_field, session, text_field)
```
Then this can be reverted as well

integreat_cms/summ_ai_api/summ_ai_api_client.py
```diff
@@ -207,6 +222,16 @@ def translate_queryset(self, queryset: list[Page], language_slug: str) -> None:
...
+        # Set the language level to "plain" if the region prefers Plain German and it is a bulk action
+        is_plain_german_region = (
+            self.request.region.slug in settings.SUMM_AI_PLAIN_GERMAN_REGIONS
+        )
+        bulk_mt_path = "machine-translate"
+        is_bulk_action = self.request.path.split("/")[-2] == bulk_mt_path
+        output_language_level = (
+            "plain" if is_plain_german_region and is_bulk_action else "easy"
+        )
+
```
timobrembeck, line 234:
```suggestion
```
And this could be moved into `translate_text_field()`.

integreat_cms/summ_ai_api/summ_ai_api_client.py
```diff
@@ -142,7 +147,10 @@ async def translate_text_field(
         return text_field
 
     async def translate_text_fields(
-        self, loop: AbstractEventLoop, text_fields: Iterator[TextField]
+        self,
+        loop: AbstractEventLoop,
+        text_fields: Iterator[TextField],
+        output_language_level: str = "easy",
```
timobrembeck, line 153:
```suggestion
        self, loop: AbstractEventLoop, text_fields: Iterator[TextField]
```

integreat_cms/summ_ai_api/summ_ai_api_client.py
```diff
@@ -152,6 +160,7 @@ async def translate_text_fields(
 
         :param loop: The asyncio event loop
         :param text_fields: The text fields to be translated
+        :param output_language_level: Plain German or Easy German
```
timobrembeck, line 163:
```suggestion
```

integreat_cms/summ_ai_api/summ_ai_api_client.py
```diff
@@ -192,6 +206,7 @@ def abort_function(task: partial, reason: Any) -> None:
             all_results = chain(worker_results)
             return all_results
 
+    # pylint: disable=too-many-locals
```
timobrembeck, line 209:
```suggestion
```

timobrembeck (changes requested):
@MizukiTemma Thanks a lot! :muscle: 

Sorry if I was a bit unclear in the issue description, we do not have to differ between the bulk action and the forms, I just meant that we didn't have to support the form stuff if it is more work. But it does not make sense to insert extra code lines just to make it not work in the forms.
So to make it a bit more general, I would suggest to move the code right before the API call itself, so it will just be respected always, also when it's called via management command.

MizukiTemma:
@JoeyStk @PeterNerlich @timobrembeck

Thank you so much for the suggestions and comments 😸 I hope all the changes are applied correctly 🙈

integreat_cms/core/settings.py
```diff
@@ -1011,6 +1011,14 @@
     strtobool(os.environ.get("INTEGREAT_CMS_SUMM_AI_IS_INITIAL", "True"))
 )
 
+# Slugs of regions that prefer Plain German over Easy German in the management command
+SUMM_AI_PLAIN_GERMAN_REGIONS: Final[str | list[str] | None] = [
```
david-venhoff, line 1015:
```suggestion
SUMM_AI_PLAIN_GERMAN_REGIONS: Final[list[str]] = [
```
This is always a list, right?

integreat_cms/summ_ai_api/summ_ai_api_client.py
```diff
@@ -92,6 +92,12 @@ async def translate_text_field(
         logger.debug("Translating %r", text_field)
         # Use test region for development
         user = settings.TEST_REGION_SLUG if settings.DEBUG else self.region.slug
+        # Set the language level to "plain" if the region prefers Plain German and it is a bulk action
```
david-venhoff, line 95:
```suggestion
        # Set the language level to "plain" if the region prefers Plain German
```

timobrembeck (approved):
Great, thanks! :+1:
Could you squash your commits before merging?

Issue #1168, opened 2022-02-15
Update file paths in content when media file is replaced


timobrembeck:
### Describe the Bug

At the moment, we keep the file path identical when we're replacing media files.
This has the disadvantage that the app needs to download the files again to check whether it has changed. It would be more efficient if we treat filenames more like unique ids of files and change them in the content when they are replaced.

### Steps to Reproduce

1. Insert media file into content of page
2. Replace file in media library
3. Refresh preview of page
4. Even though the timestamp in the media library itself is updated, the content remains the same.

### Expected Behavior

When a media file is replaced in the media library, the new image should also be shown in the apps.

We can use the `Link` model of the linkcheck app to get the mapping between media files and page translations which contain the file. Then, we can update the url in the content.

### Actual Behavior

Since the file path doesn't change, it is cached at the moment

### Additional Information

Related issues:
- #1851

osmers:
Referring here to issue 1742 - it would still be helpful to have an updated and an original date, so that I know that the file has been updated - it if is not too much hassle

steffenkleinle:
Could you notify the app team in [this issue](https://issues.tuerantuer.org/browse/IGAPP-1193) once this is done please? Thanks :)

sarahsporck:
I'm wondering what makes most sense here. Currently files are stored like this: `{subdirectory}/{strftime('%Y/%m')}/{filename}` and the new (replacing) file only modifies certain attributes of the `media_file`.

- upload the new file at e.g. `{subdirectory}/{int(time())}/{filename}`
- delete the old file at the other path

Am I missing something? :)

timobrembeck:
Hmm, I guess I would stick with the year/month naming scheme instead of `int(time())`, so maybe it would already suffice to just remove the following lines?
https://github.com/digitalfabrik/integreat-cms/blob/cd9508fd7700b663ebb76a78a04d9c9c1d55e83f/integreat_cms/cms/models/media/media_file.py#L46-L55

And I think Django does not automatically delete the old file, yes so we'd have to do that manually.
And after that, we need to to a search & replace the paths in the whole content to make sure we don't get a bunch of 404 errors on the old file path.

sarahsporck:
Thank you for the hint. Was a little bit stuck here.

PR #2705, opened 2024-03-19
Update media file paths

david-venhoff:
### Short description

Update media file paths when uploading a new file in order to avoid caching problems.

### Proposed changes

Instead of making sure that a path gets reused when a media file is replaced, always create a new path for the new file, so that no caching problems occur.
This is done by prepending the epoch time to the filename.

Then replace all internal links to the old path with links to the new
path.

### Side effects

- Hopfully none, old file paths should still work

### Resolved issues

Fixes: #1168

### DIFF
--- a/integreat_cms/cms/forms/media/replace_media_file_form.py
+++ b/integreat_cms/cms/forms/media/replace_media_file_form.py
@@ -11,7 +11,8 @@ from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 
 from ...constants import allowed_media
-from ...models import MediaFile
+from ...models import MediaFile, User
+from ...utils.linkcheck_utils import replace_links
 from ...utils.media_utils import generate_thumbnail
 from ..custom_model_form import CustomModelForm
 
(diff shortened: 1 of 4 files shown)


david-venhoff:
Note to self: update https://github.com/digitalfabrik/integreat-app/issues/2130 when this is merged

MizukiTemma:
Is there any possibility to reproduce the bug locally?
Trying the steps to reproduce in the issue description with Redis aktivated, it won't work 🤔

david-venhoff:
For more information about the current workaround the app team currently uses, you can take a look at their issue: https://github.com/digitalfabrik/integreat-app/issues/2130
With this pr, the path always changes which removes the need to redownload the files all the time.

MizukiTemma (approved):
Cool 😻 LGTM

deen13 (commented):
I know that other frameworks offer support for cache invalidation, which might help in this case, but I don't know if that is supported in Django. However, adding a time to the path seems like a simple and solid solution to me. 😊

JoeyStk (approved):

Issue #2549, opened 2023-11-13
Machine translations do not publish when translation always was draft


svenseeberg:
### Describe the Bug
When a page translation in a translated language never was published before and the machine translation is used for the source language, the newly created version has the status draft.

### Steps to Reproduce

1. Create a new page and publish the page in the main language.
2. Create a target translation and save as draft.
3. Use the machine translation of the main language.
4. The target translation status is still draft after the machine translation.

### Expected Behavior
When a machine translation is used, the status of the translated page should always be the same as in the source language.

### Actual Behavior
The status of a translation that was draft is not changed to public.

### Additional Information
- https://chat.tuerantuer.org/digitalfabrik/pl/q8mf9h9bkjyadd84afit9aic5e
- #1883: "Automatic translation with trustful automatic services within our CMS (e.g. DeepL) -> Import should result in status "automatic translation" and should be directly visible in our app (publication status: PUBLIC)"
- https://chat.tuerantuer.org/digitalfabrik/threads/q8mf9h9bkjyadd84afit9aic5e

PeterNerlich:
Why exactly is the expected behaviour that machine translation should always be public? What advantage does it have over the assumption that machine translations should inherit the status of the page they were translated from?

JoeyStk:
@svenseeberg Do you maybe have some more insights?

svenseeberg:
> @svenseeberg Do you maybe have some more insights?

I added a link with reports from users.

svenseeberg:
> What advantage does it have over the assumption that machine translations should inherit the status of the page they were translated from?

I guess that is not what is happening right now. But maybe this needs to be re-validated.

osmers:
I want to add that I tested this today and apparently, as soon as a page is saved as draft, all previous versions become draft versions even if they were published at some point. I can see that in the version history. 
I think @PeterNerlich is on the right track that translations should behave as the parent page does. So when I save a German page as draft with MT enabled, the translations should be translated and saved as draft. If I then publish the German page with MT enabled, the translations should be translated and published. Same goes for MT enabled and updating the German page. But I think that is not the problem right now.
If I save a German page as draft/or update it with MT disabled, the translations should maintain their publication status and change their translation status to outdated.

Does that make sense?

JoeyStk:
Discussed in the grooming and questions could be resolved and were added to the issue description

vthejay:
A new example reported by Stadt Augsburg:
[...]

PR #2710, opened 2024-03-20
Make machine translations always adopt the status of the source translation

PeterNerlich:
### Short description

Make machine translations always adopt the status of the source translation

### Proposed changes

- Remove code to use status of an existing translation in that language

### Side effects

- 

### Resolved issues

Fixes: #2549

### DIFF
--- a/integreat_cms/deepl_api/deepl_api_client.py
+++ b/integreat_cms/deepl_api/deepl_api_client.py
@@ -133,11 +133,7 @@ class DeepLApiClient(MachineTranslationApiClient):
                     continue
 
                 data = {
-                    "status": (
-                        existing_target_translation.status
-                        if existing_target_translation
-                        else source_translation.status
-                    ),
+                    "status": source_translation.status,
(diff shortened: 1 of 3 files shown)


david-venhoff (commented):
Looks good to me. I think however that it would make sense to unify the status behavior across all translations, and currently summ ai always translates into `status.REVIEW`. Or is it intended that we have different behavior depending on which translation provider is used?

osmers:
Uff, good catch - I think SUMM.AI translates into review bcs the municipalities need to (and can) correct the translations. That is something they cannot do with the foreign languages due to missing language skills. With SUMM.AI however, it is necessary bcs mostly the translation cannot be used as is.

david-venhoff (approved):
> With SUMM.AI however, it is necessary bcs mostly the translation cannot be used as is.

Alright, this looks good to me then

MizukiTemma (approved):
Looks good 👍 Thank you 😸

Issue #2690, opened 2024-03-03
Adjust title of translation box in content forms


dkehne:
### Motivation
Too make the UI cleaner and use as less space as possible we may remove the explanation text from the "Automatic Translation" because it's intuitve after we have made it to one of the key featurese/benefits of Integreat the last month. Also there is some kind of expansion on all languages after clicking the button so it's really clear what the function does.

### Proposed Solution
- ~Remove the explanation text of the "Automatic Translation" checkbox.~
- Adjust also the German title of the block to "Auswirkungen auf Übersetzungen".

See screenshot

<img width="807" alt="adjust-translation-text" src="https://github.com/digitalfabrik/integreat-cms/assets/4182469/b96609f9-0bff-4b40-ac39-87e5b09f404f">

osmers:
after talking to @dkehne we'll leave the description text as it is (so no removing necessay) but can definitely change the title of the box

PR #2713, opened 2024-03-26
Adjust translation box title of content form

lunars97:
### Short description

Remove the unnecessary word "another" from the title of the automatic translation text of **minor edit** box so that it was cleaner

### Proposed changes

- Adjust the German title of the block to "Auswirkungen auf Übersetzungen"

### Resolved issues

Fixes: #2690 

### DIFF
--- a/integreat_cms/cms/templatetags/content_filters.py
+++ b/integreat_cms/cms/templatetags/content_filters.py
@@ -99,9 +99,7 @@ def minor_edit_label(region: Region, language: Language) -> Promise:
     language_node = region.language_node_by_slug[language.slug]
     if language_node.is_leaf():
         return _("Implications on the translation status")
-    if language_node.is_root():
-        return _("Implications on other translations")
-    return _("Implications on the status of this and other translations")
+    return _("Implications on translations")
 
 
 @register.simple_tag
--- a/integreat_cms/cms/templates/imprint/imprint_sbs.html
+++ b/integreat_cms/cms/templates/imprint/imprint_sbs.html
@@ -199,7 +199,7 @@
                         </label>
                         {% render_field imprint_translation_form.content|add_error_class:"border-red-500" id="target_translation_tinymce" class="tinymce_textarea" %}
                         <label>
-                            {% translate "Implications on other translations" %}
+                            {% translate "Implications on translations" %}
                         </label>
                         {% render_field imprint_translation_form.minor_edit %}
                         <label for="{{ imprint_translation_form.minor_edit.id_for_label }}"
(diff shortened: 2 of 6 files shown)


integreat_cms/cms/templatetags/content_filters.py
```diff
@@ -100,7 +100,7 @@ def minor_edit_label(region: Region, language: Language) -> Promise:
     if language_node.is_leaf():
         return _("Implications on the translation status")
     if language_node.is_root():
-        return _("Implications on other translations")
+        return _("Implications on translations")
     return _("Implications on the status of this and other translations")
```
timobrembeck, line 104: @dkehne I assume that you also wanted to make the heading more clear for the languages other than the default language (e.g. English, Arabic etc.), right?

At the moment, this looks like this:

Screenshot 2024-03-26 at 17-55-22 Integreat Redaktionssystem

I guess this should probably also be changed to "Implications on translations"?

integreat_cms/cms/templatetags/content_filters.py
```diff
@@ -100,7 +100,7 @@ def minor_edit_label(region: Region, language: Language) -> Promise:
     if language_node.is_leaf():
         return _("Implications on the translation status")
     if language_node.is_root():
```
timobrembeck, line 102:
```suggestion
```
Then we could simplify this `if` statement.

integreat_cms/cms/templatetags/content_filters.py
```diff
@@ -100,7 +100,7 @@ def minor_edit_label(region: Region, language: Language) -> Promise:
...
-        return _("Implications on other translations")
+        return _("Implications on translations")
```
timobrembeck, line 103:
```suggestion
    return _("Implications on translations")
```

integreat_cms/cms/templatetags/content_filters.py
timobrembeck, line 104:
```suggestion
```

timobrembeck (changes requested):
@lunars97 Awesome, thanks a lot for your first PR! :tada: :muscle: 

In general it looks very good! :+1:
I just assume that we could simplify the box heading for non-root languages as well.

integreat_cms/cms/templatetags/content_filters.py
lunars97: yep, thank you for your review. I will work on the rest :)

MizukiTemma (approved):

timobrembeck (commented):
Sorry about the long delay!
Could you squash your commits into one? Then I think this is ready to be merged! :tada:

timobrembeck (approved):

Issue #2722, opened 2024-04-02
Cannot update a sent push notification sometimes


david-venhoff:
### Describe the Bug

Sometimes an error message shows up that a required field is not set, even though the mode is set to fallback to default language:

![grafik](https://github.com/digitalfabrik/integreat-cms/assets/78504586/43059db7-5d68-47fd-a2d5-176cae47757e)

### Steps to Reproduce
1. Create a push notification in a region that has a root language which has an id that is bigger than the id of another language of that region
2. send it
3. try to update it
4. notice the error

### Expected Behavior

The pn can be updated normally

### Actual Behavior

The pn cannot be updated

### Additional Information

The problem seems to be that the code assumes that the default language of the region is the first language of the formset, while the actual first language of the formset is the language with the smallest id

PR #2723, opened 2024-04-02
Use the correct default language for pnt forms

david-venhoff:
### Short description

This pr fixes a bug where sometimes sent push notifications cannot be updated, because the cms incorrectly calculates the default language, which causes the wrong field to be required in the push notification translation formset.

### Proposed changes

- Don't assume that the default language is the first language in the query
- Explicitly filter for the default language

### Side effects

/

### Resolved issues

Fixes: #2722


david-venhoff:
Closing this because I noticed some bigger problems with the way we handle this currently. I will open a new pr to address those.

PR #2727, opened 2024-04-03
Require the title field for all region default languages in push notification form

david-venhoff:
### Short description

This pr makes the title field in the push notification form required for every language that is a default language in one of the push notifications regions. This fixes two issues:
- Previously, only the title field of the first language of the query was required, even if it differed from the primary regions language.
- With the first problem fixed, it was still undefined what happened if a push notification was assigned to two regions with different default languages. The title was arbitrarily required for the language of either region, but not both.

### Proposed changes

- Require the title field for all region default languages in push notification form
- Do a small refactor on the push notification form view post handler, to make it easier to add this functionality

### Side effects

It is still arbitrary what will be used as the default language if a push notification without title and text gets send. Maybe we should always use the default language of the region from which the push notification was sent?

### Resolved issues

Fixes: #2722 

### DIFF
--- a/integreat_cms/cms/views/push_notifications/push_notification_form_view.py
+++ b/integreat_cms/cms/views/push_notifications/push_notification_form_view.py
@@ -145,8 +145,6 @@ class PushNotificationFormView(TemplateView):
                 if language not in existing_languages
             ],
         )
-        # Make title of default language required
-        pnt_formset[0].fields["title"].required = True
 
         if details["disable_edit"]:
             # Mark fields disabled when push notification was already sent
@@ -167,7 +165,6 @@ class PushNotificationFormView(TemplateView):
             },
         )
 
-    # pylint: disable=too-many-branches, too-many-statements
     def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
         r"""
         Save and show form for creating or editing a push notification. Send push notification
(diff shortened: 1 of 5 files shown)


MizukiTemma (approved):
Works as expected 👍 

The only thing I would suggest is adding tests, though this can be done as another issue/PR too 🙂

JoeyStk:
> It is still arbitrary what will be used as the default language if a push notification without title and text gets send. Maybe we should always use the default language of the region from which the push notification was sent?

This is probably a question for @osmers or for the next weekly :)

integreat_cms/cms/views/push_notifications/push_notification_form_view.py
```diff
@@ -167,7 +165,6 @@ def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
             },
         )
 
-    # pylint: disable=too-many-branches, too-many-statements
```
JoeyStk, line 170: Nice! Thank you!

integreat_cms/cms/views/push_notifications/push_notification_form_view.py
```diff
@@ -344,6 +286,96 @@ def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
...
+        not_accessible_regions_warning = __(
+            _(
+                "This news is also assigned to regions you don't have access to: {}."
+            ).format(", ".join(map(str, details["other_regions"]))),
+            _("Thus you cannot edit or delete this news."),
+            _(
+                "Please ask the one responsible for all of these regions or contact an administrator."
+            ),
+        )
+        messages.error(
```
JoeyStk, line 316: From our Barcamp on Clean Code I remember that messages should be separated from the logic as they are on different levels. In addition to this I think this function is again a bit too long and does too many things. Do you think it makes sense to split this up a little bit?

JoeyStk (commented):
Thank you for your PR. Please find my comment below :)

integreat_cms/cms/views/push_notifications/push_notification_form_view.py
david-venhoff: I agree that it makes sense to split up the `validate_and_save_forms` function, and i pushed a commit with a separate function for validation and saving. I am not convinced that we should changed the way messages are created, because I think that it would increase complexity, rather than lower for those concrete functions. Or do you have a specific implementation of that in mind?

JoeyStk: Fair enough. I like it better like this already :)

JoeyStk (approved):

Issue #2726, opened 2024-04-03
Push notification title is not displayed sometimes


david-venhoff:
### Describe the Bug

When a push notification is not available in the backend language, its name will not be displayed.

![grafik](https://github.com/digitalfabrik/integreat-cms/assets/78504586/e0af9b0d-3f02-4118-ba70-3940dad5ab66)

PR #2728, opened 2024-04-03
Fix push notification bug

david-venhoff:
### Short description

This pr fixes a problem that the `best_translation` method of push notifications prefers translations without title over translations with title.

### Proposed changes

- Only return the backend translation if it has a title

### Side effects

- 
- 

### Resolved issues

Fixes: #2726

### DIFF
--- a/integreat_cms/cms/fixtures/test_data.json
+++ b/integreat_cms/cms/fixtures/test_data.json
@@ -4451,6 +4451,17 @@
     }
   },
   {
+    "model": "cms.pushnotificationtranslation",
+    "pk": 9,
+    "fields": {
+      "title": "",
+      "text": "English translation without title",
+      "language": 2,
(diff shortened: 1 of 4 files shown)


ulliholtgrave (approved):

JoeyStk (commented):
Looks good to me so far. I remember that in a team call we said that we want to add a test for every bug that occurs in order to make sure it doesn't happen again. Do you think a test would also make sense in this case or is this an exception to the rule?

david-venhoff:
@JoeyStk I have added a test for this bug

JoeyStk (approved):
Thank you for adding the test. To me this looks good now.
I also like the comments in your test. The way I think about comments yours are really helpful!

david-venhoff:
Thanks :)

Issue #2698, opened 2024-03-14
Add Poi category icons


f1sh1918:
### Motivation
For malte and netzwerk obdach we need two new category icons

### Proposed Solution
Please make the new icons available via dropdown and deliver it to the frontend via api:

- mental_health
- shelter

Find new svg icons here:
[integreat_categories_cms.zip](https://nextcloud.tuerantuer.org/index.php/f/3345518)

api should deliver `<icon_name>_<colorHex>`

MizukiTemma:
@f1sh1918 Are the new icons broken? When I open the svg data of a new icon nothing will be shown, while existing icons are correctly shown... 🤔

PR #2729, opened 2024-04-03
Add new POI category icons

MizukiTemma:
### Short description

This PR adds two new POI category icons (mental health and shelter)

### Proposed changes

- Add constants and svg files for the new icons
- <s>(Still in clearing: adjust API response, see the issue page)</s> Add a new API response (In case the current ´icon´ and ´color´ are not used in future, we can remove them)

### Side effects

- What should be the german translation of "shelter"? Is it "Shelter" in German too? 😁 

### Resolved issues

Fixes: #2698 


lunars97:
I think Unterkunft is a suitable word for shelter :)

JoeyStk:
> What should be the german translation of "shelter"? Is it "Shelter" in German too? 😁

I think "Unterkunft" certainly is an option, but in this case I'd probably prefer "Obdach" as it seems to have an immediate connection to the "Vernetztes Obdach" project. But maybe @f1sh1918 can give us some additional context on what the best translation for "shelter" would be :)

FritjofK:
@MizukiTemma and @JoeyStk 
Please use "**Notunterkünfte**" as the German translation for shelter. That is the exact wording that is desired for "VernetztesObdach".

JoeyStk (approved):
LGTM! I think in the near future we should think about whether we want our categories to be in plural or singular terms, but for now it's consistent with what we have :)

integreat_cms/release_notes/current/unreleased/2698.yml
```diff
@@ -0,0 +1,2 @@
+en: Add mental health and shelter as new POI category icons
+de: Füge neue POI-Kategoriesymbole hinzu
```
seluianova, line 2:
```suggestion
de: Füge Psychische Gesundheit und Notunterkünfte als neue POI-Kategoriesymbole hinzu
```

seluianova (approved):
LGTM, thanks!

PR #2730, opened 2024-04-04
Add a nix flake providing a development shell

charludo:
### Short description

(Since this has nothing to do with any issue or any plans we've discussed as a team, I feel that I need to preface this with: everything related to this was done on my own time, not during working hours.)

This PR adds a [nix flake](https://wiki.nixos.org/wiki/Flakes). It allows developers to utilize a standardized and completely reproducible development environment.

You might find this useful if you currently manage multiple python and/or node projects (possibly requiring conflicting versions), are a nix user already, or frequently re-setup the project; but more importantly I strongly believe this makes setting our project up significantly easier for new developers (speaking from experience from my own onboarding two years ago 😉):

Regardless of the system a person is on - any kind of Linux; MacOS; even Windows with WSL enabled - getting up and running with our project would boil down to

1.  clone this repo
2. install nix (and **only** nix, not python/node/...), for example with the awesome installer script from [zero-to-nix.com](https://zero-to-nix.com):
    ```bash
    curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | sh -s -- install
    ```
3. run `nix develop` in your shell

As a result, you'll have the exact same development environment available as me/anyone else following these steps.

### Proposed changes

- add a nix flake and corresponding lock file 

### Side effects

- none, except:
- I volunteer to give a bar camp about nix/nix develop/NixOS at the next conference 😄 

### Resolved issues

Fixes: n/a

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #2732, opened 2024-04-05
Add devcontainer definition

deen13:
### Short description

This Pull Request adds a devcontainer definition to make the project setup easier and streamline the onboarding process of new devs.

### Proposed changes

I encountered some difficulties during onboarding and initial setup of the project, which I aim to automate with this addition. Through discussions in the chat, I learned that most people here develop with VSCode, so I thought that a devcontainer would be the appropriate solution.

#### Challenges Encountered:

- Initially, I installed a too-new patch of Python 3.11, which caused some issues and required downgrading.
- The translate.sh script expects the GNU variant of sed, which needs to be installed separately on macOS. Additionally, it needs to be either changed in the script to gsed instead of sed or overridden with an alias. The base image of the devcontainer includes the correct sed directly.
- The project requires the programs gettext and libmagic as native dependencies, which are also included in the base image of the devcontainer.
- After starting the project for the first time, I couldn't log in because the database was still empty. The devcontainer automatically loads the test data.

### Side effects

There should be no major side effects besides that we might need to take care of both the nix-flake (#2730) and the devcontainer definition.

### Optimizations

The Docker image built for running the devcontainer is currently quite large, at 2.1 GB, and it contains a nested Docker installation. I propose we create our own Dockerfile that utilizes a native PostgreSQL setup and  implementing image caching if this feature is utilized by a significant number of users.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,116 @@
+{
+  "version": "2.0.0",
+  "tasks": [
+    {
+      "label": "Run Shell Script",
+      "type": "shell",
+      "command": "sh",
+      "args": ["${file}"],
+      "group": {
+        "kind": "build",
+        "isDefault": true
+      },
+      "problemMatcher": []
+    },
+    {
+      "label": "Run all python tests",
+      "type": "shell",
+      "command": "./tools/test.sh",
+      "problemMatcher": [],
+      "group": {
+        "kind": "test",
+        "isDefault": true
+      }
+    },
+    {
+      "label": "Run current python test",
+      "type": "shell",
+      "command": "./tools/test.sh ${file}",
+      "problemMatcher": [],
+      "group": {
+        "kind": "test",
+        "isDefault": true
+      }
+    },
+    {
+      "label": "Run all frontend tests",
+      "type": "shell",
+      "command": "npm run test",
+      "problemMatcher": [],
+      "group": {
+        "kind": "test",
+        "isDefault": true
+      }
+    },
+    {
+      "label": "Check code style",
+      "type": "shell",
+      "command": "./tools/code_style.sh",
+      "problemMatcher": [],
+      "group": {
+        "kind": "test",
+        "isDefault": true
+      }
+    },
+    {
+      "label": "Run prettier tool",
+      "type": "shell",
+      "command": "./tools/prettier.sh",
+      "problemMatcher": [],
+      "group": {
+        "kind": "test",
+        "isDefault": true
+      }
+    },
+    {
+      "label": "Run mypy tool",
+      "type": "shell",
+      "command": "./tools/prettier.sh",
+      "problemMatcher": [],
+      "group": {
+        "kind": "test",
+        "isDefault": true
+      }
+    },
+    {
+      "label": "Run ruff tool",
+      "type": "shell",
+      "command": "./tools/ruff.sh",
+      "problemMatcher": [],
+      "group": {
+        "kind": "test",
+        "isDefault": true
+      }
+    },
+    {
+      "label": "Run eslint tool",
+      "type": "shell",
+      "command": "./tools/eslint.sh",
+      "problemMatcher": [],
+      "group": {
+        "kind": "test",
+        "isDefault": true
+      }
+    },
+    {
+      "label": "Run devserver",
+      "type": "shell",
+      "command": "./tools/run.sh",
+      "problemMatcher": [],
+      "group": {
+        "kind": "test",
+        "isDefault": true
+      }
+    },
+    {
+      "label": "Run translation check",
+      "type": "shell",
+      "command": "./tools/check_translations.sh",
+      "problemMatcher": [],
+      "group": {
+        "kind": "test",
+        "isDefault": true
+      }
+    }
+  ]
+}
--- a/README.md
+++ b/README.md
@@ -12,36 +12,63 @@
 
 This content management system helps local integration experts to provide multilingual information for newcomers.
 
-## TL;DR
+## Development Setup
 
-### Prerequisites
-
-Following packages are required before installing the project (install them with your package manager):
-
-* `npm` version 7 or later
-* `nodejs` version 18 or later
-* `python3` version 3.11 or later
-* `python3-pip` (Debian-based distributions) / `python-pip` (Arch-based distributions)
-* `python3-venv` (only on Debian-based distributions)
-* `gettext` to use the translation features
-* Either `postgresql` **or** `docker` to run a local database server
-
-### Installation
+This section provides a brief overview of setting up the development environment. We support various environments: you can set up everything locally using your preferred package manager, use it as a devcontainer, or utilize the nix-flake ([coming soon](https://github.com/digitalfabrik/integreat-cms/pull/2730)). Please clone the repository with the following snippet before starting to 
+setup your development environment.
 
 ````
 git clone git@github.com:digitalfabrik/integreat-cms.git
 cd integreat-cms
-./tools/install.sh
 ````
+
+### Local Setup
+
+To configure your development environment on your system, please follow these steps carefully.
+
+1. Ensure that the following packages are installed alongside your preferred IDE:
+   - `npm` version 7 or later
+   - `nodejs` version 18 or later
+   - `python3` version 3.11 or later
+   - `python3-pip` (Debian-based distributions) / `python-pip` (Arch-based distributions)
+   - `python3-venv` (only on Debian-based distributions)
+   - `gettext` for translation features
+   - Either `postgresql` **or** `docker` to run a local database server
+2. Execute `tools/install.sh` to download all dependencies.
+3. Execute `tools/migrate.sh` to apply all database schema migrations.
+4. Optionally, run `tools/loadtestdata.sh` to apply a predefined set of test data.
+
+### Devcontainer
+
+To configure your development container, please follow these steps carefully.
+
+1. Make sure you have Docker and VSCode (not VSCodium) installed on your machine.
+2. Open the project in VSCode.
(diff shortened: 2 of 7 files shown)


deen13:
Just learned that .devcontainers are now also supported by Jetbrains Gateway, so this is not only for the VSCode folks. 🥳

https://youtrack.jetbrains.com/issue/IDEA-292050/Support-for-.devcontainer#focus=Comments-27-9448589.0-0

timobrembeck (commented):
Interesting idea! :+1:

Personally, I have never used .devcontainers, but I see how they make things easier with regard to different developer setups, managing the dev tools for different OS was a bit messy sometimes.
That said, I think most of your challenges could just be solved by a docker compose setup, not necessarily a .devcontainer setup. But it's an interesting idea, so I agree that it basically makes sense this way.

A few things I would suggest before merging such a change:
- Add information about this in the documentation
- Adapt the dev tools

> There should be no major side effects besides that we might need to take care of both the nix-flake (https://github.com/digitalfabrik/integreat-cms/pull/2730) and the devcontainer definition.

I didn't check out the other PR yet, but does it essentially tackle the same problems? If so, I'd suggest taking a vote in the team for the preferred solution and only provide one default dev setup which works independently of the user's OS. Having multiple "supported" options to setup the dev environment might not make onboarding easier in the future.

> The Docker image built for running the devcontainer is currently quite large, at 2.1 GB, and it contains a nested Docker installation.

But the nested docker installation can probably be removed by removing this line, right?

```
"ghcr.io/devcontainers/features/docker-in-docker:2": {},
```

In my opinion, concepts like docker-in-docker should be avoided if there is no good reason for it, and in our case the only reason would be the (soon to be legacy) tools, or am I missing something?

> I propose we create our own Dockerfile that utilizes a native PostgreSQL setup and implementing image caching if this feature is utilized by a significant number of users.

Not sure if I would invest the time to get a native postgres db up and running inside the main python devcontainer. In my opinion, the best practice in a container environment is to keep the separate services in separate containers. The same goes e.g. for optional redis servers etc. So I'd prefer the suggested docker compose setup over a combined all-in-one image.

charludo:
> I didn't check out the other PR yet, but does it essentially tackle the same problems? If so, I'd suggest taking a vote in the team for the preferred solution and only provide one default dev setup which works independently of the user's OS. Having multiple "supported" options to setup the dev environment might not make onboarding easier in the future.

I'd like to chhime in here, because yes, both solutions solve the same problem, just in a different way. I'm not sure if maintaining both isn't still an option though: For people using JetBrains or VSCode, the devcontainer setup might be easiest to get running, while the nix shell requires one additional dependency (nix itself) but is editor-agnostic. It should definitely be mentioned in the README though, that one or the other way can be chosen depending on one's current setup.

> That said, I think most of your challenges could just be solved by a docker compose setup, not necessarily a .devcontainer setup. 

TBF getting to rid my system of docker is one of the main benefits of the flake/devcontainer route (at least to me) 😅

deen13:
Hey @timobrembeck,

I brought up the topic in our weekly call and we reached a consensus to maintain both versions. It's worth noting that limiting ourselves to just one solution wouldn't necessarily simplify matters, as developers tend to configure their environments locally according to their preferences anyway. By retaining both options, we acknowledge this tendency and ensure that developers can still tailor their setups while also benefiting from shared solutions. Additionally, these solutions cater to different user groups; some may prefer more integrated tooling like devcontainers, while others feel more comfortable with an editor-agnostic setup in their ide.

charludo:
Well said @deen13. One thing I'd like to add: I can see an maintainence-effort argument against keeping both solutions, but the maintainence required should be minimal, since in both cases only the environment required to run the project is defined, the single source of truth for all python/node packages remains with the existing config files.

timobrembeck (commented):
@deen13 @charludo Alright makes sense, thanks for elaborating! :+1:

README.md
```diff
@@ -12,36 +12,61 @@
...
-* Either `postgresql` **or** `docker` to run a local database server
-
-### Installation
+This section provides a brief overview of setting up the development environment. We support various environments: you can set up everything locally using your preferred package manager, use it as a devcontainer, or utilize the nix-flake ([coming soon](https://github.com/digitalfabrik/integreat-cms/pull/2730)). Please clone the repository with the following snippet before starting to 
+setup your development environment.
 
 ````
 git clone git@github.com:digitalfabrik/integreat-cms.git
 cd integreat-cms
 ./tools/install.sh
```
charludo, line 23:
```suggestion
```
I think this gets covered below, right?

README.md
```diff
@@ -12,36 +12,61 @@
...
+
+For more information, refer to [this issue](https://github.com/microsoft/vscode-remote-release/issues/1729) in the VSCode Remote Extension repository.
+
+##### Test discovery 
+
+Testing is not yet fully integrated into our Visual Studio Code Container, but we provide convenient task definitions to facilitate testing. For example, you can simply type task current python test and hit enter to run the current Python test file. Take a look at the `.vscode/tasks.json` for more helpful tasks and extend this file as needed.
```
charludo, line 65:
```suggestion
Testing is not yet fully integrated into our Visual Studio Code Container, but we provide convenient task definitions to facilitate testing. For example, you can simply type `task current python test` and hit enter to run the current Python test file. Take a look at the `.vscode/tasks.json` for more helpful tasks and extend this file as needed.
```
(I assume this is what was meant? TBH I don't quite get this sentence, but then again I'm not really vamiliar with VSCode)

.vscode/tasks.json
```diff
@@ -0,0 +1,116 @@
...
+    {
+      "label": "Run all python tests",
+      "type": "shell",
+      "command": "./tools/test.sh",
+      "problemMatcher": [],
+      "group": {
+        "kind": "test",
+        "isDefault": true
+      }
+    },
```
charludo, line 24: Just out of curiosity, are these just for documentation purposes?

tools/install.sh
```diff
@@ -115,37 +115,41 @@ echo "Installing JavaScript dependencies..." | print_info
...
 
-# Check if virtual environment exists
-if [[ -d ".venv" ]] && [[ "$(.venv/bin/python3 --version)" != "$(${PYTHON} --version)" ]]; then
-    echo "The given $(${PYTHON} --version) version differs from $(.venv/bin/python3 --version) of virtual environment." | print_warning
-    echo "Deleting the outdated virtual environment..." | print_info
-    rm -rf .venv
-fi
+if [ "$(whoami)" = "vscode" ]; then
+    echo "Skipping the initialization of the virtual environment because this devcontainer is already an isolated environment on its own." | print_info
+else
```
charludo, line 120: There's at least one other place in the `require_installed` function  where the venv is being invoked. I'm not sure if it might not be easier to keep the venv-in-container setup, even if it's not really necessary.

(Unless there's a reason I don't know about here, of course)

charludo (changes requested):
Will give the actual setup a try later, below are just a couple of nitpicks :)

.vscode/tasks.json
deen13: No they're purpose is speed up the workflow. :) 

https://www.loom.com/share/1561aac259c347a9b9d35b335d728830

README.md
deen13: You've got it right! It's possible to run tests within vscode using an extension that adds a small run button next to each test definition, but I couldn't make it work in our setup. However, the loom I linked in my other comment demonstrates the alternative path using tasks, which makes it a bit more convenient to run the scripts. It's not perfect, but it's better than before. :D

tools/install.sh
deen13: The reason was that the mentioned test support in vscode seemed to work without the virtual environment (venv) because it was able to find the correct test runner. However, it only executed a handful of tests and skipped the rest. 😢

deen13: I reverted the commit so that we now use venv in devcontainer again. 👍

charludo: Hm, weird. Does it not find the tests, or do they fail to setup? I've had a similar-ish issue with the flake, requiring to set two environment variables to properly find and run all tests:
```bash
export DJANGO_SETTINGS_MODULE="integreat_cms.core.settings"
export INTEGREAT_CMS_DEBUG=1
```

.vscode/tasks.json
charludo: https://github.com/digitalfabrik/integreat-cms/assets/47758554/a2527d39-7007-4ee0-9df2-252f99eef920

:D

deen13: image

deen13: image

tools/install.sh
deen13: The setup worked, as I suggested because some tests were successful, while others were found but marked as skipped. :/

.vscode/tasks.json
charludo: 😂 

(Took more time than I care to admit to setup though)

charludo (commented):
Nice! I justed tested it out (to the best of my abilities - I've never really used VSCode before, so some of my notes may be 100% my fault):
- mabe a note would be good that this requires VSCode, not VSCosium? Since the devcontainer extension isn't available there
- speaking of which: I was not prompted to install the extension, and searching for it myself, I only found the `Dev Containers` extension, not as indicated the `Remote - Containers` extension
- the user **must** be in the `docker` group on linux, VSCode does not allow to optionally enter sudo password
- VSCode marked all non-project imports as unable to find them; probably related to the move back to the `venv`?

Apart from that though, very pleasant setup process! Good job!

deen13:
> VSCode marked all non-project imports as unable to find them; probably related to the move back to the venv?

Good catch @charludo! I fixed the `defaultInterperterPath`

charludo (approved):
🎉 

(Just remember to squash before merge :D)

MizukiTemma (approved):
🎉

Issue #2733, opened 2024-04-07
No serverside guard against deleting a POICategory which is still being used


PeterNerlich:
### Describe the Bug

Deleting POICategories which are still used by POIs are not supposed to be deleted. To prevent users from deleting them, the delete button in the list view is disabled. However, there seem to be no server-side checks in place, sending a request to the delete endpoint deletes the category anyway.
Discovered during #2720

### Steps to Reproduce

1. Go to *Network Management*
2. Go to *Location Categories*
3. See disabled delete button on categories still in use (*Number of locations*)
4. Select POICategory for edit
5. Note id of POICategory in URL
6. Send POST request without additional data to the delete endpoint (edit URL but ending in `/delete/` instead of `/edit/`)
7. Reload *Location Categories* and see that the POICategory was deleted

### Expected Behavior

POICategory cannot be deleted, server instead shows an error message

### Actual Behavior

POICategory was deleted

PR #2734, opened 2024-04-07
Display helpful message when deleting POICategory fails

PeterNerlich:
### Short description

Deleting a POICategory still in use by POIs is discouraged by disabling the delete button in the list view. If a request reaches the delete endpoint anyways, an internal server error is displayed. This PR catches the exception and displays a helpful message instead.

### Proposed changes

- Implement a custom `POICategoryDeleteView` which subclasses `CustomDeleteView` and acts as a simple wrapper around its `post()` logic:
- If a `ProtectedError` is raised, a helpful message is added to the request context and the browser is redirected back to the list view.

### Side effects

- If in the future there are other reasons how a `ProtectedError` could arise, the current message might be misleading.

### Resolved issues

Fixes: #2733

### DIFF
--- a/integreat_cms/cms/urls/protected.py
+++ b/integreat_cms/cms/urls/protected.py
@@ -362,9 +362,7 @@ urlpatterns: list[URLPattern] = [
                             ),
                             path(
                                 "delete/",
-                                delete_views.CustomDeleteView.as_view(
-                                    model=POICategory,
-                                ),
+                                poi_categories.POICategoryDeleteView.as_view(),
                                 name="delete_poicategory",
                             ),
                         ]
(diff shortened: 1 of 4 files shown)


MizukiTemma (approved):
Thank you for the improvement :) Looks good!

timobrembeck (approved):
Thanks for the quick fix! :muscle:

deen13 (approved):
Looking good from what I can tell. Is this change something to be mentioned in `integreat_cms/release_notes/current/unreleased/2734.ym`l?

JoeyStk:
> Looking good from what I can tell. Is this change something to be mentioned in integreat_cms/release_notes/current/unreleased/2734.yml

As this change is only going to be noted by the service-team and not by the municipality, we left it out yesterday, but also weren't 100% sure about it

PR #2736, opened 2024-04-08
Add tools/prune_pdf_cache.sh

PeterNerlich:
### Short description

This PR adds a helpful tool to prune the pdf file cache so you don't have to remember the which directory to delete whenever PDF tests fail but probably not because of you.

### Proposed changes

- adds `tools/prune_pdf_cache.sh` which essentially just runs `rm -rfv "integreat_cms/pdf"`
- list the tool in the documentation
- prune pdf cache before running tests

### Side effects

- pruning the PDF cache deletes the whole cache, not only the files specific to the tests


JoeyStk (approved):
LGTM!

MizukiTemma (approved):
Thank you 😸 

This nice feeling of finally having the thing you had been thinking about for a long time 😁

Issue #2619, opened 2024-01-24
[30.06.] Integreat Chat / Digital Counseling


svenseeberg:
### Motivation
Integreat mostly provides one way communication (except for e-mail addresses provided in feedback forms). We want to extend that functionality with a digital counseling solution that enables continuous bi-directional communication between app users and region managers/counselors.

### Proposed Solution
- The Integreat app search functionality is being extended to act like a chat. Messages are sent to the CMS API.
- The able to fetch responses from the API.
- Zammad acts as a back end (database) for storing messages. Messages (Strings) are sent from the app to an API endpoint of the CMS. The CMS stores the message in Zammad via the Zammad API.
- The Integreat CMS is responsible for mapping the app (temporary chat ID?) to Zammad ticket IDs. App users can close and start a new chat which will delete the ID on the app.
- Except for a region setting to enable the functionality and set a Zammad URL, no GUI in the CMS is required for this feature.
- To prevent abuse, an IP based rate limit has to be implemented.
- The Integreat CMS is a proxy for messages between the Integreat app and Zammad (per region).
- In the future the Integreat CMS can query other components (automatic translation services, a content search engine & LLMs) to create automated responses and translation services.
- If automatic responses are sent to the user, a copy is also attached to the Zammad ticket.

To sum up:
- We need to implement a pull and post messages API endpoint.
- The CMS needs a new model to store a map of device IDs to Zammad ticket IDs.
- The CMS uses Zammad issues to store messages.

Flow for posting messages from the app:
- The app sends a message (string + device ID) to the CMS
- The CMS looks in its mapping model if there already is an existing Zammad issue
- If not, a new issues is created and the message is stored
- if a ticket exists, the new message is appended

Flow for getting answers:
- the app sends a device ID
- The CMS maps the device ID into a Zammad ticket ID
- If the ticket exists, all messages are returned to the app

### Alternatives
- N/A

### User Story
As a migrant I want to directly chat with regional counseling services to avoid having to physically travel to a help desk.

### Additional Context
- https://nextcloud.tuerantuer.org/index.php/apps/drawio/edit?fileId=5750544
- https://chat.tuerantuer.org/digitalfabrik/pl/rci55kyyg7b83d9bu1yfqqxpoo
- https://docs.zammad.org/en/latest/api/ticket/index.html#create
- https://docs.zammad.org/en/latest/api/ticket/index.html#update

### Design Requirements
- None, no GUI

JoeyStk:
[...]

PR #2738, opened 2024-04-08
[14.06.] Implement Chat API

charludo:
### Short description

This PR adds an implementation for the Chat-API.

In the issue description, a deletion function is mentioned; however, there's no reasonable way to actually delete Zammad tickets/articles via API, so this will have to be solved in some other way.

### Proposed changes

- add a new model `UserChat` storing mappings of `device_id`s -> `zammad_id`s
- add `ZammadChatAPI`, a wrapper around `zammad_py`, plus corresponding API endpoints for our usecase
- implement ratelimiting
- add tests

### Side effects

- none I am aware of, this feature is very stand-alone 

### Resolved issues

Fixes: #2619 (partially)


svenseeberg:
> In the issue description, a deletion function is mentioned; however, there's no reasonable way to actually delete Zammad tickets/articles via API, so this will have to be solved in some other way.

You're refering to "App users can close and start a new chat which will delete the ID on the app."? This is basically just intended as a front end thing to start a new chat. This should not remove Zammad tickets.

JoeyStk:
Question for @svenseeberg. Does this PR make [#2568](https://github.com/digitalfabrik/integreat-cms/pull/2568) redundant?

charludo:
> Question for @svenseeberg. Does this PR make [#2568](https://github.com/digitalfabrik/integreat-cms/pull/2568) redundant?

Yes, but I opted for creating a new PR since these two only share the same goal, not any implementation details.

charludo:
> > In the issue description, a deletion function is mentioned; however, there's no reasonable way to actually delete Zammad tickets/articles via API, so this will have to be solved in some other way.
> 
> You're referring to "App users can close and start a new chat which will delete the ID on the app."? This is basically just intended as a front end thing to start a new chat. This should not remove Zammad tickets.

Right, looks like the only mention of deletion in the issue, my bad. I do however remember discussing this at the last conference, and felt it was right to point this out on here.

JoeyStk:
> Yes, but I opted for creating a new PR since these two only share the same goal, not any implementation details.

No worries. My underlying question is if we can close [#2568](https://github.com/digitalfabrik/integreat-cms/pull/2568) then :)

docs/src/api-docs.rst
```diff
@@ -1055,3 +1055,81 @@ Body:
...
+   }
+
+The response to ``GET``-ing the endpoint without an ``attachment_id`` is a list containing all chat messages.
+
+.. code:: javascript
+
+   {
+      "messages" : [                   // A list containing the chat messages
+         "id": Number,                 // message id
+         "body": String,               // the actual message content
```
svenseeberg, line 1112: Just a note here: we should be clear that Zammad uses HTML for message formatting.

integreat_cms/cms/models/chat/user_chat.py
```diff
@@ -0,0 +1,96 @@
...
+        )
+
+
+class UserChat(AbstractBaseModel):
+    """
+    A model for a user (app) chat, mapping a device ID to a Zammad ticket ID
+    """
+
+    device_id = models.CharField(max_length=200)
+    zammad_id = models.IntegerField()
```
svenseeberg, line 44: Just another design note for @steffenkleinle : one app can have multiple sessions, as the device_id must not be unique. In theory the app could provide an option to switch between multiple tickets.

svenseeberg (commented):
I looked into the overall design and I think it does what we need. I will next create a test setup and try it out.

charludo:
> I looked into the overall design and I think it does what we need. I will next create a test setup and try it out.

Great! Let me know if there's trouble, but I think the only important note I have is the one remarked in the Region form fields, i.e. both the "bot" account and any personnel wishing to be notified must have full group permissions for the `settings.USER_CHAT_TICKET_GROUP` group.

docs/src/api-docs.rst
charludo: Good point, will add a note.

docs/src/api-docs.rst
```diff
@@ -1055,3 +1055,81 @@ Body:
...
+
+    GET /api/v3/{region_slug}/{language_slug}/chat/{device_id}/ HTTP/2
+
+.. code:: http
+
+    GET /api/v3/{region_slug}/{language_slug}/chat/{device_id}/{attachment_id}/ HTTP/2
+
+.. code:: http
+
+   POST /api/v3/{region_slug}/{language_slud}/chat/{device_id}/ HTTP/2
```
steffenkleinle, line 1077:
```suggestion
   POST /api/v3/{region_slug}/{language_slug}/chat/{device_id}/ HTTP/2
```

steffenkleinle (commented):
LGTM, thank you :yellow_heart:

integreat_cms/cms/models/chat/user_chat.py
charludo: @steffenkleinle yes, but creating a new chat from the same device is already supported in this implementation.

charludo:
Blocked due to #2763  

Since this is ready for review apart from the mentioned issue, I've marked it as such regardless of the blocked status.

svenseeberg:
We had a UX Design meeting last week and we decided it would be best to have a random user exposure with a defined chance of the feature being visible to app users. That means we want to be able to set a chance value per region that allows the app to decide whether it should show the feature or not. Would it be possible to include a `chat_feature_probability` value per region that is also returned in the API? The app would do the rest.

charludo:
> We had a UX Design meeting last week and we decided it would be best to have a random user exposure with a defined chance of the feature being visible to app users. That means we want to be able to set a chance value per region that allows the app to decide whether it should show the feature or not. Would it be possible to include a `chat_feature_probability` value per region that is also returned in the API? The app would do the rest.

Maybe doing this in a separate PR would make more sense? The required changes seem simple enough, but would probably draw out the review process even more. Plus it woud be nice to have a single PR/commit for reference when the feature is rolled back at a later point.

charludo:
The test finally pass... 😭 🎉 

This is finally ready for review :)

steffenkleinle:
Hey, thank you so much for this PR! :green_heart: Is there already a timeline for this? We'd need the API to complete work in the frontend as well with sadly the same deadline :see_no_evil:
Would it be possible to do a release with this to the test cms already?

steffenkleinle (commented):
Did we decide on a way to notify the CMS that the user wants to talk to a human counselor? I seem to remember that we talked about a specific "message", right? Probably also @svenseeberg. I am not sure if this works well in the multilingual setting. Perhaps we should add an additional attribute `request_counselor` or similar to the POST body.

Since BOT answers are not part of the initial MVP, this doesn't have to be done here though.

steffenkleinle:
Regarding the device_id:
We decided to generate a UUID in the frontend (e.g. using [randomUUID](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID)) and save it to the local storage for that purpose, right?
In theory, what happens if a user changes city/language and tries to open a new chat? Then previous messages are "lost" and a new chat is started, right?

deen13 (commented):
Thank you for the effort and hard work in integrating the Chat API into our backend system. 🥳 This feature will undoubtedly enhance our app's communication capabilities. However, there are several architectural and technical considerations that I believe need to be addressed. **I want to note that I wasn't involved in the preliminary discussions that led to this architectural decision, so please take my feedback with that context in mind.** ☺️ 

#### Architectural Considerations

1. **Architecture Decision Record (ADR):**
   To provide clarity and context behind this architectural choice, I suggest documenting the decision using an Architecture Decision Record (ADR). This will help other developers understand the reasoning, the stakeholders involved, and the key decision points. I recommend using [ADR tools](https://github.com/npryce/adr-tools) to facilitate this process.

2. **Dependence on Zammad Without Abstraction:**
   I'm concerned about directly basing our Chat API on Zammad without an abstraction layer. Here are some points to consider:
   - **Future Migration:** Can we migrate away from Zammad in the future if needed? It might be challenging without an abstraction layer.
   - **Non-Zammad Users:** What about communes who do not use Zammad but still want to use the chat feature?
   - **Single Responsibility Principle:** Using tickets as containers for chats might violate the Single Responsibility Principle.

#### Open Questions and Thoughts

- Can our app detect if a Zammad backend is configured? This might be necessary in order to hide the chat feature in case there is no backend configured.
- How will we handle live updates and push notifications for chat messages? This is crucial for real-time communication.
- Before storing the Zammad Access Token, we should validate it with a test case to ensure its correctness. Failing fast in this case also has the benefit of allowing us to show the user an error message if, for example, they made a typo.
- We need to ensure that messages are not lost if the Zammad backend is temporarily unavailable (e.g., during an update). Implementing a retry mechanism might be necessary.
- What happens if the distributed state becomes unsynchronized, and a chat_id in our database no longer exists in the Zammad backend?
- If the Zammad Access Token becomes invalid over time (e.g., due to deletion), how will our backend handle it? We should display an alert or send a notification in such cases.
- What is the device id, and how can a user restore their chat if they change their device? Are we certain that the device id cannot be taken over by someone who recycles the device?

svenseeberg:
> Did we decide on a way to notify the CMS that the user wants to talk to a human counselor? I seem to remember that we talked about a specific "message", right? Probably also @svenseeberg. I am not sure if this works well in the multilingual setting. Perhaps we should add an additional attribute `request_counselor` or similar to the POST body.

I think we decided to implement this on the Zammad side with triggers. No need to have anything on the CMS?!

svenseeberg:
> * Can our app detect if a Zammad backend is configured? This might be necessary in order to hide the chat feature in case there is no backend configured.

That will be done manually anyways.
 
> * How will we handle live updates and push notifications for chat messages? This is crucial for real-time communication.

We actively decided to not implement push in the first iteration. We will work with polling on the app side.

> * Before storing the Zammad Access Token, we should validate it with a test case to ensure its correctness. Failing fast in this case also has the benefit of allowing us to show the user an error message if, for example, they made a typo.

While this would be nice, this will usually be done by admins/experts. They would need to set up Zammad anyways.

> * We need to ensure that messages are not lost if the Zammad backend is temporarily unavailable (e.g., during an update). Implementing a retry mechanism might be necessary.

Yes, this is currently an issue. But for a prototype this could be sufficient. I have no strong opinion here.

> * What happens if the distributed state becomes unsynchronized, and a chat_id in our database no longer exists in the Zammad backend?
>
> * If the Zammad Access Token becomes invalid over time (e.g., due to deletion), how will our backend handle it? We should display an alert or send a notification in such cases.
> 
> * What is the device id, and how can a user restore their chat if they change their device? Are we certain that the device id cannot be taken over by someone who recycles the device?

Valid points. But I'd suggest to move that into new issues and address them over time.

deen13:
> While this would be nice, this will usually be done by admins/experts. They would need to set up Zammad anyways.

In my humble opinion admins and experts can benefit from input validation just as any other user.

deen13:
> > * Before storing the Zammad Access Token, we should validate it with a test case to ensure its correctness. Failing fast in this case also has the benefit of allowing us to show the user an error message if, for example, they made a typo.
> 
> While this would be nice, this will usually be done by admins/experts. They would need to set up Zammad anyways.

In my humble opinion admins and experts can benefit from input validation just as any other user.

svenseeberg:
> In my humble opinion admins and experts can benefit from input validation just as any other user.

Sure. Its just a matter of priorties. We only have one test region for now and currently I can live w/o input validation. I know that we're potentially building some kind of tech debt. But this is just a prototype functionality. We do not even know yet if we're going to use it productively.

> Why should we address them gradually instead of immediately? Especially considering that points related to data structure and architecture, such as the first one, will become increasingly challenging to tackle once we've shipped the first version.

Because we need some kind of usable prototype soonish (end of June). We are a very long way from using it in wide spread production. And I would rather choose to get some first feedback from some users than producing an elaborate architecture without knowing whether we need it.

To address your concerns: I definitely agree that we should not postpone improving things. I would just rather get this merged, improve with a new PR and in parallel collect first feedback.

charludo:
> Hey, thank you so much for this PR! 💚 Is there already a timeline for this? We'd need the API to complete work in the frontend as well with sadly the same deadline 🙈 Would it be possible to do a release with this to the test cms already?

Tricky. Is it possible to complete most of the work without the concrete API endpoints, according to the API docs added in this PR?

> Did we decide on a way to notify the CMS that the user wants to talk to a human counselor? I seem to remember that we talked about a specific "message", right? Probably also @svenseeberg. I am not sure if this works well in the multilingual setting. Perhaps we should add an additional attribute `request_counselor` or similar to the POST body.
> 
> Since BOT answers are not part of the initial MVP, this doesn't have to be done here though.

Not implemented yet, but as you said, it's also not relevant for the MVP. I'll open an issue.

> Regarding the `device_id`: We decided to generate a UUID in the frontend (e.g. using [randomUUID](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID)) and save it to the local storage for that purpose, right? In theory, what happens if a user changes city/language and tries to open a new chat? Then previous messages are "lost" and a new chat is started, even though the `device_id` stays the same?

Changing the region would have to start a new chat, yes, since the Zammad backend used is dependant on the region. The best way to handle this scenario (wipe chat? inject error message informing the user that the chat cannot be continued? offer to switch back to the previous region?) is probably a question for UX. "As is", if the user has a chat history with region A, then switches to region B, then
- if B has the chat backend configured, a new chat is initialized with B when the user sends a message, and an error ("The requested chat does not exist. Did you delete it?") is shown when trying to get the messages
- if B does not have the backend configured, an error ("No chat server is configured for your region.") is shown.

Changing the language does not have any effect. The language the user has the app set to when initializing the chat is included as part of the ticket subject, but not checked afterwards.

charludo:
> 1. **Architecture Decision Record (ADR):**

I am not at all opposed to this, but I do think we should bring it up in the next team call and decide how we want to employ this. I think adding new tooling and processes to this specific (already huge...) PR is not a good idea.

> 2. **Dependence on Zammad Without Abstraction:**

IMO this is a non-issue. Right now, the feature is explicitly wanted in the context of Zammad, and we don't have evidence that other regions want to use a different backend. Until (if) that changes, another layer of abstraction does not provide any benefit.

(If we do want to add it at a later point though, all the pieces are already in place: we'd just need to add an abstract `ChatAPIProvider` class from which `ZammadChatAPI` and any other providers inherit, then change the line `client = ZammadChatAPI(request.region)` in `api/.../user_chat.py` to use the region's configured provider.)

> Can our app detect if a Zammad backend is configured? This might be necessary in order to hide the chat feature in case there is no backend configured.

In addition to what Sven said, `{"status": 503, "error": "No chat server is configured for your region."}` is returned if Zammad is not configured. Should be sufficient IMO.

> How will we handle live updates and push notifications for chat messages? This is crucial for real-time communication.

(see Sven's answer)

> Before storing the Zammad Access Token, we should validate it with a test case to ensure its correctness. Failing fast in this case also has the benefit of allowing us to show the user an error message if, for example, they made a typo.

Validate how exactly? We could probably check the token length, but other than that? I don't think checking Zammad connectivity on save of the region form is a good idea. We could maybe add a "test connection" button, but I'd rather move that to a new issue/PR

> We need to ensure that messages are not lost if the Zammad backend is temporarily unavailable (e.g., during an update). Implementing a retry mechanism might be necessary.

`{ "status": 500, "error": "An error occurred while attempting to connect to the chat server." }` is returned in that case. Since we do definitely not want to store any actual messages, the client needs to handle this case (for example by something like "tap to retry sending")

> What happens if the distributed state becomes unsynchronized, and a chat_id in our database no longer exists in the Zammad backend?
`{ "status": 404, "error": "Couldn't find Ticket with 'id'=<the id>" }` is returned by the Zammad server. The client should probably just create a new chat in this case.

> If the Zammad Access Token becomes invalid over time (e.g., due to deletion), how will our backend handle it? We should display an alert or send a notification in such cases.

`{ "status": 500, "error": "An error occurred while attempting to connect to the chat server." }`

Besides that, I do not think it's our responsibility to monitor region's Zammad instances. I guess the "test connection" functionality mentioned above could be run periodically in theory, though. But yes, I agree with Sven, new issue for this.

> What is the device id, and how can a user restore their chat if they change their device? 

The `device_id` is a random string generated by the client (app). Chats cannot be restored on new devices. Currently, AFAIK, this is by design; if we did want to be able to transfer chats to new devices, that would need to be handled 100% on client side.

> Are we certain that the device id cannot be taken over by someone who recycles the device?

Yes, it's probably worth mentioning that `device_id` is not actually tied to the physical device, but just a random ID which we use to associate clients with Zammad tickets.

deen13 (approved):
Thanks for the answers @svenseeberg and @charludo. I don't want to block the pull request further. ☺️

deen13:
@charludo Sorry, I didn't want to add another new tool to the project. Maybe there's already a place in the project suited for some architecture documentation? 👀

charludo:
> Thanks for the answers @svenseeberg and @charludo. I don't want to block the pull request further. ☺️

No worries. Having to merge this fast is a not-too-great situation to be in. Btw though, if you have implementation-specific concerns or suggestions I am all ears! 😅 



> @charludo Sorry, I didn't want to add another new tool to the project. Maybe there's already a place in the project suited for some architecture documentation? 👀

Not sure if our docs would be the right place for this, this seems a bit too fine-grained for them. I did read up on ADRs, but am a bit lost about what exactly there's to document here. AFAIK the way API requests are handled for the chat are in line with all other API request handling in this project.

steffenkleinle:
> Deadline was pushed forward to Friday (not for the release but for merging to the test CMS), as the Frontend needs at least two weeks to test :) If we don't have a second review until Friday we should just merge it with only one approve :)

Thanks! Its maybe important to note that this is not a final product but instead just an MVP for some early testing. Bugs and non-perfect behavior is okay! Therefore, if you have any issues you don't like about the current implementation, feel free to continue to work on this while we test and implement the frontend side.

Thank you all for your help :heart:

svenseeberg (approved):
LGTM, thanks :rocket: 

One thing I would change in the future is not showing internal comments as official responses:

image

But for testing purposes this is totally sufficient.

charludo:
> LGTM, thanks 🚀
> 
> One thing I would change in the future is not showing internal comments as official responses. The red bordered message is also returned in the API, while Zammad defines this as "internal" information.
> 
> image
> 
> But for testing purposes this is totally sufficient.
> 
> *edit: I just opened a follow up PR for this: #2844

Good catch!

And also, great, I'll solve the conflict and then merge... Let's see what happens 😅

steffenkleinle:
Thanks to all of you! Amazing :heart: 
Could you please notify us once this is ready to test (in the test cms)?

Issue #2721, opened 2024-04-02
`&quot;` instead of `"` after MT by Google Translate in some languages


MizukiTemma:
### Describe the Bug

In some languages HTML Entities like`&quot;` are not well unescaped.

### Steps to Reproduce

1. Use `"` in a page title or content (for example "Mein Kind kommt in die Schule")
2. Translate it into Arabic by Google Translate
3. See in the translation `&quot;طفلي يذهب إلى المدرسة&quot;`

### Expected Behavior

`"` (or another quotation which is appropriate for the target language)

### Actual Behavior

`&quot;` in the translation

### Additional Information

The problem was found in [in this page](https://cms-test.integreat-app.de/kreisviersen/pages/ar/12694/edit/) in Arabic, Farsi, Kurmanji, Tingrinya and Sorani.

<details>

  <summary>Traceback</summary>

  ```
  ```

</details>

PR #2740, opened 2024-04-09
Fix HTML Entities after Google Translate

MizukiTemma:
### Short description

This PR fixes the problem that `"` is saved as `&quot;` after Google Translation

### Proposed changes

- `unescape` again before saving

### Side effects

- I hope none 🙈 did some check with other symboles like `€` or `<` and HTML tags looked uninfluenced, but more eyes will help a lot 👁️ 

### Resolved issues

Fixes: #2721


david-venhoff (changes requested):
Thanks! I tested the pr and the title is indeed correct after translation with those changes. However, I think the solution might have to be a bit more complicated, because this can break the translation of the content after a translation in a minor way:
If i enter the following text as a page content:
```
Hello, "Quotes" &quot;
```
Then with this pr the German translation becomes
```
Hallo, "Zitate" "
```

While I think it should be
```
Hallo, "Zitate" &quot;
```

I think the most robust way to fix this would be to only pass `html` as format when we translate the content and use `text` for non-html fields. That way the exact text that was entered should get preserved.

For the basic translation api, the following code seems to fix all issues. Here I removed both unescape calls.
```python
                            format_ = "html" if attr == "content" else "text"
                            data[attr] = (
                                self.translator_v2.translate(
                                    values=[
                                        getattr(source_translation, attr)
                                    ],
                                    target_language=target_language_key,
                                    source_language=source_language.slug,
                                    format_=format_,
                                )[0]["translatedText"])
```
Although we might want to not hardcode which fields are html and which not, but that is probably not too important.

MizukiTemma:
@david-venhoff Nice catch and suggestion 😄 applied!

timobrembeck (approved):
Thanks! :+1:

david-venhoff (approved):
:+1:

Issue #2442, opened 2023-10-09
Replace pylint + isort by ruff


timobrembeck:
### Motivation

Our CI workflows take quite a while, mostly due to the tests, but also pylint is taking a considerable amount of time.
I think we could reduce this by migrating from pylint to ruff:

![232603514-c95e9b0f-6b31-43de-9a80-9e844173fd6a](https://github.com/digitalfabrik/integreat-cms/assets/16021269/8cdc2203-9e6a-4cac-ad84-bbab3a757461)

### Proposed Solution

- Replace pylint and isort by ruff
- Check whether there's an automated tool which can migrate the `pylint: disable=` statements

### Alternatives

Maybe it isn't worth the effort...

### Additional Context

Pylint is not completely implemented by ruff yet (see https://github.com/astral-sh/ruff/issues/970), so we would loose a few rules (but also ruff has a lot of extra checks that pylint doesn't, mostly those of flake8)...

JoeyStk:
Not yet fully resolved. The following steps are still needed:

- [ ] enable ruff format
- [ ] enabled option to sort imports via `extend-select = ["I"] `

deen13:
We also need to drop the pylint leftovers to close this ticket.

JoeyStk:
@charludo this is a duplicate / done, right? :) If you agree, you can just close this issue :)

PR #2744, opened 2024-04-15
Add dev tool for ruff

JoeyStk:
### Short description

This PR adds a dev tool for ruff

### Proposed changes

- Add dev tool for ruff

### Side effects

- At the moment we use ruff still in parallel to the other formatters and linters. AFAIK over time we want to move towards using only one?
- Yet only check is enabled also with fix as extra parameter. Maybe over time we want to include formatting and also include the option to pass arguments such as --fix manually

### Resolved issues

Fixes: One part of #2442

### DIFF
--- a/tools/ruff.sh
+++ b/tools/ruff.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+# This script can be used to format the python code with ruff.
+
+# For the transitional phase we use both pylint and ruff.
+
+# Import utility functions
+# shellcheck source=./tools/_functions.sh
+source "$(dirname "${BASH_SOURCE[0]}")/_functions.sh"
+
+require_installed
+
+# Run ruff
+echo "Starting code linting and formatting with ruff..." | print_info
+ruff check --fix "${BASE_DIR}"
+# ruff format "${BASE_DIR}"
+echo "✔ Code formatting finished" | print_success


PeterNerlich (approved):
LGTM

charludo (approved):
🎉

timobrembeck:
Hmm, as described in #7242, I think adding more dev tools to perform tasks that already can be done by one pre-commit command is kind of superfluous...

JoeyStk:
Okay. Do you suggest to close this then? :)

charludo:
> Hmm, as described in #7242, I think adding more dev tools to perform tasks that already can be done by one pre-commit command is kind of superfluous...

Invoking the pre-commit manually in order to check for things you need to fix is a bit burdensome though, + the same could be said for most existing dev tools

timobrembeck:
> Okay. Do you suggest to close this then? :) 

Yes, I personally would suggest that, but I'm open for other opinions. 

> Invoking the pre-commit manually in order to check for things you need to fix is a bit burdensome though, + the same could be said for most existing dev tools

Hmm, I'd say that 
```
pre-commit run ruff
```
is not much longer than:
```
./tools/ruff.sh
```

If you still think it's inconvenient, we could set bash aliases for this command rather than separate bash scripts?

PeterNerlich:
My two cents: I basically ignored the pre-commit hooks entirely, always running the code style tools individually or as `code_style.sh`, because it took way too long and got in the way especially during rebases, if I remember correctly. So I at least got very accustomed to `tools/`. But my opinion on the pre-commit hooks might very well change once rust fully replaces most tools.

charludo:
> Hmm, I'd say that
> 
> ```
> pre-commit run ruff
> ```
> 
> is not much longer than:
> 
> ```
> ./tools/ruff.sh
> ```

True, but then why not do the same for eslint, prettier, black, isort,...?

I'd wager most of us are used to using the dev tools for these tasks, and I am having a little trouble seeing the benefit in changing this workflow.

> My two cents: I basically ignored the pre-commit hooks entirely, always running the code style tools individually or as `code_style.sh`, because it took way too long and got in the way especially during rebases, if I remember correctly. So I at least got very accustomed to `tools/`.

👍🏼

PeterNerlich:
I think we can just merge this, `tools/` became kind of the default location I look for, well, tools like that. If it's not in a script, I just straight up assume it's not set up for the project.

If we organize this differently in the future, I guess it makes little difference whether to also have to delete this `ruff.sh` or not.

Issue #2634, opened 2024-02-05
Media items used in past event should be deletable


osmers:
### Motivation
Sometimes media items are used only once for an event - e.g. pictures of flyers 
They should be deletable, if the usage in the media library refers only to an events page.

### Proposed Solution
Allow media items which are only used in /events/ to be deleted

### Alternatives
Make users delete the event first and then the media icon - more work, less user friendly

### User Story
As a CMS user I want to be able to clean up my media library without much work in order to have a better overview/less clutter.

### Additional Context
Check this icon for example: https://admin.integreat-app.de/marburg-biedenkopf/media-library/#/50/

PR #2745, opened 2024-04-15
Deletable media files that are attached to past events only

deen13:
### Short description

This pull request changes the `deletable` flag of the [media_file.py](https://github.com/digitalfabrik/integreat-cms/blob/7f9085f9b8eb13d6639aaafd1efaad1625868c77/integreat_cms/cms/models/media/media_file.py#L2) in a such a way that it is true when only used by events in the past.

### Resolved issues
- [Media items used in past event should be deletable](https://github.com/digitalfabrik/integreat-cms/issues/2634)

Fixes:  #2634

### DIFF
--- a/integreat_cms/cms/models/media/media_file.py
+++ b/integreat_cms/cms/models/media/media_file.py
@@ -277,7 +277,7 @@ class MediaFile(AbstractBaseModel):
         """
         usage_counts = [
             self.past_event_usages.count(),
-            len(self.icon_usages),
+            (len(self.icon_usages) + self.content_usages.count()),
             self.events.count(),
         ]
 
--- a/integreat_cms/cms/models/media/media_file.py
+++ b/integreat_cms/cms/models/media/media_file.py
@@ -260,6 +260,39 @@ class MediaFile(AbstractBaseModel):
         return Link.objects.filter(url__url=self.url).distinct("object_id")
 
     @cached_property
+    def past_event_usages(self) -> QuerySet:
+        """
+        Count in how many past events this file is used
+
+        :return: count of usages in past events
+        """
+        return self.events.filter(end__lt=timezone.now().date())
+
+    @cached_property
+    def is_only_used_in_past_events(self) -> QuerySet:
+        """
+        Check if a media file is used in past events only
+
+        :return: if a media file is only used in past events
+        """
+        usage_counts = [
+            self.past_event_usages.count(),
+            len(self.icon_usages),
+            self.events.count(),
+        ]
+
+        return all(c == usage_counts[0] for c in usage_counts)
+
+    @cached_property
+    def is_deletable(self) -> bool:
+        """
+        Check if a media file deletable
+
+        :return: Whether a file is deletable
+        """
+        return not self.is_used or self.is_only_used_in_past_events
+
+    @cached_property
     def is_embedded(self) -> bool:
         """
         Check if a media file is embedded in the content
@@ -337,7 +370,7 @@ class MediaFile(AbstractBaseModel):
             "lastModified": localize(timezone.localtime(self.last_modified)),
             "isGlobal": not self.region,
             "isHidden": self.is_hidden,
-            "deletable": not self.is_used,
+            "deletable": self.is_deletable,
         }
 
     @classmethod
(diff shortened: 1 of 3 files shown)


JoeyStk (approved):
Thank you!
Love the tests!

MizukiTemma (changes requested):
Thank you for qucik PR 👍 

I just found one case where it is not working as expected: media files that are used only in an non-event translation as **content** can be deleted, although they should be protected from deletion.


1. Upload a picture into the media library
2. Insert it into the "content" of a page (not as icon) and save
3. Go back to the media library and open the side menu for the picture
4. See it's deletable

I guess the cause is that ´content_usages´ is not taken into account.

deen13:
Hey @MizukiTemma, great catch, and thank you for bringing it up! I've addressed that edge case and also included an additional test. Unfortunately, I haven't been able to write a test yet that generates a page with content utilizing a media file.

integreat_cms/cms/models/media/media_file.py
```diff
@@ -259,6 +259,39 @@ def content_usages(self) -> QuerySet:
...
+
+        :return: if a media file is only used in past events
+        """
+        usage_counts = [
+            self.past_event_usages.count(),
+            (len(self.icon_usages) + self.content_usages.count()),
+            self.events.count(),
+        ]
+
+        return all(c == usage_counts[0] for c in usage_counts)
```
MizukiTemma, line 284: Sorry, my suggestion introduced another bug/edge case 🙈 
`past_event_usages` finds only events in which the media file is used as icon and does not find use case as content 🙈 

So, I'd suggest to adjust `is_only_used_in_past_events` as below

```suggestion
        past_event_translation_ids = [
            event_translation.id
            for event_translation in EventTranslation.objects.filter(
                event__end__lt=timezone.now().date()
            )
        ]

        return (
            len(self.icon_usages) == self.past_event_usages.count()
            and not self.content_usages.exclude(
                content_type__id=ContentType.objects.get(model="eventtranslation").id,
                object_id__in=past_event_translation_ids,
            ).count()
            > 0
        )
```

if this is applied, `from ..events.event_translation import EventTranslation` and `from django.contrib.contenttypes.models import ContentType` need to be imported

MizukiTemma (changes requested):
🙈

tests/cms/models/media/test_media_file.py
```diff
@@ -0,0 +1,248 @@
...
+            region=region,
+        )
+
+        german_language = Language.objects.create(slug="de", bcp47_tag="de", native_name="Deutsch", english_name="German", text_direction="ltr", primary_country_code="DE", table_of_contents="Inhaltsverzeichnis")
+
+        EventTranslation.objects.create(event=past_event, language=german_language, content=f'<p><img src="http://localhost:8000/media/{file.file}"></p>')
+
+        find_all_links()
+
+        assert file.past_event_usages.count() == 1 and file.is_deletable
```
MizukiTemma, line 248:
```suggestion
        file.is_deletable
```
This fails because `past_event_usages` only detects usage as icon, and the media file is used as content in this test. We can omit this in my opinion.

tests/cms/models/media/test_media_file.py
```diff
@@ -0,0 +1,248 @@
...
+            is_hidden=False,
+        )
+
+        past_event = Event.objects.create(
+            start=timezone.now() - timedelta(days=2),
+            end=timezone.now() - timedelta(days=1),
+            region=region,
+        )
+
+        german_language = Language.objects.create(slug="de", bcp47_tag="de", native_name="Deutsch", english_name="German", text_direction="ltr", primary_country_code="DE", table_of_contents="Inhaltsverzeichnis")
```
MizukiTemma, line 242: I guess we are getting an error here because German exists already in test data.
We could either load the fixture and use the german language objects from it or create a brand new language with a slug which is nowhere used.

MizukiTemma (changes requested):

tests/cms/models/media/test_media_file.py
deen13: Yes this was for me to verify that it counts correctly and is deletable but I now understand the behavior better and realized that it is absent because of the linkchecker.

MizukiTemma (approved):
Thank you!

david-venhoff:
We try to keep the commit history clean, so please remember to squash your commits next time :)

Issue #2583, opened 2023-12-05
Concept UI test


MizukiTemma:
### Motivation

In T32K in Novemver 2023 we have decided to introduce test for frontend. 

### Proposed Solution

TBD

### Additional Context

Related to #2584

MizukiTemma:
Blocked by #3777

PR #2716, opened 2024-03-28
Update webauthn to version 2.0

david-venhoff:
### Short description

That version contains some breaking changes, and this pr changes the affected code.
Since I haven't worked much with 2fa yet, it probably makes sense that someone with more experience reviews this.
This pr follows the upgrade guidelines from webauthn: https://github.com/duo-labs/py_webauthn/releases/tag/v2.0.0

### Proposed changes

- Add a new bytes field `webauthn_id` to the user model that is automatically generated by webauthn and use it instead of the numerical user id
- Update some renamed function calls

### Side effects

/

### Resolved issues

Fixes: #2614

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #2747, opened 2024-04-16
deps/upgrade webauthn test support

deen13:
### Short Description

This pull request configures vitest and adds a simple test as suggested in my review of [deps/upgrade_webauthn_2_0](https://github.com/digitalfabrik/integreat-cms/pull/2716).

### Proposed Changes

- Added a test task to `package.json`
- Included required vitest dependency
- Implemented a simple test for URL and base64 encoding
- Configure vitest to ignore python dependencies
- Configure import/resolver to resolve imports correctly
- Add job to circleci workflow

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -194,6 +194,17 @@ jobs:
       - run:
           name: Check formatting of CSS & JS files
           command: npx prettier --check .
+  vitest:
+    docker:
+      - image: "cimg/node:lts"
+    resource_class: small
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - run:
+          name: Run frontend tests
+          command: npm run test
   eslint:
     docker:
       - image: "cimg/node:lts"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -604,6 +604,9 @@ workflows:
       - prettier:
           requires:
             - npm-install
+      - vitest:
+          requires:
+            - npm-install
       - webpack:
           requires:
             - npm-install
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,7 +1,24 @@
+const path = require('path');
+
 module.exports = {
     env: {
         es6: true,
         browser: true,
+    },
+    settings: {
+        'import/resolver': {
+            node: {
+                paths: ['src'],
+                extensions: ['.js', '.jsx', '.ts', '.d.ts', '.tsx'],
+            },
+            typescript: {
+                project: './tsconfig.json',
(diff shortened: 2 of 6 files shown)


david-venhoff (commented):
Thanks :tada: 
I think it is a great step to enable js testing. It seems like this pr would close #2583, so it probably makes sense that @seluianova also takes a look at this.

When I tested this pr, I got an error that vitest exceeds the maximum number of file watchers on my system, because it seems to be watching every file in this repo (Including files of dependencies) for some reason. I managed to get rid of the error using this `vitest.config.ts` file:

```ts
import { defineConfig } from 'vitest/config';

export default defineConfig({
  test: {
    include: ['integreat_cms/static/src/js/__tests__/**'],
    watch: false,
  }
})
```

But I find it very weird that the `watch: false` option seems to be required to prevent this error, and that the glob is not enough to prevent it from watching every file.

Apart from that I think it makes sense to run npm test as part of our `./tools/test.sh`, or is there a reason against that?

deen13:
Just for the record: @david-venhoff and I successfully resolved the mentioned issue in [commit 09be3e5](https://github.com/digitalfabrik/integreat-cms/pull/2747/commits/09be3e560b860cd4b0f514edc151f69d9ab862e3).

deen13:
> Apart from that I think it makes sense to run npm test as part of our ./tools/test.sh, or is there a reason against that?

That's a good point! I took a look at ./tools/test.sh and noticed that it has quite a few options that it passes on to pytest like the option to invoke a single test (./tools/test.sh tests/cms/my_test.py -k my_test_case). I would prefer to keep this file as is because it is tailored to pytest, and more importantly, I don't want to break anyone's workflow. Do you think it's worth adding another script that runs all python and ts tests but accepts no flags?

deen13:
> It seems like this pr would close https://github.com/digitalfabrik/integreat-cms/issues/2583, so it probably makes sense that @seluianova also takes a look at this.

While the pull request concentrates on adding unit tests, the mentioned issue involves ui tests. Nonetheless, I'd be eager to contribute to ui testing if it's still necessary. Otherwise, feel free to disregard my comment.

seluianova:
> I believe the pull request only partially resolves the issue because it focuses solely on adding unit tests, whereas the mentioned issue pertains to UI tests

@david-venhoff right, in UI tests we wanted to add end-2-end kind of tests, that simulate user interaction. Like opening the browser, filling in fields, clicking on buttons, etc. At least that was my understanding of the task

david-venhoff:
Ah okay :+1:

PR #2748, opened 2024-04-16
Add mock_server fixture for tests

seluianova:
### Short description
Add mock_server fixture that can be re-used in tests.
This is a wrapper for pytest_httpserver, which can be extended with useful features.
For example, currently it allows to count incoming requests (to be re-used in https://github.com/digitalfabrik/integreat-cms/pull/2669)

### Proposed changes
- add mock_server fixture
- use mock_server fixture in deepl_api_test.py

### Side effects
- no?

### Resolved issues
Fixes: #

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,6 +10,7 @@ import pytest
 from django.contrib.auth import get_user_model
 from django.core.management import call_command
 from django.test.client import AsyncClient, Client
+from pytest_httpserver.httpserver import HTTPServer
 
 from integreat_cms.cms.constants.roles import (
     APP_TEAM,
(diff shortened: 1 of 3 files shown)


deen13 (commented):
Hey @seluianova,

That looks great! I have a question: Is it possible to extend the request counting so that one can query not only "how many requests have I received" but also "how many requests have I received on path /my/path that returned the status code 200"? I believe this enhancement would allow us to write assertions in a more explicit manner.

seluianova:
> Is it possible to extend the request counting so that one can query not only "how many requests have I received" but also "how many requests have I received on path /my/path that returned the status code 200"?

@deen13 yes, sure. do you suggest to implement it now or? :)

deen13:
@seluianova, if it aligns with your current priorities and workload, I believe it is worth implementing right away. However, the pull request seems to have a good scope and accomplishes the task, so there is no need to include it in this branch. ☺️

MizukiTemma (approved):
Cool improvement 😸

david-venhoff (approved):
:+1:

PR #2749, opened 2024-04-17
Ensure that webpack bundle exists before running tests

deen13:
### Description

This pull request introduces a precondition to the `test.sh` tool, ensuring that the webpack output is available. This precaution is necessary because otherwise, the tests would fail when someone runs `./tools/test.sh` in an environment where the application hasn't been initialized.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,8 @@
   "private": true,
   "scripts": {
     "dev": "NODE_ENV=development webpack watch --mode development --color --stats=errors-warnings",
-    "prod": "NODE_ENV=production webpack --mode production"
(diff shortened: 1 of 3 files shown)


PeterNerlich (approved):
Yes please!

Verified that starting out with tests after initial clone and `install.sh` creates a problem on develop, but not here.
My only concern is that we kind of have duplicate functionality with how webpack is run in `run.sh`, but that one is not only started in the background while other things still set up but also sticks around, watching for file changes to recompile. I'm not sure whether we can reasonably generalize and take advantage of the non-blocking idea here as well…?

MizukiTemma (approved):

Issue #2706, opened 2024-03-20
Use a consistent format for date and time


osmers:
### Motivation
Having a standardized date for last update makes a quick overview easier

### Proposed Solution
Don't show the time and write the date using only numbers - e.g. 04.03.2024 instead of 3. März 2024 - use German date system, not the American one

### Alternatives
Keep it, as it is

### User Story
As a user of the CMS I want to be able to see that last update date clearly without distracting information so that I can get a quick overview of outdated pages

![image](https://github.com/digitalfabrik/integreat-cms/assets/64901227/e205a146-55a1-4201-9e2e-82c8d755292a)

timobrembeck:
Hmm, I'd argue that time is actually relevant in many cases?
So maybe `20.02.2024 16:40` or some kind of separator between date and time, e.g. `|`, `–` or `, `?

lunars97:
from User's perspective maybe it could be cleaner and not so clattered with TMI if we could remove time for not so relevant things like pages except for versions like @osmers suggested

lunars97:
If we change `last update`, we also need to change `created` and `last login` as well in order to be consistent with one chosen date format

PR #2752, opened 2024-04-22
Use a consistent format for date and time

lunars97:
### Short description

Change current timestamp format for last update ( I realized that we also need to change all timestamps as well like `created` `last login`, `last usage`, and `date added`) to make a quick overview easier and not so cluttered with long date string. It might be better to remove time for not so relevant components like pages, languages, regions, feedback and include time to chat, user last login, events, and content versions. 

### Proposed changes

- Change timestamp format for short and remove time for not relevant components
- Change timestamp format for short and add time for relevant components, like chat, content versions, events and user's last login

### Side effects

None?

### Resolved issues

Fixes: #2706 

### DIFF
--- a/integreat_cms/core/formats/en/formats.py
+++ b/integreat_cms/core/formats/en/formats.py
@@ -10,7 +10,7 @@ from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     from typing import Final
 
-#: The default formatting to use for displaying datetime fields in any part of the system when using the German locale.
+#: The default formatting to use for displaying datetime fields in any part of the system when using the English locale.
 DATETIME_FORMAT: Final[str] = "d.m.Y, H:i"
 
 #: The short formatting to use for displaying datetime fields when the time is irrelevant
--- a/integreat_cms/release_notes/current/unreleased/2706.yml
+++ b/integreat_cms/release_notes/current/unreleased/2706.yml
@@ -1,0 +1,2 @@
+en: Use a consistent format for date and time
+de: Verwende ein einheitliches Format für Datums- und Zeitangaben
--- a/integreat_cms/cms/templates/feedback/admin_feedback_list_row.html
+++ b/integreat_cms/cms/templates/feedback/admin_feedback_list_row.html
@@ -78,6 +78,6 @@
         {% endif %}
     </td>
     <td class="pr-2 text-right whitespace-nowrap">
-        {{ feedback.created_date }}
+        {{ feedback.created_date|date:"SHORT_DATE_FORMAT" }}
     </td>
 </tr>
--- a/integreat_cms/cms/templates/feedback/region_feedback_list_row.html
+++ b/integreat_cms/cms/templates/feedback/region_feedback_list_row.html
@@ -72,6 +72,6 @@
         {% endif %}
     </td>
     <td class="pr-2 text-right whitespace-nowrap">
-        {{ feedback.created_date }}
+        {{ feedback.created_date|date:"SHORT_DATE_FORMAT" }}
     </td>
 </tr>
--- a/integreat_cms/cms/templates/generic_auto_save_note.html
+++ b/integreat_cms/cms/templates/generic_auto_save_note.html
@@ -2,5 +2,5 @@
 <span id="auto-save"
       class="{% if form_instance.status != AUTO_SAVE %} hidden {% endif %}">
     <i icon-name="check"></i>
-    {% trans "Automatically saved on:" %} <span id="auto-save-time">{{ form_instance.last_updated }}</span>
+    {% trans "Automatically saved on:" %} <span id="auto-save-time">{{ form_instance.last_updated|date:"SHORT_DATE_FORMAT" }}</span>
 </span>
--- a/integreat_cms/cms/templates/languages/language_list_row.html
+++ b/integreat_cms/cms/templates/languages/language_list_row.html
@@ -25,10 +25,10 @@
         {{ language.get_text_direction_display }}
     </td>
     <td class="pr-2">
-        {{ language.created_date }}
+        {{ language.created_date|date:"SHORT_DATE_FORMAT" }}
     </td>
     <td class="pr-2">
-        {{ language.last_updated }}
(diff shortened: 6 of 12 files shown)


integreat_cms/static/src/js/media-management/component/edit-directory-sidebar.tsx
```diff
@@ -51,6 +51,31 @@ const EditDirectorySidebar = ({
...
+        };
+
+        const [monthStr, day, yearStr] = dateString.split(" ");
+        const [year] = yearStr.split(",");
+        const month = months[monthStr];
+        const formattedDay = day.replace(",", "");
+
+        return `${formattedDay}.${month}.${year}`;
+    };
+    const formattedDate = formatDate(directory.CreatedDate.toLocaleString());
```
deen13, line 77: I suggest using [date-fns](https://date-fns.org/) for all things related to dates as it is a proven utility library. It's less likely that we'll run into bugs compared to building it ourselves. :)

integreat_cms/release_notes/current/unreleased/2706.yml
```diff
@@ -0,0 +1,2 @@
+en: Add standardized timestamp for all dates like last update, created at, last usage, last login, date added
+de: Füge einen standardisierten Zeitstempel für alle Daten, wie letzte Aktualisierung, erstellt am, letzte Nutzung, letzte Anmeldung, Datum des Hinzufügens hinzu
```
deen13, line 2: I suggest using only nouns, as this contributes to a more formal and precise expression.

```suggestion
de: Füge einen standardisierten Zeitstempel für alle Daten, wie letzte Aktualisierung, Erstellungsdatum, letzte Nutzung, letzte Anmeldung, Datum des Hinzufügens hinzu
```

deen13 (changes requested):
I have two minor comments but looks good overall 🥳

timobrembeck (commented):
Thanks a lot! :+1:

Instead of the numerous changes, I think it would make sense to change the default representation here:

https://github.com/digitalfabrik/integreat-cms/blob/16e925de07d412d89b56e8ee76b258c9a4a236b1/integreat_cms/core/formats/de/formats.py#L14

to e.g.
```
DATETIME_FORMAT: Final[str] = "d.m.Y, H:i"
```

and also add a new setting
```
SHORT_DATETIME_FORMAT: Final[str] = "d.m.Y"
```
for all places where the time is not relevant. Then, we can use this format via e.g.
```
{{ feedback.created_date|date:"SHORT_DATE_FORMAT" }}
```
without having to duplicate the changes in case we want to change the format in the future again.

Also, I don't think it makes sense to use the a.m./p.m. for German users – the 24h format is way more common here, and it's also shorter!

lunars97:
> Thanks a lot! 👍
> 
> Instead of the numerous changes, I think it would make sense to change the default representation here:
> 
> https://github.com/digitalfabrik/integreat-cms/blob/16e925de07d412d89b56e8ee76b258c9a4a236b1/integreat_cms/core/formats/de/formats.py#L14
> 
> to e.g.
> 
> ```
> DATETIME_FORMAT: Final[str] = "d.m.Y, H:i"
> ```
> 
> and also add a new setting
> 
> ```
> SHORT_DATETIME_FORMAT: Final[str] = "d.m.Y"
> ```
> 
> for all places where the time is not relevant. Then, we can use this format via e.g.
> 
> ```
> {{ feedback.created_date|date:"SHORT_DATE_FORMAT" }}
> ```
> 
> without having to duplicate the changes in case we want to change the format in the future again.
> 
> Also, I don't think it makes sense to use the a.m./p.m. for German users – the 24h format is way more common here, and it's also shorter!

Thank you! Good catch :)

integreat_cms/static/src/js/media-management/component/edit-directory-sidebar.tsx
lunars97: @deen13 thank you for your review :) I think now there is no need in using library. It works without any additional formating

integreat_cms/core/formats/en/formats.py
```diff
@@ -0,0 +1,15 @@
...
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Final
+
+#: The default formatting to use for displaying datetime fields in any part of the system when using the English locale.
+DATETIME_FORMAT: Final[str] = "d.m.Y, H:i"
+SHORT_DATE_FORMAT: Final[str] = "d.m.Y"
```
lunars97, line 15: It was necessary to add a format for English default UI language to overwrite the long string format in media library

integreat_cms/cms/templates/chat/_chat_messages.html
```diff
@@ -22,7 +22,7 @@
             <i icon-name="trash-2"></i>
         </button>
         {# djlint:on #}
-        <span class="text-gray-600 float-right">{{ message.sent_datetime }}</span>
+        <span class="text-gray-600 float-right">{{ message.sent_datetime|date:"DATETIME_FORMAT" }}</span>
```
timobrembeck, line 25:
```suggestion
        <span class="text-gray-600 float-right">{{ message.sent_datetime }}</span>
```
This is superfluous. You only need to specify the format if it differs from the default format. But since `DATETIME_FORMAT` is the default, you don't need this here. You only need it in places where `SHORT_DATE_FORMAT` should be used instead of the default format.

integreat_cms/cms/templates/content_versions.html
```diff
@@ -53,7 +53,7 @@ <h1 class="heading">
         {% for translation in translations %}
             <div class="w-full hidden revision-wrapper"
                  id="revision-{{ translation.version }}"
-                 data-date="{{ translation.last_updated }}"
+                 data-date="{{ translation.last_updated|date:"DATETIME_FORMAT" }}"
```
timobrembeck, line 56:
```suggestion
                 data-date="{{ translation.last_updated }}"
```

deen13 (approved):

timobrembeck (commented):
Thanks! :+1:
Could  you squash your commits?

timobrembeck (approved):

PR #2758, opened 2024-04-23
Configure ruff import sorting and drop isort support

deen13:
### Short description

This pull request enables import sorting for ruff and tick some of the remaining boxes in #2442.

### Side effects
While Ruff claims to be nearly equivalent to Isort when profile = "black", it formats the imports slightly differently. This pull request removes support for isort and deprecates black in our documentation to address the conflicting formatters.

Further Information: https://docs.astral.sh/ruff/faq/#how-does-ruffs-import-sorting-compare-to-isort


PeterNerlich:
I don't think tracking `.vscode/` is a good idea. Could we do something like putting it into `example-configs/` but have it still feel intuitive *(so you don't set up your whole project and notice at the very end that you could have based it on the default config)*?

charludo:
Hm, I'm with @PeterNerlich on this. Not sure we should be adding editor-specific configs. Wouldn't most people have set this up globally in their editor already anyways (and if not, probably prefer it that way)?

PeterNerlich:
My concern was mostly about some user changing settings and then accidentally committing them to the project. This wouldn't happen if the config in the repo was under some other name *(like `vscode.example/`)* that you'd copy to `.vscode` in order to use and maybe customize further – any changes to the example would have to be deliberate by design *(except if you use symlinks, but then that's your own fault)*

deen13:
I think it's very unlikely for someone to change the file by accident because changes made in Visual Studio Code are not written to the settings.json. I agree with you that it feels weird to commit dotfiles, but in this case, it adds value to the developer experience and onboarding without any real negative consequences. Let me know if I've overlooked any cases where having the file would be problematic. 🙈

deen13:
I have an idea for a tradeoff. How about removing it from here and adding it to the devcontainer configuration, as it already contains some rough-related stuff? @PeterNerlich @charludo

charludo:
> I have an idea for a tradeoff. How about removing it from here and adding it to the devcontainer configuration, as it already contains some rough-related stuff? @PeterNerlich @charludo

Sounds like a good idea to me!!

docs/src/tools.rst
```diff
@@ -124,13 +124,13 @@ with :github-source:`tools/integreat-cms-cli`::
 Code Quality
 ============
 
-Automatically sort python import statements with :github-source:`tools/isort.sh`::
+(Deprecated) Automatically apply our python style with :github-source:`tools/black.sh`::
```
charludo, line 127:
```suggestion
Automatically apply our python style with :github-source:`tools/black.sh`::
```

I'd argue it is not deprecated yet, since using `ruff format` leads to a very lage diff right now.

charludo (changes requested):
LGTM, apart from one tiny thing

docs/src/tools.rst
cclauss: You might want to consider `ruff format` as a highly compatible but much faster replacement for `black`.
* https://docs.astral.sh/ruff/formatter/#black-compatibility

cclauss:
Are you recommending the use of __Ruff extension for Visual Studio Code__?
* https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff

docs/src/tools.rst
charludo: The diff is largely due to currently wrapped lines (black's default line length is 88) being put on a single line (pyproject.toml configures line lengths of 249 for ruff, apparently due to differences in the way comment strings are habndled, at least so I've been told)

cclauss: https://docs.astral.sh/ruff/formatter/black/#trailing-end-of-line-comments

charludo: From that article:

> We agree that Ruff's formatting (that matches Black's 23) is hard to read and needs improvement. But we aren't convinced that parenthesizing long nested expressions is the best solution, especially when considering expression formatting holistically. That's why we want to defer the decision until we've explored alternative nested expression formatting styles.

However, setting 
```toml
[tool.ruff]
line-length = 88
```

indeed does introduce only a handful of changes - @timobrembeck, did I misunderstand the reason for setting it to `279`? Because just from this it looks like we could indeed shorten this.

deen13:
@cclauss yep it is predefined in our setup for code users

docs/src/tools.rst
timobrembeck: I guess this is the corresponding issue on ruff's side: https://github.com/astral-sh/ruff/issues/9926

cclauss: Use `ruff format` instead of `black`?  Highly compatible and much faster.

charludo: +1 for ruff format @timobrembeck that throws 0 errors and only reformats 15 files

cclauss: Awesome...  Use `ruff check --select=PL` instead of `pylint`?

deen13: @cclauss, this seems like a very good idea to me in general, but Ruff does not yet support all Pylint rules. Therefore, I'd prefer to wait a bit and ensure a smooth migration later on.

deen13: @charludo ruff formatting does work on my local setup as well, but it fails within the pre-commit hook and on circleci as well. I guess we have to wait with the formatter migration until the bug in the rule is fixed.

docs/src/tools.rst
```diff
@@ -124,11 +124,11 @@ with :github-source:`tools/integreat-cms-cli`::
 Code Quality
 ============
 
-(Deprecated) Automatically apply our python style with :github-source:`tools/black.sh`::
+Automatically apply our python style with :github-source:`tools/black.sh`::
```
cclauss, line 127:
```python
Automatically apply our Python style with :github-source:`tools/black.sh`::
```

docs/src/tools.rst
```diff
@@ -124,11 +124,11 @@ with :github-source:`tools/integreat-cms-cli`::
...
 
     ./tools/black.sh
 
-Automatically apply our python style with :github-source:`tools/ruff.sh`::
+Automatically run python linting with :github-source:`tools/ruff.sh`::
```
cclauss, line 131:
```suggestion
Automatically run Python linting with :github-source:`tools/ruff.sh`::
```

charludo (approved):
👍🏼

david-venhoff (approved):
I haven't followed this pr closely, but the changes look good to me.
I also really like that the typing imports get formatted now :+1:

Issue #1853, opened 2022-11-10
Migrate firebase API from legacy to v1


timobrembeck:
### Motivation

At the moment, we use the [legacy fcm endpoint](https://firebase.google.com/docs/cloud-messaging/http-server-ref), but [v1](https://firebase.google.com/docs/reference/fcm/rest) was released quite a while ago and seems more future proof.

### Proposed Solution

- Follow [this guide](https://firebase.google.com/docs/cloud-messaging/migrate-v1) to migrate the api client

### Alternatives

Even though the API is called "legacy", I'm pretty sure it's still very commonly used and won't just suddenly disappear. I wouldn't be surprised if the legacy endpoint lives longer than our CMS.

sarahsporck:
I'd be open to do this, but I don't have the necessary permissions, to get the new 1) projectId and 2) the new authentication token.

timobrembeck:
maybe @svenseeberg can help?

dkehne:
@svenseeberg what is the status here?

svenseeberg:
The Firebase project is mainly managed by the app developers.

PR #2759, opened 2024-04-23
Update firebase cloud messaging to v1

david-venhoff:
### Short description

This pr updates the firebase push notification api from legacy to v1.

To test that this pr works you can follow these steps:
1. Get credentials from passbolt, you might need to request permission for that file 
2. Store the credentials.json file on your computer and set the `FCM_CREDENTIALS` accordingly
3. In `firebase_api_client.py` set `validate_only` to False, so that messages can be sent
4. Open Testumgebung in the mobile app (Hidden regions can be selected if you type `wirschaffendas` in the search box)
5. Run the server, navigate to Testumgebung and send a new push notification. Make sure that the push notification only gets sent to Testumgebung, otherwise users will see the test message!

I added the label `prio: high` because the api we currently use will be removed in june, and we need to have published a release with support for the newer api by then

### Proposed changes

- Update the POST payload format
- Update the authentification method and add a dependency on `google.auth` for that
- Fix the behavior in debug mode: Only use the testregion and per default don't actually send the messages
- Update the tests

### Side effects

Ideally everything should work as before 

### Resolved issues

Fixes: #1853 


svenseeberg:
The Integreat CMS config is prepared: https://git.tuerantuer.org/DF/salt/commit/faf95965ab6dfff83c54880bd0be057d096d937e

deen13 (approved):
Tested on my iOS device and it worked perfectly! 👍 

Please remember that if your device is set to English language, you'll need to publish English news to receive notifications. 😬

charludo:
@david-venhoff I'm getting an error from firebase `Firebase Cloud Messaging API has not been used in project [12-digit number here] before or it is disabled.`

Have I missed a step somewhere? The credentials file is being loaded correctly, as far as I can tell, although I am not 100% sure it's the *correct*  credentials.json file 😄

david-venhoff:
We currently have two credential files, one for google translate and one for fcm. The correct one should be titled `Integreat FCM Service Account Credentials` in passbolt

charludo:
> We currently have two credential files, one for google translate and one for fcm. The correct one should be titled `Integreat FCM Service Account Credentials` in passbolt

Oh, alright. Then I'm indeed using the wrong one. Can't seem to find the other one though, how do I request access to it? 🙈

david-venhoff:
Apparently I can give you access to that :smile: 
You should be able to see it now

charludo:
> Apparently I can give you access to that 😄 You should be able to see it now

Yes, nice, thank you!

charludo (approved):
Very nice! Everything seems to work 😄

Issue #2756, opened 2024-04-23
HIX-threshold for translation should use the uprounded value


dkehne:
Page: Residents' Registration Office (https://admin.integreat-app.de/dahme-spreewald/pages/de/32725/edit/). Hix value to see 15. color yellow. 
Page: National Visa (D-Visa) (https://admin.integreat-app.de/dahme-spreewald/pages/de/33077/edit/). Hix value to see 15. color green. 

![Screenshot (1489)](https://github.com/digitalfabrik/integreat-cms/assets/4182469/9a12eed3-ea7d-43d2-94a4-94ea0a611097)

If you move the mouse over it:
Where it is yellow, the value is: 14.995. Where it is green, the value is: 15.002.

If the HIX-Value is shown as 15 because it is rounded up from 14.995 it should be marked green and translations should be possible. So maybe we can adjust the logic here that the decision on the HIX-threshold is also done based on the rounded-up-value.

That would be great.

dkehne:
@PeterNerlich @MizukiTemma thanks for the implemenation but i guess the function is now doing sth. completly different now. 
It is now downgrading every hix-value. and also if i have a hix-value of 13.94 it is shown as 13. 

So sth. is not working here, not sure if it comes from this update here or another.

dkehne:
Two things that have changed but were not proposed to change:
- HIX-value is shown as "14" if it is "14.99".
- Detailled mouseover was removed (so it is not possible to see the exact hix value anymore)

See for example: https://admin.integreat-app.de/testumgebung/pages/de/37330/edit/

The only thing we wanted to be changed here is that if the "rounding up"-logic shows e.g. 14 (which is enough to translate) then it should be possible to translate.

MizukiTemma:
@dkehne sorry for the unwanted changes 🥺 

In this example
- HIX value in the hix bar must be 14,99
- Mouseover should show 14,988696629148205

Am I understanding correctly?

PR #2760, opened 2024-04-24
Use uprounded HIX value

MizukiTemma:
### Short description

This PR adjusts the hix widget and hix score check to align with the score shown in the hix widget.

### Proposed changes

- Use up-rounded value when
     1. toggling the color of HIX bar
     2. toggling the check box for machine translation
     3. checking HIX score for machine translation

### Side effects

- <s>There are three ways of rounding up HIX value: `|floatformat:2` (existing) in template, `Math.round(hixValue * 100) / 100` (existing) in Typescript and `round(self.hix_score * 100) / 100` in python (new). There might be cases where they do not align each another 🙈 </s>

### Resolved issues

Fixes: #2756 

### DIFF
--- a/integreat_cms/cms/models/abstract_content_translation.py
+++ b/integreat_cms/cms/models/abstract_content_translation.py
@@ -16,6 +16,7 @@ if TYPE_CHECKING:
     from .regions.region import Region
 
 from ..constants import status, translation_status
+from ..utils.round_hix_score import round_hix_score
 from ..utils.translation_utils import gettext_many_lazy as __
 from .abstract_base_model import AbstractBaseModel
 from .languages.language import Language
@@ -518,6 +519,13 @@ class AbstractContentTranslation(AbstractBaseModel):
         return self.foreign_object.hix_ignore
 
     @cached_property
+    def rounded_hix_score(self) -> float | None:
+        """
+        return rounded-up hix_score
+        """
+        return round_hix_score(self.hix_score)
+
+    @cached_property
     def hix_sufficient_for_mt(self) -> bool:
         """
         Whether this translation has a sufficient HIX value for machine translations.
@@ -525,7 +533,10 @@ class AbstractContentTranslation(AbstractBaseModel):
 
         :return: Wether the HIX value is sufficient for MT
         """
-        return self.hix_score is None or self.hix_score >= settings.HIX_REQUIRED_FOR_MT
+        return (
+            self.hix_score is None
+            or self.rounded_hix_score >= settings.HIX_REQUIRED_FOR_MT
+        )
 
     def __str__(self) -> str:
         """
--- a/integreat_cms/release_notes/current/unreleased/2756.yml
+++ b/integreat_cms/release_notes/current/unreleased/2756.yml
@@ -1,0 +1,2 @@
+en: Use rounded hix value
+de: Verwende gerundeten Hix-Wert
--- a/integreat_cms/cms/templates/analytics/_page_hix_row.html
+++ b/integreat_cms/cms/templates/analytics/_page_hix_row.html
@@ -1,6 +1,6 @@
 {% load i18n %}
 {% with page_translation.page.id as page_id %}
-    {% with page_translation.hix_score as hix %}
+    {% with page_translation.rounded_hix_score as hix %}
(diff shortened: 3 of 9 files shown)


integreat_cms/cms/models/abstract_content_translation.py
```diff
@@ -517,6 +517,15 @@ def hix_ignore(self) -> bool:
...
         return self.foreign_object.hix_ignore
 
+    @cached_property
+    def rounded_hix_score(self) -> float | None:
+        """
+        return rounded-up hix_score
+        """
+        if self.hix_score is None:
+            return None
+        return round(self.hix_score * 100) / 100
```
david-venhoff, line 527:
```suggestion
        return round(self.hix_score, ndigits=2)
```
I think we should use the `ndigits` parameter here, to make it more clear what happens

david-venhoff (commented):
Looks good to me in general, I have only one question:
Do we even need the unrounded hix score anymore? If not, I think it would be preferable to only change the value that gets returned from `TextlabApiClient.benchmark`, because we would not need to keep track of two different versions of the hix score then. Also the rounding logic would only be in one place.

deen13 (changes requested):
I believe we should avoid calculating the value in multiple locations. Doing so could introduce bugs that are extremely difficult to identify. Is it possible to utilize the [rounded_hix_score](https://github.com/digitalfabrik/integreat-cms/blob/db6e5e965ffb396f31e133821d2798ba10668f94/integreat_cms/cms/models/abstract_content_translation.py#L521) within the TypeScript code? ☺️

MizukiTemma:
@david-venhoff @deen13 
After [the discuttion in the chat](https://chat.tuerantuer.org/digitalfabrik/pl/uieg1sh8fpr9bxih79nak9we6h):
- An util function `round_up_hix_score` is introduced and used for ajax HIX value update and the cached propery `rounded_hix_score`
- To templates and Typescript only the rounded up values are passed
- `|floatformat:2` and `Math.round(hixValue * 100) / 100` are removed
- But for each translation the raw HIX value will be stored beause `page_translation_save_handler` does not use `round_up_hix_score`
I hope this centralizes the rounding up calculation but brings no side effect 🙈

integreat_cms/cms/utils/round_up_hix_score.py
```diff
@@ -0,0 +1,8 @@
+def round_up_hix_score(raw_score: float | None) -> float | None:
+    """
+    Function to round up HIX score
+    """
+
+    if raw_score is None:
+        return None
+    return round(raw_score, ndigits=2)
```
deen13, line 8: I would argue that this function could be better served by inlining it, as it is not particularly complex. This approach avoids the need for a new file (which readers would have to navigate to), and it results in fewer lines compared to its current implementation.

deen13 (changes requested):

integreat_cms/cms/utils/round_up_hix_score.py
MizukiTemma: I know your feeling 😅 Though I decided for a separate file because the function is used both in the model and the view and to avoid circular import...

deen13 (approved):

integreat_cms/release_notes/current/unreleased/2756.yml
```diff
@@ -0,0 +1,2 @@
+en: Use uprouned hix value
+de: Verwende aufgerundeten Hix-Wert
```
david-venhoff, line 2:
```suggestion
en: Use rounded hix value
de: Verwende gerundeten Hix-Wert
```

integreat_cms/cms/utils/round_up_hix_score.py
```diff
@@ -0,0 +1,8 @@
+def round_up_hix_score(raw_score: float | None) -> float | None:
+    """
+    Function to round up HIX score
```
david-venhoff, line 3:
```suggestion
def round__hix_score(raw_score: float | None) -> float | None:
    """
    Function to round HIX score
```

david-venhoff (approved):
Thanks, looks almost good now :tada:

I would just propose to change the terminology from `round up` to just `round`, because rounding up is something different than what is done here: https://en.wikipedia.org/wiki/Rounding#Rounding_up

Issue #2358, opened 2023-07-24
Giving feedback on the HIX value


seluianova:
### Motivation
TextLab API returns some analysis details on a POST ​/benchmark​/{id} request, that could potentially be delivered to users.

For example, there is a "dataTerms" attribute, which contains the following "categories":
- _**Filler Words**: Filler words fill up your sentences without adding any content to it. <br>In most cases you can simply drop them to improve your text._
- _**Finance and Insurance**: Many financial or insurance terms are tough to understand for laypersons.<br>TextLab helps you to detect such terms.<br>Replace them with more common terms!_

etc.

If some word falls into one of these categories, it will be listed in the "results" section of the category with (or without) the replacement suggestions. For example:

```
            "result": [
                {
                    "length": [
                        1
                    ],
                    "position": [
                        0
                    ],
                    "priority": 10,
                    "replacement": [],
                    "term": {
                        "check_words": 0,
                        "description": "Dieser Begriff kann negativ wirken.|Tipp: Versuchen Sie, auf den Begriff zu verzichten und neutral zu formulieren.",
                        "lemma": [
                            "unwiderruflich"
                        ],
                        "settings": {},
                        "tag": [
                            "ADJD"
                        ],
                        "term_id": 29531,
                        "wordcount": 1,
                        "words": [
                            "unwiderruflich"
                        ]
                    }
                }
            ]
```

### Proposed Solution
I guess we could build some kind of report for users with this data?

### Alternatives
TBD

### User Story
TBD

### Additional Context
Full response example: [benchmark-response.txt](https://github.com/digitalfabrik/integreat-cms/files/12148416/benchmark-response.txt)
The text being checked (just a set of "difficult" words): 
_Unwiderrufliches Unterversicherung Unterjährigkeitszuschlag Rechtsschutzversicherungsgesellschaften_

The response also includes a "languageTool" attribute with typos and grammatical errors, but I assume they don't affect the HIX value (do they?)

### Design Requirements
TBD

seluianova:
Sometimes, however, I don't quite understand the HIX calculation logic.
For example, if I analyze the sentence:

_Willkommen in Augsburg_

The HIX value is 11.72 but there is no "results" under "dataTerms".

osmers:
> Sometimes, however, I don't quite understand the HIX calculation logic. For example, if I analyze the sentence:
> 
> _Willkommen in Augsburg_
> 
[...]

PR #2761, opened 2024-04-24
Provide HIX feedback details

seluianova:
### Short description
Textlab API returns some analysis details in a json response to the benchmark request.
We would like to display them in the HIX widget according [to the design](https://www.figma.com/file/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?type=design&node-id=1013%3A646&mode=design&t=CoWs1wUEUXzZiKM4-1).

The following categories are handled in this PR:

> Schachtelsätze
> Lange Sätze
> Lange Wörter
> Sätze im Passiv
> Infinitiv-Konstruktionen
> Sätze im Nominalstil
> Sätze im Futur

image

### Proposed changes
- Add hix_feedback field to the page translation model
- Extract feedback details from the Textlab API response and save into DB

For example, this attribute represents sentences in nominal style:
```
   "countNominalStyle": [
        [
            34,
            41, 
            0
        ]
    ],
```
One array inside means that there is 1 such sentence .
For 2 sentences there will be 2 arrays:
```
   "countNominalStyle": [
        [
            34,
            41, 
            0
        ],
        [
            76,
            98, 
            1
        ],
    ],
```
etc.
- Add feedback block to hix widget html template

### Side effects
- I haven't found any but careful testing is welcome

### Resolved issues
Fixes: #2358


integreat_cms/static/src/js/analytics/hix-widget.ts
```diff
@@ -2,39 +2,43 @@ import { getContent } from "../forms/tinymce-init";
 import { getCsrfToken } from "../utils/csrf-token";
 
 let initialContent: string = null;
-let initialHixValue: number = null;
+let initialHixScore: number = null;
 
-/* Display the HIX value using a bar chart */
-const updateHixBar = (value: number, setOutdated: boolean) => {
-    const roundedHixValue = Math.round(value * 100) / 100;
+/**
```
seluianova, line 7: I started to do some refactoring here to maybe improve readability a bit, because it's quite a clutter now 🙈
But decided to finish it later in order not to delay the main feature as it's high prio

integreat_cms/static/src/js/analytics/hix-widget.ts
```diff
@@ -62,10 +66,60 @@ const setHixLabelState = (state: string) => {
     document.getElementById("hix-loading")?.classList.add("hidden");
 };
 
-const getHixValue = async () => {
+/**
+ * Mapping between category names in the HIX widget and in the Textlab API response
+ */
+const feedbackCategoriesMapping = new Map<string, string>([
```
seluianova, line 72: not sure if it's a best place for this kind of mapping though

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -5656,6 +5660,42 @@ msgstr "Der HIX-Wert konnte nicht berechnet werden."
 msgid "There is no page content"
 msgstr "Es gibt keinen Seiteninhalt"
 
+#: cms/templates/hix_widget.html
+msgid "For review:"
+msgstr "Zur Überprüfung"
```
JoeyStk, line 5665:
```suggestion
msgstr "Zur Überprüfung:"
```
Super nit-picky :see_no_evil:

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -5656,6 +5660,42 @@ msgstr "Der HIX-Wert konnte nicht berechnet werden."
...
+msgid "nested sentences"
+msgstr "Schachtelsätze"
+
+#: cms/templates/hix_widget.html
+msgid "sentences that are too long"
+msgstr "Sätze, die zu lang sind"
+
+#: cms/templates/hix_widget.html
+msgid "words that are too long"
+msgstr "Wörte, die zu lang sind"
```
JoeyStk, line 5681:
```suggestion
msgstr "Wörter, die zu lang sind"
```

JoeyStk (changes requested):
Thanks! This is a very cool feature!
So far I couldn't find any issues. I only have two minor comments about the translation file :)

integreat_cms/cms/templates/hix_widget.html
```diff
@@ -64,7 +65,61 @@
...
+            <div id="hix-feedback" class="hidden">
+                <label class="secondary">
+                    {% translate "For review:" %}
+                </label>
+                <div hix-feedback-category="hix-count-nested-sentences"
+                     class="help-text mt-2 hidden">
+                    <i icon-name="bar-chart-big" class="size-6 pr-1"></i>
+                    {% translate "The text contains" %}
+                    <span class="font-bold"></span>
+                    <span class="font-bold">{% translate "nested sentences" %}</span>
```
deen13, line 77: This seems unnecessary. I believe it would be better to look for an alternative solution if this is being used for layout.

integreat_cms/core/settings.py
```diff
@@ -291,6 +291,17 @@
...
+#: List of categories for which we want to process HIX feedback details provided by Textlab API
+TEXTLAB_API_FEEDBACK_CATEGORIES: Final[list[str]] = [
+    "moreSentencesInClauses",
+    "moreSentencesInWords",
+    "moreWordsInLetters",
+    "countPassiveVoiceInSentence",
+    "countInfinitiveConstructions",
+    "countNominalStyle",
+    "countFutureTenseInSentence",
+]
```
deen13, line 303: I haven't found anything like this, but if there's a way to generate it from their documentation (maybe swagger), that would be incredibly helpful to prevent api changes from silently breaking our code.

integreat_cms/static/src/js/analytics/hix-widget.ts
```diff
@@ -2,39 +2,43 @@ import { getContent } from "../forms/tinymce-init";
...
         backgroundColor = "rgb(16, 111, 254, 0.3)";
-    } else if (value > hixThresholdGood) {
+    } else if (score > hixThresholdGood) {
         backgroundColor = "rgb(74, 222, 128)";
-    } else if (value > hixThresholdOk) {
+    } else if (score > hixThresholdOk) {
         backgroundColor = "rgb(250, 204, 21)";
     } else {
         backgroundColor = "rgb(239, 68, 68)";
     }
```
deen13, line 30: This could be extracted into a function so that we get rid of mutable state. 

#### Calling Side
```js
const backgroundColor = calculateBackgroundColor(score)
```

#### Function Side
```js
const calculateBackgroundColor = (score, setOutdated) => {
  if (setOutdated) {
        return "rgb(16, 111, 254, 0.3)";
    } else if (score > hixThresholdGood) {
        return "rgb(74, 222, 128)";
    } else if (score > hixThresholdOk) {
        return "rgb(250, 204, 21)";
    } else {
        return "rgb(239, 68, 68)";
    }
}
```

This code introduces a pure function, `calculateBackgroundColor`, which is easy to read and also reduces complexity on the calling side for readers who don't need to understand how the color is calculated. Additionally, the `let` variable has been replaced with a `const`.

integreat_cms/static/src/js/analytics/hix-widget.ts
```diff
@@ -58,14 +62,71 @@ const setHixLabelState = (state: string) => {
...
+/**
+ * Request current HIX data
+ * @returns HIX score and HIX feedback
+ */
+const getHixData = async (): Promise<[number, string]> => {
     const updateButton = document.getElementById("btn-update-hix-value");
-    let result;
     const sentContent = getContent().trim();
+
+    let result;
```
deen13, line 128: The result can be assigned directly and then also changed from `let` to `const`. ☺️ 

```
const result = await fetch(updateButton.dataset.url, ...)
    .then((response) => response.json())
    .then((json) => {
        if (json.error || json.score === null) {
            updateHixStateLabel("error");
            return [];
        } else {
            updateHixStateLabel("updated");
            initialContent = sentContent;
            return [json.score, JSON.stringify(json.feedback)];
        }
    });
```

integreat_cms/static/src/js/analytics/hix-widget.ts
```diff
@@ -58,14 +62,71 @@ const setHixLabelState = (state: string) => {
...
+            category.classList.add("hidden");
+        }
+    });
+
+    const feedbackBlock: HTMLElement = document.getElementById("hix-feedback");
+    if (feedbackCount === 0) {
+        feedbackBlock.classList.add("hidden");
+    } else {
+        feedbackBlock.classList.remove("hidden");
+    }
```
deen13, line 117: I'd suggest to filter the feedbackCategories at the start of this function and later on use the length of that array to make the if condition more clear.

#### Sample
```
    // Filter categories with feedback
    const categoriesWithFeedback = Array.from(feedbackCategories).filter((category) => {
        const categoryName = category.getAttribute("hix-feedback-category");
        const feedbackByCategory = feedbackJson[feedbackCategoriesMapping.get(categoryName)];
        return feedbackByCategory?.length > 0;
    });

    // Loop through each category
    categoriesWithFeedback.forEach((category) => {
        const categoryCount = category.querySelector("span");
        const categoryName = category.getAttribute("hix-feedback-category");
        const feedbackByCategory = feedbackJson[feedbackCategoriesMapping.get(categoryName)];
        categoryCount.textContent = feedbackByCategory.length.toString(); // Convert to string before setting
    });

    // Show/hide feedback block based on feedback count
    const feedbackBlock = document.getElementById("hix-feedback");
    if (categoriesWithFeedback.length === 0) {
        feedbackBlock.classList.add("hidden");
    } else {
        feedbackBlock.classList.remove("hidden");
    }
```

deen13 (changes requested):
Good work! 🥇

integreat_cms/core/settings.py
deen13: Oh my initial comment was a bit misleading: I was thinking about accessing enum types instead of string definitions.

seluianova: Do you mean moving this from settings to a separate enum class?

deen13: Oops! My initial comment might have been a bit misleading. What I meant was accessing enum types instead of string definitions directly in our code. This way, we can ensure better type safety and catch any api changes early on.

deen13: I had a quick look at their OpenAPI definition, and it seems that these values are loosely typed in their system either. Please consider this thread solved. 🙈

seluianova: Ah, right, it's hardly possible to parse their docs. We had to contact a supporting person to localize the right fields in their API response 🙈

integreat_cms/cms/templates/hix_widget.html
david-venhoff: These spans get populated from javascript: https://github.com/digitalfabrik/integreat-cms/pull/2761/files#diff-3334c5517cdff73b29d44ccb02b3d4b1b384c84ccaa19345543ca485bcaa5a30R105

deen13: That makes sense. Thanks for the explanation @david-venhoff!

integreat_cms/core/settings.py
```diff
@@ -291,6 +291,17 @@
...
 
+#: List of categories for which we want to process HIX feedback details provided by Textlab API
+TEXTLAB_API_FEEDBACK_CATEGORIES: Final[list[str]] = [
+    "moreSentencesInClauses",
+    "moreSentencesInWords",
+    "moreWordsInLetters",
+    "countPassiveVoiceInSentence",
+    "countInfinitiveConstructions",
+    "countNominalStyle",
+    "countFutureTenseInSentence",
```
david-venhoff, line 302: I know these names come from the textlab api, but could you add some comment explaining their meaning?
I don't really understand what e.g. `moreSentencesInWords` is supposed to mean.

Also I am not really sure if the settings are the best place for this. For example if we wanted to add a new category, it would not be sufficient to only update the settings. We would still have to make a new release where the mapping in `static/src/js/analytics/hix-widget.ts` is updated.
If you decide not to use the settings, maybe you could have a mapping in the textlab api client?

Roughly like this:
```python
hix_feedback_mapping = {
    "moreSentencesInClauses": "hix-count-nested-sentences",
    "moreSentencesInWords": "hix-count-long-sentences",
    "moreWordsInLetters": "hix-count-long-words",
    # ...etc
}
feedback_details = {
    cms_name: response.get(api_name, [])
    for api_name, cms_name in hix_feedback_mapping.items()
}
```
Then you could get rid of the mapping in the typescript code and directly use these values

david-venhoff (commented):
Thanks :rocket: 
This looks mostly good to me, I would just suggest to not use `JsonResponse` as return type of the textlab api function

integreat_cms/cms/models/pages/page_translation.py
```diff
@@ -55,6 +55,12 @@ class PageTranslation(AbstractBasePageTranslation):
         verbose_name=_("HIX score"),
     )
 
+    hix_feedback = models.JSONField(
+        null=True,
```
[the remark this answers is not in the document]
seluianova: I believe we can't have a [] state now.
It's either null (if hix is disabled or an error occured) or json with the list of categories, but their values are []

integreat_cms/textlab_api/textlab_api_client.py
```diff
@@ -63,7 +64,18 @@ def benchmark(
         data = {"text": unescape(text), "locale_name": "de_DE"}
         path = f"/benchmark/{text_type}"
         response = self.post_request(path, data, self.token)
-        return response.get("formulaHix")
+
+        feedback_details = {
```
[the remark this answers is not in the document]
seluianova: Jein 😄
I have this explanation from Textlab, but tbh it didn't help me much

> At sentence level, the numbers are simply placed one below the other, i.e. start of the match, end of the match, length of the match. The length varies from the unit depending on the parameter. For nested sentences it indicates the number of sentence parts, for long sentences the number of words.
> 
> Long words only show the position of the hit and the length, in this case the number of letters.

integreat_cms/static/src/js/analytics/hix-widget.ts
seluianova: Then the result type will be Promise<Response>, while we need Promise<[number, string]>.

integreat_cms/static/src/js/analytics/hix-widget.ts
seluianova: const categoryName and const feedbackByCategory are created in 2 loops in this example

integreat_cms/static/src/js/analytics/hix-widget.ts
seluianova: thanks, I have applied that

integreat_cms/core/settings.py
seluianova: > For example if we wanted to add a new category, it would not be sufficient to only update the settings

True, but now we can remove categories without updating the code. 
No idea if this could be a useful though 🙈

seluianova: I don't have a clear opinion on where to put these categories. I'll try asking on the channel, maybe there are more opinions on this there.

integreat_cms/textlab_api/textlab_api_client.py
```diff
@@ -63,7 +64,18 @@ def benchmark(
...
+            category: response.get(category, [])
+            for category in settings.TEXTLAB_API_FEEDBACK_CATEGORIES
+        }
+
+        return JsonResponse(
+            {
+                "score": response.get("formulaHix"),
+                "feedback": feedback_details,
+            }
+        )
```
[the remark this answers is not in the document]
seluianova: as brilliant as it's simple.

integreat_cms/core/settings.py
seluianova: > Then you could get rid of the mapping in the typescript code and directly use these values

Maybe that's not bad, I will think of it next week.

PeterNerlich (commented):
I started a review but didn't continue and now it's no longer needed, so I'm just dumping my two cents here~~

integreat_cms/locale/de/LC_MESSAGES/django.po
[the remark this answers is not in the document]
seluianova: It seems you have reviewed an outdated version of the code somehow. This  has been already fixed 🙂

integreat_cms/locale/de/LC_MESSAGES/django.po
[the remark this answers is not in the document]
seluianova: And this is not actual as well

integreat_cms/static/src/js/analytics/hix-widget.ts
deen13: Sorry, here's an example that should work properly and address the following two points:

- It makes the result a const instead of a let, which removes mutability and makes the code slightly easier to read since the reader doesn't have to think about subsequent places that mutate the variables.
- It also adds error handling on the promise level, which was missing before. This could have led to a case where the server responds with anything except a status code in the range of 200 to 399, and then the client would have failed silently, if I understand correctly.

What are your thoughts on this?

```js
/**
 * Request current HIX data
 * @returns HIX score and HIX feedback
 */
const getHixData = async (): Promise<[number?, string?]> => {
    const updateButton = document.getElementById("btn-update-hix-value");
    const sentContent = getContent().trim();

    try {
        const response = await fetch(updateButton.dataset.url, {
            method: "POST",
            headers: {
                "X-CSRFToken": getCsrfToken(),
            },
            body: JSON.stringify({
                text: sentContent,
            }),
        });

        const json = await response.json();

        if (json.error || json.score === null) {
            updateHixStateLabel("error");
            return [];
        } else {
            updateHixStateLabel("updated");
            initialContent = sentContent;
            return [json.score, JSON.stringify(json.feedback)];
        }
    } catch (error) {
        updateHixStateLabel("error");
        return [];
    }
};
```

integreat_cms/textlab_api/textlab_api_client.py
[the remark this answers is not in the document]
david-venhoff: that sounds awful!

integreat_cms/cms/views/utils/hix.py
```diff
@@ -94,6 +94,7 @@ def get_hix_score(request: HttpRequest, region_slug: str) -> JsonResponse:
         logger.warning("Received invalid text: %r", text)
         return JsonResponse({"error": f"Invalid text: '{text}'"})
 
-    if score := lookup_hix_score(text):
-        return JsonResponse({"score": round_hix_score(score)})
```
seluianova, line 98: Need to revise this rounding change, because it's not applicable here anymore

integreat_cms/core/settings.py
seluianova: Moved the mapping to the textlab api client
Revised feedback json structure to match standard json style (property names in camel case)

integreat_cms/static/src/js/analytics/hix-widget.ts
seluianova: Also missing is the behavior of hiding/displaying relevant feedback categories (divs) depending on categoryCount.
When count is 0 - the feedback category should be hidden in the widget, otherwise - it should be displayed.

integreat_cms/static/src/js/analytics/hix-widget.ts
seluianova: > This could have led to a case where the server responds with anything except a status code in the range of 200 to 399

Currently the server only responds with 200, even if there is an error.
I believe this could be changed to be more in line with the http protocol, but not within the scope of this PR because we wanted to deliver HIX feedback with a high prio and refactoring existing functionality can be done later.

integreat_cms/static/src/js/analytics/hix-widget.ts
seluianova: Sorry but it also doesn't work as expected.
Hiding/displaying relevant feedback categories (divs) depending on categoryCount is missing.
When count is 0 - the feedback category should be hidden in the widget, otherwise - it should be displayed.

integreat_cms/textlab_api/textlab_api_client.py
```diff
@@ -15,6 +15,21 @@
 logger = logging.getLogger(__name__)
 
 
+"""
+Mapping between fields in Textlab API response containing feedback data
+and the corresponding category names used in the CMS
+"""
```
david-venhoff, line 21: Nitpick: this should probably be a normal comment, as docstrings are only used for module, class, method or function documentation (https://peps.python.org/pep-0257/#what-is-a-docstring)

david-venhoff (approved):
I think this is good to go after the comments from the other reviewers got resolved :+1:

JoeyStk:
I will leave it to @deen13 to do the second review :)

deen13 (approved):

PR #2766, opened 2024-04-29
Allow dynamic export of database objects

### LINKED ISSUE #2516: META: Large synthetic dataset for performance evaluation

david-venhoff:
### Short description

This is an experimental pr which adds a new api endopint that creates a data export which can be imported using `./tools/integreat-cms-cli loaddata path/to/json`

To preserve user privacy, only whitelisted models get exported and user account information will not be included in the export. Also only administrators are allowed to use this endpoint. 

This is currently a proof of concept and only exports regions, languages, pages and page translations

Please take a look at this and let me know whether this approach is something that could be incorporated :)

### Proposed changes

- Add a `dump` function that exports all objects of selected models as json.
- Add a new api endpoint `/v3/database` which uses this function to create a data export.

### Resolved issues

Fixes: #2516

### DIFF
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -9582,9 +9582,9 @@ msgstr "Englisch"
 msgid "Dutch"
 msgstr "Niederländisch"
 
-#: data_dump/apps.py
-msgid "Data dump"
-msgstr "Datenexport"
+#: database_dump/apps.py
+msgid "Database dump"
+msgstr "Datenbank-Export"
 
 #: deepl_api/apps.py
 msgid "DeepL API"
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -9582,6 +9582,10 @@ msgstr "Englisch"
 msgid "Dutch"
 msgstr "Niederländisch"
 
+#: data_dump/apps.py
+msgid "Data dump"
+msgstr "Datenexport"
+
 #: deepl_api/apps.py
 msgid "DeepL API"
 msgstr "DeepL API"
--- a/integreat_cms/api/urls.py
+++ b/integreat_cms/api/urls.py
@@ -8,7 +8,7 @@ from typing import TYPE_CHECKING
 
 from django.urls import include, path, re_path
 
-from .v3.database import database
+from .v3.database_dump import database_dump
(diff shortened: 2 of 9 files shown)


integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -9582,6 +9582,10 @@ msgstr "Englisch"
 msgid "Dutch"
 msgstr "Niederländisch"
 
+#: data_dump/apps.py
+msgid "Data dump"
```
ulliholtgrave, line 9586: I guess I prefer the term "Database dump" and "Datenbank-Export" in German.

ulliholtgrave (changes requested):
Nice idea and so far it works as expected. However, I have 2 questions:
1. I did test it with our default test data and it was sufficiently fast, but did you do some performance bench marking for a  bigger database?
2. We generally want to use this file as an import for our local setups, right?

david-venhoff:
> I did test it with our default test data and it was sufficiently fast, but did you do some performance bench marking for a bigger database?

No, I did not test it with a bigger database. I do expect it to be quite slow however, since we have so many page translations on the production database and a for-loop that has to go over every single one of them. Not sure if there is a way to make it faster.
I would argue that it should be fine however, since this operation should usually only needed once every few months per developer probably.

> We generally want to use this file as an import for our local setups, right?

Probably only for performance testing, unless we can optimize the cms enough that working with realistic test data is bearable.

ulliholtgrave (approved):
Thank! I was able to export a dataset and import it :+1: 
I guess we still have to think about how and in which form we want to make them available to our devs, but this is a good start

david-venhoff:
We decided to close this, because it is also possible to use database dumps, and the advantage of this approach is not really big enough to justify it

Issue #2757, opened 2024-04-23
Add script to create POIs from CSV


timobrembeck:
### Motivation

For Malte, we need the possibility to convert Excel files to POI objects.
Since Excel files might be hard to work with, I guess we can assume that it's always possible to export the excel file into CSV and use it that way.

### Proposed Solution

- Think about how to translate POI objects into tables (especially what about opening hours?)
    - If passing the opening hours as columns is too hard, maybe we can use our internal structure and pass them via a serialized json string in one csv cell?
- Write a short script that can be executed inside `integreat-cms-cli shell` or add a management command (preferred if the time suffices) to convert a csv into POI objects.
- At least one translation in the default language has to be created

### User Story

Users want to get POIs from Excel into the CMS

### Design Requirements

No UI necessary

osmers:
What's the progress with this issue? @timobrembeck @PeterNerlich ?

PR #2038, opened 2023-01-30
Add organization and barrier-free fields to POI

MizukiTemma:
### Short description

This PR adds two new fields `organization` and `barrier_free` to POI.

### Proposed changes

- add those fields to POI model, form, template and API response.
- adjust and expand the test by creating a test organization and setting this to the test location.

### Side effects

- None (I hope)

### Resolved issues

Fixes: #2037

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #2768, opened 2024-04-30
Add possibility to import locations from CSV files

timobrembeck:
### Short description

Add a management command to import locations from CSV files

To test this, execute e.g.:
```
./tools/integreat-cms-cli import_pois_from_csv ./tests/core/management/commands/assets/pois_to_import.csv testumgebung root
```

Thanks in advance to the hero who sacrifices their holiday to review this! :pray:

### Proposed changes

- Fix logging for unsaved content objects
- Add possibility to import locations from CSV files

### Caveats

- Tests are missing, but due to the deadline, I will contribute them in another PR

### Resolved issues

Fixes: #2757


JoeyStk (commented):
Thank you very much! This is a really cool feature. I'm quite excited for it :blush: 

There a few things I don't fully understand yet. Please find my questions about them below :slightly_smiling_face: 

1. When opening `pois_to_import.csv` in Libre Calc, I get the following result. 

grafik

In row BB for two rows there is "true" filled in, however there is no title in the respective column's head. Is this just a typo in the test or are a few column's titles missing?

2. You're using the CSV library, which totally makes sense in this use case. For the feedback we used the import-export-library. Honest question, because I don't know: Is it okay to use two different libraries which have more or less the same goal, or should we decide on one csv-library in the near future to stay consistent?

3. I used the command `./tools/integreat-cms-cli import_pois_from_csv ./tests/core/management/commands/assets/pois_to_import.csv testumgebung root` to import the test POIs and they successfully appeared in the POI list (I didn't go through every column yet to double-check if the content is correct). Then I went inside the file (first in VS Code, then in Libre Calc) and changed Wertachstraße 29 to Wertachstraße 28 and ran the command again. Both times it didn't update the row as I would have expected. Then I added a line for the Elbphilharmonie and ran the command once again. And also this time it wasn't imported. Again, a maybe somewhat naive but honest question, is this supposed to happen or is this a bug?

timobrembeck:
> 1. When opening `pois_to_import.csv` in Libre Calc, I get the following result.
> 
> In row BB for two rows there is "true" filled in, however there is no title in the respective column's head. Is this just a typo in the test or are a few column's titles missing?

Ah, good catch! This was an error in the test rows, I somehow managed to invent the 8-day-week :grin:
Should be fixed now!

> 2. You're using the CSV library, which totally makes sense in this use case. For the feedback we used the import-export-library. Honest question, because I don't know: Is it okay to use two different libraries which have more or less the same goal, or should we decide on one csv-library in the near future to stay consistent?

Didn't even think about it, good question for the long term, in general you're right and we should decide on one library. For now, I'd stick with the inbuilt csv library of python for this PR to be able to merge it today.

> 3. I used the command `./tools/integreat-cms-cli import_pois_from_csv ./tests/core/management/commands/assets/pois_to_import.csv testumgebung root` to import the test POIs and they successfully appeared in the POI list (I didn't go through every column yet to double-check if the content is correct). Then I went inside the file (first in VS Code, then in Libre Calc) and changed Wertachstraße 29 to Wertachstraße 28 and ran the command again. Both times it didn't update the row as I would have expected. Then I added a line for the Elbphilharmonie and ran the command once again. And also this time it wasn't imported. Again, a maybe somewhat naive but honest question, is this supposed to happen or is this a bug?

I did not implement a de-duplication mechanism, since there are POIs with the same name yet have multiple places. So all rows are always imported as new POIs and existing POIs are not updated. Could you check the output of the management command when importing the Elbphilharmonie POI? If something went wrong, it should display an error in the console.

JoeyStk:
> I somehow managed to invent the 8-day-week 😁

:sweat_smile:

> Should be fixed now!

Yes! Looks good! Thank you!

> For now, I'd stick with the inbuilt csv library of python for this PR to be able to merge it today.

:+1: 

> Could you check the output of the management command when importing the Elbphilharmonie POI? If something went wrong, it should display an error in the console.

Ah, I found out what it was. The message in the terminal was a success message and also restarting the server didn't help, but clearing my redis-cache and then restarting the server did do the trick. I should have thought of that! Sorry!
Now all of the data shows up as expected!

I did some more manual checking and think everything looks good, only one question arose: is there a GUI element for barrier-free? I wasn't able to find the element neither in the HTML source code nor in the GUI itself.

JoeyStk (approved):
:tada:

timobrembeck:
> Ah, I found out what it was. The message in the terminal was a success message and also restarting the server didn't help, but clearing my redis-cache and then restarting the server did do the trick. I should have thought of that! Sorry! Now all of the data shows up as expected!

Thanks for the hint, I will clear the cache after importing POIs on the prod server :+1:

> I did some more manual checking and think everything looks good, only one question arose: is there a GUI element for barrier-free? I wasn't able to find the element neither in the HTML source code nor in the GUI itself.

Oh good catch, apparently the barrier-free & organization fields were added in #2038 and shortly thereafter removed from the form as an unintended side-effect of #2043, probably a mistake of a conflict resolution? If you want, you can open an ticket for this bug.

JoeyStk:
> Oh good catch, apparently the barrier-free & organization fields were added in https://github.com/digitalfabrik/integreat-cms/pull/2038 and shortly thereafter removed from the form as an unintended side-effect of https://github.com/digitalfabrik/integreat-cms/pull/2043, probably a mistake of a conflict resolution? If you want, you can open an ticket for this bug.

Oh no! Sorry! 
Issue and PR are opened!

PR #2773, opened 2024-05-02
Use rounded HIX score for dashboard

MizukiTemma:
### Short description

This is a follow up PR for #2760, in which still raw HIX score is used in the dashboard/analytics view.

How to reproduce:
1. Enable HIX in Augsburg and fill its pages with HIX value by `hix_bulk`.
2. Go to Augsburg, find 9 in "Pages with low hix score" in ToDo list and "Overview of the text understandability of the pages".
3. Add the two pages listed in the Issue #2756 
4. See 10 instead of 9 in "Pages with low hix score" and "Overview of the text understandability of the pages", while both the pages shuold be good enough in the HIX score check.

### Proposed changes

- Use rounded HIX score

### Side effects

- Tell me please if you find other points that are to be adjusted too 👀 

### Resolved issues

Fixes: #

### DIFF
--- a/integreat_cms/cms/views/dashboard/dashboard_view.py
+++ b/integreat_cms/cms/views/dashboard/dashboard_view.py
@@ -160,7 +160,7 @@ class DashboardView(TemplateView, ChatContextMixin):
             "relevant_url": invalid_url,
         }
 
-    def get_low_hix_value_context(self) -> dict[str, QuerySet]:
+    def get_low_hix_value_context(self) -> dict[str, list[PageTranslation]]:
         r"""
         Extend context by info on pages with low hix value
 
@@ -169,12 +169,18 @@ class DashboardView(TemplateView, ChatContextMixin):
         if not settings.TEXTLAB_API_ENABLED or not self.request.region.hix_enabled:
             return {}
 
-        translations_under_hix_threshold = PageTranslation.objects.filter(
+        translations = PageTranslation.objects.filter(
             language__slug__in=settings.TEXTLAB_API_LANGUAGES,
             id__in=self.latest_version_ids,
             page__hix_ignore=False,
             hix_score__lt=settings.HIX_REQUIRED_FOR_MT,
         )
+
+        translations_under_hix_threshold = [
+            translation
+            for translation in translations
+            if not translation.hix_sufficient_for_mt
+        ]
 
         return {
             "pages_under_hix_threshold": translations_under_hix_threshold,
--- a/integreat_cms/cms/views/analytics/translation_coverage_view.py
+++ b/integreat_cms/cms/views/analytics/translation_coverage_view.py
@@ -148,7 +148,7 @@ class TranslationCoverageView(TemplateView):
 
         # Get the number of translations which are not ready for MT
         not_ready_for_mt_count = sum(
-            pt.hix_score < settings.HIX_REQUIRED_FOR_MT
+            pt.rounded_hix_score < settings.HIX_REQUIRED_FOR_MT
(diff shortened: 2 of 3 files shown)


integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -169,13 +169,18 @@ def get_low_hix_value_context(self) -> dict[str, QuerySet]:
         if not settings.TEXTLAB_API_ENABLED or not self.request.region.hix_enabled:
             return {}
 
-        translations_under_hix_threshold = PageTranslation.objects.filter(
+        translations = PageTranslation.objects.filter(
             language__slug__in=settings.TEXTLAB_API_LANGUAGES,
             id__in=self.latest_version_ids,
             page__hix_ignore=False,
```
david-venhoff, line 175:
```suggestion
            page__hix_ignore=False,
            hix_score__lt=settings.HIX_REQUIRED_FOR_MT
```
It would probably make sense to speed this up a bit, since we have a lot of page translations on the production server and probably want to avoid iterating through many

david-venhoff (approved):
:+1:

JoeyStk (changes requested):
Thank you!
Could you please add a test for the rounded hix score? :) 

And secondly, I'm not really an expert on caches and loading time, so I'm just going to ask. Do you think this PR could negatively influence the loading times on the dashboard or does the caching equalize this?

MizukiTemma:
@JoeyStk 
A small test is added 😸 

 >And secondly, I'm not really an expert on caches and loading time, so I'm just going to ask. Do you think this PR could negatively influence the loading times on the dashboard or does the caching equalize this?

Me neither can judge it, thought I don't come up with any alternative 🤔

JoeyStk:
> A small test is added 😸

Thank you. Looks good.

> Me neither can judge it, thought I don't come up with any alternative 🤔

An alternative that has been discussed in the past was to do it via AJAX, but I'm not sure if it applies here too. At the moment I don't feel confident enough to approve this PR. I think we should raise this issue in the team call and hopefully we get a clear answer there.

charludo:
> negatively influence the loading times on the dashboard

Maybe I'm missing something, but this just adds one iteration through the already-queried pages, right?

charludo (approved):
> > > negatively influence the loading times on the dashboard
> > 
> > 
> > Maybe I'm missing something, but this just adds one iteration through the already-queried pages, right?
> 
> yes :)

:+1:

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -169,13 +169,19 @@ def get_low_hix_value_context(self) -> dict[str, QuerySet]:
...
             hix_score__lt=settings.HIX_REQUIRED_FOR_MT,
         )
 
+        translations_under_hix_threshold = [
```
JoeyStk, line 179: Why do we do this calculation here? To me it's doing the exact same thing as before :thinking:

MizukiTemma: `hix_score__lt=settings.HIX_REQUIRED_FOR_MT` collects translation with *raw* HIX score < threshold, but `not translation.hix_sufficient_for_mt` in the next step checks whether *rounded* HIX score satisfies the threshold or not. 

Both `rounded_hix_score` and `hix_sufficient_for_mt` are cached_property and cannot used in the query, but we wanted to avoid checking all the translation by loop. So the code is first filtering translations out that definetely have good HIX score and checks the edge cases of up-rounding after that.

Issue #2714, opened 2024-03-27
The "search & replace" function for the Links does not change the link, but says it does.


vthejay:
### Steps to Reproduce

1. Go to 'Landkreis München. Go to Broken Links.'
2. Click on "suchen & ersetzen". Try to chenge this Link "https://www.donnamobile.org/_files/ugd/c2311b_bad0cc64a70047e0b636dbd00689a810.pdf" with this one "https://www.donnamobile.org/_files/ugd/c2311b_f6c7f7aeb7044ec5b0317a8401997f35.pdf"
3. After some time, the message is shown that the links were replaced successfully.
4. However, on the page the old link is still shown.
https://integreat.app/lkmuenchen/de/willkommen-lkmuenchen/m%C3%A4dchen-und-frauen-2/gruppen-und-kurse-f%C3%BCr-frauen-und-m%C3%A4dchen
The Link for "Afghanisch" was replaced there.

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

vthejay:
Please feel free to put it in the appropriate Milestone...

PR #2779, opened 2024-05-06
Make input fields of link replacement longer

MizukiTemma:
### Short description

This PR fixes the probelem that the central link replacement showed the success message but the replacement was not done as expected.

The cause of reported problem is the length of fields for old and new links: the replacement was conducted successfully but the link was changed only in the first letters.

### Proposed changes

- Make the input fields accept longer links

### Side effects

- None? 😅 

### Resolved issues

Fixes: #2714

### DIFF
--- a/integreat_cms/cms/forms/linkcheck/link_replace_form.py
+++ b/integreat_cms/cms/forms/linkcheck/link_replace_form.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 from django import forms
+from django.conf import settings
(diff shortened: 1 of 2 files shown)


seluianova (approved):
LGTM!
Release note left ?

david-venhoff (commented):
Should we maybe use `settings.LINKCHECK_MAX_URL_LENGTH` for that instead of hardcoding the value? Otherwise we might have the same issue again if we decide to change the link max length

david-venhoff (approved):

PR #2780, opened 2024-05-06
Fix outdated references to FCM_KEY

david-venhoff:
### Short description

In #2759 I forgot to update references to `FCM_KEY` which got renamed in that pr. 

### Proposed changes

- Rename all occurences of `FCM_KEY` to `FCM_CREDENTIALS`, which is the new name

### DIFF
--- a/integreat_cms/core/settings.py
+++ b/integreat_cms/core/settings.py
@@ -200,10 +200,10 @@ FCM_URL: Final[str] = (
 )
 
 #: Path to the saved credential json file
-FCM_CREDENTIALS: Final[str | None] = os.environ.get("INTEGREAT_CMS_FCM_CREDENTIALS")
+FCM_CREDENTIALS: str | None = os.environ.get("INTEGREAT_CMS_FCM_CREDENTIALS")
 
 #: Whether push notifications via Firebase are enabled.
-#: This is ``True`` if :attr:`~integreat_cms.core.settings.FCM_KEY` is set, ``False`` otherwise.
+#: This is ``True`` if :attr:`~integreat_cms.core.settings.FCM_CREDENTIALS` is set, ``False`` otherwise.
 FCM_ENABLED: bool = bool(FCM_CREDENTIALS)
 
 #: The available push notification channels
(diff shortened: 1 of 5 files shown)


tools/_functions.sh
```diff
@@ -181,9 +181,9 @@ function require_installed {
             INTEGREAT_CMS_DEBUG=1
             export INTEGREAT_CMS_DEBUG
             # Set dummy FCM key to test functionality
-            if [[ -z "${INTEGREAT_CMS_FCM_KEY}" ]]; then
-                INTEGREAT_CMS_FCM_KEY="dummy"
-                export INTEGREAT_CMS_FCM_KEY
+            if [[ -z "${INTEGREAT_CMS_FCM_CREDENTIALS}" ]]; then
```
david-venhoff, line 184: Not sure why that did not cause any test failures :shrug:

timobrembeck: Why should it? If FCM is not enabled, it will just skip the tests, so they won't fail.

integreat_cms/core/settings.py
```diff
@@ -200,10 +200,10 @@
 )
 
 #: Path to the saved credential json file
-FCM_CREDENTIALS: Final[str | None] = os.environ.get("INTEGREAT_CMS_FCM_CREDENTIALS")
+FCM_CREDENTIALS: str | None = os.environ.get("INTEGREAT_CMS_FCM_CREDENTIALS")
 
 #: Whether push notifications via Firebase are enabled.
-#: This is ``True`` if :attr:`~integreat_cms.core.settings.FCM_KEY` is set, ``False`` otherwise.
+#: This is ``True`` if :attr:`~integreat_cms.core.settings.FCM_CREDENTIALS` is set, ``False`` otherwise.
 FCM_ENABLED: bool = bool(FCM_CREDENTIALS)
```
timobrembeck, line 207: Maybe it would make sense to only enable it here if the variable is a valid cred file which exists and is readable?
To me, it does not make sense that a dummy path of "dummy" which does not exist just enables the push news without any errors.
But ok, this also makes it harder to set dummy values :sweat_smile:

david-venhoff: Maybe we could add some warnings in case of invalid configuration?

But that should probably be done in another pr, because this would also apply to at least `GOOGLE_TRANSLATE_ENABLED`

charludo (approved):
Nice, LGTM!

MizukiTemma (approved):
👍

Issue #2763, opened 2024-04-29
Adding a new model breaks multiple tests


charludo:
### Describe the Bug

Simply adding a new model leads to >60 test failures. The error source appears to be the changing of `content_type` and/or `polymorphic_ctype` ids compared to the `test_data.json` fixture. However, changing these to natural keys does not solve the problem.

### Steps to Reproduce

1. add a new model (can be completely empty)
2. import it in `integreat_cms/cms/models/\_\_init\_\_.py``
3. run `./tools/test.sh`

### Expected Behavior

Tests pass, since the inclusion of a new model should not interfer with the current tests

### Actual Behavior

\>60 failures, + numerous errors. The (probably?) most relevant failures are of the type
`FAILED tests/cms/views/feedback/test_region_feedback_actions.py::test_restore_region_feedback[ROOT] - polymorphic.models.PolymorphicTypeInvalid: ContentType 24 for <class 'integreat_cms.cms.models.pages.page_translation.PageTranslation'> #5 does not point to a subclass!`

### Additional Information

Relevant discussion [on Mattermost](https://chat.tuerantuer.org/digitalfabrik/pl/tqcgcn6477rqd8na18opdtq35e), MRE on [this branch](/digitalfabrik/integreat-cms/tree/bug/userchat-errors)

MRE test failures with our current fixtures: [CircleCI](https://app.circleci.com/pipelines/github/digitalfabrik/integreat-cms/11279/workflows/4c37d3bc-39dc-4245-b3bf-ed7a02a09c7d/jobs/135867)
MRE test failures with natural keys\* for `content_type` and `polymorphic_ctype`: [CircleCI](https://app.circleci.com/pipelines/github/digitalfabrik/integreat-cms/11379/workflows/3a4c339f-b57d-4b0f-9d53-4c786c3550ac/jobs/137586) 

\* currently, the test data is exported using hardcoded IDs for these fields. To use natural keys instead: `./tools/integreat-cms-cli dumpdata --natural-primary --natural-foreign`

PR #2781, opened 2024-05-06
Fix loadtestdata not working after adding a new model

david-venhoff:
### Short description

This pr attempts to fix the problem that the test data cannot be loaded anymore after a new model gets added.

I apologize for the huge diff :see_no_evil: 
is there maybe a way to reduce the diff, other than doing it manually?

### Proposed changes

- Regenerate the test data with `./tools/integreat-cms-cli dumpdata --natural-primary --natural-foreign --o test_data.json`
- Handle an exception that occurs when loading feedback and the base object is not available yet

### Resolved issues

Fixes: #2763

### DIFF
--- a/integreat_cms/cms/fixtures/test_data.json
+++ b/integreat_cms/cms/fixtures/test_data.json
@@ -1,244 +1,82 @@
 [
   {
-    "model": "cms.user",
+    "model": "cms.offertemplate",
     "pk": 1,
     "fields": {
-      "password": "pbkdf2_sha256$150000$6GBOYJT9gM6Z$fUrVoa16jJuWcG5G/cEjEGUZ14e9DWJAus8ygWHN4nk=",
-      "last_login": null,
-      "is_superuser": true,
-      "username": "root",
-      "first_name": "Root",
-      "last_name": "User",
-      "email": "root@root.root",
-      "is_staff": true,
-      "is_active": true,
-      "date_joined": "2019-11-13T09:21:34.960Z",
-      "groups": [5],
-      "user_permissions": [],
-      "organization": null,
-      "regions": [],
-      "expert_mode": true,
-      "page_tree_tutorial_seen": true
+      "name": "Sprungbrett",
+      "slug": "sprungbrett",
+      "thumbnail": "https://cms.integreat-app.de/wp-content/uploads/extra-thumbnails/sprungbrett.jpg",
+      "url": "https://web.integreat-app.de/proxy/sprungbrett/app-search-internships?location=",
+      "post_data": {},
+      "use_postal_code": "GET",
+      "supported_by_app_in_content": false,
+      "is_zammad_form": false,
+      "created_date": "2019-10-09T16:42:45.256Z",
+      "last_updated": "2019-10-09T16:44:36.254Z"
     }
   },
   {
-    "model": "cms.user",
+    "model": "cms.offertemplate",
     "pk": 2,
     "fields": {
-      "password": "pbkdf2_sha256$150000$6GBOYJT9gM6Z$fUrVoa16jJuWcG5G/cEjEGUZ14e9DWJAus8ygWHN4nk=",
-      "last_login": null,
-      "is_superuser": false,
-      "username": "management",
-      "first_name": "Region",
-      "last_name": "Manager",
-      "email": "management@example.com",
-      "is_staff": false,
-      "is_active": true,
-      "date_joined": "2019-11-13T09:21:34.960Z",
-      "groups": [1],
-      "user_permissions": [],
-      "organization": null,
-      "regions": [1],
-      "expert_mode": true,
-      "page_tree_tutorial_seen": true
+      "name": "Lehrstellenradar",
+      "slug": "lehrstellen-radar",
+      "thumbnail": "https://cms.integreat-app.de/wp-content/uploads/extra-thumbnails/lehrstellen-radar.jpg",
+      "url": "https://www.lehrstellen-radar.de/5100,0,lsrlist.html",
+      "post_data": { "search-ls": "1", "search-pr": "1", "search-radius": "50" },
+      "use_postal_code": "POST",
+      "supported_by_app_in_content": false,
+      "is_zammad_form": false,
+      "created_date": "2019-10-09T16:44:06.983Z",
+      "last_updated": "2019-10-09T16:44:06.992Z"
     }
   },
   {
-    "model": "cms.user",
+    "model": "cms.offertemplate",
     "pk": 3,
     "fields": {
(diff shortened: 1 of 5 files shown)


charludo (commented):
Thank you so much!! (Including for the caching explanation comment.)

I *think* only feedback and linkcheck fixtures are affected, so maybe a way to reduce the diff would be to only update those?

timobrembeck (commented):
> is there maybe a way to reduce the diff, other than doing it manually?

Since we created the database dump mainly manually, without a consistent ordering by ids/alphabet, I doubt there is. But I guess it's an acceptable drawback given the better consistency now.
For the future, we could stick to the internal database order by re-exporting the test data every time we add something to it.
Should we document this workflow somewhere?

> I think only feedback and linkcheck fixtures are affected, so maybe a way to reduce the diff would be to only update those?

I think the usage of natural keys is beneficial for all models, because the ids of the `ContentType` objects might change when new cms models are added.

charludo:
> Should we document this workflow somewhere?

I actually think a devtool might be necessary - to ensure that all wanted objects are imported, always with natural keys, without logs,...

charludo (approved):
Since my suggestion about only updating the relevant bits appears to be outdated now, I'll change this to an approval :)

Again, thank you very much!!

david-venhoff:
> I actually think a devtool might be necessary (in a new PR)

I agree too. I'll open a new issue for that

svenseeberg:
We had a UX Design meeting last week and we decided it would be best to have a random user exposure with a defined chance of the feature being visible to app users. That means we want to be able to set a chance value per region that allows the app to decide whether it should show the feature or not. Would it be possible to include a `chat_feature_probability` value per region that is also returned in the API? The app would do the rest.

charludo:
@david-venhoff could you re-do the export after rebasing on develop?

integreat_cms/core/signals/feedback_signals.py
```diff
@@ -61,7 +61,12 @@ def feedback_create_handler(sender: ModelBase, **kwargs: Any) -> None:
...
-        feedback_ptr := getattr(instance, "feedback_ptr", None)
-    ):
-        invalidate_obj(feedback_ptr)
+    try:
+        if (instance := kwargs.get("instance")) and (
+            feedback_ptr := getattr(instance, "feedback_ptr", None)
+        ):
+            invalidate_obj(feedback_ptr)
+    except Feedback.DoesNotExist:
+        # When installing fixtures, the related feedback might not yet be initialized.
```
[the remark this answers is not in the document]
david-venhoff: I assume that the occurrence of this bug depends on the order in which the test data are loaded. Since the test data file got completely reshuffled with this pr, that might be a likely explanation. I do not know for sure though.

charludo:
Awesome that this appears to work now @david-venhoff!! 🥳 
@timobrembeck @seluianova could either of you have another look at this? Apparently this bug is blocking multiple issues now, not just the chat.

seluianova (approved):
I tried several times to reload the data and re-run the tests, everything worked smoothly.

integreat_cms/core/signals/feedback_signals.py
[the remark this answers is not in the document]
seluianova: a bit puzzling to me, but okay :)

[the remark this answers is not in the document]
seluianova: ah. wait, I've got an idea

integreat_cms/cms/fixtures/test_data.json
```diff
@@ -3062,1742 +3841,2597 @@
...
+      "embedded_offers": []
     }
   },
+  { "model": "cms.regionfeedback", "pk": 1, "fields": { "feedback_ptr": 1 } },
+  { "model": "cms.regionfeedback", "pk": 4, "fields": { "feedback_ptr": 4 } },
+  { "model": "cms.regionfeedback", "pk": 7, "fields": { "feedback_ptr": 7 } },
+  { "model": "cms.regionfeedback", "pk": 8, "fields": { "feedback_ptr": 8 } },
+  { "model": "cms.poifeedback", "pk": 6, "fields": { "feedback_ptr": 6, "poi_translation": 1 } },
+  { "model": "cms.pagefeedback", "pk": 2, "fields": { "feedback_ptr": 2, "page_translation": 1 } },
+  { "model": "cms.pagefeedback", "pk": 5, "fields": { "feedback_ptr": 5, "page_translation": 1 } },
```
seluianova, line 3854:
```suggestion
```

integreat_cms/cms/fixtures/test_data.json
```diff
@@ -3062,1742 +3841,2597 @@
...
+      "machine_translated": false,
+      "version": 4,
+      "minor_edit": false,
+      "last_updated": "2023-10-11T16:57:08.587Z",
+      "creator": ["root"],
+      "automatic_translation": false,
+      "page": 1
     }
   },
+  { "model": "cms.eventfeedback", "pk": 3, "fields": { "feedback_ptr": 3, "event_translation": 1 } },
```
seluianova, line 6003:
```suggestion
```

david-venhoff: I would like to avoid manually changing test data, because that would make it very hard to add new test data later (We would have to manually type out the new test data, or automatically generate it and the apply your suggested changes every time). 
I have searched a bit and found this issue: https://code.djangoproject.com/ticket/8399
There, the proposed solution seems to add a check to signal handlers for a `raw` argument and return if that is True. We could add a decorator that does this to our signal handlers. Would you be okay with that?

seluianova: For example, if feedback_create_handler() is executed after "cms.eventfeedback" is created, but before "cms.feedback", and we just "pass", perhaps this would cause inconsistent feedback state in the cache, or something ?

seluianova: > We could add a decorator that does this to our signal handlers. Would you be okay with that?

The concept sounds good to me

david-venhoff: I think that should be fine. This specific code is only relevant when we update an existing feedback. But in general, if we ever find issues with that, we could just invalidate the entire cache after loading test data (Maybe that even happens already?)

david-venhoff: I also just found this note in the django documentation which explicitly states that we should not query other models during loaddata: https://docs.djangoproject.com/en/5.0/ref/signals/#pre-save

david-venhoff:
@charludo @seluianova Could you have a look at the new commit? If everything is fine I will squash and merge :)

charludo (approved):
I do not 100% grasp how this works now, but since all checks are green and it looks like this is closer to what the docs encourage. 

It also looks like adding a new model works without problems. Mission accomplished IMO :D

seluianova (approved):
Looks perfect to me, thanks!
Would you also update the docs to describe the data generation process, or create a task for that?

PR #2788, opened 2024-05-07
Fix test discovery in vscode

deen13:
### Short description

This pull request addresses the test configuration and discovery issues in VSCode while eliminating the need for the workaround involving tasks.

### Proposed changes
- Add environment variables that are required for test execution to devcontainer.json
- Eliminate the test tasks, as they were employed solely as a workaround
- Enable pytest in devcontainers
- Update README.md

### Side effects
- Happier devs 🤗 

### Screenshot
Screenshot 2024-05-07 at 16 21 03


.env
charludo, line 1: Hm, I'm not sure I like the introduction of this file. Including only this subset of configuration options is rather inconsistent, while including all configuration options could hinder existing workflows. I'd actually much rather add `.env` to the `.gitignore` and add these values to the setup instructions.

It's also generally discouraged to track `.env` files in git.

charludo: Alternatively, could you instead [set these values through the devcontainer `Dockerfile`](https://stackoverflow.com/a/69520132)?

charludo (changes requested):

.env
deen13: Good idea! I moved the environment variables to the [remoteEnv](https://containers.dev/implementors/json_reference/) field of [devcontainer.json](https://github.com/digitalfabrik/integreat-cms/blob/fix/vscode-test-discovery/.devcontainer/devcontainer.json).

david-venhoff (approved):
Thanks!

As we discussed, adding the default interpreter path could be helpful, but that can also be done in another pr :)

Issue #2778, opened 2024-05-06
Add abbreviations count to HIX feedback


seluianova:
### Motivation
Follow up to #2358

As part of #2358, we added a display of HIX feedback extracted from the Textlab API response.
Currently we display the following feedback categories:
```
Schachtelsätze
Lange Sätze
Lange Wörter
Sätze im Passiv
Infinitiv-Konstruktionen
Sätze im Nominalstil
Sätze im Futur
```
All of them are extracted **from the root properties** from the response.
For example, to count the number of sentences in the nominative case, we read the property "countNominalStyle", where the number of arrays means the number of sentences.
1 array - means 1 matching sentence:
```
    "countNominalStyle": [
        [
            34,
            41,
            0
        ]
    ],
```
2 arrays mean 2 sentences:
```
    "countNominalStyle": [
        [
            2,
            65,
            56
        ],
        [
            70,
            133,
            56
        ]
    ],
```

Besides that, there is another "source" of feedback that can be extracted from the "dataTerms" properties.
For example:
```
    "dataTerms": {
        "1289": {
            "category": {
                "category_type": "negative",
                "description": {
                    "de": "Abkürzungen erschweren den Lesefluss und werden häufig auch nicht verstanden. <br>Deshalb markiert TextLab alle Abkürzungen im Text. <br>Tipp: Verwenden Sie Abkürzungen möglichst selten. <br>Falls Sie wenig gebräuchliche Abkürzungen verwenden wollen, <br>erklären Sie sie dem Leser.",
                    "en": "Abbreviations make it harder to read and understand your text.<br>Tip: Try to use abbreviations spraringly. <br>If you need to use lesser known abbreviations, explain them to your readers."
                },
                "id": 1289,
                "locale_name": "de_DE",
                "name": {
                    "de": "Abkürzungen",
                    "en": "Abbreviations for Cockpit Integreat"
                },
                "settings": {}
            },
            "result": [
                {
                    "length": [
                        1,
                        1
                    ],
                    "position": [
                        9,
                        38
                    ],
                    "priority": 6,
                    "replacement": [
                        {
                            "description": "",
                            "global_visible": 1,
                            "id": 358272,
                            "lemma": [
                                "zu+die",
                                "Beispiel"
                            ],
                            "settings": {},
                            "tag": [
                                "APPRART",
                                "NN"
                            ],
                            "wordcount": 2,
                            "words": [
                                "zum",
                                "Beispiel"
                            ]
                        }
[...]

PR #2794, opened 2024-05-10
Add abbreviation count to HIX feedback

MizukiTemma:
### Short description

This PR adds the number of abbreviations used in the content text to the HIX feedback

### Proposed changes

- Use jmespath to access deeply nested dictionary values in one step
- Add abbreviations to `textlab_api_feedback_categories`

### Side effects

- None? 👀  

### Resolved issues

Fixes: #2778 


seluianova (changes requested):
That's a cool and lean solution :)
But I can see one issue:
If the text contains 2 (or more) abbreviations, the widget will display it as "Text contains 1 abbreviation".

Sorry if my issue description was confusing, because I only described the calculation logic for previously added categories. But for "data terms" it seems to be different and requires analysis.

**UPD**. I assume the "results" property is an array whose size is the number of **unique** abbreviations.
At the same time, if the same abbreviation occurs twice, it can be "read" from the "length" property.
So, it seems we need to count the "results" size and also the "results.length" size.

MizukiTemma:
@seluianova Thank you for review 😸 

Can you tell me with what text the feedback showed "1 abbreviation" despite multiple of them?

I checked it in the test enviroment, and it says "Der Text enthält 2 Abkürzungen" for the page "Ausländerbehörde" where "FAQ" and "Z.B" are caught 🤔

MizukiTemma:
hmm 🤔 Textlab doesn't count "TÜV" and "BAföG" as abbreviation while it catches "FAQ" ...

seluianova:
> Can you tell me with what text the feedback showed "1 abbreviation" despite multiple of them?
> I checked it in the test enviroment, and it says "Der Text enthält 2 Abkürzungen" for the page "Ausländerbehörde" where "FAQ" and "Z.B" are caught

@MizukiTemma ah, I actually only checked the repeated abbreviation when I was writing the first comment. So the problem is only here:

> At the same time, if the same abbreviation occurs twice, it will be reflected in the "result.length" property

If you use "Z.B." 2 times, it will count as 1, while textlab.online shows 2 in this case.
I would think, that we should provide the same result as textlab.online. Or would you object to that? :)

MizukiTemma:
@seluianova 
I've adjusted for abbreviations that appear more than once.
Now not the arrays but the number of abbreviation itself is directly delivered to the typescript part. I hope this doesn't harm anything 😅

integreat_cms/textlab_api/textlab_api_client.py
```diff
@@ -77,10 +79,23 @@ def benchmark(
         path = f"/benchmark/{text_type}"
         response = self.post_request(path, data, self.token)
 
-        feedback_details = [
-            {"category": cms_name, "result": response.get(textlab_name, [])}
-            for textlab_name, cms_name in textlab_api_feedback_categories.items()
-        ]
+        feedback_details = []
+
+        for textlab_name, cms_name in textlab_api_feedback_categories.items():
```
seluianova, line 84: Optional idea from readability perspective:
Maybe it would be better to create a separate mapping for data-terms categories? (with the only one category so far)
And iterate over them sequentially?
I have a feeling that in that case it would be easier to follow/read the code.

seluianova: I tried to make something up and came up with this:

```
        textlab_api_feedback_dataterms = {
            "1289": "abbreviations",
        }

        feedback_details = [
            {"category": cms_name, "result": response.get(textlab_name, [])}
            for textlab_name, cms_name in textlab_api_feedback_categories.items()
        ]
        
        data_terms = response.get('dataTerms', {})

        for dataterm_id, cms_name in textlab_api_feedback_dataterms.items():
            results = data_terms.get(dataterm_id, {}).get('result', [])
            total_length = sum(len(item.get('length', '')) for item in results)
            feedback_details.append({
                "category": cms_name,
                "result": total_length,
        })
```

Not sure if it's any better though.
So just ignore this comment if you want.

integreat_cms/textlab_api/textlab_api_client.py
```diff
@@ -77,10 +79,23 @@ def benchmark(
         path = f"/benchmark/{text_type}"
         response = self.post_request(path, data, self.token)
 
-        feedback_details = [
-            {"category": cms_name, "result": response.get(textlab_name, [])}
-            for textlab_name, cms_name in textlab_api_feedback_categories.items()
-        ]
+        feedback_details = []
```
seluianova, line 82: Maybe worth moving this to a separate function?

seluianova (approved):
I've re-tested it again and found no more issues.
Added couple suggestions, but in general looks good to me.
Thanks for the applied fixes :)

MizukiTemma:
> I've re-tested it again and found no more issues. Added couple suggestions, but in general looks good to me. Thanks for the applied fixes :)

Thank you so much for review and improvement ideas 😸 They are applied too.

integreat_cms/textlab_api/textlab_api_client.py
```diff
@@ -25,6 +27,7 @@
     "countInfinitiveConstructions": "infinitive-constructions",
     "countNominalStyle": "nominal-sentences",
     "countFutureTenseInSentence": "future-tense-sentences",
+    'dataTerms."1289".result': "abbreviations",
```
seluianova, line 30: I think this mapping could also be moved to utils, as it's not used here anymore

JoeyStk (approved):
To me this also looks good. No comments from my side :)

PR #2798, opened 2024-05-13
Fix info message after saving page content twice

### LINKED ISSUE #2673: Page with content reports "successfully updated" instead of "no changes"
### Describe the Bug

Because we try to normalize the page content before saving it to the database in a format different to the one that tinymce exports, django always reports that the content field changed (except on empty pages), even when just clicking "Update".

This should largely be fixed by #2668

### Steps to Reproduce

1. Go to a page with >1 lines of content
2. Compare the source in tinymce with the content in the database (e.g. via django admin) → They are different (e.g. `<br>` vs. `\r\n`)
3. Click on *Update*

### Expected Behavior

The page should show *No changes detected, but date refreshed*

### Actual Behavior

The page shows <i>Page "</i>&lt;page name&gt;<i>" was successfully updated</i>

deen13:
### Short description

This pull request addresses the issue of Tinymce adding whitespace to the content by setting indent: false in the Tinymce configuration. Currently, the added whitespace by Tinymce causes our backend to incorrectly determine whether the content has been edited or not. This is evident when saving a newly created page for the second time, as it erroneously indicates that changes have been saved even when there are none.

Further Explanation: https://stackoverflow.com/a/39489073
Demonstration: https://github.com/digitalfabrik/integreat-cms/assets/13850172/e40e868e-87a5-4972-818d-378c8014aae5

### Proposed changes
- Set indent: false in the Tinymce configuration to prevent the addition of whitespace to the content.

### Resolved issues

Fixes: #2673

### DIFF
--- a/integreat_cms/static/src/js/forms/tinymce-init.ts
+++ b/integreat_cms/static/src/js/forms/tinymce-init.ts
@@ -100,6 +100,7 @@ window.addEventListener("load", () => {
             link_default_protocol: "https",
             target_list: false,
             default_link_target: "",
+            indent: false,
             document_base_url: tinymceConfig.getAttribute("data-webapp-url"),
             relative_urls: false,
             remove_script_host: false,


seluianova (commented):
It solves the issue ✅
But there is one side effect:
We lose the "source code" formatting of the content.
This can be checked in the editor via View --> Source code.

However I don't know if anyone uses this menu item, and may be upset at the loss of formatting. 
Probably not, but maybe someone else in the team has a clearer opinion on this?

deen13:
> It solves the issue ✅
> But there is one side effect:
> We lose the "source code" formatting of the content.
> This can be checked in the editor via View --> Source code.
> 
> However I don't know if anyone uses this menu item, and may be upset at the loss of formatting.
> Probably not, but maybe someone else in the team has a clearer opinion on this?

Hey @osmers, do you have any insights on how technically savvy the users are?

osmers:
@deen13 one question - Tory writes that we lose the source code formatting - I assume that does not count for the whole text but only for a section of it?
In general, I would say that the source code menu is not used that often by most of the municipalities, since they have no need to use it.

deen13:
We have reached a consensus that the time and resources required to address it do not outweigh the expected benefits. We will continue to monitor the situation and re-evaluate the priority of this issue in the future if necessary.

PR #2802, opened 2024-05-15
GVZ: search region by name

svenseeberg:
### Short description

If a region name is available but no AGS, the AGS should be determined with the GVZ API.

### Proposed changes
- Search for AGS if region name is available but AGS is not
- Only set region name from API if it does not override an existing name (most likely entered manually)

### Side effects
- N/A

### Resolved issues
Fixes: https://chat.tuerantuer.org/digitalfabrik/threads/tczb6bt38pg9xy8i3f5z7sfrka


charludo (approved):
LGTM

PR #2803, opened 2024-05-15
Fix nonexistant language MT provider leading to error in tests

charludo:
### Short description

For a while now, I've been getting test failures relating to the MT management view. I'm not entirely sure *why* those happen (esp. since they do not *always* happen, and apparently never happens on CircleCI), but if they do happen, it's due to a key error when querying for the MT provider for a given language.

If someone has a idea how to fix the root cause of this, I am all ears 😄 

### Proposed changes

- instead of directly accessing the dict value via `[]`, `get()` it with fallback `None`

### Side effects

- none 

### Resolved issues

Fixes: n/a


MizukiTemma (approved):
hmmm 🤔 I could reproduce the problem, but me neither have any idea why...... 😅 

Though thank you for the fix 👍

seluianova (approved):
I haven’t seen the issue, but if this helps, I think it’s good to go 👍

PR #2804, opened 2024-05-15
pre-commit: Remove prettier because the repo is now archived

cclauss:
https://github.com/pre-commit/mirrors-prettier
* prettier/prettier#16229

One of the other options could be used:
* https://github.com/prettier/prettier/blob/main/docs/precommit.md

```diff
  - repo: https://github.com/pre-commit/mirrors-prettier
    rev: v4.0.0-alpha.8
    hooks:
      - id: prettier
        additional_dependencies:
          - prettier@3.2.5
```

- - -

Unrelated CodeQL warnings: https://github.blog/changelog/2024-01-23-codeql-2-16-python-dependency-installation-disabled-new-queries-and-bug-fixes/

### Proposed changes

- 
- 

### Side effects

- 
- 

### Resolved issues

Fixes: #

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,12 +49,6 @@ repos:
           - eslint@8.56.0
           - "@typescript-eslint/eslint-plugin@6.21.0"
           - "eslint-plugin-prefer-arrow@1.2.3"
-  - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v4.0.0-alpha.8
-    hooks:
-      - id: prettier
-        additional_dependencies:
-          - prettier@3.2.5
   - repo: local
     hooks:
       - id: translations


charludo:
Thanks for the PR! I'd argue Option 4 would probably make the most sense, given that the install script makes sure prettier is available anyways?

cclauss:
I am a bit out of my depth with those options.  If you know how to implement an appropriate solution then we can close this PR with unmerged commits.

charludo:
> I am a bit out of my depth with those options. If you know how to implement an appropriate solution then we can close this PR with unmerged commits.

Is it alright with you if I push a commit with the script to this branch?

cclauss:
Sure!

charludo:
Sorry for taking so long to get back to you!

I just realized I can't easily commit to your fork, so let's do it like this instead:

This works great (unmodified from the link you provided; put in `.git/hooks/pre-commit` and `chmod +x`).
```bash
#!/bin/sh
FILES=$(git diff --cached --name-only --diff-filter=ACMR | sed 's| |\\ |g')
[ -z "$FILES" ] && exit 0

# Prettify all selected files
echo "$FILES" | xargs ./node_modules/.bin/prettier --ignore-unknown --write

# Add back the modified/prettified files to staging
echo "$FILES" | xargs git add

exit 0
```

I think the previous behavior was to write, but not re-add the changed files (delete the line with `echo "$FILES" | xargs git add`). This makes the hook fail if it has to re-format some file(s). I think this is the better behaviour, since this way you can manually check the changes before commiting.

All that's left is then to include 
```yaml
 - repo: local
    hooks:
      - id: prettier
        name: prettier (custom script)
        entry: ./.git/hooks/pre-commit
        language: script
        pass_filenames: false
```
in the `.pre-commit-config.yaml` file.

timobrembeck:
@charludo do you think we could include this functionality into our [./tools/prettier.sh](https://github.com/digitalfabrik/integreat-cms/blob/204a66f9da45d906a7471baa8fb94ca580718563/tools/prettier.sh) instead of the `.git/hooks/pre-commit` which is not part of the version control itself? :thinking:

charludo:
> @charludo do you think we could include this functionality into our [./tools/prettier.sh](https://github.com/digitalfabrik/integreat-cms/blob/204a66f9da45d906a7471baa8fb94ca580718563/tools/prettier.sh) instead of the `.git/hooks/pre-commit` which is not part of the version control itself? 🤔

Oops, yes, absolutely. 

Actually, we could probably switch all of the pre-commit hooks for which corresponding devtools exist over to this approach...? Not sure if there are downsides, but it would ensure versions between devtool/pre-commit are always consistent, plus we could finally add a pre-commit hook for pylint

charludo (approved):
> Actually, we could probably switch all of the pre-commit hooks for which corresponding devtools exist over to this approach

I'll open a new PR for this, so I think this one is good to go! Thank you @cclauss

timobrembeck (approved):
Great, thanks! :+1:

(and sorry for the long delay between opening the PR and merging it :innocent:)

Issue #2025, opened 2023-01-24
Improve FCM API Error handling


svenseeberg:
### Motivation
Timeouts of the FCM API trigger an error cascade. The handling of this information should be improved.

### Proposed Solution
Timeout errors should be catched and transformed into a clear front end message. The status of the message should be set back to draft, if sending failed.

### Alternatives
Live with (expected) internal server errors.

### Additional Context
```
Jan 24 11:18:47 ERROR   django.request - 500 Internal Server Error: /testumgebung/push-notifications/de/397/edit/
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/urllib3/connectionpool.py", line 449, in _make_request
    six.raise_from(e, None)
  File "<string>", line 3, in raise_from
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/urllib3/connectionpool.py", line 444, in _make_request
    httplib_response = conn.getresponse()
  File "/usr/lib/python3.9/http/client.py", line 1347, in getresponse
    response.begin()
  File "/usr/lib/python3.9/http/client.py", line 307, in begin
    version, status, reason = self._read_status()
  File "/usr/lib/python3.9/http/client.py", line 268, in _read_status
    line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
  File "/usr/lib/python3.9/socket.py", line 704, in readinto
    return self._sock.recv_into(b)
  File "/usr/lib/python3.9/ssl.py", line 1241, in recv_into
    return self.read(nbytes, buffer)
  File "/usr/lib/python3.9/ssl.py", line 1099, in read
    return self._sslobj.read(len, buffer)
socket.timeout: The read operation timed out

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/requests/adapters.py", line 489, in send
    resp = conn.urlopen(
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/urllib3/connectionpool.py", line 787, in urlopen
    retries = retries.increment(
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/urllib3/util/retry.py", line 550, in increment
    raise six.reraise(type(error), error, _stacktrace)
  File "/opt/integreat-cms/.venv/lib/python3.9/site-packages/urllib3/packages/six.py", line 770, in reraise
    raise value
[...]

Issue #2790, opened 2024-05-07
`JSONDecodeError` when FCM_URL is wrong


MizukiTemma:
### Describe the Bug

It's possible now to specify `FCM_URL`, and an error occurs when a wrong URL is given there.

### Steps to Reproduce

1. Go to `integreat-cms.ini`
2. Add something wrong as `FCM_URL` (for example, https://integreat-app.de/)
3. Go to a region, create and send a push notification
4. See the error

### Expected Behavior

An user friendly message is shown

### Actual Behavior

Unfriendly yellow error wall

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

MizukiTemma:
`MissingSchema` will be raised if for example "asdf" is given as URL.

MizukiTemma:
Solved as part of #2805

PR #2805, opened 2024-05-16
Catch and handle errors of FCM

MizukiTemma:
### Short description

This PR add an error handling for invalid FCM URL.

**update**
This takes care of timeout exception too. I've totally forgotten that those two issues are related each other.

### Proposed changes

- Catch errors and show an error message to users 
- The error message is left universal (not specific to any error type) because many types are possible. 

### Side effects

- The rendering after failing of a push notification is partially broken. 
- A question regarding this rendering: don't we want to redirect not no the create view but to the edit view becuase the push notification is saved in this case?
- I'll suggest to fix the rendering in a separate PR/issue.

### Resolved issues

Fixes: #2025 , #2790 

### DIFF
--- a/integreat_cms/firebase_api/firebase_api_client.py
+++ b/integreat_cms/firebase_api/firebase_api_client.py
@@ -8,6 +8,7 @@ import requests
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from google.oauth2 import service_account
+from requests.exceptions import RequestException
 
 from ..cms.constants import push_notifications as pnt_const
 from ..cms.forms.push_notifications.push_notification_translation_form import (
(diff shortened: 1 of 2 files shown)


david-venhoff (changes requested):
Thanks, this fixes both linked issues. 
I think, as part of the clean code initiative, we should avoid introducing new `pylint: disable=` comments. Maybe one option would be to handle  the errors directly at `self.send_pn`

MizukiTemma:
@david-venhoff 
That's a nice idea 😸 Applied!

david-venhoff (approved):
Nice :)

ulliholtgrave (approved):
Thanks for adding this extra handling :+1:

Issue #765, opened 2021-04-10
Tests and more Tests


svenseeberg:
### Motivation
Create a stable product. Write unit tests, end to end tests and so on.

osmers:
Closing issue as completed - ideas from above are saved and can be converted into issues if deemed necessary. Issues related to this topic have the label topic:tests

PR #2810, opened 2024-05-22
Add vitest for frontend testing

deen13:
### Short Description

This pull request configures vitest and adds a simple test as suggested in my review of [deps/upgrade_webauthn_2_0](https://github.com/digitalfabrik/integreat-cms/pull/2716).

### Proposed Changes

- Added a test task to `package.json`
- Included required vitest dependency
- Implemented a simple test for URL and base64 encoding
- Configure vitest to ignore python dependencies
- Configure import/resolver to resolve imports correctly
- Add job to circleci workflow

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -210,6 +210,17 @@ jobs:
       - run:
           name: Check formatting of CSS & JS files
           command: npx prettier --check .
+  vitest:
+    docker:
+      - image: "cimg/node:lts"
+    resource_class: small
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - run:
+          name: Run frontend tests
+          command: npm run test
   eslint:
     docker:
       - image: "cimg/node:lts"
@@ -607,6 +618,9 @@ workflows:
             branches:
               ignore: main
       - prettier:
+          requires:
+            - npm-install
+      - vitest:
           requires:
             - npm-install
       - webpack:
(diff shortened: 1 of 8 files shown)


JoeyStk (commented):
Thank you a lot :rocket: I think this is a good step forward.

1. Small note: Once this PR is merged you can check the checkbox in our [meta issue for tests](https://github.com/digitalfabrik/integreat-cms/issues/765). 
2. I personally like vitest more than jest for our usecase. I think Jest has a lot of features that we don't need as we are web only and therefore I rather prefer the speed of vitest.
3. Lastly the reason why I don't approve yet. Do we need a dev tool for this? I'm kind of lost in the whole discussion about pre-commits and dev commands :confused: Once this question is answered I will approve this PR as it looks good to me so far :)

Special thanks for the dummy test. These things always help me to get started with a new feature :) In the mid-future we can remove it again, but for the beginning it's much appreciated (at least on my side) :)

deen13:
@JoeyStk I've added a devtool and a pre-commit hook that calls it. ☺️

JoeyStk (approved):
I'm not an expert on this matter, but would like to move this forward.
To me this PR works. I didn't face issues when installing this and running the dev-tool. Thank you very much :)

ulliholtgrave:
Thank you very much :+1: This really is an improvement! Could you squash the first two commits into one? Then, this is good to go :)

PeterNerlich (approved):

Issue #2707, opened 2024-03-20
Add `available_languages` to news endpoint


steffenkleinle:
### Motivation

In order to allow users in the apps to switch the language of news, we need to get information about the other languages from the CMS.

### Proposed Solution

Deliver the ids/slugs of the news in all other languages the same way its done for pages, events and pois.
Use an `available_languages` property.

### Alternatives

None.

### Additional Context

App issue: https://github.com/digitalfabrik/integreat-app/issues/2732

### Design Requirements

None.

PR #2811, opened 2024-05-22
Add available_languages to news endpoint

lunars97:
### Short description

Currently it is not possible to switch languages of the news. Available languages property should be added to push_notifications endpoint.

### Proposed changes

- Add `available languages` to push notifications endpoint
- Make proper changes in push_notification and push_notification_translation models

### Side effects

-None?
### Resolved issues

Fixes: #2707 

### DIFF
--- a/docs/src/api-docs.rst
+++ b/docs/src/api-docs.rst
@@ -626,6 +626,12 @@ RESPONSE
       "timestamp": String,     // Deprecated field
       "last_updated": String,  // The date&time when the push notification was last updated
       "channel": String,       // The channel the push notification was sent to (e.g. "News")
+      "available_languages": [           // The available languages of the push notification
+            "<language_slug>": {
+               "id": Number | null,           // The id of the translation
+            },
+            ...
+         ]
    }
 
 
--- a/integreat_cms/api/v3/push_notifications.py
+++ b/integreat_cms/api/v3/push_notifications.py
@@ -57,6 +57,10 @@ def transform_notification(pnt: PushNotificationTranslation) -> dict[str, Any]:
     :param pnt: A push notification translation
     :return: data necessary for API
     """
+    available_languages_dict = {
+        translation.language.slug: {"id": translation.id}
+        for translation in pnt.push_notification.translations.all()
+    }
     return {
         "id": str(pnt.pk),
         "title": pnt.get_title(),
@@ -64,4 +68,5 @@ def transform_notification(pnt: PushNotificationTranslation) -> dict[str, Any]:
         "timestamp": pnt.last_updated,  # deprecated field in the future
         "last_updated": timezone.localtime(pnt.last_updated),
         "channel": pnt.push_notification.channel,
+        "available_languages": available_languages_dict,
     }


lunars97:
I got stuck in model adjustment. Any assistance would be highly valued and appreciated :pray: :pray:

cclauss:
Do you have [`pre-commit`](https://pre-commit.com) installed locally?  It should be autofix the errors from black, isort, and ruff.

lunars97:
> Do you have [`pre-commit`](https://pre-commit.com) installed locally? It should be autofix the errors from black, isort, and ruff.

I think I haven't installed it yet. Thanks for your advice and review :+1:

charludo:
Additionally, there's plugins available for all of those vor VSCode which should just pick up our project's style configurations and apply them automatically when you save a file! :)

charludo:
The prompt to enter a default value for "region" you get when loading the test data is due to `AbstractContentModel` (from which you are letting `PushNotification` inherit) having a field `region` which is not allowed to be null/empty. Because `PushNotification` has not had that field before, no `region` values for it exist.

Instead, `PushNotification` has a field `regions` (apparently a PN can belong to multiple regions?)

This is not trivial to work around. As far as I understand it, you would have to re-implemented most of the functionality from `AbstractContentTranslation` and `AbstractBaseContent`, or find a sensible default value for `PushNotification.region`. But since that region is then used to determine available languages etc. for the PN, it would somehow have to aggregate all the PN's `regions` languages.

I think you should bring this up in the next weekly. Maybe someone else who knows more about this part of the CMS has a better idea than what I just described.

lunars97:
> The prompt to enter a default value for "region" you get when loading the test data is due to `AbstractContentModel` (from which you are letting `PushNotification` inherit) having a field `region` which is not allowed to be null/empty. Because `PushNotification` has not had that field before, no `region` values for it exist.
> 
> Instead, `PushNotification` has a field `regions` (apparently a PN can belong to multiple regions?)
> 
> This is not trivial to work around. As far as I understand it, you would have to re-implemented most of the functionality from `AbstractContentTranslation` and `AbstractBaseContent`, or find a sensible default value for `PushNotification.region`. But since that region is then used to determine available languages etc. for the PN, it would somehow have to aggregate all the PN's `regions` languages.
> 
> I think you should bring this up in the next weekly. Maybe someone else who knows more about this part of the CMS has a better idea than what I just described.

Many many thanks for having a look at it :heart:

david-venhoff:
It is probably best if `PushNotificationTranslation` does not inherit from `AbstractContentTranslation`, because there are many differences, as @charludo pointed out.
Instead, you could do something like this in the api endpoint:
```python
def transform_notification(pnt: PushNotificationTranslation) -> dict[str, Any]:
    """
    Function to create a JSON from a single push notification translation Object.

    :param pnt: A push notification translation
    :return: data necessary for API
    """
    available_languages_dict = {translation.language.slug: {"id": translation.id} for translation in pnt.push_notification.translations.all()}
    return {
        "id": str(pnt.pk),
        "title": pnt.get_title(),
        "message": pnt.get_text(),
        "timestamp": pnt.last_updated,  # deprecated field in the future
        "last_updated": timezone.localtime(pnt.last_updated),
        "channel": pnt.push_notification.channel,
        "available_languages": available_languages_dict,
    }

```

deen13 (approved):
Looks good to me, aside from the change David already suggested! 🥳

PR #2821, opened 2024-06-03
Start docker container without -t flag on macOS

deen13:
### Short description

This pull request updates the `wait_for_docker_container` function in our shell script to fix the waiting mechanism on macOS, which currently hangs forever, as demonstrated in the attached video.

### Proposed changes

- Removed the docker run flags, as the waiting mechanism works without them.
- Added echo statements to provide feedback during the waiting period.

### Testing
The script was tested on macOS to ensure it does not hang and correctly waits for the Docker container to be ready. It'd be nice if someone can test the script on a Linux environment to verify that it maintains compatibility.

### DIFF
--- a/tools/_functions.sh
+++ b/tools/_functions.sh
@@ -317,14 +317,14 @@ function wait_for_docker_container {
         DOCKER_EXEC_FLAGS="-it"
     fi
 
-    echo "Waiting for Docker container ${DOCKER_CONTAINER_NAME} to be ready..."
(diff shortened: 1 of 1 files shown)


PeterNerlich (approved):
Works on Linux! *(using Podman, which claims to be full drop-in replacement of docker)*

david-venhoff (approved):
I've had the same problem on my linux laptop for a few weeks and used pretty much the same solution. So LGTM after you sqash the commits!

PR #2828, opened 2024-06-07
Define pre-commit hooks in devtool scripts

charludo:
### Short description

This is a followup to the discussion in #2804. This PR swaps out the existing pre-commit hooks for calls to our existing devtools. @cclauss I cannot request you as a reviewer, but you're more than welcome to also take a look!!

### Proposed changes

- a new helper function `run_as_precommit` is called at the start of each relevant devtool
- if the tool was called with the `--as-precommit` flag, the tool is only run against staged files
- `.pre-commit-config` uses the local tools and sets the `--as-precommit` flag
- this *should* be a drop-in replacement for the current pre-commit hooks
- additionally, `pylint` can now be run as a pre-commit hook 🎉 

### Side effects

- possibly the behavior doesn't match 100%? It's a bit hard to tell, honestly.
- no more mismatches between pre-commit / installed tool versions!

### Resolved issues

Fixes: n/a

### DIFF
--- a/tools/_functions.sh
+++ b/tools/_functions.sh
@@ -362,7 +362,7 @@ function cleanup_docker_container {
 }
 
 function ensure_webpack_bundle_exists {
-    if [ ! -d "${PACKAGE_DIR}/static/dist/" ] || [ ! "$(ls -A "${PACKAGE_DIR}"/static/dist/ 2>/dev/null)" ]; then
+    if [ ! -d "${PACKAGE_DIR}/static/dist/" ] || [ ! "$(ls -A "${PACKAGE_DIR}"/static/dist/ 2> /dev/null)" ]; then
         echo "Building webpack bundle..." | print_info
         npm run build > /dev/null
     fi
@@ -486,4 +486,24 @@ function join_by {
     if shift 2; then
         printf %s "$f" "${@/#/$d}"
     fi
+}
+
+# This function checks if the flag "--as-precommit" was set, and if so,
+# runs the specified pre-commit hook against all changed files
+function run_as_precommit {
+    local command="$1"
+    shift
+
+    for arg in "$@"; do
+        if [[ "$arg" == "--as-precommit" ]]; then
+            shift
+            [ "$#" -eq 0 ] && exit 0
+
+            for file in "$@"; do
+                eval "$command \"$file\""
+            done
+
+            exit 0
+        fi
+    done
 }
--- a/tools/djlint.sh
+++ b/tools/djlint.sh
@@ -8,6 +8,9 @@ source "$(dirname "${BASH_SOURCE[0]}")/_functions.sh"
 
 require_installed
 
+# Run djlint as a pre-commit hook
+run_as_precommit "djlint --profile=django -e=html --reformat --quiet --lint" "$@"
+
 # Run djlint
 echo "Starting code formatting with djlint..." | print_info
 djlint --reformat --quiet --lint "${PACKAGE_DIR}"
--- a/tools/eslint.sh
+++ b/tools/eslint.sh
@@ -6,6 +6,9 @@
 # shellcheck source=./tools/_functions.sh
 source "$(dirname "${BASH_SOURCE[0]}")/_functions.sh"
 
+# Run eslint as a pre-commit hook
+run_as_precommit "npx eslint" "$@"
+
 # Run prettier
 echo "Starting code linting with eslint..." | print_info
 npx eslint .
--- a/tools/prettier.sh
+++ b/tools/prettier.sh
@@ -6,6 +6,9 @@
 # shellcheck source=./tools/_functions.sh
 source "$(dirname "${BASH_SOURCE[0]}")/_functions.sh"
 
+# Run prettier as a pre-commit hook
+run_as_precommit "npx prettier --write" "$@"
+
 # Run prettier
 echo "Starting code formatting with prettier..." | print_info
 npx prettier --write .
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,35 +16,56 @@ repos:
         exclude: .github/PULL_REQUEST_TEMPLATE.md|tests/cms/views/dashboard/expected_output/chat.html
       - id: no-commit-to-branch
         args: [--branch, main, --branch, develop]
-  - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.4.7
+  - repo: local
     hooks:
       - id: ruff
+        name: ruff (custom script)
+        entry: ./tools/ruff.sh
+        args: ["--as-precommit"]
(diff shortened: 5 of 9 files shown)


tools/eslint.sh
```diff
@@ -6,6 +6,9 @@
 # shellcheck source=./tools/_functions.sh
 source "$(dirname "${BASH_SOURCE[0]}")/_functions.sh"
 
+# Run eslint as a pre-commit hook
+run_as_precommit "./node_modules/.bin/eslint" "$@"
```
david-venhoff, line 10:
```suggestion
run_as_precommit "npx eslint" "$@"
```
Is there a difference when calling it with the exact path compared to using npx? If not I'd prefer to call it like in line 14 for consistency

tools/prettier.sh
```diff
@@ -6,6 +6,9 @@
 # shellcheck source=./tools/_functions.sh
 source "$(dirname "${BASH_SOURCE[0]}")/_functions.sh"
 
+# Run prettier as a pre-commit hook
+run_as_precommit "./node_modules/.bin/prettier --ignore-unknown --write" "$@"
```
david-venhoff, line 10:
```suggestion
run_as_precommit "npx prettier --ignore-unknown --write" "$@"
```

tools/_functions.sh
```diff
@@ -473,3 +473,24 @@ function join_by {
...
+
+# This function checks if the flag "--as-precommit" was set, and if so,
+# runs the specified pre-commit hook against all changed files
+function run_as_precommit {
+    local command="$1"
+    shift
+
+    for arg in "$@"; do
+        if [[ "$arg" == "--as-precommit" ]]; then
+            FILES=$(git diff --cached --name-only --diff-filter=ACMR | sed 's| |\\ |g')
```
david-venhoff, line 485: I don't fully understand why we need the `| sed 's| |\\ |g'` part, but this escaping causes some problems when we have files with spaces:
image

Without this part, the file can be found again

david-venhoff (commented):
Thanks. Really nice that we can run pylint now :partying_face: 

The python devtools cannot really handle being called with a path to a non-python file.
When I tested this with a changed .ts file, those following tools crashed with a syntax error, because they tried to interpret the code as python: ruff, mypy and pylint.

I also get an error from eslint, but that looks like #2848.

tools/eslint.sh
deen13: Yes, there can be a difference. If node_modules is not in place, npx prettier will fetch and run the latest version of Prettier from the npm registry. This is a rare scenario, but if it happens, it could cause inconsistencies and be confusing.

Screenshot 2024-07-15 at 18 23 40

tools/djlint.sh
```diff
@@ -6,6 +6,9 @@
 # shellcheck source=./tools/_functions.sh
 source "$(dirname "${BASH_SOURCE[0]}")/_functions.sh"
 
+# Run djlint as a pre-commit hook
+run_as_precommit "djlint --profile=django -e=html --reformat --quiet --lint" "$@"
+
 require_installed
```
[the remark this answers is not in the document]
deen13: +1 not only for consistency but also because the pre-commit hooks might fail if the user is currently not within the virtual environment.

deen13 (approved):
Looks good to me, except for the points that @david-venhoff already mentioned. Great work, @charludo, and good catches, @david-venhoff! 😄

tools/djlint.sh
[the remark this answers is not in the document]
deen13: +1 not only for consistency but also because the pre-commit hooks might fail if the user is currently not within the virtual environment.

tools/eslint.sh
cclauss: Would pinning the version like `npx eslint@v9.7.0` remove those inconsistencies?

tools/_functions.sh
charludo: I think this is a mistake in the original script, you are right. Pretty sure it should just be `| sed 's| |\ |g'`.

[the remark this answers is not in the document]
charludo: No, you are right. `pass_filenames` with `types_or` makes way more sense. This part of the PR is adapted from [here](https://github.com/prettier/prettier/blob/main/docs/precommit.md), and I am not sure why this is not also the suggestion there.

In any case. Thank you for the suggestion :D

tools/_functions.sh
```diff
@@ -473,3 +473,24 @@ function join_by {
...
+            [ -z "$FILES" ] && exit 0
+
+            while IFS= read -r file; do
+                eval "$command \"$file\""
```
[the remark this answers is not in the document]
charludo: SOme of the commands are happy to take multiple files at once, some aren't. But since none of them protest being given just a single file.... 🤷🏼‍♀️ Lowest common denominator 😄

charludo:
Thank you for your thorough testing and suggestions, @david-venhoff! I hope I was able to address all of your points. (At least no more crashes with ts files... 😁)

tools/_functions.sh
```diff
@@ -473,3 +473,24 @@ function join_by {
...
+                git add "$file"
```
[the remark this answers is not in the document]
david-venhoff: I am a bit worried that the changes may not be noticed by the committer. For testing, I have modified a ts file and added some code with bad indentation, and after writing `git commit`, the commit message editor opened and the indentation in the file was fixed, without informing me about it:
image

Personally, I would prefer to not automatically add the files back, so that I can know what was changed by pre-commit. Although I would be fine with it if there was at least some kind of feedback that a file has been changed.

david-venhoff (approved):
Very cool! Apart from my one comment, this looks good to me.

cclauss:
> I am a bit worried that the changes may not be noticed by the committer.

If you have run `pre-commit install` in your local directory then `git commit` will fail.  This makes it kinda hard to miss.

Of course, you can `git commit --no-verify` to get the commit to succeed anyway but then the automated test will fail on GitHub.

david-venhoff:
Personally, I really like to know when something changed so I can have a final quick look at it before committing.
But I also don't want to insist on this if it turns out that everybody else doesn't really care about that :sweat_smile:
Maybe it would make sense to talk about this in the team call?

charludo:
> Personally, I really like to know when something changed so I can have a final quick look at it before committing. But I also don't want to insist on this if it turns out that everybody else doesn't really care about that 😅 Maybe it would make sense to talk about this in the team call?

That's completely fair. I guess there's also an argument to be made for staying with the current behavior. Let's do a poll in the meeting tomorrow, but unless an overwhelming majority really WANTS the change, let's go with your suggestion and the old behavior.

charludo:
This should be ready to be merged now - I'll leave it open for a couple of days though, in case someone wants to test it out / has objections

david-venhoff (approved):
:tada:

Issue #2724, opened 2024-04-02
HTTP Error 500 from KeyError when image lacks src attribute


PeterNerlich:
### Describe the Bug

When checking the `src` value for images in pages that were just saved, the server assumes that this attribute exists. This is not necessarily the case.

### Steps to Reproduce

1. Go to any page of any region or go to create a new one
2. Change the editor from WYSIWYG to HTML by clicking on the *View* menu and select *Source view*
3. Paste in `<img>`
4. Click on *Update*
5. See error

### Expected Behavior

The page form view loads again, and maybe displays a hint that there are empty images

### Actual Behavior

The server encounters an uncaught KeyError

### Additional Information

<details>
  <summary>Traceback</summary>

```
Apr 02 16:03:53 ERROR django.request - 500 Internal Server Error: /augsburg/pages/de/1/edit/
Traceback (most recent call last):
  File "integreat-cms/.venv/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "integreat-cms/.venv/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "integreat-cms/.venv/lib/python3.11/site-packages/django/views/generic/base.py", line 104, in view
    return self.dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "integreat-cms/.venv/lib/python3.11/site-packages/django/utils/decorators.py", line 46, in _wrapper
    return bound_method(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "integreat-cms/.venv/lib/python3.11/site-packages/django/contrib/auth/decorators.py", line 23, in _wrapper_view
    return view_func(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "integreat-cms/.venv/lib/python3.11/site-packages/django/views/generic/base.py", line 143, in dispatch
    return handler(request, *args, **kwargs)
[...]

PR #2829, opened 2024-06-07
Prevent interruption by KeyError with empty img tag

MizukiTemma:
### Short description

This PR prevents interruption of content save caused by empty img tag.

### Proposed changes

- Add `try` and `except`

### Side effects

- None?
- No message for users is implemented, as it won't be directly added in the form together with the suggested change and I guess this error case is pretty rare and it'ts not worth additional effort 😅 

### Resolved issues

Fixes: #2724 


integreat_cms/cms/forms/custom_content_model_form.py
```diff
@@ -174,22 +174,27 @@ def clean_content(self) -> str:
...
+                    Q(file=relative_url) | Q(thumbnail=relative_url)
+                ).first()
+                # Replace alternative text
+                if media_file and media_file.alt_text:
+                    self.logger.debug(
+                        "Image alt text replaced: %r", media_file.alt_text
+                    )
+                    image.attrib["alt"] = media_file.alt_text
+            except KeyError:
+                self.logger.warning("Empty img tag was found.")
```
david-venhoff, line 197:
```suggestion
            if src := image.attrib.get("src"):
                self.logger.debug(
                    "Image tag found in content (src: %s)", src
                )
                # Remove host
                relative_url = urlparse(src).path
                # Remove media url prefix if exists
                if relative_url.startswith(settings.MEDIA_URL):
                    relative_url = relative_url[len(settings.MEDIA_URL) :]
                # Check whether media file exists in database
                media_file = MediaFile.objects.filter(
                    Q(file=relative_url) | Q(thumbnail=relative_url)
                ).first()
                # Replace alternative text
                if media_file and media_file.alt_text:
                    self.logger.debug("Image alt text replaced: %r", media_file.alt_text)
                    image.attrib["alt"] = media_file.alt_text
```

david-venhoff (approved):
Thanks, this fixes the issue. The try just seems a bit broad to me, and we could avoid it by using similar code to the other checks in this function (by checking if the images has the required src attribute)
What do you think about this?

integreat_cms/cms/forms/custom_content_model_form.py
```diff
@@ -98,7 +98,7 @@ def clean(self) -> dict[str, Any]:
 
         return cleaned_data
 
-    # pylint: disable=too-many-branches
```
JoeyStk, line 101: 96 lines for one function, two disabled pylint rules and a function that does a lot of things are not really great in my opinion. Could you please open up an issue to refactor this function? Maybe for a milestone in the near future? :)

JoeyStk (approved):
Looks good :)

integreat_cms/cms/forms/custom_content_model_form.py
MizukiTemma: Agree 😅 See #2849

PR #2831, opened 2024-06-07
Ensure node_modules are in place before running vitest

deen13:
### Short description

This pull request addresses the issue where the Vitest tools fail if the node_modules directory is missing.

### Proposed changes
- If node_modules is not found, npm install is executed silently (with output redirected to /dev/null).

### DIFF
--- a/tools/_functions.sh
+++ b/tools/_functions.sh
@@ -372,14 +372,8 @@ function ensure_webpack_bundle_exists {
 
 function ensure_node_modules_exist {
     if [ ! -d "${BASE_DIR}/node_modules" ]; then
-        echo "Missing node_modules - running npm install..." | print_info
-
-        if ! npm install > /dev/null 2>&1; then
-            echo "Error during npm install" | print_error
-            exit 1
-        fi
-
-        echo "✔ npm install completed successfully" | print_success
+        echo "Missing node_modules - please run ./tools/install.sh" | print_info
+        exit 1
     else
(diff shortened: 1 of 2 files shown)


tools/_functions.sh
```diff
@@ -370,6 +370,21 @@ function ensure_webpack_bundle_exists {
...
+function ensure_node_modules_exist {
+    if [ ! -d "${BASE_DIR}/node_modules" ]; then
+        echo "Missing node_modules - running npm install..." | print_info
+
+        if ! npm install > /dev/null 2>&1; then
+            echo "Error during npm install" | print_error
+            exit 1
+        fi
+
+        echo "✔ npm install completed successfully" | print_success
```
charludo, line 382: I think we should rather do

```suggestion
        echo "Missing node_modules - please run ./tools/install.sh" | print_info
        exit 1
```

(or something similar). The install script does a bunch of additional checks, incl. correct versions of node and npm.

Maybe we could however add a `--node` flag to that script, similar to the existing `--python` flag, which would exit the script after installing node dependencies?

charludo (changes requested):

tools/_functions.sh
deen13: Sounds good @charludo! ☺️

deen13:
I'll squash the branch before merging it, fyi :)

charludo (approved):
LGTM!

david-venhoff (approved):

Issue #2797, opened 2024-05-13
Automatic slug update does not work on SBS form


seluianova:
### Describe the Bug
When the user changes the page title in the editor, the page slug is automatically updated (see slugify_ajax() function).
But this doesn't happen when the page is edited in the side-by-side translation view.

### Steps to Reproduce

1. Go to any page form
2. Open "Translator View" for any language with an existing translation
3. Try to change the title for the page translation

### Expected Behavior
The same behavior as on the page form:
Slug should be updated automatically in the form, and the updated slug should be saved when saving or publishing changes.

### Actual Behavior
Slug is not updated automatically, nor is it updated when saving/publishing a translation

### Additional Information
![image](https://github.com/digitalfabrik/integreat-cms/assets/115008338/70b979d5-b53e-455a-8dcd-3e80e8599c95)

PR #2842, opened 2024-06-11
Fix automatic slug update on sbs form

lunars97:
### Short description

Currently the sbs form slug of the target language is not updated after the title change. It should be updated after the title of the page changes.
### Proposed changes

- Connect title input value with the slug input value so that the slug will change if the title changes

### Side effects

- should be none

### Resolved issues

Fixes: #2797 


MizukiTemma (changes requested):
Thank you for the PR 👍 

Hmm, it would be cool if `slugify_ajax()` could be used here too instead of building another function 🤔 Having two different functions, we probably forget to update one of them if the logic changes in the future.

theresantonie:
@MizukiTemma wouldn't using the slugify_ajax() method mean that the slug is only updated when saving the form (backend)?
Implementing this as a client functionality seems more elegant to me. 
It also looks to me as if slugifyStr and slugify_ajax() are different in their functionality.

MizukiTemma:
@theresantonie 
`slugify_ajax` is callbed in the normal form by front-end events namely changes in the title field.
We can probably do the same for side-by-side view?

    if (
        document.getElementById("id_title") &&
        (document.querySelector('[for="id_title"]') as HTMLElement)?.dataset?.slugifyUrl
    ) {
        document.getElementById("id_title").addEventListener("focusout", ({ target }) => {
            const currentTitle = (target as HTMLInputElement).value;
            const url = (document.querySelector('[for="id_title"]') as HTMLElement).dataset.slugifyUrl;
            slugify(url, { title: currentTitle }).then((response) => {
                /* on success write response to both slug field and permalink */
                slugField.value = response.unique_slug;
                updatePermalink(response.unique_slug);
            });
        });
    }

theresantonie:
@MizukiTemma If this works the same, it is probably better to use an existing function. 👍

lunars97:
> @theresantonie `slugify_ajax` is callbed in the normal form by front-end events namely changes in the title field. We can probably do the same for side-by-side view?
> 
> ```
> if (
>     document.getElementById("id_title") &&
>     (document.querySelector('[for="id_title"]') as HTMLElement)?.dataset?.slugifyUrl
> ) {
>     document.getElementById("id_title").addEventListener("focusout", ({ target }) => {
>         const currentTitle = (target as HTMLInputElement).value;
>         const url = (document.querySelector('[for="id_title"]') as HTMLElement).dataset.slugifyUrl;
>         slugify(url, { title: currentTitle }).then((response) => {
>             /* on success write response to both slug field and permalink */
>             slugField.value = response.unique_slug;
>             updatePermalink(response.unique_slug);
>         });
>     });
> }
> ```

`slugify_ajax` is used in the page form and it uses page as a `model_type`. Should it target page as a model_type? Because I am getting error `NoReverseMatch` and slugifyUrl returning undefined when I am using it. :thinking:

MizukiTemma:
> 
> `slugify_ajax` is used in the page form and it uses page as a `model_type`. Should it target page as a model_type? Because I am getting error `NoReverseMatch` and slugifyUrl returning undefined when I am using it. 🤔

Yes, in `page_sbs` too :)

MizukiTemma (approved):
Looks good 🎉 Thank you 😸 

☝️ Only release note is left 😁

charludo (changes requested):
@lunars97 could you please rebase this on develop? #2636 was recently merged and introduces changes and fixes both to the `slugify_ajax` method and the way it should be called from `update-permalink.ts`.

integreat_cms/static/src/js/forms/update-permalink.ts
```diff
@@ -86,4 +86,19 @@ window.addEventListener("load", () => {
...
+        const targetLinkElement = document.getElementById("target_slug-link") as HTMLInputElement;
+        const dataset = (document.querySelector('[for="id_slug"]') as HTMLElement).dataset;
+        slugify(dataset.slugifyUrl, { title: targetTitleElement, model_id: dataset.modelId })
+            .then((response) => {
+                /* on success write response to both slug field and permalink */
+                targetLinkElement.value = response.unique_slug;
+                updatePermalink(response.unique_slug);
+            })
+            .finally(() => submissionLock.release());
+    });
```
charludo, line 103: This is (almost) exact same function from line 47ff:
```typescript
        document.getElementById("id_title").addEventListener("focusout", ({ target }) => {
            const submissionLock = new SubmissionPrevention(".no-premature-submission");
            const currentTitle = (target as HTMLInputElement).value;
            const dataset = (document.querySelector('[for="id_title"]') as HTMLElement).dataset;
            slugify(dataset.slugifyUrl, { title: currentTitle, model_id: dataset.modelId })
                .then((response) => {
                    /* on success write response to both slug field and permalink */
                    slugField.value = response.unique_slug;
                    updatePermalink(response.unique_slug);
                })
                .finally(() => submissionLock.release());
        });
    }
```

It's generally not super maintainable to have such duplicate pieces of code. Do you think you could write a function that accomodates both usecases, i.e. is called with arguments for the element ID and the target element?

charludo (changes requested):
Thanks for the changes! :)
Just one more thing

integreat_cms/static/src/js/forms/update-permalink.ts
lunars97: @charludo you are completely right, I have now merged them into one :smiley_cat:

integreat_cms/static/src/js/forms/update-permalink.ts
```diff
@@ -40,21 +40,25 @@ window.addEventListener("load", () => {
...
-            const currentTitle = (target as HTMLInputElement).value;
-            const dataset = (document.querySelector('[for="id_title"]') as HTMLElement).dataset;
-            slugify(dataset.slugifyUrl, { title: currentTitle, model_id: dataset.modelId })
-                .then((response) => {
-                    /* on success write response to both slug field and permalink */
-                    slugField.value = response.unique_slug;
-                    updatePermalink(response.unique_slug);
-                })
-                .finally(() => submissionLock.release());
+    if (document.querySelectorAll("#id_title")) {
```
charludo, line 43: I think you can remove this line (and the cloding bracket, of course) - if there is no `#id_title`, then `document.querySelectorAll("#id_title").forEach((item) => { ... }` will simply run 0 times.

charludo (approved):
Super sorry for the lonf turnaround time with this review :(

This looks great now, thank you! 😄 🎉  I only have one nitpick left (see below), but will already approve this, since it's very minor.

integreat_cms/static/src/js/forms/update-permalink.ts
JoeyStk: Thank you a lot for your feedback! This is should be resolved now! :blush:

Issue #2843, opened 2024-06-12
Hide internal Zammad messages in API


svenseeberg:
### Motivation
We use Zammad as a back end for the Integreat chat. Currently, our API returns all messages of a ticket to the Integreat App. However, Zammad has a concept of internal messages, that should not be visible to users:

![image](https://github.com/digitalfabrik/integreat-cms/assets/15197000/d240b534-3e4c-43f2-b6b6-108a7f3c5e54)

### Proposed Solution
The Integreat chat API should not return internal Zammad messages to Integreat app users.

### Alternatives
N/A

### User Story
As a councilor I want to store internal information for tickets that are not visible to app users so that I can collaborate with colleagues on extensive answers.

### Additional Context
Zammad TicketArticle objects have a boolean `internal` that indicates if a message should be visible to an app user.

PR #2844, opened 2024-06-12
Keep internal chat messages internal

svenseeberg:
### Short description
Do not return internal Zammad messages as chat responses to app users.

### Proposed changes
Filter messages by internal attribute.

### Side effects
- None

### Resolved issues

Fixes: #2843

### DIFF
--- a/integreat_cms/api/v3/chat/zammad_api.py
+++ b/integreat_cms/api/v3/chat/zammad_api.py
@@ -144,14 +144,17 @@ class ZammadChatAPI:
     # pylint: disable=method-hidden
     def get_messages(self, chat: UserChat) -> dict[str, dict | list[dict]]:
         """
-        Get all messages for a given ticket
+        Get all non-internal messages for a given ticket
 
         :param chat: UserChat instance for the relevant Zammad ticket
         """
-        response = self._parse_response(
-            self._attempt_call(self.client.ticket.articles, chat.zammad_id)
-        )
+        raw_response = self._attempt_call(self.client.ticket.articles, chat.zammad_id)
+        if not isinstance(raw_response, list):
+            return self._parse_response(raw_response)  # type: ignore[return-value]
 
+        response = self._parse_response(
+            [article for article in raw_response if not article.get("internal")]
+        )
         for message in response:
             if "attachments" in message:
                 message["attachments"] = [


svenseeberg:
For testing purposes, use the https://passbolt.tuerantuer.org/app/passwords/view/a9e7cb57-2a92-453b-bb3d-c89fe8bbe1f9 credentials to log in on zammad-integreat.tuerantuer.org and set up a API key in https://zammad-integreat.tuerantuer.org/#profile/token_access.

Set this API token for testumgebung in your CMS. Then you can send the following test request:

```bash
curl http://127.0.0.1:8000/api/v3/testumgebung/de/chat/8831b7d3-eb39-43bf-8750-b009edc531f5/
```

integreat_cms/api/v3/chat/zammad_api.py
```diff
@@ -149,7 +149,13 @@ def get_messages(self, chat: UserChat) -> dict[str, dict | list[dict]]:
...
         """
         response = self._parse_response(
-            self._attempt_call(self.client.ticket.articles, chat.zammad_id)
+            [
+                article
+                for article in self._attempt_call(
+                    self.client.ticket.articles, chat.zammad_id
+                )
+                if not article.get("internal")
+            ]
```
charludo, line 158: `_attempt_call` can either return a list of response items, or a single error dict. In the latter case, that dict gets flattened into a list of its keys by this list comprehension 😅 

Actually, the code I wrote right below this is also guilty of this. So even though I dislike the `isinstance()` pattern, I think we need to change the entire function to (untested!!):
```python
# pylint: disable=method-hidden
    def get_messages(self, chat: UserChat) -> dict[str, dict | list[dict]]:
        """
        Get all messages for a given ticket

        :param chat: UserChat instance for the relevant Zammad ticket
        """
        raw_response = self._attempt_call(self.client.ticket.articles, chat.zammad_id)
        if not isinstance(raw_response, list):
            return self._parse_response(raw_response)  # type: ignore[return-value]

        response = self._parse_response(
            [article for article in raw_response if not article.get("internal")]
        )
        for message in response:
            if "attachments" in message:
                message["attachments"] = [
                    self._transform_attachment(chat, message["id"], attachment)
                    for attachment in message["attachments"]
                ]

        return {"messages": response}
```

charludo:
@svenseeberg are you still working on this, or do you want me to take over the PR?

svenseeberg:
@charludo that would be awesome. It would also be great if the `to` attribute could be set for incoming messages (https://github.com/digitalfabrik/integreat-chat/issues/20#issue-2369941962).

charludo:
> @charludo that would be awesome. It would also be great if the `to` attribute could be set for incoming messages ([digitalfabrik/integreat-chat#20 (comment)](https://github.com/digitalfabrik/integreat-chat/issues/20#issue-2369941962)).

No luck, unfortunately. If you don't mind, I'd suggest we complete this PR for now, and take another look at the issue you linked afterwards.

JoeyStk (approved):
Thank you very much for taking over this ticket and moving it forward!
This looks good to me. 

I only have one nit: We could do is change the comment of the method a bit, like `Get all not internal messages for a given ticket` or something, but it's not a must :)

charludo:
> I only have one nit: We could do is change the comment of the method a bit, like `Get all not internal messages for a given ticket` or something, but it's not a must :)

Good point! Changed.

MizukiTemma (approved):
Looks good 👍 Thank you so much 😸

deen13 (approved):
LGTM :)

Issue #2827, opened 2024-06-06
Enable A/B testing for the Integreat Chat


charludo:
### Motivation

We want to have a random user exposure with a defined chance of the chat feature being visible to app users. We want to be able to set a chance value per region that allows the app to decide whether it should show the feature or not. 

### Proposed Solution

- Add a `chat_feature_probability` field to the region model region 
- include the probability in the response to checking if the chat functionality is available

### Alternatives

- only return `True` from the endpoint in `chat_feature_probability`% of cases.

### Additional Context

Blocked by 
- [x] #2619

PR #2846, opened 2024-06-13
Add A/B testing for the Integreat Chat

charludo:
### Short description

(Temporarily) add A/B testing for the Integreat Chat.

### Proposed changes

- in the region form, a percentage for the amount of users who should be able to test the chat feature can be set
- from the `is_chat_enabled_for_user` API endpoint, `true` is returned in that percentage of cases
- additionally, the requesting user ID is stored in the DB, and if the device is already known, the cached `is_tester` value is returned instead. This allows for consistent A/B groups.
- the `is_chat_enabled` API endpoint behavior has not changed, it still just returns if the region supports chats

~~Note that the API endpoint is "stateless", i.e. if the same client makes multiple requests, it might receive multiple different responses. My suggestion is that the app caches the first responses and does not check again afterwards. The alternative is that we change our API endpoint to also take a `device_id`, then add a new model keeping track of what devices received as their first response. This would be both more work now, as well as more work in the future, when these changes are to be rolled back.~~ After consultation with @steffenkleinle, this is no longer true. The new behavior is as described above.

Additionally, note that I've skipped client side validation for the percentage. IMO it's just not necessary for a non-user facing field which will be rolled back soon-ish.

### Side effects

- none

### Resolved issues

Fixes: #2827 


integreat_cms/cms/migrations/0095_chat_beta_testing.py
```diff
@@ -0,0 +1,60 @@
+# Generated by Django 4.2.13 on 2024-06-17 08:30
+
+import django.core.validators
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    """
+    Add fields fpr chat beta testing functionality
```
JoeyStk, line 10:
```suggestion
    Add fields for chat beta testing functionality
```
Is it a typo or is it an abbreviation I don't know about? :thinking:

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -4083,6 +4091,17 @@ msgstr ""
...
+msgid "Chat beta tester percentage"
+msgstr "Chat Beta-Tester Prozentsatz"
+
+#: cms/models/regions/region.py
+msgid ""
+"Percentage of users selected as beta testers for the Integreat Chat feature"
+msgstr ""
+"Prozentsatz der Nutzer, die als Beta-Tester für den Integreat-Chat "
+"ausgewählt werden sollen"
+
```
JoeyStk, line 4104: This line in the translation file `Ein Zugangs-Token wurde für diese Region existiert, wird hier jedoch aus Sicherheitsgründen nicht angzeigt.` has a typo. I think we can just delete the "wurde". Maybe it makes sense to also correct it in this PR? :)

JoeyStk (approved):
Looks good :)

integreat_cms/cms/migrations/0095_chat_beta_testing.py
charludo: Haha, no, that's a typo alright :D

integreat_cms/locale/de/LC_MESSAGES/django.po
charludo: Yes, good catch, thank you!

MizukiTemma (approved):
🪨

Issue #2849, opened 2024-06-14
Refactor `clean_content` in `custom_content_model_form`


MizukiTemma:
### Motivation

The function `clean_content` in `custom_content_model_form` has 97 lines with two pylint rules disabled. It is probably time to refactor it for clean code and better code standard.

### Proposed Solution

Devide the tasks of this fucntion into multiple of smaller functions?

### Alternatives

Other approaches are welcmed too :)

MizukiTemma:
Fell free to move to any milestone.

PR #2850, opened 2024-06-16
Refactor `clean_content`

JoeyStk:
### Short description

This PR attempts to clean up the code of `clean_content` a little bit. As this is my very first attempt at refactoring empathic and honest feedback is much appreciated. In general I tried to follow instructions and ideas from "Clean Code" by Robert C. Martin, but I'm not 100% sure if I was successful.

### Proposed changes

- Split up `clean_content` into multiple sub-methods
- Remove pylint ignores
- Write a test for the method and it's sub-methods

### Side effects

- None that I'm aware of. Thorough double-checking is much appreciated.

### Resolved issues

Fixes: #2849

### DIFF
--- a/integreat_cms/cms/forms/custom_content_model_form.py
+++ b/integreat_cms/cms/forms/custom_content_model_form.py
@@ -1,21 +1,14 @@
 from __future__ import annotations
 
 from typing import TYPE_CHECKING
-from urllib.parse import urlparse
 
 from django import forms
-from django.conf import settings
 from django.contrib import messages
 from django.core.exceptions import ObjectDoesNotExist
-from django.db.models import Q
 from django.utils.translation import gettext_lazy as _
-from lxml.etree import LxmlError
-from lxml.html import fromstring, tostring
 
 from ..constants import status
-from ..models import MediaFile
-from ..utils import internal_link_utils
-from ..utils.linkcheck_utils import fix_content_link_encoding
+from ..utils.content_utils import clean_content
 from ..utils.slug_utils import generate_unique_slug_helper
 from .custom_model_form import CustomModelForm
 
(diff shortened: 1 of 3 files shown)


integreat_cms/cms/forms/custom_content_model_form.py
```diff
@@ -132,7 +210,22 @@ def clean_content(self) -> str:
                 tostring(monospaced, encoding="unicode"),
             )
 
-        # Set link-external as class for external links
+    def update_links(self, content: HtmlElement) -> None:
```
JoeyStk, line 213: I'm undecided if it is good to first gather all methods for update_links in one method and then call this one. On one hand I think it's cleaner on the other hand I can see myself being confused at first. What do you think? Does anyone have a clear preference?

integreat_cms/cms/forms/custom_content_model_form.py
```diff
@@ -172,7 +275,12 @@ def clean_content(self) -> str:
                         "Updated link url from %s to %s", href, translated_url
                     )
 
-        # Scan for media files in content and replace alt texts
+    def process_images(self, content: HtmlElement) -> None:
```
JoeyStk, line 278: I could find a better title other than this generic one. Do you have any suggestions for a better one?

integreat_cms/cms/forms/custom_content_model_form.py
```diff
@@ -172,7 +275,12 @@ def clean_content(self) -> str:
...
+        """
+        This function processes images by scanning for media files and replacing alt texts.
```
JoeyStk, line 280: To be honest I don't yet fully understand what we need this function for. Therefore the comment is not yet containing the explanation of why we do what we do here

david-venhoff (commented):
Could you extract this functionality into its own file? Right now it is required to construct the form to clean content strings, but that is not really necessary. 
Decoupled content cleaning is for example required for #2530 ([here](https://github.com/digitalfabrik/integreat-cms/pull/2530/files#diff-17e8f7fc11e45516e427aaaf0dff9d02bf77518e9ef9cf6379e6059c4935e53e)) and the feature/add-ical-subscription branch

david-venhoff:
only the functionality to clean the content field :)

integreat_cms/cms/forms/custom_content_model_form.py
```diff
@@ -98,11 +94,10 @@ def clean(self) -> dict[str, Any]:
 
         return cleaned_data
 
-    # pylint: disable=too-many-branches, too-many-locals
     def clean_content(self) -> str:
         """
         Validate the content field (see :ref:`overriding-modelform-clean-method`) and applies changes
-        to ``<img>``- and ``<a>``-Tags to match the guidelines.
+        to ``<h1>``, ``<pre>``, ``<code>``, ``<img>``- and ``<a>``-Tags to match the guidelines.
```
david-venhoff, line 100: Nit: I think it would be easier to just say that it can update any element, otherwise we will always have to remember to keep this doc comment in sync with the implementation

integreat_cms/cms/utils/content_utils.py
```diff
@@ -0,0 +1,147 @@
...
+
+from django.conf import settings
+from django.db.models import Q
+from lxml.html import HtmlElement, tostring
+
+from ..models import MediaFile
+from ..utils import internal_link_utils
+
+
+def clean_content(content: HtmlElement, language_slug: str) -> None:
```
david-venhoff, line 12: I would prefer if this function would not have side-effects and instead just take in the content as a string and return the cleaned content also as string. Since we don't really store  `HtmlElement` anywhere, every place that calls this function would have to convert the content to that first.

integreat_cms/cms/utils/content_utils.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def update_links(content: HtmlElement, language_slug: str) -> None:
+    """
+    Super method that gathers all methods related to updating links
+
+    :param content: The content whose links should be updated
+    :param language_slug: Slug of the current language
+    """
+    set_classes_for_links(content)
```
david-venhoff, line 62: Not too important I guess, but I think it would be nice if this function already contained the for loop and called the other functions from within that loop. That would save some processing work and would also be a bit more easy to understand I think.

Then this function could look roughly like this:
```python
    for link in content.iter("a"):
        set_classes_for_links(link)
        remove_target_attribute(link)
        update_internal_links(link, language_slug)
```

tests/cms/views/pages/test_page_model.py
```diff
@@ -0,0 +1,76 @@
+import pytest
```
david-venhoff, line 1: I am not convinced on the file name. If this file should generally contain tests for the page model, shouldn't it be in the `tests/cms/models/pages` directory?

tests/cms/views/pages/test_page_model.py
```diff
@@ -0,0 +1,76 @@
...
+
+@pytest.mark.parametrize(
+    "login_role_user", PRIV_STAFF_ROLES + [MANAGEMENT, EDITOR], indirect=True
+)
+@pytest.mark.django_db
+def test_clean_content(
+    load_test_data: None,
+    login_role_user: tuple[Client, str],
+) -> None:
+    client, role = login_role_user
```
david-venhoff, line 17: Ideally this test should only test that the `clean_content` function works, so I don't think it should send a post request. Instead, it could just call the `clean_content` function

david-venhoff (commented):
Thanks, this looks mostly good. I have some ideas for improvement

integreat_cms/cms/forms/custom_content_model_form.py
david-venhoff: I think both is fine.

integreat_cms/cms/forms/custom_content_model_form.py
david-venhoff: Since this function currently only does one thing, something specific like `fix_alt_texts` could work

integreat_cms/cms/forms/custom_content_model_form.py
david-venhoff: I think this is useful when someone links to some image that is stored in the media library. Then we want to have an alt text for that image for better accessibility and since we store alt texts in the media library, we can just use those. In most cases, tinymce should already add the correct alt texts, but this function should take care for the remaining cases (and it also fixes outdated alt texts when the alt texts in the media library were changed)

integreat_cms/cms/utils/content_utils.py
david-venhoff: I am not entirely sure I understand. What I proposed is that the signature of `clean_content` looks roughly like this:
```python
def clean_content(content: str, language_slug: str) -> str:
```
So there's no need to convert to an HtmlElement outside of this function

integreat_cms/cms/forms/custom_content_model_form.py
```diff
@@ -132,7 +210,22 @@ def clean_content(self) -> str:
...
+        """
+        Super method that gathers all methods related to updating links
+
+        :param content: The content whose links should be updated
+        """
+        self.set_classes_for_links(content)
+        self.remove_target_attribute(content)
+        self.update_internal_links(content)
+
+    def set_classes_for_links(self, content: HtmlElement) -> None:
```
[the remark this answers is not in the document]
david-venhoff: Maybe `mark_external_links`?

integreat_cms/cms/utils/content_utils.py
JoeyStk: I'm sorry for making it not understandable enough. I felt strange about changing it that way as in [these lines](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/forms/custom_content_model_form.py#L111-L115) we just converted it from a string to an `HtmlElement` and then would need to change it back to a string right away. That made feel weird. Did I make my point clearer this time? :)

tests/cms/views/pages/test_page_model.py
JoeyStk: Do you suggest changing the directory or changing the name of the file? :)

integreat_cms/cms/utils/content_utils.py
david-venhoff: There is no need to do that though. Just pass the string directly to `clean_content`.
For reference, here is what it looks like on the ical branch currently: https://github.com/digitalfabrik/integreat-cms/compare/develop...feature/add-ical-subscription#diff-5f47db8ad88d6c2319f6c38c2dc1cd112d8a4235e0f857514b816280d1b70fd2R103-R105

tests/cms/views/pages/test_page_model.py
david-venhoff: hmm, maybe both? Probably depends on what kind of tests this file should contain in the future. If its only about testing the clean_content functionality, I'd move this test to `tests/cms/utils/test_content_utils.py`.

integreat_cms/cms/utils/content_utils.py
JoeyStk: I think now I got it :)

tests/cms/views/pages/test_page_model.py
JoeyStk: Yes, makes sense!

tests/cms/views/pages/test_page_model.py
JoeyStk: Good point. Is applied now :)

david-venhoff (approved):
Very nice, thanks :)

JoeyStk:
This PR blocks the PR for iCal, as iCal requires the refactored method :)

MizukiTemma (approved):
Thank you so much 💪 It's sooo clean now 😻 

Just don't forget to squash the commits 😉

Issue #2645, opened 2024-02-10
Don't send news when they're overdue


timobrembeck:
### Motivation

Currently, the push notification sending management command sends all push notifications that are scheduled for any point in time in the past, but did not yet get sent:

https://github.com/digitalfabrik/integreat-cms/blob/46bfc49f99f87fb5ba0b37c5828ade0f0ec98103/integreat_cms/core/management/commands/send_push_notifications.py#L57

This means, that even if the command fails for a longer period of time, as soon as it's starting to work again, very old messages might be sent, which could lead to confusion.

### Proposed Solution

- Add a new setting for the maximum overdue time (e.g. 1 or 2 days)
- Only send news which are scheduled for a date in this time frame

Optionally: Indicate in the CMS that a scheduled message is overdue and won't get sent anymore, even if the command would start working again.

### User Story

As an end user, I don't want to receive old news

PR #2861, opened 2024-06-18
Don't send news when they're overdue

deen13:
### Short Description

This pull request introduces a new setting to control the retention time for notifications. Notifications that exceed the specified retention time will no longer be sent and will be marked accordingly in the user interface.

### Proposed Changes

- Add `NOTIFICATION_RETAIN_TIME_IN_HOURS` that can also be set through environment variables as recommended in [The Twelve-Factor App](https://12factor.net/config)
- Implement `is_overdue` function in the PushNotification class to be used by the HTML template
- Add new case to the [push_notification_list_row.html](https://github.com/digitalfabrik/integreat-cms/pull/2861/files#diff-83cf1391d28d5cc93fd2655748cb3664e7ac4fcffe32b7fc71aa0542ec13c402)
- Add retention time filter to [send_push_notifications.py](https://github.com/digitalfabrik/integreat-cms/pull/2861/files#diff-1b2c44019336467e050fd926f3c8ce9a8dccdb4bef702b8b064d3afb1a42b193)
- Implement unit test

### Resolved Issues

- #2645

### DIFF
--- a/integreat_cms/core/settings.py
+++ b/integreat_cms/core/settings.py
@@ -1310,3 +1310,8 @@ USER_CHAT_WINDOW_LIMIT: Final[int] = 50
 
 #: Zammad ticket group used for Integreat chat messages
 USER_CHAT_TICKET_GROUP: Final[str] = "integreat-chat"
+
+#: Duration (in hours) that we retain pending push notifications for retry attempts before discarding them
+NOTIFICATION_RETAIN_TIME_IN_HOURS: Final[int] = int(
+    os.environ.get("INTEGREAT_CMS_NOTIFICATION_RETAIN_TIME_IN_HOURS", 24)
+)
--- a/integreat_cms/core/settings.py
+++ b/integreat_cms/core/settings.py
@@ -227,6 +227,11 @@ assert (
     not 60 % FCM_SCHEDULE_INTERVAL_MINUTES
 ), "Interval must be <= 60 and a divisor of 60"
 
+#: Duration (in hours) that we retain pending push notifications for retry attempts before discarding them
+FCM_NOTIFICATION_RETAIN_TIME_IN_HOURS: Final[int] = int(
+    os.environ.get("INTEGREAT_CMS_NOTIFICATION_RETAIN_TIME_IN_HOURS", 24)
+)
+
 ###########
 # GVZ API #
 ###########
@@ -1310,8 +1315,3 @@ USER_CHAT_WINDOW_LIMIT: Final[int] = 50
 
 #: Zammad ticket group used for Integreat chat messages
 USER_CHAT_TICKET_GROUP: Final[str] = "integreat-chat"
-
-#: Duration (in hours) that we retain pending push notifications for retry attempts before discarding them
-NOTIFICATION_RETAIN_TIME_IN_HOURS: Final[int] = int(
-    os.environ.get("INTEGREAT_CMS_NOTIFICATION_RETAIN_TIME_IN_HOURS", 24)
-)
--- a/integreat_cms/firebase_api/firebase_api_client.py
+++ b/integreat_cms/firebase_api/firebase_api_client.py
@@ -17,7 +17,6 @@ from ..cms.forms.push_notifications.push_notification_translation_form import (
 from ..cms.models import Region
 
 if TYPE_CHECKING:
-
     from ..cms.models.push_notifications.push_notification import PushNotification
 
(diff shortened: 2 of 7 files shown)


deen13:
During testing of this pull request, I discovered a potential new edge case. Notifications that are intended to be sent immediately but fail to send due to some error remain as draft messages. This occurs because if [push_sender.send_all()](https://github.com/digitalfabrik/integreat-cms/blob/1b3fb33f9da43c50797cb880f3509df661b2a104/integreat_cms/cms/views/push_notifications/push_notification_form_view.py#L498) returns False, the calling side also returns False and does not reach the code line that overrides the default value of Draft. This, in my opinion, is a bug.

I would propose that the management command should also pick up notifications that were meant to be sent immediately but failed due to an error. This would require extending both [send_push_notifications.py](integreat_cms/core/management/commands/send_push_notifications.py) and the is_overdue method.

What are your thoughts on this, @david-venhoff, @PeterNerlich, and @timobrembeck?

david-venhoff:
I think that it makes sense to add a retry mechanism to failed push notifications that were meant to be sent immediately :+1:

PeterNerlich:
I think it makes sense. Is there a meaningful distinction between *clicked on send immediately and failed* and *was scheduled and failed*? I think it is also unreasonable for scheduled push notifications to be sent hours later, but I have no idea what a reasonable threshold would be. I probably depends on the content, but there is no algorithm to reliably assume that.

tests/core/management/commands/test_send_push_notifications.py
```diff
@@ -1,33 +1,155 @@
...
+
+    def setup_method(self) -> None:
+        self.patch = patch.object(
+            FirebaseApiClient, "_get_access_token", return_value="secret access token"
+        )
+        self.patch.start()
+
+    def teardown_method(self) -> None:
+        self.patch.stop()
+        self.patch = None
```
david-venhoff, line 72: I think it is more readable if the setup related stuff is at the top

tests/core/management/commands/test_send_push_notifications.py
```diff
@@ -1,33 +1,155 @@
...
+        ) -> dict[str, str | int]:
+            nonlocal called
+            called += 1
+
+            return {
+                "name": "projects/integreat-2020/messages/1",
+                "status_code": 200,
+            }
+
+        requests_mock.post(
```
david-venhoff, line 91: Couldn't we just mock the `send_pn` method to return True?

integreat_cms/firebase_api/firebase_api_client.py
```diff
@@ -17,7 +17,6 @@
 from ..cms.models import Region
 
 if TYPE_CHECKING:
-
```
david-venhoff, line 20: Could you remove this file from the pr? Since there is no actual change

integreat_cms/core/settings.py
```diff
@@ -1310,3 +1310,8 @@
 
 #: Zammad ticket group used for Integreat chat messages
 USER_CHAT_TICKET_GROUP: Final[str] = "integreat-chat"
+
+#: Duration (in hours) that we retain pending push notifications for retry attempts before discarding them
+NOTIFICATION_RETAIN_TIME_IN_HOURS: Final[int] = int(
+    os.environ.get("INTEGREAT_CMS_NOTIFICATION_RETAIN_TIME_IN_HOURS", 24)
+)
```
david-venhoff, line 1317: Could you prefix this setting and add it to the firebase push notification chapter? For example below `FCM_SCHEDULE_INTERVAL_MINUTES`

david-venhoff (commented):
Thanks, this looks mostly good to me :+1:

tests/core/management/commands/test_send_push_notifications.py
deen13: Absolutely :)

tests/core/management/commands/test_send_push_notifications.py
deen13: Maybe but I spent a significant amount of time setting up the mock and was really relieved when it worked. 😄

david-venhoff: :eyes:

david-venhoff (approved):

MizukiTemma (approved):
🎉 Looks good 😸 

Don't forget to squash the commits ;)

Issue #1815, opened 2022-10-31
Refactor language tabs and make them responsive


JoeyStk:
### Motivation
We want to improve the CMS views for mobile screen. One major remaining issue are the language tabs. Up until now they still look wild (s. screenshot). However because they were initially written in Python this needs major changes. Therefore we want to rebuild the language tabs to use Javascript.
Based on that we want check via Javascript how much space there is and fill the space with as many tabs as possible (as by now this is hard-coded at four tabs) and store the left-over tabs in a dropdown "Other languages". In case there is not enough space to even show the current language and "other languages" the dropdown has the higher hierarchy and the current language gets hidden.

### Proposed Solution
- Refactor language tabs using Javascript
- Make language selection usable on mobile screens

### Alternatives
- We can ask UI/UX for better solutions. So far we have only discussed this inside the CMS team

### Additional Context
This is part of the Meta Issue https://github.com/digitalfabrik/integreat-cms/issues/170 to make the CMS responsive.

This is blocked by: 
- #2108

### Design Requirements

![image](https://user-images.githubusercontent.com/72705147/198981354-248a0cff-a303-4154-aa43-71b5750394f2.png)

JoeyStk:
Suggestion from @timoludwig 

My suggestion would be the following:
- On very wide screens which fit all of them, we
  <details>
    <summary>show all tabs</summary>

  ![Screenshot 2023-03-12 at 13-21-11 Integreat Editorial System](https://user-images.githubusercontent.com/16021269/224544534-f642f180-dad7-4813-84a7-8b7b7359b9d0.png)

  </details>
  
- On all screens which fit at least 2 tabs (but not all), we
  <details>
    <summary>show the current language plus as many as possible, and the "other languages" dropdown</summary>

  ![Screenshot 2023-03-12 at 13-21-58 Integreat Editorial System](https://user-images.githubusercontent.com/16021269/224544623-7dfa98bf-a071-47ce-b632-cc466f84b9a1.png)
  ![Screenshot 2023-03-12 at 13-22-29 Integreat Editorial System](https://user-images.githubusercontent.com/16021269/224544629-b84f2ca1-55bf-47bd-afa8-7220b217e072.png)

  </details>
    
- On all screens which only fit one tab, we either
  <details>
    <summary>replace the "other languages" text with another icon</summary>

[...]

PR #2866, opened 2024-06-19
Refactor language tabs

JoeyStk:
### Short description

This PR attempts to refactor the language tabs in JS. 

### Proposed changes

- Delete the hard coded limit of shown language tabs
- Show as many language tabs as there is space

### Side effects

- There is still one weird side effect I couldn't get rid of. When resizing to a bigger screen there is still a strange transition happening. JS wizard @charludo do you maybe have a solution / an idea for this? :thinking: 
- Eventhough I attempted one refactoring myself, I ended up with some spaghetti code again. If you have any suggestions how to resolve it, I highly appreciate it :blush: 

### Resolved issues

Fixes: #1815 

### DIFF
--- a/integreat_cms/cms/templates/_form_language_tabs.html
+++ b/integreat_cms/cms/templates/_form_language_tabs.html
@@ -3,10 +3,10 @@
 {% load i18n %}
 {% load content_filters %}
 {% with translation_states|sort_translation_states:language as sorted_translation_states %}
-    <ul class="flex flex-wrap items-end px-4">
-        {% for other_language, other_status in sorted_translation_states|slice:":4" %}
+    <ul id="tab-wrapper" class="flex flex-wrap items-end px-4">
+        {% for other_language, other_status in sorted_translation_states %}
             <li class="mr-2 -mb-[2px] {% if other_language == language %}z-10{% endif %}">
-                <div class="bg-white font-bold">
+                <div class="font-bold bg-white">
                     <div class="border-b-2 border-white">
                         {% if other_language == language %}
                             <div class="py-[calc(0.75rem+2px)] px-4 border-l border-t border-r border-blue-500 rounded-t text-blue-500 cursor-default">
@@ -15,7 +15,8 @@
                                    href="{% build_url target=target region_slug=request.region.slug language_slug=other_language.slug content_field=content_field content_id=instance.id %}">
                                 {% endif %}
                                 {% if other_status == translation_status.MACHINE_TRANSLATED %}
-                                    <span title="{% translate "Machine translated" %}">
+                                    <span class="translation-status"
+                                          title="{% translate "Machine translated" %}">
                                         <i icon-name="bot"></i>
                                     </span>
                                 {% elif other_status == translation_status.IN_TRANSLATION %}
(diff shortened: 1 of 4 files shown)


integreat_cms/static/src/js/language-tabs.ts
```diff
@@ -0,0 +1,96 @@
...
+                    if (anchor.closest("#language-switcher-list") === null) {
+                        anchor.classList.remove("px-3", "py-2", "bg-white", "whitespace-nowrap", "hover:bg-water-500");
+                        anchor.classList.add(
+                            "px-4",
+                            "py-[calc(0.75rem+2px)]",
+                            "bg-water-500",
+                            "rounded-t",
+                            "hover:bg-white",
+                            "hover:text-blue-500"
+                        );
```
PeterNerlich, line 57: I'd suggest move the list of classes to toggle to variables at the start, to eliminate the risk of accidentally changing only one part in the future resulting in inconsistent states, as well as making the logic slightly more readable.

```ts
const anchorClassesWhenShown = ["px-4", "py-[calc(0.75rem+2px)]", "bg-water-500", "rounded-t", "hover:bg-white", "hover:text-blue-500"];
const anchorClassesWhenHidden = ["px-3", "py-2", "bg-white", "whitespace-nowrap", "hover:bg-water-500"];
```

```suggestion
                    anchor.classList.remove(...anchorClassesWhenShown);
                    anchor.classList.add(...anchorClassesWhenHidden);
                }
                languageSwitcherList.append(tab);
            } else {
                tab.classList.add("mr-2");
                if (anchor) {
                    if (anchor.closest("#language-switcher-list") === null) {
                        anchor.classList.remove(...anchorClassesWhenHidden);
                        anchor.classList.add(...anchorClassesWhenShown);
```

integreat_cms/static/src/js/language-tabs.ts
```diff
@@ -0,0 +1,96 @@
...
+        let tabCount = 0;
+
+        tabs.forEach((tab) => {
+            const marginOfTab = parseFloat(window.getComputedStyle(tab).marginRight);
+            const anchor = tab.getElementsByTagName("a")[0];
+
+            if (
+                usedSpace + tab.clientWidth + marginOfTab + buffer > availableWidth &&
+                tab !== languageSwitcherListElement &&
+                tabCount > 0
```
PeterNerlich, line 30: It's difficult to visualize what's going on here, maybe pulling out the width calculation with an explicit name can make it easier. Also, I'd advocate for using [offsetWidth](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetWidth) over [clientWidth](https://developer.mozilla.org/en-US/docs/Web/API/Element/clientWidth), since it includes the border *(and potentially scroll bar, though not the margin)*.

```suggestion
        tabs.forEach((tab: HTMLElement) => {
            const marginOfTab = parseFloat(window.getComputedStyle(tab).marginRight);
            const anchor = tab.getElementsByTagName("a")[0];
            const fitsWithLanguageSwitcher = usedSpace + tab.offsetWidth + marginOfTab + buffer <= availableWidth;

            if (
                !fitsWithLanguageSwitcher &&
                tab !== languageSwitcherListElement &&
                tabCount > 0
```

PeterNerlich (commented):
not a review, just some thoughts

JoeyStk:
Thank you @PeterNerlich for your suggestions. I in cooperated them, but also did some more refactoring of things that I didn't like. Feel free to do a full review now :)

charludo:
> * JS wizard @charludo

wait __what__? 😂 

In any case, I could not reproduce the transition issue, sorry :( Everything resizes and reorganizes perfectly fine and snappy when resizing. Do you maybe have a screen recording?

JoeyStk:
> I have no idea what (if anything) has been decided with UI/UX, but do you think instead of changing the "Other languages" to the globe, you could just hide the language name and only show the flag on very narrow screens? The dropdown list needs the width, and this is the only way I can think of to fit everything neatly in one row.

Yes, thank you very much for your feedback. I can reproduce it. I was told not to spent too much time on it, especially on the very small screens. But I think just showing the flags is a quick fix and should be okay. Thank you a lot for your idea, I will be implementing it :)

JoeyStk:
@charludo I added a commit that hides language names on small devices. Can you please check again?

charludo (approved):
@JoeyStk I originally meant to only show the flag for the selected language, and keep flag+name for the dropdown content and the globe, but I guess this works just as well :D :D :D

integreat_cms/cms/templates/_form_language_tabs.html
```diff
@@ -68,49 +73,15 @@
...
-                                </div>
-                                <div>
-                                    <span class="fp fp-rounded fp-{{ other_language.primary_country_code }} ml-2"></span>
-                                    {% if other_language.secondary_country_code %}
-                                        <span class="fp fp-rounded fp-{{ other_language.secondary_country_code }}"></span>
-                                    {% endif %}
-                                </div>
-                            </a>
-                        {% endfor %}
+                    <div id="language-switcher-list" class="z-10 hidden h-0">
```
lunars97, line 84: It would be better to change it to `ul` tag because it contains list items :thinking:

lunars97:
Everything looks nice except for these two remarks. Thank you for your PR! :green_heart:

lunars97 (approved):
Works nice, thank you :+1:

Issue #2853, opened 2024-06-17
Fix HIX value display


MizukiTemma:
### Describe the Bug

After #2760 some unwanted changes have been introduced together: HIX value is shown as 9 instead of 9.73 and mouseover does not show the detailed HIX value anymore

### Steps to Reproduce

1. Activate HIX in Region Augsburg (in the local emvironment), set the user language to German
2. Populate pages with HIX value by `./tools/integreat-cms-cli hix_bulk augsburg`
3. Go to the page "Gesundheitsamt" (for example), whose raw HIX score is 9,72960372960373.
4. See "9" in the HIX bar and mouseover doesn't show anything

### Expected Behavior

It shows "9.73" and mourseover reveals the exact HIX score "9,72960372960373"

### Actual Behavior

It shows 9 and mouseover doesn't do anything

### Additional Information

The problem 9 vs 9.73 can only be observed if German is selected in the user setting. This is due to the different style between German (9,73, with comma) and English (9.73 with dot) and `parseFloat()`.

For more information and context, see #2756 

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

dkehne:
The mouseover is not necessary as long
1. you show HIX-value with 2 comma digits e.g. 9.73 and NOT as 9
2. The rounding up works well. So 9.735 is shown AND handled as 9.74. So if we would decrease the MT-threshold to 9.74, a value with 9.735 should work to translate

MizukiTemma:
@dkehne Thank you for quick reply 👍

PR #2867, opened 2024-06-21
Fix HIX value display in the HIX widget

MizukiTemma:
### Short description

This PR fixes the problem that HIX score in the hix widget was shown with decimals cut off if German is chosen in the settings (form example, 14 instead of 14, 99). 

### Proposed changes

- No change in the calculation logic
- Replace `,` to `.` so `parseFloat` works well.

### Side effects

- HIX score is shown with `.` even in German (Currenlty in German 14,99 with comma, but now 14.99 with point). If this is too bad I'll do something for that 😅 

### Resolved issues

Fixes: #2853 

### DIFF
--- a/integreat_cms/release_notes/current/unreleased/2853.yml
+++ b/integreat_cms/release_notes/current/unreleased/2853.yml
@@ -1,0 +1,2 @@
+en: Fix HIX value display in the HIX widget
+de: Korrigiere die Darstellung des HIX-Wertes im HIX-Widget
--- a/integreat_cms/cms/templates/hix_widget.html
+++ b/integreat_cms/cms/templates/hix_widget.html
@@ -28,7 +28,7 @@
             </div>
             <div id="hix-container"
                  data-content-changed
-                 data-initial-hix-score="{{ page_translation_form.instance.rounded_hix_score }}"
+                 data-initial-hix-score="{{ page_translation_form.instance.rounded_hix_score|safe }}"
                  data-initial-hix-feedback="{{ page_translation_form.instance.hix_feedback }}">
                 <div id="hix-bar">
                     <span id="hix-bar-fill">


david-venhoff:
I cannot reproduce the issue:
image

As shown in the image, the decimal are displayed correctly for me in German. Is there anything else I have to do to reproduce the problem?

david-venhoff (commented):
If I understand correctly, the previous behavior for German was that the initial hix score was displayed with a comma and dynamically updated hix scores were displayed with a dot. So its nice that this pr improves the consistency by always displaying a dot (It might make sense to open an issue for correct localization though).

I would just propose to fix the value already in the html template, instead of the ts code. From my testing, adding a `|safe` does the trick, like this:

hix_widget.html
```html
data-initial-hix-score="{{ page_translation_form.instance.rounded_hix_score|safe }}"
```

deen13 (commented):
> HIX score is shown with . even in German (Currenlty in German 14,99 with comma, but now 14.99 with point). If this is too bad I'll do something for that 

In my opinion, it's worth the time to fix that issue because it is definitely incorrect behavior and should be easy to address. 😊

MizukiTemma:
> If I understand correctly, the previous behavior for German was that the initial hix score was displayed with a comma and dynamically updated hix scores were displayed with a dot. So its nice that this pr improves the consistency by always displaying a dot (It might make sense to open an issue for correct localization though).

Ah, ok. Thank you for analysis 👍 I'll then leave it with comma for consistency.

osmers:
Hi there :) I added prio high bcs we should close/merge this pr soon as municipalities are also asking about this behaviour

david-venhoff (approved):

JoeyStk:
Am I seeing it correctly that this is currently not testable due to the hix error? :)

MizukiTemma:
> Am I seeing it correctly that this is currently not testable due to the hix error? :)

It's fixed :) Could you review? @JoeyStk

integreat_cms/release_notes/current/unreleased/2853.yml
```diff
@@ -0,0 +1,2 @@
+en: Fix HIX value display in the HIX widget
+de: Korregiere die HIX-Wert-Anzeige in der HIX-Widget
```
JoeyStk, line 2:
```suggestion
de: Korrigiere die Darstellung des HIX-Wertes im HIX-Widget
```

JoeyStk (commented):
> HIX score is shown with . even in German (Currenlty in German 14,99 with comma, but now 14.99 with point). If this is too bad I'll do something for that 😅

I'll ask about it :)

UPDATE: @osmers says it's fine :)

JoeyStk (approved):

Issue #2877, opened 2024-06-25
Fix Hixfeedback display


MizukiTemma:
### Describe the Bug

There are some bugs in the HIX feedback display.

- A: Only "For review" is shown without details (number of nested sentences etc), if the HIX score is not `None` but HIX feedback is.
- B: Nothing is shown even though the feedback detail exist, if it is in the old format

### Steps to Reproduce A

1. Set a random number to `data-initial-hix-score` and "None" to `data-initial-hix-feedback` in the template.
2. Activate HIX in a region
3. Open a page
4. See "For review" appears in the HIX widget without details

### Steps to Reproduce B

1. Set a random number to `data-initial-hix-score` and `data-initial-hix-feedback="{{ dummy_feedback }}"`
2. In `page_form_view` render "dummy_feedback" with HIX feedback in the old format (see example below)
3. Activate HIX in a region
4. Open a page
5. See nothing about the feedback is shown.

 
### Expected Behavior

The HIX widget shows the details after "For review" if the feedback is not empty.
"For review" is not shown if the feedback is empty. 

### Actual Behavior

As described above.

### Additional Information

The old HIX feedback format before #2794 is for example
"[{\"category\": \"nested-sentences\", \"result\": [[291, 313, 4]]}, {\"category\": \"long-sentences\", \"result\": [[40, 61, 21], [104, 128, 23]]}, {\"category\": \"long-words\", \"result\": [[86, 17], [174, 18], [262, 23], [292, 17], [347, 18]]}, {\"category\": \"passive-voice-sentences\", \"result\": [[259, 265, 17]]}, {\"category\": \"infinitive-constructions\", \"result\": [[225, 240, 15], [241, 245, 15]]}, {\"category\": \"nominal-sentences\", \"result\": [[40, 61, 4], [62, 79, 5], [80, 93, 6], [104, 128, 8], [129, 139, 9], [167, 179, 11], [212, 224, 14], [276, 286, 19]]}, {\"category\": \"future-tense-sentences\", \"result\": []}]"

The new format is:
[...]

PR #2878, opened 2024-06-25
Fix feedback display

MizukiTemma:
### Short description

This PR fixes HIX feedback display in the page form.

### Proposed changes

- For bug A: check whether the HIX feedback is empty so `feedbackCount` does not go up
- For bug B: accept the old format and recast `feedbackResult` in the case.

### Side effects

- I hope none 🙈 

### Resolved issues

Fixes: #2877 

### DIFF
--- a/integreat_cms/static/src/js/analytics/hix-widget.ts
+++ b/integreat_cms/static/src/js/analytics/hix-widget.ts
@@ -76,15 +76,29 @@ const updateHixStateLabel = (state: string) => {
  * Display the HIX feedback details
  */
 const updateHixFeedback = (hixFeedback: string) => {
-    const feedbackSections = document.querySelectorAll("[hix-feedback-category]");
-    const feedbackJson = JSON.parse(hixFeedback);
-
     let feedbackCount = 0;
+    let feedbackJson: any;
+    const feedbackContainer = document.getElementById("hix-feedback");
+
+    try {
+        feedbackJson = JSON.parse(hixFeedback);
+    } catch (error) {
+        feedbackContainer.classList.add("hidden");
+        console.error("Invalid HIX feedback", error);
+        return;
+    }
+
+    const feedbackSections = document.querySelectorAll("[hix-feedback-category]");
 
     feedbackSections.forEach((feedbackSection) => {
         const categoryName = feedbackSection.getAttribute("hix-feedback-category");
         const feedbackEntry = feedbackJson.find((item: { category: string }) => item.category === categoryName);
-        const feedbackResult = feedbackEntry ? feedbackEntry.result : [];
+        let feedbackResult = feedbackEntry ? feedbackEntry.result : [] || Number;
+
+        // Cast array into number if HIX feedback is in the old format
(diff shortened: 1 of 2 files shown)


integreat_cms/static/src/js/analytics/hix-widget.ts
```diff
@@ -76,25 +76,30 @@ const updateHixStateLabel = (state: string) => {
  * Display the HIX feedback details
  */
 const updateHixFeedback = (hixFeedback: string) => {
-    const feedbackSections = document.querySelectorAll("[hix-feedback-category]");
-    const feedbackJson = JSON.parse(hixFeedback);
-
     let feedbackCount = 0;
+    if (hixFeedback !== "None") {
+        const feedbackSections = document.querySelectorAll("[hix-feedback-category]");
+        const feedbackJson = JSON.parse(hixFeedback);
```
charludo, line 82: `None ` no longer triggers the empty "for review" part, but leaving it completely empty does (and so do some other non-sense values, I presume).

IMO it's probably fine, but a more robust solution might be attempting to parse the JSON, and in case that fails, to then not go in the inner loop?

charludo (approved):
Nice, thank you!! 🎉 

One small concern (see below), although I'm not really sure if it's necessary.

integreat_cms/static/src/js/analytics/hix-widget.ts
MizukiTemma: That sounds nice 😸 Applied!

david-venhoff (approved):
Thanks, this fixes the issue.

Approve since this is prio:high, but I don't really like that our database store two different versions of `hixFeedback` currently. I think the more robust way is to create migrations when we change the data that is stored in the db. Then there would be no need to handle different versions of hix feedback in the code. Do you think this is worth opening an issue for?

integreat_cms/static/src/js/analytics/hix-widget.ts
```diff
@@ -76,25 +76,34 @@ const updateHixStateLabel = (state: string) => {
...
+            if (feedbackResult > 0) {
+                const categoryCount = feedbackSection.querySelector("span");
+                categoryCount.textContent = feedbackResult;
+                feedbackSection.classList.remove("hidden");
+                feedbackCount += 1;
+            } else {
+                feedbackSection.classList.add("hidden");
+            }
+        });
+    } catch (error) {
```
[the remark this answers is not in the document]
david-venhoff: Ah, I didn't see the comment by @charludo. If you want to keep the current solution, I would recommend to make the try block only contain the relevant code line

MizukiTemma:
> Thanks, this fixes the issue.
> 
> Approve since this is prio:high, but I don't really like that our database store two different versions of `hixFeedback` currently. I think the more robust way is to create migrations when we change the data that is stored in the db. Then there would be no need to handle different versions of hix feedback in the code. Do you think this is worth opening an issue for?

Opened an issue before forgetting :) #2881 Feel free to modify the issue

Issue #1639, opened 2022-08-17
Enable dynamic linking to fallback translations


ulliholtgrave:
### Motivation

As duplicated content objects are only existing in the default language (e.g. german), our user are only able to link the default translation. This should be updated dynamically to link to their other "fake" translations.

Also the links don't get suggested when you try to use the linking mechanism in the TinyMCE. This should be changed.

Note: This is only required for "real" fallback translations where the content is really copied (so for events and locations). For pages, the fallback mechanism is only a placeholder to list the available languages, so it would not make sense to link to them I guess. But maybe it should be possible to link to the default translation of those?

seluianova:
@ulliholtgrave hey Ulli, could you please describe the use case (reproduction steps) and expected result to better understand the problem?
What do you mean with "duplicated content objects" ?

ulliholtgrave:
This refers to the fallback translations that are for example used for events or [POIs](https://github.com/digitalfabrik/integreat-cms/blob/8236701962221b2fe437b04b9af5985eff7b2a75/integreat_cms/cms/models/pois/poi.py#L130). These objects aren't translated for most of the time, but we still want them to appear in the app. Therefore, we kinda "fake" them and return the default translation if there is no translation for a specific language.
However, this behavior isn't covered by our linking functionality in the TinyMCE. That means that you can only create links to existing translations but our fallback translations are skipped. 
In reality this means that our user are editing a page in e.g. french but always have to link to the german POI if there are no french translations available. 
As a solution we should add this fallback translation for our link suggestion to stick with the same language.

PR #2880, opened 2024-06-27
Return fallback translations in search

david-venhoff:
### Short description

This pr modifies the search of pois and events to include fallback translations.

### Proposed changes

- Additionally search for objects in the default language and include them in the result if they don't have a translation in the requested language
- In search_ajax, check if the translation matches the expected language, and if not, use its fallback translation

### Side effects

- The generated SQL query looks a bit scary :eyes:
- This implementation does not make use of the `fallback_translation_enabled` property and thus adds a bit of overhead if we decide to change for which objects fallback translations are enabled
- Since the `search` method is used in other places too, these will now receive results that are potentially not in the request language. AFAIK this only affects the search functionality of the filter form and I think it is a nice improvement that it is now possible to search for fallback translations there too.

### Resolved issues

Fixes: #1639

### DIFF
--- a/integreat_cms/cms/models/events/event_translation.py
+++ b/integreat_cms/cms/models/events/event_translation.py
@@ -4,6 +4,7 @@ from typing import TYPE_CHECKING
 
 from django.contrib.contenttypes.fields import GenericRelation
 from django.db import models
+from django.db.models import Q
 from django.urls import reverse
 from django.utils.functional import cached_property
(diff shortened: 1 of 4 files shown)


MizukiTemma (approved):
So nice and cool 😻 
Thank you!

JoeyStk (approved):
Thank you very much. This should do it :)

Issue #2870, opened 2024-06-24
Support all languages in PDF export


MizukiTemma:
### Describe the Bug

After testing the PDF export function through all the languages in the live system, it turned out that languages listed below are not supported. 

- [ ] Dari (Burgenlandkreis)
- [x] Tigrinya (Stadt Dachau) #2885 
- [ ] Urdu (Testumgebung)
- [x] Simplified Chinese (Ingolstadt) #2885

### Steps to Reproduce

1. Go to a region which uses the languages (given next to each language)
2. Select some published pages
3. Export them as PDF
4. See boxes instead of letters

### Expected Behavior

The target language is correctly shown.

### Actual Behavior

Boxes (especially in the index)

### Additional Information

See #2857 #1961 

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

MizukiTemma:
@osmers @nassabay 
Dari and Urdu have LTR as textdirection in the live system. Are they but RTL?

osmers:
@MizukiTemma yes, definitely - I'll change it in the system :)

MizukiTemma:
Moved to the backlog, as Dari and Urdu are RTL languages and we decided to deactivate PDF export itsself for them (see #2890)

PeterNerlich:
Support has to be added in upstream library

osmers:
Closing as stale and we are rethinking the PDF export anyways

PR #2885, opened 2024-07-02
Support Tigrinya and Simplified Chinese in PDF export

MizukiTemma:
### Short description

This PR adds font and settings for PDF export in Tigrinya and Simplified Chinese

### Proposed changes

- Add new font data for Simplified Chinese
- Add the slug "ti" in the settings for Geʽez font (Tigrinya is written in Ge'es characters like Amharic, so far as I googled)

### Side effects

- Sentences are going too far on the right side of each page in the exported PDF in Simplified Chinese. I guess this is an upstream issue, though feel free to suggest if you have a idea for (quick local) fix 😺 

### Resolved issues

Fixes: (partially) #2870 


JoeyStk:
For Chinese:
I translated "Deutsch selber lernen" to Chinese and then generated a PDF for it. On the right side of the PDF there is a bit of an overhang (screenshot below)

grafik

JoeyStk (approved):
Except for my comment about the overhang, I think this is good to go! :rocket:

lunars97:
maybe it is possible to add `overflow-wrap: break-word` :thinking: Everything works fine except this overflow. In LibraOffice it is showing that the sentence is way out of the border
image

MizukiTemma:
@JoeyStk @lunars97 
Stupid me 🙈 I only had to check xhtml2pdf repository more 🙈 There was already a discussion about this overflow problem and even a solution `-pdf-word-wrap: CJK;` 😅 

Could you review this PR again? Not only Simplified Chinese and Tigrinya, but also no side effect to other languages?

lunars97 (approved):
Good catch with the word-wrap, it looks good now! :heart_eyes_cat:

JoeyStk (approved):
Nice, works well! Thank you very much :)

Issue #2857, opened 2024-06-18
PDF export not working for Georgian


MizukiTemma:
### Describe the Bug

The newly introduced language Georgian is not supported in our PDF export function. Generated PDF shows boxes instead of Georgian alphabets.

### Steps to Reproduce

1. Go to the region Testumgebung (in the live system)
2. Open the page tree and select Georgian on the top tab
3. Select pages available in Georgian and export as PDF
4. See boxes in titles

### Expected Behavior

Georgian letters appear

### Actual Behavior

Boxes 😿 

### Additional Information

We had the problem in other languages too. This can probably be solved like [this PR](https://github.com/digitalfabrik/integreat-cms/pull/1961).

![georgian PDF](https://github.com/digitalfabrik/integreat-cms/assets/82345046/d0fce322-ecca-4024-ae36-072deb7d04a7)

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

MizukiTemma:
Some other languages may have the same problem 🤔

PR #2887, opened 2024-07-02
Fix pdf export for the georgian language

lunars97:
### Short description

Currently the export function is not showing correct georgian characters, instead it is showing boxes. 

### Proposed changes

- Add fonts for georgian language
- Make necessary adjustments in other components so that georgian language is supported in various browsers and it is versions

### Side effects

- Should be none

### Resolved issues

Fixes: #2857 


lunars97:
@MizukiTemma do you happen to know what is the problem with the testing failures? :pray:  I have really no idea :thinking:

MizukiTemma:
> @MizukiTemma do you happen to know what is the problem with the testing failures? 🙏 I have really no idea 🤔

You can probably remove the new language object from the test data 👀

lunars97:
> > @MizukiTemma do you happen to know what is the problem with the testing failures? 🙏 I have really no idea 🤔
> 
> You can probably remove the new language object from the test data 👀

I also think about, but I thought that maybe it is because of the font or sth like that. I will try to remove it and see :) Thank you a lot for getting back to me :heart_hands:

MizukiTemma (approved):
Looks good 😸 Thank you!

integreat_cms/release_notes/current/unreleased/2857.yml
```diff
@@ -0,0 +1,2 @@
+en: Fix PDF export function for the georgian language
+de: Behebe der PDF-Exportfunktion für die georgische Sprache
```
JoeyStk, line 2:
```suggestion
de: Behebe die Probleme bei der PDF-Exportfunktion für Georgisch
```

JoeyStk (approved):
Thank you very much :blush: 
I have a small comment on the release note, but other than that it looks good :) 
After you squashed the commits, I think this PR is good to go :rocket:

PR #2891, opened 2024-07-04
Remove 'auth' and 'contenttypes' objects from testdata

david-venhoff:
### Short description

This pr fixes a bug where it was impossible to modify groups in migrations, because they were always overwritten by their values from the test data. It was an oversight from #2781 that the new test data file contains auth values.
The updated command I used to generate the test data is `./tools/integreat-cms-cli dumpdata --natural-primary --natural-foreign --exclude auth --exclude contenttypes --o test_data.json`

### Proposed changes

- Exclude `auth` models and `contenttypes` models from the test data (Unfortunately this caused some shuffling around in the test data)

### Side effects

- Test data now also contain some values that were forgotten to add in previous pull requests

### Resolved issues

/

### DIFF
--- a/integreat_cms/cms/fixtures/test_data.json
+++ b/integreat_cms/cms/fixtures/test_data.json
@@ -202,6 +202,7 @@
       "zammad_url": "",
       "zammad_access_token": "",
       "zammad_chat_handlers": "",
+      "chat_beta_tester_percentage": 0,
       "offers": [3, 4, 2]
     }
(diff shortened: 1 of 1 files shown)


MizukiTemma (approved):
👍

deen13 (approved):
I'm not sure how to review this, but since grepping for "auth" in the file returns no results, it seems good to me. 😊

Issue #2893, opened 2024-07-08
Broken HIX request


MizukiTemma:
### Describe the Bug

HIX request is currently not working. See [this conversation](https://chat.tuerantuer.org/digitalfabrik/pl/1c9uxytxrpn85pejeydiiogtze).

### Steps to Reproduce

1. Activate HIX for a region
2. Go to the region
3. Create a page and update HIX by clicking the button
4. See Error

You can als see the error when opening an existing page with no HIX value

### Expected Behavior

HIX value is received, stored and shown

### Actual Behavior

Error with `HTTPError 403: 'Forbidden'`

### Additional Information

![broken HIX](https://github.com/digitalfabrik/integreat-cms/assets/82345046/0bc1eaca-4be9-41fc-a5e0-5f5e67a73a5b)

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

MizukiTemma:
It seems to be failing at `login()` during initialization of `TextlabClient`.

As I accessed the API "/user/login" per command line, it went successsfully, but it won't sadly per `urlopen`.

PR #2896, opened 2024-07-08
Fix blocked HIX request

MizukiTemma:
### Short description

This PR wants to solve the problem that HIX feature is currently not working in Integreat (and in Malte too?).

So far as I googled, the value which will be set as `User-Agent` to requests without `User-Agent` setting in the header, is causing TextLab to reject our requests
 

### Proposed changes

- Set `User-Agent` into the header

### Side effects

- None?
- Suggestion for a better value setting for `User-Agent` will be welcomed :)

### Resolved issues

Fixes: #2893 

### DIFF
--- a/integreat_cms/release_notes/current/unreleased/2893.yml
+++ b/integreat_cms/release_notes/current/unreleased/2893.yml
@@ -1,0 +1,2 @@
+en: Fix blocked HIX request
+de: Behebe blockierte HIX-Anfrage
--- a/integreat_cms/textlab_api/textlab_api_client.py
+++ b/integreat_cms/textlab_api/textlab_api_client.py
@@ -95,5 +95,6 @@ class TextlabClient:
         if auth_token:
             request.add_header("authorization", f"Bearer {auth_token}")
         request.add_header("Content-Type", "application/json")
+        request.add_header("User-Agent", "")
         with urlopen(request) as response:
             return json.loads(response.read().decode("utf-8"))


JoeyStk (approved):
Thank you very much! This fixes the bug for me! Let's hope it works!

Do you think this bug is related to the introduction of the abbreviations? :)

MizukiTemma:
> Do you think this bug is related to the introduction of the abbreviations? :)

I don't think it is likely. It seems that the TextLab server is rejecting requests with the default user agent values, which often appear in requests from bots or other machinal procedures.

JoeyStk:
@MizukiTemma Do you think it makes sense to add a test (after this PR is through) to make sure this bug doesn't happen again? :)

MizukiTemma:
> @MizukiTemma Do you think it makes sense to add a test (after this PR is through) to make sure this bug doesn't happen again? :)

Hmm 🤔 If it were fine to call a real API in each run of the test, we could add a test for this problem, though I don't think it's okay... and, as Textlab is mocked in the test, we cannot really test it 🤔

JoeyStk:
Ok, thank you for explaining :)

MizukiTemma:
**Update from Textlab IT Team**

Their firewall was blocking more than they should they have fixed it. So the HIX feature should be working again.

Though we can mearge this PR, in my opinion, anyway, as this is not harming any functionality and can make our system inmune to this problem in case it happens again in future.

lunars97 (approved):
I also agree that it will not harm to keep it in case something similar happens again. Thank you a lot for finding a solution when we were in need :heart:

Issue #589, opened 2020-11-18
Show number of registered push notification users


timoludwig:
### Motivation

Region managers who send push notifications should be able to see the number of registered push notification users.
This is determined by the number of subscribers to a specific channel.

### Proposed Solution

Find out how  to retrieve this information from the Firebase API.

dkehne:
Any updates here @timobrembeck ?

timobrembeck:
> Any updates here @timobrembeck ?

No, nobody is assigned to this issue, thus nobody is working on it.

svenseeberg:
https://gist.github.com/timrijckaert/218be90541a7285e21455fafa2962dea

JoeyStk:
We understand that this would be really helpful, but after some research from our side it came out that this is not easy to do and now we're wondering how much effort should be spend on this. Also we cannot promise that it is possible at all.
@hauf-toni could you please find out how important this issue is and give us a rough estimate on how many hours we should spend on this to decide whether we should continue work on this. Thank you in advance :)

hauf-toni:
It’s a nice-to-have rather than a must-have. If the number of push notification subscribers would directly impacts regional managers decision-making, then it might be worth exploring. if it’s only for informational purposes and doesn’t change how notifications are sent, the high implementation cost does not justify it. @laravoelk what do you think?
Based on my insights right now would recommend closing this as not planned.

PR #2897, opened 2024-07-10
Attach region as analytics label to fcm call

deen13:
### Short description
To implement ticket #589 and display the number of users who will be reached by a notification (news), I've added the region as an analytics label to all Firebase Cloud Messaging (FCM) calls. The original plan was to call an API that returns the count of subscribed devices per Firebase topic. Unfortunately, Firebase does not offer such an API.

By adding the analytics label, we can leverage the [Firebase Cloud Messaging Data API](https://firebase.google.com/docs/reference/fcmdata/rest) to get an approximate number of delivered messages per topic. It will take some time for users on the live system to send push notifications, and for the API to deliver real data grouped by the analytics label.

Therefore, I propose merging this pull request first so that I can work on the analytics aspect in parallel and hopefully test with real data before merging the next pull request.

##### Resources
- [Blog Post about Analytics Label](https://firebase.blog/posts/2021/09/analytics-labels-app-messaging-campaigns/)
- [Firebase Cloud Messaging Data API](https://firebase.google.com/docs/reference/fcmdata/rest)

### DIFF
--- a/integreat_cms/firebase_api/firebase_api_client.py
+++ b/integreat_cms/firebase_api/firebase_api_client.py
@@ -17,7 +17,6 @@ from ..cms.forms.push_notifications.push_notification_translation_form import (
 from ..cms.models import Region
 
 if TYPE_CHECKING:
-
     from ..cms.models.push_notifications.push_notification import PushNotification
 
 logger = logging.getLogger(__name__)
@@ -135,7 +134,9 @@ class FirebaseApiClient:
                 "android": {
                     "ttl": "86400s",
                 },
-                "fcm_options": {"analytics_label": region.slug},
+                "fcm_options": {
+                    "analytics_label": f"{region.slug}-{pnt.language.slug}"
+                },
             },
         }
         headers = {
--- a/integreat_cms/firebase_api/firebase_api_client.py
+++ b/integreat_cms/firebase_api/firebase_api_client.py
@@ -135,6 +135,7 @@ class FirebaseApiClient:
                 "android": {
                     "ttl": "86400s",
                 },
+                "fcm_options": {"analytics_label": region.slug},
             },
         }
         headers = {


integreat_cms/firebase_api/firebase_api_client.py
```diff
@@ -135,6 +135,7 @@ def send_pn(self, pnt: PushNotificationTranslation, region: Region) -> bool:
                 "android": {
                     "ttl": "86400s",
                 },
+                "fcm_options": {"analytics_label": region.slug},
```
svenseeberg, line 138: Why are we not using the full topic here as well? That would immediately provide the necessary details?
```suggestion
                "fcm_options": {"analytics_label": f"{region.slug}-{pnt.language.slug}-{self.push_notification.channel}",
```

deen13: It seems to me that this would increase complexity without providing any real benefit, as the channel remains unchanged and the language should not impact the analytics. If the business requirements call for analytics based on the selected language, we can adjust accordingly. However, unless that is the case, I’d prefer to keep it as simple as possible. What do you think, @svenseeberg?

svenseeberg: AFAICT it would be very useful to have subscribers per language, not only per region.

deen13: Hey @svenseeberg, could you provide concrete tickets or data to support the assumption that this requirement is imminent? @osmers, is there currently a specific need for tracking the received push notifications per region and language?

I’m not trying to be overly rigid, but I believe it’s important to challenge our assumptions and ensure we are focusing on current, user-centered needs. Implementing this change now, without clear evidence of its necessity, might add unnecessary complexity to our codebase.

Looking forward to your thoughts! ☺️

osmers: @deen13 municipalities often ask for numbers of subscribed users so they can have an idea of how many people they are reaching with their push notifications - this information let's them decide whether to continue their work, increase awareness of the app, put work into reaching specific language groups. So I agree with @svenseeberg that separating the number by languages would be something the municipalities will ask for/are already asking for.

PeterNerlich: If this suggestion is to be implemented, is `-` the appropriate separator here? Do we have region slugs containing this (or alternative separator that might be suggested)?

deen13: Hey @osmers, thanks for the clarification! I've added the language slug to the analytics label.

deen13: > @deen13 municipalities often ask for numbers of subscribed users so they can have an idea of how many people they are reaching with their push notifications - this information let's them decide whether to continue their work, increase awareness of the app, put work into reaching specific language groups.

@osmers I've already implemented a text display showing the average number of users reached within the news tab. I propose adding a new section to the analytics area featuring a line chart that displays the number of users reached per language and day. This would give users a clear overview of how each language is performing, helping them decide which languages to target with the next push notifications. What do you think? ☺️

osmers: Hi @deen13 :) 
can you show me what the already implemented part looks like? 
Not sure if showing it per day is necessary - generally we recommend sending push notifications not more than once a week. It's probably not feasible to just show it for every message? Maybe even within the message page? (Not sure if that is the best place to show it :D)
Also, generally they should send the push in every language - so I'd say them seeing the users reached per language is more an info to figure out which language groups need to be informed more about the existence of the app in general

PR #2832, opened 2024-06-07
Improve the performance of the linkcheck list

david-venhoff:
### Short description

This pr reduces the loading time when opening the link list view. More performance optimizations are possible, but would required bigger changes.
This pr also fixes a bug that the list view basically did not load at all when the django debug toolbar was activated.

### Proposed changes

- Filter urls in sql instead of in python
- Don't use the prefetch optimization, which does not appear to be worth it

### Side effects

- Due to not using the prefetch optimization the total number of database queries for the link list has increased 

### Resolved issues

Fixes: #2539 (Maybe? I could never reproduce the issue where loading takes 20 seconds. Even before this pr I 'only' had to wait for 10 seconds at most)

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #2898, opened 2024-07-10
[Idea wanted] Add centralised link list in the network management [long loading time]

### LINKED ISSUE #1443: Show broken links in network management
### Motivation
We want a centralized overview of broken links.

### Proposed Solution
Show the broken link checker in the network management. Add a column that shows how many regions are affected by the broken link. The table should be sortable by occurrences and region.

MizukiTemma:
### Short description

This is the (probably) easiest and laziest implementation of centralised link list and the classical problem exists here too: the loading time is unacceptably long.

This PR is not intended as final proposal and opened to demonstrate the problem and to ask for ideas/suggestion how to solve it.

- Do you see any possibility to improve loading time with this implementation?
- If we structure the centalised link checker in another way (different from region link list), what do you imagine?

See [this cool insctuction](https://github.com/digitalfabrik/integreat-cms/pull/2832#issuecomment-2182293775) for how to test with realistically large data.

### Proposed changes

- Reuse the template and view of region link list

### Side effects

- Extremely long loading time

### Resolved issues

Fixes: #1443

### DIFF
--- a/integreat_cms/cms/templates/_base.html
+++ b/integreat_cms/cms/templates/_base.html
@@ -254,6 +254,13 @@
                     <i icon-name="layout-grid"></i>
                     {% translate "Admin Dashboard" %}
                 </a>
+                {% if perms.cms.view_broken_links %}
+                    <a href="{% url 'linkcheck_landing' %}"
+                        class="{% if current_menu_item == 'linkcheck' %} active{% endif %}">
+                        <i icon-name="link"></i>
+                        {% translate "Broken Links" %}
+                    </a>
+                {% endif %}
                 {% if perms.cms.view_region %}
                     <a href="{% url 'regions' %}"
                        class="{% if current_menu_item == 'regions' %} active{% endif %}">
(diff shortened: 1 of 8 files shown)


david-venhoff:
I have done some optimizing and pushed everything to a branch: feature/global_link_list_performance
Originally, the global link list did not load at all for me, and with this branch it takes about 7 seconds.
Feel free to merge if it makes sense to you :)

The actual database queries only take about 500ms, so there is still a lot of potential for optimization, but I think that would require a bit more code restructuring. The lowest hanging fruit it probably removing the hot python loop in `filter_urls`. Maybe it would be possible to use something like https://docs.djangoproject.com/en/5.0/ref/models/expressions/#window-functions for partitioning everything in a query, i will take a look at the the next days.

---

The main problem with the current approach was that the python loop in `filter_urls` performed a database query for every single of those ~100k urls, which was way too slow. Moving everything into a single query causes the page to load for me with production data. Also not calling `prefetch` is not required when loading all links, and removing that also gave a nice performance boost.

MizukiTemma:
As succeeded by #2971

PR #2899, opened 2024-07-10
Display avg delivered messages per topic 

deen13:
### Short Description

This pull request implements a new feature that enables region managers to view the average number of accepted push notifications (news) per region. The user story proposes to display the exact count of devices currently subscribed to a topic, but Firebase provides no API for that.

### Open Tasks
- [ ] Test the new service with real data (blocked by #2897)

### Proposed Changes

- Add an analytics label to FCM calls (#2897)
- Add a new role (Firebase Cloud Messaging Data Viewer) with the permission `fcmdata.deliverydata.list` to [GCP](https://console.cloud.google.com/iam-admin/iam?authuser=1&orgonly=true&project=integreat-2020&supportedpurview=organizationId)
- Implement a service to call the API and calculate the average per region
- Implement caching since the API result should only change once per day
- Extract `_get_access_token` as it is now used by both the `FirebaseApiClient` and `FirebaseDataClient`

### Alternative Solutions

- Reverse engineering the call to an unofficial API, but that seems unstable and unreliable for building features. (See: [Gist by timrijckaert](https://gist.github.com/timrijckaert/218be90541a7285e21455fafa2962dea))
- Storing the instance IDs of registered devices and using the IID endpoint. However, this feature is deprecated, would be challenging to maintain, and could potentially violate user privacy if we store the instance ids.

### Resolved issues

Fixes: #589


integreat_cms/firebase_api/firebase_data_client.py
```diff
@@ -0,0 +1,108 @@
...
+        self.timestamp: datetime | None = None
+
+    def is_valid(self) -> bool:
+        """
+        :return: Whether the cache contains a valid entry from today
+        """
+        if self.timestamp is None:
+            return False
+
+        return datetime.now().date() == self.timestamp.date()
```
deen13, line 35: Is there an established method for handling caching in our system? So far, I've only found @cached_property, which differs slightly as it does not expire after a certain time.

timobrembeck: Well, I guess you could just use Django's inbuilt caching framework?

https://docs.djangoproject.com/en/5.0/topics/cache/#basic-usage

No need to re-implement it yourself.

PR #2900, opened 2024-07-11
Add `.sh` tool to load database dump

MizukiTemma:
### Short description

It seems we will need more and more often testing with database dump, so I would like to share this small tool I made after forgetting how to do it and comming back to the instruction again and again.

### Proposed changes

- Gather all the steps in [this instruction](https://github.com/digitalfabrik/integreat-cms/pull/2832#issuecomment-2182293775)

- Use it just like other tools with the path to the extracted SQL data. Ex: `./tools/load_database_dump.sh ../integreat-2024-05-25.sql`

### Side effects

- The error `rm: cannot remove '/home/xxxxxxx/integreat-cms/.postgres/data': Permission denied` will appear. You can just use `sudo rm -r .postgres/data`. If you know how to overcome this inconvenience, let me know please in the review 👀 
- I'm not familier with thema Docker etc and not sure whether this tool can work in others' environment.

### Resolved issues

Fixes: # (related to #2516 )

### DIFF
--- a/tools/load_database_dump.sh
+++ b/tools/load_database_dump.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+# This script can be used to load downloaded database dump.
+
+# Import utility functions
+# shellcheck source=./tools/_functions.sh
+source "$(dirname "${BASH_SOURCE[0]}")/_functions.sh"
+require_installed
+
+SQL_DATA_PATH=$1
+
+bash "${DEV_TOOL_DIR}/prune_database.sh"
+
+docker run -d --name "${DOCKER_CONTAINER_NAME}" -e "POSTGRES_USER=integreat" -e "POSTGRES_PASSWORD=password" -e "POSTGRES_DB=integreat" -v "${BASE_DIR}/.postgres:/var/lib/postgresql" -p "${INTEGREAT_CMS_DOCKER_LISTEN_PORT}":"${INTEGREAT_CMS_DB_PORT}" postgres > /dev/null
+wait_for_docker_container
+
+docker exec -i "${DOCKER_CONTAINER_NAME}" psql -U integreat < "${SQL_DATA_PATH}"
+
+bash "${DEV_TOOL_DIR}/migrate.sh"
+
+bash "${DEV_TOOL_DIR}/create_superuser.sh"
+
+echo -e "✔ Done😻" | print_success
+echo -e "Use ./tools/run.sh as always" | print_info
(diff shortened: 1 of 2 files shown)


deen13 (approved):
That's great! I've been in this situation at least once too. 😄

tools/load_database_dump.sh
```diff
@@ -0,0 +1,24 @@
...
+require_installed
+
+SQL_DATA_PATH=$1
+
+bash "${DEV_TOOL_DIR}/prune_database.sh"
+
+docker run -d --name "${DOCKER_CONTAINER_NAME}" -e "POSTGRES_USER=integreat" -e "POSTGRES_PASSWORD=password" -e "POSTGRES_DB=integreat" -v "${BASE_DIR}/.postgres:/var/lib/postgresql" -p "${INTEGREAT_CMS_DOCKER_LISTEN_PORT}":"${INTEGREAT_CMS_DB_PORT}" postgres > /dev/null
+wait_for_docker_container
+
+docker exec -i "${DOCKER_CONTAINER_NAME}" psql -U integreat < "${SQL_DATA_PATH}"
```
charludo, line 17: Please do not assume docker 😅

tools/load_database_dump.sh
```diff
@@ -0,0 +1,24 @@
...
+
+bash "${DEV_TOOL_DIR}/migrate.sh"
+
+bash "${DEV_TOOL_DIR}/create_superuser.sh"
```
charludo, line 21: Wondering if we could pass default values to the tool somehow?

tools/load_database_dump.sh
charludo: Edit: I think it might be easiest to adapt this line from `loadtestdata.sh`? 
```bash
deescalate_privileges integreat-cms-cli loaddata "${PACKAGE_DIR}/cms/fixtures/test_data.json" --verbosity "${SCRIPT_VERBOSITY}"
```

timobrembeck: `loaddata` probably won't work since it only can load json exports that were created with `dumpdata`.
There is however the management command [dbshell](https://docs.djangoproject.com/en/5.0/ref/django-admin/#dbshell) which could reduce the whole dev tool to:
```
./tools/integreat-cms-cli dbshell path_to_db_dump.sql
```
and migrate & create_superuser can be run subsequently via the shell, too.
In my opinion, we don't necessarily need an extra dev tool for this.

MizukiTemma:
Closed as unnecessary.

Issue #2801, opened 2024-05-14
Hide links to images without alt-tag for screen readers


LeandraH:
### Motivation

The problem are images that don't have an alt-tag, we automatically give them an alt-tag of an empty string. That then means that a screen reader skips them. But the images are wrapped in an anchor tag linking to the image location. Since those anchor tags don't have any text to read out (remember, the content of the anchor is an image that the screen reader skips), the screen reader reads out the last part of the address of the image, which is the name that the file had when it was uploaded. That's pretty confusing.

### Proposed Solution

We decided in the weekly tech meeting that images that don't need to be described by the screen reader also don't need to have the ability to click on them, especially since the linking to image location is mainly so that a user can enlarge the picture. If the picture isn't important to see, a bigger version of it is unlikely to be important either.

The way to do that is to add the attribute `aria-hidden="true"` to the surrounding anchor tag.

So the idea is: if an image has no alt-tag or an alt-tag of an empty string, also give it an aria-hidden="true".

### Alternatives

We could also remove the links entirely for images without an alt-tag since those probably don't need to be enlarged.

### Additional Context

We also discussed this in https://github.com/digitalfabrik/integreat-app/pull/2765

MizukiTemma:
@LeandraH 
Could you assign this issue to a milestone?

JoeyStk:
> Could you assign this issue to a milestone?

We had a chat and agreed that is a nice-to-have in the near future, but not urgent. Therefore I added it to 24Q4

Issue #2491, opened 2023-10-27
Add management command for fixing content translations


david-venhoff:
### Motivation

When saving a content translation, we run [some](https://github.com/digitalfabrik/integreat-cms/blob/b7db61d477cd66384ceb6faa8a23f41c10921524/integreat_cms/cms/forms/custom_content_model_form.py#L93) transformations on it to ensure its consistency.
However, when we change these transformations, content translations will be in an inconsistent state until a new version of them gets created.

### Proposed Solution

Add a management command that performs those transformations on all current content translations and fixes them if required.

### Alternatives

- Do nothing
- Perform these actions manually

### User Story

/

### Design Requirements

/

PR #2901, opened 2024-07-12
Hide links to images with empty alt attributes and hide images with e…

lunars97:
…mpty alt from screen-readers

### Short description

Currently images in media files by default have an empty alt tag and they are getting wrapped around anchor tags with a `src` leading to the source of an image, thus making it not hidden to screen-readers and screen-readers are reading out the path which leads to the file which worsens the user experience of a person who uses the screen readers. If an image is not important to be pronounced by the screen readers, it is better to be hidden together with anchor tag around it.

### Proposed changes

- Images with empty alt text should be hidden from screen-reader with `aria-hidden=true`
- Anchor tags around images also should be hidden if there is no need in enlarging an image (anchor tags are used here as a possibility to enlarge images)
-If an image has an empty tag, the anchor tag should be not clickable. By default anchor tags are focusable and in order to completely hide it from accessibility tree, it is needed to be set to false. That way it can be also hidden to people who use keyboards.

### Side effects

- should be none

### Resolved issues

Fixes: #2801 


integreat_cms/cms/forms/media/replace_media_file_form.py
```diff
@@ -161,5 +161,4 @@ def save(self, commit: bool = True) -> MediaFile:
             user=self.user,
             region=self.instance.region,
         )
-
```
david-venhoff, line 164: Since this file is not relevant in this pr, could you undo this change?

integreat_cms/cms/forms/media/upload_media_file_form.py
```diff
@@ -135,7 +135,6 @@ def clean(self) -> dict:
             ):
                 cleaned_data["file"] = optimized_image
                 cleaned_data["thumbnail"] = generate_thumbnail(file)
-
```
david-venhoff, line 138: same here

integreat_cms/static/src/js/tinymce-plugins/mediacenter/plugin.js
```diff
@@ -19,6 +24,11 @@
                             const imageEl = document.createElement("img");
                             imageEl.src = file.url;
                             imageEl.alt = file.altText;
+                            if (imageEl.alt === "" || imageEl.alt === null) {
+                                linkEl.addEventListener("focus", anchorFocusBlur);
```
david-venhoff, line 28: I am not quite sure what this means. As far as I understand, this tries to immediately unfocus the link when it has been focused?
If that is the case, then shouldn't setting the tabindex attribute be enough?
Also, this would only the html code of the tinymce editor. But since this event listener is not stored in the html code, it will not exist anymore when a client downloads the html page.

david-venhoff (changes requested):
Thanks for this pr! Unfortunately I am not sure this is the right approach though, since updating the images happens on the client side only, which means that users can (accidentally) bypass this, for example inserting an image without using the media library. This will also have no effect on the current images without alt text.

We actually already do some processing on submitted pages on the server, to make sure that it has all required properties: https://github.com/digitalfabrik/integreat-cms/blob/c550646ab873795bec50798cceaf374a0f562e1b/integreat_cms/cms/utils/content_utils.py#L14

If we add the functionality to this function instead, hiding links to images without alt text should be more reliable, and (once we implement #2491) will also automatically fix all existing images.

integreat_cms/static/src/js/tinymce-plugins/mediacenter/plugin.js
lunars97: @david-venhoff thanks for detailed clarification. Now it makes sense. To your question why not only setting tabindex, tabindex makes unfocusable to the users of tab key, but if someone uses mouse or touchscreen, AFAIK tabindex will not work in those cases.

lunars97: @david-venhoff we do not need any changes here anymore then? If so, I will remove the changes on this file

david-venhoff: Yes, I think have the functionality once at `content_utils.py` is sufficient

integreat_cms/cms/utils/content_utils.py
```diff
@@ -162,3 +163,33 @@ def fix_alt_texts(content: HtmlElement) -> None:
...
+    """
+    This function checks whether an image tag wrapped by an anchor tag has an empty alt tag, if so it hides anchor tag from screen-reader and tab-key
+
+    :param content: the content which has an anchor tag wrapped around an img tag
+    """
+
+    for anchor in content.iter("a"):
+        href = anchor.attrib.get("href")
+        # Check if the href attribute ends with the specified image extensions
+        if href.endswith(("png", "jpeg", "svg")):
```
david-venhoff, line 178:
```suggestion
        if href and href.endswith(("png", "jpeg", "svg")):
```
Since anchor are not guaranteed to have the `href` attribute

david-venhoff, line 178: Actually, is there any problem if we just remove the check for file type? To me it seems like this would be generally useful for all `` tags.

david-venhoff (changes requested):
Thanks, this already looks pretty good!

There is just an issue that this can happen:
1. A user creates a link to a media file without alt tag
2. The cms automatically marks the image with `aria-hidden` and `tabindex=-1`
3. Later, someone could go to the media library and add an alt tag
4. When the page is saved the next time, [this function](https://github.com/digitalfabrik/integreat-cms/pull/2901/files#diff-528fbe2ebc54617c45a434a36d049deff4839b5b6c5bec3a5686060abea45c3cR142) gets executed and adds the correct alt tag
5. However, we now have a link to a media file with alt tag, that still has the `aria-hidden` and `tab-index` properties

A way to fix this would be to handle the case when the image already has an alt text. Then any existing `aria-hidden` and `tab-index` properties should probably get removed.

Apart from that, we probably don't need the changes to the `file-entry.tsx` file anymore.

Also probably needs a release note :)

integreat_cms/cms/utils/content_utils.py
lunars97: I think apart from image files, we do not use any anchor tag, so ýou are right :+1:

david-venhoff: Good point. In that case, wouldn't it make sense to just remove the for loop through anchors and the href check for filetype and instead just loop through all images?

integreat_cms/cms/utils/content_utils.py
```diff
@@ -162,3 +163,47 @@ def fix_alt_texts(content: HtmlElement) -> None:
...
+
+    :param content: the content which has an anchor tag wrapped around an img tag
+    """
+
+    for anchor in content.iter("a"):
+        href = anchor.attrib.get("href")
+        # Check if the href attribute ends with the specified image extensions
+        if href and href.endswith(("png", "jpeg", "svg")):
+            # Find all img tags within the anchor tag
+            for img in anchor.iter("img"):
```
david-venhoff, line 180:
```suggestion
    for img in content.iter("img"):
```
Like this

integreat_cms/cms/utils/content_utils.py
```diff
@@ -162,3 +163,47 @@ def fix_alt_texts(content: HtmlElement) -> None:
...
+                # Check if the image's alt attribute is empty
+                if not (alt_text := img.attrib.get("alt")):
```
david-venhoff, line 182: Nit: I think it would be a bit cleaner to swap the two cases, so that the first condition does not include a `not`:
`if (alt_text := img.attrib.get("alt")):` 
and then handle the set aria part in the else

integreat_cms/cms/utils/content_utils.py
lunars97: @david-venhoff but we need to find the href endings with those filetypes so that we can hide them from screen readers (specifically anchor tags) They are not needed when the image has no alt-text :thinking:

david-venhoff: Ah, I get it now :see_no_evil: 

Though I still would prefer if we did not perform a check for file type. Instead maybe we could test if the anchor tag has no text and only  children?

lunars97: sounds good :+1:

david-venhoff (approved):
Very nice, thanks!

There would still be a slight problem if we had a link with multiple images, where some images have alt tags and some don't, but I think this is probably not too important and also not very likely to happen.

integreat_cms/cms/utils/content_utils.py
```diff
@@ -162,3 +163,45 @@ def fix_alt_texts(content: HtmlElement) -> None:
...
+                    logger.debug(
+                        "Set 'aria-hidden' to true for anchor: %r",
+                        tostring(anchor, encoding="unicode"),
+                    )
+                    # Unfocus the anchor tag from tab key
+                    anchor.set("tabindex", "-1")
+                    logger.debug(
+                        "Set 'tabindex' to -1 for anchor: %r",
+                        tostring(anchor, encoding="unicode"),
+                    )
```
PeterNerlich, line 207: Not sure how relevant this is, but if an anchor has multiple images (and no other nodes) as children, it seems like the presence of `aria-hidden` and `tabindex=-1` is only dependent on the very last image. Can we imagine any situation where this would pose a problem, and if yes, how would the desired behaviour look?

david-venhoff: This is what I commented regarding that:

>There would still be a slight problem if we had a link with multiple images, where some images have alt tags and some don't, but I think this is probably not too important and also not very likely to happen.

But fixing the problem shouldn't be too much effort either :shrug:

PeterNerlich: I think I figured out a good way to handle this: Our original issue is that for useless links to images get read out by their URL by screen readers, as there is no readable text content inside. The proposed code fixes that in a way that makes it also run for links with multiple images, but the problem it tries to address should only occur if *none* of the images have `alt` attributes. It should not occur as soon as one does have an `alt` attribute, at least if I understand the function of screen readers correctly.

```suggestion
    for anchor in content.iter("a"):
        # Check if the anchor tag has only img children and no other text content
        if (
            all(child.tag == "img" for child in anchor.iterchildren())
            and not anchor.text_content().strip()
        ):
            if all(img.attrib.get("alt", "") for img in anchor.iter("img")):
                if "aria-hidden" in anchor.attrib:
                    del anchor.attrib["aria-hidden"]
                    logger.debug(
                        "Removed 'aria-hidden' from anchor: %r",
                        tostring(anchor, encoding="unicode"),
                    )
                if "tabindex" in anchor.attrib:
                    del anchor.attrib["tabindex"]
                    logger.debug(
                        "Removed 'tabindex' from anchor: %r",
                        tostring(anchor, encoding="unicode"),
                    )
            else:
                # Hide the anchor tag by setting aria-hidden attribute if the image alt text is empty
                anchor.set("aria-hidden", "true")
                logger.debug(
                    "Set 'aria-hidden' to true for anchor: %r",
                    tostring(anchor, encoding="unicode"),
                )
                # Unfocus the anchor tag from tab key
                anchor.set("tabindex", "-1")
                logger.debug(
                    "Set 'tabindex' to -1 for anchor: %r",
                    tostring(anchor, encoding="unicode"),
                )
```

lunars97: I think it should occur even if one out of multiple images has not an empty alt tag. A screenreader might detect them as a decorative and read out its own `href`path out of context, it might be better to hide an anchor tag completely :thinking: The whole purpose of an anchor tag in this context, AFAIK, it serves as a possibility to enlarge them

lunars97: can we just say then `any ` instead of all? what do you think @PeterNerlich @david-venhoff ?

PeterNerlich: > The whole purpose of an anchor tag in this context, AFAIK, is to enlarge them

That is where I'm completely unsure. My assumption was that the only naturally occurrence of an anchor meant for enlarging the image is when it contains a single image.

david-venhoff: I mean, we could start by only applying the aria attribute if the image is the only child of an anchor. If it later turns out that we have a usecase for an anchor with multiple images, we can still change the code later.

PeterNerlich: I guess it could look like this then. Though I'm not too sure whether the excessive use of walrus operators actually makes this more readable than alternatives.

I also flipped the cases fore adding/removing `aria-hidden` and `tabindex=-1`, I think it was still the way it was meant to be when the condition was negated.

```suggestion
    for anchor in content.iter("a"):
        # Check if the anchor tag has only img children and no other text content
        if (
            len(children := [child for child in anchor.iterchildren()]) == 1
            and (img := children[0]).tag == "img"
            and not anchor.text_content().strip()
        ):
            if img.attrib.get("alt", ""):
                # Hide the anchor tag by setting aria-hidden attribute if the image alt text is empty
                anchor.set("aria-hidden", "true")
                logger.debug(
                    "Set 'aria-hidden' to true for anchor: %r",
                    tostring(anchor, encoding="unicode"),
                )
                # Unfocus the anchor tag from tab key
                anchor.set("tabindex", "-1")
                logger.debug(
                    "Set 'tabindex' to -1 for anchor: %r",
                    tostring(anchor, encoding="unicode"),
                )
            else:
                if "aria-hidden" in anchor.attrib:
                    del anchor.attrib["aria-hidden"]
                    logger.debug(
                        "Removed 'aria-hidden' from anchor: %r",
                        tostring(anchor, encoding="unicode"),
                    )
                if "tabindex" in anchor.attrib:
                    del anchor.attrib["tabindex"]
                    logger.debug(
                        "Removed 'tabindex' from anchor: %r",
                        tostring(anchor, encoding="unicode"),
                    )
```

lunars97: > I guess it could look like this then. Though I'm not too sure whether the excessive use of walrus operators actually makes this more readable than alternatives.
> 
> I also flipped the cases fore adding/removing `aria-hidden` and `tabindex=-1`, I think it was still the way it was meant to be when the condition was negated.

I thought we can leave it like this for now, but your suggestion seems to be bettter :)

integreat_cms/cms/utils/content_utils.py
```diff
@@ -162,3 +163,47 @@ def fix_alt_texts(content: HtmlElement) -> None:
...
+                    logger.debug(
+                        "Removed 'aria-hidden' from anchor: %r",
+                        tostring(anchor, encoding="unicode"),
+                    )
+                if "tabindex" in anchor.attrib:
+                    del anchor.attrib["tabindex"]
+                    logger.debug(
+                        "Removed 'tabindex' from anchor: %r",
+                        tostring(anchor, encoding="unicode"),
+                    )
```
PeterNerlich, line 209:
```suggestion
            if img.attrib.get("alt", ""):
                if "aria-hidden" in anchor.attrib:
                    del anchor.attrib["aria-hidden"]
                    logger.debug(
                        "Removed 'aria-hidden' from anchor: %r",
                        tostring(anchor, encoding="unicode"),
                    )
                if "tabindex" in anchor.attrib:
                    del anchor.attrib["tabindex"]
                    logger.debug(
                        "Removed 'tabindex' from anchor: %r",
                        tostring(anchor, encoding="unicode"),
                    )
            else:
                # Hide the anchor tag by setting aria-hidden attribute if the image alt text is empty
                anchor.set("aria-hidden", "true")
                logger.debug(
                    "Set 'aria-hidden' to true for anchor: %r",
                    tostring(anchor, encoding="unicode"),
                )
                # Unfocus the anchor tag from tab key
                anchor.set("tabindex", "-1")
                logger.debug(
                    "Set 'tabindex' to -1 for anchor: %r",
                    tostring(anchor, encoding="unicode"),
                )
```

PeterNerlich (changes requested):
Okay, turns out I was being dumb and read the code without testing it, and of course misrecognized its function when I suggested to switch the cases. I read `img.attrib.get("alt", "")` as *check if the alt tag is empty*, but it just gets the alt tag, and if it doesn't exist at all, returns `""` as a fallback. Then the condition functions exactly the other way around as I thought it did, and your code before I injected myself into the conversation was correct. :see_no_evil: 

Otherwise this looks good and will get my green light immediately once it gets rebased onto develop / receives a proper commit. If you are confused how you even got these changes inside a merge commit *(I know I am, though I also didn't investigate it)* just take the diff as it is and put it on a fresh commit based on develop.

PeterNerlich (commented):
The commit is clean on top of develop, you only missed a few lines from the diff:

in `integreat_cms/cms/utils/content_utils.py`
```diff
     convert_monospaced_tags(content)
     update_links(content, language_slug)
     fix_alt_texts(content)
+    hide_anchor_tag_around_image(content)

     content_str = tostring(content, encoding="unicode", with_tail=False)
     return fix_content_link_encoding(content_str)
```

and the release notes `integreat_cms/release_notes/current/unreleased/2973.yml`
```diff
+en: Fix error of POI creation in event form
+de: Behebe Fehler beim Erstellen eines Ortes im Veranstaltungformular
```

I'm at a loss when it comes to the three failing tests though, as It Works On My Machine™

PeterNerlich (approved):
The only things still failing are the three tests that I suspect are a false positive. I think we might just force merge this PR and see how the pipeline does on develop, if the rest of the team agrees. So we'll take it from here. Thank you and have a great vacation!

Issue #2890, opened 2024-07-02
Inform user of known long-term export problems for specific languages


PeterNerlich:
### Motivation

The export for a few languages *(including but not limited to right-to-left languages, see #2020/#2021/#2022/#2385, #2870 as well as #1498)* is defective and will not be able to be fixed anytime soon for technical reasons. This might pose a pitfall for users of the system, who should generally be able to expect the system to either work as intended or be fixed in due time.

### Proposed Solution

Add a notice when the user is about to export a page to a language where problems are known *(that won't be resolved any time soon)*. Optionally decline to complete the action altogether.

### Alternatives

- Continue leaving it to the user to stumble across export problems, likely in situations where they are depending on it and where the problem becomes apparent on second or third glance only. This can mean high stress for the person who thought they got help, frustration, additional work and loss of trust for the user of our system as well as a continued baseline of load for the service team.

### User Story

As a *user*, I prefer to be *notified of problem areas within the system beforehand so I can work around them* to *finding out something went wrong after the fact*.

### Additional Context

It is to be discussed with the service team whether this desirable and to be implemented.

### Design Requirements

Proposed to take the form of simple warning messages within the existing message system, which would not require additional design.

MizukiTemma:
To be discussed:

- For which languages the PDF export function should be deactivated
- How the info sencense should be.

david-venhoff:
Most problems are caused by the xhtml2pdf library right?
If so, we could also go for an intermediate solution and redirect the user to the html page that we would normally feed into xhtml2pdf instead. Then we could use some js like `window.print()` to use the browsers built-in export to pdf functionality. 
Disadvantages would be that the user experience would be a bit worse and that there are probably slight differences how different browsers convert html to pdf.
Advantage would be that we would have working export and could be pretty sure that the browser handles all languages correctly.
For the ui, we could also show a warning that normal export does not work and offer this as an alternative.

@PeterNerlich @MizukiTemma Do you think something like this could be useful?

<details>

Here is a proof of concept:
![image](https://github.com/digitalfabrik/integreat-cms/assets/78504586/6a1bd9ff-5dff-4b7b-82e8-af7e5ac09aa3)

</details>

MizukiTemma:
@david-venhoff 
Hmm, technically it should work and provide a temporary solution, but I'm not sure if it is "intuitiv" for users, especially when the existing export bulk action is working for some languages, while they have to use the new procedure for other languages 🤔 


[...]

PR #2902, opened 2024-07-13
Deactivate PDF export for some languages

MizukiTemma:
### Short description

This PR deactivates PDF ecport for languages that are specified in the newly introuced variable `PDF_DEACTIVATED_LANGUAGES`.

I know there is an ongoing discussion about the in-browser PDF function as alternative, but opened this PR because the implementation for deactivating was already done before it started 😅 We can of course discard this if we decide for the in-browser PDF function and it can be implemented quickly, or use this as temporary bridge until it comes into the production.

### Proposed changes

- Add a new variable in which slugs of languages can be listed for which the PDF function should be deactivated
- Gray out the PDF bulk action if it is disabled for the current language and show an warning as tooltip

### Side effects

- None?

### Resolved issues

Fixes: #2890 

### DIFF
--- a/example-configs/integreat-cms.ini
+++ b/example-configs/integreat-cms.ini
@@ -187,3 +187,7 @@ GOOGLE_CREDENTIALS = <path-to-credential-json>
 GOOGLE_PROJECT_ID = <project-id>
 # Location
 GOOGLE_TRANSLATE_LOCATION = <location>
+
+[xhtml2pdf]
+# Slugs of languages for which PDF export should be deactivated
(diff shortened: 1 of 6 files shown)


JoeyStk (approved):
Thank you very much!
This looks good for me so far :rocket: Please don't forget to notify our PO about this and what this means :)

deen13 (approved):
Looks good to me! We maybe should add a release note for this, as users might get frustrated if a feature disappears without any prior notice. 😊

MizukiTemma:
> Looks good to me! We maybe should add a release note for this, as users might get frustrated if a feature disappears without any prior notice. 😊

Is already there 😉

PR #2913, opened 2024-07-16
Enable linkcheck for imprint

MizukiTemma:
### Short description

This PR enbales linkcheck for imprint.

This issue was not ranked highly but just implemented because it is (probably) not so complicated and #2530 is kind of affected by this (and anyway I was on the way due to the centralized link checker).

### Proposed changes

- Add it, in the same way with page translation, event translation and POI translation.

### Side effects

- None?

### Resolved issues

Fixes: #1627 

### DIFF
--- a/integreat_cms/cms/linklists.py
+++ b/integreat_cms/cms/linklists.py
@@ -9,6 +9,7 @@ from linkcheck import Linklist
 from .models import (
     Event,
     EventTranslation,
+    ImprintPageTranslation,
     LanguageTreeNode,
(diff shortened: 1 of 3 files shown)


david-venhoff (approved):
Thanks! I am not 100% confident in my review, as there might be something I might have missed, but from my testing everything looks good :tada:

JoeyStk (approved):
I think some more testing makes sense, once this is on our test system, but so far it looks good.
Thank you very much! :)

Issue #2906, opened 2024-07-15
POIs can be archived although they are in-use


JoeyStk:
### Describe the Bug

While spending some time in our system @osmers and I discovered that it's not possible to delete POIs when they are in-use somewhere, but it's still possible to archive them. This didn't make sense to us.

### Steps to Reproduce

1. Go to any region on our test system
2. Click on POIs
3. See delete icon is disabled, archive icon is enabled
4. Click on archive & see that it works

### Expected Behavior

It shouldn't be possible to archive POIs that are in-use somewhere

### Actual Behavior

It is possible to archive POIs that are in-use

### Additional Information

<details>
None

PR #2914, opened 2024-07-17
Prevent POIs in use from being archived

MizukiTemma:
### Short description

This PR deactivates archiving for POIs that are used in events.

### Proposed changes

- Use the same logic with deactivating deletion for POIs in use

### Side effects

- Almost same information is shown in the action box in the form both for archiving and deleting when those are deactivated with only slight wording difference "delete" vs "archive". Unifying both section came into consideration but the design has been kept and the warning is shown twice, because in my opinion users are used to seeing two different sections for each deleting and archiving.

### Resolved issues

Fixes: #2906 

### DIFF
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -7204,6 +7204,19 @@ msgid "Archive location"
 msgstr "Ort archivieren"
 
 #: cms/templates/pois/poi_form_sidebar/action_box.html
+msgid "You cannot archive a location which is referenced by an event."
+msgstr ""
+"Der Ort kann nicht archiviert werden, da er von einer Veranstaltung "
+"verwendet wird."
+
+#: cms/templates/pois/poi_form_sidebar/action_box.html
+msgid "To archive this location, you have to delete this event first:"
+msgid_plural "To archive this location, you have to delete these events first:"
+msgstr[0] "Löschen Sie zuerst diese Veranstaltung, um den Ort zu archivieren:"
+msgstr[1] ""
+"Löschen Sie zuerst diese Veranstaltungen, um den Ort zu archivieren:"
+
+#: cms/templates/pois/poi_form_sidebar/action_box.html
 msgid "Archive this location"
 msgstr "Diesen Ort archivieren"
 
@@ -7354,6 +7367,11 @@ msgid ""
 msgstr ""
 "Dieser Ort kann nicht in der Web-App geöffnet werden, weil er nicht "
 "veröffentlicht ist."
+
+#: cms/templates/pois/poi_list_row.html
+msgid "You cannot archive a location which is used by an event."
+msgstr ""
+"Der Ort kann nicht archiviert werden, da er von einer Veranstaltung verwendet wird."
 
 #: cms/templates/push_notifications/push_notification_form.html
 #, python-format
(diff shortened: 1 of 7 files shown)


david-venhoff (approved):
LGTM

MizukiTemma:
@ulliholtgrave 
Let me know please if I have forgotten something discussed in the grooming :)

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -7203,6 +7203,18 @@ msgstr "Diesen Ort wiederherstellen"
 msgid "Archive location"
 msgstr "Ort archivieren"
 
+#: cms/templates/pois/poi_form_sidebar/action_box.html
+msgid "You cannot archive a location which is referenced by an event."
+msgstr ""
+"Der Ort kann nicht archiviert werden, da er von einem Event verwendet wird."
```
JoeyStk, line 7209:
```suggestion
"Der Ort kann nicht archiviert werden, da er von einer Veranstaltung verwendet wird."
```

integreat_cms/release_notes/current/unreleased/2906.yml
```diff
@@ -0,0 +1,2 @@
+en: Prevent POIs in use from being archived
+de: Verhindere, dass verwendete POIs archiviert werden
```
JoeyStk, line 2:
```suggestion
de: Verhindere, dass verwendete Orte archiviert werden können
```

JoeyStk (changes requested):
Thank you very much! Works well!
I only have two minor comments. Additionally I think adding a small test for this could also make sense and also maybe some documentation somewhere  (this could also just be a comment in a method)  _why_ we did this change, so we don't run into the same problem of guessing why something behaves the way it does again :)

ulliholtgrave:
From a brief look, this seems to cover everything, but @JoeyStk brought up some good points, before we should merge it

MizukiTemma:
@JoeyStk @ulliholtgrave 
Thank you for review! Following changes are added newly.

- Add the usage check also in the bulk action (archiving)
- Not only hide the buttons, but also check the usage and prevent archiving in python level
- Add test 🧪

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -7355,6 +7368,11 @@ msgstr ""
 "Dieser Ort kann nicht in der Web-App geöffnet werden, weil er nicht "
 "veröffentlicht ist."
 
+#: cms/templates/pois/poi_list_row.html
+msgid "You cannot archive a location which is used by an event."
+msgstr ""
+"Der Ort kann nicht archiviert werden, da er von einem Event verwendet wird."
```
JoeyStk, line 7374:
```suggestion
"Der Ort kann nicht archiviert werden, da er von einer Veranstaltung verwendet wird."
```
Was there a reason why you chose to stick with "einem Event" or did you just forgot/oversee it? :)

Maybe a bit of background for my reasoning: While it's certainly not unusual to use Event also as a German word and municipalities would definitely understand, I think staying consistent within our system is important (and we usually use "Veranstaltung") :)

tests/cms/views/poi/test_poi_form.py
```diff
@@ -30,3 +56,162 @@ def test_barrier_free_and_organization_box_appear(
...
+    event.location = used_poi
+    event.save()
+
+    assert used_poi.events.count() > 0
+
+    return used_poi.id
+
+
+@pytest.mark.django_db
+def test_poi_in_use_not_archived(
```
JoeyStk, line 103:
```suggestion
def test_poi_in_use_cannot_be_archived(
```
Nit: I prefer this wording as I find it a bit more self-explanatory

JoeyStk (approved):
Nice!
I really like the test, I like the messaging with the bulk action & the wording.
Apart from two nits I wanted to let you know about, I have nothing more to add :rocket:

integreat_cms/locale/de/LC_MESSAGES/django.po
MizukiTemma: Nö. no special meaning 😅

PR #2919, opened 2024-07-21
Add utility `find_free_tree_id()` to future proof tests creating objects decended from AbstractTreeNode

PeterNerlich:
### Short description

Tests not depending on test data fixtures but creating DB objects prior to assertions are self contained and thus very easy to understand. While values like the primary key are suitibly selected by the django ORM without problems of collision and values for foreign keys are determined by passing the desired model object to the ORM function, relationships like [the expectation that any combination of `lft` and `tree_id` values must be unique for models inheriting from `AbstractTreeNode`](https://github.com/digitalfabrik/integreat-cms/pull/2596/commits/38812657c30bf342b7b40ed2807c7867dc2cac51#diff-49c75de4c1fa67b8c725a15cd51b0b9467f659d07c1f2a552550cdf9b67dd26c) are not ensured by django. The next best thing is to hard-code *some* value.

This is fine when running those tests isolated, but [may result in problems when running against a database that already had the test data loaded by a previous test](https://app.circleci.com/pipelines/github/digitalfabrik/integreat-cms/12171/workflows/dac8de10-10ed-4550-88d5-6bae754f835f/jobs/152082).

### Proposed changes

- Add a utility function to find a free `tree_id` for the model.

### Side effects

- The function **does not** return the first free `tree_id`, but the highest existing one, incremented by 1
  E.g. given that these ids exist: `1`, `2`, `3`, `5`, `8`
  `find_free_tree_id()` would **not** return `4`, but `9`.
  This is because it is easier to implement and I don't see any reason to want exactly the *next free* one.

### Alternatives

- #2921 

### DIFF
--- a/tests/core/management/commands/test_send_push_notifications.py
+++ b/tests/core/management/commands/test_send_push_notifications.py
@@ -24,6 +24,7 @@ from integreat_cms.cms.models import (
 )
 from integreat_cms.firebase_api.firebase_api_client import FirebaseApiClient
 
+from ....utils import find_free_tree_id
 from ..utils import get_command_output
 
 
@@ -107,7 +108,12 @@ class TestSendPushNotification:
         region = Region.objects.create(name="unit-test-region")
 
         LanguageTreeNode.objects.create(
-            language=german_language, lft=1, rgt=2, tree_id=1, depth=1, region=region
+            language=german_language,
+            lft=1,
+            rgt=2,
+            tree_id=find_free_tree_id(LanguageTreeNode),
+            depth=1,
+            region=region,
         )
 
         push_notification = PushNotification.objects.create(
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -10,6 +10,8 @@ from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     from _pytest.logging import LogCaptureFixture
 
+    from integreat_cms.cms.models.abstract_tree_node import AbstractTreeNode
+
 
 def get_messages(caplog: LogCaptureFixture) -> list[str]:
     """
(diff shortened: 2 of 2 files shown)


PeterNerlich:
Someone should propose to crack down on Issues and PRs with titles of excessive length in our organization. There's a certain individual who might become a problem in the future

timobrembeck:
Why do you want to create mptt objects without using the treebeard methods?
No need to re-implement library code:
- https://django-treebeard.readthedocs.io/en/latest/api.html#treebeard.models.Node.add_root
- https://django-treebeard.readthedocs.io/en/latest/api.html#treebeard.models.Node.add_child
- https://django-treebeard.readthedocs.io/en/latest/api.html#treebeard.models.Node.add_sibling

PeterNerlich:
> @timobrembeck 
> Why do you want to create mptt objects without using the treebeard methods?
> No need to re-implement library code

I'm betting a large pizza that @deen13, so shortly after joining the team and kicking ass with his work until now, has not already also dug himself into the depths of what is treebeard and mptt and what methods exist. The reviewers of that PR and me, who has spent some time to develop a reasonable understanding as of recently, didn't think of that way either until you pointed it out.
From an average devs¹ point of view, you create new database objects by calling `Model.objects.create()` and that `Model` inherits from `treebeard.models.Node` over 2 or more steps does just not feel relevant.

Should we maybe extend `AbstractTreeNote` to implement `.objects.create()` to also log something to console to recommend using `add_root()` instead, and that you can just put in any other parameters as you would with `create()`?


¹ *For a specific definition of* average dev *as dictated by my gut feeling*

JoeyStk (approved):
Nice! Thank you! :)

deen13 (approved):
Anything that makes writing tests easier is always highly appreciated. ❤️

PR #2920, opened 2024-07-22
Store delivered push notification count per region and language

deen13:
### Short Description

This pull request introduces a cron job that calls the Firebase Data API daily to fetch the count of delivered push notifications per region and language. This feature was developed following discussions with @dkehne and @osmers.

The implementation lays a solid foundation for future development of statistical analysis and reporting. Additionally, it enables us to provide regional managers with accurate data, should they require it for political purposes.

### Open Tasks
- [x] Test the new service with real data (blocked by #2897)
- [x] Add cron job definition to salt stack
- [x] Implement migration

### Proposed Changes

- Add an analytics label to FCM calls (#2897)
- Add a new role (Firebase Cloud Messaging Data Viewer) with the permission `fcmdata.deliverydata.list` to [GCP](https://console.cloud.google.com/iam-admin/iam?authuser=1&orgonly=true&project=integreat-2020&supportedpurview=organizationId)
- Implement a service to call the API and calculate the count per region
- Implement a cron job that calls the api service on a daily base and stores the result in the model
- Extract `_get_access_token` as it is now used by both the `FirebaseApiClient` and `FirebaseDataClient`


integreat_cms/cms/templates/push_notifications/push_notification_form.html
```diff
@@ -160,6 +160,13 @@ <h3 class="heading font-bold text-black">
                                 {{ push_notification_form.schedule_send.label }}
                             </label>
                         </div>
+                        {% if push_notification_form.regions|length == 1 %}
+                            <span>
+                                {% blocktranslate trimmed %}
+                                    This region reached an average of {{ avg_accepted_notifications }} devices last week.
+                                {% endblocktranslate %}
+                            </span>
+                        {% endif %}
```
deen13, line 169: The issue is about region managers, so I decided to display the text only if the user has access to exactly one region. Does that approach make sense?

PeterNerlich: I'm confused why this should be displayed exclusively to region managers and not to all roles region managers *and higher* (like staff), but I am new to the discussion around the original feature request so I will stop at making this regard only.

I think functionality about authentication / showing data only for certain roles should be done on the python side, there it's both easier to implement and able to be done so explicitly. I don't know if your assumption would hold up, but this should be more robust in any case.
For communicating that to the template I suggest setting `avg_accepted_notifications` to `None` if the user is not allowed to view that data, and then displaying it in the template based on that. If you want to handle the case that data might not be available (different from `None`), I guess using a separate boolean instead for saying to show it at all would be the next best option.

integreat_cms/firebase_api/firebase_data_client.py
```diff
@@ -0,0 +1,184 @@
...
+                "languages": region_data.get_averages(),
+            }
+            for region, region_data in regions.items()
+        }
+
+        cache.set(
+            "firebase_data",
+            output,
+            60 * 60 * 24 * 7,
+        )
```
deen13, line 182: Cache should expire at the start of the next day instead of 24 hours

integreat_cms/firebase_api/firebase_data_backup_cron_job.py
```diff
@@ -0,0 +1,54 @@
...
+
+    This job is scheduled to run at a specified time daily, as defined by the `RUN_AT_TIMES` attribute.
+    It retrieves notification statistics from Firebase and updates or creates records in the database
+    for each region and language combination.
+    """
+
+    RUN_AT_TIMES = ["04:00"]
+
+    schedule = Schedule(run_at_times=RUN_AT_TIMES)
+    code = "integreat_cms.firebase_data_backup_cron_job"
```
deen13, line 22:
```suggestion
    code = "FirebaseDataBackupCronJob"
```

integreat_cms/firebase_api/firebase_data_client.py
```diff
@@ -0,0 +1,172 @@
...
+
+        for item in response_data:
+            if "countNotificationsAccepted" in item["data"]:
+                analytics_label = item.get("analyticsLabel")
+                count_accepted = int(item["data"]["countNotificationsAccepted"])
+                if analytics_label:
+                    region, language = analytics_label.split("-")
+                    regions[region].add(language, count_accepted)
+
+        output: Dict[str, Dict[str, Union[float, Dict[str, float]]]] = {
```
JoeyStk, line 164: Replace average with actual count

integreat_cms/firebase_api/firebase_statistic.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class FirebaseStatistic(AbstractBaseModel):
+    """
+    Data model representing a push notification translation
+    """
+
+    date = models.DateField()
+    region = models.CharField(max_length=100)
+    language = models.CharField(max_length=2)
```
JoeyStk, line 13:
```suggestion
    language_slug = models.CharField(max_length=2)
```
I'd rather call this slug

pyproject.toml
```diff
@@ -231,6 +232,7 @@ pinned = [
     "xlwt==1.3.0",
     "yarl==1.9.4",
     "zammad-py==3.0.0",
+    "django-cron==0.6.0"
```
JoeyStk, line 235: Currently it's under optional-dependencies. Is this correct?

tests/firebase_api/test_firebase_data_client.py
```diff
@@ -0,0 +1,378 @@
...
+    from requests_mock.mocker import Mocker
+
+
+class TestFirebaseDataClient:
+    """
+    Test for :class:`~integreat_cms.firebase_api.firebase_api_client.FirebaseDataClient`
+    """
+
+    def teardown_method(self) -> None:
+        cache.delete("firebase_data")
```
JoeyStk, line 22: This also needs to be removed, because this feature won't be used in the GUI at the moment and as this functionality will be called once per day there is no cache anymore anyway

JoeyStk (changes requested):
Thank you very much! Looks good so far. I only have a few comments and questions :)

PeterNerlich (approved):
Fix black being mad at the empty line after deleting djangos automatic migration creation comment and this can probably be merged. I have a few nitpicks / suggestion I am myself unsure about.

```diff
diff --git a/integreat_cms/cms/migrations/0098_firebasestatistic.py b/integreat_cms/cms/migrations/0098_firebasestatistic.py
index c793103b0..12158237e 100644
--- a/integreat_cms/cms/migrations/0098_firebasestatistic.py
+++ b/integreat_cms/cms/migrations/0098_firebasestatistic.py
@@ -1,4 +1,3 @@
-
 from django.db import migrations, models
 
 
```

What I didn't check after all is whether it actually works with the official API. I'm not confident enough to test around using the production credentials for FCM

PR #2921, opened 2024-07-22
Add stdout message discouraging `AbstractTreeNode.objects.create()`

PeterNerlich:
suggest to use Node.add_root(), node.add_child() and node.add_sibling() instead

### Short description

Tests not depending on test data fixtures but creating DB objects prior to assertions are self contained and thus very easy to understand. While values like the primary key are suitibly selected by the django ORM without problems of collision and values for foreign keys are determined by passing the desired model object to the ORM function, relationships like [the expectation that any combination of `lft` and `tree_id` values must be unique for models inheriting from `AbstractTreeNode`](https://github.com/digitalfabrik/integreat-cms/pull/2596/commits/38812657c30bf342b7b40ed2807c7867dc2cac51#diff-49c75de4c1fa67b8c725a15cd51b0b9467f659d07c1f2a552550cdf9b67dd26c) are not ensured by django. Treebeard supplies separate methods handling this, but one has to be aware that those exist.

### Proposed changes

- Add a stdout message to `AbstractTreeNode.objects.create()` discouraging its use in favour of `Node.add_root()`, `node.add_child()` and `node.add_sibling()`:

image

### Side effects

- This message should not show up in logs, except if stdout is captured in addition to the logging facilities. This is intended, as the message is mostly useful during active development, especially when dealing with tests, not on production servers.

### Alternatives

- #2919 

### DIFF
--- a/integreat_cms/cms/models/abstract_tree_node.py
+++ b/integreat_cms/cms/models/abstract_tree_node.py
@@ -10,7 +10,7 @@ from django.db import models
 from django.utils.functional import cached_property
 from django.utils.translation import gettext_lazy as _
 from treebeard.exceptions import InvalidPosition
-from treebeard.ns_tree import NS_Node
+from treebeard.ns_tree import NS_Node, NS_NodeManager
 
 if TYPE_CHECKING:
     from typing import Any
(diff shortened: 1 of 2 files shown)


deen13 (approved):
Looking good!

david-venhoff (approved):

PeterNerlich:
I originally wanted to wait for @timobrembeck's opinion, but as I anticipate them much preferring this over #2919 and hope there is little to object to, I'm going to merge this already.

Issue #2903, opened 2024-07-15
Refactor calculation of pages with low HIX value on dashboard & analytics


JoeyStk:
### Steps to Reproduce

1. Go on the test server to Testumgebung. See on the dashboard it says there are 52 pages with an HIX value of < 15.
2. Click on translation report, see there it says 59 out of 110 pages are ready for translation (which means 51 pages have a HIX value > treshhold)

### Expected Behavior

Numbers should be the same

### Actual Behavior

Numbers are diverging, 52 != 51

### Additional Information

None

MizukiTemma:
@PeterNerlich @JoeyStk 
It seems pages with hix core = 0.0 are causing this difference.

JoeyStk:
Yes, thank you very much! We found out the same thing. We will refactoring the calculation for HIX a little bit & then use this refactored function as a util function both in the analytics section and the dashboard

PR #2933, opened 2024-07-25
Unifiy calculation of pages (not) ready for mt

PeterNerlich:
### Short description

Move calculation of pages ready / not ready for mt from multiple places to singular location

### Proposed changes

- Move implementation from region dashboard and translation coverage analytics to view utils
  - `get_translations_relevant_to_hix()`
  - `get_translation_under_hix_threshold()`
  - `get_translation_over_hix_threshold()`
- Calculate effects of rounding / threshold on raw hix value to more efficiently determine set of PageTranslations directly on the database
- Add test validating implementation of `round_hix_value()` based on this new method of calculation
- Add tests validating util functions
- Add decorator to disable the hix calculation post save hook
- Add tests validating functionality of said decorator

### Side effects

- There are still two ways with distinct implementations of the same functionality: Our new utility functions and iterating over python PageTranslation objects, filtering them by the `hix_sufficient_for_mt` property. While `hix_sufficient.for_mt` relies on the `round_hix_score()` function, our utility functions predict the effect of the rounding, calculating where the threshold must be on the raw `hix_score`, and using it directly on a database filter. While these are intended to yield the same results *(and we try to validate it in `test_round_hix_score()`)*, they are still different implementations of the same thing which is not ideal.
- One might argue that the `disable_hix_post_save_signal()` decorator might be useful even outside of tests *(e.g. for management commands)* and should be placed among the utils in the `integreat_cms` package. As nothing in the package uses it yet, we still opted to leave it in the test utils.

### Resolved issues

Fixes: #2903 

### DIFF
--- a/integreat_cms/cms/views/utils/hix.py
+++ b/integreat_cms/cms/views/utils/hix.py
@@ -15,6 +15,8 @@ from django.conf import settings
 from django.http import JsonResponse
 from django.views.decorators.http import require_POST
 
+from integreat_cms.cms.models.pages.page_translation import PageTranslation
+from integreat_cms.cms.models.regions.region import Region
 from integreat_cms.cms.utils.round_hix_score import round_hix_score
 
 from ....api.decorators import json_response
@@ -23,6 +25,7 @@ from ....textlab_api.textlab_api_client import TextlabClient
 if TYPE_CHECKING:
     from typing import Final
 
+    from django.db.models.query import QuerySet
     from django.http import HttpRequest
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


integreat_cms/cms/views/utils/hix.py
```diff
@@ -104,3 +107,60 @@ def get_hix_score(request: HttpRequest, region_slug: str) -> JsonResponse:
         )
 
     return JsonResponse({"error": "Could not retrieve hix score"})
+
+
+def get_translations_for_hix_threshold(region: Region) -> QuerySet:
```
david-venhoff, line 112: I initially was a bit confused by the name and only figured out what this function does after reading its body. I am not really sure what a better name would be though. Maybe something like `get_translations_with_hix`? The problem with this name is that it sounds like it would exclude pages with hix score = NULL, though.

tests/cms/utils/test_disable_hix_post_save_signal.py
```diff
@@ -0,0 +1,47 @@
...
+    with disable_hix_post_save_signal():
+        demo_page_translation = create_demo_page_translation()
+
+    assert demo_page_translation.hix_score == 16
+
+
+def test_rule_out_false_positive(db: None) -> None:
+    demo_page_translation = create_demo_page_translation()
+
+    assert demo_page_translation.hix_score is None
```
david-venhoff, line 16: This tests that in hix signals the hix score will be set to none because `hix_enabled` of the region is False, right?
If so, then maybe it makes sense to add `hix_enabled=False` when creating the region, just to be explicit.

david-venhoff (approved):

david-venhoff:
Oh, don't forget the release note :)

MizukiTemma (approved):
Thank you, looks good to me 😸 

Don't forget to squash the commits, please 👍

integreat_cms/cms/views/utils/hix.py
PeterNerlich: What do you think of `get_translations_relevant_to_hix()`?

PeterNerlich: @david-venhoff

david-venhoff: Yes, sounds good to me :)

Issue #2929, opened 2024-07-25
POI page does not show up under pages managed by organisation


osmers:
### Describe the Bug
If I assign a POI to an organisation, the POI page should show up on the organisation information page.

### Steps to Reproduce

1. Go to https://admin.integreat-app.de/testumgebung/organizations/tür-an-tür/edit/
2. See that only one page is listed, even though the POI (https://admin.integreat-app.de/testumgebung/pois/de/3871/edit/) is also managed by this organisation

### Expected Behavior
POI page should show up in list of organisation

### Actual Behavior
It does not

### Additional Information
![image](https://github.com/user-attachments/assets/28832dd8-7616-4b0d-bd55-1e20eb5ffacf)

Issue #2073, opened 2023-02-17
Add new section where you can add contact information


JoeyStk:
### Motivation
We want to introduce a centralized contact database which allows maintaining contact information in one place. The contact addresses can then be embedded in pages or attached to locations.

### User Stories
- As a CMS user I want to be able to centrally manage contact information in order to minimize work when information changes
- As a CMS user I want adresses to be non-translateable without having to mark each address as "do not translate" in order to have consistency and minimal work
- As a CMS user I want the same layout incl. icons for all addresses in order not to worry about consistency
- As a user I want all addresses to be in the same format so that I can always find all necessary information in the same space

### Proposed Solution

- Add a new menu item "Contacts" next to pages and locations.
- The new view allows maintaining existing and adding new contacts.
- The contact information should be typed. The following types should be available: phone, e-mail, city, zip, street, house number, name, website
- The contact information could be merged with the existing data from POIs.
- The TinyMCE editor needs a new button to insert a contact address. It should be up to the editor to decide which kind of information should be inserted into the content.
- When a contact is being updated, pages with embedded content should be updated as well.
- Contacts should be related to POIs and events.

### Additional Context
This needs further specification

osmers:
What kind of further information is needed? Anything we can help with?
This issue would probably solve the translations' problem we have with automatic translations translating addresses as well.

svenseeberg:
@osmers could you please provide some user stories? :-)

osmers:
As a CMS user I want to be able to centrally manage contact information in order to minimize work when information changes
As a CMS user I want adresses to be non-translateable without having to mark each address as "do not translate" in order to have consistency and minimal work
As a CMS user I want the same layout incl. icons for all addresses in order not to worry about consistency
As a user I want all addresses to be in the same format so that I can always find all necessary information in the same space
@karthik-integreat @nassabay @dkehne can you see if you have any other user stories for this?

dkehne:
I would add:
- as a CMS user i want to be able to place the contact information flexibel in the content area
- as a CMS user it should be intuitive to me how i can add/remove the contact information section
- as a CMS user i want to be able to control the design (e.g. sorting of contact information [centrally configured for the region])
- as an APP user i just want to see the contact information in the text that are maintained with content / filled

osmers:
Are there any questions remaining - should we ask Toni for UI/UX input?
[...]

PR #2934, opened 2024-07-26
Show organization locations

MizukiTemma:
### Short description

This PR fixes the problem that locations which belong to an organitation are not shown in its form.

### Proposed changes

- Apply the same procedure with the pages belonging to organizations
- Add test to check both pages and locations of an organization are shown in the form

### Side effects

- None?

### Resolved issues

Fixes: #2929 

### DIFF
--- a/integreat_cms/cms/models/users/organization.py
+++ b/integreat_cms/cms/models/users/organization.py
@@ -64,12 +64,12 @@ class Organization(AbstractBaseModel):
         return f"<Organization (id: {self.id}, slug: {self.slug}, region: {self.region.slug})>"
 
     @property
-    def num_pages(self) -> int:
+    def num_contents(self) -> int:
         """
 
         :return: the current number of maintained pages of an organization object
         """
-        return self.pages.count()
+        return self.pages.count() + self.pois.count()
 
     @property
     def num_members(self) -> int:
--- a/integreat_cms/cms/templates/_related_contents_table.html
+++ b/integreat_cms/cms/templates/_related_contents_table.html
@@ -1,0 +1,65 @@
+{% load i18n %}
+{% load content_filters %}
+<div class="table-listing w-full px-4 pb-4">
+    <div>
+        <label>
+            {{ table_title }}
+        </label>
+    </div>
+    <table class="w-full mt-4 rounded border border-solid border-gray-200 shadow bg-white">
+        <thead>
+            <tr class="border-b border-solid border-gray-200">
+                <th class="text-sm text-left uppercase py-3 pl-4 pr-2">
+                    {% translate "Name in " %} {{ region_default_language.translated_name }}
+                </th>
+                {% if region_default_language != backend_language %}
+                    <th class="text-sm text-left uppercase py-3 pl-4 pr-2">
+                        {% translate "Name in " %} {{ backend_language.translated_name }}
+                    </th>
+                {% endif %}
+            </tr>
+        </thead>
+        <tbody>
+            {% for content in contents %}
+                {% get_translation content region_default_language.slug as content_translation %}
+                {% get_translation content backend_language.slug as backendlang_content_translation %}
+                <tr class="border-b border-solid border-gray-200 hover:bg-gray-100">
+                    {% if content_translation %}
+                        <td class="text-sm text-left py-3 pl-4 pr-2">
+                            <a href="{{ content_translation.backend_edit_link }}"
+                               class="block py-1.5 px-2 overflow-hidden max-w-xs whitespace-nowrap text-ellipsis text-gray-800"
+                               title="{% translate "View content" %}">
+                                {{ content_translation.title }}
+                            </a>
+                        </td>
+                    {% else %}
+                        <td class="text-sm text-left py-3 pl-4 pr-2">
+                            <i>{% translate "Translation not available" %}</i>
+                        </td>
+                    {% endif %}
+                    {% if region_default_language != backend_language %}
+                        {% if backendlang_content_translation %}
+                            <td class="text-sm text-left py-3 pl-4 pr-2">
+                                <a href="{{ backendlang_content_translation.backend_edit_link }}"
+                                   class="block py-1.5 px-2 overflow-hidden max-w-xs whitespace-nowrap text-ellipsis text-gray-800"
+                                   title="{% translate "View content" %}">
+                                    {{ backendlang_content_translation.title }}
+                                </a>
+                            </td>
+                        {% else %}
+                            <td class="text-sm text-left py-3 pl-4 pr-2">
+                                <i>{% translate "Translation not available" %}</i>
+                            </td>
+                        {% endif %}
+                    {% endif %}
+                </tr>
+            {% empty %}
+                <tr>
+                    <td colspan="2" class="px-4 py-3">
+                        {% trans no_content_message %}
+                    </td>
+                </tr>
+            {% endfor %}
+        </tbody>
+    </table>
+</div>
--- a/integreat_cms/cms/templates/organizations/organization_form.html
+++ b/integreat_cms/cms/templates/organizations/organization_form.html
@@ -85,67 +85,11 @@
                         <div class="p-4 rounded bg-water-500">
                             <h3 class="heading font-bold text-black">
                                 <i icon-name="layout" class="mr-2"></i>
-                                {% translate "Pages" %}
+                                {% translate "Pages and locations" %}
                             </h3>
                         </div>
-                        <div class="table-listing w-full p-2">
-                            <table class="w-full mt-4 rounded border border-solid border-gray-200 shadow bg-white">
-                                <thead>
-                                    <tr class="border-b border-solid border-gray-200">
-                                        <th class="text-sm text-left uppercase py-3 pl-4 pr-2">
-                                            {% translate "Name in " %} {{ region_default_language.translated_name }}
-                                        </th>
-                                        {% if region_default_language != backend_language %}
-                                            <th class="text-sm text-left uppercase py-3 pl-4 pr-2">
-                                                {% translate "Name in " %} {{ backend_language.translated_name }}
-                                            </th>
-                                        {% endif %}
-                                    </tr>
-                                </thead>
-                                <tbody>
-                                    {% for page in form.instance.pages.all %}
-                                        {% get_translation page region_default_language.slug as page_translation %}
-                                        {% get_translation page backend_language.slug as backendlang_page_translation %}
-                                        <tr class="border-b border-solid border-gray-200 hover:bg-gray-100">
-                                            {% if page_translation %}
-                                                <td class="text-sm text-left py-3 pl-4 pr-2">
-                                                    <a href="{% url "edit_page" page_id=page.id region_slug=request.region.slug language_slug=region_default_language.slug %}"
-                                                       class="block py-1.5 px-2 overflow-hidden max-w-xs whitespace-nowrap text-ellipsis text-gray-800"
-                                                       title="{% translate "View page" %}">
-                                                        {{ page_translation.title }}
-                                                    </a>
-                                                </td>
-                                            {% else %}
-                                                <td class="text-sm text-left py-3 pl-4 pr-2">
-                                                    <i>{% translate "Translation not available" %}</i>
-                                                </td>
-                                            {% endif %}
-                                            {% if region_default_language != backend_language %}
-                                                {% if backendlang_page_translation %}
-                                                    <td class="text-sm text-left py-3 pl-4 pr-2">
-                                                        <a href="{% url 'edit_page' page_id=page.id region_slug=request.region.slug language_slug=backend_language.slug %}"
-                                                           class="block py-1.5 px-2 overflow-hidden max-w-xs whitespace-nowrap text-ellipsis text-gray-800"
-                                                           title="{% translate "View page" %}">
-                                                            {{ backendlang_page_translation.title }}
-                                                        </a>
-                                                    </td>
-                                                {% else %}
-                                                    <td class="text-sm text-left py-3 pl-4 pr-2">
-                                                        <i>{% translate "Translation not available" %}</i>
-                                                    </td>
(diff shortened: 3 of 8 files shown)


integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -6579,6 +6584,14 @@ msgstr "Weitere Seiten-Infos"
 msgid "This organization currently has no maintained pages."
 msgstr "Von dieser Organisation werden zur Zeit keine Seiten verwaltet."
 
+#: cms/templates/organizations/organization_form.html
+msgid "View location"
+msgstr "Weitere Ort-Infos"
```
JoeyStk, line 6589:
```suggestion
msgstr "Ort ansehen"
```
Hm, this translation is not really fitting :thinking:

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -6579,6 +6584,14 @@ msgstr "Weitere Seiten-Infos"
...
+
+#: cms/templates/organizations/organization_form.html
+msgid "This organization currently has no maintained locations."
+msgstr "Von dieser Organisation werden zur Zeit keine orte verwaltet."
```
JoeyStk, line 6593:
```suggestion
msgstr "Von dieser Organisation werden zur Zeit keine Orte verwaltet."
```
Nit: The "Tekamolo"-principle in German would suggest to put "Zur Zeit" at the beginning of the sentence as in "Zur Zeit werden von dieser Organisation keine Orte verwaltet". But your translation works just as well and also is commonly used :)

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -9438,8 +9452,8 @@ msgstr "Dieser XLIFF Import ist nicht länger verfügbar."
 #: cms/views/pois/poi_actions.py
 msgid "This location cannot be archived because it is referenced by an event."
 msgstr ""
-"Dieser Ort kann nicht archiviert werden, da er von einer "
-"Veranstaltung verwendet wurde."
+"Dieser Ort kann nicht archiviert werden, da er von einer Veranstaltung "
+"verwendet wurde."
```
JoeyStk, line 9456:
```suggestion
"verwendet wird."
```
Am I getting this right, that this is supposed to be in past tense not in present tense? :slightly_smiling_face: :thinking:

tests/cms/views/organizations/test_organization_form.py
```diff
@@ -0,0 +1,140 @@
...
+    load_test_data: None,
+    login_role_user: tuple[Client, str],
+    settings: SettingsWrapper,
+) -> None:
+    """
+    Test that no contents are shown in the organization form if it does not have any content assinged
+    """
+    client, role = login_role_user
+
+    # Test for english messages
```
JoeyStk, line 28: Could you please change this comment to describe why we are doing this? :)

tests/cms/views/organizations/test_organization_form.py
```diff
@@ -0,0 +1,140 @@
...
+def test_poi_form_shows_associated_contents(
+    load_test_data: None,
+    login_role_user: tuple[Client, str],
+    settings: SettingsWrapper,
+) -> None:
+    """
+    Test Organization in Augsburg has two pages and one location.
+    They must be shown in the form.
+    """
+    # Choose an organitation which has a page and a poi assigned
```
JoeyStk, line 78:
```suggestion
    # Choose an organization which has a page and a poi assigned
```

tests/cms/views/organizations/test_organization_form.py
```diff
@@ -0,0 +1,140 @@
...
+    ORGANIZAION_ID = 1
```
JoeyStk, line 79:
```suggestion
    ORGANIZATION_ID = 1
```

tests/cms/views/organizations/test_organization_form.py
```diff
@@ -0,0 +1,140 @@
...
+        assert (
+            response.headers.get("location")
+            == f"{settings.LOGIN_URL}?next={edit_organization}"
+        )
+    else:
+        assert response.status_code == 403
+
+
+@pytest.mark.django_db
+def test_poi_form_shows_associated_contents(
```
JoeyStk, line 69:
```suggestion
def test_organization_form_shows_associated_contents(
```
I'm not entirely understanding this test yet. Am I understanding this correctly that we're testing the organization form, not the poi form?

tests/cms/views/organizations/test_organization_form.py
```diff
@@ -0,0 +1,140 @@
...
+    Test Organization in Augsburg has two pages and one location.
+    They must be shown in the form.
+    """
+    # Choose an organitation which has a page and a poi assigned
+    ORGANIZAION_ID = 1
+
+    # Choose a page and a POI
+    #   - 1) that do not belong to the organization chosen above
+    # and
+    #   - 2) whose titles are not contained in the titles of pages and organizations that belong to the organization
```
JoeyStk, line 84: To be honest, I don't fully understand why we do this. Could you please explain? :)

tests/cms/views/organizations/test_organization_form.py
```diff
@@ -0,0 +1,140 @@
...
+    # Test for english messages
+    settings.LANGUAGE_CODE = "en"
+
+    organization = Organization.objects.filter(id=ORGANIZAION_ID).first()
+    organization_pages = list(organization.pages.all())
+    organization_pois = list(organization.pois.all())
+
+    region = organization.region
+
+    # Make sure this organization has pages and contents
```
JoeyStk, line 99:
```suggestion
```
Nit: I think your code is clean and self-explanatory enough to delete this line :)

integreat_cms/cms/templates/organizations/organization_form.html
```diff
@@ -146,6 +151,69 @@ <h3 class="heading font-bold text-black">
                                 </tbody>
                             </table>
                         </div>
+                        <div class="table-listing w-full p-2">
```
JoeyStk, line 154: Do you think it makes sense with regard to #2073 to already extract this to a separate django template so we can save these lines here and reuse them here and in #2073? :)

JoeyStk (changes requested):
Thank you very much for the quick fix!
I think this looks already quite good. I especially like your tests!
I have a few suggestions how we could improve this a bit more. Please find my suggestions below :)

integreat_cms/cms/templates/organizations/organization_form.html
MizukiTemma: What to you mean with "with regard to #2073"? I don't get the point.

integreat_cms/locale/de/LC_MESSAGES/django.po
MizukiTemma: It was inherited from the already existing "View page".
The suggestion is applied in both.

tests/cms/views/organizations/test_organization_form.py
MizukiTemma: Hmm, I can't imagine what you exactly expect but this part is updated too.

integreat_cms/locale/de/LC_MESSAGES/django.po
MizukiTemma: This is inherited from the existing translation too. Should we once check all the translations for update/modification?

tests/cms/views/organizations/test_organization_form.py
MizukiTemma: Improved (so far I could)

tests/cms/views/organizations/test_organization_form.py
MizukiTemma: Yes, to test is the organization form. Sorry.

tests/cms/views/organizations/test_organization_form.py
JoeyStk: Yes, now I get it. Thank you :)

JoeyStk (approved):
Nice, thank you :)

integreat_cms/cms/templates/organizations/organization_form.html
JoeyStk: In #2073 we will also have a section where we you can see where the contact is being used. I'm just trying to think if we can already refactor this section in a way that will be beneficial then. I hope this makes it clearer :)

tests/cms/views/organizations/test_organization_form.py
JoeyStk: Yes, makes it more clearer, thank you.
I have been wondering why we didn't pick German :)

integreat_cms/locale/de/LC_MESSAGES/django.po
JoeyStk: Oh, I didn't realize this. Hm, on one hand it would be definitely helpful but I assume this will be a tedious task no one will ever do. So I'm unsure. Let's raise it in the next grooming :)

integreat_cms/locale/de/LC_MESSAGES/django.po
JoeyStk: Strange! But thank you very much! :)

tests/cms/views/organizations/test_organization_form.py
MizukiTemma: I started using English for testing messages just because there were already tests doing so. We could of course use German but not all of potential external contributers of our open repository speak German and tests can fail because something is not working in the translation but not the function itsself.

integreat_cms/cms/templates/organizations/organization_form.html
MizukiTemma: Ah 💡 Now I got it.

MizukiTemma:
I'm doing the refactoring. Please wait until further notice.

integreat_cms/cms/templates/_related_contents_table.html
```diff
@@ -0,0 +1,65 @@
...
+                            <td class="text-sm text-left py-3 pl-4 pr-2">
+                                <i>{% translate "Translation not available" %}</i>
+                            </td>
+                        {% endif %}
+                    {% endif %}
+                </tr>
+            {% empty %}
+                <tr>
+                    <td colspan="2" class="px-4 py-3">
+                        {% trans no_content_message %}
```
MizukiTemma, line 59: Somehow this won't be translated into German, while `{% translate content_type_plural|title %}` with either `content_type_plural="pages` or `content_type_plural="locations` are working 🤔

david-venhoff: The problem appears to be that the translation strings for "This organization currently has no maintained pages." and "This organization currently has no maintained locations." are not defined. You can even see in the diff that the translation tool has removed the existing translation. This is probably because it does not recognize that those two message should be translated anymore. I guess the easiest fix is to remove the `{% trans … %}` block here and translate `no_content_message`  already before passing it to `_related_contents_table.html`
TIL django supports using [this](https://docs.djangoproject.com/en/5.1/topics/i18n/translation/#string-literals-passed-to-tags-and-filters) syntax for that:
```python
{% include "../_related_contents_table.html" with contents=form.instance.pages.all content_type_plural="pages" no_content_message=_("This organization currently has no maintained pages.") %}
{% include "../_related_contents_table.html" with contents=form.instance.pois.all content_type_plural="locations" no_content_message=_("This organization currently has no maintained locations.") %}
```

MizukiTemma: @david-venhoff Thank you so much 🙏 It's working now 😸

david-venhoff (approved):
Thanks, LGTM. I only found some small changes

integreat_cms/cms/models/users/organization.py
```diff
@@ -69,7 +69,7 @@ def num_pages(self) -> int:
 
         :return: the current number of maintained pages of an organization object
         """
-        return self.pages.count()
+        return self.pages.count() + self.pois.count()
```
PeterNerlich, line 72: This executes two database queries and adds the resulting integers in python. I went a little crazy and wrote a utility function for getting the count of multiple querysets as a single db query.

```suggestion
        return combined_count(self.pages, self.pois)
```

some_utils.py:
```python
def combined_count(*args: *Queryset) -> int:
    """
    Efficiently calculate the combined count of the given querysets on the database.
    This runs a single query, as opposed to ``A.objects.count() + B.objects.count()``
    which results in two queries and does the adding in python.
    """
    if not args:
        return 0
    if len(args) == 1:
        full = args[0]
    else:
        # Since we are only interested in the count, and the querysets might be from different models,
        # only selects the primary keys and reset any previous sorting using order_by without arguments
        querysets = [q.values("pk").order_by() for q in args]
        # PKs may be from different models, so we want to allow duplicates → all=True
        full = querysets[0].union(*querysets[1:], all=True)
    return full.count()
```

This might be a bit overkill since the only other use I found with a quick simple grep is `integreat_cms/cms/models/media/directory.py:68`, but who knows. I was curious about this problem, why throw away what I learned?

david-venhoff: Tbh, I would be a bit hesitant towards adding this, since it increases the complexity of the code without really having a performance improvement (sub 5ms on a region with 55 organization, burgenlandkreis). I guess what we could do if we really wanted to improve the performance of the region list would be to use `annotate` on the organizations query, to load the counts for every organization in one query? Though that should probably be its own issue.

MizukiTemma: I suggest to do it as a separate issue too (if we do)🤔

integreat_cms/cms/templates/_related_contents_table.html
```diff
@@ -0,0 +1,65 @@
+{% load i18n %}
+{% load content_filters %}
+<div class="table-listing w-full p-2">
+    <div class="px-4">
```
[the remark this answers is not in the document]
MizukiTemma: You have really good eyes 👀

Issue #2918, opened 2024-07-19
Fix language tabs in regions that use long language names


JoeyStk:
### Describe the Bug

When the name of a language is too long, the language tab jump around and are difficult to use. 

### Steps to Reproduce

1. Go to Testumgebung
2. Click on any page
3. Try to click on "other languages"
4. See tab jumps around because one language name is too long

### Expected Behavior

Language tabs shouldn't jump around

### Actual Behavior

They do

### Additional Information

-

PR #2936, opened 2024-07-28
Fix jumping language tabs

JoeyStk:
### Short description

This PR attempts to fix the bug that the "Other languages" dropdown doesn't open up when too long languages are one of the dropdown options.

### Proposed changes

- Decouple the list of "Other languages" from the button

### Side effects

- Graphical less pleasing 

### Resolved issues

Fixes: #2918

### DIFF
--- a/integreat_cms/cms/templates/_form_language_tabs.html
+++ b/integreat_cms/cms/templates/_form_language_tabs.html
@@ -58,10 +58,10 @@
                                         <i icon-name="x"></i>
                                     </span>
                                 {% endif %}
-                                <span class="language-name">{{ other_language.translated_name }}</span>
-                                <span class="fp fp-rounded fp-{{ other_language.primary_country_code }}"></span>
+                                <span class="flex-grow language-name">{{ other_language.translated_name }}</span>
+                                <span class="flex-end w-8 fp fp-rounded fp-{{ other_language.primary_country_code }}"></span>
                                 {% if other_language.secondary_country_code %}
-                                    <span class="fp fp-rounded fp-{{ other_language.secondary_country_code }}"></span>
+                                    <span class="fp ml-1 w-8 fp-rounded fp-{{ other_language.secondary_country_code }}"></span>
                                 {% endif %}
                                 {% if other_language != language %}
                                 </a>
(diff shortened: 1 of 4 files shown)


integreat_cms/cms/templates/_form_language_tabs.html
```diff
@@ -76,14 +76,14 @@
...
+                         class="flex font-bold text-right bg-water-500 py-[calc(0.75rem+2px)] px-4 border-l border-t border-r border-blue-500 rounded-t cursor-default">
                         <i id="language-switcher-globe" icon-name="globe" class="hidden"></i>
                         <span id="language-switcher-text">{% translate "Other Languages" %}</span>
                         <i icon-name="chevron-down"></i>
                     </div>
-                    <ul id="language-switcher-list" class="z-10 hidden h-0">
-                    </ul>
                 </div>
+                <ul id="language-switcher-list" class="z-10 hidden h-0">
+                </ul>
```
deen13, line 86:
```suggestion
                <ul id="language-switcher-list" class="z-10 hidden h-0" />
```

deen13 (approved):
LGTM!

MizukiTemma (approved):
Nice 😌

lunars97 (changes requested):
Seems like it moves right when the dropdown appears 
image

PeterNerlich:
I had a very quick try at an alternative CSS-only way. I did not look for the corresponding classes, but the plain CSS is quite simple:

- *Other Languages* button (`#language-switcher`): `position: relative;`
- dropdown list (`#language-switcher-list`): `position: absolute; right: 0; bottom: 0;`
    - we'd likely want to ask UI/UX for an opinion on adding a visual indicator on top of the dropdown (e.g. slightly more width on the line) to clearly separate it from the other tabs

image

theresantonie (commented):
For me it works, but I've noticed something else: When the browser window is made smaller, the tabs for the languages that don't fit disappear. When the window is enlarged again by click these tabs appear smaller than the others.
Screenshot 2024-07-31 161612
Maybe this deserves its own issue.

integreat_cms/cms/templates/_form_language_tabs.html
JoeyStk: Unfortunately this doesn't work. It seems to break the entire view (s. screenshot)
grafik
I hope it's okay if I keep it the way it is :)

JoeyStk:
@lunars97 This issue should be fixed now :)
@theresantonie Very good catch. I wasn't able to fix it yet. I think this requires another refactoring, but that is probably for another issue :see_no_evil:

lunars97:
I am still getting a slightly left shifted dropdown hmm :thinking: 
image

lunars97:
> I am still getting a slightly left shifted dropdown on Chrome (Firefox works perfectly) hmm 🤔 image

It works perfectly :+1:  Sorry the above comment is not relevant anymore. The problem was that I did not switch to remote branch.

lunars97 (approved):
Thank you for your PR! :+1:

PR #2938, opened 2024-07-30
Change text and display of in total in broken link row on dashboard

JoeyStk:
### Short description

This PR changes the text and hide the "in total" string in the broken link row on our dashboard while waiting for the broken link row to load.

### Proposed changes

- Hide all "in total" strings
- Show those of the other rows again
- Show "in total" of broken link row once it has been loaded

### Side effects

- I'm a bit insecure about doing the toggling of the "in total" strings. We have to do it this way (I think) due to the Django template that we use. I'm unsure, because there might be municipalities that have JS disabled and then this would look quite messy (I think). Plus the code doesn't look great

### Resolved issues

Fixes: -

### DIFF
--- a/integreat_cms/static/src/js/dashboard/broken-links.ts
+++ b/integreat_cms/static/src/js/dashboard/broken-links.ts
@@ -16,7 +16,20 @@ const getContent = async (url: string): Promise<Content> => {
     return response.json();
 };
 
+const showAllTotalNumbers = () => {
+    const elements = document.querySelectorAll<HTMLElement>(".total-results");
+
+    elements.forEach((element) => {
+        if (!element.closest("#broken-links")) {
+            const el = element;
+            el.classList.remove("hidden");
+        }
+    });
+};
+
 window.addEventListener("load", async () => {
+    showAllTotalNumbers();
+
     const brokenLinksElement = document.getElementById("broken-links");
 
     if (!brokenLinksElement) {
--- a/integreat_cms/cms/templates/dashboard/_todo_dashboard_row.html
+++ b/integreat_cms/cms/templates/dashboard/_todo_dashboard_row.html
@@ -18,7 +18,7 @@
             {# djlint:off #}<a href="{% spaceless %}{% block todo_dashboard_title_link %}{% endblock todo_dashboard_title_link %}{% endspaceless %}"
             class="font-bold text-blue-500 underline">{% spaceless %}{% block todo_dashboard_title %}{% endblock todo_dashboard_title %}{% endspaceless %}</a>{# djlint:on #}
             {% block todo_dashboard_number %}
-                <span class="total-results">({% translate "in total" %} <span>{{ total }}</span>)</span>
+                <span class="hidden total-results">({% translate "in total" %} <span>{{ total }}</span>)</span>
             {% endblock todo_dashboard_number %}
             <p>
                 {% block todo_dashboard_description %}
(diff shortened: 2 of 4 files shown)


JoeyStk:
This is not super-duper urgent, but we want to release a new version today and for that release this would be important.
Otherwise we would need to hide this row and postpone it until the next release

MizukiTemma (approved):
Thank you for quick fix 😸

integreat_cms/static/src/js/dashboard/broken-links.ts
```diff
@@ -16,7 +16,20 @@ const getContent = async (url: string): Promise<Content> => {
...
+        if (!element.closest("#broken-links")) {
+            const el = element;
+            el.classList.remove("hidden");
+        }
+    });
+};
+
 window.addEventListener("load", async () => {
+    showAllTotalNumbers();
+
```
PeterNerlich, line 32: Wouldn't it make much more sense to change the definition of `updateNumberOfPages()` below instead? It is the one updating the total once the data has been retrieved. Putting `showAllTotalNumbers()` at the top of the `load` event listener instead removes the class `hidden` on page load, before we even start fetching the auxiliary data.

I would propose to entirely remove the definition of `showAllTotalNumbers()` in favour of changing `updateNomberOfPages()` as follows:

```diff
     const updateNumberOfPages = (numberOfPages: number) => {
-        (brokenLinksElement.querySelector(".total-results span") as HTMLElement).innerText = numberOfPages.toString();
+        const results = brokenLinksElement.querySelector(".total-results") as HTMLElement;
+        (results.querySelector("span") as HTMLElement).innerText = numberOfPages.toString();
+        results.classList.remove("hidden");
     };
```

PeterNerlich (changes requested):

integreat_cms/static/src/js/dashboard/broken-links.ts
```diff
@@ -16,7 +16,20 @@ const getContent = async (url: string): Promise<Content> => {
...
+    const elements = document.querySelectorAll<HTMLElement>(".total-results");
+
+    elements.forEach((element) => {
+        if (!element.closest("#broken-links")) {
+            const el = element;
+            el.classList.remove("hidden");
+        }
+    });
+};
+
```
deen13, line 29: I can only find one element with the class total-results, iiuc we should be able to query one element and don't need to iterate over a list of elements.

deen13 (changes requested):

integreat_cms/static/src/js/dashboard/broken-links.ts
PeterNerlich: This suggestion is invalid and I have not read the originally suggested changes attentively enough

integreat_cms/static/src/js/dashboard/broken-links.ts
JoeyStk: Hm, I'm not sure that works. All six rows should contain the class "total-results" for the string "(Insgesamt x)". I want to display five out of the six rows, just the one for broken-links not

deen13 (approved):

PeterNerlich (approved):

PeterNerlich:
Weird, hiding the totals element on DOMContentLoaded like this instead works for me in both Chromium and Firefox. Not sure what the weird issue we saw earlier was.

```ts
window.addEventListener("DOMContentLoaded", () => {
    (document.querySelector("#broken-links .total-results") as HTMLElement).classList.add("hidden");
});
```

JoeyStk:
Would you like more to refactor it this way? Do you consider this way to be cleaner? :)

Issue #2943, opened 2024-07-30
Content editor: Trailing whitespace in translate="no" collapsed


PeterNerlich:
### Describe the Bug

When text in the editor is marked with `translate="no"` and the `span` element contains the whitespace after it, but the next character after the closing `span` tag is not a space, the space is collapsed and both words appear pressed right up to each other.

This is a regression introduced in 2024.7.1 by #2873 and caused by the addition of `display: inline-block` in order to be rendered with the set `dir="ltr"` order even when inside contexts rtl paragraphs. However, since this makes the element behave like a separate block in certain ways, leading and trailing whitespace is ignored, disrupting the presentation of the line as in this case.

### Steps to Reproduce

1. Go to edit any page or content edited with TinyMCE
2. Replace the source with the following to observe the effect (notice the placement of the spaces):
    ```html
    <p>Diese <span class="notranslate" translate="no">Änderungen </span>wurden <span class="notranslate" translate="no">zum </span><span class="notranslate" translate="no">Review </span> vorgelegt</p>
    <p dir="rtl"><a class="link-external" href="tel:+49123456789" data-bcup-haslogintext="no"><span class="notranslate" translate="no">+49 (0) 123456789</span></a></p>
    ```
3. Go back to the WYSIWYG view
4. See collapsed spaces

### Expected Behavior

Text is rendered as:
```
| Diese Änderungen wurden zum Review  vorgelegt      |
|                                  +49 (0) 123456789 |
```

### Actual Behavior

Text is rendered as:
```
| Diese Änderungenwurden zumReview  vorgelegt        |
|                                  +49 (0) 123456789 |
```

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

PR #2944, opened 2024-07-30
Fix disappearing spaces next to Do Not Translate markers

PeterNerlich:
### Short description

Specify whitespace handling for `translate="no"` elements

### Proposed changes

- Add `white-space: pre-wrap;` to the `[translate="no"]` rule

### Side effects

- Spaces are *not* collapsed when at the beginning or end of a line if the element with `white-space: pre-wrap;` is starting/ending there
  - I don't see a better solution without implementing custom text layouting/rendering in the browser, which is likely a very bad idea
  - The impact of this is negligible, especially as it only concerns our editor in the backend

### Resolved issues

Fixes: #2943 


jarlhengstmengel (approved):
Looks good to me

Issue #2917, opened 2024-07-18
HIX calculation for pages with no real text


MizukiTemma:
### Describe the Bug

Pages with only html tags in its content are not handled as empty pages, and HIX calculation/refresh is activated and it fails to retriev HIX value.

### Steps to Reproduce

1. Opne a new page
2. Insert per "View -> Source code" html tags (example below)
3. Click the HIX update button or save the page
4. See the error in the box "Text understandability"

### Expected Behavior

Such a page is handled as empty page (and show "There is no page content")

or

Manage to fetch HIX value

### Actual Behavior

HIX retrieving is activ but fails.

### Additional Information

See this similar bug too #2916 

Exmple1:
\<div>
\<p> \</p>
\<p> \</p>
\</div>

Exmaple2:
\<div>
\<p> \</p>
\<p> \</p>
\<p> \</p>
\<h4> \</h4>
\<h4> \</h4>
\<p>\<br>\<br>\</p>
\</div>

Error like `HIX benchmark API call failed: <HTTPError 400: 'Bad Request'>` was not observed but `Failed to retrieve the hix score for <PageTranslation (.........)>`

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

MizukiTemma:
Solution from the issue grooming: do not send the text to TextLab if the page doesn not contain real text.

Issue #2916, opened 2024-07-18
Error at retrieving HIX value when the page content has only a video


MizukiTemma:
### Describe the Bug

HIX value cannot be retrieved when a page has only a video in its content.

### Steps to Reproduce

1. Go to Testumgebung in the test system.
2. Go to the page "Integreat in Gebärdensprache (Video)" 
3. See the error " HIX value could not be calculated. Please try again later."

This is locally reproducable. Copy the source code of the page and paste it in a page in the local system, you"ll see `HIX benchmark API call failed: <HTTPError 400: 'Bad Request'>`

### Expected Behavior

No error

### Actual Behavior

Error appears

### Additional Information

See also #2917 
If some words are added into the content, HIX value is retrieved and saved successfully.

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

PeterNerlich:
I think we could use the combination of two conditions to determine whether we regard a page as **empty**, only **non-textual content** or **content to evaluate by TextLab**:

- **Extract plain text – is that empty?** The lxml library features [`HtmlElement.text_content()`](https://lxml.de/apidoc/lxml.html.html#lxml.html.HtmlElement.text_content), a method that traverses **all child nodes** to extract the plain text. This is an inexpensive operation. We should trim any leading and trailing whitespace before comparing to `""` for robustness.
- **Do any non-text tags exist?** We could walk the parsed tree ourselves and check if it contains any tags that we do not associate with text. I think it makes more sense to check for anything *not expected to contain text* instead of checking for a set list of *specific tags* like `img` and `video`, because I think the latter is much harder to make an exhaustive list. This might not hold true if we find that we already limit our content to a very small subset of possible tags during our normalization step.
  Tags coming to mind quickly include: `div`, `p`, `br`, `span`, `i`, `b`, `em`, `strong`, `a`, `blockquote`, `ul`, `ol`, `li`, `mark`, `s`, `sub`, `sup`, `del`, `ins`

PR #2968, opened 2024-08-05
Fix hix for pages without text

david-venhoff:
### Short description

This pr fixes hix calculation for pages without text.

### Proposed changes

- Treat `None` as an indicator that the page has no relevant content
- If the page has no text content, don't even send a request to textlab, since we already know the result

### Side effects

Ideally none other

### Resolved issues

Fixes: #2917
Fixes: #2916 

### DIFF
--- a/integreat_cms/cms/views/utils/hix.py
+++ b/integreat_cms/cms/views/utils/hix.py
@@ -20,7 +20,7 @@ from integreat_cms.cms.models.regions.region import Region
 from integreat_cms.cms.utils.round_hix_score import round_hix_score
 
 from ....api.decorators import json_response
-from ....textlab_api.textlab_api_client import TextlabClient
+from ....textlab_api.textlab_api_client import TextlabClient, TextlabResult
 
 if TYPE_CHECKING:
     from typing import Final
@@ -40,7 +40,7 @@ class CacheMeIfYouCan(Exception):
 
 
 @lru_cache(maxsize=512)
-def lookup_hix_score_helper(text: str) -> dict:
+def lookup_hix_score_helper(text: str) -> TextlabResult:
     """
     This function returns the hix score for the given text.
     It either performs an api request or returns the value from cache,
@@ -62,7 +62,7 @@ def lookup_hix_score_helper(text: str) -> dict:
         raise CacheMeIfYouCan from e
 
 
-def lookup_hix_score(text: str) -> dict | None:
+def lookup_hix_score(text: str) -> TextlabResult | None:
     """
     This function returns the hix score for the given text.
     It either performs an api request or returns the value from cache.
--- a/integreat_cms/cms/views/utils/hix.py
+++ b/integreat_cms/cms/views/utils/hix.py
@@ -14,6 +14,8 @@ from urllib.error import URLError
 from django.conf import settings
 from django.http import JsonResponse
 from django.views.decorators.http import require_POST
+from lxml.etree import LxmlError
+from lxml.html import fromstring
 
 from integreat_cms.cms.models.pages.page_translation import PageTranslation
 from integreat_cms.cms.models.regions.region import Region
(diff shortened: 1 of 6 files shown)


integreat_cms/cms/views/utils/hix.py
```diff
@@ -47,6 +49,19 @@ def lookup_hix_score_helper(text: str) -> dict:
     :param text: The text to calculate the hix score for
     :return: The score for the given text
     """
+    # If the html has no text, don't send a request to textlab and immediately return the `None` score,
```
JoeyStk, line 52: I was taught that comments should say _what_ code is doing but instead _why_ it's are doing something. Also in this particular case I think your code is quite understandable, therefore I would either delete the comment all together or change it to answer why we don't want to send empty/None requests to TextLab :)

integreat_cms/core/signals/hix_signals.py
```diff
@@ -62,16 +62,10 @@ def page_translation_save_handler(instance: PageTranslation, **kwargs: Any) -> N
...
-            logger.debug("Storing hix score %s for %r", score, instance)
-            instance.hix_score = score
-
-            if feedback := data.get("feedback"):
-                instance.hix_feedback = json.dumps(feedback)
-
-        else:
-            logger.warning("Failed to retrieve the hix score for %r", instance)
-
+        if feedback := data.get("feedback"):
```
JoeyStk, line 68: Stupid question: Is there a difference between feedback and hix_feedback? :thinking:

integreat_cms/static/src/js/analytics/hix-widget.ts
```diff
@@ -10,9 +10,11 @@ const calculateBackgroundColor = (score: number, setOutdated: boolean): string =
 
     if (setOutdated) {
         return "rgb(16, 111, 254, 0.3)";
-    } else if (score >= hixThresholdGood) {
+    }
+    if (score >= hixThresholdGood) {
```
JoeyStk, line 14: Thanks for fixing the warnings! :grin:

integreat_cms/static/src/js/analytics/hix-widget.ts
```diff
@@ -35,9 +37,8 @@ const updateHixBar = (score: number, setOutdated: boolean) => {
 
 /**
  * Display a label depending on the current HIX state
- * States are "updated", "outdated", "no-content" and "error"
  */
-const updateHixStateLabel = (state: string) => {
+const updateHixStateLabel = (state: "updated" | "outdated" | "no-content" | "error") => {
```
JoeyStk, line 41: :+1:

JoeyStk (approved):
Thank you very much!
I thought your code was very clean and it works as expected :+1:
I only have a few nits (and compliments) :D

Dear second reviewer: I have focused in my review on the code style and didn't test the functionality thoroughly. I would be very nice if you could focus on that :blush:

integreat_cms/cms/views/utils/hix.py
```diff
@@ -37,7 +39,7 @@ class CacheMeIfYouCan(Exception):
 
 
 @lru_cache(maxsize=512)
-def lookup_hix_score_helper(text: str) -> dict:
+def lookup_hix_score_helper(text: str) -> TextlabResult:
```
deen13, line 42: It appears that this method is only called by the lookup_hix_score method. I assume this is because the method is being cached. If that’s correct, I suggest renaming the method to `lookup_hix_score_cached` or something similar

integreat_cms/cms/views/utils/hix.py
deen13: I've added a test case that is better suited for documentation purposes and, more importantly, helps safeguard us from regressions. :)

deen13 (approved):
Looking goood

integreat_cms/core/signals/hix_signals.py
david-venhoff: What do you mean by `hix_feedback`? The data dictionary can only contain a feedback key.

david-venhoff: Or do you refer to the field of PageTranslation? Then there is no difference, as `data["feedback"]` gets assigned to `instance.hix_feedback`

integreat_cms/cms/views/utils/hix.py
david-venhoff: I think the current name is fine, since it is clear from the name that it should not be called. With `lookup_hix_score_cached` this is not the case.

Issue #2952, opened 2024-08-01
Create a new model and form for contact information


MizukiTemma:
### Motivation

We are introducing a new feature "contact information". To save a set of informations like name, telephone number etc we need a new model with corresponding fields and form for that.

### Proposed Solution

Create a new model `contact` with following fields
- Title (e.g. "Integrationsbeauftragter:in")
- Name
- POI
- E-mail address
- Telephone number
- Website
- Address
- Postcode
- City
- Country
- Latitude?
- Longitude?

### Alternatives

None. We do it.

MizukiTemma:
[ Thoughts about the field region ]
We could add a field "region", it is but also possible to take "region" of the linked location, which is a mandatory input.
There are also some cases where one contact will be/is being used in multiple regions. Should it be rather "regions" as many to many field?

david-venhoff:
> There are also some cases where one contact will be/is being used in multiple regions. Should it be rather "regions" as many to many field?

How would that work? Right now, as I understand it, a contact can only have one poi, so it could only be used in one region. Also, there is currently no way to add the same contact to multiple pois. Is that a problem?

david-venhoff:
Also, have we already decided about how to represent contacts that are embedded in pages? This seems nontrivial to me, tbh and could maybe influence how the model should be laid out.

MizukiTemma:
> > There are also some cases where one contact will be/is being used in multiple regions. Should it be rather "regions" as many to many field?
> 
> How would that work? Right now, as I understand it, a contact can only have one poi, so it could only be used in one region. Also, there is currently no way to add the same contact to multiple pois. Is that a problem?

@osmers ?

osmers:
I'm very confused by this region label :D what is the purpose of these if the CMS user does not input the data?
One contact can only belong to one POI - that is ok bcs we decided that even if the same person works in two different locations, he/she likely has different job descriptions, different adresses, contact details etc

JoeyStk:
@osmers The region is just implicit information, that's required for the system to place the contact correctly. This is not label/field that is going to be shown to the user. I hope that clarifies this a bit :)

Issue #2959, opened 2024-08-01
[2024-11-07] Add contact as choice for "insert" in TinyMCE


MizukiTemma:
### Motivation

Contact information referred in contents are changed and updated centrally. This is why we implement a new model "contact". Let's give it a place where it shines.

### Proposed Solution

- Add an "contact" to the option "Insert" in TinyMCE, like Media Library and Link. 
- Where each contact is embedded, must be calculatable and handed over in `contact_form.view` (search for `# to implement later, how do we collect such pages?`).

### Additional Context

Wait for #2952

david-venhoff:
One idea how we could do it:
Represent contacts as urls and add a new url endpoint to the cms that returns contacts. When a user inserts a contact in tinymce we would just insert the url to it (plus maybe some parameters that specify which fields of the contact should be shown).
Advantage: 
- We won't have to update any pages when a contact changes
- We easily figure out in which pages contacts are used using linkcheck and prevent users from deleting contacts if they are still used
- We don't include any styling in the html and leave that up to the app

Disadvantage
- The App team would have to write some custom code to correctly render these urls as contacts in the app

PeterNerlich:
We can make non-editable elements in TinyMCE by adding the HTML attribute `contenteditable="false"`. I would propose to put the div with all that should be displayed for the content inline this way.

Issue #2958, opened 2024-08-01
[2024-11-07] Finalize permission setting for contact information


MizukiTemma:
### Motivation

So far we have kept the new feature contact information only available for CMS Team and Service Team for test purpose by a very strict permission setting. Now we make it available for users.

### Proposed Solution

Change the permission setting of the item menu in the navigation bar and eventually add permission checks in the related views.

---- Munincipality users ------------------------------------
| role | view | create | edit | archive | delete |
| ---- | ----- | ------- | ---- | -------- | ------- |
| Management    | ✅ | ✅ | ✅ | ✅ | ❌ |
| Editor                | ✅ | ✅ | ✅ | ✅ | ❌ | 
| Author              | ✅ | ✅ | ✅ | ✅ | ❌ |
| Event manager | ❌ | ❌ | ❌ | ❌ | ❌ |
| Observer           | *1 | *1 | *2 | ❌ | ❌| 

*1 only if they can edit at least one page
*2 only the contacts embedded in the pages they can edit

---- Staff users ----------------------------------------------
| role | view | create | edit | archive | delete |
| ---- | ----- | ------- | ---- | -------- | ------- |
| CMS Team          | ✅ | ✅ | ✅ | ✅ | ✅ | 
| Service Team      | ✅ | ✅ | ✅ | ✅ | ✅ | 
| App Team           | ✅ | ✅ | ✅ | ❌ | ❌ |  
| Marketing Team | ✅ | ❌ | ❌ | ❌ | ❌ | 

The principle is "If I can edit at least one page, I can view contacts and create a new one. I can edit the existing contacts that are embedded in the pages I can edit."

The table for munincipality users is made based on this principle. I put ✅ for Marketing Team × View, although they cannot edit anything, becuase it is probably useful for them to "view" as staff. 

 Delete and archive permissions are given if they can delete/archive pages.

### Alternatives

Keep this cool feature a secret among us or allow everyone to do whatever they want to with contact objects.

MizukiTemma:
The tests introduced by #3162 will probably need adjustment too.

JoeyStk:
Since we're getting close to releasing the contact feature, this issue is finally ready to be worked on

JoeyStk:
@MizukiTemma What exactly does `view_contain` contain? Can I see the contact list with `view_contain`?

charludo:
[...]

Issue #2951, opened 2024-08-01
Add a new menu item "Contact" to the navigation side bar


MizukiTemma:
### Motivation

We are introducing a new feature "contact information". As the first step, a menu item for it must be added.

### Proposed Solution

- Add a menu item "Contact" to the navigation (in this issue without link to the list, which will be implemented in a subsequent issue).
- Make this menu available only for CMS and Seavice Team.

![navigation menu](https://github.com/user-attachments/assets/d14c98b7-a1ce-4646-841a-0875283b6bed)

### Alternatives

No. We start.

### Additional Context

We implement this new feature step by step and partial implementation will come into the test and product system (otherwise we have to stop all the releases until this feature is completed). As the feature contact information must be first available for munincipalities **after** the implementation is fully completed with all tickets solved, we need to hide this (partially implemented) feature from most users by setting the permission very strict.

PR #2970, opened 2024-08-06
Add contact model

JoeyStk:
### Short description

This PR adds the initial db model for our new contact section.

### Proposed changes

- Add blueprint for model
  - fields as described in issue description
  - default fields such as created, last updated and others 
- Add migration file for it
- Add dummy contact to test data

### Side effects

- I decided for region as we want every contact to be (implictly) assigned to a region
- The relation POI to contact is 1 to 1 (as requested in the requirements)
- The relation contact to region is 1 to 1 (as requested in the requirements)
- Once this is merged, it unblocks a couple of other issues related to this feature

### Resolved issues

Fixes: #2952

### DIFF
--- a/integreat_cms/cms/models/contact/contact.py
+++ b/integreat_cms/cms/models/contact/contact.py
@@ -1,0 +1,65 @@
+from django.db import models
+from django.utils import timezone
+from django.utils.translation import gettext_lazy as _
+
+from ..abstract_base_model import AbstractBaseModel
+from ..pois.poi import POI
+from ..regions.region import Region
+
+
+class Contact(AbstractBaseModel):
+    """
+    Data model representing a contact
+    """
+
+    title = models.CharField(max_length=200, verbose_name=_("title"))
+    name = models.CharField(max_length=200, verbose_name=_("name"))
+    poi = models.ForeignKey(POI, on_delete=models.CASCADE, verbose_name=_("POI"))
+    region = models.ForeignKey(
+        Region, on_delete=models.CASCADE, verbose_name=_("region")
+    )
+    email = models.EmailField(
+        blank=True,
+        verbose_name=_("email address"),
+    )
+    phone_number = models.CharField(
+        max_length=250, blank=True, verbose_name=_("phone number")
+    )
+    website = models.URLField(blank=True, max_length=250, verbose_name=_("website"))
+    archived = models.BooleanField(
+        default=False,
+        verbose_name=_("archived"),
+        help_text=_("Whether or not the location is read-only and hidden in the API."),
+    )
+    last_updated = models.DateTimeField(
+        auto_now=True,
+        verbose_name=_("modification date"),
+    )
+    created_date = models.DateTimeField(
+        default=timezone.now, verbose_name=_("creation date")
+    )
+
+    def __str__(self) -> str:
+        """
+        This overwrites the default Django :meth:`~django.db.models.Model.__str__` method which would return ``Contact object (id)``.
+        It is used in the Django admin backend and as label for ModelChoiceFields.
+
+        :return: A readable string representation of the contact
+        """
+        return f"{self.title} {self.name}"
+
+    def get_repr(self) -> str:
+        """
+        This overwrites the default Django ``__repr__()`` method which would return ``<Contact: Contact object (id)>``.
+        It is used for logging.
+
+        :return: The canonical string representation of the contact
+        """
+        return f"<Contact (id: {self.id}, title: {self.title}, name: {self.name}, region: {self.region.slug})>"
+
+    class Meta:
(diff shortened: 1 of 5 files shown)


integreat_cms/cms/models/contact/contact.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class Contact(AbstractBaseModel):
+    """
+    Data model representing a contact
+    """
+
+    title = models.CharField(max_length=200, verbose_name=_("title"))
+    name = models.CharField(max_length=200, verbose_name=_("name"))
+    poi = models.ForeignKey(POI, on_delete=models.CASCADE, verbose_name=_("POI"))
```
MizukiTemma, line 17:
```suggestion
    poi = models.ForeignKey(POI, on_delete=models.PROTECT, verbose_name=_("POI"))
```
I would make this protected from deletion for safety 🦺

MizukiTemma:
@JoeyStk 
Is the list comming together in this PR or will it be implemented separately?

JoeyStk:
@MizukiTemma I was undecided about this yesterday. How about we merge this without the list and once #2951 is merged, I will open a new PR for the list? :)

MizukiTemma:
> @MizukiTemma I was undecided about this yesterday. How about we merge this without the list and once #2951 is merged, I will open a new PR for the list? :)

Ok 😉

david-venhoff (commented):
Not a full review yet, but I would suggest to also include permissions for the model here. Otherwise we would have to introduce another migration

integreat_cms/cms/models/contact/contact.py
JoeyStk: > I would make this protected from deletion for safety 🦺
It would be very painful if a contact which had been embedded everywhere were deleted😭 and we should not expect users are well aware of this denger 😅

Good point! It's adjusted now, but I need a bit more time to fix the test

integreat_cms/cms/models/contact/contact.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    Data model representing a contact
+    """
+
+    title = models.CharField(max_length=200, verbose_name=_("title"))
+    name = models.CharField(max_length=200, verbose_name=_("name"))
+    poi = models.ForeignKey(POI, on_delete=models.PROTECT, verbose_name=_("POI"))
+    region = models.ForeignKey(
+        Region, on_delete=models.CASCADE, verbose_name=_("region")
+    )
```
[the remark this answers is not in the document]
JoeyStk: As we just foudn out we we're talking about different meanings of "implicit". Now I understand what you meant and adjusted it (but didn't push it yet)

JoeyStk:
In many commits I was able to fulfill all of your suggestions. Please feel free to review again :)

integreat_cms/cms/views/regions/region_actions.py
```diff
@@ -51,7 +51,7 @@ def delete_region(
         messages.error(
             request,
             format_html(
-                "{}<ul class='list-disc pl-4'>{}</ul>",
```
david-venhoff, line 54: It probably makes sense to undo this change, since it is unrelated.
I am in favor of enforcing a consistent order of attributes, but that should probably be done in another pr for the whole repository.

david-venhoff (approved):
Thanks, I think this looks good now.

One thing to note is that we will almost certainly have to change the model again soon, to implement the remaining contact related issues (for example #2959), which feels a bit weird to me, but I don't think it makes sense to delay this further either :/

integreat_cms/cms/constants/roles.py
```diff
@@ -87,6 +87,9 @@
 AUTHOR_PERMISSIONS: Final[list[str]] = EVENT_MANAGER_PERMISSIONS + [
     "change_page",
     "view_page",
+    "change_contact",
+    "delete_contact",
+    "view_contact",
```
MizukiTemma, line 92: In the current setting authors, editors and managers can access contacts but none of the staff users, and we need to restrict access to this new feature to CMS + Service Team or CMS + the product owner until everything is finished and we finalize the permission settings in #2958

MizukiTemma (changes requested):
Looks almost good 💪 I have only one concern to the permission settings in context of hiding this feature to most of users until completed.

integreat_cms/cms/constants/roles.py
JoeyStk: I think there was some misunderstanding between David and me beforehand that caused this. However, I think this is resolved now, and therefore I changed the permissions for only service-team and cms-team to be able to access the contact model

MizukiTemma (approved):
Thank you so much 🎉

Issue #2973, opened 2024-08-07
Error when trying to create a POI from the event form


osmers:
### Describe the Bug
A server error 500 occurs when trying to create a POI from the event from

### Steps to Reproduce

1. Go to or create any event
2. Type the name for a POI and click on create
3. See error

### Expected Behavior
Form to create POI should open

### Actual Behavior
Server error 500

### Additional Information
![image](https://github.com/user-attachments/assets/fb3ea9bf-ea19-4ca5-a010-28fa624c77c7)

PR #2974, opened 2024-08-07
Fix error at creating POI in event form

MizukiTemma:
### Short description

This PR fixes Error 500 which occurs during location creation on the event form. The cause is the queryset introduced by #2932 

### Proposed changes

- Pass the current region as `additional_instance_attributes` so organizations of the region can be found without problem

### Side effects

- I hope I'm not bringing any bug this time 🙈 

### Resolved issues

Fixes: #2973 

### DIFF
--- a/integreat_cms/release_notes/current/unreleased/2973.yml
+++ b/integreat_cms/release_notes/current/unreleased/2973.yml
@@ -1,0 +1,2 @@
+en: Fix error of POI creation in event form
+de: Behebe Fehler beim Erstellen eines Ortes im Veranstaltungformular
--- a/integreat_cms/cms/views/pois/poi_form_ajax_view.py
+++ b/integreat_cms/cms/views/pois/poi_form_ajax_view.py
@@ -33,7 +33,7 @@ class POIFormAjaxView(TemplateView, POIContextMixin):
         :param \**kwargs: The supplied keyword arguments
         :return: The html template of a POI form
         """
-        poi_form = POIForm()
+        poi_form = POIForm(additional_instance_attributes={"region": request.region})
         poi_title = kwargs.get("poi_title")
         poi_translation_form = POITranslationForm(data={"title": poi_title})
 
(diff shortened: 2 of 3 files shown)


integreat_cms/release_notes/current/unreleased/2973.yml
```diff
@@ -0,0 +1,2 @@
+en: Fix error of POI creation in event form
+de: Behebe Fehler bei der POI-Erstellung im Veranstalrungformular
```
JoeyStk, line 2:
```suggestion
de: Behebe Fehler beim Erstellen eines Ortes im Veranstaltungformular
```

JoeyStk (approved):
Thank you very much for the quick fix!
Do you think it makes sense to add a test to make sure this bug doesn't happen again?
(If needed in a separate PR) :)

MizukiTemma:
> Thank you very much for the quick fix! Do you think it makes sense to add a test to make sure this bug doesn't happen again? (If needed in a separate PR) :)

Yes, test is always good 👍 To merge this PR ASAP I'll create an separate issue for the test.

david-venhoff (approved):
LGTM
+1 for adding a test

PeterNerlich:
The failing tests…
```
ERROR tests/cms/utils/test_disable_hix_post_save_signal.py::test_disable_hix_post_save_signal - django.db.utils.IntegrityError: duplicate key value violates unique constraint "cms_language_slug_key"
DETAIL:  Key (slug)=(du) already exists.
ERROR tests/cms/utils/test_disable_hix_post_save_signal.py::test_rule_out_false_positive - django.db.utils.IntegrityError: duplicate key value violates unique constraint "cms_language_slug_key"
DETAIL:  Key (slug)=(du) already exists.
```
…are @JoeyStk's and my fault *(arguably mostly mine)*. They should be sporadic and occur only when the two tests we introduced in #2933 are executed by the same container. I couldn't think of a better way to cleanly do it at the time, but it's apparent we need a proper fix or we'll make our tests more annoying again.

MizukiTemma:
@PeterNerlich 
Do you have any idea for quick fix?

PeterNerlich:
@MizukiTemma Bad but immediate fix: Mark the tests to be skipped. Short term fix: Sorry, still have to think about it. Long term fix: [Fix django so that we have proper database rollback](https://github.com/django/django/pull/17998) *(I think this applies here as well)*.

MizukiTemma:
@PeterNerlich 
Thank you 👍 I have disabled the tests so this PR (and others following in near future too) won't be blocked by failing tests. An issue for long term solution is opened, see #2978 .

Issue #2939, opened 2024-07-30
Include integreat-cms.ini in DevContainer


jarlhengstmengel:
### Motivation
The `integreat-cms.ini` file allows for customization of settings and environment variables as well as setting keys for external tools etc. on your local system or development environment. Currently the devcontainer doesn't include this file on it's usual location at `/etc/integreat-cms.ini` or the location defined with the environment variable `INTEGREAT_CMS_CONFIG`. This can create some problems if you need to work with some additional functionality like enabling the reddis-cache or while customizing folders/file paths etc.

### Proposed Solution
Include the path to the 'integreat-cms.ini' file into the devcontainer, probably with checking if the file exists.

### Alternatives
Creating the file manually inside the devcontainer, which can be quite frustrating and error prone.

### User Story
As a developer I want to have a quick and straightforward way to set and enable all variables, keys and features so that I can start or go on quickly and smoothly with actual developing.

### Additional Context
Possible way to solve this may be with using a '.dev' file as described [here](https://stackoverflow.com/questions/58227845/how-can-i-use-an-env-file-to-pass-environment-variables-into-a-standalone-vscode).

PR #2976, opened 2024-08-07
Add cms configuration file to devcontainer

jarlhengstmengel:
### Short description
This PR adds the configuration file `integreat-cms.ini` to the development container so it can be used.

### Proposed changes

- Pass path of the configuration file via the environmental variable `INTEGREAT_CMS_CONFIG`, this ensures the devcontainer to work even when the variable isn't set (in that case it's passed as blank)
- Add Information about using configuration file to README

### Side effects
- Hopefully none

### Resolved issues

Fixes: #2939 

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -2,11 +2,15 @@ version: "3.8"
 
 services:
   devcontainer:
+    env_file: ./.env
     build:
       context: .
       dockerfile: Dockerfile
     volumes:
       - ..:/workspace:cached
+      - type: bind
+        source: ${INTEGREAT_CMS_CONFIG:-}
+        target: /etc/integreat-cms.ini
     network_mode: service:db
     depends_on:
       - db
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ To configure your development container, please follow these steps carefully.
 4. Open the command palette (Ctrl + Shift + P or Cmd + Shift + P on macOS) and search for "> Remote-Containers: Open Folder in Container".
 5. VSCode will open the project in a new container, install all further required tools and load the testdata.
 
+
 #### Known Limitations
 
 The perfect is the enemy of the good; thus, this section illuminates aspects of our evolving development setup.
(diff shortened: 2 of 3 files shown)


.devcontainer/docker-compose.yml
```diff
@@ -2,11 +2,15 @@ version: "3.8"
 
 services:
   devcontainer:
+    env_file: ./.env
```
deen13, line 5: This works for me using the `environment` block without a default value, even though I haven't set the `INTEGREAT_CMS_CONFIG` variable in my shell. If we need a default value, we can also set it directly in the Compose file using the `:-` notation.

For example:
```yaml
environment:
  - INTEGREAT_CMS_CONFIG=${INTEGREAT_CMS_CONFIG}
```

or with default value

```yaml
environment:
  - INTEGREAT_CMS_CONFIG=${INTEGREAT_CMS_CONFIG:-default}
```

deen13 (approved):
Thanks for your work on the devcontainer. 🥳

.devcontainer/docker-compose.yml
jarlhengstmengel: Thanks for the hint, A thought I have is if it would make sense to have the `.env` file to have one place to pas environment variables in general in case someone would use it that way. But I don't really know how far a devcontainer should be individualized that way.
It should be fine without a default value. If the variable isn't set, it will automatically be set to blank inside the container.

README.md
```diff
@@ -91,6 +92,17 @@ and then start your code editor (`code .`, `nvim`,...) from within that same she
...
+All these settings can also be configured via environment variables with the prefix `INTEGREAT_CMS_`, e.g. `INTEGREAT_CMS_SECRET_KEY`.
+
+#### Using a CMS configuration file with devcontainer
+
+If you want to use a configuration file in your development container, pass the absolute path of the configuration file with setting the environment variable `INTEGREAT_CMS_CONFIG`. All paths set in the configuration file must exist inside the workspace. If you want to configure the settings with environment variables you need to pass them via the `.devcontainer/.env` file.
+
+#### Known limitations
+When using the example file as starting point for configuration, all variables setting ports need to be actual numbers, even without setting the rest of the respective section.
```
charludo, line 104: I am a bit confused by this sentence - does it just mean "all ports need to be defined"?

charludo:
I don't really have anything to add, but also can't really fully test this. So I'm a bit hesitant to give Approval, maybe there's someone else who uses the devcontainer setup who can do this? 🙈

charludo (approved):

Issue #2076, opened 2023-02-18
Not-Activated User Accounts should be deleted after 7 Days automatically


dkehne:
### Describe the Bug
To be conform with data privacy not-activated user accounts should be deleted after a time frame of 7 days (168 hours).

svenseeberg:
I guess we need a management command that cleans up expired registrations. We can then call the management command with a systemd timer / cron job.

PR #2980, opened 2024-08-08
Automatically remove user accounts that remain unactivated for 7 days

deen13:
### Short Description

This pull request introduces a new management command designed to delete user accounts that have not been activated for over a week. It also includes a cron job that executes this command every hour to ensure regular checks of user accounts.

### Proposed Changes
- Implement a management command to delete inactive user accounts older than one week.
- Set up a cron job that runs every hour to invoke the management command.
- Configure the `django_cron` app.

### Side Effects

- This feature may result in the deletion of a large number of inactive accounts upon release.

### Testing

To test this feature, follow these steps:
- Create a new user account.
- Access the Django admin panel and set the account's creation date to one week ago.
- Run the cron job using the command: `./tools/integreat-cms-cli runcrons "integreat_cms.core.jobs.inactive_user_cleanup_job.InactiveUserCleanupJob" --force`.
- Verify that the user account has been deleted.

### Resolved Issues

Fixes: #2076


integreat_cms/release_notes/current/unreleased/2076.yml
```diff
@@ -0,0 +1,2 @@
+en: A background job has been introduced that deletes unactivated accounts after 7 days.
+de: Es wurde ein Hintergrundjob eingeführt, der nicht aktivierte Accounts nach 7 Tagen löscht.
```
JoeyStk, line 2: Nit: We had agreed in the past to write release notes in imperative form. So in this case it would be something like

```suggestion
de: Führe Hintergrundjob ein, der nicht aktivierte Accounts nach 7 Tagen löscht
```

JoeyStk (approved):
Thank you very much! I'm not an expert on this, but I think this looks good!

integreat_cms/release_notes/current/unreleased/2076.yml
deen13: Thank you @JoeyStk :)

JoeyStk:
@ulliholtgrave had some concerns about introducing a new dependency that does something we already do with a different dependency for the push notifications. @timobrembeck / @david-venhoff Do you remember why we chose the way via a cronjob on the server and have not used django_cron? Which way do you think is better? :)

deen13:
Do we have an existing infrastructure-as-code setup where we could define these native cron jobs (if we do it that way)? That would be ideal!

david-venhoff:
> Do we have an existing infrastructure-as-code setup where we could define these native cron jobs (if we do it that way)? That would be ideal!

Currently we define cron jobs via the salt-config. For example: https://git.tuerantuer.org/DF/salt/src/commit/ed1591855a53ebe88dd387b4e21b0cd03265d539/states/integreat-cms/init.sls#L240-L247

I am not sure what the advantages of using `django_cron` are, but I think we should either
- not use `django_cron` and define everything via salt as we did before or
- use `django_cron` if we decide that this is the better solution but then do that in a separate pr and include all other cron jobs that are currently defined in the salt config

deen13:
Hey @david-venhoff, I've outlined some pros and cons for your consideration:

#### Pros
- The cron job is defined close to the business logic, which simplifies the overall structure.
- Cron job activation aligns with rollouts, streamlining the release process.
- Developers can write scheduled tasks without needing access to our Salt repository or any prior Salt expertise.

#### Cons
- This approach deviates from our current sensible default.
- The library appears to be poorly maintained ([link](https://github.com/Tivix/django-cron)).

Let me know what you think!

david-venhoff:
Hmm, I don't really want to switch to a library that did not have a commit in 2 years. So I'd prefer to keep our current solution and use salt for cron jobs. However I still think that a library could have value, so if you think the poor maintenance is not such a big problem then we could probably talk about it together in the next weekly?

deen13:
Yeah I agree on that @david-venhoff. I'll add a cron job to the salt config.

pyproject.toml
```diff
@@ -75,7 +75,7 @@ dependencies = [
     "webauthn",
     "xhtml2pdf==0.2.8",
     "reportlab<4.0.0",
-    "zammad-py",
+    "zammad-py"
```
charludo, line 78: Interesting, I assume this somhoe comes from adding then removing `django-cron`? 

Just checked locally, and black/ruff don't seem to care about the comma either way, whereas they are opinionated about them everywhere else in the project. Huh.

I can't believe how nitpicky this is, but since all *other* lists in `pyproject.toml` end with a comma,... 😅 😂

Issue #2982, opened 2024-08-08
`replace_links` does not replace links of Imprints


david-venhoff:
### Describe the Bug

When we implemented #2913, we forgot to update `replace_links` to include ImprintTranslations, which means that we cannot replace links there.

### Steps to Reproduce

1. Create a link in an imprint
2. Use `replace_links` to replace the url with another url
3. Notice that nothing happens

PR #2983, opened 2024-08-09
Include links of imprints into link replacement

MizukiTemma:
### Short description

This PR adds `ImprintPageTranslation` into the target models of `replace_links` so links in imprints can be replaced too by the "Search&Replace" on the uppr right corner of the broken link list.

### Proposed changes

- Add `ImprintPageTranslation` into the list `models` in the function `replace_links`

### Side effects

- Should we probaly think of centralised check for such process of adding/removing a model to/from our link features? Somthing like, listing models in one single list (as an environment variable, for exmaple) and we add a check whether all models of the list are included at the points where you need to add changes while removing/adding a model to the link check feature.

### Resolved issues

Fixes: #2982

### DIFF
--- a/integreat_cms/release_notes/current/unreleased/2982.yml
+++ b/integreat_cms/release_notes/current/unreleased/2982.yml
@@ -1,0 +1,2 @@
+en: Include links in imprints for link replacement
+de: Beziehe Links in Impressen in der Link-Ersetzung ein
--- a/integreat_cms/cms/utils/linkcheck_utils.py
+++ b/integreat_cms/cms/utils/linkcheck_utils.py
@@ -252,7 +252,7 @@ def replace_links(
         region_msg,
         user_msg,
     )
-    models = [PageTranslation, EventTranslation, POITranslation]
+    models = [PageTranslation, EventTranslation, POITranslation, ImprintPageTranslation]
     with update_lock:
         for model in models:
             filters = {}


integreat_cms/release_notes/current/unreleased/2982.yml
```diff
@@ -0,0 +1,2 @@
+en: Include links in implints in link replacement
+de: Beziehe Links in Implikationen in Link-Ersetzung ein
```
david-venhoff, line 2:
```suggestion
en: Include links in imprints for link replacement
de: Beziehe Links in Impressen in der Link-Ersetzung ein
```

david-venhoff (approved):
Thanks!

> Should we probaly think of centralised check for such process of adding/removing a model to/from our link features? Somthing like, listing models in one single list (as an environment variable, for exmaple) and we add a check whether all models of the list are included at the points where you need to add changes while removing/adding a model to the link check feature.

I think this is a good idea. For this case though I think it would be best to just completely rewrite the `replace_links` function. Over time, it has become quite messy and now we have the `get_region_links` function that we could use instead to make `replace_links` both faster and cleaner.
Do you want to do that in this pr? Otherwise I could also do that in a follow-up pr.

MizukiTemma:
@david-venhoff Thank you for review and idea 😸 I'd suggest to do it in another PR to stay in the policy "one PR, one purpose".

deen13 (approved):
👍 😄

Issue #2807, opened 2024-05-16
Translation with status "maschine translated" outdates after date refreshing


MizukiTemma:
### Describe the Bug

Translations with status "maschine translated" are "outdated" after refreshing the date of their source translation.

### Steps to Reproduce
1. Set two languages (for example English and Arabic) directly hanging from German (root language) in the language tree of a region.
2. Create a German translation.
3. Create an Arabic translation by MT, and create an English translation manually (not by MT)
4. See in the page tree that the German and English traslations are up-to-date (✅) and the Arabic one maschine-translated (🤖)
5. Open the German translation and click the button "Mark this page as up-to-date"
6. Go back to the page tree, see the German and English translations are up-to-date but the Arabic one is marked as outdated (⚠️)

### Expected Behavior

The Arabic translation should be up-to-date too.

### Actual Behavior

The Arabic translation is outdated.

### Additional Information

👀 The cause seems to be the order in which the translations are refreshed: after this reproduction procedure the German translation (version 2) and the English one (version 1) have time stamp of 16:04:12, where as the Arabic one 16:03:38.

💡It is recommended to implement/review this issue with redis activated.

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

lunars97:
@MizukiTemma seems like it should be like that because when we update the source language without ticking an automatic translation and updating it, the machine translated stays unchanged and therefore it outdates if the source translation was updated without ticking an automatic translation. Shouldn't it behave like that?

osmers:
No, Mizuki is right bcs we are not talking about the update button in the top right hand corner but about the "mark page as up to date" button (further down in the settings of the page) which presumes that nothing was changed, all information is up to date and only should update the date on all pages that are marked as up to date (with a tick in the translation status).

MizukiTemma:
@osmers Thank you for explaining 👍 

@lunars97 Sorry, I have overlooked your question 😿 Yes, it about the button "Mark this page as up-to-date" in the side box "Actions"

PR #2984, opened 2024-08-12
Fix machine translations being considered as not up-to-date

charludo:
### Short description

When marking pages as up-to-date, their machine-translated translations where changed to be out-of-date. This PR fixes this.

### Proposed changes

- in `AbstractContentTranslation`, change the definition of "up to date" to include machine translations
- surprisingly, this breaks 0 tests 🤔 

### Side effects

- none, but I think we might need a discussion about outdated/up-to-date definitions. Right now, there are multiple `translation_status` states which are neither up-to-date nor outdated, which is a bit weird.

### Resolved issues

Fixes: #2807

### DIFF
--- a/integreat_cms/cms/models/abstract_content_translation.py
+++ b/integreat_cms/cms/models/abstract_content_translation.py
@@ -441,11 +441,14 @@ class AbstractContentTranslation(AbstractBaseModel):
     def is_up_to_date(self) -> bool:
         """
         This property checks whether a translation is up to date.
-        A translation is considered up to date when it is not outdated and not being translated at the moment.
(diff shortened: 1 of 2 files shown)


MizukiTemma (approved):
Looks good 😸 Thank you!

david-venhoff (approved):
Thanks, LGTM

Could you also update the docstring of the method? (Unfortunately github won't let me create a suggestion for it)

charludo:
> Could you also update the docstring of the method? (Unfortunately github won't let me create a suggestion for it)

Oh right, thanks! Is done now.

Issue #2369, opened 2023-07-25
Enable link checker for organization website URLs


timobrembeck:
### Motivation

When the website links of organizations are no longer working, we don't notice it

### Proposed Solution

Include the website fields of urls in the link checker

### Alternatives

Don't do it

### User Story

As a region manager, I want to know when the URL of one of my organizations is no longer working

MizukiTemma:
Do not forget adding it to `replace_link` too. See #2982

juliankehne:
@timobrembeck @osmers

I tested this and have some feedback:

1. Inconsistency across user roles: As far as I know only the manager role should be able to edit organizations. With this feature other roles with access to the Broken Links can change the organization URL from the Broken Links overview, e.g. the editor role.
2. In the Broken Link overview you can click on the source of an URL to get redirected to the page its used in. For an organization URL I would expect it to do the same and redirect to the Organization, at least for the manager role as only they can access organizations. Currently nothing happens and there is no visual indication why it doesn't work, e.g. tooltip when hovering over it.
![2024-10-07 13_06_03-Window](https://github.com/user-attachments/assets/f44346d6-0ecd-47a5-b85f-5cbaa849ba45)
3. There is no clear indication in the UI that an URL belongs to an organization

I'm not 100% sure about this, but the URLs from the imprint behave similar, also regarding the user role that should be able to edit the imprint.

Otherwise the organization URLs do show up in the right categories (valid, invalid, etc.) and can be changed in the Broken Links page directly. Ignoring, Unignoring and Rechecking do work with bulk execution.

osmers:
> 1. Inconsistency across user roles: As far as I know only the manager role should be able to edit organizations. With this feature other roles with access to the Broken Links can change the organization URL from the Broken Links overview, e.g. the editor role.

That's a very good catch. Needs to be changed for the imprint as well.

PR #2985, opened 2024-08-12
Add organization URLs to link check

MizukiTemma:
### Short description

This PR adds URLs of organizations to our link check/replace feature.

The issue is planned not for Q3 but Q4, but in my opinion it's better to implement this before we start refactoring of link check/replace feature for performance and the centralised link management system.

### Proposed changes

- Add the model `Organization` to the link checker
- Add a new function `replace_urls` to the model `Organization` to be compatible with the refactoring from #3003 
- <s>Adjust functions for link replacement (because organization objects do not have any translations)</s>

### Side effects

- <s>I hope it functions at least. We definately need some refactoring (but suggest to do it in another PR)</s> It's merged now :)

### Resolved issues

Fixes: #2369 

### DIFF
--- a/integreat_cms/cms/utils/linkcheck_utils.py
+++ b/integreat_cms/cms/utils/linkcheck_utils.py
@@ -14,6 +14,7 @@ from linkcheck.models import Link, Url
 from integreat_cms.cms.models import (
     EventTranslation,
     ImprintPageTranslation,
+    Organization,
     PageTranslation,
     POITranslation,
     Region,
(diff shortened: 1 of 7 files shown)


MizukiTemma:
Better to wait for #3003

integreat_cms/cms/utils/linkcheck_utils.py
```diff
@@ -252,9 +255,31 @@ def replace_links(
...
+    other_models = [Organization]
     with update_lock:
-        for model in models:
+        for model in other_models:
+            objects = model.objects.all()
+            if region:
+                objects = objects.filter(region=region)
+            for obj in objects:
+                url = obj.website
+                print(obj.links)
```
charludo, line 272:
```suggestion
```

😁

charludo (dismissed):
Thank you! 🎉 
As far as I can tell this looks good - I am not super fond of the split into `translation_models` and `other_models` and the `isinstance()` pattern, but I don't know any other way to do this either. Maybe a thing to keep an eye on for the mentioned refactor :D

Would it be possible to add a broken organization link to the test data?

timobrembeck (approved):
Thanks a lot, works like a charm! :tada:

david-venhoff (approved):
LGTM!

I would just propose to rename the `find_target_url_per_translation` function to something like `find_target_url_per_content`

PR #2997, opened 2024-08-14
Allow MT only when user can publish

### LINKED ISSUE #2961: Allow MT only when user has `publish_page` right
### Describe the Bug

Author is an user role with no publishing right that can only save pages for review. Currently users with role author can automatically traslate pages. That means, users who cannot finalize pages can translate. This does not make sense (if a page is not approved, it must be changed and trasnlated again). MT should be available only when the user has `publish_page` right.

### Steps to Reproduce

1. Log in as an author user
2. Go to a page
3. See you can neither save as draft nor publish the page but translate by MT.

### Expected Behavior

MT is available only if the user can publish the page

### Actual Behavior

MT is available even if the user can only submit for review

### Additional Information

This "submit for review" feature exists only for pages. 
MT should be available also for author users, if they are either listed as editor (page base right) or a member of the organization responsible for the page (access per organization membership).

For a suggested solution see [here](https://github.com/digitalfabrik/integreat-cms/pull/2997#pullrequestreview-2257122206)

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

MizukiTemma:
### Short description

This PR makes our MT feature available only for users who are allowed to publish the target contents.

### Proposed changes

- Change `is_permitted` to check whether the current user can "publish" the content(s) (required: `can_change` for Event and POI as before, `can_publish` for Page either per role >= Editor or per page based permission or as organization member)
- Deaktivate MT bulk action completely for observers and authors for pages

### Side effects

- `is_permitted`, `DeeplApiClient` and `GoogleTranslateApiClient` are extended with a variable. I hope this won't affect other parts, for it is "optional" with default value
- 

### Resolved issues

Fixes: #2961 

### DIFF
--- a/integreat_cms/cms/forms/machine_translation_form.py
+++ b/integreat_cms/cms/forms/machine_translation_form.py
@@ -15,7 +15,7 @@ if TYPE_CHECKING:
 
 from ...core.utils.machine_translation_provider import MachineTranslationProvider
 from ...textlab_api.utils import check_hix_score
-from ..models import LanguageTreeNode
+from ..models import LanguageTreeNode, PageTranslation
 from .custom_content_model_form import CustomContentModelForm
 
 logger = logging.getLogger(__name__)
@@ -63,8 +63,13 @@ class MachineTranslationForm(CustomContentModelForm):
             )
             return
 
+        content = (
+            self.instance.page
+            if self._meta.model == PageTranslation and self.instance.id
+            else None
+        )
         if not MachineTranslationProvider.is_permitted(
-            self.request.region, self.request.user, self._meta.model
+            self.request.region, self.request.user, self._meta.model, content
         ):
             return
 
@@ -134,6 +139,9 @@ class MachineTranslationForm(CustomContentModelForm):
         language_nodes = self.cleaned_data["mt_translations_to_create"].union(
             self.cleaned_data["mt_translations_to_update"]
         )
+        content = (
+            self.instance.page if isinstance(self.instance, PageTranslation) else None
+        )
         if commit and language_nodes and check_hix_score(self.request, self.instance):
             for language_node in language_nodes:
                 logger.debug(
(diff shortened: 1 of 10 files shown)


deen13:
Hey @MizukiTemma, can you provide a step-by-step guide on how to test this pull request? That'd be great. 🤗

MizukiTemma:
@deen13 
Thank you for your interest in this confusing PR 😸 

I think it's the best to log in as different roles and see whether MT is available.

-----The wished state-----
Events and locations: if you have permission "change_event"/"change_poi" you can use MT both by bulk action and check box in the form
Pages: if you have permission "publish_page", you can use MT both by bulk action and in the form.

-----The tasty point-----
Observers have neither "change_page" nor "publish_page" and authors have "change_page" but no "publish_page". As they are not allowed to finalize/publish any page, MT should not be available for them.
However, they can be entatiled for "publishing" certain pages if they are lised as editors in the area "Additional permissions for this page" in the form or their organization is assigned as "Responsible organization".
In such special cases, even for those restricted roles, MT should be available in the form. Bulk action MT should be though not available.


Tipp: the "Update" Button on the upper right corner of the form must be accompanied by checkbox for MT.
Only "Submit for approval" is available or no button for upate appears, then MT should not be available.

charludo (changes requested):
I just spent quite a while trying to understand the purpose of the `content` variable being passed around, but think I finally got it. Unfortunately, I think passing it around just for the purposes of the one check in `is_permitted()` is a bit convoluted and hard to follow 🙈 

What do you think of the following idea:
    - bulk action translations are only allowed for staff users, and since in the call to `mt_provider.is_permitted()`, `content` is *not* handed down, I think it's safe to say this check isn't relevant for bulk actions
    - so how about moving the PageTranslation-specific checks to the `PageTranslationForm` instead?
    
This way `content` can be removed as a parameter again, and I *think* this would be equivalent (and, IMO, a little cleaner).

Do call me out though if this is nonsense 😅 😅

MizukiTemma:
@charludo Thank you for review, I agree with your idea 👍 That will be a way cleaner solution 😸 

This PR will be closed as its changes must be now reverted.

charludo:
@MizukiTemma do you want to open a second PR for this, or would you like someone else to work on the issue?

Issue #2998, opened 2024-08-14
AutoLink doesn't trigger when text contains soft hyphens 


PeterNerlich:
### Describe the Bug

Editors not only write page content from scratch, but copy segments from other sources like PDFs of leaflets etc. This might be the reason how invisible characters like soft hyphens are introduced to the content. The characters don't show up but don't match the pattern used to recognize links and automatically add anchor elements.

### Steps to Reproduce

1. Go to any page
2. Click inside the TinyMCE editor field
4. Copy and paste `www.test.de`
5. Type a space
6. Copy and paste `ww­w.test.de` *(which contains an invisible soft hyphen after the second `w`, as seen in production)*
7. Type a space

### Expected Behavior

Both links are recognized as such and given an anchor element

### Actual Behavior

Only the first link is given an anchor element, the second one containing a soft hyphen is ignored. Note how while stepping through character by character the cursor seems to stop before the third `w`.

### Additional Information

[What is a soft hyphen?](https://en.wikipedia.org/wiki/Soft_hyphen)

PR #2999, opened 2024-08-14
Ignore soft hyphens when detecting links

PeterNerlich:
### Short description

Ignore soft hyphens when auto detecting links in TinyMCE

### Proposed changes

- Ignore soft hyphens in detection only, but don't strip them out of the text content.

### Side effects

- Nonsensically placed soft hyphens will not be removed from the page. But on the flip side, sensibly placed ones will also stay.

### Resolved issues

Fixes: #2998

### DIFF
--- a/integreat_cms/release_notes/current/unreleased/2998.yml
+++ b/integreat_cms/release_notes/current/unreleased/2998.yml
@@ -1,0 +1,2 @@
+en: Ignore soft hyphens when detecting links
+de: Ignoriere bedingte Trennstriche bei der Erkennung von Links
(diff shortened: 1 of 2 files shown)


JoeyStk (dismissed):
Thank you very much!
I couldn't reproduce any of the reported issues, but as I understand it, this should be a good attempt at solving them :)

PeterNerlich:
> I couldn't reproduce any of the reported issues

Do you mean the original issue or what I mentioned under *Side effects*?

JoeyStk:
> Do you mean the original issue or what I mentioned under Side effects?

The original reported issue

PeterNerlich:
@JoeyStk did you use the *Steps to reproduce* from #2998?

deen13 (approved):
I was able to reproduce the issue and confirm that this pull request resolves it. The fix works as expected. 👍

MizukiTemma (approved):
I could reproduce the reported problem and this PR seems to have fixed it 🎉  

Thank you so much for your research, analysis and implementation 🥹 🙏

Issue #3001, opened 2024-08-15
Checkbox to auto update link always selected


PeterNerlich:
### Describe the Bug

Our current implementation seems to violate the following paragraph from the original outline in #1878:
> When inserting a link, we already have a dialog for selecting an internal content to link to. This dialog needs something like a check box "automatically update link description". This check box should be checked when having no text selected. When having a text selected, the checkbox should be unchecked.

Instead, the checkbox is also checked when text is selected.

### Steps to Reproduce

1. Go to any page
2. Select a word that is not yet linked
3. Create a link
4. The Link dialog pops up

### Expected Behavior

As existing text is selected, the checkbox should not be checked and the created link should retain it as link text.

### Actual Behavior

The checkbox is selected, and if a target is selected and the dialog is submitted in this state, the link retains the originally selected text. However, when submitting the content to the server, the link is immediately updated, as the link text does not reflect the title of the linked page.

### Additional Information

To reduce confusion along with fixing the bug, the following changes are proposed:
1. when text is selected, the checkbox should be unchecked
2. when the checkbox is checked on closing the dialog, the link text should be updated right away
3. the checkbox description could be more explicit
4. nice to have: when the checkbox is checked, the text field could be disabled and already show the title of the selected content

PR #3002, opened 2024-08-15
fix link auto update checkbox always checked

PeterNerlich:
### Short description

Fix link auto update checkbox always checked

### Proposed changes

- When the dialog is opened without an existing anchor being selected, determine whether the checkbox should be checked by checking if the selection is empty
- Disable the text field when the checkbox is checked
- Always set the text field to the targets title, and set text and URL to nothing whenever there is no auto completion when the checkbox is checked
- Reword the checkbox description
- Omit the `data-integreat-auto-update` attribute on links if `autoupdate` is `false` in order to not unnecessarily clutter the source code for normal links

### Side effects

- When the checkbox should be checked when the dialog is first opened, the text field is not yet disabled *(only once the update function is run)* – tell me if you know how to initially disable a field
- `data-integreat-auto-update` is not changing our "normal form" of content in the database. But maybe it should be? *(as in, stripping out `data-integreat-auto-update="false"` that do nothing but take up space)*

### Resolved issues

Fixes: #3001

### DIFF
--- a/integreat_cms/static/src/js/tinymce-plugins/custom_link_input/plugin.js
+++ b/integreat_cms/static/src/js/tinymce-plugins/custom_link_input/plugin.js
@@ -62,7 +62,10 @@ import { getCsrfToken } from "../../utils/csrf-token";
             const anchor = getAnchor();
             const initialText = anchor ? anchor.textContent : editor.selection.getContent({ format: "text" });
             const initialUrl = anchor ? anchor.getAttribute("href") : "";
-            const initialAutoUpdateValue = anchor ? anchor.getAttribute("data-integreat-auto-update") === "true" : true;
+            const initialAutoUpdateValue = anchor
+                ? anchor.getAttribute("data-integreat-auto-update") === "true"
+                : initialText === "";
+            let prevAutoupdateValue = initialAutoUpdateValue;
 
             const textDisabled = anchor ? anchor.children.length > 0 : false;
             let prevSearchText = "";
@@ -84,6 +87,19 @@ import { getCsrfToken } from "../../utils/csrf-token";
 
             const updateDialog = (api) => {
                 let data = api.getData();
+
+                if (data.autoupdate) {
+                    api.disable("text");
+                    if (!prevAutoupdateValue) {
+                        api.setData({
+                            url: "",
+                            text: "",
+                        });
+                    }
+                } else {
+                    api.enable("text");
+                }
+                prevAutoupdateValue = data.autoupdate;
 
                 let urlChangedBySearch = false;
                 // Check if the selected completion changed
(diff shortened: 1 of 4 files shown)


MizukiTemma:
@PeterNerlich @osmers 
A question: when an user selects a word and links an internal content to it, the selected word gets changed to the title of the linked content. In my opinion the selected text should remains unchanged 🤔

PeterNerlich:
> when an user selects a word and links an internal content to it, the selected word gets changed to the title of the linked content

Oops, that should only happen when the checkbox is checked

integreat_cms/static/src/js/tinymce-plugins/custom_link_input/plugin.js
```diff
@@ -85,6 +87,16 @@ import { getCsrfToken } from "../../utils/csrf-token";
             const updateDialog = (api) => {
                 let data = api.getData();
 
+                if (data.autoupdate) {
+                    api.disable("text");
+                    api.setData({
+                        url: "",
+                        text: "",
+                    });
```
david-venhoff, line 95: I think this does not work as is right now: For example, when a user opens the link dialog and enters "Willko" in the search bar, then the url and text will only be visible when the page "Willkommen in augsburg" has been found initially. Subsequent keystrokes however will clear those fields again, making it impossible to click on submit.

Unfortunately I

david-venhoff (changes requested):
> When the checkbox should be checked when the dialog is first opened, the text field is not yet disabled (only once the update function is run) – tell me if you know how to initially disable a field

I vaguely remember that I had the same problem when I added this checkbox, and could not find a valid solution (I think it was a bug of tinymce). That's why I gave up and decided to make the text field not disabled :(
I found this not really reassuring comment on the original pr by me: https://github.com/digitalfabrik/integreat-cms/pull/2530#issuecomment-1848682794

integreat_cms/static/src/js/tinymce-plugins/custom_link_input/plugin.js
PeterNerlich: Good catch. I'm getting the urge to refactor the `updateDialog()` function, but that assumes that the logic indeed can be reorganized in a better fashion :sweat_smile:

david-venhoff (commented):
The behavior when searching works now :)

I still found one (probably minor) problem:
- Searching for a text and then toggling the `automatically update link` checkbox off and on again disables the submit button and removes title and url

Another thing I found is that it currently possible to manually input a url if the checkbox is checked. Do you think it makes sense to only make the checkbox checkable if the search has returned a content object?

PeterNerlich:
> - Searching for a text and then toggling the `automatically update link` checkbox off and on again disables the submit button and removes title and url

Yes. *(where disabling the submit button is only a consequence of title and url being deleted)*
I chose not to bother with that yet, but if you think this to be very weird UX, I'll have a look whether this can be easily improved.

> Another thing I found is that it currently possible to manually input a url if the checkbox is checked.

As I suggested disabling the field for the link text, I also suggested disabling the one for the URL. However, this removes the possibility to scroll through long content, and links are often longer than the field can show. Another thing that would be fairly easy if we were not relying on TinyMCEs dialog API… Though I have not checked what possibilities it offers, I just feel like this feature has a low probability of being considered.

> Do you think it makes sense to only make the checkbox checkable if the search has returned a content object?

I don't think this makes sense. @osmers was even in favour of making it not an interactable checkbox at all and have the user just always create an auto updating link when they didn't have anything selected when running the function or always create a permanently "manual" link if they did. I personally think it makes the most sense to still give the option to the user but have sane defaults based on initial text selection.

david-venhoff (approved):
Alright, I think this can be merged then :)

MizukiTemma (approved):
Looks good 👍

Issue #2993, opened 2024-08-13
Refactor the `replace_links` function


david-venhoff:
### Motivation

This function is quite slow and hard to maintain. Because we now have `get_region_links`, we could rewrite `replace_links` in terms of this function, instead of maintaining two parallel implementations.

### Proposed Solution

Rewrite `replace_links` in terms of  `get_region_links` 

### Alternatives

Do nothing

### Additional Context

https://github.com/digitalfabrik/integreat-cms/pull/2983#pullrequestreview-2231868744

PR #3003, opened 2024-08-15
Refactor `replace_links` by using `get_region_links`

MizukiTemma:
### Short description

This PR refactors the function `replace_links` by using `get_region_links`

### Proposed changes

- Use `get_region_links` to collect link objects of the current region

### Side effects

- The function looks cleaner
- It doesn't have to go through all the translations of latest version anymore, now only checking translations related to links whose URL contains "search" string and is in the selected types.

### Resolved issues

Fixes: #2993 

### DIFF
--- a/integreat_cms/cms/models/abstract_content_translation.py
+++ b/integreat_cms/cms/models/abstract_content_translation.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+from copy import deepcopy
 from html import escape
 from typing import TYPE_CHECKING
 
@@ -11,6 +12,7 @@ from django.utils import timezone
 from django.utils.functional import cached_property
 from django.utils.translation import gettext_lazy as _
 from linkcheck.listeners import disable_listeners
+from lxml.html import rewrite_links
 
 from ..utils.tinymce_icon_utils import get_icon_html, make_icon
 
@@ -23,8 +25,10 @@ if TYPE_CHECKING:
 
     from .abstract_content_model import AbstractContentModel
     from .regions.region import Region
+    from .users.user import User
 
 from ..constants import status, translation_status
+from ..utils.link_utils import fix_content_link_encoding
 from ..utils.round_hix_score import round_hix_score
 from ..utils.translation_utils import gettext_many_lazy as __
 from .abstract_base_model import AbstractBaseModel
@@ -585,6 +589,41 @@ class AbstractContentTranslation(AbstractBaseModel):
         :return: All slugs that have been used by at least on version of this translation
         """
         return self.all_versions.values_list("slug", flat=True)
+
+    def create_new_version_copy(
+        self, user: User | None = None
+    ) -> AbstractContentTranslation:
+        """
+        Create a new version by copying
+        """
+        new_translation = deepcopy(self)
+        new_translation.pk = None
+        new_translation.version += 1
+        new_translation.minor_edit = True
+        new_translation.creator = user
+        logger.debug("Created new translation version %r", new_translation)
+
+        return new_translation
+
+    def replace_urls(
+        self,
+        urls_to_replace: dict[str, str],
+        user: User | None = None,
+        commit: bool = True,
+    ) -> None:
+        """
+        Function to replace links that are in the translation and match the given keyword `search`
+        """
+        new_translation = self.create_new_version_copy(user)
+        logger.debug("Replacing links of %r: %r", new_translation, urls_to_replace)
+        new_translation.content = rewrite_links(
+            new_translation.content,
+            lambda content_url: urls_to_replace.get(content_url, content_url),
+        )
+        new_translation.content = fix_content_link_encoding(new_translation.content)
+        if new_translation.content != self.content and commit:
+            self.links.all().delete()
+            new_translation.save()
 
     def __str__(self) -> str:
         """
--- a/integreat_cms/cms/utils/content_translation_utils.py
+++ b/integreat_cms/cms/utils/content_translation_utils.py
@@ -6,7 +6,6 @@ from __future__ import annotations
 
 import logging
 import operator
-from copy import deepcopy
 from functools import reduce
 from typing import TYPE_CHECKING
 
@@ -15,7 +14,6 @@ from linkcheck.models import Url
 
 from .content_utils import clean_content
 from .internal_link_utils import get_public_translation_for_link
-from .linkcheck_utils import save_new_version
 
 if TYPE_CHECKING:
     from ..models import User
@@ -49,9 +47,12 @@ def update_links_to(
         if new_content == outdated_content_translation.content:
             continue
 
-        fixed_content_translation = deepcopy(outdated_content_translation)
+        fixed_content_translation = (
+            outdated_content_translation.create_new_version_copy(user)
+        )
         fixed_content_translation.content = new_content
-        save_new_version(outdated_content_translation, fixed_content_translation, user)
+        outdated_content_translation.links.all().delete()
+        fixed_content_translation.save()
 
         logger.debug(
             "Updated links to %s in %r",
--- a/integreat_cms/cms/utils/linkcheck_utils.py
+++ b/integreat_cms/cms/utils/linkcheck_utils.py
@@ -1,19 +1,15 @@
 from __future__ import annotations
 
 import logging
-import re
 import time
-from copy import deepcopy
-from functools import partial
-from typing import TYPE_CHECKING
-from urllib.parse import ParseResult, unquote, urlparse
+from collections import defaultdict
+from typing import DefaultDict, TYPE_CHECKING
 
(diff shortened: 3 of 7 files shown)


integreat_cms/cms/utils/linkcheck_utils.py
```diff
@@ -252,36 +252,33 @@ def replace_links(
...
-            filters = {}
-            if region:
-                filters[f"{model.foreign_field()}__region"] = region
-
-            for translation in model.objects.filter(**filters).distinct(
-                model.foreign_field(), "language"
-            ):
-                new_translation = deepcopy(translation)
-                for link in translation.links.select_related("url"):
+        links = get_region_links(region) if region else Link.objects.all()
```
MizukiTemma, line 256: Because `replace_links` is used in the management command too and it is possible to leave "region" blank there, `Link.objects.all()` was added for the case of `region=None`

JoeyStk: Could you please write a comment in the code for that so we don't forget? :)

JoeyStk (changes requested):
Thank you very much for this PR!

However I would like to be a bit more ambitious here and refactor a bit more of this method. 
I think that ideally we could get rid of `# pylint: disable=too-many-locals`. Aditionally I find 60 lines for one method still a bit excessive. Therefore I think we should split this method up into multiple smaller methods.
My idea would be to split it up like this:

Lines 246 - 254 into an own method and name it maybe something like log_replacing_info() or log_replacement_is_starting() which takes the required arguments search, replace, region_msg and user_msg.

Then I would put lines 257 to 262 in another method and call this one something like `fetch_links`. Usually I would opt for `get_links`but since this is not a getter-method, I think we shouldn't use it here.

This is where I would be fine with stopping as the big for loop is the main essence of `replace_links` to me, pylint-ignore should be fixed by then and I think I find 30 lines of code for one method acceptable (at least as I envision it in my head).

integreat_cms/cms/utils/linkcheck_utils.py
MizukiTemma: Good idea :) done

MizukiTemma:
@JoeyStk 
Thank you for the idea 👍 Your suggestions are applied 😸

integreat_cms/cms/utils/linkcheck_utils.py
```diff
@@ -309,3 +291,55 @@ def fix_content_link_encoding(content: str) -> str:
...
+    Collects link objects which match the selected link types and contain string given as search in the url,
+    and find translations which contain those links
+
+    :param search: The (partial) URL to search
+    :param region: Optionally limit the replacement to one region (``None`` means a global replacement)
+    :param link_types: Which kind of links should be replaced
+
+    :return: A tuple of links and translations
+    """
+    # This function is used in replace_links, which is used in the management command, where region can be None, too.
```
JoeyStk, line 334: Nit: As this to me is concerning the entire method and not just a single line, I would add it in line 327 between description and types. But this is just my personal preference.

JoeyStk (approved):
Thank you very much! 
To me the code looks good now!

MizukiTemma:
Prio medium as this is blocking #2971

david-venhoff (changes requested):
I think ideally we should also add some more generic support for replacing links, as we are starting to have links that are not part of `content` (For example the website field of organizations). If we add linkcheck support for the website of organizations, then the linkcheck method will be relatively messy again. 

I have one idea how we could structure the code to avoid that, but that is just a suggestion :)

1. Add a method to every model that has checked links, which will replace all matching links of the object with the replace str. 
2. refactor `find_links_and_translations` to only return translations (actually if we include organizations, `translations` will not be a very fitting name anymore)
3. In the main loop, just call this new method of step 1

What do you think of that?

MizukiTemma:
@david-venhoff Thank you for review 😸 That's a good idea, now we can do `translation.replace_links()`

There are two divergence from your original suggestion: 1. some functions are moved to another file to avoid circular import, 2. we need `links_to_replace` to check whether the link matches the selected URL type (or do the same check again in `AbstractContentTranslation.replace_links`.

integreat_cms/cms/utils/linkcheck_utils/linkcheck_utils.py
```diff
@@ -243,69 +209,43 @@ def replace_links(
...
+        for link in links
+        if search in link.url.url and (not link_types or link.url.type in link_types)
+    ]
+    translations = {link.content_object for link in links_to_replace}
 
-    :param content: The input content
-    :return: The fixed content
-    """
-    return re.sub(r"(?<=[\"'])(https?://.+?)(?=[\"'])", fix_domain_encoding, content)
+    return links_to_replace, translations
```
david-venhoff, line 251: If you agree that it makes sense to change the signature of replace_links, then we should probably also change find_links_and_translations (Also we should change the name, because it will not just return translations in the future). I think it could just return a list of objects, plus their links that should be replaced. In the future we could even try to do this in completely in the database, but that should probably be another pr. So grouping the links by object in python is probably fine for now.

Here is how it could roughly look like:
```python
    links = (get_region_links(region) if region else Link.objects.all()).filter(url__url__contains=search).select_related("url")
    if link_types:
        links_to_replace = [link for link in links if link.url.type in link_types]
    else:
        links_to_replace = links

    content_objects = defaultdict(set)
    for link in links_to_replace:
        content_objects[link.content_object].add(link.url)

    return content_objects
```

david-venhoff (changes requested):
Thanks, I think this is already a really nice improvement over the current api :tada:

I noticed that we have some duplicate code at `linkcheck_list_view`. However, now we can deduplicate that relatively easily :) 
Can you include that in this pr?

I think we don't need to create a submodule `linkcheck_utils`. Looking at the functions in `linkcheck_util_helper_functions.py`:
- `log_replacement_is_starting`: This is an internal function and can be moved to `linkcheck_utils.py`
- `fix_content_link_encoding` (And its helper `fix_domain_encoding`): I'd argue that this function is not related at all to linkcheck utils and could better live at another file in utils (Maybe a new file `link_utils`?)
- `replace_link_helper` is now only used at abstractContent translation and could maybe be moved to there (Or just be a lambda)

These changes would make it possible to keep the old file structure.

integreat_cms/cms/models/abstract_content_translation.py
```diff
@@ -586,6 +595,49 @@ def get_all_used_slugs(self) -> Iterable[str]:
         """
         return self.all_versions.values_list("slug", flat=True)
 
+    def save_new_version(
```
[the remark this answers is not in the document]
MizukiTemma: Hmm, what is a benefit of that change over the current implementation? 👀

MizukiTemma:
@david-venhoff 
I didn't imagine we would go such far but we can and it's done 😸 I hope I've understood correctly what you had in your head.

integreat_cms/cms/models/abstract_content_translation.py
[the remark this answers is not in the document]
david-venhoff: The problem I have with the current version of `save_new_version` is that it is really easy to misuse.
The caller needs to make sure that
* `new_translation` has the same language and the same related object as `self` and
* The `version` field of `new_translation` is equal to the version field `self`

We could of course document those requirements, but IMO it would be even better if these requirements to the caller would not exist at all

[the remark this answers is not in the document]
david-venhoff: Actually, is there a reason why we need a parameter to that function? If I understand correctly, it would be an option to just accept self and then modify all fields of self, right?

david-venhoff (approved):
> I hope I've understood correctly what you had in your head.

Yes, this is exactly what I thought of :)
I think this pr is good to go, except for some small comments I have :tada:

MizukiTemma:
@JoeyStk 
Could you review this PR again? It was changed a lot😅

integreat_cms/cms/utils/content_translation_utils.py
```diff
@@ -49,9 +47,12 @@ def update_links_to(
         if new_content == outdated_content_translation.content:
             continue
 
-        fixed_content_translation = deepcopy(outdated_content_translation)
+        fixed_content_translation = (
```
JoeyStk, line 50: I think I missed some context here, why is it better to use our own implementation of `create_new_version_copy()` instead of deepcopy? :)

JoeyStk (approved):
Thank you very much! I think this looks a lot cleaner now! :)

integreat_cms/cms/models/abstract_content_translation.py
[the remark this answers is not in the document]
JoeyStk: I think it'd be a good idea to document this somewhere, I'm not sure if we'll be looking in this PR again to understand this decision again in a few months or years

PR #3007, opened 2024-08-16
Add contact list view

JoeyStk:
### Short description

Add list view for the new section "Contacts" 

### Proposed changes

- Add list views for contacts
  - List view for not archived contacts
  - List view for archived contacts
- Add option to archive, delete, restore and copy contacts 
- Add test for these four options

### Side effects

I think none?

### Resolved issues

Fixes: #2952


integreat_cms/cms/templates/contacts/contact_list_archived_row.html
```diff
@@ -0,0 +1,45 @@
...
+           target="_blank"
+           rel="noopener noreferrer"
+           class="text-blue-500 hover:underline"
+           title="{% translate "To website" %}">
+            {{ contact.website }}
+        </a>
+    </td>
+    <td class="pl-2 pr-4 text-right min">
+        {% has_perm 'cms.change_contact' request.user as can_edit_contact %}
+        {% if can_edit_contact %}
```
MizukiTemma, line 35:
```suggestion
        {% if perms.cms.change_contact %}
```
I think we can simply do like .

Maybe an delete option here too? We have it in the archived poi/event list.

integreat_cms/cms/templates/contacts/contact_list_row.html
```diff
@@ -0,0 +1,57 @@
...
+        <a href="{{ contact.website }}"
+           target="_blank"
+           rel="noopener noreferrer"
+           class="text-blue-500 hover:underline"
+           title="{% translate "To website" %}">
+            {{ contact.website }}
+        </a>
+    </td>
+    <td class="pl-2 pr-4 text-right min">
+        {% if perms.cms.edit_contact %}
```
MizukiTemma, line 33:
```suggestion
        {% if perms.cms.change_contact %}
```
Maybe remains from copy&paste?

integreat_cms/cms/templates/contacts/contact_list_row.html
```diff
@@ -0,0 +1,57 @@
...
+    </td>
+    <td>
+        <a href="" class="block py-3 px-2 text-gray-800">
+            {{ contact.name }}
+        </a>
+    </td>
+    <td>
+        <a href="" class="block py-3 px-2 text-gray-800">
+            {{ contact.poi }}
+        </a>
```
MizukiTemma, line 16: it probably helps users if we add here link to the referenced POI?

MizukiTemma (changes requested):
Thank you for the PR 👍 Beside the suggestions in commets, I have two questions:

1. Do we need the "copy" functionality? I see the copy option is drawn in the design but we've never talked about it so far I remember 🤔 Of course it doesn't harm anything, but I'm asking for clarifying.

2. Currently URLs of websites are shown in the column "Phone number". Should we add another column for "Website" although this is not picted in the design?

integreat_cms/cms/templates/contacts/contact_list_row.html
JoeyStk: I think it was more that I was having the wrong thought here :)

integreat_cms/cms/templates/contacts/contact_list_row.html
JoeyStk: Alright. Sounds reason-able. I will style it like a link to make it obvious :)

JoeyStk:
> Do we need the "copy" functionality? I see the copy option is drawn in the design but we've never talked about it so far I remember 🤔 Of course it doesn't harm anything, but I'm asking for clarifying.

I asked @osmers, and the copy-functionality is still desired as it might make it easier for municipalities to create contacts.

> Currently URLs of websites are shown in the column "Phone number". Should we add another column for "Website" although this is not picted in the design?

Oh, good point! I added a website column and I'm showing the phone_number now too :)

osmers:
> I asked @osmers, and the copy-functionality is still desired as it might make it easier for municipalities to create contacts.

Or rather - I'm sure that if we don't add it, we'll have a municipality asking for it and since I don't see any harm in it (please correct me, if I'm overlooking something), I don't see why we shouldn't add it :)

MizukiTemma (approved):
🎉 Looks good 😸

integreat_cms/cms/templates/contacts/contact_list_row.html
```diff
@@ -0,0 +1,63 @@
...
+        {% if perms.cms.change_contact %}
+            <a href="{% url 'copy_contact' contact_id=contact.id region_slug=request.region.slug %}"
+               title="{% translate "Copy contact" %}"
+               class="btn-icon">
+                <i icon-name="copy"></i>
+            </a>
+            <button title="{% translate "Archive contact" %}"
+                    class="confirmation-button btn-icon"
+                    data-confirmation-title="{{ archive_dialog_title }}"
+                    data-confirmation-subject="{{ contact.best_translation.title }}"
```
MizukiTemma, line 48:
```suggestion
                    data-confirmation-subject="{{ contact.name }}"
```
I've just noticed now 😅

david-venhoff (changes requested):
Thanks, this looks mostly good to me.
Don't worry about the high number of suggestions, most of them are just related to styling :see_no_evil: 

Also looks like the archived contact list is missing the website column. Maybe it would make sense to deduplicate contact_list and contact_list_archived? Though we could also do that later if the time is too low now.

david-venhoff:
Oh, and it looks like pagination is not working. Could you please add some pagination code to `contact_list_view`?
To test, just duplicate contacts until you have more than 10. If you then select max 10 entries per page, the paginator should actually limit the number of contacts to that and show two pages total.

david-venhoff:
I think it would make sense if we would refactor all the other views at some point too, so consistency would not be a problem here imo. But I think it would be ok to either refactor the template in this pr or in another pr :)

JoeyStk:
@david-venhoff Thank you for all your feedback! I think it should be all applied now (+ a few more changes)

> Also looks like the archived contact list is missing the website column. Maybe it would make sense to deduplicate contact_list and contact_list_archived? 

Applied :)

integreat_cms/cms/views/contacts/contact_list_view.py
```diff
@@ -0,0 +1,78 @@
...
+        )
+        chunk = request.GET.get("page")
+        contact_chunk = paginator.get_page(chunk)
+
+        return render(
+            request,
+            self.template_name,
+            {
+                **self.get_context_data(**kwargs),
+                "is_archive": bool(self.archived),
```
JoeyStk, line 72:
```suggestion
                "is_archive": self.archived,
```

JoeyStk:
@david-venhoff I think this PR is ready for review again. A bit of back information:

- I added the check that pois can't be deleted if a contact uses it
- I deactivated the buttons to delete a poi
- Therefore I had to adapt the test for deleting a poi. As it's not possible to delete a poi anymore that is in use, no `ProtectedError` will be thrown anymore. Now I'm only checking for a redirect (status 302). However I'm quite unsure, if this is a good practice - I have a feeling it's not. What do you think?

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -5043,6 +5047,125 @@ msgstr "Chat"
...
+#: cms/templates/languages/language_form.html
+#: cms/templates/offertemplates/offertemplate_list.html
+#: cms/templates/organizations/organization_list.html
+#: cms/templates/poicategories/poicategory_list.html
+#: cms/templates/regions/region_list.html cms/templates/roles/role_list.html
+msgid "Name"
+msgstr "Name"
+
+#: cms/templates/contacts/contact_list.html
+msgid "Name of related POI"
```
david-venhoff, line 5078:
```suggestion
msgid "Name of related location"
```

Do we have a naming guide how to call pois in english? I find it a bit confusing that the table says poi, while the sidebar says location.
However, I noticed that we have quite a lot of occurrences of "poi" in this file, as well as a lot of occurrences of "location"

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -9787,6 +9875,24 @@ msgstr "Bitte bestätigen Sie, dass dieser Ort gelöscht werden soll"
 msgid "All translations of this location will also be deleted."
 msgstr "Alle Übersetzungen dieses Ortes werden gelöscht."
 
+#: cms/views/pois/poi_context_mixin.py
+msgid ""
+"You cannot archive a poi which is used by an event or a contact. \n"
```
david-venhoff, line 9880:
```suggestion
"You cannot archive a location which is used by an event or a contact. \n"
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -9787,6 +9875,24 @@ msgstr "Bitte bestätigen Sie, dass dieser Ort gelöscht werden soll"
...
+"This also involves archived events and contacts"
+msgstr ""
+"Sie können keinen Ort archivieren, der von einer Veranstaltung oder einem "
+"Kontakt verwendet wird. \n"
+"Dies betrifft auch archivierte Veranstaltungen und Kontakte"
+
+#: cms/views/pois/poi_context_mixin.py
+msgid ""
+"You cannot delete a poi which is used by an event or a contact. \n"
```
david-venhoff, line 9889:
```suggestion
"You cannot delete a location which is used by an event or a contact. \n"
```

integreat_cms/cms/migrations/0098_add_contact.py
```diff
@@ -96,6 +96,7 @@ class Migration(migrations.Migration):
                     "poi",
                     models.ForeignKey(
                         on_delete=django.db.models.deletion.PROTECT,
+                        related_name="contacts",
```
david-venhoff, line 99: I don't think we should be modifying an existing migration, but I really don't know what kind of problems occur if we do it or if we don't do it :sweat_smile: 

Could you test if everything still works if you remove this line? If not it is probably best to create a new migration.

integreat_cms/cms/models/pois/poi.py
```diff
@@ -145,19 +148,36 @@ def get_translation_model() -> ModelBase:
...
+        Deletes the poi
+        :param \*args: The supplied arguments
+        :param \**kwargs: The supplied keyword arguments
+        """
+        if not self.is_used:
+            super().delete(*args, **kwargs)
+        else:
+            logger.debug(
+                "Can't be deleted because this poi is used by an event or a contact"
+            )
```
david-venhoff, line 162: I think for consistency it would make sense to apply the same solution here as in the organisation pr (return whether the poi was deleted) 
Right now, you can see that a success message is shown if you delete a used poi from the poi form, even though this method makes sure that it does not get deleted

integreat_cms/cms/views/contacts/contact_list_view.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @property
+    def template_name(self) -> str:
+        """
+        Select correct HTML template, depending on :attr:`~integreat_cms.cms.views.pois.contact_list_view.ContactListView.archived` flag
+        (see :class:`~django.views.generic.base.TemplateResponseMixin`)
+
+        :return: Path to HTML template
+        """
+        return self.template_archived if self.archived else self.template
```
david-venhoff, line 40:
```suggestion
    template_name = "contacts/contact_list.html"
    archived = False
```

integreat_cms/cms/views/contacts/contact_list_view.py
```diff
@@ -0,0 +1,78 @@
...
+        Render contact list
+
+        :param request: The current request
+        :param \*args: The supplied arguments
+        :param \**kwargs: The supplied keyword arguments
+        :return: The rendered template response
+        """
+        region = request.region
+
+        contacts = Contact.objects.filter(poi__region=region, archived=self.archived)
```
david-venhoff, line 53:
```suggestion
        contacts = Contact.objects.filter(poi__region=region, archived=self.archived).select_related("poi")
```

tests/cms/views/poi/test_poi_form.py
```diff
@@ -167,8 +167,8 @@ def test_poi_in_use_not_deleted(
             == f"{settings.LOGIN_URL}?next={delete_poi}"
         )
     elif role in HIGH_PRIV_STAFF_ROLES:
-        with pytest.raises(ProtectedError):
-            response = client.post(delete_poi)
```
david-venhoff, line 171: I think that is mostly fine. You can add an assertion that the poi with `poi_id` still exists after the post request, or check for the `messages.error` text if you add that

david-venhoff (changes requested):
Thanks, this is almost good to go. There is just one missing thing I found.

The poi form delete button is still available if the poi has a contact, while it shows a detailed explanation why it cannot be deleted if it has a linked event. It would expect that something similar to this is shown:
image

integreat_cms/cms/templates/pois/poi_form_sidebar/action_box.html
```diff
@@ -47,6 +47,24 @@
...
+                        <br />
+                        {% blocktranslate count counter=poi_form.instance.contacts.count trimmed %}
+                            To archive this location, you have to delete this contact first:
+                        {% plural %}
+                            To archive this location, you have to delete these contacts first:
+                        {% endblocktranslate %}
+                    </p>
+                </div>
+                {% for contact in poi_form.instance.contacts.all %}
+                    <a href="#" class="block pt-2 hover:underline">
```
JoeyStk, line 64: @MizukiTemma could you please set the real URL for editing a contact in your PR? :)

MizukiTemma: Thank you for letting me know it 👍 I'll add it once this PR gets merged 😸

JoeyStk:
@MizukiTemma could you please check if you think that #3010 is also already resolved by this PR or if you had something different in mind when you created the issue? :)

MizukiTemma:
@JoeyStk Thank you so much for all the work, this is solving #3010 too 😸 

One thing to fix is that the success message "Ort "XXX" wurde erfolgreich archiviert" is shown when only a POI which is referenced by a contact is selected and bulk archived.
The POI is anyway not deleted, so I'm fine if this will be fixed in another issue.

MizukiTemma (approved):
🎉

integreat_cms/cms/views/pois/poi_actions.py
```diff
@@ -42,7 +42,7 @@ def archive_poi(
     """
     poi = POI.objects.get(id=poi_id)
 
-    if poi.events.count() > 0:
+    if poi.events.count() > 0 or poi.contacts.count() > 0:
```
david-venhoff, line 45: I think it would be better to include that check in `poi.archive`, so that it works the same as `poi.delete` (Also we can use `poi.is_used` here)

integreat_cms/cms/views/pois/poi_actions.py
```diff
@@ -108,8 +108,15 @@ def delete_poi(
 
     poi = POI.objects.get(id=poi_id)
     logger.debug("%r deleted by %r", poi, request.user)
```
david-venhoff, line 110:
```suggestion
```

tests/cms/views/poi/test_poi_form.py
```diff
@@ -167,8 +167,8 @@ def test_poi_in_use_not_deleted(
...
+        response = client.post(delete_poi)
+        assert poi_id is not None
```
david-venhoff, line 171: Even if the poi would have been deleted, poi_id would not change, since it is never updated

david-venhoff (changes requested):

integreat_cms/cms/views/pois/poi_actions.py
JoeyStk: I think I will change this line to `is_used` then, I think this is the better solution

david-venhoff (approved):

Issue #2094, opened 2023-02-22
Refactor calculation of recurring events


timoludwig:
### Motivation

At the moment, the calculation of event recurrences is spread over the
- events API:
  https://github.com/digitalfabrik/integreat-cms/blob/a15be1005910cdacb0e62e8d11f5149b8da75d71/integreat_cms/api/v3/events.py#L129-L167
- Event model:
  https://github.com/digitalfabrik/integreat-cms/blob/a15be1005910cdacb0e62e8d11f5149b8da75d71/integreat_cms/cms/models/events/event.py#L216-L238
- RecurrenceRule model:
  https://github.com/digitalfabrik/integreat-cms/blob/a15be1005910cdacb0e62e8d11f5149b8da75d71/integreat_cms/cms/models/events/recurrence_rule.py#L126-L174

### Proposed Solution

Refactor the code so it
1. Isn't spread over different modules anymore (probably the recurrence rule model is the most sensible place for it)
2. Uses the iCal rrule to calculate the recurrences instead of the custom function

### Additional Context

Also, there is a small bug currently that recurring events which spread over multiple days are not shown in the API when the occurrence is ongoing but started e.g. yesterday.

PR #2992, opened 2024-08-13
Correctly represent the last week in rrules

david-venhoff:
### Short description

This pr fixes a bug where an event that repeats monthly on the last week would produce an invalid rrule. Instead of saying `-1<Weekday>`, which represent <Weekday> on the last week of the month, our code produced `5<Weekday>`, specifying  <Weekday> in the 5th Week of the month.

### Proposed changes

- Add a mapping from week constant to rrule weekday
- Specify that `weeks.LAST` has value -1 in an rrule

### Side effects

/

### Resolved issues

Fixes: #2989

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

Issue #2989, opened 2024-08-13
Adding an event on the last weekday of every month only adds it in months with 5 weeks


osmers:
### Describe the Bug
If an event is set to take place every last (e.g.) Wednesday of the month, it only shows dates for months with 5 weeks.

### Steps to Reproduce

1. Go to https://admin.integreat-app.de/testumgebung/events/de/6253/edit/
2. See event set for every Wednesday during the last week of every month
3. Go to https://integreat.app/testumgebung/de/events/jeden-letzten-mittwoch-im-monat
4. See that only months with 5 weeks are shown

### Expected Behavior
The last Wednesday of every month should be shown regardless of whether the month has 4 or 5 weeks

### Actual Behavior
Only months with 5 weeks are shown

### Additional Information

PR #3026, opened 2024-08-27
Correctly represent the last week in rrules Part 2

### LINKED ISSUE #3022: Recurrence rule in release is broken
### Describe the Bug

The recurrence rule in #3018 is (still) broken. As long as this is not resolved, our release is blocked

### Steps to Reproduce

1. Log on to our test system, go to events
2. Create a new event, set as recurrence rule every last Thursday starting in August (every month, last week, Thursday, every 1st...)
3. Go to the web app
4. See that for every month the 29th is shown

### Expected Behavior

Show for an event that happens every last Thursday the list of last Thursdays (so August 29th, September 26th, October 31st, November 28th, etc.).

### Actual Behavior

 (1) The current state **on production** is that only August 29th, October 31st and January are shown (months that have a fifth week). (2) [#2992](https://github.com/digitalfabrik/integreat-cms/pull/2992) attempted to fix this, but **on the test system** now the 29th of a month is shown.

### Additional Information

grafik State after #2992
grafik State on production, before #2992

PeterNerlich:
### Short description

#2992 Found and fixed a bug in rendering out recurring events to iCal format. However, our function to internally generate the next event instances using the recurrence rule was still broken.
This PR attempts to fix the implementation of that function and add tests to it which seemed to be entirely missing until now.

### Proposed changes

- I decided to change how we store and handle the intention of "last occurrence in month" to using negative values to count from the end of the month instead of the start, rendering the functional changes of #2992 unnecessary. But it should be closer to how iCal does it and be less ambiguous.
- As a consequence, specifying `5` to select the number of occurrence of the weekday in the month *is still valid*, but will now really only mean months with enough days to have a fifth e.g. Friday. Months too short will be skipped and not produce these events. This is however not selectable through the UI but something the system in theory supports but which we won't expose to users at this moment in time.
  - Add a data migration to change all existing RecurrenceRules with frequency *monthly* and week_for_monthly of `5` to a value of `-1`
- We now also support arbitrary counting from the end of month, so I took the liberty to also expose a value of `-2` under the option of *Second last week*.
- I added tests for the `iter_after()` function. It seems like only `to_ical_rrule()` was being tested until now.

### Side effects

- *none come to mind*

### Resolved issues

Fixes: #3022, #2989


david-venhoff (commented):
I think this will not fix our issue, because I think the actual bug is on the frontend, as I commented here: https://github.com/digitalfabrik/integreat-cms/issues/3022#issuecomment-2315046828

@PeterNerlich Could you please try to verify this?

Still, this pr fixes another bug, so it would still be worth merging, just with a lower prio

david-venhoff (changes requested):
Assuming that this is not a critical fix anymore, our manual `iter_after` implementation seems error-prone and hard to maintain to me. Also it appears that dateutil library that we are already using has support for exactly the same functionality. So it seems like we could just replace this whole code with a simple `self.to_ical_rrule().xafter()` oneliner: https://dateutil.readthedocs.io/en/stable/rrule.html#dateutil.rrule.rrule.xafter
Could you try if that could work for our usecase?

david-venhoff:
I think this would also close #2094

JoeyStk:
Close because it was stale for a while

Issue #2953, opened 2024-08-01
[2024-11-07] Create form for contacts


MizukiTemma:
### Motivation

We are introducing a new feature "contact information". A form is needed for it.

### Proposed Solution

Implement a form to create a new contact and to edit an existing one.

<s>❓ Make it sure at least one of E-mail address, telephone number or website is given.</s>
💡 Reuse venue_box.html for the step of selecting/creating a POI.
💡 Reuse _related_contents_table.html to show where the contact is embedded.

### Alternatives

None. We do it.

### Additional Context

Wait for #2952 

### Design Requirements

See the design [here](https://www.figma.com/design/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=1752-1636&t=5FU8fiMws9yWz1S8-0)
⚠️ There are two versions in the link above. **New Version 30.07** is the one we have to implement.

MizukiTemma:
The cancel (Abbrechen) button will be implemented in #2930. It is not consistent with other contents and may be confusing for users if the cancel button is only available in the contact form.

MizukiTemma:
blocked by #3015

MizukiTemma:
The design suggests to mark the fields "title" and "name" with * to make it clear that they are mandatory, but mandatory fields of other forms are not marked at all. Should we probably introduce the `*` systemwide at once? #3028

MizukiTemma:
TBD: removing E-mail, phone number and website from POI.

We have those three fields not only in POIs but also in contacts. If those fields are managed only in contacts, we can avoid redundancy, maintenance like #3033 and check boxes in #3034 .

@david-venhoff Feel free to add commnets 😸

MizukiTemma:
> TBD: removing E-mail, phone number and website from POI.
> 
> We have those three fields not only in POIs but also in contacts. If those fields are managed only in contacts, we can avoid redundancy, maintenance like #3033 and check boxes in #3034 .
> 
> @david-venhoff Feel free to add commnets 😸 @PeterNerlich You're invited to the discussion too 😸

This idea was confirmed and will be implemented. See #3086

Issue #3033, opened 2024-09-02
[7.11.24] POI update and contacts


MizukiTemma:
### Motivation

There will be such cases that a POI is updated with a new E-mail address, phone number or website and some contacts are taking those infos from the POI. What should happen in such a case?

### Proposed Solution

1. Update the related contacts with the new E-mail, phone number and website of the updated POI if they are taking those fields from the POI.
     Pros: they always synchronise.
     Cons: less flexibility for users (Maybe sometimes they need to update the POI but don't want to automatically influence on (all) the contacts.) 

And show a message "The followting contacts are updated: A, B, ......, Z" after successful update of the POI.

<s>
2. Do not update the related contacts (+ show an warning that they are divergent in the contact list or form).
     Pros: users have more control.
     Cons: users have to check (and adjust) the contacts by themselves.

3. Let users choose which contacts must be updated together with the POI
     Pros: more flexibility for users
     Cons: too much effort/steps for users? more technical work

### Alternatives

</s>

david-venhoff:
One alternative would be to completely disallow users from updating email-address, phone number and website details once we land contacts. Contacts completely replace the functionality provided by these fields and we should not support two equivalent solutions to achieve the same thing. 

Unfortunately we cannot write a migration that automatically creates a contact from pois with these fields, because we cannot extract the required field `title`. However, we could deprecate these fields. Then they would not be visible to users when they create a new poi, or when they edit an existing poi that has a linked contact. They would only be visible when a user edits a poi without a contact, but there we would prevent users from changing these fields unless they create a contact. This way, we could eventually phase-out these poi fields and at some point in the future remove the fields from the poi model.

The big advantage would be no redundancy.

MizukiTemma:
@david-venhoff 
I'm not sure we are sharing the same context. It's neither about letting users change name, address, e-mail etc of POI in the contact form nor creating a new contact automatically when creating/updatig a POI .The target scenario is as below:

1. There is a POI with an website "some-website.com"
2. There is a contact, which is related to this POI and inheriting the website URL.
3. Now the POI is updated with a new website "new-website.com"
4. When saving this change, the website field of the contact must be changed to "new-website.com", when `website_from_location` is `True`

david-venhoff:
> 3. Now an user opened this POI, gave a new website "new-website.com" and saved it.

Well, if we don't let users change the email, etc. of pois, then point 3 cannot apply and we don't have this problem.

[...]

PR #3034, opened 2024-09-03
[7.11.24] Add form for contact information 

MizukiTemma:
### Short description

This PR implements a new form for contact objects.

**This PR is not yet finalized.** Changes from #3007 have been taken over but are not actual in this branch. Once it is merged, this PR will be updated too.

### To Do after #3007 is merged
- [x] Add the real URL to contacts listed in `poi_form_sidebar/action_box.html`

### Proposed changes

- Add three new boolean fields to the model contact to store whether E-mail, phone number and website must be imported from the related POI. 
- Make the fields for E-mail, phone number and website read-only when the corresponding check box is checked.
- Create a new form `ContactForm` with check whether those three fields match with those of the selected POI if checked for import from the POI
- Create a new template for the form, reusing `poibox.html` and `_related_contents_table.html`

### Side effects

- Change the field name "poi" to "location" in the contact model to be compatible with `poi_box.html`

### Memo
- There is not yet any implementation to find contents in which the contact is used. Third box "Contact information" is currently only a place hlder. This will be implemented during #2959 

### Resolved issues

Fixes: #2953 


MizukiTemma:
Update: Name is not mandatory.

integreat_cms/cms/forms/contacts/contact_form.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def clean(self) -> dict[str, Any]:
+        """
+        Validate the fields
+
+        :return: The cleaned form data
+        """
+        cleaned_data = super().clean()
+
+        related_location = POI.objects.filter(id=self.data["location"]).first()
```
david-venhoff, line 63:
```suggestion
        related_location = self.instance.region.pois.filter(id=self.data["location"]).first()
```

I think we should also filter by region, just to make sure that no one can create a contact that references a poi from another region. That would require also passing the additional instance attributes in the post method of the view.

integreat_cms/cms/views/contacts/contact_form_view.py
```diff
@@ -0,0 +1,145 @@
...
+        :raises ~django.core.exceptions.PermissionDenied: If user does not have the permission to change contact
+
+        :return: The rendered template response
+        """
+        region = request.region
+
+        contact_instance = Contact.objects.filter(id=kwargs.get("contact_id")).first()
+        contact_form = ContactForm(
+            data=request.POST,
+            instance=contact_instance,
```
david-venhoff, line 102:
```suggestion
            instance=contact_instance,
            additional_instance_attributes={"region": region},
```

integreat_cms/cms/templates/ajax_poi_form/poi_box.html
```diff
@@ -11,17 +11,24 @@
...
+        <label class="secondary !inline"
+               for="{{ form.has_not_location.id_for_label }}">
+            {{ form.has_not_location.label }}
+        </label>
+    {% endif %}
+    {% if current_menu_item == "contacts" %}
+        <div class="help-text mt-4 font-bold">
+            {{ help_text }}
+        </div>
+    {% endif %}
```
david-venhoff, line 25: Since this was designed as a generic component, we should not include checkes here from where this is used imo. In this case, could you just move this code to the places from where the poi box is included?

integreat_cms/cms/templates/ajax_poi_form/poi_box.html
```diff
@@ -34,14 +41,16 @@
...
                         {% if form.disabled %}disabled{% endif %}>
                     <i icon-name="trash-2" class="h-5 w-5"></i>
                 </button>
             </div>
         </div>
         <p class="text-sm italic block mt-2 mb-2">
-            {% translate "Create an event location or start typing the name of an existing location. Only published locations can be set as event venues" %}.
+            {% if current_menu_item == "events_form" %}
+                {{ help_text }}
+            {% endif %}
```
david-venhoff, line 53: Same here. I would either move this code to the event form, or if that does not look similar, maybe pass a help text variable to this component

integreat_cms/cms/templates/contacts/contact_form.html
```diff
@@ -0,0 +1,145 @@
...
+                        {% translate "Create new contact" %}
+                    {% endif %}
+                </h1>
+                <div class="flex flex-wrap justify-between gap-4">
+                    {% if perms.cms.change_contact and not contact_form.instance.archived %}
+                        <div class="flex flex-wrap gap-4">
+                            <button class="btn">
+                                {% translate "Save" %}
+                            </button>
+                        </div>
```
david-venhoff, line 26: Should we differentiate here between save and update? Since this is what we currently do for abstract content translations

integreat_cms/cms/templates/contacts/contact_form.html
```diff
@@ -0,0 +1,145 @@
...
+                        <label for="{{ contact_form.name.id_for_label }}">
+                            {{ contact_form.name.label }}
+                        </label>
+                        {% render_field contact_form.name|add_error_class:"border-red-500" %}
+                        <label for="{{ contact_form.email.id_for_label }}">
+                            {{ contact_form.email.label }}
+                        </label>
+                        {% render_field contact_form.email|add_error_class:"border-red-500" %}
+                        {% render_field contact_form.email_from_location %}
+                        <label for="{{ contact_form.emial_from_location.id_for_label }}"
```
david-venhoff, line 56:
```suggestion
                        <label for="{{ contact_form.email_from_location.id_for_label }}"
```

integreat_cms/cms/templates/contacts/contact_form.html
```diff
@@ -0,0 +1,145 @@
...
+                               class="secondary !inline">
+                            {{ contact_form.email_from_location.help_text }}
+                        </label>
+                        <label for="{{ contact_form.phone_number.id_for_label }}">
+                            {{ contact_form.phone_number.label }}
+                        </label>
+                        {% render_field contact_form.phone_number|add_error_class:"border-red-500" %}
+                        {% render_field contact_form.phone_number_from_location %}
+                        <label for="{{ contact_form.phone_number.id_for_label }}"
```
david-venhoff, line 65:
```suggestion
                        <label for="{{ contact_form.phone_number_from_location.id_for_label }}"
```

integreat_cms/cms/templates/contacts/contact_form.html
```diff
@@ -0,0 +1,145 @@
...
+                {% if contact_form.instance.id %}
+                    <div class="mb-4 rounded border border-solid border-blue-500 shadow-2xl bg-white">
+                        <div class="p-4 rounded bg-water-500">
+                            <h3 class="heading font-bold text-black">
+                                <i icon-name="message-square" class="mr-2"></i> {% translate "Contact information" %}
+                            </h3>
+                        </div>
+                        <div class="p-4">
+                            <div class="help-text mt-4 font-bold">
+                                {% trans "This contact is referred in those contents." %}
```
david-venhoff, line 89:
```suggestion
                                {% trans "This contact is referred to in those contents." %}
```

integreat_cms/cms/templates/contacts/contact_form.html
```diff
@@ -0,0 +1,145 @@
...
+                            </div>
+                        </div>
+                        {% include "../_related_contents_table.html" with contents=referring_pages table_title=_("Pages") no_content_message=_("This contact is currently referred in no page.") %}
```
david-venhoff, line 92:
```suggestion
                        {% include "../_related_contents_table.html" with contents=referring_pages table_title=_("Pages") no_content_message=_("This contact is not currently referred to on any page.") %}
```

integreat_cms/cms/templates/contacts/contact_form.html
```diff
@@ -0,0 +1,145 @@
...
+                        {% include "../_related_contents_table.html" with contents=referring_locations table_title=_("Locations") no_content_message=_("This contact is currently referred in no location.") %}
```
david-venhoff, line 93:
```suggestion
                        {% include "../_related_contents_table.html" with contents=referring_locations table_title=_("Locations") no_content_message=_("This contact is not currently referred to in any location.") %}
```

integreat_cms/cms/templates/contacts/contact_form.html
```diff
@@ -0,0 +1,145 @@
...
+                        {% include "../_related_contents_table.html" with contents=referring_events table_title=_("Events") no_content_message=_("This contact is currently referred in no event.") %}
```
david-venhoff, line 94:
```suggestion
                        {% include "../_related_contents_table.html" with contents=referring_events table_title=_("Events") no_content_message=_("This contact is not currently referred to in any event.") %}
```

integreat_cms/cms/views/contacts/contact_form_view.py
```diff
@@ -0,0 +1,145 @@
...
+        r"""
+        Render contact form for HTTP GET requests
+
+        :param request: Object representing the user call
+        :param \*args: The supplied arguments
+        :param \**kwargs: The supplied keyword arguments
+        :return: The rendered template response
+        """
+        region = request.region
+        contact_instance = Contact.objects.filter(id=kwargs.get("contact_id")).first()
```
david-venhoff, line 43:
```suggestion
        contact_instance = Contact.objects.filter(id=kwargs.get("contact_id"), location__region=region).first()
```

integreat_cms/cms/views/contacts/contact_form_view.py
```diff
@@ -0,0 +1,145 @@
...
+
+        :param request: Object representing the user call
+        :param \**kwargs: The supplied keyword arguments
+        :raises ~django.core.exceptions.PermissionDenied: If user does not have the permission to change contact
+
+        :return: The rendered template response
+        """
+        region = request.region
+
+        contact_instance = Contact.objects.filter(id=kwargs.get("contact_id")).first()
```
david-venhoff, line 99:
```suggestion
        contact_instance = Contact.objects.filter(id=kwargs.get("contact_id"), location__region=region).first()
```

integreat_cms/cms/views/contacts/contact_form_view.py
```diff
@@ -0,0 +1,145 @@
...
+            additional_instance_attributes={"region": region},
+        )
+
+        help_text = (
+            _("This location is used for the contact.")
+            if contact_instance
+            else _(
+                "Select a location to use for your contact or create a new location. Only published locations can be set."
+            )
+        )
```
david-venhoff, line 70: I think we typically put these things into `get_context_data`

integreat_cms/cms/views/contacts/contact_form_view.py
```diff
@@ -0,0 +1,145 @@
...
+
+        return render(
+            request,
+            self.template_name,
+            {
+                **self.get_context_data(**kwargs),
+                "contact_form": contact_form,
+                "poi": contact_instance.location if contact_instance else None,
+                "current_menu_item": "contacts",
```
david-venhoff, line 79: This can also go into `get_context_data`

integreat_cms/cms/views/contacts/contact_form_view.py
```diff
@@ -0,0 +1,145 @@
...
+            )
+
+        return render(
+            request,
+            self.template_name,
+            {
+                **self.get_context_data(**kwargs),
+                "contact_form": contact_form,
+                "poi": contact_instance.location if contact_instance else None,
+                "current_menu_item": "contacts",
```
david-venhoff, line 140: same here

integreat_cms/cms/views/events/event_form_view.py
```diff
@@ -112,6 +112,11 @@ def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
         )
 
         url_link = f"{settings.WEBAPP_URL}/{region.slug}/{language.slug}/{event_translation_form.instance.url_infix}/"
+
+        help_text = _(
+            "Create an event location or start typing the name of an existing location. Only published locations can be set as event venues."
+        )
+
```
david-venhoff, line 119: And this too

david-venhoff (changes requested):
Cool :tada:

I have a few suggestions

integreat_cms/cms/migrations/0103_update_contact.py
```diff
@@ -0,0 +1,66 @@
...
+        ("cms", "0102_alter_contact_poi"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="contact",
+            name="name",
+            field=models.CharField(blank=True, max_length=200, verbose_name="name"),
+        ),
+        migrations.RenameField(
```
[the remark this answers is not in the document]
david-venhoff: Oops, I didn't read the pr description very well :see_no_evil: 

I have opened an issue for this, so it won't have to be addressed by this pr: 3059

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -1798,6 +1798,34 @@ msgstr "Vierte Woche"
...
+"Telefonnummer stimmt nicht mit der vom ausgewählten Ort. Bitte prüfen Sie "
+"noch mal."
+
+#: cms/forms/contacts/contact_form.py
+msgid ""
+"Website URL does not match with that of the selected location. Please check "
+"again."
+msgstr ""
+"Website URL stimmt nicht mit der vom ausgewählten Ort. Bitte prüfen Sie noch "
+"mal."
```
PeterNerlich, line 1827: I advocate against these messages altogether, but if we keep them, I suggest the following wording:
```suggestion
#: cms/forms/contacts/contact_form.py
msgid ""
"E-mail address does not match with that of the selected location. Please "
"check again."
msgstr ""
"Die E-mail Adresse stimmt nicht mit der vom ausgewählten Ort überein. Bitte "
"überprüfen Sie sie erneut."

#: cms/forms/contacts/contact_form.py
msgid ""
"Phone number does not match with that of the selected location. Please check "
"again."
msgstr ""
"Die Telefonnummer stimmt nicht mit der vom ausgewählten Ort überein. Bitte "
"überprüfen Sie sie erneut."

#: cms/forms/contacts/contact_form.py
msgid ""
"Website URL does not match with that of the selected location. Please check "
"again."
msgstr ""
"Die Website URL stimmt nicht mit der vom ausgewählten Ort überein. Bitte "
"überprüfen Sie sie erneut."
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -2700,23 +2728,40 @@ msgstr "Nutzer-Chats"
...
+msgstr "Telefonnummer vom Ort"
+
 #: cms/models/contact/contact.py cms/models/pois/poi.py
 #: cms/models/users/organization.py
 msgid "website"
 msgstr "Webseite"
 
+#: cms/models/contact/contact.py
+msgid "website from location"
+msgstr "Website vom Ort"
```
PeterNerlich, line 2763: I thing this might be a bit more clear, but I'm not sure. Just *E-mail vom Ort* seems a bit weird to me
```suggestion
#: cms/models/contact/contact.py
msgid "E-mail from location"
msgstr "E-mail des Ortes"

#: cms/models/contact/contact.py
msgid "Will be taken over from the location."
msgstr "Wird vom Ort übernommen."

#: cms/models/contact/contact.py cms/models/pois/poi.py
msgid "phone number"
msgstr "Telefonnummer"

#: cms/models/contact/contact.py
msgid "phone number from location"
msgstr "Telefonnummer des Ortes"

#: cms/models/contact/contact.py cms/models/pois/poi.py
#: cms/models/users/organization.py
msgid "website"
msgstr "Webseite"

#: cms/models/contact/contact.py
msgid "website from location"
msgstr "Website des Ortes"
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -5155,6 +5189,91 @@ msgstr "Chat"
...
+msgstr "Kontakt wiederherstellen"
+
+#: cms/templates/contacts/contact_form.html
+#: cms/templates/contacts/contact_list_row.html
+msgid "Delete contact"
+msgstr "Kontakt löschen"
+
+#: cms/templates/contacts/contact_form.html
+msgid "Delete this contact"
+msgstr "Kontakt löschen"
```
PeterNerlich, line 5275: Is this a deliberate distinction? If not, we could merge these lines.
```suggestion
#: cms/templates/contacts/contact_form.html
#: cms/templates/contacts/contact_list_row.html
msgid "Delete contact"
msgstr "Kontakt löschen"

#: cms/templates/contacts/contact_form.html
msgid "Delete this contact"
msgstr "Diesen Kontakt löschen"
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -8904,6 +9001,40 @@ msgstr ""
...
+msgid "This location is used for the contact."
+msgstr "Dieser Ort wird für den Kontakt verwendet."
+
+#: cms/views/contacts/contact_form_view.py
+msgid ""
+"Select a location to use for your contact or create a new location. Only "
+"published locations can be set."
+msgstr ""
+"Wählen Sie einen Ort aus, der für ihren Kontakt verwendet werden soll oder "
+"erstellen Sie einen neuen Ort. Nur öffentliche Orte können verwendet werden."
```
PeterNerlich, line 9022:
```suggestion
#: cms/views/contacts/contact_form_view.py
msgid ""
"Select a location to use for your contact or create a new location. Only "
"published locations can be set."
msgstr ""
"Wählen Sie einen Ort aus, der für ihren Kontakt verwendet werden soll oder "
"erstellen Sie einen neuen Ort. Nur veröffentlichte Orte können verwendet "
"werden."
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -9056,6 +9182,15 @@ msgstr ""
...
 "vorlegen."
 
+#: cms/views/events/event_form_view.py
+msgid ""
+"Create an event location or start typing the name of an existing location. "
+"Only published locations can be set as event venues."
+msgstr ""
+"Erstellen Sie einen Ort oder beginnen Sie den Namen eines bestehenden "
+"Veranstaltungsortes einzutippen. Nur öffentliche Orte können als "
+"Veranstaltungsorte verwendet werden."
```
PeterNerlich, line 9192:
```suggestion
#: cms/views/events/event_form_view.py
msgid ""
"Create an event location or start typing the name of an existing location. "
"Only published locations can be set as event venues."
msgstr ""
"Erstellen Sie einen Ort oder beginnen Sie den Namen eines bestehenden "
"Veranstaltungsortes einzutippen. Nur veröffentlichte Orte können als "
"Veranstaltungsorte verwendet werden."
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -10761,6 +10896,15 @@ msgstr ""
 "Diese Seite konnte nicht importiert werden, da sie zu einer anderen Region "
 "gehört ({})."
 
+#~ msgid "Name of related POI"
+#~ msgstr "Name des verlinkten Ortes"
+
+#~ msgid "Only published locations can be set as event venues"
+#~ msgstr "Nur öffentliche Orte können verwendet werden"
```
PeterNerlich, line 10903:
```suggestion
#~ msgid "Only published locations can be set as event venues"
#~ msgstr "Nur veröffentlichte Orte können verwendet werden"
```

integreat_cms/static/src/js/events/event-query-pois.ts
```diff
@@ -42,6 +57,10 @@ const renderPoiData = (
         );
     document.getElementById("poi-query-result").classList.add("hidden");
     (document.getElementById("poi-query-input") as HTMLInputElement).value = "";
+    showContactFieldBox();
+    updateField("email", (document.getElementById("id_email_from_location") as HTMLInputElement).checked);
+    updateField("phone_number", (document.getElementById("id_phone_number_from_location") as HTMLInputElement).checked);
+    updateField("website", (document.getElementById("id_website_from_location") as HTMLInputElement).checked);
```
PeterNerlich, line 63: I interpret the comment *checkbox by default ausgewählt, wenn Daten ausgefüllt, sonst unchecked* in the design in that way that the checkboxes should be checked by default if and only if their corresponding fields in the POI are set when creating the contact.

Also the fields taken from the model should be sorted below those that are set explicitly, though I'm not sure this is worth actually implementing and would argue that it might just add confusion.

PeterNerlich (changes requested):
I agree with most of davids comments, so mine are only what caught my eye on top of them.

integreat_cms/cms/templates/contacts/contact_form.html
david-venhoff: What I thought was that it might make sense to label the button "Update" if the user is editing an existing contact and "Save" (or maybe "Create") if the user is creating a new contact

integreat_cms/cms/models/contact/contact.py
```diff
@@ -14,18 +14,38 @@ class Contact(AbstractBaseModel):
...
+        default=True,
+        verbose_name=_("phone number from location"),
+        help_text=_("Will be taken over from the location."),
     )
     website = models.URLField(blank=True, max_length=250, verbose_name=_("website"))
+    website_from_location = models.BooleanField(
+        default=True,
+        verbose_name=_("website from location"),
+        help_text=_("Will be taken over from the location."),
+    )
```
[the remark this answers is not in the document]
MizukiTemma: I'm not sure if it is then self explaining enough for users so I prefer to set a checkbox (or something similar). Also having those fields is probably helpful for #3033

integreat_cms/cms/templates/ajax_poi_form/poi_box.html
```diff
@@ -11,17 +11,24 @@
...
+    {% endif %}
     <div id="location-block"
          class="{% if form.has_not_location.value %}hidden{% endif %}">
         <label for="{{ form.location.id_for_label }}">
             {{ form.location.label }}
         </label>
-        {% translate "Name of event location" as poi_title_placeholder %}
+        {% translate "Name of location" as poi_title_placeholder %}
         <div class="relative my-2">
             <input id="poi-query-input"
```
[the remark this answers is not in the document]
MizukiTemma: I'll leave it to a separate issue #3064 as it affects the event form too.

integreat_cms/cms/templates/ajax_poi_form/poi_box.html
MizukiTemma: I know what I mean 😓 Though those elements have to stay outside of the box if they are put in the event form or contat form before or after {% include "poi_box.html" %} , or otherwise we need to reform the box itself 🤔 Do you see any handy fix idea?

integreat_cms/static/src/js/events/event-query-pois.ts
MizukiTemma: @nikolahoff 
These two behaviours are technically not simple 😿 and we think those may be rather confusing. Are they absolutely necessary?

image

integreat_cms/cms/forms/contacts/contact_form.py
```diff
@@ -0,0 +1,106 @@
...
+            ):
+                self.add_error(
+                    "website",
+                    forms.ValidationError(
+                        _(
+                            "Website URL does not match with that of the selected location. Please check again."
+                        ),
+                        code="invalid",
+                    ),
+                )
```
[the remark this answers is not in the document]
MizukiTemma: Ohw, I've totally forgotten this code 😅 As #3033 is now syncronising "the three fields" between POI and contact we don't need those checks anymore.

integreat_cms/static/src/js/events/event-query-pois.ts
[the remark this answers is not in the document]
MizukiTemma: I'm lazy: the file has been moved out of the folder 😁

integreat_cms/static/src/js/events/event-query-pois.ts
PeterNerlich: Checking whether the checkbox should be initialized as checked should be fairly simple though: We already have the information from the location, we copy it into the field after all. We can check if it is empty or not and check the box accordingly.

MizukiTemma: @PeterNerlich You're right😁 Though, changing the order depend on if it is filled in the POI, is (not impossible but tough).

But anyway, both are confusing in my opinion 🤔

integreat_cms/cms/models/contact/contact.py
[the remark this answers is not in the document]
PeterNerlich: Oh I totally agree on the user interface, I didn't mean to change that at all. I'm not sure what you mean with the fields being helpful – from the perspective of the model it should be just as simple to query for a field being `NULL` as for a bool being true or false

integreat_cms/cms/forms/contacts/contact_form.py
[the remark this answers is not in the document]
nikolahoff: @MizukiTemma why do you think it's confusing? The reason for changing the order is, that the user sees the unfilled fields easier and doesn't oversee which fields still need input. 
But if it's too complicated why can skip it. But would it be ok to have the 2 Pflichtfelder on top always?

integreat_cms/cms/models/contact/contact.py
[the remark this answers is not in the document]
MizukiTemma: An E-mail, phone number or website is given but the corresponding check box is checked and the given value does not match with that of the POI -such case must not happen in future, becuase 1) the field will be automatically filled with the data of the selected POI and not editable when the check box is checked (this is implemented in this PR) and 2) those data will syncrhonize when they must be taken from the POI #3033

And I'm not fun of adding an extra complexity or agreement. Even if we document it, we will fail to be careful enough to it someday, and differencisting between `Null` and `""` is prone to bugs (in my opinion)

integreat_cms/static/src/js/events/event-query-pois.ts
MizukiTemma: @nikolahoff 
>But if it's too complicated why can skip it. But would it be ok to have the 2 Pflichtfelder on top always?
Yes, title and name stay on top :)

MizukiTemma: image


🤔 ?

integreat_cms/cms/models/contact/contact.py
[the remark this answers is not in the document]
david-venhoff: I am also not a fan of having these `*_from_location` fields, but I also agree with Mizuki that the distinction between NULL and "" may be error prone. In my opinion, we should not have two models with redundant data in them. So I proposed an alternative to #3033: https://github.com/digitalfabrik/integreat-cms/issues/3033#issuecomment-2360829708

In that case, we would not need the `*_from_location` fields at all.

@MizukiTemma @PeterNerlich What are your opinions on that?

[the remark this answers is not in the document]
david-venhoff: No, what I meant was that we allow users to set anything they want in contacts, but prevent users from editing the email, website, phone number fields of pois. So your example scenario would still be no problem, as users could create multiple contacts with different phone numbers for the same poi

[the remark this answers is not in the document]
MizukiTemma: >prevent users from editing the email, website, phone number fields of pois.

🤔 ??

The details of POIs will not be changed from the contact form in the current implementation too. Users can only create a brand new POI (in case the wished one does not exist yet). E-mail, phone numebr and website URL given in the contact form will not be transfered to the POI.

integreat_cms/cms/forms/contacts/contact_form.py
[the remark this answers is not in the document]
PeterNerlich: @nikolahoff I see the idea behind changing the order to have the "less relevant" fields sorted last. I just feel like there might be a user story where someone goes through many contacts in a row, and then I think it's easy to get confused when the order of the fields changes between contacts.
I don't know how relevant that case is and how strong the benefits of ordering the fields that way are, but my gut feeling is that the effort implementing it is not worth it.

integreat_cms/cms/templates/ajax_poi_form/poi_box.html
MizukiTemma: Agree 😁 #3068

MizukiTemma:
@david-venhoff @PeterNerlich 
The check boxes and related codes are removed. Could you review again?

david-venhoff (changes requested):
Looks good to me! The only thing missing are the changes to which fields are required I think.

In the details block are some nits, but I don't think they are required to merge this :)
<details>
I noticed that `Pages` does not line up with `Title`
image

But I did not find a quick fix, and it is probably not that important :shrug:

When a user starts to edit a contact and then clicks on the archive or delete button, the unsaved warning will show up, which probably should not happen. But I guess this is also not important and probably would never happen ^^
</details>

integreat_cms/cms/templates/ajax_poi_form/poi_box.html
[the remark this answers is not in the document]
MizukiTemma: I suugest to remove them together with the corresponding fields of POI in #3086 for better traceability.

PeterNerlich:
> <details><summary>@david-venhoff: I noticed that `Pages` does not line up with `Title`. But I did not find a quick fix, and it is probably not that important 🤷</summary>
> 
> image
> 
> </details>

This is because for the last column, this note is in its own `div.p-4`, while the other columns have it in a `div.p-4` along with more content. The message and the following headline each have margins set, which don't add but collapse to the greater margin of both elements, while in the last column they each fully take up space, plus the padding from both `div.p-4`.

If we don't want to change `_related_contents_table.html` to have the outermost element only be a `div.w-full.table-listing` and push the responsibility of giving appropriate spacing to wherever it is being embedded *(I don't know whether we have any conventions regarding html snippets like this)*, we could change all columns to have the hint on top to uniformly be in a separate `div.p-4.pb-0`.

integreat_cms/cms/forms/contacts/contact_form.py
```diff
@@ -0,0 +1,91 @@
...
+        if not self.data["title"] and contacts_with_empty_title:
+            self.add_error(
+                None,
+                forms.ValidationError(
+                    _("Only one contact per location can have empty title."),
+                    code="invalid",
+                ),
+            )
+
+        return cleaned_data
```
david-venhoff, line 91:
```suggestion
```

The problem with doing the validation here, is that this code is potentially prone to race-conditions (If two forms are submitted at similar times, then for both forms this check could run simultaneously, see no errors, and write to the database). Instead, I'd propose to delegate the task of checking these conditions to the database, by adding two constraints to the contact model:

```python
        constraints = [
            models.UniqueConstraint("location", condition=Q(title=""), name = "contact_singular_empty_title_per_location", violation_error_message=_("Only one contact per location can have an empty title.")),
            models.CheckConstraint(check=~Q(title="")|~Q(name="")|~Q(email="")|~Q(phone_number="")|~Q(website=""), name="contact_non_empty", violation_error_message=_("One of the following fields must be filled: title, name, e-mail, phone number, website."))
        ]
```

david-venhoff (approved):
LGTM :tada:

I'd just propose to move the new contact constraints to the database layer, instead of performing them at the form-level.

integreat_cms/cms/views/contacts/contact_form_view.py
```diff
@@ -0,0 +1,148 @@
...
+            contact_form.save()
+            if not contact_instance:
+                messages.success(
+                    request,
+                    _('Contact "{}" was successfully created').format(
+                        contact_form.instance
+                    ),
+                )
+            elif not contact_form.has_changed():
+                messages.info(request, _("No changes detected, but date refreshed"))
```
PeterNerlich, line 121:
```suggestion
                messages.info(request, _("No changes detected"))
```

PeterNerlich (approved):
Looks good to me. It feels nice to use with the way we settled to do the constraints now, I think revisiting those was worth it. Thank you very much!

Issue #3031, opened 2024-08-30
`Uncaught TypeError` in the event list


MizukiTemma:
### Describe the Bug

Everything works, but `Uncaught TypeError` is raised when visiting the event list.

### Steps to Reproduce

1. Go to the event list in any region
2. See `Uncaught TypeError` in the browser console

### Expected Behavior

No error

### Actual Behavior

Sadly an error

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

PR #3036, opened 2024-09-05
Fix Uncaught TypeError in the event list

MizukiTemma:
### Short description

This PR fixes `Uncaught TypeError` which occurs when the event list is visited. 

### Proposed changes

- Remove the code lines which contains the cause of the error `node.closest("li")`.

### Side effects

- I didn't find the case when `node.closest("li")` is **not** `null`. If thre is any case it is not `null`, this code breaks the current behaviour 😅 

### Resolved issues

Fixes: #3031 


theresantonie (approved):
Works as expected!

JoeyStk (approved):
I agree. I don't think this code was in used, so good thing we delete it :)

PR #3039, opened 2024-09-11
Change Zammad ticket type

svenseeberg:
### Short description
Allow Zammad users to answer to chat messages / tickets and remove default mention.

### Proposed changes
- Set the Zammad ticket type to "web" instead of "chat".
- Remove default handler mention as this overrides the users personal notification settings.

### Side effects
None. This should only affect Zammad.

### Resolved issues

Fixes: https://github.com/digitalfabrik/integreat-chat/issues/20


svenseeberg:
Issues now have the reply link:

image

... which returns valid messages to the front end:

image

Issue #3041, opened 2024-09-11
[Blocks release] Ugly display of poi list (and possibly others)


JoeyStk:
### Describe the Bug

After #2895 the list view of POI is ugly. This is due to the line in SCSS ` width: inherit !important;`. This needs to be reintroduced, preferable with a nice comment to help us in the future

### Steps to Reproduce

1. Go to CMS test (not a problem on production)
2. Click on POI
3. See ugly display

### Expected Behavior

Nice display

### Actual Behavior

Ugly display

### Additional Information

![grafik](https://github.com/user-attachments/assets/303b5d2b-4fba-49ef-aef2-91da3dc123bc)

PR #3044, opened 2024-09-11
Fix poi table layout

david-venhoff:
### Short description
Before #2895, we had [a css rule](https://github.com/digitalfabrik/integreat-cms/pull/2895/files#diff-d0f24712a2bf2e307d18ff548e8aa7747ef3be932bb9456c0262b32c87285e65L337) that set every table listing table's width to `inherit`. This caused `table-fixed` to have no effect, because using `table-fixed` without a specific size is an error in css and causes css to fallback to `table-auto`, which is the default value.
Now, that every table has `w-full` though, this rule is valid and causes the table to get too large. Fortunately, simply removing this rule fixes the table layout.

Additional context: https://github.com/digitalfabrik/integreat-cms/pull/2895#discussion_r1744433126

### Proposed changes

- Remove the `table-fixed` class

### Side effects

- Should be none

### Resolved issues

Fixes: #3041

### DIFF
--- a/integreat_cms/cms/templates/pois/poi_list.html
+++ b/integreat_cms/cms/templates/pois/poi_list.html
@@ -44,7 +44,7 @@
         <form id="bulk-action-form" method="post">
             {% csrf_token %}
         </form>
-        <table class="w-full mt-4 rounded border border-gray-200 shadow bg-white table-fixed">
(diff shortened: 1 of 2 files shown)


MizukiTemma (approved):
:+1:

Issue #3040, opened 2024-09-11
Machine translations for Integreat Chat messages


svenseeberg:
### Motivation
We want to enable a bi-directional communication without (or with low) language barriers between Integreat App users and councilors using the Zammad back end for the Integreat Chat.

### Proposed Solution
- This feature can/should make use of the existing machine translation functionality, including the budget limitations (for the time being).
- Incoming messages from Integreat App users should be translated into the main language of the region.
- Outgoing messages from Zammad to the Integreat App users should be translated into the selected language of the Integreat App. The CMS has to provide a Webhook endpoint that allows Zammad to inform the Integreat CMS about updated tickets (by Zammad users / counselors). The webhook message from Zammad contains a `ticket_id` attribute, which can be used by the Integreat CMS to fetch all necessary messages / data from Zammad.
- Translated messages have to be stored along the original messages in Zammad as new articles within the ticket.
- This should not require a change of any Integreat CMS model.

Sequence diagram:

![image](https://github.com/user-attachments/assets/69bc8546-72b2-4363-b2a0-ee4d44966162)

To implement the functionality, we need to add the green marked flows and change the yellow ones. Black is already implemented and blue is out of scope / next milestone.

There is one big issue / caveat: Zammad webhooks are triggered on each ticket update. By default, this does not differentiate between ticket updates by a counselor or "automated" updates by the Integreat CMS (when storing translations). To prevent and endless loop of "update webhook > translation request > ticket update", we need to set the article subject to `automatically generated message` or some other string we agree upon. This can then be used to set up a filter for webhooks within Zammad.

### Alternatives
Zammad users translate the messages manually with DeepL or Google Translate.

### User Story
- As a migrant I want to communicate with a counselor in my native language so that I can get necessary information before having learned any German.
- As a counselor I want to answer questions of migrants without speaking their native language to provide answers to important questions.

### Additional Context
* See https://github.com/digitalfabrik/integreat-chat/issues/17

### Design Requirements
None / no GUI impact.

PR #3051, opened 2024-09-14
Chat: automatic answers & translations

svenseeberg:
### Short description
Implement machine translations for chat & retrieval augmented answer generation

### Proposed changes

- Create a relation between chats, the region and language
- Make the Zammad URL unique
- Create a new end point for Zammad Webhooks, which is used to trigger translations and automatic answers
- Add automatic translations for incoming and outgoing messages
- Add a property that informs which chat messages were created automatically
- Add automatic answers to app user messages

### Side effects
- Existing chats are not assigned to any region and therefore not accessible anymore. As we only had an alpha testing phase, this should be of no concern.

### Resolved issues
Fixes: #3040


integreat_cms/api/urls.py
```diff
@@ -219,6 +219,7 @@
 #: The url patterns of this module (see :doc:`django:topics/http/urls`)
 urlpatterns: list[URLPattern] = [
     path("api/v3/regions/", include(region_api_urlpatterns)),
+    path("api/v3/webhook/zammad/", user_chat.zammad_webhook, name="zammad_webhook"),
```
svenseeberg, line 222: Please note: `/api/v3/webhook/...` is a generic new path for webhooks.

integreat_cms/cms/models/regions/region.py
```diff
@@ -417,6 +417,7 @@ class Region(AbstractBaseModel):
         max_length=256,
         blank=True,
         default="",
+        unique=True,
```
svenseeberg, line 420: We cannot allow to regions to use the same Zammad server, because we need the reverse look up of the Zammad server IP during webhooks to get the region to which the Zammad belongs. This is also a good idea for privacy reasons.

integreat_cms/cms/migrations/0103_userchat_language_userchat_region_and_more.py
```diff
@@ -0,0 +1,53 @@
...
+    dependencies = [
+        ("cms", "0102_alter_contact_poi"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="userchat",
+            name="language",
+            field=models.ForeignKey(
+                default=3,
```
svenseeberg, line 21: All existing chat which were created during the first test phase, will be assigned the German language. There is no other good solution.

integreat_cms/api/v3/chat/zammad_api.py
```diff
@@ -90,23 +93,26 @@ def _parse_response(self, response: dict | list[dict]) -> dict | list[dict]:
             return [self._parse_response(item) for item in response]  # type: ignore[misc]
 
         if author := response.get("sender"):
-            response["user_is_author"] = author == "Customer"
+            response["user_is_author"] = (
+                author == "Customer" and response["subject"] != AUTO_ANSWER_STRING
+            )
+        # response["automatic_answer"] = response["subject"] == AUTO_ANSWER_STRING # TODO: Why does this not work?
```
svenseeberg, line 99: We get a key error here. That probably means that the subject attribute does not exist.

MizukiTemma (approved):
🚀

svenseeberg:
> This pull request will bring the total coverage in the repository to **82.6%** (**-0.2%** change).

This hurts a bit, I guess :see_no_evil:

integreat_cms/api/urls.py
david-venhoff: Could you add some documentation for that at https://github.com/digitalfabrik/integreat-cms/blob/develop/docs/src/api-docs.rst?

integreat_cms/api/v3/chat/user_chat.py
```diff
@@ -99,6 +105,18 @@ def get_messages(
...
 
+def translate_message(message: str, language_slug: str) -> str:
+    """
+    Translate a string with Google Translate
+    """
+    credentials = service_account.Credentials.from_service_account_file(
+        settings.GOOGLE_APPLICATION_CREDENTIALS
+    )
+    translate_client = translate.Client(credentials=credentials)
+    result = translate_client.translate(message, target_language=language_slug)
```
david-venhoff, line 116: Should we track these translations and add them to the regions translation budget, or are they free for the regions?

david-venhoff (commented):
Not a full review, just some questions I got from a quick look.

Also, is there / should there be any protection in place to avoid unauthorized people using this api?

integreat_cms/api/v3/chat/user_chat.py
svenseeberg: For the time being it is "free". We may want to host our own translation model for this purpose soon. Also something that made me think for a moment and what I had not yet considered thoroughly enough: authentication of requests for translation purposes is done via the reverse look up. I'll add a get_or_404() for resolving the Zammad server IP to a region in https://github.com/digitalfabrik/integreat-cms/pull/3051/files#diff-a9958a0ba0d1f97dd15f8eb15d9fa3723486384f9f04b0f2bd6ba7ff48a3a177R247

PR #3052, opened 2024-09-14
Do not show translation status of the root language

MizukiTemma:
### Short description

This is a revival of #2436 

- Do not show root language column in the event(page/poi list.
- Add a new flag for German (easy) to distinguish from German

### Resolved issues

Fixes: #2389 


timobrembeck (commented):
Probably, we need to update the library first to make use of the new flags, right?

MizukiTemma:
@timobrembeck Thank you for the release in [flagpack-dart-sass](https://github.com/timobrembeck/flagpack-dart-sass) 👍

PeterNerlich (approved):
Seems to work perfectly

david-venhoff (changes requested):
Looks good to me :tada: 

I only noticed that the changes don't apply to news yet (There the root language is still shown)

Also, somebody needs to make sure that we change the German (easy) flag once this pr lands in production

MizukiTemma:
@david-venhoff nice catch 👀 The same changes are now applied for news and templates 😸

david-venhoff (commented):
Oh, I found another problem :sweat_smile: 

We also show the translation status in the archived lists (archived pages, archived events, but not in archived pois for some reason)

I see two solutions here:
* Keep archived views as-is and also hide the root language there
* Remove the translation status indicator from archived views to make them more consistent with the archived poi list

MizukiTemma:
@david-venhoff 
I'd go for the option 1 to keep everything out of the issue scope as is. An issue is opened to fix this inconsistency (#3071 )

david-venhoff (approved):
:tada:

Issue #2173, opened 2023-03-27
Add SUMM.AI budget for updating translations


timobrembeck:
### Motivation

We offer municipalities the option to purchase an extra 10.000 words for SUMM.AI update translations

### Proposed Solution

- [ ] Create a new setting `SUMM_AI_CREDITS_ADDON` to define the amount for the SUMM.AI package, similar to `DEEPL_CREDITS_ADDON`
- [ ] Add new fields `summ_ai_addon_booked`, `summ_ai_renewal_month` and `summ_ai_budget_used` to the region model, similar to `deepl_addon_booked`, `deepl_renewal_month` and `deepl_budget_used`
- [ ] Whenever the content is translated via the new checkbox in the form, increase the used words budget of the region, see:
- [ ] #2110
- [ ] Show a graph of the used budget in the [machine translation settings page](https://github.com/digitalfabrik/integreat-cms/blob/573a3c80db1ce9cb1995a5ffb527c7022c2be27e/integreat_cms/cms/views/translations/translations_management_view.py)

The initial translations via the bulk action or the management command should not drain the region's budget.

osmers:
Just a small comment - I'm pretty sure we'd need this by July 1st as well since we offer the possibility of SUMM AI update translations from that point onwards.

dkehne:
I will put that back into backlog because it is not as urgent as we thougt

osmers:
Unsure of what the prio is but since a municipality ask about their budget, and we cannot tell them anything, I feel like we should implement it rather soonish - at least the part with which we can keep track of the budget used. Displaying it in the region can come later...

PR #3065, opened 2024-09-18
Budget for Summ AI

PeterNerlich:
### Short description

This PR adds separate budget accounting for Summ AI. Until now the usage of Summ AI was not metered at all by us.

### Proposed changes

- **Add `summ_ai_budget_used`** to the `Region` model, as well as various auxiliary properties and budget related changes analogous to machine translations
- Update `summ_ai_budget_used` of the region accordingly after a batch of translations was performed
  *This required a change to how the asynchronous translation flow works:*
- Instead of putting all tasks into the `PatientTaskQueue` from the start, more are added every time the queue gets low enough
  - *This is because individual tasks don't represent entire content translation objects, but individual snippets (e.g. a single paragraph), since Summ AI doesn't handle HTML content (well enough for our purposes)*
  - *We want to avoid sending Summ AI requests that we get billed for and then finding out that the region didn't have the translation budget for it*
- Introduce the **`BudgetEstimate` class** to keep track of "allocated" budget, via an instance shared with each `TranslationHelper` – the helper class representing a single content translation object responsible for splitting it up into single translatable units and splicing it back together after translation
- **Pass `TranslationHelper` instead of `TextField` instances** around to batch translation tasks per content translation
- **Only add all translation tasks** necessary for a content translation **if it still fits into the budget** according to `BudgetEstimate`
- This on-demand expansion of the queue is facilitated by means of a **management task**, during its run any workers trying to fetch a task from an empty queue won't quit because it is likely that more tasks will be available a moment later, and the worker pool shouldn't shrink from such race conditions
- The `PatientTaskQueue` is initialized empty and `worker()` is run directly to fill it before the workers are initialized

### Side effects

- Machine translations with DeepL or Google Cloud Translate are using up a regions budget whether they are successful or not. Translations with Summ AI are metered in the same way, but whether there were retries because of API rate limiting as well as whether translations were not attempted at all because the budget was already used up is ***not taken into account***. However, I think this should not pose too much of a problem.
- This is a particularly confusing section of our code base, and I play not a small role in this. I hope I added enough and helpful enough **comments**, please tell me if you think this could be improved!
- This PR has become so large that we **didn't include tests**. The existing tests still succeed and guarantee the translation functionality, but the budget metering is not covered.

### Resolved issues

This fulfills the first half of #2173 


charludo (changes requested):
I am sorry, but this needs a major architectural overhaul. Apart from the suggestions below, the main reason I am saying this is the following:

- when `SummAiApiClient.translate_queryset` is called,
- a `lambda` function returning a closure calling `SummAiApiClient.check_usage` (which overrides `MachineTranslationApiClient.check_usage`) is used in the creation of a `BudgetHelper` object,
- which is used for the instantiation of a `TranslationHelper` object,
- which has `TranslationHelper.check_usage` method, 
- which calls its embedded `BudgetHelper.check_usage` method,
- which calls `BudgetHelper._check_usage`,
- which is the aforementioned closure.

Now back up the chain:

- `BudgetHelper.check_usage` is only called by `BudgetHelper.allocate`,
- which is only called by `TranslationHelper.allocate_budget`,
- which is only called by `SummAiApiClient.translate_text_fields`,
- which is only called by `SummAiApiClient.translate_queryset`.

Frankly, I question the need for the `BudgetEstimate` class, as well as the helper methods on `TranslationHelper`.

However, it is difficult to give more directed feedback, simply because of the amount of obfuscation going on due to these enormous call chains.

I have not yet had a chance to examine the `async` functions.

JoeyStk:
As this needs some major refactoring and time is sparse at the moment, it'll take some time until we will get back to this

JoeyStk:
I think it's time to close this PR, as we have another attempt on fixing this issue with #3586

PR #3075, opened 2024-09-25
Refactor repair_tree as utility

PeterNerlich:
*This PR ~~builds on and includes the commits of #2596 and should only be merged after that PR and a subsequent rebase~~ is ready to merge.*

### Short description

Make repair_tree into a utility that can be used automatically. Attempt to repair the tree and try again whenever a page move fails.

*I am absolutely open to discussions about how the scheme with `lft`/`rgt`/`tree_id` that treebeard/MPTT uses works. It seems confusing at first but is quite simple when you wrap your head around it and I would be glad to help and accelerate that.*

### Proposed changes

- Split core functionality of repair_tree command into cms app utils
  - Add MPTT Fixer class that calculates new lft/rgt/depth values for all tree at once
  - Add ShadowInstance class that can shadow a model, such that any changes to it have to be explicitly applied and can be diffed
- Automatically attempt to fix tree if a page move operation fails and try again
- Add tests for the repair_tree utility and management command
- Adjust the migration of #2596 to run `repair_tree()` before adding the constraints

### Missing

- Front end error messages

### Side effects

- Calling the utility automatically might not always have the desired effect
    - `DBMutexTimeoutError` and `DBMutexError` were retrofitted as the exceptions to cause a "next try" while splitting this PR off from #2596. It is really intended to catch exceptions thrown by `@tree_mutex()` coming in the third split off PR (#3076).

Might have slightly positive effects in regards to #1518


david-venhoff:
Adding prio: urgent as this is a release-blocker

david-venhoff (commented):
I loaded the database dump and tried to run on this branch, but the migration still failed unfortunately.
Error is
```
django.db.utils.IntegrityError: could not create unique index "page_unique_lft_tree"
DETAIL:  Key (tree_id, lft)=(7057, 11) is duplicated.

```

PeterNerlich:
I think I found and fixed a bug in `repair_tree()`. I decided it's better to try and recreate the hierarchical structure first and then, when all nodes are already discovered, recalculate the values for `lft`, `rgt` and `depth`.

I think the problem was pages being discovered out of order which the original implementation didn't account for, e.g. under a root page some page *A*, a sub-page of *A*, a page *B*, and then again a sub-page of *A* – at this point, the left and right values for that sub-page would be calculated and the `rgt` value of all ancestors would be updated, but *B* would have been left out.
This didn't pose much of a problem because of db level ordering, except when the `tree_id` was differing from that of the node set as parent. Then those nodes were being inserted after the fact, either because the tree of the parent had been discovered much earlier or because it would have been discovered later and the orphaned nodes had been pushed to the very end of the queue.

integreat_cms/cms/utils/repair_tree.py
```diff
@@ -0,0 +1,210 @@
...
+    for root_node in root_nodes:
+        action = "Fixing" if commit else "Detecting problems in"
+        logger.info(
+            "%s tree with id %i... (%r)",
+            action,
+            root_node.tree_id__original,
+            root_node.instance,
+        )
+        for tree_node in mptt_fixer.get_fixed_tree_of_page(root_node.pk):
+            print_changed_fields(tree_node, logging_name=logging_name)
```
david-venhoff, line 69: Instead of printing the entire tree structure, wouldn't it be more useful to just print the changed nodes?

integreat_cms/cms/utils/repair_tree.py
```diff
@@ -0,0 +1,210 @@
...
+    logger = logging.getLogger(__name__)
+
+    def __init__(self, dj_apps: Apps = apps) -> None:
+        """
+        Create a fixed tree, using a :class:`~integreat_cms.cms.utils.shadow_instance.ShadowInstance` of each page.
+        """
+        Page: Model = dj_apps.get_model("cms", "Page")
+        self.broken_nodes: deque[ShadowInstance[Page]] = deque(
+            map(ShadowInstance, Page.objects.all())
+        )
```
david-venhoff, line 113:
```suggestion
        self.broken_nodes: deque[ShadowInstance[Page]] = deque(ShadowInstance(page) for page in Page.objects.all())
```
nit

integreat_cms/cms/utils/repair_tree.py
```diff
@@ -0,0 +1,210 @@
...
+            try:
+                Page.objects.get(id=single_id)
+            except Page.DoesNotExist as e:
+                raise ValueError(
+                    f'The page with id "{single_id}" does not exist.'
+                ) from e
+        # All ids in page_id are valid, get them from mptt_fixer
+        root_nodes = [
+            mptt_fixer.get_fixed_root_node(single_id) for single_id in page_id
+        ]
```
david-venhoff, line 56:
```suggestion
        root_nodes = {
            mptt_fixer.get_fixed_root_node(single_id) for single_id in page_id
        }
```

david-venhoff (commented):
Now migration the prod data works :tada: and the general approach looks good to me.

One point we briefly talked about is whether the migration should call `repair_tree`. I am tending towards no, because we cannot guarantee that this solution will work in the future and I think it is fine to require some manual action here to fix the inconsistencies.

I think we should also merge this relatively quickly, so that we can do some extensive testing on the test server, whether the pages are still correct.

integreat_cms/cms/utils/repair_tree.py
PeterNerlich: In principle yes, but I think then it would just essentially be the same as "*I found something!*" and not much more helpful without further manual investigation of the tree structure in the database. When all nodes of the tree are printed, you have enough context to say "*Yes that makes sense*" or "*Oh now I know why something went wrong in the way it did*"

PeterNerlich:
> One point we briefly talked about is whether the migration should call `repair_tree`. I am tending towards no, because we cannot guarantee that this solution will work in the future and I think it is fine to require some manual action here to fix the inconsistencies.

I think this is **easily automatable** and therefore should not require intervention for admins anywhere in the future.
I assume you argue *we cannot guarantee that this solution will always work* because someone might change the implementation of `repair_tree()` in the future, which might break the behaviour when calling it for this migration. But if it is a **manual step**, this is **almost the same**: We cannot guarantee that the repair_tree management command *(or any comparable utility)* will be available for the admin in the future. And having them understand an ancient form of storing the data of a model that might have long changed its form and purpose – *and on top of that which problems that storage format might have had and how to return the data to 100% integrity* – can hardly be a better solution.

In my opinion, **any future PR** touching the implementation of `repair_tree()` is also responsible for **checking for compatibility** with its usage in the migration, and for moving the old code into this migration to be able to independently move on for the project itself, if necessary.

david-venhoff:
> In my opinion, any future PR touching the implementation of repair_tree() is also responsible for checking for compatibility

I think the problem goes further than that (although it might be worth to test this first): The repair_tree function uses the `Page` model as defined for django, in a migration. This is should be very problematic, as the `Page` model always encodes the latest database version, which will not be the case for long if we run it during the migration. 
Consider that we rename, add, delete some fields on the page model in the future, how should django figure out how to convert the ancient data returned by postgres into very different `Page` model?

PeterNerlich:
> The repair_tree function uses the `Page` model as defined for django, in a migration. This is should be very problematic, as the `Page` model always encodes the latest database version

No, this shouldn't be a problem, as the current state of the model according to the current point inbetween the migrations is passed through to `repair_tree()` and the page model is retrieved with
```python
Page: Model = dj_apps.get_model("cms", "Page")
```
instead of importing the current class of the model directly

jarlhengstmengel:
I think it's good. I tested it locally with altering some pages manually and it worked fine and I see no logical inconsistencies. I'm unsure how to test if migrating works. I don't really have a opinion on the discussion about calling `repair_tree` in the migration. I'm missing a bit of experience with migrations to have a strong stance on that

integreat_cms/cms/migrations/0105_languagetreenode_languagetreenode_unique_lft_tree_and_more.py
```diff
@@ -1,8 +1,42 @@
...
+
+if TYPE_CHECKING:
+    from django.apps.registry import Apps
+    from django.db.backends.base.schema import BaseDatabaseSchemaEditor
+
+
+# pylint: disable=unused-argument
+
+
+def forwards_func(
```
david-venhoff, line 20: This does not attempt to fixes broken languagetreenode objects. I guess we can leave fixing that up to the users of the cms, given that it should be really unlikely?

integreat_cms/cms/migrations/0105_languagetreenode_languagetreenode_unique_lft_tree_and_more.py
david-venhoff, line 1: I know that this pr did not create this migration, but while we are at it, could you rename the migration to be a bit more descriptive?

integreat_cms/cms/migrations/0105_languagetreenode_languagetreenode_unique_lft_tree_and_more.py
david-venhoff, line 20:
```suggestion
# pylint: disable=unused-argument
def forwards_func(
```

david-venhoff (approved):
The code looks sound to me, but I think we should do some thorough testing on the test server, given that this migration touches pretty much every tree node.

Thanks a lot for the effort!

integreat_cms/cms/migrations/0105_languagetreenode_languagetreenode_unique_lft_tree_and_more.py
PeterNerlich: I think we did not experience any kind of compromised integrity with language tree nodes yet. But yes, ideally `repair_tree()` and the migration should also handle them. However, I'm not sure whether the extra effort is worth it

integreat_cms/cms/migrations/0105_languagetreenode_languagetreenode_unique_lft_tree_and_more.py
PeterNerlich: Can we simply do that? I know that the migration files specify dependencies by the full name, not only the leading, continuously incrementing number. I do not know how this works on the db side. Does Django save only the number to determine what state the database is in? My guess is "unlikely…?"

integreat_cms/cms/migrations/0105_languagetreenode_languagetreenode_unique_lft_tree_and_more.py
PeterNerlich: I intentionally placed the newlines between the pylint directive and the function definition. Previously I was under the misconception that placing such a comment immediately before a function definition would disable the specified checks for only that function definition. We were very surprised to find out that instead, it disabled it for the remainder of the file *(i.e. also for `reverse_func()` and everything in `Migration` though that is already compliant)*. Leaving blank newlines best communicates that behaviour imho.
The only alternative to specifying it for the rest of the file is to put the comment at the end of the line for which the check should be disabled *(and it only affects the very line, not the statement it belongs to, which is why I didn't just put it next to the pretty printed function definition as I would then have to specify it multiple times for the same statement)*.

jarlhengstmengel (approved):
I tested the migration with the database dump now too and it went through. So looks good to me!

integreat_cms/cms/migrations/0105_languagetreenode_languagetreenode_unique_lft_tree_and_more.py
david-venhoff: Regardless of that, I thought it would be fine to make braking changes if we did not have a release yet?

integreat_cms/cms/migrations/0105_languagetreenode_languagetreenode_unique_lft_tree_and_more.py
david-venhoff: Wow, that's crazy :open_mouth: 

According to the [pylint documentation](https://pylint.pycqa.org/en/latest/user_guide/messages/message_control.html#block-disables), the correct way to do it is to place the directive inside of the function, because a directive applies to its containing block...

I guess we should create a issue to fix all of our pylint directives then? :laughing:

PeterNerlich: Oooh, so to only apply to the next line you need `# pylint: disable-next=…` instead of `# pylint: disable=…` – this might just have started as an oversight then

PeterNerlich: Okay, this is interesting. I prefer to have `reverse_func()` in for completeness sake, but just using
```python
def reverse_func(
    apps: Apps,
    schema_editor: BaseDatabaseSchemaEditor,
) -> None:
    """
    Data migration when reverting this migration
    (nothing to do since the data is already consistent)
    """
```
apparently *doesn't* count the disable towards the block – I get a warning for each of the unused arguments, and then one for the useless suppression of `useless-argument`.
If I only add a `pass` to the end, the comment *does* count towards the block, and pylint only explains to me that I really don't need the `pass` statement, since the docstring is a multiline string that already counts as a statement, making the function not empty already. :grin:

integreat_cms/cms/migrations/0105_languagetreenode_languagetreenode_unique_lft_tree_and_more.py
PeterNerlich: True, the other PR didn't land in a release

PeterNerlich: @david-venhoff What do you think about `0105_treebeard_models_add_deferrable_tree_constraints`?

PR #3076, opened 2024-09-25
Add @tree_mutex() decorator and trick treebeard into using transactions

PeterNerlich:
### Short description

Force treebeard to use transactions. This should remove any race conditions caused by the library. The tests seem promising.

### Proposed changes

- Add `@tree_mutex()` decorator to replace `@django.db.transactions.atomic` (still uses it inside in addition to the mutex implementation)
  - Monkeypatch treebeards Node class to use djangos database cursor for its raw SQL queries so that changes get rolled back on failure
- Add tests for `@tree_mutex()`

### Side effects

- The system might respond slower when many page move operations are attempted simultaneously, because treebeard now waits with each operation until any previous operations are finished

### Resolved issues

Fixes: #1518, possibly #1716

### DIFF
--- a/integreat_cms/cms/models/abstract_tree_node.py
+++ b/integreat_cms/cms/models/abstract_tree_node.py
@@ -1,11 +1,16 @@
+"""
+.. warning::
+    Any action modifying the database with treebeard should use ``@tree_mutex(MODEL_NAME)`` from ``integreat_cms.cms.utils.tree_mutex``
+    as a decorator instead of ``@transaction.atomic`` to force treebeard to actually use transactions.
+    Otherwise, the data WILL get corrupted during concurrent treebeard calls!
+"""
+
 from __future__ import annotations
 
 import logging
 from typing import TYPE_CHECKING
 
 from cacheops import invalidate_model
-from db_mutex import DBMutexError, DBMutexTimeoutError
-from db_mutex.db_mutex import db_mutex
 from django.db import models
 from django.db.models import CheckConstraint, Deferrable, F, Q, UniqueConstraint
 from django.utils.functional import cached_property
@@ -272,41 +277,29 @@ class AbstractTreeNode(NS_Node, AbstractBaseModel):
         :param pos: The new position of the page relative to the target
                     (choices: :mod:`~integreat_cms.cms.constants.position`)
         :raises ~treebeard.exceptions.InvalidPosition: If the node is moved to another region
-        :raises ~db_mutex.exceptions.DBMutexError: If the DB mutex could not be retrieved
-        :raises ~db_mutex.exceptions.DBMutexTimeoutError: If waiting for the DB mutex timed out
         """
         logger.debug("Moving %r to position %r of %r", self, pos, target)
-        try:
-            with db_mutex(self.__class__.__name__):
-                # Do not allow to move a node outside its region, but allow
-                # moving as siblings of root nodes (because it's a separate tree)
-                if self.region != target.region and not (
-                    target.is_root()
-                    and pos
-                    in [
-                        position.LEFT,
-                        position.RIGHT,
-                        position.FIRST_SIBLING,
-                        position.LAST_SIBLING,
-                    ]
-                ):
-                    raise InvalidPosition(
-                        _(
-                            'The node "{}" in region "{}" cannot be moved to "{}".'
-                        ).format(self, self.region, target.region)
-                    )
-                # Moving a node can modify all other nodes via raw sql queries (which are not recognized by cachalot),
-                # so we have to invalidate the whole model manually.
-                invalidate_model(self.__class__)
-                super().move(target, pos)
-        except DBMutexError as e:
-            raise DBMutexError(
-                _('Could not change position in tree of "{}".').format(self)
-            ) from e
-        except DBMutexTimeoutError as e:
-            raise DBMutexTimeoutError(
-                _('Could not change position in tree of "{}".').format(self)
-            ) from e
+        # Do not allow to move a node outside its region, but allow
+        # moving as siblings of root nodes (because it's a separate tree)
+        if self.region != target.region and not (
+            target.is_root()
+            and pos
+            in [
+                position.LEFT,
+                position.RIGHT,
+                position.FIRST_SIBLING,
+                position.LAST_SIBLING,
+            ]
+        ):
+            raise InvalidPosition(
+                _('The node "{}" in region "{}" cannot be moved to "{}".').format(
+                    self, self.region, target.region
+                )
+            )
+        # Moving a node can modify all other nodes via raw sql queries (which are not recognized by cachalot),
+        # so we have to invalidate the whole model manually.
+        invalidate_model(self.__class__)
+        super().move(target, pos)
         invalidate_model(self.__class__)
 
         # Reload 'self' because lft/rgt may have changed
--- a/integreat_cms/cms/utils/tree_mutex.py
+++ b/integreat_cms/cms/utils/tree_mutex.py
@@ -1,0 +1,97 @@
+"""
+This module contains a custom decorator for db / redis mutexes
+"""
+
+import functools
+import threading
+from contextlib import contextmanager
+from typing import Callable, Generator, ParamSpec, Type, TypeVar
+
+from django.db import DEFAULT_DB_ALIAS, transaction
+from treebeard.models import Node
+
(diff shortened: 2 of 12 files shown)


jarlhengstmengel (changes requested):
Thank you a lot for the PR! I think it looks quite good. My main concern is, that I'm not sure when to use the tree_mutex decorator. Should it be used every time a method changes something in the language or page tree? Or at which point is it needed to prevent inconsistencies?

PeterNerlich:
> I'm not sure when to use the tree_mutex decorator. Should it be used every time a method changes something in the language or page tree?

In principle, since treebeard uses neither the db cursor from django nor creates their own transactions for anything it does, it makes sense to do the monkey patching for **any function of our business logic that calls treebeard methods** *(at least those modifying the database)*.

Then, what about putting the transaction only as closely around the call to treebeard as possible *(as opposed to including all the business logic to process that request)*? Well, that would work, but if we get an exception halfway through our business logic after an initial page move or something, it might leave the database in an inconsistent state – after all, the transaction for the treebeard call was successfully committed. When the transaction is around the whole request handler instead, if any exception gets raised and not caught and dealt with inside, any change that was done to the database will be rolled back. I think this is the cleanest solution and it happens to be the easiest.

integreat_cms/cms/views/bulk_action_views.py
```diff
@@ -269,6 +270,7 @@ class BulkArchiveView(BulkActionView):
     Bulk action for restoring multiple objects at once
     """
 
+    @tree_mutex("page")
```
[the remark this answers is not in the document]
PeterNerlich: I mean, leaving it should not hurt it, but we could separate it out / only conditionally apply it somehow

charludo:
Thank you for tackling this complicated issue!! 🎉 

I have a question somewhat along the lines of what you and Jarl already discussed above. The "danger" I see is that we might forget to add the mutex decorator somewhere, or not even be aware that it is required. 

Have you considered making kind of a "middleware" between  trebeard and integreat? Something like importing treebeard, applying the mutex, and then ensuring that we only import from that middleware, never from treebeard itself? I guess `@transation.atomic` would still be required then, as you pointed out above. 

OK, thinking about it again, it's probably very little benefit, ~~apart from giving the opportunity to name that middleware "quickbeam" :D~~

integreat_cms/cms/utils/tree_mutex.py
```diff
@@ -0,0 +1,97 @@
...
+from treebeard.models import Node
+
+#: For how many seconds the lock persists, and the timeout for retrying to acquire it.
+LOCK_SECONDS = 10
+#: How long to sleep between retries to acquire the lock.
+INTERVAL = 0.1
+
+
+#: A dictionary holding separate locks for each classname to be guarded
+_LOCKS = {}
```
charludo, line 20: Are you sure this works as intended? AFAIK variables are worker-local, right? So two requests could conceivably make changes to the same model's tree. (Assuming our workers use processes... If they use threads, it should be fine, since you obtain a lock below)

PeterNerlich: Yes, the locks are only to guard against a treebeard function **getting called during an ongoing transaction** in the same thread, which would also obtain the database cursor with the opened transaction, despite not being logically part of that transaction at all. I think the only situation where this could occur is when running stuff with treebeard in an asynchronous event loop.

**For separate threads this is not a problem**, since not only the locks are not shared, but there are also separate data structures representing the (patched) treebeard module. Then, that thread would get blocked when talking to the database until the ongoing transaction finishes.

PeterNerlich:
> Have you considered making kind of a "middleware" between trebeard and integreat? Something like importing treebeard, applying the mutex, and then ensuring that we only import from that middleware, never from treebeard itself? 

I'm not sure, it might work. I have the lingering feeling that there might have been complications getting the correct current db cursor, but this PR and my mental model of what was going on at every version have both been through many revisions, so it might be worth investigating again

integreat_cms/cms/utils/tree_mutex.py
charludo: > Yes, the locks are only to guard against a treebeard function **getting called during an ongoing transaction** in the same thread, which would also obtain the database cursor with the opened transaction, despite not being logically part of that transaction at all.

Makes sense!

> For separate threads this is not a problem, since not only the locks are not shared, but there are also separate data structures representing the (patched) treebeard module. Then, that thread would get blocked when talking to the database until the ongoing transaction finishes.

That is the part I was/am uncertain about. Multiple threads belonging to the same process share the lock, and the lock works as intended. But two workers work on two sets of locks, and locking one does not lock the other. So I *think* we could still run into a situation were two coinciding (web) requests cause conflicting DB operations - but tbh I am not sure. You have spend a lot more time on this, so if you believe this can not happen, I'll happily believe it too 😁

charludo:
>  have the lingering feeling that there might have been complications getting the correct current db cursor

Makes sense!

> so it might be worth investigating again

Eh, I was mostly just curious. IMO no need to rework this, since it seems to work just as intended!

charludo (approved):
I do still have the one concern stated above, but honestly, it's probably more of a theoretical excercise than a practical issue, so I think this is good as-is!

Just one request, could you add a module docstring to the top of `abstract_tree_node.py` explaining the tree mutex, and how and when it is supposed to be used instead of `transaction.atomic`? IMO it would be a good idea to have this documented somewhere, since right now it's just "institutional knowledge" 😄

integreat_cms/cms/utils/tree_mutex.py
PeterNerlich: No, because we forced treebeard into using transactions, and made sure with the lock that only the intended treebeard call gets to do stuff in the name of the transaction. So when treebeard from a different thread attempts to access the database, the database will block it until the ongoing transaction is done.
Yes, this rabbit hole is so long that at the end you don't trust DB transactions anymore. Yes, this is insane. Thank your for still following through.

integreat_cms/cms/views/bulk_action_views.py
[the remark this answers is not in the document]
PeterNerlich: We could move the contents to a separate menthod `do_post()`, create another method `do_post_for_pages()` which just calls `do_post()` but itself is decorated with `@tree_mutex("page")`, and decide which of the two methods to call in `post()` based on introspection of the type of objects for which it is being called. It seems convoluted and not very much readable to me though, so I thought using this even for poi and events when we don't need it is fine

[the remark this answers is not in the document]
jarlhengstmengel: If it doesn't harm, I think w could keep it as it is. Does it mean then that transactions on the page tree would be locked while there wouldn't be any transactions to do on the page tree?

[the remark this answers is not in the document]
PeterNerlich: Yes, though I'm not sure whether django/the database is optimising anything here. But safe to assume so

integreat_cms/cms/models/abstract_tree_node.py
```diff
@@ -272,41 +277,29 @@ def move(self, target: AbstractTreeNode, pos: str | None = None) -> None:
...
+            ]
+        ):
+            raise InvalidPosition(
+                _('The node "{}" in region "{}" cannot be moved to "{}".').format(
+                    self, self.region, target.region
+                )
+            )
+        # Moving a node can modify all other nodes via raw sql queries (which are not recognized by cachalot),
+        # so we have to invalidate the whole model manually.
+        invalidate_model(self.__class__)
```
jarlhengstmengel, line 301: Isn't this one invalidate more then needed since the `try` block isn't there anymore?

jarlhengstmengel (approved):
Thank you very much for the work and the explanations! Looks good to me!

integreat_cms/cms/models/abstract_tree_node.py
PeterNerlich: No, that refers to the django cache assuming that if we modify one object from the database, the change only affects that one object and other cached objects are still valid. However, when we move a page, in the worst case all other pages in the same tree have different `rgt` and `lft` values, so django should discard them and lod them fresh from the database next time they are accessed instead of working with the outdated cache.

PeterNerlich: I'm not sure, so Ithink I'll play it safe and leave it in

Issue #3053, opened 2024-09-14
Data export for traffic light system gets `504 Gateway Timeout` error


david-venhoff:
### Describe the Bug

This issue lays groundwork for #3012.
Add export functionality that is only available to the service team, which allows creating csv files that contain all relevant data used for the traffic light system.

That includes:
* Number of outdated pages per region
* Number of broken links per region
* Number of pages per region with hix below 15
* Number of total pages per region
* Number of missing / outdated pages per region
* Whether the extra 1 mio packages has been booked
* Number of languages besides German

david-venhoff:
@osmers 
I have two questions:
- Regarding `Number of total pages per region`: Is this about pages or page translations?
- Regarding `Number of missing / outdated pages per region`: Is this the total number of missing/outdated page translations or the total number of pages which have a missing / outdated page translation?

osmers:
Actually, if you can provide me with the information how many german pages have at least one translation missing and or outdated, we could eliminate at least two data points:

Number of total pages per region
Number of missing / outdated pages per region

Possibly even this one, even though it's interesting:
Number of languages besides German

Bcs the OKR I wanted to check with this is this one: 
Am Stichtag 31.12.2024 sind durchschnittlich die Übersetzungen von max. 50 Seiten pro Kommune nicht aktuell.

Also, can we adjust the HIX value to 14 - so the data should show how many pages below 14 exist within a region.

david-venhoff:
> Also, can we adjust the HIX value to 14 

Is 14 the current minimum required for using machine translations? Because I am currently just using that value (Since we already calculate it in the cms)
Or do you prefer to use an actually fixed value?

osmers:
At the conference we decided to keep the minimum for machine translations at 14 - so even after 01.01.2025 that value will be the relevant one - and yes, then it makes sense to just stick to the minimum value for machine translations :) no need for a fixed value, I think

JoeyStk:
Trying to use this feature on our test CMS leads to a `504 Gateway Timeout error`. I fear that this issue is not resolved yet :/

david-venhoff:
Then I think we should merge #3120 and see if that fixes the problem

david-venhoff:
Actually, I think it is fixed now (@svenseeberg  increased the timeout to 30 minutes)

The export took almost 25 minutes though :see_no_evil: 
So we should definitely try to improve the performance

JoeyStk:
You're right! Now it works :)

[...]

Issue #3012, opened 2024-08-21
Traffic light system for the status of regions


osmers:
### Motivation
Inspired by the Excel sheet that we keep, to determine how well the municipalities are doing, we want to incorporate the same metric into the network management of the CMS. This serves a dual purpose.
1. We have data that is always up to date bcs it draws the necessary input directly from the system
2. We avoid about 10 hours of work every quarter to update the list manually

### Proposed Solution
Take the columns and calculations from the excel sheet and design the same/a similar table in the network management that is fed by the data from each region. 

### Alternatives
Continue working with excel

### User Story
As a service team member, I want to have up to date data on how the municipalities are doing in order to determine who needs support and in order to see while on the phone with a municipality, how they are doing.

### Additional Context
We would need the following numbers from each region
From the to do board: 
- number of outdated pages
- number of broken links
- number of pages with HIX below 15

From the page tree:
- number of pages in total in the region

From the Translation report:
- number of missing/outdated translations (if possible it would be great to add the number of pages from the diagram, but alternatively the number of words in the table works too (but then we need some more information like number of words in the region)

From the region settings: 
- yes or no whether the extra package for 1 Mio words is activated

From language tree:
- number of languages besides German/root language

![image](https://github.com/user-attachments/assets/d0acefc6-286f-4c9c-8feb-a87a8240da6c)

### Design Requirements

osmers:
Here is the list currently being used: https://nextcloud.tuerantuer.org/index.php/f/6502587
Be careful with the calculations and everything so nothing breaks :)

david-venhoff:
Blocked on 3053

timobrembeck:
Is this a duplicate of #78 (or are they at least partially overlapping)? Or are they just linked to each other?

david-venhoff:
I think they just unfortunately share the same name, but are about entirely different things :sweat_smile: 
The traffic light system mentioned here is about how well we think the municipalities are doing, while the traffic light system of #78 is about storage usage

JoeyStk:
@osmers 
The first iteration of the table is online on cms-test now :)
However it's currently only possible to navigate by adding "?page=2" or by increasing the number of pages shown per view by adding "?size=50" to the URL. 🙈 Could you please check it out and list down a few things you wish we would improve for the next iteration? :)
[...]

Issue #3077, opened 2024-09-26
Consider using indexes to improve database performance


david-venhoff:
### Motivation

According to the django documentation, adding indexes [should be priority number one to improve performance](https://docs.djangoproject.com/en/5.1/topics/db/optimization/#use-standard-db-optimization-techniques). However, after doing a quick code search I noticed that we did not set up a single index for any model.

### Proposed Solution

Analyze and benchmark existing slow queries and add indexes where appropriate

### Alternatives

Do nothing. Though if we want to improve the performance of the cms, we should really investigate this with the highest priority

PeterNerlich:
Interpretation from issue refinement: *Question* is to be answered by working on this issue, not a precondition to starting to work on it. Removing the label.

The following is not a question blocking work on this either, but could be helpful for where to direct efforts first for maximum impact.
@laravoelk Do you know any specific views or actions that are the primary pain points for users from a performance standpoint?

steffenkleinle:
> Besides persons working with our system, the load time for data we send to clients via our API end point is also of interest, but I suspect the [@digitalfabrik/app-team](https://github.com/orgs/digitalfabrik/teams/app-team) will know the problematic endpoints there better.

From an app team perspective, I'd definitely say that the slowest are the categories endpoints (so children and/or parent) endpoints. The others are reasonable fast imo but faster is always better here I guess :D

PeterNerlich:
maybe `get_absolute_url()` (especially for pages) would be a good research specimen

PR #3079, opened 2024-09-27
Implement region condition data export

david-venhoff:
### Short description

Adds a new admin page with a button to download a region condition export.
The page itself is very minimal and will be improved with #3012.

Speed is not great, unfortunately. An export using production data takes about 6 minutes.

### Proposed changes

- Add a new view
- Add a new endpoint to create a data export
- Move some methods from the dashboard to the region, so that they can be shared by the dashboard and the region condition export

### Side effects

None

### Resolved issues

Fixes: #3053

### DIFF
--- a/integreat_cms/cms/views/region_condition/region_condition_actions.py
+++ b/integreat_cms/cms/views/region_condition/region_condition_actions.py
@@ -5,6 +5,7 @@ from datetime import datetime
 from typing import TYPE_CHECKING
 
 import magic
+from django.contrib.admin.views.decorators import staff_member_required
 from django.db.models import Q
 from django.http import HttpResponse
 from django.utils.translation import gettext_lazy as _
@@ -13,7 +14,6 @@ from import_export import fields, resources
 from tablib.formats import registry as format_registry
 
 from ...constants import region_status, translation_status
-from ...decorators import permission_required
 from ...models import Page, Region
 from ...utils.linkcheck_utils import filter_urls
 from ..utils.hix import get_translation_under_hix_threshold
@@ -144,7 +144,7 @@ class RegionConditionResource(resources.ModelResource):
 
 
 @require_POST
-@permission_required("cms.view_region")
+@staff_member_required
 def export_region_conditions(request: HttpRequest, file_format: str) -> HttpResponse:
     """
     Creates a data export summarizing the condition of all region
--- a/integreat_cms/cms/views/region_condition/region_condition_actions.py
+++ b/integreat_cms/cms/views/region_condition/region_condition_actions.py
@@ -6,17 +6,17 @@ from typing import TYPE_CHECKING
 
 import magic
 from django.db.models import Q
+from django.http import HttpResponse
+from django.utils.translation import gettext_lazy as _
+from django.views.decorators.http import require_POST
 from import_export import fields, resources
 from tablib.formats import registry as format_registry
-from django.utils.translation import gettext_lazy as _
-from django.http import HttpResponse
-from django.views.decorators.http import require_POST
 
-from ..utils.hix import get_translation_under_hix_threshold
-from ...constants import translation_status, region_status
+from ...constants import region_status, translation_status
 from ...decorators import permission_required
-from ...models import Region, Page
+from ...models import Page, Region
 from ...utils.linkcheck_utils import filter_urls
+from ..utils.hix import get_translation_under_hix_threshold
 
 if TYPE_CHECKING:
     from django.http import HttpRequest
--- a/integreat_cms/cms/views/region_condition/region_condition_actions.py
+++ b/integreat_cms/cms/views/region_condition/region_condition_actions.py
@@ -1,0 +1,146 @@
+from __future__ import annotations
+
+import logging
+from datetime import datetime
+from typing import TYPE_CHECKING
+
+import magic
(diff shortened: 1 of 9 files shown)


integreat_cms/cms/views/region_condition/region_condition_actions.py
```diff
@@ -0,0 +1,173 @@
...
+        model = Region
+        # if we don't define the empty fields all fields are created in the default way additionally to our custom way
+        fields = ()
+
+
+@require_POST
+@permission_required("cms.view_region")
+def export_region_conditions(request: HttpRequest, file_format: str) -> HttpResponse:
+    """
+    Creates a data export summarizing the condition of all region
```
JoeyStk, line 150:
```suggestion
    Creates a data export summarizing the condition of all regions
```

MizukiTemma:
@david-venhoff Thank you for the PR, looks very good 😸 

There is one question: shouldn't the feature only be available for Service Team (+ CMS)? it doesn't harm that App Team and Marketing Team can use this functionality, though it was so specified in the issue 🤔

PeterNerlich:
Had a quick look and for the stock dev test data it takes 43 queries (6 regions). I am confident that this can be significantly reduced, best case it could be one single humongous db query that we just dump as is into the desired format. I think a reduction to one quarter of the time might be realistic.

integreat_cms/cms/views/region_condition/region_condition_actions.py
```diff
@@ -0,0 +1,173 @@
...
+logger = logging.getLogger(__name__)
+
+
+class RegionConditionResource(resources.ModelResource):
+    """
+    A resources class that connects to the django-import-export library.
+    It represents the to-be exported status of all regions.
+    """
+
+    name = fields.Field(column_name=_("Region name"), attribute="name")
```
PeterNerlich, line 35: What do you think, does it make more sense to just say "Region" in this column?
```suggestion
    name = fields.Field(column_name=_("Region"), attribute="name")
```

integreat_cms/cms/views/region_condition/region_condition_actions.py
```diff
@@ -0,0 +1,173 @@
...
+    num_pages_with_missing_or_outdated_translation = fields.Field(
+        column_name=_(
+            "Number of pages with at least one missing or outdated translation"
+        )
+    )
+
+    num_outdated_pages = fields.Field(column_name=_("Number of outdated pages"))
+
+    has_translation_package_been_booked = fields.Field(
+        column_name=_("Has translation package been booked"),
```
PeterNerlich, line 50: I think adding a question mark makes it slightly better at conveying the meaning
```suggestion
    has_translation_package_been_booked = fields.Field(
        column_name=_("Has translation package been booked?"),
```

PeterNerlich (approved):
Well, this was a disappointing investigation of mine :grin: Take my approve and a few over-perfectionist nitpicks and ideas I'm not even sure of myself, since I cannot contribute anything more helpful :stuck_out_tongue:

david-venhoff:
@PeterNerlich Thanks for the review :)

>  I think a reduction to one quarter of the time might be realistic.

I share this sentiment, and I went on to exactly the same investigation as you :smiling_face_with_tear: 
I think that by far the largest bottleneck is in mapping `Link`s to `Region`s and I ~~wasted~~ spent quite a bit of time trying to improve its performance, without much success. For example, it is possible to completely get rid of the `filter_urls` function and perform the calculations completely in the database. However my experiment created such an awful query that it crashed postgres ^^
My current hope is that we can create some indexes in the right places to speed this up, and I created #3077 for it, by I did not start investigating this yet.

JoeyStk (changes requested):
Thank you very much! This looks quite good already. However I noticed a few things when reviewing this with our test data.

**Number of missing or outdated translations**: In our test data I have eight pages that have a missing or outdated translation, the CSV says seven. I have the same error in Augsburg where I had 25 outdated translations for not empty pages, the CSV only says 19.  Maybe it's worth to have a look at [this branch](https://github.com/digitalfabrik/integreat-cms/tree/feature/add-translation-coverage-row-to-dashboard), where attempted at implementing something similar :) 

The rest seemed to be accurate. I think however that thorough testing after merging would be a great benefit :)

integreat_cms/cms/views/region_condition/region_condition_actions.py
david-venhoff: :+1:

david-venhoff:
@MizukiTemma 

> There is one question: shouldn't the feature only be available for Service Team (+ CMS)?

I agree, and it is implemented now

JoeyStk:
> Could you prune the database and try again? In my csv the number is 8 (I assume you refer to nürnberg)

Ok, this is wild. At one point I also had the correct number of 8, but after executing `prune_database.sh` it's now back to 7, but I count 8 missing translations and 5 pages with at least one missing translation (for Nürnberg). I'm not sure what's going on here... For Augsburg the numbers are similarly off (see other screenshot) :thinking: 

Screenshot from 2024-10-08 11-28-14

Screenshot from 2024-10-08 11-20-29

Screenshot from 2024-10-08 11-25-27

JoeyStk (approved):
Sorry, I was just dumb

Issue #2954, opened 2024-08-01
[2024-11-07] Add bulk action to contact list


MizukiTemma:
### Motivation

After #2952 we have a model and list view for contacts of the region. So that users can archive/restore/delete contacts, we need functions and bulk actions.

### Proposed Solution

- Implement a function to archive/restore/delete a contact object
- Connect those functions to the icons in each row of the list
- Add bulk action for those three functionalities

### Alternatives

None. We do it.

### Additional Context

Wait for #2952

Issue #2950, opened 2024-08-01
Add bulk action to organization list


MizukiTemma:
### Motivation

There is no bulk action in the organization list.

### Proposed Solution

Add a bulk action to archive/restore/delete selected organizations.

### Alternatives

Leave it as it is.

### Additional Context

Probably first implement archiving functionality and then this issue? See #2949

PR #3080, opened 2024-09-30
[2024-11-07] Add contact list bulk actions

jarlhengstmengel:
### Short description

Adds archive/restore/delete bulk actions to contact list

### Proposed changes

- Adds archive/restore/delete bulk actions `contact_bulk_action.py`
- Overwrites `get_queryset()` to filter for selected contacts with poi region

### Side effects

- None

### Resolved issues

Fixes: #2954 

### DIFF
--- a/integreat_cms/cms/templates/contacts/contact_list.html
+++ b/integreat_cms/cms/templates/contacts/contact_list.html
@@ -95,6 +95,20 @@
                 <option>
                     {% translate "Select bulk action" %}
                 </option>
+                {% if is_archive %}
+                    <option data-bulk-action="{% url "bulk_restore_contacts" region_slug=request.region.slug %}">
+                        {% translate "Restore contacts" %}
+                    </option>
+                {% else %}
+                    <option data-bulk-action="{% url "bulk_archive_contacts" region_slug=request.region.slug %}">
+                        {% translate "Archive contacts" %}
+                    </option>
+                {% endif %}
+                {% if perms.cms.delete_contact %}
+                    <option data-bulk-action="{% url "bulk_delete_contacts" region_slug=request.region.slug %}">
+                        {% translate "Delete contacts" %}
+                    </option>
+                {% endif %}
             </select>
             <button form="bulk-action-form" id="bulk-action-execute" class="btn" disabled>
                 {% translate "Execute" %}
--- a/integreat_cms/cms/views/bulk_action_views.py
+++ b/integreat_cms/cms/views/bulk_action_views.py
@@ -11,6 +11,7 @@ from cacheops import invalidate_model
 from django.contrib import messages
 from django.contrib.auth.mixins import PermissionRequiredMixin
 from django.core.exceptions import PermissionDenied
+from django.db.models import Q
 from django.http import Http404
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
@@ -90,9 +91,9 @@ class BulkActionView(PermissionRequiredMixin, MultipleObjectMixin, RedirectView)
             super()
             .get_queryset()
             .filter(
-                region=self.request.region,
                 id__in=self.request.POST.getlist("selected_ids[]"),
             )
+            .filter(self.get_extra_filters())
         )
         if not queryset:
             raise Http404(f"No {self.model._meta.object_name} matches the given query.")
@@ -101,6 +102,14 @@ class BulkActionView(PermissionRequiredMixin, MultipleObjectMixin, RedirectView)
         if self.prefetch_public_translations:
             queryset = queryset.prefetch_public_translations()
         return queryset
+
+    def get_extra_filters(self) -> Q:
+        """
+        Get extra filters for get_queryset(). Overwrite this method to specify needed filters e.g. as in `~integreat_cms.cms.views.contact.contact_bulk_actions`
+
+        :return: Q Object to be used as filter argument
+        """
+        return Q(region=self.request.region)
 
 
 class BulkMachineTranslationView(BulkActionView):
--- a/integreat_cms/cms/views/contacts/contact_bulk_actions.py
+++ b/integreat_cms/cms/views/contacts/contact_bulk_actions.py
@@ -1,0 +1,152 @@
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING
+
+from django.contrib import messages
+from django.db.models import Q
+from django.utils.translation import ngettext_lazy
+
+from ...models import Contact
+from ...utils.stringify_list import iter_to_string
+from ..bulk_action_views import BulkActionView
+
+if TYPE_CHECKING:
+    from typing import Any
+
+    from django.http import HttpRequest, HttpResponse
+
+logger = logging.getLogger(__name__)
+
+
+class ContactBulkAction(BulkActionView):
+    """
+    View for executing contact bulk actions
+    """
+
+    #: The model of this :class:`~integreat_cms.cms.views.bulk_action_views.BulkActionView`
+    model = Contact
+
+    def get_extra_filters(self) -> Q:
+        """
+        Overwrite to filter queryset for region of the location since contact gets its region from location
+        """
+        return Q(location__region=self.request.region)
+
+
+class ArchiveContactBulkAction(ContactBulkAction):
+    """
+    Bulk action to archive multiple contacts at once
+    """
+
+    def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
+        r"""
+        Archive multiple contacts at once
+
+        :param request: The current request
+        :param \*args: The supplied arguments
+        :param \**kwargs: The supplied keyword arguments
+        :return: The redirect
+        """
+
+        archive_successful = []
+
+        for content_object in self.get_queryset():
+            content_object.archive()
+            archive_successful.append(content_object)
+
+        if archive_successful:
+            messages.success(
+                request,
+                ngettext_lazy(
+                    "{model_name} {object_names} was successfully archived.",
+                    "The following {model_name_plural} were successfully archived: {object_names}",
+                    len(archive_successful),
+                ).format(
+                    model_name=self.model._meta.verbose_name.title(),
+                    model_name_plural=self.model._meta.verbose_name_plural,
+                    object_names=iter_to_string(archive_successful),
+                ),
+            )
+
+        return super().post(request, *args, **kwargs)
+
+
+class RestoreContactBulkAction(ContactBulkAction):
+    """
+    Bulk action to restore multiple contacts at once
+    """
+
+    def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
+        r"""
+        Function to restore multiple contacts at once
+
+        :param request: The current request
+        :param \*args: The supplied arguments
+        :param \**kwargs: The supplied keyword arguments
+        :return: The redirect
+        """
+        restore_sucessful = []
+        for content_object in self.get_queryset():
+            content_object.restore()
+            restore_sucessful.append(content_object)
+
+        if restore_sucessful:
+            messages.success(
+                request,
+                ngettext_lazy(
+                    "{model_name} {object_names} was successfully restored.",
+                    "The following {model_name_plural} were successfully restored: {object_names}",
+                    len(restore_sucessful),
+                ).format(
+                    model_name=self.model._meta.verbose_name.title(),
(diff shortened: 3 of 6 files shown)


integreat_cms/release_notes/current/unreleased/2954.yml
david-venhoff, line 1: In general, we have not added release notes for contacts yet. I think the plan is to only add one general relase note like `add contacts`, when we are done with the implementation instead of many small release notes.
For this reason, could you delete this file?

integreat_cms/cms/templates/contacts/contact_list.html
```diff
@@ -89,6 +89,20 @@ <h1 class="heading">
...
                 </option>
+                {% if is_archive %}
+                    <option data-bulk-action="{% url "bulk_restore_contacts" region_slug=request.region.slug %}">
+                        {% translate "Restore contacts" %}
+                    </option>
+                {% else %}
+                    <option data-bulk-action="{% url "bulk_archive_contacts" region_slug=request.region.slug %}">
+                        {% translate "Archive contacts" %}
+                    </option>
+                {% endif %}
```
david-venhoff, line 100:
```suggestion
    {% if perms.cms.change_contact %}
                {% if is_archive %}
                    <option data-bulk-action="{% url "bulk_restore_contacts" region_slug=request.region.slug %}">
                        {% translate "Restore contacts" %}
                    </option>
                {% else %}
                    <option data-bulk-action="{% url "bulk_archive_contacts" region_slug=request.region.slug %}">
                        {% translate "Archive contacts" %}
                    </option>
                {% endif %}
            {% endif %}
```

These bulk actions should only be available to users who have the permission to modify contacts

integreat_cms/cms/views/contacts/contact_bulk_actions.py
```diff
@@ -0,0 +1,176 @@
...
+from ...models import Contact
+from ...utils.stringify_list import iter_to_string
+from ...utils.translation_utils import gettext_many_lazy as __
+from ..bulk_action_views import BulkActionView
+
+if TYPE_CHECKING:
+    from typing import Any
+
+    from django.http import HttpRequest, HttpResponse
+    from django.http.response import HttpResponseRedirect
```
david-venhoff, line 21:
```suggestion
```

integreat_cms/cms/views/contacts/contact_bulk_actions.py
```diff
@@ -0,0 +1,176 @@
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING
+
+from django.contrib import messages
+from django.http import Http404
+from django.utils.translation import gettext_lazy as _
```
david-venhoff, line 8:
```suggestion
```

integreat_cms/cms/views/contacts/contact_bulk_actions.py
```diff
@@ -0,0 +1,176 @@
...
+from django.utils.translation import ngettext_lazy
+from django.views.generic.list import MultipleObjectMixin
+
+from ...models import Contact
+from ...utils.stringify_list import iter_to_string
+from ...utils.translation_utils import gettext_many_lazy as __
```
david-venhoff, line 14:
```suggestion
```

integreat_cms/cms/views/contacts/contact_bulk_actions.py
```diff
@@ -0,0 +1,176 @@
...
+            super(BulkActionView, self)
+            .get_queryset()
+            .filter(
+                poi__region=self.request.region,
+                id__in=self.request.POST.getlist("selected_ids[]"),
+            )
+        )
+        if not queryset:
+            raise Http404(f"No {self.model._meta.object_name} matches the given query.")
+        return queryset
```
david-venhoff, line 51: IMO a super call with parameters is uncommon and can be a bit confusing. What do you think about adding a new function like `def get_extra_filters(self) -> Q` to `BulkActionView`?
```python
class BulkActionView(...):
    ...

    def get_queryset(...):
        ...
        queryset = (
            super()
                .get_queryset()
                .filter(
                    id__in=self.request.POST.getlist("selected_ids[]"),
                )
                .filter(self.get_extra_filters())
        )
        ...

    def get_extra_filters(self) -> Q:
       return Q(region=self.request.region)
```

Then you would just have to overwrite this method in `ContactBulkAction` and would also reduce code duplication a bit

integreat_cms/cms/views/contacts/contact_bulk_actions.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+class ContactBulkActionMixin(MultipleObjectMixin):
+    """
+    Mixin for contact bulk actions
+    """
+
+    #: The model of this :class:`~integreat_cms.cms.views.bulk_action_views.BulkActionView`
+    model = Contact
```
david-venhoff, line 60:
```suggestion
```
I don't think this class is needed. `ContactBulkAction` already inherits from `MultipleObjectMixin` (via `BulkActionView`), so this code can simply be moved to `ContactBulkAction`

integreat_cms/cms/views/contacts/contact_bulk_actions.py
```diff
@@ -0,0 +1,176 @@
...
+    from django.http.response import HttpResponseRedirect
+
+logger = logging.getLogger(__name__)
+
+
+class ContactBulkAction(BulkActionView):
+    """
+    Class to overwrite get_queryset()
+    """
+
```
david-venhoff, line 30:
```suggestion

    #: The model of this :class:`~integreat_cms.cms.views.bulk_action_views.BulkActionView`
    model = Contact

```

integreat_cms/cms/views/contacts/contact_bulk_actions.py
```diff
@@ -0,0 +1,176 @@
...
+    Mixin for contact bulk actions
+    """
+
+    #: The model of this :class:`~integreat_cms.cms.views.bulk_action_views.BulkActionView`
+    model = Contact
+
+
+class ArchiveContactBulkAction(
+    ContactBulkActionMixin, ContactBulkAction
+):  # pylint: disable=too-many-ancestors
```
david-venhoff, line 65:
```suggestion
class ArchiveContactBulkAction(ContactBulkAction):
```

integreat_cms/cms/views/contacts/contact_bulk_actions.py
```diff
@@ -0,0 +1,176 @@
...
+                    object_names=iter_to_string(archive_successful),
+                ),
+            )
+
+        return super().post(request, *args, **kwargs)
+
+
+class RestoreContactBulkAction(
+    ContactBulkActionMixin, ContactBulkAction
+):  # pylint: disable=too-many-ancestors
```
david-venhoff, line 105:
```suggestion
class RestoreContactBulkAction(ContactBulkAction):
```

integreat_cms/cms/views/contacts/contact_bulk_actions.py
```diff
@@ -0,0 +1,176 @@
...
+                    object_names=iter_to_string(restore_sucessful),
+                ),
+            )
+
+        return super().post(request, *args, **kwargs)
+
+
+class DeleteContactBulkAction(
+    ContactBulkActionMixin, ContactBulkAction
+):  # pylint: disable=too-many-ancestors
```
david-venhoff, line 143:
```suggestion
class DeleteContactBulkAction(ContactBulkAction):
```

integreat_cms/cms/views/contacts/contact_bulk_actions.py
```diff
@@ -0,0 +1,176 @@
...
+                    model_name=self.model._meta.verbose_name.title(),
+                    model_name_plural=self.model._meta.verbose_name_plural,
+                    object_names=iter_to_string(restore_sucessful),
+                ),
+            )
+
+        return super().post(request, *args, **kwargs)
+
+
+class DeleteContactBulkAction(
```
david-venhoff, line 141: I know it was part of the issue, but this would be our first delete bulk action afaik. I think if we include the delete action here, we should also consider implementing it for the other bulk actions. @osmers Do you have an opinion on that?

integreat_cms/cms/views/contacts/contact_bulk_actions.py
```diff
@@ -0,0 +1,176 @@
...
+    ContactBulkActionMixin, ContactBulkAction
+):  # pylint: disable=too-many-ancestors
+    """
+    Bulk action to delete multiple contacts at once
+    """
+
```
david-venhoff, line 147: This action would only check for the `change_permission` (See [bulk_action_views](https://github.com/digitalfabrik/integreat-cms/blob/079fad8fa712411ffb0c04085d21bbf9c5b04aa8/integreat_cms/cms/views/bulk_action_views.py#L59)), so this action will have to overwrite the `get_permission_required` method to use a delete permission.

david-venhoff (changes requested):
Thanks!

I have a few suggestions

integreat_cms/cms/templates/contacts/contact_list.html
jarlhengstmengel: The check is already there in Line 86

integreat_cms/cms/views/contacts/contact_bulk_actions.py
jarlhengstmengel: Thanks for the suggestion. I think I will try it as you suggest. I agree that it's probably less confusing.

integreat_cms/cms/views/contacts/contact_bulk_actions.py
jarlhengstmengel: I think so too. It's at least not part of `bulk_actions_view.py`. It's also part of issue #2950

integreat_cms/cms/views/contacts/contact_bulk_actions.py
jarlhengstmengel: Thanks for the hint! Untill now I was quite unsure how the permissions work and i think I realized it now

integreat_cms/cms/views/bulk_action_views.py
```diff
@@ -102,6 +103,14 @@ def get_queryset(self) -> Any:
             queryset = queryset.prefetch_public_translations()
         return queryset
 
+    def get_extra_filters(self) -> Q:
+        """
+        Get extra filters for get_queryset(). Overwrite this method to specify needed filters
+
+        :return: Q Object to be used as filter argument
+        """
+        return Q(region=self.request.region)
```
charludo, line 112: Could you add a comment on why this is necessary?

Tbh I am at a bit of a loss here. I tried it out and indeed, when directly applying `.filter(Q(...))` it won't work. Why? 😕 😂

charludo (approved):
Thank you, works great! 🎉

integreat_cms/cms/views/bulk_action_views.py
jarlhengstmengel: `get_extra_filters` is needed to be able to overwrite the filter for `region` because `contact` gets its region from `location` due to the relationship between those two entities. So in `contact_bulk_actions.py` the method is overwritten to filter for `location__region`. If you apply `region=self.request.region` directly, the contact bulk actions won't work because `region` does not exist in the `contact`model. Returning a Q object is needed to be able to use it as argument for `.filter()`.

charludo: Ohhhhh OK. That makes sense. THakns for the explanmation :D
It's not really intuitive though just from the code (at least IMO).

I think it would be helpful to either explain that `get_extra_filters` gets overriden by some views in its docstring, or to overwrite `get_queryset` instead in the `ContactBulkAction`.

@david-venhoff also feel free to weigh in on this if I am telling nonsense here ^^

david-venhoff: I suggested to create this method because overwriting `get_queryset` (as was done originally in this pr) was a bit hard to understand imo and lead to a bit of code duplication, but more documentation can't hurt :D

david-venhoff (approved):
LGTM :)

integreat_cms/cms/views/bulk_action_views.py
jarlhengstmengel: I added a reference to `contact_bulk_action` in the docstring and some explanation in the docstring in the override in `contact_bulk_action`. Hope that it makes it better to understand.

jarlhengstmengel: Thanks for your suggestions :)

Issue #2545, opened 2023-11-11
Add typing information to documentation


timobrembeck:
### Motivation

#2459 added type hints for the entire code base and enforces annotations via the CI pipeline.
However, not all current and future team members might be familiar with the syntax and caveats.

### Proposed Solution

Add a section to the documentation with explaining the most important things and links to more detailed resources
### Additional Context

https://chat.tuerantuer.org/digitalfabrik/pl/re6oi95cofdezx4s1rwrbjggie

PR #3081, opened 2024-09-30
Add documentation for mypy

MizukiTemma:
### Short description

This PR contains information about mypy and our tool `./tools/mypy.sh`

### Proposed changes

- Add explanation for mypy in `code_style.rst` and `tools.rst`
- Change the order of tool listing `tools.rst` to align with that in `./tools/code_style.sh`

### Side effects

- Easier onboading 😸 

### Resolved issues

Fixes: #2545 


jarlhengstmengel (approved):
Thanks for the work! Looks good to me :)

PeterNerlich (changes requested):
There are a few quirks where restructuredText looks similar to markdown at first, but where this first impression then causes long term confusion because you use markdown conventions from muscle memory and then fail to see what's wrong when it doesn't end up looking as expected.

Most prominently, to write `inline fixed width text`, rst encloses the text in *double backticks*, not single backticks. Single backticks instead render to italic text.

Then I added some context about what `__future__.annotations` actually does, since we won't require it forever, and tried to write it as complete but concise as possible. Feel free to reword though if you think of something better

PeterNerlich (approved):

PR #3082, opened 2024-09-30
Update dependencies

david-venhoff:
### Short description

This pr updates our dependencies, which includes multiple breaking changes that I attempted to fix.

### Proposed changes

- Fix typescript typechecking which for some reason did not appear to be active for preact before?
- Replace deprecated sass functions by their non-deprecated counterparts
- Update the sass-loader dependency to avoid another deprecation warning
- Handle the new pylint-lint by setting the maximal amount of positional arguments to 6 (default 5) and annotating the remaining exceptions

### Side effects

I also noticed 2 direct python dependencies that had a major version bump. I did some testing and don't think any functionality of the cms was broken.

- Cryptography (https://github.com/pyca/cryptography/blob/main/CHANGELOG.rst#4300---2024-07-20): 
  * Dropped support for old OpenSSL and LibreSSL versions (@ulliholtgrave could that effect us?)
- Icalendar (https://icalendar.readthedocs.io/en/latest/changelog.html#a0-2024-07-03)
  * Now uses zoneinfo instead of pytz for timezones

Running `./tools/eslint` now produces a deprecation warning. I tried updating eslint, but we depend on `eslint-config-airbnb` which does not support modern eslint versions. At least there is an active issue for that: https://github.com/airbnb/javascript/issues/2961
I think we can accept this deprecation warning for the time being and open an issue to update eslint, but I am open to other opinions :)

### Resolved issues

Fixes: /


ulliholtgrave:
@david-venhoff Thank you very much! 😊 I don't think that the changes to cryptography are affecting us.
I'll do some more functional testing and probably will approve this in the next days :)

charludo (approved):
Wow, this looks like quite the effort, so thank you for tackling this!!
Couldn't find anything not working! 🎉

Issue #3098, opened 2024-10-02
No region without Zammad-URL can be created


MizukiTemma:
### Describe the Bug

Creating a new region without Zammad-URL fails.

### Steps to Reproduce

1. Go to "Regions" in the netowork management.
2. Create a new region with the input field "Zammad-URL" empty
3. See the error `NoneType object has no attribute split`

### Expected Behavior

Region must be successfully created even without Zammad-URL.
(Or show an user friendly message if a Zammad-URL is mandatory)

### Actual Behavior

Error

### Additional Information

  File "/home/micucu/integreat-cms/integreat_cms/cms/forms/regions/region_form.py", line 605, in clean_zammad_url
    cleaned_zammad_url = cleaned_zammad_url.split("/api/v1")[0]
<details>
  <summary>
</summary>

  ```
  ```

</details>

PR #3099, opened 2024-10-03
Fix empty Zammad field in region form

svenseeberg:
### Short description
Fix handling empty Zammad URL in region form. We allowed Null/None values for the zammad_url field in #3051 but forgot to consider this in the form sanitization.

### Proposed changes
- Check if the zammad_url field in the form is empty before doing string operations. If empty, return None (Null for database).

### Side effects
- I cannot think of any.

### Resolved issues

Fixes: #3098

### DIFF
--- a/integreat_cms/release_notes/current/unreleased/3098.yml
+++ b/integreat_cms/release_notes/current/unreleased/3098.yml
@@ -1,0 +1,2 @@
+en: Fix bug when saving a Region without a Zammad URL
+de: Fehlerkorrektur beim Speichern einer Region ohne Zammad URL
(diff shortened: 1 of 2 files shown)


MizukiTemma (approved):
Thank you for speedy fix 😸

integreat_cms/release_notes/current/unreleased/3098.yml
```diff
@@ -0,0 +1,2 @@
+en: Fix empty Zammad URL in region form
```
JoeyStk, line 1:
```suggestion
en: Fix error when creating region without Zammad URL 
```
Nit: I think this is a bit better to understand for people without a background in IT?

JoeyStk (approved):
LGTM

JoeyStk:
Ah, one thing I forgot. I think it makes sense to write a test for region creation as we encounter issues with it every once in a while. Maybe we can make this part of this PR depending on how urgent it is

Issue #3104, opened 2024-10-04
[2024-11-07] Show external calendar column in event list to all users


JoeyStk:
### Motivation

In our current implementation it's only possible for members of the CMS and service team to see which events have been imported in the event list. However this is information is also - and even more so - important to members who don't belong to these groups. Therefore it should be possible for all other user groups to see the column. 

### Proposed Solution

Show column to all staff members, authors, event manager, manager, editors and observers

### Alternatives

None others were discussed

### User Story

As a user of the CMS I want to see which events have been imported and which ones haven't.

### Additional Context

None

### Design Requirements

None

PR #3115, opened 2024-10-06
[2024-11-07] Extend user permissions for external calendar column in event list

JoeyStk:
### Short description

Allow all users to see external calendar column in event list view

### Proposed changes

- Change permission
- Add icon in case there is no external calendar

### Side effects

None I'm aware of

### Resolved issues

Fixes: #3104 

### DIFF
--- a/integreat_cms/cms/templates/events/event_list_row.html
+++ b/integreat_cms/cms/templates/events/event_list_row.html
@@ -116,13 +116,13 @@
             <i icon-name="calendar"></i> {% translate "One-time" %}
         {% endif %}
     </td>
-    {% if perms.cms.view_externalcalendar %}
-        <td class="py-3 pr-2">
-            {% if event.external_calendar %}
-                {{ event.external_calendar.name }}
-            {% endif %}
-        </td>
-    {% endif %}
+    <td class="py-3 pr-2">
+        {% if event.external_calendar %}
+            {{ event.external_calendar.name }}
+        {% else %}
+            <i icon-name="minus"></i>
+        {% endif %}
+    </td>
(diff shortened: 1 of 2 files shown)


integreat_cms/cms/templates/events/event_list_row.html
```diff
@@ -116,13 +116,13 @@
...
-                {{ event.external_calendar.name }}
-            {% endif %}
-        </td>
-    {% endif %}
+    <td class="py-3 pr-2">
+        {% if event.external_calendar %}
+            {{ event.external_calendar.name }}
+        {% else %}
+            <i icon-name="minus"></i>
+        {% endif %}
```
PeterNerlich, line 124: Is there a deliberate reasoning for showing a dash instead of just leaving it blank? Should we maybe adapt the same for empty fields in the new contacts list view?

PeterNerlich (approved):

integreat_cms/cms/templates/events/event_list_row.html
JoeyStk: It was just a spontaneous idea from me side and hasn't been discussed before. I'm willing to change it back, if it's preferred. It just had a personal preference for a dash instead of an empty field

JoeyStk: So, it was agreed that this is better than leaving it blank. We still have a discussion whether or not it's better to center it or not.
As for the other list views: In feedback we already have it, in fact, that's where I go the idea from :)

JoeyStk: It was agreed that left alignment is better than center :)

Issue #3100, opened 2024-10-04
[2024-11-07] iCal: Show information who has imported the calendar


JoeyStk:
### Describe the Bug

In our current implementation of iCal we don't show the information who imported the calendar

### Steps to Reproduce

1. Go to external calendar
2. See there is no information who imported the calendar

### Expected Behavior

I can get the information on who imported the calendar

### Actual Behavior

I don't

### Additional Information

None

PR #3116, opened 2024-10-06
[2024-11-07] Add created by field to external calendar

JoeyStk:
### Short description

This PR adds the information who created an external calendar

### Proposed changes

- Add `created_by` field to model
- Add `created_by` column to calendar list view

### Side effects

None I'm aware of

### Resolved issues

Fixes: #3100 


integreat_cms/cms/views/external_calendars/external_calendar_form_view.py
```diff
@@ -95,6 +95,7 @@ def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
                 )
         else:
             external_calendar_form.instance.region = self.request.region
+            external_calendar_form.instance.created_by = request.user
```
david-venhoff, line 98: Should this happen on every update? Or just when the external calendar is created initially?

integreat_cms/cms/templates/events/external_calendar_list.html
david-venhoff, line 1: Looks like the colspan for empty tables is outdated again

integreat_cms/cms/templates/events/external_calendar_list_row.html
```diff
@@ -30,6 +30,9 @@
             {{ calendar.events.count }}
         </a>
     </td>
+    <td>
+        {{ calendar.created_by }}
```
david-venhoff, line 34:
```suggestion
         <a href="{% url 'edit_external_calendar' region_slug=request.region.slug calendar_id=calendar.id %}" class="block px-2 py-3 text-gray-800">
            {{ calendar.created_by }}
        </a>
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -2920,6 +2920,15 @@ msgstr ""
 msgid "import errors"
 msgstr "Importfehler"
 
+#: cms/models/external_calendars/external_calendar.py
+#: cms/templates/events/external_calendar_list.html
+msgid "Created by"
+msgstr "Erstellt von"
```
david-venhoff, line 2926: should we use the same words that we use for abstract content translations (creator and Ersteller:in)?

david-venhoff (approved):
Thanks!

Only have some small suggestions / questions

integreat_cms/cms/models/external_calendars/external_calendar.py
```diff
@@ -31,6 +31,15 @@ class ExternalCalendar(AbstractBaseModel):
         ),
     )
     errors = models.CharField(verbose_name=_("import errors"), default="", blank=True)
+    created_by = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        null=True,
+        blank=True,
```
charludo, line 37: Hmm, I assume thi is because external calendars have already been created without an user?
"created_by" seems like a field that should be mandatory - do you think it woul be feasible to set a default user for the existing entries in the migration?

david-venhoff: I thought so too initially, but that would cause a problem when a user account gets removed (right now the code can simply say `SET_NULL`)

charludo: Hm, true. I guess the only way around this would be to use `SET_DEFAULT` and use the same default user as for the migration. But yeah, not sure if worth it

integreat_cms/cms/templates/events/external_calendar_list.html
david-venhoff: Yeah I think so. Unfortunately that happens quite quickly :(

JoeyStk: Ok, I changed it to 5 now, so it should be resolved :)

integreat_cms/cms/views/external_calendars/external_calendar_form_view.py
PeterNerlich: In that case, we'll for sure also want to know *when* it was last changed

integreat_cms/cms/migrations/0108_alter_externalcalendar.py
david-venhoff, line 1: Could you merge this with 0107?

integreat_cms/cms/models/external_calendars/external_calendar.py
```diff
@@ -31,6 +31,28 @@ class ExternalCalendar(AbstractBaseModel):
...
+        on_delete=models.SET_NULL,
+        related_name="created_by",
```
david-venhoff, line 39:
```suggestion
```
We probably don't want that?
(related name specifies how to follow this relationship from the user model, so in this case it would be possible to get all external calendars of a user with `user.created_by`, which seems confusing)

integreat_cms/cms/models/external_calendars/external_calendar.py
```diff
@@ -31,6 +31,28 @@ class ExternalCalendar(AbstractBaseModel):
...
+        verbose_name=_("creator"),
+        help_text=_("The account that created this external calendar."),
+    )
+    last_changed_by = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="last_changed_by",
```
david-venhoff, line 48:
```suggestion
```

integreat_cms/cms/models/external_calendars/external_calendar.py
```diff
@@ -31,6 +31,28 @@ class ExternalCalendar(AbstractBaseModel):
...
+        verbose_name=_("last changed by"),
+        help_text=_("The account that was the last to change this external calendar."),
+    )
+    last_changed_on = models.DateTimeField(
+        auto_now_add=True,
```
david-venhoff, line 53:
```suggestion
        auto_now=True,
```
Using `auto_now` will make sure that this timestamp is automatically updated on save. `auto_now_add` will only set it on creation: https://docs.djangoproject.com/en/dev/ref/models/fields/#django.db.models.DateField.auto_now_add

integreat_cms/cms/views/external_calendars/external_calendar_form_view.py
```diff
@@ -95,6 +95,9 @@ def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
                 )
         else:
             external_calendar_form.instance.region = self.request.region
+            if not external_calendar_form.instance.created_by:
+                external_calendar_form.instance.created_by = request.user
+            external_calendar_form.instance.last_changed_by = request.user
```
david-venhoff, line 100: Nit: I wonder if it would be better to make use of `additional_instance_attributes` here, and only pass the current user. To me it seems like it should be the task of the form (and not the view) to correctly set the `created_by` and `last_changed_by` fields

integreat_cms/cms/models/external_calendars/external_calendar.py
JoeyStk: I agree that this is confusing, but since both fields call to the User model and both of them would have the same default reverse accessor name, it's necessary to distinguish them. As far as I understand it's necessary for at least one of the two fields to have a related_name

JoeyStk:
@charludo This PR is finally ready for review again :)

integreat_cms/cms/views/external_calendars/external_calendar_form_view.py
JoeyStk: Done :white_check_mark:

Issue #2915, opened 2024-07-18
Add test for translation cancel bulk action


MizukiTemma:
### Motivation

This is a follow-up issue for #2007 . Let's add some test for this new functionality 🧪 

### Proposed Solution

Add tests for the new bulk action which resets `currently_in_translation` status.

### Alternatives

Leave it untested and live with higher threat of bugs 🐛

MizukiTemma:
Not blocked anymore as #2007 is completed.

PR #3117, opened 2024-10-06
Test/bulk cancel translation process

theresantonie:
### Short description
New test module to test the functionality of the `CancelTranslationProcess` class in `page_bulk_actions.py`.

### Proposed changes
Add this test module to the general testing process and the needed test data to `test_data.json`:
pages: [30, 31] (Region: nurnberg)
pagetranslations: [100,101]

### Side effects
Changed expected results for other tests:
sitemap_nurnberg_de.xml
num_expected_queries for nurnberg_de 78 -> 88

### Resolved issues
Fixes: #2915


integreat_cms/cms/fixtures/test_data.json
```diff
@@ -3390,6 +3390,52 @@
...
+    "pk": 30,
+    "fields": {
+      "created_date": "2023-08-18T10:31:48.247Z",
+      "lft": 1,
+      "rgt": 2,
+      "tree_id": 11,
+      "depth": 1,
+      "parent": null,
+      "region": 1,
+      "explicitly_archived": true,
```
david-venhoff, line 3404:
```suggestion
      "explicitly_archived": false,
```

Technically, this bulk action should not be available for archived pages (I still works for them though ^^). But I think it would be better to test the bulk action only for pages that are not archived, as that is the intended use case

integreat_cms/cms/fixtures/test_data.json
```diff
@@ -3390,6 +3390,52 @@
...
+    "pk": 31,
+    "fields": {
+      "created_date": "2023-08-18T10:31:48.247Z",
+      "lft": 1,
+      "rgt": 2,
+      "tree_id": 12,
+      "depth": 1,
+      "parent": null,
+      "region": 1,
+      "explicitly_archived": true,
```
david-venhoff, line 3427:
```suggestion
      "explicitly_archived": false,
```

david-venhoff (commented):
Thanks, this looks mostly good to me.

Could you also verify that the a message is shown that the page is not in translation if the page was not in translation before, and that a message is shown that the status was successfully cancelled if the page was in translation before?

For reference, you could do something like that: https://github.com/digitalfabrik/integreat-cms/blob/develop/tests/cms/views/feedback/test_region_feedback_actions.py#L55-L59

theresantonie:
@david-venhoff 
Thank you for the suggestion regarding the log messages.
I will implement it accordingly.

Regarding the archived pages:
There were some problems with other tests not passing, because the new pages were pulled into those tests and messed with the expected results. 
There are many necessary changes and I haven't found a solution to some of them, yet.

theresantonie:
@david-venhoff 
I have added the checks for the log messages.
During my tests, the messages were in German, so I had to add support for English and German log messages.

I also un-archived the pages and moved them to the Nürnberg region to reduce the impact on other tests.
This worked well, but the Editor and Management roles don't have the necessary rights in the Nürnberg region,  
this would have to be added in the test_data.json. (But the side effects here are unclear to me, so far.)

integreat_cms/cms/fixtures/test_data.json
theresantonie: page has been un-archived.

integreat_cms/cms/fixtures/test_data.json
theresantonie: page has been un-archived.

MizukiTemma:
@theresantonie 
We can specify the language of error/success messages. See [those lines](https://github.com/digitalfabrik/integreat-cms/blob/a8cc206938c44f5bc9ce2e136e0e3db5102bffbc/tests/mt_api/deepl_api_test.py#L92-L94)

theresantonie:
@MizukiTemma 
Thank you for the hint. I was already wondering.

JoeyStk:
@theresantonie Thank you a lot for your work thus far. Since this PR hasn't changed in a while, I was wondering if you still  want to finish this PR or if someone else should take over :) Whatever works best for you :)

tests/cms/utils/test_cancel_translation.py
```diff
@@ -0,0 +1,93 @@
...
+    for page_id in IDS_PAGES_NOT_IN_TRANSLATION_PROCESS:
+        assert (
+            not Page.objects.filter(id=page_id)
+            .first()
+            .get_translation_state(LANGUAGE_SLUG)
+            == translation_status.IN_TRANSLATION
+        )
+
+    kwargs = {"region_slug": REGION_SLUG, "language_slug": LANGUAGE_SLUG}
+    page_ids = IDS_PAGES_IN_TRANSLATION_PROCESS + IDS_PAGES_NOT_IN_TRANSLATION_PROCESS
```
JoeyStk, line 53:
```suggestion
    combined_page_ids = IDS_PAGES_IN_TRANSLATION_PROCESS + IDS_PAGES_NOT_IN_TRANSLATION_PROCESS
```
Super nit: I'm not even sure if this actually makes it better, so feel free to reject this suggestion :) But while reading the test line 79 was the only one that wasn't immediately clear to me. But again, looking up what page_ids was took a few seconds and is definitely justifiable

JoeyStk (approved):
Thank you very much. I think the test makes sense is good to read :) I only have one super nit :)

jonbulz (approved):
Comments are just nitpicks, so I'll approve already :)

tests/cms/utils/test_cancel_translation.py
```diff
@@ -0,0 +1,95 @@
...
+    settings.LANGUAGE_CODE = "en"
+
+    for page_id in IDS_PAGES_IN_TRANSLATION_PROCESS:
+        assert (
+            Page.objects.filter(id=page_id).first().get_translation_state(LANGUAGE_SLUG)
+            == translation_status.IN_TRANSLATION
+        )
+    for page_id in IDS_PAGES_NOT_IN_TRANSLATION_PROCESS:
+        assert (
+            not Page.objects.filter(id=page_id)
```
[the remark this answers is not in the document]
MizukiTemma: Uhh 🤔 Our repository contains both the way and  `code_style.sh` accepts both. Maybe we can once discuss about this thema in Weekly 😸

PR #3120, opened 2024-10-07
Improve the performance of `get_region_links`

david-venhoff:
### Short description

I noticed that postgres could not optimize our query which loads links for a specific region very well. A dramatic speedup is possible by making a multiple smaller queries and `union`ing them together, instead of making a big query with a more complex condition.

### Proposed changes

- use `union all` to speed up the `get_region_links` query by up to 90% (Augsburg)
- Add a new parameter to `get_urls` and `filter_urls` to prevent the prefetch. The prefetch typically takes more time than the query. Thus, it can save more than 50% when the prefetch is not needed.

### Side effects

- #3079 on the production data now takes one minute, instead of six minutes :tada: (Though there is still a lot of optimization potential)
- faster region linkcheck view loading 

### Future work
For testing, I added an index on the link model and got another 30-40% performance improvement. However, this will have to be a pr at django-linkcheck

### Resolved issues

Fixes: /

### DIFF
--- a/integreat_cms/cms/utils/linkcheck_utils.py
+++ b/integreat_cms/cms/utils/linkcheck_utils.py
@@ -107,14 +107,18 @@ def get_region_links(region: Region) -> QuerySet:
     organizations = Organization.objects.filter(region=region, archived=False)
     # Get all link objects of the requested region
     region_links = Link.objects.filter(
-        Q(page_translation__id__in=latest_pagetranslation_versions)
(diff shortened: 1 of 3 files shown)


MizukiTemma (approved):
Super cool 😻 Soooo quicker than before! Thank you!

PeterNerlich (approved):
Good work! Can't wait to see this live.

PR #3123, opened 2024-10-10
Normalize custom tinymce icons, re-vectorize no-translate.svg

PeterNerlich:
### Short description

I noticed our icons are kind of all over the place in terms of dimensions, as well as some having a non-black color. While cleaning up on a whim I also noticed that the no-translate icon looked very weird up close, likely having been vectorized from a low resolution pixel graphic by some automatic algorithm.

### Proposed changes

- Normalize the icons directly in `integreat_cms/static/src/`
  - plain SVG format
  - rescale to be 24×24 px
  - no fill color or extra attributes
- Re-vectorize `no-translate.svg`
  - proper, defined corners
  - make it match up with pixel borders → appear crisp
  - move upwards to appear visually centered as a button in tinymce

<details><summary>Comparison</summary>

#### TinyMCE

Before:
image

After:
image

#### No-Translate with pixel grid overlaid

Before:
image

After:
image

</details

### Side effects

- at a few points, icons are not scaled to fit their containing element but were displayed at their original size, leading to very inconsistent visual appearance between which icon was used
  - for me, this is a desired side effect and what initially prompted me to do this

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #3124, opened 2024-10-11
[2024-11-07] TinyMCE: Insert contact, track usages

### LINKED ISSUE #2959: [2024-11-07] Add contact as choice for "insert" in TinyMCE

PeterNerlich:
### Short description

We want to embed contacts on pages and in other content. This PR addresses everything needed to edit contact embeddings on content translations and to track where contacts have been embedded. The latter part has been heavily inspired by the `linkcheck` library.

*This feels like my biggest changes write up yet, even though development didn't take that long at all. I am hopeful that this just means that I got more verbose in describing the changes*

### Proposed changes

- Add contacts to **content search API** endpoint
  - This required a bit of trickery because, as a user, I want to search for any part of the details of a contact and get a list with the closest match up top *(e.g. a query `example.com mariana` should match a contact with name `Mariana Musterfrau` and email `m.musterfrau@example.com`)*

- Add a **custom TinyMCE plugin** to embed contacts via a button and dialog, similar to our custom link plugin
  - Contact dialog:
    - Features a **search bar** and a **drop down** to search for and select a contact
    - **Sort** the retrieved search results **by relevancy** – an object with few but longer matches over a few fields should be significantly higher ranked than many short matches all over the place *(I tried to leave helpful comments, please point out where you think these could still be improved)*
    - Unfortunately, if a selected value is still present in the data when the list objects for the dropdown are changed, the selection will not change to the topmost value. While desirable in other contexts, in our case we would prefer the value to be reset to the best fitting result when the query changed. This seems however not possible within TinyMCEs dialog framework
    - Embedded contacts are rendered as cards for **preview**
      - These are div boxes with **`contenteditable="false"`**, which TinyMCE respects – when clicking anywhere within the element, the outermost div box with that attribute is selected, and the box can only be deleted/replaced as a whole by users
      - The defining features of these elements and what defines them as embedded contacts is the **`data-contact`** attribute, whose value shall be a **valid contact id**
      - There is **custom CSS** to style the preview, which applies only within the editor
      - If only an id was entered directly without the contact coming up as part of a previous search, the client has no information to render the preview card. It will show up empty but won't cause problems beyond that
      - **The client side rendered preview of the contact card will be stripped and replaced on the server, only the contact id is important**
    - Selecting a contact card in the editor produces a **context toolbar** above/below the element, allowing for easy reopening of the contact dialog to **change** which contact is being embedded, as well as **removing** the element
      - The icon for changing the contact is being displayed too big. This will be fixed by #3123 

- When receiving submitted content translation forms on the server, search the content for embedded contacts and render the "official" HTML for the card
  - If the id cannot be parsed or there is no contact with that id in the system, only the attribute **`data-invalid`** is applied – ***TODO: decide on how to handle and how/whether to present invalid embedded contacts in TinyMCE and to the end user***
    -  ***TODO: add message when submitted translation contains invalid contacts***
  - ***TODO: style on app side & page preview through common stylesheet or other solutions***
  - ***TODO: restrict contacts to contacts of the region***
  - ***TODO: ensure contacts are properly displayed in other languages, especially rtl ones***

- **Track contact usages** in content
  - Add a **`ContactUsage` model**, similar to the `Link` model of `linkcheck`, associating a contact object with a generic object of a specific content type
  - Add a **`Contactlist` class**, similar to the `Linklist` class of `linkcheck`, to facilitate updating `ContactUsage` instances for a specific content type
  - **Subclass `Contactlist`** for page translations, imprint translations, event translations and ~~POI tra~~ location translations, and make them centrally accessible in `integreat_cms/cms/contactlist.py`, similar to `integreat_cms/cms/linklist.py`
  - Add **post save / post delete listeners** to content translations to update `ContactUsage` objects
  - Add a **`findcontacts` management command** to update `ContactUsage` instances for content objects across the whole database

- In the contact form, **show where contacts are being embedded**
  - Move the template variable `region_default_language` into `integreat_cms/cms/templates/_related_contents_table.html` where it belongs
  - Query content objects where the current contact is embedded and pass them to the template

- ***TODO: figure out mypy with generically specifying "subclasses of Contactlists"***

### Side effects

- Performance might degrade further, since we're doing more or less the same that `linkcheck` is already doing, except that we don't have to wait to update and check URLs since our equivalent (`Contact`) are in our database and already our single source of truth
  - I did find one or two optimizations we can do compared to linkcheck, though I did not actually do any performance checks
- Some names might seem weird and less maintenance-friendly than alternatives, but I decided to keep it close to `linkcheck` in hopes that more apparent similarities with it would help maintenance more

### Resolved issues

Fixes: #2959


integreat_cms/cms/models/contact/contact.py
```diff
@@ -54,6 +62,39 @@ def region(self) -> Region:
...
+        # e.g. by finding the length of the longest common substring between each field and the original query,
+        # taking the square of that value to obtain something representing the "contribution" of that field
+        # (so longer matches in only a few fields get a much higher value than many short matches all over)
+        # and then summing those together to obtain an overall score of how relevant that object is to the query,
+        # but that would require us find the longest common substring on the db level,
+        # and that feels a bit overkill for now  (it will likely be confusing to re-discover and maintain,
+        # especially if we were to also employ fuzzy matching – which would be much preferred, if we can do it)
+
+        return cls.objects.filter(q, location__region=region)
+
```
charludo, line 97: This is definitely a respectable approach. I think implementing fuzzy search ourself is overkill, and not really maintainable in the long run.

Did you loook into exitseng solutions for this? E.g.:
https://django-haystack.readthedocs.io/en/master/toc.html
https://gist.github.com/shahriar350/aebcaef26c79c575834f4c89058cf6f3

Or (probably a lot easier): https://docs.djangoproject.com/en/5.1/ref/contrib/postgres/search/#searchvector

integreat_cms/cms/utils/content_utils.py
```diff
@@ -139,6 +141,120 @@ def update_internal_links(link: HtmlElement, language_slug: str) -> None:
                 link.append(new_html)
 
 
+def render_contact_card(
+    contact_id: int, fetched_contacts: dict[int, Contact]
+) -> HtmlElement:
+    """
+    Produces an html element for the contact.
+
```
charludo, line 149: I think this might be where jinja shines 😄

charludo (commented):
WOW, thank you for this PR - this is a real leap forward! 😍 🎉 

I have obviously not been able to do a full review yet, so I will just leave a comment with some thoughts. Please note that I have not been part of the discussion on this for the past months, so some of the things below might just plain not be applicable.

- From a user's perspective:
    - The TinyMCE plugin works really well, including changing and deleting the contact
    - However, I am not a super big fan of the separation between the search field and the drop down, since it means you can not see more than one result while searching and have to perform additional clicks to select from the list. I had a similar problem a while back with another django project, and ended up using [TomSelect](https://github.com/orchidjs/tom-select) for this. [Here](https://github.com/charludo/ycms/blob/develop/ycms/static/src/js/utils/autocomplete.ts) is an example on how it could be used, basically allowing you to have a combined search+dropdown which gets data from an API endpoint.
    - very small issue: after deleting a contact, the "popup" remains until you click somewhere else
    - and lastly, I don't hink a field for inputting the ID should be shown to the user

Apart from that, I have no wishes/comments from a user's perspective! Really good job!!

- From a technical perspective:
    - I don't doubt that everything works as intended, but I think there is some potential to increase maintainability in the long run.
    - the two comments below are concrete cases of this
    - but (and this is the point where I might just have missed a discussion about it): why do we need the `ContactUsage` model and associated functions at all? Would it not be easier to just leave a div with `data-contact` and the contact's ID in the content, and then have the app query an api endpoint to retrieve the actual contact? Then we would not have to keep the contact itself updated throughout all the content objects.

PeterNerlich:
> WOW, thank you for this PR - this is a real leap forward! 😍 🎉

A review to a draft PR already? This must be my lucky day!

>   * However, I am not a super big fan of the separation between the search field and the drop down […] I had a similar problem a while back with another django project, and ended up using [TomSelect](https://github.com/orchidjs/tom-select) for this.

This is an incredibly helpful tip, thank you so much! I have been gnawing over this and resigned that it's just gonna be this way until someone eventually throws out TinyMCEs dialog system altogether. I will be looking at that next.

>   * very small issue: after deleting a contact, the "popup" remains until you click somewhere else

Right, I have not been able to figure that out and decided it was just a minor inconvenience and fine to leave. I'm all ears if someone knows how to fix this.

>   * and lastly, I don't think a field for inputting the ID should be shown to the user

I will see whether I can hide it, but I think the field had to be present as a field in order for it to work the way

>   * why do we need the `ContactUsage` model and associated functions at all? Would it not be easier to just leave a div with `data-contact` and the contact's ID in the content, and then have the app query an api endpoint to retrieve the actual contact? Then we would not have to keep the contact itself updated throughout all the content objects.

We asked the app team what they prefer at the conference and decided that we just keep at delivering a fully rendered HTML that they just embed. I believe this embedding of HTML into the native app is already quite a source of frustration, and having to parse that and modify it *(in HTML only, or worse, injecting native widgets inside HTML, which is bound to be a nightmare)* is unnecessary complication.
Moreover, we already have the infrastructure in `linkcheck` to do exactly this kind of thing, and where links are set to always display as the name of the target content, we employ it to replace the text of those links any time the title of the target content is changed. I view that solution and our situation with contacts as instances of exactly the same problem, and would find it more confusing if we solved it in a completely different way. Of course, I ended up reusing none of `linkcheck`s code because it was not generic enough to do without the notion of something like `URL`s, but maybe it can be generalized further and contributed back to `linkcheck`

integreat_cms/cms/models/contact/contact.py
PeterNerlich: To be honest, barely, and I stopped at trigram similarity. I did a brief search I think, but I might have already been too deep in the technical details. I do tend to fail to look beyond the documentation and capabilities of the system right in front of me for dependencies to add that might solve this

charludo:
> We asked the app team what they prefer at the conference and decided that we just keep at delivering a fully rendered HTML that they just embed. I believe this embedding of HTML into the native app is already quite a source of frustration, and having to parse that and modify it _(in HTML only, or worse, injecting native widgets inside HTML, which is bound to be a nightmare)_ is unnecessary complication. 

Thanks, this makes sense! I had not thought about the difficulty of embedding the HTML in the apps.

> Moreover, we already have the infrastructure in `linkcheck` to do exactly this kind of thing, and where links are set to always display as the name of the target content, we employ it to replace the text of those links any time the title of the target content is changed. I view that solution and our situation with contacts as instances of exactly the same problem, and would find it more confusing if we solved it in a completely different way. Of course, I ended up reusing none of `linkcheck`s code because it was not generic enough to do without the notion of something like `URL`s, but maybe it can be generalized further and contributed back to `linkcheck`

I get that it would be nice to use existing code, but if, as you say, the code was not really reusable, then I am not sure sticking to the same ideas is necessarily the right thing. As your PR shows, there's a LOT of code required for ensuring that contacts stay up-to-date, so I am not sure if another solution might not be easier:

If the app team does not want to handle replacing the contact ID with the actual contact, why don't we? 😄 
I am thinking that in the `transform_page` function in `integreat_cms/api/v3/pages.py`, we could do the replacement just-in-time.  
A lot of similar data-wrangling is already going on in there, so it would also not be a completely new paradigm to go by.

Essentially, we would only need to 
- find the contact id
- render a jinja template with that contact's DB entry
- replace the `data-contact` div with the result

PeterNerlich:
> If the app team does not want to handle replacing the contact ID with the actual contact, why don't we? we could do the replacement just-in-time.

I brought that up it the discussion about links and was told there was a decision that we don't want to interpolate data like that but rather save it all up-to-date in our database, so I ruled that out here as well, but I do agree that this would not be a bad way to solve it

charludo:
> > If the app team does not want to handle replacing the contact ID with the actual contact, why don't we? we could do the replacement just-in-time.
> 
> I brought that up it the discussion about links and was told there was a decision that we don't want to interpolate data like that but rather save it all up-to-date in our database, so I ruled that out here as well, but I do agree that this would not be a bad way to solve it

I see. And I do see the appeal of the DB being 1:1 what users see. 

But since Contacts are meant to be decoupled form the page content, I think that argument has less merit here, especially since your implementation is a testament to the additional complexities that come with *not* interpolating the contacts.

(Do not get me wrong, you did impressive works here - I am not arguing against anything you did here.)

Maybe some other people from the team have opinions on this?

david-venhoff:
@PeterNerlich 
> Of course, I ended up reusing none of linkchecks code because it was not generic enough to do without the notion of something like URLs

Why don't we just give each contact a unique url? Then we could simply reuse our existing linkcheck infrastructure.

@charludo 

> Maybe some other people from the team have opinions on this?

I am not sure I followed the conversation correctly, so please correct me if I am wrong, but I think either way we need to know if a contact is used in a page so that we can prevent deleting a contact if that is still the case. If we just store the id in the html, then we have no way to check that, except for iterating through all page translations (Which could be feasible if we optimize it tbh).

PeterNerlich:
> Why don't we just give each contact a unique url?

Please do explain. I do not see how we can make `linkcheck` apply this to non-links in HTML content or otherwise separate out links in the link sense from contacts *(we don't want references to contacts to show up in or be affected by the linklist or search&replace views, but we want to be able to query an embedded contact across all different content types – pages, events etc., while the `Linklist` class only allows specialization in regards to the db objects to be searched for links)*

PeterNerlich:
@david-venhoff 

> Why don't we just give each contact a unique url?

Please do explain. I do not see how we can make `linkcheck` apply this to non-links in HTML content or otherwise separate out links in the link sense from contacts *(we don't want references to contacts to show up in or be affected by the linklist or search&replace views, but we want to be able to query an embedded contact across all different content types – pages, events etc., while the `Linklist` class only allows specialization in regards to the db objects to be searched for links)*

david-venhoff:
Sorry. What I meant is what if we did what @charludo suggested but instead used a link so that it can be tracked?

* Embedding a contact: Inserts something like `<a href="https://integreat.app/augsburg/de/contacts/1">…</a>`
* Maybe we could also add `contenteditable=false` and still render the same preview
* Remove those links in the api and render the correct html for contacts

> we don't want references to contacts to show up in or be affected by the linklist or search&replace views

Do you think it would be a problem if it showed up there? Also If we don't update the `get_region_links` function, those links will not show up in the region link view, only the global link view

charludo:
> * Embedding a contact: Inserts something like `<a href="https://integreat.app/augsburg/de/contacts/1">…</a>`
> 
> * Maybe we could also add `contenteditable=false` and still render the same preview

That sounds fantastic IMO.

PeterNerlich:
I switched the div over to being an anchor, but it keeps getting stripped out by TinyMCE when inserting it into the content, only the children with the details remain. If anyone has dealt with a link wrapped around a div in TinyMCE yet, now is your time to shine

Issue #3105, opened 2024-10-04
[2024-11-07] Disallow copying for imported events


JoeyStk:
### Describe the Bug

In our current implementation it's possible to copy imported events. However this doesn't make sense at the moment, as only a copy of the imported event is made and not a copy that is editable. Therefore we want to disallow copying for imported events.

### Steps to Reproduce

1. Go to event list on our test system 
2. See that you can copy imported events

### Expected Behavior

You can't copy imported events.

### Actual Behavior

You can.

### Additional Information

This is only a temporary solution until #3106

PR #3126, opened 2024-10-12
[2024-11-07] Disable copying for imported events

JoeyStk:
### Short description

In the events list disallow copying by disabling the action and greying it out

### Proposed changes

- Grey out and disable copying action for events that have been imported 

### Side effects

None

### Resolved issues

Fixes: #3105

### DIFF
--- a/integreat_cms/cms/models/events/event.py
+++ b/integreat_cms/cms/models/events/event.py
@@ -29,6 +29,12 @@ if TYPE_CHECKING:
     from ..users.user import User
 
 
+class CouldNotBeCopied(Exception):
+    """
+    Exception for events that can't be copied
+    """
+
+
 class EventQuerySet(ContentQuerySet):
     """
     Custom QuerySet to facilitate the filtering by date while taking recurring events into account.
@@ -251,7 +257,12 @@ class Event(AbstractContentModel):
 
         :param user: The user who initiated this copy
         :return: A copy of this event
+
+        :raises CouldNotBeCopied: When event can't be copied
         """
+        if self.external_calendar:
+            raise CouldNotBeCopied
+
         # save all translations on the original object, so that they can be copied later
         translations = list(self.translations.all())
 
--- a/integreat_cms/cms/templates/events/event_list_row.html
+++ b/integreat_cms/cms/templates/events/event_list_row.html
@@ -141,13 +141,19 @@
         {% endif %}
         {% has_perm 'cms.change_event' request.user as can_edit_event %}
         {% if can_edit_event %}
-            <form action="{% url 'copy_event' event_id=event.id region_slug=request.region.slug language_slug=language.slug %}"
-                  method="post">
-                {% csrf_token %}
-                <button class="btn-icon" title="{% translate "Copy event" %}">
+            {% if not event.external_calendar %}
+                <form action="{% url 'copy_event' event_id=event.id region_slug=request.region.slug language_slug=language.slug %}"
+                      method="post">
+                    {% csrf_token %}
+                    <button class="btn-icon" title="{% translate "Copy event" %}">
+                        <i icon-name="copy"></i>
+                    </button>
+                </form>
+            {% else %}
+                <button title="{{ cannot_copy_title }}" class="btn-icon" disabled>
                     <i icon-name="copy"></i>
                 </button>
-            </form>
+            {% endif %}
             <button title="{% translate "Archive event" %}"
                     class="confirmation-button btn-icon"
                     data-confirmation-title="{{ archive_dialog_title }}"
--- a/integreat_cms/cms/views/events/event_actions.py
+++ b/integreat_cms/cms/views/events/event_actions.py
@@ -17,6 +17,7 @@ from django.views.decorators.http import require_POST
 from ...constants import status
 from ...decorators import permission_required
 from ...models import POITranslation, Region
+from ...models.events.event import CouldNotBeCopied
 
 if TYPE_CHECKING:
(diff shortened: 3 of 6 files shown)


integreat_cms/cms/templates/events/event_list_row.html
```diff
@@ -141,13 +141,19 @@
...
+            {% if not event.external_calendar %}
+                <form action="{% url 'copy_event' event_id=event.id region_slug=request.region.slug language_slug=language.slug %}"
+                      method="post">
+                    {% csrf_token %}
+                    <button class="btn-icon" title="{% translate "Copy event" %}">
+                        <i icon-name="copy"></i>
+                    </button>
+                </form>
+            {% else %}
+                <button title="{{ cannot_delete_title }}" class="btn-icon" disabled>
```
david-venhoff, line 153:
```suggestion
                <button title="{{ cannot_copy_title }}" class="btn-icon" disabled>
```

integreat_cms/cms/views/events/event_context_mixin.py
```diff
@@ -51,6 +51,9 @@ def get_context_data(self, **kwargs: Any) -> dict[str, Any]:
                 "help_text": _(
                     "Create an event location or start typing the name of an existing location. Only published locations can be set as event venues."
                 ),
+                "cannot_delete_title": _(
```
david-venhoff, line 54:
```suggestion
                "cannot_copy_title": _(
```

david-venhoff (commented):
Looks good to me in general. Could you also add a check in the `copy` function of the event model that the event does not have an external calendar? 
The reason why I'd like to have this is that I think that we should generally not validate things on the frontend only (As frontend validation can be easily sidestepped or forgotten)

integreat_cms/cms/views/events/event_actions.py
```diff
@@ -72,6 +72,9 @@ def copy(
     region = request.region
     event = get_object_or_404(region.events, id=event_id)
 
+    if event.external_calendar:
```
JoeyStk, line 75: @david-venhoff Is this what you had in mind?

david-venhoff: Almost, but I think it would be preferable to have this check in the `event.copy` method. The advantage would be, that with the current solution, we would have to keep in mind to always check the external calendar if we wanted to call `event.copy` from a new solution in the future.

`event.copy` could return an error or bool indicating whether the copy was created successfully.

JoeyStk: Yes, very true. I added an Exception that is thrown when one attempts to copy an event that shouldn't be copied. Can you please check whether or not I followed the conventions and placement of the exception? :)

david-venhoff: Could you also document that this method can now throw an exception?

JoeyStk: Should be applied now :)

integreat_cms/cms/models/events/event.py
```diff
@@ -29,6 +29,22 @@
...
+class CouldNotBeCopied(Exception):
+    """
+    Exception for events that can't be copied
+    """
+
+    def __init__(
+        self,
+        message="Event couldn't be copied because it's from a external calendar",
+        source=None,
+        destination=None,
```
david-venhoff, line 41:
```suggestion
```
Are these unused?

integreat_cms/cms/views/events/event_actions.py
```diff
@@ -59,7 +60,7 @@ def archive(
 @permission_required("cms.change_event")
 def copy(
     request: HttpRequest, event_id: int, region_slug: str, language_slug: str
-) -> HttpResponseRedirect:
+) -> HttpResponseRedirect | None:
```
PeterNerlich, line 63:
```suggestion
) -> HttpResponseRedirect:
```
This is not needed, right? Even when copying is not allowed we redirect the browser back to the view

JoeyStk: Oh, you're right :)

MizukiTemma (approved):
Looks good 👍 

A test for this constraint would be nice, but we can implement it later if it is not handy.

JoeyStk:
> A test for this constraint would be nice, but we can implement it later if it is not handy.

@MizukiTemma Good point! A test was added :)

integreat_cms/cms/models/events/event.py
```diff
@@ -29,6 +29,18 @@
...
+class CouldNotBeCopied(Exception):
+    """
+    Exception for events that can't be copied
+    """
+
+    def __init__(
+        self,
+        message: str = "Event couldn't be copied because it's from a external calendar",
+    ) -> None:
+        super().__init__(message)
```
david-venhoff, line 41:
```suggestion
```

I think its fine to completely get rid of this part, because this exception is not supposed to ever be printed (Because it should always get catched)

integreat_cms/cms/views/events/event_actions.py
```diff
@@ -72,7 +73,15 @@ def copy(
     region = request.region
     event = get_object_or_404(region.events, id=event_id)
 
-    event.copy(request.user)
+    try:
+        event.copy(request.user)
+    except CouldNotBeCopied:
+        messages.error(
+            request, _("Event couldn't be copied because it's from a external calendar")
```
david-venhoff, line 80:
```suggestion
            request, _("Event couldn't be copied because it's from an external calendar")
```

david-venhoff (approved):
Thanks, LGTM!

Issue #3107, opened 2024-10-04
[2024-11-07] iCal: Disallow publishing through version history


JoeyStk:
### Describe the Bug

Some imported events are imported as draft. It shouldn't be possible for authors to publish them. However it's still possible to publish them through the version history.

### Steps to Reproduce

1. Go to an imported event and open the detail view
2. Double check that the imported event is a draft
3. Go to 'Show' to open up the version history
4. See that it's possible to publish the event through the version history

### Expected Behavior

It's not possible for authors to publish events

### Actual Behavior

It is, though the version history.

### Additional Information

None

PR #3127, opened 2024-10-12
Disallow publishing imported events through version history

JoeyStk:
### Short description

This PR removes the buttons from the version history for imported events from an external calendar.

### Proposed changes

- Remove buttons

### Side effects

- I'm not 100% sure if this is the best way to do it, or if maybe some explanatory text would be helpful. What do you think?

### Resolved issues

Fixes: #3107 

### DIFF
--- a/integreat_cms/cms/views/content_version_view.py
+++ b/integreat_cms/cms/views/content_version_view.py
@@ -13,6 +13,7 @@ from django.views.generic.base import TemplateView
 from django.views.generic.detail import SingleObjectMixin
 
 from ..constants import status
+from ..models import Event
 
 if TYPE_CHECKING:
     from typing import Any
@@ -274,6 +275,24 @@ class ContentVersionView(PermissionRequiredMixin, SingleObjectMixin, TemplateVie
 
         :return: The rendered template response
         """
+        if self.model is Event and self.get_object().external_calendar:
+            logger.info(
+                "%s %s can not change its status as it was imported from an external calendar",
+                self.model_name,
+                self.get_object(),
+            )
+
+            messages.info(
+                request,
+                _(
+                    "%s %s can not change its status as it was imported from an external calendar"
+                )
+                % (
+                    self.model_name,
+                    self.get_object(),
+                ),
+            )
+            return redirect(self.versions_url)
 
         if desired_status := request.POST.get("status"):
             try:
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -5533,6 +5533,10 @@ msgid "Minor edit"
 msgstr "Geringfügige Änderung"
 
 #: cms/templates/content_versions.html
+msgid "The status of imported events can't be changed"
+msgstr "Der Status von importierten Veranstaltungen kann nicht geändert werden"
+
+#: cms/templates/content_versions.html
 msgid "Discard auto save"
 msgstr "Automatische Speicherung verwerfen"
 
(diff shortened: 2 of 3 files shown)


charludo:
I think removing the buttons is valid, another option could be to disable them / gray them out? 
In both cases, a short help text would probably be nice though, I agree.

However, could you please also add some backend validation to ensure imported events are not published?

JoeyStk:
> However, could you please also add some backend validation to ensure imported events are not published?

Fair point! I'm still trying to evaluate what the most sensible validation would be. Do we just not want some users (like authors) to not publish them or do we not want any user to be able to publish them? If so, wouldn't that defeat the purpose of importing the events in the first place? :thinking: If you have any suggestions, I would be interesting in hearing them :)

charludo:
> > However, could you please also add some backend validation to ensure imported events are not published?
> 
> Fair point! I'm still trying to evaluate what the most sensible validation would be. Do we just not want some users (like authors) to not publish them or do we not want any user to be able to publish them? If so, wouldn't that defeat the purpose of importing the events in the first place? 🤔 If you have any suggestions, I would be interesting in hearing them :)

Oops, yes, "ensure imported events are not published" was probably an unfortunate way to formulate this 😆 

If it's just the authors, I would go ahead and prevent them from publishing imported events, ever, since AFAIK they are not allowed to add external calendars anyways, right?

If it's about noone being able to publish draft events imported from external sources, I would suggest enforcing this for all endpoints, except for the initial import.

But the first one is probably easier and more accurate to the issue description

JoeyStk:
@charludo & @jarlhengstmengel I'm sorry that it took me so long to get back to this. But finally this PR is ready to be reviewed again :)

Most prominent changes:
- background check was added
- The button reappears but now is disabled with a title describing why the button is disabled 

:)

MizukiTemma:
@JoeyStk 
The desired behaviour is that no user can publish or turn imported events into draft or refresh their date, right?

JoeyStk:
Yes, at least for imported events

MizukiTemma (approved):
Looks good :smile_cat: Thank you :+1:

integreat_cms/cms/views/content_version_view.py
```diff
@@ -274,6 +275,24 @@ def post(
 
         :return: The rendered template response
         """
+        if self.model is Event and self.get_object().external_calendar:
+            logger.info(
+                "%s %s can't change its status as it's imported from an external calendar",
```
charludo, line 280:
```suggestion
                "%s %s can not change its status as it was imported from an external calendar",
```

integreat_cms/cms/views/content_version_view.py
```diff
@@ -274,6 +275,24 @@ def post(
...
+                self.model_name,
+                self.get_object(),
+            )
+
+            messages.info(
+                request,
+                _(
+                    "%s %s can't change its status as it's imported from an external calendar"
```
charludo, line 288:
```suggestion
                    "%s %s can not change its status as it was imported from an external calendar"
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -9180,6 +9184,14 @@ msgstr "Link zum {}"
 msgid "The version {} does not exist."
 msgstr "Die Version {} existiert nicht."
 
+#: cms/views/content_version_view.py
+#, python-format
+msgid ""
+"%s %s can't change its status as it's imported from an external calendar"
```
charludo, line 9190:
```suggestion
"%s %s can not change its status as it was imported from an external calendar"
```

charludo (approved):
Fantastic, thank you! 🎉 
(And sorry for taking so long to get to this 😭)

I only have one absolute nit. I checked, and we are using contractions in a bunch of other strings, but the more formal, uncontracted versions are more common. 

Ultimately, this is mostly personal preference, so 100% feel free to ignore this.

Issue #3112, opened 2024-10-04
Delete Calendar prompt appears when hitting `Enter` key in external calendar form


JoeyStk:
### Describe the Bug

When hitting `Enter` in the external calendar form the "Delete calendar" promt appears. 

### Steps to Reproduce

1. Go to external calendar form
2. Fill out the fields and hit `Enter`
3. See popup to delete the calendar

### Expected Behavior

Form gets saved

### Actual Behavior

Popup appears asking whether or not you want to delete the calendar.

### Additional Information

None

Issue #2892, opened 2024-07-07
Default action in forms inconsistent, often DELETE


PeterNerlich:
### Describe the Bug

In various forms we have multiple actions (like *Save as draft* or *Publish*, but also *Delete*). When an input field is focused, pressing <kbd>Enter</kbd> triggers the default action. This is inconsistent with the users perception of what the default action should be; in two cases it was found to be *Delete* (which popped up a confirmation dialog, thankfully).

The default action should be clearly identified throughout our system (likely the dark blue "primary" buttons), and pressing enter from within an input field should trigger that action only.

### Steps to Reproduce

1. Go to *Organizations* 
2. Click into the name field
3. Press <kbd>Enter</kbd>
4. See deletion confirmation popup

### Expected Behavior

Organization is saved with the data in the form

### Actual Behavior

The *Delete* action is triggered instead and a confirmation dialog pops up.

PR #3128, opened 2024-10-12
[2024-11-07] Fix enter saves form

JoeyStk:
### Short description

Hitting `Enter` in our organization or external calendar form opens the "Do you really want to delete this?" popup. This is not the desired behavior. Instead the form should be saved when hitting `Enter`. This issue was because the delete button is the first one on our page and therefore by default the button that's going to be executed when hitting `Enter` (at least this was the result of my research).

### Proposed changes

- Adjust button type in organization and external calendar form. The "Submit"-Button will be executed when hitting `Enter`, which fixes the issue at hand

### Side effects

- I went through the other content types (at least POI, event and pages). There we don't have this issue.
- I'm not aware of any possible side effects by this change

### Resolved issues

Fixes: #3112

### DIFF
--- a/integreat_cms/cms/templates/events/external_calendar_form.html
+++ b/integreat_cms/cms/templates/events/external_calendar_form.html
@@ -21,7 +21,8 @@
                 <div class="flex flex-wrap justify-between gap-4">
                     {% if external_calendar_form.instance.id and perms.cms.delete_externalcalendar %}
                         <div class="flex flex-wrap gap-4">
-                            <button title="{% translate "Delete external calendar" %}"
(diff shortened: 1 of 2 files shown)


charludo (approved):
Nice! 🎉  Works without issues.

david-venhoff (approved):
Thanks, works without problems now!

Technically this should have a changelog entry, because the change to organizations is noticeable to users, I think ^^

juliankehne:
@JoeyStk Tested this and works fine! #2892 can be closed as well, its a duplicate of this bug

Issue #4, opened 2018-10-25
Add translation coverage to "To-Do" dashboard


svenseeberg:
The Dashboard should contain some basic overview of the current translation status.

~This could be a pie chart that visualizes 3 numbers: pages with up to date translations, pages with outdated translations, not translated pages.~

This should be part of the new "To Do" dashboard, see https://github.com/digitalfabrik/integreat-cms/issues/2107#issuecomment-1466087901.

svenseeberg:
This is related to #9

MizukiTemma:
This feature is needed for our OKR.

JoeyStk:
Icon to use: translations
Calculation: Amount of pages that are either outdated or have missing translations
Headline: Veraltete und fehlende Übersetzungen (Link leads to translation coverage)
Paragraph: Ihre Inhalte haben momentan 423 Seiten mit fehlender oder veralteter Übersetzung. Damit die Nutzer:innen von Ihren Inhalten profitieren können, sollten Sie diese übersetzen (lassen).
Button: Also leads to translation coverage

Caveat: Keep loading times in mind. Ideally this should be one single database query as this is the first view inside that CMS-user encounter.

JoeyStk:
The current implementation would require some loops on the Python side which would add a few more database queries and therefore would considerably slow down the loading time on the dashboard. Therefore we need to look into alternative solutions. For details please checkout `feature/add-translation-coverage-row-to-dashboard`

JoeyStk:
It was agreed to solve this with AJAX

PR #3129, opened 2024-10-12
Add translation coverage row to dashboard

JoeyStk:
### Short description

This PR adds the translation coverage row to the dashboard. This branch is already quite old and needs a bit of an update, however I finally wanted to open up this PR to get some feedback in advance

### Proposed changes

- Calculate the number of outdated translations
- Load the row via AJAX to increase loading time

### Side effects

- ~Uff... the calculation doesn't seem great. Is someone aware of a better way to do it?~

### Resolved issues

Fixes: #4 

### DIFF
--- a/integreat_cms/cms/views/dashboard/dashboard_view.py
+++ b/integreat_cms/cms/views/dashboard/dashboard_view.py
@@ -17,6 +17,7 @@ from ...models import Feedback, PageTranslation
 from ...utils.linkcheck_utils import filter_urls
 from ...views.utils.hix import get_translation_under_hix_threshold
 from ..chat.chat_context_mixin import ChatContextMixin
+from ..utils import get_translation_and_word_count
 
 if TYPE_CHECKING:
     from typing import Any
@@ -59,6 +60,10 @@ class DashboardView(TemplateView, ChatContextMixin):
                 ),
                 "broken_link_ajax": reverse(
                     "get_broken_links_ajax",
+                    kwargs={"region_slug": self.request.region.slug},
+                ),
+                "translation_coverage_ajax": reverse(
+                    "get_translation_coverage_ajax",
                     kwargs={"region_slug": self.request.region.slug},
                 ),
             }
@@ -242,3 +247,26 @@ class DashboardView(TemplateView, ChatContextMixin):
             "drafted_pages": drafted_pages,
             "single_drafted_page": single_drafted_page,
         }
+
+    @json_response
+    # pylint: disable=unused-argument, disable=no-self-argument
+    def get_translation_coverage_context(
+        request: HttpRequest, region_slug: str
+    ) -> JsonResponse:
+        r"""
+        Extend context by info on translation coverage of pages
+
+        :return: Dictionary containing the context for translation coverage of pages in a region
+        """
+        total_missing = 0
+        total_outdated = 0
+
+        translation_count, _ = get_translation_and_word_count(request.region)
+
+        for counts in translation_count.values():
+            total_missing += counts.get("MISSING", 0)
+            total_outdated += counts.get("OUTDATED", 0)
+
+        total = total_missing + total_outdated
+
+        return JsonResponse(data={"number_of_missing_or_outdated_translations": total})
--- a/integreat_cms/cms/templates/dashboard/_todo_dashboard_widget.html
+++ b/integreat_cms/cms/templates/dashboard/_todo_dashboard_widget.html
@@ -30,5 +30,8 @@
             {% include "dashboard/todo_dashboard_rows/_automatically_saved_pages_row.html" %}
             {% include "dashboard/todo_dashboard_rows/_drafted_pages_row.html" %}
         {% endif %}
+        {% if perms.cms.change_page %}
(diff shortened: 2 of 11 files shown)


jarlhengstmengel:
Do you have a reason why you choose a different approach then `translation_coverage_view.py`?

jarlhengstmengel:
> Do you have a reason why you choose a different approach then `translation_coverage_view.py`?

Ah sorry, I overlooked the point about AJAX

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -259,3 +263,49 @@ def get_drafted_pages(
...
+        number_of_outdated_pages = 0
+
+        if not request.region.default_language:
+            return {}
+
+        languages = LanguageTreeNode.objects.filter(region=request.region).exclude(
+            language=request.region.default_language
+        )
+
+        pages_in_region = Page.objects.filter(region=request.region)
```
jarlhengstmengel, line 286: At the moment explicit and implicit archived pages are also counted into the pages. I think it's more consistent with the Translation Report to exclude them.

jarlhengstmengel: Otherwise the calculation looks fine

jarlhengstmengel (changes requested):

integreat_cms/cms/views/dashboard/dashboard_view.py
JoeyStk: Oh, very good point! Thanks :blush:

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -241,3 +245,55 @@ def get_drafted_pages(
...
+            return {}
+
+        languages = LanguageTreeNode.objects.filter(region=request.region).exclude(
+            language=request.region.default_language
+        )
+
+        pages_in_region = Page.objects.filter(
+            region=request.region, explicitly_archived=False
+        )
+        for page in pages_in_region:
```
jarlhengstmengel, line 271: Somehow the for loop doesn't catch implicitly archived pages right. When I test it locally with the data of 'Stadt Augsburg' it doesn't count the one implicitly archived page in the data

integreat_cms/static/src/js/dashboard/utils.ts
```diff
@@ -0,0 +1,38 @@
+/* const showAllTotalNumbers = () => {
```
jarlhengstmengel, line 1: Do you have a reason to keep this out commented section?

jarlhengstmengel (changes requested):

integreat_cms/cms/views/dashboard/dashboard_view.py
```diff
@@ -241,3 +245,55 @@ def get_drafted_pages(
...
+        for pagetranslation in published_foreign_translations:
+            if pagetranslation.is_outdated:
+                number_of_outdated_pages += 1
+
+        return JsonResponse(
+            data={
+                "number_of_missing_or_outdated_translations": number_of_missing_translations
+                + number_of_outdated_pages
+            }
+        )
```
charludo, line 299: I have tried to find a simplified calculation. It works, though the number is lower than what your calculation finds - also I am not sure it's actually faster. 



```suggestion
        non_archived_pages = request.region.non_archived_pages
        language_count = len(request.region.languages)
        possible_translations = non_archived_pages.count() * language_count

        all_translations = PageTranslation.objects.filter(
            page__in=non_archived_pages
        ).only("pk")
        non_outdated_translations = sum(
            1 for translation in all_translations if not translation.is_outdated
        )

        return JsonResponse(
            data={
                "number_of_missing_or_outdated_translations": possible_translations
                - non_outdated_translations
            }
        )
```

charludo (approved):
Nice, thank you! I decided to approve this PR even though I am not 100% sure if the calculation of the number of missing pages can be improved upon. I've added my try below, but feel free to ignore it!

integreat_cms/cms/views/dashboard/dashboard_view.py
jarlhengstmengel: The approach of @charludo seems to solve this problem, since the same approach seems to solve a similar problem in refactoring the page tree #3130

integreat_cms/cms/views/dashboard/dashboard_view.py
JoeyStk: Thank you a lot :pray:

JoeyStk:
@jarlhengstmengel I think this PR is ready for review again :)

JoeyStk:
There are still a few inconsistencies in the data. Therefore @jarlhengstmengel and I will do some more testing and that's why I converted this to a draft PR

JoeyStk:
@jarlhengstmengel I refactored the calculation for the translation coverage and now I get the correct calculations for the translation_count. Could you please check if it also works for you? :)

jarlhengstmengel (approved):
Thank you very much for the work! Now the count is working as expected!

Issue #2854, opened 2024-06-17
Add "keep direction" functionality to do not translate function


osmers:
### Motivation
With the new phone number format, phone numbers are written with spaces between them.
![image](https://github.com/digitalfabrik/integreat-cms/assets/64901227/3fae36c4-8d78-47cf-85ae-5904d1abcff6)
This means that when translating this number into rtl languages, their order gets mixed up: 
![image](https://github.com/digitalfabrik/integreat-cms/assets/64901227/3563a176-7e47-4c5f-b9a0-d19f038296dd)

### Proposed Solution
Extend the do not translate function with the function that texts keep their direction: e.g. all texts that are marked as do not translate are always ltr

### Alternatives
Add an extra button if the municipalities want to keep direction and keep the functionality seperate from do not translate (not recommended bcs it adds extra difficulty and possibilites of making mistakes)

### User Story
As a user of the app I want to read the phone number in the correct order, even when using the app in rtl languages so that I always dial the correct number. 
As a CMS user I want no extra work when using formatted phone numbers and translating content so that I can keep the time I need to work on Integreat as low as possible.

### Additional Context
Mattermost Thread: https://chat.tuerantuer.org/digitalfabrik/pl/zu3x4bscy7bcfgen8p7fgngacc
@svenseeberg already tested the function and it seems to work, even between two rtl texts...
![image (5)](https://github.com/digitalfabrik/integreat-cms/assets/64901227/4f8ff345-8072-436a-8a4b-51068c6c4112)

### Design Requirements
None since it will be an invisible functionality.

svenseeberg:
The `<span>` element can get a `dir="ltr"` to achieve that. AFAICT it should not be harmful in any way, as by definition the content is not translated (and should therefore also remain ltr).

JoeyStk:
I couldn't get it to work. I added `direction:ltr` to the `do-not-translate span`. That didn't work. Also adding it to the telephone link didn't work. 

This is because the parent element `<p style="direction: rtl; ;text-align: right;"><a class="link-external" style="direction: ltr;" href="tel:+491111111" data-bcup-haslogintext="no"><span class="notranslate" style="direction: ltr;" translate="no">+49 (0) 1111111</span></a></p>` overwrites it with rtl again. I tried to manipulate this and set the closest p to ltr, how a. it was still overwritten and b. even if it worked, then we would have problems with the other elements within that parent element. 

Any further ideas are welcome! @svenseeberg

[...]

PR #3135, opened 2024-10-16
Actually fix notranslate ltr direction

PeterNerlich:
### Short description

We discovered that #2873 fixed only the in-editor representation on CMS side, not on the app side. This is because we employed CSS to do that, which only applied on CMS side. We currently don't have any CSS we serve to be injected wherever HTML is embedded on the app side.

### Proposed changes

- Make `notranslate` button and telephone numbers render `notranslate` spans with `dir="ltr"` attribute
- Add normalization to `clean_content()` for page translations to ensure all `notranslate` spans have
  - the `notranslate` class
  - the `translate="no"` attribute
  - the `dir="ltr"` attribute

### Side effects

- If we have rtl languages as source language in the future, this will need more complex logic to dynamically adjust and work the other way around
- We have to fix all currently online page translations once

### Resolved issues

Fixes: #2854


charludo (approved):
Looks good to me! 🎉 

From the discussion in MM I gathered that this does not fix the DeepL related bug though, right? Does it still make sense to merge?

PeterNerlich:
> From the discussion in MM I gathered that this does not fix the DeepL related bug though, right?

I seems the DeepL bug with moving spans marked with `translate="no"` is entirely unrelated to and unaffected by the presence or absence of `dir="ltr"`

JoeyStk:
>  Does it still make sense to merge?

I'm wondering about the same thing :)

PeterNerlich:
I feel like there is some confusion about what this PR does and what problems exist outside of that, so I'll do a brain dump of my understanding.

- There is a rendering/layouting issue with multi-word content in `span`s marked with `translate="no"` wherever the source language is *left to right* but the target language is *right to left*.
  This is an oversight on our part and not an issue in browsers or anything, because the browser has no reason to care about `translate="no"`, it doesn't translate anything, this is just another `span` that gets rendered, and we told it that this language is *right to left* and all inline elements should be layouted that way.
    - This PR addresses this problem, telling the browser that these `span`s should always be layouted *left to right*.
- There is a bug in DeepL when translating HTML content. Elements can be marked with `translate="no"` to convey to DeepL that this content should be kept as is, but apparently sometimes the element gets put at the very start of the sentence or paragraph, not in a place that makes sense and retains the meaning.
  This appears to be different between languages and I suspect has to be trained into the translation models for every language separately, which leads me to expect that this will not be a "quick fix" for DeepL.
  <details><summary>Example</summary>
  
  image
  image
  
  </details>
  
    - This is *not* addressed by this PR and does not influence the effectiveness of this PR.
    
> Does it still make sense to merge?

So that's my impression of the situation, and if it is a sufficiently close approximation of reality, then **yes, it does make sense to merge**. However, in the process of the confusion and discussions on Mattermost we kind of established that there are larger problems with DeepL that likely won't be fixed that quickly and that we cannot do anything about ourselves.

JoeyStk (approved):
Thank you! LGTM!

Issue #3109, opened 2024-10-04
[2024-11-07] Imported all-day event shows wrong date 


JoeyStk:
### Describe the Bug

When importing an all-day event it shows the wrong end date. The end time is set to 1:00 the following date.

### Steps to Reproduce

1. Go to Nextcloud calendar and create event "Integrationskonferenz" from September 17th to 20th. 
2. Double check that Nextcloud displays it correctly.
3. Import the Nextcloud calendar using the iCal subscription.
4. See in imported events that the events ends on the 21st, not the 20th.

### Expected Behavior

Event is correctly imported

### Actual Behavior

The event is not correctly imported. It shows the wrong end date and time.

### Additional Information

None

juliankehne:
Tested successfully!

PR #3136, opened 2024-10-18
Fix end date of all-day events in iCal imports

charludo:
### Proposed changes

- simply subtract a day from the imported date if the event is all-day

### Side effects

Even though it is probably not likely to occur "in the wild", we previously had the possibility of *only* either the start or end date being a `datetime.date` instead of a `datetime.datetime`.

This could lead to either all-day events with and end time, or non-all-day events with no end time.
I opted to change the logic slightly, and now events are all-day as soon as either the start or end has no time.

### Resolved issues

Fixes: #3109

### DIFF
--- a/integreat_cms/cms/utils/external_calendar_utils.py
+++ b/integreat_cms/cms/utils/external_calendar_utils.py
@@ -51,6 +51,7 @@ class IcalEventData:
         :param logger: The logger to use
         :return: An instance of IcalEventData
         """
+        # pylint: disable=too-many-locals
         event_id = event.decoded("UID").decode("utf-8")
         title = event.decoded("SUMMARY").decode("utf-8")
         content = clean_content(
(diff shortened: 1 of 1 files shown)


integreat_cms/cms/utils/external_calendar_utils.py
```diff
@@ -35,6 +35,7 @@ class IcalEventData:
     categories: list[str]
     external_calendar_id: int
 
+    # pylint: disable=too-many-locals
```
david-venhoff, line 38: Apparently we have been doing this wrong the entire time :see_no_evil: 
More context: https://github.com/digitalfabrik/integreat-cms/pull/3075#discussion_r1801247908

Github won't let me make a suggestion, but the fix would be to move this line to the first line of the function

david-venhoff (approved):
Thanks!

Looks good to me

integreat_cms/cms/utils/external_calendar_utils.py
```diff
@@ -73,15 +74,25 @@ def from_ical_event(
...
 
+        is_all_day = not (
+            isinstance(start, datetime.datetime) and isinstance(end, datetime.datetime)
+        )
+        if is_all_day:
+            start_date, start_time = start, None
+            end_date, end_time = end - datetime.timedelta(days=1), None
+        else:
+            start_date, start_time = start.date(), start.time()
+            end_date, end_time = end.date(), end.time()
```
JoeyStk, line 85: Do you think we can put this into it's own method to make the method a bit smaller? :)

JoeyStk (approved):
Nice, thank you very much. That fixes the issue! :)
I only have one small nit :)

integreat_cms/cms/utils/external_calendar_utils.py
charludo: I originally wanted to say "yes, sure!", but - that function would either have to take `start` and `end` as input and return `start_date`, `start_time`, `end_date`, `end_time`, `is_all_day` as a giant 5-tuple, or be at least three different functions.

Not sure if that is wirth it? 😅  If you insist I'll implement it, I'm just a bit ambivalent here. Small functions are nice, but I am not sure it would help with readability.

JoeyStk: Oops, I didn't realize I was blocking this PR :/ To be honest, at least for now, I don't have a clear opinion or idea. To me this is another function that keeps growing and eventually will be hard to maintain. However I also see your point and you're absolutely right. So I don't know :shrug:

charludo: > To me this is another function that keeps growing and eventually will be hard to maintain.

That's absolutely valid. The only thing I can really give as a counterargument is that it's not a short function, but it's also not a complex function - no loops, and this is the only if/else branch. *So far* it's OK to maintain, I think :D

But I'll leave it up to you.  If the above is not convincing, LMK, and I'll use this PR to try and refactor the function.

JoeyStk: > So far it's OK to maintain, I think :D
Yeah, on second thought I agree. I think we can merge this PR :)

Issue #2991, opened 2024-08-13
Add checkbox in region creation form that allows to not copy languages and translations of the source region


osmers:
### Motivation
Add a checkbox that allows to not copy the language nodes and translations upon region creation using the copy functionality, because they are often not needed.

### Proposed Solution
Add a default-checked checkbox that upon unchecking prevents the languages nodes and translations of the source region to be copied to the new region.

### Alternatives
None

### User Story
As a Service Team memeber I need the translations of copied regions in the Malte App in order to fulfill the requirements of the Malteser, but I don't want that functionality always.

david-venhoff:
This would also be a workaround of #3048

juliankehne:
@charludo While testing the checkbox I had one case where not all pages were copied.

I tried to reproduce it, but wasn't able to:
- New checkbox was checked (to include translations)
- Region that pages were copied from: Testumgebung
https://cms-test.integreat-app.de/regions/2991-1/edit/
https://cms-test.integreat-app.de/2991-1/pages/de/

Is that the known issue of the process timing out when copying pages with translations or something else?

osmers:
I assume it's the same problem we've been having with the old process - it simply times out while trying to copy everything. For Integreat generally we will need the checkbox unchecked so that no translations get copied. Nevertheless, for Malte we do need the translations, so if the problem persists, especially within Malte, that content only gets partially copied, we need to investigate further :)

PR #3137, opened 2024-10-18
Add option to skip copying language tree and translations when duplicating a region

charludo:
### Proposed changes

- add chekcbox to "new region" form
- if a region is duplicated, and the checkbox is unchecked, only copy the root nodes of the language tree
- also only copy the translations of the languages that were copied, i.e. only of the root language tree nodes

### Side effects

- probably slightly worse runtime when duplicating a region WITH all translations/languages

### Resolved issues

Fixes: #2991

### DIFF
--- a/integreat_cms/cms/forms/regions/region_form.py
+++ b/integreat_cms/cms/forms/regions/region_form.py
@@ -114,6 +114,15 @@ class RegionForm(CustomModelForm):
         ),
     )
 
+    duplication_keep_translations = forms.BooleanField(
+        required=False,
+        label=_("Copy languages and content translations"),
+        help_text=_(
+            "Disable to skip copying of the language tree and all content translations."
+        ),
+        initial=True,
+    )
+
     duplication_pbo_behavior = forms.ChoiceField(
         choices=duplicate_pbo_behaviors.CHOICES,
         initial=duplicate_pbo_behaviors.ACTIVATE,
@@ -271,6 +280,7 @@ class RegionForm(CustomModelForm):
         if duplicate_region:
             source_region = self.cleaned_data["duplicated_region"]
             keep_status = self.cleaned_data["duplication_keep_status"]
+            keep_translations = self.cleaned_data["duplication_keep_translations"]
 
             # Determine offers to force activate or to skip when cloning pages
             required_offers = OfferTemplate.objects.filter(pages__region=source_region)
@@ -292,7 +302,9 @@ class RegionForm(CustomModelForm):
 
             # Duplicate language tree
             logger.info("Duplicating language tree of %r to %r", source_region, region)
-            duplicate_language_tree(source_region, region)
+            duplicate_language_tree(
+                source_region, region, only_root=not keep_translations
+            )
             # Disable linkcheck listeners to prevent links to be created for outdated versions
             with disable_listeners():
                 # Duplicate pages
@@ -302,6 +314,7 @@ class RegionForm(CustomModelForm):
                     region,
                     keep_status=keep_status,
                     offers_to_discard=offers_to_discard,
+                    only_root=not keep_translations,
                 )
                 # Duplicate Imprint
                 if source_region.imprint:
@@ -655,6 +668,7 @@ def duplicate_language_tree(
     source_parent: LanguageTreeNode | None = None,
     target_parent: LanguageTreeNode | None = None,
     logging_prefix: str = "",
+    only_root: bool = False,
 ) -> None:
     """
     Function to duplicate the language tree of one region to another.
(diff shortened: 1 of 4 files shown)


tests/cms/test_duplicate_regions.py
```diff
@@ -178,3 +178,91 @@ def test_duplicate_regions(
...
+                "region",
+                "parent",
+                "api_token",
+                "authors",
+                "editors",
+            ],
+        )
+        assert source_page_dict == target_page_dict
+
+        source_pages_translations = source_page.translations.filter(
```
JoeyStk, line 261:
```suggestion
        source_page_translations = source_page.translations.filter(
```
Nit: I think you meant source_page_translations - or did you mean something else? :)

tests/cms/test_duplicate_regions.py
```diff
@@ -178,3 +178,91 @@ def test_duplicate_regions(
...
+            language=source_language_root
+        )
+        target_pages_translations = target_page.translations.filter(
+            last_updated__lt=before_cloning
+        )
+        assert len(source_pages_translations) == len(target_pages_translations)
```
JoeyStk, line 267: I'm sorry, but I don't understand this line. Why do we want the length of source_pages_translations and of target_pages_translations to be equal? :thinking:

integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -729,7 +744,7 @@ def duplicate_language_tree(
             )
 
 
-# pylint: disable=too-many-locals
+# pylint: disable=too-many-locals,too-many-arguments
```
JoeyStk, line 747: I think it makes sense to create a dataclass that bundes some of the arguments, so we don't need to disable this line anymore :)

integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -832,6 +849,7 @@ def duplicate_pages(
                 row_logging_prefix,
                 keep_status,
                 offers_to_discard,
+                only_root,
             )
```
JoeyStk, line 854: To me, 100 lines are a bit too long for one method. Do you think there is a quick way to extract some parts of this method into its own method - or do we instead need a refactoring ticket?

JoeyStk (commented):
Thank you very much. To me this looks good so far. The main functionality worked for me. However I haven't tested it with a bulk of data. Maybe this would be something for the second reviewer to consider :)

JoeyStk (approved):
As I only have a few nits, I'm going to approve this PR already :9

tests/cms/test_duplicate_regions.py
charludo: Yeah, fair. The Skeleton of the test is copied from an existing one. I kept the variable names, will change in both tests!

tests/cms/test_duplicate_regions.py
charludo: The source_pages_translations are filtered to only contain languages that are roots in the language tree (and thus were copied). I will change the variable name to make it clearer.

integreat_cms/cms/forms/regions/region_form.py
charludo: But does that not *increase* cognitive overhead?

I get that sometimes, a lot of arguments can indicate that a function simply does too much, but that is not the case here - the parameters are just required.

Since this is (afaik) the only place we need this combination of parameters, extracting them into a dataclass would just mean you additionally need to look up what the dataclass contains, instead of just having the info presented to you where you expect it, in the function signature.

integreat_cms/cms/forms/regions/region_form.py
charludo: Neither, I think. Of the 105 lines, 10 are the function signature, 15 are just calls to the logger, and 31 are comments. (If anything, I would like to remove the in-line coments, they are pretty clear examples of superfluous comments stating WHAT is going on, instead of WHY, despite it being obvious.)

The actual logic of the function is really well written IMO (not talking about my changes, just the "original function" 😄).
The only thing that can really be extracted - the copying of the translations, in addition to the page itself - is already a separate function.

integreat_cms/cms/forms/regions/region_form.py
JoeyStk: Yes, I think you're right :)

integreat_cms/cms/forms/regions/region_form.py
JoeyStk: >  (If anything, I would like to remove the in-line coments, they are pretty clear examples of superfluous comments stating WHAT is going on, instead of WHY, despite it being obvious.)

In my eyes you can do so :)

> The only thing that can really be extracted - the copying of the translations, in addition to the page itself - is already a separate function.

Alright. I just wanted to have some more time to think about it, but I think you're right :) I think we can merge this PR :)

Issue #3141, opened 2024-10-21
[2024-11-07] Fix issues with contact form & list view


JoeyStk:
### Describe the Bug

This is a list of a few issues that we had when testing the contact form & list view

- [x] Put location of contact in contact list view in the first column 
- [x] Adjust translation for location
- [x] Fix typo "ihr" to "Ihr"
- [x] Fix help text shouldn't be bold
- [x] Fix help text disappears when saving contact
- [x] Make a few changes to str function for it to be a bit more speaking
- [x] Rename title to point_of_contact_for 

### Additional Information

This issues came up in the first round of testing the contact feature

juliankehne:
@JoeyStk

JoeyStk:
@juliankehne 

> Adjust translation for location

The row which links to location was called "Name des Veranstaltungsortes", which is untrue. Therefore the name should be changed to "Name des verlinkten Ortes". 

> Fix help text shouldn't be bold

If you create a new contact, you get an informational text saying "Wählen Sie einen Ort aus, der für Ihren Kontakt verwendet werden soll oder erstellen Sie einen neuen Ort. Nur veröffentlichte Orte können verwendet werden. " This used to be bold. Now it should be regular font weight :)

> Fix help text disappears when saving contact

The text "Wählen Sie einen Ort aus, der für Ihren Kontakt verwendet werden soll oder erstellen Sie einen neuen Ort. Nur veröffentlichte Orte können verwendet werden. " used to disappear, once you clicked on "Save". This should not be the case anymore. The text itself changes, but there should be some kind of information afterwards :)

> Make a few changes to str function for it to be a bit more speaking

We changed the logic of the display of Contact. When you save a contact now it should say "Kontakt für "Name of location" mit ...". What's followed after the "mit" is decided as followed: If the field "Ansprechperson für" is filled out, it should say "mit Ansprechperson für: Integration". If this field is not filled out, it should fall back on the name, if that's not filled out it should go to the email, then the telephone number and lastly to the website URL. 

Thank you a lot for testing! If anything else comes up, please don't hesitate to ask :)

juliankehne:
@JoeyStk Thanks, all 7 changes tested succesfully now!

PR #3142, opened 2024-10-21
Fixes for contact

JoeyStk:
### Short description

This PR fixes a few issues we had with contact

### Proposed changes

- Put location of contact in contact list view in the first column
- Adjust translation for location
- Fix typo "ihr" to "Ihr" in contact form
- Fix help text shouldn't be bold
- Fix help text disappears when saving contact
- Make a few changes to str function for it to be a bit more speaking
- Rename title to point_of_contact_for

These changes should fix all issues we noticed during our first round of testing 

### Side effects

These were quite a few changes, it's possible that we overlooked something

### Resolved issues

Fixes: #3141

### DIFF
--- a/integreat_cms/cms/models/contact/contact.py
+++ b/integreat_cms/cms/models/contact/contact.py
@@ -14,7 +14,9 @@ class Contact(AbstractBaseModel):
     Data model representing a contact
     """
 
-    title = models.CharField(max_length=200, blank=True, verbose_name=_("title"))
+    point_of_contact_for = models.CharField(
+        max_length=200, blank=True, verbose_name=_("point of contact for")
+    )
     name = models.CharField(max_length=200, blank=True, verbose_name=_("name"))
     location = models.ForeignKey(
         POI,
@@ -61,7 +63,31 @@ class Contact(AbstractBaseModel):
 
         :return: A readable string representation of the contact
         """
-        return f"{self.title} {self.name}"
+        location_name = str(self.location)
+        additional_attribute = self.get_additional_attribute()
+        return " ".join(part for part in [location_name, additional_attribute] if part)
+
+    def get_additional_attribute(self) -> str:
+        """
+        This function determines which string is shown for the contact
+        """
+
+        if self.point_of_contact_for:
+            return _("with point of contact for: {}").format(self.point_of_contact_for)
+
+        if self.name:
+            return _("with name: {}").format(self.name)
+
+        if self.email:
+            return _("with email: {}").format(self.email)
+
+        if self.phone_number:
+            return _("with phone number: {}").format(self.phone_number)
+
+        if self.website:
+            return _("with website: {}").format(self.website)
+
+        return ""
 
     def get_repr(self) -> str:
         """
@@ -70,7 +96,7 @@ class Contact(AbstractBaseModel):
 
         :return: The canonical string representation of the contact
         """
-        return f"<Contact (id: {self.id}, title: {self.title}, name: {self.name}, region: {self.region.slug})>"
+        return f"<Contact (id: {self.id}, point of contact for: {self.point_of_contact_for}, name: {self.name}, region: {self.region.slug})>"
 
     def archive(self) -> None:
         """
@@ -92,7 +118,7 @@ class Contact(AbstractBaseModel):
         """
         # In order to create a new object set pk to None
         self.pk = None
-        self.title = self.title + " " + _("(Copy)")
+        self.point_of_contact_for = self.point_of_contact_for + " " + _("(Copy)")
         self.save()
 
     class Meta:
@@ -105,21 +131,21 @@ class Contact(AbstractBaseModel):
         constraints = [
             models.UniqueConstraint(
                 "location",
-                condition=Q(title=""),
-                name="contact_singular_empty_title_per_location",
+                condition=Q(point_of_contact_for=""),
+                name="contact_singular_empty_point_of_contact_per_location",
                 violation_error_message=_(
-                    "Only one contact per location can have an empty title."
+                    "Only one contact per location can have an empty point of contact."
                 ),
             ),
             models.CheckConstraint(
-                check=~Q(title="")
+                check=~Q(point_of_contact_for="")
                 | ~Q(name="")
                 | ~Q(email="")
                 | ~Q(phone_number="")
                 | ~Q(website=""),
                 name="contact_non_empty",
                 violation_error_message=_(
-                    "One of the following fields must be filled: title, name, e-mail, phone number, website."
+                    "One of the following fields must be filled: point of contact for, name, e-mail, phone number, website."
                 ),
             ),
         ]
--- a/integreat_cms/cms/fixtures/test_data.json
+++ b/integreat_cms/cms/fixtures/test_data.json
@@ -956,7 +956,7 @@
     "model": "cms.contact",
     "pk": 1,
     "fields": {
-      "title": "Integrationsbeauftragte",
+      "point_of_contact_for": "Integrationsbeauftragte",
       "name": "Martina Musterfrau",
       "location": 6,
       "email": "martina-musterfrau@example.com",
@@ -971,7 +971,7 @@
     "model": "cms.contact",
     "pk": 2,
     "fields": {
-      "title": "Integrationsberaterin",
+      "point_of_contact_for": "Integrationsberaterin",
       "name": "Melanie Musterfrau",
       "location": 6,
       "email": "melanie-musterfrau@example.com",
@@ -986,7 +986,7 @@
     "model": "cms.contact",
     "pk": 3,
     "fields": {
-      "title": "Integrationsbeauftragte",
+      "point_of_contact_for": "Integrationsbeauftragte",
       "name": "Mariana Musterfrau",
       "location": 6,
       "email": "mariana-musterfrau@example.com",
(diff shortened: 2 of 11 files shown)


integreat_cms/cms/models/contact/contact.py
```diff
@@ -61,7 +63,19 @@ def __str__(self) -> str:
...
+            else ""
+        )
+        name = f"{self.name.strip()} " if self.name and self.name.strip() else ""
+        details = [
+            detail
+            for x in ["email", "phone_number", "website"]
+            if (detail := self.get(x)) and detail.strip()
+        ]
+        details_str = f"({', '.join(details)})" if details else ""
+        return f"{point_of_contact_for}{name}{details_str}"
```
david-venhoff, line 78: Wouldn't it be better to make sure in the form that the fields are already stripped, instead of having to strip them every time they are used?

integreat_cms/cms/models/contact/contact.py
```diff
@@ -61,7 +63,19 @@ def __str__(self) -> str:
...
+            f"{self.point_of_contact_for.strip()}: "
+            if self.point_of_contact_for and self.point_of_contact_for.strip()
+            else ""
+        )
+        name = f"{self.name.strip()} " if self.name and self.name.strip() else ""
+        details = [
+            detail
+            for x in ["email", "phone_number", "website"]
+            if (detail := self.get(x)) and detail.strip()
+        ]
```
david-venhoff, line 76:
```suggestion
        details = [
            detail
            for detail in [self.email, self.phone_number, self.website]
            if detail
        ]
```

Did you mean something like that? Because there exists no `get` method

david-venhoff (changes requested):
Thanks, looks mostly good to me.

 I found two problems with the new `__str__` method
* It can be a bit long:
image
Maybe we should omit the details if name or point of contact for are defined?

* It can also be a bit non-descriptive:
image
For this case, it could make sense to also print the location maybe

integreat_cms/cms/models/contact/contact.py
PeterNerlich: yes, you're absolutely right

PeterNerlich:
> I found two problems with the new `__str__` method
> 
> * It can be a bit long: […]
>   Maybe we should omit the details if name or point of contact for are defined?
> * It can also be a bit non-descriptive: […]
>   For this case, it could make sense to also print the location maybe

Originally, this was my draft for a sort of all-in-one form of a contact as part of #3124 and I just quickly ported it to python and suggested it for this as well. I agree with your points, but I worry that then it might be not enough to uniquely identify the contact in some situations. But then again, I'm not sure how often we should expect such collisions to occur from our user base and if it does, whether it will be a source of confusion at all

integreat_cms/cms/models/contact/contact.py
JoeyStk: Is there still something that should be done for this? :)

JoeyStk:
I added a new commit that changes the __str__ function. It was agreed with @osmers that the message should show something along this pattern: 

"Contact for [NAME_OF_LOCATION] with [ATTRIBUTE]: [ATTRIBUTE] was saved". The attribute could be either point_of_contact_for, name, email, telephone_number or website (in this order) but always just one. In case for test-location which has a point_of_contact_for as "integration specialist" it would be: "Contact for test-location with point of contact: Integration specialist was saved". 

Please let me know if you either there is an issue with this pattern in general or if there's something off with it's implementation :)

integreat_cms/cms/models/contact/contact.py
```diff
@@ -63,19 +63,27 @@ def __str__(self) -> str:
...
+        if self.email:
+            return _("email: %s") % self.email
+
+        if self.phone_number:
+            return _("phone number: %s") % self.phone_number
+
+        if self.website:
+            return _("website: %s") % self.website
+
+        return ""
```
PeterNerlich, line 86: This might make the message show `Contact for [NAME_OF_LOCATION] with  was saved` when this case is hit.
Though it *is* a degenerate case that we currently don't anticipate in the system, as we constrain the model to have at least one field set. So maybe it's fine to just leave it, but on the other hand dropping the ` with ` part seems easy enough

JoeyStk: I removed the with from the String now :)

PeterNerlich: Tiny nitpick, if the detail happens to be empty, the result of `__str__()` has trailing whitespace. I'm unsure how to account for that cleanly and without cluttering the code *(using `return _(" with name: %s") % self.name` would put it in the translations where it does not belong, but I'm not sure shenanigans like `return f' {_("with name: %s") % self.name}` is much better either)*

integreat_cms/cms/models/contact/contact.py
david-venhoff: The form still needs to strip the fields. I suspect that this may be quite a common issue that also affects other models, so you could also open an issue, instead of implementing that in this pr I guess

integreat_cms/cms/models/contact/contact.py
```diff
@@ -61,7 +63,27 @@ def __str__(self) -> str:
 
         :return: A readable string representation of the contact
         """
-        return f"{self.title} {self.name}"
+        location_name = str(self.location)
+        additional_attribute = self.get_additional_attribute()
+        return _("%s %s") % (location_name, additional_attribute)
```
david-venhoff, line 68: (Probably a nit)
I know the django docs use `%` based formatting, but in integreat_cms we have only used formatting with the `.format` method. There are some problems with `%`-based formatting, which are described here https://peps.python.org/pep-3101/
Even though those problems do not apply to the occurrences here, I would still prefer to use the other formatting syntax for consistency with the rest of the codebase

PeterNerlich: Wait, now that I think about it, does `_("%s %s") % (…)` even make sense at all? We're trying to translate a string that has two spots for values to be inserted. But while it is usually needed for when the sentence structure is switched around in another language so the value has to end up in a different place, here the translation will only be different from the source if we happen to come across a language that does not have spaces…?

integreat_cms/cms/models/contact/contact.py
JoeyStk: Do you have a preference out of these three options?
1. Leave it as it is, as it's only a theoretical possibility and since the detail isn't allowed to be empty anyways would never happen in reality
2. But it in the translation (which is not really clean)
3. Do stuff like return f' {_("with name: %s") % self.name}

JoeyStk: (for now I changed it to option 2)

integreat_cms/cms/models/contact/contact.py
JoeyStk: Is the current implementation solving this issue or is it still the same? :7

PeterNerlich: I'm very much unsure whether this should be considered readable code, but this would be the most exact way to implement this that I can think of:
```python
return " ".join(part for part in [location_name, additional_attribute])
```

integreat_cms/cms/models/contact/contact.py
PeterNerlich: Personally, I'm itching to do this *(here trying to not have an extra f string in every if branch)* but you guys need to decide whether that's too much visual clutter for one space *(or alternatively the `" ".join()` solution)*
```python
    def get_additional_attribute(self) -> str:
        string = None

        if self.point_of_contact_for:
            string = _("with point of contact for: {}").format(self.point_of_contact_for)

        if self.name:
            s = _("with name: {}").format(self.name)

        if self.email:
            s = _("with email: {}").format(self.email)

        if self.phone_number:
            s = _("with phone number: {}").format(self.phone_number)

        if self.website:
            s = _("with website: {}").format(self.website)

        if s:
            return f" {s}"

        return ""
```

JoeyStk: I like this solution

integreat_cms/cms/models/contact/contact.py
JoeyStk: I think this needs some internal discussion about how we want to go about it. There are multiple ways to go about this and I think this needs to be decided in the entire team. Therefore I will open up a new issue for this

integreat_cms/cms/models/contact/contact.py
```diff
@@ -61,7 +63,37 @@ def __str__(self) -> str:
...
+        if self.phone_number:
+            string = _("with phone number: {}").format(self.phone_number)
+
+        if self.website:
+            string = _("with website: {}").format(self.website)
+
+        if string:
+            return f" {string}"
+
+        return ""
```
david-venhoff, line 96:
```suggestion
        return string
```
The `" ".join(…)` already inserts a space between the location name and the additional attribute, so adding another space here causes a double space

david-venhoff (approved):

integreat_cms/cms/models/contact/contact.py
MizukiTemma, line 96: I'm not sure whether I understood the agreement with @osmars correctly, but we want to show "Contact for test-location with **point of contact: Integration specialist**" rather than "Contact for test-location with **website: https://integreat-app.de/**" if both point of contact and website are given, right? Currently the later is shown.

JoeyStk: Very true

MizukiTemma (approved):
🎉 

Just don't forget to squash the commits 😉

Issue #3045, opened 2024-09-13
Implement Glossary for DeepL in CMS


osmers:
### Motivation
Christine added a Glossary for DeepL and the API call needs to be changed to call the glossary as well.

### User Story
As a CMS user I want certain words to always be translated the same way so that there is no confusion for non-native speakers.

### Additional Context

https://chat.tuerantuer.org/digitalfabrik/pl/p36gwikmgjrbfe3x6asj5ge11r

timobrembeck:
Is this a duplicate of #1346?

david-venhoff:
> Is this a duplicate of https://github.com/digitalfabrik/integreat-cms/issues/1346?

This issue is only about modifying our api calls to deepl to tell it to use the already existing glossary.
The linked issue is about implementing a glossary managed by the cms. I have edited it to make this more clear now

ChristineBruckner:
DeepL Glossaries created from Integreat memoQ glossary
&amp; uploaded to DeepL API - May 2024, Christine
Bruckner


Glossary language combination | Glossary ID
DE-EN_2 | dba493bc-b18e-4173-ae04-664dfe1de0c9
DE-ES_2 | 588708de-eb3e-4f79-b8b5-bc8dbe19f346
DE-FR_2 | d0de3b7a-5986-4613-af00-3a42b476a6c2
DE-IT_2 | 303036a5-78e7-4550-b8eb-0bc7db446f82
DE-PL_2 | f14b3b4a-e9bb-4325-b1ca-613466e056e1
DE-RU_2 | 44930f57-f01b-4c15-a88e-8adf45cab5be

PR #3143, opened 2024-10-21
Add support for deepl glossaries

david-venhoff:
### Short description

This pr adds basic support for deepl glossaries.
If enabled, every glossary defined at deepl will be used (undefined behavior if multiple glossaries for the same source to target language pair exist)

For now, the feature is disabled by default.
Note to self: Enable the setting on the test server when merged

### How to test

- Set `DEEPL_GLOSSARIES_ENABLED` to `True` in the settings
- Configure deepl by setting the integreat deepl password to the free deepl api password from passbolt.
- Translate a page that contains glossary words from german to english 
- Check that the english translation uses the corresponding glossary entry

You can print out all available glossary entries with `print(self.translator.get_glossary_entries(glossary))`

~~Note that at the time where I am writing this the de-en dictionary is accidentally reversed (so when translating from de to en, deepl replaces english words in the source translation with german words in the target translation)~~
This is fixed now

### Proposed changes

- Load available glossaries on startup
- Use matching glossary, if it exists

### Side effects

None, since the feature is disable by default

### Resolved issues

Fixes: #3045

### DIFF
--- a/integreat_cms/core/settings.py
+++ b/integreat_cms/core/settings.py
@@ -924,6 +924,11 @@ DEEPL_API_URL: Final[str | None] = os.environ.get("INTEGREAT_CMS_DEEPL_API_URL")
 #: Authentication token for the DeepL API. If not set, automatic translations via DeepL are disabled
 DEEPL_AUTH_KEY: str | None = os.environ.get("INTEGREAT_CMS_DEEPL_AUTH_KEY")
 
+#: Whether to enable deepl glossaries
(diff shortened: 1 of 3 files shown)


MizukiTemma (approved):
Nice new feature 😻 works as expected!


Maybe rebasing before the second review?

david-venhoff:
Done

JoeyStk (approved):
Thank you very much! Looks good! :)

:rocket:

Issue #3150, opened 2024-10-23
UI/UX - In Contacts and Events the location addresses (if available) have some redundant empty space


juliankehne:
### Describe the Bug
Theres a bit of redundant empty space inbetween the Address header & content when editing a Contact or Event.

Event:
![Event](https://github.com/user-attachments/assets/1c5ff18a-07a1-4f0c-b6f8-59ce0be6c19f)

Contact:
![Contact](https://github.com/user-attachments/assets/34b2a7d5-76cf-49ff-ad7e-96fbb3ca6a85)

### Steps to Reproduce

1. Go to any contact or an event with a location linked
3. See empty space inbetween

### Expected Behavior
Less empty space inbetween the Header and adress content

![2024-10-23 15_46_28-Integreat Editorial System — Mozilla Firefox](https://github.com/user-attachments/assets/5c596cb4-c225-44cd-b6a8-02511df8c8bd)

### Actual Behavior
A few lines of empty space inbetween Header and Address content (see screenshots above)

### Additional Information

When creating a new contact and choosing the location the white space is not seen:

![2024-10-23 15_53_26-Integreat Editorial System — Mozilla Firefox](https://github.com/user-attachments/assets/30659fcc-c1aa-4b25-a114-a3db743eaa8b)

juliankehne:
Tested successfully!

PR #3153, opened 2024-10-25
Fix spacing in poi box

JoeyStk:
### Short description

This PR attempts to fix the spacing issues in poi box.

### Proposed changes

- remove class `whitespace-pre-line`
- add `<br>` instead
- Avoid constructing the poi box client side. Instead, always render it on the server so that we don't have to maintain two versions of the same thing

### Side effects

I don't know

### Resolved issues

Fixes: #3150 

### DIFF
--- a/integreat_cms/cms/templates/_poi_query_result.html
+++ b/integreat_cms/cms/templates/_poi_query_result.html
@@ -13,15 +13,8 @@
             {% for poi in poi_query_result %}
                 <option class="option-existing-poi hover:bg-gray-200 cursor-pointer"
                         title="{{ poi.address }}, {{ poi.postcode }} {{ poi.city }}"
-                        data-poi-title="{{ poi|poi_translation_title:current_language }}"
-                        data-poi-id="{{ poi.id }}"
-                        data-poi-address="{{ poi.address }}"
-                        data-poi-postcode="{{ poi.postcode }}"
-                        data-poi-city="{{ poi.city }}"
-                        data-poi-country="{{ poi.country }}"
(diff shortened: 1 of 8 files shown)


david-venhoff (changes requested):
Unfortunately this causes another problem, because we dynamically modify the contents of this div from js when a user selects a location: https://github.com/digitalfabrik/integreat-cms/blob/824cfad6f560f4983c048e7b55dcba743ec50b8e/integreat_cms/static/src/js/poi_box.ts#L36

You can see this if you create a new contact and select a location there for example. Since the assignment does not include the `<br>` tags, everything gets rendered in a single line now.

There are multiple solutions to this, for example
- Keep the `whitespace-pre-line` and build the address string in python. Then we could simply use something like this: `<div id="poi-address" class="text-lg whitespace-pre-line">{{ poi_address_text }}</div>`
- Keep the current solution and instead of assigning to text content in the js manually build html
- Completely get rid of the part where we build the address string in js and instead render everything on the server. Then `search_poi_ajax` would include the rendered poi addres div. That would also have the advantage that the layout cannot diverge anymore, as there is only a single place where it gets rendered

JoeyStk:
@david-venhoff 
Oh, good catch. I went for option 2. Do you see any downsides with that? :)

david-venhoff (approved):
:grin:

MizukiTemma (approved):
Nice 😻 Thank you so much 👍 

We can probably add a release note, as this change is visible for users 👀 
And don't forget to squash the commits.

Issue #3146, opened 2024-10-23
Event name character limit too big and copying leads to Internal Server Error.


juliankehne:
### Describe the Bug
Definitely a bit of an edge case :) 
After you have copied an event that has the reached the character limit, its possible to give the copy of the event an even longer name as the character limit has changed.
If you then change the event name to its 'new' character limit and copy it again you get an Internal Server Error ( see #3145 )

### Steps to Reproduce

Event A = Original Event
Event B = Copy of Original

1. Create a new event (Event A) with the event name set to the character limit
2. Copy Event A
3. See that (copy) was added behind the name of Event B and no Internal Server Error
4. Edit Event B and set the length of the name to its character limit
5. See that the character limit has increased
6. Save Event B
7. Copy Event B
8. See an Internal Server Error

### Expected Behavior
No Internal Server Error. The 'new' character limit should be less.

### Actual Behavior
The character limit of the copy has increased by a lot. If that name is too long and then the event is copied it results in an Internal Server Error.

### Additional Information
Version history of example in cms-test: https://cms-test.integreat-app.de/testumgebung/events/de/6917/versions/

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

juliankehne:
Tested successfully!

Issue #3145, opened 2024-10-23
Contact form: Copying a Contact with a title that is too long results in Internal Server Error


juliankehne:
### Describe the Bug
When copying an existing contact via the Copy Button the system tries to add " (copy)" without "" to the end of the contact title.
However if the title of the contact is too long the (copy) cannot be added and it returns a Internal Server Error, the contact is not copied.

### Steps to Reproduce

1. Go to 'Contacts'
2. Click on 'Create contact'
3. Set the title to the maximum amount of allowed characters
4. Click 'Create'
5. Go back to 'Contacts'
6. Click the Copy button to copy the created Contact
7. See Error

### Expected Behavior
The contact is copied and " (copy)" should be added behind the title. No Internal Server Error.

### Actual Behavior
Internal Server Error when trying to copy a contact with a title that is too long

### Additional Information
Copying Events handles this by just adding the (copy) behind the event name that is too long.
However: #3146

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

PeterNerlich:
We need to take care that unicode characters might take up more than one "space", and that an implementation that truncates the original title does not split unicode characters in half

juliankehne:
Tested successfully!

PR #3155, opened 2024-10-26
Fix data error when copying long contact

JoeyStk:
### Short description

This PR attempts to fix the `DataError` which occurs when copying a long contact name. This is because appending the suffix exceeds the maximum length of the field.

### Proposed changes

- Try & catch
- In case the `DataError` would occur truncate the pre-existing name and then append the suffix.

### Side effects

- ~~I don't have a finally block so there might be some?~~

### Resolved issues

Fixes: #3145
Fixes: #3146

### DIFF
--- a/integreat_cms/cms/models/abstract_content_translation.py
+++ b/integreat_cms/cms/models/abstract_content_translation.py
@@ -32,6 +32,7 @@ from ..utils.link_utils import fix_content_link_encoding
 from ..utils.round_hix_score import round_hix_score
 from ..utils.translation_utils import gettext_many_lazy as __
 from .abstract_base_model import AbstractBaseModel
+from .fields.truncating_char_field import TruncatingCharField
 from .languages.language import Language
 
 logger = logging.getLogger(__name__)
@@ -43,7 +44,7 @@ class AbstractContentTranslation(AbstractBaseModel):
     Data model representing a translation of some kind of content (e.g. pages or events)
     """
 
-    title = models.CharField(max_length=1024, verbose_name=_("title"))
+    title = TruncatingCharField(max_length=1024, verbose_name=_("title"))
     slug = models.SlugField(
         max_length=1024,
         allow_unicode=True,
@@ -632,7 +633,7 @@ class AbstractContentTranslation(AbstractBaseModel):
 
         :return: A readable string representation of the content translation
         """
-        return self.title
+        return str(self.title)
 
     def get_repr(self) -> str:
         """
--- a/integreat_cms/cms/migrations/0109_custom_truncating_char_field.py
+++ b/integreat_cms/cms/migrations/0109_custom_truncating_char_field.py
@@ -1,0 +1,51 @@
+from django.db import migrations
+
+import integreat_cms.cms.models.fields.truncating_char_field
+
+
+class Migration(migrations.Migration):
+    """
+    Migration file to change the field point_of_contact_for from a CharField to a TruncatingCharField
+    """
+
+    dependencies = [
+        ("cms", "0108_rename_contact_title_to_point_of_contact_for"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="contact",
+            name="point_of_contact_for",
+            field=integreat_cms.cms.models.fields.truncating_char_field.TruncatingCharField(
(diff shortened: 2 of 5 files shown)


david-venhoff (changes requested):
I think this pr should also fix the copy function for events:

https://github.com/digitalfabrik/integreat-cms/blob/b4679a00755aae87168c6a8ad8e8af238e8faf45/integreat_cms/cms/models/events/event.py#L278

In order to avoid code duplication, what do you think about a solution like this: https://stackoverflow.com/a/3460942
(I am not proposing this exact solution, since it appears to be written for python 2 and truncates differently than what we want, but I like the general idea)

[This](https://docs.djangoproject.com/en/dev/howto/custom-model-fields/#how-to-create-custom-model-fields) might also be relevant if you decide to go for this solution

charludo (commented):
LGTM! 🎉

charludo (approved):

JoeyStk:
@david-venhoff Thank you a lot for your idea. I had a first try at it, but it only went okay I think. 
This is the smallest version I found, but I already noticed a few issues with it.
1. There still some strange side effects (such as "(K(Kopie)" or "(C(Copy" which are probably caused by the overlap of the functions)
2. I'm not sure how stable this solution is in regard to translations. If we added a third language besides English and German, wouldn't there be some maintenance needed to be done?
3. With the current implementation we can't stack "(Copy)" on top of each other again. So the copy of the copy is still called "(Copy)" instead of "(Copy) (Copy)".

Do you have ideas for these issues?

I tried them already, but they got really complicated and I'm not sure if it was worth the effort to follow them all the way to the end

david-venhoff:
Why do you add the "copy" text in the `truncating_char_field`? Judging by its name, I would only expect it to truncate a text, so that it does not exceed its maximum length, but not, that it adds a "copy" text to the end.
Something like this seems closer to what we want:
```python
    def get_prep_value(self, value: str) -> str | None:
        value = super().get_prep_value(value)

        if value and len(value) > self.max_length:
            return value[:self.max_length-3] + "..."
        return value
```
Also the point of the truncating char field is that we don't have to worry about it anymore, so you can remove the truncation code from `Contact.copy` :)

> 1. There still some strange side effects (such as "(K(Kopie)" or "(C(Copy" which are probably caused by the overlap of the functions)

Yes, these will go away if you adapt the solution like described above

> 2. I'm not sure how stable this solution is in regard to translations. If we added a third language besides English and German, wouldn't there be some maintenance needed to be done?

Well, we'd need to translate the string `(copy)` into the new language. But we'd have to do that anyways :shrug:

> 3. With the current implementation we can't stack "(Copy)" on top of each other again. So the copy of the copy is still called "(Copy)" instead of "(Copy) (Copy)"

This will also be fixed if you adapt the current implementation like described above

integreat_cms/cms/migrations/0109_alter_contact_point_of_contact_for.py
david-venhoff, line 1: Could you give this migration a more descriptive name?

integreat_cms/cms/models/abstract_content_translation.py
```diff
@@ -632,7 +633,7 @@ def __str__(self) -> str:
 
         :return: A readable string representation of the content translation
         """
-        return self.title
+        return str(self.title)
```
david-venhoff, line 636:
```suggestion
        return self.title
```

self.title is already a string

integreat_cms/cms/models/truncating_char_field.py
david-venhoff, line 1: (Probably a nit) 
I feel a bit uncomfortable about putting this into the models/ directory, but I also don't know what the best place for this file would be. 
What do you think about moving this file to a models/fields/ directory, or something similar?

david-venhoff (approved):
LGTM, I just have some nits

integreat_cms/cms/models/abstract_content_translation.py
JoeyStk: I received an error from pylint without the conversion. I was confused by this too, but didn't see a way around it

integreat_cms/cms/models/truncating_char_field.py
JoeyStk: Yes, sure :)

PR #3156, opened 2024-10-26
Attempt to fix docs

JoeyStk:
### Short description

This PR attempts to fix the `build_documentation` job which keeps failing.

### Proposed changes

- Rename the call to `xhtml`

### Side effects

Very unsure, it's my first time changing the docs

### Resolved issues

Fixes: -


integreat_cms/cms/utils/pdf_utils.py
```diff
@@ -139,8 +139,8 @@ def generate_pdf(
 # pylint: disable=unused-argument
 def link_callback(uri: str, rel: str) -> str | None:
     """
-    According to xhtml2pdf documentation (see :doc:`xhtml2pdf:usage`),
-    this function is necessary for resolving the django static files references.
+    According to the `xhtml2pdf documentation <https://xhtml2pdf.readthedocs.io/en/latest/usage.html>`_,
+    this function is necessary for resolving the Django static files references.
```
PeterNerlich, line 143:
```suggestion
    According to the xhtml2pdf documentation (see `Link callback <https://xhtml2pdf.readthedocs.io/en/latest/reference/python.html#link-callback>`_,
    this function is necessary for resolving the Django static files references.
```

PeterNerlich (changes requested):
I took the opportunity to describe my mental model of the relevant behaviour of sphinx in hopes that you find it helpful and whoever else might happen to stumple upon this

#### Details on how Sphinx works and investigation on what happened

It seems the [intersphinx mapping](https://github.com/digitalfabrik/integreat-cms/blob/6167119c2db90121cd801265748563bbec2d97d5/docs/src/conf.py#L121) that is telling sphinx where the documentation for `xhtml2pdf` is located is fine, the docs didn't move and [objects.inv](https://xhtml2pdf.readthedocs.io/en/latest/objects.inv) *(the index in binary format sphinx itself reads that lists all the references)* is also still accessible. I can only assume that the project restructured the docs and the reference is now called differently. You seem to link to [`usage.html`](https://xhtml2pdf.readthedocs.io/en/latest/usage.html), which is *(now)* a redirect to [`quickstart.html#use`](https://xhtml2pdf.readthedocs.io/en/latest/quickstart.html#use), so I'd expect `usage` became `use`. Well, it's quickly proven that this doesn't work either.

When we look at the [cross-referencing syntax](https://www.sphinx-doc.org/en/master/usage/referencing.html#cross-referencing-arbitrary-locations), we find that there are no automatic labels for subsections in a document *(which* Use *is, at least now)*, so we need to find whether the project defined one manually. However, when we click on [*View page source* at the top of the hosted xhtml2pdf docs](https://xhtml2pdf.readthedocs.io/en/latest/_sources/quickstart.rst.txt), we don't see any labels defined.

This means we can either do it the way you proposed, though this is not recommended because it is much more prone to breakage than labels. Alternatively, I would suggest to just do ``According to xhtml2pdf documentation (see :doc:`xhtml2pdf:quickstart`),``, given that the section before *Use* is pretty short so *Use* will already be scrolled into view on most devices.

#### Investigating where the explanation we actually want to link to ended up

However, if a reference disappeared, it seems to be overdue to check whether what we actually want to link to is still there. Our docs continue *"this function is necessary for resolving the django static files references."*, so we expect the target to explain something related to that, even if we don't fully understand the matter at hand. In *Quickstart*, I can find nothing that seems related. However, the docstring where the broken reference comes from belongs to a function `link_callback()`, so I searched for that name in the xhtml2pdf docs. It turns out that there is an [explanation in the reference section for *Python API*](https://xhtml2pdf.readthedocs.io/en/latest/reference/python.html#link-callback) – though also without a dedicated label.

Here there *is* a significant amount of content before the section, in addition to it not really being tied to any object and possibly moving around again in the future, so I am doubtful whether ``(see :doc:`xhtml2pdf:reference/python`)`` is the best move for us. I think I'm going to suggest adding labels to sections that seem like they could be cross-referenced more often, in the meantime we can link directly to the anchor.

#### Use the source, Luke!

Of course, I could have gotten a better picture more quickly if I had gone straight to the source code and looked for suspicious commits like *Better documentations structure*, which deletes `usage.rst` among others and completely restructures the docs *(as the name already makes plenty obvious)*, but started out with the suspicion that something was broken along intersphinx and then I just got stuck in investigation mode… oh well.

PeterNerlich (approved):

david-venhoff (approved):

david-venhoff:
https://chat.tuerantuer.org/digitalfabrik/pl/phhgnpfqofgqudacx554nfhcfr

Issue #3087, opened 2024-10-01
[2024-11-07] Show related contacts in POI form


MizukiTemma:
### Motivation

Each contact objects is related to a POI. Let's show which contacts is referring the POI in the POI form.

### Proposed Solution

Probably in the same/similar way with pages and POIs in the organisation form?

### Alternatives

Don't implement.

Issue #3147, opened 2024-10-23
Make information about contacts less redundant in the location form (Archive & Delete)


juliankehne:
### Motivation
Duplicates of contacts and information texts are redundant in the "Actions" tab of a Location and take up a lot of space. 
Archive and Delete options each have a copy of all linked contacts.

If one location has a lot of contacts the Actions tab will include a lot of contact entries.

See Screenshot.

### Proposed Solution

Solution 1: Have only one list of linked contacts

Solution 2: Solution 1 + The texts "You cannot archive a location which is referenced by a contact.
To archive this location, you have to delete these contacts first: " and for deleting seem redundant as well and could be merged into one message to:

"You cannot delete or archive a location..."

### Alternatives
Keep as it is or @nikolahoff can probably come up with something better :)

### User Story
As a manager I want user interface without a lot of redundancy to quickly see which contacts are blocking the deletion of a location / archiving of a location.

### Additional Context

![2024-10-23 13_13_05-Window](https://github.com/user-attachments/assets/a5d82d2a-1126-414f-b63e-70d1f9e02a65)

osmers:
I guess for users in municipalities this should generally not present a problem, bcs they cannot delete. Nonetheless, we can think about how to optimize it for us :)

osmers:
Maybe we can also address the issue, that pois which are referenced by events and contacts have the same text twice: 
![image](https://github.com/user-attachments/assets/100a20dd-bdb5-46a1-9c09-d835357f2e38)
We can probably combine that as well, e.g. by writing one warning text and then subdivide into events and contacts using headings.

MizukiTemma:
After #3087 there will be a list of contacts which are referring the POI opened in the form. It looks like the first photo currently:

![Screenshot 2024-11-05 183515](https://github.com/user-attachments/assets/5d750e70-89dd-4d75-bad3-63fed9f200b7)

while the action box looks like this in the same example:
![Screenshot 2024-11-05 183540](https://github.com/user-attachments/assets/c3c52497-46c8-42c6-844c-90e8d9095bec)

nikolahoff:
@osmers if this is still open we can discuss it in our UX <> Dev meeting.

nikolahoff:
[...]

Issue #3086, opened 2024-10-01
[2024-11-07] Remove E-mail, phone number and website from POI


MizukiTemma:
### Motivation

Currently POI has three fields for contact information: E-mail, phone number and website. We came to an idea, however, while discussing and implementing [contact management system](https://github.com/digitalfabrik/integreat-cms/issues/2955), that those fields/information must be managed in contacts in futur.

### Proposed Solution

- Remove those three fields from POI
- Adjust ajax POI form `poi_box.html` (remove *Contact details*, etc)
- Remove the collapsible box " Contact details" form the POI form [added on 04.12]
- Create a migration to create contacts from those poi fields and remove the fields from poi
- Adjust the api so that it returns the same results (Contacts without title <s>and name</s> could be treated as primary contact, and returned instead of the now removed data)

MizukiTemma:
As "primary contact" meant is the contact objects without "point_of_contact_for". There is a constraint that for each POI max. one primary contact exists.

For this issue I have imagined following work steps:

1. For each POI which has at least one of E-mail, phone number and website, create a contact object with them ("point_of_contact_for" and "name" stay empty).
2. Remove E-mail, phone number and website fields from POI model.
3. Remove the box of those three fields from the POI form.
4. Adjust API so it finds the primary contact of the requested POI and reads those fields from the primary contact.

osmers:
Just one question before this gets closed - didn't we say that this cannot be removed before we have not migrated the information stored there to a general contact?

david-venhoff:
Yes, this will be part of this issue:

> * Create a migration to create contacts from those poi fields and remove the fields from poi

Though it looks like it is currently missing from the pr

PR #3162, opened 2024-10-31
[2024-11-07] Show related contacts in the POI form

MizukiTemma:
### Short description

This PR implements a box into the POI form which shows contacts referring the POI.

### Proposed changes

- Add a new collapsible side box like "Organization", "Barrier free" or "Actions"
- List up related contacts with a link to contact form respectively
- Add tests for this box

### Side effects

- This brings more duplicate than mentioned in #3147 but it is consistent with other models (Organization, for example) to show the list of related objects. I'll leave the duplicate problem to #3147.

### Resolved issues

Fixes: #3087 

### DIFF
--- a/integreat_cms/cms/models/contact/contact.py
+++ b/integreat_cms/cms/models/contact/contact.py
@@ -3,6 +3,7 @@ from django.db.models import Q
 from django.db.utils import DataError
 from django.utils import timezone
 from django.utils.functional import cached_property
+from django.utils.translation import gettext
 from django.utils.translation import gettext_lazy as _
 
 from ..abstract_base_model import AbstractBaseModel
@@ -90,6 +91,21 @@ class Contact(AbstractBaseModel):
             return _("with website: {}").format(self.website)
 
         return ""
+
+    def label_in_reference_list(self) -> str:
+        """
+        This function returns a display name of this contact for the poi contacts list
+        """
+
+        label = (
+            _("General contact information")
+            if not self.point_of_contact_for
+            else f"{self.point_of_contact_for} {self.name}"
+        )
+        if self.archived:
+            label += " (⚠ " + gettext("Archived") + ")"
+
+        return label
 
     def get_repr(self) -> str:
         """
--- a/integreat_cms/cms/templates/pois/poi_form_sidebar/related_contacts_box.html
+++ b/integreat_cms/cms/templates/pois/poi_form_sidebar/related_contacts_box.html
@@ -1,0 +1,28 @@
+{% extends "../../_collapsible_box.html" %}
+{% load i18n %}
+{% load widget_tweaks %}
+{% block collapsible_box_icon %}
+    message-square
+{% endblock collapsible_box_icon %}
+{% block collapsible_box_title %}
+    {% trans "Related contacts" %}
+{% endblock collapsible_box_title %}
+{% block collapsible_box_content %}
+    {% with poi_form.instance.contacts.all as contacts %}
+        <div>
+            <div class="help-text">
+                {% if contacts %}
+                    {% trans "This location is currently referred to by the following contacts." %}
+                {% else %}
+                    {% trans "This location is not currently referred to in any contact." %}
+                {% endif %}
+            </div>
+            {% for contact in contacts %}
+                <a href="{% url 'edit_contact' contact_id=contact.id region_slug=request.region.slug %}"
+                   class="block pt-2 hover:underline">
+                    <i icon-name="pen-square" class="mr-2"></i> {{ contact.label_in_reference_list }}
+                </a>
+            {% endfor %}
+        </div>
+    {% endwith %}
+{% endblock collapsible_box_content %}
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -1680,9 +1680,9 @@ msgstr "Aktiv"
 msgid "Hidden"
 msgstr "Versteckt"
 
-#: cms/constants/region_status.py cms/models/pages/page_translation.py
-#: cms/models/regions/region.py cms/templates/events/event_form.html
-#: cms/templates/pages/page_form.html
+#: cms/constants/region_status.py cms/models/contact/contact.py
+#: cms/models/pages/page_translation.py cms/models/regions/region.py
+#: cms/templates/events/event_form.html cms/templates/pages/page_form.html
 #: cms/templates/pages/page_tree_archived_node.html
 #: cms/templates/pois/poi_form.html
 msgid "Archived"
@@ -2811,6 +2811,10 @@ msgstr "mit Telefonnummer: {}"
 #: cms/models/contact/contact.py
 msgid "with website: {}"
 msgstr "mit Website: {}"
+
+#: cms/models/contact/contact.py
+msgid "General contact information"
+msgstr "Allgemeine Kontaktinformationen"
 
 #: cms/models/contact/contact.py
 msgid "(Copy)"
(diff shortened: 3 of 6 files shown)


integreat_cms/cms/templates/pois/poi_form_sidebar/related_contacts_box.html
```diff
@@ -0,0 +1,28 @@
...
+                {% if contacts %}
+                    {% trans "This location is currently referred to in those contacts." %}
+                {% else %}
+                    {% trans "This location is not currently referred to in any contact." %}
+                {% endif %}
+            </div>
+            {% for contact in contacts %}
+                <a href="{% url 'edit_contact' contact_id=contact.id region_slug=request.region.slug %}"
+                   class="block pt-2 hover:underline">
+                    <i icon-name="pen-square" class="mr-2"></i> {{ contact.title }} {{ contact.name }}
```
david-venhoff, line 23:
```suggestion
                    <i icon-name="pen-square" class="mr-2"></i> {{ contact }}
```
I think we should use the the `__str__` method for contacts as it is currently being implemented in #3142, because we decided to have some more complex logic how to display a contact

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -7669,6 +7669,18 @@ msgstr "Nur Koordinaten übernehmen"
...
+msgid "Related contacts"
+msgstr "Zugehörige Kontakte"
+
+#: cms/templates/pois/poi_form_sidebar/related_contacts_box.html
+msgid "This location is currently referred to in those contacts."
+msgstr "Dieser Ort wird in folgenden Kontakten verwendet."
+
+#: cms/templates/pois/poi_form_sidebar/related_contacts_box.html
+msgid "This location is not currently referred to in any contact."
+msgstr "Zur Zeit wird dieser Ort in keinem Kontakt verwendet."
```
david-venhoff, line 7682:
```suggestion
msgstr "Zur Zeit gibt es keine Kontakte zu diesem Ort."
```

Nit: I think this might read a bit easier, but not too important I'd say

integreat_cms/release_notes/current/unreleased/3087.yml
david-venhoff, line 1: I'm not sure about the release note. As far as I know, we did not add any other release notes about contacts yet, so maybe it would be better to remove this.
Then, when we make contacts available to all users (And not just team members), we can add a release note like `add contacts`

david-venhoff (commented):
Thanks!

I have some questions:
- Should archived contacts be shown in the related contacts box? IMO we should at least note if they are archived, or hide them completely
- Issue #3086 mentions a `primary contact` that we will deliver in the api. Do you think it makes sense to highlight the primary contact also in the related contacts box?
- Should the box be positioned closely to the existing contact details box or should we leave that up to #3086?

integreat_cms/release_notes/current/unreleased/3087.yml
MizukiTemma: Ohw, right. It's still a secret 🤫

MizukiTemma:
The box is currently hidden for users except Service Team and CMS Team, as the contact feature itself is hidden.

MizukiTemma:
> * Should archived contacts be shown in the related contacts box? IMO we should at least note if they are archived, or hide them completely

I've added "(⚠ Archived)" to the archived contacts.


> * Issue [[2024-11-07] Remove E-mail, phone number and website from POI #3086](https://github.com/digitalfabrik/integreat-cms/issues/3086) mentions a `primary contact` that we will deliver in the api. Do you think it makes sense to highlight the primary contact also in the related contacts box?
> * Should the box be positioned closely to the existing contact details box or should we leave that up to [[2024-11-07] Remove E-mail, phone number and website from POI #3086](https://github.com/digitalfabrik/integreat-cms/issues/3086)?

@osmers 
What do you think?

osmers:
Sorry, I did not see the questions before :)

I think it makes sense to highlight the primary contact - maybe naming it something like general contact information?

And in my understanding the box showing related contacts should substitute the existing contact details box - if substituting is not possible then place it just below or above the existing box so that once we delete the existing box, it will basically be in the same position :)

MizukiTemma:
This PR is currently being adjusted for the changes lately introduced.

osmers:
> <point of contact for (title)> (+ name if name is given) for all other contacts. I know this schema is different from what introduced in #3142 (Contact for [NAME_OF_LOCATION] with [ATTRIBUTE]: [ATTRIBUTE]). [NAME_OF_LOCATION] is omitted because it is on the POI form and therefore it's clear about which POI it is. E-mail, phone number or website will not be shown here, for title (point of contact for) and name (if given) are easier to understand what contact it is (in my opinion).
> 
> Is this way Okay for service team?

In theory (not showing the POI) yes, but what happens when no name is provided and only a title (which might be double)? Do we then add the phone, email or webseite? Not sure if this will happen often, it's probably an edge case, but who knows what the municipalities will come up with.

MizukiTemma:
> In theory (not showing the POI) yes, but what happens when no name is provided and only a title (which might be double)? Do we then add the phone, email or webseite? Not sure if this will happen often, it's probably an edge case, but who knows what the municipalities will come up with.

@osmers 
In the current implementation only the title will be shown. I don't think we get so much of doubles when the list is "per POI". If more than one contact with the same title are needed but no individual name should be given, users will name them like "Ansprechpartner (Nachname A-K)", "Ansprechpartner (Nachname K-P) etc for their own understandability. And a problem with adding E-mail etc. is that the display name will be too long for such a small box 🤔

MizukiTemma:
@osmers 
Did you mean "OK with the current implementation" with 👍 or is it still on a discussion/thinking?

osmers:
Sorry @MizukiTemma 
Yeah, I meant I'm ok with it bcs I find it hard to think of all the possible "problems" :D 
So we'll do it this way and then see what kind of feedback we get.

integreat_cms/cms/templates/pois/poi_form_sidebar/related_contacts_box.html
```diff
@@ -0,0 +1,28 @@
...
+{% endblock collapsible_box_icon %}
+{% block collapsible_box_title %}
+    {% trans "Related contacts" %}
+{% endblock collapsible_box_title %}
+{% block collapsible_box_content %}
+    {% with poi_form.instance.contacts.all as contacts %}
+        <div>
+            <div class="help-text">
+                {% if contacts %}
+                    {% trans "This location is currently referred to in those contacts." %}
```
jarlhengstmengel, line 15:
```suggestion
                    {% trans "This location is currently referred to by the following contacts." %}
```
I think this would be more consistent with the German translation in the translation file

integreat_cms/cms/models/contact/contact.py
```diff
@@ -91,6 +92,21 @@ def get_additional_attribute(self) -> str:
 
         return ""
 
+    def label_in_reference_list(self) -> str:
+        """
+        This function returns
```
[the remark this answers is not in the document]
MizukiTemma: ohw 🙈

integreat_cms/cms/models/contact/contact.py
```diff
@@ -91,6 +92,21 @@ def get_additional_attribute(self) -> str:
...
+        """
+
+        label = (
+            _("General contact information")
+            if not self.point_of_contact_for
+            else f"{self.point_of_contact_for} {self.name}"
+        )
+        if self.archived:
+            label += " (⚠ " + gettext("Archived") + ")"
```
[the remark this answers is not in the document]
david-venhoff: I don't think that it makes a difference here which translation function is used, so imo it would be nicer to not import `gettext_lazy`

Probably a nit though :)

integreat_cms/cms/models/contact/contact.py
```diff
@@ -91,6 +92,21 @@ def get_additional_attribute(self) -> str:
 
         return ""
 
+    def label_in_reference_list(self) -> str:
+        """
+        This function returns a display name of contact
```
david-venhoff, line 97:
```suggestion
        This function returns a display name of this contact for the poi contacts list
```

david-venhoff (approved):
Thanks, very nice!

The only thing missing I think is that the new contact box should replace the existing contact details box

integreat_cms/cms/models/contact/contact.py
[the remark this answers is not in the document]
jarlhengstmengel: What I tried it doesn't make a difference in the end. It's probably more consistent with our documentation with `gettext_lazy` and since it's already imported, it would spare one import. But it's not really that important

jarlhengstmengel (approved):
Thank you, looks good!

MizukiTemma:
@david-venhoff 
> The only thing missing I think is that the new contact box should replace the existing contact details box

It'll be done as a part of #3086

PR #3163, opened 2024-10-31
Improve social media preview

svenseeberg:
### Short description
Improve the social media previews.

More info: https://chat.tuerantuer.org/digitalfabrik/pl/rnoez75xxjyhugxo97mbjuqmte & https://tasks.tuerantuer.org/projects/infra/work_packages/3734/activity?query_id=59

Previews on Signal and WhatsApp are not yet working.

### Proposed changes
- Improve edge case handling in excerpts.

### Side effects
- N/A

### DIFF
--- a/integreat_cms/api/v3/social_media_headers.py
+++ b/integreat_cms/api/v3/social_media_headers.py
@@ -40,13 +40,19 @@ logger = logging.getLogger(__name__)
 
 def get_excerpt(content: str) -> str:
     """
-    Correctly escapes, truncates and normalizes the content of the page to display in a search result
+    Correctly escapes, truncates and normalizes the content of the page to display in a search result.
+    Then apply some crazy string operations to handle several edge cases.
 
     :param content: The content of the page
 
     :return: A page excerpt containing the first 100 characters of "raw" content
     """
-    return unescape(strip_tags(content))[:100].replace("\n", " ").replace("\r", "")
+    stripped_content = unescape(
+        strip_tags(content.replace("\n", " ").replace("\r", "").replace("<br>", " "))
+    )
+    if len(stripped_content) <= 100:
+        return stripped_content.strip().replace("  ", " ")
+    return stripped_content[:100].strip().rsplit(' ', 1)[0]+" ..."
 
 
 def get_region_title(region: Region, page_title: str) -> str:
--- a/integreat_cms/api/v3/templates/social_media_headers.html
+++ b/integreat_cms/api/v3/templates/social_media_headers.html
@@ -6,10 +6,12 @@
     <meta name="twitter:card" content="summary">
     <meta property="og:title" content="{{ title }}">
     {% if excerpt %}
+        <meta name="description" content="{{ excerpt }}">
         <meta property="og:description" content="{{ excerpt }}">
         <meta name="twitter:description" content="{{ excerpt }}">
     {% endif %}
     <meta property="og:image" content="{{ image }}" />
+    <meta name="twitter:image" content="{{ image }}" />
     <meta property="og:image:width" content="1200" />
     <meta property="og:image:height" content="630" />
     <meta property="og:url" content="{{ url }}" />


integreat_cms/api/v3/social_media_headers.py
```diff
@@ -40,13 +40,19 @@
...
     :param content: The content of the page
 
     :return: A page excerpt containing the first 100 characters of "raw" content
     """
-    return unescape(strip_tags(content))[:100].replace("\n", " ").replace("\r", "")
+    stripped_content = unescape(
+        strip_tags(content.replace("\n", " ").replace("\r", "").replace("<br>", " "))
+    )
+    if len(stripped_content) <= 100:
+        return stripped_content.strip().replace("  ", " ")
```
JoeyStk, line 54: Why do we need this line? :)

svenseeberg: For example 
```
Hallo<br>
Welt
```
will evaluate to `Hallo  Welt` as we replaced newlines and br tags with empty spaces.

PeterNerlich: Wouldn't `re.sub(r"\s+", " ", stripped_content.strip())` work better in this case? Who says we only have two spaces?

integreat_cms/api/v3/social_media_headers.py
```diff
@@ -40,13 +40,19 @@
...
+    return stripped_content[:100].strip().rsplit(' ', 1)[0]+" ..."
```
PeterNerlich, line 55:
```suggestion
    return stripped_content[:100].strip().rsplit(' ', 1)[0]+" …"
```

PeterNerlich:
Looking at https://ogp.me/, we might benefit from adding/improving
- *`og:type` (seems to already be defined elsewhere…? investigate where)*
- *`og:url` should be a fixed (= unchanging) url of the content and functions as an ID of the node on the graph.*
  Investigate whether `url` is a permalink or contains the slug that might change for a page.
- `og:locale` Language of the content *(e.g. `en_US`)*
- `og:locale:alternate` Alternate locales the content is available in *(repeat this meta tag for each alternative)*
- `og:site_name` Name of the overall website the content is part of. I guess this would just be `Integreat` and should probably not be part of `og:title` if defined here…?
- `og:image:secure_url` Described as *"An alternate url to use if the webpage requires HTTPS"* – Might be out of date, but won't hurt to add it. Will be the same value as `og:image`
- `og:image:alt` Alt text for the image *(not a caption)*. I could imagine this to be the problem on WhatsApp and/or signal, as ogp.me says *"If the page specifies an og:image it should specify `og:image:alt`."*

PeterNerlich:
Did a few tests that didn't support my hypothesis that the absence of the mentioned meta tags had any impact on the card being shown in whatsapp or not.
What does seem to make a difference is whether it is a full URL or just a path implied to be for the URL the document is queried from. Integreat is currently just sending a path.

Also, I noticed that it is unclear to me how the process looks like and whether the app side just uses our snippet as is or directly references some value *(and might therefore not adopt all values we send)*, so more investigation is needed for me

JoeyStk:
Since there has been no work on this in half a year we closed this PR as stale

svenseeberg:
We need to fix this.

integreat_cms/api/v3/social_media_headers.py
```diff
@@ -41,13 +42,25 @@
 
 def get_excerpt(content: str) -> str:
     """
-    Correctly escapes, truncates and normalizes the content of the page to display in a search result
+    Correctly escapes, truncates and normalizes the content of the page to display in a search result.
+    Then apply some crazy string operations to handle several edge cases.
```
jonbulz: I personally feel like this insertion doesn't really add any information: I'd either leave it as it is or specify which operations we added to handle what edge-case.

svenseeberg (commented):
Approve w/o approve button :rocket:

Issue #2994, opened 2024-08-13
Confusing order of push notifications in Integreat App/Website


MizukiTemma:
### Motivation

Push notifications are deleted after a certain period of time (currently 28 days), but sometimes push notifications that are older than 28 days appear in the message list in Integreat Website/app. This happened as a push notification was created more than 28 days ago but sent at a later time, which is whithin 28 days (a real example: today is 13.08.2024, a push notification was created on 21.06.2024, therefore older than 28 days but shown in the website/app, because it was sent on 18.07.2024).

This is confusing. We have to change how to sort push notifications and how to count their "age".

### Proposed Solution

TBD

### Alternatives

Leave it as it is and get confused sometimes 

### Additional Context

See the context [here](https://chat.tuerantuer.org/digitalfabrik/pl/3ae6at4n9jgpjjsuny81a3116e)

### Design Requirements

UI/UX team will give us an input how to sort/calculate age of push notifications.

osmers:
We also just received an email from a municipality exactly regarding this issue - they also suggested to show the publish/sending date instead of the creation/last update date.
I think that makes a lot of sense - especially if some messages are created well in advance

PeterNerlich:
My gut feeling is that we should simply check in the view, when displaying to the user, whether the last modified timestamp or the sent timestamp is greater and just use that for sorting

osmers:
I had a chat with @nikolahoff and she agreed that the most important thing is having the most recent notification/change up top - so in her opinion there is no problem with sorting by the most current date. 
@PeterNerlich could you maybe open an issue in the app board?

osmers:
@PeterNerlich are the labels on this issue still correct? And did you open the app board issue?

PeterNerlich:
It turns out we do sorting on CMS side after all and also don't expose the sent date via the api at all, so this becomes a task for us once again

laravoelk:
We just had a feedback from one city who stated how confusing it is that the date shown in push notifications is the date, when it was written, and not when it was send. Could you maybe have another look at this issue? Thank you so much!

laravoelk:
Yes, I mean the date shown in the App. :-)

MizukiTemma:
@laravoelk 
Ok, I created a new issue for it

JoeyStk:
@PeterNerlich could you please have another look into this? :)

PR #3164, opened 2024-10-31
Sort news/push notifications

PeterNerlich:
### Short description

News items are sent out as push notifications to subscribed devices as well as served via our API. Push notifications are saved as separate objects in our database in order to retain the information for which news items we already sent notifications and to keep that separate from the concept of the content of the news item itself.

News items have a property `last_updated`, which describes when the content was last edited. Push Notification objects have a `sent_date`, which describes when the notification for this news item in that channel and this language was sent out. In the API, we only expose `last_updated`, which is the value also being displayed in the app. This can cause confusion whenever the date the news item was last edited is far before the date when it was actually sent.

### Proposed changes

- Add a value `display_date` to each news item returned by the API which equals either `last_updated` or `sent_date` of the corresponding push notification, whichever is greater *(effectively clamping the value of `last_updated` to be no earlier than `sent_date`)*
- Sort the news entries returned by the API by the `display_date` value instead of `last_updated`

> *Alternatively, we could decide to alter `last_updated` or `timestamp` to the new behaviour. It seems that we deprecated `timestamp` in 2022 in favour of the new `last_updated`, but it is still in use by the app. There are hints that converting the timestamp to be in the local time zone of the server is the reason, but we should ask directly if we wish to pursue such an alternative.*

### Side effects

- After this PR is merged and released, the app will need to switch from ~~`last_updated`~~ `timestamp` over to `display_date`. No further change should be required.
  If it aligns with how the app team handles this, https://github.com/digitalfabrik/integreat-app/issues/2952 might be repurposed for that request.

### Resolved issues

Fixes: #2994

### DIFF
--- a/docs/src/api-docs.rst
+++ b/docs/src/api-docs.rst
@@ -673,6 +673,7 @@ RESPONSE
       "message": String,       // The message of the push notification in the given language
       "timestamp": String,     // Deprecated field
       "last_updated": String,  // The date&time when the push notification was last updated
+      "display_date": String,  // The date&time when the push notification was sent / last updated (whichever is newer)
       "channel": String,       // The channel the push notification was sent to (e.g. "News")
       "available_languages": [           // The available languages of the push notification
             "<language_slug>": {
--- a/integreat_cms/api/v3/push_notifications.py
+++ b/integreat_cms/api/v3/push_notifications.py
@@ -7,6 +7,8 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 from django.conf import settings
+from django.db.models import F
+from django.db.models.functions import Greatest
 from django.http import JsonResponse
 from django.utils import timezone
 
@@ -44,7 +46,10 @@ def sent_push_notifications(
         )
         .filter(language__slug=language_slug)
         .filter(push_notification__draft=False)
-        .order_by("-last_updated")
+        .annotate(
+            display_date=Greatest(F("last_updated"), F("push_notification__sent_date"))
+        )
+        .order_by("-display_date")
     )
     if channel != "all":
         query_result = query_result.filter(push_notification__channel=channel)
@@ -69,6 +74,7 @@ def transform_notification(pnt: PushNotificationTranslation) -> dict[str, Any]:
         "message": pnt.get_text(),
         "timestamp": pnt.last_updated,  # deprecated field in the future
         "last_updated": timezone.localtime(pnt.last_updated),
+        "display_date": pnt.display_date,
         "channel": pnt.push_notification.channel,
         "available_languages": available_languages_dict,
     }


MizukiTemma (approved):
Looks good 😸

MizukiTemma:
>After this PR is merged and released, the app will need to switch from last_updated timestamp over to display_date. No further change should be required.
If it aligns with how the app team handles this, https://github.com/digitalfabrik/integreat-app/issues/2952 might be repurposed for that request.

@steffenkleinle FYI

steffenkleinle:
> > After this PR is merged and released, the app will need to switch from last_updated timestamp over to display_date. No further change should be required.
> > If it aligns with how the app team handles this, [digitalfabrik/integreat-app#2952](https://github.com/digitalfabrik/integreat-app/issues/2952) might be repurposed for that request.
> 
> @steffenkleinle FYI

Thanks for the ping. Could we perhaps adjust the naming to `last_update` to be consistent with all other endpoints? In the end, sending a push notification is also some sort of update to the news item. Otherwise, this looks good as is.

david-venhoff (commented):
Not a review, but I noticed that this change should be documented in our api documentation

JoeyStk (changes requested):
Thank you very much! 
Could you please still add the documentation as @david-venhoff had pointed out and rename the endpoint as @steffenkleinle had pointed out? That'd be great :)
Other than that I don't have any further requests :)

JoeyStk:
@PeterNerlich what is the state of this PR? :)

JoeyStk:
@PeterNerlich quick question: What is the state of this PR? Are you still going to work on it? :)

JoeyStk (approved):
LGTM :rocket:

steffenkleinle:
> @steffenkleinle Sorry for the long delay. After procrastinating over adjusting this as you suggested, or adding `display_date` to every endpoint for consistency, or merging it as is for way too long because I didn't really feel comfortable with either option, I asked for the opinion of the team. They saw it fair to just merge it with `display_date` – in the future, we will hopefully implement a proper strategy for versioning our API, and on the next major version `last_update` will probably behave like `display_date` does now. I hope this is an acceptable solution!

Sorry, but I am a little lost here. What does that mean for us? Can you provide a short summary which properties exist now with which behavior and which one we should use for sorting? Or is the sorting done entirely in the CMS already?

PeterNerlich:
@steffenkleinle Sure thing!

- `last_update` is kept as is for now
- `display_date` is the new field with the intended behaviour of assuming either the date the notification was sent, or when the content was last updated, whichever is newer. This is the value that should be shown in the app from now on
- The CMS is delivering this list pre-sorted, newest `display_date` first. I'm not sure if you want to rely on that or make sure we don't send you bogus data though, you are the representation layer after all
- Other endpoints are not affected

Issue #3131, opened 2024-10-12
[2024-12-18] Make box "Number of online accesses" in statistics collapsible


JoeyStk:
### Motivation

As our statistics page will grow as part of #1097, we want our existing statistics to become collapsible.

### Proposed Solution

Make use of our [template for collapsible boxes](integreat_cms/cms/templates/_collapsible_box.html) and convert the box "Number of online accesses" to a collapsible box.

### Alternatives

None

### User Story

As a user of the CMS I want to not have a cluttered screen so that I find information more quickly.

### Additional Context

This is part of our meta issue #1097 

### Design Requirements

Design proposal (v3):
https://www.figma.com/file/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=1179-830

Issue #1097, opened 2022-01-06
Page access statistics


svenseeberg:
### Motivation
Municipalities may want to find out which pages are being accessed most/least. We already collect this information in Matomo. This should be accessible to content managers.

### Proposed Solution
Add a view that shows access numbers on pages in a time range.

### Alternatives
We keep generating these reports by hand or give municipalities access to Matomo.

dkehne:
![Unbenannt](https://github.com/digitalfabrik/integreat-cms/assets/4182469/febb9459-4e13-4e7e-a7fd-cf929ed6dd5a)

@hauf-toni just a rough idea how i could imagine the page access overeview.

hauf-toni:
**perhaps there is a need for discussion here:** should the "page access statistics" page appear in the sidebar as a single item under "analysis" (same emphasis as e.g. "feedback" or "translation report") or as a sub-item under the current access figures as a new box? 

imo, a more prominent placement would **increase visibility, facilitate access, encourage monitoring**, and **demonstrate the value of the feature**. on the other hand, **users might overestimate the importance of the data presented** and draw the wrong conclusions (access numbers are not synonymous with importance). of course, this problem could possibly be solved by training the communities to interpret the presented data in the right way. 

💬 **are there opinions on this that you would like to share?**

osmers:
@hauf-toni I think it would make sense to not show it as a tab at all but to enable users to get to this specific view from the statistics page itself...
If that is not prefered, then I think we should have one tab for overall statistics and one for page specific stats - i would not work with sub-tabs as it makes the menu more complex.

svenseeberg:
I think there are a couple of caveats:
- What is the relevance of a list of the most read pages? That the content is good? That does not provide any actionable information.
- Should we provide a list of pages with only 1 or 2 hits per month (even zero in probably quite a lot of instances)? What would be the actionable information here? Just delete them, because they are not being used?

I think we need to define the goals first. Then the implementation will probably be easy to derive.

svenseeberg:
> the plan is to write a guide for the municipalities on how to interpret the page statistics. 

So what is it going to tell? Then we can derive a good presentation of the data.

svenseeberg:
> So what is it going to tell? Then we can derive a good presentation of the data. I think we also need to differentiate interpretation of data from actionable information.

We can recommend to use the data as a starting point for further questions, for example "Why does a page receive so few/many visitors?"

osmers:
[...]

PR #3170, opened 2024-11-04
Add separate legend for statistics charts

charludo:
### Short description

Remove the legend at the top of the chart, replacing it with a box in the right-hand column, as shown in the [design proposal](https://www.figma.com/file/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=1179-830) in preparation of #1097.

### Proposed changes

- remove old legend
- generate new legend server-side, using jinja
- wire them together using the method recommended in the [docs for chart.js](https://www.chartjs.org/docs/master/samples/legend/html.html)

### Side effects

- none

### Resolved issues

Fixes: #3133

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #3175, opened 2024-11-05
[2024-12-18] Make the number of online accesses in statistics collapsible

lunars97:
### Short description

Make the number of online accesses collapsible as a part of the #1097 as shown in [design proposal v3](https://www.figma.com/design/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=1179-830&node-type=frame&t=vaCXaXccgqveytL3-0)

### Proposed changes

- Make use of the template for collapsible box
- Create a new file called statistics to hold all the upcoming collapsible separate boxes in statistics and the sidebar
- Change the path to the template in statistics view

### Side effects

- should be none?

### Resolved issues

Fixes: #3131 

### DIFF
--- a/integreat_cms/cms/templates/statistics/statistics_sidebar.html
+++ b/integreat_cms/cms/templates/statistics/statistics_sidebar.html
@@ -1,0 +1,52 @@
+{% load i18n %}
+{% load static %}
+{% load widget_tweaks %}
+<div class="rounded border border-solid border-blue-500 shadow-2xl bg-white hidden mb-2"
+     id="chart-legend-container">
+    <div class="rounded p-4 bg-water-500">
+        <h3 class="heading font-bold text-black">
+            <i icon-name="settings" class="pb-1"></i> {% translate "Adjust shown data" %}
+        </h3>
+    </div>
+    <div id="chart-legend">
+        {% if request.region.statistics_enabled and perms.cms.view_statistics %}
+            {% include "statistics/_statistics_legend.html" with box_id="statistics_legend" %}
+        {% endif %}
+    </div>
+</div>
+<div class="rounded border border-solid border-blue-500 shadow-2xl bg-white">
+    <div class="rounded p-4 bg-water-500">
+        <h3 class="heading font-bold text-black">
+            <i icon-name="calendar" class="pb-1"></i> {% translate "Adjust time period" %}
+        </h3>
+    </div>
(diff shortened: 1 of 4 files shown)


JoeyStk (changes requested):
Thank you for your PR! To me the boxes look at the moment like this 
grafik
This is not yet exactly what we want (s. design)
grafik.

As a general hint: I think you need to extract more parts from the `statistics_overview.html` file to the `statistics.html` file :)

lunars97:
@JoeyStk I think I need to wait for this ticket #3170 in order to properly connect them, what do you think? or is it too much for this PR?

JoeyStk:
> @JoeyStk I think I need to wait for this ticket https://github.com/digitalfabrik/integreat-cms/pull/3170 in order to properly connect them, what do you think? or will it be too much for this PR?

Yeah, that probably makes sense :)

JoeyStk (approved):
Thank you very much! It works as expected now :)
Thank you also for cleaning up the files! :) It's much appreciated :)

integreat_cms/cms/templates/statistics/statistics_sidebar.html
```diff
@@ -0,0 +1,52 @@
...
+{% load widget_tweaks %}
+<div class="rounded border border-solid border-blue-500 shadow-2xl bg-white hidden mb-2"
+     id="chart-legend-container">
+    <div class="rounded p-4 bg-water-500">
+        <h3 class="heading font-bold text-black">
+            <i icon-name="settings" class="pb-1"></i> {% translate "Adjust shown data" %}
+        </h3>
+    </div>
+    <div id="chart-legend">
+        {% if request.region.statistics_enabled and perms.cms.view_statistics %}
```
charludo, line 12:
```suggestion
```
The same check is already performed in `statistics_overview.html`, right before importing this file.

Actually, maybe wrapping the entire contents of `statistics_overview.html` (i.e., everything in the `content` block) would be an option and allow you to remove all other checks?

charludo (approved):
Thanks, this looks and works great! I just have one nit (see below), where IMO we could remove a couple of the duplicate permission checks.

integreat_cms/cms/templates/statistics/statistics_sidebar.html
lunars97: yes, you are right :+1: Thank you for bringing this up :)

Issue #3174, opened 2024-11-05
Region condition: Add total number of pages column


david-venhoff:
### Motivation

To be able to calculate percentages, the region condition data export needs an additional column: `total number of pages` 

### Proposed Solution

Add this column

### Alternatives

None

### User Story

As a service team member I need the total number of pages to calculate percentages in the data export

### Additional Context

Original request by @osmers: https://github.com/digitalfabrik/integreat-cms/issues/3053#issuecomment-2454062136

JoeyStk:
This issue is not resolved yet, as it leads to a server timeout

david-venhoff:
Superseded by #3189

juliankehne:
Tested successfully on cms-test!

PR #3178, opened 2024-11-05
Add total number of pages to region condition

JoeyStk:
### Short description

The current region condition csv doesn't contain the number of pages per region. This renders some of the information as useless, since they are relative to the number of pages. This PR attempts to fix this

### Proposed changes

- Add column for total number of pages per region to csv export

### Side effects

- None?

### Resolved issues

Fixes: #3174


MizukiTemma (approved):

lunars97 (approved):
Thank you for this PR. Works as expected :+1: One note: I think we need to create a release note for this, am I right? :)

JoeyStk:
> One note: I think we need to create a release note for this, am I right? :)

As this is a feature, that's only visible for our team (network management) I don't think this information is going to be interesting for the municipalities :thinking:

PR #3180, opened 2024-11-06
Remove contact data fields from poi

lunars97:
### Short description

Remove contact box(email, phone number and website) from the poi. These fields should be managed in the contacts.

### Proposed changes

- [x] Remove contact data from the poi model
- [x]  Remove contact data from the poi form
- [x] Migration should be created
- [x]  Adjust API so it finds the primary contact of the requested POI and reads those fields from the primary contact
- [x] Remove contact box from poi form
- [x] Adjust test data

### Side effects

- none?

### Resolved issues

Fixes: #3086 


MizukiTemma:
We should merge this PR in such a timing that it will be delivered at the same time with (or after) the final release of the contact feature for all users, as this removes the contact fields from POI and they must be then shown as primary contact in the related contact box (which is currently visible only for Service Team and CMS Team).

integreat_cms/cms/migrations/0112_remove_poi_email_remove_poi_phone_number_and_more.py
```diff
@@ -0,0 +1,55 @@
...
+if TYPE_CHECKING:
+    from django.apps.registry import Apps
+
+
+def create_primary_contact(apps: Apps) -> None:
+    """
+    Create primary contact from the poi contact data
+
+    :param apps: The configuration of installed applications
+    """
```
MizukiTemma, line 16:
```suggestion
if TYPE_CHECKING:
    from django.apps.registry import Apps
    from django.db.backends.base.schema import BaseDatabaseSchemaEditor


def create_primary_contact(
    apps: Apps,
    schema_editor: BaseDatabaseSchemaEditor,  # pylint: disable=unused-argument
) -> None:
    """
    Create primary contact for each POI and remove E-mail, phone number and website fields from POI model

    :param apps: The configuration of installed applications
    :param schema_editor: The database abstraction layer that creates actual SQL code
    """
```

To let the system work again, `schema_editor: BaseDatabaseSchemaEditor` is added.

The change in the documentation sentence is just a matter of preference, feel free to ignore if it's not appealing.

integreat_cms/cms/migrations/0112_remove_poi_email_remove_poi_phone_number_and_more.py
```diff
@@ -0,0 +1,55 @@
...
+    POI = apps.get_model("cms", "POI")
+    Contact = apps.get_model("cms", "Contact")
+    for poi in POI.objects.all():
+        if poi.email or poi.phone_number or poi.website:
```
MizukiTemma, line 20: We should probably check here whether there is already a primary contact for the POI, to go safer with the constraint that there can only be one single primary contact for each POI.

MizukiTemma (changes requested):
Thank you for opening a PR :muscle: I have some initial suggestions.

MizukiTemma:
@lunars97 

The box " Contact details" must be removed too. After rebasing, a new box with related contacts (introduced by #3162 ) should appear there.

integreat_cms/cms/migrations/0112_remove_poi_email_remove_poi_phone_number_and_more.py
lunars97: I think your version is more explanatory and thank you for solving the schema error :heart_eyes:

integreat_cms/cms/migrations/0112_remove_poi_email_remove_poi_phone_number_and_more.py
lunars97: I included the check for it, but I am not sure whether I was able to really check it :see_no_evil:

lunars97:
@MizukiTemma I am receiving test errors, is it because of the test data, specifically, the poi which should not have contact data anymore? :thinking: When I do `pre-commit` it does not show any test errors...

integreat_cms/cms/migrations/0112_remove_poi_email_remove_poi_phone_number_and_more.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+     Create primary contact for each POI and remove E-mail, phone number and website fields from POI model
+
+    :param apps: The configuration of installed applications
+    :param schema_editor: The database abstraction layer that creates actual SQL code
+    """
+    POI = apps.get_model("cms", "POI")
+    Contact = apps.get_model("cms", "Contact")
+    for poi in POI.objects.all():
+        if not Contact.objects.filter(location=poi).exists():
```
MizukiTemma, line 23:
```suggestion
        if not Contact.objects.filter(location=poi, point_of_contact_for="").exists():
```

integreat_cms/api/v3/locations.py
```diff
@@ -70,6 +70,11 @@ def transform_poi_translation(poi_translation: POITranslation) -> dict[str, Any]
     """
 
     poi = poi_translation.poi
+
+    primary_contact = Contact.objects.filter(
+        location=poi, point_of_contact_for="", name=""
```
MizukiTemma, line 75:
```suggestion
        location=poi, point_of_contact_for=""
```

Sorry, that was again a mistake in the issue description. The definition of "primary contact" is contact witohut title (point_of _contact_for) 🙈

MizukiTemma (approved):
Thank you so much 😻

Only one suggestion left, which is actually because of the innacurate issue description 🙈 🙏

JoeyStk (commented):
Thank you very much!
To me this looks very good. I want to double-check the migration in the next days and will do a full review for this PR after :)
For now I have one question: This is maybe for @osmers Is it okay to merge this PR already soon, even before the contact feature is released or should we wait until we merge this until the contact feature is released?
(as a reminder: this PR removes the contact section from pois and adds them to contacts instead)

JoeyStk (approved):
Except for my question to @osmers I think this looks good. I'm still a bit unsure about the migration, so maybe we can have someone triple-check the code snippet, but other than that I think this looks good :)

osmers:
I'd wait with merging it - unless we can guarantee that the contact function and this PR will be in the same release. Otherwise where would municipalities write down contact details if this is merged and the contact functionality not yet available?

david-venhoff (approved):

MizukiTemma:
Maybe turn this PR into draft to avoid accidental merge?

integreat_cms/cms/migrations/0112_remove_poi_email_remove_poi_phone_number_and_more.py
charludo, line 1: Needs to be renamed!

Issue #3140, opened 2024-10-21
Pylint directives are applied wrong


david-venhoff:
### Describe the Bug

Looks like we have been applying the pylint `# pylint: disable=…` directives wrong :see_no_evil: 

It is usually wrong to place a pylint directive directly on top of a function, because this directive will then apply to its enclosing block, and not just the function. The correct way to suppress pylint lints for a function is to make the directive the first line of the function. 
More information: https://pylint.pycqa.org/en/latest/user_guide/messages/message_control.html#block-disables

### Additional Information

Discovered [here](https://github.com/digitalfabrik/integreat-cms/pull/3075#discussion_r1801247908)

PR #3182, opened 2024-11-07
Fix pylint directives

MizukiTemma:
### Short description

This PR relocates the pylint directives so they are only applied for the target part.

### Proposed changes

- Move `# pylint: disable=......` from the top of the function to each target part.

### Side effects

- should be none
- There may be lines which I overlooked. Feel free to directly commit onto this branch 😸 

### Resolved issues

Fixes: #3140 

### DIFF
--- a/docs/src/code-style.rst
+++ b/docs/src/code-style.rst
@@ -58,10 +58,13 @@ Run pylint with our developer tool :github-source:`tools/pylint.sh`::
 
     ./tools/pylint.sh
 
-When you think a warning is a false positive, add a comment before the specific line::
+When you think a warning is a false positive, add a comment (see :doc:`pylint:user_guide/messages/message_control`)::
 
-    # pylint: disable=unused-argument
-    def some_function(*args, **kwargs)
+    def some_function(
+        something: SomeModel,  # pylint: disable=unused-argument
+        *args, 
+        **kwargs) -> None:
+        # pylint: disable=too-many-branches
 
 .. Note::
 
--- a/integreat_cms/api/v3/social_media_headers.py
+++ b/integreat_cms/api/v3/social_media_headers.py
@@ -183,9 +183,10 @@ def root_social_media_headers(
 
 
 @partial_html_response
-# pylint: disable=unused-argument
 def region_social_media_headers(
-    request: HttpRequest, region_slug: str, language_slug: str | None = None
+    request: HttpRequest,
+    region_slug: str,  # pylint: disable=unused-argument
+    language_slug: str | None = None,
 ) -> HttpResponse:
     """
     Generally renders the social media headers for a root region page.
@@ -211,9 +212,11 @@ def region_social_media_headers(
 
 
 @partial_html_response
-# pylint: disable=unused-argument, fixme
 def page_social_media_headers(
-    request: HttpRequest, region_slug: str, language_slug: str, path: str
+    request: HttpRequest,
+    region_slug: str,  # pylint: disable=unused-argument
+    language_slug: str,
+    path: str,
 ) -> HttpResponse:
     """
     Tries rendering the social media headers for a page in a specified region and language.
@@ -236,6 +239,7 @@ def page_social_media_headers(
     ):
         raise Http404("Page not found in this region with this language.")
 
+    # pylint: disable=fixme
     # TODO: add breadcrumb json-ld if content_translation exists
     return render_social_media_headers(
         request=request,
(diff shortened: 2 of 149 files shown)


docs/src/code-style.rst
```diff
@@ -58,10 +58,13 @@ Run pylint with our developer tool :github-source:`tools/pylint.sh`::
 
     ./tools/pylint.sh
 
-When you think a warning is a false positive, add a comment before the specific line::
+When you think a warning is a false positive, add a comment::
```
PeterNerlich, line 61: What about linking straight to the pylint documentation where different forms and their effects are explained? The section [has a convenient label](https://github.com/pylint-dev/pylint/blob/main/doc/user_guide/messages/message_control.rst?plain=1#L26), so we should be able to just reference it like this:

```suggestion
When you think a warning is a false positive, add a comment (see :doc:`pylint:block_disables`)::
```

…if we add their docs to our [intersphinx mapping](https://github.com/digitalfabrik/integreat-cms/blob/develop/docs/src/conf.py#L75) in `docs/src/conf.py`:

```diff
     "dateutil": ("https://dateutil.readthedocs.io/en/stable/", None),
     "geopy": ("https://geopy.readthedocs.io/en/stable/", None),
     "lxml": ("https://lxml.de/apidoc/", None),
+    "pylint": ("https://pylint.readthedocs.io/en/latest/", None),
     "python": (
         f"https://docs.python.org/{sys.version_info.major}.{sys.version_info.minor}/",
         None,
```

PeterNerlich: @MizukiTemma Sorry, I got confused about sphinx's roles in cross referencing. It should be `` :doc:`pylint:block_disables` `` (I checked this time)

I will not be able do a full review of the PR though, sorry

integreat_cms/cms/views/analytics/app_size_view.py
```diff
@@ -29,6 +28,7 @@ def get_context_data(self, **kwargs: Any) -> dict:
         """
         context = super().get_context_data(**kwargs)
 
+        # pylint: disable=fixme
```
JoeyStk, line 31: I don't think we need this comment anymore :D I hope/think that we have been using the correct calculation for a long time already :)

tests/core/management/commands/test_summ_ai_bulk.py
```diff
@@ -53,12 +53,12 @@ def test_summ_ai_bulk_non_existing_region() -> None:
...
 
 
-# pylint: disable=fixme
 @pytest.mark.django_db
 def test_summ_ai_bulk_disabled_region(load_test_data: None) -> None:
     """
     Ensure that calling when disabled in a region throws an error
     """
+    # pylint: disable=fixme
     # TODO: Ensure there are no race conditions with tests.summ_ai_api.summ.ai_test module
```
JoeyStk, line 62: Do we still need this comment? Is this maybe related to this issue? https://github.com/digitalfabrik/integreat-cms/issues/2119

JoeyStk (approved):
Thank you very much! This must have been very tedious and boring! So thank you for doing this for us!

I think as far as your job went, you did a good job. My only thoughts go beyond what was asked for this PR.

- Most disables are for unsued-argument. AFAIK there is no way to circumvent this. So right now I'm wondering if this rule really helps us enough to not disable it globally. But maybe it does...
- I think some of the disables should be refactored and then removed. How do we want to go about this? Create issues for it or just hope that in the future every time we see a disable we remember to think about whether or not we can refactor? :)

MizukiTemma:
@JoeyStk 

> * Most disables are for unsued-argument. AFAIK there is no way to circumvent this. So right now I'm wondering if this rule really helps us enough to not disable it globally. But maybe it does...

I'm actually for disabling `unused-argument` globally. Let's do a poll in the team channel 👀 

> * I think some of the disables should be refactored and then removed. How do we want to go about this? Create issues for it or just hope that in the future every time we see a disable we remember to think about whether or not we can refactor? :)

Yes, I saw chances of refatoring but wanted to go less-invasive in thie PR, as the goal of issue is putting disables into better positions and refactoring is always prone to bugs (and therefore reviewed indivisually focused). I'll create a generall issue for this thema.

integreat_cms/api/v3/social_media_headers.py
```diff
@@ -236,6 +239,7 @@ def page_social_media_headers(
     ):
         raise Http404("Page not found in this region with this language.")
 
+    # pylint: disable=fixme
     # TODO: add breadcrumb json-ld if content_translation exists
```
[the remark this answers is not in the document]
MizukiTemma: I'll leave it to #3211. I don't know the background for those disabled and am afraid of disturbing communication 🙈

MizukiTemma:
I'd like to suggest not to put any more effort on refactoring/improvement as we are going to remove pylint (#3234 ) but merge this PR as there will be still time until ruff starts to patrol our repository.

charludo (approved):
Oof.... those are a lot of occurrences! :D

Thank you for your effort 💛

Issue #3183, opened 2024-11-07
500 error when executing bulk action "ignore" in link checker


osmers:
### Describe the Bug
Executing bulk action ignore in link checker leads to 500 error - the check again bulk action works fine.

### Steps to Reproduce

1. Go to https://admin.integreat-app.de/testumgebung/analytics/linkcheck/invalid/
2. Check the box for any link
3. Select bulk action "ignore"
4. Click execute
5. See error

### Expected Behavior
Link gets added to ignore list

### Actual Behavior
500 error

PR #3096, opened 2024-10-02
Add tests for broken link actions

MizukiTemma:
### Short description

This PR adds tests for actions users can take in the broken link list.

### Proposed changes

- Do each action and check whether the target URLs/links are changed as expected.
- Check whether actions taken in the netowork management are affecting all the targets system-wide, while actions within a region changing on the objects that belong to the region.

### Side effects

- Two assersions are disabled, as they are failling somehow (new url/link objects are not created after action). So far as I observed, both search&replace and URL replace are working fine. Probably a special care is needed for the link check listener?

### Resolved issues

Fixes: #3062 

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #3186, opened 2024-11-07
Fix error at bulk link ignore/unignore

MizukiTemma:
### Short description

This PR fixes the bug which occurs when URLs are ignored or unignored by bulk action.

### Proposed changes

- Add `prefetch_region_links=True`

Alternatively we could probably set `prefetch_region_links` as default `True`

### Side effects

- I hope none 🙈 
- Tests will come in #3096 

### Resolved issues

Fixes: #3183 


david-venhoff (approved):
Thanks a lot for the quick fix!

I think I accidentally introduced this in #3120, because I thought the region links would not be required for this query.

MizukiTemma:
> Thanks a lot for the quick fix!
> 
> I think I accidentally introduced this in #3120, because I thought the region links would not be required for this query.

And neither I noticed it while reviewing 🙈

lunars97 (approved):
Works now as expected! :rocket: Thank you :)

Issue #3185, opened 2024-11-07
Add select all checkbox to translations for events


osmers:
### Motivation
For pages we already have a checkbox to check all languages for translation at one - we need this for events as well.

### Proposed Solution
Add the same checkbox.

### Alternatives
None.

### User Story
As a CMS user I want to be able to quickly select all languages in order to save time while translating.

Add this checkbox: 
![image](https://github.com/user-attachments/assets/03fde1bc-e63d-40e4-a8bd-2a77c9d7c1ea)

osmers:
Yeah, I think anywhere, where content can be translated :)

PR #3188, opened 2024-11-07
Extend select all checkbox to events and pois

JoeyStk:
### Short description

This PR extends the "select all" checkbox for translations to events and pois

### Proposed changes

- Add checkbox to events and pois

### Side effects

None?

### Resolved issues

Fixes: #3185

### DIFF
--- a/integreat_cms/cms/templates/events/event_form_sidebar/minor_edit_box.html
+++ b/integreat_cms/cms/templates/events/event_form_sidebar/minor_edit_box.html
@@ -44,6 +44,12 @@
                         {{ event_translation_form.mt_translations_to_create.label }}
                     </label>
                     {% render_field event_translation_form.mt_translations_to_create %}
+                    <div class="mt-4 font-bold">
+                        <input id="select-all-languages" type="checkbox" />
+                        <label for="select-all-languages">
+                            {% translate "Select all languages" %}
+                        </label>
+                    </div>
                 {% endif %}
             </div>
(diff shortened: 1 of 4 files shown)


lunars97 (commented):
Everything looks good :+1: 
But I have one question: Do we need it here as well? To uncheck, for example, all languages? 
image

JoeyStk:
Yes. I think we need it there too. Where is this? Is this the imprint? :)

lunars97:
> Yes. I think we need it there too. Where is this? Is this the imprint? :)

it is where you implemented, but after automatically translating them and checking it again you see there no option anymore to uncheck all of them...

JoeyStk:
> It is where you implemented, but after automatically translating them and checking it again you see there no option anymore to uncheck all of them...

Ah, now I get it :) I don't know, this is probably a question for @osmers :)

osmers:
What would be the use case in unchecking them all? When implementing it we decided to leave already translated languages checked bcs the assumption is, that if a translation exists, they will want to translate the changes as well, right?

JoeyStk:
> What would be the use case in unchecking them all? When implementing it we decided to leave already translated languages checked bcs the assumption is, that if a translation exists, they will want to translate the changes as well, right?

Yes, I think you're right. I don't think we need to uncheck the languages in bulk then :)

MizukiTemma (approved):
Looks good 👍

Issue #2588, opened 2023-12-12
Improve Link Checker GUI


svenseeberg:
### Motivation
Currently the broken link checker lists links as "broken". This is not always correct. Architecturally, the link checker cannot definitely determine if a link does not work in a browser.

The broken link checker supports editors to find potentially broken links.

### Proposed Solution
The GUI of the broken link checker should be less "definitive" when proclaiming URLs as broken. It should be clear that links need to be checked manually and that it is (most likely) safe to ignore broken links, if they work for the editor.

### Alternatives
Many support questsions?

### Additional Context
- See discussion in https://chat.tuerantuer.org/digitalfabrik/threads/jhqyq1hd57g65qw1617xzn8ria

### Design Requirements
- N/A

osmers:
Effort low since it only concerns renaming the categories.
For UI/UX @nikolahoff @hauf-toni I want to point out that we have a special case where links on the computer work but in the app they don't. This concerns internal links that were moved to a new location. Not sure what to do with them or how to make it clear to users that even though they appear to be working on the computer, they most likely need to be replaced. 

As discussed in the CMS/UI/UX/Service Call today, we'll also open another issue to rethink the design of the link checker to make it more intuitive.

nikolahoff:
@osmers here my proposals for the wordings.
Invalid > Verification needed (i.e. with alert icon, indicating the user that action is required)
Ignored > Link verified (i.e. check-icon indicating the link is ok)

osmers:
The service team approved the namings - in German we'd suggest: Überprüfung notwendig and Link überprüft

Question to the CMS team - does it make sense to move verified links to the general valid list? And then they get checked again once the link checker rechecks links? Or does it recheck links too often for that to be useful?
I'm asking bcs as far as I understood, the links under "Ignore" are currently not re-checked unless done manually, right? And they might break at some point but noone would notice...

MizukiTemma:
@osmers 
There is a bulk action "Ignore" and "Unignogre". Should this be renamed for example to "Mark as velified", "Mark as not vverified anymore"?

osmers:
@MizukiTemma seems reasonable :) mark as verified is good - maybe to move it to the other list, something like recheck link? And that automatically removes it from the list and then depending on the outcome of the check adds it to valid or verification needed?

MizukiTemma:
[...]

PR #3191, opened 2024-11-08
Improve link checker GUI

MizukiTemma:
### Short description

This PR updates some wording on the broken link list to make the GUI more intuitive for users.

### Proposed changes

- Link category "Invalid (Ungültig)" is now labelled as "Verification needed (Überprüfung notwendig)"
- Link category "Ignored (Ignoriert)" is now labelled as "Link verified (Link überprüft)"
- Icons are added to the labellings (In the issue only ⚠️ for "Verification needed" and ✅ for "Link verified" are specified but other categories have received an icon for consistent view) 
- Bulk action "Ignore" is now called "Mark as verified", "Unignore" is "Bring back to check"

### Side effects

- Renaming is done only on the front end: in backend they are still "ignore", "unignore", "ignored" (I guess this is intuitiver for developers because the related boolean field of the link model is named "ignore")
- The new name for the bulk action "unignore" doesn't sound great, I know. This was to avoid naming it "Recheck", resulting in having two bulk actions with the same name.

----- Technical background of "unignore" and "recheck" -----
A link has status/type "Valid", "Invalid (Verification needed)", "Email", or "Phone" and a label (marking) "ignore", which is true or false. If "ignore"=True, the link is handled as "Ignored (verified)" regardless of "valid", "invalid", E-mail", "Phone". "Recheck" action scans a link again to check whether it is valid or invalid, while "Unignore" is removing "ignored" label from the link (by setting it to false). Without removing "ignored" marking, the link stays in the list "Link verified (ignored)". In such way those two actions have different meaning, and it looks better to name them with different names.

### Resolved issues

Fixes: #2588

### DIFF
--- a/integreat_cms/cms/templates/linkcheck/links_by_filter.html
+++ b/integreat_cms/cms/templates/linkcheck/links_by_filter.html
@@ -31,28 +31,28 @@
     </div>
     <a href="{% if view.kwargs.url_filter == 'valid' %}#{% else %}{% url_for_current_region 'linkcheck' request url_filter='valid' %}{{ pagination_params }}{% endif %}"
        class="pr-2 {% if view.kwargs.url_filter == 'valid' %}text-black font-bold cursor-default{% else %}bg-transparent hover:underline text-blue-500 hover:text-blue-600 cursor-pointer{% endif %}">
-        {% translate "Valid" %}
+        ✅ {% translate "Valid" %}
         <span class="{% if view.kwargs.url_filter == 'valid' %}text-gray-500 font-normal{% endif %}">
             ({{ number_valid_urls }})
         </span>
     </a>
     <a href="{% if view.kwargs.url_filter == 'invalid' %}#{% else %}{% url_for_current_region 'linkcheck' request url_filter='invalid' %}{{ pagination_params }}{% endif %}"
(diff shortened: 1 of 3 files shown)


integreat_cms/cms/templates/linkcheck/links_by_filter.html
```diff
@@ -173,7 +173,7 @@ <h1 class="heading">
                 {% if view.kwargs.url_filter == 'ignored' %}
                     <option value="unignore"
                             data-bulk-action="{% url_for_current_region 'linkcheck' request url_filter=view.kwargs.url_filter %}{{ pagination_params }}">
-                        {% translate "Unignore" %}
+                        {% translate "Bring back to check" %}
```
david-venhoff, line 176: I think this could be confusing, because it is possible to mark e.g. valid links as ignored. unignoring them will cause them to end up at the verified tab, not the unchecked tab.

Maybe we could label it "invalidate" or "revoke"?

david-venhoff (commented):
Thanks, nice UI improvement :)

integreat_cms/cms/templates/linkcheck/links_by_filter.html
MizukiTemma: Nice naming 👍

osmers: Hmm - I find that quite difficult tbh...maybe "revoke verification"? To make it clearer what we are revoking? And in German "Verifizierung entziehen"? Or "Verifizierung rückgängig machen"?

JoeyStk (approved):
Thank you very much :)

Issue #3088, opened 2024-10-01
[2024-11-07] Add a button to create a contact in the POI form.


MizukiTemma:
### Motivation

A new feature [contact](https://github.com/digitalfabrik/integreat-cms/issues/2955) is being implemented. It helps users if there is a possibility to create a contact directly in the POI form.

### Proposed Solution

Implement an ajax form like POI creation in the event form.

### Alternatives

Don't implement.

PeterNerlich:
> Is the relationship between POI/Location and Contact a 1:0..1 relationship?

No, Location and Contact are in a 0..1:0..n relationship: A Contact MAY reference a Location, there can be Locations that are not referenced by any Contact. But there is an additional constraint on the model that there can be at most one Contact with an empty title / *point of contact for* field per referenced Location

jarlhengstmengel:
> So a contact can be created without referencing a location?

NVM

PeterNerlich:
No, that's my mistake, I think a contact always has to reference a location after all

jarlhengstmengel:
@osmers the question came up if the creation of a contact in the POI form should only be available during the editing of  a POI or also during the creation?

osmers:
@jarlhengstmengel how are we handling that with events? I think I would implement the same process here.

jarlhengstmengel:
> @jarlhengstmengel how are we handling that with events? I think I would implement the same process here.

Events is doing both I think. Contact creation during POI creation is a bit more complicated as the creation during editing since a contact needs to reference a POI, but it should be doable. I will probably split this issue into two PRs then

osmers:
@jarlhengstmengel that's a fair point though - I hadn't thought of this...
We could do it this way:
1st Prio - contact creation once POI is already saved
2nd Prio - contact creation before POI is saved (not really that urgent, I think, especially if it would delay the release. I think it's always an improvement we can release later)
What do you think?

jarlhengstmengel:
Sounds good to me :)

PR #3196, opened 2024-11-11
Add contact creation in POI form

jarlhengstmengel:
### Short description

This PR adds the creation of contacts with a collapsible box in the POI form.

### Proposed changes

- Add collapsibele box with a contact form in POI form
- Add view for the box
- Trigger creation of a contact with clicking at publish and setting the location automatically to the POI that is created parallel

### Side effects

- None

### Resolved issues

Fixes: #3088 

### DIFF
--- a/integreat_cms/cms/templates/pois/poi_form.html
+++ b/integreat_cms/cms/templates/pois/poi_form.html
@@ -191,6 +191,12 @@
             </div>
         </div>
     </form>
+    <form id="ajax_contact_form"
+          name="ajax_contact_form"
+          method="post"
+          enctype="multipart/form-data"
+          data-unsaved-warning>
+    </form>
     {{ media_config_data|json_script:"media_config_data" }}
     {% if not perms.cms.change_poi or poi_form.instance.id and poi_form.instance.archived %}
         {% include "../_tinymce_config.html" with readonly=1 %}
--- a/integreat_cms/cms/templates/pois/poi_form_sidebar/related_contacts_box.html
+++ b/integreat_cms/cms/templates/pois/poi_form_sidebar/related_contacts_box.html
@@ -17,12 +17,38 @@
                     {% trans "This location is not currently referred to in any contact." %}
                 {% endif %}
             </div>
-            {% for contact in contacts %}
-                <a href="{% url 'edit_contact' contact_id=contact.id region_slug=request.region.slug %}"
-                   class="block pt-2 hover:underline">
-                    <i icon-name="pen-square" class="mr-2"></i> {{ contact.label_in_reference_list }}
-                </a>
-            {% endfor %}
+            <div id="related-contact-list">
+                {% for contact in contacts %}
+                    <a href="{% url 'edit_contact' contact_id=contact.id region_slug=request.region.slug %}"
+                       class="block pt-2 hover:underline">
+                        <i icon-name="pencil" class="mr-2"></i> {{ contact.label_in_reference_list }}
+                    </a>
+                {% endfor %}
+            </div>
         </div>
+        {% if perms.cms.change_contact %}
+            <div>
+                <div class="help-text mt-4 font-bold">
+                    <div>
+                        <button id="show-contact-form-button"
+                                form="ajax_contact_form"
+                                class="btn w-full mt-4 mb-2"
+                                data-url="{% url 'show_contact_form_ajax' region_slug=request.region.slug poi_id=poi_id %}">
+                            {% translate "Create a new contact" %}
+                        </button>
+                    </div>
+                </div>
+                <div id="contact-ajax-success-message"
+                     class="bg-green-100 border-l-4 border-green-500 text-green-800 px-4 py-3 hidden">
+                    {% trans "The new contact was successfully created." %}
+                </div>
+                <div id="contact-ajax-error-message"
+                     class="bg-red-100 border-l-4 border-red-500 text-red-700 px-4 py-3 hidden">
+                    {% trans "An error occurred." %}
+                </div>
+                <div id="contact-form-widget">
+                </div>
+            </div>
+        {% endif %}
     {% endwith %}
 {% endblock collapsible_box_content %}
--- a/integreat_cms/cms/views/contacts/contact_form_ajax_view.py
+++ b/integreat_cms/cms/views/contacts/contact_form_ajax_view.py
@@ -1,0 +1,89 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from django.http import JsonResponse
+from django.shortcuts import render
+from django.utils.decorators import method_decorator
+from django.views.generic import TemplateView
+
+from ...decorators import permission_required
+from ...forms import ContactForm
+from .contact_context_mixin import ContactContextMixin
+
+if TYPE_CHECKING:
+    from typing import Any
+
+    from django.http import HttpRequest, HttpResponse
+
+
+@method_decorator(permission_required("cms.view_contact"), name="dispatch")
+@method_decorator(permission_required("cms.change_contact"), name="post")
+class ContactFormAjaxView(TemplateView, ContactContextMixin):
+    """
+    View for the ajax contact widget
+    """
+
+    template = "ajax_contact_form/_contact_form_widget.html"
+
+    def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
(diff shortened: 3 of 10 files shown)


integreat_cms/cms/templates/pois/poi_form.html
```diff
@@ -167,6 +167,9 @@ <h3 class="font-bold text-black heading">
                 <div id="right-sidebar-column"
                      class="flex flex-col flex-wrap 3xl:col-end-3 4xl:col-end-auto md:w-full">
                     {% include "./poi_form_sidebar/contact_box.html" with box_id="poi-contact" %}
+                    {% if not poi_form.instance.id and perms.cms.change_contact %}
```
MizukiTemma, line 170:
```suggestion
                    {% if poi_form.instance.id and perms.cms.change_contact %}
```
I guess you want to show this box only when an existing POI is being edited, not when a new POI is being created on the POI form.

MizukiTemma:
The new typescript file "contact_box.ts" needs to be added in "index.ts", otherwise it can't bring its effects 😿

jarlhengstmengel:
> The new typescript file "contact_box.ts" needs to be added in "index.ts", otherwise it can't bring its effects 😿

Thank you for the hint! That would have been my question.

MizukiTemma:
@jarlhengstmengel 
I'll join to implementation of this PR ➕

MizukiTemma:
This PR implements ony the first step ([see this](https://github.com/digitalfabrik/integreat-cms/issues/3088#issuecomment-2488402054))

The second step will be a separate issue (#3322 )

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -7802,6 +7805,14 @@ msgstr "Dieser Ort wird in folgenden Kontakten verwendet."
 msgid "This location is not currently referred to in any contact."
 msgstr "Zur Zeit gibt es keine Kontakte zu diesem Ort."
 
+#: cms/templates/pois/poi_form_sidebar/related_contacts_box.html
+msgid "Create a new contact"
+msgstr "Kontakt erstellen"
+
+#: cms/templates/pois/poi_form_sidebar/related_contacts_box.html
+msgid "The new contact was successfully created."
+msgstr "Ein neuer Kontakt wurde erfolgreich erstellt."
```
JoeyStk, line 7814:
```suggestion
msgstr "Der neue Kontakt wurde erfolgreich erstellt."
```
This is very nit picky, but I think we should keep the definitive article to stay consistent to the English version, but yes, it doesn't sound as good

integreat_cms/static/src/index.ts
```diff
@@ -104,6 +104,8 @@ import "./js/menu";
 import "./js/poi-categories/poicategory-colors-icons";
 import "./js/dashboard/broken-links";
 
+import "./js/contact_box";
+
```
JoeyStk, line 108:
```suggestion
```

integreat_cms/cms/templates/pois/poi_form_sidebar/related_contacts_box.html
```diff
@@ -17,12 +17,38 @@
...
-                <a href="{% url 'edit_contact' contact_id=contact.id region_slug=request.region.slug %}"
-                   class="block pt-2 hover:underline">
-                    <i icon-name="pen-square" class="mr-2"></i> {{ contact.label_in_reference_list }}
-                </a>
-            {% endfor %}
+            <div id="related-contact-list">
+                {% for contact in contacts %}
+                    <a href="{% url 'edit_contact' contact_id=contact.id region_slug=request.region.slug %}"
+                       class="block pt-2 hover:underline">
+                        <i icon-name="pen-square" class="mr-2"></i> {{ contact.label_in_reference_list }}
```
JoeyStk, line 24:
```suggestion
                        <i icon-name="pencil" class="mr-2"></i> {{ contact.label_in_reference_list }}
```
At least for the list views we changed the icon to the pencil, but maybe it's okay for this view because the context is slightly different? (And also: we still use this icon still in a few places)

JoeyStk (commented):
Thank you a lot to the both of you! 
This is a really good step forward. I have a few questions, but other than those it looks already very good :)

MizukiTemma:
@JoeyStk 
Thank you for the review 😸 Your suggestion are applied 👍

JoeyStk (approved):
Thank you very much!

integreat_cms/cms/views/contacts/contact_form_ajax_view.py
```diff
@@ -0,0 +1,94 @@
...
+
+        :param request: The current request
+        :param \*args: The supplied arguments
+        :param \**kwargs: The supplied keyword arguments
+        :raises ~django.http.Http404: If no language for the given language slug was found
+
+        :return: A status message, either a success or an error message
+        """
+
+        contact_instance = Contact.objects.filter(id=None).first()
```
charludo, line 66: This will always be `None`, right? There cannot be an object without an ID.

(I assume this was copied from the `POIAjaxFormView`? I think it is an error there, too)

integreat_cms/static/src/js/contact_box.ts
charludo, line 1: Could you maybe rename this to something more expressive? 😅

integreat_cms/static/src/js/contact_box.ts
```diff
@@ -0,0 +1,106 @@
...
+            successMessageField.classList.add("hidden");
+        }, timeoutDuration);
+    } else {
+        const errorMessageField = document.getElementById("contact-ajax-error-message");
+        errorMessageField.classList.remove("hidden");
+        setTimeout(() => {
+            errorMessageField.classList.add("hidden");
+        }, timeoutDuration);
+    }
+};
```
charludo, line 41: This is cool, what do you think about extracting this into its own PR and applying it globally? 😃

integreat_cms/static/src/js/contact_box.ts
```diff
@@ -0,0 +1,106 @@
...
+    const response = await fetch(btn.getAttribute("data-url"), {
+        method: "POST",
+        headers: {
+            "X-CSRFToken": getCsrfToken(),
+        },
+        body: formData,
+    });
+
+    const data = await response.json();
+    console.debug(data);
```
charludo, line 62:
```suggestion
```
same here :D

charludo (changes requested):
THank you, this already looks great! 🎉 

One question: why is the option to create a new contact only shown for existing POIs? IMO it would be very convenient to also be able to create a contact at the same time as the POI.

Additionally, I am a bit confused (and I suspect some users might be too) about the intended purpose of the "related contact" vs the "contact information". I think some clarification, or better yet some unification of the features, is necessary.

For example, could a checkbox be added below the "contact information" box to create a new contact based on that information? Then *locations* would immediately become linkable in pages/events, like @nikolahoff suggested in the last weekly.

Similarly, maybe a checkbox "set as default contact information for this location" could be added in the new ajax form this PR adds, which would then auto-fill the contact information of the POI, if it was not set beforehand?

I know this is somewhat beyond the scope of this PR, I just feel that with the addition of this feature, the POI form with its multiple overlapping ideas of contacts/contact details becomes rather confusing. Maybe @osmers also has an opinion on this? 😅

osmers:
@charludo I was just gonna mention #3180 and definitely agree with your input on being able to create a contact from the POI form :) I think that might be #3088 ?

MizukiTemma:
@osmers  #3322 is probably better place to leave comments as it's to be implemented in the next weeks.

charludo:
> @osmers #3322 is probably better place to leave comments as it's to be implemented in the next weeks.

I think I have not had enough coffee yet -.-
Sorry about the noise about things already solved in this thread...

integreat_cms/static/src/js/contact_box.ts
MizukiTemma: uff, any suggestion? 😅

integreat_cms/cms/views/contacts/contact_form_ajax_view.py
MizukiTemma: You're right 👀 😁 Fixed. The one in `POIAjaxForm` will be fixed as #3359 (as not in the scope of this PR)

MizukiTemma:
@charludo 
Thank you for review 😸 Could you check again?

charludo (approved):
Thank you, looks great!

Issue #3200, opened 2024-11-12
Internal Server Error when creating a new language tree node


juliankehne:
### Describe the Bug

Internal Server Error after trying to create a new language node. This happens only in regions for which content was duplicated from another region and translations were not duplicated using the new functionality introduced in #2991 

### Steps to Reproduce

1. Go to 'Netzwerkverwaltung'
2. Go to 'Regions'
3. Click 'Create Region'
4. Select any region to duplicate content from
5. Uncheck the checkbox to copy languages and translations
6. Create the region
7. Go to the region dashboard
8. Open the language tree
9. Add a new language node
10. See error

### Expected Behavior

The new language node is created. No error.

### Actual Behavior

No new language node is created. Internal Server Error.

### Additional Information

Only happens for regions where the checkbox is unchecked. Existing regions, new regions where no content was copied or all translations were copied work as expected.

<details>
  <summary>Traceback</summary>
Environment:

Request Method: POST
Request URL: http://localhost:8000/geklontes-augsburg/language-tree/new/

Django Version: 4.2.16
Python Version: 3.11.10
Installed Applications:
['integreat_cms.cms',
 'integreat_cms.core',
 'integreat_cms.deepl_api',
 'integreat_cms.google_translate_api',
 'integreat_cms.firebase_api',
 'integreat_cms.gvz_api',
 'integreat_cms.matomo_api',
 'integreat_cms.nominatim_api',
 'integreat_cms.summ_ai_api',
 'integreat_cms.textlab_api',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.humanize',
 'django.contrib.messages',
 'django.contrib.sessions',
 'django.contrib.sitemaps',
 'django.contrib.staticfiles',
 'corsheaders',
 'db_mutex',
 'import_export',
 'linkcheck',
 'polymorphic',
 'rules.apps.AutodiscoverRulesConfig',
 'treebeard',
 'webpack_loader',
 'widget_tweaks',
 'cacheops',
 'django.contrib.admin',
 'debug_toolbar']
Installed Middleware:
['django.middleware.gzip.GZipMiddleware',
 'debug_toolbar.middleware.DebugToolbarMiddleware',
 'corsheaders.middleware.CorsMiddleware',
 'django.middleware.security.SecurityMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.locale.LocaleMiddleware',
[...]

PR #3203, opened 2024-11-13
Fix copying of language tree nodes in the region form

david-venhoff:
### Short description

When creating a new region which copies contents from another region and only the root languages should be included, the `lft` and `rgt` values of the new language node must be set to `1` and `2`, respectively. Otherwise, it would behave as if it still contained child nodes.

### Proposed changes

- Set the correct `lft` and `rgt` values

### Side effects

- Should be none

### Resolved issues

Fixes: #3200


jarlhengstmengel (approved):
Looks good, thank you!

MizukiTemma (approved):
Thank you so much 🚀

Issue #3110, opened 2024-10-04
[2024-11-07] Recurring events are not imported correctly


JoeyStk:
### Describe the Bug

When importing recurring events only the original event is shown in the Events page, marked as a One-Time. All types of recurring events are affected, daily, weekly, etc.

### Steps to Reproduce

1. Go to Nextcloud calendars and create a new recurring event.
2. Import calendar through the iCal 
3. See event is imported as a one-time event, not as a recurring event.

### Expected Behavior

Event should be correctly imported as a recurring event.

### Actual Behavior

Event is wrongfully imported as one-time event.

JoeyStk:
(Partially) blocked. The refactoring can be done anytime already

david-venhoff:
In todays meeting, we decided against refactoring the recurrence rule model, because it would be too much work. Instead, we will make a best effort to map ical rrules to our existing recurrence rule model, and fail if that is not possible. 

For future reference, a (very wip) branch attempting to rewrite the recurrence rule model can be found here: https://github.com/digitalfabrik/integreat-cms/tree/refactor/recurrence_rule

PR #3205, opened 2024-11-13
Ical: Add support for recurrence rules

david-venhoff:
### Short description

This pr adds support for importing ical events with recurrence rules. 

### Proposed changes

- Parse the recurrence rule and try to convert it into a dictionary that can be passed to the `RecurrenceRuleForm`
- I have implemented this a strict as possible: Any recurrence rules which contains any values that are not explicitly supported right now will cause the entire event import to fail.
- Alternatively, it would be possible to make a *best effort*, so that when receiving an ical rrule that we don't understand, the event will still be imported, but potentially without recurrence rule or with a different recurrence rule. However then we would not easily find out which rules are not supported and have a higher chance of wrongly imported events, so I decided against that

### Side effects

- External calendars that previously did not have any error might now show errors on import, since the rrule is not ignored anymore
- During development, I only tested this with nextcloud generated calendars. It would be helpful if somebody could test this with calendars generated from other programs

### Resolved issues

Fixes: #3110


integreat_cms/cms/utils/external_calendar_utils.py
```diff
@@ -245,6 +406,10 @@ def import_event(
         if previously_imported_event_translation
         else None
     )
+    previously_imported_recurrence_rule = RecurrenceRule.objects.filter(
```
JoeyStk, line 409: Why do we call it "previously_"? :) What is the context of this? :)

JoeyStk (commented):
Thank you very much!
This looks mostly good to me!
~~Beside my question I only noticed, that the multi select field for weekdays is not disabled for imported events and I was able to change the value. But since I couldn't save this change anyways, it didn't do any harm :) ~~

integreat_cms/cms/utils/external_calendar_utils.py
david-venhoff: I named it just like `previously_imported_event_translation` and `previously_imported_event` defined in the lines above, to indicated that this is the recurrence rule that was previously imported in this function (For example in the last run of the import_events management command). 
This is needed for the forms, e.g. so that they can check if it has changed through this import.

david-venhoff:
@PeterNerlich 
> Would that be a suitable alternative?

Yes, definitely. In fact, I first tried to solve this issue by refactoring our recurrence rule model to only store the ical rrule string. However, that turned out to be more development effort than the approach taken in this pr, and included some ugly hacks to work with the rest of the cms. For example, I could not get rid of the `recurrence_end_date` field of the recurrence rule model, because it is  [used](https://github.com/search?q=repo%3Adigitalfabrik%2Fintegreat-cms+recurrence_rule__&type=code) in some queries. 
I am quite unsure which is the best approach though :(

(For reference, this is the wip branch of this alternate approach: [refactor/recurrence_rule](https://github.com/digitalfabrik/integreat-cms/tree/refactor/recurrence_rule))

integreat_cms/cms/utils/external_calendar_utils.py
JoeyStk: Ah, okay, now I understand :) Thanks :)

JoeyStk (approved):

Issue #3050, opened 2024-09-14
Add tests for region deletion


david-venhoff:
### Motivation

Deleting region can sometimes cause internal server errors. To catch those before they get into production we should write a test

### Alternatives

None

PR #3208, opened 2024-11-14
Test for region deletion

JoeyStk:
### Short description

This PR adds a test for region deletion

### Proposed changes

- Test to delete all regions
- Test that all regions which are mirrored can't be deleted

### Side effects

- There might be some?

### Resolved issues

Fixes: #3050


tests/cms/views/regions/delete_region.py
```diff
@@ -0,0 +1,84 @@
...
+
+        if role not in [CMS_TEAM, SERVICE_TEAM, ROOT] and not ANONYMOUS:
+            assert response.status_code == 403
+            continue
+
+        if role in [CMS_TEAM, SERVICE_TEAM, ROOT]:
+            assert response.status_code == 302
+            redirect = response.headers.get("location")
+            response = client.get(redirect)
+
```
MizukiTemma, line 47:
```suggestion
            assert "Region wurde erfolgreich gelöscht" in response.content.decode(
                "utf-8"
            )

```

We can check the message here too 😸

tests/cms/views/regions/delete_region.py
```diff
@@ -0,0 +1,84 @@
...
+def test_delete_all_regions_is_successful(
+    load_test_data: None,
+    login_role_user: tuple[Client, str],
+) -> None:
+    current_number_of_regions = Region.objects.count()
+    client, role = login_role_user
+
+    # Deletion of Augsburg will not be possible due to a cloned page in Nurnberg.
+    # We want to test this is in the test below. In this test we want to make sure this is the only reason why deleting is not possible.
+    cloned_page = Page.objects.filter(id=30)
```
MizukiTemma, line 21:
```suggestion
    cloned_page = Page.objects.filter(id=CLONED_PAGE)
```

It's not critical, but we can probably  `CLONED_PAGE = 30`? as this id number is used in the next test too and to avoid magic number. Or even `CLONED_PAGES` and `id__in=CLONED_PAGES` in case another mirrored page is added.

MizukiTemma (changes requested):
Thank you for adding test 💪 I have some suggestions, please find in the comments 💬

tests/cms/views/regions/delete_region.py
```diff
@@ -0,0 +1,84 @@
...
+            assert response.status_code == 403
+            continue
+
+        if role in [CMS_TEAM, SERVICE_TEAM, ROOT]:
+            assert response.status_code == 302
+            redirect = response.headers.get("location")
+            response = client.get(redirect)
+
+    if role in [CMS_TEAM, SERVICE_TEAM, ROOT]:
+        assert Region.objects.count() == 0
```
MizukiTemma, line 49: Could you check also there is no users and push notifications that are orphaned?

david-venhoff:
Do you remember when deleting a region caused internal server errors? Ideally we should catch these with the new tests

JoeyStk:
> Do you remember when deleting a region caused internal server errors? Ideally we should catch these with the new tests

These are the three Issues/PRs that I could find: 
- https://github.com/digitalfabrik/integreat-cms/issues/1749 & https://github.com/digitalfabrik/integreat-cms/pull/1778
- https://github.com/digitalfabrik/integreat-cms/issues/653 & https://github.com/digitalfabrik/integreat-cms/pull/654
- https://github.com/digitalfabrik/integreat-cms/issues/2462 & https://github.com/digitalfabrik/integreat-cms/pull/2506

I will add commits in the next days that also test for these use cases :)

JoeyStk:
As I'm busy with a lot of other issues/projects please feel free to contribute to this PR. I don't see myself going back to this PR anytime soon :(

MizukiTemma (approved):

tests/cms/views/regions/delete_region.py
```diff
@@ -0,0 +1,122 @@
...
+
+    if role not in [CMS_TEAM, SERVICE_TEAM, ROOT] and not ANONYMOUS:
+        assert response.status_code == 403
+
+    if role in [CMS_TEAM, SERVICE_TEAM, ROOT]:
+        assert response.status_code == 302
+        redirect = response.headers.get("location")
+        response = client.get(redirect)
+        assert (
+            "Die Region konnte nicht gelöscht werden, weil die folgenden Seiten in anderen Region gespiegelt werden:"
```
jarlhengstmengel, line 119: There is a typo here with "Region" instead "Regionen", but it's also in the translation file so I think it's probably just copied from there?

jarlhengstmengel (approved):
Thanks! Looks good to me :)

tests/cms/views/regions/delete_region.py
MizukiTemma: Hmm, I think singular is correct here, as `delete_region` delets actually one single region 👀

david-venhoff:
```suggestion
            "Die Region konnte nicht gelöscht werden, weil die folgenden Seiten in anderen Regionen gespiegelt werden:"
```
I think @jarlhengstmengel was referring to the second occurrence, which should actually be plural.

MizukiTemma: @jarlhengstmengel @david-venhoff 
Ah 🙈 Yes, it must be "Regionen" there. #3302

PR #3209, opened 2024-11-15
Fix comparison to zero

JoeyStk:
### Short description

While reviewing #3182 I saw this comparison and don't think we need to disable pylint for it. Therefore I did a quick refactoring, so we can remove the disable of pylint

### Proposed changes

- Remove comparison to zero

### Side effects

- Should be none, but let's double- and triple-check

### Resolved issues

Fixes: -

### DIFF
--- a/integreat_cms/cms/utils/internal_link_utils.py
+++ b/integreat_cms/cms/utils/internal_link_utils.py
@@ -28,7 +28,6 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-# pylint: disable=compare-to-zero
 def update_link(
     link: Element, target_language_slug: str
 ) -> Optional[tuple[str, Element | str]]:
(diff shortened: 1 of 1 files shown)


MizukiTemma (approved):
Thank you 😸 Our code quality is improved 👍

david-venhoff (changes requested):
I now get this warning: 
```
/home/user/dev/integreat-cms/integreat_cms/cms/utils/internal_link_utils.py:64: FutureWarning: The behavior of this method will change in future versions. Use specific 'len(elem)' or 'elem is not None' test instead.
  if not link and link.text and current_url.strip() == link.text.strip():
```

What this code really wants to test is that the link has no children, however doing `bool(link)` may in the future also return True if it contains text, for example, so we should not rely on it.
For reference, here is the origin of the warning: https://github.com/python/cpython/blob/e09359112e250268eca209355abeb17abf822486/Lib/xml/etree/ElementTree.py#L208

JoeyStk:
@david-venhoff Do you think that there is a fix for this or should I close the PR? :)

david-venhoff:
I think the current solution, as suggested in the warning, is already the best unfortunately :(

Issue #3213, opened 2024-11-18
Statistics in the Dashboard displays "a network error has occured" message


juliankehne:
### Describe the Bug

On the "My Dashboard" page the statistics box always displays this message " A network error has occurred. Please try again later. "

### Steps to Reproduce

1. Go to a region with statistics enabled
2. Click on 'My dashboard'
3. Scroll down to 'Statistics'
4. See message

### Expected Behavior

If the statistics loaded correctly no message should be shown.

### Actual Behavior

The graph displays data correctly and displays the error message.

### Additional Information

Currently happens on cms-test and admin. No error message on the statistics page itself, only on the dashboard.
![2024-11-18 09_33_12-Window](https://github.com/user-attachments/assets/c3778ae9-9c33-4314-aee8-cf3e0d594991)

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

PR #3220, opened 2024-11-18
 Fix false positiv error message in statistics in dashboard

MizukiTemma:
### Short description

This PR fixes the error messsage in the dashboard statistics that is shown despite the statistics is working correctly.
The cause is `TypeError` (propaties of null) . It triggers `catch (error)`.

### Proposed changes

- Avoid `TypeError`

### Side effects

- The change is held minimal to avoid side effects.

### Resolved issues

Fixes: #3213 


JoeyStk (approved):
Thank you for the quick fix :)

Issue #3227, opened 2024-11-19
External calendars created in outlook with recurring events cannot be imported


david-venhoff:
### Describe the Bug

Outlook generates some attributes in the recurrence rule that we don't support right now, which causes the import to fail.

### Steps to Reproduce

1. Add external calendar with url https://github.com/user-attachments/files/17816539/outlook.calendar.txt
3. See errors on import

### Expected Behavior

No errors

### Actual Behavior

errors

### Related Issues

Discovered here: https://github.com/digitalfabrik/integreat-cms/issues/3110#issuecomment-2485876203

david-venhoff:
The remaining rrules not fixed by #3229 contain weekdays with frequencies, like `BYDAY=-1SU`. These are currently not really supported by the `icalendar` library. Because of that, I decided to block this issue on https://github.com/collective/icalendar/issues/749.

If we really want to fix these edge cases before, we can implement this ourselves.

david-venhoff:
Update: The pr at icalendar is now merged. Unfortunately, it contained a breaking change, so there is probably no update coming soon.

PR #3229, opened 2024-11-22
Fix some problems with outlook recurrence rule

david-venhoff:
### Short description

This pr fixes two problems that happen when importing external calendars that were generated in microsoft outlook.

### Proposed changes

- Handle the `WKST` property in recurrence rules (For now, the only supported weekstart is monday)
- Handle the `BYMONTHDAY` and `BYMONTH` properties in recurrence rules: They must match the month and day of the event, because we don't support recurrence rules which start on a different day than their event
- This pr does not solve #3227 fully, because rrules with weekdays like `-1SU` are still not handled. I will open another pr for that functionality.

### Side effects

- Less import errors

### Resolved issues

Fixes: partially #3227


david-venhoff:
Raising priority, as this pr blocks another pr now

charludo (approved):
WOW, thank you for the comprehensive tests!! 🎉 🎉 

TBH I was a bit unsure how to test this test PR, but I have confirmed that the new tests pass, and would not have passed with the previous implementation. Nice! 😁

MizukiTemma:
@david-venhoff 
Thank you for keep improving iCal feature 😻 

I tried with the URL provided in the issue description and still got an error:  `Could not import event due to unsupported recurrence rule: ValueError('by_set_pos must not be None')` Is this (a part of) something you mentioned in the description as not covered (to be solved in the follow PR)?

david-venhoff:
@MizukiTemma 

> Is this (a part of) something you mentioned in the description as not covered (to be solved in the follow PR)?

Exactly. The problem with these events is that the `by_set_pos` attribute is specified as part of the `by_day` attribute, but the cms in this pr still does not understand that and thinks that `by_set_pos` is missing. This will be solved by the follow-up pr.

> Though 20 event must be imported?

A lot of the events of this calendar can not be imported because the recurrence end date is on the same day as it starts, and the recurrence rule form currently throws an error when that happens. IMO that is fine and just an error with the events, not with the cms.

MizukiTemma:
> @MizukiTemma
> 
> > Is this (a part of) something you mentioned in the description as not covered (to be solved in the follow PR)?
> 
> Exactly. The problem with these events is that the `by_set_pos` attribute is specified as part of the `by_day` attribute, but the cms in this pr still does not understand that and thinks that `by_set_pos` is missing. This will be solved by the follow-up pr.
> 
> > Though 20 event must be imported?
> 
> A lot of the events of this calendar can not be imported because the recurrence end date is on the same day as it starts, and the recurrence rule form currently throws an error when that happens. IMO that is fine and just an error with the events, not with the cms.

Thank you for explanation 🙏 Then this PR is ready to go 🚀

MizukiTemma (approved):
We appreciate you've been tackling this complicated thema 🙏 


Don't forget to squash the commits before merge😸

Issue #3223, opened 2024-11-19
Event creation not working when 'no physical location' checkbox is selected


juliankehne:
### Describe the Bug

When creating a new event an choosing the 'Event does not have a physical location' option and then publishing event it does not create the event, shows and error message and changes the button from 'Publish' to 'Update'.

Afterwards clicking the 'Update' button creates the event without an error message.

### Steps to Reproduce

1. Go to 'Events'
2. Click on 'Create event'
3. Add mandatory information (name, date, etc.)
4. Check the 'Event does not have physical location' checkbox
5. Click 'Publish' button
6. See error message and wrong behaviour

### Expected Behavior

Event is created without error message.

### Actual Behavior

- An error message is shown 'Location: Either disable the event location or provide a valid location '
- No event is created
- The 'Publish' button changed to 'Update' button

### Additional Information

When choosing a physical location the creation works as expected.

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

MizukiTemma:
It seems to be the problem that "-1" is handed over as `location`.

Furthermore, you cannot remove the event location by clickling " Event does not have a physical location" once the event gets an POI assigned.

PR #3247, opened 2024-11-28
Fix event creation without physical location

MizukiTemma:
### Short description

This PR fixes the problem that events cannot be created without POI even if "Event does not have a physical location" is checked.

### Proposed changes

- <s>Clean the POI id before passing the user input values into the form.: the cause was "-1", which is passed as POI id if no POI is selected but causes validation error, resulting in the form is judged as invalid (before reaching `clean` method). </s>
- Apply the suggestion from @david-venhoff (see the comment below)
- Add a test for this case

### Side effects

- I hope none :see_no_evil: 
- 

### Resolved issues

Fixes: #3223 

### DIFF
--- a/integreat_cms/cms/templates/ajax_poi_form/_poi_address_container.html
+++ b/integreat_cms/cms/templates/ajax_poi_form/_poi_address_container.html
@@ -4,6 +4,7 @@
         {% translate "Address" %}
     </label>
     <input name="location"
+           id="id_location"
            class="hidden"
            value="{% if poi %}{{ poi.id }}{% else %}-1{% endif %}"
            readonly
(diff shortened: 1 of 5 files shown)


lunars97 (approved):
Everything looks good and works as expected! Thank you for fixing it :yellow_heart:

david-venhoff (changes requested):
Thanks, also for the tests!

Looks like the bug is that an element can not be found anymore. I get this console error when I click the "has not location" checkbox: 
image

The commit that introduced this bug is https://github.com/digitalfabrik/integreat-cms/commit/6fb7122bd3ef8d1579692722c2b5a320a389faac

There, the id field for the input was removed, which means that the js code that controls it cannot find it anymore.

The fix is to add this to `_poi_address_container.html` line 7: 
```
id="id_location
```

Then we don't need this extra code on the python side :)

MizukiTemma:
@david-venhoff 
Ohhhh 🙈 I didn't have to dig so much 😅 

Your suggestion is applied 😸

david-venhoff (approved):
:tada:

Issue #2966, opened 2024-08-05
Change styling of "Cancel"-Button in MT overlay 


JoeyStk:
### Describe the Bug

In our current styling we use a red button for "Cancel" in our MT overlay. However it was recommended to use the ghost-button for this instead to be in-line with the rest of the system.

### Steps to Reproduce

1. Go to any region that has MT activated
2. Go to pages and select a language that is not the root language (usually German)
3. Select in that language one or more page(s) and select "Machine translate selected pages to selected language"
4. Click on "Execute" and see the popup. See that the color of "Cancel" is red

### Expected Behavior

The button should be a ghost button like here: 
![grafik](https://github.com/user-attachments/assets/727508a3-0d3a-4541-8321-95ba2a1348d6)

### Actual Behavior

It's red

### Additional Information

None

JoeyStk:
Question for @nikolahoff: Should only buttons called "Cancel" be changed to a ghost button or does this include all red buttons, so also buttons for deleting? :)

PR #3248, opened 2024-11-28
Fix cancel button color to have ghost button style

lunars97:
### Short description

Right now cancel button in machine translation overlay has a red button instead of having ghost button. Additionally fix all other occurrences of cancel button not having the ghost button style.

### Proposed changes

- Change the button style to ghost button

### Side effects

- none

### Resolved issues

Fixes: #2966 

### DIFF
--- a/integreat_cms/cms/templates/_machine_translation_overlay.html
+++ b/integreat_cms/cms/templates/_machine_translation_overlay.html
@@ -124,7 +124,7 @@
                 <div class="flex justify-between mt-4">
                     <button type="button"
                             id="btn-close-machine-translation-overlay"
-                            class="btn !bg-red-500">
+                            class="btn btn-ghost">
                         {% translate "Cancel" %}
                     </button>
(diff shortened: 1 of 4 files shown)


david-venhoff (approved):
Thanks, LGTM!

I found a few other cancel buttons that currently use different styles. Do you know whether those should also be changed?

- Generic confirmation dialog: 
image

- Link list row replace url: 
image

- Poi map selector: 
image

lunars97:
> Thanks, LGTM!
> 
> I found a few other cancel buttons that currently use different styles. Do you know whether those should also be changed?
> 
> * Generic confirmation dialog:
>   image
> * Link list row replace url:
>   image
> * Poi map selector:
>   image

Thank you for your review :) And the answer for your question is yes. I think they also definitely should be changed to be consistent with the styling

JoeyStk (approved):
Thank you very much! :)

Issue #2930, opened 2024-07-25
Add cancel button to creation form of page, poi, event, organisation, language node, push notification and users


osmers:
### Motivation
A clear button that lets users cancel the current creation process they started is more intuitive than simply clicking on a menu item to leave the form

### Proposed Solution
Add a red cancel button to each creation form listed above. Maybe also to the edit forms? At least for pages, pois and events we would have to discuss the effect of the button (i.e. what to do with automatic saves etc)

### Alternatives
Keep it as it is

### User Story
As a user of the CMS I want to have a clear button that let's me cancel the changes I have made in order to be sure that no changes are saved and nothing was created.

### Additional Context
Same button as for automatic translations bulk action: 
![image](https://github.com/user-attachments/assets/e8c3238b-d1ec-4e87-87f7-af0ee2e877ed)

MizukiTemma:
Where should it redirect after clicking the button? Probably to the list (cancel on the organigation form -> the organization list)?

osmers:
@MizukiTemma yes, exactly - basically back to the page where you clicked on the create button...

lunars97:
@nikolahoff should we change the cancel button color to `ghost-button` (instead of red which is used for delete) and from now on use this color for all cancel buttons? :thinking:

nikolahoff:
@lunars97 cancel button should have this behaviour like in this [design](https://www.figma.com/design/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=2429-1129&t=KJi9rox7rgWeYceg-1) 
Delete buttons are different. For now we keep it like it is for delete. Maybe with the overall redesign we might change it.

PR #3251, opened 2024-11-29
Add cancel button to creation and edit forms

lunars97:
### Short description

Add cancel button to create and update form where user can just go back to the list overview instead of going back through side menu navigation

### Proposed changes

- Add anchor tag with ghost-button style which returns to the list overview of the related forms

### Side effects

- should be none

### Resolved issues

Fixes: #2930 

### DIFF
--- a/integreat_cms/cms/templates/pages/page_form.html
+++ b/integreat_cms/cms/templates/pages/page_form.html
@@ -44,8 +44,15 @@
                     {% translate "Create new page" %}
                 {% endif %}
             </h1>
-            {% if not page_form.instance.id or not page_form.instance.archived %}
-                <div class="flex flex-wrap grow justify-end gap-2 items-center">
+            <div class="flex flex-wrap grow justify-end gap-2 items-center">
+                {% if page_form.instance.id and page_form.instance.archived %}
+                    <a href="{% url 'archived_pages' region_slug=request.region.slug language_slug=language.slug %}"
+                       class="btn btn-ghost">{% translate "Cancel" %}</a>
+                {% else %}
+                    <a href="{% url 'pages' region_slug=request.region.slug language_slug=language.slug %}"
+                       class="btn btn-ghost">{% translate "Cancel" %}</a>
+                {% endif %}
+                {% if not page_form.instance.id or not page_form.instance.archived %}
                     {% include "generic_auto_save_note.html" with form_instance=page_translation_form.instance %}
                     <button name="preview_page" type="button" data-preview-page data-edit-page-mode data-right-to-left={{ right_to_left }} class="btn btn-ghost">
                         {% translate "Preview" %}
@@ -73,8 +80,8 @@
                             {% translate "Submit for approval" %}
                         </button>
                     {% endif %}
-                </div>
-            {% endif %}
+                {% endif %}
+            </div>
         </div>
         <div class="3xl:grid grid-cols-2 3xl:grid-cols-[minmax(0px,_1fr)_400px] {% if page %}4xl:grid-cols-[minmax(0px,_1fr)_816px]{% endif %} gap-4">
             <div class="col-span-2 3xl:col-span-1 flex flex-wrap flex-col flex-auto">
--- a/integreat_cms/release_notes/current/unreleased/2930.yml
+++ b/integreat_cms/release_notes/current/unreleased/2930.yml
@@ -1,0 +1,2 @@
+en: Add cancel button to creation and edit forms
+de: Füge einen Abbrechen-Button zu den Erstellungs- und Bearbeitungsformularen hinzu
--- a/integreat_cms/cms/templates/contacts/contact_form.html
+++ b/integreat_cms/cms/templates/contacts/contact_form.html
@@ -18,6 +18,13 @@
                     {% endif %}
                 </h1>
                 <div class="flex flex-wrap justify-between gap-4">
+                    {% if contact_form.instance.archived %}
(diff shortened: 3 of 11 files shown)


JoeyStk (commented):
Thank you very much!
This looks quite good already. I only have two thoughts:
1. I think this deserves a release note
2. I'm unsure about the page view at the moment. I think I would have put "Cancel" at to the far left, but maybe this is a question for @nikolahoff. Do you like it the way it is at the moment?
grafik

lunars97:
> Thank you very much! This looks quite good already. I only have two thoughts:
> 
> 1. I think this deserves a release note
> 2. I'm unsure about the page view at the moment. I think I would have put "Cancel" at to the far left, but maybe this is a question for @nikolahoff. Do you like it the way it is at the moment?
>    grafik

I think you are right, but let's wait for @nikolahoff opinion :)

nikolahoff:
I think there are too many buttons and styles on the same page. But maybe we can leave it for now and wait until I have done the overall review because this would effect other pages as well. @JoeyStk @lunars97

JoeyStk:
@nikolahoff Thanks! Yes, sounds reasonable. Until then, do you have a preference whether or not it should be before or after the "Preview" button? :)

integreat_cms/cms/templates/pages/page_form.html
```diff
@@ -46,6 +46,11 @@ <h1 class="heading">
             </h1>
             {% if not page_form.instance.id or not page_form.instance.archived %}
                 <div class="flex flex-wrap grow justify-end gap-2 items-center">
+                    {% has_perm 'cms.publish_page_object' request.user page as can_publish_page %}
+                    {% if can_publish_page %}
+                        <a href="{% url 'pages' region_slug=request.region.slug %}"
+                           class="btn btn-ghost">{% translate "Cancel" %}</a>
+                    {% endif %}
```
JoeyStk, line 53: We're almost there :) Could you please push it beneath the `generic_auto_save_note`, because now it looks like this? :sweat_smile: :smiley: 
grafik

lunars97: :see_no_evil:

JoeyStk: Looks good now :)

integreat_cms/release_notes/current/unreleased/2930.yml
```diff
@@ -0,0 +1,2 @@
+en: Add cancel button to creation and edit forms
+de: Füge einen Abbrechen-Button zu den Erstellungs- und Bearbeitungsformularen hinzu
```
JoeyStk, line 2: :+1:

JoeyStk (approved):
Thank you very much :rocket:

MizukiTemma:
@osmers 
I have some questions.

- I understood the scope of issue as following: a cancel button should be added to all the forms onto which region users can land if the corresponding object has a tree or list. Therefore meida library, imprint and machine translations are exempt of this change, as they do not have any tree/list, and region, language, users (in Network Management), role, offer template and location category too as they are only available for staff users.

If this interpretation is correct, the button is missing for the form of contact, external calendar and user. I guess the first two are not listed in the issue because they didn't exist at the time of issue open.

lunars97:
> @osmers I have some questions.
> 
> I understood the scope of issue as following: a cancel button should be added to all the forms onto which region users can land if the corresponding object has a tree or list. Therefore meida library, imprint and machine translations are exempt of this change, as they do not have any tree/list, and region, language, users (in Network Management), role, offer template and location category too as they are only available for staff users.
> 
> If this interpretation is correct, the button is missing for the form of contact, external calendar and user. I guess the first two are not listed in the issue because they didn't exist at the time of issue open.
> 
> And the cancel button will not be shown if the opened object is archived or the user does not have change permission (observer role), but they probably want to go back to the list by the button (once it's introduced) after checking an archived something.

@MizukiTemma @osmers should I add cancel button for form of contact, external calendar and user? :thinking:

osmers:
@lunars97 sorry for the late reply and thanks for reminding me :) 
What @MizukiTemma says makes sense, so you can add it to the ones mentioned above. 
Also I would agree that a cancel button also makes sense when a user does not have permission to change anything or is viewing an archived object so that it takes them back to the list overview (as Mizuki suggests).

MizukiTemma (approved):
Thank you 👍

JoeyStk (commented):
Thank you a lot for the improvements! To me this PR looks good now. There is just one case that we also might want to think about: Events from external calendars: When importing events from external calendars and you go to the event form of an import event, all buttons for this event are missing. This is because you shouldn't be allowed to edit imported events. At the moment also the "Cancel" button is not showing. However I'd argue that the cancel button is an exception to the rule. Therefore I think we should add the cancel button there too. What do you think? Currently it looks like this 
grafik

lunars97:
> Thank you a lot for the improvements! To me this PR looks good now. There is just one case that we also might want to think about: Events from external calendars: When importing events from external calendars and you go to the event form of an import event, all buttons for this event are missing. This is because you shouldn't be allowed to edit imported events. At the moment also the "Cancel" button is not showing. However I'd argue that the cancel button is an exception to the rule. Therefore I think we should add the cancel button there too. What do you think? Currently it looks like this grafik

I cannot reproduce it. I tried with observer role as well. But everything is working fine... hmm

JoeyStk:
> I cannot reproduce it. I tried with observer role as well. But everything is working fine... hmm

You're right. It works. It was on me! Sorry! You can merge this PR now :)

JoeyStk (approved):

Issue #3037, opened 2024-09-09
Moving around with arrow keys in input field not working


JoeyStk:
### Describe the Bug

When clicking on (certain) input fields, it's not possible to move around with the arrow keys.

### Steps to Reproduce

1. Log on to either cms-test or the prod system
2. In network management click on 'regions'
3. Type something inside the "Look for a region" search field. 
4. Try to move around with all four arrow keys, see unexpected behavior

### Expected Behavior

Arrow Keys Left/Right: move one character left/right
Arrow Keys Up/Down: move through options in dropdown

### Actual Behavior

Arrow Keys Left/Right: nothing happens
Arrow Keys Up/Down: Jump to end/beginning of word

### Additional Information
This happens in basically every search field (POI/Pages/User List View/ Global & Regional feedback)

ulliholtgrave:
This might be related to the behavior of the event listener used for the AJAX calls interpreting the key inputs as text inputs querying our API.
Besides that, I want to emphasize that the key-up/-down behavior is browser-related and we shouldn't try to mimic this with own HTML/JS.

PR #3269, opened 2024-12-04
Fix input field arrow key navigation and make search query function accessible

lunars97:
### Short description

Currently input fields for search are not navigatable via arrow keys. They should be navigatable with arrow keys + accessible with keyboard (like Tab and enter keys)

### Proposed changes

- [x] Add aria labels where it is necessary to make them accessible via keyboard
- [x] Make list items focusable for Tab key (focus indicator is a visual indicator that “highlights” the currently focused element which is really important for keyboard users)
- [x] Add support for up and down arrows for suggestion list view
- [x] Add support for left and right arrow keys on input change 

### Side effects

- should be none

### Resolved issues

Fixes: #3037

### DIFF
--- a/integreat_cms/cms/templates/_search_input.html
+++ b/integreat_cms/cms/templates/_search_input.html
@@ -8,14 +8,17 @@
                name="query"
                autocomplete="off"
                id="table-search-input"
+               aria-haspopup="listbox"
+               aria-label="Search"
                placeholder="{% translate "Search" %}"
                class="default-value rounded-r-none"
                data-url="{% if region_slug and language_slug %}{% url 'search_content_ajax' region_slug language_slug %}{% elif region_slug %}{% url 'search_content_ajax' region_slug %}{% else %}{% url 'search_content_ajax' %}{% endif %}"
                data-object-type="{{ object_type }}"
                data-archived="{% if object_archived %}true{% else %}false{% endif %}" />
-        <div id="table-search-suggestions"
-             class="absolute hidden shadow rounded-b top-full bg-graz-200 w-full z-10 max-h-60 overflow-y-auto cursor-pointer">
-        </div>
+        <ul id="table-search-suggestions"
+            role="listbox"
+            class="absolute hidden shadow rounded-b top-full bg-gray-200 w-full z-10 max-h-60 overflow-y-auto cursor-pointer">
+        </ul>
     </div>
     {% if search_query %}
         <button id="search-reset-btn"
--- a/integreat_cms/static/src/js/search-query.ts
+++ b/integreat_cms/static/src/js/search-query.ts
@@ -31,36 +31,46 @@ const queryObjects = async (url: string, type: string, queryString: string, arch
 
     if (data) {
         // Set and display new data
-        for (const value of data.data) {
+        data.data.forEach((value: any, index: number) => {
             const child = document.createElement("li");
             child.classList.add(
                 "inline-block",
                 "whitespace-nowrap",
+                "my-0.5",
                 "px-4",
                 "py-3",
                 "text-gray-800",
+                "focus:ring-2",
                 "hover:bg-gray-300",
-                "bg-gray-200",
                 "w-full",
                 "overflow-x-hidden",
                 "text-ellipsis",
                 "align-top"
             );
+            child.setAttribute("role", "option");
+            child.setAttribute("id", `suggestion-${index}`);
+            child.setAttribute("tabindex", "0");
             child.innerText = value.title;
             suggestionList.appendChild(child);
-        }
+        });
     }
 };
 
 let scheduledFunction: number | null = null;
+let focusedIndex = -1;
 
 export const setSearchQueryEventListeners = () => {
     console.debug("Setting search query event listeners");
+    const resetButton = document.getElementById("search-reset-btn");
+    const suggestionList = document.getElementById("table-search-suggestions");
     const tableSearchInput = document.getElementById("table-search-input") as HTMLInputElement;
 
     // AJAX search
     tableSearchInput.addEventListener("keyup", (event) => {
-        event.preventDefault();
+        // Ignore navigation keys for API calls
+        if (["ArrowDown", "ArrowUp", "ArrowLeft", "ArrowRight", "Escape", "Tab"].includes(event.key)) {
+            return;
+        }
 
         // Reschedule function execution on new input
         if (scheduledFunction != null) {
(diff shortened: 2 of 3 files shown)


charludo (changes requested):
Thank you, this makes keyboard navigation much easier!!

One thing I noticed: the first time I hit Tab (latest Firefox on Linux), nothing visually changes. However, if I press Enter, the search for the first (selected) element commences. For example, here I have pressed "h", then "Tab":

image

And here, I have pressed "h", then "Tab", then "Tab" again:

image

I think it's important to mark the selected elements a bit clearer.

Using the up/down arrow keys, the "hover" effect is good at communicating this:

image

So maybe that + a full outline for Tab?

charludo:
> * [ ]  Add support for left and right arrow keys on input change

I just noticed this problem exists in other fields as well, for example: the "Source Code" view in TinyMCE. Maybe this is worth a separate PR?

lunars97:
> > * [ ]   Add support for left and right arrow keys on input change
> 
> I just noticed this problem exists in other fields as well, for example: the "Source Code" view in TinyMCE. Maybe this is worth a separate PR?

i think this PR should also fix other fields...

lunars97:
> Thank you, this makes keyboard navigation much easier!!
> 
> One thing I noticed: the first time I hit Tab (latest Firefox on Linux), nothing visually changes. However, if I press Enter, the search for the first (selected) element commences. For example, here I have pressed "h", then "Tab":
> 
> image
> 
> And here, I have pressed "h", then "Tab", then "Tab" again:
> 
> image
> 
> I think it's important to mark the selected elements a bit clearer.
> 
> Using the up/down arrow keys, the "hover" effect is good at communicating this:
> 
> image
> 
> So maybe that + a full outline for Tab?

Thank your for your review. On Tab an outlined border should appear. At least that is how it is on Chrome. I think Firefox interprets it differently and has its own default style for outline :)

integreat_cms/static/src/js/search-query.ts
```diff
@@ -78,36 +88,94 @@ export const setSearchQueryEventListeners = () => {
...
+                const newCaretPosition = Math.max(0, input.selectionStart - 1);
+                input.setSelectionRange(newCaretPosition, newCaretPosition);
+            }
+        } else if (event.key === "ArrowRight") {
+            event.preventDefault();
+            if (input.selectionStart !== null && input.selectionEnd !== null) {
+                const newCaretPosition = Math.min(input.value.length, input.selectionEnd + 1);
+                input.setSelectionRange(newCaretPosition, newCaretPosition);
+            }
+        }
```
charludo, line 111: OK, this definitely works. However, it looks like the reason this is required in the first place is this code snippet:

```js
    document.addEventListener("keydown", (event) => {
        if (event.key === "ArrowRight") {
            event.preventDefault();
            selectNextTimelineItem();
        } else if (event.key === "ArrowLeft") {
            event.preventDefault();
            selectPrevTimelineItem();
        }
    });
```
in [this file](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/static/src/js/revisions.ts#L308-L316).

I think that old code is a mistake. Looks like the intention was to prevent accidentally changing what translation version is being viewed in the version history timeline. But instead, it deactivates the arrow keys globally.

(The fact that we add ALL of our JS to EVERY page is a problematic story all on its own... 😂)

Rather than overwriting the behavior with your new code, I think it would be better to fix the old behavior. For example, you could add a check in `revisions.ts` to see if `.timeline-item` dics exist on the page, and only then add the arrow-key blocking code.

integreat_cms/static/src/js/search-query.ts
```diff
@@ -78,36 +88,94 @@ export const setSearchQueryEventListeners = () => {
...
+        } else if (event.key === "Enter") {
+            event.preventDefault();
+            const activeSuggestion = suggestions[focusedIndex];
+            if (activeSuggestion) {
+                tableSearchInput.value = activeSuggestion.textContent || "";
+                document.getElementById("search-submit-btn")?.click();
+                suggestionList?.classList.add("hidden");
+                focusedIndex = -1;
+            }
+            return;
```
charludo, line 128: This is very similar to this, only a few lines below:

```js
    suggestionList?.addEventListener("keydown", (event) => {
        if (event.key === "Enter") {
            const activeSuggestion = document.activeElement as HTMLElement;
            if (activeSuggestion && suggestionList.contains(activeSuggestion)) {
                event.preventDefault();
                tableSearchInput.value = activeSuggestion.textContent || "";
                document.getElementById("search-submit-btn")?.click();
                suggestionList?.classList.add("hidden");
            }
        }
    });
```
I think you can simply remove this (old) code - right? At least, when it's commented out, I can't seem to find any differences in behavior

charludo (changes requested):
> Update: Now Firefox uses its default outline :) I think for people who use "tab" key, the most important part is to make them focused (support the default behaviour of tabbable items)

Yes, definitely. Though even non-visually impaired users can appreciate tab-navigation (speaking from personal experience :D) 

Anyways, yes, Firefox now also shows an outline 👍🏼 

> i think this PR should also fix other fields... Though seems like it is a textfild and maybe it will require additional time 🤔

Yep, works perfectly now in the Region Search, but (e.g.) the Source Code view still does not react to arrow keys.

You are more familiar with this issue, so it's your decision if you want to add that on to this PR or open a new issue!

integreat_cms/static/src/js/search-query.ts
lunars97: @charludo thank you for pointing out this one. Because I was really wondering and was not able to find the reason behind it. It turns out that this exact code was blocking the default behaviour of other fields :crying_cat_face:

integreat_cms/static/src/js/search-query.ts
lunars97: @charludo this one `on dropdown list`, without on enter it won't work unfortunately with up and down arrows as well as with tab. The above was` on input field` enter.

integreat_cms/static/src/js/search-query.ts
lunars97: @charludo Don't you think it would be better and more accessible if the timeline versions were also available with left and right arrows? Because i find it really easy just to choose the necessary version without clicking on every version... :thinking:

charludo: I can only imagine it was added to prevent accidental selections, but yes, I am not sure how much sense that makes... 😅

charludo (approved):
Thank you, this looks good to me! 🎉

integreat_cms/cms/templates/_search_input.html
```diff
@@ -8,14 +8,17 @@
...
+               aria-label="Search"
                placeholder="{% translate "Search" %}"
                class="default-value rounded-r-none"
                data-url="{% if region_slug and language_slug %}{% url 'search_content_ajax' region_slug language_slug %}{% elif region_slug %}{% url 'search_content_ajax' region_slug %}{% else %}{% url 'search_content_ajax' %}{% endif %}"
                data-object-type="{{ object_type }}"
                data-archived="{% if object_archived %}true{% else %}false{% endif %}" />
-        <div id="table-search-suggestions"
-             class="absolute hidden shadow rounded-b top-full bg-graz-200 w-full z-10 max-h-60 overflow-y-auto cursor-pointer">
-        </div>
+        <ul id="table-search-suggestions_list"
```
JoeyStk, line 18: Is there an intention to use an underscore before the list in "table-search-suggestions_list"? :)

JoeyStk (approved):
Thank you very much!
I already approved this PR as it looks good and I only have one less relevant question :)

integreat_cms/cms/templates/_search_input.html
lunars97: @JoeyStk Thank you for your review! :) I wanted to highlight that it is a list, but there is no extra intention...

JoeyStk: @lunars97 Thank you for explaining! No worries, that was not what I meant. I was wondering, why you were changing from dash "table-search-suggestions" to underscore "_list" :)

lunars97: @JoeyStk I think I will remove it better, because the id is self explanatory :)

MizukiTemma:
Hmm, I guess this PR deserves a release note... Should we add it afterwards? @lunars97 @charludo @JoeyStk

PR #3272, opened 2024-12-06
Drop root privileges for celery worker

svenseeberg:
### Short description
The run.sh is running with root privileges. However, celery does not need those.

### Proposed changes
- Drop root privileges before executing the celery worker
- add additional celery temp files to gitignore

### Side effects
- none

### Resolved issues
Fixes: https://chat.tuerantuer.org/digitalfabrik/pl/qcemkcp5aid8dnhes6obwtiooo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,5 +71,8 @@ integreat_cms/xliff/download
 .postgres
 
 # Celery
-celerybeat-schedule.db
 celerybeat-schedule
+celerybeat-schedule.db
+celerybeat-schedule.bak
+celerybeat-schedule.dir
+celerybeat-schedule.dat
--- a/tools/run.sh
+++ b/tools/run.sh
@@ -47,7 +47,9 @@ done
 listen_for_devserver &
 
 # Run Celery worker process
-celery -A integreat_cms.integreat_celery worker -l INFO -B --concurrency=1 &
+if [ "$INTEGREAT_CMS_REDIS_CACHE" == "1" ]; then
+  deescalate_privileges celery -A integreat_cms.integreat_celery worker -l INFO -B --concurrency=1 &
+fi
 
 # Start Integreat CMS development webserver
 deescalate_privileges integreat-cms-cli runserver "localhost:${INTEGREAT_CMS_PORT}"


.gitignore
```diff
@@ -71,5 +71,7 @@ integreat_cms/xliff/download
 .postgres
 
 # Celery
-celerybeat-schedule.db
 celerybeat-schedule
+celerybeat-schedule.db
```
david-venhoff, line 75:
```suggestion
celerybeat-schedule.dat
```

david-venhoff (approved):
Thanks a lot, this fixes the warning for me :)

.gitignore
charludo: `celerybeat-schedule.db` is apparently also a thing 😅

tools/run.sh
```diff
@@ -47,7 +47,7 @@ done
 listen_for_devserver &
 
 # Run Celery worker process
-celery -A integreat_cms.integreat_celery worker -l INFO -B --concurrency=1 &
+deescalate_privileges celery -A integreat_cms.integreat_celery worker -l INFO -B --concurrency=1 &
```
charludo, line 50: This is more a question than a comment, but: are we dependent on celery for the dev env? Because as far as I can tell, I can still do everything I normally would in the CMS when commenting out this line. I get that background workers won't be available, but will this crash the dev server, or just lead to errors in the logs?

If it is "optional", I would love the option to set an environment variable to skip starting celery (and thus not need redis installed).

If it is mandatory, I think the docs should mention that redis is also a mandatory prerequisite to running the dev server.

charludo (changes requested):

tools/run.sh
svenseeberg: I have no strong opinion here. But I'd recommend to move all cron tasks to Celery scheduled tasks.

.gitignore
svenseeberg: Yeah, I'm wondering where the .dat file is coming from. I never had that :D

tools/run.sh
charludo: > In the end Celery will simply not work if Redis is not running.

Not quite - it completely stops the dev server from starting if redis is not running.

charludo (approved):
I'll approve this so it can get merged, but will open a separate PR/issue to allow redis to be disabled in the dev env.

.gitignore
charludo: Please add the .dat, I think after that you can merge :)

tools/run.sh
svenseeberg: How should we check for a running Redis server? `pidof redis-server`?

svenseeberg: Oh, we actually have an existing env var for this.

PR #3273, opened 2024-12-06
Support rrules with `BYDAY` frequencies

david-venhoff:
### Short description

~~***This pr is blocked because it depends on #3229, but it can already be reviewed***~~
***The blocker is now resolved***

This pr adds support for importing events with rrules that contain `BYDAY` entries with frequencies, for example `-1SU` (which means repeat every last Sunday).

Unfortunately the icalendar library does not parse these entries fully right now, so this pr also implements some manual parsing. However, in its next major version icalendar will add support for that.

### Proposed changes

- Check if the `BYDAY` entry has a frequency, and use that frequency as a value for `BYSETPOS`
- Handle edge cases
- Implement some manual parsing for `BYDAY` weekdays

### Side effects

/

### Resolved issues

Fixes: #3227


MizukiTemma (approved):
Cool 😻 🎉

PeterNerlich (changes requested):
`RRULE:FREQ=MONTHLY;UNTIL=20250502T220000Z;INTERVAL=1;BYMONTHDAY=3` still produces a `ValueError('by_set_pos must not be None')` *(from `outlook.calendar.txt` in #3227)*

david-venhoff:
> RRULE:FREQ=MONTHLY;UNTIL=20250502T220000Z;INTERVAL=1;BYMONTHDAY=3 still produces a ValueError('by_set_pos must not be None') (from outlook.calendar.txt in https://github.com/digitalfabrik/integreat-cms/issues/3227)

Right now, we only support recurrence rules which can be also be created from within the recurrence rule form. However, this recurrence rule repeats every third day of the month, which cannot be expressed in the recurrence rule form (We only support monthly recurrences if we specify the nth-week and a weekday). So this error is intentional

PeterNerlich (approved):

Issue #3268, opened 2024-12-04
Exclude contact card from HIX score calculation


charludo:
### Describe the Bug

The contact card negatively affects the HIX score, even though it should not.

### Steps to Reproduce

1. Save page, see HIX score
2. Insert Contact Card
3. save page, see reduced HIX score

### Expected Behavior

Hix score unaffected.

### Actual Behavior

Hix Score severely affected.

### Additional Information

Just strip the contact card from the string passed to Textlab.

### Alternative
Remove all `contentenditable="false"` blocks from the string before passing to Textlab. Would future-proof a little bit, and IMO it makes sense to not include content in the score calculation on which users have no influence? 

### Related Issues

Depends on 
- [x] #2959

osmers:
@charludo I just tested this on this page: https://cms-test.integreat-app.de/muenchen/pages/de/8258/edit/
I converted all three contacts into contact cards and inserted them and the HIX-value went down from 12.78 to 12.29 - compare version 43 and 45. Could it have to to with the website being spelled out completely? Although I thought that wouldn't matter since everything in the contact card is excluded from the HIX score calculation.

charludo:
> [@charludo](https://github.com/charludo) I just tested this on this page: https://cms-test.integreat-app.de/muenchen/pages/de/8258/edit/ I converted all three contacts into contact cards and inserted them and the HIX-value went down from 12.78 to 12.29 - compare version 43 and 45. Could it have to to with the website being spelled out completely? Although I thought that wouldn't matter since everything in the contact card is excluded from the HIX score calculation.

Yep, that's not great indeed.

It looks like for some reason, the card is indeed not being removed. I just tried copying the source of that page to my local environment, and received a HIX score of 13.67 - a marked improvement over the original 12.78! - until I intentionally broke the code filtering out the contact cards, and then the score fell to the same 12.29.

So either something is not working right, or the version on the test CMS has been overwritten by some other branch, which is my guess -.-

osmers:
@charludo if we merge it into develop, we can still exclude it from the release, right? But yeah, let's merge it and then test intensively in the next week and see what results we get :)

charludo:
@

> [@charludo](https://github.com/charludo) if we merge it into develop, we can still exclude it from the release, right? But yeah, let's merge it and then test intensively in the next week and see what results we get :)

Yes, plus it would be trivial to revert.

PR #3281, opened 2024-12-09
Exclude contact cards (and other non-editable blocks) from HIX

charludo:
### Short description

The contact cards introduced in #3169 have a severe negative impact on the HIX score. This PR excludes them from the HIX calculations.

### Proposed changes

- before sending text to Textlab, remove all divs with `contenteditable="false"`.
- this currently only affects contacts, but makes it simple to exclude other blocks in the future, should the need arise

### Side effects

- `html.text_content()` was previously only used to check for empty pages. For convenience, I have changed the code to send the result of this operation to textlab instead of the raw HTML - but I am uncertain if there has been a decision against this in the past when you added the code in question @david-venhoff - are my changes OK?  

### Resolved issues

Fixes: #3268

### DIFF
--- a/integreat_cms/cms/views/utils/hix.py
+++ b/integreat_cms/cms/views/utils/hix.py
@@ -54,6 +54,11 @@ def lookup_hix_score_helper(text: str) -> TextlabResult:
     """
     try:
         html = fromstring(text)
+
+        # remove divs which the authors have no control over (e.g. contact cards)
+        for div in html.xpath('//div[@contenteditable="false"]'):
+            div.getparent().remove(div)
+
         text_content = html.text_content()
         if not text_content.strip():
             return {
@@ -64,7 +69,7 @@ def lookup_hix_score_helper(text: str) -> TextlabResult:
         pass
 
     # Replace all line breaks with <br> because Textlab API returns different HIX value depending on the line break character
-    normalized_text = "<br>".join(text.splitlines())
+    normalized_text = "<br>".join(text_content.splitlines())
 
     try:
         return TextlabClient(


david-venhoff:
> @david-venhoff - are my changes OK?

I fear that this might change the hix scores of pages again. The last time we did this our service team had quite a nightmare dealing with municipalities that suddenly could not translate their pages anymore because the hix score was too low. 
If we do this, we should probably at least do some tests that this change does not decrease the hix score in comparison to right now.

JoeyStk (approved):
I don't have a lot of context about this piece of code. I hope @david-venhoff knows more about it. For what I can see this piece of code looks good and the logic itself makes sense, but I might not think of all possible side effects :/

charludo:
> > @david-venhoff - are my changes OK?
> 
> I fear that this might change the hix scores of pages again. The last time we did this our service team had quite a nightmare dealing with municipalities that suddenly could not translate their pages anymore because the hix score was too low. If we do this, we should probably at least do some tests that this change does not decrease the hix score in comparison to right now.

That's a very valid concern. 

I'm not sure how else to progress though, tbh. Maybe passing `tostring(html)` would be the better option, but that is not guaranteed to achieve `text == tostrong(fromstring(text))`, afaik.

Removing the divs in question without parsing the HTML is an entirely differnet can of worms.

Frankly I don't think we have a choice but to risk slightly changing HIX scores :( (With prior testing still, of course)

charludo:
I think we need @osmers to chime in on this 😅 
In short: the issue is that we must remove contact cards, otherwise the HIX score worsens a lot; but the only way to reliably do so can lead to very slight changes in the HTML content we send to Textlab, and so could result in slightly changed HIX scores compared to right now, *including* pages not containing any contact cards. I really cannot say if that change in score would be an increase or decrease.

osmers:
@charludo can I already test this in the test cms? I would then just go through each page for a region and see how much the HIX-score changes, even without inserting contact cards. Though I think it would make sense to also test it with inserted contact cards just to make sure (bcs there municipalities will expect an increase in the HIX-score).
As David said, we definitely need to test this and if the tests show rather bad results, in the sense of much lower HIX-scores, we need to rethink how to do this...

charludo:
> @charludo can I already test this in the test cms? I would then just go through each page for a region and see how much the HIX-score changes, even without inserting contact cards. Though I think it would make sense to also test it with inserted contact cards just to make sure (bcs there municipalities will expect an increase in the HIX-score). As David said, we definitely need to test this and if the tests show rather bad results, in the sense of much lower HIX-scores, we need to rethink how to do this...

Not really, no, since this is not merged yet :/

Contact Cards should definitely improve HIX compared to plain-text contact details though, yes.

osmers:
Could we merge it but possibly block it from the next release? Then Salua and me can test it :)

david-venhoff:
It is also possible to test without merging by creating a pr to a branch that ends in -publish-dev-package: https://github.com/digitalfabrik/integreat-cms/blob/102b45fe268754e5dab9d6156c063c74a1484cf4/.circleci/config.yml#L661-L668

charludo:
> It is also possible to test without merging by creating a branch that ends in -publish-dev-package:
> 
> https://github.com/digitalfabrik/integreat-cms/blob/102b45fe268754e5dab9d6156c063c74a1484cf4/.circleci/config.yml#L661-L668

Huh, I had no idea!

The build was successfull, but I cannot dinf where I can actually access it 🤔 Do we have automated deploy for these?

david-venhoff:
I think it will be deployed on the cms test server, when it updates the next time. I think there is also a way to update the test server manually, so that it runs the pr immediately

charludo:
> I think it will be deployed on the cms test server, when it updates the next time. I think there is also a way to update the test server manually, so that it runs the pr immediately

Ohhhh so this will take this build just because it's the newest? But there's no way to switch between versions on the test server?

david-venhoff:
Yeah, that is my understanding

charludo:
@david-venhoff would you mind approving/re-reviewing this in light of the discussion on #3268? 😅 :D

david-venhoff (approved):
Code looks good to me

PR #3284, opened 2024-12-10
Fix sudden ruff complaint

charludo:
### Short description

Since yesterday, `ruff` has been complaining about the following:
```
integreat_cms/core/wsgi.py:43:5: PLC0206 Extracting value from dictionary without calling `.items()`
   |
42 |       # Read config from environment
43 |       for key in environ:
   |  _____^
44 | |         if key.startswith("INTEGREAT_CMS_"):
45 | |             os.environ[key] = environ[key]
   | |__________________________________________^ PLC0206
46 |   
47 |       _application = get_wsgi_application()
   |
```
That code was last touched 3 years ago, so not sure why the sudden complaint, but it *is* a valid complaint.

### Proposed changes

- use `items()` instead of implicitly iterating over / accessing by key

### Side effects

- oof I hope none
- alternative would be to just ignore the ruff error

### Resolved issues

Fixes: n/a

### DIFF
--- a/integreat_cms/core/wsgi.py
+++ b/integreat_cms/core/wsgi.py
@@ -40,9 +40,9 @@ def application(environ: dict[str, str], start_response: Callable) -> WSGIHandle
             os.environ.setdefault(f"INTEGREAT_CMS_{KEY.upper()}", VALUE)
 
     # Read config from environment
-    for key in environ:
+    for key, value in environ.items():
         if key.startswith("INTEGREAT_CMS_"):
-            os.environ[key] = environ[key]
+            os.environ[key] = value
 
     _application = get_wsgi_application()
 


MizukiTemma (approved):
Somehow I don't encounter this complainment but will approve the PR as it seems to be working fine 😸

david-venhoff (approved):
> so not sure why the sudden complaint

Looks like this lint was added to ruff 3 weeks ago, so that would explain it:  https://github.com/astral-sh/ruff/releases/tag/0.8.0/

Either way, the code looks cleaner this way :+1:

charludo:
> > so not sure why the sudden complaint
> 
> Looks like this lint was added to ruff 3 weeks ago, so that would explain it: https://github.com/astral-sh/ruff/releases/tag/0.8.0/
> 
> Either way, the code looks cleaner this way 👍

Ah, good catch!

Issue #3151, opened 2024-10-23
When editing a contact and removing the location it hides the other grid items


juliankehne:
### Describe the Bug
When editing an existing Contact and removing the Location (trash icon) from the contact it hides the grid items "Manage contact data" and "Contact information"

### Steps to Reproduce

1. Go to 'Contacts'
2. Open an existing contact to edit it
3. Click the trash bin icon to remove the location from the contact
4. See that grid items "Manage contact data" and "Contact information" are now hidden

### Expected Behavior
When editing an already existing contact the two grid items should stay visible

### Actual Behavior
When editing an already existing contact the two grid items are hidden after the location was removed

### Additional Information

Hiding the grid items items is great during the creation of a contact, but I think it is confusing when editing a contact and some of the existing information is suddenly not visible anymore.

Before removing the location:
![2024-10-23 16_04_19-Integreat Editorial System — Mozilla Firefox](https://github.com/user-attachments/assets/1de30fb2-a5c3-4cc2-9b47-fc4e5bfa5152)

After removing the location:
![2024-10-23 16_04_33-Integreat Editorial System — Mozilla Firefox](https://github.com/user-attachments/assets/c15fa702-e92a-4f79-856d-9b329894a02b)

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

juliankehne:
If editing a contact should have the same behaviour as creating a new contact as defined in this design then this issue can be closed again:

https://www.figma.com/design/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=1752-1636

osmers:
I think we could maybe grey out the two sections that are currently disappearing to indicate that they cannot be changed without a poi being selected?

juliankehne:
As I understand it we are hiding the grids during creation to guide the user through the process. But theoretically we don't have to do that as long as a location is selected in the end and all other mandatory fields are filled out.

And if a user tries saving without a location selected it does return an error message already:
![2024-10-23 16_28_05-Integreat Editorial System — Mozilla Firefox](https://github.com/user-attachments/assets/7be8bf5c-896d-4a1b-a3fb-34eb6a2f432d)

So when editing an existing contact I don't think we need to guide the user through the process again, be it via hiding or greying out the fields.

What I can think of is that we could put an indicator that Location has to be selected, once the user removes the current location.

juliankehne:
@nikolahoff What do you think?

[...]

PR #3292, opened 2024-12-12
Stop hiding the grids when POI is removed

MizukiTemma:
### Short description

This PR fixes the prblem that the middle and right grid are closing when a POI is removed while editing an existing contact.

### Proposed changes

- Remove code lines that close the grids whenever location is unseleted
- Change the trash icon to pencil icon
- Add a message indicating location is mandatory.

### Side effects

- I hope none 🙈 
- Please check for side effects in event form too

### Resolved issues

Fixes: #3151 

### DIFF
--- a/integreat_cms/static/src/js/poi-box.ts
+++ b/integreat_cms/static/src/js/poi-box.ts
@@ -3,11 +3,11 @@ import { getCsrfToken } from "./utils/csrf-token";
 
 type FormResponse = { success: boolean; poi_address_container: string };
 
-const toggleContactFieldBox = (show: boolean) => {
+const showContactFieldBox = () => {
     const contactFieldsBox = document.getElementById("contact_fields");
-    contactFieldsBox?.classList.toggle("hidden", !show);
+    contactFieldsBox?.classList.remove("hidden");
     const contactUsageBox = document.getElementById("contact_usage");
-    contactUsageBox?.classList.toggle("hidden", !show);
+    contactUsageBox?.classList.remove("hidden");
 };
 
 const hideSearchResults = () => {
@@ -19,7 +19,7 @@ const renderPoiData = (poiTitle: string, newPoiData: string) => {
     document.getElementById("poi-address-container").outerHTML = newPoiData;
     document.getElementById("poi-query-input").setAttribute("placeholder", poiTitle);
     hideSearchResults();
-    toggleContactFieldBox(true);
+    showContactFieldBox();
 };
 
 const hidePoiFormWidget = () => {
@@ -34,6 +34,7 @@ const setPoi = ({ target }: Event) => {
     renderPoiData(option.getAttribute("data-poi-title"), option.getAttribute("data-poi-address"));
     document.getElementById("poi-address-container")?.classList.remove("hidden");
     console.debug("Rendered POI data");
+    document.getElementById("info-location-mandatory")?.classList.add("hidden");
 };
 
 const showMessage = (response: FormResponse) => {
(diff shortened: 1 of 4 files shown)


integreat_cms/static/src/js/poi_box.ts
```diff
@@ -3,11 +3,11 @@ import { getCsrfToken } from "./utils/csrf-token";
```
JoeyStk, line 3: Nit: I think we said at one point, that the underscore should be used for Python/Django files and for JS/TS files, we just a dash instead. So, if I remember correctly, I think this should be `poi-box.ts`. I hope changing this won't give you a headache :see_no_evil:

integreat_cms/static/src/js/poi_box.ts
```diff
@@ -154,8 +155,9 @@ const removePoi = () => {
     hideSearchResults();
     // Clear the poi form
     hidePoiFormWidget();
-    toggleContactFieldBox(false);
     console.debug("Removed POI data");
+    document.getElementById("poi-address-container").getElementsByTagName("input")[0].value = "-1";
```
JoeyStk, line 159: Am I understanding correctly that you're searching for the poi input field? If so, I think it would be better to get the Element by id, as it already has the id of `poi-query-input`. With the current implementation I see the potential risk, that we at one point in the future we add another input field and then this line would not be correct anymore?

JoeyStk (approved):
Thank you very much for the PR! I think it looks good so far :rocket: 

During my review I checked the following:
:heavy_check_mark: The PR follows the design and uses the correct wording of the design
:heavy_check_mark: Code is understandable to me (besides one nit and one question, see below :) )
:heavy_check_mark: Checking in the events form for side effects. By positive testing, I couldn't find a side effect. The next reviewer could focus on some negative testing :)

integreat_cms/static/src/js/poi_box.ts
MizukiTemma: The target is another input element in `_poi_address_container.html` but it has an ID too and your suggestion is applied 😸

integreat_cms/static/src/js/poi-box.ts
```diff
@@ -3,11 +3,11 @@ import { getCsrfToken } from "./utils/csrf-token";
 
 type FormResponse = { success: boolean; poi_address_container: string };
 
-const toggleContactFieldBox = (show: boolean) => {
+const toggleContactFieldBox = () => {
```
david-venhoff, line 6:
```suggestion
const searchContactFieldBox = () => {
```

Since there is not toggling going on anymore :)

integreat_cms/static/src/js/poi-box.ts
```diff
@@ -19,7 +19,7 @@ const renderPoiData = (poiTitle: string, newPoiData: string) => {
     document.getElementById("poi-address-container").outerHTML = newPoiData;
     document.getElementById("poi-query-input").setAttribute("placeholder", poiTitle);
     hideSearchResults();
-    toggleContactFieldBox(true);
+    toggleContactFieldBox();
```
david-venhoff, line 22:
```suggestion
    showContactFieldBox();
```

MizukiTemma:
> One thing I noticed is that the grids still get hidden if the user clicks submit when the location is empty. Even though, when the user selects a location again, the previous values for the other grid items are still there. Do you think it would make sense to also fix this?


@david-venhoff Yes 😸 patched to keep them open if any of the five fields is filled with something.

jarlhengstmengel (approved):
Looks good to me! Thank you!

Issue #3290, opened 2024-12-11
MalteApp: The visualization of Urdu in CMS "Nachrichten" should be RTL


nassabay:
### Describe the Bug
The Urdu language in Malte CMS is LTR, but it should be like it is in Integreat... RTL. 
Note: The app is showing it correctly incorrect, only the CMS is not.
See graphic

### Steps to Reproduce

1. Go to Malteapp -> Testumgebung
2. Click on 'Nachrichten'
3. Click on 'Nachricht erstellen'
4. Change to Urdu and write something  
6. See error language is LTR

### Expected Behavior
Language should be RTL.
![Screenshot (107)](https://github.com/user-attachments/assets/9d264ae4-1712-44b6-b90f-baff521d5679)

MizukiTemma:
It seems the problem exists also in the event and POI form in Integreat-cms.

MizukiTemma:
@nassabay 
It was becuase of the language setting that Urdu appeard in LTR in the push notifications in Malte. After changing the text direction into RTL from LRT, it is shown in RTL. The same change has been applied to Dari and Paschto too, as they were set to LTR while actually RTL.

MizukiTemma:
The original problem reported in the issue description is now solved, but RTL languages are shown LTR in the POI and event form. This needs a fix in code.

PR #3293, opened 2024-12-13
Fix text direction in form

MizukiTemma:
### Short description

This PR fixes text direction of the title and content field.

### Proposed changes

- Pass `right_to_left` in the event, POI and imprint view too, which is already used in the page form to set the text direction of TinyMCE.
- Add `text_direction` in the event, POI, imprint and page view to set the text direction of title input field in the template.

### Side effects

- `right_to_left` and `text_direction` have almost the same function but I let both exist, because `right_to_left` is already being used in several places and it does not look cool (in my opinion) to write `text_direction == "rtl"` or something repeatedly 😅 (solved for page, POI and event)
- If I remember correctly, there was once a discusion about fixing the text direction of title field but it was rejected/postponed due to some technical problem or complexity or something. Though I couldn't find such thema in the repository, it may be a wrong memory. If you remember or know such (potential) problem, please comment 👀 

### Resolved issues

Fixes: #3290 

### DIFF
--- a/integreat_cms/cms/forms/machine_translation_form.py
+++ b/integreat_cms/cms/forms/machine_translation_form.py
@@ -15,6 +15,7 @@ if TYPE_CHECKING:
 
 from ...core.utils.machine_translation_provider import MachineTranslationProvider
 from ...textlab_api.utils import check_hix_score
+from ..constants import text_directions
 from ..models import LanguageTreeNode
 from .custom_content_model_form import CustomContentModelForm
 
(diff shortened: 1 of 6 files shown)


david-venhoff (approved):
Thanks, looks good to me!

It is possible to save a bit on code by adding something like this to the `machine_translation_form.__init__`:
```python
        localized_fields = ["title"]
        for field in localized_fields:
            self.fields[field].widget.attrs["dir"] = "rtl" if self.language.text_direction == text_directions.RIGHT_TO_LEFT else "ltr"
```

Then we would not need to have the repeated `class+="dir=text_direction"` and `"text_direction": "rtl" if language.text_direction == text_directions.RIGHT_TO_LEFT else "ltr",` lines.

MizukiTemma:
@david-venhoff Thank you for review and idea 😸 It could not be applied for the imprint form as our maschine translation feature is not available for this content type currently. But I guess MT will be implemented in future for imprint too and the duplicates were removed for other contents 👍

JoeyStk (approved):
Thank you very much! LGTM! :)

Issue #3102, opened 2024-10-04
iCal: Add filter to event list to filter for imported events & not imported events


JoeyStk:
### Motivation

Since #100 it's possible to import events using the iCal format. However this can overcrowd the event list quite easily. Therefore we want to add filters to the event list. 

### Proposed Solution

The idea would be add two options to the existing filters:
- [ ] Make it possible to filter for imported events 
- [ ] Make it possible to filter for not imported events.

### Alternatives

No others were discussed

### User Story

As a user of the cms I want to have an overview of which events have been imported/have not been imported so that I'm not too overwhelmed when working with events.

### Additional Context

None

### Design Requirements

Think about a sensible input field

PR #3295, opened 2024-12-17
Add filter for external calendar

JoeyStk:
### Short description

This PR adds a filter for external calendars to the event list view. It only makes sense to merge this PR once we decide to "go live" with the external calendar feature. 

### Proposed changes

- Add two checkboxes to filter for imported events or for not imported events

### Side effects

None?

### Resolved issues

Fixes: #3102 

### DIFF
--- a/integreat_cms/cms/forms/events/event_filter_form.py
+++ b/integreat_cms/cms/forms/events/event_filter_form.py
@@ -10,12 +10,13 @@ from datetime import date, datetime, time
 from typing import TYPE_CHECKING
 
 from django import forms
+from django.utils.translation import gettext_lazy as _
 
 if TYPE_CHECKING:
-    from ..models import Region
-    from ..models.events.event import EventQuerySet
+    from ...models import Region
+    from ...models.events.event import EventQuerySet
 
-from ...constants import all_day, events_time_range, recurrence
+from ...constants import all_day, calendar, events_time_range, recurrence
 from ...models import EventTranslation
 from ..custom_filter_form import CustomFilterForm
 
@@ -69,6 +70,15 @@ class EventFilterForm(CustomFilterForm):
         initial=[events_time_range.UPCOMING],
         required=False,
     )
+
+    imported_event = forms.TypedMultipleChoiceField(
+        widget=forms.CheckboxSelectMultiple(attrs={"class": "default-checked"}),
+        choices=calendar.CHOICES,
+        initial=[key for (key, val) in calendar.CHOICES],
+        coerce=calendar.DATATYPE,
+        required=False,
+    )
+
     poi_id = forms.IntegerField(widget=forms.HiddenInput, initial=-1, required=False)
 
     query = forms.CharField(required=False)
@@ -95,6 +105,7 @@ class EventFilterForm(CustomFilterForm):
             events_time_range.ALL_EVENTS
         ):
             # Either post & upcoming or no checkboxes are checked => skip filtering
+            # We do this to check if either all options or none were selected from this group
             pass
         elif events_time_range.CUSTOM in cleaned_time_range:
             # Filter events for their start and end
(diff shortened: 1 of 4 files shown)


integreat_cms/cms/forms/events/event_filter_form.py
```diff
@@ -148,6 +158,17 @@ def apply(
...
+
+        if (
+            calendar.EVENT_NOT_FROM_EXTERNAL_CALENDAR
+            in self.cleaned_data["imported_event"]
+        ):
+            events = events.filter(external_calendar__isnull=True)
+        elif (
+            calendar.EVENT_FROM_EXTERNAL_CALENDAR in self.cleaned_data["imported_event"]
+        ):
+            events = events.filter(external_calendar__isnull=False)
```
david-venhoff, line 170: That means the default behavior is that only non-imported events are shown, since both checkboxes are checkbed by default, but only the first case applies.

For me the expected behavior would be:
- Both checked: No filter applied
- One checked: Filter either for has external calendar or has no external calendar
- None checked: I guess this would be invalid and treated as both checked, like for the other checkboxes?

JoeyStk: For me this already works as you have described. Did it not work for you? :)

MizukiTemma: What @david-venhoff meant is that filtering should be explicitly skipped by `pass` if both or none of the two choices are checked, like for the filter by recurrence type or `all_day`. 

The filter looks as it it were working fine, even without such explicit skipping, if you test by changing only the calender choices. To see it is actually needed, try the following steps:

1. Save https://github.com/user-attachments/files/17816539/outlook.calendar.txt as external calendar.
2. See 11 events in the event list (before applying any filter), all of them are recurring events.
3. Open the filter, uncheck "non-recurring events"
4. See only one event appears. (Expected: all events are shown)

MizukiTemma (changes requested):

integreat_cms/cms/forms/events/event_filter_form.py
JoeyStk: Thank you for explaining. This made it more clear to me! I hope I fixed what you meant now :) Could you please check again? :)

MizukiTemma (approved):
Thank you 😸

integreat_cms/cms/forms/events/event_filter_form.py
```diff
@@ -148,6 +158,22 @@ def apply(
         elif recurrence.NOT_RECURRING in self.cleaned_data["recurring"]:
             # Only non-recurring events
             events = events.filter(recurrence_rule__isnull=True)
+
+        if (
+            len(self.cleaned_data["imported_event"]) == len(calendar.CHOICES)
+            or not self.cleaned_data["imported_event"]
+        ):
+            pass
```
charludo, line 166: Could you add a comment about this? Maybe not necessarily right above this, but since the pattern is repeated multiple times in that function, maybe above the first occurrence? With the context "check if either all options or none were selected from this group" it makes perfect sense, but IMO not super intuitive.

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -143,6 +143,14 @@ msgstr "PPT-Dokument"
...
 msgstr "PPTX-Dokument"
 
+#: cms/constants/calendar.py
+msgid "Event not from an external calendar"
+msgstr "Veranstaltungen ohne externen Kalender"
+
+#: cms/constants/calendar.py
+msgid "Event from an external calendar"
+msgstr "Veranstaltungen aus externen Kalender"
+
```
charludo, line 153:
```suggestion
#: cms/constants/calendar.py
msgid "Event does not come from an external calendar"
msgstr "Veranstaltung kommt nicht aus einem externen Kalender"

#: cms/constants/calendar.py
msgid "Event comes from an external calendar"
msgstr "Veranstaltungen kommt aus einem externen Kalender"

```

(Nit / does not have to be exactly this)

"Veranstaltungen ohne externen Kalender" to me sounds too much like "Verantsaltungen, bei denen der externe Kalender fehlt"

charludo (approved):
Thank you, very nice! I only have two nits :)

Issue #3262, opened 2024-12-03
Hide checkboxes in archived locations for Observer Userrole


juliankehne:
### Motivation

Follow up to the implementation of #3014 in which archived locations didn't get changed.

### Proposed Solution

Remove checkboxes in Archived Locations the same way as it was done for #3014

### Alternatives

Leave as it is and live with inconsistencies.

### Related Issues

#3014 PR: #3246

Issue #3014, opened 2024-08-21
Checkbox for list views shown to users who don't have edit permission


david-venhoff:
### Describe the Bug

When a user without edit permission goes to a list view (example event list view or poi list view), but does not have permission to edit the model, the checkbox selector will still be shown:
![Screenshot 2024-08-21 at 16-58-30 Integreat Editorial System](https://github.com/user-attachments/assets/b1964d52-2cd3-41fc-8a0d-156a0bf20bb1)

Additionally, selecting the topmost checkbox will not select all other checkboxes.

### Steps to Reproduce

1. Log in as observer and open the event list or poi list

### Expected Behavior

The checkboxes should only show up if the user has permission to perform any bulk action

### Actual Behavior

Checkbox always show up, even if they cannot be used

PR #3296, opened 2024-12-17
Hide checkboxes in archived locations for observer role

lunars97:
### Short description

This is the follow up implementation of the #3014 During the implementation I forgot to remove checkboxes for archived locations as well.

### Proposed changes

- Remove checkboxes for observer role in archived locations

### Side effects

- none

### Resolved issues

Fixes: #3262

### DIFF
--- a/integreat_cms/cms/templates/pois/poi_list_archived_row.html
+++ b/integreat_cms/cms/templates/pois/poi_list_archived_row.html
@@ -2,12 +2,14 @@
 {% load rules %}
 {% load content_filters %}
 <tr class="border-t border-gray-200 hover:bg-gray-100 text-gray-800">
-    <td class="py-3 pl-4 pr-2">
-        <input type="checkbox"
-               name="selected_ids[]"
-               value="{{ poi.id }}"
-               form="bulk-action-form"
-               class="bulk-select-item" />
+    <td class="pr-2 pl-4{% if not perms.cms.change_poi %} py-5 pr-2 {% endif %}">
+        {% if perms.cms.change_poi %}
+            <input type="checkbox"
+                   name="selected_ids[]"
+                   value="{{ poi.id }}"
+                   form="bulk-action-form"
+                   class="bulk-select-item" />
+        {% endif %}
     </td>
     <td class="pr-2">
         <a title="{% if poi_translation %} {{ poi_translation.title }}{% endif %}"
--- a/integreat_cms/cms/templates/pois/poi_list_archived.html
+++ b/integreat_cms/cms/templates/pois/poi_list_archived.html
@@ -34,7 +34,9 @@
             <thead>
                 <tr class="border-b border-gray-200">
                     <th class="py-3 pl-4 pr-2 min">
-                        <input form="bulk-action-form" type="checkbox" id="bulk-select-all" />
+                        {% if perms.cms.change_poi %}
+                            <input form="bulk-action-form" type="checkbox" id="bulk-select-all" />
+                        {% endif %}
                     </th>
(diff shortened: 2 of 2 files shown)


integreat_cms/cms/templates/pois/poi_list_archived_row.html
```diff
@@ -3,11 +3,13 @@
 {% load content_filters %}
 <tr class="border-t border-gray-200 hover:bg-gray-100 text-gray-800">
     <td class="py-3 pl-4 pr-2">
-        <input type="checkbox"
-               name="selected_ids[]"
-               value="{{ poi.id }}"
-               form="bulk-action-form"
-               class="bulk-select-item" />
+        {% if perms.cms.change_poi %}
```
JoeyStk, line 6: This has a small side effect: The rows loses the padding to the top and bottom, and shrinks a little bit (s. screenshot).

Screenshot from 2024-12-17 16-23-12

I'm sure there are better ways to do it, but I couldn't come up with one and wanted to share at least one idea, so here we go :)

I think I would move the permission check one hierarchy up.


```suggestion
               {% if perms.cms.change_poi %}
                        <td class="py-3 pl-4 pr-2">
```

and then add to next element the classes in case you can't change the poi (s. below).

integreat_cms/cms/templates/pois/poi_list_archived_row.html
```diff
@@ -3,11 +3,13 @@
...
+            <input type="checkbox"
+                   name="selected_ids[]"
+                   value="{{ poi.id }}"
+                   form="bulk-action-form"
+                   class="bulk-select-item" />
+        {% endif %}
     </td>
     <td class="pr-2">
```
JoeyStk, line 14:
```suggestion
    <td class="pr-2 {% if not perms.cms.change_poi %} py-3 pl-4 pr-2 {% endif %}">
```

JoeyStk: If you chose to go with this suggestion, please don't forget to also apply this to the `poi_list_archived.html` template :)

JoeyStk (commented):
Thank you for your PR. I have one thought on the implementation. I think it's a good thing that you stuck with the implementation as we have it in the other templates. However, I can see one problem we might want to think about :) Let me know what you think :)

integreat_cms/cms/templates/pois/poi_list_archived_row.html
lunars97: @JoeyStk thank you for your suggestion :) Unfortunately, if I put it above, the whole row moves on the left side, so I used instead the class option

integreat_cms/cms/templates/pois/poi_list_archived_row.html
lunars97: the problem was with the row... so i did not change here anything :)

integreat_cms/cms/templates/pois/poi_list_archived_row.html
jarlhengstmengel: I somehow can't reproduce how it looks in the picture of @JoeyStk (Firefox/Chrome)

jarlhengstmengel (approved):
Looks good to me! Thank you!

integreat_cms/cms/templates/pois/poi_list_archived_row.html
JoeyStk: Now it looks different and fine for me as well :) I think this is ready to be merged now :)

JoeyStk (approved):
:rocket:

Issue #3304, opened 2024-12-19
Insert contact broken on test server


david-venhoff:
### Describe the Bug

On the testserver inserting a contact instead inserts the login form:
![image](https://github.com/user-attachments/assets/0e57a2b9-b43d-478e-9d3d-95462e58fd83)

### Steps to Reproduce
1. Go to cms-test.tuerantuer.org
3. Go to a page and insert contact

### Expected Behavior

Inserting a contact should work

### Actual Behavior

Inserting a contact instead inserts the login form

### Additional Information

Looks like the url used for the contact insertion is wrong: Other api calls go to `cms-test.tuerantuer.org`, but this call goes to `integreat-test.tuerantuer.org`. This is why a login screen shows up instead of a contact ^^

Found here: https://chat.tuerantuer.org/digitalfabrik/pl/z349npknajr6jkh5w7jurpjgee

Issue #3258, opened 2024-12-03
Allow users to choose what contact details are shown when inserting contacts in TinyMCE


charludo:
### Motivation

Requirement

### Proposed Solution

Idea: use URL parameters to influence what fields are contained in the rendered contact card returned by the `get_contact` endpoint.

- After selecting a contact in the TinyMCE contact insertion plugin, show (default-checked) checkboxes for each contact detail (name, email, phone,...). Which details are available can be read from the selected contact's URL, which should include all available fields (see below)
- for each checked checkbox, append a URL parameter to the contact URL. For example, if `email` and `name` are checked, append something like `?details=email,name` to the contact URL.
- adjust the `contact_card` template to check for the presence of the field names in the `details` list, rendering only the present entries

It's probably a good idea to use this opportunity to refactor the `update_contacts` function to query the `get_contact` endpoint instead of calling `render_contact_card`. 

This solution retains the contact URL as the single source of truth for the rendered contact cards, and #3169 has layn the foundation for making this a relatively easy-to-build feature.

### Alternatives

Nothing simpler IMO

### User Story

As an editor/author, I might want to include only certain contact details (for example: only name and phone number in cases where readers should make inquiries by phone, in order to dissuade them from using unsupported channels).

### Design Requirements

I don't think so, a vertical stack of checkboxes & labels should be completely functional.

### Related Issues

Depends on:
- [x] #2959

PR #3305, opened 2024-12-19
Fix contact insertion

david-venhoff:
### Short description

On the testserver we have multiple possible backend urls, and the contacts were previously loaded using their complete server url. For example, if a user was on `https://cms-test.integreat-app.de` the contact was still loaded using `https://integreat-test.tuerantuer.org`. Then the login form would be returned, because the user had of course no valid session for the other url.

So instead of using full urls, we can just use relative urls, like we do for all other links.

It's a bit hard to test this pr, because everything worked anyways on localhost, but I hope the explanation helps :)

I also considered adding `redirect: "error",` to the fetch call. This would at least have prevented that the full login form gets pasted. But maybe that is something we should add to all our fetch calls?

### Proposed changes

-  Don't use the full url

### Side effects

Should be none

### Resolved issues

Fixes: #3304

### DIFF
--- a/integreat_cms/static/src/js/forms/tinymce-init.ts
+++ b/integreat_cms/static/src/js/forms/tinymce-init.ts
@@ -106,7 +106,7 @@ window.addEventListener("load", () => {
             link_target_list: false,
             link_default_target: "",
             document_base_url: tinymceConfig.getAttribute("data-webapp-url"),
-            relative_urls: false,
+            convert_urls: false,
             remove_script_host: false,
             branding: false,
             paste_as_text: true,
--- a/integreat_cms/cms/models/contact/contact.py
+++ b/integreat_cms/cms/models/contact/contact.py
@@ -257,6 +257,18 @@ class Contact(AbstractBaseModel):
         self.point_of_contact_for = self.point_of_contact_for + " " + _("(Copy)")
         self.save()
 
+    def get_absolute_url(self) -> str:
+        """
+        Returns the absolute url to this contact:
+
+        For example::
+
+            /region_slug/contact/1
+
+        :return: The absolute url to this contact
+        """
+        return f"/{self.location.region.slug}/contact/{self.id}/"
+
     @cached_property
     def full_url(self) -> str:
         """
@@ -264,7 +276,8 @@ class Contact(AbstractBaseModel):
 
         :return: The full url
         """
-        return f"{settings.BASE_URL}/{self.location.region.slug}/contact/{self.id}/"
+        absolute_url = self.get_absolute_url()
+        return f"{settings.BASE_URL}{absolute_url}"
 
     class Meta:
         verbose_name = _("contact")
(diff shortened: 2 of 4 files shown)


charludo (commented):
Thank you for taking this on so quickly!

CUrious, do you think we could just adjust the `full_url` definition?

david-venhoff:
> Curious, do you think we could just adjust the full_url definition?

I think that should also be possible. It would probably be a bit more robust, in case the base url changes

david-venhoff:
@charludo 
I tried your approach and one problem is that apparently tinymce tries to make every url absolute. So for example if the contact cards contains the url `/augsburg/contact/5` then tinymce will convert it into `https://integreat.app/augsburg/contact/5/` and that will mean that the `get_referencing_x` functions will not find it anymore, because they only look for relative urls…

I'm not sure if there is a way to get this to work, do you have an idea?

charludo:
> @charludo I tried your approach and one problem is that apparently tinymce tries to make every url absolute. So for example if the contact cards contains the url `/augsburg/contact/5` then tinymce will convert it into `https://integreat.app/augsburg/contact/5/` and that will mean that the `get_referencing_x` functions will not find it anymore, because they only look for relative urls…
> 
> I'm not sure if there is a way to get this to work, do you have an idea?

Huh, I didn't even think to make the URLs relative - I think that may be a great solution though? Although that might mean we need to expose the contacts via API as well

charludo (approved):
I am sorry, I just re-read my comment from before the holidays, and it makes *no* sense. I have no idea what I was thinking. Super sorry about that.

Upon reconsideration, this looks like a great solution to the problem 😄 🎉

MizukiTemma (approved):

charludo:
> > Curious, do you think we could just adjust the full_url definition?
> 
> I think that should also be possible. It would probably be a bit more robust, because it would make contact cards independent of base url changes

Actually, could you implement this solution after all? 🙈 
I'm working on #3258 and it's become necessary to use the `full_url` in additional places, so I think we should solve this issue at the lowest level possible.

Sorry for the flip-flopping 😭

charludo:
@david-venhoff I made the changes we talked about, would you might taking another look if this is OK like this?

integreat_cms/static/src/js/forms/tinymce-init.ts
```diff
@@ -107,6 +107,7 @@ window.addEventListener("load", () => {
             link_default_target: "",
             document_base_url: tinymceConfig.getAttribute("data-webapp-url"),
             relative_urls: false,
```
david-venhoff, line 109: I guess we could remove this line then?

david-venhoff (commented):
Looks good to me. Lets hope that it works

integreat_cms/static/src/js/forms/tinymce-init.ts
charludo: Which one?

david-venhoff: The `relative_urls: false`
It does not really apply anymore, we keep relative urls now, so I guess it could be a bit confusing if we leave it in

Issue #3309, opened 2025-01-02
Incorrect calculation of MT budget if mid-year start of add-on package in January


MizukiTemma:
### Describe the Bug

MT budget is not correctly calculated when a region books our add-on package AND if the renewal month of the add-on starts in January, differing from the renewal date of the standard credits (50,000).

### Steps to Reproduce

1. Go to "Regions" in Network Management, select any region
2. Set "Credits renewal date for foreign language translation" to May
3. Check "Add-on package for foreign languages booked"
4. Check "Budget year start differs from the renewal date" and select January
5. See "Current total budget: 1,050,000"

* 1,050,000 = 50,000 (default budget) + 1,000,000 (add-on budget)

### Expected Behavior

"Current total budget: 383,333"

* 383,333 = 50,000 + ((4-0)/12)*1,000,000 becuase the add-on budget is given in a monthly fraction

### Actual Behavior

The fractional calculation is skipped

### Additional Information

This is a similar problem to #2598. The cause is January is actually 0.
<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

PR #3310, opened 2025-01-02
Fix MT budget calculation for add-on package booked with mid-year start in January

MizukiTemma:
### Short description

This PR fixes MT budget calculation of add-on package with mid-year start in January.

### Proposed changes

- Check explicitly for `None` to avoid January (= 0) is avaluated as if `None`

### Side effects

- None? 🙏 

### Resolved issues

Fixes: #3309 

### DIFF
--- a/integreat_cms/cms/models/regions/region.py
+++ b/integreat_cms/cms/models/regions/region.py
@@ -858,7 +858,7 @@ class Region(AbstractBaseModel):
         if not self.mt_addon_booked:
             return settings.MT_CREDITS_FREE
         # All regions which did book the add-on, but not mid-year, get the add-on credits
-        if not self.mt_midyear_start_month:
+        if self.mt_midyear_start_month is None:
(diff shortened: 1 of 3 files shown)


JoeyStk (commented):
Thank you for your PR. The code looks good :smile: 
I think it would be a good idea to have one or multiple tests for this functionality. It already broke before and I would like to ensure that we get notified when it breaks next time :) What do you think? :)

MizukiTemma:
@JoeyStk 
Yes, test is always good :) added a test for MT budget calculation.

I'm afraid there are other lines in our repository which may have the same kind of bug or bug threat. An issue for that is opened and I'll leave other tests and fixes for it. #3317

JoeyStk (approved):
Nice, looks very clean! Thank you :)

charludo (approved):
Very nice, thank you!

Issue #100, opened 2019-01-14
Events: iCal subscription


svenseeberg:
### Motivation

Municipalities are using calendars for events in their region. Until now they have been creating events twice, once in their calendar and then again in our CMS. This has been tedious for them.

### Proposed Solution

Enable subscribing to external iCal calendars. With this we mean, we want to 

- register one or many URLS to get the events from (it's own category, not in the region form)
- get the events in iCal format from this URL
- write a management command to get the iCal events and then run a cronjob for it

### Alternatives

None

### User Story

As a municipality I want to show my events in my municipality calendar and in the Integreat CMS without needing to create them twice so that users have (multi-lingual) information about the events on both the municipality's website and Integreat

### Additional Context

There are two steps for it:
1. Get the iCal events and display them only in German
2. Support MT for them (most probably this is going to be another issue and PR)

### Design Requirements

None, as this will include no UI

dkehne:
This issue becomes more and more interesting due to the feedback of e.g. Saarbrücken, Regensburg and some others who already want to use their city-calendars to be (partly) imported into Integreat.

JoeyStk:
My idea for this feature would be the following: 

1. Add an import section for events, similarly to the section we have for XLIFF import for pages ![grafik](https://github.com/digitalfabrik/integreat-cms/assets/72705147/95c902b8-c549-45e9-bbc2-6f6fa698ec86)
2. After clicking importing you get a preview on how the data would be imported, again similarly as the have it for XLIFF import
3. After checking if everything looks good, you can confirm and the import starts

Do you have any objections to this workflow or did I forget something?

After taking a look at a few examples for iCal files, I saw that also locations can be filled in. So I assume that other than creating new events also new POIs should be created, if necessary, right?

osmers:
@JoeyStk I like the workflow - easy enough.
Question: location is compulsory for creating an event - if the ical does not have location information, what would happen? 
And yes to your question - if location data is filled in, a new poi should be created. But there should be a step that checks, whether a poi at that location already exists, right?

JoeyStk:
> if the ical does not have location information, what would happen?

Good question, I think we should take the time to think about these things. I would say the import is rejected then, the user is redirected to the event list view and told via a message that they can't import an event that doesn't have a location either a new one or an existing one

> if location data is filled in, a new poi should be created

[...]

PR #3313, opened 2025-01-03
Add release note for external calendars

JoeyStk:
### Short description

This PR adds the release note for the external calendars. This PR is one of the blocker for the next release. However we should wait for the final testing of @juliankehne until we merge this PR.

### Proposed changes

- Add release note.  

### Side effects

- I used the initial issue #100 as the issue id. Do you think another issue would be better suited?

### Resolved issues

Fixes: #100 

### DIFF
--- a/integreat_cms/release_notes/current/unreleased/100.yml
+++ b/integreat_cms/release_notes/current/unreleased/100.yml
@@ -1,0 +1,2 @@
+en: Add option to import events from external calendars via iCal
+de: Füge die Möglichkeit hinzu, Veranstaltungen aus externen Kalender über iCal zu importieren


MizukiTemma:
@JoeyStk 
Do you mean we release and reveal the feature? Then permission adjustment may be needed, because currently only Service Team and CMS Team can see, change and delete external calendars (unless it is so planned).

JoeyStk:
AFAIK for now that is the desired status and only members of the service team should be able to add new calendars. But maybe @osmers can give us an update if that's still the plan :)

osmers:
Yep, that's still the plan for now :) do we need a release note then or would it only confuse municipalities when they see the note but then cannot do anything?

osmers:
@JoeyStk alrighty, that's ok by me - I think most municipalities will rather read the chat/blog than the release notes anyways :P

integreat_cms/release_notes/current/unreleased/100.yml
```diff
@@ -0,0 +1,2 @@
+en: Add option to import events from external calendars via iCal
+de: Füge die Möglichkeit hinzu Veranstaltungen aus externen Kalender über iCal zu importieren
```
PeterNerlich, line 2:
```suggestion
de: Füge die Möglichkeit hinzu, Veranstaltungen aus externen Kalender über iCal zu importieren
```

PeterNerlich (approved):

Issue #2957, opened 2024-08-01
Create contacts from existing information


MizukiTemma:
### Motivation

There are already a lot of contact informations (who and how to reach for a certain thema) written in the contents. It would be very helpful if there were a way to convert those into our new model "contact" so munincipality users do not have to manually create all the contact information.

### Proposed Solution

TBD

Every idea is welcomed 😸

MizukiTemma:
One option may be orientating us with link objects: suggest users to create a contact object for E-mails and phone numbers that are embedded in a content (and therefore scanned by the link checker) but not registered in any contact object.

Sadly this is not automated solution, bt at least users know where a potential target is.

JoeyStk:
It's an important ticket, but it doesn't **have to** be done by November 7th. A few weeks later is still fine :)

Issue #3280, opened 2024-12-09
Register a new URL and Link in Linkcheck whenever a contact is inserted in TinyMCE


charludo:
### Motivation

Contacts inserted into TinyMCE in #3169 are picked up by linkcheck *eventually*. But for users, it is not at all apparent why this doesn't happen immediately, potentially leading to confusion when they have inserted a contact in a page, but that page is not shown in the contact's list of pages it is embedded in.

It can also lead to potential inconsistencies:
- insert contact "A" into a page, save
- go to "A"s contact details
- change "A"s name to "B"
- most likely, Linkcheck has not caught the embedding of the contact in the page yet, and so that page will not be updated, still showing "A".

### Proposed Solution

- the creation of a contact through the Contact Form should create a linkcheck `URL` object, where the URL in question is the contact's `full_url`. It should either be checked immediately, or set to be valid during its creation
- inserting the contact into TinyMCE should query that `URL` object, and create a new related `Link` object

Essentially, force linkcheck to recognize the contact.

### Alternatives

None, really

### Related Issues

Blocked by:
- [x] #2959

charludo:
.....is this required? I just implemented a "fix" under the linked branch, but, uhm.....

Could someone else confirm that this is even an issue right now? Because it might just be that my dev environment has `LINKCHECK_DISABLE_LISTENERS=1`, since if I change it to `0`, contacts links are immediately updated 🤦🏼‍♀️ 

Please, if someone else can confirm this would be great

osmers:
@charludo I was just testing the contact functionality and also found the same two problems, i.e. the contact not updating automatically in a page it's inserted in and the page not showing up in the contact's linked pages. So I'd say it's definitely a problem :)

charludo:
> [@charludo](https://github.com/charludo) I was just testing the contact functionality and also found the same two problems, i.e. the contact not updating automatically in a page it's inserted in and the page not showing up in the contact's linked pages. So I'd say it's definitely a problem :)

Laright, thank you for the feedback - where did you experience this? The test cms?

osmers:
@charludo yes, the test cms. I created the POI Herkules in the Testumgebung and created two contacts related to it.

charludo:
> [@charludo](https://github.com/charludo) yes, the test cms. I created the POI Herkules in the Testumgebung and created two contacts related to it.

Oof. Just tried it, and yep, same. Oh no.

charludo:
[...]

PR #3314, opened 2025-01-03
Suggest users potential contact sources

MizukiTemma:
### Short description

This PR introduces a hint page to help users to create new contacts and replace embedded E-mail addresses and phone numbers with contact cards. This is not an amtomated solution, but I hope it can help and guide users at least. This PR is opened as suggestion of an approach and does not mean to be a finalized something. I'm open to any idea 😸 

### Proposed changes

- Create a new page "Potential contact data"
- Add a link to it next to the title of contact list with 💡 icon
- In "Potential contact data": List embedded E-mail addresses and phone numbers per page/event/location *1
- In "Potential contact data": List related contacts (either E-mail or phone number matched) too as suggestion

*1: not per translation, as the list would be then too long with a lot of duplicates. All E-mail addresses and phone numbers in all translations of the page/event/content will be shown but without duplicate even if any of them appear multiple times in multiple of translations.

### Side effects

Those points are still in progress and will be worked on, once this approach receives agreement from the team and we decide for details.
- Naming, sentences, explanations and their translation, appearance, design etc.
- Queries in the view
- Preview of the page/event/location and/or link to their form (👁️ and ✏️ icons next to content title)
- How to show suggested contact (maybe as contact card, or simply `__str__` with link?)

### Resolved issues

Fixes: #2957 


MizukiTemma:
@osmers 
Currently it looks like this


Users are expected to click here
contact suggestion 1

No suggestion is shown if no contact targets are found
contact suggestion 2

If any detected, related contacts are shown
contact suggestion 3

osmers:
> Users are expected to click here

I would not have guessed that :D somehow the sentence leads me to believe that there is an information page (e.g. from the wiki) behind it, that explains, where I can use contacts) - maybe we can rephrase it?
E.g. Vorschläge für Kontakte, die Sie erstellen können @nikolahoff maybe for a better phrasing that suggests to users that they can find suggestions on how to use the feature there



> No suggestion is shown if no contact targets are found

Can you change it to Telefonnummer? :) 



> If any detected, related contacts are shown

So this only shows up with they already have a contact created, that uses the same information, right?
Is it feasible or even useful that maybe we show them lists of contact details that are in the content, but don't have a contact yet? Probably additionally to what you have already done here....

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -5385,10 +5388,41 @@ msgstr "Kontakt wiederherstellen"
...
+"centrally managed and updated in future."
+msgstr "Hier ist die Liste der E-Mail-Adressen und Telefonnummern, die in Inhalte eingebettet sind. Wir empfehlen, diese durch eine Kontaktkarte zu ersetzen, damit sie künftig zentral verwaltet und aktualisiert werden können."
+
+#: cms/templates/contacts/contact_from_email_and_phone.html
+msgid " pages have potential contacts."
+msgstr " Seiten haben mögliche Kontakte."
+
+#: cms/templates/contacts/contact_from_email_and_phone.html
+msgid "No E-mail address and phone number detected"
+msgstr "Keine E-mail Adresse und Telephonenummer gefunden."
```
JoeyStk, line 5408:
```suggestion
msgstr "Keine E-Mail-Ad­res­se und Te­le­fon­num­mer gefunden."
```

JoeyStk:
> I would not have guessed that :D somehow the sentence leads me to believe that there is an information page (e.g. from the wiki) behind it, that explains, where I can use contacts) - maybe we can rephrase it?

I agree with @osmers. I also only got it because of your explanation. My first suggestion to rephrase it would be simply:
"Analysis of content for existing contacts in" and for the German translation "Inhaltsanalyse für bestehende Kontaktdaten". What do you think about? :)

osmers:
> 4\. This goes back a little bit to 2). I think I would have expected a button or something to convert the found phone number or email to a contact, if there is no contact found which already contains this information. A wonderful bonus addition would be, if on click on the button the information already gets transferred to the form and would be filled out already.

Yeah, this is what I meant with my last point as well :) would make a lot of sense to make creating new contacts much easier and faster.

MizukiTemma:
> > No suggestion is shown if no contact targets are found
> 
> Can you change it to Telefonnummer? :)
> 

Done, thank you @JoeyStk

MizukiTemma:
@JoeyStk 
 Thank you for your review 👍 
 
 1. Icon
Planned are preview at 👁️ and a link to the page/event/location edit form at ✏️. The later is added now. As our preview functionality is currently only avaialble for pages, it is not yet implemente in this PR. If this idea gets confirmed, the preview function must be first extended for event and page, then added to the icon.

JoeyStk:
> Icon
Planned are preview at 👁️ and a link to the page/event/location edit form at ✏️. The later is added now. As our preview functionality is currently only avaialble for pages, it is not yet implemente in this PR. If this idea gets confirmed, the preview function must be first extended for event and page, then added to the icon.

Ah, thank you for clarifying

> 2. User navigation
Alternatively to the icon ✏️ next to the content title, we can add here a button "Go to page/event/location" which brings users to the edit form of the content.
And/or another button to create a new contact per ajax? (for this we will need first https://github.com/digitalfabrik/integreat-cms/pull/3196 ) 

I think that the "button"/CTA should be to create a contact from the found information. Therefore I'd rather stick to the icon?

> 3. Search per keyword

Yeah, right. I also don't have an idea for this :sweat_smile: 

> 4. Do you mean putting detected phone number and E-mail address into a contact form?

Yes, my idea would be to have button like "Convert to contact" right of the phone number / e-mail address. When I click on the button I get to a contact form and the found information (so the found phone number/e-mail address) is already filled out, which makes it easier for me and I don't have to copy/paste anything? I'll try to prepare a commit for this, to make it clear what I mean exactly :) 

> Automatic contact generation
I dreamed of this 😓
Automatic creation is not possible, as every contact requires a POI.

Yes, that is not what I meant. I think it's a good thing, that the municipalities still have to go through the list of found information, but I would like to make it as easy for them as possible :)

JoeyStk:
@MizukiTemma I've now added a commit for the minimal effort "convert to contact" functionality. If you find the time, please let me know what you think about it and feel free to drop it again, if you don't like it :)

integreat_cms/cms/utils/link_utils.py
```diff
@@ -24,3 +28,18 @@ def fix_content_link_encoding(content: str) -> str:
     :return: The fixed content
     """
     return re.sub(r"(?<=[\"'])(https?://.+?)(?=[\"'])", fix_domain_encoding, content)
+
+
+def clean_url(url: Url) -> str:
```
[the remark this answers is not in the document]
MizukiTemma: Hmm, it depends how to go after that. Icons are just .svg data. They are almost always contact targets but it's faster to check for E-mail or phone links to show what E-mail address or number are embedded 🤔

MizukiTemma:
> @MizukiTemma I've now added a commit for the minimal effort "convert to contact" functionality. If you find the time, please let me know what you think about it and feel free to drop it again, if you don't like it :)

That looks nice 😸 I would name it rather "Create a contact from this information" or something, because the "Convert to contact" sounds as if a contact would be generated automatically and/or the link would be replaced with it.

integreat_cms/cms/templates/contacts/contact_list.html
```diff
@@ -10,6 +10,10 @@ <h1 class="heading">
                 {% else %}
                     {% translate "Contacts" %}
                 {% endif %}
+                <a href="{% url 'potential_targets' region_slug=request.region.slug %}"
+                   class="font-bold text-sm text-gray-800 flex items-center gap-1 mb-2 hover:underline">
+                    <i icon-name="lightbulb" class="align-baseline"></i>{% trans "Analysis of content for existing contacts" %}
+                </a>
```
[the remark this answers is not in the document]
MizukiTemma: Hmm, could you directly commit it?

integreat_cms/cms/utils/link_utils.py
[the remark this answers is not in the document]
JoeyStk: Makes sense!

MizukiTemma:
@osmers This is the current state.

Users find the link "Inhaltsanalyse für bestehende Kontaktdaten" (en: Analysis of content for existing contacts)
contact 0

After clicking it, the suggestion list appears.

It looks like this if there is no E-mail address or phone number detected
contact 1

and this, if any suggestion exists. All entries are closed as default.

contact 2

Users can open and see the detail
contact 3

By clicking "Zu Kontakt konvertieren", users will be taken to the contact form and can create a new contact
(The screen shot below is the case when the link next to "hello-google @ translate.me" is clicked. The selected address or number is input automatically.
contact 4

How do you find it? We can also merge this PR (it's anyway available only for us) so Searvice Team can actually test around.


**note**
One thing to fix is formatting of phone number in the contact form. Currently numbers will not be automatically reformatted to "+49........." in the contact form but in the contents. This causes that +49123456789 in a content will not be matched with a contact object that has 0123456789. We have to either not reformatt all over our system or reformatt always everywhere (or take both format into consideration while searching for possible matches).

osmers:
@MizukiTemma thank you for all the screenshots and explanation - @JoeyStk took me through the function on his system as well and we tried out a few different things which really helped to get a clearer picture :) 
We found a few things, e.g. email address is input into the phone number field when clicking convert, already added contacts are a bit hard to match to the shown contact details on the left (there needs to be some visual distinction), and when the contact is then inserted into the page as a contact card, it still shows up in the list. I think @JoeyStk will write a more detailed reply with examples :) 
That all said - I think it's great that we are offering something to the municipalities to help them in the process - so thank you very much for what you have already done on this PR :) 💪

JoeyStk:
Finally, here comes the list:
- When converting an e-mail-address to a contact, then it's also shown as a phone number. This happens, because in my quick commit I put everything in the phone field, as I thought that it was okay for a quick demonstration. Do you want to fix this or should I? :) 
- The phone number is a `CharField`, this means we don't have verification the input is really an phone number. Can we add something to make sure of this?
- It is difficult to see understand which contact information belongs to which contact (there needs to be some visual distinction)
- Once a contact is inserting, the according phone number and e-mail shouldn't be shown in the list anymore
- Could we convert the list to a table? Hopefully with this it's clearer what belongs together and what not :)
- Because we use the collapsible box template it expects an icon and because we haven't added one, now there is quite a bit of blank space next to the headline
- "Convert to contact" should disappear when an contact to this information already exists

MizukiTemma:
> "Convert to contact" should disappear when an contact to this information already exists
We don't know if any of matched contacts is a good replacement of the target E-mail address/phone number. Users maybe want to create a new contact. I'm for leaving "Convert to contact" always.

> Once a contact is inserting, the according phone number and e-mail shouldn't be shown in the list anymore
Ah, I see, the phone number and/or E-mail addresss and/or POI of an embedded contact card are detected as link and listed (both in the contact suggestion and the broken link list). Related to #3280?

MizukiTemma:
> It is difficult to see understand which contact information belongs to which contact (there needs to be some visual distinction)

What about showing matched contacts as card? (that was once suggested in the discussion, if I remember correctly)

image

osmers:
> We don't know if any of matched contacts is a good replacement of the target E-mail address/phone number. Users maybe want to create a new contact. I'm for leaving "Convert to contact" always.

I'm not so sure - I see your point about the system not knowing if the contact is a good match/replacement, but on the other hand, users don't see any progress that way....

osmers:
> > It is difficult to see understand which contact information belongs to which contact (there needs to be some visual distinction)
> 
> What about showing matched contacts as card? (that was once suggested in the discussion, if I remember correctly)
> 
> image

Better - however I still need to check which contact card belongs to which contact detail - could we maybe style it, so that it works as a table with columns? That would mean that in this example the phone number, which belongs to the second contact card, would show up further down so that it is in line with the beginning of the contact card it belongs to. What do you think? And if there is more than one contact detail that belongs to the card, they should be placed directly below each other (in the same row as the card). If they are found in more than one card, they should also show up more than one, imho.

MizukiTemma:
> > > It is difficult to see understand which contact information belongs to which contact (there needs to be some visual distinction)
> > 
> > 
> > What about showing matched contacts as card? (that was once suggested in the discussion, if I remember correctly)
> > image
> 
> Better - however I still need to check which contact card belongs to which contact detail - could we maybe style it, so that it works as a table with columns? That would mean that in this example the phone number, which belongs to the second contact card, would show up further down so that it is in line with the beginning of the contact card it belongs to. What do you think? And if there is more than one contact detail that belongs to the card, they should be placed directly below each other (in the same row as the card). If they are found in more than one card, they should also show up more than one, imho.


Do you mean: there is a phone number N and an E-mail address E in a page. There are three contacts (1, 2, 3) which has both N and E, another three (4, 5, 6) only with N and other three (7, 8, 9) only with E. In this case:

| contact detail | matched contacts|
|----------------| --------------------|
| N | Card 1 |
|  | Card 2 |
|  | Card 3 |
|  | Card 4 |
|  |Card 5 |
|   |Card 6 |
| E  | Card 1 |
|   | Card 2 |
|   |Card 3 |
|   |Card 7 |
|   |Card 8 |
|   |Card 9 |

I'm afraid of too much duplicate (in the above example cards 1, 2, 3 are shown twice). The more contact details found in a page, the more duplicate can be caused. We may think then we can shouw contact cards that contains both N and E only once on the top of the list, but the number of possible combinations of one number and one E-mail increases rapidly if there are more than two.

MizukiTemma:
@JoeyStk @osmers 
>The phone number is a CharField, this means we don't have verification the input is really an phone number. Can we add something to make sure of this?

A new issue is opened, as it's outside of scope of this issue #3381

JoeyStk:
Thank you a lot for your improvements. I think this is already better! I agree with @osmers that the user guidance could be still be a bit better, but also see your points. I have one last idea for a compromise:
What do you think about adding a checkbox in front of the found phone number or e-mail address and if the user clicks on the checkbox and marks it as "complete", then the phone number or e-mail address disappears. This might increase the user guidance as the user is "rewarded" for creating contacts and still we don't leave it to the machine but to a human instead to decide if something is done or not - is this a compromise?

After that I think this PR is done! Thank you very much for tackling this HUGE issue, especially as we didn't know what exactly we wanted at the beginning. This is very valuable work :yellow_heart:

MizukiTemma:
@JoeyStk 
> What do you think about adding a checkbox in front of the found phone number or e-mail address and if the user clicks on the checkbox and marks it as "complete", then the phone number or e-mail address disappears. 

Do you mean phone numbers and E-mail addresses that users do not want to replace with a contact card? Phone numbers and E-mail addresses disapper from the list when they are removed from the content (because replaced by contact card, or whatever)

JoeyStk (approved):
I think for the first iteration this is good. Later we can expand this, once we have some more feedback, e.g. from the municipalities themselves

MizukiTemma:
There are some updates.

- `clean_url` is moved to `contact_from_existing_data.py`
- Filter by the region when collecting related contacts
- Extract the phone number formatting by #3386 into an independent util function, as this is needed in `contact_from_existing_data` too.

charludo (changes requested):
Thank you, this is a very important topic and a really nice approach! 🎉  Apart from the small details below, I do have some concerns though.

I intentionally did not read the previous discussion on this PR before trying everything out, to go into this just as a user might be. And to be honest, it was a bit confusing 😢 Specifically:

- [ ] clicking "convert to contact" on a phone number will open the new contact page with the phone number pre-filled, but without the leading "+" - this gets converted to a wrong phone number! E.g.: if the number is `+49 (0) 17212345`, then the contact form is prefilled with `49 (0) 17212345`, which gets turned into `+49 (0) 49017212345` after saving. I did not notice this, and as a result was very confused why it still said "There is no contact suggested." for a couple of minutes 😄 
- [ ] apropos, I suggest reformulating "There is no contact suggested." to "No contact matching these details exists." Makes it clearer that there really IS no contact, not just that we don't know which one to replace it with.
- [ ] If a contact card is shown, it is not clear WHY it is shown. I know that it's there to say "this contact matches some of the details to the left", but IMO that is not intuitively understandable. Maybe you could add two headers, something like "Phone numbers and emails found in the page" and "Existing contacts matching the found details" above each left/right column?
- [ ] "Convert to contact" still being shown when a contact card for the same detail is being shown is confusing. I get that we do not 100% know if that really is the right contact, but I would at least indicate that a match was found next to the detail - for example, "(1 potential match found)" between the detail and the "convert to contact" link
- [ ] The pen icon alone is not really noticable enough. I would either add some text, or even better, add a big blue button below the found contact details with "Edit this page to exchange details for contacts". That could even be disabled if no contact cards are found for the details, with a corresponding hint on hover.
- [ ] The entire view looks and feels like it should be possible to click "replace with this contact card!" and magically have the details replaced. I know that is not feasible, but users will be expecting it. I think a bold text needs to be added, stating something like "Please note that contact details need to be replaced with contact cards manually. Found details will be removed from the list below once you have replaced them with a contact card in the corresponding page, poi, or event."

I think these are relatively small fixes, and doable by the release. The following ones are not, so I am not suggesting that you should implement them, it's just some ideas for future tweaks to this page:

- maybe add a green check mark/red X to the left of details, when a contact card matching *that specific detail* is found/not found. Makes it immediately obvious which details still need a contact.
- clicking "convert to contact", creating the contact, then clicking "save" really should redirect back to the "potential contact data" view, not the contact details view.
- the contact card should be interactable. It's the largest thing in each collapsible box, and it is just kinda strange that it does nothing 😄 At the very least, hovering over it should highlight which details it can replace.
- there needs to be a way to mark / hide false positives.



Sorry for the wall of text 😅

MizukiTemma:
>The pen icon alone is not really noticable enough. I would either add some text, or even better, add a big blue button below the found contact details with "Edit this page to exchange details for contacts". That could even be disabled if no contact cards are found for the details, with a corresponding hint on hover.


I leave the button always clickable (un-disabled? 😁) as users probably want to see the content but the preview function is implemented cuttently only for pages.

charludo:
> > The pen icon alone is not really noticable enough. I would either add some text, or even better, add a big blue button below the found contact details with "Edit this page to exchange details for contacts". That could even be disabled if no contact cards are found for the details, with a corresponding hint on hover.
> 
> I leave the button always clickable (un-disabled? 😁) as users probably want to see the content but the preview function is implemented cuttently only for pages.

Ah, makes sense!

integreat_cms/cms/templates/contacts/contact_from_email_and_phone.html
```diff
@@ -0,0 +1,64 @@
...
+{% load static %}
+{% block content %}
+    <div class="flex justify-between mb-4">
+        <h1 class="heading">
+            {% translate "Potential contact data" %}
+        </h1>
+    </div>
+    <div class="pb-4">
+        <span class="pb-4">
+            {% translate "Here is the list of E-mail addresses and phone numbers that are embedded in contents. We recommend to replace them with a contact card so they can be centrally managed and updated in future." %}
```
charludo, line 12:
```suggestion
            {% translate "Here is the list of E-mail addresses and phone numbers that are embedded in contents. We recommend to replace them with a contact card so they can be centrally managed and updated in the future." %}
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -5436,10 +5438,76 @@ msgstr "Kontakt löschen"
...
+#: cms/templates/contacts/contact_from_email_and_phone.html
+msgid ""
+"Here is the list of E-mail addresses and phone numbers that are embedded in "
+"contents. We recommend to replace them with a contact card so they can be "
+"centrally managed and updated in future."
+msgstr ""
+"Hier ist die Liste der E-Mail-Adressen und Telefonnummern, die in Inhalte "
+"eingebettet sind. Wir empfehlen, diese durch eine Kontaktkarte zu ersetzen, "
+"damit sie künftig zentral verwaltet und aktualisiert werden können."
+
```
charludo, line 5454:
```suggestion
"Hier ist die Liste der E-Mail-Adressen und Telefonnummern, die in Inhalten "
"eingebettet sind. Wir empfehlen, diese durch eine Kontaktkarte zu ersetzen, "
"damit sie künftig zentral verwaltet und aktualisiert werden können."

```

charludo (approved):
Thank you for all the fixes!! I only have two nits remaining, plus the icon styling from last time, but I will already approve now so this can be merged as soon as those are done 🙂 Thank you!! 🎉

integreat_cms/locale/de/LC_MESSAGES/django.po
charludo: Ohhhh I just realized there's two ways to read this, you can ignore my comment. Sorry 😄

MizukiTemma:
Icon and link should stand in the same line now

MizukiTemma:
Not all the suggestions are implemented, but I'll send it to the test server already, so this feature can be tested as early as possible. Improvements that are not yet impelmented will be collected in a separate issue.

charludo:
> Not all the suggestions are implemented, but I'll send it to the test server already, so this feature can be tested as early as possible. Improvements that are not yet impelmented will be collected in a separate issue.

Sounds good! And sorry, I know those were a lot of last-minute requests, thank you for honoring them!!

PR #3320, opened 2025-01-08
Allow users to select shown contact details

charludo:
### Short description

This PR allows users to choose which contact details are shown when they insert a contact into TinyMCE, as well as when updating the contact later on.

### Proposed changes

- use a GET parameter to encode which contact details should be shown on a per-contact-card basis
- show appropriate checkboxes after selecting a contact through Tomselect

### Side effects

- styling the checkboxes was... a challenge, since I frustratingly was not able to target them through the `tinymce.css` file - if someone knows why/how to fix this, I am all ears!
- suboptimal from a UX perspective: if a user selects a contact, deselects some details, then chooses a different contact, the details selection is reset. Not sure if this is worth the effort to change, though, since the shown checkboxes are necessarily contact-specific

### Resolved issues

Fixes: #3258 

### DIFF
--- a/integreat_cms/cms/models/contact/contact.py
+++ b/integreat_cms/cms/models/contact/contact.py
@@ -236,6 +236,34 @@ class Contact(AbstractBaseModel):
             for link in Link.objects.filter(url__url=self.absolute_url)
         ]
 
+    @cached_property
+    def available_details(self) -> dict:
+        """
+        Returns the available details and their human-readable representation
+
+        :return: key-value pairs of available detail and human-readable representation
+        """
+        details = {
+            "address": _("show address"),
+        }
+
+        if self.point_of_contact_for:
+            details["point_of_contact_for"] = _("show point of contact")
+
+        if self.name:
+            details["name"] = _("show name")
+
+        if self.email:
+            details["email"] = _("show email")
+
+        if self.phone_number:
+            details["phone_number"] = _("show phone number")
+
+        if self.website:
+            details["website"] = _("show website")
+
+        return details
+
     def archive(self) -> None:
         """
         Archives the contact
--- a/integreat_cms/cms/templates/contacts/contact_card.html
+++ b/integreat_cms/cms/templates/contacts/contact_card.html
@@ -3,52 +3,54 @@
 {% spaceless %}
     <div contenteditable="false"
          data-contact-id="{{ contact.pk }}"
-         data-contact-url="{{ contact.absolute_url }}"
+         data-contact-url="{{ contact.absolute_url }}?details={{ wanted|join:"," }}"
          class="contact-card notranslate"
          dir="ltr"
          translate="no">
         {# djlint:off H021 #}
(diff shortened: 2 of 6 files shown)


JoeyStk:
> suboptimal from a UX perspective: if a user selects a contact, deselects some details, then chooses a different contact, the details selection is reset. Not sure if this is worth the effort to change, though, since the shown checkboxes are necessarily contact-specific

This might be a question for @osmers :)

JoeyStk (approved):
Thank you very much! 
> styling the checkboxes was... a challenge, since I frustratingly was not able to target them through the tinymce.css file - if someone knows why/how to fix this, I am all ears!

I don't, but I think it's still readable and understandable, so I would be okay with leaving it as is :)

osmers:
> > suboptimal from a UX perspective: if a user selects a contact, deselects some details, then chooses a different contact, the details selection is reset. Not sure if this is worth the effort to change, though, since the shown checkboxes are necessarily contact-specific
> 
> This might be a question for @osmers :)

I don't see a problem with that - it's actually something I would expect, as for a new contact I might want to choose different details to show. So no need to change that from my pov.

osmers:
@charludo not sure if this is the right place for the question/comment:
Once I have inserted a contact, I have no way of getting back to the selection of information I want to show - meaning I need to either delete the contact and insert again or change the contact. 
If I really want to change the contact then the behaviour is fine - but sometimes I might just want to change which information for the already selected contact is shown. Is there any way, that we can take users back to that selection?

charludo:
Oops, yes, looks like there's a bug. Normally the contact, with the ability to change the selection, should be shown when clicking on "change contact". I will take a look later today.

osmers:
@charludo could we maybe change it so that I don't need to click "change contact" first - I feel that that might not be very intuitive as it might lead users to believe that the button will allow them to change to a different contact altogether. Or maybe we can rename the button to e.g. change contact details?

charludo:
> @charludo could we maybe change it so that I don't need to click "change contact" first - I feel that that might not be very intuitive as it might lead users to believe that the button will allow them to change to a different contact altogether. Or maybe we can rename the button to e.g. change contact details?

I know the button is broken right now, but normally it should work like this:

- click "change contact" on the contact card
- get the popup for selecting a contact, but with the current contact pre-selected and the selection of details reflecting the current contact card
- but users are indeed free to search for a different contact and replace the current one witht a new one and new details

osmers:
@charludo yeah, I get the intended functionality of the button - I just thought that maybe the naming of the button could be misleading and not prompt users to understand that they can also change which details they need/want for the current contact.

charludo:
> @charludo yeah, I get the intended functionality of the button - I just thought that maybe the naming of the button could be misleading and not prompt users to understand that they can also change which details they need/want for the current contact.

Oh OK, sorry, I misunderstood. "Change Contact or Details"?

osmers:
Yeah, maybe that would be a good wording and we'll see what feedback we get - accordingly we could change the wording again.

Issue #3279, opened 2024-12-09
Export only selected page access statistics


juliankehne:
### Motivation

Currently the csv export of page access statistics always exports everything (languages, total access etc.). However when a user deselects languages or other data (Offline Accesses, WebApp Accesses, Total Accesses) it should also only export this data as a csv, since that is the data the user wants to see.

### Proposed Solution

Export the page access statistics based on the 'Shown Languages' and 'Access' checkboxes selected.

### Alternatives

Keep it as it is.

### User Story

As a manager I want to export only the selected languages so that I don't have to do more work after the export.

### Additional Context

![2024-12-09 14_01_46-Integreat Editorial System — Mozilla Firefox](https://github.com/user-attachments/assets/f3b1dd06-8dfa-4b73-848c-9e1aea7b4fd8)

### Related Issues

Discussed in https://github.com/digitalfabrik/integreat-cms/issues/3133

PR #3324, opened 2025-01-10
Show only selected data in CSV export

MizukiTemma:
### Short description

This PR fixes the CSV export of statistics so it does not contain de-selected data (language, access type) anymore.

### Proposed changes

- Check whether the data is set to "visible" at the moment of export

### Side effects

- None?

### Resolved issues

Fixes: #3279 

### DIFF
--- a/integreat_cms/release_notes/current/unreleased/3279.yml
+++ b/integreat_cms/release_notes/current/unreleased/3279.yml
@@ -1,0 +1,2 @@
+en: Export only statistics for selected languages
+de: Exportiere nur ausgewählte Seitenzugriffsstatistiken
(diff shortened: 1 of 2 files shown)


jonbulz (approved):

integreat_cms/release_notes/current/unreleased/3279.yml
```diff
@@ -0,0 +1,2 @@
+en: Export only selected page access statistics
```
JoeyStk, line 1:
```suggestion
en: Export only statistics for selected languages
```
Since this is mostly about the language selection, I find this to be a bit more speaking, but I might be wrong as well. Also I think "page access statistics" might be a bit confusing, since to me it carries the same meaning as "seitenbasierte Zugriffe", and those are not yet implemented

JoeyStk (approved):
Thank you very much :) This fixes the issue :rocket:

Issue #3212, opened 2024-11-18
Select EventLocation for Event Filter


Nbamy:
### Steps to Reproduce

1. Go to '.Veranstaltung..'
2. Click on 'Filter einblenden....' 
3. select " Vergangegen Veranstaltungen"
4. by Event location enter Bürger
5. see the location options '....'
4. See error Location not selectable

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

Nbamy:
![Screenshot (99)](https://github.com/user-attachments/assets/4f56f586-d6bc-4212-b52f-b23493838af4)

PR #3329, opened 2025-01-13
Fix location selection in the event filter

MizukiTemma:
### Short description

This PR fixes the bug that none of the suggested locations can be selected in the event form.

### Proposed changes

- Putting the ID of selected POI to the input field to hand over to the filter
- Fix `TypeError: Cannot set properties of null` at `.outerHTML`

### Side effects

-  I hope none 🙈 Please check there is no side effect on ajax POI creation in the event and contact form too.

### Resolved issues

Fixes: #3212 

### DIFF
--- a/integreat_cms/static/src/js/poi-box.ts
+++ b/integreat_cms/static/src/js/poi-box.ts
@@ -1,7 +1,7 @@
 import { createIconsAt } from "./utils/create-icons";
 import { getCsrfToken } from "./utils/csrf-token";
 
-type FormResponse = { success: boolean; poi_address_container: string };
+type FormResponse = { success: boolean; poi_address_container: string; poi_id: string };
 
 const showContactFieldBox = () => {
     const contactFieldsBox = document.getElementById("contact_fields");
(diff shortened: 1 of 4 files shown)


integreat_cms/static/src/js/poi-box.ts
```diff
@@ -31,7 +35,12 @@ const hidePoiFormWidget = () => {
 
 const setPoi = ({ target }: Event) => {
     const option = (target as HTMLElement).closest(".option-existing-poi");
-    renderPoiData(option.getAttribute("data-poi-title"), option.getAttribute("data-poi-address"));
+    console.log(option);
```
charludo, line 38:
```suggestion
```

😁 

(Huh, eslint/prettier should in theory catch this, I have a look why it's not working)

charludo (approved):
Thank you, this works perfectly! 😄 🎉 

Also checked the event/contact ajax POI creation, those (still) work perfectly as well 👍🏼

JoeyStk (commented):
Thank you! LGTM!
Do you think it makes sense to add a test for this? :)

JoeyStk (approved):

osmers:
Quick question - was this released already? It should have been judging from when it was merged, but it only works on the test system and there's no release note concerning this PR. @MizukiTemma @JoeyStk

MizukiTemma:
> Quick question - was this released already? It should have been judging from when it was merged, but it only works on the test system and there's no release note concerning this PR. @MizukiTemma @JoeyStk

This was released in the last release 🤔 We'll clarify in the team.

osmers:
Thank you :)

MizukiTemma:
@osmers

This branch was merged shortly after the last release. It will come in the next release 🙈

osmers:
@MizukiTemma thanks for checking - that somehow wasn't clear to me from the time stamps here in GitHub but I now found where I can check the exact time an action was taken :) next time I'll do that :)

Issue #3252, opened 2024-11-29
Sort order for contact list view not specified


david-venhoff:
### Describe the Bug

Currently, the contacts in the contact list view are sorted according to their name. This is problematic for two reasons:
* Contacts for the same poi might be very far apart
* Since the name is allowed to be blank, there will be many contacts without a name. For those contacts, the sort ordering is not really specified then

### Expected Behavior

I am not sure what the best solution is. Maybe first sort by poi, then by name? This would make sure that contacts that belong to the same poi are listed together

JoeyStk:
@osmers / @nikolahoff Does one of you have a preference on how the contacts should be ordered? For example should contacts belonging to the same POI be grouped - or do you have a different preference for sorting? :)

osmers:
I think grouping by POI makes sense and then order the POIs by alphabetic order?

nikolahoff:
Yes grouping by POIs is a good idea, and within the POI the contact names are ordered alphabetically right?

PR #3334, opened 2025-01-14
Sort contacts by group and name

MizukiTemma:
### Short description

This PR changes the sort order of contacts so that they are first grouped by POI and then sorted by name alphabetically.

### Proposed changes

- Add "location" to `ordering`

### Side effects

- should be none 🙈 

### Resolved issues

Fixes: #3252 


JoeyStk (commented):
Thank you very much! This looks good already! :)

I only ran into one problem. The ordering by location is already working, but the locations are not sorted alphabetically but instead by poi_id. In my case "Test" was sorted before "Entwurf", as it had the lower id. 
But, when I tried to find a fix for it, I ran into a few problems and I couldn't find an easy solution. Does someone have an idea? :)

MizukiTemma:
> I only ran into one problem. The ordering by location is already working, but the locations are not sorted alphabetically but instead by poi_id. In my case "Test" was sorted before "Entwurf", as it had the lower id. But, when I tried to find a fix for it, I ran into a few problems and I couldn't find an easy solution. Does someone have an idea? :)


Ohw 😓 I didn't read the issue conversation carefully enough, thought that it's enough to group by POI and after that sort by name whithin each POI.

As POI objects do not have any name (names are provided by translations) I don't think sorting POI by name is handy enough. It rather makes sense to add filter by POI

MizukiTemma:
@osmers 
Should POIs be sorted by their name at any cost? I know it's better but saldy not technically cheap.

charludo (approved):
Thank you! :tada:

Re: the sorting iussue: maybe we could use this as an opportunity to add client-side sorting to all lists of this kind? I.e. allow users to cklick on any table header field to sort/reverse-sort by that key? AFAIK that should be relatively straight-forward (and a very common / expected feature anyways)

JoeyStk:
> Re: the sorting iussue: maybe we could use this as an opportunity to add client-side sorting to all lists of this kind? I.e. allow users to cklick on any table header field to sort/reverse-sort by that key? AFAIK that should be relatively straight-forward (and a very common / expected feature anyways)

I think this would be the best solution, but let's see what @osmers thinks :) While we were planning the region condition @PeterNerlich and I considered something similiar, so it might me good to implement it in all desired places at the same time :)

osmers:
Sounds good to me :)

MizukiTemma:
That's a nice idea 😸 I opened an issue for that #3347 

I'd suggest to merge this PR as is (providing only grouping by POI and alphabetical sort by contact name) and apply client-side sorting all over our system at once in another PR.

JoeyStk (approved):

JoeyStk:
Could you please wait to merge this until the release is through? :) Otherwise we would need another round of testing :)

PR #3344, opened 2025-01-18
Add basic table to region condition

JoeyStk:
### Short description

This PR adds the first batch of columns to the region condition - namely number of pages, number of outdated pages, number of languages and whether the mt add on was booked.

### Proposed changes

- Add table to region condition
- Add columns: number of pages, number of outdated pages, number of languages and booked_mt_addon

### Side effects

- I tried to take the calculations from `region_condition_actions`, not from dashboard
- As we expect a high loading time, please test this with the database dump, if you can :)

### Resolved issues

Fixes: Parts of #3012

### DIFF
--- a/integreat_cms/cms/templates/region_condition/region_condition_row.html
+++ b/integreat_cms/cms/templates/region_condition/region_condition_row.html
@@ -1,0 +1,24 @@
+{% load i18n %}
+{% load rules %}
+{% load content_filters %}
+<tr class="border-t border-solid border-gray-200 hover:bg-gray-100 text-gray-800">
+    <td class="py-3 pl-4">
+        {{ region.name }}
+    </td>
+    <td class="text-center py-3 pr-2">
+        {{ region.get_pages.count }}
+    </td>
+    <td class="text-center py-3 pr-2">
+        {{ region.outdated_pages.count }}
+    </td>
+    <td class="text-center py-3 pr-2">
+        {{ region.active_languages_without_default_language|length }}
+    </td>
+    <td class="text-center py-3 pr-2">
+        {% if region.mt_addon_booked %}
+            {% translate "Yes" %}
+        {% else %}
+            {% translate "No" %}
+        {% endif %}
+    </td>
+</tr>
--- a/integreat_cms/cms/views/region_condition/region_condition_view.py
+++ b/integreat_cms/cms/views/region_condition/region_condition_view.py
@@ -1,11 +1,24 @@
 from __future__ import annotations
 
 import logging
+from typing import TYPE_CHECKING
 
+from django.conf import settings
+from django.core.paginator import Paginator
+from django.db.models import Q
+from django.shortcuts import render
 from django.utils.decorators import method_decorator
 from django.views.generic import TemplateView
 
+from ...constants import region_status
 from ...decorators import permission_required
+from ...forms import ObjectSearchForm
+from ...models import Region
+
+if TYPE_CHECKING:
+    from typing import Any
+
+    from django.http import HttpRequest, HttpResponse
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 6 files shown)


integreat_cms/cms/views/region_condition/region_condition_view.py
```diff
@@ -20,3 +31,38 @@ class RegionConditionView(TemplateView):
...
+        :param request: The current request
+        :param \*args: The supplied arguments
+        :param \**kwargs: The supplied keyword arguments
+        :return: The rendered template response
+        """
+
+        regions = Region.objects.all()
+        query = None
+
+        search_data = kwargs.get("search_data")
```
charludo, line 48: Where do I enter this? 😭

charludo (commented):
Thank you, looks great! No more construction! 😆 I just have one question...

integreat_cms/cms/views/region_condition/region_condition_view.py
JoeyStk: Not yet at all. I was think that it might make sense in the future, but maybe it makes more sense to remove this again until it's actually implemented :sweat_smile:

charludo: Ah, OK! No, leave it in!

charludo (approved):

integreat_cms/cms/templates/region_condition/region_condition_row.html
```diff
@@ -0,0 +1,24 @@
...
+        {{ region.name }}
+    </td>
+    <td class="text-center py-3 pr-2">
+        {{ region.get_pages.count }}
+    </td>
+    <td class="text-center py-3 pr-2">
+        {{ region.outdated_pages.count }}
+    </td>
+    <td class="text-center py-3 pr-2">
+        {{ region.language_tree|length }}
```
MizukiTemma, line 15: We need `-1` here to exclude root language (assumed that the table should show the same result as in the exported  CSV)

MizukiTemma (changes requested):
Thank for the PR 👍 I tests first with the fixture data and found one thing to fix .

The PR will be tested with the dump data after the comment is solved.

integreat_cms/cms/templates/region_condition/region_condition_row.html
MizukiTemma: `{{ region.language_tree|length|add:"-1" }}` works, for exmaple. But I found some problems as I tried it:

1. `region.language_tree - 1` leads to `-1` if a region has no language (yet). This means `dehydrate_num_languages_besides_root_language` needs to be adjusted too. We didn't notice this while reviewing the CSV export function, becuase the test data region "Region without language" was excluded from the export as it is archived.

2. As this calculation is being used already at two places, we should probably build a new function like `num_language_beside_root_language` or something. It provents also we forget to do this "-1" again somewhere.

3. Archived regions must be excluded from the table too, as in the export function.

JoeyStk: > Archived regions must be excluded from the table too, as in the export function.

@osmers Currently we have diverging implementations for the table and the CSV file for the region conditions. In the table archived regions are included. In the CSV they are not. Which one is better? Should we include archived regions in the CSV or should we exclude them from the table? :)

JoeyStk: @osmers Do we want all languages for every region or do we just want the active languages? :thinking:

osmers: > Currently we have diverging implementations for the table and the CSV file for the region conditions. In the table archived regions are included. In the CSV they are not. Which one is better? Should we include archived regions in the CSV or should we exclude them from the table? :)

If I have a filter, then archived regions could be included but I see no need for that. I was also thinking about hidden regions bcs for some, we want the numbers, while for others it is of no interest to us :D e.g. Testumgebung, Finland template, etc...
What's the plan here? Will we still be able to export the data to csv? Sometimes this might be useful to calculate total number of broken links, pages with HIX below 14 etc...and I think that won't be possible in the table in the cms bcs of those hidden regions we don't need...

JoeyStk: > What's the plan here? Will we still be able to export the data to csv?

Yes, the CSV will exist as is, the table is just extra and in it's first iteration very plain :)

MizukiTemma (approved):
Thank you 👍 Looks good 😸 

Just squash the commits before merge 😉

Issue #3070, opened 2024-09-23
Change event creation settings to offer single events and Dauerveranstaltungen


osmers:
### Motivation
Municipalities are complaining about the restriction of 28 days on events bcs some events (exhibitions, Interkulturelle Woche) are longer than that. We created the restriction to make the event calendar clearer (long events are always shown first and put one day events to the back of the list).

### Proposed Solution
Divide the date and time section of event creation into two
One for one-time events and one for Dauerveranstaltungen
Standard selection will be the one-time-event.
It'll have only one date field (no start and end date, bcs they are on the same day)
Reccuring events will stay in this tab.
Events longer than 28 days imported with an external calendar are imported as long term event

Dauerveranstaltungen will be in the second tab. They will have an explanation above what can be considered Dauerveranstaltung and an explanation why it sometimes makes more sense to create recurring events instead).

Make sure to be backwards compatible for the apps (and perhaps consult the app team on the new api design).

### Alternatives
Keep restrictions or remove them completely. 
Have some kind of Freigabeprozess for long events.

### User Story
As a CMS user I want to have the autonomy to create the events that I need in order to reflect what we offer in our municipality. 

### Additional Context
App issue: https://github.com/digitalfabrik/integreat-app/issues/2978
Please notify the app team once this is done.

![image](https://github.com/user-attachments/assets/8140437d-be47-46e6-8d55-ddd57f6981b7)

### Design Requirements
@hauf-toni can you design this according to what we discussed during the conference?
The design can be found here: https://www.figma.com/design/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=2204-1049&t=tTzbVsG1piOxI0cl-1

nikolahoff:
@osmers question regarding: "Dauerveranstaltungen will be in the second tab." Do you mean that the tile will have 2 tabs?  Something like this?
<img width="524" alt="Bildschirmfoto 2024-10-07 um 12 00 20" src="https://github.com/user-attachments/assets/94afbca9-7e2b-4928-a007-dd62f8b1e5f3">

osmers:
[...]

Issue #3347, opened 2025-01-20
Add client-side sorting


MizukiTemma:
### Motivation

While reviewing #3334 it was suggested to add client-side sorting to all our list, so users can click to the header of columns to sort the objects of the list by the key in ascending/descending order.

### Proposed Solution

TBD

### Additional Context

See [this conversation](https://github.com/digitalfabrik/integreat-cms/pull/3334#pullrequestreview-2558508221)

JoeyStk:
During the grooming we had following ideas:

Option 1: client side sorting -> entire system via JS/TS
Option 2: server side sorting -> extra parameter and refresh page

laravoelk:
We also discussed order possibilities in the service team. From our side, it would be great if POIs, events and contacts could be sorted. :-) 
We thought about a sorting structure like the folders on computers. So for example the date, when it was last changed, name etc. Do you think, this would be possible?

jonbulz:
In our discussion on 21.10.2025, we decided to use server-side sorting for the following reasons:
- we already use Django's pagination class
- implementing pagination and client-side sorting with vanilla JS/TS is a lot of overhead
- transferring all data to the client (to allow client-side sorting) could increase loading times

PR #3355, opened 2025-01-21
Offer long-term event

MizukiTemma:
### Short description

This PR modifies our event feature so that users can create so-called long-term events in future.

### Proposed changes

- Introduce two tabs in the box "Date and Time" in the event form
- Set `end` of model `Event` to `blank=True`, add new field `only_weekdays`
- Migration
- Add `is_long_term` to event form (default to `False`, initialized as `True` if start date != end date when opening an existing event, toggled whenever users switche the tab)
- Remove the restriction of `MAX_EVENT_DURATION`
- Set end date <- start date if `is_long_term = False`
- New type "Long-term" in the column "Recurrence" in the event list (currently only "One time", "Weekly", "Monthly"...)
- Hide "DAILY" from recurrence frequency option
- Remove `has_recurrence_end_date` as the field of recurrence end date is shown as default
- Reject events with daily recurrence during import from external calendars  

Following changes (specified in the design or needed as result of the new event structure) are not covered currently. I would suggest to solve them in separate issues/PRs for better overview and less complexity per PR.
- Modification of display of recurrence rule (from pull-down to check boxes, etc): see [Handle weekdays_for_weekly data separately from the other data because has_changed doesn't work with CheckboxSelectMultiple widgets and ArrayFields out of the box](https://github.com/digitalfabrik/integreat-cms/blob/102b45fe268754e5dab9d6156c063c74a1484cf4/integreat_cms/cms/forms/events/recurrence_rule_form.py#L150-L167) -> #3385 
- Adjustment of filter to consider the change with end date and new event type "long-term" -> #3379
 

### Side effects

The largest changes for technical aspect are:
1. The input field for end date is not shown (and no value provided from the user) in the form if the user is eiding on the tab "One-time and recurrent". 
2. Recurrence frequency "DAILY" should not be available for users.

and they result in the new definition of one-time, recurring and long-term.

Migration of existing events, modification of validation and calendar import function is done based on this table
image

### Resolved issues

Fixes: #3070 


integreat_cms/release_notes/current/unreleased/3070.yml
```diff
@@ -0,0 +1,2 @@
+en: Add a new event type "long-term" for events longer than 28 days
+de: Fügen einen neuen Veranstaltungstyp „Fortlaufend“ für Veranstaltungen hinzu, die länger als 28 Tage dauern
```
JoeyStk, line 2:
```suggestion
de: Füge einen neuen Veranstaltungstyp „Fortlaufend“ für Veranstaltungen hinzu, die länger als 28 Tage dauern
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -1955,23 +1955,14 @@ msgstr ""
...
-#: cms/forms/events/recurrence_rule_form.py
-msgid "Recurrence ends"
-msgstr "Wiederholung endet"
-
 #: cms/forms/events/recurrence_rule_form.py
 msgid "No recurrence frequency selected"
 msgstr "Keine Häufigkeit der Wiederholung angegeben"
 
+#: cms/forms/events/recurrence_rule_form.py
+msgid "Recurrence DAILY is not allowed anymore"
```
JoeyStk, line 1963:
```suggestion
msgid "Recurrence daily is not allowed anymore"
```
Or maybe this is better in quotes? Like Recurrence "daily" is not allowed anymore

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -6093,6 +6095,35 @@ msgstr "Diese Veranstaltung löschen"
...
+msgid "These events take place daily."
+msgstr "Diese Veranstaltungen finden täglich statt."
+
+#: cms/templates/events/event_form_sidebar/date_and_time_box.html
+msgid "Please create daily repeating events under “Long-term”."
+msgstr "Täglich wiederholende Events bitte unter “Fortlaufend” anlegen."
+
+#: cms/templates/events/event_form_sidebar/date_and_time_box.html
+msgid " (optional)"
+msgstr " (optional)"
```
JoeyStk, line 6125: I think if possible we should try to avoid to translate empty spaces. Do you think it would be possible to exclude the empty space from the translation? :)

JoeyStk (commented):
Thank you very much for this PR! To me it already looks good! I only have a few notes. The topic I didn't really review is the migration, as I don't feel safe in this area. Maybe the second reviewer can have a more detailed look on this :)

### Add test data
I think adding some more test data for events would be great. I think in the end we want to have three events: one one-time, one recurring and one long-term event? :)

### Filtering
I think this is a question for @osmers: Do we want some kind of filter in the event list filter for one time and recurring events / long term events? :)

### View for larger screens
> Introduce two tabs in the box "Date and Time" in the event form

For larger (and smaller) screens, when the box moves to the right column the tabs overlap each other. I don't think this is ideal. Maybe we can come up with something for it to look better? :) 
grafik

osmers:
Regarding the filtering: I don't think that's immediately necessary, we already have recurring events and no filter for that. We could think about implementing it in the future just to make it easier to find things. 
And since for contacts we are implementing the sorting per column function, that might also be interesting for events, right? Then they would have all recurring events below each other when sorting that column.

osmers:
@JoeyStk can you open issues for implementing the filtering and/or sorting by column?

MizukiTemma:
@osmers @JoeyStk 
#3347 is there for that topi.

JoeyStk:
> In the current filter implementation, long-term events belong to "non-recurreing events", that means long-term events and one-time evens will be shown if "non-recurring events" is checked. But in the form, recurreing events and one-time events belong to the same tab. This can be confusing.

I agree :) I think a third filter would be good :)

integreat_cms/locale/de/LC_MESSAGES/django.po
MizukiTemma: Nice idea :) applied!

JoeyStk:
Looks good so far!

> View for larger screens
The tabs stay side by side now, but I can't make them have the same height 🙈 Could you help me? 

The problem, as I understand it, is that the text in the left tab is too long. Therefore my idea would have been to make the ul scrollable on the x axis. The right tab would be a little into the invisible zone, but still visible enough to see it and to be able to click on it. However, I wasn't able to make it work :confused: Maybe the second reviewer has an idea that works? :thinking:

jonbulz:
> View for larger screens
The tabs stay side by side now, but I can't make them have the same height 🙈 Could you help me?

Looks good on my screen. Can you show what the problem with the tab height is?

MizukiTemma:
@jonbulz 

The problem appears when the box "date and time" comes to the right side of mailn input field (by changing the screen size)
image

jonbulz:
Ok, thanks! I couldn't see it because in english, the tabs have the same height. Do we use css for this kind of styling? I couldn't find it in the code.

MizukiTemma:
> Ok, thanks! I couldn't see it because in english, the tabs have the same height. Do we use css for this kind of styling? I couldn't find it in the code.

I don't think we have css for tabs 🤔

jonbulz:
> I don't think we have a css for tabs 🤔

Do you think it's worth to add css because of this problem? Because I don't know how else to do it. But maybe someone else has a better solution :shrug:

MizukiTemma:
> Do you think it's worth to add css because of this problem? Because I don't know how else to do it. But maybe someone else has a better solution 🤷

Yes. 
Could you create css for that? I tried but never succeeded 😓

JoeyStk:
> It seems I managed it. Could you check once again?

Thank you. Yes, it works for me :9

JoeyStk (approved):

jonbulz (approved):

MizukiTemma:
@osmers 
This PR is now approved. Is it okay to merge it without modification of recurrence rule appearance (#3385)?

osmers:
@MizukiTemma please merge, I'd like to take a look at what it will look like - should be ok though without merging the recurrence rule at the same time. Otherwise we can still exclude it from the release in case we deem the recurrence rule appearance necessary.

JoeyStk:
> Otherwise we can still exclude it from the release in case we deem the recurrence rule appearance necessary.

AFAIK it's currently not yet possible to do so. At the moment it's still the case that everything that has been merged is getting released :confused:

osmers:
> > Otherwise we can still exclude it from the release in case we deem the recurrence rule appearance necessary.
> 
> AFAIK it's currently not yet possible to do so. At the moment it's still the case that everything that has been merged is getting released 😕

Charlotte told me yesterday that it's not that difficult to reverse...but maybe that was issue specific?

JoeyStk:
Alright, I think we can merge this then :)

charludo:
> > > Otherwise we can still exclude it from the release in case we deem the recurrence rule appearance necessary.
> > 
> > 
> > AFAIK it's currently not yet possible to do so. At the moment it's still the case that everything that has been merged is getting released 😕
> 
> Charlotte told me yesterday that it's not that difficult to reverse...but maybe that was issue specific?

Yes, with the HIX change it would just be reverting 2 lines of code, which also aren't being touched by any other current issues. This feature is a bit more complicated and a different story :D

PR #3367, opened 2025-01-27
only show MT-button if user has publish rights for a page

jonbulz:
### Short description
Show MT-button for pages only to users with publishing rights

### Proposed changes
- add additional check user_has_publish_rights in PageTranslationForm and return an empty queryset if they don't

### Side effects
- requires changes in some tests, because the AUTHOR role can now only use machine translations for specific pages

### Resolved issues

Fixes: #2961

### DIFF
--- a/integreat_cms/cms/forms/pages/page_translation_form.py
+++ b/integreat_cms/cms/forms/pages/page_translation_form.py
@@ -1,8 +1,12 @@
 from __future__ import annotations
 
 import logging
+from typing import TYPE_CHECKING
 
-from ...models import PageTranslation
+if TYPE_CHECKING:
+    from treebeard.ns_tree import NS_NodeQuerySet
+
+from ...models import LanguageTreeNode, PageTranslation
 from ..machine_translation_form import MachineTranslationForm
 
 logger = logging.getLogger(__name__)
@@ -12,6 +16,22 @@ class PageTranslationForm(MachineTranslationForm):
     """
     Form for creating and modifying page translation objects
     """
+
+    def user_has_publish_rights(self) -> bool:
+        """
+        Helper method to check if the current user has permission to publish the page
+        """
+        if page_obj := self.instance.page:
+            return self.request.user.has_perm("cms.publish_page_object", page_obj)
+        return self.request.user.has_perm("cms.publish_page")
+
+    def mt_form_is_enabled(self) -> NS_NodeQuerySet:
+        """
+        For pages, machine translations should only be enabled if the user has publishing rights
+        """
+        if not self.user_has_publish_rights():
+            return LanguageTreeNode.objects.none()
+        return super().mt_form_is_enabled()
 
     class Meta:
         """
--- a/integreat_cms/cms/views/pages/page_context_mixin.py
+++ b/integreat_cms/cms/views/pages/page_context_mixin.py
@@ -31,6 +31,8 @@ class PageContextMixin(ContextMixin):
         context = super().get_context_data(**kwargs)
         context.update(
             {
+                "MT_PERMITTED": context.get("MT_PERMITTED", False)
+                and self.request.user.has_perm("cms.publish_page_object"),
                 "current_menu_item": "pages",
                 "translation_status": translation_status,
                 "archive_dialog_title": _(
--- a/integreat_cms/release_notes/current/unreleased/2961.yml
+++ b/integreat_cms/release_notes/current/unreleased/2961.yml
@@ -1,0 +1,2 @@
+en: Only users with publish rights for pages can use machine translations.
+de: Nur Benutzer mit Veröffentlichungsrechten für Seiten können maschinelle Übersetzungen verwenden.
--- a/tests/mt_api/test_mt_button_visibility.py
+++ b/tests/mt_api/test_mt_button_visibility.py
@@ -1,0 +1,47 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from django.test.client import Client
+
+import re
+
+import pytest
+from django.apps import apps
+from django.urls import reverse
+
+from ..conftest import EDITOR, MANAGEMENT, PRIV_STAFF_ROLES
+
+
+@pytest.mark.django_db
+def test_mt_button_visibility(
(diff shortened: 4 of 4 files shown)


jonbulz:
Hi @MizukiTemma , I'm a bit lost as to why the tests are failing. I suspect it's because the message I am checking in the logs is the wrong message, but I'm not sure what to expect instead. Can you help me with that?

MizukiTemma:
@jonbulz 
It's probably because of no permission check on python side. The buttons for translation is hidden, but the related functions are not checking the permission before translating by MT provider. In GUI author users do not see the MT option and therefore cannot use it. But in the test we call the functions directly, which means author users are not blocked by "hidden" buttons.

jonbulz:
@MizukiTemma thanks, I see now. Since I don't change any actual permissions, the behavior in tests basically stays the same. I'll have to think about this, I don't think it's very clean. At the very least, I would like to have a test that investigates the expected user experience for authors

jonbulz:
@MizukiTemma I tried implementing a test to check if the button is visible for observers without publish rights. However, it seems that my approach is not working. Are you doing something similar somewhere else in the code?

jonbulz:
@MizukiTemma thanks! This is similar to what I have tried, though. I'm afraid that the part of the page I want to check is rendered later, because it never showed up in the response content. Maybe I should bring it up in the next weekly?

JoeyStk:
Thank you a lot for tackling this issue :rocket: For the page view it works as expected :)
I have one question: Do we also want to hide the translation bulk action in page tree? (maybe this is a question for @osmers?)

MizukiTemma:
> @MizukiTemma thanks! This is similar to what I have tried, though. I'm afraid that the part of the page I want to check is rendered later, because it never showed up in the response content. Maybe I should bring it up in the next weekly?

Sure. Alternatively you can pop this discussion in one of our channels. It helps us to share the context before the weekly and someone may answer quicker.

jonbulz:
@JoeyStk since I finally managed to add a test, this PR is ready for review

JoeyStk (approved):
Works as expected & test looks good! Thank you very much! :blush:

MizukiTemma (changes requested):
Looks good with the translation option in the form.

Could you adjust the bulk action too? Find a suggestion [here](https://github.com/digitalfabrik/integreat-cms/issues/2961#issuecomment-2272715891)

jonbulz:
@MizukiTemma thanks for pointing that out! I'm hiding the bulk action for users without publishing rights now, but this is starting to feel a little hacky. My feeling is that no one will remember this in the future, and people might start to wonder why MT permissions are different for pages as compared to everything else (which they are not, we are just hiding things away from the user). In #2997, you had to jump through some hoops because you wanted to check on a per page basis if the user has the appropriate permissions, right? Is it an option to just take MT away from users without general publishing rights, even if they are allowed to publish specific pages?

MizukiTemma:
> Is it an option to just take MT away from users without general publishing rights, even if they are allowed to publish specific pages?

Bulk MT can be completely deactivated/hidden for users without general publish permission, even if they have publish permission on some pages, as mentioned [here](https://github.com/digitalfabrik/integreat-cms/issues/2961#issuecomment-2272715891).

If you mean taking also MT on the form away, we have to ask Service Team.

jonbulz:
> > Is it an option to just take MT away from users without general publishing rights, even if they are allowed to publish specific pages?
> 
> Bulk MT can be completely deactivated/hidden for users without general publish permission, even if they have publish permission on some pages, as mentioned [here](https://github.com/digitalfabrik/integreat-cms/issues/2961#issuecomment-2272715891).
> 
> If you mean taking also MT on the form away, we have to ask Service Team.

The comment you have linked describes it well. With these changes, users without general publish permissions won't be able to bulk MT. Could you review again? :)

jonbulz:
Good catch! Though I think that is a bug that is unrelated to my changes. I will look further into it

jonbulz:
@MizukiTemma could you check again? I want to highlight the changes to the __init__ method of MachineTranslationForm. IMO, the permission check there was just to save resources in case users don't have permission for MT. But it was preventing users without general page permissions from using MT because the translation targets would never get computed.

MizukiTemma (approved):
Thank you 🚀

MizukiTemma (changes requested):
Sorry, found a thing to fix: go to the page "Kontakt zu App Team Augsburg", check the check box of MT in the form and click "Aktualisieren". we get an error `Machine translations are disabled for content type`

Sorry for reverting the approval 🙈

jonbulz:
@MizukiTemma Alright, gotchu. This kind of brings us back to the start unfortunately, because the api client won't initialize if the user does not have general change_page permissions. I'm starting to think that your approach in #2997 is the better one after all. @charludo what do you think? I'm trying to summarize so you don't have to do too much research:

**What am I trying to do?**
We don't want users that can't publish pages to use machine translations, because it could use up budget unnecessarily. 

**What is the problem?**
The permission check for machine translations currently only looks for general 'cms.change_{foreign_field}" permissions and is not designed for fine-grained permission checks, i.e. individual pages.

**What are possible solutions?**
- In this PR, I try to avoid the problem by not actually changing the required permissions, but instead hiding the buttons from the users. This would require additional changes to the init methods of machine translation api clients to avoid the permission check in special cases.
- In #2997 , @MizukiTemma changed the permission check for machine translations in a way that it became possible to check permissions for individual pages. However, that required to pass a 'content' variable to the machine translation provider.

IMO, it is better to pass the content down to the permission check, even though it introduces an additional variable, for 3 reasons:
1) Just hiding things without changing actual permissions is not a clean solution to start with. Whenever we change the forms in the future, we would have to keep this requirement in mind.
2) If requirements change in the future, devs might be confused why buttons don't show up in the page views if they are not familiar with these changes
3) We might have similar requirements for objects other than pages in the future and could re-use the mechanism

Sorry for the text wall :see_no_evil:

MizukiTemma:
I'm for the option to allow MT only for users with general publish permission. I'll ask Service Team in the issue.

MizukiTemma (approved):
Thank you 🚀 

Just one thing: add a release note, please 😉 Then it's ready to go!

PR #3375, opened 2025-01-29
Fix contact link tracking

charludo:
### Short description

This PR fixes bugs with contact link tracking which were introduced in #3320.

### Proposed changes

- update test data fixtures so the "Testkontakt" URL is correct (i.e. contains details information)
- adjusts the DB query for gathering referring objects to check if Link objects *start* with the same URL as the contact. This is what has been preventing contacts from being tracked correctly for the last week. 
- Similarly, adjust internal link checker to be able to handle contact URLs with details information

### Side effects

- none, but please note that this does not really solve #3280, it just fixes the regressions which #3320 caused. Once this PR is merged, contact tracking should start working locally again, and we can go back to observing if it also works on the test server 🙈 

### Resolved issues

Fixes: n/a

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -111,7 +111,7 @@
             set_if_unset INTEGREAT_CMS_SECRET_KEY "dummy"
             set_if_unset INTEGREAT_CMS_BACKGROUND_TASKS_ENABLED 0
             set_if_unset INTEGREAT_CMS_SUMM_AI_API_KEY "dummy"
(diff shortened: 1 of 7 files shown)


MizukiTemma (approved):
👍

JoeyStk (approved):
Thank you very much. It had to recheck the links initially - but I assume that's expected, right? :)

charludo:
> Thank you very much. This fixes the issue. It had to recheck the links initially - but I assume that's expected, right? :)

Yes, although pruning the database and then starting the CMS should pick the contact links up immediately!

Issue #3371, opened 2025-01-28
Archive/Delete button not greyed out when contact is linked to page/poi/event


juliankehne:
### Describe the Bug

The archive & delete buttons are not greyed out when a contact is inserted in a page/poi/event and linked
![Image](https://github.com/user-attachments/assets/dfe1731c-0a6b-4eff-a152-74a3d197487c)

The delete button in the edit contact form is not greyed out either (not clickable though):
![Image](https://github.com/user-attachments/assets/d1c331f1-b313-49b7-a006-ac3bc264a799)

### Steps to Reproduce

1. Go to Contacts
2. Find a contact that is linked in page/poi/event
3. See that buttons are not greyed out

### Expected Behavior

Greyed out buttons (in list and edit form), when clicking it nothing should happen

### Actual Behavior

Buttons are not greyed out, when clicking the list view buttons the edit contact page opens with the message that archiving the contact is not possible as it is inserted in a page/poi/event.

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

### Related Issues

Prevent orphaned contacts issue: https://github.com/digitalfabrik/integreat-cms/issues/3260
Prevent orphaned contacts PR: https://github.com/digitalfabrik/integreat-cms/pull/3282

PR #3377, opened 2025-01-29
Indicate when contact actions are not available

charludo:
### Short description

Indicate vidually when contact actions are not available

### Proposed changes

- disable the small buttons in the list view (sorry, I missed that previously)
- change order of styles in `style.scss` so that disabled button styling overwrites "red" button styling, instead of the other way round 

### Side effects

- the style changes affect all buttons, not just the contact ones. However, I think that is probably a wanted change?  

### Resolved issues

Fixes: #3371

### DIFF
--- a/integreat_cms/cms/templates/contacts/contact_form.html
+++ b/integreat_cms/cms/templates/contacts/contact_form.html
@@ -95,15 +95,18 @@
             {% if contact_form.instance.id %}
                 <div class="flex flex-wrap grow justify-end gap-2 items-center">
                     {% if perms.cms.change_contact and not contact_form.instance.archived %}
-                        <button title="{% translate "Archive contact" %}"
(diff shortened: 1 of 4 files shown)


MizukiTemma (approved):
Thank you for speedy fix 😸

charludo:
> Thank you for speedy fix 😸

Yeah sorry I totally missed those in the recent PR 🙈

JoeyStk (commented):
Thank you very much! 
This fixes the issue. I think it would be nice to give some feedback to the user why it's not possible to delete the contact. Do you think it makes sense to change the title of the button giving an explanation why the button is greyed out / disabled? :)

charludo:
> Thank you very much! This fixes the issue. I think it would be nice to give some feedback to the user why it's not possible to delete the contact. Do you think it makes sense to change the title of the button giving an explanation why the button is greyed out / disabled? :)

Good idea. I have changed the button texts, but did not want to make them too long, so most of the context is in the tooptip. Hope that's OK!

JoeyStk (approved):
Thank you very much, looks good :)

PR #3390, opened 2025-02-04
Add management command to update link text

MizukiTemma:
### Short description

This PR introduces a management command which updates the link texts of ALL link object with the given URL. 

The motivation is to fulfill [this need](https://chat.tuerantuer.org/digitalfabrik/pl/5otdpsbjmiy53fd1cwzqr6mqfr). This PR can be a step of #2226 

### Proposed changes

- Find contents which contain the specified URL, check all links in each content and replace the link text if the link matches the given URL
- Add a documentation

### Side effects

- **note**: There is no way to specify which links should get the new link text (by regin, content). Change will be applied to ALL links with the URL. ALL link objects with the URL get the SAME link text.
- 

### Resolved issues

Fixes: 

### DIFF
--- a/docs/src/management-commands.rst
+++ b/docs/src/management-commands.rst
@@ -198,6 +198,23 @@ Reset MT budget of regions whose renewal month is the current month::
 * ``--force``: Allow to reset the budget even if it's not the first day of the month
 
 
+``update_link_text``
+~~~~~~~~~~~~~~~~~~~~~~
+
+Update ALL link text of links of the given URL:
+
+    integreat-cms-cli update_link_text --target-url TARGET_URL --new-link-text NEW_LINK_TEXT [--username USERNAME]
+
+**Arguments:**
+
+* ``TARGET_URL``: Update the link text of ALL links of the url ``TARGET_URL``
+* ``NEW_LINK_TEXT``: Update the link texts to ``NEW_LINK_TEXT``
+
+**Options:**
+
+* ``USERNAME``: Associate any new created translations with ``USERNAME``
+
+
 Create new commands
 -------------------
 
--- a/integreat_cms/core/management/commands/update_link_text.py
+++ b/integreat_cms/core/management/commands/update_link_text.py
@@ -1,0 +1,136 @@
+from __future__ import annotations
+
+import logging
+import time
+from html import unescape
+from typing import TYPE_CHECKING
+
+from django.contrib.auth import get_user_model
+from django.core.management.base import CommandError
+from linkcheck.listeners import tasks_queue
+from linkcheck.models import Url
+from lxml.etree import ParserError
+from lxml.html import fromstring, tostring
+
+if TYPE_CHECKING:
+    from typing import Any
+
+    from django.core.management.base import CommandParser
+
+    from .users.user import User
+
+
+from ..log_command import LogCommand
+
+logger = logging.getLogger(__name__)
+
+
+def get_url(url: str) -> Url:
+    """
+    Get an URL object by url or raise an error if not found
+
+    :param url: url as string
+    :return: URL
+    """
+    try:
+        return Url.objects.get(url=url)
+    except Url.DoesNotExist as e:
+        raise CommandError(f'URL object with url "{url}" does not exist.') from e
+
+
+def get_user(username: str) -> User:
+    """
+    Get a user by username or raise an error if not found
+
+    :param username: Username
+    :return: User
+    """
+    try:
+        return get_user_model().objects.get(username=username)
+    except get_user_model().DoesNotExist as e:
+        raise CommandError(f'User with username "{username}" does not exist.') from e
+
+
+class Command(LogCommand):
+    """
+    Management command to update the link text of ALL links with the given URL.
+    ALL links receive the SAME new link text.
+    This applies in ALL regions.
+    """
+
+    help = "Update link text of the URL"
+
+    def add_arguments(self, parser: CommandParser) -> None:
+        """
+        Define the arguments of this command
+
+        :param parser: The argument parser
+        """
+        parser.add_argument(
+            "--target-url",
+            required=True,
+            help="The URL whose link text needs to be changed",
+        )
+        parser.add_argument("--new-link-text", required=True, help="The new link text")
+        parser.add_argument("--username", help="The username of the creator")
+
+    def handle(
+        self,
+        *args: Any,
+        target_url: str,
+        new_link_text: str,
+        username: str,
+        **options: Any,
+    ) -> None:
+        r"""
+        Try to run the command
+
+        :param \*args: The supplied arguments
+        :param target_url: URL whose link texts need to be updated
+        :param new_link_text: New link text to replace with the current link texts
+        :param username: The username of the creator
+        :param \**options: The supplied keyword options
+        :raises ~django.core.management.base.CommandError: When the input is invalid
+        """
+        self.set_logging_stream()
+
+        target_url_object = get_url(target_url)
+        user = get_user(username) if username else None
+
+        target_links = target_url_object.links.all()
+
+        target_contents = {target_link.content_object for target_link in target_links}
+
+        for target_content in target_contents:
+            if target_content._meta.default_related_name in [
+                "page_translations",
+                "event_translations",
+                "poi_translations",
+                "imprint_translations",
+            ]:
+                try:
+                    content = fromstring(target_content.content)
+                except ParserError:
+                    logger.info(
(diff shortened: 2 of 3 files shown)


MizukiTemma:
Found some fix needs.

integreat_cms/core/management/commands/update_link_text.py
```diff
@@ -0,0 +1,141 @@
...
+    from .regions.region import Region
+    from .users.user import User
+
+
+from ..log_command import LogCommand
+
+logger = logging.getLogger(__name__)
+
+
+def get_url(url: str) -> Region:
```
JoeyStk, line 29:
```suggestion
def get_url(url: str) -> URL:
```
This doesn't return a Region object, right? :thinking: If it did, I think we should rename the function

integreat_cms/core/management/commands/update_link_text.py
```diff
@@ -0,0 +1,141 @@
...
+        Define the arguments of this command
+
+        :param parser: The argument parser
+        """
+        parser.add_argument(
+            "--target-url", help="The URL whose link text needs to be changed"
+        )
+        parser.add_argument(
+            "--new-link-text", help="The URL whose link text needs to be changed"
+        )
```
JoeyStk, line 75:
```suggestion
        parser.add_argument(
            "--new-link-text", help="The new link text"
        )
```

JoeyStk (changes requested):
Thank you very much and sorry for taking so long to review this. This looks mostly good to me :) Please find my suggestions below :)

integreat_cms/core/management/commands/update_link_text.py
MizukiTemma: You're right 😸 Thak you!

JoeyStk (approved):
Thank you for your changes! I think this looks good now :)
I don't think there is anything we can do about the failing tests at the moment :/

docs/src/management-commands.rst
```diff
@@ -198,6 +198,23 @@ Reset MT budget of regions whose renewal month is the current month::
 * ``--force``: Allow to reset the budget even if it's not the first day of the month
 
 
+``update_link_text``
+~~~~~~~~~~~~~~~~~~~~~~
+
+Update ALL link text of links of the given URL:
+
+    integreat-cms-cli update_link_text [--target_url TARGET_URL] [--new_link_text NEW_LINK_TEXT] [--username USERNAME]
```
PeterNerlich, line 206: It seems the arguments use hyphens, not underscores. Also, please make the important arguments required and mark them as such
```suggestion
    integreat-cms-cli update_link_text --target-url TARGET_URL --new-link-text NEW_LINK_TEXT [--username USERNAME]
```

integreat_cms/core/management/commands/update_link_text.py
```diff
@@ -0,0 +1,138 @@
...
+    def add_arguments(self, parser: CommandParser) -> None:
+        """
+        Define the arguments of this command
+
+        :param parser: The argument parser
+        """
+        parser.add_argument(
+            "--target-url", help="The URL whose link text needs to be changed"
+        )
+        parser.add_argument("--new-link-text", help="The new link text")
```
PeterNerlich, line 72: If we mark the arguments as [required](https://docs.python.org/3/library/argparse.html#required), they are handled and rendered automatically
```suggestion
        parser.add_argument(
            "--target-url", required=True, help="The URL whose link text needs to be changed"
        )
        parser.add_argument("--new-link-text", required=True, help="The new link text")
```

integreat_cms/core/management/commands/update_link_text.py
```diff
@@ -0,0 +1,138 @@
...
+        :param \**options: The supplied keyword options
+        :raises ~django.core.management.base.CommandError: When the input is invalid
+        """
+        self.set_logging_stream()
+
+        if not target_url:
+            raise CommandError("Please specify the target URL.")
+
+        if not new_link_text:
+            raise CommandError("Please specify the new link text.")
```
PeterNerlich, line 99: and we don't need to manually raise errors anymore
```suggestion
```

integreat_cms/core/management/commands/update_link_text.py
```diff
@@ -0,0 +1,138 @@
...
+
+                new_translation = target_content.create_new_version_copy(user)
+                new_translation.content = unescape(
+                    tostring(content, encoding="unicode", with_tail=False)
+                )
+
+                if new_translation.content != target_content.content:
+                    target_content.links.all().delete()
+                    new_translation.save()
+        time.sleep(0.1)
```
PeterNerlich, line 135: What purpose does this sleep fulfill? It might be nice to add a comment

integreat_cms/core/management/commands/update_link_text.py
```diff
@@ -0,0 +1,138 @@
...
+        tasks_queue.join()
```
PeterNerlich, line 136: Same with this join. I think I know what it does, but I don't think it is immediately obvious, especially if the reader only has superficial knowledge about `linkcheck`
```suggestion
        # Wait until all background tasks of linkcheck indexing the new content objects have been processed
        tasks_queue.join()
```

PeterNerlich (changes requested):
Thank you! I only found an issue with the command logistics themselves, not the actual functionality it implements, though I also didn't test that thoroughly yet.

I'm sorry, but I'm requesting you to change your tests regarding missing arguments with this. I think you need to assert `"integreat-cms-cli update_link_text: error: the following arguments are required: --target-url, --new-link-text" in out` now, I'm not completely sure

MizukiTemma:
@PeterNerlich 

Thank you for review and suggestions 😸 They're now applied

tests/core/management/commands/test_update_link_text.py
```diff
@@ -0,0 +1,119 @@
...
+        str(exc_info.value) == f'URL object with url "{NOT_USED_URL}" does not exist.'
+    )
+
+
+USED_URL = "https://integreat.app/augsburg/de/willkommen/"
+NEW_LINK_TEXT = "New Link Text"
+NEW_LINK = f'<a href="{USED_URL}">{NEW_LINK_TEXT}</a>'
+
+
+@pytest.mark.django_db(transaction=True, serialized_rollback=True)
```
PeterNerlich, line 93: <details><summary>Unfortunately, <code>django_db(transacton=True, serialized_rollback=True)</code> has nonsensical behaviour that screws our tests over big time.</summary>

Django has its own implementation and idea of test cases, but we are using pytest to run our tests. There are a few problems stemming from the different ways both approach this.

Django implements typical class-based unit tests for everything. Each test is a class, and each class spins up a db instance, migrates it to the latest schema, fills it with initial data *(this is what Django calls fixtures)*, does any other steps to set up the test, runs the test itself, and performs teardown/cleanup.

Pytest allows the user to use classes, but also just test functions by themselves. Any setting up and tearing down is done via a very powerful concept of what pytest calls fixtures – plain functions, or context managers if the teardown is needed, that can also supply a value to the test function. Most importantly, pytest fixtures are **reusable** and not *(necessarily)* re-evaluated for each test but for the specified **scope**, which can extend to the whole session.
There is a plugin creatively called `pytest_django` that defines a few fixtures to kind of serves as an adaptor between both worlds, but there still is one difference: It always tells django to keep the database between tests, which is normally just an extra option. This is obviously more efficient, but does mean that tests can influence the starting conditions of all tests run after it. Unfortunately as well, in django `serialized_rollback=True` works by flushing the database, since the expection is that the next test will just load the required django fixture itself. We wrote the `load_test_data` pytest fixture to basically emulate that for the normal tests, but it is session scoped and thus only executed once. So we also created the `load_test_data_transactional` fixture, which does essentially the same but is function scoped and thus run before each test. This ensures the test has the test data loaded. But at the end of the test the db is still flushed and any subsequent test just using the normal `load_test_data` will find an empty database.

In the end, we have an implicit dependency of all tests without `django_db(transaction=True, serialized_rollback=True)` to run before any tests with that marker.

</details>

We need to ensure that this test is not scheduled to run before any test not using `serialized_rollback`.
```suggestion
@pytest.mark.order("last")
@pytest.mark.django_db(transaction=True, serialized_rollback=True)
```

PeterNerlich (changes requested):
Only the comment about introducing `@pytest.mark.order("last")` to the transactional test is mandatory to resolve this change request, the rest are very light suggestions

PeterNerlich (approved):
Looks good and behaves as it should :grin: Glad that the tests also play nice for once

Issue #3158, opened 2024-10-28
Test whether DeepL/Google translate known examples in expected manner daily


PeterNerlich:
This should not be implemented as part of `integreat_cms`, but rather in our server infrastructure. However, since to develop the proposed tests we need the insight into typical translation content, I'll be borrowing this issue tracker.

### Motivation

We currently experience issues with both DeepL and Google serving corrupted translations *(such as putting phrases marked with `translate="no"` at the start of the sentence rather than where they make sense)* for various source/target language pairs, which is not something we can influence.

### Proposed Solution

**Add a daily cron job to**
- **translate known examples into various languages using DeepL and Google**
- **check if the results are conforming to our expectations or have deviations**
- **send a Mattermost notification if the state changes *(deviations appeared/disappeared compared to the previous day)***

In order to accomplish that, these **known examples** have to be **defined**. They should include all important features that we expect in content translations, in enough variations and redundancy that we can be reasonably confident in the result of the reports.
I imagine something like:
- *plain text:* short word or phase, as could appear in a page title
- *html:* single paragraph of a word or phrase, as might appear in page content
- *html:* single paragraph of a span with the [power set of all possible known values](https://en.wikipedia.org/wiki/Power_set#Example) to indicate something should not be translated *(`class="notranslate"`, `translate="no"`, `notranslate`)* surrounding a word or phrase
- *html:* single paragraph of some sentence containing a span with the power set of all possible notranslate indicators
- *…*

This assumes that we can decide reasonably well whether deviations occur or not. We likely will need some sort of **fuzzy matching**, as we might not be able to capture all possible different strings the API might return for any of the known examples that we would regard as valid.
**If such an automated decision algorithm cannot be found**, it might be *good enough* to post the whole example with the translated version in Mattermost, maybe along with the translation put through the system again and translated back to the source language, and have a **human check for deviations** every day. This can be sped up by saving results that have been reviewed in the past and not mentioning those previously marked to be a good translation, whenever they get produced by the API again.

### Alternatives

- Manual spot checking, which would be very labour intensive
- No checking at all, only learn about deviations when somebody notices them and reports it to us, which is the current situation

### User Story

As a *service provider* I want *to know about quirks and problems of my upstream translation services rather quickly* so that *I can give suggestions to my clients, or at least not have to find out about quirks at a press conference*.

### Additional Context

https://github.com/digitalfabrik/integreat-cms/pull/3135#issuecomment-2424968026
#3157

[...]

PR #3392, opened 2025-02-04
Pytest: Enable using markers

PeterNerlich:
### Short description

While working on #2297 I made a mistake that left using markers dysfunctional.

### Proposed changes

- Actually pass the specified marker to pytest, not the keyword arguments

### Side effects

- ~~If you relied on the functionality to completely ignore a marker when no keyword was specified or to ignore the given marker string and replace it with whatever you specified as keyword, then this will be a breaking change for you~~

### DIFF
--- a/tools/test.sh
+++ b/tools/test.sh
@@ -88,20 +88,21 @@ if [[ -z "${CHANGED}" ]] && (( ${#TESTS[@]} == 0 )); then
 fi
 
 if [[ -n "${KW_EXPR}" ]] || [[ -n "${MARKER}" ]] || (( ${#TESTS[@]} )); then
-    MESSAGES=()
+    TEST_MESSAGE=""
     if [[ -n "${KW_EXPR}" ]]; then
-        MESSAGES+=("\"${KW_EXPR}\"")
+        TEST_MESSAGE+=" \"${KW_EXPR}\""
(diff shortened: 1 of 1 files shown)


charludo (approved):
🎉

charludo:
Btw, I also meant to ask in #3395 - do you have a working linter setup? 🙈 It's really useful to be able to get all of these messages without manually having to invoke the corresponding tools:

image

PeterNerlich:
> do you have a working linter setup?

No, I always went bare editor and terminal until now. I dislike it if my machine spins away highlighting "mistakes" where I simply didn't finish writing the line yet, restarting the process every time I change anything. I prefer running those when I explicitly intend to. I will give it a try though now that we have ruff and see whether its performance can change my mind :grin:

jonbulz (approved):

jonbulz:
Thanks, I didn't really use markers so far. Looks to me that we use the pytest builtin markers, or did we define any custom ones?

PeterNerlich:
> Thanks, I didn't really use markers so far. Looks to me that we use the pytest builtin markers, or did we define any custom ones?

I am about to introduce custom markers and disable them by default for #3158, so they need to be enabled explicitly for the associated tests to run

PR #3325, opened 2025-01-11
Refactor page tree and add page based statistics

JoeyStk:
### Short description

This PR refactors parts of the page tree and adds the page based statistics to statistics.

### Proposed changes

- Refactor page tree
  - Split page tree into `partial_page_tree`, `pages_page_tree` and `statistics_page_tree`
  - Adjust naming
- Add page based statistics
  - Call to Matomo
  - Show page based statistics in statistics view  

### Side effects

- In collaboration with @jarlhengstmengel 
- Refactoring page tree is always dangerous, thorough testing is very, very welcome
- I would propose that this PR gets at least three reviews

### Resolved issues

Fixes: #3130

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #3394, opened 2025-02-04
Export page access statistics

### LINKED ISSUE #3134: Adjust export for page access statistics
### Motivation

We want to allow users of the CMS to export page access statistics too. 

### Proposed Solution

Therefore we want to extend the existing export function to page access statistics. The export should be a CSV and contain information about the timeframe of the exported data. Each line needs to contain the information for one page. This information contains the page name, accesses per language and the total accesses.

### Alternatives

None

### User Story

As a municipality I want to export page access statistics to CSV or PNG so that I can forward them in emails or show them without the need to be logged onto our system.

### Additional Context

This is part of our meta issue #1097 

### Design Requirements

Design proposal (v3):
https://www.figma.com/file/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=1179-830

jarlhengstmengel:
### Short description

This PR adds the download function for page access statistics

### Proposed changes

- Download functionality on the statistics page is extended to be able to select which statistics should be downloaded
- Page Statistics can be downloaded as csv
- For that `updatePageAccesses`in `statistics-page-accesses.ts` is extended to create an array with the page statistics data and a method `exportPageAccessesData` is introduced
- Checkboxes for selecting wich pages should be exported are now active
- Expand and Collapse all pages is added to the page tree

### Side effects

- The Javascript in the statistics is doing quite a lot of heavy lifting. I looked into using more of the bulk action functionality we have, but decided against it for now because I think that it would require some refactoring of the bulk action form/view. It's centered around the page tree view but we start using it for more and more. Additionally, how the statistics will be looking long term is currently bit in discussion.

### How to test

To test this locally you need statistics activated:
- Run the fetch page accesses management command: integreat-cms-cli fetch_page_accesses --start-date START_DATE --end--date END_DATE --region-slug augsburg (start and end date can be the same, good pick would be yesterday)
- Wait for about 30 to 40 sec
- Select some languguages in the sidebar and some pages
- By export select "Table Document/CSV - Page Accesses"

### Resolved issues

Fixes: #3134 

### DIFF
--- a/integreat_cms/cms/templates/pages/_generic_page_tree_node.html
+++ b/integreat_cms/cms/templates/pages/_generic_page_tree_node.html
@@ -41,11 +41,14 @@
         <a href="{% url 'edit_page' page_id=page.id region_slug=request.region.slug language_slug=language.slug %}"
            class="block py-1.5 px-2 overflow-hidden max-w-xs whitespace-nowrap text-ellipsis text-gray-800"
            title="{% if page_translation %} {{ page_translation.title }}{% endif %}">
-            {% if page_translation %}
-                {{ page_translation.title }}
-            {% else %}
-                <i>{% translate "Translation not available" %}</i>
-            {% endif %}
+            <div class="title-slug"
+                 data-title-slug="{% if page_translation %}{{ page_translation.slug }}{% endif %}">
+                {% if page_translation %}
+                    {{ page_translation.title }}
+                {% else %}
+                    <i>{% translate "Translation not available" %}</i>
+                {% endif %}
+            </div>
         </a>
     </td>
     {% if not backend_language %}
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -5493,12 +5493,14 @@ msgstr "Bitte versuchen Sie es später erneut."
 #: cms/templates/chat/_chat_widget.html
 #: cms/templates/statistics/_statistics_widget.html
 #: cms/templates/statistics/statistics_chart.html
+#: cms/templates/statistics/statistics_viewed_pages.html
 msgid "A server error has occurred."
 msgstr "Ein Serverfehler ist aufgetreten."
 
 #: cms/templates/chat/_chat_widget.html
 #: cms/templates/statistics/_statistics_widget.html
-#: cms/templates/statistics/statistics_chart.html xliff/utils.py
+#: cms/templates/statistics/statistics_chart.html
+#: cms/templates/statistics/statistics_viewed_pages.html xliff/utils.py
 msgid "Please contact the administrator."
 msgstr "Bitte kontaktieren Sie eine:n Administrator:in."
 
@@ -7943,14 +7945,17 @@ msgid "No pages available yet."
 msgstr "Noch keine Seiten vorhanden."
 
 #: cms/templates/pages/pages_page_tree.html
+#: cms/templates/statistics/statistics_viewed_pages.html
 msgid "Expand all"
 msgstr "Alle ausklappen"
 
 #: cms/templates/pages/pages_page_tree.html
+#: cms/templates/statistics/statistics_viewed_pages.html
 msgid "Collapse all"
 msgstr "Alle einklappen"
 
 #: cms/templates/pages/pages_page_tree.html
+#: cms/templates/statistics/statistics_viewed_pages.html
 msgid "Pages selected"
 msgstr "Seiten ausgewählt"
 
@@ -9060,20 +9065,37 @@ msgid "Export"
 msgstr "Exportieren"
 
 #: cms/templates/statistics/statistics_sidebar.html
-msgid "Choose format"
-msgstr "Format auswählen"
+msgid "Please select at least one page to export"
+msgstr "Bitte wählen Sie mindestens eine Seite für den Export aus."
+
+#: cms/templates/statistics/statistics_sidebar.html
+msgid "Please select at least one language to export"
+msgstr ""
+"Bitte wählen Sie mindestens eine Sprache für den export aus."
+
+#: cms/templates/statistics/statistics_sidebar.html
+msgid "Please select at least one page and one language to export"
+msgstr "Bitte wählen Sie mindestens eine Seite und eine Sprache für den export aus."
+
+#: cms/templates/statistics/statistics_sidebar.html
+msgid "Choose statistics to export"
+msgstr "Wähle Statistiken zum Exportieren aus"
 
 #: cms/templates/statistics/statistics_sidebar.html
 msgid "please select"
 msgstr "bitte auswählen"
 
 #: cms/templates/statistics/statistics_sidebar.html
-msgid "Image/PNG"
-msgstr "Bild/PNG"
+msgid "Image/PNG - Total Accesses"
+msgstr "Bild/PNG - Gesamtzugriffe"
 
 #: cms/templates/statistics/statistics_sidebar.html
-msgid "Table Document/CSV"
-msgstr "Tabellendokument/CSV"
+msgid "Table Document/CSV - Total Accesses"
+msgstr "Tabellendokument/CSV - Gesamtzugriffe"
+
+#: cms/templates/statistics/statistics_sidebar.html
+msgid "Table Document/CSV - Page Accesses"
+msgstr "Tabellendokument/CSV - Seitenzugriffe"
 
 #: cms/templates/statistics/statistics_viewed_pages.html
 msgid "Page views"
@@ -12035,6 +12057,18 @@ msgstr ""
 "Diese Seite konnte nicht importiert werden, da sie zu einer anderen Region "
 "gehört ({})."
 
+#~ msgid "Total Access Statistics"
+#~ msgstr "Statistiken für Gesamtzugriffe"
+
+#~ msgid "Page Access Statistics"
+#~ msgstr "Statistiken für Seitenzugriffe"
+
+#~ msgid "Choose format"
+#~ msgstr "Format auswählen"
+
+#~ msgid "Image/PNG"
+#~ msgstr "Bild/PNG"
+
 #~ msgid "Cannot generate slug from fallback."
 #~ msgstr "Slug kann nicht aus dem Fallback abgeleitet werden."
 
@@ -14419,11 +14453,6 @@ msgstr ""
 
 #~ msgid "POI was successfully deleted."
 #~ msgstr "POI wurde erfolgreich gelöscht."
-
-#~ msgid "Please create at least one language node before creating POIs."
-#~ msgstr ""
-#~ "Bitte erstellen Sie mindestens einen Sprach-Knoten, bevor Sie POIs "
-#~ "verwalten."
 
 #~ msgid "You cannot edit this POI because it is archived."
 #~ msgstr "Sie können diesen POI nicht bearbeiten, weil er archiviert ist."
--- a/integreat_cms/static/src/js/analytics/statistics-page-accesses.ts
+++ b/integreat_cms/static/src/js/analytics/statistics-page-accesses.ts
@@ -1,3 +1,5 @@
+import { updateSelectionCount, setCheckboxRecursively } from "../bulk-actions";
+
 /* Category as returned by get_page_accesses_ajax in statistics_actions.py is either a language slug or "total_accesses" */
 type AccessesPerLanguageAndInTotal = {
     [category: string]: number;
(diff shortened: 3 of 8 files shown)


jarlhengstmengel:
It's alive :partying_face:

dkehne (commented):
Nice feature! A few inline notes — two real bugs (`exportTable` init, `data-title-slug` leading space) plus some smaller points. Full context: the export path doesn't yet surface server/network errors the way `statistics-charts.ts` does.

integreat_cms/static/src/js/analytics/statistics-page-accesses.ts
```diff
@@ -11,6 +13,45 @@ let statisticsForm: HTMLFormElement;
 let pageAccessesURL: string;
 let pageAccessesForm: HTMLFormElement;
 let ajaxRequestID: number;
+let exportTable: string[][];
```
dkehne: `exportTable` is never initialized, so on an empty AJAX response (no data, or `!response.ok` → `{}`) it stays `undefined` and the export crashes at `[exportLabels].concat(exportTable)` → `undefined.join`. Suggest `let exportTable: string[][] = [];`.

integreat_cms/static/src/js/analytics/statistics-page-accesses.ts
```diff
@@ -163,19 +204,50 @@ const updateDOMRecursivly = (
...
 
+const updateExportTable = (data: AjaxResponse) => {
+    exportTable = [];
+    const checkedPages: NodeListOf<HTMLInputElement> = document.querySelectorAll(".bulk-select-item:checked");
+    checkedPages.forEach((page: HTMLInputElement) => {
+        const exportTableEntry: string[] = [];
+        const pageId = page.value;
+        const accesses = data[pageId];
+        const pageElement = document.getElementById(`page-${pageId}`);
+        const pageTitle = pageElement.querySelector(".title-slug").getAttribute("data-title-slug");
```
dkehne: `pageElement` (and the inner `.querySelector(".title-slug")`, which can be `null`) is dereferenced here — *before* the `if (pageElement && pageTitle)` guard below, so that guard is dead code. Suggest checking before the access and skipping the row otherwise.

integreat_cms/static/src/js/analytics/statistics-page-accesses.ts
```diff
@@ -163,19 +204,50 @@ const updateDOMRecursivly = (
...
+        if (pageElement && pageTitle) {
+            // Page Title needs to be utf-8 encoded for bto to work in exportPageAccessesData()
```
dkehne: `bto` → `btoa`. Also the value comes from `data-title-slug` = `page_translation.slug`, i.e. the slug, not the title (the CSV column is even labelled "Slug"). Consider renaming `pageTitle`/`pageTitleEncoded` → `pageSlug…` to match.

integreat_cms/static/src/js/analytics/statistics-page-accesses.ts
```diff
@@ -163,19 +204,50 @@ const updateDOMRecursivly = (
...
+            const pageTitleEncoded = new TextEncoder().encode(pageTitle);
+            let allAccesses: number = 0;
+
+            exportTableEntry.push(pageId);
+            exportTableEntry.push(String.fromCharCode(...pageTitleEncoded));
+
+            visibleDatasetSlugs?.forEach((languageSlug) => {
+                if (accesses && accesses[languageSlug]) {
+                    allAccesses += accesses[languageSlug];
+                    exportTableEntry[visibleDatasetSlugs.indexOf(languageSlug) + 2] = String(accesses[languageSlug]);
```
dkehne: `indexOf` re-scans the array each iteration; the index is already available via `forEach((slug, index) => …)`. A named constant for the `+2` offset (ID + Slug columns) would also aid readability.

integreat_cms/static/src/js/analytics/statistics-page-accesses.ts
```diff
@@ -187,22 +259,54 @@ export const updatePageAccesses = async (): Promise<void> => {
...
     if (!isEmpty && requestID === ajaxRequestID) {
-        updateDOM(data, visibleDatasetSlugs);
+        updateDOM(data);
+        updateExportTable(data);
     }
     pageAccessesLoading.classList.add("hidden");
 };
 
+const downloadFile = (filename: string, content: string) => {
+    const downloadLink = document.getElementById("export-download-link");
```
dkehne, line 274: This `downloadFile` is byte-identical to the one in `statistics-charts.ts`. Worth extracting to a shared module — note charts already imports from this file, so keep it here (or a neutral `export-utils.ts`) to avoid an import cycle.

integreat_cms/static/src/js/analytics/statistics-page-accesses.ts
```diff
@@ -187,22 +259,54 @@ export const updatePageAccesses = async (): Promise<void> => {
...
+    downloadLink.setAttribute("href", content);
+    downloadLink.setAttribute("download", filename);
+    downloadLink.click();
+};
+
+const exportPageAccessesData = (): void => {
+    const exportSelection = document.getElementById("export-statistics") as HTMLSelectElement;
+    const exportLabels: string[] = ["ID"].concat(["Slug"]).concat(visibleDatasetSlugs).concat(["Total Accesses"]);
```
dkehne: Could be simplified to `["ID", "Slug", ...visibleDatasetSlugs, "Total Accesses"]`.

integreat_cms/static/src/js/analytics/statistics-page-accesses.ts
```diff
@@ -187,22 +259,54 @@ export const updatePageAccesses = async (): Promise<void> => {
...
         statisticsForm.addEventListener("submit", async (event: Event) => {
             // Prevent form submit
             event.preventDefault();
             updatePageAccesses();
         });
+        document.getElementById("export-button")?.addEventListener("click", async () => {
+            const exportStatistics = document.getElementById("export-statistics") as HTMLSelectElement;
+            if (exportStatistics.value === "page-accesses-csv") {
+                // Wait for Page Accesses to be updated to ensure up to date export table
+                await updatePageAccesses();
```
dkehne, line 322: `updatePageAccesses` relies on `getData`, which returns `{}` on `!response.ok` (rendered to the user as "no accesses") and has no `try/catch/finally` (a network error leaves the spinner spinning). `statistics-charts.ts` already surfaces server/network errors via dedicated elements — worth mirroring, especially since the export now depends on this (see the `exportTable` note). Root cause is pre-existing.

integreat_cms/cms/templates/pages/_generic_page_tree_node.html
```diff
@@ -41,11 +41,14 @@
         <a href="{% url 'edit_page' page_id=page.id region_slug=request.region.slug language_slug=language.slug %}"
            class="block py-1.5 px-2 overflow-hidden max-w-xs whitespace-nowrap text-ellipsis text-gray-800"
            title="{% if page_translation %} {{ page_translation.title }}{% endif %}">
-            {% if page_translation %}
-                {{ page_translation.title }}
-            {% else %}
-                <i>{% translate "Translation not available" %}</i>
-            {% endif %}
+            <div class="title-slug"
+                 data-title-slug="{% if page_translation %} {{ page_translation.slug }}{% endif %}">
```
dkehne: The leading space inside the `{% if %}` isn't trimmed by `getAttribute`, so every exported slug is prefixed with a space. Also, when there's no translation the attribute is `""`, making the `pageTitle` guard in `updateExportTable` falsy → the selected page is silently dropped from the export. Suggest `{% if page_translation %}{{ page_translation.slug }}{% endif %}`.

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -9046,20 +9049,24 @@ msgid "Export"
 msgstr "Exportieren"
 
 #: cms/templates/statistics/statistics_sidebar.html
-msgid "Choose format"
-msgstr "Format auswählen"
+msgid "Choose statistics to export"
+msgstr "Wähle Statistiken zum exportieren aus"
```
dkehne: nit: `zum Exportieren` (capitalised — substantivised verb).

integreat_cms/static/src/js/analytics/statistics-page-accesses.ts
jarlhengstmengel: We already catch that in `getData` with `!respons.ok` and always get to a point where we stop the spinner spinning (`pageAccessesLoading.classList.add("hidden")` in `updatePageAccesses`). But I'll add an arror message for the users

integreat_cms/static/src/js/analytics/statistics-page-accesses.ts
jarlhengstmengel: I'll use unshift for adding pageId and slug, that should remove the need for an offset

jarlhengstmengel:
@dkehne I implemented your suggestions!

hannaseithe:
Before I start the review, I want to remark on an earlier architectural decision: **reusing `_generic_page_tree_node.html` and the corresponding frontend logic in `fetch-subpages.ts` and `toggle-subpages.ts`** (introduced in #3325). 

Looking at it now, I would have preferred those concerns to be kept separate. I've created #4366 to track a proper decoupling (on the frontend side) as part of the feature module refactor.

integreat_cms/static/src/js/analytics/statistics-page-accesses.ts
```diff
@@ -163,19 +206,49 @@ const updateDOMRecursivly = (
...
+            exportTableEntry.unshift(String.fromCharCode(...pageSlugEncoded));
+            exportTableEntry.unshift(pageId);
+            exportTable.push(exportTableEntry);
+        }
+    });
+};
+
 /* The main function which updates the accesses */
 export const updatePageAccesses = async (): Promise<void> => {
     const pageAccessesLoading = document.getElementById("page-accesses-loading");
```
hannaseithe, line 252:
```suggestion
    document.getElementById("accesses-server-error")?.classList.add("hidden");
    const pageAccessesLoading = document.getElementById("page-accesses-loading");
```

error should be cleared on next manual refresh after error

hannaseithe (changes requested):
Thank you for working on this feature!. 

I feel that the user should be informed that they forgot to either select any language or any pages, instead of silently just generating a table with no data.

Also if generate the table with all pages and languages selected, and afterwards deselect all languages, it still generates the table with all the data for all languages, but the language-slug's as column names missing.

Image

jarlhengstmengel:
> Before I start the review, I want to remark on an earlier architectural decision: reusing _generic_page_tree_node.html and the corresponding frontend logic in fetch-subpages.ts and toggle-subpages.ts (introduced in https://github.com/digitalfabrik/integreat-cms/pull/3325).

> Looking at it now, I would have preferred those concerns to be kept separate. I've created https://github.com/digitalfabrik/integreat-cms/issues/4366 to track a proper decoupling (on the frontend side) as part of the feature module refactor.

Yes, I think it's a case of we didn't knew better at the time. I would only advocate to do the refactoring after this is finished to not drag releasing this further out. I know it's not clean to merge something of that we know it contains technical debt. Thank you for creating the Issue!

Thank you for the review and suggestions, I'll implement them :)

hannaseithe:
> > Before I start the review, I want to remark on an earlier architectural decision: reusing _generic_page_tree_node.html and the corresponding frontend logic in fetch-subpages.ts and toggle-subpages.ts (introduced in #3325).
> 
> > Looking at it now, I would have preferred those concerns to be kept separate. I've created #4366 to track a proper decoupling (on the frontend side) as part of the feature module refactor.
> 
> Yes, I think it's a case of we didn't knew better at the time. I would only advocate to do the refactoring after this is finished to not drag releasing this further out. I know it's not clean to merge something of that we know it contains technical debt. Thank you for creating the Issue!
> 
> Thank you for the review and suggestions, I'll implement them :)

Absolutely - I will not touch the statistics for frontend refactor before this is declared done. I was just contemplating today how things could be cleanly decoupled. But i will definitely wait for the final result.

hannaseithe (approved):
Looks good to me!

dkehne (approved):

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -9046,20 +9051,37 @@ msgid "Export"
 msgstr "Exportieren"
 
 #: cms/templates/statistics/statistics_sidebar.html
-msgid "Choose format"
-msgstr "Format auswählen"
+msgid "Please select at least one page to export"
+msgstr "Bitte wählen Sie mindestens Seite für den export aus."
```
MizukiTemma:
```suggestion
msgstr "Bitte wählen Sie mindestens eine Seite für den export aus."
```

I don't speak German as native but guess we need here "eine"? 😅

MizukiTemma (approved):
Thank you so much 😻 Looks goof 🚀 

I have a suggestion but it's very very tiny so alreday appove the PR ✅

integreat_cms/locale/de/LC_MESSAGES/django.po
hannaseithe: Let me join the nitpick :D:

+ "export" should be spelled with capital letter

PR #3333, opened 2025-01-14
Replace `black` and `pylint` with `ruff`

charludo:
### Short description

I am SO sorry, dear reviewers 🙈 
@cclauss  do you have an opinion on this? 😄 

### Proposed changes

- `black` -> `ruff`
- `pylint` -> `ruff`

### Side effects

- hahahaha
- 👀 

### Resolved issues

Fixes: #3234 

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #3395, opened 2025-02-04
tools/ruff.sh: Allow restricting checking to specific subtrees

PeterNerlich:
### Short description

With #3333 just merged, ruff became our main linting tool. However if you're like me and have a lot of messy files and experiments lying around your working directory, you might want to restrict ruff to only look at a few subtrees in your project. While the performance improvement barely matters because ruff is already so fast, you will appreciate having irrelevant files untouched and ruff only printing messages about files you care about.

### Proposed changes

- Split `ruff.sh` into `ruff_check.sh` and `ruff_format.sh`, where arguments are handed over to ruff verbatim!
- `ruff.sh` now just calls those two scripts, and hands paths over to both if you specified them:
  - Parse arguments to `tools/ruff.sh` and save them to `PATHS`
  - Set `PATHS` to `BASE_DIR` if none are supplied
  - Run both `ruff_check.sh` and `ruff_format.sh` only for `PATHS`
  - Pass `--as-precommit` as is to both if given

### Side effects

- None, this change is fully backwards compatible to the previous intended usage. But if you relied on the tool ignoring any arguments at the end, it will now either complain that the file doesn't exist or lint those files.

### DIFF
--- a/tools/ruff.sh
+++ b/tools/ruff.sh
@@ -6,11 +6,13 @@
 # shellcheck source=./tools/_functions.sh
 source "$(dirname "${BASH_SOURCE[0]}")/_functions.sh"
 
+as_precommit=()
 PATHS=()
 
 # Parse given command line arguments
 while [[ $# -gt 0 ]]; do
     case $1 in
+        '--as-precommit') as_precommit+=("$1");shift;;
         *) PATHS+=("$1");shift;;
     esac
 done
@@ -31,7 +33,7 @@ else
 fi
 
 # shellcheck source=./tools/ruff_check.sh
-"$(dirname "${BASH_SOURCE[0]}")/ruff_check.sh" "${PATHS[@]}"
+"$(dirname "${BASH_SOURCE[0]}")/ruff_check.sh" "${as_precommit[@]}" "${PATHS[@]}"
 
 # shellcheck source=./tools/ruff_format.sh
-"$(dirname "${BASH_SOURCE[0]}")/ruff_format.sh" "${PATHS[@]}"
+"$(dirname "${BASH_SOURCE[0]}")/ruff_format.sh" "${as_precommit[@]}" "${PATHS[@]}"
--- a/tools/ruff.sh
+++ b/tools/ruff.sh
@@ -6,15 +6,32 @@
 # shellcheck source=./tools/_functions.sh
 source "$(dirname "${BASH_SOURCE[0]}")/_functions.sh"
 
-require_installed
+PATHS=()
 
-# Run ruff as a pre-commit hook
-run_as_precommit "ruff check" "$@"
-run_as_precommit "ruff format --check" "$@"
+# Parse given command line arguments
+while [[ $# -gt 0 ]]; do
+    case $1 in
+        *) PATHS+=("$1");shift;;
+    esac
+done
 
-# Run ruff
-echo "Starting code linting and formatting with ruff..." | print_info
-ruff check --fix "${BASE_DIR}" || true
-ruff format "${BASE_DIR}"
+if [ ${#PATHS[@]} -gt 0 ]; then
+    # Check whether paths exist
+    for p in "${PATHS[@]}"; do
+        if [[ -e "${p%%::*}" ]]; then
+            RUFF_ARGS+=("${p}")
+        elif [[ -n "${p}" ]]; then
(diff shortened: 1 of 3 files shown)


charludo (commented):
Thank you, I think this is a great idea! I have an extension of this in mind though, which also brings some simplification. It looks like `ruff check` and `ruff format --check` do the exact same thing as `ruff check .` and `ruff format --check .`, meaning we could just default to not providing a path at all, and simply pass all arguments given to `./tools/ruff.sh` to both of those calls, the same way we're already doing for the pre-commit hooks.

#### Cons:
-  no checking if passed paths actually exist. Do you have a specific usecase for this in mind? Otherwise I think `ruff`'s own error message is already pretty succinct:
    ```
    error: Failed to format help-i-dont-exist: No such file or directory (os error 2)
    ```

#### Pros:
- much simpler logic. Just need to change 
    ```bash
    ruff check --fix . || true
    ruff format .
    ```
    to 
    ```bash
    ruff check --fix "$@" || true
    ruff format "$@"
    ```
- more importantly though, that makes it trivial to pass additional flags to ruff. For example, I have really been wanting a way to append `--unsafe-fixes` a lot lately 😁 
(OK, to be fair this will throw some errors because `format` and `check` don't share a lot of flags, but it would certainly still be servicable)

PeterNerlich:
> we could just default to not providing a path at all, and simply pass all arguments given to `./tools/ruff.sh` to both of those calls

Yeah I'm also fine with that, I just copied it from `tools/test.sh` and removed everything that was tests-specific… as in, everything but the overarching structure :D

>   (OK, to be fair this will throw some errors because `format` and `check` don't share a lot of flags, but it would certainly still be servicable)

Yes I tried to look it up but didn't immediately find a comprehensive list of available options for both commands, so I just left it at the paths. I was wanted to keep the structure for this actually, so we could add flags and what to do with it later (to add it to one or the other ruff command, or some custom logic)

charludo:
> > we could just default to not providing a path at all, and simply pass all arguments given to `./tools/ruff.sh` to both of those calls
> 
> Yeah I'm also fine with that, I just copied it from `tools/test.sh` and removed everything that was tests-specific… as in, everything but the overarching structure :D
> 
> > (OK, to be fair this will throw some errors because `format` and `check` don't share a lot of flags, but it would certainly still be servicable)
> 
> Yes I tried to look it up but didn't immediately find a comprehensive list of available options for both commands, so I just left it at the paths. I was wanted to keep the structure for this actually, so we could add flags and what to do with it later (to add it to one or the other ruff command, or some custom logic)

Hm. It would probably be easier to split the tool into two (`ruff_format.sh`, `ruff_check.sh`) and call both from `ruff.sh` than to mainting a list of what-arg-goes-where for every conceivable usecase. 

The custom logic right now is not doing anything except remove non-existing paths, and print an error similar to what ruff would do anyways, right? IMO that is not enough benefit to justify 25 lines of shellscript 🙈

charludo:
What's the state of this PR?

jonbulz (changes requested):
Thanks, calling the script from command line works as expected. I'm having trouble running it as pre-commit, though:
```/path/to/integreat-cms/tools/ruff.sh: Zeile 21: print_error: Befehl nicht gefunden```

**Comment by PeterNerlich** [comment]
> `print_error: Befehl nicht gefunden`

Right, I removed sourcing the utility functions from the main script because I didn't need it anymore and immediately put a call to one of `print_error` back in :see_no_evil:

**Review by jonbulz — COMMENTED** [review]
Right, that fixes the error, thanks! I'm still having trouble running pre-commit, though:
```
ruff (custom script).....................................................Failed
- hook id: ruff
- exit code: 1

--as-precommit: No such file or directory
```
For reference, I was trying to commit some random code that should trigger ruff formatting.

PeterNerlich:
> I'm still having trouble running pre-commit, though

I don't use the pre-commit hooks, so it's definitely a blind spot I have. I don't quite understand idea and implementation of `--pre-commit` in our tool scripts, but I think this should probably fix it. Please try again and see whether it now behaves as expected.

JoeyStk (approved):
Thank you for this PR! This works as expected :rocket:

PeterNerlich:
> This is not testable (for our testing team), right? :)

Correct, this is only about the workflow with dev tools and has no impact on the produced software itself

PR #3401, opened 2025-02-05
Check if Integreat Development Header is set for requests

ulliholtgrave:
### Short description

Check if the X-INTEGREAT-DEVELOPMENT header is set before forwarding the request to matomo.

### Proposed changes

- Check if the HTTP header "X-Integreat-Development" is set for requests to matomo.


charludo (changes requested):
The actual change looks good, but please revert themypy change 😄 

I am a bit confused though, what's the error you receive? mypy is not throwing any on your branch for me, neither through the devtool nor through pre-commit.

charludo (approved):

svenseeberg (approved):

Issue #3391, opened 2025-02-04
HIX score is calculated based on Contact details if the only content is the contact


juliankehne:
### Describe the Bug

When the only content on a page is one Contact it still calculates the HIX Score based on the content of the Contact.
When you add a 2nd Contact or more it behaves as it should and excludes the Contact from HIX score calculation no HIX score is calculated.

### Steps to Reproduce

1. Go to 'Pages'
2. Click 'Create page'
3. Set a title
4. In the editor insert any contact and nothing else
5. Click 'Publish'
6. See that HIX score was calculated

7. Insert a second contact
8. Click Publish
9. See that now that instead of HIX Score it says "There is no page content"

10. Remove second contact and publish again
11. Now it also says "There is no page content" although its back to only 1 contact

### Expected Behavior

Inserted Contacts should always be excluded from HIX Score calculation.

### Actual Behavior

Inserted Contact is used in HIX Score calculation if it is the only content on a page.

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

### Related Issues

Related to https://github.com/digitalfabrik/integreat-cms/issues/3268 and the PR

charludo:
Good catch. Looks like the reason is that if you insert a contact into a completely empty page, the entire page content is that div. In order to remove the cards before HIX, the card's parent element is queried, which does not exist in that case.

Potential fixes:
- check if a parent exists, else return `None` from the HIX lookup
- wrap the contact cards in an additional `div`

Both would be fine IMO, and probably `good first issue` can be added.

BUt also, I think this is not really critical right now - contact cards can not be translated anyways

PR #3407, opened 2025-02-06
Fix hix calculation for contact card on empty page

JoeyStk:
### Short description

We disabled the hix calculation for contact cards. However there was still one case in which this didn't work. When a contact was inserted on an otherwise empty page / newly created page HIX still got calculate. This PR attempts to fix this.

### Proposed changes

- Disable HIX calculation for contact cards on otherwise empty pages

### Side effects

- I was told that inserting divs can have impacts on the loading time - but I'm unsure how much effect this really has in this case 

### Resolved issues

Fixes: #3391


charludo (approved):
NIce and easy solution! 😁 

> I think a test would be good for this, but I'm unsure how to do it

There's `test_skip_textlab_call_if_html_is_empty` in the existing tests, something like 

```python
def test_only_noneditable_still_gets_filtered_out() -> None:
    html = "<div><div contenteditable='false'>some content</div></div>"
    result = lookup_hix_score_helper(html)

    assert result is not None
    assert result["score"] is None
```

is probably sufficient (well, maybe with a better name). Succeeds as-is, but fails if you rename the `contenteditable` attribute. Note that the "some content" is crucial, since otherwise the test passes trivially, `contenteditable` or not, because `text_content()` is empty.

Testing the negative of this is probably difficult, or rather, it's indistinguishable from checking if HIX score generation works, so I would ignore that direction.

JoeyStk:
Thank you very much!
My ideas for the name would be:

- test_contact_card_gets_filtered_out (to be specific -> it's immediately clear what it's referring to)
- test_only_noneditable_gets_ignored (saying pretty much the same, but a bit shorter than the original idea)

What do you think about the suggestions? But I would also be fine with keeping the existing name :)

charludo:
> Thank you very much! My ideas for the name would be:
> 
>     * test_contact_card_gets_filtered_out (to be specific -> it's immediately clear what it's referring to)
> 
>     * test_only_noneditable_gets_ignored (saying pretty much the same, but a bit shorter than the original idea)
> 
> 
> What do you think about the suggestions? But I would also be fine with keeping the existing name :)

I like the first one!

But both are good 😄

MizukiTemma (approved):
👍

Issue #3379, opened 2025-01-29
Adjust event filter for long-term even


MizukiTemma:
### Motivation

#3355 introduces a new event type "long-term". The event filter needs to be adjusted accordingly.

### Proposed Solution

- Add a new check box "long-term events" to "Recurring events" and "Non-recurreing events"

Other adjustments must be delievred togather if necessary.

PR #3410, opened 2025-02-07
Adjust filters for long term events

JoeyStk:
### Short description

This PR refactors the event filters, adjusts the filters for long term events and adds a test.

### Proposed changes

- Refactor event list
- Adjust filters
- Add test

### Side effects

- Please check if I didn't break anything with the refactoring
- I deleted all comments as I think the code is more speaking now - but this might be "author blindness", so please let me know if you want to restore any comments

### Resolved issues

Fixes: #3379

### DIFF
--- a/integreat_cms/cms/constants/recurrence.py
+++ b/integreat_cms/cms/constants/recurrence.py
@@ -22,11 +22,14 @@ DATATYPE: Final = int
 
 #: Events that are recurring, i.e. take place more than once
 RECURRING: Final = 1
-#: Events that are not recurring, i.e. take place only once
-NOT_RECURRING: Final = 2
+#: Events that are not recurring and only take place on one day
+ONE_TIME: Final = 2
+#: Events that are not recurring but take place over a long period of time
+LONG_TERM: Final = 3
 
 #: Choices to use these constants in a database field
 CHOICES: Final[list[tuple[int, Promise]]] = [
     (RECURRING, _("Recurring events")),
-    (NOT_RECURRING, _("Non-recurring events")),
+    (ONE_TIME, _("One time events")),
+    (LONG_TERM, _("Long term events")),
 ]
--- a/integreat_cms/cms/forms/events/event_filter_form.py
+++ b/integreat_cms/cms/forms/events/event_filter_form.py
@@ -10,6 +10,7 @@ from datetime import date, datetime, time
 from typing import TYPE_CHECKING
 
 from django import forms
+from django.db.models import F, Q
 
 if TYPE_CHECKING:
     from ...models import Region
@@ -89,6 +90,9 @@ class EventFilterForm(CustomFilterForm):
         :param events: the unfiltered events
         :param events: the filtered events
         """
+        if not self.is_enabled:
+            return events
+
         cleaned_time_range = self.cleaned_data["events_time_range"]
         if not cleaned_time_range or set(cleaned_time_range) == set(
             events_time_range.ALL_EVENTS,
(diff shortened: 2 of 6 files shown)


JoeyStk:
If some knows how to fix the tests, I'm open to suggestions! @jonbulz Mizuki gave me the hint that you might something about this :)

JoeyStk:
> Why did you re-name the events anyway?

My idea was to be explicit. The test event is a weekly recurring event and we use it for this case and while refactoring I thought it would be the cleaner way to rename it and to be clear about what the test is for. I wouldn't mind it too much if I had to name it back, but don't think it's ideal :)

integreat_cms/cms/forms/events/event_filter_form.py
```diff
@@ -82,35 +83,21 @@ class EventFilterForm(CustomFilterForm):
...
+        Filter events by time range
 
-        :param events: The list of events
-        :param region: The current region
-        :param language_slug: The slug of the current language
-        :return: The filtered list of events, the poi used for filtering, and the search query
+        :param events: the unfiltered events
+        :param events: the filtered events
         """
-        if not self.is_enabled:
```
jonbulz, line 99: I did some digging as to why tests are failing (e.g. test_view_status_code_12) and found that some EventFilterForms that are not "enabled" have no "events_time_range" in self.cleaned_data, thus throwing a KeyError here

JoeyStk: Interesting! Thank you very much!
Please feel free to commit to this branch if you find a fix :)

But aside from that I have one question. So, I still have a check [if events are enabled](https://github.com/digitalfabrik/integreat-cms/pull/3410/files#diff-10626a88b853c33e85394820b84a767b701d3b51fc0a2e9b61d4f56403a93b2eR233-R234), which is the same except for the return statement. Do you think that's the problem?

jonbulz (approved):

jonbulz:
@JoeyStk it took longer than anticipated, but I finally fixed the tests. Could you have a look at my changes?

JoeyStk:
Thank you soooo much! :heart_eyes: I'll check it out in a bit :)

integreat_cms/release_notes/current/unreleased/3379.yml
```diff
@@ -0,0 +1,2 @@
+en: Adjust event filters for long-term events
+de: Passe die Veranstaltungsfilter für langfristige Ereignisse an
```
JoeyStk, line 2:
```suggestion
de: Passe die Veranstaltungsfilter für fortlaufende Ereignisse an
```

JoeyStk:
To me it looks and works as expected! So this PR is ready to have it's second review :)

JoeyStk:
I increased the importance of this PR as it's the follow-up PR to an already existing feature

integreat_cms/cms/constants/recurrence.py
```diff
@@ -22,11 +22,14 @@
 
 #: Events that are recurring, i.e. take place more than once
 RECURRING: Final = 1
-#: Events that are not recurring, i.e. take place only once
-NOT_RECURRING: Final = 2
+#: Events that are not recurring and only take place on one day
+SINGLE_DAY: Final = 2
```
MizukiTemma, line 26: Nit: currently "One-time" is used for this kind of event. Probably we should align with the existing naming schema or change the existing one.

MizukiTemma (approved):
Works as expected 😸 Thank you!

integreat_cms/cms/constants/recurrence.py
JoeyStk: Good point. I changed them now :)

Issue #3327, opened 2025-01-13
Move the 'Page was saved automatically' text on the pages form


juliankehne:
### Motivation

When the 'Page was saved automatically' message is shown, it is placed in between the new Cancel button and the Preview button creating a gap between the two buttons and it looks out of place. 

### Proposed Solution

Sort the ui elements so that the 'saved automatically' text is shown to the left of the Cancel button.

### Alternatives

Keep as it is for now.

### Additional Context

![Screenshot from 2025-01-13 09-33-50](https://github.com/user-attachments/assets/97d3ef63-fc75-4ee3-880a-348f6f932000)

### Related Issues

The cancel button:
https://github.com/digitalfabrik/integreat-cms/issues/2930

JoeyStk:
@nikolahoff Do you have an opinion what the best solution would be? Thank you in advance :)

nikolahoff:
@JoeyStk yes the proposal of @juliankehne makes sense!

PR #3415, opened 2025-02-10
Move automatically saved note

JoeyStk:
### Short description

Since adding the "Cancel" button on content forms, the info message that the page/event/poi has been automatically saved is in between two buttons, which doesn't look great. This PR moves the note so that it shows to the far left again.

### Proposed changes

- Move automatically saved note

### Side effects

- I'm not sure for which permission to check in the event form. If you have any suggestion, please let me know :)

### Resolved issues

Fixes: #3327 

### DIFF
--- a/integreat_cms/cms/templates/pages/page_form.html
+++ b/integreat_cms/cms/templates/pages/page_form.html
@@ -45,6 +45,9 @@
                 {% endif %}
             </h1>
             <div class="flex flex-wrap grow justify-end gap-2 items-center">
+                {% if page_form.instance.id and not page_form.instance.archived and can_edit_page %}
+                    {% include "generic_auto_save_note.html" with form_instance=page_translation_form.instance %}
+                {% endif %}
                 {% if page_form.instance.id and page_form.instance.archived %}
                     <a href="{% url 'archived_pages' region_slug=request.region.slug language_slug=language.slug %}"
                        class="btn btn-ghost">{% translate "Cancel" %}</a>
@@ -53,7 +56,6 @@
                        class="btn btn-ghost">{% translate "Cancel" %}</a>
                 {% endif %}
                 {% if not page_form.instance.id or not page_form.instance.archived %}
-                    {% include "generic_auto_save_note.html" with form_instance=page_translation_form.instance %}
                     <button name="preview_page" type="button" data-preview-page data-edit-page-mode data-right-to-left={{ right_to_left }} class="btn btn-ghost">
                         {% translate "Preview" %}
                     </button>
--- a/integreat_cms/cms/templates/events/event_form.html
+++ b/integreat_cms/cms/templates/events/event_form.html
@@ -37,6 +37,9 @@
                 {% endif %}
             </h1>
             <div class="flex flex-wrap gap-4 ml-auto mr-0 items-center">
+                {% if event_form.instance.id and not disabled %}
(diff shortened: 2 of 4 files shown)


JoeyStk:
> There are some suggestion, assuming that the following is the condition to show the auto save note: the object has an ID (alredy exists) AND the object is not archived AND the user can change or publish the object.

Thank you very much! Could you please give me an example where we might run into a problem with the current implementation. I'm not doubting your suggestions, I'm just wondering :)

JoeyStk:
@MizukiTemma thank you for your suggestions! This PR is ready to be reviewed again :)

integreat_cms/cms/templates/pages/page_form.html
```diff
@@ -45,6 +45,9 @@ <h1 class="heading">
                 {% endif %}
             </h1>
             <div class="flex flex-wrap grow justify-end gap-2 items-center">
+                {% if page_form.instance.id and not page_form.instance.archived and can_edit_page %}
```
jarlhengstmengel, line 48:
```suggestion
                {% if not page_form.instance.id and not page_form.instance.archived and can_edit_page %}
```
On the old location it was inside a if statement with `if not page_form.instance.id`. Is there a specific reason to change that?

jarlhengstmengel (changes requested):
Thanks for the PR. I ran in a bug this PR seems to introduce, at least i couldn't reproduce it in develop. I couldn't navigate to the Events and POIs page and got an error with `column cms_poi.website does not exist`

integreat_cms/cms/templates/pages/page_form.html
JoeyStk: This was a requested change by Mizuki to be consistent in all three content forms

JoeyStk:
> Thanks for the PR. I ran in a bug this PR seems to introduce, at least i couldn't reproduce it in develop. I couldn't navigate to the Events and POIs page and got an error with column cms_poi.website does not exist

Could you please try to run `./tools/prune_database.sh` and then `./tools/run.sh` again? :)

jarlhengstmengel:
> > Thanks for the PR. I ran in a bug this PR seems to introduce, at least i couldn't reproduce it in develop. I couldn't navigate to the Events and POIs page and got an error with column cms_poi.website does not exist
> 
> Could you please try to run `./tools/prune_database.sh` and then `./tools/run.sh` again? :)

Dropping the Table manually fixed it, After `prune_database.sh`I somehow had the same problem still with POIs :sweat_smile:

integreat_cms/cms/templates/pages/page_form.html
```diff
@@ -45,6 +45,9 @@ <h1 class="heading">
                 {% endif %}
             </h1>
             <div class="flex flex-wrap grow justify-end gap-2 items-center">
+                {% if not page_form.instance.id and not page_form.instance.archived and can_edit_page %}
```
jarlhengstmengel, line 48:
```suggestion
                {% if page_form.instance.id and not page_form.instance.archived and can_edit_page %}
```

jarlhengstmengel (changes requested):
I'm so sorry, you were right :see_no_evil: With my former suggestion, the node isn't shown at all

jarlhengstmengel (approved):
Should be good to go, thank you very much!

Issue #3319, opened 2025-01-07
500 error when saving page after changing title


osmers:
### Describe the Bug
Sometimes a 500 error is shown after changing the title of a page in Farsi, Arabic and Russian (maybe other languages too) and trying to save the changes. Although the 500 errror is shown, the changes are still saved. 
I haven't yet found a way to reliably reproduce this...

### Steps to Reproduce

Steps to always reproduce:

1. Go to any page
2. Change the slug to anything thats different to the title (optional)
3. Click on the title field to focus it
4. Click Update
5. See Error

or

1. Go to any page
2. Click on the title field to focus it
3. Change the title to something different
4. Do not click anywhere else or lose focus of title field
5. Click Update
6. See Error

### Expected Behavior
Page is saved without 500 error.

### Actual Behavior
Sometimes 500 error shows up but changes are still saved.

### Additional Information

<details>
<summary>Traceback</summary>

```
Jan 16 18:30:27 ERROR   django.request - 500 Internal Server Error: /dortmund/pages/fa/10877/edit/
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/django/views/generic/base.py", line 104, in view
    return self.dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/django/utils/decorators.py", line 46, in _wrapper
    return bound_method(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/django/contrib/auth/decorators.py", line 23, in _wrapper_view
    return view_func(request, *args, **kwargs)
[...]

PR #3422, opened 2025-02-10
Attempt to fix issue with none existant other translation

JoeyStk:
### Short description

This PR is an attempt to fix the `NoneType` error that occurs sometimes when renaming a page. Some more testing on our side show that this issue is indeed flaky and seems to be caused because `other_translation` is `None`.  

### Proposed changes

- Catch case when `other_translation` is still `None` and show an error message instead. 

### Side effects

- This is not ideal for user experience. An alternative could be to wait for a while to make sure there is enough time for `other_translation` to be set.
- Please note that this really is only a draft and I'm shooting out ideas hoping that it sparks some more ideas from your side :sparkles: 

### Resolved issues

Fixes: Parts of #3319

### DIFF
--- a/integreat_cms/cms/views/pages/page_form_view.py
+++ b/integreat_cms/cms/views/pages/page_form_view.py
@@ -359,33 +359,46 @@ class PageFormView(
                     slug=user_slug,
                     language=language,
                 ).first()
-                other_translation_link = reverse(
-                    "page_versions",
-                    kwargs={
-                        "page_id": other_translation.page.id,
-                        "language_slug": language.slug,
-                        "region_slug": region.slug,
-                        "selected_version": other_translation.version,
-                    },
-                )
-                message = _(
-                    "The slug was changed from '{user_slug}' to '{slug}', "
-                    "because '{user_slug}' is already used by <a>{translation}</a>.",
-                ).format(
-                    user_slug=user_slug,
-                    slug=page_translation_form.cleaned_data["slug"],
-                    translation=other_translation,
-                )
-                messages.warning(
-                    request,
-                    translate_link(
-                        message,
-                        attributes={
-                            "href": other_translation_link,
-                            "class": "underline hover:no-underline",
+                if other_translation:
+                    other_translation_link = reverse(
+                        "page_versions",
+                        kwargs={
+                            "page_id": other_translation.page.id,
+                            "language_slug": language.slug,
+                            "region_slug": region.slug,
+                            "selected_version": other_translation.version,
                         },
-                    ),
-                )
+                    )
+                    message = _(
+                        "The slug was changed from '{user_slug}' to '{slug}', "
+                        "because '{user_slug}' is already used by <a>{translation}</a>.",
+                    ).format(
+                        user_slug=user_slug,
+                        slug=page_translation_form.cleaned_data["slug"],
+                        translation=other_translation,
+                    )
+                    messages.warning(
+                        request,
+                        translate_link(
+                            message,
+                            attributes={
+                                "href": other_translation_link,
+                                "class": "underline hover:no-underline",
+                            },
+                        ),
+                    )
+                else:
+                    logger.warning(
+                        "Slug was changed from the one the user provided, but we can't find the translation that already used it: %s (cleaned to %s)",
+                        user_slug,
+                        page_translation_form.cleaned_data["slug"],
+                    )
+                    messages.warning(
+                        _("The slug was changed from '{user_slug}' to '{slug}.").format(
+                            user_slug=user_slug,
+                            slug=page_translation_form.cleaned_data["slug"],
+                        )
+                    )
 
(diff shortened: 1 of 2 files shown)


JoeyStk:
@MizukiTemma What do we do with this PR? Should I close it?

MizukiTemma (approved):
@JoeyStk 
Sorry for having took a long time to come to this PR 🙏 

I sadly didn't manage to reproduce the bug 😢 but aprove this PR as nothing bad was observed and I'm not against your idea (or rather to say, I don't have alternative 😅)

integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -359,33 +359,39 @@ def post(
...
+                                "href": other_translation_link,
+                                "class": "underline hover:no-underline",
+                            },
+                        ),
+                    )
+                else:
+                    logger.error(
+                        "Renaming slug failed, because translation is not existant"
+                    )
+                    messages.error("This didn't work. Please try again in a moment.")
```
PeterNerlich, line 394: I don't think this is the correct message to give. *(I tried to be concise, but it turned into a bit of a rant… sorry)*
We **already saved the page translation form** on line 351, which was apparently successful or otherwise we would have aborted with an exception. Afterwards we just **checked whether the cleaned slug differed** from the one originally submitted by the browser. In the `clean_…()` method for that field we employed [`generate_unique_slug()` from `slug_utils.py`](https://github.com/digitalfabrik/integreat-cms/blob/dc5f02332bffbfcabf3a04ded93a662fa3ee9fec/integreat_cms/cms/utils/slug_utils.py#L89-L181), which is the part of the code that is trying to avoid collisions *(I didn't read through this, I just wanted to confirm the clean method is where this happens)*.

Let me reiterate, interestingly, this check happens after the translation has already been saved – now we're just looking to generate meaningful user facing messages. Since we only know that the slug changed, we **assume** there must have been a collision. It would be nice to tell the user which page we collided with, so we try to query it. Our error happens if, for whatever reason, we don't actually find a translation with that original slug. And this is the case that we want to convey at this point:

Not that saving it failed, or changing the slug was unsuccessful, we just didn't find a translation holding the originally requested slug.
So what should we do? We could either try to understand which other reasons there could be for `generate_unique_slug()` to come up with a slug different from the requested one and make appropriate explanation messages for that. Or we could just say that we don't know why, but the slug got changed and display that as a warning, not an error, affirming that otherwise, everything has been saved and no data is lost.

PeterNerlich:
As I wrote on the issue, I wasn't able to reproduce the bug, and using julians instructions I was only able to produce a different bug on integreat-test. So depending on whether we handle both in #3319 or split that off into a new bug issue this PR wouldn't fully resolve it.

MizukiTemma:
Another [occurence](https://chat.tuerantuer.org/digitalfabrik/pl/4xbincf5ktd93fj5wi3qd95qta) freshly reported

JoeyStk:
> So depending on whether we handle both in https://github.com/digitalfabrik/integreat-cms/issues/3319 or split that off into a new bug issue this PR wouldn't fully resolve it. 

I changed the description to only partly solving the issue (and not auto-close the issue after this PR) :)

integreat_cms/cms/views/pages/page_form_view.py
PeterNerlich: I think it's weird to tell the user "hey something happened but we don't know why our system did that" right in the message automatically given to them. But we could at least tell them it was saved with a different slug, not citing any reason. Maybe the user doesn't care about the slug, but better make it explicit.

I guess we could adapt the original message and just not mention a reason. As for the logging message, maybe something like this?
```suggestion
                    logger.warning(
                        "Slug was changed from the one the user provided, but we can't find the translation that already used it: %s (cleaned to %s)",
                        user_slug,
                        page_translation_form.cleaned_data["slug"],
                    )
                    messages.warning(
                        _("The slug was changed from '{user_slug}' to '{slug}.").format(
                            user_slug=user_slug,
                            slug=page_translation_form.cleaned_data["slug"],
                        )
                    )
```

JoeyStk:
> Another [occurence](https://chat.tuerantuer.org/digitalfabrik/pl/4xbincf5ktd93fj5wi3qd95qta) freshly reported

Yet another bug

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -4363,7 +4363,7 @@ msgstr "Angepasste Integreat Chat Datenschutzerklärung"
 msgid "Link to custom privacy policy for self-hosted Zammad server."
 msgstr ""
 "Link zur angepassten Datenschutzerklärung für einen selbstverwalteten Zammad-"
-"Server."
+"Server.Server"
```
PeterNerlich, line 4366: this looks like a mistake
```suggestion
"Server."
```

PeterNerlich (approved):

Issue #3414, opened 2025-02-10
Recurrence end date is set automatically (on test cms)


osmers:
### Describe the Bug
When adding a new event and setting the recurrence option - an end date in the past is set automatically

### Steps to Reproduce

1. Go to https://cms-test.integreat-app.de/birkenfeld/events/de/
2. Click on Create event
3. Check box recurring
4. See end date being set automatically

### Expected Behavior
Choice field is empty

### Actual Behavior
Choice field shows date in the past

### Additional Information

![Image](https://github.com/user-attachments/assets/b7e98373-f664-4ab8-aa3a-710af7c1d7b4)

JoeyStk:
I can reproduce this on CMS Test, but for some reason not locally. There I get the expected output

MizukiTemma:
@JoeyStk 
Same here, also #3413 🤔

JoeyStk:
I have an idea what it could be. I think it might because of the migration and in particular these lines here: https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/migrations/0113_alter_event_end.py#L42-L50

svenseeberg:
The date is already set when the page loads. It has nothing to do with hitting the "Recurring" check box.

Now the question is: why is the end date set as the default value for this setting? Also, why does it change? @osmers when was the screenshot taken? Today?

osmers:
@svenseeberg good to know - I couldn't see that :)
And yes, I took the screenshot today. Basically when I created the issue.

svenseeberg:
The date is already set in the response from the CMS, without any JS magic:

![Image](https://github.com/user-attachments/assets/23807c49-4cc2-4e57-ae57-99047dab1fbf)

svenseeberg:
Not that many locations where this variable is touched:
```
% grep -rn recurrence_end_date
static/dist/main.351da18e7a207ebda0f6.js:1651:  var recurrenceEndData = document.getElementById("id_has_recurrence_end_date");
api/v3/events.py:159:        event.recurrence_rule.recurrence_end_date
api/v3/events.py:160:        and event.recurrence_rule.recurrence_end_date < today
cms/migrations/0001_initial.py:1310:                    "recurrence_end_date",
cms/migrations/0113_alter_event_end.py:44:                and recurrence_rule.recurrence_end_date
cms/migrations/0113_alter_event_end.py:47:                    recurrence_rule.recurrence_end_date, end.time()
[...]

PR #3428, opened 2025-02-13
Fix recurrence end date

MizukiTemma:
### Short description

This PR fixes the initialization of recurrence end date.

### Proposed changes

- Set explicitly to the end date of the recurrence rule if exists, to `None` if not

### Side effects

- As the bug cannot be reproduced in the local environment, I'm not sure whether this fix really works. It won't hurt, however, to give it a try as long as there is no negative side effect.

### Resolved issues

Fixes: #3414 

### DIFF
--- a/integreat_cms/cms/views/events/event_form_view.py
+++ b/integreat_cms/cms/views/events/event_form_view.py
@@ -116,6 +116,13 @@ class EventFormView(
         recurrence_rule_form = RecurrenceRuleForm(
             instance=recurrence_rule_instance,
             disabled=disabled,
+            initial={
+                "recurrence_end_date": (
+                    recurrence_rule_instance.recurrence_end_date
+                    if recurrence_rule_instance
+                    else None
+                )
+            },
         )
 
         url_link = f"{settings.WEBAPP_URL}/{region.slug}/{language.slug}/{event_translation_form.instance.url_infix}/"


JoeyStk (approved):
Thank you very much! Let's try this one out! :rocket: 
If this doesn't work I think we should just try a few things out on cms-test before we open up a lot of PRs :)

integreat_cms/cms/views/events/event_form_view.py
```diff
@@ -116,6 +116,11 @@ def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
         recurrence_rule_form = RecurrenceRuleForm(
             instance=recurrence_rule_instance,
             disabled=disabled,
+            initial={
+                "recurrence_end_date": recurrence_rule_instance.recurrence_end_date
+                if recurrence_rule_instance
+                else None
+            },
```
PeterNerlich, line 123: First I was confused whether this was valid at all, until I realized that this wasn't a dict comprehension. I think this would make it more readable
```suggestion
            initial={
                "recurrence_end_date": (
                    recurrence_rule_instance.recurrence_end_date
                    if recurrence_rule_instance
                    else None
                )
            },
```

PeterNerlich (approved):
Works nicely!

Issue #3431, opened 2025-02-14
Reject all events with daily recurrence during import from external calendar


MizukiTemma:
### Motivation

Currently events with daily recurrence are imported as long-term event if any end date (event end date or recurrence end date) is provided. This result in, however, interval of recurrence goes lost. 

### Proposed Solution

Reject all events with daily recurrence, even if they have any end date.

### Alternatives

None

### Additional Context

Find the contect [here](https://chat.tuerantuer.org/digitalfabrik/pl/joh7fm8i7in6dmrq769ocnznoa)

PR #3432, opened 2025-02-14
Reject all daily events in import

MizukiTemma:
### Short description

This PR changes the event import functionality so no event with daily recurrence can be imported.

### Proposed changes

- Check the frequency and raise error if it is daily

### Side effects

- None?

### Resolved issues

Fixes: #3431 

### DIFF
--- a/integreat_cms/cms/utils/external_calendar_utils.py
+++ b/integreat_cms/cms/utils/external_calendar_utils.py
@@ -114,21 +114,14 @@ class IcalEventData:
 
         recurrence_rule = None
         if "RRULE" in event:
-            if (
-                event.decoded("RRULE")["FREQ"] == ["DAILY"]
-                and event.decoded("RRULE")["UNTIL"]
-            ):
-                if start_date == end_date:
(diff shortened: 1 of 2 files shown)


JoeyStk (commented):
Thank you very much! The code runs as expected and didn't cause problems. I only have one concern / one thing we should think of.
I imported the test calendar from Nextcloud that we used to test the imports. In there we have one event that repeats every day and one that repeats every two days. Both failed to be imported. I take it that this is expected. I just want to double check with @osmers and/or @hauf-toni if this is really what we want :)

osmers:
I can see the problem that this is not very transparent for municipalities bcs they cannot see the import option and which events failed). On the other hand, so far we have no municipality using the import function (I think)...
Is this an issue we can tackle later? As in, approve this now but open another issue on how to deal with these events to that they fit the new event design?

JoeyStk (approved):

MizukiTemma:
@osmers 
Thank you for reply. An issue is opened for the thema #3443

charludo (approved):
Thank you! 🎉

Issue #3434, opened 2025-02-17
Remove contact cards from linkchecker


JoeyStk:
### Motivation

As there are a few issues (#3280, #3398 and #3399 ) when it comes to contact cards and linkchecker, it was decided in https://github.com/digitalfabrik/integreat-cms/issues/3399#issuecomment-2653273426 that contact cards should be removed from the linkchecker as a temporary solution.

### Proposed Solution

Completely remove the contact cards from linkchecker for now.

### Alternatives

The long term solution is not yet perfectly defined, but is probably something like this: https://github.com/digitalfabrik/integreat-cms/issues/3399#issuecomment-2651352879. Needs to be discussed further.

### User Story

As a user of the CMS I don't want to get confused by outdated and faulty information from contact cards.

### Additional Context

#3280, #3398 and #3399

### Design Requirements

None

### Related Issues

#3280, #3398 and #3399

Issue #3413, opened 2025-02-10
recurrence field automatically selects daily


osmers:
### Describe the Bug
When adding a new event and setting the recurrence option - daily is automatically selected (which when trying to save the event, throws an error bcs daily cannot be selected)

### Steps to Reproduce

1. Go to https://cms-test.integreat-app.de/birkenfeld/events/de/
2. Click on Create event
3. Check box recurring
4. See daily being selected automatically

### Expected Behavior
Choice field is empty

### Actual Behavior
Choice field shows daily

### Additional Information

![Image](https://github.com/user-attachments/assets/db35037d-9600-43f9-9b05-403304a0c76f)

JoeyStk:
Next idea: set this field to None during creation might fix this

MizukiTemma:
With a dump data you can reproduce this bug locally.

The cause was orhaned recurrence rule objects.

#3355 brought a migration, which converts daily events with end date into long-term evets. Those events were deprived of their recurrence rules, and these recurrence rules were not deleted.

When an event form is opened, we pick up the rucurrence rule instance of the event, assuming we get `None` if the user is creating a new event (if it is `None`, the defaul value "Weekly" appears as pre-selected).
`recurrence_rule_instance = RecurrenceRule.objects.filter(
            event=event_instance,
        ).first()`
Now there are recurrence rule objects that have no relation to any event object. One of them is picked up when an new event is being created. All the orphaned recurrence rules are "daily". That's why now we are getting "Daily" automatically selected when creating a new event.


Solution1: just delete orphaned recurrence rule objects.
Solution2: fix the migration (related to [this thema](https://chat.tuerantuer.org/digitalfabrik/pl/4c3qjpx563yhdr18xnz5pxa6ho))

@JoeyStk FYI

MizukiTemma:
@osmers @@hauf-toni @JoeyStk 

As the migration stays unchanged, I have applied the solution 1.

JoeyStk:
@MizukiTemma thank you very much! Is there still something that needs to be done or can this issue be closed once this has been tested? :) 
Also do we also have to delete the orphaned recurrence rule objects on the live system after the release? :)

MizukiTemma:
> [@MizukiTemma](https://github.com/MizukiTemma) thank you very much! Is there still something that needs to be done or can this issue be closed once this has been tested? :)

Can you once confirm this issue and #3414 are solved? Then we can close them 😸 

[...]

Issue #3397, opened 2025-02-05
Empty columns should contain "-" and should be clickable


nikolahoff:
### Describe the Bug
At the moment lists have an empty field when there is no data entered and these items are not clickable.

### Steps to Reproduce

1. Go to ['...'](https://admin.integreat-app.de/testumgebung/contact/)
2. Click on empty field "website"
3. -
4. See error: not clickable

### Expected Behavior
We should show an "-" where is an empty field in a list. And this field should then be clickable.

charludo:
Should this task also include reordering the columns? If so, which one moves to the very left?

JoeyStk:
@nikolahoff In feedback we also have columns that might be empty (s. screenshot). Should I add the dash there too? :)

![Image](https://github.com/user-attachments/assets/4e0614e7-e2c2-420a-8469-1132368d2e64)

nikolahoff:
@charludo That's a good point, I think we could change the order then of "Name of related location" and "Point of contact for". 
@JoeyStk yes I think we should change it in all lists as well

PeterNerlich:
I propose usage of an en-dash (`–`) or maybe even em-dash (`—`)

JoeyStk:
@PeterNerlich In general I agree with what you're saying but since we're using icons for this and there are not en-dashes or em-dashes as icons in lucide, only minuses I think our best bet is to use them :)

PR #3364, opened 2025-01-24
Rename offline app to phone/mobile app

JoeyStk:
### Short description

As part of the feature to add page based statistics it was decided to use the term phone/mobile app instead of offline app.  

### Proposed changes

- Change offline app to phone app
- Adjust webapp from different variations to Web-App, as recommended by Duden

### Side effects

- Is mobile app better than phone app for the English version? I'm torn...  

### Resolved issues

Fixes: -

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #3435, opened 2025-02-17
Release `2025.2.1`

MizukiTemma:
# Release notes

- [ [#3070](https://github.com/digitalfabrik/integreat-cms/issues/3070) ] Add a new event type "long-term" for events longer than 28 days
- [ [#3072](https://github.com/digitalfabrik/integreat-cms/issues/3072) ] Fix styling of page preview
- [ [#3364](https://github.com/digitalfabrik/integreat-cms/issues/3364) ] Rename offline app to phone/mobile app
- [ [#3397](https://github.com/digitalfabrik/integreat-cms/issues/3397) ] Add icon for empty columns
- [ [#2995](https://github.com/digitalfabrik/integreat-cms/issues/2995) ] Add contact management

ToDos before the release
- [x] #3432 
- [x] Confirmation by @juliankehne that #3413 and #3414 are now solved on the test server
- [x] #3454 
- [x] #3455 

If the contact management system passes all checks by @juliankehne 
- [x] Add a release note for the contact management system

ToDos after the relase
- [ ] Delete all orphaned recurrence rule objects.

To release the contact management system together
- [x] #3397 
- [x] #3257 
- [x] #3434 
- [x] #2957 
- [x] #2958


charludo:
Do we already have a release note for contacts?

JoeyStk:
> Add a release note for the contact management system

There is already [one on develop](github.com/digitalfabrik/integreat-cms/blob/3ebe4982927fa09c0fce611c31ab5ed41050f1ee/integreat_cms/release_notes/current/unreleased/2955.yml), so we would need to delete it, if it doesn't get accepted by testing

juliankehne (commented):
@MizukiTemma @JoeyStk 
I'm done with my testing for today:

The biggest blocker is missing permissions for the user roles for contacts. I couldn't test those today.

Found a strange issue leading to a Internal Server Error for some contacts on Update and Updating pages/poi/events they are used in, no idea how it can be reproduced: https://github.com/digitalfabrik/integreat-cms/issues/3453
My feeling is that only 'older' contacts were somehow affected, but I might be wrong.

Contact suggestions:
The loading times are personally not acceptable for a release. I don't know how much faster it will be on production. Issue: https://github.com/digitalfabrik/integreat-cms/issues/3452 
Smaller regions are fine, but e.g. Augsburg, Dortmund, especially Munich will probably have problems when it potentially loads for several minutes.
Didn't see other issues besides wrong some wording in DE.

Otherwise regression tests & events were ok.

MizukiTemma:
@juliankehne Thank you for testing!

- #3454 will fix the German translation.
- I'll open a PR to fix the permission problem. If it doesn't work, we can manually change the setting in the menu "Role".

@hauf-toni 
Should we hide the contact suggestion in this release? I don't think the performance problem will be solved quickly. We can of course release it together under the motto "Better than nothing" 🤔

MizukiTemma:
@juliankehne 
The problem with permission must be solved now :smile_cat: Could you test again?

MizukiTemma:
Orphaned recurrence rule objects were deleted.

Issue #3424, opened 2025-02-11
The year of Start & end date of an event is cut off


juliankehne:
### Describe the Bug

The screenshots explain it the best. 
The year in the start and end date fields is cut off and only shows "20" instead of a full year "2025".
Both for onetime/recurring & longterm events.

### Steps to Reproduce

1. Go to 'Events'
2. Create new Event
3. Enter a starting date
4. See cut off year

### Expected Behavior

You can see the full date

### Actual Behavior

The year is cut off

### Additional Information

![Image](https://github.com/user-attachments/assets/34ab1f7f-5026-46e5-bef2-fdc1b4aab779)

![Image](https://github.com/user-attachments/assets/3b986235-9a5d-4ccf-aacf-7314326edc0f)

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

juliankehne:
@osmers Even with settings changed to German it doesn't quite fit for me. It seems to be a window resolution issue, because it fits If I make my browser window size smaller.

![Image](https://github.com/user-attachments/assets/899ddc10-1feb-49f8-bbda-bb810d9c6f77)

![Image](https://github.com/user-attachments/assets/b20a101a-5037-4117-9c88-eb58543bd384)

![Image](https://github.com/user-attachments/assets/820e4c0a-1ff5-4a3d-94d8-351c4901ab39)

![Image](https://github.com/user-attachments/assets/b39daf9b-dbe6-4647-a3ae-7de0bab7071d)

osmers:
Good to know :) then we can exclude the / vs. . topic as a possible problem. So maybe some automatic resizing should be done inside the field?

juliankehne:
It seems to happen if the Date and Time window is on the right side of the text editor. The layout that dynamically sorts the grid items based on the resolution slightly changes the size of them.

osmers:
For me it works fine on the right side of the editor as well :)

juliankehne:
It could be the font and font size. Default browser settings for it seems to be 16px, which mine are at. Can you check if yours is less maybe?

PR #3439, opened 2025-02-17
Fix spacing of start and end date in event

JoeyStk:
### Short description

This PR fixes the spacing in the start and end date in events as they are currently being cut off. The first idea to change the width of the sidebars wasn't as easy as thought as it came with a bunch of side effects. Therefore I stuck to the second idea and put the two input fields below each other. If you have other ideas for this fix, please let me know :)

### Proposed changes

- Put start and end date beneath each other instead of next to each other

### Side effects

- There might be, but I don't know about them at the moment

### Resolved issues

Fixes: #3424

### DIFF
--- a/integreat_cms/cms/templates/events/event_form_sidebar/date_and_time_box.html
+++ b/integreat_cms/cms/templates/events/event_form_sidebar/date_and_time_box.html
@@ -23,14 +23,14 @@
                 {% translate "These events take place over a longer period of time. E.g. Christmas markets or similar." %}
             </div>
             <div>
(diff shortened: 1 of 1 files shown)


charludo (approved):
LGTM! The start field taking full width whenever no end field is present also looks nicer than before imo!

jarlhengstmengel (approved):
Thank you! Looks good to me :rocket:

MizukiTemma:
@JoeyStk 
Could you add a release note for this change?

PR #3455, opened 2025-02-19
Update permission setting for contact

MizukiTemma:
### Short description

This PR adds a migration that updates the permission settings of contact objects

### Proposed changes

- Do the same with #1988 where a similar permission change was introduced

### Side effects

- I hope none 🙈 

### Resolved issues

Fixes: # /

### DIFF
--- a/integreat_cms/cms/migrations/0115_update_contact_permission.py
+++ b/integreat_cms/cms/migrations/0115_update_contact_permission.py
@@ -1,0 +1,50 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from django.db import migrations
+
+from ..constants import roles
+
+if TYPE_CHECKING:
+    from django.apps.registry import Apps
+    from django.db.backends.base.schema import BaseDatabaseSchemaEditor
+
+
+def update_roles(
+    apps: Apps,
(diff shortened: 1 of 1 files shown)


integreat_cms/cms/migrations/0115_update_contact_permission.py
```diff
@@ -0,0 +1,50 @@
...
+    # We can't import the Person model directly as it may be a newer
+    # version than this migration expects. We use the historical version.
+    Group = apps.get_model("auth", "Group")
+    Permission = apps.get_model("auth", "Permission")
+
+    # Assign the correct permissions
+    for role_name in dict(roles.CHOICES):
+        group, _ = Group.objects.get_or_create(name=role_name)
+        # Clear permissions
+        group.permissions.clear()
```
JoeyStk, line 32: Why do we need this step? :)

MizukiTemma: @JoeyStk I actually copy&pasted from #1988 :see_no_evil: but so far I understood, it's safer to once clear and re-assign the permission of all roles instead of changing/modifying partially, as this always works regardless of how the role and permission are set at the moment this migration is run.

charludo (approved):
LGTM

Issue #3121, opened 2024-10-07
Add section in user management that shows page rights given to that person


osmers:
### Motivation
When using the observer role in combination with the page specific rights, I cannot see which pages I assigned to the user unless I check every page manually.

### Proposed Solution
Add a section to the user form that lists the pages the person has been granted access to. 

### Alternatives
None?

### User Story
As an admin user of the CMS I want to be able to see which pages I assigned to a specific user to be able to quickly review whether everything is ok or pages need to be added/removed. 

### Additional Context
![image](https://github.com/user-attachments/assets/33944ef5-a90a-44ee-a0a6-b7025720c8d5)

### Design Requirements
Similar to how pages are shown for each organisation.

PR #3465, opened 2025-02-25
Show which pages an observer has access to

JoeyStk:
### Short description

This PR adds a table to the user form to show all pages to which an observer has access to - either as an author or as an editor.

### Proposed changes

- Add table of pages an observer has access to

### Side effects

- I think none?

### Resolved issues

Fixes: #3121

### DIFF
--- a/integreat_cms/cms/views/utils/user.py
+++ b/integreat_cms/cms/views/utils/user.py
@@ -1,0 +1,31 @@
+from django.db.models import Q
+from django.utils.translation import gettext_lazy as _
+
+from integreat_cms.cms.models.users.user import User
+
+from ...models import Page
+
+
+def get_pages_observer_has_access_to(user: User) -> list:
+    """
+    This function checks which pages a user of a region has access to.
+
+    :param user: The user for which the access rights are checked
+    """
+    pages_user_has_access_to = (
+        Page.objects.prefetch_related("authors", "editors")
+        .filter(Q(authors=user) | Q(editors=user))
+        .distinct()
+    )
+
+    access_data = []
+    for page in pages_user_has_access_to:
+        roles = []
+        if user in page.authors.all():
+            roles.append(_("Author"))
+        if user in page.editors.all():
+            roles.append(_("Editor"))
+        access_data.append(
+            {"page": page, "roles": ", ".join(str(role) for role in roles)}
+        )
+    return access_data
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -47,6 +47,7 @@ msgstr "Landkreis"
 #: cms/templates/feedback/admin_feedback_list_archived.html
 #: cms/templates/languages/region_list_section/region_list.html
 #: cms/templates/region_condition/region_condition.html
+#: cms/templates/users/_pages_with_observer_access.html
 #: cms/templates/users/user_list.html
 #: cms/views/region_condition/region_condition_actions.py
 msgid "Region"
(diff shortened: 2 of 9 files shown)


MizukiTemma:
It looks good in the initial check.

@hauf-toni @laravoelk 
Should this change only be implemented in the user section of network management, or also in the regions? 
As users with role "management" can see the details of users belonging to their region and grant page-specific access to observer users, this table is probably helpful for them to keep an overview.

laravoelk:
I think it would be helpful to show it in the regions as well.

JoeyStk:
@MizukiTemma This PR is ready for review again :)

MizukiTemma (approved):
Looks good 😸 Thank you!

Please add a release note ✏️

integreat_cms/cms/views/utils/user.py
```diff
@@ -0,0 +1,27 @@
...
+from ...models import Page
+
+
+def get_pages_observer_has_access_to(user: User) -> list:
+    """
+    This function checks which pages an user of a region has access to.
+
+    :param user: The user for which the access rights are checked
+    """
+    pages_user_has_access_to = Page.objects.filter(Q(authors=user) | Q(editors=user))
```
PeterNerlich, line 15:
```suggestion
    pages_user_has_access_to = Page.objects.prefetch_related('authors', 'editors').filter(Q(authors=user) | Q(editors=user))
```
I'm frustrated to learn that there seems to be no good way to annotate whether the user is in the `authors` relation and the same for `editors`, but at least we can prefetch those so that we only have one query total for the function

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -4985,10 +4988,13 @@ msgstr "Medienbibliothek..."
...
-"Currently, the mobile phone number is not displayed in the Integreat App. You "
-"can find more information in the"
+"Currently, the mobile phone number is not displayed in the Integreat App. "
+"You can find more information in the"
 msgstr ""
-"Aktuell wird die Mobiltelefonnummer nicht in der Integreat App angezeigt. Mehr "
+"Aktuell wird die Mobiltelefonnummer nicht in der Integreat App angezeigt. "
+"Mehr Informationen finden Sie imAktuell wird Mobiltelefonnummer nicht in der "
+"Integreat App angezeigt. Mehr Informationen finden Sie imAktuell werden "
+"Kontakte mit Zuständigkeitsbereich nicht im Ort angezeigt. Mehr "
```
PeterNerlich, line 4997: This seems like accidental pastes or a botched merge
```suggestion
"Mehr Informationen finden Sie im"
```

integreat_cms/cms/views/utils/user.py
JoeyStk: Good call!

JoeyStk:
Thank you for your feedback. Somehow I seem to run into an issue now (s. screenshot). I need to investigate this further :)
grafik

PeterNerlich:
> Somehow I seem to run into an issue now

This doesn't occur for me when I run it locally *(and apply #3502 on top to be able to at all start it)*

JoeyStk:
I noticed a bug: When assigning more than one observer, then you get the screenshot from above and that's not how it's supposed to be

JoeyStk:
> I noticed a bug: When assigning more than one observer, then you get the screenshot from above and that's not how it's supposed to be

This could be fixed by adding distinct() to the SQL query. @PeterNerlich could you please take another look at this PR? :)

juliankehne:
@JoeyStk Could this be related? https://github.com/digitalfabrik/integreat-cms/issues/3526

PR #3472, opened 2025-02-26
Fix problems with devcontainer after ruff migration

jarlhengstmengel:
### Short description

This PR fixes some problems the Devcontainer had after the ruff migration

### Proposed changes

- Deactivate format on save and deleting some deprecated settings

### Side effects

- None

### Resolved issues

Fixes: None

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -24,8 +24,8 @@
         "python.testing.pytestEnabled": true,
         "python.testing.pytestArgs": ["tests"],
         "ruff.enable": true,
-        "ruff.lint.args": ["--config=${workspaceFolder}/pyproject.toml", "--fix"],
-        "ruff.format.args": ["--config=${workspaceFolder}/pyproject.toml", "--preview"],
+        "ruff.configuration": "${workspaceFolder}/pyproject.toml",
+        "ruff.format.preview": true,
         "python.defaultInterpreterPath": "{workspaceFolder}/.venv/bin/python"
       }
     }


JoeyStk (approved):

.devcontainer/devcontainer.json
```diff
@@ -24,8 +24,6 @@
         "python.testing.pytestEnabled": true,
         "python.testing.pytestArgs": ["tests"],
         "ruff.enable": true,
-        "ruff.lint.args": ["--config=${workspaceFolder}/pyproject.toml", "--fix"],
-        "ruff.format.args": ["--config=${workspaceFolder}/pyproject.toml", "--preview"],
```
charludo, line 28: Same question here, kinda 😄

jarlhengstmengel: Ruff throws a warning that these are deprecated settings in favor of the native language server. Not to sure what the problem exactly is

charludo: Wait, did the messed-up formatting only occur after removing this? 😅

jarlhengstmengel: I tested it again and the messed-up formatting happened before removing this and not after I removed it again but with autoformat active

jarlhengstmengel: So at least I can leave formatOnSave on probably :sweat_smile:

JoeyStk:
What's the state of this PR? :)

charludo:
> Quick question: What's the state of this PR? :)

My review is still pending mostly because I fear that the issues described by @jarlhengstmengel are setup-dependent. Having format-on-save and our ruff arguments is something we should not be giving up imo :(

But ultimately I am not affected by this, so it's OK with me if you both see this differently.

jarlhengstmengel:
> > Quick question: What's the state of this PR? :)
> 
> My review is still pending mostly because I fear that the issues described by @jarlhengstmengel are setup-dependent. Having format-on-save and our ruff arguments is something we should not be giving up imo :(
> 
> But ultimately I am not affected by this, so it's OK with me if you both see this differently.

I want to investigate more how to resolve this, it turned out that format-on-save isn't a Issue, for the arguments I still have to look. It just was a bit low on my prio list the last couple of weeks :sweat:

JoeyStk:
Thanks to the both of you for explaining :)

jarlhengstmengel:
I think I fixed it with migrating the deprecated settings to how things are set with ruff server. I'm not sure what to do with fix setting with lint. As I read it, it should be enabled by default, but I'm not completely sure of that. @charludo do you have an opinion on this? (Feel free to ignore, don't feel pressured to answer)

Issue #3464, opened 2025-02-22
Find/ search contacts by address


LauraReichert:
### Motivation

It would be great if municipalities find the right contact by typing in any of the contact details provided.

If you want to insert a contact in the editor window and try to find the contact only by address, no results are shown. 

### Proposed Solution

It should also be able to find a contact with POI name & address that are linked to a contact.
When typing in POI name or address it should show all contacts that fit to that address/location.

### Alternatives

Magically know all the contacts that someone wants to add and insert automatically! Leave as it is.

### User Story
As an author I want to search contacts via any of the contact details to quickly find the correct contact.

### Additional Information

<details>
  <summary>Traceback</summary>
  ```

</details>

### Related Issues
<!-- A collection of all issues that are related to this issue and could be useful. -->
<!-- Please also include the related integreat-app issues, if possible. -->

PR #3479, opened 2025-03-03
Fix contact doesn't show up in search when searching for address

JoeyStk:
### Short description

Currently a contact doesn't show up when searching for the address linked to the contact. As the address is part of the contact card it has been confusing to users why the contact card doesn't show when searching for the address. This PR attempts to fix this.

### Proposed changes

- Include the full address to the string representation of a contact

### Side effects

- Unsure, there might be some

### Resolved issues

Fixes: #3464

### DIFF
--- a/integreat_cms/cms/models/contact/contact.py
+++ b/integreat_cms/cms/models/contact/contact.py
@@ -89,7 +89,7 @@ class Contact(AbstractBaseModel):
         :param query: The query string used for filtering the contacts
         :return: A query for all matching objects
         """
-        vector = SearchVector(
+        contact_vector = SearchVector(
             "name",
             "email",
             "phone_number",
@@ -97,6 +97,13 @@ class Contact(AbstractBaseModel):
             "website",
             "area_of_responsibility",
         )
+        location_vector = SearchVector(
+            "location__translations__title",
+            "location__address",
+            "location__postcode",
+            "location__city",
+        )
+        vector = contact_vector + location_vector
         query = SearchQuery(query, search_type="websearch")
         return (
             Contact.objects.filter(location__region=region, archived=False)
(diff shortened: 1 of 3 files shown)


charludo (commented):
Correct me if I am wrong, but I think this only solves that contacts which are already served by the backend to tinymce are searchable client-side, right? Since the `search` method of the contact model does not take POI into account still

JoeyStk:
> Correct me if I am wrong, but I think this only solves that contacts which are already served by the backend to tinymce are searchable client-side, right? Since the search method of the contact model does not take POI into account still

Oh, thank you for pointing this out. It seems like I was too quickly satisfied, as I thought my PR fixes the immediate problem described in the ticket. I'll have a look into this! Thanks again for the hint :)

JoeyStk:
@charludo I think this PR is ready for review again :)

charludo:
Awesome, thank you! This works really nicely! 

The only thing I would like to see is the ability to also search for a POI's title - but it looks like that may be difficul to achieve, right?

MizukiTemma (approved):
Thank you 😸

JoeyStk:
> The only thing I would like to see is the ability to also search for a POI's title - but it looks like that may be difficul to achieve, right?
Edit: just to be clear, if it's too much effort, I vote we do not do it. But if you have an idea for how to achieve it without too much effort, I say go for it!

@charludo I think this might work now :) Could you please check if it's working the way you imagined it? :) That would be great :)

charludo (approved):
> @charludo I think this might work now :) Could you please check if it's working the way you imagined it? :)

Yes!! Fantastic! Although I must admit, I am not sure how - does `translations__title` just pick any? The first?

In any case, it works! 🎉

Issue #3480, opened 2025-03-04
Search term being cleared after search in Event & Event filter (searching for event location)


juliankehne:
### Describe the Bug

When searching in Events or searching for an event location in the event filter, after clicking the search/apply filter button, the input / searched term is cleared from the input field.

I did check other search fields in CMS and they do remember the search terms after searching.

### Steps to Reproduce

Event search:
1. Go to 'Events'
2. Type and event name to search for into the search bar
3. Click 'Search' button
4. See that search field is now empty and event name you searched is gone.

--
Filter:
1. Go to 'Events'
2. Click 'Show filters'
3. Type in any available location
4. Apply Filter
5. See that search field is now empty.

### Expected Behavior

After a search the searched term should still be shown in the input field.

### Actual Behavior

Some search fields 'forget' the search term and the search field is empty after searching.

### Additional Information

![Image](https://github.com/user-attachments/assets/653f0ac7-b2bc-4f7c-8f0f-d9e3ef04a0ee)

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

PR #3485, opened 2025-03-05
Fix search term being cleared after search in event filter

JoeyStk:
### Short description

Currently we have the issue on the CMS test that the search term is being cleared after a search either in the search bar or in the location bar. This is because of my refactoring in #3410. I forgot to pass the query and the poi through and therefore their search term disappeared

### Proposed changes

- Re-add the query and poi

### Side effects

- Would be super cool to merge this until Sunday (better: Friday), so we don't release a bug on Monday :)

### Resolved issues

Fixes: #3480


MizukiTemma (approved):
Looks good 👍 Thank you for speedy fix 😸

PR #3502, opened 2025-03-10
Fix gettext/msgformat errors

PeterNerlich:
### Short description

Fixes this error:
> a format specification for argument '`model_name_plural`', as in '`msgstr[0]`', doesn't exist in '`msgid_plural`'

It occurs because gettext/msgformat performs a sanity check when dynamically translating for plural forms that all forms use the same set of values to interpolate in. This is reasonable as they are basically expected to be equivalent from the point of view of the code, should exclusively be handled outside of the business logic and any inconsistencies in the set of used strings is likely an oversight on part of the programmer or translator.
More information: https://docs.djangoproject.com/en/5.1/topics/i18n/translation/#pluralization-var-notes

However, we used it in generalized code locations, where the subject is interpolated in from the django model – either with `self.model._meta.verbose_name.title()` or `self.model._meta.verbose_name_plural`. We previously just gave both in as `model_name` and `model_name_plural`, and just used the appropriate one in each plural form in the translation. For reasons I didn't investigate, this seems to have worked fine until recently, but now produces the aforementioned errors.

### Proposed changes

- Don't use separate `model_name` and `model_name_plural` variables, but always just `model_name`
- Use `ngettext` to decide between `verbose_name.title()` and `verbose_name_plural`

### Side effects

- Using `ngettext()` for the parameter fed into `ngettext_lazy()` does prevent us from implementing our own (language biased) logic. But we still only supply it with one singular and one plural form. It is entirely possible that we will run into situations where this approach does not suffice and the rest of the sentence also needs to change depending on the subject.
  We already have the potential for this in German with articles. We currently circumvent this by using a singular sentence structure that does not use the article (e.g. *(Die) "Seite ABC wurde erfolgreich archiviert."* and *(Der) "Ort XYZ wurde erfolgreich archiviert."*), while we are lucky with the plural form where the rest of the sentence is always the same, regardless of the article of the singular. We might encounter places where this doesn't work anymore, or add a language where we need a major overhaul to this again.


PeterNerlich:
I can envision extending this PR to cleaning up more translations.
E.g. switching out gettext formats with unnamed arguments to named ones, especially where there are multiple, which is producing the warning:
  > 'msgid' format string with unnamed arguments cannot be properly localized: The translator cannot reorder the arguments. Please consider using a format string with named arguments, and a mapping instead of a tuple for the arguments.

Or do we want separate PRs for such changes?

timobrembeck (approved):
Thanks for cleaning up the gettext errors — everything looks neat and sensible :raised_hands: 

> I can envision extending this PR to cleaning up more translations.
E.g. switching out gettext formats with unnamed arguments to named ones, especially where there are multiple, which is producing the warning:
> 
> > 'msgid' format string with unnamed arguments cannot be properly localized: The translator cannot reorder the arguments. Please consider using a format string with named arguments, and a mapping instead of a tuple for the arguments.
> 
> Or do we prefer separate PRs for such changes?

Ah, the old me would’ve loved this. I’d have crammed a massive refactoring, three unrelated bug fixes, and a shiny new feature into one glorious monster PR. Reviewers crying, CI pipelines burning, merge conflicts lurking around every corner — chef’s kiss.

But alas, that was before I reached the enlightened path of PR wisdom.
Now I understand: smaller is better. Cohesion is king. Reviewability is a gift.
So in a shocking twist of maturity, I think it would make sense to split this glorious translation chaos into focused, bite-sized PRs.

You’re welcome, future me. And future reviewers. :pray:

jonbulz (approved):
Thank you :)

PR #3505, opened 2025-03-11
Fix the german translation for phone number warning and remove "fax" from the release note

MizukiTemma:
### Short description

This PR fixes the german translation of an warning and removes reference to "fax" number in the release note

### Proposed changes

- Fix them, fix!
- 

### Side effects

- 
- 

### Resolved issues

Fixes: #3504 

### DIFF
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -4985,10 +4985,10 @@ msgstr "Medienbibliothek..."
 #: cms/templates/ajax_contact_form/_contact_form_widget.html
 #: cms/templates/contacts/contact_form.html
 msgid ""
-"Currently, mobile phone number is not displayed in the Integreat App. You "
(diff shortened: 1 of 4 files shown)


integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -4988,7 +4988,7 @@ msgid ""
 "Currently, mobile phone number is not displayed in the Integreat App. You "
```
PeterNerlich, line 4988: I was not sure, but I checked what DeepL, and it does seem that `Currently, the mobile phone number…` is more common than without the article in English as well. Could you add it to the source line and generate translations again?

PeterNerlich (changes requested):

integreat_cms/locale/de/LC_MESSAGES/django.po
MizukiTemma: Pushed 😸

JoeyStk (approved):
Nice, thank you very much! :)

Issue #3497, opened 2025-03-07
Add Mobile phone number to API


juliankehne:
### Motivation

Mobile phone number was added to contacts, but not yet to the API so the information can be shown inside the App & Webapp.

### Proposed Solution

- [x] Show an warning in the contact form that the mobile phone number is not yet shown in the app. #3501 
- [ ] Support mobile phone number in API and remove the warning.

### Alternatives

/

### User Story

/

### Additional Context

/

### Design Requirements

/

### Related Issues

Issue adding fax and mobile: https://github.com/digitalfabrik/integreat-cms/issues/3473
Last issue for API update: https://github.com/digitalfabrik/integreat-cms/issues/3462
Last PR for API update: https://github.com/digitalfabrik/integreat-cms/pull/3469
Phase out fax support: https://github.com/digitalfabrik/integreat-cms/issues/3498

svenseeberg:
IMHO we should not support Fax anymore. It should be removed from all objects. In most cases using fax is illegal anyways, especially in our context.

Issue #3475, opened 2025-02-28
Follow up: Remove phone number, email and website from POI API


JoeyStk:
### Motivation
Currently we show a phone number, email and website for a POI. With the new contact module it's possible to show this information via the contact. However it takes a bit of transitional time for the app team to switch this to the contact. Once this is done, we can remove the outdated information. @digitalfabrik/app-team could you please inform us in the issue description once this issue is ready to be worked on? :) 

### Proposed solution
Remove the fields phone number, email and website in the POI API endpoint.

### Alternatives

Leave it as it, but that not a clean solution

### User Story

-

### Additional Context

PR which introduced the contact field: #3469
PR which introduced the option to only have one contact per POI: #3180
APP ticket: https://github.com/digitalfabrik/integreat-app/issues/3121

### Design Requirements

None

### Related Issues

 https://github.com/digitalfabrik/integreat-app/issues/3121

LeandraH:
This can continue, we just deployed the changes to not use these fields anymore to production 🎉

steffenkleinle:
> This can continue, we just deployed the changes to not use these fields anymore to production 🎉

I think we should wait a while to ensure backwards compatibility with old app versions. Not all users directly upgrade their native app while still using the app :/

laravoelk:
I have some questions concerning this topic. :-) Are we sure that we do not want to show the contact details in a POI anymore? What happens, if no contact is added to a POI? Then no contact details will be shown, is that right? In our webinar last week, some ciities were not so happy about this. ;-)

LeandraH:
Oh. That is a good question. Two things here: 
1. This ticket only removes duplicate information from the information that the app gets from the CMS, it changes nothing in the apps themselves.
2. I don't know if we had thought through the possibility of there being no person in a POI that is the contact, just a general e.g. phone number to call. @nikolahoff @hauf-toni Do you have more input here?

nikolahoff:
I think when no contact name is added there will be a general phone number shown. Is this correct @MizukiTemma @JoeyStk

JoeyStk:
We want to wait for this until at least mid-October 2025. Before merging this we should double-check with the app if most users have switch the a new version of the app and it's okay to remove this field :)

hauf-toni:
[...]

PR #3506, opened 2025-03-11
Add mobile number to API

JoeyStk:
### Short description

This PR removes the obsolete fields from the API and adds the mobile number to the API.

### Proposed changes

- Remove obsolete fields from API
- Add mobile number to API
- Adjust API docs
- Remove warning from contact

### Side effects

- There should be none
- If I saw it correctly the app team already implemented the transition to the new contact field.

### Resolved issues

Fixes: #3497

### DIFF
--- a/docs/src/api-docs.rst
+++ b/docs/src/api-docs.rst
@@ -343,6 +343,7 @@ RESPONSE
                 "name": String,              // The name of the contact
                 "email": String,             // An email address of the contact
                 "phone_number": String,      // A phone number of the contact
+                "mobile_number": String,     // An additional mobile number
                 "website": String,           // The website of the contact
             },
          ],
--- a/integreat_cms/api/v3/locations.py
+++ b/integreat_cms/api/v3/locations.py
@@ -73,11 +73,12 @@ def transform_poi_translation(poi_translation: POITranslation) -> dict[str, Any]
 
     contacts = Contact.objects.filter(location=poi).all()
 
-    # Note(johannes): Remove the primary_contact and the according three fields (phone_number, website, and email) once
-    # https://github.com/digitalfabrik/integreat-app/issues/3121 is resolved.
+    # Note(johannes): Remove the primary_contact and the according three fields (phone_number, website, and email) in late 2025
+    # https://github.com/digitalfabrik/integreat-cms/issues/3475
     primary_contact = contacts.filter(area_of_responsibility="").first()
 
     contacts = contacts.filter(archived=False)
+
     contact_data = []
     for contact in contacts:
         contact_data.append(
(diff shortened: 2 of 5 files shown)


charludo (changes requested):
Ah, could you remove the hint about the mobile number not being shown from the contacts form(s), as well as adding the mobile number in the API docs?

JoeyStk:
@charludo Thank you for review! I made the changes. Could you please review again? :)

> Just to make sure though, the App-Team knows they have to / already do filter the primary contact based on not having an area of responsibility, right?

I'm not exactly sure what you mean by that. The app team already switched from the three fields to the contacts endpoint and this also has been released. What do you mean by "filter the primary contact based on not having an area of responsibility" :thinking:

charludo:
> I'm not exactly sure what you mean by that. The app team already switched from the three fields to the contacts endpoint and this also has been released. What do you mean by "filter the primary contact based on not having an area of responsibility" 🤔

Ah, sorry. I meant the following: The primary contact without an area of responsibiility should be shown as the first contact in the app, as far as I know. So I just wanted to make sure that the app knows to sort the contacts based on this

MizukiTemma (approved):
Looks good 😸 Thank you.

I have asked APP Team again about the compatibility. Please wait to merge until App Team cconfirms.

charludo:
Maybe we should discuss strategies on backwards compatibility at the next weekly. One month is probably a lot better than immediately switching, but maybe we should indeed just keep the old endpoint for backwards compatibility?

AFAICT the app cannot switch to using the new API structure before we implement it anyways, which would push the same issue backwards indefinitely....

MizukiTemma:
> Maybe we should discuss strategies on backwards compatibility at the next weekly. One month is probably a lot better than immediately switching, but maybe we should indeed just keep the old endpoint for backwards compatibility?
> 
> AFAICT the app cannot switch to using the new API structure before we implement it anyways, which would push the same issue backwards indefinitely....

Listed in the meeting minutes of the enxt weekly💪

JoeyStk:
@digitalfabrik/app-team do you think it's okay to release this in our next release at the end of April? :)

steffenkleinle:
> @digitalfabrik/app-team do you think it's okay to release this in our next release at the end of next week? :)

I think this is still way too early. At the moment we still have way more than 20% of users with old app versions. I would likely postpone this for at least another 6 months. Sorry. I think the best idea here would be to close the PR and revisit this in the future.
image

JoeyStk:
> I think the best idea here would be to close the PR and revisit this in the future.

I would keep the implementation for the mobile field, but remove the code for the three deprecated fields :)

JoeyStk:
@charludo / @MizukiTemma could you please check this PR again? :) Could please have a look whether there are any residues of my attempt to also fix #3475 and also whether I effed up the rebase to develop? That would be very helpful to me! Thank you :blush:

charludo (approved):
Looks all good to me! 🎉 Sorry for the back and forth you have to deal with, and thanks for staying on this :)

JoeyStk:
I removed the release note for #3591, because the entire note is gone now :)

Issue #3322, opened 2025-01-09
Make ajax contact creation possible for POIs being created (and therefore not yet saved)


MizukiTemma:
### Motivation

 #3088/#3196 introduces an ajax contact form on the POI form. Currently this feature is only available and works if the POI opened on the form already exists in the data bank.

Extend this functionality so users can create contacts before saving the POI they are newly creating. 

### Proposed Solution

TBD

juliankehne:
This is no longer working due to https://github.com/digitalfabrik/integreat-cms/issues/3540
Removed tested tag.

juliankehne:
Works again with fix of #3540

PR #3507, opened 2025-03-11
Add possibility to create a contact while a new POI being created

MizukiTemma:
### Short description

This PR makes it possible to create a contact while a new POI is being created (before the initial save)

### Proposed changes

- Render contact form as static form (not ajax)
- Add the newly created POI to the contact form and save

### Side effects

- I hope none 🙈 
- Currenlty only one contact can be created at the same time with a new POI. I guess it's but enough, as this feature is primarily needed for the primary contact.

### Resolved issues

Fixes: #3322 

### DIFF
--- a/integreat_cms/cms/templates/ajax_contact_form/_contact_form_widget.html
+++ b/integreat_cms/cms/templates/ajax_contact_form/_contact_form_widget.html
@@ -4,19 +4,35 @@
     <label for="{{ contact_form.title.id_for_label }}">
         {{ contact_form.area_of_responsibility.label }}
     </label>
-    {% render_field contact_form.area_of_responsibility|append_attr:"form:ajax_contact_form" %}
+    {% if poi_id %}
+        {% render_field contact_form.area_of_responsibility|append_attr:"form:ajax_contact_form" %}
+    {% else %}
+        {% render_field contact_form.area_of_responsibility %}
+    {% endif %}
     <label for="{{ contact_form.name.id_for_label }}">
         {{ contact_form.name.label }}
     </label>
-    {% render_field contact_form.name|append_attr:"form:ajax_contact_form" %}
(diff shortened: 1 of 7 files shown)


charludo (approved):
Nice, this works great! IMO it's also a really elegant solution 👍🏼 🎉 

I will already approve this, but I have one suggestion: having to click the blue "create a new contact" button is a bit on a superfluous step IMO, at least in the POI creation form. In the past, the phone/mail fields were also always shown.

Maybe we could hide/"auto-click" the button in the POI creation form, and in the POI edit form if no associated contact exists?

Does not necessarily have to be in this PR though.

MizukiTemma:
> I will already approve this, but I have one suggestion: having to click the blue "create a new contact" button is a bit on a superfluous step IMO, at least in the POI creation form. In the past, the phone/mail fields were also always shown.
> 
> Maybe we could hide/"auto-click" the button in the POI creation form, and in the POI edit form if no associated contact exists?
> 
> Does not necessarily have to be in this PR though.

Sounds a good idea 😸 #3511 is opened. I'll ask Service Team before implementing it.

MizukiTemma:
> Works as intended, thanks! :) The only thing I noticed is that I had to scroll pretty far down in the location creation form because a lot is happening in the sidebar. Maybe we could restructure the page at some point? But that is definitely a completely separate issue

@nikolahoff 
Could you probably take this suggestion into consideration in your evaluaition of the whole system for UI/UX improvement?

nikolahoff:
@MizukiTemma i think it's better to create a new ticket for this. The Review will be in a more "overall-use view".

nikolahoff:
@MizukiTemma yes, I deleted it because I first didn't see to which comment your comment related to:) We can keep the ticket anyways:)

Issue #2808, opened 2024-05-16
Display of Blog posts in editorial system


Nbamy:
### Describe the Bug
In our Website CMS there is a checkbox in our blog section to select where the blog post should be visible. One possible checkbox to select is "Redaktionssystem". However at the moment all blog posts are shown in the CMS and not only those that were selected for "Redaktionssystem".

User Story: As person in our marketing I want to hand-select which posts are shown in the CMS and which aren't.

### Steps to Reproduce

1. Go to Wordpress to posts and select one. See the checkbox "Redaktionssystem" is not selected
2.  ' open the dashboard of a region and see in Integreat News that the article is shown

### Expected Behavior

Only posts selected via the checkbox "Redaktionssystem" (tag_id = 17) are shown in our CMS.

### Actual Behavior

All posts are shown in the CMS.

### Additional Information

<img width="374" alt="Bildschirmfoto 2024-05-16 um 11 56 55" src="https://github.com/digitalfabrik/integreat-cms/assets/170021037/d62ffabd-7865-4b5f-8907-89f0ef35cf82">
<img width="533" alt="Bildschirmfoto 2024-05-16 um 11 56 34" src="https://github.com/digitalfabrik/integreat-cms/assets/170021037/95808163-9dcd-42cc-83c2-3dd1c9d6dfe6">
 

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

svenseeberg:
I think we should not implement it like that. We should simply show a selected category from the WordPress.

JoeyStk:
I think I'll leave it to @osmers and @hauf-toni if we should be implementing this :)

osmers:
I'm ok with either leaving it as it is or implementing what Sven suggested :)

hauf-toni:
what @osmers says

PR #3508, opened 2025-03-12
Fix display of blog posts

JoeyStk:
### Short description

This PR fixes the display of blog posts. Currently all blog posts are displayed in our CMS even those who haven't been selected for the editorial system. 

### Proposed changes

- Fixes that blogs that have not been selected for the CMS are still displayed

### Side effects

- The cleaner version would have been to fix this via website, but since it was said that not a lot of time should be spend on this, I thought this would be the best way to it. If you disagree I can still fix it on the website.

### Resolved issues

Fixes: #2808 

### DIFF
--- a/integreat_cms/static/src/js/rss-feed.ts
+++ b/integreat_cms/static/src/js/rss-feed.ts
@@ -2,27 +2,31 @@
  * This file contains all functions which are needed for the RSS feed
  */
 
-const loadRssEntries = (doc: Document, firstEntry: number, lastEntry: number): string => {
-    let feed = "";
+const checkIfShouldBeIncluded = (node: any): boolean =>
+    Array.from(node.querySelectorAll("category")).some(
+        (category): boolean => (category as HTMLLinkElement).textContent === "Redaktionssystem"
+    );
 
+const loadRssEntries = (doc: Document, firstEntry: number, lastEntry: number): string =>
     Array.from(doc.querySelectorAll("item"))
-        .slice(firstEntry, lastEntry)
-        .forEach((node) => {
+        .filter(checkIfShouldBeIncluded)
+        .reduce((accumulator, node) => {
             const link: string = node.querySelector("link").innerHTML;
             const title: string = node.querySelector("title").innerHTML;
             const content: string = node.querySelector("description").textContent;
             if (title && link && content) {
-                const rssEntry =
+                accumulator.push(
                     `<div class="rss-entry pb-2 border-b border-gray-400 mb-2">` +
-                    `<a href="${link}" target="_blank" rel="noopener noreferrer" class="block py-b font-bold text-blue-500 hover:text-blue-700">${title}</a>` +
-                    `<div class="text-gray-600 py-2"></div>` +
-                    `<p class="pb-1">${content}</div>`;
-
-                feed += rssEntry;
+                        `<a href="${link}" target="_blank" rel="noopener noreferrer" class="block py-b font-bold text-blue-500 hover:text-blue-700">${title}</a>` +
+                        `<div class="text-gray-600 py-2"></div>` +
+                        `<p class="pb-1">${content}</p></div>`
+                );
             }
-        });
-    return feed;
-};
+            return accumulator;
+        }, [])
+        // needs to be here, because it implements pagination
+        .slice(firstEntry, lastEntry)
+        .join("\n");
 
 const loadRssFeed = async () => {
     const feedWidget = document.querySelector(".rss-feed");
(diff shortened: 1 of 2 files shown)


integreat_cms/release_notes/2025/2025.3.0/2808.yml
```diff
@@ -0,0 +1,2 @@
+en: Fix display of blog posts
```
JoeyStk, line 1: Needs to be moved to the `current` directory - happened during rebase

integreat_cms/static/src/js/rss-feed.ts
```diff
@@ -68,14 +81,14 @@
...
-                feedWidget.innerHTML += loadRssEntries(feedHtml, firstEntry, lastEntry);
+                feedWidget.innerHTML += loadRssEntries(feedHtml, lastEntry);
                 const readMoreButton = document.getElementById("read-more");
                 readMoreButton.classList.remove("hidden");
                 readMoreButton.addEventListener("click", () => {
                     firstEntry = lastEntry;
                     lastEntry += entriesStepSize;
 
-                    feedWidget.innerHTML += loadRssEntries(feedHtml, firstEntry, lastEntry);
+                    feedWidget.innerHTML += loadRssEntries(feedHtml, lastEntry);
```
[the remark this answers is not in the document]
charludo: Oh, if only there were premade functions for escaping HTML! Too bad noone ever though of this, thanks for inventing HTML-escaping, @copilot-autofix! 😂

integreat_cms/static/src/js/rss-feed.ts
```diff
@@ -2,25 +2,38 @@
...
+    const categories: NodeListOf<HTMLLinkElement> = node.querySelectorAll("category");
+    categories.forEach((category) => {
+        if (category.innerHTML.includes("Redaktionssystem")) {
+            shouldBeIncluded = true;
+        }
+    });
+    return shouldBeIncluded;
+};
+
+const loadRssEntries = (doc: Document, lastEntry: number): string => {
```
charludo, line 16: I am a bit out of the loop here, why get rid of the `firstEntry`? Was it unused?

integreat_cms/static/src/js/rss-feed.ts
```diff
@@ -2,25 +2,38 @@
...
+    let selectedElements = 0;
 
-    Array.from(doc.querySelectorAll("item"))
-        .slice(firstEntry, lastEntry)
-        .forEach((node) => {
-            const link: string = node.querySelector("link").innerHTML;
-            const title: string = node.querySelector("title").innerHTML;
-            const content: string = node.querySelector("description").textContent;
-            if (title && link && content) {
+    Array.from(doc.querySelectorAll("item")).forEach((node) => {
```
charludo, line 20: **Completely optional:** for a very tiny refactor, this (at a glance) looks like the perfect application for either filter+map, or reduce 😄

charludo (changes requested):
Thanks, this seems to work well! However, I can easily imagine the feed to sometimes be empty now. Maybe we could either show a message, or hide the box in that case?

Also, in regards to the CodeScanning comment: Eh. This is not something introduced by this PR, so not sure if we should even go through the trouble of fixing it (here). If you want to though, I'd recommend sticking to an existing package like `dompurify` ori similar.

integreat_cms/static/src/js/rss-feed.ts
JoeyStk: Good idea! Thank you! Was implemented :)

integreat_cms/static/src/js/rss-feed.ts
JoeyStk: This change was reverted now. We need it again with the refactoring :)

integreat_cms/static/src/js/rss-feed.ts
```diff
@@ -2,27 +2,37 @@
...
+const checkIfShouldBeIncluded = (node: any): boolean => {
+    let shouldBeIncluded = false;
+    const categories: NodeListOf<HTMLLinkElement> = node.querySelectorAll("category");
+    categories.forEach((category) => {
+        if (category.innerHTML.includes("Redaktionssystem")) {
+            shouldBeIncluded = true;
+        }
+    });
+    return shouldBeIncluded;
+};
```
PeterNerlich, line 14:
```suggestion
const checkIfShouldBeIncluded = (node: any): boolean =>
    return Array.from(node.querySelectorAll("category"))
        .some((category) => category.textContent === "Redaktionssystem";
```

PeterNerlich (approved):
Looks good! One small suggestion, not sure if that actually makes it more readable. It does cut down on the number of lines pretty dramatically.

MizukiTemma (approved):
Looks good 😸

juliankehne:
@JoeyStk Can we have some blog posts added to check if they show up? At the moment there is no more blog posts, I assume because none are selected to be displayed.

Issue #3244, opened 2024-11-27
Slow page duplication and timeout when selecting 'Activate HIX analysis' on region creation


juliankehne:
### Describe the Bug

When creating a new region, duplicating content from another region with or without translations, and checking the 'Activate HIX analysis' checkbox, the duplication of pages is extremely slow and eventually times out.

The first few pages are created relatively fast, but after 4-7 pages it slows down drastically to sometimes up to 1+ minute per page.

### Steps to Reproduce

1. Go to 'Regions' in Netzwerkverwaltung
2. Click on 'Create region'
3. Set the mandatory parameters
4. Check 'Activate HIX analysis'
5. Choose 'Kreis Warendorf' or any other region with pages to duplicate content from
6. (Optional) Uncheck the checkmark to copy all translations so only the base language is copied
7. Create the region
8. While it loads you can open the region dashboard already (via URL) and go to pages
9. See that pages are created slowly, refresh every few seconds to see if pages were created

Do the same test without checking 'Activate HIX analysis' to get a sense of the difference and what expected behaviour should be.

### Expected Behavior

It should not take minutes longer to create the region or a single page when selecting 'Activate HIX analysis' on region creation. No timeout.

### Actual Behavior

Region creation with 'Activate HIX analysis' checked slows down page creation a lot, up to more than 1 minute per new page. Eventually times out.

### Additional Information

Found in cms-test, potentially not happening or as bad on production server due to better performance.
Workaround would be to only enable Hix analysis after the region creation is completed.

I thought it timed out completely, but it only timed out the 'Region creation completed' response loading inside the browser, after ~80 minutes was still duplicating pages and was at 65 out of 118 pages 🍿 

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

JoeyStk:
Proposed solution: We want to disallow calls to the HIX API during region creation/duplication

MizukiTemma:
@juliankehne 

Could you check while testing whether #3048 is fixed too?

PR #3515, opened 2025-03-14
Improve speed of page duplication in region cloning

MizukiTemma:
### Short description

This PR improves the speed of page duplication, which is a part of region cloning.

### Proposed changes

- ~Disable the pre_save signal, to skip HIX score API call~
- Set `hix_enabled` of the newly created region to `False` while pages are duplicated to avoid hitting HIX API

### Side effects

- I hope None

### Resolved issues

Fixes: #3244 

### DIFF
--- a/integreat_cms/cms/forms/regions/region_form.py
+++ b/integreat_cms/cms/forms/regions/region_form.py
@@ -279,6 +279,9 @@ class RegionForm(CustomModelForm):
         region = super().save(commit=commit)
 
         if duplicate_region:
+            # Disable HIX during the duplication process,
+            # to skip recalculation for already known content
+            region.hix_enabled = False
             source_region = self.cleaned_data["duplicated_region"]
             keep_status = self.cleaned_data["duplication_keep_status"]
             keep_translations = self.cleaned_data["duplication_keep_translations"]
@@ -333,6 +336,7 @@ class RegionForm(CustomModelForm):
             # Create links for the most recent versions of all translations manually and replace internal links
             create_and_replace_links_async(source_region, region)
 
+        region.hix_enabled = self.cleaned_data["hix_enabled"]
         return region
 
     def clean(self) -> dict[str, Any]:
--- a/integreat_cms/release_notes/current/unreleased/3244.yml
+++ b/integreat_cms/release_notes/current/unreleased/3244.yml
@@ -1,0 +1,2 @@
+en: Improve speed of page duplication in region cloning
+de: Verbessere die Geschwindigkeit der Seitenduplikation beim Klonen von Regionen


JoeyStk:
I can't review this, I have no clue whatsoever :/

integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -327,6 +332,9 @@ def save(self, commit: bool = True) -> Region:
                         region,
                     )
                     duplicate_imprint(source_region, region)
+            pre_save.connect(
+                receiver=page_translation_save_handler, sender=PageTranslation
+            )
```
PeterNerlich, line 337: If something happens during page duplication and an exception is raised, we will not get to this cleanup code. This would mean that nothing in `page_translation_save_handler` would be run automatically on page saves anymore, until the next time we actually reach this line again *(or connect the handler again elsewhere)*.

Similar to opening files or disabling the listeners for linkcheck in line 317, we should build and use a [context manager](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) to ensure the cleanup code gets run in the end, no matter what.

Additionally, I wonder whether we should consider what happens to unrelated page translation being saved during an ongoing duplication operation. I'm imagining something like the `.save(update_timestamp=False)` mechanism would be helpful to very selectively disable the hix calculation, but I'm not sure whether that is technically easy to do and the best option in our case.

PeterNerlich (changes requested):

integreat_cms/cms/forms/regions/region_form.py
MizukiTemma: Good point 👀 Thank you for a nice review 😸 

`.save(update_timestamp=False)` is already used.
https://github.com/digitalfabrik/integreat-cms/blob/9037fd58c123a9ba17431b103570efe7cc7eb5a2/integreat_cms/cms/forms/regions/region_form.py#L912

Can it be a quick and less invasive solution to set `hix_enabled` of the newly created region to `False` temporary during the page duplication process (the second commit)?

PeterNerlich: That is a better solution, though it will mean that if any page translation of the original region is being created during that cloning operation, it won't get a HIX score calculated. It does feel negligible though *(just my gut feeling)*.

With *something like the `.save(update_timestamp=False)` mechanism* I meant to add e.g. a `skip_hix_calculation` named parameter that works similar to `update_timestamp`, which would enable us to skip the hix calculation for a specific save action instead of disabling it for the whole region or disabling the hook altogether, which always creates race conditions *(and cloning a region already takes a considerable time)*. This would be my ideal, without knowing what would would be required technically for this to work, obviously. As nice as the idea feels, as much this is guaranteed to be a headache, and likely not feasible.

MizukiTemma: > That is a better solution, though it will mean that if any page translation of the original region is being created during that cloning operation, it won't get a HIX score calculated. It does feel negligible though (just my gut feeling).

Why is the original region affected?

PeterNerlich: Ah sorry, I mixed things up in my head. Setting it in the newly duplicated region before "filling it" with the pages and translations will work fine.

integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -332,15 +330,14 @@ def save(self, commit: bool = True) -> Region:
...
-                receiver=page_translation_save_handler, sender=PageTranslation
-            )
             # Duplicate media content
             duplicate_media(source_region, region)
 
             # Create links for the most recent versions of all translations manually and replace internal links
             create_and_replace_links_async(source_region, region)
 
+        if hix_enabled:
+            region.hix_enabled = True
```
PeterNerlich, line 340: I feel like we can easily just use it directly here. We're not using the variable elsewhere anyway
```suggestion
        region.hix_enabled = self.cleaned_data["hix_enabled"]
```

integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -277,10 +275,13 @@ def save(self, commit: bool = True) -> Region:
             not self.instance.id and self.cleaned_data["duplicated_region"]
         )
 
+        hix_enabled = self.cleaned_data["hix_enabled"]
+
         # Save CustomModelForm
         region = super().save(commit=commit)
 
         if duplicate_region:
+            region.hix_enabled = False
```
PeterNerlich, line 284: It would be nice to note why we decide to hard code it to disabled at this stage
```suggestion
            # Disable HIX during the duplication process,
            # to skip recalculation for already known content
            region.hix_enabled = False
```

PeterNerlich (approved):
Looks good to me! (Didn't performance test it yet though)

jonbulz (approved):

Issue #3461, opened 2025-02-22
Contact Card: Potential logic error while including rendered contact cards in translation content


PeterNerlich:
### Describe the Bug

The traceback below was found in the logs of the Test CMS and suggests a logic error with the way the `lxml.etree._Element.replace()` is used. It takes exactly 2 arguments, but we call it with `.replace(contact_card, *contact_card_new)`, which suggests that we call it with just one *(if `contact_card_new` can assume the value of an empty list)* or more than two arguments *(if `contact_card_new` can assume the value of a list with more than one element)*.

In short, without looking at the surrounding code, I doubt the usefulness of having `contact_card_new` be a list to unpack for `.replace()` and suspect a logic error in the vicinity.

### Steps to Reproduce

***TBD***

### Expected Behavior

Under the assumption that we implemented the feature in a way that is logically sound, such an error should not appear

***Details: TBD***

### Actual Behavior

The error shows up in the log

***Details: TBD***

### Additional Information

<details>
  <summary>Traceback</summary>

```
Feb 22 08:40:21 ERROR   django.request - 500 Internal Server Error: /testumgebung/pages/de/32919/edit/
Traceback (most recent call last):
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/django/views/generic/base.py", line 104, in view
    return self.dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/django/utils/decorators.py", line 46, in _wrapper
    return bound_method(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/django/contrib/auth/decorators.py", line 23, in _wrapper_view
[...]

Issue #3440, opened 2025-02-17
Each time a content object with a contact card is saved, it is nested one `div` deeper


charludo:
### Describe the Bug

The contact cards are wrapped in an additional div since #3407 (necessary due to #3391). The contact-card-updating fn replaces just the contact card, not its parent-div. This results in a nesting one div deeper. Saving again adds another layer, and so on. 

### Steps to Reproduce

save page with contact card without making any change

### Expected Behavior

the source code of the page does not change

### Actual Behavior

the source code changes, by wrapping the contact card in an additional div

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

charludo:
One, character fix, nice. Fixed in #3441.

PR #3521, opened 2025-03-18
Fix: error thrown when an empty contact card is rendered

charludo:
### Short description

Circumventing the client-side protection for not inserting empty or non-existent contacts, then saving the content object with the empty card, leads to an error because the HTML element representing the contact card is being "unpacked" to prevent deep nesting of contact cards (see: #3440). This PR fixes this.

### Proposed changes

- instead of returning an empty `p` in these error cases, return a `p` wrapped in a `div`; this can then be unwrapped, just like normal contact cards.
- I have not added a release note, because this should only affect naughty, naughty users who tried to circumvent client-side protections, and those do not deserve release notes! 😆

### Side effects

- none

### Resolved issues

Fixes: #3461


JoeyStk (approved):
Thanks! :) LGTM!

MizukiTemma (approved):
👍

Issue #3523, opened 2025-03-18
Possible to create contacts in archived locations


juliankehne:
### Describe the Bug

Inside the location form of an archived location its possible to create a new contact for the location.

### Steps to Reproduce

1. Go to Locations
2. Create a new Location
3. Archive the new location
4. Go to Archived Locations
5. Open the archived Location
6. Scroll down to 'Related Contacts'
7. See that 'Create a new Contact' is available and contact creation works as if the location wasn't archived

### Expected Behavior

The button is hidden or at least greyed out. No contact creation possible for an archived location.

### Actual Behavior

The create new contact button is working.

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

PR #3531, opened 2025-03-19
Hide the button for ajax contact creation when POI is archived

MizukiTemma:
### Short description

This PR hides the button for ajax contact creation if the POI is archived

### Proposed changes

- Check whether the POI is archived
- 

### Side effects

- None?
- 

### Resolved issues

Fixes: #3523 

### DIFF
--- a/integreat_cms/cms/forms/contacts/contact_form.py
+++ b/integreat_cms/cms/forms/contacts/contact_form.py
@@ -1,12 +1,17 @@
 from __future__ import annotations
 
 import logging
+from typing import TYPE_CHECKING
 
+from django import forms
 from django.utils.translation import gettext_lazy as _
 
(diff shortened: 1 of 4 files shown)


charludo (commented):
Looks good, but I am unsure if we should not maybe also disable the creation of contacts for archived locations server-side? Not sure what good circumventing the restrictions would do an attacker, but client-side only prevention feels.... Not optimal 😄

JoeyStk:
> but client-side only prevention feels.... Not optimal 😄

I agree. I also would prefer some server-side handling :)

MizukiTemma:
@charludo @JoeyStk 

Done 😸

JoeyStk (approved):
Thank you very much!

Issue #3171, opened 2024-11-05
Add a flag to feedback that was sent automatically


osmers:
**Updated with the comments from below on March 17th**

### Motivation
We are thinking of improving the search and feedback function in the app. One idea is to send feedback for search terms that were unsuccessful automatically to the CMS. For that, the app team will add a flag to those searches. 

### Proposed Solution
In the CMS, we should add a label to those searches. 

- add a new field to the model search result feedback
- add a new column to the feedback as shown [in the design](https://www.figma.com/design/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=2803-362&t=gRyMFBptJ3BPB13s-1).

### Alternatives
Just add them to the normal feedback that was sent intentionally.

### User Story
As an admin for the Integreat app in my region, the label helps me distinguish more rapidly whether feedback was sent intentionally or not, thus showing me what to prioritize,

### Additional Context
Here the app issue: https://github.com/digitalfabrik/integreat-app/issues/2969#issuecomment-2456530034
Plus Steffis ideas on how to improve the function.
To avoid spelling mistakes, the app will add a "did you mean" function to the search, helping users spell the words correctly: https://github.com/digitalfabrik/integreat-app/issues/2981

### Design Requirements
Wir bräuchten ein design innerhalb der Feedbackübersicht für eine neue Spalte für das Label.

nikolahoff:
Ok, [here](https://www.figma.com/design/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=2803-362&t=HWW7sIxbLdB78AT6-1) is the design with the added column. Maybe it makes sense to add "unsuccessfull searchterm" to the filters?

osmers:
@nikolahoff didn't we want to add a flag that told municipalities that the feedback was sent automatically? Not that the search was unsucessful - bcs any search sent as feedback was unsucessful. Just that up till now the users had to actively send the feedback..

osmers:
@nikolahoff Not necessarily - the feedback would only be sent automatically, if a user does not decide to send it themselves. See screenshot for what unsuccessful feedback looks like. Users have the option of giving more information and then send the feedback. But we would like municipalities to also get the feedback where users decided that the barrier was too high and simply left the page. Feedback for unsuccessful searches is only sent upon closing or navigating away from that page in the screenshot.

[...]

PR #3533, opened 2025-03-19
Add flag when feedback has been automatically send

JoeyStk:
### Short description

This PR adds a flag for feedback in the search result when it has been send automatically.

### Proposed changes

- Add flag if feedback has been send automatically

### Side effects

- Please check if everything works correctly with the API. I'm not sure if I checked that correctly :/

### Resolved issues

Fixes: #3171
Fixes: #3584


integreat_cms/api/v3/feedback/search_result_feedback.py
```diff
@@ -24,6 +24,7 @@ def search_result_feedback(
     comment: str,
     rating: bool,
     is_technical: bool,
+    is_automatically_send: bool,
```
MizukiTemma, line 27:
```suggestion
```


This was the cause of failing tests (and error triggered when you try to hit the API by `curl -X "POST" "http://localhost:8000/api/v3/augsburg/de/feedback/search/" -d 'is_automatically_send=True&query=corona&comment=asdf&rating=down'`)

https://github.com/digitalfabrik/integreat-cms/blob/85a3d7e1d61e885bc50ac95079c3b9a15141a4c7/integreat_cms/api/decorators.py#L94

Here nothing is given for `is_automatically_send` but `search_result_feedback` requires it. That was causing the error. As other feedback classes do not have `is_automatically_send`, I recommend to handle it like `query`, by doing like this

> 
    if query := data.get("query"):
        is_automatically_send = data.get("is_automatically_send")
        SearchResultFeedback.objects.create(
            search_query=query,
            region=region,
            language=language,
            rating=rating,
            comment=comment,
            is_technical=is_technical,
            is_automatically_send=is_automatically_send,
        )

MizukiTemma (changes requested):
Thank you for the PR 💪 It seems I found the cause of failing tests/errors, please find thexplanation below. And there is a small suggestion to make the list look better.

integreat_cms/cms/templates/feedback/admin_feedback_list.html
```diff
@@ -73,6 +73,9 @@ <h1 class="heading">
                     <th class="text-sm text-right uppercase py-3 pr-2">
                         {% translate "Date" %}
                     </th>
+                    <th class="text-sm text-right uppercase py-3 pr-2">
+                        {% translate "Automatically send" %}
+                    </th>
```
MizukiTemma, line 78:
```suggestion
                    <th class="text-sm text-left uppercase py-3 pr-2">
                        {% translate "Date" %}
                    </th>
                    <th class="text-sm text-left uppercase py-3 pr-2">
                        {% translate "Automatically send" %}
                    </th>
```

integreat_cms/cms/templates/feedback/region_feedback_list_row.html
```diff
@@ -85,4 +85,11 @@
     <td class="pr-2 text-right whitespace-nowrap">
         {{ feedback.created_date|date:"SHORT_DATE_FORMAT" }}
     </td>
+    <td>
+        {% if feedback.automatically_send %}
```
MizukiTemma, line 89:
```suggestion
    <td class="pr-2 text-center">
        {% if feedback.is_automatically_send %}
```

integreat_cms/cms/templates/feedback/admin_feedback_list.html
```diff
@@ -70,9 +70,12 @@ <h1 class="heading">
                     <th class="text-sm text-left uppercase py-3 pr-2">
                         {% translate "Marked as read by" %}
                     </th>
-                    <th class="text-sm text-right uppercase py-3 pr-2">
+                    <th class="text-sm text-left uppercase py-3 pr-2">
                         {% translate "Date" %}
                     </th>
+                    <th class="text-sm text-left uppercase py-3 pr-2">
```
JoeyStk, line 76: I think this still should be `text-right` as we had before, but I'm fine either way

integreat_cms/api/v3/feedback/search_result_feedback.py
JoeyStk: Thank you! :)

integreat_cms/cms/templates/feedback/region_feedback_list.html
```diff
@@ -70,6 +70,9 @@ <h1 class="heading">
                     <th class="text-sm text-right uppercase py-3 pr-2">
                         {% translate "Date" %}
                     </th>
+                    <th class="text-sm text-right uppercase py-3 pr-2">
+                        {% translate "Automatically send" %}
+                    </th>
```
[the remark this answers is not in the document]
JoeyStk: I think this still should be `text-right` as we had before, but I'm fine either way

JoeyStk:
@MizukiTemma This PR is ready for review again :)

MizukiTemma (approved):
Looks good 😸 Thank you!

integreat_cms/cms/templates/feedback/admin_feedback_list.html
jarlhengstmengel, line 76: This changes only the non archived list. The non archived and the archived list both share `admin_feedback_list_row.html`. This means that automaticly send is only shown in the non archived list header, while it's visible in the row of the archived and non archived view. Should this be in both?
`admin_feedback_list.html` and `admin_feedback_list_archived.html`are probably very good candidates for refactoring since they are nearly the same, I think I will create a ticket for this

jarlhengstmengel (changes requested):
Thanks for the PR! See my comment for an Issue I found. I haven't tested the API yet

integreat_cms/cms/templates/feedback/admin_feedback_list.html
JoeyStk: Thank you for your feedback! I added a commit that refactors the feedback templates and this should fix this issue too :) Could you please review again? :) Thank you in advance :)

PeterNerlich:
> @jarlhengstmengel I haven't tested the API yet

There are fancy UIs like POSTman, but the most used traditional tool is just **curl**. Look up the manpage or [some online cheat sheet](https://linuxize.com/post/curl-post-request/) and you'll quickly get started inspecting the return of single hypertext requests and gain insight into the behaviour of their respective endpoints.
You can set the request method like `-X POST`, send data with `-d key=value -d next=value` and if it should be json data conveniently with `--json '{}'`, together with a lot more nice shell shenanigans. You'll need to extract your session and csrf cookies from your browser *(sessionid is somehow not part of what `document.cookie` returns in the JS console, you need to get it from the request headers listed in the network overview)*, and maybe also your useragent `--user-agent 'Mozilla/5.0 …'` *(get yours with `window.navigator.userAgent`)* to make the request as similar to the one sent by the browser as possible, in case it raises security suspicions otherwise.

jarlhengstmengel:
> > @jarlhengstmengel I haven't tested the API yet
> 
> There are fancy UIs like POSTman, but the most used traditional tool is just **curl**. Look up the manpage or [some online cheat sheet](https://linuxize.com/post/curl-post-request/) and you'll quickly get started inspecting the return of single hypertext requests and gain insight into the behaviour of their respective endpoints. You can set the request method like `-X POST`, send data with `-d key=value -d next=value` and if it should be json data conveniently with `--json '{}'`, together with a lot more nice shell shenanigans. You'll need to extract your session and csrf cookies from your browser _(sessionid is somehow not part of what `document.cookie` returns in the JS console, you need to get it from the request headers listed in the network overview)_, and maybe also your useragent `--user-agent 'Mozilla/5.0 …'` _(get yours with `window.navigator.userAgent`)_ to make the request as similar to the one sent by the browser as possible, in case it raises security suspicions otherwise.

Thank you, that already helps. I think my main problem is that I have no idea what urls to use. The ones in our documentation for feedback are not returning for me what I would expect from the documentation https://digitalfabrik.github.io/integreat-cms/api-docs.html#feedback

PR #3543, opened 2025-03-26
Update twine to continue being able to upload packages to pypi

PeterNerlich:
### Short description

Since this week, pipelines on the `develop` branch have failed at the step where the successfully built package is uploaded to test.pypi.org. This seems to stem from `PKG-INFO` in the tarball now listing `Metadata-Version: 2.4`, while older versions of twine only support up to `2.3`. *(I did not investigate the reason this changed in the last two weeks, I just assume the build step must have produced `2.3` until recently)*

### Proposed changes

- Upgrade `twine` only to the newest version supporting `Metadata-Version: 2.4` to speed up review
  - *This was done by, for the lack of a better understanding, manually setting the pinned version of `twine` to `6.1.0`, removing its dependency `pkginfo` as per the changelog, and running a modified version of `tools/update_dependencies.sh` that skips JS dependencies, installs the new **pinned** (dev) dependencies (instead of the newest ones) in the temporary venv and writes back the package versions from that to the `pyproject.toml`*
- Allow upgrades to other packages that still happen for some case, too, since this feels safer to me than not allowing them :shrug: 

This should affect the pipeline, since the workspace from previous steps is persisted and loaded on the publishing step, before activating the virtual environment from it

### Side effects

- Anything that might have relied on `pkginfo` but not listed it as a required dependency will break. This might conceivably include our scripts in `tools/`, though I doubt the chance is high
- The extra dependencies despite `django-celery` are slightly suspicious to me *(meaning that I merely don't understand why they are suddenly listed)*

### Resolved issues

Fixes: Our release process


PeterNerlich:
Unfortunately, I'm not quite sure how to properly test this except merging it on `develop` and seeing whether it works

charludo (approved):
> Unfortunately, I'm not quite sure how to properly test this except merging it on develop and seeing whether it works

Let's try 😁

JoeyStk (approved):

charludo:
@JoeyStk hey, Urlaub genießen! 😄

PeterNerlich:
Welp. Maybe I should have fixed the deprecation notices while I was at it. Now off to more research how this field is supposed to be…

charludo:
> Welp. Maybe I should have fixed the deprecation notices while I was at it. Now off to more research how this field is supposed to be…

Was worth a try though 😃

Issue #1153, opened 2022-02-09
Add search to organization list


timobrembeck:
### Motivation

The organization list is missing a search input

### Proposed Solution

Add the possibility to search for a specific organization in the organization list

### Additional Context

Also implement the ajax search suggestions for this object type

PR #3557, opened 2025-04-01
Add search function to the organization list

MizukiTemma:
### Short description

This PR adds seach function to the organization list

### Proposed changes

- Add the input window like other lists.
- Expand `search_content_ajax` to show suggestions
- Add `search` to the organization model
- Adjust the view and template

### Side effects

- `search_content_ajax` is now breaking three rules `PLR0915`, `PLR0912`, `C901`. Refactoring is needed.
- 

### Resolved issues

Fixes: #1153 


integreat_cms/cms/views/utils/search_content_ajax.py
```diff
@@ -71,7 +72,7 @@ def format_object_translation(
 
 
 @require_POST
-def search_content_ajax(
+def search_content_ajax(  # noqa: PLR0915, PLR0912, C901
```
JoeyStk, line 75: I don't think we should just comment these lines out. All three seem reasonable to me. Could you please take the opportunity to refactor this method and split it up into multiple methods, so we don't need these ignores? :)

JoeyStk (changes requested):
Thank you! This looks mostly good!
I wasn't able to test it on the server yet, but I will do so soon!
I would like to ask you to refactor a little bit of the code in order for the code to be more clean :)

integreat_cms/cms/views/utils/search_content_ajax.py
MizukiTemma: It'll be tackled as #3558 :)

MizukiTemma: #3562 is opened for refactoring. This does not bring a thorough refatoring but should be enough for now.

JoeyStk (approved):
In our grooming on Monday we had discussed this and said that it's fine to do a refactoring after the PR, therefore I'm going to approve this PR :)

PR #3560, opened 2025-04-02
Catch error when translating links into unsupprted languages

svenseeberg:
### Short description
When translating links of other regions, not all languages are supported by the other region. This can result in an `KeyError: 'ar'`, see https://chat.tuerantuer.org/digitalfabrik/threads/c9iytpszs7dz3mdtdrhq47raro
This seems to happen for location objects only right now, pages do not seem to be a problem. Nonetheless it probably does not hurt to introduce some general error handling that keeps the original URL if it cannot be updated.

### Proposed changes
- Catch error and keep original URL

### Side effects
- URLs in translations will keep the original link target.

### DIFF
--- a/integreat_cms/cms/utils/internal_link_utils.py
+++ b/integreat_cms/cms/utils/internal_link_utils.py
@@ -52,9 +52,17 @@ def update_link(
     if not (source_translation := get_public_translation_for_link(current_url)):
         return None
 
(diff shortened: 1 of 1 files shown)


MizukiTemma (approved):

charludo (approved):
Thank you!

PR #3565, opened 2025-04-04
Remove Chat body attribute

svenseeberg:
### Short description
To be more OpenAI like in our API, we replaced the body with a content attribute. The app will implement this with https://github.com/digitalfabrik/integreat-app/pull/3175

### Proposed changes
- Remove the legacy body attribute.

### Side effects
- API clients relying on the body attribute will break. As this never was production ready, it is acceptable.

### Resolved issues

Fixes: https://github.com/digitalfabrik/integreat-chat/issues/250

### DIFF
--- a/integreat_cms/cms/utils/zammad.py
+++ b/integreat_cms/cms/utils/zammad.py
@@ -108,7 +108,6 @@ class ZammadAPI:
             "status",
             "error",
             "id",
-            "body",
             "user_is_author",
             "automatic_answer",
             "role",


jonbulz (approved):
As [3175](https://github.com/digitalfabrik/integreat-app/pull/3175) is now merged, I guess this is ready

svenseeberg:
We should still ensure that the next app release is scheduled before the next CMS release with this feature. [Currently, there is no release scheduled](https://chat.tuerantuer.org/digitalfabrik/pl/83ygqw5btbfbjrepwnnczxepwh).

charludo (commented):
We have recently had the topic of API changes with the contacts, where the app team (rightfully) warned about outright removing API fields which old app-installs depend on.

I am not fully up to speed here - has the Chat already been released? 😅 If no, I am good with this change; but otherwise, I think we need to keep it a while longer 😢

svenseeberg:
There will be no release of the app before https://github.com/digitalfabrik/integreat-app/pull/3200 gets merged.

charludo:
> There will be no release of the app before [digitalfabrik/integreat-app#3200](https://github.com/digitalfabrik/integreat-app/pull/3200) gets merged.

Yes, but users may use outdated app versions, requiring the old field to be kept around for backwards compatibility.

svenseeberg:
> Yes, but users may use outdated app versions, requiring the old field to be kept around for backwards compatibility.

No, the app does not have the chat.

svenseeberg:
An app release is coming today, that means we can merge.

charludo:
> > Yes, but users may use outdated app versions, requiring the old field to be kept around for backwards compatibility.
> 
> No, the native app does not have the chat. As we're only in a very early beta testing phase, we never cared to implement the chat for native.

Alright, in that case that's no issue then :)

charludo (approved):

Issue #3166, opened 2024-11-01
Incorrect request when trying to import more than 100 xliff files


osmers:
### Describe the Bug
Server responds with 400 error for incorrect request when trying to upload multiple xliffs at once.

### Steps to Reproduce

1. Go to https://admin.integreat-app.de/testumgebung/pages/de/
2. Download xliffs for any language (needs to be more than 100)
3. Extract .zip file content into a new folder
4. Go to pages and scroll down to XLIFF import section
5. Click select files and browse to folder with extracted xliff files
6. Select more than 100 xliff files (e.g. 101)
7. Click Import
8. See error

### Expected Behavior
Xliffs get uploaded and the preview/confirmation screen is shown.

### Actual Behavior
400 error

### Additional Information
![image](https://github.com/user-attachments/assets/1de3125d-9095-4428-9a6f-a5c790d64720)

MizukiTemma:
@juliankehne 
Could you test whether this bug can be reproduced?

I failed both on the test/live system and the local environment.

JoeyStk:
Added `Question` label because it's unclear if this is can be reproduced

juliankehne:
@MizukiTemma @JoeyStk Yes, I was able to reproduce it. I updated the steps to reproduce and added some details I found. Hope it helps :)

JoeyStk:
FYI (probably for @juliankehne): This issue is not fixed, but there is a warning shown now

PR #3566, opened 2025-04-04
Fix error at xliff upload

MizukiTemma:
### Short description

This PR fixes the problem that users encounter the bad request error screen when more than 100 xliff files are uploaded.

### Proposed changes

- Check the number of selected file
- If it exceeds 100, show an error and keep the button "import" deactivated

### Side effects

- should be none
- 

### Resolved issues

Fixes: #3166


MizukiTemma:
To reproduce the error: either use a dump data and go to a region with a lot of pages or abuse the management command `integreat-cms-cli duplicate_pages REGION_SLUG` in the local test data.

integreat_cms/cms/templates/pages/page_tree.html
```diff
@@ -274,6 +274,10 @@ <h3 class="font-bold text-xl">
             <div class="my-2 text-s text-gray-600">
                 {% translate "Supported file extensions" %}: .zip, .xlf, .xliff
             </div>
+            <div id="xliff-error-message"
+                 class="bg-red-100 border-l-4 border-red-500 text-red-700 px-4 py-3 hidden">
+                {% translate "You can upload at once only up to 100 files." %}
```
JoeyStk, line 279:
```suggestion
                {% translate "You can upload only up to 100 files at once." %}
```
Nit

integreat_cms/release_notes/current/unreleased/3166.yml
```diff
@@ -0,0 +1,2 @@
+en: Prevent the bad request error and show an warning when users try to upload more than 100 xliff files
+de: Verhindere den Fehler "Inkorrekte Anfrage" und zeige eine Warnung an, wenn Benutzer:innen versuchen, mehr als 100 xliff-Dateien hochzuladen
```
JoeyStk, line 2: I remember that we said we should use less technical terms and one given example was xliff. So maybe it's better to write "Übersetzungsdatei" instead of xliff-file, but it's not really relevant and I'm fine either way :)

JoeyStk (approved):
Thank you very much!
I wasn't able to trigger this locally, but code looks good. I hope the second reviewer can test the message in the UI.

Do we still want to fix the underlying issue or do you think that it's not possible to fix the issue? :thinking:

MizukiTemma:
> Do we still want to fix the underlying issue or do you think that it's not possible to fix the issue? 🤔

I wanted to, but it seems the problem/restriction is on Python/Django side and the error occurs before `upload_xliff` is called but after clicking the import button 👀 There may be a solution but I didn't find any feasible way to touch the error.

integreat_cms/release_notes/current/unreleased/3166.yml
MizukiTemma: Applied :)

integreat_cms/static/src/js/pages/xliff-file-upload.ts
```diff
@@ -14,6 +14,18 @@ const selectXliffFile = ({ target }: Event) => {
     const label = document.querySelector("#xliff_file_label");
     label.classList.remove("bg-blue-500", "hover:bg-blue-600");
     label.classList.add("bg-gray-500", "hover:bg-gray-600");
+
+    if (files.length > 100) {
+        const timeoutDuration = 10000;
```
jarlhengstmengel, line 19: Just one personal nit, feel free to ignore. I personally felt during testing that the error message disappeared quite fast, but it's just a subjective feeling.

jarlhengstmengel (approved):
Thank you very much for this fix! I managed to trigger the error message and I think it looks good! Would you merge this before #3325?

integreat_cms/static/src/js/pages/xliff-file-upload.ts
MizukiTemma: Thank you for review 😸 The duration is doubled now ⏲️

JoeyStk:
@MizukiTemma do we need new issues for this? :)

MizukiTemma:
@JoeyStk 

Created #3580 :)

PR #3567, opened 2025-04-04
Add custom Chat privacy policy

svenseeberg:
### Short description
Some regions (cities) may want to use their own Zammad server. In this case the normal Integreat Privacy Policy no longer applies.

### Proposed changes
- Remove region `zammad_chat_handlers` handlers and add a new `zammad_privacy_policy` field

### Side effects
N/A

### Resolved issues

Fixes: https://github.com/digitalfabrik/integreat-chat/issues/262


svenseeberg:
The API documentation needs an update.

integreat_cms/api/v3/chat/utils/chat_bot.py
```diff
@@ -130,6 +130,7 @@ def process_user_message(
             internal=True,
             automatic_message=True,
         )
+    zammad_chat.processing_answer = False
```
charludo, line 133: I just realized - the `zammad_chat` never gets saved, right?

charludo (changes requested):
LGTM, apart from the comment below

integreat_cms/api/v3/chat/utils/chat_bot.py
svenseeberg: Ah yes, this line was committed accidentally. Initially I wanted to store this information in the model but then went on to just store it in the cache.

charludo: Alright, makes sense. What threw me was just that this is an object attribute which never gets saved. IMO a short comment explaining that this is intentional would be great.

charludo (approved):
> I added this to the description of the methods.

Perfect, thank you!

PeterNerlich (approved):
I put this off for too long already, sorry about that. I don't understand the implications of moving `zammad_chat.processing_answer = False`, but with two pairs of eyes deeming it not critical I'll not let this become another delay.

svenseeberg:
> I put this off for too long already, sorry about that. I don't understand the implications of moving `zammad_chat.processing_answer = False`, but with two pairs of eyes deeming it not critical I'll not let this become another delay.

Basically this only changes the time when the app will not show the typing indicator anymore.

PR #3569, opened 2025-04-06
Do not count excluded content attributes towards translation budget

charludo:
### Short description

Just noticed that content object attributes which get excluded from translation (currently, mostly titles, but maybe more in the future?) were still being counted towards the translation budget, because the `word_count` function used a hardcoded attribute list. 

### Proposed changes

- pass content object specific list of "wanted" attributes to that function
- 

### Side effects

- none, I think
- 

### Resolved issues

Fixes: n/a

### DIFF
--- a/integreat_cms/cms/views/utils/machine_translations.py
+++ b/integreat_cms/cms/views/utils/machine_translations.py
@@ -61,18 +61,18 @@ def build_json_for_machine_translation(
     source_language = request.region.get_source_language(language_slug)
 
     for content in selected_content:
+        source_translation = content.get_translation(source_language.slug)
+
         content.translatable_attributes = [
-            (attr, getattr(content.source_translation, attr))
+            (attr, getattr(source_translation, attr))
             for attr in TRANSLATABLE_ATTRIBUTES
-            if hasattr(content.source_translation, attr)
-            and getattr(content.source_translation, attr)
+            if hasattr(source_translation, attr)
+            and getattr(source_translation, attr)
             and not (content.do_not_translate_title and attr == "title")
         ]
         words = word_count(
             content.translatable_attributes,
         )
-
-        source_translation = content.get_translation(source_language.slug)
 
         if source_translation and check_hix_score(
             request,
--- a/integreat_cms/cms/views/utils/machine_translations.py
+++ b/integreat_cms/cms/views/utils/machine_translations.py
@@ -10,7 +10,9 @@ from typing import TYPE_CHECKING
 
 from django.http import JsonResponse
 
+from ....core.utils.word_count import word_count
 from ....textlab_api.utils import check_hix_score
+from ...constants.machine_translatable_attributes import TRANSLATABLE_ATTRIBUTES
 from ...models import Event, Page, POI
 
 if TYPE_CHECKING:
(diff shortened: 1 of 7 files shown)


MizukiTemma (approved):
Nice catch 💪 Thank you!

JoeyStk (approved):
Good catch! Thank you very much! 
I noted one thing when testing this: The number of estimated needed budget in the MT popup is 1. a bit off (but this is known, at least to me :sweat_smile:) and 2. I think it still calculates the title, even if the checkbox is not checked. But I approved this PR, assuming that this wouldn't be part of this PR, right? :thinking:

charludo:
> Good catch! Thank you very much! I noted one thing when testing this: The number of estimated needed budget in the MT popup is 1. a bit off (but this is known, at least to me 😅) and 2. I think it still calculates the title, even if the checkbox is not checked. But I approved this PR, assuming that this wouldn't be part of this PR, right? 🤔

Hmmm, good question. The *internal* budget estimate should be corrected with this PR, but I am unsure if we re-use that to show it to the user? If not, we probably should. I can look into this in this PR, or we merge this one now and I will open a new one - up to you.

JoeyStk:
I'm also fine either way :sweat_smile: But if it really is okay with you then maybe it would be better to do it in this PR, since we hopefully should get it fixed before the release on Thursday? :thinking:

charludo:
> I'm also fine either way 😅 But if it really is okay with you then maybe it would be better to do it in this PR, since we hopefully should get it fixed before the release on Thursday? 🤔

Just tried to implement it. Unfortunately, I keep getting an error message about DeepL being unavailable, so cannot really test it out.....

charludo:
@MizukiTemma should I undo the second part of the changes, so we can already merge the changes you and @JoeyStk approved?

MizukiTemma:
@charludo I'll check if the second part is working fine. If it's the case, we can deliver it together :)

integreat_cms/cms/views/utils/machine_translations.py
```diff
@@ -59,14 +61,19 @@ def build_json_for_machine_translation(
...
+            if hasattr(content.source_translation, attr)
+            and getattr(content.source_translation, attr)
+            and not (content.do_not_translate_title and attr == "title")
+        ]
+        words = word_count(
+            content.translatable_attributes,
         )
 
+        source_translation = content.get_translation(source_language.slug)
+
```
MizukiTemma, line 76:
```suggestion
        source_translation = content.get_translation(source_language.slug)

        content.translatable_attributes = [
            (attr, getattr(source_translation, attr))
            for attr in TRANSLATABLE_ATTRIBUTES
            if hasattr(source_translation, attr)
            and getattr(source_translation, attr)
            and not (content.do_not_translate_title and attr == "title")
        ]
        words = word_count(
            content.translatable_attributes,
        )

```

MizukiTemma (changes requested):

integreat_cms/cms/views/utils/machine_translations.py
charludo: Oops, yes, of course!

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -4715,7 +4715,7 @@ msgstr "Zu übersetzende %(content_type)s"
 #: cms/templates/_machine_translation_overlay.html
 #, python-format
 msgid ""
-"None of the selected %(content_type)s fullfills the required HIX score of "
+"None of the selected %(content_type)s fulfills the required HIX score of "
```
JoeyStk, line 4718: Thanks!

JoeyStk (approved):
Thank you a lot for your changes and I'm sorry it took me so long to get back to this :/

charludo:
> Thank you a lot for your changes and I'm sorry it took me so long to get back to this :/

No worries. Sorry for not being able to properly test this myself, IDK why DeepL won't work 😢

MizukiTemma (approved):
Thank you 🚀

PR #3571, opened 2025-04-07
Explicitely configure trusted source IP header

svenseeberg:
### Short description
@charludo was right all along. We should play it safe an assume that admins do not configure the CMS correctly.

### Proposed changes
- Do not trust the `X-Forwarded-For` header by default.
- Do not manually serialize list of timestamps into string as the cache can directly work with lists.

### Side effects
- Admins have to configure the trusted source IP header explicitly in the CMS and web server. 

### Resolved issues
Fixes: https://github.com/digitalfabrik/integreat-chat/issues/320

### DIFF
--- a/integreat_cms/core/settings.py
+++ b/integreat_cms/core/settings.py
@@ -206,6 +206,13 @@ BACKGROUND_TASKS_ENABLED = bool(
 #: The tag that events from external calendars need to get imported
 EXTERNAL_CALENDAR_CATEGORY: Final[str] = BRANDING
 
+#: HTTP Header that contains the client IP. This can be used behind reverse proxies
+#: Important: Use Django Header notiation, example: HTTP_X_FORWARDED_FOR. Ensure that
+#: this header cannot be sent by clients.
+TRUSTED_IP_HEADER: Final[str | None] = os.environ.get(
+    "INTEGREAT_CMS_TRUSTED_IP_HEADER", None
+)
+
 ##############################################################
 # Firebase Push Notifications (Firebase Cloud Messaging FCM) #
 ##############################################################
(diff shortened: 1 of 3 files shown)


integreat_cms/core/settings.py
```diff
@@ -206,6 +206,10 @@
 #: The tag that events from external calendars need to get imported
 EXTERNAL_CALENDAR_CATEGORY: Final[str] = BRANDING
 
+#: HTTP Header that contains the client IP. This can be used behind reverse procies
```
charludo, line 209:
```suggestion
#: HTTP Header that contains the client IP. This can be used behind reverse proxies
```

integreat_cms/core/settings.py
```diff
@@ -206,6 +206,10 @@
...
+# Important: Use Django Header notiation, example: HTTP_X_FORWARDED_FOR.
+TRUSTED_IP_HEADER = os.environ.get("INTEGREAT_CMS_TRUSTED_IP_HEADER", None)
+
```
charludo, line 212: Please assign a type here

charludo (changes requested):
> @charludo was right all along. We should play it safe an assume that admins do not configure the CMS correctly.

Hooray for paranoia! 😂 

We could take it a step further and not just apply this to ratelimiting, but actually refuse all connections without the header in case the config option is set?

svenseeberg:
> We could take it a step further and not just apply this to ratelimiting, but actually refuse all connections without the header in case the config option is set?

I hope that this case never occurs. But it should not be harmful therefore why not.

svenseeberg:
Okay, now I'm happy with the state of the feature ^^

svenseeberg:
Yes, indeed, I did miss it :grimacing:  Right now I'm trying to not burden the CMS team with PRs that are irrelevant to the normal operations of the Integreat CMS. If there are relevant interactions with "supported Integreat functionality" I do not merge w/o reviews :)

PR #3577, opened 2025-04-10
Allow svg uploads in media library

charludo:
### Short description

In preparation of the high contrast mode, allowing SVG uploads in the media library is required. This PR implements the necessary changes, most notably the handling of thumbnail generation for the SVGs.
 
### More context
https://chat.tuerantuer.org/digitalfabrik/pl/p7sif9m59pf8djpwzemsubjw9h
https://chat.tuerantuer.org/digitalfabrik/pl/kfouqiunb7drp81nkt4qiyamya

### Proposed changes

- allow svg file uploads
- modify thumbnail generation code to be able to handle svgs (by first converting them to png...)
- amend docs where appropriate to indicate the new dependency

### Side effects

- a new dependency, `libcairo2`. Not just the python packages, but the actual lib used by it!! If this is a deal-breaker, I'd understand

n/a

### Resolved issues

Fixes: #3583


JoeyStk:
I tried to upload the first level icons as svg and I still get this issue shown in the screenshot. Is there something I'm doing wrong? :thinking:

JoeyStk (approved):
Works well for me! Thanks a thousand for the quick implementation! :rocket:

charludo:
> Ah, one thing I forgot to ask. Do you think we need a release note for this? I'd say it might be interesting to the municipalities, right? But I assume we would need to create an issue for it first then 🤔

Yep, fair point :)

jonbulz (approved):
LGTM, although I had trouble checking the thumbnail generation locally. Uploading SVG works well, though

charludo:
> I had trouble checking the thumbnail generation locally. 

Could you elaborate on this, just to make sure? 😅

jonbulz:
> > I had trouble checking the thumbnail generation locally.
> 
> Could you elaborate on this, just to make sure? 😅

I think that in my local environment, the URL of images in the media library is not actually pointing to the correct file (because it is directing to a domain different from localhost), and thus the thumbnail was not loading. That was affecting png and svg files

charludo:
> > > I had trouble checking the thumbnail generation locally.
> > 
> > 
> > Could you elaborate on this, just to make sure? 😅
> 
> I think that in my local environment, the URL of images in the media library is not actually pointing to the correct file (because it is directing to a domain different from localhost), and thus the thumbnail was not loading. That was affecting png and svg files

Alright, sounds like that is indeed unrelated. However, it shouldn't happen either, so possibly your dev setup has a weird setting somewhere? 😅

charludo:
uuuuuuuuuuhm

I wanted to enable auto-merge, but the button changed to "merge now"an instant before I clicked it, and now the branch is merged without going through the CircleCI pipeline again

WTAF

PR #3579, opened 2025-04-11
Clone region with Celery

svenseeberg:
### Short description
Demo of how we can move the cloning operations into the Celery worker and avoid timeouts during region cloning. Feel free to take over.

### Proposed changes
- Move cloning content from the form save() method into a dedicated function that can be processed asynchronously.

### Side effects

- Maybe fewer than #3578
- As this will be processed in the background, we should notify the user that this can take up to several hours.
- We should ensure that the celery task timeout is high enough. AFAICT currently its 1h and should be increased to maybe 3h?
- Cloning the region blocks the Celery worker for a long time. This would lead to chats not being processed. That means we need at least a second worker process, which should not be an issue besides increasing the argument for Celery.

### Resolved issues

Fixes: #3642 


JoeyStk:
What still needs to be done in order to get this PR ready/done? Are you just waiting for reviews?

svenseeberg:
Currently this is just an idea/proposal for an implementation. This is not ready and as written in the initial message anyone can take over.

JoeyStk:
Add region status: "currently in cloning"

MizukiTemma:
Following changes are added:

- Add a new region status `Currently in cloning`
- Show an warning on the form when the cloning procedure is not yet finished
- Adjust the region status and hix setting to the original input after cloning is done

jonbulz:
@MizukiTemma I did some refactoring, I hope you're okay with it. The main thing is we don't think it's necessary to use threading for create_and_update_links anymore since it's run in a background task anyway. Additionally, we separated adjusting hix and region status from this create_and_update_links function.
There's 2 more things @PeterNerlich and I talked about:
1) It would be nice to have some kind of warning for users if the cloning tasks fail somehow. The only thing that I came up with thus far is to have another region status "Cloning failed" or something that gets set in case something goes wrong. Then we could have some kind of flag or warning message, what do you think?
2) These changes now would make Redis more or less mandatory, even in a dev environment. I will look into possible ways to set a fallback broker for Celery if Redis is not installed

That's a quick update, feel free to intervene if you don't like where this is going. I think I have enough time in the coming days to get this PR ready

MizukiTemma (approved):
@jonbulz @PeterNerlich 
Thank you for the improvement :heart_eyes_cat: It's a good idea!

I added a step to populate HIX values at the end of the cloning tasks. Feel free to modify or drop it. (It seems the commit wasn't delivered immediatery. There should be  the 7th commit :/ )

juliankehne:
One thing I noticed is that the region status stays in "cloning in progress" for quite a while AFTER all pages and translations are already cloned completely.

I have no idea what it's doing during that time, if its expected or an issue. Any ideas?

MizukiTemma:
Uhh, that sounds extreme 😓 @PeterNerlich @jonbulz Any idea?

svenseeberg:
We can maybe also implement a "retry" function? KIll the cloning if it runs into a deadlock (after about 2h) and restart the process?

PeterNerlich:
> We can maybe also implement a "retry" function?

Yes, though for this to work properly we need to ensure that we actually can just try the same thing again. In this case, we would quickly fail because the target region already exists. So we would need to implement a rollback or steps for idempotence or something like that, which sounds like a pain

Issue #1152, opened 2022-02-09
Add search to language list


timobrembeck:
### Motivation

Since the language list is getting long, it makes sense to add a search for the language list too :)

### Proposed Solution

Add the possibility to search for a specific language in the language list

### Additional Context

Also implement the ajax search suggestions for this object type

PR #3585, opened 2025-04-14
Add search function to the language list

MizukiTemma:
### Short description

This PR adds a search function/window to the language list.

### Proposed changes

- Add a search window like other lists
- Add a search function

### Side effects

- should be none
- 

### Resolved issues

Fixes: #1152 

### DIFF
--- a/integreat_cms/cms/models/languages/language.py
+++ b/integreat_cms/cms/models/languages/language.py
@@ -203,6 +203,24 @@ class Language(AbstractBaseModel):
         """
         return self.slug not in settings.PDF_DEACTIVATED_LANGUAGES
 
+    @classmethod
+    def search(cls, query: str) -> QuerySet:
+        """
+        Searches for all languages which match the given `query` in their native name or english name.
+        :param query: The query string used for filtering the languages
+        :return: A query for all matching objects
+        """
+        result = []
+        for language in cls.objects.all():
+            if (
+                query.lower() in language.native_name.lower()
+                or query.lower() in language.english_name.lower()
+                or query.lower() in language.translated_name.lower()
+            ):
+                result += [language.id]
+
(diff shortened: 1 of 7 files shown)


integreat_cms/cms/models/languages/language.py
```diff
@@ -203,6 +203,24 @@ def can_be_pdf_exported(self) -> bool:
...
+        :return: A query for all matching objects
+        """
+        result = []
+        for language in cls.objects.all():
+            if (
+                query.lower() in language.native_name.lower()
+                or query.lower() in language.english_name.lower()
+                or query.lower() in language.translated_name.lower()
+            ):
+                result += [language.id]
```
MizukiTemma, line 220: I know this is not beautiful 🤔 Though, query filter doesn't let users search for `translated_name` of languages as it is `cached_property`, and search not working for translated names may cause confusion because most of users use the system in German (for example "Ungersich". Most users look for "Ungerisch", but its native name is "Magyar" and english name is "Hungarian". Inputs like "Unge..." won't hit the language without taking `translated_name` into account).

charludo: IMO it's perfectly fine, esp. since there will never be enough languages for this to become a performance issue 😄

charludo (approved):
Works really well, just as intended!

I was a bit confused because this only works in the network management, not for each region - is that not intended? Or is it a separate issue because the page tree views pose different implementation challenges?

integreat_cms/cms/models/languages/language.py
JoeyStk: Since this feature is only available for internal users I don't think we should spend a lot of time on this. For me this is okay too :)

JoeyStk (approved):
Thank you very much! :rocket:

Issue #3593, opened 2025-04-15
Add search function to the region condition view


MizukiTemma:
### Motivation

There is currently no search function in the region condition table. Let's add such a search function like the one in the region list.

### Proposed Solution

Add a search feature

### Screenshot

Add the search window here
![Image](https://github.com/user-attachments/assets/44967c0b-1762-4eaf-b39d-fe39b25181c5)

This is an example
![Image](https://github.com/user-attachments/assets/f428ca63-1c05-40fc-8c31-4e560cd4706e)

PR #3594, opened 2025-04-16
Add search function to region conditions

MizukiTemma:
### Short description

This PR adds the same region search feature to the region condition table.

### Proposed changes

- Add an input window
- Use the same suggest/search function with the feature for region list
- Adjust but the suggest function to exclude archived regions if it is in the region condition table.

### Side effects

- Should be none
- 

### Resolved issues

Fixes: #3593 

### DIFF
--- a/integreat_cms/cms/views/region_condition/region_condition_view.py
+++ b/integreat_cms/cms/views/region_condition/region_condition_view.py
@@ -70,3 +70,13 @@ class RegionConditionView(TemplateView):
                 "search_query": query,
             },
         )
+
+    def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
+        r"""
+        Apply the query and filter the rendered regions
+        :param request: The current request
+        :param \*args: The supplied arguments
+        :param \**kwargs: The supplied keyword arguments
+        :return: The rendered template response
+        """
+        return self.get(request, *args, **kwargs, search_data=request.POST)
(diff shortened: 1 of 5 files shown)


JoeyStk (approved):
Thank you very much! Works as expected! :rocket:

integreat_cms/cms/views/region_condition/region_condition_view.py
```diff
@@ -70,3 +70,13 @@ def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
...
+
+    def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
+        r"""
+        Apply the query and filter the rendered regions
+        :param request: The current request
+        :param \*args: The supplied arguments
+        :param \**kwargs: The supplied keyword arguments
+        :return: The rendered template response
+        """
+        return self.get(request, *args, **kwargs, search_data=request.POST)
```
jonbulz, line 82: Same as for contacts, I'd prefer that the search view would just use a GET request. But this is outside of the scope of this PR. Maybe a future refactoring target? Though that would likely be a lot of changes...

jonbulz (approved):
At some point, I'd like to talk about the way we implement search functions. Maybe it makes sense to include them in list views by default instead of individually adding them for different models? I'd also prefer using GET requests instead of pseudo POST requests that end up being handled as GET requests anyway. But I'll leave that up for future discussion. Anyway, thanks for this PR, looks good. :)

Issue #3597, opened 2025-04-16
Options in 'Archived Pages' are the same as in 'Pages'


juliankehne:
### Describe the Bug

The options in Archived Pages are the same as in pages. Preview, Edit, Archive, etc. that means there is also no restore option.

Archived Pages Options are always the same ones that a specific user role permissions for. Which means Observer sees their observer options in Archived Pages, Editor sees Editor options, etc.

### Steps to Reproduce

1. Go to 'Pages'
2. Go to 'Archived Pages'
3. See that Options are the same as in 'Pages'

### Expected Behavior

Options should be Preview, Restore, Delete (Depending on user role no delete and/or restore)
![Image](https://github.com/user-attachments/assets/d50e0724-7aaf-48a5-8c72-15b1219ec491)

### Actual Behavior

Same options as in Pages. Restore option missing.
![Image](https://github.com/user-attachments/assets/aa64f6b1-ca7b-4c75-a933-ea961709c76f)

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

Issue #3596, opened 2025-04-16
Archive option for sub-pages missing


juliankehne:
### Describe the Bug

This might be a result of the page tree refactoring introduced with page-based statistics.
Subpages of a page inside the page tree have no archive icon next to them (See screenshot)

### Steps to Reproduce

1. Go to 'Pages'
2. Expand a page with sub-pages
3. Scroll to the right to see available options
4. See no archive button for sub-pages

### Expected Behavior

Subpages have the archive option.

### Actual Behavior

Sub pages don't have the archive option.

### Screenshot

![Image](https://github.com/user-attachments/assets/96f5222a-4ede-44d3-b3c5-7b078ef96135)

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

MizukiTemma:
@jarlhengstmengel 

This may be solved together with #3597 , which I'm working on now 🤔

PR #3601, opened 2025-04-16
Fix edit icon in the page tree

MizukiTemma:
### Short description

This PR fixes the bugs with the edit icon in the page tree.

### Proposed changes

- Delete the pen-square icon, as we had the pencil icon
- Hide the survived one if it is the archived page tree (partially solves #3597 )

### Side effects

- Should be none 
- 

### Faithfulness to issue description and design

There are no intended deviations from the issue and design.

### Resolved issues

Fixes: #3598 

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #3602, opened 2025-04-16
Fix archive/restore icons in page tree

MizukiTemma:
### Short description

This PR fixes the bugs with archive/restore icons in the page tree

### Proposed changes

- Fix the conditions in the template
- 

### Side effects

- The edit icon will be hidden for the archived page tree by #3601 
- 

### Resolved issues

Fixes: #3597 
Fixes: #3596 

### DIFF
--- a/integreat_cms/cms/templates/pages/pages_page_tree_node.html
+++ b/integreat_cms/cms/templates/pages/pages_page_tree_node.html
@@ -121,22 +121,22 @@
                 <i icon-name="eye"></i>
             </button>
         {% endif %}
-        {% if page_translation.status == PUBLIC %}
-            <a href="{{ WEBAPP_URL }}{{ page_translation.get_absolute_url }}"
-               rel="noopener noreferrer"
-               target="_blank"
-               title="{% translate "Open page in web app" %}"
-               class="btn-icon">
-                <i icon-name="external-link"></i>
-            </a>
-        {% else %}
-            <button title="{% translate "This page cannot be opened in the web app, because it is not public." %}"
-                    class="btn-icon"
-                    disabled>
-                <i icon-name="external-link"></i>
-            </button>
-        {% endif %}
         {% if not is_archive %}
+            {% if page_translation.status == PUBLIC %}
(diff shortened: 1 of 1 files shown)


JoeyStk:
> Fixes: https://github.com/digitalfabrik/integreat-cms/issues/3597 
Fixes: https://github.com/digitalfabrik/integreat-cms/issues/3596

This way the issues get both closed automatically :)

JoeyStk (approved):
Thank you! This seems to fix it! :)

integreat_cms/cms/templates/pages/pages_page_tree_node.html
```diff
@@ -121,76 +121,63 @@
...
+                        disabled>
+                    <i icon-name="external-link"></i>
+                </button>
+            {% endif %}
         {% endif %}
         <a href="{% url 'edit_page' page_id=page.id region_slug=request.region.slug language_slug=language.slug %}"
            title="{% translate "Edit page" %}"
            class="btn-icon">
             <i icon-name="pencil"></i>
         </a>
```
jarlhengstmengel, line 145: I think this also shouldn't be available for archived pages

jarlhengstmengel (changes requested):
Thank you very much for fixing both Issues :rocket:  I just have on nit, see my comment

integreat_cms/cms/templates/pages/pages_page_tree_node.html
MizukiTemma: Yes, I had to wait for #3601 

Now it should be hidden in the archived page tree :)

jarlhengstmengel: Ah sorry, I forgot that that was a another PR :) Just one last nit, maybe you could move that into the if block above? Since both blocks check for the same

MizukiTemma: Done :)

jarlhengstmengel (approved):
Thank you!

PR #3607, opened 2025-04-17
Add broken links column and pages with at least one missing translation to region condition

JoeyStk:
### Short description

Add two new columns to region condition: collect how many broken links there are per region and collect pages with at least one missing translation. This is part of the next iteration for the region condition. After that we probably need one or two more.

### Proposed changes

- Add broken links column
- Add pages with at least one missing/outdated translation

### Side effects

- We need to have an eye on the loading times of this section again

Should be true to the description

### Resolved issues

Fixes: parts of #3012 

### DIFF
--- a/integreat_cms/cms/models/regions/region.py
+++ b/integreat_cms/cms/models/regions/region.py
@@ -21,6 +21,8 @@ from django.utils.functional import cached_property, keep_lazy_text
 from django.utils.translation import gettext, override
 from django.utils.translation import gettext_lazy as _
 
+from integreat_cms.cms.constants import translation_status
+
 if TYPE_CHECKING:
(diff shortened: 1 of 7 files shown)


MizukiTemma (approved):
Thank you! Looks good 😸 

I have a bit of concern about loading time, it took me with the dump data ca 100 seconds. Let's see what the second reviewer says 🤔

charludo (approved):
Code looks good, and everything appears to work when testing it out.

I have not tested with a datadump, but 100s does sound a lot... It's only an internal tool, so probably not that bad, but I do not have any ideas on how to speed it up.

One thing I noticed, the titles of the columns are very long. I think you can easily from the "Number of " prefix from all of the columns, it's rather clear from the context.

Issue #3599, opened 2025-04-16
Statistics graph in Dashboard is changing size


juliankehne:
### Describe the Bug

The statistics box in the Dashboard has different sizes and sometimes is multiple times larger than expected.

From what I was able to figure out it seems that while the statistics are loading a user can scroll down on the dashboard and the Statistics graph with change size dynamically while scrolling resulting in graphs with massive scale. Sometimes it also just increases without the user scrolling down.

It's not always reproducible.

See screenshots.

### Steps to Reproduce

1. Go to a region that has Statistics enabled
2. Click on My Dashboard
3. Scroll down to Statistics
4. See that size of Statistics varies
5. Refresh page and see that statistics size changes sometimes

### Expected Behavior

Same size for statistics every time.

### Actual Behavior

Statistics size varies.

### Screenshot

![Image](https://github.com/user-attachments/assets/16ebe3bb-d13f-4f94-82a6-6260dbe862ad)
![Image](https://github.com/user-attachments/assets/e2e2b3b0-1340-4848-a13e-3fb1e5590f95)
![Image](https://github.com/user-attachments/assets/b9ff5d99-93f6-4665-a236-86f0545d4998)
![Image](https://github.com/user-attachments/assets/7c790455-03a7-474e-b191-412662b85b1b)

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

PR #3608, opened 2025-04-17
Fix statistics graph sizing

jarlhengstmengel:
### Short description

Fix weird sizing behavior of the statistics graph in the `My Dashboard` view.

### Proposed changes

- Put the graph in it's own container and fixate the height

### Side effects

- Hopefully none

### Resolved issues

Fixes: #3599 

### DIFF
--- a/integreat_cms/cms/templates/statistics/_statistics_widget.html
+++ b/integreat_cms/cms/templates/statistics/_statistics_widget.html
@@ -31,7 +31,9 @@
         <div id="chart-loading" class="px-4 hidden">
             <i icon-name="loader" class="animate-spin"></i> {% translate "Loading..." %}
         </div>
-        <canvas id="statistics"
-                data-statistics-url="{% url 'statistics_total_visits' region_slug=request.region.slug %}"></canvas>
+        <div class="chart-container w-auto h-[45vh]">
+            <canvas id="statistics"
+                    data-statistics-url="{% url 'statistics_total_visits' region_slug=request.region.slug %}"></canvas>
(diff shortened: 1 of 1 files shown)


MizukiTemma (approved):
Looks good 👍 Thank you!

JoeyStk (changes requested):
Thank you very much! This fixes the statistics on the dashboard :rocket: 
However I noticed something strange with the statistics in the statistics section. Currently (locally & test-cms) the statistics don't take the full width available width, which looks a bit strange. By comparison on admin the width is fully taken. I added two screenshots to compare :)
grafik
grafik

I think the solution is to change this
```suggestion
<div class="chart-container w-[52vw] h-[60vh]">
                <canvas id="statistics"
                        data-statistics-url="{% url 'statistics_visits_per_language' region_slug=request.region.slug %}"></canvas>
            </div>
```

to

```suggestion
<div class="chart-container h-[60vh]">
                <canvas id="statistics"
                        data-statistics-url="{% url 'statistics_visits_per_language' region_slug=request.region.slug %}"></canvas>
            </div>
```
(remove the `w-[52vw]`) :)

jarlhengstmengel:
> (remove the `w-[52vw]`) :)

That sadly creates another problem when sizing the window. The chart and the sidebar can overlap without setting the width that way. I do not really have an Idea how to fix both at the same time, but seeing that the other chart does not have that behavior, I think it should be fixable somehow. But since I'm away this week, I would maybe look into that with another PR, what do you think?

image

Issue #3537, opened 2025-03-20
Add search to various list views


PeterNerlich:
### Motivation

We have a small number of list views that can be filtered using a search term. Others do not offer this functionality. The language list view would definitely benefit from gaining a search bar, but when adding it to one, the remaining effort to add it to the others should be almost negligible.

Thus this issue serves as a pool for all existing issues adding search functionality to list views *(where implementation is expected to be similar and similarly easy)*.
(This list is not exhaustive)

Investigation is needed how far this could be done in one template as well as one abstract base class for all list views.
*(I briefly opened `integreat_cms/cms/views/list_views.py` to find the class `ModelListView` but didn't immediately see it used as a base class in e.g. `class EventListView(TemplateView, EventContextMixin, MachineTranslationContextMixin)` of `integreat_cms/cms/views/events/event_list_view.py` – though maybe it is just daisy-chain–inherited through many classes and abstraction layers, just as we enjoy it elsewhere in our system)*

MizukiTemma:
We can add a search to the contact list too.


----- Lists that have no search function but I don't think need any as the number of items is not so large
External calendar list
Role list
Offer list
POI Category list

MizukiTemma:
> Investigation is needed how far this could be done in one template as well as one abstract base class for all list views.
(I briefly opened integreat_cms/cms/views/list_views.py to find the class ModelListView but didn't immediately see it used as a base class in e.g. class EventListView(TemplateView, EventContextMixin, MachineTranslationContextMixin) of integreat_cms/cms/views/events/event_list_view.py – though maybe it is just daisy-chain–inherited through many classes and abstraction layers, just as we enjoy it elsewhere in our system)


I recomment to tackle this point in a separate issue, as it affects the existing search functions.

MizukiTemma:
@laravoelk 

Is search function needed for the region language tree too?

laravoelk:
For us currently the most important search functions would be within contacts and POI. I think the region language tree is not so important as most municipalities do not have so many languages. :-)

JoeyStk:
> For us currently the most important search functions would be within contacts and POI

POI (= Orte) should already have a search bar. Do you mean POI categories (= Ortskategorien)? :)

laravoelk:
Sorry for the misunderstanding. I meant a seach / sort function. For example sort from A-Z, newest to latest etc. :-)

PeterNerlich:
Would this also make sense for the broken links view?

Issue #3484, opened 2025-03-04
Add Search/Filter for Contacts in Contacts List


juliankehne:
### Motivation

Currently the Contacts page / list view of Contacts cannot be searched for specific contacts or filtered for certain parameters making it tedious to find a specific contact to edit.

Especially since we migrated Contact data from POIs to Contacts on release most regions already have a lot of contacts and are unable to search/filter for specific contacts.

### Proposed Solution

Just adding possible solutions here. It needs further discussion.

1. Match the search functionality we offer users to find contacts when inserting them. Basically type anything and if it matches any information of a contact get result(s) back.

2. A filter instead of a search field where a user can enter any parameter of a contact in separate fields and it only gets results back where it matches the filter. e.g. filter for name, email, phone, address, location, etc.

3. Minimal search functionality based on the location of a contact. Basically search location = find contacts of that location.

4. Ability to search for Location & Name or other combination, but not everything.

5. Something else?

### Alternatives

TBD

### User Story

As an editor I want to quickly find a specific contact so that I can edit the contact.

nikolahoff:
I added a search bar on top of the contact list. You can find the design [here](https://www.figma.com/design/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=1632-153&t=1WZjKYuotZFLhxt8-1)
It should be Possible to search for Name of contact or Title of POI to get results. Is that feasable @MizukiTemma ?

MizukiTemma:
>It should be Possible to search for Name of contact or Title of POI to get results. Is that feasable [@MizukiTemma](https://github.com/MizukiTemma) ?


Search by name of contact: yes
Search by name of POI: hmm, no so easy but I guess there is a way 💪

Issue #3558, opened 2025-04-01
Refactor `search_content_ajax`


MizukiTemma:
### Motivation

`search_content_ajax` is very long with many duplucated/similar parts. It surpresses three rules `PLR0915`, `PLR0912`, `C901`. 

### Proposed Solution

Refactor it!

### Alternatives

Leave it as is.

PR #3622, opened 2025-04-28
Add search function to contact list

MizukiTemma:
### Short description

This PR adds a search function to the contact list

### Proposed changes

- Add the search window
- Let users search for contact name or location name

### Side effects

- To re-use the `search` function of `POITranslation`, `language_slug=LANGUAGE_CODE` is handed over. I hope it is safe enough to assume all the system languages (currently English and German) will be always covered by the content languages.
- 

### Resolved issues

Fixes: #3484 
Fixes: #3537 

### DIFF
--- a/integreat_cms/cms/views/contacts/contact_list_view.py
+++ b/integreat_cms/cms/views/contacts/contact_list_view.py
@@ -9,6 +9,7 @@ from django.utils.decorators import method_decorator
 from django.views.generic import TemplateView
 
 from ...decorators import permission_required
+from ...forms import ObjectSearchForm
 from ...models import Contact
 from .contact_context_mixin import ContactContextMixin
 
@@ -37,6 +38,7 @@ class ContactListView(TemplateView, ContactContextMixin):
         :return: The rendered template response
         """
         region = request.region
+        query = None
 
         contacts = Contact.objects.filter(
             location__region=region,
@@ -47,6 +49,13 @@ class ContactListView(TemplateView, ContactContextMixin):
             location__region=region,
             archived=True,
         ).count()
+
+        search_data = kwargs.get("search_data")
+        search_form = ObjectSearchForm(search_data)
+        if search_form.is_valid():
+            query = search_form.cleaned_data["query"]
+            contact_keys = Contact.search_for_query(region, query).values("pk")
+            contacts = contacts.filter(pk__in=contact_keys)
 
         chunk_size = int(request.GET.get("size", settings.PER_PAGE))
         paginator = Paginator(
@@ -66,5 +75,16 @@ class ContactListView(TemplateView, ContactContextMixin):
                 "contacts": contact_chunk,
                 "archived_count": archived_count,
                 "current_menu_item": "contacts",
+                "search_query": query,
             },
         )
+
+    def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
+        r"""
+        Apply the query and filter the rendered contacts
+        :param request: The current request
+        :param \*args: The supplied arguments
+        :param \**kwargs: The supplied keyword arguments
+        :return: The rendered template response
+        """
+        return self.get(request, *args, **kwargs, search_data=request.POST)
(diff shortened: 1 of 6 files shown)


JoeyStk:
I think with this PR you can also close #3537 :)

integreat_cms/cms/views/contacts/contact_list_view.py
```diff
@@ -66,5 +75,16 @@ def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
...
+
+    def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
+        r"""
+        Apply the query and filter the rendered contacts
+        :param request: The current request
+        :param \*args: The supplied arguments
+        :param \**kwargs: The supplied keyword arguments
+        :return: The rendered template response
+        """
+        return self.get(request, *args, **kwargs, search_data=request.POST)
```
JoeyStk, line 90: This looks dangerous to me and might lead to an indefinite recursion. I think because of `search_data` this is not going to happen, but it still doesn't look ideal to me. I would be interested to know what @jonbulz thinks

integreat_cms/cms/views/utils/search_content_ajax.py
```diff
@@ -105,6 +106,34 @@ def search_content_ajax(  # noqa: PLR0915, PLR0912, C901
     results: list[dict[str, Any]] = []
 
     user = request.user
+
+    if "contact" in object_types:
```
JoeyStk, line 110: I know that you're already working on a refactoring for this, but maybe this is still helpful. 
I think this function (only the contact section) by itself already does a lot. For example I think it would be good to separate logical units and would extract the check for permissions to another function. But as said, I'm going to let it pass during this review :)

JoeyStk (commented):
Thank you very much! This is a cool step forward, especially since a few municipalities have requested this!

I'm a bit worried about the call to the `get` function, but might not have all of the insights here. So I'll just wait for @jonbulz's review. 
Additionally I'm not sure yet if everything works as expected, for example is it expected that I can only search for the name? I would have at least expected to also be able to search for the area of responsibility. What do you think/know about this? :)

integreat_cms/cms/views/contacts/contact_list_view.py
jonbulz: I don't think I see the infinite recursion, could you elaborate? It is already implemented this way for other models, i.e. POIs. However, I do agree that it is not ideal, mainly because in the "get" method of a view I would always expect a GET request. To have a POST request end up in the "get" method is a bit unintuitive to me. IMO the search_input_form should also submit a GET request, but changing that now would be a huge pain.

integreat_cms/cms/views/utils/search_content_ajax.py
```diff
@@ -105,6 +106,34 @@ def search_content_ajax(  # noqa: PLR0915, PLR0912, C901
...
+        )
+        if TYPE_CHECKING:
+            assert language_slug
+        results.extend(
+            {
+                "title": poi_translation.title,
+                "url": None,
+                "type": "contact",
+            }
+            for poi_translation in POITranslation.search(
```
jonbulz, line 132: Why POITranslation? What is the relation to contacts? Further down we do a similar query if pois are in object_types, but with more filters. It seems to me we should at least apply the same filters, and do a permission check for pois here as well.

integreat_cms/cms/views/contacts/contact_list_view.py
MizukiTemma: will be solved as #3558

jonbulz (approved):
LGTM :)

integreat_cms/cms/views/contacts/contact_list_view.py
JoeyStk: > I don't think I see the infinite recursion, could you elaborate?

I think I was wrong :)

JoeyStk (approved):

Issue #3628, opened 2025-04-29
Change wordings for chat


MizukiTemma:
### Motivation

In the region form there are two check boxes related to "chat".
Change "Integreat-Chat aktivieren/Activate author chat" (the fist one in the screenshot) to "Autor:innen-Chat aktivieren/Activate author chat and "Integreat Chat aktivieren/Enable Integreat Chat"(the second one in the screenshot) to "KI-Chat aktivieren/Enable AI chat"

### Proposed Solution

Rename them.

### Screenshot

![Image](https://github.com/user-attachments/assets/4600ea47-4cb1-49e4-8bf5-682b9cd2508a)

### Summary of discussion and updates to the description

"AI chat/KI chat" was rejected. it must be "Public chat/Öffentlichen Chat"

![Image](https://github.com/user-attachments/assets/9ca9d223-01ef-4d68-bc42-c4249725a48a)

PR #3629, opened 2025-04-29
Rename chats

JoeyStk:
### Short description

Currently we have two types of chats and both of them are called "Integreat Chat"/"Integreat-Chat". It was requested to make this more speaking. This PR renames some instances in our code base. See my explanation for it in the side effects section :).

### Proposed changes

- Rename Integreat Chats to either Author Chat or AI-Chat

### Side effects

- I didn't rename instances in our code base, where I was unsure if I should do so (for example: `INTEGREAT_CHAT_BACK_END_DOMAIN` in `settings.py`, `INTEGREAT_CHAT_BACK_END_TIMEOUT` in `settings.py`, `return JsonResponse({"status": "Integreat Chat disabled"})` in `user_chat.py, and other instances). I wanted to double check, mostly with @svenseeberg first, if it's okay of me to do so :).

### Resolved issues

Fixes: #3628


JoeyStk:
> Wasnt the internal chat supposed to be deprecated? 

This is the first time that I'm hearing of this. I would think no, but maybe I missed something?

> Beyond that I think something along the lines "(Public|LLM) Chat (Bot)" might be better suited.

@dkehne would you be okay with that too or would you still prefer AI-chat? :)

svenseeberg:
I understood that the new Discourse forum was discussed as an alternative to the internal chat. But maybe I got this wrong.

svenseeberg:
> I didn't rename instances in our code base, where I was unsure if I should do so (for example: `INTEGREAT_CHAT_BACK_END_DOMAIN` in `settings.py`, `INTEGREAT_CHAT_BACK_END_TIMEOUT` in `settings.py`, `return JsonResponse({"status": "Integreat Chat disabled"})` in `user_chat.py, and other instances). I wanted to double check, mostly with @svenseeberg first, if it's okay of me to do so :).

I think its okay. But we have to double check that the changes do not affect the web app.

JoeyStk:
> But we have to check that the API changes do not affect the web app.

@svenseeberg What do you mean by that? Because all I changed for API and it's docs is to change the comments. That shouldn't have any effect, right?

MizukiTemma (approved):
Looks good 😸 

I have updated **Summary of discussion and updates to the description** in the issue description, because the new naming was changed after some discussion. Please check if it is correct 👍

Issue #3122, opened 2024-10-08
Add delete bulk actions


david-venhoff:
### Motivation

Since #3080, we now have delete bulk actions. For consistency, they should also get added to all other existing bulk actions. It should be possible to bulk delete all selected entries for: 
- poi
- events
- pages
- language tree

To circumvent accidental deleting of objects, we need the confirmation popup for this task, if possible.

### Proposed Solution

Add delete bulk actions

### Alternatives

Non really, if we have delete bulk actions for contacts, there is no reason not to have them for other objects.

jarlhengstmengel:
This Issue should also probably look into integrating the permission to delete requirement into `get_permission_required()` in `bulk_action_views.py` to reduce the times this method needs to be overwritten.

jarlhengstmengel:
I'm wondering if adding the confirmation dialog should be a seperate Issue. Until now we never used it for bulk actions and adding to all wouldn't be that hard but I think outside of the scope of this Issue

PR #3646, opened 2025-05-05
Add bulk delete to entries

hannaseithe:
### Short description
This adds a bulk delete option to events, pages, pois and language tree nodes

### Proposed changes

- On the list form template for pois, events, pages and language tree nodes you can now select to delete multiple entries
- After chosing to execute a confirmation dialogue will show that asks you to confirm the action
- the functionality is bundled in a general bulk_action_delete function that is used in the view of the entries
- I have added to the models of the entries (where necessary) a helper method "canBeDeleted" and adjusted the delete methods, so that the general bulk_action_delete can reuse those
- for the tests I have aimed for more self-containment as such that the entries (pois, pages, ...) are dynamically generated and not inserted into test_data.json

### Side effects

- there have been some changes to the models of poi, event, languagetreenode and page (as mentionded above)

### Resolved issues

Fixes: #3122

### DIFF
--- a/integreat_cms/cms/models/pages/page.py
+++ b/integreat_cms/cms/models/pages/page.py
@@ -310,6 +310,18 @@ class Page(AbstractTreeNode, AbstractBasePage):
             return self._relative_depth
         return self.depth
 
+    def can_be_deleted(self) -> tuple[bool, str | None]:
+        """
+        Checks whether the page can be deleted
+        """
+        if self.children.exists():
+            return False, _("you cannot delete a page which has subpages.")
+        if self.mirroring_pages.exists():
+            return False, _(
+                "you cannot delete a page that is embedded as live content by another page."
+            )
+        return True, None
+
     def move(self, target: Page, pos: str | None = None) -> None:
         """
         Moving tree nodes potentially causes changes to the fields tree_id, lft and rgt in :class:`~treebeard.ns_tree.NS_Node`
--- a/integreat_cms/cms/templates/events/event_list.html
+++ b/integreat_cms/cms/templates/events/event_list.html
@@ -208,6 +208,17 @@
                         {% translate "Restore events" %}
                     </option>
                 {% endif %}
+                {% if perms.cms.delete_event %}
+                    <option class="bulk-confirmation-dialog"
+                            data-bulk-action="{% url 'delete_multiple_events' region_slug=request.region.slug language_slug=language.slug %}"
+                            data-popup-title="{% blocktranslate %}You are about to delete multiple selected {{content_type}}{% endblocktranslate %}"
+                            data-popup-subject="{% blocktranslate %}This action cannot be undone{% endblocktranslate %}"
+                            data-popup-text="{% blocktranslate %}Do you want to continue?{% endblocktranslate %}">
+                        {% blocktranslate trimmed %}
+                            Delete {{ content_type }}
+                        {% endblocktranslate %}
+                    </option>
+                {% endif %}
             </select>
             <button form="bulk-action-form" id="bulk-action-execute" class="btn" disabled>
                 {% translate "Execute" %}
--- a/integreat_cms/cms/views/bulk_action_views.py
+++ b/integreat_cms/cms/views/bulk_action_views.py
@@ -10,13 +10,14 @@ This module contains the base view for bulk actions
 from __future__ import annotations
 
 import logging
+from collections import defaultdict
 from typing import TYPE_CHECKING
 
 from cacheops import invalidate_model
 from django.contrib import messages
 from django.contrib.auth.mixins import PermissionRequiredMixin
 from django.core.exceptions import PermissionDenied
-from django.db.models import Q
+from django.db.models import ProtectedError, Q
 from django.http import Http404
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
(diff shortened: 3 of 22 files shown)


integreat_cms/cms/templates/events/event_list.html
```diff
@@ -208,6 +208,17 @@ <h1 class="heading">
                         {% translate "Restore events" %}
                     </option>
                 {% endif %}
+                {% if perms.cms.delete_event %}
+                    <option class="bulk-confirmation-dialog"
+                            data-bulk-action="{% url 'delete_multiple_events' region_slug=request.region.slug language_slug=language.slug %}"
+                            data-popup-title="{% blocktranslate %}You are about to delete multiple selected {{content_type}}{% endblocktranslate %}"
+                            data-popup-subject="{% blocktranslate %}This action cannot be undone{% endblocktranslate %}"
+                            data-popup-text="{% blocktranslate %}Do you want to continue?{% endblocktranslate %}">
```
hannaseithe, line 216: I am very unsure about the phrasing here, so I am grateful for suggestions

integreat_cms/cms/views/bulk_action_views.py
```diff
@@ -518,3 +518,75 @@ def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
...
+        :param \**kwargs: The supplied keyword arguments
+        :return: The redirect
+        """
+        delete_sucessful = []
+        delete_failure = []
+        can_delete = True
+
+        for content_object in self.get_queryset():
+            if hasattr(content_object, "can_be_deleted"):
+                can_delete, error_msg = content_object.can_be_deleted()
```
MizukiTemma, line 551: `error_msg` may be re-witten by the next objects. 

Senario: First language was not deletable, the next one is deletable. `error_msg` is once set to the warning message but overwritten by an empty string by the second object. How to reproduce: choose English and Greek in Augsburg and bulk delete

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -3386,6 +3386,10 @@ msgstr "Anbieter für maschinelle Übersetzungen"
 msgid "Preferred provider for translations into this language"
 msgstr "Bevorzugter Anbieter für Übersetzungen in diese Sprache"
 
+#: cms/models/languages/language_tree_node.py
+msgid "it is the source language of other language(s)."
+msgstr "sie die Ausgangssprache für andere Sprachen ist."
```
MizukiTemma, line 3391:
```suggestion
msgstr "sie die Quell-Sprache für andere Sprachen ist."
```

"Quell-Sprache" is used for other cases of "source language"

integreat_cms/cms/models/pages/page.py
```diff
@@ -310,6 +310,16 @@ def relative_depth(self) -> int:
...
         return self.depth
 
+    def can_be_deleted(self) -> tuple[bool, str | None]:
+        """
+        checks if the page can be deleted
+        """
+        if self.children.exists():
+            return False, _("you cannot delete a page which has subpages.")
+        if self.mirroring_pages.exists():
+            return False, _("it was embedded as live content from another page.")
```
MizukiTemma, line 320: This can overwrite `error_msg` too.

How to reproduce: choose "Kontakt zu App Team Augsburg" and "Behörden" in Augsburg and bulk delete.

MizukiTemma (changes requested):
Thank you for tackling this big issue 💪 

I haven't yet finished the review completely, but will leave first comments already.

JoeyStk:
@hannaseithe When you were starting to work on this PR, bulk actions for push notifications didn't exist yet. By now they do exist and we would also need a bulk action for deleting push notifications. What do you think would be the best way to go about this? Should I add the bulk action to my PR, do you want to add it in this PR or should we wait until after this PR and then do it in a third PR? Which way would you prefer? I'm fine with all three options :)

MizukiTemma (approved):
Looks good 😸 Thank you so much for tackling this big issue 👍

tests/cms/views/pages/test_page_actions.py
```diff
@@ -0,0 +1,228 @@
...
+
+    deletable_pages = [
+        create_page(
+            "augsburg",
+            lft=2 * i + 1,
+            rgt=2 * i + 2,
+            depth=1,
+            tree_id=20,
+            name_add=f"-{i}",
+        )
```
PeterNerlich, line 192: Why exactly are you specifying the exact `lft`, `rgt`, `depth` and `tree_id` values again instead of using treebeards functions? If this is out of them being suspicious at first and you wanting to be explicit with what will end up in the database, I feel you. But I think this is effort spent in the wrong place – as soon as our test data grows to 20 tree structures, this will produce a collision with manually setting the `tree_id`. But we're not even testing at the level of values internal to treebeard here, just the pages involved being in the correct parent-child relationship with each other is sufficient. So throughout these tests I'm second guessing this choice here

tests/cms/views/poi/test_poi_form.py
```diff
@@ -101,6 +103,39 @@ def create_used_poi(region_slug: str) -> int:
...
 
 
+def create_unused_poi(region_slug: str, name_add: str = "") -> int:
+    """
+    A helper function to create a new POI that is unused (and therefore deletable)
+    """
+    region = Region.objects.filter(slug=region_slug).first()
+    poi_category = POICategory.objects.first()
+
+    used_poi = POI.objects.create(
```
PeterNerlich, line 113: You probably want this to be named `unused_poi`?
```suggestion
    unused_poi = POI.objects.create(
```

tests/cms/views/poi/test_poi_form.py
```diff
@@ -101,6 +103,39 @@ def create_used_poi(region_slug: str) -> int:
...
+        slug="ort" + name_add,
+        status=status.PUBLIC,
+        content="",
+        language=german_language,
+        poi=used_poi,
+    )
+
+    assert used_poi.events.count() == 0
+
+    return used_poi.id
```
PeterNerlich, line 136:
```suggestion
    assert unused_poi.events.count() == 0

    return unused_poi.id
```

integreat_cms/release_notes/current/unreleased/3122.yml
```diff
@@ -0,0 +1,2 @@
+en: Add multiple delete option to locations, events, language nodes, pages
+de: Füge mehrfache Löschoption zu Orten, Veranstaltungen, Sprachknoten, Seiten hinzu
```
PeterNerlich, line 2:
```suggestion
en: Add bulk delete option to locations, events, language nodes, pages
de: Füge Mehrfach-Löschoption zu Orten, Veranstaltungen, Sprachknoten, Seiten hinzu
```

hannaseithe (commented):
This is my own test review, to get my comments our of pending

Issue #3649, opened 2025-05-05
Internal Server Error when opening 'Potential contact data'


juliankehne:
### Describe the Bug

Trying to open the page for 'Analysis of content for existing contacts' leads to an Internal Server Error (on cms-test [Version: 2025.5.3-alpha](https://cms-test.integreat-app.de/testumgebung/release-notes/), working on production ([Version: 2025.5.0](https://admin.integreat-app.de/testumgebung/release-notes/)))

https://cms-test.integreat-app.de/testumgebung/contact/potential_targets

### Steps to Reproduce

1. Go to 'Contacts'
2. Click on 'Analysis of content for existing contacts'
3. See error

### Expected Behavior

No Error. Page opens correctly.

### Actual Behavior

Internal Server Error

### Screenshot

![Image](https://github.com/user-attachments/assets/b9e6aeed-b643-47d4-803a-2eef40cde5d3)

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

PR #3541, opened 2025-03-24
Refactor linkcheck utils

jarlhengstmengel:
### Short description

This PR refactors `linkcheck_utils.py`, mainly the `get_urls` method

### Proposed changes

- Reduce the things `get_urls` does and externalize some functionality to new methods
- Same for `filter_urls`
- Rename and extend the functionality of `get_region_links` to be able to handle multiple regions
- Exclude archived regions when multiple regions are requested

### Side effects

- This PR will also resolve #3458 since there was only one line of code that would have been changed after refactoring
- Otherwise everything revolving the linkchecker should work as always

### Resolved issues

Fixes: #3488 
Fixes: #3458

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #3651, opened 2025-05-05
Fix internal server error in suggest contacts

JoeyStk:
### Short description

This PR fixes the `Internal Server Error` on the page `potential-contacts` and in the management command to replace internal links. This bug was caused by #3541. The refactored `get_link_query` tries to access the first item within a Queryset of regions, if only a region is given this causes a `TypeError` which was represented by the `Internal Server Error`.

### Proposed changes

- But two occurrences of region in a list

### Side effects

- This probably needs a test, right?

### Resolved issues

Fixes: #3649

### DIFF
--- a/integreat_cms/cms/urls/protected.py
+++ b/integreat_cms/cms/urls/protected.py
@@ -1541,7 +1541,7 @@ urlpatterns: list[URLPattern] = [
                                 ),
                             ),
                             path(
-                                "potential_targets",
+                                "potential-targets",
                                 contacts.PotentialContactSourcesView.as_view(),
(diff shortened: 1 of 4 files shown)


MizukiTemma (approved):
Thank you for speedy fix!

JoeyStk:
@MizukiTemma / @jarlhengstmengel I added a test for the potential contacts view. Could you please have another look at this PR? :)

MizukiTemma:
> @MizukiTemma / @jarlhengstmengel I added a test for the potential contacts view. Could you please have another look at this PR? :)

Looks good 😸

Issue #3512, opened 2025-03-13
Restrict restoring pages to users that can archive pages


jonbulz:
### Describe the Bug

With changes introduced in #3491, users without publishing rights can no longer archive pages, but still have permission to restore pages that have been archived by other users. Following the principle of coherence of responsibility, users without the permissions to archive a page should not have the permission to restore an archived page.

### Steps to Reproduce

1. Log in with user role "Author"
2. Go to pages, see that user does not have permission to archive pages
3. Go to archived pages, see that user can restore pages

### Expected Behavior

Users who can't archive pages should not be able to restore archived pages

### Actual Behavior

Author users can still restore pages

### Related Issues

#3486

PR #3652, opened 2025-05-05
restrict restoring pages to users that have publishing permissions

jonbulz:
### Short description

Users without publish rights are not able to archive pages, but thus far have been able to restore pages. To adhere to the principle of coherent responsibility, users that can't archive pages should not be able to restore pages.

### Proposed changes

- check publish page permissions when restoring pages
- hide restore page icon in forms and tree views

### Side effects

- Due to an error in the page_tree_node, users with publish rights to specific pages (author role) were not able to archive these pages, even though they had the appropriate permissions, because the icon was always hidden. With this PR, icons for archiving and restoring are shown as expected

### Resolved issues

Fixes: #3512

### DIFF
--- a/integreat_cms/cms/templates/pages/page_form.html
+++ b/integreat_cms/cms/templates/pages/page_form.html
@@ -59,7 +59,6 @@
                     <button name="preview_page" type="button" data-preview-page data-edit-page-mode data-right-to-left={{ right_to_left }} class="btn btn-ghost">
                         {% translate "Preview" %}
                     </button>
-                    {% has_perm 'cms.publish_page_object' request.user page as can_publish_page %}
                     {% if can_publish_page %}
                         <button name="status"
                                 value="{{ DRAFT }}"
@@ -214,7 +213,7 @@
                 {% if page %}
                     <div id="right-sidebar-column"
                          class="md:w-full 3xl:col-end-3 4xl:col-end-auto flex flex-wrap flex-col">
-                        {% if can_edit_page %}
+                        {% if show_actions_box %}
                             {% include "./page_form_sidebar/actions_box.html" with box_id="page-actions" %}
                         {% endif %}
                         {% include "./page_form_sidebar/translator_view_box.html" with box_id="page-side-by-side" %}
--- a/integreat_cms/cms/views/pages/page_bulk_actions.py
+++ b/integreat_cms/cms/views/pages/page_bulk_actions.py
@@ -18,7 +18,7 @@ from ...utils.pdf_utils import generate_pdf
 from ...utils.stringify_list import iter_to_string
 from ...utils.translation_utils import gettext_many_lazy as __
 from ...utils.translation_utils import translate_link
-from ..bulk_action_views import BulkActionView, BulkArchiveView
+from ..bulk_action_views import BulkActionView, BulkArchiveView, BulkRestoreView
 from .page_actions import cancel_translation_process_ajax
 
 if TYPE_CHECKING:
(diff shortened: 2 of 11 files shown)


MizukiTemma (changes requested):
Thank you for the PR, it looks almost good 😸 

There is one case I would like you to fix: if you open an archived page as author, you'll find the "action" box but with no menu button. The whole box can be simply hidden in this case.

image

MizukiTemma:
> > There is one case I would like you to fix: when you open an archived page as author, you'll find the "action" box but with no menu button. The whole box can be simply hidden in this case.
> 
> I see. The solution I can think of right now is to check in `page_form.html` if the page is archived and then only show the actions_box if the user has publishing rights. Because for non-archived pages, also users without publishing rights have actions in their action box. That makes the logic in the template a bit more complicated and possibly confusing, though. Or do you have a better idea? :)


@jonbulz
Sorry for taking a long time to come back 🙏 
It probably makes the code more readable if we check the page status and publishing rights in `page_form_view.py` and rednder the result as `show_action_box` or soemhting 🤔

JoeyStk (changes requested):
Thank you very much! This looks mostly good to me and almost ready to be merged :rocket: I only have a few questions :)

integreat_cms/cms/templates/pages/page_form.html
```diff
@@ -214,7 +214,8 @@ <h1 class="heading">
                 {% if page %}
                     <div id="right-sidebar-column"
                          class="md:w-full 3xl:col-end-3 4xl:col-end-auto flex flex-wrap flex-col">
-                        {% if can_edit_page %}
+                        {% if can_edit_page and not page_form.instance.archived or can_publish_page %}
```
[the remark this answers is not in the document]
jonbulz: Yes, you are correct and yes, this is what we want. The reason is that users with editing but without publish permissions won't have any actions in the action box for archived pages. Thus, we want to hide the action box in that case. They do, however, have items in their action box on none-archived pages. In other words, we want to show the action box if:
the user has publish page permissions (because there will always be something in the action box, also in archived pages)
OR
the user has edit page permissions, but then we only want to show the box if the page is not archived

I will probably move this logic to ```page_form_view.py``` as by @MizukiTemma suggestion, though.

MizukiTemma (approved):
Thank you!

integreat_cms/cms/views/pages/page_bulk_actions.py
```diff
@@ -81,6 +81,15 @@ def get_permission_required(self) -> tuple[str]:
         return ("cms.publish_page_object",)
 
 
+class PageBulkRestoreView(PageBulkActionMixin, BulkRestoreView):
```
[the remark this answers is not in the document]
JoeyStk: Got it! Thank you for explaining :)

integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -530,3 +530,19 @@ def get_side_by_side_language_options(
                     },
                 )
         return side_by_side_language_options
+
+    def get_context_data(self, **kwargs: Any) -> dict[str, Any]:
+        context = super().get_context_data(**kwargs)
+        region = self.request.region
+        page_instance = region.pages.filter(id=kwargs.get("page_id")).first()
```
JoeyStk, line 537: Nit: I think it's better use .get() or even better use `get_object_or_404` instead of filter().first() :)

JoeyStk (approved):
Thank you very much for your changes! I think this is ready to be merged now :) :rocket:

integreat_cms/cms/views/pages/page_form_view.py
jonbulz: Thanks, yes, I think it does make sense to raise 404 if no page is found here. Could you just quickly confirm the most recent changes cover what you meant? Then I'll squash the commits and merge

jonbulz: Oh, I now remember what the problem with that was (thanks tests for failing rightfully). :sweat_smile: 
When creating a new page, this will raise an exception, as the page is not yet in the database. We could do .get and then handle the DoesNotExist exception, but in this case I'd actually prefer sticking to .filter(...).first(). Is that okay for you @JoeyStk

JoeyStk: Yes, I suppose that's fine :)

Issue #2133, opened 2023-03-21
Prevent combination of inactive but visible language tree nodes


timobrembeck:
### Motivation

Languages are not shown in the app when they are inactive but visible (because of [this](https://github.com/digitalfabrik/integreat-cms/blob/20d2dc7e667472c194e3200cf2fbd6225d1140b1/integreat_cms/cms/models/regions/region.py#L522) filter).
This is the intended behavior (after all, it makes no sense to show content in the app which cannot be edited in the cms), but in order to avoid confusion, I'd suggest to prevent this combination in the language form.

### Proposed Solution

Add a database constraint which disables this combination and migrate all languages with this existing combination to make them invisible.

### Alternatives

Keep it the way it is (there weren't any complaints yet I think)

hannaseithe:
I am wondering if we should also automatically uncheck and disable the checkbox for "visible"  in the form for creating a new language tree node, if "active" is unchecked? @nikolahoff . 

Right now I have implemented it so far, that you are still able to select in the form the combination **active=>unchecked & visible => checked**. But if you try to create such a language tree node it fails and an error warning from the server is shown.

PR #3658, opened 2025-05-07
Add constraint for Language Tree Node

hannaseithe:
### Short description

This adds a constraint to the Language Tree Node model, so that creating a node that is both inactive and visible is not possible anymore. 

### Proposed changes

- the constraint is added in the model
- migrations both for the new constraint and mutating possible nodes with active=False & visible=True to visible=False were added
- a test for the constraint was created

### Side effects

- no side effects

### Resolved issues

Fixes: #2133


hannaseithe:
I am wondering if we should also automatically uncheck and disable the checkbox for "visible"  in the form for creating a new language tree node, if "active" is unchecked? @nikolahoff . 

Right now I have implemented it so far, that you are still able to select in the form the combination **active=>unchecked & visible => checked**. But if you try to create such a language tree node it fails and an error warning from the server is shown.

nikolahoff:
@hannaseithe yes, as you described the behaviour makes sense to me.

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -10050,6 +10058,30 @@ msgstr "aktiviert"
...
+msgid_plural ""
+"The following {model_name} could not be deactivated, because inactive "
+"language tree nodes cannot be visible: {object_names}"
+msgstr[0] ""
+"{model_name} {object_names} konnte nicht deaktiviert werden, da die Quell-"
+"Übersetzung fehlt."
+msgstr[1] ""
+"Die folgenden {model_name} konnten nicht deaktiviert werden, da ihre Quell-"
+"Übersetzungen fehlen: {object_names}"
+
```
MizukiTemma, line 10084: Maybe a copy&paste error? The translation for the reason must be adjusted 😅

MizukiTemma (changes requested):
Looks good! There is only two things to be adjusted 🇩🇪

integreat_cms/locale/de/LC_MESSAGES/django.po
hannaseithe: Thank you for the CR!

integreat_cms/cms/templates/languagetreenodes/languagetreenode_form.html
```diff
@@ -21,7 +21,8 @@ <h1 class="heading overflow-hidden text-ellipsis">
             {% if perms.cms.change_languagetreenode %}
                 <div class="flex flex-wrap gap-2">
                     <a href="{% url 'languagetreenodes' region_slug=request.region.slug %}"
-                       class="btn btn-ghost">{% translate "Cancel" %}</a>
+                       class="btn btn-ghost">{% translate
+                    "Cancel" %}</a>
```
MizukiTemma, line 25: This made the translation broken and resulted in such appearance of the button

image


```suggestion
                       class="btn btn-ghost">{% translate "Cancel" %}</a>
```

MizukiTemma (changes requested):
Thank you for updates 💪 

There is one thing left in the bulk action to fix: making inactive languages visible by bulk action leads to an error.

hannaseithe:
> Sorry for requesting changes again and again. I didn't find all points to fix in the last review 🙈
Dont feel bad for being thorough! That way I learn and we get good quality code and releases. I genuinely appreciate it :smile_cat:

JoeyStk (approved):
Thank you very much for your improvements! To me this looks ready to be merged :rocket:

integreat_cms/cms/migrations/0135_languagetreenode_language_tree_node_inactive_requires_invisible.py
```diff
@@ -0,0 +1,32 @@
...
+def fix_inactive_visible_nodes(
+    apps: Apps, _schema_editor: BaseDatabaseSchemaEditor
+) -> None:
+    LanguageTreeNode = apps.get_model("cms", "LanguageTreeNode")
+    LanguageTreeNode.objects.filter(active=False, visible=True).update(visible=False)
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("cms", "0134_alter_region_status"),
```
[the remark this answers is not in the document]
hannaseithe: I am very much looking forward to this ... not :D But thank you for the reminder!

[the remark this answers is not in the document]
JoeyStk: It can be daunting for sure! I hope that no inconveniences come up :crossed_fingers:

MizukiTemma:
> Sorry for requesting changes again and again. I didn't find all points to fix in the last review 🙈
> 
> * Migration fails, if there is any language node in the DB that is inactive but vidible, as it can fulfill the new constraint 😿
>   How to reproduce:
> 
> 1. Check out to the branch develop
> 2. ./tools/run.sh and make a language node inactive but visible
> 3. Check out to the branch of this PR
> 4. ./tools/run.sh (without ./tools/prune_database.sh)
> 5. See django.db.utils.IntegrityError
> 
> As we cannot guarantee that there is no such language node at the moment of deployment, I suggest to add a function to deal with such a case.
> 
> [This migration](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/migrations/0138_configurable_mt_budget.py) would be an example for that. Write a function that turns all inactive and visible language nodes into inactive and invisible, like `migrate_mt_addon_booked` in the exmaple and apply this before `migration.AddConstraint`. This way it should work (but I didn't try by myself 🙈 sorry, if it won't go well).
> 
> * There was once case found when a bulk action faces an user-unfriendly error:
> 
> 1. Make a language node inactive and invisible
> 2. Make the language node visible by bulk Action
> 3. You'll face this
> 
> image

@hannaseithe 
Seems this is not yet fixed? 👀

hannaseithe:
@MizukiTemma / @JoeyStk

For Mizuki's change request I modified the BulkUpdateBooleanFieldView's post request to catch the Integrity Error directly and create an error message (I deleted my changes to BulkDisableView's post request as this will also be covered in the super class). When doing so I ran into an TransactionManager Error, which was caused by us wrapping LanguageTreeBulkActionView in an atomic transaction (with the @tree_mutex() decorator). After looking into it I am quite confident that the wrapper is unnecessary here, since we do not call any treebeard methods from this view (we only change the values of boolean fields) - that is why I have deleted the decorator and handle the IntegrityError with a simple try block inside BulkUpdateBooleanFieldView's post method. 

I would kindly ask you both to review the changes.

@PeterNerlich also mentioning you here since we talked about whether the decorator can be deleted or not

PeterNerlich:
Just giving my two cents here: Over the past we have learnt to be very cautious of anything related to treebeard and that we rather want to accept some inefficiencies that we are able to just throw more computing resources on than potentially opening up more holes where treebeard could come back to bite us. Although *(or especially because)* the exact mechanism is not understood by me – and you assert that treebeard does not overwrite any of the save, update etc. methods or use signals – we have observed treebeard sporadically performing big changes that go far beyond the item to save itself, silently wreaking havoc when used in a transaction that got rolled back *(but treebeards changes persisting back then)*, possibly involving outdated data in the cache or django memory, or just concurrent treebeard operations. After #3076 this is still not fully resolved *(though my personal hope is that it got less)*.

If we can get away with leaving the decorator in out of caution, I would prefer that. [According to everything I'm reading online](https://stackoverflow.com/a/69127707) it should be possible to roll back only part of an atomic transaction as you do. I don't understand where the `TransactionManagementError` is coming from

hannaseithe:
Ok I have implemented it now with the decorator in place. This is the version I came up with together with @JoeyStk. 

There is an alternate version to be found here: https://github.com/digitalfabrik/integreat-cms/tree/enhancement/prevent-inactive-and-visible-ltn-alt . This also keeps the decorator, but sets the savepoint and handles the IntegrityError inside `LanguageTreeBulkActionView`. But after learning that setting a `with transaction.atomic():` context around a single update doesnt make a difference when there is no decorator (which could happen in the future if another class than `LanguageTreeBulkActionView` inherits from `BulkUpdateBooleanView`), I prefer the solution presented here.

Now its really ready for review :)

MizukiTemma:
@hannaseithe 

> I prefer the solution presented here.

Which one is it? Does "here" menas  the branch `enhancement/prevent-inactive-and-visible-ltn` or `enhancement/prevent-inactive-and-visible-ltn-alt`?

hannaseithe:
> @hannaseithe
> 
> > I prefer the solution presented here.
> 
> Which one is it? Does "here" menas the branch `enhancement/prevent-inactive-and-visible-ltn` or `enhancement/prevent-inactive-and-visible-ltn-alt`?

I meant the branch `enhancement/prevent-inactive-and-visible-ltn` . This is what I prefer

Issue #3276, opened 2024-12-09
Incorrect style for "Restore as draft" button in version history of external calendar event


juliankehne:
### Describe the Bug
The 'Restore as draft' button the version history of an external calendar event is not fully disabled/greyed out on hover the style changes.

![2024-12-09 11_06_51-Integreat Editorial System — Mozilla Firefox](https://github.com/user-attachments/assets/1a443a9f-68a8-47fd-902c-67830205df78)
![2024-12-09 11_06_58-Window](https://github.com/user-attachments/assets/183ac90b-17a3-4755-a08a-17d01a41262c)

### Steps to Reproduce

1. Import an external calendar you own
2. Change an event in your calendar tool
3. Reimport the external calendar in the cms
4. (Alternative to Step 1-3) Use an existing imported event that has a version history with at least two versions)
5. Go to 'Events'
6. Open an external calendar event with multiple versions
7. Open the Version history
8. Click on the previous version of the event
9. See 'Restore as draft' button with wrong style

### Expected Behavior

Button has the correct style for being disabled (White text, no blue outline, style doesn't change when hovering over it)

(Same as 'Restore and Publish' button)

### Actual Behavior

Blue text, blue outline and style changes when hovering over it.

Style of the button is not correct

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

### Related Issues

https://github.com/digitalfabrik/integreat-cms/pull/3127

PR #3661, opened 2025-05-09
Fix style of disabled outline button

JoeyStk:
### Short description

This PR fixes the style of disabled outline buttons. They look now like normal disabled buttons.

### Proposed changes

- Apply button classes also for disabled outline buttons

### Side effects

I expect none

### Resolved issues

Fixes: #3276 


MizukiTemma (approved):
Looks good!

JoeyStk:
>I think we should change this part (along with line 186, where the background for any hovered disabled button is changed). From my understanding a disabled button should not have any style changes on hover, even if subtle (as Julian also points out in the original issue). But that is also a bit nitpicky, so feel free to ignore, if you dont want to.

@hannaseithe Thank you for your review and suggestion. Sounds good. It's applied :)

hannaseithe (approved):
Looks great, thank you

Issue #3049, opened 2024-09-14
Region deletion times out


david-venhoff:
### Describe the Bug

Deleting a region is much slower than it should be, taking multiple minutes and often times out

### Steps to Reproduce
1. Use the database dump
2. Delete a region with a lot of content
4. Notice timeout

### Expected Behavior

Region deletion should not take much time

### Actual Behavior

Region deletion takes a lot of time

MizukiTemma:
It seems cascade deleting is taking too long and causing the time out

https://github.com/digitalfabrik/integreat-cms/blob/2a40627f82495ce4587dce640620904b6c187495/integreat_cms/cms/views/regions/region_actions.py#L108-L110


Does anyone have an idea what we can do?

charludo:
> It seems cascade deleting is taking too long and causing the time out
> 
> [integreat-cms/integreat_cms/cms/views/regions/region_actions.py](https://github.com/digitalfabrik/integreat-cms/blob/2a40627f82495ce4587dce640620904b6c187495/integreat_cms/cms/views/regions/region_actions.py#L108-L110)
> 
> Lines 108 to 110 in [2a40627](/digitalfabrik/integreat-cms/commit/2a40627f82495ce4587dce640620904b6c187495)
>  with disable_listeners(): 
>      # Delete region and cascade delete all contents 
>      deleted_objects = region.delete() 
> 
> Does anyone have an idea what we can do?

We could add a celery task for the deletion instead... That would probably avoid the timeout, but the regoin still exists when the page is loaded after hitting "delete", and will only disappear a while later.

Maybe that is OK? But it's not nice...

Ooooor we could just disable the timeout for that endpoint 😶 (But not sure if that's to be done in the CMS or on the prod server? In any case, not really nice either)

charludo:
Alternatively, we could get rid of `on_delete=models.CASCADE` for fields referencing regions, changing it to `models.SET_NULL` instead. Then the region deletion would be instant, but we would need to add celery tasks for deleting all objects refering `NULL` regions.

Potentially, this could be the best solution, but I have no idea if that adds bad side effects.

MizukiTemma:
@charludo 
Thank you for the ideas 😸

>We could add a celery task for the deletion instead... That would probably avoid the timeout, but the regoin still exists when the page is loaded after hitting "delete", and will only disappear a while later.

[...]

PR #3666, opened 2025-05-09
Delete region with Celery

MizukiTemma:
### Short description

This PR implements region deletion as Celery task in the same way with #3579 

### Proposed changes

- Exctract deletion part into another function (Celery task)
- Adjust the message that appears after clicking the delete button
- Add a new info message that the deletion can take several minutes
- Adjust the affected test

### Side effects

- Should be none: no change in the logic and conditions

### Resolved issues

Fixes: #3049 

### DIFF
--- a/tests/cms/views/regions/delete_region.py
+++ b/tests/cms/views/regions/delete_region.py
@@ -1,3 +1,5 @@
+import time
+
 import pytest
 from django.conf import settings
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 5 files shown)


tests/cms/views/regions/delete_region.py
```diff
@@ -70,11 +72,15 @@ def test_delete_all_regions_is_successful(
...
-                "utf-8",
+            assert (
+                "Dieser Vorgang kann mehrere Minuten dauern."
+                in response.content.decode(
+                    "utf-8",
+                )
             )
 
     if role in [CMS_TEAM, SERVICE_TEAM, ROOT]:
+        time.sleep(10)
```
svenseeberg, line 83: Maybe the speed and also the robustness of the test could be improved by having a flexible wait:
```suggestion
        for n in range(0, 30):
            time.sleep(1)
            if Region.objects.count() == 0:
                break
```

svenseeberg (approved):
Thank you :tada: I have not tested this but the Celery implementation looks correct :rocket:

JoeyStk:
Is there a way to test this? :)

MizukiTemma:
I testes with the local data:
- Deletion of a region with contents referenced by other regions is stopped with the warning message
- Deletion of a region not referenced by other regions works successfully

whether the performance is really okay, can be only checked after merging into the test system 🤔 We can probably test with dump data, but conqurrence with other celery related functions and normal user requests cannot be tested anyway.

Issue #3669, opened 2025-05-12
Fix failing test_find_missing_versions_failure


svenseeberg:
### Describe the Bug

FAILED tests/core/management/commands/test_find_missing_versions.py::test_find_missing_versions_failure - AssertionError: assert equals failed

### Steps to Reproduce

1. https://app.circleci.com/pipelines/github/digitalfabrik/integreat-cms/15375/workflows/49d4d715-d945-4b64-9c66-4a64355fe901/jobs/206798?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
load_test_data = None

    @pytest.mark.django_db
    def test_find_missing_versions_failure(load_test_data: None) -> None:
        """
        Ensure that inconsistencies are listed
        """
        # Create version inconsistency
        page_translation = Page.objects.get(id=1).get_translation("de")
        page_translation.version += 1
        page_translation.save()
        out, err = get_command_output("find_missing_versions", "page")
        assert (
            out
            == "\x1b[0;34mChecking the model PageTranslation for version inconsistencies...\x1b[0m\n"
        )
>       assert err == (
            "\x1b[0;31mThe latest version of <Page (id: 1, region: augsburg, slug: willkommen)>"
            " in <Language (id: 1, slug: de, name: German)>"
            " is 3, but there are only 2 translation objects!\x1b[0m\n"
        )
E       AssertionError: assert equals failed
E         '\x1b[0;31mThe latest version o  '\x1b[0;31mThe latest version o 
E         f <Page (id: 1, region: augsbur  f <Page (id: 1, region: augsbur 
E         g, slug: w^el^come)> in <Language  g, slug: w^il^lkom+me+n)> in <Langu 
E          (id: 1, slug: de, name: German  age (id: 1, slug: de, name: Ger 
E         )> is 3, but there are only 2 t  man)> is 3, but there are only  
E         ranslation objects!\x1b[0m\n'    2 translation objects!\x1b[0m\n 
E                                          '
  ```

</details>

PR #3671, opened 2025-05-12
Fix test_find_missing_versions_failure

svenseeberg:
### Short description
It seems that some other previously running test manipulates the slug of version 1 of the augsburg/de/willkommen page. In order for this test to successfully complete, we need to ensure that the slug is correct.

### Proposed changes
- Set the slug of version 1 to `willkommen` before the test starts.

### Side effects
- Some other COULD depend on the slug being `welcome`? But I doubt that currently.

:100: :joy: 

### Resolved issues

Fixes: #3669 

### DIFF
--- a/tests/core/management/commands/test_find_missing_versions.py
+++ b/tests/core/management/commands/test_find_missing_versions.py
@@ -47,6 +47,11 @@ def test_find_missing_versions_failure(load_test_data: None) -> None:
     """
     # Create version inconsistency
     page_translation = Page.objects.get(id=1).get_translation("de")
+
+    # Ensure that the title is as expected in the output later. Some other test seems
+    # to change the title, which results in flaky CI tests.
+    page_translation.slug = "willkommen"
+
     page_translation.version += 1
     page_translation.save()
     out, err = get_command_output("find_missing_versions", "page")
@@ -59,3 +64,6 @@ def test_find_missing_versions_failure(load_test_data: None) -> None:
         " in <Language (id: 1, slug: de, name: German)>"
         " is 3, but there are only 2 translation objects!\x1b[0m\n"
     )
+    # restore previous state to not interfere with other tests.
+    page_translation.version -= 1
+    page_translation.save()


JoeyStk (approved):
Thank you! LGTM

PeterNerlich:
Rather than changing it in the test, wouldn't it be cleaner to change it in djangos test data fixture?

tests/core/management/commands/test_find_missing_versions.py
```diff
@@ -47,6 +47,8 @@ def test_find_missing_versions_failure(load_test_data: None) -> None:
     """
     # Create version inconsistency
     page_translation = Page.objects.get(id=1).get_translation("de")
+    page_translation.slug = "willkommen"
```
PeterNerlich, line 50: A comment would be nice. Thrown in a cheeky one for good measure
```suggestion
    # Prevent flaky test fails from the slug being different than expected
    # Yes this is a bad way to do this, feel free to rewrite our entire test suite yourself
    page_translation.slug = "willkommen"
```

tests/core/management/commands/test_find_missing_versions.py
```diff
@@ -47,6 +47,8 @@ def test_find_missing_versions_failure(load_test_data: None) -> None:
...
+    page_translation.save()
```
PeterNerlich, line 51: We don't need to save it twice, do we?
```suggestion
```

PeterNerlich (approved):
I can't find it anywhere as well in the tests, neither the string `welcome` nor where `.save(` is called. So I only have two nitpicks

tests/core/management/commands/test_find_missing_versions.py
svenseeberg: We do. Because we need to override version one, not the next version.

PeterNerlich: Then I'm not sure I understand the original problem. Here is my brain dump for what my understanding was until now:

- The page with `id=1` has $n \geq 1$ German translations associated with it
- Those have unique integers in `version` and can be ordered by that key without skipped numbers, always starting at 1
*(the translation with the highest `version` per page-language pairing is henceforth called the newest translation)*
- They each have a `slug`. Any page-language pairing for which a translation exists is identified only by the slug of the
- `page.get_translation("de")` gives us the newest translation for this specific page-language pairing
- The original test takes the newest translation and modifies its `version` to its original value plus one. This effectively introduces a gap in the implied ordering expected to start at 1. The test proceeds with asserting whether a management command reports this fact
- The assertion relies on the exact string content of the message. The reference value is hard-coded in the test and contains the slug representing page-language pairing, i.e. the slug of the newest translation
- The assertion fails because the slug of the newest translation doesn't match the one in the hard-coded reference value. This is despite being correct in the test data fixture and being changed seemingly nowhere in the test suite
- Your change is to update the slug back to the expected one at the start of the test. If we assume it to start with `(id=5, slug="welcome", version=4)`, your change would update it to `(id=5, slug="willkommen", version=4)`, followed by an update to `(id=5, slug="willkommen", version=5)` by the code of the original test, in which case the intermediate `.save()` is redundant

svenseeberg: You're right. I assumed that the second save() would create a new version, but it doesnt. I'll investigate further.

svenseeberg:
Rebased on current develop.

svenseeberg:
It seems that this patch fixed the "old" develop branch, the current develop branch and also #3677 / #3658. I think this is worth merging.

PeterNerlich:
@svenseeberg  What is the state of this PR – do we just need to merge (or re-review changes) or did this turn out to be not effective after all and to need further investigation?

svenseeberg:
Its hard to tell. I personally would merge it and see what happens. The changes should not be harmful in any case.

PR #3675, opened 2025-05-14
Fix timeout during loading central linkchecker

jarlhengstmengel:
### Short description

After merging #3654 and testing it on the test server, it was discovered that the central linkchecker times out. Probably because the introduced query to filter for active regions can get too big together with query to filter for the urls (it basicly filters more or less the same objects twice)

### Proposed changes

- Filter for active regions directly in `get_link_query` without an additional query that gets evaluated

### Side effects

- Hopefully none

### Resolved issues

Fixes: # n.a.


MizukiTemma (approved):
Hmm, I'm not sure whether it's fast enough...Let's try 😅

JoeyStk (approved):
I'm also a little bit skeptical, but it's worth a try! :rocket:

Issue #3679, opened 2025-05-19
News and Archived News (Push Notifications) are not separated in the search


juliankehne:
### Describe the Bug

When searching through Push Notifications in the cms you currently get results for all push notifications (not archived & archived).

In News you get search results for archived news.
In Archived News you get search results for non archived news.

### Steps to Reproduce

1. Go to 'News'
2. Archive a news object (If there is none, create a new one and archive it)
3. Search for the archived name
4. See that it also shows up in the search results

The other way around for 'Archived News'

5. Go to 'Archived News'
6. Restore the archived news object
7. Search for the restored name
8. See that it also shows up in the search results

### Expected Behavior

Only the correct results are shown. Either archived news or non-archived news.

### Actual Behavior

All news show up in the search field.

### Screenshot

![Image](https://github.com/user-attachments/assets/ebc4ff3d-9984-4b08-b19e-27cdf2a39a0d)

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

### Related Issues

https://github.com/digitalfabrik/integreat-cms/issues/3588
https://github.com/digitalfabrik/integreat-cms/pull/3641

PR #3682, opened 2025-05-19
Separate archived and not-archived push notifications in the search suggestion

MizukiTemma:
### Short description

This PR fixes the bug that the status of push notifications (archived or not) was not considered in the search suggestion

### Proposed changes

- Filter by archived status

### Side effects

- Should be none
- 

### Resolved issues

Fixes: #3679 

### DIFF
--- a/integreat_cms/cms/templates/push_notifications/push_notification_list.html
+++ b/integreat_cms/cms/templates/push_notifications/push_notification_list.html
@@ -33,7 +33,11 @@
         <div class="flex flex-wrap justify-between gap-4">
             <div class="flex flex-wrap gap-4">
                 {% include "generic_language_switcher.html" with target="push_notifications" %}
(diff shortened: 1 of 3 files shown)


JoeyStk (approved):
Oh, that makes a lot of sense! Didn't think of it. Thank you very much! Usually we want a test for bugs - but in this case we don't need one, because I had just forgotten to implement this, right? :)

JoeyStk:
Just out of curiosity: Why did you remove the release note? :thinking:

MizukiTemma:
@JoeyStk 
> Just out of curiosity: Why did you remove the release note? 🤔

Because the archive feature of PNs itself is not yet released to users :)

hannaseithe (approved):
Thank you! The search results show now as expected.

PR #3684, opened 2025-05-20
Add to documentation handling pyhon3.12 during installation

hannaseithe:
### Short description
I ran into this problem with python 3.12 during development environment setup on my Ubuntu laptop. These additions to the documentation might be helpful for the next person

### Proposed changes
- install.sh needs python 3.9 to 3.11 (3.12 fails)
- so best way is to install python3.11 alongside, but keep python3 pointing at 3.12
- end then run install.sh by setting the python parameter to python 3.11

### Side effects
none

### Resolved issues

Fixes: #


JoeyStk (approved):
Thank you very much! :tada:

MizukiTemma (approved):
Thank you so much :)

Me too came to the idea of adding this hint into the documentation yesterday as I set up a new laptop with Python 3.12 😹

Issue #3673, opened 2025-05-13
Save Page Based statistics to Page Model


jarlhengstmengel:
### Motivation

#3325 introduced page based statistics into the CMS. During testing it turned out that the requests to Matomo are too big and crashes the matomo server when data for too much pages are requested. To solve this, it was proposed to save load the page access statistics into our own data structure to heavily reduce loading times for the municipalities and avoiding crashing the matomo server.

### Proposed Solution

- [x] Add Page Accesses to its own Model
- [x] Add Method to Region for loading Page Accesses into a regions pages when statistics are enabled
- [ ] Add method to save PageAccess to database
- [ ] Alter `get_page_accesses_ajax` in `statistics_actions` to accommodate loading from the database

### Alternatives

Caching was also considered (see #3634) but since loading page accesses for a longer time frame then 30 days is one of the biggest use cases, not really a lasting solution.

### User Story

As a user of the CMS I want to see the number of accesses to individual pages in a hierarchical way so that I know which content is more generally relevant for the majority of users. The time frame of the accesses needs to be long enough to enable analyzing accesses for the last year.

### Additional Context

#3674 will handle to load page accesses regularly from the matomo server via celery

### Related Issues

This Issue is part of the Meta Issue #1097

PR #3688, opened 2025-05-21
Add page accesses model

jarlhengstmengel:
### Short description

This PR adds a model for page based statistics.

### Proposed changes

- Adds PageAccesses model
- Adds method to region to get page accesses for requested languages

### Side effects

- Hopefully none
- 

Innitial Idea was to add page accesses to the page model, but page accesses also contains information about the language that was accessed. So it needs to have it's own table. This partially solves #3673

### Resolved issues

Fixes: 

### DIFF
--- a/integreat_cms/cms/migrations/0133_add_page_accesses.py
+++ b/integreat_cms/cms/migrations/0133_add_page_accesses.py
@@ -1,0 +1,71 @@
+import django.db.models.deletion
+import django.utils.timezone
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    """
+    Initial migration file for the Page Accesses model
+    """
+
+    dependencies = [
+        ("cms", "0132_archive_pushnotification"),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name="PageAccesses",
+            fields=[
+                (
+                    "id",
+                    models.BigAutoField(
+                        auto_created=True,
+                        primary_key=True,
+                        serialize=False,
+                        verbose_name="ID",
+                    ),
+                ),
+                (
+                    "access_date",
+                    models.DateTimeField(verbose_name="Date of the page accesses"),
+                ),
+                (
+                    "language",
+                    models.ForeignKey(
+                        on_delete=models.CASCADE,
+                        to="cms.language",
+                        verbose_name="Languages of the page that was accessed",
+                    ),
+                ),
+                (
+                    "page",
+                    models.ForeignKey(
+                        on_delete=models.CASCADE,
+                        to="cms.page",
+                        verbose_name="Accessed page",
+                    ),
+                ),
+                (
+                    "accesses",
+                    models.IntegerField(
+                        validators=[django.core.validators.MinValueValidator(0)],
+                        verbose_name="Page accesses",
+                    ),
+                ),
+            ],
+            options={
+                "default_permissions": ("change", "delete", "view"),
+                "default_related_name": "page_accesses",
+                "ordering": ["pk"],
+                "verbose_name": "page accesses",
+                "verbose_name_plural": "page accesses",
+            },
+        ),
+        migrations.AddConstraint(
+            model_name="PageAccesses",
+            constraint=models.UniqueConstraint(
+                fields=("page", "language", "access_date"),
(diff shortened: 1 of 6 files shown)


integreat_cms/cms/models/statistics/page_accesses.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    Data Model representing the accesses of a page
+    """
+
+    access_date = models.DateTimeField(
+        verbose_name="Date of the page accesses",
+    )
+    language = models.ForeignKey(
+        "cms.Language",
+        on_delete=models.CASCADE,
```
hannaseithe, line 20: Just a question about cascading: Is this a general policy that we cascade-delete to keep our database clean? One could make the case that a statistic might still be of interest even if language or page dont exist anymore, but since this table is re-filled by celery regularly (correct?) thats an overkill I assume. In Matomo the data still persists, right?

integreat_cms/cms/migrations/0133_add_page_accesses.py
```diff
@@ -0,0 +1,71 @@
...
+                (
+                    "accesses",
+                    models.IntegerField(
+                        validators=[django.core.validators.MinValueValidator(0)],
+                        verbose_name="Page accesses",
+                    ),
+                ),
+            ],
+            options={
+                "default_permissions": ("change", "delete", "view"),
```
hannaseithe, line 57: Question: The Celery process already has add permissions I assume? (Sorry for my half-knowledge)

hannaseithe (changes requested):
This was a fun Code Review, I think I learned quite a new few things - Especially working with the shell!

To me this looks all good:
1. Migration works
2. Created a page_access object in the shell -> works as expected
3. Tested the unique contraint in the shell -> failed as expected

I have one change request though:
I think integreat_cms > cms > models >__init__.py  should import the page_access model, if I understand correctly. If I am mistaken please let me know why?

integreat_cms/cms/models/regions/region.py
```diff
@@ -1038,6 +1039,29 @@ def last_content_update(self) -> datetime:
...
 
+    def get_page_accesses_by_language(
+        self,
+        pages: list[Page],
+        start_date: date,
+        end_date: date,
+        languages: list[Language],
+    ) -> dict:
+        """
+        Get the page accesses of the requested pages of this region during the specified timerange
```
JoeyStk, line 1050:
```suggestion
        Get the page accesses of the requested pages of this region during the specified time range
```

integreat_cms/cms/models/statistics/page_accesses.py
```diff
@@ -0,0 +1,44 @@
...
+    )
+    page = models.ForeignKey(
+        "cms.Page",
+        on_delete=models.CASCADE,
+        verbose_name="Accessed page",
+    )
+    accesses = models.IntegerField(
+        validators=[MinValueValidator(0)], verbose_name="Page accesses"
+    )
+
```
JoeyStk, line 31: At the moment there is no __str__() method. Therefore the objects of this model appear as "PageAccesses object (1)" in the Django Admin. I think it might be helpful to have a __str__() method to have a more speaking name. This could look something like this 

```suggestion
def __str__(self):
        return f"{self.page} - Accesses: {self.accesses}, language: {self.language}, date: {self.access_date}"
```

JoeyStk (approved):
Thank you very much for your work! :rocket: This is a really good step forward!
This looks good to me, I just noticed two minor things :)

integreat_cms/cms/models/statistics/page_accesses.py
jarlhengstmengel: I think we delete very litle. The municipalities are only allowed to archive nearly everywhere. When things are deleted then by us or the service team. At that point it's quite certain stuff is not needed anymore and we can cascade. But I had that thought too, yes

integreat_cms/cms/migrations/0133_add_page_accesses.py
jarlhengstmengel: I think it does, but not too certain

hannaseithe (approved):

Issue #3655, opened 2025-05-07
Alias names for districts are not filled in a region


dkehne:
### Describe the Bug
Alias names for regions are not filled in automatically

### Steps to Reproduce

1. Go to Region Settings e.g. Landkreis Ammerland or StädteRegion Aachen
2. Scroll down to 'Alias'
3. See empty field

### Expected Behavior
After adding the Community Identification Number in the region settings for a district, normally the alias names should be filled after saving. 

### Actual Behavior
This is currently not happening. At least for 14 month not working...

### Screenshot
Correct: 

Not Correct:

### Additional Information
One requirement that those aliases are created is the Community Identification Number which is used for API call to gvz.integreat-app.de. But this seems to work: May 07 09:03:48 INFO    integreat_cms.gvz_api.apps - GVZ API is available at: 'https://gvz.integreat-app.de'

PR #3689, opened 2025-05-21
Fix GVZ API aliases

JoeyStk:
### Short description

This PR fixes the aliases for GVZ

### Proposed changes

- Load api_available via cache

### Side effects

api_available was always False and therefore the code wasn't executed. We think this issue was because `api_available` wasn't stored anywhere. Our idea was to store it in the cache to have it saved somewhere. 

### Resolved issues

Fixes: #3655 

### DIFF
--- a/integreat_cms/core/settings.py
+++ b/integreat_cms/core/settings.py
@@ -267,6 +267,11 @@ GVZ_API_ENABLED: Final[bool] = True
 #: (see :mod:`~integreat_cms.gvz_api` for more information).
 GVZ_API_URL: Final[str] = "https://gvz.integreat-app.de"
 
+#: The Time To Live for caching the availibility status of the API in seconds.
+#: ``None`` will cache the status forever (until the server is restarted), ``0`` will invalidate it immediately.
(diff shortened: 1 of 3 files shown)


MizukiTemma:
Thank you for the PR 😸 

I have tested with the data of StädteRegion Aachen. `apps.get_app_config("gvz_api").api_available` is now `True` and the block is run, but it seems `gvz_region.child_coordinates` is empty and `aliases` is not filled 🤔

JoeyStk:
> I have tested with the data of StädteRegion Aachen. apps.get_app_config("gvz_api").api_available is now True and the block is run, but it seems gvz_region.child_coordinates is empty and aliases is not filled 🤔

Thank you for testing! To me it looks as if there is an incorrect Community identification number in our test system. If you use the right one, it works for me :)

MizukiTemma (approved):
Thank you so much 👍

jarlhengstmengel (approved):
Looks good, with the right numbers it works (I made the same mistake as Mizuki :laughing: ). Thank you!

Issue #3687, opened 2025-05-21
SVGs are not delivered as SVGs


steffenkleinle:
### Describe the Bug

It seems like SVGs uploaded and added to the CMS media library are not kept and delivered as SVGs and instead at some step (possibly the upload?) converted to PNGs. This only concerns the file content as the file name stays `.svg`.
While the web app seems to successfully handle this, the native app breaks and is not able to disable those icons.
Only tested for 1st level icons atm.

### Steps to Reproduce

1. Go to https://webnext.integreat.app/augsburg-svg/de
2. Inspect a 1st level icon (e.g. blob:https://webnext.integreat.app/f1fd1c5f-1040-4cfc-b111-6d40821d154b)
3. See browser error `XML Parsing Error: not well-formed`
4. See the network tab and download one of the .svgs
5. Print the contents (`cat accommodation.svg`) and see PNG content

### Expected Behavior

The content of svgs should also contain svgs.

### Actual Behavior

The content of svgs contain pngs.

### Screenshot

N/A

### Additional Information

<details>
  <summary>SVG content</summary>

  ```
�PNG

IHDR,y}�uRxIDATx��w�$Uտ�[U��'����]`3�L�DI��П���_Ā�JPQE�HN"q�vgvr�PU��G����t����s��ϊ��]]u�O�s��s�HrSuyR:hҞ/5m;��D77�/�����O()�%@>#��I��Ek%�Ɏ�����نf_�`�jw%�1�V!�_M�/�s��_��m���!�����jc%�R
                                                                             SR��;�,��k'�L��MBȳM����x���Mς�m�Ahw����LIi�k�+s�_)���I�Hl��
{H��X�T���())e�j^4,y�Xز{Ā_W{���SU�())�Q�m�I��=�
VJJJj���rr��c�X��S�E�F����R?T	�s�ĳ�%$47՞&��j�����e
!�6�4=:d�Jl�=JH��S�VRR�b O�ϴ����\{�f�
�$17N�֠B���@�"ꉠ�����D���_뷇%�,r"�C�JIIi�d����\��i��:�ʆJJJ��5e�~����)3L�\��o���4�����(S��VҴnV�RRR!yu��qAV|C�|M�kPe����FN�-��\s������
VJJJ#,M��7szX����FԓA%%��WҴ�<����=�Ծ�
[...]

PR #3690, opened 2025-05-21
Fix svg upload and thumbnails

PeterNerlich:
### Short description

Ensure SVGs are indeed saved as such on the server.
Mimetypes of files handed out by the server seem to be determined by the file extension only, so we need to manually set the extension for thumbnails of SVGs to PNG.

### Proposed changes

- Don't generate an optimized image for SVGs
- Use the normal thumbnail generation *(which we always need)* to determine if the file is corrupted instead of the generation of an optimized image
- Change the file extension for thumbnails to `.png` if the original files are `.svg`

And as a bonus:
- Allow scaling up SVGs for thumbnail generation even if the stored dimensions of the file are smaller, since SVGs do not blur anyway *(unless someone embedded a raster image and defeated the whole purpose)*

### Resolved issues

Fixes: #3687 

### DIFF
--- a/integreat_cms/cms/forms/media/upload_media_file_form.py
+++ b/integreat_cms/cms/forms/media/upload_media_file_form.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 import mimetypes
 from os.path import splitext
-from typing import TYPE_CHECKING
+from typing import Any, TYPE_CHECKING
 
 import magic
 from django import forms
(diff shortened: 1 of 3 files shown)


integreat_cms/cms/forms/media/upload_media_file_form.py
```diff
@@ -124,19 +124,20 @@ def clean(self) -> dict:
...
             and img_type.startswith("image")
         ):
-            if optimized_image := generate_thumbnail(
-                file,
-                settings.MEDIA_OPTIMIZED_SIZE,
-                False,
-            ):
-                cleaned_data["file"] = optimized_image
-                cleaned_data["thumbnail"] = generate_thumbnail(file)
+            if thumbnail := generate_thumbnail(file):
```
JoeyStk, line 132: Since this method is already getting a little bit out of hand, could you please put this functionality into it's own method? :) If you want to go the extra mile, you could add another commit refactoring this method a little bit (s. scout rule)[https://deviq.com/principles/boy-scout-rule] :)

integreat_cms/cms/utils/media_utils.py
```diff
@@ -41,7 +41,9 @@ def generate_thumbnail(
         if original_image.content_type == "image/svg+xml":
```
JoeyStk, line 41: I'm struggling to understand this method. Thus, at least in my eyes this method could use a little refactoring too. 
This would be my suggestion: 

As far as I understand this method it does four things:
1. Check if content type and if it's SVG convert it into a PNG
2. Check image size. If it's too big -> crop it
3. Compress image into BytesIO

all wrapped into error handling (4). I think we could split this into three sub-method with speaking names, which are called within the error handling. I hope that this way we would be able to get rid off some of the comments too :) 

If you have other ideas for this, please be free to include them :)

JoeyStk (changes requested):
Thank you very much for this PR! The issue is fixed for me, which is a really cool step forward!

The functionality works for me, but some of the code surrounding your code could use some refactoring. Since this PR isn't huge yet, I think you could refactor just within this PR (just in a separate commit) :)

integreat_cms/cms/utils/media_utils.py
PeterNerlich: I don't really have ideas, I tried to apply it as sensibly as I could, but it's at least as much of a mess as the original code on its own. The steps are not as clear cut in functionality and require parameters that are not intuitive looking at the function call.

<details><summary>Rough draft:</summary>

```python
def generate_thumbnail(
    original_image: InMemoryUploadedFile,
    size: int = settings.MEDIA_THUMBNAIL_SIZE,
    crop: bool = settings.MEDIA_THUMBNAIL_CROP,
) -> InMemoryUploadedFile | None:
    """
    Creates a thumbnail for a given media_file.

    :param original_image: Original image for the thumbnail
    :param size: Desired size of the thumbnail (maximum for height and with), defaults to
                 :attr:`~integreat_cms.core.settings.MEDIA_THUMBNAIL_SIZE`
    :param crop: Whether the thumbnail should be cropped (resulting in a square regardless of the initial aspect ratio),
                 defaults to :attr:`~integreat_cms.core.settings.MEDIA_THUMBNAIL_CROP`
    :return: The created thumbnail in memory
    """
    logger.debug(
        "Generating thumbnail for %r with max size: %spx%s",
        original_image,
        size,
        " (cropped)" if crop else "",
    )
    try:
        image = read_image(original_image)

        # Save format, as this information will be lost when resizing the image
        image_format = image.format
        if TYPE_CHECKING:
            assert image_format
        if crop:
            image = apply_crop(image, size)
        else:
            # Resize the image so that the longer side is at most the specified size
            image.thumbnail((size, size), resample=Image.LANCZOS)  # type: ignore[attr-defined]
        buffer = convert_to_buffer(image, image_format)
        # Construct InMemoryUploadedFile from buffer
        thumbnail = InMemoryUploadedFile(
            file=buffer,
            field_name="thumbnail",
            name=original_image.name,
            content_type=Image.MIME[image_format],
            size=buffer.tell(),
            charset=None,
        )
        logger.debug("Successfully generated thumbnail %r", thumbnail)
    except OSError:
        logger.exception("Thumbnail generation for %r failed", original_image)
        return None
    else:
        return thumbnail


def read_image(original_image: InMemoryUploadedFile, size: int) -> Image:
    if original_image.content_type == "image/svg+xml":
        original_image.seek(0)
        svg_bytes = original_image.read()
        png_bytes = cairosvg.svg2png(
            bytestring=svg_bytes, output_width=size, output_height=size
        )
        return Image.open(BytesIO(png_bytes))
    return Image.open(original_image)


def apply_crop(image: Image, size: int) -> Image:
    # Get minimum of original size of the image because ImageOps.fit would otherwise increase the image size
    size = min(image.width, image.height, size)
    # Resize and crop the image into a square of at most the specified size.
    return ImageOps.fit(image, (size, size), method=Image.LANCZOS)  # type: ignore[attr-defined,assignment]


def convert_to_buffer(image: Image, image_format: ?) -> BytesIO:
    # Write PIL image to BytesIO buffer
    buffer = BytesIO()
    # Use optimize option to reduce the image size. Higher quality parameter reduces compression
    image.save(buffer, format=image_format, optimize=True, quality=65)
    return buffer
```

</details>

JoeyStk: Hmm... I agree... This doesn't look much better. I think you can reject the suggestion and resolve this conversation then :)

integreat_cms/cms/forms/media/upload_media_file_form.py
PeterNerlich: I tried, but I'm not really satisfied as it does not feel very clean cut. I'm open for any suggestions!

JoeyStk: Giving this function another look, I don't feel this way anymore :) For me we can resolve this conversation! Thank you for the improvement :)

JoeyStk (approved):
Thank you very much for the refactoring! Really cool feature! Let's get this merged soon! :rocket:

integreat_cms/cms/forms/media/upload_media_file_form.py
PeterNerlich: Leaving it as it now is, or leaving it as it originally was, without outsourced `process_…()` methods?

JoeyStk: Leaving at it is now :)

MizukiTemma (approved):
Thank you so much 😸 Seems to be fixed!

💡 Don't forget to squash the commits before merge 😉

Issue #3270, opened 2024-12-05
Allow copying of POIs


osmers:
### Motivation
If the description of a POI is the same for multiple POIs, e.g. for pharmacies or Kleiderkammern, copying the POI and simply changing the address is a big help.

### Proposed Solution
Add the option to copy POIs, same as for events.

### Alternatives
Keep it, as it is.

### User Story
As a CMS user I would like to be able to copy POIs with the same description to save time on copy and pasting.

### Additional Context
Implement this function
![image](https://github.com/user-attachments/assets/05c4c704-548f-4798-9111-d4e927ba097e)

PR #3699, opened 2025-05-24
Copy pois

JoeyStk:
### Short description

This PR adds the option to copy POIs

### Proposed changes

- Add option to copy POIs

### Side effects

None

### Resolved issues

Fixes: #3270 

### DIFF
--- a/integreat_cms/cms/models/pois/poi.py
+++ b/integreat_cms/cms/models/pois/poi.py
@@ -7,10 +7,17 @@ from django.core.validators import MaxValueValidator, MinValueValidator
 from django.db import models
 from django.utils.functional import cached_property
 
+from integreat_cms.cms.constants import status
+from integreat_cms.cms.utils.slug_utils import generate_unique_slug
+
 if TYPE_CHECKING:
     from typing import Any
 
     from django.db.models.base import ModelBase
+
+    from integreat_cms.cms.models.users.user import User
+
+    from ...utils.slug_utils import SlugKwargs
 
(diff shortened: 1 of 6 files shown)


MizukiTemma (approved):
Thank you 😸 works as expected!

Maybe a release note?

jonbulz (approved):
Looks good to me, thanks! Maybe as a future refactoring idea: Can we move the logic for copying into AbstractContentModel? Or is there too much model-specific logic involved?

JoeyStk:
> Can we move the logic for copying into AbstractContentModel? Or is there too much model-specific logic involved?

I should definitively be possible, but I'm not sure if either I'm just doing something wrong or if it's colliding with our architecture. With the suggested way I get a lot of circular imports. Please have a look at the commented bits in my last commit. The way I grasp the architecture is, that page (and therefore AbstractContentModel) is above  user and slug_utils, which cause the circular import. I assume this is intended. Do you think we should restructure bits of the architecture or rather find a workaround? What do you think? :)

jonbulz:
Sorry, I wanted to come around to check this out, but I didn't find the time :/
If this is getting too complex, I'd suggest to revert the changes, and leave this topic as a future refactoring target. After all, this would possibly affect other models as well.
I already approved your initial suggestions and I stand with that assessment ;)

JoeyStk:
> Can we move the logic for copying into AbstractContentModel? Or is there too much model-specific logic involved?

I think I found a way how to do it that works. The only remaining issue is flaky behavior from mypy

Issue #3626, opened 2025-04-28
Simplify the search when adding a contact into a site


laravoelk:
### Describe the Bug
At the moment, if a contact (which was created via the Kontakte-Modul) is searched via the popup"Kontakte hinzufügen" in a site, it is necessary to spell the whole name. Otherwise the contact can not be found sometimes., which can be pretty confusing (see video). Additionally, sometimes it can also not be found even though, name or place etc. is spelled correctly.

As discussed in our "Kontakte-Modul Austausch" earlier, here are two examples via video to see the problem.

### Expected Behavior
It would be expected to find all contacts when starting to type. At the moment, it is written "no contact can be found". 

Please finde more informations in the added videos.

https://github.com/user-attachments/assets/870e6b27-ccd4-4b39-97c1-998a1249fd9f
https://github.com/user-attachments/assets/650542f3-bb13-4e20-a98f-5f2cc775632d

  ```
  ```

</details>

PR #3704, opened 2025-05-26
Improve contact search for contact card insertion

MizukiTemma:
### Short description

This PR improves search suggestion for users inserting a contact card

### Proposed changes

- Add `.distinct()` to prevent a contact object appearing multiple times in the result
- Increase `MAX_RESULT_COUNT` so matches in rear positions will not be cut off

Those changes seem to have solved the reported cases and not bring side effects, but I cannot provide logical explanation that they are generally an improvement. See the explanation below.

** What I have observed **
- Sometimes one contact object appeared multiple times in the search result -> use `.distinct()`
- Contact objects that are actually matched with `query` string were not shown to users -> use `.distinct()` and increase `MAX_RESULT_COUNT`

** What I tried but did not work**
- Weight the search vectors (`name`, `area_of_responsibility` and `location__title` with `weight="A"`, other fields with `weight="B"`, `website` with `weight="C"`) -> Still objects that are matched by website ULR are shown higher than those matched by name or location name.
- Remove `website` from the vector -> Objects are still matched by website URL and shown in the result

### Side effects

- 
- 

### Resolved issues

Fixes: #3626 

### DIFF
--- a/integreat_cms/cms/models/contact/contact.py
+++ b/integreat_cms/cms/models/contact/contact.py
@@ -122,6 +122,7 @@ class Contact(AbstractBaseModel):
             Contact.objects.filter(location__region=region, archived=False)
             .annotate(rank=SearchRank(vector, query))
             .order_by("-rank")
+            .distinct()
         )
 
     @classmethod
--- a/integreat_cms/cms/views/utils/contact_utils.py
+++ b/integreat_cms/cms/views/utils/contact_utils.py
@@ -18,7 +18,7 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
-MAX_RESULT_COUNT: int = 20
+MAX_RESULT_COUNT: int = 50
 
 
 @require_POST
--- a/integreat_cms/release_notes/current/unreleased/3626.yml
+++ b/integreat_cms/release_notes/current/unreleased/3626.yml
@@ -1,0 +1,2 @@
+en: Improve contact search for contact card insertion
+de: Verbessere die Kontaktsuche für das Einfügen von Kontaktkarten


hannaseithe (commented):
Thank you so much for the fix. And I learned so much from this code review!

I dug really deep on this, and I am not sure I am grasping the issue completely (so please correct me if I am wrong). But from what I could gather: 

At the core of the problem is that when there isnt a full text search match we generate search results with all contacts, that all have the rank 0.0, on which of course ordering makes no sense. 

What is the likely reason for the 0 ranking? 
We use SearchQuery which does full-text search. So typing "s" or "so", will find nothing, but return everything with rank 0.0 (as they are not considered lexems I guess). 

What is the likely reason for returning all contacts? We do not actually filter in our search, so always all contacts will be returned. (But I am assuming that this part is likely intended)

Those results are then sent to the front end (limited to MAX_RESULT_COUNT). In the frontend then the "actual" filtering happens and it only shows those contacts where there is a partial match in the "name" field of the returned json data (which in most cases - based on probability - are none)

Doing distinct() and raising the number of results to 50, raises the chance that among the first 50 of all contacts that are returned (which are "ordered" by rank, but rank is 0.0 for all <- if there wasnt a full text match) is a partial match, which our frontend tinyMCE filter then catches and shows.

Only in the cases where our Django SearchQuery identifies actual full search matches, those get a rank higher than 0.0, and appear at the top of the results.

What do you think? Maybe you guys are already aware of this issue and this is simply an accepted work-around?

MizukiTemma:
@hannaseithe Thank you for the review and new issue with detailed explanation 😸 We'll release this PR as is to improve the situation sooner but come back to #3736 later

juliankehne:
@MizukiTemma Still getting the same issue as before that I have to spell out the full name/location/etc. to find some contacts. So I don't think the fix works. Best retested in a region with a large amount of contacts.

Issue #3144, opened 2024-10-22
After deleting all sub-pages it's not possible to delete the parent page


juliankehne:
### Describe the Bug
Its not directly possible to delete a page after deleting its sub-pages. The parent page first has to be updated or the page tree needs to be refreshed (e.g. by moving a page or creating a new page) before the page can be deleted.

### Steps to Reproduce

1. Go to 'Pages'
2. Create a new page
3. Create a second page as a sub-page for the first page
4. Go back to 'Pages'
5. Delete the sub-page
6. Delete the parent page
7. See that "Delete" icon is inactive (screenshot below)

### Expected Behavior
"Delete" button should be clickable without having to update the page or page tree after deleting the sub-pages

### Actual Behavior
After deleting the last sub-page of a page the "Delete" icon is greyed out and cannot be clicked. It only becomes available once page is updated or the page tree was refreshed, e.g. by moving a page or creating a new one.

### Additional Information
Found in Version: 2024.10.12-alpha

![cannot delete page without subpage](https://github.com/user-attachments/assets/0f62918e-487a-4979-8cce-86569007ceee)

david-venhoff:
Cannot reproduce this locally, but I can on the test server. Since I have redis disabled locally, it is probably a redis caching issue.

osmers:
I mentioned during the call that I could reproduce it on the test system, even so far that I could not delete the page, even after refreshing - I found the problem with that @PeterNerlich: there were archived pages still attributed to this parent page so it appeared empty in the page tree but for system logic was not, bcs the archived pages are connected to it. I only saw that once I opened the page and tried to delete it there... so probably it is what David says :)

osmers:
We could think about that but definitely low prio as it is a problem that only affects us and not the municipalities. I think we have bigger issues to work on right now :)

PR #3711, opened 2025-05-28
Fix: pages cannot be deleted even after deleting all the sub-pages

MizukiTemma:
### Short description

This PR fixes the bug that ghost of deleted child pages are protecting their parent page from delete action.

### Proposed changes

- Use `invalidate_model` after deleting a page
-

### Side effects

- Should be none.
- 

### Resolved issues

Fixes: #3144 

### DIFF
--- a/integreat_cms/release_notes/current/unreleased/3144.yml
+++ b/integreat_cms/release_notes/current/unreleased/3144.yml
@@ -1,0 +1,2 @@
+en: Fix the bug that pages cannot be deleted even with no child page
+de: Behebe den Fehler, dass Seiten nicht gelöscht werden können, auch wenn keine untergeordnete Seite vorhanden ist
--- a/integreat_cms/cms/views/pages/page_actions.py
+++ b/integreat_cms/cms/views/pages/page_actions.py
@@ -15,6 +15,7 @@ import time
 import uuid
 from typing import TYPE_CHECKING
 
+from cacheops import invalidate_model
 from django.conf import settings
 from django.contrib import messages
 from django.core.exceptions import PermissionDenied
@@ -260,6 +261,7 @@ def delete_page(
         logger.info("%r deleted by %r", page, request.user)
         page.delete()
         messages.success(request, _("Page was successfully deleted"))
+        invalidate_model(Page)
 
     return redirect(
         "pages",


integreat_cms/release_notes/current/unreleased/3144.yml
```diff
@@ -0,0 +1,2 @@
+en: Fix the bug that pages cannot be deleted even with no child page
```
JoeyStk, line 1:
```suggestion
en: Fix the bug that pages cannot be deleted even if there isn't a child page anymore
```

integreat_cms/release_notes/current/unreleased/3144.yml
```diff
@@ -0,0 +1,2 @@
...
+de: Behebe den Fehler, dass Seiten nicht gelöscht werden können, auch wenn keine untergeordnete Seite vorhanden ist
```
JoeyStk, line 2:
```suggestion
de: Behebe den Fehler, dass Seiten nicht gelöscht werden können, auch wenn keine untergeordnete Seite mehr vorhanden ist
```
This is a super nit: I was just thinking that the current release note sounds like deleting is never possible, but for pages that never had a child page deleting works (at least for me). That's why I suggest to use a wording that makes it clearer that deleting didn't work if at any point there was a child page. But as mentioned in the intro: It's not highly relevant :)

JoeyStk (approved):
Thank you very much for this PR! :rocket: I tested this and everything works as expected. I also tried to implement this via the django cache and couldn't get it to work. Thus, in my current understanding, this is the best way to do it :)

juliankehne:
1. I don't think it needs a release note, since municipalities can't delete pages.
2. The fix does not work for me. The delete icon in page tree is staying greyed out.

Does cms-test have the fix deployed? 🤔

juliankehne:
@MizukiTemma Thanks, looks like it wasn't updated. Now it works :)

MizukiTemma:
@juliankehne 

We would keep the release note, as most of them [voted for creating a release note also for features that are only available for us](https://chat.tuerantuer.org/digitalfabrik/pl/emyc81xjw3bttn5joppec6tyya)

juliankehne:
> @juliankehne 
> 
> We would keep the release note, as it was voted for creating a release note also for features that are only available for us. See [this thread](https://chat.tuerantuer.org/digitalfabrik/pl/emyc81xjw3bttn5joppec6tyya)

Ah, sounds good :) I like it

Issue #2615, opened 2024-01-22
Configurable machine translation budget per region


svenseeberg:
### Motivation
Malte wants to allocate different fractions of its total translation budget to the regions.

### Proposed Solution
Add an manual override for the total translation budget per region.

### Alternatives
None that I am aware of.

### User Story
As a Integreat CMS admin I want to allocate different translation budgets to regions to satisfy the translation needs of regions while keeping the overall budget as low as possible.

### Additional Context
None

JoeyStk:
Requirements still need to be specified:

- How many variants of budgets are there and how many could there be
- Should there be a freely settable budget? (Malte?)

JoeyStk:
> How many variants of budgets are there and how many could there be

Following input has been given in the meantime:
We need five / maybe six options in the future. These options are: 
- 50.000
- 100.000
- 150.000
- 250.000
- 1 million
- (and maybe) 5 million (can be left out as of May 2025)

> Should there be a freely settable budget? (Malte?)

For now, no

PR #3718, opened 2025-06-01
Make MT budget configurable

MizukiTemma:
### Short description

This PR introduces a configurable MT budget.

### Proposed changes

- Add a new field `mt_budget_booked` to region with constant `machine_translation_budget` covering possible choices
- Remove `mt_addon_booked`
- Adjust te related template and form
- Make a migration transferring `mt_addon_booked` into the menu L (1 mio)
- Adjust `@property` `mt_budget`
- In the region condition table: the colmun "add-on booked" -> "Booked MT budget"
- Adjust affected tests
- Remove `MT_CREDITS_FREE` and  `MT_CREDITS_ADDON`
- ~~Add `MT_BUDGET_BUFFER` to replace `MT_CREDITS_ADDON` with 5% of the booked budget~~

### Side effects

- I hope none 🙈 
- 

### Resolved issues

Fixes: #2615 


integreat_cms/cms/constants/machine_translation_budget.py
```diff
@@ -0,0 +1,27 @@
...
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Final
+
+
+MINIMAL: Final = 50000
+XS: Final = 100000
```
JoeyStk, line 14:
```suggestion
TINY: Final = 100000
```

integreat_cms/cms/constants/machine_translation_budget.py
```diff
@@ -0,0 +1,27 @@
...
+S: Final = 150000
```
JoeyStk, line 15:
```suggestion
SMALL: Final = 150000
```

integreat_cms/cms/constants/machine_translation_budget.py
```diff
@@ -0,0 +1,27 @@
...
+M: Final = 250000
```
JoeyStk, line 16:
```suggestion
MEDIUM: Final = 250000
```

integreat_cms/cms/constants/machine_translation_budget.py
```diff
@@ -0,0 +1,27 @@
...
+L: Final = 1000000
```
JoeyStk, line 17:
```suggestion
LARGE: Final = 1000000
```

integreat_cms/cms/constants/machine_translation_budget.py
```diff
@@ -0,0 +1,27 @@
...
+XL: Final = 5000000
```
JoeyStk, line 18:
```suggestion
HUGE: Final = 5000000
```

I think we should rename this constants for three reasons: a. At the moment I don't find "L" very readable (for example in the migration file) and it's going to be difficult to be found in the code later on. b. apart from the readability I think it makes it easier for us to add new sizes in the future (for example ENORMOUS for bigger than huge or SLIM for a size between small and medium and c. just out of consistency, because we already use MINIMAL for one of the sizes and I think it'd be good if we stick to one convention only :)

integreat_cms/cms/constants/machine_translation_budget.py
```diff
@@ -0,0 +1,27 @@
...
+
+CHOICES: Final[list[tuple[int, str]]] = [
+    (MINIMAL, "50,000"),
```
JoeyStk, line 21:
```suggestion
    (MINIMAL, "50.000"),
```
I think we should stick with the German convention in this case, because so far mostly German municipalities use the CMS. The German convention separates via dot instead of comma :)

integreat_cms/cms/constants/machine_translation_budget.py
```diff
@@ -0,0 +1,27 @@
...
+    (XS, "100,000"),
```
JoeyStk, line 22:
```suggestion
    (XS, "100.000"),
```

integreat_cms/cms/constants/machine_translation_budget.py
```diff
@@ -0,0 +1,27 @@
...
+    (S, "150,000"),
```
JoeyStk, line 23:
```suggestion
    (S, "150.000"),
```

integreat_cms/cms/constants/machine_translation_budget.py
```diff
@@ -0,0 +1,27 @@
...
+    (M, "250,000"),
```
JoeyStk, line 24:
```suggestion
    (M, "250.000"),
```

integreat_cms/cms/constants/machine_translation_budget.py
```diff
@@ -0,0 +1,27 @@
...
+    (L, "1,000,000"),
```
JoeyStk, line 25:
```suggestion
    (L, "1.000.000"),
```

integreat_cms/cms/constants/machine_translation_budget.py
```diff
@@ -0,0 +1,27 @@
...
+    (XL, "5,000,000"),
```
JoeyStk, line 26:
```suggestion
    (XL, "5.000.000"),
```

integreat_cms/core/settings.py
```diff
@@ -935,20 +935,16 @@
...
-
-#: A floating point that specifies the percentage of MT_CREDITS_FREE used as a soft margin
+#: A floating point that specifies the percentage of MT budget used as a soft margin
 MT_SOFT_MARGIN_FRACTION: Final[float] = float(
     os.environ.get("INTEGREAT_CMS_MT_SOFT_MARGIN", 0.01),
 )
 
-#: The actual number of words which are used as soft margin
-MT_SOFT_MARGIN: Final[int] = int(MT_SOFT_MARGIN_FRACTION * MT_CREDITS_FREE)
+MT_BUDGET_BUFFER: Final[float] = float(
```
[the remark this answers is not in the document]
jonbulz: Ok, I think I get it now. When we check against the remaining budget, we first subtract a SOFT_MARGIN, which is defined as a fraction (1%) of the free MT credits, correct? With your changes, this behavior won't change, which I guess is the desired outcome.
I still don't quite understand the story behind it, though. Is it to allow minimal changes, even when there is no remaining budget? Anyway, looks good to me now! :)

jonbulz (approved):

Issue #3638, opened 2025-05-02
Reintroduce phone and e-mail fieds to POI


MizukiTemma:
### Motivation

As we released the contact feature, the three fields of POI (phone number, e-mail and website) were removed and taken over by so-called primary contact (def: contact without name and area of responsibility).

This is not optimal for some munincipalities/users. They wish that the contact feature stays "optional", not a compalsory part of the location feature.

### Proposed Solution

- Reintroduce those fields
- Revive those infos by migration from the corresponding primary contents

### User Story

This is a part of [the first feedback from the munincipalities](https://nextcloud.tuerantuer.org/index.php/apps/files/files/6614783?dir=/&openfile=true).

### Additional Context

Things to clear before working on this issue:

- [ ] @nikolahoff You had some question. You can leave them here if these are not yet solved.
- [ ] @laravoelk Restoring phone number, e-mail and website of POIs should be possible (as long as the automatically generated primary contats have not been deleted). @digitalfabrik/cms-team any opinion?
- [ ] Whether we really do this

### Design Requirements

Does anyone have a screen shot of the side box in which those three fields were shown in the POI form?

### Related Issues

#3086

PeterNerlich:
> - Restoring phone number, e-mail and website of POIs should be possible (as long as the automatically generated primary contats have not been deleted). any opinion?

Purely technically, we could do that with all primary contacts. I don't think we are able to distinguish between those originally generated automatically and those created manually after the fact, but it shouldn't matter. If there is additional information that we cannot represent in those three fields, we just don't delete that contact, right?

MizukiTemma:
From Issue Meeting on 02.06

- Make it possible to deactivate the contact feature in the region form (like POI)
- We do not migrate primary contacts into the revived three fields of POI, but we "read" and render from the primary contacts (Users see three fields for E-mail, phone number and website as before the release of contact system).

JoeyStk:
Requirement from service team: it should be possible to edit contacts in a POI, even when the contact module is enabled. I will ask more questions about it :) @jonbulz / @MizukiTemma Do you have additional questions about it? :)

PR #3727, opened 2025-06-09
Option to activate contacts & reintroduce option to manage contacts in locations

JoeyStk:
### Short description

This PR adds the option to de-/activate contacts & reintroduce the option to mange contacts in a location.

### Proposed changes

- Option to de-/activate contacts in region form
- For regions with deactivate contacts: add the option to manage contacts in a location
- Do it via the contact model, so the module can be easily switch on/off again without the need to migrate anything

### Side effects

- I'm double-checking the requirements with UI/UX and service team
- I'm not sure if I already thought of all cases, please test thoroughly and be nit-picky :) that's highly appreciated :)

- Archive/delete of POIs are so adjusted that POIs can be archived/deleted even if they are referred by a contact if the contact module is deactivaed in the region (otherwise users cannot archivde/delete a POI if they filled any of the contact details, becase then a contact object referring the POI exists but no one can delete the contact because the contact module is not available 🤷 )
- The option to insert a contact card is not shown if the contact module is deactivated in the region.

### Resolved issues

Fixes: #3638 


JoeyStk:
This week we had some more talks about what still needs to be done: 

- [ ] The box in locations should also be displayed when contacts are activated
- [ ] We want to introduce a checkbox in the contact form that you can check to say "this contact is location checkbox" (wording can be adjusted, maybe a question for UI/UX) -> if this checkbox is checked, show the according contact on the poi_form 
- [ ] Also the other way around if a contact is created in the poi_form, the checkbox is automatically checked


As I'm busy with a lot of different things, please feel free to commit to this branch :)

MizukiTemma:
@hauf-toni @laravoelk @jonbulz 

I have started to bring this PR a bit forward and have some question.

> 
> * [ ]  The box in locations should also be displayed when contacts are activated

After this is implemented, if contact is activated in the region, users see the primary contact in the contact detail box and also in the box of related contacts. This is probably confusing. Should we add some explanation about that to the detail box?

> * [ ]  ~We want to introduce a checkbox in the contact form that you can check to say "this contact is location checkbox" (wording can be adjusted, maybe a question for UI/UX) -> if this checkbox is checked, show the according contact on the poi_form~
> * [ ]  ~Also the other way around if a contact is created in the poi_form, the checkbox is automatically checked~

I guess we don't need such check box anymore as we can distinguish primary contacts, right?

laravoelk:
I think, the easiest the better. ;-) So if it is possible to do it without a checkbox, that would be good.

MizukiTemma:
From today's meeting with @laravoelk and @hauf-toni the following changes are requred:

- Deactivate the field for `area_of_responsibility` in the contact form if it is a primary contact (to avoid the contact informations unintentionally disappear from the POI form)
- When a POI is saved with contact informations, the primary contact automatically takes the name of POI as its name (to make it clear that it is a primary contact)
- Move the box "Contact details" directry anove the box "Related contacts".
- Show in the box "Contact details" an info text "Die hier angegebenen Kontaktdaten werden standardmäßig angezeigt. Für bestimmte Zuständigkeiten können zusätzliche Kontakte hinterlegt werden."

integreat_cms/cms/views/contacts/contact_list_view.py
```diff
@@ -140,3 +143,19 @@ def redirect_to_poi_list(self, request: HttpRequest) -> None:
...
+
+    def redirect_to_dashboard(self, request: HttpRequest) -> None:
+        """
+        This function redirects to the dashboard if contacts are not enabled.
+
+        :param request: The current request
+
+        :return: template of region dashboard
+        """
+        messages.error(request, _("Contacts are not enabled for this region."))
```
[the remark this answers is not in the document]
MizukiTemma: The only (likely) scenario would be that user changes the region slug directly from a region with contact activated to another with contact deactivated in the URL bar of the browser.

I changed it to ` info`.

integreat_cms/static/src/js/tinymce-plugins/custom_contact_input/plugin.js
```diff
@@ -3,6 +3,12 @@ import { getCsrfToken } from "../../utils/csrf-token";
 
 (() => {
     const tinymceConfig = document.getElementById("tinymce-config-options");
+    const contactEnabled = tinymceConfig.getAttribute("data-contact-module-activated");
+
+    if (contactEnabled === "False") {
+        return;
+    }
```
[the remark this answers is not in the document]
MizukiTemma: @PeterNerlich 
Nice check! Applied 😸

integreat_cms/cms/forms/pois/poi_form.py
```diff
@@ -72,6 +83,17 @@ def __init__(self, **kwargs: Any) -> None:
         """
         super().__init__(**kwargs)
 
+        # Look explicitly for the primary contact, not the any first one,
+        # as we do not delete non-primary contacts when deactivating contact in a region.
+        # "get()" is not used as it raises an exception if there is no primary contact.
```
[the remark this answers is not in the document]
JoeyStk: I meant the exception. Is it difficult to handle exception? :thinking:

nassabay:
Hey @MizukiTemma. I’m not entirely sure I follow all the technical details, but I fully agree that any special behavior should be documented clearly. Users need to understand the concept in order to work correctly with the module. In fact, we already have some reference material here: https://wiki.integreat-app.de/kontakte-modul#konzeptkontakte_und_ihre_zuordnung_zu_orten_erste_informationen

Some points for consideration:
1. Deletion of related contacts when deleting a POI with the module deactivated:
Keep in mind that in Integreat, users cannot actually delete information — they can only archive it. This applies to "Seiten", "Veranstaltungen", "Orte", and "Kontakte". Therefore, from the user’s perspective, nothing ever truly “disappears” from the system. This is also how we explain it to them.

2. Contacts linked to multiple locations
What happens if the same contact is associated with two different locations? If one of those locations is deleted in a region where the contact module is deactivated, and as a result all its contacts are deleted, would that contact also disappear from the second location? 

3. Definition of primary vs. secondary contact
I’m not entirely clear on how the primary contact is defined for a location that essentially is the primary contact itself. Example: a town hall with its full contact details (address, phone, email), plus several employees who work there. In such a case, is the POI itself considered the primary contact, or should an additional primary contact be created?

JoeyStk:
For me this is ready to be uploaded to cms-test and then we can get some more feedback from PO and service team :)

PeterNerlich:
Should be deployed now, please confirm. I also hacked the template to display a different version in the lower left, so as long as it shows `Version: #3727 rollback-contact-in-poi (06c5899)` you know that it should still reflect the state of this PR and has not been overwritten

MizukiTemma:
> Should be deployed now, please confirm. I also hacked the template to display a different version in the lower left, so as long as it shows `Version: #3727 rollback-contact-in-poi (06c5899)` you know that it should still reflect the state of this PR and has not been overwritten

Thank you so much 😻 As there is now a check box to activate/deactivate contacts in the region form in the test system, I confirm it is deployed.

MizukiTemma:
> Hey @MizukiTemma. I’m not entirely sure I follow all the technical details, but I fully agree that any special behavior should be documented clearly. Users need to understand the concept in order to work correctly with the module. In fact, we already have some reference material here: https://wiki.integreat-app.de/kontakte-modul#konzeptkontakte_und_ihre_zuordnung_zu_orten_erste_informationen
> 
> Some points for consideration:
> 
> 1. Deletion of related contacts when deleting a POI with the module deactivated:
>    Keep in mind that in Integreat, users cannot actually delete information — they can only archive it. This applies to "Seiten", "Veranstaltungen", "Orte", and "Kontakte". Therefore, from the user’s perspective, nothing ever truly “disappears” from the system. This is also how we explain it to them.


> 2. Contacts linked to multiple locations
>    What happens if the same contact is associated with two different locations? If one of those locations is deleted in a region where the contact module is deactivated, and as a result all its contacts are deleted, would that contact also disappear from the second location?

There shold be no contact that is assosiacted with more than one location. Each contact has exactly one location.

> 3. Definition of primary vs. secondary contact
>    I’m not entirely clear on how the primary contact is defined for a location that essentially is the primary contact itself. Example: a town hall with its full contact details (address, phone, email), plus several employees who work there. In such a case, is the POI itself considered the primary contact, or should an additional primary contact be created?

A primary contact is a contact whose `area_of_responsibility` is empty. For each location there can be max. one primary contact. A locationcan has no primary contact or one. The intention of primary contact is the "generall" contact information for the POI, not each department or persons working in the POI. It is expected to store the central E-mail Address or the phone number of reception as primary contact, and to create contacts with `area_of_responsibility` for contact information of each department or persons liike "are_of_responsbility: CMS Team Coodination, name: Mizuki Temma, E-mail: some-address-xyz@mail"

MizukiTemma:
> Sorry to ask, but I'm confused about the current state of this PR after the deployment on cms_test. Would it be helpful if I review the current state or are there still some changes that need to be implemented first @MizukiTemma ?

@JoeyStk 

I, @laravoelk and @hauf-toni discussed it today, and we are going to merge this PR after the release `2025.10.0` 🚀 

You can review again or approve without review.

MizukiTemma (approved):
🚀 🚀 🚀

integreat_cms/cms/forms/contacts/contact_form.py
```diff
@@ -57,15 +57,20 @@ class Meta:
...
         if instance and instance.id:
             adopt_hours = instance.opening_hours is None
             if adopt_hours:
                 instance.opening_hours = instance.location.opening_hours
+
         super().__init__(**kwargs)
         self.fields["use_location_opening_hours"].initial = adopt_hours
 
+        if instance and not instance.area_of_responsibility:
+            self.fields["area_of_responsibility"].disabled = True
```
JoeyStk, line 72: Why do we need this line? :)

tests/cms/views/contacts/test_contact_form.py
```diff
@@ -97,7 +97,12 @@ def test_edit_a_contact(
     settings.LANGUAGE_CODE = "en"
 
     region = Region.objects.filter(slug=REGION_SLUG).first()
-    contact_id = Contact.objects.filter(location__region=region).first().id
+    contact_id = (
+        Contact.objects.filter(location__region=region)
+        .exclude(area_of_responsibility="")
```
JoeyStk, line 102: Why do we exclude primary contacts here? :)

JoeyStk (commented):
Thank you very much! Since this is my PR, I can't approve it, but I think it looks good :)

integreat_cms/cms/forms/contacts/contact_form.py
MizukiTemma: It was reuqested by Service Team to prevent users accidentally changing a primary contact to a normal contact

integreat_cms/cms/forms/pois/poi_form.py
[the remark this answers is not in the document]
MizukiTemma: Nop. It's a pure preference thing: `get()` is to be used for the case (in my opinion) when there must be (at least/exactly) one matching object. We can handle of cours an exeption and set `None`, but I don't think we need extra code lines for that.

integreat_cms/cms/views/pois/poi_form_view.py
```diff
@@ -228,6 +234,28 @@ def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
                     ),
                 )
 
+            # Look explicitly for the primary contact, not any first one,
+            # as we do not delete non-primary contacts when deactivating contact in a region.
+            # "get()" is not used as it raises an exception if there is no primary contact.
+            contact = poi.contacts.filter(area_of_responsibility="").first()
```
jonbulz, line 240: Maybe we could think about making this a manager method? This would then look something like:

```
class ContactQuerySet(model.QuerySet):
    def get_primary(self):
        return self.filter(area_of_responsibility="").first()

class Contact(AbstractBaseModel):
    ...
    objects = models.Manager.from_queryset(ContactQuerySet)()
```
Then we could do something like:
```
contact = poi.contacts.get_primary()
```
It's more code now, but maybe more extendable? :shrug: But we can definitely also keep it the way you did it, I don't think that this is too much duplicate code. Just a suggestion from the sideline :)

integreat_cms/cms/forms/pois/poi_form.py
```diff
@@ -33,6 +33,17 @@ class POIForm(CustomModelForm):
     #: The distance in km between the manually entered coordinates and the coordinates returned from Nominatim
     nominatim_distance_delta = 0
 
+    primary_website = forms.URLField(max_length=250, required=False, label=_("website"))
```
jonbulz, line 36: I'd suggest "Website" with capital letters as label :)

integreat_cms/cms/forms/pois/poi_form.py
```diff
@@ -33,6 +33,17 @@ class POIForm(CustomModelForm):
...
+    primary_email = forms.EmailField(
+        label=_("email address"),
```
jonbulz, line 38: Same here, I think capital letters work better as label

integreat_cms/cms/forms/pois/poi_form.py
```diff
@@ -33,6 +33,17 @@ class POIForm(CustomModelForm):
...
+        required=False,
+    )
+    primary_phone_number = forms.CharField(
+        max_length=250,
+        required=False,
+        label=_("phone number"),
```
jonbulz, line 44: Same here, I think capital letters work better as label

jonbulz (approved):
I think this is fine as it is, so if you want you can just merge it. I still left a few suggestions, but feel free to ignore them. Thank you! :)

integreat_cms/cms/views/pois/poi_form_view.py
MizukiTemma: Good idea 😸 Opened an issue for that #3922

Issue #3730, opened 2025-06-10
Creating/Updating contacts not possible due to missing permissions for opening hours


juliankehne:
### Describe the Bug

The following error message is shown when creating a new contact or updating a contact for Manager, Editor and Author role:

> Opening hours: An error occurred while saving the opening hours. Please contact an administrator. 

The contact is not created or updated.

Contact creation is possible without the error when creating a new location.
Contact creation when editing a location is not possible due to another issue #3728 

### Steps to Reproduce

Prerequisite: Manager, Editor or author role

1. Go to 'Contacts'
2. Click on 'Create contact'
3. Select a location
4. Fill at least one field
5. Click 'Create'
6. See Error

Updating:

1. Go to 'Contacts'
2. Open an existing contact
3. Click 'Update'
6. See Error

### Expected Behavior

Creation and Update of a contact works for Manager, Editor and Author role

### Actual Behavior

Creation and Update of a contact doesn't work for Manager, Editor and Author role

### Screenshot

![Image](https://github.com/user-attachments/assets/16e8a0af-3846-4346-92c7-ffcec41cba49)

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

Issue #3728, opened 2025-06-10
Creation of contacts not possible due to an error when editing a location


juliankehne:
### Describe the Bug

When editing a location its no longer possible to create new contacts as a message is shown "An error occurred." when clicking the Submit button to create a new contact.

### Steps to Reproduce

Prerequisites: An existing location

1. Go to 'Locations'
2. Open an existing location
3. Scroll down to the contact ajax form and click 'Create new contact'
4. Enter some info into the fields
5. Click 'Submit' to create the contact
6. See Error message

### Expected Behavior

No error message & contact is created.

### Actual Behavior

An error message when clicking submit and no contact is being created.

### Screenshot

![Image](https://github.com/user-attachments/assets/60e1a07d-ec00-44a6-826a-b814ae24343a)

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

PR #3731, opened 2025-06-10
Fix: contact update and create without beta permission

MizukiTemma:
### Short description

This PR fixes the bug that users without beta permission can neither create nor edit contacts.

### Proposed changes

- Skip the validation of opening hours for users without beta permission (by handling as `None`)
- 

### Side effects

- I hope none 🙈 
- 

### Resolved issues

Fixes: #3730 


JoeyStk (approved):

jonbulz (approved):

integreat_cms/cms/forms/contacts/contact_form.py
```diff
@@ -93,6 +94,10 @@ def validate_opening_hours(self) -> list[dict[str, Any]] | None:
 
         :return: The valid opening hours
         """
+        # Remove when opening hours become available for all users
+        if not self.request.user.has_perm("cms.test_beta_features"):
```
JoeyStk, line 98:
```suggestion
        if self.request and not self.request.user.has_perm("cms.test_beta_features"):
```
This is a little strange, but if I understand correctly the request is a 'NoneType' for tests. Therefore the test is failing at the moment. I think this should do it. At least it does for me locally :thinking:

JoeyStk (changes requested):

integreat_cms/cms/forms/contacts/contact_form.py
MizukiTemma: Thank you for speedy reply 😸 

Hmm, I'm not for adjusting the code for the test and did otherway around 🤔 Could you check the second commit?

JoeyStk (approved):
Thank you a lot for your fix and this PR in general! It looks good to me now :rocket:

jonbulz (changes requested):
Right, good catch! I think contact_form_ajax_view:65 also needs the request now

MizukiTemma:
> Right, good catch! I think contact_form_ajax_view:65 also needs the request now

OK, we'll fix #3728 too here 💪

integreat_cms/cms/forms/contacts/contact_form.py
```diff
@@ -102,6 +107,11 @@ def validate_opening_hours(self) -> list[dict[str, Any]] | None:
             _("Please contact an administrator."),
         )
         cleaned_opening_hours = self.cleaned_data["opening_hours"]
+
+        # Remove when opening hours become available for all users or after implementing opening hours for the ajax contact form too
+        if cleaned_opening_hours is None:
+            return None
+
```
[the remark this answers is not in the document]
JoeyStk: Yes, I think that makes sense :)

MizukiTemma:
@JoeyStk @jonbulz 
Again ready for review 😸

JoeyStk:
> Fixes: https://github.com/digitalfabrik/integreat-cms/issues/3730
Fixes: https://github.com/digitalfabrik/integreat-cms/issues/3728

Amazing! Thank you so much for your valuable work :blush:

jonbulz (approved):
Oops, sorry. I wasn't aware that the other issue was supposed to be handled separately. 
Thanks, looks good to me! :)

JoeyStk (approved):
Both issues are resolved for me! Thank you so much! :)

Issue #3570, opened 2025-04-07
Option to archive POIs, even if they are connected to a passed event


laravoelk:
### Motivation
Currently POIS can not be archived if they are connected to any event. In our opinion, this makes sense if the event is upcoming. However, we have been notified from one municipality that they could not archive a POI as an event, which happened some time ago, was connected to that POI.

### Proposed Solution
Provide the option, that POIS can be archived if they are connected to past events.

### Alternatives
TBD

MizukiTemma:
This is probably one of the good issues for @hannaseithe

PR #3732, opened 2025-06-10
Allow pois used by past events to be archived

hannaseithe:
### Short description
Pois (locations) that are used by past events can be archived now. This works both in single actions and bulk actions.

### Proposed changes
- I added a new property is_currently_used that tests for being used by a contact or an upcoming (!) event to check if archiving is allowed (instead is_used, which is still needed for delete actions)
- added tests

### Side effects
- none that I am aware of

### Resolved issues

Fixes: #3570 


MizukiTemma (approved):
Looks good 😸 Nice updating of docs 👍 Thank you!

MizukiTemma:
> Thank you very much! This is a really cool feature :) Code looks good and works as expected :)
> 
> I just have one question about the requirements of this feature: I thought it was implemented the way it was, because we wanted to educate the municipalities that they should unlink events and pois after they are done / before they archive them. Do we think differently about this now? And if so, should we also change this behavior for the service team when they want to delete POIs? :)

Maybe a question to @laravoelk

laravoelk:
From what I know, municipalities hardly unlink events and pois after an event. So from our side, we do think differently about this now. :-)

juliankehne:
Just tested this on cms-test. Works great.

One potential issue I see is when a cms user reuses a past event and updates it with an upcoming date. If that event has an archived location attached there is no indicator for it.
This might also happen if such an event is copied, date updated and the location isn't adjusted.
The end user in the app won't be able to see the location for an event as the location link will lead to a 'non-existent' location.

I also found that the event filter when searching for a location doesn't show archived locations, which makes sense, but now there can be events with archived locations.

Both of these probably don't require any immediate actions. Maybe good to keep in mind though.

Issue #3712, opened 2025-05-28
The contact module doesn’t seem to handle the phone number format correctly.


AnnaPerelazova:
### Describe the Bug
The format of the phone number in the contact module is changed automatically, but the number cannot be reached. The automatic format in the contact module is the same as if I enter the number manually on the page, but in the contact module it seems to cause issues.

Example: number looks the same both in the contact module and on the page:
https://integreat.app/testumgebung/de/locations/bibliothek/ (via contact module)
https://integreat.app/testumgebung/de/willkommen/wichtige-aemter/test-2

On a mobile phone, the number from the first link triggers a “number not assigned” message and shows “zero” after +49 while the number from the second page (not entered via the contact module) works fine – it is dialed correctly and shown without the zero.

![Image](https://github.com/user-attachments/assets/2029b3a3-3a71-43ea-8292-d0f5ad20da86)

![Image](https://github.com/user-attachments/assets/d86874d0-2c2f-4d29-ae99-fb8d510f7953)

Can you please check?

PeterNerlich:
I wonder what the `(0)` was originally inspired by. From what I can see, it was first brought up in #3383 but I don't think I have ever seen it myself. It is not mentioned in [E.164](https://en.wikipedia.org/wiki/E.164).

I expect that `(0)` is supposed to explain to the reader that they can simply substitutte the `+49` for a `0` if they are inside Germany, but this way of writing it feels weird to me and can lead to problems like the one present here.

My personal preference would be to drop the `(0)` altogether, but I think the minimal change would be to ensured it is dropped from the `href="tel:+49…` attribute only

PeterNerlich:
> the minimal change would be to ensure it is dropped from the `href="tel:+49…` attribute only.

wait, this is already the case

PeterNerlich:
Okay, the difference is

- In the editor, we expect the numbers to be entered like: `0123456789` or `+49123456789` *(without spaces)*. Both will produce a link to `tel:+49123456789` with the text `+49 (0) 123456789`
- In the contact form, we have a simple text field for entering the number. We don't really do validation or parsing on it since we had much more pressing concerns at the time of implementation, like *"Will this completely new architecture even work at all?"* and opted to get to a working Proof of Concept as quickly as possible, and continue from there.
  Hence the problem, as the server side card rendering does differentiate between *actually dial-able* and *nice eye candy (personal opinions beside)* format and just takes whatever is in the field for both.

This means that our TODO is:
[...]

PR #3741, opened 2025-06-12
Remove (0) from phone number formatting

MizukiTemma:
### Short description

This PR changes the behaviour of phone number formatting

### Proposed changes

- Remove ` (0) ` from the formatting
- Add a migration to adopt the (mobile) phone numbers of the existing contacts into the new format
- Adjust the test

### Side effects

- I hope none 🙈 
- 

### Resolved issues

Fixes: #3712 

### DIFF
--- a/integreat_cms/cms/migrations/0137_adopt_new_phone_number_format.py
+++ b/integreat_cms/cms/migrations/0137_adopt_new_phone_number_format.py
@@ -1,0 +1,42 @@
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING
+
+from django.db import migrations
+
+if TYPE_CHECKING:
+    from django.apps.registry import Apps
+    from django.db.backends.base.schema import BaseDatabaseSchemaEditor
+
+logger = logging.getLogger(__name__)
+
+
+def migrate_phone_number_format(
+    apps: Apps,
+    _schema_editor: BaseDatabaseSchemaEditor,
+) -> None:
+    """
+    Adopt existing phone numbers into the new format (remove "(0)")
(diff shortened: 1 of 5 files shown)


PeterNerlich:
**I don't think this will fix the issue**. I already manually stripped most of the ` (0) ` from the phone numbers in contacts, if this cropped up again in >5 instances *(that was how many edge cases I left and didn't bother handling)* then **this will just be a second one-time action**, and is not fully sufficient for either data getting added in the future or data already in the system *(e.g. there are cases where the value entered for `phone_number` reads `+49123456789 for Miss X from 9 to 12 or +49123456789 for Miss Y from 13 to 16`)*.

You do not need to modify the **display form for phone numbers entered in content objects** at all, that feature works without issue and is entirely *(and confusingly)* unrelated to `phone_number` in contacts.

In my opinion, a **long term solution** is only to be **more strict on the `phone_number` field** to only allow phone numbers that we can parse and clean them to a canonical form *(then we can also make the contact card use the nice version for display)*. However, in that case we will be rejecting a lot of real-world edge cases in favour of our oversimplified model.
If we want to instead refine our model to better match reality we will likely need considerable development time for the added complexity. I imaging being able to attach multiple phone numbers with optional *(separate)* notes for things like availability times.

MizukiTemma:
I would say, it is a fault of users if they enter something not valid as phone number (e.g. 123qwer456) and therefore the app cannot dial up or the link doesn't work or whaever. What I would suggest as a kind of parsing garantie would be allowing only numbers in the (mobile) phone number fields if we abolish `(0)`.

PR #3751, opened 2025-06-23
Add launch.json for vs code debugging

hannaseithe:
### Short description
This is a setup where we commit the launch.json and have optional configuration for debugging with docker and without docker

### Proposed changes
- start stracking launch.json
- add start docker and stop docker tasks to task.json
- edit the documenation to explain the setup

### Side effects
- individual changes to launch.json won't be possible anymore, but from my understanding it is currently only Jarl and Me who use VSCode. And we use launch.json only for the debugging setup

### Resolved issues

If this works, this is an improvement since launch.json won't have to be carried around all the time and new developers who use vs code can just start debugging straight away

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,35 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Python Debugger: Django (with Docker)",
+      "type": "debugpy",
+      "request": "launch",
+      "args": ["runserver"],
+      "django": true,
+      "autoStartBrowser": false,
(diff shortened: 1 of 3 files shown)


jarlhengstmengel (approved):
I tested without docker inside the Devcontainer and that works. Docker isn't installed inside the devcontainer so with docker didn't work (as to be expected). Maybe @JoeyStk could test that :)

JoeyStk:
Unfortunately starting up the debugger takes forever to load and doesn't get to an end :/ Could you please have another look at this? :)

Issue #3760, opened 2025-06-25
Function to copy POIs, conatcs and events from a region to another


MizukiTemma:
### Motivation

The current region cloning duplicates only pages. It has been asked multiple times to copy POIs of a region to another. Now contacts and events mut be sometimes copied. Making a management commad for this purpose or even make it avaibale in GUI will help us.

### Proposed Solution

Following objects are often duplicated from a region to another.
- [ ] POIs
- [ ] Contacts
- [ ] Events

How to implement: management command and/or feature in GUI 

### Alternatives

Do it everytime manually when asked

### User Story

I, as a developer of CMS Team, want to make this procedure partially or completely automated

PR #3762, opened 2025-06-25
Add management command to copy pois

JoeyStk:
### Short description

This PR adds management commands to copy pois, optionally along with events and/or contacts.

### Proposed changes

- Management command to copy pois
- Through `--events` and `--contacts` it will also copy events and contacts, respectively
- If `--add-suffix` is given, created duplicates will get the suffix ` (Copy)` appended to their name/title/…
- With `--user` the user can be specified that will be recorded as the creator of the duplicates
- Internally, the new implementation of `.copy()` on `AbstractContentModel`s is leveraged

### Resolved issues

Fixes: #3760
Fixes: #3764

### DIFF
--- a/docs/src/management-commands.rst
+++ b/docs/src/management-commands.rst
@@ -76,6 +76,22 @@ Duplicate all currently existing pages to make it easier to create production-li
     This command inherits from :class:`~integreat_cms.core.management.debug_command.DebugCommand`, so it is only available in debug mode.
 
 
+
+``copy_pois``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Duplicate all POIs of the template region, also all events and contacts if specified, into the target regions
+
+    integreat-cms-cli copy_pois TEMPLATE-SLUG TARGET-SLUG [TARGET-SLUG ...] [--username USERNAME] [--contacts] [--events] [--add-suffix]
+
+**Options:**
+
+* ``--username``: The user as which to create the objects
+* ``--contacts``: Whether contacts should be copied too
+* ``--events``: Whether events should be copied too
+* ``--add-suffix``: Whether the suffix ' (Copy)' should be added to copied objects
+
+
 ``find_large_files``
 ~~~~~~~~~~~~~~~~~~~~
 
--- a/integreat_cms/core/management/commands/copy_pois.py
+++ b/integreat_cms/core/management/commands/copy_pois.py
@@ -1,0 +1,179 @@
+from __future__ import annotations
+
+import logging
+import os
+import re
+from copy import deepcopy
+from typing import TYPE_CHECKING
+
+from django.contrib.auth import get_user_model
+from django.core.management.base import BaseCommand, CommandError
+from django.db import transaction
+from django.utils.translation import gettext_lazy as _
+
+from integreat_cms.cms.models import Event, POI, Region, User
+
+if TYPE_CHECKING:
+    from typing import Any
+
+    from django.core.management.base import CommandParser
+    from django.db.models import FileField
+
+
+logger = logging.getLogger(__name__)
+
+
+def make_hardlink(file_field: FileField, region_id: int) -> FileField:
+    """
+    Create a hardlink to the original file in the media path for the target region
+    """
+    base = file_field.storage.base_location
+    orig_path = file_field.name
+    new_path = re.sub(r"^regions/[0-9]+/", f"regions/{region_id}/", orig_path)
+    # make the directory
+    os.makedirs(os.path.dirname(os.path.join(base, new_path)), exist_ok=True)
+    # create the hard link
+    try:
+        os.link(os.path.join(base, orig_path), os.path.join(base, new_path))
+    except FileExistsError as e:
+        if (
+            os.stat(os.path.join(base, orig_path)).st_ino
+            == os.stat(os.path.join(base, new_path)).st_ino
+        ):
+            logger.info(
+                "Hard link already exists between %r and %r",
+                new_path,
+                orig_path,
+            )
+        else:
+            raise FileExistsError("Failed to create hard link") from e
+    # modify the file information
+    file_field.name = new_path
+    return file_field
+
+
+def copy_icon(region: Region, obj: POI | Event) -> None:
+    """
+    Ensure the icon of an object is accessible from the target region as well
+    """
+    if obj.icon and obj.icon.region is not None:
+        # Theoretically, there might be orphaned media files
+        obj.icon.pk = None
+        obj.icon.region = region
+        try:
+            obj.icon.file = make_hardlink(obj.icon.file, region.pk)
+        except FileNotFoundError:
+            logger.info("Could not copy icon for %r: FileNotFound", obj)
+        obj.icon.save()
+
+
+def get_user(username: str) -> User:
+    """
+    Get a user by username or raise an error if not found
+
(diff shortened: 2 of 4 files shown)


JoeyStk:
Since I'm unsure when I'm going to be free to continue working on this, feel free to finish off this PR :)

MizukiTemma:
The second commit includes:

- Doc
- Refactoring to cover `copy_pois` by `copy_pois_and_contacts`, by adding a flag `--contacts`. `copy_pois` can be deleted if the refactored version looks OK

MizukiTemma:
@JoeyStk @jonbulz 

Could you check once if the current all-in-one comannd looks okay?

docs/src/management-commands.rst
```diff
@@ -76,6 +76,20 @@ Duplicate all currently existing pages to make it easier to create production-li
     This command inherits from :class:`~integreat_cms.core.management.debug_command.DebugCommand`, so it is only available in debug mode.
 
 
+
+``copy_pois_and_contacts``
```
JoeyStk, line 80: I don't think this is an ideal name anymore, since a. it's possible to not copy contacts and b. it's also possible to copy events. But I'm also indecisive what name is better :thinking:
Either `copy_pois_events_and_contacts` which is long, or just `copy_pois` because the rest is optional, or `copy_content` but that way it sounds like pages could also be copied? :thinking:

docs/src/management-commands.rst
```diff
@@ -76,6 +76,20 @@ Duplicate all currently existing pages to make it easier to create production-li
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Duplicate all POIs of the template region, also all contacts if specified, into the target regions
```
JoeyStk, line 83:
```suggestion
Duplicate all POIs of the template region, also all events and contacts if specified, into the target regions
```

integreat_cms/core/management/commands/copy_pois_contacts_events.py
```diff
@@ -0,0 +1,203 @@
...
+    help = "Duplicate POIs and contacts into one or many target regions"
+
+    def add_arguments(self, parser: CommandParser) -> None:
+        """
+        Define the arguments of this command
+
+        :param parser: The argument parser
+        """
+        parser.add_argument(
+            "template_slug", type=str, help="Der Slug der Template-Region"
```
JoeyStk, line 126:
```suggestion
            "template_slug", type=str, help=_("The slug of the template region")
```

integreat_cms/core/management/commands/copy_pois_contacts_events.py
```diff
@@ -0,0 +1,203 @@
...
+        )
+        parser.add_argument(
+            "target_slugs", type=str, nargs="+", help="Die Slugs der Ziel-Regionen"
```
JoeyStk, line 129:
```suggestion
            "target_slugs", type=str, nargs="+", help="The slugs of the target regions"
```

JoeyStk (commented):
Thank you so much for your commits! This is not a full review yet. I will do so in the next days :)

docs/src/management-commands.rst
MizukiTemma: I decided for `copy_pois_events_and_contacts` :)

integreat_cms/core/management/commands/copy_pois_events_and_contacts.py
```diff
@@ -0,0 +1,207 @@
...
+
+if TYPE_CHECKING:
+    from typing import Any
+
+    from django.core.management.base import CommandParser
+    from django.db.models import FileField
+
+    from integreat_cms.cms.models import EventTranslation, POITranslation
+
+    from .users.user import User
```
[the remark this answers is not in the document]
MizukiTemma: what?

@jonbulz Do you have any problem with this import?

MizukiTemma (approved):

PeterNerlich (changes requested):
When I run the command multiple consecutive times, I expect it to not find anything new to add to it, but for me it still contains these messages:

```
Saved translation to target region in <Language (id: 2, slug: en, name: English)>: <POITranslation (id: 14, poi_id: 11, language: en, slug: test-location)>
Saved translation to target region in <Language (id: 2, slug: en, name: English)>: <POITranslation (id: 15, poi_id: 12, language: en, slug: draft-location)>
```
and on the next iteration
```
Saved translation to target region in <Language (id: 2, slug: en, name: English)>: <POITranslation (id: 16, poi_id: 13, language: en, slug: test-location)>
Saved translation to target region in <Language (id: 2, slug: en, name: English)>: <POITranslation (id: 17, poi_id: 14, language: en, slug: draft-location)>
```

integreat_cms/core/management/commands/copy_pois_events_and_contacts.py
```diff
@@ -0,0 +1,207 @@
...
+                "Saved translation to target region in %r: %r",
+                tr.language,
+                tr,
+            )
+        else:
+            logger.info(
+                "Not saving translation in language not in target region (%r): %r",
+                tr.language,
+                tr,
+            )
```
PeterNerlich, line 101: Why not make this a continue statement? It might make this a bit more readable
```suggestion
        if tr.language not in region.languages:
            logger.info(
                "Not saving translation in language not in target region (%r): %r",
                tr.language,
                tr,
            )
            continue

        # Ensure partial attempts don't make everything more difficult
        try:
            dup = obj.translations.get(language=tr.language, version=tr.version)
        except obj.translations.model.DoesNotExist:
            pass
        else:
            logger.info(
                "Not saving translation, already exists in %r: %r",
                tr.language,
                dup,
            )
            continue
        tr.pk = None
        setattr(tr, related_name, obj)
        # Don't update the timestamp
        tr.save(update_timestamp=False)
        logger.info(
            "Saved translation to target region in %r: %r",
            tr.language,
            tr,
        )
```

PeterNerlich:
The only test left failing this time was
```
ERROR tests/cms/utils/test_repair_tree.py::test_repair_tree - django.db.utils.IntegrityError: duplicate key value violates unique constraint "django_content_type_app_label_model_76bd3d3b_uniq"
DETAIL:  Key (app_label, model)=(cms, firebasestatistic) already exists.
```
which I don't think I've seen before, but is in no way connected to this PR. I fear the number of CI oddities is going up :face_exhaling:

PeterNerlich:
@MizukiTemma @jonbulz Could at least one of you have a quick check whether you consider your review still valid with the changes since?

We cut out the function to manually copy the translation objects and now use the `copy()` method of any `AbstractContentModel` as well as `Contact` introduced in #3699 and opted to make a full deep copy of the objects before creating the duplicates to avoid any side effects when copying to multiple target regions *(imagine a poi getting ` (Copy)` added to the name for region A first, and then the same object being used as a template for region B, adding another ` (Copy)` to it)*.
About the test, I'm torn whether the parametrization is more readable now or not. If you accept that the expression generates the full list of possible flag combinations from none to all three *(`--contacts --events --add-suffix`)*, then it is kind of readable, and I made sure that the test names do not end up as `::test_copy_pois_succeeds[parameters6]` but `::test_copy_pois_succeeds[--contacts --events]`. What might be overkill, I made sure that the test does not produce false positives from having no objects in the template region, when of course the resulting number in the target region would match up as expected, but without it telling anything about whether `copy()` and the management command are working

Issue #3721, opened 2025-06-02
Search & Replace always returns "Links were replaced successfully."


juliankehne:
### Describe the Bug

Result notification of Search & Replace always return "Links were replaced successfully." no matter what you enter in the two fields fields.

### Steps to Reproduce

1. Go to 'Broken Links'
2. Click on 'Search & Replace'
3. Choose any Link types (doesn't seem to matter)
4. Enter any random characters into the search field e.g. "QWERTASDAZXC" and anything into the replace field
5. Click 'Replace'
6. See "Links were replaced successfully."

### Expected Behavior

A clear and concise message of the actual result and no false positive messaging.

### Actual Behavior

Always the same message even if something didn't work.

### Screenshot

![Image](https://github.com/user-attachments/assets/5e2cb703-f565-4deb-ad3b-6b0c5e01112d)

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

Issue #2313, opened 2023-06-07
Add confirmation step for link replacement


MizukiTemma:
### Motivation

With #1851 link replacement functionality will be introduceed. It would be nice to have a confirmation step, where users can confirm or cancell changes before they are actually conducted.

### Proposed Solution

Run the replacement first dry, show how links will be changed in a pop up or something similar, let users either cancel or confirm, run the replacement with `commit=true` when confirmed.

### Alternatives

Leave it as it is.

laravoelk:
Currently, I think that it is more or less clear to the municipalities. We also describe the process in our Wiki and show it to them from time to time.

PR #3770, opened 2025-07-01
Improve messsage of the search & replace of links

MizukiTemma:
### Short description

This PR improves the message that is shown after using "Search&Replace" of links.

### Proposed changes

- Return how many ~~links were found~~ translations are changed due to link replace
- Show a message either "no change" or "Links were replaced successfully."

### Side effects

- Should be none (no change to mechanismus of link replacement)
- 

### Resolved issues

Fixes: #3721

### DIFF
--- a/integreat_cms/cms/utils/linkcheck_utils.py
+++ b/integreat_cms/cms/utils/linkcheck_utils.py
@@ -362,9 +362,10 @@ def replace_links(
     user: User | None = None,
     commit: bool = True,
     link_types: list[str] | None = None,
-) -> None:
+) -> bool:
     """
-    Perform search & replace in the content links
+    Perform search & replace in the content links.
+    Return number of translations in which links are replaced, to give users feedback whether any links were replaced or nothing was changed.
 
     :param search: The (partial) URL to search
     :param replace: The (partial) URL to replace
@@ -375,6 +376,7 @@ def replace_links(
     """
     log_replacement_is_starting(search, replace, region, user)
     content_objects = find_target_url_per_content(search, replace, region, link_types)
+
     with update_lock:
         for content, urls_to_replace in content_objects.items():
             content.replace_urls(urls_to_replace, user, commit)
@@ -384,6 +386,8 @@ def replace_links(
     time.sleep(0.1)
     tasks_queue.join()
     logger.info("Finished replacing %r with %r in content links", search, replace)
+
+    return len(content_objects) > 0
 
 
 def find_target_url_per_content(
--- a/integreat_cms/cms/forms/linkcheck/link_replace_form.py
+++ b/integreat_cms/cms/forms/linkcheck/link_replace_form.py
@@ -44,12 +44,12 @@ class LinkReplaceForm(forms.Form):
         self.region = kwargs.pop("region")
         super().__init__(**kwargs)
 
-    def save(self) -> None:
+    def save(self) -> bool:
         """
         This method extends the default ``save()``-method of the base :class:`~django.forms.ModelForm` to call replace_link function
         """
 
-        replace_links(
+        return replace_links(
             self.cleaned_data["search"],
             self.cleaned_data["replace"],
             region=self.region,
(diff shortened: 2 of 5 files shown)


JoeyStk (commented):
Additionally, I'm not sure if this fixes the issue described in the bug report. As I understand it, the bigger issue is, that the error message is wrong from time to time, rather than not precise :thinking:

MizukiTemma:
> Additionally, I'm not sure if this fixes the issue described in the bug report. As I understand it, the bigger issue is, that the error message is wrong from time to time, rather than not precise 🤔

I understood the scope of issue is to prevent "Links were replaced successfully." when no url was matched. @juliankehne what do you suggest?

It would be beautiful to deliver every details like "n urls were replaced in the following contents: <page 1>,...> <page x>, <event y>........, <locations z>", but should we give so much of effort only for the message? 🤔

MizukiTemma:
@JoeyStk 

it was the number of translations changed by this search&replace. not the number of links changed, that was counted in this approach 🙈 The variable name was adjusted accordingly.

MizukiTemma:
@juliankehne 
 
> I think an overview of the changes, before the links are changed, makes sense, too.

It was #2313 and sent to the idea graveyard.

JoeyStk (approved):
Thank you very much for your changes! To be honest, I'm still not one hundred percent happy with the solution, but since I don't see a better way to do it, I'll approve this PR :) Maybe the second reviewer has an idea for it? :)

integreat_cms/cms/utils/linkcheck_utils.py
```diff
@@ -362,9 +362,10 @@ def replace_links(
     user: User | None = None,
     commit: bool = True,
     link_types: list[str] | None = None,
-) -> None:
+) -> int:
```
hannaseithe, line 365: I would suggest not to return `int`here ( since we dont use the number). Especially from `save` I would expect a `boolean` if not a tuple of `(success:boolean, message: str)`. `Boolean`could be easily achieved (see below).

---------

But my bigger issue is that we do not check whether 
a) the rewrite of the translations was successful 
nor b) if the update to the DB succeeded
nor c) if the linkchecker succeeded in celery
But in the end we say: "Links were replaced succesfully."  (unless no results were found). That does not feel good, since right now we only know if some links were found.

To implement that replace_urls on the Abstract ContentTranslation would need to return `false`
a) for  `if urls_to_replace and new_translation.content == self.content`
b) and 
```
try: 
self.links.all().delete
new_translation.save()
except Exception as e:
return False
```
c) I have no idea how to do this

I leave it up to you though, to judge whether this is worth the effort or not.

integreat_cms/cms/utils/linkcheck_utils.py
```diff
@@ -385,6 +387,8 @@ def replace_links(
     tasks_queue.join()
     logger.info("Finished replacing %r with %r in content links", search, replace)
 
+    return len(content_objects)
```
hannaseithe, line 390: As a minimal change I would suggest to return `len(content_objects) > 0 `

hannaseithe (changes requested):
Thank you for this PR, this definitely improves user experience and resolves the issue described in the original issue. I would suggest though to not return the `int` but simply a boolean, otherwise it might be confusing.

Still I am not really happy with us sending the message 'Links were replaced successfully' when all we really know is that a match was found (not if we succeeded in replacing it). On the other hand I hear you about not wanting to invest too much effort for simply wanting to improve the messaging.

integreat_cms/cms/utils/linkcheck_utils.py
MizukiTemma: @hannaseithe 
That would be definately a meaningful improvement. I'll keep it as a separate issue for further discussion.

hannaseithe: Yes I will do that :)

hannaseithe: #3796

hannaseithe (approved):

Issue #3781, opened 2025-07-07
`get_referencing_translations()` searching for slugs at arbitrary position in links


PeterNerlich:
### Describe the Bug

Whenever we save a content translation, we also check whether there are any other content objects linking to this page and, if needed, update those links. This is necessary e.g. when the slug was changed with a new content translation version.

Because calculating which content object a given URL points to is relatively expensive *(at least when performing it hundreds or thousands of times)*, we first cut it down to all urls containing the page slug. The past week has proven this fatally insufficient, as a page with the **slug `integreat`** meant we also matched **all URLs starting with `https://integreat.app/`** and sending our system to check a huge number of URLs, each causing at least two distinct queries to the database.

### Steps to Reproduce

1. Change the slug of any content translations to something that is part of an overwhelming amount of links *(a slug of `e` might give near maximum carnage)*
2. Publish/update the translation
3. Notice the long wait time and eventual timeout

### Expected Behavior

The only URLs checked should be those that actually look like they have the slug of the updated translation at an appropriate location.

### Actual Behavior

All URLs that contain the slug anywhere are checked, even if the slug as (part of) the domain, region slug, or a similar slug containing it as a substring. If a very large volume of URLs match, even with an aborted request, the WSGI process continues flooding the database with queries until apache is restarted.

### Summary of discussion and updates to the description

- 2025-07-08T00:20+0200: Retroactively created issue for documentation purposes *(release note)*

PR #3782, opened 2025-07-07
Bugfix update_links_to()

PeterNerlich:
### Short description

This is the bugfix PR to fix the issue that occurred on production the past week. While a similar patch is already applied in production, this is the PR to bring a fix to our code base as well.

### Proposed changes

- Further limit any URLs to be considered to end in the slug instead of just containing it
- Require the slug to be delimited with slashes
- Eliminate the extra call to `url.internal` by directly requiring any URL to start with the `WEBAPP_URL` configured in the settings
- Even further limit any URLs to contain the fitting region and language slug

### Side effects

- If there are multiple possible domains, this implementation will no longer suffice and using linkchecks `url.internal` would be more complete. However, our `settings.py` clearly only intends one domain and so skipping the extra roundtrip through python seems a more than justified improvement to me.

### Resolved issues

Fixes: #3781

### DIFF
--- a/integreat_cms/cms/utils/content_translation_utils.py
+++ b/integreat_cms/cms/utils/content_translation_utils.py
@@ -9,6 +9,7 @@ import operator
 from functools import reduce
 from typing import TYPE_CHECKING
 
+from django.conf import settings
 from django.db.models import Q
 from linkcheck.models import Url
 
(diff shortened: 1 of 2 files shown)


MizukiTemma (approved):
Thank you 🚀

PR #3788, opened 2025-07-10
Revive MT budget column in region condition export

MizukiTemma:
### Short description

This PR fixes the bug that the column for MT budget is missing in the CVS export of region conditions.

### Proposed changes

- Revive the code lines accidentally deleted by #3718 with name adjustment 
- 

### Side effects

- Should be none
- 

### Resolved issues

Fixes: #


integreat_cms/cms/views/region_condition/region_condition_actions.py
```diff
@@ -48,6 +48,11 @@ class RegionConditionResource(resources.ModelResource):
 
     num_outdated_pages = fields.Field(column_name=_("Number of outdated pages"))
 
+    has_translation_package_been_booked = fields.Field(
+        column_name=_("Booked MT budget"),
```
JoeyStk, line 52:
```suggestion
        column_name=_("MT budget"),
```

integreat_cms/cms/views/region_condition/region_condition_actions.py
```diff
@@ -48,6 +48,11 @@ class RegionConditionResource(resources.ModelResource):
...
+        attribute="mt_budget_booked",
```
JoeyStk, line 53:
```suggestion
```

integreat_cms/cms/views/region_condition/region_condition_actions.py
```diff
@@ -48,6 +48,11 @@ class RegionConditionResource(resources.ModelResource):
 
     num_outdated_pages = fields.Field(column_name=_("Number of outdated pages"))
 
+    has_translation_package_been_booked = fields.Field(
```
JoeyStk, line 51:
```suggestion
    translation_budget = fields.Field(
```
I know "has_" as a prefix only for BooleanFields. Since this is not one anymore, I think that we should also change the variable name (if that's easily doable :) )

JoeyStk (approved):
Thank you very much! This fixes the issue for me and it looks mostly good to me (please find my suggestions below)! :)
Additionally - Since this bug has been already released, I think that a release note would be good :)

integreat_cms/cms/views/region_condition/region_condition_actions.py
MizukiTemma: I guess `attribute` corresponds to a field or a property of the model. If we change it to `mt_budget`, not the booked budget (50k, 100k, etc) is shown in the column, but the property `mt_budget` which is the number of words calculated with `mt_renewal_month` and `mt_midyear_start_month`. 

In my opinion this column should show which MT budget was booked, not how much budget each region has this year, as it was shown there before the change whether the add-on was booked or not.

Issue #3767, opened 2025-06-30
Script / Management command to change `.png` icons to `.svg` icons


JoeyStk:
### Motivation

Currently we use .png files as first and secondary level icons. In order to activate the high contrast mode, we would need to change these icons to .svgs. 

### Proposed Solution

Automatically change the references from .png to .svg so we don't have to do it manually.

- [ ] Script to upload media files in a zip folder and get path of these files
- [x] Script to exchange the reference of media files according to rows in CSV #3790 

### Screenshot

None

### Alternatives

None

### User Story

As a CMS user I want to use svgs instead of pngs and not have to change it manually

### Additional Context

High contrast mode

### Design Requirements

None

### Related Issues

Blocks app issue: https://github.com/digitalfabrik/integreat-app/issues/3352
Related app issue: https://github.com/digitalfabrik/integreat-app/issues/1021

### Summary of discussion and updates to the description

https://chat.tuerantuer.org/digitalfabrik/pl/rmcd5p16ftr85kqbxf5ufewibc

PR #3790, opened 2025-07-11
Script to change the references from `.png` to `.svg`

JoeyStk:
### Short description

This PR adds a script to change the references of media files from `.png` to `.svg` in our pages. However this is not 100% final yet, as we need to further specify some of the details.

### Proposed changes

- Add script to replace references to media files

### Side effects

- This is part one of two of #3767 

### Resolved issues

Fixes: Half of #3767

### DIFF
--- a/integreat_cms/core/management/commands/bulk_replace_page_icons.py
+++ b/integreat_cms/core/management/commands/bulk_replace_page_icons.py
@@ -1,0 +1,117 @@
+import csv
+import logging
+import re
+import select
+import sys
+from typing import Any
+
+from django.conf import settings
+from django.core.management.base import CommandParser
+
+from integreat_cms.cms.models.media.media_file import MediaFile
+from integreat_cms.cms.models.pages.page import Page
+
+from ..log_command import LogCommand
+
+logger = logging.getLogger(__name__)
+
+
+def remove_protocol(url: str) -> str:
+    """
+    Removes the protocol (http:// or https://) from the BASE_URL setting.
+
+    :param url: The URL to remove the protocol from
+    """
+    return re.sub("^[^:/]+://", "", url)
+
+
+DOMAIN = remove_protocol(settings.BASE_URL)
+
+
+class Command(LogCommand):
+    """
+    Management command to change references of media files from a source path to a target path.
+    This command reads a CSV file and then changes the reference from the first and second column (source region and path)
+    to the one of the third and fourth column (target region and path).
+    """
+
+    help = "Replace references of media files from source paths with target paths based on a CSV file."
+
+    def add_arguments(self, parser: CommandParser) -> None:
+        """
+        Define the arguments of this command
+
+        :param parser: The argument parser
+        """
+        parser.add_argument(
+            "csv",
+            nargs="?",
+            default=None,
+            help="The path to the csv file we want to read the replacements from",
+        )
+
+    def should_stdin_be_used(self, options: dict) -> bool:
+        return bool(
+            sys.platform != "win32"
+            and not sys.stdin.isatty()
+            and select.select([sys.stdin], [], [], 0)[0]
+            and not options["csv"]
+        )
+
+    def handle(self, *args: Any, **options: Any) -> None:
+        successful = 0
+        failed = 0
+
+        with (
+            open(options["csv"], newline="")
+            if not self.should_stdin_be_used(options)
+            else sys.stdin as csvfile
+        ):
+            reader = csv.reader(csvfile, delimiter=",", quotechar="|")
+            next(reader, None)
+
+            for row in reader:
+                path = {
+                    "old": (
+                        remove_protocol(row[0])
+                        .removeprefix(DOMAIN)
+                        .removeprefix(settings.MEDIA_URL)
+                    ),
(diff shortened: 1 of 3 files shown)


integreat_cms/core/management/commands/bulk_replace_page_icons.py
```diff
@@ -0,0 +1,148 @@
...
+                SELECT mf.id, mf.name, mf.parent_directory_id AS parent_id, (dt.full_path || '/' || mf.name)::TEXT AS full_path, FALSE AS is_directory
+                FROM {mediafile_table} mf
+                JOIN directory_tree dt ON mf.parent_directory_id = dt.id
+            )
+
+            -- Combine the results from both directories and media files
+            SELECT id, full_path, is_directory FROM directory_tree
+            UNION ALL
+            SELECT id, full_path, is_directory FROM mediafile_tree;
+        """)
```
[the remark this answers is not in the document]
PeterNerlich: For context: This is a raw SQL query because it needs to use CTEs *(Common Table Expressions, aka. the `WITH` `AS` thing)* to recursively build the paths of the directory structure in the media library. I think this could be immensely useful to have as a utility, but we didn't end up actually needing it for this management command.

PeterNerlich:
> 1. change `.removeprefix("https://")` to `.removeprefix("http://")` and add `.removeprefix("localhost:8000/media/")` for local test

This reminds me that we should take the domain from the settings and not hard code it

integreat_cms/core/management/commands/bulk_replace_page_icons.py
```diff
@@ -0,0 +1,145 @@
...
+        failed = 0
+
+        with (
+            open(options["path"], newline="")
+            if not self.should_stdin_be_used(options)
+            else sys.stdin as csvfile
+        ):
+            reader = csv.reader(csvfile, delimiter=",", quotechar="|")
+            next(reader, None)
+            full_path_lookup = get_mixed_tree_paths()
```
jonbulz, line 100: Is this variable used anywhere?

integreat_cms/core/management/commands/bulk_replace_page_icons.py
```diff
@@ -0,0 +1,145 @@
...
+        return {row[1]: {"id": row[0], "is_directory": row[2]} for row in rows}
+
+
+class Command(LogCommand):
+    """
+    Management command to change references of media files from a source path to a target path.
+    This command reads a CSV file and then changes the reference from the first column (source path) to the second column (target path).
+    """
+
+    help = "Replace references of media files from source paths with target paths based on a CSV file."
```
jonbulz, line 66: I'm not quite sure I completely get it. We want to use this to change .png to .csv, right? I guess this is meant to be more generic so that we can re-use this management command for other purposes in the future. But how exactly is the csv supposed to look for our use-case? I tried something like ```*.png,*.svg```, but that didn't seem to work :sweat_smile:

PeterNerlich: The input CSV needs to provide the exact mapping of the paths as stored on the server, e.g.

| old |new | comment |
|-|-|-|
| https://cms.integreat-app.de/media/sites/83/2015/09/phone-handset_icon-icons.com_48252-300x300.png | /sites/0/2025/08/phone_85134.svg | this is a comment that will be discarded |
| https://cms.integreat-app.de/media/sites/83/2016/04/calendar60.png | https://cms.integreat-app.de/media/sites/0/2025/08/calendar_26397.svg | etc. |

This means for every "old" file, identified by effectively its physical path on disk *(after stripping away any domain parts)* we assign one "new" file. The new file has to **already exist** in the media library, e.g. has to actually exist on that location on disk and have a database entry and stuff like a thumbnail generated, so essentially it has to have gone through the upload process *(I'm making a second management command `bulk_media_upload` for that, which takes a directory or zip file and spits out a list of all upload paths of the files inside after "uploading" it to the media library)*.
The mapping here can map **multiple** different "old" files to one "new" file to replace them, that is a need since we will have many logically identical icons in copies for each region individually which we also want to replace with the new version in the global library.

Given the paths to the file on disk, the associated MediaFile object is uniquely identifyable *(as opposed to only giving the file name like `a.png`, which might match many objects across multiple regions in multiple directories at the same time)*. Then that reference can be changed over to point to the new MediaFile object without any unintended side effects.

integreat_cms/core/management/commands/bulk_replace_page_icons.py
PeterNerlich: Sorry, we didn't end up using it as we found it not useful for what the script needs to be after all. I didn't want to throw `get_mixed_tree_paths()` away and forget about it, so I left it in for now. I want to migrate it to somewhere where it makes sense before merging this.

PeterNerlich:
> I'd prefer to move get_mixed_tree_paths somewhere else or to remove it from the code

Yes, I did that now. I moved it to #3870, it would fit better there – at least for not to not need to come up with it from scratch later

> Although changing the media file was imminent in the database, I had to restart apache for the changes to be visible in the form

That sounds like a caching issue to me. I need to investigate this

Issue #3804, opened 2025-07-23
Provide poi category icons without background


steffenkleinle:
### Motivation

We display the poi category icons in different places and different designs. Not in all cases we want a grey background, see e.g. https://github.com/digitalfabrik/integreat-app/pull/3372#discussion_r2221439543 (compare to the [design](https://www.figma.com/design/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=4931-2&p=f&t=qXGzktZRAJ53PrUS-0)).

Furthermore, the grey background conflicts with our new design system and/or the high contrast mode.

### Proposed Solution

Provide the icons without the background in the api. Backgrounds (if applicable) should be set directly in the frontend.

Please only change the icons in the testumgebung first and notify such that we can test for breaking changes.

### Screenshot

<img width="628" height="416" alt="Image" src="https://github.com/user-attachments/assets/a5b75b4f-7810-4f6b-b837-581c6af899f9" />

### Alternatives

N/A

### User Story

As a user I want a clean UI that follows material UI best practices.

### Additional Context

The current icons can be found here: https://github.com/digitalfabrik/integreat-cms/tree/develop/integreat_cms/static/src/svg/poi-category-icons
https://chat.tuerantuer.org/digitalfabrik/pl/jdyox7j6e7ym78ioa43cksebhw

### Related Issues

App issue: https://github.com/digitalfabrik/integreat-app/issues/2885

PR #3809, opened 2025-07-25
3804: Remove grey background from poi category icons

steffenkleinle:
### Short description

This PR removes the grey background and makes the foreground color of poi category icons configurable.

### Proposed changes

- Remove the grey circle background
- Change foreground color from white to `currentColor`

### Side effects

- This removes the background and changes the foreground color both in the CMS and the app. So perhaps you have to manually reset the color and background in the CMS now? In the app we still have to decide how it should look like with our new design system.

On top of the issue, the foreground color was changed to `currentColor`.

### Resolved issues

Fixes: #3804


MizukiTemma:
@digitalfabrik/service-team 

It looks like this after this PR
image

Should we probably add a new column for color?

JoeyStk:
There is one issue in the poi categories list. The icons appear like this now: 
image. I didn't find a fix for it just now. @digitalfabrik/cms-team Could someone please have a look at it during my absence?

MizukiTemma (approved):
@hannaseithe 

Thank you for the fix 😸

jonbulz (approved):
It looks exactly the same to me as it did before. I guess that is what Hanna's fix is supposed to do? I can confirm that the icons without the background load successfully, so I guess this works as intended :)

JoeyStk:
Big thanks to you, @hannaseithe! :blush:

Issue #3265, opened 2024-12-03
After exporting pages as pdf it opens another window when executing a different bulk action


juliankehne:
### Describe the Bug

In pages after selecting 1 or more published pages and exporting them as a pdf, then executing a different bulk action it will open a new tab.

### Steps to Reproduce

1. Go to 'Pages'
2. Select multiple published pages
3. Select bulk action 'Export published pages as pdf' & Execute
4. See that new tab opens with the export
5. Select 'Publish pages' & Execute
6. See that a new pages tab opens

### Expected Behavior

After exporting pages as pdf and then choosing a different bulk action it does not open a new tab if its not required

### Actual Behavior

Open a new tab

### Additional Information

Found in Mozilla Firefox, not sure if different browsers behave differently

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

MizukiTemma:
Successfully reproduced in the local emvironment and the prod system with Chrome.

PR #3813, opened 2025-07-29
Prevent unnecessary new windows at bulk action after PDF export

MizukiTemma:
### Short description

This PR fixes the bug that a new window always opens at every bulk action event if it is not PDF export, if once PDF export was used.

### Proposed changes

- Reset the value of `form.target` after each bulk action
- 

### Side effects

- Should be none
- 

### Resolved issues

Fixes: #3265 

### DIFF
--- a/integreat_cms/static/src/js/bulk-actions.ts
+++ b/integreat_cms/static/src/js/bulk-actions.ts
@@ -80,6 +80,7 @@ export const bulkActionExecute = (event: Event) => {
     event.preventDefault();
     const bulkAction = document.getElementById("bulk-action") as HTMLSelectElement;
     const form = event.target as HTMLFormElement;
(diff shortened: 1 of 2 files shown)


integreat_cms/static/src/js/bulk-actions.ts
```diff
@@ -98,6 +99,9 @@ export const bulkActionExecute = (event: Event) => {
         // Submit form and execute bulk action
         form.submit();
     }
+    // Reset the target to prevent a new tab from always opening regardless of the next actions
+    // after once PDF export is used
```
JoeyStk, line 103:
```suggestion
    // after the PDF export is used once
```

JoeyStk (approved):
Thank you very much! This is a simple but effect fix! :rocket:

PR #3815, opened 2025-07-30
Support debug messages for specific users

svenseeberg:
### Short description

To help debugging issues on production services and avoid the log file being spammed with thousands of debug messages by all requests, we want an option to log only requests made by specific users.

### Proposed changes
- Add a setting (list of usernames) for which debug messages should be logged
- The middleware provides the request as context for a logger filter
- the logger filter throws away all debug messages except for the list of users that should have debug messages logged

### Side effects
- If this functionality would be enabled by default, there would be added compute for each incoming request. Therefore, if disabled, I tried to skip over all relevant code paths.

:100: 

### Resolved issues

Fixes: https://chat.tuerantuer.org/digitalfabrik/pl/j7bh4qmd3b8amp3et56rw3bfwe

### DIFF
--- a/integreat_cms/core/logging_filter.py
+++ b/integreat_cms/core/logging_filter.py
@@ -1,0 +1,30 @@
+import logging
+
+from django.conf import settings
+
+
+class DebugRequestFilter(logging.Filter):
+    def filter(self, record: logging.LogRecord) -> bool:
+        """
+        Filter log records based on the current request.
+
+        If REQUEST_DEBUG_USERS is set in the Django settings,
+        this filter will only allow log records with level INFO or higher,
+        or if the current user's username is in the REQUEST_DEBUG_USERS setting.
+
+        :param record: The log record to filter
+        :return: True if the record should be logged, False otherwise
+        """
+        if settings.REQUEST_DEBUG_USERS:
+            from integreat_cms.core.middleware.debug_request import get_request
+
+            request = get_request()
+
+            if (
+                request is not None
+                and request.user.username in settings.REQUEST_DEBUG_USERS
+            ):
+                record.msg = f"Debug User: {request.user.username} - {record.msg}"
+                return True
+            return record.levelno >= getattr(settings, "LOG_LEVEL_NORMAL", logging.INFO)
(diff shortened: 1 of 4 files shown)


integreat_cms/core/logging_filter.py
```diff
@@ -0,0 +1,17 @@
...
+import logging
+import threading
+
+class DebugRequestFilter(logging.Filter):
+    def filter(self, record):
+        request = threading.local()
+        if request and request.user.is_authenticated and request.user.is_superuser:
+            if request.GET.get('debug') == '1' and record.levelno >= logging.INFO:
+                record.levelno = logging.WARNING
+                record.levelname = 'WARNING'
```
PeterNerlich, line 15: By default, `INFO` has a value of 20 and `WARNING` has a value of 30. Those can be changed I think and new levels might be introduced. You likely rather want to ensure the log level is at most `WARNING`
```suggestion
            if request.GET.get('debug') == '1' and record.levelno > logging.WARNING:
                record.levelno = logging.WARNING
                record.levelname = 'WARNING'
```

integreat_cms/core/logging_filter.py
```diff
@@ -0,0 +1,16 @@
...
+
+class DebugRequestFilter(logging.Filter):
+    def filter(self, record):
+        if settings.REQUEST_DEBUG_ADMIN_IPS:
+            from integreat_cms.core.middleware.debug_request import get_request
+
+            request = get_request()
+            return record.levelno >= logging.INFO or (
+                request is not None
+                and request.user.username in settings.REQUEST_DEBUG_USERS
```
svenseeberg, line 14: @PeterNerlich now that we have the request object we can actually use the usernames instead of IPs. That is much more convenient. We can even enabled debug logging for region editors if we cannot reproduce the bug ourselves.

svenseeberg: Alternatively, we could add a setting to the user preferences. However, I would actually prefer to solve this on the server side. `systemctl reload apache2` should work as well and should not be disruptive.

PR #3826, opened 2025-08-06
Implement database trigger constraint for slug uniqueness

### LINKED ISSUE #3060: Multiple events have the same path
### Describe the Bug

Due to multiple events using the same path, the search in the frontends is atm broken for the Ostallgäu: https://integreat.app/ostallgaeu/de/search

Another example for pois: https://integreat.app/schwaebischhall/ru/locations (`Stadtbibliothek`)

We rely on the path/url of all pages (categories, events, pois, ...) to be unique to be able to use them as IDs. This assertion seems to be broken, but should be enforced in the CMS.

### Steps to Reproduce

1. Request https://cms.integreat-app.de/api/v3/ostallgaeu/de/events/?combine_recurring=True
2. See two items with path: `/ostallgaeu/de/events/handarbeits-werkstatt-mit-begegnung-und-austausch-5`

### Expected Behavior

No events/categories/pois to have the same id.

### Actual Behavior

Two events have the same id.

### Additional Information

App issue: https://github.com/digitalfabrik/integreat-app/issues/2928
Mattermost discussion: https://chat.tuerantuer.org/digitalfabrik/pl/53jp75ehfj8mbdtwh3u56xhrua

hannaseithe:
### Short description
This is part 2 of this PR #3784. I use a database trigger before `INSERT` and `UPDATE` on pagetranslations, eventtranslations and poitranslations to assure that the slug does not already and if it does I raise a database exception.

In part 1 I have implemented application level safeguards, but as mentioned there, they are not really sufficient to guarantee a slugs uniqueness. 

##### Why is a simple DB constraint not implementable as `UniqueConstraint`?
A unique constraint basically creates an Index (can be partial when used with condition). But we need the condition "where page_id differs from other page_id". Meaning we need to be comparing two rows, which is not doable neither in Django nor SQL.

But there are ways to ensure uniqueness of slugs on the database level especially with PostgreSQL. The different tools we need are
- Database Triggers
- Denormalization of a column (not so much a tool but a pattern)
- ExclusionConstraint

A **database trigger** allows us to (before-)hook into every `INSERT`or `UPDATE` on for example the cms_pagetranslation table and then call a function that checks for the uniqueness of the slug against the other pagetranslation rows and throws an exception if its not unique.

#### Alternative solutions
 I have doubts though that we wont see noticable performance decrease on production level, especially with bulk actions. If that is the case, another option is to denormalize the region field onto the translations.

Denormalization would look like that
Create a field region on pagetranslation (or rather on AbstractContentTranslation) like this `    region = models.ForeignKey("cms.Region", null=True, editable=False, on_delete=models.CASCADE) ` 
- make it uneditable so it wont show up in forms
- and create two triggers that keep it in sync with page.region:
1. When page should ever change its region
2. If a pagetranslation should ever change its page
Both rare cases, but we need to be safe

Once we would have region denormalized on pagetranslation, we then can either chose to implement the trigger from above without the `JOIN` or use the `ExclusionConstraint` 

### How to test
Inside the shell:
```
PageTranslation.objects.filter(id=xyz).update(slug='slug-already-in-use-by other-page*')
EventTranslation.objects.filter(id=xyz).update(slug='slug-already-in-use-by other-event*')
POITranslation.objects.filter(id=xyz).update(slug='slug-already-in-use-by other-poi*')

*In the same region and language
```

All these three should throw a database error

### Side effects

- This might decrease performance significantly. Bulk Actions / Management Commands that create or modifiy pages, events and pois should be tested
- ~~the migration `make_slugs_unique` runs less than 1 minutes with current production data the first time and then less than 10 minutes the second time when all slugs are unique already. See the follow up sectio here: https://github.com/digitalfabrik/integreat-cms/pull/4082~~ EDIT: I have taken the migration out after realizing it takes more than 10 minutes, and simply added a printout during migration, to let the admin know to run the management command after deployment.

### Resolved issues

Fixes: #3060
Relates to: #3917

2026-03-31: Relevant thread about the integration of migration or running it as a management command: https://chat.tuerantuer.org/digitalfabrik/pl/ry4szsbk8frqbcgt7aw1zhe8br

~~The run of the (non dry-run) command took 45.461s for 675 updated slugs, this should be doable as a migration~~ This was not reproducable locally (see comment section below)


hannaseithe:
### Summary of options for Solution

#### Option 1: Database trigger with JOIN
- on every *-translation insert/update
- check all rows (and do a JOIN for each row), that there isnt another *-translation with a different page, the same language and the same region => otherwise throw database exception

#### Option 2: Denormalize region and Database trigger
- denormalize region from * into *-translation table (use two triggers to keep them in sync)
- create database trigger like in Option 1 without the JOIN

#### Option 3: Denormalize region & Exclusion constraint (implemented in the PoC)
- denormalize like in Option 2
- create an Exclusion Constraint

#### Option 4: Denormalize region & Create is_current field & Unique Constraint (with partial index) only for is_current version
- denormalize as in Option 2
- create is_current field on *-translation
- create unique constraint  mit WHERE is_current=TRUE bzw. condition=Q(is_current=True)

#### Pseudo-Option: Materialized View (instead of column denormalization) & Database Trigger:
We would not guarantee absolute syncronicity of Materialized Table

hannaseithe:
In Django 5.1 we get `GeneratedField` which should be a better substitute than the denormalization solution. We would not need a trigger at all and could use the GeneratedField in combination with an exclusionConstraint (Generated Field basically has internal triggers that would keep the column up to date). Since there is a PR for upgrading to Django5.2 (#3837) . I will set a BLOCKED label and defer the implementation until we have upgraded to Django5.2

dkehne (approved):
Looks good to my pr-review skill so i will spend an approval here ;-) The advisory lock approach correctly solves the race condition raised earlier, and the schema stays clean (no denormalization). 

Claude wants to do two minor things you can most probably ignore.
- renaming the trigger names from set_region_on_insert to something like enforce_slug_uniqueness (with the inline comment corrected to reflect that the trigger fires on both INSERT and UPDATE)
- and tackle a 99,99% theoretical INSERT-conflict test that could be added for each content type e.g., two consecutive objects.create() calls with the same slug, language, and region, wrapped in pytest.raises(IntegrityError). may never happen in real life...

tests/cms/models/events/test_event_translation.py
```diff
@@ -30,3 +31,32 @@ def test_when_creating_even_translations_to_automatically_create_unique_slugs()
     )
 
     assert event_translation1.slug != event_translation2.slug
+
+
+@pytest.mark.order("last")
+@pytest.mark.django_db(transaction=True)
+def test_db_trigger_prevents_duplicate_slug_on_event_translations() -> None:
+    region = Region.objects.create(name="trigger-test-region")
```
jarlhengstmengel:
```suggestion
    region = Region.objects.create(slug="trigger-test-region")
```
The tests work, but when I tested with some similar code in the shell, creating a region with 'name', created problems with the database. I'm not entirely sure what happened but I my impression was that the creating a region only with name has the potential to overwrite another region (which feels like that it shouldn't be possible but that's not part of this PR) When creating with slug it should be on the save side I think since slug is unique.

tests/cms/models/pages/test_page_translation.py
```diff
@@ -165,3 +166,24 @@ def test_get_translatable_attributes_first_translation(
...
     assert "title" in attr_names
     assert "content" in attr_names
+
+
+@pytest.mark.order("last")
+@pytest.mark.django_db(transaction=True)
+def test_db_trigger_prevents_duplicate_slug_on_page_translations(
+    create_page: Callable[..., Page],
+) -> None:
+    region = Region.objects.create(name="trigger-test-region")
```
jarlhengstmengel:
```suggestion
    region = Region.objects.create(slug="trigger-test-region")
```
See comment above

tests/cms/models/pois/test_poi_translation.py
```diff
@@ -39,3 +40,42 @@ def test_creating_poi_translations_to_automatically_create_unique_slugs() -> Non
         poi=poi2, language=language, slug="new-slug"
     )
     assert poi_translation1.slug != poi_translation2.slug
+
+
+@pytest.mark.order("last")
+@pytest.mark.django_db(transaction=True)
+def test_db_trigger_prevents_duplicate_slug_on_poi_translations() -> None:
+    region = Region.objects.create(name="trigger-test-region")
```
jarlhengstmengel:
```suggestion
    region = Region.objects.create(slug="trigger-test-region")
```
See comment above

jarlhengstmengel (changes requested):
Thank you very much for the PR! This was fun, reviewing actual SQL Code :D Code generally looks good to me. I have some nits about the tests. I tested also in the shell manually and that threw the expected error. I tested with a database dump. Result of the first run of the migration: `INFO integreat_cms.cms.utils.slug_utils - Finished >make_all_slugs_unique< after: 874.360s with 939 updated slugs`. I didn't trigger the migration a second time with the command. Since the database is migrated on the first startup after importing the dump, it should never run into that situation a second time. So as long as we are aware that it takes long time to run when we trigger the command manually, this shouldn't bother us anymore. (I was worried that it could become difficult to test things with a datadump after this is merged, but that doesn't make sense in hindsight 😅 ).
With the dump I also tested if there is a noticeable difference in performance with bulk actions. I couldn't detect a difference. With a lot of objects selected it takes long either way.

hannaseithe:
> * and tackle a 99,99% theoretical INSERT-conflict test that could be added for each content type e.g., two consecutive objects.create() calls with the same slug, language, and region, wrapped in pytest.raises(IntegrityError). may never happen in real life...

I am not entirely sure I got this point, but I added a test each that tests on .bulk_create() . Create() itself will never run into a IntegrityError since it passes through save()

hannaseithe:
> Result of the first run of the migration: `INFO integreat_cms.cms.utils.slug_utils - Finished >make_all_slugs_unique< after: 874.360s with 939 updated slugs`.

That is strange, because when I ran the command on the Testsystem I got a time duration of 45s with about 600 updated slugs.  I think I will have a look into this again, just so we know what exactly to expect during deployment.

integreat_cms/cms/migrations/0152_enforce_slug_uniqueness.py
```diff
@@ -0,0 +1,56 @@
...
+                    hash="fb919c90b1a5e6a12e4fda652741a618eb68c2f9",
```
jarlhengstmengel: Question out of interest, what do this hashes do?

jarlhengstmengel (commented):
I had another run on a database dump (dump from 27.04.2026) and it came out with this: `Apr 29 12:12:03 INFO integreat_cms.cms.utils.slug_utils - Finished >make_all_slugs_unique< after: 946.519s with 939 updated slugs`
Maybe it's just my local setup that makes it slower.

integreat_cms/cms/migrations/0152_enforce_slug_uniqueness.py
hannaseithe: To be frank: I didn't know. But claude says: 

>  The hashes (e.g. hash="fb919c90b1a5e6a12e4fda652741a618eb68c2f9") are generated by the django-pgtrigger library. They are SHA-1 hashes of the trigger's SQL definition.  

 >  What they're for: pgtrigger uses these hashes to detect when a trigger's SQL has changed since it was last installed. When you run migrations or python manage.py pgtrigger install, it     
  compares the hash stored in the migration against what the trigger currently looks like — if they differ, it knows the trigger is out of date and needs to be reinstalled.                  

> The pgid field (e.g. pgtrigger_enforce_slug_uniqueness_6fca2) is the identifier used in the database, and the last few hex characters of that ID are actually derived from the hash itself  
  (the 6fca2 in _6fca2 comes from the start of the hash fb919c90... — well, more precisely pgtrigger generates a short fingerprint from the trigger name + hash to ensure uniqueness).     
   
 > In short: they're integrity checksums managed automatically by pgtrigger — your coworker doesn't need to set or change them manually, they're generated when running makemigrations.

jarlhengstmengel (approved):
From my point of view this looks good! We probably just need to be aware that the next migration can take some time? Don't know if you are happy with the runtime results I had during my tests of the migration. But I also don't really have an Idea why that could be except for different machines.
The failing test looks like a flaky one?

hannaseithe:
> From my point of view this looks good! We probably just need to be aware that the next migration can take some time? Don't know if you are happy with the runtime results I had during my tests of the migration. But I also don't really have an Idea why that could be except for different machines. The failing test looks like a flaky one?

I have just got a data dump from prod myself, and I will run the command twice now. What I want to figure out is,  if locally things really are just 10x slower (that would mean running the command a second time would take more than an hour). Or if I simply misread the time when I ran it on the Test-System (like was ist 45s or 450s). I just want to be able to give @svenseeberg a reliable approximation of how long the migration is going to take

jarlhengstmengel:
I ran the command manually on a migrated data dump earlier today and it took something like 740 seconds

hannaseithe:
Hello @jarlhengstmengel , I have actually decided now to take out the migration and just have a printout to the terminal letting the Admin know that running the command after deployment is necessary. I think with > 10 minute migration time this is the better choice (as admittedly @PeterNerlich  suggested from the beginning :) )Thank you for your persistence and patience!

jarlhengstmengel (approved):
> Hello @jarlhengstmengel , I have actually decided now to take out the migration and just have a printout to the terminal letting the Admin know that running the command after deployment is necessary. I think with > 10 minute migration time this is the better choice (as admittedly @PeterNerlich suggested from the beginning :) )Thank you for your persistence and patience!

Understood, thanks for your persistant work! Still looks good to me!
Just one minor naming nit, I think I would like it more when the migration would just be called something like `0152_enforce_slug_uniqueness`. Eventtranslation implies a bit too much to me that it's mainly about events, while it's a general thing.

integreat_cms/cms/models/pages/page_translation.py
```diff
@@ -452,3 +453,40 @@ class Meta:
...
+                DECLARE
+                    new_region_id INTEGER;
+                BEGIN
+                    -- Look up the region for the new/updated page
+                    SELECT region_id INTO new_region_id
+                    FROM cms_page
+                    WHERE id = NEW.page_id;
+
+                    -- Set advisory lock (Postgresql specific)
+                    PERFORM pg_advisory_xact_lock(new_region_id);
```
seluianova: optional suggestion:
we could make the lock more granular by including region_id, language_id, and slug.
locking only by region_id might be overly broad, since requests with different slugs or languages don’t actually conflict.
not sure how much benefit this would bring in our case, though.
up to you whether it’s worth it.


```suggestion
                    PERFORM pg_advisory_xact_lock(
                        hashtextextended(NEW.slug || ':' || NEW.language_id || ':' || new_region_id, 0)
                    );
```

ADR/0003-database-triggers-for-complex-database-constraint.md
```diff
@@ -0,0 +1,34 @@
...
+
+# Possible Solutions
+
+- Database Triggers (CHOSEN)
+- Denormalization of the region column + database triggers
+- ExclusionConstraint + Database Triggers
+
+## Decision
+
+We use two database triggers that run on INSERT and UPDATE, to guarantee that it
```
seluianova: do you mean "three triggers" ? (page, event, poi)

ADR/0003-database-triggers-for-complex-database-constraint.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+Slugs on translation instances are assumed to be unique per language and region (but not 
+per foreign object - e.g. page_translations of the same page, can have the same slug). This
+needs to especially guaranteed as it breaks the search in the App. It is not possible to use
```
seluianova:
```suggestion
needs to be especially guaranteed as it breaks the search in the App. It is not possible to use
```
or must be

hannaseithe:
Thank you for your review @seluianova! I implemented your suggestions - especially the more granular lock was a great idea! Could you check again?

seluianova:
> Thank you for your review @seluianova! I implemented your suggestions - especially the more granular lock was a great idea! Could you check again?

@hannaseithe could you please also check weither `loadtestdata.sh` command work for you?
when I tested the change I suggested without reloading test data, everything worked, but now I can't run `loadtestdata.sh`, it fails with the error `django.db.utils.ProgrammingError: Problem installing fixture '/Users/tory/Integreat/integreat-cms/integreat_cms/cms/fixtures/test_data.json': Could not load cms.PageTranslation(pk=1): function hashtextextend(text, integer) does not exist`
not sure yet weither it's my local issue or the general one

upd. I localized the problem, see below

integreat_cms/cms/models/events/event_translation.py
```diff
@@ -196,3 +197,39 @@ class Meta:
...
+                DECLARE
+                    new_region_id INTEGER;
+                BEGIN
+                    -- Look up the region for the new/updated event
+                    SELECT region_id INTO new_region_id
+                    FROM cms_event
+                    WHERE id = NEW.event_id;
+
+                    -- Set advisory lock (Postgresql specific)
+                    PERFORM pg_advisory_xact_lock(hashtextextend(NEW.language_id || ':' || new_region_id || ':' || NEW.slug, 0));
```
seluianova:
```suggestion
                    PERFORM pg_advisory_xact_lock(hashtextextended(NEW.language_id || ':' || new_region_id || ':' || NEW.slug, 0));
```

ah, thats why it fails :)

integreat_cms/cms/models/pages/page_translation.py
```diff
@@ -452,3 +453,40 @@ class Meta:
...
+                DECLARE
+                    new_region_id INTEGER;
+                BEGIN
+                    -- Look up the region for the new/updated page
+                    SELECT region_id INTO new_region_id
+                    FROM cms_page
+                    WHERE id = NEW.page_id;
+
+                    -- Set advisory lock (Postgresql specific)
+                    PERFORM pg_advisory_xact_lock(hashtextextend(NEW.language_id || ':' || new_region_id || ':' || NEW.slug, 0));
```
seluianova:
```suggestion
                    PERFORM pg_advisory_xact_lock(hashtextextended(NEW.language_id || ':' || new_region_id || ':' || NEW.slug, 0));
```

integreat_cms/cms/models/pois/poi_translation.py
```diff
@@ -217,3 +218,39 @@ class Meta:
...
+                DECLARE
+                    new_region_id INTEGER;
+                BEGIN
+                    -- Look up the region for the new/updated poi
+                    SELECT region_id INTO new_region_id
+                    FROM cms_poi
+                    WHERE id = NEW.poi_id;
+
+                    -- Set advisory lock (Postgresql specific)
+                    PERFORM pg_advisory_xact_lock(hashtextextend(NEW.language_id || ':' || new_region_id || ':' || NEW.slug, 0));
```
seluianova:
```suggestion
                    PERFORM pg_advisory_xact_lock(hashtextextended(NEW.language_id || ':' || new_region_id || ':' || NEW.slug, 0));
```

integreat_cms/cms/models/events/event_translation.py
hannaseithe: Thank you :)

seluianova (approved):
So, besides the typo, it looks very good to me.
Thanks for the applied adjustments 🙂

During testing, I also found an issue which is related but actually more general.

Currently, when we check slugs for uniqueness, we check them case-sensitively.
This means it is possible to create slugs like "test-slug" and "Test-Slug" - they are not treated as duplicates.
However, URLs are effectively case-insensitive, which means "test-slug" and "Test-Slug" are actually the same slug.

I created a separate ticket for that, but I think the new triggers should also be adjusted accordingly. https://github.com/digitalfabrik/integreat-cms/issues/4269
Would you prefer to adjust the triggers now or it is fine to fix them later?

I will add my approval anyway but feel free to ping me if you need re-test :)

hannaseithe:
> So, besides the typo, it looks very good to me. Thanks for the applied adjustments 🙂
> 
> During testing, I also found an issue which is related but actually more general.
> 
> Currently, when we check slugs for uniqueness, we check them case-sensitively. This means it is possible to create slugs like "test-slug" and "Test-Slug" - they are not treated as duplicates. However, URLs are effectively case-insensitive, which means "test-slug" and "Test-Slug" are actually the same slug.
> 
> I created a separate ticket for that, but I think the new triggers should also be adjusted accordingly. #4269 Would you prefer to adjust the triggers now or it is fine to fix them later?
> 
> I will add my approval anyway but feel free to ping me if you need re-test :)

What a good catch. I think this is something that should be done seperately, because not only will the triggers need to be adjusted, but the management-command and the `generate_unique_slug` functionality as well. I dont think we should tighten the database constraint before the app/form level validation is adjusted, since the db constraint really is only the "fallback" here. I will edit your issue to make sure that all three levels will be dealt with when this is fixed

PR #3827, opened 2025-08-06
Fix debug user logging

svenseeberg:
### Short description
list.remove() does not return the list.


jarlhengstmengel:
Ruff fails at the moment

svenseeberg:
I think the failed test is unrelated? Can someone confirm this?

integreat_cms/core/settings.py
```diff
@@ -611,7 +611,8 @@
 REQUEST_DEBUG_USERS = [
     item.strip()
     for item in os.environ.get("INTEGREAT_CMS_REQUEST_DEBUG_USERS", "").split(",")
-].remove("")
+    if item.strip() != ""
```
PeterNerlich, line 614: Why the extra comparison with an empty string? This should be [more pythonic](https://docs.python.org/3/library/stdtypes.html#truth-value-testing):
```suggestion
    if item.strip()
```

Issue #3736, opened 2025-06-11
Improve Contact Search


hannaseithe:
### Motivation
In #3704  it became apparent that the search method on the contacts model is not optimally used. SearchQuery() is a full text search, but we attempt to use it as a partial search. This results in the backend - whenever a lexem is not found (eg. searching for "sonj" will not find "sonja") - returning all the first 50 contacts with a 0.0 rank each. And then the frontend tinyMCE plugin does the actual search work and filters from those and only shows those that partially match the query. 

Please also refer to my comment here: https://github.com/digitalfabrik/integreat-cms/pull/3704#pullrequestreview-2915628548

### Steps to Reproduce

The exact behavior as it was observed in the issue #3626 is only observable on a large data set of contacts. There exists a data dump by @MizukiTemma. But the zero ranking and simply passing on the first 50 contacts to the frontend can be observed on smaller sets

1. Make sure that you have a few contacts in your region, if not create some
2. Open the browsers dev tools and go to network (on chromium)
3. Go to a page
4. Click on Einfügen in the toolbar above the content part > Contact
5. In the Dialog that opened start typing
6. Observe inside the dev tools the return value for the "contacts" get request 
     Option A -> All contacts are returned without any sorting if there is no full text search match
     Option B -> All contacts are returned with the exact match(es) in front

### Proposed Solution
A) I would expect our backend Contact.search() method to either do actual partial filtering

### Alternatives
B) Or if we stick with the full text search for some reason, to filter out all the rank 0.0 ones
C) Another option I can see is that we don't search on the backend at all, and just return all the contact instances and let the frontend do all the work (which is what is basically happening now already, just that we limit the amount of contacts we return)

maybe D) maybe there is a way to switch between partial search and full text search within the method.

### Screenshot

![Image](https://github.com/user-attachments/assets/2d74f48e-20ee-4beb-95c3-f27c9255c1a8)

![Image](https://github.com/user-attachments/assets/4a97957e-213e-4c99-b226-f040479bc957)

In the second screenshot I print out the rankings of the first 20 results on the query 's', which are all 0

PR #3832, opened 2025-08-10
include trigram similarity to improve contact search

jonbulz:
### Short description

Include trigram similarity to improve contact search.

### Proposed changes

Django by default uses full-text search, which is based on lexemes. Essentially, it breaks text into tokens and tries to match whole words (stems). In our case, this can lead to issues when inserting contact cards, which are described in detail in #3736 
This PR introduces TrigramSimilarity to the contact search, which compares character 3-grams, thus supporting partial matches and fuzzy matches.

### Side effects

- Possibly negatively affects performance. If response times get too long, a trigram index could be introduced to improve performance

### Resolved issues

Fixes: #3736 


JoeyStk (approved):
Thank you very much! This solves the issue for me! I think we could test this out on test-cms and see how the performance is there. Could you please add a release note for this? :)

MizukiTemma (approved):
Thank you so much 🚀 I tested with dump data in Region Saarbrücken with "Annika", "Zukunftsarbeit Molschd" and "Caritas", and confirm an inprovement compared to the develo branch 😻

PR #3837, opened 2025-08-12
Upgrade to Django 5.2 & support Python 3.13

svenseeberg:
### Short description
In general it seems to work quite well. Some problems occur.

### Proposed changes
- Upgrade to Django 5.2
- Add Python 3.13 support
- We have to switch to a HTTP POST method for logging out.

### Side effects
- [x] Creating new pages breaks (`ValueError`: Model instances passed to related filters must be saved., [see conversation](https://chat.tuerantuer.org/digitalfabrik/pl/1wqgnpwonpdhzbywei4dw1r4xe)
- [x] Logout breaks: https://chat.tuerantuer.org/digitalfabrik/pl/sniyhe9ei7y47p83a96tt8u5fc
- [x] `ValueError` in terminal because Page does not exist (happens for example when creating a new page)
- [x] Creating or updating pages button isn't there anymore for roles: CMS/Service team, management or editors -> **Should be fixed with rebase!**

### How to Test
- basically everything needs to be tested both in python3.11 and python3.13
- in particular for page, event and poi we need to assure that the rollback works. Meaning when the `event/page/poi` is valid but the translation_form  or recurrence_form or contact_form not, the event, page, poi also should not be created.

### Resolved issues

Fixes: #
Blocks: #3826 #4208 


JoeyStk:
@svenseeberg Could you please rebase this PR to a more recent version? That would be great! :blush:

JoeyStk:
>  Creating new pages breaks: https://chat.tuerantuer.org/digitalfabrik/pl/1wqgnpwonpdhzbywei4dw1r4xe

I'm working on a fix for it :+1: 
I noticed one more thing: I logged in as root and when I wanted to update a page, I didn't had the option to update the page anymore, but instead only the option to "submit for review" - which is a function available to authors. I assume this could be related to the mentioned issue for new pages, but it also might be a different issue

JoeyStk:
@MizukiTemma Could you please test this branch in the next week and just check if this PR breaks any functionality. I already did it myself, but I think we need another set of eyes here :) Since I mostly tested as root, it would be great if you could test other user roles :)

JoeyStk:
### TODOS

- [ ] Creating new POI fails <details>`ValueError at /augsburg/pois/de/new/ Model instances passed to related filters must be saved.` </details>

MizukiTemma:
> ### TODOS
> * [ ]  Creating new POI fails Create a new POI, after clicking Save -> `ValueError at /augsburg/pois/de/new/ Model instances passed to related filters must be saved.`

Same with a new page and event 😿

JoeyStk:
Thank you for the review

JoeyStk:
We have somewhat of a deadline for this: support for 4.2LTE ends in April of 2026 :/

JoeyStk:
We need to update treebeard as well, because the version that we use (4.7.1) doesn't support Django5.2.

JoeyStk:
Intermediate result: We're down to 32 failing tests (which is another -30) :tada:
We are also aware of one downside of the new implementation: We're now required to save the `content_form` before the `content_translation_form` can be validated. This means that when the `content_translation_form` is not valid, we also need to find a way to roll back the save of `content_form`. The entire `post` method is wrapped inside a `transaction.atomic` block via Tree Mutex, but this does not roll back, because `post` doesn't raise an exception. Instead it renders the form to the user again with any errors.

**Next TODOS:**
- [ ] Fix remaining 32 failing tests
- [ ] Fix roll back for page/poi and event form

svenseeberg (commented):
Thanks for your work @PeterNerlich and @JoeyStk 🚀.  This seems to work well 🥳. I did not test all existing CMS functionality but clicked through most of the forms, saved objects, activated new users, etc and everything worked. I think this change is ready to be tested on the test system.

On the test system we should make sure that especially test filled out page/poi/event forms but missing translation data to see what happens. AFAICT the JS catches missing translations and in practice this should never become an issue.

Last but not least we would not have needed  all those `if not self.pk:` checks as we now create the related objects first. But they do not harm either so I would keep that for sanity reasons.

Consider this an approve :heavy_check_mark: 

*edit: the commits should be squashed though ;)

JoeyStk (approved):
(Approve by Sven)

MizukiTemma (approved):
Works as beofre so far as I tested 😸 

Thank you so much 👍

hannaseithe:
So we have refactored the views now in a way that 
a) make sense with the new constraint of **a form not being valid if its foreign key does not exist yet**:
  - a translation_form will show as invalid if its foreign_object_form hasn't been saved yet (when creating a new instance) -> e.g. event_form.save() needs to happen before we call event_translation_form.is_valid()
  - but this means also that we have ensure that an already saved foreign_object_form gets rolled-back if its dependent forms (that have foreign keys on it) turn out to be invalid.

b) the tests are not complaining anymore
c) ruff and the likes are happy as well

Therefore please re-review the PR to make sure something hasn't been broken.

@MizukiTemma

MizukiTemma (approved):
Thank you so much for all the effort 😻

integreat_cms/cms/views/pois/poi_form_view.py
```diff
@@ -193,121 +191,115 @@ def post(  # noqa: PLR0915
...
-                        translate_link(
-                            message,
-                            attributes={
-                                "href": other_translation_link,
-                                "class": "underline hover:no-underline",
-                            },
-                        ),
-                    )
+                    # only attempt to create a related contact if we are in creation mode and relevant fields have been added
+                    if new_poi_and_new_related_contact:
```
JoeyStk, line 221: I like the variable name! I think it describes itself pretty well. For me it even made the comment above obsolete

integreat_cms/cms/views/events/event_form_view.py
```diff
@@ -162,163 +165,182 @@ def post(self, request: HttpRequest, **kwargs: Any) -> HttpResponse:
...
-            event = event_form.save()
-            event_translation_form.instance.event = event
-            event_translation_instance = event_translation_form.save(
-                foreign_form_changed=(
-                    event_form.has_changed() or recurrence_rule_form.has_changed()
-                ),
-            )
+            success = True
+            with transaction.atomic():
+                sid = transaction.savepoint()
```
JoeyStk, line 204: What a good solution to the problem! Nicely done! :)

JoeyStk (approved):
Tested and worked well

jarlhengstmengel (changes requested):
Saddly page creation still seems to break: 
TypeError at /augsburg/pages/de/new/
argument of type 'method' is not iterable
Get's thrown here integreat_cms/core/utils/machine_translation_provider.py, line 119

jarlhengstmengel:
Thank you very much for thePR! The code generally looks good to me, but I encountered the bug above while clicking through the CMS and trying to create a new page.

MizukiTemma:
Hmm 🤔 I cannot reproduce the bug.

hannaseithe:
Thank you @jarlhengstmengel   I am also having trouble reproducing the bug. Can you still reproduce? And if so: Could you let me know the detailed steps you did to get to the bug?

jarlhengstmengel:
I tried to reproduce again. I could only reproduce when updating my local python to 3.13 and installing the cms with 3.13. With 3.11 everything seems to work fine. I made sure to select the right inpreter. I encountered the bug from when I was in the page tree view and clicked on `Create Page`
With 3.13 I also got this warning in the console:
`/workspace/.venv/lib/python3.13/site-packages/polymorphic/__init__.py:9: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.`
Bit of a weird warning since we pin it to 80.9. Don't know if it's related, doesn't seems so from what I get from a quick search. But I wanted at least mention it :sweat_smile:

hannaseithe:
> I tried to reproduce again. I could only reproduce when updating my local python to 3.13 and installing the cms with 3.13. With 3.11 everything seems to work fine. I made sure to select the right inpreter. I encountered the bug from when I was in the page tree view and clicked on Create Page

Its a bit embarassing, but we have not tested yet for python3.13. Doing that now and running into issues while installing already. I guess devcontainer handles that more smoothly. Once I get to your error, I will get back to you. 

And thank you so much for testing with python3.13

Edit: I got the CMS installed with python3.13 inside the venv, and I cannot reproduce your error @jarlhengstmengel . I am guessing devcontainer relies on pyproject.toml just as us normies do? Could you do `tools/install.sh --python python3.13` with my updated version now (I had to adapt the versioning of some packages).  I am eager for your feedback.

And about the deprecation warning: Claude said that is from the django-polymorphic package and that needs to be fixed by them

jarlhengstmengel:
Still getting it, but I'm closing in. It only happens when there are languages (besides german) in the language tree where machine translatable is checked. 'Testumgebung' in our testdata for example, only has german, so no problem there. But for 'Augsburg' for example I run into the bug. When I add a language to the language tree in 'Testumgebung' and select 'Machine translatable', I get the bug also in 'Testumgebung'. Trying to edit the language node afterwards, I run into a index error: IndexError at /testumgebung/language-tree/19/edit/

jarlhengstmengel:
I have the feeling that the same thing what was changed with the other forms, needs to be done with the language tree form. I managed to get the following while trying to add another language node to the language tree (after doing the things in my comment above):

```Apr 17 14:21:07 ERROR django.request - 500 Internal Server Error: /testumgebung/language-tree/new/
  File "/workspace/.venv/lib/python3.13/site-packages/django/db/backends/base/base.py", line 303, in _commit
    return self.connection.commit()
           ~~~~~~~~~~~~~~~~~~~~~~^^
  File "/workspace/.venv/lib/python3.13/site-packages/psycopg/connection.py", line 256, in commit
    self.wait(self._commit_gen())
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/workspace/.venv/lib/python3.13/site-packages/psycopg/connection.py", line 394, in wait
    return waiting.wait(gen, self.pgconn.socket, interval=interval)
           ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "psycopg_binary/_psycopg/waiting.pyx", line 213, in psycopg_binary._psycopg.wait_c
  File "/workspace/.venv/lib/python3.13/site-packages/psycopg/_connection_base.py", line 577, in _commit_gen
    yield from self._exec_command(b"COMMIT")
  File "/workspace/.venv/lib/python3.13/site-packages/psycopg/_connection_base.py", line 472, in _exec_command
    raise e.error_from_result(result, encoding=self.pgconn._encoding)
psycopg.errors.UniqueViolation: duplicate key value violates unique constraint "languagetreenode_unique_lft_tree"
DETAIL:  Key (tree_id, lft)=(6, 2) already exists.

  File "/workspace/.venv/lib/python3.13/site-packages/django/core/handlers/exception.py", line 55, in inner
  File "/workspace/.venv/lib/python3.13/site-packages/django/core/handlers/base.py", line 197, in _get_response
  File "/workspace/.venv/lib/python3.13/site-packages/django/views/generic/base.py", line 104, in view
    return self.dispatch(request, *args, **kwargs)
           ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/.venv/lib/python3.13/site-packages/django/contrib/auth/mixins.py", line 109, in dispatch
    return super().dispatch(request, *args, **kwargs)
           ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/.venv/lib/python3.13/site-packages/django/views/generic/base.py", line 143, in dispatch
    return handler(request, *args, **kwargs)
  File "/workspace/integreat_cms/cms/utils/tree_mutex.py", line 98, in innermost_function
    transaction.atomic(using=DEFAULT_DB_ALIAS, durable=False),
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/.venv/lib/python3.13/site-packages/cacheops/transaction.py", line 77, in __exit__
    self._no_monkey.__exit__(self, exc_type, exc_value, traceback)
    ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/.venv/lib/python3.13/site-packages/django/db/transaction.py", line 263, in __exit__
    connection.commit()
    ~~~~~~~~~~~~~~~~~^^
  File "/workspace/.venv/lib/python3.13/site-packages/django/utils/asyncio.py", line 26, in inner
    return func(*args, **kwargs)
  File "/workspace/.venv/lib/python3.13/site-packages/django/db/backends/base/base.py", line 327, in commit
    self._commit()
    ~~~~~~~~~~~~^^
  File "/workspace/.venv/lib/python3.13/site-packages/django/db/backends/base/base.py", line 302, in _commit
    with debug_transaction(self, "COMMIT"), self.wrap_database_errors:
                                            ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/.venv/lib/python3.13/site-packages/django/db/utils.py", line 91, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/workspace/.venv/lib/python3.13/site-packages/django/db/backends/base/base.py", line 303, in _commit
    return self.connection.commit()
           ~~~~~~~~~~~~~~~~~~~~~~^^
  File "/workspace/.venv/lib/python3.13/site-packages/psycopg/connection.py", line 256, in commit
    self.wait(self._commit_gen())
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/workspace/.venv/lib/python3.13/site-packages/psycopg/connection.py", line 394, in wait
    retu  [truncated from 4779 characters]
Apr 17 14:21:07 ERROR django.server - "POST /testumgebung/language-tree/new/ HTTP/1.1" 500 26323
```

hannaseithe:
Ok, I got the bug myself now, just noting exact steps to reproduce here: 
1. Fresh database after pruning
2. python version of venv: 3.13.13
3. Go to list view of pages
4. Select Arabic
5. See error 
```
Exception Type:	TypeError
Exception Value:	
argument of type 'method' is not iterable
Exception Location:	/home/hanna-seithe/integreat-cms/integreat_cms/core/utils/machine_translation_provider.py, line 119, in is_enabled
```

EDIT: Note to self - once this error is fixed the pdfs for the tests need to be recreated. For instructions see comment in test_pdf_export.py

hannaseithe:
Thank you so much for your testing @jarlhengstmengel . I should have got the other issue fixed now (regarding "method not being iterable"), but now I cannot reproduce the error you got in the the language tree form. I had no problems adding new tree nodes or editing them. Could you check if you can still reproduce and if yes, give me a detailed description of the steps for reproduction?

jarlhengstmengel (approved):
Thank you very much! I think everything is resolved. I can't reproduce the language tree error anymore with your latest commits.
Don't forget to let run ruff before merging!

hannaseithe:
Note to self: PDF export fix did not work

hannaseithe:
@MizukiTemma would you mind having a look at `poi_form_view.py` and `poi_form_ajax_view.py` again, just because I ran into some trouble when merging your changes there into my branch. I am positive I fixed it, but I would like your opinion as well. Thank you!

MizukiTemma (approved):

Issue #3838, opened 2025-08-13
'Submit for approval' button instead of 'Publish/Update' button


juliankehne:
### Describe the Bug

When creating a new page or updating an existing one and logged in as any user role that should be able to Publish/Update pages: Instead of the 'Publish' or 'Update' button it shows the 'Submit for approval' button.

Which means no pages can be updated or published at the moment.

### Steps to Reproduce

Login as e.g. Manager, Editor or CMS-Team User

Creating a new page:

1. Go to 'Pages'
2. Click on 'Create Page'
3. See 'Submit for approval' button

Editing a page:

1. Go to 'Pages'
2. Edit an existing page
3. See 'Submit for approval' button

### Expected Behavior

- Publish button when creating a new page.
- Publish button when editing a page that is saved as a draft.
- Update button when editing a page that is published.

### Actual Behavior

- Submit for approval button at all times when editing or creating a page for all user roles with page publishing permissions.

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

### Related Issues

Potentially related to https://github.com/digitalfabrik/integreat-cms/pull/3652

PR #3841, opened 2025-08-14
Fix page submit button

MizukiTemma:
### Short description

This PR fixes the bug that users cannot publish a page even if they have permission for "publish"

### Proposed changes

- Revive `{% has_perm 'cms.publish_page_object' request.user page as can_publish_page %}` which was deleted accidentally.
- 

### Side effects

- Release not blocked anymore 

### Resolved issues

Fixes: #3838 


PeterNerlich (approved):
I confirm this works again for the root user and infer from the code together with the diff of the PR that is thought to originally have introduced this that this is an appropriate fix and should not cause regressions for users with other roles/permissions

PR #3853, opened 2025-08-20
Extend contact trigram similarity search to all search fields

PeterNerlich:
### Short description

#3832 introduced trigram similarity for searching contacts, but only for the name field. This PR extends this to all other relevant fields of the contact and its location, as originally intended.

### Proposed changes

- Define the `similarity` score as the max trigram similarity between the search query and each of the relevant fields
  I decided to just use the expressions in the search vector itself *(it seems `TrigramSimilarity()` can take any expression, not just a string describing a field)* so we only define the fields we want once and don't run the risk of getting another hard to discover desync like this when we only change one of the definitions in the future

### Side effects

- This only affects the contact search and should be limited to showing more results, never less than without this change.
- Maybe the ideal aggregation would be not picking a max value but a kind of sum of trigram similarities. However, we would need a way to eliminate attributions to multiple fields, e.g. if a contact with name `Martina` has this name also appear as part of other details, e.g. email `martina@example.com` and website `myorg.example.com/martina/` – with a sum of all trigram similarities the `similarity` of this contact would be about 3× higher in a search for `marti` than a contact with the name `Martin` and no further details, which could pose a problem when making more complex searches.
  Maybe I'm overcomplicating things, but I think this might warrant some more discussion, especially as it would be nice to eventally use this in all other places where we implement a search, too. Though as long as we don't have such a method of "eliminating attributions to multiple fields" as I called it, I think using `Greatest()` is probably the best approximation

### Resolved issues

not opened as issue yet

### DIFF
--- a/integreat_cms/cms/models/contact/contact.py
+++ b/integreat_cms/cms/models/contact/contact.py
@@ -10,6 +10,7 @@ from django.contrib.postgres.search import (
 )
 from django.db import models
 from django.db.models import Q
+from django.db.models.functions import Greatest
 from django.urls import reverse
 from django.utils import timezone
 from django.utils.functional import cached_property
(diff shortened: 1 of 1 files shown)


MizukiTemma (approved):
🚀

PeterNerlich:
Well, tests are borked, there was some early side effect in container 13 that emptied the database. But nothing that seems to be in connection with this PR, so I'm merging this

PR #3865, opened 2025-08-30
do not override existing content object slugs unless explicitely updated

### LINKED ISSUE #3856: Too many internal broken links
### Describe the Bug

Since some weeks some region have an extremely large number of broken links in their link list. It seems (almost) all of them are internal links and were caused as contents were translated by MT.

### Steps to Reproduce

1. Go to locations and create a new location: "Test Ort"
2. Go to pages, create a new page
3. Link the new location in the page
4. Translate the page into a different language, e.g. english
5. See that the location is linked as e.g. /en/locations/test-ort
6. Translate the location into the same language
7. See that the location link changes with the translation, e..g /en/locations/test-location
8. See that the link in the page still references the german location link and is broken
### Expected Behavior

If a link to a location exists and is valid (i.e. /en/locations/test-ort), it should not be changed by translating the location

### Actual Behavior

The location slug is changed by the translation and generates broken links

### Additional Information

The thread (original problem report): https://chat.tuerantuer.org/digitalfabrik/pl/mberkukhni848p4smhtmt8noka
The generall problem between the link checker and our MT feature may have somthing to do: #3831 

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

jonbulz:
### Short description

Content object slugs are generated based on attributes, i.e. title or name. If an object already exists and its title is changed, this also changes the slug. If this object is referenced somewhere, this generates broken links. Especially in combination with automatic translations, this can result in a large amount of broken links.

### Proposed changes

- Changing the title of objects still triggers the slugify function, but if the object instance already exists and has a slug, the existing slug is used instead of generating a new one

### Side effects

- It is now possible to have a mismatch between slugs and titles of objects. This should be communicated to users carefully.

### Resolved issues

Fixes: #3856 

### DIFF
--- a/integreat_cms/cms/constants/machine_translatable_attributes.py
+++ b/integreat_cms/cms/constants/machine_translatable_attributes.py
@@ -2,4 +2,4 @@
 This module defines all attributes which can be machine-translated.
 """
 
-TRANSLATABLE_ATTRIBUTES = ["title", "content", "meta_description"]
+TRANSLATABLE_ATTRIBUTES = ["title", "content", "meta_description", "slug"]
--- a/integreat_cms/cms/utils/slug_utils.py
+++ b/integreat_cms/cms/utils/slug_utils.py
@@ -68,7 +68,8 @@ def generate_unique_slug_helper(
     :return: An unique slug identifier
     """
     kwargs: SlugKwargs = {
-        "slug": form_object.cleaned_data["slug"],
+        "slug": form_object.cleaned_data["slug"]
+        or getattr(form_object.instance, "slug", ""),
         "cleaned_data": form_object.cleaned_data,
         "manager": form_object.Meta.model.objects,
         "object_instance": form_object.instance,
--- a/integreat_cms/cms/views/utils/slugify_ajax.py
+++ b/integreat_cms/cms/views/utils/slugify_ajax.py
@@ -63,6 +63,12 @@ def slugify_ajax(
     ):
         raise PermissionDenied
 
+    if object_instance and getattr(object_instance, "slug", None):
+        # if a content object already has an existing slug
+        # we don't want to override it by accident when changing e.g. the title
+        existing_slug = object_instance.slug
+        return JsonResponse({"unique_slug": existing_slug})
+
     form_title = slugify(json_data["title"], allow_unicode=True)
     kwargs: SlugKwargs = {
         "slug": form_title,


hannaseithe:
> AFAICT, the situation is a bit different: This should not result in "broken" links but redirects. We flag links with redirects so that they can be updated. However, they are not broken in the strictest sense.

As we learned from @dkehne / @ulliholtgrave  internal redirects are actually "broken links" as they do not work on the native app

hannaseithe (commented):
I agree with your changes, but I would also have changed update_permalink.ts, Where the actual post request to slugify_ajax is triggered on every focusout event on the title input (line 43). If possible we should also stop those unnecessary post requests, once a page (or poi or ...) has been created. But I am not entirely sure if we can actually tell on the frontend if we are dealing with a new page (or etc) or not, but if we can, we should definitely do that.

hannaseithe:
> What is the current state of this PR after the discussion [here](https://chat.tuerantuer.org/digitalfabrik/pl/kn39rqxazf8j7x8cmis8xos4zh)?

Overall it appears that the handling of the >slug update on title change< needs a bit more discussion.

This PR is not really a fix to the original problem aka "preventing more broken internal links in newly created municipalities". We have learned that there is actually a functionality that does update internal links if the slug has changed - see #2530. The large of amount of not updated internal links was due to the facts that those translations had not been updated since after the 2530 PR was merged, therefore that mechanism was never triggered. 

Still I do see value in this PR (3865), because changing the slug (and thereby the URL) of a page/poi/event, should not be made arbitrarily, because any external links (or bookmark) to that page/event/poi will be broken on the native app. For the web app the redirects should simply forward to the actual content though.

@svenseeberg on the other hand brought up the SEO argument, that the URL should not diverge too much from the actual content of a site, and that he doubts that the people updating the content are aware of that and will therefore update in such cases the slug manually. 

But maybe a compromise could be found to have a message shown when the title was changed, whether they want to change the slug as well, but to make it clear that they should only do so if the slug has become meaningless.

jonbulz:
I did some more research as to why slugs are changed by machine translations. As far as I'm aware, it is still open for debate what the expected behavior should be. However, I updated the code in this PR to at least be consistent. A quick summary:
### Current situation
Currently, slugs are automatically changed for content objects when their name/title changes. There are 2 mechanisms to ensure this:
- When a user edits the form, there is an event listener that updates the slug field *in the form* when the title field is changed. This way, when the form is submitted, the changes to the slug will be in the updated values.
- Through ContentTranslationForm.is_valid -> ... -> CustomContentModelForm.clean_slug, we generate the unique slug for the object, which, unless a value is explicitly passed in the updated values, uses the fallback (i.e. title or name).

### So why does MT change my slugs?
In MachineTranslationApiClient, we call content_translation_form.is_valid, but the translated data does not contain the slug. Thus, we revert to the fallback, which is generating a new slug from the title/name.

### What do these changes do?
This adds checks in both aforementioned mechanisms to only generate a new slug if there is no existing slug. Thus, if the title is changed (either by a user or by machine translation), the slug remains unchanged. Slugs can still be changed explicitly by users. Slugs are now also translatable attributes such that users can use MT to update slugs in multiple languages.

PR #3870, opened 2025-09-07
Script / Management command to bulk upload to media library

PeterNerlich:
### Short description

Add a management command to bulk upload files from a directory or zip file to the media library

### Proposed changes

- Add a management command to run every file in a directory and/or zip file through the media library upload form
  This is needed to ensure each file is processed just as if it was uploaded individually by a user *(like downscaling exceedingly large images or thumbnail creation)*

### Side effects

- As this is a management command, the rest of the system should not be affected
- As images to be added to the media library are being "uploaded" through the media upload form, there should be no deviation from manually uploaded media and thus no potential for side effects in other sub systems from unexpected data

### How to test

1. Read the new section `bulk_media_upload` in the documentation (`management-commands.rst`). Check that it fits the document in style and content, and that there are no broken or likely unintentional reStructured text sequences. Check that it makes sense and is helpful *(revisit this step once you played around with the command)*.
2. Run `tools/integreat-cms-cli bulk_media_upload` without any arguments. Check that it only shows the help text. Check that the help text makes sense, is helpful and does not contradict the documentation.
3. Run the command with various combinations of arguments, checking that it does what you would expect and reports what it does correctly in stdout and the generated csv file. Try combinations that you would expect to be valid as well as invalid ones.
   You will want to prepare different situations in the media library and directory / zip file. Here is an incomplete list of things you might test *(spot checking, the full power set of all possible combinations will far exceed any time we have available for testing)*:
   - file in the input zip file / directory / both simultaneously
     *(not critical for what we currently need this for)*
   - file in the root of the input zip file / directory
   - file does/doesn't already exist in the media library / already exists but as a directory
   - target directory already exists / doesn't exist in the media library
   - …

### Resolved issues

Fixes: #3767

### DIFF
--- a/integreat_cms/core/management/commands/bulk_media_upload.py
+++ b/integreat_cms/core/management/commands/bulk_media_upload.py
@@ -1,0 +1,303 @@
+import io
+import logging
+import mimetypes
+from pathlib import Path
+from typing import Any
+from zipfile import Path as ZipPath
+from zipfile import ZipFile
+
+from django.core.files.uploadedfile import InMemoryUploadedFile
+from django.core.management.base import CommandError, CommandParser
+from django.db.utils import IntegrityError
+from django.http import QueryDict
+from django.utils.datastructures import MultiValueDict
+
+from integreat_cms.cms.forms.media.upload_media_file_form import UploadMediaFileForm
+from integreat_cms.cms.models.media.media_file import Directory, MediaFile
+from integreat_cms.cms.models.regions.region import Region
+
+from ..log_command import LogCommand
+
+logger = logging.getLogger(__name__)
+
+
+class Command(LogCommand):
+    """
+    Management command to upload files to the media library in bulk and produce a CSV of the upload locations.
+    """
+
+    help = "Bulk upload files to the media library and produce a CSV of the upload locations."
+
+    def add_arguments(self, parser: CommandParser) -> None:
+        """
+        Define the arguments of this command
+
+        :param parser: The argument parser
+        """
+        group_input = parser.add_argument_group("Input")
+        group_input.add_argument(
+            "--zip",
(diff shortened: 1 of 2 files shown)


PeterNerlich:
> Do you think you could rebase onto develop, so that we can test the current state of this? :)

Done!

> Additionally, do you think you could leave a short description on how to test this section? :)

I added a segment. Please ask me if it leaves any open questions. Testing this will unfortunately be an unsatisfying job, you will need to rely on reading and understanding the code beyond spot checking to deduce that it will probably work across all possible values.

> And lastly, is there still something blocking this PR or is it already ready for review? :)

We could already merge the PR as is, even though I left in two TODO marks that ruff will complain about, but it is functional.
I would like to explore the avenue of trying to make a QuerySet returning a mix of multiple model types, so we can interact with mixed Directory and MediaFile objects using a [`Path`-style API](https://docs.python.org/3.11/library/pathlib.html#pathlib.Path). That didn't seem very urgent yet though, so I left it as is, with the raw query we built just shoved into `media_utils.py` without using it anywhere…

integreat_cms/core/management/commands/bulk_media_upload.py
```diff
@@ -0,0 +1,297 @@
...
+            "csv",
+            help="Path to which to write the CSV information of successfully uploaded files, along with their new location on disk",
+        )
+
+    def handle(self, *args: Any, **options: Any) -> None:
+        # INPUT VALIDATION & PREPARATION
+        if not (options["zip"] or options["dir"]):
+            # We need at least one of the options, else we cannot do anything
+            raise CommandError("Specify --zip or --directory")
+        if options["region"] is not None:
```
jonbulz: Is it a supported use-case that no region is given?

PR #3872, opened 2025-09-09
Experiment: Clone DB to isolate tests

### LINKED ISSUE #3777: Isolate tests on database level
### Motivation

Our tests are sometimes flaky and can influence each other. This is last but not least fostered by the different approaches the django framework and pytest have to testing, and the way *pytest-django* tries to frankenstein the two together.
One effect is that test data is loaded into the database once and then every test is just run one after the other. While this works for tests that do not rely on being able to complete transactions, since then the test suite can wrap the whole test in a transaction and roll back any changes it made afterwards, tests that do just empty the database afterwards, since django test class expects to be under the django system where whatever test data is needed is loaded into the database at the start of each test anyways – which doesn't happen under in the *pytest-django* environment, dooming all non-transactional tests running after the first transactional one. While we try to patch that up by marking all transactional tests to run last, there seem to be still other effects where objects expected to be in the database suddenly cannot be found at some point.

### Proposed Solution

While an alternative for `transactional=True, serialized_rollback=True` *(that only half-works with pytest-django)* has been proposed on the django project along with a proof-of-concept that also showed greatly improved performance, we will not want to wait for this to be released and available to us, since we are still quite behind the newest django version. Instead, and possibly also easier given our altered situation under *pytest-django*, we should use the concept of cloning database instances used in the proof-of-concept and implement it in a form best suited for our use.

I propose a set of *(pytest)* fixtures to create cloned instances of the current database as some sort of stack, with which data can be prepared for use in multiple tests as a template. Each tests will finally create a last clone to create an environment that can be safely discarded after the test finishes, keeping the parent state unchanged for any subsequent tests to use as basis.

### Alternatives

- Wait for the solution to land upstream.
  This seems infeasible as it might take a while for this to be finished and land in a django release, at which point *pytest-django* will need to adapt to expose any new flags, while the possibility exists that, much like currently the purging of data after a test is and the loading of data before the next test isn't handled by django or the pytest-django glue code, the new functionality will not be exposed in an entirely usable capacity. Furthermore, we will still need to navigate the dependency jungle to get our project to the newest django version when it arrives
- Deal with the current state of affairs.
  With a slowly growing test suite, we are getting more and more instances where pipelines fail because of side effects and race conditions that should not occur. If we want to be able to rely on and in any reasonable capacity work with our test suite in the future, we need to address this

### User Story

As a **developer** I want **to be able to quickly check whether my changes impacted some unexpected aspect of the system** so that **I don't code up medusa.git but continually improve the quality of the project**.

### Related Issues

#765 #2119 #2711 #2978 #3343 #3370

PeterNerlich:
### Short description

This is one ongoing experiment trying to reign in the growing chaos in our tests

### Proposed changes

- Implement a function `snapshot_db()` to clone the current database and switch out djangos connection, and reverting it afterwards
- Provide a *session* scoped fixture to make an empty database snapshot *(all other fixtures depend on this one way or another so we can be sure that this will actually be an empty one)*
- Provide a *session* scoped fixture to make a snapshot and fill it with the test data
- Provide a *function* scoped fixture to just make a snapshot. This is what every test needing the db should use, and replaces the `load_test_data` fixture for tests

This experiment is still **Work In Progress**.

### Side effects

- Side effects between tests should be mitigated with this

### Resolved issues

Fixes: #3777

### DIFF
--- a/tests/api/test_api_chat.py
+++ b/tests/api/test_api_chat.py
@@ -26,7 +26,10 @@ default_kwargs = {
 
 
 @pytest.mark.django_db
-def test_api_is_chat_enabled_for_user(load_test_data: None) -> None:
+def test_api_is_chat_enabled_for_user(
+    test_data_db_snapshot: None,
+    db_snapshot: None,
+) -> None:
     """
     Check that whether a user is chat beta tester is stored in the DB
 
@@ -48,7 +51,9 @@ def test_api_is_chat_enabled_for_user(load_test_data: None) -> None:
 @pytest.mark.django_db
 @patch("integreat_cms.api.v3.chat.user_chat.UserChat.zammad_request")
 def test_api_chat_missing_auth_error(
-    zammad_request: Mock, load_test_data: None
+    zammad_request: Mock,
+    test_data_db_snapshot: None,
+    db_snapshot: None,
 ) -> None:
     """
     Check that missing/wrong auth information leads to an error
@@ -79,7 +84,8 @@ def test_api_chat_missing_auth_error(
 )
 def test_api_chat_incorrect_server_error(
     mock_zammad_url: Mock,
-    load_test_data: None,
+    test_data_db_snapshot: None,
+    db_snapshot: None,
 ) -> None:
     """
     Check that incorrect server url leads to an error
@@ -124,7 +130,8 @@ def test_api_chat_first_chat(
     messages: Mock,
     create_ticket: Mock,
     get_zammad_user_mail: Mock,
-    load_test_data: None,
+    test_data_db_snapshot: None,
+    db_snapshot: None,
 ) -> None:
     """
     Check that sending a message from a never seen-before device_id creates a new chat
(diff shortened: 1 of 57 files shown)


MizukiTemma:
This PR has not been worked for a long time but is usuful as inspiration for test improvement: see the conversation [here](https://chat.tuerantuer.org/digitalfabrik/pl/6wwy6hi9w7ymxn54r94xw3i71e)

MizukiTemma:
@jonbulz 
Is this PR still inspiring?

jonbulz:
we decided to got with `@pytest.mark.order("last")` in #4441 to try and fix the ordering dependency issues. AFAIK, that seems to work. I never got around to compare the transactional approach to the clone DB approach in terms of performance. 
I think the flakiness issue is the more pressing one, and if that is actually fixed with #4441 , we can close this PR after #4441 has landed and is confirmed working.

MizukiTemma:
> we decided to got with `@pytest.mark.order("last")` in #4441 to try and fix the ordering dependency issues. AFAIK, that seems to work. I never got around to compare the transactional approach to the clone DB approach in terms of performance. I think the flakiness issue is the more pressing one, and if that is actually fixed with #4441 , we can close this PR after #4441 has landed and is confirmed working.

Alles klar 👍 
Thank you for the updates 😸

jonbulz:
looks like things are more stable now, after #4441. might re-open if the ordering-dependency issue re-surfaces, but safe to close for now

PR #3877, opened 2025-09-12
Mark `test_copy_pois_succeeds` as last

hannaseithe:
### Short description
Our test were failing with errors such as `User matching query does not exist` etc. because the test `test_copy_pois_succeeds` was not marked as last. That resulted in non-transactional tests (aka those that use the fixture `load_test_data`) that were executed after that test, starting with an empty database and therefore failing with errors.

### Proposed changes

- added the marker to be last to the test
- also set the database to `transaction=True, serialized_rollback=True` (I am not entirely sure that is necessary, but since we do that everwhere else where we use `load_test_data_transactional`, I added it here for consistency

### Side effects

- I hope none

This PR has no issue

### Resolved issues

Fixes: This PR has no issue

### DIFF
--- a/tests/core/management/commands/test_copy_pois.py
+++ b/tests/core/management/commands/test_copy_pois.py
@@ -43,7 +43,8 @@ parameters = tuple(
 @pytest.mark.parametrize(
     "parameters", parameters, ids=[" ".join(x) for x in parameters]
 )
-@pytest.mark.django_db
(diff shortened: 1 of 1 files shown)


PeterNerlich (approved):

MizukiTemma (approved):
😹 🚀 🎉 

Thank you!

PR #3883, opened 2025-09-16
Add publiccode.yaml

svenseeberg:
### Short description
Add publiccode.yaml for listing in https://interoperable-europe.ec.europa.eu/eu-oss-catalogue

### Proposed changes
Add publiccoded.yaml defined by https://yml.publiccode.tools/schema.core.html

### Side effects
- None

N/A

### Resolved issues
N/A


publiccode.yml
```diff
@@ -0,0 +1,62 @@
...
+          It manages all pages in the same hierarchical structure across all languages. It supports automatic and manual translations.
+    documentation: "https://wiki.integreat-app.com/"
+    apiDocumentation: "https://digitalfabrik.github.io/integreat-cms/api-docs.html"
+    features:
+       - Content Management
+       - Multilingual
+       - Multitenancy
+       - Translation Management
+    screenshots:
+        - "https://integreat-app.de/wp-content/uploads/integreat-cms-screenshot.png"
```
svenseeberg, line 35: AFAICT @JoeyStk added screenshots to the repo at some point. I did not find them immediately. It would probably be better to use screenshots in this repo.

MizukiTemma: #3883

publiccode.yml
```diff
@@ -0,0 +1,62 @@
+publiccodeYmlVersion: "0"
```
JoeyStk, line 1: I'm sure this is a dumb question but why do we start with 0? :)

publiccode.yml
```diff
@@ -0,0 +1,62 @@
...
+  - Germany - City of Munich
+  - Germany - City of Dortmund
+developmentStatus: stable
+softwareType: standalone/web
+description:
+  en:
+    shortDescription: The multilingual content management system for the Integreat App
+    longDescription: >
+          The Integreat CMS is a headless content management system for the Integreat App (https://github.com/digitalfabrik/integreat-app).
+          It manages all pages in the same hierarchical structure across all languages. It supports automatic and manual translations.
```
JoeyStk, line 26:
```suggestion
          It manages all pages in the same hierarchical structure across all languages. It supports machine and manual translations.
```
Super duper nit: We try to use machine translation instead of automatic translation in the system as we once noticed that automatic translation is not clear to everyone right away, but I'm sure there are people who would disagree

publiccode.yml
JoeyStk: Do we need to change the path of the screenshot?

JoeyStk (approved):
Thank you very much! Looks good to me! :rocket:

JoeyStk:
This is not testable for people who don't have access to the code, right? Then I would add the `not testable` label :)

publiccode.yml
svenseeberg: You're right, the spec is currently at "v0.4.0".

publiccode.yml
```diff
@@ -0,0 +1,62 @@
+publiccodeYmlVersion: "0"
+name: Integreat CMS
+applicationSuite: Integreat Platform
+url: "https://github.com/digitalfabrik/integreat-cms.git"
+landingURL: "https://integreat-app.de/en/"
+isBasedOn: "https://github.com/django/django.git"
+softwareVersion: "2025.9.0"
+releaseDate: "2025-09-11"
+logo: "https://integreat-app.de/wp-content/uploads/logo-aktuell.svg"
```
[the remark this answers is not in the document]
svenseeberg: This specifically is the landing page CMS where I uploaded the logo.

Generally: I think we don't need it anymore. But for some time the app was still using the old API scheme.

svenseeberg:
I think we should add an automatic version bump before merging this.

svenseeberg:
> Is this PR ready to be merged or does it need some more work? :)

I think we should add an automatic version bump before merging this.

PR #3890, opened 2025-09-17
Upgrade node to version 22

hannaseithe:
### Short description
As part of our frontend dependency upgrades and on advice from Leandra we have decided to require Nodejs version 22 or higher to be installed. This PR adds the requirement to our `package.json` and update the example instructions in the docs

### Proposed changes

- Instructions to install nodejs 22 or higher

### Side effects
- Previously, our package did not specify an engine requirement. Side effects might occur where it technically still worked on an older version but now refuses to. We expect our project to be running on reasonably modern setups though, so this should mostly be of no concern, hopefully

### Resolved issues

This relates to: #3821 

Fixes: #

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ To configure your development environment on your system, please follow these st
 
 1. Ensure that the following packages are installed alongside your preferred IDE:
    - `npm` version 7 or later
-   - `nodejs` version 18 or later
+   - `nodejs` version 22 or later
    - `python3` version 3.11 or later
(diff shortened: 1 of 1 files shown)


jarlhengstmengel (changes requested):
Thanks for the PR! I saw that Node Version is also documented in the README, could you also update that?

jarlhengstmengel (approved):
Thank you! Looks good to me!

Issue #3773, opened 2025-07-03
`Uncaught TypeError` in `addOpeningHoursListener`


MizukiTemma:
### Describe the Bug

`document.getElementById("id_use_location_opening_hours")` is often null and causing an error.

It's nice to fix it, although there is no influence on users.

### Steps to Reproduce

1. Go to any page of the CMS system (except the contact form)
2. Open a developer tool
3. See the error "Uncaught TypeError: Cannot read properties of null (reading 'addEventListener') at HTMLDocument.addOpeningHoursListener (index.tsx:115:61)"

### Expected Behavior

No error

### Actual Behavior

An error

### Screenshot

![Image](https://github.com/user-attachments/assets/fe1255b2-2bd3-49df-8e41-7bc63cfd03ff)

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

PR #3894, opened 2025-09-18
Fix addOpeningHoursListener null error when element is not there

dash-baijayanti:
### Short description

We are trying to attach the eventListener to the `useLocationOpeningHours`-element without making sure first that it is really there. Additionally we also do not check if `openingHoursTextarea` is there afterwards, which is also fixed in this PR.

### Proposed changes

- we check that `useLocationOpeningHours` is there and only attach the eventListener if it is there
- we check that `openingHoursTextarea`  is there before the MutationObserver is created in which it is used

### Side effects

- N/A

- After fixing the error reported in the issue the creation of MutationObserver threw another error, that we decided to fix here as well, as it was just "hidden" by the first error

### How to test

- check that the error in the console of your browser is gone
- check in the contact form that everything is still working (especially the eventListener & MutationObserver)

### Resolved issues

Fixes: #3773 

### DIFF
--- a/integreat_cms/static/src/js/pois/opening-hours/index.tsx
+++ b/integreat_cms/static/src/js/pois/opening-hours/index.tsx
@@ -112,35 +112,47 @@ export const addOpeningHoursListener = () => {
 
     let updateLatch = 0;
 
-    document.getElementById("id_use_location_opening_hours").addEventListener("change", (e) => {
-        // Only reset if we changed to adopt the opening hours from the location
-        if ((e.target as HTMLInputElement).checked) {
-            updateLatch += 1;
-            const thisTimeAround = updateLatch;
-            setTimeout(() => {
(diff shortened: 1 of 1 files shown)


hannaseithe:
Hello Banni, thank you for the PR. I am glad to see you got the cloning of the repo working!

(As we have already discussed) it seems like you are fixing other things in this PR that are not related to the original issue. Basically the best practice here is: 
Create one branch for one issue >
commit and publish your changes on this branch > 
create a PR based on this branch

If you chose to deviate from the original issue for good reason (and simply mixing up your branches is not a good reason ;) ), you need to comment on the reasoning for that.

Also the PR template is not sufficiently filled in.

Please do not close this PR again to open a new one, but let's focus on fixing this as it is. If you are unsure how to seperate the relevant changes in your branch, from changes that belong to other issues, I suggest we can do that together on Wednesday.

PR #3899, opened 2025-09-23
Update core-js

PeterNerlich:
### Short description

Update core-js, supplying polyfills for older browsers (no major version changes)

### Proposed changes

- Update core-js

### Side effects

- Probability of impact rising the more outdated any browser visiting our CMS is
- This also includes the `"engines": {"node": ">=22"}` in the `package-lock.json` that should really have been part of #3890

### How to test

Comprehensively testing this might be difficult, as polyfills fill in functionality when running on older browsers that don't implement it yet. Conversely, on modern browsers, all calls will run the browser implementation, not the "backup" one by core-js. Completely testing this would require (very) old browsers, likely also older hardware or at least operating systems, with no defined boundary to what we consciously do not want to support anymore.

Since this only affects municipalities and not the target audience "in the field" dependent on the information in our system, I propose to regard it sufficient to have reviewers test with one or two of their browsers and call it a day. Any serious bug resulting from this *(which I regard as very improbable)* will be reported by municipalities, who will also be in a better position to work around these bugs and switch to different browsers than scattered people "in the field".

### Resolved issues

Related to #3821

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "chart.js": "^4.5.1",
     "classnames": "^2.5.1",
     "copy-webpack-plugin": "^13.0.1",
-    "core-js": "^3.37.1",
+    "core-js": "^3.45.1",
     "css-loader": "^7.1.2",
     "css-minimizer-webpack-plugin": "^7.0.2",
     "dropzone": "^6.0.0-beta.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "chart.js": "^4.5.1",
         "classnames": "^2.5.1",
         "copy-webpack-plugin": "^13.0.1",
-        "core-js": "^3.37.1",
+        "core-js": "^3.45.1",
         "css-loader": "^7.1.2",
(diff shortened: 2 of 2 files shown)


JoeyStk (approved):
Thank you very much! I went through the modules that use `.map()` and checked them. I couldn't find any strange behavior. I think this is okay to be merged :) :rocket:

jarlhengstmengel:
Could you rebase? This still contains the make slugs unique migration :see_no_evil: sorry for taking me so long to look at this again :sweat_smile:

package.json
```diff
@@ -24,7 +24,7 @@
     "chart.js": "^4.4.3",
     "classnames": "^2.5.1",
     "copy-webpack-plugin": "^13.0.1",
-    "core-js": "^3.37.1",
+    "core-js": "^3.45.1",
```
JoeyStk: I think we could even go up to 3.46.0 :) For more context see [here](https://github.com/digitalfabrik/integreat-cms/pull/3953)

Issue #3797, opened 2025-07-21
Improved date validation for events


dkehne:
### Describe the Bug
Eventsdates like 01.01.0005 cause app to break - see screenshot

### Steps to Reproduce
Create an event with year 0005

### Expected Behavior
Only dates in future should be allowed with valid years

### Screenshot
<img width="659" height="164" alt="Image" src="https://github.com/user-attachments/assets/a5e0a13d-9e6a-4533-a8f1-df4ff958a6be" />

PR #3919, opened 2025-10-02
Added new date validation for events

dash-baijayanti:
### Short description

- This PR fixes the issue where events could be created with invalid dates such as  `01.02.00588` . The validation has been updated to ensure that event dates are realistic and consistent with the current time zone.

### Proposed changes

- Added ` timezone ` import for consistent date and time handling.

- Implemented validation logic to enforce:

   -  `start_date` cannot be in the past.

   -  `end_date` must not be earlier than start_date.

- Prevented entry of invalid years (e.g., 00588, 0022) that were previously accepted.

- Adjusted the event form logic so that the default `start_date` is today and `end_date` cannot precede it.

-  In Translation Context for `django.po`  added a new user-facing error message:

     "The start of the event can't be before today. Please choose today or a future date."

### Side effects

- Forms with outdated or invalid event dates may now be blocked.

- Any feature or workflow that depends on event dates will now respect the updated validation rules.

- The implementation follows the issue description and the intended design. No deviations were introduced.

### How to test
 

- Open the event creation or editing form.

- Try setting a start_date in the past → should trigger validation error.

- Try setting a year with unrealistic digits (e.g., 00588, 0022) → should be rejected.

- Set start_date to today and end_date to today or later → should save successfully.

- Repeat tests in different time zones to verify consistent behavior.

### Resolved issues

Fixes: #3797 


hannaseithe (changes requested):
Thank you very much for this PR Banni, especially dealing with the failing tests. :pray:

The validation is working as expected and the tests are not failing anymore!

I have just a few minor change requests. 

Also please remember that later before you merge (after you got two approved Code Reviews) you should squash your commits into one. We do that to keep our commit history clean. There are several ways to do that, but I would recommend an interactive rebase locally, and then force pushing to remote (and then merge).

Issue #2789, opened 2024-05-07
Error `PushNotificationTranslation matching query does not exist`


MizukiTemma:
### Describe the Bug

This will be a supplement to #2722 

### Steps to Reproduce

1. Create a region which has only English as their language
2. Create a push notification in this region and select "Use main language..." and check "Stadt Augsburg"
3. Save and send
4. See the error

* It doesn't have to be "Stadt Augsburg", but a region whose name appears earlier than that of the english-only region in the alphabetical order.

No unexpected error occrus while creating and sending a push notification in Augsburg with only German translation and the English-only region selected (basically, otherway around). In this case you get an warning that the title for English translation is required and everything is fine, thanks to #2727

I know such a case like this will mostlikely not exist in the live system since the regions have usually German as default language (or at least one of their languages), but that's only an assumption 😅 🙈 

### Expected Behavior

Warning shown that title for other languages are required.

### Actual Behavior

Error

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

MizukiTemma:
@juliankehne 
Yes, this problem was already known at the time of review & merge. will be fixed as #3926

Issue #3926, opened 2025-10-08
Fix redirect after failing to send/schedule a push notification


MizukiTemma:
### Describe the Bug

If a push notification is not sent successfully, users are not successfully redirected and the form is not rendered well.

### Steps to Reproduce

1. Manipulate the code line as shown below
2. Go to "News", click "Create News"
3. Enter some title tec.
4. Click "Save & Schedule"
5. See the form is not rendered well

Modify the code to fail to send a push notification

https://github.com/digitalfabrik/integreat-cms/blob/c967ee69174ec127a27ff05c9aa948c87cde133d/integreat_cms/firebase_api/firebase_api_client.py#L51-L52

Rurn the `if` condition to `if True`

### Expected Behavior

Error is shown, user lands on the form opening the saved (but not sent) pushed notification

### Actual Behavior

Error is shown, the form is not rendered correctly

### Screenshot

<img width="1602" height="838" alt="Image" src="https://github.com/user-attachments/assets/c6c89773-7c27-44d7-abe0-a470a8ae3b11" />

### Additional Information

This is not the only case that causes this issue. Users face this whenever `send_pn` returns false.
As push notifications are indeed saved (just not yet sent), we can redirect users to `/<region_slug>/push-notifications/en/edit/<id>` instread of leaving them in `/<region_slug>/push-notifications/en/new/`

<details>
  <summary>Traceback</summary>

  ```
Oct 07 14:45:18 ERROR integreat_cms.cms.views.push_notifications.push_notification_form_view - News could not be sent due to a configuration error
Traceback (most recent call last):
  File "/home/micucu/integreat-cms/integreat_cms/cms/views/push_notifications/push_notification_form_view.py", line 484, in send_pn
    push_sender = FirebaseApiClient(pn_form.instance)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/micucu/integreat-cms/integreat_cms/firebase_api/firebase_api_client.py", line 59, in __init__
    raise ImproperlyConfigured("Push notifications are disabled")
django.core.exceptions.ImproperlyConfigured: Push notifications are disabled
Oct 07 14:45:18 ERROR integreat_cms.core.storages - Nachricht "yyyy" konnte aufgrund eines Konfigurationsproblems nicht gesendet werden.
Oct 07 14:45:18 INFO django.server - "POST /english-region/push-notifications/en/new/ HTTP/1.1" 200 7177
  ```

</details>

PR #3927, opened 2025-10-08
Handle error no translation in one region of push notification

MizukiTemma:
### Short description

This PR fixes the bug that user face the merciless internal server error when they try to send a push notification in a region which has only English in ther language tree. 

### Proposed changes

- Catch the error
- Show a message that a translation is missing

### Side effects

- Users are treated more gently

- I guess there are several ways to solve the issue. This is probably the easiest and the least invasive option.
- Users will not be redirected after catching an exception and find a half-broken form under the error message. This behaviour exists already in the system and is reported in #3926 

### Resolved issues

Fixes: #2789


integreat_cms/cms/views/push_notifications/push_notification_form_view.py
```diff
@@ -489,6 +493,18 @@ def send_pn(
...
+        logger.exception(
+            "News could not be sent due to a translation missing",
+        )
+        messages.error(
+            request,
+            _('News "{}" could not be sent due to a translation missing.').format(
+                pn_form.instance,
+            ),
+        )
+        return False
```
PeterNerlich:
```suggestion
    except ObjectDoesNotExist:
        logger.exception(
            "News could not be sent due to a missing translation",
        )
        messages.error(
            request,
            _('News "{}" could not be sent due to a missing translation.').format(
                pn_form.instance,
            ),
        )
        return False
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -11191,6 +11191,12 @@ msgstr ""
 "Nachricht \"{}\" konnte aufgrund eines Konfigurationsproblems nicht gesendet "
 "werden."
 
+#: cms/views/push_notifications/push_notification_form_view.py
+msgid "News \"{}\" could not be sent due to a translation missing."
```
PeterNerlich:
```suggestion
#: cms/views/push_notifications/push_notification_form_view.py
msgid "News \"{}\" could not be sent due to a missing translation."
```

PeterNerlich (approved):
I can confirm this works for both sending and scheduling news when the other region introducing new languages is a new change introduced with this form save. Saving as draft still always works *(as it should)*, even with configurations that are invalid for sending. I didn't explicitly test how the system reacts when sending already scheduled push notifications in the worker but a translation is missing, but I expect it to work from looking at the code – as in, an error will be added to the log. There is not much we can do otherwise anyway.

Thank you for the fix!

Issue #3904, opened 2025-09-23
Failing to create link when adding phone numbers in the middle of a line


PeterNerlich:
### Describe the Bug

When the user adds a phone number in the middle of a line as opposed to at the end, we mark it as *do not translate* but fail to create the link.

### Steps to Reproduce

1. Go to the edit form a page, event or location
2. Write a line of text *(make it `ab cd` if you want)*
3. Place your cursor inside the line *(ensure there is not a non-whitespace character before the cursor or the phone number will obviously not be matched)*
4. Type something recognized as a phone number *(e.g. `+4912345678`)*
5. Type a space or enter

### Expected Behavior

The phone number is transformed into a link around a *do not translate* `span` containing its "nice" representation. The space or newline character is placed immediately after the link. The cursor is placed immediately after this character and before any other content, with nothing selected.

### Actual Behavior

The phone number is transformed into a *do not translate* `span` containing its "nice" representation, but no link. The space or newline character is placed at the start of the next element *(or the end of the current element if there is no next element)*. The cursor is placed immediately after this character and before any other content, with nothing selected.

### Screenshot

here a screenshot in my inverted browser as usual

<img width="201" height="57" alt="Image" src="https://github.com/user-attachments/assets/1f6a5387-7c67-4d93-b875-b081f34f1ca1" />

### Additional Information

Specifically, 
```ts
editor.execCommand("createlink", false, callableNumber);
```
does not create a link object. The character and cursor placement are just a secondary consequence of this.

### Related Issues

Blocks #3905

PR #3905, opened 2025-09-23
Update tinymce and tom-select

PeterNerlich:
### Short description

Update tinyMCE *(major version upgrade)* and tom-select (no major version changes)

### Proposed changes

- Update tinymce *(our rich text editor)*
    - Change deprecated occurrences of `editor.selection.setContent()` to `editor.insertContent()` *(to be removed in 9.0)*
    - Change deprecated occurrence of `editor.fire()` to `editor.dispatch()` *(to be removed in 9.0)*
- Update tom-select *(a rich featured combo select box, currently only used in the Insert contact dialog)*

### Side effects

- Phone number normalization now exhibits #3904 even at the end of a line

### Faithfulness to issue description and design

There are no intended deviations from the issue and design.

### How to test

- Check all editor features
  - Basic text editing
  - Adding and removing simple markup *(bold, do not translate etc.)*, also in combinations that might provoke edge cases if you get such a gut feeling
  - Complex/custom behaviour / UI dialogs *(link, contact, insertion and edit/delete)*
  - Opening the source code view at various points during content editing and ensuring the representation is what you expect
  - Closing the source code view at various points during source editing and ensuring the rich text representation is what you expect
- Check used tom-select works as expected *(if you were thorough, congrats! this was already part of checking the contact insertion/editing)*
  - Does it look correctly initially and while waiting for results?
  - Does it highlight the matches correctly in the results?
  - Can you select the result you want as you expect with both keyboard and mouse?
  - Is the contact that gets rendered actually the one you selected?

### Resolved issues

Related to #3821

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #3928, opened 2025-10-09
Fix phone number linking

MizukiTemma:
### Short description

This PR fixes two problem with phone number linking:
- #3904 (linking in the middle of line)
- Linking not working also at the end of a line by #3905

### Proposed changes

- Do linking together with formatting (I gave up using `createlink` command of tinyMCE after having tackled the caret position, offset, selection etc 🙈 )
- This is blocking #3905 

### Side effects

- I hope none

### Resolved issues

Fixes: #3904 

### DIFF
--- a/integreat_cms/static/src/js/tinymce-plugins/autolink_tel/plugin.js
+++ b/integreat_cms/static/src/js/tinymce-plugins/autolink_tel/plugin.js
@@ -86,10 +86,7 @@
         }
         /* eslint-enable no-magic-numbers */
 
-        return [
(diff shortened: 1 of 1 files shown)


JoeyStk (approved):
Thank you very much! I think this works very well! :rocket:

I noticed only one minor thing: If I type "0123456789.", then hit `space` the phone number gets properly formatted, but the space gets added between -789 and the dot `[+49 (0) 15735447619](tel:+4915735447619) .`.  I thought that was a bit strange, but maybe that's no big deal?

Issue #3858, opened 2025-08-25
Use the same logic for word counting


MizukiTemma:
### Describe the Bug

There are two sources where users can see how many words (selected) contents have: in the popup which appears before bulk translation, and the table that shows number of words per content type per status. It seems they do not match each other.

### Steps to Reproduce

1. Go to Region Augsburg (in the local environment)
2. Go to the page tree, switche to English
3. Select all the 22 pages, select bulk translation and execute
4. See all the 22 pages are ready for translation* and 1211 Words to be consumed from the MT budget
5. Go to the menu "Machine Translations", find the table "Number of words in the German contents"
6. Find "1211 words" nowhere in the table

* Check all the selected pages are ready, to exclude HIX-Check to cause any difference. 

### Expected Behavior

The same calculation logic is being used both in the MT popup and the table

### Actual Behavior

Two different calculation logic are being used.

### Screenshot

<img width="543" height="667" alt="Image" src="https://github.com/user-attachments/assets/ac2df6d4-9a9f-42dc-86fb-43ff135f37dc" />

<img width="553" height="657" alt="Image" src="https://github.com/user-attachments/assets/b14edd76-9276-430d-8b31-881ea04a7376" />

<img width="1050" height="388" alt="Image" src="https://github.com/user-attachments/assets/b0bc4f6f-3859-4777-816a-04c96183a00d" />

### Additional Information

How words are counted in the table
https://github.com/digitalfabrik/integreat-cms/blob/55328cdd57f3a43c793d8ad368a75042dcb3b9ee/integreat_cms/cms/views/translations/translations_management_view.py#L66-L72

How words are counted in the popup
https://github.com/digitalfabrik/integreat-cms/blob/55328cdd57f3a43c793d8ad368a75042dcb3b9ee/integreat_cms/cms/views/utils/machine_translations.py#L63-L75

https://github.com/digitalfabrik/integreat-cms/blob/55328cdd57f3a43c793d8ad368a75042dcb3b9ee/integreat_cms/core/utils/word_count.py#L8-L24

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

### Related Issues

#1845

PR #3946, opened 2025-10-14
Fix word count

hannaseithe:
### Short description

We have been using two different methods for counting words in pages for machine translation, which results in different amounts of words to be shown. This PR applies the same method in both places. 

### Proposed changes

- in `translations_management_view.py` we changed the method to be used for word counting. Now it uses also the word_count method.

### Side effects

- hopefully none

### How to test
 

If you are testing locally: Make sure that Machine Translation is activated 
- check in the "machine translation overlay" that pops up when you select for bulk machine translation inside the page tree for all (!) pages the amount of "Benötigtes Budget" (see screenshot)
image

- and compare it with the "Summe Pro Art des Inhalts" in the "Seiten" row (see screenshot)
image

### Resolved issues

Fixes: #3858

### DIFF
--- a/integreat_cms/release_notes/current/unreleased/3858.yml
+++ b/integreat_cms/release_notes/current/unreleased/3858.yml
@@ -1,0 +1,2 @@
+en: Fix the incorrect calculation of total number of words in the translation management
+de: Behebe den Fehler, dass die Gesamtanzahl an Wörter in der Übersicht zu maschineller Übersetzung falsch berechnet wurde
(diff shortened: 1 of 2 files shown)


PeterNerlich (approved):
Confirmed the count matches for pages (published, draft, pending approval) and published locations and events.

We probably want to include a release note that the word count is now consistent throughout our system?

integreat_cms/release_notes/current/unreleased/3858.yml
```diff
@@ -0,0 +1,2 @@
+en: Fix the incorrect calculation of total number of words
+de: Behebe den Fehler, dass die Gesamtanzahl an Wörter falsch berechnet wurde
```
PeterNerlich: I think the bigger issue for people was that it was saying different numbers in different places
```suggestion
en: Fix the incorrect calculation of total number of words in the translation dashboard
de: Behebe den Fehler, dass die Gesamtanzahl an Wörter in der Übersetzungsübersicht falsch berechnet wurde
```
Maybe we have a better / more correct term than *translation dashboard*…?

Issue #3900, opened 2025-09-23
Catch matomo loading errors when matomo is down


jarlhengstmengel:
### Describe the Bug
<!-- A clear and concise description of what the bug is. -->
When loading the page accesses during the night, it sometimes happens that matomo is down for short periods of time. Currently, the script throws an error and then continues fetching page accesses, scipping some. To prevent inconsinsitenties in the page accesses data in our postgres database, we should catch when matomo is unreachable, wait for some time and then try again for the same parameters to ensure the data is fetched from matomo. See example of this in #3750 

### Steps to Reproduce


### Expected Behavior
<!-- A clear and concise description of what you expected to happen. -->
All data is fetched without any gaps from matomo

### Actual Behavior
<!-- A clear and concise description of what actually happened. -->
Sometimes there are gaps in the data, since post requests are scipped when error returns from matomo

### Screenshot
<!-- Screenshots immediately before (if applicable) and after the bug occurs -->


### Additional Information
<!-- Add any other context (e.g. logs, etc.) about the problem here. -->
Traceback is from when the similar Issue arose during our fetching of historic data

<details>
  <summary>Traceback</summary>

  ```
Fetching page accesses for augsburg for 2025-08-22                                                                                                    
---------------------------------------------------------------------------                                                                           
MatomoException                           Traceback (most recent call last)                                                                           
Cell In[5], line 12                                                                                                                                   
     10     LOGGER.info(f"Fetching page accesses for {region.slug} for {(current_date - timedelta(days=6)).date()}")                                  
     11     print(f"Fetching page accesses for {region.slug} for {(current_date - timedelta(days=6)).date()}")                                        
---> 12     fetch_page_accesses((current_date - timedelta(days=6)).date(), current_date.date(), region)                                               
     13 current_date -= timedelta(days=6)                                                                                                             
                                                                                                                                                      
  ```

</details>

### Related Issues
<!-- A collection of all issues that are related to this issue and could be useful. -->
<!-- Please also include the related integreat-app issues, if possible. -->
Part of #1097 


### Summary of discussion and updates to the description
<!-- Summarize the discussion if a long conversation was made that changed something. E.g. -->
<!-- - <TIMESTAMP>: added stack trace from cms-test -->
<!-- Leave this section blank during the initial issue creation. -->

PR #3947, opened 2025-10-15
Catch matomo loading errors

jarlhengstmengel:
### Short description

This PR catches Matomo errors when fetching page accesses from Matomo and restarts the fetching for the affected region. This ensures all data is fetched and with the integrity of the data in our database.

### Proposed changes

- Add try-except block to catch matomo exceptions and trigger fetching again after two minutes
- Edit wrong documentation

### Side effects

- I'm not sure how this could effect the currently running process for fetching historic page accesses on production, since that does the same thing (see discussion under #3900 )

### How to test
 

Not really testable in a practical way. The only really reliable test, is with a database dump.
In case you want to test with a database dump:
1. When database dump is loaded: run ./tools/run.sh
2. Then run in different console:  `./tools/integreat-cms-cli fetch_page_accesses --start-date START_DATE --end-date END_DATE` (select same date for start and end to minimize loading time)
3. Look at the messages in the first console. There should pop up error messages when there are problems reaching matomo. See that the fetching for the region where fetching failed should be restartet after 2 minutes. If everything runs without errors (yes that sometimes happens :D), repeat step 2 some time later.

### Resolved issues

Fixes: #3900 

### DIFF
--- a/docs/src/management-commands.rst
+++ b/docs/src/management-commands.rst
@@ -236,7 +236,7 @@ Update ALL link text of links of the given URL:
 
 Fetches page accesses from Matomo and store them in the database
 
-    integreat-cms-cli fetch_page_accesses --start-date START_DATE --end-date END_DATE --period PERIOD [--region-slug REGION_SLUG] [--sync SYNC]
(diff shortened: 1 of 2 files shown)


MizukiTemma (changes requested):
Thank you so much :) it looks generally good.

I have only one suggestion.

PR #3957, opened 2025-10-20
Update sass and related packages

PeterNerlich:
### Short description

Update sass and sass-loader, enabling us to write SCSS and providing the facilities to let webpack compile it down to CSS (no major version changes)
Update mini-css-extract-plugin, "extracts CSS into separate files" and "creates a CSS file for each JS file that contains CSS" (no major changes)
Update postcss-cli, postcss-import and postcss-loader, facilitate manipulating CSS with JS (no major changes)
Update postcss-preset-env, providing all necessary polyfills according to the specified range of target browsers (major version upgrade)

### Proposed changes

- Update sass, sass-loader, mini-css-extract-plugin, postcss-cli, postcss-import, postcss-loader, postcss-preset-env

### Side effects

- This should only affect webpack compilation and styles across our system

### How to test

Testing this is a mixed bag. I expect us to be able to very quickly be confident that everything is working correctly, as just starting the webserver and seeing webpack compile successfully should cover most that could go wrong. However, there remains a small possibility that some CSS rules end up slightly broken, possibly only on older browsers, making it infeasable to test completely. I propose spot checking a few views for anything that looks off, and maybe spots that you think are more rare *(e.g. the CSS involved with the menu on the left or the top bar is always shown, but there might be some that is only used on a specific page under certain conditions, like warning messages after an action)*
The rest 1% chance we will just leave to fix when we become aware of it.

### Resolved issues

Related to #3821, #3916, #3953


JoeyStk (approved):
Thank you very much! I couldn't find errors in the terminal and the styling looks as expected :) Let's merge 🚀

Issue #3945, opened 2025-10-14
internal Server Error when no start time but end time is given for event


MizukiTemma:
### Describe the Bug

User faces an internal server error when tries to create an event with no start time but with end time

### Steps to Reproduce

1. Go to the event list, click "Create event"
2. Enter title, select a date in future, leave start time emtpy, give end time (for example 15:00), select "Event does not have a physical location"
3. Click "Publish"
4. See Internal Server Error (in the test/prod system) or the yellow error wall (in a local environment)

### Expected Behavior

User gets an warning

### Actual Behavior

Error

<img width="780" height="456" alt="Image" src="https://github.com/user-attachments/assets/aba32eb6-b23e-4fc5-96bb-c8f33706d4c3" />

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  File "/home/micucu/integreat-cms/integreat_cms/cms/forms/events/event_form.py", line 173, in clean
    elif cleaned_data["start_time"] == time.min and cleaned_data[
         ~~~~~~~~~~~~^^^^^^^^^^^^^^
KeyError: 'start_time'
  ```

</details>

PR #3965, opened 2025-10-22
Fix internal server error when no event start time is given

MizukiTemma:
### Short description

This PR fixes the bug that users face the internal server error screen when start time is empty but end time is given in the event form.

### Proposed changes

- Adjust condition check

### Side effects

- should be none

(This issue was originally not scheduled in the mile stone T40K but I decided to open a PR as I found by chance the bug source 😅)

### How to test
 

Reproduce as written in the issue in the branch `develop`, test whether it is solved on this branch

### Resolved issues

Fixes: #3945

### DIFF
--- a/integreat_cms/cms/forms/events/event_form.py
+++ b/integreat_cms/cms/forms/events/event_form.py
@@ -170,9 +170,14 @@ class EventForm(CustomModelForm):
                         code="required",
                     ),
                 )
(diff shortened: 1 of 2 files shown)


JoeyStk (approved):
Thank you very much for fixing this :)

> (This issue was originally not scheduled in the mile stone T40K but I decided to open a PR as I found by chance the bug source 😅)

If you haven't already, I think we should give the PO a quick heads up about that :)

Issue #3943, opened 2025-10-13
Support POI time zones for opening hours


lunars97:
### Describe the Bug

Currently, `timeSlots` in `opening_hours` assumes that opening hours are always in the user’s local time.
This works if the user is physically at the POI, but it fails if the user views POIs in different time zones.
As a result, the system cannot reliably show if a POI is open when browsing remotely.
### Expected Behavior

Each timeslot should include the Poi’s local time zone (for ex. Europe/Berlin, Asia/Seoul).
### Actual Behavior

User's local time

### Additional Information
See app team discussion https://github.com/digitalfabrik/integreat-app/pull/3580#discussion_r2418660236

### Related Issues
App issue: https://github.com/digitalfabrik/integreat-app/issues/3619
https://github.com/digitalfabrik/integreat-app/issues/3574

steffenkleinle:
> Each timeslot should include the Poi’s local time zone (for ex. Europe/Berlin, Asia/Seoul).

I think each timeslot should use an ISO8601 time format, so e.g.
``` json
{
  "timeSlots": [{
    "start": "08:00:00+02:00",
    "end": "12:00:00+02:00"
  }]
}
```

Minor: I am actually a little surprised that its `appointmentOnly` and `timeSlots` instead of the usual snake_case `appointment_only` and `time_slots`, we could also fix this in this issue.

steffenkleinle:
I'd say not urgent, this is only a problem if a user would be in a different timezone than the poi in which case it would not be possible to go there anyway. Since this should be pretty straight forward and is perhaps a good first issue, it would be nice to fix this in the next milestone perhaps?

Magz94:
Hello everyone 
I’m a university student, and for one of our assignments we were asked to contribute to an open-source GitHub project.
I found this issue interesting and would like to work on fixing it as my first contribution.
My plan is to add a timezone field for each POI (using the IANA timezone format, e.g. Europe/Berlin) and include ISO-8601 formatted times (start_iso8601, end_iso8601) in the API response, while keeping the existing HH:mm fields for compatibility.

Please let me know if this approach sounds good , I’d be happy to start working on it right away 😊

PR #3970, opened 2025-10-23
api(v3): add ISO8601 start_time/end_time to POI timeSlots (keep legacy start/end) #3943

Magz94:
### Short description

Expose ISO-8601 offset-aware times for POI opening hours in API v3 behind an opt-in flag so legacy responses stay unchanged.

### Proposed changes
- **Add (opt-in)** `start_time` / `end_time` to each `timeSlot` as ISO-8601 strings with numeric offset (e.g., `09:00:00+02:00`), derived from the region’s timezone.  
- **Keep** legacy `start` / `end` fields as local wall time (`HH:MM` or `HH:MM:SS`) to avoid breaking existing clients.  
- **Add (opt-in)** `timezone` (IANA key, e.g., `Europe/Berlin`) to each `timeSlot` for clarity/debugging and future client logic.
- New fields are returned only when the query param  `?iso_times=1`  is provided.

Example slot:
```json
{
  "start": "09:00",
  "end": "12:30",
  "start_time": "09:00:00+02:00",
  "end_time": "12:30:00+02:00",
  "timezone": "Europe/Berlin"
}
```

### Side effects
- None expected. This change is **additive** only (no removals/renames), so current consumers using `start`/`end` continue to work unchanged.
- No database migrations. No changes to permissions or caching behavior.

Provides offset-aware times per slot (plus the region’s IANA timezone) so clients can implement “open now” logic across time zones/DST, while preserving backward compatibility by default.

### How to test
**Prereq:** run the CMS locally and ensure the `augsburg` region (or any region with POIs and opening hours) is available.

1. Default (legacy only — no ISO fields):
```bash
curl -sS "http://localhost:8000/api/v3/augsburg/de/locations/" \
| python -c 'import sys,json
d=json.load(sys.stdin)
for p in d:
    oh=p.get("opening_hours") or []
    slots=[s for day in oh for s in (day.get("timeSlots") or [])]
    if slots:
        print(p["title"], "—", p["path"])
        for s in slots[:4]:
            print(" ", s.get("start"), "→", s.get("end"),
                  "|", s.get("start_time"), "→", s.get("end_time"),
                  "|", s.get("timezone"))
        break'
```

2. Opt-in ISO fields:
```bash
curl -sS "http://localhost:8000/api/v3/augsburg/de/locations/?iso_times=1" \
| python -c 'import sys,json
d=json.load(sys.stdin)
for p in d:
    oh=p.get("opening_hours") or []
    slots=[s for day in oh for s in (day.get("timeSlots") or [])]
    if slots:
        print(p["title"], "—", p["path"])
        for s in slots[:4]:
            print(" ", s.get("start"), "→", s.get("end"),
                  "|", s.get("start_time"), "→", s.get("end_time"),
                  "|", s.get("timezone"))
        break'
```

**Expected:**  
When calling **without** the flag, each `timeSlot` shows only legacy wall times:
- `start` / `end` (wall time `HH:MM` or `HH:MM:SS`)

When calling **with**`?iso_times=1`, each `timeSlot` additionally includes:
- `start_time` / `end_time` (ISO-8601 with numeric offset)
- `timezone` (IANA key, e.g., `Europe/Berlin`)

**Validation:**
- The region’s configured timezone matches the `"timezone"` field in the response.
- `start_time` / `end_time` use the correct **current** UTC offset for that timezone (incl. DST).
- Legacy wall times `start` / `end` remain unchanged.

<details>
<summary><strong>Screenshots for Validation when calling with the iso Flag </strong> (click to expand)</summary>

**1. Situation**  

1t

**2. Situation**  

2t

**3. Situation**  

3t

</details>

### Resolved issues
Fixes: #3943

### DIFF
--- a/integreat_cms/api/v3/locations.py
+++ b/integreat_cms/api/v3/locations.py
@@ -26,6 +26,28 @@ if TYPE_CHECKING:
 
     from ...cms.models import POI, POITranslation
 
+from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
+
+
+def _ensure_zoneinfo(tz_candidate: str | ZoneInfo | None) -> ZoneInfo:
+    """
+    Convert an optional string/ZoneInfo to a valid :class:`ZoneInfo`, defaulting to settings.TIME_ZONE.
+
+    :param tz_candidate: IANA zone string, ZoneInfo, or None.
+    :return: Validated ZoneInfo.
+    """
+    default_tz = getattr(settings, "TIME_ZONE", "Europe/Berlin")
+    if isinstance(tz_candidate, ZoneInfo):
+        return tz_candidate
+    if isinstance(tz_candidate, str):
+        tz_str = tz_candidate.strip()
+        if tz_str:
+            try:
+                return ZoneInfo(tz_str)
+            except ZoneInfoNotFoundError:
+                pass
+    return ZoneInfo(default_tz)
+
 
 def transform_poi(poi: POI | None) -> dict[str, Any]:
     """
@@ -65,14 +87,19 @@ def transform_poi(poi: POI | None) -> dict[str, Any]:
     }
 
 
-def transform_poi_translation(poi_translation: POITranslation) -> dict[str, Any]:
+def transform_poi_translation(
+    poi_translation: POITranslation,
+    *,
+    region_tz: ZoneInfo,
+) -> dict[str, Any]:
     """
-    Function to create a JSON from a single poi_translation object.
+    Create JSON for a POI translation and enrich opening hours with ISO-8601 times.
 
-    :param poi_translation: The poi translation object which should be converted
-    :return: data necessary for API
+
+    :param poi_translation: POI translation to convert.
+    :param region_tz: Validated time zone used to compute DST-aware numeric offsets.
+    :return: Data for the APIv3 locations endpoint.
     """
-
     poi = poi_translation.poi
 
     contacts = Contact.objects.filter(location=poi).all()
@@ -99,11 +126,19 @@ def transform_poi_translation(poi_translation: POITranslation) -> dict[str, Any]
                 "appointment_url": contact.appointment_url or None,
             }
         )
-
     # Only return opening hours if they differ from the default value and the location is not temporarily closed
     opening_hours = None
     if not poi.temporarily_closed and poi.opening_hours != get_default_opening_hours():
-        opening_hours = poi.opening_hours
+        tz_key = getattr(region_tz, "key", str(region_tz))
+        opening_hours = [
+            day
+            | {
+                "timeSlots": [
+                    (slot | {"timezone": tz_key}) for slot in day.get("timeSlots", [])
+                ]
+            }
+            for day in (poi.opening_hours or [])
+        ]
     return {
         "id": poi_translation.id,
         "url": settings.BASE_URL + poi_translation.get_absolute_url(),
@@ -188,9 +223,10 @@ def locations(
             return JsonResponse({"error": str(e)}, status=400)
         pois = pois.filter(location_on_map=location_on_map)
 
+    region_tz = _ensure_zoneinfo(getattr(region, "timezone", None))
     for poi in pois:
         if translation := poi.get_public_translation(language_slug):
-            result.append(transform_poi_translation(translation))
+            result.append(transform_poi_translation(translation, region_tz=region_tz))
 
     return JsonResponse(
         result,
--- a/docs/src/api-docs.rst
+++ b/docs/src/api-docs.rst
@@ -382,6 +382,7 @@ RESPONSE
                   {
                      "start": String,    // The start time of the timeslot, in the format `HH:MM`, 24 Hour time
                      "end": String,      // The end time of the timeslot
+                     "timezone": String      // IANA timezone key, e.g. "Europe/Berlin"
                   },
                   ...
                ],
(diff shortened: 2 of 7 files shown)


Magz94:
Hello everyone ,
I fixed the issues that came such as Ruff,Prettier etc.. but  after adding three new fields to timeSlots (start_time, end_time, timezone in ISO-8601), the API tests fail because fixtures don’t include them.
Proposed solution: keep the default response unchanged and make these fields opt-in via 'iso_times=1'
Is this approach okay, or would you prefer a different solution 
I would be happy to hear your feedback and tell me what to do next, in order to get this merged!

integreat_cms/api/v3/locations.py
```diff
@@ -103,7 +161,32 @@ def transform_poi_translation(poi_translation: POITranslation) -> dict[str, Any]
...
+                    start = slot.get("start")
+                    end = slot.get("end")
+                    slot_copy["start_time"] = _iso_local_time(start, tz_key)
+                    slot_copy["end_time"] = _iso_local_time(end, tz_key)
+                    slot_copy["timezone"] = tz_key
+                new_slots.append(slot_copy)
+            new_day["timeSlots"] = new_slots
+            enriched_days.append(new_day)
+
+        opening_hours = enriched_days
```
PeterNerlich: If we don't want to include the iso time field, we don't need to go through these loops at all, so I would like to pull the condition out.

Additionally, we prefer [comprehensions](https://docs.python.org/3/howto/functional.html#generator-expressions-and-list-comprehensions) wherever they can be used, often times it looks cleaner than doing it imperatively. Combined with non-destructively merging dictionaries with `|` we can simplify this to a surprising degree:
```suggestion
        if not include_iso_times:
            opening_hours = poi.opening_hours
        else:
            # Enrich timeSlots with ISO-8601 times using the region's timezone (opt-in)
            default_tz: str = getattr(settings, "TIME_ZONE", "Europe/Berlin")
            tz_key: str = _tz_key(region_tz_name) or default_tz

            opening_hours = [
                day | {
                    "timeSlots": [
                        slot | {
                            "start_time": _iso_local_time(slot.get("start"), tz_key),
                            "end_time": _iso_local_time(slot.end("end"), tz_key),
                            "timezone": tz_key,
                        }
                        for slot in day.get("timeSlots", [])
                    ]
                }
                for day in poi.opening_hours or []
            ]
```
This form might be a bit confusing for anyone not already familiar with it, but you got to agree that it looks simpler and cleaner. It uses less than 45% as much non-whitespace characters as the original *(yes I know we're not playing code golf, but I just had to point this out :grin:)*

PeterNerlich: Aah, I get it. Yes, in that case we would say the intention of the code changed, so this warrants changing the tests to reflect that new intention.

integreat_cms/api/v3/locations.py
```diff
@@ -188,9 +271,28 @@ def locations(
...
+
+    # Compute once (works whether Region.timezone is a str or ZoneInfo)
+    region_tz_name = _tz_key(
+        getattr(region, "timezone", None) or getattr(region, "timezone_name", None)
+    )
     for poi in pois:
-        if translation := poi.get_public_translation(language_slug):
-            result.append(transform_poi_translation(translation))
+        translation = poi.get_public_translation(language_slug)
+        if translation:
```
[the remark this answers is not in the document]
Magz94: I'm not sure why I took walrus operator out, I think that I was just getting headaches because of those tests, and despaired on trying everything 😆 
I'm taking it back, thank you once again 😄

Magz94:
> Now, I'm sure I've pretty much bulldozed you with that huge wall of text 😅 Please do feel free to enter discussion where you can't follow why a certain thing is supposedly better or worse – even if that is a part that you decide to scrap anyways in order to implement a different approach. That's how we learn 😁 I'll do my best to stay responsive so you don't end up tinkering on this PR forever because of all the down time waiting for responses

@PeterNerlich 
Thanks for the detailed review! I’ve updated the PR to align with your points:
What changed:
* No GET flag: removed `iso_times`, we now always add the extra fields while keeping legacy fields intact.
* Backward compatibility: `start`/`end` remain unchanged; new fields are added: `start_time`, `end_time`, `timezone`.
* Validated timezone: introduced `_ensure_zoneinfo()` and now pass around a real ZoneInfo instance (not a string).
* DST-aware semantics: `start_time` and `end_time` are computed for the next occurrence of each weekday in the region TZ so the numeric offset reflects DST transitions (including “today” if it’s the matching weekday).
* Docs updated: `docs/src/api-docs.rst` now documents the three new fields and includes a note on the “next-occurrence” (DST) semantics.

Style note:
* I originally experimented with a tiny lambda/IIFE inside the list comprehension to keep things super compact 🙈 but Ruff (PLC3002) flagged it. I switched to a comprehension that binds `anchor` via a generator clause, so it’s still clean and passes linting.
This was the lambda version:
```typescript
        opening_hours = [
            (lambda a, d=day: d | {
                "timeSlots": [
                    slot | {
                        "start_time": _iso_local_time(slot.get("start"), region_tz, on_date=a),
                        "end_time": _iso_local_time(slot.get("end"), region_tz, on_date=a),
                        "timezone": tz_key,
                    }
                    for slot in d.get("timeSlots", [])
                ]
            })(
                _next_occurrence_date(region_tz, (idx + WEEK_START) % 7)
            )
            for idx, day in enumerate(poi.opening_hours or [])
        ]
```

I’m happy to tweak anything further. Also: totally open to follow-ups on whether we should eventually add the same ISO fields for contact opening hours as well, but I kept this PR scoped to the location opening hours to avoid surprises.
Thanks again for the thoughtful feedback, really helpful! 😊

Issue #3812, opened 2025-07-28
Include used library licenses in release information


svenseeberg:
### Describe the Bug
Many Open Source Licenses do require naming the used license and also the copyright holders. If we bundle these libraries, it is our obligation to show the license information. Example: https://integreat.app/licenses

### Steps to Reproduce

1. No information is available in the back end or on PyPi.

### Expected Behavior
Show the license information.

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

PR #3971, opened 2025-10-24
Update SBOM with dependencies

svenseeberg:
Result of sbom-tool w/o venv, which now also includes our Python dependency.

This is hopefully the last PR on this matter and therefore fixes the issue.

Fixes: #3884

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #3972, opened 2025-10-24
Add open source licences

JoeyStk:
### Short description

This PR creates a list of used libraries in our repository from a automatically generated SBOM file.

### Proposed changes

- Add view and template for open source 
- Generate list of used libraries from JSON

### Side effects

None I'm aware of

### How to test
 

How to get to the new view:
- Click on the release number in the bottom left corner of the CMS and then click on the link right of the version number. 

However at the moment it won't be shown a lot, because we had issues with our SBOM file. Those will be resolved once #3971 is merged :)

### Resolved issues

Fixes: #3812

### DIFF
--- a/integreat_cms/cms/templates/imprint/open_source_licenses.html
+++ b/integreat_cms/cms/templates/imprint/open_source_licenses.html
@@ -1,0 +1,44 @@
+{% extends "_base.html" %}
+{% load i18n %}
+{% block content %}
+    <h1 class="heading">
+        {% translate "Open Source Licenses" %}
+    </h1>
+    <div class="table-listing">
+        <table class="w-full mt-4 rounded border border-solid border-gray-200 shadow bg-white">
+            <thead>
+                <tr class="border-b border-solid border-gray-200">
+                    <th class="text-sm text-left uppercase py-3 pl-4 pr-2">
+                        {% translate "Packages" %}
+                    </th>
+                </tr>
+            </thead>
+            <tbody>
+                {% for package in packages %}
+                    <tr class="border-b border-solid border-gray-200 hover:bg-gray-100">
+                        <td class="px-4 py-3">
+                            <p class="font-bold">
+                                {{ package.name }}
+                            </p>
+                            <p>
+                                {% translate "Version" %}:
+                                {{ package.versionInfo }}
+                            </p>
+                            <a href="{{ package.url }}" target="_blank" class="text-blue-500">
+                                <i icon-name="external-link" class="inline-block align-center"></i>
+                                {% translate "See license" %}
+                                {{ package.license }}
+                            </a>
+                        </td>
+                    </tr>
+                {% empty %}
+                    <tr>
+                        <td colspan="1" class="px-2 py-3">
+                            {% translate "No license information available." %}
+                        </td>
+                    </tr>
+                {% endfor %}
+            </tbody>
+        </table>
+    </div>
+{% endblock content %}
--- a/integreat_cms/cms/views/imprint/open_source_licenses_view.py
+++ b/integreat_cms/cms/views/imprint/open_source_licenses_view.py
@@ -1,0 +1,51 @@
+import json
+from typing import Any
+
+from django.conf import settings
+from django.views.generic import TemplateView
+
+
+class OpenSourceLicensesView(TemplateView):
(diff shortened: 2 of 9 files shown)


integreat_cms/cms/views/imprint/open_source_licenses_view.py
```diff
@@ -0,0 +1,49 @@
...
+
+        licenses = data["packages"]
+
+        for license in licenses:
+            license["versionInfo"] = license["versionInfo"] if license["versionInfo"] else "N/A"
+            license["url"] = self.create_url_from_reference_locator(license["externalRefs"][0]["referenceLocator"])
+
+        return {
+            "licenses": licenses,
+        }
```
PeterNerlich: Technically, these are not licenses but information about packages. I think it would be easier to read if this variable was named accordingly *(e.g. in `license["versionInfo"]` I would expect to find the `2.0` in `Apache License 2.0`)*
```suggestion
        packages = data["packages"]

        for package in packages:
            package["versionInfo"] = package["versionInfo"] if package["versionInfo"] else "N/A"
            package["url"] = self.create_url_from_reference_locator(package["externalRefs"][0]["referenceLocator"])

        return {
            "packages": packages,
        }
```

PeterNerlich (changes requested):
Thank you for this PR!

I fear this still needs a rethink, as currently there exists only one SBOM file for the current version and you seem to envision displaying the license information for every past version. As is, this is very misleading, and the information for the currently running version is always shown *(the link is the same, the added API endpoint has no parameter to specify for which version to show the information and no file to use, as I already mentioned)*.

If we want to show license information for the current version only, we should only show the link for the current version. If we want it available for every past version, we need more background work to, idk, rename/copy old SBOMs similar to how we deal with release notes.

Finally, – you might be already aware of this – the SBOM manifest does not seem to mention the licenses itself, it just says `NOASSERTION` a lot. I think it would make sense to postpone this PR until that is fixed, because it makes way more sense to me to show actual license information per package instead of just linking to their store page and leaving the user to find where the license information appears there and kind of leave them to their own research.

Last comment, I notice `integreat-cms` itself not only appears in the list *(which might be considered as a bug of this PR…?)* but is also a version behind *(which is likely a separate issue of how when we generate the SBOM as part of our release pipeline)*

integreat_cms/cms/templates/imprint/open_source_licenses.html
```diff
@@ -0,0 +1,44 @@
...
+                                {% translate "Version" %}:
+                                {{ license.versionInfo }}
+                            </p>
+                            <a href="{{ license.url }}" target="_blank" class="text-blue-500">
+                                <i icon-name="external-link" class="inline-block align-center"></i>
+                                {% translate "See license" %}
+                                {{ license.license }}
+                            </a>
+                        </td>
+                    </tr>
```
PeterNerlich:
```suggestion
                {% for package in packages %}
                    <tr class="border-b border-solid border-gray-200 hover:bg-gray-100">
                        <td class="px-4 py-3">
                            <p class="font-bold">
                                {{ package.name }}
                            </p>
                            <p>
                                {% translate "Version" %}:
                                {{ package.versionInfo }}
                            </p>
                            <a href="{{ package.url }}" target="_blank" class="text-blue-500">
                                <i icon-name="external-link" class="inline-block align-center"></i>
                                {% translate "See license" %}
                                {{ package.license }}
                            </a>
                        </td>
                    </tr>
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -6972,6 +6974,22 @@ msgstr "Noch nicht gespeichert"
...
+msgid "License"
+msgstr "Lizenz"
+
+#: cms/templates/imprint/open_source_licenses.html
+msgid "See license"
+msgstr "Lizenz ansehen"
+
+#: cms/templates/imprint/open_source_licenses.html
+msgid "No licenses available."
+msgstr "Noch keine Lizenzen vorhanden."
```
PeterNerlich:
```suggestion
#: cms/templates/imprint/open_source_licenses.html
msgid "No license information available."
msgstr "Keine Lizenzinformationen vorhanden."
```

PeterNerlich (changes requested):

svenseeberg (approved):
:rocket:

PeterNerlich (approved):

Issue #2236, opened 2023-04-19
Meta: :chart_with_downwards_trend: API performance


svenseeberg:
### Motivation
The Integreat App currently has long loading times according to the Google Search console. We should improve loading times were we can by reducing API response latency where possible.

- [x] #2169 
- [x]  [Cache sites endpoint](https://git.tuerantuer.org/DF/salt/pulls/48)
- [ ] #2167

timoludwig:
Apart from the listed issues, we could also revisit all API endpoints and optimize our SQL queries and Python calculations. I'm pretty sure the views have quite some potential for improvement.

Also, is there a JIRA issue about the app-side of things that could be linked here? The changes in #2167 are very complex, so it might be easier to refactor some stuff in the app and provide simpler api endpoints that don't deliver unnecessary data.

osmers:
Closing as completed - all further API related issues will have the label topic:api

Issue #2169, opened 2023-03-27
Improve load time of children endpoint


SteffiStuffel:
### Describe the Bug
https://cms.integreat-app.de/augsburg/en/wp-json/extensions/v3/children/?depth=1
needs more than 2 seconds to load

### Steps to Reproduce

request https://cms.integreat-app.de/augsburg/en/wp-json/extensions/v3/children/?depth=1
with postman or with the browser

### Expected Behavior
should need less than 500ms

### Actual Behavior
needs 3s

timoludwig:
@SteffiStuffel Do you have other ideas for optimizations? Otherwise, I would just open a PR with this change, should be a pretty significant improvement already.

SteffiStuffel:
> @SteffiStuffel Do you have other ideas for optimizations? Otherwise, I would just open a PR with this change, should be a pretty significant improvement already.

Thanks for taking care of this. I would have checked the insights, why this request needs a lot of time and fix it depending on the result. So fine for me how you suggested.

PR #3977, opened 2025-10-27
Speedup children API endpoint

michael-markl:
_Disclaimer: I didn't read the contribution guidelines in completion since I don't plan to contribute regularly. Feel free to point out any changes I should make._

### Short description

I was a bit annoyed that the münchen instance takes so long to load, so I checked whether we can speed up the slow children API endpoint (on production the endpoint takes ~2-5s, on cms-test the endpoint takes ~8s).

This PR should speed up the children API endpoint by a factor between 3 and 50 ( :D  according to my non-scientific local measurements ) for large instances.

### Proposed changes

I tried to reduce the data queried from the DB to only those pages that are necessary to formulate the response.
This means, only a subgraph is loaded (namely, parent (if it exists), current page, and children up to the requested relative depth) and, then only a fraction of all pages is processed in Python.
For this, I made some minor adjustments to cache_tree so that it works also on such partial trees.

### Side effects

n/a 

n/a

### How to test
I let an LLM create a script that bloats the test data (see attached)
[bloat_test_data.py](https://github.com/user-attachments/files/23161075/bloat_test_data.py)
.
I then used the test data to  "benchmark"  the endpoint locally.

### Resolved issues

Related: #943 #2169 #2236

### DIFF
--- a/integreat_cms/api/v3/pages.py
+++ b/integreat_cms/api/v3/pages.py
@@ -285,7 +285,10 @@ def children(
         .prefetch_related("embedded_offers")
         .filter(Q(explicitly_archived=False) & filters)
         .cache_tree_dict(
-            archived=False, language_slug=language_slug, subtree_root_id=subtree_root_id
+            archived=False,
+            language_slug=language_slug,
+            subtree_root_id=subtree_root_id,
+            should_prefetch_nonpublic_translations=False,
         )
     )
     result = [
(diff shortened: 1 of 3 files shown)


jarlhengstmengel (approved):
Thank you very much for this PR! This looks very good to me! The failing testpipeline looks like the current flakynes of the pipeline and has probably nothing to do with your changes.
I tested with a database dump locally and had a speedup factor of 3 when the cache was empty and of 7 and upwards when things were in the cache.
I also tried to test if there are any side effects on parts of the CMS where the page tree is loaded (pages/statistics) and I couldn't find any problems.

PR #3979, opened 2025-10-27
Append LOCATION to DESCRIPTION when importing external calendar events

### LINKED ISSUE #3967: External calendar: append LOCATION to DESCRIPTION on import
### Motivation

The main value in external calenders lies in getting informed about events and where to take part (online using some special link or in some physical location). We currently have no way of representing this in our model and don't import this.

Until we have a proper solution, we need a quick solution to make do.

### Proposed Solution

- Append the LOCATION field when copying the DESCRIPTION to the event translation *(separated by one or two newlines)*

### Alternatives

- Leave it until we have a proper solution. This would almost completely diminish the value of the *external calendar* feature.

### User Story

As an ***end user*** I want ***to see events and also know how to take part*** so that ***I can take part in events relevant to me***.
As an ***municipality*** I want ***the location information to show up in Integreat*** so that ***I don't need to manually add it to the event description***.

### Related Issues

#3966

miski-h:
**What this does**

- Updates IcalEventData.from_ical_event() so that if an iCal event has a LOCATION field, that value is appended to the DESCRIPTION (separated by two newlines) before being cleaned.
- This ensures the location or meeting link appears in the event description.

**Why**

- External calendars often store participation info (address / meeting link) in LOCATION.
- This change makes that visible to users and saves municipalities from manually copying it.

**How I tested**

- Verified functionality in a Django shell:
```
print(event_obj.content)
→ <p>This is a test description<br><br>City Hall, Room 12</p>
```
- Confirming LOCATION was appended after DESCRIPTION.

**Note**

- The Django admin currently throws RelatedObjectDoesNotExist: RecurrenceRule has no event when creating events manually.
- This is an existing issue unrelated to this fix, but it prevents visual testing through the UI.
- The code change itself works correctly in the import flow.

Fixes #3967

### DIFF
--- a/integreat_cms/release_notes/current/unreleased/3967.yml
+++ b/integreat_cms/release_notes/current/unreleased/3967.yml
@@ -1,2 +1,2 @@
-en: Append LOCATION to DESCRIPTION for imported calendar events to show participation info (address or link)
-de: Füge LOCATION zu DESCRIPTION für importierte Kalendereinträge hinzu, um Teilnahmeinformationen (Adresse oder Link) anzuzeigen
+en: Append location to the description of imported calendar events
+de: Hänge den Veranstaltungsort an die Beschreibung importierter Kalendereinträge an


integreat_cms/release_notes/current/unreleased/3967.yml
```diff
@@ -0,0 +1,2 @@
+en: Append LOCATION to DESCRIPTION for imported calendar events to show participation info (address or link)
+de: Füge LOCATION zu DESCRIPTION für importierte Kalendereinträge hinzu, um Teilnahmeinformationen (Adresse oder Link) anzuzeigen
```
PeterNerlich:
```suggestion
en: Append location to the description of imported calendar events
de: Hänge den Veranstaltungsort an die Beschreibung importierter Kalendereinträge an

```

PeterNerlich (approved):

Issue #3995, opened 2025-11-03
Make countAccesses to a backend task


jarlhengstmengel:
### Motivation

When page access statistics are loaded, the pipeline is as follows:
1. Typescript requests page accesses from the backend with the `getData` method in `statistics-page-accesses.ts`
2. `getData` fetches from the backend
3. The fetch requests calls `get_page_accesses_ajax` in `statistics_actions.py`
4. `get_page_accesses_ajax` calls `get_page_accesses_by_language` in `region.py`
5. `get_page_accesses_by_language` makes the database query for the page accesses and returns them
6. Now `get_page_accesses_ajax` builds a dictionary in the format `{page_id:{language_slug:{date:accesses}}}` and returns it
7. The returned data is used to build and update the page access statistics in the frontend

The main problem here, is step 6. A lot of time is wasted building a dictionary with more information than the typescript actually would need. While experimenting it turned out to be here where the most time is spent (for Production data Augsburg 30-40 sec. for data of one month), while the database requests themselves could be handled in fractions of seconds.

### Proposed Solution

The goal of the typescript is to get charts that contain the summed accesses information for a certain timeframe per language per page. It reaches this with counting the accesses per language inside the in step 6 returned dictionary. This process of counting could be done in the backend with e.g. the tools Django QuerySet provide us (like sum). This would eliminate the need of iterating over al the selected page accesses per day to build the dictionary in step 6, dramatically reducing the complexity by the factor of requested days and probably enabling a easier implementation to build the dictionary possible.

### Related Issues

Part of #1097

PR #4000, opened 2025-11-06
Count accesses in the backend

jarlhengstmengel:
### Short description

This PR moves the `countAccesses` method into the backend as part of the query for page accesses. This allows to reduce the complexity of `get_page_accesses_ajax` and thus hugely decrease loading times for page access statistics.

### Proposed changes

- Change PageAccesses query to count accesses per page and language
- Adapt code to handle the changed format of the returned dict ({page_id: language_slug: total_accesses})
- Move the `setPageAccessesEventListeners` call to `fetch-subpages.ts` since page accesses are now loading faster then the subpages in the page tree

### Side effects

- Hopefully none

### How to test
 

I'm not sure if the effect is that noticeable with test data. Maybe this is best tested with a data dump. Then the effect is very obvious.
To test this without loading a data dump you need statistics to be activated. 
- Wait for about 30 to 40 sec
- See fast loading

### Resolved issues

Fixes: #3995 

### DIFF
--- a/integreat_cms/cms/models/regions/region.py
+++ b/integreat_cms/cms/models/regions/region.py
@@ -12,7 +12,7 @@ from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.validators import MaxValueValidator, MinValueValidator
 from django.db import models
-from django.db.models import Q, Subquery
+from django.db.models import Q, Subquery, Sum
 from django.http import Http404
 from django.template.defaultfilters import floatformat
 from django.urls import reverse
@@ -1061,7 +1061,7 @@ class Region(AbstractBaseModel):
             self.last_updated,
         )
 
-    def get_page_accesses_by_language(
+    def get_page_access_count_by_language(
         self,
         pages: list[Page],
         start_date: date,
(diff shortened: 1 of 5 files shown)


integreat_cms/cms/models/regions/region.py
```diff
@@ -1061,28 +1061,32 @@ def last_content_update(self) -> datetime:
             self.last_updated,
         )
 
-    def get_page_accesses_by_language(
+    def get_page_accesses_count_by_language(
```
PeterNerlich:
```suggestion
    def get_page_access_count_by_language(
```

integreat_cms/cms/views/statistics/statistics_actions.py
```diff
@@ -157,36 +156,21 @@ def get_page_accesses_ajax(request: HttpRequest, region_slug: str) -> JsonRespon
...
-                    access["access_date"].strftime("%Y-%m-%d"): access["accesses"]
-                }
+            page_accesses_dict[page_id][language_slug] = access_sum["total_accesses"]
         else:
-            page_accesses_dict[page_id] = {
-                language_slug: {
-                    access["access_date"].strftime("%Y-%m-%d"): access["accesses"]
-                }
-            }
+            page_accesses_dict[page_id] = {language_slug: access_sum["total_accesses"]}
```
PeterNerlich: In situations like this I like to use not an *if else* but a single *if* statement in the case that the container structure does not exist yet and then have the same line add the actual data point, being executing in any case:
```diff
     page_accesses_dict: dict[Any, Any] = {}
     for access_sum in page_access_sums:
         page_id = str(access_sum["page__id"])
         language_slug = str(access_sum["language__slug"])
-        if page_id in page_accesses_dict:
-            page_accesses_dict[page_id][language_slug] = access_sum["total_accesses"]
-        else:
-            page_accesses_dict[page_id] = {language_slug: access_sum["total_accesses"]}
+        if page_id not in page_accesses_dict:
+            page_accesses_dict[page_id] = {}
+        page_accesses_dict[page_id][language_slug] = access_sum["total_accesses"]
```

Alternatively, we could also use [`collections.defaultdict`](https://docs.python.org/3.11/library/collections.html#collections.defaultdict):
```diff
-    page_accesses_dict: dict[Any, Any] = {}
+    page_accesses_dict: dict[str, dict[str, int]] = defaultdict()
     for access_sum in page_access_sums:
         page_id = str(access_sum["page__id"])
         language_slug = str(access_sum["language__slug"])
-        if page_id in page_accesses_dict:
-            page_accesses_dict[page_id][language_slug] = access_sum["total_accesses"]
-        else:
-            page_accesses_dict[page_id] = {language_slug: access_sum["total_accesses"]}
+        page_accesses_dict[page_id][language_slug] = access_sum["total_accesses"]
```

JoeyStk (approved):
Thank you very much! LGTM! :rocket:

PR #4007, opened 2025-11-11
Log sources for Chat RAG responses

svenseeberg:
### Short description

To help editors understand why certain pages are used for generating RAG responses, we want to post a sources list as an internal note to Zammad.

### Proposed changes
- Use the reason_inclusion statement from the `extract_answer` endpoint
- Format reasons as list
- save as internal Zammad note

### Side effects
 - Longer Zammad tickets. In the future, we can make this optional by setting an additional Zammad flag. Right now it seems it is better to have this always on.

### Resolved issues

Fixes: https://github.com/digitalfabrik/integreat-chat/issues/336


svenseeberg:
Works, see https://zammad-ig-testumgebung-frag-integreat.tuerantuer.org/#ticket/zoom/24

image

Issue #2684, opened 2024-03-02
Evaluate HTMX / _hyperscript as a way to simpify our codebase


charludo:
### Motivation

Our JS/TS needs are, in most cases, pretty basic and usually fall into one of two categories:
- fetch new data from the backend to populate some element(s)
- simple reactive things (if the user clicks this button, open that modal)

We could probably significantly reduce its usage by employing [HTMX](https://htmx.org) and [_hyperscript](https://hyperscript.org).

As a concrete example, take the following file:

https://github.com/digitalfabrik/integreat-cms/blob/d2e629f294d85dae1b83ed87a19a1346c4566652/integreat_cms/static/src/js/pages/page-mirroring.ts#L1-L47

All this does is fetch an endpoint, evaluate if an error occured, and update the DOM/change some classes.

The HTMX way of doing this would be to completely scrap this file, and change the input field in `/pages/page_form_siedebar/settings_box.html` to something along the line of:

```html
<input hx-trigger="change" hx-get="/mirrored_page" hx-target="#mirrored_page_div" hx-swap="innerHTML">
```
Things like computing the correct endpoint to fetch from and handling connection errors could still be handled in TS, or in case we decide to give _hyperscript a try:

```html
<input (... as before ...) _="on htmx:sendError call (evt) => throw new Error(evt.detail.error)">
```

Why is this desirable for us?
- the obvious: far fewer lines of code
- esp. less boilerplate
- HTML generation is **only** handled in django templates, no (explicit) DOM manipulation in TS
- escaping, multilanguage support,... can be handled as usual in django/the templates instead of duplicating these structures in TS

Overall, HTMX/_hyperscript could
- move application logic back where it belongs, in python/django/templates
- remove a lot of potential error sources
- streamline development of new features

### Proposed Solution

Evaluate how big the impact of HTMX and/or _hyperscript would be on our project. This should not only include checking how much of the current code can easily be replaced, but also what the impact would be if we refactor into something more closely aligned with HTMX's philosophy.

### Alternatives

Of course, no solution is perfect, and especially in case of _hyperscript, the tech is still close to being experimental.

MizukiTemma:
We do this issue if we do not get Integreat-Berlin.

hannaseithe:
We will discuss this at the conference and @hannaseithe will do some preliminary research into options for a frontend refactoring

PR #4009, opened 2025-11-12
Refactor frontend

### LINKED ISSUE #4224: Refactor Frontend
This issue tracks the refactoring of the frontend to introduce a feature  modules pattern. TypeScript files are structured as self-contained modules, each scoped to a specific root HTML element identified by  data-js-<module-name> attributes. A bootstrapModules() function handles  dynamic discovery and initialization of these modules.

hannaseithe:
### Short description

This refactor provides a pattern for further frontend refactoring. Ts files are divided into ~~template files~~ and feature modules. Furthermore we now automatically generate a documentation for the frontend through integration of typedoc with our sphinx documentation for the backend files.

The goal of this refactor (and the subsequent refactoring of all feature modules) is/will be:
1.) To make it clear which ts file are used in which template (and vice versa)
2.) To reduce the amount of selector methods we crawl the DOM for
3.) Scope feature modules to a specific root element, to reduce the chance of accidentally attaching an eventListener to the wrong element
4.) To scope template modules to one specific template only

### Proposed changes (main ones)

- ~~a node script `generateRegistry.ts` (as npm script: `generate:registry`) that generates a file `registry.ts` which contains a record of registered feature modules and make it part of the build process~~
- a `bootstrapModules` function in index.ts ~~that dynamically imports and initializes feature modules that are need for the current page~~ that uses webpacks require.context to import feature module fiels and bootstrap them to the HMTLElement that has been marked with `data-js-<feature-module-name>`
- `search-query.ts` and `ajay-contact-form.ts` refactored as a feautre module and `data-js-search-query` attribute added on root element in template
- ~~all TS files that are only used in one template moved to the `/template` folder and set as entry-points in `webpack.config.json` and then added the respective bundles in the templates~~
- ~~delete all imports from `index.ts` that are template modules or feature modules~~
- create a npm script `docs:frontend` that uses typedoc to generate *.md documentation files for the ts files and then use the myst-parser plugin to generate *.rst files with sphinx as part of our documentation 

### Side effects

- if a template ts file is actually also used in another template (than the one I identified), it will be missed there. I have tried to be very thorough in my search, but I still might have missed one or two
- by using more entry points than before (for the template modules), the frontend will have to do one additional fetch for one more bundle

### How to test
 

- basically the whole frontend with its functionalities needs to be tested for all ~~template modules and~~ the two feature modules. Those are the files inside:
    - `integreat_cms/static/src/js/feature/`

 - make sure the validation works during generation by adding a "wrong" feature module (with wrong pattern)
- make sure the bootstrapModules function works during runtime

- generate the documentation with `tools/make_docs.sh` and see that it works

### Next steps for Refactor
- All ts files inside the `js` folder that work based on eventListeners should be turned into feature modules (that excludes exported utility functions and the files that create preact components). Follow the instructions in our documentation under `Frontend Bundling`
- use `htmx` wherever we fetch from the backend
- use `webpack-bundle-analyzer` to consider if we could get rid of big external libraries that are not essential for us or at least only partially bundle them, to reduce bundle size
- create automatic frontend tests

### Resolved issues

Fixes: #4224 
Relates to: #2684 

### Changes to PR after first round of reviews

2026-03-23:
- **Template Modules are no more**. The need to learn and maintain two patterns while feature modules could cover both, outweighed the ease of implementation template modules
- **Registry is gone**: The risk of `generateRegistry` still not being deterministic and the registry itself not being very readable in its last state, suggested to discover and add featureModules during runtime based on their location in the /feature folder and having a default function export
- **Added one more feature module: ajax-contact-fomt.ts** to test for myself if the new pattern for adding feature modules works and is easy to refactor


JoeyStk (commented):
Thank you very much for this PR! I think it's a great step forward for us in helping us keeping our codebase clean :rocket: I think you did a wonderful job :yellow_heart: 

This isn't a full review yet, but I already wanted to give you some preliminary feedback :). After running the server, I get the following error in the web console.

<details>

Uncaught (in promise) DOMException: Document.querySelectorAll: '[data-js-*]' is not a valid selector
    _callee2 index.ts:103
    i main.51245facc06e27bf6459.js:557
    o main.51245facc06e27bf6459.js:558
    __awaiter main.51245facc06e27bf6459.js:587
    __awaiter main.51245facc06e27bf6459.js:569
    bootstrapModules index.ts:100
    ts index.ts:163
    ts index.ts:159
    Webpack 5
main.51245facc06e27bf6459.js:687
    __awaiter main.51245facc06e27bf6459.js:569
    bootstrapModules index.ts:100
    ts index.ts:163
    (Async: EventListener.handleEvent)
    ts index.ts:159
    Webpack 5

</details>

JoeyStk (commented):
Still not a full review, but here is one thought of mine :)

JoeyStk (commented):
Additionally I think it might help to rebase soon, but it's probably not going to be a nice experience :grimacing:

jonbulz (commented):
My two cents:
In general, I like the idea of having a registry.ts as source of truth. But I'd strongly suggest that we decide on one of two possible paths:
- 1) We auto-generate the registry, but don't commit the file
- 2) We commit the file, but in a human-readable form. If the file is still generated by some tool or automation, we need to make sure that the generated file is identical across different environments and operating systems. Alternatively, we could edit the registry.ts by hand if a new module is added (which shouldn't happen all that often)

I personally lean towards option 2 and "edit-by-hand", unless there is a very easy way to generate the file. If we want to ensure that the registry does not contain some typo or other mishap, we can add a pre-commit hook and/or a CI job that checks that.

In general, I think you did a great job in making the frontend more structured and maintainable! Thank you very much! :)

I still didn't get around to investigate the way you implemented what you call "template modules", but the "feature modules" are a good and useful abstraction IMO

jonbulz (commented):
So, as far as I understand, we still commit the auto-generated ``registry.ts``? Is this a final decision? As I mentioned in my last comment, I'd prefer either manually editing the registry, or auto-generate but not commit. Or did I miss something? :)

tools/npm/markOrphan.ts
```diff
@@ -0,0 +1,25 @@
...
+const walkFiles = (dir: string) => {
+    for (const fileName of fs.readdirSync(dir)) {
+        const fullPath = path.join(dir, fileName);
+        if (fs.statSync(fullPath).isDirectory()) {
+            walkFiles(fullPath);
+        } else if (fullPath.endsWith(".md") && fullPath !== path.join(docsDir, mainFile)) {
+            addOrphan(fullPath);
+        }
+    }
+};
```
PeterNerlich: This is not a change of functionality. I remember seing examples of recursive functions in university and being perplexed why the professor seemed to prefer that structure – maybe you have the same reaction, I think I just somehow assimilated this, I cannot tell.

This is why I think this might be a helpful *(= readable)* alternative:
- Instead of starting with a for loop *(starting to think about the algorithm from the perspective of "I have all these files and directories I need to traverse")*, it concentrates on the case of encountering a node. This has two obvious cases which are now the first thing you see when reading the function: either it is a md file that is not the main one, or it is a directory.
- I used both `isFile()` and `isDirectory()`, which I believe is best practice, though I'm not sure whether there is more to it than how symbolic links get treated and would have to read up on the behaviour in this case. But this makes it easier to spot and change in the future, maybe?
```suggestion
const mainPath = path.join(docsDir, mainFile);
const walkFiles = (fullPath: string) => {
    if (fs.statSync(fullPath).isFile() && fullPath.endsWith(".md") && fullPath !== mainPath) {
        addOrphan(fullPath);
    } else if (fs.statSync(fullPath).isDirectory()) {
        for (const fileName of fs.readdirSync(fullPath)) {
            walkFiles(path.join(fullPath, fileName));
        }
    }
};
```
Another improvement might be to use the async versions of the file system calls to take advantage of parellization, but it runs so quickly already that this is probably a worthless avenue right now

PeterNerlich (commented):
I don't understand why we would want to differentiate between feature and template modules. Without deeply having looked into it, I feel like we could build everything as feature modules.

*What's still missing from this review is anything regarding documentation*

hannaseithe:
@jonbulz @PeterNerlich I have simplified the refactor to a great degree. Please refer to the section **Change to PR ..** in the PR description. But basically the registry is gone as well as the template modules. 

Therefore could you please re-review the PR?

integreat_cms/static/src/js/feature/ajax-contact-form.ts
```diff
@@ -0,0 +1,130 @@
...
+};
+
+export default defineFeature((root) => {
+    root.querySelector("#show-contact-form-button")?.addEventListener("click", (event) => {
+        event.preventDefault();
+        clearPreviousMessages(root);
+        renderContactForm(root);
+    });
+    if (root.querySelector("#id_area_of_responsibility")?.classList.contains("border-red-500")) {
+        root.querySelector("contact-form-widget").classList.remove("hidden");
```
jonbulz:
```suggestion
        root.querySelector("#contact-form-widget").classList.remove("hidden");
```

integreat_cms/static/src/js/feature/ajax-contact-form.ts
```diff
@@ -0,0 +1,130 @@
...
+        root.querySelector("show-contact-form-button").classList.add("hidden");
```
jonbulz:
```suggestion
        root.querySelector("#show-contact-form-button").classList.add("hidden");
```

integreat_cms/static/src/js/feature/ajax-contact-form.ts
```diff
@@ -0,0 +1,130 @@
...
+        body: formData,
+    });
+
+    const data = await response.json();
+    showMessage(root, data);
+};
+
+const renderContactForm = async (root: HTMLElement) => {
+    const showContactFormButton = root.querySelector("#show-contact-form-button");
+    const submitContactFormButton = root.querySelector("#submit-contact-form-button");
```
jonbulz: this will be `null` here because the HTML is not yet fetched. needs to move below line 109

integreat_cms/static/src/js/feature/search-query.ts
```diff
@@ -59,11 +76,11 @@ const queryObjects = async (url: string, type: string, queryString: string, arch
 let scheduledFunction: number | null = null;
```
jonbulz: can module level variables be problematic now because we could have multiple `data-js-search-query` elements in a single page now? these should maybe move to `setSearchQueryEventListeners`

integreat_cms/static/src/js/media-management/library.tsx
```diff
@@ -174,14 +174,17 @@ const Library = ({
     }, [mediaLibraryContent]);
 
     // Set the search query event listeners after a refresh
-    useEffect(setSearchQueryEventListeners, [mediaLibraryContent]);
+    useEffect(() => {
+        const container = document.querySelector("[data-js-search-query]") as HTMLElement;
+        setSearchQueryEventListeners(container);
```
jonbulz: maybe add a `null` check here?

jonbulz (changes requested):
Nice work! 😊 I can confirm that:
- bootstrapping the feature modules works
- the linter works
- the documentation is generated as expected

I'm not a big TypeScript guy, but (with the help of Claude), I found some things. I think the most significant one is to move the declaration of `submitContactFormButton`, otherwise it's not possible to create contacts from locations.

I really like that you got rid of the registry! IMO, this is really close to being finished. Just one more thought: Have you considered lazy loading for the feature modules? If so, why did you decide against it?

hannaseithe:
Thank you very much for your re-review Jonas! And I am happy that you agree that things look much cleaner without the registry.

> I really like that you got rid of the registry! IMO, this is really close to being finished. Just one more thought: Have you considered lazy loading for the feature modules? If so, why did you decide against it?

I had lazy loading (or rather dynamic importing) implemented before together with the registry: I would call  `const mod = await registry [moduleName] ()` which would call `import(...).then(...)` on the registered feature modules. Now without the registry I needed a functionality that scans a whole directory (recursively) for matching feature module files. Webpacks `require.context()` offers that, but as you correctly recognized it does not dynamically load anymore, but bundles everything together. But since our own typescript files are very small in terms of size, I felt that is something we can happily live with. What matters to me more is not using a registry anymore and making sure that only those feature modules are executed that are registered on elements in the DOM.

integreat_cms/static/src/js/feature/ajax-contact-form.ts
hannaseithe: Thank you :)

integreat_cms/static/src/js/feature/search-query.ts
hannaseithe: Good catch, thank you

integreat_cms/static/src/js/media-management/library.tsx
hannaseithe: I added a nullcheck but also use an AbortController to throw away old eventlisteners before new ones get added

integreat_cms/static/src/js/feature/search-query.ts
hannaseithe: I also changed the loop a little so that a module now gets imported only once and not  several times if the attribute (data-js-*) appears more than once

hannaseithe:
I have implemented your change requests @jonbulz .

integreat_cms/static/src/index.ts
```diff
@@ -75,10 +83,6 @@ import "./js/offers/zammad";
 
 import "./js/analytics/statistics-charts";
 import "./js/analytics/statistics-page-accesses";
-import "./js/analytics/translation_coverage";
-import "./js/analytics/hix-list";
-
-import "./js/translations/budget-graph";
```
jonbulz, line 81: can we really drop these imports? `search-query` and `ajax-contact-form` are feature modules now, but what about these?

hannaseithe: Thank you for being thorough :green_heart:  The registry revert was a very messy commit and yeah, that was missed!

jonbulz (approved):
Alrighty, I think this is good to go now 😊 Thank you very much, I think this is a great improvement!

integreat_cms/static/src/index.ts
```diff
@@ -107,12 +116,64 @@ import "./js/poi-categories/poicategory-colors-icons";
 import "./js/dashboard/broken-links";
 import "./js/dashboard/translation-coverage";
 
-import "./js/ajax-contact-form";
 // IE11: fetch
 /* eslint-disable-next-line @typescript-eslint/no-require-imports */
 require("element-closest").default(window);
 
+const initedModules = new WeakMap<HTMLElement, Set<string>>();
```
PeterNerlich: This is probably a rather dumb nitpick—there is no need to change this, but I get low key triggered that this file uses *inited* instead of *initialized*. It is even [on Wiktionary](https://en.wiktionary.org/wiki/inited) as an actual in-use phrase which I didn't expect

typedoc.json
```diff
@@ -0,0 +1,11 @@
+{
+  "entryPointStrategy": "expand",
+  "entryPoints": ["./integreat_cms/static/src"],
+  "exclude": ["**/*.test.ts"],
+  "excludePrivate": true,
+  "plugin": ["typedoc-plugin-markdown"],
+  "out": "docs/src/frontend_md",
+  "name": "Frontend Modules",
+  "categorizeByGroup": true,
+  "excludeNotDocumented": true
```
PeterNerlich: Do we really want to exclude anything that does not have a docstring? To me it sounds like we are pretending those items don't exist. But maybe I'm the one having a weird angle on this because this seems to be a popular option of typedoc and there main typedoc doesn't even consider anyone would want to document functions that are not exported.

integreat_cms/static/src/js/feature/search-query.ts
```diff
@@ -0,0 +1,213 @@
...
+        (_event) => {
+            suggestionList.classList.remove("hidden");
+        },
+        options
+    );
+
+    tableSearchInput.addEventListener(
+        "blur",
+        (_event) => {
+            suggestionList.classList.remove("hidden");
```
svenseeberg, line 195: Probably not scope of this PR, but this looks a bit odd: why are the blur and focus handler both removing the hidden class?

svenseeberg:
Is there a specific reason for using Node.js 22? AFAICT it should also work with v20, which is the major release shipped with Debian Trixie. Seems to work in https://github.com/digitalfabrik/integreat-cms/pull/4309.

svenseeberg (approved):
Looks good, I did not find any blocking issues. The front end seems to work as expected. I did not test every last feature but the core functionality seems to work without issues.

integreat_cms/static/src/js/pages/page-api-token.ts
```diff
@@ -1,5 +1,6 @@
 /*
  * Generate and (un)set the API token.
+ *
```
svenseeberg: This change seems unnecessary, but 🤷

integreat_cms/static/src/js/feature/search-query.ts
hannaseithe: The changes for that came with this [PR](https://github.com/digitalfabrik/integreat-cms/pull/3269). I would assume (without looking further into it), that blur is fired on arrow key navigation and focus on mouse navigation

hannaseithe:
> Is there a specific reason for using Node.js 22? AFAICT it should also work with v20, which is the major release shipped with Debian Trixie. Seems to work in #4309.

You are referring to my changes in the circle ci `publish-package` job?

I am guessing this was because we use node 22 overall. See also: https://github.com/digitalfabrik/integreat-cms/pull/3890

seluianova (approved):
I didn't find any issues, look good to me 👍

Issue #3064, opened 2024-09-17
POI search field in `poi_box.html` gets never disabled


MizukiTemma:
### Describe the Bug

There is an input field where a POI name can be given to search for matched POIs. This field never gets disabled even when the object is archived.

### Steps to Reproduce

1. Create an event with an event location (or add a location to an existing event)
2. Archive this event
3. Go to the archived event list and open this event
4. See the input field for POI search is not disabled

### Expected Behavior

The search input field should be disabled too like other fields when the object is archived.

### Actual Behavior

It is not disabled.

### Additional Information

This bug was found during implementation of contact form. See [this comment](https://github.com/digitalfabrik/integreat-cms/pull/3034#discussion_r1759721804).

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

MizukiTemma:
There is no search input fields for location shown anymore if the opened event had a location.

<img width="414" height="366" alt="Image" src="https://github.com/user-attachments/assets/4fe04147-af9e-48e0-85b8-da462b7e53ee" />

However, the field for online event link should be diabled.
<img width="415" height="249" alt="Image" src="https://github.com/user-attachments/assets/8945923e-97b8-4dac-b7b9-9c961120abaa" />

PR #4013, opened 2025-11-13
Disable online event link field for archived events

MizukiTemma:
### Short description

This PR disables the input field for online event link if the event is archived

### Proposed changes

- Show the URL simply, instead of rendering an input field
(In case there is no URL, nothing will be shown. This might look a bit odd but it's the same behavior with the icon box. Improvement for this aspect would be a separate issue, if we would).
image

### Side effects

- None?

### Resolved issues

Fixes: #3064 

### DIFF
--- a/integreat_cms/cms/forms/events/event_form.py
+++ b/integreat_cms/cms/forms/events/event_form.py
@@ -128,6 +128,8 @@ class EventForm(CustomModelForm):
             self.fields["external_event_id"].initial = self.instance.external_event_id
             if self.instance.start_local.date() != self.instance.end_local.date():
                 self.fields["is_long_term"].initial = True
+            if self.instance.archived:
+                self.fields["meeting_url"].widget.attrs["readonly"] = True
 
     def clean(self) -> dict[str, Any]:
         """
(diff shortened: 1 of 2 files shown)


MizukiTemma:
@JoeyStk 
Thank you for review 😸 

> I thought about maybe solving this with a disabled input field. I tried to solve this is in the `event_form` and there set the field `meeting_url` to true if it's an archived event and disabled if it's not archived.
> 
> However this line didn't solve it for me, although I thought it should. Do you have an idea how it might work? `self.fields["meeting_url"].disabled = self.instance.archived`

That approach was not so handy, if I remember correctly. Furthermore, I would like to have the appearance same/similar with the disabled POI selection field (like the screenshot)

image

PeterNerlich (approved):
Thank you for your fix!

Just a note to be aware of: If the link is longer than fits on screen, it will be cut off. The only way to find out the full link on an archived event will be to awkwardly drag from outside (!) the input element to get the browser to select its contents and paste it somewhere else to read it *(or open the dev tools to read the value directly from the HTML)*, both ways not really user friendly *(but I do still prefer this over displaying it as a link because that looks weird)*.

Issue #4001, opened 2025-11-07
Remove the template _broken_links_widget.html and linkcheck-widget.ts


hannaseithe:
### Motivation

From a brief search: It seems like the two files are not in use anymore since the commit `ccae232dadc3befc9997809239f445e7dcbb3898` ("Add a new dashboard and move HIX overview to translation coverage"  from 12th of July 2023)
They should be removed if they are not needed anymore.

MizukiTemma:
Related to #3043

PR #4014, opened 2025-11-14
Remove unused files

MizukiTemma:
### Short description

This PR removed two files that are not used anymore

### Proposed changes

- Remove `_broken_links_widget.html` and `linkcheck-widget.ts`

### Side effects

- Should be none 🙈 Please double check while reviewing 🔍 

### Resolved issues

Fixes: #4001 


JoeyStk (commented):
@hauf-toni / @osmers Just double-checking, this module is not wanted anymore and we can delete it, right? 

Screenshot from 2025-11-14 17-40-29

We had hidden it after we introduced the dashboard two years ago, because we said all of the information is now available on the dashboard. How do we think about this today? :)

JoeyStk (commented):
@MizukiTemma Thank you very much for this PR! I think this looks all well! I'm just waiting for the reply of the PO and also I have one nit: I think we might want to rename the commit a little bit. If we decide to bring this module back in two or three years, I'm not sure that it's going to be easy to find it by the current commit description. Maybe something like "Delete broken links box" would be better? Or maybe you or the second reviewer have a better idea? :)

osmers:
@JoeyStk sounds good I think - I don't see why we'd need this anymore :)

Issue #4017, opened 2025-11-17
Link and do-not-translate are mututally exclusive


osmers:
### Describe the Bug
Adding a do not translate to a link, removes the link and the other way round. 

### Steps to Reproduce

1. Go to any page
2. Choose a part of the text and add either a link or a do not translate
3. Now try adding the other one on the same section of text.
4. See the former being removed
5. Video attached for clarity.

### Expected Behavior
When adding do not translate to a link, the link remains and a do not translate is added. When adding a link to a text marked as do not translate, the do not translate remains and the link is added. 

### Actual Behavior
The first (either link or do not translate) is removed and the newly added feature is added (either link or do not translate)

### Screenshot

https://github.com/user-attachments/assets/276dc40b-090a-408d-a7d1-20f9145f09b6

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

PR #4023, opened 2025-11-19
Fix "do-not-translate" and link are mutally exclusive

MizukiTemma:
### Short description

This PR fixes the bug that `<a href>` and "do-not-translate" marking are removing the other tag each other.

### Proposed changes

- Manually set the attributes for "do-not-translate" instead of inserting

### Side effects

- Less irritated users

### Resolved issues

Fixes: #4017 

### DIFF
--- a/integreat_cms/cms/migrations/0149_alter_contact_opening_hours.py
+++ b/integreat_cms/cms/migrations/0149_alter_contact_opening_hours.py
@@ -5,7 +5,7 @@ from django.db import migrations, models
 
 class Migration(migrations.Migration):
     dependencies = [
-        ("cms", "0145_userchat_created_timestamp_and_more"),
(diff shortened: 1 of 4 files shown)


MizukiTemma:
@PeterNerlich 
Thank you for review 😸 

I hope the new approach is doing fine 🙈

PeterNerlich (approved):
Thank you so much for tackling this! I felt a bit bad when leaving my comment about the whole paragraph turning *do not translate* because I could see it being an absolute nightmare finding something that works for any selection. I'm astonished that  the solution is so compact and clean looking. Good job and I hope you stumbled across it sooner rather than later!

I did notice that if the user selects some text and across half a contact card, it breaks the contact card – but I see this as an edge case we should not attempt to fix. If a user messes up like that and does not abort their edits, it's their own fault 😅

PR #4028, opened 2025-11-20
Limit length of log entries

svenseeberg:
### Short description
We sometimes have log entries with multiple megabytes of base64 encoded images. This breaks reading logs.

### Proposed changes
Limit log messages to 4096 characters.

### Side effects
- may cut messages.

### Resolved issues

### DIFF
--- a/integreat_cms/core/logging_formatter.py
+++ b/integreat_cms/core/logging_formatter.py
@@ -71,4 +71,7 @@ class RequestFormatter(logging.Formatter):
                 else:
                     # If the string consists of one single line, just append it to the end
                     message += f"?{query}"
+        if len(message) > 4096:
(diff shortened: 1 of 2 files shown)


svenseeberg:
The failed test (generate unique slug) is unrelated to the PR.

PeterNerlich:
I'm not sure how I feel about adding this blanket solution here. I opened up a PR on linkcheck instead, this targets specifically the problematic log message, and specifically data URLs: https://github.com/DjangoAdminHackers/django-linkcheck/pull/219

I'm not sure what to expect regarding when it will be merged and land in a release however

svenseeberg:
We can increase the length to 10k chars or whatever. But truncating too long log entries is something that many programs do to prevent problems.

Issue #3831, opened 2025-08-08
Links are sometimes not tracked/recognised by the link checker


MizukiTemma:
This is a collection of issues that seem to be caused by the link checker not always successfully finding links in the contents.

After any fix is done for that promlem, each of those issues must be re-evaluated.

- [ ] #3398 
- [ ] #3706 
- [ ] #3707 
- [ ] #3708 
- [ ] #3720

juliankehne:
@PeterNerlich @MizukiTemma Initial feedback:
Contacts, urls, emails, phonenumbers seem to be recognized now as soon as something is published.

However the 'Search & Replace' is problematic:

1. At least on cms-test the 'Search & Replace' does time out / takes a long time to execute even for region with few pages / links.
2. If  'Search & Replace' does complete the link vanishes from the Broken Link overview, until the page it is in is updated again.

I made a new region to test this. Only 3 pages, 1 contact, 1 location, 1 event. Only 1 URL in the page content used on two pages. So it's not like I'm currently testing on Munich, the search & replace, if it does find a valid link seems to get stuck.

juliankehne:
> [@PeterNerlich](https://github.com/PeterNerlich) [@MizukiTemma](https://github.com/MizukiTemma) Initial feedback: Contacts, urls, emails, phonenumbers seem to be recognized now as soon as something is published.
> 
> However the 'Search & Replace' is problematic:
> 
>     1. At least on cms-test the 'Search & Replace' does time out / takes a long time to execute even for region with few pages / links.
> 
>     2. If  'Search & Replace' does complete the link vanishes from the Broken Link overview, until the page it is in is updated again.
> 
> 
> I made a new region to test this. Only 3 pages, 1 contact, 1 location, 1 event. Only 1 URL in the page content used on two pages. So it's not like I'm currently testing on Munich, the search & replace, if it does find a valid link seems to get stuck.

The issue for this: https://github.com/digitalfabrik/integreat-cms/issues/3980

PR #4030, opened 2025-11-22
Move linkcheck into celery: Switch to our fork of linkcheck

PeterNerlich:
### Short description

We will propose changes to our dependency [django-linkcheck](https://github.com/DjangoAdminHackers/django-linkcheck) that enable chosing that linkcheck jobs should run in celery.

Reviewing this PR also concerns those changes, though not necessarily as thoroughly as we do for our own well understood project. Anything that does not get caught from simply not working can be left to the upstream project. It is merely preferrable to have our PR there already be as complete as possible to help the process.

### Proposed changes

- Switch to our fork of django-linkcheck
  *We will switch this back once our changes landed upstream*

### [Proposed changes to django-linkcheck](https://github.com/DjangoAdminHackers/django-linkcheck/compare/master...digitalfabrik:django-linkcheck:celery)

- Move the core routines to update data from `listeners.py` to `worker_tasks.py`
    - rename `instance_pre_delete()` → `do_instance_pre_delete()` for consistency
- Add a setting `LINKCHECK_IN_CELERY`
- Define shared tasks in `celery.py` accepting the model and id for the instance to be checked
- When `LINKCHECK_IN_CELERY` is truthy, import `celery.py` in `listeners.py`. This means that otherwise, nothing from celery is imported, preventing celery from becoming a hard dependency of django-linkcheck
- When `LINKCHECK_IN_CELERY` is truthy, don't spawn a new thread to run any linkcheck tasks, but submit them to celery
  *Since arguments for celery tasks have to be primitive types (not python objects representing database entries) we have to juggle around a bit and find out the app and model name of the object to send along with the id, and on the other side use it to fetch the object again to hand to the function doing the actual work*

### Side effects

- Hopefully no missed links because of db deadlocks anymore
    - …as long as celery works reliably
- We shouldn't forget to change back our dependency after upstream caught up

### How to test

- ~~*integreat-django-linkcheck* is currently only on test.pypi.org – to install you need to specify it as an extra index:~~
  ```bash
  . .venv/bin/activate
  pip install -e .[dev-pinned,pinned] --extra-index-url https://test.pypi.org/simple
  ```
  ~~If you'd rather test the whole `tools/install.sh`, you can just add `--extra-index-url https://test.pypi.org/simple` to its pip command in there (around line 150).~~
  *Before we release this, we will publish our fork on the official pypi.org* :heavy_check_mark: 

### Resolved issues

Fixes: #3856, #3831

### DIFF
--- a/tools/run.sh
+++ b/tools/run.sh
@@ -48,7 +48,7 @@ listen_for_devserver &
 
 # Run Celery worker process
 if [ "$INTEGREAT_CMS_REDIS_CACHE" == "1" ]; then
-  deescalate_privileges celery -A integreat_cms.integreat_celery worker -l INFO -B --concurrency=4 -Q default,chat,statistics &
+  deescalate_privileges celery --app integreat_cms.integreat_celery worker --loglevel INFO -B --concurrency=4 --queues default,celery,chat,statistics &
 fi
 
 # Start Integreat CMS development webserver


PeterNerlich:
> A bit of an interesting review, as it is not 100% clear what exactly the scope of the review is.

Yeah, I know. I didn't want to force everyone to first need to study how exactly django-linkcheck works internally, from whoever has not dealt with it yet just reading the diff and commenting if anything weird stands out to them is enough for me.
You do already posses quite the intricate knowledge about django-linkcheck, so from you I would appreciate if you would also critically investigate my changes in that code base.

> We do need to add an additional task queue as Celery does not seem to pick up the tasks as long as `-Q default,chat,statistics` is specified.

You leave me at a loss for where that is specified and how I should search in the celery documentation to understand the behaviour and devise a plan what the best action would be *(if I understand correctly, then my gut feeling would be that we should place the task in the default queue instead of open a dedicated one, but I have not thought about queues and the dynamics of this yet)*

> When I do run the task, I get the following error […]

Oops, I forgot to pass `**kwargs` to celery through the broker. This may break for arbitrary entries though, so if someone already knows whether this just works with celery and its broker please pitch in:

```diff
     elif LINKCHECK_IN_CELERY:
         do_post_save.apply_async(
-            kwargs={
+            kwargs=kwargs | {
                 "app": sender._meta.app_label,
                 "model": sender.__name__,
                 "id": instance.id,
             },
         )
```

PeterNerlich:
I see. [The default seems to be `celery`](https://docs.celeryq.dev/en/stable/userguide/configuration.html#task-default-queue).

> ```
>   […]
>   File "/home/sven/Code/integreat-cms/.venv/lib/python3.13/site-packages/celery/app/task.py", line 566, in apply_async
>     check_arguments(*(args or ()), **(kwargs or {}))
>     ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> TypeError: do_check_link() got an unexpected keyword argument 'signal'
> ```

OH. Now this is the other task, `do_check_link()`. That *does* make sense. Pending fix should be

```diff
 @shared_task
-def do_check_link(app: str, model: str, id: int):
+def do_check_link(app: str, model: str, id: int, **kwargs):
     sender = apps.get_app_config(app).get_model(model)
```

svenseeberg:
Okay, it works at least partially :rocket: However, it still seems to break when multiple links are contained in one translation:

1 link:
```
Nov 25 10:44:57 INFO celery.worker.strategy - Task linkcheck.celery.do_check_link[777c7526-3834-464c-8076-e2fb88fa40fd] received
Nov 25 10:44:57 INFO linkcheck.models - checking external link: https://example.com/foobar1
Nov 25 10:44:58 INFO celery.app.trace - Task linkcheck.celery.do_check_link[777c7526-3834-464c-8076-e2fb88fa40fd] succeeded in 0.8399955040076748s: None
```

2 links:
```
Nov 25 10:45:10 INFO celery.worker.strategy - Task linkcheck.celery.do_post_save[e793f5d0-1af8-4ae6-881c-6a0c8d1ebd9d] received
Nov 25 10:45:10 ERROR celery.app.trace - Task linkcheck.celery.do_post_save[e793f5d0-1af8-4ae6-881c-6a0c8d1ebd9d] raised unexpected: KeyError('created')
  File "/home/sven/Code/integreat-cms/.venv/lib/python3.13/site-packages/celery/app/trace.py", line 453, in trace_task
    R = retval = fun(*args, **kwargs)
                 ~~~^^^^^^^^^^^^^^^^^
  File "/home/sven/Code/integreat-cms/.venv/lib/python3.13/site-packages/celery/app/trace.py", line 736, in __protected_call__
    return self.run(*args, **kwargs)
           ~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/sven/Code/integreat-cms/.venv/lib/python3.13/site-packages/linkcheck/celery.py", line 26, in do_post_save
    do_instance_post_save(sender, instance, linklist_cls, **kwargs)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sven/Code/integreat-cms/.venv/lib/python3.13/site-packages/linkcheck/worker_tasks.py", line 64, in do_instance_post_save
    if kwargs['created'] or (not active):
       ~~~~~~^^^^^^^^^^^
KeyError: 'created'
```

~~Seems to work with adding a kwargs in the celery.py:~~
```python
@shared_task
def do_check_link(app: str, model: str, id: int, **kwargs):
```

edit: no, there are still things not fully working as expected. I cannot see the pattern yet though. When saving translations while adding and removing links, it starts breaking at some point and then everything is borked. Sometimes I see this error in the celery log:
```
Nov 25 10:52:08 ERROR celery.app.trace - Task linkcheck.celery.do_post_save[10008789-6252-413d-a418-9bf8b6d4d02e] raised unexpected: DoesNotExist('PageTranslation matching query does not exist.')
  File "/home/sven/Code/integreat-cms/.venv/lib/python3.13/site-packages/celery/app/trace.py", line 453, in trace_task
    R = retval = fun(*args, **kwargs)
                 ~~~^^^^^^^^^^^^^^^^^
  File "/home/sven/Code/integreat-cms/.venv/lib/python3.13/site-packages/celery/app/trace.py", line 736, in __protected_call__
    return self.run(*args, **kwargs)
           ~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/sven/Code/integreat-cms/.venv/lib/python3.13/site-packages/linkcheck/celery.py", line 24, in do_post_save
    instance = sender.objects.get(id=id)
  File "/home/sven/Code/integreat-cms/.venv/lib/python3.13/site-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/sven/Code/integreat-cms/.venv/lib/python3.13/site-packages/django/db/models/query.py", line 639, in get
    raise self.model.DoesNotExist(
        "%s matching query does not exist." % self.model._meta.object_name
    )
integreat_cms.cms.models.pages.page_translation.PageTranslation.DoesNotExist: PageTranslation matching query does not exist.
```

svenseeberg:
~~I'm not exactly sure what is going on right now, but the link checks are performed in the Django application not in the Celery worker right now. I'll look into it later.~~

I had to explicitly install/update integreat-djang-linkcheck again ;)

PeterNerlich:
> @PeterNerlich do you have an idea why the tests don't like this? Looks like some kind of timeout.

Thank you for bringing this up! I had ignored it as "ah yes that failure mode, that's known" – meanwhile we are expecting this PR to eliminate it as a side effect! :see_no_evil: 

---

> *For the remainder of this comment, I am only talking about `tests/core/management/commands/test_replace_links.py::test_replace_links_commit`*

For our tests *(in `tests/conftest.py`, the `configure_celery_for_tests()` autouse fixture)* we **configure celery to synchronously execute tasks inline** and propagate any errors to the test runner, which makes the test wait for linkcheck to complete, even though the code being tested schedules a celery task with the intention of it being processed in the background. So far, this is all as it should be.

It seems like linkcheck is stuck waiting for its `update_lock`. This means that
- the **`save()`** call of the transaction was completed successfully, except for the last step of
- running all the receivers that subscribed to the models **`post_save` signal**.
- The only receiver is the linkcheck one, starting with a call to **`transaction.on_commit()`**, my last addition, that intends to ensure the celery task is only scheduled once any transaction the save action was part of is actually committed. For this management command, there is **no open transaction** *(the rationale here is probably that the command may process huge amounts of links and translations, and working through them halfway before encountering an error is not a deal breaker leaving the database in an inconsistent state, but a partial success that reduced the "problem" by half)*. So
- the **celery task is scheduled** and because of our test configuration also **executed immediately**,
- running **`do_check_instance_links()`** which as a first action
- tries to **acquire the `update_lock`**.
- **However, we wrapped the core of `replace_links()` in a block holding the `update_lock` so linkcheck does not cause chaos while we're still busy replacing links:**
  ```py
      with update_lock:
          for content, urls_to_replace in content_objects.items():
              content.replace_urls(urls_to_replace, user, commit)
  ```
  Normally the linkcheck task would wait until `replace_links()` is done and releases the lock, but since we configured celery to run inline, we also don't make any progress with our for loop. **This is a deadlock.**

My verdict: **This is a problem with our test setup only**, the code for normal operation seems good and I expect it to behave well.

PeterNerlich:
I was unable to find a way to make the tests work yet. I'm going to merge this PR so we can test in a production-like environment, and if we end up releasing this maybe open a PR on fixing the tests later

PR #4032, opened 2025-11-23
Skip page accesess with 0 value

svenseeberg:
### Short description
Do not save 0 page accesses to database. As we only use entries in sums (https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/models/regions/region.py#L1080-L1089), leaving out the rows with a value of 0 should not affect the result.

### Proposed changes
- Do not save 0 page accesses to database.

### Side effects
- We do not know exactly which data has already been imported from Matomo as 0 values are the same as non existing values. However, this should be acceptable, as we can always re-import existing data.

N/A

### How to test
Run the import

### Resolved issues

Fixes: 99% of irrelevant pageaccess entries.


jarlhengstmengel (approved):
Works 🥳 Thank you!

Issue #2639, opened 2024-02-06
Simplify (push-)News Process for Draft/Templates/Translation


dkehne:
### Motivation
The push news process became more and more complex through to 2 requirements: enabling drafts for integreat municipalities on the one hand and enabling templates for the malte-app on the other hand. both features are online and lead to confusion. 

Also we struggle with translations for push-news because of its technical architecture related to firebase. It would be more easy if a push news is translated first, then saved and the pushed to firebase.

### Proposed Solution
From @osmers and my point of view the template-requirement nearly fullfills all functions the draft-requirement has. We are thinking of the processes similar to "Überweisungsvorlagen" in daily banking activies. You first create a draft and then reuse (one or multiple times) until the draft is manually deleted. Also you could send a push notifaction and then afterwards decide if you want to send it as a kind of template (and re-use it as a draft). Newly created news can either be direct send or saved as draft/template.

The solution we propose is to rename "Vorlagen" to "Entwürfe/Vorlagen" and handle them in the same way. The current functionality to save drafts of push notifications should be dropped. From the Entwürfe/Vorlagen-View you can choose any and use it to create a push news from it. 

Things to do:
- [x] Enable the deletion of single drafts/vorlagen from the vorlagen-list #3724 
- [ ] News which are saved as draft or template shoulld show up in the draft/template list, not in the news list, as they currently do
- [ ] rename the labels from "Vorlagen" to "Entwürfe/Vorlagen"
- [ ] Drop the Entwurf-Funcitonalitiy from code

### Alternatives
Keep it as it is but we might struggle with permission-things/revision-topics and translation-issues (see blocked by issues) if we do not simpliy the process.

### User Story
Creating drafts/templates for push news should be as easy and clear like creating templates/drafts for money transfer in my bank account 

### Additional Context
Currently Labels like "Die Nachricht wurde am 04. August 2022 um 14:26 Uhr gesendet, befindet sich jetzt aber im Entwurfsstatus" are only understable for people who are insight the process. The whole problematic can be solved here as well.

JoeyStk:
Requirements aren't fully clear to me yet

nikolahoff:
@JoeyStk [here](https://www.figma.com/design/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=3542-104&t=Ug0xra9YIsUb2Kzi-1) is a draft for the flow of push notifications.

[...]

Issue #2640, opened 2024-02-06
Translation of (Push-)News with MT


dkehne:
blocked by #2639 

### Motivation
Translation should be also possible for push news/notifications. Only drafts/templates of translations should be translatable to ensure a seemless integration with firebase and external translation providers.

### Proposed Solution
News, which are saved in the template/draft-section can be auto-translated by ticking a checkbox similar to pois/events.

### Additional Context
In the Machine-translation section there should be a new checkbox to enable/disable the translation for Push News in General.

### Summary of discussion and updates to the description
Design: https://www.figma.com/design/6U7R7Xj4wL7sbjxKRmOG9D/CMS-Project?node-id=3542-104&t=xFJi4HYouPWnv7TE-0

- Add a box " Implications on translations" 
- The box should bring the same/analog effect like in the form like in page, POI, event

juliawohlhueter:
The city of Augsburg requested this feature again last week.

JoeyStk:
It's an idea for the roadmap T39K, but still can be moved. It was an idea, because it has been requested frequently

laravoelk:
Hello together, I have heard this request again twice in the last birthday calls. I was therefore wondering, if it is still the plan to work on this? Thanks a lot for a short update for us. :-)

JoeyStk:
@laravoelk Thank you for letting us know about the requests. The bugs for contact have prio 1, but it's still the plan to work on this until August :)

PR #4052, opened 2025-12-02
Simplify push notification workflow by merging templates and draft status

MizukiTemma:
### Short description

This PR introduced a new workflow for the push notification feature.

### Proposed changes

- Remove `draft`, `is_template` and `template_name` from `PushNotification` (incl. migration)
- Remove parts related to those fields from the template, form, view, Typescript file and management command
- Chanhe the workflow as in the design
- Add "Copy" option in each row of the push notification list
- Adjust API, fixture, tests

### Side effects

- Should be none 🙈 

- During the design review we all forgot that there is an option to schedule a push notification. "Send" button is called "Schedule" as before, if the opened push notification is not yet sent and has a scheduled send date.
- MT is not yet integrated into the workflow. It will be implemented as #2640 once this issue is completed.

### Resolved issues

Fixes: #2639 


integreat_cms/cms/views/push_notifications/push_notification_list_view.py
```diff
@@ -36,7 +36,7 @@ class PushNotificationListView(TemplateView):
     """
 
     #: If true, shows the template push notification list
-    templates = False
+    not_sent = False
     archived = False
     #: The template to render if templates is False
     template = "push_notifications/push_notification_list.html"
```
hannaseithe:
```suggestion
    #: The template to render if not_sent is False
    template = "push_notifications/push_notification_list.html"
```
Also in Line 41

hannaseithe (commented):
Thank you for this big PR. Everything works well and the code is clean. I just have a few thoughts about the flow of "scheduling" and of "copying" - which I could not really grasp from the design plan, but feel should be considered.

1.) I feel that the workflow with scheduling is still not really optimized. Like: if I should only be able to schedule a pn after it has been saved, I feel I shouldnt be able to select "Automatisch senden" and set the date and time during the save process. It should only show when I have the option to schedule. The other way around it is confusing that when I only click on "Speichern", but have selected a date and time, it will save that and show it in the list as "Scheduled for" that date&time. 

1a.) If we already have scheduled (actually scheduled, not just saved with a date and time) a notification, is it intended that the notification can be re-scheduled? If yes, maybe the button should show as "Re-schedule"? If no , the "schedule" button should not show

2.) Maybe when creating a copy, the form should open with that newly created copy. (Instead of just appear in the list)

MizukiTemma:
@hannaseithe 
Thank you for review 😸 I'd leave the question to @osmers @hauf-toni and @laravoelk but will provide some context.


> 1.) I feel that the workflow with scheduling is still not really optimized. Like: if I should only be able to schedule a pn after it has been saved, I feel I shouldnt be able to select "Automatisch senden" and set the date and time during the save process. It should only show when I have the option to schedule. The other way around it is confusing that when I only click on "Speichern", but have selected a date and time, it will save that and show it in the list as "Scheduled for" that date&time.

We actually didn't look deeply into the workflow with scheduling at the time of design discussion (as we all forgot that option 😅)

The background of the birth of this "saving" step before sending is to gurantee translations in other languages are surely there before user dispatchs the push notification (MT will be connected to the PN form in #2640 ). 


> 1a.) If we already have scheduled (actually scheduled, not just saved with a date and time) a notification, is it intended that the notification can be re-scheduled? If yes, maybe the button should show as "Re-schedule"? If no , the "schedule" button should not show

Re-scheduling is possible in the current implementation (before this PR) too.

> 2.) Maybe when creating a copy, the form should open with that newly created copy. (Instead of just appear in the list)

That sounds a good idea 👀

hauf-toni:
1. Thank you for your input. I think that makes even more obvious that what @hannaseithe stated is right from a users perspective. The flow should be 
New Push Notification → Edit PN → Save PN → Schedule for Sending at a later Point. This also considers our technical specification that for translation it has to be saved first and makes it scalable for when MT will be introduced at a later point. 

1a.) I agree with @hannaseithe. @MizukiTemma maybe you can clarify how exactly that would be possible? Unfortunately I can't access this functionality in the Testumgebung. 

2. Yes, agreed.

hannaseithe (approved):
Works well now. Thank you :)

osmers:
@MizukiTemma what happens to template messages that already exist in the current system? Will they be moved to the general list and marked as not sent?

MizukiTemma:
> @MizukiTemma what happens to template messages that already exist in the current system? Will they be moved to the general list and marked as not sent?

Yes, they will appear both in the lists "All News" and "Not sent".

Issue #4041, opened 2025-11-26
Selected time period won't get updated when a new period is chosen


MizukiTemma:
### Describe the Bug

Users can choose a time period in the page based statistics and it is shown on the upper right corner of "Page views" box. 

It won't be updated, however, when user chooses a new period.

### Steps to Reproduce

1. Go to a region with statistics activated
2. Go to the menu "statistics"
3. See the detault period (the last one month)
4. Scroll down and go to the box "Adjust time period"
5. Select another period and click "Customize view"
6. Go back to the box "Page views"
7. See the time period not updated

### Expected Behavior

ACCESSES (dd.mm.yyyy - dd.mm.yyyy) is updated soon after a new period is chosen

### Actual Behavior

Not updated

### Screenshot

<img width="1566" height="843" alt="Image" src="https://github.com/user-attachments/assets/5d6e883b-b4b5-401b-a32e-dbcf2ae19b46" />

### Additional Information

Probably related to #3941 

<details><summary>Traceback</summary>

```
```

</details>

Issue #4040, opened 2025-11-26
"Loading" sign does not stop rotating in the page based statistics


MizukiTemma:
### Describe the Bug

It seems the "Loading" sign in the page bases statistics does not stop/disappear in a correct timing.

### Expected Behavior

The sign stops/disappears (exactly) when all the access data is loaded

### Actual Behavior

It keeps rotating forever (?)

### Screenshot

<img width="1162" height="682" alt="Image" src="https://github.com/user-attachments/assets/2a1d8139-1778-4e8b-8fbf-139a10652222" />

### Additional Information

<details><summary>Traceback</summary>

```
```

</details>

Issue #4025, opened 2025-11-19
Uncaught TypeError page access statistics when loading bigger timeframes


jarlhengstmengel:
### Describe the Bug

When loading bigger timeframes in the statistics, an uncaught type error gets thrown and the page access statistics won't load. This bug came up after #4000 was merged. Probable cause is that the page access statistics now load faster then the graph for bigger timeframes. In `statistics-page-accesses.ts` during the execution of `createLanguageDatasetLookup`, language indexes from the graph are used to generate a lookup, with which the returned page access data is filtered for shown languages.

### Steps to Reproduce

1. Go to 'statistics'
2. Adjust time Period to a couple of months
3. See error in the console and page accesses not loading

### Expected Behavior

Everything gets loaded without errors

### Actual Behavior

Page accesses aren't loaded and uncaught type error gets thrown

### Additional Information

<details><summary>Traceback</summary>

```
```

</details>

### Related Issues

#3995 
#4000

PR #4056, opened 2025-12-04
Fix time period update

jarlhengstmengel:
### Short description

This PR fixes that the time period of the page accesses table is updated after the page accesses are loaded or isn't updated at all sometimes. This seems only to be happening when large time periods are loaded, since this is buggy at the moment.

### Proposed changes

- Move update of time period to before data is getting loaded

### Side effects

- None

### How to test

Not really testable locally, only with a database dump but that seems overkill for a oneline change.
For Julian:
- Go to statistics
- Select large time period
- See that the timeframe got updated, despite running into the loading bug (#4040 #4025)

### Resolved issues

Fixes: #4041


MizukiTemma (approved):
😁

PR #4059, opened 2025-12-09
Fix redirect after push notification send/schedule fail

hannaseithe:
### Short description
This fixes the redirect after sending or scheduling a push notification failed (after it had succesfully been created)

### Proposed changes

- If the scheduling or sending fails we now redirect to the edit version of the form, so that the fields of the new instance (that failed to send or schedule) are available inside the template

### Side effects

- None

### How to test

Make sure that the form for 
     - creating a new push notification still works 
     - the form for an existing pn still works
     - the redirect and form still works if pn was scheduled or sent
     - the redirect and from works now if pn was created but not sent (follow the steps to reproduce from the original issue)
     

### Resolved issues

Fixes: #3926

### DIFF
--- a/integreat_cms/cms/views/push_notifications/push_notification_form_view.py
+++ b/integreat_cms/cms/views/push_notifications/push_notification_form_view.py
@@ -279,9 +279,9 @@ class PushNotificationFormView(TemplateView):
                     )
                 success = False
             elif "submit_schedule" in request.POST:
-                success = send_pn(request, pn_form, schedule=True)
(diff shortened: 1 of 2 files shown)


MizukiTemma (approved):
Thank you 😸

juliankehne:
I only tested this on a local env for now since I don't know how to make a notification fail on purpose in the live environment. It did work.

Testing the push notification form in more detail will be part of the testing for this issue today & tomorrow #2639

Issue #4036, opened 2025-11-24
No access information is shown when there is no `PageAccesses` objects for the page in the selected interval


MizukiTemma:
### Describe the Bug

Sometimes no information about accesses is shown for some pages. This is confusing for users, not knowing if it is an error, or indeed some kind of result.

### Steps to Reproduce

1. Activate Statistics for Augsburg (local environment)
2. Chose an interval (for example last one month, as default)
3. Create some `PageAccesses` objects for some (not all) pages within the interval you chose
4. Go to the stistics, select all languages
5. See "x Zugriffe insgesamt" for the pages for which you created `PageAccesses`, see empty row for the pages without `PageAccesses` objects

### Expected Behavior

1. Consider no `PageAccesses` as 0 access, show "keine Zugriffe" 
2. <s>Explain there is no access history data for the page</s> See [this discussion](https://chat.tuerantuer.org/digitalfabrik/pl/jyrcrb18eibo3m9zy6be6bckee)

Option 1 is simple but does not cover the case when there is a gap in the access history (example: from some reason the access number for the page xyz on the day dd.mm.yyyy was not collected). **BUT**: the current implementation does not care whether the history is seamless anyway as long as there is at least one `PageAccesses` in the selected interval. 

<s>Option 2 reports potentiall errors and gaps in the access history but is of more effort and probably cases questions to users. </s>

I'm for option 1, personally.

### Actual Behavior

Nothing is shown.

### Screenshot

<img width="952" height="736" alt="Image" src="https://github.com/user-attachments/assets/c1ec6fd6-ede3-4749-9b9f-708c8a5e5c65" />

In this case, no `PageAccesses` object of the selected interval exists for the pages "Über die App Integreat Augsburg", Kontakt zu App Team Augsburg"). There are `PageAccesses` objects with 0 access for the pages "Wilkommen, "Behörden und Beratung".

### Additional Information

<details><summary>Traceback</summary>

```
```

</details>

### Related Issues

Blocked by: #4025

jarlhengstmengel:
#4060 is merged now. There is a quite high possibility this Issue is also resolved by #4060, since the probably underlying problem (not all page tree nodes get iterated over) was resolved sonce #4060 introduces the iteration over the shown page tree nodes instead of only the ones whose ids are in the returned data. This is should be testable when #4060 is on the test system.

hannaseithe:
[...]

PR #4060, opened 2025-12-09
Fix uncaught type error when loading bigger timeframes

jarlhengstmengel:
### Short description

Fixes the the typescript error when loading bigger timeframes for page access statistics. Also includes small refactor to only request data for languages that are actually selected from the backend. And finally it get get rid of the dependency of the page acces statistics from the total accesses chart. They are now completly independent of each other. Both only depend on the language legend.

### Proposed changes

- Refactor request to backend to accomandate a list of selected languages
- Changes to backend to be able to querry with list of language slugs and only non archived pages
- Change `updateDOM` to looping through page tree instead of returned data to ensure all nodes are updated right
- Get list of selected language slugs from language legend
- Get rid of every mention of chart in `statistics-page-accesses.ts`

### Side effects

- Should be none

### How to test

To reproduce the original error and see if its fixed you actually need a database dump. But you can test if everything works as it should without a database dump:
To test everything works fine without loading a data dump you need statistics to be activated. 
- Wait for about 30 to 40 sec
- See that everything works fine and no uncaught type error gets thrown for page accesses

For Julian:
Uncaught Type Error for pages accesses should be gone now when selecting bigger timeframes like couple of months. Loading animation should also stop.

### Resolved issues

Fixes: #4025 
Fixes: #4036 
Fixes: #4040 

### DIFF
--- a/integreat_cms/cms/models/regions/region.py
+++ b/integreat_cms/cms/models/regions/region.py
@@ -1066,14 +1066,14 @@ class Region(AbstractBaseModel):
         pages: list[Page],
         start_date: date,
         end_date: date,
-        languages: list[Language],
+        language_slugs: list[str],
     ) -> dict:
         """
         Get the sum of page accesses per page and language of this region during the specified time range
         :param pages: List of requested pages
         :param start_date: Earliest date
         :param end_date: Latest date
-        :param languages: List of requested languages
+        :param language_slugs: List of slugs for the requested languages
 
         :return: Sum of page accesses per page and language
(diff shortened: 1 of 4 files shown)


hannaseithe (approved):
Thank you very much for this elegant PR - I am coming to actually like the `statistics-page-accesses.ts` file now. :laughing:  Decoupling from chart.js is such a smart move.

I included one suggestion for further improvement, but I leave it up to you to implement or not, as its not essential

Issue #4050, opened 2025-12-01
No primary contact per "Related contacts" box, if "the box Contact details" is empty


MizukiTemma:
### Describe the Bug

Currently users cannot create a primary contact per "Related contacts" box, if "the box Contact details" is empty.

### Steps to Reproduce

1. Go to "Locations", click "Create location"
2. Fill the title, address and category
3. Keep the box "Contact details" empty
4. Click on "Create a new contact" within the box "Related contacts"
5. Fill some field but keep "Area of responsibility" empty
6. Save the location
7. See the contact is not created

 
### Expected Behavior

Input given in the box "Related contacts" is saved as contact

### Actual Behavior

No contact created if "Area of responsibility" is empty.

### Additional Information

This bug was found here: https://github.com/digitalfabrik/integreat-cms/pull/4039#pullrequestreview-3524494959
<details><summary>Traceback</summary>

```
```

</details>

MizukiTemma:
The root of problem and confusion (which even I as developer often experience) is the concurrency between two boxes "Contact details" (blue box) and "Related contact" (red box) as both can create an primary contact.

<img width="271" height="819" alt="Image" src="https://github.com/user-attachments/assets/2ca7e01c-f987-457d-a903-1402feabf294" />

In the reported case (this issue), a primary contact created by the red box is overwritten (deleted) by the blue box, which says with emptry fields "no primary contact". The cause is that the red box is processed before the blue box.

We can probably then process blue box first then the red box. This looks first good, but brings another confusion: when user opens an existing POI, create and save a primary contact in the red box, then save the POI with the blue box empty, the primary contact which user created in the red box, disappers, because the blue box is empty and processed later. This is in my opinion a correct behaviour (thinking the blue box is superior), but is definetely confusing. 

Another problematic scenario is, when both the blue box and red box receive inputs that become a primary contact but differ each other (the blue box is not empty, and in the red box "area of responsibility" is empty but one or more of other fields are filled). We do not know which should win. We can define the blue box as winner, but this makes users definetely confused.


So, long story in short word: the red box should not create any primary contact.

@osmers @hauf-toni @laravoelk @digitalfabrik/cms-team 
I want to hear your opinion 💡 👀

laravoelk:
I have read this issue twice and I am really unsure about what could be the best option, I am sorry. I would really like to wait and hear Tonis opinion.

hauf-toni:
[...]

PR #4063, opened 2025-12-10
Remove (0) from phone number formatting 

hannaseithe:
### Short description

This PR changes the behaviour of phone number formatting and adds form validation for the phone number and mobile number field of the contact.
Note: This PR is an extension of #3741 . I had to open a new PR since reopening the old one was not possible

### Proposed changes

- Remove ` (0) ` from the formatting and set the saving format as e.g. "+49 123456789" (`+<country-code> <phone-number>`)
- Add a migration to adopt the (mobile) phone numbers of the existing contacts into the new format
- Add validation for the contact form
- Adjust the test

- I also added a little change to launch.json to enable debugging into the django files

### Side effects

- None

### How to Test

Test in
1. the contact form
2. the poi form
     - the primary contact box
     - the related contact box 
3. the autolink feature of the page (type a valid phone number + ENTER -> and see that it is correctly formated and linked)
  
- Try entering invalid phone numbers like "12345678", "abcdefg", "+49 (0) 123456789", ... make sure they are marked as invalid
- Enter valide phone numbers and see that they are formatted correctly: "+49 123456789", "0049 123456789", "0123456789", "+49123456789", "0049123456789" -> should all be formated to "+49 123456789"

### Resolved issues

Fixes: #3712 


integreat_cms/cms/forms/pois/poi_form.py
```diff
@@ -39,7 +45,8 @@ class POIForm(CustomModelForm):
         required=False,
     )
     primary_phone_number = forms.CharField(
-        max_length=250,
+        validators=[phone_validator],
+        max_length=40,
```
jarlhengstmengel: What is the reason to limit this to 40 instead of 250? Intuitively it sounds reasonable but I couldn't find anything in the discussion of the original Issue

jarlhengstmengel (approved):
Thank you! Couldn't find any non german numbers that weren't migratet anymore :)

hannaseithe:
> Maybe on top of https://github.com/digitalfabrik/integreat-cms/issues/4050 its also another bug since I found that even if contact information is invalid it creates the location, but will not create the contact.

This is actually a bug that must have been there before I believe, but we did not catch it, because there was no easy way to invalidate the contact. Right now I am not entirely sure, what the expected behavior is:

During poi creation (while directly also creating a related contact), if the poi is valid but the related contact is not, what should happen?

1.) Nothing is saved, we get an error message that the phone field needs to be valid and only after everything is valid poi and contact are created at once

2.) The poi is saved, but the contact is not, we get an error for contact and on the poi form we still show the related contact form with the invalid data

3.) ?

@hauf-toni  @osmers

hauf-toni:
> > Maybe on top of #4050 its also another bug since I found that even if contact information is invalid it creates the location, but will not create the contact.
> 
> This is actually a bug that must have been there before I believe, but we did not catch it, because there was no easy way to invalidate the contact. Right now I am not entirely sure, what the expected behavior is:
> 
> During poi creation (while directly also creating a related contact), if the poi is valid but the related contact is not, what should happen?
> 
> 1.) Nothing is saved, we get an error message that the phone field needs to be valid and only after everything is valid poi and contact are created at once
> 
> 2.) The poi is saved, but the contact is not, we get an error for contact and on the poi form we still show the related contact form with the invalid data
> 
> 3.) ?
> 
> @hauf-toni @osmers

From a ui/ux perspective, I would recommend that nothing is saved. we get an error message that the phone field needs to be valid + only after everything is valid poi and contact are created at once. this provides strong data consistency + (more importantly) a simple mental model: “everything worked/nothing worked". would be my favourite solution.

hauf-toni:
> > From a ui/ux perspective, I would recommend that nothing is saved. we get an error message that the phone field needs to be valid + only after everything is valid poi and contact are created at once. this provides strong data consistency + (more importantly) a simple mental model: “everything worked/nothing worked". would be my favourite solution.
> 
> I can get on board if that does not mean that everything is erased. It's important that the users simply get a pinpoint of what to fix and then can fix that and save again without having to enter all the other data entered previously.

definitely. the version you described follows ui/ux best practice. the error should be pointed out but the data should remain the same until edited by the user

PR #4045, opened 2025-11-28
Refactor page form view

JoeyStk:
### Short description

This PR refactors the `post` method of `page_form_view` a little bit by adding some functionality to sub methods.   

### Proposed changes

- Move some functionality of the post method into sub methods

### Side effects

- We decided to keep the conditional statement in the main method, because we think it's better to read this way. However with this in mind, we were only able to reduce the cyclomatic complexity marginally from 28 to 26.

### Faithfulness to issue description and design

There are no intended deviations from the issue and design.

### Resolved issues

Fixes: None
Hopefully this helps us a little bit with #3837

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #4067, opened 2025-12-13
Refactor `get` in `page_form_view`

JoeyStk:
### Short description

After refactoring the `post` method of the the `page_form_view`, I decided to also refactor the `get` method.
Similarly to the [previous PR](https://github.com/digitalfabrik/integreat-cms/pull/4045), I also used the [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) as a metric to track progress. This time I was able to reduce the CC by half (from 18 (=C) to 9 (=B)).

### Proposed changes

- Refactor `get` method in `page_form_view`

### Side effects

- I'm not entirely happy with the method yet, but think it's a good step forward. Let me know if there is still something left in the method that you think should be refactored
- I'm not 100% sure if I really caught every test case, so thorough testing is very welcome

### Resolved issues

Fixes: /

### DIFF
--- a/integreat_cms/cms/views/pages/page_form_view.py
+++ b/integreat_cms/cms/views/pages/page_form_view.py
@@ -80,7 +80,6 @@ class PageFormView(
             only_active=True,
         )
 
-        # Get page and translation objects if they exist
         page = (
             region.pages.filter(id=kwargs.get("page_id"))
             .prefetch_translations()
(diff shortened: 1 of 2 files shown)


integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -89,103 +88,27 @@ def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
...
-                        request,
-                        translate_link(
-                            message,
-                            attributes={
-                                "href": revision_url,
-                                "class": "underline hover:no-underline",
-                            },
-                        ),
-                    )
+        disabled = self.get_initial_disable_state(request, page)
```
jonbulz: The value that is returned here is never used. If this is just to handle warning messages if the page is archived, I'd prefer renaming the method to something more fitting

integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -89,103 +88,27 @@ def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
...
                 request,
-                _(
-                    "You don't have the permission to publish this page, but you can propose changes and submit them for review instead.",
-                ),
+                page=page,
+                page_translation=page_translation,
+                language=language,
             )
 
+        disabled = self.handle_permissions(request, page)
```
jonbulz: So ``handle_permissions`` returns ``True`` if the user does not have permissions to edit the page? I find that a bit confusing. 
I have two suggestions:
1) Split it such that ``disabled = not request.user.has_perm(...)`` is in a different line and ``handle_permissions`` only manages the messages or
2) Rename to ``check_is_disabled`` or something like that

jonbulz (approved):
Thank you for your work, I think this makes things a lot cleaner! I approve because I agree with the concepts and separation of logic. However, I have a few comments regarding the naming. But naming is hard and always subjective, so I wouldn't say that my suggestions are better. I just wanted to give some food for thought :)

JoeyStk:
@jonbulz do you think you could have another look at this? :)

Issue #3941, opened 2025-10-13
Multiple requests for page based statistics loading at the same time


juliankehne:
### Describe the Bug

When opening the statistics page the page based statistics start loading for the default time range of one month.

This request for data is not stopped when another request is sent during that time by adjusting the time range and clicking 'customize view'.

If the manual request is e.g. just a few days and loads faster than the initial data of one month it will be overwritten by it.

### Steps to Reproduce

1. Go to 'Statistics'
2. Scroll down to see Page Based statistics started loading
3. While it is loading change the time range to only a few days (e.g. today & yesterday) instead of the default range of the last month (It has to load faster than the data for last month)
4. Click 'Customize view'
5. Wait

### Expected Behavior

Only the data is loading for the days that were manually requested and is not overwritten by statistics that were requested automatically before that.

### Actual Behavior

After a short while the data for the short time range is displayed. This is overwritten some time afterwards by the data for the initial request that is sent automatically when the statistics page is opened.

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

### Related Issues

#1097

jarlhengstmengel:
I have the feeling this isn't really an Issue anymore since the loading times are now so fast, I can't click that fast to trigger this behavior. @MizukiTemma what would you think? Should we close this? Or fix this to be sure?

PeterNerlich:
> It's possible to abort POST requests in typescript, maybe that's an option to fix this

I know that in search suggestions we [take care to discard responses to requests that have been superseded](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/static/src/js/tinymce-plugins/custom_link_input/plugin.js#L165-L169)

jarlhengstmengel:
> I know that in search suggestions we [take care to discard responses to requests that have been superseded](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/static/src/js/tinymce-plugins/custom_link_input/plugin.js#L165-L169)

Ah that solution looks easy enough :) thank you for the hint!

PR #4072, opened 2025-12-16
Handle multiple requests pge accesses at the same time

jarlhengstmengel:
### Short description

Add handling of multiple requests for page access statistics.

### Proposed changes

- Add `ajaxRequestID` to track if there was another request after the current one

### Side effects

- None

### How to test

The main thing to test here, is if everything works as expected. The original bug is really hard to reproduce after the speedup of the requests. But since the requests are async, it's probably not to bad to have this watertight.

To test everything works fine without loading a data dump you need statistics to be activated. 
- Wait for about 30 to 40 sec
- See that everything works fine

Bugs with the loading indication and uncaught type errors are handeld by #4060 

### Resolved issues

Fixes: #3941 


MizukiTemma (approved):
Works as expected 😸 Thank you!

Issue #3609, opened 2025-04-21
Fix width of statistics box


JoeyStk:
### Describe the Bug

Since merging the page based statistics to cms test the already existing "total accesses" box has a limited width.

### Steps to Reproduce

1. Go to statistics either on cms-test or locally
2. See "total accesss" doesn't fill the entire available width

### Expected Behavior

Fill entire width, like on admin

![Image](https://github.com/user-attachments/assets/218e9cbd-2de1-45f2-b415-b6aed4340edf)

### Actual Behavior

Limited width on cms-test and locally

![Image](https://github.com/user-attachments/assets/0472cce3-d349-48a9-bbf4-d6358e4c0b35)

### Screenshot

None

### Additional Information

This was probably introduced by #3325 

### Related Issues

None

MizukiTemma:
> [@MizukiTemma](https://github.com/MizukiTemma) is this something that is ready to be worked on?

I'll forward the question to the assignee @jarlhengstmengel

PR #4075, opened 2025-12-18
Fix width of statistics box

MizukiTemma:
### Short description

This PR fixes the bug that the graph of "Total Accesses" does not use the full width of the box.

### Proposed changes

- Remove `w-[52vw]`
- Control by the (outer) box

### Side effects

- I hope no negative side effects

- Two more unpleasant behaviours are fixed together:

"Total accesses" and "Page views" extends over the boundary in a small screen/when the screen is made narrower.
Screenshot 2025-12-18 165009

The left and right columns are overwrapping when changing the size of screen 

Screenshot 2025-12-18 180514

### Resolved issues

Fixes: #3609 

### DIFF
--- a/integreat_cms/cms/templates/_collapsible_box.html
+++ b/integreat_cms/cms/templates/_collapsible_box.html
@@ -1,5 +1,5 @@
 {# Set collapsed=True when including this template to show the box collapsed in the beginning #}
-<div class="rounded {% if 2_cols %} 2xl:col-span-2 {% endif %} shadow-2xl border border-blue-500 bg-white mb-4">
+<div class="max-w-full rounded {% if 2_cols %} 2xl:col-span-2 {% endif %} shadow-2xl border border-blue-500 bg-white mb-4">
     <div {% if box_id %}id="{{ box_id }}"{% endif %}
          class="collapsible cursor-pointer rounded p-4 bg-water-500 text-black font-bold flex justify-between"
          {% if collapsed %}data-collapsed{% endif %}>
--- a/integreat_cms/cms/templates/statistics/statistics_chart.html
+++ b/integreat_cms/cms/templates/statistics/statistics_chart.html
@@ -9,7 +9,7 @@
     {% translate "Total accesses" %}
 {% endblock collapsible_box_title %}
 {% block collapsible_box_content %}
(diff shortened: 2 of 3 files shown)


dkehne:
## Review Summary

### What it does
- Removes hardcoded `w-[52vw]` from chart container
- Adds `max-w-full` to collapsible boxes
- Adjusts responsive grid classes

### Analysis of the current fix

The current approach is on the right track (removing `w-[52vw]` is correct), but it won't fully solve the 4K issue because:

1. **Missing `w-full` on chart container** - Removing `w-[52vw]` without replacing it with `w-full` leaves the chart without any width constraint. The canvas will rely on its intrinsic size instead of filling the container.

2. **Missing `min-w-0` on flex container** - This is the key fix for 4K. By default, flex/grid items have `min-width: auto` which prevents them from shrinking below their content size. Without `min-w-0`, the grid column cannot shrink even with `minmax(0px, 1fr)`.

3. **The `3xl:block 4xl:flex` changes are confusing** - Mixing display modes (`block`/`flex`) with grid `col-span` classes can cause unexpected behavior and doesn't address the root overflow issue.

### Suggested fix

**1. `statistics_chart.html`** - Use container-relative width:
```html
<!-- Change from: -->
<div class="chart-container w-[52vw] h-[60vh]">

<!-- To: -->
<div class="chart-container w-full h-[60vh]">
```

**2. `statistics_overview.html`** - Add `min-w-0` to allow shrinking:
```html
<!-- Change from: -->
<div class="flex flex-col flex-wrap col-span-1 3xl:col-span-1">

<!-- To: -->
<div class="flex flex-col flex-wrap col-span-1 3xl:col-span-1 min-w-0">
```

**3. `_collapsible_box.html`** - The `max-w-full` addition is fine, keep it.

### Why this works
- `w-full` makes the chart 100% of its **container** width (not viewport)
- `min-w-0` overrides the default `min-width: auto` and allows the grid column to actually respect `minmax(0px, 1fr)`
- Simpler than mixing `block`/`flex` display modes with grid classes

hannaseithe (approved):
First of all: I am not an expert on CSS, so I cannot judge the code from that perspective, but I have resized and clicked around quite a bit (also checked other places where we use collapsible_box), and it looks good to me. 

Thank you very much!

PR #4076, opened 2025-12-21
Refactor `generate_unique_slug`

JoeyStk:
### Short description

This PR attempts to refactor the method `generate_unique_slug` in `utils/slug_utils.py`. We came across this method during our work on [the Django update](https://github.com/digitalfabrik/integreat-cms/pull/3837). There were two main issues I saw with the method: firstly I thought the [CC](https://en.wikipedia.org/wiki/Cyclomatic_complexity) was quite high (17) and secondly I thought the function was a bit too long (97 LOC). I tried to reduce both. 

--- Info ---
This PR was handed over to @MizukiTemma 

### Proposed changes

- Reduce CC to 5
- Reduce LOC to 68

### Side effects

- As with all refactorings, I tried my best to pay attention to all the details, but of course there is a high chance of human error, so horough testing is highly appreciated.

### How to test

- Check if a new slug is properly generated
- Check in case a slug already exists, the new slug is changed to SLUG-2
- Check in case a slug is only used in a previous version of this page, that no error occurs

### Resolved issues

Fixes: #4149 (to be confirmed)

### DIFF
--- a/integreat_cms/cms/utils/slug_utils.py
+++ b/integreat_cms/cms/utils/slug_utils.py
@@ -94,7 +94,7 @@ def generate_unique_slug_helper(
         "foreign_model": foreign_model,
         "fallback": "name",
     }
-    if foreign_model in ["page", "event", "poi"]:
+    if foreign_model in ALLOWED_OBJECTS:
         kwargs.update(
             {
                 "region": form_object.instance.foreign_object.region,
@@ -125,7 +125,6 @@ def generate_unique_slug(**kwargs: Unpack[SlugKwargs]) -> str:
 
     :return: An unique slug identifier
     """
-    content_models: list[str] = ["page", "event", "poi"]
     slug: str = kwargs.get("slug", "")
     foreign_model: str | None = kwargs.get("foreign_model")
     foreign_object: AbstractContentModel | None = kwargs.get("foreign_object")
@@ -136,13 +135,15 @@ def generate_unique_slug(**kwargs: Unpack[SlugKwargs]) -> str:
     language: Language | None = kwargs.get("language")
     manager: Any = kwargs["manager"]
 
-    handle_logging(content_models, foreign_model, region, language, slug)
+    logger.debug("foreign_model: %r", foreign_model)
+    if foreign_model in ALLOWED_OBJECTS:
+        logger.debug("%r, %r", region, language)
+    logger.debug("slug: %r", slug)
 
     base_slug = generate_base_slug(
         slug, fallback, cleaned_data, object_instance, foreign_model
     )
     pre_filtered_objects = get_prefiltered_queryset(
-        content_models,
         manager,
         foreign_model,
         region,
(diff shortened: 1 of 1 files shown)


integreat_cms/cms/utils/slug_utils.py
```diff
@@ -116,69 +120,139 @@ def generate_unique_slug(**kwargs: Unpack[SlugKwargs]) -> str:
...
+    foreign_object: AbstractContentModel | None,
+    content_models: list[str],
+) -> QuerySet[T]:
+    """
+    Excludes the current object from the given queryset to avoid false positives when checking for slug uniqueness
+    """
+    if not object_instance:
+        return qs
+
+    if foreign_model in content_models:
```
PeterNerlich: Aah, I remember why that had to be in here, because we cannot just pass objects without an ID with the new django version anymore.
In that case, the foreign object (page, event or location) does not exist in the db yet, meaning that we are trying to create it. When it doesn't exist yet, we also don't have any other slugs in the database to exclude when looking for naming conflicts, wo skipping that step and running the lines following is valid.
```suggestion
    if (
        foreign_model in content_models
        and (foreign_object or object_instance.foreign_object).id
    ):
```

dkehne:
## Review Summary

### What it does
- Refactors `generate_unique_slug` by extracting helper functions:
  - `handle_logging()` - handles debug logging
  - `generate_base_slug()` - generates initial slug from fallback
  - `get_prefiltered_queryset()` - filters queryset by region/language
  - `is_reserved_slug()` - checks reserved slugs
  - `exclude_current_object()` - excludes current object from uniqueness check
- Reduces cyclomatic complexity (17 → 5) and LOC (97 → 68)

### Key observations

1. **Good extraction of helper functions** - Each function has a single responsibility and is well-named

2. **Improved readability** - The main function is now much easier to follow

3. **Existing review comment from PeterNerlich** - Points out a potential bug in `exclude_current_object`:
   > When the foreign object doesn't exist in the DB yet (no ID), we shouldn't try to exclude it. 
   
   Suggested fix:
   ```python
   if (
       foreign_model in content_models
       and (foreign_object or object_instance.foreign_object).id
   ):
   ```
   This needs to be addressed before merging.

4. **Minor: `exclude_current_object` has defensive check** - The `if not object_instance: return qs` check is redundant since the caller already checks `if object_instance:`, but it's harmless defensive coding

5. **TypeVar `T` usage is good** - Preserves QuerySet type through the `exclude_current_object` function

**Blocking issue:** PeterNerlich's feedback about the missing ID check needs to be addressed.

MizukiTemma:
> > Is this actually comming in your new [branch](https://github.com/digitalfabrik/integreat-cms/compare/develop...bugfix/test_duplicate_pages)?
> 
> Uh, yeah I just re-read my own comment, and it seems I did the exact same investigation with the same conclusion, again. Sorry that I didn't take better notice
> 
> I haven't caught up on all the context here again, but maybe while #4149 is a good quick fix, this is the proper solution. I'll leave that to the reviewers – if this gets greenlit first and does fix #4076, maybe I'll just close #4149

@PeterNerlich 
Thank you for reply. Then let's the team review this PR first.

jonbulz:
Looks good, thanks for applying my suggestions! I already approved, so take my thumbs up instead 👍

PR #4077, opened 2025-12-21
Refactor form views

### LINKED ISSUE #4208: Refactor `post` in `integreat_cms/cms/views/pois/poi_form_view.py`
### Background and current state

The post function in `integreat_cms/cms/views/pois/poi_form_view.py` has received an warning "too-many-statements (PLR0915)" by ruff. It's currently suppressed (to separate refactoring from #4136 ).

### ToDo List

- [ ] Remove suppression
- [ ] Refactor `post` to reduce statements

### Additional Context

https://github.com/digitalfabrik/integreat-cms/pull/4136#issuecomment-4098354552

Wait for #3837

### LINKED ISSUE #4264: "The slug was changed from...to..." is not shown in the event/page form
### Bug Summary

The message "The slug was changed from...to..." must be shown to user if they selected a slug manually and it was changed at saving due to conflict to other translation (uniqueness required). This message is however not shown currently at all.

### Expected Behavior

If user select a slug manually by entering it into "Event link" or "Page link" and it is changed due to uniqueness requirement, the mssage "The slug was changed from .. to..." is shown.

### Actual Behavior

If user select a slug manually by entering it into "Event link" or "Page link" and it is changed due to uniqueness requirement, the mssage "The slug was changed from .. to..." is NOT shown.

### Steps to Reproduce

1. Create an event/page with slug "xxxx"
2. Create another event/page, fill all mandatory fields
3. Type to "Title" "xxxx"
4. Click on the pencil icon next to "Event link"/"Page link"
5. Type "xxxx" and save by clicking the floppy icon
6. Save the event/page
7. See the slug is changed but the message "The slug was changed from...to..." it not shown

### Reproducibility

Always (100%)

### Environment

Local (2026.4.1)

_No response_

### Severity

None

### Regression Information

_No response_

### Technical Context

_No response_

- [ ] The message is shown as described in "expected behaviour"
- [ ] No change for other cases

### Verification / Testing

- [ ]  Reproduce the bug, confirm it does not exist anymore

### Related Issues

The same bug was found in the POI form during review of #4210

JoeyStk:
### Short description

This PR is based off of #3837 and can only be reviewed (properly) after the Django update is open :). During our work on the upgrade, we had to do a few changes to the `poi_form_view` and `event_form_view`. Those changes increased the complexity of the files (for `poi_form_view` it even made it necessary to disable [a ruff rule](https://docs.astral.sh/ruff/rules/too-many-statements/)). This PR attempts to fix these issues.

For an easier view to review this, [this link](https://github.com/digitalfabrik/integreat-cms/compare/feature/python3.13-django-5.2...refactor-form-views) might be helpful :) 

--- Info ---
This PR was handed over to @MizukiTemma 

### Proposed changes

- Reduce CC of `event_form_view` from 30 to 26
- Reduce LOC of `event_form_view` from 201 to 177
- Reduce CC of `poi_form_view` from 33 to 14
- Reduce LOC of `poi_form_view` from 232 to 141
- Reduce statements in `poi_form_view`, so that rule can be enabled again

### Side effects

- This refactoring is by no means "perfect", these were just a few low hanging fruits I wanted to collect for us.
- Feel free to either extend this refactoring or refactor these files even more in the future
- As with all refactorings, I tried my best to pay attention to all the details, but of course there is a high chance of human error, so horough testing is highly appreciated.

### How to test

- All actions in event form and poi form :)

### Resolved issues

Fixes: #4208
Fixes: #4264 

### DIFF
--- a/integreat_cms/cms/models/chat/chat_message.py
+++ b/integreat_cms/cms/models/chat/chat_message.py
@@ -73,6 +73,9 @@ class ChatMessage(AbstractBaseModel):
 
         :return: The canonical string representation of the chat message
         """
+        class_name = type(self).__name__
+        if not self.pk:
+            return f"<{class_name} (unsaved instance)>"
         return f"<ChatMessage (id: {self.id}, sender: {self.sender.username}, date: {self.sent_datetime.strftime('%Y-%m-%d %H:%M')})>"
 
     class Meta:
--- a/integreat_cms/cms/views/events/event_form_view.py
+++ b/integreat_cms/cms/views/events/event_form_view.py
@@ -66,7 +66,6 @@ class EventFormView(
             only_active=True,
         )
 
-        # get event and event translation objects if they exist, otherwise objects are None
         event_instance = region.events.filter(id=kwargs.get("event_id")).first()
         event_translation_instance = language.event_translations.filter(
             event=event_instance,
@@ -74,38 +73,8 @@ class EventFormView(
         recurrence_rule_instance = RecurrenceRule.objects.filter(
             event=event_instance,
         ).first()
-        # Make form disabled if event is archived or user doesn't have the permission to edit the event
-        if event_instance and event_instance.archived:
-            disabled = True
-            messages.warning(
-                request,
-                _("You cannot edit this event because it is archived."),
-            )
-        elif event_instance and event_instance.external_calendar:
-            disabled = True
-            messages.warning(
-                request,
-                _(
-                    "You cannot edit this event because it was imported from an external calendar.",
-                ),
-            )
-        elif not request.user.has_perm("cms.change_event"):
-            disabled = True
-            messages.warning(
-                request,
-                _("You don't have the permission to edit events."),
-            )
-        elif not request.user.has_perm("cms.publish_event"):
-            disabled = False
-            messages.warning(
-                request,
-                _(
-                    "You don't have the permission to publish events, but you can propose changes and submit them for review instead.",
-                ),
-            )
-        else:
-            disabled = False
 
+        disabled = self.check_if_event_is_locked(request, event_instance)
         event_form = EventForm(
             instance=event_instance,
             disabled=disabled,
@@ -151,6 +120,7 @@ class EventFormView(
             },
         )
 
+    @transaction.atomic
     def post(self, request: HttpRequest, **kwargs: Any) -> HttpResponse:
         r"""
         Save event and ender event form for HTTP POST requests
@@ -182,165 +152,77 @@ class EventFormView(
             )
         )
 
-        # Check publish permissions
-        if request.POST.get("status") in [
-            status.DRAFT,
-            status.PUBLIC,
-        ] and not request.user.has_perm("cms.publish_event"):
-            raise PermissionDenied(
-                f"{request.user!r} does not have the permission 'cms.publish_event'",
-            )
+        # Save the user selected slug before is_valid() is called and it is changed for uniqueness
+        user_slug = event_translation_form.data.get("slug")
 
-        if (
-            event_translation_form.instance.status == status.AUTO_SAVE
-            and not event_form.has_changed()
-            and not event_translation_form.has_changed()
-            and not recurrence_rule_form.has_changed()
+        if self.is_qualified_for_save(
+            request, event_form, event_translation_form, recurrence_rule_form
         ):
-            messages.info(request, _("No changes detected, autosave skipped"))
-        else:
-            success = True
-            with transaction.atomic():
-                sid = transaction.savepoint()
+            sid = transaction.savepoint()
 
-                if not event_form.is_valid():
-                    event_form.add_error_messages(request)
-                    success = False
-                else:
-                    event = event_form.save()
-
-                if not event_translation_form.is_valid():
-                    event_translation_form.add_error_messages(request)
-                    success = False
-                elif success:
-                    event_translation_form.instance.event = event
-                    event_translation_instance = event_translation_form.save(
-                        foreign_form_changed=(
-                            event_form.has_changed()
-                            or recurrence_rule_form.has_changed()
-                        ),
-                    )
-
-                if success:
-                    if (
-                        event_form.cleaned_data.get("is_recurring")
-                        and not recurrence_rule_form.is_valid()
-                    ):
-                        recurrence_rule_form.add_error_messages(request)
-                        success = False
-                    elif event_form.cleaned_data.get("is_recurring"):
-                        # If event is recurring, save recurrence rule
-                        event_form.instance.recurrence_rule = (
-                            recurrence_rule_form.save()
-                        )
-                        event_form.instance.save(update_fields=["recurrence_rule"])
-                    elif event_form.instance.recurrence_rule:
-                        # If the event is not recurring but it was before, delete the associated recurrence rule
-                        event_form.instance.recurrence_rule.delete()
-                        event_form.instance.recurrence_rule = None
-                        event_form.instance.save(update_fields=["recurrence_rule"])
-
-                if success:
-                    # If any source translation changes to draft, set all depending translations/versions to draft
-                    if event_translation_form.instance.status == status.DRAFT:
-                        language_tree_node = region.language_node_by_slug.get(
-                            language.slug
-                        )
-                        languages = [language] + [
-                            node.language
-                            for node in language_tree_node.get_descendants()
-                        ]
-                        event_translation_form.instance.event.translations.filter(
-                            language__in=languages,
-                        ).update(status=status.DRAFT)
-
-                    elif (
-                        event_translation_form.instance.status == status.PUBLIC
-                        and event_translation_form.instance.minor_edit
-                    ):
-                        event_translation_form.instance.event.translations.filter(
-                            language=language,
-                        ).update(status=status.PUBLIC)
-
-                    # Show a message that the slug was changed if it was not unique
-                    user_slug = event_translation_form.data.get("slug")
-                    if (
-                        user_slug
-                        and user_slug != event_translation_form.cleaned_data["slug"]
-                    ):
-                        other_translation = EventTranslation.objects.filter(
-                            event__region=region,
-                            slug=user_slug,
-                            language=language,
-                        ).first()
-                        other_translation_link = other_translation.backend_edit_link
-                        message = _(
-                            "The slug was changed from '{user_slug}' to '{slug}', "
-                            "because '{user_slug}' is already used by <a>{translation}</a> or one of its previous versions.",
-                        ).format(
-                            user_slug=user_slug,
-                            slug=event_translation_form.cleaned_data["slug"],
-                            translation=other_translation,
-                        )
-                        messages.warning(
-                            request,
-                            translate_link(
-                                message,
-                                attributes={
-                                    "href": other_translation_link,
-                                    "class": "underline hover:no-underline",
-                                },
-                            ),
-                        )
-
-                    # Add the success message and redirect to the edit page
-                    if not event_instance:
-                        messages.success(
-                            request,
-                            _('Event "{}" was successfully created').format(
-                                event_translation_form.instance,
-                            ),
-                        )
-                    elif (
-                        not event_form.has_changed()
-                        and not event_translation_form.has_changed()
-                        and not recurrence_rule_form.has_changed()
-                    ):
-                        messages.info(
-                            request, _("No changes detected, but date refreshed")
-                        )
-                    else:
-                        # Add the success message
-                        event_translation_form.add_success_message(request)
-
-                    # Invalidate event translation cache to refresh API result
-                    invalidate_model(EventTranslation)
-
-                    return redirect(
-                        "edit_event",
-                        **{
-                            "event_id": event_form.instance.id,
-                            "region_slug": region.slug,
-                            "language_slug": language.slug,
-                        },
-                    )
-
-                # Failure: rollback and re-instantiate for clean re-render
-                transaction.savepoint_rollback(sid)
-                event_instance, event_translation_instance, recurrence_rule_instance = (
-                    self.get_instances(language=language, event_id=event_id)
+            if (
+                self.validate_and_save_event(request, event_form)
+                and self.validate_event_translation(request, event_translation_form)
+                and self.validate_recurrence_rule(
+                    request, event_form, recurrence_rule_form
                 )
-                event_form, event_translation_form, recurrence_rule_form = (
-                    self.instantiate_forms(
-                        request=request,
-                        event_instance=event_instance,
-                        recurrence_rule_instance=recurrence_rule_instance,
-                        event_translation_instance=event_translation_instance,
-                        poi=poi,
-                        language=language,
-                        region=region,
-                    )
+            ):
+                event_translation_instance = event_translation_form.save(
+                    foreign_form_changed=(
+                        event_form.has_changed() or recurrence_rule_form.has_changed()
+                    ),
                 )
+
+                self.update_recurrence_rule(event_form, recurrence_rule_form)
+
+                # If any source translation changes to draft, set all dependent translations/versions to draft
+                self.adjust_status_of_dependent_translations(
+                    event_translation_form,
+                    language,
+                    region,
+                )
+
+                self.show_slug_changed_message(
+                    request, language, region, user_slug, event_translation_form
+                )
+
+                forms_unchanged = (
+                    not event_form.has_changed()
+                    and not event_translation_form.has_changed()
+                    and not recurrence_rule_form.has_changed()
+                )
+                self.add_success_message(
+                    request, event_instance, forms_unchanged, event_translation_form
+                )
+
+                # Invalidate event translation cache to refresh API result
+                invalidate_model(EventTranslation)
+
+                return redirect(
+                    "edit_event",
+                    **{
+                        "event_id": event_form.instance.id,
+                        "region_slug": region.slug,
+                        "language_slug": language.slug,
+                    },
+                )
+
+            # Failure: rollback and re-instantiate for clean re-render
+            transaction.savepoint_rollback(sid)
+            event_instance, event_translation_instance, recurrence_rule_instance = (
+                self.get_instances(language=language, event_id=event_id)
+            )
+            event_form, event_translation_form, recurrence_rule_form = (
+                self.instantiate_forms(
+                    request=request,
+                    event_instance=event_instance,
+                    recurrence_rule_instance=recurrence_rule_instance,
+                    event_translation_instance=event_translation_instance,
+                    poi=poi,
+                    language=language,
+                    region=region,
+                )
+            )
         url_link = f"{settings.WEBAPP_URL}/{region.slug}/{language.slug}/{event_translation_form.instance.url_infix}/"
         return render(
             request,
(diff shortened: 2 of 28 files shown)


integreat_cms/cms/views/pois/poi_form_view.py
```diff
@@ -470,7 +460,62 @@ def set_success_messages(
...
+        return disabled
+
+    def adjust_status_of_depending_translations(
+        self,
+        poi_translation_form: POITranslationForm,
+        language: Language,
+        region: Region,
+    ) -> None:
+        """
+        Change the status of all depending translations of the POI according to the status of the current translation.
```
seluianova:
```suggestion
        Change the status of all dependent translations of the POI according to the status of the current translation.
```

integreat_cms/cms/views/pois/poi_form_view.py
```diff
@@ -470,7 +460,62 @@ def set_success_messages(
...
+        ):
+            poi_translation_form.instance.poi.translations.filter(
+                language=language,
+            ).update(status=status.PUBLIC)
+
+    def warn_if_coordinates_too_far(
+        self, request: HttpRequest, poi_form: POIForm
+    ) -> None:
+        """
+        Warn the user if the coordinates are too far from the region's center.
```
seluianova:
```suggestion
        Warn the user if the manually entered coordinates are too far from the entered address.
```

integreat_cms/cms/views/pois/poi_form_view.py
```diff
@@ -470,7 +460,62 @@ def set_success_messages(
...
+        elif not request.user.has_perm("cms.change_poi"):
+            disabled = True
+            messages.warning(
+                request,
+                _("You don't have the permission to edit locations."),
+            )
+
+        return disabled
+
+    def adjust_status_of_depending_translations(
```
seluianova:
```suggestion
    def adjust_status_of_dependent_translations(
```

integreat_cms/cms/models/chat/chat_message.py
```diff
@@ -73,6 +73,9 @@ def get_repr(self) -> str:
 
         :return: The canonical string representation of the chat message
         """
+        class_name = type(self).__name__
```
seluianova, line 76: If we put this into the AbstractContentModel class, there will be no need to duplicate it in all child classes.
Or am I missing something?

integreat_cms/cms/views/pois/poi_form_view.py
seluianova: btw there are more typos like this depending -> dependent

MizukiTemma:
Update: after rebasing etc, #4264 must be fixed too.

integreat_cms/cms/views/events/event_form_view.py
```diff
@@ -66,46 +66,14 @@ def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
...
-            messages.warning(
-                request,
-                _(
-                    "You don't have the permission to publish events, but you can propose changes and submit them for review instead.",
-                ),
-            )
-        else:
-            disabled = False
-
+        event_instance = self.get_first_or_none(Event, id=kwargs.get("event_id"))
```
seluianova: 🪲 it was a different statement before:
`region.events.filter(id=kwargs.get("event_id")).first()`

using `get_first_or_none` introduces a regression bug.
users can access events by ID that are not related to their region.
for example, I'm a region admin in Augsburg.
there is no event with ID = 2 in Augsburg, but there is one in Nürnberg.
and I can load it in the form like this:

Image

integreat_cms/cms/views/events/event_form_view.py
```diff
@@ -66,46 +66,14 @@ def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
...
+        recurrence_rule_instance = self.get_first_or_none(
+            RecurrenceRule, event=event_instance
+        )
+        event_translation_instance = self.get_first_or_none(
```
seluianova: it was also a different statement before:
```
        event_translation_instance = language.event_translations.filter(
            event=event_instance,
        ).first()
```
`language.event_translations` uses a different queryset.
even so I don't see any functional impact of this change, I think the previous version was more explicit...

integreat_cms/cms/views/events/event_form_view.py
MizukiTemma: Nice catch 🤯 

`get_first_or_none` is reverted. I don't think it's really making codes shorter or more readable to be honest😅

integreat_cms/cms/views/pois/poi_form_view.py
MizukiTemma: Applied also for `event_form_view`

integreat_cms/cms/views/events/event_form_view.py
MizukiTemma: Agree. The old version is restored ↩️

integreat_cms/cms/views/pois/poi_form_view.py
MizukiTemma: Applied :)

integreat_cms/cms/models/chat/chat_message.py
MizukiTemma: The child classes are inheriting from either `AbstractContentModel` (`Event`, `POI`, `Imprint` and `AbstractBasePage -> Page`) or `AbstractContentTranslation` (`EventTranslation`, `POITranslation`, `AbstractBasePageTranslation -> ImprintTranslation` and `AbstractBasePageTranslation -> PageTranslation`) or `AbstractBaseModel` (all other child classes).

 `AbstractContentModel` and `AbstractContentTranslation` have its own `get_repr` and duplicates in their child classes are already avoided.

We could do the same with `AbstractBaseModel` for all other child classes, but they need individual cuscomisation (because they have different fields each another). I don't think we can genelise more than the actual status 🤔

MizukiTemma:
@seluianova 
Thank you for the review 😸 Your suggestions are applied or commented back. Ready for re-review 🔍

integreat_cms/cms/views/events/event_form_view.py
```diff
@@ -413,3 +332,128 @@ def get_instances(
...
+            languages = [language] + [
+                node.language for node in language_tree_node.get_descendants()
+            ]
+            event_translation_form.instance.event.translations.filter(
+                language__in=languages,
+            ).update(status=status.DRAFT)
+
+        elif (
+            event_translation_form.instance.status == status.PUBLIC
+            and event_translation_form.instance.minor_edit
```
seluianova, line 418: _I see this was copied as is, so this is just a question for my understanding:_

If a translation is updated to public and the “minor edit” checkbox is NOT set, we do not update all previous versions to public.

But if a translation is updated to public and the “minor edit” checkbox IS set, we update all previous versions to public.

Why? 🤔

integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -218,91 +218,60 @@ def post(
...
+                "edit_page",
+                **{
+                    "page_id": page_form.instance.id,
+                    "region_slug": region.slug,
+                    "language_slug": language.slug,
+                },
+            )
 
             # Failure: rollback and re-instantiate for clean re-render
             transaction.savepoint_rollback(sid)
```
seluianova, line 273: 🪲 this code is not reachable (regression issue, was fine before)

integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -690,3 +659,43 @@ def instantiate_forms(
...
+    def pre_validate_page_update(
+        self,
+        request: HttpRequest,
+        page_form: PageForm,
+        page_translation_form: PageTranslationForm,
+    ) -> bool:
+        """
+        Pre-validates whether the page and page translation forms can be saved.
+        """
+        page = None
```
seluianova:
```suggestion
```
this is not needed?

integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -690,3 +659,43 @@ def instantiate_forms(
...
+        if not page_form.is_valid():
+            page_form.add_error_messages(request)
+            return False
+        page = page_form.save()
```
seluianova, line 692: I wonder why we save the form before the `page_translation_form.is_valid()` check, _but it was like this before as well..._

integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -546,63 +515,63 @@ def check_autosave_or_pending_review(
...
+                    ),
+                )
+            else:
+                logger.warning(
+                    "Slug was changed from the one the user provided, but we can't find the translation that already used it: %s (cleaned to %s)",
+                    user_slug,
+                    page_translation_form.cleaned_data["slug"],
+                )
+                messages.warning(
+                    _("The slug was changed from '{user_slug}' to '{slug}.").format(
```
seluianova:
```suggestion
                    _("The slug was changed from '{user_slug}' to '{slug}'.").format(
```

integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -690,3 +659,43 @@ def instantiate_forms(
...
+        if not page_form.is_valid():
+            page_form.add_error_messages(request)
+            return False
+        page = page_form.save()
+
+        if not page_translation_form.is_valid():
+            page_translation_form.add_error_messages(request)
+            return False
+
+        if not request.user.has_perm(
```
seluianova, line 668: permissions are checked after `page = page_form.save()`. 
this is a regression issue; on develop, the permission check happens before saving.

integreat_cms/cms/views/pages/page_form_view.py
hannaseithe: This is necessary now after the upgrade to Django 5.2. A form will show as invalid if a foreign key does not exist yet.

integreat_cms/cms/views/pages/page_form_view.py
MizukiTemma: Fixed 🔨

integreat_cms/cms/views/pages/page_form_view.py
MizukiTemma: Applied

integreat_cms/cms/views/pages/page_form_view.py
MizukiTemma: Ohw 🙈 Nice catch 👍 

Should be fixed by the commit "make rollback reachable"

integreat_cms/cms/views/events/event_form_view.py
MizukiTemma: @digitalfabrik/team-cms 
Does anyone remember? 🙈 I don't find any description from the old git PRs

integreat_cms/cms/views/events/event_form_view.py
```diff
@@ -241,77 +213,24 @@ def post(self, request: HttpRequest, **kwargs: Any) -> HttpResponse:
...
                         and not event_translation_form.has_changed()
-                        and not recurrence_rule_form.has_changed()
-                    ):
-                        messages.info(
-                            request, _("No changes detected, but date refreshed")
-                        )
-                    else:
-                        # Add the success message
-                        event_translation_form.add_success_message(request)
+                    )
```
hannaseithe:
```suggestion
                    and not recurrence_rule_form.has_changed()
                    )
```

integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -690,3 +658,42 @@ def instantiate_forms(
             changed_by_user=request.user,
         )
         return page_form, page_translation_form
+
+    def pre_validate_page_update(
```
hannaseithe: My suggestion on how to make this part more readable

```suggestion
    def pre_validate_page_update(
        self,
        request: HttpRequest,
        page_form: PageForm,
        page_translation_form: PageTranslationForm,
    ) -> bool:
        """
        Pre-validates whether the page and page translation forms can be saved.
        """
        if not request.user.has_perm(
            "cms.publish_page_object",
            page_form.instance,
        ) and request.POST.get("status") in [
            status.DRAFT,
            status.PUBLIC,
        ]:
            raise PermissionDenied(
                f"{request.user!r} does not have the permission to publish {page_form.instance!r}"
            )
        
        if (
            page_translation_form.instance.status == status.AUTO_SAVE
            and not page_form.has_changed()
            and not page_translation_form.has_changed()
        ):
            messages.info(request, _("No changes detected, autosave skipped"))
            return False
        
    def validate_and_save_page(self,
                          request: HttpRequest,
                          page_form: PageForm):
        
        if not page_form.is_valid():
            page_form.add_error_messages(request)
            return False
        page = page_form.save()
        return page is not None

    def validate_page_translation(
            self,
            request: HttpRequest,
            page_translation_form
    ):

        if not page_translation_form.is_valid():
            page_translation_form.add_error_messages(request)
            return False
```

integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -218,105 +218,73 @@ def post(
...
-            status.DRAFT,
-            status.PUBLIC,
-        ]:
-            raise PermissionDenied(
-                f"{request.user!r} does not have the permission to publish {page_form.instance!r}"
+        # Save the user selected slug before is_valid() is called and it is changed for uniqueness
+        user_slug = page_translation_form.data.get("slug")
+
+        sid = transaction.savepoint()
+        if self.pre_validate_page_update(request, page_form, page_translation_form):
```
hannaseithe:
```suggestion
        if (self.pre_validate_page_update(request, page_form, page_translation_form)
            and self.validate_save_page(request,page_form)
            and self.validate_page_translation(request,page_translation_form)):
```

integreat_cms/cms/views/pois/poi_form_view.py
```diff
@@ -470,7 +460,62 @@ def set_success_messages(
         else:
             # Add the success message
             poi_translation_form.add_success_message(request)
-        # Show warning if nominatim result is more than 10km away from manually entered coordinates
+
+    def check_if_content_is_locked(
```
hannaseithe:
```suggestion
    def check_if_poi_is_locked(
```
To be consistent with the naming of the method in `event_form_view.py`

hannaseithe (changes requested):
Thank you for working on this. This certainly shortens the post methods int the form views. Unfortunately I feel readability has suffered in places and I would suggest a few improvements:

1.) I think it is a good idea to get rid of the success pattern, for the sake of readybility as you did in `page_form_view.py`.  But If we get rid of the `success` pattern in page_form_view-py, I think we should also get rid of it in the other two form views for consistency sake. Overall those three form views should follow the same pattern (which would mean implementing something like the validation methods in page_form_view.py as well). 

2.) pre_validate_update_page is too much of a blackbox for me. It worsens readability. I dont like that permission validation, page save and form validation is all in one method that then is called something that implies to me we are doing some validation before we start the saving process. See my suggestions below

integreat_cms/cms/views/pois/poi_form_view.py
MizukiTemma: Applied :)

integreat_cms/cms/views/events/event_form_view.py
MizukiTemma: Nice catch 😸

integreat_cms/cms/views/events/event_form_view.py
```diff
@@ -413,3 +333,128 @@ def get_instances(
...
+                        "href": other_translation_link,
+                        "class": "underline hover:no-underline",
+                    },
+                ),
+            )
+
+    def add_success_message(
+        self,
+        request: HttpRequest,
+        event_instance: EventTranslation,
```
seluianova:
```suggestion
        event_instance: Event | None,
```

integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -690,3 +658,42 @@ def instantiate_forms(
...
+        if not page_translation_form.is_valid():
+            page_translation_form.add_error_messages(request)
+            return False
+
+        if (
+            page_translation_form.instance.status == status.AUTO_SAVE
+            and not page_form.has_changed()
+            and not page_translation_form.has_changed()
+        ):
+            messages.info(request, _("No changes detected, autosave skipped"))
```
seluianova, line 684: I think returning pre_validate_page_update = False if "no changes detected" is not semantically correct, because we use this value to trigger a rollback, which does not make sense in this case.
So we do: page_form.save() first, then check if "no changes detected", if true - do rollback.

On the develop branch, we did this check before starting the transaction, and I think that was more logical semantically. Because this was a successful path before, but now this is handled like an error path.

integreat_cms/cms/views/pages/page_form_view.py
seluianova: It does not break anything, but I’m a bit concerned that we do the permission check **after** opening the transaction.

Maybe the permission check should not be part of this function...? 🤔

Even though this does not have any functional impact, I think it would be cleaner (and a bit more efficient) to check permissions before opening the transaction, as it was before.

**UPD**. Also, this is currently handled differently in `event_form_view`, where we do the permission check before opening the transaction. So I would suggest unifying that behavior.

integreat_cms/cms/views/events/event_form_view.py
seluianova: if nobody needs that, maybe it should be removed 🙈

integreat_cms/cms/views/pages/page_form_view.py
seluianova: btw this check is not quite correct itself.
it checks the previous status, not the new one...

`page_translation_form.instance.status == status.AUTO_SAVE`

it means when the page is auto-saved but the previous status was public or draft, this condition won't be met...

but it is **not** a regression issue, it was like that before...

**UPD**. I created a separate issue for that https://github.com/digitalfabrik/integreat-cms/issues/4311

integreat_cms/cms/views/events/event_form_view.py
MizukiTemma: I'm not sure whether other devs are reading here, so opened an issue #4317 , leaving it outside of the scope of this PR as it exists in `develop` branch

integreat_cms/cms/views/pages/page_form_view.py
MizukiTemma: The impelemntation is changed so the permission check is done before `transaction.savepoint()` 🔨

integreat_cms/cms/views/pages/page_form_view.py
MizukiTemma: > I think returning pre_validate_page_update = False if "no changes detected" is not semantically correct, because we use this value to trigger a rollback, which does not make sense in this case.
So we do: page_form.save() first, then check if "no changes detected", if true - do rollback.

On the develop branch, we did this check before starting the transaction, and I think that was more logical semantically. Because this was a successful path before, but now this is handled like an error path.

It's again skipping `save()`s and the transaction if no change is detected. I hope it is now more preserving the old logic 🔨

MizukiTemma:
I'll re-structure `event_form_view` and `poi_form_view` once `page_form_view` is completed.

integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -218,61 +218,33 @@ def post(
...
+            ) and self.validate_page_translation(request, page_translation_form):
+                # Only save page form if page does not yet exist or if translation is no auto save
                 if (
                     not page_instance
                     or page_translation_form.instance.status != status.AUTO_SAVE
                 ):
-                    page_translation_form.instance.page = page
+                    page_translation_form.instance.page = page_form.instance
 
                 page_translation_instance = page_translation_form.save(
```
seluianova, line 230: this assignment is unused

**UPD**. it was also unused before 🙈
but I think the variable can be removed anyway.

integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -690,3 +662,47 @@ def instantiate_forms(
...
         return page_form, page_translation_form
+
+    def is_qualified_for_save(
+        self,
+        request: HttpRequest,
+        page_form: PageForm,
+        page_translation_form: PageTranslationForm,
+    ) -> bool:
+        """
+        Pre-validates whether the page and page translation forms can be saved.
```
seluianova:
```suggestion
        Checks whether the current request should proceed with saving the page and translation.
```
for validation we have separate functions

seluianova: optional: alternatively maybe it would make sense to split this function into 2, e.g. `enforce_publish_permission()` and `is_duplicate_autosave()`.
currently it checks two different things and the name `is_qualified_for_save()` is quite vague

seluianova:
> I'll re-structure event_form_view and poi_form_view once page_form_view is completed.

@MizukiTemma I feel like I could endlessly suggest nitpicks for this refactoring, but maybe I should stop 🙈
in the current version of `page_form_view`, I don't see any regression issues anymore.
so in my opinion, it's good to go.

integreat_cms/cms/views/pages/page_form_view.py
MizukiTemma: I'll keep the structure as is but update the description as suggsted ✏️

integreat_cms/cms/views/pages/page_form_view.py
MizukiTemma: 🙈 removed

integreat_cms/cms/views/pages/page_form_view.py
```diff
@@ -218,61 +218,26 @@ def post(
...
-                if (
-                    user_slug
-                    and user_slug != page_translation_form.cleaned_data["slug"]
-                ):
-                    self.report_conflicting_slug(
-                        request, page_translation_form, region, user_slug, language
-                    )
-
                 if page_translation_form.instance.status == status.DRAFT:
                     self.set_translations_to_draft(
```
hannaseithe: For readability's sake (if we are already at it), I would rename the method as `set_dependent_translations_to_draft`. And also leave a comment why we change all versions of dependent translations to draft.

hannaseithe (commented):
I have one more minor readability request for post in  page_form_view.py. I feel the readability of the method has greatly improved. Someone not knowledgable with the code, should be able to quickly get what we are doing reading from top to bottom. That makes me happy!

MizukiTemma:
@hannaseithe @seluianova 
I think all your suggestions are now applied incl. `event_form_view` and `poi_form_view`

Please re-review 🙏

integreat_cms/cms/views/events/event_form_view.py
```diff
@@ -182,165 +152,77 @@ def post(self, request: HttpRequest, **kwargs: Any) -> HttpResponse:
...
-                # Failure: rollback and re-instantiate for clean re-render
-                transaction.savepoint_rollback(sid)
-                event_instance, event_translation_instance, recurrence_rule_instance = (
-                    self.get_instances(language=language, event_id=event_id)
+            sid = transaction.savepoint()
+
+            if (
+                self.validate_and_save_event(request, event_form)
+                and self.validate_event_translation(request, event_translation_form)
+                and self.validate_and_save_recurrence_rule(
```
hannaseithe:
```suggestion
                and self.validate_recurrence_rule(
```
It does not save it

hannaseithe (approved):
I just have minor change request about naming. Other than that everything looks good to me. I genuinely believe this PR is a great improvement in terms of readability! Thank you very much for working on it!

integreat_cms/cms/views/events/event_form_view.py
```diff
@@ -151,6 +120,7 @@ def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
             },
         )
 
+    @transaction.atomic
```
seluianova, line 123: this opens the transaction for the whole post function, right?
which means the transaction will occupy the database connection longer than before (when we only opened the transaction to save event, translation and recurrence rule objects), but do we need this?
might be a nit-pick, up to you

integreat_cms/cms/views/pois/poi_form_view.py
```diff
@@ -481,3 +470,68 @@ def set_success_messages(
...
+
+    def validate_and_save_related_contact(
+        self,
+        new_poi_and_new_related_contact: bool,
+        poi_form: POIForm,
+        data: dict[str, str | list[str]],
+        request: HttpRequest,
+    ) -> bool:
+        # only attempt to create a related contact if we are in creation mode and relevant fields have been added
+        if new_poi_and_new_related_contact:
```
seluianova, line 515: minor regression issue:
when I create poi with new related contact, the following exception appears in the log
```
Jun 08 09:48:55 DEBUG integreat_cms.cms.forms.contacts.contact_form - ContactForm validated with cleaned data {'area_of_responsibility': 'Different Questions', 'name': '', 'location': <POI (id: 23)>, 'email': '', 'phone_number': '', 'mobile_phone_number': '', 'website': '', 'opening_hours': None, 'appointment_url': '', 'use_location_opening_hours': False}
Jun 08 09:48:55 DEBUG integreat_cms.cms.models.abstract_base_model - repr() for object <POI (id: 23)> failed because of AttributeError: 'NoneType' object has no attribute 'slug' (If you think this is no problem, please exclude this exception in the repr() method of the AbstractBaseModel.)
  File "/Users/tory/Integreat/integreat-cms/integreat_cms/cms/models/abstract_base_model.py", line 50, in __repr__
    return self.get_repr()
           ^^^^^^^^^^^^^^^
  File "/Users/tory/Integreat/integreat-cms/integreat_cms/cms/models/abstract_content_model.py", line 628, in get_repr
    translation_slug = f", slug: {self.best_translation.slug}" if self.id else ""
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'slug'
```
not reproduced on develop.
but I see no functional impact from this, so up to you

seluianova (approved):
looks nice to me 👍
just two more findings that change the current flow a bit, without any visible functional impact though. up to you whether you'd like to address them.

integreat_cms/cms/views/events/event_form_view.py
seluianova: I’m just thinking… if we wrap the whole function in @transaction.atomic, maybe we can remove the manual rollback control?
or we could analyze this later as a separate task.
I just have a feeling that we probably don’t need both manual rollback and @transaction.atomic.

MizukiTemma: I have outsourced the topic to #4334

MizukiTemma:
@hannaseithe @seluianova 
Thank you for all your reviews and patience 🙏 

@JoeyStk 
We made it, finaly 😹

Issue #3179, opened 2024-11-06
Media library: slow loading time


juliankehne:
### Describe the Bug
When clicking on a folder inside the media library, depending on the amount of content/images inside the folder, the loading speed of it is quite slow. For the folders 'Icons (2.Ebene)' and 'Region Logos' it takes up to 3 seconds.

The loading indicator shows up for a split second and then vanishes.

### Steps to Reproduce

Loading Time (Video 1):
1. Go Pages and create or edit any page (Or any other place in which the media library is used)
2. Edit the Page Icon (Media library opens)
3. Open one of the subfolders ('Icons 2. Ebene' or 'Region Logos' as they have similar amount of images)
4. See that the folder doesn't open instantly and it takes 2-3 seconds for it to open

### Expected Behavior
For example on windows: The folder opens instantly. The first images shown to the user are already loaded. If there are too many images in a folder to show at once: As the user scrolls down the unloaded images/thumbnails are load.

Example on Nextcloud: The folder opens instantly. Icon and image thumbnails are only loaded after the folder has openened.
If the folder doesn't load instantly there is a loading indicator.

A folder doesn't take multiple seconds to open.

The loading indicator shows for the whole loading duration.

### Actual Behavior ###
After clicking on a subfolder with many images in it, it can take multiple seconds to load.

The loading indicator vanishes after a short amount of time while a folder is still loading.

See Video below.

### Additional Information

Another loading time example:
1. Go to https://cms-test.integreat-app.de/albdonaukreis/media-library/
2. See that media library loads multiple seconds when openend

Video 1 - Long loading time:

https://github.com/user-attachments/assets/2139aa1f-008d-440d-b6db-6e4ccb0d09d6

juliankehne:
Please let me know if there are questions and if I should split it into multiple issues :)

PR #4079, opened 2026-01-13
Media library: Fix loading indicator and improve loading time

hannaseithe:
### Short description
This PR fixes the bug in the media library that
a) the loading indicator vanishes before the subfolder is loaded
b) folders take a long time to load

### Proposed changes

- change the function `ajaxRequest` (used inside preact) to not control the behavior of the loading indicator
- instead have the "eventlisteners" (useEffect and toggleFileUsages) control it

- modify the mediaFile query to use prefetch_related since inside the `serialize` method of the MediaFile model the field `is_deletable` would trigger the property-method `icon_usage` which has 5 related_fields queries (on production @svenseeberg was able to prove that this reduces a 12s call to 2s)

For reference, you can read up here on, how we debugged the endpoint: https://chat.tuerantuer.org/digitalfabrik/pl/yfrc5dtxfp8x9czbdtso4tiyar

### Side effects

- None

### How to test

- check that the loading indicator sticks around until the folder is loaded (the actual images might take longer, if they are not cached though) -> compare to the video in the Bugs issue
- the loading time improvement is most obvious to be seen with production data and in regions with loads of events, pois, pages that use a lot of media files & a lot of media files
- locally you can check for the improvement by logging with `logger.error(time.time())` before and after the list comprehension in line 99 of `media_actions.py`

### Resolved issues

Fixes: #3179

### DIFF
--- a/integreat_cms/cms/views/media/media_actions.py
+++ b/integreat_cms/cms/views/media/media_actions.py
@@ -90,7 +90,7 @@ def get_directory_content_ajax(
     media_files = MediaFile.objects.filter(
         Q(region=region) | Q(region__isnull=True, is_hidden=False),
         Q(parent_directory=directory),
-    )
+    ).prefetch_related("icon_organizations", "icon_regions", "events", "pages", "pois")
     directories = Directory.objects.filter(
         Q(region=region) | Q(region__isnull=True, is_hidden=False),
         parent=directory,
--- a/integreat_cms/static/src/js/media-management/component/edit-sidebar.tsx
+++ b/integreat_cms/static/src/js/media-management/component/edit-sidebar.tsx
@@ -53,7 +53,7 @@ type Props = {
         urlParams: URLSearchParams,
         successCallback: (data: any) => void,
         loadingSetter?: Dispatch<StateUpdater<boolean>>
-    ) => void;
+    ) => Promise<void>;
     isLoading: boolean;
     canDeleteFile: boolean;
     canReplaceFile: boolean;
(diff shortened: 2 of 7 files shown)


dkehne:
## Review Summary

### What it does
1. **Backend:** Adds `prefetch_related` for related fields to prevent N+1 queries (12s → 2s improvement)
2. **Frontend:** Refactors loading state management so components control their own loading indicators

### Key observations

1. **Performance fix is solid** - The `prefetch_related("icon_organizations", "icon_regions", "events", "pages", "pois")` directly addresses the N+1 query problem that caused slow loading

2. **Loading state management is cleaner** - Moving loading state control from `ajaxRequest` to individual components ensures the indicator stays until the component is truly ready

3. **Async/await pattern is correct** - `ajaxRequest` now returns `Promise<void>` and callers properly await it

4. **Good use of Promise.all** - In `directory-content-library.tsx`, loading path and content in parallel is efficient

5. **Inconsistent error handling** - `directory-content-library.tsx` correctly uses `try/finally` to ensure `setLoading(false)` is called even on errors. However:
   - `filter-result-library.tsx` - `setLoading(false)` is outside try block
   - `search-result-library.tsx` - Same issue
   
   If an error occurs in these components, the loading indicator may get stuck.

**Suggestion:** Wrap `setLoading(false)` in `finally` blocks in `filter-result-library.tsx` and `search-result-library.tsx` for consistency:
```tsx
try {
    // ... ajax call
} finally {
    setLoading(false);
}
```

Overall good performance fix and cleaner architecture. Minor inconsistency in error handling across components.

jarlhengstmengel (approved):
Nice, thank you very much for the PR! Looks good to me. I tested with testdata and with a datadump. The loading symbol stayed as I would expect. I tested the sidebar, the normal view, filtered and with the search. With the datadump I only had the problem that I hadn't the actual picture files wich starts a error loop in the console as the files can't be found, but that is I think nothing that was touched by this PR.

MizukiTemma (approved):
Looks good 😸 

I tested with the local default data and a lot of pictures uploaded (and used in some contents). View, loading, filter, and insert popup in a page worked fine 👍

Issue #3917, opened 2025-10-01
Remove migration `0142_make_existing_slugs_unique` and run `make_slugs_unique` as script


MizukiTemma:
During testing of the release `2025.10.0` it turned out that the migraion `0142_make_existing_slugs_unique` requires a lo of calculation and run time. For a safe release and deployment, we remove the migration and run the change as script.

- [x] Remove `0142_make_existing_slugs_unique` from the migration #3918 
- [x] Create Management Command to run migration and optimize for speed #4082 
- [x] Run Management Command on Testsystem - if it is significantly faster, consider including it as migration in #3826 
- [ ] (if not included as migration): Run `make_slugs_unique` as script after the release. Do not forget Malte and Netzwerk Obdach. 

Relates to: #3826 
Blocks: #3060

hannaseithe:
I am closing this issue as deciding whether it should be a migration or not and running the make_slugs_unique command or not is part of #3060

PR #4082, opened 2026-01-15
Management Command for async script to make all slugs unique

hannaseithe:
### Short description

This PR is based on: #3917 
- a (@shared_task) script that makes sure all ~~latest~~ translations of events, pois and pages have unique slugs (EDIT: I have decided to delete the query for only latest translations, since the database constraint will guarantee it for all)
- a management command that queues the script with celery

### Proposed changes

- the script is optimized for speed compared to the original migration script
- if this command runs fast enough, I will turn it into a migration and add it to #3826 
- I have added no tests for now, as its unsure we will keep the command anyways and because I see no way to create a test state with duplicate slugs with the application level constraints already in place and the database constraints coming up soon

### Side effects

- None

### How to test

- You will have to manipulate the test-data.json to get duplicate slugs (within one region and language!)
- Or enter the slugs into the database directly (django-admin wont work)
- in the shell you can use the .update method or the bulk_create method as they circumvent the application level constraints that have already been implemented

Then run the command with `tools/integreat-cms-cli make_slugs_unique`

### Resolved issues

Fixes: #
Blocks: #3826 
Relates to #3917 

### Follow Up

- [x] Run command on Test-System after 2026-02-17 `integreat-cms-cli make_slug_unique --dry-run` and check for the celery logs: https://chat.tuerantuer.org/digitalfabrik/pl/gf855khc3bg47jq78kwgo4egqw
- 2026-03-31: 
  - duration dry-run: 489.63871692s    / slugs identified:  1051 identified slugs
  - duration real run: 45.461s   / slugs fixed:   675 updated slugs 
  - !!!BUT!!!   duration second real run: 489.3568178640003s / slugs fixed: 0 updated slugs
  That means the less duplicate slugs we have the longer it takes

- 2026-05-04:
  - the quick duration on first run from 2026-03-31 could not be replicated locally with a production database dump
  - on two different machines: the runs lasted app. 10 to 15 minutes for 939 updated slugs

### DIFF
--- a/integreat_cms/cms/utils/slug_utils.py
+++ b/integreat_cms/cms/utils/slug_utils.py
@@ -5,17 +5,27 @@ This module contains helpers regarding unique string identifiers without special
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING
+from time import time
+from typing import Final, Literal, TYPE_CHECKING
 
+from celery import shared_task
+from django.apps import apps
 from django.conf import settings
 from django.core.exceptions import ValidationError
+from django.db import transaction
 from django.utils.text import slugify
 from django.utils.translation import gettext_lazy as _
 
 if TYPE_CHECKING:
-    from typing import Any, Literal, NotRequired, TypeAlias, TypedDict, Unpack
+    from typing import (
+        Any,
+        NotRequired,
+        TypeAlias,
+        TypedDict,
+        Unpack,
+    )
 
-    from django.db.models import Manager
+    from django.db.models import Manager, QuerySet
     from django.forms import ModelForm
     from django.http.request import QueryDict
 
@@ -23,6 +33,9 @@ if TYPE_CHECKING:
     from ..models.abstract_base_model import AbstractBaseModel
     from ..models.abstract_content_model import AbstractContentModel
 
+SlugObject = Literal["page", "event", "poi"]
+DEFAULT_OBJECTS: Final[tuple[SlugObject, ...]] = ("event", "poi", "page")
+ALLOWED_OBJECTS = {"event", "poi", "page"}
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


dkehne:
## Review Summary

### What it does
- Creates a Celery `@shared_task` called `make_all_slugs_unique` that iterates through all page/event/POI translations and regenerates slugs to ensure uniqueness
- Adds a management command `make_slugs_unique` that queues this task
- Supports `--dry-run` flag and filtering by object type (`--objects page event poi`)

### Key observations

1. **Uses existing `generate_unique_slug`** - Reuses the established slug generation logic, which is good for consistency

2. **Transaction handling** - Uses `transaction.atomic()` to batch changes, good for data integrity

3. **Celery task with `.delay()`** - The command queues the task asynchronously, which is appropriate for potentially long-running operations

4. **Missing `from __future__ import annotations`** - The new management command file is missing this required import (per project guidelines)

5. **Type import issue** - `Final` is imported inside `TYPE_CHECKING` block but used at module level for `DEFAULT_OBJECTS`. This will cause a runtime `NameError`:
   ```python
   DEFAULT_OBJECTS: Final[tuple[SlugObject, ...]] = ("event", "poi", "page")
   ```
   `Final` needs to be imported outside the `TYPE_CHECKING` block.

6. **No tests** - Author acknowledges this and explains why (can't easily create duplicate slug state with constraints in place)

7. **Dry-run passed to Celery** - The `dry_run` parameter is passed to the async task, which works but means you can't see real-time output of what would change

**Potential bug:** The `Final` type annotation will cause a `NameError` at runtime since it's only imported under `TYPE_CHECKING`.

integreat_cms/cms/utils/slug_utils.py
```diff
@@ -182,3 +195,60 @@ def generate_unique_slug(**kwargs: Unpack[SlugKwargs]) -> str:
...
+                slug=translation.slug,
+                manager=type(translation).objects,
+                object_instance=translation,
+                foreign_model=foreign_attr,
+                foreign_object=foreign_obj,
+                region=region,
+                language=translation.language,
+            )
+            if old_slug != translation.slug:
+                counter += 1
```
jarlhengstmengel: Minor nit: In a dry run the counter can be a bit confusing since the changes aren't saved, the counter gets +1 for every slug it encounters that is not unique. In a normal run it gives the count of the changed slugs as expectedt

jarlhengstmengel (commented):
Thank you for the work! I tested the script with different parameters and it seems to work fine. All not unique slugs that I added, where changed as expected. I also started a run with a datadump. It still takes a lot of time (I abborted after about 10 mins). So I would argue to leave this as a script and not turning it into a migration. In that case I think it would be good to add documenatation to `management-command.rst`. For the rest I just had one minor nit about the counter of updatet slugs, but I don't think it's that relevant.

jarlhengstmengel (approved):
> I do agree, but I still want to test this myself once this is merged on develop. Depending on the time it takes, I will create a second PR that either adds release note and any other necessary documentation on how and why to use this command OR create a migration.

Then this looks fine to me and I approve :) thank you!

MizukiTemma (approved):
Looks good and works as expected 😸 Thank you!

You can probably add this command to `docs/src/management-commands.rst`

hannaseithe:
> Looks good and works as expected 😸 Thank you!
> 
> You can probably add this command to `docs/src/management-commands.rst`

My plan is to merge it first and test how much time it takes on the Testsystem and depending on that either add documentation for the command or add a migration to the PR for the database constraints

PR #4083, opened 2026-01-16
Fix slug being overwritten when editing page title

### LINKED ISSUE #3961: editing page titles overrides slug
### Describe the Bug

We allow manually setting a slug for a content object (e.g. a page). When a slug is set in this way and deviates from the title, it will be overridden as soon as the "Title" field is selected and deselected

### Steps to Reproduce

1. Go to a page
2. Explicitly set the slug to something different then the title
3. Save the page
4. Edit the title of the page and click somewhere outside of the "Title" field
5. See that the slug is changed as well

### Expected Behavior

Manually set slugs should not be overridden by the title

### Actual Behavior

The title always sets a slug as soon as the title field is edited in the form

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

### Related Issues

Discovered in the context of #3856

dkehne:
## Summary
- Only auto-generate the slug from the title if the slug field is empty
- Prevents manually set slugs from being overwritten when editing the title

Fixed as discussed in #3961. Slug will only be adjusted/created when the field is empty. If the title changes it will not be affected.

## Test plan
- [ ] Edit an existing page with a custom slug, change the title, verify slug is NOT overwritten
- [ ] Create a new page, enter a title, verify slug IS auto-generated
- [ ] Works for pages, events, and POIs

Fixes #3961

### DIFF
--- a/integreat_cms/release_notes/current/unreleased/3961.yml
+++ b/integreat_cms/release_notes/current/unreleased/3961.yml
@@ -1,0 +1,2 @@
+en: Fix slug being overwritten when editing page title
+de: Behebe das Überschreiben des Slugs beim Bearbeiten des Seitentitels
--- a/integreat_cms/static/src/js/forms/update-permalink.ts
+++ b/integreat_cms/static/src/js/forms/update-permalink.ts
@@ -41,13 +41,20 @@ window.addEventListener("load", () => {
     };
 
     document.querySelectorAll("#id_title").forEach((item) => {
+        /* Store original title to detect actual changes */
+        let originalTitle: string;
+
+        item.addEventListener("focusin", ({ target }) => {
+            originalTitle = (target as HTMLInputElement).value;
+        });
+
         item.addEventListener("focusout", ({ target }) => {
-            // Only auto-generate slug if the field is empty
-            if (slugField?.value) {
+            const currentTitle = (target as HTMLInputElement).value;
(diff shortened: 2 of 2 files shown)


timobrembeck (approved):
🎉 Amazing work, Daniel! 🎉

This is such a thoughtful and solid fix! By ensuring that the slug is only generated when the slug field is empty, you’ve stopped the title from accidentally overriding manually set slugs — meaning custom URLs now stay exactly the way editors intend them. This keeps our content consistent and avoids surprising changes when someone tweaks a title 🤩 (great catch!).

The logic here is clean, precise, and just what we needed: auto-generation where appropriate, and respect for intentional configuration where it matters. 🛠️

It’s also just awesome to see you diving into “vibe coding” — bringing that energetic, intuitive programmer mindset we all secretly wish we had. It’s clearly paying off: elegant fixes that feel just right and improve the developer experience for the whole team. 🚀

Thanks again for this contribution — please keep up the fantastic work! The team and the project benefit tremendously from your leadership and hands-on problem solving. 💪✨

svenseeberg:
AFAICT we discussed this somewhere before: I think it is useful to update the slug if the title changes. URLs that fit the content do improve search results. It is relatively likely, that if  the title changes, the author thinks that the old title does not represent the content of the page that well. In such cases we should ensure that the slug is also being updated.

PR #4085, opened 2026-01-17
Fix duplicate_pages slugify to preserve unicode characters

dkehne:
## Problem 

Flaky tests are caused because random input causes intermittent failures and duplicate_pages doesn't handle non-ASCII languages properly.

So here's what's happening:

Normal slug creation: Uses slugify(text, allow_unicode=True) → Arabic characters preserved
duplicate_pages.py: Uses slugify(new_title) with default allow_unicode=False → Arabic characters stripped
This means if you duplicate a page with Arabic content, the duplicated page's slug would lose the Arabic characters while the original keeps them.

There's also a secondary issue on line 50:
rand_str = "".join(random.choices(string.printable, k=5))

string.printable includes problematic characters like whitespace, newlines, and tabs, which could create invalid or unpredictable slugs when passed through slugify.

So the bug is specifically in duplicate_pages.py - it doesn't respect the allow_unicode=True setting that the rest of the codebase uses. The flaky test issue would stem from string.printable containing characters that slugify handles inconsistently.

## Root Cause
  object_instance = <PageTranslation (id: None, page_id: 51, language: ar, slug: )>
  slug = ''

## Summarized 

Arabic translation (language: ar) is being duplicated
slugify(new_title) strips all Arabic characters (non-ASCII)
The slug becomes empty (slug = '')
generate_unique_slug tries to use fallback but cleaned_data is empty → ValidationError

## Solution
- Fixed `slugify()` call to use `allow_unicode=True`, preserving Arabic and other unicode characters when duplicating pages
- Changed random string generation from `string.printable` to `string.ascii_lowercase + string.digits` to avoid problematic characters (whitespace, newlines, tabs) that could cause inconsistent slugs

## Test plan
- [ ] Duplicate a page with Arabic content and verify the slug preserves Arabic characters
- [ ] Verify duplicated pages have consistent, URL-safe slugs

### DIFF
--- a/integreat_cms/core/management/commands/duplicate_pages.py
+++ b/integreat_cms/core/management/commands/duplicate_pages.py
@@ -2,12 +2,9 @@ from __future__ import annotations
 
 import copy
 import logging
-import random
-import string
 from typing import TYPE_CHECKING
 
 from django.core.management.base import CommandError
-from django.utils.text import slugify
 
 from ....cms.models import Page, Region
 from ..debug_command import DebugCommand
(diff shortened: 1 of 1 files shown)


hannaseithe:
In a PR, that is not based on any issue (as there is no issue linked to ). I would expect a description of what the actual bug was / why this needed to be fixed.

hannaseithe:
I don't know what the use cases for the duplicate_pages command are. It seems its part of the asynchronous region cloning, which is called on .save in the `RegionForm`. I don't know who uses that and how often that is used and for whom. Maybe when someone creates a region based on another region?

PeterNerlich (approved):

PR #4086, opened 2026-01-17
Validate event dates within ISO 8601 range (1583-9999)

### LINKED ISSUE #3999: Prevent invalid dates for events
### Describe the Bug

It currently seems to be possible to create events in the year `0025` which is obviously a typo and breaks the app since it is not valid for our date library.
Years before 1583 are not automatically allowed by ISO8601 either https://en.wikipedia.org/wiki/ISO_8601#cite_ref-19:
> However, years before 1583 (the first full year following the introduction of the Gregorian calendar) are not automatically allowed by the standard. Instead, the standard states that "values in the range [0000] through [1582] shall only be used by mutual agreement of the partners in information interchange".

### Steps to Reproduce

Reproducibility: Always

1. Go to Events
2. Create new Event
3. Select Long-Term Event
4. Set Start date before 1583 and End Date in the future
5. Set a Title for the Event
6. Publish
7. Open the Webapp page
8. See error message

### Expected Behavior

Dates that are not valid ISO8601 are not allowed for start and end dates of events.

### Actual Behavior

Invalid dates are allowed leading to completely disabling the Events Tab inside the Web app. The new event cannot be opened and no other events either.

### Screenshot

Image

Image

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

dkehne:
## Summary
- Add date range validation to event forms enforcing ISO 8601 compliant dates (1583-9999)
- Apply validators to `start_date`, `end_date`, and `recurrence_end_date` fields
- Include both server-side (Django validators) and client-side (HTML min/max attributes) validation

## Test plan
- [ ] Create an event with start date 1583-01-01 (should succeed)
- [ ] Create an event with start date 1582-12-31 (should fail with validation error)
- [ ] Create an event with end date 9999-12-31 (should succeed)
- [ ] Create a recurring event with recurrence end date before 1583 (should fail)
- [ ] Verify HTML date pickers show min/max constraints in the browser

Fixes #3999

### DIFF
--- a/integreat_cms/cms/forms/events/event_form.py
+++ b/integreat_cms/cms/forms/events/event_form.py
@@ -2,10 +2,11 @@ from __future__ import annotations
 
 import logging
 import zoneinfo
-from datetime import datetime, time
+from datetime import date, datetime, time
 from typing import TYPE_CHECKING
 
 from django import forms
+from django.core.validators import MinValueValidator
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 
@@ -17,6 +18,9 @@ from ..icon_widget import IconWidget
 
 if TYPE_CHECKING:
     from typing import Any
+
+#: Minimum date for ISO 8601 compliance (start of Gregorian calendar)
+MIN_DATE = date(1583, 1, 1)
 
 logger = logging.getLogger(__name__)
 
@@ -46,6 +50,7 @@ class EventForm(CustomModelForm):
     start_date = forms.DateField(
         label=_("start date"),
         widget=forms.DateInput(format="%Y-%m-%d", attrs={"type": "date"}),
+        validators=[MinValueValidator(MIN_DATE)],
     )
     end_date = forms.DateField(
         required=False,
--- a/tests/cms/views/events/test_event_form.py
+++ b/tests/cms/views/events/test_event_form.py
@@ -424,6 +424,75 @@ def test_no_orpahned_recurrence_rule(
         assert response.status_code == 403
 
 
+# ISO 8601 date range validation test parameters for start_date
+# (<start_date>, <end_date>, <should_succeed>, <error_field>)
+date_range_validation_parameters = [
+    ("1583-01-01", "2583-01-01", True, None),  # Minimum valid date
+    ("2030-01-01", "2030-12-31", True, None),  # Normal valid dates
+    ("1582-12-31", "2030-01-01", False, "start_date"),  # Below minimum start date
+]
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize("parameter", date_range_validation_parameters)
+def test_event_start_date_range_validation(
+    load_test_data: None,
+    login_role_user: tuple[Client, str],
+    settings: SettingsWrapper,
+    caplog: LogCaptureFixture,
+    parameter: tuple[str, str, bool, str | None],
+) -> None:
+    """
+    Test that event start dates are validated to be within ISO 8601 range (>= 1583)
+    """
+    client, role = login_role_user
+    start_date, end_date, should_succeed, error_field = parameter
+    settings.LANGUAGE_CODE = "en"
+
+    new_event = reverse(
+        "new_event",
+        kwargs={
+            "region_slug": REGION_SLUG,
+            "language_slug": "de",
+        },
+    )
+    data = {
+        "title": "Date Range Test Event",
+        "content": "Lorem ipsum...",
+        "start_date": start_date,
+        "end_date": end_date,
+        "is_all_day": True,
+        "is_long_term": True,
+        "status": status.PUBLIC,
+        "has_not_location": True,
+    }
+    response = client.post(
+        new_event,
+        data=data,
+    )
+
+    if role in PRIV_STAFF_ROLES + WRITE_ROLES:
+        if should_succeed:
+            assert response.status_code == 302
+            assert_message_in_log(
+                'SUCCESS  Event "Date Range Test Event" was successfully created',
+                caplog,
+            )
+        else:
+            assert response.status_code == 200
+            # Check that validation error was shown
+            content = response.content.decode("utf-8")
+            assert "Ensure this value is greater than or equal to 1583-01-01" in content
+
+    elif role == ANONYMOUS:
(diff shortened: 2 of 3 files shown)


integreat_cms/cms/forms/events/event_form.py
```diff
@@ -45,12 +51,20 @@ class EventForm(CustomModelForm):
     # These Fields will be used for form the start and date fields of the event model
     start_date = forms.DateField(
         label=_("start date"),
-        widget=forms.DateInput(format="%Y-%m-%d", attrs={"type": "date"}),
+        widget=forms.DateInput(
+            format="%Y-%m-%d",
+            attrs={"type": "date", "min": "1583-01-01", "max": "9999-12-31"},
```
hannaseithe:
```suggestion
attrs={
    "type": "date",
    "min": MIN_DATE.strftime("%Y-%m-%d"),
    "max": MAX_DATE.strftime("%Y-%m-%d")
}
```

Just a minor suggestion

tests/cms/views/events/test_event_form.py
```diff
@@ -424,6 +424,77 @@ def test_no_orpahned_recurrence_rule(
...
 
 
+# ISO 8601 date range validation test parameters
+# (<start_date>, <end_date>, <should_succeed>, <error_field>)
+date_range_validation_parameters = [
+    ("1583-01-01", "1583-01-01", True, None),  # Minimum valid date
+    ("9999-12-31", "9999-12-31", True, None),  # Maximum valid date
+    ("2030-01-01", "2030-12-31", True, None),  # Normal valid dates
+    ("1582-12-31", "2030-01-01", False, "start_date"),  # Below minimum start date
+    ("2030-01-01", "1582-12-31", False, "end_date"),  # Below minimum end date
```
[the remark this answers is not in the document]
dkehne: Good point - removed this redundant test case.

integreat_cms/cms/forms/events/event_form.py
```diff
@@ -2,13 +2,19 @@
...
 
 from django import forms
+from django.core.validators import MaxValueValidator, MinValueValidator
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 
+#: Minimum date for ISO 8601 compliance (start of Gregorian calendar)
+MIN_DATE = date(1583, 1, 1)
+#: Maximum date for ISO 8601 compliance
+MAX_DATE = date(9999, 12, 31)
```
PeterNerlich: You create these constants in the event form and import them from here in the recurrence rule form. It would be cleaner to create them in a utils file. You could place them in a new utils file like `integreat_cms/cms/forms/events/utils.py` or `integreat_cms/cms/utils/event_utils.py`, but here I think it makes sense to define them in `integreat_cms/cms/forms/events/__init__.py` and import it like `from . import MIN_DATE, MAX_DATE` in both forms.

tests/cms/views/events/test_event_form.py
```diff
@@ -424,6 +424,149 @@ def test_no_orpahned_recurrence_rule(
...
+            assert response.status_code == 302
+            assert_message_in_log(
+                'SUCCESS  Event "Date Range Test Event" was successfully created',
+                caplog,
+            )
+        else:
+            assert response.status_code == 200
+            # Check that validation error was shown
+            content = response.content.decode("utf-8")
+            assert "1583" in content or "Ensure this value is greater than or equal to" in content
```
PeterNerlich: Sorry, but I feel like checking for four successive numeric characters anywhere in the content doesn't make for a particularly trustworthy test condition for the kind of HTML documents we are generating here. If this is trying to refer to an error message, it would be better to check for the whole message

tests/cms/views/events/test_event_form.py
```diff
@@ -424,6 +424,149 @@ def test_no_orpahned_recurrence_rule(
...
+            assert response.status_code == 302
+            assert_message_in_log(
+                'SUCCESS  Event "Recurrence Date Range Test Event" was successfully created',
+                caplog,
+            )
+        else:
+            assert response.status_code == 200
+            # Check that validation error was shown
+            content = response.content.decode("utf-8")
+            assert "1583" in content or "Ensure this value is greater than or equal to" in content
```
PeterNerlich: See comment for line 486

integreat_cms/cms/forms/events/recurrence_rule_form.py
```diff
@@ -22,6 +24,19 @@ class RecurrenceRuleForm(CustomModelForm):
...
+            format="%Y-%m-%d",
+            attrs={
+                "type": "date",
+                "min": MIN_DATE.strftime("%Y-%m-%d"),
+                "max": MAX_DATE.strftime("%Y-%m-%d"),
+            },
+        ),
+        validators=[MinValueValidator(MIN_DATE), MaxValueValidator(MAX_DATE)],
+    )
+
```
hannaseithe: Here applies the same comment as above in regards to the end_date

tests/cms/views/events/test_event_form.py
```diff
@@ -424,6 +424,149 @@ def test_no_orpahned_recurrence_rule(
         assert response.status_code == 403
 
 
+# ISO 8601 date range validation test parameters
```
hannaseithe: Depending on whether we keep the two end_date validations or not, the tests need to be adjusted

hannaseithe (changes requested):
Validations looks ok to me now. There is on deletion there though, that should not be

PeterNerlich (approved):
Quick note:
> - [ ] Verify HTML date pickers show min/max constraints in the browser

In firefox I can scroll past December 9999 and submit the form, the date is caught by the server side validation only. Same for before January 1583. I don't see that as such a big problem, but since you mention it in the test section of your description I wanted to draw attention to it.

---

Before merging, please manually squash your commits into a single one. Alternatively few commits encapsulating logical sub parts of the change are also accepted, but this issues feels like just one single logical unit.

I'm already approving this because I have no 'veto' at this point, but please read and make a decision about the above and the note in `test_event_form.py`

dkehne:
@PeterNerlich or @hannaseithe - as is don't have the permission to merge PRs, could someone of you take over here?

PR #4087, opened 2026-01-19
Connect MT to Push Notification

MizukiTemma:
### Short description

This PR makes it possible for users to translate push notifications by machine translation, like in page, event, or POI.

### Proposed changes

- Add Push Notification to MT setting of region (No one, only manager, everyone)
- Use indivisual form like page, poi, event instead of formset, to use the existing code for MT and to avoid conflict between manually edited translation and MT generated version (example: user has manually created/updated the English translation but "English" is checked for MT in the "Implications on translations")
- Adopt `MaschineTranslationForm` partially into `PushNotificationTranslationForm` to use the existing MT process (could not simply inherit, due to some parts like `minor_edit` or `slug`, which do not apply for PN)
- Adjust `deepl_api_client` and `google_translate_api_client` to adopt for push notification translations
- Outsource "Setting" box into a separate file for better readability

### Side effects

- `PushNotificationTranslationForm`, `deepl_api_client` and `google_translate_api_client` need to be refactored later. I would like to keep influence on other components as minimus as possible in thie PR and suggest to refactor in a later PR, as this PR is blocking the release.

Other points to be refactored
- "pushnotification" (ex. permission `change_pushnotification`, `foreign_field` of `PushNotificationTranslation`) vs "push_notification (ex: a field in `PushNotificationTranslation`): 
- "text" in push notification vs "content" in page/event/poi

- No checkbox for "Geringfügige Änderung", as this does not exist in push notifications
- Do not show "Implications on translations" if the current language has no child language node to prevent empty box
- Only the German tab appears in a new PN: other languages will appear after the first save

### Resolved issues

Fixes: #2640 


MizukiTemma:
@dkehne 

> * In push_notification_form.html (lines ~881-899), there's an SEO section that references push_notification_translation_form.meta_description. But PushNotificationTranslationForm doesn't include a meta_description field (fields are only ["title", "text", "automatic_translation"]), and the PushNotificationTranslation model doesn't have this field either

🙈 Legacy of copy&paste. Removed.

> * do_not_translate_title field is added to the model and form, but I don't see where it's actually checked in the translation logic (maybe i oversee it).

It was already taken into account in the existing code. I didn't have to add anywhere newly.
https://github.com/digitalfabrik/integreat-cms/blob/b6319b35b837fd55aa324df227f3034fe61fb6a4/integreat_cms/core/utils/machine_translation_api_client.py#L255-L278


> * In push_notification_translation.py:541-546, the docstring mentions "poi":

Is `integreat_cms/cms/models/push_notifications/push_notification_translation.py` the file meant? Then there is no line 541 - 546. But the docstring of `foreign_field` mentioned to "poi". Fixed to "push notification".

> * The test test_validate_forms_with_no_german_title was removed. Validation that required translations exist for each region's default language moved from form validation to send_pn(). This means that users can maybe save incomplete push notifications, but will get an error when trying to send

Yes, this is the intended behaviour in the current implementation of this PR. We used formset earlier so it was possible to save multiple translations at once (User edits the German translation, then the English one, and press "Save" button at the end to save both the translations). Now we do not use formset anymore to make use of the existing MT codes, so users has to save translations from a language to another (edit German -> save German -> edit English -> Save English).

MizukiTemma:
> Multiple places use string comparison to check the model type. If someone changes the model's verbose_name, this will silently break.A safer approach would be to check the model class directly, e.g., isinstance(content_object, PushNotification) or check content_object._meta.model_name == "pushnotification"


Applied!

Issue #4051, opened 2025-12-02
Incossistent page accesses between CMS and Matomo


jarlhengstmengel:
### Describe the Bug

The numbers of page accesses differ between what is shown in the CMS and Matomo

### Steps to Reproduce

1. Go to 'Statistics' in 'Augsburg'
2. Select a time period before 10.11.2025 (preferably one day)
3. Select 'German' in the legend
4. See Accesses
5. Then log into statistics.integreat-app.de
6. Select 'Augsburg' and previously choosen day
7. Go to 'Behaviour' and 'Pages'
8. Type in the slug of a page in the search bar
9. See that the page views differ from the accesses in the CMS

### Expected Behavior

Page views in Matomo and page accesses in the CMS should be the same.

### Actual Behavior

Page views in Matomo and page accesses in the CMS differ.

### Additional Information

Summery of my investigation: Page accesses are fetched from matomo and saved in our database daily. We request data from Matomo for all PageTranslation objects with their complete url. For parent pages the following seems to happen. Matomo returns  a summary of the views of all child pages and itself. This is saved to our database.

<details><summary>Traceback</summary>

```
```

</details>

### Related Issues

#1097 

### Summary of discussion and updates to the description

When a page is expanded to show its children, the accesses of the children should be subractet from the parent page to show only the accesses on the parent page itself. When a page is collapsed, the accesses of the children should be added again.
Example: collapsed: parent: 10 expanded: parent: 2, child 1: 5, child2: 3

jarlhengstmengel:
Ok, thank you! Then lets stick with the dynamic option: show sum when collapsed and when expanded only actual accesses for the parent :)

jarlhengstmengel:
@osmers #4160 and #4161 are still being worked on, so inconsistencies are still expected

osmers:
@jarlhengstmengel alright, so we need to wait for those as well as the new bug that @svenseeberg and I found to be able to test this completely, right?

jarlhengstmengel:
From what I can see, solving #4161 would make the page based statistics indipendent from that bug

PR #4088, opened 2026-01-19
Fix inconsistent page accesses

jarlhengstmengel:
### Short description

This PR fixes the inconsistencies between page based accesses, total access statistics and matomo itself.

### Proposed changes

- Fix date range error
- Add dynamic recalculation, depending on if a page is in the page tree is expanded by it's children or collapsed. When collapsed, it's children accesses and it's own are added up, otherwise subtracted. When expanded, accesses for a page are consistent with matomo
- Add list of page IDs that are requested from the database as parameter to the backend request
- Add field to the by backend returned dict with the total accesses of page over all languages

### Side effects

- Should be none

The following requirements from the meta issue #1097 are implemented with this fix:
- If a (sub-)tree is collapsed, it will show the sum of all sub pages and its own page access statistics.
- If a tree is expanded, the parent node no longer shows the sum but only its own direct access statistics.

### How to test

It's probably best to test this with a database dump.

To test this locally you need statistics activated:
- Run the fetch page accesses management command: integreat-cms-cli fetch_page_accesses --start-date START_DATE --end--date END_DATE --region-slug augsburg (start and end date can be the same, good pick would be yesterday)
- Wait for about 30 to 40 sec

When you test look for the following:
- compare if for a page the accesses in total with accesses per language is consistent
- check that total and and accesses per language change accordingly when expanding and collapsing subpages
- compare numbers of expanded pages with matomo directly
- Make sure that the page tree in the page view is still intact

How to access matomo and check statistics there:
- Go to [matomo](https://statistics.integreat-app.de) and login
- Select "Augsburg" and the date you want to check
- Navigate to "Behaviour" and select "Pages"
- Search for a page with the page slug (not ID)
- You will see all pages that contain the slug you searched for with accesses of the selected date
- Keep in mind that we translate page slugs

### Resolved issues

Fixes: #4051 

### DIFF
--- a/integreat_cms/cms/views/statistics/statistics_actions.py
+++ b/integreat_cms/cms/views/statistics/statistics_actions.py
@@ -141,6 +141,7 @@ def get_page_accesses_ajax(request: HttpRequest, region_slug: str) -> JsonRespon
     region = request.region
     data = request.POST
     language_slugs = data.getlist("language_slugs")
+    page_ids = data.getlist("page_ids")
 
     if not region.statistics_enabled:
         logger.error("Statistics are not enabled for this region.")
@@ -157,7 +158,7 @@ def get_page_accesses_ajax(request: HttpRequest, region_slug: str) -> JsonRespon
             status=400,
         )
 
-    region_pages = Page.objects.filter(region=region)
+    region_pages = Page.objects.filter(region=region, id__in=page_ids)
 
     page_access_sums = region.get_page_access_count_by_language(
         pages=region_pages,
@@ -171,6 +172,8 @@ def get_page_accesses_ajax(request: HttpRequest, region_slug: str) -> JsonRespon
         page_id = str(access_sum["page__id"])
         language_slug = str(access_sum["language__slug"])
         page_accesses_dict[page_id][language_slug] = access_sum["total_accesses"]
+        page_accesses_dict[page_id].setdefault("total_accesses", 0)
+        page_accesses_dict[page_id]["total_accesses"] += access_sum["total_accesses"]
 
     return JsonResponse(page_accesses_dict, safe=False)
 
--- a/integreat_cms/static/src/js/analytics/statistics-page-accesses.ts
+++ b/integreat_cms/static/src/js/analytics/statistics-page-accesses.ts
@@ -42,6 +42,17 @@ const resetTotalAccessesField = (accessFields: HTMLCollectionOf<Element>, isEmpt
     }
 };
 
+const updateAllAccessesField = (accessesField: Element, allAccesses: number) => {
+    const allAccessesField = accessesField;
+    if (allAccesses === 0) {
+        allAccessesField.textContent = `${allAccessesField.getAttribute("data-translation-no-accesses")}`;
+    } else if (allAccesses === 1) {
+        allAccessesField.textContent = `${allAccesses} ${allAccessesField.getAttribute("data-translation-singular")}`;
+    } else {
+        allAccessesField.textContent = `${allAccesses} ${allAccessesField.getAttribute("data-translation-plural")}`;
+    }
+};
+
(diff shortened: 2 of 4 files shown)


dkehne (approved):
Reviewed - looks good! ✅

PeterNerlich (changes requested):
I spent an embarrassingly long time finding out that I had to select languages because I got stuck at "wait why am I getting no data" instead of proceeding with the test where I would have added languages anyway :see_no_evil: And that even though I knew and discussed this behaviour once before

I did notice problems that I think must lie with the script importing the data from matomo. Some data points like for `augsburg-und-sein-besonderes-wasser-ein-unesco-weltkulturerbe` or `contact-app-team-augsburg` on `2026-01-27` are not even in our database, while matomo reports 1 and 2 hits, respectively, even after multiple runs of the script fetching an interval involving the surrounding dates as well.

In the end, I settled on deleting any PageAccesses altogether and just produce a very boring but predictable baseline of giving every page in every language exactly 1 view. 
<details><summary>Since that made checking for errors so much easier, <b>I'm recommending this as the default way of checking for page access display stuff from now on</b></summary>

```python
from integreat_cms.cms.models import *

pages = Page.objects.filter(id__in=(1,2, […] ))
languages = Language.objects.filter(slug__in=["de", "en", "ar"])

for page in pages:
    for lang in languages:
        PageAccesses(language=lang, page=page, access_date='2026-01-27', accesses=1).
save()
```

</details>

With this, it seemed to work fine at first, as I expanded the whole tree and then added languages to display. However, if some subtrees are collapsed and languages are added, the parent page will show not the total views for its subtree, but only its individual views. What's worse, if then expanded, it will suddenly show negative values:

image


I really think we should refactor this to a more separated approach with in memory data and a mostly decoupled presentation layer instead of setting the data as attributes to the DOM nodes representing the individual data points themselves at some point. I feel like this is growing less readable.
But for now just another bandaid is maybe the way to go

integreat_cms/static/src/js/analytics/statistics-page-accesses.ts
```diff
@@ -122,6 +129,43 @@ const updateDOM = (data: AjaxResponse, visibleDatasetSlugs: string[]) => {
     });
```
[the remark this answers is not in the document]
jarlhengstmengel: It basicly does, it processes all pages and the accesses per language. The Page tree initializes with everything but the root pages hidden, but the they're all there. I think the main struggle is, that we get the accesses for parent pages summed up with their children from the backend, which is basicly the state the page tree starts in. Or do you mean not triggering `calculateToggleSubpages` in `toggle-subpages.ts` but somewhere here?

integreat_cms/cms/views/statistics/statistics_actions.py
```diff
@@ -171,6 +172,13 @@ def get_page_accesses_ajax(request: HttpRequest, region_slug: str) -> JsonRespon
         page_id = str(access_sum["page__id"])
         language_slug = str(access_sum["language__slug"])
         page_accesses_dict[page_id][language_slug] = access_sum["total_accesses"]
+        if "total_accesses" not in page_accesses_dict[page_id]:
+            page_accesses_dict[page_id]["total_accesses"] = access_sum["total_accesses"]
+        else:
+            page_accesses_dict[page_id]["total_accesses"] = (
+                page_accesses_dict[page_id]["total_accesses"]
+                + access_sum["total_accesses"]
+            )
```
jonbulz: Could be simplified to:
```
page_accesses_dict[page_id].setdefault("total_accesses", 0)
page_accesses_dict[page_id]["total_accesses"] += access_sum["total_accesses"]
```

jonbulz (changes requested):

integreat_cms/cms/views/statistics/statistics_actions.py
jarlhengstmengel: Applied

PR #4092, opened 2026-01-20
Allow updating news after scheduled push notification was sent

### LINKED ISSUE #3373: Push notifications sent via schedule cannot be changed afterwards
### Describe the Bug
A push notification that has been sent previously using the schedule function, cannot be changed afterwards even though the text above and known behaviour from push notifications sent manually indicated that it should work. 

Image

### Steps to Reproduce

1-6 can be skipped if scheduled and already sent news are available.

1. Go to 'News'
2. Click 'Create News'
3. Enter something in Title and Content
4. Click 'Schedule News' and select current day, current time
5. Save & Send
6. Wait for news to be sent out
7. Go to 'News' and open the news which was scheduled and sent out
8. Change something in the content
9. Click on 'Update'
10. See error

### Expected Behavior
The changes should be taken into account and the text displayed in the app should be changed accordingly.

### Actual Behavior
Following error message is shown: 

Image

### Summary of discussion and updates to the description
Update is refused by this code block:
https://github.com/digitalfabrik/integreat-cms/blob/b6319b35b837fd55aa324df227f3034fe61fb6a4/integreat_cms/cms/forms/push_notifications/push_notification_form.py#L118-L127

- Re-sending (dispatching a notification on user devices) must be stay stopped as it is
- Updating of a push notification must be allowed

dkehne:
## Summary
- Fix for issue where news objects became uneditable after a scheduled push notification was sent
- The push notification itself (already sent to devices) remains unchanged
- Only the news content displayed in the app can now be updated

## Root Cause
When updating an already-sent scheduled notification, the disabled `schedule_send` field still returned `True` from its initial value. The validation saw this and blocked the entire form submission, even though the user was just updating content.

## Solution
Added a check to skip the "already sent" validation when the `schedule_send` field is disabled, since:
- The user can't change a disabled field anyway
- They're just updating the news content
- Re-sending is still prevented (scheduling fields stay disabled)

## Test plan
- [ ] Create a news with a scheduled push notification (set to current time)
- [ ] Wait for the notification to be sent
- [ ] Try to edit the news content
- [ ] Verify the update succeeds and content is changed in the app

Fixes #3373

### DIFF
--- a/integreat_cms/cms/forms/push_notifications/push_notification_form.py
+++ b/integreat_cms/cms/forms/push_notifications/push_notification_form.py
@@ -8,8 +8,6 @@ from django import forms
 from django.conf import settings
 from django.forms import CheckboxSelectMultiple
 from django.utils import timezone
-from django.utils.formats import localize
-from django.utils.timezone import localtime
 from django.utils.translation import gettext_lazy as _
 
 from ...constants import region_status
@@ -117,19 +115,13 @@ class PushNotificationForm(CustomModelForm):
                 ),
             )
 
-        if cleaned_data.get("schedule_send") and self.instance.sent_date:
-            self.add_error(
-                "schedule_send",
-                forms.ValidationError(
-                    _('News "{}" was already sent on {}').format(
-                        self.instance,
-                        localize(localtime(self.instance.sent_date)),
-                    ),
-                ),
-            )
-
         # Combine the scheduled send day and time into one timezone aware field
-        if not self.errors and cleaned_data.get("schedule_send"):
+        # Skip when notification was already sent
+        if (
+            not self.errors
+            and cleaned_data.get("schedule_send")
+            and not self.instance.sent_date
+        ):
             tzinfo = zoneinfo.ZoneInfo(str(timezone.get_current_timezone()))
             time = cleaned_data["scheduled_send_date_time"] or datetime.time()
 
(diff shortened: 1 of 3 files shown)


dkehne:
my local database but it will be fixed the next days...

hannaseithe:
I am having an interesting problem/bug? 

After I sent the scheduled push notification with the management command, it shows in the list overview as sent, but when I open the notification its not registered as sent (fields not disabled etc.) - no matter how I force a reload.

Then after I stop the server and run it again, and when I open the pushnotification then it finally is registered as sent.

So I dont think that is caused by your issue. But someone who is more knowledgable about push notifications (maybe @MizukiTemma ) should have a look at this.

See the screenshots here:
#1 and #2 one is after succesful sent with management command
#3 after restarting the server
Bildschirmfoto vom 2026-01-28 17-03-38
Bildschirmfoto vom 2026-01-28 17-04-05
Bildschirmfoto vom 2026-01-28 17-05-49

integreat_cms/cms/forms/push_notifications/push_notification_form.py
```diff
@@ -115,19 +113,13 @@ def clean(self) -> dict[str, Any]:
...
-                    _('News "{}" was already sent on {}').format(
-                        self.instance,
-                        localize(localtime(self.instance.sent_date)),
-                    ),
-                ),
-            )
-
         # Combine the scheduled send day and time into one timezone aware field
-        if not self.errors and cleaned_data.get("schedule_send"):
+        # Skip when fields are disabled (notification already sent)
```
hannaseithe: The comment should be adjusted

hannaseithe (changes requested):
First of all: Thank you very much :)

See the comment below, and in addition the translations need to be fixed, since we now don't need the Message for the Validation anymore. If the AI struggles to change the translations, I can also do that for you.

In addition to all that: The Bug/Problem I commented in the PR should be also be confirmed, that this is really not caused by your changes.

Issue #4093, opened 2026-01-21
During poi creation related contact fails silently


hannaseithe:
### Describe the Bug

During poi creation, if we create a related contact at the same time and enter an invalid data to the phone number, and save, then the contact is simply not created without any error messages given

### Steps to Reproduce

1. Go to 'Locations'
2. Click 'Create location'
3. Enter mandatory location info (title, address, category)
4. In 'related contacts' click 'Create new contact'
5. enter anything in 'Area of responsibility', 'Name', and enter an invalid phone number / invalid mobile phone number
6. Click 'Create'

The location is created, however not the contact and there is no error message why it didn't create the contact.

### Expected Behavior

If any of the related contact fields are invalid, neither the POI nor the contact should be saved, we get an error message and the already entered data is preserved

### Actual Behavior

If any of the related contact fields are invalid, the POI is created, the contact isnt, the contact form is closed and we get no error messages

### Additional Information

This bug was discovered during testing here: https://github.com/digitalfabrik/integreat-cms/pull/4063#issuecomment-3772348084

### Related Issues

#4063

PR #4099, opened 2026-01-22
Fix related contact creation bug

hannaseithe:
### Short description
A fix for the bug that when "a poi is created with a related contact and the contact data is invalid", that the poi is created and the contact not, and that there are no error messages

### Proposed changes

- wrap the whole saving part of the post method of the `poi_form_view.py` in a transaction.atomic(): context wrapper and set a savepoint that gets rolled back, if the contact_form was invalid
- create error messages for contact_form
- some parts of the post method were extracted into submethods
- introducing two new arguments for the templates to make distinction between edit mode and creation mode as well as whether the contact_widget should be shown explicit

### Side effects

- None

### How to test

- test that poi creation works for the following conditions:
  - a valid poi without a related contact
  - an invalid poi without a related contact (make the poi invalid by deleting the `required` attribute of the title in chrome devtools and leaving the title field empty
  - a valid poi with a valid related contact (only fill out the name of the related contact)
  - an invalid poi with a valid related contact
  - a valid poi with a invalid related contact (fill out the name and fill out the phone number with letters)
  - a invalid poi with a invalid related contact

Make sure that everything is rendered correctly after clicking save, that the correct error messages are there, and also make sure that if anything is invalid neither the poi nor the contact were saved

### Resolved issues

Fixes: #4093 

### DIFF
--- a/integreat_cms/cms/views/pois/poi_form_view.py
+++ b/integreat_cms/cms/views/pois/poi_form_view.py
@@ -244,10 +244,9 @@ class POIFormView(
                 if new_poi_and_new_related_contact:
                     # Flush opening_hours which is meant for the location
                     # The contact can also have opening hours itself, but here we want it to always mirror those of the location
-                    success = self.create_related_contact(
+                    success, contact_form = self.create_related_contact(
                         data=data,
                         request=request,
-                        contact_form=contact_form,
                         poi=poi,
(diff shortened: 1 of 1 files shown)


dkehne (approved):
Reviewed - looks good! ✅

Issue #4098, opened 2026-01-21
Refactor suggestions to use trigram similarity


PeterNerlich:
Refactor suggestions to use trigram similarity
(see https://github.com/digitalfabrik/integreat-cms/pull/3832 – not sure how much this will complicate stuff on the completion phrases calculation side of things)

Blocked by: #4095

PR #4105, opened 2026-01-26
refactor search suggestions

### LINKED ISSUE #4095: Refactor suggestions to general implementation
Refactor suggestions to have **one general implementation** which supports drawing them from **multiple fields** of the model

See parent issue #3992

### LINKED ISSUE #4096: Search suggestions: Process phrases instead of DB objects
### Motivation

In #3963 we refactored how the search works. We want a similar approach to search suggestions, which are confusingly similar for now. The refactoring should also make these suggestions more useful.

Our suggestions *(I'm for renaming this to completions #4097 and will use this term in the rest of this issue)* currently correspond 1:1 to contacts. Instead, the method should find out which matching terms in concrete items are most prevalent, those terms should be ranked by the number of contacts using them and *that* should form the list of completions.

<details><summary>Example</summary>

Search term: `mar`

Assume these matches:
- `Integrationsberatung`: `Maria Musterfrau`
- `Mario`
- `Ina Irgendwas` (`ina@marburg-biedenkopf.de`)
- `Marge Simpson`
- `Alles rund um Marburg Biedenkopf`: `Kenny`

Ranked completions:

score 2:
- `marburg biedenkopf`
score 1:
- `maria`
- `mario`
- `marge`

The completions listed as *score n* form a category and don't have any defined order between themselves.

In *score 2* is `marburg biedenkopf` and not only `marburg`, because in the two matching contacts that word is both times immediately followed by `biedenkopf`. I'm not sure whether I would see it as helpful if `marburg` was also an additional option with *score 2*, as both are valid completions for `mar`.
If there was a third contact matching that would have something else *(or nothing)* after `marburg`, just `marburg` would have *score 3* and thus be ranked above `marburg biedenkopf`.

</details>

I'm not sure how performance heavy a solution exactly like this would be. We might find this to be not feasible in our case and settle on a compromise.

### Proposed Solution

- Complement suggestions by a second step to calculate and rank completion phrases from the objects returned from the database, as matching the phrase to be completed

### Alternatives

- Leave it at #4095 and #4098 

### User Story

As a ***user*** I want ***useful completions to my search terms*** so that ***I don't have to remember how to spell it in order to search for it***.
As a ***user*** I want ***search completions to show me the most likely search phrase first*** so that ***I, on average, need the least amount of clicks and key presses to complete my work***.

### Related Issues

#958, #3445, #3452, #3878, #3923, #3988

jonbulz:
### Short description

Currently, suggesting search term completions in the Search input field and record suggestions when linking to internal references use the same logic. This PR introduces `SearchSuggestMixin` to cleanly separate the logic of search term suggestions from suggesting records. Instead of just producing a list of record names, `suggest_tokens` extracts useful search tokens from the existing records and ranks them by similarity score.

### Proposed changes

- Introduce `SearchSuggestMixin` to extend search term suggestion to different models
- add new endpoint `search/suggest` to separate search term completion from record suggestion
- move `search_content_ajax.py` from `views/utils` to `views/search` for convenience
- add matcher, scorer and tokenizer to extract and rank tokens from existing records
- define `SEARCH_FIELDS` in a central place to handle configuration
- introduce `region_filter_field` and `archived_filter_field` to allow filtering by region and archived state
- add tests

### Side effects

- The `search_content_ajax` endpoint and its `suggest()` model methods are unchanged. List view search inputs now hit the new `search/suggest/` endpoint instead. The "Insert link" editor dialog continues to use `search_content_ajax`.

There are no intended deviations from the issue and design. The old `suggest()` / `search()` methods on individual models are intentionally left in place for the `search_content_ajax` endpoint and will be unified in a follow-up.

### How to test

- Open any list view (pages, events, contacts, etc.), type in the search box, and verify autocomplete suggestions appear
- Open the rich text editor, use "Insert link" → search for internal content, and verify record suggestions still work
- Verify suggestions respect region boundaries (e.g., searching in one region doesn't surface content from another region)
- Verify archived records are excluded from suggestions by default

### Resolved issues

Fixes: #4095 
Fixes: #4096 


hannaseithe (changes requested):
Currently the search seems to not be working. In so far as for some reason after the search returns the search results, the `partial-page-tree` enpoint seems to get called (twice), which eventually simply returns the full page tree. See the screenshot from the log

- your branch:
Image



- develop:
Image

PS: I will wait with the review of the code until this is fixed

jonbulz:
@hannaseithe yes this is actually what was fixed in #4190 . So it's unrelated to the changes here, but the fix was missing from the branch because we merged it into the release candidate first, and not develop. Since it's now on develop, I merged it back into the branch so that you can test properly :)

tests/cms/search/test_suggest.py
```diff
@@ -0,0 +1,177 @@
...
+    @pytest.mark.django_db
+    def test_query_is_case_insensitive(self, load_test_data: None) -> None:
+        """Test that query matching is case insensitive"""
+        result_lower = Contact.suggest_tokens(query="test")
+        result_upper = Contact.suggest_tokens(query="TEST")
+
+        # Both should return the same suggestions
+        assert len(result_lower.get("suggestions", [])) == len(
+            result_upper.get("suggestions", [])
+        )
```
hannaseithe: Testing for length only here seems a bit "weak" to me here. What if for another reason both lists are empty?

integreat_cms/cms/search/scorer.py
```diff
@@ -0,0 +1,24 @@
...
+) -> float:
+    """
+    Compute the score contribution of a single token.
+
+    :param similarity: The trigram similarity score (0.0 to 1.0)
+    :param weight: The field weight multiplier
+    :param prefix_match: Whether the token starts with the query
+    :return: The weighted score for this token
+    """
+    score = similarity * weight
```
hannaseithe, line 21: I have thought about this a little bit: With the the weights being unbounded this could easily become unbalanced (but maybe that is intended?). Maybe it would be better to bound weights also between 0.0 and 1.0.

integreat_cms/cms/models/mixins.py
```diff
@@ -0,0 +1,170 @@
...
+        # Scores are accumulated per token - duplicate suggestions from multiple
+        # fields or objects are automatically merged by summing their scores
+        scores: defaultdict[str, float] = defaultdict(float)
+
+        # Build filter for any field containing the query
+        q_filter = Q()
+        for field in fields:
+            q_filter |= Q(**{f"{field}__icontains": query})
+
+        qs: QuerySet[Any] = cls.objects.filter(q_filter)
```
hannaseithe: I kind of brought that up before at the search PR, and wasnt really happy with things there and I see the same problem popping up here again:

We filter on all translations, not just the latest versions, so we get suggestions from older versions, which then if we select them return nothing in the case of Page, and for the others it returns the newer version (which does not match with our search term)

tests/cms/views/search/test_search_suggest.py
```diff
@@ -0,0 +1,312 @@
...
+    )
+
+    assert response.status_code == 200
+    data = response.json()
+    suggestions = data["data"]["suggestions"]
+    if len(suggestions) > 1:
+        scores = [s["score"] for s in suggestions]
+        assert scores == sorted(scores, reverse=True), (
+            "Suggestions should be sorted by score descending"
+        )
```
hannaseithe: On our test data the length is always 0

hannaseithe (changes requested):
Thank you for the PR. Great work

As this is a very big PR, its hard to check everything in detail. But overall things seem to be very sound and clean to me, which I like. The points I am bringing up in the review, might be things you have already considered and chosen for a good reason, if so I'd appreciate if you could explain me to, so I can understand.

Edit: Thanks for the merge btw, could have thought of this myself :D

tests/cms/search/test_suggest.py
jonbulz: good catch, I assert that the lists are not empty now :)

integreat_cms/cms/search/scorer.py
jonbulz: Well, us devs define the weights. They're also defined in a single place, so I think it's quite easy to see if something is just way out of scope. If you think this is really something that needs to be more firm, I could add type hints. But it seems to me that the risk is rather low and the effort to introduce this is not worth it (I'd need to change the type in every existing function signature). I'd be against a hard validation, I don't think that would be necessary. What do you think? :)

tests/cms/views/search/test_search_suggest.py
jonbulz: another good catch! turns out using such a broad query results in similarity scores that are too low

integreat_cms/cms/models/mixins.py
jonbulz: Let's talk about this. I get your point, and I have a draft for version filtering that I could commit. It's added complexity, but that is IMO not the problem in this case. In my mind, searching the older versions is actually not a bug, but a feature, because users could maybe remember the old (translation) name of a content object, and only remember the new name when the search returns the object. In general, I feel like a search that returns (a little) more results than expected is far more useful than a search that returns too little. But that is arguably an opinion.
What should not happen is that a search term is suggested, but then the search yields no results. Why do you think this is the case for pages, but not the other content models?

MizukiTemma (changes requested):
Thank you so much, that's a great work 😻 and it comes even with clean code, good structure and tests 🚀 

I have though two concerns:

1. In Media library (both network management and region) it seems the completion is not impelmented. If it was not intented, we should add it, as it's currently available on the prod system.

2. It is not a bug but should be informed to Service Team that users may have to type more to see the first suggestion (completion) compared to the current implementation. 

Currently users get suggestions directly after the first letter tipped (if any maches), but after this PR they have to type at least two letters and no completion may be suggested with very two letters. I observed completion appeard first with >3 letters (as tested locally with test data. The performance may be better with real data). 

This makes them maybe confused. Service Team should inform users of this change and also of an option to hit the enter key or click the lupe icon to search for short string.

jonbulz:
> Thank you so much, that's a great work 😻 and it comes even with clean code, good structure and tests 🚀
> 
> I have though two concerns:
> 
>     1. In Media library (both network management and region) it seems the completion is not impelmented. If it was not intented, we should add it, as it's currently available on the prod system.
> 
>     2. It is not a bug but should be informed to Service Team that users may have to type more to see the first suggestion (completion) compared to the current implementation.
> 
> 
> Currently users get suggestions directly after the first letter tipped (if any maches), but after this PR they have to type at least two letters and no completion may be suggested with very two letters. I observed completion appeard first with >3 letters (as tested locally with test data. The performance may be better with real data).
> 
> This makes them maybe confused. Service Team should inform users of this change and also of an option to hit the enter key or click the lupe icon to search for short string.

Thank you very much! I wanted to keep `MediaFile` pointed toward the old endpoint, but that did not work with the new request parameters and response shape I intoduced 🙈 Thanks for finding this bug!

I changed the `DEFAULT_MIN_SIMILARITY` to `0.1` to include more results earlier. Do you think it is better this way? In any case, this should be fairly simple to tweak in case users prefer one option over the other (more results early vs precise results later)

MizukiTemma (approved):
Thank you so much 😸 Looks good 🚀

integreat_cms/cms/models/mixins.py
hannaseithe: This is really just my personal opinion, but I feel searching in older version is not ideal. Overall I feel that is a UX/UI question. Maybe @m0ndra you have an opinion on that?

integreat_cms/cms/search/scorer.py
hannaseithe: Why I would prefer the bounded weights: This way , when adding a new weight, the actual weight of other weights would not move since the maximum is fixed. With unbounded weights, it shifts the relative weigthiness of every other weight if we add a new "max" weight. But this is also not such a strong opinion for me, that if you want to stay with the current form, that I would be unhappy about it

integreat_cms/cms/models/mixins.py
hannaseithe: How to reproduce the page situation: 
1. Create a new page with title "Title first version", save it. 
2. Change the title to "Title second version" save it. 
3. Go to the list view. Type "first version" See suggestion "Title first version" Select that, see that no search results are shown

hannaseithe (approved):
Thank you

integreat_cms/cms/models/mixins.py
jonbulz: @hannaseithe after conversing with @m0ndra we decided that it is less confusing for users if the search and search term suggestions are limited to the latest versions only, as you suggested. However, there is only so much I want to do in this PR, and the current situation of the search functionality is still a bit messy (some models already use the newly-introduced mixin, some still have their own search functions because they are used in a context that expects a different signature). 
Long story short: after the latest two commits, the suggested search terms should be generated only from the latest version, if a model has versioning. The search in some list views works the same (e.g. pages), but not in others. Thus, there are some inconsistencies that will be fixed as the refactoring of the search function goes on. I'll create new issues for that soon.

integreat_cms/cms/models/media/directory.py
```diff
@@ -17,12 +18,34 @@
...
 
 
-class Directory(AbstractBaseModel):
+class Directory(AbstractBaseModel, SearchSuggestMixin):
     """
     Model representing a directory containing documents. This is only a virtual directory and does not necessarily
     exist on the actual file system. Each directory is tied to a region.
     """
 
+    search_fields = {"name": {"weight": 1, "tokenize": False}}
```
hannaseithe: I would expect this to be imported from search_fields.py

integreat_cms/cms/models/media/media_file.py
```diff
@@ -131,13 +132,68 @@ def filter_unused(self) -> MediaFileQuerySet:
...
+        Generate search suggestions from both media files and directories.
+
+        :param query: The search query string
+        :param region: The region to filter by (optional)
+        :param archived: Whether to include archived records (default: False)
+        :param kwargs: Additional arguments (unused, for compatibility)
+        :return: Dict with "suggestions" key containing list of {suggestion, score} dicts
+        """
+        file_result = super().suggest_tokens(query, region=region, archived=archived)
+        dir_result = Directory.suggest_tokens(query, region=region, archived=archived)
```
hannaseithe: I am a bit uncomfortable with how Directory is handled here. This seems like a bit hacky solution. I think the core issue comes from the archtiectural assumption that we only search on one object_type in the view, which forces you to hide the directory away here inside media_file.py Maybe a better way would be to let the view accept a list of object_types and then merge and sort them together at the end - especially since all the fields are not tokenized, so accumulating score over tokesn wont be a cocnern.

integreat_cms/cms/models/media/media_file.py
```diff
@@ -131,13 +132,68 @@ def filter_unused(self) -> MediaFileQuerySet:
...
+    search_fields = MEDIA_FILE_SEARCH_FIELDS
+    region_filter_field = "region"
+    archived_filter_field = None
+
+    @classmethod
+    def get_suggest_queryset(
+        cls,
+        region: Region | None = None,
+        archived: bool = False,  # noqa: ARG003
+    ) -> QuerySet[Any]:
```
hannaseithe: I haven't looked at all the model files - so this might occur elsewhere too, but I stumbled across the ordering here. I expected the model fields to come first and the classmethods later. More like in page.py

integreat_cms/cms/models/pages/page.py
```diff
@@ -256,6 +259,39 @@ def get_translation_model() -> ModelBase:
...
+        )
+        # FilteredRelation creates a JOIN constrained to latest translations only,
+        # which the ``latest_translations__title`` lookups in PAGE_SEARCH_FIELDS
+        # then resolve against — keeping old revisions out of the trigram annotation.
+        return qs.annotate(
+            latest_translations=models.FilteredRelation(
+                "translations",
+                condition=models.Q(translations__id__in=latest_translation_ids),
+            ),
+        )
```
hannaseithe, line 307: I get the error

```
May 18 10:56:45 DEBUG integreat_cms.cms.views.search.search_suggest - Search suggest for 'page' with query 'Ka'
May 18 10:56:46 ERROR django.request - 500 Internal Server Error: /augsburg/ajax/de/search/suggest/
Traceback ....

....

integreat-cms/.venv/lib/python3.11/site-packages/django/db/models/sql/query.py", line 116, in get_child_with_renamed_prefix
    raise ValueError(
ValueError: Passing a QuerySet within a FilteredRelation is not supported.
```

integreat_cms/cms/models/mixins.py
```diff
@@ -0,0 +1,183 @@
...
+
+        matcher = TrigramMatcher()
+        # Scores are accumulated per token - duplicate suggestions from multiple
+        # fields or objects are automatically merged by summing their scores
+        scores: defaultdict[str, float] = defaultdict(float)
+
+        # Build filter for any field containing the query
+        q_filter = Q()
+        for field in fields:
+            q_filter |= Q(**{f"{field}__icontains": query})
```
hannaseithe, line 159: I think there is an issue with the filtering pipeline. 

1. you do an icontains. That filters out any fuzzyness from the start
2. . You apply trigram similarity when we only have exact matches already (so similarity only gets applied by accident to alternative fields - where we already have an exact match on another field of the same object)

Like in this case:
Image

I remember you saying that the fuzzyness is intentional for typos and such. In that case icontains is the wrong prefilter.

hannaseithe (changes requested):
Thank you for implementing the change requests. I had a second a little more thorough look and came up with a few more issues:

integreat_cms/cms/models/mixins.py
jonbulz: The scope of this PR was initially to only use similarity for scoring, not for finding the matches. Using similarity for finding matches was planned to be handled in a separate PR. But since the PR is already huge, it doesn't really hurt to include it here as well, i.e. replace the icontains with similarity matching. If you'd prefer that, I can do it. The corresponding issue is #4098

integreat_cms/cms/models/media/directory.py
jonbulz: That's correct, my bad 😅

integreat_cms/cms/models/media/media_file.py
jonbulz: the issue is that, when a user is searching the media library, we don't know if they are searching for a directory or a media file object. from a technical perspective, the `object_type` is hardcoded to `mediafile` [here](https://github.com/digitalfabrik/integreat-cms/blob/44b811ee0fb358b9cf2bcb008fe7e470b16c5d51/integreat_cms/static/src/js/media-management/library.tsx#L225). we do resolve the `object_type` into a model class in `integreat_cms/cms/views/search/utils.py:get_model_class_from_object_type`, and we could change our logic to accept a list of classes such that the `mediafile` object type resolves to `[MediaFile, Directory]`, _however_, the utils function is currently also used by the legacy `search_content_ajax`, which expects a different return type. so the options if see are:
- keep the (arguably hacky) approach that the `Directory` tokens are merged into the `MediaFile` tokens inside the `MediaFile.suggest_tokens` method
- wrap `get_model_class_from_object_type` for `search_suggest` to handle the `mediafile` special case for search token suggestions only while keeping the logic for the legacy endpoint untouched

which do you think is better?

integreat_cms/cms/models/pages/page.py
jonbulz: seems like this worked in Django 4.2, but is now rejected in newer versions. I'll check for a different solution

integreat_cms/cms/models/media/media_file.py
jonbulz: yes, it seems we mostly follow that convention. I moved the classmethods down to the existing ones

jonbulz:
@MizukiTemma could you also have a look again? quite some things changed since you reviewed 🙈

MizukiTemma:
@jonbulz 
Thank you for continuing working on this topic 💪 The current implementation looks good from my side.

It's not a bug but I found one concern: completions for users are not filtered by the region and users may see names or e-mail addresses of users of the other regions.

Example and how to reproduce locally
There is one user in the region Artland with emal address management_artland[at]example.com in our default data. If you go to the region Augsburg and search for "mana" in the user list, you get "management_artland[at]example.com" in the completions. This can also happen with names. Such chances are generally low (because you have to know at least partially the names/e-mail addresses) but may be breaking a general privacy protection expectation?

integreat_cms/cms/models/media/media_file.py
hannaseithe: My suggestion would actually to be, to change the return value from the frontend:

- so in library.tsx line 225:
```
data-object-type="mediafile,directory"
```
- in search-query.ts line 85:
```
tableSearchInput.getAttribute("data-object-type").split(","),
```
- and search-query.ts from line 3
```
const queryObjects = async (url: string, types: string[], queryString: string, archived: boolean) => {
    if (queryString.trim().length === 0) {
        document.getElementById("table-search-suggestions").classList.add("hidden");
        return;
    }

    const response = await fetch(url, {
        method: "POST",
        headers: {
            "X-CSRFToken": getCsrfToken(),
        },
        body: JSON.stringify({
            query_string: queryString,
            object_types: types,
            archived,
        }),
    });
```
- and then we could just loop over the list in search_suggest.py and therfore be able to split the two suggest_tokens to the two models where they belong.

I hope I didn't miss anything, but this way we would not touch the old ajax endpoint, but still end up with a cleaner architecture. What do you think?

integreat_cms/cms/models/mixins.py
hannaseithe: If this is not too much work for you, I would prefer to have it implemented here. But I leave it up to you to decide.

integreat_cms/cms/models/abstract_content_translation.py
```diff
@@ -502,13 +505,20 @@ def get_suggest_queryset(
...
-        Restrict suggestion sources to the latest translation per (object, language)
-        so that token suggestions reflect what list views actually display, instead
-        of mixing in tokens from outdated revisions.
+        Restrict suggestion sources to the latest *published* translation per
+        (object, language) so that token suggestions reflect what list views
+        actually display, instead of mixing in tokens from outdated revisions or
+        unpublished drafts.
         """
-        qs = super().get_suggest_queryset(region=region, archived=archived)
+        qs = super().get_suggest_queryset(
```
hannaseithe, line 516: This tightly couples the SearchSuggestMixin to the abstract_content_translation class. Maybe it would be better to inherit from the SearchSuggestMixing directly here in the AbstractContentTranslation class instead of the concrete child classes?

hannaseithe (commented):
The language filter and the archived filter make sense to me. I left a minor architectural comment. That might be worth considering.

jonbulz:
converted to draft until this is ready to be picked up again

Issue #4107, opened 2026-01-27
Create Location enters broken Update state after invalid phone number input


juliankehne:
### Describe the Bug

When creating a location and also try to create a contact during creation with an invalid phone number, e.g. "123abc" it will correctly give an error message that the number is invalid and not create the location/contact, however some fields that shouldn't be available in the "Create Location" view become visible and the "Publish" button changes to "Update".

This was noticed after the initial fix: The page shows that there is a version for the location including button to Show it.

Changing the phone number to a valid one afterwards will still create the location and contact (granted the area of responsibility is also entered, see #4050)

### Steps to Reproduce

1. Go to 'Location'
2. Click on 'Create Location'
3. Enter a 'Title'
4. Select any category for the location
5. In "Position" select any location on the map inside Germany
6. Select "Update address and coordinates" 
7. Click "Confirm"
8. In ' Related contacts ' click 'Create a new contact'
9. Type an invalid phone or mobile phone number e.g. "123abc"
10. Click 'Publish' button to publish the location

### Expected Behavior

The page (buttons, visible fields, etc.) stay the same.
No version for a page that is not published.

### Actual Behavior

No location or contact is created, however the page changes:

- 'Publish' button changes to 'Update'
- 'Implications on translations' now visible
- 'Actions' now visible (clicking 'Archive' now will result in Internal Server Error)

-  A 'Version' is now visible with 'Publication Status: Published', the 'Show' button can be clicked and it opens 404 page not found error.

### Screenshot

<img width="1632" height="720" alt="Image" src="https://github.com/user-attachments/assets/16533283-9fa5-4b8b-ad00-79497f4f2206" />
<img width="1905" height="988" alt="Image" src="https://github.com/user-attachments/assets/9bcc50c0-a79f-4a9a-aefc-c9edc0801fb7" />

### Additional Information

<details><summary>Traceback</summary>

```
```

</details>

### Related Issues

https://github.com/digitalfabrik/integreat-cms/pull/4099

PR #4116, opened 2026-01-28
Reinstantiate Forms after Fail

hannaseithe:
### Short description
This is part of the fix for #4107 after a first part has already been merged. In order to get the forms back into the state our template can handle I reinstantiate them after a fail

### Proposed changes

- use is_edit in places of the template where we need to distunguish between edit state and create state
- reinstantiate the forms after rollback, since calling .is_invalid and .save on them mutates them in such a way that the template is breaking

### Side effects

- None

### How to test

Check all ways a poi can be saved:
- with or without a related contact
- in valid or invalid state
- also check updating an existing poi
- check if creating a Draft and Autosave also work as intended

### Resolved issues

Fixes: #4107 

### DIFF
--- a/integreat_cms/cms/views/pois/poi_form_view.py
+++ b/integreat_cms/cms/views/pois/poi_form_view.py
@@ -4,7 +4,6 @@ A view representing an instance of a point of interest. POIs can be created or u
 
 from __future__ import annotations
 
-import json
 import logging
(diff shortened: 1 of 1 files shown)


MizukiTemma (changes requested):
Thank you for updates 😸 

Sadly I'm facing this error 🙈 

Image

PR #4124, opened 2026-01-30
Devcontainer: Fix missing yarn gpg key

jarlhengstmengel:
### Short description

This fixes missing public gpg key for yarn when building the docker image in the devcontainer. I used the solution as mentioned in the comment section of [this](https://github.com/Azure/static-web-apps-cli/issues/982) bug report. This came up while I tried to rebuild my devcontainer in VS code.

### Proposed changes

- Curl for public gpg key for yarn and store it so that apt finds it

### Side effects

- Should be none. I'm not entirely shure if this is something that is going to be solved externally or if this is a good long term fix for us here

### How to test

Build the devcontainer in VS Code

### Resolved issues

Fixes: # n.a.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,6 +1,6 @@
 FROM mcr.microsoft.com/vscode/devcontainers/python:3.11-bookworm
 
-RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | gpg --dearmor > /usr/share/keyrings/yarn-archive-keyring.gpg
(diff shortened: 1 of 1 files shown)


dkehne:
Since yarn is never installed via apt in this Dockerfile (Node comes through `n`), would it be simpler to just remove the stale source list instead?

```dockerfile
RUN rm -f /etc/apt/sources.list.d/yarn.list
```

That way we don't depend on an external URL and avoid the same issue if the key expires again.

MizukiTemma:
From Weekly on 08.04: we'll review and merge this PR as it is necessary for people using dev container.

PR #4146, opened 2026-02-13
Fix #3774: Change geocoding error to orange warning

### LINKED ISSUE #3774: Creating address with postbox creates a "red warning"
### Describe the Bug
If someone adds a location with a postbox, instead of an address, the user get's a red error message.

### Steps to Reproduce

1. Go to "Location"
2. Create a Location, use a postbox e.g. "Postfach 1234" in the field "Straße und Hausnummer"

### Expected Behavior
The message "address was not found" should be orange because it is a warning. Saving is possible nevertheless. 

### Actual Behavior
The message "address was not found" is red (=error)

### Screenshot

Image

dkehne:
## Summary
- Change the geocoding failure message styling from red (error) to orange (warning), since saving is still possible when coordinates can't be found
- Update message text from "Address could not be found" to "Coordinates could not be found" which better describes the actual issue
- Update German translation accordingly

Closes #3774

## Test plan
- [ ] Create a new POI with a postbox address (e.g. "Postfach 1234") and verify the warning appears in orange
- [ ] Verify the message text reads "Coordinates could not be found" (or "Koordinaten wurden nicht gefunden." in German)
- [ ] Verify saving still works as expected
- [ ] Verify that when "show on map" is checked and coordinates are missing, the form validation error still correctly blocks saving

### DIFF
--- a/integreat_cms/cms/templates/pois/poi_form_sidebar/position_box.html
+++ b/integreat_cms/cms/templates/pois/poi_form_sidebar/position_box.html
@@ -15,7 +15,7 @@
             {% translate "Address" %}
         </label>
         <div id="nominatim-error"
-             class="bg-red-100 border-l-4 border-red-500 text-red-700 px-4 py-3 my-2 hidden"
+             class="bg-orange-100 border-l-4 border-orange-500 text-orange-700 px-4 py-3 my-2 hidden"
              role="alert">
         </div>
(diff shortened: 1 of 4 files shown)


MizukiTemma (approved):
Thank you 😸

dasgoutam (approved):
Does as expected

Issue #4149, opened 2026-02-16
test_duplicate_pages fails


PeterNerlich:
### Describe the Bug

`test_duplicate_pages` currently always fails with `django.core.exceptions.ValidationError: ['Slug kann nicht aus dem Titel abgeleitet werden.']`

### Steps to Reproduce

1. Run test
2. See error

### Expected Behavior

The test is successful, it is just duplicating pages

### Actual Behavior

The test fails with `Slug kann nicht aus dem Titel abgeleitet werden.` despite there being a title

### Additional Information

<details><summary>Traceback</summary>

```
```

</details>

### Related Issues

#4076

dkehne:
I looked into this and traced the root cause:

**Problem 1: `duplicate_page()` uses `slugify()` without `allow_unicode=True`**

At `duplicate_pages.py:55`, the slug is generated with `slugify(new_title)` (defaults to `allow_unicode=False`), but the `PageTranslation.slug` field has `allow_unicode=True`. The Augsburg test data has 29 translations with non-Latin titles (Arabic, Farsi, Ukrainian, Amharic) — `slugify()` strips all non-ASCII, leaving the slug empty unless the random suffix happens to contain ASCII alphanumeric characters.

**Problem 2: `generate_unique_slug()` requires `cleaned_data` but `clean()` doesn't pass it**

When the slug is empty and `clean()` → `generate_unique_slug()` is called with `fallback="title"` but no `cleaned_data`, the function checks `if "title" not in {}` → raises `ValidationError`, despite `self.title` being available on the instance.

**Possible alternative to #4150:** Instead of passing a constructed `cleaned_data` dict from every caller, the fallback logic in `generate_unique_slug()` itself could be made resilient:

```python
# Current (broken without cleaned_data):
if fallback not in cleaned_data:
    raise ValidationError(...)
slug = slugify(cleaned_data[fallback], allow_unicode=allow_unicode)

# Alternative:
fallback_value = cleaned_data.get(fallback) or getattr(object_instance, fallback, None)
if not fallback_value:
    raise ValidationError(...)
slug = slugify(fallback_value, allow_unicode=allow_unicode)
```

Single point of change, preserves form-based behavior, and eliminates the need for callers to construct `cleaned_data` dicts. Additionally, `duplicate_pages.py:55` should pass `allow_unicode=True` to `slugify()`.

@PeterNerlich what do you think, does this seem reasonable?

PeterNerlich:
[...]

PR #4150, opened 2026-02-16
Fix generate_unique_slug() without Form

PeterNerlich:
### Short description

`test_duplicate_pages` currently always fails with `django.core.exceptions.ValidationError: ['Slug kann nicht aus dem Titel abgeleitet werden.']`. It seems that `generate_unique_slug()` is built with forms in mind, but also called directly from `PageTranslation.clean()`, where `kwargs["cleaned_data"]` is not supplied and thus regarded `{}`, meaning that the title is not where `generate_unique_slug()` looks for a fallback.

### Proposed changes

- In `PageTranslation.clean()`, gather the current attributes of the object into a dict to pass in as `cleaned_data`

### Side effects

- This might be a source of confusion later, the alternative is to consider a refactoring of `generate_unique_slug()` to not name it `cleaned_data` and not imply the involvement of a form.
- There might be differences in the way a form produces `cleaned_data` and how the naive gathering of attributes via `._meta.fields` works. Fields may be missing, produce extra entries and unexpected values. It currently works for slugs but expecting this to be generalizable to any attribute is a dangerous assumption.

### How to test

Run `test_duplicate_pages`

### Resolved issues

Fixes: #4149

### DIFF
--- a/integreat_cms/cms/models/abstract_content_model.py
+++ b/integreat_cms/cms/models/abstract_content_model.py
@@ -207,6 +207,7 @@ class AbstractContentModel(AbstractBaseModel):
                 "foreign_model": self._meta.model_name,
                 "region": self.region,
                 "language": translation.language,
+                "fallback": self.created_date.isoformat(),
             }
             translation.slug = generate_unique_slug(**kwargs)
             translation.currently_in_translation = False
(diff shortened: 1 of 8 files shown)


MizukiTemma:
Merged as it was already approved a week ago and there is no following discussion.

PR #4152, opened 2026-02-17
Fix URL replacement for percent-encoded URLs

### LINKED ISSUE #4129: URL replacement function not working correctly
## Description

The central URL replacement function in the broken link checker does not seem to work correctly. A URL in Munich that is used 20+ times in different places cannot be replaced.

## Information Needed

@nassabay Could you please provide:
1. The exact URL you're trying to replace
2. What you're replacing it with
3. Which interface you're using (Search & Replace form or pencil icon on specific URL)
4. What happens when you try (success message but no change? error? "no links found"?)
5. If possible, check the server logs when attempting the replacement for any relevant messages

This will help identify whether the issue is URLs not being found, or found but not matching during HTML replacement.

dkehne:
## Summary
- Fixes #4129: The linkchecker's "Replace URL" button silently fails for URLs containing non-ASCII characters (e.g., Cyrillic) due to an encoding mismatch between lxml's `rewrite_links()` callback and the linkcheck `Url.url` dict keys
- Normalizes URL encoding in `replace_urls()` by building a lookup dict that maps both raw and `unquote()`-decoded URLs to their replacements
- Re-enables previously commented-out test assertions that verify replacement results

## Test plan
- [ ] Run `tools/test.sh tests/cms/views/link_replace/test_link_actions.py` to verify existing and re-enabled assertions pass
- [ ] Manually test replacing a URL containing non-ASCII characters (e.g., Cyrillic) via the linkchecker's pencil icon

### DIFF
--- a/integreat_cms/cms/models/abstract_content_translation.py
+++ b/integreat_cms/cms/models/abstract_content_translation.py
@@ -4,6 +4,7 @@ import logging
 from copy import deepcopy
 from html import escape
 from typing import TYPE_CHECKING
+from urllib.parse import unquote
 
 from django.conf import settings
 from django.db import models, transaction
@@ -617,9 +618,22 @@ class AbstractContentTranslation(AbstractBaseModel):
         """
         new_translation = self.create_new_version_copy(user)
         logger.debug("Replacing links of %r: %r", new_translation, urls_to_replace)
+
+        # Build a normalized lookup: map decoded URLs to their replacements too
+        # so that percent-encoded URLs in HTML match against decoded dict keys
+        normalized: dict[str, str] = {}
+        for old_url, new_url in urls_to_replace.items():
+            normalized[old_url] = new_url
+            normalized[unquote(old_url)] = new_url
+
+        def resolve_url(content_url: str) -> str:
+            return normalized.get(
+                content_url, normalized.get(unquote(content_url), content_url)
+            )
+
         new_translation.content = rewrite_links(
             new_translation.content,
-            lambda content_url: urls_to_replace.get(content_url, content_url),
+            resolve_url,
         )
         new_translation.content = fix_content_link_encoding(new_translation.content)
         if new_translation.content != self.content and commit:
--- a/integreat_cms/release_notes/current/unreleased/4129.yml
+++ b/integreat_cms/release_notes/current/unreleased/4129.yml
@@ -1,0 +1,2 @@
+en: Fix URL replacement not working for percent-encoded URLs
+de: URL-Ersetzung für prozent-kodierte URLs repariert
--- a/tests/cms/views/link_replace/test_link_actions.py
+++ b/tests/cms/views/link_replace/test_link_actions.py
@@ -90,9 +90,7 @@ def test_url_replace(
         )
 
         assert Link.objects.filter(url__url=OLD_URL).count() == after
-        """
         assert Link.objects.filter(url__url=NEW_URL).count() == before - after
-        """
 
     elif role == ANONYMOUS:
         assert response.status_code == 302
@@ -184,12 +182,10 @@ def test_search_and_replace_links(
         )
 
         assert Link.objects.filter(url__url=SEARCH_REPLACE_TARGET_URL).count() == after
-        """
         assert (
-           Link.objects.filter(url__url=TARGET_URL_AFTER_REPLACE).count()
-           == before - after
+            Link.objects.filter(url__url=TARGET_URL_AFTER_REPLACE).count()
+            == before - after
         )
-        """
 
     elif role == ANONYMOUS:
         assert response.status_code == 302


hannaseithe (commented):
I have to say that I struggle with reviewing this quite a bit, for the following reasons:

From the original bug, I cannot tell for sure what the actual issue was: 
- did using a decoded url for replacing not work or a encoded? 
When working on an issue the bug description should clearly describe how to reproduce the bug.

When trying to reproduce I also run into the problem on local that Link objects are not always created on save, and so far I could not figure out why. I was able to create link objects when adding a link on an arabic page, but not on an Ukrainian. See: https://chat.tuerantuer.org/digitalfabrik/pl/z1btgf91m388ddratay3u9ac4r

Regarding your code, as far as I can tell for now:
Unquoting LTR languages might actually work the way you suggested it, but at least with RTL languages like arabic this will not solve it, as unquoting a percentage encoded url gives you the arabic letters in the wrong (LTR) order. Furthermore I am not sure if this is maybe also Unicode normalization issue (NFC vs NFD)

integreat_cms/cms/models/abstract_content_translation.py
```diff
@@ -617,9 +618,22 @@ def replace_urls(
...
+        def resolve_url(content_url: str) -> str:
+            return normalized.get(
+                content_url, normalized.get(unquote(content_url), content_url)
+            )
+
         new_translation.content = rewrite_links(
             new_translation.content,
-            lambda content_url: urls_to_replace.get(content_url, content_url),
+            resolve_url,
         )
```
PeterNerlich, line 637: I don't think just trying both the old URL as is and its unquoted version is a good approach, because it muddies our separation of realms and is introducing opportunities for obscure bugs where the system misbehaves in a way that is hard to detect.

Instead, we should expect the list of old links and their respective replacements to be in human readable format and transparently handle the conversion between that and lxml.

```suggestion
        def quoted_replacement_url(old_quoted: str) -> str:
            old_unquoted = unquote(old_quoted)
            if old_unquoted not in urls_to_replace:
                return old_quoted
            replacement = urls_to_replace.get(old_unquoted)
            return quote(replacement)
        
        new_translation.content = rewrite_links(
            new_translation.content,
            quoted_replacement_url,
        )
```

From this it becomes apparent that we need to adjust our views as well. `linkcheck`—just like lxml—works only with quoted strings and we need to ensure we do the conversions properly where we want to work in human readable form. This includes the URL model and the string saved on it.

integreat_cms/cms/models/abstract_content_translation.py
```diff
@@ -4,6 +4,7 @@
 from copy import deepcopy
 from html import escape
 from typing import TYPE_CHECKING
+from urllib.parse import unquote
```
PeterNerlich, line 7:
```suggestion
from urllib.parse import quote, unquote
```

dkehne:
@nassabay do you have any example cases at the moment that would enable the cms-team to test? in the original issue (https://github.com/digitalfabrik/integreat-cms/issues/4129) you already delivered some cases, could you deliver more (current ones)?

dkehne:
@hannaseithe @jarlhengstmengel after I have rejected another PR from me where i worked with AI on a difficult topic and @PeterNerlich found sth. that i could in the end confirm by doing a lot of testing, i also invested some more time in here and tested the scenarios on my local instance.

The encoding mismatch doesn't trigger in normal usage. Both Python's html.parser (which linkcheck uses to populate Url.url) and lxml's rewrite_links (used in replace_urls) extract href values in exactly the same format as they sit in the HTML. 

The only edge case where it could fail is: content was originally saved with decoded Cyrillic in the href → linkcheck scanned it → Url.url = decoded → the content was later re-saved through a form (which runs clean_content → lxml's tostring → encodes the href) → linkcheck did NOT re-scan (e.g. saved with DISABLE_LISTENERS) → now Url.url is still decoded but the HTML has encoded URLs. That's a very narrow race condition.

The commented-out assertions fail for a completely different reason: linkcheck's tests_running flag is False under pytest (it only triggers for manage.py test), so link scanning happens in a background daemon thread. The time.sleep(0.5) in the view is not long enough to guarantee the thread finishes before the test assertion runs. The new URL links simply aren't there yet.

 **Conclusions for the PR**
1. The PR's defensive unquote fix is harmless but solves a problem that essentially doesn't occur in practice.
2. The commented-out assertions should stay commented out (or be properly fixed with a tasks_queue.join() / thread synchronization). Re-enabling them as the PR does will likely produce flaky tests.
3. The actual bug salua reported (replace_urls silently doing nothing) is still unexplained - worth reproducing on a real production instance with the actual Munich data to find the true root cause. I will have another more detailed look...

PeterNerlich (changes requested):
*The statement below the line is not true anymore, as following the closing of this PR the review was aborted before a proper working alternative suggestion could be cobbled together. I decided to still submit the review in its intermediate state, as it might be helpful to further investigation and working on the underlying issue.*

*I most recently ended up concluding that the exact separation that I describe in the following paragraphs is in fact not what happens to a user e.g. when looking in the browsers URL bar. My working theory is that, just as lxml converts **every individual character** that is not an ASCII character and thus not allowed in an URL into its percent encoded equivalent, when the browser navigates to a URL and displays it, it parses each individual percent encoding and displays that **as long as it falls outside the ASCII range**. This could make the function bidirectionally stable. This claim and whether browsers work this way still has to be verified. And I think whatever browsers do in this case is a great indicator for what the user expects and what we should emulate.*

---

This is a difficult issue to get right and I found myself spending more and more time on yet more research during this review. In the end, instead of the tiny suggestion to take a different approach, I ended up with a few suggestions reaching well beyond the files you requested to change.

URL encoding should only be done at very precise points in the program that separate it into a realm dealing with and tranferring URL encoded data and not URL encoded data. To blindly try both variants is bad practice imo because something like `l%C3%A4rmschutz%3F%20ja%21` is **only usually** the encoded form of the human readable string `lärmschutz? ja!`. That has to be URL encoded to to be safely passed around as part of an URL *(e.g. `page/l%C3%A4rmschutz%3F%20ja%21?search=free`)*. But there is no guarantee that it is not just a randomly generated 26 character slug that is alphanumeric and contains special characters which produced only valid escape sequences by pure luck.

**We can only ever reliably tell if a string is definitely not URL encoded**: If it contained another unencoded special character, it would be clear that all those %s are down to chance.
**If it doesn't, it could be both—it's always a guessing game**.

So where are the points in our software that devide it into the two realms? It should be wherever a link gets embedded into a HTML string, and wherever it gets parsed out—so roughly where UI composes content and where we process HTML nodes internally, like the link checker. But because it is a confusing topic lxml and linkcheck don't do automatic conversion—this would only provoke more mistakes.

<details><summary>Where (partial) conversions already happen and why they aren't enough</summary>
Our link plugin in TinyMCE does this automatically in most cases *(there might be potential to ensure that UI doesn't mess something up already)*, but even if I ensure it gets sent to the server unencoded in the HTML, in the response it is URL encoded again. This is done when it is parsed by lxml in `clean_content()` *(`integreat_cms/cms/utils/content_utils.py`)* before it gets saved to the database, but not in the way one would expect.

Same with lxmls `rewrite_links()`: It processes the links as they are in the content, so URL encoded, and it expects the same from replacement strings. If it is given a non-conforming character, it transforms that into its corresponding code to gracefully continue working somehow. What I noticed and find worrying is that this is really just the individual character, not the rest of the URL:

```py
>>> # Human readable, non-URL-encoded URL that just happens
>>> # to contain something that looks like a URL encoded character
>>> s = "https://example.com/lärm%3F"
>>> # URL encoded version
>>> urllib.parse.quote(s)
'https://example.com/l%C3%A4rm%253F'
>>> # Giving rewrite links a non-URL-encoded version breaks,
>>> # producing a URL that corresponds to neither version
>>> rewrite_links('<a href="x"></a>', lambda _: "https://example.com/lärm%3F"
'<a href="https://example.com/l%C3%A4rm%3F"></a>'
```

So lxml does not take care to encode / decode URLs as you correctly discovered, and it does feel appropriate, since now we as developers can deal with this cleanly and with intent instead of fighting against lxml. We can put the library firmly on the encoded side.
</details>

PR #4153, opened 2026-02-17
Fix link update mechanism failing on slug changes

### LINKED ISSUE #4119: Link update mechanism fails to resolve links with historical slugs
## Summary

The automatic link update mechanism implemented in PR #2530 (Issue #1878) fails to fix outdated links when slugs change, particularly affecting machine-translated content. This causes "Die URL ist nicht aktuell" errors in the linkchecker.

## Root Cause

There's a mismatch between how the system **finds** pages with outdated links and how it **resolves** those links:

| Function | Uses Historical Slugs? |
|----------|----------------------|
| `get_referencing_translations()` | ✅ Yes (via `get_all_used_slugs()`) |
| `get_public_translation_for_link()` | ❌ No (only current slugs) |

### The Problematic Flow

1. **Page A** (Arabic) has slug history: `["مرحبا", "أهلا"]` (changed over time via machine translation)
2. **Page B** (Arabic) contains a link to Page A using the OLD slug: `.../ar/مرحبا/`
3. Page A is re-translated, and `update_links_to()` is called
4. `get_referencing_translations()` correctly finds Page B because it searches **historical slugs** via `get_all_used_slugs()` ([content_translation_utils.py:80](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/utils/content_translation_utils.py#L80))
5. `clean_content()` is called to fix Page B's content
6. `get_public_translation_for_link()` tries to resolve `.../ar/مرحبا/`
7. **The search fails** because the query at [internal_link_utils.py:152](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/utils/internal_link_utils.py#L152) looks for `page__translations__slug = "مرحبا"`, but the CURRENT Arabic translation now has slug `"أهلا"`
8. Since resolution fails, `update_internal_links()` returns `None` ([internal_link_utils.py:52-53](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/utils/internal_link_utils.py#L52-L53))
9. **The link is NOT updated**

## Why Machine Translation Exacerbates This

When machine translation runs:

1. **No slug is provided** in the translation data ([deepl_api_client.py:78-83](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/deepl_api/deepl_api_client.py#L78-L83) - only `title`, `content`, `meta_description` are included)
2. The form's `clean_slug()` generates a **new slug from the translated title** ([custom_content_model_form.py:111-117](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/forms/custom_content_model_form.py#L111-L117))
3. If the translation API returns slightly different translations (common with LLMs), different slugs are generated each time
4. The OLD slug becomes "orphaned" - it exists in the version history but not as a current slug
5. Links to the old slug can't be resolved by `get_public_translation_for_link()`

## Steps to Reproduce

1. Create Page A with a German translation (slug: `willkommen`)
2. Machine-translate Page A to Arabic (gets slug: `مرحبا`)
3. Create Page B with a German translation that links to Page A
4. Machine-translate Page B to Arabic (the link now points to `.../ar/مرحبا/`)
5. Re-machine-translate Page A to Arabic (if title translation varies, gets new slug: `أهلا`)
6. Check Page B Arabic - the link still points to `.../ar/مرحبا/` which is now broken
7. The linkchecker reports "Die URL ist nicht aktuell"

## Proposed Solution

Modify `get_public_translation_for_webapp_link_parts()` in `internal_link_utils.py` to search for **any version** of a translation with the matching slug, not just current translations.

**Option A**: Search all translation versions for the slug
```python
# Current (only finds current slugs):
filter_args[f"{foreign_object}__translations__slug"] = object_slug

# Proposed (finds historical slugs too):
# First, find any translation (current or historical) with this slug
# Then get its foreign object and find the current public translation
```

**Option B**: Pre-compute a slug replacement map in `update_links_to()` before calling `clean_content()`, using the known historical-to-current slug mapping.

**Option C**: When machine-translating, preserve the existing slug if a translation already exists (only generate new slug for first-time translations).

## Related Issues/PRs

- Issue #1878 - Auto create/update link texts to Integreat content
- PR #2530 - Automatically update internal links

## Impact

- Affects all machine-translated content where slugs change between translation runs
- Causes broken internal links that appear in the linkchecker as "Die URL ist nicht aktuell"
- The more frequently content is re-translated, the more likely links become broken

dkehne:
Fixes #4119

## Summary

When a page translation is saved with a changed slug, the CMS should automatically update all internal links pointing to it. Due to a stale `@cached_property`, this update was silently skipped on every normal form save.

**Fix**: Call `invalidate_cached_translations()` on the foreign object before `update_links_to()` in `CustomContentModelForm.save()`, so the cached-property chain reflects the just-saved version rather than the pre-save state.

Note: the slug preservation change originally included in this PR has been dropped — #4173 already prevents unnecessary slug regeneration by only sending changed fields to the MT API.

## Test Plan

- [ ] Create Page A and Page B, where Page B contains an internal link to Page A
- [ ] Machine-translate both pages to another language (e.g., Arabic)
- [ ] Change the German title of Page A (which updates its slug) and publish
- [ ] Check Page B in the translated language — the link should now point to the updated URL
- [ ] The linkchecker should not report "Die URL ist nicht aktuell" for Page B
- [ ] Run `tools/test.sh tests/cms/views/link_replace/` to verify existing tests pass

### DIFF
--- a/integreat_cms/core/utils/machine_translation_api_client.py
+++ b/integreat_cms/core/utils/machine_translation_api_client.py
@@ -424,15 +424,6 @@ class MachineTranslationApiClient(ABC):
                     "push_notification": ctx.instance,
                 }
             )
-        # Preserve the existing slug when re-translating, so that
-        # machine translation does not generate a new slug from the
-        # (potentially non-deterministic) translated title each time
-        if (
-            "slug" not in translation_data
-            and content_object.existing_target_translation
-        ):
-            translation_data["slug"] = content_object.existing_target_translation.slug
-
         content_translation_form = self.form_class(
             data=translation_data,
             instance=ctx.existing_target_translation,


jarlhengstmengel:
I honestly not sure what to do with this. After testing, I get the feeling that it solves a different problem then the Issue or PR discribtion implies. I did the following steps:
1. Create new page
2. Added a link to a existing page
3. Translated the page

On develop this generated links in the translations that didn't change. They were still pointing at the link to the german content, no matter the language. On the Branch of this PR, these links get changed in the right way to the the translated links. Also the actual slugs, don't get translated anymore when the german slug changed. This seems intentional when I read the comment in the Issue. But I have a bad feeling about this strategy for difficult to forsee bugs. Was this discussed at some point?
Last point, when I change the german title of the page that is linked to, the german slug updates. The link in the german version of the new page doesn't change. I would expect that from the test plan in the PR disciption.

dkehne:
@jarlhengstmengel 

1. The behavior you observed (links correctly updating to translated versions) is exactly the bug being fixed — the stale @cached_property caused all link updates to be silently skipped on develop, so links always stayed pointing at the German content. 
2. The slug preservation only applies during machine translation re-runs (save_translation() in the MT client), not during manual edits. It prevents non-deterministic slug generation from breaking links on every re-translation.
4. The manual title change case should work based on the code — update_links_to is triggered whenever title or slug is in changed_data and the translation is PUBLIC. Did you save/publish the page, or save it as a draft?

dkehne:
You're right on both points. #4173 already handles slug churn correctly by only sending changed fields to the MT API — if the source title hasn't changed, the title won't be re-translated and no new slug is generated. Our blanket "preserve old slug" approach would wrongly suppress slug updates when the source title legitimately changes, so we've dropped that block. The only remaining change in this PR is the cache invalidation fix in `CustomContentModelForm.save()`.

jarlhengstmengel:
Thanks for the update! Could you maybe update the original Issue Description to match all the comments you made with a special focus on the observed behaviour in the CMS itself (not only the flow inside the code) and how to reproduce the bug, since this is the basis for the reviews and the later Testing by Julian. at the moment it's a bit messy to understand.

For the rest it sounds logical to me now, but I can't really test this at the moment since internal links are currently not created right on the develop branch (#4196).

dkehne:
@jarlhengstmengel i edited the description.

PeterNerlich:
Could not reproduce the issue in the first place, see https://github.com/digitalfabrik/integreat-cms/issues/4119#issuecomment-4205707064

dkehne:
Closed because of solving sth. that is not relevant, see https://github.com/digitalfabrik/integreat-cms/issues/4119#issuecomment-4213196132

PR #4155, opened 2026-02-17
Validate recurring event start date against recurrence rule

### LINKED ISSUE #1967: Initial date of recurring events is not validated against the recurrence rule
### Describe the Bug

When creating recurring events, it is not validated whether the initial event matches its own recurrence rule.

### Steps to Reproduce

1. Go to event form
2. Create an event which takes place on Monday and is recurring every week on Tuesday
3. Save

### Expected Behavior

An error should appear since the initial event does not fit into its own recurrence rule

### Actual Behavior

The event is normally saved and the recurrence rule is only valid for all subsequent events

### Additional Information

See https://github.com/digitalfabrik/integreat-cms/issues/1965#issuecomment-1347090272

dkehne:
## Summary
- Validates that the event start date matches the selected recurrence rule when creating/editing recurring events
- For **weekly** recurrence: the start date's weekday must be among the selected weekdays
- For **monthly** recurrence: the start date must match both the selected weekday **and** the selected week of the month (e.g. a start date on the 3rd Tuesday is rejected for a rule saying "1st Tuesday")

The monthly week-of-month check reuses the existing `iter_after()` logic via a temporary `RecurrenceRule` instance, which correctly handles the "last week of month" edge case.

Closes #1967

## Test plan
- [ ] Create a weekly recurring event where the start date's weekday is **not** in the selected weekdays → should show validation error
- [ ] Create a weekly recurring event where the start date's weekday **is** in the selected weekdays → should save normally
- [ ] Create a monthly recurring event where the start date's weekday does **not** match → should show validation error on weekday field
- [ ] Create a monthly recurring event where the weekday matches but the week does **not** (e.g. start on 3rd Tuesday, rule says 1st Tuesday) → should show validation error on week field
- [ ] Create a monthly recurring event where both weekday and week match → should save normally
- [ ] Edit an existing recurring event and verify validation still works

### DIFF
--- a/integreat_cms/release_notes/current/unreleased/1967.yml
+++ b/integreat_cms/release_notes/current/unreleased/1967.yml
@@ -1,0 +1,2 @@
+en: Validate that the start date of recurring events matches the recurrence rule
+de: Validiere, dass das Startdatum wiederkehrender Veranstaltungen zur Wiederholungsregel passt
--- a/integreat_cms/cms/forms/events/recurrence_rule_form.py
+++ b/integreat_cms/cms/forms/events/recurrence_rule_form.py
@@ -84,16 +84,29 @@ class RecurrenceRuleForm(CustomModelForm):
                     _('Recurrence "daily" is not allowed anymore'), code="invalid"
                 ),
             )
-        elif cleaned_data.get("frequency") == frequency.WEEKLY and not cleaned_data.get(
-            "weekdays_for_weekly",
-        ):
-            self.add_error(
-                "weekdays_for_weekly",
-                forms.ValidationError(
-                    _("No weekdays for weekly recurrence selected"),
-                    code="required",
-                ),
-            )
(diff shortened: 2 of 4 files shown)


integreat_cms/release_notes/current/unreleased/1967.yml
```diff
@@ -0,0 +1,2 @@
+en: Validate that the start date of recurring events matches the recurrence rule
+de: Validierung, dass das Startdatum wiederkehrender Veranstaltungen zur Wiederholungsregel passt
```
MizukiTemma:
```suggestion
de: Validiere, dass das Startdatum wiederkehrender Veranstaltungen zur Wiederholungsregel passt
```

We use imperative form in release notes 📝

dkehne:
Applied your release note suggestion. @MizukiTemma could you merge when you get a chance? 🙏

PR #4194, opened 2026-03-12
add sortable table headers in POI list view

### LINKED ISSUE #1936: Filter/sort function for POIs
In order to find the locations more quickly or to have an overview of all schools, for example:
our suggestion would be that we place the same filter function here as for events or pages, i.e. there is the possibility to filter locations, but this is possible on the basis of the categories. 

So -> We wish that there is a filter by category option in the editorial system.

jonbulz:
### Short description

Add sort function to POI list view

### Proposed changes

- allow sorting by Title, Street, Postal Code, Country and Category
- because the list view features the 'Publication status', the standard `render_table_header` helper function could not be used. instead, the `sort_link` tag is used directly in the sortable table headers

### Side effects

- Because the table header cannot be rendered directly from the `table_fields` property, it needs to be ensured that every change in the table headers in the template is also reflected in the `table_fields`

### How to test

- go to POI list view
- check the sorting works for Title, Street, Postal Code, Country and Category

### Resolved issues

Fixes: #1936 

### DIFF
--- a/integreat_cms/cms/templates/pois/poi_list.html
+++ b/integreat_cms/cms/templates/pois/poi_list.html
@@ -2,6 +2,7 @@
 {% load i18n %}
 {% load rules %}
 {% load static %}
+{% load sort_tags %}
 {% load content_filters %}
 {% block content %}
     {% get_current_language as LANGUAGE_CODE %}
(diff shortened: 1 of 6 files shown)


MizukiTemma (changes requested):
Thank you for the PR 😸 Generally it looks good. I found one thing :eyes

I tested with those 4 POIs
Image

Desc sorting seems to work good
Image

but asc sorting shows three times "Test-Ort" somehow.
Image

Could you check once again?


💡 and I think we can add a release note in this PR, as the changes are visible for users.

jonbulz:
@MizukiTemma thanks for your review! I thought I had that already covered, but it turns out the way Django ORM works in this respect is more complex than I thought. I learned something new from that 😄 
Anyway, I think I fixed it now :)

MizukiTemma (approved):
Thank you for updates 👍 Works as expected, the codes look also nice 😸

jarlhengstmengel (commented):
Thank you very much for the PR! Generally looks good to me. I think Sina has some Design Feedback and will come to you with that. I have one question, why not sort by Publication status? Or would you say thats just not enough for sorting?

jarlhengstmengel:
Ah and maybe add an Release Note? Since #4055 had one?

jonbulz:
I added a release note, included publication status in the sortable fields, and changed the styling as @m0ndra suggested in their [comment](https://github.com/digitalfabrik/integreat-cms/issues/1936#issuecomment-4250935063). Could you please re-review? 🙂

jarlhengstmengel (approved):
Thank you very much for the changes! @m0ndra and I had a look together and looks good!

PR #4195, opened 2026-03-16
Refactor for common news endpoint

### LINKED ISSUE #4174: Provide common news endpoint / Refactor TüNews
### Motivation
We will be adding Amal News as a news source, so we need to provide a common news endpoint for the app to access.

### Proposed Solution

Instead of crawling multiple different endpoints, there should be one CMS endpoint providing a list of all recent news, i.e. including local news, tünews and amal news in the same response (depending on which news sources are enabled in the CMS).

All results should have the same properties (preferably analogous to other endpoints, e.g. events) and a field to indicate the name and url of the source.

### User Story
As an Integreat user I want to have all news from different providers in one tab to avoid missing out and switching tabs.

### Related Issues
App issue: https://github.com/digitalfabrik/integreat-app/issues/2537

MizukiTemma:
### Short description

This PR implements a common endpoint for news (push notifcation) and Tü News.

### Proposed changes

- Add a new endpoint `all-news/` that delivers both PNs and posts of Tü News
- Add `source` flag to distinguish PNs form Tü News posts
- Keep the old endpoint `fcm/`,`sent_push_notifications`) not to break the app
- Add a celery task to collect and save Tü news posts as cache
- ~Add Tü News related models and management command to integrate functionalities of our [Tü News repository](https://github.com/digitalfabrik/tunews)~
- ~The current approach is based on that we will store Tü News posts in our Integreat CMS database and the memory cache solution proposed [here](https://github.com/digitalfabrik/integreat-cms/issues/4174#issuecomment-3959935110) comes later~

### Side effects

- ~A field `translations` is added which does not exists in [digitalfabrik/tunews](translations) to provide [reference to translations like for PNs](https://github.com/digitalfabrik/integreat-cms/blob/a9d9dff0da6cc1171eada8e30dab61f4bbe0564b/integreat_cms/api/v3/push_notifications.py#L79)~

- ~German verbose names should probably be changed to English 🤔 Is there any harm @svenseeberg @steffenkleinle ?~

- ~I applied [the 28 days rule](https://github.com/digitalfabrik/integreat-cms/blob/a9d9dff0da6cc1171eada8e30dab61f4bbe0564b/integreat_cms/api/v3/push_notifications.py#L44-L47) to nes from Tü News too. Is that okay or unnecessary?~

- Currently [Tü News Bridge](https://tunews.integreat-app.de/v1/news/languages) is supporting only German, English, Farsi and Arabic. This PR checkes Tü News posts for all languages that are available in the CMS system. 

- ~We probably need a new section in CMS where Service Team can decide for which language we fetch posts from Tü News (not implemented yet). Currently German, English, Farsi and Arabic are handled in [our Tü News](https://github.com/digitalfabrik/tunews?tab=readme-ov-file)~

### How to test

1. Use the command `INTEGREAT_CMS_REDIS_CACHE=1 tools/integreat-cms-cli import_external_news`
2. Check "Enable external news" in a region
3. Try the new endpoint (e.g.`http://localhost:8000/api/v3/augsburg/en/all-news/`) - also check if pagination(`?page=1&size=2`) and filtering works (`?source=tuNews`)

❓ Does anyone have an idea how to test the social media header endpoints (`news/<slug:news_type>/<slug:slug>/`) ?
Self-answer from 24.04.2026: can be tested like this in the console:

image

 
### Resolved issues

Fixes: #4174 


jonbulz (changes requested):
This PR doesn't sit right with me. I don't think this is a good approach. I think there are 3 ways to handle this, all with their pros and cons:
1) Store News in cache, not database: Have some kind of service fetch news form different sources and store them as JSON in our redis cache. Then have our API simply fetch the cached news from external sources in addition to our internal PushNotifications
2) Properly store the external news in our database by transforming them to PushNotifications (with some kind of `source` flag for clarity)
3) If the deadline is urgent, copy-paste the code from tunews as a Django app into our repository and dump the content from the existing database. IMO, this only makes sense as an indermediate step, i.e. the code should be removed at some point once we implemented one of the above solutions. this is much easier if we keep it as a separate app.

Introducing a new `NewsLanguage` model in our cms app feels like a bad idea to me. This is how you get a messy database. No one wants a messy database

integreat_cms/api/v3/social_media_headers.py
```diff
@@ -315,35 +316,56 @@ def news_social_media_headers(
...
+        post = next((post for post in posts if post["id"] == slug), None)
+        if not post:
+            raise Http404("Tü news post not found in this region with this news ID.")
+
+        return render_social_media_headers(
+            request=request,
+            title=post["title"],
+            language_code=language.bcp47_tag,
+            excerpt=post["content"],
+            url=post["link"],
```
MizukiTemma, line 366: @steffenkleinle 

We don't have `get_absolute_url()` for Tü News posts as they are not part of our contents. Is it acceptable to substitute `url` with `link` flag for Tü News posts? ([examples of Tü News](https://tuenews.de/wp-json/wp/v2/posts/?lang=de))

MizukiTemma:
@jonbulz 
Thank you for detailed suggestions and explanation 😍 I hope it's now implemented better 💪

integreat_cms/core/management/commands/import_tuenews.py
```diff
@@ -0,0 +1,89 @@
...
+            news = []
+
+            for post in posts:
+                if not post["acf"]["integreat"]:
+                    continue
+
+                news.append(self.transform_post(post))
+
+            cache.delete(f"tuenews:{language.slug}")
+            cache.set(f"tuenews:{language.slug}", news, timeout=60 * 60 * 24)
```
jonbulz: maybe TTL of 24h is a bit short if we run this command once a day. actually, I guess we could remove the timeout entirely

integreat_cms/api/v3/social_media_headers.py
```diff
@@ -315,35 +316,56 @@ def news_social_media_headers(
...
+        posts = cache.get(f"tuenews:{language_slug}", [])
+        post = next((post for post in posts if post["id"] == slug), None)
+        if not post:
+            raise Http404("Tü news post not found in this region with this news ID.")
+
+        return render_social_media_headers(
+            request=request,
+            title=post["title"],
+            language_code=language.bcp47_tag,
+            excerpt=post["content"],
```
jonbulz: I think the posts don't have a `content` attribute after `transform_post`, so `post["content"]` will raise a `KeyError`

integreat_cms/api/v3/social_media_headers.py
```diff
@@ -315,35 +316,56 @@ def news_social_media_headers(
...
+            title=get_region_title(region, pn_translation.get_title()),
+            language_code=language.bcp47_tag,
+            excerpt=get_excerpt(pn_translation.get_text()),
+            url=f"{settings.WEBAPP_URL}{pn_translation.get_absolute_url()}",
+        )
+    if news_type == "tuenews":
+        if not region.external_news_enabled:
+            raise Http404("Tü News is not enabled in this region.")
+        posts = cache.get(f"tuenews:{language_slug}", [])
+        post = next((post for post in posts if post["id"] == slug), None)
```
jonbulz: slug here is a `str`, are you sure `post["id"]` is as well? if it's an integer, the condition will always be false. Maybe safer to use `str(post["id"])`, or vice versa?

integreat_cms/api/urls.py
```diff
@@ -190,6 +195,11 @@
                     region_social_media_headers,
                     name="social_region_reserved_tunews",
                 ),
+                path(
+                    "news/all-news/",
+                    region_social_media_headers,
+                    name="social_region_reserved_local_news",
```
jonbulz: shadows `"news/local/"` on `urls.py:191`

integreat_cms/api/v3/push_notifications.py
```diff
@@ -53,10 +46,74 @@ def sent_push_notifications(
...
+    language_slug: str,
+) -> list[dict]:
+    """
+    Function to collect Tü News posts
+
+    :param region_slug: slug of a region
+    :param language_slug: language slug
+    :return: list of Tü news posts
+    """
+    if not Region.objects.get(slug=region_slug).external_news_enabled:
```
jonbulz: I think it's safer to pass `request.region` instead of `region_slug`. If you want to keep it like this, you should catch and handle `RegionDoesNotExist`

integreat_cms/core/management/commands/import_tuenews.py
```diff
@@ -0,0 +1,89 @@
...
+        """
+
+        for language in Language.objects.all():
+            response = requests.get(
+                f"http://tuenews.de/wp-json/wp/v2/posts/?lang={language.slug}",
+                timeout=10,
+            )
+
+            if response.status_code != 200:
+                continue
```
jonbulz: silently failing the import could give future us a headache. I'd log at least a warning here

integreat_cms/core/management/commands/import_tuenews.py
```diff
@@ -0,0 +1,89 @@
...
+    def handle(self, *args: Any, **options: Any) -> None:
+        r"""
+        Imports posts from Tü News and save in the cache
+
+        :param \*args: The supplied arguments
+        :param \**options: The supplied keyword options
+        """
+
+        for language in Language.objects.all():
+            response = requests.get(
```
jonbulz: it's probably safer to wrap all of this in a `try/except` block to handle `RequestExceptions` (`ConnectionError`, `Timeout`, `DecodeError`)

integreat_cms/core/management/commands/import_tuenews.py
```diff
@@ -0,0 +1,89 @@
...
+
+            logger.info(
+                "Got %s posts.",
+                len(posts),
+            )
+
+            news = []
+
+            for post in posts:
+                if not post["acf"]["integreat"]:
```
jonbulz: a malformed post will raise an unhandled `KeyError`. safer to wrap in `try/except` and catch `KeyError` and `TypeError`

integreat_cms/core/management/commands/import_tuenews.py
```diff
@@ -0,0 +1,89 @@
...
+        r"""
+        Imports posts from Tü News and save in the cache
+
+        :param \*args: The supplied arguments
+        :param \**options: The supplied keyword options
+        """
+
+        for language in Language.objects.all():
+            response = requests.get(
+                f"http://tuenews.de/wp-json/wp/v2/posts/?lang={language.slug}",
```
jonbulz: no `https`?

jonbulz (changes requested):
Thanks for working on this topic, this is a solid start! I do have some concerns, though:

#### Architecture / Extensibility
We already plan to add a third news source (Amal news) in the near future. There may be more to come after that. With this pattern, we have growing `if/elif` chains in several places. Also, the code for external news lives in `push_notifications.py`, which is a bit confusing. I think even though we decided against a database model, this deserves its own module and class. I don't want to be too specific with the proposed solution, but if you want, we can discuss this in a meeting.

#### Resilience
There are several possible unhandled exceptions in the code, see my additional comments. I think we need to be a bit more defensive when dealing with content from external sources. A malformed response should not crash our process.

#### Other Bugs
There are some other bugs I noticed, e.g. the duplicate URL name, see my comments. Additionally, I think we are still lacking pagination, and the results are not sorted again after tuenews have been added. I'd also prefer if we stick to one naming convention `tuenews` or `tue_news` :)

MizukiTemma:
@jonbulz
Thank you for re-review 😸  Beside your code suggestions, the following changes are applied:
- Rename `api/v3/push_notifications.py` to `api/v3/news.py`
- Stick to `tuenews`
- Add sorting to `all_news` endpoint

> We already plan to add a third news source (Amal news) in the near future. There may be more to come after that. With this pattern, we have growing if/elif chains in several places. Also, the code for external news lives in push_notifications.py, which is a bit confusing. I think even though we decided against a database model, this deserves its own module and class. I don't want to be too specific with the proposed solution, but if you want, we can discuss this in a meeting.

I'm not sure a separate module reduces `if/else` and/or bring a clear intuitive structure 🤔  Maybe I'm not reaching what you imagine.

> Additionally, I think we are still lacking pagination
What do you mean with "pagination"?

jonbulz:
> What do you mean with "pagination"?

@steffenkleinle said in this [comment](https://github.com/digitalfabrik/integreat-cms/issues/4174#issuecomment-4143291301):

> Otherwise a simple endpoint that supports pagination is enough.

This endpoint always returns an array of all news items, not a paginated response. Supporting pagination means we have to accept additional request parameters, e.g. `offset/limit` or `page/page_size` and wrap the response accordingly, e.g.

```
Response:
  {
    "count": 137,
    "next": "https://…/all-news/?limit=50&offset=50&channel=all",
    "previous": null,
    "results": [ ...]
  }
 ```

Or is pagination not needed after all?

File: integreat_cms/core/management/commands/import_tuenews.py [hunk 11]
```diff
@@ -0,0 +1,106 @@
...
+                        if not post["acf"]["integreat"]:
+                            continue
+                        news.append(self.transform_post(post))
+                    except (KeyError, TypeError):
+                        logger.error(  # noqa: TRY400
+                            "Malformed post. Skipped."
+                        )
+
+                cache.delete(f"tuenews:{language.slug}")
+                cache.set(f"tuenews:{language.slug}", news)
```
jonbulz: Actually, I was wrong about the timeout. I found out that setting no timeout means we fall back to the default timeout, which apparently for Django is 300s, if we don't override it for our project (which I don't think we do). So what we want is actually `timeout=None`. Sorry about that 😅

steffenkleinle:
> Or is pagination not needed after all?

I guess this depends on the amount of items we are expecting. But would be good practice at least I guess :)

jonbulz:
Concerning

> I'm not sure a separate module reduces if/else and/or bring a clear intuitive structure

What I mean is this: Imagine we want to add another news source. What we'd have to do now is to edit at least 4 files: 
- `news.py` to add the source to the result list (in the `all_news` endpoint)
- `social_media_headers.py` to extend our `if news_type == ...` chain
- we need a new celery task
- we need a new management command

Each of those is a place where the source-specific logic is coupled to the general logic. If any of those steps is forgotten, it breaks the new source.
I think in the long run we'd be in a much better position if each source were a self-contained thing that the rest of the code could treat uniformly. So generating the result list in `all_news` would become something like:
```
items: list[NewsItem] = []
for source in NEWS_SOURCES:
    if source.is_enabled_for(region):
        items.extend(source.get_items_list(...))
```
or similar. I'd leave the exact implementation up to you, but I think that this abstraction would be very useful. I'm happy to discuss this further if you'd like :)

MizukiTemma:
> > What do you mean with "pagination"?
> 
> @steffenkleinle said in this [comment](https://github.com/digitalfabrik/integreat-cms/issues/4174#issuecomment-4143291301):
> 
> > Otherwise a simple endpoint that supports pagination is enough.
> 
> This endpoint always returns an array of all news items, not a paginated response. Supporting pagination means we have to accept additional request parameters, e.g. `offset/limit` or `page/page_size` and wrap the response accordingly, e.g.
> 
> ```
> Response:
>   {
>     "count": 137,
>     "next": "https://…/all-news/?limit=50&offset=50&channel=all",
>     "previous": null,
>     "results": [ ...]
>   }
> ```
> 
> Or is pagination not needed after all?

Thank you for explanations 😸  I got it what was meant 🙈 

I'm but not sure whether the pagination works well with cliend-side filtering. Our Response contains both push nofitication news and Tü News posts. The app filters them by `source` flag. Shouldn't they already filtered on the CMS side to be pagination enability?

MizukiTemma:
> Concerning
> 
> > I'm not sure a separate module reduces if/else and/or bring a clear intuitive structure
> 
> What I mean is this: Imagine we want to add another news source. What we'd have to do now is to edit at least 4 files:
> 
> * `news.py` to add the source to the result list (in the `all_news` endpoint)
> * `social_media_headers.py` to extend our `if news_type == ...` chain
> * we need a new celery task
> * we need a new management command
> 
> Each of those is a place where the source-specific logic is coupled to the general logic. If any of those steps is forgotten, it breaks the new source. I think in the long run we'd be in a much better position if each source were a self-contained thing that the rest of the code could treat uniformly. So generating the result list in `all_news` would become something like:
> 
> ```
> items: list[NewsItem] = []
> for source in NEWS_SOURCES:
>     if source.is_enabled_for(region):
>         items.extend(source.get_items_list(...))
> ```
> 
> or similar. I'd leave the exact implementation up to you, but I think that this abstraction would be very useful. I'm happy to discuss this further if you'd like :)


I see. Thank you for sharing your idea :)

I would like to suggest to build first an agreement in team how the it should be implemented exactly. Otherwise everyone imagines something diffenrent and this PR then results in a table of long discussion 😓  and also want to avoid this PR staying unmegred for a long time as App Team is waiting for us.

MizukiTemma:
@jonbulz @hannaseithe 
Thank you for the idea exchange meeting 😸 Here is my try with a new structure 🔧 It's now similar to MT providers.

svenseeberg:
> I guess these aren't used in the app but want to be sure 😅 Maybe rather a question to @svenseeberg if these are being used somewhere.

Well, the TuNews Number is shown in each article: https://integreat.app/tuebingen/de/news/tu-news/2290

But you need to check if we render the number into the article or whether this is already part of the content. I think at some point we had the idea of creating a nice, structured footer with the structured data. But I think its likely that we never did that and never will do ;)

steffenkleinle:
> > I guess these aren't used in the app but want to be sure 😅 Maybe rather a question to @svenseeberg if these are being used somewhere.
> 
> Well, the TuNews Number is shown in each article: https://integreat.app/tuebingen/de/news/tu-news/2290
> 
> But you need to check if we render the number into the article or whether this is already part of the content. I think at some point we had the idea of creating a nice, structured footer with the structured data. But I think its likely that we never did that and never will do ;)

Yes, its already part of the content, we don't actively use `enewsno` atm.

integreat_cms/api/v3/news.py
```diff
@@ -0,0 +1,81 @@
...
+) -> JsonResponse:
+    """
+    Function to collect Tü News posts
+
+    :param region_slug: slug of a region
+    :param language_slug: language slug
+    :return: list of Tü news posts
+    """
+    result = TuenewsManager().collect_news_items(region_slug, language_slug)
+    return JsonResponse(result, safe=False)
```
hannaseithe: That can be deleted I think

integreat_cms/cms/constants/news_sources.py
```diff
@@ -0,0 +1,20 @@
...
+
+from ...news_managers.pushnews_manager import PushnewsManager
+from ...news_managers.tuenews_manager import TuenewsManager
+
+if TYPE_CHECKING:
+    from typing import Final
+
+    from ...news_managers.abstract_news_manager import AbstractNewsManager
+
+CHOICES: Final[list[tuple[str, type[AbstractNewsManager]]]] = [
```
hannaseithe:
```suggestion
CHOICES: Final[list[AbstractNewsManager]] = [
```
or 

```suggestion
CHOICES: Final[list[type[AbstractNewsManager]]] = [
```

Meaning: 
1) I would get rid of the str in the tuple and make that a `name` property on the News Manager, 
2) and either already save instances of the Manager in CHOICES (suggestion 1) or make the methods on the News Manager @classmethod s , so that we can follow a singleton(-ish) pattern here. 

Reason for 1: The name of the manager should be encapsulated in the manager itself. It seems redundant/unnecessary to keep it as an external lookup key

Reason for 2: Instantiating a new manager every time we want to access a method seems not very clean, even if it works here because we do not have any instance state (yet).

integreat_cms/news_managers/abstract_news_manager.py
```diff
@@ -0,0 +1,45 @@
...
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from django.http import HttpRequest
+
+    from ..cms.models import Language, Region
+
+
+class AbstractNewsManager(ABC):
```
hannaseithe, line 27:
```suggestion
class AbstractNewsManager(ABC):
   name: str
```
This is where I would add the name of the news managers

integreat_cms/api/v3/social_media_headers.py
```diff
@@ -315,35 +313,27 @@ def news_social_media_headers(
...
 
-    if not (
-        pn_translation := PushNotificationTranslation.objects.filter(
-            language__slug=language.slug,
-            push_notification__id=slug,
-            push_notification__regions=region,
-        ).first()
-    ):
-        raise Http404("Push Notification not found in this region with this language.")
+    news_manager = dict(news_sources.CHOICES).get(news_type, None)
```
hannaseithe:
```suggestion
    news_manager = next((manager for manager in news_sources.CHOICES if manager.name == news_type), None)
```

This suggestions would apply if you turn CHOICES into a list of managers only

integreat_cms/news_managers/abstract_news_manager.py
```diff
@@ -0,0 +1,45 @@
...
+        Imports news items from the source
+
+        To be implemented in the inheriting model
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def collect_news_items(
+        self, region_slug: str, language_slug: str, channel: str
+    ) -> list[dict]:
```
hannaseithe:
```suggestion
    ) -> list[NewsItem]:
```
I would use typing here to validate/guarantee the implicit assumptions about what a NewsItem needs to look like

integreat_cms/news_managers/abstract_news_manager.py
```diff
@@ -0,0 +1,45 @@
...
+
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from django.http import HttpRequest
+
+    from ..cms.models import Language, Region
+
+
```
hannaseithe, line 13:
```suggestion
@dataclass
class NewsItem():
   display_date: Any #we rely upon that existing in news.py
```

But if we actually want to guarantee more than that as a data structure in the API (which I assume we do), this should be expanded.

integreat_cms/api/v3/news.py
```diff
@@ -0,0 +1,81 @@
...
+
+    result = []
+    for _, news_manager in news_sources.CHOICES:
+        result.extend(
+            news_manager().collect_news_items(region_slug, language_slug, channel)
+        )
+
+    sorted_result = sorted(result, key=lambda i: i["display_date"], reverse=True)
+
+    return JsonResponse(sorted_result, safe=False)
```
hannaseithe: I have a problem here with the fact that we are breaking our established pattern of using a transform function inside the endpoint which is our guarantor of the data strcuture that we expose on our API endpoint. Now we have two hidden transform functions that create different data structures and that one has no direct access to on the endpoint module. 

I would expect to have a transform_news_item function that is applied to every element in sorted results. So that if I want to look up, which strucutret the endpoint follows, I would just take a look at the transform function (basically they fullfill the a similar function to the serializers in the django rest framework)

integreat_cms/news_managers/abstract_news_manager.py
```diff
@@ -0,0 +1,45 @@
+from __future__ import annotations
```
hannaseithe, line 1: news_manager as an App needs a __init__.py

integreat_cms/news_managers/abstract_news_manager.py
```diff
@@ -0,0 +1,45 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def social_media_headers(
+        self,
+        request: HttpRequest,
+        region: Region,
+        language: Language,
+        slug: str,
+    ) -> None:
```
hannaseithe: Should have a return type of HttpResponse

hannaseithe (commented):
Hello Mizuki, thank you for working this. I do like that you implemented an abstraction layer for the news managers. This makes is much cleaner. In my review I have focussed on improvements on the abstraction and haven't gone much into detail in regards to specific implementations.

hannaseithe:
One thing I forgot: I feel we should have a test for the new endpoint

integreat_cms/api/v3/news.py
MizukiTemma: I understand your point, but then Tü News posts are every time transformed when the API is called. Instead they can be directly stored in the form they are delivered.  And it's enough consisntent in my opinion if we put the transform functions for news items into respective news managers: the news managers then build an exception that their transform functions are not in the endpoint but they all have their transform function in them. In addition, jumping between a news manager and the endpoint only for this very last step breaks the stream of work.

This is very preference matter. I stick with the current version as long as no large motivation appears for separating transformation from the news managers into the endpoint.

integreat_cms/cms/constants/news_sources.py
MizukiTemma: I go for the suggestion 1 👍

integreat_cms/news_managers/abstract_news_manager.py
MizukiTemma: Hmm 🤔 Do you mean we don't return dict anymore?

hannaseithe: Yes my idea is to return something of type list[NewsItem]. 

... But then there is also my comment about breaking the transform function pattern (elsewhere). So I am a bit undecided myself. I think if it was me - I would implement both this typing of NewsItem as the return on collect_news_items and then on top of it have a transform function in the endpoint that takes the NewsItem and transforms it into whatever we expose in the API. 

My personal gutfeeling is: we neither guarantee to the consumers of the newsManagers any certain structure of the return of collect_news_items. Nor do we inside the endpoint again guarantee a certain structure of the API to the consumers of the API. All that is hidden away in the internals of TueNewsManager and PushNewsManager.

MizukiTemma: Typecheck with `TypedDict` is added (see the commit "Use TypedDict for type check of news items") 😸 I hope this ensures the structure of news items while keeping the existing codes.

MizukiTemma:
> One thing I forgot: I feel we should have a test for the new endpoint

Tests are always good 😸 two (minimum, but the most important) tests are added

MizukiTemma:
@hannaseithe @jonbulz It's again ready for review 🔍

tests/api/test_all_news_endpoint.py
```diff
@@ -0,0 +1,100 @@
...
+        "last_updated": timestamp2.isoformat(),
+        "display_date": timestamp2,
+        "channel": None,
+        "available_languages": None,
+        "source": "tuenews",
+        "link": "https://dummy.com",
+    }
+    cache.set("tuenews:de", [news_post], timeout=None)
+
+    result_order = [dummy_tuenews_id, *result_order]
```
hannaseithe: result_order is empty list

tests/api/test_all_news_endpoint.py
```diff
@@ -0,0 +1,100 @@
...
+
+    kwargs = {"region_slug": test_region_slug, "language_slug": test_language}
+    api = reverse("api:all_news", kwargs=kwargs)
+    response = client.get(api)
+    result = response.json()
+
+    assert len(result) == 1
+    i = 0
```
hannaseithe: Unnecessary?

tests/news_managers/tuenews_manager_test.py
```diff
@@ -0,0 +1,81 @@
...
+
+@pytest.mark.django_db
+def test_import_news_item(load_test_data: None) -> None:
+    with patch("requests.get") as fake_tuenews_server:
+        fake_tuenews_server.return_value.status_code = 200
+        fake_tuenews_server.return_value.json.return_value = dummy_news_items
+
+        assert not cache.get("tuenews:de")
+
+        # Get TuenewsManager via the constant instead of direct import
```
hannaseithe: I think the circualr import issue should be tackled where its caused: I would extract from the social_media_headers.py all utility functions into their own file, to break the circle. Then You have 

social_media_headers -> news_sources -> PushNewsManager/TueNewsManager -> social_media_header_utility

AND

tuenews_test -> TuenewsManager -> social_media_header_utility

integreat_cms/news_managers/pushnews_manager.py
```diff
@@ -0,0 +1,129 @@
...
+            available_languages_dict = {
+                translation.language.slug: {"id": translation.id}
+                for translation in pnt.push_notification.translations.all()
+            }
+            return {
+                "id": pnt.pk,
+                "title": pnt.get_title(),
+                "message": pnt.get_text(),
+                "timestamp": pnt.last_updated,  # deprecated field in the future
+                "last_updated": timezone.localtime(pnt.last_updated),
```
hannaseithe: That is inconsistent with the NewsItem Type

integreat_cms/news_managers/pushnews_manager.py
```diff
@@ -0,0 +1,129 @@
...
+            """
+            available_languages_dict = {
+                translation.language.slug: {"id": translation.id}
+                for translation in pnt.push_notification.translations.all()
+            }
+            return {
+                "id": pnt.pk,
+                "title": pnt.get_title(),
+                "message": pnt.get_text(),
+                "timestamp": pnt.last_updated,  # deprecated field in the future
```
hannaseithe: That is inconsistent with the NewsItem type

tests/news_managers/tuenews_manager_test.py
```diff
@@ -0,0 +1,81 @@
...
+        "link": "https://tuenews.de/news-post-4/",
+    }
+]
+
+dummy_news_items = [not_integreat_news, malformed_news_1, malformed_news_2, valid_news]
+
+
+@pytest.mark.django_db
+def test_import_news_item(load_test_data: None) -> None:
+    with patch("requests.get") as fake_tuenews_server:
```
hannaseithe: From Claude:   In tuenews_manager_test.py, patch("requests.get") patches the library globally rather than the specific usage site. The correct patch target is                                             
  integreat_cms.news_managers.tuenews_manager.requests.get. With parallel test execution (pytest-xdist), global patching can leak between workers.

tests/news_managers/test_tunews_manager.py
```diff
@@ -0,0 +1,81 @@
+from __future__ import annotations
```
hannaseithe, line 1: This folder needs a `__init__.py` as well

tests/news_managers/tuenews_manager_test.py
```diff
@@ -0,0 +1,81 @@
...
+        fake_tuenews_server.return_value.status_code = 200
+        fake_tuenews_server.return_value.json.return_value = dummy_news_items
+
+        assert not cache.get("tuenews:de")
```
hannaseithe: This is fragile. Cache is not automatically cleaned between tests. Depending on execution order ( if test_all_news_endpoints runs first) this will already fail

tests/api/test_all_news_endpoint.py
```diff
@@ -0,0 +1,100 @@
...
+        "channel": None,
+        "available_languages": None,
+        "source": "tuenews",
+        "link": "https://dummy.com",
+    }
+    cache.set("tuenews:de", [news_post], timeout=None)
+
+    result_order = [dummy_tuenews_id, *result_order]
+
+    kwargs = {"region_slug": test_region_slug, "language_slug": test_language}
```
hannaseithe: kwargs computation is redundant here and below

hannaseithe (changes requested):
Hello Mizuki, thank you for working in my change requests.  I had a closer look at the tests and also a suggestion about how to break up the circular import issue

jonbulz:
@steffenkleinle pagination is still missing. things we need to decide on before we continue:
  1. **Response shape**: bare array or envelope with `count` / `next` / `previous`?
  2. **Style** : page-based, offset-based, or cursor-based? (Cursor is drift-stable for the recency-sorted feed; page is simplest.)
  3. **Defaults**: default page size and max page size?

  Notes:
  - Pagination only slices output; server still fetches and merges all sources before slicing.
  - `fcm/` stays unpaginated unless you say otherwise.
  - Each item carries a `source` field (`"local"` / `"tunews"`). Filtering client-side by source is not handled yet.

@hannaseithe I applied my changes (which I think covers everything that you suggested in your reviews). I'd suggest that we wait for the decisions regarding pagination before you do a re-review

MizukiTemma:
@steffenkleinle 
We want to hear your opinion :)

steffenkleinle:
> @steffenkleinle pagination is still missing. things we need to decide on before we continue:
> 
>     1. **Response shape**: bare array or envelope with `count` / `next` / `previous`?

Since we track pagination in the frontend already, a bare array is enough and easier for us.

>     2. **Style** : page-based, offset-based, or cursor-based? (Cursor is drift-stable for the recency-sorted feed; page is simplest.)

Let's just stick to simple `page-based` since its easiest and consistent with the current tüNews endpoint implementation.

>     3. **Defaults**: default page size and max page size?

I don't mind about max page size. Default page size 20 sounds good to me.

> Notes:
> 
>     * Pagination only slices output; server still fetches and merges all sources before slicing.
> 
>     * `fcm/` stays unpaginated unless you say otherwise.
> 
>     * Each item carries a `source` field (`"local"` / `"tunews"`). Filtering client-side by source is not handled yet.

Perfect! Are there plans to implement filtering by source? I think we'd need all local news and all non-local news (i.e. tüNews and amal news), so it also needs to be possible to allow filtering by multiple sources (or a negation).

image

steffenkleinle:
Question: What is your naming for the different news?
localNews, tuNews and amalNews and corresponding source labels `local`, `tuNews` and `amalNews`? I think it would be good if we could agree on something consistent here.

hannaseithe:
> Question: What is your naming for the different news? localNews, tuNews and amalNews and corresponding source labels `local`, `tuNews` and `amalNews`? I think it would be good if we could agree on something consistent here.

We currently have `local`and `tunews`. I can change it to `tuNews` if that suits you better

steffenkleinle:
> > Question: What is your naming for the different news? localNews, tuNews and amalNews and corresponding source labels `local`, `tuNews` and `amalNews`? I think it would be good if we could agree on something consistent here.
> 
> We currently have `local`and `tunews`. I can change it to `tuNews` if that suits you better

Yes, please, more consistency with future `amalNews`. How about your internal naming?

hannaseithe:
> Yes, please, more consistency with future `amalNews`. How about your internal naming?

You mean like classes etc.? 
`tuNews`:  Its TunewsManager and tunews_manager.py
And for the `local` source it is PushnewsManager and pushnews_manager.py

steffenkleinle:
> > Yes, please, more consistency with future `amalNews`. How about your internal naming?
> 
> You mean like classes etc.? `tuNews`: Its TunewsManager and tunews_manager.py And for the `local` source it is PushnewsManager and pushnews_manager.py

Jep. Seems in the code like you have quite a few usages of `tueNews` (mind the `e`).

hannaseithe:
> Jep. Seems in the code like you have quite a few usages of `tueNews` (mind the `e`).

Ah now I see what you were getting at. I couldnt find any tueNews in the code. Checked on the current branch though.

steffenkleinle:
> > Jep. Seems in the code like you have quite a few usages of `tueNews` (mind the `e`).
> 
> Ah now I see what you were getting at. I couldnt find any tueNews in the code. Checked on the current branch though.

Ah sorry, looks like I looked at an old revision. Than all good, thanks for checking!

hannaseithe:
Ok, I have added pagination and filtering now to `all_news` endpoint. So maybe @MizukiTemma and @jonbulz could re-review? Its a bit of a merry-go-round this PR. Everyone gets to review, everyone gets to code :D

In the meantime I will check if I am happy with my change requests being applied and approve accordingly

steffenkleinle:
> Ok, I have added pagination and filtering now to `all_news` endpoint. So maybe @MizukiTemma and @jonbulz could re-review? Its a bit of a merry-go-round this PR. Everyone gets to review, everyone gets to code :D
> 
> In the meantime I will check if I am happy with my change requests being applied and approve accordingly

Would you mind adjusting the endpoint to just `news` instead of `all_news` since this is going to be the standard news endpoint (and filtering allows to get not all news)?

hannaseithe:
> > Ok, I have added pagination and filtering now to `all_news` endpoint. So maybe @MizukiTemma and @jonbulz could re-review? Its a bit of a merry-go-round this PR. Everyone gets to review, everyone gets to code :D
> > In the meantime I will check if I am happy with my change requests being applied and approve accordingly
> 
> Would you mind adjusting the endpoint to just `news` instead of `all_news` since this is going to be the standard news endpoint (and filtering allows to get not all news)?

Since you are asking so nicely, I will do that :D

hannaseithe (approved):
I have approved now all up to pagination and filter

hannaseithe:
> Would you mind adjusting the endpoint to just `news` instead of `all_news` since this is going to be the standard news endpoint (and filtering allows to get not all news)?

Sorry I missed the filtering question. Filter will work like this: `/?source=local&source=amalNews` will get you local and amalNews (and not tuNews)

MizukiTemma:
@hannaseithe 

Thank you for bringing the PR forward and for the re-review ❤️ I played around and it works as expected 😻

MizukiTemma:
@jonbulz 
Thank you so much for taking care of this PR while I was away 🙏 

Could you re-review?

integreat_cms/api/v3/news.py
```diff
@@ -0,0 +1,81 @@
...
+
+    :param request: Django request
+    :param region_slug: slug of a region
+    :param language_slug: language slug
+    :return: JSON list of news items merged across all sources defined in :data:`~integreat_cms.news_managers.registry.CHOICES`
+    """
+    channel = request.GET.get("channel", "all")
+
+    result = []
+    logger.debug(registry.CHOICES)
```
jonbulz, line 62: do we want to keep this in prod or was this for local debugging?

jonbulz (approved):
I took the liberty of just committing the changes that I think are not something that we have to debate (feel free to correct me if you disagree):
- remove orphaned `news/all-news/` endpoint
- fix naming inconsistencies
- tighten tests
- sanitize the `page_size` request parameter

If you don't like any of that, please accept my apology and I'll revert the commit, but I felt like this would be faster. Except for removing the debugging statement I commented on, I think this is ready to merge. But it doesn't hurt to leave it in, and I wasn't sure if it's intentional, so I left it there for now and I'll approve already 👍

integreat_cms/api/v3/news.py
MizukiTemma: I would keep it for future debagging, it won't harm 😁

Issue #4089, opened 2026-01-20
Contact creation failure clears selected location instead of preserving it


juliankehne:
### Describe the Bug

This behavior is also on Admin and probably has been around since the introduction of contacts.
Now it is more noticeable due to stricter phone number rules.

When creating a new contact and creation fails due to:

- invalid phone numbers (or mobile)
- having no Area of responsibility if there already is a primary contact for the location
- something else

the selected location will be cleared and the user has to reselect the location.

### Steps to Reproduce

Reproducibility: Always

1. Go to 'Contacts'
2. Click on 'Create'
3. Select a location
4. Enter an invalid phone number e.g. "abc"
5. Enter something in 'area of responsibility'
6. Click 'Create'

### Expected Behavior

Creation fails due to the invalid phone number, but the selected location is saved and does not have to be reselected

### Actual Behavior

After the creation of the contact fails the location is empty again

### Additional Information

<details><summary>Traceback</summary>

```
```

</details>

PR #4198, opened 2026-03-17
Prevent selected POI from being cleared when contact creation fails

MizukiTemma:
### Short description

This PR fixes the bug that the POI user has selected gets cleared when contact creation fails

### Proposed changes

- Store which POI was selected
- Hand it over to the ajax POI form so the POI is shown after redirect

### Side effects

- Shold be none

### Resolved issues

Fixes: #4089 

### DIFF
--- a/integreat_cms/cms/views/contacts/contact_form_view.py
+++ b/integreat_cms/cms/views/contacts/contact_form_view.py
@@ -161,6 +161,12 @@ class ContactFormView(TemplateView, ContactContextMixin):
             additional_instance_attributes={"region": region},
         )
 
+        selected_poi_id = int(request.POST.get("location"))
+        # request.POST.get("location") == -1 if no POI is selected
+        selected_poi = (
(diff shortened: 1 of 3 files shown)


integreat_cms/cms/views/contacts/contact_form_view.py
```diff
@@ -161,6 +161,11 @@ def post(
             additional_instance_attributes={"region": region},
         )
 
+        selected_poi_id = int(request.POST.get("location"))
+        selected_poi = (
+            POI.objects.get(id=selected_poi_id) if selected_poi_id > 0 else None
```
hannaseithe, line 167: I stumbled across this in terms of readability, maybe a comment that request.POST.get("location") == -1 means nothing is selected would help

MizukiTemma: Done 😸

PR #4206, opened 2026-03-19
avoid UniqueViolation upon saving a new content translation version

### LINKED ISSUE #4073: `psycopg.errors.UniqueViolation: duplicate key value violates unique constraint "pagetranslation_unique_version"` when saving a translation
### Describe the Bug

Sometimes this error occurs when saving a content. It has been observed lately more often. Service Team could not use Bulk MT due to this error.

This bug does not occur always. Trigger conditions are unclear. 

However, there is a tendency in bulk MT: it seems not to be (completely) random. Looks like, once a page faces this error, it keeps suffering from it, regardless of which language is chosen for a bulk MT next time.

Interesting was that MT worked fine as only this suffering page alone is selected for a bulk MT action or it is trasnlated into multiple languages on the page form (by translation side box).

For context there is a thread here: https://chat.tuerantuer.org/digitalfabrik/pl/senf6ufdzt8iud8gnbhr989w1a

### Steps to Reproduce

---- What I observed
1. Go to a location (A)
2. Change something
3. Save it
4. Repeat this relatively repidly 
5. Face the error
6. Return to the locations list, open the location (A)
7. Change something and save
8. See the error again 

---- What Service Team experienced --------------
1. Go to the page tree
2. Select pages (including one or more "suffering" page(s))
3. Use bulk MT
4. See the error

### Expected Behavior

No error, MT is not blocked

### Actual Behavior

Error, MT blocked

### Additional Information

```
`Dec 15 07:45:54 ERROR   django.request - 500 Internal Server Error: /kaufbeuren/pages/hr/machine-translate/
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/django/db/backends/utils.py", line 89, in _execute
    return self.cursor.execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/psycopg/cursor.py", line 97, in execute
    raise ex.with_traceback(None)
psycopg.errors.UniqueViolation: duplicate key value violates unique constraint "pagetranslation_unique_version"
DETAIL:  Key (page_id, language_id, version)=(9507, 22, 4) already exists.

  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
               ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, callback_args, callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/django/views/generic/base.py", line 104, in view
    return self.dispatch(request, *args, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/django/contrib/auth/mixins.py", line 109, in dispatch
    return super().dispatch(request, args, *kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/django/views/generic/base.py", line 143, in dispatch
    return handler(request, args, *kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/integreat_cms/cms/views/bulk_action_views.py", line 226, in post
    api_client.translate_queryset(to_translate, language_node.slug)
  File "/usr/lib/python3.11/contextlib.py", line 81, in inner
    return func(args, kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/integreat_cms/core/utils/machine_translation_api_client.py", line 161, in translate_queryset
    self.invoke_translation_api()
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/integreat_cms/google_translate_api/google_translate_api_client.py", line 130, in invoke_translation_api
    self.save_translation(content_object, data)
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/integreat_cms/core/utils/machine_translation_api_client.py", line 321, in save_translation
    content_translation_form.save()
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/integreat_cms/cms/forms/machine_translation_form.py", line 151, in save
    self.instance = super().save(commit, foreign_form_changed)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/integreat_cms/cms/forms/custom_content_model_form.py", line 143, in save
    result = super().save(commit=commit)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/integreat_cms/cms/forms/custom_model_form.py", line 133, in save
    return super().save(commit=commit)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/django/forms/models.py", line 542, in save
    self.instance.save()
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/integreat_cms/cms/models/pages/page_translation.py", line 435, in save
    super().save(*args, kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/integreat_cms/cms/models/abstract_content_translation.py", line 668, in save
    super().save(args, *kwargs)
  File "/opt/integreat-cms/.venv/lib/python3.11/site-packages/django/db/models/base.py", line 814, in save
    self.save_b  [truncated from 6726 characters]`
```

</details>

jonbulz:
### Short description

Fix server error of concurrent requests saving translations with the same version number, e.g. during bulk MT or rapid consecutive updates

### Proposed changes

- wrap `save()` calls in `save_new_version_with_retry` in `custom_content_model_form` and `publication_status`
- catch `IntegrityError` in wrapper and bump version number based on current database value
- the fix is deliberately not in the model's save() because the XLIFF import relies on IntegrityError propagation to detect conflicting file imports

### Side effects

- None

There are implementation suggestions in a comment on the issue that suggest replacing `self.instance.version += 1` by a database lookup in [custom_content_model_form](https://github.com/digitalfabrik/integreat-cms/blob/15214c62fcc49f6dbf987e5752f0d1b1197fe247/integreat_cms/cms/forms/custom_content_model_form.py#L143). However, this would only partially solve the issue, because concurrent requests could still query the same MAX(version) and trigger the unique violation. The time-of-check-to-time-of-use gap would be smaller (DB query to save vs. request start to save), but the underlying problem remains. Catching the collision and retrying in a loop completely avoids this issue, unless there are so many concurrent requests that the retry attempts get exhausted, which is extremely unlikely under our expected load

### How to test

Since this fixes a race condition, there is no deterministic way to test. Follow the Steps to Reproduce from the issue and see that the error does not occur anymore.

### Resolved issues

Fixes: #4073 

### DIFF
--- a/integreat_cms/cms/utils/content_translation_utils.py
+++ b/integreat_cms/cms/utils/content_translation_utils.py
@@ -25,6 +25,8 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
+MAX_RETRY = 3  # number of retry attempts when hitting the translation version number race condition
+
 
 def save_new_version_with_retry(
     translation: AbstractContentTranslation,
@@ -48,7 +50,7 @@ def save_new_version_with_retry(
     foreign_field = translation.foreign_field()
     foreign_id = getattr(translation, f"{foreign_field}_id")
 
-    for attempt in range(3):
+    for attempt in range(MAX_RETRY):
         try:
             with transaction.atomic():
                 return save_fn()
--- a/integreat_cms/cms/utils/content_translation_utils.py
+++ b/integreat_cms/cms/utils/content_translation_utils.py
@@ -6,7 +6,6 @@ from __future__ import annotations
 
 import logging
 import operator
-from collections.abc import Callable
 from functools import reduce
 from typing import Any, TYPE_CHECKING
 
@@ -19,6 +18,8 @@ from .content_utils import clean_content
 from .internal_link_utils import get_public_translation_for_link
 
 if TYPE_CHECKING:
+    from collections.abc import Callable
+
     from ..models import User
     from ..models.abstract_content_translation import AbstractContentTranslation
 
--- a/integreat_cms/cms/utils/content_translation_utils.py
+++ b/integreat_cms/cms/utils/content_translation_utils.py
@@ -6,11 +6,13 @@ from __future__ import annotations
 
 import logging
 import operator
+from collections.abc import Callable
 from functools import reduce
-from typing import TYPE_CHECKING
+from typing import Any, TYPE_CHECKING
 
 from django.conf import settings
-from django.db.models import Q
+from django.db import IntegrityError, transaction
+from django.db.models import Max, Q
 from linkcheck.models import Url
 
 from .content_utils import clean_content
@@ -21,6 +23,55 @@ if TYPE_CHECKING:
     from ..models.abstract_content_translation import AbstractContentTranslation
 
 logger = logging.getLogger(__name__)
+
+
+def save_new_version_with_retry(
+    translation: AbstractContentTranslation,
+    save_fn: Callable[[], Any],
+) -> Any:
+    """
+    Call *save_fn* inside a savepoint, retrying with a recomputed version
+    number when a :class:`~django.db.IntegrityError` on the
+    ``unique_version`` constraint is raised (up to 3 attempts).
+
+    This handles race conditions where concurrent requests both try to
+    insert a new version with the same ``(foreign_object, language, version)``
+    tuple.
+
+    :param translation: The content translation whose ``version`` field will
+        be corrected on conflict
+    :param save_fn: A zero-argument callable that persists *translation*
+        (e.g. ``lambda: form.save(commit=True)`` or ``translation.save``)
+    :return: Whatever *save_fn* returns
+    """
+    foreign_field = translation.foreign_field()
+    foreign_id = getattr(translation, f"{foreign_field}_id")
+
+    for attempt in range(3):
+        try:
+            with transaction.atomic():
+                return save_fn()
+        except IntegrityError as e:
+            if "unique_version" not in str(e) or attempt >= 2:
+                raise
+            # Recompute the next version from the database
+            max_version = (
+                (
+                    type(translation)
+                    .objects.filter(
+                        **{foreign_field: foreign_id, "language": translation.language}
+                    )
+                    .aggregate(max_v=Max("version"))["max_v"]
+                )
+                or 0
+            )
+            translation.version = max_version + 1
+            logger.info(
+                "Version conflict on %r, retrying with version %d",
+                translation,
+                translation.version,
+            )
+    return None  # unreachable, but satisfies type checkers
 
 
 def update_links_to(
--- a/integreat_cms/cms/utils/content_translation_utils.py
+++ b/integreat_cms/cms/utils/content_translation_utils.py
@@ -55,7 +55,7 @@ def save_new_version_with_retry(
             with transaction.atomic():
                 return save_fn()
         except IntegrityError as e:
-            if "unique_version" not in str(e) or attempt >= 2:
+            if "unique_version" not in str(e) or attempt >= MAX_RETRY - 1:
                 raise
             # Recompute the next version from the database
             max_version = (
--- a/integreat_cms/cms/utils/content_translation_utils.py
+++ b/integreat_cms/cms/utils/content_translation_utils.py
@@ -107,8 +107,19 @@ def update_links_to(
             outdated_content_translation.create_new_version_copy(user)
         )
         fixed_content_translation.content = new_content
-        outdated_content_translation.links.all().delete()
-        fixed_content_translation.save()
+        try:
+            with transaction.atomic():
+                outdated_content_translation.links.all().delete()
+                save_new_version_with_retry(
+                    fixed_content_translation, fixed_content_translation.save
+                )
+        except IntegrityError:
+            logger.exception(
+                "Could not update links in %r referencing %r",
+                outdated_content_translation,
+                content_translation,
+            )
+            continue
 
         logger.debug(
             "Updated links to %s in %r",
--- a/integreat_cms/cms/views/utils/publication_status.py
+++ b/integreat_cms/cms/views/utils/publication_status.py
@@ -11,6 +11,7 @@ from django.utils.translation import ngettext, ngettext_lazy
 
 from ...constants import status
 from ...models import Language
+from ...utils.content_translation_utils import save_new_version_with_retry
 from ...utils.stringify_list import iter_to_string
 
 if TYPE_CHECKING:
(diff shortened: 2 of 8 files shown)


integreat_cms/cms/utils/content_translation_utils.py
```diff
@@ -7,22 +7,74 @@
...
+    :param translation: The content translation whose ``version`` field will
+        be corrected on conflict
+    :param save_fn: A zero-argument callable that persists *translation*
+        (e.g. ``lambda: form.save(commit=True)`` or ``translation.save``)
+    :return: Whatever *save_fn* returns
+    """
+    foreign_field = translation.foreign_field()
+    foreign_id = getattr(translation, f"{foreign_field}_id")
+
+    for attempt in range(3):
```
MizukiTemma:
```suggestion
    MAX_RETRY = 3
    for attempt in range(MAX_RETRY):
```

A nitpick 🙈

MizukiTemma (approved):
Thank you for the analyse and PR 😸 No side effect was observed.

I could not reproduce the problem and therefore cannt test this PR against ir, but the changes make sense and the correct version number was calclated after a fail as the version number was manipulated.👍

integreat_cms/cms/utils/content_translation_utils.py
jonbulz: I made MAX_RETRY a module level constant now :)

MizukiTemma: Nice! Thank you for update 😸

integreat_cms/cms/utils/content_translation_utils.py
```diff
@@ -7,21 +7,75 @@
...
+    """
+    foreign_field = translation.foreign_field()
+    foreign_id = getattr(translation, f"{foreign_field}_id")
+
+    for attempt in range(MAX_RETRY):
+        try:
+            with transaction.atomic():
+                return save_fn()
+        except IntegrityError as e:
+            if "unique_version" not in str(e) or attempt >= 2:
```
hannaseithe, line 58:
```suggestion
            if "unique_version" not in str(e) or attempt = MAX_RETRY - 1:
```
From Claude

integreat_cms/cms/views/utils/publication_status.py
```diff
@@ -51,7 +52,7 @@ def change_publication_status(
                         translation.all_versions.filter(status=status.PUBLIC).update(
                             status=status.DRAFT,
                         )
-                    translation.save()
+                    save_new_version_with_retry(translation, translation.save)
                     successful.append(translation.title)
             else:
```
hannaseithe, line 57: This has not been introduced by this PR, but might still be a good place to fix: 

We do not catch the errors raise from save_new_version_with_retry. Therefore we end up with partially executed state and a 500 if we fail on retry. Maybe both a try/catch for each single translation and inside the try a transaction.atomic() would be an acceptable way to handle the error here?

hannaseithe (changes requested):
Thank you.

I have asked Claude if there are other places we should also use the new function. Maybe you have considered them all already, but I thought it was worth checking. This is the answer by Claude:

There are three more places where a new version is saved after create_new_version_copy
(or directly after version += 1 / pk = None):

| Location                                                              | Concurrent risk? |
|-----------------------------------------------------------------------|------------------|
| xliff/utils.py:459                                                    | Already has a try/except IntegrityError with transaction.atomic() — handled, but doesn't retry with a recomputed version |
| cms/models/abstract_content_translation.py:638 (replace_urls)        | Called e.g. during link rewriting — no protection |
| cms/utils/content_translation_utils.py:111 (update_links_to area)    | Called after MT — no protection, and this is exactly the concurrent scenario the PR is targeting |
| core/management/commands/fix_internal_links.py:192                   | Management command — runs serially, low risk |
| core/management/commands/update_link_text.py:130                     | Management command — runs serially, low risk |

The most important gap is content_translation_utils.py:111 — that's the MT bulk-save path
where the race condition is most likely to occur (it's even the scenario mentioned in the PR
commit message). replace_urls is also used during MT and link rewriting, so it's exposed too.

The management commands are lower priority since they run serially and aren't triggered by
concurrent web requests.

integreat_cms/cms/views/utils/publication_status.py
jonbulz: I'm not entirely sure if you mean the same thing as I, but wrapping all actions in a transaction and displaying a meaningful error message to users if one of the database updates failed is definitly a good idea, thanks 😊

jonbulz:
> Thank you.
> 
> I have asked Claude if there are other places we should also use the new function. Maybe you have considered them all already, but I thought it was worth checking. This is the answer by Claude:
> 
> There are three more places where a new version is saved after create_new_version_copy (or directly after version += 1 / pk = None):
> Location 	Concurrent risk?
> xliff/utils.py:459 	Already has a try/except IntegrityError with transaction.atomic() — handled, but doesn't retry with a recomputed version
> cms/models/abstract_content_translation.py:638 (replace_urls) 	Called e.g. during link rewriting — no protection
> cms/utils/content_translation_utils.py:111 (update_links_to area) 	Called after MT — no protection, and this is exactly the concurrent scenario the PR is targeting
> core/management/commands/fix_internal_links.py:192 	Management command — runs serially, low risk
> core/management/commands/update_link_text.py:130 	Management command — runs serially, low risk
> 
> The most important gap is content_translation_utils.py:111 — that's the MT bulk-save path where the race condition is most likely to occur (it's even the scenario mentioned in the PR commit message). replace_urls is also used during MT and link rewriting, so it's exposed too.
> 
> The management commands are lower priority since they run serially and aren't triggered by concurrent web requests.

Doesn't hurt to use this in more places, but not `xliff/utils.py`. The unique version constraint is used there to check for and reject duplicate uploads

hannaseithe:
Hello @jonbulz  I would say, pretty clean implementation of the change requests. Thank you very much :)

hannaseithe (approved):

jonbulz:
Merged into develop, thus this will not be part of the upcoming release @juliankehne 
This is intentional to not overload the release candidate with last-minute additions

PR #4214, opened 2026-03-24
Fix 404 when changing entries-per-page from a non-first page

### LINKED ISSUE #3249: Changing "number of entries per page
### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```

</details>

dkehne:
Closes #3249

## Summary

When a user is on page 2+ of a paginated list and changes the entries-per-page size, the URL keeps `page=2` while the new page size may result in fewer total pages — causing a 404.

**Fix**: Pre-compute a `base_url` with `page=1` reset before building the size option URLs, so changing the page size always navigates to page 1.

## Test Plan

- [ ] Go to Languages (or any paginated list), navigate to page 2
- [ ] Change the entries-per-page selector to 50, 100, or 500
- [ ] Verify it navigates to page 1 with the correct size instead of a 404

### DIFF
--- a/integreat_cms/cms/templates/pagination.html
+++ b/integreat_cms/cms/templates/pagination.html
@@ -5,6 +5,7 @@
 <!-- in general there are links for the 2 previous and next pages -->
 <!-- plus additional links to start, end, immediate next and immediate previous -->
 {% with num_pages=chunk.paginator.num_pages chunk_size=request.GET.size url=request.get_full_path %}
+    {% add_queries url 'page' 1 as base_url %}
     <div class="pagination">
         <div>
             <label for="chunk-size">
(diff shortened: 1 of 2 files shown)


dkehne:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

MizukiTemma (commented):
Looks goold 😸

MizukiTemma (approved):

jonbulz:
Actually, this is already fixed by #3963 , because it catches an `EmptyPage` exception and [redirects to the last valid page](https://github.com/digitalfabrik/integreat-cms/blob/8464b6b2f2bd3712fa5afe5e9179292576401246/integreat_cms/cms/views/mixins.py#L181-L182) 
I was not aware of this issue then, sorry about that. So the difference with this PR essentially becomes
- Currently: User gets silently redirected to the last page
- With these changes: User always gets redirected to page 1 when changing page size

I don't know if this is worth the added template complexity. Probably a UX decision, though

dkehne:
I would say, let's keep it simple and close here when https://github.com/digitalfabrik/integreat-cms/pull/3963 is solving the problem.

MizukiTemma:
@jonbulz @dkehne 

#3963 was closed long time ago and the bug still exists, right? Then don't we anyway have to add a fix?

jonbulz:
I could not reproduce the bug

MizukiTemma:
I can still reproduce the bug for the POI category list and offer template list 👀

jonbulz:
True, POI category and offer template use `ModelListView` instead of our `PaginationMixin` and thus are not covered by the existing fix. I'll create a different PR that will include the existing fix in the `ModelListView` pagination

jonbulz:
See #4229

MizukiTemma:
@jonbulz 

Sorry, I thought it was the issue page here 🙈 closed again

PR #4219, opened 2026-03-26
Prevent reset button activation on Enter key press in search

### LINKED ISSUE #4187: Pressing 'Enter' when searching in a list view resets the search query
### Describe the Bug

In **some** list views, if the search query is not empty (a user already searched for something previously) pressing 'Enter' when the search field is selected resets the query.

My guess is pressing 'Enter' is selecting the "Reset Search" button, X icon, but it could be something else.

Reproducible in the following list forms:

- languages
- organizations
- locations
- push notification
- region
- region condition
- users

### Steps to Reproduce

1. Go to contacts list view
2. Type anything into the search input
3. Search (Click the Search button or press Enter)
4. See correct "?search_query=your_search_input" in the URL
5. Click the search input field again
6. Press 'Enter' to seach again (do not click the search button)
7. See "?search_query=" in URL

### Expected Behavior

It searches for whatever was in the search input after pressing 'Enter' and it should behave the same as if the search button was clicked.

### Actual Behavior

When specifically searching by pressing 'Enter' the search query is cleared resulting in a URL that ends with "?search_query=" even though there was input inside the search field.

### Additional Information

```
```

</details>

### Related Issues

https://github.com/digitalfabrik/integreat-cms/pull/4164/changes

### Summary of discussion and updates to the description

Specifying where this bug appears as its not in every list view search as initially written.

dasgoutam:
### Short description
Fixes the search bar resetting to empty when pressing Enter after modifying the search input a second time.

### Proposed changes

The bug occurs when the user searches, then edits the query and searches again(by pressing Enter), the search resets to empty instead of filtering the table with the new value.
- Handle Enter before the suggestions.length === 0 check in search-query.ts so it is always intercepted.
- Add event.preventDefault() to the Enter keydown handler in search-query.ts to prevent the native Enter event from carrying over to the newly loaded page and activating the reset button
- Move the search-submit-btn click outside the suggestion check so Enter always submits regardless of whether a suggestion is highlighted

### Side effects

- None

### How to test

- Go to any list view (e.g. Regions)
- Type a search query and press Enter — table filters correctly
- Edit the query and press Enter again — table filters with the new value instead of resetting
- Should behave as a normal search bar over any number of iterations

### Resolved issues

Fixes: #4187 

### DIFF
--- a/integreat_cms/cms/templates/_search_input.html
+++ b/integreat_cms/cms/templates/_search_input.html
@@ -21,7 +21,8 @@
         </ul>
     </div>
     {% if search_query %}
-        <button id="search-reset-btn"
+        <button type="button"
+                id="search-reset-btn"
                 title="{% translate "Reset search" %}"
                 class="bg-gray-200 hover:bg-gray-300 text-gray-800 py-2 px-3">
             <i icon-name="x" class="w-5"></i>
--- a/integreat_cms/release_notes/current/unreleased/4187.yml
+++ b/integreat_cms/release_notes/current/unreleased/4187.yml
@@ -1,0 +1,2 @@
+en: Fix enter key resets search query
+de: Fix Enter-Taste setzt Suchanfrage zurück


jonbulz (changes requested):
This fix works, great work!
I do believe there is a simpler solution, though 🙈 
If we add a `type="button"` in [_search_input.html:25](https://github.com/digitalfabrik/integreat-cms/blob/8464b6b2f2bd3712fa5afe5e9179292576401246/integreat_cms/cms/templates/_search_input.html#L25), the native form should handle this correctly. The root cause of this issue is apparently that, since no `type` attribute is set on the reset button, it defaults to `type="submit"`. This is a problem because once everything is wrapped in a `<form>`, pressing enter activates the first `submit` button in DOM order, which then is the reset button. 
Once we assign the correct `button` type, it does not participate in form submission any more and does not get in the way of our JS search query logic.
I only found this when reviewing your PR, because I noticed that I could not reproduce the initial bug on every list view, i.e. `Events` was already working as expected before the fix. This is because `event_list.html` does not use `search_input_form` and thus does not wrap the reset button in a `<form>`, which 'accidentaly' avoids this issue.

dasgoutam:
Thanks @jonbulz for the suggestion and explanation. That is indeed a much simpler and cleaner fix to the problem and works perfectly 😀

jonbulz (approved):
Great, thank you very much! 😊

MizukiTemma (approved):
Thank you so much 😸 Wroks as expected 👍 

Just add a release note (as this fix is visible for users) and rebase before merge ;)

PR #4064, opened 2025-12-11
Make primary contact only editable by the contact fields if user is in POI form

MizukiTemma:
### Short description

This PR makes it possible for users in the POI form only in "Contact details" to create or edit the primary contact, to avoid two side boxes ("Contact details" and "Related contacts") conflicts each other for the primary contact.

### Proposed changes

- Make `area_of_responsibility` mandatory for contact creation in "Related contacts"
- For a new (not yet saved) POI: Add 5 fields (`area_of_responsibility`, `name`, `email`, `phone_number`, `mobile_phone_number`, and `website`) to the POI form to validate `area_of_responsibility` before saving POI and POI translation so no input goes lost in case `area_of_responsiblity` is not given (but POI and POI translations were saved) and user gets redirected to the POI form.

### Side effects

- Less confusion?

### Faithfulness to issue description and design

There are no intended deviations from the issue and design.

### How to test

- There is no chance to create a POI without `area_of_responsibility` in "Related contacts"
- An error is shown that `area_of_responsibility` is mandatory 
- When creating a new (not yet saved) POI: if `area_of_responsibility` is left empty but at lest one of the follower fields is filled, user get an warning, and input in this box and the POI form won't go lost
- It should be possible as before to create/save POI without any new contact or information for primary contact.
- The primary contact is updated accordingly when anything in "Contact details" is changed.

### Resolved issues

Fixes: #4050 

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #4226, opened 2026-03-31
Fix failing ajax POI creation

### LINKED ISSUE #4199: POI creation per ajax fails if E-mail adress, phone number or website is given
### Describe the Bug

Users can create a new POI on spot while they are creating/editing a contact. This ajax POI form seems to have a problem.

### Steps to Reproduce

1. Go to "contacts", click "Create contact"
2. Give a location name which does not exist in the region
3. Click "Create location <location name>
4. Fulfill "Name of the location", "Street and house number", "Postal code", "City" and "Country"
5. Fulfill at lease one of "Email adress", "Phone number", "Website"
6. Select a category
7. Click "Publish"
8. See error "An error occurred."
### Expected Behavior
A POI is created (as long as given values are valid)

### Actual Behavior

No chance to create a POI as long as at least one of "Email adress", "Phone number", "Website" is filled

### Additional Information

```
```

</details>

### Related Issues

#4064

MizukiTemma:
### Short description

This PR fixes the bug that POI creation per ajax form in the event/contact form is failing if phone number, e-mail or website is given.

### Proposed changes

- Make the input fields for phone numebr, email address and website of the ajax form match to `primary_phone_number`, `primary_email` and `primary_website` of `POIFrom` instead `phone_number`, `email` and `website`

* `primary_phone_number`, `primary_email` and `primary_website` are there for the primary contact
image

- `phone_number`, `email` and `website` are there for cases user wants to create a non-primary contact at the same time with creation of a contact
image

- Refactor handling of primary contact into a new function `generate_primary_contact_from_poi` to implement the same mechanism in `POIFormAjaxView` and `POIFormView`
- No release note as this is a fix for #4064 which is not yet released into the prod system.

### Side effects

- should be none

### How to test

- Create a new POI in the contact/event form and check the primary contact is successfully created
- Test the POI form works as before and check the primary contact is successfully created
- Test creation of non-primary contact in the POI form works as before

### Resolved issues

Fixes: #4199 

### DIFF
--- a/integreat_cms/cms/views/pois/poi_form_ajax_view.py
+++ b/integreat_cms/cms/views/pois/poi_form_ajax_view.py
@@ -10,6 +10,7 @@ from django.views.generic import TemplateView
 from ...forms import POIForm, POITranslationForm
 from ...models import Language
 from ...models.pois.poi import get_default_opening_hours
+from ..utils.contact_utils import generate_primary_contact_from_poi
 from .poi_context_mixin import POIContextMixin
 
 if TYPE_CHECKING:
@@ -86,6 +87,10 @@ class POIFormAjaxView(TemplateView, POIContextMixin):
             changed_by_user=request.user,
         )
 
+        phone_number = poi_form.data.get("primary_phone_number")
+        email = poi_form.data.get("primary_email")
+        website = poi_form.data.get("primary_website")
+
         if not poi_form.is_valid() or not poi_translation_form.is_valid():
             return JsonResponse(
                 data={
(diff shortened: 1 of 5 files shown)


seluianova:
@MizukiTemma maybe I'm missing something, but when I follow the steps from the issue description, I get the same error on this branch... 🤔
image

**UPD**. ah wait, I understood what the problem was...

**UPD-2**: so, the problem I faced was a separate issue. I created another ticket for that: https://github.com/digitalfabrik/integreat-cms/issues/4242

seluianova (approved):
tested, lgtm! 👍

integreat_cms/cms/views/pois/poi_form_ajax_view.py
```diff
@@ -86,23 +87,38 @@ def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
...
                 data={
                     "success": False,
                 },
             )
 
-        poi_translation_form.instance.poi = poi_form.save()
-        poi_translation_form.save()
+        poi = poi_form.save()
+        poi_translation_form.instance.poi = poi
+        poi_translation_form = poi_translation_form.save()
```
jonbulz: I think this is misleading. The `save` method of a form does return a Model instance, in this case a `POITranslation`. It should not override the form variable. Maybe do `poi_translation_instance = poi_translation_form.save()`, similar to how it is done in the regular POI view? 
Also, the regular POI view passes `foreign_form_changed=poi_form.has_changed()` as an argument to `save`. I don't think it's very important, but we maybe should also include this here to be consistent. Or is there a reason you left it out?

jonbulz (changes requested):
I think you should wrap POI save + translation save + contact generation in a `transaction.atomic()` block. Otherwise, if one of the steps fails, it could lead to an inconsistent database state, e.g. because the POI and translation are already committed, but the contact generation gets aborted.
I also have concerns regarding the way you save the form instance, see my additional comment.
Other than that, looks good 👍

integreat_cms/cms/views/pois/poi_form_ajax_view.py
MizukiTemma: Thank you for nice improvement suggestions 😺 They're now applied.

jonbulz (approved):
Nice, looks great now! 👍

PR #4229, opened 2026-04-07
add get_safe_page helper method for ModelListView

### LINKED ISSUE #3249: Changing "number of entries per page

jonbulz:
### Short description

Changing the page size in a list view can result in a 404 exception if the page does not exist with the new entries per page setting. `PaginationMixin` (introduced in #3963 ) handles this gracefully, but our existing `ModelListViews` don't use `PaginationMixin`. This PR introduces a shared helper method to handle `EmptyPage` exceptions in all list views.

### Proposed changes

- add `get_safe_page` helper method and use it in `PaginationMixin` and `ModelListView`

### Side effects

- None

### How to test

Ensure that no 404 occurs on any list view when changing the page size

### Resolved issues

Fixes: #3249 

### DIFF
--- a/integreat_cms/cms/views/list_views.py
+++ b/integreat_cms/cms/views/list_views.py
@@ -11,7 +11,11 @@ from django.conf import settings
 from django.contrib.auth.mixins import PermissionRequiredMixin
 from django.views.generic.list import ListView
 
-from .mixins import ModelConfirmationContextMixin, ModelTemplateResponseMixin
+from .mixins import (
+    get_safe_page,
+    ModelConfirmationContextMixin,
+    ModelTemplateResponseMixin,
+)
 
 if TYPE_CHECKING:
     from typing import Any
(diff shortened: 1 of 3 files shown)


MizukiTemma (approved):
Looks good 😸 Thank you for speedy fix!

We can add a release note, this fix is visible for users 📝

jarlhengstmengel:
Bit unsure what to test, I can't reproduce the originall bug on the test system, is that bug fixed somewhere else and this only making sure that it never arises again? :sweat_smile:

MizukiTemma:
@jarlhengstmengel 
With which list did you try?
The bug is already fixed for the language list (which is mentioned in the issue description) but still exists in the POI category list and offer template list (mentioned in the issue conversation).

hannaseithe (approved):
Thank you for this fix, this works for poi categories. 

Just for my own understanding: Why are we not using PaginationMixin for the ModelListView?

jonbulz:
`ModelListView` inherits Django's built in `ListView`, which already handles pagination. We need `PaginationMixin` for our other list views, because they only inherit `TemplateView` <- `View`, which has no implementation of pagination yet. But simply pluging our `PaginationMixin` into the `ModelListView` also doesn't work because the pagination logic follows different signatures

PR #4231, opened 2026-04-08
fix bumpver for rc

PeterNerlich:
### Short description

Our rc pipeline failed at the bumpver step

Previously, the conditional in the bumpver step was only necessary for alpha builds from develop. On main, the new version was always a jump in the major version, there was nothing to compare. For alpha builds, the conditional is needed for when there just was a release of a new major and the current build was about to be uploaded as the first alpha build after that, i.e. the first alpha build containing the new major release number. But the alpha builds are uploaded to test.pypi.org, where there are only alpha builds, never the major releases. The version in the code never exactly matched the last available version number there.

Our newly introduced rc builds live on the proper pypy.org, as pre-releases among the real ones. Now the last available version can be the same one mentioned in the code, and we did not treat this correctly historically since it didn't matter.

### Proposed changes

- Properly handle the case when the versions are equal: Only exit with a non-zero code when a new major release needs to be set before bumping the rc number

### Side effects

- 

### How to test

Merge it and see whether the pipeline passes this time

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -417,7 +417,7 @@ jobs:
             echo "Current production version: ${CURRENT_VERSION}"
             # Bump version to current alpha version if it is newer
             # Attention: exit(True) in Python means exit(1) which is False in Bash :)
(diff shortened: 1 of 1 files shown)


MizukiTemma (approved):
Thank you 🚀

jonbulz (approved):
👍

PR #4245, opened 2026-04-16
Fix page path in API response

### LINKED ISSUE #4220: Children responses contain wrong page paths
### Describe the Bug
The CMS redirects the app to wrong paths (omitting the first level slug) when opening pages on the third level.

### Steps to Reproduce

1. Open https://cms.integreat-app.de/api/v3/hochtaunuskreis/de/children/?depth=2&url=/hochtaunuskreis/de/willkommen/asyl-und-gefl%C3%BCchtete/asylsozialberatung
2. The `path` returned in the response is `/hochtaunuskreis/de/asyl-und-geflüchtete/asylsozialberatung/` (the same path is used in the `url` attribute.  The `/willkommen` part in the beginning is missing.

### Expected Behavior
The CMS should not redirect the app to the path w/o the top level slug -> `"path": "/hochtaunuskreis/de/willkokmmen/asyl-und-geflüchtete/asylsozialberatung/"`

### Actual Behavior
The CMS redirects the app to a path without the top level slug -> `"path": "/hochtaunuskreis/de/asyl-und-geflüchtete/asylsozialberatung/"`

### Screenshot

Image

### Additional Information
* https://chat.tuerantuer.org/digitalfabrik/pl/h6ufjt747fnrjgsnrjqhpnihor
* https://chat.tuerantuer.org/digitalfabrik/pl/ks1dbkta7tdxpnobctbp8qb1fr

MizukiTemma:
### Short description

This PR fixes the bug that slugs of ancestors (higher than parent) are missing in the page pathes in the `children` endpoint response.

### Proposed changes

- Include not only the parent but also higher ancestors into the query `pages` in `children`
- Exclude them but in the loop with `transform_page` so they are not included in the response

### Side effects

- Should be none
- There is probably a more beautiful/thorough solution. This is but in my opinion less invasive and less risk of breaking something else while the added loop is small. We could adjust `cache_tree_dict` or use `get_ancestors()` instead of `get_cached_ancestors()`, but the `cache_tree` is calling `cache_tree_dict` and being used in several places, and `get_ancestors()` may affect loading time negatively.

### How to test

- Start the CMS in `develop` branch
- Create a page tree with more than 2 levels
image

- Call the endpoint `children` (example: http://localhost:8000/api/v3/testumgebung/de/children/?depth=3&url=/testumgebung/de/parent/child-1/child-2/child-3)
- See the slugs of the first and second level page (`/root/child-1/` in the example) is missing in the path
image

- Check out to this branch and try the endpoint
- See the slugs of the first and second level appear
image

### Resolved issues

Fixes: #4220

### DIFF
--- a/integreat_cms/cms/models/pages/page.py
+++ b/integreat_cms/cms/models/pages/page.py
@@ -111,6 +111,10 @@ class PageQuerySet(NS_NodeQuerySet, ContentQuerySet):
                     # Set the relative depth to the relative depth of the parent + 1
                     page._relative_depth = result[page.parent_id].relative_depth + 1
                 else:
+                    if page.parent_id:
+                        page._cached_ancestors = list(
+                            page.get_ancestors().prefetch_public_translations()
+                        )
                     # Set the relative depth to 1
                     page._relative_depth = 1
                 result[page.id] = page


hannaseithe (commented):
Thank you very much for your fix. It works well.

I would like to suggest though a different fix, since I see the problem actually rather lying with `cache_tree_dict` and not the children endpoint itself. From what I understand we rely everywhere on `._cached_ancestors` (the cached property produced by cache_tree_dict and used be transform_page) showing the FULL ancestor path of a page (up to its actual root page), therefore `cache_tree_dict` should guarantee this directly even if the subtree_root_id is not an actual root page. Therefore I suggest that instead of your changes in `pages.py` it might be cleaner to add above line 114 the following:

```
                    if page.parent_id:
                        page._cached_ancestors = list(page.get_ancestors())
                    page._relative_depth = 
```

I would be curious to hear your thoughts on this? Which version do you think is cleaner?

MizukiTemma:
> Thank you very much for your fix. It works well.
> 
> I would like to suggest though a different fix, since I see the problem actually rather lying with `cache_tree_dict` and not the children endpoint itself. From what I understand we rely everywhere on `._cached_ancestors` (the cached property produced by cache_tree_dict and used be transform_page) showing the FULL ancestor path of a page (up to its actual root page), therefore `cache_tree_dict` should guarantee this directly even if the subtree_root_id is not an actual root page. Therefore I suggest that instead of your changes in `pages.py` it might be cleaner to add above line 114 the following:
> 
> ```
>                     if page.parent_id:
>                         page._cached_ancestors = list(page.get_ancestors())
>                     page._relative_depth = 
> ```
> 
> I would be curious to hear your thoughts on this? Which version do you think is cleaner?

Thank you for review 😸 

I avoided changing `cache_tree_dict` becuase it is being called by ` chace_tree`, which is used in muptiple places, and I'm not sure how much of calc load `get_ancestors()` bring.

hannaseithe:
> I avoided changing cache_tree_dict becuase it is being called by  chace_tree, which is used in muptiple places, and I'm not sure how much of calc load get_ancestors() bring.

Just one observation: If I am not mistaken cache_tree always passes a tree queryset that starts from a root page. Therefore cache_tree should never trigger the get_ancestors method. It would only be used for the children endpoint. 
That said, I'm happy to go with whichever approach you are more comfortable maintaining

MizukiTemma:
> > I avoided changing cache_tree_dict becuase it is being called by  chace_tree, which is used in muptiple places, and I'm not sure how much of calc load get_ancestors() bring.
> 
> Just one observation: If I am not mistaken cache_tree always passes a tree queryset that starts from a root page. Therefore cache_tree should never trigger the get_ancestors method. It would only be used for the children endpoint. That said, I'm happy to go with whichever approach you are more comfortable maintaining

Let's hear the second review then :)

MizukiTemma:
Updates to the background context: https://chat.tuerantuer.org/digitalfabrik/pl/d7yqnjcqibdufmj69uhcrgjwdc

The merge target was changed from develop to main.

jarlhengstmengel:
Thanks for the PR! This is not a full review, mainly a question to understand the working of the children endpoint. I don't fully understand why your suggested changes work because I'm confused by the relation between `transform_page` and `children`. `transform_page` calls `absolute_url` for the url and the path the endpoint returns per page_translation object. In my understanding this call should always return the same path for the same page_translation. Why do your changes influence what is returned here?

jarlhengstmengel:
Ah nvm, I think I understand now. Your changes enforces chaching the ancestors fully so that they show up in the absolute_url?

MizukiTemma:
> Ah nvm, I think I understand now. Your changes enforces chaching the ancestors fully so that they show up in the absolute_url?

Yes 😸

jonbulz:
I agree with @hannaseithe , this is why:
As was already pointed out, we rely on `_cached_ancestors` to be a complete list of ancestors, otherwise our absolute paths and urls are not computed correctly (which is causing this bug).  So it should not be possible to end up with a truncated list of ancestors from `cache_tree_dict`. Especially because it explicitly supports the use-case of working with a partial tree (the existance of `subtree_root_id` as an argument already proves that).
Your fix works here (thanks btw), because it ensures that for this endpoint, we don't work with a partial tree. But IMO, the fact that this doesn't work is a broken promise from `cache_tree_dict`, so we should tackle the issue there. Otherwise we might end up with a similar bug for a different endpoint in the future.

One addition to @hannaseithe 's suggestion: I'd do prefetch_public_translations. Otherwise, `ancestor.get_public_translation(self.language.slug)` will hit the DB again for each ancestor. This way, it's a `dict` lookup that only adds on extra query, and the performance difference should be negligible. So:
```
if page.parent_id:
    page._cached_ancestors = list(page.get_ancestors().prefetch_public_translations())
```
in [/integreat_cms/cms/models/pages/page.py:114](https://github.com/digitalfabrik/integreat-cms/blob/49f07153faa193b01d4847bca176aee0909de3b4/integreat_cms/cms/models/pages/page.py#L114)

jarlhengstmengel:
I tested it and as a quick fix it works fine  :) But depending on the prio (if we need a quickfix now), I would much more prefer this being resolved in `cache_tree_dict` or maybe in `ancestor_path` in `page_translation.py`. It feels very weird to me, to be depended on what is manually cached beforehand on what `absolute_url` returns.

MizukiTemma:
@hannaseithe @jonbulz @jarlhengstmengel 

Thank you for all the feedback and review 😻 The PR is now with the approach @hannaseithe and @jonbulz suggested.

hannaseithe (dismissed):
Thank you very much! This looks good to me

MizukiTemma:
Rebase has discarded an alrady given approval 😿 

I would anyway wait for the secont approval by @jonbulz or @jarlhengstmengel for double check

hannaseithe (approved):
:)

jonbulz (approved):
👍

Issue #4246, opened 2026-04-19
Broken release pipeline from #4205


PeterNerlich:
### Describe the Bug

The pipeline for our [most recent release](https://github.com/digitalfabrik/integreat-cms/pull/4230) failed after the changes introduced by #4205.

This PR contains the necessary changes to `tools/make_release_notes.sh` that should have been included in #4205.

#4205 changed how we handle release notes for developer convenience, specifically how release notes are moved from `current/unreleased/` to `2026/2026.X.X`, now always leaving the `unreleased` directory around. `tools/make_release_notes.sh` previously only picked the "last" release note directory, relying on the directories to always contain release notes, and thus also that the `unreleased` directory doesn't exist on full releases.

### Additional Information

https://app.circleci.com/pipelines/gh/digitalfabrik/integreat-cms/17456/workflows/7e112d02-2c81-4be9-89c8-d999b68c7a34/jobs/245601

PR #4230, opened 2026-04-08
Release 2026.4.0

MizukiTemma:
# Release notes

- [ [#1967](https://github.com/digitalfabrik/integreat-cms/issues/1967) ] Validate that the start date of recurring events matches the recurrence rule
- [ [#3997](https://github.com/digitalfabrik/integreat-cms/issues/3997) ] Fix timeout when deleting a language with many translations
- [ [#4050](https://github.com/digitalfabrik/integreat-cms/issues/4050) ] Make primary contact only editable by the contact details box if user is in POI form
- [ [#4089](https://github.com/digitalfabrik/integreat-cms/issues/4089) ] Prevent selected location from being cleared when contact creation fails
- [ [#4135](https://github.com/digitalfabrik/integreat-cms/issues/4135) ] Copy appointment link to contact when created via location's contact details
- [ [#4173](https://github.com/digitalfabrik/integreat-cms/issues/4173) ] Skip machine translation for fields that have not changed since the last translation

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #4205, opened 2026-03-19
CI: Release notes

PeterNerlich:
### Short description

Prevent merge conflics after releases caused by moving release notes.

*(Also only run the build-rc-package step on `rc/` branches, [which I apparently missed](https://app.circleci.com/pipelines/github/digitalfabrik/integreat-cms/17379/workflows/1d05a94b-4652-40d7-8a25-c9770f34aefc) in #4193)*

### Proposed changes

- When moving release notes, don't make git think of it as a rename operation
    - Move the files explicitly instead of the directory
    - Keep the unreleased directory by leaving a `.gitkeep` there
- Fix loading release notes to only read yaml files *(and not fail trying to interpret `.gitkeep` as yaml)*
- Don't show the *Unreleased* headline if there are no release notes *(previously it wasn't shown because the directory didn't exist)*
- *Fix the CI configuration to only run the *build-rc-package* step on `rc/` branches*

### Side effects

- If I messed something up, the next release will need manual intervention

### Faithfulness to issue description and design

The *build-rc-package* fix was included because it was simpler than opening another issue plus PR for this

### How to test

We will see at the next release

### Resolved issues

Fixes: #4204

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #4247, opened 2026-04-19
pretend unreleased dir not there if without release notes

PeterNerlich:
### Short description

The pipeline for our [most recent release](https://github.com/digitalfabrik/integreat-cms/pull/4230) failed after the changes introduced by #4205.

This PR contains the necessary changes to `tools/make_release_notes.sh` that should have been included in #4205.

#4205 changed how we handle release notes for developer convenience, specifically how release notes are moved from `current/unreleased/` to `2026/2026.X.X`, now always leaving the `unreleased` directory around. `tools/make_release_notes.sh` previously only picked the "last" release note directory, relying on the directories to always contain release notes, and thus also that the `unreleased` directory doesn't exist on full releases.

### Proposed changes

- Modify arguments to `find` commands to not throw an error for empty directories
- When automatically determining the most recent version to show release notes for, ensure the `unreleased` directly is not included when it does not contain any `*.yml` release notes

### Side effects

- This PR can be merged directly into `main` without further disrupting the release process. The version bump step on the previous merge commit for the release PR did not complete and the bump commit was not pushed to `main`. Thus the code on `main` currently still specifies the old version (`2026.3.0`) and merging another PR into it won't result in a version like `2026.5.0` or `2026.4.1` but the expected `2026.4.0`.

### Resolved issues

Fixes: #4246

### DIFF
--- a/tools/make_release_notes.sh
+++ b/tools/make_release_notes.sh
@@ -104,7 +104,7 @@ function format_release_notes_version {
 }
 
 function format_release_notes_year {
-    find "$1/" -maxdepth 1 -type d | sort --version-sort --reverse | while read -r VERSION; do
+    find "$1/" -mindepth 1 -maxdepth 1 -type d | sort --version-sort --reverse | while read -r VERSION; do
         format_release_notes_version "${VERSION}"
     done
 }
@@ -130,7 +130,7 @@ if [[ -n ${ONLY_VERSION} ]]; then
     format_release_notes_version "${RELEASE_NOTES_DIR}/${ONLY_VERSION_PATH}/${ONLY_VERSION}" | sed '$ d' | sed '$ d' >> "${OUTPUT}"
 elif [[ -n ${ALL} ]]; then
     # Append each year's entries
-    find "${RELEASE_NOTES_DIR}/" -maxdepth 1 -type d | sort --reverse | while read -r YEAR; do
+    find "${RELEASE_NOTES_DIR}/" -mindepth 1 -maxdepth 1 -type d | sort --reverse | while read -r YEAR; do
         format_release_notes_year "${YEAR}" >> "${OUTPUT}"
     done
 else
@@ -147,6 +147,10 @@ else
     $empty && note_dirs=("${note_dirs[@]/${RELEASE_NOTES_DIR}'/current/unreleased'}")
     # Only return the latest version by default
     latest_version_dir="$(printf '%s\n' "${note_dirs[@]}" | sort --version-sort | tail -n1)"
+    if [[ -z "${latest_version_dir}" ]]; then
+        echo "No release notes found" > /dev/stderr
+        exit 1
+    fi
     format_release_notes_version "$latest_version_dir" | sed '$ d' | sed '$ d' >> "${OUTPUT}"
 fi
 
--- a/tools/make_release_notes.sh
+++ b/tools/make_release_notes.sh
@@ -97,14 +97,14 @@ function format_release_notes_version {
             fi
         fi
     fi
-    find "$1"/* -maxdepth 0 -type f | sort --version-sort | while read -r ENTRY; do
+    find "$1/" -maxdepth 1 -type f -name '*.yml' | sort --version-sort | while read -r ENTRY; do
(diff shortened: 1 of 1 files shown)


tools/make_release_notes.sh
```diff
@@ -97,14 +97,14 @@ function format_release_notes_version {
             fi
         fi
     fi
-    find "$1"/* -maxdepth 0 -type f | sort --version-sort | while read -r ENTRY; do
+    find "$1/" -maxdepth 1 -type f -name '*.yml' | sort --version-sort | while read -r ENTRY; do
```
PeterNerlich, line 100: We need to match `*.yml` files explicitly to avoid `.gitkeep` being fed to `format_release_notes_entry()`, which somehow causes empty release notes to be generated

MizukiTemma (dismissed):
Thank you so much for such a speedy fix 🚀

tools/make_release_notes.sh
```diff
@@ -97,14 +97,14 @@ function format_release_notes_version {
...
         format_release_notes_entry "${ENTRY}"
     done
     echo -e "\n"
 }
 
 function format_release_notes_year {
-    find "$1"/* -maxdepth 0 -type d | sort --version-sort --reverse | while read -r VERSION; do
+    find "$1/" -maxdepth 1 -type d | sort --version-sort --reverse | while read -r VERSION; do
```
hannaseithe:
```suggestion
    find "$1/" -mindepth 1 -maxdepth 1 -type d | sort --version-sort --reverse | while read -r VERSION; do
```
Suggestion by Claude, so we do not include the starting directory

tools/make_release_notes.sh
```diff
@@ -130,12 +130,24 @@ if [[ -n ${ONLY_VERSION} ]]; then
     format_release_notes_version "${RELEASE_NOTES_DIR}/${ONLY_VERSION_PATH}/${ONLY_VERSION}" | sed '$ d' | sed '$ d' >> "${OUTPUT}"
 elif [[ -n ${ALL} ]]; then
     # Append each year's entries
-    find "${RELEASE_NOTES_DIR}"/* -maxdepth 0 -type d | sort --reverse | while read -r YEAR; do
+    find "${RELEASE_NOTES_DIR}/" -maxdepth 1 -type d | sort --reverse | while read -r YEAR; do
```
hannaseithe:
```suggestion
    find "${RELEASE_NOTES_DIR}/" mindepth 1 -maxdepth 1 -type d | sort --reverse | while read -r YEAR; do
```

hannaseithe (dismissed):
I don't know shell script, so it's hard for me to review the validity of your changes. But I at least let claude run against it, and it claims to have found one minor bug and a minor edge case:

- **Minor edge case:** if `note_dirs` ends up containing only empty strings (i.e., the only versioned dir was `current/unreleased` and it's empty), `latest_version_dir` would be empty and `format_release_notes_version ""` would be called with unpredictable results. Practically impossible (there's always at least one released version), but worth noting.

tools/make_release_notes.sh
PeterNerlich: Yes, makes sense—when it was `-maxdepth 0` giving a mindepth was redundant after all, but now it's not

PeterNerlich:
> if `note_dirs` ends up containing only empty strings (i.e., the only versioned dir was `current/unreleased` and it's empty), `latest_version_dir` would be empty and `format_release_notes_version ""` would be called

Thanks! At first I was just going to dismiss it, but since we then go on to call `find "$1/" -maxdepth 1 -type f -name '*.yml'` with `"$1/"` expanding to `"/"`, dumping random files from root does not sound critical, but it surely feels very weird. I included a check to prevent this.

PeterNerlich:
@MizukiTemma @hannaseithe Accepting the suggestions revoked your reviews. Please again have a more or less detailed look, depending on your confidence and how invested you feel, and submit another quick review! Thanks!

hannaseithe (approved):
Thank you for considering "my claude-code review. From what I understand it looks good to me :)

MizukiTemma (approved):

PR #4251, opened 2026-04-21
fix releases without release notes

PeterNerlich:
### Short description

After #4247, there is still a bug in the bump version CI step for the case a release is created without any release notes. This is essentially the same bug, but in the CI step definition:

When moving the release notes *(now the individual files, not the whole `current/unreleased/` directory)* as part of the release pipeline step, the move command fails when there are no release notes, as `"$1"/*.yml` expands to nothing.

### Proposed changes

- Check whether there are any release notes to move first. If not, don't attempt to move anything
- Use this opportunity to add a release note for #4245, making the fix not technically necessary, but a release without any release notes is indeed weird

### Side effects

- It is funny / We are lucky that we noticed this right away by trying to release another fix without adding any release note for the release. Not that I expect this would have been a hugely time consuming investigation if it ever came up at a later point

### Resolved issues

Fixes: #4246


MizukiTemma (commented):
Let's go 🚀

MizukiTemma (approved):

Issue #4257, opened 2026-04-29
Sortable table header: Order of location categories sorted via ID and not alphabetically


juliankehne:
### Bug Summary

When sorting the locations in ascending or descending order via the table header sorting based on the location category, the sorting is not alphabetical and no clear reason has been found just by looking at the front-end information as to how they are ordered.

### Expected Behavior

When sorting the location via category they should be sorted based on the selected language

### Actual Behavior

Categories are sorted via ID of the category.

### Steps to Reproduce

1. Go to locations
2. Click Arrows for sorting next to category either ascending or descending
3. Observe sorting order

### Reproducibility

Always (100%)

### Environment

- cms-test
- [Version: 2026.4.10-alpha](https://cms-test.integreat-app.de/testumgebung/release-notes/)

### Screenshots / Videos / Logs

<img width="513" height="642" alt="Image" src="https://github.com/user-attachments/assets/e6e66448-e345-4b59-addf-8f01888cd204" />

### Severity

🟡 Minor — Non-core functionality affected, acceptable workaround exists

### Regression Information

- Not a regression

### Acceptance Criteria for Fix

- [ ] alphabetical sorting works in ascending and descending order
- [ ] sorting works for multiple languages (at least DE & EN), since categories can have different names (translated) based on the users selected CMS language

### Related Issues

https://github.com/digitalfabrik/integreat-cms/issues/1936

juliankehne:
Was fixed with #4261

PR #4261, opened 2026-04-29
fix sort by related m2m fields in POI list view

jonbulz:
### Short description

Fix sorting in list views when the sort field is a reverse relation

### Proposed changes

Sorting a queryset is trivial when sorting by a field that is on the model, but not trivial when the field is on a related model, especially when the relation produces multiple rows per parent (a reverse foreign key like POI ↔ POITranslation). The current (generic) approach attempted to handle this in the `FilterSortMixin` by collapsing those rows using `Min`, annotate it to the base queryset and then sort by the annotated field. Unfortunately, this does not yield any useful results for our use cases for 2 reasons:

- The collapsed value picks the lexicographically smallest `translations__status` (or `translations__title`) across all languages and all versions of a POI, not the translation in the language the user has selected in their view
- status is stored as the choice key (AUTO_SAVE, DRAFT, PUBLIC, REVIEW), so even after collapsing the alphabetic order has no relation to the localized labels the user sees

What we want instead is to sort the queryset based on the latest translation in the user-selected language. So with this PR, we

- remove the attempt to handle reverse-relation sort fields generically from the `FilterSortMixin`
- add custom logic to `POIListView` to handle sorting by title, publication status and category name (Note that POICategory has no name column: the property goes through `best_translation.name`. So the category sort uses a Subquery against POICategoryTranslation in the active backend language, matching what is rendered in the row)

### Side effects

- Since no other list views sort by related fields yet, only the `POIListView` is affected by these changes
- The mixin still passes __ sort keys through to plain `order_by`, which is correct for single-row joins (FKs from the listed model). A future view that wants to sort by a reverse-relation field will need to provide its own annotation, otherwise the joined rows will be duplicated, and similar bugs as the ones fixed here may be introduced

### How to test

- In a region with at least two languages, create POIs whose translations differ between languages — e.g. one POI is Published in DE but Draft in EN
- Go to 'Locations' in EN and sort by Publication Status ascending — verify Draft rows come before Published rows, and that the order reflects the EN status, not the DE one.
- Sort by Title ascending/descending in EN and in DE — verify the ordering matches the titles actually rendered in the row, not the title in the other language.
- Sort by Category ascending — verify it's alphabetical by the category name shown in the row (i.e. by the active backend language's translation), not by category id.
- Verify rows are not duplicated and pagination still works under each sort.

### Resolved issues

Fixes: #4257 

### DIFF
--- a/integreat_cms/cms/views/pois/poi_list_view.py
+++ b/integreat_cms/cms/views/pois/poi_list_view.py
@@ -1,25 +1,26 @@
 from __future__ import annotations
 
-import logging
 from typing import TYPE_CHECKING
 
 from django.contrib import messages
+from django.db.models import Case, IntegerField, OuterRef, Subquery, Value, When
 from django.shortcuts import redirect, render
 from django.utils.decorators import method_decorator
+from django.utils.translation import get_language
 from django.utils.translation import gettext_lazy as _
 from django.views.generic import TemplateView
 
+from ...constants import status
 from ...decorators import permission_required
-from ...models import POI, POITranslation
+from ...models import POI, POICategoryTranslation, POITranslation
 from ..mixins import FilterSortMixin, MachineTranslationContextMixin, PaginationMixin
 from .poi_context_mixin import POIContextMixin
 
 if TYPE_CHECKING:
     from typing import Any
 
+    from django.db.models.query import QuerySet
     from django.http import HttpRequest, HttpResponse
-
-logger = logging.getLogger(__name__)
 
 
 @method_decorator(permission_required("cms.view_poi"), name="dispatch")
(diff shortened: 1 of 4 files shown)


MizukiTemma (approved):
Looks good 😸

integreat_cms/cms/views/pois/poi_list_view.py
```diff
@@ -4,19 +4,23 @@
...
 from ..mixins import FilterSortMixin, MachineTranslationContextMixin, PaginationMixin
 from .poi_context_mixin import POIContextMixin
 
 if TYPE_CHECKING:
     from typing import Any
 
+    from django.db.models.query import QuerySet
     from django.http import HttpRequest, HttpResponse
 
 logger = logging.getLogger(__name__)
```
seluianova: not actually related (feel free to ignore), just noticed that this logger is unused

integreat_cms/cms/views/pois/poi_list_view.py
```diff
@@ -42,15 +46,53 @@ class POIListView(
...
+        )
+        queryset = queryset.annotate(
+            _sort_title=Subquery(latest_translation.values("title")[:1]),
+            _sort_status=Subquery(ranked_status.values("_rank")[:1]),
+            _sort_category_name=Subquery(category_translation.values("name")[:1]),
+        )
+        return super().get_filtered_sorted_queryset(queryset)
+
     def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
         r"""
```
seluianova, line 94:
```suggestion
        """
```
one more unrelated nit-pick 🙃

integreat_cms/cms/views/pois/poi_list_view.py
```diff
@@ -42,15 +46,53 @@ class POIListView(
...
     ]
 
+    def get_filtered_sorted_queryset(self, queryset: QuerySet) -> QuerySet:
+        """
+        Annotate sort keys that match what the user actually sees in the list:
+
+        * ``_sort_title`` and ``_sort_status``: latest translation (any version,
+          including auto-saves) in the language slug from the URL — this is the
+          translation rendered by ``poi_list_row.html``.
+        * ``_sort_status`` ranks status values by alphabetic order in German (AUTO_SAVE, DRAFT,
```
hannaseithe: Just a thought: If we truly want to sort by the german translations of the status value, it might be better to look them up first and simply assign the German translations in the annotation. Instead of hardcoding the rank like you did below. On the other hand the ranking you used mirrors the workflow of the status, which I would also consider a legtitimate way of sorting, but then that should maybe show here in the docstring

seluianova:
I found a tricky issue.

Preconditions:
- **The user’s language is set to English**.
- There are POI categories with both English and German translations, as well as categories with only German translations.

In this case sorting by category looks like this:

image

Categories with English translations are sorted as expected.
But categories that have only German translations look randomly sorted.

I'm not sure it's worth making the sorting logic more complex to address this.
From a user’s perspective, though, the order may appear inconsistent.

I will add my approval anyway since the original issue is fixed.
Looks good to me 👍

seluianova (approved):

hannaseithe (approved):
Since you already have two reviewers, I just approve. Just wanted to add my one comment below and maybe a test for the get_filtered_sorted_queryset on the POIListView would be helpful. I leave it up to your own discernment if you want to consider those.

integreat_cms/cms/views/pois/poi_list_view.py
jonbulz: So would you keep the order and just change the wording in the docstring? It's true that we have to change this any time we change the possible status values, that's maybe not ideal...

hannaseithe: I am not sure, but I personally think its ok to keep the order, but just change the wording.

jonbulz:
> I found a tricky issue.
>
> Preconditions:
> - The user’s language is set to English.
> - There are POI categories with both English and German translations, as well as categories with only German translations.


@seluianova thanks for finding this. The root cause is that the categories are sorted by their name in the language the user has selected. In the example you described, the name will be `None` for those without an english translation, thus the random ordering. I think this is an edge case that is safe to ignore for now, and if it should occur, we can explain it to the users.

integreat_cms/cms/views/pois/poi_list_view.py
jonbulz: I'd prefer to keep it that way because there are `\` in the docstring and I like to follow this [convention](https://www.docutils.org/docs/peps/pep-0257.html#:~:text=Use%20r%22%22%22raw%20triple%20double%20quotes%22%22%22%20if%20you%20use%20any%20backslashes%20in%20your%20docstrings)

seluianova: oopsi, it means "raw" 🙈 I thought it was a typo
(I will have to get used to Python a bit...)

integreat_cms/cms/views/pois/poi_list_view.py
jonbulz: Done

integreat_cms/cms/views/pois/poi_list_view.py
jonbulz: Seems reasonable, I changed it

PR #4262, opened 2026-04-29
Refactor network area and team memberships

### LINKED ISSUE #2451: Refactor network area and team memberships
### Motivation

When I designed the access control system, I thought app team users might want to have read-only access to the entire CMS, also to debug specific app-related problems in other regions.
However, in reality they need write-access, but only to the testumgebung and not to other regions.
This causes some confusion when new app team members are created in the CMS.

### Proposed Solution

The simplest solution would be to just remove the app team from the possible team choices and to rename the checkbox "Team member" to "Access to network area" to point out that it is possible to create team members who do not have access to the network area.
Then, app team members would just be created as managers of the testumgebung.

### Alternatives

A more complex solution would be to keep the current app team with the network access, but make all permissions read-only again and somehow make it possible that they have write access for the testumgebung, despite having only read access to other regions.
This could even be used to allow them to edit the region form (in the network area) for "testumgebung", but not for the other regions.

seluianova:
### Short description
`App Team` members should not have access to the `Network Management` area.
Instead, existing `App Team` users should be migrated to the `Manager` role in `Testumgebung`.

### Proposed changes
- Add a migration to move all `App Team` users to the `Manager` role in `Testumgebung`, and remove the `App Team` group and role from the database.
- Rename the checkbox in the user settings from `Team member` to `Access to network area`.

### Side effects
no?

### How to test
- Verify that the `App Team` is no longer available for selection.
image

- Verify that the existing `App Team` members no longer have access to `Network Management`, but their role is changed to `Manager` and the `Testumgebung` region is assigned.

### Resolved issues
Fixes: #2451


MizukiTemma (approved):
Works as exepcted 😸 Thank you!

integreat_cms/locale/nl/LC_MESSAGES/django.po
```diff
@@ -1900,8 +1900,8 @@ msgid "Administrator"
 msgstr "Beheerder"
 
 #: cms/forms/users/user_filter_form.py:30 cms/forms/users/user_form.py:109
-msgid "Integreat team member"
-msgstr "Integreat-teamlid"
+msgid "Access to network area"
+msgstr "Toegang tot het netwerkgebied"
```
jarlhengstmengel:
```suggestion
msgstr "Toegang tot het netwerkbereik"
```
Sounds more intuitive to me in dutch, but since we don't really maintain dutch translations, feel free to ignore :)

jarlhengstmengel (approved):
Thank you very much for the PR! Looks good to me :)

PR #4276, opened 2026-05-06
Increase PDF link text truncation limit from 50 to 80 characters

### LINKED ISSUE #4275: Increase PDF link text truncation limit from 50 to 80 characters
## Problem

Link texts in generated PDFs are truncated at 50 characters per word token. This means any link whose text contains a word (or URL) longer than 50 characters gets visibly cut off with an ellipsis, even when the text would still fit on the line.

The limit was introduced in #1904 as a workaround: long URLs used as link text would overflow the line boundary in xhtml2pdf, since the PDF renderer does not support `word-break: break-all` or `overflow-wrap: break-word`. Truncating at 50 characters was a conservative choice to prevent overflow.

## Why 50 is too tight

On A4 paper with 2 cm margins at `font-size: large`, approximately 80–90 characters fit on a single line. A limit of 50 causes many URLs and descriptive link texts that would wrap fine to be truncated unnecessarily.

Note: the truncation applies **per whitespace-separated token**, not per full link text. Ordinary readable link texts like *"Klicken Sie hier für mehr Informationen"* are unaffected. Only individual tokens longer than the limit are truncated — in practice, raw URLs used as link text.

## Proposed fix

Raise the `max_chars` argument in `page_pdf.html` from `50` to `80`. This is a one-file change (3 occurrences).

A CSS-based fix (`word-break`, `overflow-wrap`) is **not possible** with xhtml2pdf — these properties are not implemented in the library. The truncation workaround remains necessary; only the threshold changes.

dkehne:
## Summary

- Raises the per-token link text truncation limit in generated PDFs from 50 to 80 characters
- The previous value of 50 was chosen arbitrarily; on A4 with standard margins at the PDF font size (~large), approximately 80–90 characters fit per line, making 80 a more appropriate threshold
- The truncation workaround remains necessary since xhtml2pdf has no support for `word-break` or `overflow-wrap` CSS properties

Closes #4275

## Test plan
- [ ] Generate a PDF for a page that contains a link (raw url) between 51 and 80 characters — verify it is no longer truncated
- [ ] Generate a PDF for a page that contains a link (raw url) over 80 characters — verify it is still truncated (no line overflow)
- [ ] Generate a PDF for a page that contains a link with anchor text longer than 80 characters - verify that the text is not truncated
- [ ] Run existing PDF tests: `tools/test.sh tests/pdf/`

### DIFF
--- a/integreat_cms/release_notes/current/unreleased/4275.yml
+++ b/integreat_cms/release_notes/current/unreleased/4275.yml
@@ -1,0 +1,2 @@
+en: Increase PDF link text truncation limit from 50 to 80 characters to reduce premature truncation of long URLs
+de: Erhöht die Kürzungsgrenze für Linktexte in PDFs von 50 auf 80 Zeichen, um vorzeitiges Abschneiden langer URLs zu reduzieren
--- a/integreat_cms/cms/templates/pages/page_pdf.html
+++ b/integreat_cms/cms/templates/pages/page_pdf.html
@@ -113,11 +113,11 @@
                 <h1 class="level-{{ page.depth|add:"-1" }}">{{ page_translation.title }}</h1>
                 <div class="content">
                     {% if page.mirrored_page_first %}
-                        {{ page_translation.mirrored_translation_text|pdf_strip_fontstyles|pdf_truncate_links:50|safe }}
+                        {{ page_translation.mirrored_translation_text|pdf_strip_fontstyles|pdf_truncate_links:80|safe }}
                     {% endif %}
-                    {{ page_translation.content|pdf_strip_fontstyles|pdf_truncate_links:50|safe }}
+                    {{ page_translation.content|pdf_strip_fontstyles|pdf_truncate_links:80|safe }}
                     {% if not page.mirrored_page_first %}
-                        {{ page_translation.mirrored_translation_text|pdf_strip_fontstyles|pdf_truncate_links:50|safe }}
+                        {{ page_translation.mirrored_translation_text|pdf_strip_fontstyles|pdf_truncate_links:80|safe }}
                     {% endif %}
                 </div>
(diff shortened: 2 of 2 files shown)


seluianova:
> Increase PDF link text truncation limit from 50 to 80 characters to reduce premature truncation of meaningful link texts

Why do we need truncation for meaningful link texts?
If the link text is actually readable text (with words and spaces), maybe it makes sense to display it as normal text?

image


**UPD**. ah, wait. the issue description says:

> Note: the truncation applies per whitespace-separated token, not per full link text. Ordinary readable link texts like "Klicken Sie hier für mehr Informationen" are unaffected.

So readable link text is **not** affected. Then the release note is a bit misleading?
I would also suggest updating the test instructions accordingly.

integreat_cms/release_notes/current/unreleased/4275.yml
```diff
@@ -0,0 +1,2 @@
+en: Increase PDF link text truncation limit from 50 to 80 characters to reduce premature truncation of meaningful link texts
+de: Erhöht die Kürzungsgrenze für Linktexte in PDFs von 50 auf 80 Zeichen, um vorzeitiges Abschneiden von sinnvollen Linktexten zu reduzieren
```
seluianova: smthg like this?

```suggestion
en: Increase PDF link text truncation limit from 50 to 80 characters to reduce premature truncation of long URLs
de: Erhöht die Kürzungsgrenze für Linktexte in PDFs von 50 auf 80 Zeichen, um vorzeitiges Abschneiden langer URLs zu reduzieren
```

seluianova (approved):
Tested, looks good to me 👍

PeterNerlich (approved):
LGTM

Issue #4267, opened 2026-05-04
New workflow for user author


osmers:
### User Story

As an author within the CMS system
I want to be able to save pages as draft
So that I can hand them in only once they are completely done and ready for review

#### Current State
Authors cannot save pages as draft. 

#### Pain Point
When authors work on content and want to create a new page, they can only submit it for review. Sometimes the page is not ready to be reviewed though and they only want to save what they worked on so far. This is not possible. 
Authors cannot save pages as draft (reason being, that saving a page as draft would take it offline, if it was online/published before). If it was never published, there is no damage that could be done.

#### Current workaround
Submit a new page for review and go back to edit it. Change something and leave the page so that changes get saved automatically. This removes the page from the list of pages to be reviewed. It is not very intuitive and requires some technical understanding.

#### Functional Criteria
- [ ] Authors can save a page as "draft" - this should not lead to unpublishing pages

#### UI/UX Criteria
- [ ] UI Design is followed

### User Flow

This is a bit complicated - @m0ndra we can have an exchange on this, if needed from your side
We'd need an understandable workflow that does not change once a page has been published.

### Definition of Ready

- [ ] Acceptance criteria are clear and testable
- [ ] UI design / mockups are available (if applicable)
- [ ] Dependencies are identified and unblocked
- [ ] Story has been refined / estimated by the team
- [ ] Edge cases and error states have been considered

MizukiTemma:
@osmers 

Do you mean authors? Editors don't need approval for pages.

osmers:
Sorry, I will change it :D my bad

PR #4279, opened 2026-05-11
Exclude pages without public translations from mirrored page selection

### LINKED ISSUE #4259: Pages in Draft Status can be selected in "Live Inhalte"
### Bug Summary

If municipalities want to include "Live Inhalte" in their page, they can chose from a variety of pages from other municipalities or from the Integreat Inhalte. We have just seen, that pages in "draft" status can also be selected, which we were not aware of.

### Expected Behavior

Published pages should only be shown in the Live Inhalte section. Pages in draft status should not be included in the list.

### Actual Behavior

All pages are shown, regardless their actual status.

### Steps to Reproduce

1. Go to any page
2. Click on "Live Inhalte einbinden"
3. Chose the region
4. Chose one page and see, that also pages in draft status are displayed.

### Reproducibility

Always (100%)

### Environment

In the CMS system

Image
Page "Rechtskreiswechsel Ukraine" is currently in draft status.

### Severity

### Regression Information

_No response_

### Technical Context

_No response_

- [ ] Only published pages appear as choice
- [ ] Everything else works as before

_No response_

### Verification / Testing

_No response_

### Related Issues

_No response_

seluianova:
### Short description
Exclude pages without public translations from the mirrored page selection.

### Proposed changes
- Added a `is_publicly_visible` property to Page
- Filtered mirrored page choices to exclude pages without public translations

### Side effects
Hopefully no

### How to test
1. Create a page without any public translations (in draft or pending approval status)
2. Open another page and go to “Live Inhalte einbinden”
3. Verify that the page created in step 1 is not available for selection

### Resolved issues
Fixes: #4259


jarlhengstmengel (commented):
Thank you for the fix! The code looks good and it works. It is already the case that when embedded pages are in Draft that the endpoint doesn't include them in the response. While Testing I only observed that when there was already a page embedded which was in draft, the first page in the `Embed live content` dropdown would be shown in the page edit view. When hitting update this "wrong" page would then be embedded. I'm not sure if it's really the case that there are embedded pages that are in draft in production data, but I can imagine that this mistake could happen. So I think I'm in favor to add an migration for this reason.

seluianova:
> When hitting update this "wrong" page would then be embedded.

@jarlhengstmengel right, good point.
I have added the migration, please have a look now!

seluianova:
I found one more related issue.

1. Create page A with public translation.
2. Open page B and select page A in the "Live Inhalte einbinden" section, then click "Publish".
3. Go to the page A and change its status to "Draft".

As a result, the page A will still be embedded in the database as live content. However, since it is no longer available in the dropdown, when page B is opened in the editor and modified, the embedded page will be silently replaced with the first page in the list, without any notifications.

I guess the expected behavior should be either:

1. If a draft page was already embedded, it should remain selected in the form (so that it is not overwritten when the page is edited). However, it should not be available in the dropdown.

or

2. When a page status is changed to "Draft", we should perform a check to determine whether the page is currently embedded, show a warning, and remove it from the embedded content if necessary.

If we go for the first option (sounds more reasonable to me), I guess we also don't actually need a migration. @jarlhengstmengel what do you think?

**UPD**: I added one more commit that removes the migration and keeps draft pages selected on the form (while hiding them from the dropdown).

seluianova:
Another question:
If parent page is draft (and excluded from live content selection) - does this mean the child pages should also be excluded❓

hannaseithe:
> Another question: If parent page is draft (and excluded from live content selection) - does this mean the child pages should also be excluded❓

Are public child pages of draft parent pages still accessible in the app somehow? Is it reasonable for the CMS user to assume that when they turn a parent page into draft, that all children are kind of "implied draft" until their parent has become public again? I think I would base the handling on that. 

> If a draft page was already embedded, it should remain selected in the form (so that it is not overwritten when the page is edited). However, it should not be available in the dropdown.

I like this solution. And I think it would be good if we would visibly mark such a pre-selected "draft-page" as such in the selected field as currently in draft.

seluianova:
> Are public child pages of draft parent pages still accessible in the app somehow?

as I understand - no.

> Is it reasonable for the CMS user to assume that when they turn a parent page into draft, that all children are kind of "implied draft" until their parent has become public again?

no idea 😀
but there is no warning or something... maybe we need one 🤔

hannaseithe:
Actually then let's put this question back to @laravoelk @osmers @m0ndra 

Should you be able to select a public child page of a parent page that is currently in draft status for mirroring ("Live-Inhalte einbinden") into another page?

m0ndra:
> Should you be able to select a public child page of a parent page that is currently in draft status for mirroring ("Live-Inhalte einbinden") into another page?

My gut feeling after first reading this question is: **YES** 

But: I have not yet a full understanding about how all the different states (public, draft) affect the integreat app and as far as my understanding goes, the current underlying concept of publishing/saving as drafts/auto saving has some....flaws? weird behaviours? (Also: related to this ticket: https://github.com/digitalfabrik/integreat-cms/issues/4267). I'd like to improve that, but of course not within this issue... So my answer "YES" might lead to inconsistencies for the time being AND we might have to change it again if we have a better understanding and idea how to improve. But I stick with "Yes" for now. I hope this helps

laravoelk:
I would also agree with a Yes currently and would say that only the draft pages should be excluded.

osmers:
> Should you be able to select a public child page of a parent page that is currently in draft status for mirroring ("Live-Inhalte einbinden") into another page?

Actually, I would say **no** - reason being, that as Tory said above, you cannot see pages that have parents in draft status. Thus, how can the municipality that is implementing the page, check the content before implementation?

We specifically removed the possibility to have published children of parents in draft status as to not have parent-less pages in the first level (like we used to have in wordpress). 

> > Is it reasonable for the CMS user to assume that when they turn a parent page into draft, that all children are kind of "implied draft" until their parent has become public again?
> 
> no idea 😀 but there is no warning or something... maybe we need one 🤔

Yeah, a warning might be good. We sometimes have the case that users are confused why their page isn't visible (especially in the translation) and it is because the parent page isn't public.

seluianova:
> And I think it would be good if we would visibly mark such a pre-selected "draft-page" as such in the selected field as currently in draft.

@hannaseithe I'm not sure how to do that to be honest.
my concern is that we list pages in the dropdown, but the "draft" status is related to translations.
so it wouldn't be entirely correct to add a "draft" label to the page itself, because its translations can have different statuses.
e.g. one language could be "pending approval", while another could be "draft" or "auto-save".

**UPD**: Also, page names are often quite long (compared to region names), so there isn't much space in the field for additional labels.

image

I could imagine adding smthg like “(⚠ No public translations)” to the title, but it’s quite long and wouldn’t be visible in many cases. so I’m not sure whether it makes sense to add it.

what do you think?

seluianova:
> Actually, I would say no - reason being, that as Tory said above, you cannot see pages that have parents in draft status. Thus, how can the municipality that is implementing the page, check the content before implementation?

> We specifically removed the possibility to have published children of parents in draft status as to not have parent-less pages in the first level (like we used to have in wordpress).

Since I had the same concern, I decided to exclude pages whose parent pages do not have a public translation from the selection as well. See the last commit.

_But feel free to protest._

I think everything is addressed now and ready for another review :)

hannaseithe:
So we assume that if at least one page translation is public (in any language), that page should be mirrorable. But the rest of the translations can be missing, or be draft or any other status, . I just want to make sure if that is the original intention of the issue? Because on the other hand one might want to argue that if at least one page translation is in draft, the page as a whole should be considered draft (meaning: not mirrorable), since it is being worked on. @osmers could you give your input here?

jarlhengstmengel:
> If a draft page was already embedded, it should remain selected in the form (so that it is not overwritten when the page is edited). However, it should not be available in the dropdown.

I like this solution, it removes the need for a mirgation for me :)

> Since I had the same concern, I decided to exclude pages whose parent pages do not have a public translation from the selection as well. See the last commit.

I think this a good choice, the endpoint wouldn't deliver them either way I think, but then there is no misunderstanding. The other option would be to add some sort of visiual indication that a page is draft/will not be shown in the App as you already dicussed above (or is acitve in discussion?)

seluianova:
As discussed in the last weekly with @osmers, we want to show pages in the mirrored pages dropdown only if they have a public German translation.

However, I tested the following scenario:

* Page A has a **draft** German translation and a **public** English translation.
* Page B has Page A configured as a mirrored page.

In this case:

* When I open Page B in the app in German, the mirrored content is not displayed --> this looks correct to me, since the German translation of Page A is in draft status.
* When I open Page B in the app in English, the mirrored content **is** displayed --> this also looks correct to me, since the English translation of Page A is public.

So I'm wondering whether we actually want to exclude pages that have a draft German translation but public translations in other languages from the selection, considering that this scenario already seems to work correctly in the app? @osmers 

**UPD**. and one more concern:
If we still want to check the German translation status:

- If the user's language is English, should we check the English translation instead of the German one?
- What if the region's default language is English?
- Or do we really want to hard-code German in all cases?

seluianova:
Discussion summary with @m0ndra:

💡 The mirrored pages dropdown should only contain pages that have a public translation in the default region language (which, in most if not all cases, will be German).

A side effect of this approach is that if a page has a public translation in the default (German) language but no public translations in other languages, it can still be mirrored in those languages. In that case, it will be displayed with empty content.

However, we discussed that this is acceptable because we expect the missing translations to be added soon.

osmers:
> A side effect of this approach is that if a page has a public translation in the default (German) language but no public translations in other languages, it can still be mirrored in those languages. In that case, it will be displayed with empty content.

We do have Platzhaltertexte - meaning that if a translation is missing, that page would show a text saying that the content is not available but is available in the following languages.

seluianova:
> We do have Platzhaltertexte - meaning that if a translation is missing, that page would show a text saying that the content is not available but is available in the following languages.

@osmers I mean the following case:

- Page A has a public German translation and a draft English translation.
- Page B embeds content from Page A and has both a public German and a public English translation.

In this case, Page B will be visible to users, but the embedded content from Page A will be empty in the English version.

As I understand Platzhaltertexte won't apply in this case?

osmers:
> As I understand Platzhaltertexte won't apply in this case?

Ahh maybe - I always thought they would, but maybe you are right :) could be a discussion whether to include draft content as not available translation and show the Platzhaltertext. 

For now, I think I am ok with this state of the PR.

seluianova:
**UPD**:

Now the `Live-Inhalte einbinden` dropdown only includes pages with a public translation **in the default region language**.

> And I think it would be good if we would visibly mark such a pre-selected "draft-page" as such in the selected field as currently in draft.

@hannaseithe I decided to address your idea in the following way, let me know if this works for you:

If a page was public when it was selected as a mirrored page, but its status is later changed to draft, the page remains selected in the form.

However, when the user opens the dropdown, the page is shown as disabled and a tooltip is displayed:
`This page cannot be mirrored because it has no public translation in the default language`

image


ℹ️ Ready for another review round.

integreat_cms/cms/forms/pages/mirrored_page_field_widget.py
```diff
@@ -67,4 +68,20 @@ def create_option(
...
+        default_language = instance.region.default_language
+        if (
+            instance
+            and str(instance.mirrored_page_id) == str(value)
+            and default_language
+            and not instance.mirrored_page.get_public_translation(default_language.slug)
+        ):
+            option_dict["attrs"]["disabled"] = True
+            option_dict["attrs"]["title"] = _(
+                "This page cannot be mirrored because it has no public translation in the default language"
```
hannaseithe: The tooltip is a great solution, I am just hesitating about the text UX wise (which is not my greatest strength). But maybe something like: "This is the selected mirrored page. It is currently being edited, and will only show up after it has been published." Also I am not so sure about the disabled status anymore. Like if it is the currently selected mirrored page, I dont really see why accidentally clicking on another page and then wanting to reselect the original one, should not be possible. So maybe only greyed in and tooltip? Sorry for being so undecided about this. If you disagree, I am also ok with keeping it as is.

hannaseithe (approved):
I really like the tooltip solution! I have some more ideas about that, see below. But I leave it up to you to implement those or not :)

integreat_cms/cms/forms/pages/mirrored_page_field_widget.py
seluianova: > It is currently being edited

well I think this may sound like somebody edits the page right now 🤔

@m0ndra tagging you if you have some nice ideas for this tooltip text

my another thought would be:

`This page is currently mirrored but not visible to the users because its default translation is not public.`

m0ndra: I like the tooltip idea! 
For the text, I'd go for something like: 
DE: "Die aktuell als Live-Inhalt ausgewählte Seite ist für Nutzende zur Zeit nicht sichtbar, da sie sich im Entwurfs-Status befindet. Sobald die Seite veröffentlicht wird, werden die Inhalte wieder sichtbar." 
EN: "The page currently selected as live content is not visible to users at this time because it is in draft status. Once the page is published, the content will be visible again."
It's rather long but people forget stuff when they don't log in. I'd also not mention stuff like "default language" or "Standardsprache" since it is complicated (even if its technically correct). But users understand the status, since they often work with drafts/publishing.

hannaseithe: I like both of yours suggestions - definitely better than mine :)

jarlhengstmengel:
How is the state of the PR, can I review or is the Idea about tooltip still in progress?

seluianova:
> How is the state of the PR, can I review or is the Idea about tooltip still in progress?

@jarlhengstmengel I will update the tooltip text, but I would keep the rest as is. imo the PR can be reviewed.

> Also I am not so sure about the disabled status anymore. Like if it is the currently selected mirrored page, I dont really see why accidentally clicking on another page and then wanting to reselect the original one, should not be possible.

If the user accidentally selects a different page, they don’t have to save the changes, or they can deselect the live content (should not be a problem since the draft live content is not displayed anyway).
I think making the item look disabled (greyed out) but not actually disabled is also not ideal 🤔

seluianova:
The tooltip text has been adjusted
image

jarlhengstmengel (commented):
Thank you very much! I like the tooltip :)
I observed the following behaviour which seems to be in conflict to the PR Discription to me, but maybe I overlooked something in the discussion 😅 I can still embed pages that have status "Pending Approval"

Image

Image

seluianova:
> I can still embed pages that have status "Pending Approval"

Does this page have a public translation?
If yes - it works as expected.
If "pending approval" is the only translation of the page - then it's a problem.

jarlhengstmengel (approved):
> Does this page have a public translation?

Yes it does. Then everything looks good to me! Thank you!

PR #4295, opened 2026-05-18
Fix machine translation when page content is cleared

### LINKED ISSUE #4277: Machine translation silently fails when page content is cleared
### Bug Summary

When a page's content field is cleared (e.g. turning a content page into a heading-only parent page), triggering machine translation incorrectly reports "no changes to the source translation" and does not update target language translations. The target translations are left with stale content while the source no longer has any.

### Expected Behavior

When content is deleted from the source translation, target translations should either:
- have their content cleared as well (propagating the deletion), or
- be marked as outdated so editors know they need attention.

The error message "Page X was not translated into Y because there were no changes to the source translation" should not appear — a deletion _is_ a change.

### Actual Behavior

1. Page has title + content in the source language (e.g. German).
2. Editor deletes the content, leaving only the title, and saves.
3. Editor triggers machine translation to other languages.
4. Error: _"Page X was not translated into Y, because there were no changes to the source translation."_
5. Target language translations still contain the old (now deleted) content.
6. In the Integreat app the German page shows no content while other languages show the stale old content.

### Root Cause

In `get_translatable_attributes` (`abstract_content_model.py`), the early-exit `if not value: continue` skips empty fields entirely — including fields that were _just cleared_. Since neither `content` (now empty) nor `title` (unchanged) ends up in `translatable_attributes`, `word_count` becomes 0.

`filter_unchanged_translations` in `machine_translation_api_client.py` then treats `word_count == 0` as "nothing changed", fires the misleading error, and skips the translation update entirely.

### Steps to Reproduce

1. Create a page with title and content in the source language.
2. Remove the content field, leaving only the title, and save.
3. Trigger machine translation to any target language.
4. Observe the error and that target translations retain stale content.

### Suggested Fix

- In `get_translatable_attributes`: when a field is now empty but was non-empty in `old_source_translation`, add it as `(attr, "")` to signal the deletion.
- In `filter_unchanged_translations`: change the condition from `word_count > 0` to `bool(translatable_attributes)` so a cleared field still counts as a change.
- In the API clients (DeepL, Google Translate): skip calling the external API for empty values but write `data[attr] = ""` so the target translation's field is cleared on save.

### Reproducibility

Always (100%)

### Related Issues

Shares the same area of code as #4254 but is a distinct bug (false negative vs. false positive in change detection).

seluianova:
### Short description
When a page’s content field is cleared (e.g. turning a content page into a heading-only parent page), triggering machine translation incorrectly reports "no changes to the source translation" and does not update target translations. As a result, target translations contain stale content while the source translation is empty.

**Suggested fix**: target translations are now also updated - cleared, marked with the machine-translated icon, and set to up to date.

### Proposed changes
- In `get_translatable_attributes`: when a field is now empty but was non-empty in `old_source_translation`, add it as `(attr, "")` to signal the deletion.
- In `filter_unchanged_translations`: change the condition from `word_count > 0` to `bool(translatable_attributes)` so a cleared field still counts as a change.
- In the API clients (DeepL, Google Translate): skip calling the external API for empty values but write `data[attr] = ""` so the target translation's field is cleared on save.
- Add new test

### Side effects
no?

### How to test
- Create a page with non-empty content, enable `Automatic translation` and `Create new translations` checkboxes, click "Publish".
- Clear the content of the source translation, enable `Automatic translation` and `Update existing translations` checkboxes, click "Publish".
- Verify that the target translation is updated: content is empty and the translation is marked as up to date.

### Resolved issues

Fixes: #4277

### DIFF
--- a/tests/mt_api/mt_api_test.py
+++ b/tests/mt_api/mt_api_test.py
@@ -838,12 +838,16 @@ do_not_translate_title = [True, False]
 
 @pytest.mark.django_db
 @pytest.mark.parametrize("do_not_translate_title", do_not_translate_title)
+@pytest.mark.parametrize(
+    "login_role_user",
+    [*PRIV_STAFF_ROLES, MANAGEMENT, EDITOR],
+    indirect=True,
+)
 def test_do_not_translate_title(
     load_test_data: None,
     login_role_user: tuple[Client, str],
     do_not_translate_title: bool,
     settings: SettingsWrapper,
-    caplog: LogCaptureFixture,
     mock_server: MockServer,
 ) -> None:
     """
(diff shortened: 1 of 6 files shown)


tests/mt_api/mt_api_test.py
```diff
@@ -858,104 +862,232 @@ def test_do_not_translate_title(
     """
     client, role = login_role_user
 
-    if role in [*PRIV_STAFF_ROLES, MANAGEMENT, EDITOR]:
```
seluianova, line 861: just moved the roles to @pytest.mark.parametrize and adjusted indentation

hannaseithe (approved):
This looks good to me! Thank you for the fix and the test

MizukiTemma (approved):
Works as expected and nice test improvement😸 Thank you!

PR #4248, opened 2026-04-20
Fix Tiny-MCE New Link Creation

hannaseithe:
### Short description

Switch from using `editor.dom.create -> setNode` to `editor.dom.createHTML -> setContent` which prevents selection being deleted before the new Link can be assigned, as well as fixing the setting of attributes inside `createHTML`

### Proposed changes

- Switch from using `editor.dom.create -> setNode` to `editor.dom.createHTML -> setContent` which prevents selection being deleted before the new Link can be assigned
- use correct the setting of attributes inside `createHTML`, so that auto-update attribute is applied correctly

### Side effects

- None

### Faithfulness to issue description and design

There are no intended deviations from the issue and design.

### How to test

- follow instructions to create a bullet list with links inside #4228 and see that the bullet points do not get deleted anymore
- add a link with "Automatisch den Titel des verlinkten Inhalts als Linktext verwenden" and look at the link with "Quelltext anzeigen" and see that the <a ...> element is created with the right `href` attribute and the `data-integreat-auto-update="true"` attribute as well (and NOT as inside the href value as before)
- make sure that the `do-not_translate` <span> can also be freely assigned and unassigned to a link of the same selection (see bug report below by @MizukiTemma )

### Resolved issues

Fixes: #4196 
Fixes: #4228

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #4318, opened 2026-05-29
normalize urls in internal link checker

### LINKED ISSUE #4235: Links comming back to "Verification needed" (invalid) even once manually marked as valid
### Context & Background

Internal and external links are sometimes judged as invalid even if they are actually functioninig. Munincipalities mark such links as valid by hand.
However, those links come back to the category "invalid" sometimes. Then munincipalities have to manually verify them again. They often update (re-save) the contents which contain such links manually, as the links are then somehow jusged as valid. This is a large effort and waste of time.

### Research Question(s)

There is a cron job with `checklinks`. Probably it is overwriting the manual verification.
After my observation, the realistic solutions are the [shortcodes](https://github.com/digitalfabrik/integreat-cms/issues/3745) and [white list](https://github.com/digitalfabrik/integreat-cms/issues/3859).

Unclear is why the links get jusged as valid when the contents which contain them are re-saved.

### Expected Outcome / Deliverables

- Is there another possible cause? (second opinion)
- If any found:  what is a solution for it?
- Explanation why re-saving a content results in the correct judgement of links in it

### Time Box

1 day

### Scope & Boundaries

** In scope: **
- External and internal links

**Out of scope:**
/

### Definition of Done

- [ ] The questions are answered or no answer/new insight is found after the time box

### Current Uncertainty Level

🟡 Medium — We have some leads but significant unknowns

### Prior Art & Known Resources

_No response_

### Risks & Open Questions

_No response_

### Findings: Summary

_No response_

### Findings: Key Discoveries

_No response_

### Findings: Recommendation

_No response_

### Findings: Follow-up Actions

_No response_

jonbulz:
### Short description

Internal links are often marked as invalid by linkcheck, even though the link resolves. There are several possible causes for this, and it is likely that different mechanisms are overlapping. This PR attempts to eliminate a few possible causes by normalising URLs for the comparison in `check_translation_link` and `check_internal`

### Proposed changes

Normalise links in `check_translation_link` and `check_internal` to 
- drop query parameters and fragments
- percent-decode
- NFKC-normalise
- drop unicode charcater in category 'C' (control & format)
- lower-case
- strip outer slashes

Add tests

### Side effects

Each normalisation step is one that django.utils.text.slugify(value, allow_unicode=True) already applies when the DB slug is generated, and one that check_object_link already applies when looking the target up. It is possible that the app applies stricter URL routing than the CMS does. If so, a link that was formerly marked as invalid could now be marked valid while 404-ing in production (it would have 404-ed before as well, though. so instead of placing it in a wrong bin, we just wave it through now).

### How to test

Check that internal links containing none-ascii characters resolve and are not flagged as invalid. Note that the internal link picker is currently affected by the bug #4248 

### Resolved issues

Fixes: #4235 (hopefully)


jonbulz:
Closed because it did not result in noticeable improvements

PR #4321, opened 2026-06-02
Make slug uniqueness validation & constraint case insensitive

### LINKED ISSUE #4269: Slug uniqueness check is case-sensitive, which allows duplicate slugs to be created
### Bug Summary

It is possible to create the same slug with different casing, e.g. "test-slug" and "Test-Slug".
In our application, these URLs resolve to the same page, so slugs are effectively treated case-insensitively, resulting in duplicate slugs in practice.

Even though manually editing slugs is probably a rare case (?), we still provide this functionality, so the uniqueness check should account for it.

### Expected Behavior

Slugs should be normalized to lowercase before checking for uniqueness.
The normalized slugs should be saved to the database.

### Actual Behavior

Slugs with different casing can be saved successfully, resulting in duplicates.

### Steps to Reproduce

1. Create page/event/poi with the slug "test-slug"
2. Create another page/event/poi (in the same language and region) with the slug "Test-Slug" (slug should be edited manually in this case)

### Reproducibility

Always (100%)

### Environment

Environment: all

_No response_

### Severity

None

### Regression Information

_No response_

### Technical Context

This will need to be implemented on the form-level, the app-level and the db constraint. Analogous to how #3060 was fixed. 
- `generate_unique_slug` will need to be adjusted
- the management command `make_slugs_unique` will need to be adjusted and then run once db constraint has been tightened
- the database triggers will need to account for that

_No response_

### Verification / Testing

_No response_

### Related Issues

_No response_

hannaseithe:
### Short description

So far the app and database levels for slug uniqueness have been treating slugs as case sensitive, but they need to by case insensitive, because the app treats url with case insenstive identical slugs as the same. 

### Proposed changes

- data migration, that lowers all slugs and makes sure they are unique
  - on prod data dump it: 
  ```
  Jun 02 12:48:41 INFO integreat_cms.cms.utils.slug_utils - Finished >make_all_slugs_lowercase< after: 311.830s with 369 updated slugs
  ```
- check constraint on translation models
- ~~adjust database trigger to ensure case insensitive slug uniqueness~~ not necessary if we have the check constraints
- adjust `generate_unique_slug()` to generate only lowercase slugs

### Side effects

- None

### How to test

- try creating uppercase slugs in GUI for page, poi, events -> should not be possible
- try creating uppercase non-unique slugs in GUI -> should not be possibel
- in the shell update a translation with an uppercase slug -> should give IntegrityError

### Resolved issues

Fixes: #4269

### DIFF
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -10093,6 +10093,16 @@ msgstr ""
 "veröffentlichen, aber Sie können Änderungen vornehmen und diese zur Freigabe "
 "vorlegen."
 
+#: cms/views/events/event_form_view.py cms/views/pages/page_form_view.py
+#: cms/views/pois/poi_form_view.py
+#, python-brace-format
+msgid ""
+"The slug was changed from '{user_slug}' to '{slug}', because uppercase "
+"letters are not allowed."
+msgstr ""
+"Der URL-Parameter wurde von '{user_slug}' zu '{slug}' geändert, da "
+"Großbuchstaben nicht erlaubt sind."
+
 #: cms/views/events/event_form_view.py cms/views/pois/poi_form_view.py
 #, python-brace-format
 msgid ""
--- a/integreat_cms/cms/fixtures/test_data.json
+++ b/integreat_cms/cms/fixtures/test_data.json
@@ -5528,7 +5528,7 @@
     "pk": 89,
     "fields": {
       "title": "ስለ Augsburg መታወቅ ያለበት",
-      "slug": "ስለ-Augsburg-መታወቅ-ያለበት",
+      "slug": "ስለ-augsburg-መታወቅ-ያለበት",
       "status": "PUBLIC",
       "content": "<p>Augsburg የሚገኘው በ Bayern ግዛት ውስጥ እና በመካከለኛው franken አውራጃ ውስጥ ነው፡፡ Augsburg 500,000 አካባቢ ነዋሪዎች ሲኖሩት Bayern ውስጥ ከሙኒክ ቀጥሎ ሁለተኛው ትልቅ ከተማ ነው፡፡ Pegnitz የሚባል ወንዝ ከተማ ውስጥ ይፈሳል፡፡</p>",
       "language": 8,
--- a/integreat_cms/cms/migrations/0154_enforce_case_insensitive_slug_uniqueness.py
+++ b/integreat_cms/cms/migrations/0154_enforce_case_insensitive_slug_uniqueness.py
@@ -1,0 +1,56 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import django.db.models.functions.text
+from django.db import migrations, models
+
+from integreat_cms.cms.utils.slug_utils import make_all_slug_lowercase
+
+if TYPE_CHECKING:
+    from django.apps.registry import Apps
+    from django.db.backends.base.schema import BaseDatabaseSchemaEditor
+
+
+def forwards(_apps: Apps, _schema_editor: BaseDatabaseSchemaEditor) -> None:
+    make_all_slug_lowercase()
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("cms", "0153_enforce_slug_uniqueness"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            forwards,
+            migrations.RunPython.noop,
+        ),
+        migrations.AddConstraint(
+            model_name="eventtranslation",
+            constraint=models.CheckConstraint(
+                condition=models.Q(
+                    ("slug", django.db.models.functions.text.Lower("slug"))
+                ),
+                name="eventtranslation_slug_lowercase",
+            ),
+        ),
+        migrations.AddConstraint(
+            model_name="pagetranslation",
+            constraint=models.CheckConstraint(
+                condition=models.Q(
+                    ("slug", django.db.models.functions.text.Lower("slug"))
+                ),
+                name="pagetranslation_slug_lowercase",
+            ),
+        ),
+        migrations.AddConstraint(
+            model_name="poitranslation",
+            constraint=models.CheckConstraint(
+                condition=models.Q(
+                    ("slug", django.db.models.functions.text.Lower("slug"))
+                ),
+                name="poitranslation_slug_lowercase",
+            ),
+        ),
+    ]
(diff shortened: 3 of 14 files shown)


MizukiTemma (approved):
Looks good 😸 Nice PR with tests and migration!

jarlhengstmengel (changes requested):
Thank you very much for the PR! I encoutered `AttributeError: 'NoneType' object has no attribute 'backend_edit_link'` in the POI View. What I did: Create or edit a POI, edit the link manually with a uppercase character. Click on publish, see error. Saving as Draft is possible and I only encountered it for POIs

Complete Callback:
```
Jun 17 13:22:46 ERROR django.request - 500 Internal Server Error: /augsburg/pois/de/504/edit/
  File "/workspace/.venv/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
               ^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/.venv/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/.venv/lib/python3.11/site-packages/django/views/generic/base.py", line 104, in view
    return self.dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/.venv/lib/python3.11/site-packages/django/utils/decorators.py", line 48, in _wrapper
    return bound_method(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/.venv/lib/python3.11/site-packages/django/contrib/auth/decorators.py", line 59, in _view_wrapper
    return view_func(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/.venv/lib/python3.11/site-packages/django/views/generic/base.py", line 143, in dispatch
    return handler(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/.venv/lib/python3.11/site-packages/django/utils/decorators.py", line 48, in _wrapper
    return bound_method(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/.venv/lib/python3.11/site-packages/django/contrib/auth/decorators.py", line 59, in _view_wrapper
    return view_func(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/integreat_cms/cms/views/pois/poi_form_view.py", line 260, in post
    self.show_slug_changed_message(
  File "/workspace/integreat_cms/cms/views/pois/poi_form_view.py", line 353, in show_slug_changed_message
    other_translation_link = other_translation.backend_edit_link
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'backend_edit_link'
Jun 17 13:22:46 ERROR django.server - "POST /augsburg/pois/de/504/edit/ HTTP/1.1" 500 22802
```

hannaseithe:
> Thank you very much for the PR! I encoutered `AttributeError: 'NoneType' object has no attribute 'backend_edit_link'` in the POI View. What I did: Create or edit a POI, edit the link manually with a uppercase character. Click on publish, see error. Saving as Draft is possible and I only encountered it for POIs

I think I have fixed that now, would you mind having a second look?

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -10093,6 +10093,16 @@ msgstr ""
...
 "vorlegen."
 
+#: cms/views/events/event_form_view.py cms/views/pages/page_form_view.py
+#: cms/views/pois/poi_form_view.py
+#, python-brace-format
+msgid ""
+"The slug was changed from '{user_slug}' to '{slug}', because uppercase "
+"letters are not allowed."
+msgstr ""
+"Der URL-Parameter wurde von '{user_slug}' zu '{slug}' geändert, da "
```
jarlhengstmengel, line 10103: Bit unsure about the translation for slug, but I think we use a couple of different translations in the cms (URL-Paramet, Code, Alias, just slug, were the ones I found with a quick search). Maybe we should make that more uniform, but probably not inside the scope of this PR. Feel free to ignore.

jarlhengstmengel (commented):
Thanks for fixing the Issue! I tested changing slugs in UI while creating and editing and changing slugs in the shell for pages, events and pois and all behaved as expected.

I'm a bit unsure about one thing, when you try to change the url by editing it in the UI, you can click on save directly besides the field and it keeps the slug as typed with uppercase. The uppercase get's changed only after publishing. I have the feeling that could be confusing. But maybe this is more of a question @m0ndra?

Manual changing the slug to uppercase:
Image

Still uppercase after hitting the save symbol:
Image

Changed to lowercase after publishing:
Image

integreat_cms/locale/de/LC_MESSAGES/django.po
hannaseithe: Yeah same, i went searching in the translations for german translations, and I think I just picked the one that appeared most often

hannaseithe:
> Thanks for fixing the Issue! I tested changing slugs in UI while creating and editing and changing slugs in the shell for pages, events and pois and all behaved as expected.
> 
> I'm a bit unsure about one thing, when you try to change the url by editing it in the UI, you can click on save directly besides the field and it keeps the slug as typed with uppercase. The uppercase get's changed only after publishing. I have the feeling that could be confusing. But maybe this is more of a question @m0ndra?

I agree this might not be optimal UX behavior, but I would make fixing this not part of this PR, as this is pre-existing behavior. I would recommend creating a new issue, if Sina thinks this should be changed

m0ndra:
> I'm a bit unsure about one thing, when you try to change the url by editing it in the UI, you can click on save directly besides the field and it keeps the slug as typed with uppercase. The uppercase get's changed only after publishing. I have the feeling that could be confusing. But maybe this is more of a question @m0ndra?

@jarlhengstmengel @hannaseithe in terms of consistency it is not the best behavior but I think we can just keep it as it is. No need for a new issue here. Also there is an appropriate hint which explains it so we're fine for now.

jarlhengstmengel (approved):
All good from my side then! Thank you!

PR #4326, opened 2026-06-04
Add previous slugs to the page related API responses

### LINKED ISSUE #1587: Add previous_paths property to all pages endpoints
### Motivation
In the cms changes of the slug and therefore the path of pages are possible. This currently breaks opening of the previous urls in the native apps since a find by path fails.

### Proposed Solution
A new property `previous_paths: string[]` should be added containing all previous valid paths. If there are no previous paths, the array should be empty.

### Alternatives
None

### Additional Context
This blocks https://issues.tuerantuer.org/browse/IGAPP-1050, please update the issue once this is done and released.
For the web app this was not needed as we just used (for now) the current implementation of the children endpoint which returns the page with the correct new path as first item in the response even if queried with one of the previous paths.

MizukiTemma:
### Short description

This PR adds previous slugs to page entries in the API responses.

### Proposed changes

- Collect slugs of all versions of a requested page by `get_translation_slug`
- Remove duplicates and the same slugs with the given translation version
- Return the result as a part of each page entry by `transform_page`

- Publication status of each version is not checked. Let me know if it harms.

### Side effects

- The number of performed queries grows up dramatically. I'm afraid this worsens response time 👀 

### How to test

- Select or create a page which has 1. only one version, 2. several versions with different slugs 3. several version with the same slug in all or some versions
- Call `children/`, `pages/`, `parents`, `page` endpoints
- See the field `previous_slugs` is filled as expected

### Resolved issues

Fixes: #1587 

### DIFF
--- a/integreat_cms/api/v3/pages.py
+++ b/integreat_cms/api/v3/pages.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import json
 import logging
+from collections import defaultdict
 from typing import TYPE_CHECKING
 
 from django.conf import settings
@@ -19,7 +20,7 @@ from django.utils.text import slugify
 from django.views.decorators.csrf import csrf_exempt
 
 from ...cms.forms import PageTranslationForm
-from ...cms.models import Page
+from ...cms.models import Page, PageTranslation
 from ...cms.utils.shortcodes import expand_shortcodes
 from ..decorators import json_response, matomo_tracking
 from .offers import transform_offer
@@ -29,13 +30,16 @@ if TYPE_CHECKING:
 
     from django.http import HttpRequest
 
-    from ...cms.models import Page, PageTranslation
+    from ...cms.models.pages.page import PageQuerySet
 
 logger = logging.getLogger(__name__)
 
 
 def transform_page(
-    page_translation: PageTranslation, context: dict[str, Any] | None = None
+    page_translation: PageTranslation,
+    page: Page | None = None,
+    context: dict[str, Any] | None = None,
+    slug_history: list[str] | None = None,
 ) -> dict[str, Any]:
     """
     Function to create a dict from a single page_translation Object.
@@ -51,10 +55,13 @@ def transform_page(
         "path": None,
     }
 
-    parent_page = page_translation.page.cached_parent
+    if page is None:
+        page = page_translation.page
+    parent_page = page.cached_parent
+    language = page_translation.language
     if parent_page and not parent_page.explicitly_archived:
         if parent_public_translation := parent_page.get_public_translation(
-            page_translation.language.slug,
+            language.slug,
         ):
             parent_absolute_url = parent_public_translation.get_absolute_url()
             parent = {
@@ -63,21 +70,21 @@ def transform_page(
                 "path": parent_absolute_url,
             }
             # use left edge indicator of mptt model for ordering of child pages
-            order = page_translation.page.lft
+            order = page.lft
         else:
             logger.info(
                 "The parent %r of %r does not have a public translation in %r",
                 parent_page,
-                page_translation.page,
-                page_translation.language,
+                page,
+                language,
             )
             raise Http404("No Page matches the given url or id.")
     else:
         parent = fallback_parent
         # use tree id of mptt model for ordering of root pages
-        order = page_translation.page.tree_id
+        order = page.tree_id
 
-    organization = page_translation.page.organization
+    organization = page.organization
     absolute_url = page_translation.get_absolute_url()
     return {
         "id": page_translation.id,
@@ -93,9 +100,7 @@ def transform_page(
         "parent": parent,
         "order": order,
         "available_languages": page_translation.available_languages_dict,
-        "thumbnail": (
-            page_translation.page.icon.url if page_translation.page.icon else None
-        ),
+        "thumbnail": (page.icon.url if page.icon else None),
         "organization": (
             {
                 "id": organization.id,
@@ -109,9 +114,17 @@ def transform_page(
         ),
         "hash": None,
         "embedded_offers": [
-            transform_offer(offer, page_translation.page.region)
-            for offer in page_translation.page.embedded_offers.all()
+            transform_offer(offer, page.region) for offer in page.embedded_offers.all()
         ],
+        "slug_history": list(slug_history)
+        if slug_history is not None
+        else list(
+            dict.fromkeys(
+                page_translation.all_versions.order_by("version").values_list(
+                    "slug", flat=True
+                )
+            )
+        ),
     }
 
 
@@ -135,24 +148,30 @@ def pages(
     result = []
     # The preliminary filter for explicitly_archived=False is not strictly required, but reduces the number of entries
     # requested from the database
-    for page in (
+    pages = (
         region.pages.select_related("organization__icon")
         .prefetch_related(
             "embedded_offers",
         )
         .filter(explicitly_archived=False)
         .cache_tree(archived=False, language_slug=language_slug)
-    ):
+    )
+
+    slug_history: defaultdict[int, list[str]] = get_slug_history(pages, language_slug)
+
+    for page in pages:
         if page_translation := page.get_public_translation(language_slug):
             result.append(
                 transform_page(
                     page_translation,
+                    page,
                     context={
                         "region_slug": region_slug,
                         "language_slug": language_slug,
                         "content_object": page_translation,
                         "request": request,
                     },
+                    slug_history=slug_history.get(page.id, []),
                 )
             )
     return JsonResponse(
@@ -246,12 +265,20 @@ def single_page(
         return JsonResponse(
             transform_page(
                 page_translation,
+                page,
                 context={
                     "region_slug": region_slug,
                     "language_slug": language_slug,
                     "content_object": page_translation,
                     "request": request,
                 },
+                slug_history=list(
+                    dict.fromkeys(
+                        page_translation.all_versions.order_by("version").values_list(
+                            "slug", flat=True
+                        )
+                    )
+                ),
             ),
             safe=False,
         )
@@ -318,15 +345,20 @@ def children(
             should_prefetch_nonpublic_translations=False,
         )
     )
+
+    slug_history: defaultdict[int, list[str]] = get_slug_history(pages, language_slug)
+
     result = [
         transform_page(
             page_translation,
+            page,
             context={
                 "region_slug": region_slug,
                 "language_slug": language_slug,
                 "content_object": page_translation,
                 "request": request,
             },
+            slug_history=slug_history.get(page.id, []),
         )
         for page in pages.values()
         if root_page is None or page.id != root_page.parent_id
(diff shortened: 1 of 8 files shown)


hannaseithe:
I am sorry for bringing this up again (especially since I fear, I might be simply misunderstanding things).

But:

From my understanding the app creates a link to a page based on 
A) the page tree structure + B) the involved slugs. So page url would look like this if it has a parent:
`https://integreat.app/<region_slug>/<language_slug>/<parent_page_translation_slug>/<own_page_translation_slug>`

So there can be several reasons for a page_translation url to change
1) Its own slug changes
2) the parent slug changes (or ancestor's depending on the nesting)
3) the page_translation is moved elsewhere in the page tree structure (its parent/ancestors change)

If we only return the previous slugs of a page translation (not that of its parents, nor any information about changes to page tree structure), is that truly sufficient for the app to be able to reconstruct previous paths? @steffenkleinle

steffenkleinle:
> I am sorry for bringing this up again (especially since I fear, I might be simply misunderstanding things).
> 
> But:
> 
> From my understanding the app creates a link to a page based on A) the page tree structure + B) the involved slugs. So page url would look like this if it has a parent: `https://integreat.app/<region_slug>/<language_slug>/<parent_page_translation_slug>/<own_page_translation_slug>`

The result is the same, but we actually don't construct these paths/links on our own but just use the `path` property delivered in the corresponding endpoints.

> So there can be several reasons for a page_translation url to change
> 
>     1. Its own slug changes
> 
>     2. the parent slug changes (or ancestor's depending on the nesting)
> 
>     3. the page_translation is moved elsewhere in the page tree structure (its parent/ancestors change)
> 
> 
> If we only return the previous slugs of a page translation (not that of its parents, nor any information about changes to page tree structure), is that truly sufficient for the app to be able to reconstruct previous paths? @steffenkleinle

Currently, we only match pages using the full path, i.e. including slugs of parents/ancestors. If the slug (and therefore also the path) is changed, we have currently no way to match the correct page with the old slug/path. If matching the old/legacy slug with the last segment of the path we are trying to resolve is enough (and considers all three cases mentioned), you need to tell me :D If not, I am more than happy to match whole paths as we currently do.

hannaseithe:
Thank you very much for explaining things to me here and in the call @steffenkleinle . The PR should work for the purpose of being able to redirect requests to old (non-existent) urls to its current url.

For documentation purposes I note down here what I understood, how the mechanism in the app works:
- a request comes into the app with a path, that cannot be matched, then:
- take the last part of the path (so the actual slug of an object) and find it in the previous slug property of all objects available to them (e.g. if its a page path like `page/de/augsburg`, they will only search all German page translations for Augsburg) - that way they can identify if such a page exists and redirect to its actual current link. 
- since our database constraint gurantees that even an old slug (that has been overwritten by a newer slug for that object), wont be allowed to be used for a different object again, this should guarantee that a slug will always be able to clearly identify which object (within a language and region) it once belonged to. 

One question for the app, that remains for me: Will the app be caching this previous slugs property for all objects? Because otherwise how can you be searching for it?

steffenkleinle:
> Thank you very much for explaining things to me here and in the call @steffenkleinle . The PR should work for the purpose of being able to redirect requests to old (non-existent) urls to its current url.
> 
> For documentation purposes I note down here what I understood, how the mechanism in the app works:
> 
>     * a request comes into the app with a path, that cannot be matched, then:
> 
>     * take the last part of the path (so the actual slug of an object) and find it in the previous slug property of all objects available to them (e.g. if its a page path like `page/de/augsburg`, they will only search all German page translations for Augsburg) - that way they can identify if such a page exists and redirect to its actual current link.
> 
>     * since our database constraint gurantees that even an old slug (that has been overwritten by a newer slug for that object), wont be allowed to be used for a different object again, this should guarantee that a slug will always be able to clearly identify which object (within a language and region) it once belonged to.

Nice summary, no objections!

> One question for the app, that remains for me: Will the app be caching this previous slugs property for all objects? Because otherwise how can you be searching for it?

Yes, we actually "cache" all the data on native (we save them offline), but I think the relevant factor here is actually that we don't use the individual endpoints but instead the general page endpoint that returns all pages at once. Which is also why the solution for web (where requests are automatically redirected by the CMS) does not work for native.

MizukiTemma:
@hannaseithe thank you for the summary and @steffenkleinle thank you for explaining 😸

integreat_cms/api/v3/pages.py
```diff
@@ -63,21 +66,27 @@ def transform_page(
...
         parent = fallback_parent
         # use tree id of mptt model for ordering of root pages
-        order = page_translation.page.tree_id
+        order = page.tree_id
 
-    organization = page_translation.page.organization
+    all_translation_versions = list(PageTranslation.objects.filter(page=page))
+    previous_slugs = set(
+        get_translation_slug(all_translation_versions)[page.id][language.slug].values()
+    )
```
hannaseithe:
```suggestion
if all_slugs is None:
  all_slugs = page_translation.get_all_used_slugs()
```

integreat_cms/api/v3/pages.py
```diff
@@ -63,21 +66,27 @@ def transform_page(
...
+    previous_slugs.discard(page_translation.slug)
```
hannaseithe: In my conversation with @steffenkleinle , he mentioned that he would actually prefer to have the current slug in the list/array . But then a renaming of that property would likely make sense - maybe just `all_slugs` or `used_slugs`

integreat_cms/api/v3/pages.py
hannaseithe:
```suggestion
    page_translation: PageTranslation, 
    page: Page | None = None,
    context: dict[str, Any] | None = None,
    all_slugs: set[str] | None = None
```

integreat_cms/api/v3/pages.py
```diff
@@ -51,10 +52,12 @@ def transform_page(
         "path": None,
     }
 
-    parent_page = page_translation.page.cached_parent
+    page = page_translation.page
```
hannaseithe:
```suggestion
if page is None:
    page = page_translation.page
```

integreat_cms/api/v3/pages.py
hannaseithe, line 175:
```suggestion
transform_page(
page_translation,
page,
context={
"region_slug": region_slug,
"language_slug": language_slug,
"content_object": page_translation,
"request": request,
},
all_slugs=slug_history.get(page.id,set())
)
```

hannaseithe:
```suggestion
    slug_history = defaultdict(set)
    slug_query = PageTranslation.objects.filter(page__region=region,language_slug=language_slug).values("page_id", "slug")
    for row in slug_query:
        slug_history[row["page_id"]].add(row["slug"])

    pages_list = (
        region.pages.select_related("organization__icon")
        .prefetch_related(
            "embedded_offers",
        )
        .filter(explicitly_archived=False)
        .cache_tree(archived=False, language_slug=language_slug)
    )

    slug_history: defaultdict[int, set[str]] = defaultdict(set)
    for row in PageTranslation.objects.filter(page_id__in=pages_list,language__slug=language_slug).values("page_id","slug"):
        slug_history[row["page_id"]].add(row["slug"])

    for page in pages_list:
```

hannaseithe (changes requested):
Thank you for your PR, I think the approach is correct, but the problem is that with accessing the slugs inside transform_page we run into a n+1 query problem (both on the pages endpoint as well as the children endpoint.), by querying the slugs for all versions inside the transform_page method.

Therefore I suggest that we query all slugs in a single pre-query on all pages before we call transform_page. Also I hand in the page, so we can save that extra query as well. I have added my suggestions for the pages endpoint, but suggest to implement the same for the children endpoint as well

integreat_cms/api/v3/pages.py
steffenkleinle: Or `slug_history` :)

MizukiTemma: Thank you for sharing it with me :)

MizukiTemma:
@hannaseithe 

Thank you for review and improvement idea 😸 They are now applied.

integreat_cms/api/v3/pages.py
```diff
@@ -29,13 +30,17 @@
 
     from django.http import HttpRequest
 
-    from ...cms.models import Page, PageTranslation
+    from ...cms.models import Page
+    from ..models.pages.page import PageQuerySet
```
hannaseithe:
```suggestion
    from ...cms.models.pages.page import PageQuerySet
```

integreat_cms/api/v3/pages.py
```diff
@@ -29,13 +30,17 @@
 
     from django.http import HttpRequest
 
-    from ...cms.models import Page, PageTranslation
+    from ...cms.models import Page
```
hannaseithe: This import is redundant

integreat_cms/api/v3/pages.py
```diff
@@ -447,3 +475,23 @@ def push_page_translation_content(
         return JsonResponse({"status": "success"})
     logger.error("Push Content: failed to validate page translation.")
     return JsonResponse({"status": "error"}, status=405)
+
+
+def get_slug_history(
+    pages: PageQuerySet, language_slug: str
```
hannaseithe:
```suggestion
    pages: PageQuerySet | list[Page], language_slug: str
```
because of the call insie `get_public_ancestor_translations`

hannaseithe (approved):
I really like your improvements on my suggestions! Only some minor change requests left, that claude found. I approve ahead of time

integreat_cms/api/v3/pages.py
```diff
@@ -109,9 +114,11 @@ def transform_page(
         ),
         "hash": None,
         "embedded_offers": [
-            transform_offer(offer, page_translation.page.region)
-            for offer in page_translation.page.embedded_offers.all()
+            transform_offer(offer, page.region) for offer in page.embedded_offers.all()
         ],
+        "slug_history": list(slug_history)
+        if slug_history
+        else page_translation.get_all_used_slugs(),
```
jonbulz: I think this `else` branch is currently unreachable by construction (we don't allow empty slugs, and every page that reaches `transform_page` has at least one public translation in this language, thus the set can't be empty). But it would crash if it ever executed because `get_all_used_slugs` returns a lazy `Queryset`, which is not json serializable. Simple fix is to wrap the `list()` around the `else` branch as well

jonbulz (changes requested):
Looks very good in general! +1 for keeping an eye on performance :)
There's one thing I think needs fixing, other than that we're good to go!

integreat_cms/api/v3/pages.py
MizukiTemma: Nice catch 👁️ 

Fixed!

MizukiTemma:
@jonbulz 
It's ready for re-review 😸

MizukiTemma:
It seems the order of slugs in `slug_history´ is flaky in the tests 😭

hannaseithe:
> It seems the order of slugs in `slug_history´ is flaky in the tests 😭

I found the problem, the devil is in the usage of `set`, which does not guarantee order. As discussed I implemented a fix and will commit it.

MizukiTemma:
@hannaseithe 

I double-checked and it sitll works as expected ✅ 

Thank you for analysis and fix implementation 🙏

jonbulz (approved):
Thanks, looks good! 👍

MizukiTemma:
@steffenkleinle FYI: merged 😁

PR #4328, opened 2026-06-08
Add new app slugs

steffenkleinle:
### Short description

With our renaming in https://github.com/digitalfabrik/integreat-app/issues/4045, we changed some of our routes:
- https://integreat.app/landing -> https://integreat.app/regions
- https://integreat.app/testumgebung/de/locations -> https://integreat.app/testumgebung/de/places

This PR adds the new slugs `regions` and `places` to the reserved slugs.

### Proposed changes

- Add `regions` to `RESERVED_REGION_SLUGS`
- Add `places` to `RESERVED_REGION_PAGE_PATTERNS`

### Side effects

N/A

No issue.

### How to test

Test that it is not possible anymore to create a page with slug `places` and it is not possible to create a region with the name `regions`.

### Resolved issues

Fixes: N/A

### DIFF
--- a/integreat_cms/core/settings.py
+++ b/integreat_cms/core/settings.py
@@ -72,7 +72,8 @@ TEST_REGION_SLUG: Final[str] = "testumgebung"
 
 #: Reserved region slugs, that are used in the webapp
 RESERVED_REGION_SLUGS: Final[list[str]] = [
-    "landing",
+    "landing",  # Legacy, now regions
+    "regions",
     "recommend",
     "licenses",
     "main-disclaimer",
@@ -86,7 +87,8 @@ RESERVED_REGION_PAGE_PATTERNS: Final[list[str]] = [
     IMPRINT_SLUG,
     "news",
     "events",
-    "locations",
+    "locations",  # Legacy, now places
+    "places",
     "offers",
     "search",
 ]


MizukiTemma:
@steffenkleinle 

Thank you for the PR 😸 It looks good.

> Test that it is not possible anymore to create a page with slug places and it is not possible to create a region with the name regions.

Both constraints now works but it's possible that there is (already) a page/region with slug "places"/"regions"in the system. I recommend to add a migration to go safely.

steffenkleinle:
> @steffenkleinle
> 
> Thank you for the PR 😸 It looks good.
> 
> > Test that it is not possible anymore to create a page with slug places and it is not possible to create a region with the name regions.
> 
> Both constraints now works but it's possible that there is (already) a page/region with slug "locations"/"regions"in the system. I recommend to add a migration to go safely.

@MizukiTemma would you mind taking over?

MizukiTemma:
> > @steffenkleinle
> > Thank you for the PR 😸 It looks good.
> > > Test that it is not possible anymore to create a page with slug places and it is not possible to create a region with the name regions.
> > 
> > 
> > Both constraints now works but it's possible that there is (already) a page/region with slug "locations"/"regions"in the system. I recommend to add a migration to go safely.
> 
> @MizukiTemma would you mind taking over?

Done 💪

MizukiTemma (approved):

integreat_cms/cms/migrations/0154_replace_newly_added_app_slugs.py
```diff
@@ -0,0 +1,55 @@
...
+                "slug": translation.slug,
+            }
+            translation.slug = generate_unique_slug(**slug_kwargs)
+            translation.save()
+
+    regions = Region.objects.filter(slug="regions")
+    if regions:
+        counter = 1
+        for region in regions:
+            region.slug = "regions-" + str(counter)
```
dkehne: `Region.slug` is `unique=True`, so this raises an `IntegrityError` if a region with slug `regions-1` already exists — and since at most one region can have the slug `regions`, the counter can never go beyond 1 anyway. Using `generate_unique_slug` like in the page branch above would handle both:

```python
region.slug = generate_unique_slug(
    slug=region.slug,
    manager=Region.objects,
    object_instance=region,
    foreign_model="region",
)
```

integreat_cms/cms/migrations/0154_replace_newly_added_app_slugs.py
```diff
@@ -0,0 +1,55 @@
...
+                "foreign_model": "page",
+                "foreign_object": translation.page,
+                "language": translation.language,
+                "manager": PageTranslation.objects,
+                "object_instance": translation,
+                "region": translation.page.region,
+                "slug": translation.slug,
+            }
+            translation.slug = generate_unique_slug(**slug_kwargs)
+            translation.save()
```
dkehne: Calling the live model's `save()` here has two side effects:
- the HIX `pre_save` signal fires for every matched version; for old versions whose content differs from the latest it makes a synchronous Textlab API call inside the migration transaction and overwrites `hix_score`,
- `last_updated` is bumped on all matched versions (the queryset includes historic versions, not just the latest).

Wrapping the loop in `disable_hix()` and calling `save(update_timestamp=False)` (as `cleanup_autosaves` does) would avoid both.

integreat_cms/cms/migrations/0154_replace_newly_added_app_slugs.py
```diff
@@ -0,0 +1,55 @@
...
+    from django.db.backends.base.schema import BaseDatabaseSchemaEditor
+
+
+def replace_newly_added_app_slugs(
+    _apps: Apps, _schema_editor: BaseDatabaseSchemaEditor
+) -> None:
+    """
+    Page translations with slug `places` and any region with slug `regions` will receive a new slug
+    """
+    translations = PageTranslation.objects.filter(slug="places")
```
dkehne: These renames change public app URLs, and since the reverse migration is a noop, the old→new mapping is lost once applied. A short `print()` per rename (similar to `print_action_required` in 0153 or the logging in `make_all_slugs_unique`) would give operators an audit trail for setting up redirects or notifying affected regions.

integreat_cms/cms/migrations/0154_replace_newly_added_app_slugs.py
```diff
@@ -0,0 +1,55 @@
+# Generated by Django 5.2 on 2026-04-27 14:42
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from django.db import migrations
+
+from ..models import PageTranslation, Region
```
dkehne: Most of our other data migrations use `apps.get_model()` instead of importing the live models, so they keep working for fresh installs after future model changes. I realize `generate_unique_slug` needs the real model instances, so this isn't a trivial swap — but it means this migration will run the then-current `save()`/`clean()` logic against the 0154-era schema, which may be worth a short comment documenting the trade-off.

integreat_cms/cms/migrations/0154_replace_newly_added_app_slugs.py
```diff
@@ -0,0 +1,55 @@
...
+) -> None:
+    """
+    Page translations with slug `places` and any region with slug `regions` will receive a new slug
+    """
+    translations = PageTranslation.objects.filter(slug="places")
+
+    if translations:
+        for translation in translations:
+            slug_kwargs = {
+                "fallback": None,
```
dkehne: nit picky: `"fallback": None` doesn't match `SlugKwargs` (`fallback: NotRequired[str]`) — the key could simply be omitted. Same for the `if translations:` / `if regions:` guards, which are redundant around the loops.

dkehne (commented):
Took a closer look at the migration — a few notes inline.

MizukiTemma:
@dkehne 
Thank you for review 😸 

Hmm 🤔 It seems to be not worth to implement a migration. The change does not break the database, and I suggested a migration only for clean consistency with the new constraints. After checking all the points you mentioned I decided to remove it.

dkehne:
Thanks to Fabel 5 (test ballon) :D I also messaged @steffenkleinle that i did some testing here...

jonbulz (approved):
LGTM 👍

PR #4329, opened 2026-06-08
Fix timezone awareness for external calendar times

### LINKED ISSUE #4312: Event import from .ics ignores UTC timezone offset (Z suffix)
### Bug Summary

When importing events from external calendars (.ics), UTC timestamps (indicated by the `Z` suffix per RFC 5545) are displayed with the wrong time — the UTC offset is not applied.

**Example:** An event with `DTSTART:20260526T063000Z` (06:30 UTC) is shown as **06:30** in Integreat, but should be **08:30** (UTC+2, Europe/Berlin in summer).

### Expected Behavior

UTC times (and any timezone-aware times) from imported ICS files should be converted to the region's local timezone before being stored. `DTSTART:20260526T063000Z` in a region with `Europe/Berlin` (UTC+2 in summer) should result in an event displayed at **08:30**.

The fix must handle all three RFC 5545 time forms correctly:

| ICS form | Example | What `icalendar` returns | Correct handling |
|---|---|---|---|
| UTC (`Z` suffix) | `20260526T063000Z` | aware datetime (UTC) | convert via `.astimezone(region_tz)` |
| `TZID=...` | `DTSTART;TZID=America/New_York:20260526T063000` | aware datetime (that zone) | convert via `.astimezone(region_tz)` |
| floating / naive | `20260526T063000` | naive datetime | leave as-is (interpret as local wall-clock — current behavior) |

### Actual Behavior

UTC-aware datetimes are not converted to the region timezone before being split into date/time components. The timezone info is lost, and the naive time is then re-labelled (not converted) as the region timezone:

In [`external_calendar_utils.py`](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/utils/external_calendar_utils.py) (around line 120):

```python
# start = datetime(2026, 5, 26, 6, 30, 0, tzinfo=UTC)  ← timezone-aware, UTC
start_date, start_time = start.date(), start.time()
# start_time = time(6, 30, 0)  ← tzinfo is LOST here
```

Later, in [`event_form.py`](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/forms/events/event_form.py) (around line 249–253):

```python
tzinfo = zoneinfo.ZoneInfo(self.instance.timezone)  # e.g. "Europe/Berlin"
cleaned_data["start"] = datetime.combine(
    cleaned_data["start_date"],
    cleaned_data["start_time"],  # time(6, 30) — UTC context already lost
).replace(tzinfo=tzinfo)  # treats 6:30 *as* Berlin time instead of converting from UTC
```

`.replace()` re-labels the time without converting, so 06:30 UTC becomes 06:30 Europe/Berlin instead of 08:30 Europe/Berlin.

Note: for **manually** entered events `.replace()` is correct — the user types "08:30" meaning local time. The `EventForm` is shared with the manual UI, so the conversion must happen at the **import boundary**, not in the form.

### Steps to Reproduce

1. Create an external calendar in a region with timezone `Europe/Berlin`
2. Import an ICS file with a UTC-time event:
   ```
   DTSTART:20260526T063000Z
   DTEND:20260526T083000Z
   ```
3. The imported event shows **06:30–08:30** instead of the correct **08:30–10:30**

### Reproducibility

Always (100%)

### Environment

- **Environment:** Development / Production
- **URL (if applicable):** —
- **App version / commit:** develop
- **Browser:** —
- **OS:** —
- **Device:** —
- **User role / permissions:** Any user with calendar import access

_No visual evidence required — the bug is deterministic and reproducible with any ICS file containing UTC timestamps._

### Severity

**Workaround:** Manually edit the imported event time after import.

### Regression Information

- **Is this a regression?** Unknown — the `Z` suffix handling may never have worked correctly.
- **Last known working version / date:** Unknown
- **Suspected cause:** The `EventForm` was designed for manual wall-clock input and `.replace()` is correct there; when it was reused for ICS import the UTC conversion was never added.

#### Background: How datetimes work in the system

- `core/settings.py`: `USE_TZ = True`, `TIME_ZONE = "UTC"` → every `DateTimeField` is stored as an **absolute UTC instant** in the DB.
- `Event.start`/`end` are plain `DateTimeField`s → also stored in UTC.
- Conversion happens only at **display** time via the `start_local`/`end_local` properties ([`event.py:215–234`](https://github.com/digitalfabrik/integreat-cms/blob/develop/integreat_cms/cms/models/events/event.py)), which activate `region.timezone` and call `timezone.localtime()`.

#### Suggested Fix

Convert at the import boundary, before splitting into date/time components, and only when the datetime is timezone-aware:

```python
region_tz = zoneinfo.ZoneInfo(region.timezone)
if start.tzinfo is not None:          # aware (Z or TZID) → convert
    start = start.astimezone(region_tz)
start_date, start_time = start.date(), start.time()
```

The `tzinfo is not None` guard makes both UTC and `TZID` cases correct while leaving floating times untouched. This should be applied consistently to `DTSTART`, `DTEND`, and the `UNTIL` value in recurrence rules (which also carries a `Z` suffix per RFC 5545).

- [ ] ICS event with `DTSTART:...Z` (UTC) is imported with the correct local time for the region timezone
- [ ] ICS event with `DTSTART;TZID=...` is imported with the correct local time for the region timezone
- [ ] ICS event with a floating/naive `DTSTART` (no `Z`, no `TZID`) continues to work as before
- [ ] `DTEND` and `UNTIL` in recurrence rules are handled consistently with `DTSTART`
- [ ] Manually created events are unaffected

### Verification / Testing

- [x] Reproduce original bug with `DTSTART:20260526T063000Z` → confirm displayed time is now 08:30 instead of 06:30
- [x] Test with `DTSTART;TZID=America/New_York:20260526T063000` → confirm correct conversion to region timezone
- [x] Test with floating `DTSTART:20260526T063000` → confirm no change in behavior
- [x] Test recurring event with `UNTIL=...Z` → confirm correct handling
- [ ] Run existing event import tests

### Related Issues

_None known._

hannaseithe:
### Short description

- This fixes the bug that timezone aware times were ignored when importing external calendars

### Proposed changes

- oarse end and start time with timezone awareness
- add test

### Side effects

- None

### How to test

- create a calendar in next_cloud with an event that has a different timezone time for the start and end event -> copy link to calendar
- import calendar through CMS GUI and see that the correct times were created

### Resolved issues

Fixes: #4312 


hannaseithe:
Converted to draft as tests are failing

MizukiTemma (approved):
Clean implementation and test 😸 Thank you!

jarlhengstmengel (approved):
Thank you very much, looks good! Times were correctly imported from my testcalender :)

PR #4341, opened 2026-06-10
preserve Link.ignore flag through content version updates

### LINKED ISSUE #3859: White list for functioning URLs
### Motivation

Sometimes an URL gets judged as broken link and lands on the group "Verification needed", even if it is working fine. Users can mark such links as verified by bulk action. However, when a content with such a link is updated, the URL appears again in "Verification needed". This is a bad user experience.

### Proposed Solution

Create a white list, where URLs can be stored. All links referencing one of those URLs get judged as valid.

Any alternative idea is welcomed.

### Additional Context

How to enjoy this sad user experience:

1. Go to a page
2. Embed a link https://www.flaticon.com/
3. Find this URL in "Verification needed"
4. Select the URL and execute the bulk action "Mark as verified"
5. See the URL moved to "Link verified"
6. Go to the page where you embedded the URL
7. Change something in the content and update the page
8. See the URL is again in "Verification needed"

jonbulz:
### Short description

Broken links can be marked as verified in the link checker dashboard, which sets the `Link.ignore` flag on the model. Thus far, these flags did not survive content updates or management commands like `fix_internal_links`, because the `Link` and `Url` objects in the database are deleted and re-created, thus losing the `ignore=True` information. This PR attempts to carry the `ignore` flag across different versions of the same link.

### Proposed changes

There are several issues that make this not straight forward:
- the `Link` and `Url` objects are created by `linkcheck.worker_tasks:do_check_instance_links`, which is in the `linkcheck` library and not our own codebase
- we manually remove the `Link` objects on a content translation object when a new version is created to prevent accumulating `Link`s in the DB. this happens before the new content translation object is created, and thus before the new `Link` objects exist
- `linkcheck` runs in celery on production, so the process that creates the new `Link` objects is different then the process that deletes the old ones

So this PR 
- adds a context manager `preserve_ignored_links` that snapshots `Link` objects with `ignore=True` that are deleted and writes them to the cache (TTL 10 minutes)
- monkey-patches `do_check_instance_links` to read the cache and transfer the `ignore` flag to the newly created `Link`s

Note that the links are identified by their target, which is the content object they resolve to for internal links and the url for external links. Using the Url row id for reference was not sufficient since the Url rows also get deleted and re-created.
### Side effects

- every linkcheck trigger now runs through the new wrapper. worth keeping an eye on performance
- if linkcheck executes after the 10 minute TTL expired (celery worker down, or big queue backlog), the flag won't be transfered

XLIFF import path is not covered, thus the `ignore` flags are still lost there

### How to test

- create a page with a broken link
- mark the link as verified in the dashboard
- edit the page and update to create a new version
- see that the broken link remains listed as verified

### Resolved issues

Fixes: #3859 

### DIFF
--- a/integreat_cms/cms/apps.py
+++ b/integreat_cms/cms/apps.py
@@ -9,7 +9,9 @@ from django.utils.translation import gettext_lazy as _
 if TYPE_CHECKING:
     from typing import Final
 
+    from django.db.models import Model
     from django.utils.functional import Promise
+    from linkcheck import Linklist
 
 logger = logging.getLogger(__name__)
 
@@ -29,11 +31,57 @@ class CmsConfig(AppConfig):
     verbose_name: Final[Promise] = _("CMS")
 
     def ready(self) -> None:
-        """
-        Monkeypatch the checking of internal URLs
-        """
-        from linkcheck.models import Url
+        from django.core.cache import cache
+        from linkcheck import celery as linkcheck_celery
+        from linkcheck import listeners as linkcheck_listeners
+        from linkcheck import worker_tasks as linkcheck_worker_tasks
+        from linkcheck.models import Link, Url
 
+        from .models.abstract_content_translation import AbstractContentTranslation
         from .utils.internal_link_checker import check_internal
+        from .utils.link_ignore_preservation import cache_key_for, target_key
 
         Url.check_internal = check_internal
+
+        # Reapply ignore=True to Link rows that were stashed by a
+        # preserve_ignored_links context manager at the delete site.
+        upstream = linkcheck_worker_tasks.do_check_instance_links
+
+        def do_check_instance_links_preserving_ignore(
+            sender: type[Model],
+            instance: Model,
+            linklist_cls: type[Linklist],
+            wait: bool = False,
+        ) -> None:
+            cache_key = (
+                cache_key_for(instance)
+                if isinstance(instance, AbstractContentTranslation)
+                else None
+            )
+            stashed = cache.get(cache_key) if cache_key else None
+            upstream(sender, instance, linklist_cls, wait=wait)
+            if stashed:
+                ignored_keys = {tuple(k) for k in stashed}
+                cache.delete(cache_key)
+                to_re_ignore = [
+                    link.pk
+                    for link in Link.objects.filter(
+                        content_type=linklist_cls.content_type(),
+                        object_id=instance.pk,
+                    ).select_related("url")
+                    if target_key(link.url.url) in ignored_keys
+                ]
+                if to_re_ignore:
+                    Link.objects.filter(pk__in=to_re_ignore).update(ignore=True)
+
+        # Patch all three references — listeners and celery imported the
+        # name at module load.
+        linkcheck_worker_tasks.do_check_instance_links = (
+            do_check_instance_links_preserving_ignore
+        )
+        linkcheck_listeners.do_check_instance_links = (
+            do_check_instance_links_preserving_ignore
+        )
+        linkcheck_celery.do_check_instance_links = (
+            do_check_instance_links_preserving_ignore
+        )
--- a/integreat_cms/cms/utils/link_ignore_preservation.py
+++ b/integreat_cms/cms/utils/link_ignore_preservation.py
@@ -1,0 +1,69 @@
+"""
+Carry editor-set ``Link.ignore`` flags across the delete-and-recreate
+cycle that runs on every translation save.
+"""
+
+from __future__ import annotations
+
+from contextlib import contextmanager
+from typing import TYPE_CHECKING
+
+from django.contrib.contenttypes.models import ContentType
+from django.core.cache import cache
+
+from . import internal_link_utils
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+    from ..models.abstract_content_translation import AbstractContentTranslation
+
+
+_CACHE_TTL = 600
+_CACHE_PREFIX = "linkcheck_pending_ignore"
+
+
+def target_key(url_string: str) -> tuple[str, object]:
+    """
+    Stable identity for a URL's target. Internal URLs that resolve to a
+    content object share a key across slug rewrites; everything else
+    falls back to the URL string.
+    """
+    resolved = internal_link_utils.get_public_translation_for_link(url_string)
+    if resolved is None:
+        return ("ext", url_string)
+    return (type(resolved).__name__, resolved.foreign_object.pk)
+
+
+def cache_key_for(translation: AbstractContentTranslation) -> str:
+    foreign = translation.foreign_object
+    foreign_ct = ContentType.objects.get_for_model(foreign)
+    return f"{_CACHE_PREFIX}:{foreign_ct.id}:{foreign.pk}:{translation.language_id}"
+
+
+@contextmanager
+def preserve_ignored_links(
+    translation: AbstractContentTranslation,
+) -> Iterator[None]:
+    """
+    Snapshot target-keys of ignored Links on ``translation`` so the next
+    ``do_check_instance_links`` for the same (foreign_object, language)
+    can restore the flag on the freshly reconciled Links.
+
+    Wrap the delete-and-save sequence:
+
+        with preserve_ignored_links(self):
+            self.links.all().delete()
+            new_translation.save()
+
+    The follow-up reconciliation may run synchronously or via Celery on
+    transaction commit; the stash is shared via the Django cache so both
+    paths see it.
+    """
+    keys = [
+        target_key(link.url.url)
+        for link in translation.links.filter(ignore=True).select_related("url")
+    ]
+    if keys:
+        cache.set(cache_key_for(translation), keys, timeout=_CACHE_TTL)
+    yield
--- a/integreat_cms/cms/forms/custom_content_model_form.py
+++ b/integreat_cms/cms/forms/custom_content_model_form.py
@@ -14,6 +14,7 @@ from ..utils.content_translation_utils import (
     update_links_to,
 )
 from ..utils.content_utils import clean_content
+from ..utils.link_ignore_preservation import preserve_ignored_links
 from ..utils.slug_utils import generate_unique_slug_helper
 from .custom_model_form import CustomModelForm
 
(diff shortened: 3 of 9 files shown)


integreat_cms/cms/apps.py
```diff
@@ -29,11 +31,58 @@ class CmsConfig(AppConfig):
...
-        """
-        from linkcheck.models import Url
+        from django.core.cache import cache
+        from linkcheck import celery as linkcheck_celery
+        from linkcheck import listeners as linkcheck_listeners
+        from linkcheck import worker_tasks as linkcheck_worker_tasks
+        from linkcheck.models import Link, Url
 
         from .utils.internal_link_checker import check_internal
+        from .utils.link_ignore_preservation import cache_key_for, target_key
```
andrew8er, line 42: ❓ Is there a specific reason these imports are not at the top level?

integreat_cms/cms/apps.py
```diff
@@ -29,11 +31,58 @@ class CmsConfig(AppConfig):
...
+            cache_key = (
+                cache_key_for(instance)
+                if hasattr(instance, "foreign_object")
+                and hasattr(instance, "language_id")
+                else None
+            )
+            stashed = cache.get(cache_key) if cache_key else None
+            upstream(sender, instance, linklist_cls, wait=wait)
+            if not stashed:
+                return
```
andrew8er: 🔧 Purely stylistic, but I would prefer this condition reversed and the following code indented under it. This way it is clear from the indentation, that this code is only executed when this condition is true, without having to scan the code for early returns.

integreat_cms/cms/apps.py
```diff
@@ -29,11 +31,58 @@ class CmsConfig(AppConfig):
...
+        from django.core.cache import cache
+        from linkcheck import celery as linkcheck_celery
+        from linkcheck import listeners as linkcheck_listeners
+        from linkcheck import worker_tasks as linkcheck_worker_tasks
+        from linkcheck.models import Link, Url
 
         from .utils.internal_link_checker import check_internal
+        from .utils.link_ignore_preservation import cache_key_for, target_key
 
         Url.check_internal = check_internal
```
andrew8er, line 44: 🔧 Unrelated to this PR, but this version does not get flagged by type checkers:
```suggestion
        @functools.wraps(check_internal)
        def check_internal_method(self: Url) -> bool | None:
            return check_internal(self)

        Url.check_internal = check_internal_method
```

integreat_cms/cms/apps.py
```diff
@@ -29,11 +31,58 @@ class CmsConfig(AppConfig):
...
+            sender: type[Model],
+            instance: Model,
+            linklist_cls: type[Linklist],
+            wait: bool = False,
+        ) -> None:
+            cache_key = (
+                cache_key_for(instance)
+                if hasattr(instance, "foreign_object")
+                and hasattr(instance, "language_id")
+                else None
```
andrew8er, line 59: ❓ I get a type checking error here:
```
Argument of type "Model" cannot be assigned to parameter "translation" of type "AbstractContentTranslation" in function "cache_key_for"
  "Model" is not assignable to "AbstractContentTranslation"
```
Would this version also work, or is it too strict?
```suggestion
                cache_key_for(instance)
                if isinstance(instance, AbstractContentTranslation)
                else None
```

integreat_cms/cms/utils/link_ignore_preservation.py
```diff
@@ -0,0 +1,69 @@
...
+def target_key(url_string: str) -> tuple[str, object]:
+    """
+    Stable identity for a URL's target. Internal URLs that resolve to a
+    content object share a key across slug rewrites; everything else
+    falls back to the URL string.
+    """
+    resolved = internal_link_utils.get_public_translation_for_link(url_string)
+    if resolved is None:
+        return ("ext", url_string)
+    return (type(resolved).__name__, resolved.foreign_object.pk)
```
andrew8er, line 35: 🔧 
```suggestion
    return (
        ("ext", url_string)
        if resolved is None
        else (type(resolved).__name__, resolved.foreign_object.pk)
    )
```

andrew8er (commented):
So, the implementation looks pretty good to me.

It is however quiet complex, and I wonder if we could achieve this in a simpler way. From what I could gather, the `linkchecker` library is a separate project, but we are already using our [own fork here](https://github.com/digitalfabrik/django-linkcheck).

Since we have quiet unique requirements in our CMS, would it not be beneficial to adapt our fork? We would then not be able to benefit from upstream again, but looking at the [commits](https://github.com/DjangoAdminHackers/django-linkcheck/commits/master/) it seems that this is already a practically "finished" project. There are hardly any new commits there and I would not expect any bigger updates that we might benefit from. I would even vendor it into this repository actually.

Would this be possible or did I miss a problem here?

Also, could you rebase the branch on `develop`?

integreat_cms/cms/apps.py
jonbulz: Yes. We'd end up with an import chain like: 
apps.py -> link_ignore_preservation -> internal_link_utils -> from ..models import (...) 
but every model that goes through `ModelBase` can only be imported after `check_apps_ready` is True, which can't possibly happen before `apps.py` even initialized

integreat_cms/cms/apps.py
jonbulz: what type checker are your using? I don't see why this would make any difference 🤔 functionally, it should be the same anyway

integreat_cms/cms/apps.py
jonbulz: Good catch, probably better to be strict here. Needs the import though 😉

integreat_cms/cms/utils/link_ignore_preservation.py
jonbulz: Is this just a style thing? Because I actually find my version more readable 😅

integreat_cms/cms/apps.py
jonbulz: I got used to early returns and I actually kinda like it, but I don't have a strong opinion. So I changed it if you think this makes it more readable

jonbulz:
@andrew8er thanks for your review! I answered to your comments and applied (most of) your suggestions. Happy to discuss things further if you want.

Yes, I suppose we could save ourselves the monkey-patching if we committed to using our fork from now on, but last I asked, we were still hoping to return to upstream linkcheck at some point. Since[ #220](https://github.com/DjangoAdminHackers/django-linkcheck/pull/220#issuecomment-3646751134) got rejected, unfortunately we'll likely have to wait for Django 6.0 for that. That would require some rework in the methods introduced here either way. But I'd like to get rid of the fork if there is any hope

integreat_cms/cms/apps.py
andrew8er: We could consider upgrading to Python 3.15 and use lazy imports for these situations: https://blog.jetbrains.com/pycharm/2026/06/explicit-lazy-imports-are-coming-to-python-3-15/

integreat_cms/cms/apps.py
andrew8er: I used Zed with uses [basedpyright](https://github.com/DetachHead/basedpyright) by default. Do we have any type checker requiered for this project? I think we should have one actually.

integreat_cms/cms/utils/link_ignore_preservation.py
andrew8er: Functionality wise, it boils down to a style thing, yes. I just prefer the expression syntax more than the imperative form, but you can keep it this way.

andrew8er (approved):
Looking at the upstream project, it does seem to be pretty dormant. Not much discussion in the PRs, and a few are pretty old but never merged/declined. I would suggest merging this PR now, and then we could think about how to go on further. 

Personally, I would not mind vendoring it in and adapt it to our needs.

MizukiTemma:
Thank you for tackling this topic, I find the implementation pretty cool 😸 and it's also working as expected ✅ 

There is one concern: when a link is embedded into a German page, get marked as "verifired" and the page is translated into another language by MT, the link appears again in "Verification needed" because of the translation in the new language. I do not consider this as bug within the scope of current implementation but it needs some explanation to Service Team (or we cover such cases too) 🤔

jonbulz:
> Thank you for tackling this topic, I find the implementation pretty cool 😸 and it's also working as expected ✅
> 
> There is one concern: when a link is embedded into a German page, get marked as "verifired" and the page is translated into another language by MT, the link appears again in "Verification needed" because of the translation in the new language. I do not consider this as bug within the scope of current implementation but it needs some explanation to Service Team (or we cover such cases too) 🤔

If I understand you correctly, the link to the content in the new language appears in "Verification needed". But that is the expected and correct behavior. The new translation will point to the content in a different language, which is not only a different href but really a different content object (because it is a different ContentTranslation instance). So just binning every link from every language and marking everything as verified would not be correct IMO. Or do I missunderstand something?
Either way, I wouldn't crack my head open about it because internal links will be replaced by shortcode soon (hopefully), and then all of that will be different yet again

MizukiTemma:
> > Thank you for tackling this topic, I find the implementation pretty cool 😸 and it's also working as expected ✅
> > There is one concern: when a link is embedded into a German page, get marked as "verifired" and the page is translated into another language by MT, the link appears again in "Verification needed" because of the translation in the new language. I do not consider this as bug within the scope of current implementation but it needs some explanation to Service Team (or we cover such cases too) 🤔
> 
> If I understand you correctly, the link to the content in the new language appears in "Verification needed". But that is the expected and correct behavior. The new translation will point to the content in a different language, which is not only a different href but really a different content object (because it is a different ContentTranslation instance). So just binning every link from every language and marking everything as verified would not be correct IMO. Or do I missunderstand something? Either way, I wouldn't crack my head open about it because internal links will be replaced by shortcode soon (hopefully), and then all of that will be different yet again

Thank you for reply 😸 You understood what I meant, and I agree it's a correct behaviour. The concern is rather about users and Service Team probably expect URLs/links stay in the verified column once they have manually marked, regardless of language etc. But I guess such cases are not so often and we can communicate with them.

MizukiTemma (approved):

juliankehne:
> > > Thank you for tackling this topic, I find the implementation pretty cool 😸 and it's also working as expected ✅
> > > There is one concern: when a link is embedded into a German page, get marked as "verifired" and the page is translated into another language by MT, the link appears again in "Verification needed" because of the translation in the new language. I do not consider this as bug within the scope of current implementation but it needs some explanation to Service Team (or we cover such cases too) 🤔
> > 
> > 
> > If I understand you correctly, the link to the content in the new language appears in "Verification needed". But that is the expected and correct behavior. The new translation will point to the content in a different language, which is not only a different href but really a different content object (because it is a different ContentTranslation instance). So just binning every link from every language and marking everything as verified would not be correct IMO. Or do I missunderstand something? Either way, I wouldn't crack my head open about it because internal links will be replaced by shortcode soon (hopefully), and then all of that will be different yet again
> 
> Thank you for reply 😸 You understood what I meant, and I agree it's a correct behaviour. The concern is rather about users and Service Team probably expect URLs/links stay in the verified column once they have manually marked, regardless of language etc. But I guess such cases are not so often and we can communicate with them.

I also just found this behavior when testing it. I'd also say it's not scope of this issue and agree we should communicate with the service team that this is how it works when a page gets translated again (for now).

PR #4344, opened 2026-06-11
Update eslint&packages and switch out airbnb config

### LINKED ISSUE #3539: Update `eslint` & migrate `eslint` config
### Motivation

In the newest version of `eslint`, `.eslintrc.js` is deprecated in favor of `eslint.config.mjs`. We need to migrate in order to update `eslint`.

### Proposed Solution

There is an automatic migration tool, but unfortunately only for `.eslint.json` files, not `.js` files. There might be workarounds, or we might have to migrate manually.

Once this migration is complete, all `eslint` related packages need to be updated; *hopefully*, this resolves the update issues we have been having with e.g. #3536. Investigating this should be part of a PR fixing this issue. If it does not solve these problems, a separate issue should be opened.

### Alternatives

none, this blocks updating JS deps

### Related Issues

#3536

hannaseithe:
### Short description

By switching out the old `eslint-config-airbnb` package for `eslint-config-airbnb-extended` (which is actively maintained), we are now able to upgrade to eslint 9 and switch to a flat config. eslint 10 is still blocked because `eslint-config-airbnb-extended` does not support it yet

### Proposed changes

- switch `eslint-config-airbnb` package for `eslint-config-airbnb-extended`
- upgrade eslint to `9.39.4`
- update a few more eslint plugins
- migrate `.eslintrc.js` to `eslint.config.mjs` (flat config)
- fix a few minor warnings and errors

### Side effects

- none

### How to test

- this is only testable by the reviewers in the code:
  - check if eslint.sh gives no errors
  - spot test if rules are still applied, eg:
  
  ```
    Equality/comparison
  - == instead of === (non-null) → should error

  Variables
  - var x = 1 → should error (no-var)
  - let x = 1 where x is never reassigned → should error (prefer-const)

  Functions
  - Regular function expression const f = function() {} → should error (prefer-arrow/prefer-arrow-functions)

  Imports
  - import something from "./file" where file doesn't exist → should error (import-x/no-unresolved)

  React/Preact
  - A component without a return type that returns different types → check react-hooks/exhaustive-deps by using a hook dependency that's missing

  TypeScript
  - @ts-ignore without description → should error (ban-ts-comment)
  - Unused variable → should error (@typescript-eslint/no-unused-vars)
  - let x: string = "hello" where type is inferrable → this one we turned off intentionally

  Catch a regression from the rename
  - import/order — try importing in wrong order (this rule may also be import-x/order now)




### Resolved issues
<!-- List all issues which should be closed when this PR is merged. -->

Fixes: #3539 


<!-- Keep this link for the potential reviewer -->


eslint.config.mjs
```diff
@@ -0,0 +1,236 @@
...
+    {
+        name: "js/config",
+        ...js.configs.recommended,
+    },
+    plugins.stylistic,
+    plugins.importX,
+    ...configs.base.recommended,
+]);
+
+// strip plugin registration from preactConfig to avoid confligt with eslint-config-airbnb-extended
```
seluianova:
```suggestion
// strip plugin registration from preactConfig to avoid conflict with eslint-config-airbnb-extended
```

eslint.config.mjs
```diff
@@ -0,0 +1,236 @@
...
+            "package.json",
+            "package-lock.json",
+            "vitest.config.ts",
+        ],
+    },
+    {
+        languageOptions: {
+            parser: typescriptParser,
+            parserOptions: {
+                ecmaVersion: 6,
```
seluianova:
```suggestion
                ecmaVersion: "latest",
```
optional nit-pick (unless v.6 is intentional)?

> In most cases, we recommend using the default of "latest" to ensure you’re always using the most recent ECMAScript version.
https://eslint.org/docs/latest/use/configure/language-options

eslint.config.mjs
```diff
@@ -0,0 +1,236 @@
...
+                    ignoreArrayIndexes: true,
+                },
+            ],
+            "no-mixed-operators": "error",
+            "no-plusplus": ["error", { allowForLoopAfterthoughts: true }],
+            "prefer-destructuring": ["error", { array: false }],
+            "prefer-arrow/prefer-arrow-functions": "error",
+            "prefer-object-spread": "error",
+            "prefer-template": "error",
+            "react/function-component-definition": ["error", { namedComponents: "arrow-function" }],
```
seluianova, line 156: nit-pick question: do we use any react components?
not sure how to reproduce this check.
I feel like since we are using preact, we don't need this one.
_but I might be missing something, so feel free to ignore_

seluianova (approved):
tested a few rules, looks nice to me 👍

MizukiTemma (approved):
No issue found, linting works as before 😸

eslint.config.mjs
hannaseithe: I will follow your suggestion, but be aware that the ts and the js overrides actually apply in those cases.

eslint.config.mjs
hannaseithe: From the little research I did: There is no alternative preact specific rule we could use instead. Furthermore `eslint-config-preact` depends itself on the react plugin. So it really only extends but doesnt replace them.

PR #4356, opened 2026-06-12
Refactor bulk-actions.ts as feature module

### LINKED ISSUE #4355: [Task]: Refactor `bulk-actions.ts` as feature module
### Task Description

This refactors the file ~~auto-file-download.ts~~ bulk-actions.ts as a feature module. No change in functionality

### Task Category

- Refactor based on feature module pattern
- decouple where necessary

### Acceptance Criteria

- feature module exists and works
- old file and file imports deleted

### Technical Details

_No response_

### Risks & Impact

_No response_

### Target Environment

_No response_

### Verification / Testing

_No response_

### Additional Context

_No response_

_No response_

### Definition of Done

- [x] Work is complete and meets acceptance criteria
- [x] Code reviewed / peer reviewed (if applicable)
- [x] Documentation updated (if applicable)
- [x] Changes deployed to target environment
- [x] Verified in target environment
- [x] Parent issue / stakeholders notified

hannaseithe:
### Short description

After #4009 this PR migrates bulk-actions.ts from a legacy global module to a feature module (js/feature/bulk-actions.ts), replacing per-element event listener attachment with event delegation and a subpages-loaded custom event.

### Proposed changes

- Delete js/bulk-actions.ts and replace with js/feature/bulk-actions.ts using the defineFeature pattern
- Add data-js-bulk-actions wrapper to all list view templates that use bulk actions
- Add meaningful docstring for documentation generation

- Switch from per-checkbox event listeners to event delegation on the root element, so dynamically inserted rows (async page tree loading) are handled without re-initialisation
 - Remove the setBulkActionEventListeners import and call from fetch-subpages.ts; instead dispatch a subpages-loaded event after async loading completes:
  - bulk-actions.ts listens for subpages-loaded (once) to remove cursor-wait from bulk checkboxes
- Fix pre-existing bug: cursor-wait on #bulk-select-all was never removed on the statistics page because setBulkActionEventListeners returned early when #bulk-action was absent; fixed in statistics-page-accesses.ts
- Make cursor-wait on bulk checkboxes conditional in templates: only applied when not filter_form.is_enabled (i.e. only when async subpage loading is active)

### Side effects

- fetch-subpages.ts now dispatches a subpages-loaded event on document after all subpages are loaded — future feature modules can listen for this instead of being called directly

### How to test

  Visit the following pages and verify:
  1. Select-all checkbox selects/deselects all rows
  2. Per-item checkboxes update the selection count
  3. Execute button stays disabled until both items and an action are selected
  4. A bulk action submits correctly; confirmation dialog fires where applicable

  Pages with bulk actions: page tree, events, POIs, contacts, organisations, push notifications, feedback, language tree, broken links.

  Additional page tree checks:
  - Checking a parent propagates checked state to children recursively
  - With active filters (filter_form.is_enabled): no cursor-wait on checkboxes, no async loading

In page-accesses make sure the cursor-wait shows after subpages are loaded.

### Resolved issues

Fixes: #4355


MizukiTemma (changes requested):
Not yet a full review, I'll leave comments when something is found 👁️ 

The bulk actions are not set correctyl in the contact lists. 

`[bootstrapModules] Failed to init bulk-actions for [object HTMLDivElement] TypeError: Cannot read properties of null (reading 'addEventListener')
    at Object.default (bulk-actions.ts:114:16)
    at index.ts:159:69
    at Generator.<anonymous> (style.scss:25:78)
    at Generator.next (style.scss:25:78)
    at style.scss:25:78
    at new Promise (<anonymous>)
    at style.scss:25:78
    at index.ts:158:37
    at index.ts:161:26
    at Generator.<anonymous> (style.scss:25:78)`

MizukiTemma (approved):
Didn't find any further malfunction in the lists affected by this PR, looks good 😸

integreat_cms/static/src/js/feature/bulk-actions.ts
```diff
@@ -0,0 +1,155 @@
...
+        childrenIds.forEach((childId) => setCheckboxRecursively(root, childId, checked));
+    }
+};
+
+export default defineFeature((root) => {
+    console.debug("Set event handlers for bulk actions");
+
+    const selectAllCheckbox = root.querySelector<HTMLInputElement>("#bulk-select-all");
+    const bulkAction = root.querySelector<HTMLSelectElement>("#bulk-action");
+    const form = root instanceof HTMLFormElement ? root : root.querySelector<HTMLFormElement>("#bulk-action-form");
```
jonbulz, line 105: any of these being `null` throws. steps to reproduce:

- log in as observer role
- go to events list view
- see exception in the console (because observer has view_event, but not change_event permission)

would be nice to have a guard like 
```
if (!bulkAction || !selectAllCheckbox || !form) {return;}
```

integreat_cms/static/src/js/feature/bulk-actions.ts
```diff
@@ -0,0 +1,155 @@
...
+    if (!form) {
+        return;
+    }
+    const initialTarget = form.target;
+    const selectedAction = bulkAction.options[bulkAction.selectedIndex];
+    const action = selectedAction.getAttribute("data-bulk-action");
+    if (action) {
+        const url = new URL(action, window.location.origin);
+        if (url.origin === window.location.origin) {
+            form.action = url.pathname;
```
jonbulz: here's one more that claude found: 
thus far, executing a bulk action redirected to the user-selected page in the linkcheck list view. this only holds for linkcheck, every other list view always redirected to the list view without query parameters, i.e. the default first page. 
after these changes, bulk actions in linkcheck also redirect to the default first page, because the query string does not get passed to the POST handler anymore. to fix this, `url.pathname` would need to become `url.pathname + url.search`

I find this a bit tricky, because 
a) it is a regression for the linkcheck list view bulk actions. I can see it confusing users that are used to staying on the same page after triggering a bulk action, especially since the linkcheck list view tends to have a lot of pages where it can be hard to find and track individual entries

but...
b) it makes the linkcheck list view bulk actions consistend with every other list view, because for all of them we redirect to the default first page

IDK, maybe @m0ndra has an opinion?

jonbulz (changes requested):
Thanks, tested it locally and it works great in general! I only have a few concerns, see my above comments.
I also want to note that, while we have the `cursor-wait` as indication that not all sub-pages are loaded yet, toggling select all on a large page tree will not actually select everything until all sub-pages are fetched. There is nothing that prevents users from ignoring the cursor-wait and using select all when it is not yet ready. Maybe we should disable it until all sub-pages are loaded?

integreat_cms/static/src/js/feature/bulk-actions.ts
hannaseithe: Good catch

integreat_cms/static/src/js/feature/bulk-actions.ts
hannaseithe: Actually this was not an attempt from my side to make linkcheck consistent with the rest of bulk actions, but to only somehow satisfy the "github-advanced-security" bot (see earlier commits). I actually do agree we should keep `url.search` . I will have to experiment and see how to not both cause a regression (as you described for the linkchecker) as well as make the bot happy.

hannaseithe: Ok, supressing the alert seems to have worked. Since we do the url parsing and origin check, url.search should not be significantly taintable anymore. Or you ok, with this @jonbulz ?

And @mondra in case you feel we SHOULD actually align linkcheck behavior with the rest of the list views, I would recommend tackling this in a different issue

m0ndra: From my side it's fine as it is right now, I won't open another issue but will document that we have a slight inconsistency and have a look at it hollistically in the future.

jonbulz: Yes, sounds good :)

jonbulz (approved):
Thanks for applying my suggestions! Once the merge conflicts are resolved, this is good to go IMO 👍

MizukiTemma:
I added the old version of `updateSelectionCount` and `setCheckboxRecursively` into `statistics-page-accesses.ts` as their new versions in `/feature/bulk-actions` are not compatible with `statistics-page-accesses.ts`. #4430 This change came afterwards as #3394 was merged.

PR #4368, opened 2026-06-18
Bugfix: Add region scope to contact and raw contact endpoint

hannaseithe:
### Short description

This fixes the endpoints `get_contact` and `get_contact_raw` in contact_utils.py. 

### Proposed changes

- add region scope to endpoints
- tests

### Side effects

- None

### Faithfulness to issue description and design

There are no intended deviations from the issue and design.

### How to test

- follow the steps for reproduction in original issue

### Resolved issues

Fixes: #4367 

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

Issue #4289, opened 2026-05-13
[Task]: Shortcodes in TinyMCE


PeterNerlich:
*Until the feature is ready to be released, any work on it is to be done in / merged into [`feature/shortcodes-tinymce`](https://github.com/digitalfabrik/integreat-cms/tree/feature/shortcodes-tinymce)*

*The decision has been made to test shortcodes itself without special handling in javascript, by delivering the rendered HTML to TinyMCE and detecting any internal links in submitted content and substituting them with the equivalent shortcode. This issue exists only for reference / in case this is picked up again later, after Shortcodes themselves have proved valuable.*

### What
Adjust TinyMCE to
- insert shortcodes instead of raw html whereever possible and
- render shortcodes in some way (but not overwrite them in the content)

This is a major new feature, consisting of
- [x] ground work to make implementing UI for individual shortcodes possible and managable
- [ ] cleanup of that ground work
- [x] implementation of the UI for the `page` shortcode (replacing direct links to internal pages)
- [ ] implementation of the UI for `contact` shortcode (replacing direct insertion of contact card HTML)

### Why
Our users are not too technical and should be able to create and edit links as before, if not better.

### Task Category

🏗️ Implementation — Sub-task of a story or feature

### Steps / Implementation Plan

- [x] draft system
- [ ] remaining implementation of `page` shortcode
- [ ] remaining implementation of `contact` shortcode
- [ ] clean up system

### Acceptance Criteria

- [ ] the `page` shortcode is working as well as the mechanism it replaces does
- [ ] the `contact` shortcode is working as well as the mechanism it replaces does
- [ ] there are no major known bugs in the shortcodes UI framework

### Affected Systems / Services

- TinyMCE editor
- saved content
- API

### Risks

Risks are minimized, as shortcodes are going to be polished in their feature branch until ready

### Verification / Testing

- [ ] the `page` shortcode is working as well as the mechanism it replaces does
- [ ] the `contact` shortcode is working as well as the mechanism it replaces does
- [ ] there are no major known bugs

### Parent / Related issue and Work

- #3745

### Definition of Done

- [ ] Work is complete and meets acceptance criteria
- [ ] Code reviewed / peer reviewed (if applicable)
- [ ] Tests added or updated (if applicable)
- [ ] Documentation updated (if applicable)
- [ ] Changes deployed to target environment
- [ ] Verified in target environment
- [ ] Parent issue / stakeholders notified

PR #4371, opened 2026-06-18
Bugfix: Fix region scope for render_contact_card

### LINKED ISSUE #4370: Cross-region IDOR: `render_contact_card` embeds contact PII from other regions into public content
### Bug Summary

`render_contact_card` in `content_utils.py` retrieves a contact by primary key without scoping to the current region. When an editor saves content containing a contact card div, the contact is rendered and stored at save time. If the contact belongs to a different region, its PII gets permanently stored in the content and exposed via the public API.

### Expected Behavior

Only contacts belonging to the current region should be rendered into content. References to contacts from other regions should be ignored or result in an error placeholder.

### Actual Behavior

Any contact ID can be embedded in content regardless of region, causing PII (name, email, phone, address, etc.) from other regions to be stored and served publicly.

### Steps to Reproduce

1. Note the ID of a contact belonging to **Region A**
2. As an editor in **Region B**, open a page for editing and insert the following in the TinyMCE source editor:
   ```html
   <div data-contact-id="<id>" data-contact-url="/region-a/contact/<id>/?details=name,email"></div>
   ```
3. Save/publish the page
4. Observe that the saved content now contains Region A's contact PII
5. Request `GET /api/v3/region-b/pages/` and observe the PII in the response

### Reproducibility

Always (100%)

### Environment

- **User role / permissions:** Any editor-level role with content editing permissions

### Severity

### Technical Context

**Affected Components:** `integreat_cms/cms/utils/content_utils.py` lines 162 and 190

**Relevant Endpoints (where the stored PII is exposed):**
- `GET /api/v3/<region>/pages/`
- `GET /api/v3/<region>/locations/`
- `GET /api/v3/<region>/events/`
- `GET /api/v3/<region>/imprint/`

**Suspected Root Cause:** Missing `location__region` filter in `render_contact_card`. The function has no access to the current region, so the region will need to be passed in as a parameter through `update_contacts` → `clean_content`.

**Relevant Code:**
```python
# current (vulnerable) — content_utils.py:162
"contact": Contact.objects.get(pk=contact_id),

# fix (region needs to be threaded through from clean_content)
"contact": Contact.objects.get(pk=contact_id, location__region=region),
```

**Related:** Same missing `location__region` filter as in `get_contact` / `get_contact_raw`, but in the public content path.

- [ ] Contact cards for contacts from other regions are not rendered into stored content
- [ ] Contact cards for contacts belonging to the current region still render correctly
- [ ] Existing public API responses do not expose cross-region contact PII

### Verification / Testing

- [ ] Reproduce original bug → confirm cross-region contact PII is no longer stored in content
- [ ] Existing contact card tests still pass

hannaseithe:
### Short description

This fixes the security issue (IDOR) bug that it is possible for an editor from Region A to include contact cards from Region B into their pages.

### Proposed changes

- region scope the function `render_contact_card`
- tests

### Side effects

- None

### How to test

- follow the steps for reproduction on the original bug issue

### Resolved issues

Fixes: #4370

### DIFF
--- a/integreat_cms/cms/utils/content_utils.py
+++ b/integreat_cms/cms/utils/content_utils.py
@@ -20,7 +20,9 @@ logger = logging.getLogger(__name__)
 _xss_cleaner = Cleaner(scripts=True, javascript=True, safe_attrs_only=False)
 
 
-def clean_content(content: str, language_slug: str) -> str:
+def clean_content(
+    content: str, language_slug: str, region_id: int | None = None
+) -> str:
     """
     This is the super function to clean content
 
@@ -40,7 +42,7 @@ def clean_content(content: str, language_slug: str) -> str:
     fix_alt_texts(content)
     fix_notranslate(content)
     hide_anchor_tag_around_image(content)
-    update_contacts(content)
+    update_contacts(content, region_id)
 
     content_str = tostring(content, encoding="unicode", with_tail=False)
     return fix_content_link_encoding(content_str)
@@ -149,7 +151,9 @@ def update_internal_links(link: HtmlElement, language_slug: str) -> None:
 
 
 def render_contact_card(
-    contact_id: int | str | None, wanted_details: Iterable[str]
+    contact_id: int | str | None,
+    wanted_details: Iterable[str],
+    region_id: int | None = None,
 ) -> HtmlElement:
     """
     Produces a rendered html element for the contact.
@@ -162,19 +166,32 @@ def render_contact_card(
     """
     template = loader.get_template("contacts/contact_card.html")
     try:
+        if region_id is None:
+            contact = Contact.objects.get(pk=contact_id)
+        else:
+            contact = Contact.objects.get(pk=contact_id, location__region_id=region_id)
+
         context = {
-            "contact": Contact.objects.get(pk=contact_id),
+            "contact": contact,
             "wanted": wanted_details,
         }
     except Contact.DoesNotExist:
-        logger.warning("Contact with id=%r does not exist!", contact_id)
-        # Provide a "dummy contact"
-        # This is super hacky and barely renders a card,
-        # but makes it obvious to the user that there should be a contact here, but there was an error
+        if region_id is not None and Contact.objects.filter(pk=contact_id).exists():
+            message = _(
+                "This contact belongs to a different region and cannot be displayed."
+            )
+            logger.warning(
+                "Contact with id=%r belongs to a different region",
+                contact_id,
+            )
+        else:
+            message = _("Oops! Error displaying contact data")
+            logger.warning("Contact with id=%r does not exist!", contact_id)
         context = {
             "contact": {
-                "name": _("Oops! Error displaying contact data"),
+                "name": message,
                 "absolute_url": f"/{None}/contact/{contact_id}/",
+                "pk": contact_id,
             },
             "wanted": ["name"],
         }
@@ -191,7 +208,7 @@ def render_contact_card(
         return Element("pre", raw_element)
 
 
-def update_contacts(content: HtmlElement) -> None:
+def update_contacts(content: HtmlElement, region_id: int | None = None) -> None:
     """
     Inject rendered contact html for given ID
 
@@ -210,7 +227,7 @@ def update_contacts(content: HtmlElement) -> None:
             wanted_details = []
 
         contact_card_new = (
-            render_contact_card(contact_id, wanted_details)
+            render_contact_card(contact_id, wanted_details, region_id)
             if any(detail for detail in wanted_details)
             else fromstring("<div><p></p></div>")
         )
--- a/tests/cms/utils/test_content_utils.py
+++ b/tests/cms/utils/test_content_utils.py
@@ -1,7 +1,9 @@
 import pytest
 from django.test.client import Client
+from django.utils import translation
+from lxml.html import tostring
 
-from integreat_cms.cms.utils.content_utils import clean_content
+from integreat_cms.cms.utils.content_utils import clean_content, render_contact_card
 from tests.conftest import EDITOR, MANAGEMENT, PRIV_STAFF_ROLES
 
 
@@ -16,7 +18,7 @@ def test_clean_content(
     login_role_user: tuple[Client, str],
 ) -> None:
     raw_content = '<h1>Das ist eine H1</h1><pre>Das ist vordefinierter Text</pre><code>Das ist vordefinierter Code</code><a href="https://www.integreat-app.de"></a><a href="http://localhost:8000/augsburg/pages/de/5" class="link-external"></a>'
-    cleaned_content = clean_content(raw_content, "de")
+    cleaned_content = clean_content(raw_content, "de", 1)
 
     # Test convert_heading works
     assert "<h1>Das ist eine H1</h1>" not in cleaned_content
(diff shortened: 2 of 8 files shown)


MizukiTemma:
@hannaseithe 
Isn't it already covered by #4368 ? #4368 prevents users from embedding a contact that does not belong to their region 🤔

hannaseithe:
> @hannaseithe Isn't it already covered by #4368 ? #4368 prevents users from embedding a contact that does not belong to their region 🤔

#4368 ~~blocks it from being embedded as a contact card in TinyMCE through url manipulation~~ (edit: I dont think it has influence on that at all), but it does not stop you from going into the source-code editor and adding a `<div data-contact-id="<id>" data-contact-url="/region-a/contact/<id>/?details=name,email"></div>`  tag. After publishing, the contact from the other region will be added

edit2: Just for a better explanation: TinyMCE already makes it impossible for users in the normal flow to add a other-region contact inside the dialog: The search is region scoped (getCompletions > search_contact_ajax). GetContactHtml (> get_contact) - which is called on dialog submit - is not region scoped, but it would have only been handed a selection from the search results.
Also during editing a contact - search again is region scoped.

When adding a contact `<div>` tag through the source code directly, getContact nor getContactRaw are touched, but the content in clean_content() goes through a chain of procedures before saving of which one is update_contacts() > render_contact_cards() , where now with the region scope, a contact from a different region simply will not be loaded nor rendered - instead you get "Oops! Error displaying contact data"

hannaseithe:
> When adding a contact `<div>` tag through the source code directly, getContact nor getContactRaw are touched, but the content in clean_content() goes through a chain of procedures before saving of which one is update_contacts() > render_contact_cards() , where now with the region scope, a contact from a different region simply will not be loaded nor rendered - instead you get "Oops! Error displaying contact data"

This is actually bad UX - if someone added a contact-`<div>` from a different region it should simply be silently removed. I will add those changes.

MizukiTemma:
Thank you for explanation 😸 The region scope works 👍 

I'm not sure whether it's the best to remove cross-region contact cards silently 🤔 A possible senario is region cloning.

There is a region A. It has a contact C that belongs to it.
Region A has a page P, in which C is embedded.
A new region B is created with region A as base. All pages of A are cloned into B.
Now B has the page P', which is the clone of P. That means B has a page with a cross-region contact card.

When the page P' is updated, the embedded contact card disappears silently.


This is probably a question to @laravoelk @osmers @m0ndra

laravoelk:
Is it possible to include a short text with a hint that there used to be a contact from another region here instead of removing it silently? Generally I do think, that this scenario does not happen a lot, but it could of course happen from time to time.

osmers:
@laravoelk do we work with contact cards in Deutschlandvorlage at all? Would it make sense to work with them? I'm thinking whether it always makes sense to remove contacts from other regions or whether there are cases, where we'd want them to be copied into contacts and left in the page. But it probably only applies when copying the Deutschlandvorlage and there only if we use contacts for general information.

But it out of scope for this PR anyways. 

I like Laras suggestion with the hint - but I'm not sure it feasible or when to show it.

hannaseithe:
> Is it possible to include a short text with a hint that there used to be a contact from another region here instead of removing it silently? Generally I do think, that this scenario does not happen a lot, but it could of course happen from time to time.

I could just switch back to the already implemented UX of showing a pseudo-contact card with the message "Oops! Error displaying contact data"? 
Or a different message: "Contacts from other regions cannot be displayed"?

MizukiTemma:
> I could just switch back to the already implemented UX of showing a pseudo-contact card with the message "Oops! Error displaying contact data"? 

I'm for that :)

> Or a different message: "Contacts from other regions cannot be displayed"?

🤔 The sentence can be changed if someone wishes.

@laravoelk @osmers 
FYI: It looks like this
image

hannaseithe:
> @laravoelk @osmers FYI: It looks like this image



Just want to add to this: The editor will then have to delete this Pseudo-Contact Card - so that it wont be published in the page and shown to the user

m0ndra:
For the time being, I'd like to show the user a hint which displays WHAT the problem is and HOW it can be fixed, so something like:

> Diese Kontaktdaten können nicht mehr angezeigt werden. Integreat Nutzende kommen hier nicht weiter. Erstellen Sie einen neuen, passenden Kontakt und verlinken Sie diesen hier, oder löschen Sie diese Kontaktkarte.

Not ideal, since cloning a region is meant to reduce workload and you have to look at every single page to find contact cards which aren't working. But I suppose you look at all pages anyway since no region is 100% the same in terms of content, but it might still be error prone.

hannaseithe:
Implemented those changes @MizukiTemma

osmers:
> Not ideal, since cloning a region is meant to reduce workload and you have to look at every single page to find contact cards which aren't working. But I suppose you look at all pages anyway since no region is 100% the same in terms of content, but it might still be error prone.

We used to have place holders for addresses before as well or even kept the contacts from a cloned region, because we didn't use contact cards yet - so the problem of having to find all of them is not new :)

laravoelk:
As we do not work with contact cards in the Deutschlandvorlage so far, it would not happen a lot (yet). However, if people start with an Integreat-App from another municipality, it would be helpful to have the error message. I like the suggested text where users immediatly see what they need to do now.

tests/cms/utils/test_content_utils.py
```diff
@@ -37,3 +38,34 @@ def test_clean_content(
...
+@pytest.mark.django_db
+def test_clean_content_removes_cross_region_contact(load_test_data: None) -> None:
+    """
+    Test that clean_content silently removes a contact card from a different region
+    """
+    raw_content = '<div><div data-contact-id="3" data-contact-url="/augsburg/contact/3/?details=name"></div></div>'
+    # Region 8 (berlin) — contact 3 belongs to augsburg
+    result = clean_content(raw_content, "de", region_id=8)
+    assert "data-contact-id" not in result
+    assert "Mariana Musterfrau" not in result
```
MizukiTemma:
```suggestion
    
    message = "Dieser Kontakt geh&#246;rt zu einer anderen Region und kann nicht angezeigt werden. Erstellen Sie einen neuen, passenden Kontakt und verlinken Sie diesen hier. Oder l&#246;schen Sie diese Kontaktkarte."
    assert message in tostring(render_contact_card(3, ["name"], region_id=8)).decode("utf-8")

```

MizukiTemma (approved):
Thank you, I already approve the PR :)

tests/cms/utils/test_content_utils.py
hannaseithe: Thank you very much for catching that.

integreat_cms/cms/utils/content_utils.py
```diff
@@ -158,19 +162,32 @@ def render_contact_card(
...
+                "Contact with id=%r belongs to a different region",
+                contact_id,
+            )
+        else:
+            message = _("Oops! Error displaying contact data")
+            logger.warning("Contact with id=%r does not exist!", contact_id)
         context = {
             "contact": {
-                "name": _("Oops! Error displaying contact data"),
+                "name": message,
```
PeterNerlich, line 192: I appreciate the helpful message to the editor, but I am not sure how I feel about rendering it inside the contact card that is going to be displayed to normal users, too.

Maybe @m0ndra and @osmers have some input on this?

PeterNerlich (approved):
Looks good! Thank you for this change

integreat_cms/cms/utils/content_utils.py
MizukiTemma: @PeterNerlich 
Thank you for review. I took over this PR.

For the main case where a contact card is cross-region embedded, users see this messgae. "Oops!..." is for all other cases, which are very very rare in my opinion.

So I would say the handling is enough. Cross-region embedding itsself shuold not happen actually often (for users using the CMS in an ordinary way).

image


Could you re-review?

m0ndra: > I appreciate the helpful message to the editor, but I am not sure how I feel about rendering it inside the contact card that is going to be displayed to normal users, too.

@PeterNerlich you are totally right, the error message would be a bit weird for app users, since I wrote it for cms user 🙈 is there a possibility to show the cms users something else than the app users? otherwise we'd have to remove the helpful how to fix this hint and just go with "Dieser Kontakt gehöt zu einer anderen Region und kann nicht angezeigt werden."

MizukiTemma: @PeterNerlich 
🙈 sorry, I misunderstood your feedback 🙈 

@m0ndra 
Thank you for making the point clear 🙏 
Changing the message depending on App or CMS, is not so trivial, and I do not recommend to spend much effort for that, as cross-region contact cards are definetely a very very edge case 🤔 

So, I'm for the general message "Dieser Kontakt gehöt zu einer anderen Region und kann nicht angezeigt werden." :) The sencond sentence "Erstellen Sie ... " can be added to the Wiki ✏️

MizukiTemma: Updated ✏️ 


image

PeterNerlich: > is there a possibility to show the cms users something else than the app users?

Adding this capability was envisioned as part of #4289, but without being tied to shortcodes *in the editor* I don't feel like this is a reasonable investment of effort and complexity. If we are happy with Shortcodes in half a year decide we want to do the Browser/Editor side handling after all, this would be trivial.

PeterNerlich (approved):
Thanks!

PR #4372, opened 2026-06-18
Add an endpoint for single news post

### LINKED ISSUE #4339: [Task]: Follow-up Common News Endpoint 1
### Task Description

A new endpoint was added by #4195 to support multiple source for our news feature.

News posts imported from external sources (Currently only Tü News but Amal News will follow) are saved as cache in the CMS with the `id` given to them in the original source. This means it's possible that more than one news have the same `id` if they come from different news sources (one push news post with id=42, one Tü News post with id=42 and one Amal News post with id=42 can co-exist in the system in the current implementation).

As an element endpoint must be provied now (#4340 ), the `id` of news posts must be inter-source unique. 

### Task Category

TBD

### Acceptance Criteria

- [ ] Every news has an unique `id`, regardless of source

### Technical Details

_No response_

### Risks & Impact

_No response_

### Target Environment

_No response_

### Verification / Testing

_No response_

### Additional Context

_No response_

_No response_

### Definition of Done

### LINKED ISSUE #4340: [Task]: Follow-up Common News Endpoint 2
### Task Description

A new endpoint was added by #4195 to support multiple source for our news feature.
This topic needs however some upgrades.

- Add an element endpoint `/api/v3/<region_slug>/<language_slug>/news/<id>` which identifies a news by the given id and returns it
- Update the API documentation accordingly with the changes so far made for news

### Task Category

- Wait for #4339
- Implement an element entpoint as specified anove
- Update the API documentation to include changes of this issue and #4195 

### Acceptance Criteria

- [ ] Endpoint `/api/v3/<region_slug>/<language_slug>/news/<id>` is available and functioning
- [ ] API documentation contains all the changes so far made

### Technical Details

_No response_

### Risks & Impact

_No response_

### Target Environment

_No response_

### Verification / Testing

_No response_

### Additional Context

_No response_

_No response_

### Definition of Done

MizukiTemma:
### Short description

This PR makes `id` of news items unique and adds a new entpoint which return a single news object.

### Proposed changes

- Add a prefix `tuNews-` for news posts from Tü News at import (`name` of each news manager as prefix)
- Return push news with prefix `local-`
- Add a new entpoint `news/<news_id>` which return exact one news item that matches the id 
- Add a test for the new endpoint
- Update API documentation

**updates**
- Two separate transform functions `transform_notification_for_fcm` and `transform_notification` not to break the existing `fcm/` endpoint

### Side effects

@steffenkleinle 
- Is it okay to add a prefix `local-` to push news? Does it break anything on the app?
The common endpoint `news` will return push news with `"id": "local-123"` instead of `"id": 123` and `news/<news_id>/` endpoit is expected to ask for a push news like `news/local-123/`

I splitted unique id and the endpoint into two separate issues but noticed it makes sense to implement both together to check it actually works 😅 

### How to test

- Run the command `import_external_news` to see prefix `tuNews-` is added
- Hit the new endpoint `news/<news_id>` to see the correct news item is returned

### Resolved issues

Fixes: #4339 
Fixes: #4340 

### DIFF
--- a/docs/src/api-docs.rst
+++ b/docs/src/api-docs.rst
@@ -705,6 +705,78 @@ RESPONSE
    }
 
 
+News
+====
+
+Get all news posts from all available sources
+
+REQUEST
+~~~~~~~
+
+.. code:: http
+
+   GET /api/v3/{region_slug}/{language_slug}/news/ HTTP/2
+
+Response
+~~~~~~~~
+
+.. code:: javascript
+
+   [
+      {
+         "id": String,                          // The id of the news
+         "title": String,                       // The title of the news
+         "content": String,                     // The content of the news
+         "last_updated": String,                // Date of last update
+         "display_date": String,                // Date of publication or last update, of which later
+         "channel": String | null,              // The channel of the push notification, null for external news
+         "available_languages": [               // The available languages of the push notification
+               "<language_slug>": {
+                  "id": String | null,          // The id of the translation
+               },
+               ...
+            ],
+         "source": String,                      // The source of the news ('local', 'tuNews' or 'amalNews')
+         "externalUrl": String,                 // The link to the news
+      }
+   ]
+
+
+Single News
+===========
+
+Get a single news
+
+REQUEST
+~~~~~~~
+
+.. code:: http
+
+   GET /api/v3/{region_slug}/{language_slug}/news/{news_id}/ HTTP/2
+
+Response
+~~~~~~~~
+
+.. code:: javascript
+
+   {
+      "id": String,                          // The id of the news
+      "title": String,                       // The title of the news
+      "content": String,                     // The content of the news
+      "last_updated": String,                // Date of last update
+      "display_date": String,                // Date of publication or last update, of which later
+      "channel": String | null,              // The channel of the push notification, null for external news
+      "available_languages": [           // The available languages of the push notification
+            "<language_slug>": {
+               "id": String | null,           // The id of the translation
+            },
+            ...
+         ],
+      "source": String,                      // The source of the news ('local', 'tuNews' or 'amalNews')
+      "externalUrl": String,                 // The link to the news
+   }
+
+
 Imprint / Disclaimer
 ====================
 
--- a/integreat_cms/api/v3/news.py
+++ b/integreat_cms/api/v3/news.py
@@ -9,7 +9,7 @@ import logging
 from typing import TYPE_CHECKING
 
 from django.core.paginator import Paginator
-from django.http import JsonResponse
+from django.http import Http404, JsonResponse
 
 from integreat_cms.cms.views.mixins import get_safe_page
 
@@ -37,7 +37,9 @@ def sent_push_notifications(
     :return: JSON object according to APIv3 push notifications definition
     """
     channel = request.GET.get("channel", "all")
-    result = registry.PUSHNEWS.collect_news_items(region_slug, language_slug, channel)
+    result = registry.PUSHNEWS.collect_news_items_for_fcm(
+        region_slug, language_slug, channel
+    )
     return JsonResponse(result, safe=False)
 
 
@@ -82,3 +84,42 @@ def news(
     page = get_safe_page(paginator, page_num)
 
     return JsonResponse(list(page.object_list), safe=False)
+
+
+@json_response
+def single_news(
+    request: HttpRequest,
+    region_slug: str,
+    language_slug: str,
+    news_id: str,
+) -> JsonResponse:
+    """
+    Function to collect and return an news item which matched the given news id.
+
+    :param request: Django request
+    :param region_slug: slug of a region
+    :param language_slug: language slug
+    :param news_id: id of the requested news item
+    :return: JSON single NewsItem
+    """
+    region = request.region
+    language = region.get_language_or_404(language_slug, only_active=True)
+
+    parts = news_id.rsplit("-", 1)
+    if len(parts) != 2 or not parts[1].isdigit():
+        raise Http404("Invalid news id.")
+    news_type, raw_id = parts
+
+    news_manager = next(
+        (
+            news_manager
+            for news_manager in registry.CHOICES
+            if news_manager.name == news_type
+        ),
+        None,
+    )
+
+    if not news_manager:
+        raise Http404("No matching news source was found.")
+
+    return news_manager.get_single_news(request, region, language, raw_id)
--- a/integreat_cms/news_managers/abstract_news_manager.py
+++ b/integreat_cms/news_managers/abstract_news_manager.py
@@ -6,22 +6,21 @@ from typing import TYPE_CHECKING, TypedDict
 if TYPE_CHECKING:
     from datetime import datetime
 
-    from django.http import HttpRequest, HttpResponse
+    from django.http import HttpRequest, HttpResponse, JsonResponse
 
     from ..cms.models import Language, Region
 
 
 class NewsItem(TypedDict):
-    id: int
+    id: str
     title: str
-    message: str
-    timestamp: datetime
+    content: str
     last_updated: datetime
     display_date: datetime
     channel: str | None
     available_languages: dict | None
     source: str
-    link: str | None
+    externalUrl: str
 
 
 class AbstractNewsManager(ABC):
@@ -53,10 +52,23 @@ class AbstractNewsManager(ABC):
         request: HttpRequest,
         region: Region,
         language: Language,
-        slug: str,
+        news_raw_id: str,
     ) -> HttpResponse:
         """
         Tries rendering the social media headers for a news page in a specified region and language
         To be implemented in the inheriting model
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_single_news(
+        self,
+        request: HttpRequest,
+        region: Region,
+        language: Language,
+        news_id: str,
+    ) -> JsonResponse:
+        """
+        Returns a news item that is imported from the source and matches the id
         """
         raise NotImplementedError
--- a/integreat_cms/news_managers/pushnews_manager.py
+++ b/integreat_cms/news_managers/pushnews_manager.py
@@ -5,7 +5,7 @@ from typing import TYPE_CHECKING
 from django.conf import settings
 from django.db.models import F
 from django.db.models.functions import Greatest
-from django.http import Http404
+from django.http import Http404, JsonResponse
 from django.utils import timezone
 
 from ..cms.models import PushNotificationTranslation
@@ -17,35 +17,12 @@ from ..cms.utils.social_media_utils import (
 from .abstract_news_manager import AbstractNewsManager, NewsItem
 
 if TYPE_CHECKING:
+    from typing import Any
+
     from django.db.models.query import QuerySet
     from django.http import HttpRequest, HttpResponse
 
     from ..cms.models import Language, Region
-
-
-def _transform_notification(pnt: PushNotificationTranslation) -> NewsItem:
-    """
-    Function to create a JSON from a single push notification translation Object.
-
-    :param pnt: A push notification translation
-    :return: data necessary for API
-    """
-    available_languages_dict = {
-        translation.language.slug: {"id": translation.id}
-        for translation in pnt.push_notification.translations.all()
-    }
-    return {
-        "id": pnt.pk,
-        "title": pnt.get_title(),
-        "message": pnt.get_text(),
-        "timestamp": pnt.last_updated,  # deprecated field in the future
-        "last_updated": timezone.localtime(pnt.last_updated),
-        "display_date": pnt.display_date,
-        "channel": pnt.push_notification.channel,
-        "available_languages": available_languages_dict,
-        "source": "local",
-        "link": None,
-    }
 
 
 def _query_sent_translations(
@@ -86,17 +63,17 @@ class PushnewsManager(AbstractNewsManager):
         self, region_slug: str, language_slug: str, channel: str
     ) -> list[NewsItem]:
         """
-        Returns push notification news
+        Returns push notification news for common news endpoint
         """
         query_result = _query_sent_translations(region_slug, language_slug, channel)
-        return list(map(_transform_notification, query_result))
+        return list(map(self.transform_notification, query_result))
 
     def social_media_headers(
         self,
         request: HttpRequest,
         region: Region,
         language: Language,
-        slug: str,
+        news_raw_id: str,
     ) -> HttpResponse:
         """
         Tries rendering the social media headers for a news page in a specified region and language
@@ -104,7 +81,7 @@ class PushnewsManager(AbstractNewsManager):
         if not (
             pn_translation := PushNotificationTranslation.objects.filter(
                 language__slug=language.slug,
-                push_notification__id=slug,
+                push_notification__id=news_raw_id,
                 push_notification__regions=region,
             ).first()
         ):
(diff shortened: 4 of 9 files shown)


steffenkleinle:
> Is it okay to add a prefix local- to push news? Does it break anything on the app?
The common endpoint news will return push news with "id": "local-123" instead of "id": 123 and news/<news_id>/ endpoit is expected to ask for a push news like news/local-123/

I am really not sure about the naming here what you mean by push notifications? Do you mean
1. the actual push notification sent via firebase?
2. or the local news in the new messages endpoint that are first sent as push notifications?

If you mean 1.: Changing the id in the actual push notifications in firebase would be a breaking change and should definitely be avoided!

If you mean 2.: We'd somehow need the new news id in the content of push notifications as well to correctly link to the corresponding local news. However, it would be possible to just add the prefix on the app side though for push notifications.

MizukiTemma:
> 
> 1. the actual push notification sent via firebase?
> 2. or the local news in the new messages endpoint that are first sent as push notifications?
> 
> If you mean 1.: Changing the id in the actual push notifications in firebase would be a breaking change and should definitely be avoided!
> 



@steffenkleinle 
Sorry for confusion 🙈 

> 1. the actual push notification sent via firebase?
> If you mean 1.: Changing the id in the actual push notifications in firebase would be a breaking change and should definitely be avoided!

No, I don't mean this. Push notification objects and push notification translation objects keep their (auto-provided) `int` ids in the databse :)

 > 2. or the local news in the new messages endpoint that are first sent as push notifications?
 > If you mean 2.: We'd somehow need the new news id in the content of push notifications as well to correctly link to the corresponding local news. However, it would be possible to just add the prefix on the app side though for push notifications.
 
 Yes, it's about local news 📰 
 The current implementation of this PR delivers local news always with the prefix. If it changes during review I'll let you know.

integreat_cms/news_managers/pushnews_manager.py
```diff
@@ -119,3 +94,56 @@ def social_media_headers(
...
+            )
+            .first()
+        ):
+            raise Http404(
+                "Push Notification not found in this region with this language."
+            )
+
+        return JsonResponse(self.transform_notification(pn_translation), safe=False)
+
+    def transform_notification(self, pnt: PushNotificationTranslation) -> NewsItem:
```
hannaseithe, line 143: I would create this as an abstract method on `abstract_news_manager.py` as well

integreat_cms/api/v3/news.py
```diff
@@ -82,3 +82,38 @@ def news(
...
+    :param request: Django request
+    :param region_slug: slug of a region
+    :param language_slug: language slug
+    :param news_id: id of the requested news item
+    :return: JSON list of news items merged across all sources defined in :data:`~integreat_cms.news_managers.registry.CHOICES`
+    """
+    region = request.region
+    language = region.get_language_or_404(language_slug, only_active=True)
+
+    news_type, news_id = tuple(news_id.split("-"))
```
hannaseithe: I think we should first validate that news_id contains  ONE  "-"  between other characters
+ that the int part is an int
```suggestion
    parts = news_id.split("-", 1)
  if len(parts) != 2 or not parts[1].isdigit():
      raise Http404("Invalid news id.")
  news_type, raw_id = parts
```

integreat_cms/api/v3/news.py
```diff
@@ -82,3 +82,38 @@ def news(
     page = get_safe_page(paginator, page_num)
 
     return JsonResponse(list(page.object_list), safe=False)
+
+
+@json_response
+def single_news(
```
hannaseithe, line 90: I feel that naming the endpoint view function the same as the   manager method is not the best choice, this can create confusion.

integreat_cms/news_managers/tunews_manager.py
```diff
@@ -152,17 +133,63 @@ def social_media_headers(
         """
         Tries rendering the social media headers for a news page in a specified region and language
         """
-        if not region.external_news_enabled:
-            raise Http404("External news are not enabled in this region.")
-        posts = cache.get(f"tunews:{language.slug}", [])
-        post = next((post for post in posts if str(post["id"]) == slug), None)
-        if not post:
-            raise Http404("Tü news post not found in this region with this news ID.")
+        post = self.find_post(region, language, slug)
```
hannaseithe:
```suggestion
        post = self.find_post(region, language, id)
```

I feel slug should be named `id`. There is no slug field/attribute

tests/api/test_news_endpoint.py
```diff
@@ -135,7 +137,7 @@ def test_news_endpoint_pagination(load_test_data: None, clean_news_cache: None)
     page2_ids = [item["id"] for item in result_page2]
     assert ids[2] in page2_ids
     assert ids[0] not in page2_ids
-    assert ids[1] not in page2_ids
+    assert [1] not in page2_ids
```
hannaseithe: I think this is a typo

integreat_cms/api/v3/news.py
```diff
@@ -82,3 +82,38 @@ def news(
...
+    news_id: str,
+) -> JsonResponse:
+    """
+    Function to collect and return an news item which matched the given news id.
+
+    :param request: Django request
+    :param region_slug: slug of a region
+    :param language_slug: language slug
+    :param news_id: id of the requested news item
+    :return: JSON list of news items merged across all sources defined in :data:`~integreat_cms.news_managers.registry.CHOICES`
```
hannaseithe:
```suggestion
    :return: JSON single NewsItem
```

docs/src/api-docs.rst
```diff
@@ -705,6 +705,70 @@ RESPONSE
...
+   [
+      {
+         "id": String,                          // The id of the news
+         "title": String,                       // The title of the news
+         "message": String,                     // The content of the news
+         "timestamp": String,                   // Date of last update, will be deprecated
+         "last_updated": String,                // Date of last update
+         "display_date": String,                // Date of publication or last update, of which later
+         "channel": String | null,              // The channel of the push notification, null for external news
+         "available_languages": String |null,   // Languages in which a translation of the news is available
```
hannaseithe: returns a dict

integreat_cms/news_managers/tunews_manager.py
```diff
@@ -152,17 +133,63 @@ def social_media_headers(
...
 
         return render_social_media_headers(
             request=request,
             title=post["title"],
             language_code=language.bcp47_tag,
             excerpt=post["message"],
-            url=post["link"],
+            url=post["link"],  # type: ignore[arg-type]
```
hannaseithe: I'd rather change the call signature of render_social_media_headers to url: str | None

hannaseithe (changes requested):
Thank you very much for working on this important new feature,  I have a few change requests.

integreat_cms/api/v3/news.py
MizukiTemma: Nice idea 😸 Applied

integreat_cms/api/v3/news.py
MizukiTemma: The manager method is renamed to `get_single_news` ✏️

tests/api/test_news_endpoint.py
MizukiTemma: 🙈

integreat_cms/news_managers/pushnews_manager.py
MizukiTemma: hmm 🤔 Not a bad idea but I'll leave it as is, becuse it is not a function that is supposed to be callsed for all news managers like `import_news_items` and others.

docs/src/api-docs.rst
MizukiTemma: You mean "available_languages"?

integreat_cms/news_managers/tunews_manager.py
MizukiTemma: I have rather changed the signature of `NewsItem` to have `link` always as `str` (See the commit "Adjust NewsItem")

steffenkleinle: Would it be a big change to use `externalUrl` instead of `link`?

MizukiTemma: @steffenkleinle Done ✏️

integreat_cms/news_managers/tunews_manager.py
MizukiTemma: Applied (see the commit "change slug -> news_raw_id")

docs/src/api-docs.rst
```diff
@@ -705,6 +705,80 @@ RESPONSE
...
+Response
+~~~~~~~~
+
+.. code:: javascript
+
+   [
+      {
+         "id": String,                          // The id of the news
+         "title": String,                       // The title of the news
+         "message": String,                     // The content of the news
```
steffenkleinle: 🤔 Could we perhaps also update this to `content` to keep it consistent with other endpoints?

docs/src/api-docs.rst
```diff
@@ -705,6 +705,80 @@ RESPONSE
...
+         "last_updated": String,                // Date of last update
+         "display_date": String,                // Date of publication or last update, of which later
+         "channel": String | null,              // The channel of the push notification, null for external news
+         "available_languages": [               // The available languages of the push notification
+               "<language_slug>": {
+                  "id": Number | null,          // The id of the translation
+               },
+               ...
+            ],
+         "source": String,                      // The source of the news (push news, Tü News, Amal News)
```
steffenkleinle:
```suggestion
         "source": String,                      // The source of the news ('local', 'tu-news' or 'amal-news')
```

docs/src/api-docs.rst
```diff
@@ -705,6 +705,80 @@ RESPONSE
...
+         "id": String,                          // The id of the news
+         "title": String,                       // The title of the news
+         "message": String,                     // The content of the news
+         "timestamp": String,                   // Date of last update, will be deprecated
+         "last_updated": String,                // Date of last update
+         "display_date": String,                // Date of publication or last update, of which later
+         "channel": String | null,              // The channel of the push notification, null for external news
+         "available_languages": [               // The available languages of the push notification
+               "<language_slug>": {
+                  "id": Number | null,          // The id of the translation
```
steffenkleinle: 🤔 Shouldn't this also be the stringified id?
```suggestion
                  "id": String | null,          // The id of the translation
```

docs/src/api-docs.rst
```diff
@@ -705,6 +705,80 @@ RESPONSE
...
+~~~~~~~~
+
+.. code:: javascript
+
+   [
+      {
+         "id": String,                          // The id of the news
+         "title": String,                       // The title of the news
+         "message": String,                     // The content of the news
+         "timestamp": String,                   // Date of last update, will be deprecated
```
steffenkleinle: Since its a new endpoint, you could just remove this already

MizukiTemma:
@steffenkleinle 

I just noticed one thing: the endpoint `fcm/` will return the id field with prefix.
Is the app using this endpoint and will that be a problem?

steffenkleinle:
> @steffenkleinle
> 
> I just noticed one thing: the endpoint `fcm/` will return the id field with prefix. Is the app using this endpoint and will that be a problem?

Yes the app is using this endpoint and yes, this is a breaking change. Most things MIGHT work anyway, but at least the linking from push notifications to local news will break. Please avoid that.

docs/src/api-docs.rst
MizukiTemma: Do you mean "timestamp"?
It is not needed anymore in `news` and `news<id>`?

steffenkleinle: In this endpoint it would not be needed since it is new. However, for the `fcm` endpoint we still need it, so if it is easier to just keep it in both places thats okay as well.

MizukiTemma:
@steffenkleinle @hannaseithe 
All suggestions are now either implemented or answered. Please re-review 🙏

hannaseithe (commented):
This looks good to me.

But I have two questions about the UX:
1.) the single news endpoint returns old news, which the general news endpoint seems to filter out. Is that intended?
2.) the single news endpoint also returns unsent push_notifications, while the general news endpoint does not. Is that intended?

MizukiTemma:
> This looks good to me.
> 
> But I have two questions about the UX: 1.) the single news endpoint returns old news, which the general news endpoint seems to filter out. Is that intended? 2.) the single news endpoint also returns unsent push_notifications, while the general news endpoint does not. Is that intended?

@steffenkleinle 
Should we filter such cases out?

steffenkleinle:
> > This looks good to me.
> > But I have two questions about the UX: 1.) the single news endpoint returns old news, which the general news endpoint seems to filter out. Is that intended? 2.) the single news endpoint also returns unsent push_notifications, while the general news endpoint does not. Is that intended?
> 
> @steffenkleinle Should we filter such cases out?

Not necessary from our side, I think its more a question for regions if they are okay with old news/unpublished news to still be (in theory) accessible. So most likely a question for @osmers / the service team.

MizukiTemma:
> > > This looks good to me.
> > > But I have two questions about the UX: 1.) the single news endpoint returns old news, which the general news endpoint seems to filter out. Is that intended? 2.) the single news endpoint also returns unsent push_notifications, while the general news endpoint does not. Is that intended?
> > 
> > 
> > @steffenkleinle Should we filter such cases out?
> 
> Not necessary from our side, I think its more a question for regions if they are okay with old news/unpublished news to still be (in theory) accessible. So most likely a question for @osmers / the service team.

@steffenkleinle 
In what scenario can an old news be accessed by/shown to users? This may help @osmers to decide 👀

steffenkleinle:
> > > > This looks good to me.
> > > > But I have two questions about the UX: 1.) the single news endpoint returns old news, which the general news endpoint seems to filter out. Is that intended? 2.) the single news endpoint also returns unsent push_notifications, while the general news endpoint does not. Is that intended?
> > > 
> > > 
> > > @steffenkleinle Should we filter such cases out?
> > 
> > 
> > Not necessary from our side, I think its more a question for regions if they are okay with old news/unpublished news to still be (in theory) accessible. So most likely a question for @osmers / the service team.
> 
> @steffenkleinle In what scenario can an old news be accessed by/shown to users? This may help @osmers to decide 👀

Only if they know the correct id, either by accessing an old URL which contains the id or by guessing :)

osmers:
If something was not sent or was removed, it should not be accessible anymore/at all. So to me that would mean that they should not be delivered and need to be filtered out.

MizukiTemma:
@hannaseithe @steffenkleinle 

The same filter is now applied to the single endpoitn and the common endpoint. Archived, not yet sent or old news are not returned.

hannaseithe (approved):
The filtering looks good to me

----
I am sorry to have another remark - I hope I am not being too annoying! 

But it seems we do not apply any timeframe filtering to tuenews. Maybe there we should also only show the news from within the last `settings.FCM_HISTORY_DAYS`?

Since this is a smaller detail, I pre-approve and l leave it up to you :)

MizukiTemma:
> The filtering looks good to me
> 
> I am sorry to have another remark - I hope I am not being too annoying!
> 
> But it seems we do not apply any timeframe filtering to tuenews. Maybe there we should also only show the news from within the last `settings.FCM_HISTORY_DAYS`?
> 
> Since this is a smaller detail, I pre-approve and l leave it up to you :)

You are not annoying, no worries 😃 

That point was once or twice raised somewhere and it was decided just to collect all available news posts.
I'll create a SRS for the external news in general to document it and other points.

integreat_cms/news_managers/abstract_news_manager.py
```diff
@@ -6,22 +6,21 @@
...
-    message: str
-    timestamp: datetime
+    content: str
     last_updated: datetime
     display_date: datetime
     channel: str | None
     available_languages: dict | None
     source: str
-    link: str | None
+    externalUrl: str
```
jonbulz, line 23: this change also applies to the existing `fcm` endpoint. Is that intentional?

integreat_cms/api/v3/news.py
```diff
@@ -82,3 +84,42 @@ def news(
...
+    :param request: Django request
+    :param region_slug: slug of a region
+    :param language_slug: language slug
+    :param news_id: id of the requested news item
+    :return: JSON single NewsItem
+    """
+    region = request.region
+    language = region.get_language_or_404(language_slug, only_active=True)
+
+    parts = news_id.split("-", 1)
```
jonbulz: this breaks if the `NewsItem` name contains a -. We could use `rsplit("-", 1)` instead, that would still work in that case

docs/src/api-docs.rst
```diff
@@ -705,6 +705,78 @@ RESPONSE
...
+         "last_updated": String,                // Date of last update
+         "display_date": String,                // Date of publication or last update, of which later
+         "channel": String | null,              // The channel of the push notification, null for external news
+         "available_languages": [               // The available languages of the push notification
+               "<language_slug>": {
+                  "id": String | null,          // The id of the translation
+               },
+               ...
+            ],
+         "source": String,                      // The source of the news ('local', 'tu-news' or 'amal-news')
```
jonbulz: comment say `tu-news`, it`s actually `tuNews`. it's just a comment, but we should try to keep these accurate

docs/src/api-docs.rst
```diff
@@ -705,6 +705,78 @@ RESPONSE
...
+      "last_updated": String,                // Date of last update
+      "display_date": String,                // Date of publication or last update, of which later
+      "channel": String | null,              // The channel of the push notification, null for external news
+      "available_languages": [           // The available languages of the push notification
+            "<language_slug>": {
+               "id": String | null,           // The id of the translation
+            },
+            ...
+         ],
+      "source": String,                      // The source of the news ('local', 'tu-news' or 'amal-news')
```
jonbulz: same as above, should be `tuNews`

integreat_cms/news_managers/pushnews_manager.py
```diff
@@ -86,25 +63,25 @@ def collect_news_items(
...
+        news_raw_id: str,
     ) -> HttpResponse:
         """
         Tries rendering the social media headers for a news page in a specified region and language
         """
         if not (
             pn_translation := PushNotificationTranslation.objects.filter(
                 language__slug=language.slug,
-                push_notification__id=slug,
+                push_notification__id=news_raw_id,
```
jonbulz, line 84: are you sure it's `push_notification__id`? everywhere else, we treat `news_raw_id` as a `PushNotificationTranslation` id

jonbulz (changes requested):
A few minor change requests :) I think the most important one is to clarify if changing `link` to `externalUrl` in the `fcm` endpoint breaks anything in the app

docs/src/api-docs.rst
MizukiTemma: Updated ✏️

integreat_cms/news_managers/abstract_news_manager.py
MizukiTemma: Aha, right 👀 as we have now a separate transform function for `fcm` endpoint we can get rid of "source" and externalUrl 💡 removed from `transform_notification_for_fcm`

integreat_cms/news_managers/pushnews_manager.py
MizukiTemma: It seems this has been implemented in this way since the time before #4195 👀 So it shuold not break any existing behaviours related to this part.

But it's not inconsistent that `PushnewsManager.social_media_headers` searchs by push notification id while `TunewsManager.social_media_headers` by translation id 🤔 I'll talk with App Team and will do adjustments if needed but as separate PR, as it's not breaking and it's outside of scope of the original issues.

MizukiTemma:
@jonbulz 
Thank you that you reviews such soon 😸 The suggestions are either applied or replied.

integreat_cms/news_managers/abstract_news_manager.py
steffenkleinle: We are not using the `link` from the fcm endpoint, so this could safely be renamed.

integreat_cms/news_managers/pushnews_manager.py
jonbulz: yes, seems like this is not changing existing behaviour. definitely seems wrong to me, though, so worth a follow-up 😅

jonbulz (approved):
Since the one remaining issue (`PushNotification__id` vs `PushNotificationTranslation__id` for `news_raw_id`) is not introduced here, I think this is good to go. But we should investigate that issue separately, otherwise we may hand out wrong social media headers

MizukiTemma:
> Since the one remaining issue (`PushNotification__id` vs `PushNotificationTranslation__id` for `news_raw_id`) is not introduced here, I think this is good to go. But we should investigate that issue separately, otherwise we may hand out wrong social media headers

A thread is opened 🧵 
https://chat.tuerantuer.org/digitalfabrik/pl/45n114usn3yzpqfatbipo5aq5r

MizukiTemma:
@steffenkleinle FYI: merged 😸

PR #4376, opened 2026-06-18
Bugfix: Add js sanitization to clean_content

### LINKED ISSUE #4373: Security: Stored XSS — script tags and event handlers saved to DB and served via public API
## Bug Summary

Script tags (`<script>`) and JavaScript event handler attributes (e.g. `onerror`, `onclick`) submitted via the page/event/POI content editor are persisted to the database without sanitization and served raw through the public REST API (`/api/v3/`).

The current protections are entirely client-side:
- The CMS edit view relies on TinyMCE to strip scripts before submission (bypassable via direct HTTP request)
- The React web app sanitizes HTML before DOM insertion

Neither protection is applied at the server/API layer, so any API consumer that does not sanitize — including third-party integrations and potentially the native Integreat app's WebView — receives raw executable HTML.

## Expected Behavior

`clean_content` (called during form save) should strip `<script>` tags and JavaScript event handler attributes before the content is stored. The public API should never serve executable script content.

## Actual Behavior

Submitting the following payload via a direct POST to the page edit endpoint (bypassing TinyMCE):

```html
<p><img src='x' onerror='alert("XSS")'> and <script>alert('XSS')</script></p>
```

Results in both payloads being stored in the database and returned verbatim by the public API:

```json
{
  "content": "<p><img src=\"x\" onerror=\"alert(&quot;XSS&quot;)\"> and <script>alert('XSS')</script></p>"
}
```

## Steps to Reproduce

1. Log in to the CMS as any user with page edit permissions
2. Navigate to a page's edit view, e.g. `/testumgebung/pages/de/45285/edit/`
3. Open browser DevTools → Network tab, submit the form once to capture a valid CSRF token
4. In the DevTools console, run a modified fetch replacing the `content` field with the XSS payload
5. Call the public API: `GET /api/v3/testumgebung/de/pages/`
6. Find the edited page — both `onerror` and `<script>` are present in the `content` field

## Reproducibility

Always (100%)

## Environment

- **Environment:** Staging (cms-test.integreat-app.de)
- **URL:** `/testumgebung/pages/de/45285/edit/`
- **Browser:** Chromium 147
- **User role:** Any role with page edit permissions (Editor, Management, Staff)

### Network Response (API)
```json
{
  "content": "<p><img src=\"x\" onerror=\"alert(&quot;XSS&quot;)\"> and <script>alert('XSS')</script></p>"
}
```

## Severity

Native app impact (React Native WebView) is under investigation — WebViews execute both `<script>` and event handler attributes by default unless explicitly disabled. The `onerror` event handler fires even when `Content-Security-Policy: script-src 'none'` is set.

### Affected Components / Services
- `integreat_cms/cms/utils/content_utils.py` — `clean_content()` performs no sanitization of script tags or event handlers
- `integreat_cms/cms/forms/custom_content_model_form.py` — calls `clean_content()` on save

### Relevant Endpoints
- `POST /<region>/pages/de/<id>/edit/` — write path (no sanitization)
- `GET /api/v3/<region>/de/pages/` — read path (serves raw content)
- Same pattern applies for events and POIs

### Suspected Root Cause
`clean_content()` uses lxml for HTML transformation but does not invoke `lxml.html.clean.Cleaner` or equivalent to strip scripts and event handler attributes.

## Acceptance Criteria for Fix

- [ ] `clean_content()` strips `<script>`, `<iframe>`, `<object>` and similar tags
- [ ] `clean_content()` strips JavaScript event handler attributes (`onerror`, `onclick`, `onload`, etc.)
- [ ] The public API never returns executable script content
- [ ] No regression in legitimate HTML content (links, images, contact cards, formatting)

## Verification / Testing

- [ ] Reproduce original bug (direct POST with script payload) → API response no longer contains script/event handler
- [ ] Existing unit tests in `tests/cms/utils/test_content_utils.py` still pass
- [ ] Add test: `clean_content` with `<script>` input → output does not contain `<script>`
- [ ] Add test: `clean_content` with `onerror` attribute → output does not contain `onerror`

## Related Issues

- #4370 — IDOR: Contact card endpoint does not enforce region scoping (related finding from same security scan)

hannaseithe:
### Short description

This fixes the security vulnerability bug, where we did not js-sanitize page/event/poi content on the server side.

### Proposed changes

- add lxml html cleaner to clean_content function
- tests

For context: https://chat.tuerantuer.org/digitalfabrik/pl/fjx9bkmxxtn4im4y1yeb9atymy

### Side effects

- Pre-existing injected js have been fixed for latest_versions

### How to test

- follow the steps to reproduction on original issue

### Resolved issues

Fixes: #4373 


MizukiTemma (approved):
Thank you 😸 I really like you always put tests together 🧪 

Should we maybe add the same mechanism to the push notifications?

seluianova (approved):
cool finding 👍
the described issue is not reproduced ✅

hannaseithe:
> Should we maybe add the same mechanism to the push notifications?

Good point - I went ahead and checked on the app - and on native the text field of a pn_translation gets actually executed - which the app has to fix.

But I am not sure if with push notifications we should also provide backend sanitization: Does the responsibility actually lie with us? We treat the text field on the translation of a push notification as simply a text field. In contrast: The content field on page/event/poi intentionally is html-string - we expect it to be rendered as html. So not sure what is the best approach. What do you think?

EDIT: That is the bug on the app, to sanitize on their end for pages and pushnotifications as well 
https://github.com/digitalfabrik/integreat-app/issues/4199

MizukiTemma:
> > Should we maybe add the same mechanism to the push notifications?
> 
> Good point - I went ahead and checked on the app - and on native the text field of a pn_translation gets actually executed - which the app has to fix.
> 
> But I am not sure if with push notifications we should also provide backend sanitization: Does the responsibility actually lie with us? We treat the text field on the translation of a push notification as simply a text field. In contrast: The content field on page/event/poi intentionally is html-string - we expect it to be rendered as html. So not sure what is the best approach. What do you think?
> 
> EDIT: That is the bug on the app, to sanitize on their end for pages and pushnotifications as well [digitalfabrik/integreat-app#4199](https://github.com/digitalfabrik/integreat-app/issues/4199)

Ah, ok :) Thank you for checking!

PR #4383, opened 2026-06-24
Improve Frag Integreat message processing

### LINKED ISSUE #480: Zammad field for chat ID
To be more flexible with the mapping between regions and Zammad installations, we should save the chat ID in an additional field of the ticket that cannot be modified.

This allows the webhook to send the chat ID and therefore the identification of the chat w/o any additional information. This in turns allows multiple regions to use the same Zammad server.

We have to change 3 things:
- update the Zammad setup steps
- change all existing Zammad servers
- change the CMS function that handles the Zammad web hook to fetch the ticket by chat ID

svenseeberg:
### Short description
Improve Frag Integreat message processing by removing unnecessary cache and flushing cache in case of human responses. This also adds support for a single Zammad that can be used across all regions of the CMS.

### Proposed changes

* The automatic chat answers attribute was cached and therefore changing the value in Zammad did not immediately take effect. As it is only accessed when a user posts a message, we can directly read it from Zammad and avoid any Cache problems.
* The generating answer flag should only be true if we actually generate and answer.
* Add support for single Zammad for all regions by supporting a device_id attribute in Zammad tickets.
* Set the device_id attribute when multi region Zammad setting is true.
* Flush message cache when a new human message webhook triggers

### Side effects
- There should be none. The removed cache is negligible.

### How to test
Due to the difficulty of setting up the full stack, I would recommend to temporarily replace the files on the test server:
```
cd /opt/integreat-cms
cp ./.venv/lib/python3.13/site-packages/integreat_cms/api/v3/chat/user_chat.py ./.venv/lib/python3.13/site-packages/integreat_cms/api/v3/chat/user_chat.py.bak
cp ./.venv/lib/python3.13/site-packages/integreat_cms/cms/utils/zammad.py ./.venv/lib/python3.13/site-packages/integreat_cms/cms/utils/zammad.py.bak
wget https://github.com/digitalfabrik/integreat-cms/blob/fix/chat-cache-issues/integreat_cms/cms/utils/zammad.py -O ./.venv/lib/python3.13/site-packages/integreat_cms/cms/utils/zammad.py
wget https://github.com/digitalfabrik/integreat-cms/blob/fix/chat-cache-issues/integreat_cms/api/v3/chat/user_chat.py -O ./.venv/lib/python3.13/site-packages/integreat_cms/api/v3/chat/user_chat.py
systemctl restart apache2.service
```

Test chat on https://beta.integreat.app/testumgebung/de, then undo changes:

```
mv ./.venv/lib/python3.13/site-packages/integreat_cms/api/v3/chat/user_chat.py.bak ./.venv/lib/python3.13/site-packages/integreat_cms/api/v3/chat/user_chat.py
cp ./.venv/lib/python3.13/site-packages/integreat_cms/cms/utils/zammad.py.bak ./.venv/lib/python3.13/site-packages/integreat_cms/cms/utils/zammad.py
systemctl restart apache2.service
```

### Resolved issues
Fixes: https://github.com/digitalfabrik/integreat-chat/issues/480
Fixes: reported problem were changed "automatic_answers" does not take effect immediately

### DIFF
--- a/integreat_cms/api/v3/chat/user_chat.py
+++ b/integreat_cms/api/v3/chat/user_chat.py
@@ -11,6 +11,7 @@ from typing import TYPE_CHECKING
 
 import requests
 from django.conf import settings
+from django.core.cache import cache
 from django.http import HttpResponse, JsonResponse
 from django.shortcuts import get_object_or_404
 from django.utils.dateparse import parse_datetime
@@ -102,9 +103,9 @@ def process_chat_payload(
         )
         user_chat.language = language
         user_chat.save()
-        user_chat.processing_answer = True  # type: ignore[assignment]
         if response is not None:
             if user_chat.automatic_answers:
+                user_chat.processing_answer = True  # type: ignore[assignment]
                 celery_translate_and_answer_question.apply_async(
                     args=[
                         parse_datetime(response["updated_at"]),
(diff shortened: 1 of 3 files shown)


svenseeberg:
Manual changes and reviewed by Opus 4.8

hannaseithe:
I tried to run the commands on the test-systems but I am lacking permissions and a few more errors:
```
cp: cannot create regular file './.venv/lib/python3.13/site-packages/integreat_cms/api/v3/chat/user_chat.py.bak': Permission denied
cp: cannot create regular file './.venv/lib/python3.13/site-packages/integreat_cms/cms/utils/zammad.py.bak': Permission denied
./.venv/lib/python3.13/site-packages/integreat_cms/cms/utils/zammad.py: Permission denied
./.venv/lib/python3.13/site-packages/integreat_cms/api/v3/chat/user_chat.py: Permission denied
Failed to execute /usr/bin/pkttyagent: No such file or directory
Failed to restart apache2.service: Access denied
See system logs and 'systemctl status apache2.service' for details.
```

Would using `sudo` everywhere solve that?

svenseeberg:
Yes, I assumed an initial `sudo -i`

hannaseithe:
I realized that I am not sure what I am supposed to be testing. I assumed that toggling the "Automatic Chatbot Answers" field on the Ticket in Zammad would be the point, but that still showed the typing indicator for a while and then it disappeared.

svenseeberg:
> I realized that I am not sure what I am supposed to be testing.

Its mostly "the Chat still works". There are 2 new features:

- The Zammad Chat object can be fetched by the device_id. This requires configuring the Zammad server to have the additional device_id object attribute. However, this is just the first part which uses the attribute. We need a second PR that actually sets this value. Maybe I can amend it here :thinking: 
- Removing the cache should allow the "Automatic Answers" attribute in Zammad Tickets to take immediately. If you want to test it, write questions in the Chat and switch between on/off in the Zammad ticket. That should lead to either responses being created or not.

svenseeberg:
> * The Zammad Chat object can be fetched by the device_id. This requires configuring the Zammad server to have the additional device_id object attribute. However, this is just the first part which uses the attribute. We need a second PR that actually sets this value. Maybe I can amend it here 🤔

This is now done when the `INTEGREAT_CMS_MULTI_REGION_ZAMMAD` variable is true.

MizukiTemma (approved):

Issue #4284, opened 2026-05-12
After saving a published page as a draft all the existing translations become "outdated" (from draft to publish happens the same)


AnnaPerelazova:
### Bug Summary

if you save a page with existing translations as draft, the translations get an "outdated"-status 

### Expected Behavior

the translations shouldnt be shown as "outdated", as the the content of the page didnt change

### Actual Behavior

as described above

### Steps to Reproduce

1. Go to on any published page with existing automatic translations e.g. https://admin.integreat-app.de/testumgebung/pages/de/44929/edit/
2. save it as draft ("Als Entwurf speichern")
3. all the translations get an outdated (exclamation mark) status 

### Reproducibility

Always (100%)

### Environment

cms-bug, supposedly reproduceable in every browser

### Severity

None

### Related Issues

https://github.com/digitalfabrik/integreat-cms/issues/4277
https://github.com/digitalfabrik/integreat-cms/issues/4254

Issue #4254, opened 2026-04-23
Existing translation outdated after machine translation of unchanged source translation


osmers:
### Bug Summary

It should be an edge case. 
This is a video for better understanding:

https://github.com/user-attachments/assets/15581da3-8046-44ff-a61d-81edfd5f56a1

### Expected Behavior

Spanish should keep the robot icon. The translation was up to date before and still is, since nothing changed in German/source translation.

Red warning message can still appear, but it should read the following (adjusted to this use case, i.e. mentioning the languages affected): 

<img width="497" height="67" alt="Image" src="https://github.com/user-attachments/assets/a15e25fa-8e57-4526-8475-b49830a373ed" />

This screenshot is from bulk translating up to date translations.

### Actual Behavior

We see a red message and the translation status is outdated:

<img width="1576" height="878" alt="Image" src="https://github.com/user-attachments/assets/c04ade7d-b050-4365-8cd2-1f30ad891fc5" />

### Steps to Reproduce

1. Go to [Testumgebung](https://admin.integreat-app.de/testumgebung/pages/de/54280/edit/)
2. Check box Automatische Übersetzungen
3. Check one additional new language
4. Click Update
5. See that previous robot icons (e.g. Russian and Spanish) now show status outdated

### Reproducibility

Always (100%)

### Environment

- **Environment**: CMS production and test
- **Version**: 2026.4.0 (production), 2026.5.2-alpha (test)

### Severity

None

osmers:
While creating this issue, I also tested on test-cms and could not reproduce it. I guess that means a new release would mean, that is is not a bug anymore @MizukiTemma ?

osmers:
Never mind that about the test system - I just tested it there again and could reproduce it. Will change the issue description accordingly.

dkehne:
## Root Cause

The bug is triggered by `CustomContentModelForm.save()` in `integreat_cms/cms/forms/custom_content_model_form.py`. It unconditionally creates a **new version** of the source translation (version increment + `pk = None`) with a fresh `last_updated` timestamp — even when the editor changed nothing in the content.

The outdated status is then determined in `AbstractContentTranslation.translation_state` (`abstract_content_translation.py`, line 491):

```python
translation.last_updated <= source_translation.last_updated
```

Since the source (e.g. German) now has a newer `last_updated` than the existing target translations (e.g. Spanish, Russian), those are incorrectly marked as outdated.

## Proposed Fix

[...]

PR #4387, opened 2026-06-24
Bugfix: Update translation state to up to date when an outdated translation is published without changes

### LINKED ISSUE #4385: Publishing an outdated translation does not update its status to up-to-date
### Bug Summary

Publishing an outdated translation displays a success message indicating that the date was updated, but the translation status remains outdated instead of being marked as up-to-date.

### Expected Behavior

The English translation should be marked as up-to-date after publishing.

### Actual Behavior

The message `Keine Änderungen vorgenommen, aber Datum aktualisiert` is displayed, but the English translation remains outdated.

### Steps to Reproduce

1. Create a new page with a German translation and click `Publish`.
2. Add an English translation for the page (either manually or via automatic translation).
3. Edit the German translation so that the English translation becomes outdated.
4. Open the English translation and click `Update`.

### Reproducibility

Always (100%)

### Environment

Current develop and RC branches.
_Not reproduced on production._

Image

### Severity

None

### Regression Information

_No response_

### Technical Context

_No response_

_No response_

### Verification / Testing

_No response_

### Related Issues

_No response_

seluianova:
### Short description
When evaluating `translation_state`, we use the latest major version of a translation.
When a translation is saved without content changes, it is marked as a "minor edit". 
This leads to inconsistent behavior: 
Publishing an outdated translation without changes does not update it to "up to date".

### Proposed changes
- Adjust `translation_state` property implementation: determine whether a translation is outdated using the latest public or draft version instead of the latest major version.

### Side effects
Previously, the translation state was evaluated against the latest major version. This logic has been in place for many years.
Changing this behavior may affect other parts of the system, although I have not identified any issues so far.
Careful testing is required.

### How to test
- Follow the reproduction steps described in the issue.
- Think of any other scenarios where the `translation_state` can be used & properly displayed.

### Resolved issues

Fixes: #4385

### DIFF
--- a/integreat_cms/cms/models/abstract_content_translation.py
+++ b/integreat_cms/cms/models/abstract_content_translation.py
@@ -393,6 +393,15 @@ class AbstractContentTranslation(AbstractBaseModel):
         return self.foreign_object.get_translation(self.language.slug)
 
     @cached_property
+    def latest_public_or_draft_version(self) -> AbstractContentTranslation | None:
+        """
+        This property is a link to the most recent public or draft version of this translation.
+
+        :return: The latest public or draft revision of the translation
+        """
+        return self.foreign_object.get_public_or_draft_translation(self.language.slug)
+
+    @cached_property
     def public_version(self) -> AbstractContentTranslation | None:
         """
         This property is a link to the most recent public version of this translation.
@@ -488,7 +497,11 @@ class AbstractContentTranslation(AbstractBaseModel):
             # If the source translation is already outdated, this translation is as well
             or source_translation.translation_state == translation_status.OUTDATED
             # If the translation was edited before the last major change in the source language, it is outdated
-            or translation.last_updated <= source_translation.last_updated
+            or (
+                self.latest_public_or_draft_version
+                and self.latest_public_or_draft_version.last_updated
+                <= source_translation.last_updated
+            )
         ):
             return translation_status.OUTDATED
         if translation.machine_translated:
--- a/integreat_cms/cms/models/abstract_content_translation.py
+++ b/integreat_cms/cms/models/abstract_content_translation.py
@@ -490,6 +490,8 @@ class AbstractContentTranslation(AbstractBaseModel):
         if not self.source_language:
             # If the language of this translation is the root of this region's language tree, it is always "up to date"
             return translation_status.UP_TO_DATE
+
+        latest_translation = self.latest_public_or_draft_version or translation
         if (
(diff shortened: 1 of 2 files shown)


seluianova:
❌ CMS became very slow on this branch. It seems the fix should be reworked.

**UPD**. fixed

MizukiTemma (approved):
Thank you for speedy analysis and fix 😸 

The reported bug is solved and #4254 and #4284 remain solved too ✅

PeterNerlich (approved):
This works and I don't think there should be any weird side effects. The usages of the `translation_state` property and the `is_outdated` and `is_up_to_date` properties where it's used seem limited and innocuous.

jonbulz:
This does introduce additional queries, because only major translations are prefetched e.g in `page_tree_view` and `partial_page_tree_view`. After these changes, we'll have one additional query per page when computing the translation state.
We have a `prefetch_public_or_draft_translations` helper, but it is never used. I think we should do that for every queryset that iterates multiple objects and computes translation state. We could also do this in a separete PR that takes a look on performance. I think there are also other some views that currently don't prefetch at all.

There is also an edge case if no public or draft version exist for a translation. If the only version is in the auto_save or review state, it will also show as up to date.

But as an urgent fix for a release blocker, I'm fine with releasing this. I'd be in favor of opening an issue for the performance improvement then, though.

seluianova:
> There is also an edge case if no public or draft version exist for a translation. If the only version is in the auto_save or review state, it will also show as up to date.

true, I can fix this one quickly by changing the condition to:
if there is a public or draft version - use it in the check.
if not - check the last major version.

integreat_cms/cms/models/abstract_content_translation.py
```diff
@@ -481,14 +490,16 @@ def translation_state(self) -> str:
         if not self.source_language:
             # If the language of this translation is the root of this region's language tree, it is always "up to date"
             return translation_status.UP_TO_DATE
+
+        latest_translation = (self.latest_public_or_draft_version or translation)
         if (
             # If the source language does not have a major public version, the translation is considered "outdated",
             # because the content is not in sync with its source translation
             not (source_translation := self.major_source_translation)
```
seluianova, line 498: btw I think this condition is also a bit questionable (?)
e.g. if Arabic depends on English, why can't I manually create Arabic translation without having English translation? why Arabic translation should always be "outdated" in this case?

seluianova:
Task for the performance improvement created ✅ https://github.com/digitalfabrik/integreat-cms/issues/4392

Issue #4315, opened 2026-05-27
Media Library - Searching a file with its alternative text breaks the media library


juliankehne:
### Bug Summary

Its possible to give files in the media library an alternative text.
Searching for the alternative text gives the file name as a search suggestion, clicking search or pressing enter to execute the search will result in the media library being stuck in loading indefinitely and the media library becomes unresponsive.

### Expected Behavior

The search works showing the correct results. The media library stays responsive after the search is completed.

### Actual Behavior

The search doesn't work for alternative texts.
The media library becomes unresponsive and is stuck in loading.

### Steps to Reproduce

1. Go to 'Media Library'
2. Give a file any 'Alternative text'
3. Search the media library for the alternative text
4. See that the search suggestion includes the file that has the alternative text
5. Click search

### Reproducibility

Always (100%)

### Environment

Environment: all

### Screenshots / Videos / Logs

Unresponsive media library after searching for an alternative text:
<img width="1606" height="779" alt="Image" src="https://github.com/user-attachments/assets/d5434b7d-c840-4c73-a3df-5e6c90ce32af" />

### Severity

🟡 Minor — Non-core functionality affected, acceptable workaround exists

### Regression Information

Regression? Unknown

### Acceptance Criteria for Fix

- [ ] Steps to reproduce cannot be reproduced
- [ ] Search in the media library works for file name and alternative texts

### Related Issues

There are other issues where the media library becomes unresponsive and they might be related:

https://github.com/digitalfabrik/integreat-cms/issues/4271
https://github.com/digitalfabrik/integreat-cms/issues/4157

hannaseithe:
Cloased as PR was merged

juliankehne:
Cannot reproduce on 2026.7.3-alpha.

Issue #4271, opened 2026-05-06
Media library unresponsive after deleting a file that was part of search result or in a subfolder


juliankehne:
### Bug Summary

After deleting a file (image, pdf, ... ) that was part of a search or is in a subfolder the media library becomes unresponsive and the page has to be refreshed to keep working. 

Deleting a file without searching for it when its in the main directory works.

### Expected Behavior

After searching a file and then deleting it the media library should keep working.
When deleting a file that is in a folder the media library should keep working.

### Actual Behavior

After searching a file and then deleting it the media library stops working.
When deleting a file that is in a folder the media library stops working.

### Steps to Reproduce

1. Go to Media Library
2. Upload a file
3. Search the file name
4. Select the file
5. Delete the file

and

1. Go to Media Library
2. Create a new directory
3. Upload a file to the new directory
4. Select the file
5. Delete the file

### Reproducibility

Always (100%)

### Environment

production / cms-test / integreat-rc 
Version: 2026.5.2-alpha
Version: 2026.5.0-rc

### Severity

🟡 Minor — Non-core functionality affected, acceptable workaround exists

### Regression Information

Unknown

### Acceptance Criteria for Fix

- [ ] Steps to reproduce cannot be reproduced
- [ ] Any file that is not being used or read-only can be deleted without the media library stopping to work

### Related Issues

Potentially related: The media library behaviour after the deletion is similar to the outcome of this bug: https://github.com/digitalfabrik/integreat-cms/issues/4157

hannaseithe:
Cloased as PR was merged

juliankehne:
Cannot reproduce it with either of the steps on 2026.7.3-alpha.

Issue #4157, opened 2026-02-18
Media Library breaks when closing it while images are still loading


juliankehne:
### Describe the Bug

When opening the media library in pages, events, locations, organization and other places where it opens in a separate window on top of your current page and then closing it while it is loading images. 
After opening it again it opens a broken version in which no images are loaded and most buttons are unresponsive.

Found on - Version: 2026.2.18-alpha
I wasn't able to reproduce this on production - Version 2026.2.0

### Steps to Reproduce

1. Go to 'Pages' (Or any place where content can have a logo added)
2. Open an existing page with a logo
3. Click 'Change logo'
4. In the opened media library open one of the folders that has images, e.g. "region logos"
5. While it is loading the images close the media library (X top right)
6. Click 'Change logo' to open the media library again

### Expected Behavior

- The media library opens and images load for the last opened folder
- navigation inside it works, e.g. clicking 'Home'
- Search is usable
- All buttons are responsive

### Actual Behavior

- The media library opens, but no images load
- navigation inside it is no longer possible
- Search doesn't work
- Filter unused images, Create Directory and Upload File buttons are unresponsive

### Screenshot

What it looks like in broken state:
<img width="1606" height="797" alt="Image" src="https://github.com/user-attachments/assets/7260809b-c175-4cb6-ac6c-ac30425d3ce5" />

### Additional Information

<details><summary>Traceback</summary>

```
```

</details>

### Related Issues

This might have introduced this bug: https://github.com/digitalfabrik/integreat-cms/pull/4079

This bug might be related in some way: https://github.com/digitalfabrik/integreat-cms/issues/3226

osmers:
I could just reproduce it on production - Version 2026.3.0

juliankehne:
Cannot reproduce on 2026.7.3-alpha.
Media Library works as expected.

PR #4394, opened 2026-06-29
Media Library: Nullify file index before re-render is triggered

### LINKED ISSUE #4384: Media Library - Unresponsive after opening a file and then opening a folder
### Bug Summary

In the media library, after opening any file (image or pdf) and then opening a folder, the media library will be stuck in loading and unresponsive. A user needs to refresh the page to get it working again.

### Expected Behavior

Folder opens after a file was selected before it.

### Actual Behavior

After opening any file (image or pdf) and then opening a folder, the media library will be stuck in loading and unresponsive.

### Steps to Reproduce

1. Go to Media Library
2. Click on any file
3. Click on a folder to navigate into it

### Reproducibility

Always (100%)

### Environment

Environment: Production 2026.5.2, RC 2026.6.2rc

_No response_

### Severity

### Regression Information

Unknown

### Technical Context

_No response_

_No response_

### Verification / Testing

_No response_

### Related Issues

_No response_

hannaseithe:
### Short description

Set file index to null before re-render is triggered. Otherwise we run into a situation where the sidebar re-render fails on `edit-sidebar.tsx:94` because `file` is null.

### Proposed changes

- nullify the fileIndex right from the start before re-render is triggered

### Side effects

- None

### How to test

- follow the instructions from the original issue for reproduction

### Resolved issues

Fixes: #4384 #4271 #4315 possibly #4157 (hard to reproduce locally)

### DIFF
--- a/integreat_cms/static/src/js/media-management/directory-content-library.tsx
+++ b/integreat_cms/static/src/js/media-management/directory-content-library.tsx
@@ -33,6 +33,7 @@ const DirectoryContentLibrary = (props: LibraryProps) => {
     useEffect(() => {
         const loadDirectory = async () => {
             setLoading(true);
+            setFileIndex(null);
             const urlParams = new URLSearchParams({});
             if (directoryId) {
                 console.debug(`Loading directory with id ${directoryId}...`);
@@ -48,9 +49,6 @@ const DirectoryContentLibrary = (props: LibraryProps) => {
             } finally {
                 setLoading(false);
             }
-
-            // Close the file sidebar
-            setFileIndex(null);
         };
         loadDirectory();
         /* eslint-disable-next-line react-hooks/exhaustive-deps */
--- a/integreat_cms/release_notes/current/unreleased/4384.yml
+++ b/integreat_cms/release_notes/current/unreleased/4384.yml
@@ -1,0 +1,2 @@
+en: Fix media library getting stuck on the loading indicator when a file was selected.
+de: Behebe, dass die Medienbibliothek im Ladeindikator hängen bleibt, wenn eine Datei ausgewählt war.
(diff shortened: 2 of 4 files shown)


MizukiTemma (approved):
The bug is fixed and I did not notice any side effect 😸 Thank you!

💡 A release note that the media library won't fall into unresponsiveness anymore?

hannaseithe:
> The bug is fixed and I did not notice any side effect 😸 Thank you!
> 
> 💡 A release note that the media library won't fall into unresponsiveness anymore?

I would suggest not? The bug is such an edge case and furthermore I thought we do not really create release notes for bug fixes?

MizukiTemma:
The rule is "A release not is needed when the changes are visible for users". Users can actually face the bug and so the changes are visible. But I'm fine without a release note too, as it was not reported by users but found internally.

hannaseithe:
> The rule is "A release not is needed when the changes are visible for users". Users can actually face the bug and so the changes are visible. But I'm fine without a release note too, as it was not reported by users but found internally.

Na I'll add it, was just surprised to create release note for a bug. Thank you for explaining the rule to me

integreat_cms/static/src/js/media-management/directory-content-library.tsx
```diff
@@ -33,6 +33,7 @@ const DirectoryContentLibrary = (props: LibraryProps) => {
     useEffect(() => {
         const loadDirectory = async () => {
             setLoading(true);
+            setFileIndex(null);
```
seluianova, line 36: I think we need the same fix in the `filter-result-library.tsx`
If you click on any file and then on the “Show only used” filter button, the UI gets stuck as well.
Let me know if you prefer this as a separate ticket, and I will create one.

seluianova (approved):
The reported issue is fixed ✅
Just one nice-to-have suggestion.
Approving anyway.

integreat_cms/static/src/js/media-management/directory-content-library.tsx
hannaseithe: Super good catch - that actually fixes a few other issues as well. I will fix all of them here

PR #4399, opened 2026-06-30
Bugfix: Update translation state to machine_translated when an outdated translation is machine translated without changes

### LINKED ISSUE #4320: Translation status is not updated when source and target translations are empty
### Bug Summary

Translation status is not updated when both the source translation and the target translation are empty, even though the target translation is marked as outdated.

### Expected Behavior

- The English translation status should be updated to MACHINE_TRANSLATED.
- No translation request should be sent to the translation provider because the content is empty.

### Actual Behavior

The error message
`Page "Translation test" was not translated into 'English' because there were no changes to the source translation.`
is displayed, and the English translation remains outdated.

### Steps to Reproduce

1. Create a page with **empty** content and click "Publish".
2. Enable the "Automatic translation" and "Create new translations" checkboxes e.g. for the English language, and click "Publish" to create a new English translation --> **both translations are now empty and not outdated**
3. Enter some content in the German translation, **do not set** "Automatic translation" checkbox, and click "Publish" --> **the English translation is now marked as outdated**
4. Clear the content of the German translation again, enable "Automatic translation" and "Update translations" for the English language, and click "Publish".

### Reproducibility

Always (100%)

### Environment

All environments

Image

### Severity

None

### Regression Information

_No response_

### Technical Context

_No response_

_No response_

### Verification / Testing

_No response_

### Related Issues

_No response_

seluianova:
### Short description
This PR fixes the following issue:

1. Create two translations, e.g. German and English. The English translation should be either UP_TO_DATE or MACHINE_TRANSLATED.
2. Change the German translation, making the English translation outdated.
3. Change the German translation again **to restore the same content as in step 1** and enable automatic translation for English.

It was not possible to machine-translate the page to English because of the following error:

`Page "Translation test" was not translated into 'English' because there were no changes to the source translation`.

### Proposed changes
- Introduce the `refreshed_translations` flow in the `MachineTranslationApiClient` class for outdated pages with no changes.
- When a translation is OUTDATED and has no changes, create a new minor edit version without sending a new request to the MT API to update the translation state.

### Side effects
hopefully none

### How to test
Follow the steps in the bug description.

### Resolved issues
Fixes: #4320

### DIFF
--- a/integreat_cms/core/utils/machine_translation_api_client.py
+++ b/integreat_cms/core/utils/machine_translation_api_client.py
@@ -18,6 +18,7 @@ from django.utils.translation import ngettext, ngettext_lazy
 
 from ...cms.constants.machine_translatable_fields import TRANSLATABLE_FIELDS
 from ...cms.constants.machine_translation_budget import MINIMAL
+from ...cms.models.abstract_content_translation import AbstractContentTranslation
 from ...cms.utils.stringify_list import iter_to_string
 from ...textlab_api.utils import check_hix_score
 from .word_count import word_count
@@ -31,7 +32,6 @@ if TYPE_CHECKING:
         Region,
     )
     from ...cms.models.abstract_content_model import AbstractContentModel
-    from ...cms.models.abstract_content_translation import AbstractContentTranslation
 
 
 logger = logging.getLogger(__name__)
@@ -59,6 +59,7 @@ class MachineTranslationApiClient(ABC):
     form_class: ModelFormMetaclass
     #: Successful translations
     successful_translations: list[TranslationContext] = []
+    refreshed_translations: list[TranslationContext] = []
     #: Translations with an attached API failure
     failed_translations: list[str] = []
     #: Content objects untranslatable since no actual changes were made
@@ -98,6 +99,7 @@ class MachineTranslationApiClient(ABC):
         shown to the user.
         """
         self.successful_translations = []
+        self.refreshed_translations = []
         self.failed_translations = []
         self.failed_because_no_changes_made = []
         self.failed_because_no_source_translation = []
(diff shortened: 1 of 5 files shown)


MizukiTemma (approved):
Looks good 😸 Nice with a new test added 👍

integreat_cms/core/utils/machine_translation_api_client.py
```diff
@@ -172,15 +174,28 @@ def translate_queryset(
...
-        context = self.filter_insufficient_hix_score(context)
-        context = self.filter_exceeds_limit(context)
+        context_to_translate = self.filter_no_source_translation(changed_context)
+        context_to_translate = self.filter_insufficient_hix_score(context_to_translate)
+        context_to_translate = self.filter_exceeds_limit(context_to_translate)
 
-        # Provider-API-specific implementation
-        self.invoke_translation_api(context)
+        # Provider-API-specific implementation for changed contexts
+        self.invoke_translation_api(context_to_translate)
```
PeterNerlich: Super tiny nitpick, I would use the plural here always
```suggestion
        # Prepare all content objects
        all_contexts = self.prepare_content_objects()

        # Separate contexts into those with changes and those without
        changed_contexts = []
        unchanged_contexts = []

        for ctx in all_contexts:
            if bool(ctx.translatable_attributes):
                changed_contexts.append(ctx)
            else:
                unchanged_contexts.append(ctx)

        self.handle_unchanged_contexts(unchanged_contexts)

# Filter out content objects which can not be translated
        contexts_to_translate = self.filter_no_source_translation(changed_contexts)
        contexts_to_translate = self.filter_insufficient_hix_score(contexts_to_translate)
        contexts_to_translate = self.filter_exceeds_limit(contexts_to_translate)

        # Provider-API-specific implementation for changed contexts
        self.invoke_translation_api(contexts_to_translate)
```

integreat_cms/core/utils/machine_translation_api_client.py
```diff
@@ -190,38 +205,42 @@ def translate_queryset(
...
+            if (
+                isinstance(target_translation, AbstractContentTranslation)
+                and target_translation.is_outdated
+            ):
+                target_translation.save_new_version(self.request.user)
+                self.refreshed_translations.append(ctx)
             else:
                 self.failed_because_no_changes_made.append(
                     ctx.instance.best_translation.title
                 )
```
PeterNerlich:
```suggestion
    def handle_unchanged_contexts(
        self, unchanged_contexts: list[TranslationContext]
    ) -> None:
"""
        Handle content objects without changes.

        If there are no changes but the target translation is outdated, a new
        translation version is created with ``minor_edit=True`` to refresh its
        status without flagging it as a content change.
        Otherwise, the translation is not updated because no changes were detected.

        Push notifications are excluded from this processing because they do not
        use ``AbstractContentTranslation`` and therefore have no outdated status
        to refresh.
       """
        for ctx in unchanged_contexts:
            target_translation = ctx.existing_target_translation
            if (
                isinstance(target_translation, AbstractContentTranslation)
                and target_translation.is_outdated
            ):
                target_translation.save_new_version(self.request.user)
                self.refreshed_translations.append(ctx)
            else:
                self.failed_because_no_changes_made.append(
                    ctx.instance.best_translation.title
                )
```

tests/mt_api/mt_api_test.py
```diff
@@ -570,6 +571,93 @@ def test_manual_update_mt_page(
         assert en_translation.minor_edit is True
 
 
+@pytest.mark.django_db
+@pytest.mark.parametrize("login_role_user", [EDITOR], indirect=True)
```
PeterNerlich, line 575: Why only for the editor? I have not really worked with the different roles and their permissions yet, but wouldn't you at least also include root?

PeterNerlich (approved):
Thank you! I have two nitpicks, neither of which keep this from being merged as-is, if need be.

tests/mt_api/mt_api_test.py
seluianova: good question :)

I think testing the same scenarios with multiple roles is excessive. it just increases the test run time without adding much value.

I believe permissions should be tested separately: for each function, we verify which roles have access and which don’t. there’s no need to repeat the same **functional** scenarios for different roles.

e.g., if translation works for a ROOT user, it works regardless of the specific scenario, whether we’re testing “do not translate” tags, empty pages, or non-empty pages, etc.

however, it’s a question of the general testing approach, and it might be confusing to apply it to some tests but not to others 🙈

integreat_cms/core/utils/machine_translation_api_client.py
seluianova: adjusted to plural

PR #1931, opened 2022-12-01
Fix possibility to mark page as up-to-date without performing changes

timoludwig:
### Short description

Fix possibility to mark page as up-to-date without performing changes

### Proposed changes

- Fix possibility to mark page as up-to-date without performing changes
- Fix change detection in event form
- Fix change detection in location form

### Resolved issues

Fixes: #1864

__________________________________________________

[Pull Request Review Guidelines](https://digitalfabrik.github.io/integreat-cms/pull-request-review-guide.html)

PR #4273, opened 2026-05-06
Fix existing translations incorrectly marked as outdated after MT into new language

dkehne:
Fixes: #4254
Fixes: #4284 

## Summary

- Adds `minor_edit = True` in `MachineTranslationForm.save()` before calling `super().save()` when no text content (`title`/`content`) changed
- Minor-edit versions are excluded from `major_source_translation` lookups, so the outdated timestamp comparison is done against the last content-bearing source version, not the just-created empty save
- `CustomContentModelForm` is deliberately left unchanged to preserve the behaviour from #1864 (editor can mark an outdated translation as up-to-date by saving without content changes)

## Root Cause

See [comment on #4254](https://github.com/digitalfabrik/integreat-cms/issues/4254#issuecomment-4387736177) for the full analysis.

**Why not fix it in `CustomContentModelForm`?**
The condition `{"title", "content"}.isdisjoint(self.changed_data) and foreign_form_changed` was introduced intentionally to fix #1864: without it, an editor opening an outdated translation and saving without content changes would get `minor_edit=True` forced, keeping the translation stuck as outdated forever. The fix therefore targets only `MachineTranslationForm` — the code path triggered when saving a source translation purely to kick off MT for a new language.

## Test plan

- [ ] Open a page that has multiple up-to-date machine-translated languages (e.g. Spanish, Russian)
- [ ] Save the page with MT enabled for one additional new language, without changing any content
- [ ] Verify existing translations still show as up to date (robot icon), not outdated (red icon)
- [ ] Verify the new language was correctly machine-translated
- [ ] Open an outdated translation, save without changing content → should now appear up-to-date (regression test for #1864)
- [ ] Run `tools/test.sh tests/cms/models/ tests/cms/views/pages/`

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

PR #4411, opened 2026-07-06
Bugfix: Show UP_TO_DATE translation status when a translation is manually updated without changes

### LINKED ISSUE #4397: A previously machine translated page that is outdated will go back to 'machine translated' status after being Updated.
### Bug Summary

Related to https://github.com/digitalfabrik/integreat-cms/issues/4385

When updating an outdated translation that was translated with Google Translate that was previously machine translated it will get the 'machine translation' status instead of 'Up-to-date' status.

Affects:
- Right To Left Languages
- Translations translated with Google Translate
- Same behavior in Pages, Events and Locations

### Expected Behavior

The status of the translation is 'up-to-date' after being updated.

### Actual Behavior

The status of the translation is 'machine translated' after being updated.

### Steps to Reproduce

Prerequisites:
Google Translate as Translation Provider & Right To Left Language added in Language Tree

1. Go To Machine Translations -> For all languages change all translation providers to Google Translate
2. Create a new page with a German translation and click Publish.
3. Add an Arabaic / Paschto / Person (any right to left language) translation for the page via Machine Translation
4. Edit the German translation so that the translation becomes outdated.
5. Open the translation and click Update

### Reproducibility

Always (100%)

### Environment

[Version: 2026.6.3-rc](https://integreat-rc.tuerantuer.org/202662-rc/release-notes/)

What the result looks like for Right to Left languages translated with Google Translate & DeepL comparison:
Image
Image
Image

### Severity

None

### Regression Information

_No response_

### Technical Context

_No response_

_No response_

### Verification / Testing

_No response_

### Related Issues

_No response_

seluianova:
### Short description
For some historical reasons, when a user saved a translation without content changes, we saved it as a major edit.
Only if there were any other changes on the form, except for the title or content fields, did we save it as a minor edit.

After [this change](https://github.com/digitalfabrik/integreat-cms/pull/4273) was introduced, we started saving translations without content changes as minor edits. 
This introduced the following side effect:
When a user manually updated a machine-translated page without any content changes, it did not change the status from "machine translated" to "up-to-date" (check mark).
Because we defined translation state based on the last major version.

### Proposed changes
- When a translation has no content changes, always save it as a minor edit _(Previous behavior: only when translation had no content changes BUT other fields on the form were changed, it was saved as a minor edit)_
- Adjust the `translation_state` property: use the last public or draft version to determine the translation state _(Previous behavior: the last major version was used)_

### Side effects
I could not find any, but careful testing is needed.

### How to test

- Check that the reported issue is not reproduced https://github.com/digitalfabrik/integreat-cms/issues/4397
- Check that the previously fixed related issue was not broken again:
https://github.com/digitalfabrik/integreat-cms/issues/4254
https://github.com/digitalfabrik/integreat-cms/issues/4284
- Analyze whether any other cases could be affected.

### Resolved issues

Fixes: #4397

### DIFF
--- a/integreat_cms/cms/forms/custom_content_model_form.py
+++ b/integreat_cms/cms/forms/custom_content_model_form.py
@@ -119,13 +119,12 @@ class CustomContentModelForm(CustomModelForm):
         self.data["slug"] = unique_slug
         return unique_slug
 
-    def save(self, commit: bool = True, foreign_form_changed: bool = False) -> Any:
+    def save(self, commit: bool = True) -> Any:
         """
         This method extends the default ``save()``-method of the base :class:`~django.forms.ModelForm` to set attributes
         which are not directly determined by input fields.
 
         :param commit: Whether or not the changes should be written to the database
-        :param foreign_form_changed: Whether or not the foreign form of this translation form was changed
         :return: The saved content translation object
         """
 
@@ -133,8 +132,9 @@ class CustomContentModelForm(CustomModelForm):
             # Delete now outdated link objects
             self.instance.links.all().delete()
 
-        # If none of the text content fields were changed, but the foreign form was, treat as minor edit (even if checkbox isn't clicked)
-        if {"title", "content"}.isdisjoint(self.changed_data) and foreign_form_changed:
+        # If none of the text content fields were changed, treat as minor edit (even if checkbox isn't clicked)
+        # so that existing translations in other languages are not incorrectly flagged as outdated
+        if {"title", "content"}.isdisjoint(self.changed_data):
             self.logger.debug("Set 'minor_edit=True' since the content did not change")
             self.instance.minor_edit = True
 
--- a/integreat_cms/cms/forms/machine_translation_form.py
+++ b/integreat_cms/cms/forms/machine_translation_form.py
@@ -139,22 +139,14 @@ class MachineTranslationForm(CustomContentModelForm):
     def save(
         self,
         commit: bool = True,
-        foreign_form_changed: bool = False,
     ) -> EventTranslation | (PageTranslation | POITranslation):
         """
         Create machine translations and save them to the database
 
         :param commit: Whether or not the changes should be written to the database
-        :param foreign_form_changed: Whether or not the foreign form of this translation form was changed
         :return: The saved content translation object
         """
-        # If no text content changed, mark the source translation as minor edit so that
-        # existing translations in other languages are not incorrectly flagged as outdated
-        if {"title", "content"}.isdisjoint(self.changed_data):
-            self.instance.minor_edit = True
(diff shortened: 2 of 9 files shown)


integreat_cms/cms/forms/custom_content_model_form.py
```diff
@@ -134,7 +133,8 @@ def save(self, commit: bool = True, foreign_form_changed: bool = False) -> Any:
             self.instance.links.all().delete()
 
         # If none of the text content fields were changed, but the foreign form was, treat as minor edit (even if checkbox isn't clicked)
-        if {"title", "content"}.isdisjoint(self.changed_data) and foreign_form_changed:
```
seluianova, line 137: `foreign_form_changed` param was added here in 2022 to fix the bug: https://github.com/digitalfabrik/integreat-cms/pull/1931
I assume this is obsolete, the bug is not reproduced anymore.

I reverted that because the logic itself looked questionable for me. why no content changes + other form changes = minor edit, but no content changes + no form changes (just timestamp update) = major edit ?

MizukiTemma (approved):
Looks good 😄

integreat_cms/cms/forms/custom_content_model_form.py
```diff
@@ -134,7 +133,8 @@ def save(self, commit: bool = True, foreign_form_changed: bool = False) -> Any:
             self.instance.links.all().delete()
 
         # If none of the text content fields were changed, but the foreign form was, treat as minor edit (even if checkbox isn't clicked)
```
hannaseithe: This comment should be edited, since we do not use the foreign form check anymore

hannaseithe (commented):
This change looks good to me.

I would suggest to add some tests though, both for the translation_state logic and for the minor_edit flag - just to reduce the regression risk around this rather complex business logic. At the minimum I would suggest a test that checks for the "veraltete machine translated status -> no content update -> UP_TO_DATE state"

seluianova:
> At the minimum I would suggest a test that checks for the "veraltete machine translated status -> no content update -> UP_TO_DATE state"

I agree, I added exactly this test in my another PR: https://github.com/digitalfabrik/integreat-cms/pull/4399
It just needs to be updated a bit after this PR is merged (currently it checks the previous state - "veraltete machine translated status -> no content update -> up-to-date machine translated")
I will also add a check for minor_edit flag there.

**UPD**. okay it is not exactly the same, but I would prefer to continue with tests in 4399 anyway, because I also did some test refactoring there to simplify test creation. Would this be fine for you? @hannaseithe

integreat_cms/cms/forms/custom_content_model_form.py
seluianova: right, thanks.
I updated the comment and added a release note

hannaseithe (approved):

PR #4414, opened 2026-07-06
change error message for incorrect date in events

### LINKED ISSUE #4243: Incorrect error message when creating a one-time event with a start date in the past
### Describe the Bug
When creating a one-time event with a start date in the past, an incorrect validation error message is shown. The error refers to the **end date** instead of the **start date**.

### Steps to Reproduce

1. Go to 'Events'
2. Click on 'Create Event'
3. Fill the title and select `start date` on the `One-time or recurrent` tab in the past, for example:
Image

4. Select any location and click "Publish"
5. See error

### Expected Behavior
The error message should refer to the **start date** field, for example:
`start date: The event can't be in the past. Please choose today or a future date and time.`

### Actual Behavior
The error message incorrectly refers to the **end date**:
`end date: The end of the event can't be in the past. Please choose today or a future date and time.`

### Screenshot

Image

### Additional Information

```
```

</details>

dasgoutam:
### Short description

Fixes error message when creating a one-time event with a start date in the past

### Proposed changes

- The error message should refer to the 'start_date'. Currently it is referring to the 'end_date'

### Side effects

- None

### How to test

- Go to 'Events'
- Click on 'Create Event'
- Fill the title and select start date on the One-time or recurrent tab in the past.
- Select any location and click "Publish"
- Check Error message. It should be - `start date: The event can't be in the past. Please choose today or a future date and time`

### Resolved issues

Fixes: #4243

### DIFF
--- a/integreat_cms/cms/forms/events/event_form.py
+++ b/integreat_cms/cms/forms/events/event_form.py
@@ -235,10 +235,10 @@ class EventForm(CustomModelForm):
                 )
             ):
                 self.add_error(
-                    "end_date",
+                    "start_date",
                     forms.ValidationError(
                         _(
-                            "The end of the event can't be in the past. Please choose today or a future date and time."
+                            "The event can't be in the past. Please choose today or a future date and time."
                         ),
                         code="invalid",
                     ),
--- a/integreat_cms/cms/forms/events/event_form.py
+++ b/integreat_cms/cms/forms/events/event_form.py
@@ -227,22 +227,34 @@ class EventForm(CustomModelForm):
                     ),
                 )
             today = timezone.now()
-            if not cleaned_data.get("is_recurring") and (
-                end_date < today.date()
-                or (
+            if not cleaned_data.get("is_recurring"):
+                if end_date < today.date():
+                    self.add_error(
+                        (
+                            "end_date"
+                            if cleaned_data.get("is_long_term")
+                            else "start_date"
+                        ),
+                        forms.ValidationError(
+                            _(
+                                "The event can't be in the past. Please choose today or a future date and time."
+                            ),
+                            code="invalid",
+                        ),
+                    )
+                elif (
                     end_date == today.date()
                     and cleaned_data.get("end_time") < today.time()
-                )
-            ):
-                self.add_error(
-                    "start_date",
-                    forms.ValidationError(
-                        _(
-                            "The event can't be in the past. Please choose today or a future date and time."
+                ):
+                    self.add_error(
+                        "end_time",
+                        forms.ValidationError(
+                            _(
+                                "The event can't be in the past. Please choose today or a future date and time."
+                            ),
+                            code="invalid",
                         ),
-                        code="invalid",
-                    ),
-                )
+                    )
 
         # If everything looks good until now, combine the dates and times into timezone-aware datetimes
         if not self.errors:
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -2015,11 +2015,11 @@ msgstr "Das Ende der Veranstaltung kann nicht vor ihrem Beginn sein"
 
 #: cms/forms/events/event_form.py
 msgid ""
-"The end of the event can't be in the past. Please choose today or a future "
-"date and time."
+"The event can't be in the past. Please choose today or a future date and "
+"time."
 msgstr ""
-"Das Ende der Veranstaltung darf nicht in der Vergangenheit liegen. Bitte "
-"wählen Sie heute oder ein zukünftiges Datum und Uhrzeit."
+"Die Veranstaltung darf nicht in der Vergangenheit liegen. Bitte wählen Sie "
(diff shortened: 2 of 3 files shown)


MizukiTemma:
Thank you for the PR 😸 It works as expected 👍 

I have one idea which is outside of the issue scope but good to solve together.

The corrected message "The event can't be in the past. Please choose today or a future date and time." is shown in two cases in the current implementation of the PR: 1) when the end **date** is before today and 2) the end date is equal today and the end **time** is before the current time. In both cases the **date field** is marked red.

It's toally correct for the case 1 but it may be better to highlight the end time field for the case 2 🤔 


The screenshot below is showing an example of case 2, creating an event at the time of 08:30 on 07.07.2026
image


@m0ndra 

I would like to hear your opinion here 👂

m0ndra:
So as far as I understood the cases:

1. We have a single-day event, so the **start-date is the same as the end date**. If this selected single-day date is in the past, it should be highlighted in red for the user.
2. We have a single-day event: the date is **today**, but the **end-time is already in the past**. In this case, **only the end-time form should be highlighted** in red for the user (since the date is correct).
3. We have a more-days-event and the end-date is in the past: Same behaviour as in case 1: the **date should be highlighted** in red to the user
4. We have a more-days-event, the end-date is today but the end-time is already in the past: same behaviour as in case 2: **the end-date form should be highlighted** in red for the user.

So summary: **Only highlight the forms that are actually WRONG and need to be fixed by the user.** The error message in all cases can be: "**The event can't be in the past. Please choose today or a future date and time.**"

@MizukiTemma can we improve it within this issue?

MizukiTemma:
> So as far as I understood the cases:
> 
> 1. We have a single-day event, so the **start-date is the same as the end date**. If this selected single-day date is in the past, it should be highlighted in red for the user.
> 2. We have a single-day event: the date is **today**, but the **end-time is already in the past**. In this case, **only the end-time form should be highlighted** in red for the user (since the date is correct).
> 3. We have a more-days-event and the end-date is in the past: Same behaviour as in case 1: the **date should be highlighted** in red to the user
> 4. We have a more-days-event, the end-date is today but the end-time is already in the past: same behaviour as in case 2: **the end-date form should be highlighted** in red for the user.
> 
> So summary: **Only highlight the forms that are actually WRONG and need to be fixed by the user.** The error message in all cases can be: "**The event can't be in the past. Please choose today or a future date and time.**"
> 
> @MizukiTemma can we improve it within this issue?

@dasgoutam 
Would it be fine for you? Sorry that I've opened a discussion widening the issue scope 🙈

m0ndra:
I mean, I would open a bug otherwise since it is not correct system behaviour at the moment ☺️ So I thought it made sense to fix it within this issue since we are looking at the code right now anyway. @MizukiTemma I think it is great to always try to "leave the place better as we found it" so I personally like that you opened the discussion. I hope it will be OK for @dasgoutam as well 🙂

integreat_cms/cms/forms/events/event_form.py
```diff
@@ -235,10 +235,10 @@ def clean(self) -> dict[str, Any]:
                 )
             ):
                 self.add_error(
-                    "end_date",
+                    "start_date",
```
seluianova: This if-condition only checks `end_date` here, so I don't think it is correct to simply change `end_date` to `start_date` in the error message, because the `start_date` is not actually checked here.

This would only work for single-day events, where `start_date` = `end_date`.

Otherwise if `end_date` is invalid but `start_date` is valid, the message will be incorrect.

dasgoutam:
> I mean, I would open a bug otherwise since it is not correct system behaviour at the moment ☺️ So I thought it made sense to fix it within this issue since we are looking at the code right now anyway. @MizukiTemma I think it is great to always try to "leave the place better as we found it" so I personally like that you opened the discussion. I hope it will be OK for @dasgoutam as well 🙂

@MizukiTemma @m0ndra Thank you for the discussion and going through the different cases, and I agree with them :) I think it makes sense to fix it within this issue. I can update the description after the fix. Thanks!

integreat_cms/cms/forms/events/event_form.py
dasgoutam: @seluianova Thanks for the review! You're right, this behaviour is wrong. I think I didn't think through all the cases clearly here. I'll do that and get back with a correction

dasgoutam: @seluianova I've pushed a new commit. For single-day events, where `start_date` = `end_date`, the error message would reference the 'start_date'. But for long-term events, if the `end_date` is invalid and the `start_date` is valid, the message would reference the `end_date`.

Let me know if you think this covers all the cases :)

dasgoutam:
@MizukiTemma You can test the new fix now which should be able differentiate the red highlight depending on whether the `end_time` or `end_date` is in the past.

MizukiTemma (commented):
Works as expected 😸 Thank you for updates 👍 

💡 You can add a release note 🗒️

MizukiTemma (approved):

seluianova (approved):
looks good to me, thanks for the adjustments 👍

PR #4415, opened 2026-07-07
prefetch public_or_draft_translations without content

jonbulz:
### Short description

hotfix: prevent per-page DB queries loading full page content in page-tree rendering

### Proposed changes

the latest release changed the translation-status calculation (translation_state) to also look up the newest draft-or-public version of each translation . the page-tree renderer (cache_tree_dict) doesn't prefetch this set of translations, so every page-tree render (pages list, statistics page tree) falls back to one extra DB query per page, each loading full translations.
this fix adds the missing prefetch in cache_tree_dict (integreat_cms/cms/models/pages/page.py): the latest draft/public translation per page and language is now fetched in one query up front, with the content field deferred since it's not needed.

### Side effects

- 

### Resolved issues

Fixes: #

### DIFF
--- a/integreat_cms/cms/models/pages/page.py
+++ b/integreat_cms/cms/models/pages/page.py
@@ -12,6 +12,7 @@ from django.utils.translation import gettext_lazy as _
 from linkcheck.models import Link
 from treebeard.ns_tree import NS_NodeQuerySet
 
+from ...constants import status
 from ...utils.translation_utils import gettext_many_lazy as __
 from ..abstract_content_model import ContentQuerySet
 from ..abstract_tree_node import AbstractTreeNode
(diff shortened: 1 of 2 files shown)


MizukiTemma:
@jonbulz @svenseeberg 
I think we can merge this as hot fix directly into `main` (= the prod system)?

jonbulz:
unfortunately messes up the query count for the tests. I'll look into it. would not like to force merge a broken pipeline into main

jarlhengstmengel:
Messing around with the page tree and creating pages with a database dump, I managed to crash CMS locally on the develop branch. So in theory it could be tested locally.

jonbulz:
the increase in query counts are expected, so I adjusted the expected results. tests are green now and I changed the base branch to `main` so this can land before the next deployment reverts the changes on the server

MizukiTemma:
@jonbulz 

It seems this PR got to contain changes from other PRs that are diff between `main` and `develop`. Could you clean up?

jonbulz:
Whoops, my bad 😅 Guess that happens when you change the base branch...
Unfortunately, the status quo in `api_config.py` is different between `main` and `develop`, so this will probably result in merge confilcts when we merge `main` back into `develop` after this fix landed. I guess I'll have to resolve that manually then. I still think it's better that this state is self-contained with a green test suite

svenseeberg:
> I think we can merge this as hot fix directly into `main` (= the prod system)?

Currently, this is not strictly necessary, as the prod system is already patched manually. We need to include this PR in the next release though.

jonbulz:
Ah ok, I thought a highstate would revert the changes, but I suppose it only does so when we`force-reinstall`. In that case, I'd suggest we got through the usual workflow. I'll close this PR and make a new one for develop

PR #4422, opened 2026-07-08
prefetch public_or_draft_translations without content

jonbulz:
### Short description

hotfix: prevent per-page DB queries loading full page content in page-tree rendering

duplicate of #4415 , but into `develop` and thus with different expected query counts in `api_config.py`

### Proposed changes

the latest release changed the translation-status calculation (translation_state) to also look up the newest draft-or-public version of each translation . the page-tree renderer (cache_tree_dict) doesn't prefetch this set of translations, so every page-tree render (pages list, statistics page tree) falls back to one extra DB query per page, each loading full translations.
this fix adds the missing prefetch in cache_tree_dict (integreat_cms/cms/models/pages/page.py): the latest draft/public translation per page and language is now fetched in one query up front, with the content field deferred since it's not needed.

### Side effects

- 

### Resolved issues

Fixes: #

### DIFF
--- a/integreat_cms/cms/models/pages/page.py
+++ b/integreat_cms/cms/models/pages/page.py
@@ -12,6 +12,7 @@ from django.utils.translation import gettext_lazy as _
 from linkcheck.models import Link
 from treebeard.ns_tree import NS_NodeQuerySet
 
+from ...constants import status
 from ...utils.translation_utils import gettext_many_lazy as __
 from ..abstract_content_model import ContentQuerySet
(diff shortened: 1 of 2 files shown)


jarlhengstmengel (approved):
Thank you for the speedy fix! Besides the real world validation we already have, it also works with a database dumb locally :)

MizukiTemma (approved):
Thank you so much for investigating and fixing it 🚀 🙏

PR #4425, opened 2026-07-10
Improve PDF performance

svenseeberg:
* Make check for existing cached PDF faster.
* Execute compute heavy queries only when no cached PDF exists.
* Generate PDF file in temp dir and move to download dir when done to mitigate race condition between 2 processes.

### DIFF
--- a/integreat_cms/cms/utils/pdf_utils.py
+++ b/integreat_cms/cms/utils/pdf_utils.py
@@ -1,16 +1,16 @@
 from __future__ import annotations
 
+import contextlib
 import hashlib
 import logging
 import os
+import tempfile
 from typing import TYPE_CHECKING
 from urllib.parse import unquote, urlparse
 
 from django.conf import settings
 from django.contrib.staticfiles import finders
-from django.core.files.base import ContentFile
 from django.core.files.storage import FileSystemStorage
-from django.db.models import Min
 from django.http import HttpResponse
 from django.shortcuts import redirect
 from django.template.loader import get_template
@@ -18,8 +18,8 @@ from django.utils.translation import gettext_lazy as _
 from xhtml2pdf import pisa
 from xhtml2pdf.default import DEFAULT_CSS
 
-from ..constants import text_directions
-from ..models import Language, Page
+from ..constants import status, text_directions
+from ..models import Language, Page, PageTranslation
 from .text_utils import truncate_bytewise
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


integreat_cms/cms/utils/pdf_utils.py
```diff
@@ -48,10 +48,17 @@ def generate_pdf(
     :param pages: at least on page to render as PDF document
     :return: Redirection to PDF document
     """
+    # Build a lightweight queryset with the (large) translation content deferred, so that the
+    # cache key computation and the existence check below do not load the page contents. The
+    # full content is only fetched further down if the PDF actually has to be rendered.
+    pages = pages.prefetch_related(None)
```
timobrembeck: This makes `GeneratePdfView.prefetch_public_translations = True` in `cms/views/pages/page_bulk_actions.py:50` dead config — the prefetch lookup is stripped here before the queryset is ever evaluated, so it never executes. If you want to, you could remove that flag in this PR so it doesn't mislead future readers into thinking the bulk view still prefetches translations.

integreat_cms/cms/utils/pdf_utils.py
```diff
@@ -48,10 +48,17 @@ def generate_pdf(
...
+
     # first all necessary data for hashing are collected, starting at region slug
     # region last_updated field taking into account, to keep track of maybe edited region icons
     pdf_key_list = [region.slug, region.last_updated]
-    for page in pages:
+    excluded_page_ids = []
+    for page in hash_pages:
         # add translation id and last_updated to hash key list if they exist
         page_translation = page.get_public_translation(language_slug)
         if page_translation and not page.archived:
```
timobrembeck: Non-blocking, pre-existing: `page.archived` resolves `implicitly_archived` → `get_cached_ancestors()`, which runs one `get_ancestors()` query per page since nothing warms `_cached_ancestors` here. So the cache-hit path is still O(N) queries even with the content deferred. Also, for the whole-region branch the check is redundant — `region.get_pages()` already returns `non_archived_pages`. Might be worth a follow-up issue rather than this PR.

timobrembeck (approved):
Looks good, thanks a lot! 🚀 

A minor suggestion: the existing tests guard correctness but not the performance goal. A small test using pytest-django's `django_assert_num_queries` (or asserting the hash-path SQL does not select `content`) would protect it. Optional, but cheap insurance.

integreat_cms/cms/models/abstract_content_model.py
```diff
@@ -77,6 +85,22 @@ def prefetch_public_translations(
...
+    ) -> ContentQuerySet:
+        """
+        Like :meth:`prefetch_public_translations`, but defers the (potentially large)
+        translation ``content`` field. Useful when only translation metadata (e.g. ``id`` and
+        ``last_updated`` for a cache key) is required and the content is not accessed.
+
+        :return: The queryset of content objects
+        """
+        return self.prefetch_translations(
+            to_attr="prefetched_public_translations",
```
jonbulz: I'd prefer to use a different attribute name to avoid confusion. Something like `prefetched_public_translations_without_content` maybe? I won't die on that hill, though.

jonbulz (commented):
Looks nice, thanks! I think once @timobrembeck's comments are addressed this is ready to go. 👍

integreat_cms/cms/utils/pdf_utils.py
svenseeberg: Thanks, implemented now :+1:

integreat_cms/cms/utils/pdf_utils.py
svenseeberg: Thanks, implemented now 👍

integreat_cms/cms/models/abstract_content_model.py
svenseeberg: Thanks, implemented now 👍

svenseeberg:
Actually, the approach does not really work. Just deferring the content field does not really yield that much of an improvement. I'll try to improve the hash loop.

svenseeberg:
image

andrew8er:
@jonbulz Can this be merged now? Looks like all your remarks are addressed.

integreat_cms/cms/utils/pdf_utils.py
```diff
@@ -33,6 +33,73 @@
...
+            "title",
+            "id",
+            "last_updated",
+        )
+        .distinct("page__id")
+    )
+    rows = list(rows)
+    rows.sort(key=lambda r: (r[4] or 0, r[2] or 0))
+
+    explicit = [(r[2], r[3]) for r in rows if r[5]]
```
andrew8er, line 83: 💭 Could be named `explicitly_archived` for clarity.

integreat_cms/cms/utils/pdf_utils.py
```diff
@@ -33,6 +33,73 @@
...
+            "page__tree_id",
+            "page__explicitly_archived",
+            "title",
+            "id",
+            "last_updated",
+        )
+        .distinct("page__id")
+    )
+    rows = list(rows)
+    rows.sort(key=lambda r: (r[4] or 0, r[2] or 0))
```
andrew8er, line 81: This version avoids a rebind of the variable.

```suggestion
    page_translation_rows = list(
        PageTranslation.objects.filter(
            language__slug=language_slug,
            status=status.PUBLIC,
            page__in=pages,
        )
        .order_by("page__id", "-version")
        .values_list(
            "page__id",
            "page__depth",
            "page__lft",
            "page__rgt",
            "page__tree_id",
            "page__explicitly_archived",
            "title",
            "id",
            "last_updated",
        )
        .distinct("page__id")
    )
    page_translation_rows.sort(key=lambda r: (r[4] or 0, r[2] or 0))
```

integreat_cms/cms/utils/pdf_utils.py
andrew8er, line 107: ❌ Unrelated to this PR, but the return code must be `HttpResponseRedirect | HttpResponse`.

andrew8er, line 149: 💭 Unrelated to this PR, but this calculation is valid for the duration of the process (assuming no mounts happen on the storage), so it could be lifted into the module's top level.

andrew8er, line 204: ❌ Unrelated to this PR, but since this function is internal to the module, it should be called `_link_callback`.

integreat_cms/cms/utils/pdf_utils.py
```diff
@@ -109,32 +164,40 @@ def generate_pdf(
...
-                "The following PDF could not be rendered: %r, %r, %r",
-                region,
-                language,
-                pages,
-            )
-            return HttpResponse(
-                _("The PDF could not be successfully generated."),
-                status=500,
-            )
+        final_path = pdf_storage.path(filename)
```
andrew8er, line 171: 💭 Could be named `local_path` to better reflect its purpose and match the wording of `Storage.path()`.

integreat_cms/cms/utils/pdf_utils.py
andrew8er, line 151: 💭 Could be named `public_path` to better reflect its purpose.

integreat_cms/cms/utils/pdf_utils.py
```diff
@@ -33,6 +33,73 @@
 pdf_storage = FileSystemStorage(location=settings.PDF_ROOT, base_url=settings.PDF_URL)
```
andrew8er, line 33: 💭 Unrelated to this PR, but since this is an internal variable, it should be named `_pdf_storage`.

andrew8er (commented):
A few suggestions.

PR #4428, opened 2026-07-13
Keep filter state after copy and delete action in Events

### LINKED ISSUE #2396: Keep filter status after copying events
### Motivation

Currently the filter will be reset after copying events.

For example, filter for past events, past event are shown, do copying, then only future events will be shown after copying.
In this case only past events should be shown after copy operation as before that.

### Proposed Solution

Keep the filter status/result after copying events.

### Alternatives

Leave it as it is.

### Additional Context

The idea of keeping filter status/result can be applied for not only event copying but also other actions.

dasgoutam:
### Short description

Preserve the filter state in "Events" after `Copy` or `Delete` of an event, so that the user returns to the same filtered view before after the action.

### Proposed changes

- Pass the current query string as a hidden `query_string` in the form. There is a separate form for copy(`event_list_row.py:159`) and delete(`generic_confirmation_dialog.html`)
- The copy and delete views are updated in `event_actions.py` to add it to the redirect URL. This should preserve the filter state on redirect.

### Side effects

- None

There are no intended deviations from the issue and design. It was decided to add "Delete" in addition to "Copy" as actions which preserve the filter state.

### How to test

- Open 'Events' for any region and language
- Apply any filter (e.g. select "Past events" in the time range filter and submit)
- Click "Copy" or "Delete" on any event in the list
- Verify that after the copy, the events list still shows only past events or any other filter you added (i.e. filter is preserved)
- Also verify that with no filters active, copy and delete still redirect to the plain unfiltered events list

### Resolved issues

Fixes: #2396 

### DIFF
--- a/integreat_cms/cms/templates/events/event_list_row.html
+++ b/integreat_cms/cms/templates/events/event_list_row.html
@@ -156,10 +156,9 @@
                     <i icon-name="pencil"></i>
                 </a>
                 {% if not event.external_calendar %}
-                    <form action="{% url 'copy_event' event_id=event.id region_slug=request.region.slug language_slug=language.slug %}"
+                    <form action="{% url 'copy_event' event_id=event.id region_slug=request.region.slug language_slug=language.slug %}?{{ request.GET.urlencode }}"
                           method="post">
                         {% csrf_token %}
-                        <input type="hidden" name="query_string" value="{{ request.GET.urlencode }}">
                         <button class="btn-icon" title="{% translate "Copy event" %}">
                             <i icon-name="copy"></i>
                         </button>
@@ -195,7 +194,7 @@
                     data-confirmation-title="{{ delete_dialog_title }}"
                     data-confirmation-text="{{ delete_dialog_text }}"
                     data-confirmation-subject="{{ event_translation.title }}"
-                    data-action="{% url 'delete_event' event_id=event.id region_slug=request.region.slug language_slug=language.slug %}">
+                    data-action="{% url 'delete_event' event_id=event.id region_slug=request.region.slug language_slug=language.slug %}?{{ request.GET.urlencode }}">
                 <i icon-name="trash-2"></i>
             </button>
         {% endif %}
--- a/integreat_cms/cms/templates/events/event_list_row.html
+++ b/integreat_cms/cms/templates/events/event_list_row.html
@@ -159,6 +159,7 @@
                     <form action="{% url 'copy_event' event_id=event.id region_slug=request.region.slug language_slug=language.slug %}"
                           method="post">
                         {% csrf_token %}
+                        <input type="hidden" name="query_string" value="{{ request.GET.urlencode }}">
                         <button class="btn-icon" title="{% translate "Copy event" %}">
                             <i icon-name="copy"></i>
                         </button>
--- a/integreat_cms/cms/templates/generic_confirmation_dialog.html
+++ b/integreat_cms/cms/templates/generic_confirmation_dialog.html
@@ -25,7 +25,6 @@
             </p>
             <form method="post" action="#">
                 {% csrf_token %}
-                <input type="hidden" name="query_string" value="{{ request.GET.urlencode }}">
                 <div class="flex justify-end gap-2">
                     <button id="close-confirmation-popup" class="btn btn-ghost">
                         {% translate "Cancel" %}
--- a/integreat_cms/cms/templates/generic_confirmation_dialog.html
+++ b/integreat_cms/cms/templates/generic_confirmation_dialog.html
@@ -25,6 +25,7 @@
             </p>
             <form method="post" action="#">
                 {% csrf_token %}
+                <input type="hidden" name="query_string" value="{{ request.GET.urlencode }}">
                 <div class="flex justify-end gap-2">
                     <button id="close-confirmation-popup" class="btn btn-ghost">
                         {% translate "Cancel" %}
--- a/integreat_cms/cms/utils/event_utils.py
+++ b/integreat_cms/cms/utils/event_utils.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 from typing import TYPE_CHECKING
-from urllib.parse import parse_qsl, urlencode
 
 from django.shortcuts import reverse
 
@@ -24,6 +23,6 @@ def get_filtered_events_url(
     event_url = reverse(
         "events", kwargs={"region_slug": region_slug, "language_slug": language_slug}
     )
-    if params := parse_qsl(request.POST.get("query_string", "")):
-        event_url = f"{event_url}?{urlencode(params)}"
+    if query_string := request.GET.urlencode():
+        event_url = f"{event_url}?{query_string}"
     return event_url
--- a/integreat_cms/cms/utils/event_utils.py
+++ b/integreat_cms/cms/utils/event_utils.py
@@ -1,0 +1,29 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+from urllib.parse import parse_qsl, urlencode
+
+from django.shortcuts import reverse
+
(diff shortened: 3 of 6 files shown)


integreat_cms/cms/views/events/event_actions.py
```diff
@@ -79,25 +79,26 @@ def copy(
     region = request.region
     event = get_object_or_404(region.events, id=event_id)
 
+    query_string = request.POST.get("query_string", "")
+    events_url = reverse(
+        "events", kwargs={"region_slug": region_slug, "language_slug": language_slug}
+    )
+    if query_string:
+        events_url = f"{events_url}?{query_string}"
```
timobrembeck: If we re-encode the parameters instead of trusting the raw string, CodeQL should be satisfied and we're guaranteed a same-site redirect:
```suggestion
    events_url = reverse(
        "events", kwargs={"region_slug": region_slug, "language_slug": language_slug}
    )
    if params := parse_qsl(request.POST.get("query_string", "")):
        events_url = f"{events_url}?{urlencode(params)}"
```

integreat_cms/cms/views/events/event_actions.py
```diff
@@ -10,7 +10,7 @@
```
timobrembeck, line 10:
```suggestion
from urllib.parse import parse_qsl, urlencode

```

integreat_cms/cms/views/events/event_actions.py
```diff
@@ -159,13 +160,13 @@ def delete(
...
-        **{
-            "region_slug": region_slug,
-            "language_slug": language_slug,
-        },
+    query_string = request.POST.get("query_string", "")
+    events_url = reverse(
+        "events", kwargs={"region_slug": region_slug, "language_slug": language_slug}
     )
+    if query_string:
+        events_url = f"{events_url}?{query_string}"
```
timobrembeck: The same block appears here; can't we extract a small helper (e.g. `get_filtered_events_url(request, region_slug, language_slug)`) instead of duplicating it?

timobrembeck (changes requested):
Thanks, nice quality-of-life improvement! :+1:

Could you address the CodeQL warnings before we merge this? Even though CSRF protection makes this hard to exploit, we shouldn't put unvalidated POST data into a redirect.

Also, if you want to, could you extend one of the existing copy/delete tests to assert that the redirect keeps the query string? Then the behaviour doesn't regress silently.

integreat_cms/cms/views/events/event_actions.py
dasgoutam: @timobrembeck Thanks for the suggestions! I have applied them and the CodeQL warnings seem good now :)

integreat_cms/release_notes/current/unreleased/2396.yml
```diff
@@ -0,0 +1,2 @@
+en: Keep filter status after copy/delete action of events
+de: Filtermodus nach dem Kopieren oder Löschen von Ereignissen beibehalten
```
timobrembeck:
```suggestion
en: Keep filter state after copying or deleting events
de: Filter nach dem Kopieren oder Löschen von Veranstaltungen beibehalten
```
Minor detail: I think the release note could be rephrased.

timobrembeck (approved):
Thanks for your changes, looks good to me! :tada:

I'd reword the release note, but that's just a minor detail.

integreat_cms/cms/templates/events/event_list_row.html
```diff
@@ -159,6 +159,7 @@
                     <form action="{% url 'copy_event' event_id=event.id region_slug=request.region.slug language_slug=language.slug %}"
                           method="post">
                         {% csrf_token %}
+                        <input type="hidden" name="query_string" value="{{ request.GET.urlencode }}">
```
hannaseithe:
```suggestion
<form action="{% url 'copy_event' event_id=event.id region_slug=request.region.slug language_slug=language.slug %}?{{ request.GET.urlencode }}"
method="post">
{% csrf_token %}
```

integreat_cms/cms/templates/generic_confirmation_dialog.html
```diff
@@ -20,6 +20,7 @@ <h3 id="confirmation-subject" class="text-center mt-4 mb-6 break-words">
             </p>
             <form method="post" action="#">
                 {% csrf_token %}
+                <input type="hidden" name="query_string" value="{{ request.GET.urlencode }}">
```
hannaseithe:
```suggestion
```

integreat_cms/cms/utils/event_utils.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+from urllib.parse import parse_qsl, urlencode
```
hannaseithe:
```suggestion
```

integreat_cms/cms/utils/event_utils.py
```diff
@@ -0,0 +1,29 @@
...
+    :param region_slug: Slug of the region
+    :param language_slug: Current GUI language slug
+
+    :return: The events list URL with optional filter query string
+    """
+    event_url = reverse(
+        "events", kwargs={"region_slug": region_slug, "language_slug": language_slug}
+    )
+    if params := parse_qsl(request.POST.get("query_string", "")):
+        event_url = f"{event_url}?{urlencode(params)}"
```
hannaseithe:
```suggestion
    if query_string := request.GET.urlencode():
        event_url = f"{event_url}?{query_string}"
```

integreat_cms/cms/templates/events/event_list_row.html
hannaseithe:
```suggestion
                    data-action="{% url 'delete_event' event_id=event.id region_slug=request.region.slug language_slug=language.slug %}?{{ request.GET.urlencode }}">
```

hannaseithe (approved):
Hi Goutam, this is working well. So I approve in advance: If you want to keep it as is, I am ok with it. 

But I still want to suggest mychanges below, as  I feel this might be a bit cleaner.  Since I got stuck on this and invested all the time figuring out a different way to do this, I might as well share it with you. The advantage of my approach is that it actually simply uses the url parameters of the request to transmit the filter arguments AND (more importantly) it saves us from sending the query_string from anywhere else but the delete confirmation dialog of the event_list.html view. 

But I leave it up to you :) Your version is fine too

dasgoutam:
@hannaseithe Thanks for looking so thoroughly into this and finding an alternate solution! I was hesitant to change the code initially, but I agree now that your version is a much cleaner change and I think we would much rather have that. There was one change missing from your suggestions. The search params would also need to go through `confirmationPopup` for the `Delete` action to work. However this change introduces a new CodeQL warning. I'm not sure yet if the warning is a valid problem.

hannaseithe:
> However this change introduces a new CodeQL warning. I'm not sure yet if the warning is a valid problem.

I had the same security warning here and decided that it was sufficient: https://github.com/digitalfabrik/integreat-cms/pull/4356/changes#diff-a72678609cdd134e587b7c594d2562085024fcabba5610c0f3f40ebc61ec821fR68 (Line 68 in `bulk-action.ts`)

The `new URL(...)` followed by `url.origin ==window.location.origin` is what does the sanitization.

dasgoutam:
> > However this change introduces a new CodeQL warning. I'm not sure yet if the warning is a valid problem.
> 
> I had the same security warning here and decided that it was sufficient: https://github.com/digitalfabrik/integreat-cms/pull/4356/changes#diff-a72678609cdd134e587b7c594d2562085024fcabba5610c0f3f40ebc61ec821fR68 (Line 68 in `bulk-action.ts`)
> 
> The `new URL(...)` followed by `url.origin ==window.location.origin` is what does the sanitization.

Thanks for the info 👍 Then I would also dismiss this alert

Issue #4403, opened 2026-07-01
[Story]: Add confirmation pop up for contact bulk action


MizukiTemma:
### User Story

As CMS user I want to be once warned before deleting/restoring/archiving contacts by bulk action.

### Problem Description

Users can bulk delete/archive/restore contacts without confirming anything. It can lead to accidental status change or even deletion and/or data lost due to irriversibility.

### Acceptance Criteria

- [ ] Show a confirmation pop up before processing bulk deletion/archiving/resotoring of contacts, like it does for bulk page deletion

### User Flow

1. User selects at least one contact object
2. User select deletion in the bulk action menü and click "execute"
3. User sees an warning popup and select either "Cancel" or "Confirm"
4-1. Come back to the contact list, if canceleld. 
4-2. The selected contacts are deleted, if confirmed

### Additional Context

https://chat.tuerantuer.org/digitalfabrik/pl/p9w6m9fam3bz8j39duukpg7nue

### Definition of Ready

- [ ] Acceptance criteria are clear and testable
- [ ] UI design / mockups are available (if applicable)
- [ ] Dependencies are identified and unblocked
- [ ] Story has been refined / estimated by the team
- [ ] Edge cases and error states have been considered

PR #4436, opened 2026-07-15
Allow users to archive places if all referencing events and contacts are archived

### LINKED ISSUE #4154: Not possible to archive POI unless the contact, that belongs to it, is deleted
### Describe the Bug
It is not possible to archive POI unless contact that belong is deleted, but municipalities are not allowed to delete anything, only to archive. 

### Steps to Reproduce

1. Go to POI -> e.g. https://admin.integreat-app.de/testumgebung/pois/de/13328/edit/
2. Archive the contact that belongs to the POI
3. It is still not possible to archive the POI "Der Ort kann nicht archiviert werden, da er von einem Kontakt verwendet wird. Löschen Sie zuerst diesen Kontakt, um den Ort zu archivieren"

Example: https://admin.integreat-app.de/testumgebung/pois/de/9382/edit/

### Expected Behavior
it should be possible to archive the POI after the contact/contacts that belong to it is/are archived

### Actual Behavior
it is possible to archive the ort, only after the contact is deleted but it is not possible for the municipalities to delete content

### Screenshot

Image

### Additional Information

```
```

</details>

MizukiTemma:
### Short description

This PR makes it possible for users to archive a place (location/POI) if all events and contacts that are referencing it are archived (without deletion).

### Proposed changes

- Change the condition for archiving

### Side effects

- Should be none

### How to test

- Create a place
- See it can be archived
- Create or select an event or contact, make it reference the newly created place
- See the place cannot be archived
- Archive all referencing contacts and events
- See the place can be archived
 
### Resolved issues

Fixes: #4154 

### DIFF
--- a/integreat_cms/cms/models/events/event.py
+++ b/integreat_cms/cms/models/events/event.py
@@ -62,6 +62,14 @@ class EventQuerySet(ContentQuerySet):
             ),
         )
 
+    def filter_upcoming_non_archived(self) -> Self:
+        """
+        Filter all upcoming events that are not archived
+
+        :return: The Queryset of events that are upcoming and not archived
+        """
+        return self.filter_upcoming().exclude(archived=True)
+
     def filter_completed(self, to_date: date | None = None) -> Self:
         """
         Filter all events that are not ongoing and don't have any occurrences in the future. This is, per definition, if
@@ -294,6 +302,9 @@ class Event(AbstractContentModel):
         for translation in self.translations.distinct("event__pk", "language__pk"):
             # The post_save signal will create link objects from the content
             translation.save(update_timestamp=False)
+
+        if self.location and self.location.archived:
+            self.location.restore()
 
     def delete(self, *args: Any, **kwargs: Any) -> tuple[int, dict[str, int]]:
         """
--- a/integreat_cms/cms/models/pois/poi.py
+++ b/integreat_cms/cms/models/pois/poi.py
@@ -11,6 +11,7 @@ from integreat_cms.cms.models.utils import get_default_opening_hours
 
 if TYPE_CHECKING:
     from django.db.models.base import ModelBase
+    from django.db.models.query import QuerySet
 
 
 from django.utils.translation import gettext_lazy as _
@@ -146,7 +147,7 @@ class POI(AbstractContentModel):
         Archives the poi and removes all links of this poi from the linkchecker
         """
         was_successful = False
-        if not self.is_currently_used:
+        if self.can_be_archived:
             self.archived = True
             self.save()
             # Delete related link objects as they are no longer required
@@ -173,12 +174,21 @@ class POI(AbstractContentModel):
     @property
     def referring_contacts_exist(self) -> bool:
         """
-        return: whether there are contact objects referring the POI.
-        If the contact module is disabled in the region, it's always False
+        :return: whether there are contact objects referring the POI.
+            If the contact module is disabled in the region, it's always False
         """
         if not self.region.contacts_enabled:
             return False
         return self.contacts.exists()
+
+    @property
+    def non_archived_contacts(self) -> QuerySet:
+        """
+        :return: contacts that are not archived and referencing the POI
+        """
+        if not self.region.contacts_enabled:
+            return self.contacts.none()
+        return self.contacts.exclude(archived=True)
 
     @property
     def is_used(self) -> bool:
@@ -188,13 +198,14 @@ class POI(AbstractContentModel):
         return self.events.exists() or self.referring_contacts_exist
 
     @property
-    def is_currently_used(self) -> bool:
+    def can_be_archived(self) -> bool:
         """
-        :return: whether this poi is used by a contact or an upcoming event
+        :return: whether this poi can be archived
         """
-        upcoming_events = self.events.filter_upcoming()
-
-        return self.referring_contacts_exist or upcoming_events.exists()
+        return (
+            not self.non_archived_contacts.exists()
+            and not self.events.filter_upcoming_non_archived().exists()
+        )
 
     @cached_property
     def short_address(self) -> str:
--- a/integreat_cms/cms/templates/contacts/contact_form.html
+++ b/integreat_cms/cms/templates/contacts/contact_form.html
@@ -165,6 +165,7 @@
                                     data-confirmation-title="{{ restore_dialog_title }}"
                                     data-confirmation-text="{{ restore_dialog_text }}"
                                     data-confirmation-subject="{{ contact_form.instance.name }}"
+                                    {% if contact_form.instance.location.archived %} data-confirmation-referenced-location-title="{{ referenced_location_title }}" data-confirmation-referenced-location-subject="{{ contact_form.instance.location }}" {% endif %}
                                     data-action="{% url 'restore_contact' contact_id=contact_form.instance.id region_slug=request.region.slug %}">
                                 <i icon-name="refresh-ccw" class="mr-2"></i> {% translate "Restore this contact" %}
                             </button>
--- a/integreat_cms/cms/templates/pois/poi_form_sidebar/action_box.html
+++ b/integreat_cms/cms/templates/pois/poi_form_sidebar/action_box.html
@@ -28,22 +28,22 @@
             <label class="mt-0">
                 {% translate "Archive location" %}
             </label>
-            {% if poi_form.instance.is_currently_used %}
-                {% if poi_form.instance.events.filter_upcoming.exists %}
+            {% if not poi_form.instance.can_be_archived %}
+                {% if poi_form.instance.events.filter_upcoming_non_archived.exists %}
                     <div class="mb-5">
                         <div class="bg-orange-100 border-l-4 border-orange-500 text-orange-700 px-4 py-3 mb-5"
                              role="alert">
                             <p>
-                                {% translate "You cannot archive a location which is referenced by an event." %}
+                                {% translate "You cannot archive a location which is referenced by a non-archived event." %}
                                 <br />
-                                {% blocktranslate count counter=poi_form.instance.events.count trimmed %}
-                                    To archive this location, you have to delete this event first:
+                                {% blocktranslate count counter=poi_form.instance.events.filter_upcoming_non_archived.count trimmed %}
+                                    To archive this location, you have to archive or delete this event first:
                                 {% plural %}
-                                    To archive this location, you have to delete these events first:
+                                    To archive this location, you have to archive or delete these events first:
                                 {% endblocktranslate %}
                             </p>
                         </div>
-                        {% for event in poi_form.instance.events.filter_upcoming %}
+                        {% for event in poi_form.instance.events.filter_upcoming_non_archived %}
                             <a href="{% url 'edit_event' event_id=event.id region_slug=request.region.slug language_slug=language.slug %}"
                                class="block pt-2 hover:underline">
                                 <i icon-name="pen-square" class="mr-2"></i> {{ event.best_translation.title }}
@@ -51,21 +51,21 @@
                         {% endfor %}
                     </div>
                 {% endif %}
-                {% if request.region.contacts_enabled and poi_form.instance.contacts.exists %}
+                {% if request.region.contacts_enabled and poi_form.instance.non_archived_contacts.exists %}
                     <div class="mb-5">
                         <div class="bg-orange-100 border-l-4 border-orange-500 text-orange-700 px-4 py-3 mb-5"
                              role="alert">
                             <p>
-                                {% translate "You cannot archive a location which is referenced by a contact." %}
+                                {% translate "You cannot archive a location which is referenced by a non-archived contact." %}
                                 <br />
-                                {% blocktranslate count counter=poi_form.instance.contacts.count trimmed %}
-                                    To archive this location, you have to delete this contact first:
+                                {% blocktranslate count counter=poi_form.instance.non_archived_contacts.count trimmed %}
+                                    To archive this location, you have to archive or delete this contact first:
                                 {% plural %}
-                                    To archive this location, you have to delete these contacts first:
+                                    To archive this location, you have to archive or delete these contacts first:
                                 {% endblocktranslate %}
                             </p>
                         </div>
-                        {% for contact in poi_form.instance.contacts.all %}
+                        {% for contact in poi_form.instance.non_archived_contacts.all %}
                             <a href="{% url 'edit_contact' contact_id=contact.id region_slug=request.region.slug %}"
                                class="block pt-2 hover:underline">
                                 <i icon-name="pen-square" class="mr-2"></i> {{ contact.label_in_reference_list }}
--- a/integreat_cms/locale/de/LC_MESSAGES/django.po
+++ b/integreat_cms/locale/de/LC_MESSAGES/django.po
@@ -8100,29 +8100,42 @@ msgid "Archive location"
 msgstr "Ort archivieren"
 
 #: cms/templates/pois/poi_form_sidebar/action_box.html
-msgid "You cannot archive a location which is referenced by an event."
+msgid ""
+"You cannot archive a location which is referenced by a non-archived event."
 msgstr ""
-"Der Ort kann nicht archiviert werden, da er von einer Veranstaltung "
-"verwendet wird."
+"Der Ort kann nicht archiviert werden, da er von einer nicht-archivierten "
+"Veranstaltung verwendet wird."
 
 #: cms/templates/pois/poi_form_sidebar/action_box.html
-msgid "To archive this location, you have to delete this event first:"
-msgid_plural "To archive this location, you have to delete these events first:"
-msgstr[0] "Löschen Sie zuerst diese Veranstaltung, um den Ort zu archivieren:"
-msgstr[1] ""
-"Löschen Sie zuerst diese Veranstaltungen, um den Ort zu archivieren:"
-
-#: cms/templates/pois/poi_form_sidebar/action_box.html
-msgid "You cannot archive a location which is referenced by a contact."
-msgstr ""
-"Der Ort kann nicht archiviert werden, da er von einem Kontakt verwendet wird."
-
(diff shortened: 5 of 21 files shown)


integreat_cms/cms/models/pois/poi.py
```diff
@@ -180,6 +181,15 @@ def referring_contacts_exist(self) -> bool:
             return False
         return self.contacts.exists()
 
+    @property
+    def non_archived_contacts(self) -> QuerySet:
+        """
+        return: contacts that are not archived and referencing the POI
+        """
+        if not self.region.contacts_enabled:
+            return False
```
timobrembeck: This crashes when trying to archive a POI in a region where `contacts_enabled` is disabled, because `can_be_archived` calls `.exists()` on the return value:

```
AttributeError: 'bool' object has no attribute 'exists'
```

```suggestion
            return self.contacts.none()
```

integreat_cms/cms/models/events/event.py
```diff
@@ -62,6 +62,13 @@ def filter_upcoming(self, from_date: datetime | None = None) -> Self:
             ),
         )
 
+    def filter_upcoming_non_archived(self) -> Self:
+        """
+        Filter all events that take place after the given date and are not archived
+        return: The Queryset of events that are upcoming and not archived
```
timobrembeck:
```suggestion
        :return: The Queryset of events that are upcoming and not archived
```
Could you use `:return:` instead of `return:` so Sphinx picks it up?

integreat_cms/cms/models/pois/poi.py
```diff
@@ -180,6 +181,15 @@ def referring_contacts_exist(self) -> bool:
             return False
         return self.contacts.exists()
 
+    @property
+    def non_archived_contacts(self) -> QuerySet:
+        """
+        return: contacts that are not archived and referencing the POI
```
timobrembeck:
```suggestion
        :return: contacts that are not archived and referencing the POI
```

The docstring of `referring_contacts_exist` is also incorrect, but not changed in this PR and outside of the suggestion window... but if you want to, you could fix it as well.

integreat_cms/cms/templates/pois/poi_form_sidebar/action_box.html
```diff
@@ -28,44 +28,44 @@
...
-            {% if poi_form.instance.is_currently_used %}
-                {% if poi_form.instance.events.filter_upcoming.exists %}
+            {% if not poi_form.instance.can_be_archived %}
+                {% if poi_form.instance.events.filter_upcoming_non_archived.exists %}
                     <div class="mb-5">
                         <div class="bg-orange-100 border-l-4 border-orange-500 text-orange-700 px-4 py-3 mb-5"
                              role="alert">
                             <p>
-                                {% translate "You cannot archive a location which is referenced by an event." %}
+                                {% translate "You cannot archive a location which is referenced by an non-archived event." %}
```
timobrembeck:
```suggestion
                                {% translate "You cannot archive a location which is referenced by a non-archived event." %}
```

integreat_cms/cms/models/events/event.py
```diff
@@ -62,6 +62,13 @@ def filter_upcoming(self, from_date: datetime | None = None) -> Self:
             ),
         )
 
+    def filter_upcoming_non_archived(self) -> Self:
+        """
+        Filter all events that take place after the given date and are not archived
```
timobrembeck:
```suggestion
        Filter all upcoming events that are not archived
```

integreat_cms/cms/templates/pois/poi_form_sidebar/action_box.html
```diff
@@ -28,44 +28,44 @@
...
                                 <br />
                                 {% blocktranslate count counter=poi_form.instance.events.count trimmed %}
```
timobrembeck:
```suggestion
                                {% blocktranslate count counter=poi_form.instance.events.filter_upcoming_non_archived.count trimmed %}
```

integreat_cms/cms/templates/pois/poi_form_sidebar/action_box.html
```diff
@@ -28,44 +28,44 @@
...
-                {% if request.region.contacts_enabled and poi_form.instance.contacts.exists %}
+                {% if request.region.contacts_enabled and poi_form.instance.non_archived_contacts.exists %}
                     <div class="mb-5">
                         <div class="bg-orange-100 border-l-4 border-orange-500 text-orange-700 px-4 py-3 mb-5"
                              role="alert">
                             <p>
-                                {% translate "You cannot archive a location which is referenced by a contact." %}
+                                {% translate "You cannot archive a location which is referenced by a non-archived contact." %}
                                 <br />
                                 {% blocktranslate count counter=poi_form.instance.contacts.count trimmed %}
```
timobrembeck:
```suggestion
                                {% blocktranslate count counter=poi_form.instance.non_archived_contacts.count trimmed %}
```

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -11122,12 +11134,10 @@ msgstr "Alle Übersetzungen dieses Ortes werden gelöscht."
...
 "You cannot archive a location which is used by an upcoming event or a "
-"contact. \n"
-"This also involves archived events and contacts"
+"contact."
 msgstr ""
-"Sie können keinen Ort archivieren, der von einer zukünftigen Veranstaltung "
-"oder einem Kontakt verwendet wird. \n"
-"Dies betrifft auch archivierte Veranstaltungen und Kontakte"
+"Der Ort kann nicht archiviert werden, da er von einer Veranstaltung "
+"verwendet wird."
```
timobrembeck:
```suggestion
"Sie können keinen Ort archivieren, der von einer zukünftigen Veranstaltung "
"oder einem Kontakt verwendet wird."
```

timobrembeck (changes requested):
Thanks a lot! :muscle:

Unfortunately this crashes for all regions with disabled contacts. Could you also add a test for the new archiving conditions (e.g. that an archived event/contact no longer blocks archiving, and a region with disabled contacts)?

integreat_cms/cms/models/pois/poi.py
MizukiTemma: 🙈 Totally forgotten that the conact feature can be turned off 🙈

integreat_cms/cms/models/pois/poi.py
```diff
@@ -173,13 +174,22 @@ def restore(self) -> None:
     @property
     def referring_contacts_exist(self) -> bool:
         """
-        return: whether there are contact objects referring the POI.
+        :return: whether there are contact objects referring the POI.
         If the contact module is disabled in the region, it's always False
```
timobrembeck: This unindented continuation line ends the `:return:` field list and makes the doc build fail:
`WARNING: Field list ends without a blank line; unexpected unindent.`
(previously `return:` was plain text, now `:return:` is an RST field, so the follow-up line must be indented)

```suggestion
            If the contact module is disabled in the region, it's always False
```

integreat_cms/cms/models/events/event.py
```diff
@@ -62,6 +62,13 @@ def filter_upcoming(self, from_date: datetime | None = None) -> Self:
             ),
         )
 
+    def filter_upcoming_non_archived(self) -> Self:
+        """
+        Filter all upcoming events that are not archived
+        :return: The Queryset of events that are upcoming and not archived
```
timobrembeck, line 69: Nit: blank line between description and `:return:`, matching other docstrings in this file.

```suggestion

        :return: The Queryset of events that are upcoming and not archived
```

timobrembeck (changes requested):
Thanks for your changes! 💪 

After thinking a bit about the new workflow, I noticed a small problem:
Since users can revert an archiving operation, the following is now possible:

archive event/contact -> archive POI -> restore event/contact ⚠️ 

Then, we would have an event or contact which references an archived POI (which is not delivered via the API), so the app is confronted with an inconsistency in our data set.
I would suggest to prevent events & contacts from being restored as long as their referenced POI is archived as well, or what do you think?
Sorry I did not notice this in my initial review.

MizukiTemma:
> Thanks for your changes! 💪
> 
> After thinking a bit about the new workflow, I noticed a small problem: Since users can revert an archiving operation, the following is now possible:
> 
> archive event/contact -> archive POI -> restore event/contact ⚠️
> 
> Then, we would have an event or contact which references an archived POI (which is not delivered via the API), so the app is confronted with an inconsistency in our data set. I would suggest to prevent events & contacts from being restored as long as their referenced POI is archived as well, or what do you think? Sorry I did not notice this in my initial review.

Nothing to be sorry about, I had to notice that problem 😣 

I'll forward the question to @osmers

osmers:
> After thinking a bit about the new workflow, I noticed a small problem: Since users can revert an archiving operation, the following is now possible:
> 
> archive event/contact -> archive POI -> restore event/contact ⚠️
> 
> Then, we would have an event or contact which references an archived POI (which is not delivered via the API), so the app is confronted with an inconsistency in our data set. I would suggest to prevent events & contacts from being restored as long as their referenced POI is archived as well, or what do you think?

Either that option or we could restore the POI automatically upon restoring the contact/event. Maybe with a warning/information or something for the user. Sorry to pass this even further on, but @m0ndra should decide what makes sense looking at the overall system workflows and what's intuitive/understandable for users or works with her ideas of workflows.

m0ndra:
I am confused. In what order is the user able to restore a contact then? Can the user begin with the POI and THEN restore the contact? So is this just a sequence the user has to stick to? This whole POI+Contacts entanglement is already so super confusing for everyone, I'd like to keep it as simple as possible. 

👉 So if we can restore BOTH at the same time, with a proper hint for the user that the system is doing it, I'd go for that approach. @MizukiTemma is that doable?

MizukiTemma:
> 👉 So if we can restore BOTH at the same time, with a proper hint for the user that the system is doing it, I'd go for that approach. @MizukiTemma is that doable?

Thank you for the input 😸 

Should be, and so implemented in the new commit ✏️

m0ndra:
@MizukiTemma How does the hint look like? Can you show me what the user sees after performing the action?

MizukiTemma:
> @MizukiTemma How does the hint look like? Can you show me what the user sees after performing the action?

Hint (sentence or visual something) is not yet there. I meant only the functionality.

MizukiTemma:
@m0ndra 
I would say, we can and should pack this explanation into wiki until the whole reform of POI- Event/Contact relation🤔 
Or maybe adding a new sentence like "The Location which is referenced by this contact/Event will be restored too." to the tooltip? 
image
image

m0ndra:
The screenshots you're showing refer to an event 🙈 I'm not sure if the workflows are identical? I'd be happy if we can think about a visual concept together in a 15min call.

m0ndra:
We decided to update the hint overlays that appear as soon as someone wants to restore pages. The overlay should display all the connected content and their respective titles that will be restored, so that the user knows what the system does in the background:  
grafik

MizukiTemma:
The popup is updated to mention to the POI if referenced ✏️

m0ndra:
@MizukiTemma can post a screenshot, how the popup looks like now when a POI is referenced? Also I think we need to tell the Service People that something changed with the workflow, right?

MizukiTemma:
If the referenced place is not archived:

image
image

If archived:
image

image


The same popup appears when user tries to restore a place/event in a form, too 🍿

MizukiTemma:
@laravoelk @almahoefler @osmers 

Here is a short update about changed not originally included in the issue: when an archived contact or event is restores, if the referenced place is also archchived, the place will be restored together with the event/contact, and user see before proceeding such a popup like anove screenshots.

integreat_cms/cms/templates/contacts/contact_form.html
```diff
@@ -165,6 +165,7 @@ <h3 class="heading font-bold text-black">
                                     data-confirmation-title="{{ restore_dialog_title }}"
                                     data-confirmation-text="{{ restore_dialog_text }}"
                                     data-confirmation-subject="{{ contact_form.instance.name }}"
+                                    {% if contact.location.archived %} data-confirmation-referenced-location-title="{{ referenced_location_title }}" data-confirmation-referenced-location-subject="{{ contact_form.instance.location }}" {% endif %}
```
timobrembeck:
```suggestion
                                    {% if contact_form.instance.location.archived %} data-confirmation-referenced-location-title="{{ referenced_location_title }}" data-confirmation-referenced-location-subject="{{ contact_form.instance.location }}" {% endif %}
```
`contact` is not available in the context of the `contact_form`.

integreat_cms/static/src/js/utils/confirmation-popup.ts
```diff
@@ -72,6 +72,14 @@ export const showConfirmationPopup = (event: Event, onSubmit?: (event: Event) =>
         button.getAttribute("data-confirmation-text"),
         onSubmit
     );
+    const referencedLocationTitle = button.getAttribute("data-confirmation-referenced-location-title");
+    if (referencedLocationTitle) {
+        document.getElementById("confirmation-referenced-location-title").textContent = referencedLocationTitle;
+    }
```
timobrembeck:
```suggestion
    document.getElementById("confirmation-referenced-location-title").textContent = referencedLocationTitle ?? "";
```
When we only set the attribute when it's present, it will be stale when the confirmation popup is closed and another popup with empty `confirmation-referenced-location-title` attribute is opened, e.g. the deletion confirmation:

Image

integreat_cms/static/src/js/utils/confirmation-popup.ts
```diff
@@ -72,6 +72,14 @@ export const showConfirmationPopup = (event: Event, onSubmit?: (event: Event) =>
...
+    const referencedLocationSubject = button.getAttribute("data-confirmation-referenced-location-subject");
+    if (referencedLocationTitle) {
+        document.getElementById("confirmation-referenced-location-subject").textContent = referencedLocationSubject;
+    }
```
timobrembeck:
```suggestion
    document.getElementById("confirmation-referenced-location-subject").textContent = referencedLocationSubject ?? "";
```

timobrembeck (changes requested):
Thanks, the archiving conditions and the crash fix look good now! 💪 Two things from the restore cascade:

- The referenced-location notice never shows in the contact **form**, because the guard uses `contact`, which isn't in the form context (it's `contact_form.instance` there). See suggestion.
- The confirmation dialog is shared, so the referenced-location fields keep stale content in the next dialog opened on the same page. See suggestion.

And optionally: could you also add a test for the cascade restore itself (restore an archived event/contact whose location is archived → the location gets restored)?

MizukiTemma:
@timobrembeck 
Thank you for feedback 😸 The suggestions are applied and small tests are added ✅

tests/cms/models/contacts/test_contacts.py
```diff
@@ -85,3 +93,48 @@ def test_restoring_contact_works(
 
     assert Contact.objects.all().count() == 6
     assert contact.archived is False
+
+
+@pytest.mark.django_db
+def test_event_restore_with_referenced_poi() -> None:
```
timobrembeck:
```suggestion
def test_contact_restore_with_referenced_poi() -> None:
```
Just a cosmetic detail

timobrembeck (approved):
Awesome, thanks a lot for addressing my feedback, I have nothing left to complain about :wink:

seluianova (commented):
I have 2 thoughts:

1. We show this warning when the user restores a contact from the edit form:

Image

However, it's also possible to restore a contact via the bulk actions. In that case, there is no warning, and **the linked place is restored silently without any feedback** to the user.

_Actually, we don't show a confirmation dialog for archiving or restoring content via the bulk actions at all, so perhaps this should be handled in a separate issue..._

**UPD**. I have just realized that you posted screenshots from bulk actions and you have the warning? I don't know why I don't have any 🤷🏻‍♀️

2. I also find the `place → contact` constraint a bit confusing.

When I create a place, I can enter contact information, which automatically creates a primary contact in the database. 
After that, I can't simply delete or archive the place anymore. I first have to go to the `Contacts` and delete or archive the linked contact, even though I didn't explicitly create it.

Since contacts are always linked to places (they can't exist without a place), maybe they should always be treated as a single entity?

* User deletes a place → the contacts are deleted automatically.
* User archives a place → the contacts are archived automatically.

This isn't really related to this PR, though. Just thinking out loud. 🤔

seluianova (approved):
Besides the general concerns, the changes look good to me 👍

MizukiTemma:
> 2. I also find the `place → contact` constraint a bit confusing.
> 
> When I create a place, I can enter contact information, which automatically creates a primary contact in the database. After that, I can't simply delete or archive the place anymore. I first have to go to the `Contacts` and delete or archive the linked contact, even though I didn't explicitly create it.
> 
> Since contacts are always linked to places (they can't exist without a place), maybe they should always be treated as a single entity?
> 
> * User deletes a place → the contacts are deleted automatically.
> * User archives a place → the contacts are archived automatically.
> 
> This isn't really related to this PR, though. Just thinking out loud. 🤔

@m0ndra is analysing and designing a new workflow 🛠️ This PR is a solution until then (becuase it takes a while)

MizukiTemma:
> I have 2 thoughts:
> 
> 1. We show this warning when the user restores a contact from the edit form:
> 
> Image
> However, it's also possible to restore a contact via the bulk actions. In that case, there is no warning, and **the linked place is restored silently without any feedback** to the user.
> 
> _Actually, we don't show a confirmation dialog for archiving or restoring content via the bulk actions at all, so perhaps this should be handled in a separate issue..._
> 
> **UPD**. I have just realized that you posted screenshots from bulk actions and you have the warning? I don't know why I don't have any 🤷🏻‍♀️
> 


It's not a screenshot of bulk action warning, but an warning for the restore icon 🔄 

Yes, there is no confirmation popup for bulk action of contacts.  I have extended the description of #4403

seluianova:
> It's not a screenshot of bulk action warning, but an warning for the restore icon 🔄

ah, I missed that we have even 3 ways to restore content objects 🙈

PR #4440, opened 2026-07-16
Fetch all available news from Tü News

### LINKED ISSUE #4439: [Task]: Fetch all available news from Tü News
### Task Description

We are currently fetching only the first 10 Posts from Tü News which is the default behaviour if no `per_page` and `page` parameters are set.

Instead fetch all available news posts.

### Task Category

Use `page` and `per_page` parameter or another one if there is any better, and fetch all posts

### Acceptance Criteria

- Not only the first 10 but all available posts are fetched
- Filter by `["acf"]["integreat"]` stays as before

### Technical Details
FInd the API Documentation [here](https://tuenews.de/wp-json/)
### Risks & Impact

_No response_

### Target Environment

_No response_

### Verification / Testing

_No response_

### Additional Context
This change was requested by App Team (@steffenkleinle )

This issue is blocking the next release in the app.

https://github.com/digitalfabrik/integreat-app/issues/2537

### Definition of Done

MizukiTemma:
### Short description

This PR updates `import_news_items` of `TunewsManager` so all the available news posts are fetched.

### Proposed changes

- Use `page` and `per_page` parameters to check all news posts
- Adjust log message

### Side effects

- None

~~There are no intended deviations from the issue and design.~~
During review we decided for some further changes:
- Fetch only news posts of the last `TUNEWS_HISTORY_DAYS` days (with default `TUNEWS_HISTORY_DAYS = FCM_HISTORY_DAYS = 28`)
- Deliver all news posts without pagination if no `size` parameter is given in the common news endpoint `news/`

### How to test

- run `integreat-cms-cli import_externa_news`
- See more news are imported

### Resolved issues

Fixes: #4439 

**Updates**
Debug code lines from (this commit](https://github.com/digitalfabrik/integreat-cms/commit/ae3bfd69962c6a57992ea978f078db5014bb09f1) were moved.

### DIFF
--- a/integreat_cms/api/v3/news.py
+++ b/integreat_cms/api/v3/news.py
@@ -75,11 +75,14 @@ def news(
             result for result in sorted_result if result["source"] in sources
         ]
 
-    page_num = request.GET.get("page", 1)
     try:
         page_size = min(int(request.GET.get("size", 20)), 500)
     except (TypeError, ValueError):
         page_size = 20
+
+    if "page" not in request.GET and "size" not in request.GET:
+        return JsonResponse(list(sorted_result), safe=False)
+    page_num = request.GET.get("page", 1)
     paginator = Paginator(sorted_result, page_size)
     page = get_safe_page(paginator, page_num)
 
--- a/integreat_cms/core/circleci_settings.py
+++ b/integreat_cms/core/circleci_settings.py
@@ -40,6 +40,8 @@ CACHEOPS_ENABLED = False
 LINKCHECK_DISABLE_LISTENERS = True
 # Disable background tasks during testing
 BACKGROUND_TASKS_ENABLED = False
+# Disable auto re-import of external news on demand
+EXTERNALNEWS_DISABLE_AUTO_REIMPORT = True
 #: Enable logging of all entries from the messages framework
 MESSAGE_LOGGING_ENABLED = True
 #: Use debug logging on CircleCI
--- a/integreat_cms/core/settings.py
+++ b/integreat_cms/core/settings.py
@@ -281,6 +281,24 @@ FCM_NOTIFICATION_RETAIN_TIME_IN_HOURS: Final[int] = env_int(
     24,
 )
 
+#################
+# External News #
+#################
+
+#: How many days back news posts should be fetched from Tü News
+#: Default to FCM_HISTORY_DAYS
+TUNEWS_HISTORY_DAYS: Final[int] = env_int(
+    "INTEGREAT_CMS_TUNEWS_HISTORY_DAYS",
+    FCM_HISTORY_DAYS,
+)
+
+#: Disable re-importing of external news posts on demand e.g. in test context
+EXTERNALNEWS_DISABLE_AUTO_REIMPORT: bool = bool(
+    strtobool(
+        os.environ.get("INTEGREAT_CMS_EXTERNALNEWS_DISABLE_AUTO_REIMPORT", "False")
+    ),
+)
+
 ###########
 # GVZ API #
 ###########
--- a/integreat_cms/news_managers/tunews_manager.py
+++ b/integreat_cms/news_managers/tunews_manager.py
@@ -1,10 +1,11 @@
 from __future__ import annotations
 
 import logging
-from datetime import datetime
+from datetime import datetime, timedelta
 from typing import TypedDict
 
 import requests
+from django.conf import settings
 from django.core.cache import cache
 
 from ..cms.models import Language
@@ -42,54 +43,76 @@ class TunewsManager(AbstractNewsManager):
         """
         Imports Tü News posts and save them as cache
         """
+        per_page = 100
+        oldest_date = (
+            datetime.now() - timedelta(days=settings.TUNEWS_HISTORY_DAYS)
+        ).isoformat()
+
         for language in Language.objects.all():
-            try:
-                response = requests.get(
-                    f"https://tuenews.de/wp-json/wp/v2/posts/?lang={language.slug}",
-                    timeout=10,
-                )
-
-                if response.status_code != 200:
-                    logger.error(
-                        "Could not find posts in %s.",
-                        language,
+            news = []
+            page_count = 1
+            import_failed = False
+            while True:
+                try:
+                    response = requests.get(
+                        f"https://tuenews.de/wp-json/wp/v2/posts/?lang={language.slug}&per_page={per_page}&page={page_count}&after={oldest_date}",
+                        timeout=10,
                     )
-                    continue
+                except requests.exceptions.RequestException:
+                    logger.exception(
+                        "Failed to fetch posts from TuNews in %s.", language
+                    )
+                    import_failed = True
+                    break
 
-                posts = response.json()
+                result = response.json()
+                if response.status_code != 200:
+                    code = result.get("code")
+                    if code == "rest_invalid_param" and "lang" in result.get(
+                        "data", {}
+                    ).get("params", {}):
+                        # Language genuinely not served by Tü News, not a failure:
+                        # cache the (empty) result to avoid retrying import at every request.
+                        logger.debug(
+                            "Tü News does not serve %s; caching empty result.", language
+                        )
+                    elif code == "rest_post_invalid_page_number":
+                        # No more pages, not a failure: keep the posts gathered so far.
+                        logger.debug("Reached end of pagination for %s.", language)
+                    else:
+                        logger.error(
+                            "Could not fetch page %s in %s.", page_count, language
+                        )
+                        import_failed = True
+                    break
 
                 logger.info(
-                    "Got %s posts in %s.",
-                    len(posts),
+                    "Got %s result in %s.",
+                    len(result),
                     language,
                 )
-
-                news = []
-
-                for post in posts:
+                for post in result:
                     try:
                         if not post["acf"]["integreat"]:
                             continue
                         news.append(self.transform_post(post))
                     except (KeyError, TypeError, ValueError):
                         logger.exception(
-                            "Malformed %s post (id=%s); skipped.",
-                            self.name,
+                            "Malformed Tü News post (id=%s); skipped.",
                             post.get("id"),
                         )
 
+                if len(result) < per_page:
+                    break
+                page_count += 1
+
+            if not import_failed:
                 cache.set(f"{self.short_name}:{language.slug}", news, timeout=None)
-                logger.info("Saving %s news in %s", len(news), language)
                 logger.info(
                     "Saved %s news in %s",
                     len(cache.get(f"{self.short_name}:{language.slug}")),
                     language,
                 )
-                for post in cache.get(f"{self.short_name}:{language.slug}"):
-                    logger.info(post.get("title"))
-
-            except requests.exceptions.RequestException:
-                logger.exception("Failed to fetch posts in %s.", language)
 
     def transform_post(self, post: _TunewsPost) -> NewsItem:
         """
--- a/tests/news_managers/test_tunews_manager.py
+++ b/tests/news_managers/test_tunews_manager.py
@@ -1,11 +1,12 @@
 from __future__ import annotations
 
 from datetime import datetime, UTC
-from unittest.mock import patch
+from unittest.mock import Mock, patch
 
 import pytest
 from django.core.cache import cache
 
+from integreat_cms.cms.models import Language
 from integreat_cms.news_managers.abstract_news_manager import clean_html
 from integreat_cms.news_managers.amalnews_manager import AmalnewsManager
 from integreat_cms.news_managers.tunews_manager import TunewsManager
@@ -116,24 +117,54 @@ dummy_news_items = [
 
 
 @pytest.mark.django_db
-def test_import_news_item(load_test_data: None, clean_news_cache: None) -> None:
+def test_import_tunews_item(clean_news_cache: None) -> None:
+    response_1 = Mock()
+    response_1.status_code = 200
+    response_1.json.return_value = dummy_news_items
+
+    response_2 = Mock()
+    response_2.status_code = 400
+
+    first_language_slug = Language.objects.first().slug
+
     with patch(
         "integreat_cms.news_managers.tunews_manager.requests.get"
     ) as fake_tunews_server:
-        fake_tunews_server.return_value.status_code = 200
-        fake_tunews_server.return_value.json.return_value = dummy_news_items
+        fake_tunews_server.side_effect = [response_1, response_2] * (
+            Language.objects.all().count()
+        )
 
-        assert not cache.get(f"{tu_short_name}:de")
+        assert not cache.get(f"{tu_short_name}:{first_language_slug}")
 
         TunewsManager().import_news_items()
 
-        assert cache.get(f"{tu_short_name}:de") == expected_result_tunews
+        assert (
+            cache.get(f"{tu_short_name}:{first_language_slug}")
+            == expected_result_tunews
+        )
 
-        assert not cache.get(f"{amal_short_name}:de")
+
+@pytest.mark.django_db
+def test_import_amalnews_item(clean_news_cache: None) -> None:
+    response_1 = Mock()
+    response_1.status_code = 200
+    response_1.json.return_value = dummy_news_items
+
+    first_language_slug = Language.objects.first().slug
+
+    with patch(
+        "integreat_cms.news_managers.amalnews_manager.requests.get"
+    ) as fake_tunews_server:
+        fake_tunews_server.side_effect = [response_1] * (Language.objects.all().count())
+
+        assert not cache.get(f"{amal_short_name}:{first_language_slug}")
 
         AmalnewsManager().import_news_items()
 
-        assert cache.get(f"{amal_short_name}:de") == expected_result_amalnews
+        assert (
+            cache.get(f"{amal_short_name}:{first_language_slug}")
+            == expected_result_amalnews
+        )
 
 
 def test_clean_html_keeps_plain_text() -> None:
--- a/integreat_cms/api/v3/social_media_headers.py
+++ b/integreat_cms/api/v3/social_media_headers.py
@@ -112,7 +112,7 @@ def render_error_headers(request: HttpRequest, error: str) -> HttpResponse:
 @partial_html_response
 def root_social_media_headers(
     request: HttpRequest,
-    language_slug: str = settings.LANGUAGE_CODE,
+    language_slug: str | None = None,
 ) -> HttpResponse:
     """
     Renders the social media headers for a root page
@@ -122,7 +122,10 @@ def root_social_media_headers(
 
     :return: HTML social meta headers required by social media platforms
     """
-    language = get_object_or_404(Language, slug=language_slug)
+    language = get_object_or_404(
+        Language,
+        slug=language_slug or settings.LANGUAGE_CODE,
+    )
     title = language.social_media_webapp_title or settings.BRANDING_TITLE
     url = site_url(request)
 
--- a/integreat_cms/news_managers/abstract_news_manager.py
+++ b/integreat_cms/news_managers/abstract_news_manager.py
@@ -5,6 +5,7 @@ from abc import ABC, abstractmethod
 from io import StringIO
 from typing import TYPE_CHECKING, TypedDict
 
+from django.conf import settings
 from django.core.cache import cache
 from django.http import Http404, JsonResponse
 from lxml import etree
@@ -75,9 +76,14 @@ class AbstractNewsManager(ABC):
         cache_key = f"{self.short_name}:{language_slug}"
         posts = cache.get(cache_key, _CACHE_MISS)
         if posts is _CACHE_MISS:
-            logger.info("Cache miss for %s; importing news items on demand.", cache_key)
-            self.import_news_items()
-            posts = cache.get(cache_key, [])
+            if not settings.EXTERNALNEWS_DISABLE_AUTO_REIMPORT:
+                logger.info(
+                    "Cache miss for %s; importing news items on demand.", cache_key
+                )
+                self.import_news_items()
+                posts = cache.get(cache_key, [])
+            else:
+                return []
         return posts
 
     def collect_news_items(
--- a/tests/api/test_news_endpoint.py
+++ b/tests/api/test_news_endpoint.py
@@ -44,7 +44,9 @@ def _create_push_notification(
 
 
 @pytest.mark.django_db
-def test_news_endpoint(load_test_data: None, clean_news_cache: None) -> None:
+def test_news_endpoint(
+    load_test_data: None, clean_news_cache: None, disable_auto_news_reimport: None
+) -> None:
     """
     The combined endpoint merges items from every news source and sorts them
     by ``display_date``, newest first.
@@ -121,7 +123,9 @@ def test_news_endpoint(load_test_data: None, clean_news_cache: None) -> None:
 
 
 @pytest.mark.django_db
-def test_news_endpoint_pagination(load_test_data: None, clean_news_cache: None) -> None:
+def test_news_endpoint_pagination(
+    load_test_data: None, clean_news_cache: None, disable_auto_news_reimport: None
+) -> None:
     """
     The combined endpoint returns only one page of results at a time.
     Page 1 contains the most recent items; page 2 contains the next batch.
@@ -170,7 +174,7 @@ def test_news_endpoint_pagination(load_test_data: None, clean_news_cache: None) 
 
 @pytest.mark.django_db
 def test_news_endpoint_source_filter(
-    load_test_data: None, clean_news_cache: None
+    load_test_data: None, clean_news_cache: None, disable_auto_news_reimport: None
 ) -> None:
     """
     The combined endpoint supports filtering by source.
@@ -260,7 +264,9 @@ def test_news_endpoint_source_filter(
 
 
 @pytest.mark.django_db
-def test_single_news_endpoint(load_test_data: None, clean_news_cache: None) -> None:
+def test_single_news_endpoint(
+    load_test_data: None, clean_news_cache: None, disable_auto_news_reimport: None
+) -> None:
     """
     The single news endpoint returns one news that matches the given id.
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -176,3 +176,11 @@ def clean_news_cache(load_test_data: None) -> Generator[None]:
     yield
     for key in keys:
         cache.delete(key)
+
+
+@pytest.fixture()
+def disable_auto_news_reimport(settings: SettingsWrapper) -> None:
+    """
+    Disable re-import of external news on demand to avoid hitting the real APIs and getting real news posts
+    """
+    settings.EXTERNALNEWS_DISABLE_AUTO_REIMPORT = True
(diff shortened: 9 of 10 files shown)


MizukiTemma:
I would like to include this PR into the release `2026.7.0` as it's blocking the app release.

jonbulz (changes requested):
You're setting up an infinite loop in the tests. The mock `news_manager` return value is always 200, thus the exit condition of the loop is never met.
Also, if they ever introduce rate-limiting, we are likely to fall into it. This loop fires rapid requests until if fetches _everything_ for _all_ languages. I just checked and arabic alone went to page 15, meaning 15 requests and ~1500 news items just for this one language. Maybe we should sleep between requests to avoid accidentaly DoS-ing them?
Also, a general remark: Do we really need all news items? The API has a `after` parameter that would allow us to fetch news of e.g. the latest month or so. Wouldn't that be sufficient as news are, you know, supposed to be news?

MizukiTemma:
> Also, a general remark: Do we really need all news items? The API has a after parameter that would allow us to fetch news of e.g. the latest month or so. Wouldn't that be sufficient as news are, you know, supposed to be news?


@steffenkleinle 
Now it's not only me who think fetching all is too much 😅 And I think `after` parameter is a good idea. We have a similar freshness check for local news (currently 28 days?). Is it really necessary to collect them all?

steffenkleinle:
As far as I understood it, the plan was to cache/save the news in the CMS, so wouldn't it be possible to fill the database over time to avoid rate limiting etc? If that is the case, I am not sure why we would need to limit it to one month only. But if it helps you, obviously implement a cutoff. I was just mentioning that currently only 5 tünews are included in the news endpoint, which is barely one week and imo too little.
Implementing a cutoff is a breaking change for us in any case as previously it was possible to get every and any tünews. In the end this decision is up to @osmers imo.

jonbulz:
In the discussion in the original issue #4174 we decided against saving news in our database. Can you describe why it is a breaking change if not every and all tünews are available? As far as I understood, the point that we only need the more recent news was made as an argument in favor of the cache option

steffenkleinle:
> In the discussion in the original issue #4174 we decided against saving news in our database.

Can't say anything to that.

> Can you describe why it is a breaking change if not every and all tünews are available?

Breaking change was maybe the wrong naming, however, it will be a change to users since currently all tünews are available in the app.

> As far as I understood, the point that we only need the more recent news was made as an argument in favor of the cache option

I don't know about that, but on the conference our talks seemed like it would not be a problem to cache all news/load more news from the tünews backend on demand. In the end, I actually don't mind, just thought it would be nice/in line with the current app capabilities and I don't see a downside from an app/user perspective that all news are available. However, if it is a problem on the CMS side, we can just do the last month.

steffenkleinle:
@MizukiTemma not sure if its related, but currently the endpoint is again not returning any tünews:https://cms-test.integreat-app.de/api/v3/testumgebung/de/news/?page=1&count=20&source=tunews

MizukiTemma:
> @MizukiTemma not sure if its related, but currently the endpoint is again not returning any tünews:https://cms-test.integreat-app.de/api/v3/testumgebung/de/news/?page=1&count=20&source=tunews

I ran the management command. It should be working again.

jonbulz:
I'm sorry, but it's just really hard to make good architecture decisions if the requirements come one by one, or are changing. The logic is already a bit bloated because we had to implement an additional single news endpoint and filtering, something which I specifically asked about [here](https://github.com/digitalfabrik/integreat-cms/issues/4174#issuecomment-4126663852) and got an answer [here](https://github.com/digitalfabrik/integreat-cms/issues/4174#issuecomment-4143291301). This would have been a lot cheaper if we had saved the news in a Django model from the start.
The way the caching currently works is one key per source per language, e.g. `tunews:ar`, and the value then is an array of news items. Which means we can't simply add more items to the pile, we'd have to either:
- fetch everything, everytime, and hope they don't block our IP

or
- retrieve the cache
- fetch the news from the tunews api
- merge with the existing news array and de-duplicate
- hope that our cache doesn't reset and implement a fetch-all logic in case it ever does

It is possible to do this, but it is not a one line change. Fetching and caching only the latest month or two would hopefully mean that the amount of requests we send stays more or less constant, that's why I proposed it. 

I'm just highlighting this because it is a good example of how we are inefficient due to a lack of proper communcation. For me, it was again not clear after the first comment if having access to all news is a hard requirement, or just nice to have. Maybe we can learn from this and properly map out the requirements at the start next time, not just the what but also the why, so everybody is on the same page from the beginning.

jonbulz:
So while my last comment reads like a bit of a vent, I did some math and with the current number of posts on tunews, it should be possible to always fetch all news if we think that's necessary. I still want to highlight that this does not scale well, though. The way it's implemented in this PR, we do the requests sequentially, so we won't overload their server with it (unless the response status code is != 200, but that's something we can fix). But another issue is that this also steadily increases the processing and allocated memory on our side on every read. The allocated memory per `news` and `news/news_id` request is probably what's going to bite us first. I don't know how many requests we are expecting to our endpoints, though we should have some headroom.

MizukiTemma:
Thank you for opinions 🙏 

We are discussing based on the fectors that are neither officially stated nor we can influence: whether rate limiting will be introduced (if so, what is the max rate), how much news there will be in future in Tü News (whether old news are removed regually or news posts accumulate endlessly), etc.

If the compliance to the current behaviour is the only reason to fetch all news posts, I would like to discuss validity of this requirement first. @osmers

osmers:
I checked whether we have any external requirements to fetch all and apparently we do not. So how about we use the same limit as for internal news, i.e. 28 days? Anything older gets removed/deleted?

MizukiTemma:
> I checked whether we have any external requirements to fetch all and apparently we do not. So how about we use the same limit as for internal news, i.e. 28 days? Anything older gets removed/deleted?

@steffenkleinle 

I would like to go for `after` parameter and 28 days freshness criteria. Is that acceptable for you?
An example of 28days from 22 Jul 2026: https://tuenews.de/wp-json/wp/v2/posts/?lang=de&per_page=100&page=1&after=2026-06-24T00:00:00Z

steffenkleinle:
> I'm sorry, but it's just really hard to make good architecture decisions if the requirements come one by one, or are changing. The logic is already a bit bloated because we had to implement an additional single news endpoint and filtering, something which I specifically asked about [here](https://github.com/digitalfabrik/integreat-cms/issues/4174#issuecomment-4126663852) and got an answer [here](https://github.com/digitalfabrik/integreat-cms/issues/4174#issuecomment-4143291301).

I am sorry for that, but I guess that was a misunderstanding then. I only took the question in regards to endpoints with multiple news (e.g. `one for each tab`) too which I replied its not necessary. I just assumed its clear that we keep/replicate the current endpoints, i.e. one for the news list and one endpoint for single news. Generally, its just always necessary to have an endpoint for a single news with id if you use pagination imo. Nevertheless, I should have spotted that and told you that we also need another endpoint.

This would have been a lot cheaper if we had saved the news in a Django model from the start. The way the caching currently works is one key per source per language, e.g. `tunews:ar`, and the value then is an array of news items. Which means we can't simply add more items to the pile, we'd have to either:
> 
>     * fetch everything, everytime, and hope they don't block our IP
> 
> 
> or
> 
>     * retrieve the cache
> 
>     * fetch the news from the tunews api
> 
>     * merge with the existing news array and de-duplicate
> 
>     * hope that our cache doesn't reset and implement a fetch-all logic in case it ever does
> 
> 
> It is possible to do this, but it is not a one line change. Fetching and caching only the latest month or two would hopefully mean that the amount of requests we send stays more or less constant, that's why I proposed it.

Thanks for explaining. With that information, I'd have said from the start that its not necessary to have all news. I just took the talks on the conference in a very different way since there it was mentioned that is basically no problem to just cache all news and populate it piece by piece.

> I'm just highlighting this because it is a good example of how we are inefficient due to a lack of proper communcation. For me, it was again not clear after the first comment if having access to all news is a hard requirement, or just nice to have. Maybe we can learn from this and properly map out the requirements at the start next time, not just the what but also the why, so everybody is on the same page from the beginning.

Yes, totally agree. There was basically no communication and just assumptions on both sides :D

steffenkleinle:
> > I checked whether we have any external requirements to fetch all and apparently we do not. So how about we use the same limit as for internal news, i.e. 28 days? Anything older gets removed/deleted?
> 
> @steffenkleinle
> 
> I would like to go for `after` parameter and 28 days freshness criteria. Is that acceptable for you? An example of 28days from 22 Jul 2026: https://tuenews.de/wp-json/wp/v2/posts/?lang=de&per_page=100&page=1&after=2026-06-24T00:00:00Z

For sure, that works. In that case, however, we should maybe just always fetch all news the app instead of using pagination. Shouldn't be too many items in that case. I know that this is then another changing requirement, but would it then be possible to just return all items if no `size` query param is set?

I assume that we then also apply the same 28 day logic for amal news?

MizukiTemma:
> I assume that we then also apply the same 28 day logic for amal news?

The number of days can be set per news source. It doesn't have to be the same for all news sources.

We don't know how much news posts there will be in Amal News. They are currently providing only one single post for testing for us. If they regularly clean up ther post lists We probably don't need any freshness restriction and fetch all without thinking  about rate limit etc.

MizukiTemma:
I have updated the description with the deviations decided in the above discussion.

@jonbulz 
It's ready for re-review 🔍

integreat_cms/core/settings.py
```diff
@@ -275,6 +275,16 @@
...
 )
 
+#################
+# External News #
+#################
+
+#: How many days back news posts should be fetched from Tü News
+#: Default to FCM_HISTORY_DAYS
+TUNEWS_HISTORY_DAYS: Final[int] = int(
+    os.environ.get("INTEGREAT_CMS_FCM_HISTORY_DAYS", FCM_HISTORY_DAYS),
```
jonbulz: are we reading the correct env variable here? I think it makes more sense to either use a different environment variable, or to just set `TUNEWS_HISTORY_DAYS = FCM_HISTORY_DAYS`, or just use `FCM_HISTORY_DAYS` in the tunews manager.
also, we started using the `env_int` helper function for most of the settings, so we might want to do this here as well for consistency

integreat_cms/news_managers/tunews_manager.py
```diff
@@ -42,54 +43,62 @@ def import_news_items(self) -> None:
...
+                            if not post["acf"]["integreat"]:
+                                continue
+                            news.append(self.transform_post(post))
+                        except (KeyError, TypeError, ValueError):
+                            logger.exception(
+                                "Malformed Tü News post (id=%s); skipped.",
+                                post.get("id"),
+                            )
+
+                except requests.exceptions.RequestException:
```
jonbulz, line 61: with the `try/except` in the `while` loop like this, if we get a `RequestException` continuously, we increase the page number indefinetly and fire infinite requests because our exit condition is unreachable. we need some `break` condition in the `except` block. maybe with `max_retries` if you prefer?
maybe we even want to introduce a `max_page_count` as an additional safeguard.

integreat_cms/news_managers/tunews_manager.py
```diff
@@ -42,54 +43,62 @@ def import_news_items(self) -> None:
...
+                    logger.exception("Failed to fetch posts in %s.", language)
+
+                page_count += 1
+
+            cache.set(f"{self.short_name}:{language.slug}", news, timeout=None)
```
jonbulz: we unconditionally set the cache here, and `news = []` is initialised before the loop. so if the tunews api is temporarly unavailable during our import run, we wipe the existing cache with an empty list, and we don't trigger a cache miss anymore after that because the empty list is a hit. would be more stable to only set the cache if the import was a success

integreat_cms/news_managers/tunews_manager.py
```diff
@@ -42,54 +43,62 @@ def import_news_items(self) -> None:
...
+            news = []
+            page_count = 1
+            while True:
+                try:
+                    response = requests.get(
+                        f"https://tuenews.de/wp-json/wp/v2/posts/?lang={language.slug}&per_page=100&page={page_count}&after={oldest_date}",
+                        timeout=10,
+                    )
+
+                    if response.status_code != 200:
```
jonbulz: we assume `status_code != 200` on a page other than page 1 means we successfully reached the end of the feed. a 500 or something else halfway through would mean we silently drop a portion of the news. maybe checking if `len(posts) < per_page` would be a more stable condition? then we could treat any non-200 status code as a failure

integreat_cms/core/settings.py
```diff
@@ -275,6 +275,23 @@
     24,
 )
 
+#################
+# External News #
```
jonbulz, line 285: there already is an external news settings section at `:1565`

integreat_cms/core/settings.py
```diff
@@ -275,6 +275,23 @@
...
+#################
+
+#: How many days back news posts should be fetched from Tü News
+#: Default to FCM_HISTORY_DAYS
+TUNEWS_HISTORY_DAYS: Final[int] = int(
+    os.environ.get("INTEGREAT_CMS_FCM_HISTORY_DAYS", FCM_HISTORY_DAYS),
+)
+
+#: Disable re-importing of external news posts on demand e.g. in test context
+EXTERNALNEWS_DISABLE_AUTO_REIMPORT: bool = bool(
```
jonbulz, line 296: should be `Final[bool]`

integreat_cms/api/v3/news.py
```diff
@@ -75,11 +75,14 @@ def news(
...
         ]
 
-    page_num = request.GET.get("page", 1)
     try:
-        page_size = min(int(request.GET.get("size", 20)), 500)
+        page_size = min(int(request.GET.get("size", 0)), 500)
     except (TypeError, ValueError):
         page_size = 20
+
+    if not page_size:
```
jonbulz: I think the more consistent behavior would be to check
```
if "page" not in request.GET and "size" not in request.GET:
    return JsonResponse(list(sorted_result), safe=False)
```
Currently, a consumer that keys only by page gets the full list on every page=1, 2, 3... 
Also, the fallback is a bit inconsistend, because the `except` still sets page_size=20, so 
no size -> all
size = 0 -> all
size = 'something weird' -> 20

tests/news_managers/test_tunews_manager.py
```diff
@@ -112,12 +113,21 @@
...
+    response_2 = Mock()
+    response_2.status_code = 400
+
     with patch(
         "integreat_cms.news_managers.tunews_manager.requests.get"
     ) as fake_tunews_server:
-        fake_tunews_server.return_value.status_code = 200
-        fake_tunews_server.return_value.json.return_value = dummy_news_items
+        # prepare 2 responses for each language and each external news manager
+        fake_tunews_server.side_effect = [response_1, response_2] * (
```
jonbulz, line 133: ok, I had to spend some time to understand this test setup and I might get it wrong. but I think it is actually a coincidence this test passes as is, at least the amal news section.
first off, we test both, the amal news and tunews manager in `test_tunews_manager.py`, which is not ideal, but pre-existing. the patch of `integreat_cms.news_managers.tunews_manager.requests.get` actually patches **both**, the amal news and tunews import, because they share the requests module. otherwise, the amal news manager would hit the live endpoint.
but that also means that they both consume the same queue of fake news responses, which is alternating response 200, 400, 200, 400 and so on... tunews_manager does 2 calls per language, so after it completes, we still have a chain of 200, 400, 200, 400, ...
amal news then only does one request per language, so half the languages get a 200 and half get a 400, and which half depends on the language ordering. we only check if the `de` import was successful, but that only works by coincidence, because german happens to sit on an even index in the language list. before the changes on this branch, every mock response was always a 200, but now we converted it to a side effect that makes all of this very brittle. I think now is the time to move the amal news tests to a dedicated module, otherwise it's only a matter of time until we see some weird flaky side effects of these tests

integreat_cms/core/circleci_settings.py
```diff
@@ -44,6 +44,8 @@
 LINKCHECK_DISABLE_LISTENERS = True
 # Disable background tasks during testing
 BACKGROUND_TASKS_ENABLED = False
+# Disable auto re-import of external news on demand
+EXTERNALNEWS_DISABLE_AUTO_REIMPORT = True
```
jonbulz, line 44: this is a good addition, but local test runs never see this setting, because `test.sh` sources `integreat_cms.core.settings`. so local runs could still do on-demand imports against the live endpoints. we could set the env var in `test.sh`. adding a pytest fixture would be even more robust in case someone runs the tests directly and not through the script

jonbulz (changes requested):
Here's my re-review. Sorry, it's a lot 😅 
Thanks for taking this on, I think we are definitely on the right track!
But what needs fixing is the infinite loop if we receive `RequestExceptions`. The cache wipe on failed imports also doesn't sit right with me, and the tests are very brittle like this, see my comments. I wouldn't feel comfortable merging until that is addressed. 
The rest is a bit more nit-picky. Feel free to reach out in case you want to discuss anything! :)

integreat_cms/core/settings.py
MizukiTemma: Yes, I notice it too and it seems I fixed it during you were reveiwing 😅

integreat_cms/core/settings.py
MizukiTemma: 🙈 What I wanted to do was read `TUNEWS_HISTORY_DAYS` and use `FCM_HISTORY_DAYS` as default/fallback 🙈 Updates will follow

integreat_cms/core/settings.py
MizukiTemma: I would leave it as it, we ressign the variable in `integreat_cms/core/circleci_settings.py` (mypy noticed it 😅 )

integreat_cms/news_managers/tunews_manager.py
MizukiTemma: I decided to let it `break` without retrying, since there is an automatic re-import on demand.

integreat_cms/api/v3/news.py
MizukiTemma: Applied 👍

tests/news_managers/test_tunews_manager.py
MizukiTemma: You got it right, and that was messy and depended on the language order. So, the test was separated into two (very similar) tests respectively for Amal News and Tu News, and they should be less independent of language order. I hope this makes it better.

integreat_cms/news_managers/tunews_manager.py
MizukiTemma: Checking by `len(post)` is not so Handy as the Response with `status_code != 200` Looks like this and has `length=3`

{
  "code": "rest_post_invalid_page_number",
  "message": "Die angeforderte Seitennummer ist größer als die Anzahl der verfügbaren Seiten.",
  "data": {
    "status": 400
}

We could proceed with such result and reject them as malformed post but it doesn't look correct for me, and I couldn't find any source that error responses always have this structure or what variety of possible error messages they have. For me it doesn't look trivial to distinguish between responses with `status != 200` containing valid posts, reponses with `status != 200` with potential of next page with valid posts and other `status != 200` Responses. And...I'm not convinced it's worh it to build Rescue measure for such cases 🤔 Let me know if I misunderstood your intention.

MizukiTemma:
I'm investigating the faling tests

MizukiTemma:
@jonbulz 
The test is resolved. Ready to re-review 🙏

integreat_cms/api/v3/news.py
```diff
@@ -75,11 +75,14 @@ def news(
             result for result in sorted_result if result["source"] in sources
         ]
 
-    page_num = request.GET.get("page", 1)
     try:
-        page_size = min(int(request.GET.get("size", 20)), 500)
+        page_size = min(int(request.GET.get("size", 0)), 500)
```
jonbulz: Thanks for applying my suggestions! :) But like this, the default can't be 0, because `Paginator` can't handle `page_size=0'. Can't it just stay at 20?

integreat_cms/news_managers/tunews_manager.py
jonbulz: Well, yes and no 😅 I'll give you a code snippet to illustrate what I mean, not because I think you need to implement it line-by-line:
```
for language in Language.objects.all():
    news = []
    page_count = 1
    per_page = 100
    import_failed = False
    while True:
        try:
            response = requests.get(
            f"https://tuenews.de/wp-json/wp/v2/posts/?lang={language.slug}&per_page={per_page}&page={page_count}&after={oldest_date}", timeout=10,
        )
        except requests.exceptions.RequestException:
            logger.exception("Failed to fetch posts in %s.", language)
            import_failed = True
            break
        if response.status_code != 200:
            if response.status_code == 400 and page_count == 1:
                logger.debug("Tü News does not serve %s; caching empty result.", language)
            else:
                logger.error("Could not fetch page %s in %s.", page_count, language)
                import_failed = True
            break
        posts = response.json()
        ...
        if len(posts) < per_page:      # short page = last page
            break
        page_count += 1

    if not import_failed:
        cache.set(f"{self.short_name}:{language.slug}", news, timeout=None)
```
This way:
- we break before we reach the 400, and every failure is treated as a failure, so we don't override a good cached result with a partial one
- we still set the cache for languages for which there are genuinely no results

I won't insist on the first point, I think it is rather unlikely that happens. But the second one is important, because currently for a language that tunews does not serve, we have:
-> no posts in the response -> no cache is set -> every request to our news api in that language is a cache miss -> every cache miss triggers a re-import for **every** language

before the latest change, what we had was:
if tunews api is unavailable during import run -> empty result replaces good cached result

I think the former is worse than the latter, so if you think I'm over-engineering, I'd prefer you revert the latest change and we set the cache unconditionally 😅

jonbulz (changes requested):
Thanks again! Almost there 😅 
I tried to be more clear about what I mean, I hope that helps. Except for the last 2 comments, everything else is resolved now :)

integreat_cms/news_managers/tunews_manager.py
MizukiTemma: Aha 😸 That looks cool! I adoped your code just as is 😀

MizukiTemma:
@jonbulz Thank you for explanation 👍 Your sugegstions are applied :)

jonbulz (approved):
Thank you for sticking with me and applying all my suggestions! Looks good now :)

integreat_cms/news_managers/tunews_manager.py
```diff
@@ -42,54 +43,65 @@ def import_news_items(self) -> None:
...
+            import_failed = False
+            while True:
+                try:
+                    response = requests.get(
+                        f"https://tuenews.de/wp-json/wp/v2/posts/?lang={language.slug}&per_page={per_page}&page={page_count}&after={oldest_date}",
+                        timeout=10,
                     )
-                    continue
+                except requests.exceptions.RequestException:
+                    logger.exception("Failed to fetch posts in %s.", language)
```
hannaseithe: Maybe a more specific log would be more helpful here:
```suggestion
                    logger.exception("Failed to fetch posts from TuNews in %s.", language)
```

integreat_cms/news_managers/tunews_manager.py
```diff
@@ -42,54 +43,65 @@ def import_news_items(self) -> None:
...
+                if response.status_code != 200:
+                    if response.status_code == 400 and page_count == 1:
+                        logger.debug(
+                            "Tü News does not serve %s; caching empty result.", language
+                        )
+                    else:
+                        logger.error(
+                            "Could not fetch page %s in %s.", page_count, language
+                        )
+                        import_failed = True
```
hannaseithe, line 86:
```suggestion
                   import_failed = True
```
I think we should set import_failed=True for both cases of the if condition above

integreat_cms/news_managers/tunews_manager.py
```diff
@@ -42,54 +43,65 @@ def import_news_items(self) -> None:
...
                         news.append(self.transform_post(post))
                     except (KeyError, TypeError, ValueError):
                         logger.exception(
-                            "Malformed %s post (id=%s); skipped.",
-                            self.name,
+                            "Malformed Tü News post (id=%s); skipped.",
                             post.get("id"),
                         )
 
+                if len(posts) < per_page:
```
hannaseithe:
```suggestion
                if len(posts) <= per_page:
```
Otherwise we do an extra call. e.g.  `if  100 < 100` we start an extra loop run, get returned nothing and then all posts gathered are discarded

hannaseithe (changes requested):
Thank you, I have some small change requests.

integreat_cms/news_managers/tunews_manager.py
MizukiTemma: I think we now want to save the result of "No news for this language (though no error)" as cache to avoid endless re-import. (I should not have folded [this comment](https://github.com/digitalfabrik/integreat-cms/pull/4440#discussion_r3735799933)

integreat_cms/news_managers/tunews_manager.py
MizukiTemma: Hmm 🤔 We cannot then tell the end of page from other `400` errors and we end up saving posts regardless of it can be a partiall result or actually the complete one, or am I missing something? @jonbulz

integreat_cms/news_managers/tunews_manager.py
hannaseithe: I see, thank you for explaining. In that case, I would suggest actually decoding the reponse body instead of of relying on the `page_count == 1` condition. I checked the responses for both cases (this and regarding the comment below).  

One gives: 
```
{"code":"rest_invalid_param","message":"Ung\u00fcltige(r) Parameter: lang","data":{"status":400,"params":{"lang":"lang ist nicht eins von , de, uk, ar, en, fa und he."},"details":{"lang":{"code":"rest_not_in_enum","message":"lang ist nicht eins von , de, uk, ar, en, fa und he.","data":null}}}}
```

And the pagination issue gives
```
{"code":"rest_post_invalid_page_number","message":"Die angeforderte Seitennummer ist gr\u00f6\u00dfer als die Anzahl der verf\u00fcgbaren Seiten.","data":{"status":400}}
```

Therefore I would suggest this handling:
```
if response.status_code != 200:
      body = response.json()
      code = body.get("code")
      if code == "rest_invalid_param" and "lang" in body.get("data", {}).get("params", {}):
          # Language genuinely not served by Tü News, not a failure:                                                                                                                            
          # cache the (empty) result instead of retrying every request. 
          logger.debug("Tü News does not serve %s; caching empty result.", language)
      elif code == "rest_post_invalid_page_number":
          # No more pages, not a failure: keep the posts gathered so far.
          logger.debug("Reached end of pagination for %s.", language)
      else:
          logger.error("Could not fetch page %s in %s (code=%s).", page_count, language, code)
          import_failed = True
      break
```
p.s. I especially care about the comments here, so the next dev would be saved from the same confusion as I ran into

integreat_cms/news_managers/tunews_manager.py
hannaseithe: Thank you for pushing back, my suggestion makes no sense, I was only thinking about when we reach the end of all posts, not about what happens when we are just breaking for a new page. But I still think this is a real bug (even if we assume it to be an edge case), therefore I suggested my handling above.

integreat_cms/news_managers/tunews_manager.py
MizukiTemma: Thank you for detailed suggestion. The comments are definetely helpful for us in future 😁 

I'm personally not so happy to distinguish cases by the code string but let's just do so, assuming they rarely change them 😸

MizukiTemma:
@hannaseithe 
Thank you for suggestions and reply. It's ready for re-review 🙏

hannaseithe (approved):
This looks good to me. Thank you Mizuki!

PR #4424, opened 2026-07-10
Update dependencies

timobrembeck:
### Short description

Bumps CI to Python 3.13 / PostgreSQL 17 (cherry-picked from #4307), updates **all** Python and npm dependencies to their most recent versions and adapts the code base to the new majors. This closes all open Python security alerts and most npm ones (see below).

### Proposed changes

**Adaptations to new major versions**

- **django-treebeard 5**: `PageManager` now subclasses `NS_NodeManager`; the `Node._get_database_cursor` monkeypatch in `tree_mutex` is gone (treebeard no longer issues raw, non-transactional SQL — `transaction.atomic` alone suffices, and the "corrupt the tree without the mutex" test can no longer provoke corruption and was removed); the reworked `MoveNodeForm` field names/semantics (`treebeard_ref_node`/`treebeard_position`, ModelChoiceField validated against a region-limited queryset) are reflected in the page & language-tree forms, templates, `page-order.ts` and test data. The now-unneeded `get_private_member` template filter was removed.
- **pytest 9 / pytest-django 4.12**: `pytest-circleci-parallelized` is abandoned and incompatible — replaced by CircleCI-native test splitting (`circleci tests glob | circleci tests split --split-by=timings --timings-type=filename`, junit XML with `junit_family=xunit1`), so no root `conftest.py` plugin is needed at all. All transactional tests now use `serialized_rollback` uniformly, since with file-based splitting a plain `transaction=True` flush before a `serialized_rollback` restore violates the content-type unique constraint.
- **Sphinx 9**: adapted the apidoc patch, guarded sphinxcontrib-django's docstring hook against PEP 695 type-parameter bounds (the plain `Model` bound of `ShadowInstance[T: Model]` crashed the build), fixed treebeard 5's malformed `tree_model` docstring via an autodoc hook and disambiguated an `object` annotation. Docs build with zero warnings again.
- **djlint 1.40**: reformatted templates, fixed a real `{% endblock %}` name mismatch in `poi_view.html` and extended the `djlint:off` rules in `page_pdf.html` for the new H007 rule.
- **icalendar 7**: `event.decoded()` returns `str` for text properties; `interval` annotated as plain `int`.
- **lxml 6**: markup-less content is parsed into a `<span>` instead of a `<p>` — `clean_content` restores the block-level wrapping (keeps machine-translated plain-text content unchanged).
- **reportlab 4.5**: wraps over-long unbreakable words instead of truncating with an ellipsis — reference PDF regenerated.
[...]

PR #4221, opened 2026-03-26
Enhancement/refactor tests

jonbulz:
### Short description

Refactoring of our test suite to improve test reliability and execution time. Adds a comprehensive guide on how to use the introduced arguments and how to add new tests.

#### Unify CI setup and local runs
- introduce `test_settings` module to export the same settings to `circleci_settings` and `test.sh`, eliminating duplicated dummy API keys and feature flags
- explicitly export `DJANGO_SETTINGS_MODULE` in `test.sh` after `require_database`, which was overriding with the base settings
#### Fix flaky tests and eliminate ordering dependencies
- remove the complex relative after_tests tuples and inter-test dependencies. Retain a simple `@pytest.mark.order("last")` on all transactional tests (needed because the --circleci-parallelize plugin reorders tests by module after pytest-django's collection).
- testmon + xdist race condition — `--testmon-noselect` conflicted with parallel workers, causing sporadic `User.DoesNotExist` errors during fixture setup. Fixed by disabling testmon in parallel mode (`-p no:testmon`). 
- POI category test bug — `test_edit_poi_category_was_successful` was asserting the wrong category name in the success message, and the session-scoped client's message cookie would overflow in the full parallel suite. Fixed by checking database state directly instead of HTML messages.
- re-enable tow HIX tests that were skipped by converting the session-scoped dummy_region fixture to function-scope
- remove `serialized_rollback=True` from django_db markers. it is no longer needed because ordering dependencies are eliminated, and removing this argument significantly speeds up transactional tests
- mark `test_rule_out_false_positive` as `xfail(strict=False)` because it is non-deterministic by nature
#### Improve test performance
- introduce unit and slow test markers
- add `QUICK_ROLES` for faster local iteration
- auto-mark all status-code view tests as slow via directory-level conftest.py
#### Improve test infrastructure
- add `tests/factories.py` for cleaner test setups
- extract role constants into `tests/constants.py` (importable without triggering fixture registration side effects)
- allow to update snapshots in a single test run with `update_snapshots` argument
#### Documentation
- add comprehensive testing guide in `docs/src/testing.rst`

### Side effects

- Introduces test coverage threshold. If we fail to add tests for new features, the pipeline will fail in the future

### Faithfulness to issue description and design

There are no intended deviations from the issue and design.

### How to test

Run the test suite on your local machine. See that tests are now deterministic. Try the different options and see that they reduce test execution time:
- `./tools/test.sh -m unit` — unit tests only (seconds)
[...]

PR #4441, opened 2026-07-18
Refactor test suite

### LINKED ISSUE #4293: [Task]: Make tests more useful and reliable
### What: 
Unify the test setup so local runs match CI, eliminate test-ordering dependencies, and make the suite faster and more useful for everyday development.
### Why: 
Tests were brittle and environment-coupled — they passed in CI but flaked locally (and vice versa), required scattered env-var setup, mixed transactional and non-transactional tests in unsafe orders, and offered no way to run a fast subset during iteration.

### Task Category

  - Consolidate test-only config (dummy API keys, disabled listeners, background tasks off) into integreat_cms/core/test_settings.py instead of tools/test.sh exports + circleci_settings.py duplication.
  - Introduce unit and slow pytest markers; document a 3-tier model (unit / standard integration /transactional).
  - Restore @pytest.mark.order("last") on every transaction=True test to defeat the --circleci-parallelize reordering that mixes them with savepoint tests.
  - Add QUICK_ROLES=1 mode (4 representative roles instead of 11) for fast local parametrized runs.
  - Add tests/factories.py (factory functions with sensible defaults) and tests/constants.py (role identifiers / groups) so new tests don't import from conftest.py.
  - Forward unknown long flags (--update-snapshots etc.) from tools/test.sh to pytest; fix snapshot-writer indentation to match existing 2-space convention.
  - Harden infra around tools/test.sh: skip Docker fallback when INTEGREAT_CMS_DB_PORT ≠ 5432; propagate INTEGREAT_CMS_DOCKER_LISTEN_PORT → INTEGREAT_CMS_DB_PORT when Docker is used.
  - Disable testmon under pytest-xdist (root cause of sporadic User.DoesNotExist during fixture setup).
  - Stop AbstractContentTranslation.get_repr from issuing extra queries (reduces N+1 noise in API tests).
  - Add a "Running Tests" / "Markers" / "Fixtures" / "Pitfalls" section to docs/src/testing.rst.

### Acceptance Criteria

  - ./tools/test.sh passes on CI and on a clean local checkout without per-developer env-var setup.
  - ./tools/test.sh -m unit runs in ~1 s and touches no database.
  - ./tools/test.sh -m "not slow" skips the 16 test_view_status_code_*.py files and region-duplication tests.
  - QUICK_ROLES=1 ./tools/test.sh runs with 4 roles only.
  - ./tools/test.sh -v -k test_api_result --update-snapshots regenerates snapshots without errors.
  - No test depends on the order in which pytest collects its peers (verified by running with -p no:randomly and a random seed).
  - New testing docs are published from docs/src/testing.rst.

### Affected files / modules
  - integreat_cms/core/test_settings.py (new) — single source of truth for test settings
  - integreat_cms/core/circleci_settings.py — slimmed down, inherits from test_settings
  - tests/conftest.py — reverted django_db_setup override that broke CI fixture loading; cleaned up fixture scopes
  - tests/factories.py (new), tests/constants.py (new)
  - tools/test.sh, tools/_functions.sh — env propagation, pytest passthrough, testmon/xdist guard
  - pyproject.toml — register unit / slow markers; reduced coverage gate
  - integreat_cms/cms/models/abstract_content_translation.py — query-free get_repr
  - All tests/**/test_* files using transactional DB now carry @pytest.mark.order("last")

 ### Commands
  `./tools/test.sh -m unit`                       # fast unit tier
  `./tools/test.sh -m "not slow" `                # skip parametrized view tests
  `QUICK_ROLES=1 ./tools/test.sh`                # 4 roles instead of 11
  `./tools/test.sh -v -k test_api_result --update-snapshots`

### Risks & Impact

_No response_

### Target Environment

_No response_

### Verification / Testing

_No response_

### Additional Context

- Residual environment-coupled failures still exist on some local boxes (PDF rendering byte-equality, one exact assert_num_queries count) — tracked separately and not blockers for this task.
  - Translation .mo files are still not committed; if test_csv_export_feedback fails locally, run `./tools/translate.sh` first.
  - See docs/src/testing.rst for the full developer-facing write-up.

- #2711 
- #3777 
- #4181 

### Definition of Done

- [x] Work is complete and meets acceptance criteria
- [x] Code reviewed / peer reviewed (if applicable)
- [x] Tests added or updated (if applicable)
- [x] Documentation updated (if applicable)

timobrembeck:
### Short description

Refactoring of our test suite to improve reliability and execution time: unified test settings, deterministic test ordering, `unit`/`slow` markers, and a dockerized test runner that mirrors CI. Continues @jonbulz's work in #4221 (supersedes it), rebased onto `update/dependencies` (#4424) with all review findings addressed.

#### Unify CI setup and local runs
- introduce a `test_settings` module used by both `circleci_settings` and `test.sh`, eliminating duplicated dummy API keys and feature flags
- never use redis/cacheops in test runs: force `LocMemCache` and disable the cacheops entry points in the test settings — this was the root cause of the sporadic "Expected N queries but fewer were done" failures reported in #4221

#### Run tests in Docker by default
- `./tools/test.sh` now runs the test suite inside a container (`tools/docker/Dockerfile.test` + `docker-compose.test.yml`) that mirrors the CI images (Python 3.13, PostgreSQL 17), so "works locally, fails in CI" discrepancies disappear; pass `--local` to run on the host as before
- the container runs as the invoking host user (no root-owned files in the working tree), compiles only the project's own translations, and ensures the webpack bundle exists before running view tests
- Docker compose setup is decoupled from the local `.env` file; a custom `INTEGREAT_CMS_DB_PORT` no longer spins up a broken postgres container

#### Fix flaky tests and eliminate ordering dependencies
- remove the fragile relative `after=` ordering tuples and `serialized_rollback=True` from `django_db` markers; retain a simple `@pytest.mark.order("last")` on all transactional tests
- clear caches between tests via an autouse fixture (locmem rate-limit counters leaked across tests); restore the `clean_news_cache` fixture
- run `--changed` tests serially, since testmon conflicts with pytest-xdist workers
- fix `test_edit_poi_category_was_successful` to assert database state instead of an overflowing message cookie; re-enable two skipped HIX tests by converting the session-scoped `dummy_region` fixture to function scope; mark the inherently non-deterministic `test_rule_out_false_positive` as `xfail(strict=False)`
- avoid database hits in `get_repr` (noisy DB-blocker stack traces during logging)

#### Improve test performance
- introduce `unit` and `slow` markers; auto-mark all status-code view tests as slow
- `QUICK_ROLES=1` runs view tests with only 4 representative roles (works in Docker mode too, parsed as a proper boolean)
- run pytest-xdist inside the CI test containers and halve the CircleCI shard count; drop the redundant migrate step from the test job

#### Coverage gate
- introduce a test coverage threshold, enforced only on full test runs (filtered runs like `-m unit` or `-k` are exempt, so they don't fail on partial coverage)

#### Test infrastructure & documentation
- add `tests/factories.py` for cleaner test setups and extract role constants into `tests/constants.py` (importable without fixture registration side effects)
- allow updating snapshots in a single run via `--update-snapshots` (with fixed indentation of generated snapshot files)
- add a comprehensive testing guide in `docs/src/testing.rst`, covering the Docker test mode, the coverage gate and the test ordering rules

### Side effects

- `./tools/test.sh` now requires Docker by default — contributors who want to keep running on the host need to pass `--local`
- the coverage threshold makes the pipeline fail in the future if we add features without tests
- `get_repr` now prints foreign-key ids instead of resolving related objects, so some log lines change format slightly

### How to test

Run the test suite and see that tests are deterministic and match the CI environment:
- `./tools/test.sh` — full suite in the CI-mirroring Docker container (default)
- `./tools/test.sh --local` — full suite on the host as before
- `./tools/test.sh -m unit` — unit tests only (seconds)
- `./tools/test.sh -m "not slow"` — skip the slow parametrized view tests
- `QUICK_ROLES=1 ./tools/test.sh` — test only 4 representative roles
- `./tools/test.sh --changed` — only tests affected by your changes (serial)
- `./tools/test.sh --update-snapshots` — regenerate stale snapshots in one run

### Resolved issues

Fixes: #4293


jonbulz:
tested locally, but re-converted to draft until #4424 is merged

.circleci/config.yml
hannaseithe, line 251: Maybe we should add a comment here that reminds the devs to keep the config in sync with docker-compose.test.yml (also in line 223 for check-migrations?)

docs/src/testing.rst
```diff
@@ -17,9 +17,285 @@ For more information, see :doc:`django:topics/testing/index` and :doc:`django:to
...
+   transactional test (which flushes the database) before non-transactional
+   ones. ``@pytest.mark.order("last")`` from pytest-order re-enforces the
+   correct ordering after all collection hooks have run.
+
+2. **Don't use** ``serialized_rollback=True`` — it has FK ordering issues
+   with PostgreSQL. Use ``load_test_data_transactional`` instead.
+
+3. **Don't modify session-scoped data** in non-transactional tests — the
+   session-scoped fixtures are shared across all tests in a worker. If your
+   test needs to modify data, use a transactional test.
```
hannaseithe: I did not understand this part from the get go - I had to ask claude for an example. As I understand this mainly applies to in-memory state, which gets shared between non-transactional tests, correct? Maybe an example would help here to make things clearer

tests/core/management/commands/test_fix_internal_links.py
hannaseithe:
```suggestion
@pytest.mark.django_db(transaction=True)
```
This test is actually failing for me locally (with docker)

```
ERROR tests/core/management/commands/test_fix_internal_links.py::test_fix_internal_links_commit_skips_version_conflicts - django.db.utils.IntegrityError: duplicate key value violates unique constraint "django_content_type_app_label_model_76bd3d3b_uniq"
```

tests/factories.py
```diff
@@ -0,0 +1,187 @@
+"""
```
hannaseithe: So far we are not using those factories anywhere. As I suggested before in an earlier PR. We have examples for that in conftest.py still `create_page` and `create_language`. So I suggest using that pattern here and then also using these factory fictures at least where those two had been used

hannaseithe (changes requested):
First of all: THANK YOU for this PR - so so much. Both to @timobrembeck and @jonbulz . This is such a improvement towards code quality / maintainability - it makes me genuinely happy. 

I have minor suggestions in the comments. But a few more points here:

1) I think this would be a valid place to add an ADR for the decision to run the tests inside docker? One could argue this is more about infrastructure, but I feel an ADR might come in handy here in the future

2) Overall I have felt the whole time that we overly rely on test_data.json and its fixtures. This of course is another architecture decision, but I would want to encourage people at least with newer tests to rely on the fixture-factories rather (if possible), than adding new data into test_data.json. Maybe that is a topic for issue discussion...

3) "not slow" is actually pretty slow. Executing all test takes about 6 minutes on my machine, and when I ran "non slow" those seemed to be about half of all tests and they taook 5 minutes!

4) I ran with --changed twice - without any code change in between - and it re-ran all tests both times - I would have expected no tests to be run, the second time

------

I am not familiar with bash-script, so I only skimmed through those files

.circleci/config.yml
timobrembeck: Added, on both the `test` and `check-migrations` jobs.

docs/src/testing.rst
timobrembeck: You're right, what survives is the Python object a session-scoped fixture returns, most importantly the cookies of the shared `login_role_user` client. I added an example to the docs.

tests/core/management/commands/test_fix_internal_links.py
timobrembeck: Correct, and already fixed in 553e4bcc — that was the last surviving `serialized_rollback=True`, which replays the serialized fixture contents on top of an already populated database (hence the duplicate `django_content_type` row). Could you re-run that file to confirm it's green on your machine too?

tests/factories.py
timobrembeck: Fair point, the `create_page` and `create_language` fixtures were duplicating `make_page`/`make_language` anyway, so they're gone and their tests now call the factories directly. Do you think this is sufficient for now, or should I have a look at other tests as well?

timobrembeck:
@hannaseithe Thanks for your in-depth review! :muscle: 

> 1. I think this would be a valid place to add an ADR for the decision to run the tests inside docker? One could argue this is more about infrastructure, but I feel an ADR might come in handy here in the future

ADR: I'd rather not. `ADR/README.md` scopes these to architecture decisions, "but not implementation details", and the three existing ones are all about the product's architecture. How we run the tests locally is dev tooling, and the rationale (match CI by construction) is recorded in `docs/src/testing.rst`. Happy to write one anyway if you feel strongly.

> 2. Overall I have felt the whole time that we overly rely on test_data.json and its fixtures. This of course is another architecture decision, but I would want to encourage people at least with newer tests to rely on the fixture-factories rather (if possible), than adding new data into test_data.json. Maybe that is a topic for issue discussion...

Fully agree on `test_data.json`, and that's exactly what `tests/factories.py` is for. Let's track it in a separate issue; this PR only adds the tooling for it.

> 3. "not slow" is actually pretty slow. Executing all test takes about 6 minutes on my machine, and when I ran "non slow" those seemed to be about half of all tests and they taook 5 minutes!

Fixed, at least partly: filtered runs were still instrumented with coverage (`-m`/`-k` only got `--cov-fail-under=0`, not `--cov` removed). They now run without the coverage tracer at all. Beyond that, the `slow` marker currently only covers the status-code view tests, the region duplication tests and `test_tree_mutex`, so a lot of genuinely slow tests are still in the "not slow" set. Also, tests that do external requests might take longer based on the network connection and external services... I'm currently on a low spec machine and poor internet connection, so I'm unable to do a proper performance evaluation and provide a good baseline for which tests are really slow and which aren't.

> 4. I ran with --changed twice - without any code change in between - and it re-ran all tests both times - I would have expected no tests to be run, the second time

Good catch, this was broken. `--changed` passed `--testmon-nocollect`, which never writes back what testmon learned, and this PR additionally disabled testmon for normal (parallel) runs. So the dependency database was written exactly once by your very first `--changed` run, and then frozen. On top of that, a `.testmondata` collected in a different environment (your old host runs vs. the new docker runs) makes testmon report "system packages changed" and select everything, with no way to ever recover. `--changed` now collects while it selects, and docker and `--local` runs keep their data in separate namespaces. If you still have the old `.testmondata` lying around, the first run after this will be a full one and every run after that should be short.

tests/factories.py
hannaseithe: Sorry, my comment was incomplete. What I was also trying to get at is, that I would rather have those factories be fixtures than simple functions that we import. Similar to the pattern of that `create_page` fixture. I think having them as fixtures would make it easier for the devs to use. But I leave it up to you. I can live with them as they are as well.

EDIT: And to answer your question: I have just greped for ".create(" in our tests folder and from a short glance I see quite some candidates that could use the factories. But maybe this could be done in a follow up task

tests/core/management/commands/test_fix_internal_links.py
hannaseithe: Is green now

docs/src/testing.rst
hannaseithe: I can follow that explanation now. Thank you :)

.circleci/config.yml
hannaseithe: Thank you!

hannaseithe:
> ADR: I'd rather not. `ADR/README.md` scopes these to architecture decisions, "but not implementation details", and the three existing ones are all about the product's architecture. How we run the tests locally is dev tooling, and the rationale (match CI by construction) is recorded in `docs/src/testing.rst`. Happy to write one anyway if you feel strongly.

Yeah I get your point. And since we have the record in the documentation, I hope that is going to be enough. Let's leave it as is then.

> Fully agree on `test_data.json`, and that's exactly what `tests/factories.py` is for. Let's track it in a separate issue; this PR only adds the tooling for it.

If you want I can create the issue


> Fixed, at least partly: filtered runs were still instrumented with coverage (`-m`/`-k` only got `--cov-fail-under=0`, not `--cov` removed). They now run without the coverage tracer at all. Beyond that, the `slow` marker currently only covers the status-code view tests, the region duplication tests and `test_tree_mutex`, so a lot of genuinely slow tests are still in the "not slow" set. Also, tests that do external requests might take longer based on the network connection and external services... I'm currently on a low spec machine and poor internet connection, so I'm unable to do a proper performance evaluation and provide a good baseline for which tests are really slow and which aren't.
> 

Also I want to add that I observed later that the time of execution for a normal run (tools/test.sh) wildly fluctuated on my machine yesterday (between 6 minutes and 11 minutes). So might have also been due to that


> Good catch, this was broken. `--changed` passed `--testmon-nocollect`, which never writes back what testmon learned, and this PR additionally disabled testmon for normal (parallel) runs. So the dependency database was written exactly once by your very first `--changed` run, and then frozen. On top of that, a `.testmondata` collected in a different environment (your old host runs vs. the new docker runs) makes testmon report "system packages changed" and select everything, with no way to ever recover. `--changed` now collects while it selects, and docker and `--local` runs keep their data in separate namespaces. If you still have the old `.testmondata` lying around, the first run after this will be a full one and every run after that should be short.

Beautiful! Works now

jonbulz (approved):
Thank you so so much for working on this! 😊 
Tested locally and as far as I can tell, this seems to solve all the ordering-dependency and environment-related issues. I only have two small suggestions, both none blocking, so I'll approve already :)
I'd include these into `docker-compose.test.yml`:
```
INTEGREAT_CMS_PDF_ROOT: /tmp/pdf
INTEGREAT_CMS_LOGFILE: /tmp/integreat-cms.log
```
for the following reasosn:
- `INTEGREAT_CMS_PDF_ROOT`: nice to have to avoid comparing to a stale pdf, which essentially saves us the occasional `./tools/prune_pdf_cache.sh`
-  `INTEGREAT_CMS_LOGFILE`: log hygiene, and a small performance improvement due to less crosses to the bind mount

hannaseithe (changes requested):
Sorry for the delay, I thought I had already re-reviewed. All looks good to me. Thank you very much for implementing my change requests.
The only thing I would still like changed (or at least discussed) is that comment I left under the factories topic. I missed mentioning that on my first review.

jonbulz:
@hannaseithe I re-worked the factories into fixtures, as you suggested. let's get this thing merged 💪 
@timobrembeck I squashed some of my commits, but left yours untouched. I remember you saying you preferred it that way somewhere

hannaseithe (approved):
This looks good to me! I am really excited

PR #4445, opened 2026-07-20
4300: Add published date to endpoints

### LINKED ISSUE #4300: Add published date to endpoints
### User Story

As a user of the integreat-app I want to see on one glance which events/news/categories/places are new such that I can check those new documents for relevant info for myself.

### Problem Description

In order to implement this in the integreat-app (currently planned: events in https://github.com/digitalfabrik/integreat-app/issues/2112 and news in https://github.com/digitalfabrik/integreat-app/issues/4043) we need to include the actual time of when the document/document translation got published to decide if a document is "new". 

### Acceptance Criteria

- For pages, events and places (locations) the first publication date can be stored
- In the endpoins the value of this new field is delivered as `published_at` for Events, places and pages
- For local news (push news) use the date of `sent_date` as `published_at` too
- For existing events, places and pages, the new field is filled by the date of last update
- API documentation is updated
- Test is adjusted (ggf also added)

Following endpoints are targeted: `pages/`, `locations/`, `events/`, `page/`, `children/`, `parents/`, `fcm/`, `news/`, `news/<news_id>/`

### Functional Criteria
- Add a `published` or `publishedAt` ISO 8601 property to the events in the events endpoint which resembles the point in time the event translation was published
- It would be really nice to consistently include this to all endpoints (pages, places, news)
- For local news `sent_date` already exists which is apparently the same, but it would be really nice if we could keep it consistent and rename to/add `published`/`publishedAt` as well here
- For tuNews and amalNews it should also be added
- Since we don't know the published date anymore for already published documents, I'd propose to just set it to `lastUpdate` in those cases or some time before. Please avoid the property being null.

### Error & Edge Case Handling

_No response_

### User Flow

_No response_

### Technical Notes
We currently do not save the first date and time of publication. It is not always equal to the creation timestamp. So we have to add a new field to `Event` model to store when the event translation fo the root language went online (published) for the first time.
For existing events this field can be populated with lastUpdate

### Out of Scope

_No response_

### Testing Guidance

_No response_

### Additional Context

_No response_

### Related Issues

Blocks: https://github.com/digitalfabrik/integreat-app/issues/2112, https://github.com/digitalfabrik/integreat-app/issues/4043

### Definition of Ready

- [x] Acceptance criteria are clear and testable
- [ ] UI design / mockups are available (if applicable)
- [ ] Dependencies are identified and unblocked
- [ ] Story has been refined / estimated by the team
- [x] Edge cases and error states have been considered

bahaaTuffaha:
### Short description
As a user of the integreat-app I want to see on one glance which events/news/categories/places are new such that I can check those new documents for relevant info for myself. We need to include the actual time of when the document/document translation got published to decide if a document is "new".

### Proposed changes

- Added `published_at` field to `AbstractContentTranslation` as a nullable `DateTimeField` with `verbose_name="publication date"`. And `None` at the `copy()` method.
- Added `published_at` key to `transform_event_translation`, `transform_page`, and `transform_poi_translation`, falling back to `last_updated` when `published_at` is unset.
- Updated `transform_notification` and `transform_notification_for_fcm` in `pushnews_manager.py` to include `published_at`, using `sent_date` with fallback to `last_updated`.
- Updated `transform_post` in `tunews_manager.py` and `amalnews_manager.py` to include `published_at`.
- Updated all API expected test output JSON files to include `published_at`.
- Created a data migration `0157` to populate `published_at` for existing content using the earliest public translation's `last_updated`.
- Adjusted the existing `0155` migration to lowercase slugs via the historical models instead of the live model, so the new translation field does not break a fresh migration.
- Created a test for publication date covering the draft, non-default-language, first-publish, and re-publish cases.
- Documented the new `published_at` field across the pages, locations, events, and news schemas in `api-docs.rst`.

### Side effects

- Adjusted `0155` migration to use historical models so fresh database migrates do not fail after `published_at` is added in a later migration.

### How to test

- Launch the docker and run the cms server.
- open Postman or any alternative and try to query a Get request for ex:
```
- Events: http://localhost:8000/api/v3/augsburg/de/events/
- Pages: http://localhost:8000/api/v3/augsburg/de/pages/
- Locations: http://localhost:8000/api/v3/augsburg/de/locations/
```

### Resolved issues

Fixes: #4300


integreat_cms/cms/models/abstract_content_translation.py
```diff
@@ -697,6 +697,13 @@ def save(self, *args: Any, **kwargs: Any) -> None:
         if kwargs.pop("update_timestamp", True):
             self.last_updated = timezone.now()
         super().save(*args, **kwargs)
+        if (
+            self.status == status.PUBLIC
+            and not self.foreign_object.first_published_at
+            and self.language == self.foreign_object.region.default_language
+        ):
+            self.foreign_object.first_published_at = self.last_updated
+            self.foreign_object.save(update_fields=["first_published_at"])
```
MizukiTemma:
```suggestion
        if (
            self.status == status.PUBLIC
            and not self.foreign_object.first_published_at
        ):
            self.foreign_object.first_published_at = self.last_updated
            self.foreign_object.save(update_fields=["first_published_at"])
```

Itdoesn't have to be the first publication in the default language, it can be in any language, right? @steffenkleinle

integreat_cms/cms/migrations/0158_populate_first_published_at.py
```diff
@@ -0,0 +1,51 @@
...
+from django.db.models import OuterRef, Subquery
+
+from ..constants import status
+
+if TYPE_CHECKING:
+    from django.apps.registry import Apps
+    from django.db.backends.base.schema import BaseDatabaseSchemaEditor
+
+
+def populate_first_published_at(
```
MizukiTemma: You can squash the both migration files into one. `AddField` and `populate_first_published_at` can be done one after the other like in 0138_configurable_mt_budget.py ✏️

MizukiTemma (changes requested):
Thank you so much for such a large work 💪 😸 

Please find my suggestions below.

integreat_cms/cms/models/abstract_content_model.py
```diff
@@ -133,6 +133,11 @@ class AbstractContentModel(AbstractBaseModel):
         default=timezone.now,
         verbose_name=_("creation date"),
     )
+    first_published_at = models.DateTimeField(
```
andrew8er: 💭 Since the API field is named `published_at`, wouldn't it be more consistent to name this column the same way? The "first" seems a bit redundant anyway.

integreat_cms/cms/models/abstract_content_translation.py
```diff
@@ -697,6 +697,9 @@ def save(self, *args: Any, **kwargs: Any) -> None:
         if kwargs.pop("update_timestamp", True):
             self.last_updated = timezone.now()
         super().save(*args, **kwargs)
+        if self.status == status.PUBLIC and not self.foreign_object.first_published_at:
+            self.foreign_object.first_published_at = self.last_updated
+            self.foreign_object.save(update_fields=["first_published_at"])
```
andrew8er: ❓ So saving a (public) translation sets the `first_published_at` field to the translation's `last_updated` field? Wouldn't that create a bit of a confusion? The translations last update date is mostly unrelated to the publication date IMHO, and a publication date can legally be a really crucial piece information. I would rather leave it unset for cases we have no data for.

andrew8er (commented):
LGTM, a small question.

dkehne:
Opus 5 opinion on that PR:

Nice touch that the backfill uses the earliest PUBLIC version rather than the blunt "last update" the AC asked for.

One question I'd settle before the details, since it may reshape the implementation: `published_at` is per content object, so it's identical for every language. A page published in German in 2019 whose Arabic translation goes live in 2026 still reports 2019 on `/api/v3/<region>/ar/pages/` — so the audience the "what's new" signal exists for won't see it as new. The issue text talks about when the *translation* was published. Is per-object the intended contract, or should this sit on `AbstractContentTranslation`?

Two things I stumbled over in the current implementation:

- `copy()` resets the field, but region duplication doesn't go through it: `region_form.py:1029` deepcopies and resets only `pk`/`tree_id`/`api_token` (same at `:1140` for the imprint). A duplicated region would inherit the source's date — and because the hook checks `not self.foreign_object.first_published_at` (`abstract_content_translation.py:700`), publishing there later won't correct it, so the wrong date sticks permanently.
- `test_data.json` has no `first_published_at`, so the column is NULL throughout the test DB and the updated expected-outputs only exercise the `or …last_updated` fallback. As far as I can tell the API tests would still pass if the field were never written at all — adding it to the fixture would give the feature its first end-to-end coverage.

Smaller: the migration updates all rows of four content tables in one transaction (a `first_published_at__isnull=True` filter would shorten the locks during deploy), and the docs say "published for the first time" while the API silently substitutes `last_updated` when the field is NULL — worth documenting the fallback or allowing `null`.

integreat_cms/cms/models/abstract_content_model.py
bahaaTuffaha: `published_at` in the API isn't the same value as the column. It falls back to `last_updated` when the column is empty, so giving them the same name would suggest that they always match. 🤔

integreat_cms/cms/models/abstract_content_translation.py
bahaaTuffaha: Screenshot 2026-07-27 at 7 06 28 PM

It was stated in the criteria to use the `last_updated`...

bahaaTuffaha:
> One question I'd settle before the details, since it may reshape the implementation: `published_at` is per content object, so it's identical for every language. A page published in German in 2019 whose Arabic translation goes live in 2026 still reports 2019 on `/api/v3/<region>/ar/pages/` — so the audience the "what's new" signal exists for won't see it as new. The issue text talks about when the _translation_ was published. Is per-object the intended contract, or should this sit on `AbstractContentTranslation`?

- I think I can make it per-translation instead of per-object so it would move to `AbstractContentTranslation`.
-  @steffenkleinle said "I am more than happy if we only have one first publication date which is the same for all languages! So yes, go for the root language for all languages." 😆 so am not sure if he has different opinion on this.

integreat_cms/cms/models/abstract_content_model.py
MizukiTemma: @bahaaTuffaha You can decide as PR authour 🟢

integreat_cms/cms/models/abstract_content_translation.py
MizukiTemma: I meant (or I understood) this statement as request/specification of a migration for all contents existing at the time of deployment of this PR.

But anyway, `self.last_updated = timezone.now()` in this code block, so `first_published_at` will be set to `timezone.now()` and in my opinion this is so wished. To avoid confusion, it can be but changed to `self.foreign_object.first_published_at = timezone.now()`

bahaaTuffaha: > But anyway, `self.last_updated = timezone.now()` in this code block, so `first_published_at` will be set to `timezone.now()` and in my opinion this is so wished. To avoid confusion, it can be but changed to `self.foreign_object.first_published_at = timezone.now()`

could it affect the `cleanup_autosaves` at `abstract_content_translation.py` ?

integreat_cms/cms/models/abstract_content_translation.py
steffenkleinle: Yes, I guess the first publication date in any language works :)

steffenkleinle:
I don't have the headspace right now to weigh the pros and cons of the different approaches. Feel free to discuss this yourself and come up with the best solution :) In the end, we need it to show users new events/news. You can decide if it makes more sense to have a per-language or per-object timestamp for that.

bahaaTuffaha:
> Opus 5 opinion on that PR:

@dkehne thank you for the heads up!

integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -1111,6 +1112,8 @@ def duplicate_page_translations(
         # Set the translation to draft if keep_status is false
         if keep_status is False:
             page_translation.status = status.DRAFT
+            # Reset the publication date because the duplicate has not been published yet
+            page_translation.published_at = None
```
MizukiTemma: I'd refresh the date always independent of `keep_status` 👀 because the content is new in the new region.

MizukiTemma (approved):
Thank you so much for updates 😸 
There is only one small suggestion left from my side, and I already approve the PR in advance ✅

integreat_cms/cms/migrations/0155_enforce_case_insensitive_slug_uniqueness.py
```diff
@@ -3,17 +3,55 @@
...
+from django.db.models.functions import Lower
 
 if TYPE_CHECKING:
     from django.apps.registry import Apps
     from django.db.backends.base.schema import BaseDatabaseSchemaEditor
 
 
-def forwards(_apps: Apps, _schema_editor: BaseDatabaseSchemaEditor) -> None:
-    make_all_slug_lowercase()
+def forwards(apps: Apps, _schema_editor: BaseDatabaseSchemaEditor) -> None:
```
seluianova, line 16: Does updating an old migration have any effect?
I thought it won't be reapplied anyway, unless the database is recreated from scratch.
Or am I missing something?

docs/src/api-docs.rst
```diff
@@ -249,6 +249,7 @@ RESPONSE
          "title": String,              // The title of the page
          "modified_gmt": String,       // Deprecated field
          "last_updated": String,       // When the page translation was last updated, in ISO 8601
+         "published_at": String,       // Publication date of this page translation, in ISO 8601
```
seluianova:
```suggestion
         "published_at": String,       // First publication date of this page translation, in ISO 8601
```
optional. just thought it might be not very obvious

integreat_cms/cms/migrations/0155_enforce_case_insensitive_slug_uniqueness.py
bahaaTuffaha: Yes this is correct but it could affect the CI test at `config.yml`  `name: Migrate database` L266 I think this test will migrate through them all.

bahaaTuffaha: Also migrate_database at `run.sh`. it goes through them on fresh DB.

seluianova (approved):
Tested everything except news (not sure how to configure them)
Looks good to me 👍

PR #4449, opened 2026-07-22
Add SSI content endpoints, fix #4448

### LINKED ISSUE #4448: [Task]: Content for Server Side Includes
### Task Description

Bot traffic / LLM agent is increasing. But most LLMs only get our static empty web app content. We need to extend our server side includes in the web app with actual page content. In the best case, this works for all content types.

### Task Category

Add a new view that can copy the news_social_media_headers view which returns the actual content.

### Acceptance Criteria

API endpoint that returns raw HTML content for at least pages. News, contacts, events should also be supported in the best case.

### Technical Details

_No response_

### Risks & Impact

_No response_

### Target Environment

_No response_

### Verification / Testing

_No response_

### Additional Context

_No response_

_No response_

### Definition of Done

svenseeberg:
### Short description

Add endpoints that return raw HTML of content models for web app server side includes.

### Proposed changes
- Add API endpoints that serve raw HTML content of all content models: `GET /api/v3/raw-content/...`
- Prevent social media headers endpoint from delivering archived region content

### Side effects
- Hopefully none. Does not touch existing endpoints.

### How to test

curl the endpoints:
- http://localhost:8000/api/v3/raw-content/augsburg/en/welcome/
- http://localhost:8000/api/v3/raw-content/augsburg/de/events/einmalige-veranstaltung/
- http://localhost:8000/api/v3/raw-content/augsburg/de/locations/test-ort/
- http://localhost:8000/api/v3/raw-content/augsburg/de/test-links/

### Resolved issues

Fixes: #4448 


integreat_cms/api/v3/llm_content.py
```diff
@@ -0,0 +1,283 @@
...
+        raise Http404("Invalid news type is given.")
+
+    if not (
+        news_translation := get_public_translation_for_webapp_link_parts(
+            region.slug,
+            language_slug,
+            ["news", news_type, news_raw_id],
+        )
+    ):
+        raise Http404("News not found in this region with this language.")
```
MizukiTemma: I guess `get_public_translation_for_webapp_link_parts` is not compatible with news and it tries to find a `PageTranslation`.

Probably `find_post` can be used instead?

MizukiTemma (changes requested):
Thank you for the PR 💪 

I found one issue with `news_content` (see the comment below) and have a question.

In the current implementation archived or hidden regions and their contents can be exposed. Is that so planned?

svenseeberg:
> In the current implementation archived or hidden regions and their contents can be exposed. Is that so planned?

Good question. I suppose for hidden regions this is the inteded bahvior. We expose them in the normal API as well. But for archived regions this is almost certainly not intended. This probably also affects the social media headers. I'll add a fix for both in this PR.

integreat_cms/api/v3/llm_content.py
svenseeberg: Do you think it would be useful to hook the find_posts into the get_public_translation_for_webapp_link_parts to have one single function that can resolve paths to content?

svenseeberg: No, not a good idea. It should now be implemented as in the social media headers.

MizukiTemma (approved):
Looks good 😸 Thank you!

dkehne:
Opus 5 opinion on that PR:

Good that this also closes the archived-content gap in the existing social endpoints — that fixes a real pre-existing leak.

Three things I'd check before merge:

- `raw_content.html` renders `{{ content|safe }}`. That holds for page/event/POI content, which is sanitized in `CustomContentModelForm.clean_content`. But push-notification text isn't sanitized anywhere as far as I can tell (`PushNotificationTranslationForm.Meta.fields` is just `["title", "text", "automatic_translation"]`, no `clean_content`), and `AbstractNewsManager.clean_html` only strips elements whose class contains `pvc_`, so external news HTML passes through as well. Since the response is HTML served from the CMS origin, could that be an XSS path for those two sources? Sanitizing (or escaping) in the view might be the safer place — the gap isn't introduced by this PR, but the new endpoint is what makes it reachable. Push text also loses its line breaks through `|safe`, so something like `|linebreaks` may be what's wanted anyway.
- `foreign_object.archived`: `ImprintPage` doesn't have that attribute, and `internal_link_utils.py:144` maps `disclaimer` to `ImprintPageTranslation`, which the `<path:path>/` catch-all routes to the page view. So `/disclaimer/<something>/` looks like it would raise `AttributeError` → 500 — and since the check is new in `social_media_headers.py` too, that would also regress the existing endpoint. Would `getattr(..., "archived", False)` do?
- `pushnews_manager.raw_content` filters on `push_notification__id`, while `get_single_news` right below it uses the translation id plus the `archived=False` and `FCM_HISTORY_DAYS` filters. Since the webapp URLs and the news JSON API both use the translation id, raw_content would resolve to a different notification than the one linked — and without those two filters, unsent and scheduled notifications become readable in full. Should it mirror `get_single_news` there?

Smaller ones: the 404 path routes through `render_error_headers` → `webapp_server_side_include.html`, which emits a deliberately unclosed `<html><head>` — fine for a head fragment, but this endpoint's success case is a body fragment, so a miss would inject a nested `<html><head>` mid-document. `expand_shortcodes` isn't applied (`api/v3/pages.py:99` does), so contact shortcodes come through as literal `[contact 5 email]` — and #4448 lists contacts as in scope. `site_url(request)` strips the `/api/v3/social` prefix, which never appears in raw-content paths, so `url`/`language_code` are unused-but-wrong in the template context. And the release note seems to be missing — `API_SOCIAL_ENDPOINTS` in `tests/api/api_config.py` also looks like it would cover these endpoints almost for free.

jonbulz (commented):
Thanks! I think `PushnewsManager.raw_content` possibly serving archived and unpublished `PushNotifications` should be fixed. The XSS-vulnerability that @dkehne described also seems valid, and running external news content through `_xss_cleaner` at import couldn't hurt. After those 2 are fixed, I think this is ready to merge :)
The shortcode expansion we can do in a follow-up. Also, I think the disclaimer path routes to region content, which returns 200 with ` <h1>Augsburg | Integreat</h1>`, not the `AttributeError`. Not ideal that we can't reach the content imprint via the `raw-content` path, but I don't think it's critical

svenseeberg:
@jonbulz the issues should now be addressed.

jonbulz (approved):
Thanks, I guess fixing everything also works 😆 

Two more findings for completeness, both none blocking IMO:
- the imprint route is wrong: it should be [http://localhost:8000/api/v3/raw-content/augsburg/de/disclaimer/](http://localhost:8000/api/v3/raw-content/augsburg/de/disclaimer/) as defined in the `settings.py:67` default. in your test, you test against [http://localhost:8000/api/v3/raw-content/augsburg/de/disclaimer/impressum/](http://localhost:8000/api/v3/raw-content/augsburg/de/disclaimer/impressum/.) the thing is, `/api/v3/raw-content/augsburg/de/disclaimer` is consumed by the reserved pattern route to region_content and returns 200  `<h1>Augsburg | Integreat</h1>`, while `/api/v3/raw-content/augsburg/de/disclaimer/literally-anything/` will give the correct imprint body. so there is a gap and the test is actually misleading. but given that we didn't have any SSI endpoints up until now, I'd say that gap is acceptable
- `RegionMiddleware.get_current_region` raises `Http404` before any view runs, so the exception is not caught, and we return a html document instead of a body fragment, e.g. for [http://localhost:8000/api/v3/raw-content/some-silly-name/de/willkommen/](http://localhost:8000/api/v3/raw-content/some-silly-name/de/willkommen/)

PR #4451, opened 2026-07-23
inherit link ignore flag on new content objects within region

jonbulz:
### Short description

The broken links list view displays `Url` objects. If a `Url` is verified manually from the list view, the `Link.ignore` flag is set to `True` on all `Link` objects within the region that reference this `Url`. When a new content object with the same url is created, a new corresponding `Link` object is created. This will have the default `Link.ignore = False`, thus causing the `Url` to re-surface in the "Verification needed" tab in the broken links view.
This PR introduces a mechanism that, upon creating new `Link` objects, checks if the referenced `Url` has been marked as verified before, and sets the `Link.ignore` flag accordingly.  This behavior is region-scoped.

### Proposed changes

- introduce `inherit_ignore_within_region` method and `region_of` and `_region_links` helper methods
- call `inherit_ignore_within_region` in `do_check_instance_links`
- add tests

### Side effects

- Extra DB work on `do_check_instance_link_preserving_ignore`. But since this runs in a Celery worker, it should not affect system performance for users

### How to test

- create a `PageTranslation` that contains a broken link
- mark the url as verified from the broken link view
- create a new `PageTranslation` that contains the same link
- see that the url remains in the "Link verified" tab in the broken link view

### Resolved issues

Follow up to #3859 

### DIFF
--- a/integreat_cms/cms/apps.py
+++ b/integreat_cms/cms/apps.py
@@ -39,7 +39,12 @@ class CmsConfig(AppConfig):
 
         from .models.abstract_content_translation import AbstractContentTranslation
         from .utils.internal_link_checker import check_internal
-        from .utils.link_ignore_preservation import cache_key_for, target_key
+        from .utils.link_ignore_preservation import (
+            cache_key_for,
+            inherit_ignore_within_region,
+            region_of,
+            target_key,
+        )
 
         Url.check_internal = check_internal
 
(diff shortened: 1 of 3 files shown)


dkehne:
Opus 5 opinion on that PR:

`_region_links` reads a wider scope than the UI writes: both the "ignore" and "unignore" actions operate on `url.regions_links`, i.e. `get_link_query` (`linkcheck_list_view.py:219-238`), which restricts events to `filter_upcoming()`. So a link on a past event can still count as verification evidence, but no UI action can reach it any more — once an editor revokes verification, that link keeps `ignore=True` and the next content object with the same URL inherits it again.

Is the wider scope deliberate? If past events should keep counting, the unignore action probably needs to cover the same set; if not, reusing `get_link_query(Region.objects.filter(pk=region.pk))` would make read and write scope identical by construction — and would also settle the non-latest-version and archived-POI differences. The docstring's own reasoning ("hidden from the broken-links dashboard … must not suppress the same URL on live content") reads like the latter was the intent.

jonbulz:
Yes that's actually a good suggestion. The idea was that `_region_link` had better performance and was more scalable, but I measured with dummy data and it actually doesn't make much of a difference.

seluianova (approved):
Looks good to me 👍
Couldn't find any issues

MizukiTemma (approved):
Soooo cooooool improvement 😻Thank you!

PR #4452, opened 2026-07-23
Bugfix: select all action in feedback outside checkbox

### LINKED ISSUE #4237: Select all not triggering
### Describe the Bug

Select all does not trigger if I press slightly outside the checkbox. However, the checkbox still changes.

### Steps to Reproduce

1. Go to https://cms-test.integreat-app.de/testumgebung/feedback/
2. Press slightly outside of the checkbox

### Expected Behavior

If the select all checkbox gets selected, all other entries should as well.

### Actual Behavior

Only the select all checkbox changes its state. All other entries stay unaffected.

### Screenshot

https://github.com/user-attachments/assets/f32b5a3d-0d09-4c5b-8a64-dadd868d0227

### Additional Information

N/A

### Related Issues

N/A

lunars97:
### Short description

This PR fixes `select all` action checkbox being checked outside of checkbox.

### Proposed changes

- Remove redundant attribute from the table and related to it js file

### Side effects

- should be none

### How to test

Go to Feedback and try to click outside of checkbox

### Resolved issues

Fixes: #4237 


MizukiTemma (approved):
Thank you!

dkehne:
Opus 5 opinion on that PR:

The select-all handler at `bulk-actions.ts:122` listens for `click`, while `toggle-grid-checkbox` dispatched `change` — I think that's the actual mismatch. Row clicks on data rows did work, thanks to the `change` delegation at `bulk-actions.ts:143`; only the header checkbox flipped without propagating.

If that holds, a one-line change would fix #4237 while keeping the row-click convenience:

```diff
- selectAllCheckbox.addEventListener("click", () => {
+ selectAllCheckbox.addEventListener("change", () => {
```

— i.e. keeping `data-enable-row-checkbox-toggle` and `toggle-grid-checkbox.ts` as they are. Real clicks on a checkbox fire both events, so nothing is lost. And since `#bulk-select-all` appears in 10 templates, this would close the same latent gap everywhere; `statistics-page-accesses.ts:41` has the identical `click` pattern.

It would also match the issue's expected behaviour ("all other entries should be selected as well") rather than making the click a no-op.

Dropping row-click is of course a valid product call — I just wanted to flag that the attribute wasn't dead code (it was added in 2d9a01130 for the feedback lists), so it's a small UX change worth a conscious yes. If you do keep this approach, maybe the release note could mention that clicking a row no longer toggles the checkbox, since editors will notice. WDYT?

seluianova (changes requested):
The described issue is fixed, but I noticed one side effect so far:

**Before:**
When a user clicked on a feedback row, e.g. here, the feedback was selected (the checkbox was checked).

Image

**After:**
Now this no longer works, and the user has to click directly on the checkbox to select the feedback.

I don't know how important this behavior was or whether we want to preserve it. _Probably we do?_

@MizukiTemma, do you know?

MizukiTemma:
@seluianova 
You are really a super reviewer 😻 

I don't think we have to keep the old behaviour, becuase it was actually inconsistent with other content lists, and I neither remember it was explicitely requested nor find any trace it was implemented on purpose. So I would say we can go with the current implementation ✅

seluianova (approved):
> So I would say we can go with the current implementation ✅

Thanks for checking! No more concerns from my side. Lgtm 👍

PR #4453, opened 2026-07-24
Fetch tunews if cache is empty

svenseeberg:
### Short description
Fill news cache if cache is missed.

### Proposed changes
- We cannot rely on the news cache being warmed up. We do not persist Redis on a hard disk and also the database migrations trigger a flush of the cache.
- This change fetches the news from the remote source if the cache key is not found. If the cache key has an empty value, we assume that there are currently no news.

### Side effects
- Users missing the cache have to wait a bit longer.

### How to test
1. Flush the cache with `redis-cli flushall`
2. curl the news endpoint

### Resolved issues

Fixes: https://chat.tuerantuer.org/digitalfabrik/pl/6oqmwagroty38xk5nr85zuorny

### DIFF
--- a/integreat_cms/news_managers/abstract_news_manager.py
+++ b/integreat_cms/news_managers/abstract_news_manager.py
@@ -21,6 +21,8 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
+_CACHE_MISS = object()
+
 
 class NewsItem(TypedDict):
     id: str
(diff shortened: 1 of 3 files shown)


MizukiTemma (approved):
Thank you so much!

jonbulz (approved):
I think this is good enough for the testing phase if we need it fast, but we should make the cache-miss language specific as soon as possible. If we get a non-200 status response or RequestException for a language (which is not unlikely, considering that we don't sync our languages with the news providers), we skip that key. Requests to news in that language will then always trigger a re-import of every language.
Additionally, I think the tests now hit the real endpoint. We should switch to a mock response as soon as possible.

jonbulz:
Plus a test for the cache-miss case would be nice :)

MizukiTemma:
@jonbulz 

Thank you for review and improvement suggestions 😸 I'll arleady let this PR go into the test system so App Team can develop things on their side using news posts in the test system without asking us to import them manually.

Issue #4138, opened 2026-02-08
API: Add CRM metadata endpoint for region feature flags


dkehne:
## Motivation

The [CustomCRM](https://github.com/digitalfabrik/customcrm) tracks Integreat product deployments per municipality (Account). After go-live, account managers need to see which features are enabled for a region (chat, map, events, MT contract, etc.). Currently this data is maintained manually in the CRM, leading to stale information.

A lightweight API endpoint would allow the CRM to sync feature flags directly from the CMS, ensuring data stays accurate.

## Proposal

New public endpoint (no auth, consistent with existing API):

```
GET /api/v3/<region_slug>/crm-metadata/
```

### Response

```json
{
  "slug": "augsburg",
  "name": "Stadt Augsburg",
  "live": true,
  "events_enabled": true,
  "locations_enabled": true,
  "push_notifications_enabled": true,
  "integreat_chat_enabled": false,
  "mt_budget_booked": 1000000,
  "last_updated": "2026-01-15T10:30:00+00:00"
}
```

### Why a separate endpoint?

The existing `/api/v3/regions/<slug>/` doesn't expose all needed fields:
- `mt_budget_booked` (raw integer) is not exposed at all
- `integreat_chat_enabled` is only available as composite `is_chat_enabled` (includes Zammad check)
- `summ_ai_enabled` is missing

A dedicated endpoint avoids polluting the app-facing API with administrative metadata.

### Implementation

- New file `integreat_cms/api/v3/region_crm_metadata.py` following the same pattern as `regions.py` (`@json_response` decorator, `request.region` via middleware)
- Register URL in `urls.py` under `api/v3/<slug:region_slug>/` alongside `languages/` and `offers/`
- The endpoint uses `request.region` resolved by `RegionMiddleware`

### Field mapping in the CRM

| API field | CRM usage |
|---|---|
| `integreat_chat_enabled` | Suchassistent (bool) |
| `locations_enabled` | Kartenfunktion (bool) |
| `push_notifications_enabled` | Nachrichtenfunktion (bool) |
| `events_enabled` | Veranstaltungsfunktion (bool) |
| `summ_ai_enabled` | Leichte Sprache (bool) |
| `mt_budget_booked` | MT-Vertrag 1 Mio (`== 1_000_000`) / 5 Mio (`== 5_000_000`) |

CRM-side implementation: digitalfabrik/customcrm (separate issue)

osmers:
[...]

PR #4466, opened 2026-07-27
Add personal API tokens for users

dkehne:
### Short description

First step of the CRM integration (#4138): a personal API token per user, so the upcoming region settings and statistics endpoints can be authenticated without bypassing the RBAC structure.

The CMS API currently has **no authentication at all** — this PR adds the foundation, not yet any protected endpoint.

Per the discussion in #4138, this deliberately avoids a per-region token: the `region_slug` in the URL already provides the CRM↔region mapping, and a token bound to a normal user keeps that user's permissions intact.

### Implementation

- **`UserApiToken`** model (modelled on `FidoKey`): user FK, name, `prefix`, `token_hash`, `created_at`, `last_usage`
  - only a SHA-256 hash is stored; the plaintext is shown **once** after creation and is unrecoverable
  - the random `prefix` allows looking up a token without hashing every row; the comparison itself is constant-time via `secrets.compare_digest`
  - a fast hash is appropriate here because the token carries 32 bytes of entropy — a slow password hash would only add latency to every API request
- **`api_token_required(permission=None)`** decorator in `api/decorators.py`: reads `Authorization: Bearer <token>`, resolves and validates it, rejects tokens of deactivated users, optionally enforces a permission, records `last_usage` and sets `request.user`
- **Account settings UI**: new section to create tokens (name), list them (name / last usage / created at) and delete them. Deletion is a POST form with CSRF protection and is scoped to `request.user.api_tokens`, so nobody can delete somebody else's token.

### Testing

11 tests in `tests/cms/test_user_api_token.py` covering: plaintext is never persisted, token lookup (valid / wrong secret / unknown prefix / malformed / empty), successful authentication + `last_usage` update, rejection of missing/malformed/unknown/non-bearer headers, rejection of deactivated users, and permission enforcement (unprivileged user gets 403, privileged passes).

`ruff`, `mypy`, `djlint` and `check_translations` all pass locally.

### Follow-ups (separate PRs, see #4138)

1. `POST/GET /api/v3/<region_slug>/settings/` — generic region settings endpoint + read-only form for CRM-managed regions + removal of the `mt_midyear_start_month` logic incl. data migration
2. `/api/v3/<region_slug>/statistics/` — async webhook replacing the region-condition CSV

Part of #4138 · CRM side: digitalfabrik/customcrm#19


dkehne:
### Code review

Found 3 issues, all fixed in 6c8d397:

1. The plaintext token was passed through the messages framework, which writes it to the log. `MESSAGE_STORAGE` is `MessageLoggerStorage`, whose `add()` calls `logger.log(level, message)` whenever `MESSAGE_LOGGING_ENABLED` is set — it defaults to `DEBUG`, is forced `True` in `circleci_settings.py`, and is operator-toggleable in production via `INTEGREAT_CMS_MESSAGE_LOGGING_ENABLED`. The usable secret would have landed verbatim in the logs, defeating the point of storing only a hash. Now handed over via the session and rendered once in the template, the same way `TOTPRegisterView` passes its TOTP secret; a regression test asserts the plaintext never reaches the log.

https://github.com/digitalfabrik/integreat-cms/blob/322b84a0d8c1094c27b23645c36fdb286b6d1e11/integreat_cms/cms/views/settings/user_settings_view.py#L149-L159

2. Model naming: `UserApiToken` reintroduced the `User` prefix that was deliberately dropped in #2257 ("it's also called `Organization` instead of `UserOrganization`, `Role` instead of `UserRole`") — its siblings in the same package are `FidoKey`, `Organization` and `Role`. Renamed to `ApiToken`.

https://github.com/digitalfabrik/integreat-cms/blob/322b84a0d8c1094c27b23645c36fdb286b6d1e11/integreat_cms/cms/models/users/user_api_token.py#L38-L41

3. `ApiTokenForm.__init__` declared and forwarded `*args`, but `CustomModelForm.__init__` accepts keyword arguments only, so any positional call would raise `TypeError`. The docstring documented `*args` as supported. Reduced to `**kwargs` like the sibling forms.

https://github.com/digitalfabrik/integreat-cms/blob/322b84a0d8c1094c27b23645c36fdb286b6d1e11/integreat_cms/cms/forms/users/user_api_token_form.py#L28-L38

Also addressed: tests moved to `tests/cms/models/users/` and `tests/api/` to mirror the app structure, an explicit note that `api_token_required` does not check region membership (region-scoped endpoints must do that themselves), and corrections to the prefix-lookup rationale and `token_hash` help text.

integreat_cms/cms/models/users/api_token.py
```diff
@@ -0,0 +1,155 @@
...
+    prefix = models.CharField(
+        max_length=16,
+        unique=True,
+        verbose_name=_("token prefix"),
+        help_text=_("The public part of the token which is used to identify it."),
+    )
+    token_hash = models.CharField(
+        max_length=64,
+        verbose_name=_("token hash"),
+        help_text=_("The hash of the full token."),
```
andrew8er: ❌ See above.

integreat_cms/cms/models/users/api_token.py
```diff
@@ -0,0 +1,155 @@
...
+        on_delete=models.CASCADE,
+        related_name="api_tokens",
+        verbose_name=_("user"),
+    )
+    name = models.CharField(max_length=200, verbose_name=_("token name"))
+    prefix = models.CharField(
+        max_length=16,
+        unique=True,
+        verbose_name=_("token prefix"),
+        help_text=_("The public part of the token which is used to identify it."),
```
andrew8er: ❌ 
- The encoding (hexdigest) of the token should be documented as well, since it is crucial for interpreting the data here. 
- Another option would be to store the prefix (and hash) directly as [`bytea`](https://www.postgresql.org/docs/current/datatype-binary.html#DATATYPE-BINARY-BYTEA-HEX-FORMAT) data type so we could omit the string encoding with `hexdigest()`, using [`digest()`](https://docs.python.org/3/library/hashlib.html#hashlib.hash.digest) instead.

integreat_cms/cms/models/users/api_token.py
```diff
@@ -0,0 +1,155 @@
...
+    from .user import User
+
+#: Number of random bytes used for the secret part of a token
+TOKEN_SECRET_BYTES: int = 32
+
+#: Number of random bytes used for the prefix which identifies a token
+TOKEN_PREFIX_BYTES: int = 6
+
+
+def hash_token(token: str) -> str:
```
andrew8er: 💭 Since this function is only called from this file, I'd suggest naming it `_hash_token`. Also, to avoid misuse, it could take `prefix` and `secret` parameters and would do the appropriate string formatting itself.

integreat_cms/cms/models/users/api_token.py
```diff
@@ -0,0 +1,155 @@
...
+
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        related_name="api_tokens",
+        verbose_name=_("user"),
+    )
+    name = models.CharField(max_length=200, verbose_name=_("token name"))
+    prefix = models.CharField(
+        max_length=16,
```
andrew8er: ❓ When using `varchar` as column type, wouldn't the length be `TOKEN_PREFIX_BYTES * 2`, since each byte gets encoded into a double char hex representation? The `token_hash` column seems correct in this regard.

integreat_cms/cms/models/users/api_token.py
```diff
@@ -0,0 +1,155 @@
...
+        unique=True,
+        verbose_name=_("token prefix"),
+        help_text=_("The public part of the token which is used to identify it."),
+    )
+    token_hash = models.CharField(
+        max_length=64,
```
andrew8er: 💭 `max_length=TOKEN_SECRET_BYTES * 2`

andrew8er (commented):
General implementation is good, but the column documentation is a bit lacking. There is also an alternative that might be preferable (regarding storage and compute).

integreat_cms/cms/models/users/api_token.py
dkehne: Adopted — `prefix` and `token_hash` are `bytea` now and hashing uses `digest()` instead of `hexdigest()`. Only the prefix inside the plaintext token stays hex encoded (it has to travel in an `Authorization` header), and both help texts now spell the encoding out. 0ebfe91

integreat_cms/cms/models/users/api_token.py
dkehne: Same commit: the column stores the raw digest now, and the help text says which value is hashed and in which form.

integreat_cms/cms/models/users/api_token.py
dkehne: Done: `_hash_token(prefix, secret)` builds the plaintext itself, so storage and lookup can no longer disagree on the format.

integreat_cms/cms/models/users/api_token.py
dkehne: Correct, it would have had to be `TOKEN_PREFIX_BYTES * 2`. With `bytea` the question disappears — `max_length` is the number of bytes now.

integreat_cms/cms/models/users/api_token.py
dkehne: The 64 was right only by coincidence: a SHA-256 hex digest is 64 characters for any input length, and `TOKEN_SECRET_BYTES` happens to be 32. It is `TOKEN_HASH_BYTES = hashlib.sha256().digest_size` now, so it follows the hash algorithm instead of the token size.

dkehne:
Thanks for the careful look at the columns, @andrew8er — all five points are in 0ebfe91:

- `prefix` and `token_hash` are `bytea` now with `digest()` instead of `hexdigest()`, which halves the storage and drops the encoding step from every lookup. The only remaining hex is the prefix inside the plaintext token, since it has to travel in an `Authorization` header — both help texts state the encoding explicitly.
- `_hash_token(prefix, secret)` assembles the plaintext itself, so storage and lookup cannot disagree on the format. As a side effect the lookup now rejects prefixes of the wrong length or with non-hex characters before touching the database.
- The hash length comes from `hashlib.sha256().digest_size` rather than from the token size, which is what the old `64` accidentally matched.

Also rebased onto the current `develop` (the SUMM.AI removal moved the migration to `0159`). Could you have another look and approve if this is what you had in mind?

andrew8er (approved):
Nice work!

seluianova (approved):
looks good to me 👍

PR #4467, opened 2026-07-27
Add region settings API endpoint and replace mid-year budget start

dkehne:
### Short description

Second step of the CRM integration (#4138): the region settings endpoint, plus the replacement of the mid-year budget mechanism.

> [!IMPORTANT]
> Stacked on #4466 (personal API tokens) — please merge that one first. The diff here is against that branch.

### Endpoint

`GET/POST /api/v3/<region_slug>/settings/`, authenticated with a personal API token and requiring `cms.change_region`.

Deliberately generic rather than CRM-scoped, but it only exposes an explicit allowlist in [`cms/constants/region_api_settings.py`](integreat_cms/cms/constants/region_api_settings.py): `events_enabled`, `locations_enabled`, `contacts_enabled`, `external_news_enabled`, `integreat_chat_enabled`, `push_notifications_enabled`, `mt_budget_booked`, `mt_renewal_month`, `mt_budget_adjustment`. Core fields like `name` and `slug` identify the region and are rejected with 400.

That allowlist is the single source of truth for both the endpoint and the region form: once a region has been written through the API it counts as externally managed, and exactly those fields become read-only with a banner pointing at the connected system.

`GET` mirrors the writable fields plus the derived `mt_budget` / `mt_budget_used` / `mt_budget_remaining`, so a consumer can fetch, change one value and post the same structure back.

### Mid-year start replaced by an explicit adjustment

`mt_midyear_start_month` and its pro-rating are gone. `mt_budget` is now `mt_budget_booked + mt_budget_adjustment`, floored at zero — the adjustment is a signed cumulative correction for the current budget year, so sending the cumulative value keeps the endpoint idempotent.

- **Data migration**: converts the pro-rated amount of existing mid-year regions into an equivalent booked budget, so their effective budget is unchanged.
- **`reset_mt_budget`**: now resets the adjustment alongside the used budget, since it only applies to one budget year and the CRM re-pushes it for the new one.
- **`mt_budget_booked`** no longer restricts values to the package sizes on the model (the API may push arbitrary word counts). The form defines the choices explicitly, so CMS-maintained regions keep the familiar dropdown.

Removed along the way: the mid-year field in the region form, template and `conditional-fields.ts`, and the field from the test fixture.

### Testing

11 endpoint tests in `tests/api/test_region_settings.py`: GET shape, POST writes + sync timestamp, arbitrary budget values accepted, rejection of core/unknown/read-only fields, invalid values (bad renewal month, non-integer budget), missing/invalid token, insufficient permission, unknown region. `test_region_mt_management.py` rewritten for the adjustment arithmetic (including the floor at zero), `test_reset_deepl_budget.py` updated.

`ruff`, `mypy`, `djlint`, `eslint` and `check_translations` pass locally.

### Notes for review

- **Region membership is not checked.** `api_token_required` only verifies the global permission. For this endpoint that is intentional (it is aimed at a service user with `cms.change_region`), but it is worth a deliberate decision — if we want per-region restriction, the check belongs here against `request.region`.
- `tests/api/test_api_result.py::test_api_result[...regions.json...]` fails locally, but it fails identically on `develop` without these changes, so it is unrelated.
- There is a latent test-ordering fragility in the repo: `load_test_data` is session-scoped while `load_test_data_transactional` truncates the tables, so depending on xdist distribution a later test can find an empty database. Not introduced here, but it surfaced while running these suites together.

Part of #4138 · CRM side: digitalfabrik/customcrm#19


dkehne:
### Code review

Found 1 issue, fixed in a4e6f6c:

1. The form always rebuilt `mt_budget_booked` as a choice field limited to the six predefined package sizes — including for API-managed regions, whose budget may be any value. The disabled select then has no matching `<option>`, so the browser renders the first one and shows staff "50.000" instead of the real budget. This is exactly the case the change is meant to support (see `test_post_accepts_budget_outside_the_predefined_sizes`, which writes 77777), and it also contradicts the note in #4138 that the field "should be rendered as a plain value" for externally managed regions. API-managed regions now keep the plain integer field; CMS-maintained ones still get the dropdown. Regression test added.

https://github.com/digitalfabrik/integreat-cms/blob/cbb44a716709f59a8632b1c54e30b6cdeb1ca104/integreat_cms/cms/forms/regions/region_form.py#L254-L268

Checked and found **not** to be issues:

- **Cross-region writes via the URL slug.** The endpoint only checks the global `cms.change_region` permission and not region membership, which looks like it would let a token write any region. In practice `change_region` is granted exclusively to `SERVICE_TEAM` and `CMS_TEAM` ([`roles.py:152-163`](https://github.com/digitalfabrik/integreat-cms/blob/cbb44a716709f59a8632b1c54e30b6cdeb1ca104/integreat_cms/cms/constants/roles.py#L151-L164)) — both staff roles with access to every region anyway, so this grants nothing extra. Still worth a conscious decision if the permission is ever given to a region-scoped role; the constraint is documented in the `api_token_required` docstring.
- The data migration arithmetic reproduces the old `mt_budget` property exactly, and the operation order keeps `mt_midyear_start_month` readable while `RunPython` executes.
- The allowlist cannot be bypassed: non-dict bodies, unknown fields and core fields are all rejected before any `setattr`, and `full_clean(exclude=...)` validates precisely the submitted fields.

integreat_cms/api/v3/region_settings.py
```diff
@@ -0,0 +1,119 @@
...
+    # Derived values which are useful for the consumer but not writable
+    settings["mt_budget"] = region.mt_budget
+    settings["mt_budget_used"] = region.mt_budget_used
+    settings["mt_budget_remaining"] = region.mt_budget_remaining
+    settings["api_settings_synced_at"] = (
+        region.api_settings_synced_at.isoformat()
+        if region.api_settings_synced_at
+        else None
+    )
+    return settings
```
andrew8er: 💭 Could be expressed more concise in a single expression:
```suggestion
    return {
        **(
            {
                field: getattr(region, field)
                for field in region_api_settings.WRITABLE_FIELDS
            }
        ),
        "mt_budget": region.mt_budget,
        "mt_budget_used": region.mt_budget_used,
        "mt_budget_remaining": region.mt_budget_remaining,
        "api_settings_synced_at": (
            region.api_settings_synced_at.isoformat()
            if region.api_settings_synced_at
            else None
        ),
    }
```

integreat_cms/api/v3/region_settings.py
```diff
@@ -0,0 +1,119 @@
...
+    from typing import Any
+
+    from django.http import HttpRequest
+
+    from ...cms.models import Region
+
+logger = logging.getLogger(__name__)
+
+
+def transform_region_settings(region: Region) -> dict[str, Any]:
```
andrew8er: 💭 A typed dictionary `Settings` would be nice.

integreat_cms/api/v3/region_settings.py
```diff
@@ -0,0 +1,119 @@
...
+    region.save(
+        update_fields=[*data.keys(), "api_settings_synced_at"],
+    )
+    logger.info(
+        "Settings of %r updated via API by %r (fields: %s)",
+        region,
+        request.user,
+        ", ".join(sorted(data)),
+    )
+    return JsonResponse(transform_region_settings(region))
```
andrew8er: 💭 While having a little bit more indentation, I think the following version is more clear. To have some symmetry for the different branches, the POST part could either be moved into its own function `set_region_settings()` (suggestion) similar to `transform_region_settings()` for the GET part, or both part could be inlined into their respective case blocks.

```suggestion
    match request.method:
        case "GET":
            return JsonResponse({
                **(
                    {
                        field: getattr(region, field)
                        for field in region_api_settings.WRITABLE_FIELDS
                    }
                ),
                "mt_budget": region.mt_budget,
                "mt_budget_used": region.mt_budget_used,
                "mt_budget_remaining": region.mt_budget_remaining,
                "api_settings_synced_at": (
                    region.api_settings_synced_at.isoformat()
                    if region.api_settings_synced_at
                    else None
                ),
            })
        case "POST":
            try:
                data = json.loads(request.body.decode())
            except json.JSONDecodeError:
                return JsonResponse({"error": "Invalid JSON body."}, status=400)

            if not isinstance(data, dict):
                return JsonResponse({"error": "Expected a JSON object."}, status=400)

            if unknown_fields := sorted(
                set(data) - set(region_api_settings.WRITABLE_FIELDS)
            ):
                return JsonResponse(
                    {
                        "error": f"Unknown or read-only fields: {', '.join(unknown_fields)}",
                    },
                    status=400,
                )

            for field, value in data.items():
                setattr(region, field, value)

            # Enforce the model level validation, e.g. that the budget values are integers and that the
            # renewal month is one of the valid choices. Only the submitted fields are validated, so a
            # request is not rejected because of pre-existing invalid data in an unrelated field.
            untouched_fields = [
                field.name for field in region._meta.fields if field.name not in data
            ]
            try:
                region.full_clean(exclude=untouched_fields, validate_unique=False)
            except ValidationError as e:
                return JsonResponse({"error": e.message_dict}, status=400)

            region.api_settings_synced_at = timezone.now()
            region.save(
                update_fields=[*data.keys(), "api_settings_synced_at"],
            )
            logger.info(
                "Settings of %r updated via API by %r (fields: %s)",
                region,
                request.user,
                ", ".join(sorted(data)),
            )
            return JsonResponse(transform_region_settings(region))
        case _:
            return JsonResponse({"error": "Invalid request."}, status=405)
```

integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -254,9 +251,25 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+                choices=machine_translation_budget.CHOICES,
+                coerce=int,
+                label=self.fields["mt_budget_booked"].label,
+                initial=machine_translation_budget.MINIMAL,
+            )
+        # Settings which are pushed through the API must not be edited here — the next sync would
+        # overwrite them anyway
+        if is_api_managed:
+            for field_name in region_api_settings.WRITABLE_FIELDS:
+                self.fields[field_name].disabled = True
```
andrew8er: 💭 The two if statements represent mutually exclusive alternatives based on the same condition, so an `if else` statement is apt:

```suggestion
        if bool(self.instance.id and self.instance.is_api_managed):
            # Settings which are pushed through the API must not be edited here — the next sync would
            # overwrite them anyway
            for field_name in region_api_settings.WRITABLE_FIELDS:
                self.fields[field_name].disabled = True
        else:
            # The model no longer restricts the budget to the predefined package sizes, because the
            # API may push arbitrary values. Regions maintained in the CMS should still get the
            # dropdown with the known sizes, so the choices are defined on the form instead.
            # API-managed regions keep the plain integer field: their budget may be any value, and a
            # dropdown without a matching option would silently display the wrong one.
            self.fields["mt_budget_booked"] = forms.TypedChoiceField(
                choices=machine_translation_budget.CHOICES,
                coerce=int,
                label=self.fields["mt_budget_booked"].label,
                initial=machine_translation_budget.MINIMAL,
            )
```

integreat_cms/cms/migrations/0158_mt_budget_adjustment.py
```diff
@@ -0,0 +1,97 @@
...
+                null=True,
+                verbose_name="last settings sync via API",
+            ),
+        ),
+        migrations.AlterField(
+            model_name="region",
+            name="mt_budget_booked",
+            field=models.PositiveIntegerField(
+                default=50000,
+                help_text="The booked budget in words. For regions managed via the API this can be any value, so no choices are enforced on the model.",
```
andrew8er: ❓ Does this actually mean "The booked budget in _number_ of words"?

integreat_cms/locale/de/LC_MESSAGES/django.po
```diff
@@ -12126,6 +12146,30 @@ msgstr ""
...
+#~ "Bitte geben Sie ein gültiges Startdatum für den Abrechnungszeitraum der "
+#~ "Fremdsprachenübersetzung an."
+
+#~ msgid "Budget year start date for foreign language translation"
+#~ msgstr ""
+#~ "Unterjähriger Start des Abrechnungszeitraums für Fremdsprachenübersetzung"
+
+#~ msgid "Month from which the add-on package was booked"
+#~ msgstr "Monat ab welchem das Add-On Paket gebucht wurde"
+
```
andrew8er, line 12170: ❓ Do we need this later?

integreat_cms/cms/fixtures/test_data.json
andrew8er, line 1: 💭 We could also use some ready made test API tokens here, so testing the API with an API client becomes easier.

andrew8er (approved):
LGTM. A few suggestions.

MizukiTemma:
@dkehne 

What is the state of this PR? Let us know if support helps 👁️‍🗨️

integreat_cms/api/v3/region_settings.py
dkehne: Added as `Settings`. The writable half is still read from `WRITABLE_FIELDS` so the allowlist stays the single source of truth, which the type checker cannot verify — `test_get_region_settings` now asserts that the response keys, the allowlist and the keys of `Settings` agree.

integreat_cms/api/v3/region_settings.py
dkehne: Done, single expression now.

integreat_cms/api/v3/region_settings.py
dkehne: Went with `match` plus the extracted `set_region_settings()` — that keeps the handler down to the three cases without the extra nesting.

integreat_cms/cms/forms/regions/region_form.py
dkehne: Done.

integreat_cms/cms/migrations/0158_mt_budget_adjustment.py
dkehne: Yes — corrected in the model and the migration.

integreat_cms/locale/de/LC_MESSAGES/django.po
dkehne: Yes, those should stay: the pro-rated budgets are still how the CMS works, the pro-rating just moves to the connected system now and arrives as `mt_budget_adjustment`. If we ever render it in the CMS again the wording is right there.

integreat_cms/cms/fixtures/test_data.json
dkehne: Good idea — the test data now contains two tokens, `746573740001.test-token-root` for `root` and `746573740002.test-token-service-team` for `service_team` (the latter has `cms.change_region`), and the README shows a curl call against the settings endpoint.

dkehne:
Thanks for the review and the approval, @andrew8er — all suggestions are in 77f2dd6: the typed `Settings` dictionary, the single-expression response, `match` on the request method with the write path extracted into `set_region_settings()`, the `if`/`else` in the region form, the corrected "number of words" help text and two ready-made API tokens in the test data (documented in the README).

@MizukiTemma with the approval here in place, a second review would be very welcome — this one is stacked on #4466 (personal API tokens), which is still waiting for a first review, so that would be the more useful place to start. Both branches are rebased onto the current `develop`; the SUMM.AI removal touched the same budget code, so the migrations are `0159`/`0160` now.

integreat_cms/release_notes/current/unreleased/4138-settings-api.yml
seluianova, line 1: not sure if this has any impact, but other release notes don't include the description in the file name, so it's just `4138.yml`

integreat_cms/cms/migrations/0160_mt_budget_adjustment.py
```diff
@@ -0,0 +1,97 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="region",
+            name="mt_budget_adjustment",
+            field=models.IntegerField(
+                blank=True,
+                default=0,
+                help_text="Cumulative manual adjustment for the current budget year, added on top of the booked budget. Negative values reduce the budget.",
```
seluianova: nitpick: might just be my personal misunderstanding, but `for the current budget year` sounds like it applies only to the current year. 
in fact it is "applied until changed or removed."

integreat_cms/cms/migrations/0160_mt_budget_adjustment.py
```diff
@@ -0,0 +1,97 @@
...
+        region.mt_budget_booked = int(multiplier * region.mt_budget_booked)
+        region.save(update_fields=["mt_budget_booked"])
+
+
+def revert_midyear_conversion(
+    _apps: Apps,
+    _schema_editor: BaseDatabaseSchemaEditor,
+) -> None:
+    """
+    Do nothing when migrating backwards
```
seluianova: can we migrate backwards? 🙈

seluianova:
late-coming question: currently, any user can create a token. 
is this intended? maybe we want to limit this functionality to particular user roles only?

integreat_cms/api/v3/region_settings.py
```diff
@@ -0,0 +1,158 @@
...
+        "Settings",
+        {
+            **{
+                field: getattr(region, field)
+                for field in region_api_settings.WRITABLE_FIELDS
+            },
+            "mt_budget": region.mt_budget,
+            "mt_budget_used": region.mt_budget_used,
+            "mt_budget_remaining": region.mt_budget_remaining,
+            "api_settings_synced_at": (
```
seluianova: question regarding this property:
for me, it sounds more like `api_settings_updated_at`, because we only change this timestamp with a POST request.
when we receive a GET request and the data is updated on the other side, it is also a synchronization, but we don't update the property in this case?

integreat_cms/cms/templates/regions/region_form.html
```diff
@@ -25,6 +25,20 @@ <h1 class="heading overflow-hidden text-ellipsis">
                 </button>
             {% endif %}
         </div>
+        {% if form.instance.is_api_managed %}
+            <div class="bg-yellow-100 border-l-4 border-yellow-500 text-orange-700 px-4 py-3 mb-4"
+                 role="alert">
+                <p>
+                    {% with last_sync=form.instance.api_settings_synced_at %}
+                        {% blocktranslate trimmed %}
+                            The booked features and the translation budget of this region are managed
```
seluianova, line 34: nipick: Activate events / Activate locations / Activate contacts features are disabled as well, but there is no hint why. might be a bit confusing perhaps

seluianova (approved):
tested, looks good to me!
just added a few optional thoughts

@dkehne can you resolve the conflict?

integreat_cms/release_notes/current/unreleased/4138-settings-api.yml
bahaaTuffaha: @MizukiTemma there is 2 release notes with the same number is that ok?
is that a mistake from claude or another ticket?

seluianova: Ah, that's why the text was added

bahaaTuffaha: then I have to remove one right? which one do think it's better to keep?

integreat_cms/cms/migrations/0160_mt_budget_adjustment.py
bahaaTuffaha: I used `migrations.RunPython.noop` instead

integreat_cms/release_notes/current/unreleased/4138-settings-api.yml
seluianova: yes I think we could keep one release note for both PRs, otherwise it looks like this
image
maybe it could be the general one?
`Allow users to create personal API tokens and add an API endpoint to read and write the region settings`

bahaaTuffaha:
There is unrelated Test failing at develop... 🤔 do I merge anyway?

jonbulz:
We just merged #4441, which hopefully fixes some flaky test. Please rebase and see if that fixes it :)

PR #4470, opened 2026-07-30
Update Documentation and Devcontainer to python 3.13

jarlhengstmengel:
### Short description

After #4424 Python 3.13 is a mandatory. This PR updates the Devcontainer to 3.13 and updates our documentation that python 3.13 is required.

### Proposed changes

- Bumps up Devcontainer to Python 3.13
- Updates docs
- Some Devcontainer tweeks: No standard loading of testdata after building the container (can cause issues with celery), standard activation of linkcheck listeners (default in the cms config)

### Side effects

- None

### How to test

If you are using the devcontainer, rebuild it (VS code should prompt you to do so).
If you don't use devcontainer, maybe do a sanity check and check the changed docs.

### Resolved issues

Fixes: -

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/vscode/devcontainers/python:3.11-bookworm
+FROM mcr.microsoft.com/vscode/devcontainers/python:3.13-bookworm
 
 RUN rm -f /etc/apt/sources.list.d/yarn.list
 RUN apt-get update && \
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/vscode/devcontainers/python:3.13-bookworm
+FROM mcr.microsoft.com/devcontainers/python:3.13-bookworm
 
 RUN rm -f /etc/apt/sources.list.d/yarn.list
 RUN apt-get update && \
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -30,7 +30,7 @@
       }
     }
   },
-  "postCreateCommand": "./tools/install.sh --python python3.11 && ./tools/migrate.sh && ./tools/loadtestdata.sh",
+  "postCreateCommand": "./tools/install.sh --python python3.13 && ./tools/migrate.sh && ./tools/loadtestdata.sh",
   "postStartCommand": "git config --global --add safe.directory ${containerWorkspaceFolder} && redis-server --daemonize yes --unixsocket /workspace/redis-server.sock --unixsocketperm 770",
   "remoteEnv": {
     "DJANGO_SETTINGS_MODULE": "integreat_cms.core.settings",
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -30,7 +30,7 @@
       }
     }
   },
-  "postCreateCommand": "./tools/install.sh --python python3.13 && ./tools/migrate.sh && ./tools/loadtestdata.sh",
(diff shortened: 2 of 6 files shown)


.devcontainer/devcontainer.json
```diff
@@ -40,6 +40,6 @@
     "INTEGREAT_CMS_SECRET_KEY": "dummy",
     "INTEGREAT_CMS_BACKGROUND_TASKS_ENABLED": "0",
     "INTEGREAT_CMS_SUMM_AI_API_KEY": "dummy",
-    "INTEGREAT_CMS_LINKCHECK_DISABLE_LISTENERS": "1"
+    "INTEGREAT_CMS_LINKCHECK_DISABLE_LISTENERS": "0"
```
MizukiTemma, line 43: why changed to disable it?

.devcontainer/devcontainer.json
```diff
@@ -30,7 +30,7 @@
       }
     }
   },
-  "postCreateCommand": "./tools/install.sh --python python3.11 && ./tools/migrate.sh && ./tools/loadtestdata.sh",
+  "postCreateCommand": "./tools/install.sh --python python3.13 && ./tools/migrate.sh",
```
timobrembeck, line 33: When not loading the test data anymore, we should also adapt this part in the README:
> 5. VSCode will open the project in a new container, install all further required tools ~and load the testdata~.

docs/src/installation.rst
timobrembeck, line 1: In line 18 of this file, there is still an old reference:
```
* `python3 <https://www.python.org/>`_ version 3.9 to 3.11 (for above 3.11 see below)
```

README.md
```diff
@@ -40,7 +40,7 @@ To configure your development environment on your system, please follow these st
 1. Ensure that the following packages are installed alongside your preferred IDE:
    - `npm` version 7 or later
    - `nodejs` version 22 or later
-   - `python3` version 3.11 or later
+   - `python3` version 3.13 or later
```
timobrembeck:
```suggestion
   - `python3` version 3.13
```
the "or later" is not really true, we do not officially support 3.14...

.devcontainer/devcontainer.json
timobrembeck: @MizukiTemma setting the "disable" config to `0` actually enable the listeners, but I also don't know why. Maybe we had a limitation on earlier python versions that is now resolved?

.devcontainer/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/vscode/devcontainers/python:3.11-bookworm
+FROM mcr.microsoft.com/vscode/devcontainers/python:3.13-bookworm
```
timobrembeck: `mcr.microsoft.com/vscode/devcontainers` is now deprecated, see: https://github.com/microsoft/vscode-dev-containers/issues/532
Maybe we could use this opportunity to migrate to the new location?

```suggestion
FROM mcr.microsoft.com/devcontainers/python:3.13-bookworm
```

timobrembeck (changes requested):
Thanks a lot! :+1:

I found a last remainder in publiccode.yml#L59:

https://github.com/digitalfabrik/integreat-cms/blob/706148000cc9afeb72e542f3448a10428d5d1ae8/publiccode.yml#L59

jarlhengstmengel:
@timobrembeck applied your suggestions. Also found a outdated requirement in `install.sh` for required node version (18 instead of 22)

timobrembeck (approved):
Thanks for your changes :+1:

seluianova (approved):
I don't use Testcontainers (I tried it, but it's somehow very slow, and I'm not sure what the problem is), the docs lgtm.

PR #4483, opened 2026-08-05
Add timeout for textlab API call

### LINKED ISSUE #4482: Publishing a page can hang for ~30 min and time out if the Textlab (HIX) API is slow/unresponsive
### Bug Summary

Publishing a page can hang for a very long time (~30 minutes) before the browser shows a timeout notification. The edit itself is saved to the database, but the page is left un-published from the user's perspective, since the client never received a response. Which is possible the case here: https://chat.tuerantuer.org/digitalfabrik/pl/79fqx6878idp3gkm5nrumcguhe

### Expected Behavior

When trying to publish a page a timeout for the hix api should fire after 30 seconds

### Actual Behavior

- after publishing a page a unresponsive API can result in a 30 minute "hang" of the software

### Steps to Reproduce

Not easy to reproduce directly on the Running instance, as the error occurs when an external API is failing. But to test it locally you can do:

Run with `tools/run.sh`
Second Terminal: `nc -l 9999`
Third Terminal: 
```
tools/integreat-cms-cli shell
  >>> from django.conf import settings
  >>> settings.TEXTLAB_API_URL = "http://127.0.0.1:9999"
  >>> settings.TEXTLAB_API_ENABLED = True
  >>> from integreat_cms.cms.views.utils.hix import lookup_hix_score
  >>> lookup_hix_score("some test text")   # this should just hang forever
```

Observe that the call to lookup_hix_score only returns after you kill the nc with STR+C in the second terminal

### Reproducibility

Always (100%)

### Environment

- Production

_No response_

### Severity

None

### Regression Information

_No response_

### Technical Context

_No response_

_No response_

### Verification / Testing

_No response_

### Related Issues

_No response_

hannaseithe:
### Short description

This adds a timeout for the textlab API call in `textlab_api_client.py`

### Proposed changes

- 30 second timeout for the API call
- logging when running into timeout
- test for timeout

### Side effects

- none

### How to test

Follow the instructions in the bug ticket

### Resolved issues

Fixes: #4482

### DIFF
--- a/tests/cms/views/utils/test_hix.py
+++ b/tests/cms/views/utils/test_hix.py
@@ -1,10 +1,14 @@
 from __future__ import annotations
 
+import json
 import math
+import socket
+import time
 from copy import deepcopy
 from typing import TYPE_CHECKING
 
 import pytest
+from werkzeug.wrappers import Request, Response
 
 from integreat_cms.cms.constants.administrative_division import MUNICIPALITY
 from integreat_cms.cms.constants.region_status import ACTIVE
@@ -27,6 +31,8 @@ from tests.utils import disable_hix_post_save_signal
 
 if TYPE_CHECKING:
     from pytest_django.fixtures import SettingsWrapper
+
+    from tests.mock import MockServer
 
 
 def create_dummy_region() -> tuple[Language, Region]:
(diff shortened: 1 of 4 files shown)


MizukiTemma (approved):
`black_hole` 😹 

looks good, thank you!

hannaseithe:
> `black_hole` 😹

Claude can be funny :D

tests/cms/views/utils/test_hix.py
```diff
@@ -349,3 +352,41 @@ def test_contact_card_gets_filtered_out() -> None:
...
+            conn, _ = listener.accept()
+            with conn:
+                # Deliberately never send a response, simulating a stalled API
+                time.sleep(2)
+
+    threading.Thread(target=black_hole, daemon=True).start()
+
+    settings.TEXTLAB_API_ENABLED = True
+    settings.TEXTLAB_API_URL = f"http://127.0.0.1:{port}"
+    settings.TEXTLAB_API_TIMEOUT = 0.2
```
PeterNerlich, line 395: While this is perfectly fine on its own, I'm not sure about how I feel about introducing an entirely new method of mocking requests. We already use `MockServer` in other tests, which is our own convenience wrapper subclassing `pytest_httpserver.httpserver.HTTPServer`, which we can partially re-use:

```suggestion
def test_hix_lookup_times_out_instead_of_hanging(
    settings: SettingsWrapper,
    mock_server: MockServer,
) -> None:
    """
    If the Textlab API accepts the connection but never responds, ``lookup_hix_score()``
    must give up after :attr:`~integreat_cms.core.settings.TEXTLAB_API_TIMEOUT` instead
    of blocking forever (see #<issue-number>).
    """

    def black_hole(request: Request) -> Response:
        mock_server.requests_counter += 1
        time.sleep(2)
        # If this is being handled, the test already failed: We expected to time out
        return Response(
            json.dumps(
                {
                    "formulaHix": 15.12345678,
                    "moreSentencesInClauses": [4, 5, 6],
                    "moreSentencesInWords": [],
                }
            ),
            status=200,
        )

    mock_server.configure("/user/login", 200, {"token": "dummy"})
    mock_server.http_server.expect_request("/benchmark/420").respond_with_handler(
        black_hole
    )

    settings.TEXTLAB_API_ENABLED = True
    settings.TEXTLAB_API_URL = f"http://localhost:{mock_server.port}"
    settings.TEXTLAB_API_TIMEOUT = 0.2
```

*(With this, you need to add `import json`, `from werkzeug.wrappers import Request, Response` and into `if TYPE_CHECKING:` the line `from tests.mock import MockServer`, and can discard `import socket` and `import threading`)*

PeterNerlich (changes requested):
Thank you for the fix! I have a conceptually small nitpick regarding the way we mock the hanging request.

tests/cms/views/utils/test_hix.py
seluianova: optional nit-pick:
`mock_server.requests_counter += 1` can be omitted because we don't check the requests count in this test.
and this `settings.TEXTLAB_API_ENABLED = True` also can be omitted because we set INTEGREAT_CMS_TEXTLAB_API_KEY in test.sh, so it's already true

PeterNerlich: This is true. I included increasing the counter because the property exists on the class, and we simulated an incoming request. In a hypothetical future where someone looks at this test, they might be confused about the intention behind the class. No idea what the probability on this hypothetical is, but since my suggestion introduces a previously unseen way to use the class, I thought consistency was the safest bet.
Cleaner from the class perspective would be to introduce a new method to blackhole requests, but the test is most readable in the current form in my opinion.

hannaseithe: Your change suggestions have been applied. Could you please check again? @PeterNerlich

PeterNerlich (approved):
Can you squash the apply suggestions commit into the other one? *(as a fixup commit, i.e. one that doesn't change the commit message, or its header at least)*

PR #4487, opened 2026-08-05
2308: Return server errors in json for API

### LINKED ISSUE #2308: Server errors in API return HTML response instead of JSON
### Describe the Bug

Whenever an error occurs in the API namespace (no matter whether that's a 404 or 500), the response is our rendered HTML error page.

### Steps to Reproduce

1. Access e.g. https://cms.integreat-app.de/api/regions/non-existing/
2. See rendered HTML error page

### Expected Behavior

In my opinion, the API should always return JSON, even if an error occurs

### Actual Behavior

If the error is not explicitly returned as JSON in the API views, the fallback error page is rendered as HTML

### Additional Information

Probably, we could force a JSON response in a custom middleware?

bahaaTuffaha:
### Short description
The API should always return JSON, even if an error occurs.

### Proposed changes

- Added an `is_api_request` helper that checks `request.resolver_match.app_name` against `"api"`, falling back to a `request.path` prefix check when the URL couldn't be resolved.
- Added a `render_error_response` function that returns a `JsonResponse` for API requests and otherwise falls back to `HttpResponse` wrapping the existing `render_error_template`.
- Updated `handler400`, `handler403`, `handler404`, `handler500`, and `csrf_failure` to use `render_error_response` instead of constructing the response classes directly.
- Changed the return type for the handlers from `HttpResponseBadRequest`, `HttpResponseForbidden`, `HttpResponseNotFound`, and `HttpResponseServerError` to the generic `HttpResponse`.

### Side effects

None.

### How to test

- Launch the server in non debug mode I used this command (Debug=0):
```
source .venv/bin/activate

DJANGO_SETTINGS_MODULE=integreat_cms.core.docker_settings \
INTEGREAT_CMS_DEBUG=0 \
INTEGREAT_CMS_SECRET_KEY=dummy \
DYLD_FALLBACK_LIBRARY_PATH=/opt/homebrew/lib \
integreat-cms-cli runserver --insecure localhost:8000
```
Note: `DYLD_FALLBACK_LIBRARY_PATH` is a macOS/Homebrew workaround for `cairocffi` skip it on Linux.

- Test non-existent url paths like:
    - For the Api : http://localhost:8000/api/v3/non-existing
    - For any other pages: http://localhost:8000/regions/foo

### Resolved issues

Fixes: #2308

### DIFF
--- a/integreat_cms/cms/views/error_handler/error_handler.py
+++ b/integreat_cms/cms/views/error_handler/error_handler.py
@@ -49,7 +49,7 @@ def render_error_template(context: dict[str, Any]) -> SafeString:
     return render_to_string("error_handler/http_error.html", context)
 
 
-def render_error_response(context: dict[str, Any]) -> HttpResponse:
+def _render_error_response(context: dict[str, Any]) -> HttpResponse:
     """
     Render the HTTP error response
 
@@ -78,7 +78,7 @@ def handler400(request: HttpRequest, exception: BadRequest) -> HttpResponse:
         "message": _("There was an error in your request."),
     }
     logger.debug(exception)
-    return render_error_response(context)
+    return _render_error_response(context)
 
 
 def handler403(
@@ -99,7 +99,7 @@ def handler403(
         "message": _("You don't have the permission to access this page."),
     }
     logger.debug(exception)
-    return render_error_response(context)
+    return _render_error_response(context)
 
 
 def handler404(request: HttpRequest, exception: Http404) -> HttpResponse:
@@ -117,7 +117,7 @@ def handler404(request: HttpRequest, exception: Http404) -> HttpResponse:
         "message": _("The page you requested could not be found."),
     }
     logger.debug(exception)
-    return render_error_response(context)
+    return _render_error_response(context)
 
 
 def handler500(request: HttpRequest) -> HttpResponse:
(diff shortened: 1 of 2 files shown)


MizukiTemma (approved):
Looks good, thank you 😸

integreat_cms/cms/views/error_handler/error_handler.py
```diff
@@ -23,6 +18,19 @@
...
+    """
+    Check whether the given request is directed at the API
+
+    :param request: Object representing the user call
+    :return: Whether the request belongs to the API namespace
+    """
+    if request.resolver_match:
+        return request.resolver_match.app_name == "api"
+    # If the url could not be resolved at all, fall back to the url prefix of the API
+    return request.path.startswith("/api/")
```
andrew8er, line 31: ❓ Under which circumstances can this happen?

integreat_cms/cms/views/error_handler/error_handler.py
```diff
@@ -41,13 +49,27 @@ def render_error_template(context: dict[str, Any]) -> SafeString:
     return render_to_string("error_handler/http_error.html", context)
 
 
-def handler400(request: HttpRequest, exception: BadRequest) -> HttpResponseBadRequest:
+def render_error_response(context: dict[str, Any]) -> HttpResponse:
```
andrew8er: ❌ Since this function is only meant to be called within this file, it should be named `_render_error_response`.

andrew8er (approved):
LGTM with one trivial naming error.

integreat_cms/cms/views/error_handler/error_handler.py
bahaaTuffaha: For cases where there is no pattern matches for example:
`/api/regions/non-existing/`

PR #4489, opened 2026-08-05
fix: allow regular patch updates of the CircleCI runner image

### LINKED ISSUE #4488: CI: pip cache is not invalidated when the Python version changes
### Bug Summary

The pip cache key in `.circleci/config.yml` contains only the checksum of `pyproject.toml`, with no identifier for the Python interpreter the virtualenv was built against:

```yaml
- restore_cache:
    key: pip-{{ checksum "pyproject.toml" }}-v1
```

Since `pip-install` skips installing entirely when `.venv` already exists, any change of interpreter restores a `.venv` built by the *previous* one and never rebuilds it. The cache stays poisoned for every subsequent pipeline until someone manually bumps the `-v1` suffix.

This is currently masked, not fixed: the image is pinned to a full patch version (`cimg/python:3.13.11`), so the interpreter only ever changes when a human edits the tag. The cost of that mitigation is that we cannot patch the runner image on a regular schedule — which is how CVE fixes in the base image and the interpreter would otherwise reach CI.

### Expected Behavior

The Python patch level of the CI runner image can be updated regularly — ideally automatically — without anyone having to remember to invalidate the pip cache by hand. Any change of interpreter misses the cache and rebuilds the virtualenv.

### Actual Behavior

The virtualenv built by the old interpreter is reused under the new image. `pyvenv.cfg` and the `lib/python3.X/site-packages` path no longer match the interpreter in the image, so downstream jobs fail with import errors that point nowhere near the actual cause — an unrelated-looking `ModuleNotFoundError` in whichever job happens to run first. This was previously observed as `FileNotFoundError: /home/circleci/project/.venv/bin/integreat-cms-cli` in #1732.

The fix (bump the cache-key suffix) is invisible from the failure, so in practice the image tag is simply left alone.

### Steps to Reproduce

The bug is latent as long as the image tag is pinned and untouched. To trigger it:

1. Let a pipeline run on `develop` with `cimg/python:3.13.11`, so a cache is stored under `pip-<pyproject-checksum>-v1`.
2. Open a PR that changes the interpreter without touching `pyproject.toml` — either bump the tag to a new patch/minor, or switch it to a floating `cimg/python:3.13`.
3. Watch the `pip-install` job: the cache key is unchanged, the old `.venv` is restored, `[[ -d ".venv" ]]` is true, and the job logs `Virtual environment restored from cache, skipping pip install`.
4. Watch any downstream job that runs `.venv/bin/…`.

### Reproducibility

Always (100%)

### Environment

- **Environment:** CI only (CircleCI) — no runtime/deployment environment involved
- **Pipeline:** `digitalfabrik/integreat-cms`, all branches
- **Affected job:** `pip-install`, plus every downstream job that consumes the persisted `.venv`
- **Config:** `.circleci/config.yml` on `develop` @ 3636f6607
- **Executor image:** `cimg/python:3.13.11`
- **Browser / OS / Device / Screen / User role / Tenant:** n/a

_No response_

### Severity

### Regression Information

- **Is this a regression?** No — known and previously reported as #1732 (*"CircleCI cache of virtual environment outdated"*, closed 2022-10-26).
- **Previous mitigation:** d650b9d2d ("Downgrade CircleCI docker images to python 3.9.14") pinned the image to a full patch version. That does not invalidate the cache on an interpreter change — it merely makes the change rare, by coupling every patch bump to a manual decision.
- **Why it is still open in practice:** the mitigation holds only as long as the runner image is never patched independently of `pyproject.toml`, so it trades a maintenance practice we want for a cache-correctness property we should get for free.
- **Prior attempt at a real fix:** branch `circleci/python-version` (2022-10-27, three commits, never opened as a PR) tried deriving the version at runtime into the cache key via `$BASH_ENV` and `{{ .Environment.PYTHON_VERSION }}`. That does not work — cache-key interpolation does not see variables exported by an earlier step — and the branch was abandoned unfinished.
- **Related recent changes:** none — the trap survived the pipenv → pip/venv migration, which rewrote the cache key without adding an interpreter component.

### Affected Components / Services
- `.circleci/config.yml` — image tag at L9, cache key at L14 and L27

### Suspected Root Cause
- The cache key identifies only the dependency manifest, not the interpreter that resolved it.

### Proposed Fix
Key the cache on a file the job writes at runtime. `{{ checksum "…" }}` only requires the file to exist in the working directory when the cache step runs — it does not have to be tracked in git — and unlike `{{ .Environment.VAR }}` it does see values produced by an earlier `run` step. This makes a floating tag safe by construction:

```diff
   pip-install:
     docker:
-      - image: cimg/python:3.13.11
+      - image: cimg/python:3.13
     resource_class: small
     steps:
       - checkout
+      - run:
+          name: Record interpreter identity for the cache key
+          command: python3 -VV > .python-build-id
       - restore_cache:
-          key: pip-{{ checksum "pyproject.toml" }}-v1
+          key: pip-{{ checksum ".python-build-id" }}-{{ checksum "pyproject.toml" }}-v1
       …
       - save_cache:
-          key: pip-{{ checksum "pyproject.toml" }}-v1
+          key: pip-{{ checksum ".python-build-id" }}-{{ checksum "pyproject.toml" }}-v1
```

`python3 -VV` prints the full patch version plus build date and compiler string, so the key also moves when the image is rebuilt on the same Python patch — desirable, since a rebuilt base image can change the C libraries the wheels linked against.

### Notes and Trade-offs
- `.python-build-id` must be added to `.gitignore` so jobs that assert a clean tree stay clean. It does not leak downstream: `persist_to_workspace` lists explicit paths, and `pip-install` is the only job using this cache key.
- A floating tag means an upstream image patch can fail an unrelated PR. The failure mode is at least honest — the venv is rebuilt against the new interpreter and any real incompatibility surfaces as a real dependency error, instead of a phantom `ModuleNotFoundError` from a stale venv.
- Expect an occasional full reinstall (~1 min, a few times a year) when upstream patches the image.
- Optional cleanup: move the tag into a pipeline parameter, which also removes the ~15 duplicated `image:` lines and makes a minor-version bump a one-line change.

### Alternatives Considered
- **Keep the pin and automate the bumps.** Dependabot cannot do this — `.github/dependabot.yml` covers `github-actions`, `npm` and `devcontainers`, and Dependabot has no manager for Docker images inside `.circleci/config.yml`. Renovate's `circleci` manager can, but that means adopting a second bot.
- **Treat `-v1` as a manual cache epoch** and bump it with every image change. Status quo; unenforced and easy to forget.
- **Hardcode the version into the key** (`pip-3.13.11-…`). Visible in the diff of a bump, but still manual and still incompatible with a floating tag.
- **Stop caching `.venv`** and cache only `~/.cache/pip`. Correct by construction, at the cost of a venv rebuild in every pipeline. Largely obsolete if the `uv` migration lands, since `uv` keys its own cache by interpreter and builds venvs fast enough that caching `.venv` stops being worthwhile.

- [ ] Changing the interpreter without touching `pyproject.toml` misses the cache and reinstalls
- [ ] Unchanged interpreter + unchanged `pyproject.toml` still hits the cache (no loss of CI speed)
- [ ] The runner image can receive patch updates without any manual cache-invalidation step
- [ ] No leftover untracked files in the working directory of any job

### Verification / Testing

- [ ] On a scratch branch, change the image tag only → `pip-install` reinstalls instead of logging "skipping pip install"
- [ ] Re-run the same pipeline unchanged → second run restores from cache
- [ ] `git status` is clean in jobs that inspect the working tree (`check-translations`, `compile-translations`)
- [ ] Full pipeline green on both runs

### Related Issues

- **integreat-cms issues:**
  - #1732 — same root cause, closed 2022-10-26 by pinning the image to a full patch version rather than fixing the cache key

timobrembeck:
### Short description

The CircleCI pip cache key did not identify the Python interpreter, so a virtualenv built by a different interpreter could be restored and silently reused. Fixing the key allows the runner image to be referenced by its minor version (`cimg/python:3.13`) and thus to receive patch updates automatically.

### Proposed changes

- Write the output of `python3 -VV` to `.python-build-id` in the `pip-install` job and add its checksum to the pip cache key. Unlike variables exported to `$BASH_ENV`, files are visible to the interpolation of cache keys.
- Reference the Docker image as `cimg/python:3.13` instead of `cimg/python:3.13.11` in all 17 places. Since the virtualenv is built once in `pip-install` and handed to the other jobs via `persist_to_workspace`, all Python jobs have to resolve the same tag — otherwise the mismatch is merely moved somewhere else.
- Ignore `.python-build-id` and document the mechanism in the `pip-install` section of `docs/src/continuous-integration.rst`, including a warning not to retry the `$BASH_ENV` approach.

### Side effects

- The first pipeline after merging will do a full `pip install`, because the shape of the cache key changed and no cache exists under the new key. Subsequent runs are cached as before.
- Whenever the upstream image is patched, one pipeline pays for a full reinstall (~1 min). This is the intended behaviour, not a regression.
- A floating tag means an upstream image update can break an unrelated PR. The failure mode is at least honest now: the virtualenv is rebuilt against the new interpreter, so a real incompatibility surfaces as a real dependency error instead of a `ModuleNotFoundError` from a stale environment.
- If a new patch version is published while a pipeline is running, `pip-install` and a later job could theoretically resolve different patch versions. This is rare and resolves itself on a rerun, but it is the one guarantee the pinned tag gave us.
- `pyproject.toml` declares `requires-python = ">=3.13,<3.14"`, so the floating tag cannot drift out of the supported range.
- The `-v1` suffix of the cache key is kept: it is still the only way to manually evict a cache after changes which no checksum covers, e.g. a different set of `save_cache` paths or different extras in the install command.

### How to test

There is nothing to test locally, the change only affects CI. To verify the cache behaviour on this branch:

1. Let the pipeline run once, so a cache is written under the new key. `pip-install` performs a full install.
2. Re-run the pipeline without any change. `pip-install` logs `Virtual environment restored from cache, skipping pip install`, so caching still works.
3. Push a commit which changes only the image tag (e.g. to `cimg/python:3.12`), and confirm that `pip-install` installs from scratch instead of restoring the cache. Revert it afterwards.

Also worth a look in the job output of the first step:

```
python3 -VV > .python-build-id
```

The cached key in the `restore_cache`/`save_cache` step output should contain two checksums.

### Resolved issues

Fixes: #4488

Related: #1732 (same root cause, closed in 2022 by pinning the patch version instead of fixing the cache key)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,12 +6,15 @@ orbs:
 jobs:
   pip-install:
     docker:
-      - image: cimg/python:3.13.11
+      - image: cimg/python:3.13
     resource_class: small
     steps:
       - checkout
+      - run:
+          name: Record interpreter identity for the cache key
+          command: python3 -VV > .python-build-id
       - restore_cache:
-          key: pip-{{ checksum "pyproject.toml" }}-v1
+          key: pip-{{ checksum ".python-build-id" }}-{{ checksum "pyproject.toml" }}-v1
       - run:
           name: Install pip dependencies
           command: |
@@ -24,7 +27,7 @@ jobs:
               pip install -e .[dev-pinned,pinned]
             fi
       - save_cache:
-          key: pip-{{ checksum "pyproject.toml" }}-v1
+          key: pip-{{ checksum ".python-build-id" }}-{{ checksum "pyproject.toml" }}-v1
           paths:
             - .venv
             - integreat_cms.egg-info
@@ -36,7 +39,7 @@ jobs:
             - integreat_cms.egg-info
   djlint:
     docker:
-      - image: cimg/python:3.13.11
+      - image: cimg/python:3.13
     resource_class: small
     steps:
       - checkout
@@ -50,7 +53,7 @@ jobs:
           command: djlint --check --lint integreat_cms
   ruff:
     docker:
-      - image: cimg/python:3.13.11
+      - image: cimg/python:3.13
     resource_class: small
     steps:
       - checkout
@@ -66,7 +69,7 @@ jobs:
             ruff format --check
   mypy:
     docker:
-      - image: cimg/python:3.13.11
+      - image: cimg/python:3.13
     resource_class: small
     steps:
       - checkout
@@ -77,7 +80,7 @@ jobs:
           command: ./tools/mypy.sh
   check-release-notes:
     docker:
-      - image: cimg/python:3.13.11
+      - image: cimg/python:3.13
     resource_class: small
     steps:
       - checkout
@@ -97,7 +100,7 @@ jobs:
           command: ./tools/make_release_notes.sh --format=raw --all
   check-translations:
     docker:
-      - image: cimg/python:3.13.11
+      - image: cimg/python:3.13
     resource_class: small
     steps:
       - checkout
@@ -111,7 +114,7 @@ jobs:
           command: ./tools/check_translations.sh
   compile-translations:
     docker:
-      - image: cimg/python:3.13.11
+      - image: cimg/python:3.13
     resource_class: small
     steps:
       - checkout
@@ -222,7 +225,7 @@ jobs:
             - integreat_cms/webpack-stats.json
   check-migrations:
     docker:
-      - image: cimg/python:3.13.11
+      - image: cimg/python:3.13
       - image: cimg/postgres:17.10
         environment:
           POSTGRES_USER: integreat
@@ -244,7 +247,7 @@ jobs:
           command: integreat-cms-cli makemigrations cms --check --settings=integreat_cms.core.circleci_settings
   test:
     docker:
-      - image: cimg/python:3.13.11
+      - image: cimg/python:3.13
       - image: cimg/postgres:17.10
         environment:
           POSTGRES_USER: integreat
@@ -296,7 +299,7 @@ jobs:
           path: integreat-cms.log
   build-package:
     docker:
-      - image: cimg/python:3.13.11
+      - image: cimg/python:3.13
     resource_class: small
     steps:
       - checkout
@@ -317,7 +320,7 @@ jobs:
             - dist
   publish-package:
     docker:
-      - image: cimg/python:3.13.11
+      - image: cimg/python:3.13
     resource_class: small
     steps:
       - checkout
@@ -331,7 +334,7 @@ jobs:
           command: twine upload --non-interactive ./dist/integreat_cms-*.tar.gz
   build-documentation:
     docker:
-      - image: cimg/python:3.13.11
+      - image: cimg/python:3.13
     resource_class: large
     steps:
       - checkout
@@ -357,7 +360,7 @@ jobs:
             - docs/dist
   deploy-documentation:
     docker:
-      - image: cimg/python:3.13.11
+      - image: cimg/python:3.13
     resource_class: small
     environment:
       BRANCH: gh-pages
@@ -401,7 +404,7 @@ jobs:
             git push origin $BRANCH
   bump-dev-version:
     docker:
-      - image: cimg/python:3.13.11
+      - image: cimg/python:3.13
     resource_class: large
     steps:
       - checkout
@@ -451,7 +454,7 @@ jobs:
             - integreat_cms/_manifest
   bump-rc-version:
     docker:
-      - image: cimg/python:3.13.11
+      - image: cimg/python:3.13
     resource_class: large
     steps:
       - checkout
@@ -502,7 +505,7 @@ jobs:
             - integreat_cms/_manifest
   bump-version:
     docker:
-      - image: cimg/python:3.13.11
+      - image: cimg/python:3.13
     resource_class: large
     steps:
       - checkout
@@ -582,7 +585,7 @@ jobs:
           command: git checkout develop && git merge main --commit --no-edit && git push
   create-release:
     docker:
-      - image: cimg/python:3.13.11
+      - image: cimg/python:3.13
     resource_class: small
     steps:
       - checkout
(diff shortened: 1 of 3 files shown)


andrew8er (approved):
Implementation looks good.

I wonder however, if the intended effect is not more easily (and more comprehensively) solved by using a proper lockfile. A lockfile solves exactly this problem but it includes _every_ detail of all dependencies, which this method lacks. E.g. it also includes the detailed download URL for wheels that are platform specific.

Pip has experimental lockfile support. Uv has lockfile support since basically forever. Every sane package manager these days has lockfile support.

timobrembeck:
> Implementation looks good.
> 
> I wonder however, if the intended effect is not more easily (and more comprehensively) solved by using a proper lockfile. A lockfile solves exactly this problem but it includes _every_ detail of all dependencies, which this method lacks. E.g. it also includes the detailed download URL for wheels that are platform specific.
> 
> Pip has experimental lockfile support. Uv has lockfile support since basically forever. Every sane package manager these days has lockfile support.

Sure, I think migrating our custom pyproject.toml dependency locking to a proper lockfile like uv would be a great improvement! It would however not resolve this issue, since this is about unpinning and automatically updating the CI runner image, and invalidating the dependency cache when the python version of the CI runner image is updated.

andrew8er:
I could imagine a system, where Python is installed via uv (uv can install Python as well) or [mise](https://mise.en.dev/), so the development setup always matches exactly the one of the CI environment _and_ the distributed artifacts. This would also remove the need to change the base image in 17 different places. We could just use one generic base image.

Including the runtime in the distribution would make us more independent of the production system, where we are currently stuck with Python 3.13 (from Debian Trixie). Kind of like running a Docker image that also works locally.

seluianova (approved):
caching works, lgtm!

Image

PR #4513, opened 2026-08-12
Critical dependency updates

andrew8er:
### Short description
Updates Python and NPM packages to resolve vulnerability alerts.

### Proposed changes

- Updates most Python packages (see below for exceptions)
- Updates most NPM packages (see below for exceptions)
- Enable dependabot updates of `maplibre-gl` again.
- Fix some newly appearing linter/formatter/type checker errors.

### Side effects
None

### How to test

- Run tests
- Check icons, POI map

#### Python dependencies

aiohttp 3.14.1 has 6 known vulnerabilities:

- GHSA-mfx4-hv73-q22v: AIOHTTP: HTTP request smuggling via WebSocket upgrade

  Fixed in: 3.14.2

  Advisory information: https://github.com/aio-libs/aiohttp/security/advisories/GHSA-mfx4-hv73-q22v

- GHSA-cq5v-8q36-5273: AIOHTTP: Out-of-bounds heap read in C HTTP response parser error path (malformed chunked response)

  Fixed in: 3.14.3

  Advisory information: https://github.com/aio-libs/aiohttp/security/advisories/GHSA-cq5v-8q36-5273

- GHSA-mq44-7p77-q5h7: AIOHTTP: WebSocket client accepts compressed frames without negotiated permessage-deflate

  Fixed in: 3.14.2

  Advisory information: https://nvd.nist.gov/vuln/detail/CVE-2026-59881

- PYSEC-2026-3545: AIOHTTP: Out-of-bounds heap read in C HTTP response parser error path (malformed chunked response)

  Fixed in: 3.14.3

  Advisory information: https://github.com/advisories/GHSA-cq5v-8q36-5273

- PYSEC-2026-3547: AIOHTTP: WebSocket client accepts compressed frames without negotiated permessage-deflate

  Fixed in: 3.14.2

  Advisory information: https://nvd.nist.gov/vuln/detail/CVE-2026-59881

- PYSEC-2026-3546: AIOHTTP: HTTP request smuggling via WebSocket upgrade

  Fixed in: 3.14.2

  Advisory information: https://github.com/advisories/GHSA-mfx4-hv73-q22v

cryptography 49.0.0 has 2 known vulnerabilities:

- PYSEC-2026-3552: cryptography: PKCS#7 EnvelopedData decryption exposes a Bleichenbacher oracle through distinguishable errors and timing

  Fixed in: 50.0.0

  Advisory information: https://github.com/advisories/GHSA-g6cj-pr64-35w5

- GHSA-g6cj-pr64-35w5: cryptography: PKCS#7 EnvelopedData decryption exposes a Bleichenbacher oracle through distinguishable errors and timing

  Fixed in: 50.0.0

  Advisory information: https://github.com/pyca/cryptography/security/advisories/GHSA-g6cj-pr64-35w5

pypdf 6.14.2 has 4 known vulnerabilities:

- PYSEC-2026-3656: pypdf: Possible long runtimes/large memory usage for large CID font width ranges

  Fixed in: 6.15.0

  Advisory information: https://github.com/advisories/GHSA-fwg2-594c-jp42

- GHSA-fwg2-594c-jp42: pypdf: Possible long runtimes/large memory usage for large CID font width ranges

  Fixed in: 6.15.0

  Advisory information: https://github.com/py-pdf/pypdf/security/advisories/GHSA-fwg2-594c-jp42

- GHSA-fp3f-mc75-235c: pypdf: Possible large memory usage for large /ToUnicode streams

  Fixed in: 6.15.0

  Advisory information: https://github.com/py-pdf/pypdf/security/advisories/GHSA-fp3f-mc75-235c

- PYSEC-2026-3655: pypdf: Possible large memory usage for large /ToUnicode streams

  Fixed in: 6.15.0

  Advisory information: https://github.com/advisories/GHSA-fp3f-mc75-235c

#### NPM dependencies

brace-expansion  <=1.1.17 || 4.0.0 - 5.0.8
Severity: high
brace-expansion: DoS via unbounded expansion length causing an out-of-memory process crash - https://github.com/advisories/GHSA-mh99-v99m-4gvg
brace-expansion: DoS via unbounded expansion length causing an out-of-memory process crash - https://github.com/advisories/GHSA-mh99-v99m-4gvg
brace-expansion: DoS via unbounded intermediate arrays, bypassing the CVE-2026-14257 mitigation - https://github.com/advisories/GHSA-rgw5-rvv9-x895
brace-expansion: DoS via unbounded intermediate arrays, bypassing the CVE-2026-14257 mitigation - https://github.com/advisories/GHSA-rgw5-rvv9-x895
fix available via `npm audit fix`
node_modules/brace-expansion
node_modules/glob/node_modules/brace-expansion

fast-uri  3.0.0 - 3.1.4
Severity: high
fast-uri vulnerable to host confusion via literal backslash authority delimiter - https://github.com/advisories/GHSA-v2hh-gcrm-f6hx
fast-uri vulnerable to host confusion via backslash authority introducer - https://github.com/advisories/GHSA-7p8r-x3mc-p8w7
fix available via `npm audit fix`
node_modules/fast-uri

js-yaml  4.0.0 - 4.3.0
Severity: high
JS-YAML: Quadratic CPU consumption in !!omap resolution (3.x and 4.x) — CVE-2026-59870 fix not backported - https://github.com/advisories/GHSA-5p4m-2wfm-xmqj
fix available via `npm audit fix`
node_modules/js-yaml

nanoid  <3.3.17
Severity: high
nanoid: custom generators can loop indefinitely when size is zero - https://github.com/advisories/GHSA-2v37-7h3g-55p8
fix available via `npm audit fix`
node_modules/nanoid

svgo  4.0.0 - 4.0.1
Severity: high
SVGO removeScripts plugin leaves some executable scripts intact - https://github.com/advisories/GHSA-2p49-hgcm-8545
fix available via `npm audit fix`
node_modules/svgo

5 high severity vulnerabilities

#### Python dependency notes
 - xhtml2pdf kept at 0.2.17 → reportlab stays at 4.5.1. reportlab 5.0.0 is available but xhtml2pdf doesn't support it yet; keeping the actively-used xhtml2pdf at latest is the better tradeoff.
 - build==1.5.0 (not 1.5.1) — 1.5.1 was yanked on PyPI.
 - django-treebeard 5.3.0 → 7.0.0 is a major version jump — worth a targeted test of the page/tree functionality.
 - Package-set changes from the fresh resolution: dropped decorator, sgmllib3k, and pip (pip isn't a project dependency); added feedparser-sgmllib (feedparser 6.0.14 swapped its sgmllib dep); colorama moved into the runtime pinned set and pywin32-ctypes was added to dev-pinned (universal resolution).

##### tailwindcss 3 → 4 not updated:

- Sass incompatibility: Tailwind 4 officially does not support running inside a Sass/PostCSS-preprocessor pipeline (it's meant to replace Sass). This project's CSS is a 727-line style.scss using @use, @layer, nesting and @import "tailwindcss/base|components|utilities" — all of which change in v4 (@tailwindcss/postcss plugin, @import "tailwindcss", @theme/@config, new @source-based content detection).
- Template-wide class changes: v4 renames/removes utilities (e.g. flex-shrink→shrink, bg-opacity-*, etc.), affecting the Tailwind classes across all the Django templates.

##### maplibre-gl updated 5.24.0 → 6.3.0. 

Build passes with TypeScript type-checking clean (the v6 API surface used here is unchanged).

Two things of note:
- v6 is ESM-only, and this project's TS pipeline emits CommonJS, so webpack couldn't match maplibre's exports conditions. I added a targeted resolve.alias in `webpack.config.js` pointing the bare maplibre-gl import at its .mjs entry (scoped with $ so the maplibre-gl/dist/maplibre-gl.css import is unaffected).
- The remaining Critical dependency: the request of a dependency is an expression warning comes from maplibre's internal worker-URL handling — a well-known benign warning when bundling maplibre-gl with webpack; it doesn't affect functionality.

##### typescript updated 5.9.3 → 6.0.3

the latest usable version. Standalone tsc --noEmit, the webpack build's type check, and the tests all pass clean.

Why not 7.0.2 (the absolute latest): TypeScript 7 is the new native compiler, and the two tools that consume TypeScript's API don't support it yet — both already at their latest releases:
 - typescript-eslint 8.67.0 → peer typescript <6.1.0
 - typedoc 0.28.20 → peer typescript 6.0.x

So 6.0.3 is the highest version compatible with the current lint + docs toolchain. Once typescript-eslint and typedoc ship TS 7-compatible releases, we can revisit the jump to 7.


integreat_cms/cms/templates/_base.html
```diff
@@ -347,13 +340,9 @@
             <a href="{% if request.region %}{% url "release_notes" region_slug=request.region.slug %}{% else %}{% url "release_notes" %}{% endif %}"
```
PeterNerlich: TODO: test whether single quotes work and make the syntax parsing easier

```suggestion
            <a href="{% if request.region %}{% url 'release_notes' region_slug=request.region.slug %}{% else %}{% url "release_notes" %}{% endif %}"
```

integreat_cms/cms/templates/imprint/imprint_form.html
```diff
@@ -185,17 +159,17 @@ <h3 class="heading font-bold text-black">
...
-    </form>
-    {{ media_config_data|json_script:"media_config_data" }}
-    {% if not perms.cms.change_imprintpage %}
-        {% include "../_tinymce_config.html" with readonly=1 %}
-    {% else %}
-        {% include "../_tinymce_config.html" %}
-        {% include "../_content_edit_lock.html" %}
-    {% endif %}
-    {% include "../generic_confirmation_dialog.html" %}
+        {% include "../generic_confirmation_dialog.html" %}
```
PeterNerlich, line 148: I don't quite understand this change in indentation. Was this automatically generated by some rules or is this an oversight?
```suggestion
            </div>
        {% endif %}
    </form>
    {{ media_config_data|json_script:"media_config_data" }}
    {% if not perms.cms.change_imprintpage %}
        {% include "../_tinymce_config.html" with readonly=1 %}
    {% else %}
        {% include "../_tinymce_config.html" %}
        {% include "../_content_edit_lock.html" %}
    {% endif %}
    {% include "../generic_confirmation_dialog.html" %}
```

integreat_cms/cms/templates/push_notifications/push_notification_form.html
```diff
@@ -58,23 +58,23 @@ <h1 class="heading overflow-hidden text-ellipsis">
...
                                     {% else %}
-                                    </div>
-                                {% endif %}
+                                        </div>
+                                    {% endif %}
+                                </div>
                             </div>
-                        </div>
-                    </li>
+                        </li>
```
PeterNerlich, line 77: Was this generated by rules? I think this latter part should also have another indentation removed, just as the *endif* just before
```suggestion
                                <span class="flex-grow language-name">{{ other_language.translated_name }}</span>
                                <span class="flex-end w-8 fp fp-rounded fp-{{ other_language.primary_country_code }}"></span>
                                {% if other_language.secondary_country_code %}
                                    <span class="fp ml-1 w-8 fp-rounded fp-{{ other_language.secondary_country_code }}"></span>
                                {% endif %}
                                {% if other_language != language %}
                                    </a>
                                {% else %}
                                    </div>
                                {% endif %}
                            </div>
                        </div>
                    </li>
```

PeterNerlich (commented):
Thank you so much!

This Review carefully examined the HTML/Template files and can attest functional equivalence.
I only have two very minor comments about indentation, of all things 😅 

I didn't look too closely at the actual dependencies. The server starts and content editing, using the map, the statistics diagrams all pass my spot checking.

jonbulz (commented):
I actually found an issue:
In `treebeard 7`, the `move`, `add_child`, `add_root` etc. methods moved from the node model to the manager. Thus, we need to move our cache invalidation logic  to match this change.

Context: `treebeard` updates nodes with a plain `Queryset.update()`. By itself, that does not invalidate the cache. Thus, to prevent a stale cache, we currently override the methods in `abstract_tree_node.py` and add cache invalidation. 
With `treebeard 7`, the calls switch to the manager methods, thus we need to add cache invalidation on those methods as well. Otherwise, the updates can base off a stale cache and mess up the page tree. A way to provoke this is:

- With caching enabled, go to a page tree with
``` 
- Parent
   - Page A
   - Page B
   - Page C
 ```
- warm up the cache with ```curl 'http://localhost:8000/api/v3/<region>/<lang>/children/?id=<page B id>&depth=1```
- add a child page to page A by creating a page through the form
- move page C up in the page tree, above B
- see that page C is not above B: it has been moved into page A's subtree, as a child of the page that was just created. `Page.objects.find_problems()` now reports it (4th list = wrong depth)

Note that warming the cache by opening page B in the CMS does not work, because the edit view fetches the page with prefetches and therefore caches under a different key.
The damage is permanent. `repair_tree --commit` will 'fix' the mismatch by correcting the depth, not by moving page C to where it was supposed to be, because there is no way to recover this information.

This is actually sort of pre-existing on develop, because our current cache invalidation does not cover the `delete` case. But with moving to `treebeard 7`, the surface of this bug would become much bigger.
I'd propose that I'll introduce a fix for the existing bug that is forward compatible with `treebeard 7`. The fix would then preferably land before this branch is merged, or at least before these changes are released.

File: integreat_cms/cms/templates/imprint/imprint_form.html (hunk 2 above)
Comment (andrew8er) [reply integreat_cms/cms/templates/imprint/imprint_form.html:148]: Yes, the files were reformatted with `djlint --reformat`. I know that the formatting is pretty borked in some situations, but that's just how djlint does it.

File: integreat_cms/cms/templates/push_notifications/push_notification_form.html (hunk 3 above)
Comment (andrew8er) [reply integreat_cms/cms/templates/push_notifications/push_notification_form.html:77]: What do you mean by "rules". I did not bother fighting djlint's formatting, I just used the preexisting config and updated the formatting, according to the new djlint version.

**Comment by andrew8er** [comment]
@jonbulz Good find. Then let's wait for your PR to be merged before merging this.

**Review by PeterNerlich — APPROVED** [review]
Attesting functional equivalence of HTML and python files 👍

File: pyproject.toml [hunk 4]
```diff
@@ -210,61 +210,59 @@ pinned = [
...
-    "yarl==1.24.2",
+    "yarl==1.24.5",
 ]
 dev-pinned = [
     "alabaster==1.0.0",
-    "ast_serialize==0.6.0",
+    "ast_serialize==0.8.0",
     "babel==2.18.0",
-    "build==1.5.1",
+    "build==1.5.0",
```
jonbulz: is the downgrade intentional?

jonbulz (approved):
What I tested:
- quick regression tests with test data
- quick check with a prod db dump
- moving tree nodes works as expected
- local test suite runs green

The one downgrade I flagged looks like an oversight to me. But either way, everything looks like it works as expected. Nice work, thanks! :)

pyproject.toml
andrew8er: Possibly a rebase error. Fixed now.

PR #4528, opened 2026-08-20
invalidate cache on treebeard node deletion and manager methods

jonbulz:
### Short description

`treebeard` uses `Queryset.update()` to update nodes when changing a tree, e.g. by moving a page in the page tree or adding a new child page. This does not invalidate the cache, thus we override the methods in `abstract_tree_node.py` to add cache invalidation manually. This PR serves two purposes:
- close the existing gap that node deletion was not covered in our cache invalidation
- make the overrides forward compatible with `treebeard 7`, which will be introduced by #4513 

### Proposed changes

- override the `delete` method of `NS_NodeQuerySet` to add cache invalidation
- override `add_child`, `add_sibling` and `move` methods on `NS_NodeManager` to add cache invalidation for `treebeard >= 7`
- introduce AbstractTreeNode.invalidate_tree_cache() as the single hook all layers call, replacing the inline `invalidate_model(self.__class__)` calls in the model methods
- re-base PageQuerySet on AbstractTreeNodeQuerySet and PageManager on AbstractTreeNodeManager (previously NS_NodeQuerySet / NS_NodeManager) because Page.objects is a PageManager, so overriding only the abstract manager would fix language trees and leave pages uncovered
- AbstractTreeNodeManager.get_queryset() returns the new queryset class
- remove stale `Page.move` method
- add regression tests

### Side effects

- page translations are now invalidated on add_child, add_sibling and delete too, not only on move as before. Slightly more cache churn
- PageManager now inherits AbstractTreeNodeManager, so Page.objects.create() now emits "Don't use Page.objects.create()" dev warning it previously escaped

### How to test

Start the CMS with the cache enabled and the test data loaded:

    INTEGREAT_CMS_REDIS_CACHE=True ./tools/run.sh

Start with a page tree like
```
...
- page A
- page B
...
```

**Warm the cache for page B**:

    curl -s 'http://localhost:8000/api/v3/augsburg/de/children/?id=<page B id>&depth=1' > /dev/null

**Delete page A** via bulk delete.
Compare cache and database:

    INTEGREAT_CMS_REDIS_CACHE=True integreat-cms-cli shell -c "
    from integreat_cms.cms.models import Region
    region = Region.objects.get(slug=<region>)
    print('cache   ', region.pages.get(id=<page B id>).lft)
    print('database', region.pages.filter(id=<page B id>).nocache().get().lft)
    "

Before this PR the two disagree, with the fix they match.
See this [comment](https://github.com/digitalfabrik/integreat-cms/pull/4513#pullrequestreview-4983113978) on how to test the bug with treebeard 7 without the fix.

### Resolved issues

Fixes: #

### DIFF
--- a/integreat_cms/cms/models/abstract_tree_node.py
+++ b/integreat_cms/cms/models/abstract_tree_node.py
@@ -16,12 +16,10 @@ from django.db.models import CheckConstraint, Deferrable, F, Q, UniqueConstraint
 from django.utils.functional import cached_property
 from django.utils.translation import gettext_lazy as _
 from treebeard.exceptions import InvalidPosition
-from treebeard.ns_tree import NS_Node, NS_NodeManager
+from treebeard.ns_tree import NS_Node, NS_NodeManager, NS_NodeQuerySet
 
 if TYPE_CHECKING:
     from typing import Any, Self
-
-    from treebeard.ns_tree import NS_NodeQuerySet
 
 from ..constants import position
 from .abstract_base_model import AbstractBaseModel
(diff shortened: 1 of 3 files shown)


jonbulz:
tests have failed for the last few commits on develop. maybe a change shard composition issue?

jonbulz:
failing tests pass after a rebase. ready for review

PeterNerlich (approved):
Makes sense to me and the provided test instructions yielded good results.

andrew8er (approved):
Seems to be working for me:

```
In [2]: from integreat_cms.cms.models import Region

In [3]: region = Region.objects.get(slug='augsburg')

In [4]: print('cache   ', region.pages.get(id=14).lft)
cache    1

In [5]: print('database', region.pages.filter(id=14).nocache().get().lft)
database 1
```

Issue #2021, opened 2023-01-23
Bullet points are on the wrong side for RTL languages in PDF export


seluianova:
### Describe the Bug
Bullet points for right-to-left languages are on the left.
![image](https://user-images.githubusercontent.com/115008338/214126494-53f310d0-c570-4bbd-a98b-1b2d48e5449d.png)

### Steps to Reproduce

1. Go to 'Pages'
2. Select any right-to-left language, for example Arabic
3. Select any page containing a bulleted list
4. Select bulk action --> Export as PDF
5. Check the exported PDF

### Expected Behavior
Bullet points are on the right
![image](https://user-images.githubusercontent.com/115008338/214126949-af9a9e9c-e313-48b8-a769-a97583691848.png)

### Actual Behavior
Bullet points are on the left

### Additional Information
The issue should be probably fixed in xhtml2pdf library:
https://github.com/xhtml2pdf/xhtml2pdf/issues/663

Within this task we can update the library version and our tests as soon as the issue is fixed.

Within this task we can update the library version and our tests as soon as the issue is fixed.

osmers:
Stale and blocked and considering other options for PDF export

Issue #2020, opened 2023-01-23
Wrong text direction in ToC for RTL languages in PDF export


seluianova:
### Describe the Bug
ToC has wrong text direction for right-to-left languages in PDF export
![image](https://user-images.githubusercontent.com/115008338/214123204-d734dad3-b2b7-4c20-9bc8-e56854b2d0c7.png)

### Steps to Reproduce

1. Go to 'Pages'
2. Select any right-to-left language, for example Arabic
3. Select any parent page containing child pages, for example "German language"
4. Select bulk action --> Export as PDF
5. Check Table of Content in the exported PDF

### Expected Behavior
Table of Content has right-to-left text direction (as well as pages themselves)

### Actual Behavior
Table of Content has left-to-right text direction

### Additional Information
ToC is generated by xhtml2pdf
`<div dir="rtl">`
`  <pdf:toc/>`
`</div>`

The issue should be probably fixed on their side:
https://github.com/xhtml2pdf/xhtml2pdf/issues/664

Within this task we can update the library version and our tests as soon as the issue is fixed.

osmers:
Stale and blocked and considering other options for PDF export

PR #4541, opened 2026-08-28
2908: Replace xhtml2pdf with weasyprint

### LINKED ISSUE #2908: Consider temporary alternative PDF export
### Motivation

Due to bugs which need to be fixed in the external repository, there are problem with PDF export for some languages (mostly RTL languages) and we have disabled the PDF export bulk actionfor those languages (see #2890 ). In the discussion there it was suggested to use in-browser PDF printing as temporary solution until the bugs in the repository are fixed. We might even want to move away from the current library completely and use another solution permanently

### Proposed Solution/Scope of Issue

Evaluate in-browser PDF printing, or consider another alternativ (check out https://github.com/Kozea/WeasyPrint as mentioned below in the comments).
Timebox: 2 hours
Output: Recommendation on In-Browser printing or other alternative, new issue to implement recommendation

bahaaTuffaha:
### Short description
Due to bugs which need to be fixed in the external repository, there are problem with PDF export for some languages (mostly RTL languages) and we have disabled the PDF export bulk action for those languages.

### Proposed changes

- Replaced `xhtml2pdf` with `weasyprint` in the dependencies and dropped `reportlab`, which was only needed by the old renderer.
- Installed `libpango-1.0-0`, `libpangoft2-1.0-0` and `libharfbuzz0b` in `.circleci/config.yml`, `Dockerfile.test`, `README.md`, `installation.rst` and `prod-server.rst`, and added `pango` and `harfbuzz` to `LD_LIBRARY_PATH` in `flake.nix`, since WeasyPrint opens them at runtime.
- Replaced the `link_callback` function with a `PdfUrlFetcher` class and a `resolve_pdf_uri` helper. WeasyPrint expects a `URLFetcherResponse` instead of a path string, so local static and media files are read from disk and wrapped in one, while everything else is delegated to the parent fetcher via `super().fetch()`.
- At `pdf_utils.py` I split the `generate_pdf` into `render_pdf_html`, `build_pdf_filename` and `write_pdf`, and moved the `save` after the rendering, so a failed export no longer leaves an empty file behind that `pdf_storage.exists` would serve from the cache forever.
- Switched RTL PDFs from `DejaVu` to `Lateef` and latin body text from `DejaVu` to `Open Sans`, which WeasyPrint can load (xhtml2pdf could not). `zh` / `am` / `ti` / `ka` keep `Noto Sans SC`, `Noto Sans Ethiopic`, and `Noto Sans Georgian`.
- Added dotted leaders and page numbers to TOC entries in `pdf_page_export.css` with `leader(".")` and `target-counter(attr(href), page)`.
- Passed `presentational_hints=True` to `write_pdf`, so the `width` and `height` attributes on images are applied instead of ignored.
- Dropped the xhtml2pdf tags (`<pdf:toc />`, `<pdf:pagenumber />`, `<pdf:language />`, `<pdf:nextpage />`) from `page_pdf.html` and replaced them with an HTML table of contents, `dir` on `<html>`, and CSS page numbers.
- Added `lateef-pdf.css` so right-to-left text uses `Lateef` instead of `DejaVu Sans`, which is the main fix of this issue.
- Added `bookmark-level` rules for `.level-N > h1`, `h2`–`h6` and `#title_page` to restore the nested outline and keep content headings out of it.
- Added `:lang()` rules for the `arabic-indic` and `persian` counter styles, covering all six right-to-left languages of `integreat-app`.
- Moved the `level-` class onto the `<section>` and added `padding` rules per `.level-N` to indent pages by their depth in the page tree.
- Added `[dir="rtl"]` counterparts for the `.toc` and `.level-N` indentation, because `postcss-preset-env` rewrites logical properties like `padding-inline-start` to their left-to-right equivalent.
- Increased the font size for `.font-rtl` up to `1.3em`, since `Lateef` renders smaller than the latin fonts.
- Dropped the `artbox`, `bleedbox`, `cropbox`, and `get_contents()` checks in `test_pdf_export`. They compared xhtml2pdf's internal page boxes and drawing stream, not the exported document; `mediabox` and `extract_text()` already cover page size and content
- Regenerated the reference PDFs in `tests/pdf/files/`.

### Side effects

- I hope none.

### How to test
I generated a new pdfs under `tests/pdf/files/` so we can compare the ones from develop with the ones at this branch:
So to make it easier I created these copy commands to copy the pdfs to a place for example at /desktop:
on develop:
```
mkdir -p ~/Desktop/integreat-pdfs-old
```
```
for pdf in tests/pdf/files/*/*.pdf; do cp "$pdf" ~/Desktop/integreat-pdfs-old/; done
```

on this branch:
```
mkdir -p ~/Desktop/integreat-pdfs-new
```
```
for pdf in tests/pdf/files/*/*.pdf; do cp "$pdf" ~/Desktop/integreat-pdfs-new/; done
```

### Resolved issues

Fixes: #2908


seluianova:
Minor finding (if not quickly fixable, I think we could live with it) - the bulleted and numbered lists are slightly shifted to the left:

image

seluianova:
Another regression change (not sure if it is a problem though, just noticing as a change to consider):
If a page contains an image, the image is clickable in the PDF and opens in the same tab when the user clicks it.
_(The images were not clickable before)_

seluianova (approved):
Tested different languages and formatting. The summary of my findings:

* the fonts are different
* bulleted and numbered lists are slightly shifted to the left ⁉️
* images are now clickable and open in the same tab when clicked ⁉️
* the existing RTL issues are fixed ✅

Overall, the changes look good to me. 
_Not sure whether we need to fix the issues mentioned above, either now, separately, or at all._

bahaaTuffaha:
> * images are now clickable and open in the same tab when clicked

I can't click on images I even tried to open the pdf from the browser and also not clickable.
I tested it on this logo:
Screenshot 2026-09-07 at 7 34 49 PM

seluianova:
> I can't click on images I even tried to open the pdf from the browser and also not clickable.
I tested it on this logo

I meant images in the page content, when you add an image in the editor

bahaaTuffaha:
> > I can't click on images I even tried to open the pdf from the browser and also not clickable.
> > I tested it on this logo
> 
> I meant images in the page content, when you add an image in the editor

Screenshot 2026-09-08 at 1 19 55 PM

It automatically added a link to it from the editor... I just removed the link and it's not clickable anymore in the PDF. so I think this is expected.

seluianova:
> It automatically added a link to it from the editor...

yes but pictures are not clickable on the current develop

integreat_cms/cms/templates/pages/page_pdf.html
```diff
@@ -3,62 +3,16 @@
...
 {% load pdf_filters %}
 {% load render_bundle from webpack_loader %}
-{% comment %}
-    These file does not inherit from _raw.html,
-    because xhtml2pdf does not work in combination with tailwind css,
-    so that no resources like tailwind from _raw.html can be incorporated.
-    Tailwind css uses the @media without a mediatype, but xhtml2pdf expects the media identifier.
-    See https://github.com/xhtml2pdf/xhtml2pdf/issues/498 for more information.
-    Once this issue is fixed, we can use tailwind in the pdf context.
-{% endcomment %}
```
hannaseithe, line 13: We are still not using _raw.html. What is the reason for not using _raw.html again and thereby switching back to using tailwind? 

If there is a good reason still for not using _raw.html it should be stated here I think. Otherwise switching to using _raw.html should be considered I think

hannaseithe:
In the issue it says, that rtl languages are disabled for pdf export. And in the settings.py we have 

```
PDF_DEACTIVATED_LANGUAGES: Final[list[str]] = [
    slug
    for slug in os.environ.get("INTEGREAT_CMS_PDF_DEACTIVATED_LANGUAGES", "").split(",")
    if slug
]

```
But it seems this env var was never set in the salt repo (git log --all -S "PDF_DEACTIVATED_LANGUAGES" --oneline). So I am confused: Did we actually never disable any languages or is it defined elsewhere?

This is the PR that added the line above: #2902 

@team-cms

bahaaTuffaha:
> In the issue it says, that rtl languages are disabled for pdf export. And in the settings.py we have
> 
> ```
> PDF_DEACTIVATED_LANGUAGES: Final[list[str]] = [
>     slug
>     for slug in os.environ.get("INTEGREAT_CMS_PDF_DEACTIVATED_LANGUAGES", "").split(",")
>     if slug
> ]
> ```
> 
> But it seems this env var was never set in the salt repo (git log --all -S "PDF_DEACTIVATED_LANGUAGES" --oneline). So I am confused: Did we actually never disable any languages or is it defined elsewhere?
> 
> This is the PR that added the line above: #2902
> 
> @team-cms

I checked the web app we actually disabling it using RTL language check (`hasRTLScript()`) not from a CMS flag 😅

.circleci/config.yml
```diff
@@ -85,7 +85,7 @@ jobs:
           at: .
       - run:
           name: Install requirements
-          command: sudo apt-get update && sudo apt-get install pcregrep libcairo2
+          command: sudo apt-get update && sudo apt-get install pcregrep libcairo2 libpango-1.0-0 libpangoft2-1.0-0 libharfbuzz0b
```
hannaseithe, line 88: Have you actually checked that we need this for all the jobs you added it to?  From a quick overview I can only see that the following jobs need it:
- test - obviously
- check-migrations, check-translations -> management commands trigger a system check which eventually tries importing pdf_export
- build-documentation -> Sphinx imports every module

But the rest I could not quickly find why they need it

hannaseithe:
> I checked the web app we actually disabling it using RTL language check (`hasRTLScript()`) not from a CMS flag 😅


So once this PR is released, the app can go ahead any enable pdf export for rtl languages? Is there an issue already for this in the app repo?

hannaseithe (changes requested):
Thank you very much for working on this. The pdf rendering works and they look good.

1.) I have two legit (change) requests in regards to the new system libraries:

- we will need a PR on the salt repo that installs these new libraries there on all the servers concerned - check `states/integreat-cms/init.sls` in the salt repo (this should be merged before we merge this PR into develop otherwise the Testsystem server will crash)

- I think the Dockerfile for the .devcontainer needs to also be adapted to install the new system libraries

2.) Depending on whether this `PDF_DEACTIVATED_LANGUAGES` is set somewhere, that should also be undone (wherever it was set). And regardless of that, I am not sure we need the the whole `PDF_DEACTIVATED_LANGUAGES` & `language.can_be_pdf_exported` ... but I would wait for someone else from the team to reply to my question there, to make a decision on that.

bahaaTuffaha:
> > I checked the web app we actually disabling it using RTL language check (`hasRTLScript()`) not from a CMS flag 😅
> 
> So once this PR is released, the app can go ahead any enable pdf export for rtl languages? Is there an issue already for this in the app repo?

This issue is related:
https://github.com/digitalfabrik/integreat-app/issues/4220

integreat_cms/cms/templates/pages/page_pdf.html
bahaaTuffaha: The tailwind blocker is gone, but _raw.html hardcodes three things the PDF needs to set and none of them inside a {% block %}:

1- lang : it uses the UI language, the PDF needs the content language, because the CSS picks the page number digits from it (٢ for Arabic and Farsi).
2- dir : not set at all. the PDF needs `dir="rtl"` for right-to-left text.
3- `<body class="font-default">`  the PDF sets its own font class per language.

bahaaTuffaha:
> * I think the Dockerfile for the .devcontainer needs to also be adapted to install the new system libraries

Done.

.circleci/config.yml
bahaaTuffaha: Removed the rest 👍

andrew8er:
> 1.) I have two legit (change) requests in regards to the new system libraries:
>
>    we will need a PR on the salt repo that installs these new libraries there on all the servers concerned - check states/integreat-cms/init.sls in the salt repo (this should be merged before we merge this PR into develop otherwise the Testsystem server will crash)

I added a PR here: https://git.tuerantuer.org/DF/salt/pulls/940

bahaaTuffaha:
> > 1.) I have two legit (change) requests in regards to the new system libraries:
> > we will need a PR on the salt repo that installs these new libraries there on all the servers concerned - check states/integreat-cms/init.sls in the salt repo (this should be merged before we merge this PR into develop otherwise the Testsystem server will crash)
> 
> I added a PR here: https://git.tuerantuer.org/DF/salt/pulls/940

Thanks... I created an issue for it earlier:
https://tasks.tuerantuer.org/projects/infra/work_packages/5956/activity?query_id=59

andrew8er:
> Thanks... I created an issue for it earlier:
> https://tasks.tuerantuer.org/projects/infra/work_packages/5956/activity?query_id=59

Ah sorry, I didn't see this. I'll link the issue in the PR.

MizukiTemma:
The PR in salt-repo was merged, not blocked anymore.

MizukiTemma (approved):
Thank you so much for the research and implementation 😻 This PR seems to fix the problems like #2020, #2021 and I didn't find any regression 👍Checked for some RTL languages (Farsi, Dari, Urdu) and also form LTR languages (German, Ukrainian, Amharic, Chinese)

PR #4547, opened 2026-09-04
3059: Rename poi to places

### LINKED ISSUE #3059: Rename pois / locations to "place"
### Describe the Bug

We have a model called POI, which is sometimes referred to using `poi` and sometimes referred to using `location` (For example, the event model stores `location`, but region uses `pois`). 
This can be confusing and is inconsistent. Instead, we should decide for one name and stick to it.

As far as I know, in the codebase we mostly use `poi`, while we mostly use `location` in the ui. However, there are exceptions for both.

**Updates**
Rename "location" and "poi" to "place".

bahaaTuffaha:
### Short description
We have a model called POI, which is sometimes referred to using poi and sometimes referred to using location (For example, the event model stores location, but region uses pois).
This can be confusing and is inconsistent. Instead, we should decide for one name and stick to it.

### Proposed changes

- Renamed `poi` / `location` to `place` across models, forms, views, templates, JS, tests, docs, roles, and fixtures (code and CMS UI; public webapp/API paths kept).
- Renamed the folders `cms/models/pois`, `cms/models/poi_categories`, `cms/forms/pois`, `cms/forms/poi_categories`, `cms/views/pois`, `cms/views/poi_categories`, `cms/templates/pois`, `cms/templates/poicategories`, `cms/templates/ajax_poi_form`, `static/src/js/pois`, `static/src/js/poi-categories`, `tests/cms/models/pois`, `tests/cms/views/poi` and `tests/cms/views/poi_categories` to their `place` equivalents.
- Renamed `POI`, `POITranslation`, `POICategory`, and `POIFeedback` to `Place`, `PlaceTranslation`, `PlaceCategory`, and `PlaceFeedback`.
- Added `0164_rename_poi_to_place` which drops the `enforce_slug_uniqueness` trigger and the four constraints referring to the old field names applies the `RenameModel`/`RenameField` operations, updates the verbose names and help texts, then re-adds the trigger and constraints under their new names.
- Appended an `update_roles` `RunPython` to that migration, it emits the post-migrate signal so the permissions of the renamed models exist, deletes the nine obsolete `(change|delete|view)_poi*` permissions then re-assigns each role from `roles.PERMISSIONS`.
- Updated `docs/src/management-commands.rst`, `api-docs.rst`, `frontend-bundling.rst` and `conf.py`.
- Updated the translation files.
- The following deliberately left unchanged:
  - JSON keys `"pois"` (`api/v3/regions.py`) and `"location"` / `"location_path"` (`api/v3/places.py`, `events.py`)
  - API paths `/locations/`, `/location-categories/`, `/feedback/poi/`
  - `svg/poi-category-icons/` it's path is emitted as `icon_url` in the API response.
  - PlaceTranslation.url_infix still returns "locations" (place_translation.py).
  - `region.locations_enabled` column (CMS label is `"activate places"`; writable region-settings JSON still uses `"locations_enabled"` via `WRITABLE_FIELDS`)

### Side effects

-  I'm not sure if the migration could be handled on our huge dataset without crashes or out of memory problems.

### How to test

- Check Place Categories.
- Check the Places `augsburg/places/de/` and all of it's buttons like archived places, copy, archive, restore, delete tool tips.
- Check create / edit place.
- Check Events create/edit (Venue section).
- Check Contacts create/edit (Connect a place).
- Check Region settings.
- Check Edit organization page.
- Test public APIs that they work without issues (paths should stay the same).

### Resolved issues

Fixes: #3059


integreat_cms/core/management/commands/copy_places.py
```diff
@@ -27,7 +27,7 @@ def make_hardlink(file_field: FileField, region_id: int) -> FileField:
     """
     Create a hardlink to the original file in the media path for the target region
     """
-    base = file_field.storage.base_location
+    base = file_field.storage.base_place
```
hannaseithe:
```suggestion
    base = file_field.storage.base_location
```
This is a cached_property on django's `FileSystemStorage`. If you execute copy_places on a template region that contains a place with an icon set from that region, you will run into an error otherwise.

hannaseithe (changes requested):
Thank you very much for this big rename work. I have checked the cms is still working and your intentional unchanged exclusions (all related to the public API) also make sense. Claude found one bug, see below.

integreat_cms/core/management/commands/copy_places.py
bahaaTuffaha: Nice catch! 👀

hannaseithe (approved):
I approve since my one change request has been applied.

integreat_cms/cms/forms/contacts/contact_form.py
```diff
@@ -45,9 +45,9 @@ class ContactForm(CustomModelForm):
         label=_("Mobile Phone number"),
     )
 
-    use_location_opening_hours = forms.BooleanField(
+    use_place_opening_hours = forms.BooleanField(
         required=False,
-        label=_("Adopt opening hours from linked location as office hours"),
+        label=_("Adopt opening hours from linked place as office hours"),
```
andrew8er: ❌ "Adopt opening hours from _the_ linked place as office hours"

integreat_cms/cms/migrations/0164_rename_poi_to_place.py
```diff
@@ -0,0 +1,411 @@
...
...       -- Look up the region for the new/updated place\n                    SELECT region_id INTO new_region_id\n                    FROM cms_place\n                    WHERE id = NEW.place_id;\n\n                    -- Set advisory lock (Postgresql specific)\n                    PERFORM pg_advisory_xact_lock(hashtextextended(NEW.language_id || ':' || new_region_id || ':' || NEW.slug, 0));\n\n                    -- Check if there's a conflict (same slug/language/region but different place)\n                    IF EXISTS (\n                        SELECT 1\n                        FROM cms_placetranslation t\n                        JOIN cms_place p ON t.place_id = p.id\n                        WHERE t.slug = NEW.slug\n                        AND t.language_id = NEW.language_id\n                        AND p.region_id = new_region_id\n                        AND t.place_id <> NEW.place_id\n                    ) THEN\n                        RAISE EXCEPTION 'Slug must be unique per language and region across different places.' USING ERRCODE = 'unique_violation'; -- SQLSTATE 23505\n                    END IF;\n\n                    RETURN NEW;\n                END;\n                ",
```
andrew8er: 🔧 Could use a multiline string here, using [`dedent()`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent).

integreat_cms/cms/models/regions/region.py
```diff
@@ -159,8 +159,8 @@ class Region(AbstractBaseModel):
     )
     locations_enabled = models.BooleanField(
```
andrew8er: 💭  Should `locations_enabled` also be changed?

integreat_cms/cms/templates/contacts/contact_form.html
```diff
@@ -141,7 +141,7 @@ <h3 class="heading font-bold text-black">
                                     data-confirmation-title="{{ restore_dialog_title }}"
                                     data-confirmation-text="{{ restore_dialog_text }}"
                                     data-confirmation-subject="{{ contact_form.instance.name }}"
-                                    {% if contact_form.instance.location.archived %} data-confirmation-referenced-location-title="{{ referenced_location_title }}" data-confirmation-referenced-location-subject="{{ contact_form.instance.location }}" {% endif %}
+                                    {% if contact_form.instance.place.archived %} data-confirmation-referenced-place-title="{{ referenced_location_title }}" data-confirmation-referenced-place-subject="{{ contact_form.instance.place }}" {% endif %}
```
andrew8er: ❌ Should probably read `referenced_place_title`.

andrew8er (changes requested):
What a huge PR, thanks for the effort.

There's one bug in a form, otherwise stylistic suggestions. The directory `/integreat_cms/static/src/svg/poi-category-icons` could also be renamed.

integreat_cms/cms/models/regions/region.py
bahaaTuffaha: I think yes after I checked the app again...

andrew8er (approved):

PR #4556, opened 2026-09-09
Update Python dependencies, fix tests

andrew8er:
### Short description

Updates some Python packages to fix some critical security reports, fixes tests that originated from updated packages and also fixes some subtle errors in the way we run tests on Docker. Some updates where held back to not conflict with #4541.

### Proposed changes
- Updates some Python packages to fix some critical security reports.
- Fix djlint settings. We had quiet a few config options that applied to pylint, but were somehow copied to the djlint config.
- Remove a special djlint exemption, since the associated issue was resolved.
- Fix test fixtures.
- Let prettier ignore `/tests/api/expected-outputs`, since these need be formatted exactly as they are.
- Fixes a few subtle bugs that originate in the way we run tests on Docker. I have constant problems with this setup and I'd dare to say that it is currently absolutely overcomplicated and brittle beyond repair (e.g. caching the pip installed `.venv` directory is a total minefield). The following sections contains discovered problems and their fixes. I have some more errors, that are not fixed in this PR.

#### Problem with running tests in Docker 1
Error when running `tools/test.sh`:
```
[Errno 2] No such file or directory: '/home/circleci/venv/bin/python'
```
##### Analysis by Claude
- The cached venv's `bin/python` symlink points at `/home/circleci/.pyenv/versions/3.13.11/bin/python`.
- The recent commit 34d73b3f7 ("Track floating cimg/python:3.13 tag in Dockerfile.test") let the base image's Python float, and the rebuilt image now ships pyenv 3.13.15. So that symlink is now dangling.
- `test_entrypoint.sh` did detect the broken interpreter (`[[ ! -x ]]` is true for a dangling symlink) and ran `python -m venv "$VENV_DIR"` — but without `--clear`, venv leaves existing symlinks in place, so `bin/python` stayed dangling and every pip/pytest invocation referencing it failed with Errno 2. I reproduced this exactly in the cached volume.

##### Fix
`tools/docker/test_entrypoint.sh:29` — the recreate branch now also runs when the interpreter can't execute, and rebuilds with `--clear`

#### Problem with running tests in Docker 2

Running tests report:
```
E           webpack_loader.exceptions.WebpackError:
E                       Error in
E                       [tsl] ERROR
E                 TS18003: No inputs were found in config file 'tsconfig.json'. Specified 'include' paths were '["**/*"]' and 'exclude' paths were '[]'.
```

##### Analysis by Claude
Important finding: the build succeeds even without the include (committed state, status done). So the include addition isn't the true fix. Let me restore and dig into what actually produces the error.

Only two tsconfigs, both now have include. The TS18003 fallback to `["**/*"]` only occurs when include is absent — so the user's additions are the correct hardening. The remaining gap is that a once-failed build leaves an error status in `webpack-stats.json`, and `ensure_webpack_bundle_exists` only rebuilds when static/dist is empty — so it silently reuses the broken bundle on every later run. Let me verify current state and that the build/tools still work.

##### Fix
1. The real fix — explicit include.
 TS18003 reporting include paths of `["**/*"]` only happens when a `tsconfig.json` has no explicit include, so TypeScript falls back to the `**/*` glob and finds no inputs in that config's context. Your uncommitted additions of include to both `tsconfig.json` and `tools/npm/tsconfig.json` are the correct fix. Verified: `npm run build`/`npm run prod` produce a valid bundle (status: done) and `tsc -p tools/npm/tsconfig.json` passes cleanly.
2. Why it kept failing every run — a stale broken bundle got reused.

`ensure_webpack_bundle_exists` (`tools/_functions.sh:367`) only rebuilt when static/dist was empty. Once a build failed and wrote an error-status webpack-stats.json, django-webpack-loader re-raised that recorded error on every subsequent test run, even after you fixed the tsconfig. I hardened the helper to also rebuild when the stats file reports "status": "error", so a one-off failure can no longer poison later runs. Shellcheck passes.

### Side effects
None

### Resolved issues


andrew8er:
~~I think there is a real problem with our tests. I cannot pin it down exactly, but here are some observations:~~

~~- On CI, tests fail more or less randomly. These are all commits on main:~~
~~- cb879b6af4674e7d3ce36031636b3a7857282f76 fails~~
~~- 6c0e2166b43339f5661e182606b7d0cea3637734 passes~~
~~- 474f02e8567dcb086183c45a2cf95ea50959bce1 fails~~
 
~~- Locally, running with~~
~~- `tools/test.sh` fails, but its different failing tests with every invocation.~~
~~- `tools/test.sh <specific_test_file.py>` always works.~~
~~- Running all tests through my (yet unreleased) tooling always works.~~

~~@digitalfabrik/team-cms can you run the tests locally and see what you get? I'm at loss.~~


There were multiple errors, partially fixed by #4562 and some are fixed by this PR.

hannaseithe:
> I think there is a real problem with our tests. I cannot pin it down exactly, but here are some observations:
> 
> * On CI, tests fail more or less randomly. These are all commits on main:
>   
>   * [cb879b6](https://github.com/digitalfabrik/integreat-cms/commit/cb879b6af4674e7d3ce36031636b3a7857282f76) fails
>   * [6c0e216](https://github.com/digitalfabrik/integreat-cms/commit/6c0e2166b43339f5661e182606b7d0cea3637734) passes
>   * [474f02e](https://github.com/digitalfabrik/integreat-cms/commit/474f02e8567dcb086183c45a2cf95ea50959bce1) fails
> * Locally, running with
>   
>   * `tools/test.sh` fails, but its different failing tests with every invocation.
>   * `tools/test.sh <specific_test_file.py>` always works.
>   * Running all tests through my (yet unreleased) tooling always works.
> 
> @digitalfabrik/team-cms can you run the tests locally and see what you get? I'm at loss.

I think I have found what is going on here: The original failing test `test_api_result` in `test_api_raw_content.py` is actually flaky (the flakiness is caused by different ordering and distribution to CI test workers between runs). It relies on the default language being set to German, but there are other tests (if run before that test cause it to fail), that falsely mutate `settings.LANGUAGE_CODE`  (instead of using the settings fixture) which is not reset between tests. I will create a PR that fixes this issue hopefully.

andrew8er:
@bahaaTuffaha This PR will conflict with your PR #4541. I can refactor it, so it will not cause any conflicts, if your PR will be merged soon?

bahaaTuffaha:
> @bahaaTuffaha This PR will conflict with your PR #4541. I can refactor it, so it will not cause any conflicts, if your PR will be merged soon?

Thank you!

MizukiTemma (approved):
Found no issue. No objection.

Thank you 😸

PR #4559, opened 2026-09-09
Allow releasing a region from API management

### LINKED ISSUE #4558: Release a region from being managed via the API
### User Story

**As a** member of the service or CMS team,
**I want to** take a region back out of external management,
**so that** I can maintain its booked features and translation budget in the CMS again when the external system no longer manages it.

### Problem Description

Since #4467 a region counts as managed by an external system as soon as `api_settings_synced_at` is set (`Region.is_api_managed`), and the synced settings are read-only in the region form. That is correct, but there is no way back.

After the first `POST` to `/api/v3/<slug>/settings/`, `events_enabled`, `locations_enabled`, `contacts_enabled`, `external_news_enabled`, `integreat_chat_enabled`, `push_notifications_enabled` and the budget fields stay locked permanently. Neither the form nor the API can reset the timestamp — the allowlist in `cms/constants/region_api_settings.py` deliberately does not contain it — so the only route is a direct database write.

This matters when the external management of a region ends: an expired contract, a switch of the managing system, or a link created by mistake while testing. We ran into exactly the last case on `cms-test`, where a region is now locked although the link only existed for a test.

### Acceptance Criteria

- [ ] A user with `cms.change_region` can release a region from API management in the region form
- [ ] The release only lifts the lock; no setting is changed by it
- [ ] After the release, all fields in `region_api_settings.WRITABLE_FIELDS` are editable again
- [ ] Releasing a region that is not API-managed is a no-op with an informational message, not a success
- [ ] The release is not permanent: a subsequent push by the external system marks the region as API-managed again
- [ ] Users without `cms.change_region` cannot release a region

### Error & Edge Case Handling

- **Unknown slug** → 404, not a server error
- **`GET` on the endpoint** → not allowed; the release must be a `POST`
- **Region not API-managed** → informational message, nothing changed
- **A budget outside the package sizes** (the API may push arbitrary word counts) must survive the release. The region form offers a dropdown of package sizes for CMS-maintained regions, so a value with no matching option must not be silently replaced on the next save.
- **Two users releasing concurrently** → the second gets the informational message

### User Flow

1. The region is API-managed; the region form shows the yellow banner explaining that the settings are managed externally and read-only
2. The user clicks the release button **inside that banner**, so the way out is where the problem is stated
3. A confirmation dialog names the consequence: the external system is no longer the source, and a later push overwrites the values maintained here and locks them again
4. On confirmation the user returns to the region form, the banner is gone and the fields are editable
5. Messages state what happened and that a later push will take over again

### Technical Notes

- The action only needs to set `api_settings_synced_at = None`; `is_api_managed` is derived from it
- Modelled on `delete_region` in `cms/views/regions/region_actions.py`: `@require_POST`, `@permission_required("cms.change_region")`, redirect back to `edit_region`
- No region membership check is needed: the URL lives under `/regions/<slug>/`, so `request.region` is `None` and `AccessControlMiddleware` already restricts the area to staff
- The release being overridable by the next push is intended. A release the API could not override would be the more dangerous design, because both systems would then disagree without anyone noticing

### Out of Scope

- Preventing the external system from taking the region over again
- Any change to the API endpoint or its allowlist
- The CRM side: it needs its own counterpart so it stops pushing (digitalfabrik/customcrm#95)

### Testing Guidance

- Push settings for a test region from the CRM, then release it in the CMS and check that the fields are editable and the values unchanged
- Use a budget that is **not** one of the package sizes, then save the form once and verify the budget is still what the API pushed
- Push again from the CRM and verify the region is locked once more

### Additional Context

The CRM side has a counterpart ("Verbindung lösen", digitalfabrik/customcrm#95) which stops further pushes but cannot lift the lock here — its message says so explicitly. Both sides together make a complete way back.

### Related Issues

- Part of #4138
- Introduced the lock: #4467
- CRM counterpart: digitalfabrik/customcrm#95

dkehne:
### Short description

Adds a way to release a region from being managed via the API, so its booked features and translation budget can be maintained in the CMS again.

### Proposed changes

- New action `release_api_management` in `cms/views/regions/region_actions.py`, modelled on `delete_region`: `@require_POST`, `@permission_required("cms.change_region")`, redirecting back to the region form. It sets `api_settings_synced_at = None` and changes nothing else.
- A release button **inside the yellow banner** that explains the lock, so the way out sits where the problem is stated. It uses the existing `confirmation-button` idiom, with `type="button"` so it cannot submit the region form if the confirmation JS has not loaded yet.
- The confirmation text and the messages state the full consequence: the external system is no longer the source, and a later push overwrites the values maintained here in the meantime and locks them again.
- `region_form.py` now uses the package-size dropdown for `mt_budget_booked` **only when the current value is one of the choices**. A budget the API pushed is rarely one of the six sizes, and a value with no matching option made the browser select the first one, so the next save of the region form — even one only toggling an unrelated switch — silently wrote `50000`. The reasoning was already in the code for the API-managed branch; the release created exactly the state that comment warns about.
- Release note `4558.yml`, and German translations for all new strings.

### Side effects

- **`region_form.py` affects every region, not only released ones.** A CMS-maintained region whose `mt_budget_booked` is not one of the package sizes now gets a plain integer input instead of the dropdown. Before this PR such a region could only come from an API push; the model has carried no `choices` since #4467, so the value is valid either way.
- **`api_settings_synced_at` is the only state the release touches.** The settings keep the values the external system pushed last — the release states who maintains them from now on, not what they should be.
- **`last_updated`** is not advanced, because `update_fields` excludes the `auto_now` field. This matches what the API endpoint does and is left as is.
- **The unsaved-warning interaction** applies to this button as it does to the delete button: with unsaved changes on the page, confirming triggers the native `beforeunload` dialog. Handled separately in #4560 / #4561, not here.

### How to test

1. Pick a test region and push settings to it via `POST /api/v3/<slug>/settings/` with a personal API token of a user holding `cms.change_region`. Use a budget that is **not** one of the package sizes, for example `{"mt_budget_booked": 77777}`.
2. Open the region form. The yellow banner appears, the synced fields are read-only, and the budget shows 77777.
3. Click "Edit these settings here again" and confirm.
4. The banner is gone, the fields are editable, and every value is unchanged — including the budget.
5. Save the form once without touching the budget, then reload: the budget must still be 77777. (Before this PR it became 50000 here.)
6. Push again via the API and verify the region is locked once more.

Also worth checking as a non-privileged user: without `cms.change_region` the button is not rendered and a direct `POST` returns 403.

### Resolved issues

Fixes: #4558

---

**Testing:** full suite through `./tools/test.sh` (the CI-mirroring container): 4822 passed, 2 skipped, 1 xfailed. `ruff check`, `ruff format`, `djlint` and `mypy` clean; no empty or fuzzy translation entries.

**Note on the translation file:** it contains one hunk that is not mine — a rewrap of an existing `text-blue-600` string. `makemessages` produces it in the CI image, and the workflow passed with it (`check-translations` included), so it is what CI expects rather than drift against it. Resetting the file to `develop` and regenerating reproduces the rewrap.

### DIFF
--- a/integreat_cms/cms/forms/regions/region_form.py
+++ b/integreat_cms/cms/forms/regions/region_form.py
@@ -236,10 +236,16 @@ class RegionForm(CustomModelForm):
             # silently display the wrong one.
             for field_name in region_api_settings.WRITABLE_FIELDS:
                 self.fields[field_name].disabled = True
-        else:
+        elif self.instance.mt_budget_booked in dict(machine_translation_budget.CHOICES):
             # The model no longer restricts the budget to the predefined package sizes, because the
             # API may push arbitrary values. Regions maintained in the CMS should still get the
             # dropdown with the known sizes, so the choices are defined on the form instead.
+            #
+            # A budget outside those sizes keeps the plain integer field, for the same reason the
+            # API-managed branch above does: a dropdown without a matching option displays the
+            # wrong one, and the next save of this form would silently write that wrong value.
+            # This is not hypothetical -- it is exactly the state of a region that was released
+            # from API management while the API had pushed an arbitrary word count.
             self.fields["mt_budget_booked"] = forms.TypedChoiceField(
                 choices=machine_translation_budget.CHOICES,
                 coerce=int,
--- a/integreat_cms/cms/views/regions/region_actions.py
+++ b/integreat_cms/cms/views/regions/region_actions.py
@@ -158,7 +158,8 @@ def release_api_management(
     messages.info(
         request,
         _(
-            "If the connected system pushes settings again, they will become read-only once more.",
+            "If the connected system pushes settings again, it overwrites the values maintained "
+            "here and they become read-only once more.",
         ),
     )
 
--- a/integreat_cms/cms/views/regions/region_actions.py
+++ b/integreat_cms/cms/views/regions/region_actions.py
@@ -109,6 +109,62 @@ def delete_region(
     return redirect("regions")
 
 
+@require_POST
+@permission_required("cms.change_region")
+def release_api_management(
+    request: HttpRequest,
+    *args: Any,
+    **kwargs: Any,
+) -> HttpResponseRedirect:
+    r"""
+    This view releases a region from being managed by an external system via the API.
+
+    It only clears :attr:`~integreat_cms.cms.models.regions.region.Region.api_settings_synced_at`,
+    which is what marks a region as API-managed and makes the synced settings read-only in the
+    region form. Nothing else is changed: the settings themselves keep the values the external
+    system pushed last.
+
+    The release is deliberately not permanent. If the connected system pushes again, the region is
+    marked as API-managed once more and the fields become read-only again -- which is the desired
+    semantics: the release states "we maintain these settings here from now on", and a subsequent
+    push visibly contradicts that instead of failing silently.
+
+    :param request: The current request
+    :param \*args: The supplied arguments
+    :param \**kwargs: The supplied keyword arguments
+    :return: A redirection to the region form
+    """
+    region = get_object_or_404(Region, slug=kwargs.get("slug"))
+
+    if not region.is_api_managed:
+        messages.info(
+            request,
+            _('The settings of the region "{}" are not managed via the API.').format(
+                region.name,
+            ),
+        )
+        return redirect("edit_region", slug=region.slug)
+
+    region.api_settings_synced_at = None
+    region.save(update_fields=["api_settings_synced_at"])
+    logger.info("%r released %r from API management", request.user, region)
+
+    messages.success(
+        request,
+        _('The settings of the region "{}" can be edited here again.').format(
+            region.name,
+        ),
+    )
+    messages.info(
+        request,
+        _(
+            "If the connected system pushes settings again, they will become read-only once more.",
+        ),
+    )
+
+    return redirect("edit_region", slug=region.slug)
+
+
 @shared_task
 def async_delete_region(region_id: int) -> None:
     region = Region.objects.get(id=region_id)
--- a/integreat_cms/cms/templates/regions/region_form.html
+++ b/integreat_cms/cms/templates/regions/region_form.html
@@ -36,10 +36,11 @@
                     {% endwith %}
                 </p>
                 {% if perms.cms.change_region %}
-                    <button title="{% translate "Edit these settings here again" %}"
+                    <button type="button"
+                            title="{% translate "Edit these settings here again" %}"
                             class="btn btn-blue mt-2 confirmation-button"
                             data-confirmation-title="{% translate "Please confirm that you want to maintain these settings here again." %}"
-                            data-confirmation-text="{% translate "The connected system will no longer be the source for the booked features and the translation budget." %} {% translate "If it pushes settings again, they will become read-only here once more." %}"
+                            data-confirmation-text="{% translate "The connected system will no longer be the source for the booked features and the translation budget." %} {% translate "If it pushes settings again, it overwrites the values maintained here and they become read-only once more." %}"
                             data-confirmation-subject="{{ form.name.value }}"
                             data-action="{% url 'release_api_management' slug=form.instance.slug %}">
                         <i icon-name="unlock"></i>
--- a/integreat_cms/cms/templates/regions/region_form.html
+++ b/integreat_cms/cms/templates/regions/region_form.html
@@ -35,6 +35,17 @@
                         {% endblocktranslate %}
                     {% endwith %}
                 </p>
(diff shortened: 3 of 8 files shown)


dkehne:
### Code review

Nine findings, six addressed in 55addf029. Two are pre-existing behaviour shared with the delete button, and one turned out to be a false alarm — with evidence.

**1. Silent data loss, in exactly the case this feature exists for.** Reproduced before fixing:

```
field=TypedChoiceField value=77777
not in choices=[50000, 100000, 150000, 250000, 1000000, 5000000]
```

An API-managed region keeps the plain integer field for `mt_budget_booked`; a CMS-maintained one gets the package-size dropdown. Releasing moved a region from the first state into the second **without touching the value** — deliberately. But a budget the API pushed is rarely one of the six sizes, and a value with no matching option makes the browser select the first. The next save of the region form, even one that only toggles an unrelated switch, then wrote `50000`. No error, no hint, and the region's actual translation quota shrank.

The reasoning was already in the code: the API-managed branch keeps the integer field because „a dropdown without a matching option would silently display the wrong one". My release created precisely the state that comment warns about. Fixed at the cause — the dropdown is now used only when the current value is one of the choices, which has always been true for regions maintained in the CMS.

**4. The missing test was the one that mattered.** `test_release_keeps_the_synced_settings` stopped one step before the loss. Added: that the settings are really editable again after the release — the point of the whole action, asserted on the form rather than inferred from the absent banner — and it is the test that would have caught finding 1.

**3. The banner block was rendered by no test.** `view_config.py` only calls `edit_region` for `augsburg`, which is not API-managed, and the form test never touches the template. A broken `{% url %}` in the block would have meant HTTP 500 for exactly the regions this feature targets. Added a render test, plus 404 for an unknown slug.

**6. `type="button"` added.** Without it the HTML default is submit and only `showConfirmationPopup`'s `preventDefault` stood in the way — a click before the `load` event would have saved the region form instead. Together with finding 1 that was the save which reset the budget. (The delete button has the same shape; I did not want to propagate it.)

**8. The message named only half the consequence.** The next push does not just make the settings read-only again, it also overwrites whatever was maintained here in the meantime. As long as the CRM side is still linked, that is the normal case, not the exception.

**5.** Dropped the assertion that no code path could break.

### Finding 2 — false alarm, and worth recording why

The concern was that the foreign `.po` rewrap would make `check_translations` red. Two pieces of evidence against it:

- The `check-translations` job runs in `cimg/python:3.13.11` with apt `gettext` (`.circleci/config.yml:99-111`) — the same base image and package as `tools/docker/Dockerfile.test`. The gettext version matches by construction.
- The full workflow on `6c5662ca9`, which contained the rewrap, passed **including `check-translations`**.

Cross-checked the other way as well: resetting the file to `develop` and re-running `makemessages` in that container reproduces the rewrap. So it is not drift against CI — it is what CI produces. Removing it would have made the job fail.

### Not changed

Findings 7 (`data-unsaved-warning` interaction) and 9 (`last_updated` not advanced) are pre-existing behaviour, identical for the delete button and for the API endpoint respectively. Both are worth their own issue rather than a change smuggled in here.

Thanks for the review — finding 1 in particular was a real bug in the scenario the feature was built for, and it is the kind that would have surfaced as an unexplained budget drop months later.

Full suite through `./tools/test.sh`: **4822 passed**. `ruff check`, `ruff format`, `djlint` and `mypy` clean, no empty or fuzzy translation entries.

integreat_cms/cms/forms/regions/region_form.py
```diff
@@ -236,10 +236,16 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
             # silently display the wrong one.
             for field_name in region_api_settings.WRITABLE_FIELDS:
                 self.fields[field_name].disabled = True
-        else:
+        elif self.instance.mt_budget_booked in dict(machine_translation_budget.CHOICES):
```
seluianova, line 239: > Save the form once without touching the budget, then reload: the budget must still be 77777. (Before this PR it became 50000 here.)

the side effect here is that it is also possible to set any value in the form afterwards. for example, I can change `77777` to `66666` in the UI.

is this intended?

the choice becomes limited again only when I change it to `50000` (or any other value from `machine_translation_budget.CHOICES`).

integreat_cms/cms/views/regions/region_actions.py
```diff
@@ -109,6 +109,63 @@ def delete_region(
     return redirect("regions")
 
 
+@require_POST
+@permission_required("cms.change_region")
+def release_api_management(
+    request: HttpRequest,
+    *args: Any,
+    **kwargs: Any,
```
andrew8er, line 117: 💭 Instead of `kwargs`, we could probably use `slug: str` here.

andrew8er (approved):
Looks good.

PR #4561, opened 2026-09-09
Do not let the unsaved-warning cancel a confirmed action

### LINKED ISSUE #4560: Unsaved-warning silently cancels a confirmed action
### Bug Summary

`unsaved-warning.ts` clears its dirty flag only on `submit` of the form carrying `data-unsaved-warning`. The confirmation dialog is a separate form, so a confirmed action navigates away with the flag still set and the browser asks a second time. Choosing "Stay" then cancels the action without any message.

This affects every `confirmation-button`, for example deleting a region in the region form.

### Expected Behavior

After confirming an action in the confirmation dialog, the action is carried out. The dialog already names the title, the text and the subject, so no further question is needed and the browser should not ask again.

### Actual Behavior

The native `beforeunload` dialog appears ("This content is not saved. Would you leave the page?"). Both outcomes are wrong:

- **"Stay"** cancels the navigation, so **nothing happens at all** — no message, no hint — although the action was just explicitly confirmed. This is the actual defect.
- **"Leave"** carries out the action but discards the typed edits. Correct in itself, but presented as a warning that suggests the user is doing something wrong.

### Steps to Reproduce

1. Open the region form of any region (`/regions/<slug>/edit/`) as a user with `cms.change_region`
2. Change any field, for example the region name — do **not** save
3. Click "Delete this region" (or, on an API-managed region, "Edit these settings here again")
4. Confirm in the confirmation dialog
5. The browser asks "This content is not saved. Would you leave the page?"
6. Choose "Stay"
7. Observe that nothing happened: the region still exists, no message is shown

### Reproducibility

Always (100%)

### Environment

- **Environment:** independent of environment, this is client-side behaviour
- **App version / commit:** `develop` at 474f02e85
- **Browser:** any browser implementing `beforeunload` (Chrome, Firefox, Safari)
- **User role / permissions:** any role that sees a `confirmation-button`, e.g. SERVICE_TEAM or CMS_TEAM in the region form

### Severity

**Workaround:** save or reload the page before triggering a confirmation button, so the dirty flag is not set.

### Regression Information

- **Is this a regression?** No — the behaviour has existed as long as `unsaved-warning.ts` and the generic confirmation dialog have coexisted.

#### Affected Components

- `integreat_cms/static/src/js/unsaved-warning.ts`
- `integreat_cms/cms/templates/generic_confirmation_dialog.html`
- every template using `confirmation-button`, e.g. `cms/templates/regions/region_form.html`

#### Suspected Root Cause

`unsaved-warning.ts` registers its listeners on the form with `data-unsaved-warning` and clears the flag only there:

```ts
form?.addEventListener("submit", () => {
    dirty = false;
});
```

The dialog submits its own form (`#confirmation-dialog form`), which never reaches that listener. There is already an `autosave` listener clearing the flag for the same class of reason.

#### Proposed Fix

```ts
document.querySelector("#confirmation-dialog form")?.addEventListener("submit", () => {
    dirty = false;
});
```

- [ ] Confirming an action in the confirmation dialog carries it out, even with unsaved changes on the page
- [ ] No `beforeunload` dialog appears after a confirmation
- [ ] The warning still appears when navigating away without saving and without a confirmation
- [ ] Autosave keeps clearing the flag as before

### Verification / Testing

- [ ] Reproduce the original bug → confirm it no longer occurs
- [ ] Edit a field, then navigate away without confirming → warning still appears
- [ ] Delete a region with unsaved changes → deletion happens on first confirmation
- [ ] vitest covers the sequence (edit → confirm → unload)

### Related Issues

- Found while reviewing #4559
- CRM counterpart of that work: digitalfabrik/customcrm#95

dkehne:
### Short description

Clears the unsaved-warning flag when the confirmation dialog is submitted, so a confirmed action is no longer cancelled silently by the browser's `beforeunload` prompt.

### Proposed changes

- `unsaved-warning.ts` now also clears `dirty` on `submit` of `#confirmation-dialog form`, next to the existing `autosave` listener that exists for the same class of reason.
- A vitest test drives the actual sequence in jsdom — edit a field, confirm the dialog, unload — and asserts the warning appears after the edit and is gone after the confirmation.
- Release note `4560.yml`.

### Side effects

- **Affects every `confirmation-button`, not just one page.** Deleting a region, deleting a page, discarding an offer embedding: after this PR none of them is interrupted by a second browser prompt. That is the intent, and it applies repo-wide.
- **The warning itself is unchanged** for the ordinary case: editing a field and navigating away without saving and without a confirmation still warns.
- **Typed edits are still discarded** when a confirmed action navigates away — that has always been the case. The difference is that it now happens without a prompt that could cancel the confirmed action.
- **No new dependency.** jsdom is already present, and the module needed no refactoring to be testable.

There are no intended deviations from the issue and design. The fix is the one proposed in #4560.

### How to test

1. Open the region form of any region as a user with `cms.change_region`
2. Change any field, for example the region name — do **not** save
3. Click "Delete this region" and confirm in the dialog
4. The deletion happens on the first confirmation; no `beforeunload` prompt appears

Regression check, so the warning is not lost altogether:

5. Change a field again and navigate away via the menu without confirming anything
6. The warning must still appear

### Resolved issues

Fixes: #4560

---

**Testing:** all 3 vitest files pass, `eslint` and `prettier` clean, webpack bundle builds. Counter-check: without the new listener the test fails with `expected true to be false`.

### DIFF
--- a/integreat_cms/release_notes/current/unreleased/4560.yml
+++ b/integreat_cms/release_notes/current/unreleased/4560.yml
@@ -1,0 +1,2 @@
+en: Fix a confirmed action being cancelled silently when the page had unsaved changes
+de: Behebe, dass eine bestätigte Aktion stillschweigend abgebrochen wurde, wenn die Seite ungespeicherte Änderungen hatte
--- a/integreat_cms/static/src/js/__tests__/unsaved-warning.test.ts
+++ b/integreat_cms/static/src/js/__tests__/unsaved-warning.test.ts
@@ -1,0 +1,38 @@
+// @vitest-environment jsdom
+import { expect, test } from "vitest";
+
+/**
+ * Dispatch a cancelable beforeunload event and report whether the warning kicked in.
+ */
+const warningShown = (): boolean => {
+    const event = new Event("beforeunload", { cancelable: true });
+    window.dispatchEvent(event);
+    return event.defaultPrevented;
+};
+
+test("a confirmed action is not interrupted by the unsaved-warning", async () => {
+    document.body.innerHTML = `
+        <form data-unsaved-warning>
+            <input id="field" type="text" />
+        </form>
+        <div id="confirmation-dialog">
+            <form id="confirmation-form" method="post"></form>
+        </div>
+    `;
+
+    // Registers the beforeunload handler and, on load, the form listeners
+    await import("../unsaved-warning");
+    window.dispatchEvent(new Event("load"));
+
+    expect(warningShown()).toBe(false);
+
+    // The user edits something
+    document.getElementById("field").dispatchEvent(new Event("input", { bubbles: true }));
+    expect(warningShown()).toBe(true);
+
+    // The user triggers a confirmation button and confirms. The dialog is a separate form, so
+    // without clearing the flag the browser would ask a second time -- and staying on the page
+    // would silently swallow the action the user just confirmed.
+    document.getElementById("confirmation-form").dispatchEvent(new Event("submit", { bubbles: true }));
+    expect(warningShown()).toBe(false);
+});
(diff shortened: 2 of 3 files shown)


dkehne:
Sorry, this is way to theortical. missinterpreted sth...

PR #4562, opened 2026-09-10
Fix flaky test: `test_api_raw_content.py::test_api_result`

hannaseithe:
### Short description

The test ` test_api_raw_content.py::test_api_result` flakes on the CI (for example here: https://app.circleci.com/pipelines/github/digitalfabrik/integreat-cms?branch=develop ) , because it relies on the default language being set to German, but there is one other test - which if run before that test cause it to fail -, that falsely mutate settings.LANGUAGE_CODE, instead of using the settings fixture, which is not reset between tests. Also the endpoint `root_content` bakes in the default for the language_slug into the function signature, instead of loading it freshly during runtime.

### Proposed changes

- Fix root_content baking a stale settings.LANGUAGE_CODE into its default
- Use settings fixture instead of directly mutating django.conf.settings 
- explicitly set default language for the flaky test on the settings fixture
- add a 'Common Pitfalls' paragraph to the documentation

### Side effects

- None

### How to test

Observer over the next 20 or so pipeline runs, that the test wont fail again. 

### Resolved issues

Fixes: 

### DIFF
--- a/docs/src/testing.rst
+++ b/docs/src/testing.rst
@@ -267,23 +267,45 @@ Common Pitfalls
 2. **Don't use** ``serialized_rollback=True`` — it has FK ordering issues
    with PostgreSQL. Use ``load_test_data_transactional`` instead.
 
-3. **Clean up state you attach to session-scoped fixtures** — database changes
-   are rolled back after every non-transactional test, but the Python objects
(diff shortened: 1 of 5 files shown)


andrew8er (approved):
LGTM.

seluianova (approved):
nice finding!

